Issue #435, opened 2016-01-03
resurrectable species in suitability


vincent-legoll:
They don't appear in suitability report until you build the first colony whereas they should appear when you build the xenoresurrection center at their discovery site.

dbenage-cx:
Should be fixed by cebf07a

Vezzra:
Can anyone confirm that cebf07a5cdc9a158e312af872c8739514c1144ed fixed this issue? Don't have the time to test that myself right now.

vincent-legoll:
Bug is fixed, tested with today's cb44cafde562faa86fa2765ae59ee3cf28597742

Vezzra:
Thanks!

PR #559, opened 2016-03-23
Display extinct species in suitability report

dbenage-cx:
Displays extinct species in planet suitability report after construction of Xenoresurrection Lab.

Quite a few steps here, alternatives could be an empire->HasSpecial(string) or adding unlocks to specials or buildings.  I chose an approach with fewer code changes.

Issue https://github.com/freeorion/freeorion/issues/435


UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2287,8 +2287,16 @@ namespace {
         for (SpeciesManager::iterator it = species_manager.begin();
                 it != species_manager.end(); ++it)
         {
-            if (it->second && (it->second->Tags().find("CTRL_ALWAYS_REPORT") != it->second->Tags().end()))
+            if (it->second) {
+                if (it->second->Tags().find("CTRL_ALWAYS_REPORT") != it->second->Tags().end()) {
                     species_names.insert(it->first);
+                } else if (it->second->Tags().find("CTRL_EXTINCT") != it->second->Tags().end()) {
+                    std::string extinct_bld = "BLD_COL_" + it->first.substr(3);
```
geoffthemedio: not fond of specific (or generated) building names in C++ code...

dbenage-cx:
reworked to use building tags.  rebased

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2287,8 +2287,29 @@ namespace {
         for (SpeciesManager::iterator it = species_manager.begin();
                 it != species_manager.end(); ++it)
         {
-            if (it->second && (it->second->Tags().find("CTRL_ALWAYS_REPORT") != it->second->Tags().end()))
-                    species_names.insert(it->first);
+            if (it->second) {
```
geoffthemedio: prefer

```
if (!it->second)
    continue;
```

and then don't indent the following code.

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2287,8 +2287,29 @@ namespace {
...
+                const std::string& species_str = it->first;
+                const std::set<std::string>& species_tags = it->second->Tags();
+                if (species_tags.find("CTRL_ALWAYS_REPORT") != species_tags.end()) {
+                    species_names.insert(species_str);
```
geoffthemedio: could also put a continue in this if block to avoid using and else block and thus avoid indenting unnecessarily

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2287,8 +2287,29 @@ namespace {
         for (SpeciesManager::iterator it = species_manager.begin();
                 it != species_manager.end(); ++it)
         {
-            if (it->second && (it->second->Tags().find("CTRL_ALWAYS_REPORT") != it->second->Tags().end()))
-                    species_names.insert(it->first);
+            if (it->second) {
+                const std::string& species_str = it->first;
+                const std::set<std::string>& species_tags = it->second->Tags();
+                if (species_tags.find("CTRL_ALWAYS_REPORT") != species_tags.end()) {
```
geoffthemedio: Could you put some comments amongst this code to explain what you're checking / doing? it's a bit dense to read through and interpret otherwise...

default/scripting/buildings/XENORESURRECTION_LAB.focs.txt
```diff
@@ -13,6 +13,26 @@ BuildingType
         ]
     ]
     EnqueueLocation = [[ENQUEUE_BUILD_ONE_PER_PLANET]]
+    effectsgroups = [
+        EffectsGroup
+            scope = And [
+                Source
+                ContainedBy HasSpecial name = "EXTINCT_BANFORO_SPECIAL"
```
geoffthemedio: Perhaps better to put the ContainedBy condition in the activation condition, so the scope is just Source?
Also, maybe make it

```
And [
    Object id = Source.PlanetID
    HasSpecial name = "..."
]
```

so it's not checking any other objects for the presence of the special. The ContainedBy condition doesn't presently limit the containers it searches very well... and isn't necessary in a case like this where you only want the container of a single directly-referencable object (the source), and not anything that contains and object matching a more complicated condition.

dbenage-cx:
```
scope = Source
activation = And [
    Object id = Source.PlanetID
    HasSpecial name = "EXTINCT_BANFORO_SPECIAL"
]
```

Does not work, it parses ok(no errors sent) but does not activate.

geoffthemedio: Activation conditions test only the source object (and not the planet a source building is on), so you need to check something like

```
activation = ContainedBy And [
    Object id = Source.PlanetID
    HasSpecial name = "EXTINCT_BANFORO_SPECIAL"
]
```

dbenage-cx:
rebased

wheals:
Does `col_bld_gen.py` need to get updated for this?

PR #562, opened 2016-03-24
Add pedia links into beginner game hints

Ouaz:
- Replace "production" by "industry" when needed, as the related Focus is "Industry Focus"
- Add pedia links to the basic Scout, Colony Ship and Outpost Ship articles, as they are the only ones available by default for a beginner.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -3921,11 +3921,11 @@ Outposts are unmanned stations, which can be founded in places too hostile for a
 FOCUS_TITLE
 Planetary Focus
 FOCUS_TEXT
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I think you could just commit a patch like this yourself, without making a pull request...

MatGB:
I concur, stringtable cleanup is always fine (unless you're inserting Oxford Commas then I'll complain ;-) ) and it's good work.

Ouaz:
Oh OK, I thought that when it's about the EN stringtable, it was better that MatGB checks if everything is right, even if it's mainly some cleanup.

Issue #564, opened 2016-03-25
FreeOrion app hangs on program exit on OSX


Vezzra:
Commit 545449ccb53cf3bf88f17e05c6d5b50cdec1e0c2 (part of PR #519) introduced a weird bug on OSX: when exiting the application, it does not terminate cleanly, but hangs. The only way to terminate the program completely is to kill the process.

When starting the program from the command line, I get the following output in the shell immediately after program launch:

```
2016-03-25 13:27:04.712 freeorion[13473:87446] _createMenuRef called with existing principal MenuRef already associated with menu
2016-03-25 13:27:04.713 freeorion[13473:87446] (
    0   CoreFoundation                      0x96686c63 __raiseError + 195
    1   libobjc.A.dylib                     0x932e9a2a objc_exception_throw + 276
    2   CoreFoundation                      0x96686b7d +[NSException raise:format:] + 141
    3   AppKit                              0x9ad2a5b3 -[NSCarbonMenuImpl _createMenuRef] + 70
    4   AppKit                              0x9ad29d44 -[NSCarbonMenuImpl _instantiateCarbonMenu] + 155
    5   AppKit                              0x9ad29c9e -[NSCarbonMenuImpl setupCarbonMenuBar] + 40
    6   AppKit                              0x9ad281fd -[NSApplication finishLaunching] + 1398
    7   AppKit                              0x9ad278a0 -[NSApplication run] + 143
    8   AppKit                              0x9ac9cfa0 NSApplicationMain + 2082
    9   freeorion                           0x00079f96 main + 408
    10  freeorion                           0x0003e385 start + 53
)
```

freeorion.log contents:

```
2016-03-25 13:27:05.465621 [debug] Client : Logger initialized
2016-03-25 13:27:05.465786 [debug] Client : v0.4.5+ HEAD [build 2016-03-12.545449c] Xcode 0640
2016-03-25 13:27:05.465828 [debug] Client : GL Version String: 2.1 ATI-1.32.25
2016-03-25 13:27:05.465878 [debug] Client : Dependency versions from headers:
2016-03-25 13:27:05.465890 [debug] Client : Boost: 1_56
2016-03-25 13:27:05.465901 [debug] Client : FreeType: 2.3.11
2016-03-25 13:27:05.465912 [debug] Client : PNG: 1.5.12
2016-03-25 13:27:05.465922 [debug] Client : Python: 2.7.6
[...]

PR #519, opened 2016-03-12
Fix multimonitor persistence

LGM-Doyle:
This patch further improves the multi-monitor presentation.  It allows for persistence of
window sizes, when presenting freeorion in one big window across multiple monitors.  

Previously, window sizes were not persisting because the validation  of config.xml sizes was using a maximum size from a single monitor.  

This patch adds the ability for SDLGUI to report a maximum possible width/height if
all available monitors are lined up either horizontally/vertically.  That size is used for the validation.  Hence,
no possible valid window sizes are rejected.

The patch does not adjust the initial sizes chosen for windows, which are still not great for multi-monitor presentation.

PR #568, opened 2016-03-27
Move initialization of app window options out of static initialization

LGM-Doyle:
This is an attempt to address bug #564.

I can not test it as I'm running linux.  It may not help. Please allow Vezzra to test it.

Starting with PR #519 OSX crashes on exit.  PR #519 initialized SDL during the static initialization.  Prior to PR #519 initialization happened in the HumanClientApp constructor.  The initialization was moved in order to detect the screen sizes before the HumanClientApp was constructed. 

I've moved the initialization out of static initialization and into main().

Note, technically SDL requires that the main() function be replaced with SDL_main().  Freeorion was not using SDL_main() before PR #519 and is still not using SDL_main(), so I do not think that absence of SDL_main() is directly causing the new OSX crash.  However, adding SDL_main() might fix it because then all initialization will be as SDL expects.  I'm trying this fix first, because it is simpler and closest to the way things were.

### DIFF
--- a/GG/GG/SDL/SDLGUI.h
+++ b/GG/GG/SDL/SDLGUI.h
@@ -160,7 +160,8 @@ protected:
 
 private:
     void            RelayTextInput (const SDL_TextInputEvent& text, Pt mouse_pos);
-    // Bare minimum SDL video initialization to allow queries to display sizes etc.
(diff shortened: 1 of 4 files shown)


Vezzra:
Works. Thanks!

PR #569, opened 2016-03-27
Cleanup

Cjkjvfnby:
- call `write` to file in single place in code
- use dict to collect data
- move add of line breaks to template
- remove extra empty line from generator and files

### DIFF
--- a/default/scripting/buildings/colonies/col_bld_gen.py
+++ b/default/scripting/buildings/colonies/col_bld_gen.py
@@ -113,7 +113,8 @@ t_main = string.Template('''BuildingType
 
 #include "/scripting/common/misc.macros"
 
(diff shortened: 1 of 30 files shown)


default/scripting/buildings/colonies/col_bld_gen.py
```diff
@@ -113,7 +113,9 @@
 
 #include "/scripting/common/misc.macros"
 
-#include "/scripting/common/upkeep.macros"''')
+#include "/scripting/common/upkeep.macros"
+
```
adrianbroher: No need for that extra line here or in the generated output.

Cjkjvfnby:
@adrianbroher, empty line removed files updated

adrianbroher:
Looks good.

PR #575, opened 2016-03-29
Tech control icon buttons

dbenage-cx:
- [x] Rewrite button to use an existing method
- [x] Needs better artwork:
- Ship Hulls
- Ship Weapons
- Engineering (all other ship parts)

Use icons with tooltips for research screen control buttons.

Adds a new CUIToggleButton control which changes color and texture based on the checked state.
The icon sizes can be adjusted in options/config (since re-sizing window changes the layout).
Unified Tree and List view buttons to single toggle.


dbenage-cx:
tech_buttons1
tech_buttons4 
tech_buttons6

UI/CUIControls.cpp
```diff
@@ -1924,3 +1925,136 @@ void RotatingGraphic::Render() {
     glPopClientAttrib();
     glPopMatrix();
 }
+
+////////////////////////////////////////////////
+// ToggleButton
+////////////////////////////////////////////////
+ToggleButton::ToggleButton (GG::SubTexture icon_, GG::Clr color, GG::Clr checked_clr, const std::string& tooltip_str,
```
geoffthemedio: How is this different from StateButton?
Regardless, I suggest using StateButton or a derivate, and styling the buttons like the top-right menu bar buttons.

geoffthemedio:
Make the tree view icon a symbolic tree, not a literal tree... branching lines.

UI/CUIControls.cpp
adrianbroher: I agree with @geoffthemedio  here.  If you want some state use a StateButton with a new Representer.

dbenage-cx:
rebased
tech_buttons_8
Icons are not great, but they are a little more consistent.  Not real happy with the 'All' button, but it is much better than an english word.
Made the disabled/unchecked color for this control slightly darker as it was hard to see a marked difference.

Final commit is potentially a separate PR if needed.

UI/CUIControls.h
```diff
@@ -594,4 +595,48 @@ class RotatingGraphic : public GG::StaticGraphic {
...
+    virtual GG::Pt MinUsableSize(const CUIToggleButton& button) const;
+};
+
+/** \brief Toggled StateButton
+ * 
+ * This StateButton renders a SubTexture depending on the checked state.
+ * With the default single texture it renders the same icon recolored towards grey.
+ * When a second SebTexture is assigned, it will render the appropriate SubTexture.
+ */
+class CUIToggleButton : public GG::StateButton {
```
adrianbroher: What is the purpose of this class?

dbenage-cx: Poor cleanup on my part, what remained only handled sound.
Removed and handled sound within CUIToggleButton

adrianbroher: You got it wrong. What is the purpose of CUIToggleButton that can't be handled by CUIToggleButtonRepresenter?

dbenage-cx: briefly, references to textures.

From what I can understand, representers are designed for rendering a specific state, but not to hold any specific references.  Initially I had all of the render logic here, but while the buttons worked and displayed tooltip, I was not  able to produce a call to CUIToggleButtonRepresenter->Render().

Should I look into restructuring TechTreeWnd to utilize a representer?  This appeared to add more complexity since the representer needs a reference to a specific button anyway, but I am obviously missing something crucial here.

adrianbroher: > briefly, references to textures.

I don't see why this shouldn't be possible.

> From what I can understand, representers are designed for rendering a specific state, but not to hold any specific references. Initially I had all of the render logic here, but while the buttons worked and displayed tooltip, I was not able to produce a call to CUIToggleButtonRepresenter->Render().

The representers are designed to represent a button. This includes visual representation. If you need to store texture references to show a representation then the representer is a good place to store references to textures.

> Should I look into restructuring TechTreeWnd to utilize a representer?

Yes.

> This appeared to add more complexity since the representer needs a reference to a specific button anyway, but I am obviously missing something crucial here.

Why would the representer need a reference to a button?

dbenage-cx: [GG/GG/Button.h:133](https://github.com/freeorion/freeorion/blob/master/GG/GG/Button.h#L133)

>    The StateButtonRepresenter is a stub interface to implement a visual
>     representation of a StateButton instance.  A representer can be shared
>     amongst multiple instances of a StateButton, so no subclass should
>     store instance specific attributes of a state button.

I realize this is your code, but this is where I came to the conclusion not to include the references within CUIToggleButtonRepresenter.

adrianbroher: Yeah, but how is this an argument against my own request? `Should` is not `must` also the comment is written under the prerequisite that the ButtonRepresenter instance is shared, which isn't the case for the image variant. If there is a valid reason to violate the rule (here: there doesn't exist a storage for storing instance specific rendering data yet), then do it.

dbenage-cx: Will do.  Thanks for nudging me along.

dbenage-cx:
If someone more competent wants to run with this, please do so.

I am currently stalled trying to figure out how this is supposed to be implemented and find myself blankly staring at the screen.

Will close and open as a feature request tomorrow if no one objects.

adrianbroher:
> I am currently stalled trying to figure out how this is supposed to be implemented and find myself blankly staring at the screen.

What is the problem you're facing?

dbenage-cx:
TLDR; I do not know the goal I'm heading for, so I end up with a pile of rubbish.

Here is what I have inferred so far:
1. Inherit from StateButtonRepresenter
2. Store texture/color in representer
3. When utilized, the representer should not need to refer to a StateButton object
So I currently need the following attributes for the representer:

```
SubTexture    m_checked_icon
SubTexture    m_unchecked_icon
Clr           m_checked_clr
Clr           m_unchecked_clr
string        m_tooltip
Rect          m_render_area
bool          m_is_checked
bool          m_is_mouseover
```

The bool attributes could be dropped if we build separate checked and unchecked representers.  I stopped when I started to add Rect to this class, but potentially requires any other attributes needed during `Render()`, since I should not use `Render(button)`.

The only use of inheriting StateButtonRepresenter that I can see, is so a StateButton can reference it.  Without the shared nature the other representers have, and needing to duplicate the attributes of StateButton, I do not understand the purpose in this class.

Maybe I misunderstood the third point, and I should still be using `Render(button)`.  This leads me to needing to refer to something like `Representer[i]->Render(button[j])`.  But again, why?  Where is the efficiency loss in adding the render logic to an inherited StateButton.

I am trying not to dispute points others bring up, since I do not get to code much and I do not want to be a distraction.  However, I am at a complete loss as to why this needs to be such a complicated process and what benefit it could have.

adrianbroher:
> TLDR; I do not know the goal I'm heading for, so I end up with a pile of rubbish.

Well, not problem.  I will try to explain the reason and motivation behind that.

> 1. Inherit from StateButtonRepresenter

That's okay.

> 1. Store texture/color in representer

That's okay.

> 1. When utilized, the representer should not need to refer to a StateButton object

That's not okay.  It is okay to query attributes from the GG::StateButton like 'are you currently checked?', 'Where are you located?' because that's what a GG::StateButton logical does, it places an area and links a boolean variable to that area, which now is represented in some way (like an image or a check mark or an "Checked", "Unchecked" text) and allows the user to interact in some way with that area to change the boolean state.

So while the analysis for the required attributes would be correct under the assumption that 'we should never refer to the GG::StateButton', but that's not true.

The attributes:

> Rect          m_render_area
> bool          m_is_checked
> bool          m_is_mouseover

can (and should) be queried from from the GG::StateButton, because those describe the logical state of the button.

The representation however (things like 'the checked state is represented by image.checked.png' and 'the unchecked one by image.unchecked.png' or 'play beep.ogg on the uncheck to check state change') do not belong to the logical StateButton.  They could be replaced without the StateButton acting any different, just looking (be represented in) different (way).  For example take the CUICheckBoxRepresenter, CUIRadioRepresenter or CUITabRepresenter. They all represent very different button types, but the logical behaviour behind all of them is consistent.

And that's the reason why I added this in the first place.  Previously we had a lot of different StateButton, StateButton-ish thing, that were not a StateButton at all and all of them behaved a bit different.  Splitting out the representation from the behaviour removed those differences.  Unfortunately I'm still not done with the whole idea of making the widgets consistent (otherwise such horrible Workarounds like SettableInWindowCUIButton or CUIArrowButton wouldn't even exist) so it may be not that obvious to some other programmer.

> Where is the efficiency loss in adding the render logic to an inherited StateButton.

The is not about efficiency but first an foremost about establishing consistent behaviour across all widgets of a certain type.

> I am trying not to dispute points others bring up, since I do not get to code much and I do not want to be a distraction. However, I am at a complete loss as to why this needs to be such a complicated process and what benefit it could have.

It's okay to ask.

dbenage-cx:
Should we be assigning specific attributes to a renderer?  It seems like it breaks with the spirit of that class sub-set and inconsistent.

The textures might be held either by an inherited StateButton, mapped to the state by StateButton, or stored elsewhere and passed as an argument to Render.

Color is not as big of an issue, since it is likely to be shared among different buttons, but could follow the same approach along with tooltip.

dbenage-cx:
hopefully more inline with standard.

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,139 @@ void RotatingGraphic::Render() {
...
+{ return m_checked_color; }
+
+void CUIToggleRepresenter::SetTooltip(const std::string& tooltip)
+{ m_tooltip = tooltip; }
+
+/** \brief Set the SubTexture for both checked and unchecked states
+ * 
+ * @param[in] icon SubTexture to assign
+ */
+void CUIToggleRepresenter::SetIcon(boost::shared_ptr<GG::SubTexture> icon) {
```
geoffthemedio: Put the comments in the header...

Why do you need this function, when the next two together do the same thing?
Why do you need these three functions, when constructor parameters could be used to set the icons?

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,139 @@ void RotatingGraphic::Render() {
...
+{ m_unchecked_color = c; }
+
+void CUIToggleRepresenter::SetCheckedColor(GG::Clr c)
+{ m_checked_color = c; }
+
+const bool CUIToggleRepresenter::HasSharedColor() const
+{ return (m_unchecked_color == m_checked_color); }
+
+void CUIToggleRepresenter::Render(const CUIToggleButton& button) const {
+    bool render_checked(!HasSharedIcon() && (!button.Checked()^button.IsMouseover()));  // use specific icon if not shared, swap on mouseover
```
geoffthemedio: why ^ ?

geoffthemedio: Why do you need this testing of whether there is a shared icon? Why not just render whichever texture is appropriate for the state of the button (checked or unchecked)?

UI/CUIControls.h
```diff
@@ -594,4 +595,61 @@ class RotatingGraphic : public GG::StaticGraphic {
...
+/** \brief Represents a CUIToggleButton
+ * 
+ * Renders a SubTexture depending on the checked state of the CUIToggleButton.
+ * When the \a button has single texture and is unchecked, it renders the same icon recolored towards grey.
+ * When a second SubTexture is assigned to the \a button, it will render the appropriate SubTexture.
+ */
+class CUIToggleRepresenter : public GG::StateButtonRepresenter {
+public:
+    CUIToggleRepresenter(const std::string& tooltip, boost::shared_ptr<GG::SubTexture> icon, const GG::Clr& highlight_clr); ///< ctor
+    const bool                              HasSharedIcon() const;
```
geoffthemedio: why "const" bool?

UI/TechTreeWnd.cpp
```diff
@@ -80,25 +81,38 @@ namespace {
         return true;
     }
 
-    struct ToggleCategoryFunctor {
-        ToggleCategoryFunctor(TechTreeWnd* tree_wnd, const std::string& category) : m_tree_wnd(tree_wnd), m_category(category) {}
-        void operator()() { m_tree_wnd->ToggleCategory(m_category); }
+
+    struct SetCategoryViewFunctor {
+        SetCategoryViewFunctor(TechTreeWnd* tree_wnd, const std::string& category) :
+                               m_tree_wnd(tree_wnd), m_category(category) {}
```
geoffthemedio: indent initializer list one tab (4 spaces), not (necessarily) lined up with parameters.
preferably put each initializer on its own line, and the {} on a line after, indented to match the function name.

UI/TechTreeWnd.cpp
```diff
@@ -80,25 +81,38 @@ namespace {
...
+        void operator()(bool b) { b ? m_tree_wnd->ShowCategory(m_category) : m_tree_wnd->HideCategory(m_category); }
```
geoffthemedio: might want a safety check on m_tree_wnd != 0

dbenage-cx:
rebased

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,119 @@ void RotatingGraphic::Render() {
     glPopClientAttrib();
     glPopMatrix();
 }
+
+////////////////////////////////////////////////
+// CUIToggleButton
+////////////////////////////////////////////////
+CUIToggleRepresenter::CUIToggleRepresenter(boost::shared_ptr<GG::SubTexture> icon,
+                                           const GG::Clr& highlight_clr) :
+                                           m_unchecked_icon(icon),
```
geoffthemedio: indent initializer list one tab from class name, not to be in line with parameters

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,119 @@ void RotatingGraphic::Render() {
...
+void CUIToggleButton::LClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_keys)
+{
+    if (!Disabled()) {
+        SetCheck(!Checked());
+        CheckedSignal(Checked());
+    }
+}
+
+void CUIToggleButton::Render()
+{
```
geoffthemedio: { to previous line

UI/CUIControls.h
```diff
@@ -594,4 +595,58 @@ class RotatingGraphic : public GG::StaticGraphic {
...
+                         boost::shared_ptr<GG::SubTexture> checked_icon, const GG::Clr& checked_clr); ///< ctor
+    //@}
+
+    /** \name Mutators */ //@{
+    virtual void                            RenderUnchecked(const CUIToggleButton& button) const;
+    virtual void                            RenderChecked(const CUIToggleButton& button) const;
+    //@}
+
+private:
+    boost::shared_ptr<GG::SubTexture>       m_unchecked_icon;
```
geoffthemedio: deindent member names one tab

UI/CUIControls.h
```diff
@@ -594,4 +595,58 @@ class RotatingGraphic : public GG::StaticGraphic {
...
+class CUIToggleRepresenter : public GG::StateButtonRepresenter {
+public:
+    /** \name Structors */ //@{
+    CUIToggleRepresenter(boost::shared_ptr<GG::SubTexture> icon, const GG::Clr& highlight_clr); ///< ctor
+    CUIToggleRepresenter(boost::shared_ptr<GG::SubTexture> unchecked_icon, const GG::Clr& unchecked_clr,
+                         boost::shared_ptr<GG::SubTexture> checked_icon, const GG::Clr& checked_clr); ///< ctor
+    //@}
+
+    /** \name Mutators */ //@{
+    virtual void                            RenderUnchecked(const CUIToggleButton& button) const;
```
geoffthemedio: deindent substantially; no need to go so far over for these

UI/CUIControls.h
```diff
@@ -594,4 +595,58 @@ class RotatingGraphic : public GG::StaticGraphic {
...
+ * When the \a button has single texture and is unchecked, it renders the same icon recolored towards grey.
+ * If a second SubTexture is assigned to the \a button, it will render the appropriate SubTexture.
+ * When a mouse pointer hovers over this button, the opposing SubTexture and color are used and highlighted
+ */
+class CUIToggleRepresenter : public GG::StateButtonRepresenter {
+public:
+    /** \name Structors */ //@{
+    CUIToggleRepresenter(boost::shared_ptr<GG::SubTexture> icon, const GG::Clr& highlight_clr); ///< ctor
+    CUIToggleRepresenter(boost::shared_ptr<GG::SubTexture> unchecked_icon, const GG::Clr& unchecked_clr,
+        boost::shared_ptr<GG::SubTexture> checked_icon, const GG::Clr& checked_clr); ///< ctor
```
geoffthemedio: parameters should be indented to match the first to the right of the (

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,118 @@ void RotatingGraphic::Render() {
...
+    SetBrowseInfoWnd(boost::shared_ptr<GG::BrowseInfoWnd>(new TextBrowseWnd(tooltip, "")));
+    GG::Connect(CheckedSignal, &PlayButtonCheckedSound, -1);
+}
+
+void CUIToggleButton::SetTooltip(const std::string& tooltip)
+{ SetBrowseText(tooltip); }
+
+void CUIToggleButton::LClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_keys)
+{
```
geoffthemedio: to previous line

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,118 @@ void RotatingGraphic::Render() {
...
+CUIToggleRepresenter::CUIToggleRepresenter(boost::shared_ptr<GG::SubTexture> icon,
+    const GG::Clr& highlight_clr) :
+    m_unchecked_icon(icon),
+    m_checked_icon(icon),
+    m_unchecked_color(highlight_clr),
+    m_checked_color(highlight_clr)
+{}
+
+CUIToggleRepresenter::CUIToggleRepresenter(boost::shared_ptr<GG::SubTexture> unchecked_icon,
+    const GG::Clr& unchecked_clr,
```
geoffthemedio: indent to match (

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,118 @@ void RotatingGraphic::Render() {
...
+{
+    if (!Disabled()) {
+        SetCheck(!Checked());
+        CheckedSignal(Checked());
+    }
+}
+
+void CUIToggleButton::Render() {
+    if (m_representer)
+        (Checked()^IsMouseover()) ? m_representer->RenderChecked(*this) : m_representer->RenderUnchecked(*this);
```
geoffthemedio: really, what's with the ^ ?

geoffthemedio:
should probably put the tech category renaming stuff in a separate pull request

UI/CUIControls.cpp
dbenage-cx: uses opposite icon on mouseover

geoffthemedio: add a comment explaining it then...

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,117 @@ void RotatingGraphic::Render() {
...
+
+CUIToggleButton::CUIToggleButton(const std::string& tooltip, boost::shared_ptr<CUIToggleRepresenter> representer) :
+    StateButton("", ClientUI::GetFont(), GG::FORMAT_NONE, GG::CLR_ZERO, representer),
+    m_representer(representer)
+{
+    SetBrowseInfoWnd(boost::shared_ptr<GG::BrowseInfoWnd>(new TextBrowseWnd(tooltip, "")));
+    GG::Connect(CheckedSignal, &PlayButtonCheckedSound, -1);
+}
+
+void CUIToggleButton::SetTooltip(const std::string& tooltip)
```
adrianbroher: Never used, unnecessary thin wrapper for SetBrowseText.  Should be removed.

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,117 @@ void RotatingGraphic::Render() {
...
+{ SetBrowseText(tooltip); }
+
+void CUIToggleButton::LClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_keys) {
```
adrianbroher: Already implemented in GG::StateButton.  This implementation derives in behaviour.  Should be removed.

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,117 @@ void RotatingGraphic::Render() {
...
+    if (!Disabled()) {
+        SetCheck(!Checked());
+        CheckedSignal(Checked());
+    }
+}
+
+void CUIToggleButton::Render() {
```
adrianbroher: Already implemented in GG::StateButton.  This implementation derives in behaviour.  Should be removed.

Overwrite the CUIToggleButtonRepresenter::Render(comst GG::StateButton&) method instead to call the RenderChecked/RenderUnchecked methods instead.

UI/CUIControls.h
```diff
@@ -594,4 +595,58 @@ class RotatingGraphic : public GG::StaticGraphic {
...
+    /** \name Mutators */ //@{
+    void                                    SetTooltip(const std::string& tooltip);  ///< Set the text for the tooltip of this button.
+    virtual void                            Render();
+    //@}
+
+protected:
+    virtual void                            LClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_keys);
+
+private:
+    boost::shared_ptr<CUIToggleRepresenter> m_representer;
```
adrianbroher: Already exists in GG::StateButton.  No need to store a reference to the representer at all.

UI/CUIControls.h
```diff
@@ -594,4 +595,58 @@ class RotatingGraphic : public GG::StaticGraphic {
...
+ * 
+ * This StateButton can utilize a separate SubTexture and
+ * With the default single texture it renders the same icon recolored towards grey.
+ * If a second SebTexture is assigned, it will render the appropriate SubTexture.
+ * When a mouse pointer hovers over this button, the opposing SubTexture and color are used and highlighted
+ */
+class CUIToggleButton : public GG::StateButton {
+public:
+    /** \name Structors */ //@{
+    CUIToggleButton(const std::string& tooltip, boost::shared_ptr<CUIToggleRepresenter> representer); ///< ctor
```
adrianbroher: No need for the tooltip parameter.  Call SetBrowseText where necessary.

UI/OptionsWnd.cpp
```diff
@@ -489,65 +489,66 @@ OptionsWnd::OptionsWnd():
     GG::Connect(show_font_texture_button->LeftClickedSignal, &ShowFontTextureWnd);
 
     CreateSectionHeader(current_page, 0, UserString("OPTIONS_FONT_SIZES"));
-    IntOption(current_page, 0, "UI.font-size",       UserString("OPTIONS_FONT_TEXT"));
```
adrianbroher, line 492: Completely unrelated cosmetic change.  Next time please make a separate PR for things like that.

UI/TechTreeWnd.cpp
```diff
@@ -238,139 +255,155 @@ class TechTreeWnd::TechTreeControls : public CUIWnd {
     // calculated by SizeMove, and stored, so that start and end positions don't need to be recalculated each
     // time Render is called.
 
-    std::vector<CUIButton*>             m_category_buttons;
-    std::map<TechStatus, CUIButton*>    m_tech_status_buttons;
-    CUIButton*                          m_list_view_button;
-    CUIButton*                          m_tree_view_button;
+    boost::shared_ptr<GG::SubTexture>            GetManagedSubTexture(const std::string& fn);  ///< return a shared_ptr of a managed subtexture for tech controls
+    CUIToggleButton*                             m_view_type_button;
```
adrianbroher: Pointer to GG::ToggleButton to all of the variables in this attribute block is more than sufficient.

UI/TechTreeWnd.cpp
```diff
@@ -238,139 +255,155 @@ class TechTreeWnd::TechTreeControls : public CUIWnd {
...
+    CUIToggleRepresenter*                        m_view_type_representer;
+    CUIToggleButton*                             m_all_cat_button;
+    CUIToggleRepresenter*                        m_all_cat_representer;
```
adrianbroher: No need to store a pointer to a CUIToggleRepresenter, just initialize it and pass it to the CUIToggleButton.

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,117 @@ void RotatingGraphic::Render() {
...
+
+    // render border
+    AngledCornerRectangle(border_ul, border_lr, bg_clr, border_clr, CUIBUTTON_ANGLE_OFFSET, BORDER_THICKNESS);
+
+    // render icon
+    glColor(icon_clr);
+    m_unchecked_icon->OrthoBlit(icon_ul, icon_lr);
+}
+
+void CUIToggleRepresenter::RenderChecked(const CUIToggleButton& button) const {
```
adrianbroher: RenderChecked and RenderUnchecked are so similar that they should be one method.

UI/CUIControls.cpp
```diff
@@ -29,6 +30,9 @@ namespace {
     void PlayButtonCheckSound()
     { Sound::GetSound().PlaySound(GetOptionsDB().Get<std::string>("UI.sound.button-click"), true); }
 
+    void PlayButtonCheckedSound(bool b)
```
adrianbroher: Duplicate of PlayButtonCheckSound.  Can be removed.

UI/CUIControls.h
```diff
@@ -594,4 +595,58 @@ class RotatingGraphic : public GG::StaticGraphic {
     float   m_phase_offset;
 };
 
+class CUIToggleButton;
```
adrianbroher: Forward declaration not needed at all.

UI/TechTreeWnd.cpp
```diff
@@ -238,139 +255,155 @@ class TechTreeWnd::TechTreeControls : public CUIWnd {
...
-        GG::Clr background_clr = ClientUI::CategoryColor(cats[i]);
-        GG::Clr border_clr     = background_clr;
-        AdjustBrightness(background_clr, -50);
-        m_category_buttons.push_back(new CUIButton(UserString(cats[i]), background_clr, border_clr));
-        AttachChild(m_category_buttons.back());
+        GG::Clr icon_clr = ClientUI::CategoryColor(cats[i]);
+        boost::shared_ptr<GG::SubTexture> icon = boost::make_shared<GG::SubTexture>(ClientUI::CategoryIcon(cats[i]));
+        m_cat_representers[cats[i]] = new CUIToggleRepresenter(icon, icon_clr);
+        m_cat_buttons[cats[i]] = new CUIToggleButton(UserString(cats[i]),
+                                                     boost::shared_ptr<CUIToggleRepresenter>(m_cat_representers[cats[i]]));
```
adrianbroher: Instead of `boost::shared_ptr<Type>(new Type(arg1, arg2))` you can use `boost::make_shared<Type>(arg1, arg2)` to reduce code duplication.

UI/CUIControls.cpp
```diff
@@ -1924,3 +1928,117 @@ void RotatingGraphic::Render() {
...
+    glColor(icon_clr);
+    m_checked_icon->OrthoBlit(icon_ul, icon_lr);
+}
+
+CUIToggleButton::CUIToggleButton(const std::string& tooltip, boost::shared_ptr<CUIToggleRepresenter> representer) :
+    StateButton("", ClientUI::GetFont(), GG::FORMAT_NONE, GG::CLR_ZERO, representer),
+    m_representer(representer)
+{
+    SetBrowseInfoWnd(boost::shared_ptr<GG::BrowseInfoWnd>(new TextBrowseWnd(tooltip, "")));
+    GG::Connect(CheckedSignal, &PlayButtonCheckedSound, -1);
```
adrianbroher: Use the `void CUIToggleButtonRepresenter::OnChecked(bool checked) const` method instead to call PlayButtonCheck.  See `void CUICheckBoxRepresenter::OnChecked(bool checked) const` (UI/CUIControls.cpp:363) as an example.

dbenage-cx:
Ty adrianbroher, I'll work on making these changes.

When you have the some free time, I have a couple of issues I will bring up where I probably missed something.

UI/CUIControls.h
dbenage-cx: Not sure what my alternative is here?  Without a forward, I will see multiple

```
...error: ‘CUIToggleButton’ does not name a type
virtual void RenderUnchecked(const CUIToggleButton& button) const;
                                   ^
```

And the reverse for CUIToggleButtonRepresenter if I put CUIToggleButton first.

UI/CUIControls.cpp
geoffthemedio: presumably this was due to not know how to use boost::bind to connect signals to functions with different signatures (return + parameter types)

UI/CUIControls.h
dbenage-cx: Just to note, and I will try this again, I was not successful using this before.
I suspected that in passing through GG::StateButton::m_representer, I discarded the attributes from the CUIToggleButtonRepresenter (hence the textures were never passed).

Storing the inherited form seemed to resolve this, but I will try to find where my fault was.

UI/CUIControls.h
adrianbroher: The `virtual void Render{Un,}Checked(const CUIToggleButton& button) const;` don't need a CUIToggleButton as parameter, using a GG::StateButton is sufficient.

If you take a look at the other comment this will become more clear.

> Overwrite the CUIToggleButtonRepresenter::Render(comst GG::StateButton&) method instead.

For reference you can take a look at UI/CUIControls.cpp:416. There I overwrite the `GG::StateButtonRepresenter:Render(comst GG::StateButton&) const` to do the rendering.  You should do the same for the `CUIToggleButtonRepresenter`.

UI/CUIControls.cpp
dbenage-cx: Indeed, I did not realize it as an option here. Ty

dbenage-cx:
Rebased.

Control window width would increase to 150 on game load, if it was smaller.  Since it is a minor edit and only surfaces as an issue here, not sure if you want as separate  PR or squashed?

geoffthemedio:
Fails to link for me...

```
TechTreeWnd.obj : error LNK2001: unresolved external symbol "class GG::TextureManager & __cdecl GG::GetTextureManager(void)" (?GetTextureManager@GG@@YAAAVTextureManager@1@XZ)
```

Not sure why you want to add GetManagedSubTexture to TechTreeControls, but even if it has a use, it doesn't belong in that class. It should also be implemented like ClientUI::BuildingIcon or similar functions.

adrianbroher:
> Fails to link for me...

That's interesting because it works for me.  The linker probably the linker complains because the `GetManagedSubTexture` is not declared in `TechTreeControls`.

> It should also be implemented like ClientUI::BuildingIcon or similar functions.

I would say a function inside the anonymous namespace in `TechTreeWnd.cpp` should be fine,  adding this function for the ClientUI just bloats the API for no reason and the function is just used locally.

UI/OptionsWnd.cpp
adrianbroher, line 492: Completely unrelated to the PR itself and it's by no means a request to change this now, but it maybe better to explain WHY I don't want to see such changes in PRs:  It took literally 10 reads for me until I realized that you've added a new OptionsDB entry.

 This whole ' tabularizing' entries is just some horrible habit we should ban for this very reason from our code base, it makes changes much harder to understand. As I said this is not against your change specifically @dbenage-cx.

geoffthemedio:
Implemented, not located (ie. doesn't need to be in ClientUI)

eg. just call

```
ClientUI::GetTexture(ArtDir() / graphic_name, true);
```

without an explicit call to GG::GetTextureManager()

UI/TechTreeWnd.cpp
```diff
@@ -80,22 +81,40 @@ namespace {
...
-        ToggleCategoryFunctor(TechTreeWnd* tree_wnd, const std::string& category) : m_tree_wnd(tree_wnd), m_category(category) {}
-        void operator()() { m_tree_wnd->ToggleCategory(m_category); }
+
+    struct SetCategoryViewFunctor {
+        SetCategoryViewFunctor(TechTreeWnd* tree_wnd, const std::string& category) :
+            m_tree_wnd(tree_wnd),
+            m_category(category)
+        {}
+        void operator()(bool b) {
+            if (m_tree_wnd != 0)
```
geoffthemedio: just need

```
if (!m_tree_wnd) return;
```

UI/TechTreeWnd.cpp
```diff
@@ -80,22 +81,40 @@ namespace {
...
-    struct ToggleAllCategoriesFunctor {
-        ToggleAllCategoriesFunctor(TechTreeWnd* tree_wnd) : m_tree_wnd(tree_wnd) {}
-        void operator()() { m_tree_wnd->ToggleAllCategories(); }
+    struct SetAllCategoryViewsFunctor {
+        SetAllCategoryViewsFunctor(TechTreeWnd* tree_wnd) :
+            m_tree_wnd(tree_wnd)
+        {}
+        void operator()(bool b) {
+            if (m_tree_wnd != 0)
+                b? m_tree_wnd->ShowAllCategories() : m_tree_wnd->HideAllCategories();
```
geoffthemedio: if (!m_tree_wnd)
space between b and ?

dbenage-cx: assuming you meant `(!m_tree_wnd == 0)`

geoffthemedio: no; I mean if (!a_pointer)

dbenage-cx: pushed.
Am I missing some other point here?  These only proceed if the window doesn't exist (?), so the buttons will never trigger the functionality.

adrianbroher:
At least from my side all issues are addressed or are there any other pointers for improvement @freeorion/developers?

I'm fetching the most recent version now and give it a test and if there aren't any other complains I will rebase and merge.  Replacing the currently suggested graphics  can be done IMO another time.

dbenage-cx:
Currently: Clicking the buttons will not trigger as expected (except for list/tree view).  I am waiting for @geoffthemedio input on why they should only trigger on null ptr.  I did not want to push the issue and have tried to figure out what the intent is but come up blank, [comment thread](https://github.com/freeorion/freeorion/pull/575#discussion_r59015296)

adrianbroher:
@dbenage-cx I assume his intent was to remove the comparision with 0.  The suggested condition is wrong as you already found out.  So `if(m_the_window)` is the correct implementation, not `if(!m_the_window)`.

It also looks like the button don't honour the tooltip delay.

dbenage-cx:
@adrianbroher ty, now uses option defined delay

adrianbroher:
Works for me. I will rebase and merge then.

PR #577, opened 2016-03-30
Windows Wide Path Fixes

geoffthemedio:
Currently, setting the resource directory to a path with non-ascii characters will cause errors in path resolution on Windows. This attempts to fix it, hopefully in a way that doesn't break anything else. Some testing reports from others (Windows or non-Windows) would be useful.

### DIFF
--- a/GG/src/GIL/extension/io/png_dynamic_io.hpp
+++ b/GG/src/GIL/extension/io/png_dynamic_io.hpp
@@ -120,6 +120,30 @@ inline void png_read_image(const std::string& filename,any_image<Images>& im) {
 }
 
 /// \ingroup PNG_IO
+/// \brief reads a PNG image into a run-time instantiated image
+template <typename Images>
+inline void png_read_image(FILE* file,any_image<Images>& im) {
+    detail::png_reader_dynamic m(file);
+    m.read_image(im);
+}
+
+/// \ingroup PNG_IO
+/// \brief reads a PNG image into a run-time instantiated image
+template <typename Images>
+inline void png_read_image(const boost::filesystem::path& path,any_image<Images>& im) {
+#if defined (_WIN32)
+    boost::filesystem::path::string_type path_native = path.native();
+    FILE* file = _wfopen(path_native.c_str(), L"rb");
+    detail::png_reader_dynamic m(file);
+    m.read_image(im);
+    fclose(file);
+#else
+    std::string filename = path.generic_string());
+    png_read_image(filename.c_str(),im);
+#endif
+}
+
+/// \ingroup PNG_IO
 /// \brief Saves the currently instantiated view to a png file specified by the given png image file name.
 /// Throws std::ios_base::failure if the currently instantiated view type is not supported for writing by the I/O extension 
 /// or if it fails to create the file.
--- a/GG/src/GIL/extension/io/png_dynamic_io.hpp
+++ b/GG/src/GIL/extension/io/png_dynamic_io.hpp
@@ -138,7 +138,7 @@ inline void png_read_image(const boost::filesystem::path& path,any_image<Images>
     m.read_image(im);
     fclose(file);
 #else
-    std::string filename = path.generic_string());
+    std::string filename = path.generic_string();
     png_read_image(filename.c_str(),im);
 #endif
 }
--- a/GG/src/GIL/extension/io/png_io.hpp
+++ b/GG/src/GIL/extension/io/png_io.hpp
@@ -193,7 +193,7 @@ inline void png_read_and_convert_image(const boost::filesystem::path& path,Image
     gil::png_read_and_convert_image(file,im);
     fclose(file);
 #else
-    std::string filename = path.generic_string());
+    std::string filename = path.generic_string();
     png_read_and_convert_image(filename.c_str(),im);
 #endif
 }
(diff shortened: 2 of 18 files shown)


GG/src/GIL/extension/io/png_dynamic_io.hpp
```diff
@@ -120,6 +120,30 @@ inline void png_read_image(const std::string& filename,any_image<Images>& im) {
...
+template <typename Images>
+inline void png_read_image(const boost::filesystem::path& path,any_image<Images>& im) {
+#if defined (_WIN32)
+    boost::filesystem::path::string_type path_native = path.native();
+    FILE* file = _wfopen(path_native.c_str(), L"rb");
+    detail::png_reader_dynamic m(file);
+    m.read_image(im);
+    fclose(file);
+#else
+    std::string filename = path.generic_string());
```
dbenage-cx: excess closing bracket

GG/src/GIL/extension/io/png_io.hpp
```diff
@@ -176,6 +176,29 @@ inline void png_read_and_convert_image(const std::string& filename,Image& im) {
...
+/// \brief Allocates a new image whose dimensions are determined by the given png image file path, loads and color-converts the pixels into it.
+template <typename Image>
+inline void png_read_and_convert_image(const boost::filesystem::path& path,Image& im) {
+#if defined (_WIN32)
+    boost::filesystem::path::string_type path_native = path.native();
+    FILE* file = _wfopen(path_native.c_str(), L"rb");
+    gil::png_read_and_convert_image(file,im);
+    fclose(file);
+#else
+    std::string filename = path.generic_string());
```
dbenage-cx: excess closing bracket

GG/src/Texture.cpp
```diff
@@ -252,11 +253,33 @@ void Texture::OrthoBlit(const Pt& pt1, const Pt& pt2, const GLfloat* tex_coords/
...
+        throw BadFile("Texture \"file\" \"" + path.generic_string() + "\" is not a file");
+    }
+
+    // convert path into UTF-8 format filename string
+#if defined (_WIN32)
+    boost::filesystem::path::string_type path_native = path.native();
+    std::string filename;
+    utf8::utf16to8(path_native.begin(), path_native.end(), std::back_inserter(filename));
+#else
+    std::string filename = path.generic_string());
```
dbenage-cx: excess closing bracket

dbenage-cx:
Still building windows.

For Fedora en_US.UTF8: 
~~unable to save `[error] Client : SaveFileDialog.cpp:763 : SaveFileDialog::Confirm: problem with result: /opt/projects/freeorion/userdir/save/save-20160330_142941`~~ Issue in master too, checking
But overwrite existing files when saving will work.

created a sub-directory within userdir: zzzПрыгать
able to select directory from options dialog(only the first 3 chars render, others are blank)
auto-save, loading, and overwriting saves works.
changing to locale ru_RU.iso88595, russian, or ru_RU.utf8 had no noticable difference

edit: custom ship designs save ok, same with changing the user-dir to the above sub-dir

dbenage-cx:
win7p en_US;English:
- created userdir sub-dir with same name: zzzПрыгать
  can over-write saves and load games from this directory(still only renders first 3 chars)

Attempted to add the above sub-dir to Directories.cpp:258 and save file as unicode.  Compiler does not seem to handle this (C4566) with current code page.  Build completes but game immediately crashes at Directoies.cpp:246, likely not a valid test without changing system codepage.

If no one else is able to test, I will see if cloning the instance is possible before attempting locale/codepage/user name changes.

adrianbroher:
Fails to compile because of the 'excess brackets' @dbenage-cx  mentioned.

```
In file included from /home/adrian/Projekte/freeorion/freeorion.git/GG/src/GUI.cpp:44:0:
/home/adrian/Projekte/freeorion/freeorion.git/GG/src/GIL/extension/io/png_io.hpp: In Funktion »void boost::gil::png_read_and_convert_image(const boost::filesystem::path&, Image&)«:
/home/adrian/Projekte/freeorion/freeorion.git/GG/src/GIL/extension/io/png_io.hpp:196:49: Fehler: expected »,« or »;« before »)« token
     std::string filename = path.generic_string());
```

geoffthemedio:
Hopefully fixed extra brackets...

geoffthemedio:
@dbenage-cx

I wouldn't expect to be able to hard-code a non-wide character string into the C++ source code explicitly. I think you could do a series of escape codes, though.

But I don't think being able to do that is necessary... the default relative pathes in C++ can remain ascii characters... All I want to achieve is that user directories or custom-set resource or save directories with wide characters work.

adrianbroher:
Using the latest PR code I would say it works as expected. Replaced the 'default' asset directory with 'データ' and I have yet to encounter missing resources even after changing all resource references in the config file.  The only issue so far is that the 'データ' string isn't rendered in the resource path input field or the file selection dialog.  The characters itself are there so it seems like a issue with the font don't containing the glyphs (nothing to worry about).

geoffthemedio:
@adrianbroher On what OS are you testing?
Assuming not Windows, is the this branch's behaviour any different from master?

adrianbroher:
> On what OS are you testing?

Fedora Linux with de_DE.UTF-8 as LANG

> Assuming not Windows, is the this branch's behaviour any different from master?

I would need to test that.

adrianbroher:
> Assuming not Windows, is the this branch's behaviour any different from master?

Comparing to a190c02 (common ancestor between this PR and master) I don't see any difference in behaviour.

PR #578, opened 2016-03-30
STATIC_CODE_ANALYSIS_FIXES

matthoppe:
Minor code adjustments to handle null pointer dereferences and some copy and paste errors.


GG/src/GroupBox.cpp
```diff
@@ -59,9 +59,11 @@ GroupBox::GroupBox(X x, Y y, X w, Y h, const std::string& label, const boost::sh
     m_label(label.empty() ? 0 : GUI::GetGUI()->GetStyleFactory()->NewTextControl(label, m_font, m_text_color, FORMAT_LEFT | FORMAT_TOP)),
     m_set_client_corners_equal_to_box_corners(false)
 {
-    m_label->MoveTo(Pt(X0, -m_font->Lineskip()));
-    m_label->MoveTo(Pt(X1, m_font->Lineskip()));
-    AttachChild(m_label);
+    if (!m_label) {
+        m_label->MoveTo(Pt(X0, -m_font->Lineskip()));
```
geoffthemedio, line 63: huh? If the pointer is null, only then dereference it?

UI/MapWnd.cpp
```diff
@@ -364,7 +364,7 @@ class MapWnd::MapScaleLine : public GG::Control {
             if (TemporaryPtr<const Fleet> fleet = GetFleet(*it)) {
                 if (fleet->Speed() > 20)
                     fixed_distances.insert(fleet->Speed());
-                for (std::set< int >::iterator ship_it = fleet->ShipIDs().begin(); ship_it != fleet->ShipIDs().end(); ship_it++){
+                for (std::set< int >::iterator ship_it = fleet->ShipIDs().begin(); ship_it != fleet->ShipIDs().end(); ++ship_it){
```
geoffthemedio: does this actually matter on any modern compiler?

UI/MapWnd.cpp
```diff
@@ -1915,7 +1915,7 @@ void MapWnd::RenderMovementLine(const MapWnd::MovementLineData& move_line, float
 
     // set vertex positions to outline a quad for each move line vertex
     for (std::vector<MovementLineData::Vertex>::const_iterator verts_it = vertices.begin();
-         verts_it != vertices.end(); ++verts_it)
+         verts_it != --vertices.end(); ++verts_it)
```
geoffthemedio: This seems wrong...

UI/ProductionWnd.cpp
```diff
@@ -531,7 +531,9 @@ namespace {
         }
         if (m_block_size_selector) {
             m_block_size_selector->MoveTo(GG::Pt(left, GG::Y(MARGIN)));
-            left += m_quantity_selector->Width();
+	    if (m_quantity_selector) {
```
geoffthemedio: fix indenting

combat/CombatEvents.cpp
```diff
@@ -54,7 +54,8 @@ AttackEvent::AttackEvent() :
     round(-1),
     attacker_id(INVALID_OBJECT_ID),
     target_id(INVALID_OBJECT_ID),
-    damage(0.0f)
+    damage(0.0f),
+    attacker_owner_id(-1)
```
geoffthemedio: use ALL_EMPIRES

combat/CombatEvents.cpp
```diff
@@ -110,7 +111,8 @@ void AttackEvent::serialize<freeorion_xml_iarchive>(freeorion_xml_iarchive& ar,
 //////////////////////////////////////////
 IncapacitationEvent::IncapacitationEvent() :
     bout(-1),
-    object_id(INVALID_OBJECT_ID)
+    object_id(INVALID_OBJECT_ID),
+    object_owner_id(-1)
```
geoffthemedio: ALL_EMPIRES

universe/Condition.cpp
```diff
@@ -2350,10 +2350,12 @@ void HasTag::Eval(const ScriptingContext& parent_context,
         // evaluate number limits once, use to match all candidates
         TemporaryPtr<const UniverseObject> no_object;
         ScriptingContext local_context(parent_context, no_object);
-        if (!m_name)
+        if (!m_name){
```
geoffthemedio: space before {

universe/Condition.cpp
```diff
@@ -4957,13 +4959,13 @@ void DesignHasPartClass::Eval(const ScriptingContext& parent_context,
 }
 
 bool DesignHasPartClass::RootCandidateInvariant() const
-{ return (m_low->RootCandidateInvariant() && m_high->RootCandidateInvariant()); }
+{ return (m_low && m_high ? m_low->RootCandidateInvariant() && m_high->RootCandidateInvariant() : false); }
```
geoffthemedio: need more complicated logic here to handle presence or absense of either independnetly. and if both aren't there, then it should return true.

GG/src/GroupBox.cpp
matthoppe, line 63: Oops... I'll nix the reverse logic.

UI/MapWnd.cpp
matthoppe: Later on we do

``` cpp
if (dots_added_to_buffer >= BUFFER_CAPACITY)
            break; // can't fit any more!

        // get next two vertices
        const MovementLineData::Vertex& vert1 = *verts_it;
        verts_it++;
        const MovementLineData::Vertex& vert2 = *verts_it;
```

verts_it can become end() which is adjacent memory not belonging to the vector. --end() ensures we are one off from the edge of the vector.

universe/ShipDesign.cpp
```diff
@@ -495,7 +495,9 @@ const int       ShipDesign::MAX_ID            = 2000000000;
 ShipDesign::ShipDesign() :
     m_id(INVALID_OBJECT_ID),
     m_name(),
+    m_description(),
     m_designed_on_turn(UniverseObject::INVALID_OBJECT_AGE),
+    m_designed_by_empire(-1),
```
geoffthemedio: ALL_EMPIRES

util/Order.cpp
```diff
@@ -104,7 +104,9 @@ void RenameOrder::ExecuteImpl() const {
 // CreateFleetOrder
 ////////////////////////////////////////////////
 NewFleetOrder::NewFleetOrder() :
-    Order()
+    Order(),
+    m_fleet_names(),
+    m_system_id(-1)
```
geoffthemedio: INVALID_OBJECT_ID

UI/MapWnd.cpp
geoffthemedio: I understand your motivation, but:
1) wouldn't that also cause problems if the vector is empty?
2) there is a test earlier in the function to make sure there are an odd number of entries in the vector...

https://github.com/matthoppe/freeorion/blob/null_pointer_fixes/UI/MapWnd.cpp#L1893

matthoppe: Agreed, the check above should handle those edge cases nicely.

universe/Condition.cpp
matthoppe: Right ... how about 

``` cpp
bool DesignHasPartClass::RootCandidateInvariant() const {
    bool ret = true;

    if (m_low) {
        ret = m_low->RootCandidateInvariant();
    }
    if (m_high) {
        ret = m_high->RootCandidateInvariant();
    }
    if (m_low && m_high) {
        ret = m_low->RootCandidateInvariant() && m_high->RootCandidateInvariant();
    }

    return ret;
}
```

geoffthemedio:
could you rebase this without all the merge commits?

universe/Condition.cpp
```diff
@@ -4956,14 +4958,27 @@ void DesignHasPartClass::Eval(const ScriptingContext& parent_context,
     }
 }
 
-bool DesignHasPartClass::RootCandidateInvariant() const
-{ return (m_low->RootCandidateInvariant() && m_high->RootCandidateInvariant()); }
+bool DesignHasPartClass::RootCandidateInvariant() const { 
+    bool ret = true;
+
+    if (m_low) {
```
geoffthemedio: rework this to avoid repeat calls to ->RootCandidateInvariant(). eg.

```
return (!m_low || m_low->RootCandidateInvariant()) && (!m_high || m_high->RootCandidateInvariant());
```

universe/Condition.cpp
```diff
@@ -4956,14 +4958,27 @@ void DesignHasPartClass::Eval(const ScriptingContext& parent_context,
...
+    if (m_low && m_high) {
+        ret = m_low->RootCandidateInvariant() && m_high->RootCandidateInvariant();  
+    }
+
+    return ret;
+}
 
 bool DesignHasPartClass::TargetInvariant() const
-{ return (m_low->TargetInvariant() && m_high->TargetInvariant()); }
+{ return (m_low && m_high ? m_low->TargetInvariant() && m_high->TargetInvariant() : false); }
```
geoffthemedio: these also need fixing as was the case for the above

util/Order.cpp
```diff
@@ -104,7 +104,9 @@ void RenameOrder::ExecuteImpl() const {
 // CreateFleetOrder
 ////////////////////////////////////////////////
 NewFleetOrder::NewFleetOrder() :
-    Order()
+    Order(INVALID_OBJECT_ID),
```
geoffthemedio: the Order constructor parameters is an empire id, not an object id, and the default is already to use ALL_EMPIRES

matthoppe:
I've squashed my commits.

Vezzra:
@matthoppe, I've added you to the credits. Is there any real name you want to supply so we can add that to your entry?

matthoppe:
No, that looks good.

PR #579, opened 2016-03-31
Re-allow saving to new file, fix file ext

dbenage-cx:
Saving to new file broke throughout several different commits.(win and rhel)

When attempting to save to new filename, is_regular_file returns false since file does not exist.

.sav extension was no longer added by default, files not saved with this extension do not display for file loading dialog.


MatGB:
I've compiled this and it fixes the problem for me, as it's backend changes I can't even read the code let alone review it, but I think it ought to go in ASAP.

PR #580, opened 2016-04-02
Refactor ResourcesAI.py

LGM-Doyle:
This PR is a major refactor of ResourcesAI.py. 

I was trying to determine why the AI appears to be precessing around a threshold when toggling between research and industry focus every third turn.  It turns out adj_round 3 intentionally trades 3 industry points for 1 research point when targeting low a research/industry ratio.

The PR fixes one small bug preventing more than one high industry planet from being prevented from switching to research and otherwise doesn't intentionally change any logic.

The refactor is mostly of three parts.
1.   Rewrote set_planet_resource_foci into a function that calls 5 sub-functions: use_planet_growth_specials, use_planet_production_and_research_specials, set_planet_protection_foci, set_planet_happiness_foci (,a placeholder) and set_planet_industry_and_research_foci.  They all do the obvious duty in the same manner as before.  They do not access any non-local variables other than fo, the freeorion global.
2.  Replaced the file scope variables new_targets, current_focus, current_output  and planet_map with a focus_manager.  The focus_manager includes planet info for each planet encompassing the previous variables.  focus_manager manages all updating or changes of state of the focus setting.  Most significantly focus_manager uses bake_future_focus to change a planet's status from raw to baked as its focus is set.  Previously, the behavior was implicit and if a planet was set, but not removed from the list it would be overwritten later.  All setting happens in this one function as opposed to throughout the file.
3.  Moved all reporting into a Reporter object.  The reporter.capture_section_info is called after each stage of the set_planet_resource_foci to capture the information to make table sections: Unfocusable, Specials, Protection, Happiness and Typical.

I think the refactored code is clearer to read.
The code is more local in scope and hence easier to modify in future.

### DIFF
--- a/default/AI/ResourcesAI.py
+++ b/default/AI/ResourcesAI.py
@@ -11,10 +11,7 @@ from freeorion_tools import tech_is_complete
 
 resource_timer = Timer('timer_bucket')
 
-newTargets = {}
-currentFocus = {}
-currentOutput = {}
-planetMap = {}
+#Local Constants
 IFocus = FocusType.FOCUS_INDUSTRY
 RFocus = FocusType.FOCUS_RESEARCH
 MFocus = FocusType.FOCUS_MINING  # not currently used in content
@@ -29,17 +26,22 @@ limitAssessments = False
 
 lastFociCheck = [0]
 
+new_targets = {}
+current_focus = {}
+current_output = {}
+planet_map = {}
+
 
 def _fill_data_dicts(planet_ids):
     universe = fo.getUniverse()
-    newTargets.clear()
-    currentFocus.clear()
-    currentOutput.clear()
-    planets = [(pid, planetMap[pid]) for pid in planet_ids]
+    new_targets.clear()
+    current_focus.clear()
+    current_output.clear()
+    planets = [(pid, planet_map[pid]) for pid in planet_ids]
     for pid, planet in planets:
-        currentFocus[pid] = planet.focus
-        currentOutput.setdefault(pid, {})[IFocus] = planet.currentMeterValue(fo.meterType.industry)
-        currentOutput[pid][RFocus] = planet.currentMeterValue(fo.meterType.research)
+        current_focus[pid] = planet.focus
+        current_output.setdefault(pid, {})[IFocus] = planet.currentMeterValue(fo.meterType.industry)
+        current_output[pid][RFocus] = planet.currentMeterValue(fo.meterType.research)
         if IFocus in planet.availableFoci and planet.focus != IFocus:
             fo.issueChangeFocusOrder(pid, IFocus)  # may not be able to take, but try
     universe.updateMeterEstimates(planet_ids)
@@ -47,13 +49,13 @@ def _fill_data_dicts(planet_ids):
         itarget = planet.currentMeterValue(fo.meterType.targetIndustry)
         rtarget = planet.currentMeterValue(fo.meterType.targetResearch)
         if planet.focus == IFocus:
-            newTargets.setdefault(pid, {})[IFocus] = (itarget, rtarget)
-            newTargets.setdefault(pid, {})[GFocus] = [0, rtarget]
+            new_targets.setdefault(pid, {})[IFocus] = (itarget, rtarget)
+            new_targets.setdefault(pid, {})[GFocus] = [0, rtarget]
         else:
-            newTargets.setdefault(pid, {})[IFocus] = (0, 0)
-            newTargets.setdefault(pid, {})[GFocus] = [0, 0]
-        # if currentFocus[pid] == MFocus:
-        # newTargets[pid][MFocus] = (mtarget, rtarget)
+            new_targets.setdefault(pid, {})[IFocus] = (0, 0)
+            new_targets.setdefault(pid, {})[GFocus] = [0, 0]
+        # if current_focus[pid] == MFocus:
+        # new_targets[pid][MFocus] = (mtarget, rtarget)
         if RFocus in planet.availableFoci and planet.focus != RFocus:
             fo.issueChangeFocusOrder(pid, RFocus)  # may not be able to take, but try
     universe.updateMeterEstimates(planet_ids)
@@ -62,22 +64,22 @@ def _fill_data_dicts(planet_ids):
         itarget = planet.currentMeterValue(fo.meterType.targetIndustry)
         rtarget = planet.currentMeterValue(fo.meterType.targetResearch)
         if planet.focus == RFocus:
-            newTargets.setdefault(pid, {})[RFocus] = (itarget, rtarget)
-            newTargets[pid][GFocus][0] = itarget
+            new_targets.setdefault(pid, {})[RFocus] = (itarget, rtarget)
+            new_targets[pid][GFocus][0] = itarget
         else:
-            newTargets.setdefault(pid, {})[RFocus] = (0, 0)
-            newTargets[pid][GFocus][0] = 0
-        if can_focus and currentFocus[pid] != planet.focus:
-            fo.issueChangeFocusOrder(pid, currentFocus[pid])  # put it back to what it was
+            new_targets.setdefault(pid, {})[RFocus] = (0, 0)
+            new_targets[pid][GFocus][0] = 0
+        if can_focus and current_focus[pid] != planet.focus:
+            fo.issueChangeFocusOrder(pid, current_focus[pid])  # put it back to what it was
     universe.updateMeterEstimates(planet_ids)
     # Protection focus will give the same off-focus Industry and Research targets as Growth Focus
     for pid, planet in planets:
-        newTargets[pid][PFocus] = newTargets[pid][GFocus]
+        new_targets[pid][PFocus] = new_targets[pid][GFocus]
 
 
 def get_resource_target_totals(empirePlanetIDs):
-    pp = sum(x.currentMeterValue(fo.meterType.targetIndustry) for x in planetMap.values())
-    rp = sum(x.currentMeterValue(fo.meterType.targetResearch) for x in planetMap.values())
+    pp = sum(x.currentMeterValue(fo.meterType.targetIndustry) for x in planet_map.values())
+    rp = sum(x.currentMeterValue(fo.meterType.targetResearch) for x in planet_map.values())
     _fill_data_dicts(empirePlanetIDs)
     return pp, rp
 
@@ -134,34 +136,34 @@ def weighted_sum_output(outputs):
 
 def assess_protection_focus(pid):
     """Return True if planet should use Protection Focus"""
-    this_planet = planetMap[pid]
+    this_planet = planet_map[pid]
     sys_status = foAI.foAIstate.systemStatus.get(this_planet.systemID, {})
     threat_from_supply = (0.25 * foAI.foAIstate.empire_standard_enemy_rating *
                           min(2, len(sys_status.get('enemies_nearly_supplied', []))))
     print "Planet %s has regional+supply threat of %.1f" % ('P_%d<%s>' % (pid, this_planet.name), threat_from_supply)
     regional_threat = sys_status.get('regional_threat', 0) + threat_from_supply
     if not regional_threat:  # no need for protection
-        if currentFocus[pid] == PFocus:
+        if current_focus[pid] == PFocus:
             print "Advising dropping Protection Focus at %s due to no regional threat" % this_planet
         return False
-    cur_prod_val = weighted_sum_output([currentOutput[pid][IFocus], currentOutput[pid][RFocus]])
-    target_prod_val = max(map(weighted_sum_output, [newTargets[pid][IFocus], newTargets[pid][RFocus]]))
-    prot_prod_val = weighted_sum_output(newTargets[pid][PFocus])
+    cur_prod_val = weighted_sum_output([current_output[pid][IFocus], current_output[pid][RFocus]])
+    target_prod_val = max(map(weighted_sum_output, [new_targets[pid][IFocus], new_targets[pid][RFocus]]))
+    prot_prod_val = weighted_sum_output(new_targets[pid][PFocus])
     local_production_diff = 0.8 * cur_prod_val + 0.2 * target_prod_val - prot_prod_val
     fleet_threat = sys_status.get('fleetThreat', 0)
     # TODO: relax the below rejection once the overall determination of PFocus is better tuned
     if not fleet_threat and local_production_diff > 8:
-        if currentFocus[pid] == PFocus:
+        if current_focus[pid] == PFocus:
             print "Advising dropping Protection Focus at %s due to excessive productivity loss" % this_planet
         return False
     local_p_defenses = sys_status.get('mydefenses', {}).get('overall', 0)
     # TODO have adjusted_p_defenses take other in-system planets into account
-    adjusted_p_defenses = local_p_defenses * (1.0 if currentFocus[pid] != PFocus else 0.5)
+    adjusted_p_defenses = local_p_defenses * (1.0 if current_focus[pid] != PFocus else 0.5)
     local_fleet_rating = sys_status.get('myFleetRating', 0)
     combined_local_defenses = sys_status.get('all_local_defenses', 0)
     my_neighbor_rating = sys_status.get('my_neighbor_rating', 0)
     neighbor_threat = sys_status.get('neighborThreat', 0)
-    safety_factor = 1.2 if currentFocus[pid] == PFocus else 0.5
+    safety_factor = 1.2 if current_focus[pid] == PFocus else 0.5
     cur_shield = this_planet.currentMeterValue(fo.meterType.shield)
     max_shield = this_planet.currentMeterValue(fo.meterType.maxShield)
     cur_troops = this_planet.currentMeterValue(fo.meterType.troops)
@@ -172,18 +174,18 @@ def assess_protection_focus(pid):
     use_protection = True
     reason = ""
     if (fleet_threat and  # i.e., an enemy is sitting on us
-              (currentFocus[pid] != PFocus or  # too late to start protection TODO: but maybe regen worth it
+              (current_focus[pid] != PFocus or  # too late to start protection TODO: but maybe regen worth it
               # protection forcus only useful here if it maintains an elevated level
               all([AIDependencies.PROT_FOCUS_MULTIPLIER * a <= b for a, b in def_meter_pairs]))):
         use_protection = False
         reason = "A"
-    elif ((currentFocus[pid] != PFocus and cur_shield < max_shield - 2 and
+    elif ((current_focus[pid] != PFocus and cur_shield < max_shield - 2 and
                not tech_is_complete(AIDependencies.PLANET_BARRIER_I_TECH)) and
               (cur_defense < max_defense - 2 and not tech_is_complete(AIDependencies.DEFENSE_REGEN_1_TECH)) and
               (cur_troops < max_troops - 2)):
         use_protection = False
         reason = "B1"
-    elif ((currentFocus[pid] == PFocus and cur_shield*AIDependencies.PROT_FOCUS_MULTIPLIER < max_shield-2 and
+    elif ((current_focus[pid] == PFocus and cur_shield*AIDependencies.PROT_FOCUS_MULTIPLIER < max_shield-2 and
                not tech_is_complete(AIDependencies.PLANET_BARRIER_I_TECH)) and
               (cur_defense*AIDependencies.PROT_FOCUS_MULTIPLIER < max_defense-2 and
                    not tech_is_complete(AIDependencies.DEFENSE_REGEN_1_TECH)) and
@@ -200,7 +202,7 @@ def assess_protection_focus(pid):
         use_protection = False
         reason = "E"
     elif (safety_factor * regional_threat <= combined_local_defenses and
-              (currentFocus[pid] != PFocus or
+              (current_focus[pid] != PFocus or
               (0.5 * safety_factor * regional_threat <= local_fleet_rating and
                    fleet_threat == 0 and neighbor_threat < combined_local_defenses and
                    local_production_diff > 5))):
@@ -212,7 +214,7 @@ def assess_protection_focus(pid):
           local_production_diff > 5):
         use_protection = False
         reason = "G"
-    if use_protection or currentFocus[pid] == PFocus:
+    if use_protection or current_focus[pid] == PFocus:
         print ("Advising %sProtection Focus (reason %s) for planet %s, with local_prod_diff of %.1f, comb. local"
                " defenses %.1f, local fleet rating %.1f and regional threat %.1f, threat sources: %s") % (
             ["dropping ", ""][use_protection], reason, this_planet, local_production_diff, combined_local_defenses,
@@ -228,7 +230,7 @@ def use_planet_growth_specials(empirePlanetIDs, newFoci):
             for special in [aspec for aspec in AIDependencies.metabolismBoostMap.get(metab, []) if aspec in ColonisationAI.available_growth_specials]:
                 rankedPlanets = []
                 for pid in ColonisationAI.available_growth_specials[special]:
-                    planet = planetMap[pid]
+                    planet = planet_map[pid]
                     cur_focus = planet.focus
                     pop = planet.currentMeterValue(fo.meterType.population)
                     if (pop > metabIncPop - 2 * planet.size) or (GFocus not in planet.availableFoci):  # not enough benefit to lose local production, or can't put growth focus here
@@ -254,10 +256,10 @@ def use_planet_growth_specials(empirePlanetIDs, newFoci):
                         newFoci[spPID] = GFocus
                         if spPID in empirePlanetIDs:
                             del empirePlanetIDs[empirePlanetIDs.index(spPID)]
-                            print "%s focus of planet %s (%d) at Growth Focus" % (["set", "left"][cur_focus == GFocus], planetMap[spPID].name, spPID)
+                            print "%s focus of planet %s (%d) at Growth Focus" % (["set", "left"][cur_focus == GFocus], planet_map[spPID].name, spPID)
                         break
                     else:
-                        print "failed setting focus of planet %s (%d) at Growth Focus; focus left at %s" % (planetMap[spPID].name, spPID, planetMap[spPID].focus)
+                        print "failed setting focus of planet %s (%d) at Growth Focus; focus left at %s" % (planet_map[spPID].name, spPID, planet_map[spPID].focus)
 
 
 def use_planet_production_and_research_specials(empirePlanetIDs, newFoci):
@@ -265,7 +267,7 @@ def use_planet_production_and_research_specials(empirePlanetIDs, newFoci):
     universe = fo.getUniverse()
     already_have_comp_moon = False
     for pid in empirePlanetIDs:
-        planet = planetMap[pid]
+        planet = planet_map[pid]
         if "COMPUTRONIUM_SPECIAL" in planet.specials and RFocus in planet.availableFoci and not already_have_comp_moon:
             curFocus = planet.focus
             newFoci[pid] = RFocus
@@ -276,7 +278,7 @@ def use_planet_production_and_research_specials(empirePlanetIDs, newFoci):
                     universe.updateMeterEstimates(empirePlanetIDs)
             if curFocus == RFocus or result == 1:
                 already_have_comp_moon = True
-                print "%s focus of planet %s (%d) (with Computronium Moon) at Research Focus" % (["set", "left"][curFocus == RFocus], planetMap[pid].name, pid)
+                print "%s focus of planet %s (%d) (with Computronium Moon) at Research Focus" % (["set", "left"][curFocus == RFocus], planet_map[pid].name, pid)
                 if pid in empirePlanetIDs:
                     del empirePlanetIDs[empirePlanetIDs.index(pid)]
                 continue
@@ -289,7 +291,7 @@ def use_planet_production_and_research_specials(empirePlanetIDs, newFoci):
                 if result == 1:
                     universe.updateMeterEstimates(empirePlanetIDs)
             if curFocus == IFocus or result == 1:
-                print "%s focus of planet %s (%d) (with Honeycomb) at Industry Focus" % (["set", "left"][curFocus == IFocus], planetMap[pid].name, pid)
+                print "%s focus of planet %s (%d) (with Honeycomb) at Industry Focus" % (["set", "left"][curFocus == IFocus], planet_map[pid].name, pid)
                 if pid in empirePlanetIDs:
                     del empirePlanetIDs[empirePlanetIDs.index(pid)]
                 continue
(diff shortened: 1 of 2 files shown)


default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,261 @@
...
+            focusable = pinfo.planet.currentMeterValue(fo.meterType.targetPopulation) > 0
+            if not focusable:
+                self.baked_planet_info[pid] = self.raw_planet_info.pop(pid)
+
+
+    def bake_future_focus(self, pid, focus, update=True):
+        """Set the focus and moves it from the raw list to the baked list of planets
+        Return success or failure"""
+        pinfo = self.raw_planet_info.get(pid)
+        success = (pinfo != None and
```
Cjkjvfnby: https://www.python.org/dev/peps/pep-0008/#id45:
Comparisons to singletons like None should always be done with is or is not , never the equality operators. 

PS. Just run code checker to fix all style issues at once.

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,261 @@
...
+                   (pinfo.current_focus == focus
+                    or (focus in pinfo.planet.availableFoci
+                        and fo.issueChangeFocusOrder(pid, focus) == 1)))
```
Cjkjvfnby: `fo.issueChangeFocusOrder(pid, focus)` return int. But in python you can treat it as bool. 

`and fo.issueChangeFocusOrder(pid, focus) == 1` => `and fo.issueChangeFocusOrder(pid, focus)`

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,261 @@
...
+        # Protection focus will give the same off-focus Industry and Research targets as Growth Focus
+        for pid, pinfo, planet in planet_infos:
+            pinfo.possible_output[PFocus] = pinfo.possible_output[GFocus]
+
+class Reporter(object):
+    """Reporter contains some file scope functions to report"""
+
+    def __init__(self, focus_manager):
+        self.focus_manager = focus_manager
+        self.sections = list()
```
Cjkjvfnby: `self.sections = []`

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,261 @@
...
+        self.captured_ids = set()
+
+    def capture_section_info(self, title):
+        """Grab ids of all the newly baked planets
+        """
+        new_captured_ids = set(self.focus_manager.baked_planet_info)
+        new_ids = new_captured_ids - self.captured_ids
+        if len(new_ids) > 0:
```
Cjkjvfnby: PEP8:

```
For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

Yes: if not seq:
     if seq:

No: if len(seq):
    if not len(seq):
```

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,261 @@
...
+        warnings = {}
+        #TODO combine this with previous table to reduce report duplication?
+        print "Planet Resources Foci:"
+        for planetID in empirePlanetIDs:
+            planet = universe.getPlanet(planetID)
+            planetPopulation = planet.currentMeterValue(fo.meterType.population)
+            maxPop = planet.currentMeterValue(fo.meterType.targetPopulation)
+            if maxPop < 1 and planetPopulation > 0:
+                warnings[planet.name] = (planetPopulation, maxPop)
+            statusStr = "  ID: " + str(planetID) + " Name: % 18s -- % 6s % 8s " % (str(planet.name), str(planet.size), str(planet.type))
```
Cjkjvfnby: Mix of formatting and string concatenation in single line?  

`str(planet.name), str(planet.size), str(planet.type)` you use `%s`, no need to wrap arguments to string.

LGM-Doyle:
Thanks for the review.

What code checker are you using?  

I used pylint.  I ignored variable naming violations, missing docstrings, too many branches and too many statements as there were already violations in the old code. It did not pick up the violations that you pointed out, use of None etc.

The last string formatting code was in the legacy code and was only moved.  I didn't write it and wish to dodge the blame.  I did fix it as requested.

Cjkjvfnby:
I use embedded checker in PyCharm (Community edition is free).  It highlight warnings in code.

| The PR fixes one small bug preventing more than one high industry planet from being prevented from switching to research and otherwise doesn't intentionally change any logic.

Can you rephrase phrase in description because I don sure that I understand it properly

You can use `insert code` button at the title of edit area to wrap code names in description it will add them visual highlighting.

See also [Styling with Markdown is supported](https://guides.github.com/features/mastering-markdown/) link at the bottom of edit area, there is a lot of useful text formatting features. You can use real list instead of manually creating them. And make sublists inside them.

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,259 @@
...
+            self.baked_planet_info[pid] = self.raw_planet_info.pop(pid)
+
+            if update and pinfo.current_focus != focus:
+                # TODO determine if this should update all planets (for reporting) or just the remaining unplaced planets
+                universe = fo.getUniverse()
+                universe.updateMeterEstimates(self.raw_planet_info.keys())
+                itarget = pinfo.planet.currentMeterValue(fo.meterType.targetIndustry)
+                rtarget = pinfo.planet.currentMeterValue(fo.meterType.targetResearch)
+                pinfo.possible_output[focus] = (itarget, rtarget)
+        return success
```
Cjkjvfnby: PyCharm generates me next hint:

`Inferred type: (self: PlanetFocusManager, pid: Any, focus: Any, update: bool) -> Union[bool, int]`

Just wrap `success` to `bool` before return it.

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,259 @@
...
+        success = (pinfo is not None and
+                   (pinfo.current_focus == focus
+                    or (focus in pinfo.planet.availableFoci
+                        and fo.issueChangeFocusOrder(pid, focus))))
+        if success:
+            pinfo.future_focus = focus
+            self.baked_planet_info[pid] = self.raw_planet_info.pop(pid)
+
+            if update and pinfo.current_focus != focus:
+                # TODO determine if this should update all planets (for reporting) or just the remaining unplaced planets
```
Cjkjvfnby: I don't get this `TODO`, we have only one planet here.

Can you clarify update in docstring, why we did not update manager?

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,259 @@
...
+                universe.updateMeterEstimates(self.raw_planet_info.keys())
+                itarget = pinfo.planet.currentMeterValue(fo.meterType.targetIndustry)
+                rtarget = pinfo.planet.currentMeterValue(fo.meterType.targetResearch)
+                pinfo.possible_output[focus] = (itarget, rtarget)
+        return success
+
+
+    def calculate_planet_infos(self, pids):
+        """ Calculates for each possible focus the target output of each planet and stores it in planet info
+        It excludes baked planets from
```
Cjkjvfnby: from what?

default/AI/ResourcesAI.py
```diff
@@ -211,30 +381,227 @@ def assess_protection_focus(pid):
...
+            if ratio < 1.1 and foAI.foAIstate.aggression > fo.aggression.cautious:  # but wait, RP is still super cheap relative to PP, maybe will take more RP
+                if priorityRatio < 1.5 * (curTargetRP / (curTargetPP + 0.0001)):  # yeah, really a glut of RP, stop taking RP
+                    break
+            else:  # RP not super cheap & we have enough, stop taking it
+                break
+        II, IR = pinfo.possible_output[IFocus]
+        RI, RR = pinfo.possible_output[RFocus]
+        # if focus_manager.current_focus[pid] == MFocus:
+        # II = max( II, focus_manager.possible_output[MFocus][0] )
+        if (not do_research and (
```
LGM-Doyle: Commit [0e8bd8b676f45b 'Fix overly indented code in set_planet_industry_and_foci' ](https://github.com/freeorion/freeorion/pull/580/commits/0e8bd8b676f45b05d084149d841880f1bf5f4d01) fixes the bug.

Previously, the code to prevent planets satisfying the conditions now at line 568-573 was indented to the depth inside the `if` of the line  574 `if not printedHeader`.  It was only executed when the header was printed, instead of for each planet that satisfied the entry conditions.

The important (non reporting) line is the `continue` on line 583, which needs to run to skip the next planet.  In the buggy code it was falling through and having its focus set th RFocus.

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,259 @@
...
+        It excludes baked planets from
+        Note: The results will not be strictly correct if any planets have global effects
+        """
+        #TODO this function depends on the specific rule that off-focus meter value are always the minimum value
+        universe = fo.getUniverse()
+        unbaked_pids = [pid for pid in pids if not pid in self.baked_planet_info]
+        planet_infos = [(pid, self.all_planet_info[pid], self.all_planet_info[pid].planet) for pid in unbaked_pids]
+        for pid, pinfo, planet in planet_infos:
+            if IFocus in planet.availableFoci and planet.focus != IFocus:
+                fo.issueChangeFocusOrder(pid, IFocus)  # may not be able to take, but try
```
Cjkjvfnby, line 114: This action is not expected for me inside method method with name `calculate_planet_infos`

IMHO `fo.issueChangeFocusOrder` should be called in one place.  That we need manager for.  I found total 6 call of it in AI code.

default/AI/ResourcesAI.py
```diff
@@ -211,30 +381,227 @@ def assess_protection_focus(pid):
...
+            if not printedHeader:
+                printedHeader = True
+                print "Rejecting further Research Focus choices as too expensive:"
+                print "%34s|%20s|%15s |%15s|%15s |%15s |%15s" % ("                      Planet ", " current RP/PP ", " current target RP/PP ", "current Focus ", "  rejectedFocus ", " rejected target RP/PP ", "rejected RP-PP EQF")
+            oldFocus = pinfo.current_focus
+            cPP, cRP = pinfo.current_output[IFocus], pinfo.current_output[RFocus]
+            otPP, otRP = pinfo.possible_output[oldFocus]
+            ntPP, ntRP = pinfo.possible_output[RFocus]
+            print "pID (%3d) %22s | c: %5.1f / %5.1f | cT: %5.1f / %5.1f |  cF: %8s | nF: %8s | cT: %5.1f / %5.1f | %.2f" % (pid, pinfo.planet.name, cRP, cPP, otRP, otPP, fociMap.get(oldFocus, 'unknown'), fociMap[RFocus], ntRP, ntPP, ratio)
+            continue  # RP is getting too expensive, but might be willing to still allocate from a planet with less PP to lose
```
LGM-Doyle, line 579: Commit [0e8bd8b676f45b 'Fix overly indented code in set_planet_industry_and_foci' ](https://github.com/freeorion/freeorion/pull/580/commits/0e8bd8b676f45b05d084149d841880f1bf5f4d01) fixes the bug.

Previously, the code to prevent planets satisfying the conditions now at line 568-573 was indented to the depth inside the `if` of the line  574 `if not printedHeader`.  It was only executed when the header was printed, instead of for each planet that satisfied the entry conditions.

The important (non reporting) line is the `continue` on line 583, which needs to run to skip the next planet.  In the buggy code it was falling through and having its focus set th RFocus.

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,259 @@
...
+        for pid, pinfo, planet in planet_infos:
+            if IFocus in planet.availableFoci and planet.focus != IFocus:
+                fo.issueChangeFocusOrder(pid, IFocus)  # may not be able to take, but try
+
+        universe.updateMeterEstimates(unbaked_pids)
+        for pid, pinfo, planet in planet_infos:
+            itarget = planet.currentMeterValue(fo.meterType.targetIndustry)
+            rtarget = planet.currentMeterValue(fo.meterType.targetResearch)
+            if planet.focus == IFocus:
+                pinfo.possible_output[IFocus] = (itarget, rtarget)
```
Cjkjvfnby, line 121: - [x] `pinfo.possible_output[IFocus]` and `pinfo.possible_output[GFocus]` data collections look pretty familiar for me. But in one case you use list and in other you use tuple.  Please use make them same type for consistency. Tuple is preferred.

default/AI/ResourcesAI.py
LGM-Doyle, line 114: The three issueChangeFocusOrder calls in calculate_planet_info are from the legacy code.  

The intent of them is to calculate the changed industry and research production when some planets are changed to one of the special focuses (growth, computronic, etc.) that have global effects.  The third one in that function restores the initial state on entering the function.  

The only issueChangeFocusOrder that has a lasting effect after generate_resource_orders exits is the one on line 90 in bake_future_focus.

Cjkjvfnby:
- [ ] Format docstrings according to https://www.python.org/dev/peps/pep-0257/

default/AI/ResourcesAI.py
LGM-Doyle, line 121: This distinction was in the legacy code to facilitate changing the GFocus list in the second block of code.  I've changed it to a tuple.

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,258 @@
...
-        # newTargets[pid][MFocus] = (mtarget, rtarget)
-        if RFocus in planet.availableFoci and planet.focus != RFocus:
-            fo.issueChangeFocusOrder(pid, RFocus)  # may not be able to take, but try
-    universe.updateMeterEstimates(planet_ids)
-    for pid, planet in planets:
-        can_focus = planet.currentMeterValue(fo.meterType.targetPopulation) > 0
+class PlanetFocusInfo(object):
+    """ The current, possible and future foci and output of one planet."""
+    def __init__(self, planet):
+        self.planet = planet
```
Morlic-fo, line 46: Most of the python code takes ids instead of actual objects, probably we should do so here as well? @Cjkjvfnby opinion?

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,258 @@
...
+        self.current_focus = planet.focus
+        self.current_output = {}
+        self.current_output[IFocus] = planet.currentMeterValue(fo.meterType.industry)
+        self.current_output[RFocus] = planet.currentMeterValue(fo.meterType.research)
+        self.possible_output = {}
```
Morlic-fo, line 50: This is weird for me (eventhough consistend with legacy code). From `current_output[Focus]` I would expect the output (probably both research and production) if the planet has the `Focus` activated.

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,258 @@
...
+
+    def __init__(self):
+        universe = fo.getUniverse()
+
+        resource_timer.start("getPlanets")
+        planet_ids = list(PlanetUtilsAI.get_owned_planets_by_empire(universe.planetIDs))
+
+        # resource_timer.start("Shuffle")
+        # shuffle(generalPlanetIDs)
+
```
Morlic-fo: `# shuffle(planet_ids)`?

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,258 @@
...
+        resource_timer.start("Targets")
+        planet_infos = [PlanetFocusInfo(universe.getPlanet(pid)) for pid in  planet_ids]
+        self.all_planet_info = {}
+        self.all_planet_info.update(zip(planet_ids, planet_infos))
+
```
Morlic-fo, line 68: I guess can be replaced with
`self.all_planet_info = {pid: PlanetFocusInfo(universe.getPlanet(pid)) for pid in planet_ids}`?

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,258 @@
...
+        self.raw_planet_info = dict(self.all_planet_info)
+        self.baked_planet_info = dict()
+
```
Morlic-fo, line 71: `baked` seems weird. Maybe `processed`?

Also `= dict() --> = {}`

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,258 @@
...
+        for pid, pinfo in self.raw_planet_info.items():
+            focusable = pinfo.planet.currentMeterValue(fo.meterType.targetPopulation) > 0
+            if not focusable:
```
Morlic-fo: Why is this a check for `targetPopulation` instead of `population`? (If you took it from existing code and don't know, please add `TODO why not current population`).

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,258 @@
...
+            pinfo.future_focus = focus
+            self.baked_planet_info[pid] = self.raw_planet_info.pop(pid)
+
+            if update and pinfo.current_focus != focus:
+                # TODO determine if this should update all planets (for reporting) or just the remaining unplaced planets
+                universe = fo.getUniverse()
+                universe.updateMeterEstimates(self.raw_planet_info.keys())
+                itarget = pinfo.planet.currentMeterValue(fo.meterType.targetIndustry)
+                rtarget = pinfo.planet.currentMeterValue(fo.meterType.targetResearch)
+                pinfo.possible_output[focus] = (itarget, rtarget)
```
Morlic-fo, line 97: I'm unsure about the logic of the entire block here.
First, you remove `pid` out of `raw_planet_info`, then you update meter estimates but only for the `raw_planet_info`, thus not `pid`, yet try to update `pinfo` which is paired with `pid`?

default/AI/ResourcesAI.py
```diff
@@ -29,98 +38,258 @@
...
+                pinfo.possible_output[GFocus] = rtarget
+            else:
+                pinfo.possible_output[IFocus] = (0, 0)
+                pinfo.possible_output[GFocus] = 0
+            if RFocus in planet.availableFoci and planet.focus != RFocus:
+                fo.issueChangeFocusOrder(pid, RFocus)  # may not be able to take, but try
+
+        universe.updateMeterEstimates(unbaked_pids)
+        for pid, pinfo, planet in planet_infos:
+            can_focus = planet.currentMeterValue(fo.meterType.targetPopulation) > 0
```
Morlic-fo: That is not the only place this check is needed. Maybe define it as function in `PlanetFocusInfo`?

default/AI/ResourcesAI.py
```diff
@@ -211,30 +380,227 @@ def assess_protection_focus(pid):
...
                " defenses %.1f, local fleet rating %.1f and regional threat %.1f, threat sources: %s") % (
-            ["dropping ", ""][use_protection], reason, this_planet, local_production_diff, combined_local_defenses,
-            local_fleet_rating, regional_threat, sys_status['regional_fleet_threats'])
+                   ["dropping ", ""][use_protection], reason, this_planet, local_production_diff, combined_local_defenses,
+                   local_fleet_rating, regional_threat, sys_status['regional_fleet_threats'])
     return use_protection
 
 
+def use_planet_growth_specials(focus_manager):
+    '''set resource foci of planets with potentially useful growth factors. Remove planets from list of candidates.'''
```
Morlic-fo: `"""`

default/AI/ResourcesAI.py
LGM-Doyle, line 46: I used planet instead of pid to avoid the cost of a `fo.getUniverse()` call in every PlanetFocusInfo every game turn.  This way fo.getUniverse() is called once per turn in PlanetFocusManager().

I didn't profile it, so it is probably premature optimization.
Changing it is trivial. 

Should I change it?

default/AI/ResourcesAI.py
LGM-Doyle, line 50: current_output is only for the current focus.

However, I would like to change from the legacy code to a tuple as that is more consistent with the treatment of possible_output which store tuples.

default/AI/ResourcesAI.py
LGM-Doyle: It is from the legacy code. 

I think it might have been to test if the algorithm was stable independent of planet id.  It is not.  Changing the order of the pids changes the results.

I'll remove those lines.

default/AI/ResourcesAI.py
Morlic-fo, line 50: That's the point. Why does `current_output` have both industry and research focus as key if focus isn't a consideration.
Changing to tuple definitely makes sense to me.

default/AI/ResourcesAI.py
Morlic-fo: Probably should be noted somewhere as `TODO`.

default/AI/ResourcesAI.py
LGM-Doyle, line 71: I was looking for a triplet where.
1. the unprocessed and processed lists were two different words that wouldn't be mistyped or misread as the same word.  
2. the verb implied the transistion.  

Hence.

raw -> bake -> baked

raw -> bake -> processed satisfies condition 1 but not  2.

default/AI/ResourcesAI.py
LGM-Doyle: I took it from the existing code.

However, I think that the check is for target population to avoid setting focus for planets with a 0 target population but a currently non-zero population.  Those planets are not viable.

I've added the comment.

default/AI/ResourcesAI.py
LGM-Doyle, line 97: The pinfo object is transferred from the raw_planet_info to baked_planet_info.  That same pinfo object is the one that is updated.

Morlic-fo, line 97: Yes... But pid is not included in the `updateMeterEstimates` call, is that intentional?

LGM-Doyle, line 97: You are correct.
It was intentional, but then I moved the code to grab the meterValues for the current planet to after the point where pinfo was transferred.

default/AI/ResourcesAI.py
Cjkjvfnby, line 46: For me `fo.planet` instead of 'int' looks ok.  
- we are sure that we don't get any wrong id. I hate that constant check fo `"NonePointerException"`
- we cannot save this into savegame

default/AI/ResourcesAI.py
```diff
@@ -43,9 +43,8 @@ class PlanetFocusInfo(object):
     def __init__(self, planet):
         self.planet = planet
         self.current_focus = planet.focus
-        self.current_output = {}
-        self.current_output[IFocus] = planet.currentMeterValue(fo.meterType.industry)
-        self.current_output[RFocus] = planet.currentMeterValue(fo.meterType.research)
+        self.current_output = (planet.currentMeterValue(fo.meterType.industry)
+                               , planet.currentMeterValue(fo.meterType.research))
```
Cjkjvfnby: Leave coma at previous line

default/AI/ResourcesAI.py
```diff
@@ -51,6 +51,12 @@ def __init__(self, planet):
         self.possible_output[self.current_focus] = (itarget, rtarget)
         self.future_focus = self.current_focus
 
+    def can_change_focus(self):
+        """Can the planet change focus?  Is it's population non zero?"""
+        #TODO Could this be changed to current population instead of target population?
```
Cjkjvfnby: Add space after comment  `# TODO`

default/AI/ResourcesAI.py
```diff
@@ -51,6 +51,12 @@ def __init__(self, planet):
...
+        #It uses the target population to avoid situations where an unsustainable population is dying?
```
Cjkjvfnby: Looks like answer to previous TODO.

default/AI/ResourcesAI.py
Cjkjvfnby, line 97: `calculate_planet_infos` already calculates all planet info so we can remove `pinfo.possible_output[focus] = (itarget, rtarget)` form here

Dilvish-fo:
Hi, LGM-Doyle, the refactoring looks good to me.  Also, nice job catching the bug, indentation errors can be tricky enough that even after your initial note above I still had to look pretty close to spot the bug and fix.  That might actually wind up being a pretty notable bug fix, I'm interested in seeing how that works out.

Also, in that section of code the entire use of the variable "do_research" is just a holdover from a previous way of tracking preset focii; in the current code (and your refactoring) all uses of do_research could simply be removed.

If you had responded to Cj's question about the bug description, I'm not seeing it.  I would phrase it as "The PR fixes an indentation error which prevented more than one high production planet from bypassing Research Focus during the final assessment of the local research_potential : industry_potential ratio."

To confirm that the refactoring itself did not cause any unintended changes of focus assignments, did you do any save-game testing without the bugfix in place?

default/AI/ResourcesAI.py
```diff
@@ -55,8 +55,8 @@ def __init__(self, planet):
 
     def can_change_focus(self):
         """Can the planet change focus?  Is it's population non zero?"""
-        #TODO Could this be changed to current population instead of target population?
-        #It uses the target population to avoid situations where an unsustainable population is dying?
+        # It uses the target population versus current population to avoid situations
+        # where an unsustainable population is dying?
```
Dilvish-fo: It's my recollection that a while in the past, when a colony died down to an outpost it still retained its last-set focus, which could not be changed, hence a test like the one here (though I don't recall why it would look at target pop rather than current pop).  Nowadays the focus is reset to nothing in that case, so I think this test would no longer be needed (or perhaps simply be changed to look for nonempty AvailableFoci).

LGM-Doyle:
Dilvish, thanks for the review.

I have removed do_research and can_change_focus().

I answered CJ's question above.  It is one of the collapsed comments on my browser.  Do you want me to change the commit message?  I thought the change in the code was clear.

To check that I didn't break anything as I refactored I looked for automated tests. There only appears to be a testing build target in the CMakefiles for the parser.

So, here is how I tested.

Before I changed anything I ran 150 turns of a random but specific universe.  I picked two empires and observed their progress: one that grew successfully (AI4) and one that did not (AI3).  I kept their AI_N.log files as an example of a correctly working resource allocation.  I ran the same universe with ResourcesAI completely disabled and took note of the results of complete failure.

Then I added noisy debugging print statements, since removed, to adj_round 2, 3 and 4 and the ratio loop just past their guard conditions.  That is how I noticed that the header printing was masking further research assignments in the ratio portion of the algorithm.

After the refactor before I removed the noisy debugging statements AI3 and AI4 were still developed along similar lines and were toggling foci in a similar manner.

As the refactor was extensive, and there is no automated testing I can NOT say that I haven't introduced any new bugs.

Cjkjvfnby:
I still have question:

> calculate_planet_infos already calculates all planet info so we can remove pinfo.possible_output[focus] = (itarget, rtarget) form here

Last time I try to automate test, I was stopped by some randomness that can-not be controlled by AI.   I don't remember what exactly it was.

default/AI/ResourcesAI.py
LGM-Doyle, line 97: calculate_planet_infos() is called after all of the planet foci with global effects are set.  

However, bake_future_focus() is called to set the planets with global effects (before calculate_planet_infos is called) and at that point they still need their pinfo.possible_output[focus] to be calculated.

The method avoids having to re-calculate all of the planets possible_outputs after every focus setting event.

LGM-Doyle:
Cj, here are some of my quick thoughts/questions on testing the AI.

For testing the AI, I have only looked at ResourcesAI.py.  It is definitely influenced by the globals, foAI, empire and universe.  One way to make testing repeatable would be to have fake/mock foAI, empire and universe objects that just feed in expected test data and respond non-randomly to expected changes or fail the test.

I would be interested in adding testing as a developer tool to the python: adding a test_AI target to the CMake files etc.

I would not be interested in trying to write tests for all of the legacy code.

What is your preferred python testing framework?

Should I create an AI testing github issue even if it is an umbrella issue that is never checked off?

Cjkjvfnby:
@LGM-Doyle, you mean unittest for code, executed outside of the AI?  This require to split AI to modules (which is nice).

For python I prefer [pytest](http://pytest.org/latest/). It full of magic but do a lot of work for you.  On windows machine I does not manage to make it work inside AI.  (freeorion use own python for game proposes for win builds)

According to branch I have no comments left. Lets merge it.

Cjkjvfnby:
@Dilvish-fo, @Morlic-fo, any objections about merge it?

default/AI/ResourcesAI.py
```diff
@@ -211,30 +376,225 @@ def assess_protection_focus(pid):
...
+                newplanet = universe.getPlanet(pid)
+                print ("Error: Failed setting %s for planet %s (%d) with species %s and current focus %s, but new planet copy shows %s" %
+                       (focus_manager.new_foci[pid], planet.name, pid, planet.speciesName, planet.focus, newplanet.focus))
+
+
+def set_planet_happiness_foci(focus_manager):
+    """Assess and set planet focus to preferred focus depending on happiness"""
+    # TODO Assess need to set planet to preferred focus to improve happiness
+
+
```
Morlic-fo, line 483: function seems missing a `pass`  statement.

default/AI/ResourcesAI.py
```diff
@@ -211,30 +376,225 @@ def assess_protection_focus(pid):
...
+                    planet = focus_manager.all_planet_info[spPID].planet
+                    if focus_manager.bake_future_focus(spPID, GFocus):
+                        print "%s focus of planet %s (%d) at Growth Focus" % (["set", "left"][cur_focus == GFocus], planet.name, spPID)
+                        break
+                    else:
+                        print "failed setting focus of planet %s (%d) at Growth Focus; focus left at %s" % (planet.name, spPID, planet.focus)
+
+
+def use_planet_production_and_research_specials(focus_manager):
+    """Use production and research specials as appropriate.  Remove planets from list of candidates."""
```
Morlic-fo: Please be more specific in the docstring.

What is happening in the function? Does it set any foci? Does it only calculate them? How does it "use" specials.

default/AI/ResourcesAI.py
```diff
@@ -29,98 +40,252 @@
...
+    def print_table(self, priorityRatio):
+        """Prints a table of all of the captured sections of assignments"""
+        self.print_table_header()
+
+        for title, id_set in self.sections:
+            print Reporter.table_format % (("---------- " + title + " ------------------------------")[:33], "", "", "", "", "")
+            id_set.sort()  # pay sort cost only when printing
+            for pid in id_set:
+                pinfo = self.focus_manager.baked_planet_info[pid]
+                oldFocus = pinfo.current_focus
```
Morlic-fo, line 226: variable naming style here and elsewhere still violates PEP8.

`oldFocus` -> `old_focus`

Please also check the rest of the code for this (I suppose you have some decent IDE that may check for this kind of things?)

default/AI/ResourcesAI.py
```diff
@@ -211,30 +376,225 @@ def assess_protection_focus(pid):
...
+                        print "Considering Growth Focus for %s (%d) with special %s; planet has pop %.1f and %s metabolism incremental pop is %.1f" % (
+                            planet.name, pid, special, pop, metab, metabIncPop)
+                        if pinfo.current_focus == GFocus:
+                            pop -= 4  # discourage changing current focus to minimize focus-changing penalties
+                            rankedPlanets.append((pop, pid, pinfo.current_focus))
+                if not rankedPlanets:
+                    continue
+                rankedPlanets.sort()
+                print "Considering Growth Focus choice for special %s; possible planet pop, id pairs are %s" % (metab, rankedPlanets)
+                for spSize, spPID, cur_focus in rankedPlanets:  # index 0 should be able to set focus, but just in case...
```
Morlic-fo: what does `sp` stand for?
Also first index is pop, not size of the planet?

default/AI/ResourcesAI.py
```diff
@@ -211,30 +376,225 @@ def assess_protection_focus(pid):
...
+
+    ratios = []
+    # for each planet, calculate RP:PP value ratio at which industry/Mining focus and research focus would have the same total value, & sort by that
+    # include a bias to slightly discourage changing foci
+    curTargetPP = 0.001
+    curTargetRP = 0.001
+    resource_timer.start("Loop")  # loop
+    has_force = tech_is_complete("CON_FRC_ENRG_STRC")
+    # cumulative all industry focus
+    ctPP0, ctRP0 = 0, 0
```
Morlic-fo, line 497: instead of comment, use self-explaining variable names.

default/AI/ResourcesAI.py
Cjkjvfnby, line 483: pass required in no other code is present.  It just like  close bracket for indent.

default/AI/ResourcesAI.py
Cjkjvfnby, line 497: It is copy paste of old code.  But since it is local variables it can be changed here.

default/AI/ResourcesAI.py
Morlic-fo, line 483: Do comments actually count as well? Cause there sure isn't any actual code within the function.

default/AI/ResourcesAI.py
Morlic-fo, line 497: That's what you get when spending 2 minutes to check a PR due to lack of time... Anyway, I guess not strictly needed to change it within this PR then, his call.

default/AI/ResourcesAI.py
Cjkjvfnby, line 483: Comments does not count.  Function have docstring.  It is part of function declaration and can be accessed in runtime via `__doc__`

default/AI/ResourcesAI.py
Cjkjvfnby, line 497: I plan to cleanup variable names and formatting myself.  It is quite easy with IDE refactor tools and have less chances to got mistake.

LGM-Doyle:
@Morlic-fo  thanks for the review.  I changed everything except the variable names, which I left the same as before the refactor.

Morlic-fo:
I have no further objections to merging this @Cjkjvfnby

PR #584, opened 2016-04-03
Add some support for freedesktop XDG Base Directory Specification

LGM-Doyle:
The XDG Base Directory Specification is at
https://specifications.freedesktop.org/basedir-spec/basedir-spec-
latest.html

It requires that configuration files be stored in
$XDG_CONFIG_HOME/freeorion which defaults to
$HOME/.local/share/freeorion

It requires that user app data be stored in
$XDG_DATA_HOME/freeorion which defaults to
$HOME/.config/freeorion

This commit splits GetUserDir() into GetUserConfigDir() and
GetUserDataDir().  On linux and BSD they return the correct XDG
directories.  On Windows and OSX they return the same directory
as GetUserDir() before this commit.

This commit does not support XDG_DATA_DIRS, XDG_CONFIG_DIRS,
XDG_CACHE_HOME or XDG_RUNTIME_DIRS.

This does not break saves/configs, but if you do not move them to the new location 
or set the save dir to your old location you will not see them.

### DIFF
--- a/default/AI/charting/charts.py
+++ b/default/AI/charting/charts.py
@@ -20,7 +20,7 @@ import sys
 from glob import glob
 import traceback
 
-dataDir = os.environ.get('HOME', "") + "/.freeorion"
+dataDir = (os.environ.get('HOME', "") + "/.freeorion" ) if os.name != 'posix' else (os.environ.get('XDG_DATA_HOME', os.environ.get('HOME', "") + "/.local/share") + "/freeorion") 
 graphDir = dataDir
 
 fileRoot = "game1"
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -116,7 +116,7 @@ namespace {
 
     const std::string DESIGN_FILENAME_EXTENSION = ".txt";
     const std::string UNABLE_TO_OPEN_FILE = "Unable to open file";
-    boost::filesystem::path SavedDesignsDir() { return GetUserDir() / "shipdesigns"; }
+    boost::filesystem::path SavedDesignsDir() { return GetUserDataDir() / "shipdesigns"; }
 
     class SavedDesignsManager {
     public:
--- a/UI/OptionsWnd.cpp
+++ b/UI/OptionsWnd.cpp
@@ -609,7 +609,7 @@ OptionsWnd::OptionsWnd():
     // Directories tab
     current_page = CreatePage(UserString("OPTIONS_PAGE_DIRECTORIES"));
     DirectoryOption(current_page, 0, "resource-dir", UserString("OPTIONS_FOLDER_SETTINGS"), GetRootDataDir());  // GetRootDataDir() returns the default browse path when modifying this directory option.  the actual default directory (before modifying) is gotten from the specified option name "resource-dir"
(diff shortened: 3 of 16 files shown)


default/AI/charting/charts.py
```diff
@@ -20,7 +20,7 @@
 from glob import glob
 import traceback
 
-dataDir = os.environ.get('HOME', "") + "/.freeorion"
+dataDir = (os.environ.get('HOME', "") + "/.freeorion" ) if os.name != 'posix' else (os.environ.get('XDG_DATA_HOME', os.environ.get('HOME', "") + "/.local/share") + "/freeorion")
```
adrianbroher, line 23: Is there are reason why you don't use `getUserDataDir` here?

LGM-Doyle, line 23: charts.py appears to be a stand-alone charting app, that parses the AI logs.  It has no dependencies on any freeOrion code.  I didn't want to introduce new dependencies.

adrianbroher, line 23: That reason is just fine with me.  I just asked because I don't have a  complete overview over the whole code base and that part of the part the PR looks a bit weird without looking too deep into the context.

adrianbroher:
Looks good so far and works as expected.  Is there anybody who opposes this change?

Also should there be some migration code, which copies the save files over to the new default location?

Vezzra:
> Also should there be some migration code, which copies the save files over to the new default location?

If it's not too much trouble, that could probably be done... but still, I think for a game in alpha stage that isn't really necessary.

LGM-Doyle:
I don't think that there should be migration code.  

I think an announcement about the change on the bulletin board for devs/testers benefit is the correct solution.

For regular users, the change from previous version to the next version, is a breaking change and neither the config files nor the save will be compatible. So, no need for migration code.

For devs/testers the situation is complicated, because all options may have unpleasantly surprising results.  

Doing the migration correctly means moving saves and saved designs to $XDG_DATA_HOME/freeorion/, moving config and persistent configuration to $XDG_CONFIG_HOME/freeorion/ and then deleting ~/.freeorion/.  

Devs may be storing, linking  or doing other things in ~/.freeorion.  I, for example camp(autorevert) on the AI_x.log files while I'm fiddling with the AI code.  If it suddenly disappeared I would be surprised.  If it suddenly stops updating, I would also be surprising but less annoyed. 

Doing nothing.means that the next freeorion startup ignores your persistent_config, a big visible change.  The dev immediately notices, checks the git log,  sees something about XDG Directory support, read the commit message and moves their files or updates their scripts as appropriate.  

I think that the least surprising options is to *_announce the change on the bulletin board *_ but not write migration code.

However, I can write the migration code.  You decide and I will do, but not today.

LGM-Doyle:
I'm working on the migration code.  I'll make it have a popup message, so that it doesn't change anything unexpectedly.

adrianbroher:
Works for me.  But I would like to see rather `MIGRATION.README` instead of `README_CONFIGURATION_FILES_MIGRATED` as readme file.

LGM-Doyle:
MIGRATION.README is a better name.
Thanks.

PR #576, opened 2016-03-29
[Testing Required] AI: Combat Ratings Update + Fighter Adaptations

Morlic-fo:
## WIP _Not ready to merge_

As far as my playtesting is concerned, the AI didn't throw any errors and performed similar to the old state. However, these kind of changes require long-term testing in multiple situations to really find all bugs and issues. Feel free to test these changes and provide feedback.

This PR should be stable enough to provide a framework for balancing the fighter parts.
Note that while this PR should let the AI handle and build bay/hangar parts, it still needs to actually research the relevant techs. So if you add new techs, make sure to add them at a suitable location in TechlistsAI.py as well.

Note that in this PR, there are some fighter parts unlocked for testing purposes.

Performance is something I haven't looked too much into yet, probably some stuff can/should be cached.
# Changelog

**Reimplementation of the existing rating system**
- Introduced new module CombatRatingsAI.py to calculate military ship- and fleet-ratings
- Unified rating calculations (should generally be close to previous values)
- Updated species modifier for weapon strength

**Introduced basic Fighter support for the AI**
- The AI will now take into account fighters when rating fleet strength
- The AI will now make use of fighter hangars and bays, i.e. build carrier-style ships
- The AI will try to replenish fighters under the same conditions as refueling ships
- The AI will use only use a single hangar type per ship.
- Preunlocked fighter parts for testing purposes

PR #585, opened 2016-04-03
Add cache by session function

Cjkjvfnby:
At current time we have a lot of values that cached in local variables. This function will allow to make this code more friendly.

I expect first usages will be introduced in this PR https://github.com/freeorion/freeorion/pull/576

### DIFF
--- a/default/AI/freeorion_tools.py
+++ b/default/AI/freeorion_tools.py
@@ -132,6 +132,25 @@ def chat_human(message):
     print "\nChat Message to human: %s\n" % remove_tags(message)
 
 
+def cache_by_session(function):
+    """
+    Cache a function value by session.
+    Wrap only functions with hashable arguments.
+    """
+    _cache = {}
+
+    @wraps(function)
+    def wrapper(*args, **kwargs):
+        key = (function, args, tuple(kwargs.items()))
+        if key in _cache:
+            return _cache[key]
+        res = function(*args, **kwargs)
+        _cache[key] = res
+        return res
+    wrapper._cache = _cache
+    return wrapper
+
+
 def cache_by_turn(function):
     """
     Cache a function value by turn, stored in foAIstate so also provides a history that may be analysed. The cache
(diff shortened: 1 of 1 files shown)


default/AI/freeorion_tools.py
```diff
@@ -132,6 +132,25 @@ def chat_human(message):
     print "\nChat Message to human: %s\n" % remove_tags(message)
 
 
+def cache_by_session(function):
+    """
+    Cache a function value by session.
+    Wrap only functions with hashable arguments.
```
Morlic-fo: Wraps

Cjkjvfnby:
Fixed.

Github have new feature squash and rebase instead of merge, it perfectly suits for this PR

Morlic-fo:
> Github have new feature squash and rebase instead of merge, it perfectly suits for this PR

I still only have the option to do a classic merge.

MatGB:
Same here, it needs to be enabled by an admin, I've commented on the thread discussing it that it should be an option that we should use rarely, there are times when I'd much prefer to see a single commit at the time it went in.

Cjkjvfnby:
Looks like it is not disabled for me. 
- click Merge pull request.
- click arrow near `Confirm merge`
- select squash and merge
- edit commit message

Vezzra:
The option has been enabled by default when github introduced the feature, no one has turned it off as AFAICT. @MatGB, @Morlic-fo, apparently you didn't realize that you need to click "Merge pull request" first, and only then get the option to choose...?

And apparently @Morlic-fo figured it out after @Cjkjvfnby comment and did a "squash and merge" with this commit, right?

MatGB:
Pretty much, yes, thanks for checking. Every time I think I've learnt what I need to know about git/github something changes ;-)

Morlic-fo:
> @MatGB, @Morlic-fo, apparently you didn't realize that you need to click "Merge pull request" first, and only then get the option to choose...?

Yep, that was it. I expected to have the dropdown already available from the start...

PR #589, opened 2016-04-04
Use separate icons for ship tech categories

dbenage-cx:
Renames categories for tech window control ordering


geoffthemedio:
Changing the icons doesn't require renaming the category labels. I'm not sure what purpose doing such renaming serves...

dbenage-cx:
original: tech_controls_basic

with renaming:tech_controls_proposed

PR #603, opened 2016-04-11
Fix null pointers in GG::Wnd DeleteChild, DeleteChildren and RemoveLayout

LGM-Doyle:
1. Both DeleteChild and DeleteChildren kept pointer references in
   m_children to deleted pointers.
2. DeleteChild did not call RemoveLayout when the child was the m_layout
   object.
3. DeleteChildren did not null m_layout.
4. RemoveLayout should null m_layout before calling DeleteChild.

### DIFF
--- a/GG/src/Wnd.cpp
+++ b/GG/src/Wnd.cpp
@@ -606,17 +606,26 @@ void Wnd::DetachChildren()
 
 void Wnd::DeleteChild(Wnd* wnd)
 {
-    if (wnd && std::find(m_children.begin(), m_children.end(), wnd) != m_children.end()) {
+    if (!wnd)
+        return;
+
+    if (wnd == m_layout)
+        RemoveLayout();
+    std::list<Wnd*>::iterator found = std::find(m_children.begin(), m_children.end(), wnd);
+    if (found != m_children.end()) {
+        m_children.erase(found);
         delete wnd;
     }
 }
 
 void Wnd::DeleteChildren()
 {
+    m_layout = 0;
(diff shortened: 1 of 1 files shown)


GG/src/Wnd.cpp
```diff
@@ -606,17 +606,25 @@ void Wnd::DetachChildren()
 
 void Wnd::DeleteChild(Wnd* wnd)
 {
-    if (wnd && std::find(m_children.begin(), m_children.end(), wnd) != m_children.end()) {
-        delete wnd;
+    if (wnd) {
```
geoffthemedio: prefer

```
if (!wnd)
    return

// more code not indended unnecessarily
```

adrianbroher:
Looks like a good change and seemingly doesn't introduce new bugs on a first test.  What made you do this change?  Just digging in the code or some more sophisticated analysis like using valgrind?

LGM-Doyle:
While fiddling in the code, I tried to delete all windows and add some new windows to a layout  and crashed.  Running gdb said "normal exit" which led me to PR #604 fixing the exception handlers.

I did run valgrind.  There were 1830 potential leaks. 

The first was an obviously intentionally leaked statically initialized pointer.

Then there were some long template style potential errors from the effects code.  I couldn't tell if those were intentional, or more false positives.  

After that I decided that 1830 potetntial leaks meant freeorion didn't use valgrind to police memory leaks.  I decided a better approach was to ask about C++11 and try to use fewer naked pointers.

PR #604, opened 2016-04-11
Remove exception handlers from chmain that were masking stack traces

LGM-Doyle:
The exception handlers were trapping all the errors, handling none of them and printing a message saying something bad happened.  This had the result of deleting the stack trace because freeorion had exited normally.

I have removed the fake exception handlers and now freeorion produces a stack trace for unhandled exceptions.

### DIFF
--- a/client/human/chmain.cpp
+++ b/client/human/chmain.cpp
@@ -19,6 +19,19 @@
 
 #include <iostream>
 
+#if defined(FREEORION_LINUX)
+/* Freeorion aims to have exceptions handled and operation continue normally.
+An example of good exception handling is the exceptions caught around config.xml loading.
+After catching and informing the user it continues normally with the default values.
+
+An exception that can not be handled should allow freeorion to crash and keep
+a complete stack trace of the intial exception.
+Some platforms do not support this behavior.
+
+When FREEORION_CHMAIN_KEEP_BACKTRACE is defined, do not catch an unhandled exceptions,
+unroll and hide the stack trace, print a message and still crash anyways. */
+#define FREEORION_CHMAIN_KEEP_STACKTRACE
+#endif
 
 // The STORE_FULLSCREEN_FLAG parameter below controls whether the fullscreen
 // option is stored in the XML config file.  On Win32 it is not, because the
@@ -92,7 +105,9 @@ int mainConfigOptionsSetup(const std::vector<std::string>& args) {
     InitDirs((args.empty() ? "" : *args.begin()));
 
     // read and process command-line arguments, if any
+#ifndef FREEORION_CHMAIN_KEEP_STACKTRACE
     try {
+#endif
         // add entries in options DB that have no other obvious place
         GetOptionsDB().AddFlag('h', "help",                 UserStringNop("OPTIONS_DB_HELP"),                  false);
         GetOptionsDB().AddFlag('g', "generate-config-xml",  UserStringNop("OPTIONS_DB_GENERATE_CONFIG_XML"),   false);
@@ -187,6 +202,7 @@ int mainConfigOptionsSetup(const std::vector<std::string>& args) {
             GetOptionsDB().Set<bool>("show-fps",                true);
         }
 
+#ifndef FREEORION_CHMAIN_KEEP_STACKTRACE
     } catch (const std::invalid_argument& e) {
         std::cerr << "main() caught exception(std::invalid_argument): " << e.what() << std::endl;
         boost::this_thread::sleep_for(boost::chrono::seconds(3));
@@ -199,10 +215,8 @@ int mainConfigOptionsSetup(const std::vector<std::string>& args) {
         std::cerr << "main() caught exception(std::exception): " << e.what() << std::endl;
         boost::this_thread::sleep_for(boost::chrono::seconds(3));
         return 1;
-    } catch (...) {
-        std::cerr << "main() caught unknown exception." << std::endl;
-        return 1;
     }
+#endif
 
     return 0;
 }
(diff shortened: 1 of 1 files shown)


client/human/chmain.cpp
```diff
@@ -187,22 +187,22 @@ int mainConfigOptionsSetup(const std::vector<std::string>& args) {
...
 
-    } catch (const std::invalid_argument& e) {
+    /* Restore this code when C++11 is adopted and rethrow e will preserve the stack trace.
+       Otherwise these Exception non-handlers just hide the stack trace and make debugging harder.
+    }
+    catch (const std::invalid_argument& e) {
         std::cerr << "main() caught exception(std::invalid_argument): " << e.what() << std::endl;
         boost::this_thread::sleep_for(boost::chrono::seconds(3));
-        return 1;
+        rethrow e;
```
adrianbroher: I'm having a hard time finding any material about this `rethrow` keyword. Can you point out the specification paper?

LGM-Doyle:
With rethrow e, I was thinking of was this.
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2179.html
http://www.cplusplus.com/reference/exception/rethrow_exception/

I never compiled or checked that it worked as expected.. I intended it to rethrow the exception and preserve the stack. 

I should have been bolder in the first PR and simply removed all of the non-exception handling code.

I think that the best approach is to let un-handled exceptions crash freeorion with a useful stack.  Then regular users can be directed to run gdb and post the backtrace.

However, the original code looked like it was trying to hide the exceptions and exit cleanly with only a message in freeorion.log.  This isn't useful to developers, but may be friendlier to regular users.

I could re-write the patch to mask the exceptions in the RELEASE version and not-handle them in the DEBUG version.  

I should re-write this patch.

Please pick an alternative:
a. (BEST) - Don't handle exceptions and remove the commented out exception masking.
b. Mask exceptions in RELEASE version and don't handle exceptions in DEBUG version.
c. Reject PR and keep the old behavior.

adrianbroher:
> I never compiled or checked that it worked as expected.. I intended it to rethrow the exception and preserve the stack.

Unfortunately it doesn't work this way.  As far as I know (and fact checked with some research) every time an exception is `catch`ed every stack frame above this is unwind and cleaned up.  Also it is implementation defined behaviour if the `std::terminate` handler has access to the stack trace.

> I think that the best approach is to let un-handled exceptions crash freeorion with a useful stack. Then regular users can be directed to run gdb and post the backtrace.

The project is cross platform, therefor you can't assume gdb or g++. 

I agree that _disabling_ the coredumps isn't a good idea at all, but I would assume that this code was added to provide more information on unix-ish platforms.  For example Windows doesn't seem to report the exception message but rather '&lt;exception_type&gt; was raised' in  the default handler and otherwise the error handling on Windows seems to fall into the category [horrible mess](http://www.codeproject.com/Articles/207464/Exception-Handling-in-Visual-Cplusplus).

Unless you have a better idea I would rather 'fix' this for now by adding a conditional `#define` and rethink if either the exception handling should be reimplemented to expose the good features of the platforms or if some third party library like _google-breakpad_ couldn't help here.

LGM-Doyle:
I agree every catch unwinds the stack, which is why I don't want to catch.  The stack information is more important than echoing `e.what()` in freeorion.log.

I read the article that you linked on the Windows exception handling.  It was a good way to highlight the difficulties of multi-platform exception handling. I'm glad I can't volunteer to help out with that mess.

I didn't assume that everyone has gdb.  I was trying to be less verbose.  I am assuming that every platform has a debugger and that the stack information is useful regardless of platform.

I am running linux.  I don't think the current setup is useful, unless you are trying to hide complexity from user.  Its benefit is copying the e.what() message to freeorion.log, but the message will appear in the console and crash reporting tool. 

I agree that ideally we would have an opt-in multi-platform crash reporter.  

Right now, I'm shooting for the lower bar of stacktraces for devs and simple, rich, manual crash reporting for  users.  I will have to settle for linux only.

If we let the un-handled exceptions fall through, then we can add the following simple instructions to the website and regular linux users can post useful backtraces.
1. Type `gdb -- freeorion`
2. At gdb prompt type `run`
3. Make free orion crash.
4. At gdb prompt type `backtrace`
5. Post backtrace on  freeorion.org.

No matter the platform, the current exception handlers are not handling the exception.  Execution always terminates.  That is why I still think the best alternative is removing the exception handlers.

The narrowest fix is to only remove the exception handlers within sections defined by. 
#if defined(FREEORION_LINUX) && defined(BUILD_DEBUG)

Do we use DEBUG and RELEASE defines to compiles different versions of the code? 

How do you want me to section the code?

Vezzra:
I'm surprised @geoffthemedio hasn't chimed in yet, so I'll try to sum up the approach behind how FO handles exceptions (AFAIK): You've already guessed at it, the main purpose is to spare regular users crashes and try to continue whenever possible. The idea is to silently log error messages in the background and only let FO crash if program execution really can't continue anymore.

That approach has been subject to discussions in the past, but has remained project policy till now, which I guess is why you found what you found. (I remember a discussion I had with @geoffthemedio on building FO with or without `NDEBUG` defined, which disables assertions. He insisted on keeping NDEBUG to avoid crashes for regular users just because some odd assertion deep in GG code failed.)

If you guys want to propose changes on how exceptions should be handled, you need to check back with @geoffthemedio. After all, he is our project and programming lead, and I remember him to be pretty firm on the subject.

> The narrowest fix is to only remove the exception handlers within sections defined by.
> #if defined(FREEORION_LINUX) && defined(BUILD_DEBUG)

Something like that sound like a good approach, and IMO doesn't need to be restricted to the Linux platform (unless you plan to implement Linux specific exception handling). One question though: can we tie the code sections with the improved exceptions handling to a separate symbol ("IMPROVED_EXCEPTION_HANDLING" or whatever you see fit) instead of BUILD_DEBUG? Otherwise I can't use it on OSX, as DEBUG builds are kind if impossible on that platform (the weekly test builds on OSX are done with a special "test release" build environment).

geoffthemedio:
I assumeed this was just about the top-level exception handlers in chmain or similar places, which would only be relevant when an exception hasn't been handled locally, and the program is going to crash out anyway. If that's not the case, then yes, as Vezzra notes, the priority if to avoid crashing if possible. Modifications to preserve a stack trace in the event of an unavoidable crash are OK, though, I think...

adrianbroher:
> If we let the un-handled exceptions fall through, then we can add the following simple instructions to the website and regular linux users can post useful backtraces.

Are you aware of the existence of `catchsegv`? That's easier to use. `gdb` is interesting in the context of coredumps.

> The narrowest fix is to only remove the exception handlers within sections defined by. 
> #if defined(FREEORION_LINUX) && defined(BUILD_DEBUG)

Remove the `BUILD_DEBUG` condition, it is never defined and I always want coredump anyway, regardless of any potential build mode.

LGM-Doyle:
@geoffthemedio your thoughts are correct.

@Vezzra thanks for your response on the forums about the merge message.  A "Well Duh" moment on my behalf.

@adrianbroher, I did not know about catchsev.  I tried it out.  It works as advertised on the box, catching seg faults.  However, it doesn't catch general exceptions.  I wrote `throw "Some String"`, which crashes as expected but is not in turn caught by catchsev.

Code wise, I have bracketed and excluded the exception non-handlers in linux only.

I have restored the `return 1;` return code instead of `throw e;` as the error exit method of the non-linux handlers.  This means that non-linux exceptions will be masked and more difficult to debug.

adrianbroher:
Merged with 1fe183d3.

PR #607, opened 2016-04-14
Code cleanup and simplification for branch Fighters

Cjkjvfnby:
May be my commits should be cherry-picked to Fighters manually.


adrianbroher:
Why don't you request your PR against the freeorion:Fighters branch? Something like:

> Cjkjvfnby  wants to merge xx commits into freeorion:Fighters from Cjkjvfnby:even-more-cleanup

You can select which branch you want to merge against in the Webgui and the client

MatGB:
Yeah, if we cherry pick across they get different hashes, the point of the branch was so we can have all the work in the same place and then merge it in together, cherry picking between branches causes issues.

Cjkjvfnby:
@adrianbroher, I planned to create it against Fighter, but forget to select it.

@MatGB, cherry picking between branches does not causes issues more than rebasing. It can be used as alias to make patch from commit and apply this patch to other place, like we done in svn workflow.  There is a lot of ways in git to mess thing up :)

Cjkjvfnby:
Closed to be reincarnated against Fighter.

Cjkjvfnby:
@adrianbroher , this is improvement to branch that is in development. It should have same Milestone.

Vezzra:
Well, you closed it commenting that it will be replaced with another PR, so this one here isn't going anywhere anymore (hence the "Gateway to the Void" milestone, I like @adrianbroher's sense of humor... 😄).

PR #609, opened 2016-04-16
Fix tech tree arcs null pointer reference and memory leak

LGM-Doyle:
This pull request fixes two errors:
- a null pointer reference in TechTreeArcs::Render
  when m_impl has not been allocated. 
- a memory leak in Reset() where m_impl is zeroed without first unallocating memory.


geoffthemedio:
looks reasonable...

Issue #538, opened 2016-03-19
Your Mines should generate Sitreps when they damage enemy ships


cpeosphoros:
Refer to: http://www.freeorion.org/forum/viewtopic.php?f=15&t=10006&view=unread&sid=96ccf93875079e58510e2de75d3978b7#unread

MatGB said:

> players, especially new ones, don't really get to learn how mines work, this would help.

Vezzra:
Implemented by #616

PR #616, opened 2016-04-21
Restructure mine damage effectgroups

### LINKED ISSUE #360: Check if mine techs are applied correctly/consistently to space monsters
Brought up here: http://www.freeorion.org/forum/viewtopic.php?p=80779#p80779

Created an issue so we do not forget to do that check eventually.

dbenage-cx:
Fixes #360 & #538 
Ensure sitreps are generated for both planet and ship owner when mines damage or destroy a ship.

A message is generated for the planet owner for invisible ship damage, but only to note that something set them off.

### DIFF
--- a/default/scripting/specials/FORTRESS.focs.txt
+++ b/default/scripting/specials/FORTRESS.focs.txt
@@ -33,82 +33,7 @@ Special
                 SetShield value = Value + min(5.0, max(Value, 0.25*Target.Construction))
                 SetDefense value = Value + 1
             ]
-
-        // Mines Start
-        EffectsGroup
-            scope = And [
-                Ship
-                ContainedBy And [
-                    System 
-                    Contains And [
-                        Source
-                        Not OwnerHasTech name = "DEF_SYST_DEF_MINE_2"
-                    ]
-                ]
-                (RootCandidate.Owner != Source.Owner)  // particularly to exclude an unowned Fotress with unowned monsters
-                Or [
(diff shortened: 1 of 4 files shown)


Vezzra:
The mine effect does not work for Fortress specials on unowned planets. Just ran a test, my ships didn't receive any damage. Looking at the mines macro I don't think you can use the same macro for both the mine effects of both the mine techs and the Fortress special, because `Source` evaluates to very different things in tech vs special effectsgroups contexts...

dbenage-cx:
Weird, swore that was the last check I did :/

Testing checklist
- Unowned fortress damages player ships (3 damage)
- Unowned fortress does not damage unowned ships
- Owned fortress:
  - Enemy ships damaged
  - Unowned ships damaged
  - Fortress damage(3) with no mine tech.
  - Fortress damage(3) with mine tech 1 researched.
  - Damage from tech only for mines 2 or 3 researched.
- Mine techs:
  - Damage only applies from the highest level
  - Enemy ships damaged
  - Unowned ships damaged
  - Ally/Owned ships not damaged
- Ships are destroyed when mine damage > ship structure.
- Sit Reps:
  - Generated for all of the above, if a ship or planet owned by the player is involved.
  - Ships invisible to planet owner only display that mines were triggered to planet owner.
  - One sitrep per fleet for mine damage.
  - Sitrep for each destroyed ship

Vezzra:
Ok, as you apparently did quite some testing, and my brief testing also indicates that things work fine now, I'm going to merge.

PR #617, opened 2016-04-22
Made Acirema home star blue

### LINKED ISSUE #592: Acirema should have a blue star
"Portal is a Radiated world close to an unusually bright Blue Giant Star"

matt474:
Fixes #592 by changing giving the Acirema an effect to make their star blue at the start of a game.

### DIFF
--- a/default/scripting/species/SP_ACIREMA.focs.txt
+++ b/default/scripting/species/SP_ACIREMA.focs.txt
@@ -48,6 +48,14 @@ Species
             effects = CreateShip designname = "SM_ACIREMA_GUARD" species = "SP_ACIREMA"
 
         EffectsGroup
+            scope = And [
+                Object id = Source.SystemID
+                System
+            ]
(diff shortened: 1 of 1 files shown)


MatGB:
Have you tested this to see if it causes any weird problems? I can't think of any it would but I never had the time to set up a few games and just run them through to make sure it was fine.

matt474:
I did some testing and I do not see any obvious problems. The script is also very similar to one used for phototropic species in the population macros so I do not think this will result in any new problems.

MatGB:
Yeah, that's what I expect, the only possible issue is having them both turn up in the same system, but that theoretically can't happen as photrophe natives don't do the changing anymore. Cool, thanks.

Vezzra:
@matt474, do you want to provide a real name for the credits, or should I just add you under your nick?

matt474:
I would prefer to use just my nickname for now.

PR #621, opened 2016-04-24
Adds effectsgroup to XENOPHOBIC_SELF macro to make other-species-popu…

geoffthemedio:
…lated planets visible.

### DIFF
--- a/default/scripting/species/common/xenophobic.macros
+++ b/default/scripting/species/common/xenophobic.macros
@@ -28,13 +28,13 @@ XENOPHOBIC_SELF
                 SetTargetHappiness value = Value - (
                     Count condition = And [
                     PopulationCenter
-                    WithinStarlaneJumps jumps = 5 condition = Source
                     OwnedBy empire = Source.Owner
                     Not OR [ 
                         Species name = Source.Species
                         Species name = "SP_EXOBOT"
                     ]
                     Not Population high = 0
+                    WithinStarlaneJumps jumps = 5 condition = Source
                 ]
                 )
             ]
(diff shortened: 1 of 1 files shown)


MatGB:
Playing this through, and I'm not sure it should apply to scout ships, possibly just population centres.

On the other hand, if we used this to _replace_ the existing vision bonus they get it could be rather useful and more characterful.

Vezzra:
Are we talking about Trith in particular or xenophobic races in general now? Because this PR introduces "xenophobic visibility" for all xenophobic species (if I understand correctly) - is that what you had in mind, @MatGB?

geoffthemedio:
Also, it doesn't currently apply to ships... just planets (or population centres...).

MatGB:
When I suggested it I meant for just Trith (well, self sustaining xenophobes, currently just Trith), Eaxaw definitely don't need the boost.

And it makes planets visible if you've a ship in range, I had meant just in range of Trith planets but I think this is actually better if we tone down their vision boost.

Vezzra:
@MatGB: just made a PR which grants this kind of detection ability only to Trith: #627. Is that more in line with what you had in mind?

(I made a new species trait "Telepathic Detection" and added that to the Trith species.)

geoffthemedio:
Shall I close / delete this then? (or Vezza, feel free to do so if it's decided to use a different implementation...)

Vezzra:
> (or Vezza, feel free to do so if it's decided to use a different implementation...)

Ok, will do. Once my PR gets merged, I will close this one here and delete the XenophobicVisibility branch.

Vezzra:
#627 has been merged, closing this.

PR #623, opened 2016-04-24
Allow unhandled exceptions to leave a complete stack trace in camain …

LGM-Doyle:
…and cdmain.

Remove exception handlers from camain and dmain that were not handling
the exception and only printing a message to the log.
These non-handlers just unroll the stack.

Allow unhandled exceptions to crash freeorion and produce a complete
stack trace.

This applies to linux only.

These are the same changes for the same reasons as PR #604.

### DIFF
--- a/client/AI/camain.cpp
+++ b/client/AI/camain.cpp
@@ -9,6 +9,20 @@
 
 #include <boost/filesystem/fstream.hpp>
 
+#if defined(FREEORION_LINUX)
+/* Freeorion aims to have exceptions handled and operation continue normally.
+An example of good exception handling is the exceptions caught around config.xml loading.
+After catching and informing the user it continues normally with the default values.
+
+An exception that can not be handled should allow freeorion to crash and keep
+a complete stack trace of the intial exception.
+Some platforms do not support this behavior.
+
+When FREEORION_CAMAIN_KEEP_BACKTRACE is defined, do not catch an unhandled exceptions,
+unroll and hide the stack trace, print a message and still crash anyways. */
+#define FREEORION_CAMAIN_KEEP_STACKTRACE
+#endif
+
 #ifndef FREEORION_WIN32
 int main(int argc, char* argv[]) {
     InitDirs(argv[0]);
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Can someone else on Linux confirm this works as intended?

dbenage-cx:
Based on the description, this seems to have to opposite effect? (or my testing is flawed)

UI/ResearchWnd.cpp:445 `throw "a";` 
launch game->quickstart->add research to queue

PR:

```
$ gdb build/freeorion
(gdb) r
...
main() caught unknown exception.
[Thread 0x7fffe17cf700 (LWP 13861) exited]
[Thread 0x7fffe5fd1700 (LWP 13860) exited]
[Inferior 1 (process 13854) exited normally]
(gdb) bt
No stack.
```

Vezzra:
> (or my testing is flawed)

This, I guess. You're testing the _human_ client, this PR removes the top level exception handlers from the _AI_ client and the _server_.

Vezzra:
I can build and run this PR on my Linux box, works fine. But I don't know how to make the AI client and the server throw unhandled exceptions, and even if I managed that, I don't know how to get the stack...

LGM-Doyle:
Here is how I tested. 

You need to 
1. cause the freeorionca or the freeoriond executables to crash. 
2. at the same time you need to capture the stack trace.
1. To cause the crashes on the first turn update I added either 

`case Message::TURN_UPDATE: {
         throw "a";`

to AIClientApp.cpp:249, or

`ProcessingTurn::ProcessingTurn(my_context c) :
    my_base(c)`

`{ if (TRACE_EXECUTION) DebugLogger() << "(ServerFSM) ProcessingTurn"; 
throw "a";}`

to ServerFSM.cpp:1442 

Only add one or the other change not both.

2 . To grab the backtrace I used gdb as follows:
Start a freeorion game with a single AI.
Start gdb with the pid of either freeoriond or freeorionca.
Continue the now observed process running.
Hit next turn.
Look at backtrace.

`freeorion --quickstart&`  assuming your config files starts a game with single AI.
`ps -elf | grep freeorionca`
`gdb ./freeorionca --pid <pid from ps>`
`(gdb) continue`
Hit next turn.
`(gdb) backtrace`

dbenage-cx:
Thank you LGM-Doyle, both of the provided have backtrace.
Will assume the first (invalid) test differs from master due to commits since this PR formed.

Vezzra:
Yep, thanks @LGM-Doyle, with those instructions I've been able to test this too and can confirm PR works as expected. Going to merge.

Issue #513, opened 2016-03-09
Small UI "bug" (or misbehavior)


cpeosphoros:
I'm playing in a Linux Mint box, with FO compiled from trunk 2 days ago.

The thing is, when I close the Production UI by hitting "ESC" the SitRep window resets to the first topic. When closing it by clicking the cog icon in the top of the screen the SitRep doesn't change its state (which I think is the desirable behavior).

---

All contributions are released under GPL or LGPL v2 or later, or under appropriate Creative Commons licence, consistent with project guidelines.

Vezzra:
Can you specify the exact build (or any other build with which you can reproduce the issue)?

cpeosphoros:
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

Vezzra:
Confirmed on OSX (build a0f6702ad9e22edcda7ad69d7d9509d6511be8e5).

Issue #597, opened 2016-04-06
Indefinitely dismissing a SitRep makes the SitRep window scroll back to the first line.


cpeosphoros:
As per the title.

dbenage-cx:
related #513 ?

PR #625, opened 2016-04-24
Fix bug513 bug597 resetting sitrep

LGM-Doyle:
This PR fixes bug #513 and bug #597, which both deal with the sitrep window resetting when either the MapWnd closes the production window or when the filtering of sitreps is changed.

It also fixes two un-reported bugs in GG::List SetFirstRowShown and BringRowIntoView where it was using the height of the target row in place of the height of each individual rows when re-positioning the rows.  This fix was necessary for the fix of the resetting of the sitrep window with changing filtering.

The fix for the map window is simply to not update when toggling the sitrep window.

The fix for the resetting with filtering is two steps.
1. always restore sitrep->list positioning on each update. 
2. Only change which sitrep is displayed in ShowSitRepsForTurn(), which also sets the sitrep window to display the first row.

### DIFF
--- a/GG/src/ListBox.cpp
+++ b/GG/src/ListBox.cpp
@@ -987,7 +987,7 @@ void ListBox::BringRowIntoView(iterator it)
     if (m_vscroll) {
         Y acc(0);
         for (iterator it2 = m_rows.begin(); it2 != m_first_row_shown; ++it2)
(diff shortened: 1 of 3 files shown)


Vezzra:
Builds and runs fine on OSX, works as intended.

Can be merged as far as I'm concerned.

Issue #610, opened 2016-04-17
ShipHull Icon and Graphic are swapped


dbenage-cx:
`HullType` graphic name is stored as icon, and icon as graphic.

```
diff --git a/UI/ClientUI.cpp b/UI/ClientUI.cpp
index 804de7f..53fc166 100644
--- a/UI/ClientUI.cpp
+++ b/UI/ClientUI.cpp
@@ -311,6 +311,7 @@ boost::shared_ptr<GG::Texture> ClientUI::HullIcon(const std::string& hull_name)
         if (texture_name.empty())
             texture_name = hull->Graphic();
     }
+    std::cout << "hull: " << hull_name << "\n\ticon: " << hull->Icon() << "\n\tgraphic: " << hull->Graphic() << std::endl;
     if (texture_name.empty())
         return ClientUI::GetTexture(ArtDir() / "icons" / "ship_hulls"/ "generic_hull.png", true);
     return ClientUI::GetTexture(ArtDir() / texture_name, true);
```

returns:

```
hull: SH_SMALL_ASTEROID
    icon: hulls_design/small_asteroid_hull.png
    graphic: icons/ship_hulls/small_asteroid_hull_small.png
...
```

GCC 5.3.1
Boost 1.58

dbenage-cx:
Easiest comparison without changes is looking at solar hull in design window, it has a unique pov between icon and graphic.

Works around the issue, though the unmodified order is correct:

```
diff --git a/parse/ShipHullsParser.cpp b/parse/ShipHullsParser.cpp
index 1c66bb3..bad29b6 100644
--- a/parse/ShipHullsParser.cpp
+++ b/parse/ShipHullsParser.cpp
@@ -107,7 +107,7 @@ namespace {
                 >   parse::detail::common_params_parser()       [ _d = _1 ]
                 >   parse::label(Icon_token)    > tok.string    [ _f = _1 ]
                 >   parse::label(Graphic_token) > tok.string
-                    [ insert_hull(_r1, new_<HullType>(_c, _d, _a, _e, _1, _f)) ]
+                    [ insert_hull(_r1, new_<HullType>(_c, _d, _a, _e, _f, _1)) ]  //TODO: graphic and icon are swapped during construction
                 ;

             start
```

MatGB:
[...]

PR #626, opened 2016-04-25
Fix bug #610 ShipHull Icon and Graphic are swapped in ShipHullsParser.

LGM-Doyle:
This fixes bug #610.


Vezzra:
Seems to work fine.

PR #627, opened 2016-04-27
Give Trith new Telepathic Detection species trait

Vezzra:
Added new "Telepathic Detection" species trait which grants partial visibility of populated planets within a certain number of jumps of a ship or planet of the species with this trait and added the new trait to species Trith. Removed the Great Detection species trait from Trith for balance reasons.

### DIFF
--- a/default/scripting/species/SP_TRITH.focs.txt
+++ b/default/scripting/species/SP_TRITH.focs.txt
@@ -27,12 +27,12 @@ Species
         [[AVERAGE_SUPPLY]]
         [[AVERAGE_DEFENSE_TROOPS]]
 
-        [[GREAT_DETECTION]]
+        [[GOOD_DETECTION]]
         [[GOOD_STEALTH]]
 
         [[XENOPHOBIC_SELF]]
-
         [[XENOPHOBIC_OTHER(TRITH)]]
+        [[TELEPATHIC_DETECTION(5)]]
 
         // not for description
         [[AVERAGE_PLANETARY_SHIELDS]]
--- a/default/scripting/species/common/telepathic.macros
+++ b/default/scripting/species/common/telepathic.macros
@@ -1,0 +1,12 @@
+TELEPATHIC_DETECTION
+'''     EffectsGroup
+            description = "TELEPATHIC_DETECTION"
+            scope = And [
+                PopulationCenter
+                Not Population high = 0
+                WithinStarlaneJumps jumps = @1@ condition = Source
+                Not VisibleToEmpire empire = Source.Owner
+            ]
+            activation = OwnedBy affiliation = AnyEmpire
+            effects = SetVisibility visibility = Basic empire = Source.Owner
+'''
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -11237,6 +11237,11 @@ GREAT_ASTEROID_INDUSTRY
 
 LIGHT_SENSITIVE
 −	Light Sensitive: Population is reduced in systems with blue and to a lesser extend white stars.
+
+TELEPATHIC_DETECTION
++	Telepathic Detection: can sense nearby populated planets.
+
+
 #############################################################
 ####          A C C O U N T I N G    L A B E L S         ####
 #############################################################


Vezzra:
This is an attempt to provide an alternative implementation to #621.

default/scripting/species/common/telepathic.macros
```diff
@@ -0,0 +1,15 @@
+TELEPATHIC_DETECTION
+'''     EffectsGroup
+            scope = And [
+                PopulationCenter
+                Not OR [
+                    Species name = Source.Species
+                    Species name = "SP_EXOBOT"
+                ]
+                Not Population high = 0
+                WithinStarlaneJumps jumps = @1@ condition = Source
```
MatGB, line 7: How does this work, am I missing something? I can't see where @1@ is defined.

MatGB:
This is I think what I had in mind, perhaps let them have Good visibility (same as Happybirthday) instead of Great (ie tone it down but not remove it), but would want to test it more.

default/scripting/species/SP_TRITH.focs.txt
```diff
@@ -27,12 +27,12 @@ Species
         [[AVERAGE_SUPPLY]]
         [[AVERAGE_DEFENSE_TROOPS]]
 
-        [[GREAT_DETECTION]]
         [[GOOD_STEALTH]]
 
         [[XENOPHOBIC_SELF]]
-
         [[XENOPHOBIC_OTHER(TRITH)]]
+        [[TELEPATHIC_DETECTION(5)]]
```
Vezzra, line 35: @MatGB: The "5" which is passed as a parameter to the TELEPATHIC_DETECTION macro here is what the "@1@" gets replaced with.

default/scripting/species/common/telepathic.macros
Vezzra, line 7: Um, that's a macro parameter. They don't get explicitely defined, you just use them. When the macro is called, you need to pass a parameter, which will replace the "@1@" here. See my comment below the line where this macro is called for the Trith species.

I implemented it this way to provide more flexibility, so we can add that trait to other species with varying ranges...

Vezzra:
> This is I think what I had in mind, perhaps let them have Good visibility (same as Happybirthday) instead of Great (ie tone it down but not remove it), but would want to test it more.

Ok, made the adjustment (gave them good visibility), & squashed commits.

Happy testing :smile:

default/scripting/species/common/telepathic.macros
MatGB, line 7: Ah, yes, that new thing I haven't got around to learning to use yet ;-)

Morlic-fo:
I would strongly suggest to make this work for planets only (`Activation = Planet` would work, right?). 

Otherwise, owning a single Trith ship gives you quite some advantage. So you conquer a Trith planet, build some scouts and wipe out the entire population afterwards and still get all the benefits (i.e. let scouts follow your invasion fleets). Seems quite too strong for me.

default/scripting/species/common/telepathic.macros
```diff
@@ -0,0 +1,16 @@
+TELEPATHIC_DETECTION
+'''     EffectsGroup
+            description = "TELEPATHIC_DETECTION"
+            scope = And [
+                PopulationCenter
+                Not OR [
+                    Species name = Source.Species
+                    Species name = "SP_EXOBOT"
```
dbenage-cx: If other Trith are immune, should other telepaths be immune as well?
Could exclude by TELEPATHIC tag and maybe TELEPATHIC_IMMUNE for Exobot if you foresee other species having that trait later.

MatGB:
@Morlic-fo, that was my initial reaction as well when Geoff implemented this in his version (and was my intent when I suggested it), however, having been playing with it I think it's interesting and not overpowered. Thus far, I need to play a couple more games, the idea is that they get the xenophobic/self sustaining malus from other planets but are also aware of those planets.

Having been playing with it, it's not that powerful in practice, sure, you're more aware of inhabited worlds than before but it's not massively moreso. I'm not going to suggest this merges in immediately, but from a balance perspective I currently think it's fine, and a lot less of a problem than I initially thought it would be.

MatGB:
OK, partway into my second game I discovered a fairly major flaw: granting Partial visibility via scout ships completely breaks planetary stealth, which given it's a major tech line we want to do more with is a bad idea.

In testing I've switched it to Basic, which appears to be working quite well so far, will need to run more tests to be sure on it, might be a plan to grant colonies Partial but ships only Basic, only Basic for both might be enough though.

default/scripting/species/common/telepathic.macros
MatGB: Flavour wise, I'm happy with that as is, Trith learn to mask their thoughts, other telepaths don't have to/need to. Balance wise, the effect gives visibility of the planets that are causing the population malus, so it makes sense as is.  That's not to say it should be as is, merely that for initial testing _and_ for flavour this makes the most sense.

Vezzra: I think that's actually a piece of code that got carried over from the original Xenophobic macros, where this line made sure the xenophobic effect doesn't get applied to the xenophobic species itself (xenophobes freaking themselves out, so to speak).

Which means the line probably should be removed here - after all, why should telepaths not be able to detect other telepaths/other pops of the same species as themselves? Of course, we can add in a TELEPATHIC_IMMUNE tag for species that are supposed to be immune to telepathic detection (which makes sense for dumb automatons like exobots).

That would be my solution - comments?

Vezzra:
If basic visibility works well enough, I'd vote for both colonies and ships granting only basic.

MatGB:
I've played this through several games both with and against Trith now, and I like it, Basic visibility isn't overpowered and adds character to the Trith, and reducing their built in vision bonus means there's a stronger differentiation between them and Laenfa.  I vote merge this and close #621

default/scripting/species/common/telepathic.macros
Vezzra: Ok, decided to remove that condition entirely - which means, Telepathic Detection works on all populated planets regardless of species. This shouldn't be a problem here (on the contrary, excluding planets with the same species as the "detecting" species doesn't make sense), unlike the xenophobic effects, where your own species and exobots of course need to be excluded.

If we want certain species to be excluded from Telepathic Detection, we can still add this in later on, but for simplicities sake I'm going to remove that bit from this PR.

Vezzra:
Based on my own quick tests, giving only basic visibility (and have that for both planets and ships) seems to work well enough, and is a bit less powerful, which is good. I'm going to adjust the PR accordingly and merge.

PR #629, opened 2016-04-28
Pedia Detail Panel-Ship Design: Skip species assignment for 'Generic' ship entry

dbenage-cx:
When displaying a ship pedia entry, an error log message is generated:

```
[error] Client : Ship.cpp:587 : Ship::SetSpecies couldn't get species with name 
```

PR bypasses Ship->SetSpecies() for empty string (the initial value for Generic).  Generic entry is still displayed and stats are populated.

### DIFF
--- a/UI/EncyclopediaDetailPanel.cpp
+++ b/UI/EncyclopediaDetailPanel.cpp
@@ -2108,7 +2108,6 @@ namespace {
                 if (!this_ship->SpeciesName().empty())
                     additional_species.insert(this_ship->SpeciesName());
         std::vector<std::string> species_list(additional_species.begin(), additional_species.end());
-        species_list.insert(species_list.begin(), "");
         // detailed_description = GetDetailedDescription(temp, design, cost);
         detailed_description = GetDetailedDescriptionBase(design);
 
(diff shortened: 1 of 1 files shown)


UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2119,7 +2119,7 @@ namespace {
 
         // apply various species to ship, re-calculating the meter values for each
         for (std::vector< std::string >::iterator spec_it = species_list.begin(); spec_it != species_list.end(); spec_it++) {
-            temp->SetSpecies(*spec_it);
+            if (!(*spec_it == "")) temp->SetSpecies(*spec_it);
```
geoffthemedio: I'd prefer

```
if (!spec_it->empty())
    temp->SetSpecies(*spec_it);
```

(assuming it compiles...)

However, if the species to set is empty, then probably the species needs to be cleared before doing the following calculations, as otherwise it'll report the results for the previously-set species... Unless that's never a problem due to the species_list being initialized from a set that would have "" always sorted to the start... But some handling of this issue here still seems worthwhile for clarity...

dbenage-cx:
Handled a little differently to keep from adding more conditionals.

For explicitly clearing the species, would you prefer a bypass token in Ship.cpp:587 ? e.g.

```
if (species_name == "FUNCTIONAL_GENERIC_SPECIES")
    m_species_name = "";
```

geoffthemedio:
No need for that... just avoid the error message if the passed in species was "", as in that case there's no expectation that it will correspond to a parsed species.

PR #638, opened 2016-05-01
Add use the MapZoom hotkey to zoom research window techtree

LGM-Doyle:
1. Disable the map zoom hotkey when research or design window is
   visible.
2. Enable the map zoom hot key to zoom the research window tech tree.

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -6086,10 +6086,10 @@ namespace {
     /// On when the MapWnd window is visible and not covered
     //  by one of the full screen covering windows
     class VisibleMapWndCondition : public HotkeyCondition {
-        protected:
+    protected:
         const MapWnd& target;
 
-        public:
+    public:
         VisibleMapWndCondition(const MapWnd& tg) : target(tg) {}
         virtual bool IsActive() const {
             return target.Visible() && !target.InResearchViewMode() && !target.InDesignViewMode();
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -1377,6 +1377,12 @@ void MapWnd::GetSaveGameUIData(SaveGameUIData& data) const {
 bool MapWnd::InProductionViewMode() const
 { return m_in_production_view_mode; }
 
+bool MapWnd::InResearchViewMode() const
+{ return m_research_wnd->Visible(); }
+
+bool MapWnd::InDesignViewMode() const
+{ return m_design_wnd->Visible(); }
+
 ModeratorActionSetting MapWnd::GetModeratorActionSetting() const
 { return m_moderator_wnd->SelectedAction(); }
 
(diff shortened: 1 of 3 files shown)


UI/MapWnd.cpp
```diff
@@ -6069,6 +6075,22 @@ bool MapWnd::ZoomToSystemWithWastedPP() {
...
 
+namespace {
+
+    /// On when the MapWnd window is visible and not covered
+    //  by one of the full screen covering windows
+    class VisibleMapWndCondition : public HotkeyCondition {
+        protected:
+        const MapWnd& target;
+
+        public:
```
geoffthemedio: deindent to match "class"

UI/MapWnd.h
```diff
@@ -70,6 +70,16 @@ class MapWnd : public GG::Wnd {
       * production screen */
     bool                        InProductionViewMode() const;
 
+    /** returns true iff this MapWnd is visible and usable for interaction, but
+      * the allowed interactions are restricted to those appropriate to the
```
geoffthemedio: I don't understand the meaning of this comment after "but"

UI/TechTreeWnd.cpp
```diff
@@ -564,6 +565,9 @@ class TechTreeWnd::LayoutPanel : public GG::Wnd {
     void TreeZoomedSlot(int move);
     void TreeZoomInClicked();
```
geoffthemedio, line 569: having two sets of these functions seems unnecessary, but probably OK to avoid awkward uses of boost::bind to discard return values when not needed...

UI/MapWnd.h
LGM-Doyle: I think that I have clarified the comment.

UI/TechTreeWnd.cpp
LGM-Doyle, line 569: In order to not add boost::lambda to the dependencies I left this as is.  When freeorion moves to C++11 it should be cleaned up with a lambda function to add the dummy true return value.

LGM-Doyle:
I have just noticed the conflicts on the branch.  I will fix them and re-push

UI/MapWnd.h
```diff
@@ -65,19 +65,22 @@ class MapWnd : public GG::Wnd {
...
-      * production screen */
+    /** returns true if MapWnd is visible and usable behind a production window.
+     * MapWnd interactions are restricted to those appropriate to the production window */
     bool                        InProductionViewMode() const;
 
-    /** returns true iff this MapWnd is visible and usable for interaction, but
-      * the allowed interactions are restricted to those appropriate to the
-      * research screen */
+    /** returns true if MapWnd is visible and usable behind a research window.
+     * MapWnd interactions are restricted to those appropriate to the research window.
```
geoffthemedio, line 73: It's longer now, but I still don't know why the 2nd and later lines of this comment are here... What does this discussion have to do with this function?

LGM-Doyle, line 73: The original comment was on InProductionViewMode from 2005-09-01.

The use of InProductionViewMode is to disable Fleet::MouseHere() when in the production window with the map window visible.

The use of the two new functions is to disable map hotkeys when in the research/production windows.

The second part of the comment is to indicates expected usage of the functions to create 4 modal states for the user interface: map only, production, research and design.  In each user interface state the map window may have a different subset of available actions.

Issue #618, opened 2016-04-22
Unnecessary writes to config folder?


andresmrm:
I just got an SSD disk so I'm trying to reduce unnecessary writes to it.
So I set "saves" folder to an extenal HDD (since I knew autosave would do many writes).
After that, Freeorion was still writing ~2MB/turn on my SSD (where my main system is).

Using `find -mmin -1` I saw some files (logs?) in `~/.local/share/freeorion` were changing every turn, and they have a few MB. So I linked that folder to a folder on `/tmp` (my /tmp stays on RAM). Even after that it was still writing ~2MB/turn.

Then I tried to start Freeorion with `HOME=<path to a folder on external HDD> ./freeorion`.
It started to write the log files to this new "fake home", but was still writing ~2MB/turn to the SSD.

The last file that was still changing in my true home folder was `.config/freeorion/config.xml`. It has only a few KB, so I didn't think it was the cause of the 2MB/turn. But I tried to transfer `.config/freeorion` to RAM (`/tmp`) and it worked. No more writes to my SSD every turn.

I see two _possible_ bugs there:
- Starting Freeorion with a modified $HOME var changed the path to the `.local` folder, but not to the config folder.
- I can't see why it's writing ~2MB/turn to the config folder, if there is no file that big there. Maybe the autosave is first been saved there and than moved to the right folder?

Well, maybe this is a specific problem that will only affect me... Anyway, it seemed a strange behavior, so I thought it would be better to report. =P

dbenage-cx:
Hi @andresmrm, not a dev, but just a quick checklist:
- Assuming you do not have a `.config/freeorion/persistent_config.xml` file
- `$XDG_CONFIG_HOME` is set to your desired external directory, or is unset
- Temporarily rename or mv `.config/freeorion/config.xml`

andresmrm:
> - Assuming you do not have a .config/freeorion/persistent_config.xml file

No.

> - $XDG_CONFIG_HOME is set to your desired external directory, or is unset

Good point! Changing this var Freeorion started using the `.config` in the "fake home". Thanks!

> - Temporarily rename or mv .config/freeorion/config.xml

Not sure about what do you mean with this point...

So the remaining question is: why is it writing about 2MB/turn in the config folder?

Vezzra:
I've to admit I have no idea what's going on here... @andresmrm, I'm a bit confused: how can you see that 2MB get written to a certain folder without seeing which file(s) those 2MB get written to?

[...]

PR #639, opened 2016-05-01
Add a dirty bit to OptionsDB and only Commit() when something has cha…

LGM-Doyle:
…nged.

This fixes issue #618.

### DIFF
--- a/util/OptionsDB.h
+++ b/util/OptionsDB.h
@@ -308,10 +308,10 @@ private:
                const boost::any& default_value_, const std::string& description_,
                const ValidatorBase *validator_, bool storable_, bool flag_, bool recognized_);
 
-        // SetFromValue returns true if value changed
+        // SetFromValue returns true if this->value is successfully changed
         template <typename T>
-        bool            SetFromValue(const T& value);
-        // SetFromString returns true if value changed
+        bool            SetFromValue(const T& value_);
+        // SetFromString returns true if this->value is successfully changed
         bool            SetFromString(const std::string& str);
         std::string     ValueToString() const;
         std::string     DefaultValueToString() const;
--- a/util/OptionsDB.h
+++ b/util/OptionsDB.h
@@ -113,7 +113,8 @@ public:
     bool        OptionExists(const std::string& name) const
     { return m_options.find(name) != m_options.end() && m_options.at(name).recognized; }
 
-    /** write back the optionDB's state to the XML config file */
+    /** write back the optionDB's state to the XML config file
+        if it has changed since it was last saved. */
     void        Commit();
 
     /** validates a value for an option. throws std::runtime_error if no option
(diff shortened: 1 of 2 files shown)


util/OptionsDB.h
```diff
@@ -338,8 +344,23 @@ class FO_COMMON_API OptionsDB
...
     std::map<std::string, Option>   m_options;
     static OptionsDB*               s_options_db;
+    bool                            m_dirty; //< has OptionsDB changed since last Commit()
 
     friend OptionsDB& GetOptionsDB();
 };
 
+template <typename T>
+bool OptionsDB::Option::SetFromValue(const T& value_) {
+    if (value.type() != typeid(T))
```
geoffthemedio, line 354: this makes me a bit worried about portability...

I'd probably rather try the cast within a try-catch block and react to whether bad_any_cast is thrown...

util/OptionsDB.h
```diff
@@ -306,7 +308,11 @@ class FO_COMMON_API OptionsDB
                const boost::any& default_value_, const std::string& description_,
                const ValidatorBase *validator_, bool storable_, bool flag_, bool recognized_);
 
-        void            SetFromString(const std::string& str);
+        // SetFromValue returns true if value changed
+        template <typename T>
+        bool            SetFromValue(const T& value);
```
geoffthemedio: inconsistent parameter name here...

util/OptionsDB.h
```diff
@@ -338,8 +344,23 @@ class FO_COMMON_API OptionsDB
 
     std::map<std::string, Option>   m_options;
     static OptionsDB*               s_options_db;
+    bool                            m_dirty; //< has OptionsDB changed since last Commit()
 
     friend OptionsDB& GetOptionsDB();
 };
 
+template <typename T>
+bool OptionsDB::Option::SetFromValue(const T& value_) {
```
geoffthemedio, line 353: ...and here

util/OptionsDB.h
LGM-Doyle, line 354: I can change this but I'd like to keep it:
1. It was moved from OptionsDB.cpp:290, so it has been in the code base for some time.
2. It is the standard idiom for boost::any, right on the front page example.  (The very next example on the same page is try{} catch(boost::bad_any_cast &) )
3. It is less verbose.

util/OptionsDB.h
LGM-Doyle, line 353: Fixed.

LGM-Doyle:
@geoffthemedio you closed the issue, but did not merge this PR.  Was that intentional?

ghost:
I'm getting a consistent segfault with this patch when enabling fullscreen mode.  If I am reading things correctly, validator appears to be null.

bool changed =  validator->String(value) != validator->String(value_);

Linux 64 bit, compiled from git sources.

LGM-Doyle:
@spikethehobbitmage thanks for the bug report.  I've pushed a fix in PR #700.

PR #640, opened 2016-05-01
Add combat log details

LGM-Doyle:
This PR adds expandable details to the combat log.  The point is to make the specific rules governing combat available immediately in the combat log, in particular the interaction of shields and stealth.  It creates a framework for other events and writing the combat log in a more narrative structure.

Please comment on this idea and/or the implementation.

The motivation is in questions from beginners (like myself) on the forum.  Two often repeated questions are: Why was my massive fleet with mass drivers unable to damage the enemy?; Why did my stealth fleet suddenly get wiped out in round 2? The answer is usually read the manual.  This tries to provide the answer right where the player is already looking.

I've posted two partially expanded combat logs on the [http://freeorion.org/forum/viewtopic.php?f=6&t=10058](forum), showing shields preventing damage and a ship being destroyed in round 2 after it has been decloaked.  Note that damage is grouped by attacker.

In the shield case the details show attack strength, shield strength and final damage implying that 9 strength - 3 shield = 6 damage.

In the cloaking case, the details show which (friendly) planets and ships are cloaked at the start of combat and which ones decloaked and why.  It does not list enemy cloaked ships or planets.

To implement this, I have pushed the XML generation into the CombatEvent, which must generate a CombatLogDescription() and may generate CombatLogDetails() and SubEvents() if appropriate.  This better localizes the code for each type of event.  The combat log itself only checks each log and displays it as plain text or in an accordion window.

CombatSystem was changed in order to include the new events. ObjectAttackableBy[Monsters/Empire] was refactored to create ObjectUntargettableBy[Monsters/Empire] to create the initial stealth event.

Some code other than the CombatLog, CombatSystem and the events was changed.

The TextControl and Layout were changed so that TextControl can report a minusable size based on a proposed width and the Layout can then use that value.

AccordionPanel was changed to be more general: report a client area, settable interior color, report state of collapse.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -56,6 +56,10 @@ Yes
 NO
 No
 
+# ELLIPSIS is used as a placeholder for unexpanded content in the combat log window.
+ELLIPSIS
+...
+
 EMPIRE
 Empire
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -3868,6 +3868,17 @@ The%1%ship %2% was destroyed
 ENC_COMBAT_INITIAL_STEALTH_LIST
 %2% empire can not target:
 
+# %1 attacker %2 defender %3 empire
+ENC_COMBAT_STEALTH_DECLOAK_ATTACK
+%1% decloaks while attacking %2%.
+
+ENC_COMBAT_STEALTH_DECLOAK_ATTACK_1_EVENTS
+%2% empire can now target,
+
+ENC_COMBAT_STEALTH_DECLOAK_ATTACK_MANY_EVENTS
+%2% empire can now target:
+
+
 ###########################################
 # Combat Report #
 ###########################################
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -3878,6 +3878,18 @@ ENC_COMBAT_STEALTH_DECLOAK_ATTACK_1_EVENTS
 ENC_COMBAT_STEALTH_DECLOAK_ATTACK_MANY_EVENTS
 %2% empire can now target:
 
+ENC_COMBAT_PLATFORM_DAMAGE_1_EVENTS
+%2% damages the
+
+ENC_COMBAT_PLATFORM_DAMAGE_MANY_EVENTS
+%2% damages %1% targets:
+
+ENC_COMBAT_PLATFORM_NO_DAMAGE_1_EVENTS
+%2% could not damage the
+
+ENC_COMBAT_PLATFORM_NO_DAMAGE_MANY_EVENTS
+%2% could not damage %1% targets:
+
 
 ###########################################
 # Combat Report #
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -3837,6 +3837,9 @@ Combat at %1% on turn %2%:
 ENC_COMBAT_ATTACK_STR
 %1% attacks %2% and does %3% damage
 
+ENC_COMBAT_ATTACK_DETAILS
+Attack strength: %1%  Shield strength: %2%  Final damage: %3%
+
 ENC_COMBAT_ATTACK_SIMPLE_STR
 %1% attacks %2%
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -3865,6 +3865,9 @@ The%1%planet %2% was incapacitated
 ENC_COMBAT_DESTROYED_STR
 The%1%ship %2% was destroyed
 
+ENC_COMBAT_INITIAL_STEALTH_LIST
+%2% empire can not target:
+
 ###########################################
 # Combat Report #
 ###########################################
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -8724,6 +8724,9 @@ Interstellar Lighthouse
 SPY_LIGHTHOUSE_DESC
 By using a complex pattern of EM pulses and particle streams to reveal the location of objects in a system, it is possible for even very stealthy objects to be illuminated and detected with moderate detection equipment.
 
+DEF_DEFENSE
+Planetary Defense
+
 DEF_DEFENSE_NET_1
 Planetary Defense Network 1
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -207,6 +207,63 @@ Unowned
 NOWHERE
 Cannot be produced
 
+################################## 
+# FORMAT_LIST items are for lists of the form
+# FORMAT_LIST_HEADER FORMAT_LIST_x_ITEMS
+# i.e.
+# There is none.
+# There is only: banana
+# There are: apple and banana
+##################################
+
+FORMAT_LIST_DEFAULT_PLURAL_HEADER
+There are:
+
+FORMAT_LIST_DEFAULT_SINGLE_HEADER
+There is one:
+
+FORMAT_LIST_DEFAULT_EMPTY_HEADER
+There is none.
+
+FORMAT_LIST_DEFAULT_DUAL_HEADER
+There are:
+
+FORMAT_LIST_0_ITEMS
+%1%
+
+FORMAT_LIST_1_ITEMS
+%1% %2%.
+
+FORMAT_LIST_2_ITEMS
+%1% %2% and %3%.
+
+FORMAT_LIST_3_ITEMS
+%1% %2%, %3% and %4%.
+
+FORMAT_LIST_4_ITEMS
+%1% %2%, %3%, %4% and %5%.
+
+FORMAT_LIST_5_ITEMS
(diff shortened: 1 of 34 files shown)


default/stringtables/en.txt
```diff
@@ -207,6 +207,63 @@ Unowned
...
+##################################
+
+FORMAT_LIST_DEFAULT_PLURAL_HEADER
+There are:
+
+FORMAT_LIST_DEFAULT_SINGLE_HEADER
+There is one:
+
+FORMAT_LIST_DEFAULT_EMPTY_HEADER
+There is none.
```
geoffthemedio: There are none.

default/stringtables/en.txt
```diff
@@ -3784,6 +3844,32 @@ The%1%planet %2% was incapacitated
...
+%1% decloaks while attacking %2%.
+
+ENC_COMBAT_STEALTH_DECLOAK_ATTACK_1_EVENTS
+%2% empire can now target,
+
+ENC_COMBAT_STEALTH_DECLOAK_ATTACK_MANY_EVENTS
+%2% empire can now target:
+
+ENC_COMBAT_PLATFORM_DAMAGE_1_EVENTS
+%2% damages the
```
geoffthemedio: lose the "the"

default/stringtables/en.txt
```diff
@@ -207,6 +207,63 @@ Unowned
...
+FORMAT_LIST_0_ITEMS
+%1%
+
+FORMAT_LIST_1_ITEMS
+%1% %2%.
+
+FORMAT_LIST_2_ITEMS
+%1% %2% and %3%.
+
+FORMAT_LIST_3_ITEMS
```
geoffthemedio, line 244: Rather than dealing with all these list formatting issues, why not just have a space-separated list with no extra special formatting required to add an extra item to the end? works well enough in the pedia...

default/stringtables/en.txt
```diff
@@ -3756,6 +3813,9 @@ Combat at %1% on turn %2%:
 ENC_COMBAT_ATTACK_STR
 %1% attacks %2% and does %3% damage
 
+ENC_COMBAT_ATTACK_DETAILS
+Attack strength: %1%  Shield strength: %2%  Final damage: %3%
```
geoffthemedio: Lose "final"

default/stringtables/en.txt
```diff
@@ -3784,6 +3844,32 @@ The%1%planet %2% was incapacitated
...
 
+ENC_COMBAT_INITIAL_STEALTH_LIST
+%2% empire can not target:
+
+# %1 attacker %2 defender %3 empire
+ENC_COMBAT_STEALTH_DECLOAK_ATTACK
+%1% decloaks while attacking %2%.
+
+ENC_COMBAT_STEALTH_DECLOAK_ATTACK_1_EVENTS
+%2% empire can now target,
```
geoffthemedio: , or :

default/stringtables/en.txt
```diff
@@ -3784,6 +3844,32 @@ The%1%planet %2% was incapacitated
...
+%2% empire can now target:
+
+ENC_COMBAT_PLATFORM_DAMAGE_1_EVENTS
+%2% damages the
+
+ENC_COMBAT_PLATFORM_DAMAGE_MANY_EVENTS
+%2% damages %1% targets:
+
+ENC_COMBAT_PLATFORM_NO_DAMAGE_1_EVENTS
+%2% could not damage the
```
geoffthemedio: no the

util/i18n.h
```diff
@@ -44,4 +45,106 @@ FO_COMMON_API std::string DoubleToString(double val, int digits, bool always_sho
...
+/** returns a boost::format pre-filled with a list of up to 10 items with some introductory text similar to
+ These are 3 fruit: apples, pears and bananas.
+ The Container types need to support size(), begin() and end().
+ The headers are boost::format strings fed the size of the words list and then any header words passed in.
+ TODO: Adopt an out of house i18n framework and handle of empty, singular, pairs, triplets etc. correctly. */
+
+template<typename HeaderContainer, typename ListContainer>
+FO_COMMON_API boost::format FlexibleFormatList(
+    const HeaderContainer& header_words
+    , const ListContainer& words
```
geoffthemedio: , to end of line, not start of line

util/i18n.h
```diff
@@ -44,4 +45,106 @@ FO_COMMON_API std::string DoubleToString(double val, int digits, bool always_sho
...
+        header_template = plural_header_template;
+    }
+
+    boost::format header_fmt = FlexibleFormat(header_template) % boost::lexical_cast<std::string>(words.size());
+
+    for (typename HeaderContainer::const_iterator it = header_words.begin(); it != header_words.end(); ++it) {
+        header_fmt % *it;
+    }
+
+    std::string template_str = words.size()<=10
```
geoffthemedio: spaces around <=

util/i18n.h
```diff
@@ -44,4 +45,106 @@ FO_COMMON_API std::string DoubleToString(double val, int digits, bool always_sho
...
+        fmt % *it;
+    }
+
+    return fmt;
+}
+
+template<typename Container>
+FO_COMMON_API boost::format FlexibleFormatList(const Container& words) {
+    return FlexibleFormatList(std::vector<std::string>(), words
+                              , UserString("FORMAT_LIST_DEFAULT_PLURAL_HEADER")
```
geoffthemedio: , to line before

Vezzra:
This PR has already received very positive feedback in the forum thread, and I also think that this is a good enhancement to the combat log.

Given that this is a quite extensive change I want to get opinions on wether to include it in 0.4.6 or postpone it. So everyone, please give your feedback on this. :smiley:

MatGB:
If it works put it in, haven't tested it but it's the sort of thing that most helps new players, who mostly start on Release versions and it might be ages until the next one.

default/stringtables/en.txt
LGM-Doyle, line 244: This was intended to be easier for the translators.  It makes no assumptions about left to right or right to left.  Single, dual, many and none cover most languages pluralities.  

It is easy from the C++ side, both the header and the list parts will iterate over anything that yields a value that boost::format understands:list, set, unordered_set.

Vezzra:
Alright, sounds reasonable, setting milestone to 0.4.6.

geoffthemedio:
Initial layout of the log is broken:
combat log layout

Sub-entries should have the left-edge indented from the enclosing entry.
need indent

Maybe also put the click point to expand/collapse on the left, instead of the right.

Maybe also organize the events by round, then by owner of the attacker, not order in which the events were generated. This helps make it not seem to players like there is any order of events within a combat round.

LGM-Doyle:
I thought that my last push [   PR #640 changes](https://github.com/freeorion/freeorion/pull/640/commits/e7f6edd510bf23dac6d6b74eea7b72eff8bdee9b) fixed the initial layout problems.  

Am I lucky enough that you didn't pull it?

I put the expansion/collapse on the right so that it had the same look feel as the Planet SidePanel and things were more consistent.  

Do you still want it moved to the left?

BTW, I created a bug in the Planet SidePanel, one of the expand/collapse buttons disappears, although it still works correctly.  I will fix that in the next push.

geoffthemedio:
I built the latest version.
Putting the expand/collapse on the left is more consistent with the objects list, which this more closely resembles.

LGM-Doyle:
I've fixed the bug with the planet side panel accordion expander sometimes disappearing.
I've made the empire tags link to the Pedia.
I've moved the expansion symbols to the left.
I've made simultaneous events (attacks, incapacitations and fighter launches/recoveries) display in the order own empire, ALL_EMPIREs and then other empires in id order.
And I hope I have finally fixed the re-flowing of the logs on initial display and re-sizing.

geoffthemedio:
Looking better.

There needs to be a gap between the expand/collapse buttons and the text though, especially since that text can be a clickable link.

The log should scroll to the top when first clicked; presently the vertical scroll position seems to be retained from the last viewed log, even if there's not enough stuff to show to scroll down that far.

Might also be good if the detail for an attack indicated what weapon it was, by name.

LGM-Doyle:
I'm working on these.  

I've also found that large combats (>150 combatants) load very slowly.  The problem is the parsing of the XML tags in `GG::Font::DetermineLinesImpl()` using regular expressions.  

Fixing GG::Font is too much work, so I'm working around it by making the display of logs lazy.

 It will take a couple of days.

LGM-Doyle:
Updated.

geoffthemedio:
Can you get rid of the extra space in this message, before the target name?

extra_space

LGM-Doyle:
Fixed.

I rebased the change into the PR #640 changes commit which was where the extra spaces were introduced into all of the single item events.

geoffthemedio:
I have some remaining stylistic quibbles, but OK for now.

Issue #534, opened 2016-03-18
Fatal exceptions when loading old saves not crashing FO


cpeosphoros:
I just did a "git pull", recompiled, etc.

When trying to load a recent save from my latest build, hoping it would still work, it didn't (as expected, it happens most of the time.)

What didn't happen was the expected FO crash.

The client UI's Messages window says:

```
Creating AI clients
Loading
```

And stays there forever.

The terminal from which I launched FO says:

```
$ ./freeorion
FreeOrion server waiting for network events
main() caught exception(std::runtime_error): Invalid empire ID specified for order.
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
main() caught exception(std::runtime_error): Invalid empire ID specified for order.
main() caught exception(std::runtime_error): Invalid empire ID specified for order.
main() caught exception(std::runtime_error): Invalid empire ID specified for order.
```

And also stays there forever.

Shouldn't those exceptions just crash everything and throw me back to the terminal prompt?

Possibly related to: https://github.com/freeorion/freeorion/issues/524

Comment from Geoff on the forum post:

> It already happens in some cases, that if the server crashes, the client notices and aborts back to the intro screen. Sometimes the servers ends up hung waiting for some message that will never come, though, leading to this situation. It would be good to add some periodic ping messages between clients and server, to verify everything's still connected, but that hasn't been implemented.

---

All contributions are released under GPL or LGPL v2 or later, or under appropriate Creative Commons licence, consistent with project guidelines.

Vezzra:
Ok, I boldly assume that #642 fixed this issue and therefore close it. Should it turn out that it isn't fixed, please report and we can reopen.

PR #642, opened 2016-05-02
Fix bug524 wait for ai crash

### LINKED ISSUE #524: Server keeps waiting for crashed AI on startup
So I was messing up an import in AI-scripts - AI could not be initialized:

```
2016-03-13 18:37:02.198722 [error] AI : AIClientApp.cpp:51 : AIClientApp::AIClientApp : Failed to initialize Python AI.  Exiting.
2016-03-13 18:37:02.198722 [error] AI : camain.cpp:61 : main() caught exception(std::runtime_error): AIClientApp::AIClientApp : Failed to initialize Python AI.  Exiting.
```

However, the server kept waiting:

```
2016-03-13 18:37:01.244667 [debug] Server : ai-path set to 'AI'
2016-03-13 18:37:01.244667 [debug] Server : ai-config not set.
2016-03-13 18:37:01.248668 [debug] Server : done starting C:\Users\FabianAdmin\FreeOrionDevelopment\FreeOrion\freeorionca.exe
2016-03-13 18:37:01.248668 [debug] Server : (ServerFSM) WaitingForSPGameJoiners.CheckStartConditions
2016-03-13 18:38:13.496800 [debug] Server : ServerNetworking::DisconnectImpl : disconnecting player 1
```

(Last Entry: Me shutting down FO)

Expected Result: Server shuts down if the AI can not be initialized and some error gets displayed.

LGM-Doyle:
I think this fixes #524 and #534. 

It produces the MessageBox saying "ERROR: Python AI for AI_1 crashed."  as request by @Morlic-fo when I used the python code provided by @Cjkjvfnby during initialization of the AI.

### DIFF
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -6,6 +6,7 @@
 #include "../../util/OptionsDB.h"
 #include "../../util/Directories.h"
 #include "../../util/Serialize.h"
+#include "../../util/i18n.h"
 #include "../../network/Message.h"
 #include "../util/Random.h"
 
@@ -44,14 +45,6 @@ AIClientApp::AIClientApp(const std::vector<std::string>& args) :
 
     InitLogger(AICLIENT_LOG_FILENAME, "AI");
     DebugLogger() << PlayerName() + " logger initialized.";
-
-    PythonAI* python_ai = new PythonAI();
-    if (!(python_ai->Initialize())) {
-        const std::string err_msg = "AIClientApp::AIClientApp : Failed to initialize Python AI.  Exiting.";
-        ErrorLogger() << err_msg;
-        throw std::runtime_error(err_msg);
-    }
-    m_AI = python_ai;
 }
 
 AIClientApp::~AIClientApp() {
@@ -77,6 +70,29 @@ const AIBase* AIClientApp::GetAI()
 { return m_AI; }
 
 void AIClientApp::Run() {
+
+    ConnectToServer();
+
+    StartPythonAI();
+
+    // join game
+    Networking().SendMessage(JoinGameMessage(PlayerName(), Networking::CLIENT_TYPE_AI_PLAYER));
+
+    // respond to messages until disconnected
+    while (1) {
+        if (!Networking().Connected())
+            break;
+        if (Networking().MessageAvailable()) {
+            Message msg;
+            Networking().GetMessage(msg);
+            HandleMessage(msg);
+        } else {
+            boost::this_thread::sleep_for(boost::chrono::milliseconds(250));
+        }
+    }
+}
+
+void AIClientApp::ConnectToServer() {
     // connect
     const int MAX_TRIES = 10;
     int tries = 0;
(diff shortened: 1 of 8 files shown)


client/AI/AIClientApp.cpp
```diff
@@ -77,6 +70,29 @@ const AIBase* AIClientApp::GetAI()
 { return m_AI; }
 
 void AIClientApp::Run() {
+
```
adrianbroher: Unnecessary whitespace.

server/ServerApp.cpp
```diff
@@ -318,7 +318,6 @@ int ServerApp::GetNewDesignID()
 
 void ServerApp::Run() {
     DebugLogger() << "FreeOrion server waiting for network events";
-    std::cout << "FreeOrion server waiting for network events" << std::endl;
```
geoffthemedio, line 321: Why?

With this line, there is some indication of what's happening if the server process is run from command line.

server/ServerFSM.cpp
```diff
@@ -1321,6 +1370,11 @@ sc::result WaitingForTurnEndIdle::react(const SaveGameRequest& msg) {
     return transit<WaitingForSaveData>();
 }
 
+sc::result WaitingForTurnEndIdle::react(const Error& msg) {
```
geoffthemedio: This appears to be overriding the containing state equivalent, but with nothing changed. Is it necessary?

client/AI/AIClientApp.cpp
```diff
@@ -96,22 +112,21 @@ void AIClientApp::Run() {
...
-        } else {
-            boost::this_thread::sleep_for(boost::chrono::milliseconds(250));
-        }
+void AIClientApp::StartPythonAI() {
+    PythonAI* python_ai = new PythonAI();
+    if (!(python_ai->Initialize())) {
+        const std::string err_msg = "AIClientApp::AIClientApp : Failed to initialize Python AI.  Exiting Soon.";
+        ErrorLogger() << err_msg;
+        Networking().SendMessage(
+            ErrorMessage(str(FlexibleFormat(UserString("ERROR_PYTHON_AI_CRASHED")) % PlayerName()) , true));
```
geoffthemedio, line 125: The server presumably already knows the player name... Is this necessary?

If the text is going to be shown in the human client, I'd rather do the stringtable lookup in that process, rather than before sending the message to the server.

client/human/HumanClientFSM.cpp
```diff
@@ -445,19 +469,23 @@ boost::statechart::result PlayingGame::react(const Error& msg) {
...
+
+    //Note: Any transit<> transition must occur before the MessageBox().
+    // MessageBox blocks and can allow other events to transit<> to a new state
+    // which makes this transit fatal.
+    boost::statechart::result retval = fatal ? transit<IntroMenu>() : discard_event();
     ClientUI::MessageBox(UserString(problem), true);
 
     if (fatal) {
-        Client().GetClientUI()->GetMessageWnd()->HandleGameStatusUpdate(UserString("RETURN_TO_INTRO") + "\n");
+        Client().EndGame(true);
```
geoffthemedio, line 482: I renamed the AI script directory, then started freeorion, then quickstarted. I got a crash around here, though I'm now unable to reproduce it.

geoffthemedio:
If a first game is run, then a second that has an AI crash, or a game gets far enough into init to have opened the messages / empires window, then the AI crashes, I sometimes get the windows left open when back on the intro screen...
windows_left_open

server/ServerApp.cpp
LGM-Doyle, line 321: I didn't think of running the server process standalone.  I've restored the line.

server/ServerFSM.cpp
LGM-Doyle: HandleErrorMessage() doesn't return if the error is fatal.  I've added a comment to its definition.

client/AI/AIClientApp.cpp
LGM-Doyle, line 125: At this point in the initialization, before the AI has tried to join the game the server's PlayerConnection is not initialized with the AI id or name.  It gets initialized when the AI joins.  I added a comment to clarify and left the code as is.

LGM-Doyle:
I hid the messages and empire windows, which are owned by the client and not the MapWnd, during the transition.

server/ServerFSM.cpp
geoffthemedio: How is that relevant to my question that it follows (regarding overriding functions seemingly unnecessarily)?

server/ServerApp.cpp
geoffthemedio, line 321: It's useful for attaching the debugger to catch something before whatever state the server is in on the galaxy setup screen.

server/ServerFSM.cpp
LGM-Doyle: I misunderstood your question.

You were correct. The handlers for the inner states were redundant.  I have removed them and pushed the changes.  

I re-tested for the same cases we have already used, starting and re-starting the AI a few times.

PR #643, opened 2016-05-03
Add missing Unresearchable techs for AI

dbenage-cx:
Expands temporary AIDependencies list of unresearchable techs for all current techs.


Cjkjvfnby:
In the future need to reformat file in next way: all collection have single item on line with trailing coma. This will allow have smaller diffs. Iwill do it when be back from vacation.

PR #644, opened 2016-05-04
Closable pedia for research window

dbenage-cx:
Allows pedia window in research screen to be toggled opened/closed.

Menu pedia button behavior consistent with production window.


geoffthemedio:
Can you add similar pedia lookup right-click commands for producible items on the production screen, and to both production and research queue item right click menus?

dbenage-cx:
Will do

PR #645, opened 2016-05-05
Add right click context for production wnd pedia

dbenage-cx:
Adds a right click context for production window pedia lookup.


geoffthemedio:
Doesn't seem to work for me... I get no pedia option in the production item list when right clicking on an item...

dbenage-cx:
Added to queue as well (just in case), sorry for the wasted effort.
I don't see anything hindering cross-platform outright, will start a build on windows tonight and figure out what I missed.

dbenage-cx:
Working here?
prod_3

geoffthemedio:
Seems to work for me now. Would be good to also have pedia lookup on tech queue items as well.

Also, some handling of a case where the item.build_type isn't one of the two cases you've handled would be good. Add an error line to the log and abort?

PR #649, opened 2016-05-07
Prohibit spawning Scrying Sphere in system with Exp. Outpost

### LINKED ISSUE #647: Scrying sphere needs to be gated to not appear on Experimentor homeworld
Multiple ways this could be done but it's definitely not a good combination.

dbenage-cx:
Fixes #647


geoffthemedio:
looks reasonable

Issue #537, opened 2016-03-19
Guardian hulls need sensor parts


cpeosphoros:
Refer to: http://www.freeorion.org/forum/viewtopic.php?f=15&t=10006&view=unread&sid=96ccf93875079e58510e2de75d3978b7#unread

MatGB said:

> Stealthed scoutships should be able to sneak relatively easily passed Sentries and Maintenance ships, but not Wardens, all currently get the same detection (based on vision range for unowned stuff). I recently set it to 50, but ideally it should start at 10 and go up for different things, so individual parts are needed.

Vezzra:
Implemented by #650.

PR #650, opened 2016-05-08
Reduce detection strength of guard monsters

dbenage-cx:
Guard monster vision changes regarding #537.

For Release milestone, in the event new parts are not polished in time.

### DIFF
--- a/default/scripting/ship_hulls/monster/SH_IMMOBILE_FACTOR.focs.txt
+++ b/default/scripting/ship_hulls/monster/SH_IMMOBILE_FACTOR.focs.txt
@@ -15,7 +15,7 @@ Hull
     Unproducible
     location = All
     effectsgroups = [
-        [[GOOD_VISION]]
+        [[UNOWNED_OWNED_VISION_BONUS(WEAK,10,10)]]
         EffectsGroup
             scope = Source
             activation = Not Turn low = 1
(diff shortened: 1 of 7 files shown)


default/scripting/ship_hulls/monster/SH_IMMOBILE_FACTOR.focs.txt
```diff
@@ -15,7 +15,7 @@ Hull
     Unproducible
     location = All
     effectsgroups = [
-        [[GOOD_VISION]]
+        [[UNOWNED_OWNED_VISION_BONUS(WEAK,15,10)]]
```
MatGB: Can't have value 15, detection strengths need to always be a multiple of 10, stealth values then fit between them (normally ending in 5) to ensure there's never doubt about whether something is higher or lower.

dbenage-cx: rebased.  dropped to 10 here and raised to 40 for Acirema/Sentinel (catching +stealth organics without cloak)

MatGB:
Sensible, merged in, thanks—I do of course reserve the right to change all the numbers when I get around to rebalancing the whole thing but let's face it that'll be awhile ;-)

Vezzra:
Assuming this addresses #537 adequately I'm going to close #537.

MatGB:
Knew I'd forgotten to do something. Yup, does what was needed and then some, will be easier to change moving forward as well, which is nice.

Issue #563, opened 2016-03-25
Star systems sometimes too far apart


Vezzra:
Reportedly, with certain settings (mostly spiral or cluster layouts), excessively long starlanes get generated, which surpass the max starlane length limit by far. This can cause issues with the Experimentors, as well as other game mechanics/elements which are balanced with a certain starlane length limit in mind.

EDIT: This is caused by star systems which are placed too far away from any other system during universe generation, which is what needs to be fixed to prevent the excessively long starlanes /EDIT

http://www.freeorion.org/forum/viewtopic.php?p=83137#p83137

geoffthemedio:
I don't consider this a bug... The limit is intentionally not enforced if it would disconnect a system from the rest of the galaxy, and sometimes a long lane will be needed.

Vezzra:
Yes, I know. The issue is not that the limit isn't _strictly_ enforced, and that you can get the occasional extra long starlane. The problem I see is that apparantly sometimes these extra long starlanes can get really _excessively_ long, like twice the max length limit or even more.

I assumed that's not within the tolerances you had in mind when not enforcing the limit in the starlane generation code (correct me if I'm wrong), which is why I consider this an issue that should be addressed. I don't know how to classify this other than as a bug... we can, of course, introduce some kind of label for things which are issues we want to address, but are not bugs in the strict sense of the word - any ideas?

geoffthemedio:
I didn't have any particular tolerances in mind... Sometimes a lane is needed.

This might be better described as an issue with the system location generation, rather than the lane itself being too long, as the latter in a consequence of the former.

MatGB:
> This might be better described as an issue with the system location generation,

Agreed, but it is still an issue—I've commented on similar before in Clusters where it's possible to have a system or cluster so far away there's a stupidly restrictive chokepoint that can create very unfun games, extremely long starlanes are part of that problem

Vezzra:
> This might be better described as an issue with the system location generation

Yep, you're right of course. Edited the issue title and OP accordingly.

Vezzra:
Fixed by #651.

PR #567, opened 2016-03-27
update script for colony building entries

dbenage-cx:
Missed free standing python utility: https://github.com/freeorion/freeorion/pull/559#issuecomment-201966609

PR #651, opened 2016-05-08
Fix bug #563 stars systems too far apart

LGM-Doyle:
Fixes issue #563.

This fixes the system placement by using disjoint sets after initial placement to find isolated groups of systems and placing addition systems between the closest two systems to stitch them together.  

It imposes an additional requirement that MAX_STARLANE_LENGTH >= 2 \* MIN_SYSTEM_SEPARATION so that there is always room for the extra system to join the groups.

I also added Timers from the AI utilities to check that this doesn't add undue overhead.  It increased the time to create a 999 star system by about 0.25%.  

For large galaxies most of the time is consumed in monster placement by the Not WithinStarlaneJumps of Empires condition.  I reduced the reporting of monster possible placement to only the first 100 systems which improved timing by ~40% for large 999 system galaxies.

Overall, this PR guarantees MAX_STARLANE_LENGTH separation and speeds up large galaxy generation by about 40%.


default/python/universe_generation/empires.py
```diff
@@ -239,6 +238,27 @@ def add_planets_to_vicinity(vicinity, num_planets, gsd):
...
 
+def get_systems_within_jumps(origin_system, jumps):
+    """
+    Returns all systems within jumps jumps of system origin_system (including origin_system).
+    If jumps is 0, return list that only contains system origin_system.
+    If jumps is negative, return empty list.
+    """
+    if jumps < 0:
+        return []
+    matching_systems = [origin_system]
```
Cjkjvfnby: This is always same as next_origin_systems.

default/python/universe_generation/empires.py
```diff
@@ -239,6 +238,27 @@ def add_planets_to_vicinity(vicinity, num_planets, gsd):
...
+    next_origin_systems = [origin_system]
```
Cjkjvfnby: May be use set?

Vezzra:
Ok, that's a lot to review. Will take me some time, pls be patient... :wink:

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
         print "Adjacency Grid: width", self.width, ", cell size", self.cell_size
 
+    def cell(self, pos):
+        return (int(pos[0] / self.cell_size), int(pos[1] / self.cell_size))
+
     def insert_pos(self, pos):
-        self.grid[int(pos[0] / self.cell_size)][int(pos[1] / self.cell_size)].append(pos)
+        self.grid[self.cell(pos)].append(pos)
+
+    def remove(self, pos):
```
Cjkjvfnby: Name is not in same style as insert method

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+
+        # stores None or the parent
+        self.parent = None
+
+        # starts at 0, never decreases and only increases
+        # when this object is the parent and two equal rank sets are merged
+        self.rank = 0
+
+    def bind_parent(self, parent):
+        assert self != parent
```
Cjkjvfnby: You should change it to 'is parent' to check if it is same object or define compare method.

default/python/universe_generation/galaxy.py
```diff
@@ -45,32 +307,10 @@ def calc_universe_width(shape, size):
...
-    next_origin_systems = [origin_system]
-    while jumps > 0:
-        origin_systems = list(next_origin_systems)
-        next_origin_systems = []
-        for system in origin_systems:
-            neighbor_systems = [s for s in fo.sys_get_starlanes(system) if s not in matching_systems]
-            next_origin_systems.extend(neighbor_systems)
-            matching_systems.extend(neighbor_systems)
-        jumps -= 1
-    return matching_systems
```
Vezzra, line 66: Although it's only used be the `empire.py` module, this function has been deliberatly placed here in the `galaxy.py` module. After all, this module is where all the stuff that deals with the galaxy map belongs, and this function is provided not only for what it's used in `empire.py`, but potential future uses in other modules as well (e.g. users who want to customize/tweak the universe generation scripts might need it for whatever things the want to accomplish).

So IMO it should stay here.

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+    See https://en.wikipedia.org/wiki/Disjoint-set_data_structure
+    """
+    def __init__(self):
+        self.dsets = {}
+
+    def add(self, pos):
+        if pos not in self.dsets:
+            self.dsets[pos] = DSet(pos)
+
+    # Creates a link between the sets containing p1 and p2
```
Cjkjvfnby: Move this to docstring.

default/python/universe_generation/galaxy.py
Vezzra: Yep. Either call it `remove_pos` or rename `insert_pos` to `insert`. I strongly prefer the former, as it more clearly says what it does (inserting/removing a _position_).

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+    return [p3] + stitching_positions(p1, p3) + stitching_positions(p2, p3)
+
+
+def enforce_max_distance(positions, adjacency_grid):
+    """
+    Inserts extra planets to guarrantee that no clusters of planets
+    are separated from all other clusters of planets by more than
+    universe_tables.MAX_STARLANE_LENGTH
+
+    Find all clusters of planets.
```
Cjkjvfnby: find, insert? Can you rephrase docstring

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+    For each cluster
+      Remove it from the adjacency grid and find its nearest neighbor
+      Add planets to make all spacing less than the maximum
+    """
+    # Find all clusters
+    clusterer = Clusterer(positions, adjacency_grid)
+    clusters = sorted(clusterer.clusters(), key=len)
```
Cjkjvfnby: Can we just replace Clasterer class with function?

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
 
 import freeorion as fo
 import util
 import universe_tables
 
 
-class AdjacencyGrid:
+class AdjacencyGrid(object):
     def __init__(self, universe_width):
+        assert universe_tables.MIN_SYSTEM_SEPARATION < universe_tables.MAX_STARLANE_LENGTH
```
Vezzra, line 13: In the description of your PR you say that this additional requirement to be
`MAX_STARLANE_LENGTH >= 2 * MIN_SYSTEM_SEPARATION`
So either there is a `*2` missing here after `MIN_SYSTEM_SEPARATION`, or the description is wrong, or I'm missing something obvious...?

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+        (_, p1, p2) = min([(util.distance(pos, nn), pos, nn)
+                           for nn in adjacency_grid.nearest_neighbor(pos)
+                           for pos in smallest_cluster])
+
+        for pos in smallest_cluster:
+            adjacency_grid.insert_pos(pos)
+        # Add extra planets
+        extra_planets = stitching_positions(p1, p2)
+        for i_extra, p3 in enumerate(extra_planets):
+            print "Adding", i_extra, "of", len(extra_planets), "extra planets at",\
```
Cjkjvfnby: Use string formatting

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+        return (int(pos[0] / self.cell_size), int(pos[1] / self.cell_size))
+
     def insert_pos(self, pos):
-        self.grid[int(pos[0] / self.cell_size)][int(pos[1] / self.cell_size)].append(pos)
+        self.grid[self.cell(pos)].append(pos)
+
+    def remove(self, pos):
+        self.grid[self.cell(pos)].remove(pos)
+
+    def _square_indices_containing_cell(self, (cell_x, cell_y), radius):
```
Vezzra, line 33: Please add docstrings also for these private functions. Applies to all subsequent functions/classes.

default/python/universe_generation/timers.py
```diff
@@ -0,0 +1,148 @@
+# TODO Find a common location for debug tools used by all python AI, and engine
```
Cjkjvfnby: Not only debug. Making new tiket about it would be great.

default/python/universe_generation/universe_tables.py
```diff
@@ -141,6 +141,7 @@
 }
 
 # The maximum length of any starlane the galaxy generator can create (in uu).
+# MAX_STARLANE_LENGTH >= 2 * MIN_SYSTEM_SEPARATION
```
Cjkjvfnby: ?

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+        while ring and i_radius <= i_radius_max:
+            inner = outer
+            outer = set(self._square_indices_containing_cell(cell, i_radius))
+            ring = outer - inner
+            yield ring
+            i_radius += 1
+
+    # All neighbors within max_dist
+    def neighbors(self, p):
+        _neighbors = []
```
Vezzra, line 58: The comment above should be a docstring.

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+        for i, ring in enumerate(self._generate_rings_around_point(p, self.max_dist)):
+            if i <= i_radius_close_enough:
+                _neighbors += [pos for cell in ring for pos in self.grid[cell]]
+            else:
+                _neighbors += [pos for cell in ring for pos in self.grid[cell] if
+                               util.distance(pos, p) <= self.max_dist]
+        return _neighbors
+
+    # nearest neighbor at any distance
+    def nearest_neighbor(self, p):
```
Vezzra, line 68: Ditto for this and all subsequent cases.

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+
+    def inc_rank(self):
+        self.rank += 1
+
+
+class DisjointSets(object):
+    """
+    A set of disjoint sets.
+
+    It suuports the operations of:
```
Vezzra: "supports" (typo)

Cjkjvfnby:
If you can submit your cleanup commit in separate pull request it will be great.

Can you also add real time to persentage. 40% says nothing for me. It can depend on python version/platform.

I am from mobile now and just scroll code. I will back from vacation at 11 may.

Vezzra:
Just did a first very quick pass, made a few comments. More thorough passes to come. :smiley:

default/python/universe_generation/empires.py
```diff
@@ -239,6 +238,27 @@ def add_planets_to_vicinity(vicinity, num_planets, gsd):
         num_planets -= 1
 
 
+def get_systems_within_jumps(origin_system, jumps):
+    """
+    Returns all systems within jumps jumps of system origin_system (including origin_system).
```
Cjkjvfnby: Use double apostrophe to wrap arguments inside docstrings.

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,297 @@
...
+                p3, "to fix max spacing error between", p1, "and", p2
+            adjacency_grid.insert_pos(p3)
+            positions.append(p3)
+            clusterer.add(p3)
+        clusters = sorted(clusterer.clusters(), key=len)
+
+        if len(clusters) == 1:
+            print "All systems now positioned in a single connected cluster"
+        else:
+            print len(clusters), "clusters separated by more the MAX_STARLANE_LENGTH.  Continuing to fill gaps."
```
Vezzra: The use of "planet" in the above function when actually _positions_ are meant is a bit confusing. Pls rename everything "planet*" to "pos*" or "position*".

LGM-Doyle:
Thank you for taking the time to review this PR.

You've found several mistakes.  I'm adding comments to the line notes, where I would like to discuss things further.  

I'll make all of the changes at once after we have consensus.

default/python/universe_generation/empires.py
```diff
@@ -239,6 +238,27 @@ def add_planets_to_vicinity(vicinity, num_planets, gsd):
         num_planets -= 1
```
LGM-Doyle, line 241: get_system_within_jumps() has nothing to do with placing the positions in the galaxy.  I would rather not change it in this PR.  I expand on why I think it should be moved below.

default/python/universe_generation/galaxy.py
LGM-Doyle, line 13: AdjacencyGrid only requires MAX_STARLANE_LENGTH > MIN_SYSTEM_SEPARATION so the too_close_to_other_positions() and neighbors() mean different regions.  stitching_positions() below requires that MAX_STARLANE_LENGTH > 2*MIN_SYSTEM_SEPARATION in order to place positions(s) between the nearest neighbors in groups of positions that are too far apart. I reported the most restrictive bound in the PR request.

default/python/universe_generation/galaxy.py
LGM-Doyle: I'll change it to `remove_pos()`  It is self-documenting.

default/python/universe_generation/galaxy.py
LGM-Doyle, line 68: I am happy to add the docstrings for internal functions where the docstring clarifies something about usage.

However, adding docstrings for self documenting function/variables names like `remove_pos(pos)` where the docstring will only repeat the function/variable name merely adds visual noise, in the best case.  Typically, they end up being de-coupled and becoming erroneous and noisy.  

I'd like to add doctrings, everywhere that I have a comment clarifying usage, and not add any docstrings which do not expand on the function name.

default/python/universe_generation/universe_tables.py
LGM-Doyle: `stitching_positions()` only works if there is at least 2*MIN_SYSTEM_SEPARATION between the nearest stars in two groups, in order to generate a position for the extra star(s).  I put the comment here so that it is right at the location where someone will change the MAX_STARLANE_LENGTH.

default/python/universe_generation/galaxy.py
LGM-Doyle, line 66: `galaxy.py` is a misnomer.  It should more properly by named `galaxy_morphology.py`.  `get_systems_within_jumps()` does not belong in `galaxy.py`.

All of the functions/classes in `galaxy.py`, are directly or indirectly called by `calc_star_system_positions()` to generate position data.  Except `get_systems_within_jumps()`.

None of the functions know anything about starlanes or deal with starlanes.    Except `get_systems_within_jumps()`.

`get_systems_within_jumps()` doesn't belong in this file.  The only place that it is called is in `empires.py`.  There are similar functions in `empires.py` which also call `fo.` to access starlane information, such as `systems_min_jumps_from()`.  I moved the definition of `get_systems_within_jumps()` to just before it is used. 

I could:
1. Do nothing.  Leave `get_systems_within_jumps()` in `empires.py`.  I like this option.
2. Restore `get_systems_within_jumps()` to `galaxy.py`. 
3. Rename `galaxy.py` to `galaxy_morphology.py`.  Create `galaxy_topology.py` and move `get_systems_within_jumps()` into it.  

I'll do whichever option you prefer.

default/python/universe_generation/galaxy.py
Cjkjvfnby, line 68: Docstring and comment are two ways to document code. Comment is wrong way and should no be used.

default/python/universe_generation/universe_tables.py
Cjkjvfnby: Put this explanation to code. From my context it look like debug artefact, which may be possible issue.

LGM-Doyle:
I think that I have addressed all of the comments.

I restored `get_systems_within_jumps()` to `galaxy.py`, but I didn't make the changes suggested by @Cjkjvfnby .  I agree with the observations, but I am working on another PR to speed up universe generation and I will include the changes with that PR.

I snuck in two bits of new code to this PR update:
1. Changed list to set in the AdjacencyGrid.
2. Added combining code the clusterer.  It answers @Cjkjvfnby's question, "Why is Clusterer not a function?"  Because it can use its state information to efficiently combine the groups of systems without re-running the entire disjoint set algorithm each time.  I forgot to include it first time around.

default/python/universe_generation/galaxy.py
```diff
@@ -15,19 +15,23 @@ def __init__(self, universe_width):
         self.max_dist = universe_tables.MAX_STARLANE_LENGTH
         self.cell_size = min(max(universe_width / 50, self.min_dist), self.max_dist/sqrt(2))
         self.width = int(universe_width / self.cell_size) + 1
-        self.grid = defaultdict(list)
-        print "Adjacency Grid: width", self.width, ", cell size", self.cell_size
+        self.grid = defaultdict(set)
+        print "Adjacency Grid: width {}, cell size {}".format(self.width, self.cell_size)
 
     def cell(self, pos):
+        """Returns cell index"""
```
Cjkjvfnby: The docstring is a phrase ending in a period. https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring

default/python/universe_generation/galaxy.py
```diff
@@ -15,19 +15,23 @@ def __init__(self, universe_width):
...
         return (int(pos[0] / self.cell_size), int(pos[1] / self.cell_size))
 
     def insert_pos(self, pos):
-        self.grid[self.cell(pos)].append(pos)
+        """Inserts pos"""
+        self.grid[self.cell(pos)].add(pos)
```
Cjkjvfnby, line 27: Looks slightly wired. We have `pos` in key and in value.  If we always can recreate key by value when we don`t need dict here.    I did not dive deep into it, may be there is some reasons.

I assume that if you have getter methods for `self.grid` then you did not access it directly and can make it private by adding double underscore to its name `__`.  This really help during refactoring, because you can be sure that attribute is not used outside.

default/python/universe_generation/galaxy.py
```diff
@@ -205,16 +217,50 @@ def __init__(self, positions, adjacency_grid):
...
 
-    def clusters(self):
-        """Returns a list of lists of the clusters of positions"""
-        return self.dsets.complete_sets()
+    def smallest_isolated_cluster(self):
+        """
+        Return None if all of the positions are in one cluster
+        or the smallest cluster
+        """
+        return None if len(self.clusters) < 2 else min(self.clusters)[1]
```
Cjkjvfnby, line 242: Are you sure that in case of 3 clusters smallest are always second?

IMHO if you reverse conditions it will look more pretty

``` python
return min(self.clusters)[1] if len(self.clusters) > 1 else None
```

default/python/universe_generation/galaxy.py
```diff
@@ -205,16 +217,50 @@ def __init__(self, positions, adjacency_grid):
...
             # print "Clusterer adding ", pos,  " and ", neighbor
             self.dsets.link(pos, neighbor)
 
-    def clusters(self):
-        """Returns a list of lists of the clusters of positions"""
-        return self.dsets.complete_sets()
+    def smallest_isolated_cluster(self):
+        """
+        Return None if all of the positions are in one cluster
+        or the smallest cluster
```
Cjkjvfnby: Are you sure that in case of 3 clusters smallest are always second?

default/python/universe_generation/galaxy.py
```diff
@@ -205,16 +217,50 @@ def __init__(self, positions, adjacency_grid):
...
+        """
+        return None if len(self.clusters) < 2 else min(self.clusters)[1]
+
+    def stitch_clusters(self, p1, p2, stitches):
```
Cjkjvfnby, line 244: IMHO function name should match documentation string. `stitch` = `join`. I prefer to rename function to `join_clusters`.

default/python/universe_generation/galaxy.py
LGM-Doyle, line 242: The elements of self.clusters are tuples of `(len(cluster_a), cluster_a)`.  I was unaware of `min(x, key=func)`.  I will change this line of code to `min(self.clusters, key=len)` which is clearer.

The order is so that the desired and typical result `None` is first.

default/python/universe_generation/galaxy.py
LGM-Doyle, line 244: I will change the comment to use the phrase "stitch together" because it evokes the visual image of two clusters stitched with a path of systems, which is exactly how it looks with clusters more than one star apart.

default/python/universe_generation/galaxy.py
LGM-Doyle, line 27: The value is the set containing pos, not pos itself.  

The `AdjacencyGrid` provides coarse chunking of the universe so that you can compare positions of systems with other systems in their vicinity without searching the entire galaxy of systems.  So given the position you can always find the set containing that position and nearby systems.

default/python/universe_generation/galaxy.py
genestack-solomatin, line 242: Sorry I missed `min`  

`min(self.clusters, key=len)` will sort by length of tuple.  So you need to change `self.clusters` structure too.

geoffthemedio:
Are there any remaining Python-style concerns about this?

Vezzra:
> Are there any remaining Python-style concerns about this?

Well, as I said:

> Just did a first very quick pass, made a few comments. More thorough passes to come.

I admit I've been dragging my feet about this, because it's a lot to go over. There definitely are some issues/concerns I want to address, so this is not yet ready to merge.

geoffthemedio:
bump / reminder

Vezzra:
Thanks for nagging me about that... will _try_ to get to it this week...

default/python/universe_generation/galaxy.py
Vezzra, line 13: Ok, I see.

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,349 @@
...
+        self.clusters.add((len(new_set), new_set))
+
+
+def stitching_positions(p1, p2):
+    """
+    Returns a list of positions between p1 and p2 between MIN_SYSTEM_SEPARATION and MAX_STARLANE_LENGTH apart
+    """
+    assert 2 * universe_tables.MIN_SYSTEM_SEPARATION < universe_tables.MAX_STARLANE_LENGTH,\
+        "MAX_STARLANE_LENGTH must be twice MIN_SYSTEM_SEPARATION to allow extra positions to \
+        be added to enforce MAX_STARLANE_LENGTH"
```
Vezzra: `MAX_STARLANE_LENGTH` and `MIN_SYSTEM_SEPARATION` are supposed to be parameters that can be customized by players, so failing here with an assertion isn't a good idea. General policy is to abort execution only if it can't be avoided at all, otherwise log an error and try to continue.

In case of the other assertion check above (`AdjacencyGrid` init) failing hard probably can't be avoided, because if that condition isn't met, star system positioning is screwed. But here you can try to continue, just abort the process of stitching together the disjointed clusters. Log an error using the `report_error` function of the `util` module, which will ensure the player gets a proper warning at the start of the game.

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,349 @@
...
+        between ''p1'' and ''p2''
+        """
+        len_dset1 = None
+        len_dset2 = None
+        for len_dset in self.clusters:
+            if p1 in len_dset[1]:
+                len_dset1 = len_dset
+            if p2 in len_dset[1]:
+                len_dset2 = len_dset
+        assert len_dset1 and len_dset2, "p1 and p2 must be points in disjoint sets of positions"
```
Vezzra: See my other comment a few lines down. No reason to fail hard here, just log an error and abort the insertion of additional systems to enforce the max starlane length limit. This will result in getting a map with overly long starlanes, but the player can still play. If you fail with an assertion here, the entire universe generation process will be aborted, resulting in an unplayable map.

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,349 @@
...
+                     radius_0space * p1_p2_scale * sin(angle_0space + p1_p2_theta))
+
+    p3 = (rotated_point[0] + p1_p2_translation[0], rotated_point[1] + p1_p2_translation[1])
+
+    return [p3] + stitching_positions(p1, p3) + stitching_positions(p2, p3)
+
+
+def enforce_max_distance(positions, adjacency_grid):
+    """
+    Adds extra positions between groups of positions to guarrantee
```
Vezzra: guarantee

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,349 @@
...
+        ensure every position has at least one neighbor within
+        MAX_STARLANE_LENGTH.
+    """
+    # Find all clusters
+    clusterer = Clusterer(positions, adjacency_grid)
+
+    if len(clusterer) == 1:
+        print "All systems positioned in a single connected cluster"
+    else:
+        print "{} clusters separated by more the MAX_STARLANE_LENGTH.  Starting to fill gaps.".format(len(clusterer))
```
Vezzra: "...by more than the..."

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,349 @@
...
+                           for nn in adjacency_grid.nearest_neighbor(pos)
+                           for pos in smallest_cluster])
+
+        for pos in smallest_cluster:
+            adjacency_grid.insert_pos(pos)
+        # Add extra positions
+        extra_positions = stitching_positions(p1, p2)
+        for i_extra, p3 in enumerate(extra_positions):
+            print "Adding {} of {} extra positions at {} to enforce max spacing between {} and {}."\
+                .format(i_extra, len(extra_positions), p3, p1, p2)
```
Vezzra: `i_extra + 1`, otherwise you get log messages like this:

```
Adding 0 of 1 extra positions at (1329.2735707173076, 599.3023108153775) to enforce max spacing between (1311.4786665332913, 636.6901601665741) and (1353.0400945377253, 562.9319676906877).
```

I assume this should read "Adding 1 of 1 extra positions..."

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,349 @@
...
 
 
-class AdjacencyGrid:
+class AdjacencyGrid(object):
     def __init__(self, universe_width):
+        assert universe_tables.MIN_SYSTEM_SEPARATION < universe_tables.MAX_STARLANE_LENGTH
         self.min_dist = universe_tables.MIN_SYSTEM_SEPARATION
-        self.cell_size = max(universe_width / 50, self.min_dist)
+        self.max_dist = universe_tables.MAX_STARLANE_LENGTH
+        self.cell_size = min(max(universe_width / 50, self.min_dist), self.max_dist/sqrt(2))
```
Vezzra: spaces around the `/` in `self.max_dist/sqrt(2)`

default/python/universe_generation/galaxy.py
```diff
@@ -1,35 +1,349 @@
...
+    while len(clusterer) > 1:
+        smallest_cluster = clusterer.smallest_isolated_cluster()
+        print "Searching for nearest neighbor position to a cluster with {} positions.".format(len(smallest_cluster))
+        for pos in smallest_cluster:
+            adjacency_grid.remove_pos(pos)
+        # Find nearest neighbour
+        (_, p1, p2) = min([(util.distance(pos, nn), pos, nn)
+                           for nn in adjacency_grid.nearest_neighbor(pos)
+                           for pos in smallest_cluster])
+
```
Vezzra, line 349: I assume getting the return value of `adjacency_grid.nearest_neighbor` as a list with only one element instead of returning its result just as a single value allows you to pack searching for the nearest neighbor in a compact and efficient statement like the above, but making `adjacency_grid.nearest_neighbor` return its result as a list is somewhat clunky/counter-intuitive.

Wouldn't that also work if making `adjacency_grid.nearest_neighbor` return just a single value and writing the above statement like this:

```
        (_, p1, p2) = min([(util.distance(pos, nn), pos, nn)
                           for nn in [adjacency_grid.nearest_neighbor(pos)]
                           for pos in smallest_cluster])
```

...?

default/python/universe_generation/galaxy.py
```diff
@@ -109,15 +422,13 @@ def spiral_galaxy_calc_positions(positions, arms, size, width):
             break
 
         if not attempts:
-            print "Spiral galaxy shape: giving up on placing star", i,\
-                  ", can't find position sufficiently far from other systems"
+            print "Spiral galaxy shape: giving up on placing star {}, can't find position sufficiently far from other systems.".format(i)
```
Vezzra: Changing the print statement to use the new recommended formatting is fine, but why did you remove the line break? We're trying to conform to PEP8, and this results in a slightly too long line (and the notorious PEP8 warning).

Same for all the other cases where you did this.

default/python/universe_generation/galaxy.py
```diff
@@ -440,45 +746,41 @@ def calc_star_system_positions(gsd):
...
     elif gsd.shape == fo.galaxyShape.cluster:
-        # Typically a galaxy with 100 systems should have ~5 clusters
-        avg_clusters = gsd.size / 20
-        if avg_clusters < 2:
-            avg_clusters = 2
-        # Add a bit of random variation (+/- 20%)
-        clusters = randint((avg_clusters * 8) / 10, (avg_clusters * 12) / 10)
-        if clusters >= 2:
-            cluster_galaxy_calc_positions(positions, clusters, gsd.size, width)
+        cluster_galaxy_calc_positions(positions, adjacency_grid, gsd.size, width)
```
Vezzra, line 805: Moving that piece of code into `cluster_galaxy_calc_positions` is probably a good idea. If anything, there should be a parameter specifying the numbers of stars per cluster, instead of the numbers of clusters.

default/python/universe_generation/galaxy.py
Vezzra, line 68: As @Cjkjvfnby said, this is important for documentation. We are working on a process to create reference documentation for all the FO Python code automatically, which retrieves the docstrings in the code. So even the simplest of functions should provide them.

Well, moot point, as you added those docstrings already. Thanks for bearing with our nitpicking :wink:

default/python/universe_generation/galaxy.py
Vezzra, line 66: Ok, now I get where you're coming from. And considering that `galaxy.py` only contains star system placement code besides `get_systems_within_jumps` your conclusion is understandable, however, the `galaxy.py` module is intended to contain _all_ the galaxy map related stuff, hence the name. Which coincidentally is just star system placement and the lonely `get_systems_within_jumps` function.

When I wrote that code I was actually aware of the slight ackwardness of that combination, and considered splitting up things differently. But `get_systems_within_jumps` doesn't belong into `empires.py`, this module contains empire setup stuff, not galaxy map stuff. And creating a module just for that single function seemed overkill.

So I ended up putting these together into a module that's supposed to contain the map related stuff. I'm open to suggestions for a better name though, if you think `galaxy.py` is misleading.

Vezzra:
Ok, _finally_ got around to do a more thorough review of this. Had a few more comments/corrections/nitpicks, otherwise looks fine. You're certainly a better coder than me :wink:

Also did some brief testing, seems to work smoothly.

One final thing: while it's fine to sneak in some minor changes that aren't strictly related to the purpose of the patch, all the timer stuff you added in here really should go in a separate PR. It's good stuff, because it provides valuable debug information, but it doesn't really belong here. I know why you included it, but before we merge you should split that off.

And because of my extreme tardiness it's probably a good idea to rebase the whole thing, even if there are no conflicts so far.

Vezzra:
[EDIT] Just saw that this has been a double post. Dunno how that happened, but deleted the contents of the duplicate.

Vezzra:
There is another thing about the universe generation Python scripts I want to point out here, concerning structure and coding style. Instead of repeating a very long forum post I made a long time ago, I put a link in here: http://freeorion.org/forum/viewtopic.php?p=67524#p67524

I don't know if what I tried to achive is possible, but that post sums up an important part of my original intentions with that whole Python scripted universe generation thing. Just wanted to bring that to your attention, because I guess you might do more in that department in the future.

And yes, I know, that should probably go onto a wiki page, instead of being buried in an ancient forum thread...

default/python/universe_generation/galaxy.py
LGM-Doyle: nemesis word

default/python/universe_generation/galaxy.py
LGM-Doyle, line 349: You are correct.

However, an empty adjacency_grid yields no nearest neighbor.

I have fixed it to return None for no neighbor.

default/python/universe_generation/galaxy.py
LGM-Doyle: The print statement output was wider than 79 characters.  It seemed wrong to break them, when they would not be broken in the log output.

Fixed in any case.

default/python/universe_generation/galaxy.py
LGM-Doyle: I agree with this change.  

A user configuration error should not cause the whole universe generation to fail.

default/python/universe_generation/galaxy.py
LGM-Doyle: I made this change but I disagree with it for the following reason.

This is different from the other case in that it is a programmer error not a configuration error.

This assertion failing means that either Clusterer is designed incorrectly or it is being used incorrectly by the programmer currently using it.

If the failure is a soft failure then broken code either the Clusterer itself or code using the Clusterer will get checked in and perhaps merged into master.  

In the future some other programmer, not the one currently breaking the build will have to fix the problem.  That programmer will have to infer what the current programmer's intent is from the commit log.   

The fix I provided is a poor fix, because it will result in the loop terminating without informing the surrounding code that it failed.

default/python/universe_generation/galaxy.py
LGM-Doyle, line 805: Currently it specifies approximately 20 stars per cluster.

I'd like to change that so that the number of stars per cluster increases gradually as the galaxy size increases.  

I think that aesthetically it would reveal that the clusters are rotated ellipses as the galaxy is explored on larger maps.  It would be like the shape of the spirals is revealed as you explore the spiral galaxies.

I like that.  I often play spiral galaxies for that reason.

Secondly, it would make cluster map play out differently on small galaxies versus large galaxies.  

On small galaxies it would play the same as it does now.  Each player tries to grab a cluster, isolate them selves and then explode onto the galaxy at large.

On large galaxies it would play more like a series of elliptical galaxies.  A struggle in the early game to master a cluster with no good choke points.  Followed by a late game with big guns, big empires and an isolationist option.

LGM-Doyle:
Thanks a lot for the review.  I know that you are pressed for time.

All changes are done and it is rebased.

As mentioned above I continued working on improving startup while I thought this
PR review was imminent.

I now plan on submitting a PR of startup improvements after 0.4.6 is finalized.

The changes are in both the python code and the C++ code.

On my local version startup of a new game with
999 systems and 39 AIs takes 21 seconds versus 111 seconds
for master.  Smaller universes are also faster, but I needed a consistent test case.

This is not an apples to apples change.  The distribution of specials and
perhaps monsters will be
slightly different and may require an adjustment of the numbers in
SPECIALS_FREQUENCY and MONSTER_FREQUENCY that I haven't done.

Please decide if I should start cleaning up those commits into a
PR for submission.

LGM-Doyle:
tl:dr Assorted comments on your code style post. Doesn't need to be read or responded to.

Thank you for the compliment.

Coding is an entirely learned skill.  Everything I know I've learned from reading texts or code.

I've read through your post and here are some comments.

Don't take anything that I say here as a defense of my coding style as an
example of clarity and easily understood code.

Encouraging everyone to hack on everything is a laudable goal.

I've found that more than anything else, novices want an assurance that nothing
they can do will break anything and that they can always get back to where they
started.  
- Readability

I think readability of code to an inexperienced coder depends on how closely its
paradigm matches the code that they learned on.  And that depends in part on
when they learned to code.
- 80s Procedural with global variables
- 90s Object oriented
- 00s Functional
- all time Declarative

Chasing readability for novices with a small pool of novices commenting risks
an echo chamber effect.  Novices who found the code difficult to understand
won't comment.  Novices who do comment favorably are probably saying the code
looks like code they have read before.

There is a difference between simple and simplistic.

Sometimes it is a matter of taste.  However, at some point each line of code is
so simple that anyone can understand it, and the intent of code is lost in the
details.  The code becomes so verbose that no one knows, by inspection where a
change needs to be made.

Sometimes it goes beyond taste to performance.

One way to defend against this is to pull any code that risks worst than O(N)
performance back across the C++ boundary in C++ land.  This solution makes the
unwarranted assumption that the C++ code is less likely to contain algorithmic
defects. 

For example the disjoint set code in this PR should probably be re-written
using boost::graph code. When I wrote this I didn't realize that freeorion
already used boost::graph. 

Using boost::graph is a better solution because that disjoint set algorithm is
tested.
- Encapsulation and Modularization

I agree with what you say in these sections.

I think that clarity is provided in clever naming and encapsulation.  If each
level is short, well named then it is either easy to understand how it works or
easy to understand what it does and ignore how it works.

In python and C++ files and folders are free.

If you need to describe a folder/file/class/function with it does this, that
and the other thing, perhaps it should be 3 folders/files/classes/functions
that do this, that and the other thing. 

I've looked at the universe generation code and the way that it is broken up
makes sense.  It strikes terror into my heart to imagine that is was one
monstrous file.

A key to encapsulation is a well defined interface. It should be clear how
information enters and exits the object.  It is easier to reason about the
correctness of an algorithm if it is clear where all the inputs and when they
can change.

An important aspect of this is no global state.  That means no global
variables, no singletons etc.  State that can be changed outside of lexical
scope makes it difficult or impossible to be certain an algorithm is correct.

Global state is an easy, insidious way to break otherwise good encapsulation.

default/python/universe_generation/galaxy.py
Vezzra: I understand what you mean, and you certainly have a point here. However, it has been project policy for as long as I can remember that as long as there is any possibility to continue execution we _never_ fail hard (which is why there are so few instances where exceptions are thrown, and even where that happens, the majority gets caught and tried to handle at a higher level, at least AFAIK). Regardless if we're dealing with potential user config/input or programmer errors.

The reasoning behind that is that the game shouldn't crash on the player unless it's absolutely impossible to avoid. Even if game execution is impaired. Of course, provided these errors get logged and the logs carefully checked (especially before committing/merging things into master), to avoid the problems you mentioned.

As you can probably guess, not everyone was/is in agreement with that policy, and it has been repeatedly disputed. The reasons you cite above usually being the main argument against it.

However, that hasn't led to a change to this policy, so it's still in force. That said, I don't think that it can't be challenged again. As I said above, I do see your point, and I'm not adamant about it. But I'm not the one you need to convince, that would be @geoffthemedio, our project and programming lead. If you want to put that topic up for discussion, I suggest doing so on the forum (probably with a link to this little exchange here).

As far as our specific case here is concerned however, I still tend to prefer soft failure, because this is a part of the universe generation process that isn't vitally important to be able to carry on. If we fail here (regardless if of an user config or programmer error), all that happens is that the max starlane length doesn't get enforced, and we might end up with a map that has some overly long starlanes. That shouldn't cause universe generation to fail entirely.

That said, I expect the risk of that assertion to ever fail very low, so we can change that at a later point, no need to hold up merging the PR because of it.

default/python/universe_generation/galaxy.py
Vezzra: AFAIK the PEP8 max line length isn't 79/80, but 120. So that would be why the original statement was fine, while after your change it would have been just a bit too long. I'm not 100% sure about that however.

default/python/universe_generation/galaxy.py
Vezzra, line 805: Sounds good 😃

Vezzra:
Thx, excellent work. Finally merged 😃 

> Please decide if I should start cleaning up those commits into a
> PR for submission.

The improvements are tempting, but I expect the required fine-tuning of the specials and monster frequency universe tables to be non-trivial. With @Morlic-fo finally back and getting to those crtitical AI updates we have now a chance to really make some headway with the release, so I'm reluctant to throw that in. @MatGB and the other content scripters are busy enough sorting stuff out.

So I tend to postpone that after the release. But then this work you've done there should definitely go in.

> Assorted comments on your code style post. Doesn't need to be read or responded to.

Well, I think your comments are valuable input, so if I can squeeze it in at some point I'd like to come back to this. But that discussion actually belongs on the forum, so if/once I get back to it, I'll continue our discourse there 😃

default/python/universe_generation/galaxy.py
dbenage-cx: [72/79](https://www.python.org/dev/peps/pep-0008/#maximum-line-length), optionally expanded to 72/99. It was mentioned to me [previously](https://github.com/freeorion/freeorion/pull/567#issuecomment-202140779) to expand to 120.  The wiki entry could use a mention of this if 120 should be the guideline.

Issue #652, opened 2016-05-09
Hidden child windows intermittently obstruct interaction


dbenage-cx:
When a child window is hidden, the area under it is not always completely intractable until the window is fully brought into focus.

For me, this is mostly apparent when using a large 'pedia window in the research screen (and this is ~90% reproducible here):
- Open research screen
- Resize 'Pedia to fill a majority of the area.
- Hide the 'Pedia window
- Hover over a tech that would have been under the 'Pedia window (tooltip shows)
- Attempt to double click on the same tech (fails to add to queue, hover text no longer shows)
- Click on an area outside of the 'Pedia window area.
- Double click on same tech (succeeds) 

Production screen is not reliably reproducible (there is less interaction with map view), the common occurrences appear to happen on the first instance of viewing the production screen.

geoffthemedio:
I question whether your claimed cause is always / entirely true. The production wnd has some special-case code in its InWindow functions, which call other Wnds' InWindow functions directly...

Regardless, a more-general fix involving adding texts of visibility before allowing a click to be considered in a window might be better... Moving child winds around the Z-order could potentially lead to rendering problems later.

dbenage-cx:
During expansion of my description, I have mistakenly left out that this is an _intermittent_ issue. 
Adjusted description and title accordingly.

PR #653, opened 2016-05-09
Adjust z-order when toggling Pedia in Research and Production screens

dbenage-cx:
Workaround for #652

Specifically reorders for research and production screens.


dbenage-cx:
Closing as undesired behavior

PR #657, opened 2016-05-10
Reject pointer selection for hidden windows

### LINKED ISSUE #652: Hidden child windows intermittently obstruct interaction

dbenage-cx:
Fixes #652

### DIFF
--- a/GG/src/ZList.cpp
+++ b/GG/src/ZList.cpp
@@ -191,6 +191,8 @@ Wnd* ZList::PickWithinWindow(const Pt& pt, Wnd* wnd, const std::set<Wnd*>* ignor
     // any of them (or their children)
     std::list<Wnd*>::reverse_iterator end_it = wnd->m_children.rend();
     for (std::list<Wnd*>::reverse_iterator it = wnd->m_children.rbegin(); it != end_it; ++it) {
+        if (!(*it)->Visible())
+            continue;
         Wnd* temp = 0;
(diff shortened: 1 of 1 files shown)


dbenage-cx:
Minimum change for the issue.  If you'd rather, this might be handled in GG::Wnd and then re-checked in any overriden InWindow.

geoffthemedio:
A modification of this sort just for CUIWnd seems rather odd and prone to causing unexpected behaviour. I'm not sure what the base GG code is assuming InWindow function will do. I think a better fix would be to check the visibility in places where a child Wnd's InWindow is directly called, and not bother checking InWindow in cases where a non-visible Wnd shouldn't be checked.

dbenage-cx:
Ty Geoff

geoffthemedio:
looks reasonable and simple enough, though probably has unexpected consequence, but I suppose we'll just have to figure out what those are...

Issue #660, opened 2016-05-12
Save files become inaccessible if renamed


Cjkjvfnby:
Steps to reproduce:
- start game
- rename saves form `*.sav` to `xxx` or anything without .`sav` at the end
- try to load game

Observed result:
- there is no save game with this name
- file present in saves directory

Expected result:
- if `.sav` is absent it should be added before saving

```
Windows 7x64
v0.4.5+ [build 2016-05-10.2a6b7b4] MSVC 2013
```

dbenage-cx:
PR addresses correctly appending the extension when saving

> rename saves form *.sav to xxx or anything without .sav at the end

Assuming these are renamed outside of the game, the observed result is expected.

Vezzra:
Ok, thanks. Closing this issue.

PR #661, opened 2016-05-12
Append appropriate file extension when saving

dbenage-cx:
Addresses the expected result portion of #660

### DIFF
--- a/UI/SaveFileDialog.cpp
+++ b/UI/SaveFileDialog.cpp
@@ -605,6 +605,12 @@ void SaveFileDialog::Init() {
     GG::Button* cancel_btn = new CUIButton(UserString("CANCEL"));
 
     m_name_edit = new CUIEdit("");
+    if (m_extension != MP_SAVE_FILE_EXTENSION && m_extension != SP_SAVE_FILE_EXTENSION) {
+        std::string savefile_ext = HumanClientApp::GetApp()->SinglePlayerGame() ? SP_SAVE_FILE_EXTENSION : MP_SAVE_FILE_EXTENSION;
+        DebugLogger() << "SaveFileDialog passed invalid extension " << m_extension << ", changing to " << savefile_ext;
+        m_extension = savefile_ext;
+    }
+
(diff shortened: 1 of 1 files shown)


Cjkjvfnby:
@geoffthemedio, can you take care about this PR?

geoffthemedio:
This doesn't quite work as intended. If there is a file "test", then it will pop up a warning about overwriting, even though the file that will be written is "test.sav". If there is a "test.sav", and the user types just "test", it won't pop up a warning, but will overwrite the "test.sav" file.

dbenage-cx:
Moved to CheckChoiceValidity(), appending the extension as the user types (saving only).

geoffthemedio:
Now when I put the cursor at the end of a filename and press backspace, or otherwise type anything without .sav at the end, it appends it very awkwardly.
extension

dbenage-cx:
Was commit 2b5d9ca accidental?  I can rebase either way but wanted to check before scrubbing a commit.

New commit behavior disables the OK button if the extension is not valid.  To assist quickly changing the name, only the non-extension part of the filename is initially selected.

If that is not preferred, should we just hide the extension from the user?

geoffthemedio:
yes, accidental. switching between branches apparently didn't remove the old branch modifications before applying the new branch stuff, and then after an unrelated commit I pushed.

geoffthemedio:
Disabling the OK button without any indication why is not useful.
It should just treat any save file name without ".sav" at the end as if it had ".sav" at the end when the user clicks OK. That includes popping up a warning that "test.sav" already exists if the user types just "test".

dbenage-cx:
Rebased.  Of note, multiplayer does not confirm overwrite.  This is the same behavior as master so I did not want to muddy the waters further if this is a separate issue.

Vezzra:
> Addresses the expected result portion of #660

Is #660 completely fixed by this (and can be closed), or are further fixes required to address #660?

dbenage-cx:
@Vezzra I've commented on issue, remainder is expected behavior.

PR #662, opened 2016-05-13
Add Environment pedia article

Ouaz:
See http://www.freeorion.org/forum/viewtopic.php?p=83931#p83931

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -4071,6 +4071,20 @@ TROOP_TEXT
 
 The Troop meter on a planet represents the power of the defending ground forces. The invading force must send a more powerful force (i.e. a higher cumulative number) to conquer a planet.  Ships must be built with special parts for carrying troops, and such troops cannot invade if the defending planet has active [[encyclopedia SHIELDS_TITLE]].'''
 
+ENVIRONMENT_TITLE
+Environment
+
+ENVIRONMENT_TEXT
+'''Eleven different environments can be encountered when exploring the planetary systems on the galactic map: Inferno, Radiated, Toxic, Barren, Desert, Tundra, Swamp, Terran, Ocean, Asteroids, and Gas Giant.
+
+Each [[encyclopedia ENC_SPECIES]] has its own environmental preferences, which determine on each planet type the suitability for the species, according to the following list (from worst to best):
+
+Uninhabitable < Hostile < Poor < Adequate < Good
+
+The planet suitability report can be displayed by right-cliking on a planet in the system sidepanel. A green number following the suitability data indicates the max population bonus value if the species colonizes the planet, whereas a red number indicates that the population will stay idle or will decrease, to finally perish if the species attempts to colonize the planet.
+
+When terraforming a planet (by research or if the planet has the [[encyclopedia GAIA_SPECIAL]] special) in order to better suit the species' environmental preferences, the planet's original environment is modified by stages, until it finally reaches the Good suitability for the species who wants to live on the planet. The terraforming stages can be checked on the suitability wheel displayed in the Planet Suitability report.'''
+
 ORGANIC_SPECIES_TITLE
 Organic Metabolism
 ORGANIC_SPECIES_TEXT
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -4075,15 +4075,15 @@ ENVIRONMENT_TITLE
 Environment
 
 ENVIRONMENT_TEXT
-'''Eleven different environments can be encountered when exploring the planetary systems on the galactic map: Inferno, Radiated, Toxic, Barren, Desert, Tundra, Swamp, Terran, Ocean, Asteroids, and Gas Giant.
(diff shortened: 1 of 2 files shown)


default/stringtables/en.txt
```diff
@@ -4071,6 +4071,20 @@ TROOP_TEXT
 
 The Troop meter on a planet represents the power of the defending ground forces. The invading force must send a more powerful force (i.e. a higher cumulative number) to conquer a planet.  Ships must be built with special parts for carrying troops, and such troops cannot invade if the defending planet has active [[encyclopedia SHIELDS_TITLE]].'''
 
+ENVIRONMENT_TITLE
+Environment
+
+ENVIRONMENT_TEXT
+'''Eleven different environments can be encountered when exploring the planetary systems on the galactic map: Inferno, Radiated, Toxic, Barren, Desert, Tundra, Swamp, Terran, Ocean, Asteroids, and Gas Giant.
```
Cjkjvfnby: May be avoid to mention exact number of planet types, jsut to change less if new one will come?

Should we use links to planet types names? `[[PT_SWAMP]]`?

default/stringtables/en.txt
```diff
@@ -4071,6 +4071,20 @@ TROOP_TEXT
...
+
+Each [[encyclopedia ENC_SPECIES]] has its own environmental preferences, which determine on each planet type the suitability for the species, according to the following list (from worst to best):
+
+Uninhabitable < Hostile < Poor < Adequate < Good
+
+The planet suitability report can be displayed by right-cliking on a planet in the system sidepanel. A green number following the suitability data indicates the max population bonus value if the species colonizes the planet, whereas a red number indicates that the population will stay idle or will decrease, to finally perish if the species attempts to colonize the planet.
+
+When terraforming a planet (by research or if the planet has the [[encyclopedia GAIA_SPECIAL]] special) in order to better suit the species' environmental preferences, the planet's original environment is modified by stages, until it finally reaches the Good suitability for the species who wants to live on the planet. The terraforming stages can be checked on the suitability wheel displayed in the Planet Suitability report.'''
```
Cjkjvfnby: `the species' environmental`  single quote mark

default/stringtables/en.txt
```diff
@@ -4071,6 +4071,20 @@ TROOP_TEXT
...
 
+ENVIRONMENT_TITLE
+Environment
+
+ENVIRONMENT_TEXT
+'''Eleven different environments can be encountered when exploring the planetary systems on the galactic map: Inferno, Radiated, Toxic, Barren, Desert, Tundra, Swamp, Terran, Ocean, Asteroids, and Gas Giant.
+
+Each [[encyclopedia ENC_SPECIES]] has its own environmental preferences, which determine on each planet type the suitability for the species, according to the following list (from worst to best):
+
+Uninhabitable < Hostile < Poor < Adequate < Good
```
Cjkjvfnby: Use `# PlanetEnvironments` keys?

default/stringtables/en.txt
MatGB: > May be avoid to mention exact number of planet types, jsut to change less if new one will come?

Normally, I'd agree completely, but I don't see any plans for new environments and, unless there's a _really_ good reason, would argue for there to be a Special instead, we don't have a strong enough creature mix for what we've got.  

I would suggest splitting it into 9 standard habitable planets and two special types that are normally not colonised but have different in game effects.

I agree on the keys though.

Ouaz:
Modifications done.

Sorry for the number of commits in this PR. I'm not used yet to Git, so I still work from the Github website.

MatGB:
That's fine, we're all learning things about, it's both very powerful and very confusing. Anyway, done, thanks.

default/stringtables/en.txt
```diff
@@ -4071,6 +4071,20 @@ TROOP_TEXT
...
+
+Each [[encyclopedia ENC_SPECIES]] has its own environmental preferences, which determine on each planet type the suitability for the species, according to the following list (from worst to best):
+
+[[PE_UNINHABITABLE]] < [[PE_HOSTILE]] < [[PE_POOR]] < [[PE_ADEQUATE]] < [[PE_GOOD]]
+
+The planet suitability report can be displayed by right-cliking on a planet in the system sidepanel. A green number following the suitability data indicates the max population bonus value if the species colonizes the planet, whereas a red number indicates that the population will stay idle or will decrease, to finally perish if the species attempts to colonize the planet.
+
+When terraforming a planet (by research or if the planet has the [[encyclopedia GAIA_SPECIAL]] special) in order to better suit the species environmental preferences, the planet's original environment is modified by stages, until it finally reaches the Good suitability for the species who wants to live on the planet. The terraforming stages can be checked on the suitability wheel displayed in the Planet Suitability report.'''
```
Cjkjvfnby, line 4086: `Good` - > `best`

Cjkjvfnby: Nevermind

Ouaz, line 4086: This "Good" = [[PE_GOOD]].

This paragraph is more or less a reminder of the GAIA_SPECIAL_DESC

> If the world is a Good environment a substantial population bonus is added, if it is not then it will slowly terraform itself in stages to eventually reach Good (NB: [[species SP_EXOBOT]] do not have a Good environment so it cannot improve beyond Adequate for them).

MatGB, line 4086: I'm thinking specifically referencing Exobots every time they've an exclusion is going to get annoying and forgotten, better to have _their_ article give all their quirks, and have them referenced only where we know confusion happens.

Cjkjvfnby, line 4086: It is related to `until it finally reaches the Good suitability`  Both `[[PE_GOOD]]` and  `best` suits here.

PR #667, opened 2016-05-15
Use correct stringtables entry for aggression

Cjkjvfnby:
- use correct stringtables entries for aggression
- print aggression name to log

This PR should be merged via `squash and rebase`


Vezzra:
Hm, shouldn't this also be flagged as a bugfix? After all, using wrong stringtable keys strikes me as a bug, and correcting that as a bugfix...?

PR #668, opened 2016-05-15
Fix KeyError bug with possible_output in ResourcesAI.py

LGM-Doyle:
The reporting function is able to reference undefined resource keys in possible_output.  I changed the references to .get().

### DIFF
--- a/default/AI/ResourcesAI.py
+++ b/default/AI/ResourcesAI.py
@@ -191,11 +191,11 @@ class Reporter(object):
             current_industry_target += fPP
             current_research_target += fRP
 
-            iPP, iRP = pinfo.possible_output[IFocus] if IFocus in planet.availableFoci else (fPP, fRP)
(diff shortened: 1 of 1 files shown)


Cjkjvfnby:
May be find real reason and fix?  `.get` will sillently hide this issue

LGM-Doyle:
The real reason is that it is trying to report on values that have not been set, as stated.

I could use defaultdict and set the default to (0,0).  That is probably the better solution.

Thoughts?

LGM-Doyle:
@Cjkjvfnby, I apologize.  

You are entirely correct, my previous PR was hasty because I was concerned about breaking master and people's games.  

This PR fixes the problem.

In detail, the problem was that planets that have their focus set by the growth, production or research specials, then do not have both industry and research foci set in `calculate_planet_infos()` and hence can't be included in the report of min/max research/industry setting.  

The solution is to only report on the available results.

Cjkjvfnby:
@LGM-Doyle, no need to apologize.  It is normal working process.

Issue #606, opened 2016-04-14
Import AI code only after freeOrionAIInterface is ready


Cjkjvfnby:
We initialize python in 2 stages:

```
- import code (which execute some code like config creation)
- prepare AI in c++ code and call `initFreeOrionAI`
```

Can we prepare AI first and only after that import python libs? 
This will make `initFreeOrionAI` obsolete and make things less complicated.

Vezzra:
Thx!

Cjkjvfnby:
ping

Vezzra:
Ok, I've squeezed in a quick look at the initialization code sections. I've to admit that I'm not sure what should be done differently - what exactly is this "prepare AI in C++ code" you're referring to, that should happen before the import of the `FreeOrion.py` module?

Because there isn't really happening much in terms of "preparing the AI" after the import of `FreeOrion.py` which could be done before that import. The import is actually one of the last things happening after everything else has been initialized. The only thing that does some kind of AI preparation after `FreeOrion.py` has been imported is the call to `initFreeOrionAI` - and that call of course requires the `FreeOrion.py` to be imported, otherwise it can't be called (as it's defined in `FreeOrion.py`).

So, I'm a bit confused about what exactly you want me to do...?

Cjkjvfnby:
If you put `print fo.empireID()` to module code or `initFreeOrionAI` you will get `-1`, in generateOrders you will get real number.

I was wrong assuming that this value is already set when `initFreeOrionAI` called.

I want `empireId` and other constants were accessible at the moment of import.

Vezzra:
Ah, now I understand what you are getting at.

> I want empireId and other constants were accessible at the moment of import.

I'm sorry, but that is not possible. The way the AI client is designed and implemented, the initialization and import of the `FreeOrionAI.py` module has to happen when the client _application_ is initialized, _before_ the game is started/loaded. At that stage no game specific data (like empire id and the other constants you're referring to) is available, simply because game initialization happens in a separate step _after_ the client app has been initialized. That can't be done differently, it's not possible to start/load the game before the Python AI has been initialized.

The purpose of `initFreeOrionAI` is to provide a function where AI scripters can run optional game independent initialization code. Any initialization code that requires access to game specific data has to happen in `startNewGame` (called in case a new game is being started) and `resumeLoadedGame` (called in case a saved game is being loaded).

Cjkjvfnby:
Got you. Lets close this issue.

PR #669, opened 2016-05-17
Add python logger support to AI

LGM-Doyle:
This adds support for the standard python logging.logger, by redirecting its output to an appropriate stream of the freeorion log files.

There are comments in the code base that print statements need to be removed or turned off for the release build.  The python logger supports debug levels and hierarchically named loggers.  Both of these features will be useful with detailed control of logging.  It doesn't help if the cost is in computing the logged data.

While refactoring to add this feature I removed some previously hard-coded python from CommonFramework.cpp.

I also moved initFreeOrionAI() into FreeOrion.py which might address issue #606 :"Import AI code only after freeOrionAIInterface is ready"

### DIFF
--- a/default/python/AI/AIDependencies.py
+++ b/default/python/AI/AIDependencies.py
@@ -20,6 +20,8 @@ Example usage:
     my_industry = AIDependencies.INDUSTRY_PER_POP * my_population
 """
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client  # pylint: disable=import-error
+from common.configure_logging import convenience_function_references_for_logger
+(debug, info, warn, error, fatal) = convenience_function_references_for_logger(__name__)
 
 # Note re common dictionary lookup structure, "PlanetSize-Dependent-Lookup":
 # Many dictionaries herein (a prime example being the building_supply dictionary) have a primary key (such as
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -5,7 +5,6 @@ import sys
 import random
 
 from common import configure_logging
-configure_logging.redirect_logging_to_freeorion_logger(__name__)
 
 import logging
 logger = logging.getLogger(__name__)
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -49,8 +49,8 @@ except ImportError:
 
 
 user_dir = fo.getUserDataDir()
-print "Path to folder for user specific data: %s" % user_dir
-print 'Python paths', sys.path
+logger.debug("Path to folder for user specific data: %s" % user_dir)
+logger.debug('Python paths %s' % sys.path)
 
 
 # Mock to have proper inspection and autocomplete for this variable
@@ -74,29 +74,29 @@ def startNewGame(aggression_input=fo.aggression.aggressive):  # pylint: disable=
         return
 
     if empire.eliminated:
-        print "This empire has been eliminated. Ignoring new game start message."
+        logger.info("This empire has been eliminated. Ignoring new game start message.")
         return
 
     turn_timer.start("Server Processing")
 
     # initialize AIstate
     global foAIstate
-    print "Initializing foAIstate..."
+    logger.debug("Initializing foAIstate...")
     foAIstate = AIstate.AIstate(aggression_input)
     aggression_trait = foAIstate.character.get_trait(Aggression)
-    print "New game started, AI Aggression level %d (%s)" % (
-        aggression_trait.key, get_trait_name_aggression(foAIstate.character))
+    logger.debug("New game started, AI Aggression level %d (%s)" % (
+        aggression_trait.key, get_trait_name_aggression(foAIstate.character)))
     foAIstate.session_start_cleanup()
-    print "Initialization of foAIstate complete!"
-    print "Trying to rename our homeworld..."
+    logger.debug("Initialization of foAIstate complete!")
+    logger.debug("Trying to rename our homeworld...")
     planet_id = PlanetUtilsAI.get_capital()
     universe = fo.getUniverse()
     if planet_id is not None and planet_id != INVALID_ID:
         planet = universe.getPlanet(planet_id)
         new_name = " ".join([random.choice(possible_capitals(foAIstate.character)).strip(), planet.name])
-        print "    Renaming to %s..." % new_name
+        logger.debug("    Renaming to %s..." % new_name)
         res = fo.issueRenameOrder(planet_id, new_name)
-        print "    Result: %d; Planet is now named %s" % (res, planet.name)
+        logger.debug("    Result: %d; Planet is now named %s" % (res, planet.name))
 
     diplomatic_corp_configs = {fo.aggression.beginner: DiplomaticCorp.BeginnerDiplomaticCorp,
                                fo.aggression.maniacal: DiplomaticCorp.ManiacalDiplomaticCorp}
@@ -120,7 +120,7 @@ def resumeLoadedGame(saved_state_string):  # pylint: disable=invalid-name
     global foAIstate
     print "Resuming loaded game"
     if not saved_state_string:
-        print_error("AI given empty state-string to resume from; this is expected if the AI is assigned to an empire "
+        logger.error("AI given empty state-string to resume from; this is expected if the AI is assigned to an empire "
                     "previously run by a human, but is otherwise an error. AI will be set to Aggressive.")
         foAIstate = AIstate.AIstate(fo.aggression.aggressive)
         foAIstate.session_start_cleanup()
@@ -133,7 +133,7 @@ def resumeLoadedGame(saved_state_string):  # pylint: disable=invalid-name
             # assigning new state
             foAIstate = AIstate.AIstate(fo.aggression.aggressive)
             foAIstate.session_start_cleanup()
-            print_error("Fail to load aiState from saved game: %s" % e)
+            logger.error("Fail to load aiState from saved game: %s", e)
 
     aggression_trait = foAIstate.character.get_trait(Aggression)
     diplomatic_corp_configs = {fo.aggression.beginner: DiplomaticCorp.BeginnerDiplomaticCorp,
@@ -154,10 +154,10 @@ def prepareForSave():  # pylint: disable=invalid-name
         return
 
     if empire.eliminated:
-        print "This empire has been eliminated. Save info request"
+        logger.info("This empire has been eliminated. Save info request")
         return
 
-    print "Preparing for game save by serializing state"
+    logger.info("Preparing for game save by serializing state")
 
     # serialize (convert to string) global state dictionary and send to AI client to be stored in save file
     try:
@@ -165,7 +165,7 @@ def prepareForSave():  # pylint: disable=invalid-name
         print "foAIstate pickled to string, about to send to server"
         fo.setSaveStateString(dump_string)
     except:
-        print_error("foAIstate unable to pickle save-state string; the save file should be playable but the AI may have a different aggression.", trace=True)
+        logger.error("foAIstate unable to pickle save-state string; the save file should be playable but the AI may have a different aggression.", exc_info=True)
 
 
 @chat_on_error
@@ -178,10 +178,10 @@ def handleChatMessage(sender_id, message_text):  # pylint: disable=invalid-name
         return
 
     if empire.eliminated:
-        print "This empire has been eliminated. Ignoring chat message"
+        logger.debug("This empire has been eliminated. Ignoring chat message")
         return
 
-    # print "Received chat message from " + str(senderID) + " that says: " + messageText + " - ignoring it"
+    # logger.debug("Received chat message from " + str(senderID) + " that says: " + messageText + " - ignoring it")
     # perhaps it is a debugging interaction
     if handle_debug_chat(sender_id, message_text):
         return
@@ -201,7 +201,7 @@ def handleDiplomaticMessage(message):  # pylint: disable=invalid-name
         return
 
     if empire.eliminated:
-        print "This empire has been eliminated. Ignoring diplomatic message"
+        logger.debug("This empire has been eliminated. Ignoring diplomatic message")
         return
 
     diplomatic_corp.handle_diplomatic_message(message)
@@ -217,7 +217,7 @@ def handleDiplomaticStatusUpdate(status_update):  # pylint: disable=invalid-name
         return
 
     if empire.eliminated:
-        print "This empire has been eliminated. Ignoring diplomatic status update"
+        logger.debug("This empire has been eliminated. Ignoring diplomatic status update")
         return
 
     diplomatic_corp.handle_diplomatic_status_update(status_update)
@@ -252,7 +252,7 @@ def generateOrders():  # pylint: disable=invalid-name
         return
 
     if empire.eliminated:
-        print "This empire has been eliminated. Aborting order generation"
+        logger.debug("This empire has been eliminated. Aborting order generation")
         try:
             # early abort if already eliminated. no need to do meter calculations
             # on last-seen gamestate if nothing can be ordered anyway...
@@ -262,16 +262,16 @@ def generateOrders():  # pylint: disable=invalid-name
         return
 
     # This code block is required for correct AI work.
-    print "Meter / Resource Pool updating..."
+    logger.info("Meter / Resource Pool updating...")
     fo.initMeterEstimatesDiscrepancies()
     fo.updateMeterEstimates(False)
     fo.updateResourcePools()
 
     turn = fo.currentTurn()
     turn_uid = foAIstate.set_turn_uid()
-    print "\n\n\n", "=" * 20,
-    print "Starting turn %s (%s) of game: %s" % (turn, turn_uid, foAIstate.uid),
-    print "=" * 20, "\n"
+    logger.debug("\n\n\n" + "=" * 20)
+    logger.debug("Starting turn %s (%s) of game: %s" % (turn, turn_uid, foAIstate.uid))
+    logger.debug("=" * 20 + "\n")
 
     turn_timer.start("AI planning")
     # set the random seed (based on galaxy seed, empire name and current turn)
@@ -313,7 +313,7 @@ def generateOrders():  # pylint: disable=invalid-name
 
     foAIstate.refresh()  # checks exploration border & clears roles/missions of missing fleets & updates fleet locs & threats
     foAIstate.report_system_threats()
-    print("Calling AI Modules")
+    logger.debug("Calling AI Modules")
     # call AI modules
     action_list = [ColonisationAI.survey_universe,
                    ProductionAI.find_best_designs_this_turn,
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -4,8 +4,11 @@ import pickle  # Python object serialization library
 import sys
 import random
 
-from common.configure_logging import redirect_logging_to_freeorion_logger
-redirect_logging_to_freeorion_logger()
+from common import configure_logging
+configure_logging.redirect_logging_to_freeorion_logger(__name__)
+
+import logging
+logger = logging.getLogger(__name__)
 
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client  # pylint: disable=import-error
 
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -9,8 +9,7 @@ from common.configure_logging import redirect_logging_to_freeorion_logger, conve
 
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client  # pylint: disable=import-error
 
-log_level = logging.DEBUG if fo.getOptionsDBOptionBool("verbose-logging") else logging.INFO
-redirect_logging_to_freeorion_logger(log_level)
+redirect_logging_to_freeorion_logger()
 (debug, info, warn, error, fatal) = convenience_function_references_for_logger()
 
 from common.option_tools import parse_config
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -5,7 +5,7 @@ import sys
 import random
 
 from common.configure_logging import convenience_function_references_for_logger
-(debug, info, warning, error, fatal) = convenience_function_references_for_logger()
+(debug, info, warn, error, fatal) = convenience_function_references_for_logger()
 import logging
 logger = logging.getLogger(__name__)
 
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -3,13 +3,15 @@ these methods in turn activate other portions of the python AI code."""
 import pickle  # Python object serialization library
 import sys
 import random
-
-from common.configure_logging import convenience_function_references_for_logger
-(debug, info, warn, error, fatal) = convenience_function_references_for_logger()
 import logging
-logger = logging.getLogger(__name__)
+
+from common.configure_logging import redirect_logging_to_freeorion_logger, convenience_function_references_for_logger
 
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client  # pylint: disable=import-error
+
+log_level = logging.DEBUG if fo.getOptionsDBOptionBool("verbose-logging") else logging.INFO
+redirect_logging_to_freeorion_logger(log_level)
+(debug, info, warn, error, fatal) = convenience_function_references_for_logger()
 
 from common.option_tools import parse_config
 parse_config(fo.getOptionsDBOptionStr("ai-config"), fo.getUserConfigDir())
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -31,7 +31,7 @@ import ResearchAI
 import ResourcesAI
 import TechsListsAI
 from AIDependencies import INVALID_ID
-from freeorion_tools import chat_on_error, print_error, handle_debug_chat, AITimer, init_handlers
+from freeorion_tools import chat_on_error, handle_debug_chat, AITimer, init_handlers
 from common.listeners import listener
 from character.character_module import Aggression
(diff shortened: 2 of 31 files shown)


Cjkjvfnby:
#606 was about making AI ready at the moment when import done. I don't think this changes should be done in this PR.

For loggers I really don`t know what to say.  Can you attach example file with how loggers will look in python and result here.

LGM-Doyle:
I have removed the commit related to issue #606.

tldr;  I hope that logging will replace the use the print in the code base.

I have included links to the python 2 logging examples and the library.

[python 2 logging examples](https://docs.python.org/2/howto/logging.html)

[python 2 logging library](https://docs.python.org/2/library/logging.html)

You can create arbitrary hierarchical loggers, or arbitrary levels, which can be filtered and redirected to arbitrary handlers.  I've just set it up to create handlers for DEBUG, INFO, WARN, ERROR, CRITICAL and anything else and write them to the appropriate freeorion, freeoriond or ai_N logs.  

The hierarchical nature means that you can turn on/off logging for the entire app or just portions of the hierarchy with just a config file.  For example for release vs debug configurations.  Or turning on logging for a narrow portion of the ai like freeorionAI.research.weapons.plasma.

I did not create any configuration file and I only added the one log message at the end of configure_logging.py

Cjkjvfnby:
I just want to see how it will look like in project. There is not usages of new logging in PR.

I like `print` because it in global namespace and short. Changing prints to logs will make code look more complicated.  IMHO 2 logs `stdout` and `stderr` is enough.  `stderr` can be turned off for release. And adding logger for it looks reasonable.

Vezzra:
@LGM-Doyle, is it ok with you if I come back to this after the release? I don't think I will be able to review this any time soon - I'm still dealing with RL issues that are a bit overwhelming right now, and getting the release out is going to be a considerable challenge as it is.

As you've seen, I haven't even been able to get back to the other PRs where you're waiting for my input (the universe generation stuff etc.). Sorry for the delay, I'm really struggling ATM... 😓

LGM-Doyle:
@Vezzra, thank you for the heads up.  It is considerate.  Please attend to or enjoy your RL issues without guilt.  After all, freeorion is a game supported by voluntary contributions.

Why not delay the 0.4.6 release?  

Most of the devs are absent.  C++ code is not being reviewed or committed at all.  

The 0.4.6 branch hasn't yet been created.

Most of the outstanding issues marked of 0.4.6 are for the AI tuning for research,  weapons, dry dock logic and new supply code. There are no accompanying PRs.  The AI being non-adaptive to changes in FOCS scripts is gated by the balancing which is on-going.  I don't think that these PRs are imminent.

Vezzra:
The 0.4.6 release will already be delayed (because of prolonged absences of almost all lead devs, myself included), and I have a lot of other things I should take care of, so I try to minimize any additions to that list. 😉 

Otherwise, we won't ever get this release out. Currently I expect the release to be finished in July/August (depending on how quickly we get the outstanding issues and PRs done, and how much bugs will turn up during the RC testing phase).

The release branch usually gets created once we are ready to produce the first release candidate, or shortly before that. Currently, we're not even close to that...

geoffthemedio:
needs rebase

LGM-Doyle:
Rebased.

LGM-Doyle:
Rebased.

Vezzra:
Just wanted to drop a note that I still have this on my radar/todo list, but due to being incessantly overhelmed with stuff I still have to ask for your patience, I won't get to it anytime soon - sorry... :frowning_face:

But maybe someone else of our devs can help @LGM-Doyle by reviewing the proposed changes...?

Morlic-fo:
> I like print because it in global namespace and short. Changing prints to logs will make code look more complicated. IMHO 2 logs stdout and stderr is enough. stderr can be turned off for release. And adding logger for it looks reasonable.

I think it generally is a good addition.
As far as log-levels are concerned, I think the following makes sense (for the AI at least):

**Fatal / Error**: Should exclusively used along with ```freeorion_tools.print_error``` to show the error in the chat window as well. Not sure which function should call which one.
**Warning**: Should be used for "minor" errors such as not getting a ship object from a (supposedly valid) ship_id.
**Log**(?): Any server update / message as well as orders issued by the AI client
**Info**: Anything that logs the current AI gamestate such as the table about colonization possibilities or threat assessment.
**Debug**: Strictly for debugging or logging implementation details. Explicitly not for printing high-level game-related info.

So when rating a fleet, the contained ship stats and the final rating may be printed in Info level, any calculation details only on Debug level.

default/python/common/configure_logging.py
```diff
@@ -1,26 +1,124 @@
...
+import utils.configure_logging
+< Use python logging or print and have it re-directed >
+
+Notes on using the standard python logger:
+The python standard library is composed of two basic parts: loggers and handlers.
+Loggers generate the log and associate a whole bunch of level, timing, file, function and
+other information with the log.  Handlers do somthing like stream to file, console or email.
+
+* The simplest is to use the root logger.
+logging.debug(msg)
```
Cjkjvfnby: Please add description to each log where it should be used.  

IMHO 3 loggers will be enough, other can be marked as not recommended.

Cjkjvfnby:
- As developer I did not find instruction how to run game with different log levels.
- Please ensure that log level is printed to game log at any log level.

Vezzra:
Uh, another PR that got buried in my todo list... @LGM-Doyle, I still want to get back to this, but no chance I'll be able to do it anytime soon. However, if our AI team approves of these changes, go ahead and merge... it might be a long time before I can really get back to this.

dbenage-cx:
Really like this addition.

Not necessarily for this PR, but passing the filename and line number back to c++ along with a [similiar](https://github.com/dbenage-cx/freeorion/commit/b0ca703a3e2ebcd864a8bd226e73b55c0a5cc05c) change to logging should cover #1240 

@Cjkjvfnby Unless you are commenting about lack of documentation: the level is handled by the config setting `log-level`.  Lines are formatted to include the level (at least when run from within the ai client)

default/python/common/configure_logging.py
```diff
@@ -1,26 +1,124 @@
+"""
+configure_logging redirects print and the logger to use the freeorion server.
```
Cjkjvfnby: please add:

`log level can be set by freeorion command line option `--log-level`, by default it is DEBUG for developer version`

Cjkjvfnby: Do we use different log level for release builds?

dbenage-cx: > Do we use different log level for release builds?

Not yet, guessing closer to v1.0 it would switch to `INFO`

adrianbroher: > Do we use different log level for release builds?

Not yet and good part of the C++ internal logs to debug when another log level would be more appropriate, but I claim that it shouldn't stay this way.

Cjkjvfnby:
I try to summarize my remaining requests:
- need explanation for each log level when it should be used
- need to add comment how to change log level to `configure_logging.py`
- I prefer to import functions and call them like `error(msg, exec_info=True)`, `debug(msg)`, not making a logger in each file.
```diff
-       logger.debug("This empire has been eliminated. Ignoring diplomatic status update")
+      from configure_logging import debug
+      debug("This empire has been eliminated. Ignoring diplomatic status update")
```

LGM-Doyle:
@Cjkjvfnby I will get back to this.

The design of the hierarchical loggers makes it easy to isolate the log messages from one section of code.

For example if you are debugging the research code, you could turn down the root log level to error and then turn the research logger level to debug.  Then your log file would only show debug, info and warning messages from the research module.

I will try and clarify that when I address your concerns.

LGM-Doyle:
I've made the following changes to this PR.
- Expanding the module docstring to include descriptions of the log levels: FATAL, ERROR, WARNING, INFO and DEBUG.  I paraphrased from the descriptions @Morlic-fo provided.  I left out LOG because, I think that it is a specific separate logger and not a log level of its own.  

  @Cjkjvfnby its is debatable whether we need both FATAL and ERROR because fatal will always crash back to freeorion except in test environment. However, I think that the definitions are clear and fairly standard.  I agree with @adrianbroher that a clearer distinction between debug, info and warning would be useful.

- @Morlic-fo, I added a sink `ConsoleLogHandler` that sends error level messages to the console.  It replaces `print_error`.  It is called for all error messages.  It will always have the location information captured by the standard library logger. `exc_info=True` will make it generate a stack trace for the latest exception.   

- @Cjkjvfnby on startup it now print the initial log level, which as @dbenage-cx pointed out comes from the `OptionsDB` `verbose-logging` option.  @dbenage-cx must be able to see in the future as I only added that today. :grin:

- @dbenage-cx I made a partial fix to #1240.  The debug level logs now print file, function and line numbers for the python module.  However, in order to fix the odd ": : :" in the output it needs a separate sink on the server side, which might conflict with your log rollover work and I decided is outside of the scope of this PR.

- @Cjkjvfnby I added `(debug, info, warning, error, fatal) = convenience_function_references_for_logger()` which creates aliases to the individual level logging functions as you requested. 

For release we might want to turn logging all the way down to error or at least warn for performance reasons.  However, for the weekly test builds which are built with the release tag(?) we should keep the log level at info to maintain the quality of the testing feedback.

geoffthemedio:
If numerous log levels are going to start being used to filter what actually get logged, there should be an expanded option in the gui and optionsdb to set the level, onstead of the verbose logging bool option.

LGM-Doyle:
Yes.

My original plan was:
- this PR.
- a PR to create hierarchical adhoc loggers in the backend which would insert themselves into OptionsDB or have their initial level set by an existing value in OptionsDB.  Use this to replace the combat logger.  This would reduce the current visible gating combat logs and verbose combat logs, because that would happen in the boost::logging framework. Just send the debug logs to the combat logger and if it is set to warnings only then the logs are not even parsed/assembled.
- a PR to add to the options panel a way to automatically populate with the existing loggers some drop down lists that enable switching the log level on the fly.  
- a PR to make the python loggers first class citizens attached to the options panel. 

These changes were intended to be developer focused, so that you can dial the game in to just generate debug logs of only the section of code that you are currently working on.

geoffthemedio:
Just ignoring any logging text is probably not enough, as assembling that text can involve substantial calculation or collection of objects to request info from...

LGM-Doyle:
You are correct if you do the work to assemble the text before passing it to the logging framework.

Both the python and the boost logging frameworks account for that.

The python logger if you pass the inputs in the recommended way,  `debug("A formatted %s contained the number nine %d", "string", 9)` does not examine either the format string or the parameters until after it has determined it is generating the log.  You can query the logger itself as to the current log level, so it is the responsibility of the calling code to ensure that the effort to generate the unparsed parameters is minimal.

Similarly, if I recall correctly from a year ago the boost logging framework uses some macro magic to avoid touching any parameters before determining if the log is needed.

So depending on how attentive to detail you are prepared to be, the cost of an un-generated log can be low, about an single if statement. 

In any case, I thought while not perfect it was a change that incrementally improve the code enough to start with this PR.

default/python/AI/FreeOrionAI.py
```diff
@@ -3,12 +3,16 @@
...
 import random
+import logging
 
-from common.configure_logging import redirect_logging_to_freeorion_logger
-redirect_logging_to_freeorion_logger()
+from common.configure_logging import redirect_logging_to_freeorion_logger, convenience_function_references_for_logger
 
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client  # pylint: disable=import-error
 
+log_level = logging.DEBUG if fo.getOptionsDBOptionBool("verbose-logging") else logging.INFO
```
dbenage-cx: Does not seem to be routed to the intended channel?

Results of AI_1.log for the `initialized` line + next `: : :`

--log-level INFO --verbose-logging 0
```
2017-02-10 16:55:42.294527 [info] AI : : :  configure_logging.py:redirect_logging_to_freeorion_logger():199  - The python logger is initialized with a log level of INFO
2017-02-10 16:55:45.159839 [info] AI : : :  FreeOrionAI.py:generateOrders():211  - Meter / Resource Pool updating...
```

--log-level INFO --verbose-logging 1
```
2017-02-10 16:56:53.941233 [info] AI : : :  configure_logging.py:redirect_logging_to_freeorion_logger():199  - The python logger is initialized with a log level of DEBUG
2017-02-10 16:56:56.832946 [info] AI : : :  FreeOrionAI.py:generateOrders():211  - Meter / Resource Pool updating...
```

--log-level DEBUG --verbose-logging 0
```
2017-02-10 16:57:50.251641 [info] AI : : :  configure_logging.py:redirect_logging_to_freeorion_logger():199  - The python logger is initialized with a log level of INFO
...
2017-02-10 16:57:53.703864 [debug] AI : : : ((4.0, 1), 15, 0, 0, 0, 0)
```

--log-level DEBUG --verbose-logging 1
```
2017-02-10 17:01:07.298026 [info] AI : : :  configure_logging.py:redirect_logging_to_freeorion_logger():199  - The python logger is initialized with a log level of DEBUG
2017-02-10 17:01:07.480485 [debug] AI : : : FreeOrionAI.py:<module>():54  - Path to folder for user specific data: /home/cxdamian/.local/share/freeorion
```

Not certain what the full intent here is, hopefully it is not to override the `log-level` config setting.

geoffthemedio:
My point was that there will need to be some if blocks that test the log level wrapping stuff like for loops or calls to Dump functions or log output formatting code, and that it's not enough to just ignore passed-in text in the logging code.

Cjkjvfnby:
@geoffthemedio, I see next approach: code logging mostly as functions and add check to them for log level.
```
def if_info(function):
    """Decorator that skips execution of function if log level is not INFO"""

@if_info
def print_info(msg...):
    """Print message if log level is INFO"""
    print msg...
```

default/python/AI/AIDependencies.py
```diff
@@ -20,6 +20,8 @@
     my_industry = AIDependencies.INDUSTRY_PER_POP * my_population
 """
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client  # pylint: disable=import-error
+from common.configure_logging import convenience_function_references_for_logger
+(debug, info, warn, error, fatal) = convenience_function_references_for_logger(__name__)
```
Cjkjvfnby, line 24: May be we can import this https://docs.python.org/2/library/logging.html#logging.Logger.info  and use it?

I wan to have option to import single log function:  `from common.configure_logging import debug`, but this will require to play with frames to get proper filename.

default/python/AI/FreeOrionAI.py
Cjkjvfnby: Move this line to  `redirect_logging_to_freeorion_logger`

Vezzra:
I see that work on that PR is finally moving forward, which is good. Sorry if I held it up with my constant promises to get back to it and never being able to get around to that. Please don't wait for input from me anymore, otherwise I'm afraid this PR won't get merged for years... 😉

default/python/AI/AIDependencies.py
LGM-Doyle: Unless I misunderstand what you want.  The indicated line does what you requested.  If you do not want a given logger, then replace it with `_`

LGM-Doyle: Unless I misunderstood what you requested, the indicated line does what you want.  If you don't want a given logger, replace it with `_`.

default/python/AI/FreeOrionAI.py
LGM-Doyle: I did not wire up `--log-level`, just `--verbose-logging` and I only checked that it worked from `persistent_config.xml`  not the command line.  I'll look into that.

I don't want to get too deep into wiring up the back end until you finish the log rollover code.  This PR was planned to be just the python portion and controlled from within the python code itself.

LGM-Doyle: The logger code is common code and does not have access to the `fo.` server functions.  So at least the lookup has to remain in the AI code.

The universe generation and the AI won't necessarily have the same default log level.

Could we break up the FreeOrionAI into two modules: one that is the entry point for the C++ and a second that is the top-level code for the AI.  Currently, FreeOrionAI.py serves both purposes and we end up hiding initialization code in imports in order to make the AI code look nice.  If we separated the entry point from the AI controller, then the initialization code in the entry point would be expected, in a central location and easier to improve.

LGM-Doyle (commented):
@geoffthemedio your point is correct.  It will be interesting to see if it impacts performance.

default/python/AI/FreeOrionAI.py
dbenage-cx: Ok, probably a misunderstanding on my part.

default/python/AI/AIDependencies.py
Cjkjvfnby: It is look very ugly, and uncommon.
`(debug, _, _, _, _) = convenience_function_references_for_logger(__name__)`

I didn't ever use python logging as raw code, only as part of frameworks.  I don't know how to do it properly.

I see next way:
`from freeorion_logging import debug`

Log setup will be done in freeorion_logging module (execute on import).  Have import from it -> have logging setuped, no import no setup.  

This function should have same signature as logging `debug`,   Just save link it in module. `debug = logging.debug`

LGM-Doyle:
I've brought this PR up to date with the changes to `Logger` on the C++ side.  
- the python `print` and `logger` messages are routed to a `Logger(python)` instance.
- the `Logger(python)` can be controlled via the Options window the same as the other loggers.
- the python `print` and `logger` messages report the filename and line number of the python file not the C++ wrapper for the `Logger`.  This fixes issue https://github.com/freeorion/freeorion/issues/1240.

Cjkjvfnby:
For python part look ready for merge.

python/LoggingWrapper.cpp
```diff
@@ -1,57 +1,159 @@
...
-                log_buffer = "";
-            } else {
-                log_buffer += text[i];
-            }
+    static const std::size_t MAX_SINGLE_CHUNK_TEXT_SIZE = 4096;
+
+    // Python sends text as several null-terminated array of char which need to be
+    // concatenated before they are output to the logger.  There's probably a better
+    // way to do this, but I don't know what it is, and this seems reasonably safe...
+    void send_to_log(std::stringstream & ss, const std::string & input, const std::function<void(const std::string&)>& logger) {
```
geoffthemedio: remove spaces before `&`s

python/LoggingWrapper.cpp
```diff
@@ -1,57 +1,159 @@
...
-            if (text[i] == '\0') break;
-            if (text[i] == '\n' || i == MAX_SINGLE_CHUNK_TEXT_SIZE - 1) {
-                DebugLogger() << log_buffer;
-                log_buffer = "";
-            } else {
-                log_buffer += text[i];
-            }
+    static const std::size_t MAX_SINGLE_CHUNK_TEXT_SIZE = 4096;
+
+    // Python sends text as several null-terminated array of char which need to be
```
geoffthemedio: Is this comment relevant now? The parameter type is no longer `char *` Does the `std::string input` still have null characters in it? If not, then the search within it seems unnecessary...

python/LoggingWrapper.cpp
```diff
@@ -1,57 +1,159 @@
...
+    // concatenated before they are output to the logger.  There's probably a better
+    // way to do this, but I don't know what it is, and this seems reasonably safe...
+    void send_to_log(std::stringstream & ss, const std::string & input, const std::function<void(const std::string&)>& logger) {
+        if (input.empty()) return;
+        ss <<  ((input.size() < MAX_SINGLE_CHUNK_TEXT_SIZE) ? input : input.substr(0, MAX_SINGLE_CHUNK_TEXT_SIZE));
```
geoffthemedio: why is `ss` a parameter and not a local?

python/LoggingWrapper.cpp
```diff
@@ -1,57 +1,159 @@
...
+
+    // Assemble a python message that is the same as the C++ message format.
+    void PythonLogger(const std::string& msg,
+                      const LogLevel log_level,
+                      const std::string& python_logger,
+                      const std::string& filename,
+                      // const std::string& function_name,
+                      const std::string& lineno)
+    {
+        // Keeping the string assemble in of this switch statement means that it is gated by
```
geoffthemedio: weird grammar happening in here... rewrite?

LGM-Doyle (commented):
@Cjkjvfnby thanks for looking over the python.

I've made the requested changes.

I also removed a line of logging in Logger.cpp `StoreConfigurerWithLoggerName()` which was causing the boost::logging singleton's mutex to sometimes deadlock during static initialization.

Dilvish-fo:
@LGM-Doyle Ack!
This had seemed to work fine generally, but I just ran into some strange trouble with it, after rebasing an experimental branch to current master I am getting an AI crash with an unhelpfully terse logfile; the end of it is as follows (the preceding portion appeared to just all be fine setup reporting). 
```
20:28:12.965800 [debug] log : LoggerWithOptionsDB.cpp:105 : Configure log source "python" from optionsDB using threshold debug
20:28:12.965815 [debug] python : configure_logging.py:185 - Python stdout and stderr are redirected to ai process.
20:28:12.966006 [info] root : configure_logging.py:197 - The python logger is initialized with a log level of %s
20:28:13.005596 [error] ai : AIClientApp.cpp:188 : AIClientApp failed due to error in python AI code for AI_1.  Exiting Soon. id = -1
```

Even setting the AI logging to trace just gave me 22k of extra lines of intro prep reporting, but did not add anything relevant to the crash, the final two lines are the same.

If you want to test this out yourself, the [branch is one from Morlic](https://github.com/Morlic-fo/freeorion/tree/UniverseGraphTheory), to use some graph theory in AI planning, and I've been fiddling with it a bit also.   By itself it runs fine, and if I rebase to the point in master where 1656 was merged in, then it still runs fine, but if I rebase to the point where this PR is merged in and then start a single player game with a single AI, the AI immediately crashes per above.

LGM-Doyle:
@Dilvish-fo I think that the problem lies in the new file, `UniverseStrategyAI.py`.  `print_error` no longer exists.  This patch might fix the problem, although I was not able to test it since I do not have networkx.

I also fixed the string interpolation of the `he python logger is initialized with a log level of %s` message.

````patch
diff --git a/default/python/AI/UniverseStrategyAI.py b/default/python/AI/UniverseStrategyAI.py
index 6d6bec7..7e5543b 100644
--- a/default/python/AI/UniverseStrategyAI.py
+++ b/default/python/AI/UniverseStrategyAI.py
@@ -4,7 +4,6 @@ from functools import wraps
 
 import freeOrionAIInterface as fo
 import FreeOrionAI as foAI
-from freeorion_tools import print_error
 from graph_interface import Graph
 
 
@@ -23,7 +22,7 @@ class BrokenUniverseGraphException(Exception):
 
     def __init__(self, fnc_name=""):
         self.message = "Function %s broke the UniverseGraph instance" % fnc_name
-        print_error(self.message)
+        error(self.message)
 
 
 class UniverseGraph(Graph):
@@ -231,7 +230,7 @@ def find_defensive_positions_min_cut(weight_owned, weight_enemy):
         # note that the finally-block is executed even if we exit the function using a return statement
         return __universe_graph.minimum_st_node_cut(SOURCE, SINK, weight_fnc)
     except Exception as e:
-        print_error(e)
+        error("Some other message", e, exc_info=True)
         return set()
     finally:
         # remove the previously added nodes and edges
@@ -294,7 +293,7 @@ def find_inner_systems(g=None):
                 inner_systems.update(subnodelist)
         return inner_systems
     except Exception as e:
-        print_error(e)
+        error("Some message", e, exc_info=True)
         return set()
     finally:
         # restore the previously added nodes and edges
````

How does your python interpreter spit out exceptions?  My python exceptions are passed through abrt and then show up in the AI_x.log as `[error] : python abrt_exception_handler.py:`.

**Comment by Dilvish-fo** [comment]
> This patch might fix the problem

It did indeed stop the crash, so at least that is taken care of, thanks much.  Although I think it also needed 
```
from common.configure_logging import convenience_function_references_for_logger
(debug, info, warn, error, fatal) = convenience_function_references_for_logger(__name__)
```
at the top?  (although then I get a doubled-up log line about ```driver.py:120 - Generating grammar tables from %s``` but without it if I try adding something like ```error("test")``` that would cause an exception and crash)

@Morlic-fo  without the above patch, if you try running your UniverseGraphTheory branch after rebasing it to master, do you get the same near-silent AI crash that I got, or does the AI log still give you the full exception information identifying the problem import?

I have a strong memory that it *used* to be the case that if the AI had an uncaught exception, the full exception info would be printed in the AI log file.  Checking now, though, it's apparent that has changed for me at some point prior to this PR.  If I force an exception in  a test script run from the command line that just says 
```from nothing import foo```

then I get regular exception reporting like I strongly recall I used to get in my AI log files from such things:
```
~/programs/freeorion_builds/kdev$ python test.py
Traceback (most recent call last):
  File "test.py", line 1, in <module>
    from nothing import foo
ImportError: No module named nothing
```
However, if I add that same line to an AI python file and then run FO, I just get a bareboned crash like in my first post above.  @LGM-Doyle if you try adding the above import to one of the AI scrpts do you get detailed exception info in your logs?   

It's my understanding that on Ubuntu Apport pretty much corresponds to Abrt on fedora/redhat, but I thought that's really for fullblown crash dumps, not simple exception reporting like this.

LGM-Doyle:
On my machine python errors still generate the Traceback.  The UniverseGraphTheory branch generated this:

````ImportError: cannot import name print_error````

It was printing and still does print the Traceback for me.

Could you check if the Traceback was printed in the log on Ubuntu just prior to merging this PR?  If possible could you do a bisection to determine which commit broke Ubuntu's python Traceback?

Morlic-fo:
> does the AI log still give you the full exception information identifying the problem import?

Can't recompile at the moment to check this particular case. In general, import errors on initialization will crash the AI  and then terminate the server without any log entries for me and did so for quite some time.
An import error at a later time (e.g. during debug mode or by importing within a function call) gives the expected traceback.

Cjkjvfnby:
> Even setting the AI logging to trace just gave me 22k of extra lines of intro prep reporting, but did not add anything relevant to the crash, the final two lines are the same.

It is default windows behavior. I already report about it (somewhere) long time ago.  (Tested on `v0.4.7+ [build 2017-08-02.b9fdf9e] MSVC 2015`)

Dilvish-fo:
> Checking now, though, it's apparent that has changed for me at some point prior to this PR

I'll do the git bisect process.

PR #673, opened 2016-05-20
Move profiling to separate function.

Cjkjvfnby:
Disable it for `find_best_designs_this_turn`


Cjkjvfnby:
Looks like this function is useless without some analyzing tools. 
Need store information between couple of runs, and scripts to measure optimization.
Will make this PR back again with them.

PR #675, opened 2016-05-25
Move profiling to separate function

Cjkjvfnby:
- extract profile to separate function
- disable it for find_best_designs_this_turn

Here is possible usecase:
- add decorator to `generateOrders`
- play game on same settings couple of times
- change something
- play couple of times
- run script:
  
  ``` python
  import os
  import re
  import matplotlib.pyplot as plt
  
  base = 'c:/Users/sa/AppData/Roaming/FreeOrion/profiling/'
  
  call_regexpr = re.compile('(\d+) function calls \(\d+ primitive calls\) in ([0-9.]+) seconds')
  
  
  def extract_time(path):
  times = []
  with open(path) as f:
      for line in f:
          if 'function calls' in line:
              times.append(call_regexpr.search(line).groups())
  return times
  
  
  def show_time_graph():
  all_times = []
  for name in os.listdir(base):
      all_times.append(extract_time(os.path.join(base, name)))
  
  for i, t in enumerate(all_times):
      plt.plot([float(x[1]) for x in t])
  
  plt.show()
  
  
  show_time_graph()
  ```

Result I got:
test

I did not find any bottleneck that can be fixed easily.

### DIFF
--- a/default/AI/ProductionAI.py
+++ b/default/AI/ProductionAI.py
@@ -11,10 +11,6 @@ import PriorityAI
 import ColonisationAI
 import MilitaryAI
 import ShipDesignAI
-from time import clock
-import cProfile
-import pstats
-import StringIO
 
 from EnumsAI import (PriorityType, EmpireProductionTypes, MissionType, get_priority_production_types,
                      FocusType, ShipRoleType, ShipDesignTypes)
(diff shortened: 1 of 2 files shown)


LGM-Doyle:
The code looks good to me and it seems like a good idea.

Issue #665, opened 2016-05-15
Add `different` option for AI aggression for testing purposes 


Cjkjvfnby:
We have 100500 "switches" for AI aggression in code.
Game creation config option say `Max AI aggression` but it works as `AI aggression`:  all AI agression are equal.

I want to have something like `aggression_list[empire_id % len(agression_list)]` to have all 6 aggression if game started with 6 players.

It will speedup testing if something is not broken.

I don`t sure how to implement it yet.  Via python config or game config option.

Cjkjvfnby:
Issue example: Some aggression names are not displayed in greeting string.

```
AI_3: Affiliation (): In both of our interests, let us leave each other alone.
AI_1: Territory (Beginner Mode): What have we done to deserve your wrath, oh mighty ones?
AI_2: Control (Aggressive): Welcome, dear enemy, our fearless warriors are waiting for you...
AI_6: Seigniory (): To do battle with you will be an honor.
AI_4: Totality (Maniacal): Alien scum! We shall cleanse the galaxy of your filthy presence!
AI_5: Union (Turtle): Stay the hell away from us!
```

Fix for it: https://github.com/freeorion/freeorion/pull/667

PR #677, opened 2016-05-27
Add AI behavior and character.

LGM-Doyle:
Please use the following as the merge message 

**************\* Merge Message Cut Here ***********************************
Create AI Character composed of Behaviors to encompass variation 
between AIs.

Character solves the problem of having AI personalization added
piecemeal throughout the AI code.  Now each type of personalization is
encapsulated in a single Behavior.  The Behaviors are combined to form a
single unique Character per AI.

Refactor current variation in AI into two Behaviors, Aggression and 
EmpireIDBehavior to encompass AI variation due to the aggression setting 
and the empire id.

Add debugging bypass to randomized Character and Behavior.

The bypass can force individual AIs to have a specific
behavior and/or force all other AIs to have the same default behavior. 
********************\* End of Merge Message ****************************

I think this PR addresses issue #665.

It is a first step to more testable personalization of the AI.  If each behavior is kept orthogonal to the other behaviors then they can be tested in isolation, for functionality and performance.  This will keep testing manageable.


default/AI/Behavior/Character.py
```diff
@@ -0,0 +1,305 @@
...
+    def may_research_tech(self, tech):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to research ''tech''"""
+        return True
+
+    def may_research_tech_classic(self, tech):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to research ''tech'' in the classic algorithm"""
+        return True
+
+
+class Character(Behavior):
```
Cjkjvfnby: Why it is inherited from `Behavior` if it is collection of `Behaviors`?

default/python/test/AI/Behavior/freeOrionAIInterface.py
```diff
@@ -0,0 +1,20 @@
+# Fake fo for testing Behavior purposes
+
+# TODO Find a better way to import the static portions of fo without the test chaos, overhead and random state of fo.
```
Cjkjvfnby: This is mock for autocomplete. It is importable.   You can take `aggression` from there
https://github.com/freeorion/freeorion/blob/master/default/AI/freeorion_debug/ide_tools/result/freeOrionAIInterface.py

default/python/test/README.testing
```diff
@@ -0,0 +1,12 @@
+To run tests from freeorion/default/python you first need to set the
+python path to include AI.  For example:
+
+linux:
+PYTHONPATH=../AI bash -c 'py.test'
+
+OSX:
+PYTHONPATH=../AI bash -c 'py.test'
+
+windows:
```
Cjkjvfnby: I will add win part.

Cjkjvfnby:
Arr, 904cc649fec32938db6e4a7db34abb052c4354a4 how does this related to PR? Will merge this commit myself to master.

This is not related to this PR too. f1bdb2d098c36d5028266b4f4d343ca0cdee261e, please make it separate PR and don't forget about usages in c++ code

It will take me some time to review this PR.

Cjkjvfnby:
You use outdated master as base for your branch.

default/stringtables/en.txt
```diff
@@ -1803,9 +1816,9 @@ GSETUP_BEGINNER
 Beginner Mode
 GSETUP_TURTLE
 Turtle
-GSETUP_DEFENSIVE
+GSETUP_CAUTIOUS
 Cautious
-GSETUP_MODERATE
```
dbenage-cx: Might also change the references in UI/GalaxySetupWnd.cpp:269 and util/EnumText.cpp:40

default/AI/Behavior/Character.py
LGM-Doyle: It is a collection of Behaviors that implements the interface of Behavior.  It is both.

I chose not to use abstract base class (abc) for the inheritance because that forces you to re-implement every method.  Most Behaviors should only need to re-implement a few methods.  Re implementing all of them adds noise, hides the purpose of the Behavior and can only introduce bugs.

default/python/test/AI/Behavior/freeOrionAIInterface.py
LGM-Doyle: I tried to take `agression, userString and userStringList` from where you indicated.  It only works if I have empty **init**.py files in ./freeorion_debug/ and ./freeorion_debug/ide_tools/results/.  The imports in ./freeorion_debug/**init**.py assume that a running freeorion session exists.  I made a comment in the code, but did not change the **init**.py file, because that will break the interactive debugger.

Once the imports are sorted out it works perfectly.

If you have some insight in how to fix it then I will fix it.  Otherwise, I'll leave it as is and fix it after I understand how the interactive portion of the freeorion_debug module works.

default/python/test/README.testing
LGM-Doyle: Thank you.

default/stringtables/en.txt
LGM-Doyle: Done.

LGM-Doyle:
https://github.com/freeorion/freeorion/commit/904cc649fec32938db6e4a7db34abb052c4354a4 seemed easier to tag on this PR than to have a whole PR of its own.  Sorry.

https://github.com/freeorion/freeorion/commit/f1bdb2d098c36d5028266b4f4d343ca0cdee261e, is necessary for  `make_aggression_table` in CharacterStrings.py to work, so I left it in.

I rebased to https://github.com/freeorion/freeorion/commit/e888bedff1ee7415f783385fdfe5fa8892e55954, before I pushed yesterday.  Probably, I made a mistake with fixing the conflicts with the camel_case.  I've looked and I can't find them.  Please, point out where I made a mistake and I'll fix them.

A merge would have been better than a rebase, because there would have been a merge commit and finding the problem would be easy.

@Cjkjvfnby , clearly you have a vision for how testing should be setup.  If you want me to remove the tests from this PR, I'm good with that.  However, I do think that starting with some tests and modifying them into better tests is a way to keep progressing towards a more completely tested app.

Cjkjvfnby:
@LGM-Doyle, your branch `origin/add_ai_behavior`  is not changed.  Your `origin/master` is far away form `upstream/master`.

> 904cc64 seemed easier to tag on this PR than to have a whole PR of its own. Sorry.

I already cherry-picked this comment to master and pushed. 

> f1bdb2d, is necessary for make_aggression_table in CharacterStrings.py to work, so I left it in.  

You can cherry-pick this commits to new branch from master.  I want this in separate branch for single reason: It can be merged much faster then other code. 

> A merge would have been better than a rebase

Personally I agree with that, but project rules require rebasing.

PS. How common you ar with git?   I can help with rebasing. Do you use console for git?

> clearly you have a vision

no, i did not try to lunch it yet

Cjkjvfnby:
Before force push rebased branch make `git diff master origin/master > rebase.patch` and check what was really changed.

Cjkjvfnby:
Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability.  https://www.python.org/dev/peps/pep-0008/#id38

Cjkjvfnby:
For example test can be run from python file:

Put file to `freeorion\default\python\test\AI` folder and run it with python

```python
import os
import sys

import pytest

this_dir = os.path.dirname(__file__)

sys.path.append(os.path.join(this_dir, '..', '..', '..', 'AI/'))

pytest.main(['-v', '--tb', 'long', os.path.join(this_dir, 'Behavior')])
```

Tests are failes for me:
```
================================== FAILURES ===================================
_________________________ test_make_aggression_table __________________________

    def test_make_aggression_table():
        """Check that the table adds the correct prefixes, suffixes, runs the postfix function,
        has the UNKNOWN value entry and is the correct length.
        """
        table = make_aggression_table("prefix_", lambda x: "post_proc_%s" % x)
    
        character = Character([Aggression(0)])
>       assert table[character] == "post_proc_prefix_BEGINNER"
E       assert 'post_proc_prefix__BEGINNER' == 'post_proc_prefix_BEGINNER'
E         - post_proc_prefix__BEGINNER
E         ?                  -
E         + post_proc_prefix_BEGINNER

Behavior\CharacterStrings_test.py:12: AssertionError
---------------------------- Captured stdout call -----------------------------
CharacterTable returns  post_proc_prefix__BEGINNER
___________________ test_make_aggression_table_no_post_func ___________________

    def test_make_aggression_table_no_post_func():
        """Check that the table doesn't run the post function if it doesn't exist"""
        table = make_aggression_table("prefix_")
    
        character = Character([Aggression(0)])
>       assert table[character] == "prefix_BEGINNER"
E       assert 'prefix__BEGINNER' == 'prefix_BEGINNER'
E         - prefix__BEGINNER
E         ?       -
E         + prefix_BEGINNER

Behavior\CharacterStrings_test.py:28: AssertionError
---------------------------- Captured stdout call -----------------------------
CharacterTable returns  prefix__BEGINNER
===================== 2 failed, 7 passed in 0.05 seconds ======================
```
```

**Comment by Cjkjvfnby** [comment]
We use python style for our code (PEP8). Please follow it for all files.

I use IDE PyCharm (community edition is free) it allow to reformat files according python style.  Style issue that are present after reformatting are highlighted.  You can easy fix them using `quick fix` action.

**Comment by LGM-Doyle** [comment]
I have posted a new separate PR #683 with the stringtable changes.

I see why there is confusion about the commit on which I based this PR. 

I fetch directly from freeorion/master to my workspace.  The
LGM-Doyle/master branch is rarely current.  But if you check the
add_ai_behavior branch then you can see that it is based on the current
freeorion/master. 

I will clarify what I mean about the difference between rebasing and
merging. 

When you rebase 10 commits and have to fix some rebase conflicts, then
the changes from the fixes may end up in any of the 10 new rebase commits.

When you merge all of the changes from the merge conflicts end up in the
one single merge commit.

It is easier to verify one merge commit that only contains merge changes
and fix any errors with a single fixup commit, than to have to hunt
through several rebase commits that also include all of the changes from
the PR as well as the rebasing errors.

I added the python tests to my push-hook, so hopefully test failures
will not slip through in future.  BTW the test failed, but the code
worked. 

I left the modules names as CamelCase because most of the modules are
named in that fashion and all the modules that it interacts with are
named in that fashion.  Plus, in the first section of PEP8 
"A Foolish Consistency is the Hobgoblin of Little Minds ". it says being
consistent with the surrounding code counts.

If freeorion wants to change the file names perhaps it should all be
done at once in a single commit instead of haphazardly.

Do we want to have different naming conventions for the C++ files and
the python files?

**Comment by Cjkjvfnby** [comment]
I am against rebasing too, but we live in cruel world.

> BTW the test failed, but the code worked. 
Tests or code are broken and need to be fixed.

> If freeorion wants to change the file names perhaps it should all be done at once in a single commit instead of haphazardly.

Want. But this have a big chance to have conflicts with branches in progress. It is low priority task and can wait for the good time.

> Do we want to have different naming conventions for the C++ files and the python files?
Yes. I have plan to rename C++ bindings, but it is low priority too.

**Comment by LGM-Doyle** [comment]
@Cjkjvfnby I will send you a note if I do a sneaky force push to update a PR.

I have just done a sneaky push on this PR.  I removed the string table elements that you just merged with master from the first commit of this PR.  The new commit is [url](https://github.com/freeorion/freeorion/pull/677/commits/9db51dc49b273589bf492934f599c718d33d25b8)

By the way the tests have been passing since my last push.  I didn't mean to imply that the tests not passing and the code working was a good thing.  I was merely pointing out the irony of me committing the first set of tests and failing to run the tests in my push-hook.

**Comment by MatGB** [comment]
It was my understanding the 'no merges' rule was more properly 'no merges from master', as we had a few incidents where someone had updated their fork via merge and it created errors, however my git-fu is low, I see no personal issue with merging a branch into a different branch in order to keep track of changes within an ongoing project as long as you never merge master into it, rebasing is definitely best there.

**Comment by Cjkjvfnby** [comment]
AI have 3 Behaviors in same time. Looks like human brain disease.

AI should have single Behevior.

I think we can replace combination with join:

Character will have exact number of behaviors, for example `ProductionBehavior`, `ResearchBBehavior`, `ColonizationBehavior`. All this behavior does not intersect by methods, and we just join them together.

Method combination can be hard to understand in case of reading logs. You need to find in logs which Behavior ai used, check theirs methods and join them in you head.

File: default/AI/Behavior/Character.py [hunk 5]
```diff
@@ -0,0 +1,492 @@
...
+    """
+
+    def __init__(self, behaviors):
+        self.behaviors = behaviors
+        if not all([isinstance(x, Behavior) for x in behaviors]):
+            raise TypeError("All behaviors must be sub-classes of Behavior")
+
+    @property
+    def key(self):
+        return super.key
```
Cjkjvfnby: 1) it does not work at all
2) it have no sense to override method of parent with method that call method of parent

default/AI/AIstate.py
```diff
@@ -110,6 +110,37 @@ def __init__(self, aggression=fo.aggression.typical):
...
+                print "Set  %s to a value to force the behavior of all AIs not individually specified." % (all_id_option,)
+                behavior = default
+            print "%s behavior bypassed and set to %s for %s" % (name, repr(behavior), fo.playerName())
+            return behavior
+
+        no_value = -1
+        _behavior_aggression = get_behavior_bypass_value("aggression", aggression, no_value)
+        _behavior_id = get_behavior_bypass_value("empire-id", self.empireID, no_value)
+
+        self.character = create_character(_behavior_aggression, _behavior_id)
```
Morlic-fo: The entire AIstate instance is pickled into save files.

Is it a good idea to store a (complex) object which implementation may be subject to future changes here?
I'd consider to recreate the character object on loading and only store the most important information (needed to recreate the object) in save files.

default/AI/Behavior/Character.py
LGM-Doyle: You are correct.  That is held over from when Behavior was an abc and every property needed an override.

LGM-Doyle:
It is meant to be like a brain, quirky.  But it is also meant to be easy to debug.

I agree that the combiner Character has no log output and it needs it.  The logs should always be clear exactly which behavior resulted in a denial.  You are right, you should never be figuring out which behavior is dominant when the Character can just log it.

The behavior methods are intended to be of two flavors.  Permission type may_<activity>() permits a certain activity. Preference type preferred_<activity>(alternatives) selects the most preferred of the alternatives.  

The combiners in Character are simple.

The may_<activity>() combiner forbids activity if any behavior forbids the activity.  The log could be like this for a rejected call to character.may_research(LRN_TRANSCEND) :

```
`Character forbids researching LRN_TRANSCEND because
          permits: EmpireIDBehavior, SomeOtherFutureBehaviorNotYetImplemented
          forbids: Aggression`
```

The specifics for why Aggression forbids would appear in the Aggression's own logging code presumably in the previous line.

The preferred_<activity>(alternatives) combiner just counts up votes.  Currently all of the examples of preferred_ are in the EmpireIDBehavior which I think should be removed.  I will use a hypothetical example for the log.  A log for a request to preferred_invasion_target([planet1, planet2, planet3]) could look like:

```
`Character prefers attacking planet2 because
Alternatives    planet1     planet2     planet3
Votes              1            3          0
                             Nemesis
                Loyal 
                             Aggression
                             Genocidal
Abstentions: Chattiness`
```

Similarly, the individual behaviors will have logged their calculations previously in the log.

I will add logging as described above to the PR.

I will illustrate why behavior needs to be common across the modules.

If we are going to implement behavior for each module individually, then it should be left in the specific module's code as it is today.  If a choice is specific enough that it is only needed in one module and is only dependent on one aspect of behavior then it should not be in behavior.  You can retrieve a single behavior from Character with get_behavior, for example character.get_behavior(Aggression) for Aggression.

The current code is simply a refactoring of all the aggression and (empire id % n) conditional logic from throughout the AI into a single location.  That was messy and this still is messy.  There are too many methods that ask almost the exact same question of the behaviors.  There were some cases where the previous code was just inconsistent.

Many of the current method are too specific and tied directly to arbitrary coefficients from the original module.  A good example is may explore system, which is based on a 'monster_threat' threshold which is only defined in one module.  These need to be refactored to ask more generic useful questions.

Others like may_research() and may_research_classic duplicate() functionality with slightly different outputs.  They will need be merged into a single method.

A good example of generic behavior method that are useful across many modules are: may_output(), may_colonize(), and may_invade().  Obviously, research, ship designer, military planner and colonization AI all benefit from a uniform answer to those questions.

The next refactor should move towards all the behavior methods being good methods and refactoring the AI modules to use the more generic results appropriately.

Each Behavior should be moved to its own file.

Since each Behavior is derived from Behavior it only needs to re-implement the methods it cares about.  It should be easy for a dev to look over a method and figure out what say Aggression affects.

Since the combiner implements all of the methods it is a simple single calling point for all of the AI modules.

LGM-Doyle:
@MatGB first, I have abided by the no merging rule.  I was clarifying that my rebasing to master may have caused errors in any of the commits I had rebased, due to the many renamed variables, but that I couldn't see them myself.

I don't know the incidents from the past so I'm not commenting on them or trying to change the decision to only rebase.

However, I'd like to clarify that the advantages of merge over rebasing are particularly evident when keeping a long running branch tracking another long running branch such as master and a feature.  A rebasing only policy strongly discouraged long running branches.

The advantage to merge is that all the fixes for the merge conflicts are in the single merge commit.  The whole merge can be undone by removing the one commit.  Any errors in the merge can be fixed by fixing one commit.

With rebase all of the fixes for the rebase conflicts are scattered in any of the commits and the ancestors are not kept.  

I'll try to explain with ascii drawing, but you could also read about rebasing vs merging in https://www.atlassian.com/git/tutorials/merging-vs-rebasing/ or http://mislav.net/2013/02/merge-vs-rebase/.  An older but popular article on git workflow is http://nvie.com/posts/a-successful-git-branching-model/.

Here is the notation I will use.  There are two branches: master and feature. Commits on master are m1, m2, m3, etc. and on feature f1, f2, f3 etc.  Rebasing creates new commits like f1.1 or m2.1.  Merge commits are indicated by x.m1.f3.

The scenario that I'll depict is a long running feature branch that has master merged/rebased into it once with conflicts and then is later merged into master conflict free.

Starting condition (same for both cases):
Each of these commits has been tested by the dev who committed them.  If they don't change they do not need to be reverified.

```
`m0 -- m1 -- m2 -- m3
  \
   \
    f1 -- f2 -- f3`
```

Merge case:

Merge master into feature and fix conflicts:
This creates a merge commit with the fixes, but leaves f1, f2, and f3 unchanged.
It doesn't matter which f1, f2, or f3 commit caused the conflict all the conflict resolution is in the x.m3.f3 commit. The only new commit that needs to be verified is x.m3.f3.  It should be verified by both the f3 and m3 committers/testers, to make sure their changes stuck.

```
`m0 -- m1 -- m2 -- m3
  \                 \
   \                 \
    f1 -- f2 -- f3 -- x.m3.f3`
```

Merge again and clean up more conflicts:

```
`m0 -- m1 -- m2 -- m3 -- m4 -- m5 -- m6 
  \                 \                  \
   \                 \                  \
    f1 -- f2 -- f3 -- x.m3.f3 -- f4 --  x.m6.f4`
```

Complete the feature by merging into master:

```
`m0 -- m1 -- m2 -- m3 -- m4 -- m5 -- m6 -- x.m6.(x.m6.f4) --
  \                 \                 \   /
   \                 \                 \ /
    f1 -- f2 -- f3 -- x.m3.f3 -- f4 --  x.m6.f4`
```

The importance here is that f1, f2, f3 and f4 are the exact same commits that the dev verified.  The state at any point prior to any merge is accessible by checking out an f1, f2, f3 or f4 commit.  Only the two merge commits need to be verified, by some of the devs involved.  x.m6.(x.m6.f4) has no conflicts and is there only to prevent the master minder from doing merge work.  If you allow fast-forward merges, which are a good thing, then vacuous merges can be omitted.

Rebase case:

Rebase master into feature and fix conflicts:
This creates new commits for f2 and f3 because there were rebasing conflicts.
There is no indication in the commit record that the rebase happened.
I wrote f1.1 with no suffix to indicate it had no rebase conflicts and so is the same as f1.  In reality it in a new commit with a new hash because it has a new parent.
The new commits f1, f2.1 and f3.1 need to be verified by all committers/testers of m1, m2, m3, f1, f2 and f3 because they are 3 new commits that contain or all of the changes from the previous 6 commits.

```
`m0 -- m1 -- m2 -- m3
                    \
                     \
                      f1 -- f2.1 -- f3.1 `
```

Merge again and clean up more conflicts:
In this case I assume the conflicts start in f3.1, so only f3.1 and f4 have new suffixes.

```
`m0 -- m1 -- m2 -- m3 -- m4 -- m5 -- m6 
                                      \
                                       \
                                        f1 -- f2.1 -- f3.2 -- f4.1`
```

Complete the feature by merging into master:

```
`m0 -- m1 -- m2 -- m3 -- m4 -- m5 -- m6  --------------------------x.m6.f4.1 --
                                      \                            /
                                       \                          /
                                        f1 -- f2.1 -- f3.2 -- f4.1`
```

The advantage to the rebase is that the history is visually simple.
There are several disadvantages.  The is no simple way to reverse a bad rebase.  If there is a particularly tricky rebase conflict you may have to fix the same error in multiple commits, giving you ample opportunity to botch it in one commit and then fix it in another.  Five commits need to be verified by all of the devs involved.

default/AI/AIstate.py
LGM-Doyle: That is a good point.  Re implementing the same behavior should not break saves.

I'll fix that.

Add a totally new type of behavior will still inevitably break the save.

Morlic-fo: Not sure why that would be the case.

I would expect `create_character(aggression, id)` to return the same behavior object even if another new type of behaviour is implemented (as subclass?).
Or do you mean you want to expand future use to something like `create_character(aggression, id, id2)`. In that case you could provide a default behaviour with default (i.e. no) decision-making in case no argument is specified (as e.g. not stored in savestring).

Cjkjvfnby: No need to break saves: https://docs.python.org/2/library/pickle.html#object.__getstate__

Cjkjvfnby:
> The combiners in Character are simple.

  Not for everyone. Adding log for every decision is bad idea, because it will overflow logs.

```
Class Char:
    def foo():
        return combine(self.a.foo, self.b.foo)
    def boo()
        return combine(self.a.boo, self.b.boo)

char1 = Char(BehA, BehB)
char2 = Char(BehA, BehC)
char3 = Char(BehB, BehC)
```

```
Class Char:
    def foo():
        return self.a.foo
    def boo()
        return self.b.boo

char1 = Char(BehFooA, BehBooA)
char2 = Char(BehFooA, BehBooB)
char3 = Char(BehFooB, BehBooB)
```

Both approaches will produce same result: Character with number of functions.
I want you to agree on that point. If you don't,  I will provide bigger example with method realization.

default/AI/DiplomaticCorp.py
```diff
@@ -98,7 +92,8 @@ def evaluate_diplomatic_attitude(self, other_empire_id):
         num_peace_requests = len(foAI.foAIstate.diplomatic_logs.get('peace_requests', {}).get(log_index, []))
         num_war_declarations = len(foAI.foAIstate.diplomatic_logs.get('war_declarations', {}).get(log_index, []))
         # Too many requests for peace irritate the AI, as do any war declarations
-        irritation = (foAI.foAIstate.aggression * (2.0 + num_peace_requests / 10.0 + 2.0 * num_war_declarations) + 0.5)
+        aggression = foAI.foAIstate.character.get_behavior(Aggression)
```
Cjkjvfnby: I have 3 instances of `Behavior` here: `character`, `character.get_behavior(Aggression)`, and `character.get_behavior(EmpireIDBehavior)`.  How can I understand which one I should take?  Why we need to have any function composition if we manually choose which behavior to use?

Morlic-fo:
> Not for everyone. Adding log for every decision is bad idea, because it will overflow logs.

IMHO, for a new future like this which in its current design may become rather complex, I am definitely supporting extensive logging - but gated by a `verbose_logging` switch defaulted to False so it will/may be used for debugging / understanding purposes for devs.

default/AI/DiplomaticCorp.py
Morlic-fo: This seems weird to me, too.

Irritation level or at least modifiers should be returned by a `character.function`directly I think.

default/AI/FreeOrionAI.py
```diff
@@ -75,27 +64,30 @@ def startNewGame(aggression=fo.aggression.aggressive):  # pylint: disable=invali
...
+        aggression.key, aggression_name[foAIstate.character])
     foAIstate.session_start_cleanup()
     print "Initialized foAIstate class"
     planet_id = PlanetUtilsAI.get_capital()
     universe = fo.getUniverse()
     if planet_id is not None and planet_id != -1:
         planet = universe.getPlanet(planet_id)
-        new_name = " ".join([random.choice(_capitals.get(aggression, []) or [" "]).strip(), planet.name])
-        print "Capitol City Names are: ", _capitals
+        possible_capitals = aggression_capitals[foAIstate.character]
```
Morlic-fo: If the character defines the capitals, why not let it be handled by the character object?`character.get_possible_capitals()`

default/AI/AIstate.py
Morlic-fo: Yeah, probably should make use of these functions then.

default/AI/Behavior/Character.py
```diff
@@ -0,0 +1,492 @@
...
+        """Applies funcnamei from each behavior to the alternatives and return the most preferred."""
+        prefs = [y for y in [getattr(x, funcnamei)(alternatives) for x in self.behaviors] if y is not None]
+        if not prefs:
+            return None
+        if len(prefs) == 1:
+            return prefs[0]
+        counts = defaultdict(int)
+        for pref in prefs:
+            counts[pref] += 1
+        return max(counts)
```
Morlic-fo:
```
from collections import Counter
counts = Counter(prefs)
return Counter.most_common(1)[0][0]
```

default/AI/AIstate.py
LGM-Doyle: I will use `__getstate__`.

Currently, it only needs to store aggression, which is exactly the same as previously.  However, since I haven't used `__getstate__`, it is storing aggression twice and empire_id once.

In future I do want to expand it to something like:
    `def create_character(<initialization stuff>) {

```
     return Character([ behavior from All_Behaviors
            if behavior in Mandatory_Behaviors
            or if behavior in Mandatory_Behaviors(my_species)
            ] + [ 1 to 3 random behaviors])
}`
```

The only mandatory behavior I expect to be Difficulty, which will be refactored out of Aggression.

default/AI/DiplomaticCorp.py
LGM-Doyle: You are right.  That needs to be abstracted into a Behavioral question.  I remember looking at it and it was so specific that I couldn't think of an abstracted behavioral question that encompassed this equation.  So, I left it for later and later became never.  I will fix that.

default/AI/FreeOrionAI.py
LGM-Doyle: There will be a get_capital_name, get_name and get_greeting, however I think it should go in CharacterStrings.

I think of the break between Character and CharacterStrings as the difference between the logic which goes into Character and the UI goes into CharacterStrings.  Strings are used to convey information to the human players, about the "thoughts" of the AI.

There will be an interface to CharacterStrings similar to Character.

Since I only had the one example of a Behavior with strings I only did aggression_name, aggression_capital and aggression_greetings.

default/AI/Behavior/Character.py
LGM-Doyle: Thanks.  I appear to have reinvented the wheel.

LGM-Doyle:
PR #669 adds support to freeorion for the python standard library logger.

The logger library has built in support for arbitrary scoped loggers and many levels of logs.

Logging statements that are disabled are not even parsed, saving that overhead.

It globally tracks which loggers are enabled/disabled and you can query the loggers about their state, so no need to store additional booleans.

It is a good incremental replacement for print statements and adding more `verbose_<thing>` booleans.  This being new code is a good place to start.

LGM-Doyle:
@Cjkjvfnby Your method will produce the same results.  In fact it could produce more complex results.

Character with combiner produces a more decoupled design.  Changes in the AI modules, the Character or the Behaviors are independent of each other.  This is more maintainable and easier to extend.

I can explain this with more ASCII art.

```
`| N |              | 1 |             | L |
 |   |              |   |             |   |
 | A |              | C |             | B |
 | I |              | h |    provide  | e |
 |   | ask any of   | a |    some of  | h |
 | m |     M        | r |       M     | a |
 | o | questions -> | a | <- answers  | v |  
 | d |              | c |             | i |
 | u |              | t |             | o |
 | l |              | e |             | r |
 | e |              | r |             | s |
 | s |              |   |             |   |`
```

There are N AI modules, L Behaviors and 1 character.  They are interact through the M questions.

AI Modules

Any single AI Module only asks the subset of the M questions that it cares about.
It typically asks character questions, not individual Behaviors.  There is no guarantee that any particular behavior is present.
It is independent of the Behaviors and adding questions.  Adding/removing Behaviors or adding questions does not change the AI module.

Behavior

Each Behavior only provides answers to the subset of questions that it cares about.
It is independent of the AI modules and of other Behaviors.  Adding/removing an AI module or a new Behavior does not change any of the Behaviors.
Behaviors are decoupled enough to be tested in isolation.
The complexity of Behaviors is limited because they do not depend on other Behaviors.

Character

There is only 1 character.  It is composed of some number of Behaviors.  If that number is zero it falls through to the base class Behavior.  Character does not have one instance of each behavior, it only has the behaviors particular to this AI instance.

Adding a new question/answer to Character,  adds 1 combiner from a list of standardized combiners, but does not change any of the Behaviors or the AI modules.

Character's standardized combiners are already tested in the test code in the PR.

Part of the problem with the old implementation is that the aggression and (empire_id % 2 or 3) related code was scatter throughout the different AI modules.  The solution is to localize all of the aggression or (empire_id % 2 or 3) code in a single class.

Character with combiner provides better code localization to answer the question "What does behavior X do?".  All of the behavior X code is in a single class.

Your version is better localized to answer the question "How do all of the behaviors affect the single question/answer may_research('TECH')?".  All of that code to answer one question is in a single method of arbitrary complexity.

Calculating the answer in one method invites increases in code complexity like `EmpireIDBehavior.get_research_index()` which depends on aggression, (empire_id % 2) and (empire_id % 3).  This growth in complexity does not scale well when you increase the number of behaviors beyond two. 

Character with combiner provides better code localization to limit the growth of complexity.  The combiner is either the preference counting combiner which looks for maximum votes, or the single function combiner using one of `min()`, `max()`, `any()` or `all()`.  I think that with a debug message all 5 of those are simple enough to be understood by inspection. 

All 5 versions of the combiner code are tested in the test code in this PR.  So if it works now and the tests are correct, then it will keep working until the tests say otherwise.  And if it is broken now, then adding a new test and fixing the bug will put it back into the state working with passing tests.

Vezzra:
I recommend moving the discussion about our rebase vs merge policy to the forums. It doesn't really belong to the discussion of this PR, and is likely do be buried/overlooked here.

Cjkjvfnby:
FYI:  AI code moved, `AI -> python/AI` don't forget to move new files.

Morlic-fo:
Any plans for this?

LGM-Doyle:
Yes plans.

Once freeorion was in release mode and this was shelved I switched to working on UI improvements that would not require any balance or play testing.  I have several more that I want to clean up and push.

I had thought that the AI team was on vacation.

After that I had planned to come up with some new species specific behaviors that would be more exciting that just replacing the current behavior.

However, if you are around to ok this PR, I can rebase it soon.  Then anyone/everyone can add new behaviors.

Morlic-fo:
Yeah, no problem, it was a good call from you. Feel free to finish up any other projects you have in mind first. I was just wondering what's the status on this PR lest it is forgotten.

default/AI/AIstate.py
LGM-Doyle: I looked into using `__getstate__` and `__setstate__` but each of the two `Behaviors` do not contain any derived member data so saving and loading only save and loads aggression and empire id.  So the information in the save file currently is the essential information.

LGM-Doyle:
Rebased.

I added a commit so that the history analyzer assessment code pulls aggression from the character.

LGM-Doyle:
Rebased.

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return "Behavior"
+
+    # @abc.abstractproperty
+    @property
+    def key(self):  # pylint: disable=no-self-use,unused-argument
+        """Return an key for look up tables."""
+        return None
+
+    def may_explore_system(self, monster_threat):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to explore system given monster threat."""
```
Morlic-fo: a system with given monster threat

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return True
+
+    def may_surge_industry(self, totalPP, totalRP):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to surge industry."""
```
Morlic-fo: needs updated docstring. What is meant by surge industry? In doubt, add # TODO

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return True
+
+    def may_maximize_research(self):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to maximize research."""
+        return True
+
+    def preferred_research_cutoff(self, alternatives):  # pylint: disable=no-self-use,unused-argument
+        """Return preferred research cutoff."""
```
Morlic-fo: What are alternatives?

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return True
+
+    def invasion_priority_scaling(self):  # pylint: disable=no-self-use,unused-argument
+        return 1.0
+
+    def military_priority_scaling(self):  # pylint: disable=no-self-use,unused-argument
+        return 1.0
+
+    def preferred_colonization_portion(self, alternatives):  # pylint: disable=no-self-use,unused-argument
+        """Select a fraction less than 1"""
```
Morlic-fo: ^

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return None
+
+    def preferred_outpost_portion(self, alternatives):  # pylint: disable=no-self-use,unused-argument
+        """Select a fraction less than 1"""
```
Morlic-fo: ^

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return None
+
+    def preferred_building_ratio(self, alternatives):  # pylint: disable=no-self-use,unused-argument
+        """Select a fraction less than 1 at maximum ratio of PP for buildings"""
```
Morlic-fo: ^

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return None
+
+    def preferred_discount_multiplier(self, alternatives):  # pylint: disable=no-self-use,unused-argument
+        """Select a discount multiplier for use in evaluate planet"""
```
Morlic-fo: ^

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return None
+
+    def max_defense_portion(self):  # pylint: disable=no-self-use,unused-argument
+        """Return maximum fraction of production for defense"""
+        return 1.0
+
+    def check_orbital_production(self):  # pylint: disable=no-self-use,unused-argument
+        """Return if orbital production is checked this turn"""
```
Morlic-fo: Meaning?

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return False
+
+    def target_number_of_orbitals(self):  # pylint: disable=no-self-use,unused-argument
+        """Return target number of orbitals"""
```
Morlic-fo: orbital what?

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+    def may_build_building(self, building):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to build ''building''"""
+        return True
+
+    def may_produce_troops(self):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to produce troop ships"""
+        return True
+
+    def military_safety_factor(self):  # pylint: disable=no-self-use,unused-argument
+        """Return military safety factor"""
```
Morlic-fo: What is that?

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+    def may_research_heavily(self):  # pylint: disable=no-self-use,unused-argument
+        """Return true if allowed to target research/industry > 1.5"""
+        return True
+
+    def may_travel_beyond_supply(self, distance):  # pylint: disable=no-self-use,unused-argument
+        """Return True if able to travel distance hops beyond empire supply"""
+        return True
+
+    def get_research_index(self):  # pylint: disable=no-self-use,unused-argument
+        """Deprecated"""
```
Morlic-fo: ?

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return None
+
+    def may_research_xeno_genetics_variances(self):  # pylint: disable=no-self-use,unused-argument
+        """Return True if able to vary xeno genetics research"""
```
Morlic-fo: Meaning?

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return True
+
+    def may_research_tech(self, tech):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to research ''tech''"""
+        return True
+
+    def may_research_tech_classic(self, tech):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to research ''tech'' in the classic algorithm"""
+        return True
+
```
Morlic-fo: If you can not give more descriptive docstrings because you do not know the underlying code, that is no problem and expected. Simply refering to the place(s) where the function is invoked would be already a great help, e.g. "Toggles Bla in Research.py" is more helpful than  "Return True if able to vary xeno genetics research" with no indication to whatever that is and whereever it is used.
More descriptive docstrings can be added later, TODOs would not hurt though.

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return 0.5 if self.aggression == fo.aggression.beginner else 1.0
+
+    def military_priority_scaling(self):
+        return 1.0 if self.aggression > fo.aggression.typical else ((1.0 + self.aggression) / (1.0 + fo.aggression.typical))
+
+    def max_defense_portion(self):
+        return [0.7, 0.4, 0.3, 0.2, 0.1, 0.0][self.aggression]
+
+    def check_orbital_production(self):
+        aggression_index = max(1, self.aggression)
```
Morlic-fo: arguably deserves its own function as it is used more than once and is non-trivial.

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+        return ((fo.currentTurn() % aggression_index) == 0) and self.aggression < fo.aggression.maniacal
+
+    def target_number_of_orbitals(self):
+        aggression_index = max(1, self.aggression)
+        return min(int(((fo.currentTurn() + 4) / (8.0 * aggression_index**1.5))**0.8), fo.aggression.maniacal - aggression_index)
+
+    building_table_static = {"BLD_SHIPYARD_AST": fo.aggression.beginner,
```
Morlic-fo: `BUILDING_TABLE` would be more in line with rest of the codebase.

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+                             "BLD_SHIPYARD_AST_REF": fo.aggression.maniacal,
+                             "BLD_SHIPYARD_ORG_CELL_GRO_CHAMB": fo.aggression.aggressive,
+                             "BLD_SHIPYARD_ORG_XENO_FAC": fo.aggression.aggressive,
+                             "BLD_SHIPYARD_ENRG_COMP": fo.aggression.aggressive,
+                             "BLD_SHIPYARD_ENRG_SOLAR": fo.aggression.maniacal,
+                             "BLD_NEUTRONIUM_FORGE": fo.aggression.cautious}
+
+    @property
+    def building_table(self):
+        return type(self).building_table_static
```
Morlic-fo: why?

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+    def may_research_xeno_genetics_variances(self):
+        return self.aggression >= fo.aggression.cautious
+
+    def prefer_research_defensive(self):
+        return self.aggression <= fo.aggression.cautious
+
+    def prefer_research_low_aggression(self):
+        return self.aggression < fo.aggression.typical
+
+    tech_lower_threshold_static = {"LRN_TRANSCEND": fo.aggression.aggressive}
```
Morlic-fo: as before

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+
+    def may_research_tech(self, tech):
+        return type(self).tech_lower_threshold_static.get(tech, fo.aggression.beginner) <= self.aggression
+
+    tech_upper_threshold_classic_static = {"DEF_DEFENSE_NET_1": fo.aggression.cautious,
+                                           "DEF_GARRISON_1": fo.aggression.cautious,
+                                           "GRO_XENO_GENETICS": fo.aggression.cautious,
+                                           "GRO_GENETIC_ENG": fo.aggression.cautious}
+
+    tech_lower_threshold_classic_static = {"SHP_DEFLECTOR_SHIELD": fo.aggression.aggressive,
```
Morlic-fo: as before

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+
+    def preferred_research_cutoff(self, alternatives):
+        if not alternatives:
+            return None
+        return alternatives[self.id % 2] if len(alternatives) >= 2 else alternatives[0]
+
+    def preferred_colonization_portion(self, alternatives):
+        if not alternatives:
+            return None
+        return alternatives[self.id % 2] if len(alternatives) >= 2 else alternatives[0]
```
Morlic-fo: If the original intent is that there are n alternatives to choose from, then simply use

```
try:
    return alternatives[self.id % n]
catch IndexError as e:
    print_error(e)
    return None
```

which may be implemented as private class method.

Requires the additional import
`from freeorion_tools import print_error`

Morlic-fo (changes requested):

LGM-Doyle:
@Morlic-fo thanks for the review.  I will work on these changes.

default/python/AI/AIstate.py
```diff
@@ -93,6 +94,37 @@ def __init__(self, aggression):
...
+            if not fo.getOptionsDBOptionBool(force_option):
+                return default
+
+            per_id_option = "AI.config.behavior.%s.%s" % (name, fo.playerName())
+            all_id_option = "AI.config.behavior.%s.all" % (name,)
+
+            behavior = fo.getOptionsDBOptionInt(per_id_option)
+            if behavior is None or behavior == sentinel:
+                behavior = fo.getOptionsDBOptionInt(all_id_option)
+            # print 'NOT picv = ', per_id_option , ' bev = ', behavior
```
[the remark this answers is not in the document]
LGM-Doyle: No. Removed.

default/python/AI/AIstate.py
```diff
@@ -93,6 +94,37 @@ def __init__(self, aggression):
         self.empire_standard_enemy = (4, ((4, 1),), 0.0, 10.0)  # TODO: track on a per-empire basis
         self.empire_standard_enemy_rating = 40  # TODO: track on a per-empire basis
 
+        # Check the optionsDB for the behavior bypass values and create
+        # the character.
+        def get_behavior_bypass_value(name, default, sentinel):
+            """Fetch the bypassed behavior value or return the default"""
```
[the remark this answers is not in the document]
LGM-Doyle: done.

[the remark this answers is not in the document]
LGM-Doyle: done

default/python/AI/AIstate.py
```diff
@@ -93,6 +94,37 @@ def __init__(self, aggression):
...
+
+            per_id_option = "AI.config.behavior.%s.%s" % (name, fo.playerName())
+            all_id_option = "AI.config.behavior.%s.all" % (name,)
+
+            behavior = fo.getOptionsDBOptionInt(per_id_option)
+            if behavior is None or behavior == sentinel:
+                behavior = fo.getOptionsDBOptionInt(all_id_option)
+            # print 'NOT picv = ', per_id_option , ' bev = ', behavior
+            if behavior is None or behavior == sentinel:
+                print ("Bypass of %s behavior failed. Returning default value %s." % (name, str(default)))
```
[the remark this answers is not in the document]
LGM-Doyle: fixed.

default/python/AI/AIstate.py
LGM-Doyle: I moved this function definition from within AIState to be a free function in Character.py.  Its definition in AIState was misplaced as it only currently deals with bypassing character behaviors.

default/python/AI/AIstate.py
```diff
@@ -93,6 +94,37 @@ def __init__(self, aggression):
...
+                print "To force %s behavior set %s to 1" % (name, force_option,)
+                print "Set %s to a value to force just %s's behavior." % (per_id_option, fo.playerName())
+                print "Set  %s to a value to force the behavior of all AIs not individually specified." % (all_id_option,)
+                behavior = default
```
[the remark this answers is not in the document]
LGM-Doyle: None of the cases are errors.  If the config value is missing then the dev (presumably) does not want to hard code the AI behavior in the config.xml file.  The print statements were only informative.  I moved the information into the docstring of the function.  Anyone hacking on the AI Character code will hopefully find it there.

default/python/AI/AIstate.py
```diff
@@ -93,6 +94,37 @@ def __init__(self, aggression):
...
+            print "%s behavior bypassed and set to %s for %s" % (name, repr(behavior), fo.playerName())
+            return behavior
+
+        no_value = -1
```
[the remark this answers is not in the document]
LGM-Doyle: fixed

default/python/AI/AIstate.py
```diff
@@ -93,6 +94,37 @@ def __init__(self, aggression):
...
+        _behavior_aggression = get_behavior_bypass_value("aggression", aggression, no_value)
+        _behavior_id = get_behavior_bypass_value("empire-id", self.empireID, no_value)
+
+        self.character = create_character(_behavior_aggression, _behavior_id)
+
```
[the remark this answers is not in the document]
LGM-Doyle: I moved the bypass functionality to be a free function in Character.py.

I created `create_character(aggression, empire_id)`.  I kept aggression and empire_id as parameters to preserve a logical flow of control.  Aggression and empire id are inputs to AIState, which then sets up Character using those inputs.

default/python/AI/Behavior/Character.py
```diff
@@ -0,0 +1,517 @@
...
+"""
+
+import abc
+from collections import defaultdict, Counter
+import random
+
+import freeOrionAIInterface as fo  # pylint: disable=import-error
+
+
+class Behavior(object):
```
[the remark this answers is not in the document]
LGM-Doyle: Do you mean to replace `Behavior` everywhere with `Trait`?

I'm OK with that, but I want to double check before I change it everywhere.

LGM-Doyle (commented):
I've made all of the changes as noted, except changing Behavior to Trait.  As mentioned I'm fine with that change, but I wanted to check before I did the work and found out I misunderstood your intent.

LGM-Doyle:
@Morlic-fo  I added an additional tap `warship_adjusted_production_cost_exponent()` to `Behavior `which pulls in the aggression based exponent calculation in the warship costs that are being used in the `Fighters` branch.  This seemed like the easiest way to make this simple to merge with the `Fighters` branch to minimize work for people who are regularly rebasing `Fighters` onto master, if this PR is merged.

Morlic-fo:
Yes, I meant changing ```Behavior``` into ```Trait``` everywhere unless you have a better idea of a name. Trait is arguably kind of an awkward name when using the empireID for decision-making. That one should in the long-term be removed for many decisions and replaced by more appropriate Traits.

LGM-Doyle:
`Character` `Trait`  works for me.  I'd update the PR in a few days.

LGM-Doyle:
Renamed everywhere to `Trait` and rebased.

LGM-Doyle:
Rebased.

default/python/AI/PriorityAI.py
```diff
@@ -281,19 +279,19 @@ def _calculate_outpost_priority():
...
 
-    not_sparse, enemy_unseen = 0, 0
-    is_sparse, enemy_seen = 1, 1
-    allotted_portion = {
-        (not_sparse, enemy_unseen): (0.6, 0.8),
-        (not_sparse, enemy_seen): (0.3, 0.4),
-        (is_sparse, enemy_unseen): (0.8, 0.9),
-        (is_sparse, enemy_seen): (0.3, 0.4),
-    }[(galaxy_is_sparse, any(enemies_sighted))][fo.empireID() % 2]
+    not_sparce, enemy_unseen = 0, 0
```
Morlic-fo: Any reason to rename this or typo? ```not_sparse``` seems correct.

default/python/AI/PriorityAI.py
```diff
@@ -281,19 +279,19 @@ def _calculate_outpost_priority():
...
+    is_sparce, enemy_seen = 1, 1
+    allotted_portion = {(not_sparce, enemy_unseen): (0.6, 0.8),
+                        (not_sparce, enemy_seen): (0.3, 0.4),
+                        (is_sparce, enemy_unseen): (0.8, 0.9),
+                        (is_sparce, enemy_seen): (0.3, 0.4), }[
+        (galaxy_is_sparse, any(enemies_sighted))]
```
Morlic-fo: The way the dict-definition was formatted previously is the style that is mostly used throughout the codebase and thus preferred, i.e.
```
d = {
    k: v,
    k: v,
}[indexing]
```

Edit: fix space

default/python/AI/ResearchAI.py
```diff
@@ -952,7 +947,7 @@ def generate_classic_research_orders():
     #
     # assess if our empire has any non-lousy colonizers, & boost gro_xeno_gen if we don't
     if True:  # just to help with cold-folding / organization
-        if got_ggg_tech and got_sym_bio and (not got_xeno_gen) and foAI.foAIstate.aggression >= fo.aggression.cautious:
+        if got_ggg_tech and got_sym_bio and (not got_xeno_gen):
```
Cjkjvfnby, line 950: Any reason why you remove this aggression check?

And same for other such checks.

default/python/AI/Trait/Character.py
```diff
@@ -0,0 +1,687 @@
...
+        return True
+
+    def may_research_tech_classic(self, tech):  # pylint: disable=no-self-use,unused-argument
+        """Return True if permitted to research ''tech''.  This is called in the classic research algorithm."""
+        # TODO remove this tap when the classic research algorithm is removed.
+        return True
+
+    def attitude_to_empire(self, other_empire_id, diplomatic_logs):  # pylint: disable=no-self-use,unused-argument
+        """Return a value from [-10 .. 10] representing attitude towards other empire."""
+        return None
```
Morlic-fo: Maybe return 0 as neutral default value?

default/python/AI/Trait/Character.py
```diff
@@ -0,0 +1,687 @@
...
+
+    def warship_adjusted_production_cost_exponent(self):  # pylint: disable=no-self-use,unused-argument
+        """Return an exponent to scale the production cost of a warship in ShipDesignAI."""
+        return None
```
Morlic-fo: Maybe return 1 as neutral default value?

default/python/AI/Trait/Character.py
```diff
@@ -0,0 +1,687 @@
...
+        if self.aggression >= fo.aggression.aggressive:
+            research_index = 2 + (self.id % 3)  # so indices [2,3,4]
+        elif self.aggression >= fo.aggression.typical:
+            research_index += 1
+        return research_index
+
+
+class Character(Trait):
+    """
+    A collection of behaviours.
```
Morlic-fo: behaviours -> traits

default/python/AI/Trait/Character.py
```diff
@@ -0,0 +1,687 @@
...
+            raise TypeError("All traits must be sub-classes of Trait")
+
+    def get_trait(self, type_of_trait):
+        """Return the requested trait or None"""
+        trait = [x for x in self.traits if isinstance(x, type_of_trait)]
+        return trait[0] if trait else Trait()
+
+
+# Complete the Character class by adding all of the combiners to combine the outputs of the
+# individual traits.  Character tries to combine results in the way most limiting to the AI
```
Morlic-fo: Either define the functions within the class or add a short comment there indicating that the class definition is incomplete.

default/python/AI/Trait/Character.py
```diff
@@ -0,0 +1,687 @@
...
+
+# Create combiners for traits that take any result
+for funcname in ["check_orbital_production"]:
+    setattr(Character, funcname, _make_single_function_combiner(funcname, any))
+
+
+# Create combiners for traits that averages all not None results
+def average_not_none(llin):
+    ll = [x for x in llin if x]
+    return sum(ll) / len(ll)
```
Morlic-fo: Check for None explicitly. The list may contain zeros.

E.g. average of (0, 1) is 0.5 but your function returns 1.

default/python/AI/AIstate.py
```diff
@@ -15,6 +15,7 @@
 import PlanetUtilsAI
 from freeorion_tools import dict_from_map
 from universe_object import System
+from Trait.Character import create_character
```
Morlic-fo: ```Trait.Character``` feels awkward, The package should probably be named Character.

default/python/AI/freeorion_tools/charts_handler.py
```diff
@@ -15,14 +15,14 @@ def charting_text():
     planet = None
     if planet_id is not None:
         planet = universe.getPlanet(planet_id)
-    aggression_name = fo.aggression.values[foAI.foAIstate.aggression].name
+    aggression_name_local = aggression_name[foAI.foAIstate.character]
```
Cjkjvfnby: As developer I prefer to see something like `foAI.foAIstate.character.name`

default/python/AI/FreeOrionAI.py
```diff
@@ -83,29 +68,31 @@ def startNewGame(aggression=fo.aggression.aggressive):  # pylint: disable=invali
...
 
     turn_timer.start("Server Processing")
-    print "New game started, AI Aggression level %d (%s)" % (aggression, UserString(_aggression_names[aggression]))
 
     # initialize AIstate
     global foAIstate
     print "Initializing foAIstate..."
-    foAIstate = AIstate.AIstate(aggression)
+    foAIstate = AIstate.AIstate(_aggression)
+    aggression = foAIstate.character.get_trait(Aggression)
```
Morlic-fo: Using all 3 ```_aggression, aggression, Aggression``` is kind of unclear. 

Maybe use ```aggression``` as parameter and ```aggression_trait``` for the second variable.

default/python/AI/FreeOrionAI.py
```diff
@@ -83,29 +68,31 @@ def startNewGame(aggression=fo.aggression.aggressive):  # pylint: disable=invali
...
+        new_name = " ".join([random.choice(possible_capitals).strip(), planet.name])
         print "    Renaming to %s..." % new_name
         res = fo.issueRenameOrder(planet_id, new_name)
         print "    Result: %d; Planet is now named %s" % (res, planet.name)
 
     diplomatic_corp_configs = {fo.aggression.beginner: DiplomaticCorp.BeginnerDiplomaticCorp,
                                fo.aggression.maniacal: DiplomaticCorp.ManiacalDiplomaticCorp}
     global diplomatic_corp
-    diplomatic_corp = diplomatic_corp_configs.get(aggression, DiplomaticCorp.DiplomaticCorp)()
-    TechsListsAI.test_tech_integrity()
```
Morlic-fo, line 109: Why is this test removed here?

Cjkjvfnby:
I have felling that Traits are missing.

Also see this comment: https://github.com/freeorion/freeorion/pull/677#issuecomment-222333619

Morlic-fo (changes requested):
Seems fine now for the most part.

default/python/AI/PriorityAI.py
Morlic-fo: Same for ```is_sparse```

default/python/AI/PriorityAI.py
Cjkjvfnby: space: `k: v`

LGM-Doyle (commented):
@Cjkjvfnby  and @Morlic-fo  thanks for the prompt and thorough review.  I've made changes or explained why I chose this implementation for all of your comments.

I've renamed the modules and packages as requested.

default/python/AI/Character/Character.py
```diff
@@ -567,15 +567,17 @@ def func(self, *args, **kwargs):
 
 # Create combiners for traits that averages all not None results
 def average_not_none(llin):
-    ll = [x for x in llin if x]
+    ll = [x for x in llin if x != None]
```
Morlic-fo: PEP8: ```x is not None```

default/python/AI/Character/Character.py
```diff
@@ -567,15 +567,17 @@ def func(self, *args, **kwargs):
...
+        return 0
     return sum(ll) / len(ll)
 
 for funcname in ["attitude_to_empire"]:
     setattr(Character, funcname, _make_single_function_combiner(funcname, average_not_none))
 
 # Create combiners for traits that use the geometic mean of all not None results
 def geometric_mean_not_none(llin):
-    ll = [x for x in llin if x]
+    ll = [x for x in llin if x != None]
```
Morlic-fo: PEP8: ```x is not None```

default/python/AI/Character/Character.py
```diff
@@ -567,15 +567,17 @@ def func(self, *args, **kwargs):
 
 # Create combiners for traits that averages all not None results
 def average_not_none(llin):
-    ll = [x for x in llin if x]
+    ll = [x for x in llin if x != None]
+    if not ll:
+        return 0
     return sum(ll) / len(ll)
```
Morlic-fo: Needs explicit float conversion

default/python/AI/Character/Character.py
```diff
@@ -567,15 +567,17 @@ def func(self, *args, **kwargs):
...
 for funcname in ["attitude_to_empire"]:
     setattr(Character, funcname, _make_single_function_combiner(funcname, average_not_none))
 
 # Create combiners for traits that use the geometic mean of all not None results
 def geometric_mean_not_none(llin):
-    ll = [x for x in llin if x]
+    ll = [x for x in llin if x != None]
     if not ll:
         return 1
     return math.exp( sum(map(math.log, ll)) / len(ll))
```
Morlic-fo: Please add a sanity check if input values are greater than zero. Filter and log an error if not.
Also not sure if math.log ensures that a float number is returned, might need explicit conversion.

default/python/AI/Character/Character.py
```diff
@@ -569,9 +570,18 @@ def average_not_none(llin):
...
-for funcname in ["attitude_to_empire", "warship_adjusted_production_cost_exponent"]:
+for funcname in ["attitude_to_empire"]:
     setattr(Character, funcname, _make_single_function_combiner(funcname, average_not_none))
 
+# Create combiners for traits that use the geometic mean of all not None results
+def geometric_mean_not_none(llin):
+    ll = [x for x in llin if x]
+    if not ll:
+        return 1
+    return math.exp( sum(map(math.log, ll)) / len(ll))
```
Morlic-fo: remove excess whitespace behind first (

LGM-Doyle:
I've fixed the geometric combiner.

default/python/AI/character/character_module.py
```diff
@@ -567,20 +568,24 @@ def func(self, *args, **kwargs):
...
+    return sum(ll) / float(len(ll))
 
 for funcname in ["attitude_to_empire"]:
     setattr(Character, funcname, _make_single_function_combiner(funcname, average_not_none))
 
 # Create combiners for traits that use the geometic mean of all not None results
 def geometric_mean_not_none(llin):
-    ll = [x for x in llin if x != None]
+    ll_not_none = [x for x in llin if x is not None]
+    ll = [x for x in ll_not_none if x >= 0]
```
Morlic-fo: ```x > 0``` as log(0) is undefined.

LGM-Doyle: Fixed again.

Morlic-fo (approved):

default/python/AI/Trait/CharacterStrings.py
```diff
@@ -0,0 +1,51 @@
...
+
+_aggression_label_suffix = {fo.aggression.beginner: "_BEGINNER",
+                            fo.aggression.turtle: "_TURTLE",
+                            fo.aggression.cautious: "_CAUTIOUS",
+                            fo.aggression.typical: "_TYPICAL",
+                            fo.aggression.aggressive: "_AGGRESSIVE",
+                            fo.aggression.maniacal: "_MANIACAL"}
+
+
+def make_aggression_table(prefix, post_process_func=None):
```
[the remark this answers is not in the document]
LGM-Doyle: I changed the function name to  `get_trait_name_aggression(character)` .  I made functions not used outside of `character_strings_module.py`, except in tests, private.

default/python/AI/freeorion_tools/charts_handler.py
LGM-Doyle: 1. I will gladly reword it, because I can only write/understand a few rudimentary sentences in languages other than English.
2. The correct expression is "forced idiom".  I mistyped. Idiom is the style of way of expressing an idea in a language.
3. Perhaps we should do a day in each devs mother tongue to put the English in their place. 😃 

My style for this code was unnatural,  inelegant and awkward.  Using something that looks like an array `table[character]` when `character` is not obviously an index is the wrong approach.  As you pointed out a function is the more natural choice.

I switched to a public function  `get_trait_name_aggression(character)` and expanded the module comments.

LGM-Doyle:
I have:
- renamed some functions in `character_strings_module.py` to make them private.
- renamed `trait_name_aggression()` to `get_trait_name_aggression()`
- **fixed** `Character.get_trait()` to return `None` when a character does not have a trait because that is the documented behavior and it was actually returning an blank `Trait()`
- expanded the comments in `character_strings_module.py`
- added a test for `get_trait_name_aggression()`

default/python/AI/freeorion_tools/charts_handler.py
Vezzra: > Perhaps we should do a day in each devs mother tongue to put the English in their place.

Brilliant idea - @Cjkjvfnby surely is going to have a blast with us trying to communicate in Russian... 😂

Morlic-fo:
@Cjkjvfnby: I know you are busy at the moment. Could you give a short statement if you think all your concerns have been adressed and the PR is generally good to merge or if you need to review it in more detail once you have time again?

Cjkjvfnby:
@Morlic-fo,  lets merge it.

Vezzra:
Just to be sure it doesn't get overlooked once this gets merged, because the discussion thread has become quite long: @LGM-Doyle requested to use this text in the merge commit message:

*************** Merge Message Cut Here ***********************************
Create AI Character composed of Behaviors to encompass variation
between AIs.

Character solves the problem of having AI personalization added
piecemeal throughout the AI code. Now each type of personalization is
encapsulated in a single Behavior. The Behaviors are combined to form a
single unique Character per AI.

Refactor current variation in AI into two Behaviors, Aggression and
EmpireIDBehavior to encompass AI variation due to the aggression setting
and the empire id.

Add debugging bypass to randomized Character and Behavior.

The bypass can force individual AIs to have a specific
behavior and/or force all other AIs to have the same default behavior.
********************* End of Merge Message ****************************

Alternatively, let @LGM-Doyle do the merge himself, then he can put in whatever merge commit messge he deems fitting.

Morlic-fo:
@LGM-Doyle Needs a final rebase, think we are good to merge then.

LGM-Doyle:
I've rebased it, updated the comment and merged it.

Morlic-fo:
Thanks. Do you plan on adding some AI traits yourself?

LGM-Doyle:
If you are asking because you want to add/implement some Traits yourself, then please do.  

I'll slowly work my way through the items on the TODO list in character_module.py, which don't require a functioning diplomacy/treaty system.

But first I'll play some games and see what jumps out at me.

Morlic-fo:
Great.

I don't have any immediate plans to add traits myself as I have multiple other half-finished stuff I want to put in first. I would have somewhat prioritized it in my todo list if you did not want to implement anything as I don't want the feature to basically lie around for a year to be forgotten and not used.

I might add some traits when getting good ideas doing other stuff though.

PR #663, opened 2016-05-14
fix terraform queue

ghost:
Don't allow multiple terraforming to be queued, or on worlds with Gaiaform.

PR #678, opened 2016-05-27
Fix terraform queue

ghost:
Recreation of pull request #663 with requested changes.
Also added exclusions for Artificial Planet.


MatGB:
Works for me, thanks for this.

Guarantee at some point we'll have someone complaining about being unable to enqueue their gaia transformation because there's something else there, but, y'know.

PR #684, opened 2016-05-30
Fix bugs where TELEPATHIC_DETECTION downgrades visibilty of planets.

LGM-Doyle:
TELEPATHIC_DETECTION had two bugs.

First it hid the species with telepathic detection's own capital planet
on the first turn.

Secondly, it prevented normal visibility rules from applying on the
first turn.

This fixes the problem by not applying to its own planets and by
delaying telepathic detection by one turn.

### DIFF
--- a/default/scripting/species/common/telepathic.macros
+++ b/default/scripting/species/common/telepathic.macros
@@ -4,9 +4,11 @@ TELEPATHIC_DETECTION
             scope = And [
                 PopulationCenter
                 Not Population high = 0
-                WithinStarlaneJumps jumps = @1@ condition = Source
                 Not VisibleToEmpire empire = Source.Owner
+                Not OwnedBy empire = Source.Owner
+                WithinStarlaneJumps jumps = @1@ condition = Source
             ]
-            activation = OwnedBy affiliation = AnyEmpire
+            // Activate the turn after source creation to allow regular vision to prevail
+            activation = Turn low = Source.CreationTurn + 1
             effects = SetVisibility visibility = Basic empire = Source.Owner
 '''
--- a/default/scripting/species/common/telepathic.macros
+++ b/default/scripting/species/common/telepathic.macros
@@ -3,12 +3,12 @@ TELEPATHIC_DETECTION
             description = "TELEPATHIC_DETECTION"
             scope = And [
                 PopulationCenter
-                Not Population high = 0
                 Not VisibleToEmpire empire = Source.Owner
+                Not Source
                 Not OwnedBy empire = Source.Owner
+                Not Population high = 0
                 WithinStarlaneJumps jumps = @1@ condition = Source
             ]
-            // Activate the turn after source creation to allow regular vision to prevail
-            activation = Turn low = Source.CreationTurn + 1
+            activation = OwnedBy affiliation = AnyEmpire
             effects = SetVisibility visibility = Basic empire = Source.Owner
 '''


default/scripting/species/common/telepathic.macros
```diff
@@ -4,9 +4,11 @@ TELEPATHIC_DETECTION
...
                 PopulationCenter
                 Not Population high = 0
-                WithinStarlaneJumps jumps = @1@ condition = Source
                 Not VisibleToEmpire empire = Source.Owner
+                Not OwnedBy empire = Source.Owner
+                WithinStarlaneJumps jumps = @1@ condition = Source
             ]
-            activation = OwnedBy affiliation = AnyEmpire
+            // Activate the turn after source creation to allow regular vision to prevail
+            activation = Turn low = Source.CreationTurn + 1
```
geoffthemedio: What happens if a planet has a telepathic species added to it by an effect and has its ownership set on the same turn so that an empire should get vision of a planet for the first time on the same turn it has that species, but it's not the planet's creation turn?

LGM-Doyle:
Welcome back.

I have not tested that case and I'm not very familiar with FOCS, so I do not know.

I think that the planet will be set to basic vision in that case and they will not be able to use the planet.  That was what was happening to the Trith.  

I did not go through the Effects code to see if the problem was that the condition on this effect was processed before owner ship of the planet was transferred to the Trith, or if it was processed before visibility of the planet was calculated or both.  The result was that the Trith start out with a planet set to basic visibility. 

If you know of a FOCS Condition that means first ownership turn + 1, that would be better than 
Source.CreationTurn +1.

LGM-Doyle:
It is meant to be like a brain, quirky.  But it is also meant to be easy to debug.

I agree that the combiner Character has no log output and it needs it.  The logs should always be clear exactly which behavior resulted in a denial.  You are right, you should never be figuring out which behavior is dominant when the Character can just log it.

The behavior methods are intended to be of two flavors.  Permission type may_<activity>() permits a certain activity. Preference type preferred_<activity>(alternatives) selects the most preferred of the alternatives.  

The combiners in Character are simple.

The may_<activity>() combiner forbids activity if any behavior forbids the activity.  The log could be like this for a rejected call to character.may_research(LRN_TRANSCEND) :

    `Character forbids researching LRN_TRANSCEND because
          permits: EmpireIDBehavior, SomeOtherFutureBehaviorNotYetImplemented
          forbids: Aggression`

The specifics for why Aggression forbids would appear in the Aggression's own logging code presumably in the previous line.

The preferred_<activity>(alternatives) combiner just counts up votes.  Currently all of the examples of preferred_ are in the EmpireIDBehavior which I think should be removed.  I will use a hypothetical example for the log.  A log for a request to preferred_invasion_target([planet1, planet2, planet3]) could look like:

    `Character prefers attacking planet2 because
    Alternatives    planet1     planet2     planet3
    Votes              1            3          0
                                  Nemesis
                    Loyal 
                                 Aggression
                                 Genocidal
    Abstentions: Chattiness`

Similarly, the individual behaviors will have logged their calculations previously in the log.

I will add logging as described above to the PR.



I will illustrate why behavior needs to be common across the modules.

If we are going to implement behavior for each module individually, then it should be left in the specific module's code as it is today.  If a choice is specific enough that it is only needed in one module and is only dependent on one aspect of behavior then it should not be in behavior.  You can retrieve a single behavior from Character with get_behavior, for example character.get_behavior(Aggression) for Aggression.

The current code is simply a refactoring of all the aggression and (empire id % n) conditional logic from throughout the AI into a single location.  That was messy and this still is messy.  There are too many methods that ask almost the exact same question of the behaviors.  There were some cases where the previous code was just inconsistent.

Many of the current method are too specific and tied directly to arbitrary coefficients from the original module.  A good example is may explore system, which is based on a 'monster_threat' threshold which is only defined in one module.  These need to be refactored to ask more generic useful questions.

Others like may_research() and may_research_classic duplicate() functionality with slightly different outputs.  They will need be merged into a single method.

A good example of generic behavior method that are useful across many modules are: may_outpost(), may_colonize(), and may_invade().  Obviously, research, ship designer, military planner and colonization AI all benefit from a uniform answer to those questions.

The next refactor should move towards all the behavior methods being good methods and refactoring the AI modules to use the more generic results appropriately.

Each Behavior should be moved to its own file.

Since each Behavior is derived from Behavior it only needs to re-implement the methods it cares about.  It should be easy for a dev to look over a method and figure out what say Aggression affects.

Since the combiner implements all of the methods it is a simple single calling point for all of the AI modules.

geoffthemedio:
Are you sure you want to remove the ownership by any empire requirement from the activation? Without that, a native Trith planet will cause the xenophobic effect even when not owned by any player...

Regardless, this seems to work ok for me for the case of starting with Trith:

```
TELEPATHIC_DETECTION
'''     EffectsGroup
            description = "TELEPATHIC_DETECTION"
            scope = And [
                PopulationCenter
                Not VisibleToEmpire empire = Source.Owner
                Not Source
                Not OwnedBy empire = Source.Owner
                Not Population high = 0
                WithinStarlaneJumps jumps = @1@ condition = Source
            ]
            activation = OwnedBy affiliation = AnyEmpire
            effects = SetVisibility visibility = Basic empire = Source.Owner
'''
```

MatGB:
Currently you can't have native Trith, Playable/Native is an exclusive toggle—I've thought it might be worth changing that, and allowing species not selected by an AI to then be native, but we don't currently do that.

Having said that, I think if there were native xenophobes they should cause the disruption, it's meant to be an annoyance.

LGM-Doyle:
Your changes work for me.

I pushed the changes so you can decide whether to merge this PR or your own commit.  You fixed the bug after all.

geoffthemedio:
pushed it myself to avoid needing to rebase / squash this

LGM-Doyle:
In the future in cases like this with changes I consider to be trivial, should I just make an unannounced rebase and squash?  What is the freeorion policy?  What is easiest for the devs?

geoffthemedio:
For your own branch which is just used for one pull request, as far as I know you can rebase whenever you feel like it, and perhaps add a comment to the pull request after donig so. If it's a shared dev branch, then be more worried. adrian broher might have other views / advice / preference, though.

Vezzra:
As @geoffthemedio said. As far as your own branches are concerned, you can squash, rebase etc. to your hearts desire, accompanying comments are appreciated of course.

Shared dev branches are a very different thing, but we only had (have) one so far (the Fighters branch). The policy for this branch is rebase only after announcement (so every involved dev can push their changes, clean up their local copies etc. to avoid things getting messed up).

PR #685, opened 2016-05-30
Remove several vacuous try..catch from UniverseWrapper

LGM-Doyle:
Removing these empty catch statements made fixing PR #684 possible.

### DIFF
--- a/client/human/HumanClientFSM.cpp
+++ b/client/human/HumanClientFSM.cpp
@@ -125,10 +125,11 @@ boost::statechart::result WaitingForSPHostAck::react(const Error& msg) {
     // MessageBox blocks and can allow other events to transit<> to a new state
     // which makes this transit fatal.
     boost::statechart::result retval = fatal ? transit<IntroMenu>() : discard_event();
-    ClientUI::MessageBox(UserString(problem), true);
 
-    if (fatal)
+    if (fatal) {
+        ClientUI::MessageBox(UserString(problem), true);
         Client().GetClientUI()->GetMessageWnd()->HandleGameStatusUpdate(UserString("RETURN_TO_INTRO") + "\n");
+    }
 
     return retval;
 }
@@ -165,10 +166,11 @@ boost::statechart::result WaitingForMPHostAck::react(const Error& msg) {
     // MessageBox blocks and can allow other events to transit<> to a new state
     // which makes this transit fatal.
     boost::statechart::result retval = fatal ? transit<IntroMenu>() : discard_event();
-    ClientUI::MessageBox(UserString(problem), true);
 
-    if (fatal)
+    if (fatal) {
+        ClientUI::MessageBox(UserString(problem), true);
         Client().GetClientUI()->GetMessageWnd()->HandleGameStatusUpdate(UserString("RETURN_TO_INTRO") + "\n");
+    }
 
     return retval;
 }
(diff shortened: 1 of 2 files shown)


geoffthemedio:
If this is necessary for fixing something else, why isn't it included in that pull request?

What specifically is the problem with empty catch blocks? They are intended to prevent crashes when attempts are made to find distances between things that don't have defined distances between them, which tends to result in weird output from the internally-used boost graph functions.

LGM-Doyle:
It was necessary to find where the error was for #684.  Because the Trith could not see their own homeworld they used -1 as a system id, which did reach all the way to the boost::graph functions which threw when trying to convert the system_id to a graph index.  However, these catch all clauses prevent the AI from generating any error.

Removing these catch clauses did not crash the AI, but the AI did produce a line number error.  I could not have found the error with these in place because nothing reported the error.

I'll repeat that this falling through to python did not crash the game.  It even reported the correct exception.

I also have a commit with more expansive debug message when get_capital() fails to find a homeworld, but I didn't bother submitting those.  Do you want that?

geoffthemedio:
That's all fine, but the problem is that these functions are not just used in the AI... Calls to effects can cause some of them to run, and in that case they need to not cause crashes.

LGM-Doyle:
I see what you mean.

I went and checked `create_universe()` and `execute_turn_events()`.  By adding `raise "a"` to each of them I was able to determine that neither of them crash the application.  They do produce an ugly popup with a message like "Python scripted turn events executed with errors. ...".

I removed MessageBox popup for non fatal errors.  The error messages still shows up in the logs, with line number information.  Before we were effectively masking the errors by not throwing, so the player has lost nothing.

Fatal errors still produce a popup.

This solution yields useful error information for the devs, and quiet running for players.

geoffthemedio:
There should still be player notifications in the event of non-fatal errors. That change is unrelated this this discussion.

You seem to be only thinking about what this does from calls within Python scripts. These functions can (I think) also be called when effects / conditions are executed / evaluated, or possibly when doing fleet routing / movement. Problems might occur when the universe connectivity changes due to effects adding / remove starlanes or systems.

LGM-Doyle:
These catch statements are in python/UniverseWrapper.cpp.  It is the python wrapper and I think is only called from python code.  

C++ code accesses Universe directly.

geoffthemedio:
Ah, ok then. Needs rebase, though...

geoffthemedio:
Bump... rebase needed.

LGM-Doyle:
I hear your bump.

I want to do a debugging pass and look for any weirdness that is exposed by removing these catches.  There won't any crashes, but if python is passing garbage in and never getting an error, this change may uncover other problems.  I don't want this change to be perceived as the source of problems, even if they are long-standing.

I let this slide so that I could work on the combat logger which is probably more likely to hold up the release.

Vezzra:
@LGM-Doyle, status on this PR? Still waiting for something else that needs to be addressed first, or is it postponed until further notice, or should it be closed?

Specifically, should it stay tagged for the release?

LGM-Doyle:
The PR should not be closed.

The PR will produce better error messages.

In that vein it has exposed some areas of the AI code that are passing bad system ids, specifically at least the exploratory module.  I was planning on fixing that and any others that I can find before I push a new version of this PR.

That being said, merging this PR will give me more eyes on collecting error messages.  It will not crash anything, only have the python AI catch the exceptions from the pathfinding functions when the AI passes bad system IDs.

It is only waiting for me spending a day running games with AIs to generate error message, finding them and fixing them.  (i.e. not getting distracted and just playing)

This is not critical for the release, but it should go if possible since it doesn't introduce and new behavior and it will improve error handling.

geoffthemedio:
Why do you need to fix errors this PR would reveal before pushing it?

LGM-Doyle:
Only so that I don't get blamed for causing errors that this PR reveals.

geoffthemedio:
I think if you note that several errors are un-hidden by the change in the description, you can avoid "blame", even if someone traces the appearance of them to your pull request.

LGM-Doyle:
I rebase it shortly, but not today.

Vezzra:
@LGM-Doyle, take your time, I just wanted make sure this PR didn't get forgotten 😉

LGM-Doyle:
Please use this as the merge message 

--------------------------------------Snip here--------------------------------------------------------
This branch removed empty catch clauses that were masking errors generated when python passed invalid system ids to the pathfinding functions in universe.  

It will cause new errors to be reported until the bad system ids are flushed from the python code.

With the errors unmasked, the python errors will report line numbers as usual, making fixes straightforward.

---

Issue #671, opened 2016-05-18
Add documentation for "Update Old Design" in Design Window


LGM-Doyle:
PRs #506 and #552 added the ability to persistently sort designs in the design window and to update old designs.

These new features need documentation in the UI section of the pedia and also tooltips explaining the usage.

See the discussion in [http://freeorion.org/forum/viewtopic.php?f=25&t=10090&sid=d961978af4ed03a5c22e0d7bcad1a375](url) for details.

Vezzra:
Has been addressed by #690, closing.

PR #690, opened 2016-06-04
Add Pedia article for Design Window in User Interface Category.

LGM-Doyle:
This is intended to address #671.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -4230,6 +4230,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]], 
 
 On the right hand side of the control bar are a number of icons for toggling display of other significant UI windows. Tooltips are available for these.'''
 
+DESIGN_WINDOW_ARTICLE_TITLE
+Design Window
+DESIGN_WINDOW_ARTICLE_TEXT
+'''The design window is for viewing or altering the designs of monsters or ships.
+
+There are four sub windows:  Design Starts, Ship Parts, Design Details and the Pedia.
+
+
+Design Starts has four tabs: empty Hulls, Finished designs, Saved designs from previous games and Monster designs for inspection.  Drag or double click hulls, designs or monsters to edit/examine them on the workbench.
+
+Finished designs appear in the same order as on the Production Window.  Designs can be drag-and-dropped within the list to change their order.  Delete designs with Obsolete from the context menu.  Ships already queued for production will continue to be built with the old design.  Save designs to disk for future games with Save from the context menu.
+
+Right click Saved designs and copy a single or all saved designs to the empire's known designs.
+
+
+Ship Parts displays parts. Parts whose associated technology has been researched may be used in current designs.  Other parts can not be used because either the associated technology has not been researched or the part is only available to monsters.
+
+The Available and Unavailable filters on Design Starts and Ship Parts toggle the display of parts that the empire can currently use or that are restricted.  Some unavailable parts may be unlocked by researching associated technologies.
+
+The Ship Parts window has filters to display/hide entire categories of parts.
+
+
+Design Details is the workbench for the current design.
+
+The current design may have internal, external and/or core slots.  Drag and drop or double click parts from Ship Parts to add/remove them to/from the current design.
+
+Add a name and optionally a description to the design.
+
+Click Add New Design to add the design to the end of the list of Finished designs.
+
+Click Update Old Design to replace the most recently selected finished design with a new design.  Update Old Design is only active if this design was started from a design in the Finished list.  Update Old Design only changes the design, any ships already queued for production or already produced will continue to use the old design.
+'''
+
 FLEET_MOVEMENT_ARTICLE_TITLE
 Fleet Movement
 FLEET_MOVEMENT_ARTICLE_TEXT
--- a/default/scripting/encyclopedia/DESIGN_WINDOW.focs.txt
+++ b/default/scripting/encyclopedia/DESIGN_WINDOW.focs.txt
@@ -1,0 +1,6 @@
+Article
+    name = "DESIGN_WINDOW_ARTICLE_TITLE"
+    category = "CATEGORY_USER_INTERFACE"
+    short_description = "DESIGN_WINDOW_ARTICLE_TITLE"
+    description = "DESIGN_WINDOW_ARTICLE_TEXT"
+    icon = "icons/tech/galactic_exploration.png"


default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
 
 On the right hand side of the control bar are a number of icons for toggling display of other significant UI windows. Tooltips are available for these.'''
 
+DESIGN_WINDOW_ARTICLE_TITLE
+Design Window
+DESIGN_WINDOW_ARTICLE_TEXT
+'''The design window is for viewing or altering blueprints of monsters or ships.
```
geoffthemedio: Why introduce "blueprints"? Not used anywhere else / before, is it? Seems like unnecessary extra jargon.

LGM-Doyle: I'm trying to create a difference in meaning between blueprints as the plan for a ship, the queued production of a ship and the ship which comes off the end of the production line.  

It is to get ahead of the question, "Why doesn't changing the design change all of the already queued designs as well?".  If people are asking themselves, "Why doesn't changing the blueprint, change the design in the queue?", then I was hoping the answer would be clear.

The question was asked [here.](http://freeorion.org/forum/viewtopic.php?f=25&t=10090&sid=d961978af4ed03a5c22e0d7bcad1a375)  I think that it is a question that will keep arising unless the UI is more obvious in the way it works.

So, I could either remove the word blueprint, or go all in and make the design window the blueprint window?

geoffthemedio: Remove the word blueprint.

LGM-Doyle:
@geoffthemedio I force pushed a no blueprint version.

default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
...
+
+
+Design Starts has four tabs: empty Hulls, Finished designs, Saved designs from previous games and Monster designs for inspection.  Drag or double click hulls, designs or monsters to edit/examine them on the workbench.
+
+Finished designs appear in the same order as on the Production Window.  Drag and drop designs within the list to change the order.  Delete designs with Obsolete from the context menu.  Ships already queued for production will continue to be built with the old design.  Save designs to disk for future games with Save from the context menu.
+
+Right click Saved designs and copy a single or all saved designs to the empire's known designs.
+
+
+Ship Parts shows parts that can be added to the current design.
```
geoffthemedio: Not really true. Parts that can't be added to a design are also shown.

default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
...
+
+The Available and Unavailable filters on Design Starts and Ship Parts toggle the display of parts that the empire can currently use or that are restricted.
```
geoffthemedio: some note about why parts would be restricted, and how to unrestrict them, would be useful...

default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
...
+
+The Ship Parts window has many filters to display/hide entire categories of parts.
```
geoffthemedio: nix "many"

default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
...
+
+
+Design Details is the workbench for the current design.
+
+The current design may have internal, external and/or core slots.  Drag and drop or double click parts from Ship Parts to add/remove them from the design.
+
+Add a name and possibly a description to the design.
```
geoffthemedio: possibly -> optionally

default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
...
+
+Click Add New Design and the design will be added to the end of the list of Finished designs.
+
+Click Update Old Design to replace the most recently selected finished design with a new design.  Update Old Design is only active if this design was started from a design in the Finished list.  Update Old Design only changes the design, any ships already queued for production will continue to use the old design.
```
geoffthemedio: and existing / already produced ships will be unchanged.

default/stringtables/en.txt
MatGB: Yes, I also wonder if there should be an 'unresearchable' toggle or other hiding mechanism for the stuff that isn't meant to be on player designs, there's a big difference for a player between "you haven't got this yet" and "you'll never have this"

LGM-Doyle:
@MatGB I agree.

It also concerns me that one of the states when both Available and Unavailable are untoggled shows no parts.  In that state if the player doesn't realize how they ended up with no parts displayed the design window is unusable.  It seems like an unnecessary gotcha.

Here is a thought.
Ship Parts could have 3 expandable/collapsable sections Available, Unresearched and Foreign.  Foreign is for unresearchable items like monster tech, experimenter tech and possibly other empire tech if we eventually do randomized or per species tech trees. 

If all sections are collapsed it would be clearer why there are no parts displayed.

It removes the available/unavailable toggles from the visually identical category buttons.

The unavailable parts wouldn't be interleaved with the available parts.  This may be a good/bad thing, depending on exactly how your are using the interface at the moment.

default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
...
 
+DESIGN_WINDOW_ARTICLE_TITLE
+Design Window
+DESIGN_WINDOW_ARTICLE_TEXT
+'''The design window is for viewing or altering the designs of monsters or ships.
+
+There are four sub windows:  Design Starts, Ship Parts, Design Details and the Pedia.
+
+
+Design Starts has four tabs: empty Hulls, Finished designs, Saved designs from previous games and Monster designs for inspection.  Drag or double click hulls, designs or monsters to edit/examine them on the workbench.
```
geoffthemedio, line 4241: capitalization of Starts of words Seems Random...

default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
...
+
+Finished designs appear in the same order as on the Production Window.  Drag and drop designs within the list to change the order.  Delete designs with Obsolete from the context menu.  Ships already queued for production will continue to be built with the old design.  Save designs to disk for future games with Save from the context menu.
```
geoffthemedio: oddly mixes imperative and descriptive sentences. should be in separate paragraphs, or precede the instructions with "you can", or better yet, rephrase to "Designs can be drag-and-dropped within the list to change their order."

default/stringtables/en.txt
```diff
@@ -4219,6 +4219,39 @@ Next are icons showing the player empire total [[encyclopedia INDUSTRY_TITLE]],
...
+
+Right click Saved designs and copy a single or all saved designs to the empire's known designs.
+
+
+Ship Parts displays parts. Parts whose associated technology has been researched may be used in current designs.  Other parts can not be used because either the associated technology has not been researched or the part is only available to monsters.
+
+The Available and Unavailable filters on Design Starts and Ship Parts toggle the display of parts that the empire can currently use or that are restricted.  Researching the associated technology makes parts available.  Some parts will never be available because they are only used for monsters.
```
geoffthemedio: Is that the only reason are part could not be possible to unlock?

default/stringtables/en.txt
LGM-Doyle, line 4241: I've capitalized words that are tab names from the UI, or context menu commands.  I could remove the capitalization or use quotes.

default/stringtables/en.txt
LGM-Doyle: I'm not certain but I think so.  That is assuming monsters includes experimentors.

Do you know otherwise?

dbenage-cx: Might drop it and the previous sentence.  
AFAIK, the only part not available from techs or strictly a monster is Krill Spawner.  Which could be completely unavailable if the player sets games to no specials.  This part might still fit the description, but future additions could come from a variety of conditions (granted with species; single-player only).

default/stringtables/en.txt
geoffthemedio, line 4241: That makes sense as it is then; it's fine.

default/stringtables/en.txt
geoffthemedio: I'd just write something like "Some unavailable parts may be unlocked by researching associated technologies."

geoffthemedio:
Can you squash into one commit?

LGM-Doyle:
Rebased to master and squashed.

PR #692, opened 2016-06-05
Production pedia: Fix non-stringtable name lookups

dbenage-cx:
Custom ship design names show "ERROR:..." for right click context pedia lookup.

PR uses the raw name entry when there is no stringtable reference.


geoffthemedio:
looks fine...

PR #694, opened 2016-06-05
Research queue: Add right click context for pedia lookup

dbenage-cx:
Adds option for pedia lookup when right clicking an item in the queue.

### DIFF
--- a/UI/ResearchWnd.cpp
+++ b/UI/ResearchWnd.cpp
@@ -196,12 +196,78 @@ namespace {
 }
 
 //////////////////////////////////////////////////
+// ResearchQueueListBox                         //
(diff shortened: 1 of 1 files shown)


geoffthemedio:
No easy way to move the common code in ItemRightClickedImpl int QueueListBox, I assume...

dbenage-cx:
Possible, not quite sure on easy or if it would be beneficial.
[Design](https://github.com/freeorion/freeorion/blob/master/UI/DesignWnd.cpp#L1810) would stick with existing overrides.  [Production](https://github.com/freeorion/freeorion/blob/master/UI/ProductionWnd.cpp#L611) is nearly identical except for rally and item.

PR #697, opened 2016-06-05
 Update fr.txt

Cjkjvfnby:
- remove trailing spaces
- remove unnecessary quoting
- update comments according en.txt

### DIFF
--- a/default/stringtables/fr.txt
+++ b/default/stringtables/fr.txt
@@ -218,7 +218,7 @@ SD_CARRIER_DESC
 Modèle prototype propulsant des chasseurs en cours de bataille en lieu et place d'une puissance de feu directe.
 
 SD_FLAK
-Prototype Antiaérien 
+Prototype Antiaérien
 SD_FLAK_DESC
 Modèle prototype équipé de canons antiaériens comme moyen de riposte contre les chasseurs.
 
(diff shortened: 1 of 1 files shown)


default/stringtables/fr.txt
```diff
@@ -889,6 +889,9 @@ L'Univers généré contient des erreurs. Consultez le fichier log pour le déta
 SERVER_TURN_EVENTS_ERRORS
 Les événements de tour scriptés contiennent des erreurs. Consultez le fichier log pour le détail des messages d'erreur. La partie peut continuer, mais le gameplay en sera probablement impacté.
 
+#ERROR_PYTHON_AI_CRASHED
+#Python AI for %1% crashed.
+
```
dbenage-cx: Might omit untranslated entries.  They will default to en.txt and is more noticeable for translators.

geoffthemedio:
Are the quoting changes consistent with the english stringtable?

Cjkjvfnby:
Quoting changes consistent with logic,  I remove quote there they are not necessary.

Ouaz:
Ok with all the modifications but not with the removal of extra blank lines (e.g. line 4304). The number of lines has to be exactly the same between the english and the french stringtables, because it's way more convenient for me when I want to compare the two files. If a given text is not at the same place (i.e. same line) in the two files, it will be a pain to handle.

If the same blank lines removals are brought to en.txt, it will be OK then. The others modifications should be also brought to en.txt (unnecessary triple quotes).

EDIT:
As I commited my weekly fr.txt update, there is now some conflicts with this PR (new translations). I created a branch on my fork with your modifications implemented in fr.txt (minus the blank lines removals and the new translations) > https://github.com/Ouaz/freeorion/commit/6985d84a88ff9137bea9ea9011bf59ac0dfe31c6
so you can patch this PR (I don't know how to do it :p) and it will be ok to merge it.

Cjkjvfnby:
>  I created a branch on my fork with your modifications implemented in fr.txt 

In that case just merge this yourself with next fr update.

PR #700, opened 2016-06-07
Fix bug referencing NULL validator in OptionsDB.

LGM-Doyle:
flag options have no validator.  Old code referenced the non-existent
validator.

### DIFF
--- a/util/OptionsDB.cpp
+++ b/util/OptionsDB.cpp
@@ -93,8 +93,18 @@ OptionsDB::Option::Option(char short_name_, const std::string& name_, const boos
 }
 
 bool OptionsDB::Option::SetFromString(const std::string& str) {
-    boost::any value_ = flag ? boost::lexical_cast<bool>(str) : validator->Validate(str);
-    bool changed =  validator->String(value) != validator->String(value_);
+    bool changed = false;
+    boost::any value_;
+
+    if (!flag) {
+        value_ = validator->Validate(str);
+        changed = validator->String(value) != validator->String(value_);
+    } else {
+        value_ = boost::lexical_cast<bool>(str);
+        changed = (boost::lexical_cast<std::string>(boost::any_cast<bool>(value))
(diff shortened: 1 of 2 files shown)


LGM-Doyle:
@spikethehobbitmage  thanks for reporting this bug with PR #639.  I believe that this fixes the problem.

util/OptionsDB.cpp
```diff
@@ -93,8 +93,18 @@ OptionsDB::Option::Option(char short_name_, const std::string& name_, const boos
 }
 
 bool OptionsDB::Option::SetFromString(const std::string& str) {
-    boost::any value_ = flag ? boost::lexical_cast<bool>(str) : validator->Validate(str);
-    bool changed =  validator->String(value) != validator->String(value_);
+    bool changed = false;
+    boost::any value_;
+
+    if (!flag) {
```
geoffthemedio, line 99: Why test if flag is true, rather than if validator is non-null?

LGM-Doyle, line 99: tl;dr   `OptionsDB.Add()` enforces that all `!flag` options have Validators and all `flag` options do not.

As stated earlier in the code 
     `///< a validator for the option. Flags have no validators; lexical_cast boolean conversions are done for them.`

This is enforced in the construction of Option, by `OptionsDB.Add()`

The rest of the code uses that same test `if (!flag)`, to check for the absence of a Validator.

So to remain consistent with the intent of Option I followed suit.

I think that the reason there are no Validators for flags is to avoid standardizing the value of booleans in the config.xml file that is written to disk.  I don't think this is useful since in config.xml there is no comment about which values are acceptable for booleans, so a user still has to try something and hope it sticks.

I think that the best solution is to re-factor Option so that it always has a Validator.  Flags can use a new Validator based on `boost::lexical_cast<std::string>(boost::any_cast<bool>(value)`.  The chief benefit is code simplicity.

I did not pursue this fix because without behavioral tests in place for OptionsDB, the more minimal solution is least likely to introduce bugs.

PR #706, opened 2016-06-09
FOCS: Adjust mine damage macro for limited scope

### LINKED ISSUE #703: Mine damage in system with no mines
Build 2016-05-30.fc49608
On turn 68 an AI troopship wandered to one of my planets (Horst α) and was destroyed. SitRep reported that the enemy troopship was destroyed by mines, but I had not researched ANY defense techs yet (no Planetary Defense, Mines, etc., only Force Filed Harmonics), and the only ship in the system was an outpost ship. The system had no hidden natives, nor Fortress special.

The same situation happened again in a different system a few turns later.

Attached are screenshot for beginning of turn 68 and savegame of turn 69 (save of turn 68 not available).

mines
[save-20160607_093444.zip](https://github.com/freeorion/freeorion/files/305530/save-20160607_093444.zip)

dbenage-cx:
Fixes #703 

Mine damage from an owned fortress was applied to all empire systems.

### DIFF
--- a/default/scripting/techs/defense/mines.macros
+++ b/default/scripting/techs/defense/mines.macros
@@ -1,26 +1,19 @@
+//#debug_macros "mines_macro.log"
 // EffectsGroups for System Defense Mines
 // arg1 total damage
 // arg2 priority
+// arg3 scope of systems effected: SOURCE, EMPIRE
 EG_SYSTEM_MINES
 '''        // damage ship; sitreps issued below
         EffectsGroup
             scope = And [
                 Ship
-                ContainedBy And [
-                    System 
-                    Contains Or [
-                        Source
-                        And [
-                            Planet
-                            OwnedBy empire = Source.Owner
-                        ]
-                    ]
-                ]
+                [[SYSTEM_MINES_SCOPE(@3@)]]
                 Or [
                     OwnedBy affiliation = EnemyOf empire = Source.Owner
                     Unowned
                 ]
-                (RootCandidate.Owner != Source.Owner)
+                Not OwnedBy empire = Source.Owner
                 Structure low = @1@ + 0.0001
             ]
             stackinggroup = "DEF_MINES_DAMAGE_STACK"
@@ -32,21 +25,12 @@ EG_SYSTEM_MINES
         EffectsGroup
             scope = And [
                 Ship
-                ContainedBy And [
-                    System 
-                    Contains Or [
-                        Source
-                        And [
-                            Planet
-                            OwnedBy empire = Source.Owner
-                        ]
-                    ]
-                ]
+                [[SYSTEM_MINES_SCOPE(@3@)]]
                 Or [
                     OwnedBy affiliation = EnemyOf empire = Source.Owner
                     Unowned
                 ]
-                (RootCandidate.Owner != Source.Owner)
+                Not OwnedBy empire = Source.Owner
                 Structure high = @1@
             ]
             stackinggroup = "DEF_MINES_DAMAGE_STACK"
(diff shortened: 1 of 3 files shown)


default/scripting/techs/defense/mines.macros
```diff
@@ -1,6 +1,7 @@
 // EffectsGroups for System Defense Mines
 // arg1 total damage
 // arg2 priority
+// arg3 planet scope (-1=all systems with an empire planet, 1=source system only)
```
geoffthemedio: I question whether this is a good use of macro parameters, both for readability and evaluation efficiency reasons...

default/scripting/techs/defense/mines.macros
```diff
@@ -9,8 +10,12 @@ EG_SYSTEM_MINES
                 ContainedBy And [
                     System 
                     Contains Or [
-                        Source
                         And [
```
geoffthemedio: Given that most of this scope condition is reused numerous times in this macro, a sub-macro for it seems reasonable...

dbenage-cx:
Rebased.

default/scripting/techs/defense/mines.macros
```diff
@@ -271,3 +190,17 @@ EG_SYSTEM_MINES
                     empire = Target.Owner
             ]
 '''
+
+SYSTEM_CONTAINS_SOURCE_OR_OWNED_PLANET
+'''                ContainedBy And [
+                    System 
+                    Contains Or [
+                        Source
```
geoffthemedio: Why is the Source condition used?

Wouldn't the And [] case cover that (assuming the source is the empire's capital planet)?

default/scripting/techs/defense/mines.macros
```diff
@@ -271,3 +190,17 @@ EG_SYSTEM_MINES
...
+                        And [
+                            Planet
+                            OwnedBy empire = Source.Owner
+                            OwnerHasTech name = ThisTech
```
geoffthemedio: Why is this has tech condition needed? If this macro is being used in a tech script, then the source's owner will always have this tech. (And if it's used in something other than a tech, "ThisTech" doesn't make sense...)

default/scripting/techs/defense/mines.macros
dbenage-cx: When source is a special (fortress), Source is matching for the containing system.
When source is a tech, it may also match for the home system, which would apply for the And[] case anyway.

default/scripting/techs/defense/mines.macros
dbenage-cx: OwnerHasTech is included to keep specials from applying empire-wide. This is wasted effort for the techs, attempts at ObjectType did not pan out (bad syntax? `(Source.ObjectType type != Special)`).
`ThisTech` was only chosen due to using it with OwnerHasTech.  Prefer `CurrentContent`?

geoffthemedio: How does a has-tech condition affect whether something applies empire-wide? Either an empire has a tech or it doesn't; there's no distinction about this from some of the things the empire owns.

Specials are not objects. Specials just have some effectsgroups, and whatever object they are applied to will be the source object.

Avoid using != when there is a suitable condition. eg. instead of

```
(RootCandidate.Owner != Source.Owner)
```

Use

```
Not OwnedBy empire = Source.Owner
```

default/scripting/techs/defense/mines.macros
geoffthemedio: Source is never a special or a tech; specials and techs are not objects.

The source could be an object a special was applied to, such as a planet, or the capital of an empire in the case of a tech.

If you have two very different cases - a special effect and a tech effect - that have different scopes - "in the system with the special", and "in a system that also contains a planet owned by the empire that knows the tech" - then use two different conditions. It's much clearer that way, and you can optimize the conditions, such as by using

```
    InSystem id = Source.SystemID
```

for the special scope, and

```
    ContainedBy And [
        System
        Contains And [
            Planet
            OwnedBy empire = Source.Owner
        ]
    ]
```

for the tech scope

dbenage-cx:
Rebased

Issue #656, opened 2016-05-10
Add a common location for python utility/tool files.


LGM-Doyle:
There are three python portions of freeorion: universe_generation, turn_events and the AI.  Currently, they can only access files in their starting directory and below.  Consequently, utility/debugging code common to all three gets copied into each location.

There should be a common location for python utility classes, to avoid this code duplication.

Cjkjvfnby:
This path can be added as `PYTHONPATH` at start of python.

Vezzra:
Yes, absolutely agree, and I've already (long term) plans to implement something like that.

This includes moving the AI script directories to the `python` subfolder the `turn_events` and `universe_generation` script folders are already in - @freeorion/ai-team, is that ok with you? Then we can either place the "common" Python modules directly into the `python` folder, or create a dedicated subfolder (called "common" or "shared" or something like that). Any preferences/votes on that? Personally I'd prefer the latter option (a dedicated "common" subfolder).

ghost:
I'll vote a common folder.  Arranging the scripts is actually pretty easy, but the Boost binding code does some strange things and will take time to fix.
[edit]
We don't need PYTHONPATH.  We can use `freeorion.__path__` to set up our imports.
It works like this:
1) Native freeorion module is just a stub.
2) Set `freeorion.__path__` to where our native extensions are and load the ones we want.
4) Set `freeorion.__path__` to $RESOURCEDIR/python
5) Profit.
[/edit]

Cjkjvfnby:
This ticket is about pure python modules. 

``` diff
- import freeorion_utils
+ import commons.freeorion_utils
```

LGM-Doyle:
All of the python started from within freeorion (generate universe, effects and AI) is already setup with the default/python directory as a common path.  I didn't realize this when I initially posted this issue.

Setting up a common location for code should be as simple as creating the directory under default/python.

Cjkjvfnby:
> Setting up a common location for code should be as simple as creating the directory under default/python.

and add this directory as PYTHONPATH

ghost:
Currently, all pure python modules are being loaded at global scope, leading to potential conflict with other packages.
Putting them all in default/python and adding that path as `freeorion.__path__` makes them all submodules of the `freeorion` module.

ghost:
[...]

PR #707, opened 2016-06-10
Modularize python

ghost:
This is a WIP for Issue #656.
Pure python modules have been updated to a single tree.  Native code still needs to be updated to the new layout.


Vezzra:
These are quite considerable changes... will need review by the @freeorion/ai-team and myself.

Any objections to postponing this until after the 0.4.6 release?

Cjkjvfnby:
Post release.

`from __future__ import absolute_import` should not be added in this PR.

ghost:
Post release is pretty much a given.  There is no way this would be ready in time.

I've already run into a major snag in the AI module with cyclic imports.  To be honest, there must be serious bugs in python's import code for it to load at all.  Quite frankly, the way Freeorion handles python scares me.

What I'm going to have to do is create another PR just for the import loops, and then redo this on top.

@Vezzra any suggestions, advice, or requests from you or the team are most welcome.

@Cjkjvfnby May I ask what the problem is with `absolute_import`?  Just curious.

Cjkjvfnby:
@spikethehobbitmage, changes in commit with name `move default/AI to default/python/AI` is separate feature and can be done as single PR.  This should not add new conflicts for merging. But I don't know how it will work with rebasing.  Best way is to announce move of modules and wait till developers update all theirs branches to master.  I don't mind to move modules now, but lets @Vezzra, decide.

Problem with absolute_import is next:
- no need to mix it inside this pr
- adding this to part of files is not good.  We should use old or new imports, not both.

ghost:
@Cjkjvfnby Thanks for explaining.  Missing details like those are reasons I shouldn't code at 3:00AM.

Vezzra:
> Quite frankly, the way Freeorion handles python scares me.

Hm, what exactly do you mean by that? The way the Python embedding is done, or just how things are implemented in the Python scripts themselves (like the cyclic imports you mentioned)? I'm certainly open for fixing messed up things (that is, after release of course :wink:).

> any suggestions, advice, or requests from you or the team are most welcome.

I'll try - post release... :wink:

Vezzra:
> I don't mind to move modules now, but lets @Vezzra, decide.

Erm, I definitely prefer post release (I'm saying "post release" a lot lately, don't I...? 😉)

ghost:
The Python embedding has numerous throwbacks to the old Boost Python C interface that never really worked right.
As for the import loops, I'm officially an idiot.  Either the python import semantics have changed sometime in the last decade, or I got confused somewhere.  After re-reading the spec, I was doing it wrong.  Old style import allows loops, but don't guarantee the module will actually be ready, while new style is the reverse.  :/
So, post release.  OK.  I'll shelve this for now.

Vezzra:
> The Python embedding has numerous throwbacks to the old Boost Python C interface that never really worked right.

That's probably due to the time when the FO Python interface was first implemented - I guess the old Boost Python C interface was the current one back then.

How well do you know Boost Python? I want to revise/harmonize/unifiy the AI and server Python interfaces, but my grasp on Boost Python isn't exactly, well, _strong_, to put it mildly. So I've been dragging my feet a lot in that regard. If you're game, you're welcome to help with that task (post release of course). 😃

ghost:
I actually tried fixing the FO Python interface back when the Boost C Interface was state-of-the-art.  I gave up because it was actually impossible, due to Boost.  It will take me some time to learn the C++ interface, but what I've seen is so much better already.
Regarding the interfaces, the AI and Player interfaces should be unified, but the server interface is inherently different.  The modularization work was intended as a precurser to that.  It is probably a good thing it is being delayed some, since I need to rethink my approach, but such is life.

Vezzra:
> the server interface is inherently different

Well, yes, of course, but there is still a large "common" part which is the same for client and server, so doesn't need to be implemented twice (currently we have a lot of redundant code). My idea is to have a base "common" interface, and derive the client and server interfaces from that (implementing only the differences in those derived interfaces).

Hence the `CommonFramework`, `AIFramework` and `ServerFramework` (and related) C++ modules, these have been my very first step in that direction.

ghost:
True that.  On the other hand, unless the AI is granted special privileges, the player and AI should share an interface.

Vezzra:
I absolutely agree that the human and AI _client_ interfaces should be unified. When I first discussed my plans to revise the Python interface with @Dilvish-fo, he specifically brought this up.

He even wants to go so far as to provide some kind of scripting/plugin interface for the human client that can be used by players to provide custom Python scripts which "extend" the human client (if I understood his ideas correctly).

ghost:
A user scripting interface wouldn't be all that hard to implement.  It could even be made interactive by including a python shell.  The hard part would be allowing the GUI itself to be scripted, but that is stuff for the future.

geoffthemedio:
Any chance this will be useful, or should it be closed?

Dilvish-fo:
Looks to me like some of the ideas have already been adopted, and the rest apparently no one saw much need for.  Being Closed, with the label Superseded, seems appropriate to me.

PR #709, opened 2016-06-12
Remove secondary file check of savegames

### LINKED ISSUE #693: Unable to load multiplayer saves.
With a valid multiplayer savefile:
- Select 'Load'->'...' on Multiplayer setup screen.
- Select file to load (xx.mps)
- Click 'OK'

Result: Nothing (screen remains)
  Log: "Client : SaveFileDialog.cpp:765 : SaveFileDialog::Confirm: problem with result: ./xx.mps"

Expected: Proceed with setup with selected filename.

Platform: Fedora 23 / gcc 5.3.1

Filing issue as I'm not sure what ./ should resolve to under each circumstance (e.g. non-local clients).

dbenage-cx:
Fixes #693

User is still kept from loading a non-existing file, as CheckChoiceValidity() does not enable the OK button.

### DIFF
--- a/UI/SaveFileDialog.cpp
+++ b/UI/SaveFileDialog.cpp
@@ -774,9 +774,6 @@ void SaveFileDialog::Confirm() {
             ErrorLogger() << "SaveFileDialog::Confirm: Invalid status for file: " << Result();
             return;
         }
-    } else if (!fs::is_regular_file(chosen_full_path)) {
-        ErrorLogger() << "SaveFileDialog::Confirm: problem with result: " << Result();
-        return;
     }
 
     CloseClicked();


dbenage-cx:
Comment due to server error during PR submit (no notifications generated).

geoffthemedio:
Keeping such extra checks around can be useful, in cases where the filesystem is modified between when a selection is made and an attempt is made to use the selection...

geoffthemedio:
Having looked into the issue, I now understand the motivation, and agree.

dbenage-cx:
@geoffthemedio Having just finished testing, I noticed the merge.  I've pushed to a [separate branch](https://github.com/dbenage-cx/freeorion/tree/load_server_savegame) if it looks more suitable.

PR #710, opened 2016-06-12
Remove initFreeOrionAI

Cjkjvfnby:
This function have no sense since same can be done at import so just remove it. 
- I did not tested if it works 
- it should be merged close to test build
- need to update wiki after merge

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -37,6 +37,12 @@ except ImportError:
     pass
 
 
+ai_config = fo.getAIConfigStr()
+print "Initialized FreeOrion Python AI with ai_config string '%s'" % ai_config
+user_dir = fo.getUserDataDir()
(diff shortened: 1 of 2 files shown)


Vezzra:
AFAICT you have a point here, `initFreeOrionAI` isn't strictly necessary. Looking at the C++ code I can't think of anything that could only be done in this function and not during import of `FreeOrionAI.py`.

I think `initFreeOrionAI` has been provided as an optional, convenient place where AI scripters can run some code that does some basic initialization of the AI client, before the actual game is initialized.

If there aren't any objections, I guess we can remove it. Comments, opinions anyone? @freeorion/ai-team in particular?

ghost:
Tested.  Works.  This function can be safely removed.

Vezzra:
@Cjkjvfnby, please resolve conflicts and rebase. Feel free to merge when done.

Cjkjvfnby:
@Vezzra, done

PR #712, opened 2016-06-13
Parse: Replace FOCS macros with single whitespace

dbenage-cx:
Macros are currently replaced with whitespace for each character, which can lead to long lines.

### DIFF
--- a/parse/Parse.cpp
+++ b/parse/Parse.cpp
@@ -269,8 +269,8 @@ namespace parse {
                     ErrorLogger() << "Duplicate macro key foud: " << macro_key << ".  Ignoring duplicate.";
                 }
 
-                // remove macro definition from text by replacing with whitespace that is ignored by later parsing
-                text.replace(text_it + match.position(), text_it + match.position() + match.length(), match.length(), ' ');
+                // remove macro definition from text by replacing with a newline that is ignored by later parsing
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I think my motivation for how it is was to keep the numbers of lines in the file the same after macro removal. This doesn't help much if multi-line macros are inserts into the scripts, though...

dbenage-cx:
Stumbled on when checking mines macro (dumping to file after macro inclusions).  In that case it outputs a single line of 7400+ spaces as replacement.  Close and post as issue for later?

geoffthemedio:
A concern: does replacing the text with a smaller bit of text cause a bunch of expensive reallocations of the string? Replacing with spaces should avoid that... if it's replacing with the right number of spaces...

dbenage-cx:
Honestly don't know for this signature, though changing to `.replace(start_pos, length, ' ')` should just be a single reallocation.

geoffthemedio:
I mean an allocation for each macro, not each character.

Though I suppose it would just be a copies, not reallocations, since it's getting smaller.

geoffthemedio:
Since macros can contain an arbitrary number of newlines, the line numbers aren't going to line up anyway... So it seems fine to just erase instead of replace the macro text from the string.

dbenage-cx:
Rebased

PR #713, opened 2016-06-17
Create common folder for AI code

### LINKED ISSUE #656: Add a common location for python utility/tool files.

Cjkjvfnby:
- name for shared modules folder is `common` 
- `print_utils` is moved with out changes
- imports updated

I think it is final changes required to close https://github.com/freeorion/freeorion/issues/656
Moving other modules to common can be done when it will be required.


ghost:
This patch is missing an `__init__.py` file.  I'll add it to PR #714.

Cjkjvfnby:
@spikethehobbitmage thank you.

Fixing this PR in other place is not good idea: each PR should have working code.

I rebased this PR with missed `__init__`

ghost:
@Cjkjvfnby acknowledged.  PR updated.

PR #714, opened 2016-06-18
move default/AI to default/python/AI

ghost:
This should be fast-forwarded onto master, then PR #713 applied on top.


ghost:
Scratch that, this applies on top of #713.

PR #716, opened 2016-06-18
In SidePanel confine planet rename context menu item inside planet name.

LGM-Doyle:
The SidePanel has a context menu item to rename the planet if it is
owned by the player.  Change the condition so that it only appears when
the cursor is over the planet name.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -1851,8 +1851,11 @@ void SidePanel::PlanetPanel::RClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_
     }
 
     GG::MenuItem menu_contents;
-    if (planet->OwnedBy(HumanClientApp::GetApp()->EmpireID()) && m_order_issuing_enabled )
+    if (planet->OwnedBy(HumanClientApp::GetApp()->EmpireID()) && m_order_issuing_enabled
+        && m_planet_name->InClient(pt))
+    {
         menu_contents.next_level.push_back(GG::MenuItem(UserString("SP_RENAME_PLANET"), 1, false, false));
+    }
 
     menu_contents.next_level.push_back(GG::MenuItem(UserString("SP_PLANET_SUITABILITY"), 2, false, false));
 


geoffthemedio:
why?

TheSilentOne1:
I have to say I also often click "rename" when I actually want to look up the planet's suitability. Moving "rename" to the bottom of the menu would work for me though.

LGM-Doyle:
Rename planet is a context menu item that is contextually appropriate when you are hovering over the planet name.  

Checking the suitability of a planet is always in the context menu and almost always the item that the player wants to click on.

As @TheSilentOne1 said another solution is to just bump it to the second item and perhaps put a line between it and suitability.

Another solution would be to make it the double click item when over the planet name box.

Renaming a planet is rare but I think right clicking on the name is sufficiently intuitive and discoveralbe.

p.s. A final related quibble is I think that the renaming should be inline.  Opening a dialog box to change a text item that can be seen on the screen seems inelegant.

geoffthemedio:
If you'd like to rework things so the renaming can be done inline, and can find a way to do it that doesn't require the name text to look like an edit box all the time, and want to do it the same way for everything else that can be renamed, and can do THAT in a way that doesn't conflict with drag-dropping ships and fleets around, then OK.

LGM-Doyle:
As you point out, it is non-trivial, and probably requires a deep dive into the bowels of GG.  So I hope someone else can see an easy way to do it.

Vezzra:
I doubt that there is an _easy_ way to do a change like _that_. Meaning, if we don't want to incorporate the suggested change as currently implemented by this PR, I propose to flag it for post release.

Comments?

UI/SidePanel.cpp
```diff
@@ -1851,7 +1851,8 @@ void SidePanel::PlanetPanel::RClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_
     }
 
     GG::MenuItem menu_contents;
-    if (planet->OwnedBy(HumanClientApp::GetApp()->EmpireID()) && m_order_issuing_enabled )
+    if (planet->OwnedBy(HumanClientApp::GetApp()->EmpireID()) && m_order_issuing_enabled
```
geoffthemedio, line 1854: if this is split over two lines, put { } around the following statement

LGM-Doyle:
I've added the brackets.

@Vezzra The second part of the discussion about inlining the edit boxes was an aside.  I did preface it with p.s. (postscript).  This PR is simply about restricting the context menu for changing the planet name to the area over the planet name.

UI/SidePanel.cpp
```diff
@@ -1851,8 +1851,10 @@ void SidePanel::PlanetPanel::RClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_
...
+        && m_planet_name->InClient(pt)) {
```
geoffthemedio: put the { on a new line, indented to match the start of the if and the }

```
if (something &&
    something_else)
{ do_something(); }

if (something &&
    something_else)
{
    // also acceptable
    like_this();
}
```

LGM-Doyle:
Repushed.

LGM-Doyle:
@MatGB I noticed while playing that the HONEYCOMB special deletes planets in the surrounding systems, which results in stars with no planets.

Do you want to create bug report to fix this?

Vezzra:
Hm, definitely a case we've overlooked, but needs to be addressed. However, I consider this a design decision - do we want to make a exception here to the "no stars without planets" rule for the Honeycomb special, or shall we set the star type of the affected systems to "no star"?

That discussion actually belongs on the design forum, so I've [opened a corresponding thread](http://freeorion.org/forum/viewtopic.php?f=6&t=10147). Please continue discussion there.

PR #718, opened 2016-06-18
Require systems with non black hole stars to have a satellite.

LGM-Doyle:
@MatGB suggested this change.  

Once he suggested it I started to notice how much time I was spending clicking on systems with stars, because I couldn't remember if they had planets or not.

@MatGB may want to rebalance it to have fewer stars, since they all now have at least one planet/asteroids.

I still allowed black holes to have no planets.  I think that they are rare and that flavor-wise the absence of planets works.

### DIFF
--- a/default/python/universe_generation/planets.py
+++ b/default/python/universe_generation/planets.py
@@ -34,11 +34,11 @@ def base_chance_of_planet(planet_density, galaxy_shape, star_type, orbit, planet
             + tables.ORBIT_MOD_TO_PLANET_SIZE_DIST[orbit][planet_size]
             + tables.GALAXY_SHAPE_MOD_TO_PLANET_SIZE_DIST[galaxy_shape][planet_size])
 
+
 def can_have_planets(star_type, orbits, planet_density, galaxy_shape):
     """
     Return True if a system can have any planets.
     """
-
     for orbit in orbits:
         base_chance_none = base_chance_of_planet(planet_density, galaxy_shape, star_type, orbit, fo.planetSize.noWorld)
         base_chance_any  = max([base_chance_of_planet(planet_density, galaxy_shape, star_type, orbit, size)
@@ -53,7 +53,6 @@ def calc_planet_size(star_type, orbit, planet_density, galaxy_shape):
     """
     Calculate planet size for a potential new planet based on planet density setup option, star type and orbit number.
     """
-
     # try to pick a planet size by making a series of "rolls" (1-100)
     # for each planet size, and take the highest modified roll
     planet_size = fo.planetSize.unknown
@@ -88,7 +87,6 @@ def calc_planet_type(star_type, orbit, planet_size):
 
     TODO: take into account star type and orbit number for determining planet type.
     """
-
     # check specified planet size to determine if we want a planet at all
     if planet_size in planet_sizes:
         # if yes, determine planet type based on planet size...
@@ -106,7 +104,6 @@ def generate_a_planet(system, star_type, orbit, planet_density, galaxy_shape):
     """
     Place a planet in an orbit of a system. Return True on success
     """
-
     planet_size = calc_planet_size(star_type, orbit, planet_density, galaxy_shape)
     if planet_size not in planet_sizes:
         return False
--- a/default/python/universe_generation/planets.py
+++ b/default/python/universe_generation/planets.py
@@ -28,6 +28,27 @@ planet_types_real = (fo.planetType.swamp,  fo.planetType.radiated,  fo.planetTyp
                      fo.planetType.ocean)
 
 
+def base_chance_of_planet(planet_density, galaxy_shape, star_type, orbit, planet_size):
+    return (tables.DENSITY_MOD_TO_PLANET_SIZE_DIST[planet_density][planet_size]
+            + tables.STAR_TYPE_MOD_TO_PLANET_SIZE_DIST[star_type][planet_size]
+            + tables.ORBIT_MOD_TO_PLANET_SIZE_DIST[orbit][planet_size]
+            + tables.GALAXY_SHAPE_MOD_TO_PLANET_SIZE_DIST[galaxy_shape][planet_size])
+
+def can_have_planets(star_type, orbits, planet_density, galaxy_shape):
+    """
+    Return True if a system can have any planets.
+    """
+
+    for orbit in orbits:
+        base_chance_none = base_chance_of_planet(planet_density, galaxy_shape, star_type, orbit, fo.planetSize.noWorld)
+        base_chance_any  = max([base_chance_of_planet(planet_density, galaxy_shape, star_type, orbit, size)
+                                for size in planet_sizes])
+
+        if base_chance_any > base_chance_none - 100:
+            return True
+    return False
+
+
 def calc_planet_size(star_type, orbit, planet_density, galaxy_shape):
     """
     Calculate planet size for a potential new planet based on planet density setup option, star type and orbit number.
@@ -38,15 +59,12 @@ def calc_planet_size(star_type, orbit, planet_density, galaxy_shape):
     planet_size = fo.planetSize.unknown
     try:
         max_roll = 0
-        for candidate in planet_sizes_all:
-            roll = random.randint(1, 100) \
-                + tables.DENSITY_MOD_TO_PLANET_SIZE_DIST[planet_density][candidate] \
-                + tables.STAR_TYPE_MOD_TO_PLANET_SIZE_DIST[star_type][candidate] \
-                + tables.ORBIT_MOD_TO_PLANET_SIZE_DIST[orbit][candidate] \
-                + tables.GALAXY_SHAPE_MOD_TO_PLANET_SIZE_DIST[galaxy_shape][candidate]
+        for size in planet_sizes_all:
+            roll = (random.randint(1, 100)
+                    + base_chance_of_planet(planet_density, galaxy_shape, star_type, orbit, size))
             if max_roll < roll:
                 max_roll = roll
-                planet_size = candidate
+                planet_size = size
     except:
         # in case of an error play it safe and set planet size to invalid
         planet_size = fo.planetSize.unknown
@@ -82,3 +100,22 @@ def calc_planet_type(star_type, orbit, planet_size):
             return random.choice(planet_types_real)
     else:
         return fo.planetType.unknown
+
+
+def generate_a_planet(system, star_type, orbit, planet_density, galaxy_shape):
+    """
+    Place a planet in an orbit of a system. Return True on success
+    """
+
+    planet_size = calc_planet_size(star_type, orbit, planet_density, galaxy_shape)
+    if planet_size not in planet_sizes:
+        return False
+    # ok, we want a planet, determine planet type and generate the planet
+    planet_type = calc_planet_type(star_type, orbit, planet_size)
+    if planet_type == fo.planetType.unknown:
+        return False
+    if fo.create_planet(planet_size, planet_type, system, orbit, "") == fo.invalid_object():
+        # create planet failed, report an error
+        util.report_error("Python generate_systems: create planet in system %d failed" % system)
+        return False
+    return True
--- a/default/python/universe_generation/starsystems.py
+++ b/default/python/universe_generation/starsystems.py
@@ -84,13 +84,37 @@ def generate_systems(pos_list, gsd):
                               % (position[0], position[1]))
             continue
         sys_list.append(system)
-        for orbit in range(fo.sys_get_num_orbits(system)):
-            # check for each orbit if a planet shall be created by determining planet size
-            planet_size = planets.calc_planet_size(star_type, orbit, gsd.planet_density, gsd.shape)
-            if planet_size in planets.planet_sizes:
-                # ok, we want a planet, determine planet type and generate the planet
-                planet_type = planets.calc_planet_type(star_type, orbit, planet_size)
-                if fo.create_planet(planet_size, planet_type, system, orbit, "") == fo.invalid_object():
-                    # create planet failed, report an error and try to continue with next orbit
-                    util.report_error("Python generate_systems: create planet in system %d failed" % system)
+
+        orbits = range(fo.sys_get_num_orbits(system))
+
+        if not planets.can_have_planets(star_type, orbits, gsd.planet_density, gsd.shape):
+            continue
+
+        # Try to generate planets in each orbit.
+        # If after each orbit is tried once there are no planets then
+        # keep trying until a single planet is placed.
+        # Except for black hole systems, which can be empty.
+
+        at_least_one_planet = False
+        random.shuffle(orbits)
+        for orbit in orbits:
+            if planets.generate_a_planet(system, star_type, orbit, gsd.planet_density, gsd.shape):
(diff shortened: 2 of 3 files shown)


MatGB:
As we've recently changed the Low Planets setting to make empty space more common anyway, it'll hopefully balance out, I'll run this for my next game (I'm currently proving to myself just how overpowered Egassem are at the moment).

I would like, for game balance purposes, to have even black holes have something you can put an outpost on, it's always annoying when you've got the generator tech, multiple black holes in supply, but none have anything outpostable. On the other hand, my recent changes to the starlane drive give you two ways to create a black hole system, so it might work out, and there's still a chance even if not a good one. 

Thanks for this.

default/python/universe_generation/planets.py
```diff
@@ -28,6 +28,27 @@
...
+    return (tables.DENSITY_MOD_TO_PLANET_SIZE_DIST[planet_density][planet_size]
+            + tables.STAR_TYPE_MOD_TO_PLANET_SIZE_DIST[star_type][planet_size]
+            + tables.ORBIT_MOD_TO_PLANET_SIZE_DIST[orbit][planet_size]
+            + tables.GALAXY_SHAPE_MOD_TO_PLANET_SIZE_DIST[galaxy_shape][planet_size])
+
+def can_have_planets(star_type, orbits, planet_density, galaxy_shape):
+    """
+    Return True if a system can have any planets.
+    """
+
```
Cjkjvfnby: No empty line after docstring (here and in other places)

default/python/universe_generation/starsystems.py
```diff
@@ -84,13 +84,37 @@ def generate_systems(pos_list, gsd):
...
+        random.shuffle(orbits)
+        for orbit in orbits:
+            if planets.generate_a_planet(system, star_type, orbit, gsd.planet_density, gsd.shape):
+                at_least_one_planet = True
+
+        if star_type == fo.starType.blackHole:
+            continue
+
+        ii_safety = 1000
+        while not at_least_one_planet and ii_safety > 0:
```
Cjkjvfnby: you can change `ii_safety > 0` to `ii_safety`

default/python/universe_generation/starsystems.py
```diff
@@ -84,13 +84,37 @@ def generate_systems(pos_list, gsd):
...
+            ii_safety -= 1
+            for orbit in orbits:
+                if planets.generate_a_planet(system, star_type, orbit, gsd.planet_density, gsd.shape):
+                    at_least_one_planet = True
```
Cjkjvfnby: You can drop some variables and simplify this code

``` python
for x in xrange(1000):
    if condition:
        break
else:  # if no break was in cycle
    print >> sys.stderr
```

default/python/universe_generation/starsystems.py
```diff
@@ -84,13 +84,37 @@ def generate_systems(pos_list, gsd):
...
+                    break
+
+        if not at_least_one_planet:
+            # Intentionally non-modal.  Should be a warning.
+            print >> sys.stderr, ("Python generate_systems: place planets in system at position (%f, %f) failed"
```
Cjkjvfnby: Do we need to print full float, maybe `%.1f`?

default/python/universe_generation/starsystems.py
```diff
@@ -84,13 +84,37 @@ def generate_systems(pos_list, gsd):
...
+        # Except for black hole systems, which can be empty.
+
+        at_least_one_planet = False
+        random.shuffle(orbits)
+        for orbit in orbits:
+            if planets.generate_a_planet(system, star_type, orbit, gsd.planet_density, gsd.shape):
+                at_least_one_planet = True
+
+        if star_type == fo.starType.blackHole:
+            continue
```
LGM-Doyle, line 115: Thanks for catching that.  I have fixed it.

Two cool notes:
- Planet graphics in no-star systems are rendered un-illuminated.  Who though about that.  Awesome.
- In real life, there are probably billions of no-star systems with enough mass to form gas giants and capture a satellite/planetary system, but not enough mass for fusion.

default/python/universe_generation/starsystems.py
LGM-Doyle: Since it was a double loop, I used `itertools.product` with `else:`.  I hope that is sufficiently elegant.

LGM-Doyle:
I've addressed all of the issues.

@MatGB, I think that empty black holes are an in game problem with multiple in game solutions.  

Personally, I keep a red dwarf on standby if I'm planning on using solar hulls.

default/python/universe_generation/starsystems.py
```diff
@@ -101,20 +111,16 @@ def generate_systems(pos_list, gsd):
...
-        ii_safety = 1000
-        while not at_least_one_planet and ii_safety > 0:
-            ii_safety -= 1
-            for orbit in orbits:
-                if planets.generate_a_planet(system, star_type, orbit, gsd.planet_density, gsd.shape):
-                    at_least_one_planet = True
-                    break
-
-        if not at_least_one_planet:
+        for _unused_recursion_guard, orbit in product(range(1000), orbits):
```
Cjkjvfnby: `_unused_recursion_guard` why not `_`?

MatGB:
I am about half way through a 350 system game on Low with this, and overall I like it, however I do suspect that with it there are too many planets on Low, we may need to adjust the Low numbers a bit but I'd like more feedback before we do.

Basically, if the code's good, I'm happy for this to go in with the note that we might need to tweak the settings to ensure Low stays Low. Because it's Python, I'm not just going to hit merge because, well, Python is only "easier to read than C" for me so far.

default/python/universe_generation/starsystems.py
Cjkjvfnby: For me `product` is less clear, since i need to check what does it do. But I don't insist to change it.

default/python/universe_generation/starsystems.py
LGM-Doyle: I wanted a name where it was clear why it was there and also clear why any code checkers were warning that it is unused.

Cjkjvfnby: PyCharm code checker does not triggered on `_`

default/python/universe_generation/starsystems.py
```diff
@@ -70,6 +70,16 @@ def name_planets(system):
...
 
 
+def can_have_no_planets(star_type):
+    """
+    Return True is system is allowed to have no planets.
+    This intentionally only checks the STAR_TYPE_MOD_TO_PLANET_SIZE_DIST so that
+    the following code forces most stars to have a planetary body so that in the
+    GUI the presence of a star implies a planet
+    """
+    return universe_tables.STAR_TYPE_MOD_TO_PLANET_SIZE_DIST[star_type][fo.planetSize.noWorld] != 0
```
Vezzra: Nifty, but also tricky - however, I think you should at least test for `> 0`, not `!= 0` (unless you want star types with _reduced_ probability for no planet also to be classified as stars that can have no planets, which strikes me as upside down :wink:).

I admit, basing the check if a star type should allow for an empty system on the probability for no planet in that universe table has a certain appeal. However, even checking for a probability `> 0` for no planet seems a bit... excessive? 

Currently, that would now include neutron stars, which have only a slightly increased probability for no planet. Which translates to "neutron stars should have less planets than normal stars" IMO, but not, like black holes, having no planets at all.

So maybe check for `> 20` or something like that?

default/python/universe_generation/starsystems.py
```diff
@@ -84,13 +94,33 @@ def generate_systems(pos_list, gsd):
...
+            if planets.generate_a_planet(system, star_type, orbit, gsd.planet_density, gsd.shape):
+                at_least_one_planet = True
+
+        if at_least_one_planet or can_have_no_planets(star_type):
+            continue
+
+        for _unused_recursion_guard, orbit in product(range(1000), orbits):
+            if planets.generate_a_planet(system, star_type, orbit, gsd.planet_density, gsd.shape):
+                at_least_one_planet = True
+                break
```
Vezzra, line 120: Just to be sure: does that check each orbit, and repeat that a 1000 times, or check one orbit a 1000 times before moving onto the next one?

If the latter, that wouldn't be quite optimal, because planet size is influenced by orbit (universe table `ORBIT_MOD_TO_PLANET_SIZE_DIST`), so one orbit being tried a 1000 times will prefer the probabilities of that orbit - we quit the moment a planet is generated, so the other orbits wouldn't get tried.

Oh, and setting `at_least_one_planet` to `True` isn't strictly necessary at that point, is it? After the refactoring of the double loop that flag doesn't get checked afterwards.

But probably can't hurt to set it nonetheless, less error prune in case that loop gets reworked at some later point I guess...

Vezzra:
> Because it's Python, I'm not just going to hit merge because, well, Python is only "easier to read than C" for me so far.

...aaaaaand because that PR is actually assigned to me, if I may be so bold. No PR nickin' here, ya hear me? 😉 

Me wanna do that merge 😎

default/python/universe_generation/starsystems.py
LGM-Doyle: `_` does not explain why this variable is unused.  It is not the typical reason of being an unused member of a tuple, so I made it a self-documenting variable name.  

I think that the code is clearer with it in.

default/python/universe_generation/starsystems.py
LGM-Doyle: I added a comment to `universe_tables.py` to clarify that a positive value in the none column means a chance of empty systems of that type.

The check should be > 0 and I fixed that.

The check should not be > 20 or anything other than 0.  Change the value in `universe_tables.py` for neutron stars to 0 if there should be no chance of neutron stars having no planets.

An "engine" should deal with the "source material", in this case `universe_tables.py` honestly.    

I left it as is, because the chief thing that it complicates is the acquisition of neutronium, which is a late game problem, whose solution is part of the "fun".

default/python/universe_generation/starsystems.py
LGM-Doyle, line 120: The inner loop is orbits.

However, earlier on I randomized orbits to guarantee that the single planet even in the single pass case will be in a random orbit.

I removed the redundant setting.

MatGB:
> that PR is actually assigned to me,

Hah, yeah, so it is, someday I'll learn to look at that column, but not this day.

OK, playtest feedback: on Low, it works fine, although I suspect we'll need to adjust some numbers as it does massively increase the number of non-empty systems in a way that makes it feel less 'low', that I suspect is a universe tables tweak.

On High, it does still create a fair number of starless systems with planets in, and while I'm very happy that such things probably exist in reality, I think it's unlikely that there'd be three in a 250 system galaxy relatively close to each other (seed 0, clusters, 250 systems). And regardless of likelihood, I think players will object, we've had gripes about starless planets in the past and I don't like to suspend disbelief too much if we can avoid it, even if the disbelief is probably unwarranted.

default/python/universe_generation/starsystems.py
Cjkjvfnby: `_` commonly used in two cases: 
- cycle variable is unused
- collection of fixed size unpacked to variables

I see typical case here. 

Can you explain about recursion here? for me it is `attempt_count`

Cjkjvfnby:
Should universe generation have different label on github than `scripting`

May be add code to print table of all systems with number of planets and stars?

default/python/universe_generation/starsystems.py
Vezzra: Hm, the original purpose of avoiding star systems (which are not deep space nodes) with no planets is to reduce annoying micromanagement (clicking through a lot of star systems to find the ones actually containing planets). Making an exception for black hole systems is reasonable, because they are sufficiently rare. Neutron stars, while still not as common as the other star types, are certainly more common than black holes.

But probably still uncommon enough so that probably won't be too much of a problem. That, and with the suggested comments in `universe_tables.py`, the fix for checking `> 0` I guess we can leave it at that. If things doesn't work well enough, we can make further adjustments later.

> An "engine" should deal with the "source material", in this case universe_tables.py honestly.

What exactly do you mean by that?

Vezzra:
@LGM-Doyle, did you push your latest changes? Maybe there is something malfunctioning on my end, but I don't see the `> 0` fix for the universe table check, nor the comment you said you added to `universe_tables.py` etc.

Vezzra:
@MatGB:

> on Low, it works fine, although I suspect we'll need to adjust some numbers as it does massively increase the number of non-empty systems in a way that makes it feel less 'low', that I suspect is a universe tables tweak.

Pretty much. With that PR going in, I guess we most likely need to readjust _all_ the numbers in the universe tables for the planet density setting (some a bit more, others probably not that much).

> On High, it does still create a fair number of starless systems with planets in, and while I'm very happy that such things probably exist in reality, I think it's unlikely that there'd be three in a 250 system galaxy relatively close to each other (seed 0, clusters, 250 systems).

Well, that actually is an unrelated issue, as this PR should not have any impact on what you witnessed here. And can probably also be addressed by tweaking the universe tables accordingly.

> And regardless of likelihood, I think players will object, we've had gripes about starless planets in the past and I don't like to suspend disbelief too much if we can avoid it, even if the disbelief is probably unwarranted.

In that case we should just set the numbers in the universe tables so that starless planets don't occur at all.

Vezzra:
@Cjkjvfnby:

> Should universe generation have different label on github than scripting

I thought about that too already. But then we'd also need a different label for turn events... maybe introduce a "python" tag (that could be used for both universe generation and turn events, but OTOH would overlap with AI of course, so maybe also not ideal...), and rename "scripting" to "FOCS"?

> May be add code to print table of all systems with number of planets and stars?

Why not, might help debug the relevant universe tables...

LGM-Doyle:
@MatGB thanks for the play-testing.

The problems with high should be fixed in the current version of the PR.

I just pushed a new version, unlike the other day when I commented on stuff, but didn't push any changes.

default/python/universe_generation/starsystems.py
LGM-Doyle: I've refactored this line so it should be a non-issue.

I think of the inner loop as a recursive function that keeps getting called until it generates a planet or it hits the recursion limit.

default/python/universe_generation/starsystems.py
LGM-Doyle: I mean that in the universe_table (source material), if someone sets the values so that 5% of the neutron stars have no planets, then the code (engine) should not have another secret threshold that prevents them from achieving their goal.  

Five percent should mean five percent not zero.

Change the universe_table to reflect the result that we want.

Vezzra: Ah, I see where you're coming from. Well, the values in the `STAR_TYPE_MOD_TO_PLANET_SIZE_DIST` table are modifiers to "dice rolls" for the planet size of each orbit, so the value in the "none" column are not the percent chance of a system being empty. Insofar I wouldn't necessarily expect a greater than zero value there to determine if a star system is allowed to be empty or not, so 0 is more or less as arbitrary a treshold for potentially empty systems as 20.

But that's a minor quibble. My preferred solution would be to decouple the "can be empty" setting for star types from the `STAR_TYPE_MOD_TO_PLANET_SIZE_DIST` table and have an extra table of boolean values for that. Just a vague idea, your solution should work well enough for now.

PR #719, opened 2016-06-18
Speedup redraw of TechTreeWnd when toggling techs.

LGM-Doyle:
In TechTreeWnd::LayoutPanel::Layout() creating the new TechPanel is time
consuming.

This commit compares the time of creation of the TechPanel with
CurrentTime() and keeps the old panel if it was created this turn.

### DIFF
--- a/UI/TechTreeWnd.cpp
+++ b/UI/TechTreeWnd.cpp
@@ -507,7 +507,9 @@ public:
     virtual void SizeMove(const GG::Pt& ul, const GG::Pt& lr);
 
     void Update();  ///< update indicated \a tech panel or all panels if \a tech_name is an empty string, without redoing layout
-    void Clear();                               ///< remove all tech panels
+    void Clear();
+    virtual void Show(bool children = true);    ///< remove all tech panels
+    virtual void Hide(bool children = true);    ///< remove all tech panels
     void Reset();                               ///< redo layout, recentre on a tech
     void SetScale(double scale);
     void ShowCategory(const std::string& category);
@@ -581,8 +583,8 @@ private:
     std::string             m_browsed_tech_name;
     TechTreeLayout          m_graph;
 
-    boost::unordered_map<std::string, TechPanel*>   m_techs;
-    int                     m_techs_update_turn;
+    boost::unordered_map<std::string, TechPanel*>
+                            m_techs;
     TechTreeArcs                        m_dependency_arcs;
 
     LayoutSurface* m_layout_surface;
@@ -1024,7 +1026,6 @@ TechTreeWnd::LayoutPanel::LayoutPanel(GG::X w, GG::Y h) :
     m_browsed_tech_name(),
     m_graph(),
     m_techs(),
-    m_techs_update_turn(CurrentTurn() - 1),
     m_layout_surface(0),
     m_vscroll(0),
     m_hscroll(0),
@@ -1166,21 +1167,13 @@ void TechTreeWnd::LayoutPanel::Clear() {
     GG::SignalScroll(*m_vscroll, true);
     GG::SignalScroll(*m_hscroll, true);
 
-    // Delete old all panels and hide all current panels.
-    if (m_techs_update_turn != CurrentTurn()) {
-        for (boost::unordered_map<std::string, TechPanel*>::const_iterator panel_it = m_techs.begin()
-                 ; panel_it != m_techs.end(); ++panel_it) {
-            delete panel_it->second;
-        }
-        m_techs.clear();
-
-    } else {
-        for (boost::unordered_map<std::string, TechPanel*>::const_iterator panel_it = m_techs.begin()
-                 ; panel_it != m_techs.end(); ++panel_it) {
-            panel_it->second->Hide();
-            if (GG::Wnd * parent = panel_it->second->Parent())
-                parent->DetachChild(panel_it->second);
-        }
+    //Hide all current panels.
+    for (boost::unordered_map<std::string, TechPanel*>::const_iterator panel_it = m_techs.begin();
+         panel_it != m_techs.end(); ++panel_it)
+    {
+        panel_it->second->Hide();
+        if (GG::Wnd* parent = panel_it->second->Parent())
+            parent->DetachChild(panel_it->second);
     }
 
     m_graph.Clear();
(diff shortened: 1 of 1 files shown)


UI/TechTreeWnd.cpp
```diff
@@ -579,7 +581,8 @@ class TechTreeWnd::LayoutPanel : public GG::Wnd {
     std::string             m_browsed_tech_name;
     TechTreeLayout          m_graph;
 
-    std::map<std::string, TechPanel*>   m_techs;
+    boost::unordered_map<std::string, TechPanel*>   m_techs;
+    int                     m_techs_update_turn;
```
geoffthemedio: make indentation consistent for adjacent line declarations

UI/TechTreeWnd.cpp
```diff
@@ -1156,11 +1166,25 @@ void TechTreeWnd::LayoutPanel::Clear() {
     GG::SignalScroll(*m_vscroll, true);
     GG::SignalScroll(*m_hscroll, true);
 
-    // delete all panels
-    for (std::map<std::string, TechPanel*>::const_iterator it = m_techs.begin(); it != m_techs.end(); ++it)
-        delete it->second;
-    m_techs.clear();
+    // Delete old all panels and hide all current panels.
+    if (m_techs_update_turn != CurrentTurn()) {
+        for (boost::unordered_map<std::string, TechPanel*>::const_iterator panel_it = m_techs.begin()
```
geoffthemedio:
```
for (long<declaration stuff> varname;
     varname < something; ++varname)
{
}
```

UI/TechTreeWnd.cpp
```diff
@@ -1156,11 +1166,25 @@ void TechTreeWnd::LayoutPanel::Clear() {
...
+                 ; panel_it != m_techs.end(); ++panel_it) {
+            delete panel_it->second;
+        }
+        m_techs.clear();
+
+    } else {
+        for (boost::unordered_map<std::string, TechPanel*>::const_iterator panel_it = m_techs.begin()
+                 ; panel_it != m_techs.end(); ++panel_it) {
+            panel_it->second->Hide();
+            if (GG::Wnd * parent = panel_it->second->Parent())
```
geoffthemedio: remove space between Wnd and *

geoffthemedio:
I don't like the tests against current turn, as it likely has edge cases you've not considered. Why not just delete all the panels at the end of each turn and clear the whole container of them, and recreate them if they don't already exist whenever each needs to be rendered or the tree layout is redone?

LGM-Doyle:
I assumed for the `for` that you did not want the `!=` replaced with a `<`.

Regarding the test against current turn, this code effectively deletes all the panels each turn.

The previous code was deleting all of the panels every time the layout was updated.  That was what caused the slowness.  This code now checks the current turn during that same call to clear every time that the layout is drawn.  So unless the turn can end without CurrentTurn changing this code will delete all the panels every time the layout is drawn after the turn changes.    

I could instead look for a turn end signal and delete the TechPanels while handling turn end.  However, m_graph and m_dependency arcs depend on the clear called each layout, which would then only be a partial clear.  

Neither solution is ideal.

geoffthemedio:
Not sure what you're referring to about != vs <. If it's the code bit I commented with,

```
for (long<declaration stuff> varname;
     varname < something; ++varname)
{
}
```

the point was just the formatting of the for loop structure.

geoffthemedio:
I know what you've changed and why from the initial description; my comment about checking the turn value  stands: it probably has edge cases that aren't considered, so just avoid doing it like that, and explicitly do the clearing of the panels as part of the turn cycling of the UI, and just recreate them whenever they need to be rendered if they don't already exist.

I don't understand the relevance of what you wrote about m_graph and m_dependency though. This is, I thought, about not deleting and recreating the panels.

LGM-Doyle:
I understand what you are saying about edge cases.  Neither, solution guards against edge cases where a Tech changes during a turn.  Provided that `CurrentTurn()` tells the truth both solutions update each turn.  

What I mean about `m_graph` and `m_dependency` is that I did not go through the `m_graph` and `m_dependency` code to see if they need to be cleared every time before re-layout.  So it might be possible to move all of the code in clear() into an end of turn handler, but that is more change than needed to fix this problem.

Right now all of the code related to the layout is cleared or not cleared in one function.  It is trivial to see how the timing of that clearing is related to the regeneration of the layout.  Moving some but not all of the code to an end of turn handler makes that timing less clear, if someone is fixing a future bug.

I can re-work the solution into an end of turn handler.  I'm not defending this solution as elegant.  It is just the smallest change that fixes the laggy re-draw problem.

geoffthemedio:
There could be situations where the tech tree changes but the turn hasn't, or the turn you're storing isn't valid, so don't store the turn. If once-per-turn refreshes are wanted / needed, then do that as part of the turn refresh.

LGM-Doyle:
I've moved the panel deleting code into the Hide() and Show() function son the assumption that the tech panels do not change while the tech winow is visible

geoffthemedio:
They could change if the turn cycles while the tech window is visible, I think?

LGM-Doyle:
@geoffthemedio I'm going to drop this PR.  

I apologize with taking up your time reviewing a PR that didn't and doesn't work.

geoffthemedio:
The idea seemed reasonable; what's the problem?

LGM-Doyle:
The essential problem is that there is no push behavior for the effects code.  

The only way to determine if a tech has changed  is to run TechPanel::Update() for every visible tech, which is what is consuming the time.

Since techs can change in the middle of the turn and while the window is visible if it spans a turn change, what I need is a signal "TECH_<id>_HAS_CHANGED" to know to update only that tech box.

If there is a signal like that, then I can refactor this to use that signal and also the knowledge of what techs are selected for the list box (I noticed that I botched that as well) to only update the tech boxes that are changed.

geoffthemedio:
The proposed changes wouldn't make during-the-turn changes any worse; it would already be an issue, and not a reason to avoid optimizing where possible.

LGM-Doyle:
The reason it is currently, sluggish is that it polls every visible tech anytime anything changes in the tech tree.  This also avoids the during-the-turn-changes (mostly) because any change the user makes will update the tech tree.

I'd rather drop it and think of a more correct solution.  

A more correct solution entails a signal to indicate when a tech has changed so that only that tech panel needs to be updated.

PR #720, opened 2016-06-18
Fix regression where research window hotkey zoom also zooms map window.

LGM-Doyle:


UI/MapWnd.cpp
```diff
@@ -6119,19 +6119,19 @@ void MapWnd::ConnectKeyboardAcceleratorSignals() {
     hkm->Connect(this, &MapWnd::ShowMenu,               "map.menu",
                  new AndCondition(new VisibleWindowCondition(this), new NoModalWndsOpenCondition()));
     hkm->Connect(this, &MapWnd::KeyboardZoomIn,         "map.zoom_in",
-                 new AndCondition(new VisibleWindowCondition(this), new NoModalWndsOpenCondition()));
+                 new AndCondition(new VisibleMapWndCondition(*this), new NoModalWndsOpenCondition()));
```
geoffthemedio, line 6122: The signature is

```
VisibleWindowCondition(const GG::Wnd* tg)
```

Changing to *this would be passing a MapWnd& reference instead. What am I missing here?

LGM-Doyle:
Different function.  VisibleMapWndCondition != VisibleWindowCondition.

VisibleMapWndCondition checks to see if the map window is covered by the fullscreen windows: ResearchWnd or DesignWnd.

It prevents the map from zooming when it is not visible.

geoffthemedio:
Maybe that condition should be renamed then? It's not doing the same test as other visible window conditions, which actually poll the Visible member of the indicated Wnds...

LGM-Doyle:
How about `OccludedMapWndCondition()` ?

geoffthemedio:
It'd have to NotOccluded... wouldn't it?

PR #726, opened 2016-06-22
only add test subdirectory if BUILD_TESTS is true

vincele:
Signed-off-by: Vincent Legoll vincent.legoll@gmail.com


geoffthemedio:
sounds reasonable, looks relatively harmless...

PR #727, opened 2016-06-22
Add missing .clear() call

vincele:
Signed-off-by: Vincent Legoll vincent.legoll@gmail.com


vincele:
If this is not missing, at least a big fat comment should be put there explaining why...

Issue #462, opened 2016-02-03
AI client(?) creates default_config.ini in a potentially non writable location.


adrianbroher:
Some part of the application writes a file called `default_config.ini` in the `default/AI` subdirectory.

I would like to point out that the whole 'default' subdirectory is usually not writable for users of the game.  Think of FreeOrion installations in the `/usr/`, `/application/`, `C:\Program Files` or similar directories.  Those directories are only writable for super users (root, Administrator) during installation time of the game.  When you need to write configuration data please use a user writable directory.

 The c++ `GetUserDir`function located in `util/Directories.cpp` gives you a path, that is intended for user specific configuration and save data, so please use this function and its Python binding (if available).

Cjkjvfnby:
There is not python binding for `GetUserDir`
If someone add binding for it I will make python part.

Binding should looks like this

``` c++
def("getUserDir", Directories.GetUserDir, return_value_policy<return_by_value>(), "Return path to user directory, AI have write permission to this directory.");
```

Vezzra:
Will try to get to that tomorrow.

Vezzra:
@Cjkjvfnby, implementation of Python binding for GetUserDir committed: 162e8cd92729789da6f0881952276e852e292bde. Had to do it a bit differently than you suggested (for whatever reason returning the path to Python as std::string crashes the AI client, returning as boost::python:str however works).

You can go ahead and implement the required adjustments on the Python side. Re-assigned this issue to you.

Vezzra:
On a side note: when storing the AI config file (`default_config.ini`) in this folder for user specific data, please rename it to something that indicates it's an **AI** config file. That will make its purpose clearer, as that location isn't for AI data only.

Cjkjvfnby:
@Vezzra can you merge `python/AI/AIWrapper.cpp` changes to master?

Vezzra:
@Cjkjvfnby: um, I already have? Commit 162e8cd92729789da6f0881952276e852e292bde has already been pushed to master... I can't see anything amiss?

Cjkjvfnby:
My bad.

Cjkjvfnby:
I try to ensure that only one AI write config file and this required a lot of changes.

We initialize python in 2 stages:
- import (and execute some code including config creation) 
- prepare AI in c++ code and call `initFreeOrionAI`

Can we prepare AI first and only after that import python libs?  This will make `initFreeOrionAI`  obsolete and make things less complicated.

[...]

PR #728, opened 2016-06-23
Don't ignore user specified config

Cjkjvfnby:
Fix issue introduced in fix for #462


dbenage-cx:
@Cjkjvfnby no issues here

PR #729, opened 2016-06-24
Replace icons paths in some Pedia articles to be the same icons than on the…

Ouaz:
… main interface

For Design Window, Fleet Movement, Greetings, Hidden Settings, Quick Start articles


Vezzra:
Anyone any objections to merging this?

MatGB:
Looks very sensible to me, if we can have different icons, actually doing so makes sense.

Ouaz:
@ spikethehobbitmage

The last time, you found that my commit changed the line endings (CR+LF instead of LF only)

> https://github.com/freeorion/freeorion/pull/679

Weird, because before commiting, I checked that lines endings were fine (LF in Notepad2 or Unix in Notepad++). Same for this commit. Can you check though?

Thanks.

PR #730, opened 2016-06-24
Improve CMake status messages.

LGM-Doyle:
Added two messages to build.

Build type CMAKE_BUILD_TYPE set to ${CMAKE_BUILD_TYPE}
Building tests BUILD_TESTS set to ${BUILD_TESTS}

This is just so that it is clear to me when I'm building with the tests on or off.


geoffthemedio:
looks simple enough; hopefully @adrianbroher will object if it's problematic.

Issue #631, opened 2016-04-30
[AI] Tweak Drydock Repair logic


Morlic-fo:
As discussed in comments in #554, changes probably need further enhancement.

Morlic-fo:
#732 seems to have covered all the urgent issues. Thanks @LGM-Doyle.

Issue #633, opened 2016-04-30
AI research path needs to be checked and updated


Morlic-fo:
With so many changes to the game since the last major overhaul, research choices and priorities need to be checked and adapted.

If you notice weird behaviour or have general ideas on how to improve the AI by getting different techs earlier/later, feel free to speak up.

dbenage-cx:
A few thoughts:
- Microgravity and Orbital Construction are not gated.  These should be de-prioritized, at least until a suitable system is selected to acquire. (single AI, no natives/monsters, will still research these after Algo Elegance regardless of not colonizing systems with either)
- Weapon upgrades should have lower priority if no threat detected (part of other issue). Parts > upgrades in 'peace-time'.
- Possibly have each AI focus toward a single ship type.  This is both to help decrease some production loss from shipyard spam, and provide some flavor/consistency.  Delay the decision until turn 20-30+, then base off the potential systems.

Cjkjvfnby:
We already developed other way of research, which is turn off by default.  This ticket looks like improvement of current one. I am confused.

Morlic-fo:
The research path needs to be updated before 0.4.6 release. I don't expect the new research system to be in an acceptable state at that point in time. So for 0.4.6 the classical/hardcoded approach will be used and thus its research path needs to be updated.

The other issue deals with the new research approach and is for the time after 0.4.6 (and has the according milestone).

Vezzra:
@Morlic-fo, @Cjkjvfnby, aside from the issues concerning building FO on Linux with Boost 1.60/1.61, that one here is maybe the last potentially critical, and the last one concerning the AI. How important do you consider this having to be addressed for the release, at least with some quick temporary adjustments?

Or, going the other way round: how likely is it for you being able to come up with something that can address this, within the next two weeks?

I need to know if I can remove the "Release v0.4.6" milestone, or if you consider it critical for the release.

Cjkjvfnby:
Next week I will make some research history. After that I will need some help to analyze it.

Vezzra:
Good. Keep in mind that you need to be finished by Sept 4th at the latest. Earlier would be better, but only if that's not too much for @Morlic-fo and you.

Worst case we need to do without that adjustment, but that's something I want to avoid of course.

Cjkjvfnby:
Here is some research table for AI (turn, research started, research finished, research in progress)
It is generated by code in this branch  https://github.com/Cjkjvfnby/freeorion/pull/16

[...]

Issue #632, opened 2016-04-30
[AI] Weapon Upgrades aren't fully supported yet


Morlic-fo:
Some of the code in ShipDesignAI.py needs to update its calculation.
In particular, the species upgrades are wrong.

Considering future potential upgrades when designing/building ships also is a good idea.

Morlic-fo:
Covered needed changes for 0.4.6 in #814

Issue #630, opened 2016-04-30
Adaptations to new supply propagation required


Morlic-fo:
The AI doesn't understand the supply propagation changes introduced in #488.

Morlic-fo:
With #896, this is no longer critical for release.

PR #732, opened 2016-06-24
Add checks for safe passage to drydocks in AI.

LGM-Doyle:
This expands checks for drydocks from current happiness and distance to
include all of:
- current happiness
- target happiness
- distance in hops
- safe passage by comparing fleet rating to the rating of the entire
  path.

It fixes issue #631.

### DIFF
--- a/default/python/AI/MoveUtilsAI.py
+++ b/default/python/AI/MoveUtilsAI.py
@@ -1,3 +1,4 @@
+import sys
 import freeOrionAIInterface as fo  # pylint: disable=import-error
 import FreeOrionAI as foAI
 import AIstate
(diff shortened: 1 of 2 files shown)


default/python/AI/MoveUtilsAI.py
```diff
@@ -164,37 +165,67 @@ def get_nearest_supplied_system(start_system_id):
...
     drydock_system_ids = set()
     for sys_id, pids in ColonisationAI.empire_dry_docks.iteritems():
+        if sys_id == -1:
+            print >> sys.stderr, "get_best_drydock_system_id passed bad dry dock sys_id."
+            continue
         for pid in pids:
             planet = universe.getPlanet(pid)
-            if planet and planet.currentMeterValue(fo.meterType.happiness) >= 5:
+            if (planet
+                  and planet.currentMeterValue(fo.meterType.happiness) >= HAPPINESS_THRESHOLD
```
Cjkjvfnby, line 200: Should we check only target happiness?

default/python/AI/MoveUtilsAI.py
```diff
@@ -164,37 +165,67 @@ def get_nearest_supplied_system(start_system_id):
...
+        print >> sys.stderr, "get_best_drydock_system_id passed bad fleet id."
+        return None
+
+    HAPPINESS_THRESHOLD = 5
+
     universe = fo.getUniverse()
+    start_sys = universe.getSystem(start_system_id)
     drydock_system_ids = set()
     for sys_id, pids in ColonisationAI.empire_dry_docks.iteritems():
+        if sys_id == -1:
```
Cjkjvfnby, line 194: Any reason why you add extra check?  Should `ColonisationAI.empire_dry_docks` be free from `-1`

Vezzra:
@Cjkjvfnby, the corresponding issue #631 is flagged as "enhancement", not "bug", that should be the same. As #631 addresses a required adjustment of the AI to changed game mechanics, IMO that isn't really a bug, so I think the "enhancement" label fits better...?

Cjkjvfnby:
@Vezzra, this PR fixes AI issue that was added by changes in core mechanics. 
Labels like `bug` and `enchantment` does not affect my workflow.

Vezzra:
@Cjkjvfnby, yes, I know, this is more of an administrative nitpick...

default/python/AI/MoveUtilsAI.py
LGM-Doyle, line 194: No particular reason.

However, while I was working on cleaning up the python when removing the empty exception handlers in PR #685, some lists (I'm not exactly sure which) were carrying around and passing `-1` system ids into the C++ path finding functions.

So some code somewhere is using `-1` in lists of system ids.

default/python/AI/MoveUtilsAI.py
LGM-Doyle, line 200: Ideally, we check happiness at the expected arrival time and for the expected duration of the repairs.  

This version means the AI targets planets that are happy now and will be happy when it arrives.

It doesn't get trapped sitting at a planet that will be happy enough in 5 turns, but for now is a death trap.

Cjkjvfnby:
I totally forget about this pr, looks like it is ready to be merged. @LGM-Doyle?

LGM-Doyle:
Yes, I think that it is ready to be merged.

Thank you for doing the review.

Vezzra:
Did this PR address #631 fully, so that this issue can be closed, or is further work required on #631?

LGM-Doyle:
I think that it fully addresses issue #631.

However @Morlic-fo created the issue, so only @Morlic-fo can say it certainly addresses everything that they had in mind.

Speaking of @Morlic-fo , issues #630, #632 and #633 were created by @Morlic-fo and are marked for v0.4.6, but no one appears to be handling them.  If no one picks them up what is your plan?

Vezzra:
I sincerely hope that it won't come to this, but if indeed no one picks up those issues we might have to just make the release without those AI adjustments. That's a "worst case" scenario, as it would result in a release with an impaired AI.

That's a decision I won't make on my own, this certainly would be a subject of discussion. Considering that we've already July, we probably need to have that discussion in the not too far future...

The heart of the problem is that our lead AI dev @Dilvish-fo has been absent since last September/October, @Morlic-fo hasn't been around much recently too, the only remaining AI dev is @Cjkjvfnby. We're critically low on AI dev manpower...

PR #738, opened 2016-06-26
State

Cjkjvfnby:
I extract module level variables and code to set and get them to single module.

Next step will be extracting `ColonizationAI.survey_universe` to method of this module. 

I want this to be merged in this release.

I did not assign this to exact person but want to hear all @freeorion/ai-team comments.

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -10,6 +10,7 @@ import PlanetUtilsAI
 import ProductionAI
 import TechsListsAI
 import MilitaryAI
+from turn_state import state
 from EnumsAI import MissionType, FocusType, EmpireProductionTypes, ShipRoleType, PriorityType
 from freeorion_tools import dict_from_map, tech_is_complete, get_ai_tag_grade, cache_by_turn
 from freeorion_debug import Timer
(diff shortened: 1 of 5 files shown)


Vezzra:
Well, that is just @Morlic-fo (besides yourself), as @Dilvish-fo is still MIA...

LGM-Doyle:
As always the code is elegantly done.

I like the getters and setters.

I'm concerned with moving the variables out of ColonisationAI.

With the variables in ColonisationAI, there is better encapsulation.  It is clear that ColonisationAI controls and is responsible for those variables.  In a global state with many variables it is no longer clear which module controls which variable.

Why not put the getters and setters in ColonisationAI?

It would prevent other modules from interfering with the variables and provide better encapsulation.

Either way the code looks well written to me.

Cjkjvfnby:
Thank you.

> Why not put the getters and setters in ColonisationAI?

Because they are not belong to colonization, whole `ColonizationAI.survey_universe` is not belong to colonization.  

> It would prevent other modules from interfering with the variables and provide better encapsulation.

This variables are note belong to single module.

LGM-Doyle:
Sorry.

I thought that all of the setting was being done in one file and the getting was being done in many files.

Encapsulation seemed natural from the code.

Issue #715, opened 2016-06-18
Move population effects from species/common/population.macros when possible


MatGB:
The icons in the tech tree are generated automatically from the techs: [Post](http://www.freeorion.org/forum/viewtopic.php?p=84454#p84454). They were previously all grouped together in species.txt in order that they could be applied in the correct order, now we have an effects priority system and this has been applied to all population effects it is no longer needed for them to be where they are and it would be better for them to be in their originating tech.

Some effects/macros may need to remain where they are if shared between multiple kinds of scripted content, but if at all possible, especially for techs, they should be moved either to the source itself or to a macro file in that folder.

geoffthemedio:
(edited your description a bit)

MatGB:
Fine by me, that's clearer and I was struggling to word it anyway, thanks.

dbenage-cx:
Some planet shields, defense and troop tech effects could move over with priorities assigned.  Some specials and other effects (homeworld) could be shifted as well.

geoffthemedio:
I'd like to encourage attention to this issue. Particularly for population macros that are applied to species, where part of the activation is whether the owner has a particular tech, I suspect that the current arrangement is a major contributor to effects processing time increasing with number of ships or planets.

https://github.com/freeorion/freeorion/blob/master/default/scripting/species/common/population.macros#L104

As of now, with these macros in place, the game has to evaluate dozens of effect activation conditions for every object that has a species attached, including all these population effectsgroups for all the ships, which will never be acted on by them. I suspect it would be a lot faster to test the activation and evaluate the scope conditions once per empire, rather than once per ship or planet, as would be the case of they were moved into the tech being tested for in the current activation conditions.

dbenage-cx:
@geoffthemedio Is this issue now `won't fix`?  I'm inclined to close the PR, but don't want to appear upset.

geoffthemedio:
No? I just explained why it's worth doing. Why should / would it be marked won't fix or closed?

dbenage-cx:
Ok, i took your statement to mean that any move into tech would now result in a less efficient process.  Will look into reworking scope/activation.

geoffthemedio:
I don't know what you mean by "reworking scope/activation"... The suggested fix is to move tech effects into techs instead of having them in species.

MatGB:
@dbenage-cx that was my initial reaction on first read of @geoffthemedio 's comment as well, but on reread I worked out he meant the exact opposite.

[...]

PR #741, opened 2016-06-29
FOCS: Move defense effects from species

dbenage-cx:
Moves planetary defense, troops, and shield effects from species.

Tag as work towards #715

### DIFF
--- a/default/scripting/buildings/IMPERIAL_PALACE.focs.txt
+++ b/default/scripting/buildings/IMPERIAL_PALACE.focs.txt
@@ -94,12 +94,19 @@ BuildingType
                 SetTargetHappiness value = Value + 5 // min(10, max(0, 10 - Target.DistanceToSource/20))
         
         EffectsGroup
-            scope = Object id = Source.PlanetID
+            scope = And [
+                Object id = Source.PlanetID
+                Planet
+                OwnedBy empire = Source.ProducedByEmpireID
+            ]
             stackinggroup = "PALACE_DEFENSE_STACK"
-            priority = [[EARLY_PRIORITY]]
-            effects = SetMaxDefense value = Value + 5 accountinglabel = "BLD_IMPERIAL_PALACE"
-        ]
-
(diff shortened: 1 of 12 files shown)


MatGB:
Something's wrong with this, I'm getting zero shields on a lot of AI worlds, and only getting Barrier Shield 1 on my homeworld, should be getting Force Field Dynamics and the basic +1 from Self Defence (which ought actually to be removed for players that've got FFD but that's not been discussed as planned).

I suspect I'm ahead on research than my two Egassem neighbours which means Shield1 is working but the earlier two aren't.

dbenage-cx:
@MatGB Should have caught that, took forever to figure out what was wrong.

Not sure how far you are on stringtable updates, would you want a PR for updates on these entries? (DEF_ROOT_DEFENSE mentions troops and not shield).

MatGB:
We are, as a team, perpetually forgetting to update stringtables, anytime you catch something a quick fix is always welcome—I'm especialyl bad at remembering the unlocked-at-start stuff as I never look at it while playing.

I haven't yet started on the full review of the stringtable, I'm behind on the balancing work I wanted in already, I normally do a stringtable review after we've closed off for new work which is delayed a bit now.

geoffthemedio:
Are there any remaining issues with these scripting changes?

dbenage-cx:
Just verification from someone other than me ;)

Vezzra:
Well, @MatGB has taken responsibility for this PR, so we're more or less waiting for him I guess...

MatGB:
I had missed the fixes had been done before going away, I'll be able to test it over the weekend but if anyone else is able to today or tomorrow that'd be fine, it looks fine but the last bug was a weird one

MatGB:
K, tested on a few savegames and a quick playthrough to a reasonable level, appears to work fine so merging.

Hard to be sure, but fairly certain there's a significant lag reduction in the mid to late game from it, as predicted, which is nice.

dbenage-cx:
Ty MatGB, I'll get the last of it rebased and tested.

PR #742, opened 2016-07-02
Fix bug python system exit not an error

LGM-Doyle:
This PR addresses issue #524.

It makes handling of python exceptions more consistent, by handling them in a single function and allowing the exceptions to propagate to the level where they can be handled completely.

It creates HandleErrorAlreadySet() which differentiates between exception which are errors and SystemExit types of python exceptions, which are not errors and indicate that the python interpreter has shutdown.  

The errors are printed out with PyErr_Print() and in cases outside of the initialization, the execution loop continues after the error is logged.

All cases of SystemExit exceptions and python errors during initialization are fatal and cause the freeorion server to exit.

The only cases that I know of that can still cause the freeorion server to hang are using the python os module to merge threads that have not been forked, or to cause an exit via terminate.  I don't know how to prevent these.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -5,7 +5,6 @@ import sys
 import random
 
 from common import configure_logging
-configure_logging.redirect_logging_to_freeorion_logger()
 
 # IMPORTANT! this import also execute python code to update freeOrionAIInterface interface,
 # removing this import will brake AI in unexpected way.
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -4,6 +4,9 @@ import pickle  # Python object serialization library
 import sys
 import random
 
+from common import configure_logging
+configure_logging.redirect_logging_to_freeorion_logger()
+
 # IMPORTANT! this import also execute python code to update freeOrionAIInterface interface,
 # removing this import will brake AI in unexpected way.
 
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -1,0 +1,26 @@
+import sys
+import logging
+import freeorion_logger
+
+
+class dbgLogger(object):
+    """A stream-like object to redirect stdout to C++ process"""
+    @staticmethod
+    def write(msg):
+        freeorion_logger.log(msg)
+
+
+class errLogger(object):
+    """A stream-like object to redirect stderr to C++ process"""
+    @staticmethod
+    def write(msg):
+        freeorion_logger.error(msg)
+
+
+def redirect_logging_to_freeorion_logger():
+    """Redirect stdout, stderr and the logging.logger to hosting process' freeorion_logger"""
+
+    sys.stdout = dbgLogger()
+    sys.stderr = errLogger()
+    print 'Python stdout and stderr redirected'
+
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -1,16 +1,15 @@
 import sys
-import logging
 import freeorion_logger
 
 
-class dbgLogger(object):
+class DbgLogger(object):
     """A stream-like object to redirect stdout to C++ process"""
     @staticmethod
     def write(msg):
         freeorion_logger.log(msg)
 
 
-class errLogger(object):
+class ErrLogger(object):
     """A stream-like object to redirect stderr to C++ process"""
     @staticmethod
     def write(msg):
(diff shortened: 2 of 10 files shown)


Vezzra:
@LGM-Doyle, looks like I've to get back to all the Python/Python interface stuff PRs you're providing - _somehow_ I've to find the time for it, things are piling up... 😉

default/python/common/configure_logging.py
```diff
@@ -0,0 +1,26 @@
...
+        freeorion_logger.error(msg)
+
+
+def redirect_logging_to_freeorion_logger():
+    """Redirect stdout, stderr and the logging.logger to hosting process' freeorion_logger"""
+
+    sys.stdout = dbgLogger()
+    sys.stderr = errLogger()
+    print 'Python stdout and stderr redirected'
+
```
Vezzra, line 25: While we are at it, we can adjust the names of the logger classes to PEP8 (`DbgLogger` and `ErrLogger`).

default/python/common/configure_logging.py
```diff
@@ -0,0 +1,26 @@
+import sys
+import logging
```
Vezzra: This module doesn't get used here, or am I missing something obvious?

default/python/AI/FreeOrionAI.py
```diff
@@ -4,6 +4,9 @@
 import sys
 import random
 
+from common import configure_logging
+configure_logging.redirect_logging_to_freeorion_logger()
```
Vezzra: Hm, can we move that line into `configure_logging.py`? If it gets executed immediately after import everywhere we use `configure_logging.py` anyway, it can be placed in the module to be executed on import automatically as well.

default/python/universe_generation/universe_generator.py
```diff
@@ -1,5 +1,8 @@
 import random
 
+from common import configure_logging
+configure_logging.redirect_logging_to_freeorion_logger()
```
Vezzra: Same as in `FreeOrionAI.py`.

python/CommonFramework.h
```diff
@@ -15,20 +15,43 @@ class PythonBase {
...
+    void         ExecScript(const std::string script); // executes a Python script or throws error_already_set
+    void         SetCurrentDir(const std::string dir); // sets Python current work directory or throws error_already_set
+    void         AddToSysPath(const std::string dir);  // adds directory to Python sys.path or throws error_already_set
     void         SetErrorModule(boost::python::object& module); // sets Python module that contains error report function defined on the Python side
     std::vector<std::string> ErrorReport();            // wraps call to error report function defined on the Python side
 
 private:
+    // A copy of the systemExit exception to compare with returned
+    // exceptions.  It can't be created in the exception handler.
+    boost::python::object systemExit;
```
Vezzra: I know it's just a helper object, but IMO it should still conform to standard naming style: `m_system_exit`. Or `m_system_exit_exc`, or whatever you like.

python/CommonFramework.cpp
```diff
@@ -63,93 +65,83 @@ bool PythonBase::Initialize()
...
+void PythonBase::SetCurrentDir(const std::string dir) {
     std::string script = "import os\n"
     "os.chdir(r'" + dir + "')\n"
     "print 'Python current directory set to', os.getcwd()";
-    if (!ExecScript(script)) {
-        ErrorLogger() << "Unable to set Python current directory";
-        return false;
-    }
-    return true;
+    exec(script.c_str(), m_namespace, m_namespace);
```
Vezzra, line 134: Why don't use the `ExecScript` member function? That's the point of having it after all. If you use `exec` directly here and in `AddToSysPath`, it's actually obsolete and can be removed.

server/ServerApp.cpp
```diff
@@ -1283,9 +1299,27 @@ void ServerApp::GenerateUniverse(std::map<int, PlayerSetupData>& player_setup_da
...
+    bool success(false);
+    try {
+        m_python_server.SetCurrentDir(GetPythonTurnEventsDir());
+        // Call the main Python turn events function
+        success = m_python_server.ExecuteTurnEvents();
+    } catch (boost::python::error_already_set err) {
+        success = false;
+        m_python_server.HandleErrorAlreadySet();
+        if (!m_python_server.IsPythonRunning()) {
+            ErrorLogger() << "Python interpreter is no longer running.  Attemting to restart.";
```
Vezzra: Typo: Attem**p**ting

LGM-Doyle:
Fixed and rebased.

Vezzra:
Thx!

default/python/AI/FreeOrionAI.py
Cjkjvfnby: And PyCharm mark it as unused import.

Cjkjvfnby:
How about `turn_events`?

Vezzra:
What do you mean?

PR #744, opened 2016-07-03
improve sidepanel/scrollbar behaviour

TheSilentOne1:
- use whole sidepanel width for planet panels when scrollbar is hidden
- hide scrollbar when system is empty/unexplored
- remove scrollbar position inaccuracy

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -568,6 +568,8 @@ public:
 
     void            RefreshAllPlanetPanels();           //!< updates data displayed in info panels and redoes layout
 
+    virtual void    ShowScrollbar();
+    virtual void    HideScrollbar();
     virtual void    SizeMove(const GG::Pt& ul, const GG::Pt& lr);
 
     /** Enables, or disables if \a enable is false, issuing orders via the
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I have a minor usability concern, which is that if expanding or collapsing a panel (to show/hide meters/icons) causes the scrollbar to be shown or hidden, then the button to expand/collapse moves as a result, making it a bit awkward to click multiple buttons or on the same button multiple times.

TheSilentOne1:
I see your point, but I wouldn't think it's a concern. At least I don't switch forth and back between expand/collapse often, and usually there's enough space on the sidepanel anyway. And even if a "jump" occurs, I didn't find it irritating. But that's me, so maybe someone else wants to give this a try.

Vezzra:
I guess this is intended for inclusion into 0.4.6?

TheSilentOne1:
It's only a small change, so I'd say yes.

UI/SidePanel.cpp
```diff
@@ -2975,6 +2991,13 @@ void SidePanel::DoLayout() {
     lr = GG::Pt(ClientWidth() - 1, ClientHeight() - GG::Y(INNER_BORDER_ANGLE_OFFSET));
     m_planet_panel_container->SizeMove(ul, lr);
 
+    // hide scrollbar if there is no planets in the system
+    if (TemporaryPtr<const System> system = GetSystem(s_system_id))
+        if (system->PlanetIDs().size() == 0)
```
geoffthemedio: prefer !system->PlanetIDs().empty() ?

UI/SidePanel.cpp
```diff
@@ -2531,6 +2539,14 @@ void SidePanel::PlanetPanelContainer::RefreshAllPlanetPanels() {
         (*it)->Refresh();
 }
 
+void SidePanel::PlanetPanelContainer::ShowScrollbar(){
```
geoffthemedio: for single line functions, do like

```
void SidePanel::PlanetPanelContainer::HideScrollbar()
{ DetachChild(m_vscroll); }
```

UI/SidePanel.cpp
```diff
@@ -2531,6 +2539,14 @@ void SidePanel::PlanetPanelContainer::RefreshAllPlanetPanels() {
...
+	DoPanelsLayout();
+}
+
+void SidePanel::PlanetPanelContainer::HideScrollbar(){
```
geoffthemedio: If this did have a { at the end of a line, it should have a space between ) and {

UI/SidePanel.cpp
```diff
@@ -2403,30 +2405,36 @@ void SidePanel::PlanetPanelContainer::DoPanelsLayout() {
 
     GG::Y starting_y = y;
 
+    // how much vertical space will be consumed and how much is available?
+    GG::Y available_height = Height();
+    GG::Y used_height = GG::Y(0);
+
+    for (std::vector<PlanetPanel*>::iterator it = m_planet_panels.begin(); it != m_planet_panels.end(); ++it) {
```
geoffthemedio: I'd split this for over two lines (and put { on the third line)

PR #747, opened 2016-07-05
Change not auto fleet scouts and cols

LGM-Doyle:
Please use the following as the merge message.

---------------------snip --------------------------------------
1. Names fleets based on their function: Colony, Recon, Troop, Bomber, Combat and Herd.
2. Prevent non-combat fleets from automatically merging.  Colony and Recon ships are immediately useful for sending to new planets.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -148,6 +148,24 @@ Fleet %1%
 NEW_FLEET_NAME_NO_NUMBER
 Fleet
 
+NEW_MONSTER_FLEET_NAME
+Herd %1%
+
+NEW_COLONY_FLEET_NAME
+Colony Fleet %1%
+
+NEW_RECON_FLEET_NAME
+Recon Fleet %1%
+
+NEW_TROOP_FLEET_NAME
+Troop Fleet %1%
+
+NEW_BOMBARD_FLEET_NAME
+Bomber Fleet %1%
+
+NEW_BATTLE_FLEET_NAME
+Battle Fleet %1%
+
 # new asteroids or new planet names...
 # used to generate name for effect-created planets, when no name = (string) is specified.
 # When specifying an object name for these effects, if the specified string matches a
--- a/universe/Fleet.cpp
+++ b/universe/Fleet.cpp
@@ -1397,6 +1397,94 @@ std::string Fleet::GenerateFleetName() {
     if (ID() == INVALID_OBJECT_ID)
         return UserString("NEW_FLEET_NAME_NO_NUMBER");
 
+    std::vector<TemporaryPtr<const Ship> > ships;
+    for (std::set<int>::iterator it = m_ships.begin(); it != m_ships.end(); ++it) {
+        if (TemporaryPtr<const Ship> ship = GetShip(*it)) {
+            ships.push_back(ship);
+        }
+    }
+
+    std::vector< TemporaryPtr<const Ship> >::iterator it = ships.begin();
+
+    // TODO C++11 replace all of these loops with std::all_of
+    bool all_monster(true);
+    while (it != ships.end()) {
+        if(!(*it)->IsMonster()) {
+            all_monster = false;
+            break;
+        }
+        ++it;
+    }
+
+    if (all_monster)
+        return boost::io::str(FlexibleFormat(UserString("NEW_MONSTER_FLEET_NAME")) % boost::lexical_cast<std::string>(ID()));
+
+    bool all_colony(true);
+    it = ships.begin();
+    while (it != ships.end()) {
+        if(!(*it)->CanColonize()) {
+            all_colony = false;
+            break;
+        }
+        ++it;
+    }
+
+    if (all_colony)
+        return boost::io::str(FlexibleFormat(UserString("NEW_COLONY_FLEET_NAME")) % boost::lexical_cast<std::string>(ID()));
+
+    bool all_non_coms(true);
+    it = ships.begin();
+    while (it != ships.end()) {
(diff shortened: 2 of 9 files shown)


Vezzra:
For release or post release?

LGM-Doyle:
I would include it in the release because:
- It is purely an minor UI improvement.
- It does not change any rules and hence does not affect balance.
- Play testing of the release candidate has not started yet.

This PR prevents non-combatant, scouts, colony and outpost ships from automatically forming fleets when they are produced in the same turn at the same planet.  

The changes are useful by turn 3 with high production races like the Egassem.  

If an Egassem empire adds 2 scouts to the production queue on turn 1, on turn 3 they produce 2 scouts.  Currently, the scouts group into a fleet, which is not what the player wants.  The player wants 2 scouts to head off in two different directions and find an Inferno planet.

MatGB:
I'd agree with this, it's not something I ever thought to ask for but as soon as I saw the description I realised it was something I definitely wanted. I want my warships grouped, I don't want my scouts or colonies grouped _unless_ I've told them to go to the same place with a rally point, and even that's not necessary.

universe/Fleet.cpp
```diff
@@ -1391,13 +1391,101 @@ void Fleet::ShortenRouteToEndAtSystem(std::list<int>& travel_route, int last_sys
...
+        if (TemporaryPtr<const Ship> ship = GetShip(*it)) {
+            ships.push_back(ship);
+        }
+    }
+
+    std::vector< TemporaryPtr<const Ship> >::iterator it = ships.begin();
+
+    // TODO C++11 replace all of these loops with std::all_of
+    bool all_monster(true);
+    while (it != ships.end()) {
```
geoffthemedio, line 1411: why while and not for?

universe/Fleet.cpp
```diff
@@ -1391,13 +1391,101 @@ void Fleet::ShortenRouteToEndAtSystem(std::list<int>& travel_route, int last_sys
...
+        if(!(*it)->IsMonster()) {
```
geoffthemedio: space after if

universe/Fleet.cpp
```diff
@@ -1391,13 +1391,101 @@ void Fleet::ShortenRouteToEndAtSystem(std::list<int>& travel_route, int last_sys
...
 
-    return boost::io::str(FlexibleFormat(UserString("NEW_FLEET_NAME")) % boost::lexical_cast<std::string>(new_fleet_id));
+    std::vector<TemporaryPtr<const Ship> > ships;
+    for (std::set<int>::iterator it = m_ships.begin(); it != m_ships.end(); ++it) {
+        if (TemporaryPtr<const Ship> ship = GetShip(*it)) {
+            ships.push_back(ship);
+        }
+    }
+
+    std::vector< TemporaryPtr<const Ship> >::iterator it = ships.begin();
```
geoffthemedio: remove extra space after <

universe/Fleet.cpp
```diff
@@ -1391,13 +1391,101 @@ void Fleet::ShortenRouteToEndAtSystem(std::list<int>& travel_route, int last_sys
...
+        ++it;
+    }
+
+    if (all_non_coms)
+        return boost::io::str(FlexibleFormat(UserString("NEW_RECON_FLEET_NAME")) % boost::lexical_cast<std::string>(ID()));
+
+    bool all_troop(true);
+    it = ships.begin();
+    while (it != ships.end()) {
+        if(!(*it)->CanHaveTroops()) {
```
geoffthemedio: space after if

universe/Fleet.cpp
```diff
@@ -1391,13 +1391,101 @@ void Fleet::ShortenRouteToEndAtSystem(std::list<int>& travel_route, int last_sys
...
+        ++it;
+    }
+
+    if (all_troop)
+        return boost::io::str(FlexibleFormat(UserString("NEW_TROOP_FLEET_NAME")) % boost::lexical_cast<std::string>(ID()));
+
+    bool all_bombard(true);
+    it = ships.begin();
+    while (it != ships.end()) {
+        if(!(*it)->CanBombard()) {
```
geoffthemedio: ...

default/stringtables/en.txt
```diff
@@ -148,6 +148,24 @@ Fleet %1%
...
+
+NEW_TROOP_FLEET_NAME
+Troop Fleet %1%
+
+NEW_BOMBARD_FLEET_NAME
+Bomber Fleet %1%
+
+NEW_BATTLE_FLEET_NAME
+Battle Fleet %1%
+
```
geoffthemedio, line 168: Maybe also have a "Mixed Fleet" case, for use if all the ships aren't consistently anything?

LGM-Doyle, line 168: If it falls through everything, then the fleet name is just Fleet %1%

universe/Fleet.cpp
LGM-Doyle, line 1411: No technical reason, but two style reasons.

First.

`for` implies that something should be done for all elements.

`while` implies that something (in this case a check for a ship type) should be done for some elements until something happens (find a specific type).  The reader expects the break statement.  

And second.

The code that I wanted to write was like:

```
`if (std::all_of(ship.begin(), ship.end(), [](const Ship & s){return s.IsMonster();}))
      return "Monster Fleet";`
```

`std::all_of` is only available in C++11 and it expands semantically into the while loop that I used.

LGM-Doyle:
Thanks for the review.

Changes done and pushed.

PR #760, opened 2016-07-07
[AI] Move common code to common folder

Cjkjvfnby:
- move almost all content of `freeorion_debug` to commons
- refactor `extend_free_orion_AI_interface` to be called by function call, not import

I don't assign anyone, if someone will spend some time re review it, it would be great.


dbenage-cx:
Curious on directory structure. `default/python/common` implies potential use by AI and server.  For portions strictly related to AI, would `default/python/AI/common` be more apt?

LGM-Doyle:
Good work making it more generic.

I have a few concerns/questions.
- As @dbenage-cx said, why are the freeOrionAIInterface wrappers being moved to common?  They are only for the AI.
- The same with the interactive shell.  It needs the python ai -> freeorionai -> freeoriond -> HumanClient chain.
- The common code should not call the `freeOrionAIInterface` interface, since different pieces of python code have different interfaces to the C++ code.
  
  The file `default/python/common/handlers/__init__.py` imports the AI interface into all of the common code.
  
  For a specific example in the `timers.py` code it imports.
  `import freeOrionAIInterface as fo`
  
  It then uses `fo` to retrieve the userDataDir the current turn, the empire id.  
- The file `default/python/common/handlers/__init__.py` will be imported into all of the common code.  Is it minimal?  Will every module in the common dir need all of these imports and code?  Otherwise, this code should be separately imported only in the modules that need these imports.

Cjkjvfnby:
@dbenage-cx, @LGM-Doyle, thank you. will rethink this PR.

Cjkjvfnby:
I do next things:
- `freeorion_tools` became folder module, AI content related to `freeorion_debug` folder moved to it
- move timers back to AI (moving them to common will touch a lot of files, and should be done in separate PR)
  - ger rid of AI specific imports from common

PS. Handlers like `inspect_freeOrionAIInterface.py` are not related to AI.  This is developer tools, not AI code. In other PR I will add support for `universe generation` and `turn events`

Cjkjvfnby:
This changes cause my master crash because compiled python files still present on disc.
- Pycharm: context menu on folder "Clean python compiled files"
- unix: 
  
  ``` bash
  find . -name "*.pyc"   # check
  
  find . -name "*.pyc" -delete  # delete
  ```

Vezzra:
Just to emphasize what @dbenage-cx & @LGM-Doyle said: `common` should _only_ contain Python modules that are supposed to be used by both the client/AI and the server Python interfaces. Everything else needs to go into the respective "specialized" interface. Anything but strict adherence to that policy is a recipe for trouble...

Cjkjvfnby:
Handlers AI specific part moved to AI,
Mock generation code moved to separate folder

LGM-Doyle:
The changes address the concerns that I raised.

Vezzra:
As far as I am concerned, looks fine, go ahead.

PR #641, opened 2016-05-02
Link planet density setting to nebulae planet creation.

dbenage-cx:
Issue #628 

Separate effectgroup for each planet density setting, for nebulae planet creation macro.

Feedback requested on desired outcome of each setting.

PR #761, opened 2016-07-08
Append planet ordinal on creation

dbenage-cx:
PR is to address new planet names having a fixed ordinal (outside of universe generation), escalated by #641

Default behaviour when no name is specified will now append a roman number for the orbit position.
Asteroids are named according to their position within the system.

Universe generation is changed to utilize orbital position naming.

OribtalSuffix is added for FOCS, returning the same value for the suffix as the default/no name behaviour.

### DIFF
--- a/default/python/universe_generation/starsystems.py
+++ b/default/python/universe_generation/starsystems.py
@@ -54,16 +54,21 @@ def name_planets(system):
     """
     planet_number = 1
     # iterate over all planets in the system
-    for planet in fo.sys_get_planets(system):
+    for planet in fo.sys_get_planets_by_orbit(system):
+        if planet == fo.invalid_object():
+            continue
+        planet_type = fo.planet_get_type(planet)
+        name = fo.user_string("NEW_PLANET_NAME")
+        name = name.replace("%1%", fo.get_name(system))
         # use different naming methods for "normal" planets and asteroid belts
-        if fo.planet_get_type(planet) == fo.planetType.asteroids:
-            # get localized text from stringtable
-            name = fo.user_string("PL_ASTEROID_BELT_OF_SYSTEM")
-            # %1% parameter in the localized string is the system name
-            name = name.replace("%1%", fo.get_name(system))
+        if planet_type == fo.planetType.asteroids:
+            # get defined orbit suffix
+            name = name.replace("%2%", fo.planet_get_orbital_suffix(planet))
         else:
             # set name to system name + planet number as roman number...
-            name = fo.get_name(system) + " " + fo.roman_number(planet_number)
+            # TODO This has the potential of duplicate names
+            #   if a new planet is created in this system later
+            name = name.replace("%2%", fo.roman_number(planet_number))
             # ...and increase planet number
             planet_number += 1
         # do the actual renaming
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -191,15 +191,17 @@ Bomber Fleet %1%
 NEW_BATTLE_FLEET_NAME
 Battle Fleet %1%
 
-# Name for a newly created asteroid belt.
-# %1% name of the system this asteroid belt is created in.
-# %2% orbit number where this asteroid belt is created on.
-NEW_ASTEROIDS_NAME
-%1% %2% %3%
-
+# This string remains as a non-compile mechanisim to override the default name of new planets.
+# Planet suffix is a roman number, except for asteriods which are unique (see NEW_ASTEROID_SUFFIX_ORBIT_I)
+# During universe generaton, the roman number is ordinal with other non-asteroids, in order of orbit
+# For all other instances, the roman number is the orbital position of the planet
+# %1% represents the system name
+# %2% represents the planet suffix
 NEW_PLANET_NAME
 %1% %2%
 
+# Asteroid suffixes are appended to the default name when a new asteroid is created.
+# Each orbit position is able to have a unique suffix.
 NEW_ASTEROID_SUFFIX_ORBIT_I
 [[PL_ASTEROID_BELT]]
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -195,13 +195,31 @@ Battle Fleet %1%
 # %1% name of the system this asteroid belt is created in.
 # %2% orbit number where this asteroid belt is created on.
 NEW_ASTEROIDS_NAME
-%1% %2% remnants
+%1% %2% %3%
 
-# Name for a newly created planet. When creating a new planet it is assumed this
-# planet is the first one within the system.
-# %1% name of the system this planet is created in.
 NEW_PLANET_NAME
-%1% I
+%1% %2%
+
+NEW_ASTEROID_SUFFIX_ORBIT_I
+[[PL_ASTEROID_BELT]]
+
+NEW_ASTEROID_SUFFIX_ORBIT_II
+[[PL_ASTEROID_BELT]]
+
+NEW_ASTEROID_SUFFIX_ORBIT_III
+[[PL_ASTEROID_BELT]]
+
+NEW_ASTEROID_SUFFIX_ORBIT_IV
+[[PL_ASTEROID_BELT]]
+
+NEW_ASTEROID_SUFFIX_ORBIT_V
+[[PL_ASTEROID_BELT]]
+
+NEW_ASTEROID_SUFFIX_ORBIT_VI
+[[PL_ASTEROID_BELT]]
+
+NEW_ASTEROID_SUFFIX_ORBIT_VII
+[[PL_ASTEROID_BELT]]
 
 EMPTY_SPACE
 Deep Space
--- a/universe/Planet.cpp
+++ b/universe/Planet.cpp
@@ -871,3 +871,29 @@ void Planet::ClampMeters() {
     UniverseObject::GetMeter(METER_REBEL_TROOPS)->ClampCurrentToRange();
     UniverseObject::GetMeter(METER_DETECTION)->ClampCurrentToRange();
 }
+
+const std::string Planet::OrbitalSuffix() const {
+    TemporaryPtr<const System> system_ptr = GetSystem(this->SystemID());
+    int orbit = system_ptr->OrbitOfPlanet(this->ID());
+    if (m_size == SZ_ASTEROIDS) {  // asteroids are named according to orbit position
+        while (!UserStringExists("NEW_ASTEROID_SUFFIX_ORBIT_" + RomanNumber(orbit + 1))) {
+            --orbit;
+            if (orbit < 0) {
+                orbit = system_ptr->NumPlanets();
+                break;
+            }
+        }
+        return UserString("NEW_ASTEROID_SUFFIX_ORBIT_" + RomanNumber(orbit + 1));
+    } else {
+        if (orbit < 0) {
+            orbit = 0;
+            const std::set<int>& system_planet_ids = system_ptr->PlanetIDs();
+            for (std::set<int>::const_iterator it = system_planet_ids.begin(); it != system_planet_ids.end(); ++it) {
+                if (TemporaryPtr<const Planet> it_planet = GetPlanet(*it))
+                    if (it_planet->Size() != SZ_ASTEROIDS)
+                        ++orbit;
+            }
+        }
+        return RomanNumber(orbit + 1);
+    }
+}
--- a/universe/ValueRef.cpp
+++ b/universe/ValueRef.cpp
@@ -1013,6 +1013,17 @@ std::string Variable<std::string>::Eval(const ScriptingContext& context) const
         if (!empire)
             return "";
         return empire->TopPriorityEnqueuedTech();
+    } else if (property_name == "OrbitalSuffix") {
+        if (TemporaryPtr<const Planet> planet = boost::dynamic_pointer_cast<const Planet>(object)) {
+            return planet->OrbitalSuffix();
+        } else if (TemporaryPtr<const System> system_ptr = boost::dynamic_pointer_cast<const System>(object)) {
+            if (TemporaryPtr<const Planet> planet = boost::dynamic_pointer_cast<const Planet>(GetPlanet(*system_ptr->PlanetIDs().rbegin())))
+                return planet->OrbitalSuffix();
+            else
+                return RomanNumber(system_ptr->PlanetIDs().size());
+        }
+        ErrorLogger() << "OrbitalSuffix requested for non-planet or system object: " << "(" << object->ID() << ")" << object->Name();
+        return "";
     }
 
     ErrorLogger() << "Variable<std::string>::Eval unrecognized object property: " << TraceReference(m_property_name, m_ref_type, context);
--- a/default/scripting/fields/fields.macros
+++ b/default/scripting/fields/fields.macros
@@ -99,19 +99,19 @@ EFFECT_CREATE_PLANET
 '''CreatePlanet
     type = OneOf(Barren, Desert, Inferno, Ocean, Radiated, Swamp, Terran, Toxic, Tundra)
     planetsize = OneOf(tiny, small, medium, large)
-    name = UserString("NEW_PLANET_NAME") % Target.Name
+    name = UserString("NEW_PLANET_NAME") % Target.Name % Target.OrbitalSuffix
 '''
 
 EFFECT_CREATE_ASTEROID
 '''CreatePlanet
     type = asteroids
     planetsize = asteroids
-    name = UserString("NEW_ASTEROIDS_NAME") % Target.Name % Source.Name
+    name = UserString("NEW_ASTEROIDS_NAME") % Target.Name % Source.Name % Target.OrbitalSuffix
 '''
 
 EFFECT_CREATE_GASGIANT
 '''CreatePlanet
     type = GasGiant
     planetsize = GasGiant
-    name = UserString("NEW_PLANET_NAME") % Target.Name
+    name = UserString("NEW_PLANET_NAME") % Target.Name % Target.OrbitalSuffix
 '''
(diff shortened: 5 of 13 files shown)


universe/System.cpp
```diff
@@ -270,6 +270,15 @@ StarType System::NextYoungerStarType() const {
     return StarType(m_star - 1);
 }
 
+int System::NumPlanets() const {
+    int retval = 0;
+    for (std::set<int>::const_iterator it = m_planets.begin(); it != m_planets.end(); ++it) {
+        if (*it != INVALID_OBJECT_ID)
```
geoffthemedio: Is this extra check needed ever? (When?)
Alternatively, just return m_planets.size();

default/stringtables/en.txt
```diff
@@ -157,11 +157,10 @@ Fleet
 # these are applied to the %1% style placeholders according the order in which the
 # parameters are specified in the effect
 NEW_ASTEROIDS_NAME
-%1% %2% remnants
+%1% %2% %3% remnants
```
geoffthemedio: Asteroids aren't normally give numbers, are they?

dbenage-cx: They are not, without an exclusion variable you would have a gap in planet numbering.  In the event of multiple asteroid belts, they would all have the same name.

If duplicate asteroid names are not an issue, I can drop the valueref and change the default naming to exclude asteroids.

universe/System.cpp
dbenage-cx: Will reduce to alternate header only change.

geoffthemedio:
Default universe generation doesn't give asteroids numbers so far as I'm aware...

dbenage-cx:
Rebased, dropping OccupiedOrbitsAsRoman in favour of relying on no name behaviour.

geoffthemedio:
The numerals were fine for planets; I was just questioning their use for asteroid belts...

dbenage-cx:
Would you prefer a valueref `NonAsteroidPlanetsAsRoman`?

geoffthemedio:
Whatever is most consistent with the default universe generation...

MatGB:
I've always wondered why multiple asteroid fields aren't numbered in generation, never enough to card though. If you're on High Planets you can end up with multiples in the same system and sitreps can be confusing. That's reduced now sitrep clicking selects the correct one but it's still sometimes annoying. I think I'd prefer numbers for both. 

Slightly different idea, could this allow high orbit asteroid fields to get names like Oort Cloud or similar instead? That would be cool

dbenage-cx:
Current is consistent?
Universe generation (master)
planet_creation_default
Universe generation (PR)
planet_creation_pr
Nebula spawn (PR)
planet_creation_pr_nebula1

dbenage-cx:
@MatGB I've a slightly different preference on asteroid numbering, can be a separate PR after some discussion?  Can roll in positional dependant naming with it.

dbenage-cx:
Rebased, expanding to positional asteroid names.
Updating description to reflect changes.

@MatGB Asteroid suffixes are placeholder, shorter names would be preferred.

universe/Planet.cpp
```diff
@@ -871,3 +871,23 @@ void Planet::ClampMeters() {
...
+    if (m_size == SZ_ASTEROIDS) {  // asteroids are named according to orbit position
+        if (orbit < 1) {
+            orbit = system_ptr->NumPlanets();
+        }
+        return UserString("NEW_ASTEROID_SUFFIX_ORBIT_" + RomanNumber(orbit));
+    } else {
+        if (orbit < 1) {
+            const std::set<int>& system_planet_ids = system_ptr->PlanetIDs();
+            for (std::set<int>::const_iterator it = system_planet_ids.begin(); it != system_planet_ids.end(); ++it) {
+                if (GetPlanet(*it)->Size() != SZ_ASTEROIDS)
```
geoffthemedio: I'd add a safety check on the result of GetPlanet before dereferencing the pointer...

universe/Planet.cpp
```diff
@@ -871,3 +871,23 @@ void Planet::ClampMeters() {
     UniverseObject::GetMeter(METER_REBEL_TROOPS)->ClampCurrentToRange();
     UniverseObject::GetMeter(METER_DETECTION)->ClampCurrentToRange();
 }
+
+const std::string Planet::OrdinalSuffix() const {
```
geoffthemedio: is it intended that "Ordinal" and "Orbital" are both used?

default/python/universe_generation/starsystems.py
```diff
@@ -53,21 +53,21 @@ def name_planets(system):
...
+    orbit_number = 1
     # iterate over all planets in the system
     for planet in fo.sys_get_planets(system):
         # use different naming methods for "normal" planets and asteroid belts
         if fo.planet_get_type(planet) == fo.planetType.asteroids:
             # get localized text from stringtable
-            name = fo.user_string("PL_ASTEROID_BELT_OF_SYSTEM")
-            # %1% parameter in the localized string is the system name
-            name = name.replace("%1%", fo.get_name(system))
+            name = fo.get_name(system) + " " + fo.user_string("NEW_ASTEROID_SUFFIX_ORBIT_" + (fo.roman_number(orbit_number)))
```
geoffthemedio: maybe clamp the orbit number to be no bigger than the highest number present in the stringtable, in case more planets are added than 7?

universe/Planet.cpp
dbenage-cx: It was, since they are so similar and the reasoning was minor, will stick with Orbital.

dbenage-cx:
@geoffthemedio Thank you, rebased

universe/ValueRef.cpp
```diff
@@ -999,6 +999,14 @@ std::string Variable<std::string>::Eval(const ScriptingContext& context) const
         if (!empire)
             return "";
         return empire->TopPriorityEnqueuedTech();
+    } else if (property_name == "OrbitalSuffix") {
+        if (TemporaryPtr<const Planet> planet = boost::dynamic_pointer_cast<const Planet>(object)) {
+            return planet->OrbitalSuffix();
+        } else if (TemporaryPtr<const System> system_ptr = boost::dynamic_pointer_cast<const System>(object)) {
+            return GetPlanet(*system_ptr->PlanetIDs().rbegin())->OrbitalSuffix();
+        }
+        DebugLogger() << "OrbitalSuffix requested for non-planet or system object: " << "(" << object->ID() << ")" << object->Name();
```
geoffthemedio: Should probably be an error...

universe/ValueRef.cpp
```diff
@@ -999,6 +999,14 @@ std::string Variable<std::string>::Eval(const ScriptingContext& context) const
         if (!empire)
             return "";
         return empire->TopPriorityEnqueuedTech();
+    } else if (property_name == "OrbitalSuffix") {
+        if (TemporaryPtr<const Planet> planet = boost::dynamic_pointer_cast<const Planet>(object)) {
+            return planet->OrbitalSuffix();
+        } else if (TemporaryPtr<const System> system_ptr = boost::dynamic_pointer_cast<const System>(object)) {
+            return GetPlanet(*system_ptr->PlanetIDs().rbegin())->OrbitalSuffix();
```
geoffthemedio: should validate planet pointer before using...

default/python/universe_generation/starsystems.py
```diff
@@ -59,10 +59,18 @@ def name_planets(system):
...
+                if (orbit_number - i) < 1:
+                    asteroid_suffix = "NEW_ASTEROID_SUFFIX_ORBIT_I"
+                    break
+                asteroid_suffix = "NEW_ASTEROID_SUFFIX_ORBIT_" + (fo.roman_number(orbit_number - i))
+                i -= 1
+            name = fo.get_name(system) + " " + fo.user_string(asteroid_suffix)
         else:
             # set name to system name + planet number as roman number...
-            name = fo.get_name(system) + " " + fo.roman_number(min(7, planet_number))
+            name = fo.get_name(system) + " " + fo.roman_number(planet_number)
```
geoffthemedio: what happens if there are more than 7 planets? (thought I already commented about that...?)

dbenage-cx: Referring to [this comment](https://github.com/freeorion/freeorion/pull/761#discussion-diff-71681435)?  If so, there is no stringtable lookup performed here, planet count would continue to progress.

geoffthemedio: Referring to stringtable lookup with VIII or greater in the key.

universe/Planet.cpp
```diff
@@ -876,8 +876,12 @@ const std::string Planet::OrbitalSuffix() const {
     TemporaryPtr<const System> system_ptr = GetSystem(this->SystemID());
     int orbit = system_ptr->OrbitOfPlanet(this->ID());
     if (m_size == SZ_ASTEROIDS) {  // asteroids are named according to orbit position
-        if (orbit < 1) {
-            orbit = system_ptr->NumPlanets();
+        while (!UserStringExists("NEW_ASTEROID_SUFFIX_ORBIT_" + RomanNumber(orbit))) {
```
geoffthemedio: looks a bit dangerous... infinite loop if a lowest-numbered stringtable keys are missing...

universe/Planet.cpp
```diff
@@ -876,8 +876,12 @@ const std::string Planet::OrbitalSuffix() const {
...
+            --orbit;
+            if (orbit < 1) {
+                orbit = system_ptr->NumPlanets();
+                break;
+            }
```
dbenage-cx, line 884: Is this insufficient to catch an infinite loop?  Falling back to UserString handling on non-existent lookup entries.

default/python/universe_generation/starsystems.py
dbenage-cx: I'm am not able to find where a lookup occurs in this instance.  The resulting string in `name` is not modified after this line.

universe/Planet.cpp
geoffthemedio, line 884: seems fine...  whole thing is structured a bit weird to me, but I guess it's OK

geoffthemedio:
I don't understand the logic of these asteroid names...

asteroids_names

dbenage-cx:
Temporary unique names for verification.  Change them all to simply `asteroids` until suitable names are decided?

geoffthemedio:
I don't object to the use of different names. My confusion is about the ordering (low vs. unspecific) and inconsistency (asteroids vs. asteroid belt)

dbenage-cx:
@geoffthemedio Thank you.  I've removed that commit for now, will update once a sufficient proposal is reached.

Others:
PR is pending some suitable names for asteroids: http://www.freeorion.org/forum/viewtopic.php?f=15&t=10149

dbenage-cx:
Rebased.

Vezzra:
Ok, after one and a half months I finally get around to have a look at this. I don't know what the implementation had been prior to @geoffthemedio initial comments (that apparently made you drop numbering of asteroid belts and maybe change something in the way you determine orbit number?), but in it's current state this PR results in planet naming like this:

orbitalnumbering

That's not really what we want, and is a result of using the absolute "orbital slot" number, so to speak. The problem here is, occupied and empty orbits are mixed up, it's not that in a system with three planets the first three orbits are occupied by the planets and the rest is empty, but you can have some empty orbits, then some occupied, then some empty, and so on. In the example above planets have been placed in orbits 4 and 6 (and named accordingly, the ordering issue is fixed in #869 if I understand correctly, so not an issue here).

However, "Bhanu Beta IV" should be "Bhanu Beta I", and "Bhanu Beta VI" should be "Bhanu Beta II".

To achieve this, you'll have to extend the determination of the orbit number a bit. I suggest to add a member function `PlanetIDsByOrbit` to the `System` class, which returns the id's of all the planets of a star system ordered by orbit. You can then determine the planet number by the position of the planets id in that ordered set in `OrbitalSuffix` (so, if it's the third planet in the "Orion" system, it will be named "Orion III").

Expose both member functions (`PlanetIDsByOrbit` and `OrbitalSuffix`) to Python, and we can use them for universe generation as well, making it easy to use the same naming method there.

For simplicity sake I'd also apply that method for the asteroid belts in this PR (so we have "Orion asteroid belt I" etc., and using independent numbering for asteroid belts and planets), and make a separate one for the fancy naming. This PR tries to fix an issue that should go in for the release, while the fancy naming is nice, but not critical, and I don't want to delay the release because we don't reach a consensus here in time.

default/stringtables/en.txt
```diff
@@ -185,15 +185,13 @@ Bomber Fleet %1%
...
 Battle Fleet %1%
 
-# Name for a newly created asteroid belt.
-# %1% name of the system this asteroid belt is created in.
-# %2% orbit number where this asteroid belt is created on.
-NEW_ASTEROIDS_NAME
-%1% %2% %3%
-
+# This string remains as a non-compile mechanisim to override the default name of new planets.
+# The default naming scheme for new planets is "System Suffix"
```
adrianbroher: Please use the '%1% description' documentation schema.

dbenage-cx:
@Vezzra Thank you for the review.

I've made a number of the changes you requested, but not all of them.  The current state numbers planets in order only for universe generation.  Since I will need to revisit this from the start to overcome FOCS planet creation, tomorrow I will try to concentrate on a simpler fix/PR suitable for release.

dbenage-cx:
I had hoped to create a smaller changeset to address nebula planet creation.
This has not panned out so far, might try deferring planet creation in nebulas to one planet per turn.

Barring another idea coming up, either:
-  Work out any remaining issues in this PR
-  Revert abafca7e and set this and 869 to `post 0.4.6` (or void, if I've simply given everyone a headache )

For this PR, nebulas still name planets for their orbit.  Attempting to label them with the same scheme as universe generation keeps bringing me back full circle (nebula planets all with same name).
Asteroids need to change in universe generation as well, which is relatively simple.

Vezzra:
> Work out any remaining issues in this PR

I think this requires to first fix the issues you tried to address in #869. Once the mess with the orbits is fixed, we can continue here.

> Revert abafca7 and set this and 869 to post 0.4.6

Hm, I'd leave abafca7 in place. First of all, reverting it would just lead to systems created by nebulas not honoring the planet density game setup option, the naming of the planets/asteroid belts would still be faulty. Or did I miss something...?

And we can still try to get things to work. Should it turn out to be too complicated to be done in time for the release, we can still set it to "post 0.4.6".

> (or void, if I've simply given everyone a headache )

Naaah... I mean, headache yes, but that's not your fault, your attempt to fix the planet naming just uncovered an underlying, apparently somewhat elusive issue. Which needs to be fixed anyway, and once that has been addressed, getting this PR to work shouldn't be that hard anymore.

dbenage-cx:
Reverting to the previous behavior of nebulas results in a max of 2 asteroids and one planet or gas giant, effectively masking the issue.  While I would not like to see any reverts, honoring density settings here is not highly perceivable to the end user.

Vezzra:
Ah ok, I see. Still, I'd prefer to keep abafca7, it's a nice improvement and the faulty naming isn't that big of an issue.

Vezzra:
@dbenage-cx, #869 is in, we can proceed with this patch here (or its successor). I've already spotted your new [641_defunct_planet_creation_name_cardinal](https://github.com/dbenage-cx/freeorion/commits/641_defunct_planet_creation_name_cardinal) branch. So I assume you're going to replace this PR here with a new one based on that new branch?

I've peeked into it, and have to say looks good so far! Seems only a few bits are missing (stringtable update as far as I can tell), I guess it will be ready soon?

Just a heads up: I'll be gone over the weekend, and won't be able to do FO stuff (or at least not much) until I return Monday evening. So no worries if I'm a bit unresponsive... 😉

dbenage-cx:
@Vezzra My issue with that branch is it is that the current implementation of `CardinalSuffix` is not sufficient for FOCS.
I am currently stumped on trying to create that desired outcome, without somehow flagging each planet for a deferred batch rename which propagates before the turn start.

Without resolving that, there is an option to implement `OrbitalSuffix` in that branch and use it for nebulas.  With the planets pushed to the lowest free orbit now, the inconsistencies to the user would be in systems with asteroids.

> So I assume you're going to replace this PR here with a new one based on that new branch?

Not sure if you prefer this branch be replaced (update this PR), or a new PR is created?  Each time I read it, I think it means the other one :smirk:.  This is assuming that branch is/becomes a more viable branch.

Vezzra:
@dbenage-cx:

> My issue with that branch is it is that the current implementation of CardinalSuffix is not sufficient for FOCS.

My friend, what are you talking about?? I just checked out your branch, rebased it onto master and ran a couple of tests - works like a charm! The only issues left I noticed are:
- If there are more than one asteroid belts in the newly created system the first one does not get a cardinal, so you get "Orion asteroid belt", "Orion asteroid belt II", etc.
- The ordering is a bit "unflexible" - if there is an asteroid belt, it's in the lowest orbit most of the time.

Both issues are to be expected considering the implementation you cooked up so far, but both are of the very minor cosmetic kind. Compared with what we had before this is already a vast improvement, and certainly more than sufficient for the release (IMO).

So, if you don't mind, I suggest you put up what you've got so far as a PR and we merge that, so we are set for the release, at least as far as this issue is concerned.

You are then more than welcome to further tinker with that thing to address the remaining glitches, but as you said, the obvious solution is what is also done in universe generation: create the systems and after that do a batch rename once the composition of the systems is fixed. That will also allow to switch back to random free orbit selection in `System::Insert`, which will take care of the "unflexible" ordering of planets/asteroids.

But that approach will take a bit more effort than we can squeeze in before the release I think, so we should get in what we already have, and continue from there.

dbenage-cx:
> works like a charm
> Well that is just aggravating :p I had it sidelined due to testing with random orbit placement.

Created new PR, this one should probably be rejected if that is merged.  Thanks @geoffthemedio and @Vezzra for the reviews and assistance.

Vezzra:
Replaced by #922, closing this one.

PR #763, opened 2016-07-10
Move species collection to state

Cjkjvfnby:
Replace some module level variables to methods in state.
- `empire_species` -> Splited for two methods:
  - `state.get_species_planets()`
  - `state.get_planets_for_species(species_name)`
- `empire_species_by_planet` - > `state.planets_with_species()`
- `empire_species_systems` -> `state.get_empire_species_systems`,
  structure changed: `{sid: {'pids': [pid, ...]}}` - > `{sid: [pid, ...]}`

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -17,9 +17,7 @@ from freeorion_tools import dict_from_map, tech_is_complete, get_ai_tag_grade, c
 
 colonization_timer = Timer('getColonyFleets()')
 
-empire_species = {}
-empire_species_by_planet = {}
-empire_species_systems = {}  # TODO: as currently used, is duplicative with combo of foAI.foAIstate.popCtrSystemIDs and foAI.foAIstate.colonizedSystems
+
 empire_colonizers = {}
 empire_ship_builders = {}
 empire_shipyards = {}
@@ -236,8 +234,6 @@ def survey_universe():
         empire_status.update({'industrialists': 0, 'researchers': 0})
         AIstate.empireStars.clear()
         empire_outpost_ids.clear()
-        empire_species.clear()
-        empire_species_by_planet.clear()
         empire_colonizers.clear()
         empire_ship_builders.clear()
         empire_shipyards.clear()
@@ -247,7 +243,6 @@ def survey_universe():
         active_growth_specials.clear()
         if tech_is_complete(TechsListsAI.EXOBOT_TECH_NAME):
             empire_colonizers["SP_EXOBOT"] = []  # get it into colonizer list even if no colony yet
-        empire_species_systems.clear()
         AIstate.popCtrIDs[:] = []
         AIstate.popCtrSystemIDs[:] = []
         AIstate.outpostIDs[:] = []
@@ -257,7 +252,7 @@ def survey_universe():
         unowned_empty_planet_ids.clear()
         facilities_by_species_grade.clear()
         system_facilities.clear()
-        state.cleanup()
+        state.update()
 
     # var setup done
 
@@ -300,10 +295,7 @@ def survey_universe():
                 elif this_spec:
                     empire_has_qualifying_planet = True
                     AIstate.popCtrIDs.append(pid)
-                    empire_species_systems.setdefault(sys_id, {}).setdefault('pids', []).append(pid)
                     empire_has_pop_ctr_in_sys = True
-                    empire_species_by_planet[pid] = spec_name
-                    empire_species.setdefault(spec_name, []).append(pid)
                     for metab in [tag for tag in this_spec.tags if tag in AIDependencies.metabolismBoostMap]:
                         empire_metabolisms[metab] = empire_metabolisms.get(metab, 0.0) + planet.size
                     if this_spec.canProduceShips:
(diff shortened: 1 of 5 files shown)


default/python/AI/turn_state.py
```diff
@@ -18,9 +24,60 @@ def __init__(self):
...
+
+        :rtype: dict[int, list[int]]
+        """
+        # TODO: as currently used, is duplicative with combo of foAI.foAIstate.popCtrSystemIDs and foAI.foAIstate.colonizedSystems
+        empire_id = fo.empireID()
+        planets_with_species = (x for x in self.__planets.itervalues() if x.owner == empire_id and x.species)
+        result = {}
+        for x in planets_with_species:
+            result.setdefault(x.system_id, []).append(x.pid)
+        return result
```
Morlic-fo, line 57: I'm trying to figure out your design concept and intentions for this module. 

Is there a reason why this (and the following) is an explicit public function instead of calling the code once in `update()` and implementing the dict as class property?

default/python/AI/turn_state.py
```diff
@@ -18,9 +24,60 @@ def __init__(self):
...
+    def get_planets_for_species(self, species):
+        """
+        Return list of empire planet ids with species.
+
+        :rtype: list[int]
+        """
+        if not species:
+            return []
+        empire_id = fo.empireID()
+        return [x.pid for x in self.__planets.itervalues() if x.owner == empire_id and x.species == species]
```
Morlic-fo: Why does this return a `list` while `get_inhabitated_species` returns a `set`?

`get_species_planets` builds a dict which includes this information. Is there any reason to build the list here again instead of having the dict built by `get_species_planets` (during an `update` call) and simply return the mapped value of the species key?

default/python/AI/turn_state.py
```diff
@@ -18,9 +24,60 @@ def __init__(self):
         self.__have_computronium = False
         self.__best_pilot_rating = 1e-8
         self.__medium_pilot_rating = 1e-8
+        self.__planets = {}
 
-    def cleanup(self):
+    def update(self):
```
Morlic-fo: Docstring of the class needs to be updated to reflect this change.

default/python/AI/turn_state.py
Cjkjvfnby, line 57: I don't sure you talk about `@property` or `attribute` 

As public interface I  prefer method because it does not mutate inner state and have documentation (autocomplete and hints in IDE).

default/python/AI/turn_state.py
Cjkjvfnby: Good point, will update

Cjkjvfnby: > Why does this return a list while get_inhabitated_species returns a set?

I checked usage and select best suitable collection. By the way I changed them to imitable analogues.

> Is there any reason to build the list here again instead of having the dict built by
> done

Cjkjvfnby:
PS. Not ready for merge. After we will have agreement on public interfaces I will do some research on inner implementation to speed up this code if it will be necessary.

Cjkjvfnby:
@Morlic-fo, do you remember about me?

default/python/AI/turn_state.py
```diff
@@ -51,21 +51,22 @@ def get_inhabited_planets(self):
         """
         Return set of empire planet ids with species.
 
-        :rtype: set
+        :rtype: set[int]
         """
         empire_id = fo.empireID()
-        return {x.pid for x in self.__planets.itervalues() if x.owner == empire_id and x.species}
+        return frozenset(x.pid for x in self.__planets.itervalues() if x.owner == empire_id and x.species)
```
Morlic-fo: `:rtype:` should be updated, I guess?

Morlic-fo:
> After we will have agreement on public interfaces

Not sure what exactly you want to discuss. If it is about the public function vs `@property` implementation, I am perfectly fine with your approach. I was mostly asking to make sure my future additions are consistent.

Vezzra:
@Cjkjvfnby, @Morlic-fo: status update on this? Looking at the discussion, it seems that you guys reached some kind of consensus, but then nothing happened anymore. Any ideas on how you want to proceed?

Cjkjvfnby:
I rebased this branch.

@Morlic-fo, I want to store data in raw format, so we can grep it in way we need, without changes in update.

PS. I did not run game after rebase.

default/python/AI/ColonisationAI.py
```diff
@@ -395,10 +387,10 @@ def survey_universe():
         table_name="Empire species roster"
     )
 
-    for spec_name in empire_species:
+    for spec_name, planets in state.get_species_planets().items():
```
Morlic-fo: These are planet ids not planet objects, so should be named `pids` or `planet_ids` but probably not planets.

Morlic-fo (changes requested):
Most of the new get_x functions need to have a clearer name indicating what they return.

While some of the old variable names were just as unspecific/misleading, I think we should take this refactoring as opportunity to provide a clear and intuitive interface.

default/python/AI/turn_state.py
```diff
@@ -18,9 +24,61 @@ def __init__(self):
...
+    def update(self):
         self.__init__()
+        universe = fo.getUniverse()
+        for pid in universe.planetIDs:
+            planet = universe.getPlanet(pid)
+            self.__planets[pid] = PlanetInfo(pid, planet.speciesName, planet.owner, planet.systemID)
+
+    def get_empire_species_systems(self):
+        """
+        Return dict from system id to planet ids of empire with species.
```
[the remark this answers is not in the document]
Cjkjvfnby: |  The summary line may be on the same line as the opening quotes or on the next line. 

I prefer to put it in next line for multistring comments

Morlic-fo (approved):

MatGB:
I'm getting AI crashes on both new and saved games with this when I compile from master, if I branch from before this is merged everything appears to run fine.

Cjkjvfnby:
I did not finished it yet. Will try to fix asap

Cjkjvfnby:
Crash fixed in master

PR #764, opened 2016-07-11
add tech turn requirement and turn cost to unresearched techs

TheSilentOne1:
also increase tech panel width and decrease (standard) horizontal distance between techs:
tech_screen

### DIFF
--- a/UI/TechTreeWnd.cpp
+++ b/UI/TechTreeWnd.cpp
@@ -36,7 +36,7 @@ namespace {
 
     // command-line options
     void AddOptions(OptionsDB& db) {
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Doesn't fit in the available space with closer-to-current-default font size.

TheSilentOne1:
I think the turn/RP cost really should be on the panel. Solutions to a larger font size would be:
- variable panel width/spacing depending on font size
- place cost label under tech name, increase panel height with font size:
  tech_panel

geoffthemedio:
Could also put it at the bottom-right, like on the queue panels.

Deepsloth:
Bottom-right sounds like a good idea.

TheSilentOne1:
Currently caught up IRL, but I'll get back to this soon.

TheSilentOne1:
I corrupted my branch of this pull-request, so I created a new one here:
https://github.com/freeorion/freeorion/pull/791

PR #765, opened 2016-07-12
Fix scope logic for damage control on moving ships

dbenage-cx:
Skip battlecheck for ships in motion.

```
[error] Client : ValueRef.cpp:79 : FollowReference : Unable to get system for object
[error] Client : ValueRef.cpp:753 : Variable<int>::Eval unable to follow reference: LocalCandidate.System.LastTurnBattleHere :  | Local Candidate: Ship 1538 (  )  |  System (-1):  |  LastTurnBattleHere  | 
[error] Client : ValueRef.cpp:755 : source: Planet 63 ( Minerva I )
```

### DIFF
--- a/default/scripting/ship_hulls/robotic/SH_NANOROBOTIC.focs.txt
+++ b/default/scripting/ship_hulls/robotic/SH_NANOROBOTIC.focs.txt
@@ -36,9 +36,15 @@ Hull
         EffectsGroup
             scope = And [
                 Source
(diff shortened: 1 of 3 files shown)


MatGB:
Ah, yes, I should've noticed that in testing, oops, I had similar problems when I did gated Advanced Damage Control and that's one of the reason that got the InSystem requirement. Good catch, thanks.

PR #769, opened 2016-07-15
Add dismiss sensor ghost

LGM-Doyle:
The UI displays known objects that have been detected in previous turns as partially greyed out.

Sometimes these fleets disappear/merge/die outside of the view of the player empire.  Then the ghost fleets are displayed in their last known state forever, even though the player is fairly certain that the fleet no longer exists and is very certain that they want the detected fleet removed from the display.

This PR adds a context menu option "Dismiss Sensor Ghost" to ghost fleets.  The option removes the fleet from the table of known objects. If the fleet is detected again it is re-added to the table of known objects as normal.

This issue has been brought up multiple times before now, most recently by Starionx and @MatGB  in [the forum](http://freeorion.org/forum/viewtopic.php?f=28&t=10148&sid=ef956232905082d5a4b97150b6a278aa&start=15).

### DIFF
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -3584,14 +3584,8 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
             GetUniverse().ForgetKnownObject(ALL_EMPIRES, fleet->ID());
 
             // Force a redraw
-
-            // Just signaling StateChanged doesn't work because MapWnd
-            // only resets/deletes paths to non-existent objects on the
-            // turn updates and not when handling StateChanged.
-            // fleet->StateChangedSignal();
-
-            //So brute force.
-            ClientUI::GetClientUI()->GetMapWnd()->MidTurnUpdate();
+            this->Refresh();
+            ClientUI::GetClientUI()->GetMapWnd()->RemoveFleet(fleet->ID());
 
             break;
         }
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -2683,7 +2683,7 @@ void FleetDetailPanel::ShipRightClicked(GG::ListBox::iterator it, const GG::Pt& 
             std::map<int, int> pending_scrap_orders = PendingScrapOrders();
             std::map<int, int>::const_iterator it = pending_scrap_orders.find(ship->ID());
             if (it != pending_scrap_orders.end())
-                HumanClientApp::GetApp()->Orders().RecindOrder(it->second);
+                HumanClientApp::GetApp()->Orders().RescindOrder(it->second);
             break;
         }
 
@@ -3523,7 +3523,7 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
                 for (OrderSet::const_iterator it = orders.begin(); it != orders.end(); ++it) {
                     if (boost::shared_ptr<ScrapOrder> order = boost::dynamic_pointer_cast<ScrapOrder>(it->second)) {
                         if (order->ObjectID() == ship_id) {
-                            HumanClientApp::GetApp()->Orders().RecindOrder(it->first);
+                            HumanClientApp::GetApp()->Orders().RescindOrder(it->first);
                             // could break here, but won't to ensure there are no problems with doubled orders
                         }
                     }
@@ -3550,7 +3550,7 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
                     boost::dynamic_pointer_cast<GiveObjectToEmpireOrder>(it->second))
                 {
                     if (order->ObjectID() == fleet->ID()) {
-                        HumanClientApp::GetApp()->Orders().RecindOrder(it->first);
+                        HumanClientApp::GetApp()->Orders().RescindOrder(it->first);
                         // could break here, but won't to ensure there are no problems with doubled orders
                     }
                 }
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -3435,13 +3435,14 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
     }
 
 
-    //Allow dismissal of stale visibility information
+    // Allow dismissal of stale visibility information
     if (!fleet->OwnedBy(client_empire_id)){
         Universe::VisibilityTurnMap visibility_turn_map =
             GetUniverse().GetObjectVisibilityTurnMapByEmpire(fleet->ID(), client_empire_id);
         Universe::VisibilityTurnMap::const_iterator last_turn_visible_it = visibility_turn_map.find(VIS_BASIC_VISIBILITY);
         if (last_turn_visible_it != visibility_turn_map.end()
-            && last_turn_visible_it->second < CurrentTurn()) {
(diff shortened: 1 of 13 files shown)


UI/FleetWnd.cpp
```diff
@@ -3550,11 +3561,34 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
...
+        }
+
+        case 13: { // Remove visibility information for this fleet from
+                   // the empire's visibility table.
+            //server info for future effect
+            HumanClientApp::GetApp()->Orders().IssueOrder(
+                OrderPtr(new ForgetOrder(client_empire_id, fleet->ID()))
+            );
+            //local info for immediate effect
+            GetUniverse().ForgetKnownObject(ALL_EMPIRES, fleet->ID());
```
geoffthemedio, line 3584: should be redundant; the order should be executed locally immediately when issued

universe/Universe.cpp
```diff
@@ -2506,6 +2506,39 @@ void Universe::ApplyEffectDerivedVisibilities() {
     }
 }
 
+void Universe::ForgetKnownObject(int empire_id, int object_id) {
```
geoffthemedio, line 2509: ah... ok... local client needs to remove from the main objectmap, whereas on the server it applies to the empire-specific object knowledge... should clarify the comment when issuing about why the extra call is needed.

universe/Universe.cpp
```diff
@@ -2506,6 +2506,39 @@ void Universe::ApplyEffectDerivedVisibilities() {
...
+    // Note: Client calls this with empire_id == ALL_EMPIRES to
+    // immediately forget information without waiting for the turn update.
+    ObjectMap& objects(EmpireKnownObjects(empire_id));
+
+    if (objects.Empty())
+        return;
+
+    TemporaryPtr<UniverseObject> obj = objects.Object(object_id);
+    if(!obj) {
```
geoffthemedio: space before (

util/Order.cpp
```diff
@@ -1304,3 +1304,35 @@ bool GiveObjectToEmpireOrder::UndoImpl() const {
...
+
+ForgetOrder::ForgetOrder(int empire, int object_id) :
+    Order(empire),
+    m_object_id(object_id)
+{}
+
+void ForgetOrder::ExecuteImpl() const {
+    ValidateEmpireID();
+    int empire_id = EmpireID();
+    TemporaryPtr<Fleet> fleet = GetFleet(m_object_id);
```
geoffthemedio: should also handle forgetting planets and systems

universe/Universe.cpp
LGM-Doyle, line 2509: I expanded the comments at the calling location.

util/Order.cpp
LGM-Doyle: Do not look behind the curtain.  We are making sausage here.

I started from and only tested it with the ghost fleets.  It is fixed now.

LGM-Doyle:
Rebased.

UI/FleetWnd.cpp
```diff
@@ -3435,6 +3435,17 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
     }
 
 
+    //Allow dismissal of stale visibility information
+    if (!fleet->OwnedBy(client_empire_id)){
+        Universe::VisibilityTurnMap visibility_turn_map =
+            GetUniverse().GetObjectVisibilityTurnMapByEmpire(fleet->ID(), client_empire_id);
+        Universe::VisibilityTurnMap::const_iterator last_turn_visible_it = visibility_turn_map.find(VIS_BASIC_VISIBILITY);
+        if (last_turn_visible_it != visibility_turn_map.end()
+            && last_turn_visible_it->second < CurrentTurn()) {
```
geoffthemedio: { to next line

UI/FleetWnd.cpp
```diff
@@ -3435,6 +3435,17 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
     }
 
 
+    //Allow dismissal of stale visibility information
```
geoffthemedio: space before A

UI/FleetWnd.cpp
```diff
@@ -3550,11 +3561,38 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
...
+
+            // Force a redraw
+
+            // Just signaling StateChanged doesn't work because MapWnd
+            // only resets/deletes paths to non-existent objects on the
+            // turn updates and not when handling StateChanged.
+            // fleet->StateChangedSignal();
+
+            //So brute force.
+            ClientUI::GetClientUI()->GetMapWnd()->MidTurnUpdate();
```
geoffthemedio: I think MidTurnUpdate is doing a lot more than is needed at this point, including recalculating the system graph connectivity and refreshing all the system icon and fleet icons and their rendering stuff. Should be possible to something more specific, like

```
            this->Refresh();
            ClientUI::GetClientUI()->GetMapWnd()->RefreshFleetButtons();
```

MapWnd::RefreshFleetButtons is private though, so would need to be moved to the public section of the MapWnd definition. It's also refreshing all the fleet buttons, which is probably not necessary, but fixing that would be a bit more involved.

Vezzra:
I assume this should go in for the release...

UI/FleetWnd.cpp
LGM-Doyle: I added `MapWnd::RemoveFleet(id)` to fix the problem. 

`MidTurnUpdate()` was overkill.  That is why I described it as the brute force solution.

However, `RefreshFleetButtons()` does not reset or update fleet movement lines.  The case where you dismiss the sensor ghost of an in transit fleet needs the fleet lines deleted as well as the fleet icon.

UI/MapWnd.cpp
```diff
@@ -3931,6 +3938,12 @@ void MapWnd::RemoveProjectedFleetMovementLine(int fleet_id) {
         m_projected_fleet_lines.erase(it);
 }
 
+void MapWnd::RemoveFleetMovementLine(int fleet_id) {
+    std::map<int, MovementLineData>::iterator it = m_fleet_lines.find(fleet_id);
```
geoffthemedio: can't you just do

```
    m_fleet_lines.erase(fleet_id);
```

?
don't really need a separate function for that, if so..,

LGM-Doyle: No need at all.  

I removed it and also RemoveProjectedFleetMovementLine which was never called outside of MapWnd.cpp.

universe/Universe.cpp
```diff
@@ -2506,6 +2506,50 @@ void Universe::ApplyEffectDerivedVisibilities() {
...
+    }
+
+    if (int container_id = obj->ContainerObjectID() != INVALID_OBJECT_ID) {
+        if (TemporaryPtr<UniverseObject> container = objects.Object(container_id)) {
+            if (TemporaryPtr<System> system = boost::dynamic_pointer_cast<System>(container))
+                system->Remove(object_id);
+            else if (TemporaryPtr<Planet> planet = boost::dynamic_pointer_cast<Planet>(container))
+                planet->RemoveBuilding(object_id);
+            else if (TemporaryPtr<Fleet> fleet = boost::dynamic_pointer_cast<Fleet>(container))
+                fleet->RemoveShip(object_id);
```
geoffthemedio, line 2546: might need some extra handling here if the last ship is removed from a fleet... not sure, though...

PR #774, opened 2016-07-19
The Honeycomb special will now turn nearby planets into asteroids ins…

Deepsloth:
…tead of destroying them.

Suggestion by Mat here: http://www.freeorion.org/forum/viewtopic.php?f=6&t=10147

### DIFF
--- a/default/scripting/specials/HONEYCOMB.focs.txt
+++ b/default/scripting/specials/HONEYCOMB.focs.txt
@@ -38,7 +38,10 @@ Special
                 ]
             ]
             activation = Turn high = 0 
-            effects = Destroy
+            effects = [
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Is it intended that the new asteroid belts will keep their planet-type names?

Would make sense for events happening mid-game, but probably not for stuff during galaxy generation / on turn 1 before any players can see it.

MatGB:
It is intended, yes, as discussed on forum thread

geoffthemedio:
ah, there it is. thought I remembered it, but didn't see it on the page.

Vezzra:
Looks very simple and straightforward, anything preventing this from being merged?

PR #775, opened 2016-07-20
[AI] Cleanup: update log output

Cjkjvfnby:
This should be merged via squash.

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -937,10 +937,10 @@ class AIstate(object):
                 elif this_sys:
                     status['sysID'] = this_sys.id
                 else:
-                    main_missin = self.get_fleet_mission(fleet_id)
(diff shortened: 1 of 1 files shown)


geoffthemedio:
For stuff like this, I think you could just push it yourself without bothering with a pull request, but I don't deal with Python code much myself, so maybe it's preferred otherwise?

Vezzra:
What @geoffthemedio said. @Cjkjvfnby, you're currently pretty much the only active AI dev left, so it's essentially your call. I've taken a quick glance at the changes, they look fine to me, for what it's worth.

Cjkjvfnby:
It is common for me to make some stupid mistakes, so I prefer to make review once again after some times come.

Vezzra:
> It is common for me to make some stupid mistakes

Well, welcome to the club 😉 I'm definitely prone to that too. With simple changes like this however, I'd say just commit it, and worst case, if there is a stupid mistake, fix it with another commit - did that often enough...

PR #791, opened 2016-07-24
add tech turn and RP requirement to tech panel (2)

TheSilentOne1:
adds tech RP and turn cost to the lower right of tech panels, for unresearched techs.
also adjusts tech panel width, and distance between the panels

tech_tree

### DIFF
--- a/UI/TechTreeWnd.cpp
+++ b/UI/TechTreeWnd.cpp
@@ -36,7 +36,7 @@ namespace {
 
     // command-line options
     void AddOptions(OptionsDB& db) {
-        db.Add("UI.tech-layout-horz-spacing",   UserStringNop("OPTIONS_DB_UI_TECH_LAYOUT_HORZ_SPACING"), 1.0,  RangedStepValidator<double>(0.25, 0.25, 4.0));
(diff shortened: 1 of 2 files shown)


LGM-Doyle:
Once I tried this change, I liked it.

MatGB:
Agreed, it falls under the "why didn't we think of this years ago" category of improvements. Only thing I dislike is for the really cheap techs it displays 8.0 RP not just 8 RP which looks a bit weird.

TheSilentOne1:
Yeah, I also don't like the .0 thing. Probably could simply check if the string ends with ".0" and truncate it, or is there a more elegant method?

MatGB:
No idea, it's not taking the .0 from the tech focs file, Physical Brain is 4 \* multiplier and that displays as 8.0, so it's always showing two numerals for some reason.

geoffthemedio:
DoubleToString returns a string representing the passed in number, with the specified number of digits, in this case 1, which is actually bumped up to 2 in the function because you can't meaningfully display numbers with just 1 digit.

https://github.com/freeorion/freeorion/blob/master/UI/TechTreeWnd.cpp#L1031

PR #794, opened 2016-07-25
fleet screen: increase fleet drop target width and add an icon to it

TheSilentOne1:
fleet_screen


Vezzra:
Looks simple enough, so tagged it for the release.

geoffthemedio:
looks simple enough; assuming it builds w/o testing

PR #797, opened 2016-07-25
Improve mapwnd supply propagation performance

LGM-Doyle:
tl:dr  This PR changes `MapWnd::InitStarlaneRenderingBuffers()` from a O(N^3) algorithm to a O(N) algorithm.

This PR refactors `MapWnd::InitStarlaneRenderingBuffers()` to improve performance.  

`MapWnd::InitStarlaneRenderingBuffers()` calculates which systems and starlanes to display with wide lines in the map window to to indicate being part of a supply network.

`MapWnd::InitStarlaneRenderingBuffers()` is the principle source of lag in the map window when the supply network of systems is large.  I determined this by profiling with `perf`.  

The problem is obvious when updating the production queue with a large number of systems in any empire's supply network.  freeorion It briefly freezes.

"Large" number of systems is a fuzzy number dependent on each machine.  It might be 100 systems on a slow machine or 500 on a fast one.

**Algorithm Comparison.**

Given N resource systems, M supply network systems and P average starlanes per system.

The old method was roughly:

```
`from N resources systems
   to each of N resource systems
      calculate the shortest path with P average starlanes
      using a breadth first search of M supply network systems.`
```

The old method was O(N^2_M_P).  In the late game P << N and N ~ M, so this reduces to O(N^3).

The new method is roughly:

```
`do a single breadth first search from all N resource systems to any of N
resource systems through M network systems with P average starlanes.`
```

The new method is O((N+M)*P).  The new method handles each system once.  In the late game this reduces to O(N). 

**Performance Testing:**
You can easily verify the performance difference in the algorithms by checking the log for the string `MapWnd::InitStarlaneRenderingBuffers time:`.  The changes are save game compatible, so pick a late game save where the production window is obviously laggy and load the game, record some of the times, compile this PR, reload the game and compare the times.

The more systems in the supply network the bigger the difference in performance.  In the test case I used to develop the PR my improvement was 4400 ms vs 10 ms.  

**Behavioral change:**
  There is one user visible change in the behavior of the algorithms.  The old algorithm only displayed the shortest path between every resource to every other resource.  If two path were the same length it chose the path with the lower numbered system ids.  

The new algorithm shows all redundant supply paths as part of the supply network.

There is debug code in the PR to visualize the change. I removed the testing code in the last commit.  To visualize the change checkout https://github.com/freeorion/freeorion/commit/d09f01088e5822993fe7afe67e371b779010bbec `Clarify comments in MapWnd::InitStarlaneRenderingBuffers.` and load a saved game.  It will display the systems that are added to the rendered supply network with the new algorithm with big orange triangles.  

Note, this does not change how the server handles the supply network, just how it is rendered in `MapWnd`.

@Vezzra I'm assuming that you will ask the perennial question, should this go into 0.4.6?  
- It addresses the regular forum complaint about late game slowness.  
- It does not make any rules changes.  
- It is "tested" code to the extent that any code in freeorion is tested. The aforementioned commit compares the new method directly with the old method.  

I think it should go in.


Vezzra:
> I think it should go in.

@LGM-Doyle, for all the reasons you cited, I agree.

geoffthemedio:
needs rebase

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
-                continue;
-            if (!laneIt->second && ( ancestor[newSys] == -1 )) { //is a starlane, and not yet visited newSys //TODO: should allow wormholes here?
-                ancestor[newSys] = sysID;
-                if (newSys==end_sys_it) {
-                    int iSys = newSys;
-                    while ((ancestor[iSys] !=-1)&&( ancestor[iSys] != iSys )) {
-                        path.push_back(iSys);
-                        iSys = ancestor[iSys];
+namespace {
+    // SupplyLaneMap map keyed by system containing all system
```
geoffthemedio: systemS ?

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+        const SupplyLaneMMap& supply_lanes)
+    {
+        good_path.clear();
+
+        // No terminal points, so all paths lead nowhere.
+        if (terminal_points.empty())
+            return;
+
+        boost::unordered_map<int, AncestorInfo> ancestors;
+        std::deque<PrevCurrInfo> tryNext;
```
geoffthemedio: try_next

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+            tryNext.push_back(PrevCurrInfo(*start_it, *start_it, *start_it));
+            ancestors.insert(std::make_pair(*start_it, AncestorInfo(*start_it, *start_it)));
+        }
+
+        // Find all reachable midpoints where paths from two different
+        // terminal points meet.
+        while (!tryNext.empty() ) {
+            const PrevCurrInfo & curr = tryNext.front();
+
+            std::pair<SupplyLaneMMap::const_iterator, SupplyLaneMMap::const_iterator> supplylane_endpoints
```
geoffthemedio: prefer variable name on same line as the value its assigned

```
std::pair<SupplyLaneMMap::const_iterator, SupplyLaneMMap::const_iterator>
    supplylane_endpoints = supply_lanes.equal_range(curr.curr);
```

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+                = supply_lanes.equal_range(curr.curr);
+            for (SupplyLaneMMap::const_iterator sup_it = supplylane_endpoints.first;
+                 sup_it != supplylane_endpoints.second; ++sup_it)
+            {
+                int next = sup_it->second;
+                if (next == curr.prev)
+                    continue;
+                boost::unordered_map<int, AncestorInfo>::iterator ancestor = ancestors.find(next);
```
geoffthemedio, line 2848: Is there a reason to prefer unordered_map vs. std::map given that you're searching in it?

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+                    // Same origin so add to ancestors.
+                    if (ancestor->second.single_origin
+                        && ancestor->second.single_origin == curr.origin) {
+                        ancestor->second.ids.push_back(curr.curr);
                     }
-                    return path;
+
+                    // Different origins so mark both as ends of good paths
+                    else if (ancestor->second.single_origin
+                        && ancestor->second.single_origin != curr.origin) {
```
geoffthemedio: { to next line

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+
+                    // Same origin so add to ancestors.
+                    if (ancestor->second.single_origin
+                        && ancestor->second.single_origin == curr.origin) {
+                        ancestor->second.ids.push_back(curr.curr);
                     }
-                    return path;
+
+                    // Different origins so mark both as ends of good paths
+                    else if (ancestor->second.single_origin
```
geoffthemedio: prefer else if on same line as the } from the preceeding if block

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+                    tryNext.push_back(PrevCurrInfo(curr.curr, next, curr.origin));
+                }
+
+                // Previously visited so modify older ancestor to merge
+                // paths/create end points.
+                else {
+
+                    // Same origin so add to ancestors.
+                    if (ancestor->second.single_origin
+                        && ancestor->second.single_origin == curr.origin) {
```
geoffthemedio: { to newlinea fter multi-line if test

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+
+                    // Different origins so mark both as ends of good paths
+                    else if (ancestor->second.single_origin
+                        && ancestor->second.single_origin != curr.origin) {
+                        // Single origin becomes multi origin and these
+                        // points are marked as midpoints.
+                        ancestor->second.single_origin = boost::none;
+                        reachable_midpoints.push_back(curr.curr);
+                        reachable_midpoints.push_back(next);
+                    }
```
geoffthemedio: same here... make it clear there's an else block by not having space between the end of the if {} and the else

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+        // Queue is exhausted.
+
+        // No terminal point has a path to any other terminal point.
+        if (reachable_midpoints.empty())
+            return;
+
+        // Return all systems on any path back to a terminal point.
+        boost::unordered_set<int> ancestors_on_path;
+        for (std::vector<int>::const_iterator reached_it = reachable_midpoints.begin();
+             reached_it != reachable_midpoints.end(); ++reached_it) {
```
geoffthemedio: { to newline

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+
+            boost::unordered_map<int, AncestorInfo>::const_iterator ancestor_ii_sys;
+            int ii_sys;
+            ancestors_on_path.insert(*reached_it);
+            while (!ancestors_on_path.empty()) {
+                ii_sys = *ancestors_on_path.begin();
+                ancestors_on_path.erase(ancestors_on_path.begin());
+                if ((ancestor_ii_sys = ancestors.find(ii_sys)) != ancestors.end()
+                       && (good_path.count(ancestor_ii_sys->first) == 0 )) {
```
geoffthemedio: ...

geoffthemedio: indenting of && is odd

UI/MapWnd.cpp
```diff
@@ -2879,82 +2974,81 @@ void MapWnd::InitStarlaneRenderingBuffers() {
...
-                }
-                if (placed_pool)
-                    break;
+            // Convert supply starlanes to non-directional.
+            SupplyLaneMMap resource_supply_lanes_undirected;
+            const std::set<std::pair<int, int> > resource_supply_lanes_directed
+                = GetSupplyManager().SupplyStarlaneTraversals(client_empire_id);
+
+            for (std::set<std::pair<int, int> >::const_iterator sp_it = resource_supply_lanes_directed.begin();
+                 sp_it != resource_supply_lanes_directed.end(); ++sp_it) {
```
geoffthemedio: ...

UI/MapWnd.cpp
```diff
@@ -2879,82 +2974,81 @@ void MapWnd::InitStarlaneRenderingBuffers() {
...
-                        res_pool_to_group_map[res_pool_sys_it->first] = *rg_it;
-                        placed_pool = true;
-                        break;
-                    }
-                }
-                if (placed_pool)
-                    break;
+            // Convert supply starlanes to non-directional.
+            SupplyLaneMMap resource_supply_lanes_undirected;
+            const std::set<std::pair<int, int> > resource_supply_lanes_directed
```
geoffthemedio: variable name to line with value

UI/MapWnd.cpp
```diff
@@ -3058,15 +3152,20 @@ void MapWnd::InitStarlaneRenderingBuffers() {
                 if ( (this_client_empire) &&(res_group_core_members.find(start_system->ID()) != res_group_core_members.end()))  {//start system is a res Grp core member for this_client_empire -- highlight
                     lane_colour = this_client_empire->Color();
                     float indicatorExtent = 0.5f;
-                    if (under_alloc_res_grp_core_members.find(start_system->ID()) != under_alloc_res_grp_core_members.end() ) {
+                    if (under_alloc_res_grp_core_members
+                        && (under_alloc_res_grp_core_members->find(start_system->ID()) != under_alloc_res_grp_core_members->end() ) ) {
```
geoffthemedio: ...

UI/MapWnd.cpp
```diff
@@ -3058,15 +3152,20 @@ void MapWnd::InitStarlaneRenderingBuffers() {
...
                         (this_client_empire->SupplyObstructedStarlaneTraversals().find(lane_backward) != this_client_empire->SupplyObstructedStarlaneTraversals().end()) ||
                         !( (this_client_empire->SupplyStarlaneTraversals().find(lane_forward) != this_client_empire->SupplyStarlaneTraversals().end()) ||
                         (this_client_empire->SupplyStarlaneTraversals().find(lane_backward) != this_client_empire->SupplyStarlaneTraversals().end())   )  ) */
-                    if (res_group_cores[ member_to_pool[start_system->ID()]] != res_group_cores[ member_to_pool[dest_system->ID()]])
+                    boost::unordered_map<int, boost::shared_ptr<std::set<int> > >::const_iterator start_core = member_to_core.find(start_system->ID());
+                    boost::unordered_map<int, boost::shared_ptr<std::set<int> > >::const_iterator dest_core = member_to_core.find(dest_system->ID());
+                    if (start_core != member_to_core.end() && dest_core != member_to_core.end()
+                        && (start_core->second != dest_core->second)
+                        && (*(start_core->second) != *(dest_core->second)))
```
geoffthemedio, line 3171: need {} after multi-line if statement

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,147 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+            }
+        }
+        return;
+    }
+
+
+    /** Look a \p kkey in \p mmap and if not found allocate a new
+        shared_ptr with the default constructor.*/
+    template <typename Map>
+    typename Map::mapped_type& lookup_or_make_shared(Map & mmap, typename Map::key_type const & kkey) {
```
geoffthemedio: remove space before & s

UI/MapWnd.cpp
```diff
@@ -2847,30 +2940,32 @@ void MapWnd::InitStarlaneRenderingBuffers() {
...
-    std::set<int>                           under_alloc_res_grp_core_members;
+    boost::unordered_map<int, boost::shared_ptr<std::set<int> > >   member_to_core;
+    boost::shared_ptr<std::set<int> >       under_alloc_res_grp_core_members;
 
     if (this_client_empire) {
-        const std::set<std::set<int> >& res_groups = GetSupplyManager().ResourceSupplyGroups(client_empire_id);
         const ProductionQueue& queue = this_client_empire->GetProductionQueue();
-        std::map<std::set<int>, float> allocatedPP(queue.AllocatedPP());
-        std::map<std::set<int>, float> availablePP(this_client_empire->GetResourcePool(RE_INDUSTRY)->Available());
+        const std::map<std::set<int>, float>& allocatedPP(queue.AllocatedPP());
```
geoffthemedio: should be allocated_pp (not your own variable name, but still)

UI/MapWnd.cpp
```diff
@@ -2879,82 +2974,81 @@ void MapWnd::InitStarlaneRenderingBuffers() {
                 //DebugLogger() << "Empire " << empire_id << "; Planet (" << object_id << ") is named " << planet->Name();
 
                 int system_id = planet->SystemID();
-                res_pool_systems[it->first].insert(system_id);
+                TemporaryPtr<const System> system = Objects().Object<const System>(system_id);
```
geoffthemedio: should be something like just GetSystem() ... less verbose

UI/MapWnd.cpp
```diff
@@ -2879,82 +2974,81 @@ void MapWnd::InitStarlaneRenderingBuffers() {
...
+
+        // For each pool of resources find all paths available through
+        // the supply network.
+
+        for (boost::unordered_map<std::set<int>, boost::shared_ptr<std::set<int> > >::iterator res_pool_sys_it = res_pool_systems.begin();
              res_pool_sys_it != res_pool_systems.end(); res_pool_sys_it++)
         {
-            for (std::set<std::set<int> >::const_iterator rg_it = res_groups.begin(); rg_it != res_groups.end(); ++rg_it) {
-                bool placed_pool = false;
-                for (std::set<int>::iterator sysIt=res_pool_sys_it->second.begin(); sysIt!=res_pool_sys_it->second.end(); sysIt++) {
```
geoffthemedio, line 2897: sys_it

spaces around =

geoffthemedio:
Is it necessary to recalculate all the supply groups every time the buffers are reinitialized? If just changing a group from fully to not fully using its PP, presumably it's not necessary to recalculate what the groups are?

geoffthemedio:
Seems basically functional, haven't tested thoroughly, though.

UI/MapWnd.cpp
LGM-Doyle: `tryNext` was a variable name from the original code that survived the refactoring.  Goodbye `tryNext`.

UI/MapWnd.cpp
LGM-Doyle, line 2848: Yes.

`unordered_set`, `unordered_map` and their `multimap` versions have average worst case linear complexity, `O(1)`, when inserting, searching, deleting, whereas `set`, `map` and their `multimap` version have logarithmic complexity <code>O(log<sub>2</sub>N)</code> for the same operations.  That is the price of `set` and `map` being ordered.

So assuming that the supply plus terminal point network has `N = 64` systems in it.  Every insert, find and delete will take on average <code>log<sub>2</sub>(64)</code> = 6 times longer if I use `set` instead of `unordered_set`.

`unordered_map` and `unordered_set` are usually the better option if you don't need the sorting behavior of `set` and `map`.

UI/MapWnd.cpp
LGM-Doyle: More variables names that had survived the refactoring unscathed.

LGM-Doyle:
Thanks for the prompt review.

I've addressed everything.

I also moved calculating the non-directional supply lanes out of the inner loop.

I did not know offhand if it is necessary to recalculate the supply groups every time that the buffers are initialized.  I did not investigate that.  It does seems that there is unnecessary work being done in this function.

I wanted this PR to be an apples to apples change and save compatible.   Then I could write test code and compare this code's output directly with the old output and be fairly certain that any changes were intentional.  And I could work from the same save file and compare the execution times.

The way `MapWnd`, `Empire` and `Supply` it is necessary to recalculate which lanes are part of the resource carrying portion of the supply network.  

However, I re-ran perf on the new code and `InitStarlaneRenderingBuffers()` is only 0.34% of the running time during production window updates.  I think that falls into not perfect, but good enough to focus on another area.

UI/MapWnd.cpp
geoffthemedio, line 2848: "average worst case" ?
"linear complexity, O(1)" ?

UI/MapWnd.cpp
```diff
@@ -2778,57 +2781,149 @@ void MapWnd::ClearSystemRenderingBuffers() {
...
+                    continue;
+                boost::unordered_map<int, AncestorInfo>::iterator ancestor = ancestors.find(next);
+
+                // Unvisited system, so create new ancestor.
+                if (ancestor == ancestors.end()) {
+                    ancestors.insert(std::make_pair(next, AncestorInfo(curr.curr, curr.origin)));
+                    try_next.push_back(PrevCurrInfo(curr.curr, next, curr.origin));
+                }
+
+                // Previously visited so modify older ancestor to merge
```
geoffthemedio, line 2855: another else well past the } of the if

UI/MapWnd.cpp
LGM-Doyle, line 2848: I should have said average worst case constant complexity `O(1)`.

I was thinking linear complexity `O(N)` for filling an `unordered_set` vs log linear complexity <code>O(Nlog<sub>2</sub>N)</code> for filling a `set` with the same data.

geoffthemedio:
Some additional / expanded comments in the code would be helpful... I can't really follow most of what it's doing without a lot more effort / attention than I want to put into it.

PR #809, opened 2016-07-29
fix fleetscreen scrollbar bug(s)

TheSilentOne1:
- refresh m_fleet_detail_panel when a different fleet is selected
- refresh m_fleets_lb when ships are dropped on m_drop_target
- moved FleetDetailPanel::Refresh from private to public

While this works as intended, I'd appreciate feedback on if there would have been a "cleaner" way to implement this, or if it is fine as is.

### DIFF
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -2059,6 +2059,12 @@ public:
         //std::cout << "FleetsListBox::DragDropLeave done" << std::endl << std::flush;
     }
 
+    void            Refresh() {
+        const GG::Pt row_size = ListRowSize();
+        for (GG::ListBox::iterator it = begin(); it != end(); ++it)
+            (*it)->Resize(row_size);
+    }
+
     virtual void    SizeMove(const GG::Pt& ul, const GG::Pt& lr) {
         const GG::Pt old_size = Size();
         CUIListBox::SizeMove(ul, lr);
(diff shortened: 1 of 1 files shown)


geoffthemedio:
What problem is this intended to fix?

TheSilentOne1:
1. Create a new fleet by dropping ships on the drop target. If there is not enough space for the new fleet in the listbox on the top, a scrollbar appears, but the lb on the top is not being resized accordingly.
2. When you select a fleet in the top lb that contains more ships than the bottom lb can display, and a scrollbar newly appears on the (bottom) ship lb, the lb is not resized.

Only noticed these issues myself when playing with large fleets.

geoffthemedio:
Resized horizontally or vertically?

MatGB:
horizontally, I'm getting this too and forget to mention it, sometimes the scrollbar completely blocks view of the aggression indicator/selector and regularly blocks off the shiptype name, which given my naming pattern can be a problem.

geoffthemedio:
not having to deal with issues like this is part of why stuff didn't resize to fill the space of a not-needed scrollbar.

seems fine, though possibly misses some cases when the contents of the fleetwnd change due to other events.

Vezzra:
So, this can be merged, right?

geoffthemedio:
"seems fine" -> yes

Vezzra:
Well, usually you go ahead and merge a PR when you approve, but didn't in this case, so I wanted to check back with you just to be safe before I do the merge myself. 😉

Issue #812, opened 2016-07-30
AI not invading with regular Troop ships


Morlic-fo:
Test build from 2016/07/26, AI scripts as of current Github state.

The AI seems not to invade even undefended planets despite having spare troop ships available.
However, I did see invasions happening via Orbital Troop bases (after colonizing some planet in the system).
1. Start Game as Humans, Elliptical shape, high starlanes, 10 planets, 1 AI, no monsters (so AI finds you quickly), High natives (for alternative invasion targets)
2. Just auto-pass turns

Result: After some time, the AI will find your planet, kill your ships and park its military fleet there. It will have some troop ships possibly spread around its empire but seemingly cant be bothered to move those in to actually take the planet.

Can people reproduce this or is my version somehow corrupted? 
If its reproducible, any idea when this bug could have been introduced? AI Invasion mechanics surely worked "some months" ago.

Morlic-fo:
Seems to be introduced by commit #721.

MatGB:
I have wondered if there was something up with development, and I've seen a couple instances of AIs simply not invading natives at all even when they've got nothing else to do and have the troops.

Can't confirm but doubt it's just you.

PR #813, opened 2016-07-30
Fix AI invasions

Morlic-fo:
Fix bug #812 previously introduced by commit #721.


Cjkjvfnby:
@Morlic-fo, thank you.

PR #814, opened 2016-07-30
AI weapon upgrade calculation

Morlic-fo:
This updates the AI calculation for species weapon modifier when designing ships.

Fixes the buggy part of #632.

### DIFF
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -80,7 +80,7 @@ INVALID_DESIGN_RATING = -999  # this needs to be negative but greater than MISSI
 
 # Potentially, not adding techs to AIDependencies is intended for testing purposes.
 # Therefore, chat the player only once to inform him about the issue to prevent spam.
-__raised_warnings = []
+_raised_warnings = set()
 
 # string constants for better readability of the cache
 WITH_UPKEEP = "considering fleet upkeep"
@@ -832,7 +832,10 @@ class ShipDesigner(object):
             self.production_time = max(self.production_time,
                                        local_time_cache.get(part.name, part.productionTime(fo.empireID(), self.pid)))
             partclass = part.partClass
-            capacity = part.capacity if partclass not in WEAPONS else _calculate_weapon_strength(part)
+            if partclass in WEAPONS:
+                capacity = self._calculate_weapon_strength(part, ignore_species=ignore_species)
+            else:  # TODO Hangar part modifiers
+                capacity = part.capacity
             if partclass in FUEL:
                 self.fuel += capacity
             elif partclass in ENGINES:
@@ -852,10 +855,8 @@ class ShipDesigner(object):
             elif partclass in ARMOUR:
                 self.structure += capacity
             elif partclass in WEAPONS:
-                if capacity in self.attacks:
-                    self.attacks[capacity] += 1
-                else:
-                    self.attacks[capacity] = 1
+                shots = self._calculate_weapon_shots(part, ignore_species=ignore_species)
+                self.attacks[capacity] = self.attacks.get(capacity, 0) + shots
             elif partclass in SHIELDS:
                 shield_counter += 1
                 if shield_counter == 1:
@@ -876,8 +877,6 @@ class ShipDesigner(object):
         if self.species and not ignore_species:
             weapons_grade, shields_grade, troops_grade = foAI.foAIstate.get_piloting_grades(self.species)
             self.shields = foAI.foAIstate.weight_shields(self.shields, shields_grade)
-            if self.attacks:
-                self.attacks = foAI.foAIstate.weight_attacks(self.attacks, weapons_grade)
             if self.troops:
                 self.troops = foAI.foAIstate.weight_attack_troops(self.troops, troops_grade)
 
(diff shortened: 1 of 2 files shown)


Vezzra:
Just administrative nitpicking: if this PR fixes buggy parts, shouldn't it be tagged with the "bug" instead the "enhancement" label, or do you have a specific reason for using "enhancement"? (Especially as #632 is tagged as "bug".)

Vezzra:
Makes sense. I just want to ensure that both an issue and the PR(s) addressing it are tagged consistently. Will therefore add the "enhancement" label also to #632.

default/python/AI/ShipDesignAI.py
```diff
@@ -80,7 +80,7 @@
 
 # Potentially, not adding techs to AIDependencies is intended for testing purposes.
 # Therefore, chat the player only once to inform him about the issue to prevent spam.
-__raised_warnings = []
+_raised_warnings = set()
```
Cjkjvfnby, line 83: I have feeling that I already have seen this changes somewhere.

default/python/AI/ShipDesignAI.py
```diff
@@ -1441,6 +1440,66 @@ def _partclass_in_design(self, partclass):
         """
         return any(part.partClass in partclass for part in self.parts)
 
+    def _calculate_weapon_strength(self, weapon_part, ignore_species=False):
```
Cjkjvfnby, line 1444: maybe split to two methods:

_calculate_absolute_weapon_strength(weapon_part)
_calculate_species_weapon_strength(absolute_weapon_strength, species)

default/python/AI/ShipDesignAI.py
```diff
@@ -1441,6 +1440,66 @@ def _partclass_in_design(self, partclass):
...
+
+        # species modifiers
+        if self.species and not ignore_species:
+            weapons_grade = foAI.foAIstate.get_piloting_grades(self.species)
+            species_modifier = AIDependencies.PILOT_DAMAGE_MODIFIER_DICT.get(weapons_grade, {}).get(weapon_name, 0)
+        else:
+            species_modifier = 0
+        return base_damage + total_tech_bonus + species_modifier
+
+    def _calculate_weapon_shots(self, weapon_part, ignore_species=False):
```
Cjkjvfnby, line 1472: I did not know that number of shots are modified too.

`good` vs `ultimate` have really big difference

default/python/AI/ShipDesignAI.py
```diff
@@ -1441,6 +1440,66 @@ def _partclass_in_design(self, partclass):
...
+        # tech modifier
+        tech_modifier = 0  # none implemented yet
+        return base_shots + species_modifier + tech_modifier
+
+    def _calculate_hangar_damage(self, hangar_part, ignore_species=False):
+        hangar_name = hangar_part.name
+        base_damage = hangar_part.secondaryStat
+        # species modifier
+        if self.species and not ignore_species:
+            weapons_grade, _, _ = foAI.foAIstate.get_piloting_grades(self.species)
```
Cjkjvfnby: may be cache grades as attribute?

default/python/AI/ShipDesignAI.py
```diff
@@ -1441,6 +1440,66 @@ def _partclass_in_design(self, partclass):
         """
         return any(part.partClass in partclass for part in self.parts)
 
+    def _calculate_weapon_strength(self, weapon_part, ignore_species=False):
+        weapon_name = weapon_part.name
+        base_damage = weapon_part.capacity
+
+        # tech modifiers
+        try:
```
Cjkjvfnby, line 1449: May be move such checks to function in `AIDependencies`? 

This is not related to decision making, this is check that developer did not fail.  

IMHO check at module import and fail looks better for me.  This is critical bug and should be found as soon as possible. 

PS. Not in this PR

default/python/AI/ShipDesignAI.py
```diff
@@ -1441,6 +1440,66 @@ def _partclass_in_design(self, partclass):
...
+            weapons_grade, _, _ = foAI.foAIstate.get_piloting_grades(self.species)
+            species_modifier = AIDependencies.PILOT_ROF_MODIFIER_DICT.get(weapons_grade, {}).get(weapon_name, 0)
+        else:
+            species_modifier = 0
+
+        # tech modifier
+        tech_modifier = 0  # none implemented yet
+        return base_shots + species_modifier + tech_modifier
+
+    def _calculate_hangar_damage(self, hangar_part, ignore_species=False):
```
Cjkjvfnby, line 1489: may be `_calculate_fighters_damage` ?

default/python/AI/ShipDesignAI.py
```diff
@@ -852,10 +855,8 @@ def update_stats(self, ignore_species=False):
             elif partclass in ARMOUR:
                 self.structure += capacity
             elif partclass in WEAPONS:
-                if capacity in self.attacks:
-                    self.attacks[capacity] += 1
-                else:
-                    self.attacks[capacity] = 1
+                shots = self._calculate_weapon_shots(part, ignore_species=ignore_species)
+                self.attacks[capacity] = self.attacks.get(capacity, 0) + shots
```
Cjkjvfnby: I am lost here.

We add number of shots to attack damage.
`self.attacks[capacity] += shots`

Cjkjvfnby:
`bug` or `enhancement`. It does not matter. PR tag matches with ticket tag is enough :)

PR idea looks OK for me.  Just some cosmetic notes.

default/python/AI/ShipDesignAI.py
Morlic-fo, line 83: Yeah most of the stuff here is based on some extensive work for the fighter branch which was partially relying on a new module. Didn't get to push the most relevant changes to master back then.

default/python/AI/ShipDesignAI.py
Morlic-fo, line 1472: IIRC, that's so far only implemented for the Flak Cannon in the fighters branch. But doesn't hurt to have the framework already in master and reduce the diff later on.

default/python/AI/ShipDesignAI.py
Morlic-fo: Hmm, could probably be queried for and stored when updating the species.

default/python/AI/ShipDesignAI.py
Morlic-fo, line 1449: Seems worth doing in the future.

default/python/AI/ShipDesignAI.py
Morlic-fo, line 1489: It's not strictly the same. In this function, the damage of a single part should be calculated.

How that damage of the hangar part then translates into final fighter damage should be covered in another part of the code, IMO as that depends on another implementation detail on the C++ side (which may still be subject to change in the future considering the feature is far from established).

default/python/AI/ShipDesignAI.py
Cjkjvfnby: And check if we still need `self.species` after that

default/python/AI/ShipDesignAI.py
Cjkjvfnby, line 1489: Name is confusing for me. Hangar is harmless creature that spits and shallows fighters.

Vezzra:
@Morlic-fo, as @Cjkjvfnby has given his ok (more or less), is there anything preventing this from being merged?

PR #815, opened 2016-07-31
Update AI industry modifiers

Morlic-fo:
- Update tech/building values to the current values
- Update Calculation so that effect priority in relation to species modifiers is considered.

Couldn't find any existing code where the actual research values were used, so leaving those out for now.


default/python/AI/AIDependencies.py
```diff
@@ -399,6 +406,22 @@
     "SHP_SOLAR_CONT",
 )
 
+
+# Industry modifiers per population - [[EARLY_PRIORITY]]: Affected by species modifiers
+INDUSTRY_EFFECTS_PER_POP_MODIFIED_BY_SPECIES = {
```
Cjkjvfnby, line 411: I have feeling that this should be replaced by `function(species)` (tag extraction can be done inside)

default/python/AI/ColonisationAI.py
```diff
@@ -1199,31 +1198,30 @@ def evaluate_planet(planet_id, mission_type, spec_name, empire, detail=None):
...
-        for tech in ind_tech_map:
-            if tech_is_complete(tech):
-                ind_mult += ind_tech_map[tech]
+        has_blackhole = (len(claimed_stars.get(fo.starType.blackHole, [])) > 0)
+        ind_tech_map_before_species_mod = AIDependencies.INDUSTRY_EFFECTS_PER_POP_MODIFIED_BY_SPECIES
+        ind_tech_map_after_species_mod = AIDependencies.INDUSTRY_EFFECTS_PER_POP_NOT_MODIFIED_BY_SPECIES
+
+        ind_mult = 1
+        for tech in ind_tech_map_before_species_mod:
+            if tech_is_complete(tech) and (not tech == AIDependencies.PRO_SINGULAR_GEN or has_blackhole):
```
Cjkjvfnby: tech != AIDependencies.PRO_SINGULAR_GEN

default/python/AI/AIDependencies.py
Morlic-fo, line 411: I don't understand what you mean with this. 

What should that function return and when would I call it? What tags are you talking about?

Cjkjvfnby, line 411: Sorry for tags, it is from other places.

Extract code with cycle `for tech in ind_tech_map_before_species_mod` to function.

Vezzra:
@Morlic-fo & @Cjkjvfnby: what's the status on this PR? What's left that needs to be addressed so it can be merged?

PR #816, opened 2016-07-31
Added the native species Lembala'Lam.

Deepsloth:
Presented here: http://www.freeorion.org/forum/viewtopic.php?f=7&t=10160


default/scripting/species/SP_LEMBALALAM.focs.txt
```diff
@@ -0,0 +1,123 @@
+Species
+    name = "SP_LEMBALALAM"
+    description = "SP_LEMBALALAM_DESC"
+    gameplay_description = "SP_LEMBALALAM_GAMEPLAY_DESC"
+    Native
+
+    tags = [ "ORGANIC" "AI_TAG_NO_INDUSTRY" "AI_TAG_GOOD_RESEARCH" "AI_TAG_BAD_POPULATION" "AI_TAG_AVERAGE_SUPPLY" "AI_GOOD_STEALTH"]
```
Morlic-fo: AI_GOOD_STEALTH -> AI_TAG_GOOD_STEALTH

AI_TAG_BAD_POPULATION is reserved for the standard case. Do not use it in this case.

Morlic-fo:
While the AI will not understand the 5 pop mechanic (would probably easiest be hardcoded), out of my head the relevant calculations do not matter for non-colonizers anyway.
The everyday stuff is, IIRC, directly queried from the planet meters, so no big deal.

default/scripting/species/SP_LEMBALALAM.focs.txt
```diff
@@ -0,0 +1,123 @@
...
+
+        EffectsGroup
+            description = "FIXED_LOW_POPULATION"
+            scope = Source
+            activation = Planet
+            priority = 130
+            effects = [
+                SetTargetPopulation value = 5 accountinglabel = "IMMORTAL_LABEL"
+                SetPopulation value = Target.TargetPopulation
+            ]
```
Morlic-fo, line 61: What is the interaction with Evacuation or Concentration Camp?

What is the interaction with Bombard-type weapons?

PR #821, opened 2016-08-01
Enable energy shipyard queueing with pending artificial blackhole

### LINKED ISSUE #820: Allow Energy shipyards to be built at systems where artifical black hole queued
[FreeOrion • View topic - Energy Compressor requirements tweak](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10142&p=84575#p84575)

The Energy Compressor and Solar Containment Unit shipyards can be built if an artificial black hole in enqueued on the same planet, ideally it should be any planet in the same system.

dbenage-cx:
Fixes #820

### DIFF
--- a/default/scripting/buildings/shipyards/ENERGY_COMP.focs.txt
+++ b/default/scripting/buildings/shipyards/ENERGY_COMP.focs.txt
@@ -14,7 +14,15 @@ BuildingType
         Or [
             Star type = [White Blue BlackHole]
             Enqueued type = Building name = "BLD_ART_BLACK_HOLE"
+            ContainedBy And [
+                Star type = Red
+                Contains And [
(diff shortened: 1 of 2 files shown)


Vezzra:
@freeorion/ai-team: I guess this might need some adjustments to the AI code so the AI can take advantage of that?

MatGB:
At some point, yes, it's an extension of the changes I made a few months back that the AI isn't using but was mostly to resolve player frustration for new colonies/captures, that went straight in because it wasn't going to be massively overbalanced. 

Plus, I've never seen the AI make a black hole, it'd be cool if they started but I susepct too many variables for it to be easy.

Morlic-fo:
What Mat said.

Issue #822, opened 2016-08-01
Psionics doesn't seem to work


l29ah:
The description suggests that by researching psionics and capturing telepatic species i'll have reduced tech costs, but nope, death rays are as expensive as usual.

MatGB:
By having a telepathic species the cost to research psionics is reduced (from 600 to 100), and also telepathic species are the only ones able to use some abilities/immune to some abilities.

It's an area that needs more work at some point but for now it definitely works as intended for me, the issue might be the description (or translation?) isn't clear, point us at where you read it and we can try to imrpvoe it.

l29ah:
Yes, i've misunderstood the description in the pedia, combined with the fact that telepathic species usually have a penalty to research and never a bonus.

PR #824, opened 2016-08-01
clarify psionics bonus

l29ah:
#822


Vezzra:
Sounds reasonable... but I leave final judgement to our native speakers.

MatGB:
Very reasonable, and reading the old version I can see whre the confusion could easily come from, thanks.

PR #825, opened 2016-08-01
enable keep-alives on the connection to freeoriond

### LINKED ISSUE #819: Enable TCP keep-alives plz
ISP NATs tend to forget about connections that doesn't have any data moving thru them for a while. That's why there's SO_KEEPALIVE sockopt. It should be enabled in the connections to freeoriond to avoid spurious disconnects.
// i'll try poing it in the meantime

l29ah:
fixes #819

### DIFF
--- a/network/ClientNetworking.cpp
+++ b/network/ClientNetworking.cpp
@@ -201,6 +201,7 @@ bool ClientNetworking::ConnectToServer(
                 else
                     DebugLogger() << "ClientNetworking::ConnectToServer : this client using xml serialization.";
                 m_socket.set_option(boost::asio::socket_base::linger(true, SOCKET_LINGER_TIME));
+                m_socket.set_option(boost::asio::socket_base::keep_alive(true));
                 DebugLogger() << "ClientNetworking::ConnectToServer : starting networking thread";
                 boost::thread(boost::bind(&ClientNetworking::NetworkingThread, this));
                 break;


Vezzra:
Looks simple enough, so I'll tag it for the release.

geoffthemedio:
Could you explain what making this change will do, and any potential side effects to consider / expect?

l29ah:
http://tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/#codeneeding
http://tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/#whatis

geoffthemedio:
any objections to doing this? anyone with disconnection issues found it to help?

adrianbroher:
> any objections to doing this?

I'm not quite sure yet.

The description of TCP Keep-Alive in [RFC 1122 4.2.3.6](https://tools.ietf.org/html/rfc1122#page-101) mandates some defaults from the TCP stack implementors, that are a bit far fetched from what I would consider 'good' or useful for solving the problem @l29ah  described in #819.  For example:

> Keep-alive packets MUST only be sent when no data or
> acknowledgement packets have been received for the
> connection within an interval.  This interval MUST be
> configurable and MUST default to no less than two hours.

So the first OS Keep-Alive message is sent at the earliest after two hours.  All of Windows, MacOSX and Linux default to this value.

Also there are configuration values for the number of retries when the first Keep-Alive message doesn't come through and the interval between those retries.

All those values need to be configured for our game and as far as I investigated those parameters are not exposed in the boost asio library.

geoffthemedio:
Would it then make more sense to have an in-freeorion "ping" message be sent every minute or 5 instead?

I tried a bit to set up some timing stuff for networking code, such as auto-ending turns by the server after some limit, but I couldn't get it working properly with the FSM. Should be doable, though.

l29ah:
On Fri, Aug 19, 2016 at 03:21:04AM -0700, Marcel Metz wrote:

> > any objections to doing this?
> 
> I'm not quite sure yet.
> 
> The description of TCP Keep-Alive in [RFC 1122 4.2.3.6] mandates some defaults from the TCP stack implementors, that are a bit far fetched from what I would consider 'good' or useful for solving the problem @l29ah  described in #819.  For example:
> 
> > Keep-alive packets MUST only be sent when no data or
> > acknowledgement packets have been received for the
> > connection within an interval.  This interval MUST be
> > configurable and MUST default to no less than two hours.
> 
> So the first OS Keep-Alive message is sent at the earliest after two hours.  All of Windows, MacOSX and Linux default to this value.

It can be adjusted system-wide to suit one's needs.

> Also there are configuration values for the number of retries when the first Keep-Alive message doesn't come through and the interval between those retries.
> 
> All those values need to be configured for our game and as far as I investigated those parameters are not exposed in the boost asio library.

And should not be configured per app, yet there's an interface for this in linux at least.

## 

adrianbroher:
> And should not be configured per app […]

Who else should know the needs of the application?

l29ah:
On Fri, Aug 19, 2016 at 04:04:13AM -0700, Marcel Metz wrote:

> > And should not be configured per app […]
> 
> Who else should know the needs of the application?

Those are the needs of every application that communicate from behind the crappy NAT, and should be configured globally as they have the same limitations.

## 

adrianbroher:
> Those are the needs of every application that communicate from behind the crappy NAT, and should be configured globally as they have the same limitations.

You have a point here.

> Would it then make more sense to have an in-freeorion "ping" message be sent every minute or 5 instead?

Well, nothing prevents us to implement both solutions (enabling TCP Keep-Alive and adding Ping/Pong messages to the the network code).

geoffthemedio:
OK, then given the relative triviality of the actual changes in this pull request, is there a reason not to merge it?

Vezzra:
Especially considering our tight schedule for the release, I might add. Implementing our own ping/pong messages in addition to that, while probably not overly complicated, sounds a little bit more work than just adding one line.

Which of course we can (and probably should) do, but maybe post 0.4.6?

adrianbroher:
> OK, then given the relative triviality of the actual changes in this pull request, is there a reason not to merge it?

No reason, merging now.

PR #830, opened 2016-08-02
High and moderate tech specials will be removed when the species is g…

Deepsloth:
…one.


Deepsloth:
When a native species with the moderate tech natives special is wiped out by bombing, the special will remain. This patch fixes that.

MatGB:
Makes sense.

PR #834, opened 2016-08-03
Fix experimentor system sunders galaxy

LGM-Doyle:
This addresses the issue brought up in [this post](http://freeorion.org/forum/viewtopic.php?f=2&t=10112), that the experimentor spawn sometimes divides the galaxy into multiple disconnected pieces.

This PR prevents that by checking if the galaxy would be multiple disconnected sets before placing the experiementor system.


default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+
+
+def populate_monster_fleet(fleet_plan, system):
+    """
+    Create a monster fleet in ''system'' according to ''fleet_plan''
+    """
+
+    # create monster fleet
+    monster_fleet = fo.create_monster_fleet(system)
+    # if fleet creation fails, report an error and try to continue with next system
```
Vezzra: This comment needs to get adjusted, we don't continue with the next system here, but throw an exception.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+    if monster_fleet == fo.invalid_object():
+        errstr = "Python generate_monsters: unable to create new monster fleet %s" % fleet_plan.name()
+        util.report_error(errstr)
+        raise RuntimeError(errstr)
```
Vezzra: If you throw an exception, don't use `report_error` here (which will write to the error log), but on the level where you handle the exception. Currently this causes the error to be logged twice.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+
+    # add monsters to fleet
+    for design in fleet_plan.ship_designs():
+        # create monster, if creation fails, report an error and try to continue with the next design
```
Vezzra: Similar to the comment above, even more so, as in the above case the calling code actually continues with the next system (so the comment isn't strictly wrong, just in the wrong place), but here we don't proceed with the next design, but quit by throwing.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+        if fo.create_monster(design, monster_fleet) == fo.invalid_object():
+            errstr = "Python generate_monsters: unable to create monster %s" % design
+            util.report_error(errstr)
+            raise RuntimeError(errstr)
```
Vezzra: See above.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+
+    print "Spawn", fleet_plan.name(), "at", fo.get_name(system)
+
+monsters_that_alter_starlanes = {"SM_EXP_OUTPOST"}
```
Vezzra: The placement of this statement is a bit odd. It's a (module) global variable you initialize here, I'd expect that to be placed after the section with the import statements above, before the global functions.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+    # Find the Experimentor output fleet plan
+    plans = [plan for plan in fo.load_monster_fleet_plan_list()
+             if (plan.name() in monsters_that_alter_starlanes
+                 and plan.spawn_rate() > 0.0
+                 and plan.spawn_limit() > 0
+                 and plan.ship_designs())]
+    if not plans:
+        return
+
+    plan = plans[0]
```
Vezzra: Correct me if I'm wrong, but won't this code break if someone decides to script another starlane altering monster and just adds its name to the `monsters_that_alter_starlanes`, expecting this to be sufficient to work? Especially since most of the code you wrote here seems to be suited to allow for multiple kinds of starlane altering monsters...

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+    spawn_limit = plan.spawn_limit()
+    placed = []
+    removed_lanes = set()
+
+    starlanes = [(s1, s2) for s1 in systems for s2 in fo.sys_get_starlanes(s1) if s1 < s2]
+
+    for system in systems:
+        # Quit when no more to spawn
+        if len(placed) >= spawn_limit:
+            return
```
Vezzra: Implementing experimentor placement like that will cause different handling of the spawn limit parameter. The spawn limit is supposed to be the _max_ limit, not the actual number of fleets to place. Unless I've reading your code wrong, this will cause `spawn_limit` experimentor monsters to be placed (or at least try to, provided that enough suitable systems are present).

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+
+        if num_contiguous_regions > expected_num_contiguous_regions:
+            continue
+
+        if num_contiguous_regions < expected_num_contiguous_regions:
+            print >> sys.stderr, ("Number of contiguos regions %d is below the expected number "
+                                  "of contiguous regions %d when placing %d monster %s that can "
+                                  "break starlanes."
+                                  % (num_contiguous_regions, expected_num_contiguous_regions,
+                                     len(placed) + 1, plan.name()))
```
Vezzra: In the universe generation scripts, _always_ use `util.report_error` to log an error. This takes care of writing the error message to the log and also ensures that the player will get a proper warning at the start of the game that something during universe generation went wrong, and gameplay is potentially impaired (which won't happen if errors are only written to the log via `print`).

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+            continue
+
+        # create monster fleet
+        try:
+            populate_monster_fleet(plan, system)
+            placed.append(system)
+            removed_lanes |= local_lanes
+
+        except RuntimeError as e:
+            print >> sys.stderr, str(e)
```
Vezzra: Same here, and all other cases where you only use `print >> sys.stderr`.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,110 @@
...
+                                  "of contiguous regions %d when placing %d monster %s that can "
+                                  "break starlanes."
+                                  % (num_contiguous_regions, expected_num_contiguous_regions,
+                                     len(placed) + 1, plan.name()))
+            continue
+
+        # create monster fleet
+        try:
+            populate_monster_fleet(plan, system)
+            placed.append(system)
```
Vezzra: Again, correct me if I'm wrong, but won't that code break if someone just raises the spawn limit for the experimentor monster? It might be highly unlikely, but `placed` is a list, and in case two or more experimentor monsters are placed within the same system, `len(placed)` will count the same system multiple times, and your calculation of `expected_num_contiguous_regions` will go wrong, won't it?

MatGB: Definitely don't want things to break if someone wants more than one Experimentor system, I've been known to play with multiples just for the fun of it on occasions, and have also upped the limit for testing purposes, etc.

Vezzra:
The basic idea (using your disjoint map regions checker code here to check if a potential experimentor placing would segment the map) is good, but separating "normal" monster placement and experimentor placement like you did isn't ideal IMO.

I took your code and changed it a bit, so the resulting behaviour is closer to the original code, and (hopefully) more robust if someone decides to tinker with the experimentor spawn limit or add in more starlane changing monster types.

I branched off your commit d8da9ee5b4c03e860faf859ded9e6b2dcb38f2ee, leaving out the PEP8 adjustments and also didn't apply some of the corrections I mentioned above, to limit the differences between your implementation and mine. Hopefully this helps when you compare what I came up with your code.

I've provided my implementation as a PR: #840. It's not intended for merge, just so you can take a look at what I think is a better approach. You can use that as a basis to adjust/rewrite your PR, if you think that approach is suitable.

Vezzra:
Another thing to mention: I see both our approaches just as some kind of hack to fix the problem at hand for the release. Long term a more thorough solution is due, where the entire experimentor code (placement during universe generation as well as behaviour during the game) should be moved completely into the Python scripts (universe generation and turn events). FOCS isn't really suited to implement that kind of thing.

LGM-Doyle:
I agree with all of your comments except one.

The original PR did work with multiple experimentor systems, since that is how I tested it.  

It did not work with multiple starlane altering fleets of multiple types.  It does now and it is tested.

I have addressed all of the issues in the updated PR.

I cherry-picked your commit from PR #840 and continued from there.  I made a commit message out of your comments in the PR.  Please change that if it is inadequate.

I added the tag CanAlterStarlanes to the MonsterFleetPlan FOCS, which indicates that the monster can alter starlanes.

I added MapGenerationError as a custom Runtime error for universe generation to prevent over catching errors.

**Aside 1:**
I also viewed this as a temporary hack, not because I have any plans for the experimentors, but I wanted to prevent anyone from committing code here and then finding that I'm sitting on changes that might overwrite their work.

FOCS has grown into a Turing complete language.  Although it may not be suited to this kind of thing, if it can be done people will do it.

**Aside 2:**
Speaking of "configuration files", there are at least 8 locations for/types of "configuration files":  FOCS, config.xml, persistent_config.xml, universe_tables.py, AIDependencies.py, EnumsAI.py, buildings.inf and config.ini.  They use at least 4 different syntaxes: FOCS, XML, python, ini and line terminated double quote delineated file of strings. Some are not visible to the server.  And some duplicate or override information from other configuration files or the server.

Are things supposed to be moving in one direction or another?

default/python/universe_generation/monsters.py
```diff
@@ -22,7 +96,7 @@ def generate_monsters(monster_freq, systems):
 
     # get all monster fleets that have a spawn rate and limit both > 0 and at least one monster ship design in it
     # (a monster fleet with no monsters in it is pointless) and store them in a list
-    fleet_plans = fo.load_monster_fleet_plan_list()
+    fleet_plans = [fp for fp in fo.load_monster_fleet_plan_list()]
```
Cjkjvfnby: `fleet_plans = list(fo.load_monster_fleet_plan_list())`

default/python/universe_generation/monsters.py
```diff
@@ -33,7 +107,8 @@ def generate_monsters(monster_freq, systems):
                     if fp.spawn_rate() > 0.0 and fp.spawn_limit() > 0 and fp.ship_designs()}
 
     # map nests to monsters for ease of reporting
-    nest_name_map = dict(zip(["KRAKEN_NEST_SPECIAL", "SNOWFLAKE_NEST_SPECIAL", "JUGGERNAUT_NEST_SPECIAL"], ["SM_KRAKEN_1", "SM_SNOWFLAKE_1", "SM_JUGGERNAUT_1"]))
+    nest_name_map = dict(zip(["KRAKEN_NEST_SPECIAL", "SNOWFLAKE_NEST_SPECIAL", "JUGGERNAUT_NEST_SPECIAL"],
```
Cjkjvfnby: Cleanup?

Merge it to master and rebase.   

PS. convert them to list of pairs to avoid zip

default/python/universe_generation/monsters.py
Vezzra: Um, what's wrong with:

```
fleet_plans = fo.load_monster_fleet_plan_list()
```

...the simple original line?

default/python/universe_generation/universe_generator.py
```diff
@@ -132,6 +132,7 @@ def create_universe(psd_map):
     seed_rng(seed_pool.pop())
     generate_natives(gsd.native_frequency, systems, home_systems)
 
+
```
Vezzra: Minor nitpick: no need for two blank lines here, guess that somehow slipped in unintentionally...

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,82 @@
 import random
 import freeorion as fo
 import util
+from util import MapGenerationError
```
Vezzra: Hm, if we import `MapGenerationError` that way, it's probably more consistent if we import the rest of the stuff from `util` the same way.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,82 @@
...
+        self.removed_lanes = set()
+        self.starlanes = [(s1, s2) for s1 in systems for s2 in fo.sys_get_starlanes(s1) if s1 < s2]
+
+    def can_place_at(self, system, plan):
+        # Check if the starlane altering monster fleet ''plan'' can be placed at ''system''
+        # without disjoining the galaxy map.
+        # Compute the disjoint set of the galaxy without the starlanes
+        # from the proposed system.  Return False if there will be more
+        # connected regions than the number of placed starlane altering
+        # monsters plus one, otherwise True.
```
Vezzra: That should probably turned into docstring format. My bad, have been very sloppy with comments when I put together my alternative implementation.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,82 @@
...
+                              "of contiguous regions %d when placing %d monster %s that can "
+                              "break starlanes."
+                              % (num_contiguous_regions, expected_num_contiguous_regions,
+                                 len(self.placed) + 1, plan.name()))
+            return False
+
+        return True
+
+    def place(self, system, plan):
+        # create map altering monster fleet ''plan'' at ''system''.
```
Vezzra: Same as above.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,82 @@
...
+    """
+    Create a monster fleet in ''system'' according to ''fleet_plan''
+    """
+
+    # create monster fleet
+    monster_fleet = fo.create_monster_fleet(system)
+
+    if monster_fleet == fo.invalid_object():
+        errstr = "Python generate_monsters: unable to create new monster fleet %s" % fleet_plan.name()
+        raise MapGenerationError(errstr)
```
Vezzra: Minor nitpick: `errstr` isn't really necessary anymore, you can pass the string literal directly to `MapGenerationError`. Same for the other cases below.

default/python/universe_generation/monsters.py
```diff
@@ -1,8 +1,82 @@
...
+        # Check if the starlane altering monster fleet ''plan'' can be placed at ''system''
+        # without disjoining the galaxy map.
+        # Compute the disjoint set of the galaxy without the starlanes
+        # from the proposed system.  Return False if there will be more
+        # connected regions than the number of placed starlane altering
+        # monsters plus one, otherwise True.
+        local_lanes = {(min(system, s), max(system, s)) for s in fo.sys_get_starlanes(system)}
+
+        dsets = DisjointSets()
+        for s3 in self.systems:
```
Cjkjvfnby: s3?

default/python/universe_generation/monsters.py
Cjkjvfnby: May be just print to log error here and don't use that Exception?

Vezzra:
> The original PR did work with multiple experimentor systems, since that is how I tested it.

Sorry if I phrased my comment a bit wrong. What I meant wasn't that the original code did not work with multiple experimentor systems at all, I just wondered if it handled the very unlikely edge case where the same system would be selected multiple times during the placement process (because `placed` has been a list before, and you didn't check if a system is already in the list before adding it, thus potentially screwing the calculation of the expected disjointed regions).

Moot point, as your updated code now takes care of that.

> I made a commit message out of your comments in the PR. Please change that if it is inadequate.

The commit messages look fine to me.

> I added the tag CanAlterStarlanes to the MonsterFleetPlan FOCS, which indicates that the monster can alter starlanes.

Ok, the basic idea is actually a very good and sound one - **_but**_ the way you implemented it is IMO absolutely overkill and also doesn't really fit with how those directives are supposed to work in FOCS.

The token you added has no actual function, it's just a flag that can be queried. For such purposes a dedicated "auxiliary" parser already exists, the tags parser. It's currently used by the field, species and tech parsers, see there how to incorporate it. It enables the scripter to attach any number of custom string tags to a definition (be it field, species, tech or whatever we decide to add that tag thing to). Far more flexible and much simpler than hardcode a token for each flag we want to provide to scripters, doesn't limit them to a fixed set of flags etc.

Furthermore, the hardcoded tokens are supposed to control functionality (in the broadest sense, that includes things like just linking text, labels or names). They are not meant to be pure flags (meaning, not even switches like `CanColonize` and `CanProduceShips`). So a `CanAlterStarlanes` token might even be a bit confusing for scripters, because someone might infer from how all the other tokens work, that they can enable a monster fleet to create or destroy starlanes just by adding that token to a monster fleet definition.

The `tags` token however is supposed to provide exactly that: a custom list of arbitrary string tags that can easily be queried by FOCS and the Python scripts.So I'd recommend using that instead of a hardcoded `CanAlterStarlanes` token.

It might probably even be best to add the tags token to the ship design parser (or even the ship hull parser?) and then instead of checking the monster fleet plan iterate over the actual monsters in the fleet plan and check them for the existence of the tag. A bit more complicated, but IMO more consistent, intuitive and less prone to scripting errors (as it's the monster itself that contains the starlane altering code, so the tag should also belong there - makes it easier if you use such a monster in more than one fleet plan, as you don't have to tag each fleet plan, which can easily be overseen).

> I added MapGenerationError as a custom Runtime error for universe generation to prevent over catching errors.

Good idea, probably should use that in the rest of the universe generation scripts too (instead of using the default `Exception` object). But that's not subject of this PR of course.

> Speaking of "configuration files", there are at least 8 locations for/types of "configuration files" ... Are things supposed to be moving in one direction or another?

Good point. But that discoussion belongs on the forum, [I opened a corresponding thread](http://www.freeorion.org/forum/viewtopic.php?f=9&t=10170), lets continue there.

Vezzra:
Regarding the docstring/comments: that's actually my fault: when I took your code and rearranged things into my alternative approach, I didn't really bother with correct docstrings and comments. I completely ignored the former, and only threw in a bare minimum of the latter. That should be corrected, but I do not expect you to clean up my mess. So it's fine with me if you leave this as it is, and I will go in after merging and clean things up.

parse/MonsterFleetPlansParser.cpp
```diff
@@ -31,8 +31,9 @@ namespace {
 #endif
 
         MonsterFleetPlan* operator()(const std::string& fleet_name, const std::vector<std::string>& ship_design_names,
-                                     double spawn_rate, int spawn_limit, Condition::ConditionBase* location) const
-        { return new MonsterFleetPlan(fleet_name, ship_design_names, spawn_rate, spawn_limit, location); }
+                                     double spawn_rate, int spawn_limit, bool can_alter_starlanes,
+                                     Condition::ConditionBase* location) const
+        { return new MonsterFleetPlan(fleet_name, ship_design_names, spawn_rate, spawn_limit, location, can_alter_starlanes); }
```
Cjkjvfnby: @Vezzra, how about merging c++ part right now? Is this contains all we might need?

Vezzra: No, you need the entire package, and anyway particularly this part is something I hope @LGM-Doyle agrees to implement differently.

default/python/universe_generation/monsters.py
LGM-Doyle: Sorry.  Filtering was added.  Filtering was removed.  Original line was not restored.

Original line is now restored.

default/python/universe_generation/monsters.py
LGM-Doyle: Raising the exception here means that the level of encapsulation above that can't deal cleanly with the exceptional condition can ignore it.

`StarlaneAlteringMonsters::place()` uses `populate_monster_fleet()` without checking any error condition.  Catching the exception, reporting the error and continuing with the next monster is all handled with the single except clause at the end of the for systems loop in generate_monsters.

LGM-Doyle:
Regarding the docstrings/comments:  You clearly marked PR #840 as not intended for merging and I still used the commit.  I realize that you intended it as a reference, but I felt more comfortable using your code with a commit with your name on it.  I fixed the docstrings.

The tag idea is the better approach.  

Tags were already part of the ShipHull parser so I just needed to feed them back to monsters.py.

I made all the requested changes, except changing the one exception to report_error as noted above.

default/python/universe_generation/monsters.py
Cjkjvfnby: Think about that from other side. 

You just bring new approach for this projects. You create first custom exception. Make first handler of custom exception.

For me this approach is ok, but I don't think that we should bring it now.  You will not save any lines on it.

Vezzra: Well, it also can't hurt, and we have to start somewhere sometimes, so why not here and now? The custom exception is something I should have done from the beginning, just have been to lazy 😉 

And it's not that we really have been avoiding exceptions consequently and consistently throughout our code. So I'm fine with leaving this part as it is.

universe/UniverseGenerator.h
```diff
@@ -42,7 +42,7 @@ class FO_COMMON_API MonsterFleetPlan : public FleetPlan {
 public:
     MonsterFleetPlan(const std::string& fleet_name, const std::vector<std::string>& ship_design_names,
                      double spawn_rate = 1.0, int spawn_limit = 9999, const Condition::ConditionBase* location = 0,
-                     bool lookup_name_userstring = false) :
+                     bool can_alter_starlanes = false, bool lookup_name_userstring = false) :
```
Vezzra: Just to be sure: I guess you just missed that line when you removed the `CanAlterStarlanes` token and we don't need that additional parameter (`can_alter_starlanes`) here anymore, right?

python/UniverseWrapper.cpp
```diff
@@ -146,6 +146,16 @@ namespace {
...
 
+    const ShipDesign*       GetShipDesignByName(const Universe& universe, const std::string& name) {
+        Universe::ship_design_iterator design_it = universe.beginShipDesigns();
+        while (design_it != universe.endShipDesigns()) {
+            if (design_it->second->Name(false) == name)
+                return design_it->second;
+            ++design_it;
+        }
+        return 0;
+    }
```
Vezzra: Is there a specific reason why you don't just expose `Universe::GetGenericShipDesign`, which (at least to my eyes) does exactly the same as your helper function here?

Vezzra:
Looks fine, just the two minor quibbles above to address, then we should be ready to go.

python/UniverseWrapper.cpp
```diff
@@ -146,6 +146,9 @@ namespace {
     const Ship::PartMeterMap&
                             (Ship::*ShipPartMeters)(void) const =                               &Ship::PartMeters;
 
+    const ShipDesign*       GetGenericShipDesign(const Universe& universe, const std::string& name)
+    { return universe.GetGenericShipDesign(name); }
```
Vezzra, line 150: Just out of curiosity: why still stick to a wrapper function instead of just exposing the member function `GetGenericShipDesign` of the `Universe` class?

I'm asking because I'm confused about all those different ways member functions get exposed to Python in FO, and I wonder if you actually know more about this than me. There are member functions that get directly exposed (e.g. `Universe::SystemHasVisibleStarlanes`), others have these wrappers, and still others use this `make_function` function.

Which is used for what and why? The Boost Python documentation isn't partucularly enlightening. In some cases I can guess at the reasons, but others still leave me clueless. In this case (`Universe::GetGenericShipDesign`) I would have just exposed the member function directly...

LGM-Doyle:
I do not know more than you.  I read just enough of the `boost::python` documentation to make my changes.

I should have directly exposed the function.  A wrapper that adds no value is a questionable choice.

Thanks for reviewing and merging the PR.

Vezzra:
> I do not know more than you. I read just enough of the `boost::python` documentation to make my changes.

Just the same that I did. So I guess we're in the same boat then... 😉 

> I should have directly exposed the function. A wrapper that adds no value is a questionable choice.

No problem, [fixed](https://github.com/freeorion/freeorion/commit/b937e9d9c0857300a46c205cddcdf4f0b849c262).

> Thanks for reviewing and merging the PR.

You're welcome 😃

PR #837, opened 2016-08-05
Added Ancient Guardians species as an alternative to Sentries guardin…

Deepsloth:
…g specials.

### DIFF
--- a/default/scripting/specials/specials.macros
+++ b/default/scripting/specials/specials.macros
@@ -20,7 +20,7 @@ CHANCE_OF_GUARD_1
             scope = Source
             activation = AND [
                 Turn high = 0
-                Random probability = 0.60
+                Random probability = 0.70
                 (GalaxyMaxAIAggression >= 1)
                 (GalaxyMonsterFrequency >= 1)
                 Not ContainedBy Contains OR [
@@ -29,7 +29,16 @@ CHANCE_OF_GUARD_1
                     And [ Planet HasSpecial name = "HIGH_TECH_NATIVES_SPECIAL" ]
                 ]
             ]
-            effects = CreateShip designname = OneOf("SM_GUARD_0", "SM_GUARD_1")
+            effects = [
+                If condition = Random probability = 0.6
+                    effects = [
+                        CreateShip designname = OneOf("SM_GUARD_0", "SM_GUARD_1")
+                    ]
+                else = [
+                    SetSpecies name = "SP_ANCIENT_GUARDIANS"
+                    SetPopulation value = Target.Population
+                ]
+            ]
 '''
 
 // NOTE: Unused
--- a/default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
+++ b/default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
@@ -1,0 +1,81 @@
+Species
+    name = "SP_ANCIENT_GUARDIANS"
+    description = "SP_ANCIENT_GUARDIANS_DESC"
+    gameplay_description = "SP_ANCIENT_GUARDIANS_GAMEPLAY_DESC"
+
+    tags = [ "ROBOTIC" "AI_TAG_NO_INDUSTRY" "AI_TAG_NO_RESEARCH" "AI_TAG_AVERAGE_SUPPLY"]
+
+    foci = [
+        [[HAS_INDUSTRY_FOCUS]]
+        [[HAS_RESEARCH_FOCUS]]
+        [[HAS_GROWTH_FOCUS]]
+        [[HAS_MINING_FOCUS]]
+        [[HAS_ADVANCED_FOCI]]
+    ]
+
+    effectsgroups = [
+        [[NO_INDUSTRY]]
+        [[NO_RESEARCH]]
+
+        [[AVERAGE_POPULATION]]
+        [[AVERAGE_HAPPINESS]]
+        [[AVERAGE_SUPPLY]]
+        [[AVERAGE_DEFENSE_TROOPS]]
+
+        // not for description
+
+        [[AVERAGE_PLANETARY_SHIELDS]]
+        [[AVERAGE_PLANETARY_DEFENSE]]
+
+        // huge bonus to planetary shields
+        EffectsGroup
+            description = "ANCIENT_PLANETARY_SHIELD"
+            scope = And [
+                Source
+                Planet
+            ]
+            effects = SetMaxShield value = Value + 500
+
+        // huge bonus to ground troops
+        EffectsGroup
+            description = "ANCIENT_DEFENSE_TROOPS"
+            scope = And [
+                Source
+                Planet
+            ]
+            effects = SetMaxTroops value = Value + Target.Population * 10
+
+        // self-destruct when captured
+        EffectsGroup
+            scope = Source
+            activation = OwnedBy affiliation = AnyEmpire
+            effects = [
+                SetPopulation value = 0
+                GenerateSitRepMessage
+                    message = "EFFECT_ANCIENT_GUARDIANS_CAPTURED"
+                    icon = "icons/species/robotic-08.png"
+                    parameters = tag = "planet" data = Source.ID
+                    empire = Source.Owner
+        ]
+
+    ]
+
+    environments = [
+        type = Swamp        environment = Good
+        type = Toxic        environment = Good
+        type = Inferno      environment = Good
+        type = Radiated     environment = Good
+        type = Barren       environment = Good
+        type = Tundra       environment = Good
+        type = Desert       environment = Good
+        type = Terran       environment = Good
+        type = Ocean        environment = Good
+        type = Asteroids    environment = Good
+        type = Gasgiant     environment = Good
+    ]
+
+    graphic = "icons/species/robotic-08.png"
+
+#include "common/*.macros"
+
+#include "/scripting/common/*.macros"
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -5129,6 +5129,11 @@ On %planet%: Ancient Ruins were excavated, but only tantalizing hints of ancient
 EFFECT_ANCIENT_RUINS_EMPTY_LABEL
 Ancient Ruins Empty Excavation
 
+EFFECT_ANCIENT_GUARDIANS_CAPTURED
+On %planet%: The [[species SP_ANCIENT_GUARDIANS]] self-destructed when the planet was captured!
+EFFECT_ANCIENT_GUARDIANS_CAPTURED_LABEL
+Ancient Guardians seld destruct
+
 EFFECT_CONC_CAMP_COMLETE
 At %planet%: depopulation is complete.
 EFFECT_CONC_CAMP_COMLETE_LABEL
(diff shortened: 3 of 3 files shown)


Deepsloth:
Discussion: http://www.freeorion.org/forum/viewtopic.php?f=15&t=10162

Post v0.4.6

default/stringtables/en.txt
```diff
@@ -5129,6 +5129,11 @@ On %planet%: Ancient Ruins were excavated, but only tantalizing hints of ancient
 EFFECT_ANCIENT_RUINS_EMPTY_LABEL
 Ancient Ruins Empty Excavation
 
+EFFECT_ANCIENT_GUARDIANS_CAPTURED
+On %planet%: The [[species SP_ANCIENT_GUARDIANS]] self-destructed when the planet was captured!
+EFFECT_ANCIENT_GUARDIANS_CAPTURED_LABEL
+Ancient Guardians seld destruct
```
MatGB: self

default/stringtables/en.txt
```diff
@@ -6197,6 +6202,18 @@ SP_EXOBOT_DESC
 '''While not as efficient as fully self-directed citizens, Exobots can colonize otherwise inhospitable planets, and build more Exobots.
 '''
 
+SP_ANCIENT_GUARDIANS
+Ancient Guardians
+SP_ANCIENT_GUARDIANS_GAMEPLAY_DESC
+'''Guardian Robots constructed by the Precursor civilizations aeons ago
```
MatGB: Better to say "a Precursor", we've got several in game already, no need to say which one.

default/stringtables/en.txt
```diff
@@ -6197,6 +6202,18 @@ SP_EXOBOT_DESC
...
+[[encyclopedia ROBOTIC_SPECIES_TITLE]]
+Will self-destruct when conquered
+'''
+
+SP_ANCIENT_GUARDIANS_DESC
+'''Constructed by precusor technology these ancient machines can adapt to and fight in any kind of environment. With unwavering loyalty and duteousness they have protected the planet from any invaders, never wasting time to repair and improve the fortifications. Upon being captured every Guardian Robot will automatically self-destruct.
```
MatGB: I dislike "duteousness", I had to look up whether it was an actual word and it's very rarely used (so much so Google won't even give me an ngram of it, insisting to give me results for duteous instead). "Commitment" might be better, not sure tho.

MatGB:
These do look very fun. Not sure about troops as a multiplier of target population, that prevents partial destruction from bombardment which ought to be a valid tactic, multiplier of current pop?

(it is of course another special giving Egassem a potential early advantage...)

Deepsloth:
Troops are not dependent on target population. "Target.Population" is the current population of the target. The other would be "Target.TargetPopulation". I love FOCS.

MatGB:
Good point, yeah, see, still not grokking FOCS properly.

default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
```diff
@@ -0,0 +1,81 @@
...
+    description = "SP_ANCIENT_GUARDIANS_DESC"
+    gameplay_description = "SP_ANCIENT_GUARDIANS_GAMEPLAY_DESC"
+
+    tags = [ "ROBOTIC" "AI_TAG_NO_INDUSTRY" "AI_TAG_NO_RESEARCH" "AI_TAG_AVERAGE_SUPPLY"]
+
+    foci = [
+        [[HAS_INDUSTRY_FOCUS]]
+        [[HAS_RESEARCH_FOCUS]]
+        [[HAS_GROWTH_FOCUS]]
+        [[HAS_MINING_FOCUS]]
```
MatGB, line 12: Mining is gone, I was wandering why I hadn't found any until I checked my logs.

default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
```diff
@@ -0,0 +1,81 @@
...
+        [[AVERAGE_DEFENSE_TROOPS]]
+
+        // not for description
+
+        [[AVERAGE_PLANETARY_SHIELDS]]
+        [[AVERAGE_PLANETARY_DEFENSE]]
+
+        // huge bonus to planetary shields
+        EffectsGroup
+            description = "ANCIENT_PLANETARY_SHIELD"
```
MatGB, line 32: as of ec2f1cc this needs a _DESC suffix, @adrianbroher is on a bug hunt

default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
```diff
@@ -0,0 +1,81 @@
...
+            scope = And [
+                Source
+                Planet
+            ]
+            effects = SetMaxShield value = Value + 500
+
+        // huge bonus to ground troops
+        EffectsGroup
+            description = "ANCIENT_DEFENSE_TROOPS"
```
MatGB, line 41: same for this

default/stringtables/en.txt
```diff
@@ -11374,6 +11391,9 @@ GREAT_DEFENSE_TROOPS
 ULTIMATE_DEFENSE_TROOPS
 +++	Ultimate Defensive Ground [[encyclopedia TROOP_TITLE]]: 300%
 
+ANCIENT_DEFENSE_TROOPS
++++	Ancient Defensive Ground [[encyclopedia TROOP_TITLE]]: 10 per population
+
```
MatGB, line 11396: this and the below shield are where the merge conflicts were for me, I made a hash of sorting it out so better to be redone, both need the _desc suffix as above.

MatGB (changes requested):
I like the idea of this and want to expand on it, but better to get this in then use it for more down the line, fairly sure it needs AI updates as well.

default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
```diff
@@ -0,0 +1,81 @@
...
+            effects = SetMaxTroops value = Value + Target.Population * 10
+
+        // self-destruct when captured
+        EffectsGroup
+            scope = Source
+            activation = OwnedBy affiliation = AnyEmpire
+            effects = [
+                SetPopulation value = 0
+                GenerateSitRepMessage
+                    message = "EFFECT_ANCIENT_GUARDIANS_CAPTURED"
```
MatGB, line 55: Again, needs the label here.

MatGB:
Think they need an effect that sets their planetary population to max at start of game, if they're meant to be ancient they should have max pop not be growing on game start, especially given the troop effects.

Beyond that, finally got them working with current backend and I really like. Moving forward, the AI team need to review and possibly update the AI.

After that's been done, I'd like to expand the concept: for balance reasons I'd like the chance of guard on most of the big specials to always apply, so if a sentinel isn't generated these guys always should be, and possibly with a bonus of some sort for the bigger monsters, I've been uncomfortable for ages that ancient ruins can sometimes give you a really powerful unlock early with no defence whatsoever, etc.  But that's an ongoing balance thing, these are cool and I like them as is instead of sentries.

MatGB:
Also, Sitrep Priority needs to be added to the self destruct so it's just above the "dies out completely" message.

default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
```diff
@@ -0,0 +1,81 @@
+Species
+    name = "SP_ANCIENT_GUARDIANS"
+    description = "SP_ANCIENT_GUARDIANS_DESC"
+    gameplay_description = "SP_ANCIENT_GUARDIANS_GAMEPLAY_DESC"
+
+    tags = [ "ROBOTIC" "AI_TAG_NO_INDUSTRY" "AI_TAG_NO_RESEARCH" "AI_TAG_AVERAGE_SUPPLY"]
```
Morlic-fo, line 6: The AI_TAG_ prefix was removed in e39340464446e7bd3b268088ac4f0d03393a20e1. Just use "NO_INDUSTRY" etc.

default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
Morlic-fo: Does it make sense for a non-playable, non-conquerable species to have any foci?

default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
Morlic-fo: Also I'd like to have a new tag "DESTROYED_ON_CONQUEST" for AI purposes.

default/scripting/specials/specials.macros
```diff
@@ -29,7 +29,16 @@ CHANCE_OF_GUARD_1
                     And [ Planet HasSpecial name = "HIGH_TECH_NATIVES_SPECIAL" ]
                 ]
             ]
-            effects = CreateShip designname = OneOf("SM_GUARD_0", "SM_GUARD_1")
+            effects = [
+                If condition = Random probability = 0.6
+                    effects = [
+                        CreateShip designname = OneOf("SM_GUARD_0", "SM_GUARD_1")
+                    ]
```
Morlic-fo, line 36: Any reason to use brackets and 3 lines for a singe-line-statement?

default/stringtables/en.txt
```diff
@@ -6197,6 +6202,18 @@ SP_EXOBOT_DESC
...
+SP_ANCIENT_GUARDIANS
+Ancient Guardians
+SP_ANCIENT_GUARDIANS_GAMEPLAY_DESC
+'''Guardian Robots constructed by a Precursor civilization aeons ago
+[[encyclopedia ROBOTIC_SPECIES_TITLE]]
+Will self-destruct when conquered
+'''
+
+SP_ANCIENT_GUARDIANS_DESC
+'''Constructed by precusor technology these ancient machines can adapt to and fight in any kind of environment. With unwavering loyalty and commitment they have protected the planet from any invaders, never wasting time to repair and improve the fortifications. Upon being captured every Guardian Robot will automatically self-destruct.
```
Morlic-fo, line 6214: Not a native speaker - shouldn't there be a comma after `Constructed by precusor technology`, `With unwavering loyalty and commitment`, and `Upon being captured`?

Morlic-fo (changes requested):
This will need some AI adaptations but I wouldn't consider those to be critical enough to delay the merge.

@MatGB As before, please open another AI issue when you merge this.

Morlic-fo:
> Moving forward, the AI team need to review and possibly update the AI.
> 
> After that's been done, I'd like to expand the concept: for balance reasons I'd like the chance of guard on most of the big specials to always apply

While this PR may be merged without the AI adaptations, the concept should not be expanded until the relevant AI changes have been made. A rare special that the AI acts weird on is fine. Having half the galaxy being populated by Ancient Guardians is a different matter...

default/stringtables/en.txt
MatGB: I am, but am terrible with punctuation, I checked with the editor-I-live-with and she says yes in all three, the latter two being optional for reading-as-text but required if to be read aloud. The first definitely as it's two clauses.

I wouldn't have put any of them in either, so good catch.

MatGB:
> A rare special that the AI acts weird on is fine. Having half the galaxy being populated by Ancient Guardians is a different matter...

Fair enough: what I'm hoping for is that the existing chance-of-guard becomes chance of guard monster or chance of getting these, so that everything that gives a decent bonus gets something. But it's a longer term thing and is only a balance quibble not a major concern, it would also allow us to maintain having ancient ruins in No Monsters galaxies without me having conniptions over the balance issues there (ie the chance of a monster is determined by monster setting else you get these)

Deepsloth:
I made a new branch with a fixed and tweaked version of the ancient guardians, but there is one requested change i haven't figured out:

"Also, Sitrep Priority needs to be added to the self destruct so it's just above the "dies out completely" message."

How do i implement that?

dbenage-cx:
For now: https://github.com/freeorion/freeorion/blob/master/default/customizations/common_user_customizations.txt contains an ordered list of sitreps under `FUNCTIONAL_SITREP_PRIORITY_ORDER`.  These sitreps are displayed before those with no entry.

Issue #593, opened 2016-04-05
Hyperspatial Dam needs reworking


MatGB:
Currently it's an actively bad thing to build, so many other effects are keyed off population that the malus is debilitating.

Make the malus planet size dependent, and then redo the bonus so it gives a bonus overall if you haven't got a black hole generator but have all other techs and buildings, and a minor malus if you've got the BHG (which can't always be built in a game).

MatGB:
Update: #866 balanced the dam to make it not awful and usable, but further work is needed to differentiate it and make it interesting as discussed on the forums [here](http://www.freeorion.org/forum/viewtopic.php?f=2&t=10176)

PR #866, opened 2016-08-14
Balance Hyperspatial Dam

MatGB:
~ Population malus is set to equal planet size
~ Bonus is set to equal Black Hole Generator
~ Stacking Group added so does nothing if empire has BHG
~ Negative effects don't trigger if BHG in supply

### DIFF
--- a/default/scripting/buildings/BLACK_HOLE_POW_GEN.focs.txt
+++ b/default/scripting/buildings/BLACK_HOLE_POW_GEN.focs.txt
@@ -21,7 +21,7 @@ BuildingType
             ]
             activation = Star type = BlackHole
             stackinggroup = "BLD_BLACK_HOLE_POW_GEN_PRIMARY_EFFECT"
-            priority = [[VERY_LATE_PRIORITY]]
+            priority = [[LATE_PRIORITY]]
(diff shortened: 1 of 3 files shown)


MatGB:
OK, fairly simple, works as intended, addresses #593 as a quick fix.

@freeorion/ai-team to my knowledge the AI doesn't use the Dam (as, frankly, it was terrible), this is basically ready to go in but I created a PR so you can decide if you want to make any amendments to the AI either now or in future, if you're happy that it doesn't hurt them please do merge in ASAP.

At a basic level, virtually all players would benefit now from using the dam _if_ they haven't got a black hole system, a black hole generator is better—in the next Cycle we hope to rework both and a few other things to make them more interesting as being discussed in the [Poison advance](http://www.freeorion.org/forum/viewtopic.php?f=2&t=10176) thread.

Morlic-fo:
Fine with me. AI doesn't/won't build it but that seems not critical for release.

Vezzra:
#593 shall remain open however, right? Because the real fix is still coming...

MatGB:
Correct, but it's a lower priority, this just makes it a usable thing within current scripting style.

Issue #839, opened 2016-08-05
Meter growth tooltip doesn't display non-effect-based fuel regeneration


MatGB:
Prime example being Force Energy Structures, the correct amendments to the stats are applied each turn, but the hover text just says +1/-1 regardless, it would be better to not have that text than have it display wrongly for a lot of each game, but even better would be for it to work correctly.

This is a backend/UI bug not a scripting problem as far as I can see.

geoffthemedio:
I think the "always" is unnecessary in the title...

l29ah:
Also it thinks that outposts will grow some production due to the industrial automation.

geoffthemedio:
Original issue should be fixed now, but in doing so, a smaller similar / related issue with ship/fleet fuel regen display was created.

dbenage-cx:
Fixed? Regen is showing here

Vezzra:
Confirmed on OSX, normal regen showing up in the tooltip here too. Strange... why does this work all of a sudden?

Anyway, can anyone on Windows confirm this too?

MatGB:
Aye, I cannot remember seeing a fix for this go in but Force Energy Structures and fuel regeneration are both working properly as far as I can see.

Vezzra:
Well, seems to work on all platforms now, have just been able to confirm this myself.

Closing this as resolved.

PR #871, opened 2016-08-16
Fix bug839 meter growth tooltip incorrect

LGM-Doyle:
This PR fixes issue #839.

It does this by creating `DualMeter::CurrentProjectedTarget()` to calculate the correct projected meter estimates.  It then uses these estimates in the `MeterBrowseWnd` and `MultiMeterStatusBar`. 

I tested it with Force Energy Structures and the Megalith.

### DIFF
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -289,37 +289,39 @@ void MeterBrowseWnd::UpdateSummary() {
     const Meter* primary_meter = obj->GetMeter(m_primary_meter_type);
     const Meter* secondary_meter = obj->GetMeter(m_secondary_meter_type);
 
+    if (!primary_meter) {
+        ErrorLogger() << "MeterBrowseWnd::UpdateSummary can't get primary meter";
+        return;
+    }
+
     float breakdown_total = 0.0f;
     std::string breakdown_meter_name;
 
-    if (primary_meter && secondary_meter) {
+    if (secondary_meter) {
         if (!m_current_value || !m_next_turn_value || !m_change_value) {
             ErrorLogger() << "MeterBrowseWnd::UpdateSummary has primary and secondary meters, but is missing one or more controls to display them";
             return;
         }
 
-        // Primary meter holds value from turn to turn and changes slow each turn.
-        // The current value of the primary meter doesn't change with focus changes
-        // so its growth from turn to turn is important to show
-        float primary_current = obj->InitialMeterValue(m_primary_meter_type);
-        float primary_next = obj->NextTurnCurrentMeterValue(m_primary_meter_type);
-        float primary_change = primary_next - primary_current;
+        boost::tuple<float, float, float> current_projected_target = DualMeter::CurrentProjectedTarget(
+            *obj, m_primary_meter_type, m_secondary_meter_type);
 
-        m_current_value->SetText(DoubleToString(primary_current, 3, false));
-        m_next_turn_value->SetText(DoubleToString(primary_next, 3, false));
+        m_current_value->SetText(DoubleToString(boost::get<0>(current_projected_target), 3, false));
+        m_next_turn_value->SetText(DoubleToString(boost::get<1>(current_projected_target), 3, false));
+        float primary_change = boost::get<1>(current_projected_target) - boost::get<0>(current_projected_target);
         m_change_value->SetText(ColouredNumber(primary_change));
 
         // target or max meter total for breakdown summary
-        breakdown_total = obj->CurrentMeterValue(m_secondary_meter_type);
+        breakdown_total = boost::get<2>(current_projected_target);
         breakdown_meter_name = MeterToUserString(m_secondary_meter_type);
-
-    } else if (primary_meter) {
-        // unpaired meter total for breakdown summary
-        breakdown_total = obj->InitialMeterValue(m_primary_meter_type);
-        breakdown_meter_name = MeterToUserString(m_primary_meter_type);
     } else {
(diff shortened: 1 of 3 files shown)


UI/MeterBrowseWnd.cpp
```diff
@@ -184,6 +187,100 @@ void MeterBrowseWnd::UpdateImpl(std::size_t mode, const Wnd* target) {
...
+            if (boost::optional<const std::vector<Effect::AccountingInfo>&>
+                maybe_info_vec = GetAccountingInfo(obj.ID(), actual_meter_type))
+            {
+                const std::vector<Effect::AccountingInfo>& info_vec = *maybe_info_vec;
+
+                projected -= current;
+                for (std::vector<Effect::AccountingInfo>::const_iterator info_it = info_vec.begin();
+                     info_it != info_vec.end(); ++info_it)
+                {
+                    switch (info_it->cause_type) {
```
geoffthemedio: this would be a lot shorter as just

```
if (info_it->cause_type == ECT_UNKNOWN_CAUSE)
    continue;
```

or similar...

UI/MeterBrowseWnd.cpp
```diff
@@ -184,6 +187,100 @@ void MeterBrowseWnd::UpdateImpl(std::size_t mode, const Wnd* target) {
...
+                    case ECT_SPECIAL:
+                    case ECT_SPECIES:
+                    case ECT_SHIP_HULL:
+                    case ECT_SHIP_PART:
+                    case ECT_INHERENT:
+                        projected += info_it->meter_change;
+                        break;
+                    case ECT_UNKNOWN_CAUSE:
+                    default:
+                        ; // Don't add unknown effects.
```
geoffthemedio: why omit unknowns?

geoffthemedio:
I don't see inherent fuel regeneration on ship fuel tooltips now...
fuel

Vezzra:
Yep, confirmed. When I start a default quick game and venture with one of the starting scouts beyond supply range, I get exactly the same result as in Geoffs screenshot above.

However, I also tested your PR with my current test game, and there I've noticed that the fuel prediction works - with organic hulls:

fuelpredorgship

I also checked various other meter predictions, and everything else seems to work perfectly fine. So I'd really see that PR go in in time for the release - it fixes an age old annoyance. 👍 😃

UI/MeterBrowseWnd.cpp
LGM-Doyle: Fixed.

UI/MeterBrowseWnd.cpp
LGM-Doyle: I figured that the next turn prediction should not include information "not known" to your empire.

LGM-Doyle:
Fuel update prediction does not work currently in master.

The dual/paired/double meter handling is not enforced by anything other than convention in the code.  Sometimes the value visible to the UI are not consistent with the changes during the turn update.

Fuel is not handled the same as the other dual/paired/double meters.  Fuel is incremented and back propagated atomically in various places in Fleet.cpp and Ship.cpp.  Hence, the increments are not visible to the UI in the current and target meters for fuel.

This PR is an improvement, but not a complete fix to the problem.

Vezzra:
Personally I much prefer having all the other meter predictions now work correctly and only the standard fuel meter increase not working than the other way round, so I vote for merging this - unless there is something else left that prevents merging...?

Regarding this special handling of the default fuel increase: I propose to eleminate that completely from the backend C++ code and handle that entirely by FOCS. Then the meter prediction should work without having to add more special case handling to the C++ code in order handle default fuel increase correctly.

MatGB:
> I propose to eleminate that completely from the backend C++ code and handle that entirely by FOCS. 

Agree completely, that's definitely something with a lot of scope for scripting different approaches and putting it into scripts would make differentiating different hulls more interesting.

geoffthemedio:
If it could be handled by FOCS, it probably already would be. Problem is there's a bunch of specific special case rules for fleets flying into/out of/through systems that have supply connections and getting refuelled mid-movement, not just once per turn as would be the case with effects.

That said, the fuel increase for stationary ships probably could be scripted.

Vezzra:
> If it could be handled by FOCS, it probably already would be.

Well, the additional fuel regen of e.g. of organic ships is already handled by FOCS (which, btw, as mentioned in [my post above](https://github.com/freeorion/freeorion/pull/871#issuecomment-241220566), results in the fuel regen to be displayed correctly by the meter increase prediction), so that can't be the real problem here.

> Problem is there's a bunch of specific special case rules for fleets flying into/out of/through systems that have supply connections and getting refuelled mid-movement, not just once per turn as would be the case with effects.

Sure, but what has that to do with the default 0.1/turn fuel increase, which (AFAIK) happens completely independent of supply?

EDIT: Oh, I guess when I talked about eleminating default fuel increase completely from the C++ backend code, that sounds a bit misleading, as if referring to the entire refuel mechanic. Sorry, I only referred to this:

> That said, the fuel increase for stationary ships probably could be scripted.

That's exactly what I actually meant, and what I thought you've been referring to [in your post above](https://github.com/freeorion/freeorion/pull/871#issuecomment-241189667) as not being displayed correctly.

geoffthemedio:
mmhmm

If you/someone wants to move stationary ship +0.1 style fuel regen into scripts, then the C++ version can be disabled as here:

https://github.com/freeorion/freeorion/pull/903

MatGB:
Hah, yes, good. Post Release tho, leave it as is for now please (unless someone wants to write a macro into every existing ship hull and then test it extensively)

Issue #883, opened 2016-08-18
Supply propagation not working correctly


Vezzra:
### Environment
- **FreeOrion Version**: v0.4.5+ [build 2016-08-16.6a394b5]
- **Operating System**: Mac OS X 10.10.5
- **Graphic card used**: AMD Radeon HD 6490M 256 MB
- **Fetched as**
  - Compiled from source
### Description

![supplypropbug](https://cloud.githubusercontent.com/assets/10279417/17784210/f088f784-657b-11e6-9633-d7fed9417570.png)

In my current test game I encountered a strange situation concerning supply propagation. As you can see in the screenshot above, my system Tähti Zeta has no supply connection to the deep space node southeast of it. Which so far is correct, as supply propagates with a strength of 3 from Tähti Zeta into that deep space node, which is the same strength at which supply of the dark blue empire propagates from Trobia Alpha into it. Two empires propagating supply with the same strength into the same system means both can't get supply into it.

However, as you can see, the circle of the deep space node and the starlane to Trobia Beta are displayed in my color, as if I get supply into it, apparently from Grail Beta.

Now, if Grail Beta had strong enough supply to overhelm supply propagation of the dark blue empire, that would explain why it can propagate supply into the deep space node. _However_, first of all, that isn't the case - Grail Beta has a colony with supply 2, so it's actually _weaker_ than both Tähti Zeta and Trobia Alpha. Meaning, my empire shouldn't be able to propagate supply into the deep space node.

But even _if_ Grail Beta had a colony with sufficiently strong supply, then Tähti Zeta should be able to supply connect to the deep space node. Which it does not, as you can see by the coloring of the starlanes, Grail Beta and Tähti Zeta belong to different resource groups.

Save game:  [SupplyPropBug.sav.zip](https://github.com/freeorion/freeorion/files/425566/SupplyPropBug.sav.zip)
### Expected Result

The deep space node should not be supply connected to Trobia Beta. Its system circle and the starlane to Trobia Beta should be displayed in grey.

geoffthemedio:
Fixed?

Vezzra:
Yep, #888 fixed it. Thx!

PR #888, opened 2016-08-19
Fix obstructed system supply calculation.

LGM-Doyle:
This PR fixes issue #883.

Remove system from the set of unobstructed systems when two or more
empires have equal supply to the system.

Previously, only the starlanes were marked obstructed allowing another
system from a different starlane at a lower supply value to supply the
system.


geoffthemedio:
sounds plausible...

PR #891, opened 2016-08-20
Prevent unowned planet mines from affecting unowned guard ships

### LINKED ISSUE #873: Fortress mines kill defending Sentinel
The Fortress special generates a Guard to prevent early conquest at start of game, but the mines affect slowly kills said sentinel.

This isn't an instant fix as we now have all the mine effects using a macro, a short term hack could stop all mines from hurting sentinel, sentry, maintenance ship and warden designs. Alternatively the mines affect could be turned off unless the special is owned. I'm not sure on the ideal longer term solution.

dbenage-cx:
Fixes #873 

The following ship designs are not damaged from unowned planet mines:
-  SM_GUARD_0  -  Maintenance Ship
-  SM_GUARD_1  -  Sentry
-  SM_GUARD_2  -  Sentinel
-  SM_ACIREMA_GUARD  -  Acirema Guard Ship
-  SM_GUARD_3  -  Warden

### DIFF
--- a/default/scripting/ship_hulls/monster/SH_GUARD_0_BODY.focs.txt
+++ b/default/scripting/ship_hulls/monster/SH_GUARD_0_BODY.focs.txt
@@ -13,6 +13,7 @@ Hull
     buildCost = 1
     buildTime = 2
     Unproducible
+    tags = "UNOWNED_FRIENDLY"
     location = All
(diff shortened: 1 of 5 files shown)


MatGB:
Acirema, good spot, should've remembered them.

PR #896, opened 2016-08-20
Made supply not propagate into systems where another empire has a sup…

geoffthemedio:
…ply source and there are no ships to block the propagation.

https://github.com/freeorion/freeorion/issues/630
http://www.freeorion.org/forum/viewtopic.php?p=85127#p85127


Vezzra:
Works as intended on OSX.

PR #903, opened 2016-08-22
Disable in-C++ fuel regeneration while stationary.

geoffthemedio:
So the regeneration can be moved into effects / FOCS.

### DIFF
--- a/default/scripting/ship_hulls/Basic.focs.txt
+++ b/default/scripting/ship_hulls/Basic.focs.txt
@@ -15,6 +15,7 @@ Hull
         OwnedBy empire = Source.Owner
     ]
     effectsgroups = [
+        [[AVERAGE_BASE_FUEL_REGEN]]
         [[REGULAR_HULL_DETECTION]]
         [[SCAVANGE_FUEL_UNOWNED]]
         [[UNOWNED_GOOD_VISION]]
@@ -42,6 +43,7 @@ Hull
         OwnedBy empire = Source.Owner
     ]
     effectsgroups = [
+        [[AVERAGE_BASE_FUEL_REGEN]]
         [[REGULAR_HULL_DETECTION]]
         [[SCAVANGE_FUEL_UNOWNED]]
         [[UNOWNED_GOOD_VISION]]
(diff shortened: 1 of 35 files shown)


geoffthemedio:
Shall I / we merge this into master?

Vezzra:
Not unless there is also a patch which provides the necessary FOCS scripts replacing the current hardcoded fuel regeneration. Otherwise ships that run out of fuel are stranded indefinitely.

Of course that's not _that_ big of an issue, and maybe an incentive for the rest of us to provide said FOCS scripts. Personally I'd do these two together, but others may have different preferences.

adrianbroher:
I agree with Vezzra, this PR should only be merged when a complete replacement is available.

MatGB:
Thirded.

However, without checking IIRC Dilvish added a macro to all ships that currently does stuff when they're unowned, as a quick hack that could be adapted until someone has time to update all the FOCS files for every ship and decide which lines get what regen, etc.

This is going to be a very useful change when I/we go through and rebalance the hull lines again, probably after Fighters is done, it'll be nice to, for example, let Organic ships refuel rates reflect the star type of the system they're in, etc.

dbenage-cx:
Initial [FOCS additions](https://github.com/freeorion/freeorion/compare/master...dbenage-cx:903_FOCS_base_fuel_regen), currently based on master 81ab0c6.

I did not commit directly to this branch, as I thought it might be rude with an open PR.
I recall seeing PR to a branch was undesired, which is preferred? (maybe a good test for projects?)

Vezzra:
> I did not commit directly to this branch, as I thought it might be rude with an open PR.

Usually a branch in the main repo means that all committers can work on it. However, @geoffthemedio never uses branches in his own clone of the main repo for PRs, so branches he creates in the main repo for his PRs are a bit different - check back with him if it's ok to commit to this branch.

> I recall seeing PR to a branch was undesired

Hm, I'd say creating a PR to a branch is perfectly fine, and has already been done on rare occasions.

geoffthemedio:
Feel free to commit if it's in the main repo.

geoffthemedio:
Any objections to merging this now that the FOCS-based regen is in place?

Seems to work, although the next turn value predictions in the fleet panel / fleet window don't yet show it.

geoffthemedio:
Lacking any objections, I'll merge.

Vezzra:
> Seems to work, although the next turn value predictions in the fleet panel / fleet window don't yet show it.

Weird... now that this is done by FOCS like all the other similar stuff, why do the value predictions work for them but not for this? Need to keep an eye on it and open an issue I think...

geoffthemedio:
My assumption is that the relevant tooltip is different. Started looking into it, but got busy / distracted.

Issue #901, opened 2016-08-21
Situation report font size does not obey UI options


raptor:
- **FreeOrion Version**:  0.4.5+ [build 2016-08-21.68b91cc]
- **Operating System**:  openSUSE 13.2 x86_64
- **Graphic card used**:  Both Intel (using i915 module) and NVidia Quadro NVS 4200M (using nvidia binary blobs)
- **Fetched as**
  - Compiled from source
### Description

Adjusting the font size for 'text' in the UI options does not adjust the font size for the yearly situation reports.  This occurs even after restarting the game, and in different screen modes (full screen/ windowed and with different resolutions).

geoffthemedio:
There is a separate option for the sitrep font size, which is not reflected in the options screen. Why this was added, I'm not sure, but it's certainly intentional.

geoffthemedio:
https://github.com/freeorion/freeorion/commit/5f94fee8ef139abd2a11cbf0ce942a0241b3ee9c
http://www.freeorion.org/forum/viewtopic.php?p=79699#p79699
http://www.freeorion.org/forum/viewtopic.php?f=9&t=9589

Vezzra:
Hm, it's nevertheless far from ideal that the font size you can specify in the options dialog has no effect on the sitrep text, without being able to change the font size for it too via the options dialog. That you can do so by hacking config.xml isn't really helping.

@TheSilentOne1, 5f94fee8ef139abd2a11cbf0ce942a0241b3ee9c was your commit, what was the idea again? Why didn't you add sitrep font size to the options dialog?

TheSilentOne1:
No specific reason. If there's no objections, I'll do so.

adrianbroher:
From http://www.freeorion.org/forum/viewtopic.php?p=79699#p79699

Please keep the response order in mind: 3 1 2.

> and there is a forum thread with more information on the rationale.

There is no obvious link to the rationale. It's not clear if this means that there should be two font size settings (I don't believe this is intentional) or if the icon size should be configurable and not depend on the font size (sound much more reasonable to me).

TheSilentOne1:
Firstly, the option-db entry currently (also) has no effect on the sitrep font size. My bad - first contact with FreeOrion's code. However, I agree with @adrianbroher that we don't need an individual font size for the sitrep entries. So I second to remove the sitrep font size options-db entry.
If we agree on this I'll put up a fix.

TheSilentOne1:
[...]

PR #906, opened 2016-08-22
remove individual sitrep font size. use general font size instead

TheSilentOne1:
in reference to:
https://github.com/freeorion/freeorion/issues/901

### DIFF
--- a/UI/SitRepPanel.cpp
+++ b/UI/SitRepPanel.cpp
@@ -265,7 +265,7 @@ namespace {
             AttachChild(m_icon);
 
             m_link_text = new LinkText(GG::X0, GG::Y0, GG::X1, m_sitrep_entry.GetText() + " ", 
-                                       ClientUI::GetFont(ClientUI::Pts()),
+                                       ClientUI::GetFont(),
                                        GG::FORMAT_LEFT | GG::FORMAT_VCENTER | GG::FORMAT_WORDBREAK, ClientUI::TextColor());
             m_link_text->SetDecorator(VarText::EMPIRE_ID_TAG, new ColorEmpire());
             AttachChild(m_link_text);
--- a/UI/SitRepPanel.cpp
+++ b/UI/SitRepPanel.cpp
@@ -25,15 +25,11 @@ namespace {
         db.Add("verbose-sitrep", UserStringNop("OPTIONS_DB_VERBOSE_SITREP_DESC"),  false,  Validator<bool>());
         db.Add<std::string>("hidden-sitrep-templates", UserStringNop("OPTIONS_DB_HIDDEN_SITREP_TEMPLATES_DESC"), "");
         db.Add("UI.sitrep-icon-size", UserStringNop("OPTIONS_DB_UI_SITREP_ICONSIZE"), 24, RangedValidator<int>(12, 64));
-        db.Add("UI.sitrep-font-size", UserStringNop("OPTIONS_DB_UI_SITREP_FONTSIZE"), 12, RangedValidator<int>(10, 48));
     }
     bool temp_bool = RegisterOptions(&AddOptions);
 
     GG::X GetIconSize()
     { return GG::X(GetOptionsDB().Get<int>("UI.sitrep-icon-size")); }
-
-    GG::X GetFontSize()
-    { return GG::X(GetOptionsDB().Get<int>("UI.sitrep-font-size")); }
 
     std::map<std::string, std::string> label_display_map;
     std::map<int, std::set<std::string> > snoozed_sitreps;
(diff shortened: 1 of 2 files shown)


adrianbroher:
Just a note that is unrelated to the PR content: did you intent to use this author name? 
Because it isn't associated with your github account.  It doesn't really matter from a technical standpoint, but you either have used another author name for earlier commits and may want to configure your local git installation accordingly or or you want associate the new email with github.  It will still work if you choose to do nothing.

TheSilentOne1:
Thanks for pointing that out. Think I should've changed my local git client to the right account now.

UI/SitRepPanel.cpp
```diff
@@ -269,7 +265,7 @@ namespace {
             AttachChild(m_icon);
 
             m_link_text = new LinkText(GG::X0, GG::Y0, GG::X1, m_sitrep_entry.GetText() + " ", 
-                                       ClientUI::GetFont(std::max(1.0*ClientUI::Pts(), 1.0*GetOptionsDB().Get<int>("UI.sitrep-font-size"))),
+                                       ClientUI::GetFont(1.0*ClientUI::Pts()),
```
geoffthemedio: no need for 1.0\* or any parameter to GetFont... the no-parameter version defaults to the same thing.

UI/SitRepPanel.cpp
```diff
@@ -25,16 +25,12 @@ namespace {
...
         db.Add("UI.sitrep-icon-size", UserStringNop("OPTIONS_DB_UI_SITREP_ICONSIZE"), 24, RangedValidator<int>(12, 64));
-        db.Add("UI.sitrep-font-size", UserStringNop("OPTIONS_DB_UI_SITREP_FONTSIZE"), 12, RangedValidator<int>(10, 48));
     }
     bool temp_bool = RegisterOptions(&AddOptions);
 
     GG::X GetIconSize()
     { return GG::X(GetOptionsDB().Get<int>("UI.sitrep-icon-size")); }
 
-    GG::X GetFontSize()
-    { return GG::X(GetOptionsDB().Get<int>("UI.sitrep-font-size")); }
```
geoffthemedio: if removing the use of this option, remove the option addition as well? and any relevant stringtable entry...

UI/SitRepPanel.cpp
```diff
@@ -25,16 +25,12 @@ namespace {
         db.Add("verbose-sitrep", UserStringNop("OPTIONS_DB_VERBOSE_SITREP_DESC"),  false,  Validator<bool>());
         db.Add<std::string>("hidden-sitrep-templates", UserStringNop("OPTIONS_DB_HIDDEN_SITREP_TEMPLATES_DESC"), "");
         db.Add("UI.sitrep-icon-size", UserStringNop("OPTIONS_DB_UI_SITREP_ICONSIZE"), 24, RangedValidator<int>(12, 64));
-        db.Add("UI.sitrep-font-size", UserStringNop("OPTIONS_DB_UI_SITREP_FONTSIZE"), 12, RangedValidator<int>(10, 48));
```
geoffthemedio, line 28: is there a stringtable entry to remove as well?

TheSilentOne1:
Good catch with the stringtable entry. Have made the relevant changes.

Vezzra:
Looks ok to me.

UI/SitRepPanel.cpp
geoffthemedio: still unnecessarily calling multiple functions... just ClientUI::GetFont() should suffice

UI/SitRepPanel.cpp
```diff
@@ -269,7 +265,7 @@ namespace {
             AttachChild(m_icon);
 
             m_link_text = new LinkText(GG::X0, GG::Y0, GG::X1, m_sitrep_entry.GetText() + " ", 
-                                       ClientUI::GetFont(std::max(1.0*ClientUI::Pts(), 1.0*GetOptionsDB().Get<int>("UI.sitrep-font-size"))),
+                                       ClientUI::GetFont(ClientUI::Pts()),
```
geoffthemedio: not sure what changed, but it wasn't this being fixed

Vezzra: The `1.0*` directly before `ClientUI::Pts()` was still there and has been removed. But I guess what you meant is that `ClientUI::Pts()` can be removed as well, as `ClientUI::GetFont()` without any parameters will yield the exact same result anyway, correct?

geoffthemedio: yes

TheSilentOne1:
Removed that part as well, sorry.

PR #911, opened 2016-08-24
Moved new/temporary supply propagation special case into supply code …

geoffthemedio:
…instead of more-general-use empire obstructed system code.


geoffthemedio:
supply

With this change, the cyan AI might want to put a ship in Kentaurus in order to connect Minkowski to the rest of its supply network.

MatGB:
Appears to have solved both the fleet movement problem and the newly conquered system problem, but that's on a single turn test.

MatGB:
Several further turns and I think it's done the job.

PR #914, opened 2016-08-24
Add orbit dump entries for systems

dbenage-cx:
Adds orbit entries for system debug dumps.

/cc @Vezzra

### DIFF
--- a/universe/System.cpp
+++ b/universe/System.cpp
@@ -192,8 +192,19 @@ std::string System::Dump() const {
     os << UniverseObject::Dump();
     os << " star type: " << UserString(EnumToString(m_star))
        << "  last combat on turn: " << m_last_turn_battle_here
-       << "  starlanes: ";
+       << "  orbits: " << m_orbits.size() << " (";
 
+    int orbit_index = 0;
+    for (std::vector<int>::const_iterator it = m_orbits.begin();
+        it != m_orbits.end(); ++it)
+    {
+        int planet_id = *it;
+        os << "[" << orbit_index << "]" << planet_id << (it == m_orbits.end() ? "" : ", ");
+        ++orbit_index;
+    }
+    os << ")";
+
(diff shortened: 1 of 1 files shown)


universe/Planet.cpp
```diff
@@ -184,13 +184,15 @@ UniverseObjectType Planet::ObjectType() const
 { return OBJ_PLANET; }
 
 std::string Planet::Dump() const {
+    TemporaryPtr<System> cur_system = GetSystem(SystemID());
```
geoffthemedio: If you have to get another object to get some info, it's probably not something that should be in a Dump function. The idea is to output this object's state, not any related information from other objects.

dbenage-cx: Thank you, dropping planet dump and reworking system to report each orbit.

dbenage-cx:
Testing with #869

geoffthemedio:
looks ok... assuming you did some basic testing

dbenage-cx:
@geoffthemedio sorry, i meant I was still testing.  Minor issue with the trailing `,`.
(Note: with current master, systems will report no orbits)

geoffthemedio:
fix it and make another pull request then...

Vezzra:
@adrianbroher: Not related to this PR per se, just taking it as an example: what category to use for this? It's not a refactoring or bug, and to label this as feature doesn't really fit IMO. It's more something like an extension/enhancement of an already existing thing, in this case the `Dump` function of the `System` class.

Personally I'd prefer to use "enhancement" for this kind of things. That ok with you, or would you prefer something else?

Vezzra:
@dbenage-cx: Thanks! That should help with the debugging.

adrianbroher:
@Vezzra we should keep that discussion in the corresponding thread in the forum.

http://www.freeorion.org/forum/viewtopic.php?f=4&t=10199

PR #918, opened 2016-08-25
Remove 'Auto-generated effect descriptions' in favor of EffectBase::Dump

adrianbroher:
See http://www.freeorion.org/forum/viewtopic.php?f=9&t=10202

> If there are no objections to it I would like to remove the 'Auto-generated effect descriptions'. To my knowledge the output always was subpar and by adding the optional 'description' parameter to an EffectsGroup a year ago (8359b308) content writers now can provide much better description texts.

The attached PR implements this considering the remarks given by @geoffthemedio in the linked thread.

@geoffthemedio 

Could you check if the output is satisfying for your needs?


geoffthemedio:
autopedia

Some section headings in the pedia entries would be helpful, ie. Location and Effects

default/stringtables/en.txt
```diff
@@ -2594,8 +2594,8 @@ Music files
 OPTIONS_SOUND_FILE
 Sound files
 
-OPTIONS_AUTO_EFFECT_DESC
-Auto-generated effect descriptions
```
geoffthemedio, line 2603: Prefer the original name... meaning of "Dump" in the UI is unclear.

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -53,7 +53,7 @@ namespace {
     const int DESCRIPTION_PADDING(3);
 
     void    AddOptions(OptionsDB& db)
-    { db.Add("UI.autogenerated-effects-descriptions", UserStringNop("OPTIONS_DB_AUTO_EFFECT_DESC"),  false,  Validator<bool>()); }
+    { db.Add("UI.dump-effects-descriptions", UserStringNop("OPTIONS_DB_DUMP_EFFECTS_GROUPS_DESC"),  false,  Validator<bool>()); }
```
geoffthemedio: I think this string is missing, probably because I requested the original be used instead...

default/stringtables/de.txt
```diff
@@ -1901,8 +1901,8 @@ Musikdateien
 OPTIONS_SOUND_FILE
 Klangdateien
 
-OPTIONS_AUTO_EFFECT_DESC
-Automatisch generierte Effektbeschreibungen
+OPTIONS_DUMP_EFFECTS_GROUPS_DESC
+Effektgruppenbeschreibungen anzeigen
```
geoffthemedio, line 1910: something weird is going on with this pull request. this line is not consistent with the file I have here. 

I've tried repeatedly to check this out / rebase / whatever, but it's not working properly (it = git interface & me). Seems OK though, so merge if you want.

adrianbroher:
> I've tried repeatedly to check this out / rebase / whatever, but it's not working properly (it = git interface & me). Seems OK though, so merge if you want.

This probably come from the fact that I squashed the changes you requested into the commit.  You should be able to fix this by either deleting your local 'remove-autodesc' branch, fetching all remotes and checking out the remote branch, that corresponds to the remote 'remove-autodesc' branch. A screenshot from some git log history viewer with all branches visible would probably helpful (e.g. gitk --all).

geoffthemedio:
I tried deleting the local branch and re-checking it out / pulling (using the command line instructions shown just above this text entry box), but it didn't help.

Here's what I see now, though I don't presently see any references to this pull request's branch anywhere.

gitk-all

I can try again if you want, but it doesn't really matter for this pull request, I think...

adrianbroher:
> Here's what I see now, though I don't presently see any references to this pull request's branch anywhere.

Is it listed under _Files >> List references_? I Don't know how or if you fetched the PR at all, so It's hard to guess the remote name for the branch.

geoffthemedio:
I get the PR by pasting

```
git checkout -b adrianbroher-remove-autodesc master
git pull git://github.com/adrianbroher/freeorion.git remove-autodesc
```

into the command prompt while the master branch is currently checked out.

After doing that, I have references to to branch in gitk
gitk-all-references

After rebuilding, I get the same string error in the options screen:
string_error

adrianbroher:
I see…

Well there is nothing wrong the command or the view. Unless the issues is not caused by the uncommitted changes (the red dotted 'commit' in the viewer) I will check if I broke the translation unintentionally.

adrianbroher:
> After rebuilding, I get the same string error in the options screen

Seems like I missed renaming the translation key in the en, fr and ru file when rebasing the last time. The new commits contain the changes.

PR #920, opened 2016-08-25
Display message box when no acceptable GL context can be created

adrianbroher:
Some users may not have an approciate OpenGL implementation installed.
Instead of letting those users experience a broken application inform
them what happended and that the PC doesn't match the mininal
requirements.

The minimal requirements for the human client are now defined as:
- Having a graphical desktop
- Having an OpenGL implementation that supports at least OpenGL 2.0
- Having a graphic card that is able to hardware accelerate said context

### DIFF
--- a/GG/src/SDL/SDLGUI.cpp
+++ b/GG/src/SDL/SDLGUI.cpp
@@ -623,24 +623,40 @@ void SDLGUI::SDLInit()
     SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
     SDL_GL_SetAttribute(SDL_GL_STENCIL_SIZE, 2);
     SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
+    SDL_GL_SetAttribute(SDL_GL_ACCELERATED_VISUAL, 1);
 
     // Create the window with a temporary size.
     // Use the same code for the real size initialization that is used for resizing the window
     // to avoid duplicated effort.
     m_window = SDL_CreateWindow(AppName().c_str(), Value(m_initial_x), Value(m_initial_y),
-                                Value(m_app_width), Value(m_app_height), SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE);
+                                Value(m_app_width), Value(m_app_height), SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_HIDDEN);
+    if (m_window)
+        m_gl_context = SDL_GL_CreateContext(m_window);
+    GLenum glew_status = glewInit();
 
-    if (m_window == 0) {
-        std::cerr << "Video mode set failed: " << SDL_GetError() << std::endl;
+    if (!m_window || !m_gl_context || GLEW_OK != glew_status) {
+        std::string msg;
+        if (!m_window) {
+            msg = "Unable to create window.";
+            msg += "\n\nSDL reported:\n";
+            msg += SDL_GetError();
+        } else if (!m_gl_context) {
+            msg = "Unable to create accelerated OpenGL 2.0 context.";
+            msg += "\n\nSDL reported:\n";
+            msg += SDL_GetError();
+        } else {
+            msg = "Unable to load OpenGL entry points.";
+            msg += "\n\nGLEW reported:\n";
+            msg += reinterpret_cast<const char*>(glewGetErrorString(glew_status));
+        }
+
+        SDL_ShowSimpleMessageBox(
+            SDL_MESSAGEBOX_ERROR, "OpenGL initialization error", msg.c_str(), 0);
+        std::cerr << msg << std::endl;
         Exit(1);
     }
-    m_gl_context = SDL_GL_CreateContext(m_window);
 
-    GLenum error = glewInit();
-    if(error != GLEW_OK){
-        std::cerr << "Glew initialization failed: " << error << " = " << glewGetErrorString(error) << std::endl;
-        Exit(1);
-    }
+    SDL_ShowWindow(m_window);
 
     SDL_ShowCursor(false);
 
@@ -833,6 +849,11 @@ void SDLGUI::FinalCleanup()
     if (m_gl_context){
         SDL_GL_DeleteContext(m_gl_context);
         m_gl_context = 0;
+    }
+
+    if (m_window) {
+        SDL_DestroyWindow(m_window);
+        m_window = 0;
     }
 }
 


geoffthemedio:
Builds and seems to run OK on Windows, though I have GL > 2.0 so can't test the main functionality.

adrianbroher:
> Builds and seems to run OK on Windows, though I have GL > 2.0 so can't test the main functionality.

According to a [forum comment in the OpenGL](1) forum you should be able to force a 1.1
 OpenGL software only context by disabling the hardware acceleration on windows. Here is [a tutorial](2) to disable the hardware acceleration.

I tested it by requesting a high OpenGL version which can't be satisfied by the driver. Just set the `SDL_GL_CONTEXT_MAJOR_VERSION` in [GG/src/SDL/SDL_GUI.cpp](3) to 5 or 6.

geoffthemedio:
Set the required GL version to 6, results look as intended:
gl_init_message

It might be better to let the game try to run in this case, though, after showing the popup, in case someone wants to try anyway...

adrianbroher:
> It might be better to let the game try to run in this case, though, after showing the popup, in case someone wants to try anyway...

There is nothing to try. The prerequisites are not there and as soon as some code path with 1.1+ OpenGL code is executed the application **will** crash.

geoffthemedio:
There have been cases in the past where the full requirements for some level of GL aren't met, but parts of it are implemented, which let FreeOrion run, perhaps with rendering problems.

adrianbroher:
> which let FreeOrion run, perhaps with rendering problems.
> with rendering problems

So the cases are:
1. it didn't run properly. 
2. it didn't run at all.

I still don't see any reason not to bail out at initialization stage.

geoffthemedio:
Because sometimes 1) isn't enough of a problem that someone doesn't want to play anyway.

adrianbroher:
So the scanline renderer is unecessary for gameplay?

geoffthemedio:
"necessary", no. "rather helpful", yes.

but regardless, there are surely cases where it's irrelevant if a particular rendering issue is present and someone might still want to run the client.

adrianbroher:
Then why requesting OpenGL 2.0 in the first place?

Request1.5 and check if 2.0 is available when using 2.0 features.

If it is not make it a NoOp.

That's still magnitudes better than crashing.

geoffthemedio:
It's a lot more complicated to check if stuff is supported every time it's used than to just check at the start, show a warning if it's not, and then assume it is elsewhere and accept any resulting problems.

adrianbroher:
So in other words you don't care about crashing if it involves proper implementation. Unless it could crash somewhere else in the code then better ignore it with a

```
if(variable_that_must_be_initialized)
    return; // better do nothing and god forbid don't use exceptions.
```

geoffthemedio:
I would prefer it didn't crash ever, but if it only crashes after a warning to the player saying it might, then that's not so bad. If it crashes or doesn't render properly without any warning that that might occur, that's much worse.

I don't know why you closed this, as the warning popup is still good to have. What I'm suggesting now is to give the player the option to continue after they see the warning if they want.

Since putting a bunch of is-it-available tests before each use of GL stuff is a lot of unnecessary work, showing a warning instead is, I think, sufficient.

adrianbroher:
> I don't know why you closed this

Because I won't _fix_ it the way you suggested.  I don't write intentional broken code especially when seeing reports like #929 or [this thread](http://freeorion.org/forum/viewtopic.php?f=25&t=10203).

To be honest I don't feel like contributing anymore.  Most changes to the last version are minuscule IMO and removing/replacing some of the uglier code parts just doesn't happen for reasons I can't agree with.  May it be the discussion about 'python instead of FOCS' or 'use cmake instead of 3 diverging build systems to allow consistent building and automatic building' or 'use a translation system that has a proper toolchain so the translations would not bitrot' or 'better ignore internal domain model inconsistencies and run a broken game with subtle to find errors but don't scare the user with any error message' or 'gui components don't need MVC, better subclass every base control variant until it segfaults for no obvious reasons'.  On top of that I would love to see some general graphical update but with the current project climate I just won't see something like that happen soon or anytime.

geoffthemedio:
Actually that thread is a pretty good example / argument towards your approach in this case... I wasn't really considering the (actually common) case of crashes apparently causing weird issues where the game won't start again afterwards. That's a much bigger issue I would like to avoid, more so than just crashing or having rendering artifacts. It's also different to have a forced exit when first starting the game than it is to have errors crashes happen while the game is actually being played, especially in case of multiplayer, where I prefer to continue the game if possible. So, while I still doubt your claim that lacking GL 2.0 means it will definitely crash, I would merge the PR as it is now. An OSX test might be good first, though. 

The rest of this should probably go in a forum post, so I'll move it there if you'd prefer, though feel I need to respond here also...

I'm concerned to hear you're so unhappy with how things are going. Certainly there are disagreements, but you're have been making a lot of valuable contributions over the last year, both to game code itself and about git workflow and support stuff like the SDK update. I don't want to lose your interest and effort...

Has there been a discussion about MVC GUI stuff that rejected your suggestions? I just recall the StateButton changes where you explained to 
dbenage-cx how to do what I assume is what you're referring to ( https://github.com/freeorion/freeorion/pull/575 ). Were there other discussions about how this sort of thing should done / reworked in the existing code?

About the error messages, I don't think anyone wants to never show them or to ignore any problems that might lead to them... Rather, it just shouldn't force the game to end. I support adding sanity checks and error logging to make sure things are working as intended. Implementing automated testing of content would also be very useful, and has been suggested / started a few times, but is a huge project that needs someone really motivated to make it useful, I think...

I thought with all your recent stringtable key changes around the code that you were getting things near to a state where other translation tools could be used? I still don't have a good sense of how something like gettext would work in the context of FreeOrion, and I would have thought what we have now is what we'd have wanted to move to anyway if we had been using something like gettext due to the extra custom functionality we now support, and you've posted saying essentially the same. There were also recent discussions where you brought you removing GUI settings from the stringtable, and it was agreed that that should change... I could easily have missed some discussions about this as I've been away / busy a lot in the last couple months... Could you clarify what you're frustrated about with this?

I was also hoping in particular that you'd get back to the CMake SDK build system soon... that would be very useful to automate, but last I checked still had a few issues with Python headers.

The graphical code is a case where a big overhaul would probably be very welcome. I don't think anyone who really knows what they're doing has looked at it in quite a while... certainly I'm just figuring out how to do the GL stuff as I go, and would rather appreciate someone more experienced with it to fix it up and do it properly.

Regarding Python vs. FOCS, was there another discussion where you gave some explanation beyond "kill FOCS", as here?

http://www.freeorion.org/forum/viewtopic.php?p=84941#p84941

That thread didn't reject switching to Python, but did note that it's perhaps not as simple to choose as you implied, and switching to Python won't fix all the issues with FOCS. I'd still like some elaboration on what switching to Python actually means, at least to you...

adrianbroher:
@geoffthemedio I will respond to your message in the corresponding forum threads.

@Vezzra could you please test if the code compiles in it's current state and also check if the message box is displayed when changing the major version as outlined in https://github.com/freeorion/freeorion/pull/920#issuecomment-242961713?

Vezzra:
@adrianbroher, have been already on it. Compiles, runs, and (if I change the major version as suggested above) shows the message box as intended:

opengliniterror

After clicking OK the app terminates cleanly.

As far as I'm concerned, can be merged.

PR #922, opened 2016-08-26
Revise naming method for new planets

dbenage-cx:
This PR is to address planet naming in a manner suitable for release (supersedes #761).

Names planets based on their proximity to the center of the system, in relation to other planets in their group.
Groups are defined as a) asteroid type planets and b) all others.

A couple of notable issues:
-  The first asteroid planet created by a nebula will not have a roman numeral, even with other asteroids in the system. (minor issue for release)
-  Creation of planets outside of the current default order (lowest free orbit) may cause duplicate or un-ordered names for future planets. (not a conflict for anything targeted for 0.4.6)
-  `CardinalSuffix` is fixed to group between asteroids and non-asteroids.  Allowing dynamic sets of PlanetTypes to group by is deferred for post-release.

### DIFF
--- a/default/python/universe_generation/starsystems.py
+++ b/default/python/universe_generation/starsystems.py
@@ -52,21 +52,12 @@ def name_planets(system):
     unless it's an asteroid belt, in that case name is system
     name + 'asteroid belt' (localized).
     """
-    planet_number = 1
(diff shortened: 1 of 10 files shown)


Vezzra:
Thx, perfect! 😃

PR #923, opened 2016-08-27
make species attribute description uniform

TheSilentOne1:
species desc
PR changes en, de, fr, ru stringtable so numbers range from 50-300%, switch -25% to 75% and -50% to 50%. Also some minor tweaks to de stringtable. Fixed some spec. attrib. numbers (industry -50% to 75%, etc.) for de and ru stringtable.


TheSilentOne1:
Not sure if "refactoring" is the right category for this, change as appropriate pls.

geoffthemedio:
didn't check carefully, but seems reasonable

Issue #925, opened 2016-08-27
resource panel cut off at the right


TheSilentOne1:
### Description

![resource_panel](https://cloud.githubusercontent.com/assets/12985960/18027172/10aa5098-6c5d-11e6-98ad-7999728278ae.jpg)
Resource panel is cut off under the collapse arrow.
### Expected Result

Have panel extend to the right as the other panels do.
### Steps to reproduce

Start a new game, open sidepanel, click (un)collapse arrow of resource panel.

PR #926, opened 2016-08-27
Fix bug925 resource sidepanel cutoff

LGM-Doyle:
Fix for #925.

This PR does two things:
- It fixes the child clipping of the resource sidepanel panel.
- It allows the MultiMeterStatusBar to rescale to handle values larger than 100.0, by increasing the range of the MultiMeterStatusBar in increments of 100.0.  It still draws the grey lines every 20 units to show the scale.

@TheSilentOne1, I was not sure if you wanted this fixed or you wanted to fix it yourself.  I chose to fix it, since I added the offending line of code incorrectly clipping the resource Client.

### DIFF
--- a/UI/ResourcePanel.cpp
+++ b/UI/ResourcePanel.cpp
@@ -43,8 +43,6 @@ ResourcePanel::ResourcePanel(GG::X w, int object_id) :
     if (!res)
         throw std::invalid_argument("Attempted to construct a ResourcePanel with an UniverseObject that is not a ResourceCenter");
 
-    SetChildClippingMode(ClipToClientAndWindowSeparately);
-
     GG::Connect(m_expand_button->LeftClickedSignal, &ResourcePanel::ExpandCollapseButtonPressed, this);
 
     // small meter indicators - for use when panel is collapsed


TheSilentOne1:
I don't mind you fixing it. Thanks.

geoffthemedio:
Two separate changes like this should probably be in separate pull requests.

LGM-Doyle:
Both commits are part of the fix for #925.

The first commit fixes the problem for resource side panels without values larger than 100.0.
The second commit fixes it for panels with bars longer than 100.0 units.

Do you want each commit in a separate PR?

geoffthemedio:
I don't see what the number 100 has to do with that issue. It's about the right side of the bars being cut off by some rendering artifact that doesn't appear in the other panels.

LGM-Doyle:
When I replicated the issue, I used a save with over 100 industry resource points.  It displayed two separate problems both causing rendering artifacts on the right side of the bars.

The first, evident in the example provided by @TheSilentOne1 is that the clipping of the resource panel is different from the other panels.  It has been since 2009.  The first commit makes the clipping the same as the other panels.

The second problem is that any meters larger than 100, render outside of the edge of the box.  This problem is not evident in the example provided by @TheSilentOne1.  It affects all of the side panel windows.  It is fixed in the second commit. 

100 is the maximum renderable value in the `MultiMeterStatusBar` before it draws over the edge of the boundary.

I considered both commits to be related to the same problem, incorrect right side of window clipping.

geoffthemedio:
The second problem is unrelated to the referenced issue, and should be a separate pull request, particularly as it changes how meter values are displayed.

LGM-Doyle:
OK.  I'll sort that, tomorrow.

LGM-Doyle:
Split into two PRs.  The second PR is #930.

PR #928, opened 2016-08-28
Beginner bonus

geoffthemedio:
Gives bonus industry, research, and troops (as multiplier to species effects) to human players when Beginner mode AI aggression is selected.

### DIFF
--- a/default/scripting/species/common/industry.macros
+++ b/default/scripting/species/common/industry.macros
@@ -5,7 +5,6 @@ BASIC_INDUSTRY
 '''        EffectsGroup
             scope = Source
             activation = And [
-                Source
                 Planet
                 TargetIndustry low = 0
                 Focus type = "FOCUS_INDUSTRY"
@@ -13,6 +12,17 @@ BASIC_INDUSTRY
             accountinglabel = "FOCUS_INDUSTRY_LABEL"
             priority = [[EARLY_PRIORITY]]
             effects = SetTargetIndustry value = Value + Target.Population * [[INDUSTRY_PER_POP]]
+
+        EffectsGroup      // gives human bonuses when AI Aggression set to Beginner
+            scope = Source
+            activation = And [
+                Planet
+                Human
+                (GalaxyMaxAIAggression = 0)
+            ]
+            accountinglabel = "DIFFICULTY"
+            priority = [[VERY_LATE_PRIORITY]]
+            effects = SetTargetIndustry value = Value * 4
 
         EffectsGroup                    // removes residual production from a dead planet
             scope = Source
--- a/default/scripting/species/common/industry.macros
+++ b/default/scripting/species/common/industry.macros
@@ -17,12 +17,12 @@ BASIC_INDUSTRY
             scope = Source
             activation = And [
                 Planet
-                Human
+                Human     // human player, not human species
                 (GalaxyMaxAIAggression = 0)
             ]
             accountinglabel = "DIFFICULTY"
             priority = [[VERY_LATE_PRIORITY]]
-            effects = SetTargetIndustry value = Value * 4
+            effects = SetTargetIndustry value = Value * 2
 
         EffectsGroup                    // removes residual production from a dead planet
             scope = Source
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -133,6 +133,9 @@ Paused
 RESUME
 Resume
 
+DIFFICULTY
+Difficulty
+
 # Name for a newly created general purpose fleet.
 # %1% represents a unique number.
 NEW_FLEET_NAME
--- a/default/scripting/species/SP_ABADDONI.focs.txt
+++ b/default/scripting/species/SP_ABADDONI.focs.txt
@@ -30,6 +30,7 @@ Species
         // not for description
         [[AVERAGE_PLANETARY_SHIELDS]]
         [[AVERAGE_PLANETARY_DEFENSE]]
+        [[STANDARD_SHIP_SHIELDS]]
     ]
 
     [[INFERNO_STANDARD_EP]]
--- a/default/scripting/species/SP_ACIREMA.focs.txt
+++ b/default/scripting/species/SP_ACIREMA.focs.txt
@@ -54,13 +54,7 @@ Species
             activation = Turn high = 0
             effects = SetStarType type = Blue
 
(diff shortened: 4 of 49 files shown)


geoffthemedio:
Suggestions for changes to the size of the bonus or for other things to give the bonus to are welcome...

default/scripting/species/common/industry.macros
```diff
@@ -5,15 +5,25 @@ BASIC_INDUSTRY
...
             scope = Source
             activation = And [
-                Source
                 Planet
                 TargetIndustry low = 0
                 Focus type = "FOCUS_INDUSTRY"
             ]
             accountinglabel = "FOCUS_INDUSTRY_LABEL"
-            priority = [[EARLY_PRIORITY]]
+            priority = [[DEFAULT_PRIORITY]]
```
MatGB: No. Default priority is explicitly reserved for the species trait multipliers, due to the vagaries of timing anything that is set to happen at the same time can happen before or after, and getting the balance right on these was a pain. Please keep as Early.

default/scripting/species/common/industry.macros
```diff
@@ -5,15 +5,25 @@ BASIC_INDUSTRY
...
+            scope = Source
+            activation = And [
+                Planet
+                Human
+                (GalaxyMaxAIAggression = 0)
+            ]
+            accountinglabel = "DIFFICULTY"
+            priority = [[VERY_LATE_PRIORITY]]
+            effects = SetTargetIndustry value = Value * 4
+
```
MatGB, line 26: That...

Will be a very substantial bonus if it's set to go off last.

default/scripting/species/common/industry.macros
geoffthemedio: Then why does the equivalent effect for research use default priority?

https://github.com/freeorion/freeorion/blob/master/default/scripting/species/common/research.macros#L10

MatGB: Looks like I messed up somewhere then and got, well, lucky in testing that it never processed in teh wrong order, they both need to apply before the multipliers (nothing ought to be prioritised at the same time as a multiplier)

I'll fix it tomorrow.

Vezzra:
Is the inclusion of the variable name refactoring in `UniverseGenerator` in this PR intentional?

geoffthemedio:
semi-intentional, can be removed if objectionable

Vezzra:
Well, it hasn't really anything to do with this PR, so it should probably be committed separately.

But you are the programming lead, so it's your call after all how strictly you want these things to be handled. I guess @adrianbroher would prefer a strict handling... 😉

geoffthemedio:
This pull request likely requires substantial adjustments anyway. Value *4 is a rather big bonus, though whether it is too big (or not big enough?) requires some consideration. Also, the few meters I adjusted are somewhat arbitrary. It also would be better if this was all in a tech instead of species effects (along with all the similar effects actually) due to the previously-discussed effects processing benefits from removing effectsgroups from species, which get check for all ships with that species as well.

default/stringtables/en.txt
```diff
@@ -124,6 +124,9 @@ Back
 LAST
 Last
 
+DIFFICULTY
+Difficulty
+
```
dbenage-cx, line 138: Can this label be renamed\titled to avoid future confusion? (`Newbie Bonus`?)
I was under the assumption AI Aggression is not intended as a difficulty setting (although used that way sometimes due to lack of options).

MatGB: I'd avoid newbie, although I've never understood why I've seen it interpreted as an insult too many times, I'm guessing there are groups /cultures out there that use it as such that I've managed to avoid. New Player would be OK but Beginner is clearer. 

We had a suggestion to have a catch all difficulty recommended settings option to set various things to "easier" options, I'm not sure it's needed but I don't think it matters. All the AI types except Beginner can be a challenge but the meaning of that specific setting is clear. 

Not 100% sure this is the approach to take but something's needed and we can improve it or replace it in future.

default/scripting/species/common/industry.macros
Vezzra: @MatGB, did you get around to fix that issue as you intended? I don't recall seeing any commits regarding this...

MatGB: That'd be because I forgot, sorry, friend visiting. [f259042](https://github.com/freeorion/freeorion/commit/f2590428b3a716ec40e547953972b195f2d22735)

Probably ought to have some comments somewhere explaining priorities better, ideally the default multipliers wouldn't be at default priority as any new effects may clash, ah well, hindsite, something to fix in future.

default/stringtables/en.txt
geoffthemedio: I didn't want to limit the label to beginner bonus, as higher difficulties might give players penalties instead of bonuses.

MatGB:
I've just mentioned this idea on the forum [FreeOrion • View topic - Need a harder difficulty](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10289&p=86771#p86771) as a possible solution to a different problem as well, it might be a plan to have this sort of thing and have it apply to either players (for an easy game) or to AIs (for a much harder game), I think it would be better to give the AIs a bonus rather than neuter a player as they do struggle to keep up against an experienced player in the early to midgame, but having it instead neuter the player and thus give the same relative bonus would also work.

geoffthemedio:
rebased

dbenage-cx:
Some minimum troops for the case of early colonization might be good.
I frequently forget I have no defensive troops in the early game, the AI is good at reminding me.
Also favor +1 shields on ships to help with combat, without over reaching with a weapon bonus.

I agree with eventually moving to a handicap setting (or group of settings), where these values would be transferred to a `+1` handicap.  (Content defined setting, so not objecting to current approach)

geoffthemedio:
Something like
````
    SetMaxTroops value = Value * 2 + 6
````
?

dbenage-cx:
That or even `value = max(6, Value * 2)`
Though if the existing effect applies to ships as well, then some separate effectsgroup `= max(6, Value)`
(Edit: or `= Value + 6`, as you see fit)

geoffthemedio:
Any further comments / objections / suggestions?

Vezzra:
Not from my side, as far as I'm concerned, go ahead and merge.

PR #930, opened 2016-08-29
Make MultiMeterStatusBar rescale to show largest value.

LGM-Doyle:
MultiMeterStatusBar will draw a bar off the edge of the sidepanel if any
of its displayed values exceed 100.0.

This commit rescales the display in increments of 100.0 to correctly
display the largest value in the MultiMeterStatusBar.

It fixes a bug uncovered while working on issue #925, and was split from PR #926.

### DIFF
--- a/UI/MultiMeterStatusBar.cpp
+++ b/UI/MultiMeterStatusBar.cpp
@@ -69,7 +69,7 @@ namespace {
     const int       BAR_PAD(1);
     const GG::Y     BAR_HEIGHT(10);
 
-    const double    MULTI_METER_STATUS_BAR_DISPLAYED_METER_RANGE = 100.0;
+    const double    MULTI_METER_STATUS_BAR_DISPLAYED_METER_RANGE_INCREMENT = 100.0;
 }
 
 MultiMeterStatusBar::MultiMeterStatusBar(GG::X w, int object_id, const std::vector<std::pair<MeterType, MeterType> >& meter_types) :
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I understand the motivation, but am also hesitant to chance the scaling. Part of the idea is that the visible meters have the same size for the same value. It was probably also originally the idea to keep meter values under 100.

Does anyone else have an opinion on whether the meter bar scale should change to accommodate very large values? Or perhaps the solution is to prevent such large values from occurring on a planet? I suppose that's probably impractical given the different properties of various meters, including shields or troops with supply.

If the meter bar is to rescale, then should it default to max 100, or perhaps something lower by default?

Also, is any multiple of 100 good, or perhaps it should be "round" numbers, like 10, 20, 50, 100, 200, 500, etc.

So, probably there should be a forum thread first.

MatGB:
We currently have so many planet meters that go way past 100 that that horse is way past bolted. I regularly get production over 300 on my better planets and while we have plans to massively reduce those totals it's not going to be soon, and we'll still have shields and similar that are scripted to go that high and have been since way before my time.

Only other potential solution is if '100' is set as the max target value either in that system or in the empire and it displays as a percentage but I don't think that's ideal.

geoffthemedio:
So @MatGB would you prefer to merge this?

Vezzra:
Regarding merging this PR as it is, I'd vote for yes. IMO it's definitely an improvement to the current situation (as @MatGB pointed out, it's not exactly the rare exception that a meter grows way past 100), and the simple approach to just use multiples of 100 should work well enough for the release (considering current numbers).

That doesn't stop us from trying to come up with better/more thorough/finetuned solutions afterwards.

Discussion on that however should, as suggested, continue on the forum, @LGM-Doyle has already opened a thread:
http://www.freeorion.org/forum/viewtopic.php?f=6&t=10210

Vezzra:
Did a brief test on OSX, builds and runs fine, works like a charm.

So, anything preventing this from being merged? Keeping 100 and multiples of 100 should be good enough for the release, and we can always improve the scale increase later on.

PR #935, opened 2016-08-30
Move various weapon stat-based effects into techs

geoffthemedio:
Should have similar effect processing optimization results as moving species effects.

Updated to work for various weapons and upgrades, and handle case of unowned ships.

The idea is that instead of processing this effect as a separate single-target effectsgroup for each ship with the part, it can be processed once per empire.

### DIFF
--- a/default/scripting/techs/ships/Weapons.focs.txt
+++ b/default/scripting/techs/ships/Weapons.focs.txt
@@ -11,6 +11,19 @@ Tech
         Item type = ShipPart name = "GT_TROOP_POD"
         Item type = ShipPart name = "SR_WEAPON_0_1"
     ]
+    effectsgroups =
+        EffectsGroup
+            scope = And [
+                Ship
+                OwnedBy empire = Source.Owner
+                DesignHasPart  name = "SR_WEAPON_1_1"
+            ]
+            accountinglabel = "SR_WEAPON_1_1"
+            effects = [
+                SetMaxCapacity partname = "SR_WEAPON_1_1" value = Value + 5
+                SetMaxSecondaryStat partname = "SR_WEAPON_1_1" value = Value + 1
+            ]
+
     graphic = "icons/tech/planetary_colonialism.png"
 
 Tech
--- a/default/scripting/techs/ships/Weapons.focs.txt
+++ b/default/scripting/techs/ships/Weapons.focs.txt
@@ -12,17 +12,7 @@ Tech
         Item type = ShipPart name = "SR_WEAPON_0_1"
     ]
     effectsgroups =
-        EffectsGroup
-            scope = And [
-                Ship
-                OwnedBy empire = Source.Owner
-                DesignHasPart  name = "SR_WEAPON_1_1"
-            ]
-            accountinglabel = "SR_WEAPON_1_1"
-            effects = [
-                SetMaxCapacity partname = "SR_WEAPON_1_1" value = Value + 5
-                SetMaxSecondaryStat partname = "SR_WEAPON_1_1" value = Value + 1
-            ]
+        [[WEAPON_BASE_EFFECTS(SR_WEAPON_1_1)]]
 
     graphic = "icons/tech/planetary_colonialism.png"
 
@@ -47,14 +37,7 @@ Tech
     tags = [ "PEDIA_SR_WEAPON_TECHS" ]
     prerequisites = "SHP_ROOT_AGGRESSION"
     effectsgroups =
-        EffectsGroup
-            scope = And [
-                Ship
-                OwnedBy empire = Source.Owner
-                DesignHasPart  name = "SR_WEAPON_1_1"
-            ]
-            accountinglabel = "SR_WEAPON_1_2"
-            effects = SetMaxCapacity partname = "SR_WEAPON_1_1" value = Value + 1
+        [[WEAPON_UPGRADE_CAPACITY_EFFECTS(SR_WEAPON_1_1, 1)]]
     graphic = "icons/ship_parts/mass-driver-2.png"
 
 Tech
@@ -67,14 +50,7 @@ Tech
     tags = [ "PEDIA_SR_WEAPON_TECHS" ]
     prerequisites = "SHP_WEAPON_1_2"
     effectsgroups =
-        EffectsGroup
-            scope = And [
-                Ship
-                OwnedBy empire = Source.Owner
-                DesignHasPart name = "SR_WEAPON_1_1"
-            ]
-            accountinglabel = "SR_WEAPON_1_3_DESC"
-            effects = SetMaxCapacity partname = "SR_WEAPON_1_1" value = Value + 1
+        [[WEAPON_UPGRADE_CAPACITY_EFFECTS(SR_WEAPON_1_1, 1)]]
     graphic = "icons/ship_parts/mass-driver-3.png"
 
 Tech
@@ -87,14 +63,7 @@ Tech
     tags = [ "PEDIA_SR_WEAPON_TECHS" ]
     prerequisites = "SHP_WEAPON_1_3"
     effectsgroups =
-        EffectsGroup
-            scope = And [
-                Ship
-                OwnedBy empire = Source.Owner
-                DesignHasPart  name = "SR_WEAPON_1_1"
-            ]
-            accountinglabel = "SR_WEAPON_1_4"
-            effects = SetMaxCapacity partname = "SR_WEAPON_1_1" value = Value + 1
+        [[WEAPON_UPGRADE_CAPACITY_EFFECTS(SR_WEAPON_1_1, 1)]]
     graphic = "icons/ship_parts/mass-driver-4.png"
 
 Tech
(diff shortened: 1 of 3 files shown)


default/scripting/techs/ships/Weapons.focs.txt
```diff
@@ -9,6 +9,19 @@ Tech
...
+    effectsgroups =
+        EffectsGroup
+            scope = And [
+                Ship
+                OwnedBy empire = Source.Owner
+                DesignHasPart  name = "SR_WEAPON_1_1"
+            ]
+            accountinglabel = "SR_WEAPON_1_1"
+            effects = [
+                SetMaxCapacity partname = "SR_WEAPON_1_1" value = Value + 5
```
geoffthemedio: should use a lookup for the relevant part's stats, rather than re-hard-coding here.

Vezzra:
I think "component:internal" is supposed to be used for additions/changes to the innards of the backend code which do not fit into the other "component" categories... so "component:scripting" should be completely sufficient here?

geoffthemedio:
just edit the labels; no need to post about it first.

Vezzra:
Ok... it's just that I want to check back before I remove a label someone else set...

geoffthemedio:
Updated. Testing requested.

LGM-Doyle:
Works for me.

Mass driver and laser damage scaled correctly when researching new levels.

geoffthemedio:
Do all monster parts still work? In particular, any non-player owned ships with SR_WEAPON_?_1 on them?

Would also be interesting if in a later-stage game with lots of player-owned ships, if there's any notable change in turn processing / meter updating time.

MatGB:
AFAIK, it shouldn't affect monsters at all, we specifically rewrote all the monster parts and hulls, including the guards, so they gave any bonuses directly and/or used unique monster stuff.

Going to test this now, curious why put it into Techs and not the parts themselves? A test for whether the owner has a tech is easy to script (and I'm testing a script using it now) but I don't know about the resource drain.

MatGB:
OK, playing in the Fighters branch rebased, which AFAICS shouldn't affect things at all, and all my ranged weapon stats are reduced to zero, have hit end turn a couple times to see if it/s a load/update issue and it's not, even Flak Cannons are gone

geoffthemedio:
The idea I'm basing this on is that if a weapon part has an effect, which only acts on the ship that has the part, then each ship has a separate effectsgroup to evaluate and execute, meaning for 200 ships with 5 parts, that's 1000 effectsgroups' activation conditions and scopes to evaluate (several of which end up being unnecessary due to stacking rules). If instead the effect is in a tech, then there's just one scope to evaluate. Even though single-ship Source scopes are trivial to evaluate, doing so ~1000 times involves a lot of creation and passing around of containers of pointers, as well as creating threads to execute the effects groups, probably some extra accounting stuff, etc. I suspect that doing a single effects group with 1000 targets once is a lot more efficient.

geoffthemedio:
@MatGB Did you rebuild from source first? There are important changes in ValueRef.cpp

LGM-Doyle:
For monsters I fought Kraken and Snowflakes, which don't have SR_WEAPON_ in their descriptions.

I fought one Dyson forest and a Maintenance ship, which I think is SM_TREE and SM_GUARD_0, both with SR_WEAPON_1_1.  I had basic shields and nothing unusual happened.  This does not test the weapon tech upgrades.

I only played 100 turns, so I didn't noticed any change in lag.

So just a test of basic functionality.

MatGB:
Ah, I missed that there were backend changes, no, didn't recompile and am in the middle of work in Fighters so can't do that immediately, will do so when I'm done what I'm doing.

geoffthemedio:
Bump!

Particularly if someone likes profiling, it would be nice to know if it makes a difference in games with lots of ships with lots of weapon parts.

LGM-Doyle:
How did you profile it while you were developing it?

I usually find a problem and then write tests to measure the performance, so that I can see if things are improving or have improved. 

I have not noticed this scripting issue as a problem in my playing or profiling so I have not drilled down into it.  Where were you seeing an improvement and how big was it?

geoffthemedio:
I didn't profile explicitly for this, but I recall there were previous cases where a similar change (the "moving species effects" mentioned in the first post) was made with substantial improvements, either noted qualitatively or through someone profiling / timing effect execution. I don't routinely play games long enough to have a relevant test case for this, though.

LGM-Doyle:
I assume that this PR, change the save state so you can't take a late game save, profile it, compile the new code and re-profile it.  If that is the case then comparison is difficult.

The easy way would be to setup a test bench with just this code under test.  Then you would know that it is input/output correct, performs the same function, and have performance information to determine which code was better.  It is a targeted test.  You can keep the code and add it to a test suite.

Another less satisfactory way is to run both sets of code in parallel through an entire game.  Again you can compare both input/output and performance.  This is how I tested/refactored the `MapWnd` supply lane coloring code.  The test code is useless and will have to be tossed. 

Right now option B is our only option.

Which effect processing loop do you expect to see a performance in?
How hard would it be to double up this code, so that the effect is processed twice, once by the old code and once by the new code?

geoffthemedio:
I don't think this would break a save.

LGM-Doyle:
Good.  

So the next question is where do you think is a good place to measure the performance change?  

I can merge this PR with my latest long play and add some test code at that place.  The improvement may be large, but if it is buried in the AI turn processing time, then I may miss it if I'm not looking in the correct location.

geoffthemedio:
A game with lots of ships that have lots of weapon 1 to 4 parts in their designs should see the most benefit for effect updates... Basically Universe::UpdateMeterEstimatesImpl. Hopefully turn updates should go faster in particular, and perhaps focus changes or production queue manipulations.

LGM-Doyle:
I'll try profiling this and report back.

LGM-Doyle:
Here are the times from profiling.

The most recent long game that I have saves for is based on the Fighters branch with the AI behavior and some of my sidepanel branches merged in.  I added this PR for WeaponStatEffects (WSE).  The profiling code was already there.  

I played 3 turns, which include 2 battles, 1 invasion and some pathing.  The bigger battle was  127 ships vs 18 ships, with 1.6K power of 69 Laser4 weapons with Solar Web vs 564 power of mixed Plasma and Laser4, with 162 fighter vs 107 fighters.

I only ran each case once and copied the results here.  

`InitTurn` times in ms, with totals at the bottom.

Without WSE | With WSE
-----------------|---------------
4330 | 4452
4256 | 4152
4219 | 4021
4128 | 4360
---|---
**16933** | 16985

`Universe::UpdateMeterEstimatesImpl` times in ms.  

I don't know for sure if each of these cases actually aligns.  I moved the 296 ms number down two rows, because it obviously aligns with the 426 row since they are the only estimates with 94 targets.

Without WSE | With WSE | Number of targets
-----------------|---------------|------------
1109 | 1195 | 0
1228 | 1288 | 0
1095 | 1166 | 0
1529 | 1283  | 0
1047 | 1125 | 0
1376 | 1215 | 0
296 | 426  | 94
1089 |1212 | 0
1157 |1501 | 0
---|---|---
**9926** | 10411 |

geoffthemedio:
Not much difference at that number of ships. I was thinking of somewhat larger fleets, though.

LGM-Doyle:
That was the game I had sitting around.

However, the timers for `InitTurn` and `UpdateMeterEstimatesImpl` are currently in master.  Anyone can merge this PR in a check the performance difference.

Vezzra:
What's the status on this PR? Judging by the comments, it could be merged, OTOH it doesn't offer much improvement apparently. Any ideas what to do with it?

LGM-Doyle:
It may/may not improve performance, but it does standardize the weapons effect in FOCS and that improves the code quality.

geoffthemedio:
Would also be interesting to test on a much bigger game with lots more ships and weapon effects happening, such as one where someone reports getting "5 minute" turns ( http://freeorion.org/forum/viewtopic.php?f=25&t=10414 )

Vezzra:
> It may/may not improve performance, but it does standardize the weapons effect in FOCS and that improves the code quality.

I agree, and am in favor of merging.

geoffthemedio:
rebased

MatGB:
I can't follow the code well, but *owned* flak cannons aren't doing anything (showing strength 0 and not shooting), the unowned effect works and things like Sentries work as intended.

I think there's just a little thing missing but I can't figure out what.

Morlic-fo:
Try adding
```
[[WEAPON_BASE_EFFECTS(SR_WEAPON_0_1)]]
```
to the effectgroups of the SHP_ROOT_AGGRESSION tech (where the macro for SR_WEAPON_1_1 is as well).

MatGB:
While it absolutely makes sense, the game won't load and the logs say nothing, looks like it makes the tech invalid but the server crashes on me before the log gets written.

Morlic-fo:
Did you also add the relevant brackets required by syntax?

```
effectsgroups = [
    [[WEAPON_BASE_EFFECTS(SR_WEAPON_1_1)]]
    [[WEAPON_BASE_EFFECTS(SR_WEAPON_0_1)]]
]
```

MatGB:
D'oh! Committing fix now, thanks. (weirdly, the fix didn't start working until I built a new ship with the part, at which point all my existing ships got the upgrade)

Issue #811, opened 2016-07-30
Tech tree labels are slow


l29ah:
When i enable rendering of all technologies, fps drops to 5 on Intel(R) Core(TM) i7 CPU L 620 @ 2.00GHz, regardless the actual amount of visible labels. When i zoom out enough the labels vanish, fps grows to 30.

Vezzra:
Been noticing that for a long time too, but never got around to open an issue about it. Thx.

Not sure how complicated it is to fix this, so I didn't assign a milestone yet.

l29ah:
This reproduces better on debug builds btw: it becomes multiple times slower, like 0,7fps with full tech tree enabled.

Vezzra:
Fixed by #936

PR #936, opened 2016-08-30
Add TechTreeWnd::LayoutPanel::TechPanel::PreRender().

LGM-Doyle:
This fixes issue #811.

It has no other user visible changes except speeding up rendering of the tech tree.

@l29ah thanks for pointing out the difference when zooming.  It was key to finding the source of the problem.

TechTreeWnd::LayoutPanel::TechPanel::Render() was using SetText() and
SizeMove() on TextControls which are both expensive operations.  This
slowed frame rate to << 10fps.

Refactor PreRender() from Render() by removing label SetText() and
SizeMove() operations.  PreRender() should only be run when label text
or zoom changes.  PreRender() is only intended run once per render cycle
when something has changed.

Create RequirePreRender() which forces the next PreRender() to be run.

Use RequirePreRender() in TechTreeWnd::LayoutPanel::TechPanel::Update()
and TechTreeWnd::LayoutPanel::SetScale(), which change label text and
scale respectively.

### DIFF
--- a/UI/TechTreeWnd.cpp
+++ b/UI/TechTreeWnd.cpp
@@ -604,6 +604,11 @@ public:
     virtual         ~TechPanel();
 
     virtual bool    InWindow(const GG::Pt& pt) const;
+
+    /** Require that layout and format be updated before the next Render()*/
+    virtual void    RequirePreRender();
+    /** Update layout and format only if required.*/
+    virtual void    PreRender();
     virtual void    Render();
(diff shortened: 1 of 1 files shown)


Vezzra:
If this could go in in time for the release, that would be really nice... 😃

Vezzra:
Did a quick test run on OSX - much better, doubles my FPS on the research screen.

geoffthemedio:
code looks ok. merge if it works for you.

Issue #924, opened 2016-08-27
Xenophobic frenzy sometimes gives bonus to population


Voker57:
Example here: http://dump.bitcheese.net/files/livyvun/xenofun.mps

Empire NBL, planet Yoyo 3 I

Vezzra:
Confirmed on OSX.

@MatGB, IIRC you've been the last one to tinker with the whole xenophobic stuff, any idea what's going on?

MatGB:
It's something related to the Trith self sustaining malus, I've encountered it but not for awhile and had forgotten about it to be honest. Self sustaining and the related malus have had weird issues in edge cases for ages. 

To see it, play trith and look at poor or hostile worlds before you've got the growth techs. I don't think it's release stopping but I'll have another look at it.

Vezzra:
Ok, so it's an edge case. Well, that xenophobic stuff has had these quirks quite a while, and it has been worse. So definitely not release stopping.

If you can figure out the problem and fix it, good, but it's not a priority (for the release).

Morlic-fo:
Should be fixed by #944

PR #944, opened 2016-09-02
Xenophobic bugfixes

Morlic-fo:
The PR covers two fixes related to the xenophobic frenzy:
#### 1. Fixes issue #924

Bug was existing as population malus was proportional to current population. So negative populations received a "negative malus", i.e. a bonus.
Note that the formula is not changed but merely recasted to an actually readable form:

> Previous form of the formula: `Max(A,B)`
> with `B = Value - 3*size =: Value - b`,
> with `A = Value * (0.6 + (0.4 * (0.8 ^ n)))`
> 
> A is easily recasted to
> 
> ```
> A  =  Value *  [ (1-0.4) + 0.4 * 0.8^n] 
>    =  Value * (1 - 0.4 * (1 - 0.8^n))
>    =  Value - Value * 0.4*(1-0.8^n)
>    =: Value - a
> ```
> 
> It is then easy to see that `Max(A, B)` is equivalent to `Value - Min(a, b)`, i.e. what is used in the new code.
#### 2. Fixes Happiness malus:

Activation for the condition ignores planets with Concentration Camps. However, planets with concentration camps where still counted for the calculation of the happiness malus.

### DIFF
--- a/default/scripting/species/common/xenophobic.macros
+++ b/default/scripting/species/common/xenophobic.macros
@@ -1,9 +1,12 @@
 XENOPHOBIC_SELF
-'''     EffectsGroup
+'''
+        // Give the concentration camp tech at game start
+        EffectsGroup
             scope = Source
             activation = Turn high = 0
             effects = GiveEmpireTech name = "CON_CONC_CAMP"
 
+        // Give malus to industry, research and happiness for other species nearby
         EffectsGroup
             scope = Source
             activation = And [
@@ -40,6 +43,7 @@ XENOPHOBIC_SELF
                 )
             ]
 
+        // Give a pop malus for self-sustaining xenophobic species based on the number of nearby planets with other species
         EffectsGroup
             scope = Source
             activation = And [
@@ -74,7 +78,7 @@ Count condition = And [
 
 // single argument should be the name of the species capitalized
 XENOPHOBIC_OTHER
-'''     EffectsGroup
+'''     EffectsGroup 
             scope = And [
                 PopulationCenter
                 OwnedBy empire = Source.Owner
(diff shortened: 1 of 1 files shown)


default/scripting/species/common/xenophobic.macros
```diff
@@ -31,15 +31,16 @@ XENOPHOBIC_SELF
...
+                        PopulationCenter
+                        OwnedBy empire = Source.Owner
+                        Not OR [ 
+                            Species name = Source.Species
+                            Species name = "SP_EXOBOT"
+                        ]
+                        Not Population high = 0
+                        Not Contains Building name = "BLD_CONC_CAMP"
+                        WithinStarlaneJumps jumps = 5 condition = Source
                     ]
```
Morlic-fo, line 43: Note that I changed the order of conditions to be consistent with the activation condition above.

MatGB:
Right...

I was on entirely the wrong track looking for what was wrong with that code, thank you.

I do want to completely rework the way we do xenophobia at some point (so that it's always a per planet malus the way happiness already is), but having the existing maths actually work is a good thing. Thank you.

I won't be able to test this for the next 48hours or so but it ought to go into the release if possible, assuming it works feel free to merge in.

Morlic-fo:
> I won't be able to test this for the next 48hours or so but it ought to go into the release if possible, assuming it works feel free to merge in.

Worked in my tests:
-Playing Trith and colonizing a hostile planet with some nearby natives gives 0 bonus/malus 
-Homeworld (with nearby natives) has same target population as before the changes.

Vezzra:
Sounds reasonable. Took only a very brief look at the changes, seem to be ok too as far as I can tell.

Go ahead and merge. If you don't get around to it and no one raises any objections, I'll merge it before I create the release branch on Sunday (as you set the "Release v0.4.6" milestone).

Vezzra:
Perfect, thanks!

PR #947, opened 2016-09-03
Remove debug print

Cjkjvfnby:
Every time I run game on windows,  black window with annoying message appears.

I don't know if we change logging level for release. If yes, then just close this PR.  Or may be it should be changed to print message only for not valid names?

PS. Game icon is blank,  it is not always blank, and this happens for a long time, should I create an issue separate issue for that?

PPS. I don't sure about tags.  May be add theirs descriptions to `guidelines for contributing`

debug_at_start


adrianbroher:
Is the build a 'Release' one or a 'Debug' one?

geoffthemedio:
Uh, I thought this was an issue, not a pull request. Already fixed:

https://github.com/freeorion/freeorion/commit/313f92f9dec464a21a9137b7d3df8258d42319cd

PR #953, opened 2016-09-07
Fix gg handles drag drop enter event inconsistently

LGM-Doyle:
GG used `DragDropEnter` and `CheckDrops` events inter-changeably and
inconsistently.

`ChecksDrops` should be used to check if drops are acceptable.

`DragDropEnter` should be used to inform windows of drag drop entry.

In `HandleMouseDrag()` `DragDropEnter` was used to check if drops are
acceptable.

In `CheckedGetWindowUnder()` the drag drop here window was being updated
without first causing the `DragDropEnter` event.

This commit make all three locations where the drag drop here window is
updated first use `CheckDrops` to capture drop acceptability and then
inform the window with `DragDrop` Enter.

This PR will cause new `DragDropEnter` events that were not being generated before the patch.

It can be cherry-picked to version 0.4.6. with or without PR #954 as you see fit.


Vezzra:
Tested this with together with #954, builds and runs fine on OSX. Works as intended. Code looks ok too as far as I can tell.

Issue #952, opened 2016-09-07
Display glitch when dragging ship parts on the ship design screen


Vezzra:
### Environment
- **FreeOrion Version**: master
- **Operating System**: OSX
- **Fetched as**
  - Compiled from source
### Description

When dragging ship parts on the design screen, if you drag a ship part over an already occupied slot (but don't drop it!) the ship part in that slot disappears.

This is purely a display glitch, internally the ship part is still in that slot. You can verify that by trying to add a ship part by double clicking, it won't insert the part (provided that the now apparently empty slot is the only empty one of course), or by adding the design and reloading it from the finished designs list - the part that was in the slot that appeared empty is displayed again.
### Expected Result

Obviously a ship part should still be displayed even if you drag another on over it (unless you drop that other part, but in this case the new one should be displayed, not an empty slot be left behind).
### Steps to reproduce
- Start new default quickstart game.
- Go to ship design screen.
- Load one of the predefined ship designs.
- Grab a ship part from the ship part list and drag it over one of the occupied slots, but don't drop it.
- As soon as you move away from the slot while still dragging the part, the part in the slot disappears.

Vezzra:
Not sure if I should really consider this as release-blocking, but on the other hand this is the kind of bugs we're trying to find and address when doing release candidates. And this one can be rather confusing for players.

So I went ahead and tagged it for the release for now, but if you guys think that's a bit too strict, we can remove the milestone.

Vezzra:
Fixed by #954.

PR #954, opened 2016-09-07
Add dragging ship part over replacable placed part hides part in DesignWnd.

LGM-Doyle:
This makes dragging a ship part over a part, that it could replace, hide
the part while the drag drop is hovering.

It fixes issue #952 . 

Together with PR #953 it works as described above.

Without PR #953 it just makes the ship parts already placed on the ship stop disappearing.

It can be cherry-picked to version 0.4.6. with or without PR #953 as you see fit.


Vezzra:
As I said in #953: tested these two PR together with #953, everything ok, code looks fine (at least to me). So I suggest to merge both and cherry pick them into the release branch (as I consider both of them fixes).

If no one objects and no one else merges it, I'm going to do that on Friday before I produce RC1.

PR #959, opened 2016-09-08
Add prompt to uninstall previous version in NSIS Installer

dbenage-cx:
Provides a prompt on startup of windows installer to uninstall the previous version.

As pointed out by xlightwavex: http://freeorion.org/forum/viewtopic.php?f=15&t=10217#p85406
A previous install may leave behind conflicting content files.

### DIFF
--- a/Installer/FreeOrion_Install_Script.nsi.in
+++ b/Installer/FreeOrion_Install_Script.nsi.in
@@ -135,3 +135,21 @@ Section Uninstall
   DeleteRegKey HKLM "$${PRODUCT_DIR_REGKEY}"
   SetAutoClose true
 SectionEnd
+
+Function .onInit
+  ReadRegStr $$R0 HKLM "$${PRODUCT_UNINST_KEY}" "UninstallString"
+  StrCmp $$R0 "" fin_onInit
+
+  MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION \
+  "$${PRODUCT_NAME} is already installed. $$\n$$\nClick `OK` to run the \
+  uninstall dialog (recommended) $$\nor click `Cancel` to continue." \
+  IDOK uninst_onInit
+  Goto fin_onInit
+
+  uninst_onInit:
+  ClearErrors
+  Exec "$$R0"
+
+  fin_onInit:
+
+FunctionEnd


Vezzra:
Hm, as a "fix" that's a bit of an edge case - but as this is a thing of quite some importance especially for releases, I guess you mean that to go in 0.4.6 nonetheless, right?

geoffthemedio:
I think the problem in the linked post is that something crashed, and then the uninstall didn't complete because some files were unmodifyable because the related program was still running. Given version string checking in config.xml, having older versions installed should not cause any problems, as each should default to its own default content directory, as long as there are independent install directories. If someone's installing one version over the other, I assume there's already a warning popup about that when picking the install location...

dbenage-cx:
For release 0.4.6 preferred, the default installer will install on top of the existing files (not replacing some) if the user does not change the directory.

The dialog allows installation to continue after clicking cancel, so is only a minor burden to users that change the default install directory (win7: always defaulted to `C:\Program Files (x86)\FreeOrion`).

Changing `Overwrite` from `try` was a smaller change, but I felt the barrier was much higher from the multiple error dialogs that would display (and does not resolve `*.pyc`, if those are the issue, or any other conflicts)

Vezzra:
@dbenage-cx, tested that on my old XP machine, and although FO is already installed there, the installer didn't show any message boxes offering me to uninstall the existing installation.

dbenage-cx:
Is this with an NSIS script install? (0.4.5 Release/0.4.5+)
If so, could you search the registry for `FreeOrion`, (I assume this is 32-bit XP).  Looking for _likely_ the 2nd-3rd match, grouped with `UninstallString`.

Currently, using the file _resolved_ from `HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\Uninstall\FreeOrion` `"UninstallString"`.  For 64-bit os, the key actually lands in HKLM `Software\Wow6432Node\...`, but should resolve correctly with a 32-bit app.

I was trying to keep from relying on `$INSTDIR\Uninstall.exe` being the target, so as to not add to the registry confusion and maintain a path for any future script changes to work from.

Vezzra:
> Is this with an NSIS script install? (0.4.5 Release/0.4.5+)

Yep, this XP machine is the one where I do all my tests of the weekly test, release candidates and release builds for Windows. So the FO version installed there was the most recent test build I installed (and I always use the installer package created by NSIS).

> If so, could you search the registry for `FreeOrion`...

Well, it wasn't exactly the 2nd or 3rd, more the thousandst match (well, I _do_ install and uninstall FO _a lot_ on that machine... ;)), but yes, the key `UninstallString` is there, in `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\UnInstall\FreeOrion`. Value: `C:\Programme\FreeOrion\Uninstall.exe`.

So everything as it should be, right?

> (I assume this is 32-bit XP)

Yep, correct.

dbenage-cx:
Rebased.
Missed a set of expanding brackets, running the NSIS compiler should result in 0 warnings.

If the dialog does not show up immediately after execution, could you set `RequestExecutionLevel none` on line 11 in `FreeOrion_Install_Script.nsi`?  This sets ManifestSupportedOS to none as well (execution level was added to manifest in vista).

Vezzra:
> Missed a set of expanding brackets, running the NSIS compiler should result in 0 warnings.

Apparently that fixed the issue. Works perfectly fine now. 😃

> If the dialog does not show up immediately after execution, could you set RequestExecutionLevel none on line 11 in FreeOrion_Install_Script.nsi? This sets ManifestSupportedOS to none as well (execution level was added to manifest in vista).

That wasn't necessary, so we're good to merge.

PR #961, opened 2016-09-09
Don't assume CMAKE_INSTALL_* variables to be relative paths

hasufell:
So instead of prepending CMAKE_INSTALL_PREFIX, which may produce
an incorrect path, use CMAKE_INSTALL_FULL_*, which can be assumed
to be an absolute path.


Vezzra:
@adrianbroher, I guess you'll have to look into this.

hasufell:
Since I feel incoming nitpicking, here's the relevant information from GNUInstallDirs.cmake:

```
# Inclusion of this module defines the following variables:
#
# ``CMAKE_INSTALL_<dir>``
#
#   Destination for files of a given type.  This value may be passed to
#   the ``DESTINATION`` options of :command:`install` commands for the
#   corresponding file type.
#
# ``CMAKE_INSTALL_FULL_<dir>``
#
#   The absolute path generated from the corresponding ``CMAKE_INSTALL_<dir>``
#   value.  If the value is not already an absolute path, an absolute path
#   is constructed typically by prepending the value of the
#   :variable:`CMAKE_INSTALL_PREFIX` variable.  However, there are some
#   `special cases`_ as documented below.
```

Especially:

> The absolute path generated from the corresponding `CMAKE_INSTALL_<dir>`
> value.  If the value is not already an absolute path, an absolute path
> is constructed

Which means `CMAKE_INSTALL_<dir>` may be an absolute path. So prepending cmake prefix must not be done manually.

hasufell:
Please don't pull this. I'm closing this PR and don't want it merged.

PR #972, opened 2016-09-13
Prevent sitreps from unpopulated drydocks

### LINKED ISSUE #970: Drydocks on outposts trigger "Can't repair due to low happiness" sitreps
### Environment
- **FreeOrion Version**: 0.4.6 RC1
- **Operating System**: any
- **Fetched as**
  - Binary release
### Description

When the population of a colony with a drydock drops to zero (for whatever reason, in my case because I CC'ed it), an outpost with a drydock is left. If you park damaged ships at that outpost, you will get sitreps telling you that these ships can't be repaired due to low happiness (most likely due to the fact that the happiness meter of outposts is zero).
### Expected Result

Drydocks left on colonies that got reduced to outposts should not be able to do repairs anyway, so those sitreps should not show up at all.
### Steps to reproduce
- Well, park some damaged ships at an outpost with a drydock.
- Enjoy the sitreps. 😉

dbenage-cx:
Prevent further repair related effect processing when source planet has 0 population.

Fixes #970


default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
```diff
@@ -15,6 +15,24 @@ BuildingType
...
+            ]
+            activation = And [
+                Source
+                ContainedBy And [
+                    Planet
+                    Population high = 0
+                ]
+            ]
+            stackinggroup = "SHIP_REPAIR"
+            priority = 75
```
Vezzra, line 42: How important is the priority for this to work? If the priority here being lower than those of the other effectsgroups is required for the no repair with no pop, then this is a potential source for troubles. If anyone decides to lower the values assigned to the priority macros, that will screw up the setup of the drydock effectsgroups.

Is it possible to calculate the priority? So we can do something like: `priority = [[VERY_EARLY_PRIORITY]] - 5`?

default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
```diff
@@ -15,6 +15,24 @@ BuildingType
...
+            effects = RemoveSpecial name = ""
```
Vezzra: Just a sidenote: we need to provide a noop effect. While constructs like this do the job, it's not very pretty.

Vezzra:
Aside from my comments above, looks good and works. I want to wait for @MatGB's response in #970 before I merge though, as we still might decide that instead of giving no sitrep at all, we provide one stating that no repairs are possible because the drydock is on an outpost.

default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
geoffthemedio: Effectsgroup priority is stored as a simple int, and there is no parse-time calculation of anything.

https://github.com/freeorion/freeorion/blob/master/universe/Effect.h#L41
https://github.com/freeorion/freeorion/blob/master/parse/Parse.cpp#L87

default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
geoffthemedio: Add an issue...

geoffthemedio:
Separate sitrep can be separately hidden / filtered by the player, which is helpful if they still want to receive the no-repair due to happiness ones.

default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
dbenage-cx: Critical it is before any effects it should prevent, though that is the case with most effects with a specified priority.
There was talk about using the same stackinggroup for effects in other files, so I wanted to leave room before the noop when they are not planet based.

priority only accepts a constant currently.

dbenage-cx:
Will add a new sitrep later tonight.

Vezzra:
> Will add a new sitrep later tonight.

Excellent. Shouldn't be too hard, replacing the current noop effect with a proper sitrep should do the job I guess?

default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
Vezzra: Hm, guess we'll have to live with that solution then, but I still foresee potential issues...

dbenage-cx:
All issues so far should be addressed.

@Ouaz : When structural changes occur, is it easier for the obsolete entry to be removed or changed as above?

Ouaz:
@dbenage-cx 

Generally, as I check new strings or modifications in en.txt each week,  I prefer that the FR file is not  modified.^^

But, as it's a special case (fix for the release, that needs to be cherry-picked), it's OK and more convenient to include the FR modifications in the PR.

If you can just modify the new FR strings like that and update the commit (better translation and missing EFFECT_DRYDOCK_SHIP_REPAIR_NOPOP).

Thanks!

line 5336 to 5353:

```

SITREP_SHIP_REPAIR_DOCK_NONE
Réparation en Cale Sèche - Aucune

SITREP_SHIP_REPAIR_DOCK_NOPOP
Réparation en Cale Sèche - Pas de Population

SITREP_SHIP_REPAIR_DOCK_PARTIAL
Réparation en Cale Sèche - Partielle

SITREP_SHIP_REPAIR_DOCK_COMPLETE
Réparation en Cale Sèche - Achevée

EFFECT_DRYDOCK_SHIP_REPAIR_NONE
Sur %planet% : la structure %building% ne peut pas démarrer les réparations sur l'astronef %ship%, en raison d'un indicateur [[encyclopedia HAPPINESS_TITLE]] trop faible.

EFFECT_DRYDOCK_SHIP_REPAIR_NOPOP
Sur %planet% : la structure %building% étant abandonnée, plus aucun personnel n'est présent pour effectuer des réparations sur l'astronef %ship%.
```

dbenage-cx:
Thank you @Ouaz, updated and rebased.

default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
Vezzra: That should work, thanks.

Still not ideal, because a change to the values of the priority macros could potentially mess up the interaction with the other repair techs/mechanics (which use those macros), but probably far better and more robust than the lonely "75" value we had before. At least changing the priority macros can't mess up drydock repair itself now.

Good enough for now. A more thorough solution is beyond the scope of his PR anyway, as that's a more general issue.

default/customizations/common_user_customizations.txt
```diff
@@ -109,6 +109,10 @@ SITREP_MONSTER_FLEET_ARRIVED_AT_DESTINATION_LABEL
 SITREP_MONSTER_SHIP_ARRIVED_AT_DESTINATION_LABEL
 SITREP_FLEET_ARRIVED_AT_DESTINATION_LABEL
 SITREP_PLANET_BOMBARDED_AT_SYSTEM_LABEL
-SITREP_UNOWNED_PLANET_BOMBARDED_AT_SYSTEM_LABEL'''
+SITREP_UNOWNED_PLANET_BOMBARDED_AT_SYSTEM_LABEL
+SITREP_SHIP_REPAIR_DOCK_NOPOP
+SITREP_SHIP_REPAIR_DOCK_NONE
+SITREP_SHIP_REPAIR_DOCK_PARTIAL
+SITREP_SHIP_REPAIR_DOCK_COMPLETE'''
```
Vezzra, line 116: This will have the side effect that drydock repair sitreps appear above all the sitreps that have no entry in this list (thus getting moved up), but that shouldn't be an issue.

Vezzra:
Seems to work as intended, will merge.

dbenage-cx:
Thanks for reviewing

PR #975, opened 2016-09-14
Fix Wnd::DetachChild(wnd) to null m_layout, if wnd equals m_layout.

LGM-Doyle:
DetachChild(wnd) does not set m_layout to null if the child, wnd, is the
m_layout. Consequently, calling DetachChildren() will eventually cause a
crash because GetLayout() returns m_layout whose Parent() is null.

This commit fixes that bug.

I exposed this bug with new code that I am working on.  I do not think it is critical that it be included in the 0.4.6 candidate, but you may be of a different mind.

### DIFF
--- a/GG/src/Wnd.cpp
+++ b/GG/src/Wnd.cpp
@@ -590,6 +590,8 @@ void Wnd::DetachChild(Wnd* wnd)
         return;
     m_children.erase(it);
     wnd->SetParent(0);
+    if (wnd == m_layout)
+        m_layout = 0;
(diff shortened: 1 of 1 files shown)


geoffthemedio:
looks / sounds reasonable...

Vezzra:
>  I do not think it is critical that it be included in the 0.4.6 candidate, but you may be of a different mind.

A fix is a fix, we need to produce another RC anyway, so I think it should go in. Will cherry pick.

Issue #977, opened 2016-09-17
Duplicate SitRep labels in Filters list


dbenage-cx:
### Environment
- **FreeOrion Version**: 0.4.6
- **Operating System**: N/A
- **Graphic card used**: N/A
- **Fetched as** 
  - Compiled from source
### Description

The following labels are used in sitreps, with a paired entry that contains the same default (en) text.
This leads to confusion as to which label to select if a user wishes to omit one.

`SITREP_COMBAT_SYSTEM_LABEL` | `SITREP_COMBAT_SYSTEM_ENEMY_LABEL`

`SITREP_GROUND_BATTLE_LABEL` | `SITREP_GROUND_BATTLE_ENEMY_LABEL`

These also conflict in the french stringtable.

In the german stringtable, only the first has a conflict.
### Expected Result

Labels should contain a unique entry.
### Steps to reproduce
- Play a game until the player has combat both with a single empire and with multiple empires at the same time.
- Click Filters

Vezzra:
I consider that solved by #978, so closing this issue.

PR #978, opened 2016-09-18
Change conflicting EN sitreps labels to unique entries

dbenage-cx:
Changes the default stringtable entries related to #977


dbenage-cx:
Appears `SITREP_COMBAT_SYSTEM_LABEL` is also used in the case of combat with ships with no known empire (monsters/stealth).

PR #981, opened 2016-09-23
Move GG/Config.h into the GG tree

Vezzra:
Currently there is a copy in the msvc2013 directory tree, apparently with the idea that each build system should have it's own. However, it does not look like this is really required, so moving this file into the GG tree should simplify things.


Vezzra:
The question about this file came up [here](https://github.com/freeorion/freeorion-sdk/issues/6#issuecomment-248345529). Are there any reasons not to just move it to `GG/GG/Config.h` and remove it from all other locations?

geoffthemedio:
Historically the intention was probably to set up that file depending what you built GG with, or for it to be generated as part of the GG configuration... not sure though.

Vezzra:
@geoffthemedio: so you're ok with applying this change?

geoffthemedio:
I have no strong opinion.

Vezzra:
Ok, merged it. The change is simple enough to easily undo should that become necessary.

l29ah:
It is being regenerated at the build time with some comments. Maybe it should be added to .gitignore?
```
--- a/GG/GG/Config.h
+++ b/GG/GG/Config.h
@@ -1,3 +1,4 @@
+// -*- C++ -*-
 /* GG is a GUI for SDL and OpenGL.
    Copyright (C) 2003-2008 T. Zachary Laine
 
@@ -22,9 +23,13 @@
    Zach Laine
    whatwasthataddress@gmail.com */
 
+/** \file Config.h \brief Contains build-generated configuration macros used
+    throughout GG. */
+
 #ifndef _GG_Config_h_
 #define _GG_Config_h_
 
+
 #define GG_HAVE_LIBPNG 1
 #define GG_HAVE_LIBTIFF 0
 
```

Vezzra:
I don't think that adding a file under version control to `.gitignore` is a good idea. However, assuming that it will always get generated with the same contents, committing it once should take care of the issue that it always shows up as modified.

The other solution would be to remove that file from version control entirely and indeed add it to `.gitignore`. That would actually be the right, clean solution. However, AFAIK, the manually maintained MSVC and Xcode projects don't generate that file, but still need it. So we can't do this until @adrianbroher manages to get the cmake scripts to the point where we can replace the manually maintained projects with cmake generated ones.

@adrianbroher?

adrianbroher:
@l29ah
Build out of source.  Other setups are not supported.  @Vezzra already explained the reason why it was done this way.

l29ah:
On Fri, Apr 21, 2017 at 07:18:31AM -0700, Marcel Metz wrote:
> Build out of source.  Other setups are not supported.

Why?

-- 

Vezzra:
Building out of source is cleaner, has less potential for issues.

Issue #907, opened 2016-08-22
Purpose of FUNCTIONAL_* user string entries


adrianbroher:
### Description

Currently we use the UserString.\* API to query some configuration values from files.  I do think this is a subpar solution for querying those configuration and was wondering what was the rationale behind this and why are those values not part of the OptionsDB.  The problems I see are:
- Tools to extract translation keys from the code  need to be more complex to exclude FUNCTIONAL_ entries.
- UserString self documentation implies that you give a key and get a localized string. Except for the FUNCTIONAL_ entries.
- StringTables now need to support (needlessly) a `#include` syntax to keep the configuration values out of the translators reach.
- Translation keys can't be chosen abitrary, but need to stay away from the `FUNCTIONAL_` namespace.

tl;dr: The current implementation intermingles to the tasks of 'give key, get translation' and 'give key, get configuration value'.

Only references I found so far are:

http://freeorion.org/forum/viewtopic.php?p=83590#p83590

Vezzra:
Agree to everything you said. The discussion you linked to already contains proposals how to fix this. I agree with @Dilvish-fo that we should avoid code duplication by providing two separate interfaces in the backend code that do essentially the same technically ("give string key, get some string value").

So the idea to provide a generic implementation that can be used for value retrieval by key for both localized text and configuration values sounds very reasonable to me, we just need to move all those FUNCTIONAL_\* entries out of the stringtables and put them in dedicated configuration files (which are stored at a location that makes more sense than the stringtables...).

> what was the rationale behind this

AFAIK, because it has been quick and easy. I don't think that there is really any oder reason behind it.

adrianbroher:
Thanks for the answer Vezzra, but unfortunately it doesn't answer why are those valuee not part of the OptionsDB.

> […] that do essentially the same technically ("give string key, get some string value").

But this isn't the case. One use case converts to float and one converts it to a list of string, which is something very different than a string.

geoffthemedio:
The OptionsDB probably doesn't presently allow or at least handle correctly multi-line strings as a data type.

The stuff in
https://github.com/freeorion/freeorion/blob/master/default/customizations/common_user_customizations.txt
seems to be UI controls, though, so would make sense to put in OptionsDB. (I would avoid moving anything content-related into the OptionsDB, until there's a general game-rules-configuration mechanism in place).

MatGB:
Ah, that's where they are now.

[...]

PR #984, opened 2016-09-25
Move shipyard building list from UserString to definition tag

dbenage-cx:
Removes the UserString defined list in favor of `CTRL_UNDERLINE` definition tag.

Partial proposal towards #907

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -1376,9 +1376,7 @@ void SidePanel::PlanetPanel::Refresh() {
             continue;
         if (known_destroyed_object_ids.find(*building_it) != known_destroyed_object_ids.end())
             continue;
-        std::list<std::string> shipyards_list;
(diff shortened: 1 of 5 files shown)


adrianbroher (changes requested):
The `TAG_UNDERLINE` contains no semantic.  `SHIPYARD` would be more appropriate as the underline below a system represents the fact that this system or planet contains a shipyard.

MatGB:
Part of the way it was done before is to allow users to change things so other things are underlined instead of shipyards—it would be best to keep that ability, I've in the past changed it to underline transformers and stargates which I found more useful and also once changed it to drydocks. In the future when we've reduced the number of shipyards further it would be preferable to be able to easily change what is or isn't underlined in scripting.

This is especially useful for script test purposes where you need an easy identifier for where a specific building is.

dbenage-cx:
Rebased using `TAG_SHIPYARD`/`CTRL_SHIPYARD`

PR #985, opened 2016-09-25
Move ship design redundancy settings from UserString

dbenage-cx:
PR relocates settings used with the `Redundant` parts filter within ship design.

Moves the bargain, cost, and time ratios to OptionsDB.

Removes skip list in favor of definition tags.

Partial proposal towards #907

### DIFF
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -52,6 +52,19 @@ namespace {
     const GG::Y         SLOT_CONTROL_HEIGHT(60);
     const int           PAD(3);
 
+    void AddOptions(OptionsDB& db) {
+        db.Add<float>("UI.ship-design.part-filters.ratio-bargain-min",
+                      UserStringNop("OPTIONS_DB_UI_DESIGN_PART_FILTER_MIN_BARGAIN"),
+                      1.0,     RangedStepValidator<float>(0.1, 1.0, 10.0));
+        db.Add<float>("UI.ship-design.part-filters.ratio-cost-max",
+                      UserStringNop("OPTIONS_DB_UI_DESIGN_PART_FILTER_MAX_COST"),
+                      1.0,     RangedStepValidator<float>(0.1, 1.0, 10.0));
+        db.Add<float>("UI.ship-design.part-filters.ratio-time-max",
+                      UserStringNop("OPTIONS_DB_UI_DESIGN_PART_FILTER_MAX_TIME"),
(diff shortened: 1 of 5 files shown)


UI/DesignWnd.cpp
```diff
@@ -491,9 +508,18 @@ PartsListBox::PartsListBox(void) :
...
     m_show_superfluous_parts(true),
-    m_previous_num_columns(-1)
+    m_previous_num_columns(-1),
+    m_ratio_bargain_min(1.0),
+    m_ratio_cost_max(1.0),
+    m_ratio_time_max(1.0)
 { SetStyle(GG::LIST_NOSEL); }
 
+void PartsListBox::InitOptions() {
+    m_ratio_bargain_min = GetOptionsDB().Get<float>( "UI.ship-design.part-filters.ratio-bargain-min");
```
geoffthemedio: get rid of the extra space between ( and "

default/stringtables/en.txt
```diff
@@ -1725,6 +1725,56 @@ Network port to use for server discovery.
 OPTIONS_DB_NETWORK_MESSAGE_PORT
 Network port to use for client-server messaging.
 
+OPTIONS_DB_UI_DESIGN_PART_FILTER_DESC
+'''DesignWnd Redundant Part Filtering
+
+Deems a first part redundant to a second only if the second is all of:
+
+(i) more powerful
```
geoffthemedio, line 1751: what about otherwise equal but less expensive, or otherwise equal but faster to build parts?

UI/DesignWnd.cpp
```diff
@@ -52,6 +52,19 @@ namespace {
     const GG::Y         SLOT_CONTROL_HEIGHT(60);
     const int           PAD(3);
 
+    void AddOptions(OptionsDB& db) {
+        db.Add<float>("UI.ship-design.part-filters.ratio-bargain-min",
+                UserStringNop("OPTIONS_DB_UI_DESIGN_PART_FILTER_MIN_BARGAIN"),
```
geoffthemedio: indent parameters to match left-side of first parameter

geoffthemedio (changes requested):

default/stringtables/en.txt
dbenage-cx: I aimed to relocate the documentation here, along with minor formatting adjustments.
According to this, both cases fail the first check, so the first part would not be deemed redundant.

Vezzra:
@dbenage-cx, a short comment why you closed this as invalid would be good...

Issue #855, opened 2016-08-10
loading save list is slow


Voker57:
If I launch freeoriond on remote server and host on it, loading list of autosaves (~50-100) takes several minutes and freeorion appears to be frozen in progress. It probably consumes a lot of bandwidth too (unless it's CPU bound?)

geoffthemedio:
What version?

Voker57:
0cab3265af9777a0f5fb9 with unrelated, can test on latest if needed

l29ah:
It takes a lot of time even locally, consuming 100% of one cpu.

geoffthemedio:
What happens if you switch to binary save files? Clear out the saves folder, enable "Create Binary Save Files" in the options/other tab, load a big game, play a few turns, make a few copies of the save files so you have ~100 of them or so, then try opening that directory in the save dialog.

geoffthemedio:
@l29ah @Voker57 Does switching to binary serialization and removing all the XML-based saves help the save list load time at all?

Vezzra:
Status on this? @l29ah, @Voker57, can you provide feedback to @geoffthemedio's question above?

sumpfralle:
I also experienc a performance issue whenever the load/save dialog opened.

I did the following:
* run freeorion
* choose "Load Game"
* immediate interactive result: the file open dialog showing the current savegame directory (which is empty - there are only games in the "auto" sub-directory)
* change to "auto" directory
* everything freezes for a number of seconds
* the list of files and their details are shown

The savegames are stored as xml files (not binary).

I tested, if the number of files is related to the duration of the high-load event. I could confirm this thesis (for me: 9 seconds per file).

I watched the load of the freeorion process via "top". It went up to 100% for the duration described above.

Regarding the process scheduling: my computer has two CPU cores. Thus a high-load situation caused by a single process is usually not a problem. In this specific case it seemed like freeorion managed to actively prevent the usual process scheduling, since I could not move the mouse pointer during the stressful time and I even could not switch to the non-X login consoles (which always works here even under high load). Additionally the freeorion process did not react on a kill signal (SIGTERM, 15). "SIGKILL" worked, of course.

The savegames have a filesize between one and two megabytes.

Switching between workspaces/applications causes another savegame parse run, whenever the freeorion dialog is refreshed / shown again. Maybe it would be good to cache the result of a parse operation (at least during runtime).

[...]

PR #988, opened 2016-09-26
Remove binding of SaveFileDialog::UpdatePreviewList to FocusChangedSignal.

LGM-Doyle:
This might contribute to long loading times, since it will reload all of
the data in the SaveFileDialog if focus changes.

This would contributes to long loading times in #855, if the user shifted focus away from freeorion while waiting.

### DIFF
--- a/UI/SaveFileDialog.cpp
+++ b/UI/SaveFileDialog.cpp
@@ -665,7 +665,6 @@ void SaveFileDialog::Init() {
     GG::Connect(m_file_list->DoubleClickedSignal,               &SaveFileDialog::DoubleClickRow,    this);
     GG::Connect(m_name_edit->EditedSignal,                      &SaveFileDialog::FileNameEdited,    this);
     GG::Connect(m_current_dir_edit->EditedSignal,               &SaveFileDialog::DirectoryEdited,   this);
-    GG::Connect(GG::GUI::GetGUI()->FocusChangedSignal,          boost::bind(&SaveFileDialog::UpdatePreviewList, this));
 
     if (!m_load_only) {
         m_name_edit->SetText(std::string("save-") + FilenameTimestamp() + m_extension);


geoffthemedio:
This would cause loss of functionality - updating the list after switching windows, which could potentially have allowed the user to change the directory contents.

LGM-Doyle:
I had not thought of that case.

However, the way it is now if the load takes long enough for the user to minimize freeorion or switch to another task and check back, then it will never finish loading.

With this commit, if the user changes the directory, they will probably close and re-open the dialog or change directories, which will re-read the files.

Ideally, we need a directory changed signal and handler.

I think that including this commit is the better of the two non-ideal options.

MatGB:
How viable would it be to have a warning of some sort if the directory has a large number of saves in it? I habitually move older savegames into a different folder after awhile—we know there's a group using a fileserver somewhere to store their saves, if the folder they're using is big then it's bound to fail.

LGM-Doyle:
@MatGB after the last PR in this series many of those problem in #855 should improve a lot.  

From my profiling most of the time was used in laying out and sorting SaveFile::Rows that were not visible.  Parsing the file headers is fast by comparison.

The SaveFileDialog changes insert the files as a vector which will improve sorting time.

I changed the sort so that it doesn't require a layed out ListBox::Row.

The new style of ListBox only lays out visble rows, so display time should scale with the number of visible rows not the number of files in the directory to a first order.

I did not test it with a large number of save files to determine when the second order effect of number of files in the directory begins to dominate.

Vezzra:
@LGM-Doyle, what's the status on this? It seems to be the last of your listbox improving PRs that hasn't been merged. I actually agree with your assessment:

> I think that including this commit is the better of the two non-ideal options.

So I think this should be merged. The issue @geoffthemedio mentioned:

> This would cause loss of functionality - updating the list after switching windows, which could potentially have allowed the user to change the directory contents.

...can be dealt later IMO. I'd rather include this in 0.4.7 is possible, so I'm keeping the "0.4.7" milestone.

geoffthemedio:
I don't have major objections...

Vezzra:
@LGM-Doyle, I guess you missed my above inquiry for a status update on this. So I'll do a bump and ask again, can we merge this? See my comments above, I think this should be merged as is.

As @geoffthemedio gave his consent, I'm going to merge unless you object.

LGM-Doyle:
I've gone ahead and merged it.

PR #989, opened 2016-09-26
Improve GG::Font::DetermineLines()

LGM-Doyle:
This PR is a collection of improvements to `GG::Font` to improve its
performance.

Previously `Font::DetermineLines()` consumed 25% of the processing time when
the UI was stalling freeorion.  Within `DetermineLines()` most of the time was
spent parsing the XML strings into TextElements used by Font to break the
string into known style tags, unknown style tags and text. 

The PR breaks out the XML to TextElements computation into
`ComputeTextElements()`.  It then removes gotcha variants of
`DetermineLines()` so that Font can be used efficiently.  It short-circuits
logic for common operations like parsing an empty string.

Repeated use of `ComputeTextElements()` in tight loops is still a
performance hazard. I was unable to improve the usage of `boost::xpressive`
as much as I had hoped.

### DIFF
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -272,7 +272,7 @@ public:
 
     /** \brief TextAndElementsAssembler is used to assemble a matched pair of text and a vector of
         TextElement, without incurring the computational cost of parsing the text with
-        ComputeTextElements().
+        ExpensiveParseFromTextToTextElements().
 
         The pair of string and vector returned by Text() and Elements() are consistent with each
         other and can be used with the fast constructor or the fast SetText variant of TextControl.
@@ -571,8 +571,9 @@ public:
         string \p text to be consistent even if the \p new_text is longer/shorter than the original
         TEXT type TextElement.
 
-        This does not recompute the text_elements. It is faster than ComputeTextElements on a new
-        string. It will not find white space in the inserted text.
+        This does not recompute the text_elements. It is faster than
+        ExpensiveParseFromTextToTextElements on a new string. It will not find white space in the
+        inserted text.
 
         \p text and \p text_elements are assumed to be consistent with each other and both will be
         changed to remain consistent.
--- a/GG/GG/TextControl.h
+++ b/GG/GG/TextControl.h
@@ -95,6 +95,16 @@ public:
     //@}
     virtual ~TextControl();
 
+    /** Assignment operator.
+
+        Text Control requires an assignment operator because m_text_elements contains pointers
+        to m_text which need to be bound with Bind() to the m_text in this TextControl.
+
+        Using the assignment operator is faster than using SetText() with text from another
+        TextControl because it avoids the XML parse overhead.
+    */
+    TextControl& operator=(const TextControl& that);
+
     /** \name Accessors */ ///@{
     virtual Pt        MinUsableSize() const;
 
--- a/GG/GG/TextControl.h
+++ b/GG/GG/TextControl.h
@@ -78,8 +78,8 @@ public:
     /** Fast constructor.
 
      This constructor requires a \p str and \text_elements that are consistent with each other.
-     Font::ComputeTextElements() will not be called on \p str.  Hence this constructor is much
-     faster than the first constructor.*/
+     Font::ExpensiveParseFromTextToTextElements() will not be called on \p str.  Hence this
+     constructor is much faster than the first constructor.*/
     TextControl(X x, Y y, X w, Y h, const std::string& str,
                 const std::vector<boost::shared_ptr<Font::TextElement> >&text_elements,
                 const boost::shared_ptr<Font>& font,
--- a/GG/GG/TextControl.h
+++ b/GG/GG/TextControl.h
@@ -198,11 +198,16 @@ public:
         the newly rendered text occupies. */
     virtual void SetText(const std::string& str);
 
-    /** Sets the text displayed in this control to \a str \p text_elements
-        pair.  This is faster than SetText without \p text_elements. May resize
-        the window.  If the control was constructed with FORMAT_NOWRAP, calls
-        to this function cause the window to be resized to whatever space the
-        newly rendered text occupies. */
+    /** Sets the text displayed in this control to the \p str \p text_elements
+        pair.  This is faster than SetText without \p text_elements.
+
+        This may resize the window.  If the control was constructed with FORMAT_NOWRAP, calls to
+        this function cause the window to be resized to whatever space the newly rendered text
+        occupies.
+
+        If the \p str and \p text_elements are inconsistent and \p str is shorter than expected
+        from examining \p text_elements then it will return without changing the TextControl.
+    */
     virtual void SetText(const std::string& str,
                          const std::vector<boost::shared_ptr<Font::TextElement> >&text_elements);
 
--- a/GG/GG/TextControl.h
+++ b/GG/GG/TextControl.h
@@ -316,6 +316,10 @@ private:
     void RefreshCache();
     void PurgeCache();
 
+    /** Recompute line data, code points, text extent and minusable size cache when
+        m_text_elements changes.*/
+    void RecomputeLineData();
+
     std::string                 m_text;
     Flags<TextFormat>           m_format;      ///< the formatting used to display the text (vertical and horizontal alignment, etc.)
     Clr                         m_text_color;  ///< the color of the text itself (may differ from GG::Control::m_color)
--- a/GG/GG/TextControl.h
+++ b/GG/GG/TextControl.h
@@ -74,7 +74,12 @@ public:
     TextControl(X x, Y y, X w, Y h, const std::string& str, const boost::shared_ptr<Font>& font,
                 Clr color = CLR_BLACK, Flags<TextFormat> format = FORMAT_NONE,
                 Flags<WndFlag> flags = NO_WND_FLAGS);
-    /** Fast constructor.*/
+
+    /** Fast constructor.
+
+     This constructor requires a \p str and \text_elements that are consistent with each other.
+     Font::ComputeTextElements() will not be called on \p str.  Hence this constructor is much
+     faster than the first constructor.*/
     TextControl(X x, Y y, X w, Y h, const std::string& str,
                 const std::vector<boost::shared_ptr<Font::TextElement> >&text_elements,
                 const boost::shared_ptr<Font>& font,
--- a/GG/GG/TextControl.h
+++ b/GG/GG/TextControl.h
@@ -74,6 +74,12 @@ public:
     TextControl(X x, Y y, X w, Y h, const std::string& str, const boost::shared_ptr<Font>& font,
                 Clr color = CLR_BLACK, Flags<TextFormat> format = FORMAT_NONE,
                 Flags<WndFlag> flags = NO_WND_FLAGS);
+    /** Fast constructor.*/
+    TextControl(X x, Y y, X w, Y h, const std::string& str,
+                const std::vector<boost::shared_ptr<Font::TextElement> >&text_elements,
+                const boost::shared_ptr<Font>& font,
+                Clr color = CLR_BLACK, Flags<TextFormat> format = FORMAT_NONE,
+                Flags<WndFlag> flags = NO_WND_FLAGS);
     //@}
     virtual ~TextControl();
 
--- a/GG/GG/TextControl.h
+++ b/GG/GG/TextControl.h
@@ -80,6 +80,18 @@ public:
                 const boost::shared_ptr<Font>& font,
                 Clr color = CLR_BLACK, Flags<TextFormat> format = FORMAT_NONE,
                 Flags<WndFlag> flags = NO_WND_FLAGS);
+
+    /** Copy constructor.
+
+        Text Control requires a copy-constructor because m_text_elements contains pointers
+        to m_text which need to be bound with Bind() to the m_text in the new TextControl.
+
+        Using the copy constructor is faster than constructing a TextControl with text from another
+        TextControl because it avoids the XML parse overhead.
+
+        Since Control does not have a way to access Flags the copy using default flags.
+    */
+    explicit TextControl(const TextControl& that);
     //@}
     virtual ~TextControl();
 
--- a/GG/src/TextControl.cpp
+++ b/GG/src/TextControl.cpp
@@ -336,7 +336,6 @@ void TextControl::SizeMove(const Pt& ul, const Pt& lr)
         Pt text_sz = m_font->TextExtent(m_line_data);
         m_text_ul = Pt();
         m_text_lr = text_sz;
-        AdjustMinimumSize();
         PurgeCache();
     }
     RecomputeTextBounds();
--- a/GG/src/TextControl.cpp
+++ b/GG/src/TextControl.cpp
@@ -55,6 +55,26 @@ TextControl::TextControl(X x, Y y, X w, Y h, const std::string& str, const boost
     SetText(str);
 }
 
+TextControl::TextControl(X x, Y y, X w, Y h, const std::string& str,
+                         const std::vector<boost::shared_ptr<Font::TextElement> >& text_elements,
+                         const boost::shared_ptr<Font>& font,
+                         Clr color /*= CLR_BLACK*/, Flags<TextFormat> format /*= FORMAT_NONE*/,
+                         Flags<WndFlag> flags /*= NO_WND_FLAGS*/) :
+    Control(x, y, w, h, flags),
+    m_format(format),
+    m_text_color(color),
+    m_clip_text(false),
+    m_set_min_size(false),
+    m_code_points(0),
+    m_font(font),
+    m_render_cache(0),
+    m_cached_minusable_size_width(X0),
+    m_cached_minusable_size(Pt())
+{
+    ValidateFormat();
+    SetText(str, text_elements);
+}
+
 TextControl::~TextControl()
 {
     delete m_render_cache;
--- a/GG/src/TextControl.cpp
+++ b/GG/src/TextControl.cpp
@@ -61,10 +61,12 @@ TextControl::TextControl(X x, Y y, X w, Y h, const std::string& str,
                          Clr color /*= CLR_BLACK*/, Flags<TextFormat> format /*= FORMAT_NONE*/,
                          Flags<WndFlag> flags /*= NO_WND_FLAGS*/) :
     Control(x, y, w, h, flags),
+    m_text(),
     m_format(format),
     m_text_color(color),
     m_clip_text(false),
     m_set_min_size(false),
+    m_text_elements(),
     m_code_points(0),
     m_font(font),
     m_render_cache(0),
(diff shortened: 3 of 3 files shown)


adrianbroher:
Too massive to review and a cursory review gives me the impression that this PR is not cleaned up (adding and removing the same code in the same PR).

LGM-Doyle:
@adrianbroher , the added and then removed code are all either timers, test code or to maintain functionality.  

Since freeorion doesn't have any test code for functionality or performance I left in adding and removing the test code so that anyone interested could checkout the timers/tests and verify the performance/agreement with previous results.

In some case I did not see a way to add/change functionality while maintaining code that compiled, so I added a bridge, changed the code in reasonable steps and then removed the bridge.

The other alternatives are one monster commit, or committing code where some middle steps are non-functional.

GG/GG/Font.h
```diff
@@ -146,6 +146,9 @@ class GG_API Font
             pair.second. */
         Substring(const std::string& str_, const IterPair& pair);
 
+        /** Rebind to a different string.*/
+        void Bind(const std::string& str_);
```
geoffthemedio, line 159: What does binding to a string do, and why are there two (three?) Bind functions?

GG/GG/Font.h
```diff
@@ -446,7 +493,7 @@ class GG_API Font
                     std::size_t end_line, CPSize end_char) const;
 
     void PreRenderText(const Pt& ul, const Pt& lr, const std::string& text, Flags<TextFormat>& format,
```
geoffthemedio, line 532: needs explanitory comment

GG/GG/Font.h
```diff
@@ -461,35 +508,41 @@ class GG_API Font
     void ProcessTagsBefore(const std::vector<LineData>& line_data, RenderState& render_state,
                            std::size_t begin_line, CPSize begin_char) const;
 
-    /** Returns the maximum dimensions of the string in x and y, and populates
-        \a line_data. */
-    Pt   DetermineLines(const std::string& text, Flags<TextFormat>& format, X box_width,
-                        std::vector<LineData>& line_data) const;
+    /** Populate \p text_elements with TextElements parsed from \p
+        text. \p format onyl determines if tags are ignored. */
```
geoffthemedio: typo onyl

geoffthemedio: meaning of "\p format onyl determines if tags are ignored." is unclear.

GG/GG/Font.h
```diff
@@ -233,6 +241,40 @@ class GG_API Font
...
+    class GG_API TextElements
+    {
+    public:
+        TextElements(const Font& font);
+        ~TextElements();
+
+        /** Return the constructed text.*/
+        const std::string& Text();
+        /** Return the constructed TextElements.*/
+        const std::vector<boost::shared_ptr<TextElement> >& Elements();
```
geoffthemedio, line 289: What is the difference / purpose of having a TextElements class that returns a vector of TextElement?

GG/GG/Font.h
```diff
@@ -461,35 +508,41 @@ class GG_API Font
...
-    /** Returns the maximum dimensions of the string in x and y, and populates
-        \a line_data and \a text_elements.  Note that \a text_elements must be
-        empty. */
-    Pt   DetermineLines(const std::string& text, Flags<TextFormat>& format, X box_width,
-                        std::vector<LineData>& line_data,
-                        std::vector<boost::shared_ptr<TextElement> >& text_elements) const;
+    /** Fill \p text_elements with the font widths of characters from \p text starting from \p
+        start. */
+    void FillTemplatedText(
+        const std::string& text,
```
geoffthemedio: put first parameter on same line as function name, indent other parameters to match where it starts

GG/GG/TextControl.h
```diff
@@ -74,6 +74,12 @@ class GG_API TextControl : public Control
     TextControl(X x, Y y, X w, Y h, const std::string& str, const boost::shared_ptr<Font>& font,
                 Clr color = CLR_BLACK, Flags<TextFormat> format = FORMAT_NONE,
                 Flags<WndFlag> flags = NO_WND_FLAGS);
+    /** Fast constructor.*/
```
geoffthemedio: "Fast" ?

GG/GG/TextControl.h
```diff
@@ -165,6 +171,20 @@ class GG_API TextControl : public Control
         the newly rendered text occupies. */
     virtual void SetText(const std::string& str);
 
+    /** Sets the text displayed in this control to \a str \p text_elements
+        pair.  This is faster than SetText without \p text_elements. May resize
+        the window.  If the control was constructed with FORMAT_NOWRAP, calls
+        to this function cause the window to be resized to whatever space the
+        newly rendered text occupies. */
+    virtual void SetText(const std::string& str,
```
geoffthemedio, line 211: Does something weird happen if the contents of text_elements is inconsistent with str?

GG/GG/TextControl.h
```diff
@@ -248,6 +268,10 @@ class GG_API TextControl : public Control
     void RefreshCache();
     void PurgeCache();
 
+    /** Recompute line data, code points, text extent and minusable size cache when
+        m_text_elements changes.*/
```
geoffthemedio, line 325: Meaning of "when m_text_elements changes". Does this immediately do something, or just set a flag so something happens later if something changes?

GG/src/Edit.cpp
```diff
@@ -122,17 +122,15 @@ void Edit::Render()
 
     BeginScissorClipping(Pt(client_ul.x - 1, client_ul.y), client_lr);
 
-    if (GetLineData().empty())
```
geoffthemedio: why?

GG/src/Edit.cpp
```diff
@@ -449,9 +451,11 @@ void Edit::KeyPress(Key key, boost::uint32_t key_code_point, Flags<ModKey> mod_k
         } else if (0 < m_cursor_pos.second) {
```
geoffthemedio, line 451: some comments in here explaining what you're doing / is happening would be helpful...

GG/src/Font.cpp
```diff
@@ -514,6 +483,179 @@ std::pair<std::size_t, CPSize> GG::LinePositionOf(CPSize index, const std::vecto
...
+            m_known_tags.clear();
+        }
+
+        bool IsKnown(const std::string &tag) {
+            return m_known_tags.find(tag) != m_known_tags.end();
+        }
+
+        // Return a regex bound to \p text using the currently known
+        // tags possible \p ignore_tags and/or \p strip_unpaired_tags
+        sregex & Regex(const std::string& text, bool ignore_tags, bool strip_unpaired_tags = false) {
```
geoffthemedio: remove space before & ?

GG/src/Font.cpp
```diff
@@ -514,6 +483,179 @@ std::pair<std::size_t, CPSize> GG::LinePositionOf(CPSize index, const std::vecto
...
+        // m_tag_stack is used to track XML opening/closing tags.
+        std::stack<Font::Substring> m_tag_stack;
+
+        // The combined regular expression.
+        sregex m_EVERYTHING;
+    };
+
+    class TagHandler {
+        public:
+        TagHandler() : m_known_tags(),
```
geoffthemedio: move initializer list to newline, indented 4 spaces

GG/src/Font.cpp
```diff
@@ -548,6 +695,191 @@ StrSize Font::TextElement::StringSize() const
...
+    std::string m_text;
+    std::vector<boost::shared_ptr<TextElement> > m_text_elements;
+    bool m_are_widths_calculated;
+};
+
+
+Font::TextElements::TextElements(const Font& font) :
+    pimpl(new TextElementsImpl(font))
+{}
+// Required because Impl is defined here
```
geoffthemedio, line 902: add blank lines between } of one function and class name of the next (or comment)

GG/src/Font.cpp
```diff
@@ -1253,9 +1377,238 @@ Pt Font::DetermineLinesImpl(const std::string& text,
             std::cout << "\n";
         }
         std::cout << std::endl;
+    }
+
+    void PrintLineBreakdown(const std::string& text,
+                       const Flags<TextFormat>& format,
```
geoffthemedio: indent to match first parameter

GG/src/Font.cpp
```diff
@@ -1253,9 +1377,238 @@ Pt Font::DetermineLinesImpl(const std::string& text,
...
+                       const X box_width,
+                       const std::vector<Font::LineData>& line_data) {
```
geoffthemedio: { to next line

GG/src/Font.cpp
```diff
@@ -1253,9 +1377,238 @@ Pt Font::DetermineLinesImpl(const std::string& text,
...
+            std::cout << "    justification=" << line_data[i].justification << "\n" << std::endl;
+        }
+    }
+}
+
+std::vector<boost::shared_ptr<Font::TextElement> > Font::ComputeTextElements(
+    const std::string& text,
+    const Flags<TextFormat>& format) const
+{
+    std::vector<boost::shared_ptr<TextElement> > text_elements;
```
geoffthemedio, line 1468: can you add a bunch more comments throughout all this code to explain what it's doing?

GG/src/Font.cpp
```diff
@@ -1253,9 +1377,238 @@ Pt Font::DetermineLinesImpl(const std::string& text,
...
+    // fill in the widths of code points in each TextElement
+    FillTemplatedText(text, text_elements, text_elements.begin());
+
+#if DEBUG_DETERMINELINES
+    DebugOutput::PrintParseResults(text_elements);
 #endif
+    return text_elements;
+}
+
+
```
geoffthemedio, line 1573: remove extra blank lines?

GG/src/Font.cpp
```diff
@@ -1253,9 +1377,238 @@ Pt Font::DetermineLinesImpl(const std::string& text,
...
+    if (new_text.empty())
+        return;
+
+    int change_of_len;
+
+    // Find the target text element.
+    size_t curr_offset = 0;
+    std::vector<boost::shared_ptr<Font::TextElement> >::iterator te_it = text_elements.begin();
+    while (te_it != text_elements.end())
+    {
```
geoffthemedio: can move to end of previous line

GG/src/MultiEdit.cpp
```diff
@@ -860,11 +882,11 @@ void MultiEdit::KeyPress(Key key, boost::uint32_t key_code_point, Flags<ModKey>
     case GGK_RIGHT: {
         if (MultiSelected() && !shift_down) {
             m_cursor_begin = m_cursor_end = HighCursorPos();
-        } else if (m_cursor_end.second <
+        } else if (!GetLineData().empty() && m_cursor_end.second <
                     GetLineData()[m_cursor_end.first].char_data.size() -
                     (LineEndsWithEndlineCharacter(GetLineData(), m_cursor_end.first, Text()) ? 1 : 0)) {
```
geoffthemedio: { to newline after multi-line if logic

geoffthemedio (dismissed):
I don't understand what most of this is doing, so can't really meaningfully review that aspect...

LGM-Doyle:

GG/GG/Font.h
adrianbroher: I think that the TextElement is a tree-like structure. When having a formatted string like:

`The <i>quick <rgba 102 51 0 255>brown</rgba></i> fox jumps over the <b>lazy</b> dog.`

it wood result in something like:

object instances

geoffthemedio: OK, but that doesn't address the question about TextElements (with s on the end) and TextElement (without s).

LGM-Doyle: As far a I know the vector of TextElement is always flat.  However, when it is doing pre-rendering it uses a stack to maintain hierarchical information of which tags are active. 

For your example the vector of TextElement, listed from top to bottom, with one TextElement per line, would be:
- TEXT text = "The"
- WHITESPACE text = " " 
- OPEN_TAG text = "<i>", tag_name = "i"
- TEXT text = "quick" 
- WHITESPACE text = " " 
- OPEN_TAG text = "<rgba 102 51 0 255>" tag_name = "rgba",  four params = 102,51,0,255
- TEXT text = brown
- CLOSE_TAG text = </rgba>, tag_name = rgba
- CLOSE_TAG, text = </i> tag_name = i
- TEXT text = "fox" 
- WHITESPACE text = " " 

etc.

I've changed the poorly chosen name of `TextElements` to `TextAndElementsAssembler`, which clarifies that it constructs a matched pair of Text and a vector of TextElements.

geoffthemedio: Does this class do the assembly, or does it provide an interface that other code uses to add elements to the vector of TextElement ?

GG/src/Font.cpp
```diff
@@ -548,6 +697,191 @@ StrSize Font::TextElement::StringSize() const
...
+bool Font::TextElement::operator==(const TextElement &rhs) const
+{
+    return (text == rhs.text && widths == rhs.widths
+            && whitespace == rhs.whitespace && newline == rhs.newline);
+}
+
+///////////////////////////////////////
+// class GG::Font::TextAndElementsAssembler
+///////////////////////////////////////
+class Font::TextAndElementsAssembler::TextAndElementsAssemblerImpl
```
geoffthemedio, line 753: Are there two TextAndElementsAssembler classes defined?

GG/src/Font.cpp
```diff
@@ -548,6 +697,191 @@ StrSize Font::TextElement::StringSize() const
...
+};
+
+
+Font::TextAndElementsAssembler::TextAndElementsAssembler(const Font& font) :
+    pimpl(new TextAndElementsAssemblerImpl(font))
+{}
+// Required because Impl is defined here
+Font::TextAndElementsAssembler::~TextAndElementsAssembler()
+{}
+const std::string& Font::TextAndElementsAssembler::Text()
```
geoffthemedio, line 906: add blank lines between function

LGM-Doyle:
@geoffthemedio thanks for making the time to review this PR.

I though I sent this explanation two days ago, the first time you asked about `Bind()`.  I did not hit send.  It explains why my explanation was not helpful at all.  I've edited the response to use the new `TextAndElementsAssembler` name.

I'll try and answer all of your questions with one clear explanation.

The `Font` XML string to rendered text pipline works as follows: 
- XML string is parsed into a vector of `TextElement`
- The vector of `TextElement` and the original string are used to fill the TEXT type of `TextElement` with glyphs.
- The vector of `TextElement`, the original string and the formatting are used to produce a vector of `LineData`.
- The vector of `LineData`,  the original string and the formatting are used to prerender the text.
- The prerendered text is used to render the text.

90% of the time is consumed in the first step parsing the XML into `TextElement`s.  In spots where the UI was slow due to `Font::DetermineLines()` it amounted to 25% of the computational time of the whole game.

There are three methods to improve this performance bottleneck:
1. Don't recreate the vector of `TextElement` for known strings.  
   
   This is done outside of `Font.cpp`, by not calling `ComputeTextElements` unnecessarily, particularly inside loops.
2. Re-use where possible `TextElements` if they will be the same even though the text has changed.  
   
   Re-using a `TextElement` uses the `Bind()` function as I explain below.
   
   Each `TextElement` contains a `Substring` which is pointer to a string with a start and end index.  `FormattingTag` is a `TextElement` with an additional `Substring`.  Hence all three, `Substring`, `TextElement` and `FormattingTag` have their own `Bind()` functions to change the pointer, in order to re-use an already parsed vector of `TextElement` with a different string.
   
   Another solution would have been to add a `std::string` parameter to each `Substring`, `TextElements` and `FormattingTag` member function, but I decided this was the least amount of change from what was already working.
   
   To answer your question about the risk of mismatch, if a vector of `TextElement` is bound to a string that is too short, eventually it will run into one of the asserts in `Substring` and die.  This is how it always was.  Otherwise, it just displays weirdly rendered text.
   
   I changed `TextControl::SetText(str, text_element)` so that it will not use a `str`, `text_element` pair where `str` is too short and will cause a crash.  This will prevent accidental mismatches. 
   
   The choice to tightly bind the `Substring` to a particular instance of `std::string` predates this PR.
3. Directly construct the vector of `TextElement` and avoid the XML parsing costs.
   
   `TextAndElementsAssembler` allows the vector of `TextElement` to be directly constructed using `AddOpenTag()`, `AddText()` etc.  Then use the fast constructor or fast `SetText` to directly create `TextControl` from the matched pair of string and vector of `TextElement`.
   
   Check PR #992 in [CUIControls.cpp](https://github.com/LGM-Doyle/freeorion/blob/87cd544312af329417cf6608d8fc400bdeb79bc5/UI/CUIControls.cpp#L1339-1375) for an example of the use.

GG/src/Font.cpp
LGM-Doyle: No.  There is `TextAndElementsAssembler` and its impl.

adrianbroher:
I'm unable to compile the code in the current state because the using namespace statement creates ambiguous symbols.

```
./GG/src/Font.cpp: In constructor '{anonymous}::CompiledRegex::CompiledRegex(const boost::unordered::unordered_set<std::__cxx11::basic_string<char> >&, bool)':
./GG/src/Font.cpp:535:38: error: call of overloaded 'ref(const string*&)' is ambiguous
                     [PushP(ref(m_text), ref(m_tag_stack), ref(m_ignore_tags), tag_name_tag)] |
                                      ^
```

I have a fix available but seem to have disabled the [allow edits from maintainers](https://github.com/blog/2247-improving-collaboration-with-forks) feature in this PR.

LGM-Doyle:
Sorry, I was feeling unduly possessive of this code.

Fixed.

adrianbroher:
This is the fifth or sixth time I want to start a review, but I'm loosing any motivation to touch this as soon I switch to the file list and start scrolling. Please break this down into consumable parts.

LGM-Doyle:
I will do this.
It may take a few days due to other commitments, but I am not ignoring this request.

LGM-Doyle:
I have broken this down into a series of PRs from #1029 to #1042.  This PR #989 is still the last in the series.

Each PR is single which either refactors or improve `Font::DetermineLines()` performance.

LGM-Doyle:
This PR #989 is the last PR in the series of PRs.  

It handles the changes to TextControl to add a constructor that uses the TextElement vector directly.

adrianbroher:
> This PR #989 is the last PR in the series of PRs.

Sorry, I didn't the whole comment when closing this PR.

Can you rebase this PR to master?

geoffthemedio:
If next, rebase?

LGM-Doyle:
Rebased.

geoffthemedio:
It is mentioned in the summary and comments, but I can't find any ComputeTextElements() function.

Also, even with this patch applied, I don't see any improvement in text layout performance when resizing the pedia on pages such as the techs list, or the zero point generation article. Continuously resizing the pedia window while showing these pages still noticeably drops the FPS into the single digits. Should there be some improvement by now?

Edit: Nevermind... the slow updating with pedia resizing is fixed by adding a PreRender step.

LGM-Doyle:
ComputeTextElements() was changed to ExpensiveParseFromTextToTextElements() in a review of a previous PR, but not in comments in this PR.  I've fixed that.

All of the changes in these PRs was only focused on improving ListBox.  However, it is gratifying that adding PreRender to other windows works as expected.  I was going to change those after the ListBox changes were accepted.

GG/GG/TextControl.h
```diff
@@ -74,9 +74,42 @@ class GG_API TextControl : public Control
...
+
+    /** Copy constructor.
+
+        Text Control requires a copy-constructor because m_text_elements contains pointers
+        to m_text which need to be bound with Bind() to the m_text in the new TextControl.
+
+        Using the copy constructor is faster than constructing a TextControl with text from another
+        TextControl because it avoids the XML parse overhead.
+
+        Since Control does not have a way to access Flags the copy using default flags.
```
geoffthemedio, line 97: using -> will use

Perhaps you should add a copy constructor to Wnd and Control to copy the flags with then?

PR #990, opened 2016-09-26
Add GG::Wnd::PreRender() to do layout once per frame.

LGM-Doyle:
Adding `Wnd::PreRender()` to the event loop before `Render()` formalizes the
method of consolidating all layout changes once per video frame.

Consolidating change provides performance improvements from a single layout
per frame. It may also fix bugs where multiple UI events are triggered by a
single user action and early changes modify the layout, causing
subsequently processed actions to hit different or removed UI elements
resulting in unexpected behavior.

`PreRender()` is only run on visible windows that have requested a prerender
with `RequirePrender()`.

`PreRender()` processes children first so that parents can check child sizes
while in `PreRender()` and get the correct size, not the size from the last 
rendered frame.

### DIFF
--- a/GG/GG/GUI.h
+++ b/GG/GG/GUI.h
@@ -376,6 +376,7 @@ public:
     //@}
 
     static GUI*  GetGUI();                  ///< allows any GG code access to GUI framework by calling GUI::GetGUI()
+    static void  PreRenderWindow(Wnd* wnd); ///< PreRender \p wnd's children and the \p wnd.
     static void  RenderWindow(Wnd* wnd);    ///< renders a window (if it is visible) and all its visible descendents recursively
     virtual void RenderDragDropWnds();      ///< renders Wnds currently being drag-dropped
 
@@ -417,6 +418,8 @@ protected:
 
     /** \name Mutators */ ///@{
     void           ProcessBrowseInfo();    ///< determines the current browse info mode, if any
+    /** Allow all windows in the z-list to update data before rendering. */
+    virtual void   PreRender();
     virtual void   RenderBegin() = 0;      ///< clears the backbuffer, etc.
     virtual void   Render();               ///< renders the windows in the z-list
     virtual void   RenderEnd() = 0;        ///< swaps buffers, etc.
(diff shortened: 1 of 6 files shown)


geoffthemedio:
Appears to not do much in of itself, but just sets up for other changes...?

GG/GG/Wnd.h
```diff
@@ -583,6 +610,16 @@ class GG_API Wnd : public boost::signals2::trackable
...
 
+    /** Update Wnd prior to Render().
+        PreRender() is called before Render() if RequirePreRender() was called.
+        The default PreRender() resets RequirePreRender.  Wnd::PreRender should
+        be called in any overrides to reset RequirePreRender().
+        PreRender() of child windows will be called before this PreRender(). */
+    virtual void PreRender();
+
+    /** Require that layout be updated before the next Render(). */
+    virtual void        RequirePreRender();
```
geoffthemedio: de-indent function name?

geoffthemedio (changes requested):

LGM-Doyle:
De-indented.
I also moved the RequirePreRender() in FleetWnd.cpp you noticed in PR #991 to the correct commit.

GG/GG/GUI.h
```diff
@@ -376,6 +376,7 @@ class GG_API GUI
     //@}
 
     static GUI*  GetGUI();                  ///< allows any GG code access to GUI framework by calling GUI::GetGUI()
+    static void  PreRenderWindow(Wnd* wnd); ///< PreRender \p wnd's children and the \p wnd.
```
geoffthemedio: explain what PreRenderWindow does, not just that is does itself on a Wnd's children...

GG/GG/Wnd.h
```diff
@@ -194,6 +194,30 @@ extern GG_API const WndFlag NO_WND_FLAGS;
...
+      + events from the GUI and the data sources are handled
+      + PreRender() is called to update layout if required.
+      + Render() is called.
+
+    Wnd does not distinguish between events from the GUI and events from the
+    data sources.
+
+    PreRender() and Render() are only called if the window is visible.
+
+    PreRender() defers all layout changes until after all events have been
```
geoffthemedio, line 210: this is a bit confusing... does PreRender do things inherently, or should various derived classes set themselves up so that their Layout changes in response to (eg.) SizeMove events are only done in PreRender?

LGM-Doyle:

PR #991, opened 2016-09-26
Add GG::DeferredLayout to do layout in PreRender.

LGM-Doyle:
`DeferredLayout` is the same as `Layout` except that the layout
only happens in `PreRender()`.

### DIFF
--- a/GG/GG/DeferredLayout.h
+++ b/GG/GG/DeferredLayout.h
@@ -1,6 +1,6 @@
 // -*- C++ -*-
-/** \file Layout.h \brief Contains the Layout class, which is used to size and
-    align GG windows. */
+/** \file DeferredLayout.h \brief Contains the DefferedLayout class, which is used to size and
+    align GG windows in the PreRender() phase. */
 
 #ifndef _GG_DeferredLayout_h_
 #define _GG_DeferredLayout_h_
@@ -20,7 +20,7 @@ public:
     /** \name Structors */ ///@{
     /** Ctor. */
     DeferredLayout(X x, Y y, X w, Y h, std::size_t rows, std::size_t columns,
-           unsigned int border_margin = 0, unsigned int cell_margin = INVALID_CELL_MARGIN);
+                   unsigned int border_margin = 0, unsigned int cell_margin = INVALID_CELL_MARGIN);
     //@}
 
     /** \name Accessors */ ///@{
@@ -39,9 +39,9 @@ protected:
     //@}
 
 private:
-    Pt                              m_ul_prerender;    ///< ul for PreRender
-    Pt                              m_lr_prerender;    ///< lr for PreRender
-    bool                            m_stop_deferred_resize_recursion;
+    Pt   m_ul_prerender;        ///< ul for PreRender
+    Pt   m_lr_prerender;        ///< lr for PreRender
+    bool m_stop_deferred_resize_recursion;
 };
 
 } // namespace GG
--- a/GG/GG/DeferredLayout.h
+++ b/GG/GG/DeferredLayout.h
@@ -1,0 +1,46 @@
+// -*- C++ -*-
+/** \file Layout.h \brief Contains the Layout class, which is used to size and
+    align GG windows. */
+
+#ifndef _GG_DeferredLayout_h_
+#define _GG_DeferredLayout_h_
+
+#include <GG/Layout.h>
+
+namespace GG {
+
+/** \brief An invisible Wnd subclass that arranges its child Wnds during PreRender.
+
+    A DeferredLayout is a layout that does all of the expensive layout operations once per frame
+    during PreRender().
+*/
+class GG_API DeferredLayout : public Layout
+{
+public:
+    /** \name Structors */ ///@{
+    /** Ctor. */
+    DeferredLayout(X x, Y y, X w, Y h, std::size_t rows, std::size_t columns,
+           unsigned int border_margin = 0, unsigned int cell_margin = INVALID_CELL_MARGIN);
+    //@}
+
+    /** \name Accessors */ ///@{
+    //@}
+
+    /** \name Mutators */ ///@{
+    virtual void SizeMove(const Pt& ul, const Pt& lr);
+    virtual void PreRender();
+    //@}
+
+protected:
+    /** \name Mutators */ ///@{
+    virtual void   RedoLayout();
+    //@}
+
+private:
+    Pt                              m_ul_prerender;    ///< ul for PreRender
+    Pt                              m_lr_prerender;    ///< lr for PreRender
+};
+
+} // namespace GG
+
+#endif
--- a/GG/GG/DeferredLayout.h
+++ b/GG/GG/DeferredLayout.h
@@ -39,8 +39,8 @@ protected:
     //@}
 
 private:
-    Pt   m_ul_prerender;        ///< ul for PreRender
-    Pt   m_lr_prerender;        ///< lr for PreRender
+    Pt   m_ul_prerender;
+    Pt   m_lr_prerender;
     bool m_stop_deferred_resize_recursion;
 };
 
--- a/GG/GG/DeferredLayout.h
+++ b/GG/GG/DeferredLayout.h
@@ -29,16 +29,19 @@ public:
     /** \name Mutators */ ///@{
     virtual void SizeMove(const Pt& ul, const Pt& lr);
     virtual void PreRender();
+    virtual void SetMinSize(const Pt& sz);
+    virtual void SetMaxSize(const Pt& sz);
     //@}
 
 protected:
     /** \name Mutators */ ///@{
-    virtual void   RedoLayout();
+    virtual void RedoLayout();
     //@}
 
 private:
     Pt                              m_ul_prerender;    ///< ul for PreRender
     Pt                              m_lr_prerender;    ///< lr for PreRender
+    bool                            m_stop_deferred_resize_recursion;
 };
 
 } // namespace GG
--- a/GG/GG/Layout.h
+++ b/GG/GG/Layout.h
@@ -244,8 +244,8 @@ protected:
     virtual void KeyRelease(Key key, boost::uint32_t key_code_point, Flags<ModKey> mod_keys);
 
     /** Do all the layout work*/
-    virtual void DoLayout(const Pt& ul, const Pt& lr);
-    virtual void   RedoLayout();
+    virtual void DoLayout(const Pt ul, const Pt lr);
+    virtual void RedoLayout();
     //@}
 
 private:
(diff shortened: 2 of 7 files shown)


GG/GG/DeferredLayout.h
```diff
@@ -0,0 +1,49 @@
...
+    A DeferredLayout is a layout that does all of the expensive layout operations once per frame
+    during PreRender().
+*/
+class GG_API DeferredLayout : public Layout
+{
+public:
+    /** \name Structors */ ///@{
+    /** Ctor. */
+    DeferredLayout(X x, Y y, X w, Y h, std::size_t rows, std::size_t columns,
+           unsigned int border_margin = 0, unsigned int cell_margin = INVALID_CELL_MARGIN);
```
geoffthemedio: indent to X

GG/GG/DeferredLayout.h
```diff
@@ -0,0 +1,49 @@
...
+    virtual void SetMaxSize(const Pt& sz);
+    //@}
+
+protected:
+    /** \name Mutators */ ///@{
+    virtual void RedoLayout();
+    //@}
+
+private:
+    Pt                              m_ul_prerender;    ///< ul for PreRender
```
geoffthemedio: don't need so much indentation to variable names

GG/GG/DeferredLayout.h
```diff
@@ -0,0 +1,49 @@
+// -*- C++ -*-
+/** \file Layout.h \brief Contains the Layout class, which is used to size and
```
geoffthemedio: update comment

GG/GG/Layout.h
```diff
@@ -114,14 +114,14 @@ class GG_API Layout : public Wnd
...
-    unsigned int     CellMargin() const;                       ///< returns the number of pixels the layout leaves between the edges of windows in adjacent cells
-    double           RowStretch(std::size_t row) const;        ///< returns the stretch factor for row \a row.  Note that \a row is not range-checked.
-    double           ColumnStretch(std::size_t column) const;  ///< returns the stretch factor for column \a column.  Note that \a column is not range-checked.
-    Y                MinimumRowHeight(std::size_t row) const;  ///< returns the minimum height allowed for row \a row.  Note that \a row is not range-checked.
-    X                MinimumColumnWidth(std::size_t column) const; ///< returns the minimum height allowed for column \a column.  Note that \a column is not range-checked.
-    std::vector<std::vector<const Wnd*> >
+    virtual Flags<Alignment> ChildAlignment(const Wnd* wnd) const;     ///< returns the aligment of child \a wnd.  \throw GG::Layout::NoSuchChild Throws if no such child exists.
```
geoffthemedio: why would this need to be virtual?

geoffthemedio: or indeed most of these?

geoffthemedio:
You shouldn't make everything virtual by default... just the functions that reasonably might need to be overridden by derived classes.

GG/src/Layout.cpp
```diff
@@ -433,21 +437,37 @@ void Layout::SizeMove(const Pt& ul, const Pt& lr)
...
+        lr -= lr_margin;
 
         if (it->second.alignment == ALIGN_NONE) { // expand to fill available space
             it->first->SizeMove(ul, lr);
         } else { // align as appropriate
             Pt available_space = lr - ul;
             Pt min_usable_size = it->first->MinUsableSize();
             Pt min_size = it->first->MinSize();
+
+            // HACK! This is put here so that TextControl, which is currently GG's
```
geoffthemedio, line 461: what about graphics with propscale enabled?

geoffthemedio: and what about any other user-(of code)-created Wnd type? does having a special case here make sense?

UI/FleetWnd.cpp
```diff
@@ -1220,6 +1220,7 @@ FleetDataPanel::FleetDataPanel(GG::X w, GG::Y h, int system_id, bool new_fleet_d
     m_stat_icons(),
     m_selected(false)
 {
+    RequirePreRender();
```
geoffthemedio: why in the constructor for this class?

GG/src/Layout.cpp
LGM-Doyle: The special case for `TextControl` in the above line 247 predates this PR.

It is necessary here because `TextControl` does not have a well defined minimum size.  This makes sure that the text is re-flows to the width of this column before asking the minimum size.  Otherwise t adds padding to the wrong size of text box.

You are right.  There should probably be a category of Wnd with scalable sizes that Layout can correctly handle.

GG/GG/DeferredLayout.h
```diff
@@ -0,0 +1,49 @@
...
+    virtual void SetMaxSize(const Pt& sz);
+    //@}
+
+protected:
+    /** \name Mutators */ ///@{
+    virtual void RedoLayout();
+    //@}
+
+private:
+    Pt   m_ul_prerender;        ///< ul for PreRender
```
geoffthemedio: if you're going to comment, please do so with some explanation of what the variable is used for... this comment is pretty much useless

GG/GG/Layout.h
```diff
@@ -242,6 +242,10 @@ class GG_API Layout : public Wnd
     virtual void MouseWheel(const Pt& pt, int move, Flags<ModKey> mod_keys);
     virtual void KeyPress(Key key, boost::uint32_t key_code_point, Flags<ModKey> mod_keys);
     virtual void KeyRelease(Key key, boost::uint32_t key_code_point, Flags<ModKey> mod_keys);
+
+    /** Do all the layout work*/
+    virtual void DoLayout(Pt ul, Pt lr);
+    virtual void RedoLayout();
```
geoffthemedio, line 254: what is this for?

GG/GG/Layout.h
```diff
@@ -242,6 +242,10 @@ class GG_API Layout : public Wnd
     virtual void MouseWheel(const Pt& pt, int move, Flags<ModKey> mod_keys);
     virtual void KeyPress(Key key, boost::uint32_t key_code_point, Flags<ModKey> mod_keys);
     virtual void KeyRelease(Key key, boost::uint32_t key_code_point, Flags<ModKey> mod_keys);
+
+    /** Do all the layout work*/
+    virtual void DoLayout(Pt ul, Pt lr);
```
geoffthemedio, line 246: if you're going to comment, explain what doing layout means; don't just rephrase the function name...

GG/GG/Wnd.h
```diff
@@ -583,6 +613,16 @@ class GG_API Wnd : public boost::signals2::trackable
...
 
+    /** Update Wnd prior to Render().
+        PreRender() is called before Render() if RequirePreRender() was called.
+        The default PreRender() resets RequirePreRender.  Wnd::PreRender should
+        be called in any overrides to reset RequirePreRender().
+        PreRender() of child windows will be called before this PreRender(). */
+    virtual void PreRender();
+
+    /** Require that layout be updated before the next Render(). */
+    virtual void        RequirePreRender();
```
geoffthemedio: make indenting consistent

GG/src/DeferredLayout.cpp
```diff
@@ -0,0 +1,82 @@
+/* GG is a GUI for SDL and OpenGL.
+   Copyright (C) 2003-2008 T. Zachary Laine
```
geoffthemedio: I think if you're adding a new file, you shouldn't but Zach's copyright notice on it...

GG/src/DeferredLayout.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+   Zach Laine
+   whatwasthataddress@gmail.com */
+
+#include <GG/DeferredLayout.h>
+
+using namespace GG;
+
+DeferredLayout::DeferredLayout(X x, Y y, X w, Y h, std::size_t rows, std::size_t columns,
+               unsigned int border_margin/* = 0*/, unsigned int cell_margin/* = INVALID_CELL_MARGIN*/) :
```
geoffthemedio: indent to X

GG/src/DeferredLayout.cpp
```diff
@@ -0,0 +1,82 @@
...
+    Layout(x, y, w, h, rows, columns, border_margin, cell_margin),
+    m_ul_prerender(Pt(x, y)),
+    m_lr_prerender(Pt(x + w, y + h)),
+    m_stop_deferred_resize_recursion(false)
+{ }
```
geoffthemedio: space not needed

GG/src/DeferredLayout.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+void DeferredLayout::SizeMove(const Pt& ul, const Pt& lr)
+{
+    if (m_stop_deferred_resize_recursion)
+        return;
+
+    RequirePreRender();
+
+    m_ul_prerender = ul;
```
geoffthemedio, line 43: Layout is derived from Wnd, but you never seem to set the m_upperleft and m_lowerright members... Intentional? Will this break things if a layout is assumed by a parent class to have these properties be valid?

geoffthemedio (changes requested):

LGM-Doyle:

geoffthemedio:
rebase?

LGM-Doyle:
rebased.  I fixed the spacing and added comments in a way git couldn't trivially merge.

GG/GG/DeferredLayout.h
```diff
@@ -0,0 +1,49 @@
...
+    /** \name Mutators */ ///@{
+    virtual void SizeMove(const Pt& ul, const Pt& lr);
+    virtual void PreRender();
+    virtual void SetMinSize(const Pt& sz);
+    virtual void SetMaxSize(const Pt& sz);
+    //@}
+
+protected:
+    /** \name Mutators */ ///@{
+    virtual void RedoLayout();
```
geoffthemedio, line 38: Base class RequirePreRender() is already public, and this just calls RequirePreRender(), so what is the purpose of it, especially if protected?

LGM-Doyle: I may not be understanding your question, so if the answer misses your point I apologize.

Three facts are significant:
1. `Layout::RedoLayout()` and `DeferredLayout::RedoLayout()` do different things.
   
   `Layout::RedoLayout()` immediately causes a re-layout to happen if the size has changed.
   
   `DeferredLayout::RedoLayout()` calls `RequirePreRender()` which will cause a `DoLayout()` within `PreRender()`.
2. `Layout::RedoLayout()` is a virtual function.
   
   Since `RedoLayout()` is a virtual function every place in `Layout::SomeMutatorFunction()` that calls `RedoLayout()` will call `DeferredLayout::RedoLayout()` if the `Layout` is really a `DeferredLayout`.  This makes the changes in all of the mutator function in `Layout` cause deferred changes if the `Layout` is really a `DeferredLayout`.
3. `Layout::RedoLayout()` is protected because only `Layout` or its derived classes will know when a change should trigger a `RedoLayout()`.  Of course, anyone can call the public `RequirePreRender()`, but that doesn't do anything for `Layout`.

PR #992, opened 2016-09-26
Use Wnd::PreRender and new Font functions in various files.

LGM-Doyle:
This PR contain commits that require  `Wnd::PreRender()` or the
`Font::DetermineLines()` improvements, but are not `ListBox` related.

They are all elements that showed up as performance bottlenecks while
profiling ListBox derived classes.

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -960,30 +960,31 @@ MapWnd::MapWnd() :
     // resources
     const GG::X ICON_DUAL_WIDTH(100);
     const GG::X ICON_WIDTH(24);
-    m_population = new StatisticIcon(ClientUI::MeterIcon(METER_POPULATION), 0, 3, false);
-    m_population->Resize(GG::Pt(ICON_DUAL_WIDTH, m_btn_turn->Height()));
+    m_population = new StatisticIcon(ClientUI::MeterIcon(METER_POPULATION), 0, 3, false,
+                                     GG::X0, GG::Y0, ICON_DUAL_WIDTH, m_btn_turn->Height());
     m_population->SetName("Population StatisticIcon");
 
-    m_industry = new StatisticIcon(ClientUI::MeterIcon(METER_INDUSTRY), 0, 3, false);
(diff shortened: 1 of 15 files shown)


adrianbroher:
Requires #989 and rebase.

geoffthemedio:
Rebase?

LGM-Doyle:
This will be the next PR rebased, but it won't be ready today.

I've been refactoring and testing the SidePanel::PreRender() code.  You added SidePanel::PreRender() to manage the timing of the DoLayout() to speed up SizeMove(), which was good.  This PR added SidePanel::PreRender() to manage the timing of SidePanel::Update() which was slowing down the FleetWnd layout when the SidePanel was open.  While I was merging those two PreRender() functions, I noticed that SidePanel::Refresh() could also cause the same problem.  I'm still testing.

LGM-Doyle:
Rebased and this is next in line.

UI/MapWnd.cpp
```diff
@@ -1400,20 +1401,31 @@ ModeratorActionSetting MapWnd::GetModeratorActionSetting() const
 bool MapWnd::AutoEndTurnEnabled() const
 { return m_auto_end_turn; }
 
+void MapWnd::PreRender() {
+    GG::Wnd::PreRender();
+    // Save CPU / GPU activity by skipping rendering when it's not needed
+    // As of this writing, the design and research screens have fully opaque backgrounds.
```
dbenage-cx, line 1407: Only by default, a user may reduce the alpha value in `UI.ctrl-color` or `UI.wnd-color` (both supported in GUI options)

dbenage-cx: Apologies, I missed that this was an existing comment.  Without changes impacting such behavior, any existing related issues fall outside of the scope of this PR.

Just to note:

> How would you suggest/expect this to work?

If it can be viewed at all (UI layout changes, alpha adjustment), the MapWnd should be updated.  If updates are prevented due to obstruction, preferably hide the window (I suspect this might be impractical, possibly render a new non-configurable opaque window on top)

PR #1000, opened 2016-09-28
Added exact effect descriptions to the ion storm and molecular cloud …

Deepsloth:
…fields.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -956,7 +956,7 @@ FLD_ION_STORM
 Ion Storm
 
 FLD_ION_STORM_DESC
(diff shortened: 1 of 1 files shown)


default/stringtables/en.txt
```diff
@@ -956,13 +956,13 @@ FLD_ION_STORM
 Ion Storm
 
 FLD_ION_STORM_DESC
-Magnetic vortex of relativistic charged particles that can interfere with sensors and obscures all objects within it.
+Magnetic vortex of relativistic charged particles that can interfere with sensors and obscures all objects within it. All objects inside the storm receive an increase of [[encyclopedia STEALTH_TITLE]] by 40, but a decrease of [[encyclopedia DETECTION_RANGE_TITLE]] by 40.
```
MatGB: either "stealth increased by" or "an increase to stealth of", same for detection. Beyond that, good call, we need to get more precise descriptions into the stringtables—ideally using macros from the effects themselves in the case of stuff that can change.

MatGB:
Woohoo! Is #1000 something to celebrate? ;-)

Vezzra:
> Woohoo! Is #1000 something to celebrate? ;-)

Feel free to throw a party... 😉

PR #1009, opened 2016-10-01
Ancient guardians

Deepsloth:
Successor of #837.


default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
```diff
@@ -0,0 +1,90 @@
...
+    description = "SP_ANCIENT_GUARDIANS_DESC"
+    gameplay_description = "SP_ANCIENT_GUARDIANS_GAMEPLAY_DESC"
+
+    tags = [ "ROBOTIC" "NO_INDUSTRY" "NO_RESEARCH" "AVERAGE_SUPPLY" "DESTROYED_ON_CONQUEST"]
+
+    foci = [
+        [[HAS_INDUSTRY_FOCUS]]
+        [[HAS_RESEARCH_FOCUS]]
+        [[HAS_GROWTH_FOCUS]]
+        [[HAS_ADVANCED_FOCI]]
```
dbenage-cx, line 12: Morlic had an [unanswered question](https://github.com/freeorion/freeorion/pull/837#discussion_r81445479) here regarding foci.
At least one is likely required, a populated planet with no available focus could lead to issues.

Any consideration for using `FOCUS_PROTECTION`, and setting it as the preferred focus?
(removing all other foci, and adjusting values as you see fit)

default/scripting/specials/specials.macros
```diff
@@ -13,7 +13,7 @@ CHANCE_OF_GUARD_1
             scope = Source
             activation = AND [
                 Turn high = 0
-                Random probability = 0.60
+                Random probability = 0.70
```
dbenage-cx, line 16: Take `GalaxyMonsteFrequency` into account here?
Something akin to: `Random probability = 0.6 + (GalaxyMonsterFrequency * 0.1)`

dbenage-cx (commented):
Untested, curious if there are any issues with World Tree.

default/scripting/specials/specials.macros
Deepsloth: As discussed with Matt, this will likely change to 100% when the AI team is ready for lots of ancient guardians. Morlic wanted the chances to stay as they are for now.

default/scripting/species/SP_ANCIENT_GUARDIANS.focs.txt
MatGB: If we use Protection that would make complete sense but would mean if we change what protection does down the line we have to reasses the numbers here. Having the focuses there makes little sense as they should, in theory, never be used, but removing them isn't a priority I think.

When we do more work on them setting it to protection only and fiddling with the numbers can work, but my test games reveal the AI team need to make tweaks before we make them any more annoying—my neighbour in my current game has over 100 troops in orbit around one planet but no warship in sight.

MatGB:
Yup, works as intended for now, the AI definitely needs updating but this concept is definitely something we can expand on, I like a lot.

PR #1011, opened 2016-10-01
Organize AIDependencies.py

Morlic-fo:
- Reorganize the way information is sorted
- Add code-folding tags for easier IDE navigation


Cjkjvfnby:
@Morlic-fo, Is it your idea to use that folding style? Should it be discussed with team?  

PS. If you add some new feature like code folding it will be nice to add some link about it.

https://www.jetbrains.com/help/idea/2016.2/code-folding.html#using_folding_comments

Morlic-fo:
I will make sure to add links in the future. Sorry for that.

Yeah, to use the folding style is my idea as the previous form of AIDependencies.py was very hard to navigate through. From my perspective, it most certainly is better than those `If true:  # only for code folding` statements in other parts of the code. It is comment-base and thus easily ignored even if your IDE does not support it. The style itself is very specific in what it does.

> Should it be discussed with team?
> Well, effectively we two are the AI team right now and I assigned the PR to you so you could voice your opinion and any concerns. As you merged the PR, I guess you are ok with it - if you have better ideas for the code organization in that file or just think that that particular style is a bad idea for whatever reason, please say so.

Or do you think this should be discussed with all the devs / project lead? It does not seem to be a critical / fundamental design decision to me - but if anyone feels the need, of course I am open to discuss it.

Vezzra:
As this addition is entirely in your "realm" (so to speak) and a rather minor style thing I guess it's not absolutely necessary to involve the other devs/leads in that decision.

If we feel left out, we'll complain. 😉

PR #1015, opened 2016-10-02
Improved the Environment pedia article by adding the ep_wheel picture.

Deepsloth:


MatGB:
Yup, works fine and we got an image embedded via script, which is also good.

PR #829, opened 2016-08-02
Added the native species Lembala'Lam.

Deepsloth:
Cleaned, fixed, tweaked and with comments.

PR #1016, opened 2016-10-02
Added the native species Lembala'Lam

Deepsloth:
Successor of #829.

### DIFF
--- a/default/scripting/species/SP_LEMBALALAM.focs.txt
+++ b/default/scripting/species/SP_LEMBALALAM.focs.txt
@@ -1,0 +1,129 @@
+Species
+    name = "SP_LEMBALALAM"
+    description = "SP_LEMBALALAM_DESC"
+    gameplay_description = "SP_LEMBALALAM_GAMEPLAY_DESC"
+    Native
+
+    tags = [ "ORGANIC" "AI_TAG_NO_INDUSTRY" "AI_TAG_GOOD_RESEARCH" "AI_TAG_AVERAGE_SUPPLY" "AI_TAG_GOOD_STEALTH"]
+
+    foci = [
+        /*[[HAS_INDUSTRY_FOCUS]]*/
+        [[HAS_RESEARCH_FOCUS]]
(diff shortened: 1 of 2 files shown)


Deepsloth:
Conflicts resolved. 
AI_Tag, _Desc and label updates have been done.

MatGB:
OK, after about 4 attempts to test these guys, I've either not found them, had them the other side of the galaxy or been near the end of the tech tree when I found them.

Unless someone objects, I'm going to merge the PR in tomorrow, it runs fine with no errors and I think it's best to fix problems with them if/when we get feedback reports.

Vezzra:
Ok with me.

MatGB:
For future, you've mentioned in passing in a thread that it would be fairly easy to hack the Python code to guarantee a specific native near homeworlds: I'd love to know how because testing natives really is a pain when you've no guarantee of even getting one in a galaxy let alone it being near enough to you to test viably.

Morlic-fo:
If you do merge, please open up the standard AI issue. The required adaptations are minor and not too important (I can only think of tweaking the invasion priority right now) but it should be done at some point anyway.

Vezzra:
@MatGB, the easiest way to test a native species is probably to tweak the relevant universe generation scripts to make it so that all planets with natives get the species you want to test (and then set Native Frequency to High, obviously). To do that, change line 87 in `natives.py`:

```
natives = random.choice(natives_for_planet_type[planet_type])
```

to:

```
natives = "SP_WHATEVER_SPECIES"
```

and add the following line after line 90 (`fo.planet_set_species(candidate, natives)`):

```
fo.planet_set_type(candidate, fo.planetType.preferred_planet_type_of_whatever_species)
```

...to make sure our natives don't die out if the planet type doesn't match their preference. See `planets.py` for the list of values for the planet types.

Didn't test this, but it should work.

MatGB:
Cool, I can always tweak it if it doesn't work, I'm almost able to read simple Python now but the parsing always confuses me. Thanks.

PR #1018, opened 2016-10-03
Add right click context to hide/show SitRep templates

### LINKED ISSUE #889: Add "Hide X Sitreps" to sitrep entry right click menu
One can right click on a sitrep entry to copy its text, and one can use the filters button / list to hide / show various types of sitrep, and one can right click on the sitrep icon (not the main text area) to hide/dissmiss (not sure what the difference is) for Y turns. It would be convenient to be able to hide a class of sitrep from a right-click menu as well, which would work like the filters in the button, but not require the player to know what type / class of sitrep to pick to hide.

dbenage-cx:
Adds right click context for each SitRep row for:
- hiding SitReps of the same template type
- showing all hidden SitReps when at least one is hidden

Resolves #889

### DIFF
--- a/UI/SitRepPanel.cpp
+++ b/UI/SitRepPanel.cpp
@@ -599,36 +599,49 @@ void SitRepPanel::IgnoreSitRep(GG::ListBox::iterator it, const GG::Pt& pt, const
 
 void SitRepPanel::DismissalMenu(GG::ListBox::iterator it, const GG::Pt& pt, const GG::Flags<GG::ModKey>& mod) {
 
-    GG::MenuItem menu_contents;
+    GG::MenuItem menu_contents, submenu_ignore, submenu_block, separator_item;
     std::string sitrep_text, sitrep_template;
+    separator_item.separator = true;
     int start_turn = 0;
     SitRepRow* sitrep_row(0);
     if (it != m_sitreps_lb->end()) 
         sitrep_row = dynamic_cast<SitRepRow*>(*it);
+    submenu_ignore.label = UserString("SITREP_IGNORE_MENU");
     if (sitrep_row) {
         const SitRepEntry& sitrep_entry = sitrep_row->GetSitRepEntry();
         sitrep_text = sitrep_entry.GetText();
         start_turn = sitrep_entry.GetTurn();
         if (start_turn > 0) {
-            menu_contents.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_5_TURNS"),      1, false, false));
-            menu_contents.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_10_TURNS"),     2, false, false));
-            menu_contents.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_INDEFINITE"),   3, false, false));
+            submenu_ignore.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_5_TURNS"),      1, false, false));
+            submenu_ignore.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_10_TURNS"),     2, false, false));
+            submenu_ignore.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_INDEFINITE"),   3, false, false));
+            submenu_ignore.next_level.push_back(separator_item);
         }
+    }
+    submenu_ignore.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_CLEAR_ALL"),            4, false, false));
+    submenu_ignore.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_CLEAR_INDEFINITE"),     5, false, false));
+    menu_contents.next_level.push_back(submenu_ignore);
+
+    submenu_block.label = UserString("SITREP_BLOCK_MENU");
+    if (sitrep_row) {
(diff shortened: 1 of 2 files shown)


UI/SitRepPanel.cpp
```diff
@@ -574,22 +574,33 @@ void SitRepPanel::DismissSitRep(GG::ListBox::iterator it, const GG::Pt& pt, cons
...
+        const SitRepEntry& sitrep_entry = sitrep_row->GetSitRepEntry();
+        sitrep_text = sitrep_entry.GetText();
+        start_turn = sitrep_entry.GetTurn();
         if (start_turn > 0) {
             menu_contents.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_5_TURNS"),      1, false, false));
             menu_contents.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_10_TURNS"),     2, false, false));
             menu_contents.next_level.push_back(GG::MenuItem(UserString("SITREP_SNOOZE_INDEFINITE"),   3, false, false));
         }
+        sitrep_template = sitrep_entry.GetLabelString();
+        std::string sitrep_label = sitrep_entry.GetStringtableLookupFlag()? UserString(sitrep_template) : sitrep_template;
```
geoffthemedio: space between ) and ?

default/stringtables/en.txt
```diff
@@ -5032,7 +5032,14 @@ SITREP_SNOOZE_CLEAR_INDEFINITE
 '''  Show all SitReps indefinitely dismissed'''
 
 SITREP_SNOOZE_CLEAR_ALL
-'''  Show all hidden or dismissed SitReps'''
+'''  Show all dismissed SitReps'''
+
+# %1% localized name of SitRep template
+SITREP_HIDE_TEMPLATE
+'''  Hide all SitReps with template: %1%'''
```
geoffthemedio: avoid use of "template" in user-visible text... meaning is unclear.

MatGB: Agreed: we use the term 'label' in the scripts and having been thinking about it for a few days I haven't come up with a more appropriate term.

geoffthemedio: Is a term needed? Why not just say "Hide all %1% sitreps"?

UI/SitRepPanel.cpp
dbenage-cx: Fixed

default/stringtables/en.txt
dbenage-cx: Changed to `SitReps of type`, to be in-line with #1019.
I was unaware of the use of `label`, will search for it.

dbenage-cx: @geoffthemedio Sounds good, changed.

default/stringtables/en.txt
```diff
@@ -5032,7 +5032,14 @@ SITREP_SNOOZE_CLEAR_INDEFINITE
...
 SITREP_SNOOZE_CLEAR_ALL
-'''  Show all hidden or dismissed SitReps'''
+'''  Show all dismissed SitReps'''
+
+# %1% localized name of SitRep template
+SITREP_HIDE_TEMPLATE
+'''  Hide all %1% SitReps'''
+
+SITREP_SHOWALL_TEMPLATES
+'''  Show all hidden SitRep templates'''
```
geoffthemedio: still using "template" here

dbenage-cx: Fixed.

default/stringtables/en.txt
```diff
@@ -5032,7 +5032,14 @@ SITREP_SNOOZE_CLEAR_INDEFINITE
 '''  Show all SitReps indefinitely dismissed'''
```
geoffthemedio: make word order consistent... "SitReps" to end, consistent with following strings

dbenage-cx: Fixed

geoffthemedio:
The difference between hiding and dismissing sitreps needs to be clarified.

Can you make the right-click menu work for the whole sitrep, not just the icon part?

dbenage-cx:
Looking into adding that functionality in coming days (the SitRepPanel contains the menu, its just overlapped by SitRepRow/SitRepDataPanel).

default/stringtables/en.txt
dbenage-cx: If it is ok, I'd like to address clarification within #1019, so one PR is not dependent on the other.
I've reverted this fix here, but am happy to reapply if needed or the other is rejected.

dbenage-cx:
@geoffthemedio opened PR #1053 for menu selection

dbenage-cx:
Rebased.
Since the length of this popup has expanded, I've moved ignore/block entries into submenus.
The prefix "margins" included in the stringtable entries were removed and are now added after translation.

geoffthemedio:
Builds OK, but the help lookup line in the popup is very awkward / I don't like it.
popup

dbenage-cx:
Suggestions?
`Pedia - Ignore/Block`
`Help: Hiding SitReps`
indentation?
Place below `Copy`?

MatGB:
I think I prefer Help, and think at the bottom is a good thing: it probably would be good to standardise Pedia lookups to be at the bottom of popups as much as possible/practical.

But I don't have a strong opinion.

dbenage-cx:
Will try that, I left the title alone for now.

geoffthemedio:
I'd change "completely" to "permanently", though...

PR #1019, opened 2016-10-03
Add help article for SitRep Dismiss/Hide

dbenage-cx:
There is an understandable confusion between Dismissing a SitRep and Hiding a SitRep template (Filters).

This PR adds an article to explain those differences, and a right click context entry to make the explanation more accessible to the user.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -4711,6 +4711,36 @@ Important sections include [[encyclopedia CATEGORY_GAME_CONCEPTS]] and the recen
 
 '''
 
+SITREP_IGNORE_BLOCK_TITLE
+Ignoring and Blocking SitReps
+
+SITREP_IGNORE_BLOCK_SHORT_DESC
+Ignoring a specific SitRep and blocking SitRep types.
+
+SITREP_IGNORE_BLOCK_TEXT
+'''There are two options for hiding Situation Reports (SitReps): Ignoring and Blocking.
+
+Ignoring a SitRep is always for a specific event.
+For example, a specific ship arriving at a specific system.
+A different ship arriving at that system, or the same ship arriving at a different system are both different events, and will still be displayed.
+
+Blocking a SitRep <i>type</i> applies to any SitRep for the same type of event.
+For example, <u>[[SITREP_OWN_SHIP_ARRIVED_AT_DESTINATION_LABEL]]</u> applies to any ship owned by the players empire arriving at any system.
+
+<u>[[SITREP_SNOOZE_INDEFINITE]]</u> will hide the SitRep for all turns, starting from turn 0.
+Options to ignore for 5 or 10 turns will start counting from the turn being viewed in the Situation Report window.
+Ignoring a SitRep only applies to the current game session.
+If the client is restarted or a game is loaded, these SitReps are displayed again.
+
+Blocking a SitRep type applies to all turns and will persist between game sessions.
+Previously blocked SitRep types may be re-enabled by clicking the Filters button.
+SitRep types will have a check mark next to them if they are enabled/shown.
+SitRep types only appear in the Filters list once an event has occurred that would trigger them.
+
+Ignoring is useful for hiding a reoccurring event, such as reminders to build a [[buildingtype BLD_GAS_GIANT_GEN]] at one planet.
+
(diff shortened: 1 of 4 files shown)


dbenage-cx:
documentation seemed the closest match, although it is user facing

MatGB:
I use documentation when I'm doing Pedia based stuff, but then I'm never sure what to use or why.

default/stringtables/en.txt
```diff
@@ -4705,6 +4705,29 @@ Important sections include [[encyclopedia CATEGORY_GAME_CONCEPTS]] and the recen
 
 '''
 
+SITREP_DISMISS_HIDE_TITLE
+Dismissing and Hiding SitReps
+
+SITREP_DISMISS_HIDE_SHORT_DESC
+Dismissing a specific SitRep and hiding a SitRep template.
```
MatGB: I dislike the use of the word template, I don't think it's appropriate, but I'm not sure what would be better. Sitrep type?

dbenage-cx: Thanks, I kept trying to use 'group' as a replacement.  Type seems a better fit to a user, though slightly less accurate.

default/stringtables/en.txt
```diff
@@ -4705,6 +4705,29 @@ Important sections include [[encyclopedia CATEGORY_GAME_CONCEPTS]] and the recen
...
+SITREP_DISMISS_HIDE_TITLE
+Dismissing and Hiding SitReps
+
+SITREP_DISMISS_HIDE_SHORT_DESC
+Dismissing a specific SitRep and hiding SitRep types.
+
+SITREP_DISMISS_HIDE_TEXT
+'''There are two options for ignoring Situation Reports (SitReps): Dismissing and Hiding.
+
+Dismissing a SitRep will keep that specific SitRep for displaying.
```
geoffthemedio: This does not clarify anything for me. A sitrep is generated on and describes an event on a specific turn... how does one "dismiss" a "specific sitrep" for several turns? Does that mean if I hide it on turn 10, then on turn 13 I go back in the sitrep window to look at turn 10, a single entry will be hidden? I don't see why anyone would want to do that just for a few turns, so I assume that's not what it means. So then I suppose something else is meant by "specific sitrep", such as a particular type of message, like "enemy fleet arrived"... but then what is a "group of sitreps" if not a particular type of message?

The only difference I understand is per-game or persistent in client config hiding of message types.

default/stringtables/en.txt
```diff
@@ -4705,6 +4705,29 @@ Important sections include [[encyclopedia CATEGORY_GAME_CONCEPTS]] and the recen
...
+There are options to dismiss for a certain number of turns, or indefinitely.
+When dismissing a SitRep, it will only stay dismissed for the current game.
+If you restart the client or reload the current game, these SitReps will display again.
+Dismissing is always for a specific SitRep, such as a reminder to build a Gas Generator on planet Delphi II.
```
geoffthemedio: So if a fleet arrival sitrep for a particular fleet at a particular system, and the fleet arrives at that system again a few turns later, there'd be no sitrep? Maybe that starts to make sense, but I think you need to explicitly explain "that specific sitrep" immediately. Does it mean a sitrep with the exact same message?

How does the turns / timing limitation work? Turn on which the sitrep was generated, or the current game turn (eg. so if I hide a specific sitrep on turn 20, then on turn 21 I go back in the sitrep window to look at sitreps from turn 10, will the specific sitrep appear then if it was generated on turn 10?)

default/stringtables/en.txt
dbenage-cx: Hoping the new wording will clarify this.

default/stringtables/en.txt
dbenage-cx: Yes, meaning the exact same message.

In your scenario, if the sitrep was dismissed for X turns, it will still show on turn 10.
If it was Indefinitely Dismissed, it would hide that sitrep for all turns (changed to Completely Ignore, as indefinite is misleading).

dbenage-cx:
Rebased.

To help distinguish (since both are "hidden"), I've relabeled `dismiss` to `ignore`, and `hide` to `block`.
Ignoring always applies to a specific sitrep message (Discovered NewTechX).
Blocking always applies to filtering by template (all new tech completed messages).
Suggestions welcome if either of these terms are not suitable.

PR #1021, opened 2016-10-03
left-click on map to close fleet wnd

TheSilentOne1:
Left-clicking the map screen will close the fleet window if it is open. If no fleet window is open, it will close the sidepanel. (If you play with multiple fleet window enabled, all fleet windows will close.)

### DIFF
--- a/UI/ClientUI.cpp
+++ b/UI/ClientUI.cpp
@@ -531,7 +531,7 @@ namespace {
         // UI behavior
         db.Add("UI.tooltip-delay",              UserStringNop("OPTIONS_DB_UI_TOOLTIP_DELAY"),              500,        RangedValidator<int>(0, 3000));
         db.Add("UI.multiple-fleet-windows",     UserStringNop("OPTIONS_DB_UI_MULTIPLE_FLEET_WINDOWS"),     false);
-        db.Add("UI.window-quickclose",          UserStringNop("OPTIONS_DB_UI_WINDOW_QUICKCLOSE"),          false);
+        db.Add("UI.window-quickclose",          UserStringNop("OPTIONS_DB_UI_WINDOW_QUICKCLOSE"),          true);
         db.Add("UI.auto-reposition-windows",    UserStringNop("OPTIONS_DB_UI_AUTO_REPOSITION_WINDOWS"),    true);
 
         // UI behavior, hidden options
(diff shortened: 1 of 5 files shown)


dbenage-cx:
Is deselecting fleet(s) while viewing the production screen intended?
I imagine a user could expect either behavior.

geoffthemedio:
There is already an option in the UI tab: Quick close windows. If enabled, right clicking on the map away from a system will close the fleets window.

TheSilentOne1:
@dbenage-cx: Good point, I guess I would rather prevent deselecting fleets when the production screen is open.

@geoffthemedio: And one more thing I didn't know that it existed! But if even I as a FO veteran didn't know about this option, this small left-click patch may still be an improvement over the right-click method. Let's maybe hear some more opinions? 

... I just noticed there is also some issue where the fleet window resets its size after it has been closed, I will ofc adress that if the left-click-close feature finds approval.

TheSilentOne1:
Have adressed the aforementioned issue.

Vezzra:
> But if even I as a FO veteran didn't know about this option, this small left-click patch may still be an improvement over the right-click method. Let's maybe hear some more opinions? 

Using left-click to first close open fleet windows and then the planet sidepanel (and, if we keep the "Quick close window" option in the options dialog, default it to "on") sounds like a good idea to me, more intuitive than the current right click solution.

Any objections to changing that as @TheSilentOne1 suggested?

TheSilentOne1:
I have noticed another bug: when you drag the map screen while the mouse is over a nebula (field icon), that will cause the fleet screen to close. This seems to happen because at the end of dragging, instead of FieldIcon::LButtonUp() FieldIcon::LClick() is called. I don't have a clue why this is happening - help would be appreciated!

dbenage-cx:
GG::Wnd:LClick

> ```
> /** Respond to release of left mouse button over this Wnd, if it was also
>     originally depressed over this Wnd.  A Wnd will receive an LButtonUp()
>     message whenever a drag that started over its area ends over its area
>     as well. */
> ```

Checking `m_dragged` here should resolve the issue over nebulas (and other potential cases), something like:

```
    if (!m_dragged && !m_in_production_view_mode) {
        FleetUIManager& manager = FleetUIManager::GetFleetUIManager();
        if (manager.ActiveFleetWnd()) {
            manager.CloseAll();
        } else {
            SelectSystem(INVALID_OBJECT_ID);
            m_side_panel->Hide();
        }
    }
```

TheSilentOne1:
Checking for m_dragged did it. Thank you!

TheSilentOne1:
As far as I'm concerned, this is good to go (in).

Vezzra:
As there haven't been any objections to the proposed changes, I think you can go ahead and merge.

PR #1023, opened 2016-10-05
Include dynamic pedia categories in pedia search

### LINKED ISSUE #1022: Pedia Search Misses Articles
Type "Research" into the pedia search bar. One article shows up, which is not the "Research" article in the Game Concepts section. It should. This is probably not an isolated example.

dbenage-cx:
Adds categories, as reported by articles, to search results.

Fixes #1022


geoffthemedio:
untested, but looks plausible

Issue #755, opened 2016-07-05
Add Sidebar or other Layout Reworking to Pedia


geoffthemedio:
Improve 'Pedia navigation with a sidebar.

http://www.freeorion.org/forum/viewtopic.php?f=10&t=6597&hilit=%27pedia+GUI

PR #1024, opened 2016-10-05
Match article text in pedia searches

### LINKED ISSUE #754: Improve Pedia Search
The encyclopedia search function could be improved. Highest priority is showing a list of matches or partial matches for a search query, rather than jumping directly to the first match found. Article contents, not just titles, should also be searched.

dbenage-cx:
Matches pedia article text for search results.

Resolves #754

### DIFF
--- a/UI/EncyclopediaDetailPanel.cpp
+++ b/UI/EncyclopediaDetailPanel.cpp
@@ -52,8 +52,10 @@ namespace {
     const GG::Y TEXT_MARGIN_Y(0);
     const int DESCRIPTION_PADDING(3);
 
-    void    AddOptions(OptionsDB& db)
-    { db.Add("UI.dump-effects-descriptions", UserStringNop("OPTIONS_DB_DUMP_EFFECTS_GROUPS_DESC"),  false,  Validator<bool>()); }
+    void    AddOptions(OptionsDB& db) {
+        db.Add("UI.dump-effects-descriptions", UserStringNop("OPTIONS_DB_DUMP_EFFECTS_GROUPS_DESC"),  false,  Validator<bool>());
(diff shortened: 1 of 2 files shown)


default/stringtables/en.txt
```diff
@@ -3891,6 +3891,9 @@ Word Title Matches
 ENC_SEARCH_PARTIAL_MATCHES
 Partial-Word Title Matches
 
+ENC_SEARCH_ARTICLE_MATCHES
+Article Description Matches
```
geoffthemedio: "Article Description" is misleading... I think you mean article contents or text?

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -987,6 +1006,25 @@ void EncyclopediaDetailPanel::HandleSearchTextEntered() {
...
+         entry_it = all_pedia_entries_list.begin();
+         entry_it != all_pedia_entries_list.end(); ++entry_it)
+    {
+        std::set<std::pair<std::string, std::string> >::const_iterator dupe_it =
+            already_listed_results.find(std::make_pair(entry_it->second.second, entry_it->first));
+        if (dupe_it != already_listed_results.end())
+            continue;
+
+        EncyclopediaArticle article_entry = GetPediaArticle(entry_it->second.second);
+        if (boost::icontains(UserString(article_entry.description), search_text)) {
```
geoffthemedio: I'm concerned this could be rather slow to search through all article text...

default/stringtables/en.txt
dbenage-cx: Fixed

UI/EncyclopediaDetailPanel.cpp
dbenage-cx: Added config option.  Instead of adding another selection in the options screen, I'd like to defer the UI counterpart for #755

geoffthemedio:
seems reasonable...

dbenage-cx:
TY for reviewing

PR #1029, opened 2016-10-07
Refactor Font::DetermineLines.

LGM-Doyle:
This PR refactors Font::DetermineLines.
It creates FillTextElements to do the intensive processing of DetermineLines.
It moves all of the boost::xpressive regex creation into CompiledRegex, to reduce repetition in the code.

### DIFF
--- a/GG/src/Font.cpp
+++ b/GG/src/Font.cpp
@@ -700,6 +700,104 @@ Font::Glyph::Glyph(const boost::shared_ptr<Texture>& texture, const Pt& ul, cons
 {}
 
 namespace {
+    using namespace boost::xpressive;
+
+    /** CompiledRegex maintains a compiled boost::xpressive regular
+        expression that includes a tag stack which can be cleared and
+        provided to callers without the overhead of recompiling the
+        regular expression.*/
+    class CompiledRegex {
+        public:
+        CompiledRegex(const std::set<std::string>& known_tags, bool strip_unpaired_tags) :
+            m_text(0), m_known_tags(&known_tags), m_ignore_tags(false), m_tag_stack(),
+            m_EVERYTHING() {
+
+            // Synonyms for s1 thru s5 sub matches
+            mark_tag tag_name_tag(1);
+            mark_tag open_bracket_tag(2);
+            mark_tag close_bracket_tag(3);
+            mark_tag whitespace_tag(4);
+            mark_tag text_tag(5);
+
+            //   [<]+?   -+ 'non-greedy'   ~not   set[|] 'set', _s space
+            // anythin but space or <
+            const sregex TAG_PARAM =
+                -+~set[_s | '<'];
+
+            //+_w one or more greed word chars, () group no capture, [] semantic
+            const sregex OPEN_TAG_NAME =
+                (+_w)[check(boost::bind(&CompiledRegex::MatchesKnownTag, this, _1))];
+            // (+_w) one or more greedy word check matches stack
+            const sregex CLOSE_TAG_NAME =
+                (+_w)[check(boost::bind(&CompiledRegex::MatchesTopOfStack, this, _1))];
+            // *blank zero or more greedy whitespace, >> followed by , _ln newline,
+            // (set = 'a', 'b') [ab], +blank one or more greedy blank
+            const sregex WHITESPACE =
+                (*blank >> (_ln | (set = '\n', '\r', '\f'))) | +blank;
+
+            // < followed by not space or <   or one or more not space or <
+            const sregex TEXT =
+                ('<' >> *~set[_s | '<']) | (+~set[_s | '<']);
+
+            if (!strip_unpaired_tags) {
+                m_EVERYTHING =
+                    ('<' >> (tag_name_tag = OPEN_TAG_NAME) // < followed by TAG_NAME
+                     >> repeat<0, 9>(+blank >> TAG_PARAM)  //repeat 0 to 9  single blank followed
+                                                           //by TAG_PARAM
+                     >> (open_bracket_tag.proto_base() = '>'))  //s1. close tag and push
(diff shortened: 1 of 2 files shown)


GG/src/Font.cpp
```diff
@@ -240,6 +240,21 @@ namespace {
...
+    {
+        typedef void result_type;
+        void operator()(const std::string* str,
+                        std::stack<Font::Substring>& tag_stack,
+                        bool& ignore_tags,
+                        const boost::xpressive::ssub_match& sub) const
+        {
+            tag_stack.push(Font::Substring(*str, sub));
+            if (tag_stack.top() == PRE_TAG)
+                ignore_tags = true;
```
adrianbroher, line 257: `ignore_tags` may break the stack logic in case of nested `<pre>` tags.

GG/src/Font.cpp
```diff
@@ -685,33 +700,167 @@ Font::Glyph::Glyph(const boost::shared_ptr<Texture>& texture, const Pt& ul, cons
...
-        ActionTags().insert("u");
-        ActionTags().insert("rgba");
-        ActionTags().insert(ALIGN_LEFT_TAG);
-        ActionTags().insert(ALIGN_CENTER_TAG);
-        ActionTags().insert(ALIGN_RIGHT_TAG);
-        ActionTags().insert(PRE_TAG);
-
-        // Always know the action tags.
-        KnownTags().insert(ActionTags().begin(), ActionTags().end());
+        StaticTagHandler().Insert("i");
```
adrianbroher: This stems from old code but while you're fixing this already: the literals should be replaced with constants like it was already done for ALIGN_LEFT_TAG and relatives.

GG/src/Font.cpp
```diff
@@ -1097,61 +1242,29 @@ void Font::ThrowBadGlyph(const std::string& format_str, boost::uint32_t c)
...
-#endif
-
-    std::vector<boost::shared_ptr<TextElement> > local_text_elements;
-
-    std::vector<boost::shared_ptr<TextElement> >& text_elements =
-        text_elements_ptr ? *text_elements_ptr : local_text_elements;
-
+void Font::FillTextElements(const std::string& text,
+                            bool ignore_tags,
+                            std::vector<boost::shared_ptr<TextElement> >& text_elements) const {
```
adrianbroher: Missing line break before brace.

GG/src/Font.cpp
```diff
@@ -1097,61 +1242,29 @@ void Font::ThrowBadGlyph(const std::string& format_str, boost::uint32_t c)
...
     if (text_elements.empty()) {
         using namespace boost::xpressive;
 
-        std::stack<Substring> tag_stack;
-        bool ignore_tags = format & FORMAT_IGNORETAGS;
-        MatchesKnownTag matches_known_tag(KnownTags(), ignore_tags);
-        MatchesTopOfStack matches_tag_stack(tag_stack, ignore_tags);
+        // Fetch and use the regular expression from the TagHandler which parses all the known XML tags.
+        sregex & regex = StaticTagHandler().Regex(text, ignore_tags);
```
adrianbroher: Unnecessary space in front of ampersand.

GG/src/Font.cpp
```diff
@@ -1097,61 +1242,29 @@ void Font::ThrowBadGlyph(const std::string& format_str, boost::uint32_t c)
...
+        sregex_iterator it(text.begin(), text.end(), regex);
 
+        // These are the types found by the regular expression: XML open/close tags, text and
```
adrianbroher, line 1260: No need for change but those are not XML tags as they don't follow the specification for XML especially for attributes but 'something else'.

adrianbroher (changes requested):
Looks good. The stack should work properly except for nested `pre` tags.  Can you squash the style changes into the commits that were intoducing the style violations?

adrianbroher:
Thanks a lot for breaking the PR up.  That made it much easier to review it properly.  Could you also split     32b5149 into a 'update comments' commit and squash the style changes into the corresponding commits that introduced the style violation?

GG/src/Font.cpp
LGM-Doyle: That is correct. This is the existing behavior.  It is also documented in the comments of Font.h

GG/src/Font.cpp
LGM-Doyle: I added the new tags, as well as documentation that "s" for shadow is a recognized tag.

GG/src/Font.cpp
LGM-Doyle: And this is not an XML parser.  A "correct" XML is supposed to stop processing completely and report an error if anything about the document is incorrect.  

However, throughout freeorion's technical documentation we are referring to the various ad-hoc formats as XML, so I chose not to start referring to them as XML-like here.

LGM-Doyle:
I left the style and comment changes in a commit referencing the reviews.

It recognizes the work put in by the reviewers as valuable to the project and is worth recording in the commit log.

GG/src/Font.cpp
adrianbroher: Hm, time to file a bug then.

GG/src/Font.cpp
adrianbroher: Thanks. :+1:

adrianbroher:
> I left the style and comment changes in a commit referencing the reviews.
> 
> It recognizes the work put in by the reviewers as valuable to the project and is worth recording in the commit log.

We won't come to an agreement here.  But that's nothing that should prevent merging this PR.  Compiling and giving it a last test run before merging.

adrianbroher (approved):

adrianbroher:
So, which one is next? #1030? Can you rebase it?

LGM-Doyle:
#1030 thru #1033 are all in parallel and have been re-based.

PR #1030, opened 2016-10-07
In Font::FillTextElements skip work when text is empty.

LGM-Doyle:
This PR follows #1029

It skips doing work when `text` is empty.


adrianbroher (approved):
Not much to say about that one. Compiling and running the game doesn't show obvious flaws.

LGM-Doyle:
#1031, #1032 and #1033 should all still be conflict free, since they are independent of each other.

PR #1031, opened 2016-10-07
Use unordered set to track known tags in Font.

LGM-Doyle:
This PR follow #1029

It changes the know_tag from set to unordered_set to improve performance.

### DIFF
--- a/GG/src/Font.cpp
+++ b/GG/src/Font.cpp
@@ -39,6 +39,7 @@
 #include <boost/xpressive/xpressive.hpp>
 #include <boost/xpressive/regex_actions.hpp>
 #include <boost/assign/list_of.hpp>
+#include <boost/unordered_set.hpp>
 
 #include <cmath>
 #include <cctype>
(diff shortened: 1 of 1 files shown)


adrianbroher (commented):
Does this PR make a difference?

The amount of tags isn't that big that a hash table could play out its strength in container access time and I would assume that hashing comes with some cost too.

And the most important point: it's C++11 and we haven't migrated to that yet.

adrianbroher:
Eh, ignore the C++11 statement, I had std::unordered_set in mind, not boost::unordered_set

LGM-Doyle:
The PR makes a difference.

I tested it.  It provides a 6.5% improvement in split fleet times on the default fleet at game start.  

You can test it yourself with the "testing code" that I left in PR #989.  I tested mine with the pre-massaged version of PR #989. 

My tone may be overly strident, because I think that `unordered_set` would be the default choice of set if its implementation had been available when C++ was rolled out.  This is the case with newer languages
 where programmers default to the hash table implementation, unless ordering is specifically needed.  C++ `unordered_set` suffers from the perception that it is something extra and costly since it didn't arrive until TR1.   

Here is a longer argument for using `unordered_set` even if it doesn't show improvement on every machine today.

tldr: `set` provides no benefit (because tags aren't sorted), and `unordered_set` either costs nothing or only provides an improvement. 

We don't care about ordering in the set of tags.  So there is no requirement to use `set` for ordering.

There were more tags than I imagined. I added code to print out of the tags while I was debugging.  For example the Pedia adds tags for `empire` etc.

There is some break even point where there are enough tags in the table that a single lookup will be faster with `unordered_set`. The exact break even point does not matter because `unordered_set` is not too slow to use below the break even point  The exact break even point will depend on the library used boost, libstdc++(gcc), libc++ (clang), msvc.

It is unlikely (and pointless) to work to reduce the number of tags recognized.  So  there is some point where using unordered_set is the correct choice for all libraries because it is faster.

adrianbroher:
Okay, I take your word then.

geoffthemedio:
@adrianbroher Thanks for looking over all these changes. I've not had usable internet access on the weekends the last few weeks, so haven't been able to give this the attention it deserves.

(Also thanks @LGM-Doyle for implementing and explaining it all...)

PR #1032, opened 2016-10-07
Use precomputed space width for given Font.

LGM-Doyle:
This PR follows #1029


adrianbroher (approved):
No obvious logic errors, compiles as expected and doesn't show any display errors when running the game.

PR #1033, opened 2016-10-07
Improve text element consolidation in Font.cpp

LGM-Doyle:
This PR follow #1029


adrianbroher (approved):
I needed some time to understand the code, but it seems equivalent.

PR #1034, opened 2016-10-07
Improve Font::StripTags()

LGM-Doyle:
This PR follows # 1029 through #1033

This switches to using the regex from `CompiledRegex` for `Font::StripTags()`

### DIFF
--- a/GG/src/Font.cpp
+++ b/GG/src/Font.cpp
@@ -1175,7 +1175,7 @@ std::string Font::StripTags(const std::string& text, bool strip_unpaired_tags)
     mark_tag whitespace_tag(4);
     mark_tag text_tag(5);
 
-    std::string retval;
+    std::stringstream retval;
 
     // scan through matched markup and text, saving only the non-tag-text
     sregex_iterator it(text.begin(), text.end(), regex);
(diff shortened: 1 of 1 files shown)


LGM-Doyle:
Rebased.

GG/src/Font.cpp
```diff
@@ -1167,47 +1167,28 @@ std::string Font::StripTags(const std::string& text, bool strip_unpaired_tags)
 {
     using namespace boost::xpressive;
 
-    bool temp_bool = false;
-    std::stack<Substring> tag_stack;
-    MatchesKnownTag matches_known_tag(StaticTagHandler().KnownTags(), temp_bool);
-    MatchesTopOfStack matches_tag_stack(tag_stack, temp_bool);
+    sregex & regex = StaticTagHandler().Regex(text, false, strip_unpaired_tags);
 
     mark_tag tag_name_tag(1);
```
adrianbroher, line 1114: Not for this PR, but shouldn't those be part of TagHandler? There are at lease two other places where this is duplicated.

adrianbroher (approved):
Looks good.  The only user; clipboard related code seems to work as expected.

PR #1035, opened 2016-10-07
In Font.cpp split out debug functions.

LGM-Doyle:
This PR follows #1034

PR #1036, opened 2016-10-07
Remove gotchas variants of `Font::DetermineLines()`.

LGM-Doyle:
This PR follows #1035

This removes the gotcha variant of Font::DetermineLines that repeatedly create and discard the computationally expensive TextElement vector.

### DIFF
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -448,7 +448,7 @@ public:
 
     /** Wrapper around PreRenderText that provides dummy values for line start and end values.*/
     void PreRenderText(const Pt& ul, const Pt& lr, const std::string& text, Flags<TextFormat>& format,
-                       RenderCache& cache, const std::vector<LineData>* line_data = 0,
+                       RenderCache& cache, const std::vector<LineData>& line_data,
                        RenderState* render_state = 0) const;
 
     /** Fill the \p cache with glyphs corresponding to the passed in \p text and \p line_data.*/
@@ -474,7 +474,7 @@ public:
         call it repeatedly on a known text.
     */
     std::vector<boost::shared_ptr<Font::TextElement> > ExpensiveParseFromTextToTextElements(const std::string& text,
-                                                                                            const Flags<TextFormat>& format = FORMAT_NONE) const;
+                                                                                            const Flags<TextFormat>& format) const;
 
     /** Returns the maximum dimensions of the string in x and y, and populates
         \a line_data. */
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -477,20 +477,7 @@ public:
                                                                                             const Flags<TextFormat>& format) const;
 
     /** Returns the maximum dimensions of the string in x and y, and populates
-        \a line_data. */
-    Pt   DetermineLines(const std::string& text, Flags<TextFormat>& format, X box_width,
-                        std::vector<LineData>& line_data) const;
-
-    /** Returns the maximum dimensions of the string in x and y, and populates
-        \a line_data and \a text_elements.  Note that \a text_elements must be
-        empty. */
-    Pt   DetermineLines(const std::string& text, Flags<TextFormat>& format, X box_width,
-                        std::vector<LineData>& line_data,
-                        std::vector<boost::shared_ptr<TextElement> >& text_elements) const;
-
-    /** Returns the maximum dimensions of the string in x and y, and populates
-        \a line_data.  The contents of \a text_elements will be used, and the
-        equivalent work done by DetermineLines() will be skipped.  Supplying a
+        \a line_data.  The contents of \a text_elements will be used.  Supplying a
         \a text and a \a text_elements that are incompatible will result in
         undefined behavior. */
     Pt   DetermineLines(const std::string& text, Flags<TextFormat>& format, X box_width,
@@ -577,7 +564,7 @@ private:
                           Flags<TextFormat>& format,
                           X box_width,
                           std::vector<LineData>& line_data,
-                          std::vector<boost::shared_ptr<TextElement> >* text_elements_ptr) const;
+                          const std::vector<boost::shared_ptr<TextElement> >& text_elements) const;
 
     FT_Error          GetFace(FT_Face& face);
     FT_Error          GetFace(const std::vector<unsigned char>& file_contents, FT_Face& face);
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -476,10 +476,19 @@ public:
     std::vector<boost::shared_ptr<Font::TextElement> > ExpensiveParseFromTextToTextElements(const std::string& text,
                                                                                             const Flags<TextFormat>& format) const;
 
-    /** Returns the maximum dimensions of the string in x and y, and populates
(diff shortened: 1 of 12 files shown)


adrianbroher:
Rebased onto master.

GG/src/Font.cpp
```diff
@@ -1253,140 +1216,136 @@ namespace DebugOutput {
...
 
-                // Whitespace element
-                } else if ((*it)[whitespace_tag].matched) {
-                    boost::shared_ptr<Font::TextElement> element(new Font::TextElement(true, false));
-                    element->text = Substring(text, (*it)[whitespace_tag]);
-                    text_elements.push_back(element);
+    bool ignore_tags = format & FORMAT_IGNORETAGS;
+
+    // Fetch and use the regular expression from the TagHandler which parses all the known XML tags.
+    sregex & regex = StaticTagHandler().Regex(text, ignore_tags);
```
adrianbroher: Space in front of ampersand.

GG/src/Font.cpp
```diff
@@ -1253,140 +1216,136 @@ namespace DebugOutput {
     }
 }
 
-void Font::FillTextElements(const std::string& text,
-                            bool ignore_tags,
-                            std::vector<boost::shared_ptr<TextElement> >& text_elements) const
+std::vector<boost::shared_ptr<Font::TextElement> > Font::ComputeTextElements(
```
adrianbroher: We align parameters to the opening bracket.

GG/src/Font.cpp
```diff
@@ -1253,140 +1216,136 @@ namespace DebugOutput {
...
+    const std::string& text,
+    const Flags<TextFormat>& format) const
 {
```
adrianbroher, line 1221: General remark:  Breaking down the commit into several commits like 'detending function content', 'renaming function' and 'parameter/return value usage' would help significantly when reviewing a PR.  I wasn't sure that this is essentially the same function until I used `git difftool` and applied some vimdiff changes locally.

GG/src/Menu.cpp
```diff
@@ -465,7 +465,9 @@ void PopupMenu::Render()
             }
             std::vector<Font::LineData> lines;
             Flags<TextFormat> fmt = FORMAT_LEFT | FORMAT_TOP;
-            Pt menu_sz = m_font->DetermineLines(str, fmt, X0, lines); // get dimensions of text in menu
+            std::vector<boost::shared_ptr<Font::TextElement> > text_elements
+                = m_font->ComputeTextElements(str, fmt);
+            Pt menu_sz = m_font->DetermineLines(str, fmt, X0, text_elements, lines); // get dimensions of text in menu
```
adrianbroher, line 470: `X0` is magic. What does it do?  Should go into the `DetermineLines` documentation.

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2439,7 +2439,12 @@ namespace {
             std::string species_name = *it;
 
             std::string species_name_column1 = str(FlexibleFormat(UserString("ENC_SPECIES_PLANET_TYPE_SUITABILITY_COLUMN1")) % UserString(species_name)); 
-            max_species_name_column1_width = std::max(font->TextExtent(species_name_column1).x, max_species_name_column1_width);
+            std::vector<GG::Font::LineData> lines;
+            GG::Flags<GG::TextFormat> format = GG::FORMAT_NONE;
+            std::vector<boost::shared_ptr<GG::Font::TextElement> > text_elements =
+                font->ComputeTextElements(species_name_column1, format);
+            GG::Pt extent = font->DetermineLines(species_name_column1, format, GG::X(1 << 15), text_elements, lines);
```
adrianbroher: `GG::X(1 << 15)` is magic². What does it do?  An arbitrary huge value? A constant and some documentation would help.  Also remember there are other uses of this pattern when updating this PR.

adrianbroher (changes requested):
Looks good in general but the documentation is lacking.

LGM-Doyle (commented):
Thanks for the review.

I don't think you needed to rebase the last few PRs since I have already rebased them to the previous PR in the series, unlike the previous 3 which were in parallel.  I have also rebased the next few PRs as a series.

GG/src/Menu.cpp
geoffthemedio: Explain that in a comment then, as apparently it's not clear enough just looking at the code.

GG/GG/Font.h
```diff
@@ -465,33 +465,19 @@ class GG_API Font
...
+    std::vector<boost::shared_ptr<TextElement> > ComputeTextElements(
+        const std::string& text, const Flags<TextFormat>& format = FORMAT_NONE) const;
 
     /** Returns the maximum dimensions of the string in x and y, and populates
-        \a line_data.  The contents of \a text_elements will be used, and the
-        equivalent work done by DetermineLines() will be skipped.  Supplying a
+        \a line_data.  The contents of \a text_elements will be used.  Supplying a
         \a text and a \a text_elements that are incompatible will result in
         undefined behavior. */
     Pt   DetermineLines(const std::string& text, Flags<TextFormat>& format, X box_width,
```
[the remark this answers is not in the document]
adrianbroher: > If the function decides where linebreaks are, then maybe call it BreakTextIntoLines or DetermineLinebreakLocationsForText ?

The purpose isn't limited to line breaking, look into the implementation.
1. Break text at the target `box_width` depending on `format`.
2. Text justification depending on `format` (delegating the implementation details to `SetJustification`).

You can see it as you want, but you pour in TextElement s, a document width and justification rules and get a [typesetting](https://en.wikipedia.org/wiki/Typesetting), which is the very meaning of text layouting.

GG/src/Menu.cpp
LGM-Doyle: I have already added the following comment in the documentation of the definition of the `DetermineLines()` near [Font.h:486](https://github.com/LGM-Doyle/freeorion/blob/22632e325886231a2eb0a57cbdc58280d6361f20/GG/GG/Font.h#L486)

`A \p box_width of X0 will add a line break at every whitespace element in \p text_elements.`

I was not clear about that in the previous comment.

Do you also want an additional comment at each invocation of the function that uses `X0` as a value?

GG/GG/Font.h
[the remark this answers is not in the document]
LGM-Doyle: Pick any name and I'll change it in the code.

`DetermineLines` does not position text.  Positioning of text is done in `PreRenderText`.  `DetermineLines` only figures out which characters go on which lines and adds the justification information to each `LineData` so that `PreRenderText` can use the correct justification when positioning text.

geoffthemedio:
Is this currently the next of this series of PRs to be tested / reviewed / merged?

geoffthemedio:
Builds OK, seems to work as before.

LGM-Doyle:
Thanks for coming back around to these PRs.

The order is #1036 through #1042 and then #989 from which they were all broken out.

 For this PR I had two remaining questions:
- Do you want to pick a different name for `DetermineLines`?
- Do you want comments about the use of `X0` as a width parameter at each invocation as well as in the header definition of the function?

geoffthemedio:
I think the current name and level of documentation are OK.

PR #1037, opened 2016-10-07
Improve font determine lines streamline use of Font::TextExtent()

LGM-Doyle:
This follow PR #1036.

PR #1038, opened 2016-10-07
Improve font determine lines change a copy to a reference.

LGM-Doyle:
This PR follows #1037


LGM-Doyle:
Yes this PR just changes that pointer to a reference.

PR #1039, opened 2016-10-07
In Font.cpp create TextAndElementsAssembler.

LGM-Doyle:
This PR follows #1038.

TextAndElementsAssembler allows the creation of a matched pair of text and vector of `TextElement` efficiently.

### DIFF
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -247,20 +247,18 @@ public:
         mutable X cached_width;
     };
 
-    /** \brief Used to encapsulate a text and tokens together to be rendered
-        using GG::Font.
-
-        TextElements allows a matched pair of XML string and vector of TextElement to be directly
-        constructed while avoiding the computationally costly ComputeTextElements() function.
+    /** \brief TextAndElementsAssembler is used to assemble a matched pair of text and a vector of
+        TextElement, without incurring the computational cost of parsing the text with
+        ComputeTextElements().
 
         The pair of string and vector returned by Text() and Elements() are consistent with each
-        other and can be used with the fast constructor for TextControl.
+        other and can be used with the fast constructor or the fast SetText variant of TextControl.
     */
-    class GG_API TextElements
+    class GG_API TextAndElementsAssembler
     {
     public:
-        TextElements(const Font& font);
-        ~TextElements();
+        TextAndElementsAssembler(const Font& font);
+        ~TextAndElementsAssembler();
 
         /** Return the constructed text.*/
         const std::string& Text();
@@ -268,24 +266,24 @@ public:
         const std::vector<boost::shared_ptr<TextElement> >& Elements();
 
         /** Add an open tag iff it exists as a recognized tag.*/
-        TextElements& AddOpenTag(const std::string& tag);
+        TextAndElementsAssembler& AddOpenTag(const std::string& tag);
         /** Add an open tag iff it exists as a recognized tag.*/
-        TextElements& AddOpenTag(const std::string& tag, const std::vector<std::string>& params);
+        TextAndElementsAssembler& AddOpenTag(const std::string& tag, const std::vector<std::string>& params);
         /** Add a close tag iff it exists as a recognized tag.*/
-        TextElements& AddCloseTag(const std::string& tag);
+        TextAndElementsAssembler& AddCloseTag(const std::string& tag);
         /** Add a text element.  Any whitespace in this text element will be non-breaking.*/
-        TextElements& AddText(const std::string& text);
+        TextAndElementsAssembler& AddText(const std::string& text);
         /** Add a white space element.*/
-        TextElements& AddWhitespace(const std::string& whitespace);
+        TextAndElementsAssembler& AddWhitespace(const std::string& whitespace);
         /** Add a new line element.*/
-        TextElements& AddNewline();
+        TextAndElementsAssembler& AddNewline();
 
         /** Add an open Clr tag.*/
-        TextElements& AddOpenTag(const Clr& color);
+        TextAndElementsAssembler& AddOpenTag(const Clr& color);
 
     private:
-        class TextElementsImpl;
-        boost::scoped_ptr<TextElementsImpl> const pimpl;
+        class TextAndElementsAssemblerImpl;
+        boost::scoped_ptr<TextAndElementsAssemblerImpl> const pimpl;
     };
 
     /** \brief The type of TextElement that represents a text formatting
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -147,6 +147,9 @@ public:
             pair.second. */
         Substring(const std::string& str_, const IterPair& pair);
 
+        /** Rebind to a different string.*/
+        void Bind(const std::string& str_);
+
         /** Returns an iterator to the beginning of the substring. */
         std::string::const_iterator begin() const;
 
@@ -206,6 +209,9 @@ public:
 
         virtual ~TextElement(); ///< Virtual dtor.
 
+        /** Rebind to a different string.*/
+        virtual void Bind(const std::string& whole_text);
+
         /** Returns the TextElementType of the element. */
         virtual TextElementType Type() const;
 
@@ -241,6 +247,9 @@ public:
         /** Ctor.  \a close indicates that the tag is a close-tag
             (e.g. "</rgba>"). */
         FormattingTag(bool close);
+
+        /** Rebind to a different string.*/
+        virtual void Bind(const std::string& whole_text);
 
         virtual TextElementType Type() const;
 
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -247,6 +247,40 @@ public:
         mutable X cached_width;
     };
 
+    /** \brief Used to encapsulate a text and tokens together to be rendered
+        using GG::Font. */
+    class GG_API TextElements
+    {
+    public:
+        TextElements(const Font& font);
+        ~TextElements();
+
+        /** Return the constructed text.*/
+        const std::string& Text();
+        /** Return the constructed TextElements.*/
+        const std::vector<boost::shared_ptr<TextElement> >& Elements();
+
+        /** Add an open tag iff it exists as a recognized tag.*/
+        TextElements& AddOpenTag(const std::string& tag);
+        /** Add an open tag iff it exists as a recognized tag.*/
+        TextElements& AddOpenTag(const std::string& tag, const std::vector<std::string>& params);
+        /** Add a close tag iff it exists as a recognized tag.*/
+        TextElements& AddCloseTag(const std::string& tag);
+        /** Add a text element.  Any whitespace in this text element will be non-breaking.*/
+        TextElements& AddText(const std::string& text);
+        /** Add a white space element.*/
+        TextElements& AddWhitespace(const std::string& whitespace);
+        /** Add a new line element.*/
+        TextElements& AddNewline();
+
+        /** Add an open Clr tag.*/
+        TextElements& AddOpenTag(const Clr& color);
+
+    private:
+        class TextElementsImpl;
+        boost::scoped_ptr<TextElementsImpl> const pimpl;
+    };
+
     /** \brief The type of TextElement that represents a text formatting
         tag. */
     struct GG_API FormattingTag : TextElement
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -491,6 +491,13 @@ public:
     std::vector<boost::shared_ptr<Font::TextElement> > ExpensiveParseFromTextToTextElements(const std::string& text,
                                                                                             const Flags<TextFormat>& format) const;
 
+    /** Fill \p text_elements with the font widths of characters from \p text starting from \p
+        start. */
+    void FillTemplatedText(
+        const std::string& text,
+        std::vector<boost::shared_ptr<TextElement> >& text_elements,
+        std::vector<boost::shared_ptr<TextElement> >::iterator starting_from) const;
+
     /** DetermineLines() returns the \p line_data resulting from adding the necessary line
         breaks, to  the \p text formatted with \p format and parsed into \p text_elements, to fit
         the \p text into a box of width \p box_width.
--- a/GG/GG/Font.h
+++ b/GG/GG/Font.h
@@ -129,7 +129,11 @@ class GG_API Font
 {
 public:
     /** \brief A range of iterators into a std::string that defines a
-        substring found in a string being rendered by Font. */
+        substring found in a string being rendered by Font.
+
+        Substring is bound to a particular instance of a std::string.  If
+        that particular std::string goes out of scope or is deleted then
+        behavior is undefined, but may seg fault with the next access. */
     class GG_API Substring
     {
     public:
@@ -147,7 +151,11 @@ public:
             pair.second. */
         Substring(const std::string& str_, const IterPair& pair);
 
-        /** Rebind to \p str_, by changing the pointer str to point to \p str_.*/
+        /** Attach this Substring to \p str_.
+
+            This changes the iterators from pointing into the previous
+            std::string to pointing into \p str_.
+        */
         void Bind(const std::string& str_);
 
         /** Returns an iterator to the beginning of the substring. */
@@ -209,11 +217,26 @@ public:
 
         virtual ~TextElement(); ///< Virtual dtor.
 
-        /** Rebind to \p whole_text, by binding the SubString data member text to \p whole_text.
+        /** Attach this TextElement to the string \p whole_text, by
+            attaching the SubString data member text to \p whole_text.
 
-            Binding to a new \p whole_text is very fast compared to re-parsing the entire \p
-            whole_text and allows TextElements of TextElementType TEXT to be changed quickly if it
-            is known that the parse would be the same.
+            Binding to a new \p whole_text is very fast compared to
+            re-parsing the entire \p whole_text and allows TextElements of
+            TextElementType TEXT to be changed quickly if it is known that
+            the parse would be the same.
+
+            It is efficient when you want to do a text parse or layout
+            once, and then create several different controls that have the
+            same text layout and text contents, but each of them needs to
+            keep there own internal copy of the text. So, while the pointer
+            diffs are all the same, since the text contents are the same,
+            the pointer to the string in each TextElement (and its
+            contained Substring) needs to be set to the appropriate copy of
+            the string.
+
+            This is used in TextControl and its derived classes to re-use
+            entire vectors of TextElement with different std::strings
+            without re-parsing the std::string.
          */
         virtual void Bind(const std::string& whole_text);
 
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Can you rebase this as well? Sorry... it's just rather difficult to look over the changes with all the old ones mixed in.

LGM-Doyle:
I will rebase it.

No need to apologize.

I was trying to find a way to reduce rebasing iteration time.  I had hoped that github would show the diffs from master in `Files changed`.

LGM-Doyle:
Rebased.

geoffthemedio:
I think it does... but from the revision you branch off of... (but I'm not sure...)

GG/src/Font.cpp
```diff
@@ -527,9 +722,216 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+}
+
+///////////////////////////////////////
+// class GG::Font::TextAndElementsAssembler
+///////////////////////////////////////
+class Font::TextAndElementsAssembler::TextAndElementsAssemblerImpl
+{
+public:
+    TextAndElementsAssemblerImpl(const Font& font) :
+        m_font(font), m_text(), m_text_elements(), m_are_widths_calculated(false)
```
geoffthemedio: split initializer list over lines

GG/src/Font.cpp
```diff
@@ -527,9 +722,216 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+    {}
+
+    /** Return the constructed text.*/
+    const std::string& Text()
```
geoffthemedio: make const?

GG/src/Font.cpp
```diff
@@ -527,9 +722,216 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+
+    /** Add an open tag iff it exists as a recognized tag.*/
+    void AddOpenTag(const std::string& tag, const std::vector<std::string>* params = 0)
+    {
+        if (!StaticTagHandler().IsKnown(tag))
+            return;
+
+        m_are_widths_calculated = false;
+
+        boost::shared_ptr<Font::FormattingTag> element(new Font::FormattingTag(false));
```
geoffthemedio, line 785: comment per block of code, perhaps?

GG/src/Font.cpp
```diff
@@ -527,9 +722,216 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+        boost::shared_ptr<Font::TextElement> element(new Font::TextElement(true, false));
+        size_t begin = m_text.size();
+        size_t end = m_text.append(whitespace).size();
+        element->text = Substring(m_text,
+                                  boost::next(m_text.begin(), begin),
+                                  boost::next(m_text.begin(), end));
+        m_text_elements.push_back(element);
+    }
+
+    /** Add a white space element.*/
```
geoffthemedio: probably need to correct comment?

GG/src/Font.cpp
```diff
@@ -527,9 +722,216 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+    void AddNewline()
+    {
+        m_are_widths_calculated = false;
+
+        boost::shared_ptr<Font::TextElement> element(new Font::TextElement(false, true));
+        m_text_elements.push_back(element);
+    }
+
+    /** Add open color tag.*/
+    void AddOpenTag(const Clr& color) {
```
geoffthemedio: be consistent with { placement

GG/src/Font.cpp
```diff
@@ -527,9 +722,216 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+const std::vector<boost::shared_ptr<Font::TextElement> >& Font::TextAndElementsAssembler::Elements()
+{ return pimpl->Elements(); }
+
+Font::TextAndElementsAssembler& Font::TextAndElementsAssembler::AddOpenTag(const std::string& tag) {
+    pimpl->AddOpenTag(tag);
+    return *this;
+}
+
+Font::TextAndElementsAssembler& Font::TextAndElementsAssembler::AddOpenTag(
+    const std::string& tag, const std::vector<std::string>& params) {
```
geoffthemedio: especially after multi-line function name line, put { on newline. also do so here and around here to be consistent

geoffthemedio (changes requested):

geoffthemedio:
Should I notice improvements in text layout speed? For my standard test of continuously resizing the pedia window while it shows zero point generation, I don't see an improvement...

LGM-Doyle:
No you should not notice any changes in layout speed.

This PR is just about creating `TextAndElementsAssembler` to allow creation of the vector of `TextElements` without having to parse the XML type string.  

It is not used anywhere until [#992 Use Wnd::PreRender and new Font functions in various files.](https://github.com/freeorion/freeorion/pull/992) when it is used in the `StatisticIcon` which was one of the major factors of slowing the `FleetWnd`.  It will be useful in other locations as well, but the driving test case for this series of PRs was improving `FleetWnd` rendering time from minutes to milliseconds.

This was why I originally left in the `SecionedTimer` code, so that people could track which commits were improving things and by how much. 

I am working on the changes you requested, but I will not finish today.

GG/src/Font.cpp
```diff
@@ -478,6 +484,186 @@ std::pair<std::size_t, CPSize> GG::LinePositionOf(CPSize index, const std::vecto
...
+            m_known_tags.clear();
+        }
+
+        bool IsKnown(const std::string &tag) {
+            return m_known_tags.find(tag) != m_known_tags.end();
+        }
+
+        // Return a regex bound to \p text using the currently known
+        // tags.  If required \p ignore_tags and/or \p strip_unpaired_tags.
+        xpr::sregex & Regex(const std::string& text, bool ignore_tags, bool strip_unpaired_tags = false) {
```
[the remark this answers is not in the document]
LGM-Doyle: Fixed the space.

It can't be const because the `BindRegexToText` clears and returns either `m_regex_w_tags` or `m_regex_w_tags_skipping_unmatched`, neither of which is mutable.

LGM-Doyle:
I've made the changes, except the one change to const I noted above.

GG/GG/Font.h
```diff
@@ -206,6 +209,14 @@ class GG_API Font
 
         virtual ~TextElement(); ///< Virtual dtor.
 
+        /** Rebind to \p whole_text, by binding the SubString data member text to \p whole_text.
+
+            Binding to a new \p whole_text is very fast compared to re-parsing the entire \p
+            whole_text and allows TextElements of TextElementType TEXT to be changed quickly if it
+            is known that the parse would be the same.
```
geoffthemedio: Can you give an example when this would apply?

My initial thought was adding / removing underlines, but in that case the extra `<u>` and `</u>` would presumably move the index in the whole_text string that each TextElement should point to.

Perhaps colour changes, eg. for mouseovers, but then wouldn't the number of characters in the numbers in the rgba tag possibly change?

Or do I not get/understand how TextElement indexes into raw strings?

GG/src/Font.cpp
```diff
@@ -478,6 +484,186 @@ std::pair<std::size_t, CPSize> GG::LinePositionOf(CPSize index, const std::vecto
...
+    namespace xpr = boost::xpressive;
+
+    /** CompiledRegex maintains a compiled boost::xpressive regular
+        expression that includes a tag stack which can be cleared and
+        provided to callers without the overhead of recompiling the
+        regular expression.*/
+    class CompiledRegex {
+    public:
+        CompiledRegex(const boost::unordered_set<std::string>& known_tags, bool strip_unpaired_tags) :
+            m_text(0), m_known_tags(&known_tags), m_ignore_tags(false), m_tag_stack(),
```
geoffthemedio: split initializer list over lines...

GG/src/Font.cpp
```diff
@@ -478,6 +484,186 @@ std::pair<std::size_t, CPSize> GG::LinePositionOf(CPSize index, const std::vecto
...
+    public:
+        TagHandler() :
+            m_known_tags(),
+            m_regex_w_tags(m_known_tags, false),
+            m_regex_w_tags_skipping_unmatched(m_known_tags, true)
+        {}
+
+        /** Add a tag to the set of known tags.*/
+        void Insert(const std::string& tag) {
+            m_known_tags.insert(tag);
```
geoffthemedio: for single-line functions, can put { and } on same line as the... single line.

GG/src/Font.cpp
```diff
@@ -478,6 +484,186 @@ std::pair<std::size_t, CPSize> GG::LinePositionOf(CPSize index, const std::vecto
...
+            m_known_tags.clear();
+        }
+
+        bool IsKnown(const std::string &tag) const {
+            return m_known_tags.find(tag) != m_known_tags.end();
+        }
+
+        // Return a regex bound to \p text using the currently known
+        // tags.  If required \p ignore_tags and/or \p strip_unpaired_tags.
+        xpr::sregex& Regex(const std::string& text, bool ignore_tags, bool strip_unpaired_tags = false) {
```
geoffthemedio: for multi-line functions in GG code, should generally have the { on a new line after the function name / signature line

GG/src/Font.cpp
```diff
@@ -514,6 +703,12 @@ CPSize Font::TextElement::CodePointSize() const
 { return CPSize(widths.size()); }
 
 
+bool Font::TextElement::operator==(const TextElement &rhs) const
+{
```
geoffthemedio, line 708: ...like this...

GG/src/Font.cpp
```diff
@@ -527,9 +722,226 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+    }
+}
+
 Font::FormattingTag::TextElementType Font::FormattingTag::Type() const
 { return close_tag ? CLOSE_TAG : OPEN_TAG; }
 
+bool Font::FormattingTag::operator==(const TextElement &rhs) const
+{
+    Font::FormattingTag const* ft = dynamic_cast<Font::FormattingTag const*>(&rhs);
+    return (ft
```
geoffthemedio, line 743: not required, but I'd prefer the && at the end of lines here... just me being arbitrary, though...

GG/src/Font.cpp
```diff
@@ -527,9 +722,226 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+// Required because Impl is defined here
+Font::TextAndElementsAssembler::~TextAndElementsAssembler()
+{}
+const std::string& Font::TextAndElementsAssembler::Text()
+{ return pimpl->Text(); }
+
+const std::vector<boost::shared_ptr<Font::TextElement> >& Font::TextAndElementsAssembler::Elements()
+{ return pimpl->Elements(); }
+
+Font::TextAndElementsAssembler& Font::TextAndElementsAssembler::AddOpenTag(const std::string& tag) {
```
geoffthemedio: { to next line

GG/src/Font.cpp
```diff
@@ -527,9 +722,226 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+}
+
+Font::TextAndElementsAssembler& Font::TextAndElementsAssembler::AddOpenTag(
+    const std::string& tag, const std::vector<std::string>& params)
+{
+    pimpl->AddOpenTag(tag, &params);
+    return *this;
+}
+
+Font::TextAndElementsAssembler& Font::TextAndElementsAssembler::AddCloseTag(const std::string& tag) {
```
geoffthemedio: particularly here, need to be consistent with surrounding code wrt the { placement

geoffthemedio (changes requested):

GG/src/Font.cpp
```diff
@@ -527,9 +722,226 @@ Font::FormattingTag::FormattingTag(bool close) :
...
+    std::string m_text;
+    std::vector<boost::shared_ptr<TextElement> > m_text_elements;
+    bool m_are_widths_calculated;
+};
+
+
+Font::TextAndElementsAssembler::TextAndElementsAssembler(const Font& font) :
+    pimpl(new TextAndElementsAssemblerImpl(font))
+{}
+// Required because Impl is defined here
```
geoffthemedio, line 902: prefer blank lines between function definitions / bodies

geoffthemedio (changes requested):

GG/GG/Font.h
LGM-Doyle: I've expanded the comment for the Bind function to add use cases.

I also added a copy constructor and an assignment constructor in the PR #989 where the constructor that uses the results of `TextAndElementsAssembler` is added.  This is the second obvious use case for `Bind()`.

The operation of adding/removing tags like `<u>`, was not something that I added. It did not impact the performance of the `ListWnd` `FleetWnd` which was the focus of this set of PRs.  This functionality would be the counterpart to `ChangeTemplatedText`, but it doesn't exist.

I also made the requested style changes.

GG/GG/Font.h
```diff
@@ -206,6 +209,22 @@ class GG_API Font
...
+
+            One use case is copying or copy constructing a TextControl.
```
geoffthemedio: When is a TextControl ever copied?

GG/GG/Font.h
```diff
@@ -206,6 +209,22 @@ class GG_API Font
...
+
+            Another use case is constructing a string and TextElement vector pair with
```
geoffthemedio: I don't really understand the examples... possibly because here is too low level for the relevant context to make sense...?

GG/GG/Font.h
LGM-Doyle: `TextControl`s are currently not copied.

However, `TextControl`s are constructed or set from other `TextControl`s which is an example of unnecessary parsing which would be fixed by `Bind()`.  Since you brought it up I added the copy constructor and the comment here.  This was not a use case that came up fixing the `ListWnd`.

GG/GG/Font.h
LGM-Doyle: tldr: Would changing the name `Bind()` to something more explicit like `AttachTo(const string&)` fix the problem?

I think you are correct that this is too low level to immediately see the relevance.  `Bind()` does what its name implies and binds, reassign an internal pointer of the `Substring` to a new string.  `TextElement` which is a composite containing a `Substring` needs bind its `Substring`s members to new strings.

I would remove the cases I added to the function comment, because `Bind()` is general and will be needed for all cases where a `Substring` needs to point to a new string, because they are tightly coupled.  

The motivating case for `TextAndElementsAssembler` was from `StatisticIcon::DoLayout()` and it might clarify the use case for `Bind()`.  But, again this is two steps removed from the `Bind()` function on its own.

`StatisticIcon::DoLayout()` became a bottleneck in improving `FleetWnd`.   I will explain that here incase that helps you decide how to comment `Bind()`.  `DoLayout()` generates the icon plus number used in many places in freeorion: `FleetWnd`, `PlanetTab`, `MapWnd`'s tab bar, etc.  For the `FleetWnd` alone avoiding re-parsing the XML saves (6 per fleet) \* (6 per ship in fleet) of the parsing cost each time.  

Here is a section of the code from PR #992, which uses `TextAndElementsAssembler`, which in turn uses `Bind()`.

```
    `// Precompute text elements
GG::Font::TextAndElementsAssembler text_elements(*ClientUI::GetFont());
text_elements.AddOpenTag(ValueColor(0))
    .AddText(DoubleToString(m_values[0], m_digits[0], m_show_signs[0]))
    .AddCloseTag("rgba");

    // Create new label in correct place.
    m_text = new CUILabel(text_elements.Text(), text_elements.Elements(),
                          format, GG::NO_WND_FLAGS,
                          text_ul.x, text_ul.y, text_lr.x - text_ul.x, text_lr.y - text_ul.y);
    `
```

Would changing the name `Bind()` to something more explicit like `AttachTo(const string&)` fix the problem?

geoffthemedio: I don't mind the name Bind.

But it's not great when the comments explaining the purpose / result of the function use the functions name to explain itself... ie. saying Bind "rebinds" something doesn't help much. The rest of the comment is better, though.

The example aren't very clear, though... More detail would be better, about why situations would arise in which you'd want to change the pointed to string.

After reading this several times, I think it would make sense when you want to do a text parse or layout once, and then create several different controls that have the same text layout and text contents, but each of them needs to keep there own internal copy of the text. So, while the pointer diffs are all the same, since the text contents in the same, the pointer to the string in each TextElement (and its contained Substring) needs to be set to the appropriate copy of the string.

At this point though, I wonder why you need to index into the strings, rather than just keeping the necessary string inside each TextElement itself? If you end up making another copy of the whole string every time you make another copy of the control, then what's the benefit of indexing instead of having a bunch of smaller strings internally...?

LGM-Doyle: I prioritized working on #1083 to fix the combat log glitches introduced by the merge of #1076, since that is a problem in master.

I'll try to fix this comment again.

I'll point out again that the `Substring` design is the original implementation.  I assume that it was profiled when it was designed and that creating many `Substrings` was either an unacceptable time and/or memory overhead.

LGM-Doyle: I've made the following changes:
- I used your words from above to describe use cases
- changed the word rebind to attach
- added clarification to the `Substring` description that binding points the iterators into the new string.

GG/GG/Font.h
```diff
@@ -147,6 +151,13 @@ class GG_API Font
             pair.second. */
         Substring(const std::string& str_, const IterPair& pair);
 
+        /** Attach this Substring to \p str_.
+
+            This changes the iterators from pointing into the previous
```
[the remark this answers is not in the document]
LGM-Doyle: Yes, it is misleading.  

I do not think that the iterators are ever currently stored.  But they could be.

It could just say that previous iterators are invalidated and any use will have indeterminate effects, then any use after rebinding it the responsibility of the programmer.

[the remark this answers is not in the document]
geoffthemedio: It won't actually invalidate the iterators, though, as SubString just stores a pointer to a strong. What would invalidate would be if the pointed-to-string was manipulated in such a way that its iterators are invalidated. Code that uses this might cause this, but not this function in of itself... (I think...)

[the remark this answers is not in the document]
LGM-Doyle: It is the same behavior as the C++ standard library.  

When an operation on a container, insert,  delete etc., invalidates the iterators for that container, the iterators themselves are not changed, but they are invalid by definition and using them has undefined behavior. 

In our case, once the `Substring` has been bound to a new `std::string` any iterators stored are invalid.  They might by accident produce correct results if the old string and the new string are identical, but they are definitely pointing to the a string not bound to this `Substring`and perhaps a string that no longer exists.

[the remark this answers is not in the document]
geoffthemedio: Rebinding a SubString doesn't invalidate in the same sense as the STL, though. You could have several substrings pointing into the same std::string, and rebinding one would not affect the others, because the pointed-to string is still valid and unchanged by the substring rebinding.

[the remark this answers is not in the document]
LGM-Doyle: The iterator is invalid because it does not point to the string the its own `SubString` is pointing to.  If it still points to a string that has not been de-allocated, then dereferencing it won't cause a crash, but it is still invalid.

So with multiple Substrings all pointing to the same string, rebinding one Substring does not invalidate any iterators from the other Substrings.  It does invalidate the iterators from the rebound Substring, but happenstance prevents dereferencing those invalid iterators from causing a segmentation fault.

PR #1043, opened 2016-10-08
Housekeeping fix brackets

LGM-Doyle:
This PR fixes all of the warnings exposed by -Wparentheses by adding the requested brackets.

### DIFF
--- a/GG/GG/utf8/checked.h
+++ b/GG/GG/utf8/checked.h
@@ -120,13 +120,13 @@ namespace utf8
         }
         else if (cp < 0x10000) {              // three octets
             *(result++) = static_cast<uint8_t>((cp >> 12)           | 0xe0);
-            *(result++) = static_cast<uint8_t>((cp >> 6) & 0x3f     | 0x80);
+            *(result++) = static_cast<uint8_t>(((cp >> 6) & 0x3f)   | 0x80);
             *(result++) = static_cast<uint8_t>((cp & 0x3f)          | 0x80);
         }
         else if (cp <= internal::CODE_POINT_MAX) {      // four octets
             *(result++) = static_cast<uint8_t>((cp >> 18)           | 0xf0);
-            *(result++) = static_cast<uint8_t>((cp >> 12)& 0x3f     | 0x80);
-            *(result++) = static_cast<uint8_t>((cp >> 6) & 0x3f     | 0x80);
+            *(result++) = static_cast<uint8_t>(((cp >> 12) & 0x3f)  | 0x80);
+            *(result++) = static_cast<uint8_t>(((cp >> 6) & 0x3f)   | 0x80);
             *(result++) = static_cast<uint8_t>((cp & 0x3f)          | 0x80);
         }
         else
--- a/UI/ObjectListWnd.cpp
+++ b/UI/ObjectListWnd.cpp
@@ -1392,7 +1392,7 @@ private:
 
         for (std::size_t i = 1; i < m_controls.size(); ++i) {
             right = left + GetColumnWidth(static_cast<int>(i));
-            if (ctrl = m_controls[i])
+            if ((ctrl = m_controls[i]))
                 ctrl->SizeMove(GG::Pt(left, top), GG::Pt(right, bottom));
             left = right + PAD;
         }
--- a/parse/EffectParser1.cpp
+++ b/parse/EffectParser1.cpp
@@ -79,7 +79,7 @@ namespace {
                             // useful to specify a single recipient empire, or the allies
                             // or enemies of a single empire
                             (   (parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ])
-                            |   eps [ _d = AFFIL_SELF ]
+                            |    eps [ _d = AFFIL_SELF ]
                             )
                         >>  parse::label(Empire_token) > int_value_ref
                             [ _val = new_<Effect::GenerateSitRepMessage>(_a, _b, _c, _1, _d, _e, _f) ]
@@ -98,8 +98,8 @@ namespace {
                         )
                     |   (   // no empire id or condition specified, with or without an
                             // affiliation type: useful to specify no or all empires
-                        (   (parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ])
-                            |   eps [ _d = AFFIL_ANY ]
+                            (   (parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ])
+                            |    eps [ _d = AFFIL_ANY ]
                             )
                             [ _val = new_<Effect::GenerateSitRepMessage>(_a, _b, _c, _d, _e, _f) ]
                         )
@@ -124,7 +124,7 @@ namespace {
 
             string_and_string_ref_vector
                 =    ('[' > *string_and_string_ref [ push_back(_val, _1) ] > ']')
-                |    string_and_string_ref [ push_back(_val, _1) ]
+                |     string_and_string_ref [ push_back(_val, _1) ]
                 ;
 
             start
--- a/parse/EffectParser1.cpp
+++ b/parse/EffectParser1.cpp
@@ -78,7 +78,7 @@ namespace {
                         (   // empire id specified, optionally with an affiliation type:
                             // useful to specify a single recipient empire, or the allies
                             // or enemies of a single empire
-                            (   parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ]
+                            (   (parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ])
                             |   eps [ _d = AFFIL_SELF ]
                             )
                         >>  parse::label(Empire_token) > int_value_ref
@@ -98,7 +98,7 @@ namespace {
                         )
                     |   (   // no empire id or condition specified, with or without an
                             // affiliation type: useful to specify no or all empires
-                            (   parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ]
+                        (   (parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ])
                             |   eps [ _d = AFFIL_ANY ]
                             )
                             [ _val = new_<Effect::GenerateSitRepMessage>(_a, _b, _c, _d, _e, _f) ]
@@ -123,7 +123,7 @@ namespace {
                 ;
 
             string_and_string_ref_vector
-                =    '[' > *string_and_string_ref [ push_back(_val, _1) ] > ']'
+                =    ('[' > *string_and_string_ref [ push_back(_val, _1) ] > ']')
                 |    string_and_string_ref [ push_back(_val, _1) ]
                 ;
 
--- a/parse/ShipDesignsParser.cpp
+++ b/parse/ShipDesignsParser.cpp
@@ -72,7 +72,7 @@ namespace {
                 =    design_prefix(_a, _b, _c, _f)
                 >    parse::label(Parts_token)
                 >    (
-                            '[' > +tok.string [ push_back(_d, _1) ] > ']'
+                    ('[' > +tok.string [ push_back(_d, _1) ] > ']')
                         |   tok.string [ push_back(_d, _1) ]
                      )
                 >   -(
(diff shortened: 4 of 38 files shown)


adrianbroher (commented):
The change itself looks good, but I don't know if this should be merged or if the nagging compiler should be silenced.

adrianbroher:
The equivalent of -Wparentheses for cl.exe should be C4554.

LGM-Doyle:
I assumed that that warning appearance (in both gcc and clang) indicated that someone wanted them fixed.

I agree one or the other:  turn off warning that freeorion does not consider useful or fix warnings that result in a noisy compile that can mask real problems.

The only warnings in the compile that are unavoidable are the warnings for the unused variables that are actually used to force initialization of static variables.

Do you want me to fix the build files?

adrianbroher:
> I assumed that that warning appearance (in both gcc and clang) indicated that someone wanted them fixed.

No, I just turned up the error reports to the highest warning level available in 6540904f.

> Do you want me to fix the build files?

Lets wait for further input before you waste your time.

dbenage-cx:
Personally prefer that, by default, such warnings be kept.  Having the user/dev specifically hide these warnings gives more awareness to any issue that may be hidden from them.
Any "erroneous" warnings should be in-line commented when possible.

geoffthemedio:
Not sure if I follow all this, but I would not disable unused variable warnings just because there are a few unneeded warnings about them from when they're being used intentionally.

adrianbroher:
@geoffthemedio 

> Not sure if I follow all this

@LGM-Doyle s PR implements a fix for the GCC and Clang -Wparentheses warning.  This is emitted whenever there are logical statements that may misinterpreted by the developer.  For example:

`ìf (a || b && c || d)` would trigger such message.  The idea here that explicit adding parentheses disambiguate the logic, even when the operator precedence already defines the order of resolution.  To make the compiler happy you would have to write: `if (a || (b && c) || d)`.  Keep in mind that a to d could be long expressions and the condition itself could span over several lines.  The C4554 warning would be the equivalent to -Wparenthesis on the cl compiler (Visual Studio).

geoffthemedio (changes requested):
Adding brackets to silence warnings is fine, but please don't change the indenting when doing so

geoffthemedio:
Apparently I did follow all of it.

LGM-Doyle:
@geoffthemedio could you point out where I botched the indenting?  I looked but wasn't sure what needed fixing and what did not.

@dbenage-cx I also do not want to reduce/turn off warnings, that is why I fixed them in this PR.

However, there is one case where fixing the warnings reduced the code quality.  

Fixing this warning prevents defining a variable in the conditional clause of a conditional statement, because it requires an extra set of brackets.  It is not possible to lexically limit the scope of the defined variable to the body of the conditional clause.  The "fixed" versions of this structure leaves a known null variable floating in the namespace.

This example old version generates warnings prompting to put an additional bracket around the definition which C++ doesn't allow:

```
  `if(bool_convertible_type maybe_usable = FunctionMaybeReturningGoodValue()) {
         // code that knows maybe_usable is good
    }
    // code that can't see maybe_usable so can't misuse it`
```

The "fixed" version with the definition outside of the conditional clause generates no warnings, but allows the known bad variable to be used outside of the conditional clause.

```
  `bool_convertible_type maybe_usable = FunctionMaybeReturningGoodValue());
    if(maybe_usable) {
         // code that knows maybe_usable is good
    }
     // code that can misuse maybe_usable`
```

parse/SpeciesParser.cpp
```diff
@@ -76,7 +76,7 @@ namespace {
             foci
                 =    parse::label(Foci_token)
                 >    (
-                            '[' > +focus_type [ push_back(_r1, _1) ] > ']'
+                    ('[' > +focus_type [ push_back(_r1, _1) ] > ']')
```
geoffthemedio: here

parse/SpeciesParser.cpp
```diff
@@ -92,7 +92,7 @@ namespace {
                 ;
 
             environment_map
-                =    '[' > +environment_map_element [ insert(_val, _1) ] > ']'
+                =    ('[' > +environment_map_element [ insert(_val, _1) ] > ']')
```
geoffthemedio, line 95: here

parse/TechsParser.cpp
```diff
@@ -116,14 +116,14 @@ namespace {
 
             prerequisites
                 =   parse::label(Prerequisites_token)
-                >  (    '[' > +tok.string [ insert(_r1, _1) ] > ']'
+                >  (    ('[' > +tok.string [ insert(_r1, _1) ] > ']')
```
geoffthemedio, line 119: here

parse/TechsParser.cpp
```diff
@@ -116,14 +116,14 @@ namespace {
...
                     |   tok.string [ insert(_r1, _1) ]
                    )
                 ;
 
             unlocks
                 =   parse::label(Unlock_token)
-                >  (    '[' > +parse::detail::item_spec_parser() [ push_back(_r1, _1) ] > ']'
+                >  (    ('[' > +parse::detail::item_spec_parser() [ push_back(_r1, _1) ] > ']')
```
geoffthemedio, line 126: here

parse/SpecialsParser.cpp
```diff
@@ -67,10 +67,10 @@ namespace {
                 ;
 
             spawn
-                =    (      parse::label(SpawnRate_token)   > parse::double_ [ _r1 = _1 ]
+                =    (      (parse::label(SpawnRate_token)   > parse::double_ [ _r1 = _1 ])
```
geoffthemedio, line 70: here

parse/ShipHullsParser.cpp
```diff
@@ -90,7 +90,7 @@ namespace {
                 =  -(
                         parse::label(Slots_token)
                     >   (
-                            '[' > +slot [ push_back(_r1, _1) ] > ']'
+                        ('[' > +slot [ push_back(_r1, _1) ] > ']')
```
geoffthemedio: very much here

parse/ShipDesignsParser.cpp
```diff
@@ -72,7 +72,7 @@ namespace {
                 =    design_prefix(_a, _b, _c, _f)
                 >    parse::label(Parts_token)
                 >    (
-                            '[' > +tok.string [ push_back(_d, _1) ] > ']'
```
geoffthemedio, line 75: extremely here

geoffthemedio (changes requested):
parser indenting problems, incomplete list

parse/TechsParser.cpp
LGM-Doyle: What is supposed to be aligning with what here?  Is the newly bracketed clause supposed to be on its own line?

geoffthemedio: the start of each 'line' of parser code should match up with the start of the other lines at the same level of nesting, as though the ( ) weren't there. so

```
                >  (    '[' > +tok.string [ insert(_r1, _1) ] > ']'
                    |   tok.string [ insert(_r1, _1) ]
                   )
```

could become

```
                >  (   ('[' > +tok.string [ insert(_r1, _1) ] > ']')
                    |   tok.string [ insert(_r1, _1) ]
                   )
```

LGM-Doyle: Ok, I'll try.

I do not think that adding brackets to the parser code increases its clarity.  That is why I put all the parser stuff in a separate commit.  

I almost mentioned it specifically in the PR because the `>` operator is not a comparison operator so the warning itself is misleading.

LGM-Doyle:
Hopefully all fixed?

parse/TechsParser.cpp
geoffthemedio: Nor do I... I'd probably have added some if I did (and actually, I probably have in a few places). But I'd rather add brackets than deal with disabling warnings and resulting quirks or maintenance problems.

GG/GG/utf8/checked.h
```diff
@@ -120,13 +120,13 @@ namespace utf8
         }
         else if (cp < 0x10000) {              // three octets
             *(result++) = static_cast<uint8_t>((cp >> 12)           | 0xe0);
-            *(result++) = static_cast<uint8_t>((cp >> 6) & 0x3f     | 0x80);
+            *(result++) = static_cast<uint8_t>(((cp >> 6) & 0x3f)   | 0x80);
```
geoffthemedio, line 123: Are you sure these changes doesn't break anything? I don't know, but I'd have avoided messing with stuff like this...

UI/ObjectListWnd.cpp
```diff
@@ -1392,7 +1392,7 @@ class ObjectPanel : public GG::Control {
 
         for (std::size_t i = 1; i < m_controls.size(); ++i) {
             right = left + GetColumnWidth(static_cast<int>(i));
-            if (ctrl = m_controls[i])
+            if ((ctrl = m_controls[i]))
```
geoffthemedio, line 1395: how is an extra layer of () any different / better here?

parse/EffectParser1.cpp
```diff
@@ -98,7 +98,7 @@ namespace {
                         )
                     |   (   // no empire id or condition specified, with or without an
                             // affiliation type: useful to specify no or all empires
-                            (   parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ]
+                        (      (parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ])
                             |   eps [ _d = AFFIL_ANY ]
                             )
```
geoffthemedio, line 103: this ) should line up with its opening (

parse/EffectParser1.cpp
```diff
@@ -98,7 +98,7 @@ namespace {
                         )
                     |   (   // no empire id or condition specified, with or without an
                             // affiliation type: useful to specify no or all empires
-                            (   parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ]
+                        (      (parse::label(Affiliation_token) > parse::enum_parser<EmpireAffiliationType>() [ _d = _1 ])
```
geoffthemedio: this ( should be indented relative to its inclosing (

universe/System.cpp
```diff
@@ -501,7 +501,7 @@ bool System::RemoveStarlane(int id) {
 
 bool System::RemoveWormhole(int id) {
     bool retval = false;
-    if (retval = HasWormholeTo(id)) {
+    if ((retval = HasWormholeTo(id))) {
```
geoffthemedio, line 504: again, why?

universe/ValueRef.cpp
```diff
@@ -632,8 +632,8 @@ double Variable<double>::Eval(const ScriptingContext& context) const
     IF_CURRENT_VALUE(float)
 
     if (m_ref_type == NON_OBJECT_REFERENCE) {
-        if (property_name == "UniverseCentreX" |
-            property_name == "UniverseCentreY")
+        if ((property_name == "UniverseCentreX") |
```
geoffthemedio, line 635: I also question the necessity of wrapping these in () as well...

UI/ObjectListWnd.cpp
LGM-Doyle: The changes in this PR should not generate different code.  They only make the spurious compiler warnings go away.

The compiler wants the two operations, assignment to `ctrl` and testing of the boolean clause in separately bracketed clauses.  Hence the extra set of brackets.

parse/EffectParser1.cpp
LGM-Doyle: Fixed

universe/Condition.cpp
```diff
@@ -6606,8 +6606,8 @@ void WithinDistance::Eval(const ScriptingContext& parent_context,
                           ObjectSet& matches, ObjectSet& non_matches,
                           SearchDomain search_domain/* = NON_MATCHES*/) const
 {
-    bool simple_eval_safe = m_distance->LocalCandidateInvariant() &&
-                            parent_context.condition_root_candidate || RootCandidateInvariant();
+    bool simple_eval_safe = (m_distance->LocalCandidateInvariant() &&
```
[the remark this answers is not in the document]
LGM-Doyle: Fixed. 

That was incorrect in the original code. I went through this file and fixed two other cases of  that same pattern.

universe/ValueRef.cpp
LGM-Doyle: The change is just to remove the spurious warning, the compiler warning `-Wparentheses` wants the `&&` clause explicitly bracketed.

PR #1044, opened 2016-10-09
Document `util/XMLDoc.h` and trim file includes.

adrianbroher:
This PR updates the in-code documentation of `util/XMLDoc.h` by using Doxygen keywords, adding remarks, bug notes and todo notes and being more verbose in general.  Types, attributes or functions that were not documented are left out intentional.

I already applied a pass of aspell correction.

Also this PR cuts down the includes from files that were included by `util/XMLDoc.h` and files that included `util/XMLDoc.h` to reduce useless inclusion dependencies, which force up the compile time.  This required a minimal API change of `OptionsDB`.

### DIFF
--- a/UI/IntroScreen.cpp
+++ b/UI/IntroScreen.cpp
@@ -13,6 +13,7 @@
 #include "../util/OptionsDB.h"
 #include "../util/Serialize.h"
 #include "../util/Version.h"
+#include "../util/XMLDoc.h"
 
 #include <GG/GUI.h>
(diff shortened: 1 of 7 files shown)


dbenage-cx (changes requested):
cpp-apidoc/html/classXMLElement.html renders fine with no associated warnings

adrianbroher:
Thanks for the review @dbenage-cx.

dbenage-cx (approved):

PR #1046, opened 2016-10-09
Force the usage of boost::phoenix v3 in older boost versions.

adrianbroher:
As reported in this [thread](http://www.freeorion.org/forum/viewtopic.php?f=24&t=10276) and in commit c00f6f94c67504b1f761de1ad0ccb55270386c16 the added line and file logger attributes pull in a boost::log header, which in turn uses boost::phoenix v3.  The current compiler code however uses boost::spirit::phoenix v2 or v3 depending on the boost version.  This may cause a type collision between those two phoenix versions (thanks to @Hacklin for pointing this out).

To avoid this issue I defined  `BOOST_SPIRIT_USE_PHOENIX_V3` in the CMakeLists.txt for boost versions older than 1.56 (thanks to @dbenage-cx for researching this) and I removed some boost version conditional code, which defines the return type for rules.  This branch now compiles properly with recent boost version and with older boost version (tested on debian stable with boost 1.55).

@Hacklin: Could you please test if this fixes your problem?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,10 @@ add_subdirectory(GG)
 set(CMAKE_INSTALL_LIBDIR ${_ORIG_CMAKE_INSTALL_LIBDIR})
 unset(_ORIG_CMAKE_INSTALL_LIBDIR)
 
+add_definitions(
+    -DBOOST_SPIRIT_USE_PHOENIX_V3
+)
+
 if(${Boost_VERSION} EQUAL "106100")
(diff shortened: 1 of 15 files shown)


Hacklin:
Fixed.

I can build, run and play FreeOrion with Boost v1.54.

Thanx.

Hacklin (approved):
Fixed.
I can build, run and play FreeOrion with Boost v1.54.
Thanx.

PR #1050, opened 2016-10-13
Improved stringtable entries about ancient guardians self-destruction.

Deepsloth:


Deepsloth:
Inspired by http://www.freeorion.org/forum/viewtopic.php?f=28&t=10285

PR #993, opened 2016-09-26
Improve GG::ListBox and derived classes.

LGM-Doyle:
This PR improves the performance of `ListBox` based classes.

It refactors `ListBox` so that it will not layout rows that are not
visible.  It does this by adding `PreRender()` to both the `ListBox::Row`
and `ListBox` itself.  To use this functionality fully any classes derived
from `ListBox::Row` must do their initialization in their `PreRender()`.  

Additionally, the `ClientSize()` now correctly accounts for the scroll
bars.

This functionality replaces `ListBox::Row::AdjustLayout()`, which was a
toned down version of `PreRender()` that operated with some lexical scope
to suppress re-layout.  This cause of unpredictability when using
`ListBox::Row` because some operations where immediate and others not.

`ListBox` based classes are upgraded/fixed to work with the new functionality.

PR #1053, opened 2016-10-13
Allow SitRep rclick menu from majority of SitRep row

dbenage-cx:
Allows user to right click on most of a row to select from the context menu.
LinkText retains the original menu to either copy or open the link.


dbenage-cx:
@LGM-Doyle is #993 the only PR in that series touching ListBox?

LGM-Doyle:
Yes PR #993 contains all of the ListBox changes.

I tried this PR and it works for me.  Copy and paste from a SitRep to an editor works.

dbenage-cx:
Ty for testing

PR #1055, opened 2016-10-15
Add new turn sound option

### LINKED ISSUE #800: Beginning of the turn sound
It would be nice if the game would play a sound when a next turn begins. This would solve a problem of missing it while going away doing other things while other people are finishing their turns in a multiplayer game. This is partially solved by the sound of the spawning messages window, but some turns don't produce any messages and the window won't open.

dbenage-cx:
Adds options for a sound to play at start of new turn.

`help wanted` for default sound file

Resolves #800


adrianbroher:
Why would the button to toggle even be in the map window?

dbenage-cx:
Thought it might be desired for convenient access, I will move toggle to the Options menu.
If later desired a right click context on the turn button could work just as well.

dbenage-cx:
Rebased

Issue #971, opened 2016-09-12
Save game operation fails silently if user resigns and quits the game quickly enough


Vezzra:
### Environment
- **FreeOrion Version**: 0.4.6 RC1
- **Operating System**: OSX
- **Fetched as**
  - Binary release
### Description

Something I never noticed before, probably because I very rarely play games that get so big that the save game process takes long enough:

When you initiate a save, the operation returns immediately, but as you can observe by the messages in the message window, the save game operation actually takes some time to complete. However, you are already able to resign and quit the game, in which case the save game operation gets silently aborted.

If you're like me and hide the message window, you don't realize that the save hasn't completed yet - you play your game, want to take a break and resume later, so you save, resign, quit. When you get back, your savegame isn't there, because you quit too quickly.
### Expected Result

If you save a game, one of the following should prevent quitting the game too quickly for the operation to finish:
- Simply don't return control to the user until the save operation has finished.
- Disable all buttons in the game menu that would quit the current game ("Resign", "Load") until the operation has finished, also block all other operations that could quit the application (closing the window etc.).
### Steps to reproduce

Huh, tedious:
- Play a big game long enough to get a save that takes long to finish.
- Initiate a save.
- Resign and quit as quickly as possible.
- The save should not have been created.

Vezzra:
I tagged that for the release for now, because IMO that's a serious bug, but I want to put that decision up for a vote: do you guys agree, or should I remove the "Release v0.4.6" milestone and leave that as a "known issue" for 0.4.6?

geoffthemedio:
This is not a new observation, and is not appropriate to implement fixes for before a release. The save-related messages were added in part to resolve this (and if you're hiding the messages window, its your own fault for not seeing them...)

Vezzra:
Ok, I see. Sorry, didn't want to cause unnecessary fuss, but somehow I missed that the issue existed and the save game message is an attempt to address that.

If that's too complicated to address for the release I'll of course remove the milestone (will set it to "post 0.4.6"). 

> if you're hiding the messages window, its your own fault for not seeing them.

While that is true, the save game message solution is not really sufficient. Saving the game is a too important and essential operation to rely on the user paying attention to a message appearing in a small window in the corner of the screen. It should not be _possible_ to interrupt the operation like that, so I think the issue should remain open (unless there is already another issue about this which I overlooked?).

MatGB:
[...]

PR #1060, opened 2016-10-18
Use waiting cursor during long processing states

dbenage-cx:
Adds functionality to change the cursor to a waiting icon.
Uses this cursor during
- WaitingForTurnStart
- WaitingForTurnData
- between PlayingTurn SaveGameDataRequest and SaveGameComplete messages.

Functionality is designed to not clear a waiting cursor until all requests for that cursor are resolved.
`ResetCursorToDefault()` is provided for cases of an immediate need for the default cursor.

Related #971

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -193,7 +193,9 @@ HumanClientApp::HumanClientApp(int width, int height, bool calculate_fps, const 
     m_game_started(false),
     m_connected(false),
     m_auto_turns(0),
-    m_have_window_focus(true)
+    m_have_window_focus(true),
+    m_default_cursor(0),
+    m_wait_cursor(0)
 {
 #ifdef ENABLE_CRASH_BACKTRACE
     signal(SIGSEGV, SigHandler);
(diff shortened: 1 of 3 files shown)


client/human/HumanClientApp.cpp
```diff
@@ -1238,3 +1243,42 @@ void HumanClientApp::OpenURL(const std::string& url) {
...
 }
+
+void HumanClientApp::AddWaitingFlag(std::string reason) {
+    if (reason.empty()) {
+        ErrorLogger() << "Requested addition of empty wait flag";
+        return;
+    }
+
+    TraceLogger() << "Request to add wait flag: " << reason;
+    m_wait_flags[reason]++;
```
geoffthemedio, line 1254: is this thread safe?
if any networking events can cause this to be called, it might need to be...

client/human/HumanClientApp.cpp
```diff
@@ -1238,3 +1243,42 @@ void HumanClientApp::OpenURL(const std::string& url) {
...
+
+    SetCursor(m_wait_cursor);
+}
+
+void HumanClientApp::ClearWaitingFlag(std::string reason) {
```
geoffthemedio, line 1259: why not const std::string&

client/human/HumanClientApp.h
```diff
@@ -74,6 +75,10 @@ class HumanClientApp :
     void                HandleSaveGameDataRequest();
 
     void                OpenURL(const std::string& url);
+
+    void                AddWaitingFlag(std::string reason);         ///< Ensure the wait cursor is used until \p reason is cleared
```
geoffthemedio, line 79: ...

geoffthemedio (changes requested):

client/human/HumanClientApp.cpp
dbenage-cx: @geoffthemedio Good point.
I am guessing that adding functionality to GG instead and using signal/slots would be preferred (and provide a better avenue for other cursor changes).
I will look into making such a change unless another avenue is suggested, I need to study up on GG::GUIImpl/atomic/SDL to determine how to approach.
Will close this PR as invalid and open as issue if that sounds like the correct course.

Vezzra:
@dbenage-cx, do you intend to come up with another PR with a different implementation that fixes the issues raised, or did you drop the idea entirely? In case of the former, the status needs to be changed to "follow up".

dbenage-cx:
@Vezzra I forgot to open an issue earlier, I will do so shortly.
While I am actively researching this, it may take me some time to start a new implementation.  If someone else has an interest in tackling it, I would not want to impede them from doing so.

I assume this would remain `invalid` in this case, since there is no current PR to link to (or should I link to an issue? I will hold off linking from one to the other for now).

Vezzra:
> I assume this would remain invalid in this case, since there is no current PR to link to

After re-reading the description for the "follow up" status label I have to agree - when applying that label, a reference to the follow-up PR should be provided, so as long as there is no follow-up PR, this propbably should be labelled invalid.

Just don't forget to set the status label to "follow up" once the new PR is up.

adrianbroher:
A new implementation addressing the wait cursor will be handled in #1062.

> After re-reading the description for the "follow up" status label I have to agree.

That's another corner case. The follow up shouldn't be bound to to a PR. I will change the description accordingly.

PR #1063, opened 2016-10-20
fix name conflict between system variable and sys module

ghost:
Currently, the AI uses 'sys' as a variable name, which overshadows the 'sys' module.  This causes errors when trying to write to sys.stderr.

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -191,7 +191,7 @@ def check_supply():
     # TODO consider that this should not be more then maximal value in empire.systemSupplyRanges
     supply_distance += 6  # should not be more than max value in supplyProjections
 
-    # we should not rely on constant here, for sys, supply in supplyProjections need to add systems in supply range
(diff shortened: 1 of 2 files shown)


default/python/AI/ColonisationAI.py
```diff
@@ -1406,7 +1406,7 @@ def __print_candidate_table(candidates, mission):
         col1 = Text('Score')
         col1_value = lambda x: x[0]
     else:
-        print >> sys.stderr, "__print_candidate_table(%s, %s): Invalid mission type" % (candidates, mission)
+        print >> system.stderr, "__print_candidate_table(%s, %s): Invalid mission type" % (candidates, mission)
```
Morlic-fo: `sys.stderr`

Morlic-fo (changes requested):
One call changed wrongly, otherwise seems fine on first sight.

default/python/AI/ColonisationAI.py
ghost: Good catch.  Fixed.

Morlic-fo (approved):

PR #1065, opened 2016-10-21
Add description of production quantities in pedia

ericfurman:
Add a line at 4356. Describes how to build multiple ships in Production queue.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -4352,6 +4352,8 @@ The Industry of a planet can be used for any production projects of [[encycloped
 
 Each project has a minimum number of turns required to build it. For example, if a project that cost 15 PPs, and had a minimum time of 3 turns, 5 PPs is the maximum that could be put towards its completion per turn.
 The project at the top of the queue is given PPs first. Any remaining PP are then given to the next project in the queue, and if there is still more remaining to the next-- and so on. You can re-prioritize the items in the queue by dragging the most important items closer to the top.
+
+When ships are added to the queue they have two drop down menus that enable you to build the same ship multiple times. The first is to build that ship the specified number of times. So if 5 is selected the shipyard will build one ship and then build another ship and so forth until it has built 5 ships. If the second one to the right is selected the shipyard will build that number of ships. Note: To change these values you must left click to open the drop down menu and then left click again on the desired value to change it. You can change the numbers with your mouse scroll wheel, but it will not be recognized.
 '''
 
 RESEARCH_TITLE


default/stringtables/en.txt
```diff
@@ -4352,6 +4352,8 @@ The Industry of a planet can be used for any production projects of [[encycloped
 
 Each project has a minimum number of turns required to build it. For example, if a project that cost 15 PPs, and had a minimum time of 3 turns, 5 PPs is the maximum that could be put towards its completion per turn.
 The project at the top of the queue is given PPs first. Any remaining PP are then given to the next project in the queue, and if there is still more remaining to the next-- and so on. You can re-prioritize the items in the queue by dragging the most important items closer to the top.
+
+When ships are added to the queue they have two drop down menus that enable you to build the same ship multiple times. The first is to build that ship the specified number of times. So if 5 is selected the shipyard will build one ship and then build another ship and so forth until it has built 5 ships. If the second one to the right is selected the shipyard will build that number of ships. Note: To change these values you must left click to open the drop down menu and then left click again on the desired value to change it. You can change the numbers with your mouse scroll wheel, but it will not be recognized.
```
geoffthemedio, line 4356: I can't find the discussion, but recently something like this was proposed for addition to the research article, in which there was discussion about the mechanics of requiring supply connects for refinements to upgrade ship weapon meters. This was vaguely related to research, but not really relevant to an article about research as a resource.

In this case, I'm a bit more conflicted. The article is really about industry as a meter / resource, which is probably OK to extend to discussion about how that resource is allocated on the queue. But getting into details of the interface seems off topic, and should probably be in a separate article about the UI specifically.

geoffthemedio (changes requested):

dbenage-cx:
Instead of first/second quantities, it might be good to define some terms here.
Internally, the second quantity is `blocksize`.
That term may bring more questions than answers, personally I think of an element as a batch (or maybe a work order).
So the first quantity might be `batch iterations`, and the second `batch size`.
(I'll refer to the queue element as batch here, substitute as needed if another term is used)
Some things not noted that may lead to user confusion/frustration:
- On completion of a batch, any additional iterations retain the queue position.
- After progress is made on a queue item, decreasing the batch size may result in lost PP.
- After progress, increasing the batch size will recalculate to the items required # of turns.
- The next batch iteration will not start production until the final turn of the current iteration.

That final point may be hard to explain concisely without an explanation.
e.g. Given 120 PP output and a ship needing 300 PP/3 turns (100/turn).
During turns 1-2 the remaining 20PP goes to the next item in the queue.
At the end the third turn, the next iteration receives 20PP (and the first iteration completes), the next item in the queue would not receive any PP.

default/stringtables/en.txt
ericfurman: Yes I thought about making this into a separate Pedia article, but this was easier.
It does seem like a good idea. I'll go ahead and start that.

ericfurman:
ok dbenage-cx, you gave me some info I was unaware of. I think your wording doesn't need much if any modification. only one thing...
While Batch has a well known and defined meaning in the computer world I think Work Order would be more understandable to the average person.

default/stringtables/en.txt
geoffthemedio: bump?

Vezzra:
> I think Work Order would be more understandable to the average person.

Or maybe (more specific) "Build Order"?

geoffthemedio:
Prefer "production" over "build", please.

Vezzra:
"Production Order", sure, even better 👍

dbenage-cx:
@ericfurman any update?

Vezzra:
Also rebase needed.

Vezzra:
There has been no more feedback from @ericfurman since last October. What shall we do with this PR?

dbenage-cx:
Closing for followup
@ericfurman  Thank you for the initial work on this.

Vezzra:
Superceded by #1354.

PR #1074, opened 2016-10-27
Do not increase PLANETS_DEPOPULATED stat for Ancient Guardians

dbenage-cx:
As a special case, Ancient Guardians should not increase the Own Planets Depopulated stat after capture.


Vezzra:
Sounds a looks very reasonable, and is a fairly simple patch. Anyone any objections to merging this?

geoffthemedio:
looks reasonable.

PR #1075, opened 2016-10-27
Allow industry on outposts (pop center change)

agrrr3:
Before, all population centers with close to zero population were
depopulated and in turn got the industry reset to zero.
As a outpost is a population center with zero population, it was not
possible to have industry at an outpost. 

I need this functionality right now because I scripted a branch for supply ships which do not work on outposts without this. I was told to create this PR
Discussion about this and supply ships in the forums at 
http://www.freeorion.org/forum/viewtopic.php?f=15&t=10152&p=86319#p86319

To be able to build at supply line disconnected outposts, the code now
looks if there is a species on the planet. If it is not, it is considered
an outpost and the industry does not get set to zero.

### DIFF
--- a/universe/PopCenter.cpp
+++ b/universe/PopCenter.cpp
@@ -123,6 +123,10 @@ void PopCenter::PopCenterPopGrowthProductionResearchPhase() {
 
     //if (cur_pop > 0.0)
     //    DebugLogger() << "Planet Pop: " << cur_pop << " growth: " << pop_growth;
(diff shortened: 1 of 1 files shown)


agrrr3:
After a hint from geoff I found that outposts now produce research and industry, which is not intended.
I still would like to implement this feature but would need to refactor (which would improve the current code quite a bit, i think).
I am not sure where the discussion for that should happen. Should I close the pull request?

Issue #1073, opened 2016-10-26
Crash on opening combat log


MatGB:
### Environment
- **FreeOrion Version**: Fighters branch rebased with master
- **Operating System**: Kubuntu
- **Graphic card used**: AMD A78
- **Fetched as** 
  - Compiled from source
### Description

Reading turn sitrep, clicked combat report, game crashed.
### Steps to reproduce

Unsure as yet, not replicated on reload, didn't happen earlier in game. Logs attached.
[freeorion.zip](https://github.com/freeorion/freeorion/files/552073/freeorion.zip)

MatGB:
OK, this might be a Fighters branch specific, while the game didn't crash this time it did get horrendously laggy while I tried to read the report, and the report itself is messed up in really bad ways (large number of fighters launched by both sides).

Savegame attached, for compatibility, Fighters branch rebased to
[FreeOrion_Mat_Gotham_0179_20161026_030211.sav.zip](https://github.com/freeorion/freeorion/files/552205/FreeOrion_Mat_Gotham_0179_20161026_030211.sav.zip)
 Master as at 4f3619b

MatGB:
@LGM-Doyle , you're the most recent person to have done major work on the combat log, can you give this a look, think it's a Fighter compatibility issue with your stuff.

LGM-Doyle:
Yes I will look at this.

Could you provide a more detailed list of the problems with the combat report?  An itemized list will mean that I have at least looked at every item on the list, even if I am unable to improve the situation.

MatGB:
OK, screenshot attached, I expanded a set of results, as you can see the fighters are listed twice, independently of the collapsing per ship results and within it, plus there's a lot of "unknown" in there.

Loading the report at all takes significant time, even on relatively small fights, if either/both sides is using fighters (and as I'm deliberately testing fighters I am).

The scrolling down to read further results is very laggy, it's not all pregenerated, and if you get to the bottom you sometimes need to resize the report window to get the scrollbar to let you go down further.

For **really** large fights, the end of it either doesn't generate properly or sometimes at all, and you get some weird artifacts displaying instead of actually rendered font: loading such a fight has now caused at least three identical-in-style crashes for me.

[...]

PR #1076, opened 2016-10-27
Improve combat log with deferred layout.

LGM-Doyle:
This PR addresses some of the issues in #1073.  

For my combat window with fighters test case it improved layout time from 12 seconds to 1 second.

It does not fix any of the `Uknown` labels that should be known, or any other issues with the combat log.

It depends on the newly merged into master `DeferredLayout` class, so the Fighter branch will need to be rebased to master to see the improvements from this PR.

### DIFF
--- a/UI/CombatReport/CombatLogWnd.cpp
+++ b/UI/CombatReport/CombatLogWnd.cpp
@@ -140,7 +140,7 @@ namespace {
 
         SetBorderMargin(BORDER_MARGIN);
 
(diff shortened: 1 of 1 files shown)


Vezzra:
The changes are quite trivial and straightforward, and tests with the fighters branch apparently worked ok. So, are there any objections to merging this?

geoffthemedio:
nope

PR #1079, opened 2016-10-28
Refactor get_fleets_for_mission

Morlic-fo:
Previous implementation was some weird iterative-recursion-hybrid. It is now a purely iterative implementation of the same algorithm.

Various arguments were not used or always called the same way. I removed that unused functionality for an overall much simpler and more readable function. If needed in the future, the functionality may be trivially reimplemented. 

The PR is opened to fighter branch rather than master to avoid conflicts as the function is changed in Fighter branch. Probably will be needed for next set of fighter-related AI updates and I prefer working with a clean codebase...

### DIFF
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -69,7 +69,7 @@ def get_targeted_planet_ids(planet_ids, mission_type):
 
 
 def get_fleets_for_mission(target_stats, min_stats, cur_stats, starting_system,
-                           fleet_pool_set, fleet_list, species="", take_any=False):
+                           fleet_pool_set, fleet_list, species=""):
     """
     Implements breadth-first search through systems
     mutates cur_stats with running status, systems_to_check and systems_checked as systems are checked,
@@ -137,8 +137,7 @@ def get_fleets_for_mission(target_stats, min_stats, cur_stats, starting_system,
             )):
                 systems_to_check.append(neighbor_id)
     # we ran out of systems or fleets to check but did not meet requirements yet.
-    if take_any or (stats_meet_reqs(cur_stats, min_stats)
-                    and any(universe.getFleet(fid).shipIDs for fid in fleet_list)):
+    if stats_meet_reqs(cur_stats, min_stats) and any(universe.getFleet(fid).shipIDs for fid in fleet_list):
         return fleet_list
     else:
         return []
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -68,7 +68,8 @@ def get_targeted_planet_ids(planet_ids, mission_type):
     return targeted_planets
 
 
-# TODO: as used, cur_stats is  just another return value
+# TODO: Avoid mutable arguments and use return values instead
+# TODO: Use Dijkstra's algorithm instead of BFS to consider starlane length
 def get_fleets_for_mission(target_stats, min_stats, cur_stats, starting_system,
                            fleet_pool_set, fleet_list, species=""):
     """Get fleets for a mission.
@@ -112,7 +113,7 @@ def get_fleets_for_mission(target_stats, min_stats, cur_stats, starting_system,
         while fleets_here:
             fleet_id = fleets_here.pop(0)
             fleet = universe.getFleet(fleet_id)
-            if not fleet:
+            if not fleet:  # TODO should be checked before passed to the function
                 fleet_pool_set.remove(fleet_id)
                 continue
             # try splitting fleet
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -69,8 +69,7 @@ def get_targeted_planet_ids(planet_ids, mission_type):
 
 
 def get_fleets_for_mission(nships, target_stats, min_stats, cur_stats, species, systems_to_check, systems_checked,
-                           fleet_pool_set, fleet_list, take_any=False, extend_search=True,
-                           verbose=False, depth=0):
+                           fleet_pool_set, fleet_list, take_any=False, extend_search=True):
     """
     Implements breadth-first search through systems
     mutates cur_stats with running status, systems_to_check and systems_checked as systems are checked,
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -68,8 +68,8 @@ def get_targeted_planet_ids(planet_ids, mission_type):
     return targeted_planets
 
 
-def get_fleets_for_mission(nships, target_stats, min_stats, cur_stats, species, systems_to_check, systems_checked,
-                           fleet_pool_set, fleet_list, take_any=False, extend_search=True):
+def get_fleets_for_mission(nships, target_stats, min_stats, cur_stats, systems_to_check, systems_checked,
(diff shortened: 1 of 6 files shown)


default/python/AI/FleetUtilsAI.py
```diff
@@ -68,107 +68,97 @@ def get_targeted_planet_ids(planet_ids, mission_type):
...
+
+    Implements breadth-first search through systems starting at the **starting_sytem**.
+    In each system, local fleets are checked if they are in the allowed **fleet_pool_set** and suitable for the mission.
+    If so, they are added to the **fleet_list** and **cur_stats** is updated with the currently selected fleet summary.
+    The search continues until the requirements defined in **target_stats** are met or there are no more systems/fleets.
+    In that case, if the **min_stats** are covered, the **fleet_list** is returned anyway.
     Otherwise, an empty list is returned by the function, in which case the caller can make an evaluation of
     an emergency use of the found fleets in fleet_list; if not to be used they should be added back to the main pool.
+
+    :param target_stats: stats the fleet should ideally meet
```
Cjkjvfnby, line 85: may be rename it to `requirements`

default/python/AI/FleetUtilsAI.py
```diff
@@ -68,107 +68,97 @@ def get_targeted_planet_ids(planet_ids, mission_type):
...
+    while systems_to_check and fleet_pool_set:
+        this_system_id = systems_to_check.pop(0)
+        systems_checked.append(this_system_id)
+        accessible_fleets = foAI.foAIstate.systemStatus.get(this_system_id, {}).get('myFleetsAccessible', [])
+        fleets_here = [fid for fid in accessible_fleets if fid in fleet_pool_set]
+        # loop over all fleets in the system, split them if possible and select suitable ships
+        while fleets_here:
+            fleet_id = fleets_here.pop(0)
+            fleet = universe.getFleet(fleet_id)
+            if not fleet:
```
Cjkjvfnby: Move this check outside this function.  Lets store only valid idis.

default/python/AI/FleetUtilsAI.py
```diff
@@ -68,107 +68,97 @@ def get_targeted_planet_ids(planet_ids, mission_type):
...
-        needs_troops = 'troopCapacity' in target_stats
-        if needs_troops:
-            troop_capacity = count_troops_in_fleet(fleet_id)
-
-        use_fleet = (((not needs_species and not has_species) or meets_species_req) and
-                     ((needs_troops and troop_capacity > 0) or (not needs_troops and not troop_capacity)))
-        if use_fleet:
+    colonization_roles = (ShipRoleType.CIVILIAN_COLONISATION, ShipRoleType.BASE_COLONISATION)
+    systems_to_check = [starting_system]
+    systems_checked = []
```
Cjkjvfnby: For BFS I usually call this `visited` but up to you.

default/python/AI/FleetUtilsAI.py
```diff
@@ -68,107 +68,97 @@ def get_targeted_planet_ids(planet_ids, mission_type):
...
-        troop_capacity = 0
-        needs_troops = 'troopCapacity' in target_stats
-        if needs_troops:
-            troop_capacity = count_troops_in_fleet(fleet_id)
-
-        use_fleet = (((not needs_species and not has_species) or meets_species_req) and
-                     ((needs_troops and troop_capacity > 0) or (not needs_troops and not troop_capacity)))
-        if use_fleet:
+    colonization_roles = (ShipRoleType.CIVILIAN_COLONISATION, ShipRoleType.BASE_COLONISATION)
+    systems_to_check = [starting_system]
```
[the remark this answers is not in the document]
Morlic-fo: I don't think a `set` would work for the BFS as order isn't preserved?

default/python/AI/FleetUtilsAI.py
Morlic-fo: It is not strictly required. As long as `min_stats` are fulfilled, it will still be accepted...

I do agree however, that the old name was not ideal either. I will think about a better name.

default/python/AI/FleetUtilsAI.py
Morlic-fo: I simply used the previous variable name without much thought. `visited` seems fine to me.

default/python/AI/FleetUtilsAI.py
```diff
@@ -68,107 +68,97 @@ def get_targeted_planet_ids(planet_ids, mission_type):
...
-    mutates cur_stats with running status, systems_to_check and systems_checked as systems are checked,
-    fleet_pool_set as fleets are checked also mutates fleet_list as running list of selected fleets;
-    this list will be returned as the function return value if the target statsare met or if upon exhausting
-    systems_to_check both take_any is true and the min stats are met.
+# TODO: as used, cur_stats is  just another return value
+def get_fleets_for_mission(target_stats, min_stats, cur_stats, starting_system,
+                           fleet_pool_set, fleet_list, species=""):
+    """Get fleets for a mission.
+
+    Implements breadth-first search through systems starting at the **starting_sytem**.
```
Cjkjvfnby, line 77: What about Dijkstra's algorithm by starlane length?

```
1 -------- 2
|
3-4
```

default/python/AI/FleetUtilsAI.py
```diff
@@ -68,107 +68,97 @@ def get_targeted_planet_ids(planet_ids, mission_type):
...
+    The search continues until the requirements defined in **target_stats** are met or there are no more systems/fleets.
+    In that case, if the **min_stats** are covered, the **fleet_list** is returned anyway.
     Otherwise, an empty list is returned by the function, in which case the caller can make an evaluation of
     an emergency use of the found fleets in fleet_list; if not to be used they should be added back to the main pool.
+
+    :param target_stats: stats the fleet should ideally meet
+    :type target_stats: dict
+    :param min_stats: minimum stats the final fleet must meet to be accepted
+    :type min_stats: dict
+    :param cur_stats: (**mutated**) stat summary of selected fleet
```
Cjkjvfnby: -> `selected fleets`

default/python/AI/FleetUtilsAI.py
```diff
@@ -68,107 +68,97 @@ def get_targeted_planet_ids(planet_ids, mission_type):
...
+    :type cur_stats: dict
+    :param starting_system: system_id where breadth-first-search is centered
+    :type starting_system: int
+    :param fleet_pool_set: (**mutated**) fleets allowed to be selected. Split fleed_ids are added, used ones removed.
```
Cjkjvfnby, line 93: I think about avoiding mutable arguments:

return  None|FullyMatched|MinimumMatched , stats, fleets

Morlic-fo: Both `fleet_list` and `fleet_pool_set` are used in some parts of the code. In particular, `fleet_list` is used eventhough an empty list is returned. The use is really messy and definitely deserves another update in the future.

For now, this PR is aimed not to change other parts of the code (except for avoiding unneeded arguments).
But I agree that we should change that function in the future to not use mutable arguments and will add a TODO.

default/python/AI/FleetUtilsAI.py
Morlic-fo: Seems like a good and reasonable change for the future but beyond the scope of this PR.

Cjkjvfnby:
Looks ok.

PR #1083, opened 2016-10-30
Fix combat log accordion bad initial reflow.

LGM-Doyle:
PR #1076 re-exposed the problem with the initial text reflow of the combat accordion window title.

I have used the same workaround as before #1076 added `DeferredLayout`, but moved it into the `PreRender()` function.  

If I find a proper fix, while fixing the other combat log problems with fighters I will replace this code, but this PR is an improvement on #1076.

### DIFF
--- a/UI/CombatReport/CombatLogWnd.cpp
+++ b/UI/CombatReport/CombatLogWnd.cpp
@@ -1,5 +1,6 @@
 #include "CombatLogWnd.h"
 
+#include <GG/GUI.h>
 #include <GG/DeferredLayout.h>
 #include <GG/Scroll.h>
 #include <GG/ScrollPanel.h>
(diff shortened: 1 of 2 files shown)


UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -464,3 +462,25 @@ GG::Pt CombatLogWnd::MinUsableSize() const
 
 void CombatLogWnd::HandleMadeVisible()
 { return pimpl->HandleWndChanged(); }
+
+void CombatLogWnd::PreRender()
+{
```
geoffthemedio: { to prev line

geoffthemedio (changes requested):

UI/CombatReport/CombatLogWnd.cpp
LGM-Doyle: Fixed.

geoffthemedio (approved):

PR #1084, opened 2016-11-01
New graphics for self-gravitating hull

Captain-HM:


geoffthemedio:
I believe pull requests with merges from master will always be rejected.

Captain-HM:
OK, finally managed to get rid of it by forced push to the remote repository...

Vezzra:
@Captain-HM, can you provide the source files via PR for the assets repo (same way you already did in the past for your other models)?

Issue #1086, opened 2016-11-04
Stack Overflow When Resizing SidePanel


geoffthemedio:
Start game, expand the population, resource, buildings, and military panels on the sidepanel. (Happened for me with just the bottom two open, but not with none open or just the bottom). Start resizing the sidepanel by dragging the bottom-right around. After a few seconds, FreeOrion crashes. Attached debugger indicates this is a stack overflow. Top of the stack is similar to:
````
>	GetResourceDir() Line 424	C++
 	ClientUI::ArtDir() Line 62	C++
 	AccordionPanel::SetCollapsed(bool collapsed=false) Line 94	C++
 	BuildingsPanel::DoLayout() Line 251	C++
 	AccordionPanel::SizeMove(const GG::Pt & ul={...}, const GG::Pt & lr={...}) Line 89	C++
 	GG::Wnd::Resize(const GG::Pt & sz={...}) Line 538	C++
 	BuildingsPanel::DoLayout() Line 247	C++
 	AccordionPanel::SizeMove(const GG::Pt & ul={...}, const GG::Pt & lr={...}) Line 89	C++
 	SidePanel::PlanetPanel::DoLayout() Line 963	C++
 	boost::signals2::detail::variadic_slot_invoker<boost::signals2::detail::void_type>::operator()<boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<enum boost::signals2::detail::slot_meta_group,boost::optional<int> >,boost::signals2::slot<void __cdecl(void),boost::function<void __cdecl(void)> >,boost::signals2::mutex> > >(const boost::shared_ptr<boost::signals2::detail::connection_body<std::pair<enum boost::signals2::detail::slot_meta_group,boost::optional<int> >,boost::signals2::slot<void __cdecl(void),boost::function<void __cdecl(void)> >,boost::signals2::mutex> > & connectionBody={...}) Line 120	C++
[...]

PR #1088, opened 2016-11-06
Fix bug1086 resizing side panel causes crash

LGM-Doyle:
This PR fixes issue #1086, by preventing the stack overflow.

This section is copied from the third commit.

```Several infinitely recursive loops passing through
SidePanel::PlanetPanelContainer::DoPanelsLayout() were possible.

In one PlanetPanelContainer::DoPanelsLayout() would call
PlanetPanel::DoLayout(), which would signal ResizedSignal(), which would
call DoPanelsLayout() in an infinite loop.

A second loop passed through AccordionPanel::SetCollapsed(), which even
when the collapsed state was unchanged still signaled with the
ExpandCollapseSignal().
```

The third commit is possibly the only commit necessary to stop SidePanel having a stack overflow.

The second commit prevents AccordionPanel from doing unnecessary work and potentially causing other unnecessary work.

The first commit is required since currently SidePanel expects the unnecessary signal from AccordionPanel before it correctly sizes the focus drop down icons.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -607,7 +607,7 @@ private:
 
     GG::Scroll*                 m_vscroll; ///< the vertical scroll (for viewing all the planet panes)
 
-    bool                        m_ignore_recurisve_resize;
+    bool                        m_ignore_recursive_resize;
 };
 
 class RotatingPlanetControl : public GG::Control {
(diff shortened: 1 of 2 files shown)


UI/SidePanel.cpp
```diff
@@ -606,6 +606,8 @@ class SidePanel::PlanetPanelContainer : public GG::Wnd {
                                 m_valid_selection_predicate;
 
     GG::Scroll*                 m_vscroll; ///< the vertical scroll (for viewing all the planet panes)
+
+    bool                        m_ignore_recurisve_resize;
```
Vezzra: Typo: recurisve => recursive

Vezzra (changes requested):

geoffthemedio:
otherwise ok

LGM-Doyle:
I've fixed the spelling error.

Vezzra (approved):

LGM-Doyle:
Thanks for the review.

PR #1090, opened 2016-11-06
Industrious outposts minimal

agrrr3:
This is a minimalistic/KISS implementation for enabling production on outposts again.

What it does:
* Add a minimal population condition for all effects which add a fixed amount to research or production regardless 
* Stop depopulating planets which are already depopulated (i.e. have no species)
 
All other effects are not important, because of one of the following reason
* effect uses a FOCUS condition
* effect already has a condition which checks if there is population (e.g. minimum population, species, if there is a good environment...)
* effect adds some multiple of population (so zero pop means zero effect)
* building is cultural archives, cultural library, automated history analyzer which probably should be destroyed on depopulation (ATM cultural archives and automated history analyzer are not)
* effect changes supply (which is the same for outposts and colonies)

I checked for research, industry and happiness effects.

### DIFF
--- a/universe/PopCenter.cpp
+++ b/universe/PopCenter.cpp
@@ -117,6 +117,11 @@ void PopCenter::PopCenterResetTargetMaxUnpairedMeters() {
 }
 
 void PopCenter::PopCenterPopGrowthProductionResearchPhase() {
+	if (m_species_name.empty()) {
+		// No changes to population or happiness
+		return;
+	}
+
     float cur_pop = CurrentMeterValue(METER_POPULATION);
     float pop_growth = NextTurnPopGrowth(); // may be negative
     float new_pop = cur_pop + pop_growth;
--- a/universe/PopCenter.cpp
+++ b/universe/PopCenter.cpp
@@ -117,10 +117,10 @@ void PopCenter::PopCenterResetTargetMaxUnpairedMeters() {
 }
 
 void PopCenter::PopCenterPopGrowthProductionResearchPhase() {
-	if (m_species_name.empty()) {
-		// No changes to population or happiness
-		return;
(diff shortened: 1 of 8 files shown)


universe/PopCenter.cpp
```diff
@@ -117,6 +117,11 @@ void PopCenter::PopCenterResetTargetMaxUnpairedMeters() {
 }
 
 void PopCenter::PopCenterPopGrowthProductionResearchPhase() {
+	if (m_species_name.empty()) {
+		// No changes to population or happiness
```
geoffthemedio: replace tabs with 4 spaces

geoffthemedio (changes requested):

geoffthemedio:
> - effect adds some multiple of population (so zero pop means zero effect)

I'd still add populating gating for such effects, to avoid they appearing in the effect accounting tooltip with 0 change to the meter.

> - building is cultural archives, cultural library, automated history analyzer which probably should be destroyed on depopulation (ATM cultural archives and automated history analyzer are not)

This decisions are debatable and rather separate from the point of this pull request, and not yet made regardless, so these effects should be similarly gated by population.

agrrr3:
Hi Geoff, 

~~if I am gonna gate more effects, should I use a macro like I did in my first try (https://github.com/freeorion/freeorion/compare/master...agrrr3:industrious-outposts)
or not?~~ (i did it without macro now).

I am positive, that all effects are gated now.

geoffthemedio:
Doesn't seem substantial enough to warrant using a macro to me. I don't object if someone else feels strongly differently, though.

geoffthemedio:
otherwise looks fine...

PR #1091, opened 2016-11-08
Impose limits on population-dependent research and production bonuses.

geoffthemedio:
Imposes limits on the size of meter increases from various techs that have population-dependent bonuses. This means that bonuses grow with population, but only up to a point. This should reduce the range of output levels over the course of a game, and reduce the benefit of having high population planets.

The actual limits here are somewhat arbitrary and require balance consideration (along with the entire concept).

### DIFF
--- a/default/scripting/techs/learning/ALGO_ELEGANCE.focs.txt
+++ b/default/scripting/techs/learning/ALGO_ELEGANCE.focs.txt
@@ -13,7 +13,7 @@ Tech
                 Focus type = "FOCUS_RESEARCH"
             ]
             priority = [[EARLY_PRIORITY]]
-            effects = SetTargetResearch value = Value + Target.Population * 0.5 * [[RESEARCH_PER_POP]]
+            effects = SetTargetResearch value = Value + Max(6, Target.Population * 0.5 * [[RESEARCH_PER_POP]])
     ]
     graphic = "icons/tech/algorithmic_elegance.png"
 
--- a/default/scripting/techs/learning/ARTIF_MINDS.focs.txt
+++ b/default/scripting/techs/learning/ARTIF_MINDS.focs.txt
@@ -14,7 +14,7 @@ Tech
                 NOT TargetPopulation low = 0 high = 0
             ]
             priority = [[VERY_LATE_PRIORITY]]
-            effects = SetTargetResearch value = Value + 10 * [[RESEARCH_PER_POP]]
+            effects = SetTargetResearch value = Value + max(8, 10 * [[RESEARCH_PER_POP]])
     ]
     graphic = "icons/tech/basic_autolabs.png"
 
--- a/default/scripting/techs/learning/DISTRIB_THOUGHT.focs.txt
+++ b/default/scripting/techs/learning/DISTRIB_THOUGHT.focs.txt
@@ -15,7 +15,7 @@ Tech
                 // Focus type = "FOCUS_RESEARCH"  //Does not require research Focus
             ]
             priority = [[EARLY_PRIORITY]]
-            effects = SetTargetResearch value = Value + Target.Population * 0.5 * [[RESEARCH_PER_POP]]
+            effects = SetTargetResearch value = Value + Max(12, Target.Population * 0.5 * [[RESEARCH_PER_POP]])
     ]
     graphic = "icons/tech/distributed_thought.png"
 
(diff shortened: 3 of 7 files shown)


MatGB:
Not at all sure about the idea: I'd rather reduce the per-pop bonus for a fair few of them, which is what we've already discussed.

Getting techs and specials and boosts like Gaia, growth specials and similar are put of the fun of the game, fully maxing out a homeworld to get the best possible out of it is something I personally enjoy (and am very clear that at max the numbers are too high). Plus, we (well, I) have put a lot of effort in to balance the costs of improvements like terraforming and gaia transformation taking planet size into account of late.

If you implement something like this then the fun of trying to get your good planets truly great is reduced and there'd be a much stronger need to work out if a given improvement is worth it, in a way that I think might be detrimental: currently my decision on, for example, gaia transformation is whether the planet is already close to max output, but if it's at or close to max possible output then, unless that's VERY clearly displayed in the UI then working out if the building is worth it becomes far more complex/time consuming/micromangy/unfun.

Plus we did agree that we'd do a proper balance pass on these bonuses with/after Influence was implemented because doing it before then was basically a waste of time as it'd need doing again regardless.

There are definitely some elements that could be included for some techs, and I like the idea of the mechanic as a limiter for some things, but as a fairly blanket approach I'm not at all keen.

geoffthemedio:
Perhaps it would be better to just add this to some of the early-game tech effects, then? The result would be that early techs initially have a population-dependence, making growing planets useful at first, but later the bonus is capped so that those techs work like the fixed bonuses. Some other later-game techs could be uncapped, so there would still be reasons to continue increasing planet population in the later parts of the content tree.

Part of the point of adding this now would be to get a feel for how it affects balance.

Vezzra:
Um, as this is actually a design discussion, shouldn't that be done on the forum? I completely understand the intend to provide an experimental implementation of the proposed feature via a PR, but the corresponding design discussion shouldn't take place here.

I suggest in cases like this to open a corresponding thread on the forum, with the OP containing the proposal/description of the new feature and a link to the PR. Likewise put a link in the PR to the discussion thread on the forum.

Use the thread on the forum for the design related discussion, and the discussion area of the PR for technical/implementation related discussion.

@geoffthemedio, @MatGB, if that's ok with you, can you create a respective thread on the forum for this, set the links on both sides and repeat your posts there? I've some thoughts I'd want to add to, but would prefer to do so on the forum...

LGM-Doyle:
A fix that I have not seen suggested, is to use only the top M highest producing of all N colonies.  If you set M = sqrt(N) then you get the following table.

#### Number of colonies contributing to empire production
- 1  -  3  colonies in empire uses the best  1 planet's production.
- 4  -  8  colonies in empire uses the sum of the best  2 planet's production.
- 9  - 15  colonies in empire uses the sum of the best  3 planet's production.
- 16 - 24  colonies in empire uses the sum of the best  4 planet's production.
- etc

Its benefits are:
- It is a simple to explain mechanic, you don't even need to mention the sqrt, just show the table.  
- The total production output is additive, no difficult mental math.  
- It steeply increases the cost of the next production increase later in the game, while not completely removing growth as a strategy.  
- It make taking large key planets more strategically significant.

@Vezzra, I know that his breaks the rules, but I thought it would be presumptuous of me to start the forum thread and I wanted to chip in my two cents.  You can definitely transfer my comments to the forum.

Vezzra:
I've opened a corresponding thread on the design forum:

http://freeorion.org/forum/viewtopic.php?f=6&t=10325

Please continue all design related discussion there.

default/scripting/techs/learning/ALGO_ELEGANCE.focs.txt
```diff
@@ -13,7 +13,7 @@ Tech
                 Focus type = "FOCUS_RESEARCH"
             ]
             priority = [[EARLY_PRIORITY]]
-            effects = SetTargetResearch value = Value + Target.Population * 0.5 * [[RESEARCH_PER_POP]]
+            effects = SetTargetResearch value = Value + Max(6, Target.Population * 0.5 * [[RESEARCH_PER_POP]])
```
dbenage-cx, line 16: `max` -> `min`? (and in other cases)

adrianbroher:
Did this ever came to a conclusion?  This branch is now over a year old.

Vezzra:
> Did this ever came to a conclusion?

No. Although we might get back to these ideas. Inhowfar this PR might get "revived" I can't say (I cautiously doubt it), it could still have some value for playttesting, to get a feeling how such changes would affect gameplay.

@geoffthemedio?

geoffthemedio:
I commented on the forum thread.

PR #648, opened 2016-05-06
Transition to C++11 

LGM-Doyle:
This PR is to determine where freeorion is in its progress towards C++11 adoption.

I've added the code to the CMakeList.txt files to generate C++11 code.  I changed one file/object in each executable unit by adding a single C++11 only keyword/feature.

I'm hoping that devs will compile this on Windows and OSX and then post any error reports.  I will search for solutions to any problems that arise.

Thanks in advance for your time.

PR #1097, opened 2016-11-10
Transition to C++11

adrianbroher:
This PR is a follow up to #648, with updated Xcode project.

### DIFF
--- a/GG/GG/StaticGraphic.h
+++ b/GG/GG/StaticGraphic.h
@@ -95,7 +95,6 @@ public:
     //@}
 
 private:
-    void     Init(const SubTexture& subtexture); ///< initializes a StaticGraphic from a SubTexture
     void     ValidateStyle();   ///< ensures that the style flags are consistent
 
     SubTexture          m_graphic;
--- a/GG/src/RichText/TagParser.cpp
+++ b/GG/src/RichText/TagParser.cpp
@@ -26,7 +26,7 @@
 #include "TagParser.h"
 
 
-#include <boost/regex.hpp>
+#include <regex>
 
 namespace GG {
 
@@ -100,20 +100,20 @@ namespace GG {
                                                       std::vector<RichTextTag>* tags)
             {
                 std::string::const_iterator current = start;
-                boost::match_results<std::string::const_iterator> match;
-                boost::match_flag_type flags = boost::match_default;
+                std::match_results<std::string::const_iterator> match;
 
                 // The regular expression for matching begin and end tags. Also extracts parameters from start tags.
-                typedef boost::basic_regex<char, boost::regex_traits<char>> regex;
-                const static regex tag("<(?<begin_tag>\\w+)( "
-                                       "(?<params>[^>]+))?>|</"
-                                       "(?<end_tag>\\w+)>");
+                typedef std::basic_regex<char, std::regex_traits<char>> regex;
+                const static regex tag(R"(<(\w+)( [^>]+)?>|</(\w+)>)");
+                const int begin_tag_capture = 1;
+                const int params_capture = 2;
+                const int end_tag_capture = 3;
 
                 // Find all tags on this nesting level.
-                while (boost::regex_search(current, end, match, tag, flags)) {
+                while (std::regex_search(current, end, match, tag, std::regex_constants::match_default)) {
                     //Found a new tag. Recurse if begin tag, return if end tag.
-                    const boost::ssub_match& begin_match = match["begin_tag"];
-                    const boost::ssub_match& end_match = match["end_tag"];
+                    const std::ssub_match& begin_match = match[begin_tag_capture];
+                    const std::ssub_match& end_match = match[end_tag_capture];
 
                     if (begin_match.matched) {
                         // A new tag begins. Finish current plaintext tag if it is non-empty.
@@ -124,7 +124,7 @@ namespace GG {
                         }
 
                         // Recurse to the next nesting level.
-                        current = FinishTag(begin_match, match["params"],
+                        current = FinishTag(begin_match, match[params_capture],
                                             current + match.position() + match.length(), end, tags);
                     } else if (end_match.matched) {
                         // An end tag encountered. Stop parsing here.
@@ -132,7 +132,7 @@ namespace GG {
                     } else {
                         // This really shouldn't happen.
                         std::stringstream error;
-                        error << "Error parsing rich text tags: match not begin or end tag:" << match;
+                        error << "Error parsing rich text tags: matches neither begin or end tag:" << match.format("[$&]");
                         throw std::runtime_error(error.str());
                     }
                 }
--- a/GG/src/StaticGraphic.cpp
+++ b/GG/src/StaticGraphic.cpp
@@ -71,15 +71,22 @@ namespace {
 ////////////////////////////////////////////////
 StaticGraphic::StaticGraphic(const std::shared_ptr<Texture>& texture, Flags<GraphicStyle> style/* = GRAPHIC_NONE*/,
                              Flags<WndFlag> flags/* = 0*/) :
-    Control(X0, Y0, X1, Y1, flags),
-    m_style(style)
-{ Init(SubTexture(texture, X0, Y0, texture->DefaultWidth(), texture->DefaultHeight())); }
+    StaticGraphic(
+        SubTexture(texture, X0, Y0, texture->DefaultWidth(), texture->DefaultHeight()),
+        style,
+        flags
+    )
+{}
 
 StaticGraphic::StaticGraphic(const SubTexture& subtexture, Flags<GraphicStyle> style/* = GRAPHIC_NONE*/,
                              Flags<WndFlag> flags/* = 0*/) :
     Control(X0, Y0, X1, Y1, flags),
     m_style(style)
-{ Init(subtexture); }
+{
+    ValidateStyle();  // correct any disagreements in the style flags
+    SetColor(CLR_WHITE);
+    m_graphic = subtexture;
+}
 
 Flags<GraphicStyle> StaticGraphic::Style() const
 { return m_style; }
@@ -168,13 +175,6 @@ void StaticGraphic::SetTexture(const std::shared_ptr<Texture>& texture)
 
 void StaticGraphic::SetTexture(const SubTexture& subtexture)
 { m_graphic = subtexture; }
-
-void StaticGraphic::Init(const SubTexture& subtexture)
-{
-    ValidateStyle();  // correct any disagreements in the style flags
-    SetColor(CLR_WHITE);
-    m_graphic = subtexture;
-}
 
 void StaticGraphic::ValidateStyle()
 {
--- a/GG/test/acceptance/TestTextControl.cpp
+++ b/GG/test/acceptance/TestTextControl.cpp
@@ -11,14 +11,14 @@ int main(int argc, char* argv[])
 {
     Application app(argc, argv);
 
-    const std::string message = "Are <i>we</i> Готово yet?"; // That Russian word means "Done", ha.
-    const std::set<GG::UnicodeCharset> charsets_ = GG::UnicodeCharsetsToRender(message);
+    const auto message = "Are <i>we</i> Готово yet?"; // That Russian word means "Done", ha.
+    const auto charsets_ = GG::UnicodeCharsetsToRender(message);
     const std::vector<GG::UnicodeCharset> charsets(charsets_.begin(), charsets_.end());
 
-    const std::shared_ptr<GG::Font> font =
+    const auto font =
     GG::GUI::GetGUI()->GetStyleFactory()->DefaultFont(12, &charsets[0], &charsets[0] + charsets.size());
 
-    GG::Wnd* quit_dlg =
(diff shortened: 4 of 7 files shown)


adrianbroher:
@Vezzra Can you check if the Xcode project builds for you now with FreeOrionSDK v0.4.6-1?

geoffthemedio:
Builds and runs OK in MSVC 2013.

Vezzra:
Using the new SDK this PR builds ok, but the resulting executable crashes. I found two issues, the library id names of both the SDL2 and the Python libraries are wrong.

The first one crashes FO immediately on launch of the app as the dynamic linker can't load the SDL2 lib. That one however has been easy to fix, the cause being a simply typo in the `0007-Enable-MACOSX_RPATH-to-prefix-the-install_name-for-M.patch` patch file. I've provided a PR with a fix (freeorion/freeorion-sdk#18).

Rebuilding the SDK with that fix and rebuilding FO with the rebuilt SDK lets me launch FO, but can't start a game because the server crashes. Here things got a bit more tricky. The server didn't crash because it couldn't load the Python library at all. It couldn't load the Python library _bundled with the app_ (because of the wrong library id name), so the dynamic linker fell back on the Python framework installed in my system. Apparently there is something different between this one and the bundled one (from the new SDK) that causes one of the imports in one of the universe generation scripts to fail, which in turn causes the server to terminate.

Here is the server log:

```
2016-11-11 12:35:13.922623 [debug] Server : Logger.cpp:67 : Logger initialized
2016-11-11 12:35:13.923008 [info] Server : Logger.cpp:68 : v0.4.6+ add_cpp11 [build 2016-11-10.b2b49e1] Xcode 0640
2016-11-11 12:35:13.923065 [info] Server : DependencyVersions.cpp:82 : Dependency versions from headers:
2016-11-11 12:35:13.923124 [info] Server : DependencyVersions.cpp:87 : Boost: 1_59
2016-11-11 12:35:13.923522 [info] Server : DependencyVersions.cpp:87 : Python: 2.7.6
2016-11-11 12:35:13.923562 [info] Server : DependencyVersions.cpp:87 : zlib: 1.2.8
2016-11-11 12:35:13.923921 [debug] Server : ServerFSM.cpp:256 : (ServerFSM) Idle
2016-11-11 12:35:13.923993 [debug] Server : CommonFramework.cpp:38 : Initializing FreeOrion Python interface
2016-11-11 12:35:13.928557 [debug] Server : CommonFramework.cpp:49 : Python home set to /Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current
2016-11-11 12:35:13.928647 [debug] Server : CommonFramework.cpp:52 : Python program name set to /Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/Python
2016-11-11 12:35:14.063042 [debug] Server : CommonFramework.cpp:56 : Python initialized
2016-11-11 12:35:14.063093 [debug] Server : CommonFramework.cpp:57 : Python version: 2.7.10 (v2.7.10:15c95b7d81dc, May 23 2015, 09:33:12) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]
2016-11-11 12:35:14.063115 [debug] Server : CommonFramework.cpp:58 : Python prefix: /Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current
2016-11-11 12:35:14.063133 [debug] Server : CommonFramework.cpp:59 : Python module search path: /Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python27.zip:/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/:/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/plat-darwin:/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/plat-mac:/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/plat-mac/lib-scriptpackages:/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/lib-tk:/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/lib-old:/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/lib-dynload
2016-11-11 12:35:14.063159 [debug] Server : CommonFramework.cpp:66 : Initializing C++ interfaces for Python
2016-11-11 12:35:14.063321 [debug] Server : ServerFramework.cpp:59 : Initializing server Python modules
2016-11-11 12:35:14.294407 [debug] Server : LoggingWrapper.cpp:23 : Python stdout and stderr redirected
2016-11-11 12:35:14.322684 [error] Server : : : Traceback (most recent call last):
2016-11-11 12:35:14.322866 [error] Server : : :   File "/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Resources/default/python/universe_generation/universe_generator.py", line 8, in <module>
2016-11-11 12:35:14.322991 [error] Server : : :     from galaxy import calc_star_system_positions
2016-11-11 12:35:14.323066 [error] Server : : :   File "/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Resources/default/python/universe_generation/galaxy.py", line 4, in <module>
2016-11-11 12:35:14.323149 [error] Server : : :     from collections import defaultdict
2016-11-11 12:35:14.323212 [error] Server : : :   File "/Users/user/SoftwareProjekte/FO/test/Xcode/build/Test/FreeOrion.app/Contents/Frameworks/Python.framework/Versions/Current/lib/python2.7/collections.py", line 20, in <module>
2016-11-11 12:35:14.323348 [error] Server : : :     from _collections import deque, defaultdict
2016-11-11 12:35:14.323428 [error] Server : : : ImportError: No module named _collections
2016-11-11 12:35:14.323474 [error] Server : ServerApp.cpp:154 : Server's python interpreter failed to initialize.
2016-11-11 12:35:14.323524 [debug] Server : ServerApp.cpp:177 : Initiating Exit (code 1 - error termination)
```

As you can see, Python 2.7.10 gets loaded (which is the version of the Python framework installed on my system), instead of 2.7.12, which the new SDK builds (and is bundled with the app). Why the import of the `collections` module of the standard library fails beats me, as that should work with 2.7.10.

When I checked the library id name of the Python lib I get when I build the SDK on my system, I got:

```
/Users/user/SoftwareProjekte/FO/sdk/build/INSTALL/dep/Frameworks/Python.framework/Versions/2.7/lib/Python.framework/Versions/2.7/Python
```

Which means, trying to set an rpath for the Python lib doesn't work at all (the SDK built by Travis has this problem too, the path of course looks a bit different, but is an absolute one like the above). Of course I looked at the corresponding patch file which applies the required changes to Pythons CMakeLists.txt, but I can't find anything amiss so far. I also checked the patched CMakeLists.txt, and apparently the patch gets applied fine, everything looked as it should (AFAICT).

Will now try to fix the Python of the SDK library manually, rebuild FO with the fixed SDK and report back if that gives me a working build.

adrianbroher:
> It couldn't load the Python library bundled with the app (because of the wrong library id name)

Do you mean this line?

> 2016-11-11 12:35:14.063093 [debug] Server : CommonFramework.cpp:57 : Python version: 2.7.10 (v2.7.10:15c95b7d81dc, May 23 2015, 09:33:12) 
> [GCC 4.2.1 (Apple Inc. build 5666) (dot 3)]

There is another strange point:

> 2016-11-11 12:35:13.923522 [info] Server : DependencyVersions.cpp:87 : Python: 2.7.6

The SDK contains 2.7.12 and this value is taken from a compile time constant.

Vezzra:
> Do you mean this line? [...]

Yep, exactly.

> There is another strange point [...]

Noticed that too, but I suspect this being an issue in `DependencyVersions`, as fixing the library id name of the Python lib and rebuilding FO with the fixed lib gave me a working build. The server log also shows that the bundled Python framework is used now:

```
2016-11-11 13:58:15.090235 [debug] Server : CommonFramework.cpp:57 : Python version: 2.7.12 (default, Nov 11 2016, 12:24:31) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]
```

Getting the cmake scripts to produce a Python lib with the right library id name should be the only issue left to get the SDK working.

There are two more minor issues which don't break the SDK, but probably should be addressed (the second one actually only a cosmetic thing):
1. The `bootstrap.command` shell script doesn't move the `.travis.yml` file from the `temp` directory, which also leads to `temp` not being removed (because not empty). freeorion/freeorion-sdk#19 fixes that.
2. The SDK DMG file produced by the cmake scripts has the bootstrapper and the dep archive directly in the root directory. If it's not too complicated it would be better to have a "FreeOrionSDK" folder which contains these two files. Then you can just open the DMG and drag the "FreeOrionSDK" folder wherever you want your FO working copy to reside. With the current layout, you have to create an empty folder and copy the bootstrapper and dep archive there. Not a big deal, but a bit more comfortable 😉

dbenage-cx:
I have no familiarity with them, is a check for compatibility of [lambda functions](http://en.cppreference.com/w/cpp/language/lambda) warranted?

adrianbroher:
Latest master of freeorion/freeorion-sdk should build python with the correct install name.

 Python cmake upstream sets the `INSTALL_NAME_DIR` target property, which in turn overrides the `MACOSX_RPATH` target property when installing.  Fixing this sets install name to what we want.

@Vezzra can you retry building with this branch and latest master of freeorion/freeorion-sdk?

adrianbroher:
@dbenage-cx 

> I have no familiarity with them, is a check for compatibility of lambda functions warranted?

Sure.  I wanted to add a test for lambdas, but I can't find the code where I wanted to apply this anymore.  It was a some std::&lt;some container&gt;::find or or a std::find_if where we used a functor to implement the search filter, where the functor was way more complicated than the actual condition.  If you know what code I'm talking about feel free to add a sampler in the style @LGM-Doyle did for the other commits or tell me the place so I can do it myself.

Also does anybody know a class where we have multiple different constructors with mostly copied and pasted initialization so we can use `delegated constructors`?

http://en.cppreference.com/w/cpp/language/initializer_list#Delegating_constructor

Same goes for constructors, which only pass parameters to a base class constructor, but don't do any initialization.  This would be a use case for `inherited constructors`.

http://en.cppreference.com/w/cpp/language/using_declaration#Inheriting_constructors

LGM-Doyle:
There are places the code uses private `::Init()` member functions like a delegated constructor.  Multiple constructors call a single private `Init()` function.  I did not check all of the `Init()` functions, but `GG::StaticGraphic` looks like a reasonable candidate to make one of its two constructor delegate to the other and remove the `Init()` function.

I can't think of either a functor candidate or an inheriting constructor candidate.

adrianbroher:
Added some more samplers.  Still searching for an example use case of `inherited constructors`.

Vezzra:
> @Vezzra can you retry building with this branch and latest master of freeorion/freeorion-sdk?

Success! Library id names are correct now. 👍 

Generated the SDK with latest master of freeorion-sdk (freeorion/freeorion-sdk@b52894128fde6020cbdd6b0c2f951b88005f4ea0) and then built this branch (commit b2b49e18e89ddc15700c232ac745abc4d208bb45). Ran a quick test game, everything looks good.

dbenage-cx:
> 'inherited constructors'

Closest candidate I've found is [`CUITabBar`](https://github.com/freeorion/freeorion/blob/master/UI/CUIControls.cpp#L581)

Vezzra:
Built and ran the latest version of this PR/branch, seems to work fine so far on OSX (using the new C++11 enabled SDK of course). Except for the crashes I also get in master currently.

geoffthemedio:
Is this pull request still useful?

adrianbroher:
I use the commits as a todo list so far.

LGM-Doyle:
@adrianbroher, thank you for taking over and doing the conversion of the SDK kit to C++11.  Also thank you for undertaking the huge task of upgrading the entire code base to C++11.  I appreciate the work.  Thanks.

Vezzra:
Setting milestone to "undecided" as this issue is only kept open to use the commits as todo list.

Vezzra:
@adrianbroher, do you still work on something related with the switch to C++11 and need this PR open, or can we close it...?

adrianbroher:
Yeah, this PR and the related branch served its purpose.

PR #1099, opened 2016-11-12
New graphics for the organic hull

Captain-HM:


adrianbroher:
I like the old design better.  Also there is no source asset PR on freeorion/freeorion-assets.

geoffthemedio:
The new design doesn't look "organic"... It just looks vaguely stingray shaped, but made of metal.

Captain-HM:
Well, I tried to give it a metal-scale look and make it similar in style to the other hulls of the organic hull branch. I now put the blender file into the assets  repository in case someone else knows how to make it more "organic" and wants to give it a try...

Ouaz:
IMO, only the color is not good. The texture itself is nice (cell walls/cytoplasm), as the ship shape (maybe the tail could be more rounded -not a tapered tip- and split in 3 parts, like the "wings" of the old design).

Maybe you could use a beige color for the cell interiors, and a dark orange/red for the cell walls (which would look like a vascular system).

The organic_hull_small.png would need to be also a little bit sharpened:

organic_hull_small

As nice looks the old design, it needs to be replaced because it's a 2D art, and now, it doesn't fit well with the other ship graphics (3D art).

---

Note: it's a thing I have in my suggestion list:  all the ship hull icons (not the side views) in the Design Window are a little bit blurry, because they are displayed in a reduced size, and it's not possible to resize them to their native size (256x256) (unlike the side views)

IMO, it's a pity, because the great quality of these graphics ar somewhat lost.

robotic_hull_comparison

Vezzra:
> I now put the blender file into the assets repository

Two things: First, once we've agreed on a version, you need to update the blender file that got merged into the assets repo. Second, the filename of the files for the assets repo don't match the filenames of the graphics merged here.

Concerning the updated graphics: I think they look sufficiently "organic" now 👍

Captain-HM:
> Two things: First, once we've agreed on a version, you need to update the blender file that got merged into the assets repo.  Second, the filename of the files for the assets repo don't match the filenames of the graphics merged here. 

Since there was no further feedback on this topic, I now created a pull request updating the model in the assets repository to the latest (red scales w. tentacles) version and changing its name.

MatGB:
While I liked Sir Kevin's old art it didn't fit at all with the more recent models for the rest of the line and he didn't contribute much. So replacing it with this is a good plan in my opinion.

I am hoping to introduce a few more hulls and/or monsters at some point, having spare art to use is a Good Thing and the old art might fit very well for the 'monstrous' hull I might want to introduce to balance/fill out the existing Organic line.

I like this final version, should've said earlier, I think I'll miss the art being green but it won't take long to get used to, have a concern without testing it for visibility given the colour scheme and we mostly having dark backgrounds though, has anyone tested that?

Vezzra:
Just did a brief test, and yes, it is a bit dark. But it's actually not that much works than other dark hulls (e.g. the asteroid hulls), so IMO it should be fine. I say lets try it, we can always revert, or @Captain-HM can tweak the colors/texturing if it turns out to be too dark.

@Captain-HM, can you clean up the branch (squash the two commits)?

Vezzra:
@Captain-HM, bump?

Captain-HM:
sorry, was busy... 
clean-up done now.

Vezzra:
Well, been busy too... 😉

PR #1100, opened 2016-11-14
Add user doc for George communal vision

dbenage-cx:
George were granted communal vision in commit 7cbe2822
Currently no pedia entry mentions this feature, leading to the appearance of faulty behavior e.g. [forum post](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10315)

PR adds some descriptive entries to the George and the Detection Range articles to help mitigate confusion such features may cause.


default/stringtables/en.txt
```diff
@@ -12374,6 +12377,9 @@ LIGHT_SENSITIVE_DESC
 TELEPATHIC_DETECTION_DESC
 +	Telepathic Detection: can sense nearby populated planets.
 
+COMMUNAL_VISION_DESC
++	Communal Vision: shares visibility within the same species.
```
geoffthemedio: debatable whether it is a + ...?

PR #587, opened 2016-04-04
Fighters branch

MatGB:
As discussed, set up a branch on the main repo so that the AI team and the testing/balancing players can compare work and keep each other up to date.

It would be helpful if any fighter-specific changes/bugfixes, etc were added to this branch and not master until we've got things up to speed as there's a danger things will be overlooked.

@Morlic-fo , this is up to date with your branch as of 0230 BST 04/04/2016, it's not up to date with master there's a clash in AI/Shipdesign.py that I assume is your bugfix?

I've done an initial pass on balancing some of the parts in terms of cost/benefit, I've also toned down the species traits, they should be good, but not overpoweringly so, and I think we might need, moving forward, to split up good piloting (fighters) with good piloting (gunnery) for balance purposes but we're good for now.

Todo, balancing: create and cost unlock techs for all fighter based parts, they shouldn't be pre unlocked. Create refinement techs so fighters can stay good in the mid/late game against much beefier hulls and armour.

We also, if anyone reading this is good at art, need something for the fighter launch bays, the placeholder of the scattered asteroid art isn't working for me.

PR #1101, opened 2016-11-14
Fix combat log to not report targeted fighters as Unknown objects.

LGM-Doyle:
Fighters were being reported as unknown because they have a temporary
negative ID.  They are the only objects with temporary IDs.

This commit adds the empire associated with targeted objects to the data
stored in a weapon fire event so that temporary ids are reported as
Fighter and not Unknown.

This is part of ongoing fixes/improvments related to the Fighters branch, #1073 and #587.


geoffthemedio:
Seems fine.

PR #1102, opened 2016-11-14
Add pierced description for combat log

LGM-Doyle:
This PR changes the combat log description for fighter attacks so that the shield strength is reported as `pierced`.  This makes the fighter shield piercing ability more visible/discoverable.

### DIFF
--- a/combat/CombatEvents.cpp
+++ b/combat/CombatEvents.cpp
@@ -646,11 +646,18 @@ std::string WeaponFireEvent::CombatLogDescription(int viewing_empire_id) const {
 std::string WeaponFireEvent::CombatLogDetails(int viewing_empire_id) const {
     const std::string& template_str = UserString("ENC_COMBAT_ATTACK_DETAILS");
 
-    return str(FlexibleFormat(template_str)
-               % ShipPartLink(weapon_name)
-               % power
-               % shield
-               % damage);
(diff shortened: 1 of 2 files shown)


geoffthemedio:
I'm a bit concerned this could be annoying clutter after the initial discovery benefit wears off...

LGM-Doyle:
It replaces the shield strength number in the expanded portion of the display.  It will only show when the player has expanded a Fighter attack report to see how the damage breaks down.

I tried to attach an example image, but github balked.  Here is some ascii art of an expanded fighter combat log report, before this PR:

``````
````>> Fighter attacks AD Griffin Mk3 and does 5 damages
       Weapon: Fighter Weapon  Strength: 5 Shields: 3 Damage: 5````
``````

and after this PR:

``````
````>> Fighter attacks AD Griffin Mk3 and does 5 damages
       Weapon: Fighter Weapon  Strength: 5 Shields: pierced Damage: 5````
``````

Otherwise, it reports the shield strength with no indication that the shields are bypassed.

MatGB:
I think this is a good idea myself, having a shield strength shown when it's not going to do anything is going to confuse and this isn't really clutter given it's a collapsed line rarely looked at.

geoffthemedio:
ok then

PR #1103, opened 2016-11-14
Do not show tree in TechTreeWnd Constructor when game is not started.

LGM-Doyle:
This PR speeds up game startup.

When the ClientUI is initialized (before game creation) the TechTreeWnd
in created and immediately hidden.  Because the game has not been
created yet this generates several errors and takes a long time.  

The errors are the same ones that were recently removed by [ Tweaked research cost and time functions to prevent unnecessary debug…](https://github.com/freeorion/freeorion/commit/da2db7eb1a19a9b926ff35fb7bd6afaa9ef69e3f), which reminded me I had this sitting around.  I was afraid if I didn't see the errors every time I checked the logs, I'd forget about this commit.

The speed up still applies.  

This commit defaults the TechTreeWnd constructor to not show the tree.

### DIFF
--- a/UI/ResearchWnd.cpp
+++ b/UI/ResearchWnd.cpp
@@ -343,7 +343,7 @@ private:
 //////////////////////////////////////////////////
 // ResearchWnd                                  //
 //////////////////////////////////////////////////
-ResearchWnd::ResearchWnd(GG::X w, GG::Y h) :
+ResearchWnd::ResearchWnd(GG::X w, GG::Y h, bool initially_hidden /*= true*/) :
     GG::Wnd(GG::X0, GG::Y0, w, h, GG::INTERACTIVE | GG::ONTOP),
     m_research_info_panel(0),
     m_queue_wnd(0),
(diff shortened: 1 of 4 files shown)


geoffthemedio:
seems to work fine...

MatGB:
Don't know if this is the cause but it looks likely, I'm having some techs not display in the tree window regardless of filter settings, specifically Adaptive Automation and Orbital Generation (that I've spotted, there might be more).  If I queue a later dependent tech they go into my queue but I can't find them, in addition some dependent techs (eg Exobots) display as if they've got no dependencies.

LGM-Doyle:
Are you still seeing this problem?

Could you post a log and perhaps a picture when you are seeing the problem?

MatGB:
Still seeing it, it seems to change which techs are/aren't displayed between loads, I'm back on the same game as above and adaptive is now displaying, but singularity generation isn't: on the attached I haven't researched it yet, neither it nor N-Dimensional Engines are displaying but both should be, the same is true if I set it to show researched techs or all techs, but it makes things unreadable.
screenshot_20161118_151902
[freeorion-techtree.zip](https://github.com/freeorion/freeorion/files/600252/freeorion-techtree.zip)

I'm playing on a compiled 93847d8 with a rebased-to-master version of the Fighters branch but that shouldn't be making any difference at all.

Issue #1106, opened 2016-11-14
Can't build/compile


MatGB:
### Environment

* **FreeOrion Version**: 7f979fb
* **Operating System**: Kubunto 16.04
* **Fetched as**
  * Compiled from source

### Description
Have done a hard reset to remote HEAD to ensure it's not a problem with my Git repo, get following message:
`
CMakeFiles/freeorioncommon.dir/build.make:86: recipe for target 'CMakeFiles/freeorioncommon.dir/combat/CombatEvents.cpp.o' failed
make[2]: *** [CMakeFiles/freeorioncommon.dir/combat/CombatEvents.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:99: recipe for target 'CMakeFiles/freeorioncommon.dir/all' failed
make[1]: *** [CMakeFiles/freeorioncommon.dir/all] Error 2
Makefile:149: recipe for target 'all' failed
make: *** [all] Error 2
`
Full build log attached
### Expected Result
Game should compile

[make-log.txt](https://github.com/freeorion/freeorion/files/589555/make-log.txt)

LGM-Doyle:
Please try PR #1108 to see if it fixes the problem.

Issue #1107, opened 2016-11-14
Compile error: no matching function for call to 'make_shared'


Vezzra:
### Environment

* **FreeOrion Version**: 4dfc0553c0d065ad8a3a5857ae4cdac7ee6cd1b0
* **Operating System**: OSX
* **Fetched as**
  * Compiled from source

### Description

When trying to build FO on OSX, I get the following compile errors:
```
/Users/user/SoftwareProjekte/FO/src/combat/CombatEvents.cpp:942:9: error: no matching function for call to 'make_shared'
        boost::make_shared<WeaponFireEvent>(
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
and
```
/Users/user/SoftwareProjekte/FO/src/combat/CombatSystem.cpp:550:39: error: no matching function for call to 'make_shared'
        CombatEventPtr attack_event = boost::make_shared<WeaponFireEvent>(
                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I can build FO on Windows just fine. Didn't try on Linux.

geoffthemedio:
Try adding

```
#include <boost/smart_ptr/make_shared_object.hpp>
```

to the file? Or some variation

LGM-Doyle:
Please try PR #1108 to see if it fixes the problem.

PR #1108, opened 2016-11-15
Fix bugs #1106 and 1107, WeaponFireEvent doesn't work with boost::shared_ptr.

LGM-Doyle:
The WeaponFireEvent constructor has 10 parameters.
    
boost::make_shared<T> uses templates with variadic parameters to create
the shared pointer with T constructor.
    
Some compilers only emulate variadic templates and consequently don't
support boost::make_shared<T> with constructors with more than 9
parameters.
    
This commit uses boost::tuple to reduce the number of parameters in the
WeaponFireEvent constructor below 10.

This hopefully fixes issues #1106 and #1107.

I apologize for breaking the build.

### DIFF
--- a/combat/CombatSystem.cpp
+++ b/combat/CombatSystem.cpp
@@ -547,8 +547,10 @@ namespace {
                 DebugLogger() << "COMBAT: Fighter of empire " << attacker->Owner() << " (" << attacker->ID() << ") does " << damage << " damage to Ship " << target->Name() << " (" << target->ID() << ")";
         }
 
+        float pierced_shield_value(-1.0);
         CombatEventPtr attack_event = boost::make_shared<WeaponFireEvent>(
-            bout, round, attacker->ID(), target->ID(), weapon.part_type_name, power, -1.0f, damage,
+            bout, round, attacker->ID(), target->ID(), weapon.part_type_name,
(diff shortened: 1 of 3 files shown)


LGM-Doyle:
I rushed out the earlier fix.

This new fix is better in that it will work if people use the WeaponFireEvent in the customary way even with compilers that emulate variadic templates, with some cap on the number of parameters.

Vezzra:
Fix works for #1107, can build FO with this PR. 👍

MatGB:
Also able to build, it's backend code so not merging but ought to be ASAP

Vezzra:
Yep, no new test builds possible before this is merged...

combat/CombatSystem.cpp
```diff
@@ -547,8 +547,10 @@ namespace {
                 DebugLogger() << "COMBAT: Fighter of empire " << attacker->Owner() << " (" << attacker->ID() << ") does " << damage << " damage to Ship " << target->Name() << " (" << target->ID() << ")";
         }
 
+        float pierced_shield_value(-1.0);
```
geoffthemedio, line 550: -1.0f preferred

geoffthemedio:
looks mostly fine, reportedly builds...

PR #1110, opened 2016-11-17
[Fighters] Added refinements for fighters to weapon techs for testing

MatGB:
~Placeholder, intention is create separate techs
~Not for merging to Fighters or Master

Test commit with ideas for refinements: actual plan is to have them as separate techs but I wanted/needed to test them with the AI making use of them without having to mess in techlists.py (which I mess up too often)

### DIFF
--- a/default/scripting/techs/ships/Weapons.focs.txt
+++ b/default/scripting/techs/ships/Weapons.focs.txt
@@ -102,10 +102,44 @@ Tech
             scope = And [
                 Ship
                 OwnedBy empire = Source.Owner
-                DesignHasPart  name = "SR_WEAPON_2_1"
+                Or [
+                    DesignHasPart  name = "SR_WEAPON_2_1"
+                    DesignHasPart  name = "FT_HANGAR_1"
+                    DesignHasPart  name = "FT_HANGAR_2"
+                    DesignHasPart  name = "FT_HANGAR_3"
+                ]
             ]
             accountinglabel = "SR_WEAPON_2_2"
-            effects = SetMaxCapacity partname = "SR_WEAPON_2_1" value = Value + 2
+            effects = [
+                SetMaxCapacity partname = "SR_WEAPON_2_1" value = Value + 2
+                SetMaxSecondaryStat partname = "FT_HANGAR_1" value = Value + 1
+                SetMaxSecondaryStat partname = "FT_HANGAR_2" value = Value + 2
+                SetMaxSecondaryStat partname = "FT_HANGAR_3" value = Value + 3
+                ]
+ /*       EffectsGroup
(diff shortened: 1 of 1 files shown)


MatGB:
It took me awhile to get this working at all and I'm not at all certain on the numbers, they're roughly based on Geoff's numbers for the species traits but they're also something I'm not certain on.

Plan is to have a tech called "Mass Driver Fighters" that unlocks the basic fighters in game, then have a tech dependent on that and Laser1 called "Laser Fighters" for the first refinement, etc. I suspect these might need to be fairly expensive for balance but there definitely needs to be some progression else late game fights take far too long and, without a species trait, they remain at the base damage level for a weapon and the equivalent shield.

Blue42hand:
Oh gotcha, that makes sense.

Morlic-fo:
> This implementation does, but that's purely for testing, the AI already researches those techs so I wanted to see how the game plays with and against the refinements, it will be a separate tech when/if we put it in but that requires an AI update that's beyond my skills, I need it in use for testing.

Note that while the AI does actually use the correct ship meter / part meter stats once the ships are built, tech upgrades are unknown to the AI when rating ship designs unless specified in AIDependencies.py.

I added the relevant dicts in 50a5c8f to Fighters branch. Can you add the missing entries yourself or is the documentation insufficient? I would like to get to a point where you can add the necessary AIDependencies entries yourself to improve workflow.

/Edit: Also needs   bb31b16  to work in case you were going to cherry-pick instead of rebase.

MatGB:
@Morlic-fo thanks, been ill and only now had the strength to read the code, I **think** I know what I need to do and that is a lot clearer than it was, and I can fudge inserting the techs into techlists.py in a way that you can move them if you think I got it wrong.

I think I'm almost able to read basic Python. I think. 

Tomorrow, assuming nothing goes wrong, I'll hopefully be able to write a basic tech tree script and make the AI able to use it, then I test again...

Vezzra:
@MatGB, if I understand correctly, this PR has been superseded by another one, so I removed the "help wanted" etc. labels and added the "follow up". However, the superseding PR should be referenced here, but I don't know which one it is. Can you provide the reference...?

MatGB:
Oh, yeah, sorry, forgot this one, it was #1140

Vezzra:
Um, the "status:follow up" label needs to stay... 😉

MatGB:
Oh, OK, I've misunderstood that, thought it was for stuff we needed to follow up not things that've been superseded. And completely didn't read your previous comment saying this properly either. Sometimes I don't know where my brain is.

Vezzra:
No problem. I know the feeling... 😉

Issue #1093, opened 2016-11-09
Sidepanel Flicker During turn Updates


geoffthemedio:
In the latest master, during turn updates, the contents of the sidepanel flicker visible and not a few times. It's rather distracting. Attached is a composition of screenshots attempting to illustrate the two things that appear and disappear...

The layout also seems to be reset a few times, with the meter bars flickering a bit longer / shorter.

![sidepanel flicker](https://cloud.githubusercontent.com/assets/11386908/20137327/0dacfae2-a67a-11e6-997a-6efa184ce632.png)

LGM-Doyle:
I have replicated this problem and I will work on a solution.

MatGB:
A solution: force close sidepanels and fleet windows at the end of a turn. Having either or both open slows the game down a fair bit for a lot of players, I'd personally prefer it if they auto closed.

LGM-Doyle:
The intent of my post was just to forestall Geoff pinging me to look at this issue.

I play with side panel and the fleet window open, but the objects window closed.  I think I would be unhappy if  they closed each turn, since sometimes I watch my colonies grow.

The slow down can be mitigated.  

The current glitch is due to work to speed up the side panel that has created a mismatch in the redraw timing of the planet container and the side panel proper.  I can fix the glitch.

Vezzra:
According to the discussion @geoffthemedio and @LGM-Doyle had in #1154, this issue has been (almost) completely resolved,  except for this small thing @geoffthemedio mentioned that apparently still lingers:

> There's no visible jitter in the position of the sidepanel controls, but the stealth stat icon does seem to flicker (go black) briefly - perhaps a frame - when pressing turn. (In the case where all the sub-panels are collapsed)

Is that minor enough so we can close this nonetheless, or do you guys want to keep it open until this last bit has been addressed?

LGM-Doyle:
Keep it open or start a new issue with only the remaining issue, so that it is not lost.  The latter would be preferable so that someone can come along and understand the issue without reading the whole thread.

Vezzra:
Ok, done: #1177

PR #1113, opened 2016-11-18
Partial fix of bug #1093 sidepanel flicker.

LGM-Doyle:
This PR mostly fixes issue #1093 of sidepanel flicker during updates.

The flicker was caused by the side panel, system name, system resource indicator, planet panel container and accordion panels all having `PreRender()` function that update during the prerender phase.  However, the sidepanel re-creates or triggers data updates of the layouts of its child panels during its `PreRender()`.  These children then update their layout during the **next** prerender phase.  Since the sidepanel layout is slow, 10s of ms on my machine this delay is very visible.

This PR fixes the problem by calling `GG::GUI::PreRenderWindow()` on the system name, system resource indicator and the planet panel container after those children are modified so that they immediately execute their `PreRender()` functions.

I did not fix the glitch with the accordion panels, because it looks like a larger fix that will spill over into the combat log, so I will work on it separately.

### DIFF
--- a/GG/src/DropDownList.cpp
+++ b/GG/src/DropDownList.cpp
@@ -385,6 +385,7 @@ DropDownList::iterator DropDownList::Insert(Row* row, iterator it, bool signal/*
     row->SetDragDropDataType("");
     DropDownList::iterator ret = LB()->Insert(row, it, signal);
     Resize(Size());
+    RequirePreRender();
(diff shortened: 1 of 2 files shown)


geoffthemedio:
I don't have time to look through all the code, but the result is slightly better.

Issue #969, opened 2016-09-12
Focus dropdown list box gets clipped when it doesn't fit in the window


Vezzra:
### Environment
- **FreeOrion Version**: 0.4.6 RC1
- **Operating System**: OSX
- **Fetched as**
  - Binary release
### Description

![clippedfocuslist](https://cloud.githubusercontent.com/assets/10279417/18446516/e96f0b90-7922-11e6-9660-c28fc02504c4.png)

Hopefully the screenshot shows the problem sufficiently well.

If a system has enough planets to fill up the sidepanel down to the bottom, then it's sometimes impossible to access all focus settings of the bottommost planet. The sidepanel scrollbar doesn't allow to scroll down far enough and thus get the bottom planet up far enough so that the entire focus dropdown list box fits within the window.

Also, the box does not get expanded upwards, if it does not fit within the window when expanded, there appears no scrollbar and you can't move the selection with the cursor keys.

The only way to set the focus for such a planet is either by the object window, or if you resize the sidepanel so that the bottom planet is forced to move sufficiently far away from the bottom window border that the dropdown list box fits into the window.
### Expected Result

If the focus dropdown list box extends past the bottom window border, one or more of those should happen:
- Add some extra space below the bottommost planet so you can scroll that planet up high enough for the box to fit within the window when expanded.
- Expand the dropdown list box upward.
- Provide a vertical scrollbar.
- Enable the cursor keys so that a selection can be made using them (+space or enter to confirm the selection).
### Steps to reproduce
- Start new game.
- Resize the FO window so that the focus dropdown list box of your starting colony does not fit within the window when expanded (at least one focus should be entirely in the clipped part of the box).
- Try to select the focus that is in the clipped part of the box.
- You won't succeed... 😉

Vezzra:
As this is a quite annoying bug, I'll tag it for the release, which means there _should_ be a RC2.

However, if no one gets around to fix that suffciently quickly, I won't really block the release for this, and if no actually critical issue turns up, there will also be no RC2. I'll just try to give it some priority by tagging it for the release.

geoffthemedio:
I don't understand the problem... Just expand the buildings, population, or resources panel, then scroll a bit more down.

Vezzra:
Right. I can officially claim the "dumbest assumable user" crown from @Morlic-fo now. 😳 😉

[...]

PR #1114, opened 2016-11-18
Add DropDownList keypress while list is dropped.

LGM-Doyle:
This PR partially addresses issue #969 "Focus dropdown list box gets clipped when it doesn't fit in the window", by refactoring DropDownList to allow key presses to work while the drop down list is open.  It was the easiest of the suggestions that @Vezzra made to make work.

It refactors behavior that was already present when the drop down was closed, and moves it into the drop down so that it can also run when the drop down is both closed and open.

### DIFF
--- a/GG/src/DropDownList.cpp
+++ b/GG/src/DropDownList.cpp
@@ -510,7 +510,9 @@ void DropDownList::LClick(const Pt& pt, Flags<ModKey> mod_keys)
     }
     LB()->m_first_col_shown = 0;
 
+    DropDownOpenedSignal(true);
     m_modal_picker->Run();
+    DropDownOpenedSignal(false);
 }
 
 void DropDownList::KeyPress(Key key, boost::uint32_t key_code_point, Flags<ModKey> mod_keys)
--- a/GG/src/DropDownList.cpp
+++ b/GG/src/DropDownList.cpp
@@ -31,10 +31,11 @@
 #include <GG/StyleFactory.h>
 #include <GG/WndEvent.h>
 
+#include <boost/optional/optional.hpp>
 
 using namespace GG;
 
-class ModalListPicker : public Wnd
+class ModalListPicker : public Control
 {
 public:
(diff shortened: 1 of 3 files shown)


GG/src/DropDownList.cpp
```diff
@@ -53,13 +58,40 @@ class ModalListPicker : public Wnd
 
     mutable SelChangedSignalType SelChangedSignal; ///< the selection change signal object for this DropDownList
 
+    DropDownList::iterator CurrentItem();
+
+    /** Select \p it  in m_lb_wnd is not none and return the new selected or none if no change.*/
```
dbenage-cx: This comment is hard to read, can it be clarified?

dbenage-cx (approved):
No issue with selecting using keys or mouse wheel here.
Not sure this PR is a complete fix for the issue, since the list itself is still 'clipped'..

GG/src/DropDownList.cpp
LGM-Doyle: It looks like I left out the word "if" as well as phrased is awkwardly.  I've rephrased it.

GG/src/DropDownList.cpp
```diff
@@ -53,13 +58,41 @@ class ModalListPicker : public Wnd
...
+    virtual void    MouseWheel(const Pt& pt, int move, Flags<ModKey> mod_keys);
+
 private:
     void LBSelChangedSlot(const ListBox::SelectionSet& rows);
 
     void LBLeftClickSlot(ListBox::iterator it, const GG::Pt& pt, const GG::Flags<GG::ModKey>& modkeys);
 
     ListBox*    m_lb_wnd;
     const Wnd*  m_relative_to_wnd;
+    bool m_dropped;  ///< Is the drop down list open.
```
geoffthemedio, line 95: indent to align with other members

Vezzra:
Thanks! 👍

Issue #786, opened 2016-07-24
Label Fleets Window as "near System X"


geoffthemedio:
If a fleets window is showing fleets not at a system, indicate that the fleets are "near SystemName" instead of just saying "Fleets".

Cjkjvfnby:
May be "on way to SystemName" for fleet with known destination?

geoffthemedio:
This is about the whole window, not individual fleets. Also, fleets already are labelled with their destination.

Vezzra:
Implemented by #1115

PR #1115, opened 2016-11-18
Add fleetwnd near to system title

LGM-Doyle:
This implements #786 'Label Fleets Window as "near System X"'.

Most of the work was fixing the ability to correctly track the location of fleet in the fleet window and their owners.  The commit message of the fourth commit explains the relevant logic and I have copied it below.

Add a FleetWnd title that indicates near system X exposes the following
subtle bug in FleetWnd.  When a FleetWnd is created for a travelling
fleet and the fleet arrives at a system the FleetWnd still displays the
fleet and still thinks that it is travelling.

This commit fixes the bug by changing the tracking behavior to the
following.

Are all fleets in one location?  Use that location.

Otherwise, are all selected fleets in one location?  Use that location
and remove fleets not at that location.

Otherwise, is the current location a system?  Use that location and add
appropriate fleets at that system.

Otherwise remove all fleets as all fleets have gone in separate
directions.

### DIFF
--- a/GG/GG/PtRect.h
+++ b/GG/GG/PtRect.h
@@ -30,6 +30,7 @@
 
 #include <GG/Base.h>
 #include <GG/StrongTypedef.h>
+#include <boost/functional/hash.hpp>
 
 
 namespace GG {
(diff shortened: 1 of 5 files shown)


Vezzra:
Builds and runs fine on OSX. Works as intended.

Issue #1117, opened 2016-11-19
Tech tree displaying incompletely.


LGM-Doyle:
MatGB observed this problem and attached it to PR #1103.  I have moved MatGB's description here and started a new issue as I think it is not the result of PR #1103.

MatGB's initial post was:

Don't know if this is the cause but it looks likely, I'm having some techs not display in the tree window regardless of filter settings, specifically Adaptive Automation and Orbital Generation (that I've spotted, there might be more). If I queue a later dependent tech they go into my queue but I can't find them, in addition some dependent techs (eg Exobots) display as if they've got no dependencies.

* **FreeOrion Version**: Fighter branch rebased on weekyl master
* **Operating System**:

LGM-Doyle:
You are playing with essentially the exact same build as myself, Fighters rebased to master with 2-4 of my in progress branches merged on top.  However, I'm not seeing anything like the behavior you are describing, even though I spent some time clicking on the various research window filters starting from the same place you seem to be in the tech tree.

So I have some questions, to determine what is different.

What OS are you playing on?

Did you start this game with the exact same compiled version of freeorion, or did you start with one version,  save, re-compile and then re-load the same save with a different compiled version?  

Are you playing with custom FOCS scripts?  If so could you push a branch to your own repo and I'll look at those and see if that  is the difference.

Did you start the game with one version of FOCS files and then re-load the same game with a different set of FOCS files?

dbenage-cx:
`initially_hidden` is never false (at least, I don't see a case), so the control/status' are not initialized.

For me, the individual category icons are greyed until I toggle the 'All' button twice.

LGM-Doyle:
@dbenage-cx thank you.   You are absolutely correct.  My eyes must have glossed over and missed the second click each time.  Does PR #1119 fix that problem for you?

@MatGB does it fix hopefully or change the behavior of your bug?

MatGB:
Have recompiled and loaded the game, on initial load the problem does appear to have gone but compiling takes awhile and just finished.

Also, yes, agree with dbenage, I noticed last night as I shut down that some stuff reappeared if I toggled everything but it wasn't clear enough, was going to test today then report but done now, thanks.

dbenage-cx:
Fixed in #1119

PR #1119, opened 2016-11-19
Add TechTreeWnd::Show() to fix bug with deferred tech parsing in TechTreeWnd

LGM-Doyle:
The TechTreeWnd parsing was deferred until the client could access the
technologies, but was never intentionally activated.

This commit creates a Show() function that checks if the parsing was
deferred and parses the techs the first time that show is called.

@dbenage-cx pointed this bug with #1103 out in issue #1117.

### DIFF
--- a/UI/TechTreeWnd.cpp
+++ b/UI/TechTreeWnd.cpp
@@ -1908,7 +1908,8 @@ TechTreeWnd::TechTreeWnd(GG::X w, GG::Y h, bool initially_hidden /*= true*/) :
     m_tech_tree_controls(0),
     m_enc_detail_panel(0),
     m_layout_panel(0),
-    m_tech_list(0)
+    m_tech_list(0),
+    m_initially_hide_tree(initially_hidden)
 {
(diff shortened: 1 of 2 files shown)


UI/TechTreeWnd.cpp
```diff
@@ -1908,7 +1908,8 @@ TechTreeWnd::TechTreeWnd(GG::X w, GG::Y h, bool initially_hidden /*= true*/) :
     m_tech_tree_controls(0),
     m_enc_detail_panel(0),
     m_layout_panel(0),
-    m_tech_list(0)
+    m_tech_list(0),
+    m_initially_hide_tree(initially_hidden)
```
dbenage-cx: Minor gripe, should this be something like `m_init_flag`?  The current name suggests it reflects the initial state.

dbenage-cx (commented):
From brief testing, working fine here.

UI/TechTreeWnd.cpp
LGM-Doyle: Fixed.

PR #1121, opened 2016-11-20
Fix production wnd does not track sidepanel.

LGM-Doyle:
This PR fixes the bug where the ProductionWnd's selected planet no longer persistently tracks the system selected in the SidePanel.  The bug was introduced by changes to add PreRender() to the SidePanel.

Previously, selection and storage of the selected planet id was handled by the PlanetContainerPanel inside the SidePanel that was linked to the production window.  

Adding PreRender() changed the timing of events, so that when the Production window was informed that the system had changed and polled the PlanetContainerPanel it was still showing the old system, which the Production window then used to determine which planet's BuildDesignator to display. 

This PR promotes selection of the planet id and storage of the value to the static functions and variables that tie together the two SidePanels.  This is at a higher level of the hierarchy, outside of the PreRender() functions.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -2473,7 +2473,6 @@ void SidePanel::PlanetPanelContainer::ScrollTo(int pos) {
 void SidePanel::PlanetPanelContainer::Clear() {
     m_planet_panels.clear();
     m_selected_planet_id = INVALID_OBJECT_ID;
-    PlanetSelectedSignal(m_selected_planet_id);
     DetachChild(m_vscroll);
     DeleteChildren();
     AttachChild(m_vscroll);
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -3332,6 +3332,13 @@ void SidePanel::SelectPlanetImpl(int planet_id) {
 void SidePanel::SetSystem(int system_id) {
     if (s_system_id == system_id)
         return;
+
+    TemporaryPtr<const System> system = GetSystem(system_id);
+    if (!system) {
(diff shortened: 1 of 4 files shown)


MatGB:
On initial testing this appears to have solved my issues but it's a new game so not many systems yet.

MatGB:
Spoke slightly too soon, double clicking a specific planet in a multi-planet system doesn't open the prodwnd with that planet selected, it keeps to the previously selected/used planet.

LGM-Doyle:
@MatGB I think these new commits address the issue that you observed.

MatGB:
Appears to on first test, another new game, will confirm when I've got more than 3 planets.

MatGB:
Confirmed, working as intended on all uses I've tried so far.

UI/SidePanel.cpp
```diff
@@ -3263,33 +3252,73 @@ void SidePanel::FleetStateChanged()
...
+bool SidePanel::PlanetSelectable(int planet_id) const {
+    if (!m_selection_enabled)
+        return false;
+
+    TemporaryPtr<const System> system = GetSystem(s_system_id);
+    if (!system)
+        return false;
+
+    const std::set<int>& planet_ids = system->PlanetIDs();
+    if (!planet_ids.count(planet_id))
```
geoffthemedio, line 3267: this is a bit odd at first glance. would prefer
````
planet_ids.count(planet_id) > 0
````
or
````
planet_ids.find(planet_id) != planet_ids.end()
````

LGM-Doyle: Fixed in master.

Issue #1109, opened 2016-11-15
Human client crashes when opening production screen


Vezzra:
### Environment

* **FreeOrion Version**: 93847d84df206d6a28b2ba1919fec522ec7feba6
* **Operating System**: OSX
* **Graphic card used**: AMD Radeon HD 6490M 256 MB
* **Fetched as**
  * Compiled from source

### Description

On OSX, when opening the production screen the human client throws an exception, see these lines of `freeorion.log`:
```
2016-11-15 22:55:57.632665 [debug] Client : ProductionWnd.cpp:999 : ProductionWnd::UpdateQueue()
2016-11-15 22:55:57.679054 [debug] Client : ProductionWnd.cpp:999 : ProductionWnd::UpdateQueue()
2016-11-15 22:55:57.730758 [debug] Client : ClientNetworking.cpp:317 : ClientNetworking::NetworkingThread() : Networking thread will be terminated due to disconnect exception "Operation canceled"
2016-11-15 22:55:58.731709 [debug] Client : Process.cpp:231 : Process::Impl::Kill calling kill(m_process_id, SIGKILL)
2016-11-15 22:55:58.731924 [debug] Client : Process.cpp:233 : Process::Impl::Kill calling waitpid(m_process_id, &status, 0)
2016-11-15 22:55:58.738236 [debug] Client : Process.cpp:235 : Process::Impl::Kill done
2016-11-15 22:55:58.738307 [debug] Client : HumanClientFSM.cpp:664 : (HumanClientFSM) ~PlayingTurn
2016-11-15 22:55:58.738337 [debug] Client : HumanClientFSM.cpp:334 : (HumanClientFSM) ~PlayingGame
2016-11-15 22:55:58.859604 [error] Client : chmain.cpp:304 : main() caught exception(std::exception): Layout::DoLayout() : calculated column positions do not sum to the width of the layout
```
Curiously, this doesn't happen *always*. Apparently, depending on what I do before I go to the production screen, the crash happens or not.

I haven't been able to reproduce the crash on Windows.

### Expected Result

Well, no crash... 😉 

### Steps to reproduce

The following procedure has so far reliably reproduced the crash:

* Start new game.
* Open research screen.
* Open production screen.
* Crash.

MatGB:
This might be related and I got called away before I could open issue but I'm finding planets aren't selecting properly to start building immediately when I open prod or change system within it, have to select the planet and build lists are sometimes for a previous system. 

[...]

PR #1124, opened 2016-11-22
Fix bug #1109 drop down list modal event pump race.

LGM-Doyle:
tl;dr; A race condition between the EventPump and the ModalEventPump in the DropDownList code was fixed by adding a separate signal SelChangedWhileDroppedSignal.

This fixes issue #1109.

****The problem****

+ DropDownList uses a ModalEventPump to display the drop down list in ModalListPicker, while at the same time moving and restoring the selected row each Render() phase to render it in the drop down list anchor location.  
+ SidePanel::SystemRow displays a centered system name.  
+ SidePanel deletes and re-creates all of the rows immediately when SystemRow indicates a change in the selected row.  
+ The DropDownList and the DropDownList::ModalListPicker are both running PreRender on the same SystemRow in different EventPumps.  

A race condition happens which sometimes results in the SystemRow::Init() function running twice.  This creates an inconsistent SystemRow, which eventually leads to the crash.

Triggering this bug does not require the SidePanel to be touched in anyway.  It might require the SidePanel to have been visible at least once, but  I did not check this condition.

****The evidence****

I captured the stack inside of the SystemRow::Init() function showing that it had a) run unnecessarily a second time when m_intialized was already true and b) that it added a spurious control with a min width in Layout of 32767 !!!.  SystemRow while valiantly keeping the System name centered aliases the spurious min width with certain widths of SidePanels to create either ever larger magnitude positive/negative (m_ul_prerender, m_lr_prerender) pairs.  These are the large values that Geoff noticed that look uninitialized.  This is also what I assume was causing the crash to be very machine specific.

****A solution****

1. Remove the clamping of SystemRow to the parent size that was interacting badly with the DropDownList double render of the selected row.
2. Prevent the DropDownWnd correcting the size of the drop down box while the drop down box is visible.  This should prevent both DropDownList::PreRender() and ModalListPicker::PreRender() from racing to prerender the list box.
3. Make EndRun() virtual so that ModalListPicker can hide its list as part of its teardown within the ModalEventPump.
4. Split SelChangedSignal in to SelChangedWhileDroppedSignal and SelChangedSignal so that consumers can handle them appropriately.  I considered passing a parameter `is_dropped`, but chose two signals to emphasize that the SelChangedWhileDropped version is handled in the ModalEventPump and not the normal event pump.  The user could also use a single handler and detect status with Dropped().
5. Change the original SelChangedSignal to emit a final SelChanged when the drop down closes so that old code does not miss any selection changes events.
6. Changed the SidePanel code to separate the re-generation of the SystemRow so that it doesn't happen while the modal list is open.  

****Not perfect****

It introduces (at the least) the following quirk/bug.  If the system name drop down list is open and the names of known systems changes this will not be reflected in the systems in the list until the list is closed and re-opened.  A solution to that is to refactor DropDownList to not use  ModalEventPump.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -2817,7 +2817,8 @@ SidePanel::SidePanel(const std::string& config_name) :
     m_system_resource_summary = new MultiIconValueIndicator(Width() - EDGE_PAD*2);
     AttachChild(m_system_resource_summary);
 
-    GG::Connect(m_system_name->SelChangedSignal,                         &SidePanel::SystemSelectionChanged, this);
+    GG::Connect(m_system_name->SelChangedSignal,                         &SidePanel::SystemSelectionChangedSlot,    this);
+    GG::Connect(m_system_name->SelChangedWhileDroppedSignal,             &SidePanel::SystemSelectionChangedSlot,    this);
     GG::Connect(m_button_prev->LeftClickedSignal,                        &SidePanel::PrevButtonClicked,      this);
     GG::Connect(m_button_next->LeftClickedSignal,                        &SidePanel::NextButtonClicked,      this);
     GG::Connect(m_planet_panel_container->PlanetSelectedSignal,          &SidePanel::PlanetSelected,         this);
@@ -3237,7 +3238,10 @@ void SidePanel::SizeMove(const GG::Pt& ul, const GG::Pt& lr) {
         RequirePreRender();
 }
 
-void SidePanel::SystemSelectionChanged(GG::DropDownList::iterator it) {
+void SidePanel::SystemSelectionChangedSlot(GG::DropDownList::iterator it) {
+    /** This handles cases when the list is dropped and not dropped in the
+        same way. Refresh should not update the list of systems if the list
+        is open. */
     int system_id = INVALID_OBJECT_ID;
     if (it != m_system_name->end())
         system_id = boost::polymorphic_downcast<const SystemRow*>(*it)->SystemID();
@@ -3251,7 +3255,7 @@ void SidePanel::PrevButtonClicked() {
     if (selected == m_system_name->begin())
         selected = m_system_name->end();
     m_system_name->Select(--selected);
-    SystemSelectionChanged(m_system_name->CurrentItem());
+    SystemSelectionChangedSlot(m_system_name->CurrentItem());
 }
 
 void SidePanel::NextButtonClicked() {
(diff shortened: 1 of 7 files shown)


UI/SidePanel.cpp
```diff
@@ -3056,54 +3038,67 @@ void SidePanel::RefreshInPreRender() {
...
+void SidePanel::RefreshSystemNames() {
+    TemporaryPtr<const System> system = GetSystem(s_system_id);
+    // if no system object, there is nothing to populate with.  early abort.
+    if (!system)
+        return;
+
+    // Repopulate the system with all of the names of known systems, if it is closed.
+    // If it is open do not change the system names because it runs in a seperate ModalEventPump
+    // from the main UI.
+    if (!m_system_name->Dropped()) {
```
geoffthemedio, line 3050: prefer
````
if (test)
    return
(non-indented code)
````
to
````
if (!test) {
    (bunch of indented code)
}
````

File: UI/SidePanel.cpp [hunk 2]
```diff
@@ -3056,54 +3038,67 @@ void SidePanel::RefreshInPreRender() {
...
+        std::map<std::string, int> system_map; //alphabetize Systems here
+        for (ObjectMap::const_iterator<System> sys_it = Objects().const_begin<System>();
+             sys_it != Objects().const_end<System>(); ++sys_it)
+        {
+            if (!sys_it->Name().empty() || sys_it->ID() == s_system_id) // skip rows for systems that aren't known to this client, except the selected system
+                system_map.insert(std::make_pair(sys_it->Name(), sys_it->ID()));
+        }
+        std::vector<GG::DropDownList::Row*> rows;
+        rows.reserve(system_map.size());
+        for (std::map< std::string, int>::iterator sys_it = system_map.begin(); sys_it != system_map.end(); ++sys_it)
```
geoffthemedio, line 3063: break over line?

UI/SidePanel.cpp
```diff
@@ -3056,54 +3038,67 @@ void SidePanel::RefreshInPreRender() {
...
+        {
+            int sys_id = sys_it->second;
+            rows.push_back(new SystemRow(sys_id));
+        }
+        m_system_name->Insert(rows, false);
+
+        // select in the list the currently-selected system
```
geoffthemedio, line 3070: does inserting invalidate other row iterators? if no, can you select when adding instead of as a separate step? probably not, as the preexisting code did it this way...

geoffthemedio:
reportedly helps, so merging...

PR #1126, opened 2016-11-23
Change side panel SystemName to an unsorted list with pre-sorted rows.

LGM-Doyle:
This improves performance, because the ListWnd sort is an insertion sort.

This also addressed style issues from review #1124.

I was literally 5 minutes away from pushing these changes fixing the issues you raised in #1124 and adding a slight performance improvement, when you merged #1124.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -750,7 +750,7 @@ namespace {
         int SystemID() const { return m_system_id; }
 
         virtual SortKeyType SortKey(std::size_t column) const
-        { return GetSystem(m_system_id)->Name(); }
+        { return GetSystem(m_system_id)->Name() + boost::lexical_cast<std::string>(m_system_id); }
 
     private:
         int m_system_id;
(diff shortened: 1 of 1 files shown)


UI/SidePanel.cpp
```diff
@@ -3040,42 +3040,56 @@ void SidePanel::RefreshInPreRender() {
...
+
+    // The system names are manually sorted here and not automatically in
+    // the vectorized insert because the ListBox is currently never
+    // resorted, inserted or deleted so this was faster when profiled.  If
+    // the performance of the std::stable_sort used in ListBox improves to
+    // N logN when switching to C++11 then this should simply insert the
+    // entire vector into the ListBox.  If the approach switches to
+    // maintaing the list by incrementally inserting/deleting system
+    // names, then this approach should also be dropped.
+    std::map<std::string, int> system_map;
```
geoffthemedio: given that systems can now be renamed, this needs to account for multiple systems having the same name

geoffthemedio (changes requested):

LGM-Doyle:
I wonder who did that?  Good catch.  Fixed.

PR #1132, opened 2016-11-25
Remove AccordionPanel::PreRender().

LGM-Doyle:
From profiling it saves only microseconds per call and it is not
algorithmically triggered, so should never be trigger 1000 times during
a single render interval.

This makes further improvements to the flicker of the sidepanel in issue #1093.


Vezzra:
@LGM-Doyle, I assume you meant to set the milestone to "Next Release", not "Gateway to the Void", so I corrected that 😉

LGM-Doyle:
Thanks.  The javascript always takes some time to update and I wasn't patient enough.

PR #1134, opened 2016-11-25
Add ListBox::Show() to only show rows within the visible box.

LGM-Doyle:
This PR further speeds up and cleans up the SidePanel SystemName drop down list.

It adds `ListBox::Show(bool children)`, which unlike the default `Show()` does not show all of its children.  It only shows those rows which would be visible if the list rendered.  This means that the `PreRender()` of rows that would not be in the drop down are not called.  It still calls all of the `PreRender()` functions of the invisible items in the dropped portion of the list even when it is not dropped, so it is not optimal.

This speeds up re-generation of the SidePanel even when the SystemName is not expanded.  From profiling on a 1000 system map this improves the re-load of the SystemName caused by the Super-Testers revealing all of the systems from 400ms to 1.8ms.  In a regular game this cost is paid every time the SidePanel updates the SystemName contributing to late game sluggishness.

Making this change exposed a bug where `ListBox::Row::operator[]` creates dangling null Control* pointers, which causes problems (i.e. crashing), so that operator is removed in this PR.

### DIFF
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -1872,9 +1872,9 @@ void BasesListBox::BaseRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, 
                 if (result != "" && result != design->Name()) {
                     HumanClientApp::GetApp()->Orders().IssueOrder(
                         OrderPtr(new ShipDesignOrder(client_empire_id, design_id, result)));
-                    ShipDesignPanel* design_panel = dynamic_cast<ShipDesignPanel*>(
-                        !design_row->empty() ? design_row->at(0) : 0);
-                    design_panel->Update();
+                    if (!design_row->empty())
+                        if (ShipDesignPanel* design_panel = dynamic_cast<ShipDesignPanel*>(design_row->at(0)))
+                            design_panel->Update();
                 }
                 break;
             }
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -458,7 +458,7 @@ void PartsListBox::PartsListBoxRow::ChildrenDraggedAway(const std::vector<GG::Wn
     // find control in row
     unsigned int i = -1;
     for (i = 0; i < size(); ++i) {
-        dragged_control = (*this)[i];
+        dragged_control = !empty() ? at(i) : 0;
         if (dragged_control == control)
             break;
         else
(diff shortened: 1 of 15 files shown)


UI/DesignWnd.cpp
```diff
@@ -1872,7 +1872,8 @@ void BasesListBox::BaseRightClicked(GG::ListBox::iterator it, const GG::Pt& pt,
                 if (result != "" && result != design->Name()) {
                     HumanClientApp::GetApp()->Orders().IssueOrder(
                         OrderPtr(new ShipDesignOrder(client_empire_id, design_id, result)));
-                    ShipDesignPanel* design_panel = dynamic_cast<ShipDesignPanel*>((*design_row)[0]);
+                    ShipDesignPanel* design_panel = dynamic_cast<ShipDesignPanel*>(
```
geoffthemedio: I'd have put the ? outside rather than inside the dynamic_cast call...

UI/MultiplayerLobbyWnd.cpp
```diff
@@ -332,7 +332,8 @@ namespace {
             m_color_selector->SelectColor(m_player_data.m_empire_color);
 
             // set previous player name indication
-            boost::polymorphic_downcast<GG::Label*>(operator[](4))->SetText(it->second.m_player_name);
+            if (size() >=5)
```
geoffthemedio: space after >= before 5

UI/TechTreeWnd.cpp
```diff
@@ -1680,11 +1680,11 @@ void TechTreeWnd::TechListBox::TechRow::Update() {
         return;
 
     std::string cost_str = boost::lexical_cast<std::string>(static_cast<int>(this_row_tech->ResearchCost(HumanClientApp::GetApp()->EmpireID()) + 0.5));
-    if (GG::TextControl* tc = dynamic_cast<GG::TextControl*>((*this)[2]))
+    if (GG::TextControl* tc = dynamic_cast<GG::TextControl*>(!empty() ? at(2) : 0))
```
geoffthemedio: why not check size() >= 3 instead of !empty if you're indexing at 2?

geoffthemedio (changes requested):

geoffthemedio (approved):

LGM-Doyle:
Do you want me to merge this?

geoffthemedio:
At least to me, approving changes means I approved those changes in response to the requested changes, not that I've decided to merge, which I would have done myself.

LGM-Doyle:
I was unsure if you were changing the procedure.  Thanks for clarifying.

geoffthemedio:
rebase?

LGM-Doyle:
Rebased.

PR #1144, opened 2016-11-27
Fix bug 1128 ObjectWnd sort is not persistent.

### LINKED ISSUE #1128: Objects menu sort not working nor remembered on action
### Environment

* **FreeOrion Version**: compiled from [MatGB/freeorion at lgm-doyle-test-branch](https://github.com/MatGB/freeorion/tree/lgm-doyle-test-branch) using some fixes by @LGM-Doyle then running using the a rebased Fighters branch for default
* **Operating System**: Linux Kubuntu
* **Fetched as** 
  * Compiled from source

### Description
List boxes in general and the Objects menu in particular are especially jumpy and not always sorting correctly. Specifically, if a column is created in Objects for Environment and then clicked, it should sort all my planets by their environment, it doesn't.

Regardless of this, if I select a number of planets and right click/queue a building or right click/set a focus, the list resorts itself and loses position. However, even on columns where the sort is working correctly (sort by focus works fine) the sort is lost but the list doesn't know this, so reclicking will give a reverse sort requiring another click instead.

I've also noticed the production screen listbox jumps when an item is queued, especially buildings, this is also new behaviour and is likely linked, given that queuing one item now allows for extra items to appear at that location (eg Drydocks followed by Nanorobotic facility) this is not at all ideal and can be disruptive.

### Expected Result
Clicking a column to sort should correctly sort so that eg all Poor planets are together. Right clicking planet(s) to queue a building or set a focus should not change the listbox order or position.

LGM-Doyle:
The PreRender on the ObjectRow was preventing the ObjectPanel cache from
updating correctly for use with SortKey().

Move the PreRender() down the hierarchy into ObjectPanel, means that the
data update (faster) and the layout update (slower) can be separated.

This PR fixes #1128.

### DIFF
--- a/UI/ObjectListWnd.cpp
+++ b/UI/ObjectListWnd.cpp
@@ -1259,7 +1259,7 @@ public:
 
     void                ResourceCenterChanged() {
         RefreshCache();
-        Refresh();
+        RequirePreRender();
     }
 
     std::string         SortKey(std::size_t column) const {
@@ -1271,6 +1271,11 @@ public:
     }
 
     int                 ObjectID() const { return m_object_id; }
+
+    virtual void PreRender() {
+        GG::Control::PreRender();
+        RefreshLayout();
+    }
 
     virtual void        Render() {
         if (!m_initialized)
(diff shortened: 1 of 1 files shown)


MatGB:
Works for me, I have succesfully queued a bunch of gaia transformations and told all my research planets that we've had enough of that thank you.

Issue #1147, opened 2016-11-28
Regression: build candidates list rewinds after adding an item to the build queue


l29ah:
This makes it painful to add multiple tasks of one ship.

MatGB:
Was going to start an issue on this having mentioned it as an aside elsewhere, but yes.

When you add something to the production queue and/or if you change the order size or repeat number, the producible items list jumps, so you can't always see the same thing, especially annoying if, for example, you want several different outpost ships to be given rally points to different places.

In addition, if deleting an item from a long production list, it now jumps to the middle of the list, which makes cleaning up after capturing AI worlds building unnecessary things (eg if they've already lost a capital and are rebuilding high priority buildings) even more frustrating.

l29ah:
7ace2fe8b5362e698f84634eef9b235de6d00e11 is the first bad commit

LGM-Doyle:
@l29ah thanks for the bug report.  

I think that #1150 addresses the issue.  Your comments are welcome.

Vezzra:
Bump. @l29ah, @MatGB, has the issue been solved for you? Can we close this?

l29ah:
Yes.

PR #1150, opened 2016-11-29
Fix bug 1147 production wnd tracking of queue.

LGM-Doyle:
This fixes bug #1147.

This changes the tracking of both the production queue and the build designator queue in the production window.   Now the behavior of both when a new list is loaded is as follows:
- If at the top of the list stay at the top of the list.
- If you were not at the bottom then keep the same first row, even when rows have been added or removed.
- If you at the bottom then preserve the same relative position from the bottom.
- otherwise the list is shorter than before so show the top.


l29ah:
Looks worky.

PR #1152, opened 2016-12-01
Improve listbox initialization of rows.

LGM-Doyle:
This PR has two commits that improve the initialization of list box rows.

One sets the row's width on insert.

The second prerenders rows if and when they are normalized.


geoffthemedio:
Dunno if there is any difference I should notice, but it seems fine...

Issue #1125, opened 2016-11-22
Freeorion crashes upon adding/removing AI


Voker57:
Reproduce: start multiplayer, add AI, crash (not always, like 50%)

Version: d9ffe7dbbdc74f9bc97fa50b38a5ba4fd60edbf7

dbenage-cx:
Looks related to #1109 
[backtrace](https://github.com/freeorion/freeorion/files/608102/gdb_bt_fo_dropdown.txt)

Can not reproduce when using PR #1124 fc3dce5

Vezzra:
Can reproduce the issue on OSX, and also confirm that #1124 fixes the issue.

Therefore, closing as resolved, fixed by #1124.

Vezzra:
@Voker57, please test if that works for you, if not, report back and I'll reopen the issue.

l29ah:
Now it started to crash for me.
```
#0  0x0000038eb3ee3348 in boost::unordered::unordered_set<std::_List_iterator<GG::ListBox::Row*>, GG::ListBox::IteratorHash, std::equal_to<std::_List_iterator<GG::ListBox::Row*> >, std::allocator<std::_List_iterator<GG::ListBox::Row*> > >::empty (this=0x258) at /usr/include/boost/unordered/unordered_set.hpp:188
#1  0x0000038eb3edcb2b in ModalListPicker::CurrentItem (this=0x4c194e9a90) at /home/l29ah/projects/freeorion-teamsmac/GG/src/DropDownList.cpp:189
#2  0x0000038eb3edc8e3 in ModalListPicker::Run (this=0x4c194e9a90) at /home/l29ah/projects/freeorion-teamsmac/GG/src/DropDownList.cpp:167
#3  0x0000038eb3ee05d6 in GG::DropDownList::LClick (this=0x4c194e9460, pt=..., mod_keys=...) at /home/l29ah/projects/freeorion-teamsmac/GG/src/DropDownList.cpp:710
#4  0x0000004c1273c3f5 in CUIDropDownList::LClick (this=0x4c194e9460, pt=..., mod_keys=...) at /home/l29ah/projects/freeorion-teamsmac/UI/CUIControls.cpp:838
#5  0x0000038eb4173e54 in GG::Wnd::HandleEvent (this=0x4c194e9460, event=...) at /home/l29ah/projects/freeorion-teamsmac/GG/src/Wnd.cpp:1098
[...]

Issue #940, opened 2016-08-31
Multi player lobby window layout glitches 


Vezzra:
### Environment
- **FreeOrion Version**: master
- **Operating System**: OSX
- **Graphic card used**: Intel HD Graphics 3000 512 MB
- **Fetched as**
  - Compiled from source
### Description

![multiplayerlobby](https://cloud.githubusercontent.com/assets/10279417/18142469/f3e506c4-6fbd-11e6-9cdb-7be5e2d1f124.png)

Apparently the multiplayer lobby does not adjust itself correctly to bigger font sizes. It has always been a bit cramped, but hasn't been that bad before the change of the default font size.

Vezzra:
I'm not sure if I should tag that for the release... on the one hand, it probably _should_ be fixed, as it's a rather prominently visible glitch, on the other hand, if it can't be fixed without too much trouble, I don't think it's worth to hold up the release for this. Opinions?

adrianbroher:
Is this caused by the new Roboto font? What about reversing that commit and postpone to 0.4.6 + 1 release?

Edit: Never mind, the PR isn't merged yet.

Vezzra:
It's caused by the increased default font size. Because I never use multiplayer, I only discovered this very recently.

TheSilentOne1:
I think the multiplayer setup screen could do without the galaxy graphic. That'd make some space.

adrianbroher:
Should this still be considered an issue? My screenshot uses stock configuration with Roboto.

![bildschirmfoto von 2016-09-04 13-32-50](https://cloud.githubusercontent.com/assets/84891/18230694/4446cfb0-72a4-11e6-8dfe-e0c81a529878.png)

Vezzra:
Looks like this in OSX now (also stock config with Roboto):

![multiplayerlobby2](https://cloud.githubusercontent.com/assets/10279417/18230743/7680115c-72a5-11e6-82c1-490da438a041.png)

Not release stopping, but should be addressed.

Ouaz:
W7, Roboto, default settings:

![mp_setup_en](https://cloud.githubusercontent.com/assets/11946570/18232171/4cf3e4a6-72ca-11e6-8b5d-c9059096aedd.PNG)

But with french language:

![mp_setup_fr](https://cloud.githubusercontent.com/assets/11946570/18232177/672153ea-72ca-11e6-8c60-138a13eda1b6.PNG)

Make the window resizeable should do the trick?

[...]

PR #1153, opened 2016-12-01
Fix multiplayer lobby player list glitches.

LGM-Doyle:
This PR might address the second bug reported in #1125.  I was unable to duplicate the crash myself.

From the backtrace provided by @dbenage-cx backtrace I think that the problem was a valid color selector returned an indication that no color was selected due to the click landing between when the player list rendered and when the color selector rendered.  There are two parts to the problem: the `SelChangeSignal` handler in the color selector does not check if no row is returned, the rendering of the nest color selector is not synchronous with the player list.

The first commit fixes all the locations in the code where drop down `SelChangedSignal` handlers were not checking the returned iterator for validity against `end()`.  Returning `end()` is a valid output indicating nothing was selected.  It is unexpected by the color control, but still valid.

The second commit fixes the glitchiness of the multiplayer lobby, which was due to the nested drop down lists and `Layout` not recursively prerendering everything.  I did not fix `Layout`, I just used a forced prerender with a comment about the underlying problem.

The second commit addresses the way that I imagine the player managed to click on an invalid control, selecting a color before its rows were prerendered.  That is no longer possible, since the player only sees completely prerendered objects.

This PR also fixes the glitching in the multiplayer window.


dbenage-cx:
Color selection no longer bounces around when changing colors.

~~One issue I have not checked for in master yet, is when dropping an AI player:~~
~~```[error] Client : MultiplayerLobbyWnd.cpp:844 : MultiPlayerLobbyWnd::PlayerDataAcceptable found not empty player row with unrecognized client type?!```~~

There are some other glitches with layout I will post as well that I have not looked into yet:
*  Changing system type will change the lobby wnd layout when there are more AI than completely fit.
*  Changing the "host" player type changes the layout

Btw, thanks for all the recent work wrt ListBox

geoffthemedio:
A few quirks, may or may not be specific to this branch / patch...
- When mousewheeling through species without dropping the list, as it updates, the new species initially appears displaced down by a pixel, then jumps back up to the normal position.
- It's too easy to mousewheel over a player / AI to the "drop" option at the bottom of the list, which can unintentionally drop a player / AI / the host.
- The first colour (green) is wider than the rest in the colour picker droplist.
- The player list scrollbar doesn't appear when adding a new player to the list. It does appear after resizing the lobby window, though.
- Text in the player type column appears misaligned... should be centred a bit more left of where it is, perhaps accounting for the part of the width that is taken up by the drop arrow.
- Mouseover highlighting is a bit quirky when mousewheeling through items in lists; move cursor over -> highlight; wheelscroll -> loose highlight; move again within control -> regain highlight.
quirks

LGM-Doyle:
I've addressed the issues that I could replicate.

- Player type is now centered accounting for the `CUIDropList` triangle.
- The color bars were already fixed
- The height of the rows was made to dynamically account for `ClientUI::Pts()` which fixed the missing scroll bar.
- I added code the the `~ModalListPicker()` to call `EndRun()`.  I was not able to replicate the crash that @dbenage-cx reported, but it might be because the `ModalEventPump` is handling signals on a color selector that has been destroyed because this whole page is destroyed on a change in color.

I did not look at mouse highlighting quirkiness.

I could not replicate the jump offset in the species name.

The drop option seems to be the intended behavior of the multiplayer lobby.

geoffthemedio:
The one pixel jump seems a lot more noticeable on the later items in the list (around Trith) than at the top.

Managed to catch it with a screenshot... added line to show the offset.
trith

dbenage-cx:
With default font size, the color is abnormally tall
Overflow of species selection is much more prevalent now, first mentioned in #940
 (again, possibly not in scope)

From previous commit "Fixing glitching in Multiplayer Lobby..."
mp_setup_7d2e9f4

From current commit "In multiplayer lobby add a spacer..."
mp_setup_e5ea05a

geoffthemedio:
Overflow under the drop arrow can be resolved by widening the window. It should probably be hidden by the button though.

LGM-Doyle:
More multiplayer lobby fixes.
- Color swatch width really fixed this time.  It was broken in master and when I rebased this PR to master the bug appeared.
- Color swatch height is now smaller than the full row height.
- I fixed the species selector species name overflowing its drop down list anchor, by adding child clipping to the anchor width.  This also entailed adding separate `DroppedRowWidth()` function from the `DisplayedRowWidth()` so that the `CUIDropDownList()` can provide two different widths to account for the drop down arrow.
- I adjusted the size of the `SpeciesSlector` which may fix the vertical name glitch.  I was able to replicate this, but I may have just moved the glitch to font sizes that I did not try.


Problems that I don't understand are:
- @dbenage-cx Changing the system type (I assume you mean galaxy shape) will change the lobby layout.  I don't see this problem.
- @geoffthemedio The mouse over highlighting.  It sounds to me like you are mouse scrolling away from the selected row, which is why it disappears.

Unfixed problems that I have replicated
- Changing the host type from Human to Moderator or Observer shifts the list up/down by several pixels.

geoffthemedio:
@LGM-Doyle No; I position the mouse over the droplist, at which point it's highlighted. I can then raise the mouse so I'm sure the cursor isn't moving, then click the scrollwheel. The highlighting goes away. If I then move the cursor within the control, the highlighting comes back.

Noticed another issue with the lobby: the column headings don't resize when the list resizes. (Not due to this PR, but just noticed while testing...)

dbenage-cx:
I did mean galaxy shape, you have since fixed this but the issue was that the scroll bar would reappear when galaxy shape was changed, then disappear when a control in lobby wnd was changed.

LGM-Doyle:
I've fixed the Moderator and Observer row heights.

The column headers are not part of the list box, so they are not part of a Layout object, so they won't resize.  I'm inclined to leave them as is.

I still can't duplicate the highlighting issue.  The steps that I am following are: position mouse over an open droplist, click mousewheel, see selection highlight disappear, move cursor and see it reappear.  Does this only happen in the multiplayer droplists or is it a general problem?  If it is a general problem could we open an issue and solve it separately from cleaning up the multiplayer lobby. 

The two outstanding issues are the mousewheel on player type causing drop and the headers not resizing with the list.  I think that both are working as intended, although not in an ideal manner.

geoffthemedio:
Are those the steps you're following and the results you're seeing, or what you're unable to replicate despite following the steps? (Because that sounds like a description and results illustrating the problem).

I only see the problem in the multiplayer lobby player list, and it happens for all the droplists in the player row.

Without looking into the code, I assume it's due to the rows / their contents being recreated, and a new droplist control being created under the cursor not receiving a mousehere event to cause it to be highlighted, until the cursor moves.

LGM-Doyle:
Those are the steps that I am following that are **not** showing the problem for me.

Your assessment of the problem sounds correct, except I don't think that the new selection is sent to the server and the rows recreated until the droplist closes.  I could check that and make that change tomorrow, speculatively, if you will test it.

geoffthemedio:
Ah... just noticed: the droplist is not open when doing this. I'm mousewheeling over an undropped droplist in the players list.

UI/MultiplayerLobbyWnd.cpp
```diff
@@ -790,6 +820,12 @@ bool MultiPlayerLobbyWnd::PopulatePlayerList() {
 
     Refresh();
 
+    // This PreRender forces an 'extra' prerender to make the nested
+    // dropdown lists to prerender after  their rows have
```
geoffthemedio: extra space after after

geoffthemedio:
I can't merge this... every time I try, it says
````
Merge attempt failed
Base branch was modified. Review and try the merge again.
````
There's no indication that it can't be merge or otherwise requires a rebase, though...

**Comment by LGM-Doyle** [comment]
I fixed the spacing error, rebased it to master and did a test merge, which worked.

I figured out what is happening with focus disappearing, when you said that it happens when the drop down box is closed.

The multiplayer lobby list is recreated entirely when the server sends new information  It is not re-populated, it is re-created with new controls.  The new controls are created at the same position as the old ones, but they do not necessarily have the same data.  They also do not automatically steal focus.

I could not find a simple fix.

I think that the correct fix is to refactor the multiplayer lobby to keep persistent controls and update them with the new data from the server.  In that case, the focused control would  also be persistent.

I do not want to do that much work for this bug.

File: GG/CMakeLists.txt [hunk 2]
```diff
@@ -23,17 +23,6 @@ option(BUILD_SHARED_LIBS "Enables building of shared libraries; OFF implies stat
 
 cmake_dependent_option(USE_STATIC_LIBS "Favour static linking of depencendies." OFF BUILD_SHARED_LIBS OFF)
 
-if (BUILD_DEBUG)
```
geoffthemedio, line 26: is this supposed to be in this pull request?

LGM-Doyle:
@adrianbroher did you push, about a day ago, these three commits
https://github.com/freeorion/freeorion/pull/1153/commits/e5b27050beb79bec2f15d7501137c3496df53172,
https://github.com/freeorion/freeorion/pull/1153/commits/c6d1348c59bc8f30a055d3b54a4378d66c6095b0,
and https://github.com/freeorion/freeorion/pull/1153/commits/39e083e4bd5a90e92f06ab73be523b121bd86cd6 here by accident instead of to master?

dbenage-cx:
I had an issue with these 3 showing up on another branch as well (~~I believe they were merged to master through the Fighters branch~~ Edit: This occured in the same time frame Fighter branch was merged).
`git diff upstream/master` would not show them on the local branch, only on github compare.
Neither rebasing nor cherry-picking to a new branch with the same name seemed to resolve this, I was able to resort to a new branch name in that circumstance.

LGM-Doyle:
@dbenage-cx thanks for the comment.  I'm trying your solution.

I was perplexed, because I rebased, glanced at it in gitk, did a test merge on the command line and assumed github used the same merge resolution as basic git.

LGM-Doyle:
@adrianbroher, sorry for implying that your actions had something do do with the strange appearance of those commits in this, and apparently other PRs.

Vezzra:
@LGM-Doyle: I assume you closed this as resolved, is that correct? Assuming it is, I going to tag the issue accordingly, if I'm mistaken, please set the correct "status:*" label.

Vezzra:
@LGM-Doyle: Argh, forget my last comment, I just saw the "closed" status and thought this is an issue, instead of a PR.

However, that still leaves the issue that you didn't set any "status:*" tag to indicate why you closed the PR. Has it become obsolete (in this case tag it as "invalid"), or superseded by another PR (in this case set the "follow-up" tag and post a reference to the superseding PR)?

I assume the latter, with #1160 being the superseding PR, but I'm not 100% sure, so I leave it to you to set the tag and post the reference 😃

LGM-Doyle:
This PR was superceded by #1160.

A bug in github caused 3 PRs from the recently merged #587 Fighters branch to replicate in multiple subsequent PRs.  @dbenage-cx pointed out that changing the name of the branch and creating a new PR fixes the problem, so I did that.

Vezzra:
Thanks!

PR #1154, opened 2016-12-01
PreRender specials panel after changes during prerender.

LGM-Doyle:
This fixes the last flickering item that I saw in the sidepanel, for issue #1093.

Changing the specials panel during prerender causes changes to be
scheduled for layout in the next prerender, which appears as a visible
glitch.

This commit forces a prerender of the specials panel at the end of 
PlanetPanel::DoLayout() which is only called during the prerender phase.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -1069,6 +1069,11 @@ void SidePanel::PlanetPanel::DoLayout() {
 
     Resize(GG::Pt(Width(), std::max(y, min_height)));
 
+    // DoLayout() is only called during prerender so prerender the specials panel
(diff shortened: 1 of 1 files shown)


geoffthemedio:
For something this small, I think you can just commit without a pull request first...

LGM-Doyle:
OK.

It was also to point out that I think #1093 is fixed, unless someone points out a specific item that is flickering.

geoffthemedio:
There's no visible jitter in the position of the sidepanel controls, but the stealth stat icon does seem to flicker (go black) briefly - perhaps a frame - when pressing turn. (In the case where all the sub-panels are collapsed)

Issue #1148, opened 2016-11-28
improve visibility of sensor ghosts


TheSilentOne1:
Bug reports like this one ( https://github.com/freeorion/freeorion/issues/1130 ) bring it to attention that it can sometimes be difficult to recognize the scanlines over a ghost fleet. An easy fix would be to change the scanline color. The following screenshot shows the change of a single line of scanlines.frag, adding some green:
const vec4      half_gray = vec4(0.0, 0.8, 0.0, 0.5);   // darkened pixels

![invisible_fleets](https://cloud.githubusercontent.com/assets/12985960/20671558/118d06c6-b57e-11e6-8db9-67cde4672e12.jpg)

It is obvious that the color of the scanlines over nebulae, systems etc. would better not be changed. So I'd propose to just change the color for ghost fleets, which would require a modification of the shader so it takes a color as an argument (replacing the half-gray variable).
Other ideas are certainly welcome!

geoffthemedio:
How does that colour look interacting with the various other empire coloured detection range circles?

LGM-Doyle:
How about de-saturating the ghost's color, so that they look white, pale and ghostly.  

It would conflict with the new `White Kraken`.

TheSilentOne1:
I'm working some more on this, will report next week (probably).

TheSilentOne1:
Adressed with #1155.

PR #1155, opened 2016-12-05
fine-tuning the scanline shader

TheSilentOne1:
In reference to https://github.com/freeorion/freeorion/issues/1148:
This pull request improves the use of the scanline shader (sls) throughout the game by allowing to assign a color to it. Currently, the sls is used for the fleet and ship data panel on the fleet window, planets on the sidepanel, as well as fleet buttons, systems and fields on the map screen. For the sls of fleet buttons, the complementary color of the fleet buttons is used, which helps a lot with the sensor ghost issue referenced above. All the other uses of the sls have been fine-tuned.

As always, I'm especially grateful for advice regarding the C++ code.

sls-test

### DIFF
--- a/UI/CUIControls.cpp
+++ b/UI/CUIControls.cpp
@@ -2232,12 +2232,14 @@ namespace {
     ScanlineRenderer scanline_shader;
 }
 
-ScanlineControl::ScanlineControl(GG::X x, GG::Y y, GG::X w, GG::Y h, bool square /*= false*/) :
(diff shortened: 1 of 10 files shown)


geoffthemedio:
You appear to have just used the complementary colour on the map, but not the fleet window lists, or (presumably) the objects list... why?

TheSilentOne1:
The black scanlines work okay on the fleet window. However, due to their small size and being in front of a black background, the scanlines of the fleet buttons need a different treatment to be visible (enough). That's why they have been given the complementary color.

PR #1156, opened 2016-12-05
Fix in ListBox::KeyPress() twiddling left right causes crash.

LGM-Doyle:
As described currently master will crash if you open any single column unnormalized list box and twiddle the right and left keyboard buttons for a random length of time.

The problem was that the unnormalized ListBoxes, which now most are do not initialized the col_widths.  

This PR uses instead the column widths of the first visible row which exist because it is visible.

### DIFF
--- a/GG/src/ListBox.cpp
+++ b/GG/src/ListBox.cpp
@@ -681,17 +681,22 @@ ListBox::iterator ListBox::LastVisibleRow() const
 
 std::size_t ListBox::LastVisibleCol() const
 {
-    X visible_pixels = ClientSize().x;
-    X acc(0);
-    std::size_t i = m_first_col_shown;
-    for (; i < m_col_widths.size(); ++i) {
-        acc += m_col_widths[i];
-        if (visible_pixels <= acc)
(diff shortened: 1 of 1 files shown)


GG/src/ListBox.cpp
```diff
@@ -1533,25 +1538,29 @@ void ListBox::KeyPress(Key key, boost::uint32_t key_code_point, Flags<ModKey> mo
...
-            if (m_first_col_shown) {
-                --m_first_col_shown;
-                m_hscroll->ScrollTo(
-                    Value(std::accumulate(m_col_widths.begin(), m_col_widths.begin() + m_first_col_shown, X0)));
-                SignalScroll(*m_hscroll, true);
-            }
+            if (m_first_col_shown == 0)
+                break;
+
+            --m_first_col_shown;
```
geoffthemedio, line 1544: could do with some comments to explain this rather large mess of nested calls... or can you use some temporaries to make it more readable?

regardless, though, if it fixes a crash, you can probably just merge it as is and fix later...

LGM-Doyle:
I've added more temporaries.

PR #1158, opened 2016-12-06
Update GG/COPYING with current LGPLv2.1 contents

subpop:
The bundled libgigi COPYING file contains a LGPL-2.1 license with the old FSF address. This updates the COPYING file with the current LGPL-2.1 contents, found at http://www.gnu.org/licenses/lgpl-2.1.txt.

### DIFF
--- a/GG/COPYING
+++ b/GG/COPYING
@@ -1,8 +1,8 @@
-		  GNU LESSER GENERAL PUBLIC LICENSE
-		       Version 2.1, February 1999
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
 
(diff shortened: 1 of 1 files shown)


GG/COPYING
```diff
@@ -1,16 +1,16 @@
-		  GNU LESSER GENERAL PUBLIC LICENSE
-		       Version 2.1, February 1999
+                  GNU LESSER GENERAL PUBLIC LICENSE
```
geoffthemedio, line 1: Why the tabs -> spaces? Was the license text changed, or originally written that way?

subpop:
I copied the contents of the license into the GG/COPYING file and this was the diff it produced, so I would guess the license change its formatting from tabs to spaces.

PR #1159, opened 2016-12-07
Fix bug in SidePanel::DoPanelsLayout.

LGM-Doyle:
DoPanelsLayout is unstable when a PlanetPanel that is narrowed for the
scroll bar is also shorter.  This results in the layout flickering
between a SidePanel with vscroll and without.

PlanetPanels with expanded buildings can be shorter with the scroll bar
because it might resize the building icons to be smaller when the scroll
bar is added.

This commit just keeps the layout with the scroll bar even if adding the
scroll makes the panels short enough to not require the scroll bar.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -603,6 +603,8 @@ private:
     void            DisableNonSelectionCandidates();    //!< disables planet panels that aren't selection candidates
 
     void            DoPanelsLayout();                   //!< repositions PlanetPanels, without moving top panel.  Panels below may shift if ones above them have resized.
+    /** Resize the panels and \p use_vscroll if requested.  Return the total height.*/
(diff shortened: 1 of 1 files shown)


UI/SidePanel.cpp
```diff
@@ -2579,51 +2581,49 @@ void SidePanel::PlanetPanelContainer::DoPanelsLayout() {
 
     GG::ScopedAssign<bool> prevent_inifinite_recursion(m_ignore_recursive_resize, true);
 
-    GG::Y y = GG::Y0;
-    GG::X x = GG::X0;
+    // Determine if scroll bar is required for height or expected because it is already present.
+    GG::Y available_height = Height();
```
geoffthemedio, line 2585: Use GG::Y0

geoffthemedio (changes requested):

LGM-Doyle:
Fixed.

PR #1163, opened 2016-12-09
Fix use after free error.

o01eg:
Fix use-after-free error. Shutdown socket before closing and use shared pointer to prevent free while handlers are active.

### DIFF
--- a/network/ServerNetworking.cpp
+++ b/network/ServerNetworking.cpp
@@ -72,8 +72,14 @@ PlayerConnection::PlayerConnection(boost::asio::io_service& io_service,
     m_disconnected_callback(disconnected_callback)
 {}
 
-PlayerConnection::~PlayerConnection()
-{ m_socket.close(); }
+PlayerConnection::~PlayerConnection() {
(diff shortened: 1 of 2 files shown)


network/ServerNetworking.cpp
```diff
@@ -72,8 +72,14 @@ PlayerConnection::PlayerConnection(boost::asio::io_service& io_service,
     m_disconnected_callback(disconnected_callback)
 {}
 
-PlayerConnection::~PlayerConnection()
-{ m_socket.close(); }
+PlayerConnection::~PlayerConnection() {
+    boost::system::error_code ec;
+    m_socket.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
+    if(ec) {
```
geoffthemedio: space after if

geoffthemedio (changes requested):

o01eg:
I've added space.

geoffthemedio:
I don't really understand the relevant details, but online examples suggest this is a reasonable way to call async_read

PR #1168, opened 2016-12-11
Use human player's name as empire's name.

l29ah:


Vezzra:
Anyone any comments/opinions on this?

Personally, I'm not strongly opposed to the idea, but also not really in favor of... having the player name as the (default) empire name might satisfy the personal preferences of some players, but I wonder if it's the majority...?

l29ah:
On Fri, Dec 16, 2016 at 06:27:48AM -0800, Vezzra wrote:
> Anyone any comments/opinions on this?
> 
> Personally, I'm not strongly opposed to the idea, but also not really in favor of... having the player name as the (default) empire name might satisfy the personal preferences of some players, but I wonder if it's the majority...?

Sure it's a tiny minority, as it seems like we're the only people on this planet that play freeorion multiplayer.

-- 

Vezzra:
I don't think a preference for that feature is necessarily linked to playing multiplayer (although I can see why you'd have a stronger preference for it in that case), players of single player games might prefer that too.

As I said, I'm not strongly opposed, except for quickj test games (where I don't bother what my player/empire names are anyway), I always think up a name for my empire, never take the default.

So maybe let me rephrase my question: Has anyone any objections to merging this?

MatGB:
Will it override a persistent configuration entry? I'd rather it respected that beyond that I'm happy with the idea, I'd rather this than Terran Imperium, and I've regularly forgotten to change mine from the CHANGEME I've set.

Vezzra:
Hm, looking at the code it seems that this change only applies to multiplayer games, but I can't tell if it will override persistent configuration entries. You'll probably have to run a quick test with this PR.

Don't have the time to do this myself, I'm hopelessly swamped... :frowning_face:

geoffthemedio:
I don't think there are any config entries for multiplayer player settings

Vezzra:
Ok, just ran a quick test myself, indeed applies only to multiplayer and makes sense the way it's implemented. Seems to work as intended, so as no objections have been raised and the changes are pretty simple and straightforward, I'm going to merge.

PR #1169, opened 2016-12-11
Fix wrong tooltip in tech controls

Ouaz:
- "Partially Unlocked" tooltip is wrongly displayed instead of "Available" for the corresponding tech control in the Research Window.


Vezzra:
That's actually a bugfix, so I changed the category label accordingly.

PR #1171, opened 2016-12-12
Make planet stealth techs effect only owned planet not all system planets

LGM-Doyle:
Recent changes to allow system renaming allow systems to be owned if only a single empire owns
planets in a system.

The current version of the planetary stealth techs applies to all
planets in a system if the system is owned by the empire.  This means
that uncolonized planets in a system with colonies/outposts controlled
by a single empire pick up the stealth boost.

This commit reduces the scope of the effect to only the owned planet.

I was not sure what the original intent of the `Or` clause with system and planet was in the original code.

I also looked through all uses of `OwndedBy` in the scripting and did not see other examples of this problem.

### DIFF
--- a/default/scripting/techs/spy/STEALTH_1.focs.txt
+++ b/default/scripting/techs/spy/STEALTH_1.focs.txt
@@ -10,10 +10,7 @@ Tech
         EffectsGroup
             scope = And [
                 OwnedBy empire = Source.Owner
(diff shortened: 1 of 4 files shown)


MatGB:
I had never noticed this was the case despite much playing with/against stealth. Good catch, will review in a bit.

PR #1072, opened 2016-10-25
Add Combat Simulator

Morlic-fo:
This PR contains a Monte-Carlo style simulation for combat between two factions.
## SETUP

To use it, first check the file default/python/AI/SimulationShiplist.txt.
Make sure, the two lines
`FACTION_A` and `FACTION_B`
exist.
Any line under `FACTION_A` is interpreted as ship design name which is aligned with the first faction. If it doesn't exist or if it is spelled wrong, you will be informed when running the script.

Any line that comes under `FACTION_B` will be interpreted the same but added to the second faction.

So

```
FACTION_A
SD_MARK_1
SD_MARK_1
SD_MARK_1

FACTION_B
SD_AST_2
```

will be a battle between 3 SD_MARK_1 ship designs for faction A and a SD_AST_2 for faction B.

The design names used must be defined as usual by FOCS (see default/scripting/ship_designs)
## HOW TO RUN

1) Start Single Player Game with a single AI
2) write the command "simulate" into the game chat.
3) Read Simulation results in chat
Optional:
4) Edit SimulationShipList.txt to add/remove ships and save the file.
5) Run simulation again by writing the command "simulate" into the game chat. The new ships will be used for the simulation.
## EXAMPLE OUTPUT

```
AI_1: Adding Ships to Faction A now.
AI_1: Adding ship with name SD_MARK_1
AI_1: Adding ship with name SD_MARK_1
AI_1: Adding ship with name SD_MARK_1
AI_1: Adding Ships to Faction B now.
AI_1: Adding ship with name SD_AST_2
AI_1: Adding ship with name SD_AST_22
AI_1: ERROR: Could not find design.
AI_1: Number of ships A: 3 (cost: 140.4 PP)
AI_1: Number of ships B: 1 (cost: 118.6 PP)
AI_1: +++ SIMULATION RESULTS+++
AI_1: Simulated a total number of 150 iterations.
AI_1: Wins A: 0 (0.0%)
AI_1: Wins B: 88 (58.7%)
AI_1: Draws: 62 (41.3%)
AI_1: Losses Player A: (121.1 +- 23.1) PP
AI_1: Losses Player B: (0.0 +- 0.0) PP
```

The first lines are for debugging. Here, you can see that the design "SD_AST_22" was not found because there was a typo in it. The simulation was still performed with the remaining ships.

A summary of ship count and cost comparison is given (cost is influenced by game start fleet upkeep, not sure how to avoid this) followed by the simulation results.
[...]

PR #1175, opened 2016-12-15
Combat simulator

Morlic-fo:
Follow-up on #1072 

Not ready for merge.


Dilvish-fo:
Nice work here!

Once this gets at or close to final, it would be great to write up a Wiki article about how players could use it to test/evaluate their designs, and then we could include the URL for that in the Pedia article.

One area that seems deserving of some more attention is regarding the accuracy of the predictions and possibly a corresponding recommendation for number of iterations to use.  You do report the (sample) standard deviation for the PP_lost numbers, which is useful but different than the accuracy of prediction and wouldn't be useful for the win/loss rates (which I expect is why you don't report the standard deviation for that).   To help illustrate why standard deviation isn't enough for us regarding win/loss, and an accuracy estimator that probably *would* be helpful, let us consider the indicator variable I'll call A_Won, corresponding to the test ```r == WIN_A``` for which you report the mean value in your code.  With our default number of 150 samples the difference between the unbiased sample standard deviation and the population (or biased sample) standard deviation is quite small, and we can simplify our math a bit more using an alternate form for the population variance:
 formula

For our indicator variable, having only values of 0 or 1, < x<sup>2</sup> > is the same as < x >, and you could also reorganize that in the form more common for a series of binomial trials with p as <x> the chance of A winning: variance = p(1-p), which clearly is going to be rather high in comparison to p for any situation where one side is not tremendously overpowering the other.  If A won 100 out of the 150 trials, then  < x > is 0.6666 (we would report as "Wins A: 66.7%"); the biased sample standard deviation would be (2/3 - 4/9)<sup>0.5</sup> = 0.4714  (and the unbiased sample standard deviation would be 0.4729), nearly as high as our expected value for the win rate itself.

Furthermore, [as noted here](http://www.northeastern.edu/afeiguin/phys5870/phys5870/node71.html), simply increasing our number of trials would not generally reduce that standard deviation, so it is unhelpful in assessing accuracy.  Instead, as also noted and justified there, a typical way to assess accuracy would be to further divide that standard deviation by the square root of the number of samples.  This esimate is also sometimes referred to as [the standard error of the mean](https://en.wikipedia.org/wiki/Standard_error#Standard_error_of_the_mean).  In our particular example this works out to an error estimate of 0.038, which seems plenty small.

Given that our max variance for probability of win is going to be 0.5*(1 - 0.5) = 0.25, for a max standard deviation of 0.5, with even just 25 trials we'd have a max error estimate of 0.1.  To save on computing time if we have the AI doing things like this as a normal course of business, we could have it do something like start out with just 25 trials, calculate the estimated win probability and estimated error for that, and in most cases that would probably be enough.  (Perhaps if we had the maniacal AIs willing to take on a fight with just a 55% win estimate it would run more trials to get the estimated error below 0.05 or something.)

================

A different approach which I have been mulling for a long time would be a sort of dynamic programming approach to calculate the full distribution of all possible outcomes; I think the trick would be figuring out enough ways to collapse the number of state variables that really need to be tracked from one round to the next, and to handle really big battles it might require doing a sort of balls-and-bins approach instead of figuring out each attack separately.  But, working through the error estimate issue above, it's seeming like the approach here is likely to work out plenty accurate enough (and we could also add more tracking for different ranges of outcomes also besides just the simple win/loss/draw).

Morlic-fo:
Good summary of statistics 101, Dilvish. You are correct regarding the uncertainty in the estimation of the winrate.

However, calculating the chance to win the combat is only a minor part of what I have in mind with these simulations. 

Consider the following situation (obviously a hyperbole to emphasize the point):
We have just won a battle and now all of our 100 surviving ships are at 1 structure. The enemy has a fleet of 5 small ships in the next system, each of which has 2 Mass Drivers.
Now, we could engage the enemy and win the battle by 100%. However, that is obviously a really bad idea as we are going to lose many of our expensive ships. Retreating to repair certainly would be the right call. 
On the other hand, it is perfectly reasonable to sacrifice a small ship to destroy an enemy solar hull full of Death Rays even if the rest of the enemy fleet survives (i.e. we lose the fight).

So, what we are actually interested in most of the time is not so much the chance to win the battle but rather the losses we inflict on the enemy as compared to our own (possibly weighted by the industrial output of the empires).

Now the 0th order approximation of the losses is to calculate the mean or median losses. A more reasonable approach is to consider the statistical distribution of losses. The standard deviation is some measure for that but probably not too accurate considering the distribution will often deviate from a gaussian one. Still, it gives a first idea in a single easy-to-calculate value. Alternatively, we could look at the actual quantiles in the simulation outcomes.

It also is important to realize that an accurate estimation of the PP losses requires many times more iterations than an accurate estimation of the winrate. This follows directly from the fact that there are many different combinations of damaged/destroyed ships each of which has a different contribution to the lost PP and each of which has a very small chance to occur (because there are so many different outcomes).

For example, consider the example above: We have 100 ships each on 1 structure that all cost some random, uncorrelated amount of PP and also 1 enemy with 1 Mass Driver. We are going to lose exactly one ship. Now we need at least 100 iterations to find all possible outcomes. Obviously, in practice we will need many more iterations as many outcomes will repeat several times before others occur for the first time.
At the same time, we are relatively quick to arrive at the conclusion we have a winning chance of >99%.


Obviously, computation time is an issue in practice and we do need to impose a limit on the number of iterations. I agree with you that the number of iterations should probably be dynamically calculated based on the outcomes but also on the importance of the fight. Having the AI calculate 20 seconds 1-2 times in a game before some deciding battle may be reasonable. Having it calculate 20 seconds for some random battle each turn certainly is not.

I also hope that by using a trust-but-verify approach we can cut down the number of simulations that actually have to be performed. We already have some rough approximation of the combat outcome by our combat rating. If our combat rating suggests a clear outcome, we verify this by simulation. If the first simulation agrees with really minor (relative) losses, then that is good enough.

The better heuristic for the combat rating we can implement, the less we have to simulate and the more iterations we can run on the actual significant simulations.





> A different approach which I have been mulling for a long time would be a sort of dynamic programming approach to calculate the full distribution of all possible outcomes;

If that turns out to be feasible, that would certainly be useful. Even approximating the "state" as some statistical distributions would be fine if the loss of information is small enough so that the final result will be within some margin of error to the actual outcomes.

We would need a way to verify the implementation of such a model, so probably need a working simulation either way...

PR #1176, opened 2016-12-16
Pedia: New subcategory sorting in "Guides" and "Game Concepts"

Ouaz:
- Move "Interface" category as subcategory in "Guides"
- Create an "Encyclopedia subcategories" section in stringtables
- Create folders and subfolders corresponding to Categories and Subcategories in scripting/encyclopedia

See:  http://www.freeorion.org/forum/viewtopic.php?p=86745#p86745


dbenage-cx:
Should the articles for Damage, Detection, and Technologies be moved to `game_concepts/` as well?

Ouaz:
> Should the articles for Damage, Detection, and Technologies be moved to game_concepts/ as well?

Actually, `Damage`, `Detection` and `Technologies` are still in `Game Concepts` category in this PR.

I would like to move `Damage` in `Meter Type/Ship Meters` and` Detection` in `Meter Type/Empire Meters` (new subcategories), and `Technologies` article in top of the `Technology` category.

But these categories are hardcoded (in UI/EncyclopediaDetailPanel.cpp ?) so I'm not able to do that.

In fact, all categories (except `Guides` and `Game Concepts`) are hardcoded and built in the pedia according corresponding tags in the focs files (`FieldType`, `BuildingType`, `Tech`, etc), so I don't know how to add subcategories in them (like my list on the FO forum).

dbenage-cx:
I only intended to refer to the actual files for these entries, they are the only remaining entries in the root encyclopedia directory.  Are these different from (e.g.) `FIELDS.focs.txt` in some way?

Possibly need some keyword for listing, like `list OBJECT_TYPE TAG_NAME`, before removing hard-coded articles to definitions.  I will try to post some thoughts on this tonight/early morning on forum.

dbenage-cx:
For Technology, changing the article to:

    name = "ENC_TECH"
    category = "ENC_TECH"


Will preface the Technology category with the descriptive text before listing contents.

Ouaz:
Thanks for the trick, done.

I did the same thing to move the "Interstellar Fields" article content (from Game Concepts) in top of the Field Type list.

In FIELDS.focs.txt:

```
name = "ENC_FIELD_TYPE"
category = "ENC_FIELD_TYPE"
```

dbenage-cx:
Instances of `[[encyclopedia TECHNOLOGIES_TITLE]]` need update to new article name.

Ouaz:
> Instances of [[encyclopedia TECHNOLOGIES_TITLE]] need update to new article name.

Oops, forgot it.

But` ENC_TECH` value is `Technology` (as each pedia category is in singular form).
`TECHNOLOGIES_TITLE` value was `Technologies`.

Do I change `ENC_TECH` value to `Technologies` or simply remove the `TECHNOLOGIES_TITLE` pedia links (only 3 occurrences)?

dbenage-cx:
I'd opt for `Technologies`. (Some other directories use a plural form, e.g. `Monster Types`.)

Ouaz:
> I'd opt for Technologies.

Done (forced update).

geoffthemedio:
Is this relatively stable / complete now, other than stuff that would require C++ code changes?

dbenage-cx:
Couple of inconsistent file locations, can probably be addressed in next set of changes.
Outside of that, LGTM

Ouaz:
IMO, to do in the future :

- Shorten the "Technologies" introduction article, because now, the start of the list is too low in the pedia window
- Move technologies.focs.txt in `scripting/techs`, and fields.focs.txt in `scripting/fields`.
- Make visually more obvious that a pedia entry is either a category, subcategory, or direct article (with bullets, brackets or whatever)

Ouaz:
> move ... in or into?

Oh, there's a difference? :p Sorry for my bad english then.

I mean move the files into (?) the correct folders (`technologies.focs.txt` and `fields.focs.txt` are still in `scripting/encyclopedia` folder, whereas their content are now used in `Technologies` (`ENC_TECH`) and `Field Type` (`ENC_FIELD_TYPE`) pedia category.

In order to shorten the `Technologies` article, I will submit soon a new PR in which the text part about how a ship receives a tech upgrade, becomes a beginner hint in the pedia.

geoffthemedio:
Do you know any German grammar? "Into/in the park" is like the difference between "in den/im Park" (accusative vs. dative case). English has just a single objective case, but "into" implies a change of location / a destination for a movement, and "in" specifies a location / area in which something occurs.

Anyway, is moving the files something you can't do yourself?

dbenage-cx:
> Make visually more obvious that a pedia entry is either a category, subcategory, or direct article (with bullets, brackets or whatever)

Hopefully will be covered in #755 , removing such additions from the bottom of articles and placing them in the navigation panel/sidebar.  Need to work out how to custom define categories for generated articles.

Ouaz:
> English has just a single objective case, but "into" implies a change of location / a destination for a movement, and "in" specifies a location / area in which something occurs.

Thanks for the explanation! In french, we have "dans" et "vers" ("déplacer dans" ou "déplacer vers"). "Vers" is more correct, but "dans" works too.  So I made the mistake in english ("in" instead of "into"). ^^

> Anyway, is moving the files something you can't do yourself?

I tried, but if I move `technologies.focs.txt` into `scripting/techs` folder, the article content is no more displayed in the pedia on top of the generated tech list. Same thing with `fields.focs.txt`, if I move it into `scripting/fields`.

I don't know why, but I assume these files are declared somewhere as being currently in `scripting/encyclopedia`?

geoffthemedio:
More likely it just expects encyclopedia articles to be in the pedia directory, and tech definitions to be in the techs directory, and won't go searching through every directory to find any definitions hidden in files not in the expected place.

Ouaz:
As the "Technologies" article content is now displayed in top of the tech generated list, I thought it would be better to have `technologies.focs.txt` in root of the `scripting/techs` folder:
```
construction
defense
[...]
specials
spy
Categories.inf
technologies.focs.txt
techs.macros
```

Same with` fields.focs.txt` in `scripting/fields`.

But if doesn't make sense, I'm OK to let these two files in `scripting/encyclopedia`. ^^

geoffthemedio:
I don't understand the motivation, but regardless, https://github.com/freeorion/freeorion/blob/master/default/scripting/encyclopedia/TECHNOLOGIES.focs.txt contains a pedia article definition, so needs to be in the encyclopedia directory. If it's in https://github.com/freeorion/freeorion/tree/master/default/scripting/techs the parser would attempt to parse tech definitions from it, which won't work, and it won't be included in the pedia since the pedia parser won't know where to find it. It looks in particular directories for particular definition types, eg: https://github.com/freeorion/freeorion/blob/master/parse/TechsParser.cpp#L248

Ouaz:
That was I didn't know. Thanks again for the explanation!

geoffthemedio:
>> English has just a single objective case, but "into" implies a change of location / a destination for a movement, and "in" specifies a location / area in which something occurs.

>Thanks for the explanation! In french, we have "dans" et "vers" ("déplacer dans" ou "déplacer vers"). "Vers" is more correct, but "dans" works too.  So I made the mistake in english ("in" instead of "into"). ^^

Actually, "in" can be used in the sense of "into" as well, like "put it in(to) the box"... Either would be fine, as it's not ambiguous what is meant in that case. But when you said "move the files in the folder" it sounds like you mean "move the files, which are presently in the folder, to another unspecified location". Saying instead "move the files into the folder" removes the ambiguity.

But regardless, you mentioned a few things you want to do in future. Do those need to go into this pull request, or is it done?

Ouaz:
I think it is done and ready to merge.

The only thing I'm able to do is to shorten the Technologies article, making a part of it a new beginner hint (in another PR, though).

Note: I updated the last commit -> I moved some remaining focs files:

`TECHNOLOGIES.focs.txt` and `FIELDS.focs.txt` at root of `scripting/encyclopedia`
`DAMAGE.focs.txt` into `scripting/encyclopedia/game_concepts/ship_management`
`DETECTION.focs.txt` into `scripting/encyclopedia/game_concepts`

PR #1178, opened 2016-12-16
Fix WeaponDamageImpl for fighters

dbenage-cx:
Changes reported damage from fighters to not include a specified damage reduction.

Removes addition of additional hangar fighters, as the meter is shared/stacked across other hangars. This fixes an issue initially reported in [forum](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10344)

Posting for review to ensure there is no damage reduction other than shields, and the stacked meter value is not due to some further issue.


dbenage-cx:
Rebased with additions for `ShipDesign::AdjustedAttack()`

geoffthemedio:
As far as I know this should be correct.

PR #1180, opened 2016-12-18
Fractional progress

geoffthemedio:
Changes storing of progress on production items to a fraction of completion rather than accumulated PP. This should result in better handling of situations where costs change after partial completion, such as preventing extra pp per turn being accumulated while at high cost, then quicker-than-intended completion after the cost is reduced while accumulated PP are retained.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -86,23 +86,23 @@ namespace {
 
         item_cost *= queue_element.blocksize;
         build_turns = std::max(build_turns, 1);
-        float element_accumulated_PP = queue_element.progress;                                 // PP accumulated by this element towards producing  next item
-        float element_total_cost = item_cost * queue_element.remaining;                        // total PP to produce all items in this element
-        float additional_pp_to_complete_element = element_total_cost - element_accumulated_PP; // additional PP, beyond already-accumulated PP, to produce all items in this element
-        float additional_pp_to_complete_item = item_cost - element_accumulated_PP; // additional PP, beyond already-accumulated PP, to produce the current item of this element
+        float element_accumulated_PP = queue_element.progress*item_cost;                        // effective PP accumulated by this element towards producing next item. progress is a fraction between 0 and 1.
+        float element_total_cost = item_cost * queue_element.remaining;                         // total PP to produce all items in this element
+        float additional_pp_to_complete_element = element_total_cost - element_accumulated_PP;  // additional PP, beyond already-accumulated PP, to produce all items in this element
+        float additional_pp_to_complete_item = item_cost - element_accumulated_PP;              // additional PP, beyond already-accumulated PP, to produce the current item of this element
         float basic_element_per_turn_limit = item_cost / build_turns;
         // the extra constraints on frontload and topping up amounts ensure that won't let complete in less than build_turns (so long as costs do not decrease)
         float frontload = (1.0 + frontload_limit_factor/std::max(build_turns-1,1)) * basic_element_per_turn_limit  - 2 * EPSILON;
         float topping_up_limit = basic_element_per_turn_limit + std::min(topping_up_limit_factor * item_cost, basic_element_per_turn_limit - 2 * EPSILON);
         float topping_up = (additional_pp_to_complete_item < topping_up_limit) ? additional_pp_to_complete_item : basic_element_per_turn_limit;
-        float returnval = std::min(additional_pp_to_complete_element, std::max(basic_element_per_turn_limit, std::max(frontload, topping_up)));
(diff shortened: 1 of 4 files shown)


geoffthemedio:
@MatGB ...

MatGB:
OK, I'm liking this a lot, I've tried a couple abusive tricks and they're gone, plus if you capture a planet with a half completed colony building you get a half completed colony building, it doesn't shrink back down to barely a tenth done because your colony upkeep is higher, which is far better: it might be an idea in the future to limit this in some way but I don't have a balance quibble with it currently and the previous setting was an annoyance.

It needs a display improvement: Having something say "0.50 / 200 PP" on the mouseover isn't clear, it looks like only 0.5 PP has been invested, I think it'd be better as "50% of 200".

I've also gone in to global_settings.txt and set the frontload factor to 0, this basically makes that redundant.

Vezzra:
Correct me if I'm wrong, but if I understand the proposed changes correctly, then while preventing the "speedup completion" cheat, they introduce a cost-reducing cheat: fill your queue with ships and suspend/pause production on all ships just before completion, and add new ships to use up the PP that become available.

This way you can fill up your queue with ships built to almost completion at low build costs, and then start unpausing once you really need the ships. Under the old mechanics you would not be able to save on production costs, because the costs of ships farther down in the queue would increase without regard to percentage completed. But with the new mechanic, the percentage of completion would be taken into account, and you would have to spent the increased production costs only for the remaining X%.

I expect the cost savings gained by careful micromanagament of the build queue to be significant.

So unless I misunderstood how the changes are supposed to work, this fix only shifts how the mechanic can be exploited. IMO the only real fix to the problem is to remove all mechanics that dynamically adjust/change costs entirely. The cost + min production time combo mechanic hasn't been designed with possible dynamic cost change in mind as far as I can tell, which leads to all these issues...

geoffthemedio:
I think it's better to avoid unintended fast production (by accumulating extra PP while at high cost and then completing instantly by reducing the cost), even if it means it's possible to optimize the costs of things by almost-completing several and then finishing them all together.

Also, players could already do something similar by putting a 10x or 20x item on the queue. All the ships in it finish together, so don't increase eachothers' costs.

Also, in order to exploit things as you propose, you have to "store" the PP in incomplete ship production, which means it isn't being spent on anything else. With the spend-faster-while-expensive but finish-faster-after-scrapping exploit, one gets to have all the ships that are making things expensive, then scrap them when no longer needed in order to make the next set of ships produce faster.

geoffthemedio:
Should now show % completion on production queue item tooltips. 

Also rebased.

Vezzra:
Well, I agree that both alternatives have their pros and cons, personally I think being able to save costs is the more problematic issue than being able to save time. But I don't have a strong opinion on this, I can live with either.

My point is that this fix shouldn't be a final solution, as it merely replaces one cheat with another, which is just perceived as a lesser evil. So I hope we can fix the underlying problem on a more fundamental level somewhere down the road.

> Also, players could already do something similar by putting a 10x or 20x item on the queue.

I know, but that feature at least comes at a certain cost, as you have to decide at the time of enqueuing that you want to complete these ships as a batch, you can't back out from that later. The cheat I described is far more flexible (and therefore advantageous) than that.

But even so I see this "batch order" feature combined with our cost increasing mechanic as a small cheat I'm not particularly happy about. It's one of the things I can live with, but still something I hope can be fixed someday.

Not urgent though. Once we get the Influence thing going and maybe the current cost increasing mechanic replaced (or just removed), those issues will be gone anyway. So, as far as I'm concerned, we can merge this PR, I don't really have any serious objection.

geoffthemedio:
For the batch orders, the slightly lower cost of the batch vs. producing separately with some completing early and increasing the cost of the latter is arguably a reasonable compensation for players for giving up the flexibility by using the batch.

Vezzra:
Hm, for large batches (50 or 99) the cost savings might be a bit more substantial than "slightly", but agreed, it comes at a reasonable cost. I still doesn't like it very much, but that's probably a personal preference.

MatGB:
I remain of the opinion that this sort of upkeep mechanism shouldn't be the default system in game and we should use Influence for it, but I *do* like Geoff's idea of different government types allowing for different upkeep systems.

Re the juggling/pausing to get large cheap fleets, I've done that in the past with the existing system, before the Starlane Bore was introduced I'd scrap my entire fleet on conquering the last AI then queue huge numbers in batches of very powerful hulls (Solar, mostly), and move them down the production order until it was nearly Start_Turn, then launch them all in one turn: I did once have all but one batch complete, and that one went from 90%+ complete to about 30% complete as a result.

This system would make it easier to do that, and I hadn't actually thought to try that, but I am frankly happier with that as a potential exploit than I am with the existing system, and this can go in now with no AI rewrite or similar, so until we have Influence and Govt systems and other options the existing mechanic is better.

PR #1182, opened 2016-12-20
Random species selection during setup

### LINKED ISSUE #799: Feature request: random species in multiplayer games
It would be nice to be able to set "random" as the species for a player in a multiplayer game, so people will have to adapt to different playing styles.

dbenage-cx:
Provides option in both Multi-player and Single player setups for random species selection.

Resolves #799

### DIFF
--- a/UI/CUIControls.cpp
+++ b/UI/CUIControls.cpp
@@ -1434,25 +1434,35 @@ namespace {
         {
             if (!species)
                 return;
-            GG::Wnd::SetName(species->Name());
-            GG::StaticGraphic* icon = new GG::StaticGraphic(
-                ClientUI::SpeciesIcon(species->Name()), GG::GRAPHIC_FITGRAPHIC| GG::GRAPHIC_PROPSCALE);
-            icon->Resize(GG::Pt(GG::X(Value(h - 5)), h - 5));
+            const std::string& species_name = species->Name();
+            Init(species_name, UserString(species_name), species->GameplayDescription(), w, h,
+                 ClientUI::SpeciesIcon(species_name));
+        };
+
+        SpeciesRow(const std::string& species_name, const std::string& localized_name, const std::string& species_desc,
+            GG::X w, GG::Y h, boost::shared_ptr<GG::Texture> species_icon) :
+            GG::ListBox::Row(w, h, "", GG::ALIGN_VCENTER, 0)
+        { Init(species_name, localized_name, species_desc, w, h, species_icon); };
+
+    private:
+        void Init(const std::string& species_name, const std::string& localized_name, const std::string& species_desc,
+                  GG::X width, GG::Y height, boost::shared_ptr<GG::Texture> species_icon) {
+            GG::Wnd::SetName(species_name);
+            GG::StaticGraphic* icon = new GG::StaticGraphic(species_icon, GG::GRAPHIC_FITGRAPHIC| GG::GRAPHIC_PROPSCALE);
+            icon->Resize(GG::Pt(GG::X(Value(height - 5)), height - 5));
             push_back(icon);
-            GG::Label* species_name = new CUILabel(UserString(species->Name()), GG::FORMAT_LEFT | GG::FORMAT_VCENTER);
-            push_back(species_name);
-            GG::X first_col_width(Value(h));
+            GG::Label* species_label = new CUILabel(localized_name, GG::FORMAT_LEFT | GG::FORMAT_VCENTER);
+            push_back(species_label);
+            GG::X first_col_width(Value(height));
             SetColWidth(0, first_col_width);
-            SetColWidth(1, w - first_col_width);
+            SetColWidth(1, width - first_col_width);
             GetLayout()->SetColumnStretch(0, 0.0);
             GetLayout()->SetColumnStretch(1, 1.0);
-            SetBrowseModeTime(GetOptionsDB().Get<int>("UI.tooltip-delay"));
-            SetBrowseInfoWnd(boost::shared_ptr<GG::BrowseInfoWnd>(
-                new IconTextBrowseWnd(ClientUI::SpeciesIcon(species->Name()),
-                                      UserString(species->Name()),
-                                      species->GameplayDescription())
-                                     )
-                            );
+            if (!species_desc.empty()) {
+                SetBrowseModeTime(GetOptionsDB().Get<int>("UI.tooltip-delay"));
+                SetBrowseInfoWnd(boost::shared_ptr<GG::BrowseInfoWnd>(new IconTextBrowseWnd(species_icon, localized_name,
+                                                                                            species_desc)));
+            }
         }
     };
 }
@@ -1469,8 +1479,12 @@ SpeciesSelector::SpeciesSelector(GG::X w, GG::Y h) :
     const SpeciesManager& sm = GetSpeciesManager();
     for (SpeciesManager::playable_iterator it = sm.playable_begin(); it != sm.playable_end(); ++it)
         Insert(new SpeciesRow(it->second, w, h - 4));
-    if (!this->Empty())
+    if (!this->Empty()) {
+        // Add an option for random selection
+        Insert(new SpeciesRow("RANDOM", UserString("GSETUP_RANDOM"), UserString("GSETUP_SPECIES_RANDOM_DESC"), w, h - 4,
+                              ClientUI::GetTexture("default/data/art/icons/unknown.png")));
         Select(this->begin());
+    }
     GG::Connect(SelChangedSignal, &SpeciesSelector::SelectionChanged, this);
 }
 
--- a/UI/CUIControls.cpp
+++ b/UI/CUIControls.cpp
@@ -1440,13 +1440,14 @@ namespace {
         };
 
         SpeciesRow(const std::string& species_name, const std::string& localized_name, const std::string& species_desc,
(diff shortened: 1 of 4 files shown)


geoffthemedio:
This seems to work, but I'm wondering whether this randomness should be reproducible for a given seed value... Not sure whether it should be or not.

UI/CUIControls.cpp
```diff
@@ -1434,25 +1434,35 @@ namespace {
...
-            GG::StaticGraphic* icon = new GG::StaticGraphic(
-                ClientUI::SpeciesIcon(species->Name()), GG::GRAPHIC_FITGRAPHIC| GG::GRAPHIC_PROPSCALE);
-            icon->Resize(GG::Pt(GG::X(Value(h - 5)), h - 5));
+            const std::string& species_name = species->Name();
+            Init(species_name, UserString(species_name), species->GameplayDescription(), w, h,
+                 ClientUI::SpeciesIcon(species_name));
+        };
+
+        SpeciesRow(const std::string& species_name, const std::string& localized_name, const std::string& species_desc,
+            GG::X w, GG::Y h, boost::shared_ptr<GG::Texture> species_icon) :
```
geoffthemedio: indent to match parameters on previous row (

UI/CUIControls.cpp
```diff
@@ -1434,25 +1434,35 @@ namespace {
...
+            GG::ListBox::Row(w, h, "", GG::ALIGN_VCENTER, 0)
+        { Init(species_name, localized_name, species_desc, w, h, species_icon); };
+
+    private:
+        void Init(const std::string& species_name, const std::string& localized_name, const std::string& species_desc,
+                  GG::X width, GG::Y height, boost::shared_ptr<GG::Texture> species_icon) {
```
geoffthemedio: { to new line after multi-line declaration

client/human/HumanClientApp.cpp
```diff
@@ -501,12 +501,14 @@ void HumanClientApp::NewSinglePlayerGame(bool quickstart) {
         else
             human_player_setup_data.m_empire_color = GG::CLR_GREEN;
 
+        const SpeciesManager& sm = GetSpeciesManager();
         human_player_setup_data.m_starting_species_name = GetOptionsDB().Get<std::string>("GameSetup.starting-species");
         if (human_player_setup_data.m_starting_species_name == "1")
             human_player_setup_data.m_starting_species_name = "SP_HUMAN";   // kludge / bug workaround for bug with options storage and retreival.  Empty-string options are stored, but read in as "true" boolean, and converted to string equal to "1"
+        else if (human_player_setup_data.m_starting_species_name == "RANDOM")
+            human_player_setup_data.m_starting_species_name = sm.RandomPlayableSpeciesName();
```
geoffthemedio: maybe leave it "RANDOM" here, and instead handle the random choice during galaxy setup based on the seed value...

geoffthemedio (changes requested):

UI/CUIControls.cpp
dbenage-cx: Fixed

client/human/HumanClientApp.cpp
dbenage-cx: Looking into this

dbenage-cx:
Random selection is now tied to seed value.

UI/CUIControls.cpp
```diff
@@ -1430,29 +1430,40 @@ namespace {
     // row type used in the SpeciesSelector
     struct SpeciesRow : public GG::ListBox::Row {
         SpeciesRow(const Species* species, GG::X w, GG::Y h) :
-            GG::ListBox::Row(w, h, "", GG::ALIGN_VCENTER, 0)
+                   GG::ListBox::Row(w, h, "", GG::ALIGN_VCENTER, 0)
```
geoffthemedio: GG::ListBox::Row(...) is the base class constructor, not a parameter to SpeciesRow, so should only be indented 4 spaces, not to the (

UI/CUIControls.cpp
```diff
@@ -1430,29 +1430,40 @@ namespace {
...
-                ClientUI::SpeciesIcon(species->Name()), GG::GRAPHIC_FITGRAPHIC| GG::GRAPHIC_PROPSCALE);
-            icon->Resize(GG::Pt(GG::X(Value(h - 5)), h - 5));
+            const std::string& species_name = species->Name();
+            Init(species_name, UserString(species_name), species->GameplayDescription(), w, h,
+                 ClientUI::SpeciesIcon(species_name));
+        };
+
+        SpeciesRow(const std::string& species_name, const std::string& localized_name, const std::string& species_desc,
+                   GG::X w, GG::Y h, boost::shared_ptr<GG::Texture> species_icon) :
+                   GG::ListBox::Row(w, h, "", GG::ALIGN_VCENTER, 0)
```
geoffthemedio: same here

server/ServerApp.cpp
```diff
@@ -965,6 +966,19 @@ namespace {
...
+
+    /** Replace species selection for any players set to "RANDOM" species, with a species based on @p galaxy_seed */
+    void ValidateSpeciesSelections(std::map<int, PlayerSetupData>& player_setup_data, const std::string& galaxy_seed) {
+        SpeciesManager& sm = GetSpeciesManager();
+        int num_playable_species = sm.NumPlayableSpecies();
+        std::string species_seed = galaxy_seed + "species";
+        for (std::map<int, PlayerSetupData>::value_type& entry : player_setup_data) {
+            if (entry.second.m_starting_species_name == "RANDOM") {
+                entry.second.m_starting_species_name = sm.SequentialPlayableSpeciesName(GetIdxForSeed(num_playable_species,
+                                                                                                         species_seed) + 1);
```
geoffthemedio: weird indenting

server/ServerApp.cpp
```diff
@@ -965,6 +966,19 @@ namespace {
...
     }
+
+    /** Replace species selection for any players set to "RANDOM" species, with a species based on @p galaxy_seed */
+    void ValidateSpeciesSelections(std::map<int, PlayerSetupData>& player_setup_data, const std::string& galaxy_seed) {
+        SpeciesManager& sm = GetSpeciesManager();
+        int num_playable_species = sm.NumPlayableSpecies();
+        std::string species_seed = galaxy_seed + "species";
+        for (std::map<int, PlayerSetupData>::value_type& entry : player_setup_data) {
+            if (entry.second.m_starting_species_name == "RANDOM") {
+                entry.second.m_starting_species_name = sm.SequentialPlayableSpeciesName(GetIdxForSeed(num_playable_species,
```
geoffthemedio: what happens if multiple players pick "RANDOM"? do they get the same species?

using this method is OK for galaxy setup settings, but for multiple-evaluations, need to use stateful random functions, perhaps within Python code...

server/ServerApp.cpp
```diff
@@ -965,6 +966,19 @@ namespace {
                                    << "human player with assigned empire id: " << psd.m_save_game_empire_id;
         }
     }
+
+    /** Replace species selection for any players set to "RANDOM" species, with a species based on @p galaxy_seed */
+    void ValidateSpeciesSelections(std::map<int, PlayerSetupData>& player_setup_data, const std::string& galaxy_seed) {
```
geoffthemedio: it's not really "validating" anything, is it?

geoffthemedio (changes requested):

server/ServerApp.cpp
dbenage-cx: TY, I've simply appended the empire name for now.  That is not meant to ignore your suggestion, I need to inspect the existing Python code to see what is relevant tomorrow.

dbenage-cx:
This is not portable between boost 1.60 and 1.62, will keep digging.

server/ServerApp.cpp
```diff
@@ -965,6 +966,52 @@ namespace {
...
+                int player_id = entry.first;
+                // alter the seed for each player according to their id
+                seed_altered = true;
+                Seed(initial_seed + player_id);
+                unsigned int seed = initial_seed * RandGaussian(1.0, (player_id * 0.01));
+                Seed(seed);
+                int idx = RandWeightedInt(species_prob);
+
+                // reduce the chance of this species being picked again
+                species_prob[idx] -= weight_reduction;
```
Morlic-fo: You probably do not want to do this. 

Let n be the number of species and m the number of times a species is picked.

Then, your chance to pick the species m+1 times is ```p(m+1) = p(m) - 1/n^2 = 1/n - m/n^2```. This obviously runs into trouble if m > n.  This **will** happen if the total number of random selection exceeds n^2. (Consider only 1 playable species but 3 players that want a random selection).
Not sure if the boost function you call has some fancy stuff happening to handle weights that sum to zero or even negative probabilites but you leave mathematical definitions behind in any case.

Instead, why not reduce probability by a set factor each time the species is selects, e.g. ```species_prob[idx] *= .5```, i.e. ```p(m+1) = p(m) * 0.5 = p(initial) * 0.5^m```.
where m is number of times this species is picked. Seems more intuitive and does also not run into the potential problems above.

dbenage-cx: Thank you for this.  Even for relative weights, it would be better not to tempt an issue with negative values.  I will keep this in mind if I implement anything similar in the future.

dbenage-cx:
Apparently, I'm trying to recreate the wheel yet again.
Dropped previous to utilize existing python selection.

geoffthemedio:
seems to work

PR #1184, opened 2016-12-21
Ai tweaks for sparse galaxy

Dilvish-fo:
Hi guys, long time no see.  I have a little Christmas contribution that I think could help liven up the holidays-- some relatively modest AI tweaks that nonetheless seem to make it a fair bit more competitive, especially in the specific setups this is directed to, which are ones where the monster density is no greater than Low, and where there are at least 40 systems per empire (or 35 if the shape is elliptical).   I prefer to run this at 400 systems and 6 AIs.  

Also, independent of the sparsity, this enables the AI to use the XenoResurrection Lab and the associated extinct species.

I am still getting up to speed on some of the AI changes during my absence, so CJ et al., just let me know if you'd prefer I change the way I tweaked the techs here.

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -917,10 +917,8 @@ def evaluate_planet(planet_id, mission_type, spec_name, empire, detail=None):
             ast_val = 0
             if tech_is_complete("PRO_MICROGRAV_MAN"):
                 per_ast = 5
-            elif fo.currentTurn() > 40:
-                per_ast = 2.5
             else:
-                per_ast = 0.1
+                per_ast = 3
             if system:
                 for pid in system.planetIDs:
                     other_planet = universe.getPlanet(pid)
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -214,6 +214,9 @@ def survey_universe():
         active_growth_specials.clear()
         if tech_is_complete(TechsListsAI.EXOBOT_TECH_NAME):
             empire_colonizers["SP_EXOBOT"] = []  # get it into colonizer list even if no colony yet
+        for spec_name in AIDependencies.EXTINCT_SPECIES:
+            if empire.buildingTypeAvailable("BLD_COL_"+spec_name):
+                empire_colonizers["SP_"+spec_name] = []  # get it into colonizer list even if no colony yet
         AIstate.popCtrIDs[:] = []
         AIstate.popCtrSystemIDs[:] = []
         AIstate.outpostIDs[:] = []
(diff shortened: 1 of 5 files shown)


Cjkjvfnby:
I have exams soon, after that new year + vacation and more exams after.  I am useless until that.  :(

default/python/AI/ColonisationAI.py
```diff
@@ -214,6 +214,9 @@ def survey_universe():
         active_growth_specials.clear()
         if tech_is_complete(TechsListsAI.EXOBOT_TECH_NAME):
             empire_colonizers["SP_EXOBOT"] = []  # get it into colonizer list even if no colony yet
+        for spec_name in AIDependencies.EXTINCT_SPECIES:
+            if empire.buildingTypeAvailable("BLD_COL_"+spec_name):
+                empire_colonizers["SP_"+spec_name] = []  # get it into colonizer list even if no colony yet
```
Morlic-fo, line 219: Whitespace around operators.

Morlic-fo (changes requested):
What a nice christmas surprise! Happy to see you again!

MatGB:
I've played two games through with this now and, for the first time in over a year, one of them had to be restarted as the AI overpowered me in the early game: partially due to my own mistakes and partially because it was doing things I wasn't expecting.

I especially like the greater emphasis on detection techs in this, even if it did mess up some of my strategies, but it also made me try out a few new ideas in a way that was fun.

Overall I like the way it feels/plays.

Morlic-fo:
Ok that did not work out as intended. Tried resolving the conflicts using the web interface. Turns out that creates a merge commit.

Vezzra:
> Ok that did not work out as intended. Tried resolving the conflicts using the web interface. Turns out that creates a merge commit.

@Morlic-fo, hopefully fixed. 😃

Dilvish-fo:
Morlic, thanks much for your feedback.  I've made the corrections and fixed the conflict with current master, and was trying to do a bit of testing to confirm things are working OK before pushing the revised branch.   My testing is a little bit slowed down, though, since I ~~can't find any AI logs anymore,...~~ had trouble finding the new location for AI logs (in the .local folder), but I just now found it.

dbenage-cx:
@Dilvish-fo try `~/.local/share/freeorion` for logs (if not, check config file for location of `save-dir`, should be the parent of)

Vezzra:
@Dilvish-fo, do you think you can get this finalized in time for the release?

Cjkjvfnby:
@Dilvish-fo, closing PR by pushing other commit does not looks good for me.   IMHO PR should be closed manually with comment.

PR #1185, opened 2016-12-25
Fix Head on a spike special

Morlic-fo:
Fixes an issue described in this [forum post](http://freeorion.org/forum/viewtopic.php?f=28&t=10354&p=86863#p86863). 

Ever since the weapon upgrade rework, the special would now reduce the *current* weapon capacity by 1 **per turn** whereas the initial intention (IIRC) was to reduce the maximum damage of the weapons by 1.


MatGB:
And I would like to take the opportunity to thank Github for only displaying the changes between the two commits on the code page and not the changes between this PR and master, was going to ask what the difference was before looking at individual commits.

I had not noticed this happening at all, it's a very good catch, have you managed to replicate the problem and test this as a solution yourself?

PR #1188, opened 2016-12-28
Optional separate descriptive text for map menu icon tooltips

dbenage-cx:
Suggested by MatGB in [forum](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10348&start=15).

Sets a separate key for descriptive text on the map menu icon tooltips.
These keys are initially set to an empty value, relying on the behavior of TextBrowseWnd in not rendering the extra space for an empty string (but preserving the ability if desired later).

PR opened in event that option is not desired and these should be hard-coded as empty strings.


geoffthemedio:
looks fine...

PR #1190, opened 2016-12-31
Fix include path for VS build

Voker57:


server/ServerApp.cpp
```diff
@@ -41,6 +41,7 @@
 
 
 #include <ctime>
+#include <../util/Random.h>
```
geoffthemedio: should be with "" and with the other ../util headers

geoffthemedio (changes requested):

geoffthemedio (approved):

geoffthemedio:
What error is this supposed to fix? I don't have any compile issues.

Vezzra:
Tagged this as bugfix for now, because it says "fix" in the title, but I'm not sure either what exactly this PR is supposed to do. So please correct me if my assumptions are wrong.

Voker57:
@geoffthemedio when building with VS 2015, compiler complained about that include not found. The fix resolved the issue.

geoffthemedio:
I'm not getting any such error with MSVC 2015 (or 2013)

Voker57:
Well no idea but it occured to me so I submitted a fix (and it's proper way to include, right?). Not going to rebuild fixless code in my slow VM to recheck.

Vezzra:
If the change doesn't cause any problems for anyone else (which is highly unlikely considering it's just an added include), and apparently fixes a problem @Voker57 had, I guess we can merge this? It's a *very* simple change after all...

adrianbroher:
@Vezzra 

No, as adding unnecessary headers always increase compile time.  If this problem is specific to Vokers fork it is his problem.

Voker57:
Oh, my apologies. It's indeed related to non-included patch and supposed to correct include line, not add it. I didn't check it after cherry-picking.

Vezzra:
@adrianbroher, sorry, I thought @Voker57 had the problem when building master, didn't think of the possibility that this might be an issue with his branch.

Issue #758, opened 2016-07-07
Tech List


geoffthemedio:
Make the tech list view resizable and sortable.

dbenage-cx:
Sort and config defined column widths were added.
Still needs UI support for resizing columns.

PR #1200, opened 2017-01-04
Sort tech list view columns

### LINKED ISSUE #958: Tech list display glitch in other languages than english
### Environment

- **FreeOrion Version**: v0.4.6 pre-release [build 2016-09-04.22b124c]
- **Operating System**: Windows 7
### Description

Due to longer words, the text in the tech list display columns may overlap in languages other than english:

English:
tech_list_en

French:
tech_list_fr

Deutsch:
tech_list_de
### Expected Result

Make the tech list columns resizeable.

dbenage-cx:
Adds sorting of tech tree list view columns and config defined column widths.

As this PR does not allow in-game adjustment of columns, only partial work towards #758
fixes #958


UI/TechTreeWnd.cpp
```diff
@@ -41,6 +44,19 @@ namespace {
         db.Add("UI.tech-layout-zoom-scale",     UserStringNop("OPTIONS_DB_UI_TECH_LAYOUT_ZOOM_SCALE"),   1.0,  RangedStepValidator<double>(1.0, -25.0, 10.0));
         db.Add("UI.tech-controls-graphic-size", UserStringNop("OPTIONS_DB_UI_TECH_CTRL_ICON_SIZE"),      3.0,  RangedStepValidator<double>(0.25, 0.5,  12.0));
         db.Add("UI.windows." + RES_PEDIA_WND_NAME + ".persistently-hidden", UserStringNop("OPTIONS_DB_RESEARCH_PEDIA_HIDDEN"), false, Validator<bool>());
+
+        // TechListBox::TechRow column widths
+        int default_pts = 16;
+        db.Add(TECHROW_COL_XML_PREFIX + ".graphic",         UserStringNop(TECHROW_COL_ST_PREFIX + "_GRAPHIC"),
```
geoffthemedio: not sure, but I don't think @adrianbroher would like this... Probably better to put the whole stringtable key in explicitly each time.

UI/TechTreeWnd.cpp
```diff
@@ -1583,20 +1604,108 @@ void TechTreeWnd::TechListBox::TechRow::Render() {
...
-    retval.push_back(DESC_WIDTH);
+    // Early return for very small screen
+    if (total_width < (6 * 4 * ClientUI::Pts())) {
+        ErrorLogger() << "Not enough room for reasonable column space";
+        retval = {GG::X1, GG::X1, GG::X1, GG::X1, GG::X1, GG::X1};
+        return retval;
+    }
+
+    // Attempt to utilize user defined widths
+    // If not enough width to fit, attempt default widths
```
geoffthemedio: If there are user defined column widths, I'd just use them, and have a scrollbar to show extra width if needed.

UI/TechTreeWnd.cpp
```diff
@@ -1583,20 +1604,108 @@ void TechTreeWnd::TechListBox::TechRow::Render() {
...
+    GG::X time_width(       GetOptionsDB().Get<int>(TECHROW_COL_XML_PREFIX + ".time"));
+    GG::X category_width(   GetOptionsDB().Get<int>(TECHROW_COL_XML_PREFIX + ".category"));
+
+    GG::X cols_width_sum = graphic_width + name_width + cost_width + time_width + category_width;
+
+    GG::X desc_width(std::max(GG::X1, total_width - cols_width_sum));
+
+    if (cols_width_sum > total_width) {
+        // config values did not fit, attempt default values
+        graphic_width   = GG::X(GetOptionsDB().GetDefault<int>(TECHROW_COL_XML_PREFIX + ".graphic"));
```
geoffthemedio: all this prefix stuff isn't really saving enough space to warrant hiding the actual value amongst several variables / constants, I think...

UI/TechTreeWnd.cpp
```diff
@@ -1583,20 +1604,108 @@ void TechTreeWnd::TechListBox::TechRow::Render() {
...
+
+bool TechTreeWnd::TechListBox::TechRowCmp(const GG::ListBox::Row& lhs, const GG::ListBox::Row& rhs, std::size_t column) {
+    bool retval = false;
+    const std::string lhs_key = boost::trim_copy(lhs.SortKey(column));
+    const std::string rhs_key = boost::trim_copy(rhs.SortKey(column));
+
+    // When equal, sort by previous sorted column
+    if ((lhs_key == rhs_key) && (m_previous_sort_col != column)) {
+        retval = TechRowCmp(lhs, rhs, m_previous_sort_col);
+    } else if (column == 2 || column == 3) {
```
geoffthemedio: not a huge fan of special casing like thing... but probably OK for first imlementation

UI/TechTreeWnd.cpp
```diff
@@ -1583,20 +1604,108 @@ void TechTreeWnd::TechListBox::TechRow::Render() {
...
+        try {  // compare by int
+            retval = boost::lexical_cast<int>(lhs_key) < boost::lexical_cast<int>(rhs_key);
+        } catch (boost::bad_lexical_cast& e) {
```
geoffthemedio: should the exception variable be const?

UI/TechTreeWnd.cpp
```diff
@@ -1583,20 +1604,108 @@ void TechTreeWnd::TechListBox::TechRow::Render() {
...
+            DebugLogger() << "Unable to cast sort keys to int, lhs:" << lhs_key << " rhs:" << rhs_key << ". " << e.what();
+            retval = lhs_key < rhs_key;
+        }
+    } else {
+        retval = lhs_key < rhs_key;
```
geoffthemedio: what does < do for strings? How does it handle non-latin character sets? probably need to replicate what the objects list is doing for these cases...

geoffthemedio (changes requested):

UI/TechTreeWnd.cpp
dbenage-cx: Fixed

UI/TechTreeWnd.cpp
dbenage-cx: Note, column headers do not scroll horizontally with listbox rows.
I've changed per request as I suspect this will not be a major issue until in-game adjustments are supported.

UI/TechTreeWnd.cpp
dbenage-cx: Fixed (was following previous example mostly to avoid typos)

UI/TechTreeWnd.cpp
dbenage-cx: Not addressed.
Some sort of flag to indicate the underlying value type?
Is relying on a throw for intended behavior a sensible practice? (honest question, not meant to offend)

UI/TechTreeWnd.cpp
dbenage-cx: Fixed

UI/TechTreeWnd.cpp
dbenage-cx: I've changed to a duplicated GetLocal(), I'm not quite sure how this is a better option when the locale is manually set to `en_US` however.
We could set these to `""` for the default system locale, until a proper link with the selected language is implemented.

geoffthemedio: regardless of what language / locale is selected, as the comment in ObjectListWnd.cpp notes, setting up a locale will sort accented letters properly, which not using one won't.

UI/TechTreeWnd.cpp
```diff
@@ -1583,20 +1611,61 @@ void TechTreeWnd::TechListBox::TechRow::Render() {
...
+    const std::string rhs_key = boost::trim_copy(rhs.SortKey(column));
+
+    // When equal, sort by previous sorted column
+    if ((lhs_key == rhs_key) && (m_previous_sort_col != column)) {
+        retval = TechRowCmp(lhs, rhs, m_previous_sort_col);
+    } else if (column == 2 || column == 3) {
+        try {  // compare by int
+            retval = boost::lexical_cast<int>(lhs_key) < boost::lexical_cast<int>(rhs_key);
+        } catch (const boost::bad_lexical_cast& e) {
+            DebugLogger() << "Unable to cast sort keys to int, lhs:" << lhs_key << " rhs:" << rhs_key << ". " << e.what();
```
geoffthemedio: don't need to log this... is expected to occur often.

UI/TechTreeWnd.cpp
geoffthemedio: If you're catching an exception locally, sure... same method is used in various boost graph traversals to indicate the algorithm finished / couldn't finish. What is done in MapWnd/ObjectListWnd seems to work: attempt to cast to number to then compare, if that fails, then compare as a string.

Why is checking for columns 2 or 3 necessary?

dbenage-cx: Removed checks for specific columns.

The check was to avoid a casting attempt when those values are known not to represent an int (e.g. two techs with nomenclatures for names).  Such an event in this case only changes the order between those rows, so only a minor issue if it ever appears.

UI/TechTreeWnd.cpp
dbenage-cx: Ok, TY for explanation

UI/TechTreeWnd.cpp
dbenage-cx: Fixed

geoffthemedio:
Sorting seems to work (though didn't carefully test). The header row / buttons resize a bit oddly when the window size increases, though...
buttons

geoffthemedio (approved):

geoffthemedio:
The horizontal scrollbar behaves a bit oddly... if the window is wide enough to show all the rows, there is no bar, but just a few pixels smaller and the bar can appear, and then it can be scrolled all the way over so the last row is the first row shown. Not really a problem, but it seems odd when using the GUI.

PR #1202, opened 2017-01-05
Some fixes for CMake-generated MSVC project.

o01eg:
1. Use unicode version of WinAPI functions. I got error with STARTUPINFOA.
2. Add dependency from libpng to human client project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,10 @@ if (WIN32)
         set (CMAKE_STATIC_LIBRARY_PREFIX "lib")
         set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:LIBCMT")
         set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /NODEFAULTLIB:LIBCMT")
+        add_definitions(
+            -DUNICODE
+            -D_UNICODE
+        )
     endif ()
 elseif (APPLE)
     add_definitions(-DFREEORION_MACOSX)
--- a/parse/DoubleComplexValueRefParser.cpp
+++ b/parse/DoubleComplexValueRefParser.cpp
@@ -48,7 +48,7 @@ namespace parse {
 
             species_empire_opinion
                 = (
-                    (  tok.SpeciesOpinion_ [ _a = construct<std::string>("SpeciesEmpireOpinion") ]
+                    (  tok.SpeciesOpinion_ [ _a = construct<std::string>(std::string("SpeciesEmpireOpinion")) ]
                        >  parse::label(Species_token) >  string_value_ref [ _d = _1 ]
                     )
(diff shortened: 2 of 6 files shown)


CMakeLists.txt
```diff
@@ -330,6 +334,7 @@ if (NOT WIN32)
 endif ()
 
 add_library(freeorioncommon
+    $<$<CXX_COMPILER_ID:msvc>:STATIC>
```
geoffthemedio: it already is a static library for me...?

o01eg: For MSVC 2017 it tries to build dll.

Vezzra: > it already is a static library for me...?

Because you are using the MSVC project provided in the repo. @o01eg apparently tried to produce a MSVC project using the cmake scripts, and there the common library apparently is specified as dynamic.

geoffthemedio: The repo project files were based on (possibly older) CMake-generated project files.

Vezzra: Hm, in that case I suspect @adrianbroher changed that to be build as a dynamic lib. It's his goal to turn the common and parser libs into shared libraries after all. So, I wouldn't just change it back to static, at least not without feedback from him - @adrianbroher?

o01eg: I think it possible to change to shared libraries but now FO_COMMON_API isn't defined for windows.

o01eg:
I've deleted last commit. I want to check if it possible to fix shared library compilation.

CMakeLists.txt
adrianbroher: > The repo project files were based on (possibly older) CMake-generated project files.

That's not true.  The repo history shows that.  Also the generated projects are unfit for source tracking.

@o01eg Don't waste your time.  Making the libaries shared isn't as easy on Windows as it is on the other two platforms as there are circular symbol dependencies that need to be sorted out before whereas the other platforms have a linker flag to let the runtime linker resolve those.

parse/DoubleComplexValueRefParser.cpp
```diff
@@ -48,7 +48,7 @@ namespace parse {
 
             species_empire_opinion
                 = (
-                    (  tok.SpeciesOpinion_ [ _a = construct<std::string>("SpeciesEmpireOpinion") ]
```
adrianbroher, line 51: Not related to CMake.  Please drop this from the PR.

Aside from that is a strange way of fixing this. I wouldn't pull it this way.

adrianbroher (changes requested):

o01eg:
I succesfully built FreeOrion for Win64 so I'll close this PR.

parse/DoubleComplexValueRefParser.cpp
dbenage-cx: @adrianbroher Since this affects building with boost 1.63, could you expand on a preferred resolution? (or `-DBOOST_PHOENIX_NO_VARIADIC_EXPRESSION`)

PR #1205, opened 2017-01-06
Massively increased build costs of Transformer

Vezzra:
Fixes the long standing and repeatedly criticized issue that the build costs of the Transformer building are far too low, sometimes only a fraction of one of the buildings it can emulate (stargates being the example being frequently brought up). Raised the build costs from 100 to 800.

### DIFF
--- a/default/scripting/buildings/TRANSFORMER.focs.txt
+++ b/default/scripting/buildings/TRANSFORMER.focs.txt
@@ -1,7 +1,7 @@
 BuildingType
     name = "BLD_TRANSFORMER"
     description = "BLD_TRANSFORMER_DESC"
-    buildcost = 100
+    buildcost = 600
     buildtime = 8
     location = AND [
         Not Contains Building name = "BLD_TRANSFORMER"


MatGB:
I'm uncomfortable making it that expensive: part of the reason it exists (the only reason it exists) is to reduce micromanagement, one building scattered around not 4 (one of which is never used). Two of the buildings only cost 75, the other two each cost 500, one of which is useful in a lot of places the other is only really good at very key defensive points.

800 is too high *compared to what it's replacing*, I'd say 500 is better, 600 at most. On the other hand putting the costs of the other buildings up would also be viable as is it goes from useful replacement to a waste of PP in virtually all cases.

(aside: we really need to fix Bio Terror in some way at some point, I have zero clue how. That or remove it)

Vezzra:
> part of the reason it exists (the only reason it exists) is to reduce micromanagement, one building scattered around not 4 (one of which is never used)

Having a building (or any particular game element/mechanic) solely for the purpose of reducing micromanagement is something we should avoid IMO.

Therefore I think the Transformer should only be viable in cases where you want *at least* two of the buildings it replaces on a planet. If I only want e.g. a Stargate, building a Stargate and not a Transformer should be the better option.

Which means the Transformer need to be priced accordingly, which in turn means it must be *at least* a bit more expensive than the most expensive building it can replace. On the other hand I have too agree, you have a point, if two of the buildings the Transformer replaces only cost 75, and the other two 500, 800 is a bit too much.

But I would not go lower than 600, that's already very close to the 500 of the two more expensive buildings. As that's the upper limit you think could be viable, I guess that's an acceptable compromise, so I put that change in.

If you're ok with that, and no one else has any objections, I think we can merge.

However, we should take a look at the two 75PP buildings and maybe also reconsider their cost. Getting the build costs right for a building that replaces other buildings which such vast differences in price is very difficult...

Vezzra:
> (aside: we really need to fix Bio Terror in some way at some point, I have zero clue how. That or remove it)

Oh, I *do* have some ideas about "biological warfare", I think there can be done some very interesting (and *nasty*) things. But that's more of a complete mechnic I have in mind, I don't think the Bioterror buildings in their current form or anything similar to it can be made to work reasonably.

MatGB:
Agree on all counts basically, 600 is fine and looking at costs of all buildings is one of those longer term balancing objectives that will happen eventually anyway.

PR #1206, opened 2017-01-06
Remove build cost increase mechanic from stargate

Vezzra:


Vezzra:
As the build cost increase mechanic doesn't work well with the new fractional progress mechanic of the build queue, and by @MatGB's suggestion [on the forum](http://freeorion.org/forum/viewtopic.php?p=86961#p86961) this PR just removes the build cost increase part.

PR #1209, opened 2017-01-08
FOCS support for part based fleet upkeep

dbenage-cx:
~~PR is intended for testing of fleet upkeep utilizing number of currently owned parts.
Initial change from 1% per ship to 0.2% per part.
While there are various threads discussing changes to upkeep, the most recent revolves around [Influence & Happiness](http://www.freeorion.org/forum/viewtopic.php?f=5&t=9998)~~

Changed to only add support for such a change.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -3385,6 +3385,23 @@ void Empire::UpdateOwnedObjectCounters() {
         m_ship_designs_owned[ship->DesignID()]++;
     }
 
+    // update ship part counts
+    m_ship_part_types_owned.clear();
+    m_ship_part_class_owned.clear();
+    for (const std::map<int, int>::value_type& design_count : m_ship_designs_owned) {
+        const ShipDesign* design = GetShipDesign(design_count.first);
+        if (!design)
+            continue;
+
+        // update count of PartTypes
+        for (const std::map<std::string, int>::value_type& part_type : design->PartTypeCount())
+            m_ship_part_types_owned[part_type.first] += part_type.second * design_count.second;
+
+        // update count of ShipPartClasses
+        for (const std::map<ShipPartClass, int>::value_type& part_class : design->PartClassCount())
+            m_ship_part_class_owned[part_class.first] += part_class.second * design_count.second;
+    }
+
     // colonies of each species, and unspecified outposts
     m_species_colonies_owned.clear();
     m_outposts_owned = 0;
@@ -3424,6 +3441,16 @@ int Empire::TotalShipsOwned() const {
     for (const std::map<int, int>::value_type& entry : m_ship_designs_owned)
     { counter += entry.second; }
     return counter;
+}
+
+int Empire::TotalShipPartsOwned() const {
+    // sum counts of all ship parts owned by this empire
+    int retval = 0;
+
+    for (const std::map<ShipPartClass, int>::value_type& part_class : m_ship_part_class_owned)
+        retval += part_class.second;
+
+    return retval;
 }
 
 int Empire::TotalBuildingsOwned() const {
--- a/Empire/Empire.h
+++ b/Empire/Empire.h
@@ -567,10 +567,13 @@ public:
     void                        UpdateOwnedObjectCounters();
 
     int                         TotalShipsOwned() const;
+    int                         TotalShipPartsOwned() const;    ///< Total number of parts for all owned ships in this empire
     int                         TotalBuildingsOwned() const;
 
     std::map<std::string, int>& SpeciesShipsOwned()     { return m_species_ships_owned; }
     std::map<int, int>&         ShipDesignsOwned()      { return m_ship_designs_owned; }
+    std::map<std::string, int>&       ShipPartTypesOwned()        { return m_ship_part_types_owned; }
+    std::map<ShipPartClass, int>&     ShipPartClassOwned()        { return m_ship_part_class_owned; }
     std::map<std::string, int>& SpeciesColoniesOwned()  { return m_species_colonies_owned; }
     int&                        OutpostsOwned()         { return m_outposts_owned; }
     std::map<std::string, int>& BuildingTypesOwned()    { return m_building_types_owned; }
@@ -643,6 +646,8 @@ private:
 
     std::map<std::string, int>      m_species_ships_owned;      ///< how many ships of each species does this empire currently own?
     std::map<int, int>              m_ship_designs_owned;       ///< how many ships of each design does this empire currently own?
+    std::map<std::string, int>      m_ship_part_types_owned;    ///< how many ship parts are currently owned, indexed by PartType
+    std::map<ShipPartClass, int>    m_ship_part_class_owned;    ///< how many ship parts are currently owned, indexed by ShipPartClass
     std::map<std::string, int>      m_species_colonies_owned;   ///< how many colonies of each species does this empire currently own?
     int                             m_outposts_owned;           ///< how many uncolonized outposts does this empire currently own?
     std::map<std::string, int>      m_building_types_owned;     ///< how many buildings does this empire currently own?
--- a/parse/IntComplexValueRefParser.cpp
+++ b/parse/IntComplexValueRefParser.cpp
@@ -102,6 +102,18 @@ namespace parse {
                     ) [ _val = new_<ValueRef::ComplexVariable<int> >(_a, _b, _c, _f, _d, _e) ]
                 ;
 
+            ship_parts_owned
+                =   (
+                            tok.ShipPartsOwned_ [ _a = construct<std::string>(_1) ]
+                        >-( parse::label(Empire_token)          > int_value_ref [ _b = _1 ] )
+                        >-(     ( parse::label(Name_token)      > string_value_ref [ _d = _1 ] )
+                            |   ( parse::label(Class_token)     >>
+                                  parse::enum_parser<ShipPartClass>() [ _c = new_<ValueRef::Constant<int>>(_1) ]
+                                )
+                          )
+                    ) [ _val = new_<ValueRef::ComplexVariable<int>>(_a, _b, _c, _f, _d, _e) ]
+                ;
+
             ship_designs_destroyed
                 =   (
                             tok.ShipDesignsDestroyed_ [ _a = construct<std::string>(_1) ]
@@ -238,6 +250,7 @@ namespace parse {
                 |   outposts_owned
                 |   parts_in_ship_design
                 |   part_class_in_ship_design
+                |   ship_parts_owned
                 |   ship_designs_destroyed
                 |   ship_designs_lost
                 |   ship_designs_owned
@@ -265,6 +278,7 @@ namespace parse {
             outposts_owned.name("OutpostsOwned");
             parts_in_ship_design.name("PartsInShipDesign");
             part_class_in_ship_design.name("PartClassInShipDesign");
+            ship_parts_owned.name("ShipPartsOwned");
             ship_designs_destroyed.name("ShipDesignsDestroyed");
             ship_designs_lost.name("ShipDesignsLost");
             ship_designs_owned.name("ShipDesignsOwned");
@@ -292,6 +306,7 @@ namespace parse {
             debug(outposts_owned);
             debug(parts_in_ship_design);
             debug(part_class_in_ship_design);
+            debug(ship_parts_owned);
             debug(ship_designs_destroyed);
             debug(ship_designs_lost);
             debug(ship_designs_owned);
@@ -321,6 +336,7 @@ namespace parse {
         complex_variable_rule<int>::type    outposts_owned;
         complex_variable_rule<int>::type    parts_in_ship_design;
         complex_variable_rule<int>::type    part_class_in_ship_design;
+        complex_variable_rule<int>::type    ship_parts_owned;
         complex_variable_rule<int>::type    ship_designs_destroyed;
         complex_variable_rule<int>::type    ship_designs_lost;
         complex_variable_rule<int>::type    ship_designs_owned;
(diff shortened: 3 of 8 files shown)


UI/MapWnd.cpp
```diff
@@ -5779,16 +5779,36 @@ void MapWnd::RefreshFleetResourceIndicator() {
     const std::set<int>& this_client_known_destroyed_objects = GetUniverse().EmpireKnownDestroyedObjectIDs(empire_id);
 
     int total_fleet_count = 0;
```
geoffthemedio, line 5782: if adding other counts of ships (not actually counting fleets), this should be renamed to total_ship_count

UI/MapWnd.cpp
```diff
@@ -5779,16 +5779,36 @@ void MapWnd::RefreshFleetResourceIndicator() {
...
+            total_ship_slot_count += hull_type->NumSlots();
+        }
     }
 
     m_fleet->SetValue(total_fleet_count);
     m_fleet->ClearBrowseInfoWnd();
     m_fleet->SetBrowseModeTime(GetOptionsDB().Get<int>("UI.tooltip-delay"));
     m_fleet->SetBrowseInfoWnd(boost::shared_ptr<GG::BrowseInfoWnd>(
-        new TextBrowseWnd(UserString("MAP_FLEET_TITLE"), UserString("MAP_FLEET_TEXT"))));
+        new TextBrowseWnd(UserString("MAP_FLEET_TITLE"),
```
geoffthemedio: Can you make this a separate pull request?

parse/IntComplexValueRefParser.cpp
```diff
@@ -102,6 +102,13 @@ namespace parse {
                     ) [ _val = new_<ValueRef::ComplexVariable<int> >(_a, _b, _c, _f, _d, _e) ]
                 ;
 
+            ship_parts_owned
```
geoffthemedio, line 105: not necessary now, but planning to add a parameter to specify a part type would be good

Empire/Empire.cpp
```diff
@@ -3429,6 +3429,17 @@ int Empire::TotalShipsOwned() const {
     return counter;
 }
 
+int Empire::TotalShipPartsOwned() const {
```
geoffthemedio, line 3446: You can't recalculate this every time the value is needed. It will need to be store in the empire as state info, and updated once per turn in UpdateOwnedObjectCounters

UI/MapWnd.cpp
dbenage-cx: Done. PR #1210 (will drop for this PR when updated with other changes requested)

UI/MapWnd.cpp
dbenage-cx: Fixed in other PR

Vezzra:
> number of currently owned parts

IIRC @geoffthemedio had already added FOCS keywords that returned the total number of slots and number of used slots on a ship design, what are "owned parts" supposed to be?

Vezzra:
@freeorion/ai-team, what adjustments to the AI are required for this?

MatGB:
The AI gets the upkeep from here [freeorion/AIDependencies.py at master · freeorion/freeorion](https://github.com/freeorion/freeorion/blob/master/default/python/AI/AIDependencies.py#L54) but it's backend coded to think that's a per ship cost. I've started my first game with this now and given it's very close to what I wanted to try out (but could never get working, fairly sure Geoff's FOCS was PartsInShipDesign but it was so long ago I last tried to play around with that) unsurprisingly I like it.

If we do have an upkeep-as-govt-choice mechanic something like this is far preferable to the current blunt edged setup and this appears to work fine on initial pass. Only on turn 70 with 14 ships total though.

parse/IntComplexValueRefParser.cpp
dbenage-cx: Added

Empire/Empire.cpp
dbenage-cx: Fixed

dbenage-cx:
Added ShipPartClassOwned to support different upkeep per part class.
e.g. For upkeep to increase only for short range weapon parts
`(1 + 0.004 * ShipPartClassOwned empire = Source.Owner class = ShortRange)`

No work was done towards AI yet, would like to finalize a formula for upkeep beforehand.

> what are "owned parts" supposed to be?

Current number of ship parts for all existing ships, possibly limited to those owned by a specific empire and/or of a specific part type.  In the use case here, it is for the total number of currently owned parts for the source empire.

MatGB:
More feedback: I've thought for a long time but this now confirms it that we need something in the UI displaying the current Upkeep rates, it's confusing enough for new players we should be trying to make it more discoverable, and my brain keeps seizing up when I ask it to work out the fleet upkeep with this system. 

Having it in the Fleet Metrics mouseover is a bare minimum, also having it in the production screen somewhere, perhaps in the box where it gives your total production and prod in this resourcegroup?

Same for colony upkeep, in the same place in production if that's possible, then a colony count per species and a total colony at the bottom of the census pane? (that's outside the remit of this PR, obviously).

Empire/Empire.cpp
```diff
@@ -3386,6 +3388,20 @@ void Empire::UpdateOwnedObjectCounters() {
...
+        // parts
+        const ShipDesign* design = ship->Design();
+        if (!design)
+            continue;
+        const std::vector<std::string>& parts = design->Parts();
+        for (const std::string& part : parts) {
+            const PartType* part_type = GetPartType(part);
+            if (!part_type)
+                continue;
+            if (!part_type->Name().empty())
```
geoffthemedio: should check for blank name before looking up, not after, as the latter will never return an empty name if the lookup worked.

Empire/Empire.cpp
```diff
@@ -3372,9 +3372,11 @@ void Empire::ResetMeters() {
 }
 
 void Empire::UpdateOwnedObjectCounters() {
-    // ships of each species and design
+    // ships of each species and design, parts by type and class
     m_species_ships_owned.clear();
     m_ship_designs_owned.clear();
+    m_ship_parts_owned.clear();
```
geoffthemedio: Add a comment explaining the different containers.... one indexed by part name, one by part class, summing either could be used for total ship parts count.

Empire/Empire.h
```diff
@@ -567,10 +567,14 @@ class FO_COMMON_API Empire {
     void                        UpdateOwnedObjectCounters();
 
     int                         TotalShipsOwned() const;
+    /** Total number of parts for all owned ships in this empire */
+    int                         TotalShipPartsOwned() const;
     int                         TotalBuildingsOwned() const;
 
     std::map<std::string, int>& SpeciesShipsOwned()     { return m_species_ships_owned; }
```
geoffthemedio, line 573: make everything line up horizontally...

Empire/Empire.cpp
```diff
@@ -3386,6 +3388,20 @@ void Empire::UpdateOwnedObjectCounters() {
         if (!ship->SpeciesName().empty())
             m_species_ships_owned[ship->SpeciesName()]++;
         m_ship_designs_owned[ship->DesignID()]++;
+        // parts
+        const ShipDesign* design = ship->Design();
+        if (!design)
+            continue;
+        const std::vector<std::string>& parts = design->Parts();
```
geoffthemedio: This might be better moved into ShipDesign. Could add a ShipDesign::PartCount function that returns, or takes in by reference and fills, maps from PartClass and from string to int? That info could also be cached in the ShipDesign in a mutable member after it is calculated...

Empire/Empire.cpp
```diff
@@ -3372,9 +3372,11 @@ void Empire::ResetMeters() {
 }
 
 void Empire::UpdateOwnedObjectCounters() {
-    // ships of each species and design
+    // ships of each species and design, parts by type and class
     m_species_ships_owned.clear();
     m_ship_designs_owned.clear();
+    m_ship_parts_owned.clear();
+    m_ship_part_class_owned.clear();
     for (std::map<int, TemporaryPtr<UniverseObject> >::iterator ship_it = Objects().ExistingShipsBegin();
```
geoffthemedio, line 3375: Probably doesn't matter for a once-per-turn calculation, but could calculate number of parts in each design once, and then multiply by the number of each design the empire owns, instead of processing each ship separately.

parse/IntComplexValueRefParser.cpp
```diff
@@ -102,6 +102,30 @@ namespace parse {
...
+                =   (
+                            tok.ShipPartsOwned_ [ _a = construct<std::string>(_1) ]
+                        >-( parse::label(Empire_token) >    int_value_ref [ _b = _1 ] )
+                        >-( parse::label(Name_token)   >    string_value_ref [ _d = _1 ] )
+                    ) [ _val = new_<ValueRef::ComplexVariable<int>>(_a, _b, _c, _f, _d, _e) ]
+                ;
+
+            ship_part_class_owned
+                = (
+                            tok.ShipPartClassOwned_ [ _a = construct<std::string>(_1) ]
```
geoffthemedio: probably don't need a separate name for this value in FOCS... could construct the parts of class owned variant when a class is specified, and the parts by name owned variant when a string is specified. Shouldn't be any ambiguity between these...

parse/IntComplexValueRefParser.cpp
```diff
@@ -102,6 +102,30 @@ namespace parse {
...
+                        >-( parse::label(Empire_token) >    int_value_ref [ _b = _1 ] )
+                        >-( parse::label(Class_token)  >
```
geoffthemedio: enum_parser<ShipPartClass>() already exists

universe/ValueRef.cpp
```diff
@@ -1278,6 +1278,8 @@ namespace {
             return empire->SpeciesShipsProduced();
         if (parsed_map_name == "SpeciesShipsScrapped")
             return empire->SpeciesShipsScrapped();
+        if (parsed_map_name == "ShipPartsOwned" || parsed_map_name == "ShipPartClassOwned")
```
geoffthemedio: should these really both be calling the same function?

parse/IntComplexValueRefParser.cpp
```diff
@@ -102,6 +102,30 @@ namespace parse {
...
+                            ( tok.ShortRange_       | tok.FighterBay_   | tok.FighterWeapon_
+                            | tok.Shield_           | tok.Armour_
+                            | tok.Troops_           | tok.Detection_    | tok.Stealth_
+                            | tok.Fuel_             | tok.Colony_       | tok.Speed_
+                            | tok.General_          | tok.Bombard_      | tok.Research_
+                            | tok.Industry_         | tok.Trade_        | tok.ProductionLocation_
+                            ) [ _d = new_<ValueRef::Constant<std::string> >(_1) ]
```
geoffthemedio: should probably be consistent about whether to use >> or > > when closing nested templates

geoffthemedio (changes requested):

Morlic-fo:
> @freeorion/ai-team, what adjustments to the AI are required for this?

I will try to look into this tomorrow.

parse/IntComplexValueRefParser.cpp
dbenage-cx: `as_string` seems to return `std::vector<char>` instead of `std::string`.  I'm not certain how to overcome that in this context, hence the use of int. (edit: I'm not certain what value is expected here, but it seems `as_string` is not suitable by itself)

dbenage-cx:
Rebased commits, removing `ShipPartClassOwned` and adding a `class` value for `ShipPartsOwned`.
If both `name` and `class` are specified, only the first is utilized.

parse/IntComplexValueRefParser.cpp
geoffthemedio: hmm, looks like I had the same idea before and gave up on getting it to work, and did what you've done here (for part_class_in_ship_design)

Maybe a solution here, but probably not worth the effort: http://www.doing-stuff.org/2015/04/obtaining-both-the-attribute-and-the-raw-string-in-boostspirit/

geoffthemedio:
@MatGB Related: https://github.com/freeorion/freeorion/issues/1211

Morlic-fo:
What stops me from spamming COMsats or generally empty hulls as cannon fodder?

Such hulls definitely stay relevant even with fighters as they protect the main fleet from capital ship fire in the first turn before fighters are launched.

Vezzra:
> What stops me from spamming COMsats or generally empty hulls as cannon fodder?

Good catch. That means when calculating the cost increase factor we either need to either:

a) still also need to factor in the number of ships (so even ships without any parts on their design get counted)

b) consider the part slots regardless if their are used or not.

MatGB:
c) have a different specific upkeep for the Base hull so it can be used for intended purpose (outpost bases and colony bases), can be used for edge case uses (comsats/chaff/setting system to aggressive) but is too costly to use in large numbers.

I've thought an actual cost per turn in PP for the system it's in would be doable, each base hull in a system reduces PP output in that system by an amount (probably 1).

We could have an individual hull count as a part for counting but it'd still be a very low upkeep amount compared to current, I'm opposed to using slots available instead of slots used because part of my vision for this was to reduce the need players feel to use every slot in every design: if you don't need a fuel tank, don't design it in, but if cost is per slot then you might as well fill everything.

(we do need to work out a better UI for setting a system to aggressive than having a comsat in orbit, but that's outside this remit)

FWIW, I'm now into my 4th game with the original version of this PR that had the upkeep set to 0.2% per part, I *really* like it, it's less upkeep in the early game but a substantial increase in the late game and I'm finding myself using more smaller hulls way into late game in a way I didn't before.

Morlic-fo:
> c) have a different specific upkeep for the Base hull so it can be used for intended purpose (outpost bases and colony bases), can be used for edge case uses (comsats/chaff/setting system to aggressive) but is too costly to use in large numbers.

This is not a solution. I would simply use other cheap hulls. For supporting a battle fleet, I need to use mobile hulls anyway.

dbenage-cx:
Expect we probably want upkeep applied through effects rather than appended to `buildcost` in the definitions.  Such effects would apply to the previous linked empire meters.

PR #1210, opened 2017-01-08
Add various fleet details to fleet summary tooltip

dbenage-cx:


geoffthemedio:
fleet_tooltip
Seems a bit odd now... especially the underline... I'd move that to the title (though call it statistics, perhaps...), add a "total ships count" line that duplicates what the icon number says, add an "unarmed ships" count as well, and maybe count colony ships, outpost ships, ships with fighters, ships with troops, and perhaps monsters.

dbenage-cx:
Placed as anon in MapWnd in the interim.  Not certain if a new include is desired here, or if it is preferred the other existing includes (for map browse wnds) be consolidated somehow.

geoffthemedio:
Looks better, but there's a weird outline around the bottom section of the tooltip now...
fstt

UI/MapWnd.cpp
```diff
@@ -283,6 +284,198 @@ namespace {
...
+                return;
+
+            const std::set<int>& destroyed_objects = GetUniverse().EmpireKnownDestroyedObjectIDs(m_empire_id);
+            for (std::shared_ptr<const Ship> ship : Objects().FindObjects<Ship>()) {
+                if (!ship->OwnedBy(m_empire_id) || destroyed_objects.find(ship->ID()) != destroyed_objects.end())
+                    continue;
+                m_values["MAP_FLEET_SHIP_COUNT"]++;
+
+                if (ship->IsArmed())
+                    m_values["MAP_FLEET_ARMED_COUNT"]++;
```
geoffthemedio: all these string literals appearing in several places should be put into constants defined in one place above the relevant code.

UI/MapWnd.cpp
```diff
@@ -283,6 +284,198 @@ namespace {
...
+                for (const std::string& part : design->Parts()) {
+                    m_values["MAP_FLEET_SLOT_COUNT"] ++;
+                    if (!part.empty())
+                        m_values["MAP_FLEET_PART_COUNT"] ++;
+                }
+            }
+
+        }
+
+        void LayoutRow(const std::string& label, GG::Pt& label_ul, GG::Pt& label_lr, GG::Pt& value_ul, GG::Pt& value_lr, const GG::Pt& row_advance) {
```
geoffthemedio: break this definition into two lines... it's a bit wide now

geoffthemedio: A comment explaining what this function is doing would be helpful

UI/MapWnd.cpp
```diff
@@ -283,6 +284,198 @@ namespace {
...
+                    m_values["MAP_FLEET_COLONY_COUNT"]++;
+
+                if (ship->HasTroops())
+                    m_values["MAP_FLEET_TROOP_COUNT"]++;
+
+                if (ship->HasFighters())
+                    m_values["MAP_FLEET_CARRIER_COUNT"]++;
+
+                const ShipDesign* design = ship->Design();
+                if (design == nullptr)
```
geoffthemedio: I think this is the only place this idiom would be used... I'd stick with `if (!design)`

UI/MapWnd.cpp
```diff
@@ -283,6 +284,198 @@ namespace {
...
+        }
+
+        /** Updates the text displayed for the value of each label */
+        void UpdateLabels() {
+            UpdateValues();
+            for (const std::unordered_map<std::string, int>::value_type& value : m_values) {
+                std::unordered_map<std::string, LabelValueType>::iterator label_it = m_labels.find(value.first);
+                if (label_it == m_labels.end())
+                    continue;
+                label_it->second.second->ChangeTemplatedText(boost::lexical_cast<std::string>(value.second), 0);
```
geoffthemedio: I think std::to_string is now preferred for built-in types (in this case: int)

UI/MapWnd.cpp
```diff
@@ -283,6 +284,198 @@ namespace {
...
+            LayoutRow("MAP_FLEET_TROOP_COUNT",      label_ul, label_lr, value_ul, value_lr, next_row);
+            LayoutRow("MAP_FLEET_COLONY_COUNT",     label_ul, label_lr, value_ul, value_lr, space_row);
+            LayoutRow("MAP_FLEET_SLOT_COUNT",       label_ul, label_lr, value_ul, value_lr, next_row);
+            LayoutRow("MAP_FLEET_PART_COUNT",       label_ul, label_lr, value_ul, value_lr, GG::Pt(GG::X0, GG::Y0));
+
+            Resize(GG::Pt(value_lr.x + (m_margin * 3), value_lr.y + (m_margin * 3)));
+        }
+
+        /** Constructs and attaches new label and value controls for the given label. */
+        void NewLabelValue(const std::string& label, bool title = false) {
```
geoffthemedio: label and m_labels are a bit confusing. rename label to something more explicit, like label_text or label_name or whatever is appropriate

UI/MapWnd.cpp
```diff
@@ -283,6 +284,198 @@ namespace {
...
+                std::shared_ptr<CUILabel>{
+                    new CUILabel{"0", GG::FORMAT_RIGHT,
+                        GG::NO_WND_FLAGS, GG::X0, GG::Y0,
+                        m_col_widths.at(1) - (m_margin * 2), height
+                    }
+                }
+            ));
+
+            if (title) {  // utilize bold font for title label and value
+                m_labels.at(label).first.get()->SetFont(ClientUI::GetBoldFont());
```
geoffthemedio: are these `.get()` calls necessary? You should be able to dereference a shared_ptr directly...

geoffthemedio (changes requested):

UI/MapWnd.cpp
```diff
@@ -283,6 +284,198 @@ namespace {
...
+            NewLabelValue("MAP_FLEET_SLOT_COUNT");
+            NewLabelValue("MAP_FLEET_PART_COUNT");
+            NewLabelValue("MAP_FLEET_UNARMED_COUNT");
+            NewLabelValue("MAP_FLEET_COLONY_COUNT");
+            NewLabelValue("MAP_FLEET_CARRIER_COUNT");
+            NewLabelValue("MAP_FLEET_TROOP_COUNT");
+
+            UpdateLabels();
+            DoLayout();
+        }
```
LGM-Doyle: Have you considered moving all of this, but particularly `UpdateLabels()` into a `FleetDetailBrowseWnd::PreRender()` function.

`MapWnd::RefreshFleetResourceIndicator()` is called every `InitTurn()`, so `UpdateLabels()` will iterate through all of the ships every turn.  The player will probably only refer to this popup on occasion.

The change would be like the following:
````
        FleetDetailBrowseWnd(int empire_id, GG::X width) :
            GG::BrowseInfoWnd(GG::X0, GG::Y0, width, GG::Y(ClientUI::Pts())),
            m_empire_id(empire_id),
            m_margin(5)
        {
            GG::X value_col_width{(m_margin * 3) + (ClientUI::Pts() * 3)};
            m_col_widths = {width - value_col_width, value_col_width};

            RequirePreRender();
        }

        void PreRender() override {
            SetChildClippingMode(ClipToClient);

            NewLabelValue("MAP_FLEET_SHIP_COUNT", true);
            NewLabelValue("MAP_FLEET_ARMED_COUNT");
            NewLabelValue("MAP_FLEET_SLOT_COUNT");
            NewLabelValue("MAP_FLEET_PART_COUNT");
            NewLabelValue("MAP_FLEET_UNARMED_COUNT");
            NewLabelValue("MAP_FLEET_COLONY_COUNT");
            NewLabelValue("MAP_FLEET_CARRIER_COUNT");
            NewLabelValue("MAP_FLEET_TROOP_COUNT");

            UpdateLabels();
            DoLayout();
        }
````

**Review by LGM-Doyle — COMMENTED** [review]
The UI looks good.

I think that the last two lines of the table, could be improved.

````
Total Ship Part Slots 20
Total Ship Parts          18 
````

When I think of that I think 18 of 20 slots are full, but when I read it the order seems wrong.  Swapping the order of the lines would improve it.

Here are two suggestions that are more compact and contain the same information.  I prefer the first.
````
18 of 20 Ship Part Slots Filled
````
or
````
18 Ship Parts in 20 Slots
````

File: UI/MapWnd.cpp (hunk 7 above)
Comment (geoffthemedio) [reply UI/MapWnd.cpp]: `18 Ship Parts in 20 Slots` doesn't nicely fit into the table format.

Comment (LGM-Doyle) [reply UI/MapWnd.cpp]: It is a minor issue.  My eyes are looking for a fraction.  Swapping the line order also improves it for me.  Again it is a quibble.

**Comment by dbenage-cx** [comment]
@geoffthemedio TY for review, changes fixed. (I intend to fixup/squash these changes to the first commit)

@LGM-Doyle TY for PreRender, I've also swapped the `Parts` and `Slots` rows.

> there's a weird outline around the bottom section of the tooltip

Inset border for non-ship counts, this seemed a change more familiar with the other tooltips yet still distinguished from the ship counts.
Should the outline be made more prominent, or is the issue with the style itself? I would be glad to try multiple attempts and post shots for comparison if there are other suggestions.

File: UI/MapWnd.cpp [hunk 8]
```diff
@@ -394,12 +417,12 @@ namespace {
...
             if (title) {  // utilize bold font for title label and value
-                m_labels.at(label).first.get()->SetFont(ClientUI::GetBoldFont());
-                m_labels.at(label).second.get()->SetFont(ClientUI::GetBoldFont());
+                m_labels.at(descr).first->SetFont(ClientUI::GetBoldFont());
+                m_labels.at(descr).second->SetFont(ClientUI::GetBoldFont());
             }
 
-            AttachChild(m_labels.at(label).first.get());
-            AttachChild(m_labels.at(label).second.get());
+            AttachChild(m_labels.at(descr).first.get());
```
geoffthemedio: This might be a bad idea... Are you sure the lifetime of the shared_ptr you're getting the raw pointers from here will exist longer than the member variables storing the children in this AttachChild call? If not, then the shared_ptr might be cleared and the pointed to Wnd destructed before the children pointers are cleared, potentially leading to using or attempting to delete an invalid pointer..

geoffthemedio:
> Should the outline be made more prominent, or is the issue with the style itself?

Issue is style. The embossing / shadowing like an extruded button is odd, and doesn't look like it was intended to look like that.

But it can be fixed later...

PR #1214, opened 2017-01-13
Allow Process() to return immediately.  linux only.

LGM-Doyle:
Removes sleep(1) from the original side of the fork() in linux.

In all cases where Process() is actually used the human client or the
server contacts the child process asynchronously via networking.

sleep(1) did not guarantee the new process was fully initialized.  It
also imposed a 1 second per AI startup penalty.

### DIFF
--- a/util/Process.cpp
+++ b/util/Process.cpp
@@ -206,8 +206,7 @@ Process::Impl::Impl(const std::string& cmd, const std::vector<std::string>& argv
         break;
     }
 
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Any objections @adrianbroher  ?

adrianbroher:
No objections.

LGM-Doyle:
@adrianbroher thanks for the review.

PR #1215, opened 2017-01-13
Reduce AI polling interval from 250ms to 10ms to increase responsiveness.

LGM-Doyle:
1/4 second is too long.  The AI will never appear responsive to callbacks.  I chose 10ms as long from a processor perspective, but slightly shorter than a video refresh interval.


geoffthemedio:
ok

PR #1216, opened 2017-01-13
Improve combat log differential serialization

LGM-Doyle:
Currently, every turn update the server sends all combat logs since turn 1 to every player, human and AI.  The AIs never use the combat reports.  

Profiling on my machine showed that the time spent sending the logs was about 0.8 ms per log, per player, per turn.  If the game goes long enough parsing and sending the logs becomes the biggest time component of processing the turn.

This PR changes the implementation so that each turn all players are informed of the number of available combats logs and then only the human clients request any combat logs that they do not already have.  This means that during a game each combat log is only downloaded once per human client.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -893,7 +893,6 @@ void HumanClientApp::UpdateCombatLogs(const Message& msg){
     {
         GetCombatLogManager().CompleteLog(it->first, it->second);
     }
-
 }
 
 void HumanClientApp::HandleWindowMove(GG::X w, GG::Y h) {
@@ -999,9 +998,8 @@ void HumanClientApp::HandleTurnUpdate()
 { UpdateCombatLogManager(); }
 
 void HumanClientApp::UpdateCombatLogManager() {
-    boost::optional<std::vector<int> > incompleteIDs(GetCombatLogManager().IncompleteLogIDs());
-    if (incompleteIDs)
-        m_networking.SendMessage(RequestCombatLogsMessage(PlayerID(), *incompleteIDs));
+    if (auto incomplete_ids {GetCombatLogManager().IncompleteLogIDs()})
+        m_networking.SendMessage(RequestCombatLogsMessage(PlayerID(), *incomplete_ids));
 }
 
 namespace {
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -977,6 +977,11 @@ void HumanClientApp::StartGame() {
 
 void HumanClientApp::HandleTurnUpdate() {
 
+    if (GetCombatLogManager().HasIncompleteLogs() && m_networking.Connected()) {
+        m_networking.SendMessage(
+            RequestCombatLogsMessage(EmpireID(),
+                                     GetCombatLogManager().IncompleteLogIDs()));
+    }
 }
 
 namespace {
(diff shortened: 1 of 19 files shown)


Morlic-fo:
> The AIs never use the combat reports.

Yet. 
It would be nice to have the AI parse combat logs to estimate the strength of an enemy but that probably will not come any time soonish, so it probably makes sense not sending them for now...

client/human/HumanClientApp.cpp
```diff
@@ -878,6 +880,22 @@ void HumanClientApp::HandleSaveGameDataRequest() {
...
+    std::vector<std::pair<int, CombatLog> > logs;
+    ExtractDispatchCombatLogsMessageData(msg, logs);
+
+    // Update the combat log manager with the completed logs.
+    for (std::vector<std::pair<int, CombatLog> >::const_iterator it = logs.begin();
+         it != logs.end(); ++it)
+    {
+        GetCombatLogManager().CompleteLog(it->first, it->second);
+    }
+
```
geoffthemedio: extra blank line

client/human/HumanClientApp.cpp
```diff
@@ -973,6 +991,17 @@ void HumanClientApp::StartGame() {
...
         map_wnd->ResetEmpireShown();
+
+    UpdateCombatLogManager();
+}
+
+void HumanClientApp::HandleTurnUpdate()
+{ UpdateCombatLogManager(); }
+
+void HumanClientApp::UpdateCombatLogManager() {
+    boost::optional<std::vector<int> > incompleteIDs(GetCombatLogManager().IncompleteLogIDs());
```
geoffthemedio: rename to incomplete_ids

geoffthemedio: why does this need to be an optional?

combat/CombatLogManager.cpp
```diff
@@ -149,69 +151,218 @@ template
...
+    }
 }
 
-void CombatLogManager::SetLog(int log_id, const CombatLog& log)
+void CombatLogManager::CombatLogManagerImpl::SetLog(int log_id, const CombatLog& log)
 { m_logs[log_id] = log; }
 
+boost::optional<std::vector<int> > CombatLogManager::CombatLogManagerImpl::IncompleteLogIDs() const
+{
+    if (m_incomplete_logs.empty())
```
geoffthemedio, line 251: what's wrong with returning an empty vector...?

geoffthemedio (changes requested):

Vezzra:
> Yet.
It would be nice to have the AI parse combat logs to estimate the strength of an enemy but that probably will not come any time soonish, so it probably makes sense not sending them for now...

Well, if I understand the description correctly, the combat logs (of past turns) are only downloaded once for each game *session* (meaning, after loading a game each client would request all the past combat logs on the first turn update, but not on the subsequent ones).

So even if the AI clients would request those logs too at some point, that's still a big improvement.

client/human/HumanClientApp.cpp
LGM-Doyle: fixed.

client/human/HumanClientApp.cpp
LGM-Doyle: It does not need to be an `optional`.

`optional` has the following advantages:
- `optional` is semantically clearer regardless of the container type.
- an empty `optional` is smaller than an empty STL container.
- I was anticipating the transition to C++11.  Using `optional` with type inference means that the container type can be changed without this code changing.  I wrote this code before the transition to C++11 completed and did not go through and update anything.  I've updated this one line to illustrate why `optional` with type inference is more elegant.

````    
if (auto incomplete_ids {GetCombatLogManager().IncompleteLogIDs()})
        m_networking.SendMessage(RequestCombatLogsMessage(PlayerID(), *incomplete_ids));
````
- `std::optional` is available in C++17 as the standard way to express an optional/possible contained value.

LGM-Doyle (commented):
Thanks for taking the time to review this PR.

LGM-Doyle:
@Vezzra your understanding is correct.  

As an aside, the current code has a TODO indicating that filtering of combat logs should be done in the future.  Currently, all combat logs are available to all clients.  In future, if that filtering was added, then clients would only be able to review combat logs for combat that they observed (most generous) or participated in (most mysterious).

PR #1223, opened 2017-01-17
Add aggregation to common python timer.

LGM-Doyle:
This PR moves the portions of the AI timer that are not dependent on the `freeOrionAIInterface` into the python common code.  It then adds aggregation to the timer.  Aggregation means adding up the times of recursive sections of code and printing the sum with `print_aggregate()`.  

The original style of printing is available with `print_flat()`.  Calling `end()` still defaults to `print_flat()`.  

This addition to the `Timer` helped in profiling the universe generation code.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -249,8 +249,9 @@ def generateOrders():  # pylint: disable=invalid-name
             main_timer.stop()
         except Exception as e:
             print_error(e, location=action.__name__)
-    main_timer.end()
-    turn_timer.end()
+    main_timer.stop_and_print()
+    turn_timer.stop_and_print()
+
     # Turn timer is file scope in order to capture times from different functions
     # in this file, but it produces output each turn.  It needs to be cleared
     # once per turn so that the aggregate times only reflect this turn.
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -28,13 +28,13 @@ import ResearchAI
 import ResourcesAI
 import TechsListsAI
 from AIDependencies import INVALID_ID
-from freeorion_tools import UserStringList, chat_on_error, print_error, UserString, handle_debug_chat, Timer, init_handlers
+from freeorion_tools import UserStringList, chat_on_error, print_error, UserString, handle_debug_chat, FOTimer, init_handlers
 from common.listeners import listener
 from character.character_module import Aggression
 from character.character_strings_module import get_trait_name_aggression, possible_capitals
 
-main_timer = Timer('timer', write_log=True)
-turn_timer = Timer('bucket', write_log=True)
+main_timer = FOTimer('timer', write_log=True)
+turn_timer = FOTimer('bucket', write_log=True)
 
 using_statprof = False
 try:
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -251,6 +251,7 @@ def generateOrders():  # pylint: disable=invalid-name
             print_error(e, location=action.__name__)
     main_timer.end()
     turn_timer.end()
+    turn_timer.clear()
     turn_timer.start("Server_Processing")
 
     try:
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -249,13 +249,8 @@ def generateOrders():  # pylint: disable=invalid-name
             main_timer.stop()
         except Exception as e:
             print_error(e, location=action.__name__)
-    main_timer.stop_and_print()
-    turn_timer.stop_and_print()
-
-    # Turn timer is file scope in order to capture times from different functions
-    # in this file, but it produces output each turn.  It needs to be cleared
-    # once per turn so that the aggregate times only reflect this turn.
-    turn_timer.clear_data()
+    main_timer.stop_print_and_clear()
+    turn_timer.stop_print_and_clear()
     turn_timer.start("Server_Processing")
 
     try:
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -28,13 +28,13 @@ import ResearchAI
 import ResourcesAI
 import TechsListsAI
 from AIDependencies import INVALID_ID
-from freeorion_tools import UserStringList, chat_on_error, print_error, UserString, handle_debug_chat, FOTimer, init_handlers
+from freeorion_tools import UserStringList, chat_on_error, print_error, UserString, handle_debug_chat, AITimer, init_handlers
 from common.listeners import listener
 from character.character_module import Aggression
 from character.character_strings_module import get_trait_name_aggression, possible_capitals
 
-main_timer = FOTimer('timer', write_log=True)
-turn_timer = FOTimer('bucket', write_log=True)
+main_timer = AITimer('timer', write_log=True)
+turn_timer = AITimer('bucket', write_log=True)
 
 using_statprof = False
 try:
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -251,7 +251,10 @@ def generateOrders():  # pylint: disable=invalid-name
             print_error(e, location=action.__name__)
     main_timer.end()
     turn_timer.end()
-    turn_timer.clear()
+    # Turn timer is file scope in order to capture times from different functions
+    # in this file, but it produces output each turn.  It needs to be cleared
+    # once per turn so that the aggregate times only reflect this turn.
+    turn_timer.clear_data()
     turn_timer.start("Server_Processing")
 
     try:
--- a/default/python/AI/freeorion_tools/timers.py
+++ b/default/python/AI/freeorion_tools/timers.py
@@ -44,6 +44,15 @@ class DummyTimer(object):
     def end(self):
         pass
 
+    def clear(self):
+        pass
+
+    def print_flat(self):
+        pass
+
+    def print_aggregate(self):
+        pass
+
 
 class FOLogTimer(Timer):
     """A Timer with a FO AI engine dependent extension that logs timer results to a file each turn.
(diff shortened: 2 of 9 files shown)


default/python/AI/freeorion_tools/timers.py
```diff
@@ -123,4 +100,4 @@ def end(self):
             self._write(''.join(row))
         self.timers = []  # clear times
 
-Timer = LogTimer if USE_TIMERS else DummyTimer
+FOTimer = FOLogTimer if USE_TIMERS else DummyTimer
```
Cjkjvfnby: Maybe leave `Timer` name as it was? And rename `commons.Timer`  to `BaseTimer`

I am against `FO` prefix. It does not explain anything every part of code is `FO`. Use `AI`

default/python/common/timers.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        Stop prev timer if present and starts new.
+        """
+        self.stop()
+        self.section_name = section_name
+        self.start_time = time()
+
+    def _print_timer_table(self, time_table):
+        """
+        Print a header and a footer and fill it with the timing results from time_table
```
Cjkjvfnby: Put the come to the end of sentence.   Please check all of the docstrings.

default/python/common/timers.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        if not self.timers:
+            return
+        max_header = max(len(x[0]) for x in self.timers)
```
Cjkjvfnby: Add  TODO  to use this https://github.com/freeorion/freeorion/blob/master/default/python/common/print_utils.py#L116

default/python/AI/FreeOrionAI.py
```diff
@@ -251,6 +251,7 @@ def generateOrders():  # pylint: disable=invalid-name
             print_error(e, location=action.__name__)
     main_timer.end()
     turn_timer.end()
+    turn_timer.clear()
```
Cjkjvfnby: Why do we need `clear`?

LGM-Doyle: We don't need the clear and I have removed it.

The AI version of the timer clears its times when end() is called, so it does not need the clear().

The standard version of the timer does not, so that you can stop the timer and then use either of the print functions.  If it was the standard timer then it would need to be cleared once per turn in order to be a turn timer.

Cjkjvfnby:
We have `stop`, `end`, and `clear`  thats a little confusing.

We have two timers:
- clean on each turn (AITimer)
- don't clean on each turn (unknown timer)

Maybe we will always store full history for current session (number of calls and mean time)?

Also please provide log examples to PR descriptions.

LGM-Doyle:
The names `start`, `stop` and `end` predate this commit.

I kept `stop()` since start and stop are a strongly couple antonym pair and `stop()` stops the timer that `start()` started.

I changed `clear()` to `clear_data()`.

I changed `end()` to `stop_and_print()`.  I removed the clear functionality from the `AITimer` `stop_and_print()`.  The `AITimer.end()` function did everything, stopped the timers, printed to the log, printed to disk and cleared the timers.  Now it only does the first three.

I added `clear_data()` to the timers in AI and explained with a comment where appropriate why the clear was needed.  Clear is needed when a file scope timer is declared to collect timing data from multiple functions in the file, but only print once per turn.

Here is an example of the two styles of tables taken from the main timer in FreeOrionAI.py.

````
 : Timing for timer:
 : ======================================================
 : survey_universe                                 2 msec
 : find_best_designs_this_turn                    51 msec
 : calculate_priorities                           22 msec
 : assign_scouts_to_explore_systems                0 msec
 : assign_colony_fleets_to_colonise                0 msec
 : assign_invasion_fleets_to_invade                0 msec
 : assign_military_fleets_to_systems               1 msec
 : generate_fleet_orders_for_fleet_missions        2 msec
 : issue_fleet_orders_for_fleet_missions           1 msec
 : generate_research_orders                        5 msec
 : generate_production_orders                      7 msec
 : generate_resources_orders                      12 msec
 : survey_universe                                 2 msec
 : find_best_designs_this_turn                    35 msec
 : calculate_priorities                           45 msec
 : assign_scouts_to_explore_systems                1 msec
 : assign_colony_fleets_to_colonise                0 msec
 : assign_invasion_fleets_to_invade                0 msec
 : assign_military_fleets_to_systems               1 msec
 : generate_fleet_orders_for_fleet_missions        3 msec
 : issue_fleet_orders_for_fleet_missions           1 msec
 : generate_research_orders                       23 msec
 : generate_production_orders                      5 msec
 : generate_resources_orders                      11 msec
 : survey_universe                                 4 msec
 : find_best_designs_this_turn                    12 msec
 : calculate_priorities                           41 msec
 : assign_scouts_to_explore_systems                1 msec
 : assign_colony_fleets_to_colonise                0 msec
 : assign_invasion_fleets_to_invade                0 msec
 : assign_military_fleets_to_systems               1 msec
 : generate_fleet_orders_for_fleet_missions        5 msec
 : issue_fleet_orders_for_fleet_missions           4 msec
 : generate_research_orders                       10 msec
 : generate_production_orders                      7 msec
 : generate_resources_orders                      17 msec
 : ------------------------------------------------------
 :                                   Total:      347 msec
 : 
 : Timing for timer:
 : ======================================================
 : survey_universe                                 9 msec
 : find_best_designs_this_turn                    99 msec
 : calculate_priorities                          108 msec
 : assign_scouts_to_explore_systems                3 msec
 : assign_colony_fleets_to_colonise                1 msec
 : assign_invasion_fleets_to_invade                0 msec
 : assign_military_fleets_to_systems               4 msec
 : generate_fleet_orders_for_fleet_missions       11 msec
 : issue_fleet_orders_for_fleet_missions           8 msec
 : generate_research_orders                       39 msec
 : generate_production_orders                     20 msec
 : generate_resources_orders                      40 msec
 : ------------------------------------------------------
 :                                   Total:      347 msec
```` 

The first table is the flat style and the second is the aggregate style that this PR adds.  Note this PR does not change any of the current tables from the flat style to the aggregate style.  The flat style is often what you want, for example if you are trying to determine if an incremental algorithm is doing most of the work in earlier/later iterations.

File: default/python/AI/FreeOrionAI.py [hunk 5]
```diff
@@ -249,8 +249,13 @@ def generateOrders():  # pylint: disable=invalid-name
             main_timer.stop()
         except Exception as e:
             print_error(e, location=action.__name__)
-    main_timer.end()
-    turn_timer.end()
+    main_timer.stop_and_print()
+    turn_timer.stop_and_print()
+
+    # Turn timer is file scope in order to capture times from different functions
```
Cjkjvfnby: Main timer should not be cleared?
IMHO this comment not needed.

Cjkjvfnby (changes requested):
I see next timers commands:
 - `start(section)` start new section with name
 - `stop(additional_info="")` stops current section adds additional_info to its name
 - print(aggregate=False)  all prints goes here
 - reset() Reset timer to initial state 

But I don't see usecases where I `reset` timer and don't `print` its value.  It is place to mistake, forget to reset timer -> get growing print each time.  And I already see such place in your code.  **Any reason why print and reset should not be merged to single method?**

For aggregated add next columns to table:  number of calls and mean time.  At least TODO, because it should be done along with  adding Table support in the same time.

LGM-Doyle:
I have added a `stop_print_and_clear()` to the `AITimer`.

I added a `print_statistics()` to the `Timer` that prints the number of samples, mean, standard deviation and the maximum sample time.

Here is an example of the output:
````: Timing statistics for timer:
: ==============================================================================================================
: survey_universe                          num:      5, mean:   3.57 msec, std:   1.16 msec, max:   5.38 msec
: find_best_designs_this_turn              num:      5, mean:  21.02 msec, std:  15.01 msec, max:  42.27 msec
: calculate_priorities                     num:      5, mean:  37.40 msec, std:   9.13 msec, max:  54.73 msec
: assign_scouts_to_explore_systems         num:      5, mean:   1.53 msec, std:   0.44 msec, max:   2.19 msec
: assign_colony_fleets_to_colonise         num:      5, mean:   0.60 msec, std:   0.04 msec, max:   0.65 msec
: assign_invasion_fleets_to_invade         num:      5, mean:   0.18 msec, std:   0.03 msec, max:   0.22 msec
: assign_military_fleets_to_systems        num:      5, mean:   1.60 msec, std:   0.20 msec, max:   1.97 msec
: generate_fleet_orders_for_fleet_missions num:      5, mean:   3.88 msec, std:   0.38 msec, max:   4.38 msec
: issue_fleet_orders_for_fleet_missions    num:      5, mean:   3.32 msec, std:   1.10 msec, max:   5.27 msec
: generate_research_orders                 num:      5, mean:  15.58 msec, std:  11.92 msec, max:  36.54 msec
: generate_production_orders               num:      5, mean:   7.53 msec, std:   1.88 msec, max:  10.98 msec
: generate_resources_orders                num:      5, mean:  16.71 msec, std:   3.49 msec, max:  20.28 msec
: ==============================================================================================================
````
There have been times when I wanted to print the result, but not clear the times.

For example, when I want to print the flat and the aggregate tables together from the same data.  This allows me to use the aggregate time to determine which sections are taking the longest and then check the flat print out to for more information.  I used this when profiling the monster placement code to determine that while portions of the placement were time consuming it was really specific FOCS clauses in specific monsters that caused the problem.  Neither table on its own was sufficient.

If the times are cleared every time it prints, then you can only print one type of result.

I have also used incrementally adding to a timer and the growing printout to examine the computational complexity of an algorithm overtime.  It allows me to compare all of the times on one screen while the timer continues to accumulate results.

For those reasons, the `Timer` is more useful if `clear_data()` and `print_x()` are separate functions.

Another solution is to remove clear entirely.  Then the `Timer` stores all data until it is destroyed.  This happens naturally with timers declared as local variables within functions, which cease to exist when they go out of scope.  Then you can never make a mistake, because you can't clear.  This is almost what clear does now, except it does not reset a running timer.

I do not understand how your proposed `start(section)` is different from the current version of start which takes the section name as its parameter.

For stop(additional_info=""), why not add, in future, if needed `rename_section(new_name)` .  Then `stop()` continues to serve a single purpose and `rename_section()` could also serve a single purpose.

Why add `reset()` when it is the same as creating a new timer `some_timer = Timer(some_timer.timer_name)`

Cjkjvfnby (approved):

LGM-Doyle:
@Cjkjvfnby thanks for taking the time to review and improve this PR.

PR #1224, opened 2017-01-17
Move parsing of unlocked buildings into Parser.cpp from empires.py

LGM-Doyle:
This is a less adhoc solution than parsing the unlocked buildings in the
universe generation python code.  This parses the unlocked buildings
withinin the regular parser.


geoffthemedio:
It's not really "unlocking" the buildings... Rather, it's creating them at the start of the game... so I'd call the relevant functions / lists starting_buildings rather than unlocked_buildings.

python/server/ServerWrapper.cpp
```diff
@@ -321,6 +321,18 @@ namespace {
         return py_items;
     }
 
+    // Wrapper for starting buildings
+    list LoadStartingBuildings() {
+        list py_items;
+        std::vector<ItemSpec> buildings;
+        parse::starting_buildings(buildings);
+        for (std::vector<ItemSpec>::const_iterator building = buildings.begin()
+                 ; building != buildings.end(); ++building) {
```
geoffthemedio: weird indenting, move ; to end of line, move { to new line for a multi-line for block

default/python/universe_generation/empires.py
```diff
@@ -422,10 +422,9 @@ def setup_empire(empire, empire_name, home_system, starting_species, player_name
         print "Player", player_name, ": no available foci on homeworld for starting species", starting_species
 
     # give homeworld starting buildings
-    # use the list provided in scripting/starting_unlocks/buildings.inf
     print "Player", player_name, ": add starting buildings to homeworld"
-    for building in load_string_list(os.path.join(fo.get_resource_dir(), "scripting/starting_unlocks/buildings.inf")):
-        fo.create_building(building, homeworld, empire)
+    for item in fo.load_starting_buildings():
+        fo.create_building(item.name, homeworld, empire)
```
geoffthemedio, line 427: what happens if the parsed starting_buildings file had something other than a building in it? the item should still be parsed, and would be passed in here...

LGM-Doyle:
I've made the name changes.  I left `bulidings.inf` in the directory `scripting/starting_unlocks/` since the name implies that it is for either starting or unlocking stuff.

parse/StartingBuildingsParser.cpp
```diff
@@ -0,0 +1,50 @@
...
+            void (std::vector<ItemSpec>&),
+            parse::skipper_type
+        > start_rule;
+
+        start_rule start;
+    };
+}
+
+namespace parse {
+    bool starting_buildings(std::vector<ItemSpec>& starting_buildings_) {
```
geoffthemedio: does this need to be in a new file? could you just add this function to ItemsParser.cpp instead, and use the existing rule definition in that file? (assuming you want to keep it parting items and not just a list of strings...)

geoffthemedio (changes requested):

python/server/ServerWrapper.cpp
adrianbroher: Why don't you use a range based loop?

Vezzra:
Why move parsing of the starting buildings into the C++ backend code? When universe generation was migrated to Python scripts, the original C++ code which did that parsing has *deliberately* been removed to do the parsing entirely in Python.

So this isn't an adhoc solution. The idea has been (still is) to remove parsing from the C++ backend code for *all* content files that are solely used for universe generation, and never get called elsewhere in the C++ code. Meaning, this PR is actually a step backward.

My long term plans regarding universe generation specific content definitions is to migrate them entirely from FOCS to Python. I'd prefer to strictly separate between content scripts that specify *what* a particular piece of content (special, building, species, etc.) is/does, or *how* it works, and content scripts that define *where* content gets placed during universe generation.

Insofar the starting buildings file can be considered a stop gap solution, but with the intend to convert it to a Python module at some point, not move parsing it back to C++.

So unless there is a specific reason why that parsing should/needs to be done in C++, I don't see the point - especially since the current implementation in Python looks simpler...

default/python/universe_generation/empires.py
Vezzra: Which is why using the unlocked items parser for this is not ideal. The starting buildings file has always been a simple list of building names, which avoided issues like that.

LGM-Doyle:
I didn't realize that I had stepped into a partially complete project.  

It looked like someone without access to a C++ compiler had added some basic parsing inline in a python file.  While improving the startup performance, I found that the universe generation code runs along until it decides to stop for... disk...  access... for... each... empire.

If the way that it is going to be is do the universe generation parsing in this manner in python then ignore this PR.

However, I will still make the arguments for the benefits of doing all parsing with a single syntax in a single location at a single time.    

- Having a single syntax for all of the content, including the universe generation content is more valuable than having a simpler syntax for a few files.  Modders are always certain how to format files, without having to read either C++/python code and searching for where this particular file is parsed.

  I understand that converting from FOCS to FOCS2 is huge undertaking.  Converting/adding individual files with syntaxes just complicated enough to do a specific job means that modders need to guess what some programmer thought was a simple syntax.  Is it comma, space, or return separated. Is case significant. Are comments allowed? Where do the error messages appear?   

  In the `starting_unlocks` directory the only file with a different non FOCS syntax is the `buildings` file.  

- The single location should be the server C++ code.  

  First, if the game rules and mechanics are parsed in each executable, then that opens the possibility of accidental/intentional errors in multiplayer environments when clients are not run on the same machine and might have separate startup files.  

  Second, while you argue that only the universe code needs to know the galaxy generation code, I've noticed that the AI performs better on cluster galaxies, than either of the spiral galaxies.  I don't know why, but it indicates that it would help the AI to know the startup parameters in order to be better.  This also applies to the AI handling of monsters.  This would require the parsing results be known to the server, so that it can inform the clients.

- The single time that required parsing should be done is as early as possible.

  In a few places in the code parsing of game content is left late and done inline with calculations.  The disk access slows down the whole show.  Parsing should be run in a thread as early as possible, even before the python engine is started up.  Once startup is running you know you need the parse results, so you might as well hide the disk access time.

 So this specific file, `buildings.inf`, it has a unique format, is unavailable to the server and is parsed once per empire inline. That is why it should be fixed.

Vezzra:
> I didn't realize that I had stepped into a partially complete project.

No problem, these things tend to happen in open source projects like this, and that particular "project" has been lying dormant for years. I never found the time and motivation to pick up where I left, there have always been more important things to attend to...

> It looked like someone without access to a C++ compiler had added some basic parsing inline in a python file.

Yeah, I realize how the current "solution" looks. It *is* nothing more than a quick and dirty hack, a placeholder for a more thorough solution (that never saw the light of day... 😉).

> While improving the startup performance, I found that the universe generation code runs along until it decides to stop for... disk... access... for... each... empire.

I admit I never gave the performace aspect of that approach a thought. Is the performance impact of this really that significant? After all, even with as many as 40 players, considering todays disk access times and how small that file is, that should be hardly noticeable...?

Anyway, *that* particular issue should be easy to fix, just read the contents of the file only on first access, then store them in a cache.

> Having a single syntax for all of the content, including the universe generation content is more valuable than having a simpler syntax for a few files.

While having the same *basic* syntax, the different types of FOCS files still differ in certain details and are of varying complexity already, so I don't quite see why we should make things needlessly complicated for `buildings.inf`. Especially if the syntax used allows for entries that don't make sense in that file and require additional checks in the Python code to safeguard against such invalid entries.

> In the starting_unlocks directory the only file with a different non FOCS syntax is the buildings file.

If the concern is mostly about that file not even following the basic syntax of FOCS, at least don't use the item spec parser which is intended for *unlocking* different types of game content at game start, but use a syntax that restricts the contents to buildings. Like this:
```
Building name = "BUILDING_NAME"
```
Which requires implementation of an extra parser just for this, even if it's a very simple one.

> The single location should be the server C++ code.
>
> First, if the game rules and mechanics are parsed in each executable, then that opens the possibility of accidental/intentional errors in multiplayer environments when clients are not run on the same machine and might have separate startup files.

Which is a compelling argument why all content needs to be parsed only by the server, and clients should acquire all relevant data by querying the server. I very much agree on that account, however, currently we're far from that, and implementing that approach is going to be a quite huge project by itself.

And I can't think of a case where it would make sense for the clients to access the scripts and content files which are *only* used to create the game universe. They have already access to the result (the created game universe), and can get all relevant data from that. What's the point of providing them with the rulesets that are used to create the game universe?

> Second, while you argue that only the universe code needs to know the galaxy generation code, I've noticed that the AI performs better on cluster galaxies, than either of the spiral galaxies. I don't know why, but it indicates that it would help the AI to know the startup parameters in order to be better. This also applies to the AI handling of monsters. This would require the parsing results be known to the server, so that it can inform the clients.

Um, the *startup parameters* are a very different matter, which can and probably should be provided to the clients. But as far as the "game universe creation rulesets" (the Python scripts and associated "config" files) are concerned, I can't think of a case where the clients (particularly the AI) could gain any information from them that they could not gain from the game universe itself.

To stay with the topic under discussion, the starting buildings: Just query the buildings you got on your starting homeworld.

Or am I missing something here...?

> The single time that required parsing should be done is as early as possible.

Ok, that's certainly something I can agree to. For universe generation related stuff that could be done by parsing all associated "config" files as a first step in `universe_generator.py` and storing them in proper data structures.

> In a few places in the code parsing of game content is left late and done inline with calculations. The disk access slows down the whole show. Parsing should be run in a thread as early as possible, even before the python engine is started up.

Is the performance hit really that bad, even on old, slow systems that an early parsing in the Python scripts wouldn't be sufficient (for stuff that only the Python scripts would need, that is)?

All that said, I want to make clear that I don't have a strong objection against moving back parsing of the starting buildings back into C++, it's just that I don't really see the need.

But if you think that's better than leaving the current (admittedly ugly) stop-gap solution, even considering that in the long run such things might be moved entirely into Python code (so it would just be a "better"/more consistent stop-gap solution), by all means be my guest. 😉 

The only thing I'd insist on is to not use the item spec parser (which would allow to define non-building items as starting buildings), but to provide a dedicated parser which ensures that only buildings can be defined in that file.

geoffthemedio:
>> The single location should be the server C++ code.
>>
>> First, if the game rules and mechanics are parsed in each executable, then that opens the possibility of accidental/intentional errors in multiplayer environments when clients are not run on the same machine and might have separate startup files.
>
>Which is a compelling argument why all content needs to be parsed only by the server, and clients should acquire all relevant data by querying the server.

The human client needs to parse the content files itself to populate the pedia when not connected to a server. The clients parsing everything themselves avoids the need to send all that information over the network, though I don't know if that's actually worth worrying about having to do.

Morlic-fo:
If it is supposed to be in the python scripts, why not simply add a trivial data structure

```
buildings_on_homeworld = ["Foo", "Bar", "Timmy"]
```
Seems like a 3 minute fix that eliminates any need for parsing and is just as editable...

Vezzra:
> The human client needs to parse the content files itself to populate the pedia when not connected to a server.

That isn't mutually exclusive with the concept that for an actual game the client needs to retrieve content data from the server. Players just need to be aware that when connecting to games hosted on remote servers content might be different from what they have in their local installation.

> The clients parsing everything themselves avoids the need to send all that information over the network

That seems like a minor drawback to me, compared to the issues that arise when players need to manually take care that in multiplayer games the the entire contents of the default directory needs to be exactly identical.

Which might not be so much of an issue as long as all participating players use the stock contents, but becomes an issue as soon as customized content is used.

The only issue I see is with graphic and sound files, sending them over the network will certainly be a problem (because of the large amount of data involved). However, a clever implementation of a protocol that ensures only data is send to the clients that is different on the server should take care of that.

I realize that implementing something like that isn't going to be easy at all.

Vezzra:
> Seems like a 3 minute fix that eliminates any need for parsing and is just as editable...

The problem isn't that we can't migrate the file that defines the starting buildings easily, the problem is that if we start with that, content scripts that define things for universe generation and stuff you get at game start get scattered around different locations, which makes things confusing for content scripters.

If we're going to open that can of worms, we need to do it thoroughly, and carefully think through what needs to go where, so it's easy for content scripters to know where to find what. Which is basically what I meant when talking about the idea to separate all strictly universe generation related stuff (that is, all scripts that define what things are placed where during universe generation) from the content scripts that define the things themselves and how they work.

E.g. having the starting buildings in a Python file in `default/python/universe_generation` and the starting fleets in a FOCS file in `default/scripting/starting_unlocks` isn't something I expect content scripters to appreciate much...

dbenage-cx:
From the arguments so far, it sounds like these could be either side effects of some other definition (e.g. tech), or a candidate for a (still theoretical) content config setting.

OT:

> The only issue I see is with graphic and sound files, sending them over the network will certainly be a problem (because of the large amount of data involved). However, a clever implementation of a protocol that ensures only data is send to the clients that is different on the server should take care of that.

I'd envision the server would require each client to have a specific version of a given content set, specifically to ensure the correct binary assets and stringtables are available.

Morlic-fo:
> The problem isn't that we can't migrate the file that defines the starting buildings easily, the problem is that if we start with that, content scripts that define things for universe generation and stuff you get at game start get scattered around different locations, which makes things confusing for content scripters.

What are the files you are talking about? Only those in ```(default)/scripting/starting_unlocks```?

All of those seem trivial to have in a single file ```starting_unlocks.py```. For buildings, 
```starting_buildings = ["FOO", "BAR", ...]```
seems fair.
For starting fleets, adding a single dict
```starting_fleets = {"SCOUT_FLEET": ["SD_SCOUT"], ...}```
seems also perfectly editable.

For starting unlocks, just move everything into techs (e.g. Mass Driver1 is unlocked that way) and define
```starting_techs = ["TECH1", "TECH2", ...]```

**Review by geoffthemedio — APPROVED** [review]

File: python/server/ServerWrapper.cpp [hunk 4]
```diff
@@ -321,6 +321,20 @@ namespace {
         return py_items;
     }
 
+    // Wrapper for starting buildings
+    list LoadStartingBuildings() {
```
geoffthemedio, line 325: If you call this function for every empire, isn't it going to re-load and parse them all for every empire still? Thought part of the point was to avoid doing the disk access repeatedly.

LGM-Doyle: Yes.

Moving it out of the python into C++ made it faster on its own.  

I said that parsing them in a central location and early would be optimal, but this PR just moves them to a central location and a standard syntax.

Collecting all the startup parses and doing them early was on the list for this branch later, but I ran out of steam.

MatGB:
>  I've noticed that the AI performs better on cluster galaxies, than either of the spiral galaxies. I don't know why

I hadn't noticed this before, but started a Clusters game yesterday and the AI is behaving remarkably better than in the last few months where I've been almost exclusively playing Spirals.

I, obviously, have zero clue why, and it's just one game so far, but if there is a backend reason, possibly including better knowledge of layout, then applying that to all would be a good plan.

Vezzra:
> What are the files you are talking about? Only those in (default)/scripting/starting_unlocks?

No, not only those. There is also `monster_fleets.inf` in "default/scripting", and the specials definitions contain data specifically for placement in universe generation.

Vezzra:
Time to get back to this, before I forget it completely... 😳 

> I've made the changes so that it uses the item parse but now generates an error message if there is an unrecognized building in the file. I think this is a reasonable solution that doesn't generate a new parser for one file ( @geoffthemedio 's concern), but it prints a clear error if there is a non-building in the file ( @Vezzra 's concern). The current solution happily parses non buildings without a warning.

I'd still prefer a separate parser (requiring the item spec syntax for something that's just a list of buildings strikes me as confusing from a content scripters POV), but if @geoffthemedio prefers to avoid that, I guess I can live with it.

Concerning the rest of the ongoing discussion, we should continue that on the forum, this really isn't the place for it. Although I probably won't get back to that soon, maybe once we start getting into migrating the content scripts to Python.

As fas as this PR is concerned, I guess we can go ahead and merge. Unless anyone has any comments/objections...?

PR #1225, opened 2017-01-17
Improve universe generation performance.

LGM-Doyle:
This PR along with #1214 provide a performance improvement to universe generation.

The performance improvement on a galaxy with 999 systems and 39 AIs is from 83 seconds to to 21 seconds.

39 seconds of that improvement is from #1214, which is linux only, but I don't think that windows machines are experiencing that slowdown.

This PR makes the improvements by making the following changes in the universe generation python code along with supporting changes in the C++ code:  
- It vectorizes `locations()` and `special_locations()`.  By searching all locations at the same time, it reduces round trip time between python and C++ and does all of the `WithinStarLaneJumps` calculations in parallel.
- It aggressively trims the candidate list for home systems.
- It rearranges the HONEYCOMB condition to do the expensive clause last.
- It uses known home systems to remove systems from consideration for monsters and natives.

While doing this two bugs were found, but only one was fixed.

In monsters and specials the systems were not randomized.  In galaxy topologies were system number is strongly correlated with location the monsters and specials ended up bunched in one area of the galaxy.  This is fixed in this PR.

In `specials.py` specials check their own clause against all existing specials before placement, but they do not recursively check all existing specials clauses against themselves.  Since, some special's clauses are written in a reciprocal way this is not a problem for all specials.  I could not think of a computationally efficient way to fix this so I made a TODO note and did not fix it.

### DIFF
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -94,19 +94,19 @@ class HomeSystemFinder(object):
 
     def find_home_systems_for_min_jump_distance(self, systems_pool, min_jumps):
         """
-        Return a good list of home systems
+        Return a good list of home systems or an empty list if there are fewer than num_home_systems in the pool.
 
-        Return a list of home systems which are at least the specified minimum number of jumps apart,
-        and with a good system merit picked randomly from the specified pool.
+        A good list of home systems are at least the specified minimum number of jumps apart,
+        with the best minimum system merit of all such lists picked randomly from the ''systems_pool''.
 
-        Return an empty list if there are fewer than num_home_systems in the pool.
+        Algorithm:
+        Make several attempts to find systems that match the condition
+        of being at least min_jumps apart.
+        Use the minimum merit of the best num_home_system systems found
+        as the merit of the current best set of systems.
+        On each attempt use the merit of the current best set of home
+        systems to truncate the pool of candidates.
         """
-        # Make several attempts to find systems that match the condition
-        # of being at least min_jumps apart
-        # Use the minimum merit of the best num_home_system systems found
-        # as the merit of the current best set of systems.
-        # On each attempt use the merit of the current best set of home
-        # systems to truncate the pool of candidates.
 
         # precalculate the system merits
         for system in systems_pool:
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -87,10 +87,11 @@ def min_planets_in_vicinity_limit(num_systems):
 
 class HomeSystemFinder(object):
     """Finds a set of home systems with a least ''num_home_systems'' systems."""
-    def __init__(self, _num_home_systems):
+    def __init__(self, _len_systems, _num_home_systems):
         # cache of sytem merits
         self.system_merit = {}
         self.num_home_systems = _num_home_systems
+        self.len_systems = _len_systems
 
     def find_home_systems_for_min_jump_distance(self, systems_pool, min_jumps):
         """
@@ -129,7 +130,7 @@ class HomeSystemFinder(object):
         # (999, 39) and (199, 19) the following was observered.  The distribution of candidate
         # length is a normal random variable with standard deviation approximately equal to
 
-        expected_len_candidate_std = (len(systems_pool) ** (1.0/3)) / 10.0 * 1.5
+        expected_len_candidate_std = (self.len_systems ** (1.0/2.0)) * 0.03
 
         # So if the mean of the length of the candidates is with 1/2 std of the target
         # num_home_systems then there is a 90% expectation of hitting the goal within 6 tries.
@@ -199,7 +200,7 @@ class HomeSystemFinder(object):
         return best_candidate
 
 
-def find_home_systems(num_home_systems, pool_list, jump_distance, min_jump_distance):
+def find_home_systems(len_systems, num_home_systems, pool_list, jump_distance, min_jump_distance):
     """
     Tries to find a specified number of home systems which are as far apart from each other as possible.
     Starts with the specified jump distance and reduces that limit until enough systems can be found or the
@@ -209,7 +210,7 @@ def find_home_systems(num_home_systems, pool_list, jump_distance, min_jump_dista
     of the pool for logging purposes as second element.
     """
 
-    finder = HomeSystemFinder(num_home_systems)
+    finder = HomeSystemFinder(len_systems, num_home_systems)
     # try to find home systems, decrease the min jumps until enough systems can be found, or the jump distance drops
     # below the specified minimum jump distance (which means failure)
     while jump_distance >= min_jump_distance:
@@ -384,7 +385,7 @@ def compile_home_system_list(num_home_systems, systems, gsd):
         # specify 0 as number of requested home systems to pick as much systems as possible
         (pool_matching_sys_limit, "pool of systems that meet at least the min systems limit"),
     ]
-    home_systems = find_home_systems(num_home_systems, pool_list, min_jumps, HS_MIN_DISTANCE_PRIORITY_LIMIT)
+    home_systems = find_home_systems(len(systems), num_home_systems, pool_list, min_jumps, HS_MIN_DISTANCE_PRIORITY_LIMIT)
 
     # check if the first attempt delivered a list with enough home systems
     # if not, we make our second attempt, this time disregarding the filtered pools and using all systems, starting
@@ -392,7 +393,7 @@ def compile_home_system_list(num_home_systems, systems, gsd):
     # systems as possible
     if len(home_systems) < num_home_systems:
         print "Second attempt: trying to pick home systems from all systems"
-        home_systems = find_home_systems(num_home_systems, [(systems, "complete pool")], min_jumps, 1)
+        home_systems = find_home_systems(len(systems), num_home_systems, [(systems, "complete pool")], min_jumps, 1)
 
     # check if the selection process delivered a list with enough home systems
     # if not, our galaxy obviously is too crowded, report an error and return an empty list
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -3,7 +3,7 @@ import random
 
 import freeorion as fo
 
-from galaxy import get_systems_within_jumps
+from galaxy_topology import get_systems_within_jumps
 from starsystems import star_types_real, pick_star_type
 from planets import calc_planet_size, calc_planet_type, planet_sizes_real, planet_types_real
 from names import get_name_list, random_name
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -87,11 +87,10 @@ def min_planets_in_vicinity_limit(num_systems):
 
 class HomeSystemFinder(object):
     """Finds a set of home systems with a least ''num_home_systems'' systems."""
-    def __init__(self, _len_systems, _num_home_systems):
+    def __init__(self, _num_home_systems):
         # cache of sytem merits
         self.system_merit = {}
         self.num_home_systems = _num_home_systems
-        self.len_systems = _len_systems
 
     def find_home_systems_for_min_jump_distance(self, systems_pool, min_jumps):
         """
@@ -130,13 +129,13 @@ class HomeSystemFinder(object):
         # (999, 39) and (199, 19) the following was observered.  The distribution of candidate
         # length is a normal random variable with standard deviation approximately equal to
 
-        # expected_len_candidate_std = (self.len_systems ** (1.0/2.0)) * 0.03
+        # expected_len_candidate_std = (len(systems) ** (1.0/2.0)) * 0.03
 
         # which is about 1 for 1000 systems.  It is likely that anylen(candidate) is within 1
         # standard deviation of the expected len(candidate)
 
-        # If we are within the miss_threshold of the target then try up to num_complete_misses more times.
-        miss_threshold = 3
+        # If we are within the MISS_THRESHOLD of the target then try up to num_complete_misses more times.
+        MISS_THRESHOLD = 3
         num_complete_misses_remaining = 4
 
         # Cap the number of attempts to the smaller of the number of systems in the pool, or 100
@@ -166,7 +165,7 @@ class HomeSystemFinder(object):
                 local_pool -= set(get_systems_within_jumps(member, min_jumps))
 
             # Count complete misses when number of candidates is not close to the target.
-            if len(candidate) < (self.num_home_systems - miss_threshold):
+            if len(candidate) < (self.num_home_systems - MISS_THRESHOLD):
                 num_complete_misses_remaining -= 1
 
             if len(candidate) < self.num_home_systems:
@@ -193,7 +192,7 @@ class HomeSystemFinder(object):
         return best_candidate
 
 
-def find_home_systems(len_systems, num_home_systems, pool_list, jump_distance, min_jump_distance):
+def find_home_systems(num_home_systems, pool_list, jump_distance, min_jump_distance):
     """
     Tries to find a specified number of home systems which are as far apart from each other as possible.
     Starts with the specified jump distance and reduces that limit until enough systems can be found or the
@@ -203,7 +202,7 @@ def find_home_systems(len_systems, num_home_systems, pool_list, jump_distance, m
     of the pool for logging purposes as second element.
     """
 
-    finder = HomeSystemFinder(len_systems, num_home_systems)
+    finder = HomeSystemFinder(num_home_systems)
     # try to find home systems, decrease the min jumps until enough systems can be found, or the jump distance drops
     # below the specified minimum jump distance (which means failure)
     while jump_distance >= min_jump_distance:
@@ -378,7 +377,7 @@ def compile_home_system_list(num_home_systems, systems, gsd):
         # specify 0 as number of requested home systems to pick as much systems as possible
         (pool_matching_sys_limit, "pool of systems that meet at least the min systems limit"),
     ]
-    home_systems = find_home_systems(len(systems), num_home_systems, pool_list, min_jumps, HS_MIN_DISTANCE_PRIORITY_LIMIT)
+    home_systems = find_home_systems(num_home_systems, pool_list, min_jumps, HS_MIN_DISTANCE_PRIORITY_LIMIT)
 
     # check if the first attempt delivered a list with enough home systems
     # if not, we make our second attempt, this time disregarding the filtered pools and using all systems, starting
@@ -386,7 +385,7 @@ def compile_home_system_list(num_home_systems, systems, gsd):
     # systems as possible
     if len(home_systems) < num_home_systems:
         print "Second attempt: trying to pick home systems from all systems"
-        home_systems = find_home_systems(len(systems), num_home_systems, [(systems, "complete pool")], min_jumps, 1)
+        home_systems = find_home_systems(num_home_systems, [(systems, "complete pool")], min_jumps, 1)
 
     # check if the selection process delivered a list with enough home systems
     # if not, our galaxy obviously is too crowded, report an error and return an empty list
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -135,6 +135,11 @@ class HomeSystemFinder(object):
             local_pool = {s for (m, s) in all_merit_system}
 
             if len(local_pool) < self.num_home_systems:
+                if len(best_candidate) < self.num_home_systems:
+                    print ("Failing in find_home_systems_for_min_jump_distance because "
+                           "current_merit_lower_bound = {} trims local pool to {} systems "
+                           "which is less than num_home_systems {}.".format(
+                               current_merit_lower_bound, len(local_pool), self.num_home_systems))
                 break
 
             attempts = min(attempts - 1, len(local_pool))
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -119,7 +119,6 @@ class HomeSystemFinder(object):
         # can be achieved.
         all_merit_system = sorted([(self.system_merit[s], s)
                                    for s in systems_pool], reverse=True)
-        (best_case_merit_lower_bound, _) = all_merit_system[self.num_home_systems - 1]
         current_merit_lower_bound = 0
 
         best_candidate = []
@@ -171,10 +170,6 @@ class HomeSystemFinder(object):
 
             # Quit if the lowest merit planet meets the minimum threshold
             if merit >= min_planets_in_vicinity_limit(get_systems_within_jumps(system, HS_VICINITY_RANGE)):
-                break
-
-            # quit if it isn't possible to improve the current accepted list
-            if merit >= best_case_merit_lower_bound:
                 break
 
             # If we have a better candidate, set the new lower bound one higher to try for a better candidate.
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -156,7 +156,8 @@ class HomeSystemFinder(object):
             # quit if it isn't possible to improve the current accepted list
             if merit >= best_case_merit_lower_bound:
                 break
-            attempts -= 1
+
+            attempts = min(attempts - 1, len(local_pool))
         return best_candidate
 
 
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -113,18 +113,24 @@ class HomeSystemFinder(object):
             if system not in self.system_merit:
                 self.system_merit[system] = calculate_home_system_merit(system)
 
-        best_case_worst_merit = sorted(self.system_merit.values())[self.num_home_systems - 1]
-        current_worst_merit = 0
+        # In the list of systems sorted by merit, if the first num_home_systems are all min_jumps
+        # apart, then the merit of the num_home_systems nth system is the
+        # best_case_merit_lower_bound that can be achieved.
+        best_case_merit_lower_bound = sorted(self.system_merit.values())[self.num_home_systems - 1]
+        current_merit_lower_bound = 0
 
         best_candidate = []
 
         # cap the number of attempts to the smaller of the number of systems in the pool, or 100
         attempts = min(100, len(systems_pool))
 
+        # Copy systems_pool to avoid aliasing of the caller's systems_pool variable
+        local_pool = set(systems_pool)
+
         while attempts:
             candidate = []
             # use a local pool of all candidate systems better than the worst threshold merit
-            local_pool = {s for s in systems_pool if self.system_merit[s] > current_worst_merit}
+            local_pool = {s for s in local_pool if self.system_merit[s] > current_merit_lower_bound}
             if len(local_pool) < self.num_home_systems:
                 break
 
@@ -133,24 +139,22 @@ class HomeSystemFinder(object):
                 candidate.append(member)
 
                 # remove all neighbors from the local pool
-                for neighbor in get_systems_within_jumps(member, min_jumps):
-                    if neighbor in local_pool:
-                        local_pool.remove(neighbor)
+                local_pool -= set(get_systems_within_jumps(member, min_jumps))
+
+            if len(candidate) < self.num_home_systems:
+                break
 
             # Calculate the merit of the current attempt.  If it is the best so far
             # keep it and update the merit_threshold
-            if len(candidate) >= self.num_home_systems:
-                merit_system = sorted([(self.system_merit[s], s)
-                                       for s in candidate])[:self.num_home_systems]
-                merit = merit_system[self.num_home_systems - 1]
-                if merit > current_worst_merit:
-                    current_worst_merit = merit
-                    best_candidate = [s for (_, s) in merit_system]
+            merit_system = sorted([(self.system_merit[s], s)
+                                   for s in candidate])[:self.num_home_systems]
+            merit = merit_system[-1]
+            if merit > current_merit_lower_bound:
+                current_merit_lower_bound = merit
+                best_candidate = [s for (_, s) in merit_system]
 
-                # quit if it isn't possible to improve the current accepted list
-                if merit >= best_case_worst_merit:
-                    break
-            else:
+            # quit if it isn't possible to improve the current accepted list
+            if merit >= best_case_merit_lower_bound:
                 break
             attempts -= 1
         return best_candidate
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -1,5 +1,6 @@
 import os.path
 import random
+import statistics
 
 import freeorion as fo
 
@@ -12,7 +13,6 @@ from options import (HS_ACCEPTABLE_PLANET_TYPES, HS_MIN_PLANETS_IN_VICINITY_TOTA
                      HS_VICINITY_RANGE, HS_MIN_SYSTEMS_IN_VICINITY, HS_ACCEPTABLE_PLANET_SIZES)
 
 from util import load_string_list, report_error
-import statistics
 
 
 def get_empire_name_generator():
@@ -321,7 +321,8 @@ def compile_home_system_list(num_home_systems, systems, gsd):
             pool_matching_sys_limit.append(system)
             if count_planets_in_systems(systems_in_vicinity) >= min_planets_in_vicinity_limit(len(systems_in_vicinity)):
                 pool_matching_sys_and_planet_limit.append(system)
-    print len(pool_matching_sys_and_planet_limit), "systems meet the min systems and planets in the near vicinity limit"
+    print (len(pool_matching_sys_and_planet_limit),
+           "systems meet the min systems and planets in the near vicinity limit")
     print len(pool_matching_sys_limit), "systems meet the min systems in the near vicinity limit"
 
     # now try to pick the requested number of home systems
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -157,11 +157,9 @@ class HomeSystemFinder(object):
             # Calculate the merit of the current attempt.  If it is the best so far
             # keep it and update the merit_threshold
             merit_system = sorted([(self.system_merit[s], s)
-                                   for s in candidate])[:self.num_home_systems]
-            merit = merit_system[-1]
-            if merit > current_merit_lower_bound:
-                current_merit_lower_bound = merit
-                best_candidate = [s for (_, s) in merit_system]
+                                   for s in candidate], reverse=True)[:self.num_home_systems]
+
+            (merit, system) = merit_system[-1]
 
             # quit if it isn't possible to improve the current accepted list
             if merit >= best_case_merit_lower_bound:
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -126,12 +126,27 @@ class HomeSystemFinder(object):
         # cap the number of attempts to the smaller of the number of systems in the pool, or 100
         attempts = min(100, len(systems_pool))
 
-        # Cap the number of attempts when the found number of systems is much less than the target
+        # Cap the number of attempts when the found number of systems is less than the target
         # num_home_systems because this indicates that the min_jumps is too large and/or the
         # systems_pool is too small to ever succeed.
-        num_complete_misses_remaining = 4
 
-        while attempts and num_complete_misses_remaining:
+        # From experimentation with cluster and 3 arm spiral galaxies, with low, med and high
+        # starlane density and (number of systems, number of home systems) pairs of (9999, 399),
+        # (999, 39) and (199, 19) the following was observered.  The distribution of candidate
+        # length is a normal random variable with standard deviation approximately equal to
+
+        expected_len_candidate_std = (len(systems_pool) ** (1.0/3)) / 10.0 * 1.5
+
+        # So if the mean of the length of the candidates is with 1/2 std of the target
+        # num_home_systems then there is a 90% expectation of hitting the goal within 6 tries.
+        # Otherwise quit and try an easier problem.
+        lowest_continuable_mean_len_of_candidates = (self.num_home_systems - 0.5 * expected_len_candidate_std)
+        print ("Len systems pool {}. Threshold STD is {}. Low threshold {}".format(
+            len(systems_pool), expected_len_candidate_std, lowest_continuable_mean_len_of_candidates))
+
+        lens_of_failed_candidates = []
+
+        while attempts:
             # use a local pool of all candidate systems better than the worst threshold merit
             all_merit_system = [(m, s) for (m, s) in all_merit_system if m > current_merit_lower_bound]
             local_pool = {s for (m, s) in all_merit_system}
@@ -154,11 +169,15 @@ class HomeSystemFinder(object):
                 # remove all neighbors from the local pool
                 local_pool -= set(get_systems_within_jumps(member, min_jumps))
 
-            # Count complete misses when number of candidates is not close to the target.
-            if len(candidate) < (self.num_home_systems - 4):
-                num_complete_misses_remaining -= 1
-
             if len(candidate) < self.num_home_systems:
+
+                # Break if the normal distribution of failed candidates is too much smaller than
+                # the target number of systems
+                lens_of_failed_candidates.append(len(candidate))
+                mean_len_of_failed_candidates = sum(lens_of_failed_candidates) / float(len(lens_of_failed_candidates))
+                if mean_len_of_failed_candidates < lowest_continuable_mean_len_of_candidates:
+                    break
+
                 continue
 
             # Calculate the merit of the current attempt.  If it is the best so far
@@ -174,6 +193,8 @@ class HomeSystemFinder(object):
                        "{}".format(current_merit_lower_bound, merit))
                 current_merit_lower_bound = merit
                 best_candidate = [s for (_, s) in merit_system]
+                lens_of_failed_candidates = []
+
                 # Quit sucessfully if the lowest merit planet meets the minimum threshold
                 if merit >= min_planets_in_vicinity_limit(get_systems_within_jumps(system, HS_VICINITY_RANGE)):
                     break
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -71,19 +71,9 @@ def count_planets_in_systems(systems, planet_types_filter=HS_ACCEPTABLE_PLANET_T
     return num_planets
 
 
-def systems_min_jumps_away_from(systems_pool, min_jumps, chosen_systems):
-    """
-    Generator that returns systems randomly picked from a pool that are at least a certain jump distance away
-    from the group systems listed in chosen_systems.
-    """
-    # on generator initialization shuffle our pool of systems to randomize the order in which the systems are returned
-    random.shuffle(systems_pool)
-    # return systems that meet our condition until we have exhausted our pool
-    for candidate in systems_pool:
-        # check if our candidate is at least the specified min jump distance away from the systems in chosen_systems
-        # if yes, return candidate, otherwise continue with next candidate
-        if all(fo.jump_distance(candidate, system) >= min_jumps for system in chosen_systems):
-            yield candidate
+def calculate_home_system_merit(system):
+    """Calculate the system's merit as the number of planets within HS_VICINTIY_RANGE."""
+    return count_planets_in_systems(get_systems_within_jumps(system, HS_VICINITY_RANGE))
 
 
 def min_planets_in_vicinity_limit(num_systems):
@@ -95,26 +85,75 @@ def min_planets_in_vicinity_limit(num_systems):
     return min(HS_MIN_PLANETS_IN_VICINITY_TOTAL, num_systems * HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM)
 
 
-def find_home_systems_for_min_jump_distance(systems_pool, min_jumps):
-    """
-    Returns a list of home systems which are at least the specified minimum number of jumps apart,
-    picked randomly from the specified pool.
-    """
-    # make several attempts to find systems that match the condition of being at least min_jumps apart
-    # set the number of attempts to the number of systems in the pool, or 100 at max
-    attempts = min(100, len(systems_pool))
-    # store the result of all attempts in a list, so we can pick the attempt that yielded the largest set of systems
-    accepted_list = []
-    while attempts:
-        accepted = []
-        # get candidates that are at least min_jumps away from the systems we already accepted
-        for candidate in systems_min_jumps_away_from(systems_pool, min_jumps, accepted):
-            accepted.append(candidate)  # add the candidate to our list of accepted systems
-        # add the result of the current attempt to our list and continue with next attempt
-        accepted_list.append(accepted)
-        attempts -= 1
-    # return the result of the attempt that yielded the most systems
-    return max(accepted_list, key=len)
+class HomeSystemFinder(object):
+    """Finds a set of home systems with a least ''num_home_systems'' systems."""
+    def __init__(self, _num_home_systems):
+        # cache of sytem merits
+        self.system_merit = {}
+        self.num_home_systems = _num_home_systems
+
+    def find_home_systems_for_min_jump_distance(self, systems_pool, min_jumps):
+        """
+        Return a good list of home systems
+
+        Return a list of home systems which are at least the specified minimum number of jumps apart,
+        and with a good system merit picked randomly from the specified pool.
+
+        Return an empty list if there are fewer than num_home_systems in the pool.
+        """
+        # Make several attempts to find systems that match the condition
+        # of being at least min_jumps apart
+        # Use the minimum merit of the best num_home_system systems found
+        # as the merit of the current best set of systems.
+        # On each attempt use the merit of the current best set of home
+        # systems to truncate the pool of candidates.
+
+        # precalculate the system merits
+        for system in systems_pool:
+            if system not in self.system_merit:
+                self.system_merit[system] = calculate_home_system_merit(system)
+
+        best_case_worst_merit = sorted(self.system_merit.values())[self.num_home_systems - 1]
+        current_worst_merit = 0
+
+        best_candidate = []
+
+        # cap the number of attempts to the smaller of the number of systems in the pool, or 100
+        attempts = min(100, len(systems_pool))
+
+        while attempts:
+            candidate = []
+            # use a local pool of all candidate systems better than the worst threshold merit
+            local_pool = {s for s in systems_pool if self.system_merit[s] > current_worst_merit}
+            if len(local_pool) < self.num_home_systems:
+                break
+
+            while local_pool:
+                member = random.choice(list(local_pool))
+                candidate.append(member)
+
+                # remove all neighbors from the local pool
+                for neighbor in get_systems_within_jumps(member, min_jumps):
+                    if neighbor in local_pool:
+                        local_pool.remove(neighbor)
+
+            # Calculate the merit of the current attempt.  If it is the best so far
+            # keep it and update the merit_threshold
+            if len(candidate) >= self.num_home_systems:
+                merit_system = sorted([(self.system_merit[s], s)
+                                       for s in candidate])[:self.num_home_systems]
+                merit = merit_system[self.num_home_systems - 1]
+                if merit > current_worst_merit:
+                    current_worst_merit = merit
+                    best_candidate = [s for (_, s) in merit_system]
+
+                # quit if it isn't possible to improve the current accepted list
+                if merit >= best_case_worst_merit:
+                    break
+            else:
+                break
+            attempts -= 1
+        return best_candidate
 
 
 def find_home_systems(num_home_systems, pool_list, jump_distance, min_jump_distance):
@@ -126,6 +165,8 @@ def find_home_systems(num_home_systems, pool_list, jump_distance, min_jump_dista
     This parameter contains a list of tuples, each tuple has a pool of systems as first element and a description
     of the pool for logging purposes as second element.
     """
+
+    finder = HomeSystemFinder(num_home_systems)
     # try to find home systems, decrease the min jumps until enough systems can be found, or the jump distance drops
     # below the specified minimum jump distance (which means failure)
     while jump_distance >= min_jump_distance:
@@ -142,7 +183,7 @@ def find_home_systems(num_home_systems, pool_list, jump_distance, min_jump_dista
                 continue
 
             # try to pick home systems
-            home_systems = find_home_systems_for_min_jump_distance(pool, jump_distance)
+            home_systems = finder.find_home_systems_for_min_jump_distance(pool, jump_distance)
             # check if we got enough
             if len(home_systems) >= num_home_systems:
                 # yes, we got what we need, return the home systems we found
@@ -320,8 +361,7 @@ def compile_home_system_list(num_home_systems, systems, gsd):
     if len(home_systems) > num_home_systems:
         # yes: calculate the number of planets in the near vicinity of each system
         # and store that value with each system in a map
-        hs_planets_in_vicinity_map = {s: count_planets_in_systems(get_systems_within_jumps(s, HS_VICINITY_RANGE))
-                                      for s in home_systems}
+        hs_planets_in_vicinity_map = {s: calculate_home_system_merit(s) for s in home_systems}
         # sort the home systems by the number of planets in their near vicinity using the map
         # now only pick the number of home systems we need, taking those with the highest number of planets
         home_systems = sorted(home_systems, key=hs_planets_in_vicinity_map.get, reverse=True)[:num_home_systems]
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -168,16 +168,15 @@ class HomeSystemFinder(object):
 
             (merit, system) = merit_system[-1]
 
-            # Quit if the lowest merit planet meets the minimum threshold
-            if merit >= min_planets_in_vicinity_limit(get_systems_within_jumps(system, HS_VICINITY_RANGE)):
-                break
-
             # If we have a better candidate, set the new lower bound and try for a better candidate.
             if merit >= current_merit_lower_bound:
                 print ("Home system merit lower bound improved from {} to "
                        "{}".format(current_merit_lower_bound, merit))
                 current_merit_lower_bound = merit
                 best_candidate = [s for (_, s) in merit_system]
+                # Quit sucessfully if the lowest merit planet meets the minimum threshold
+                if merit >= min_planets_in_vicinity_limit(get_systems_within_jumps(system, HS_VICINITY_RANGE)):
+                    break
 
         return best_candidate
 
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -1,6 +1,5 @@
 import os.path
 import random
-import statistics
 
 import freeorion as fo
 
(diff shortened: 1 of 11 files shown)


default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,75 @@ def min_planets_in_vicinity_limit(num_systems):
...
+                break
+
+            while local_pool:
+                member = random.choice(list(local_pool))
+                candidate.append(member)
+
+                # remove all neighbors from the local pool
+                for neighbor in get_systems_within_jumps(member, min_jumps):
+                    if neighbor in local_pool:
+                        local_pool.remove(neighbor)
```
Morlic-fo: Use python's set-functions for that:
```local_pool -= set(get_systems_within_jumps(member, min_jumps))``` or
```local_pool.difference_update(get_systems_within_jumps(member, min_jumps))```

https://docs.python.org/2/library/sets.html

File: default/python/universe_generation/monsters.py [hunk 2]
```diff
@@ -131,7 +132,8 @@ def generate_monsters(monster_freq, _systems, home_systems):
     systems = [s for s in _systems if s not in empire_exclusions]
 
     # Fleet plans that include ships capable of altering starlanes.
-    ## @content_tag{CAN_ALTER_STARLANES} universe_generator special handling for fleets containing a hull design with this tag.
+    # @content_tag{CAN_ALTER_STARLANES} universe_generator special handling for
+    # fleets containing a hull design with this tag.
```
dbenage-cx: For special command processing, doxygen requires `##`. [doxygen pythonblocks](http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#pythonblocks) / [cppapi doc for content_tag alias](http://freeorion.github.io/cppapi/master/content_tags_alias.html)

The param/content for `content_tag` alias is limited to the first line break.

default/python/universe_generation/specials.py
```diff
@@ -1,77 +1,178 @@
...
+from timers import Timer
+from galaxy_topology import get_systems_within_jumps
+
+REPEAT_RATE = {1: 0.08, 2: 0.05, 3: 0.01, 4: 0.00}
+
+def calc_num_placed(objs):
+    """Return a list of number placed at each obj"""
+    return [1 if random.random() > REPEAT_RATE[1] else
+            2 if random.random() > REPEAT_RATE[2] else
+            3 if random.random() > REPEAT_RATE[3] else 1
```
Morlic-fo: Seems like the last else case should yield 4 and not 1?

default/python/universe_generation/specials.py
```diff
@@ -1,77 +1,178 @@
...
 import freeorion as fo
 import statistics
 import universe_tables
+from timers import Timer
+from galaxy_topology import get_systems_within_jumps
+
+REPEAT_RATE = {1: 0.08, 2: 0.05, 3: 0.01, 4: 0.00}
+
+def calc_num_placed(objs):
+    """Return a list of number placed at each obj"""
```
Morlic-fo: number of what?

default/python/universe_generation/specials.py
```diff
@@ -1,77 +1,178 @@
...
+def place_special_fast(specials, obj, specials_timer, per_special_timer):
+    """ Place at least single special.
+    Return the number of specials placed.
+    """
+    num_placed = 0
+    # Calculate the conditional probabilities that each special is
+    # placed here given that a special will be placed here.
+    probs = [fo.special_spawn_rate(sp) for sp in specials]
+    total_prob = sum(probs)
+    thresholds = [x/total_prob for x in probs]
```
Morlic-fo: use explicit float conversion in division and make sure that you do not divide by zero.

default/python/universe_generation/specials.py
```diff
@@ -1,77 +1,178 @@
...
+        specials_timer.start("Place Special")
+        # All prerequisites and the test have been met, now add this special to this universe object.
+        fo.add_special(obj, special)
+        num_placed += 1
+        print "Special", special, "added to", fo.get_name(obj)
 
+        break
 
+    per_special_timer.start("None")
+    return num_placed
```
Morlic-fo: I am not sure what you are trying to achieve here.

It seems to me that this function will always place exactly one special - the probabilities are normalized to 1 so chance wil eventually be smaller than some threshold. After placing one special, you break out of the loop.

So num_placed will always be 1, making it both a pointless variable and return value. Docstring specifies the functio should be able to place more than 1 special. Am I missing something here?

Morlic-fo (changes requested):
The special refactoring does not look right to me. Will have to look into it in more detail though.

default/python/universe_generation/specials.py
```diff
@@ -1,76 +1,179 @@
...
+
+import freeorion as fo
 import universe_tables
+from timers import Timer
+from galaxy_topology import get_systems_within_jumps
+
+REPEAT_RATE = {1: 0.08, 2: 0.05, 3: 0.01, 4: 0.00}
+
+
+def calc_num_placed(objs):
```
Cjkjvfnby: `generate_number_for_placing`

default/python/universe_generation/specials.py
```diff
@@ -1,76 +1,179 @@
...
+    """Return a list of number placed at each obj"""
+    return [1 if random.random() > REPEAT_RATE[1] else
+            2 if random.random() > REPEAT_RATE[2] else
+            3 if random.random() > REPEAT_RATE[3] else 1
+            for __ in objs]
+
+
+def place_special_fast(specials, obj, specials_timer, per_special_timer):
```
Cjkjvfnby: Do we have slow? Otherwise just `place_special`

default/python/universe_generation/specials.py
```diff
@@ -1,76 +1,179 @@
...
+    """ Place at least single special.
```
Cjkjvfnby: According to code this function will return 0 or 1.

default/python/universe_generation/specials.py
```diff
@@ -1,76 +1,179 @@
...
+    # appart.  Place a special on each one.  Repeat until you run out of specials or objects.
+    GALAXY_DECOUPLING_DISTANCE = 6
+
+    i_cache = 0
+    while obj_tuple_needing_specials:
+        systems_needing_specials = defaultdict(set)
+        for (obj, sys, n) in obj_tuple_needing_specials:
+            systems_needing_specials[sys].add((obj, sys, n))
+
+        print " Placing in " + str(len(systems_needing_specials)) + " locations remaining"
```
Cjkjvfnby: Use string formatting.  PS. If you use PyCharm apply quick fix (alt+enter)

default/python/universe_generation/specials.py
```diff
@@ -1,76 +1,179 @@
...
+        # Find a list of candidates all spaced GALAXY_DECORRECLATION_DISTANCE apart
+        specials_timer.start("Compose spaced list")
+        candidates = []
+        while systems_needing_specials:
+            specials_timer.start("Randos")
+            random_sys = random.choice(systems_needing_specials.values())
+            member = random.choice(list(random_sys))
+            candidates.append(member[0])
+            obj_tuple_needing_specials.remove(member)
+            if member[2] > 1:
```
Cjkjvfnby: Unpack tuple to variables `objects_need_special, systems_for_object, number_to_place = memeber`

Cjkjvfnby (commented):
I make a quick look.  I will not add `Request changes` since does not know if I will have time to check this RP again.

default/python/universe_generation/specials.py
LGM-Doyle: I used format with the string.

default/python/universe_generation/specials.py
LGM-Doyle: I unpacked the tuple.

LGM-Doyle (commented):
Thanks for the thorough review.

default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,80 @@ def min_planets_in_vicinity_limit(num_systems):
...
+
+        # precalculate the system merits
+        for system in systems_pool:
+            if system not in self.system_merit:
+                self.system_merit[system] = calculate_home_system_merit(system)
+
+        # In the list of systems sorted by merit, if the first num_home_systems are all min_jumps
+        # apart, then the merit of the num_home_systems nth system is the
+        # best_case_merit_lower_bound that can be achieved.
+        best_case_merit_lower_bound = sorted(self.system_merit.values())[self.num_home_systems - 1]
```
Vezzra: Correct me if I'm wrong, but what you *want* is the merit of the system with the lowest merit out of the `num_home_system` systems with the highest merits, correct? Which makes sense considering the description of the algorithm in the docstring, the name of the variable and the actual implementation.

However, unless I'm completely confused for whatever reason, what this line actually gives is the merit of the system with the highest merit out of the `num_home_system` systems with the lowest merits (because by default `sorted` sorts in ascending order). Which makes no sense to me considering the description and implementation of the algorithm.

To get what you want you'd either have to sort in reversed order, or access index `[-self.num_home_systems]` (if I got that right).

default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,80 @@ def min_planets_in_vicinity_limit(num_systems):
...
+
+            while local_pool:
+                member = random.choice(list(local_pool))
+                candidate.append(member)
+
+                # remove all neighbors from the local pool
+                local_pool -= set(get_systems_within_jumps(member, min_jumps))
+
+            if len(candidate) < self.num_home_systems:
+                break
```
Vezzra: Why breaking out of the loop here, instead of decreasing attempts and `continue`? The reason for coming up with too few candidates might have been a very unlucky selection, so subsequent attempts might be successful.

default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,80 @@ def min_planets_in_vicinity_limit(num_systems):
...
+
+            # Calculate the merit of the current attempt.  If it is the best so far
+            # keep it and update the merit_threshold
+            merit_system = sorted([(self.system_merit[s], s)
+                                   for s in candidate])[:self.num_home_systems]
+            merit = merit_system[-1]
```
Vezzra: Again, correct me if I'm wrong, but what you *want* is the merit of the system with the *lowest* merit out of the final selection of candidate systems, right? At least, that is what I'd expect considering the descriptions in the comments and the implementation of the algorithm.

I see two issues here:

1. Doesn't `merit = merit_system[-1]` yield the *tuple* at index `[-1]`? But what you need is the actual merit value, isn't it?

1. As that list of tuples is sorted in *ascending* order, wouldn't the tuple at index `[-1]` be the one with the *highest* merit value?

default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,80 @@ def min_planets_in_vicinity_limit(num_systems):
...
+            if merit > current_merit_lower_bound:
+                current_merit_lower_bound = merit
+                best_candidate = [s for (_, s) in merit_system]
+
+            # quit if it isn't possible to improve the current accepted list
+            if merit >= best_case_merit_lower_bound:
+                break
+
+            attempts = min(attempts - 1, len(local_pool))
```
Vezzra: Why capping `attempts` at the size of the at this point reduced `local_pool` here? I mean, I can see how that will speed things up, but doesn't this cut down the number of attempts a bit to aggressively?

default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,80 @@ def min_planets_in_vicinity_limit(num_systems):
...
-        accepted = []
-        # get candidates that are at least min_jumps away from the systems we already accepted
-        for candidate in systems_min_jumps_away_from(systems_pool, min_jumps, accepted):
-            accepted.append(candidate)  # add the candidate to our list of accepted systems
-        # add the result of the current attempt to our list and continue with next attempt
-        accepted_list.append(accepted)
-        attempts -= 1
-    # return the result of the attempt that yielded the most systems
-    return max(accepted_list, key=len)
+class HomeSystemFinder(object):
```
Vezzra, line 88: Nothing important, just a style issue and anyway probably just a preference of mine, but I usually group constants, global variables, classes and global functions together (in this order). So, that class should be above all the global functions.

Unless you have a preference for keeping it here, close by the function that uses it.

default/python/universe_generation/galaxy.py
```diff
@@ -379,27 +380,6 @@ def calc_universe_width(shape, size):
...
-    next_origin_systems = [origin_system]
-    while jumps > 0:
-        origin_systems = list(next_origin_systems)
-        next_origin_systems = []
-        for system in origin_systems:
-            neighbor_systems = [s for s in fo.sys_get_starlanes(system) if s not in matching_systems]
-            next_origin_systems.extend(neighbor_systems)
-            matching_systems.extend(neighbor_systems)
-        jumps -= 1
-    return matching_systems
```
Vezzra, line 400: Moving that function out of the `galaxy` module has been proposed before, and while I can see the point, I wonder if it's really reasonable to create an extra module just for one function. The `galaxy` module has been named "galaxy" for this reason, because it originally was supposed to contain all kind of map related stuff, and not only the actual system placement functions.

Once we have more strictly topology related stuff, splitting that from `galaxy` makes more sense IMO.

Vezzra (changes requested):
You seem to have a soft spot for big PRs, don't you? 😉 That biggie could probably have been broken up into a couple of smaller PRs.

Well, this way I can't review the entire thing at once, so I'll have to break that up into several smaller reviews. Can't do all of it today, so more to come.

This first one deals with the changes to `empires.py` and splitting `get_systems_within_jumps` from `galaxy.py` to its own module.

default/python/universe_generation/empires.py
```diff
@@ -113,18 +113,24 @@ def find_home_systems_for_min_jump_distance(self, systems_pool, min_jumps):
             if system not in self.system_merit:
                 self.system_merit[system] = calculate_home_system_merit(system)
 
-        best_case_worst_merit = sorted(self.system_merit.values())[self.num_home_systems - 1]
-        current_worst_merit = 0
+        # In the list of systems sorted by merit, if the first num_home_systems are all min_jumps
+        # apart, then the merit of the num_home_systems nth system is the
+        # best_case_merit_lower_bound that can be achieved.
+        best_case_merit_lower_bound = sorted(self.system_merit.values())[self.num_home_systems - 1]
```
Morlic-fo: If I am not missing something, then this is not some "lower bound" but by your definition of the merit the maximum achievable value, i.e. the *upper bound* for the achievable merit.

default/python/universe_generation/specials.py
```diff
@@ -121,7 +121,7 @@ def distribute_specials(specials_freq, universe_objects):
         for (obj, sys, n) in obj_tuple_needing_specials:
             systems_needing_specials[sys].add((obj, sys, n))
 
-        print " Placing in " + str(len(systems_needing_specials)) + " locations remaining"
+        print " Placing in {} locations remaining.".format(str(len(systems_needing_specials)))
```
Morlic-fo: str() call seems redundant.

default/python/universe_generation/specials.py
```diff
@@ -1,76 +1,188 @@
...
+# The probabilites break down as follows:
+# Count  Probability
+# one    (1 - REPEAT_RATE[0])
+# two    REPEAT_RATE[0] * (1 - REPEAT_RATE[1])
+# three  REPEAT_RATE[0] * REPEAT_RATE[1] *(1 - REPEAT_RATE[2])
+# four   REPEAT_RATE[0] * REPEAT_RATE[1] * REPEAT_RATE[2]
+REPEAT_RATE = {1: 0.08, 2: 0.05, 3: 0.01, 4: 0.00}
+
+
+def calc_num_placed(objs):
```
Morlic-fo: Maybe "calculate_number_of_specials_to_place"?

default/python/universe_generation/specials.py
```diff
@@ -1,76 +1,188 @@
...
+    """Return a list of number placed at each obj"""
+    return [1 if random.random() > REPEAT_RATE[1] else
+            2 if random.random() > REPEAT_RATE[2] else
+            3 if random.random() > REPEAT_RATE[3] else 4
+            for __ in objs]
+
+
+def place_special(specials, obj, specials_timer, per_special_timer):
+    """ Place at most a single special.
```
Morlic-fo: I can't see when you would place 0 specials.

After your normalization, the sum of thresholds is exactly 1. random.random() returns a number [0, 1). 
If no special was previously placed, the last if check performed is 
```chance - (sum(thresholds)-threshold[end])  = chance - (1-threshold[end]) > threshold[end] <--> chance > 1```
which is false in any case.
Therefore, you will *always* place a special with the given code (unless my math failed me somewhere).



So, the function just decides which special is placed given the probability that each special has not *if* it places a special. As long as that is the intention, the math seems correct (but docstring is wrong).

File: default/python/universe_generation/empires.py (hunk 12 above)
Comment (Morlic-fo) [reply default/python/universe_generation/empires.py]: > To get what you want you'd either have to sort in reversed order, or access index [-self.num_home_systems] (if I got that right).

Sorting in descending (i.e. reversed order) seems more clear.

File: default/python/universe_generation/specials.py [hunk 22]
```diff
@@ -122,14 +130,15 @@ def distribute_specials(specials_freq, universe_objects):
             specials_timer.start("Randos")
             random_sys = random.choice(systems_needing_specials.values())
             member = random.choice(list(random_sys))
-            candidates.append(member[0])
+            (obj, sys, n) = member
```
Cjkjvfnby: `(obj, sys, n)` - > `obj, sys, n`  no need parenthesis here
`sys` shadows python builtin module, please rename to `systems`. 
`n` -> `specials_count`

default/python/universe_generation/specials.py
```diff
@@ -1,76 +1,188 @@
...
+            2 if random.random() > REPEAT_RATE[2] else
+            3 if random.random() > REPEAT_RATE[3] else 4
+            for __ in objs]
+
+
+def place_special(specials, obj, specials_timer, per_special_timer):
+    """ Place at most a single special.
+    Return the number of specials placed.
+    """
+    num_placed = 0
```
Cjkjvfnby: Remove it and just use return 1|0. (`break` -> `return 1`)

default/python/universe_generation/empires.py
Morlic-fo: Isn't ```len(local_pool)``` equivalent to zero here? Otherwise, we wouldn't have left the while loop above?

default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,80 @@ def min_planets_in_vicinity_limit(num_systems):
...
+        # cap the number of attempts to the smaller of the number of systems in the pool, or 100
+        attempts = min(100, len(systems_pool))
+
+        # Copy systems_pool to avoid aliasing of the caller's systems_pool variable
+        local_pool = set(systems_pool)
+
+        while attempts:
+            candidate = []
+            # use a local pool of all candidate systems better than the worst threshold merit
+            local_pool = {s for s in local_pool if self.system_merit[s] > current_merit_lower_bound}
```
Morlic-fo: This is an empty set after first iteration.

default/python/universe_generation/empires.py
Vezzra: Well, I guess it's a bit difficult to come up with a good proper name for this variable. Because the value assigned to it is actually the merit of the system with the *lowest* merit value out of the set of systems with the highest merit values.

And you need to somehow make that clear in the variable name, so just changing it to `best_case_merit_upper_bound` won't cut it too...

default/python/universe_generation/empires.py
Vezzra: Yep, pretty much.

Anyway, what we want here is to create a pool from the original full set of systems minus those who are already disqualified because of their merit value. So we need to copy from `systems_pool`.

Which makes the `local_pool = set(systems_pool)` statement a few lines above obsolete.

default/python/universe_generation/empires.py
Morlic-fo: ```upper_bound``` is exactly what the variable is and would be perfectly fine to use as name. You use a random heuristic to find some value. About the found values we know that
```lower_bound <= found_value <= best_possible_value <= upper_bound```.  The best known upper bound is exactly what that variable is. The best lower bound we have is the currently found value...

**Review by LGM-Doyle — COMMENTED** [review]
Things I missed in the last pass that I've fixed.
1.  Explicitly prevent division by zero in place_special.
2. Improve place_special docstring
3. Remove unneeded str() in debug message.

File: default/python/universe_generation/empires.py (hunk 18 above)
Comment (Vezzra) [reply default/python/universe_generation/empires.py]: Then how to convey the (not unimportant) fact that this is the lowest value from a "best possible set" of values? The "best_case_merit" part IMO refers already to the value being the best theoretically achievable, the "_lower_bound" part to the value being the lowest one of the theoretical best set.

I'm not saying that there can't be a better name, but `best_case_merit_upper_bound` IMO doesn't tell me that it's the lowest value from a set...?

File: default/python/universe_generation/empires.py (hunk 16 above)
Comment (Vezzra) [reply default/python/universe_generation/empires.py:88]: Ok. No need to make it a nested class.

File: default/python/universe_generation/empires.py (hunk 15 above)
Comment (Vezzra) [reply default/python/universe_generation/empires.py]: My concern was that at this point `local_pool` has been reduced to just a few systems after all the removing of systems done before (I didn't realize that it actually would be empty). So that struck me as overly aggressive.

Moving the statement addressed my concerns, so I guess that's fine now.

File: default/python/universe_generation/galaxy.py (hunk 17 above)
Comment (Vezzra) [reply default/python/universe_generation/galaxy.py:400]: No problem, if you need that to avoid circular dependencies in other stuff you're working on, and the file is going to be removed anyway eventually, then of course I have no objections.

No need to make things more difficult for you. 😉

File: default/python/universe_generation/empires.py (hunk 18 above)
Comment (Morlic-fo) [reply default/python/universe_generation/empires.py]: > I'm not saying that there can't be a better name, but best_case_merit_upper_bound IMO doesn't tell me that it's the lowest value from a set...?

The comments and / or docstrings need to explain what are the criteria used to calculate the "merit" of the selected systems once in the very beginning when explaining the algorithm. 
In the PR it is stated as "Use the minimum merit of the best num_home_system systems found  as the merit of the current best set of systems."

So what is the merit of a selection of systems is already clearly stated. There is absolutely no need to state it anywhere again especially not in a variable name. If you want to avoid confusion between single-system-merit and set-of-systems-merit, then do not use "merit" twice but switch to "score"/"rating" or whatever else comes to mind so that it reads
"The score of a selection of systems is defined as the lowest merit of the systems within that selection."

If you then call it ```score_upper_bound```, there will be no ambiguity. If you want to explain how the upper bound is calculated, then that should be a single line comment at the location where it is calculated but not a ```4_underscore_variable_name_to_state_already_stated_facts_that_are_never_needed_again``` is unneccessary and very clumsy IMO.

File: default/python/universe_generation/empires.py [hunk 25]
```diff
@@ -95,26 +85,101 @@ def min_planets_in_vicinity_limit(num_systems):
...
+
+            # Quit if the lowest merit planet meets the minimum threshold
+            if merit >= min_planets_in_vicinity_limit(get_systems_within_jumps(system, HS_VICINITY_RANGE)):
+                break
+
+            # If we have a better candidate, set the new lower bound one higher to try for a better candidate.
+            if merit >= current_merit_lower_bound:
+                print ("Home system merit lower bound improved from {} to "
+                       "{}".format(current_merit_lower_bound, merit + 1))
+                current_merit_lower_bound = merit + 1
```
Morlic-fo: ```merit``` is the lower bound as you know such a combination exists. ```merit+1``` is an arbitrary threshold as you do not know if any combination yields such a value (or higher).

Please do not add the +1 here but change the comparison in the if-operator here to a stricter greater than. If you somehow want to go with the +1, then you absolutely must change the ```all_merit_systems``` fitering (currently in line 136) to a ```>=``` operator. But not adding the +1 here is more precise.

default/python/universe_generation/galaxy.py
```diff
@@ -297,7 +298,7 @@ def stitching_positions(p1, p2):
 
     # Pick a random point in an 2:1 ratio ellipse and then rotate and slide it in between p1 and p2
     p1_p2_theta = acos((p2[0] - p1[0]) / p1_p2_dist)
-    p1_p2_scale = (p1_p2_dist - 2*min_dist) / 2
+    p1_p2_scale = (p1_p2_dist - 2 * min_dist) / 2
```
Morlic-fo: When using whitespace in formulas, it is prefered to group by mathematical order of operations.

PEP 8:
```
Yes:
x = x*2 - 1
hypot2 = x*x + y*y
c = (a+b) * (a-b)
No:
x = x * 2 - 1
hypot2 = x * x + y * y
c = (a + b) * (a - b)
```
https://www.python.org/dev/peps/pep-0008/#other-recommendations

default/python/universe_generation/empires.py
Vezzra: Ok, that makes sense. I'd probably go with the use of "score"/"rating" to avoid the confusion, that approach sounds like the most reasonable to me.

Moot point though, as `best_case_merit_lower_bound` got thrown out in the meantime...

default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,101 @@ def min_planets_in_vicinity_limit(num_systems):
...
+            candidate = []
+            while local_pool:
+                member = random.choice(list(local_pool))
+                candidate.append(member)
+
+                # remove all neighbors from the local pool
+                local_pool -= set(get_systems_within_jumps(member, min_jumps))
+
+            # Count complete misses when number of candidates is not close to the target.
+            if len(candidate) < (self.num_home_systems - 4):
```
Vezzra: Ok, here the variance on the length of candidates being within a range of about 4 you mentioned comes in, correct? So we have a complete miss when we lack more than 4 systems to meet the `num_home_system` goal.

Question: Isn't that variance dependend on the number of home systems we try to find? I'd expect that variance to be greater if I'm looking for 40 than 4 home systems. While it probably doesn't make sense to come up with complex calculations to reduce the complete miss treshold when looking for fewer systems (I guess the performance gain with a reduced treshold for e.g. 10 home systems will be negligible), we need to account for the case that the current hard limit of 40 players might be increased in the future.

We've been raising those hard limits in the past, so I expect that to happen again in the future. Then the hardcoded treshold of 4 might become to narrow, and I expect no one is going to remember that there is some hard coded treshold somewhere deep in the universe generation code that needs according adjustment.

So better make it scale properly with the number of home systems requested.

default/python/universe_generation/empires.py
```diff
@@ -95,26 +85,101 @@ def min_planets_in_vicinity_limit(num_systems):
...
+            # Calculate the merit of the current attempt.  If it is the best so far
+            # keep it and update the merit_threshold
+            merit_system = sorted([(self.system_merit[s], s)
+                                   for s in candidate], reverse=True)[:self.num_home_systems]
+
+            (merit, system) = merit_system[-1]
+
+            # Quit if the lowest merit planet meets the minimum threshold
+            if merit >= min_planets_in_vicinity_limit(get_systems_within_jumps(system, HS_VICINITY_RANGE)):
+                break
```
Vezzra: Shouldn't that happen *after* checking if `merit` is better than `current_merit_lower_bound`, and `best_candidate` has been updated? Because doing that *here* will result in the candidate under consideration in the current iteration being discarded, regardless if it's better than the current `best_candidate` or not.

Which can, worst case, lead to quitting before *any* best candidate has been determined, when the candidate of the first iteration already meets this exit criteria. Resulting in returning an empty `best_candidate`, definitely not what we want. Unless I missed something obvious of course...

That check only makes sense if a new best candidate has been found anyway, doesn't it? Otherwise the loop would have already been quit when the last best candidate had been found.

Vezzra (changes requested):
Second pass on the new home system finder algorithm. Two issues remain. Otherwise looks good.

Will follow up with separate reviews for the changes to `monsters.py` and `specials.py`.

default/python/universe_generation/monsters.py
```diff
@@ -115,62 +118,60 @@ def generate_monsters(monster_freq, systems):
...
     if not fleet_plans:
         return
 
     universe = fo.get_universe()
 
+    EMPIRE_TO_MONSTER_MIN_DIST = 2
+    empire_exclusions = set(itertools.chain.from_iterable(
+        get_systems_within_jumps(e, EMPIRE_TO_MONSTER_MIN_DIST)
+        for e in home_systems))
+    systems = [s for s in _systems if s not in empire_exclusions]
```
Vezzra: I understand the desire to speed things up, but you can't do that here (or at least, it's not a good idea to do that here at all). Unlike natives, monsters do have a location condition (yes, that's an inconsistency!), and all criteria that determine where a monster can be spawned should be defined there.

That part here effectively adds another location restriction (to the ones defined in a monsters location condition), which means the rules where a monster can be placed becomes scattered among very different files/locations. To give an example how this can screw up what a content scripter might want to do:

Imagine someone comes up with the idea to script some kind of "invisible helper monster" that can appear close to player starting homeworlds, and that secretly aids the player (by applying some boosts or whatever). The above code would prevent that monster from ever spawning, with no hint whatsoever for the content scripter to figure out why their script doesn't work as intended.

The point being, you can't make *any* assumptions about space monsters and where they should/shouldn't spawn here. Space monsters can and are used for all kind of things, not just the obvious nasty critters that are supposed to make life hard for players. 😉 

By the way an excellent example why I want to strip those location conditions from monsters and specials and integrate them into the universe generation scripts. Among other things, to get rid of inconsistencies, to allow for optimizations like you want to do here without making life hard for content scripters, etc.

default/python/universe_generation/monsters.py
```diff
@@ -115,62 +118,60 @@ def generate_monsters(monster_freq, systems):
...
+            if system in fp_location_cache[fp]:
                 tracked_plan_valid_locations[fp] += 1
 
         # filter out all monster fleets whose location condition allows this system and whose counter hasn't reached 0.
-        # Note: The WithinStarlaneJumps condition in fp.location() is
-        # the most time costly function in universe generation.
         suitable_fleet_plans = [fp for fp in fleet_plans
-                                if spawn_limits[fp]
-                                and fp.location(system)
+                                if system in fp_location_cache[fp]
```
Vezzra, line 164: I'm curious: You correctly identified the bug in `specials.py`, that by placing a special the location condition of a special that has been placed before could have been invalidated, so the location conditions of all already placed specials need to be re-checked, which isn't done.

The same actually applies to space monsters. So why would you *pre-calculate* their location conditions, which makes the current bugged distribution algorithm even worse, as now placement of a monster at a location that might have become invalidated by already placed monsters becomes possible? Isn't that a step in the wrong direction?

Your discovery of that bug made me realize how botched the current implementations are, and how complex a correct solution will be that actually works. IMO we need to redesign the entire algorithm from scratch. The problem goes even further than just monsters invalidating monsters and specials invalidating specials: *everything* that gets placed during universe generation potentially invalidates the location condition of *everything* already placed that has a location condition.

And even re-checking all location conditions of all things that have one after everything has been placed doesn't solve anything, because what to do if a re-checked location condition fails? There are several ways to solve the conflict, determining the optimal one will be non-trivial. And just weeding out everything that causes a conflict will probably be to aggressive, and lead to maps that have far less monsters and specials than the selected game settings are supposed to provide.

The fundamental issue is the enormous flexibility of the location conditions. We have a set of items with location conditions, where placement of each item *is affected by and affects* all items already placed. A post-placement step which tries to automatically resolve conflicts sounds awfully complicated, and I don't want to think of the possible performance hit.

I'm really scratching my head here what to do. A thorough solution will take considerable effort, and certainly won't be something we can design and implement quickly.

A temporary solution might be to actually go with your approach here and calculate the locations conditions for all monsters before any monster is placed, and all specials before any special gets placed, and make it a rule that you can't make location conditions which involve items of the same type.

So you can't check for space monsters in the location condition of space monsters, and the same with specials. And because of the order in which things are done during universe generation, you also can't check for specials in space monster location conditions (that is, you can, but it won't do anything), but you can check for space monsters in specials location conditions.

That's a set of rules content scripters need to be aware of and should be able to work with.

If we take that road (at least as a stop-gap solution for now), which I propose we do, then you need to apply the approach implemented here for monsters also to the distribution algorithm of the specials. IMO that's the best, most consistent and most performance friendly quick solution.

Vezzra (changes requested):
Review of the revised monster distribution.

python/server/ServerWrapper.cpp
```diff
@@ -475,18 +524,8 @@ namespace {
...
-                ErrorLogger() << "MonsterFleetPlanWrapper::Location: Couldn't get object with ID " << object_id;
-                return false;
-            }
-
-            // get location condition and evaluate it with the specified universe object
-            // if no location condition has been defined, any object matches
-            const Condition::ConditionBase* location_test = m_monster_fleet_plan->Location();
-            return (!location_test || location_test->Eval(obj));
+        list Locations(list systems) {
+            return FilterIDsWithCondition(m_monster_fleet_plan->Location(), systems);
```
Vezzra, line 505: Why *replace* `Location` with `Locations`? I'd keep the former and just add the latter, both are useful.

Having both functions (`Location` to test a single object against the location condition and `Locations` to get a list of all objects that match a location condition) for the specials would be a good idea too (if you're willing to add that one too).

python/server/ServerWrapper.cpp
```diff
@@ -906,6 +945,22 @@ namespace {
         return field->ID();
     }
 
+    // Wrappers for Object class member functions
```
Vezzra: That comment is misleading, `ObjectsGetSystems` isn't a member function of the `UniverseObject` class (wrappers for those start around line 550 by the way, so if you add a wrapper for one of them you should place it there), as it acts on a *list* of universe objects, so to speak.

If anything, it maybe can be counted with the wrappers for member functions of the `Universe` class, because it kind of acts on the game universe. Insofar this wrapper can be considered to be in the right place already. Just adjust the comment, or remove it.

Vezzra (changes requested):
Review for changes to `ServerWrapper.cpp`.

Vezzra:
Ok, so far, so good, now only `specials.py` is left. This will take a bit longer for me to wrap my head around and review, so I have to ask for your patience. I'll try to get to it within the next few days.

python/server/ServerWrapper.cpp
LGM-Doyle: `Locations()` is orders of magnitude faster than `Location()` .  Checking all of the systems at once with `Locations()` is just as fast as checking one location with `Location()`.

Removing `Location()` encourages vectorizing those tests, which is much faster.

This is high-handed, but I think that the difference is large enough to justify it.

For example replacing https://github.com/freeorion/freeorion/pull/1225/files/d8707e3f267a172e7293311847c92d54c427c858#diff-f6a824c38bd1953f33a55ae2606fc199R146,

`fp_location_cache[fleet_plan] = set(fleet_plan.locations(systems))`

with a loop adds 20.5s to the startup time of a 999 galaxy.

python/server/ServerWrapper.cpp
LGM-Doyle: Rephrased.

LGM-Doyle (commented):
I also addressed the following concern, but couldn't find the original request to respond directly.

@Morlic-fo I have changed phrasing to lowest merit in a set to avoid talking about the merit of a set.

default/python/universe_generation/empires.py
```diff
@@ -154,11 +169,15 @@ def find_home_systems_for_min_jump_distance(self, systems_pool, min_jumps):
...
-                num_complete_misses_remaining -= 1
-
             if len(candidate) < self.num_home_systems:
+
+                # Break if the normal distribution of failed candidates is too much smaller than
+                # the target number of systems
+                lens_of_failed_candidates.append(len(candidate))
+                mean_len_of_failed_candidates = sum(lens_of_failed_candidates) / float(len(lens_of_failed_candidates))
+                if mean_len_of_failed_candidates < lowest_continuable_mean_len_of_candidates:
+                    break
```
Morlic-fo: Your ```expected_len_candidate_std``` is based on 3 measurements. You have 2 variables and don't even use one in the final formula. Unless you have some sound theoretical model to accompany that formula, I don't buy it the slightest. It seems arbitrary to me (you could fit up to N-> infinity different functions that would approximate your 3 measurements well enough...). Also I can't quite believe that a Cluster with Low starlanes would yield the same std as an elliptical galaxy with high starlane density.

You write that ```lens_of_failed_candidates``` is roughly a normal distributed variable.
In that case, I think it would be more reasonable to estimate the underlying distribution here by calculating mean and standard deviations and break if ```(mean + m*std) < num_home_systems``` with a minimum number of tries...


Even assuming your estimated std was correct, your break condition is completely unstable against statistics. You need some minimum sample size before you break. A single bad first selection can screw you up badly otherwise. So what you probably would want to do is calculate and consider the uncertainty of the mean here as well (i.e. std / sqrt(n)).  But again, I don't think the approach is valid at all... Correctness should definitely go before speed for Universe Generation, imo.

default/python/universe_generation/empires.py
```diff
@@ -172,11 +172,11 @@ def find_home_systems_for_min_jump_distance(self, systems_pool, min_jumps):
             if merit >= min_planets_in_vicinity_limit(get_systems_within_jumps(system, HS_VICINITY_RANGE)):
                 break
 
-            # If we have a better candidate, set the new lower bound one higher to try for a better candidate.
+            # If we have a better candidate, set the new lower bound and try for a better candidate.
             if merit >= current_merit_lower_bound:
```
Morlic-fo: This should be a stricter greater now.

default/python/universe_generation/empires.py
```diff
@@ -113,19 +113,13 @@ def find_home_systems_for_min_jump_distance(self, systems_pool, min_jumps):
             if system not in self.system_merit:
                 self.system_merit[system] = calculate_home_system_merit(system)
 
-        # In the list of systems reverse sorted by merit, if the first
-        # num_home_systems are all min_jumps apart, then the merit of the
-        # num_home_systems nth system is the best_case_merit_lower_bound that
-        # can be achieved.
+        # The list of merits and systems reverse sorted by merit.
```
Morlic-fo: Use descending instead of "reverse".

Morlic-fo (commented):
I am not convinced that the new heuristic for early exit is valid.

default/python/universe_generation/empires.py
LGM-Doyle: Fixed.

default/python/universe_generation/empires.py
LGM-Doyle: Fixed.

default/python/universe_generation/empires.py
LGM-Doyle: tl;dr;  I recommend going with the original solution of breaking immediately if `len(candidates)` is too small.


First, we should return to the original version of the PR and quit as soon as we fail, `len(candidate) < self.num_home_systems` and proceed to try an easier pool.

The algorithm tries to find a set of _m_target_ home systems out of _n_ total.  It first tries with a spacing _s_ between home systems of 10 and a series of more relaxed systems pools.  Then it repeats for a system spacing of 9 through 5.  For a given set of _n_ systems and spacing _s_ there is some _m_max_ home systems and if it is less than the _m_target_ the algorithm can't succeed. Trying more is a wasted effort.

When the algorithm fails it produces _m_failed_ candidates with some probability distribution.  It is only reasonable to try a second attempt if there is enough overlap of _m_target_ and this distribution to expect to succeed.  Otherwise, it is faster to try one of the more relaxed pools.

The variance of _m_failed_ is small relative to _n_.  This is easy to verify by printing out the `len(candidate)` and eyeballing it.  I provide some data below.  Consequently, if the first candidate is too small then all other candidates from the same pool with the same _s_ will with high probability be too small.

So actually using the variance to decide when to continue, only provides improved results if the first failed candidate is within some reasonable distance of _m_target_ home systems.  The total improvement is an increase in _s_ of 1.  This is a lot of angst for minimal gain, which is why I would return to the original PR code and quit immediately  and try an easier pool.

Let me try to explain the theory more clearly.

The `std = k * (_n_ systems)^(1/2)` for some coefficient _k_ is not an arbitrary interpolation.  Note, I made a mistake and carried a cubed root through all of the calculation.  It is expected standard deviation of _m_failed_home_systems_ from the math.  This can be shown in three steps that the distribution is expected to be normal, to have a mean proportional to _m_systems and have a standard deviation as above.

First, for a sufficiently large galaxy we could imagine it as _p_ galaxies stitched together.  We expect the big galaxy to be able to fit _p_ times as many home systems as each individual small galaxy.  Sufficiently large means that we ignore the perimeter effects.  So we expect the distribution of _m_failed_ to have a mean proportional to _n_.

Second, as our large galaxy is composed of _p_ smaller galaxies its _m_failed_ will be the sum of all the _p_ smaller galaxies _m_failed_ random variables.  We apply the central limit theorem, and know that such a sum of independent identically distributed random variables will have a normal distribution.  We now know that _m_failed_home_systems_ will be a normal random variable with a mean proportional to _n_ systems

Third and finally, also from the central limit theorem we know that the variance will be _p_ * _variance_small_galaxy_.  The standard deviation is the sqrt of the variance.  And again for some sufficiently large galaxy we can make _p_ proportional to _n_ systems.  Hence we get the final result `std = k * (_n_ systems)^(1/2)`.

So it only remains to get an estimate for _k_.  These were the experiments that I did.  As you pointed out, _k_ depends on the galaxy topology.  The 3 arm spiral std is about half of the cluster galaxy std.  However, we are estimating this variance in order to exclude cases that would run pointless trials, so I used the smaller variance.


Why not do this fitting in game?  It takes a long time for large galaxies, completely defeating the intent of using it to avoid wasteful computation.  Furthermore, we know the distribution a priori and it will not change in game.  Empirically determine the threshold and just use it in game. 


Here are some concrete numbers from a 3-arm spiral galaxy with 999 systems (1004 actually generated) targeting 39 home systems.  These are the length of the first 100 candidates for the first pool at a `min_jumps` of 10:

[13, 15, 16, 15, 16, 17, 16, 15, 17, 14, 15, 14, 14, 16, 17, 15, 16, 15, 14, 14, 15, 15, 13, 15, 15, 14, 17, 14, 16, 15, 15, 14, 14, 16, 16, 14, 17, 15, 13, 16, 15, 16, 16, 15, 16, 15, 17, 16, 16, 16, 17, 17, 14, 15, 15, 14, 15, 13, 16, 16, 16, 14, 18, 14, 16, 15, 15, 16, 16, 15, 17, 16, 15, 15, 14, 16, 17, 15, 15, 16, 15, 15, 13, 14, 15, 16, 15, 14, 17, 15, 14, 15, 15, 15, 15, 16, 16, 15, 13, 16]

You can verify that the mean is 15.23 and the standard deviation is 1.09. The goal is 23.85 standard deviations away from the first candidate, 13.  The next 99 trails are wasted computation.  There is no need to wait for an estimator to settle in order to decide to skip to the next trial.

In summary, I recommend removing the variance portion of the exit condition, and break if any candidate is shorter than the goal.  For galaxy up to the five times the current maximum size this is the correct decision.  Even for much larger galaxies the certainty of trying with an easier pool out-weighs the minimal benefit an increase of one in the minimum home system spacing.

This exit condition is useful, particularly if there is interest in large galaxies, because these extra trials are completely wasted computation and they scale worse than linear.

python/server/ServerWrapper.cpp
Vezzra: Ok, I wasn't aware that checking all systems at once with `Locations` is almost as fast as checking only one with `Location`. So, that means a line like this:
```
if system in monster_fleet.Locations():
    ...
```
is so close to check `system` with a call to `Location` performance-wise that it's practically pointless to keep the `Location` function, right? In this case I agree.

default/python/universe_generation/empires.py
Vezzra: Ok, I'm officially out-mathed. 😉 

From the little bit I understand of the above exchange, I want to suggest this:

When I suggested to scale the fixed variance of 4, I have been thinking of something simple like `num_home_systems/10 + 1`. Judging by the numbers of the example above (999 system map with 39 home systems), a fixed variance of 4 would probably be sufficient for maps with ten thousands of systems and 100+ home systems, a limit we'll probably never hit, at least not within the next 10-20 years.

Which means, returning to the prior version, where we had a check for a "complete miss" (a complete miss being `len(candidate) < num_home_systems - 4`) and exit early after 5 such "complete misses" sounds like a reasonable approach to me.

This should be negligible performance-wise, while not as aggressive as quitting immediately when a candidate set doesn't yield at least `num_home_systems` systems. The cases of that happening because of a very unlucky random selection might be very small, but I still want that cases be avoided as best as possible.

> The total improvement is an increase in s of 1. This is a lot of angst for minimal gain

I beg to differ. Even when considering the difference between a minimum spacing of 10 (which is the highest minimum spacing we begin with) vs. 9, I wouldn't call that "minimal". And it's only getting more significant from there on.

And the "lot of angst" is deliberate. Getting the required number of home systems as far away from each other as possible, and if possible from the strictest pool has absolute priority over speed.

The latter condition is so important because using the more relaxed pools means the min system and planets within a certain vicinity limit most likely won't be met by all home systems, which means planets have to be added to the systems in the vicinity of those home systems, which I want to avoid as much as possible.

default/python/universe_generation/monsters.py
Vezzra: >  I checked all of the existing monster code and monsters only care about monster in their system and distance from an empire.

Well, doesn't that mean that with each monster placed cached location conditions might become invalidated? Even is monsters care only about monsters in the same system, with that approach monsters might be placed in systems together with other monsters they are not supposed to be in the same system with.

So, either we don't pre-calculate and cache the location conditions, or we need at least inform content scripters that until we properly fix this code, they can't rely on monster fleet location conditions to work on other monsters. Existing locations conditions need to be checked and if necessary adjusted to account for that, otherwise we're going to actually introduce a bug here.

All said, I'm still extremely wary of that change. Even if it didn't cause any issues with the current monster fleet location conditions, any modification or addition that does (and it's easy enough to come up with something that does) would give unintended/buggy results.

This is a bit much for a speed gain for universe generation, even if it's a noticeable one. To cite @Morlic-fo, correctness is way more important than speed for universe generation.

default/python/universe_generation/specials.py
```diff
@@ -1,17 +1,68 @@
...
+# four   REPEAT_RATE[0] * REPEAT_RATE[1] * REPEAT_RATE[2]
+REPEAT_RATE = {1: 0.08, 2: 0.05, 3: 0.01, 4: 0.00}
+
+
+def calculate_number_of_specials_to_place(objs):
+    """Return a list of number of specials to be placed at each obj"""
+    return [1 if random.random() > REPEAT_RATE[1] else
+            2 if random.random() > REPEAT_RATE[2] else
+            3 if random.random() > REPEAT_RATE[3] else 4
+            for __ in objs]
```
Vezzra: Just curious: why use a doube underscore here than just a single one?

adrianbroher: You're probably aware that in python it is convention to use an underscore to indicate that a  variable inside loops or an unpacking isn't used.  However this conflicts with another convention for the `gettext` library that uses a single underscore as alias for the `gettext` function which fetches a localized text by a unlocalized key.

default/python/universe_generation/specials.py
```diff
@@ -20,57 +71,87 @@ def distribute_specials(specials_freq, universe_objects):
...
+
+    # Equal to the largest distance in WithinStarlaneJumps conditions
+    # GALAXY_DECOUPLING_DISTANCE is used as follows.  For any two or more objects
+    # at least GALAXY_DECOUPLING_DISTANCE appart you only need to check
+    # fo.special_locations once and then you can place as many specials as possible,
+    # subject to number restrictions.
+    #
+    # Organize the objects into sets where all objects are spaced GALAXY_DECOUPLING_DISTANCE
+    # appart.  Place a special on each one.  Repeat until you run out of specials or objects.
+    GALAXY_DECOUPLING_DISTANCE = 6
```
Vezzra, line 100: Ok, if I understand the algorithm correctly, what you're trying to achieve is to minimize the number of calls to the expensive `special_locations` API function by this organization of the passed in list of universe objects into sets where all objects are at least `GALAXY_DECOUPLING_DISTANCE` jumps apart, which allows you to precalculate and cache the location conditions.

That's nifty, but of course works only under the assumption that no `WithinStarlaneJumps` condition is used in the location condition of any special which specifies a jump distance limit that is greater than 6.

Though somewhat reasonable, this still isn't correct, as you simply can't make this assumption safely. Nevertheless, it's good enough for a stop-gap solution I guess, so I'm ok with that approach for now. However, we need to revisit this entire placement thing (and I'm talking about monsters, specials and natives) at some point to come up with a 100% correct solution.

Sorry, long term I won't relent on that. :wink:

default/python/universe_generation/specials.py
```diff
@@ -20,57 +71,87 @@ def distribute_specials(specials_freq, universe_objects):
...
+
+    i_cache = 0
```
Vezzra: What's the purpose of this variable? Unless I missed something, you only increment it with each iteration of the `while` loop below, and do nothing else with it.

Vezzra (changes requested):
Ok, *finally* got around to review the changes to the specials distribution algorithm (sorry it took me so long). Aside from my comments looks good - as far as I can tell and have been able to wrap my brain around it. :wink:

However, as stated in my comments, this can only be a stop-gap solution, given the current compromises wrt to correctness.

Vezzra:
After going over this and the discussions we had, I'm a bit curious.

I've to admit I'm a bit surprised to what length you're willing to go to optimize all those distribution algos for speed. Especially considering that universe generation only happens once, having to wait for even 20 min instead of maybe 5 (or even just one) isn't *that* big deal in my eyes (contrary to the performance/UI lags during the game, which are a constant annoyance once they become noticeable, and can even render a game unplayable if they get too bad).

*Correctness* on the other side is a very big deal IMO, even when it comes to very rare edge cases, and should *always* takes precedence over speed/performance. While we can relent on that a bit when it comes to stop-gap solutions, especially for things where correctness has already been broken anyway, it's not acceptable for any implementation considered/intended as final.

default/python/universe_generation/monsters.py
Vezzra: Hm, this discussion is moving beyond the scope of a PR comment section, so I think we better stop here. 😉 It's not an urgent issue anyway, nothing we need to resolve right now. I plan to get back to all this on the forum, once I have a bit more time (har, har... :wink:).

Vezzra:
Had to cherry-pick b6126e90ec1fe3c925ae1642bc4cca7da67f7130 to get past the compile errors this commit fixed. However, there is still one compile error when I try to build this PR:
```
/Users/user/SoftwareProjekte/FO/src/client/human/HumanClientApp.cpp:1002:71: error: indirection requires pointer operand ('std::initializer_list<boost::optional<std::__1::vector<int, std::__1::allocator<int> > > >' invalid)
        m_networking.SendMessage(RequestCombatLogsMessage(PlayerID(), *incomplete_ids));
                                                                      ^~~~~~~~~~~~~~~

```
Which is a bit strange, since your PR doesn't touch `HumanClientApp.cpp`, but I can't remember I ever got that error with master...

LGM-Doyle (commented):
I rebased to master to pick up the mentioned bugfix commit.

I added a commit which might fix your compile problem, which I think is from Clang trying to use the list initializer for the vector, when it won't work.  I don't know why this isn't also a problem in master.

I'm not suggesting sacrificing correctness for speed.  The recursive placement bug has existed for a least a year.  No one noticed.  It is fairly benign.  When I prepared this PR, I figured that it is fast and no more broken than the existing code.

Startup time is somewhat important for three reasons:
- Perception of the game quality.  If it takes 10 minutes to generate a 100 system game on a 1GHz machine, is it reasonable to take 600e6 operations per system?  A good place to start improving game performance is at the beginning.
- Testing.  I wrote this code when I was testing something and each test case initially took 3 minutes each.  I hacked it down to 14 seconds and was able to generate many test cases.  Yah.  I hope someone else improves it further to 1 second.  
- Extensibility of the software.  If someone proposes a new feature that will double startup time, the threshold for accepting that change is different at a 10 second startup vs. 10 minutes startup time.      

Speaking of doubling startup time.  I have an extension to this PR, that is not complete yet.  It may double statrup time, but it will correct the recursive location condition problem.

Morlic-fo (approved):

default/python/universe_generation/empires.py
```diff
@@ -95,29 +85,115 @@ def min_planets_in_vicinity_limit(num_systems):
...
+        # (999, 39) and (199, 19) the following was observered.  The distribution of candidate
+        # length is a normal random variable with standard deviation approximately equal to
+
+        # expected_len_candidate_std = (self.len_systems ** (1.0/2.0)) * 0.03
+
+        # which is about 1 for 1000 systems.  It is likely that anylen(candidate) is within 1
+        # standard deviation of the expected len(candidate)
+
+        # If we are within the miss_threshold of the target then try up to num_complete_misses more times.
+        miss_threshold = 3
```
Morlic-fo: capitalize constants.

default/python/universe_generation/empires.py
Vezzra: Hm, that's certainly not optimal and something to keep in mind. There has also been another complaint on the forum about a bad starting position, where a player was backed into a corner as you describe here, and their only exit 2 systems away blocked by a maintenance ship.

All this makes me think that the entire placement mechanics/algorithms don't really work well in their current form. We've been trying to patch things up, and it got better, but it remains patchwork, and new issues continue to turn up.

I guess the entire original approach (which I more or less migrated unchanged from the original C++ implementation) is fundamentally flawed. That thing needs a major revision. Or better, a complete rewrite. But that's subject to a design discussion on the forum.

default/python/universe_generation/empires.py
```diff
@@ -95,29 +85,115 @@ def min_planets_in_vicinity_limit(num_systems):
...
-
-
-def find_home_systems(num_home_systems, pool_list, jump_distance, min_jump_distance):
+class HomeSystemFinder(object):
+    """Finds a set of home systems with a least ''num_home_systems'' systems."""
+    def __init__(self, _len_systems, _num_home_systems):
+        # cache of sytem merits
+        self.system_merit = {}
+        self.num_home_systems = _num_home_systems
+        self.len_systems = _len_systems
```
Vezzra: Unless I've missed something, `len_systems` isn't used anywhere anymore (with switching back to the previous early exit condition), so it can be removed.

default/python/universe_generation/empires.py
```diff
@@ -95,29 +85,115 @@ def min_planets_in_vicinity_limit(num_systems):
...
+                best_candidate = [s for (_, s) in merit_system]
+
+                # Quit sucessfully if the lowest merit system meets the minimum threshold
+                if merit >= min_planets_in_vicinity_limit(get_systems_within_jumps(system, HS_VICINITY_RANGE)):
+                    break
+
+        return best_candidate
+
+
+def find_home_systems(len_systems, num_home_systems, pool_list, jump_distance, min_jump_distance):
```
Vezzra: Ditto.

default/python/universe_generation/empires.py
```diff
@@ -126,6 +202,8 @@ def find_home_systems(num_home_systems, pool_list, jump_distance, min_jump_dista
     This parameter contains a list of tuples, each tuple has a pool of systems as first element and a description
     of the pool for logging purposes as second element.
     """
+
+    finder = HomeSystemFinder(len_systems, num_home_systems)
```
Vezzra: Ditto.

default/python/universe_generation/empires.py
```diff
@@ -299,15 +378,15 @@ def compile_home_system_list(num_home_systems, systems, gsd):
         # specify 0 as number of requested home systems to pick as much systems as possible
         (pool_matching_sys_limit, "pool of systems that meet at least the min systems limit"),
     ]
-    home_systems = find_home_systems(num_home_systems, pool_list, min_jumps, HS_MIN_DISTANCE_PRIORITY_LIMIT)
+    home_systems = find_home_systems(len(systems), num_home_systems, pool_list, min_jumps, HS_MIN_DISTANCE_PRIORITY_LIMIT)
```
Vezzra: Ditto.

default/python/universe_generation/empires.py
```diff
@@ -299,15 +378,15 @@ def compile_home_system_list(num_home_systems, systems, gsd):
...
 
     # check if the first attempt delivered a list with enough home systems
     # if not, we make our second attempt, this time disregarding the filtered pools and using all systems, starting
     # again with the min_jumps jump distance limit and specifying 0 as number of required home systems to pick as much
     # systems as possible
     if len(home_systems) < num_home_systems:
         print "Second attempt: trying to pick home systems from all systems"
-        home_systems = find_home_systems(num_home_systems, [(systems, "complete pool")], min_jumps, 1)
+        home_systems = find_home_systems(len(systems), num_home_systems, [(systems, "complete pool")], min_jumps, 1)
```
Vezzra: Ditto.

Vezzra:
Ok, builds and runs fine now on OSX. A quick test indicates that the algorithm seems to work well, with the default quickstart game it even seems to pick almost the same set of home systems (which I didn't expect). So far looks very good.

The only thing I'm still very unsure about are the changes to the monster distribution algorithm. My main objection remains, that with caching the location condition location conditions that prevent placing monsters within a certain distance of certain other monsters is even more broken now than before.

However, as you already pointed out, the current algorithm is already broken, and a more thorough revision is needed anyway. Also taking into account that I expect the new algorithm not to make things too much more worse, I won't let that hold up merging this PR anymore. Lets see how it works out, and make subsequent adjustments as the need arises.

One thing might be to put matching locations condition in the definitions of all monsters that are supposed to be spawned at least a certain distance away from each other (or just not be spawned in the same system).

@MatGB, I suppose you know the monster fleet definitions in question best (as I guess the only case where this actually poses a problem are the experimentors), can you check which location conditions might be in need of adjustment?

So I'm going to approve this PR, as far as I'm concened LGTM. @LGM-Doyle, I'm not merging immediately, in case want to make the last minor corrections @Morlic-fo and I pointed out in our last comments. You're free to merge yourself, otherwise I'm going to merge before I produce the next test builds.

Thanks for your patience with my maybe a bit annoying nitpicking and my taking such a long time to finally get my reviewing done. 😉

Vezzra (approved):

Vezzra:
> I'm not suggesting sacrificing correctness for speed. The recursive placement bug has existed for a least a year. No one noticed. It is fairly benign. When I prepared this PR, I figured that it is fast and no more broken than the existing code.

That's perfectly fine, as long as we're on the same page that somewhere down the road this needs to be fixed, and a final solution has to be 100% correct.

Regarding the recursive placement bug, AFAIK that actually existed since forever. The basic algorithms of the monster and special distribution has been migrated from the old C++ implementation of universe generation, and hasn't really been changed on a fundamental level, there have only been small adjustments IIRC.

It's the home system selection algorithm that has basically been rewritten since.

> Startup time is somewhat important for three reasons...

The reasons you cite are certainly valid. I mean, it's not that I haven't been annoyed on more than one occasion when I had to run a lot of test games and universe generation took its sweet time. :wink:

Still, as a player I'm still less annoyed if universe generation takes somewhat longer, but therefore more carefully ensures that starting positions are evenly spaced, not too bad wrt to the near neighborhood (that one still needs work), and also ensuring that the galaxy generated fits the parameters I choose in game setup, than the other way round.

At least, that's how I feel.

> Speaking of doubling startup time. I have an extension to this PR, that is not complete yet. It may double statrup time, but it will correct the recursive location condition problem.

Ok, when putting efforts into this just keep in mind that I consider re-designing and rewriting this entire part of universe generation from scratch in the long run. That may be several release cycles away, and who knows if I'm still around then, though. ;)

LGM-Doyle:
Thanks everyone for looking over this PR.

Special thanks to @Vezzra for asking if I could tighten up the exit conditions and to @Morlic-fo for asking for justification of the result.

PR #1226, opened 2017-01-18
Improve combat log for fighters.

LGM-Doyle:
This PR makes 4 small changes to the combat log window:
- moves the launch events into the bout event before the destroy events.  This was requested in http://freeorion.org/forum/viewtopic.php?f=28&t=10339.
- restricts debug logging to when verbose logging is requested
- makes the NEUTRAL empire name the same color as its links
- changes the destroyed object report to indicate empire with color instead of explicitly by name.  This is more in line with the other types of logs.

### DIFF
--- a/UI/CombatReport/CombatLogWnd.cpp
+++ b/UI/CombatReport/CombatLogWnd.cpp
@@ -386,11 +386,11 @@ void CombatLogWnd::CombatLogWndImpl::SetLog(int log_id) {
                            GetOptionsDB().Get<bool>("verbose-combat-logging");
 
     m_wnd.DeleteChildren();
-    GG::Layout* layout = new GG::DeferredLayout(m_wnd.UpperLeft().x, m_wnd.UpperLeft().y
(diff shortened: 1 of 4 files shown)


UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -381,6 +382,9 @@ void CombatLogWnd::CombatLogWndImpl::SetLog(int log_id) {
         return;
     }
 
+    bool verbose_logging = GetOptionsDB().Get<bool>("verbose-logging") ||
+                           GetOptionsDB().Get<bool>("verbose-combat-logging");
+
     m_wnd.DeleteChildren();
     GG::Layout* layout = new GG::DeferredLayout(m_wnd.UpperLeft().x, m_wnd.UpperLeft().y
                                         , m_wnd.Width(), m_wnd.Height()
```
geoffthemedio: not from this PR, but the indenting here is odd

LGM-Doyle:
I fixed the indenting.

PR #1227, opened 2017-01-18
Forward Declare Enums

geoffthemedio:
Works for me, but want to check with others before merging...

### DIFF
--- a/AI/AIInterface.cpp
+++ b/AI/AIInterface.cpp
@@ -12,6 +12,7 @@
 #include "../universe/ShipDesign.h"
 #include "../universe/System.h"
 #include "../universe/Tech.h"
+#include "../universe/Enums.h"
 #include "../Empire/Empire.h"
 
 #include "../util/Directories.h"
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -3320,12 +3320,8 @@ void Empire::InitResourcePools() {
 
 
     // set stockpile object locations for each resource, ensuring those systems exist
-    std::vector<ResourceType> res_type_vec;
-    res_type_vec.push_back(RE_INDUSTRY);
-    res_type_vec.push_back(RE_TRADE);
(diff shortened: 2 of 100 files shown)


adrianbroher:
I don't see the point of this change. It doesn't really reduce circular inclusion dependencies.

geoffthemedio:
It's not complete, so there are more removals of including Enums.h to be done. It does / will avoid having to include all the enum declarations into numerous headers (and thus source files) where they're not needed. I'm not concerned about circular include chains since the include guards should stop that for actually extending the resulting compiled source file.

adrianbroher:
Let me rephrase:

I don't see the point of the PR as it doesn't significantly reduce the number of included files when replacing `universe/Enums.h` with `universe/EnumsFwd.h`.  If you want to improve build time start with carving out the ValueRefBase, ConditionBase and EffectBase and place them into separate headers to avoid dependency inclusion pulled in by the concrete implementations.

geoffthemedio:
It doesn't reduce the number of included files, but is substantially reduces the complexity and length of what is included.

There already is https://github.com/freeorion/freeorion/blob/master/universe/ValueRefFwd.h

geoffthemedio:
Not sure how reproducible the Travis CI build times are, but at least right now this branch built a few min faster than the latest master.

geoffthemedio:
I've also split ConditionBase into a separate header from the rest of the Condition classes, but it isn't actually very helpful, since Condition.h isn't included in very many places where only ConditionBase is needed.

adrianbroher:
> I don't see the point of the PR as it doesn't significantly reduce the number of included files when replacing universe/Enums.h with universe/EnumsFwd.h. 

> It doesn't reduce the number of included files, but is substantially reduces the complexity and length of what is included.

I stand corrected then.

> Not sure how reproducible the Travis CI build times are, but at least right now this branch built a few min faster than the latest master.

Good job Geoff.

PR #1229, opened 2017-01-19
Imperial supply stockpile3

agrrr3:
The WIP version of imperial stockpile as discussed in

http://www.freeorion.org/forum/viewtopic.php?f=6&t=10374

The feature basically is usable:
* meter for yield (set to 0.2 at the moment)
* meter for maximum extraction from the stockpile per turn (set to ?10 PP? at the moment)
* overproduction automatically gets transferred to imperial stockpile PP using the yield ratio
* triggering an item in the build queue to allow/disallow use of the imperial stockpile
* use imperial stockpile IFF the item is allowed to use the stockpile and it is underfunded

What works but is too ugly:
* UI

Whats missing:
* tech tree, id like to have some feedback first in the discussion
* calculating expected turns

I guess the stockpile could/should be refactored to use a empire meter instead of reusing the old stockpile tied to the resource pool


universe/UniverseObject.h
```diff
@@ -29,6 +29,10 @@ FO_COMMON_API extern const int INVALID_OBJECT_ID;
 // The ID number assigned to temporary universe objects
 FO_COMMON_API extern const int TEMPORARY_OBJECT_ID;
 
+// The ID number assigned instead of the system which represents the whole imperium
+// Used for industrial stockpile
+FO_COMMON_API extern const int IMPERIAL_OBJECT_ID;
```
dbenage-cx: How does this differ from Empire ID?

Empire/Empire.cpp
```diff
@@ -1458,6 +1498,8 @@ int Empire::StockpileID(ResourceType res) const {
         return m_capital_id;
         break;
     case RE_INDUSTRY:
+        return IMPERIAL_OBJECT_ID;
```
geoffthemedio: just get rid of the concept of a stockpile location, rather than adding some fake ID to use for it...

agrrr3: Ok. Question: Should I switch to an empire meter instead using the resource pool stockpile as well?

universe/UniverseObject.h
agrrr3: This is a fake id (set to a constant value) and has no connection to empire id (which i would dynamically need to set/find out). 
This gets is used to compare against planets.
One question - are the ids of objects globally unique or just per type (so could be there an empire id 1 == and a planet id == 1 or not)?

Empire/Empire.cpp
agrrr3: Or should I rip out the current stockpile system as it is now (as it is not used). I could maybe also use the empire id as identifier (see the discussion with dbenage-cx above)

universe/UniverseObject.h
geoffthemedio: An empire could have the same id as an UniverseObject.

Empire/Empire.cpp
geoffthemedio: Storing an empire's stockpile as an empire meter makes some sense, especially if you're also storing the stockpile limit as a meter... Avoids the need for extra getter/setter functions and effects and valueref stuff, at least in principle... I think there aren't actually relevant valuerefs now to access an empire's meter values, though the EmpireMeterValue condition exists. That said, stockpile amounts should be persistent, so will have a different sort of behaviour from the other current empire meters, which are reset and re-calculated by effects each turn. It also makes sense to keep the stockpile tracking in the relevant resourcepool... By default I'd leave it in the resourcepool.

agrrr3: So, did I read you right? I should remove all the code relating to stockpile locations?

geoffthemedio: sure

agrrr3: Better this way?

universe/UniverseObject.h
Vezzra: @agrrr3, bear in mind that *empires* are not *objects* like planets, ships, etc. are. The latter are all subclasses of `UniverseObject`, and AFAIK all universe objects share the same "id pool" (so there can't be a ship and a planet with the same id).

However, as empires aren't universe objects, they have an independent "id pool", so, as @geoffthemedio already said, there can be an empire with the same id as a universe object (ship, planet, etc.).

agrrr3: Understood, thank you for the good explanations. As a sidenote: I'd prefer that all my IDs would share the same pool. Usually always corner cases emerge where it helps to be able to handle different things in the same way. If we would have kept the stockpile locations one could distinguish your own empire (imperial stock) from one of your planets (local stock) by the id.

agrrr3:
Obsolete. Imperial stockpile is developed now in #1710

PR #1234, opened 2017-01-21
Add Death Spores, Bio Terminators and Launch Bay icons.

V-ktor:
New icons for death spores, bio terminators and launch bay.
Use the old icons for dark ray and void shadow.


MatGB:
These are lovely, thank you, especially the launch bay, that's cool, reminds me of several SF shows without actually being any of them, perfect.

PR #1235, opened 2017-01-22
Add Dark Ray, Void Shadow and Space Elevator icons.

V-ktor:
Add Dark Ray, Void Shadow and Space Elevator icons.
Use the old icons for Sit Rep.


Vezzra:
Nice! Thanks for the good work! 😃

Issue #1230, opened 2017-01-20
Visible misuse of randomness


o11c:
Currently, `util/Random.cpp` contains two harmful functions:

    void Seed(unsigned int seed)
    void ClockSeed()

Seeding should *always* be done for the full state of the object (not just 32 bits of it), and using an unpredictable source (e.g. `boost::random_device`).

Particularly evil, `Seed` is called many times each turn with highly predictable input (object IDs), which leads to user-visible annoyances such as *always* finding the same things in Ancient Ruins.

Additionally, mt19937 is a particular bad RNG (especially for seeding). Something like PCG or xoroshiro would be better.

I've tested that the below patch "works" for single-player games, but I don't know if special care is required for multiplayer synchronization. Note that random map seeds do *not* go through this code path, although remember that some parts (e.g. creation of monsters and specials) happen after map creation, so do go through this.
````
diff --git a/util/Random.cpp b/util/Random.cpp
index 339baae65..c936af2c6 100644
--- a/util/Random.cpp
+++ b/util/Random.cpp
@@ -1,18 +1,24 @@
 #include "Random.h"
 
-#include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/random/random_device.hpp>
 
 namespace {
     GeneratorType gen; // the one random number generator driving the distributions below
 }
 
 void Seed(unsigned int seed) {
-    gen.seed(static_cast<boost::mt19937::result_type>(seed));
+    (void)seed;
+    ClockSeed();
 }
 
 void ClockSeed() {
-    boost::posix_time::time_duration diff = boost::posix_time::microsec_clock::local_time().time_of_day();
-    gen.seed(static_cast<boost::mt19937::result_type>(diff.total_milliseconds()));
+    static bool initialized = false;
+    if (!initialized)
+    {
+        boost::random_device seed_seq;
+        gen.seed(seed_seq);
+        initialized = true;
+    }
 }
 
 SmallIntDistType SmallIntDist(int min, int max)
````

geoffthemedio:
Updated the summary with markup around the patch to fix formatting.

geoffthemedio:
Reproducible / predictable "random" results is intentional. The idea is that you should be able to get the same galaxy setup using the same seed.

[...]

PR #1241, opened 2017-01-25
Replace mt19937 with PCG random number generator.

o11c:
In short, PCG is:

* smaller, without compromising quality (we don't need 625 dimensions, though PCG can offer arbitrary power-of-2 dimensionality anyway)
* faster
* more flexible (interesting APIs available)

The `#ifdef __SIZEOF_INT128__` is not strictly *necessary* (either algorithm *works* on all platforms), but it ensures we're using the fastest available on each platform. This *does* mean results don't match between platforms, but that's not needed.

The `_k2` variant is used on 32-bit platforms so that the state is the same size. We'll have to do our own serialization rather than use their `operator >>` though, when fixing #1230.

Also, fixed `ClockSeed` to use a better source and to fill all 256 bits of state..

***

Home page: http://www.pcg-random.org/

C++ implementation copied from: https://github.com/imneme/pcg-cpp with PRs 5 and 14 merged.

### DIFF
--- a/util/Random.cpp
+++ b/util/Random.cpp
@@ -1,18 +1,18 @@
 #include "Random.h"
 
-#include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/random/random_device.hpp>
 
 namespace {
     GeneratorType gen; // the one random number generator driving the distributions below
 }
 
 void Seed(unsigned int seed) {
-    gen.seed(static_cast<boost::mt19937::result_type>(seed));
+    gen.seed(static_cast<GeneratorType::result_type>(seed));
 }
 
 void ClockSeed() {
-    boost::posix_time::time_duration diff = boost::posix_time::microsec_clock::local_time().time_of_day();
-    gen.seed(static_cast<boost::mt19937::result_type>(diff.total_milliseconds()));
+    boost::random_device seed_seq;
+    gen.seed(seed_seq);
 }
 
 SmallIntDistType SmallIntDist(int min, int max)
--- a/util/Random.h
+++ b/util/Random.h
@@ -15,6 +15,7 @@
 #include <ctime>
 
 #include "Export.h"
+#include "pcg_random.hpp"
 
 /** \file
  *
@@ -31,7 +32,13 @@
  * call the Random*() functions.
  */
 
-typedef boost::mt19937                                                          GeneratorType;
+#ifdef __SIZEOF_INT128__
+/** If native 128-bit integers are available, use them. 64-bit output. */
+typedef pcg64                                                                   GeneratorType;
+#else
+/** Otherwise, use a different algorithm using 64-bit integers and 32-bit output. */
+typedef pcg32_k2                                                                GeneratorType;
+#endif
 typedef boost::variate_generator<GeneratorType&, boost::uniform_smallint<> >    SmallIntDistType;
 typedef boost::variate_generator<GeneratorType&, boost::uniform_int<> >         IntDistType;
 typedef boost::variate_generator<GeneratorType&, boost::uniform_real<> >        DoubleDistType;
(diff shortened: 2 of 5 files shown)


geoffthemedio:
> This does mean results don't match between platforms, but that's not needed.

We do need/want RNG results, particularly for galaxy generation, to be as consistent as possible between platforms.

Please also put all the PCG headers in their own subdirectory within util.

MatGB:
I think in the previous thread there was some talking across each other and this wasn't clearly stated.

**It is a design objective that the same seed and generation settings will create identical results across all platforms and between separate games.**

Disagreeing with that objective is fine, but a discussion of that nature does not belong on a PR or Issue thread. I strongly support the objective for a variety of reasons.

In the run up to each release, the playtest team puts a lot of effort into testing the default quickstart for balance and accessibility for new players, we have on occasion changed settings to ensure it is a good learn-to-play experience. Regardless of plans to introduce a "tutorial" scenario at some point, that remains a strong and desirable feature. As playtest/balance lead I'm not at all happy with the idea that feature should be removed at all.

dbenage-cx:
I did not see much info describing the differences between these generators.  (I would expect the sizeof check for `pcg64_k32`, but not `pcg64`?)  If only `pcg32` was utilized, do you know if there is a reasonable certainty for a portably consistent result set from a specific seed?

With GCC 6.3.1, there are link errors for `boost::random::random_device`:
```
[ 63%] Linking CXX executable ../freeoriond
../libfreeorioncommon.so: undefined reference to `boost::random::random_device::operator()()'
../libfreeorioncommon.so: undefined reference to `boost::random::random_device::~random_device()'
../libfreeorioncommon.so: undefined reference to `boost::random::random_device::random_device()'
collect2: error: ld returned 1 exit status
server/CMakeFiles/freeoriond.dir/build.make:473: recipe for target 'freeoriond' failed

geoffthemedio:
Might be worth noting that there's some additional seeding related code here: https://github.com/freeorion/freeorion/blob/master/server/ServerApp.cpp#L1235

It's re-implemented there so that the seed can be saved to the galaxy setup info.

o11c:
Okay, one if you has to make the decision: make it slower on 32-bit platforms, or make it slower on 64-bit platforms?

geoffthemedio:
I don't think the speed consideration is worth worrying about. But as @dbenage-cx mentioned, it needs to be portably reproducible.

dbenage-cx:
If there is not a drastic speed loss, I'd favor 64-bit due to 'market share', but at your discretion.

I only picked `pcg32` in the question as it seemed the least likely for conflicts if any arise.  A better question might be: do you know of a previous portability test for any of these generators?

I suppose we need to devise our own regardless, including one for the python side.

Vezzra:
Two things:

First, I just want to emphasize what has already been mentioned: whatever PRNG we decide to implement/use, it *must not* produce different results on different platforms.

Second, the universe generation scripts currently use the default PRNG of the Python standard library (which is MT19937 AFAIK). So replacing the use of boosts MT19937 PRNG with PCG in the backend C++ code won't have any effect on universe generation at all.

To achieve that, an interface for Python to the PCG implementation will have to be provided, and all Python scripts adjusted to use this instead of the default PRNG of the standard library.

Vezzra:
@geoffthemedio, how shall we proceed with this PR? There hasn't been any more progress or comments on the issues which have been pointed out (consistent results across platforms), so shall we keep that open or close as invalid?

@o11c, are you still there? Can you provide some feedback, particularly on the consistent-results-across-platforms issue?

geoffthemedio:
@Vezzra Unless the cross-platform consistency is ensured, this isn't acceptable as is.

The ClockSeed change could be done separately. It's not really a "clock" seed anymore, but the relevant function could be renamed.

Vezzra:
Well then, considering there hasn't been any more feedback from @o11c since January, I propose to reject this PR and close it as invalid. If anyone wants to pick up the ClockSeed change, they can do that in a separate PR.

PR #1247, opened 2017-01-26
Acquire scoped mutex locks before accessing PRNG generator.

geoffthemedio:
Hopefully makes accessing the random functions thread safe.

Related to:
https://github.com/freeorion/freeorion/issues/1230
https://github.com/freeorion/freeorion/pull/1241


geoffthemedio:
lacking any comments or objections...

Issue #1242, opened 2017-01-26
Merge duplicates of `SourceForEmpire`


adrianbroher:
This is an incomplete list of duplicate implementations of SourceForEmpire or similar named functions.  Those should be merged to a single function and probably be a part of the `Empire` class interface.

Executing `grep -rn 'capital?' *` has shown up the following results:

* universe/Building.cpp:226
* universe/Universe.cpp:3641
* universe/ShipDesign.cpp:162
* universe/Tech.cpp:148
* UI/EncyclopediaDetailPanel.cpp:959

PR #1250, opened 2017-01-27
Merge duplicates of source for empire.

LGM-Doyle:
This PR addresses https://github.com/freeorion/freeorion/issues/1242.

It also fixes a logic bug in that some universe objects were using invalid sources in their `ScriptingContexts`.


Empire/Empire.cpp
```diff
@@ -1468,6 +1470,43 @@ int Empire::StockpileID(ResourceType res) const {
...
 
+int Empire::SourceID() const {
+    std::shared_ptr<const UniverseObject> good_source = Source();
+    return good_source ? good_source->ID() : INVALID_OBJECT_ID;
+}
+
+std::shared_ptr<const UniverseObject> Empire::Source() const {
+    // Make it safe to use with a null this pointer.
+    // This generates a compiler warning, but it enables the natural line of code of
+    // "GetEmpire(id)->Source()" without requiring an intervening null check
```
geoffthemedio: Just have a GetSourceForEmpire(int id) function that can return nullptr. Not having a check that GetEmpire(id) didn't return nullptr is weird / misleading.

Empire/Empire.cpp
```diff
@@ -1480,8 +1519,10 @@ std::string Empire::Dump() const {
     return retval;
 }
 
-void Empire::SetCapitalID(int id)
-{ m_capital_id = id; }
+void Empire::SetCapitalID(int id) {
```
geoffthemedio, line 1519: probably shouldn't assume id is valid...

Empire/Empire.cpp
```diff
@@ -1480,8 +1519,10 @@ std::string Empire::Dump() const {
...
+    m_capital_id = id;
+    m_source_id = id;
```
geoffthemedio: for purposes of setting the source id at least

Empire/Empire.h
```diff
@@ -316,6 +317,14 @@ class FO_COMMON_API Empire {
     int                     CapitalID() const;                          ///< Returns the numeric ID of the empire's capital
     int                     StockpileID(ResourceType res) const;        ///< Returns the numeric ID of the empire's stockpile location for the resource of type \a res
 
+    /** Return an object id that is owned by the empire or
+        INVALID_OBJECT_ID. SourceID() is safe to call with a null this
+        pointer. */
+    int                     SourceID() const;
```
geoffthemedio, line 321: calling a member function on a null pointer is probably undefined behaviour... as above, no.

Empire/Empire.cpp
adrianbroher: > +    // Make it safe to use with a null this pointer.

That's nonstandard undefined behaviour. Rejected.

adrianbroher (changes requested):

Empire/Empire.cpp
LGM-Doyle: I added EmpireManager::GetSource(id) and removed the no-standard behavior.

Empire/Empire.cpp
LGM-Doyle: I've added a check that the capital exists and is owned by the empire.

Empire/Empire.cpp
LGM-Doyle: I've added a check that the source id is valid.  Source ids don't always need to exist in the code, it seems.

Empire/Empire.h
LGM-Doyle: Fixed.

Empire/EmpireManager.cpp
```diff
@@ -46,6 +46,11 @@ const Empire* EmpireManager::GetEmpire(int id) const {
     return it == m_empire_map.end() ? nullptr : it->second;
 }
 
+std::shared_ptr<const UniverseObject> EmpireManager::GetSource(int id) const {
+    auto it{m_empire_map.find(id)};
```
adrianbroher: The OSX clang compiler spitted out error messages the last time I used a similar construct.

See 89f801cc3e

Empire/Empire.cpp
```diff
@@ -1517,8 +1517,20 @@ std::string Empire::Dump() const {
...
-    m_capital_id = id;
-    m_source_id = id;
+    m_capital_id = INVALID_OBJECT_ID;
+    m_source_id = INVALID_OBJECT_ID;
+
+    if (id == INVALID_OBJECT_ID)
+        return;
+
+    // Verify that the capital exists and is owned by the empire
+    auto possible_capital {Objects().ExistingObject(id)};
```
adrianbroher: Same here.

> The OSX clang compiler spitted out error messages the last time I used a similar construct.
>
> See 89f801c

adrianbroher (approved):

Empire/EmpireManager.cpp
LGM-Doyle: New features, new quirks :smile:.  Fixed.

Empire/Empire.cpp
LGM-Doyle: Fixed.

LGM-Doyle (commented):
Thanks for the quick reviews.

adrianbroher:
Looks good.

PR #1253, opened 2017-01-28
Add EmpireMeterValue to FOCS for retrieving a value

agrrr3:
A generic way to get the value of an empire meter in FOCS. This is necessary for meters existing only in FOCS and helpful when in the process of introducing new meters.

Use like this:
EmpireMeterValue empire = Source.Owner meter = "METER_DETECTION_STRENGTH"

Be careful that the value only represents the meter value in the instant
when this gets called. E.g. other effects might change it afterwards.

Implementation Note:
Maybe the token should be EmpireMeter instead of EmpireMeterValue.

### DIFF
--- a/parse/DoubleComplexValueRefParser.cpp
+++ b/parse/DoubleComplexValueRefParser.cpp
@@ -26,6 +26,13 @@ namespace parse {
                   )     [ _val = new_<ValueRef::ComplexVariable<double> >(_a, _b, _c, _f, _d, _e) ]
                 ;
 
+            empire_meter_value
+                = (     tok.EmpireMeterValue_ [ _a = construct<std::string>(_1)]
+                     >  parse::label(Empire_token)  >  simple_int[ _b = _1]
+                     >  parse::label(Meter_token) > tok.string[ _d = new_<ValueRef::Constant<std::string> >(_1)]
+                  )     [_val = new_<ValueRef::ComplexVariable<double> >(_a, _b, _c, _f, _d, _e)]
+                ;
+
              direct_distance
                 = (     tok.DirectDistanceBetween_ [ _a = construct<std::string>(_1) ]
                      >  parse::label(Object_token) > simple_int [ _b = _1 ]
@@ -66,6 +73,7 @@ namespace parse {
 
             start
                 %=  part_capacity
+                |   empire_meter_value
                 |   direct_distance
                 |   shortest_path
                 |   species_empire_opinion
@@ -73,6 +81,7 @@ namespace parse {
                 ;
 
             part_capacity.name("PartCapacity");
+            empire_meter_value.name("EmpireMeterValue");
             direct_distance.name("DirectDistanceBetween");
             shortest_path.name("ShortestPathBetween");
             species_empire_opinion.name("SpeciesOpinion (of empire)");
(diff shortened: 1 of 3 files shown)


universe/ValueRef.cpp
```diff
@@ -1757,6 +1757,22 @@ double ComplexVariable<double>::Eval(const ScriptingContext& context) const
...
+        Empire* empire = GetEmpire(empire_id);
+        if (!empire)
+            return -1.0;
+
+        std::string empire_meter_name;
+        if (m_string_ref1)
+            empire_meter_name = m_string_ref1->Eval(context);
+        Meter * meter = empire->GetMeter(empire_meter_name);
+        if (!meter)
+            return 0.0;
```
dbenage-cx, line 1773: Consistent early return value?

parse/DoubleComplexValueRefParser.cpp
```diff
@@ -26,6 +26,13 @@ namespace parse {
                   )     [ _val = new_<ValueRef::ComplexVariable<double> >(_a, _b, _c, _f, _d, _e) ]
                 ;
 
+            empire_meter_value
```
geoffthemedio, line 29: Would be preferable to be able to write `EmpireDetection id = Source.Owner` or `Detection empire = Source.Owner` or similar. The part_capacity example refers to the meters directly, not `ShipPartMeter meter = Capacity id = Source.ID` or similar.

geoffthemedio (changes requested):

parse/DoubleComplexValueRefParser.cpp
agrrr3: This change is the small brother of SetEmpireMeter. 

EmpireDetection etc.. should probably also be done.
But not in this PR.

universe/ValueRef.cpp
agrrr3: I did not verify, but I think you could script an empire meter. So it could be that you ask for the value before the meter was created. For that I assumed a default value of 0 is sensible.
This is different from not finding the empire. That is certainly an error. I didnt see any error handling there so I returned a value that is probably recognizable as wrong (-1).

geoffthemedio: whether you could script a meter or not, the default return value should be 0.0 for consistency with similar cases like: https://github.com/freeorion/freeorion/blob/master/universe/Ship.cpp#L423

agrrr3: Yes, I wanted to let it stay 0.0.

What I think dbenage meant is that i return -1.0 if no empire can be found.

universe/ValueRef.cpp
```diff
@@ -1757,6 +1757,22 @@ double ComplexVariable<double>::Eval(const ScriptingContext& context) const
...
+        if (m_int_ref1)
+            empire_id = m_int_ref1->Eval(context);
+        Empire* empire = GetEmpire(empire_id);
+        if (!empire)
+            return -1.0;
+
+        std::string empire_meter_name;
+        if (m_string_ref1)
+            empire_meter_name = m_string_ref1->Eval(context);
+        Meter * meter = empire->GetMeter(empire_meter_name);
```
geoffthemedio: remove space between Meter and *

universe/ValueRef.cpp
geoffthemedio: don't change the 0.0. Change the -1.0 that appears elsewhere.

geoffthemedio:
looks OK

Issue #1261, opened 2017-02-01
Build/Research Costs and Times incorrect


dbenage-cx:
Build from source: d5f7f74

### Description
Costs and times for buildings and techs are incorrect, seemingly when the time or cost are defined using certain valuerefs.
Building example: Starlane bore - 999k PP / 999 turns
Research example: Planetary Cloud Cover - 999k RP / 5 turns

`git bisect` resulted in 8ac9609
reverting the portion for `universe/Building.cpp` restores the building costs to normal values.

geoffthemedio:
Yeah, that change looks wrong to me.
````
-        if (!source && !m_production_cost->SourceInvariant())
+        if (!source || m_production_cost->SourceInvariant())
             return 999999.9f;
````
The point was: if there is no source and the cost depends on the source, return a default value.

Now it's been changed to: if there is no source or the cost doesn't depend on the source, return a default value. That means that regardless of whether there is a source, if the cost doesn't depend on the source the default value will be used, which is wrong in cases like a constant cost that doesn't depend on the source.

If there is no source but the cost doesn't depend on the source, then things are fine to evaluate the valueref using a null source, since the cost doesn't depend on the source.

There could potentially be a problem is SourceInvariant() isn't implemented correctly, and an attempt is made to access the source within a supposedly-source-invariant valueref, but just not having a source is not a reason to never attempt to evaluate a valueref... many work fine without a valid source object... those are supposed to be the source-invariant ones. (Or at least the source-invariant ones are supposed to not change value depending on the source... there could be some that still attempt to access the source object even if their value doesn't depend on the source object, but that's a bug to fix within them, not here...)

LGM-Doyle:
Sorry.  I misunderstood that.  Your explanation clarifies it.

I thought that the intent was to return a reasonable default value when the source was missing or the whole thing was source invariant.  I see that this really is an error condition.

I will provide a solution.

Vezzra:
@dbenage-cx, a fix (PR #1262) has been merged, can you check if that solves the issue for you?

dbenage-cx:
@LGM-Doyle Thanks for providing a fix

PR #1262, opened 2017-02-01
Fix for issue #1261 Build/Research Costs and Times incorrect

LGM-Doyle:
I think that this fixes issue #1261 by reverting the problematic commit tracked down by @dbenage-cx .

I also expanded the error messages in the log, so that missing but expected sources show up as errors and not just weird default values.

### DIFF
--- a/universe/Building.cpp
+++ b/universe/Building.cpp
@@ -239,13 +239,20 @@ float BuildingType::ProductionCost(int empire_id, int location_id) const {
         if (m_production_cost && m_production_cost->ConstantExpr())
             return m_production_cost->Eval();
 
+        const auto arbitrary_large_number = 999999.9f;
+
         std::shared_ptr<UniverseObject> location = GetUniverseObject(location_id);
-        if (!location)
-            return 999999.9f;    // arbitrary large number
+        if (!location) {
+            ErrorLogger() << "Location is missing using production cost of " << arbitrary_large_number;
+            return arbitrary_large_number;
+        }
 
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
-        if (!source && !m_production_cost->SourceInvariant())
-            return 999999.9f;
+        if (!source && !m_production_cost->SourceInvariant()) {
+            ErrorLogger() << "Source is expected but missing using production cost of "
+                          << arbitrary_large_number;
+            return arbitrary_large_number;
+        }
 
         ScriptingContext context(source, location);
 
@@ -263,13 +270,20 @@ int BuildingType::ProductionTime(int empire_id, int location_id) const {
         if (m_production_time && m_production_time->ConstantExpr())
             return m_production_time->Eval();
 
+        const auto arbitrary_large_number = 9999;
+
         std::shared_ptr<UniverseObject> location = GetUniverseObject(location_id);
-        if (!location)
-            return 9999;    // arbitrary large number
+        if (!location) {
+            ErrorLogger() << "Location is missing using production time of " << arbitrary_large_number;
+            return arbitrary_large_number;
+        }
 
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
-        if (!source && !m_production_time->SourceInvariant())
-            return 9999;
+        if (!source && !m_production_time->SourceInvariant()) {
+            ErrorLogger() << "Source is expected but missing using production time of "
+                          << arbitrary_large_number;
+            return arbitrary_large_number;
+        }
 
         ScriptingContext context(source, location);
 
--- a/universe/Building.cpp
+++ b/universe/Building.cpp
@@ -246,11 +246,8 @@ float BuildingType::ProductionCost(int empire_id, int location_id) const {
             return arbitrary_large_number;
 
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
-        if (!source && !m_production_cost->SourceInvariant()) {
-            ErrorLogger() << "Source is expected but missing using production cost of "
-                          << arbitrary_large_number;
+        if (!source && !m_production_cost->SourceInvariant())
             return arbitrary_large_number;
-        }
 
         ScriptingContext context(source, location);
 
@@ -275,11 +272,8 @@ int BuildingType::ProductionTime(int empire_id, int location_id) const {
             return arbitrary_large_number;
 
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
-        if (!source && !m_production_time->SourceInvariant()) {
-            ErrorLogger() << "Source is expected but missing using production time of "
-                          << arbitrary_large_number;
+        if (!source && !m_production_time->SourceInvariant())
             return arbitrary_large_number;
-        }
 
         ScriptingContext context(source, location);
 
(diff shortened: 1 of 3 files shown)


universe/Building.cpp
```diff
@@ -239,13 +239,20 @@ float BuildingType::ProductionCost(int empire_id, int location_id) const {
         if (m_production_cost && m_production_cost->ConstantExpr())
             return m_production_cost->Eval();
 
+        const auto arbitrary_large_number = 999999.9f;
+
         std::shared_ptr<UniverseObject> location = GetUniverseObject(location_id);
-        if (!location)
-            return 999999.9f;    // arbitrary large number
+        if (!location) {
+            ErrorLogger() << "Location is missing using production cost of " << arbitrary_large_number;
```
geoffthemedio: not necessarily an error... pedia can show stuff even when there's no game ongoing.

LGM-Doyle: Is there a way to determine if the pedia is running off-line?

If so, I could add an earlier check `if (offline ) return arbitrary_large_number;` that does not generate an error message.

geoffthemedio:
even if it's in a game, the pedia value probably won't have a valid source. just don't treat it as an error.

Cjkjvfnby:
`Fix 1261 build times incorrect` did you cont all 1261 build times?   

PS Please add `#` to issue reference to avoid confusions.

LGM-Doyle:
I've removed the error messages on the Location clauses in Buildings.cpp.

o01eg:
What if use boost::optional instead special values?

geoffthemedio:
Then there would need to be a check later for whether the optional value wasn't provided, and in that case use arbitrary values or some other way to display something in the UI.

universe/Building.cpp
```diff
@@ -273,10 +271,8 @@ int BuildingType::ProductionTime(int empire_id, int location_id) const {
         const auto arbitrary_large_number = 9999;
 
         std::shared_ptr<UniverseObject> location = GetUniverseObject(location_id);
-        if (!location) {
+        if (!location)
             ErrorLogger() << "Location is missing using production time of " << arbitrary_large_number;
-            return arbitrary_large_number;
```
dbenage-cx: Did you mean to remove the previous line instead of this one?

LGM-Doyle: Fixed.

universe/Building.cpp
```diff
@@ -239,13 +239,18 @@ float BuildingType::ProductionCost(int empire_id, int location_id) const {
...
         std::shared_ptr<UniverseObject> location = GetUniverseObject(location_id);
         if (!location)
-            return 999999.9f;    // arbitrary large number
+            return arbitrary_large_number;
 
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
-        if (!source || m_production_cost->SourceInvariant())
-            return 999999.9f;
+        if (!source && !m_production_cost->SourceInvariant()) {
+            ErrorLogger() << "Source is expected but missing using production cost of "
```
geoffthemedio: same here... not always an error / unexpected.

the message is also poorly worded / ambiguous

universe/Building.cpp
```diff
@@ -263,13 +268,18 @@ int BuildingType::ProductionTime(int empire_id, int location_id) const {
...
         std::shared_ptr<UniverseObject> location = GetUniverseObject(location_id);
         if (!location)
-            return 9999;    // arbitrary large number
+            return arbitrary_large_number;
 
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
-        if (!source || m_production_time->SourceInvariant())
-            return 9999;
+        if (!source && !m_production_time->SourceInvariant()) {
+            ErrorLogger() << "Source is expected but missing using production time of "
```
geoffthemedio: ...

universe/ShipDesign.cpp
```diff
@@ -330,13 +330,20 @@ float PartType::ProductionCost(int empire_id, int location_id) const {
         if (m_production_cost->ConstantExpr())
             return static_cast<float>(m_production_cost->Eval());
 
+        const auto arbitrary_large_number = 999999.9f;
+
         std::shared_ptr<UniverseObject> location = GetUniverseObject(location_id);
-        if (!location)
-            return 999999.9f;    // arbitrary large number
+        if (!location) {
+            ErrorLogger() << "Location is missing using production cost of " << arbitrary_large_number;
```
geoffthemedio: ...

universe/ShipDesign.cpp
```diff
@@ -330,13 +330,20 @@ float PartType::ProductionCost(int empire_id, int location_id) const {
...
+            return arbitrary_large_number;
+        }
 
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
-        if (!source || m_production_cost->SourceInvariant())
-            return 999999.9f;
+        if (!source && !m_production_cost->SourceInvariant()) {
+            ErrorLogger() << "Source is expected but missing using production cost of "
```
geoffthemedio: ...

universe/ShipDesign.cpp
```diff
@@ -345,19 +352,26 @@ float PartType::ProductionCost(int empire_id, int location_id) const {
...
         return 1;
     } else {
         if (m_production_time->ConstantExpr())
             return m_production_time->Eval();
 
         std::shared_ptr<UniverseObject> location = GetUniverseObject(location_id);
-        if (!location)
-            return 9999;    // arbitrary large number
+        if (!location) {
+            ErrorLogger() << "Location is missing using production time of " << arbitrary_large_number;
```
geoffthemedio: ...

universe/ShipDesign.cpp
```diff
@@ -345,19 +352,26 @@ float PartType::ProductionCost(int empire_id, int location_id) const {
...
+            return arbitrary_large_number;
+        }
 
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
-        if (!source || m_production_time->SourceInvariant())
-            return 9999;
+        if (!source && !m_production_time->SourceInvariant()) {
+            ErrorLogger() << "Source is expected but missing using production time of "
```
geoffthemedio: ...

LGM-Doyle (commented):
I've removed 2 error messages in Buildings.cpp and 4 in ShipDesign.cpp.  

I've reworded the missing source error message in Tech.cpp.

universe/ShipDesign.cpp
geoffthemedio: really? I still see it in the diff

universe/Tech.cpp
```diff
@@ -160,8 +160,8 @@ float Tech::ResearchCost(int empire_id) const {
     } else {
         std::shared_ptr<const UniverseObject> source = Empires().GetSource(empire_id);
         if (!source && !m_research_cost->SourceInvariant()) {
-            ErrorLogger() << "Source is expected but missing using research cost of "
-                          << arbitrary_large_number;
+            ErrorLogger() << "Using default research cost of " << arbitrary_large_number
```
geoffthemedio: I'm also skeptical about logging an error message here for the same reasons...

LGM-Doyle:
I've removed all of the error messages.

I assumed that the ones you had not itemized were acceptable.

PR #1263, opened 2017-02-01
Refactor Pathfinder::distance_matrix_cache

LGM-Doyle:
This commit refactors the `distance_matrix_cache` to move all of the handling of the locks into the `distance_matrix_cache`.  

In the existing code, the caller of `get_or_lock_row()` passes in a lock.  If there is a cache miss then `get_of_lock_row()` locks the row, swaps the lock and returns null.  The caller is responsible for filling the row and unlocking the cache.

This PR changes the behavior so that the caller of `get_T()` passes in a cache miss handler.  If there is a cache miss then `get_T()` locks the row and calls the cache miss handler with the locked row.  

The cache miss handler is only responsible for filling the cache.

The caller of `get_T()` is only responsible for providing a cache miss handler.  In future PRs different callers can use the `distance_cache_matrix` in new ways while using a common cache miss handler.

All the handling of the locks happens in the `distance_matrix_cache`.  

This PR removes swapping of locks, which was required to transfer the lock out of `get_or_lock_row()`. This reduces the amount of locking for a cache miss from:
- shared_lock, shared_lock, shared_lock, unique_lock, (swap_shared,
swap_unique), (unlock_shared, unlock_unique)

to

- shared_lock, shared_lock, shared_lock, unique_lock.

### DIFF
--- a/universe/Pathfinder.cpp
+++ b/universe/Pathfinder.cpp
@@ -19,6 +19,7 @@
 
 #include <algorithm>
 #include <stdexcept>
+#include <limits>
 
 namespace {
     const double    WORMHOLE_TRAVEL_DISTANCE = 0.1;         // the effective distance for ships travelling along a wormhole, for determining how much of their speed is consumed by the jump
@@ -39,7 +40,7 @@ namespace {
      */
     template <class T> struct distance_matrix_storage {
         typedef T value_type;  ///< An integral type for number of hops.
-        typedef std::vector<T>& row_ref; ///< A row type protected by a mutex.
+        typedef std::vector<T>& row_ref;  ///< A reference to row type
 
         distance_matrix_storage() {};
         distance_matrix_storage(const distance_matrix_storage<T>& src)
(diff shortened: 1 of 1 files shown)


universe/Pathfinder.cpp
```diff
@@ -98,102 +99,70 @@ namespace {
...
-                m_row_lock.unlock();
-                m_lock.unlock();
-            }
-        };
-
         public:
-        /** try to retrieve an element, lock the whole row on cache miss
-          * if \p lock already holds a lock, it will be unlocked after locking the row.
-          */
-        boost::optional<T> get_or_lock_row(size_t row_index, size_t column_index, row_lock& lock) const {
```
geoffthemedio, line 135: comment on line 77 refers to "get_or_row_lock" which seemingly didn't exist before, and still doesn't exist after these changes.

geoffthemedio:
Builds and runs seemingly OK

universe/Pathfinder.cpp
```diff
@@ -74,9 +74,9 @@ namespace {
     /**distance_matrix_cache is a cache of symmetric hop distances
        based on distance_matrix_storage.
 
-    It enforces the locking convention with get_or_row_lock which
-    returns either a single intgral value or an entire write locked
-    row.
+    It enforces the locking convention with get_T which returns either a single
+    integral value or calls a cache miss handler to fill an entire write locked
+    row with data.
```
geoffthemedio: maybe "... returns either a single cached integral value, or calls a cache miss handler to fill a write-locked row with values and then returns the requested value." ?

LGM-Doyle: Comment is rephrased.

geoffthemedio:
Don't really understand the changes or have time to look through it, but seems plausible and still seems to run OK.

LGM-Doyle:
I hope that some of the changes will be clarified by the next PR in the series.

Thanks for reviewing this.

PR #1265, opened 2017-02-01
Change supply mechanic to prevent at peace empires from blocking supply.

LGM-Doyle:
This commit **changes the supply mechanic** to prevent peaceful empires from
blocking each others supply by:
- modifying the workaround for the AI that allows colonies to block
supply, to prevent that blocking for peaceful empires.
- making the top empire(s) only block the supply of empires that it is at
war with.
- only prevent propagation when the entered supply is from a hostile
empire.

This might deal with some of the issues from the forum, if people want to try it out.

Since it only relaxes the restrictions of supply it should not negatively affect the AI.

This would be a user visible change to the supply mechanic.

### DIFF
--- a/Empire/Supply.cpp
+++ b/Empire/Supply.cpp
@@ -257,16 +257,19 @@ void SupplyManager::Update() {
             }
         }
 
-        std::set<int> systems_where_others_have_supply_sources_and_current_empire_doesnt;
-        // add all systems where others have supply
+        std::set<int> systems_where_hostile_others_have_supply_sources_and_current_empire_doesnt;
+        // add all systems where hostile_others have supply
         for (std::map<int, std::map<int, float>>::value_type& empire_supply : empire_system_supply_ranges) {
             if (empire_supply.first == empire_id || empire_supply.first == ALL_EMPIRES)
+                continue;
+
+            if (Empires().GetDiplomaticStatus(empire_id, empire_supply.first) != DIPLO_WAR)
                 continue;
 
             for (const std::map<int, float>::value_type& supply_range : empire_supply.second) {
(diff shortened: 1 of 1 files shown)


Vezzra:
I'm very wary of this change. That *only one* empire can propagate supply into a system is a *core aspect* of the new supply mechanic, which removed the necessity to deal with the more complicated situation where two or more empires share a system.

However, this is actually a design issue that belongs on the forum, and needs to be discussed and resolved there. That should be done *before* we merge anything that changes the mechanic in such a profound way.

Therefore I propose to tag this PR as "work in progress" for now, until we reach a general consensus on how we want that whole supply thing (and the related ideas on stealth, distributed empires etc.) to work.

If the main concern is to have a quick solution for the problem that empires at peace still block each others supply, I'd rather enable shared supply in that case.

LGM-Doyle:
Absolutely.  That is why I marked it **changes the supply mechanic** and cross posted to the forum.  I've only provided this PR, as something concrete to discuss on the forums.  I didn't think to mark it "work in progress"

I'm happy with the supply mechanic as is and would rather it not be reverted.  

While looking at adding stealth to the supply resolution, as I proposed on the forum, I realized that the "at war" status was checked for fleet blocking, but not for the AI colony nerf.  I added that and to the per system resolution and tried a game.  

The game behavior seemed reasonable.  I could use my supply range to colonize within friendly empire's borders, but hostile empire's have solid borders.  I've pushed it so that others can compare the game play with the alternatives and decide for themselves.

Vezzra:
> and cross posted to the forum

Yeah sorry, I saw your forum post only *after* I responded here. What you stated there made my comment here more or less obsolete.

PR #1266, opened 2017-02-02
Correct ctor selection for delegated ctor - MultiIconValueIndicator

dbenage-cx:
Apple LLVM 6.1.0 fails to build with 1 error, 1 warning:
```
/Users/travis/build/freeorion/freeorion/UI/MultiIconValueIndicator.cpp:31:33: warning: too many braces around scalar initializer

/Users/travis/build/freeorion/freeorion/UI/MultiIconValueIndicator.cpp:31:5: error: constructor for 'MultiIconValueIndicator' creates a delegation cycle [-Wdelegating-ctor-cycles]
```

~~GCC 6.3.1 has a similar error and warning:~~
(Apparently, I can no longer reproduce despite `make clean && ccache -- clear`)

PR is a proposed solution to correct constructor selection of delegated constructor.

Needs testing at least for `MSVC`

 (created as branch on master to verify travis now builds)

### DIFF
--- a/UI/MultiIconValueIndicator.cpp
+++ b/UI/MultiIconValueIndicator.cpp
@@ -28,7 +28,7 @@ MultiIconValueIndicator::MultiIconValueIndicator(GG::X w) :
 
 MultiIconValueIndicator::MultiIconValueIndicator(GG::X w, int object_id,
                                                  const std::vector<std::pair<MeterType, MeterType>>& meter_types) :
-    MultiIconValueIndicator(w, {{object_id}}, meter_types)
+    MultiIconValueIndicator(w, std::vector<int>{object_id}, meter_types)
 {}
 
 MultiIconValueIndicator::MultiIconValueIndicator(GG::X w, const std::vector<int>& object_ids,


geoffthemedio:
builds OK for me with MSVC

UI/MultiIconValueIndicator.cpp
```diff
@@ -28,7 +28,7 @@ MultiIconValueIndicator::MultiIconValueIndicator(GG::X w) :
 
 MultiIconValueIndicator::MultiIconValueIndicator(GG::X w, int object_id,
                                                  const std::vector<std::pair<MeterType, MeterType>>& meter_types) :
-    MultiIconValueIndicator(w, {{object_id}}, meter_types)
+    MultiIconValueIndicator(w, std::vector<int>{object_id}, meter_types)
```
adrianbroher, line 31: ~~This calls the [fill constructor][1] for std::vector. You probably intended to write `std::vector<int>{1, object_id}`.~~

[1]: http://www.cplusplus.com/reference/vector/vector/vector/

adrianbroher (dismissed):

UI/MultiIconValueIndicator.cpp
adrianbroher: Hm, even this is wrong.  The rules regarding initializer list and uniform initialization are confusing…

Your version seems to be correct.  If a class provides a constructor with initializer lists it will prefer this over the interpretation as uniform initialization if the initializer list contains valid initialization values for a vector.

Vezzra:
Yep, that fixes an issue which breaks building on OSX. As it seems to meet approval and I need it to be able to build master I'm going ahead and merge.

PR #1268, opened 2017-02-02
Allied fleet supply

geoffthemedio:
Lets empires' ships resupply within allied empires' supply networks.

### DIFF
--- a/universe/Fleet.cpp
+++ b/universe/Fleet.cpp
@@ -21,7 +21,7 @@
 #include <cmath>
 
 namespace {
-    const bool ALLOW_ALLIED_SUPPLY = false;
+    const bool ALLOW_ALLIED_SUPPLY = true;
 
     const std::set<int> EMPTY_SET;
     const double MAX_SHIP_SPEED = 500.0;        // max allowed speed of ship movement


geoffthemedio:
Moved the changes into master, and reworked this pull request to just switch a flag that enables / disables the behaviour.

l29ah:
This plays badly with the peace by default. I guess there should be a "neutral" affiliation that doesn't attack you with military forces but doesn't resupply your ships.

geoffthemedio:
"by default" ?

l29ah:
When a multiplayer game starts you have peace with every human player.

Voker57:
Greatly simplifies allied military operations, merge plz

geoffthemedio:
@Vezzra @Dilvish-fo @Cjkjvfnby @Morlic-fo are there AI issues with enabling allied fleet supply?

Vezzra:
> are there AI issues with enabling allied fleet supply?

@geoffthemedio, *I* don't think so, but my assessment of that is probably the least qualified one compared to the others.

Vezzra:
> Greatly simplifies allied military operations, merge plz

We should probably address @l29ah's concerns before we do that. Either by introducing a third stance (as I mentioned elsewhere, I'd prefer to add an "Allied" stance and have that enable the "allied supply" mechanic), or making "War" the default stance (which btw would make the current workaround of having the AI declare war on everyone on turn 1 obsolete).

Morlic-fo:
> @Vezzra @Dilvish-fo @Cjkjvfnby @Morlic-fo are there AI issues with enabling allied fleet supply?

Probably not more than that the AI does not actively make use of it.

geoffthemedio:
Does the AI make decisions about where to send exploring ships by considering where fleets can be resupplied? If so, how does it determine where resupply is available?

Morlic-fo:
There are two kind of supply calculations in the AI. The one relevant for fleet movement does directly use a server provided set of systems:
https://github.com/freeorion/freeorion/blob/master/python/EmpireWrapper.cpp#L147

The other one uses some supply projection calculation which I forgot was migrated to the C++ side after the supply changes:
https://github.com/freeorion/freeorion/blob/master/python/EmpireWrapper.cpp#L101

So assuming the relevant functions provide the correct set of systems, the AI should understand the supply changes already.


That being said, the AI does not understand diplomacy at the moment and will consider allied fleets and systems as dangerous (i.e. the same as if they were at war with the empire) which will likely lead to the AI not using this feature actively.

geoffthemedio:
So there needs to be an allied (distinct from peace/neutral) diplomatic relationship then, into which the AI will never go.

Vezzra:
That looks like the best approach to me.

Vezzra:
@geoffthemedio, assuming you want to implement the additional "Allied" diplomatic relationship stance and tie allied supply to that before merging, do you want to get that into 0.4.7, or rather postpone it after the release?

geoffthemedio:
If I get it done fast we'll see but I wouldn't delay specifically for it.

Vezzra:
Ok, then I set it to "optional for 0.4.7".

geoffthemedio:
Pulled this pull request's changes into #1320

PR #1269, opened 2017-02-02
[AI] Cleanup

Cjkjvfnby:
Fix warnings highlighted by IDE.

@LGM-Doyle, I replaced import of executing module with explicit call of the method.  Unused imports are shown as error, and I want to avoid it, because fixing this "error" will break the game. That why I send you review request.

This PR should be merged with squash  and merge.  I can do it myself after get `approve` from reviewers.


LGM-Doyle:
@Cjkjvfnby that looks good to me.

It was like that originally.

Issue #1272, opened 2017-02-03
Crash when rendering MultiMeterStatusBar when values are zero


subpop:
### Environment

* **FreeOrion Version**: 0.4.6
* **Operating System**: Linux (Fedora 25)
* **Graphic card used**: Intel HD 530
* **Fetched as**: Compiled from source

### Description
I'm the package maintainer for freeorion in Fedora. I recently received a [crash report](https://bugzilla.redhat.com/show_bug.cgi?id=1416197) that looks like it's caused by an implicit cast from a signed int to an unsigned int (size_t). `length_error` is thrown by std::vector when `reserve` is called and the passed in `size_t` is larger than the size of the vector. That got me to inspect the values as computed inside `MultiMeterStatusBar.cpp` and noticed that when `num_segments` works out to be `0` (as is the case when all the resources of an Outpost on a planet are zero), when `bar_verts.reserve(num_segments - 1)` is called, the expression `num_segments - 1` evaluates to `-1`, but `reserve` implicitly casts this to an unsigned `size_t`. In the debugger, I was able to see these values are huge.

```
(lldb) p bar_verts.b_data.max_size()
(size_type) $5 = 4611686018427387903
(lldb) p (size_type) num_segments - 1
(unsigned long) $6 = 18446744073709551615
```

### Expected Result
The resource status bar to pull down.

### Actual Result
Crash, throwing `length_error`.

### Steps to reproduce
1. Build Outpost ship
2. Deploy the outpost on some planet
3. Click on pull-down icon in outpost detail window ( prod/research/trade/supply where all 4 values are 0 )

LGM-Doyle:
@subpop thank you for the perfect bug report.  Thanks also for packaging freeorion for Fedora.

Thanks to Peter "Pessoft" Kolínek and Link Dupont for additional details on bugzilla.

I've verified this bug.  It is a crash across all platforms, not just linux.

I've provided a fix in #1274.

Vezzra:
We've two different PRs providing a fix for this issue now: #1273 and #1274. #1274 looks more thorough to me, but that's only my opinion. What do the others say? Which one of the fixes should be merged?

@geoffthemedio?

BTW: which status label should be used for the rejected PR? Neither "follow up" nor "invalid" sounds really fitting, any ideas...?

LGM-Doyle:
I should withdraw mine, #1274.  I did not intend to post a second fix to a solved problem.  How about posting it as "duplicate".

[...]

PR #1274, opened 2017-02-03
Fix bug 1272: MultiMeterStatusBar crash with zero value.

LGM-Doyle:
This PR fixes issue #1272.

When the MultiMeterStatusBar received all zero values it would
initialize a vertex buffer with a signed int (-1) cast to a size_t,
causing a crash.

This commit forces the minimum number of segments to be 1 by
initializing largest_value with 0.1

I have tested that this patch works on both the release-0.4.6 and the master branches. @Vezzra, if this is merged then it is suitable for a hotfix patch to release-0.4.6, if we do that kind of thing.

### DIFF
--- a/UI/MultiMeterStatusBar.cpp
+++ b/UI/MultiMeterStatusBar.cpp
@@ -110,8 +110,9 @@ void MultiMeterStatusBar::Render() {
         y += BAR_HEIGHT + BAR_PAD;
     }
 
-    // Find the largest value to be displayed to determine the scale factor
-    double largest_value = 0;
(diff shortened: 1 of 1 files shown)


Vezzra:
> it is suitable for a hotfix patch to release-0.4.6, if we do that kind of thing

Most likely not, as we're too close to 0.4.7 already (which is scheduled for February/March), but still tagged it for cherry-picking to be on the safe side (we might reconsider after all).

geoffthemedio:
looks fine

PR #1282, opened 2017-02-05
Organize Tech articles into sub-categories

dbenage-cx:
[Forum thread](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10343)
Some further sub-categories were added to aid in finding content.


geoffthemedio:
seems fine after cursory test

PR #1286, opened 2017-02-06
Remove boost::gil fork from GG code

adrianbroher:
No need to maintain a fork if the FreeOrion specific changes can be used with the upstream boost::gil library without great effort.

I would appreciate it if this change could be tested on Windows and Mac because both build and texture loading may be broken by this change on those platforms.


adrianbroher:
With the CI now finally working properly there is at least one bug on MacOSX that needs to be fixed before I need to bother other people with review requests.

geoffthemedio:
Builds and seems to run OK on Windows.

Vezzra (approved):
Builds and runs fine on OSX. LGTM. 👍

adrianbroher:
Thanks for the review.

PR #1287, opened 2017-02-06
Fix bug 1237: Monsters not blockading correctly.

LGM-Doyle:
This commit fixes UpdateMonsterTravelRestrictions() in ServerApp.cpp.

The problem was that recent code changes caused the loop to loop over
all of the monsters in the galaxy instead of just monsters in one
system.

This commit enforces the following behavior on a per system basis:

1. If any unrestricted (blockading) empire fleets are present, the
monster's  arrival starlane is set to INVALID_OBJECT, so that they can't
leave the system.

2. If any monsters are blockading and any empire fleets are present, the
monster's arrival starlane is set to INVALID_OBJECT, so that they can't
maintain the blockade.

It was tricky to test, so someone else should verify that it is working.

Thanks.

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -2761,31 +2761,53 @@ void ServerApp::ProcessCombats() {
 }
 
 void ServerApp::UpdateMonsterTravelRestrictions() {
-    //std::vector<Fleet*> all_fleets =  m_universe.Objects().FindObjects<Fleet>;
-    for (std::shared_ptr<const System> system : m_universe.Objects().FindObjects<System>()) {
+    for (auto const &maybe_system : m_universe.Objects().ExistingSystems()) {
+        std::shared_ptr<const System> system = std::dynamic_pointer_cast<const System>(maybe_system.second);
+        if (!system) {
+            ErrorLogger() << "Non System object in ExistingSystems with id = " << maybe_system.second->ID();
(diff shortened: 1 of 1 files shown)


geoffthemedio:
> recent code changes

Do you know a specific change that caused the issue?

LGM-Doyle:
No.

There were several contributing changes, which would have caused different variations of the problematic behavior.    The observed "good"  behavior may have been an artefact of code outside of this function. I only did enough digging to determine that reverting any one commit wouldn't fix the problem.

Here is a summary of the changes to this function:
- The outer for loop was changed to `for ( begin() to begin()` effectively commenting out all of the code.  If the game was behaving correctly at this time then it was due to code elsewhere.  I tried restoring this commenting out, but it did not fix the problem, so the code elsewhere has been changed.
- I "fixed" the outer for loop to be `for (begin() to end())`.  
- Both loops were changed to C++11 style.  The inner loop here was now looping over all of the fleets in the galaxy, not all of the fleets in the system.  I tried just restoring looping over just system fleets, but this did not fix the problems.

Since reverting some/all of these changes did not fix the issue, there are other changes to mechanics contributing to the issue.  I don't think these are "problems" we need to search for or keep the issue open for, because the result was more regular behavior between monsters and normal fleets.

PR #1290, opened 2017-02-07
Reset ready status for players with same chosen color

### LINKED ISSUE #1284: Multiplayer game can start when players have same color

dbenage-cx:
Fixes #1284

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -644,6 +644,11 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
 
     static int AI_count = 1;
     static int nameless_player_count = 1;
+    std::map<GG::Clr, int> psd_colors;
+    // std::map<std::string, int> psd_player_names;
+    std::map<std::string, int> psd_empire_names;
+    std::set<int> conflicted_psd_ids;
+    const GG::Clr CLR_NONE = GG::Clr(0, 0, 0, 0);
 
     // assign unique names / colours to any lobby entry that lacks them, or
     // remove empire / colours from observers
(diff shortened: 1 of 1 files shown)


geoffthemedio:
There was/is a general https://github.com/freeorion/freeorion/blob/master/UI/MultiplayerLobbyWnd.cpp#L959 PlayerDataAcceptable() function that you should probably replicate all the functionality of.

dbenage-cx:
Rebased with added checks for unique player and empire names

server/ServerFSM.cpp
```diff
@@ -688,8 +693,34 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
...
         }
+
+        if (psd.m_empire_color != CLR_NONE) {
+            if (!psd_colors.emplace(psd.m_empire_color, entry.first).second) {
+                conflicted_psd_ids.insert(psd_colors.at(psd.m_empire_color));
+                conflicted_psd_ids.insert(entry.first);
+            }
+        }
+
+        if (!psd.m_player_name.empty()) {
```
geoffthemedio: requiring player names (not just empire names) to be unique is problematic because you can't change your player name after joining a server

dbenage-cx: Left the code in place but commented out due to the TODO mentioned at ln 467 (check if player name is unique)

server/ServerFSM.cpp
```diff
@@ -652,7 +657,7 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
         if (psd.m_client_type == Networking::CLIENT_TYPE_HUMAN_OBSERVER ||
             psd.m_client_type == Networking::CLIENT_TYPE_HUMAN_MODERATOR)
         {
-            psd.m_empire_color = GG::Clr(0, 0, 0, 0);
+            psd.m_empire_color = CLR_NONE;
```
geoffthemedio, line 660: there is a GG::CLR_ZERO you should probably be using here, but in a subsequent commit is fine

dbenage-cx: Is it worth the additional include?

geoffthemedio: unsure... it's not a huge header file: https://github.com/freeorion/freeorion/blob/master/GG/GG/ClrConstants.h

dbenage-cx: Will do, getting link errors after changing so doing a clean build.

dbenage-cx: Should have checked CMakeLists first...server does not link to GiGi.  Propose to leave as is.

geoffthemedio: ok

PR #1294, opened 2017-02-07
Replace get_systems_within_jumps() from galaxy.py

LGM-Doyle:
This PR replaces `get_systems_within_jumps()` with a server provided faster version based on the `Pathfinder` cache.

I mentioned this in PR #1225 about improving the galaxy generation.  The improvement without #1225 is marginal.


python/server/ServerWrapper.cpp
```diff
@@ -1291,6 +1310,8 @@ namespace FreeOrionPython {
         def("create_field",                         CreateField);
         def("create_field_in_system",               CreateFieldInSystem);
 
+        def("systems_within_jumps",                 SystemsWithinJumps);
```
genestack-solomatin: You can add documentation string as third argument here.

python/server/ServerWrapper.cpp
```diff
@@ -905,6 +905,25 @@ namespace {
...
+
+        std::vector<int> systems;
+
+        for (boost::python::stl_input_iterator<int> id(sys_ids); id != end; ++id) {
+            systems.push_back(*id);
+        }
+
+        auto near = GetPathfinder()->WithinJumps(jumps, systems);
+
+        for (auto ii : near) {
```
adrianbroher: ii?

python/server/ServerWrapper.cpp
```diff
@@ -905,6 +905,25 @@ namespace {
...
+        list py_systems;
+        boost::python::stl_input_iterator<int> end;
+
+        std::vector<int> systems;
+
+        for (boost::python::stl_input_iterator<int> id(sys_ids); id != end; ++id) {
+            systems.push_back(*id);
+        }
+
+        auto near = GetPathfinder()->WithinJumps(jumps, systems);
```
adrianbroher: near?

universe/Pathfinder.cpp
```diff
@@ -1010,6 +1015,51 @@ std::multimap<double, int> Pathfinder::PathfinderImpl::ImmediateNeighbors(int sy
 }
 
 
+void Pathfinder::PathfinderImpl::WithinJumpsCacheHit(
+    std::unordered_set<int>* result, size_t jump_limit,
+    size_t ii, const distance_matrix_storage<short>::row_ref row) const {
+
+    // Scan the LUT of system ids and add any result from the row within
+    // the neighborhood range to the results.
+    for (auto it : m_system_id_to_graph_index) {
```
adrianbroher: it?

universe/Pathfinder.cpp
```diff
@@ -1010,6 +1015,51 @@ std::multimap<double, int> Pathfinder::PathfinderImpl::ImmediateNeighbors(int sy
 }
 
 
+void Pathfinder::PathfinderImpl::WithinJumpsCacheHit(
+    std::unordered_set<int>* result, size_t jump_limit,
+    size_t ii, const distance_matrix_storage<short>::row_ref row) const {
```
adrianbroher, line 1020: ii?

adrianbroher (changes requested):
You should really reconsider your general variable naming style.

universe/Pathfinder.cpp
geoffthemedio: not that unusual for an iterator type in FO code...

adrianbroher: True, but range based for loops don't use the iterator directly but the type that the iterator dereferences to.

Vezzra (approved):
Can't comment on the C++ side of things, as your implementation goes over my head, but moving this into the backend code is certainly a good idea.

Everything else looks good to me, builds and runs ok on OSX. LGTM (as far as I'm concerned).

Cjkjvfnby:
> but moving this into the backend code is certainly a good idea
@Vezzra, why?

Python is slower when C++, but not so slow.  I expect that using cache should give more benefits then changing language. 

Code in python usually more readable.

python/server/ServerWrapper.cpp
LGM-Doyle: Thanks.

python/server/ServerWrapper.cpp
LGM-Doyle: Could you provide suggestions?

I've found that no matter what names I choose someone finds them too terse, too verbose, too cryptic or all of the above.

universe/Pathfinder.cpp
LGM-Doyle: That is the row index, typically written  `i`.  However `ii` is much easier to grep.  This one I'd like to keep.

LGM-Doyle (commented):
This is backend vs frontend issue.  

The backend can access all of the information to compute this result in a single vector.  The provided function `fo.systems_within_jumps(range, [systems])` is an intuitive  addition to the ServerWrapper.  The best place to do the calculation is where the data is.

python/server/ServerWrapper.cpp
adrianbroher: After looking into the other changes why not use 'systems_in_vicinity'?  It was already a good name and the documentation for the python `get_systems_within_jumps` just provided the information I am missing out at `WithinJumps` so it should be transferred and adapted accordingly.

> I've found that no matter what names I choose someone finds them too terse, too verbose, too cryptic or all of the above.

That's the outcome of cooperative development. Everybody has to deal with different opinions. ;)  In genreal I will always complain about terse or generalized names if the context doesn't provide any information about the code usage. In this case `near` just didn't ring a bell.

python/server/ServerWrapper.cpp
adrianbroher: `system_id` would probably better.

Vezzra:
> Python is slower when C++, but not so slow.

IMO the speed advantage C++ offers is still significant enough to justify moving the implementation only for that (and then there are the reasons @LGM-Doyle cited).

> I expect that using cache should give more benefits then changing language.

Sure, but that's no reason not to take advantage of the additional performance boost the change to C++ offers.

> Code in python usually more readable.

Yes, but that's not really relevant in case of this function. Any speed gain, even if it's only a moderate one, is more important than the not quite as good readability.

LGM-Doyle (commented):
I think that I have addressed the concerns.
I've also rebased it and used it to replace uses of get_systems_within_jumps() in the newly merged code from #1225.

Vezzra:
Just noticed a minor thing: you did not remove the now obsolete `galaxy_topology.py`. Any reason for this?

adrianbroher (approved):

LGM-Doyle:
Fixed.

The rebasing removed the original commit that removed the file.

PR #1300, opened 2017-02-09
Split Log Files!

### LINKED ISSUE #1276: Split Log Files?
In big long games, or particularly when verbose logging is enabled, the .log files generated by the game can grow very big very fast. It might be a good idea to start saving to a new log file after the current one gets excessively big... 100 MB or so, perhaps?

Some mechanism to limit the total size of the generated log files might also be good, although it's rarely a concern when not using verbose logging, which is specifically labelled as for debugging purposes.

dbenage-cx:
PR provides configurable rotation and archiving of log files.

When logs are initialized, any existing logs are move to the archive directory.
These archived files are appended with a timestamp to retain a unique name.
Timestamp matches the last write time (e.g. `AI_1-1486641524.log`)
If there are more files for a process than the config setting allows, old files are removed.

During a game, log files are written to a process specific file with a `.tmp` extension.
Once the rotation limit is reached it is appended to a process specific `.log` file.
(e.g. `logs\AI_1.tmp` and `logs\AI_1.log`)

Current logs and archived logs are in separate directories to ease the burden of sending logs.
With this layout a user could compress the `log` folder for upload elsewhere.

One removal was of the process name from each log line.  This is simple to add back if needed.
freeorion and freeoriond were given more user friendly names of Client and Server.

In a worse case, if the temporary files are not cleaned up by the collector (e.g. terminated client), the archived files may have duplicate info appended to the end.

Resolves #1276

For testing recommend: `--log-level TRACE --logging.rotate-mb 1 --logging.archive-max-files 2`

### DIFF
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -78,12 +78,11 @@ AIClientApp::AIClientApp(const std::vector<std::string>& args) :
     // read command line args
 
     m_player_name = args.at(1);
-    const std::string AICLIENT_LOG_FILENAME((GetUserDataDir() / (m_player_name + ".log")).string());
     if (args.size() >=3) {
         m_max_aggression = boost::lexical_cast<int>(args.at(2));
     }
 
-    InitLogger(AICLIENT_LOG_FILENAME, "AI");
+    InitLogger(m_player_name);
     DebugLogger() << PlayerName() + " ai client initialized.";
 }
 
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -202,9 +202,8 @@ HumanClientApp::HumanClientApp(int width, int height, bool calculate_fps, const 
 #endif
     m_fsm.reset(new HumanClientFSM(*this));
 
-    const std::string HUMAN_CLIENT_LOG_FILENAME((GetUserDataDir() / "freeorion.log").string());
-
-    InitLogger(HUMAN_CLIENT_LOG_FILENAME, "Client");
+    // TODO Add an increment here to support multiple human clients (hotseat)
+    InitLogger("HumanClient");
 
     try {
         InfoLogger() << "GL Version String: " << GetGLVersionString();
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1432,6 +1432,12 @@ Sets the root directory for the game resource files (game content and data files
 OPTIONS_DB_LOG_LEVEL
 Sets the level at or above which log messages will be output (levels in order of increasing verbosity: FATAL, ERROR, WARN, INFO, DEBUG, TRACE)
 
+OPTIONS_DB_LOG_DIR
+Directory to store log files.
+
+OPTIONS_DB_LOG_ROTATE_SIZE
+Approximate maximum size (in MegaBytes) for log files.
+
 OPTIONS_DB_STRINGTABLE_FILENAME
 Sets the language-specific string table filename.
 
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -138,9 +138,7 @@ ServerApp::ServerApp() :
     m_current_turn(INVALID_GAME_TURN),
     m_single_player_game(false)
 {
-    const std::string SERVER_LOG_FILENAME((GetUserDataDir() / "freeoriond.log").string());
-
-    InitLogger(SERVER_LOG_FILENAME, "Server");
+    InitLogger("Server");
 
     LogDependencyVersions();
 
--- a/util/Logger.cpp
+++ b/util/Logger.cpp
@@ -14,11 +14,15 @@
 #include <boost/log/utility/setup/formatter_parser.hpp>
 #include <boost/log/utility/setup/filter_parser.hpp>
 #include <boost/log/support/date_time.hpp>
+#include <boost/filesystem/path.hpp>
+#include <chrono>
+#include <boost/filesystem/operations.hpp>
 
 namespace logging = boost::log;
 namespace expr = boost::log::expressions;
 namespace attr = boost::log::attributes;
 namespace keywords = boost::log::keywords;
+namespace fs = boost::filesystem;
 
 
 namespace {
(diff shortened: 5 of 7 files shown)


geoffthemedio:
Just "Client" is ambiguous, as the AIs are also clients. "HumanClient" is preferable.

I'd rather have immediate log files keep the .log extension. Expect unexpected terminations.

dbenage-cx:
Changed to HumanClient
Immediate logs changed to a mask ending in _part.log

dbenage-cx:
Rebased, utilizing custom file collector.

geoffthemedio:
````
C:\Users\Geoff\Desktop\FreeOrionSDK3\FreeOrion>FreeOrion.exe
[2017-02-12 11:04:42.226250] [0x00004f80] [error]   main() caught exception(std::runtime_error): boost::filesystem::directory_iterator::construct: The system cannot find the path specified: "C:\Users\
Geoff\AppData\Roaming\FreeOrion\old_logs"
main() caught exception(std::runtime_error): boost::filesystem::directory_iterator::construct: The system cannot find the path specified: "C:\Users\Geoff\AppData\Roaming\FreeOrion\old_logs"
main() failed to setup or run SDL.
````

dbenage-cx:
Sorry about that, did not check early enough.

geoffthemedio:
This doesn't seem to work, and I'm not sure how to describe how...

When I run a game, I get a bunch of .log and _part.log files generated in the logs directory. The sizes of these don't make much sense, and I don't understand why there are _part files at all.

But more importantly, when I run freeorion.exe again after exiting, most but not all of the files in logs disappear, but nothing appears in old_logs

logs

I didn't follow all the description in the initial post, but it sounds overly complicated and fragile, perhaps leading to some of what I'm seeing here?

Also, when I exit from the main menu or use the top-right X to close the window, I get a popup saying FreeOrion.exe has stopped working. I'm not getting that with master.

dbenage-cx:
TY for taking a look, will see if I can resolve issues with MSVC so I can look into no archived files and error message.

To try and summarize:
*  log entries are written to `_part`
*  when `_part` roughly exceeds the size limit, it is appended to the log file and removed
*  on initialization and deconstruction, log files are moved to `old_logs`

Some way to distinguish partial logs from normal logs is needed if they are to reside in the same directory. I assumed the behavior of the default file collector would not be suitable (chiefly from needing to manually figure which log files belonged to which session).

~~In your screenshot, you should have a `_part` file for each process (if the game is running), possibly the windows file manager is not refreshing?~~ Edit: The existing log files remain because those processes were killed, they will be cleaned up once a process for them starts again.

geoffthemedio:
I don't see a point to this change then. The idea was to switch to a new log file when the current one was too big, so individual logs files don't get so huge. You're almost doing the opposite - moving contents of smaller log files into a single bigger one when the small one gets big... which doesn't seem to accomplish much except making things more complicated.

What I'd envisioned was more like having a HumanClient-1-timestamp.log and then switching the output to HumanClient-2-timestamp.log once the -1 file was over some size limit. Then delete the oldest ones if there are more than X log files for a process.

dbenage-cx:
Ok, that shouldn't need the custom collector then.  I will swap this branch out for a simple version.

dbenage-cx:
Reset to the previous file collector, (archived previous work [here](https://github.com/dbenage-cx/freeorion/tree/defunct_custom_loggin_sink))

`max_files` first appears in boost 1.61, so there is no limit set.
The other filters available in 1.59 are time based rotation and min free space.

dbenage-cx:
Not certain how to proceed here.
I don't know how to resolve excess file removal in a method external to the logger in a sane way.
Closing as invalid so it will not conflict with other work.

LGM-Doyle:
I noticed you cherry picked the warning fix to master and came here to say thanks.  Thanks.

Then I noticed that the PR had disappeared.  That's unfortunate.  If you are still motivated to work on this then please do so, my logging changes have been pending for most of a year, I can wait a little longer.

@geoffthemedio, I skimmed the PR earlier and also assumed it was working the way I imagined, which was a third method different from either of your approaches.

I assumed that the tail of the log would be kept in the original file.  The excess log would be moved to the new file.  Then refreshing the original log shows the most current error messages.

An easy way to do this might be to rename the log file when it exceeds the threshold, start a new file with the same original name.  I wouldn't timestamp anything until moved to the old_logs directory, because it is easier to script log parsers as the file names are stable.

A problem that I see with this approach is that immediately after the rollover the log file will be truncated to zero length and not useful.  I am uncertain of a fast solution to this that doesn't involve file/string handling that will slow the game.

For example, for "freeorion.log" and a file threshold of 20MB, and an old_logs threshold of 50MB a history might be:

- 10MB after startup all of the log is in the main log file
-- (freeorion.log, 10MB) (old_logs/ 0MB)
-  21MB after startup, 1MB after first overflow. The tail 1MB of log is in freeorion.log.  The original log has been renamed to freeorion.log.1. 
-- (freeorion.log, 1MB) (freeorion.log.1, 20MB) (old_logs/ 0MB)
- 43MB at crash 
-- (freeorion.log, 3MB) (freeorion.log.1, 20MB) (freeorion.log.2, 20MB) (old_logs/ 0MB)
- 22 MB after 1st restart.  The original logs have been moved to old_logs and timestamped since they are under the threshold.
-- (freeorion.log, 2MB) (freeorion.log.1, 20MB) (old_logs/ 43MB) (old_logs/freeorion-timestamp1.log, 3MB) (old_logs/freeorion-timestamp1.log.1, 20MB) (old_logs/freeorion-timestamp1.log.2, 20MB) 
- 7MB after 3rd start. Logs from the second start are timestamped and moved to old_logs.  All logs from the first startup with the timestamp1 are removed to preserve old_logs max size of 50MB.
-- (freeorion.log, 7MB) (old_logs/ 22MB) (old_logs/freeorion-timestamp2.log, 2MB) (old_logs/freeorion-timestamp2.log.1, 20MB)

geoffthemedio:
> A problem that I see with this approach is that immediately after the rollover the log file will be truncated to zero length and not useful.

As long as the logged info is retained and accessible, I don't think this is a major problem.

PR #1302, opened 2017-02-10
Correct structure check for unowned planet mines

### LINKED ISSUE #1283: Mines can't destroy ships
Version: 387a65d513bf379ae

Save: http://dump.bitcheese.net/files/gukaseh/undying.mps

Voker57 empire. At the end of turn, ship at Rheims 3 will die and resurrect.

dbenage-cx:
Fixes #1283

### DIFF
--- a/default/scripting/techs/defense/mines.macros
+++ b/default/scripting/techs/defense/mines.macros
@@ -80,7 +80,7 @@ EG_SYSTEM_MINES
                     Not HasTag name = "UNOWNED_FRIENDLY"
                     ]
                 ]
-                Structure low = @1@ + 0.0001
+                Structure high = @1@
             ]
             activation = Unowned
             stackinggroup = "DEF_MINES_DAMAGE_STACK"


default/scripting/techs/defense/mines.macros
```diff
@@ -80,7 +80,7 @@ EG_SYSTEM_MINES
                     Not HasTag name = "UNOWNED_FRIENDLY"
                     ]
                 ]
-                Structure low = @1@ + 0.0001
+                Structure high = @1@ + 0.0001
```
Dilvish-fo: seems to me that this line should be the same as line 63 in the empire-owned planet case,
``` Structure high = @1@```

dbenage-cx: TY, fixed

Dilvish-fo:
The underlying problem is rarely enough encountered that unless someone had a saved game to test this against (other than the one posted with the cited issue that you already have access to), or was willing to put time into engineering up a test case, that any additional testing of this is not likely to occur.  The fix is clear and straightforward enough that I would suggest it is ready to merge.

Vezzra:
@dbenage-cx, I can't open the provided savegame, neither on OSX nor on Linux, so I can't test it, but assuming you did test it on your system I agree with @Dilvish-fo. As far as I'm concerned, go ahead and merge.

dbenage-cx:
Ok, since it is last in the order skipping extensive testing for edge cases.

Issue #1244, opened 2017-01-26
Replace boost::lexical_cast<std::string> with std::to_string for fundamental types


adrianbroher:
Within the whole codebase we use `std::lexical_cast<std::string>` to convert fundamental types to an std::string.  Replacing those instances with a call to `std::to_string` would make the code much more readable.

However keep in mind that `std::to_string` only works for fundamental numeric types, not for custom enumerations.

Executing `grep -rn 'boost::lexical_cast<std::string>' *` should list a good list of candidates.

Vezzra:
@adrianbroher, @dbenage-cx, with #1303 merged, can this issue be considered resolved and closed?

adrianbroher:
After reviewing latest master I consider this task realized.  There are still some places missing but those involve custom types and also there is no std::string to numeral, but those cases were never in the scope of the original issue.

PR #1303, opened 2017-02-10
Remove lexical cast to std::string for fundamental types

dbenage-cx:
PR scope is to resolve remainder of #1244

### DIFF
--- a/universe/Tech.cpp
+++ b/universe/Tech.cpp
@@ -12,7 +12,6 @@
 #include "ValueRef.h"
 #include "Enums.h"
 
-#include <boost/lexical_cast.hpp>
 #include <boost/filesystem/fstream.hpp>
 
 namespace {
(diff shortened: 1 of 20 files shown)


universe/Tech.cpp
```diff
@@ -376,8 +373,7 @@ TechManager::TechManager() {
 #ifdef OUTPUT_TECH_LIST
     for (const Tech* tech : m_techs.get<NameIndex>()) {
         std::cerr << UserString(tech->Name()) << " (" 
-                  << UserString(tech->Category()) << " "
-                  << UserString(boost::lexical_cast<std::string>(tech->Type())) << ") - "
```
geoffthemedio, line 380: Appears to remove the type output...

dbenage-cx: Should have been a separate commit, `Tech` no longer has a member `Type`

adrianbroher: @geoffthemedio you removed the tech type (theory vs application) two year ago (or so).

PR #1304, opened 2017-02-12
Adapt AI to SH_SPATIAL_FLUX changes

### LINKED ISSUE #1111: Update AI for Spatial Flux Hull
### Description
With f61e3c4 I've increased the costs of the Spatial Flux to be more balanced and reduced the fuel from 6 to 4, in addition I've given it some stealth abilities that are state dependent.

As far as I can see the AI dependencies list simply had a blunt -30 stealth stat, which was always wrong, it's now a fair bit more complex and I suspect the AI will need some minor tweaking that is nevertheless outside my ability.

Morlic-fo:
This PR also extends the hardcoded effects scripting system of the ShipDesignAI to some ad-hoc implementation of conditionals needed to handle the spatial flux hulls effects.

Resolves #1111.

### DIFF
--- a/default/python/AI/AIDependencies.py
+++ b/default/python/AI/AIDependencies.py
@@ -631,6 +631,9 @@ PART_KRILL_SPAWNER = "SP_KRILL_SPAWNER"
 
 # <editor-fold desc="Effect Scripting for Shipdesigns">
 # <editor-fold desc="Tokens">
+# Syntax tokens
+IF = "IF"
+
 # known tokens the AI can handle
 REPAIR_PER_TURN = "REPAIR_PER_TURN"
 FUEL_PER_TURN = "FUEL_PER_TURN"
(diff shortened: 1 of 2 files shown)


default/python/AI/AIDependencies.py
```diff
@@ -668,7 +672,13 @@
         REPAIR_PER_TURN: 2,
     },
     "SH_SPATIAL_FLUX": {
-        STEALTH_MODIFIER: -30,
+        STEALTH_MODIFIER: [
```
Cjkjvfnby, line 675: I don't understand what does it mean.  Looks like error and unnecessary code because it always 10. 

This should be easy to understand without any comments.  Please don't add magic to code. 

Also I suggest to do next steps:
- move parsing to separate module
- make tests (freeorion\default\python\tests)

After what we can back to review.

PS. We already have some new language in research AI:  it works on function.  Maybe we can reuse it?

https://github.com/freeorion/freeorion/blob/master/default/python/AI/ResearchAI.py#L356

Cjkjvfnby (changes requested):

adrianbroher:
Oh no.  Not another custom brewed DSL.

PR #1305, opened 2017-02-12
Adjust invasion and colonization priority for non-standard species

### LINKED ISSUE #1012: Update AI to take into account Ancient Guardians
### Description

As of #1009 Ancient Guardians are a species that self destruct on conquest, the AI doesn't know how they work and in testing will deploy large fleets of troops to capture them without a backup warship to keep shields down. They have the tag "DESTROYED_ON_CONQUEST" as per @Morlic-fo's request.

### LINKED ISSUE #1058: AI needs to take into account Lembala Lem special features
### Description

Species introduced in #1016 has non-standard population effects and grants some tech bonuses, AI needs to take these into account when setting invasion priorities and similar.

Morlic-fo:
The PR

- handles fixed size populations when assigning colonization values
- increases invasion priority if a tech will be unlocked by conquering the planet
- implements the relevant settings for Lembala'Lam species.
- adjusts invasion priority if a species is destroyed on conquest (e.g. Ancient Guardians)

Resolves #1012
Resolves #1058

### DIFF
--- a/default/python/AI/InvasionAI.py
+++ b/default/python/AI/InvasionAI.py
@@ -284,6 +284,12 @@ def evaluate_invasion_planet(planet_id, empire, secure_fleet_missions, verbose=T
         bld_tally += bval
         detail.append("%s: %d" % (bldType, bval))
 
+    tech_tally = 0
+    for unlocked_tech in AIDependencies.SPECIES_TECH_UNLOCKS.get(species_name, []):
+        if not tech_is_complete(unlocked_tech):
+            rp_cost = fo.getTech(unlocked_tech).researchCost(empire_id)
+            tech_tally += rp_cost * 10
+
     p_sys_id = planet.systemID
     capitol_id = PlanetUtilsAI.get_capital()
     least_jumps_path = []
@@ -367,7 +373,8 @@ def evaluate_invasion_planet(planet_id, empire, secure_fleet_missions, verbose=T
         troop_cost = math.ceil(planned_troops/6.0) * (40*(1+foAI.foAIstate.shipCount * AIDependencies.SHIP_UPKEEP))
     else:
         troop_cost = math.ceil(planned_troops/6.0) * (20*(1+foAI.foAIstate.shipCount * AIDependencies.SHIP_UPKEEP))
-    planet_score = retaliation_risk_factor(planet.owner) * threat_factor * max(0, pop_val+supply_val+bld_tally+enemy_val-0.8*troop_cost)
+    base_score = pop_val + supply_val + bld_tally + tech_tally + enemy_val - 0.8*troop_cost
+    planet_score = retaliation_risk_factor(planet.owner) * threat_factor * max(0, base_score)
     if clear_path:
         planet_score *= 1.5
     if verbose:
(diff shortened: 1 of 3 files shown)


default/python/AI/InvasionAI.py
```diff
@@ -284,6 +286,12 @@ def evaluate_invasion_planet(planet_id, empire, secure_fleet_missions, verbose=T
         bld_tally += bval
         detail.append("%s: %d" % (bldType, bval))
 
+    tech_tally = 0
+    for unlocked_tech in AIDependencies.SPECIES_TECH_UNLOCKS.get(species_name, []):
+        if not tech_is_complete(unlocked_tech):
+            rp_cost = fo.getTech(unlocked_tech).researchCost(empire_id)
+            tech_tally += rp_cost * 10
```
Dilvish-fo: Valuing the tech at 10 times research cost seems a bit high to me.  Generally I was valuing research points at 2-3 PP each, and so it seems to me that a similar multiplier should be used here.

Dilvish-fo: I'll leave the coding-structure review to CJ, but content-wise, this looks really good to me.

default/python/AI/InvasionAI.py
```diff
@@ -367,7 +375,8 @@ def evaluate_invasion_planet(planet_id, empire, secure_fleet_missions, verbose=T
         troop_cost = math.ceil(planned_troops/6.0) * (40*(1+foAI.foAIstate.shipCount * AIDependencies.SHIP_UPKEEP))
     else:
         troop_cost = math.ceil(planned_troops/6.0) * (20*(1+foAI.foAIstate.shipCount * AIDependencies.SHIP_UPKEEP))
-    planet_score = retaliation_risk_factor(planet.owner) * threat_factor * max(0, pop_val+supply_val+bld_tally+enemy_val-0.8*troop_cost)
+    base_score = pop_val + supply_val + bld_tally + tech_tally + enemy_val - 0.8*troop_cost
```
Cjkjvfnby, line 379: Did I missed or there is not logical changes?   Please extract it as separate commit and merge it to master.  No need to mix feature and cleanups.

default/python/AI/InvasionAI.py
Morlic-fo: > Valuing the tech at 10 times research cost seems a bit high to me. Generally I was valuing research points at 2-3 PP each, and so it seems to me that a similar multiplier should be used here.

I didn't think of the value in terms of PP here but rather looked at the order of magnitude of the building values (https://github.com/freeorion/freeorion/pull/1305/files/5936cc2de6811fdbd62ae64353e6fbb2e91f3797#diff-9018f67910952aba310b8885e70cb762R231). 

Arguably, taking buildings away from the opponent adds another advantage. So maybe the ratio should be lower as you suggest. Then again, getting a tech immediately offers a power spike to the empire and seems to be more worth than the same number of RP cumulated over x turns.

Cjkjvfnby: Code structure look OK.  We only extend length of the `evaluate_planet` by 8%.

Cjkjvfnby (approved):

default/python/AI/InvasionAI.py
Morlic-fo: The ```tech_tally``` was added here. That makes the line awkwardly long.

Instead of going for a 2-line-formula which really does not make things more readable, I prefered this approach.

default/python/AI/InvasionAI.py
Morlic-fo: > We only extend length of the evaluate_planet by 8%.

The function definitely should be refactored at some point. It is way too long, has an extremely long if-else block and seems to redefine/reuse some variables from some hundred lines before... Not today and in this PR though.

Morlic-fo:
Rebased and weighted the techs with three times their RP cost.

Issue #1183, opened 2016-12-20
Scrollbar missing on focus dropdown


MatGB:
### Environment

* **FreeOrion Version**: b7f04f3
* **Operating System**: Linux
  * Compiled from source

### Description
If a planet has multiple focus options (eg through a Transformer) and is focused on an option at the bottom of a list, on opening the sidepanel with the need to scroll *up* (ie to Industry), there is no scrollbar. Selecting Research, closing the sidepanel then reopening will then display industry. If set to Industry the scrollbar appears correctly to allow a scroll down.

The scrollbar, as far as I can tell, is only missing if, on opening the sidepanel, the selected focus is low enough on the list to normally trigger a need to scroll upwards.
![screenshot_20161220_150757](https://cloud.githubusercontent.com/assets/11533206/21355642/b8f13d52-c6c6-11e6-807a-45dbef630b3b.png)

MatGB:
OK, this has now gone, so closing.

Vezzra:
So, basically, the issue disappeared for no obvious reason...?

MatGB:
And it's back with master compiled at d9352da, if you do scroll up and click select the display doesn't change as it should but the top focus is selected, it updates to show this when the UI updates the sidepanel registering the change.

As the preferred focus is almost always Industry it's a valid workaround but it's not good. As it appears to be intermittent and isn't always happening in the current game I'm not sure how to test further.

Dilvish-fo:
I think the issue is not so much that a scrollbar is missing, but that the first item is getting omitted.  Note in the screenshot above that there is a blank spot at the bottom of the list.  Also, this is very easy to replicate for me, without needing any advanced techs like stargate -- for a planet with the typical set of initial focii, set the focus to Defense (the last one) and then advance a turn, and then go to change it back:

![humancolonymissingproductionfocuschoice](https://cloud.githubusercontent.com/assets/10995939/22799118/70ad1d66-eeb9-11e6-8719-e329afee725b.png)

Vezzra:
> I think the issue is not so much that a scrollbar is missing, but that the first item is getting omitted.

You mean, a focus setting that should be there (industry?) is missing? *That* would actually be a very serious bug.

The blank line however has been around since forever I think...

LGM-Doyle:
Thank you @Dilvish-fo for the straightforward replicator.

geoffthemedio:
[...]

Issue #1196, opened 2017-01-04
Species dropdown list is drawn out of screen


l29ah:
![screenshot](https://dump.bitcheese.net/files/bilaseg/screenshot-freeorion-off-screen-dropdown.png)

Vezzra:
Should be sufficiently addressed by #1183.

l29ah:
No // 29d1da701b3f6967915f178710e43ae9f03ad436 doesn't fix it.

Vezzra:
No change at all, or what exactly is the issue that remains?

l29ah:
Species dropdown list is drawn out of screen.

Vezzra:
So, basically, #1183 didn't change anything at all. Well, in this case we need to reopen.

@LGM-Doyle, any ideas? I assume the fix you provided solved the problem on your system...

LGM-Doyle:
To be clear, when it is working correctly the drop down list has at least one row and no more rows than can fit on the available screen space.  It has a scroll bar and keyboard navigation and selection works.  If there is not room for even one row, then one  row is still drawn and this row will extend offscreen.  

@Vezzra I'm assuming that you have checked OSX and it works for you.

I'm guessing the whatever window manager is being used is not reporting screen size in a way that the app understands.

@l29ah, I'll ask the standard list of questions.
- Are you running a clean build of the latest master?
- Does the change to the SideBar focus drop down work for you?  If you shrink your side panel and drag it to the bottom of the screen does the focus drop down clip itself to the screen edge?
- What OS and version are you using?
- If you are not using windows/OSX, what desktop/window manager are you using?

Vezzra:
@LGM-Doyle, your fix works perfectly fine on OSX.

l29ah:
AFAIK, apps don't ask window managers about their window sizes, but ask X.
Yes, with your patch on top.
No, it's drawn out of screen as well.
Gentoo. What version?
xmonad

Vezzra:
> Yes, with your patch on top.

Latest master has the fix already merged (799eccd51fe22d5bcd43babedc91c047584fa1fe).

> Gentoo. What version?

Well, the version/release of Gentoo you're using. I don't know what the version/release numbering/naming scheme for Gentoo is, but there has to be one. E.g. the Linux system I have on one of my computers here is Kubuntu 16.04.

dbenage-cx:
@Vezzra Gentoo is a rolling release, so does not have version numbers per se

[...]

PR #1307, opened 2017-02-12
Fix bug #1183 in DropDownList pre-scroll to last selected item is incorrect.

LGM-Doyle:
There were several contributing problems all related to the BORDER_SIZE.
1. In the DropDownList::CorrectListSize() it was not accounting for the
borders of the list.
2. In ListBox BringRowIntoView() and FirstRowShownWhenBottomIs() both
were not accounting for the BORDER_SIZE.

### DIFF
--- a/GG/src/DropDownList.cpp
+++ b/GG/src/DropDownList.cpp
@@ -110,7 +110,7 @@ private:
     void WindowResizedSlot(X x, Y y);
 
     /** Used by CorrectListSize() to determine the list height from the current first shown row. */
-    void DetermineListHeight(Pt drop_down_size);
+    Pt DetermineListHeight(const Pt& drop_down_size);
 
     ListBox*            m_lb_wnd;
     const size_t        m_num_shown_rows;
@@ -261,7 +261,9 @@ void ModalListPicker::SignalChanged(boost::optional<DropDownList::iterator> it)
         SelChangedSignal(*it);
 }
 
-void ModalListPicker::DetermineListHeight(Pt drop_down_size) {
+Pt ModalListPicker::DetermineListHeight(const Pt& _drop_down_size) {
+    auto drop_down_size = _drop_down_size;
+
     // Determine the expected height
     auto border_thick = 2 * GG::Y(ListBox::BORDER_THICK);
     auto num_rows = std::min<int>(m_num_shown_rows, LB()->NumRows());
@@ -281,6 +283,8 @@ void ModalListPicker::DetermineListHeight(Pt drop_down_size) {
     if (!LB()->Selections().empty())
         LB()->BringRowIntoView(*(LB()->Selections().begin()));
     GUI::GetGUI()->PreRenderWindow(LB());
+
+    return drop_down_size;
 }
 
 void ModalListPicker::CorrectListSize() {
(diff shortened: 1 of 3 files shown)


GG/src/ListBox.cpp
```diff
@@ -1173,6 +1173,7 @@ void ListBox::BringRowIntoView(iterator it)
 
     Y y_offset(BORDER_THICK), it_y_offset(BORDER_THICK);
     Y first_row_y_offset(BORDER_THICK), last_row_y_offset(BORDER_THICK);
+    iterator it_end_row = --m_rows.end();
```
geoffthemedio: is this safe in the event of an empty listbox?

LGM-Doyle: Yes.

The line above
````
    // m_first_row_shown only equals end() if the list is empty, hence 'it' is invalid.
    if (m_first_row_shown == m_rows.end())
        return;
````
should be checking an invariant of the ListBox that is preserved in all the insert/remove operations.  If this is not true, other parts of ListBox are broken.

**Comment by LGM-Doyle** [comment]
I think that this fixes the following things:
- The 4 pixel gap that you observed at the bottom of the drop down list.
- the ListBox would draw a full sized selection highlight if the selected item was partially scrolled out of the top of the ListBox.  Now it only highlights the visible portion of the selected item.
- the ListBox could scroll the list even after it determined that the vscroll was not needed.  Now it displays the entire list.
- the drop down list near the bottom of the app would be draw outside of the app area.  Now it draws as many items as it can in the space remaining, but never fewer than one.  This fixes issues #1196 and #969, if you are willing to accept that a drop down box with less that one row of space will still be drawn offscreen.  This is better than not drawing anything, because keyboard navigation still works.

File: GG/src/DropDownList.cpp [hunk 2]
```diff
@@ -109,6 +109,9 @@ class ModalListPicker : public Control
         list location not tracking the anchor.*/
     void WindowResizedSlot(X x, Y y);
 
+    /** Used by CorrectListSize() to determine the list height from the current first shown row. */
+    void DetermineListHeight(Pt drop_down_size);
```
geoffthemedio: standard seems to be to pass as `const Pt&` rather than by value

GG/src/ListBox.cpp
```diff
@@ -1159,55 +1159,58 @@ void ListBox::SetSelections(const SelectionSet& s, bool signal/* = false*/)
...
         return;
 
     // m_first_row_shown only equals end() if the list is empty, hence 'it' is invalid.
     if (m_first_row_shown == m_rows.end())
         return;
 
-    // Find the y offsets of the first and last shown rows and 'it'.
-    bool first_row_found(false), last_row_found(false), it_found(false);
+    // Find the y offsets of the first and last shown rows and target.
+    bool first_row_found(false), last_row_found(false), target_found(false);
```
geoffthemedio: @adrianbroher isn't fond of multiple definitions per line

Vezzra:
> This fixes issues #1196 and #969, if you are willing to accept that a drop down box with less that one row of space will still be drawn offscreen.

I guess that's acceptable. So IMO merging this should completely resolve these issues.

Issue #1013, opened 2016-10-01
AI Invasion problems


Morlic-fo:
### Description

There are currently issues with the AI invasions.
- The AI has problems invading planets with shields. Just send a military ship already, stupid!
- The AI generally leaves native planets unconquered for way too long. This indicates the priorities are off and need an overhaul

LGM-Doyle:
ovarwa, commented on the AI invastion/ship deployment in http://freeorion.org/forum/viewtopic.php?f=28&t=10307&start=0 and I transcribed the relevant section.
- When the AI considers something a high priority, it will wisely congregate ships there. Yay! But then it sends out ships in onesy twosies to their deaths at lower priority targets, because it knows it needs to do _something_ but has no idea either going big or not going at all.
- The AI seems to know not to send an inferior force to a sector that already has a greatly superior enemy force. But it has no sense of history, and makes no use of its ability to see my detection range: If I leave the sector empty, it will repeatedly send a ship or two there, and I can move a fleet to the sector to wipe it out.
- When I have a large fleet besieging an enemy system with heavy industry, it builds things like scouts and outposts rather than ships with weapons that can abrade my force. If I plunk a colony down on an unoccupied planet of that system, it also builds lots of troop ships, which are also immediately destroyed without inflicting any damage.

MatGB:
Re the planets with shields problem: it also seems to think that fighters reduce shields and stations fighter-only ships in orbit around high tech natives.

Morlic-fo:
https://github.com/freeorion/freeorion/issues/1012#issuecomment-279231433
>I do regularly see an AI world in the same system as guardians with huge piles of troop drops in orbit, 100+ troop drops will mess up an AI production substantially, if that at the very least could be reduced then the rest shouldn't be release blocking.

Vezzra:
Ok, all that sounds rather serious... @freeorion/ai-team, do you want to tackle this for 0.4.7, or should I leave it only as "optional for 0.4.7" (that is, fine if it can get done in time, but should not block the release)?

Morlic-fo:
#1309 aims to resolve the issues regarding shields, fighters and the "ancient guardian problem". 

Testing and feedback appreciated.

Dilvish-fo:
I think that #1309  has some significant issues, and unless I am misreading something there I don't think it would address the troop base issue at all.  I'll follow up with Morlic for more discussion on that.  Whether covered by 1309 or not, I do think we can get something in soon for the troop bases.

[...]

PR #1309, opened 2017-02-13
Improve AI invasion logic

Morlic-fo:
The PR improves the AI invasion logic by requiring sufficient military resources to secure an invasion targeted system before sending the invasion fleet.

Partial progress to #1013.

### DIFF
--- a/default/python/AI/CombatRatingsAI.py
+++ b/default/python/AI/CombatRatingsAI.py
@@ -216,6 +216,12 @@ class ShipCombatStats(object):
         """
         return self.get_basic_stats(hashable=hashable) + self.get_fighter_stats()
 
+    @property
+    def total_attack(self):
+        """Sum of all conventional weapon (i.e. non-fighter) damage per turn."""
+        attacks, _, _ = self.get_basic_stats()
+        return sum(n*dmg for dmg, n in attacks.iteritems())
+
 
 class FleetCombatStats(object):
     """Stores combat related stats of the fleet."""
@@ -248,6 +254,11 @@ class FleetCombatStats(object):
         :rtype: float
         """
         return combine_ratings_list(map(lambda x: x.get_rating(enemy_stats), self.__ship_stats))
+
+    @property
+    def total_attack(self):
+        """Sum of all conventional weapon (i.e. non-fighter) damage per turn."""
+        return sum(ship.total_attack for ship in self.get_ship_combat_stats())
 
     def __get_stats_from_fleet(self):
         """Calculate fleet combat stats (i.e. the stats of all its ships)."""
--- a/default/python/AI/InvasionAI.py
+++ b/default/python/AI/InvasionAI.py
@@ -360,14 +360,32 @@ def evaluate_invasion_planet(planet_id, empire, secure_fleet_missions, verbose=T
             supply_val *= 0.2
 
     threat_factor = min(1, 0.2*MilitaryAI.totMilRating/(sys_total_threat+0.001))**2  # devalue invasions that would require too much military force
-    build_time = 4
 
-    planned_troops = troops if system_secured else min(troops+max_jumps+build_time, max_troops)
-    if not tech_is_complete("SHP_ORG_HULL"):
-        troop_cost = math.ceil(planned_troops/6.0) * (40*(1+foAI.foAIstate.shipCount * AIDependencies.SHIP_UPKEEP))
+    design_id, _, locs = ProductionAI.get_best_ship_info(PriorityType.PRODUCTION_INVASION)
+    if not locs or not universe.getPlanet(locs[0]):
+        # We are in trouble anyway, so just calculate whatever approximation...
+        build_time = 4
+        planned_troops = troops if system_secured else min(troops + max_jumps + build_time, max_troops)
+        troop_cost = math.ceil(planned_troops / 6.0) * 20 * FleetUtilsAI.get_fleet_upkeep()
     else:
-        troop_cost = math.ceil(planned_troops/6.0) * (20*(1+foAI.foAIstate.shipCount * AIDependencies.SHIP_UPKEEP))
-    planet_score = retaliation_risk_factor(planet.owner) * threat_factor * max(0, pop_val+supply_val+bld_tally+enemy_val-0.8*troop_cost)
+        loc = locs[0]
+        species_here = universe.getPlanet(loc).speciesName
+        design = fo.getShipDesign(design_id)
+        cost_per_ship = design.productionCost(empire_id, loc)
+        build_time = design.productionTime(empire_id, loc)
+        troops_per_ship = CombatRatingsAI.weight_attack_troops(design.troopCapacity,
+                                                               CombatRatingsAI.get_species_troops_grade(species_here))
+        planned_troops = troops if system_secured else min(troops + max_jumps + build_time, max_troops)
+        ships_needed = math.ceil(planned_troops / float(troops_per_ship))
+        troop_cost = ships_needed * cost_per_ship  # fleet upkeep is already included in query from server
+
+    # apply some bias to expensive operations
+    normalized_cost = float(troop_cost) / max(empire.productionPoints, 1)
+    normalized_cost = max(1, normalized_cost)
+    cost_score = (normalized_cost**2 / 50.0) * troop_cost
+
+    base_score = max(0, pop_val + supply_val + bld_tally + enemy_val - cost_score)
+    planet_score = retaliation_risk_factor(planet.owner) * threat_factor * base_score
     if clear_path:
         planet_score *= 1.5
     if verbose:
--- a/default/python/AI/InvasionAI.py
+++ b/default/python/AI/InvasionAI.py
@@ -429,13 +429,29 @@ def send_invasion_fleets(fleet_ids, evaluated_planets, mission_type):
                 continue
             else:
                 these_fleets = found_fleets
+
+        military_escort_fleets = MilitaryAI.get_military_support_for_invasion(sys_id)
+        if not military_escort_fleets:
+            print "Found Invasion fleets for target %s but have not enough military to support it." % planet
+            invasion_fleet_pool.update(found_fleets)
+            continue
+
+        # send invasion fleet
         target = universe_object.Planet(planet_id)
         print "assigning invasion fleets %s to target %s" % (these_fleets, target)
-        for fleetID in these_fleets:
-            fleet_mission = foAI.foAIstate.get_fleet_mission(fleetID)
+        for fleet_id in these_fleets:
+            fleet_mission = foAI.foAIstate.get_fleet_mission(fleet_id)
             fleet_mission.clear_fleet_orders()
             fleet_mission.clear_target()
             fleet_mission.set_target(mission_type, target)
+
+        # send military escort
+        system_to_secure = universe_object.System(sys_id)
+        for fleet_id in military_escort_fleets:
+            fleet_mission = foAI.foAIstate.get_fleet_mission(fleet_id)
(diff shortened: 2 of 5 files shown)


MatGB:
message just popped up:
`AI_4: AI_Error: AI script error in "assign_invasion_fleets_to_invade": "'list' object has no attribute 'update'"`
[AI_4-error.log.zip](https://github.com/freeorion/freeorion/files/773440/AI_4-error.log.zip)

Playing with this, #1311 and #1304 rebased to master

Morlic-fo:
Thanks. Should be fixed.

default/python/AI/CombatRatingsAI.py
```diff
@@ -216,6 +216,11 @@ def get_stats(self, hashable=False):
         """
         return self.get_basic_stats(hashable=hashable) + self.get_fighter_stats()
 
+    @property
+    def raw_attack(self):
```
Cjkjvfnby: Little docstring will be helpfull.  

"""
Sum of all weapon damage without species bonuses.
"""

default/python/AI/InvasionAI.py
```diff
@@ -360,14 +360,37 @@ def evaluate_invasion_planet(planet_id, empire, secure_fleet_missions, verbose=T
...
 
     threat_factor = min(1, 0.2*MilitaryAI.totMilRating/(sys_total_threat+0.001))**2  # devalue invasions that would require too much military force
-    build_time = 4
 
-    planned_troops = troops if system_secured else min(troops+max_jumps+build_time, max_troops)
-    if not tech_is_complete("SHP_ORG_HULL"):
-        troop_cost = math.ceil(planned_troops/6.0) * (40*(1+foAI.foAIstate.shipCount * AIDependencies.SHIP_UPKEEP))
+    design_id, _, locs = ProductionAI.get_best_ship_info(PriorityType.PRODUCTION_INVASION)
+    if not locs or not universe.getPlanet(locs[0]):
+        # We are in trouble anyway, so just calculate whatever approximation...
```
Cjkjvfnby, line 366: It is not clear in which trouble we are.

default/python/AI/InvasionAI.py
```diff
@@ -360,14 +360,37 @@ def evaluate_invasion_planet(planet_id, empire, secure_fleet_missions, verbose=T
...
+        build_time = 4
+        planned_troops = troops if system_secured else min(troops + max_jumps + build_time, max_troops)
+        if not tech_is_complete("SHP_ORG_HULL"):
```
Cjkjvfnby: `cost_multiplier = 20 if tech_is_complete("SHP_ORG_HULL") else 40`

default/python/AI/InvasionAI.py
```diff
@@ -411,14 +434,30 @@ def send_invasion_fleets(fleet_ids, evaluated_planets, mission_type):
...
-        for fleetID in these_fleets:
-            fleet_mission = foAI.foAIstate.get_fleet_mission(fleetID)
+        for fleet_id in these_fleets:
+            fleet_mission = foAI.foAIstate.get_fleet_mission(fleet_id)
             fleet_mission.clear_fleet_orders()
             fleet_mission.clear_target()
             fleet_mission.set_target(mission_type, target)
 
+        # send military escort
+        target = universe_object.System(sys_id)
```
Cjkjvfnby: rename to `system_to_secure`

default/python/AI/MilitaryAI.py
```diff
@@ -786,3 +786,44 @@ def assign_military_fleets_to_systems(use_fleet_id_list=None, allocations=None,
...
+    all_military_fleet_ids = FleetUtilsAI.get_empire_fleet_ids_by_role(MissionType.MILITARY)
+    if not all_military_fleet_ids:
+        return []
+
+    avail_mil_fleet_ids = set(FleetUtilsAI.extract_fleet_ids_without_mission_types(all_military_fleet_ids))
+    found_fleets = []
+    found_stats = {}
+    min_stats = {
+        'rating': (total_defense * (total_shields + total_defense) + fleet_threat) * 2 + 1,
+        'rawAttack': (total_shields + total_defense) / 3,
```
Cjkjvfnby: Why do we use rawAttack?

Cjkjvfnby (changes requested):

default/python/AI/InvasionAI.py
```diff
@@ -360,14 +360,37 @@ def evaluate_invasion_planet(planet_id, empire, secure_fleet_missions, verbose=T
...
+    # apply some negative bias to expensive operations
+    normalized_cost = float(troop_cost) / max(empire.productionPoints, 1)
+    if normalized_cost < 1:
+        cost_score = 0
     else:
-        troop_cost = math.ceil(planned_troops/6.0) * (20*(1+foAI.foAIstate.shipCount * AIDependencies.SHIP_UPKEEP))
-    planet_score = retaliation_risk_factor(planet.owner) * threat_factor * max(0, pop_val+supply_val+bld_tally+enemy_val-0.8*troop_cost)
+        cost_score = -(normalized_cost**2 / 50.0) * troop_cost
+
+    base_score = max(0, pop_val + supply_val + bld_tally + enemy_val - cost_score)
```
Dilvish-fo: cost_score is subtracted below, so shouldn't we leave it positive here (or vice-versa) or it will do just the opposite of what we want?

Also, your description "Increase the negative bias for invasions which are costy compared to the current production output of the AI" -- it looks to me like this calc will do that if the troop cost is more than about 7 times the total empire production, but can significantly underweight (or totally ignore) the cost if it is less than that -- I'm not really understanding the reason for this last part.

Morlic-fo: Define "underweight". 

The basic idea is that producing troops with a total cost in the order of the Empire's production output does not cause much struggle:
Assume that troop ships build in 3 turns, then if the cost is less than 3 turns of empire's production output, there will remain (some) PP for other projects.
If the cost is even within a single turn's production output, then it will take less than a third of our resources per turn. It is in some sense negliglible and it would be preferable to get the best planet at that point instead of one where we have to spend 20% less PP.

 If it is slightly higher cost per turn than we can produce, we pause the production of other things for some 3-5 turns which does not hurt too much and the opportunity cost is thus limited.
The longer we have to build troop ships because we do not have "sufficient" PP, the later we will get returns and the higher will be the opportunity cost.

An extreme case was highlighted by Mat with regard to the Ancient Guardians:
https://github.com/freeorion/freeorion/issues/1012#issuecomment-279231433

It thus makes sense to scale the ```cost_score``` with the available production.
The numeric factor of 1/50 is arbitrary in the sense that I chose some number that seemed reasonable and led to seemingly decent results ingame.
The quadratic scaling is based on the idea that the opportunity costs seem to warrant a higher-than-linear scaling while not scaling too aggressive.

If we can agree on the general concept, the numbers can be tweaked as needed.

Morlic-fo:
I addressed @Cjkjvfnby's points and fixed the ```cost_score``` by making it positive as pointed out by Dilvish and not letting it reach zero.

default/python/AI/InvasionAI.py
Dilvish-fo: I think that describing these as escorts is misleading -- it could be that the troop ships could benefit from an actual escort, but that's not what these are; these are securing missions, like what is already covered for invasions but simply established/launched at a different point in the planning cycle than what the current ones are.

Dilvish-fo:
OK well after looking through this more thoroughly, I think it should be broken up into the two parts that the intro describes for it, separately.  I had some comments on the troop cost aspect, but subject to some adjustment there I think that part could be fine to go in.  (*edit: I just saw that you made changes there, I'll go look at that now.*)

The "Escort" Military mission part, has a good main point, and perhaps most or all of this code can be used, but I think it needs to fit into the overall planning cycle differently and should not go in like this, although perhaps the fix can be relatively simple.    The problem with the approach here is that the InvasionAI gets called before the MilitaryAI does, **and this code could allocate warships to Invasions that are really more strongly needed for Defense**.  If I extract the primary idea here as "do not create an invasion mission unless we have sufficient military support available" then I DO think that we could usefully apply that idea, and perhaps this code could still be largely applicable, but would need some other significant changes also.

To step back a minute and look at the current (in master, not in this PR) sequence of planning/actions -- the InvasionAI identifies and ranks invasion targets, and then may launch some troop fleets to invade some of those.   Next up the MilitaryAI assigns troops to defend systems and if it has enough remaining it assigns them to securing invasion targets and then colonization targets.  So the main problem that this PR is trying to address is that a troop fleet might be launched even though the necessary securing military forces don't wind up available.

Currently the troop fleets could start moving whether or not the target has been secured, but they *shouldn't* be taking the last jump to the target unless that is safe.  That aspect is partly but not perfectly working and I think it could be improved by having the FleetMove order take into account the overall mission type and be more careful about that final step, and also perhaps the invasion should not even start moving at all unless the target is totally safe or there is already a military mission assigned to secure it (akin to the approach here).

Besides the above ideas on movement, here is how I think the ideas of this PR could best be applied-- instead of having the InvationAI run *before* the MilitaryAI, we have it get called *by* the MilitaryAI.  The MilitaryAI would first go through and assign its defense forces, and then it would call the InvasionAI which could approach things largely as proposed here, then back in the MilitaryAI it could assess whether any of the previously launched invasions have increased military needs and deal with those, and then, like now, move on to securing outpost/colonization targets, etc.

So mostly, relative to this PR, we would be moving the main call to InvasionAI.assign_invasion_fleets_to_invade() from being placed by FreeOrionAI to being placed by MilitaryAI, at the midpoint in the latter's planning just prior to the section where it currently deals with invasion securing missions, and then that section of MilitaryAI would also get some adjustment probably.

As for the issue with the large number of troop bases sometimes seen, they are handled by an earlier section of the InvasionAI code and it doesn't look to me like that gets touched here.  I think the best thing for them would be to simply be more restrictive on what targets can qualify for the base-invasion approach, I'll look at putting in a separate PR to handle that.

Dilvish-fo (changes requested):
as noted in previous comments

Morlic-fo:
> The problem with the approach here is that the InvasionAI gets called before the MilitaryAI does, and this code could allocate warships to Invasions that are really more strongly needed for Defense.

Good point.

> Besides the above ideas on movement, here is how I think the ideas of this PR could best be applied-- instead of having the InvationAI run before the MilitaryAI, we have it get called by the MilitaryAI.

Sounds plausible, I will look into this.

Morlic-fo:
> OK well after looking through this more thoroughly, I think it should be broken up into the two parts that the intro describes for it, separately. I had some comments on the troop cost aspect, but subject to some adjustment there I think that part could be fine to go in. (edit: I just saw that you made changes there, I'll go look at that now.)

#1341 for the cost-part.

geoffthemedio:
Is part of this still work in progress?

Dilvish-fo:
> Is part of this still work in progress?

We have covered a fair bit of this in various PRs and commits already, but there is still a portion of this yet not done, about changing where we call the invasion code and a bit of further improvement in the coordination between general military planning and invasion missions.  I'm not totally sure though if Morlic was planning on tackling that or if he was expecting me to, and I'm not sure that we need to consider that portion to be a 0.4.7 Milestone.   So we might want to change the milestone but until we get that last bit in I think this PR should stay up.

Right now it looks like Morlic is focusing on dealing with the complications about Carriers vs Planets, and I think that is a much higher priority than anything else for the AI.  I'll open a thread on the forums to talk about what else we might want or need to focus on for the AI for 0.4.7.

Morlic-fo:
Closing this. Changes in #1353 are an alternative implementation to effectively achieve the same. The general idea I had in this PR that invasion missions should require sufficient military forces before them starting is still valid and may be implemented some day but not in this form.

PR #1313, opened 2017-02-13
Fix typo resulting in the AI not being able to use colony ships

### LINKED ISSUE #1312: AI does not use the starting colony ship
### Description
The AI currently prefers building a new outpost ship and colonizing with that instead of using its starting colony ship. This obviously delays expansion.

### Expected Result
Use the colony ship...

### Steps to reproduce
Pretty much happens in any game as far as I can see.

Morlic-fo:
fixes #1312


Cjkjvfnby:
Why not update all `except:\n` cases?

Morlic-fo:
I just updated this specific except-case in this PR as it was part of the bug-hunt.

But yes, all empty except statements should be checked and updated as they hide bugs such as this one.

PR #1318, opened 2017-02-14
Refactor MultiTurnProgressBar for percentage arguments

### LINKED ISSUE #1201: Building in progress icon not displaying progress on sidepanel
### Environment

* **FreeOrion Version**: #1192 rebased to master
* **Operating System**: Linux
  * Compiled from source

### Description
I think this is probably a result of the fractional progress merge, but the sidepanel icons that should display building progress aren't showing any at all.

dbenage-cx:
Change control to require percentages of total width to fill.

Fixes #1201

### DIFF
--- a/UI/CUIControls.cpp
+++ b/UI/CUIControls.cpp
@@ -19,6 +19,7 @@
 
 #include <boost/format.hpp>
 #include <boost/regex.hpp>
+#include <boost/algorithm/clamp.hpp>
 
 #include <limits>
 #include <iomanip>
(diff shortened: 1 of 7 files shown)


UI/CUIControls.cpp
```diff
@@ -1930,61 +1930,64 @@ void ProductionInfoPanel::DoLayout() {
...
+    m_clr_bg(bg_color),
+    m_clr_outline(outline_color)
+{
+    // validate percentage values
+    if (m_perc_completed < 0.0f || m_perc_predicted < 0.0f ||
+        (m_perc_completed + m_perc_predicted) > 1.0f)
+    {
+        WarnLogger() << "Values not within percent range, clamping: "
+                     << m_perc_completed << ", " << m_perc_predicted;
+        m_perc_completed = std::min(1.0f, std::max(0.0f, m_perc_completed));
```
adrianbroher: Use `boost::algorithm::clamp` from `boost/algorithm/clamp.hpp` to make your intention clearer.  A future C++17 `std::clamp` is on the way.

UI/CUIControls.h
```diff
@@ -608,23 +608,36 @@ class ProductionInfoPanel : public CUIWnd {
...
-    MultiTurnProgressBar(int total_turns, double turns_completed, double total_cost, double turn_spending,
-                         const GG::Clr& bar_color, const GG::Clr& background, const GG::Clr& outline_color);
-    //@}
-
-    /** \name Mutators */ //@{
+    /** @name Structors */ //@{
+    /** Ctor
+    * @param[in] num_segments Number of segments in the bar
+    * @param[in] percent_completed Percent of bar to fill
+    * @param[in] percent_predicted Percent of bar to fill after
```
adrianbroher: It's not clear what it the valid input range for this.

 * 0.0 to (1.0 - percent_completed)?
 * percent_completed to 1.0?
 * 0.0 to 1.0?

UI/ResearchWnd.cpp
```diff
@@ -172,11 +172,22 @@ namespace {
...
-                                                  m_in_progress ? ClientUI::ResearchableTechFillColor() : GG::LightColor(ClientUI::ResearchableTechFillColor()) );
+        int total_time = 0;
+        float next_progress = 0.0f;
+        if (tech) {
+            total_time = tech->ResearchTime(m_empire_id);
+            float total_cost = tech->ResearchCost(m_empire_id);
+            next_progress = turn_spending / std::max(1.0f, total_cost);
+        }
+        m_progress_bar = new MultiTurnProgressBar(total_time,
+            total_time > 0 ? turns_completed / total_time : 0.0f,
```
adrianbroher: I you do most of the calculation outside the constructor parameter list you should move all of it.

adrianbroher: Also why `total_time > 0` in the first place?  Is `Tech::ResearchTime` allowed to return `<= 0`?

adrianbroher (changes requested):

UI/ResearchWnd.cpp
dbenage-cx: Not immediately clear if it may return 0, due to branches that return from Eval().

dbenage-cx:
Rebased with requested changes

adrianbroher (approved):

UI/CUIControls.cpp
```diff
@@ -1930,61 +1931,65 @@ void ProductionInfoPanel::DoLayout() {
...
+}
 
-    // draw segment separators
-    if (segmented) {
+void MultiTurnProgressBar::Render() {
+    GG::Pt ul(UpperLeft());
+    GG::Pt lr(LowerRight());
+    GG::Y bottom(lr.y);
+    // draw background
+    FlatRectangle(ul, lr, m_clr_bg, m_clr_outline, 1);
```
geoffthemedio: would be preferable to render with vertex arrays here rather than a series of calls to FlatRectangle which I think will end up in here https://github.com/freeorion/freeorion/blob/master/GG/src/DrawUtil.cpp#L54 doing separate calls to set the rendering state each time.

UI/CUIControls.h
```diff
@@ -608,23 +608,36 @@ class ProductionInfoPanel : public CUIWnd {
...
 
 private:
-    int     m_total_turns;
-    double  m_turns_completed;
-    double  m_total_cost;
-    double  m_turn_spending;
-    GG::Clr m_bar_color;
-    GG::Clr m_background;
-    GG::Clr m_outline_color;
+    int             m_num_segments;
```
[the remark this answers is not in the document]
geoffthemedio: Why? I find indenting so the variable names line up easier to read.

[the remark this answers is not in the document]
adrianbroher: * You touch lines unrelated to the commit content whenever you change a member attribute name just to align the other member attributes.  A pain to deal with when doing code archaeology/blaming.
* All member attributes lines get overlong unnecessary as soon there is a single member with a more complex type name.  A pain to deal with when reading diffs.

geoffthemedio:
otherwise seems to work in a quick test

UI/CUIControls.cpp
dbenage-cx: Should all of the funcs in DrawUtil.cpp return a set of vertexes instead?

geoffthemedio: they are still useful as is in some situations

adrianbroher: But dbenage-cx idea isn't that wrong.  When looking at the draw utils as algorithms to write/prepare the vertex buffer it's the clients choice to batch multiple draw util drawing operations into a single draw call or use a draw call for each draw util call.  This is of course out of the scope of this PR but I like the general idea.

geoffthemedio:
Trouble with that is that merging into a single draw call only works if you're rendering several things with the same or compatible render settings and formats. Doing that through a simple one call per thing to draw API is rather complicated. The existing functions work relatively independently in any order and with arbitrary combonations of rendering settings. They are slower than batched rendering as a result though, much like (but hopefully not as much as) immediate mode render calls.

dbenage-cx:
Incomplete, some bug with not correctly prerendering the collapsed version of building indicator.

dbenage-cx:
Removed use of `PreRender`.  I was able to fix the offset placement be adding `m_progress_bar->RequirePreRender()` to `BuildingIndicator::SizeMove()`.

There are multiple other issues with prerender noted: odd flickering during updates, incorrect colors in queues, and a detached/fixed progress bar when dragging queue entries.

Edit: Also, rebased to master with fixup

UI/CUIControls.cpp
```diff
@@ -1930,81 +1931,123 @@ void ProductionInfoPanel::DoLayout() {
...
+
+    // define completed and predicted bar sizes
+    GG::X comp_width(Width() * m_perc_completed);
+    GG::X pred_width(Width() * m_perc_predicted);
+    // maximum lower right points to lower right of control
+    GG::Rect comp_rect(ul, {std::min(lr.x, ul.x + comp_width), bottom});
+    GG::Rect pred_rect({std::max(ul.x, comp_rect.lr.x - 1), ul.y},
+                       {std::min(lr.x, comp_rect.lr.x + pred_width), bottom});
+
+    // predicted progress
```
geoffthemedio, line 1977: this is better, but storing the vertices in member variables that don't need to be refilled each frame would be preferable. or does doing this result in the prerender issues you mentioned?

dbenage-cx: Yes, I had these as member variables and split at line 2006.

geoffthemedio:
more optimized rendering would be preferable, but this functionality fix is probably a higher prioirity

PR #1320, opened 2017-02-14
Diplomacy Additions

geoffthemedio:
Adds allied diplomatic status, commands to use it.


universe/ValueRef.cpp
```diff
@@ -1841,6 +1841,18 @@ double ComplexVariable<double>::Eval(const ScriptingContext& context) const
         return part_type->Capacity();
 
     }
+    else if (variable_name == "PartSecondaryStat") {
```
geoffthemedio: this bit's off topic, should be picked into master at some point if this isn't merged soon

update: already merged to master for weapon stat effects

Empire/EmpireManager.cpp
```diff
@@ -80,10 +87,18 @@ std::string EmpireManager::Dump() const {
         switch (entry.second) {
         case DIPLO_WAR:     retval += "War";    break;
         case DIPLO_PEACE:   retval += "Peace";  break;
+        case DIPLO_ALLIED:  retval += "Allied";  break;
         default:            retval += "?";      break;
         }
         retval += "\n";
     }
+    retval += "Diplomatic Messages:\n";
+    for (auto message : m_diplomatic_messages) {
```
adrianbroher: This creates copies of whatever an entry in m_diplomatic_messages is.  Better use `auto& var` or `const auto& var` when `var` is not modified in the loop.

Empire/EmpireManager.h
```diff
@@ -54,10 +54,11 @@ class FO_COMMON_API EmpireManager {
     int                 NumEmpires() const;
 
     DiplomaticStatus            GetDiplomaticStatus(int empire1, int empire2) const;
-    bool                        DiplomaticMessageAvailable(int empire1, int empire2) const;
-    const DiplomaticMessage&    GetDiplomaticMessage(int empire1, int empire2) const;
+    bool                        DiplomaticMessageAvailable(int sender_id, int recipient_id) const;
+    const DiplomaticMessage&    GetDiplomaticMessage(int sender_id, int recipient_id) const;
 
     std::string         Dump() const;
+    std::string         DumpDiplomacy() const;
```
adrianbroher: No need for a public interface.

Empire/EmpireManager.cpp
```diff
@@ -177,91 +189,147 @@ void EmpireManager::SetDiplomaticMessage(const DiplomaticMessage& message) {
...
     int sender_empire_id = message.SenderEmpireID();
     int recipient_empire_id = message.RecipientEmpireID();
-    //std::pair<int, int> key = DiploKey(sender_empire_id, recipient_empire_id);
+
     DiplomaticStatus diplo_status = GetDiplomaticStatus(sender_empire_id, recipient_empire_id);
-    bool message_already_available = DiplomaticMessageAvailable(sender_empire_id, recipient_empire_id);
-    const DiplomaticMessage& existing_message = GetDiplomaticMessage(sender_empire_id, recipient_empire_id);
+    bool message_from_recipient_to_sender_available = DiplomaticMessageAvailable(recipient_empire_id, sender_empire_id); 
+    const DiplomaticMessage& existing_message_from_recipient_to_sender = GetDiplomaticMessage(recipient_empire_id, sender_empire_id);
+    bool message_from_sender_to_recipient_available = DiplomaticMessageAvailable(sender_empire_id, recipient_empire_id);
```
dbenage-cx: If this portion is mostly complete, this variable is unused. (having both status tags, not certain which errors are worth mention)

Vezzra:
@geoffthemedio, what milestone do you suggest? Required/optional for 0.4.7? I guess it's required, but wanted to check back with you first...

geoffthemedio:
probably optional... may need AI support I can't implement.

Voker57:
With that patch, ships do not refuel in allied space, but do refuel in peaceful space.

geoffthemedio:
> With that patch, ships do not refuel in allied space, but do refuel in peaceful space.

Until one or the other are merged, I more or less consider this and the "allied" refueling patch separate / independent.

geoffthemedio:
Are there any AI problems with merging this?

geoffthemedio:
pulled the allied fleet supply changes into this pull request

l29ah:
Being able to hack the upkeep mechanics with a spare player sucks.

geoffthemedio:
> Being able to hack the upkeep mechanics with a spare player sucks.

I don't know what that means, or what it has to do with this pull request, but perhaps the solution is not having any "spare player" in your game? Having a "spare player" will break game balance in many ways, presumably.

l29ah:
On Sun, Feb 19, 2017 at 11:04:28PM -0800, Geoff wrote:
> > Being able to hack the upkeep mechanics with a spare player sucks.
> 
> I don't know what that means, but perhaps the solution is not having any "spare player" in your game? Having a "spare player" will break game balance in many ways, presumably.

Playing three players FFA, one player gets advantage over the other and threatens him to ally or get crushed, distributes his upkeep with him and they waste the third one together.

-- 

geoffthemedio:
So by "upkeep" you're referring to fleet resupply?

Regardless, I don't see this as a problem with the game mechanics, or as a "hack".

If you don't like alliances between players, have a house rule against them.

l29ah:
FLEET_UPKEEP_MULTIPLICATOR

geoffthemedio:
What does that have to do with this pull request?

You can give ships to other players whether or not the "allied" diplomatic state exists.

l29ah:
On Mon, Feb 20, 2017 at 02:30:55AM -0800, Geoff wrote:
> What does that have to do with this pull request?
> 
> You can give ships to other players whether or not the "allied" diplomatic state exists.

Without the supply sharing it would be penalized with inability to move them around due to the lack of fuel.

-- 

Voker57:
This is better addressed by having fleet upkeep substract from industry rather than multiply ship cost (which makes about zero sense)

geoffthemedio:
> Without the supply sharing it would be penalized with inability to move them around due to the lack of fuel.

OK, that starts to sound more relevant to this pull request.

> This is better addressed by having fleet upkeep substract from industry rather than multiply ship cost (which makes about zero sense)

That won't happen, in part because ships are at particular locations, and production output is at other particular locations, and these locations may not be connected, so there's no way to decide where to take the production from. There have been extensive discussions on the subject in the forums.

...

Regardless, for purposes of this pull request, can you do your mulitplayer games and not make alliances and then gift eachother ships to intentionally exploit the increase in ship costs mechanism, or is avoiding that impossible?

Alternatively, I can disable ship gifting.

If you want the allied fleet supply mechanisms merged before the next release, it can't be dependent on reworking major other game mechanics.

Voker57:
Also, upkeep exploiting is balanced by fact that you give your potential opponent a huge fleet that they can use to strike back at you.

Voker57:
PR appears to be working very well.

Vezzra:
> Also, upkeep exploiting is balanced by fact that you give your potential opponent a huge fleet that they can use to strike back at you.

Absolutely, this should sufficiently counter the ship-offloading-to-ally practice so it's not really an exploit.

MatGB:
It is if you're playing 2v2 allied victory, one player controls the fleets the other the planets could be very nasty in the current setup, and definitely an exploit if the other team isn't doing it too.

dbenage-cx:
> Alternatively, I can disable ship gifting.

Possibly have some limit per turn, starting at 0.  Starting value might be user defined, leaving a possibly for content to increase it down the road.
Does not really address the issue, but defers it until other mechanics are in place.

Vezzra:
We're dealing with incomplete mechanics here, so IMO we just have to live with the fact that some special cases don't work and/or cause seriously inbalanced games (like the mentioned 2v2 allied victory game).

Disabling other, already working, mechanics or adding more stop-gap solutions like limiting ship gifting only creates other, equally annying issues in place of the one that are supposed to be alleviated by these solutions. Which only makes things more complicated, and adds stuff that needs to be reverted later on, while not really gaining anything.

So my suggestion is to just live with the exploit for now. It only really matters in some edge cases, as far as I can see, which is tolerable in my eyes.

geoffthemedio:
Supply sharing through adjacent allied networks should now work. Probably some (literal) edge cases I've not identified, though.

geoffthemedio:
Objections to merging?

dbenage-cx (commented):
Possibly some notification to the player when their status with another empire is changed or requests are received/rejected.

Vezzra:
> Objections to merging?

Fine with me.

geoffthemedio:
notification to players is a feature needed independent of this pull request (although this pull request would perhaps make it more useful / important to have)

Dilvish-fo:
Hi, somehow I had overlooked this PR for quite a while.  Geoff, your AI script adjustement in ea71604 looks fine for its limited purpose.  As you hinted at in your commit message, we should at some point add something to deal with alliance proposals, but (to my understanding) there shouldn't be any real harm in merging the PR as it is, and we can add the alliance response/rejection later (and start working on actual AI cooperation between players at some point).

Empire/EmpireManager.cpp
```diff
@@ -177,91 +207,147 @@ void EmpireManager::SetDiplomaticMessage(const DiplomaticMessage& message) {
...
     int sender_empire_id = message.SenderEmpireID();
     int recipient_empire_id = message.RecipientEmpireID();
-    //std::pair<int, int> key = DiploKey(sender_empire_id, recipient_empire_id);
+
     DiplomaticStatus diplo_status = GetDiplomaticStatus(sender_empire_id, recipient_empire_id);
-    bool message_already_available = DiplomaticMessageAvailable(sender_empire_id, recipient_empire_id);
-    const DiplomaticMessage& existing_message = GetDiplomaticMessage(sender_empire_id, recipient_empire_id);
+    bool message_from_recipient_to_sender_available = DiplomaticMessageAvailable(recipient_empire_id, sender_empire_id); 
+    const DiplomaticMessage& existing_message_from_recipient_to_sender = GetDiplomaticMessage(recipient_empire_id, sender_empire_id);
+    //bool message_from_sender_to_recipient_available = DiplomaticMessageAvailable(sender_empire_id, recipient_empire_id);
```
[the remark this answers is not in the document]
geoffthemedio: I don't know how to clarify... it's checking if there is already another message waiting from the recipient of the current message to the sender of the current message. ie. when handing a message from A to B, it checks if there was already a message from B to A sent.

PR #1321, opened 2017-02-14
introduce an inner system circle

TheSilentOne1:
As mentioned here ( http://www.freeorion.org/forum/viewtopic.php?f=6&t=10413 ) I suggest that we introduce a second, outer circle around systems that indicates if a colony is present at the sytem, in addition to the system font. This makes it a lot easier to read the galaxy map. The use of the inner circle (to indicate which empire controls supply for the system) remains unchanged.
Natives are indicated by a white segmented outer circle; empire colonies have an unbroken circle that will have multi-colored segments if colonies of different empires are present.
Also adjusts/reduces the width of starlanes when the map is zoomed far out.
Further consideration should go into representation of capitals and ship yards on the map at some point.
Review of this PR can ofc wait for after the upcoming release.


adrianbroher:
Please squash the formatting commits.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,93 @@ void MapWnd::RenderSystems() {
...
             }
 
-            // render circles around systems that have at least one starlane, if circles are enabled.
+            // render circles around systems that have at least one starlane if they are enabled (specific circles for player owned planets, capitals, neutrals)
             if (circles) {
                 if (std::shared_ptr<const System> system = GetSystem(system_icon.first)) {
                     if (system->NumStarlanes() > 0) {
-                        glColor(GetOptionsDB().Get<GG::Clr>("UI.unowned-starlane-colour"));
 
+                        bool has_empire_planet = false, has_neutrals = false;
```
adrianbroher: No multiple declarations on one line.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,93 @@ void MapWnd::RenderSystems() {
...
+                        std::map<int, int> colony_number_by_empire_id;
+                        std::vector<std::shared_ptr<const Planet>> system_planets = Objects().FindObjects<const Planet>(system->PlanetIDs());
```
adrianbroher: No need for a local variable.  This can be inlined into the loop.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,93 @@ void MapWnd::RenderSystems() {
...
+                        const std::set<int>& known_destroyed_object_ids = GetUniverse().EmpireKnownDestroyedObjectIDs(HumanClientApp::GetApp()->EmpireID());
+
+                        for (std::shared_ptr<const Planet> planet : system_planets) {
+                            int planet_id = planet->ID();
+                            if (known_destroyed_object_ids.find(planet_id) != known_destroyed_object_ids.end())
```
adrianbroher: Use `std::set<T>::count`.  You don't need the element or the iterator so there is no need for code duplication.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,93 @@ void MapWnd::RenderSystems() {
...
+
+                            // remember if this system has a player-owned planet, count # of colonies for each empire
+                            if (!planet->Unowned()) {
+                                has_empire_planet = true;
+
+                                std::map<int, int>::iterator it = colony_number_by_empire_id.find(planet->Owner()) ;
+                                if (it != colony_number_by_empire_id.end()) {
+                                    it->second++;
+                                } else {
+                                    colony_number_by_empire_id.insert(std::pair<int, int> (planet->Owner(), 1));
```
adrianbroher: Use C++11 initializer list.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,93 @@ void MapWnd::RenderSystems() {
...
+                                CircleArc(outer_circle_ul, outer_circle_lr, n * segment, (n+1) * segment, false);
+                            }
+                        }
+
+                        // systems with empire planets have an unbroken outer circle; color segments for each empire present
+                        if (has_empire_planet) {
+                            glLineWidth(1.5f);
+
+                            int colonised_planets = 0, position = 0;
+                            for (std::map<int, int>::iterator it = colony_number_by_empire_id.begin(); it != colony_number_by_empire_id.end(); ++it) {
```
adrianbroher: Use C++11 range based for loop.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,93 @@ void MapWnd::RenderSystems() {
...
 
-            // render circles around systems that have at least one starlane, if circles are enabled.
+            // render circles around systems that have at least one starlane if they are enabled (specific circles for player owned planets, capitals, neutrals)
             if (circles) {
                 if (std::shared_ptr<const System> system = GetSystem(system_icon.first)) {
                     if (system->NumStarlanes() > 0) {
-                        glColor(GetOptionsDB().Get<GG::Clr>("UI.unowned-starlane-colour"));
 
+                        bool has_empire_planet = false, has_neutrals = false;
+                        std::map<int, int> colony_number_by_empire_id;
```
adrianbroher: `Number` has a broad meaning. `Count` would be more understandable.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,93 @@ void MapWnd::RenderSystems() {
...
+                        if (has_empire_planet) {
+                            glLineWidth(1.5f);
+
+                            int colonised_planets = 0, position = 0;
+                            for (std::map<int, int>::iterator it = colony_number_by_empire_id.begin(); it != colony_number_by_empire_id.end(); ++it) {
+                                colonised_planets += it->second;
+                            }
+                            float segment = static_cast<float>(TWO_PI) / colonised_planets;
+
+                            for (std::map<int, int>::iterator it = colony_number_by_empire_id.begin(); it != colony_number_by_empire_id.end(); ++it) {
```
adrianbroher: Use C++11 range based for loop.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,93 @@ void MapWnd::RenderSystems() {
...
+
             if (fog_scanlines
                 && (universe.GetObjectVisibilityByEmpire(system_icon.first, empire_id) <= VIS_BASIC_VISIBILITY))
             {
                 m_scanline_shader.SetColor(GetOptionsDB().Get<GG::Clr>("UI.system-fog-of-war-clr"));
                 m_scanline_shader.RenderCircle(circle_ul, circle_lr);
             }
 
-            // render circles around systems that have at least one starlane, if circles are enabled.
+            // render circles around systems that have at least one starlane if they are enabled (specific circles for player owned planets, capitals, neutrals)
```
adrianbroher: Wrap overlong comments at a reasonable line length (> 80 total).

adrianbroher (changes requested):

adrianbroher:
I also think you have the order of rings the wrong way around.  The innermost should represent properties of the system (ownership and capital planets), the outermost should represent effects of the system to the rest of the galaxy (supply generation, ship facilities).

adrianbroher:
> applying adrianbroher's corrections

How are those changes related to me?

https://chris.beams.io/posts/git-commit/

geoffthemedio:
> How are those changes related to me?

At least some were suggested / requested by you.

Builds and runs OK for me. Do want the inner-outer circles switched, though.

TheSilentOne1:
@adrianbroher: It was only a temporary name for the commit before squashing, and gone now. Thank you for reviewing the code, it's appreciated.

@adrianbroher, @geoffthemedio: I'm somewhat reluctant about switching the rings, as all the rings around uncolonised systems become larger, which doesn't look right to me. Although I see your point about the rings' order, adrianbroher, I think the current order does actually feel intuitive in game. If you shouldn't have done so I recommend you try it out a little.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,94 @@ void MapWnd::RenderSystems() {
...
             }
 
-            // render circles around systems that have at least one starlane, if circles are enabled.
+            // render circles around systems that have at least one starlane, if they are enabled
             if (circles) {
                 if (std::shared_ptr<const System> system = GetSystem(system_icon.first)) {
                     if (system->NumStarlanes() > 0) {
-                        glColor(GetOptionsDB().Get<GG::Clr>("UI.unowned-starlane-colour"));
 
+                        bool has_empire_planet = false;
```
adrianbroher: This variable isn't required.  `!colony_count_by_empire_id.empty()` fits the same purpose.

adrianbroher (changes requested):

adrianbroher:
Sorry, missed that one in the first review.

> If you shouldn't have done so I recommend you try it out a little.

It's true that I didn't play it yet, but the remark was given from a more logical deduction and a 'rule of least surprise' UI design.  Sure, a player can learn the meaning, but it's easier to understand on first guess if it follows the previously mention IHMO logical assumption.

UI/MapWnd.cpp
geoffthemedio: I would argue the variable is substantially clearer and easier to understand at a glance

TheSilentOne1:
> Sure, a player can learn the meaning, but it's easier to understand on first guess if it follows the previously mention IHMO logical assumption.

The current solution can also be explained logically: the addition of circles on the outside (rather than the inside) means an increase of the size/development status of the system; adding the circles on the outside increases the icon's size and thus symbolifies growth.
Which solution works better is something that should be tried out in game. I favour the current one after comparing both.

geoffthemedio:
Looking at it in game, I still think system contents circles should be inside, and supply on the outside. It looks a bit odd that the supply lines sort of pass through the contents rings when entering a system when their colours are different.
circles
Also, the spacing between the rings is oddly close when zoomed in... might need some zoom-dependent tweaking.

o01eg:
Isn't it just duplicate for system name color? May be it better to use it for something that not presented on the map yet? I want to see marked systems where events from Situational Report took place.

TheSilentOne1:
> Looking at it in game, I still think system contents circles should be inside

Okay, I'll change it and do some fine-tuning (so the starlanes connect to the outer circle etc.). It'll be a while, I'm currently a little short on time.

> Isn't it just duplicate for system name color?

Not exactly. If you zoom out the map, the system names disappear and you are no longer able to identify colonies; so this is where the second ring helps. I also think that the second ring makes it easier to read the map at close zoom. As for map markers, see my reply on the forums.

adrianbroher:
Approving code changes so far.

adrianbroher (approved):

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,94 @@ void MapWnd::RenderSystems() {
...
                         CircleArc(circle_ul, circle_lr, 0.0, TWO_PI, false);
+
+                        // systems with neutrals and no empire have a segmented outer circle
+                        if (has_neutrals && !(has_empire_planet)) {
+                            glLineWidth(2.0f);
+                            glColor(ClientUI::TextColor());
+
+                            float segment = static_cast<float>(TWO_PI) / 24.0f;
+                            for (int n = 0; n < 24; n = n + 2) {
+                                CircleArc(outer_circle_ul, outer_circle_lr, n * segment, (n+1) * segment, false);
```
geoffthemedio: I think if you're rendering 12 arc bits, it should be doable with a single render call, rather than using this function.

UI/MapWnd.cpp
```diff
@@ -1939,26 +1939,94 @@ void MapWnd::RenderSystems() {
...
+
+                        // systems with empire planets have an unbroken outer circle; color segments for each empire present
+                        if (has_empire_planet) {
+                            glLineWidth(1.5f);
+
+                            int colonised_planets = 0;
+                            int position = 0;
+
+                            for (std::pair<int, int> it : colony_count_by_empire_id) {
+                                colonised_planets += it.second;
```
geoffthemedio: could move the `{}` to this line, around the code

TheSilentOne1:
I'll get back to this soon.

UI/MapWnd.cpp
TheSilentOne1: I don't know how to do that, can you elaborate?

TheSilentOne1:
I've switched the circles, and also I've made the line width of the inner circle zoom-dependent. It increases when the map is zoomed out, so that systems with colonies can then still be differentiated from empty systems.

UI/MapWnd.cpp
```diff
@@ -1923,26 +1927,27 @@ void MapWnd::RenderSystems() {
...
-            GG::Pt circle_half_size = GG::Pt(circle_size.x / 2, circle_size.y / 2);
-            GG::Pt circle_ul = middle - circle_half_size;
+            GG::Pt circle_ul = icon_middle - (circle_size / 2);
             GG::Pt circle_lr = circle_ul + circle_size;
 
-            GG::Pt circle_distance = GG::Pt(GG::X(4), GG::Y(4));
-            
-            GG::Pt outer_circle_ul = circle_ul - circle_distance;
-            GG::Pt outer_circle_lr = circle_lr + circle_distance;
+            GG::Pt circle_distance_pt = GG::Pt(GG::X(1), GG::Y(1)) * circle_distance;
```
geoffthemedio: use GG::X1 and GG::Y1

default/stringtables/en.txt
```diff
@@ -1248,6 +1248,18 @@ Toggles whether to draw circles around systems.
 OPTIONS_DB_UI_SYSTEM_CIRCLE_SIZE
 Sets size of circles around systems on map, relative to system icon size.
 
+OPTIONS_DB_UI_SYSTEM_INNER_CIRCLE_WIDTH
+Sets the line width of the inner system circle when the map is zoomed in.
+
+OPTIONS_DB_UI_SYSTEM_OUTER_CIRCLE_WIDTH
+Sets the line width of the outer system circle.
+
+OPTIONS_DB_UI_SYSTEM_INNER_CIRCLE_MAX_WIDTH
```
geoffthemedio, line 1257: shouldn't it be bigger when zoomed in?

TheSilentOne1: This doesn't have an effect when the map is zoomed in. But when you zoom out, systems with colonies can be identified by their thicker circle.

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
+                        // systems with empire planets have an unbroken inner circle; color segments for each empire present
+                        if (has_empire_planet) {
+                            float line_width = std::max(std::min(2 / ZoomFactor(), max_inner_circle_width), inner_circle_width);
+                            glLineWidth(line_width);
+
+                            int colonised_planets = 0;
+                            int position = 0;
+
+                            for (std::pair<int, int> it : colony_count_by_empire_id)
+                                { colonised_planets += it.second; }
```
geoffthemedio: de-indent so { aligns with left of for, or just remove the {}

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
+
+                        // systems with neutrals and no empire have a segmented inner circle
+                        if (has_neutrals && !(has_empire_planet)) {
+                            float line_width = std::max(std::min(2 / ZoomFactor(), max_inner_circle_width), inner_circle_width);
+                            glLineWidth(line_width);
+                            glColor(ClientUI::TextColor());
+
+                            float segment = static_cast<float>(TWO_PI) / 24.0f;
+                            for (int n = 0; n < 24; n = n + 2)
+                                { CircleArc(inner_circle_ul, inner_circle_lr, n * segment, (n+1) * segment, false); }
```
geoffthemedio: de indent

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
+                            int colonised_planets = 0;
+                            int position = 0;
+
+                            for (std::pair<int, int> it : colony_count_by_empire_id)
+                                { colonised_planets += it.second; }
+
+                            float segment = static_cast<float>(TWO_PI) / colonised_planets;
+
+                            for (std::pair<int, int> it : colony_count_by_empire_id) {
+                                if (Empire* empire = GetEmpire(it.first)) {
```
geoffthemedio: const Empire* ?

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
+                            glLineWidth(line_width);
+                            glColor(ClientUI::TextColor());
+
+                            float segment = static_cast<float>(TWO_PI) / 24.0f;
+                            for (int n = 0; n < 24; n = n + 2)
+                                { CircleArc(inner_circle_ul, inner_circle_lr, n * segment, (n+1) * segment, false); }
+                        }
+
+                        // systems with empire planets have an unbroken inner circle; color segments for each empire present
+                        if (has_empire_planet) {
```
geoffthemedio: change to `if (!empire_has_planet) continue` and de-indent the following

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
                             }
+                        } else {
+                            glColor(GetOptionsDB().Get<GG::Clr>("UI.unowned-starlane-colour"));
                         }
+
+                        glLineWidth(outer_circle_width);
                         CircleArc(circle_ul, circle_lr, 0.0, TWO_PI, false);
+
+                        // systems with neutrals and no empire have a segmented inner circle
+                        if (has_neutrals && !(has_empire_planet)) {
```
geoffthemedio: would prefer to change to `if (!has_neutrals || has_empire_planet) continue;` and de-indent the following, but that would skip the following if case as well

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
+                            if (planet->Unowned() && !planet->SpeciesName().empty() && planet->CurrentMeterValue(METER_POPULATION) > 0.0) {
+                                has_neutrals = true;
+                            }
+                        }
+
+                        // draw outer circle in color of supplying empire
                         int empire_id = GetSupplyManager().EmpireThatCanSupplyAt(system_icon.first);
                         if (empire_id != ALL_EMPIRES) {
                             if (Empire* empire = GetEmpire(empire_id)) {
                                 glColor(empire->Color());
```
geoffthemedio: seeming has unpredictable effect on colour when the empire isn't found? should probably have an ErrorLogger() note in this case...

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
             if (fog_scanlines
                 && (universe.GetObjectVisibilityByEmpire(system_icon.first, empire_id) <= VIS_BASIC_VISIBILITY))
             {
                 m_scanline_shader.SetColor(GetOptionsDB().Get<GG::Clr>("UI.system-fog-of-war-clr"));
                 m_scanline_shader.RenderCircle(circle_ul, circle_lr);
             }
 
-            // render circles around systems that have at least one starlane, if circles are enabled.
+            // render circles around systems that have at least one starlane, if they are enabled
             if (circles) {
```
geoffthemedio: can this be changed to `if (!circles) continue;` ?

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
                 if (std::shared_ptr<const System> system = GetSystem(system_icon.first)) {
```
geoffthemedio: looks like this could also be made into a !test continue...

UI/MapWnd.cpp
```diff
@@ -1923,42 +1927,112 @@ void MapWnd::RenderSystems() {
...
+                        }
+
+                        // draw outer circle in color of supplying empire
                         int empire_id = GetSupplyManager().EmpireThatCanSupplyAt(system_icon.first);
                         if (empire_id != ALL_EMPIRES) {
                             if (Empire* empire = GetEmpire(empire_id)) {
                                 glColor(empire->Color());
                             }
+                        } else {
+                            glColor(GetOptionsDB().Get<GG::Clr>("UI.unowned-starlane-colour"));
```
geoffthemedio: is this case necessary? what will the colour be if not set to something here?

geoffthemedio:
builds and looks good in game

UI/MapWnd.cpp
TheSilentOne1: It is, otherwise the circle will have a random previously used color (possibly set by the scanline renderer before or some previous system). However in the code above this, calling glColor was not necessary, so I removed that instead.

TheSilentOne1:
Squashed the commits. GTG from my side.

TheSilentOne1:
If there should be no more comments, I'll merge the PR tonight.

Issue #999, opened 2016-09-27
Parts and hulls requiring a strategic resource should require supply


MatGB:
### Description

Currently, Neutronium and Asteroid parts/hulls only require the building player owns a building for the correct resource, meaning building of these parts cannot be prevented by blocking supply, etc.
### Expected Result

If something requires a strategic resource, another player should be able to block supply and deny that resource, so neutronium and asteroid parts should require an active supply connection not just ownership.

This will probably require an AI update, especially for the Asteroid parts and hulls.

Vezzra:
Hrm, that's actually something we should have taken care of this release cycle, but as this requires AI updates it's a bit late for that now. But we should make sure to address this after 0.4.7.

Morlic-fo:
The proposed changes should be fine without an AI update.

I don't think there is any part of the code that assumes that strategic resources are spread globally. It is just assumed to be some normal shipyard for now.

The checks whether or not a part can be built somewhere are fully queried from the server so no issue on that front either.

Vezzra:
That's good to hear, in that case I set this to be required for 0.4.7.

Vezzra:
Fixed by #1324.

PR #1324, opened 2017-02-16
Fix bug #999 ship parts require strategic resources.

LGM-Doyle:
I fixed the issue with the neutronium forge using ResourceSupplyConnected.  I tested this and it works.

I used the same fixed for the asteroid armor and hulls.  However, I don't know if it matter with the current implementation, since the asteroid processor must be in the same system.  I tested it and it didn't make a difference.  However, if the in system requirement changes, then the fix is already in.

### DIFF
--- a/default/scripting/ship_hulls/asteroid/SH_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_ASTEROID.focs.txt
@@ -29,13 +29,6 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
-        ResourceSupplyConnected empire = Source.Owner condition = And [
-             Building name = "BLD_SHIPYARD_AST"
-             Or [
-                 OwnedBy empire = Source.Owner
-                 OwnedBy affiliation = AllyOf empire = Source.Owner
-             ]
-        ]
     ]
     effectsgroups = [
         [[AVERAGE_BASE_FUEL_REGEN]]
--- a/default/scripting/ship_hulls/asteroid/SH_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_ASTEROID.focs.txt
@@ -29,6 +29,13 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
+        ]
     ]
     effectsgroups = [
         [[AVERAGE_BASE_FUEL_REGEN]]
--- a/default/scripting/ship_hulls/asteroid/SH_CAMOUFLAGE_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_CAMOUFLAGE_ASTEROID.focs.txt
@@ -26,13 +26,6 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
-        ResourceSupplyConnected empire = Source.Owner condition = And [
-             Building name = "BLD_SHIPYARD_AST"
-             Or [
-                 OwnedBy empire = Source.Owner
-                 OwnedBy affiliation = AllyOf empire = Source.Owner
-             ]
-        ]
     ]
     effectsgroups = [
         EffectsGroup
--- a/default/scripting/ship_hulls/asteroid/SH_CAMOUFLAGE_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_CAMOUFLAGE_ASTEROID.focs.txt
@@ -26,6 +26,13 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
+        ]
     ]
     effectsgroups = [
         EffectsGroup
--- a/default/scripting/ship_hulls/asteroid/SH_CRYSTALLIZED_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_CRYSTALLIZED_ASTEROID.focs.txt
@@ -35,20 +35,6 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
-        ResourceSupplyConnected empire = Source.Owner condition = And [
-             Building name = "BLD_SHIPYARD_AST"
-             Or [
-                 OwnedBy empire = Source.Owner
-                 OwnedBy affiliation = AllyOf empire = Source.Owner
-             ]
-        ]
-        ResourceSupplyConnected empire = Source.Owner condition = And [
-             Building name = "BLD_SHIPYARD_AST_REF"
-             Or [
-                 OwnedBy empire = Source.Owner
-                 OwnedBy affiliation = AllyOf empire = Source.Owner
-             ]
-        ]
     ]
     effectsgroups = [
         [[AVERAGE_BASE_FUEL_REGEN]]
--- a/default/scripting/ship_hulls/asteroid/SH_CRYSTALLIZED_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_CRYSTALLIZED_ASTEROID.focs.txt
@@ -35,6 +35,20 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
+        ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST_REF"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
+        ]
     ]
     effectsgroups = [
         [[AVERAGE_BASE_FUEL_REGEN]]
--- a/default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
@@ -31,6 +31,13 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
+        ]
     ]
     effectsgroups = [
         [[AVERAGE_BASE_FUEL_REGEN]]
--- a/default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
@@ -31,13 +31,6 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
-        ResourceSupplyConnected empire = Source.Owner condition = And [
-             Building name = "BLD_SHIPYARD_AST"
-             Or [
-                 OwnedBy empire = Source.Owner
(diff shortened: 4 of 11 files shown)


dbenage-cx:
Ref #999 
~~@MatGB Agree with the parts, but being able to use a specialized remote facility (shipyard) to construct a ship (hull) in another location is making my head hurt~~ Edit: I'm mistaken, since these are labeled refineries, and not an actual shipyard facility.

MatGB:
I was going to comment in a few places, appears my description in #999 wasn't as clear as ideal, there are two asteroid shipyards BLD_SHIPYARD_AST & BLD_SHIPYARD_AST_REF. The former allows for the building of asteroids in that system, the latter creates a strategic resource allowing use of Crystal plating and similar.

All Asteroid hulls need to require a BLD_SHIPYARD_AST insystem, some also require a BLD_SHIPYARD_AST_REF insystem, some merely need one to be owned (and thus in supply). There are several superfluous additions in this for the first yard, it's only the reformation facility that sometimes requires a supply commitment replacing the current ownership requirement.: there might be some hulls/parts that would be better served simply needing one of them rather than insystem but I haven't given that any thought at all. 

Sorry the initial description wasn't as clear, I'd been thinking of Neutronium as a specific then remembered Crystal while writing it up.

Vezzra:
As the corresponding issue is tagged as "tweak", not "bug", I changed the tagging of this PR accordingly. Wasn't sure which way to go (set the issue to "bug" or the PR to "tweak"), but since the original implementation wasn't really "buggy", but the design somewhat counter-intuitive, I decided to go with "tweak". Feel free to change both issue and PR to "bug" if you think that's more fitting.

Vezzra:
Looks good, and from what I can see, should take care of all cases raised in #999. LGTM.

Vezzra (approved):

LGM-Doyle:
I've made to changes to make it work as I think @MatGB wanted it to work.  Asteroid reformation now works as a galactic resource and if a ship yard has a supply connection to an asteroid reformation then it can build the asteroid armor types.

I made two changes:
- I required the asteroid shipyard to built in a system with an empire controlled regular shipyard.  The asteroid shipyard works like the other ship yard improvements, except it is build on an asteroid belt.
- I allowed asteroid reformation to be build independently of the asteroid ship yard.  This allows empires to build asteroid reformation anywhere and use supply to connect it to a shipyard.

Dilvish-fo:
OK, from my cursory review of the relevant AI code it looks to me that, although the AI *expects* it can start building the asteroid shipyard even if it has not yet finished a local regular shipyard, I think it should still wind up getting the job done with this extra requirement.  But it would be nice to know that someone had tested and confirmed that -- in your testing of this did you observer the AI successfully using asteroid hulls?

Please keep in mind, also, anytime you change requirements for things like this that the AI uses, to either tell us if you have reviewed the AI code and believe it compatible, or specifically flag that an AI review is needed, to help make sure it doesn't happen to slip by without the necessary review, thanks.

MatGB:
> I required the asteroid shipyard to built in a system with an empire controlled regular shipyard. 

Not sure about this: I don't like Asteroid hulls and rarely use them unless testing, partially because they're so sloooow but also because getting them setup is a pain unless you get lucky with an asteroid belt in a homeworld system.

For a colony, the taskflow is normally build shipyard and outpost base, use base to grab asteroid belt then start building the processor, that's time consuming enough, you can get around it by sending in two outpost ships: in general I'm in favour of speeding up the ability to start building ships in colony worlds (hence the changes made to other lines to let advanced shipyards be built if a predecessor is enqueued).

I don't oppose this, but I don't think it's needed, I think it'd make me even less likely to want to throw rocks around than I already am.

LGM-Doyle:
@Dilvish-fo the AI builds asteroid hulls and Asteroid processor in both master and this branch.  I tested by building SuperTester take over and running the game until I was taken over around tun 140.  In neither case does the AI build asteroid reformation, before my empire is wiped out.

@MatGB, this change relaxes the requirements to build the asteroid reformation.  You can build it anywhere and then use the resource to build cheap asteroid armor at any supply connected shipyard.  I thought you wanted asteroid reformation to behave like a resource after it was built.

MatGB:
I do, but not sure we need to restrict further the basic processor. However, I'm not that bothered (it won't affect me much after all), so if we try it and get feedback that's good enough.

dbenage-cx:
Looks to work great, probably need some stringtable updates for the affected parts/hulls.

LGM-Doyle:
I've added stringtable changes to clarify that the asteroid reformation processor can be stand-alone.

Vezzra:
> I made two changes:
>
> * I required the asteroid shipyard to built in a system with an empire controlled regular shipyard. The asteroid shipyard works like the other ship yard improvements, except it is build on an asteroid belt.
> * I allowed asteroid reformation to be build independently of the asteroid ship yard. This allows empires to build asteroid reformation anywhere and use supply to connect it to a shipyard.

I respectfully object. This goes beyond the scope of what has been proposed in #999, which is a fix to what can be described as a design flaw. Whereas the modifications you describe here are more or less changes to the basic design of the Asteroid Processer and the Asteroid Reformation Processor.

Both aren't considered extensions to the Basic Shipyard (like all the other shipyard upgrade buildings), but more so facilities that do some preprocessing of asteroids so that a shipyard can make ships (or armor) out of them. Which is why they are not built on the same planet as the shipyard, but on the asteroid belt, and why it's ok to build them even if there is no shipyard finished in the system yet.

However, the Asteroid Reformation Processor actually *is* considered an upgrade to the Asteroid Processor the way the shipyard upgrade buildings are considered upgrades to the Basic Shipyard, and therefore of course requires an Asteroid Processor being present at the same location.

The Asteroid Reformation Processor not only (I'd even say not even primarily) provides the means to make Rock Armor, but is required to built the more advanced asteroid hulls. The idea being that the more advanced asteroid hulls require a more advanced asteroid processing facility, and like the basic one, the more advanced one needs to be in the same system as the shipyard.

IIRC it's only the Rock Armor ship part that does not require the Asteroid Reformation Processor to be in the same system as the shipyard which builds the ship that uses this part. The idea being that these parts can be distributed via supply lines, while the prefabricated "raw" hulls can't (being too large or whatever).

That's the general idea/concept behind these buildings, their design and implementation. I'm not saying that changes to the design of the asteroid processing facilities can't be discussed and changed if we come to the conclusion that a different apporach is better/more interesting, but IMO that's beyond the scope of a PR which intends to address #999.

Also, for the AI concerns @Dilvish-fo mentioned, I'd rather postpone such a change until after the release.

LGM-Doyle:
My goal was to knock an issue off the 0.4.7 list.  To do that should I pare this down to just the first commit, or the first two commits?

However, this PR as is increases the AIs use of asteroid hulls.  From the observations I took with the two games that I ran for about 140 turns, the requirement to build the asteroid shipyard in a system with a regular ship yard means that every asteroid shipyard the AI builds can be used to build ships.  The AIs used a greater variety of asteroid hull types in the game with this PR.

Vezzra:
> My goal was to knock an issue off the 0.4.7 list. To do that should I pare this down to just the first commit, or the first two commits?

First two. Both are required to address #999 completely (the first commit only addresses the Neutronium stuff).

I suggest splitting the rest off to a separate PR. Thus we can merge this one here and close #999, and continue discussion on the other changes you proposed. Preferably in a forum thread which gets linked to in the new PR.

Wrt your observation that these changes improved the AIs usage of asteroid hulls, this strikes me as odd - didn't the AI already restrict building Asteroid Processors and Asteroid Reformation Processors to systems where it has a shipyard? They don't make much sense without one, except maybe in the case where you only have asteroid belts in systems where you don't have (or can't have) a shipyard, but still want to use Rock Armor. That however seems like an edge case.

But I'm probably the wrong person to ask - @freeorion/ai-team, any idea what could be the reason for @LGM-Doyle's observations?

Anyway, as a player I definitely prefer being able to build the Asteroid Processor and the Basic Shipyard in parallel. The asteroid hulls are already very slow, at least you can get their shipyards up and running a bit faster. The way these buildings are designed and work make much sense to me, I don't want them to be changed as a "byproduct" of a fix, without proper discussion.

LGM-Doyle:
I've trimmed it down to the first two commits.

Vezzra:
Ok, looks good to me. @MatGB, can you give this trimmed down version a quick look if it's ok to merge?

default/scripting/ship_hulls/asteroid/SH_ASTEROID.focs.txt
```diff
@@ -29,6 +29,13 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
+        ]
```
MatGB: Don't get the point of this addition: the previous clause requires BLD_SHIPYARD_AST to be in the same system, this new clause requires there be one in supply, the only way the latter can apply without the former is if there's an enemy fleet blockading you: I'd have zero problem with that as a requirement in which case add that as an AND check to the first clause.

default/scripting/ship_hulls/asteroid/SH_CAMOUFLAGE_ASTEROID.focs.txt
```diff
@@ -26,6 +26,13 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
+        ]
```
MatGB: Same again

default/scripting/ship_hulls/asteroid/SH_CRYSTALLIZED_ASTEROID.focs.txt
```diff
@@ -35,6 +35,20 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
```
MatGB: and again

default/scripting/ship_hulls/asteroid/SH_CRYSTALLIZED_ASTEROID.focs.txt
```diff
@@ -35,6 +35,20 @@ Hull
...
+             ]
+        ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST_REF"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
```
MatGB: This is slightly different as it's the reformation processor but it's basically the same problem, it already requires it be in system.

default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
```diff
@@ -31,6 +31,13 @@ Hull
                 OwnedBy empire = Source.Owner
             ]
         ]
+        ResourceSupplyConnected empire = Source.Owner condition = And [
+             Building name = "BLD_SHIPYARD_AST"
+             Or [
+                 OwnedBy empire = Source.Owner
+                 OwnedBy affiliation = AllyOf empire = Source.Owner
+             ]
+        ]
```
MatGB: and again

MatGB (commented):
I'm very happy with the Neutronium changes, and the Rock/Crystal armour changes, but the asteroid hull changes are all fairly pointless, if you have to have the building insystem then it's in supply unless blockaded, if we want to stop building if there's a blockade (happy for this) then it can surely be simply added as a clause to the initial insystem requirement?

default/scripting/ship_hulls/asteroid/SH_ASTEROID.focs.txt
Vezzra: To quote @LGM-Doyle:

> I used the same fixed for the asteroid armor and hulls. However, I don't know if it matter with the current implementation, since the asteroid processor must be in the same system. I tested it and it didn't make a difference. However, if the in system requirement changes, then the fix is already in.

Obviously the rationale is that in case the "in system" condition gets changed/removed (although I don't see that happening) a fix to at least still require those buildings to be within supply will be already in place.

I don't have a problem with that, it's just pointless IMO. Because if someone would make such changes, I'd expect them to put something else in place.

However, tweaking this additional condition so that you can't continue to build asteroid hulls in blockaded systems actually makes a lot of sense (after all, the rationale is that these asteroid processing facilites "harvest" and prepare asteroids, then these prepared things get transported to the shipyard, where ships are made out of them - a blockade should disrupt that).

The question is if the AI needs to be adjusted to this change - @freeorion/ai-team?

LGM-Doyle:
I've removed all of the requirements that the asteroid ship hulls require the asteroid shipyard to be supply connected to the shipyard building the ship using the asteroid hull.

Issue #1327, opened 2017-02-17
FOCS / AI Scripting muddle between IsArmed and HasFighters


Dilvish-fo:
The presence of fighters had only been exposed to the AI for fleets, not for ships and ship designs, and there are various situations where the AI relies on analysis of individual ships and designs.   The incomplete exposure for Fighters was causing the AI to leave carrier fleets at passive rather than at aggressive if they did not also have a direct fire weapon present.

Although it might be that some of those spots could beneficially be changed over to considering characteristics at the fleet level, it seems that the information should at least also be available at the Ship and ShipDesign level, and so I am concurrently submitting a PR to resolve the AI issue in that manner.

It's possible that this bug may at least partly have arisen because while the C backend code considers IsArmed and HasFighters to be distinct (though closely related) things, the FOCS "Armed" Condition for ships conflates the two (and there is no independent Condition for either), so anyone focused on scripting and FOCS would be less likely to be aware of the distinction.  It could be that the current Armed condition should be changed en-masse to something like "HasWeapons" and then new conditions for Armed and HasFIghters could be exposed that directly track the corresponding characteristics.   I guess there would still be a question about whether HasWeapons should include means for bombardment.

Vezzra:
@Dilvish-fo, what milestone do you suggest? Should this be considered required or optional for the release, or even be postponed for afterwards?

Dilvish-fo:
Well, the the part that was definitely buggy has now been fixed by the PR referenced above, and so the remaining bit I see as purely optional -- it's possible that it @geoffthemedio   could prefer that bit stay as it is, although I hope not.  Since that is a design question I've opened [a forum question](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10425) on it (which is my understanding of our current preferred practice).

geoffthemedio:
Tweak as you see fit...

Vezzra:
Ok, in this case I tag this as "post 0.4.7", as the proposed changes sound like they'd require a quite major revision of the affected content scripts, and I think we better leave that for after the release. There is so much other stuff that needs to be addressed for 0.4.7, lets not add too much to the pile... 😉

Vezzra:
> Since that is a design question I've opened a forum question on it (which is my understanding of our current preferred practice).

Yup, I think we better keep all design related discussions to the forum, and only discuss implementation related stuff in the issue/PR comment sections. As much as possible at least, we frequently fail to stick to that... 😉

PR #1328, opened 2017-02-17
Fix AI recognition of carrier military role

Dilvish-fo:
As noted in Issue #1327, there was incomplete exposure of Fighter info to the AI, leading to some problems, which this PR fixes (but this PR does not quite address all that is addressed by that issue).

Although it might be that some of those spots could beneficially be changed over to considering characteristics at the fleet level, it seems that the information should at least also be available at the Ship and ShipDesign level, and so this PR resolves the AI issue in that manner.


Morlic-fo (approved):
Seems fine.

For reference:
https://github.com/freeorion/freeorion/commit/b6b398c1501da4f053049a5dbffbb1361f4dfca7#diff-dd18c87c8be49826259d4039c3caf139L911
split the concepts which led to the bugs / insufficient checks in the AI.

geoffthemedio:
C++ part, small as it is, seems fine...

PR #1329, opened 2017-02-17
Fix cur_best_military_design_rating() 

Dilvish-fo:
Some time ago (a long while) we inadvertently started using a cost-adjusted military ship rating (that includes some additional embellishments as well) in a place where a more plain combat rating was expected by the production priority planning code.  This PR fixes that by backing out the cost adjustment in this particular area.

### DIFF
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -97,10 +97,15 @@ def cur_best_military_design_rating():
         return _best_military_design_rating_cache[current_turn]
     priority = PriorityType.PRODUCTION_MILITARY
     if priority in _design_cache:
+        # TODO: rework the below code, CombatRatingsAI and possibly  WarShipDesigner so that the below uses a rating
+        # provided by CombatRatingsAI (so as to be more suitable for its actual use and to avoid making any assumptions
+        # about how the WarShipDesigner may apply cost adjustments to its rating).
         if _design_cache[priority] and _design_cache[priority][0]:
             rating, pid, design_id, cost = _design_cache[priority][0]
+            exponent = foAI.foAIstate.character.warship_adjusted_production_cost_exponent()
+            rating = max(rating*(cost**exponent), 0.001)
             _best_military_design_rating_cache[current_turn] = rating
-            return max(rating, 0.001)
+            return rating
         else:
             return 0.001
     else:


default/python/AI/ProductionAI.py
```diff
@@ -97,10 +97,15 @@ def cur_best_military_design_rating():
         return _best_military_design_rating_cache[current_turn]
     priority = PriorityType.PRODUCTION_MILITARY
     if priority in _design_cache:
+        # TODO: rework the below code, CombatRatingsAI and possibly  WarShipDesigner so that the below uses a rating
+        # provided by CombatRatingsAI (so as to be more suitable for its actual use and to avoid making any assumptions
+        # about how the WarShipDesigner may apply cost adjustments to its rating).
```
Morlic-fo, line 102: I agree. This is definitely the wrong place to calculate the rating "backwards".

As far as I can see, the rating you calculate here may not even be correct because of the ```_adjusted_production_cost``` used when calculating the rating:
 https://github.com/freeorion/freeorion/blob/master/default/python/AI/ShipDesignAI.py#L1457
and also some of the non-combat-stuff (fuel, speed) that are not relevant to the rating as interpreted by ```cur_best_military_design_rating()```

In my opionion, the cleanest approach is to give the ```WarShipDesigner``` an interface to calculate and return a combat rating based on the calculations in ```CombatRatingsAI```.
That rating may or may not be used within the ```rating_function``` but is not equivalent to it (as it neglects e.g. fuel)

Morlic-fo (commented):
While the value calculated by this PR seems not to be correct either, it should be significantly more accurate than before...

As the ```# TODO``` exists, I am fine with merging this as is and update it later.
If you would provide a "cleaner" implementation, that would be preferred of course.

default/python/AI/ProductionAI.py
Dilvish-fo: Ah, yes, I had overlooked that fleet upkeep contribution, and since it can get pretty big I think I should back it out also, as part of this stopgap measure.  

I submitted this stopgap rather than trying to produce the cleaner long-term solution now, because we seem on a tight-ish timeframe for 4.7, I thought it important to get at least some improvement in for 4.7, and I am not yet back up to speed enough on ShipDesigner and CombatRatingsAI to do that myself.  But I might be able to get there quick enough.  Let's see how this planning discussion goes....

Part of my uncertainty about adding this into WarShipDesigner was just how to make the necessary information available here where we need it without having to run optimize_design() again.  It could be that optimize_design could save two ratings values for the design cache-- in addition to the current one that is adjusted for cost and other considerations, it could save a 'rated_strength' value, which for warships would be the rating a la CombatRatingsAI, for troop ships would be the effective number of troops, etc.  The places that use the design cache would need to be adjusted to unpack the extra value even if they don't use it.  How does that approach sound to you?

Morlic-fo: > It could be that optimize_design could save two ratings values for the design cache-- in addition to the current one that is adjusted for cost and other considerations, it could save a 'rated_strength' value, which for warships would be the rating a la CombatRatingsAI, for troop ships would be the effective number of troops, etc.

Sounds reasonable.

MatGB:
@Dilvish-fo , following your suggestion on the forum, I'm playing with the three PRs you listed, so here's an error message that might be caused by any of them 
`AI_5: AI_Error: AI script error in "calculate_priorities": "Python argument types in
    universe.getPlanet(universe, float)
did not match C++ signature:
    getPlanet(Universe, int)"`
[AI_5.log.zip](https://github.com/freeorion/freeorion/files/797464/AI_5.log.zip)
Happened a few turns earlier to a different AI, seems to only happen when I've captured a planet from the AI then it comes up in the next turn.

Morlic-fo:
Yes, the error messages are related to an error in one of the PRs pointed out here:
https://github.com/freeorion/freeorion/pull/1352/files#r102213715

Morlic-fo:
#1387 implements the cleaner approach discussed here and in the forums, i.e. let the rating be calculated by the relevant stats of the design instead of trying to reverse-engineer the rating.

PR #1332, opened 2017-02-17
Kobuntura should prefer the industry focus

l29ah:
as they are better at industry, as such this avoids a dumb
micromanagement of switching the focus after colonization

### DIFF
--- a/default/scripting/species/SP_KOBUNTURA.focs.txt
+++ b/default/scripting/species/SP_KOBUNTURA.focs.txt
@@ -15,7 +15,7 @@ Species
         [[HAS_ADVANCED_FOCI]]
     ]
 
-    preferredfocus = "FOCUS_RESEARCH"
+    preferredfocus = "FOCUS_INDUSTRY"
 
     effectsgroups = [
         [[GREAT_INDUSTRY]]


MatGB:
Disagree. They were specifically selected as a race with an opposed preference, they're one of the most powerful natives as a result of Self Sustaining metabolism and this is an intentional balance and flavour decision.

Not everyone prefers doing what they're best at. 

Not closing this, but we've had discussions on the forum on this subject and as this is as much a balance/flavour issue it should be discussed there first, my instinct is to reject completely but I'd be open to persuasion if the relevant arguments are countered.

l29ah:
On Fri, Feb 17, 2017 at 09:57:42AM -0800, Mat Bowles wrote:
> Disagree. They were specifically selected as a race with an opposed preference, they're one of the most powerful natives as a result of Self Sustaining metabolism and this is an intentional balance and flavour decision.
> 
> Not everyone prefers doing what they're best at. 
> 
> Not closing this, but we've had discussions on the forum on this subject and as this is as much a balance/flavour issue it should be discussed there first, my instinct is to reject completely but I'd be open to persuasion if the relevant arguments are countered.

So it's a design feature that it's power is given to the player that is devoted to dutifully setting the focus each time the colony is built? I don't see any other penalties, as the "happiness" thing can just be ignored for every case i can think of.

-- 

MatGB:
Focus changing isn't micromanagement. Decisions about what a colony is meant to be doing are one of the things the game is meant to be about.

Happiness is a new feature that is barely implemented yet: one of the things it will be doing more often is affecting the ability to do things like build ships.

And I regularly want my Kobuntura colonies to be on research in the earlier stages, they're good at it compared to a LOT of other species.  

But, like I say, this can be discussed, with others, on the forums, as reversing deliberate design and balance decisions isn't something we do in PRs.

MatGB:
[FreeOrion • View topic - Chato default focus](http://www.freeorion.org/forum/viewtopic.php?f=28&t=9964&p=87532#p87532)

Dilvish-fo:
Seems to me that this should be closed, with the status invalid (I think).  We have no current intention to change this, and any such change would need to be discussed and agreed on the boards first, so at this time this PR seems invalid.

MatGB:
Agreed, I've been trying to summon up the willpower to find it and close it for a bit now.

I do think, and we could open an issue for it, that there needs to be a way, especially in the mid to late game, to override 'preferred' settings when colonising but that can be done elsewhere in an empire management system (possibly via Objects), it's an alternative solution to the perceived problem that makes more sense longer term.

l29ah:
Reopen plz.

l29ah:
Abandoned (and barely touched at all) features should have no place in releases, especially when they degrade playing experience.

Dilvish-fo:
> and any such change would need to be discussed and agreed on the boards first, so at this time this PR seems invalid.

the boards may be down at the moment, but given that it's been over a year since this issue was closed I think there is no great urgency

MatGB:
A focus override would be a good feature, but giving one of the most
powerful non playable species a reverse preference was a deliberate balance
decision as discussed before.

Give a good reason to up their power or spec out a feature for what you
actually want.


-- 
Mat Bowles
07855307993

l29ah:
Their power won't be upped as the happiness cap doesn't affect anything. It only adds micromanagement.

Voker57:
> and any such change would need to be discussed and agreed on the boards first, so at this time this PR seems invalid.

This issue is not related to balance. Happiness does not work, so balance does not change.

l29ah:
apply plz

MatGB:
Not my decision any more but unless they're given a different balancing thing then no. Introducing an empire wide override makes some sense but we should have more species that like doing what they're not best at, not none 

Both balance and flavour went into the decision 

⁣--
Mat Bowles 

Get BlueMail for Android ​

l29ah:
On Tue, Dec 31, 2019 at 08:55:25AM -0800, Mat Bowles wrote:
> Not my decision any more but unless they're given a different balancing thing then no. Introducing an empire wide override makes some sense but we should have more species that like doing what they're not best at, not none 
> Both balance and flavour went into the decision 

Regarding balance everything was told in the previous messages, yours doesn't add nothing new.
Do you mean the species were supposed to come with a "extremely dumb micromanagement" flavour?

-- 

Vezzra:
The basic problem here is, that the supposed nerf by having Kobuntura prefer what they are not best at does not work - simply because the necessary happiness mechanic isn't in place yet (or better, what is in place is so rudimentary that it does not provide the desired effect).

Which effectively results in not providing the interesting choice it is supposed to provide, but only forces the player to do a lot of annoying micromanagement, so I understand @l29ah's issue with the current state of things.

Nevertheless, this discussion shouldn't take place here, but on the forum. @l29ah, can you open a corresponding thread on the forum, and link to it here?

TheSilentOne1:
@MatGB: you're alive! That's awesome! Seriously, I was a bit worried. All the best to you, it'd be great if you'd someday hang out with us again! We have lots of fun playing multiplayer games on o01eg's long turn server.

Oberlus:
Can we apply this PR and revert it only when influence is finished and merged?
It is so annoying to have to change focus of every Kobuntura world...
I would not say the same if they were a CannotColonize species, but that's not the case.

geoffthemedio:
> Can we apply this PR and revert it only when influence is finished and merged?

seems reasonable

PR #1338, opened 2017-02-19
new galaxy shape images

TheSilentOne1:
Originally done by sam, redone by me using sam's style. This was necessary because the background of the original pictures was jumping and the galaxy icons weren't properly centered, which couldn't be fixed without the source files.
Uses picture "Andromeda Galaxy (with h-alpha).jpg" by Adam Evans (http://www.flickr.com/photos/8269775@N05) released under CC 2.0.


TheSilentOne1:
(If there's no objections, I'll merge the PR in 1-2 days.)

Vezzra:
Just to be sure: CC 2.0 is compatible with CC BY-SA 3.0?

Vezzra:
I wouldn't consider this as critical for the release, but as you're probably going to merge in a few days anyway, I guess we can leave it at that. However, if objections are raised and a discussion starts, please "downgrade" to "optional for 0.4.7".

TheSilentOne1:
> Just to be sure: CC 2.0 is compatible with CC BY-SA 3.0?

Yes, that should be ok: https://creativecommons.org/faq/#can-i-include-a-work-licensed-with-cc-by-in-a-wikipedia-article-even-though-they-use-a-cc-by-sa-license

> However, if objections are raised and a discussion starts, please "downgrade" to "optional for 0.4.7".

Sure.

geoffthemedio:
If you're taking art or other content from someone who didn't submit it directly to freeorion for use in the game, then there probably needs to be an explicit credit for it added, not just in the credits list as "graphics" or similar. I'd add a separate section saying "incorporated" or "included content" and then list specific things that were used, who they were made by, and probably also include a URL to them, and possibly more: https://creativecommons.org/licenses/by/2.0/#
````
If supplied, you must provide the name of the creator and attribution
parties, a copyright notice, a license notice, a disclaimer notice, and a
link to the material. CC licenses prior to Version 4.0 also require you
to provide the title of the material if supplied, and may have other
slight differences.
````

>> Just to be sure: CC 2.0 is compatible with CC BY-SA 3.0?
>
> Yes, that should be ok: https://creativecommons.org/faq/#can-i-include-a-work-licensed-with-cc-by-in-a-wikipedia-article-even-though-they-use-a-cc-by-sa-license

That seems to indicate that CC-BY stuff can be released under CC-BY-SA, but does it also mean CC 2.0 stuff can be rereleased under a later CC license?

**Comment by TheSilentOne1** [comment]
> That seems to indicate that CC-BY stuff can be released under CC-BY-SA, but does it also mean CC 2.0 stuff can be rereleased under a later CC license?

See here: https://wiki.creativecommons.org/wiki/License_Versions
Go to "General license features", "Compatibility mechanism in BY-SA licenses"; for CC 2.0 it states: "same license, **later versions**, or ports". So if they're "compatible" I think we're good?
I've added a section "included content" to the credits that names the author, the used file and a link to the author's web page.

File: default/credits.xml [hunk 1]
```diff
@@ -132,5 +132,8 @@
       <PERSON name="Jean-Charles Schmidt"     nick="Ouaz"             task="French Translation"/>
       <PERSON name=""                         nick="evzorge"          task="Russian Translation"/>
     </GROUP>
+    <GROUP name="INCLUDED CONTENT">
+	  <PERSON name="Adam Evans"               nick="http://www.flickr.com/photos/8269775@N05"   task="Andromeda Galaxy (with h-alpha).jpg"/>
```
geoffthemedio: spaces, not tabs

might need to specify that the content has been modified

geoffthemedio:
Anyone else have concerns about this attribution method?

adrianbroher:
In general I would consider this sufficient.  But maybe we should introduce some new XML syntax for this kind of stuff to the credits.xml.

I suggest adding a `RESOURCE` tag, which in turn contains the attributes `author`, `license`, `title`, `source`, `notes`:

* **author**: Obviously the resource author.
* **license**: Obviously the license the resource was released under.  Even if our assets are CC-BY-SA 3.0 the source maybe licensed under a different, but compatible license.
* **title**: The title the resource was released under.
* **source**: A hyperlink to the original resource.  Can be linked with the title as click text.
* **notes**: Additional notes like `modified` or `background of galaxy shapes selection`.

Except the `notes` every entry should be mandatory.

geoffthemedio:
Sounds reasonable. Might be easier to add the credits text to a pedia page, where links are already supported, rather than adding this to the credits, where it doesn't work very well, particularly as they are scrolling continuously.

TheSilentOne1:
Alternatively, we could add a external_resources.txt (or similar) file to the default folder. That's probably better than naming the authors of external material in the credits, which might create the impression that they actively contributed their work. And it's probably also a better place than the pedia (which is for gameplay information after all).

adrianbroher:
> Alternatively, we could add a external_resources.txt (or similar) file to the default folder.

Why yet another file?  Where should the attribution happen? 

>  That's probably better than naming the authors of external material in the credits, which might create the impression that they actively contributed their work.

Giving credits has nothing to with contributing.

> And it's probably also a better place than the pedia (which is for gameplay information after all).

I agree here.  Keep the pedia in universe.

geoffthemedio:
I meant it could also be included in the pedia, where links can be clickable, in addition to being in the credits.

The point of having a new section in the credits is to make it clear stuff was used without being actively contributed.

TheSilentOne1:
> I suggest adding a RESOURCE tag, which in turn contains the attributes author, license, title, source, notes

This is also fine by me. However, I don't know how to do that. Can you show me?

geoffthemedio:
> Can you show me?

Relevant code is here:
https://github.com/freeorion/freeorion/blob/master/UI/IntroScreen.cpp#L121

Looks easy enough to understand.

TheSilentOne1:
I added a resource tag as @adrianbroher suggested. From my side gtg, shall I squash the commits and merge?

credits

UI/IntroScreen.cpp
```diff
@@ -117,28 +117,52 @@ void CreditsWnd::DrawCredits(GG::X x1, GG::Y y1, GG::X x2, GG::Y y2, int transpa
...
+                if (0 == item.Tag().compare("RESOURCE")) {
+                    if (item.attributes.count("author"))
+                        credit += item.attributes.at("author");
+                    if (item.attributes.count("title")) {
+                        credit += "<rgba 153 153 153 " + std::to_string(transparency) + "> - ";
+                        credit += item.attributes.at("title");
+                        credit += "</rgba>\n";
+                    }
+                    if (item.attributes.count("license"))
+                        credit += "Released under " + item.attributes.at("license");
```
geoffthemedio: use the stringtable for in-code human-readable text

default/credits.xml
```diff
@@ -132,5 +132,9 @@
       <PERSON name="Jean-Charles Schmidt"     nick="Ouaz"             task="French Translation"/>
       <PERSON name=""                         nick="evzorge"          task="Russian Translation"/>
     </GROUP>
+    <GROUP name="EXTERNAL RESOURCES">
+      <RESOURCE author="Adam Evans"  license="CC-BY 2.0"  title="Andromeda Galaxy (with h-alpha).jpg"
+	            source="http://www.flickr.com/photos/8269775@N05"  notes="used for galaxy picture background"/>
```
geoffthemedio: tab to space

geoffthemedio:
ok then

PR #1340, opened 2017-02-19
Stopgap solution for spatial flux hull AI adaptation

### LINKED ISSUE #1111: Update AI for Spatial Flux Hull

Morlic-fo:
Good arguments were raised against #1304 so let's just go with strict hardcoded values for the 0.4.7 release until some framework is etablished to handle conditional effects.

Resolves #1111

### DIFF
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -1005,6 +1005,17 @@ class ShipDesigner(object):
             if is_hull and AIDependencies.DETECTION not in tokendict:
                 self.detection += AIDependencies.BASE_DETECTION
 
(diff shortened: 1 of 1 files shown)


Vezzra:
Resolves #1111 (should be mentioned here too, after all that is what #1304 has been about).

Dilvish-fo (approved):
Looks good to me

PR #1341, opened 2017-02-19
Increase negative bias for costy invasion operations

Morlic-fo:
This PR contains the cost-bias part of #1309.

Partial progress towards issue #1013.

### DIFF
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -6,6 +6,7 @@ import traceback
 from universe_object import Planet
 from ShipDesignAI import get_part_type
 from AIDependencies import INVALID_ID
+import AIDependencies
 __designStats = {}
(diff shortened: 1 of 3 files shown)


Dilvish-fo (approved):
Because of how you are using the normalized cost here, I am inclined to think that perhaps it should have a floor of something like 5 (or 6 or 7) rather than 1, so that the adjusted troop cost doesn't get adjusted down below at least about 1/2 the actual cost. *(And it would probably be more clear to leave the normalized cost calc itself the same and just apply this higher floor as part of the ```cost_score``` calculation.)*  But we can always continue to experiment with that later, I am OK with this going in as-is.

Vezzra:
@Morlic-fo, needs rebase, after that I guess you can go ahead and merge, as @Dilvish-fo already approved this PR.

PR #1342, opened 2017-02-19
Fix bug #1330 ~ServerApp() is never called.

### LINKED ISSUE #1330: ~ServerApp never called
### Environment
* **FreeOrion Version**: 0.4.6+
* **Operating System**: Fedora 25
* **Graphic card used**: N/A
* **Fetched as** 
  * Compiled from source b7ac413 

When resigning or exiting, `ServerApp::~ServerApp` never appears to execute.

Scenario 1 ("resign"):
*  `freeorion -q`
* Menu -> Resign

Scenario 2 ("wm_close"):
* `freeorion -q`
* Close via window (The "X" typically at the top right)

In neither scenario is there a log message [`Server::~Server`](https://github.com/freeorion/freeorion/blob/master/server/ServerApp.cpp#L161)

In scenario 2, `CleanupAIs()` is not called.  Last server log is from [`ServerFSM.cpp:241`](https://github.com/freeorion/freeorion/blob/master/server/ServerFSM.cpp#L241)
This sometimes results in ai clients and/or server remaining after exiting the game.

[logs_backtrace.zip](https://github.com/freeorion/freeorion/files/782706/logs_backtrace.zip)

Possibly related #1092
(optional for release?)

LGM-Doyle:
This PR fixes [#1330 ~ServerApp never called](https://github.com/freeorion/freeorion/issues/1330).

The problem was that the ServerApp (and the AIClient) were calling `Exit(1)` which called `exit(1)` in the normal shutdown of the server via a user message.  This in turn throws an exception, which because it is at the top level calls `terminate()` which the C++ standard require to exit the application without unrolling the stack.  Hence `~ServerApp()` is never called.

The solution is to catch the normal exit conditions.  So I added `NormalExitException` to the `AppInterface` to be thrown and caught during normal operation.

The human client is immune to this problem, because it already uses a similar construction within the SDLGUI code. 

I expanded the meaning of normal exit to include crashed python environment.  This condition is adequately reported in the log files, there is no information in any stacktrace and preventing the application from exiting cleanly is unwarranted.

### DIFF
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -99,7 +99,9 @@ void AIClientApp::operator()()
 
 void AIClientApp::Exit(int code) {
     DebugLogger() << "Initiating Exit (code " << code << " - " << (code ? "error" : "normal") << " termination)";
-    exit(code);
+    if (code)
+        exit(code);
+    throw NormalExitException();
 }
 
 int AIClientApp::EffectsProcessingThreads() const
@@ -142,6 +144,8 @@ void AIClientApp::Run() {
                     throw;
             }
         }
+    } catch (const NormalExitException&) {
+        // intentionally empty.
     } catch (boost::python::error_already_set) {
         HandlePythonAICrash();
     }
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -91,7 +91,7 @@ AIClientApp::~AIClientApp() {
     if (Networking().Connected())
         Networking().DisconnectFromServer();
 
-    DebugLogger() << "Shutting down " + PlayerName() + " ai client.";
+    DebugLogger() << "Shut down " + PlayerName() + " ai client.";
 }
 
 void AIClientApp::operator()()
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -170,7 +170,10 @@ void ServerApp::operator()()
 
 void ServerApp::SignalHandler(const boost::system::error_code& error, int signal_number) {
     if (!error)
-        Exit(1);
+        throw NormalExitException();
+
+    ErrorLogger() << "Exiting due to OS error (" << error.value() << ") " << error.message();
+    Exit(1);
 }
 
 void ServerApp::Exit(int code) {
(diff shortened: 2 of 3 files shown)


dbenage-cx:
Thank you for looking into and addressing this.

Successfully resolves the first scenario (resign), but not the second (close).
In the second scenario, the tails of logs do not look different from the initial logs.
Maybe this is working as intended, since process.cpp:231 is sending sigkill in this case.
Probably not the correct route, but potentially a different issue.

LGM-Doyle:
I don't have any chrome on my freeorion.  So I didn't test the second scenario.  If it is not tested, it probably doesn't work.

I'll look into it soon.

geoffthemedio:
If this resolves at lease some cases, and doesn't cause any problems, then it should probably be merged. The other cases can be handled in a separate pull request.

adrianbroher:
I know it's not your code, but uh.

Exceptions for flow control?

geoffthemedio:
> Exceptions for flow control?

It's not unheard of... Eg. the boost graph traversal code uses an exception to indicate success: https://github.com/freeorion/freeorion/blob/master/universe/Pathfinder.cpp#L355  (unless that doesn't count as flow control)

adrianbroher:
> It's not unheard of... 

I know.  I still consider it bad style in general.

LGM-Doyle:
@adrianbroher I agree.  There is also a member function `Exit()` that then calls system `exit()` and never returns, which is scattered throughout the code base.  Some have comments stating the `Exit()` never returns and others do not.  Just throwing an exception would clearer.  Nobody expects `throw something` to return.

The mixture of exceptions, `Exit` and  `exit`makes it unclear when and what resources are guaranteed to be recovered. 

Eventually, `Exit` and `exit` should be removed.  

The IApp interface should be given a regular execution path to handle exiting the event loop, which properly recovers resources, closes files etc.  

Exceptions should be used as exceptions, handled if possible and allow to propagate otherwise.

LGM-Doyle:
@geoffthemedio as @dbenage-cx pointed out it resolves the first case, but not the second.

geoffthemedio:
Seems to work for me. Will merge. Can be extended and improved in future.

PR #1343, opened 2017-02-19
Change asteroid building requirements.

LGM-Doyle:
This PR makes two changes to the requirements for the asteroid buildings:
- requires that the asteroid shipyard be build in a system with an empire owned shipyard.
- allows the asteroid reformation processor to be build on any asteroid belt, without an asteroid shipyard.

It more clearly delineates the usage of the two buildings.

I never realized that the asteroid reformation processor, created armor parts that you could use anywhere it was supply connected.  I knew that it was an extension to the asteroid shipyard and assumed that it needed to be in the same system as the shipyard using the parts.

With this PR the asteroid shipyard is tied to a basic shipyard, like the other hull lines shipyard extensions.

The asteroid processor clearly creates a freespace resource that can be used anywhere.

### DIFF
--- a/default/scripting/buildings/NEUTRONIUM_FORGE.focs.txt
+++ b/default/scripting/buildings/NEUTRONIUM_FORGE.focs.txt
@@ -6,6 +6,7 @@ BuildingType
     tags = "ORBITAL"
     location = And [
         Planet
+        OwnedBy empire = Source.Owner
         Not Contains Building name = "BLD_NEUTRONIUM_FORGE"
         Contains And [
             Building name = "BLD_SHIPYARD_BASE"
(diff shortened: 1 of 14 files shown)


Vezzra:
@LGM-Doyle, to avoid a design discussion here in the comment section of your PR, can you open a corresponding thread on the [Other Game Design forum](http://freeorion.org/forum/viewforum.php?f=6&sid=4d4564063d26592daacc19e00ff2e9b4) and add a link to it here?

In the opening post of the thread explain why you think this change should be done, so we can start the discussion there.

LGM-Doyle:
I'm closing this since I originally considered this a bug fix and not a content change.

Issue #949, opened 2016-09-04
Monster maturation should be gated for combat occuring


MatGB:
### Description

Currently, Krakens et al can mature on turns in which combat is taking place in their system, this then means they're not only at full health the next turn but also significantly more of a threat. Extreme example attached:
![screenshot_20160904_213859](https://cloud.githubusercontent.com/assets/11533206/18233790/6013ff30-72e8-11e6-9a60-29d658b7cace.png)
### Expected Result

Combat occurring should prevent maturation in the same way it prevents most repair effects. Potentially matured monsters should have the same proportion of health that they had before maturing, but this would require some thought and give them some sort of healing ability (and unrelated to this: could keep the same name?).

o01eg:
Was it fixed by #1351 ?

PR #1351, opened 2017-02-21
Prevent monster maturation during combat

dbenage-cx:
Addresses the bug portion of #949, preventing monsters from "maturing" during combat.


geoffthemedio:
looks reasonable

MatGB:
Agreed, and a real pain to test because it's spotting something not happening randomly, so unusually for me I vote just merge in.

PR #1352, opened 2017-02-21
increase AI restrictions on use of base trooper invasions

Dilvish-fo:
Helps address issues raised in #1012 (complementary to the related improvement already put in place by Morlic re destroyed on conquest) and #1013.  One of the restrictions added here is that base troopers would not be used if more than 20 of them would be needed.

### DIFF
--- a/default/python/AI/InvasionAI.py
+++ b/default/python/AI/InvasionAI.py
@@ -19,6 +19,9 @@ import CombatRatingsAI
 from freeorion_tools import tech_is_complete, AITimer
 from AIDependencies import INVALID_ID
 
+MAX_BASE_TROOPERS_GOOD_INVADERS = 20
+MAX_BASE_TROOPERS_POOR_INVADERS = 10
+
 invasion_timer = AITimer('get_invasion_fleets()', write_log=False)
 
 
@@ -67,6 +70,32 @@ def get_invasion_fleets():
             avail_pp = el.data()
             for pid in el.key():
                 available_pp[pid] = avail_pp
+        # For planning base trooper invasion targets we have a two-pass system.  (1) In the first pass we consider all
+        # the invasion targets and figure out which ones appear to be suitable for using base troopers against (i.e., we
+        # already have a populated planet in the same system that could build base troopers) and we have at least a
+        # minimal amount of PP available, and (2) in the second pass we go through the reserved base trooper target list
+        # and check to make sure that there does not appear to be too much military action still needed before the
+        # target is ready to be invaded, we double check that not too many base troopers would be needed, and if things
+        # look clear then we queue up the base troopers on the Production Queue and keep track of where we are building
+        # them, and how many; we may also disqualify and remove previously qualified targets (in case, for example,
+        # we lost our base trooper source planet since it was first added to list).
+        #
+        # For planning and tracking base troopers under construction, we use a dictionary store in
+        # foAI.foAIstate.qualifyingTroopBaseTargets, keyed by the invasion target planet ID.  We only store values
+        # for invasion targets that appear likely to be suitable for base trooper use, and store a 2-item list.
+        # TODO: either actually use the first value to help sort out potential mission collisions in
+        # assign_invasion_fleets_to_invade(), or else simplify to a 1-item (source PID) entry
+        # TODO: or consider using a small class with named fields instead of this 2-item list, for clarity
+        # The first item in this list is the number of base troopers to be build, and the second entry will store the
+        # location ID where they are being built, once they are actually queued.  Initially, however, we simply enter
+        # INVALID_ID (-1) as the location, to flag this target as being reserved as a base-trooper invasion target.
+        # In the second pass, if/when we actually start construction, then we modify the record to indicate the source
+        # of the base troopers.  The expectation is that this will help us be able to better manage multiple base troop
+        # invasions in the same system that could be at the same time or at least overlap in time.
+
+        secure_ai_fleet_missions = foAI.foAIstate.get_fleet_missions_with_any_mission_types([MissionType.SECURE])
+
+        # Pass 1: identify qualifying base troop invasion targets
         for pid in invadable_planet_ids:  # TODO: reorganize
             if pid in foAI.foAIstate.qualifyingTroopBaseTargets:
                 continue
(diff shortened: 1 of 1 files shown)


default/python/AI/InvasionAI.py
```diff
@@ -76,15 +78,25 @@ def get_invasion_fleets():
...
+                    best_base_trooper_here = \
+                        ProductionAI.get_best_ship_info(PriorityType.PRODUCTION_ORBITAL_INVASION, pid2)[1]
+                    troops_per_ship = best_base_trooper_here.troopCapacity
+                    species_troop_grade = CombatRatingsAI.get_species_troops_grade(planet2.speciesName)
+                    troops_per_ship = CombatRatingsAI.weight_attack_troops(troops_per_ship, species_troop_grade)
+                    if troops_per_ship > best_trooper_count:
+                        best_base_planet = pid2
+                        best_trooper_count = troops_per_ship
+            if best_base_planet != INVALID_ID:
+                foAI.foAIstate.qualifyingTroopBaseTargets.setdefault(pid, [best_trooper_count, INVALID_ID])
```
Morlic-fo: The way ```qualifyingTroopBaseTargets``` works is obscure and should be reworked at some point. It is nowhere documented what the two entries are and everytime I stumble across it I need to reread all the use cases to figure it out.

Anyway, this usage seems wrong. Before this PR it was:
-                    foAI.foAIstate.qualifyingTroopBaseTargets.setdefault(pid, [pid2, INVALID_ID])
So it seems you want to set it to ```best_base_planet``` not troop count.

default/python/AI/InvasionAI.py
```diff
@@ -76,15 +78,25 @@ def get_invasion_fleets():
             planet_partial_vis_turn = universe.getVisibilityTurnsMap(pid, empire_id).get(fo.visibility.partial, -9999)
             if planet_partial_vis_turn < sys_partial_vis_turn:
                 continue
+            best_base_planet = INVALID_ID
+            best_trooper_count = 1  # won't bother using base troopers unless better than 1 per ship
```
Morlic-fo: Not sure I like this.

Assume we have only 1 trooper per part. Still, if we want to invade a planet with, say, 6 troops, there is not really an argument to build a bigger ship (6 slots) somewhere else which might need to be moved here.

IMO, the ```if n_bases > 20:``` check should be sufficient to rule out inefficient cases.

default/python/AI/InvasionAI.py
```diff
@@ -126,6 +140,10 @@ def get_invasion_fleets():
                     'Best troop design %s can not be produces in at planet with id: %s\d' % (col_design, build_choices)
                 )
             n_bases = math.ceil((p_troops + 1) / troops_per_ship)  # TODO: reconsider this +1 safety factor
+            if n_bases > 20:  # fall back to standard troopers if would need more than 20 base troopers
```
Morlic-fo: Please introduce a module level constant for this value.

default/python/AI/InvasionAI.py
```diff
@@ -126,6 +140,10 @@ def get_invasion_fleets():
...
+                print "ruling out base invasion troopers at %s due to excessive number required." % planet.name
```
Morlic-fo: Use ```% planet``` instead of ```% planet.name``` - it gives both ID and name which is useful for debugging.

Example output: ```P_65<Kithrup Asteroids I>```

Morlic-fo (changes requested):

default/python/AI/InvasionAI.py
Dilvish-fo: Well, OK, pretty much.  Checking for at least 1 trooper per ship was a very short-cut check to decide if the base troopers should be deemed a better choice than regular ships.  If you are only getting 1 per ship then I think it can easily be better to go with regular troopers.  I added TODO about doing more analysis, and moved the floor down to anything above zero, and made the later check a more complex, so it accepts 10 or fewer ships no matter how many troopers per ship, or up to 20 ships so long as it is more than 1 trooper per ship.

default/python/AI/InvasionAI.py
```diff
@@ -76,15 +78,25 @@ def get_invasion_fleets():
...
             for pid2 in state.get_empire_inhabited_planets_by_system().get(sys_id, []):
                 if available_pp.get(pid2, 0) < 2:  # TODO: improve troop base PP sufficiency determination
                     break
                 planet2 = universe.getPlanet(pid2)
                 if not planet2:
                     continue
-                if pid not in foAI.foAIstate.qualifyingTroopBaseTargets and planet2.speciesName in ColonisationAI.empire_ship_builders:
-                    foAI.foAIstate.qualifyingTroopBaseTargets.setdefault(pid, [pid2, INVALID_ID])
-                    break
+                if planet2.speciesName in ColonisationAI.empire_ship_builders:
```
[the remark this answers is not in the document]
Dilvish-fo: OK, done

default/python/AI/InvasionAI.py
```diff
@@ -76,15 +78,25 @@ def get_invasion_fleets():
...
+                    best_base_trooper_here = \
+                        ProductionAI.get_best_ship_info(PriorityType.PRODUCTION_ORBITAL_INVASION, pid2)[1]
+                    troops_per_ship = best_base_trooper_here.troopCapacity
```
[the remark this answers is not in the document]
Dilvish-fo: OK, done

default/python/AI/InvasionAI.py
Dilvish-fo: You're right that it is/was obscure, and unnecessarily complex.  I've added comments explaining it and proposing potential simplifications.  My changing thoughts over time as to which values to use and how, account for the discrepancy you note, but I still don't really make use of the two values the way I envision eventually doing, and until such time it should perhaps just drop to one value.  But rather than do that right away I wanted to make sure I got a related bugfix in right away for anyone testing this and mull over the rest a bit more.

default/python/AI/InvasionAI.py
Dilvish-fo: OK, done.

default/python/AI/InvasionAI.py
Dilvish-fo: OK

Dilvish-fo:
I haven't yet actually gotten the AI to USE base troopers with this revised version (though I've tried), so perhaps the requirements are a bit too strict right now.  I'll keep working on that, but I'd rather err on the restrictive side than having it go back to spamming a ton of them.

default/python/AI/InvasionAI.py
```diff
@@ -67,7 +70,35 @@ def get_invasion_fleets():
...
+        # look clear then we queue up the base troopers on the Production Queue and keep track of where we are building
+        # them, and how many; we may also disqualify and remove previously qualified targets (in case, for example,
+        # we lost our base trooper source planet since it was first added to list).
+        #
+        # For planning and tracking base troopers under construction, we use a dictionary store in
+        # foAI.foAIstate.qualifyingTroopBaseTargets, keyed by the invasion target planet ID.  We only store values
+        # for invasion targets that appear likely to be suitable for base trooper use, and store a 2-item list.
+        # TODO: either actually use the first value to help sort out potential mission collisions in
+        # assign_invasion_fleets_to_invade(), or else simplify to a 1-item (source PID) entry
+        # TODO: or consider using a small class with named fields instead of this 2-item list, for clarity
```
Morlic-fo: Speaking longterm, we should hide any such implementation details and use human-readable interface functions instead, e.g. ```register_basetrooper_target(target, troop_count)```. Beyond the scope of this PR of course.

default/python/AI/InvasionAI.py
```diff
@@ -76,35 +107,61 @@ def get_invasion_fleets():
...
+                del foAI.foAIstate.qualifyingTroopBaseTargets[pid]
+                continue
             if (planet.currentMeterValue(fo.meterType.shield) > 0 and
                     this_sys_status.get('myFleetRating', 0) < 0.8 * this_sys_status.get('totalThreat', 0)):
+                # this system not secured, so ruling out invasion base troops for now
+                # don't immediately delete from qualifyingTroopBaseTargets or it will be opened up for regular troops
+                #del foAI.foAIstate.qualifyingTroopBaseTargets[pid]
                 continue
-            loc = foAI.foAIstate.qualifyingTroopBaseTargets[pid][0]
+            loc = foAI.foAIstate.qualifyingTroopBaseTargets[pid][1]
```
Morlic-fo: Here is the problem I see.

At this point, the second entry should always be INVALID_ID as it only changed to a real planet_id when the troopers are enqueued. However, that requires us knowing which planet we should build the bases on. But we always find INVALID_ID and thus will never know the location.

The previous implementation stored the source planet in the first entry and could use that one. So, considering that you do not use the troop count at the moment, easiest to just revert your changes to this structure...

Morlic-fo (changes requested):
> I haven't yet actually gotten the AI to USE base troopers with this revised version (though I've tried)

That is probably caused by a simple but serious logic error you introduced with your changes to the ```qualifyingTroopBaseTargets```...

default/python/AI/InvasionAI.py
Dilvish-fo: Yes, that is the problem; I've put in a fix and updated the comments.  I seem to really be muddling through this one, oof!

Dilvish-fo:
Yes, that is the problem; I've put in a fix and updated the comments. I seem to really be muddling through this one, oof!

default/python/AI/InvasionAI.py
```diff
@@ -128,7 +127,7 @@ def get_invasion_fleets():
                     best_base_planet = pid2
                     best_trooper_count = troops_per_ship
             if best_base_planet != INVALID_ID:
-                foAI.foAIstate.qualifyingTroopBaseTargets.setdefault(pid, [best_trooper_count, INVALID_ID])
+                foAI.foAIstate.qualifyingTroopBaseTargets.setdefault(pid, [pid2, INVALID_ID])
```
Morlic-fo: best_base_planet not pid2?

Dilvish-fo:
*sigh*, right, I can hardly believe how many mistakes I am ringing up here, thanks for checking carefully.

Morlic-fo (approved):

Morlic-fo:
Seems good for merge now.

PR #1354, opened 2017-02-21
Production Window pedia article

dbenage-cx:
Adds an article for the production window. (Guides->Interface)

Supercedes #1065

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -3478,6 +3478,116 @@ Double-click on an available item to add it to the production queue.
 PRODUCTION_WND_TITLE
 Production
 
+PRODUCTION_WINDOW_ARTICLE_TITLE
+Production Window
+
+PRODUCTION_WINDOW_ARTICLE_TEXT
+'''The Production Window is the primary interface for utilizing Production Points (PP).
+It may be accessed by:
+*  Pressing the hotkey (default: CTRL + P)
+*  Double clicking on a planet
+*  Clicking on either of the industry icons at top of the [[MAP_WINDOW_ARTICLE_TITLE]] (shown here) <img src="icons/buttons/production.png"></img>
+
+The Production Window is composed of:
+*  Production Summary panel
+*  Producible Items panel
+*  Production Queue panel
+*  The system side panel (on the right edge)
+It also allows for some interaction with the [[encyclopedia MAP_WINDOW_ARTICLE_TITLE]], primarily to change the system selection.
+
+
+<u>Production Summary panel</u>
+
+(default position: top left)
+The Production Summary panel shows available Production Points.
+Total Available Points relates to the PP available across your entire empire.
+Points Available Here are the maximum available for the selected system.  This may be less than the total available when all of your empire's systems are not [[metertype METER_SUPPLY]] connected.
+Wasted Points will be lost, if not allocated before ending the turn.
+
+
+<u>Producible Items panel</u>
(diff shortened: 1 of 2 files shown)


dbenage-cx:
Not thrilled about either example format.
The first is too verbose to follow, the second may not be verbose enough.

default/stringtables/en.txt
```diff
@@ -3478,6 +3478,111 @@ Double-click on an available item to add it to the production queue.
...
+(default position: bottom edge)
+The Producible Items panel contains items available to start a Production Order.
+Toggle the display of item types or availability by clicking on the column headers.
+Items may be added to the Production Queue as a Production Order by double clicking on them or with the right click menu.
+Each item type has a column for total cost and minimum time.
+The maximum PP that may be allotted to any Production Order is (cost / time) x quantity.
+e.g. an item type with a total cost of 30 PP and a minimum time of 3 turns can progress by up to 10 PP per item, per turn.
+Unavailable items, if displayed, will detail the reasons they are unavailable at the bottom of their tooltip in red (hover the mouse over the item).
+
+<u>Production Queue panel</u>
```
geoffthemedio, line 3519: add extra newline

dbenage-cx: Nice change, fixed.

dbenage-cx:
Rebased, lacking criticism on example I've reduced to single commit.

default/stringtables/en.txt
```diff
@@ -3478,6 +3478,116 @@ Double-click on an available item to add it to the production queue.
...
+-Turn  3: Order #1 20 / 16 -> 10
+
+-Turn  4: Order #1 30 /   6 ->   6  an additional 4 PP will go to the next repetition of this order.
+
+-Turn  5: A Survey Ship is produced (order #1)
+                Order #1   4 / 32 -> 10  now displays 1 repetition remains
+
+-Turn  6: Order #1 14 / 22 -> 10
+
+-Turn  7: Order #1 32 /   4 ->   4
```
Ouaz, line 3565: I come a bit late, but why the completed PP value raise from 14 to 32 in one turn? Although the projected PP value is only 10?

It seems you added 10 to the remaining PP of Turn 6, instead of adding them to the completed PP.

default/stringtables/en.txt
```diff
@@ -3478,6 +3478,116 @@ Double-click on an available item to add it to the production queue.
...
+                Order #2   0 / 36 ->   6
+
+-Turn  8: A Survey Ship is produced (order #1)
+                Empire production increases to 22 PP per turn (for some unrelated reason).
+                Order #2   4 / 32 -> 12  the max per turn for this project type
```
Ouaz, line 3570: Shouldn't be  `Order #2   6 / 30 -> 12` as there was 6 projected PP in Turn 7?

default/stringtables/en.txt
```diff
@@ -3478,6 +3478,116 @@ Double-click on an available item to add it to the production queue.
...
+                Order #3   0 / 36 -> 10
```
Ouaz, line 3571: Shouldn't be `Order #3   0 / 72 -> 10`, as `Order #3` is x2 ?

default/stringtables/en.txt
dbenage-cx: Thanks for catching these, apparently I forgot how to add ;/  Addressed in PR #1367

Issue #1361, opened 2017-02-23
FO sounds continues playing even when game lost focus


adrianbroher:
### Environment
* **FreeOrion Version**: master
* **Operating System**: any
* **Graphic card used**: -
* **Fetched as**
  * Compiled from source

### Description
The current implementation of the sound system continues playing the background music, even when the game application lost its focus.  

### Expected Result
When selecting another application FreeOrion should suspend its audio output until it regains the application focus.

### Steps to reproduce
* Start FreeOrion in windowed mode.
* Focus OS Desktop or another Application
* FreeOrion continues playing background music, despite not being in focus.

Voker57:
This should be disableable. New turn sounds are extremely useful, for instance, and I would like to hear them when freeorion is not focused.

TheSilentOne1:
The new turn sound is not disabled when the application loses focus, just the music.

PR #1362, opened 2017-02-23
disable music when app loses focus

TheSilentOne1:
as requested here:
https://github.com/freeorion/freeorion/issues/1361

### DIFF
--- a/UI/Sound.cpp
+++ b/UI/Sound.cpp
@@ -38,6 +38,12 @@ public:
         loops + 1 times otherwise. */
     void PlayMusic(const boost::filesystem::path& path, int loops = 0);
 
+    /** Pauses music play, to be continued from the same position */
+    void PauseMusic();
+
+    /** Resumes music play */
+    void ResumeMusic();
+
     /** Stops playing music. */
     void StopMusic();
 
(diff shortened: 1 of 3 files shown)


adrianbroher:
This isn't suspending the sound system.

TheSilentOne1:
Does it have to be? That shouldn't be complicated either, but all we want to disable the music, don't we?

geoffthemedio:
In at least some cases, I think people still want to have the turn update sound be audible, so that they know when turn processing finished with the game minimized.

adrianbroher:
>  That shouldn't be complicated either, but all we want to disable the music, don't we?

No, I don't think stopping and restarting is the right thing to do.  The music should continue playing at the point where it was paused.

TheSilentOne1:
@adrianbroher: I think you're right, I changed it.
Beware: I have no plan about (how) OpenAL (works). Testing suggests this works as intended, but no guarantees.

adrianbroher:
@TheSilentOne1 Looks good in my opinion.  And according to the API of OpenAL this should be right way to do this.  I will test the PR as soon as I have some time.  Thanks so far.

Vezzra:
The corresponding issue #1361 is assigned to the "post 0.4.7" milestone, however, with a solution already implemented that looks promising so far, this has a good chance to be ready for 0.4.7, in which case there is no reason to hold it back.

So this now looks more like "optional for 0.4.7", and both issue and PR could be set to this milestone. @adrianbroher, just wanted to check back with you, as you initially assigned the issue to "post 0.4.7" - are you ok with switching to "optional for 0.4.7"?

adrianbroher:
@Vezzra Sure, go ahead.  I didn't expect that someone would pick this up the mentioned issue this fast and any solution wouldn't be missed for v0.4.7.

geoffthemedio:
works for me

TheSilentOne1:
@adrianbroher: Did you want to have a look at this or shall I squash & merge?

adrianbroher:
@TheSilentOne1 go ahead.

PR #1364, opened 2017-02-24
Replace Cpp Api content tag filter with parser script

### LINKED ISSUE #1319: Cpp API tag filter requires doxygen v1.8.12
Travis is currently running doxygen v1.8.11 and not correctly parsing the arguments in `FILTER_PATTERNS` (doc/DoxyFile.in).

This was apparently fixed for doxygen v1.8.12 https://bugzilla.gnome.org/show_bug.cgi?id=504305

Locally running version 1.8.13, I do not have any errors for:
 `sh: 1: /usr/bin/python2.7 /src/doc/tag_filter.py: not found`

(milestone purposefully absent, does not affect build or typical user)

dbenage-cx:
Parses python and focs files and creates `all_tags.dox`

The description for content tags are allowed to span multiple lines as well as multiple instances for the same tag.

Resolves #1319


doc/CMakeLists.txt
```diff
@@ -13,11 +13,17 @@ if(DOXYGEN_FOUND)
         add_custom_target(doc)
     endif()
 
+    add_custom_target(cpp-apidoc-content-tags
```
adrianbroher: add_custom_target allows multiple COMMAND statements so there is no need for an explicit `cpp-apidoc-content-tags` target but instead you can add the execution statement as the first COMMAND to the cpp-apidoc target. 

https://cmake.org/cmake/help/v3.1/command/add_custom_target.html

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
...
+# Source and target paths for file generation
+doc_dir = os.path.join(source_dir, 'doc')
+dox_source = os.path.join(doc_dir, 'all_tags.dox.in')
+dox_target = os.path.join(doc_dir, 'all_tags.dox')
+
+# match condition in doc_target to replace with result of parsing
+dox_variable = "${CONTENT_DOCUMENTATION}"
+
+# prefix to source links (being the cpp api, internal links do not exist)
+base_link_url = 'https://github.com/freeorion/freeorion/blob/master/'
```
adrianbroher: This breaks local generations of the documentation.  Why don't you use internal linking?

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
+from __future__ import print_function
+import sys
+import os
+from collections import defaultdict
+import fnmatch
+
+# mask for files to parse
+file_masks = {'*.py', '*.focs.txt', '*.macros'}
```
adrianbroher: The source directories should be application parameters.

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
...
+
+# mask for files to exclude from parse, checked after file_masks
+exclude_masks = {'__*'}
```
adrianbroher: The exclude masks should be application parameters.

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
...
+# prefix to initiate a match condition
+tag_comment = "@content_tag{"
+
+# Root to search for files
+source_dir = sys.argv[1]
+
+# Source and target paths for file generation
+doc_dir = os.path.join(source_dir, 'doc')
+dox_source = os.path.join(doc_dir, 'all_tags.dox.in')
+dox_target = os.path.join(doc_dir, 'all_tags.dox')
```
adrianbroher: The output file should be an application parameter.  Otherwise you write into the source tree, which hinders the use of multiple build trees.

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
...
+
+# match condition in doc_target to replace with result of parsing
+dox_variable = "${CONTENT_DOCUMENTATION}"
+
+# prefix to source links (being the cpp api, internal links do not exist)
+base_link_url = 'https://github.com/freeorion/freeorion/blob/master/'
+
+
+class SourceInfo(object):
```
adrianbroher: A tuple is sufficient as a substitute.

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
...
+        :type _line_number: int
+        :type _description: str
+        """
+        assert isinstance(_file_name, str)
+        assert isinstance(_line_number, int)
+        self.file = SourceInfo(_file_name, _line_number)
+        assert isinstance(_description, str)
+        self.description = _description
+
+    def __repr__(self):
```
adrianbroher: Wrong use of `__repr__`.

https://docs.python.org/2/library/functions.html#repr

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
...
+        assert isinstance(_name, str)
+        self.name = _name
+        assert isinstance(_line, int)
+        self.line = _line
+
+    def __repr__(self):
+        return str.format("{0}#L{1}", self.name, self.line)
+
+
+class TagDescription(object):
```
adrianbroher: Again I think a tuple should be sufficient as substitute for this.

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def parse_file(_file_path, _file_name, _tags):
+    relative_name = os.path.join(_file_path, _file_name)
+    fqn = os.path.join(source_dir, relative_name)
+    with open(fqn, 'r') as f:
+        line_num = 0
+        match_line = 0
+        content = []
+        for parse_line in f:
```
adrianbroher: `enumerate` should make the intent a bit more clear:

https://docs.python.org/2/library/functions.html#enumerate

doc/tag_parser.py
```diff
@@ -0,0 +1,143 @@
...
+            line_num += 1
+            if parse_line.lstrip().startswith("#"):
```
adrianbroher: To be honest, I have no idea what is happening here (and the following lines) at first glance.  Do you collect multiline comments?

adrianbroher (changes requested):

doc/tag_parser.py
dbenage-cx: TY for review and helpful notes.

Yes, the variables are stored locally and moved/copied once an end condition is met (eof, non-comment line, new tag line).
If the next line is a comment line, it is assumed part of the tag description.  Mainly this is to allow pep8 compliance for line length and not hinder the description.  I will try to clarify this section as I work on other changes.

doc/tag_parser.py
dbenage-cx: The links for python/focs files would not exist on the `.io` site?  Will look into alternative.

adrianbroher: > The links for python/focs files would not exist on the .io site?

Well okay.  Now I get your idea.  The documentation nowhere points to the source code.  Why should it in this case?

doc/tag_parser.py
dbenage-cx: Fixed

doc/tag_parser.py
dbenage-cx: Removed

doc/tag_parser.py
dbenage-cx: Fixed

doc/tag_parser.py
dbenage-cx: Removed

doc/tag_parser.py
dbenage-cx: I've left in a link to github source, supporting the target being changed or link completely removed if either are requested.

I think some context can be helpful when looking at tags, especially if the documentation is too concise or there are multiple sources of documentation for a single tag.
With FOCS definitions there is no context to follow without pulling in the freeorion parser.
Python could arguably use the containing function/module, though I don't see any benefit over a name/line#.

doc/tag_parser.py
dbenage-cx: Fixed, moved target to build dir

doc/tag_parser.py
dbenage-cx: Fixed

doc/tag_parser.py
dbenage-cx: Fixed

doc/CMakeLists.txt
dbenage-cx: Fixed, though requires `&&` to not append doxygen command as an argument to script.
Not certain if you prefer these arguments as specific cmake variables, so I simply appended for now.

adrianbroher (approved):

adrianbroher:
Looks good to me.

dbenage-cx:
TY for review and assistance.

Vezzra:
@adrianbroher, same question as in #1319: why assigning this to a milestone? Has no relevance/impact on a release...

Issue #1336, opened 2017-02-18
Crash on changing something in the lobby


Voker57:
Happens quite often with different players when somebody changes something

Backtrace: 

```
#0  0x00007f254a1f34d1 in ModalListPicker::CurrentItem (this=0x55a56143cd80) at /home/voker57/Sources/freeorion/GG/src/DropDownList.cpp:233
#1  0x00007f254a1f30bb in ModalListPicker::Run (this=0x55a56143cd80) at /home/voker57/Sources/freeorion/GG/src/DropDownList.cpp:183
#2  0x00007f254a1f66af in GG::DropDownList::LClick (this=0x55a5612cf6b0, pt=..., mod_keys=...)
    at /home/voker57/Sources/freeorion/GG/src/DropDownList.cpp:811
#3  0x000055a55c57fb61 in CUIDropDownList::LClick (this=0x55a5612cf6b0, pt=..., mod_keys=...)
    at /home/voker57/Sources/freeorion/UI/CUIControls.cpp:840
#4  0x00007f254a39dae3 in GG::Wnd::HandleEvent (this=0x55a5612cf6b0, event=...) at /home/voker57/Sources/freeorion/GG/src/Wnd.cpp:1070
#5  0x00007f254a296e00 in GG::GUIImpl::HandleMouseButtonRelease (this=0x55a55e66a300, mouse_button=0, pos=..., curr_ticks=42469)
    at /home/voker57/Sources/freeorion/GG/src/GUI.cpp:523
#6  0x00007f254a29a269 in GG::GUI::HandleGGEvent (this=0x7ffee1801328, event=GG::GUI::LRELEASE, key=GG::GGK_UNKNOWN, key_code_point=0, 
    mod_keys=..., pos=..., rel=..., text=0x0) at /home/voker57/Sources/freeorion/GG/src/GUI.cpp:1145
#7  0x00007f254c6a1084 in GG::SDLGUI::HandleSystemEvents (this=0x7ffee1801328) at /home/voker57/Sources/freeorion/GG/src/SDL/SDLGUI.cpp:801
#8  0x000055a55c3a6064 in HumanClientApp::HandleSystemEvents (this=0x7ffee1800860)
    at /home/voker57/Sources/freeorion/client/human/HumanClientApp.cpp:819
#9  0x00007f254a227eda in GG::ModalEventPump::operator() (this=0x7ffee18007c0) at /home/voker57/Sources/freeorion/GG/src/EventPump.cpp:113
[...]

PR #1366, opened 2017-02-25
Fix crashes in ModalListPicker in the multiplayer lobby.

LGM-Doyle:
If the parent of a ModalListPicker that is dropped down and in a modal
event loop is destroyed, then signaling or returning from the Run() and
accessing any members of the parent results in a segfault.

It is not possible to throw an exception and unroll the modal event pump
cleanly.

This commit uses a pair of shared_ptrs to the ModalListPicker instance
to determine if the parent has been destroyed before signaling or
returning from Run().

This PR works for this specific case #1336.

It does not fix the underlying problem that the modal event pump is not thread safe.  There may be other paths lurking to expose this flaw.

### DIFF
--- a/GG/GG/DropDownList.h
+++ b/GG/GG/DropDownList.h
@@ -237,7 +237,7 @@ protected:
 private:
     const ListBox*  LB() const;
 
-    std::unique_ptr<ModalListPicker> const m_modal_picker;
+    const std::shared_ptr<ModalListPicker> m_modal_picker;
 };
 
 } // namespace GG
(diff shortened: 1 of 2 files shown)


Vezzra:
@LGM-Doyle, I wonder if we should open an issue about the modal event pump not being thread safe, so we don't forget about that issue...?

LGM-Doyle:
I would not open an issue.

The issue would not be a focused actionable item.  There would be no way to determine, when the goal had been achieved. 

The modal event pump is also the event pump that the freeorion GUI runs in.  The number of ways that it is not thread safe, not interruptible etc are difficult to enumerate.  When it feels ignore, I imagine it will speak up for itself, particularly as we have more multiplayer fans at the moment.

Vezzra:
Ok... although that it's apparently so hard to make something thread safe which apparently *needs* to be thread safe sounds a bit instable...

PR #1367, opened 2017-02-26
Fixup example section of Production Window article

dbenage-cx:
Corrects the example given in Production Window pedia article.


geoffthemedio:
@Ouaz comments?

Ouaz:
Perfect.

He even added `(22 - (4 + 12)` on Turn 11, which makes easier to understand the sentence `If the order will complete, some PP are allowed to start the next iteration, up to the maximum allowed per turn for that item type.` 

Maybe add it is due to the position priority of the order in the queue?

PR #1368, opened 2017-02-27
Add stealth and detection to supply resolution.

LGM-Doyle:
This PR goes with the [forum](http://freeorion.org/forum/viewtopic.php?f=6&t=10391&p=87613#p87613) post, showing images of the progress towards adding stealth and detection to the supply resolution.  

The PR is not complete, but given the difficult merge with the MapWnd.cpp changes yesterday I would like to coordinate, if possible with the changes to diplomacy and supply for allied supply.

If we decide to use some/all of this PR then I will break it up like other large PR I've proposed in the past.

This PR does the following
- Changes the supply resolution to include the supply source's stealth and the competing empire's detection strength in the calculations.  Simply, if an empire can't detect a competing empire's supply then it can't blockade nor can its supply block that supply source.  Otherwise, it works as before.
- Changes the map display of supply to only show one empire at a time.  This avoids trying to visualize overlapping supply networks.  The displayed supply network is selected by the player clicking on a planet, fleet, ship in the MapWnd or the empire name in the Empire window.
- Adds a system supply icon in the SidePanel that displays the details of the supply calculation.  It shows range, distance, stealth, the total bonus and the individual components of the bonus.  
- It extends the python wrappers to pass unordered_maps and unordered_sets.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -2005,29 +2005,6 @@ namespace
     }
 }
 
-void Empire::UpdateSystemToStealthAndSupplyRange() {
-
-    m_system_to_stealth_supply.clear();
-
-    // as of this writing, only planets can generate supply propagation
-    for (int object_id : ExistingObjectsKnownToEmpire<Planet>(EmpireID())) {
-        std::shared_ptr<const Planet> planet = GetPlanet(object_id);
-        std::shared_ptr<const UniverseObject> obj = planet;
-        // Check is it an owned planet with a valid id and a supply meter.
-        if (!planet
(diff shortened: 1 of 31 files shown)


adrianbroher:
> The PR is not complete, but given the difficult merge with the MapWnd.cpp changes yesterday I would like to coordinate, if possible with the changes to diplomacy and supply for allied supply.

> difficult merge
> 30 files
> +3,061 −800 delta

You only have to blame yourself for creating such massive PRs.

PS.: The build is broken on every platform.

LGM-Doyle:
Thanks for the heads up.

PR #1370, opened 2017-02-28
Refactor allocation part of MilitaryAI.get_military_fleets

Morlic-fo:
The new approach is object oriented and will provide a base for upcoming changes (in particular, an update to the invasion logic).

For now, the allocation code is supposed to be exactly replicated. This
is checked by running the new code in parallel to the old one and
comparing the outputs. If some discrepancies are detected, there will be some error message in the chat (but the AI will not be affected). My tests in few games indicate that the outputs are the same but as usual rare edge cases may have not occured.

If the changes in this PR are approved, I will remove the old code  before merge.

### DIFF
--- a/default/python/AI/MilitaryAI.py
+++ b/default/python/AI/MilitaryAI.py
@@ -638,34 +638,19 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
         for sys_id in other_targeted_system_ids:
             OutpostTargetAllocator(sys_id, allocation_helper).allocate()
 
-        # TODO this block is effectively not used as AIstate.opponentSystemIDs is never filled...
-        # blockade enemy systems
-        other_targeted_system_ids = []
-        targetable_ids = set(
-            ColonisationAI.systems_by_supply_tier.get(0, []) + ColonisationAI.systems_by_supply_tier.get(1, []))
-        for sys_id in AIstate.opponentSystemIDs:
-            if sys_id in targetable_ids:
-                other_targeted_system_ids.append(sys_id)
-            else:
-                for nID in universe.getImmediateNeighbors(sys_id, empire_id):
-                    if nID in targetable_ids:
-                        other_targeted_system_ids.append(sys_id)
-                        break
-        for sys_id in other_targeted_system_ids:
-            BlockadeAllocator(sys_id, allocation_helper).allocate()
+        # TODO blockade enemy systems
 
         # interior systems
-        current_mil_systems = [sid for sid, alloc, take_any, _ in allocation_helper.allocations]
+        targetable_ids = set(ColonisationAI.systems_by_supply_tier.get(0, []) +
+                             ColonisationAI.systems_by_supply_tier.get(1, []))
+        current_mil_systems = [sid for sid, _, _, _ in allocation_helper.allocations]
         interior_targets1 = targetable_ids.difference(current_mil_systems)
         interior_targets = [sid for sid in interior_targets1 if (
             allocation_helper.threat_bias + systems_status.get(sid, {}).get('totalThreat', 0) > 0.8 * allocation_helper.already_assigned_rating[sid])]
         for sys_id in interior_targets:
             InteriorTargetsAllocator(sys_id, allocation_helper).allocate()
 
-        # exploration targets
-        explo_target_ids = []
-        for sys_id in explo_target_ids:
-            ExplorationTargetAllocator(sys_id, allocation_helper).allocate()
+        # TODO Exploration targets
 
         # border protections
         visible_system_ids = foAI.foAIstate.visInteriorSystemIDs.keys() + foAI.foAIstate.visBorderSystemIDs.keys()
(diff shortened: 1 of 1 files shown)


default/python/AI/MilitaryAI.py
```diff
@@ -606,38 +1092,18 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
...
-                new_alloc += this_alloc
-                allocations.append((sid, this_alloc, False, 5))
-                remaining_mil_rating -= this_alloc
-                ot_sys_alloc += this_alloc
-            if "Main" in thisround or this_alloc > 0:
-                if _verbose_mil_reporting:
-                    print "Monster Den %4d ( %10s ) has local threat %8d ; existing military allocation %d and new allocation %8d" % (sid, universe.getSystem(sid).name, thrt, cur_alloc, this_alloc)
-        if "Main" in thisround or new_alloc > 0:
-            if _verbose_mil_reporting:
-                print "-----------------"
```
Morlic-fo: This code block is never used as ```monster_dens``` is empty. As it serves no real purpose, this part was removed to avoid unnecessary reimplementation.

default/python/AI/MilitaryAI.py
```diff
@@ -209,17 +674,19 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
...
         print "----------------------------"
-    # allocation format: ( sysID, newAllocation, takeAny, maxMultiplier )
-    # ================================
-    # --------Capital Threat ----------
+    # <editor-fold description="Capital Defense">
     if capital_sys_id is not None:
         capital_sys_status = systems_status[capital_sys_id]
         capital_threat = safety_factor*(2 * threat_bias + combine_ratings_list([capital_sys_status[thrt_key] for thrt_key in ['totalThreat', 'neighborThreat']]))
-        capital_threat += max(0, enemy_sup_factor[sys_id]*enemy_rating - capital_sys_status.get('my_neighbor_rating', 0))
+        capital_threat += max(0, enemy_sup_factor[capital_sys_id]*enemy_rating - capital_sys_status.get('my_neighbor_rating', 0))
```
Morlic-fo: existing bug in the old code.

default/python/AI/MilitaryAI.py
```diff
@@ -107,6 +569,7 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
 
     if try_reset and (fo.currentTurn() + empire_id) % 30 == 0 and thisround == "Main":
         try_again(all_military_fleet_ids, try_reset=False, thisround=thisround + " Reset")
+        return
```
Morlic-fo: Existing flaw in the old code.

Dilvish-fo:
I just wanted to note that I was too busy the past week or so to review and test this yet, but expect to do so over the next few days.

default/python/AI/MilitaryAI.py
```diff
@@ -94,6 +94,442 @@ def avail_mil_needing_repair(mil_fleet_ids, split_ships=False, on_mission=False,
...
+
+class OutpostTargetAllocator(TargetAllocator):
+    _max_alloc_factor = 3
+
+
+class BlockadeAllocator(TargetAllocator):
+    _potential_threat_factor = 0.25
+
+    def _maximum_allocation(self, threat):
+        return min(self._minimum_allocation(threat), self._allocation_helper.remaining_rating) * 1.5
```
Dilvish-fo: named factor?

default/python/AI/MilitaryAI.py
```diff
@@ -94,6 +94,442 @@ def avail_mil_needing_repair(mil_fleet_ids, split_ships=False, on_mission=False,
...
+
+class InteriorTargetsAllocator(LocalThreatAllocator):
+    _max_alloc_factor = 3
+    _min_alloc_factor = 1.5
+
+    def _calculate_threat(self):
+        return self.threat_bias + self.safety_factor * self._local_threat()
+
+    def _maximum_allocation(self, threat):
+        return 3 * min(self._minimum_allocation(threat), self._allocation_helper.remaining_rating)
```
Dilvish-fo: 3 --> _max_alloc_factor

default/python/AI/MilitaryAI.py
```diff
@@ -94,6 +94,442 @@ def avail_mil_needing_repair(mil_fleet_ids, split_ships=False, on_mission=False,
...
+class BorderSecurityAllocator(LocalThreatAllocator):
+    _min_alloc_factor = 1.2
+    _max_alloc_factor = 2
+    _allocation_group = 'accessibleTargets'
+
+    def __init__(self, sys_id, allocation_helper):
+        super(BorderSecurityAllocator, self).__init__(sys_id, allocation_helper)
+
+    def _maximum_allocation(self, threat):
+        return self.safety_factor * 2 * max(self._local_threat(), self._neighbor_threat())
```
Dilvish-fo: 2 -> _max_alloc_factor

Dilvish-fo (approved):
I noted a few places where you seem to have overlooked using the named factors you had already defined, and some others where perhaps (in keeping with the rest of the approach) you should add a named factor instead of using a constant.

Other than that it all looks fine, and it seems to work fine in my testing.

PR #1373, opened 2017-03-01
On Disconnect only sleep if disconnect attempted.

LGM-Doyle:
This tweaks the disconnect wait time, so that it is only 10ms and only occurs if a disconnect event happens.

### DIFF
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -91,8 +91,7 @@ AIClientApp::AIClientApp(const std::vector<std::string>& args) :
 }
 
 AIClientApp::~AIClientApp() {
-    if (Networking().Connected())
-        Networking().DisconnectFromServer();
+    Networking().DisconnectFromServer();
 
     DebugLogger() << "Shut down " + PlayerName() + " ai client.";
 }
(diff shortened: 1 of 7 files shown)


geoffthemedio:
is there a nontrivial advantage to waiting 0.01 instead of 1.0 seconds?

LGM-Doyle:
@geoffthemedio I am marking #1373, #1374 and #1375 as works in progress.  While I was pulling together the information for this answer I learned a few things that can be improved.

LGM-Doyle:
@geoffthemedio I think that you discovered and reported in #1375 why this is a non-trivial change.  The previous code always waited one second and then killed the server.  The various one, two and three second delays scattered around the code prevent the shutdown from happening within this one second window, or even 10 seconds.  So without addressing these directly the shutdown improvements won't work.

To correctly (I hope) address linger time on various platforms, I've split `Connected()` into two functions for tx and rx.  Now close can correctly, immediatly shutdown tx and shutdown rx when the other end finishes acknowledging the shutdown.

I've added a shared pointer to the handler in the tcp thread to anchor it in that there so it will not be destroyed before it finishes its work.

I've improved the closed connection messages so that they sound less like errors and explain why the connection was shutdown.

I've added clarifying comments about the linger option and keep alive.

geoffthemedio:
builds and seems to run OK

LGM-Doyle:
Included in #1375, which was merged.

Issue #1371, opened 2017-03-01
Time-out on connecting to server for LoadSingleGame (OSX)


memem359:
For the recent Mac OSX test build, I am unable to load saved games, as it times-out while trying to connect to the server. Starting a new game (with or without Quickstart) is fine. (This is also true for earlier test builds in February, but I don't know when it started.)

After some investigating, I believe this behavior is due to client/human/HumanClientApp.cpp
Specifically these 3 lines:
- Line 86: global variable for amount of time before a time out connecting to server
`    const unsigned int  SERVER_CONNECT_TIMEOUT = 4500; // in ms`
- Line 448: time between trying to connect to server for "New Game"
`    while (!m_networking.ConnectToLocalHostServer(boost::posix_time::seconds(2))) {`
- Line 660: default (no argument) time for "Load Game" is 5 seconds
`    while (!m_networking.ConnectToLocalHostServer()) {`

On my computer, it takes 1.8 seconds for the server to initialize and be ready for network connections.
- For New Game (normal), the human client waits 4 seconds after the server start, and then connects on 1st try.
- For New Game (quickstart), the human client tries to connect immediately after the server starts (but before it is ready). It fails to connect on the 1st try, but 2 seconds later connects on the 2nd try.
- For Load Game, the human client tries to connect immediately. It fails on the 1st try, and then times out before trying again.

## My Suggestion to the appropriate developer...
Change line 660 to be like line 448: add the 2 second argument to the ConnectToLocalHostServer call. This will let the LoadSinglePlayerGame try two more times to connect before the SERVER_CONECT_TIMEOUT, and it should succeed on the second try, which is after the 1.8 seconds for the server to initialize.

Vezzra:
Fixed by #1375.

calebjones:
Error when time-out error appears:

```
2018-10-06 15:32:08.883 FreeOrion[72716:113884306] IMKClient Stall detected, *please Report* your user scenario attaching a spindump (or sysdiagnose) that captures the problem - (imkxpc_presentFunctionRowItemTextInputViewWithEndpoint:reply:) block performed very slowly (44.91 secs).
[...]

PR #1374, opened 2017-03-01
Improve server polling timeout consistency

LGM-Doyle:
This PR are some tweaks to the network polling, that I meant to push earlier.

The PR switches some clocks from posix clock to boost::chrono, so that all clocks in freeorion are of one type.

The PR makes the network polling more reasonable, so that the polling interval is 10ms, about one refresh interval.

It might address bug #1371, but I have not tested it.

### DIFF
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -156,7 +156,7 @@ void AIClientApp::Run() {
 
 void AIClientApp::ConnectToServer() {
     typedef boost::chrono::steady_clock Clock;
-    const boost::chrono::milliseconds SERVER_CONNECT_TIMEOUT(4500);
+    const boost::chrono::milliseconds SERVER_CONNECT_TIMEOUT(10000);
     const boost::chrono::milliseconds SERVER_STARTUP_POLLING_TIME(10);
 
     bool connected = false;
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -161,17 +161,19 @@ void AIClientApp::ConnectToServer() {
 
     bool connected = false;
     Clock::time_point start_time = Clock::now();
-    boost::chrono::milliseconds connection_time(0);
+    Clock::duration connection_time{0};
     while (!connected && (connection_time < SERVER_CONNECT_TIMEOUT )) {
         connected = Networking().ConnectToLocalHostServer(SERVER_STARTUP_POLLING_TIME);
-        connection_time = boost::chrono::duration_cast<boost::chrono::milliseconds>(Clock::now() - start_time);
+        connection_time = Clock::now() - start_time;
     }
 
     if (!connected) {
-        ErrorLogger() << "AIClientApp::Run : Failed to connect to localhost server after " << connection_time << ".  Exiting.";
+        ErrorLogger() << "Timed out.  Failed to connect to server in "
+                      << boost::chrono::duration_cast<boost::chrono::milliseconds>(connection_time) << ".";
         Exit(1);
     }
-    DebugLogger() << "AI connected to local server took " << connection_time << ".";
+    DebugLogger() << "Connecting to server took "
+                  << boost::chrono::duration_cast<boost::chrono::milliseconds>(connection_time) << ".";
 }
 
 void AIClientApp::StartPythonAI() {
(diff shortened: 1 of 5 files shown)


geoffthemedio:
seems to work fine for me

Vezzra:
On OSX I can't start a game with this PR, connecting to the server times out.

Log files: [logfiles.zip](https://github.com/freeorion/freeorion/files/817044/logfiles.zip)

LGM-Doyle:
@Vezzra, thanks for the log files.  I'm working on a fix.

LGM-Doyle:
This PR should be reviewed before #1373 and #1375.

I have remove the polling code, which should not have been there since the `boost::asio` code already polls internally.  That was a misunderstanding in the current `HumanClientApp`.

I've also converted the asio interface to `std::chrono`, since the `boost::asio` code now defaults to `std::chrono` and this avoid some long redefinitions of timers.

@Vezzra I hope it now works on OSX again.

adrianbroher:
> since the boost::asio code now defaults to std::chrono

With what definition of `now`?

LGM-Doyle:
> With what definition of now?

Boost 1.49 if I did not misread their revision history.

Vezzra:
@LGM-Doyle, nope, unfortunately I'm still unable to start a game on OSX with that PR (commit 47c20e25c55842574359c1b14538d55650a7f9bb). Still getting the same timeout error when the client tries to connect to the server.

Logfiles: [logfiles.zip](https://github.com/freeorion/freeorion/files/827207/logfiles.zip)

LGM-Doyle:
@Vezzra I still don't know what the issue is.  

I've added more debugging message to this PR.  Could you try it and provide logs again, including any AI logs, if they exist?

The problem is not be directly related to this PR, since issue #1371 pre-dates this PR.

I think something changed with OSX in the relevant time period.

I've added a progressively backed off delay to the connection retry attempts in case the original round trip time through the SDL timer was effectively creating a delay in the new game connections, but a shorter delay in the load game connections.  If this works then we can keep it.

Otherwise, could you also check your system logs and firewall logs for any messages coinciding with starting up freeorion with this PR.

Thanks.

Vezzra:
>  Could you try it and provide logs again, including any AI logs, if they exist?

As expected no AI logs, but I think `freeorion.log` might contain some interesting info. Apparently all connection attempts except the very first fail because of an "invalid argument" error. The first one apparently simply fails because the server isn't ready yet.

Log files: [logfiles.zip](https://github.com/freeorion/freeorion/files/828548/logfiles.zip)

> The problem is not be directly related to this PR, since issue #1371 pre-dates this PR.
>
> I think something changed with OSX in the relevant time period.

Well, all I can say that I can at least start games with builds of master. So something you did in this PR changed things in a way that you can't even start a game on OSX now.

Obviously something doesn't work on OSX the same way it does on Windows and Linux...

> Otherwise, could you also check your system logs and firewall logs for any messages coinciding with starting up freeorion with this PR.

Can you check the FO logs I provided first, if anything in there gives you a relevant clue? Before I have to start hacking and digging through the jungle... 😨  😉 

For what it's worth, the issue turns up even when I switch off the OS firewall, so I guess the firewall isn't the culprit here.

LGM-Doyle:
@Vezzra thanks for the logs.  I'd like to try again.

I've refactored the connect attempts so that they close the socket before every attempt to connect.

If this works I'll remove the noisy debugging and rebase #1375 onto it.

Based on the comments in #1371, there is a difference between the quick and normal start game and the load game in terms of connecting on OSX.  I think that once we figure out what that difference is then both this PR  and that issue will be fixed.

Vezzra:
@LGM-Doyle, success. I'm now able to start new and load saved games with this PR. 👍

LGM-Doyle:
Thanks for testing.

Vezzra:

geoffthemedio:
if it runs, it's probly OK

LGM-Doyle:
#1375 contains all of the changes in #1374.  If #1375 is about to be merged, it will save one rebase to just wait a day or two.

Vezzra:
Ok, that's fine. 👍

Vezzra:
@LGM-Doyle, #1375 has been merged, which means we can close this one, right? I suggest to tag it as "follow up" and provide a reference to #1375 as the superseding PR.

LGM-Doyle:
Included in #1375, which was merged.

PR #1380, opened 2017-03-05
Require Cultural Archives for research bonus of Auto History Analyser

### LINKED ISSUE #1081: Scripting Bug - Auto history analyser should research iff culture library exists
### Environment
- **FreeOrion Version**: master head from this morning
### Description

Building BLD_AUTO_HISTORY_ANALYSER can only be built if BLD_CULTURE_ARCHIVES is there, but for producing research output it is not necessary. Probably one can build the analyser, then kill the population of and still get the benefit.
### Expected Result

Same as for the BLD_CULTURE_ARCHIVES, the BLD_AUTO_HISTORY_ANALYSER should not help researdch if the original species gets wiped out. 
### Implementation idea

Either auto-destroy history analyser or reduce research output to zero if BLD_CULTURE_ARCHIVES is not there.
Probably a sitrep message should be given to the user, so one knows why the building is not there anymore/does not produce research output.

dbenage-cx:
Fixes #1081

### DIFF
--- a/default/scripting/buildings/AUTO_HISTORY_ANALYSER.focs.txt
+++ b/default/scripting/buildings/AUTO_HISTORY_ANALYSER.focs.txt
@@ -19,6 +19,12 @@ BuildingType
                 NOT Population high = 0
                 OwnedBy empire = Source.Owner
             ]
+            activation = And [
+                ContainedBy And [
+                    Planet
+                    Contains Building name = "BLD_CULTURE_ARCHIVES"
+                ]
+            ]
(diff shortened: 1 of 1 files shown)


Vezzra:
@dbenage-cx, I assume you mean "Fixes #1081", right...?

Vezzra (approved):
Didn't run any tests, but looks simple and straightforward, so I guess this should work.

dbenage-cx:
@Vezzra TY, Fixed.

Issue #1285, opened 2017-02-05
AI still fails when taking over human empire


Voker57:
Save file: http://dump.bitcheese.net/files/muvukez/l_v_n_o.mps

I picked Voker57, set AI for L29Ah.

AI log: http://dump.bitcheese.net/files/nexolal/AI_4.log 

```
v0.4.6+ teamsmac [build 2017-02-05.7605a7f] CMake
...
2017-02-06 00:22:27.157337 [error] AI : : : Traceback (most recent call last):
2017-02-06 00:22:27.157368 [error] AI : : :   File "/opt/freeorion/share/freeorion/default/python/AI/freeorion_tools/_freeorion_tools.py", line 95, in wrapper
2017-02-06 00:22:27.157408 [error] AI : : :     return function(*args, **kw)
2017-02-06 00:22:27.157425 [error] AI : : :   File "/opt/freeorion/share/freeorion/default/python/AI/FreeOrionAI.py", line 139, in prepareForSave
2017-02-06 00:22:27.157457 [error] AI : : :     dump_string = pickle.dumps(foAIstate)
2017-02-06 00:22:27.157470 [error] AI : : :   File "/usr/lib/python2.7/pickle.py", line 1380, in dumps
2017-02-06 00:22:27.157581 [error] AI : : :     Pickler(file, protocol).dump(obj)
2017-02-06 00:22:27.157595 [error] AI : : :   File "/usr/lib/python2.7/pickle.py", line 224, in dump
2017-02-06 00:22:27.157628 [error] AI : : :     self.save(obj)
2017-02-06 00:22:27.157640 [error] AI : : :   File "/usr/lib/python2.7/pickle.py", line 331, in save
2017-02-06 00:22:27.157677 [error] AI : : :     self.save_reduce(obj=obj, *rv)
2017-02-06 00:22:27.157690 [error] AI : : :   File "/usr/lib/python2.7/pickle.py", line 425, in save_reduce
2017-02-06 00:22:27.157736 [error] AI : : :     save(state)
2017-02-06 00:22:27.157748 [error] AI : : :   File "/usr/lib/python2.7/pickle.py", line 286, in save
2017-02-06 00:22:27.157781 [error] AI : : :     f(self, obj) # Call unbound method with explicit self
2017-02-06 00:22:27.157793 [error] AI : : :   File "/usr/lib/python2.7/pickle.py", line 655, in save_dict
[...]

PR #1383, opened 2017-03-05
Catch exception when failing to pickle save state string.

geoffthemedio:
May prevent occasional AI crash when unable to pickle the save state. As suggested here: https://github.com/freeorion/freeorion/issues/1285

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -144,9 +144,12 @@ def prepareForSave():  # pylint: disable=invalid-name
     print "Preparing for game save by serializing state"
 
     # serialize (convert to string) global state dictionary and send to AI client to be stored in save file
-    dump_string = pickle.dumps(foAIstate)
-    print "foAIstate pickled to string, about to send to server"
-    fo.setSaveStateString(dump_string)
+    try:
+        dump_string = pickle.dumps(foAIstate)
+        print "foAIstate pickled to string, about to send to server"
+        fo.setSaveStateString(dump_string)
+    except:
+        print_error("foAIstate unable to pickle save-state string; the save file should be playable but the AI may have a different aggression.", trace=True)
 
 
 @chat_on_error


o01eg:
What if print exception data and stacktrace too?

default/python/AI/FreeOrionAI.py
```diff
@@ -138,9 +138,12 @@ def prepareForSave():  # pylint: disable=invalid-name
...
     # serialize (convert to string) global state dictionary and send to AI client to be stored in save file
-    dump_string = pickle.dumps(foAIstate)
-    print "foAIstate pickled to string, about to send to server"
-    fo.setSaveStateString(dump_string)
+    try:
+        dump_string = pickle.dumps(foAIstate)
+        print "foAIstate pickled to string, about to send to server"
+        fo.setSaveStateString(dump_string)
+    except:
+        print "foAIstate unable to picke string"
```
Dilvish-fo: picke --> pickle

and would probably be better to be a bit more specific: "unable to pickle save-state string"

Dilvish-fo: Hmm, thinking this through a bit more and perusing code-- this does look like it should stop the referenced crash-on-save problem, and the resulting save-file should be playable, but when loading from that save-file the affected AI will have lost its original aggression level and would then be playing as "Aggressive".  

Seems reasonable at least as a stopgap, but still enough of an issue that I am not sure I would want to make it as nearly invisible as this change would make it.  If instead of a simple print statement we use our print error function, then it will also send the warning message in red letters to the host.  In this case, I think the host would be getting this message every turn so we should probably also make it clear that the resulting save file should still be playable, or maybe even note the aggression issue (or point them to the Github issue?).
```        print_error("foAIstate unable to pickle save-state string; the save file should be playable but the AI may have a different aggression.", trace=True)```

Looking and thinking a bit more, I may have a fix for the underlying problem in the referenced issue, which I will try out.  But the changes discussed here could still be helpful improvements even if the immediate need went away from my hopeful fix.

Dilvish-fo: OK, 16b93b9 fixes the cited problem.   Catching any exceptions here is probably still a good thing, so long as there is at least a louder error message like I propose.

Dilvish-fo (changes requested):

geoffthemedio:
feel free to implement such as you see fit...

default/python/AI/FreeOrionAI.py
geoffthemedio: improved?

PR #1384, opened 2017-03-06
Allow colony construction for owned species gated by tech

dbenage-cx:
PR is to address not being able to construct a colony for tech gated species when a suitable colony is owned.  Reported in forum thread [here](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10443).
This issue affects extinct species ~and Exobots~.

Affected colony buildings are changed to start unlocked, with a location/enqueue requirement for either the tech or a suitable colony.


MatGB:
This all looks fine to me but it's such an edge case very difficult to test: given you did the work creating the techs &c so they'd be in the suitability report if you're happy it works I'm happy for it to go straight in.

dbenage-cx:
@MatGB TY, will wait a couple days to merge since there is no urgency.

default/scripting/buildings/colonies/SP_BANFORO.focs.txt
```diff
@@ -45,18 +45,21 @@ BuildingType
...
+                Or [
+                    And [
+                        Species name = "SP_BANFORO"
+                        Population low = [[MIN_RECOLONIZING_SIZE]]
+                        Happiness low = 5
+                    ]
+                    And [
+                        HasSpecial name = "EXTINCT_BANFORO_SPECIAL"
+                        Contains Building name = "BLD_XENORESURRECTION_LAB"
+                    ]
```
Vezzra: Correct me if I'm wrong, but shouldn't the `OwnerHasTech name = "TECH_COL_BANFORO"` condition be together with the
```
HasSpecial name = "EXTINCT_BANFORO_SPECIAL"
Contains Building name = "BLD_XENORESURRECTION_LAB"
```
block? Because if I read the implementation above correctly, having the Banforo colony tech alone would allow you to build a Banforo colony (the outer `OR` would be satisfied, and the `ResourceSupplyConnected` condition become irrelevant). Which I don't think is what you want to achieve.

So IMO it should look like this:
```
ResourceSupplyConnected empire = Source.Owner condition = And [
  Planet
  OwnedBy empire = Source.Owner
  Or [
    And [
      Species name = "SP_BANFORO"
      Population low = [[MIN_RECOLONIZING_SIZE]]
      Happiness low = 5
    ]
    And [
      OwnerHasTech name = "TECH_COL_BANFORO"
      HasSpecial name = "EXTINCT_BANFORO_SPECIAL"
      Contains Building name = "BLD_XENORESURRECTION_LAB"
    ]
  ]
]
```
Which should enable you to build a Banforo colony if either:

* you own an already established Banforo colony that meets the pop and happiness tresholds, and is supply connected to the planet on which you want to build the new colony.
* have the Banforo colony tech, and own a colony with the Banforo special and the Xenoresurrection Lab building which is supply connected to the planet on which you want to build the new colony.

Vezzra (changes requested):
Not sure about the change to the Exobots. These actually *are* supposed to be an exception to the rule that you can build a new colony via a colony building if you own an existing colony which is supply connected.

The fluff explanation being that Exobots aren't really a real species on their own, but just some enhanced autonomous robots (hence the bad stats).

To actually create a new colony (be it via the colony building or via deploying an Exobot colony ship) you need the tech.

The increased costs for the colony building and the fixed build time for it reflect that. If the Exobots could colonize just like other species, the costs and build time should be the same as for other species in case an already existing Exobot colony acts as "colonists source".

So in this particular case I'd rather stick with not being able to build an Exobot colony building or Exobot colony ship if you don't have the tech.

default/scripting/buildings/colonies/SP_BANFORO.focs.txt
dbenage-cx: TY, I will update and test Exobots accordingly.

For extinct species I envisioned that building a lab on ruins grants the knowledge to reconstruct the species without any other resource requirement from the ruins.
So either of:
*  Has tech
*  Has supply connection to suitable colony

In that case, I probably should have removed the `And` block here, resulting in:

```
Or [
    OwnerHasTech name = "TECH_COL_BANFORO"
    ResourceSupplyConnected empire = Source.Owner condition = And [
        Planet
        OwnedBy empire = Source.Owner
        Species name = "SP_BANFORO"
        Population low = [[MIN_RECOLONIZING_SIZE]]
        Happiness low = 5
    ]
]
```

Regardless of which method is desired, I should probably add some info to pedia articles reflecting the special requirements compared to a normal species.

Vezzra: Hm, IIRC it has been @Deepsloth who designed and implemented the extinct species, but I *think* the idea is that you need some kind of "remnants" of members of the extinct species (represented by the special) and the lab to clone them, this way providing an alternative source of colonists (as you don't have an existing colony).

The knowledgle alone isn't sufficient. You need sufficient "raw material" to work with. Hence the necessity to have supply connection to a colony with the "raw material" and the facility to make colonists out that.

Personally I prefer that approach, this way you can loose the species again if you loose all their colonies as well as the planet where you found their remnants originally. More interesting IMO. 

The other approach would mean once you got the tech, you can just recreate them whenever and wherever you want. Which is something you can't even do with normal species, so this strikes me as quite unbalanced.

Dilvish-fo: I concur with Vezzra's comments and proposal.

dbenage-cx: Will do ;)

dbenage-cx:
Still need to update pedia articles.

Vezzra:
If you want to do that in this PR, tag it as work in progress.

However, I think it would be better to do that in a separate PR, so we can merge this one already. The sooner it's in master, the sooner it will receive more thorough testing. 😉

dbenage-cx:
Will do a separate PR, have I missed any concerns in your review?

default/scripting/starting_unlocks/items.inf
```diff
@@ -12,20 +12,24 @@ Item type = Building name = "BLD_SHIPYARD_ORBITAL_DRYDOCK"
...
 Item type = Building name = "BLD_COL_SUPER_TEST"
 Item type = Building name = "BLD_COL_ABADDONI"
+Item type = Building name = "BLD_COL_BANFORO"
 Item type = Building name = "BLD_COL_CHATO"
 Item type = Building name = "BLD_COL_CRAY"
 Item type = Building name = "BLD_COL_DERTHREAN"
 Item type = Building name = "BLD_COL_EAXAW"
 Item type = Building name = "BLD_COL_EGASSEM"
 Item type = Building name = "BLD_COL_ETTY"
+Item type = Building name = "BLD_COL_EXOBOT"
```
Vezzra: Is that still needed here? If the Exobots are the big exception, you should only be able to build their colony building if you have the tech, so no need to pre-unlock it (unless I'm missing something).

Vezzra (commented):
Aside from my last remark on not preunlocking the Exobot colony building, looks fine as far as I can tell.

Vezzra (approved):

dbenage-cx:
TY for reviews

PR #1385, opened 2017-03-07
Use defined boost include dirs for freeorion and freeoriond

dbenage-cx:
As of cf67d52 building with a custom defined `Boost_INCLUDE_DIRS` fails to link.
The freeoriond and freeorion targets are including the system supplied boost include directory instead.

[make-cf67d52.log.txt](https://github.com/freeorion/freeorion/files/823435/make-cf67d52.log.txt)

Appears a simple change, but posting as PR in event it should be handled differently.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,6 +286,7 @@ target_compile_definitions(freeoriond
 
 target_include_directories(freeoriond SYSTEM
     PRIVATE
+        ${Boost_INCLUDE_DIRS}
         ${PROJECT_SOURCE_DIR}/GG
         ${PYTHON_INCLUDE_PATH}
 )
@@ -341,6 +342,7 @@ target_compile_definitions(freeorion
 
 target_include_directories(freeorion SYSTEM
     PRIVATE
+        ${Boost_INCLUDE_DIRS}
         ${OPENGL_INCLUDE_DIR}
         ${OPENAL_INCLUDE_DIR}
         ${SDL2_INCLUDE_DIR}


adrianbroher:
> As of cf67d52 building with a custom defined Boost_INCLUDE_DIRS fails to link.
The freeoriond and freeorion targets are including the system supplied boost include directory instead.

Just to be clear:  You're having a custom installed boost in a nonstandard location and populating the `-DBoost_INCLUDE_DIRS` value or do you leverage the FindBoost module by setting `-DBOOST_ROOT` when calling `cmake` in configuration mode?

But the change itself looks okay.

adrianbroher:
Also do I need to wonder about

```
undefined reference to `boost::this_thread::hiden::sleep_for(timespec const&)'
```

and friends in freeoriond for the same reason (nonstandard boost installation)?

adrianbroher (approved):

dbenage-cx:
I believe the undefined reference is from differing boost version during link (GiGi, common, and parse used the requested include location, hinted by lack of pragma message litter).

I've compiled boost from source for different versions and use the first option mentioned:
build$ ccmake -DBoost_LIBRARY_DIR=*custom_lib_dir* -DBoost_INCLUDE_DIR=*custom_include_dir* ..

(ccmake reports and sets component directories for `Boost_INCLUDE_DIR`, but CMakeLiists.txt uses `Boost_INCLUDE_DIRS`?)

I did not initially have success exporting `BOOST_INCLUDEDIR`, `BOOST_LIBRARYDIR`, and `Boost_ADDITIONAL_VERSIONS` so simply hacked new version numbers into the system installed FindBoost module, without changes for component dependencies from 1.61.
`ldd` shows all targets link to desired boost version.  If you want, I can list out the full process.

adrianbroher:
> ccmake reports and sets component directories for Boost_INCLUDE_DIR, but CMakeLiists.txt uses Boost_INCLUDE_DIRS

`*_DIR` or `*_LIBRARY` are 'advanced variables' by definition.  Those are intended for manual configuration if the user wants to set up a special build with manual selected libraries or when the install tree/build tree layout of the dependency differ so much that the corresponding CMake Find* module fails to recognize the dependency.  I would assume that you don't have modified the boost build system so this should not be  required.

The `*_DIRS` and `*_LIBRARIES` variables are generated by the Find module and compose the user set or automatically found `*_DIR` and `*_LIBRARY` together in a way that the project can be build.  Some dependencies may consist of multiple libraries that need to be linked in a certain order with compiler specific flags and the Find module takes care of that.  So a client project should use `*_DIRS` and `*_LIBRARIES` in their CMakeLists.txt and end users can configure special layouts by setting `*_DIR` and `*_LIBRARY` but shouldn't required to do so most of the time.

Did you install boost locally to some prefix (for example `/usr/local`)?  If yes, then you can set `BOOST_ROOT` to this installation prefix as a search hint (again example: -DBOOST_ROOT=/usr/local) and the stock FindBoost module should find the installation without problems. `BOOST_INCLUDEDIR`, `BOOST_LIBRARYDIR` and `Boost_ADDITIONAL_VERSIONS` are additional hints, that may be required if the installation layout again is non-standard or if you use an old FindBoost.cmake module with a most recent boost version.

> If you want, I can list out the full process.

Feel free to do so if you still have questions, but please do so in the programming forum.

adrianbroher:
Also the stock Find modules come with good documentation in my opinion. You can call

```
cmake --help-module <module name>
```

e.g.

```
cmake --help-module FindBoost
```

to get the documentation on the command line as reStructureText.

dbenage-cx:
Thank you for the detailed info, since this is not an issue due to my initial configuration, will go ahead and merge.  (same issue using `BOOST_ROOT` + updated FindBoost.cmake)

PR #1387, opened 2017-03-08
Refactor ShipDesignAI to allow using CombatRatingsAI functions

Morlic-fo:
Changes in this PR:

1. Introduce a new class DesignStats to ShipDesignAI which contains the stats of the (to be) optimized design including species and tech modifiers
2. Let ShipDesignAI.optimize_design() return such an instance to allow other functions access to the actual stats of the design
3. Provide an interface to get a combat rating for such DesignStats from the relevant CombatRatingsAI function to avoid code duplication in the WarshipDesigner rating functions.
4. Let ProductionAI.cur_best_military_rating() return the actual combat rating of our best military design instead of the internal rating of the ShipDesignAI. This seems to be a cleaner implementation compared to what was tried in #1329 and is intended to supersede that PR.

### DIFF
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -12,6 +12,7 @@ import PriorityAI
 import ColonisationAI
 import MilitaryAI
 import ShipDesignAI
+import CombatRatingsAI
 from turn_state import state
 
 from EnumsAI import (PriorityType, EmpireProductionTypes, MissionType, get_priority_production_types,
@@ -96,15 +97,18 @@ def cur_best_military_design_rating():
     if current_turn in _best_military_design_rating_cache:
         return _best_military_design_rating_cache[current_turn]
     priority = PriorityType.PRODUCTION_MILITARY
-    if priority in _design_cache:
-        if _design_cache[priority] and _design_cache[priority][0]:
-            rating, pid, design_id, cost, stats = _design_cache[priority][0]
-            _best_military_design_rating_cache[current_turn] = rating
-            return max(rating, 0.001)
-        else:
-            return 0.001
-    else:
-        return 0.001
+    if priority in _design_cache and _design_cache[priority] and _design_cache[priority][0]:
+        # the rating provided by the ShipDesigner does  not
+        # reflect the rating used in threat considerations
+        # but takes additional factors (such as cost) into
+        # account. Therefore, we want to calculate the actual
+        # rating of the design as provided by CombatRatingsAI.
+        _, _, _, _, stats = _design_cache[priority][0]
+        # TODO: Should this consider enemy stats?
+        rating = CombatRatingsAI.ShipCombatStats(stats=stats.convert_to_combat_stats()).get_rating()
+        _best_military_design_rating_cache[current_turn] = rating
+        return max(rating, 0.001)
+    return 0.001
 
 
 def get_best_ship_info(priority, loc=None):
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -98,7 +98,7 @@ def cur_best_military_design_rating():
     priority = PriorityType.PRODUCTION_MILITARY
     if priority in _design_cache:
         if _design_cache[priority] and _design_cache[priority][0]:
-            rating, pid, design_id, cost = _design_cache[priority][0]
+            rating, pid, design_id, cost, stats = _design_cache[priority][0]
             _best_military_design_rating_cache[current_turn] = rating
             return max(rating, 0.001)
         else:
@@ -123,7 +123,7 @@ def get_best_ship_info(priority, loc=None):
             return None, None, None
 
         for design_stats in best_designs:
-            top_rating, pid, top_id, cost = design_stats
+            top_rating, pid, top_id, cost, stats = design_stats
             if pid in planet_ids:
                 break
         valid_locs = [item[1] for item in best_designs if item[0] == top_rating and item[2] == top_id]
@@ -156,10 +156,10 @@ def get_best_ship_ratings(planet_ids):
     priority = PriorityType.PRODUCTION_MILITARY
     planet_ids = set(planet_ids).intersection(ColonisationAI.empire_shipyards)
 
-    if priority in _design_cache:  # use new framework
+    if priority in _design_cache:
         build_choices = _design_cache[priority]
-        loc_choices = [[item[0], item[1], item[2], fo.getShipDesign(item[2])]
-                       for item in build_choices if item[1] in planet_ids]
+        loc_choices = [[rating, pid, design_id, fo.getShipDesign(design_id)]
+                       for (rating, pid, design_id, cost, stats) in build_choices if pid in planet_ids]
         if not loc_choices:
             return []
         best_rating = loc_choices[0][0]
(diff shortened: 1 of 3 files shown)


Vezzra:
> This seems to be a cleaner implementation compared to what was tried in #1329 and is intended to supersede that PR.

In that case #1329 should be closed and tagged as "status:follow up". Also put a reference to the superseding PR (this one here) there.

Morlic-fo:
> In that case #1329 should be closed and tagged as "status:follow up". Also put a reference to the superseding PR (this one here) there.

Right, done.

Vezzra:
Thanks!

default/python/AI/ShipDesignAI.py
```diff
@@ -703,30 +731,157 @@ def __init__(self):
...
-        self.maximum_organic_growth = 0
-        self.repair_per_turn = 0
-        self.asteroid_stealth = 0
-        self.solar_stealth = 0
-        self.fighter_capacity = 0
-        self.fighter_launch_rate = 0
-        self.fighter_damage = 0
+
+    @property
+    def attacks(self):
```
Cjkjvfnby: I don't sure bu may be we just can give direct access to `_design_stats` object?    Looks like you provide all setters and getter to it, so there is not any benefits of encapsulation.

default/python/AI/ShipDesignAI.py
```diff
@@ -1087,7 +1242,7 @@ def optimize_design(self, additional_parts=(), additional_hulls=(),
 
         Only designs with a positive rating (i.e. matching the minimum requirements) will be returned.
 
-        :return: list of (rating,planet_id,design_id,cost) tuples, i.e. best available design for each planet
+        :return: list of (rating,planet_id,design_id,cost, design_stats) tuples, i.e. best available design for each planet
```
Cjkjvfnby: inconsistency in spaces after coma

default/python/AI/ShipDesignAI.py
```diff
@@ -1751,28 +1911,10 @@ def __init__(self):
     def _rating_function(self):
         if self.fighter_capacity < 1:
             return INVALID_DESIGN_RATING
-        # first, calculate "normal" weapon stuff
-        weapon_dmg = max(self._total_dmg_vs_shields(), self._total_dmg() / 1000)
-        effective_structure = self._effective_structure()
-
-        # now, consider offensive potential of our fighters
```
Cjkjvfnby, line 1758: Looks like this part totally dropped, I assume it is code duplication you talk about?

Cjkjvfnby (approved):
I set approve status to review, since have limited time, and probably will not be able to review fixes.

default/python/AI/ShipDesignAI.py
```diff
@@ -662,6 +662,34 @@ def convert_to_tuple(self):
                 "enemyMineDmg: %s" % self.enemy_mine_dmg)
 
 
+class DesignStats(object):
+
+    def __init__(self):
+        self.attacks = {}  # {damage: shots_per_turn}
```
dbenage-cx: `shots_per_round`?

default/python/AI/ShipDesignAI.py
Dilvish-fo: I likewise tend to have at least a bit of a preference to just using direct access to the _design_stats object, but I also see that this approach avoids having to change a zillion references to the various stats.  I don't see an easy way to do that refactoring with PyCharm's refactoring tools, and although simple search and replace might for for most or hopefully all of it, trying to do that would add some extra uncertainty/risk and need for more testing.  

I think it is probably best to go with this PR's approach for now, and we could consider refactoring more later on.

Dilvish-fo (approved):
dbenage-cx' correction about the shots_per_turn ==> shots_per_round (and CJs whitespace correction) look to me like they should be made, and other than that it all looks fine to me (and seems to work fine in testing).

default/python/AI/ProductionAI.py
```diff
@@ -96,15 +97,18 @@ def cur_best_military_design_rating():
...
-        return 0.001
+    if _design_cache.get(priority, None) and _design_cache[priority][0]:
+        # the rating provided by the ShipDesigner does not
+        # reflect the rating used in threat considerations
+        # but takes additional factors (such as cost) into
+        # account. Therefore, we want to calculate the actual
+        # rating of the design as provided by CombatRatingsAI.
+        _, _, _, _, stats = _design_cache[priority][0]
+        # TODO: Should this consider enemy stats?
+        rating = CombatRatingsAI.ShipCombatStats(stats=stats.convert_to_combat_stats()).get_rating()
```
Dilvish-fo, line 108: yes, I think it would be best for this to reference the standard enemy
```    get_rating(enemy_stats=foAI.foAIstate.get_standard_enemy())```

default/python/AI/ShipDesignAI.py
Morlic-fo: Yes, this part is covered in the CombatRatingsAI function call.

default/python/AI/ProductionAI.py
Morlic-fo: I am not too sure if this is actually the standard use case of the function.

We may compare our ship strength with an actual specific monster. So at least the enemy should be  provided as an argument. I am undecided if using a standard enemy should be the default behaviour or using None.

default/python/AI/ShipDesignAI.py
Morlic-fo: I was just lazy as the property-based approach was a 30 sec script solution. I replaced it with direct access.

default/python/AI/ProductionAI.py
Dilvish-fo: Having an argument for the enemy does sound best to me, but all the current uses would be against the standard enemy so I think the default should be to use the standard enemy rather than None.

There is only 1 current regular use, by PriorityAI.  The PriorityAI is trying to decide how much priorty to give to military ship production, comparing unmet threat (from a variety of sources) to the value returned by this function, to get an estimate of how many additional Military ships we feel an existing need for.  That could be made a more fine-grained analysis, but for now I would prefer it to use the standard enemy.

It is also currently optionally referenced in by MilitaryAI.cur_best_mil_ship_rating(), and that itself is used in a number of places; this function is getting the rating of our best existing ship (and optionally design), the ship rating is always gotten against the standard enemy and so if any of the places calling there wanted to also check designs they would need them evaluated against the standard enemy also (or, as with the PriorityAI, a future tweak could be that in some cases they both be evaluated against specific enemies).

Also, in this new code it looks to me like you have the new method _combat_rating() *always* using the standard enemy when considering warship ratings, so it seems very consistent for us to have the standard enemy as default for cur_best_military_design_rating().

Dilvish-fo: Hmm, thinking about this a bit more, the mechanics of just how we'd want to have the default be an up-to-date value for foAI.foAIstate.get_standard_enemy(), while still allowing for None to be an acceptable value, might take a  bit of thinking/talking; it's far simpler to just let the default be None as you suggest and then we just change the current calls in PriorityAI and in MilitaryAI so that they specify enemy_stats=foAI.foAIstate.get_standard_enemy() when they call into cur_best_military_design_rating

PR #1389, opened 2017-03-09
Add Missing Include/Library Paths for GiGiSDL

kdhp:


adrianbroher:
Thanks for the PR.  I assume this bug was only exposed on FreeBSD, given your other repositories?

kdhp:
@adrianbroher I noticed it on FreeBSD, but it would have been an issue on other
systems where OpenGL/GLEW are not in the default include path.

kdhp:
@adrianbroher  Oh, the variables you corrected were copied _verbatim_ from
earlier in the file.

[GG/CMakeLIsts.txt:148](https://github.com/freeorion/freeorion/blob/4a8ab850ce026ed64f519dce2dabf5b4d00102c4/GG/CMakeLists.txt#L148)

adrianbroher:
@kdhp thanks for pointing this out.  Fixed with b1fecab.

Vezzra:
@adrianbroher, shouldn't status be "merged" instead of "resolved"? If I understood correctly, "resolved" is strictly for issues, and "merged" for PRs.

adrianbroher:
This PR wasn't merged.

Vezzra:
Hm, I thought you picked the relevant parts of this PR into your commit, so I assumed it would still fall under what's covered in the description of the "status:merged" label:

> All relevant commits of this PR were merged into the master development branch. This can either happen by merging, cherry-picking or rebasing and pushing onto master. Close the PR if GitHub didn't close it automatically.

If not, and we want to use "status:resolved" for cases like this, we either need to extend the description of the "status:resolved" label...

> The Issue was resolved satisfactorily, either by answering properly or fixing the underlying bug. Close the Issue.

...accordingly, or add a new label specifically for PRs that weren't rejected, also not really merged, but closed by a commit that incorporated the relevant parts of a PR.

Another option could be to use "status:follow up" and extend that label also to apply to PRs which have been "followed up" and closed by a commit (instead of another follow up PR).

Personally I'd lean toward the last option, what do you think?

PR #1390, opened 2017-03-09
Fix bug 1382 SidePanel System name out of place.

### LINKED ISSUE #1382: System Name Misplaced
I've only observed this after loading an MP game, and not SP. When the game loads, the map is shown without the sidepanel open. Clicking on a system... possibly only ones without populated planets in them, will open the sidepanel with the system name misplaced at the top right. Viewing another system or closing and reopening the sidepanel fixes it.

deep_space_shifted

LGM-Doyle:
This fixes #1382.

The problem was that the system name width was not set in the SidePanel layout.

### DIFF
--- a/GG/src/ListBox.cpp
+++ b/GG/src/ListBox.cpp
@@ -265,10 +265,10 @@ void ListBox::Row::Render()
 {}
 
 void ListBox::Row::GrowWidthsStretchesAlignmentsTo(std::size_t nn) {
-    while (m_col_widths.size() < nn) {
-        m_col_widths.emplace_back(X(5));
-        m_col_alignments.emplace_back(ALIGN_NONE);
-        m_col_stretches.emplace_back(0.0);
+    if (m_col_widths.size() < nn) {
+        m_col_widths.resize(nn, X(5));
+        m_col_alignments.resize(nn, ALIGN_NONE);
+        m_col_stretches.resize(nn, 0.0);
     }
 }
 
--- a/GG/src/ListBox.cpp
+++ b/GG/src/ListBox.cpp
@@ -264,21 +264,25 @@ bool ListBox::Row::IsNormalized() const
 void ListBox::Row::Render()
 {}
 
+void ListBox::Row::GrowWidthsStretchesAlignmentsTo(std::size_t nn) {
+    while (m_col_widths.size() < nn) {
+        m_col_widths.emplace_back(X(5));
+        m_col_alignments.emplace_back(ALIGN_NONE);
+        m_col_stretches.emplace_back(0.0);
+    }
+}
+
 void ListBox::Row::push_back(Control* c)
 {
     m_cells.push_back(c);
-    m_col_widths.push_back(X(5));
-    m_col_alignments.push_back(ALIGN_NONE);
-    m_col_stretches.push_back(0.0);
-    if (1 < m_cells.size())
-        m_col_widths.back() = m_col_widths[m_cells.size() - 1];
-
-    std::size_t ii = m_cells.size() - 1;
+    GrowWidthsStretchesAlignmentsTo(m_cells.size());
+    auto ii = m_cells.size() - 1;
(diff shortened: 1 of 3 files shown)


UI/SidePanel.cpp
```diff
@@ -3211,7 +3211,9 @@ void SidePanel::DoLayout() {
 
     // system name / droplist
     ul = GG::Pt(GG::X(MaxPlanetDiameter()), GG::Y0);
-    lr = ul + GG::Pt(ClientWidth() - GG::X(MaxPlanetDiameter()), name_height);
+    auto system_name_width = ClientWidth() - GG::X(MaxPlanetDiameter());
```
geoffthemedio, line 3214: Typing X here would have been shorter than auto

LGM-Doyle: Thanks for taking the time to review this PR.

Yes, typing X would have been less text that `auto`.

However, the type inference from `auto` provides more benefits than saved typing:
- `auto` variables are always initialized.  The type inference of `auto` only occurs if it has a constructor from which to infer the type.  `SomeType x;` may/may not be initialized, but `auto x = <anything that compiles>` is always initialized.  Using `auto` protects against unintentionally uninitialized variables.
- `auto` variables are always just constructed.  They are not constructed and then moved or copied. 
- pointer/reference `auto` variables are never aliased/sliced down to a base type, even if the type on the right hand side changes in the future after the code is written.  The type is correctly inferred at compile time.
- `auto` variables will not call implicit type converting constructors or casts.

For these reasons `auto` is more letters than `X`, but better code.

geoffthemedio: Is any of that relevant in this case, where the type is X, a typedef of int which is immediately assigned a value?

GG/src/ListBox.cpp
```diff
@@ -264,21 +264,25 @@ bool ListBox::Row::IsNormalized() const
 void ListBox::Row::Render()
 {}
 
+void ListBox::Row::GrowWidthsStretchesAlignmentsTo(std::size_t nn) {
+    while (m_col_widths.size() < nn) {
+        m_col_widths.emplace_back(X(5));
```
geoffthemedio: why not use `vector::resize(nn, X(5))` and similar? the loop seems oddly verbose for this purpose. a check that `nn >= m_col_widths.size()` would more explicitly ensure no shrinkage if that is desired.

geoffthemedio:
builds and seems to fix issue #1382

GG/src/ListBox.cpp
LGM-Doyle: Fixed.

UI/SidePanel.cpp
LGM-Doyle: Yes.  Two reasons.

First, there are many integral type that are implicitly cast to int, some signed, some unsigned, some larger and some smaller than int.  If you write `X` there may be an implicit cast and perhaps a bug.  `auto` prevents this bug.  

Second, future proofing.  The type on the right side could change.

`GG` uses `X` to represent absolute locations, relative locations and dimensions.  A bug that I have regularly introduced and occasionally fixed 😮 is to replace one for the other.  A reasonable improvement to the code base would be to create 3 types that are not implicitly converted.  In which case `ClientWidth() - X` would return an XDim which is not convertible to X.  Using X here means that this line would just keep working correctly.

You are right though.  For any given line of code, if the inferred type is not a type only known to the compiler, if you replace the inferred type with a written out type, and you know there are no implicit conversions, then you have the exact same generated code.  The point of `auto` is to off-load all of that checking on the compiler.

Fewer opportunities for mistakes.

geoffthemedio: I think implicit casts to X are disabled?

I question whether the supposed "future proofing" would really be of much help, compared with the clarity of explicitly stating the type in such cases, but OK...

LGM-Doyle:
Thanks.

PR #1378, opened 2017-03-02
Categorize Ship Part pedia articles

dbenage-cx:
Resolves #1346 

Disabled definitions were added for Industry, Research, "Trade", and Production Location parts
(no stringtable description added for "trade")

PR #1391, opened 2017-03-09
Fix missing pedia articles in Ship Part subcategory

Ouaz:
- "Distortion Modulator" and "Dimensional Cleaver" pedia articles missing in "Ship Part/General" subcategory
(follows #1378 Categorize Ship Part pedia articles)


dbenage-cx:
Changing to `bug` as this is a correction to the category.

TY for catching and fixing @Ouaz

PR #1392, opened 2017-03-09
Refactor Message class.

o01eg:
1. Remove sender_id and receiver_id from header because we know this data in most cases from player connection.
2. Merge PLAYER_CHAT and LOBBY_CHAT messages.
3. Send broadcast Message by ServerNetworking::SendMessageAll and other
Message by PlayerConnection::SendMessage.
4. Add player_id into Message body where it was used in header.

### DIFF
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -199,11 +199,7 @@ void AIClientApp::HandleMessage(const Message& msg) {
         if (text.empty()) {
             ErrorLogger() << "AIClientApp::HandleMessage for HOST_ID : Got empty message text?!";
         } else {
-            try {
-                host_id = boost::lexical_cast<int>(text);
-            } catch (...) {
-                ErrorLogger() << "AIClientApp::HandleMessage for HOST_ID : Couldn't parese message text: " << text;
-            }
+            host_id = boost::lexical_cast<int>(text);
         }
         m_networking->SetHostPlayerID(host_id);
         break;
@@ -213,12 +209,8 @@ void AIClientApp::HandleMessage(const Message& msg) {
         if (PlayerID() == Networking::INVALID_PLAYER_ID) {
             DebugLogger() << "AIClientApp::HandleMessage : Received JOIN_GAME acknowledgement";
             const std::string& text = msg.Text();
-            try {
-                int player_id = boost::lexical_cast<int>(text);
-                m_networking->SetPlayerID(player_id);
-            } catch(...) {
-                ErrorLogger() << "AIClientApp::HandleMessage : Received erroneous JOIN_GAME acknowledgement " << text;
-            }
+            int player_id = boost::lexical_cast<int>(text);
+            m_networking->SetPlayerID(player_id);
         } else {
             ErrorLogger() << "AIClientApp::HandleMessage : Received erroneous JOIN_GAME acknowledgement when already in a game";
         }
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -341,7 +341,6 @@ void AIClientApp::HandleMessage(const Message& msg) {
     }
 
     case Message::PLAYER_CHAT: {
-        const std::string& text = msg.Text();
         std::string data;
         int player_id;
         ExtractServerPlayerChatMessageData(msg, player_id, data);
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -304,7 +304,7 @@ void AIClientApp::HandleMessage(const Message& msg) {
 
     case Message::SAVE_GAME_DATA_REQUEST: {
         //DebugLogger() << "AIClientApp::HandleMessage Message::SAVE_GAME_DATA_REQUEST";
-        Networking().SendMessage(ClientSaveDataMessage(PlayerID(), Orders(), m_AI->GetSaveStateString()));
+        Networking().SendMessage(ClientSaveDataMessage(Orders(), m_AI->GetSaveStateString()));
         //DebugLogger() << "AIClientApp::HandleMessage sent save data message";
         break;
     }
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -181,7 +181,7 @@ void AIClientApp::HandlePythonAICrash() {
     err_msg << "AIClientApp failed due to error in python AI code for " << PlayerName() << ".  Exiting Soon.";
     ErrorLogger() << err_msg.str() << " id = " << PlayerID();
     Networking().SendMessage(
-        ErrorMessage(PlayerID(), str(FlexibleFormat(UserString("ERROR_PYTHON_AI_CRASHED")) % PlayerName()) , true));
+        ErrorMessage(str(FlexibleFormat(UserString("ERROR_PYTHON_AI_CRASHED")) % PlayerName()) , true));
 }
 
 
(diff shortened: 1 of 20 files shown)


Vezzra:
@freeorion/developers, this looks like a rather extensive refactoring, so I wonder if it wouldn't be better to postpone that until after the release (meaning, assigning this to the "post 0.4.7" milestone). Or do you think it's safe to include it into 0.4.7?

Setting milestone to "Undecided/TBD" for now.

client/AI/AIClientApp.cpp
```diff
@@ -359,9 +357,14 @@ void AIClientApp::HandleMessage(const Message& msg) {
         break;
     }
 
-    case Message::PLAYER_CHAT:
-        m_AI->HandleChatMessage(msg.SendingPlayer(), msg.Text());
+    case Message::PLAYER_CHAT: {
+        const std::string& text = msg.Text();
+        std::string data;
```
LGM-Doyle: Is `text` unused?

client/human/HumanClientFSM.cpp
```diff
@@ -99,19 +99,27 @@ WaitingForSPHostAck::~WaitingForSPHostAck()
...
-    Client().GetClientUI().GetMapWnd()->Sanitize();
+        Client().Networking().SetPlayerID(host_id);
+        Client().Networking().SetHostPlayerID(host_id);
 
-    return transit<PlayingGame>();
+        Client().GetClientUI().GetMapWnd()->Sanitize();
+
+        return transit<PlayingGame>();
+    } catch (...) {
+        ErrorLogger() << "WaitingForSPHostAck::react(const HostSPGame& msg) couldn't parese message text: " << msg.m_message.Text();
```
LGM-Doyle: Spelling parese -> parse.

And I'd use a message like

`"Host id " << msg.m_message.Text() << " is not a number."`

client/human/HumanClientFSM.cpp
```diff
@@ -99,19 +99,27 @@ WaitingForSPHostAck::~WaitingForSPHostAck()
...
 
-    Client().GetClientUI().GetMapWnd()->Sanitize();
+        Client().Networking().SetPlayerID(host_id);
+        Client().Networking().SetHostPlayerID(host_id);
 
-    return transit<PlayingGame>();
+        Client().GetClientUI().GetMapWnd()->Sanitize();
+
+        return transit<PlayingGame>();
+    } catch (...) {
```
LGM-Doyle: Again, just catch the `bad_lexical_cast` or add to the serialization of this message type.

client/human/HumanClientFSM.cpp
```diff
@@ -145,17 +153,25 @@ WaitingForMPHostAck::~WaitingForMPHostAck()
...
-    Client().Networking().SetHostPlayerID(msg.m_message.ReceivingPlayer());
+    try {
+        int host_id = boost::lexical_cast<int>(msg.m_message.Text());
+
+        Client().Networking().SetPlayerID(host_id);
+        Client().Networking().SetHostPlayerID(host_id);
 
-    return transit<MPLobby>();
+        return transit<MPLobby>();
+    } catch (...) {
```
LGM-Doyle: Again, just catch the `bad_lexical_cast` or add to the serialization of this message type.

geoffthemedio:
Builds OK in MSVC, no obvious problems upon trivial test.

As this doesn't fix a problem, but could introduce some, it could wait until after v0.4.7

client/human/HumanClientFSM.cpp
```diff
@@ -189,16 +205,24 @@ WaitingForMPJoinAck::~WaitingForMPJoinAck()
...
 
-    Client().Networking().SetPlayerID(msg.m_message.ReceivingPlayer());
+    try {
+        int player_id = boost::lexical_cast<int>(msg.m_message.Text());
+
+        Client().Networking().SetPlayerID(player_id);
 
-    return transit<MPLobby>();
+        return transit<MPLobby>();
+    } catch (...) {
```
LGM-Doyle: Again, just catch the `bad_lexical_cast` or add to the serialization of this message type.

network/Message.cpp
```diff
@@ -140,41 +115,29 @@ void swap(Message& lhs, Message& rhs)
...
 }
 
 void HeaderToBuffer(const Message& message, Message::HeaderBuffer& buffer) {
     buffer[0] = message.Type();
-    buffer[1] = message.SendingPlayer();
-    buffer[2] = message.ReceivingPlayer();
-    buffer[3] = message.SynchronousResponse();
-    buffer[4] = message.Size();
+    buffer[1] = message.SynchronousResponse();
+    buffer[2] = message.Size();
```
LGM-Doyle: Could you replace these magic constants with an enumeration and then use the enumeration in place of the constants?
````
class Message {
public:
...
enum class Parts {TYPE, RESPONSE, SIZE};
}
````

I know that the old code uses the magic numbers, but this would improve things.

File: network/Message.cpp [hunk 7]
```diff
@@ -379,34 +342,34 @@ Message GameStartMessage(int player_id, bool single_player_game, int empire_id,
...
-{ return Message(Message::HOST_SP_GAME, Networking::INVALID_PLAYER_ID, player_id, ACKNOWLEDGEMENT); }
+{ return Message(Message::HOST_SP_GAME, std::to_string(player_id)); }
 
 Message HostMPAckMessage(int player_id)
-{ return Message(Message::HOST_MP_GAME, Networking::INVALID_PLAYER_ID, player_id, ACKNOWLEDGEMENT); }
+{ return Message(Message::HOST_MP_GAME, std::to_string(player_id)); }
 
 Message JoinAckMessage(int player_id)
-{ return Message(Message::JOIN_GAME, Networking::INVALID_PLAYER_ID, player_id, ACKNOWLEDGEMENT); }
+{ return Message(Message::JOIN_GAME, std::to_string(player_id)); }
```
LGM-Doyle, line 354: For all of these `Message` types that include the `player_id` could you use `boost::serialization` to stream them out and in and avoid the exceptions from the lexical casting?

network/Message.cpp
```diff
@@ -416,7 +379,7 @@ Message PlayerStatusMessage(int player_id, int about_player_id, Message::PlayerS
         oa << BOOST_SERIALIZATION_NVP(about_player_id)
            << BOOST_SERIALIZATION_NVP(player_status);
```
LGM-Doyle, line 379: Could you send/receive the other player ids in this same fashion?

LGM-Doyle (commented):
@Vezzra this is straightforward enough that it either works or it does not.  We have not branched v0.4.7 or officially started testing.  I think that any problems with this will show up in testing and benching it for weeks is unwarranted.

However, I bow to @geoffthemedio's greater experience.

@o01eg the code is good, clear and easy to understand.

You explained what the changes that you made are, but you did not explain why you made them?

Why is this better?

o01eg:
> Why is this better?

1. It is more logically as it isn't P2P application and we don't need to have own transport protocol.
2. Reduce inconsistency if player_id in Message differs from connection's player_id. Just use connection associated data.
3. It was part of reducing non-unique player_id usage in favour unique player_name which going to be used for user authentication.

client/AI/AIClientApp.cpp
```diff
@@ -221,102 +221,103 @@ void AIClientApp::HandleMessage(const Message& msg) {
...
-            } else {
-                ErrorLogger() << "AIClientApp::HandleMessage : Received erroneous JOIN_GAME acknowledgement when already in a game";
+        if (PlayerID() == Networking::INVALID_PLAYER_ID) {
+            DebugLogger() << "AIClientApp::HandleMessage : Received JOIN_GAME acknowledgement";
+            const std::string& text = msg.Text();
+            try {
+                int player_id = boost::lexical_cast<int>(text);
+                m_networking.SetPlayerID(player_id);
+            } catch(...) {
+                ErrorLogger() << "AIClientApp::HandleMessage : Received erroneous JOIN_GAME acknowledgement " << text;
```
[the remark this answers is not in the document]
o01eg: 1. I suppose it simpler for single value then serialization also it was used in HOST_ID handler.
2. OK, I'll better remove exception handler there.

client/AI/AIClientApp.cpp
o01eg: Yes, I'll remove it.

client/human/HumanClientFSM.cpp
o01eg: OK, I'll specify exception type.

client/human/HumanClientFSM.cpp
o01eg: Yep

client/human/HumanClientFSM.cpp
o01eg: Yep

client/human/HumanClientFSM.cpp
o01eg: Yep

network/Message.cpp
o01eg: Yes, I thought about it when fix errors with it.

o01eg: It looks like scoped enum requires explicit cast.

network/Message.cpp
o01eg: I took for example HOST_ID message.

network/Message.cpp
o01eg: I think it over-engineering to use boost::serialization for the single value.

o01eg:
Rebased and fixed.

network/Message.cpp
geoffthemedio: I don't think using boost::serialization will avoid the chance of a malformed message not being possible to convert into a non-string type, or of an exception being thrown. For cases like this, using it seems like massively over-complicating the implementation.

o01eg:
Cleaned up unused code.

LGM-Doyle:
@o01eg thanks for the changes and the explanation.

Vezzra:
> this is straightforward enough

True, but, you never know. Wouldn't be the first time that "oh, this is simple/straightforward enough, what can possibly go wrong?" backfires. So I'm leaning towards @geoffthemedio's assessment and mark this as post release.

Merging actually only has to be postponed until the release branch ist created, which I hope we can get done within this month, which means in two or three weeks. That shouldn't be too much of a problem.

o01eg:
Fixed conflicts and rebased.

LGM-Doyle:
@o01eg my changes to the context menu have broken your PR.  

Sorry.  

I didn't notice that you had just cleared up the conflicts.  

If you fix it one more time, I'll help get it merged quickly.

Also, send a message when you push a rebase that you want people to notice.  gitbub doesn't do that automatically.

o01eg:
I've just fixed it with latest changes in `master`.

LGM-Doyle (commented):
I've played a few hundred turns with this PR and observed no problems.

I've noted two small changes.  

After those are fixed we should merge this quickly because it touches on a lot of code and it will often need to be rebased as other code is merged.  Plus @o01eg is a newer contributor. 😄

client/AI/AIClientApp.cpp
```diff
@@ -199,119 +199,112 @@ void AIClientApp::HandleMessage(const Message& msg) {
...
-                DebugLogger() << "AIClientApp::HandleMessage : Received JOIN_GAME acknowledgement";
-                m_networking->SetPlayerID(msg.ReceivingPlayer());
-            } else {
-                ErrorLogger() << "AIClientApp::HandleMessage : Received erroneous JOIN_GAME acknowledgement when already in a game";
-            }
+        if (PlayerID() == Networking::INVALID_PLAYER_ID) {
+            DebugLogger() << "AIClientApp::HandleMessage : Received JOIN_GAME acknowledgement";
+            const std::string& text = msg.Text();
+            int player_id = boost::lexical_cast<int>(text);
+            m_networking->SetPlayerID(player_id);
```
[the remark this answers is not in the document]
o01eg: Fixed and rebased.

PR #1395, opened 2017-03-11
Fix bug #1369 OS indepence of systems_within_jumps().

### LINKED ISSUE #1369: Same game settings, different player starting locations depending on platform
### Environment

* **FreeOrion Version**: 4644c927ccf8a02f36e59326832d7cfc31e31595
* **Operating System**: Windows, OSX, Linux
* **Fetched as**
  * Compiled from source

### Description

When starting a default quickstart game, assignment of home systems to empires differs depending on platform. I've tested this on all three platforms: Windows (XP), OSX (10.10) and Linux (Kubuntu 16.04). For some strange reason I can't insert them here, so I opened a thread on the forum where I posted the screenshots:

http://www.freeorion.org/forum/viewtopic.php?f=28&t=10434

### Expected Result

The map as generated by the universe generation scripts should be identical across all platforms when using identical game settings.

### Steps to reproduce

Just start a default quickstart game on different platforms and compare the results.

LGM-Doyle:
This is another attempt to fix #1369.

This PR implements @Dilvish-fo 's suggestion based on @Vezzra's detective work to sort the results of `systems_within_jumps` if the results are not used as a set in an order dependent way.

I also changed the name to `systems_within_jumps_unordered` so it might be clearer that it is not sorted.

### DIFF
--- a/default/python/interface_mock/result/freeorion.py
+++ b/default/python/interface_mock/result/freeorion.py
@@ -4070,7 +4070,7 @@ def sys_set_star_type(number, star_type):
     return None
 
 
-def systems_within_jumps(number, item_list):
+def systems_within_jumps_unordered(number, item_list):
     """
     Return all systems within ''jumps'' of the systems with ids ''sys_ids''
 
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -72,7 +72,7 @@ def count_planets_in_systems(systems, planet_types_filter=HS_ACCEPTABLE_PLANET_T
 
 def calculate_home_system_merit(system):
     """Calculate the system's merit as the number of planets within HS_VICINTIY_RANGE."""
-    return count_planets_in_systems(fo.systems_within_jumps(HS_VICINITY_RANGE, [system]))
+    return count_planets_in_systems(fo.systems_within_jumps_unordered(HS_VICINITY_RANGE, [system]))
 
 
 def min_planets_in_vicinity_limit(num_systems):
@@ -161,7 +161,7 @@ class HomeSystemFinder(object):
                 candidate.append(member)
 
                 # remove all neighbors from the local pool
-                local_pool -= set(fo.systems_within_jumps(min_jumps, [member]))
+                local_pool -= set(fo.systems_within_jumps_unordered(min_jumps, [member]))
 
             # Count complete misses when number of candidates is not close to the target.
             if len(candidate) < (self.num_home_systems - MISS_THRESHOLD):
@@ -185,7 +185,7 @@ class HomeSystemFinder(object):
                 best_candidate = [s for (_, s) in merit_system]
 
                 # Quit sucessfully if the lowest merit system meets the minimum threshold
-                if merit >= min_planets_in_vicinity_limit(fo.systems_within_jumps(HS_VICINITY_RANGE, [system])):
+                if merit >= min_planets_in_vicinity_limit(fo.systems_within_jumps_unordered(HS_VICINITY_RANGE, [system])):
                     break
 
         return best_candidate
@@ -351,7 +351,7 @@ def compile_home_system_list(num_home_systems, systems, gsd):
     pool_matching_sys_and_planet_limit = []
     pool_matching_sys_limit = []
     for system in systems:
-        systems_in_vicinity = fo.systems_within_jumps(HS_VICINITY_RANGE, [system])
+        systems_in_vicinity = fo.systems_within_jumps_unordered(HS_VICINITY_RANGE, [system])
         if len(systems_in_vicinity) >= HS_MIN_SYSTEMS_IN_VICINITY:
             pool_matching_sys_limit.append(system)
             if count_planets_in_systems(systems_in_vicinity) >= min_planets_in_vicinity_limit(len(systems_in_vicinity)):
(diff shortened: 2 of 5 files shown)


geoffthemedio:
Why not make `std::unordered_set<int> WithinJumps` instead return `std::set<int>` so the system IDs are already sorted?

Dilvish-fo:
I just realized that I was misparsing Geoff's comment above, and that my comment here is mostly a repeat of his, oops.   But maybe it's a bit more explicit with the question, so I'll leave it....

quoting from #1369 :
> The fast method I have for providing read-only access for python to unorderd_set and unordered_map preserves the unorderedness from which those data structures derive their speedy access properties.

Is that extra bit of speed improvement worth the headache of remembering that this is a source of platform dependence?  I am concerned that simply adding "_unordered" to the function name may not really be an adequate notice to anyone not already involved in this conversation that the function is a source of platform dependence.  Even if it is noticeably helpful to use the unordered set while compiling the information, perhaps move the results to a normal ordered set before returning the result to python?

LGM-Doyle:
First this is not the fast method I mentioned.  It is your solution of sorting the returned value in python.

Second, we should establish the convention that sets, maps, dicts etc. that cross the python/C++ boundary always arrive in an undetermined order.  If a specific piece of code needs to, it sorts locally.

I think this is a performance decision with larger impact that this specific PR or section of code.  I will make the case below.

Sorting a set of size N has algorithmic complexity of O(NlogN).  This fact leads to the following consequences for the game:

1. If we choose to present data at the C++/python boundary as always sorted, then we either pay the O(NlogN) cost every time that we cross the boundary, or we keep the data in sorted order and pay a O(logN) cost with every insert/deletion/search.

2. Python's default data structures, dict and set are unsorted hash tables for precisely this performance reason.  In python there is already the expectation that to do a sorted operation on a set/dict  requires a call  to `sorted()`.  

3. The original C++ std library did not have has tables.  There is room for performance improvement in freeorion merely by replacing legacy sets and maps with unordered_sets and unordered_maps.  It is uncreative, but the benefits are real.

4. It is rare in terms of time and code location to require sorted data:
    - Sorting is usually done for presentation to humans.  It happens on a human timescales, no faster than 10's of ms.  On a machine timescale of ns, events on a ms timescale are rare.
    - The number of code locations that need sorted data is also rare. 

        Using `systems_within_jumps()`, as an concrete example it is called 7 times, but only needs sorted data in one location.

5. Sorting by object id has no in universe relevance.  Algorithms that are sorting for players, not devs, will sort alphabetically, by star type, etc.  Presorting by id is wasted computation.

    - In this specific instance it makes more sense for `add_planets_to_vicinity()` to sort by distance or direction from the homeworld, so that the extra planets are reasonably placed around the homeworld and not bunched due to the system id quirks from topology generation.  I should probably implement something like this method in any case,  since it is id independent.


The negative impact of sorting everywhere and all the time grows over the course of an individual game as N increases and will grow as we increase the complexity of the engine.  

The positive impact of using hash tables where they provide performance improvements and sorting as needed, will be a problematic performance decline that never materializes.

Adopting the convention that python/C++ should sort as needed any sets crossing the python/C++ boundary will allow us to gain the full benefit of processing with python hashes dict and set on one side and the C++ hashes unordered_set and unordered_map on the other.

geoffthemedio:
Can one not adopt such a convention - arbitrary order of information accessed across C++/Python interface - but actually supply sorted data in some cases? When it's known that the information will be sorted in Python anyway, at least most of the time, it could be faster to provide it pre-sorted, as at least some sorting algorithms have good best-case performance when given already-sorted input.

LGM-Doyle:
Yes.

Pre-processing data to minimize the amount of data transported across the interface, or to do the computation on the more efficient side of the interface for that calculation makes sense.

Dilvish-fo:
> First this is not the fast method I mentioned. It is your solution of sorting the returned value in python.

Yes, it's just that I saw both your explanation in the  other discussion of the underlying cause of platform dependence, and this related PR, at that same time, had to choose which one I put these comments in, and it seemed more suitable to put them here.

> we should establish the convention...

You make a pretty good case, and at least regarding the AI (@Morlic-fo, @Cjkjvfnby chime in if you have different thoughts) I'd be fine with adopting a rule that for any new python coding the python side will not assume it is getting ordered data from the C++ side, except perhaps for clearly-identified exceptions as Geoff alludes to.   I think we should hear from @Vezzra regarding the universe generation and events python code.   

And of course, we'd have to be careful to not just rush into changing existing parts of the interface; for each such change we would need to carefully review whether the existing python code had been relying on getting ordered data and then add a sort as needed, such as turned out to be the case here.   Such reliance may not always be easy to spot, since like here it may only be subtly relied upon for platform independence.

> merely by replacing legacy sets and maps with unordered_sets and unordered_maps

I have a nagging feeling that even just within the C++ side, there is more current reliance on ordered sets/maps than is fully compatible with the word "merely" here, but that's from vague memories a couple years old and I might be wrong.  Regardless, it does sound like a good change to make, but I think it will need careful attention.

geoffthemedio:
I didn't mean Python should ever assume it is getting ordered data. I meant that C++ could sometimes supply ordered data even if it doesn't claim to, because doing so might make the Python-side sorting faster.

LGM-Doyle:
I think that we are all on the same page and are now debating whether we are on the same word.

@Dilvish-fo when I said merely, I meant uncreative not without checking.    

Last fall, as an exercise, I replaced many of the maps and sets with unordered variants and then checked each location where they were used to see if there was order dependence.  I grouped the cases into 4 categories, that last category holds the troubling cases:
- the entire map/set is iterated over in an order independent manner. There is no local benefit/penalty to the change.  There is general benefit to not maintaining a sorted set.
- the map/set is used to insert/search for items, or to construct a set of unique items.  The change is entirely beneficial.
- the map/set is used to present ordered data to humans.  In most of cases it uses a UI element like ListWnd.  It is easy to transfer the sort to the ListWnd's built in sort.  Most of the remaining cases were debug messages in the log file, which can be sorted in place.
- the map/set is only partially iterated/consumed in an order dependent manner.  There were fewer of these case than I expected.  I don't remember any that didn't have a straightforward solution.

Vezzra:
> I think we should hear from @Vezzra regarding the universe generation and events python code.

As you've already pointed out, the order of items in Python dicts and sets is undefined and potentially implementation dependent, so whenever we need a predictable/platform independent order we need to sort them.

That means we have to go over the entire universe generation/turn event Python scripts anyway and make sure this gets done in all cases where such predictable/platform independent order has to be ensured, but currently isn't. Which usually is whereever we iterate over a dict or set.

So while we are at it, we can just check all sequences that get passed from C++ to Python as well. It's the better and more robust approach anyway not to let Python rely on any kind of order.

However, I don't want to get into that before the release. For 0.4.7 things are hopefully stable enough, but after the release I think we're good to go ahead with this.

PR #1396, opened 2017-03-11
Add max capacity effects for locked weapons

### LINKED ISSUE #1386: Dragon Tooth weapons have 0 strength.
### Environment
* **FreeOrion Version**: 0.4.6+
* **Operating System**: Fedora 25
* **Graphic card used**: N/A
* **Fetched as** 
  * Compiled from source cf67d52

### Description
When awarded a Dragon Tooth from an Ancient Ruins, the ship has a total weapon capacity of 0.

Seen with empires without Death Ray researched, possibly due to the change in default weapon effects.

dbenage-cx:
Fixes #1386 

Grants ~a reduced~ damage and shot value for weapons that are not yet researched by an empire, but owned by some means.

### DIFF
--- a/default/scripting/ship_parts/Weapons.focs.txt
+++ b/default/scripting/ship_parts/Weapons.focs.txt
@@ -10,7 +10,7 @@ Part
     tags = [ "PEDIA_PC_DIRECT_WEAPON" ]
     location = OwnedBy empire = Source.Owner
     effectsgroups =
-        [[WEAPON_BASE_UNOWNED_EFFECTS(SR_WEAPON_1_1)]]
+        [[WEAPON_BASE_DEFAULT_EFFECTS(SR_WEAPON_1_1)]]
     icon = "icons/ship_parts/mass-driver.png"
 
 Part
@@ -26,7 +26,7 @@ Part
     tags = [ "PEDIA_PC_DIRECT_WEAPON" ]
     location = OwnedBy empire = Source.Owner
     effectsgroups =
-        [[WEAPON_BASE_UNOWNED_EFFECTS(SR_WEAPON_0_1)]]
+        [[WEAPON_BASE_DEFAULT_EFFECTS(SR_WEAPON_0_1)]]
     icon = "icons/ship_parts/flak.png"
 
 Part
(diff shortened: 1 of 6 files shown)


Dilvish-fo:
I like the general structure of the fix but I am not sure about the reduction.  Am I reading this right that it would allow the weapon to operate, but with a 20% reduction to the base attack value?  Could you compare the numbers for the Dragon Tooth, which is our primary motivation for this?

I guess this would also apply to ships acquired via Psy Dom, and while the 20% reduction would seem reasonable in that case, I don't think it's really needed, and it really doesn't seem to be what we want for the Dragon Tooth; if we keep them in the same bucket via this mechanism it seems it should be at base value, not 80% of base.

dbenage-cx:
Ok, have removed the reduction.
That was primarily to address a suspicion that the issue affects gifted ships, but I've not tested to verify.

I suppose an alternate method might be to add a starting tech with such effects (possibly hidden as unresearchable).

default/scripting/ship_parts/Weapons.focs.txt
```diff
@@ -147,3 +157,18 @@ WEAPON_BASE_UNOWNED_EFFECTS
...
+                Not Unowned
+                Not OwnerHasShipPartAvailable name = "@1@"
+            ]
+            stackinggroup = "WEAPON_BASE_LOCKED_EFFECTS_@1@"
+            accountinglabel = "@1@"
+            effects = [
+                SetMaxCapacity partname = "@1@" value = Value + PartCapacity name = "@1@"
+                SetMaxSecondaryStat partname = "@1@" value = Value + PartSecondaryStat name = "@1@"
+            ]
+'''
```
Morlic-fo: The effect is now the same as the unowned macro. Seems like merging them into one macro (activation = Unowned or Not OwnerHasShipPartAvailable) would make more sense.

Dilvish-fo: > Seems like merging them into one macro (activation = Unowned or Not OwnerHasShipPartAvailable) would make more sense.

Although in the current uses it seems we are always pairing these together, it does not strike me that we would necessarily *always* want to do so; I'd be inclined to leave them as separate things as the PR currently has them.

dbenage-cx: I've merged to the one macro in the interest of moving forward.

Vezzra:
Everyone, especially @dbenage-cx: status report? Anything preventing this from being merged?

dbenage-cx:
Will merge tomorrow if no objections.

PR #1397, opened 2017-03-12
use dual stack sockets instead of ipv4-only ones

### LINKED ISSUE #1394: ipv6 support
freeoriond doesn't listen on ipv6 addresses, the game won't try to connect to an ipv6-only host either reporting a timeout. This kinda sucks because the ipv4 address pool is out and one may need to buy a vps to be able to host a game.

l29ah:
for multiplayer game hosting and client connections
fixes #1394
tested with both ipv4 and ipv6 clients

### DIFF
--- a/network/ServerNetworking.cpp
+++ b/network/ServerNetworking.cpp
@@ -485,7 +485,8 @@ void ServerNetworking::SetHostPlayerID(int host_player_id)
 { m_host_player_id = host_player_id; }
 
 void ServerNetworking::Init() {
-    tcp::endpoint endpoint(tcp::v4(), Networking::MessagePort());
+    // use a dual stack (ipv6 + ipv4) socket
+    tcp::endpoint endpoint(tcp::v6(), Networking::MessagePort());
 
     if (GetOptionsDB().Get<bool>("singleplayer")) {
         // when hosting a single player game only accept connections from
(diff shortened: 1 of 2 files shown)


network/ServerNetworking.cpp
```diff
@@ -485,7 +485,7 @@ void ServerNetworking::SetHostPlayerID(int host_player_id)
 { m_host_player_id = host_player_id; }
 
 void ServerNetworking::Init() {
-    tcp::endpoint endpoint(tcp::v4(), Networking::MessagePort());
+    tcp::endpoint endpoint(tcp::v6(), Networking::MessagePort());
```
geoffthemedio: `v6` here means `v6` or `v4` ?

l29ah: What.
It means "dual stack". It used to mean "ipv6 only" long time ago i guess but nobody cares enough to rename it properly in the socket interface headers.

geoffthemedio: In that case, a comment on the line, and any similar, indicating that this supports both v4 and v6 would be helpful.

geoffthemedio:
builds and runs OK for me with a local server.

Vezzra:
@geoffthemedio, I've set this to post release because I'm a bit wary of potential/unexpected side effects that might arise when tweaking critical components on such a low level, but if you think it's save enough to be merged we don't need to hold this up until after the release.

The change *is* very simple and straightforward after all.

geoffthemedio:
If OSX and Linux people say it works fine, and someone runs a non-local server connection test, it would be OK to merge, I think.

l29ah:
We've tested on windows and linux.

geoffthemedio:
> We've tested on windows and linux.

with an IPv6 address?

l29ah:
Both ipv4 and ipv6 on linux, ipv4 on windows as we don't have any ipv6-enabled users w/ windows.

Vezzra:
Works with OSX and Linux, although only on IPv4. Haven't been able to establish a connection with IPv6, but maybe IPv6 isn't set up correctly on my Mac.

Or maybe I'm doing something wrong when entering the IPv6 addresses. I've never bothered with IPv6 before, and don't really know what I'm doing...

geoffthemedio:
I'm able to connect to a server on localhost by pasting in the IPv6 address given by `ipconfig`. If it's not breaking anything with IPv4, then seems worth merging.

MatGB:
Observation: my ISP *still* doesn't support ipv6, so I couldn't test it regardless, worth checking with your ISP to see if they do just in case, but if it doesn't break anything we're good for now.

PR #1402, opened 2017-03-15
Categorize pedia articles for species

Ouaz:
- Classification by metabolism types (Lithic, Organic, Robotic, Phototrophic, Self-Sustaining)
- "Super Testers" still as a direct article (no category)

See [Forum Thread](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10343&p=86745) for the full list (search for PENDING PULL REQUEST in the opening post)


default/stringtables/en.txt
```diff
@@ -3939,6 +3939,9 @@ Building Type
 ENC_SPECIES
 Species
 
+ENC_SPECIES_DESC
+Each species is classified according its [[encyclopedia METABOLISM_TITLE]].
```
geoffthemedio: Species are classified according to their ...

default/stringtables/en.txt
```diff
@@ -6238,6 +6241,40 @@ Opinions of Empires:
...
+PHOTOTROPHIC_SPECIES_CLASS_DESC
+List of [[encyclopedia PHOTOTROPHIC_SPECIES_TITLE]] species:
+
+ROBOTIC_SPECIES_CLASS
+Robotic Species
+
+ROBOTIC_SPECIES_CLASS_DESC
+List of [[encyclopedia ROBOTIC_SPECIES_TITLE]] species:
+
+SELF_SUSTAINING_SPECIES_CLASS
```
geoffthemedio, line 6270: This shouldn't have "Species" in it as it looks oddly doubled as the article type in the pedia
cat

Ouaz: OK, done and rebased.

I removed "species" in each `****_SPECIES_CLASS` value, and adapted consequently the corresponding pedia links in each metabolism article.

geoffthemedio: See also other similar cases.

ship part

I guess that's a page about ship parts then...?

Ouaz: Oh, yes, ugly (it happens only in an article used as a subcategory "container")

To avoid this, the simplest fix is to leave blank the `short_description` value in `default/scripting/encyclopedia/ENC_SHIP_PART.focs.txt ` (as `short_description` is mandatory).  A little bit dirty, but it works. Is it ok for you?

```
Article
    name = "ENC_SHIP_PART"
    category = "ENC_SHIP_PART"
    short_description = "" (currently ENC_SHIP_PART)
    description = "ENC_SHIP_PART_DESC"
    icon = "icons/ship_parts/generic_part.png"
```

If it's ok, I can make a dedicated PR, because there is the same problem in:

```
default/scripting/encyclopedia/ENC_SPECIES.focs.txt
default/scripting/encyclopedia/ENC_SHIP_HULL.focs.txt
default/scripting/encyclopedia/ENC_SHIP_PART.focs.txt
default/scripting/encyclopedia/ENC_TECH.focs.txt
default/scripting/encyclopedia/FIELDS.focs.txt
```

EDIT: oh, and a quick and proper text needs to be written for `ENC_SHIP_PART_DESC`. Something like "Ship parts can be unlocked by research and  used to design ships. Each ship part falls into one of the following categories:"

geoffthemedio: as long as that doesn't generate a bunch of errors about not finding the string ""

Ouaz: Actually, it seems the blank value is rendered in-game as a single space (the undertitle line is a little bit shifted to the right -from one space, almost unnoticeable).

No error in the log, AFAIK.

undertitle_shift

geoffthemedio: I believe the space comes from this string: https://github.com/freeorion/freeorion/blob/master/default/stringtables/en.txt#L4110

PR #1415, opened 2017-03-18
Add description for Ship Hull / Ship Part pedia articles

Ouaz:
As usual, feel free to comment and make suggestions if my written english is too bad!


geoffthemedio:
could use some minor tweaks but is OK

PR #1416, opened 2017-03-18
Build freeorioncommon as shared library on Windows with CMake

adrianbroher:
This PR enables building `freeorioncommon` as a shared library on windows within the CMake build system.  This is a work in progress and unfortunately breaks both the current MSVC builds.

If someone wants to test this out please keep the following things in mind:

* You need to pass the toolset (at least v140_xp or a later one) with the cmake `-T` parameter to support WindowsXP.  When not passing the -T parameter the build will still work but the resulting executable won't be executable on WindowsXP obviously.
* Use a different build directory than the source directory.  The build will probably work if this advice is ignored but I didn't test this variant because it is considered bad style.
* The build copies the used library DLLs into the build dir and links the default directory automatically to allow an arbitrary build layout with different build configurations.  However the build expects the libraries and the python.exe to be located inside a `bin` which is at the same level as the SDK `include` and `lib` directories.  If you don't want to copy around files you can establish the same layout by downloading the most recent SDK and **not** executing the included `bootstrap.bat`.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -106,33 +106,32 @@ maintained Visual Studio Project is used for building.
 Step by step procedure:
 
  * Enter the _source directory_.
+ * Create a `build` directory, which will contain all compile FreeOrion
+   build artifacs.
+ * Change into the `build` directory on the command line.
+ * Execute cmake to generate the native build system:
+
+     ```
+     cmake -G "GENERATOR" ..
+     ```
+
+   Where `GENERATOR` should be:
+
+   * on Windows: "Visual Studio 2015"
+   * on Mac OS X: "Xcode"
+   * on Linux and other other Operating Systems: "Unix Makefiles"
+
  * On Windows:
-   * Open `msvc2015\FreeOrion.sln` with Visual Studio.
+   * Open `FreeOrion.sln` with Visual Studio.
    * Compile the whole project by selecting the `Build` -> `Build Solution`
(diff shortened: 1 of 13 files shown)


geoffthemedio:
Does this effectively just merge parse and common and then make the combination a DLL instead of static library?

adrianbroher:
> Does this effectively just merge parse and common and then make the combination a DLL instead of static library?

Somewhat.  `common` and `parse` stay different targets (or projects in VS) with different build configurations, but the build artifact is a single dll.

adrianbroher:
@LGM-Doyle the `Pending::Pending` implementation defines a move constructor and a move assign operator.  Is there a particular reason why you did not implement the corresponding copy operations?  This causes the build to fail for me with the MSVC compiler.

LGM-Doyle:
std::future is not copyable.

I could work on a copyable version next week, if it is preventing this from landing.

o01eg:
Hi, could you try to rebase it over current master? At least I took one commit to enable CI test.

adrianbroher:
Stale branch, superceded by adrianbroher/win-cmake

Vezzra:
Superseded by #2697.

adrianbroher:
@Vezzra can you delete this branch?

Vezzra:
Well, @geoffthemedio beat me to it...

PR #1420, opened 2017-03-19
Remove boost::chrono from the build files.

LGM-Doyle:


geoffthemedio:
@adrianbroher ?

adrianbroher:
Builds fine on all systems and runs on Linux so I don't expect any issues on the other two.

PR #1421, opened 2017-03-19
Reduce client networking build dependencies.

LGM-Doyle:
This reduces the dependency of compilation units downstream of ClientNetworking.h, improving incremental build times by a factor of 4 when ClientNetworking is changed.  Sadly, it make no difference in the total clean build time.

It moves several dependencies from the header into the cpp file.

It changes function signatures so that forward declarations are as good as full definitions.

It reduces the number of files that include ClientNetworking.h.

I wish I had done this before I started hacking on ClientNetworking.

### DIFF
--- a/network/ClientNetworking.cpp
+++ b/network/ClientNetworking.cpp
@@ -2,6 +2,21 @@
 
 #include "ClientNetworking.h"
 
+#include "Message.h"
+#include "MessageQueue.h"
+
+// boost::asio pulls in windows.h which in turn defines the macros Message,
+// MessageBox, min and max. Disabling the generation of the min and max macros
+// and undefining those should avoid name collisions with std c++ library and
+// FreeOrion function names.
+#define NOMINMAX
+#include <boost/asio.hpp>
+#include <boost/asio/high_resolution_timer.hpp>
+#ifdef FREEORION_WIN32
+#   undef Message
+#   undef MessageBox
+#endif
+
 #include "Networking.h"
 #include "../util/Logger.h"
 #include "../util/MultiplayerCommon.h"
@@ -119,10 +134,125 @@ namespace {
     };
 }
 
+
+
+class ClientNetworking::Impl {
+public:
+    /** The type of list returned by a call to DiscoverLANServers(). */
+    typedef std::vector<std::pair<boost::asio::ip::address, std::string> >  ServerList;
+
+    /** \name Structors */ //@{
+    Impl();
+    //@}
+
+    /** \name Accessors */ //@{
+    /** Returns true iff the client is full duplex connected to the server. */
+    bool IsConnected() const;
+
+    /** Returns true iff the client is connected to receive from the server. */
+    bool IsRxConnected() const;
+
+    /** Returns true iff the client is connected to send to the server. */
+    bool IsTxConnected() const;
+
+    /** Returns true iff there is at least one incoming message available. */
+    bool MessageAvailable() const;
+
+    /** Returns the ID of the player on this client. */
+    int PlayerID() const;
+
+    /** Returns the ID of the host player, or INVALID_PLAYER_ID if there is no host player. */
+    int HostPlayerID() const;
+
+    /** Returns whether the indicated player ID is the host. */
+    bool PlayerIsHost(int player_id) const;
+    //@}
+
+    /** \name Mutators */ //@{
+    /** Returns a list of the addresses and names of all servers on the Local
+        Area Network. */
+    ServerList DiscoverLANServers();
+
+    /** Connects to the server at \a ip_address.  On failure, repeated
+        attempts will be made until \a timeout seconds has elapsed. */
+    bool ConnectToServer(const ClientNetworking* const self,
+                         const std::string& ip_address,
+                         std::chrono::milliseconds timeout = std::chrono::seconds(10));
+
+    /** Connects to the server on the client's host.  On failure, repeated
+        attempts will be made until \a timeout seconds has elapsed. */
+    bool ConnectToLocalHostServer(const ClientNetworking* const self,
+                                  std::chrono::milliseconds timeout =
+                                  std::chrono::seconds(10));
+
+    /** Sends \a message to the server.  This function actually just enqueues
+        the message for sending and returns immediately. */
+    void SendMessage(const Message& message);
+
+    /** Gets the next incoming message from the server, places it into \a
+        message, and removes it from the incoming message queue.  The function
+        assumes that there is at least one message in the incoming queue.
+        Users must call MessageAvailable() first to make sure this is the
+        case. */
+    void GetMessage(Message& message);
+
+    /** Sends \a message to the server, then blocks until it sees the first
+        synchronous response from the server. */
+    void SendSynchronousMessage(const Message& message, Message& response_message);
+
+    /** Disconnects the client from the server. */
+    void DisconnectFromServer();
+
+    /** Sets player ID for this client. */
+    void SetPlayerID(int player_id);
+
+    /** Sets Host player ID. */
+    void SetHostPlayerID(int host_player_id);
+    //@}
+
+private:
+    void HandleException(const boost::system::system_error& error);
+    void HandleConnection(boost::asio::ip::tcp::resolver::iterator* it,
+                          const boost::system::error_code& error);
+    void CancelRetries();
+    void NetworkingThread(const std::shared_ptr<const ClientNetworking>& self);
+    void HandleMessageBodyRead(const std::shared_ptr<const ClientNetworking>& keep_alive,
+                               boost::system::error_code error, std::size_t bytes_transferred);
+    void HandleMessageHeaderRead(const std::shared_ptr<const ClientNetworking>& keep_alive,
+                                 boost::system::error_code error, std::size_t bytes_transferred);
+    void AsyncReadMessage(const std::shared_ptr<const ClientNetworking>& keep_alive);
+    void HandleMessageWrite(        boost::system::error_code error, std::size_t bytes_transferred);
+    void AsyncWriteMessage();
+    void SendMessageImpl(Message message);
+    void DisconnectFromServerImpl();
+
+    int                             m_player_id;
+    int                             m_host_player_id;
+
+    boost::asio::io_service         m_io_service;
+    boost::asio::ip::tcp::socket    m_socket;
+
+    // m_mutex guards m_incoming_message, m_rx_connected and m_tx_connected which are written by
+    // the networking thread and read by the main thread to check incoming messages and connection
+    // status. As those read and write operations are not atomic, shared access has to be
+    // protected to prevent unpredictable results.
+    mutable boost::mutex            m_mutex;
+
+    MessageQueue                    m_incoming_messages; // accessed from multiple threads, but its interface is threadsafe
+    std::list<Message>              m_outgoing_messages;
+    bool                            m_rx_connected;      // accessed from multiple threads
+    bool                            m_tx_connected;      // accessed from multiple threads
+
+    Message::HeaderBuffer           m_incoming_header;
+    Message                         m_incoming_message;
+    Message::HeaderBuffer           m_outgoing_header;
+};
+
+
 ////////////////////////////////////////////////
-// ClientNetworking
+// ClientNetworking Impl
 ////////////////////////////////////////////////
-ClientNetworking::ClientNetworking() :
+ClientNetworking::Impl::Impl() :
     m_player_id(Networking::INVALID_PLAYER_ID),
     m_host_player_id(Networking::INVALID_PLAYER_ID),
     m_io_service(),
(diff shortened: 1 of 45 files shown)


adrianbroher:
@LGM-Doyle I'm curious: did you recieve a failing message from AppVeyor via eMail?

network/ClientNetworking.cpp
```diff
@@ -119,10 +134,125 @@ namespace {
...
+    void HandleConnection(boost::asio::ip::tcp::resolver::iterator* it,
+                          const boost::system::error_code& error);
+    void CancelRetries();
+    void NetworkingThread(const std::shared_ptr<const ClientNetworking> self);
+    void HandleMessageBodyRead(const std::shared_ptr<const ClientNetworking>& keep_alive,
+                               boost::system::error_code error, std::size_t bytes_transferred);
+    void HandleMessageHeaderRead(const std::shared_ptr<const ClientNetworking>& keep_alive,
+                                 boost::system::error_code error, std::size_t bytes_transferred);
+    void AsyncReadMessage(const std::shared_ptr<const ClientNetworking>& keep_alive);
+    void HandleMessageWrite(        boost::system::error_code error, std::size_t bytes_transferred);
```
geoffthemedio: funky indenting

network/ClientNetworking.cpp
```diff
@@ -132,47 +262,53 @@ ClientNetworking::ClientNetworking() :
...
 
-ClientNetworking::ServerList ClientNetworking::DiscoverLANServers() {
+ClientNetworking::ServerNames ClientNetworking::Impl::DiscoverLANServerNames() {
     if (!IsConnected())
-        return ServerList();
+        return ServerNames();
     ServerDiscoverer discoverer(m_io_service);
     discoverer.DiscoverServers();
-    return discoverer.Servers();
+    auto servers = discoverer.Servers();
```
geoffthemedio: necessity of local copy instead of putting call in subsequent loop directly?

network/ClientNetworking.cpp
```diff
@@ -261,14 +396,15 @@ bool ClientNetworking::ConnectToServer(
...
+bool ClientNetworking::Impl::ConnectToLocalHostServer(
+    const ClientNetworking* const self,
+    const std::chrono::milliseconds& timeout/* = std::chrono::seconds(10)*/)
 {
     bool retval = false;
 #if FREEORION_WIN32
     try {
 #endif
-        retval = ConnectToServer("127.0.0.1", timeout);
+        retval = ConnectToServer(self, "127.0.0.1", timeout);
```
geoffthemedio: self vs localhost ?

network/ClientNetworking.cpp
```diff
@@ -319,7 +455,7 @@ void ClientNetworking::GetMessage(Message& message) {
                       << message;
 }
 
-void ClientNetworking::SendSynchronousMessage(Message message, Message& response_message) {
+void ClientNetworking::Impl::SendSynchronousMessage(const Message& message, Message& response_message) {
     if (TRACE_EXECUTION)
         DebugLogger() << "ClientNetworking::SendSynchronousMessage : sending message "
                                << message;
```
geoffthemedio: not this patch, but indenting odd

network/ClientNetworking.cpp
```diff
@@ -331,7 +467,7 @@ void ClientNetworking::SendSynchronousMessage(Message message, Message& response
                                << "response message " << response_message;
 }
 
-void ClientNetworking::HandleConnection(tcp::resolver::iterator* it,
+void ClientNetworking::Impl::HandleConnection(tcp::resolver::iterator* it,
                                         const boost::system::error_code& error)
```
geoffthemedio: indenting

network/ClientNetworking.cpp
```diff
@@ -387,7 +523,7 @@ void ClientNetworking::NetworkingThread(std::shared_ptr<ClientNetworking>& self)
         DebugLogger() << "ClientNetworking::NetworkingThread() : Networking thread terminated.";
 }
 
-void ClientNetworking::HandleMessageBodyRead(const std::shared_ptr<ClientNetworking>& keep_alive,
+void ClientNetworking::Impl::HandleMessageBodyRead(const std::shared_ptr<const ClientNetworking>& keep_alive,
                                              boost::system::error_code error, std::size_t bytes_transferred)
```
geoffthemedio: ...

network/ClientNetworking.cpp
```diff
@@ -400,7 +536,7 @@ void ClientNetworking::HandleMessageBodyRead(const std::shared_ptr<ClientNetwork
     }
 }
 
-void ClientNetworking::HandleMessageHeaderRead(const std::shared_ptr<ClientNetworking>& keep_alive,
+void ClientNetworking::Impl::HandleMessageHeaderRead(const std::shared_ptr<const ClientNetworking>& keep_alive,
                                                boost::system::error_code error, std::size_t bytes_transferred)
```
geoffthemedio: ... (not noting any more... you get the point... indenting)

network/ClientNetworking.h
```diff
@@ -85,21 +77,21 @@ class ClientNetworking : public std::enable_shared_from_this<ClientNetworking> {
...
     /** Connects to the server on the client's host.  On failure, repeated
         attempts will be made until \a timeout seconds has elapsed. */
-    bool ConnectToLocalHostServer(std::chrono::milliseconds timeout =
+    bool ConnectToLocalHostServer(const std::chrono::milliseconds& timeout =
                                   std::chrono::seconds(10));
 
     /** Sends \a message to the server.  This function actually just enqueues
-        the message for sending and returns immediately. */
-    void SendMessage(Message message);
+        the message for sending a nd returns immediately. */
```
geoffthemedio: ?

network/ClientNetworking.cpp
LGM-Doyle: Fixed.

network/ClientNetworking.cpp
LGM-Doyle: No need.  Fixed.

network/ClientNetworking.cpp
LGM-Doyle: fixed.

network/ClientNetworking.cpp
LGM-Doyle: fixed

network/ClientNetworking.cpp
LGM-Doyle: fixed.

network/ClientNetworking.cpp
LGM-Doyle: fixed

network/ClientNetworking.h
LGM-Doyle: fixed.

LGM-Doyle (commented):
All fixed.

@adrianbroher I did not receive an email from AppVeyor.

Also the AppVeyor website does not work on my browser, firefox.  I've enabled all scripting for it, but it just shows three pulsing blue dots.

Travis works correctly.

adrianbroher:
> @adrianbroher I did not receive an email from AppVeyor.

Hm, maybe it's because of the noreply email of GitHub.

> Also the AppVeyor website does not work on my browser, firefox. I've enabled all scripting for it, but it just shows three pulsing blue dots.

Developers lacked permissions until now.  This should be fixed.

network/ClientNetworking.h
```diff
@@ -53,10 +44,11 @@
 class ClientNetworking : public std::enable_shared_from_this<ClientNetworking> {
 public:
     /** The type of list returned by a call to DiscoverLANServers(). */
-    typedef std::vector<std::pair<boost::asio::ip::address, std::string> >  ServerList;
+    using ServerNames =  std::vector<std::string>;
```
geoffthemedio: using vs. typedef in this context?

LGM-Doyle: In most cases, and this specific case `using` and `typedef` are exactly the same.

tldr; We should replace all of our `typedefs` with `using` declarations.

`using` is more general than `typedef`.  

There is never case where employing `using` produces less functional or more verbose code than `typedef`.  

`using` can be templated and the compiler then knows that the templated thing is a type.  This results in less verbose code and better compiler errors than the typedef alternative.

For example create two templated types, one with `using` and one with `typedef`:
````
template<typename T> using UsingMappedThingType = std::map<T, Thing<T>>;
````
something similar with `typedef` involves nesting in a templated structure.
````
template<typename T> struct StructToHoldTypedef {
    typedef std::map<T, Thing<T>> TypdefMappedThingType;
};
```` 
When further utilizing those types in other templated things, the compiled knows that `UsingMapThingType<T>` is a type, but `StructToHoldTypedef<T>::TypedefMappedThingType` may be a type, member data or a member function, the compile does not know.    

In pre-C++11 code, you can add `typename` decorators to force the compiler to consider this a type, `typename StructToHoldTypedef<T>::TypedefMappedThingType`.  Sometimes adding/removing `typename` causes the compile to fail, because the member you used `typename` for is not actually a type.  In C++11 code, `UsingMappedThingType<T>` in unambiguously a type.

Changing all of the `typedef` statements to `using` declarations would make the code less verbose and more consistent. Keeping `typedef` in some places and `using` in locations where it avoids the use of `typename` adds unnecessary cognitive load.

**Comment by Vezzra** [comment]
Everyone, status report? Anything preventing this from being merged?

**Comment by geoffthemedio** [comment]
````
4>..\..\network\ClientNetworking.cpp(707): error C2039: 'SendMessageW': is not a member of 'ClientNetworking'
4>         c:\users\geoff\desktop\freeorionsdk5\freeorion\network\ClientNetworking.h(44): note: see declaration of 'ClientNetworking'
4>     1>
4>..\..\network\ClientNetworking.cpp(708): error C2065: 'm_impl': undeclared identifier
4>     1>
4>..\..\network\ClientNetworking.cpp(708): error C2227: left of '->SendMessageW' must point to class/struct/union/generic type
4>         ..\..\network\ClientNetworking.cpp(708): note: type is 'unknown-type'
4>     1>
4>..\..\network\ClientNetworking.cpp(710): error C2039: 'GetMessageW': is not a member of 'ClientNetworking'
4>         c:\users\geoff\desktop\freeorionsdk5\freeorion\network\ClientNetworking.h(44): note: see declaration of 'ClientNetworking'
4>     1>
4>..\..\network\ClientNetworking.cpp(711): error C2065: 'm_impl': undeclared identifier
4>     1>
4>..\..\network\ClientNetworking.cpp(711): error C2227: left of '->GetMessageW' must point to class/struct/union/generic type
4>         ..\..\network\ClientNetworking.cpp(711): note: type is 'unknown-type'
````

Same / similar appears in the Appveyor report: https://ci.appveyor.com/project/freeorion/freeorion/build/ci-1088

**Comment by adrianbroher** [comment]
> Anything preventing this from being merged?

Uhm, the Windows CI build is failing as you can see?

**Comment by Vezzra** [comment]
> Uhm, the Windows CI build is failing as you can see?

Oh, sorry... I actually *did* notice that, and then looked at the build log, but not thoroughly - just saw that it ended with:
```
  EmpireWrapper.cpp
  EnumWrapper.cpp
  LoggingWrapper.cpp
  UniverseWrapper.cpp
  DependencyVersions.cpp
Command exited with code 1
```
which looked like the build crashed (maybe because of overtaxing the limited resources of the build environment AppVeyor spawned, I get such crashes on my Mac and Windows systems now and then), as there are no summary lines (detailing how much errors, warnings etc. occured), which I expected though.

Should have taken the two secs to look at the messages a few lines above, then I would have noticed the errors... my bad 😳

geoffthemedio:
To be fair, there are a lot of false positive (false negative?) build failures in the automated test build results.

adrianbroher:
> I actually did notice that, and then looked at the build log, but not thoroughly - just saw that it ended with:

AppVeyor does extract warnings/errors and displays in the `messages` tab right beside the `console` tab so you can get a quick overview what is the exact cause of the issue.  Would be a nice feature for Travis, but as far as I know it doesn't exist.

> which looked like the build crashed (maybe because of overtaxing the limited resources of the build environment AppVeyor spawned, I get such crashes on my Mac and Windows systems now and then)

TravisCI does report build failure (the build broke, marked with a red X) and build errors (the build environment could not be run successfully, e.g compiler crashes, took to long to build, marked with a red !).  I don't know how AppVeyor reports this because I can't remember any breakage so far.

> To be fair, there are a lot of false positive (false negative?) build failures in the automated test build results.

Howso (especially the false positive part)?  I consider the build to be very reliable so far.

geoffthemedio:
> Howso (especially the false positive part)? I consider the build to be very reliable so far.

I (feel like I) often see / get an error about the clang build due to there not being enough time to complete it.

Or here (not sure if it's a legit problem in this specific case):
https://travis-ci.org/freeorion/freeorion/builds/214677710
It can't find `Universe::serialize` apparently?

adrianbroher:
> I (feel like I) often see / get an error about the clang build due to there not being enough time to complete it.

I see your point but I don't agree.  Maybe I can create some statistics about build success rates for the CI in the future.

Vezzra:
Also, that's the Travis CI for OSX. IIRC that hasn't been an issue with AppVeyor, and wrt to the Travis CI build for OSX I *think* @adrianbroher has already done something to get the build times down (and thereby rid of the failed builds due to timeout).

adrianbroher:
What's the whole point of this PR when you add the asio header again to `network/ClientNetworking.h`?  This completely fails the PRs point of isolating the asio dependency inside the PIMPLE implementation.

LGM-Doyle:
@adrianbroher, you are correct for the Windows implementation.  All other OSes work as described above.

Here is the problem with Windows as inferred from the AppVeyor logs and the comment that was already in the code.  (If a real Windows programmer can clarify, please do).

`boost::asio` includes `windows.h`, which defines macros for `min`, `max`, `Message`, `MessageBox` and `PlaySound`.  These conflict with the same named functions in `std` and FreeOrion.  In order to be undefined, they first need to be included.

There is undoubtedly a better way, but I am not compiling on Windows and can only get the compile logs through AppVeyor, which with queue wait times is often hours of round trip time between compiles.  I also have no way to measure the incremental build times, since AppVeyor always builds from scratch. This PR works as described for non-Windows OSes and I am not in a position to efficiently improve the Windows compile times.

adrianbroher:
> @adrianbroher, you are correct for the Windows implementation. All other OSes work as described above.

So it's only 2/3 of the way done or 1/3 still missing, however you prefer to see this.

> If a real Windows programmer can clarify, please do

Yeah here, I'm the author of those lines. What do you want to know? Wait, I just cherry pick the good parts and fix it myself tomorrow.

LGM-Doyle:
@adrianbroher, I am not requesting that anyone do work on my behalf, but thanks for the offer.  I was asking for clarification in case I had inferred a conflicting windows macro when it is a different problem. 

Because I am using AppVeyor as a compiler to get the error log I needed to work back from the broken build to a working build, where I understood what the problem was.

I will try renaming `Message`, `MessageBox` and `PlaySound` so that there are no future conflicts with the Windows `windows.h` macros.

We can't rename `std::min` and `std::max`, so I will find another workaround.

I've marked the PR WIP, and will leave it that way until Ii think that the Windows portion is done.

Vezzra:
If fixing these problems takes longer, can we reassign to the "optional for 0.4.7" milestone? This PR isn't really *critical* for the relase, after all...

LGM-Doyle:
I meant to do that when I changed it to WIP.

o01eg:
What if isolate the code uses "windows.h"?

LGM-Doyle:
@o01eg that is what I have done in the current PR.  The objection is that it makes the incremental build performance of the Windows version of `ClientNetworking.h` worse than the other OSes.

The problem is that the `windows.h` is being implicitly included in headers from `boost` libraries that we do not control.

adrianbroher:
Merged as f731059e293e1d960490dcac1759a515be50318c, leaving out most recent commits.

LGM-Doyle:
Thanks.

PR #1423, opened 2017-03-20
Improve app termination save pending

### LINKED ISSUE #971: Save game operation fails silently if user resigns and quits the game quickly enough

LGM-Doyle:
This improves/fixes [Issue #971 Save game operation fails silently if user resigns and quits the game quickly enough](https://github.com/freeorion/freeorion/issues/971).

The problem as described in the issue is that application can exit while the server is either collecting or writing the save game.

This PR solves this by popping up a dialog when there is a pending save and the user tries to return to the intro menu or completely exit the game.  The dialog says "A save game is in progress.  Click Stop Waiting to abandon the save."

If the player waits the application will quit/exit when the save completes.

If the player clicks "Stop Waiting" then the app will continue with the exit, without waiting for the save to complete.

A deficiency is that when the player elects to stop waiting the exit may wait  for 5 seconds and then kill the server.  Neither the server process or the AI processes are interruptible.  However, the player will have been informed and chosen to abandon the save and kill the server. 

I think this fixes the bug.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -1130,13 +1130,13 @@ namespace {
     // The dialog automatically closes if the save completes while the user is waiting
     class SaveGamePendingDialog : public GG::ThreeButtonDlg {
         public:
-        SaveGamePendingDialog(boost::signals2::signal<void()>& save_completed_signal) :
+        SaveGamePendingDialog(bool reset, boost::signals2::signal<void()>& save_completed_signal) :
             GG::ThreeButtonDlg(
                 GG::X(320), GG::Y(200), UserString("SAVE_GAME_IN_PROGRESS"),
                 ClientUI::GetFont(ClientUI::Pts()+2),
                 ClientUI::WndColor(), ClientUI::WndOuterBorderColor(),
                 ClientUI::CtrlColor(), ClientUI::TextColor(), 1,
-                UserString("ABORT_SAVE_AND_QUIT"))
+                (reset ? UserString("ABORT_SAVE_AND_RESET") : UserString("ABORT_SAVE_AND_EXIT")))
         { GG::Connect(save_completed_signal, &SaveGamePendingDialog::SaveCompletedHandler, this); }
 
         void SaveCompletedHandler() {
@@ -1159,7 +1159,7 @@ void HumanClientApp::ResetOrExitApp(bool reset) {
 
     if (m_save_game_in_progress) {
         DebugLogger() << "save game in progress. Checking with player.";
-        auto dlg = SaveGamePendingDialog(this->SaveGameCompletedSignal);
+        auto dlg = SaveGamePendingDialog(reset, this->SaveGameCompletedSignal);
         dlg.Run();
     }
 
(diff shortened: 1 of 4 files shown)


client/human/HumanClientApp.cpp
```diff
@@ -1117,18 +1125,46 @@ void HumanClientApp::ResetClientData() {
...
+            GG::ThreeButtonDlg(
+                GG::X(320), GG::Y(200), UserString("SAVE_GAME_IN_PROGRESS"),
+                ClientUI::GetFont(ClientUI::Pts()+2),
+                ClientUI::WndColor(), ClientUI::WndOuterBorderColor(),
+                ClientUI::CtrlColor(), ClientUI::TextColor(), 1,
+                UserString("STOP_WAITING"))
+        { GG::Connect(save_completed_signal, &SaveGamePendingDialog::SaveCompletedHandler, this); }
+
+        void SaveCompletedHandler()
+        {
```
geoffthemedio: { to previous line

client/human/HumanClientApp.cpp
```diff
@@ -1117,18 +1125,46 @@ void HumanClientApp::ResetClientData() {
...
+            DebugLogger() << "SaveGamePendingDialog:: save game completed handled.";
```
geoffthemedio: double ::  ?

default/stringtables/en.txt
```diff
@@ -899,6 +899,12 @@ Error when reading save file.
 UNABLE_TO_SAVE_NOW_TRY_AGAIN
 Unable to save during AI processing. Try again when AIs have finished turns.
 
+SAVE_GAME_IN_PROGRESS
+A save game is in progress.  Click Stop Waiting to abandon the save.
+
+STOP_WAITING
+Stop waiting.
```
geoffthemedio: Maybe something stronger-sounding? Like `Abort Save and Quit Immediately`

geoffthemedio: `Abort Save and Quit` is ambiguous. Will the quit be aborted?

Vezzra: So maybe "Abort Save, and Quit", or "Abort Save, then Quit"?

LGM-Doyle: I've split the messages in separate ones for return to main menu and exiting. 
`Return to Main Menu without saving.`
and
`Exit FreeOrion without saving.`

This is the same phrasing as the `RETURN_TO_INTRO` message.

PR #1425, opened 2017-03-21
Fix issue 1422 human client FSM transition order issues.

### LINKED ISSUE #1422: Human Client Access Violation Upon Fatal Error From Server
Was playing around with the save format, and made a save that the UI accepts, but which fails to load properly on the server. When the client receives an error message from the server about the fatal error, there is an access violation in the client, at least in Windows. The problem happens around here:

https://github.com/freeorion/freeorion/blob/master/client/human/HumanClientFSM.cpp#L550

The comment immediately above this code suggests the possible cause of the problem.

I'm not sure how the code within `Client().ResetToIntro();`, which itself posts some events and does some transitions to kill the sever and similar things, will interact with the explicit transition here, but at least the explicit transition probably needs to happen before the message box.

LGM-Doyle:
I think that this PR fixes #1422.

The problem is that moving `discard_event()` to after the transition within `ResetToIntro()` means that the discard event null transition occurs from a state that has already been exited.

The PR expands the comments clarifying the boost state machine behavior everywhere in the code. 

Hopefully this corrects, my previous unclear explanation of what was happening in this section of code.

### DIFF
--- a/client/human/HumanClientFSM.cpp
+++ b/client/human/HumanClientFSM.cpp
@@ -542,12 +542,17 @@ boost::statechart::result PlayingGame::react(const Error& msg) {
     ErrorLogger() << "PlayingGame::react(const Error& msg) error: "
                   << problem << "\nProblem is" << (fatal ? "fatal" : "non-fatal");
 
+    //Note: Any transit<> transition must occur before the MessageBox().
+    // MessageBox blocks and can allow other events to transit<> to a new state
+    // which makes this transit fatal.
+
+    auto retval = discard_event();
     if (fatal)
-        Client().ResetToIntro();    // initiates state transition, so nothing extra required in this function
+        Client().ResetToIntro();
 
     ClientUI::MessageBox(UserString(problem), fatal);
 
-    return discard_event();
+    return retval;
 }
 
 boost::statechart::result PlayingGame::react(const TurnProgress& msg) {
--- a/client/human/HumanClientFSM.cpp
+++ b/client/human/HumanClientFSM.cpp
@@ -17,6 +17,53 @@
 #include <boost/format.hpp>
 #include <thread>
 
+/** \page statechart_notes Notes on Boost Statechart transitions
+
+    \section reaction_transition_note Transistion Ordering and Reaction return values.
+
+    Transitions, transit<> or discard_event(), from states must occur from the current state.
+    Transitioning from a state a second time causes indeterminate behavior and may crash.
+    Transitioning from a state that is not the current state causes indeterminate behavior and may crash.
+
+    Reactions handle events in states.  They return a statechart::result from a transit<> or discard_event().
+
+    In a reaction the state is transitioned when transit<> or discard_event() is called, not when
+    the reaction returns.  If a function is called within the reaction that may internally cause a
+    transition, then the transistion out of the reaction's initial state must happen before the
+    function is called.  Save the return value and return it at the end of the reaction.
+
+    This code is correct:
+    \code
+        boost::statechart::result SomeState::react(const SomeEvent& a) {
+
+            // Transition before calling other code that may also transition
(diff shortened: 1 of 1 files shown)


client/human/HumanClientFSM.cpp
```diff
@@ -114,9 +114,13 @@ boost::statechart::result WaitingForSPHostAck::react(const HostSPGame& msg) {
 
 boost::statechart::result WaitingForSPHostAck::react(const Disconnection& d) {
     if (TRACE_EXECUTION) DebugLogger() << "(HumanClientFSM) PlayingGame.Disconnection";
-    //Note: Any transit<> transition must occur before the MessageBox().
-    // MessageBox blocks and can allow other events to transit<> to a new state
-    // which makes this transit fatal.
+    // Note: All transitions, transit<> or disable_event(), must occur in the order
```
geoffthemedio: disable or discard?

geoffthemedio: "must occur in the order that they cause changes to state machine" is unclear.

client/human/HumanClientFSM.cpp
```diff
@@ -114,9 +114,13 @@ boost::statechart::result WaitingForSPHostAck::react(const HostSPGame& msg) {
...
+    // that they cause changes to the state machine.
+    // - MessageBox blocks the UI, but continues handling events so other events
```
geoffthemedio: MessageBox handles events? Or the FSM continues handling events while MessageBox has the UI blocked? Why does a UI block matter to the FSM?

client/human/HumanClientFSM.cpp
geoffthemedio: the phrase I quoted is still there and still doesn't make sense (or appears tautological)

client/human/HumanClientFSM.cpp
geoffthemedio: Then say the MessageBox starts handling events, rather than continues, since it wasn't handling any events before it blocked the UI.

Vezzra:
Fixed the issue on OSX. No more in-game windows left over when returning to main menu.

client/human/HumanClientFSM.cpp
LGM-Doyle: I laughed when I read "appears tautological".

I expanded the comment so that it hopefully is clear why the incorrect code ordering crashes. 

The comment was long and repeated. I moved it to the top of the file, with doxygen notation to create a separate page and added references to it at all the locations in code where it was repeated.

client/human/HumanClientFSM.cpp
LGM-Doyle: clarified.

geoffthemedio:
better now.

the top-of-file comment still isn't great / perfect, but I don't have any immediate specific complaints.

LGM-Doyle:
I agree a non-local comment is less likely to be read when it needs to be read.

However, I felt that the comment having reached comically long proportions and being repeated 12 times risked being updated inconsistently in differing locations, resulting in one correct comment and 11 misleading comments. 

This way there is one correct comment, that no one will read :)

geoffthemedio:
I meant the current comment, which is at the top of the file, isn't great. Not that it being at the top of the file isn't great. It's fine where it is, just not written perfectly.

LGM-Doyle:
I'll need more specific criticism of the comment, if you want it changed.  I wrote it, so it seems clear to me.

I'm trying to summarize enough of the `boost::statechart` manual/tutorial so that the mechanics of reactions and the sequencing of state transitions are clear enough that the ordering requirements in the code are obvious.

geoffthemedio:
As I wrote before, I don't have any immediate specific complaints. Just a vague feeling it could be improved. But it's fine as is.

Suppose I should merge this then...

Issue #622, opened 2016-04-24
Add Editable Save File Format Option


geoffthemedio:
Recently someone asked in the forums how to edit save files, and in a reply it was noted that in v0.4.5, one could set the save file format to XML, to get a semi-human-editable save format.

In the latest builds, however, the non-binary-mode save is a readable-text header followed by compressed, and thus unreadable or editable, text.

It shouldn't be too difficult to add another option, for "uncompressed XML" saves, in which the main bulk of the gamestate is still semi-readable serialized text.

adrianbroher:
It's complete an utter nonsense to introduce another option for something that should not be configurable in the first place.  Just save the files as xml, but compress the whole file with the standard algorithm when saving and uncompress when loading.  Also get rid of the binary save files options.  It's not platform transparent and therefor garbage.

geoffthemedio:
The whole file being compressed means the whole file has or would have to be loaded before any info in it can be used. This creates a bunch of problems, most notably substantial delays when populating save file lists in the UI. I added an uncompressed header to the format in part to avoid that.

Binary saves might still be faster than loading and uncompressing the XML, and can likely work with less memory requirements. In order to handle the compressed XML, the whole compressed string and the whole uncompressed string need to be stored continguously in memory buffers. This could potentially cause problems when attempting to reserve that large of a memory block (~500 MB); at least it did before I reworked things to attempt to reserve enough space in the decompressed text buffer before doing the decompression filtering, and a particularly large save file might still have issues, due the the verbosity of tagged XML serialization.

LGM-Doyle:
Is it possible to change from a save file to a save directory?

Each section of the save can be a separate compressed XML file.  
The simplest partition is a header file and an everything else file.  Other partitions might minimize the amount of data that needs to be loaded for different executables.  Perhaps header, current turn and historical to prevent loading historical data until the human client requests it.  Perhaps even break up current turn and historical into per-client save files to minimize client loading.

XML does allow file inclusion, so the header could just include all of the other stuff, if the XML parser is lazy enough to not expanded tags until requested.

Is  a directory with two compressed XML files header and everything else and way to have small file size, fast loading header information and transparent data?

geoffthemedio:
Possible, yes, but I wouldn't do anything more complicated to deal with than a single save file.

geoffthemedio:
https://github.com/freeorion/freeorion/pull/1427

geoffthemedio:
[...]

PR #1427, opened 2017-03-21
Uncompressed Save Option

geoffthemedio:
Adds an option to save with uncompressed XML. This makes the save files much bigger, but has some advantages:
* It will probably be portable, like compressed XML.
* It will probably produce save files that can be loaded even when the compressed XML can't due to the buffer required to decompress the XML being too big to allocate in memory. The uncompressed format can probably be streamed directly from disk, and shouldn't need to be all held in memory at once.
* It can be edited with a text editor to change the gamestate in the save.

Should be backwards compatible with old save files.

save_save_different_formats

Should resolve issue #622

### DIFF
--- a/server/SaveLoad.cpp
+++ b/server/SaveLoad.cpp
@@ -123,7 +123,8 @@ int SaveGame(const std::string& filename, const ServerSaveGameData& server_save_
     ScopedTimer timer("SaveGame: " + filename, true);
 
     bool use_binary = GetOptionsDB().Get<bool>("binary-serialization");
-    DebugLogger() << "SaveGame(" << (use_binary ? "binary" : "zlib-xml") << ") filename: " << filename;
+    bool use_zlib_for_zml = GetOptionsDB().Get<bool>("xml-zlib-serialization");
+    DebugLogger() << "SaveGame(" << (use_binary ? "binary" : (use_zlib_for_zml ? "zlib-xml" : "raw-xml")) << ") filename: " << filename;
     GetUniverse().EncodingEmpire() = ALL_EMPIRES;
 
     DebugLogger() << "Compiling save empire and preview data";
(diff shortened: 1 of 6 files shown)


dbenage-cx:
Seems to work ok so far, successfully changed the size of a planet and loaded fine.
One caveat was changing to uncompressed xml (unchecking the new option) required a restart of the client to take effect.

Possibly feature request after this merges: have separate options for auto saves and manual (or an override within the save dialog).

geoffthemedio:
> required a restart of the client to take effect.

Should just require restarting the server, not the full client. If the option is changed while a game is being played, the server isn't informed (about this or any other option), so it will continue to use the setting that was in effect when the server was launched, likely at the start of the game (unless using an external / separately launched server).

geoffthemedio:
Tacked on some additional tracking of save text size to this branch. Should avoid allocating substantially more RAM for save load buffers than necessary.

adrianbroher:
Quote from [#1421][1]:

> Or here (not sure if it's a legit problem in this specific case):
> https://travis-ci.org/freeorion/freeorion/builds/214677710
> It can't find Universe::serialize apparently?

Well, it can't.  You changed some code, which caused to require said symbol to be exported by `freeorioncommon`, which currently doesn't happen.  This only occurs on OSX because we build shared libraries out of `freeorioncommon`  and because the compiler complains about undefined symbols by default, where it doesn't on Linux.

To fix this you need to explicit instantiate `Universe::serialize` in `util/SerializeUniverse.cpp`.  You can take the explicit instantiations of `SpeciesManager::serialize` as [a reference][2].

[1]: https://github.com/freeorion/freeorion/pull/1421#issuecomment-289290191
[2]: https://github.com/freeorion/freeorion/blob/master/util/SerializeUniverse.cpp#L263-L270

server/SaveLoad.cpp
```diff
@@ -167,75 +168,99 @@ int SaveGame(const std::string& filename, const ServerSaveGameData& server_save_
...
+                    boost::iostreams::stream<InsertDevice> s_sink(serial_inserter);
+
+                    // create archive with (preallocated) buffer...
+                    freeorion_xml_oarchive xoa(s_sink);
+                    // serialize main gamestate info
+                    xoa << BOOST_SERIALIZATION_NVP(player_save_game_data);
+                    xoa << BOOST_SERIALIZATION_NVP(empire_manager);
+                    xoa << BOOST_SERIALIZATION_NVP(species_manager);
+                    xoa << BOOST_SERIALIZATION_NVP(combat_log_manager);
+                    xoa << BOOST_SERIALIZATION_NVP(universe);
```
adrianbroher: You're calling `Universe::serialize` here indirectly, whereas previously only…

server/SaveLoad.cpp
```diff
@@ -167,75 +168,99 @@ int SaveGame(const std::string& filename, const ServerSaveGameData& server_save_
...
-                boost::iostreams::stream<InsertDevice> s_sink(serial_inserter);
-
-                // create archive with (preallocated) buffer...
-                freeorion_xml_oarchive xoa(s_sink);
-                // serialize main gamestate info
-                xoa << BOOST_SERIALIZATION_NVP(player_save_game_data);
-                xoa << BOOST_SERIALIZATION_NVP(empire_manager);
-                xoa << BOOST_SERIALIZATION_NVP(species_manager);
-                xoa << BOOST_SERIALIZATION_NVP(combat_log_manager);
-                Serialize(xoa, universe);
```
adrianbroher, line 206: … the global `Serialize` function was called.

geoffthemedio:
maybe fixed?

PR #1428, opened 2017-03-22
Categorize universe data articles

dbenage-cx:
Moves the '[Game - ***]', and both normal and monster ship designs pedia categories as sub-categories to a new top-level category.

PR #1429, opened 2017-03-22
Fix incomplete pedia search of categorized results.

dbenage-cx:
The categorized pedia entries were not shown in search results.

Compatible (but not required) with PR #1428

### DIFF
--- a/UI/EncyclopediaDetailPanel.cpp
+++ b/UI/EncyclopediaDetailPanel.cpp
@@ -47,7 +47,7 @@
 #include <boost/algorithm/clamp.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/replace.hpp>
-
+#include <unordered_map>
 
 using boost::io::str;
 
@@ -130,7 +130,7 @@ namespace {
                                                        meter_name.second,
                                                        meter_name.first) + "\n";
 
-        list.insert({meter_name.first, {link_string, "ENC_METER_TYPE"}});
+        list.insert({meter_name.first, {link_string, meter_name.second}});
     }
 
     const std::vector<std::string>& GetSearchTextDirNames() {
@@ -939,34 +939,56 @@ void EncyclopediaDetailPanel::HandleLinkDoubleClick(const std::string& link_type
     }
 }
 
+namespace {
+    std::unordered_map<std::string, std::pair<std::string, std::string>>
+        GetSubDirs(const std::string& dir_name, int depth = 0)
+    {
+        depth++;
+        std::unordered_map<std::string, std::pair<std::string, std::string>> retval;
+        // safety check to pre-empt potential infinite loop
+        if (depth > 50) {
+            WarnLogger() << "Exceeded recursive limit with lookup for pedia"
+                         << " category " << dir_name;
+            return retval;
+        }
+
+        std::unordered_map<std::string, std::pair<std::string, std::string>> sub_dirs;
+        std::multimap<std::string, std::pair<std::string, std::string>> sorted_entries;
+        GetSortedPediaDirEntires(dir_name, sorted_entries);
+        for (auto&& entry : sorted_entries) {
+            // explicitly exclude textures
+            if (entry.second.second == "ENC_TEXTURES")
+                continue;
+            sub_dirs.emplace(entry.second.second,
+                             std::make_pair(entry.first, entry.second.first));
+        }
+
+        // Add each article for this directory
+        for (auto&& article : sub_dirs) {
+            if (article.first == dir_name)
+                continue;
+            // If a sub-category, add this article and any nested categories
+            for (const auto& sub_article : GetSubDirs(article.first, depth))
+                retval.emplace(sub_article);
+            retval.emplace(article);
+        }
+
+        return retval;
+    }
+}
+
 void EncyclopediaDetailPanel::HandleSearchTextEntered() {
     // search lists of articles for typed text
     const std::string& search_text = m_search_edit->Text();
     if (search_text.empty())
         return;
 
-    std::string match_report;
-    std::set<std::pair<std::string, std::string>> already_listed_results;  // pair of category / article name
-
     // assemble link text to all pedia entries, indexed by name
-    std::vector<std::string> dir_names = GetSearchTextDirNames();
-    for (const std::map<std::string, std::vector<EncyclopediaArticle>>::value_type& article : GetEncyclopedia().articles) {
-        dir_names.push_back(article.first);
-    }
+    std::unordered_map<std::string, std::pair<std::string, std::string>>
+        all_pedia_entries_list;
 
-    // map from human-readable-name to (link-text, category name)
-    std::multimap<std::string, std::pair<std::string, std::string>> all_pedia_entries_list;
-
-
-    // for every directory, get all entries, add to common multimap for ease of
-    // repeated searching for different types of match...
-    for (const std::string& type_text : dir_names) {
-        std::multimap<std::string, std::pair<std::string, std::string>> sorted_entries_list;
-        GetSortedPediaDirEntires(type_text, sorted_entries_list);
-        for (std::multimap<std::string, std::pair<std::string, std::string>>::value_type& entry : sorted_entries_list) {
-            all_pedia_entries_list.insert(entry);
-        }
-    }
+    for (auto&& subdir_name : GetSubDirs("ENC_INDEX"))
+        all_pedia_entries_list.emplace(subdir_name);
 
     // find distinct words in search text
     std::set<std::string> words_in_search_text;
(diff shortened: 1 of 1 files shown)


UI/EncyclopediaDetailPanel.cpp
```diff
@@ -59,6 +59,9 @@ namespace {
     void    AddOptions(OptionsDB& db) {
         db.Add("UI.dump-effects-descriptions", UserStringNop("OPTIONS_DB_DUMP_EFFECTS_GROUPS_DESC"),  false,  Validator<bool>());
         db.Add("UI.encyclopedia.search.articles", UserStringNop("OPTIONS_DB_ENC_SEARCH_ARTICLE"),     true,   Validator<bool>());
+        db.Add("UI.encyclopedia.search.recursive-limit",
```
geoffthemedio: I don't think this needs to be an adjustable option...

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -939,6 +942,39 @@ void EncyclopediaDetailPanel::HandleLinkDoubleClick(const std::string& link_type
     }
 }
 
+namespace {
+    std::unordered_set<std::string> GetSubDirs(const std::string& dir_name,
+                                               int depth = 1)
+    {
+        std::unordered_set<std::string> retval {};
```
geoffthemedio: the {} presumably does nothing here?

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -939,6 +942,39 @@ void EncyclopediaDetailPanel::HandleLinkDoubleClick(const std::string& link_type
...
+        const int depth_limit =
+            GetOptionsDB().Get<int>("UI.encyclopedia.search.recursive-limit");
+        if (depth > depth_limit) {
+            WarnLogger() << "Exceeded recursive limit with lookup for pedia"
+                         << " category " << dir_name;
+            return retval;
+        }
+
+        const auto& article_it = GetEncyclopedia().articles.find(dir_name);
+        if (article_it != GetEncyclopedia().articles.end()) {
```
geoffthemedio: can rearrange this to exit if `article_it == ().end` and then de-indent the following code.

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -949,9 +985,11 @@ void EncyclopediaDetailPanel::HandleSearchTextEntered() {
     std::set<std::pair<std::string, std::string>> already_listed_results;  // pair of category / article name
 
     // assemble link text to all pedia entries, indexed by name
-    std::vector<std::string> dir_names = GetSearchTextDirNames();
-    for (const std::map<std::string, std::vector<EncyclopediaArticle>>::value_type& article : GetEncyclopedia().articles) {
-        dir_names.push_back(article.first);
+    std::unordered_set<std::string> dir_names;
```
geoffthemedio: should you be adding entries to an unordered set while iterating over it?

are circular subcategory relationships possible to script?

dbenage-cx: Not seeing where this occurs? `dir_names` is not iterated over until `all_pedia_entries_list` is built.
I've renamed `dir_name` to `topdir_name` in case it caused confusion.

Circular categories are possible, though they should not display to begin with.

dbenage-cx:
Rebased for review changes.

UI/EncyclopediaDetailPanel.cpp
geoffthemedio: nevermind... the outer loop is over the result of `GetSearchTextDirNames()` not over `dir_names`.

geoffthemedio:
With vs. without the patch finds different articles when searching, but neither finds all it should.

patched_search
master_search

Vezzra:
How important is this for 0.4.7? If not really necessary, I don't want to block the release for this, so unless someone objects (preferably by explaining why they think it's important enough to hold up the release), I will reassign it to the "optional for 0.4.7" milestone.

geoffthemedio:
I think it's better if, and probably pretty important that, it finds all (or at least more) articles with the searched word in the title, which it appears to do better with the patch applied. I didn't test thoroughly, though.

dbenage-cx:
I have improved results a good deal since the review, but still missing some articles so have not pushed yet (there should be at least 20 results for `Article Content Matches` searching for `empire`).

Are you wanting to release in the next few days?

dbenage-cx:
Search fails to get description results for any generated articles, I expect this was an issue before the work with new categories.
Unless there is some suggestion on how to resolve it, I propose to push that portion to post release where a larger refactor of the class may be warranted.

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -985,70 +1007,73 @@ void EncyclopediaDetailPanel::HandleSearchTextEntered() {
...
-    // search for exact title matches
-    match_report += UserString("ENC_SEARCH_EXACT_MATCHES") + "\n\n";
-    for (const std::multimap<std::string, std::pair<std::string, std::string>>::value_type& entry : all_pedia_entries_list) {
-        if (boost::iequals(entry.first, search_text)) {
-            match_report += entry.second.first;
-            already_listed_results.insert({entry.second.second, entry.first});
-        }
-    }
+    const std::string heading_idx = "";
+    std::vector<std::multimap<std::string, std::string>> reports = {
```
geoffthemedio: not a fan of this and the magic number indexing, eg `[1]` below.

Vezzra:
> Are you wanting to release in the next few days?

No, certainly not. 😉

However, currently there are a couple of issues and PRs assigned to the "0.4.7" milestone where I'm not sure if they really need to be considered release blocking. So I want to do some housekeeping and only keep issues and PRs assigned to "0.4.7" that really are critical for the release.

As nice as it would be to have all these things in 0.4.7, we have to draw the line somewhere, otherwise we'll never get this release out. It's fine to keep sneaking in things while the critical stuff is still worked on, but once all the actually release blocking things have been completed/finished, I want to create the release branch ASAP, without having to wait for another bunch of issues/PRs which are nice to have, but not really *critical*.

That said, @geoffthemedio already stated that he thinks this PR should go in, so it will remain assigned to "0.4.7".

geoffthemedio:
I suppose it's not essential for v0.4.7.

UI/EncyclopediaDetailPanel.cpp
dbenage-cx: changed to specific multimaps

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -939,34 +939,56 @@ void EncyclopediaDetailPanel::HandleLinkDoubleClick(const std::string& link_type
...
+        }
+
+        std::unordered_map<std::string, std::pair<std::string, std::string>> sub_dirs;
+        std::multimap<std::string, std::pair<std::string, std::string>> sorted_entries;
+        GetSortedPediaDirEntires(dir_name, sorted_entries);
+        for (auto&& entry : sorted_entries) {
+            // explicitly exclude textures
+            if (entry.second.second == "ENC_TEXTURES")
+                continue;
+            sub_dirs.emplace(entry.second.second,
```
geoffthemedio, line 962: can't replace `make_pair` with `{}` ?

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -985,70 +1007,97 @@ void EncyclopediaDetailPanel::HandleSearchTextEntered() {
...
+    bool first_report = true;
+    if (!exact_match_report.empty()) {
+        first_report = false;
+        match_report += UserString("ENC_SEARCH_EXACT_MATCHES") + "\n\n";
+        for (auto&& match : exact_match_report)
+            match_report += match.second;
+    }
+
+    if (!word_match_report.empty()) {
+        if (!first_report)
```
geoffthemedio: this doesn't seem like it will work right if you don't reset `first_report` to true after each heading and double `\n\n`

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -985,70 +1007,97 @@ void EncyclopediaDetailPanel::HandleSearchTextEntered() {
...
-            EncyclopediaArticle article_entry = GetPediaArticle(entry.second.second);
-            if (boost::icontains(UserString(article_entry.description), search_text)) {
-                match_report += entry.second.first;
-                already_listed_results.insert({entry.second.second, entry.first});
-                break;
-            }
+        // search for matches within article text
+        EncyclopediaArticle article_entry = GetPediaArticle(entry.first);
+        if (article_entry.description.empty()) {
+            // FIXME Articles generated through GetRefreshDetailPanelInfo
```
geoffthemedio: I don't think searching dynamically generated article text is unnecessary. Most such articles are lists of stuff that would be found in another article anyway, or are lists of in-game objects that would be better listed directly in the search results themselves. Actually searching them all would also mean constructing them all, which would be very slow, and caching the results in some sort of search index, but that would be problematic because dynamically generated article contents can and do change.

dbenage-cx: Ok, changed to a `NOTE`

UI/EncyclopediaDetailPanel.cpp
dbenage-cx: I've changed to prepend newline for all report headings.

UI/EncyclopediaDetailPanel.cpp
dbenage-cx: Not in this case, braced init list has no type.
http://stackoverflow.com/questions/24229580/why-variadic-function-cant-eat-the-list-initialization-argument-in-c11

dbenage-cx:
TY for reviews

Issue #1419, opened 2017-03-19
Disable pop consumption for building colony ships until issues with that feature can be solved


Vezzra:
Currently there are issues with the new feature that building colony ships consume population. Until we can come up with a more thorough solution to these issues, disable the mechanic so they don't block the release.

Vezzra:
Addressed by #1432.

PR #1432, opened 2017-03-22
Remove colony population consumption until outpost/colony discrepancy is fixed.

LGM-Doyle:
Fixes issue #1419.


geoffthemedio:
looks reasonable

PR #1434, opened 2017-03-23
Let AI consider fighter vs planet mechanics

Morlic-fo:
This PR introduces an additional ```rating_vs_planets``` which does not consider fighter stats for combat rating calculations.

Generally, all situations where we force combat in a system with enemy planets should check both the total fleet strength vs total enemy threat and fleet strength vs planets vs enemy planet threat.

There may be more places in the code which require those checks. The ones in this PR are all I found. In case more are found, it should be trivially to add the relevant checks...

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -456,11 +456,14 @@ class AIstate(object):
             sys_status['enemies_supplied'] = enemy_supply.get(sys_id, [])
             sys_status['enemies_nearly_supplied'] = enemy_near_supply.get(sys_id, [])
             my_ratings_list = []
+            my_ratings_against_planets_list = []
             for fid in sys_status['myfleets']:
-                this_rating = self.get_rating(fid, True, self.get_standard_enemy())  # DONE!
+                this_rating = self.get_rating(fid, True, self.get_standard_enemy())
                 my_ratings_list.append(this_rating)
+                my_ratings_against_planets_list.append(self.get_rating(fid, against_planets=True))
             if sys_id != INVALID_ID:
-                sys_status['myFleetRating'] = my_ratings_list and CombatRatingsAI.combine_ratings_list(my_ratings_list) or 0
+                sys_status['myFleetRating'] = CombatRatingsAI.combine_ratings_list(my_ratings_list)
+                sys_status['myFleetRatingVsPlanets'] = CombatRatingsAI.combine_ratings_list(my_ratings_against_planets_list)
                 sys_status['all_local_defenses'] = CombatRatingsAI.combine_ratings(sys_status['myFleetRating'], sys_status['mydefenses']['overall'])
             sys_status['neighbors'] = set(dict_from_map(universe.getSystemNeighborsMap(sys_id, self.empireID)))
             
@@ -584,7 +587,7 @@ class AIstate(object):
                 return True
         return False
 
-    def get_rating(self, fleet_id, force_new=False, enemy_stats=None):
+    def get_rating(self, fleet_id, force_new=False, enemy_stats=None, against_planets=False):
         """Returns a dict with various rating info."""
         if fleet_id in self.fleetStatus and not force_new and enemy_stats is None:
             return self.fleetStatus[fleet_id].get('rating', 0)
@@ -592,10 +595,11 @@ class AIstate(object):
             fleet = fo.getUniverse().getFleet(fleet_id)
             if not fleet:
                 return {}  # TODO: also ensure any info for that fleet is deleted
-            status = {'rating': CombatRatingsAI.get_fleet_rating(fleet_id, enemy_stats),  # TODO
+            status = {'rating': CombatRatingsAI.get_fleet_rating(fleet_id, enemy_stats),
+                      'ratingVsPlanets': CombatRatingsAI.get_fleet_rating_against_planets(fleet_id),
                       'sysID': fleet.systemID, 'nships': len(fleet.shipIDs)}
             self.fleetStatus[fleet_id] = status
-            return status['rating']
+            return status['rating'] if not against_planets else status['ratingVsPlanets']
 
     def update_fleet_rating(self, fleet_id):
         self.get_rating(fleet_id, force_new=True)
@@ -663,6 +667,7 @@ class AIstate(object):
         """Removes fleetRoles if a fleet has been lost, and update fleet Ratings."""
         for sys_id in self.systemStatus:
             self.systemStatus[sys_id]['myFleetRating'] = 0
+            self.systemStatus[sys_id]['myFleetRatingVsPlanets'] = 0
 
         universe = fo.getUniverse()
         ok_fleets = FleetUtilsAI.get_empire_fleet_ids()
--- a/default/python/AI/CombatRatingsAI.py
+++ b/default/python/AI/CombatRatingsAI.py
@@ -166,15 +166,20 @@ class ShipCombatStats(object):
         """
         return self._fighter_stats.get_stats()
 
-    def get_rating(self, enemy_stats=None):
+    def get_rating(self, enemy_stats=None, ignore_fighters=False):
         """Calculate a rating against specified enemy.
 
         :param enemy_stats: Enemy stats to be rated against
         :type enemy_stats: ShipCombatStats
+        :param ignore_fighters: If True, acts as if fighters are not launched
+        :type ignore_fighters: bool
         :return: rating against specified enemy
         :rtype: float
         """
         # adjust base stats according to enemy stats
+        def _rating():
+            return my_total_attack * my_structure
+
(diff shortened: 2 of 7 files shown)


MatGB:
K, @Dilvish-fo said to check and test this PR so I rebased it into my testing branch for a game I'd started earlier today, hadn't got to turn 100 yet. It's now turn 153 and I'm in trouble, two nearby AIs are clearly superior to me in tech and total military force: I have more colonies so I'll survive but it's the first time I've really felt challenged for months.  Some of that is obviously due to the colonisation bug but a lot of it has to be this combined with other recent changes.

I haven't noticed anything specifically bad, colonies are definitely being built properly and I've yet to see a massive pile of troop ships waiting for a warship to turn up. I have seen some strange routes taken by AI ships but that could easily be reprioritising due to unseen-by-me threats. Ergo, my initial assessment is that this is a substantial improvement.

Also, that next cycle we need to make the planetary defence techs more expensive and split them up a bit because top tier defences set on Defend when you're still using Robohulls means dead fleets, and the nearest AI to me appears to have rushed them.

Morlic-fo:
Sounds good. Most of your observations are a cumulative effect over the last changes and probably not directly caused by this PR (always good it does not obviously break something, though). 

That being said, the impact of this PR is hard to measure. It will mostly affect edge cases and those will (or: should) simply not happen anymore. Basically, as long as you do not see fighter-only fleets trying to attack planets, it seems to serve its purpose. But then again, hard to say if the AI would have otherwise without this PR...

memem359:
Short version: Haven't noticed any problems, and at least one of the AI was behaving smarter.

About two weeks ago for a particular seed, the Chato_AI conquered a Mu Ursh system, but failed to colonize Good planets 1 jump from its homeworld.
After last weeks changes, the Chato_AI colonized those Good planets, but ignored the Mu Ursh.
With the build from today, including these suggested code changes, the Chato_AI colonized AND conquered. So whatever alchemy of decisions led to this point, it is definitely an improvement over a month ago.

I did notice one thing (spying with the Super-Tester and looking through the AI_log).
It is probably unrelated to these changes, but I'll mention it just in case.
For the Mu Ursh system, it had 26 defending troops. The Chato_AI created a goal of having 26 troops. (Seemed to be procrastinating on building them, but maybe it had other concerns.) It then eventually sent those 26 troops to invade, got the Mu Ursh to 0, then started building ships for 10 more troops. After those were built, it then invaded with all of them (overkill). I'd think that the AI should have requested more than 26 troops to start with, and/or not use so much troop overkill.

Morlic-fo:
> then started building ships for 10 more troops.

If it was a single ship completely stacked with troops, then this is currently the expected (but obviously non-optimum) behaviour.

Morlic-fo:
> For the Mu Ursh system, it had 26 defending troops. The Chato_AI created a goal of having 26 troops.

Fix attempt: #1444

Vezzra:
Needs rebase.

Aside from this, anything preventing this from getting merged? @freeorion/ai-team?

Morlic-fo:
Rebased.

Dilvish-fo:
Good job, this looks to me like a great start, and playtests fine for me.  

If you have a chance, some more comment lines particularly in MilitaryAI would be very helpful -- the way these allocation classes are set up it seems I have to do an awful lot of searching and jumping around to figure out just where and how ratings vs planets are being used.

I am also noting some ideas here about some points that I think would be most helpful to focus on for continued improvement in this area post-0.4.7, though perhaps we should take that ongoing discussion to the programming forum?

- using the vs planets rating for defense planning as well as offense:  this PR puts sets up the vs planets rating to be used for offensive planning, which was the biggest point of need, but the AI military planning would be more effective if we also took these issues into account for defense planning as well.

- ~~vs planets rating excluding fighters altogether: this seems like a reasonable first pass sort of estimate, which should be a conservative one for the offensive planning purposes where it is used in this PR.  The planets can actually attack the fighters, though, even if the fighters cannot attack back, and so even in a pure ships vs planet battle the fighters should actually help the ship side a bit and it would be nice if we could more accurately capture that.~~ (nm, as has been pointed out I misunderstood that part of the combat code)

- Mixed Battle Assessment: with this PR the code takes into account both ships vs. ships and ships vs. planets; the handling for assessing ships vs a ships+planets force, if I am understanding it correctly, seems like a reasonable stopgap (setting a target rating as the max of (i) regular rating for total enemy forces or (ii) non-fighter rating vs just enemy planets rating), but still a rather rough estimate.  I think these will probably be the most significant type of battles (the ones with the most empire resources at stake).  Trying to figure out how to handle ratings for that kind of battle will be more tricky, but I expect it would be quite helpful.

Dilvish-fo (approved):

Vezzra:
> perhaps we should take that ongoing discussion to the programming forum?

That would be better. After all, the comment sections of PRs are meant for reviewing/discussing the proposed implementation. Although we go beyond that frequently enough, we should make some effort to keep that to a minimum. 😉 

> The planets can actually attack the fighters, though, even if the fighters cannot attack back

OUCH! Are you sure about that? Because the whole point of this restriction actually is that *planets* don't target *fighters* in combat, and waste their high damage shots on a myriad of low structure targets. Otherwise fighters would become the ultimate anti-planet weapon.

Of course that only works if fighters can't target planet also.

Meaning, if planets actually can and do target fighters in combat, that would be a very serious, release blocking bug.

@geoffthemedio? As the one who implemented the code in question, can you shed light on this?

MatGB:
I haven't ever noticed a planet targetting a fighter and I've just checked through my most recent several turns and can't seen an instance, the game is at the stage where planets can actually still be firing in the second or third round of combat.

Of course, not seeing something happening doesn't mean it isn't/can't happen, but it shouldn't and if it is it needs fixing, at least until we rework planetary defences (because I do think planets should have fighter bases at some point and a ROF for planetary defences would be cool)

geoffthemedio:
https://github.com/freeorion/freeorion/blob/master/combat/CombatSystem.cpp#L1174

Planets seemingly shouldn't be attacking fighters...

Dilvish-fo:
Ah, sorry, I had seen the function AttackPlanetFighter
https://github.com/freeorion/freeorion/blob/master/combat/CombatSystem.cpp#L469

and forgot to check that it was actually used (whereas I had checked that there is no AttackFighterPlanet that I could see) and so I got confused by that.  So never mind that point, I have marked that part of my comment above with strikethru now to minimize confusion for future readers.

Dilvish-fo:
ps., although now I checked, and [AttackPlanetFighter *is* actually used](https://github.com/freeorion/freeorion/blob/master/combat/CombatSystem.cpp#L573), although apparently in a pathway that can't actually be triggered due to the line that Geoff pointed to, so perhaps the AttackPlanetFighter function deserves a comment noting that it is currently never triggered due to the line that Geoff noted.

Vezzra:
Ah, ok. Good 😃 

So, LGTM, right?

MatGB:
Yup, it's working fine my end.

Dilvish-fo:
> So, LGTM, right?

Sure, I had approved the code already, I just held off from merging in case Morlic wanted to add some more comments, but that could always be done later.

geoffthemedio:
I wrote "seemingly shouldn't", not "don't" because it shouldn't but not necessarily can't happen.

Morlic-fo:
> I just held off from merging in case Morlic wanted to add some more comments, but that could always be done later.

I will update the comments by providing another PR / pushing to master directly.


Thanks for testing / review.

Morlic-fo:
> the handling for assessing ships vs a ships+planets force, if I am understanding it correctly, seems like a reasonable stopgap (setting a target rating as the max of (i) regular rating for total enemy forces or (ii) non-fighter rating vs just enemy planets rating), but still a rather rough estimate.

The intention was to use both:
(i) The attacking fleet must have a regular rating greater than the total defenses
AND
(ii) The attacking fleet must have a rating vs planets (i.e. rating without fighters) greater than the planetary defenses.

Both are required for fleet allocation:
https://github.com/freeorion/freeorion/pull/1434/files#diff-26ae651908f4d03ac27d4d28e5cac947R786

The maximum is considered [here](https://github.com/freeorion/freeorion/pull/1434/files#diff-26ae651908f4d03ac27d4d28e5cac947R192) as the rating vs planets (rvp) is always smaller or equal to the actual rating. So if we need X rvp, then our allocation must have at least Y regular rating even if our regular rating is already greater than the total rating of the enemy force.
I do this so the estimation of the remaining rating after allocation as  [in these blocks](https://github.com/freeorion/freeorion/pull/1434/files#diff-26ae651908f4d03ac27d4d28e5cac947R694) is still more or less valid even if we do not require any more regular rating (e.g. having a large fleet of fighters vs planets)

But yes, generally all the combat ratings are somewhat crude estimates and should be improved at some point.

PR #1436, opened 2017-03-24
Fix crash caused by erasing map elements while iterating in System.cpp.

LGM-Doyle:
This might fix the crash reported in [the forum](http://freeorion.org/forum/viewtopic.php?f=28&t=10456&sid=6abcfe7c1f17f67c51ee477fdfa7c7f7), where after the Experimentors are active the game crashes on OSX.

In System::Copy there is for loop iterating over the map of starlanes and deleting elements. It does not account for the deleted elements while iterating so it could step into nowhere.  

This PR fixes that problem.  However, since I can't replicate the crash it may not fix the crash.

### DIFF
--- a/universe/System.cpp
+++ b/universe/System.cpp
@@ -162,10 +162,15 @@ void System::Copy(std::shared_ptr<const UniverseObject> copied_object, int empir
             // to exist any more
 
             // remove previously known lanes that aren't currently visible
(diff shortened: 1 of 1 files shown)


geoffthemedio:
looks reasonable

Vezzra:
And just for the records, it indeed fixed the issue, at least on my system. The save game provided in that forum thread no longer crashes when executing the next turn. 👍

PR #414, opened 2015-12-04
[AI] remove compatibility with old saves

Cjkjvfnby:
Last changes in master break compatibility with old saves: https://github.com/freeorion/freeorion/commit/9de778d6e4fa0dd3229a6e377d4d53d3f1c90859#commitcomment-14781649

PR #1438, opened 2017-03-24
Some cleanup

Morlic-fo:
- Mostly converting some dicts that were used like sets to actual sets.
- Remove some redundant / unused variables
- fix some PEP8 naming style violations
- minor code simplifications

Should be squashed before merge but leaving history as is for easier review.

Tagging this as optional for 0.4.7 as it shouldn't be release blocking but the changes in AIstate break save compatibility as usual so I would strongly prefer to merge this before 0.4.7 (so we can actually use 0.4.7 saves in master for a little while...)

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -79,7 +79,12 @@ def convert_to_version(state, version):
             state['_aggression'] = state['character'].get_trait(Aggression).key
         except Exception as e:
             raise ConversionError("Error when converting to compatibility version 1: "
-                                   "Can't find aggression in character module. Exception thrown was: " + e.message)
+                                  "Can't find aggression in character module. Exception thrown was: " + e.message)
+    elif version == 2:
+        # some dicts which used only the keys were changed to sets
+        for var_name in ['visInteriorSystemIDs', 'exploredSystemIDs', 'visBorderSystemIDs', 'unexploredSystemIDs']:
+            state[var_name] = set(state.get(var_name, {}).keys())
+
     #   state["some_new_member"] = some_default_value
     #   del state["some_removed_member"]
     #   state["list_changed_to_set"] = set(state["list_changed_to_set"])
@@ -98,7 +103,7 @@ class AIstate(object):
     is playable with this AIstate version, i.e. new members must be added
     and outdated members must be modified and / or deleted.
     """
-    version = 1
+    version = 2
 
     def __init__(self, aggression):
         # Do not allow to create AIstate instances with an invalid version number.
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -232,7 +232,7 @@ class AIstate(object):
                 for key in diff:
                     del planet_dict[key]
 
-        ExplorationAI.graphFlags.clear()
+        ExplorationAI.graph_flags.clear()
         if fo.currentTurn() < 50:
             print "-------------------------------------------------"
             print "Border Exploration Update (relative to %s)" % (PlanetUtilsAI.sys_name_ids([exploration_center, INVALID_ID])[0])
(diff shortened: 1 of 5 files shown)


default/python/AI/AIstate.py
```diff
@@ -66,10 +66,7 @@ def __init__(self, aggression):
         empire = fo.getEmpire()
         self.empireID = empire.empireID
         homeworld = universe.getPlanet(empire.capitalID)
-        if homeworld:
-            self.__origin_home_system_id = homeworld.systemID
-        else:
-            self.__origin_home_system_id = INVALID_ID
+        self.__origin_home_system_id = homeworld and homeworld.systemID or INVALID_ID
```
Dilvish-fo: this does not work correctly if the homeworld.systemID is 0

but if you want a one-liner I suppose the  = if else
kind of structure could work.

Dilvish-fo:
Hmm, github is apparently hiding my comment as being for an outdated commit, but it looks to me like the line in the overall changes is the same as the one from the commit I commented on, so please be sure to check it.

Dilvish-fo (changes requested):
per previous comment

Dilvish-fo:
> but the changes in AIstate break save compatibility as usual

Changes to AIstate do not need to break save compatibility; at least, backwards compatibility can be preserved, by providing an ```AIstate.__setstate__``` method to make any necessary adjustments during the unpickling process, and I ***very much*** prefer to preserve backwards compatibility for at least a while after any change, it is often quite helpful for people to be able to test out the change with respect to an existing save file.

At one point we hadn't had any such compatibility issues for a while, though, and the method that deals with such things was stripped out in #414.  I had considered leaving a skeleton method with comments in place, but wound up just putting in a [note on the wiki](http://freeorion.org/index.php/Python_Development#AI_state_in_save_file).

This is making me rethink about leaving the skeleton method in place, even when not truly needed at the moment.

Morlic-fo:
> Changes to AIstate do not need to break save compatibility; at least, backwards compatibility can be preserved, by providing an AIstate.__setstate__ method to make any necessary adjustments during the unpickling process, and I very much prefer to preserve backwards compatibility for at least a while after any change, it is often quite helpful for people to be able to test out the change with respect to an existing save file.

Right. No reason to do this before 0.4.7

Dilvish-fo:
So, what do you say, would you add the AIstate.setstate handling so that this need not break savegame compatibility?  It would be nice to get this one merged and out of the way.

Morlic-fo:
Thanks for the reminder. Yes, I will update this PR some time this weekend.

Morlic-fo:
How do you want to handle compatibility across multiple revisions of the AIState?

My idea would be to add a version number as class attribute and do something like
```
class AIState():
    def __setstate__(self, state):
        for i in range(state["version"], AIState.version):
            convert_to_version(state, i+1)
    self.__dict__ = state

...

def convert_to_version(state, version):
    if state["version"] >= version:
        return
    if version == 1:
        state["new_member"] = some_default_value
        del state["old_member"]
    elif version == 2:
        ....
    state["version"] = version
```

Any other preference?

Morlic-fo:
> My idea would be to add a version number as class attribute

Relevant PR: #1584

Dilvish-fo (approved):

Dilvish-fo:
the compatibility version idea sounds fine; I've made some other comments in that PR.

Dilvish-fo:
It just sunk in that the approval was perhaps slightly premature, since you hadn't actually added the conversion here yet, so just take this as a reminder to do that before merging.

Vezzra:
@Morlic-fo, @Dilvish-fo, anything preventing this from being merged?

Morlic-fo:
Rebased and added the code for savegame compatibility.

Ready to merge from my side. Feel free to merge @Dilvish-fo  if you have no new concerns.

PR #1444, opened 2017-03-27
Fix AI invasion troop calculation

Morlic-fo:
Aims to fix the issue reported in this post: http://freeorion.org/forum/viewtopic.php?f=4&p=88039#p88039

We need more troops than defenders. The same number is not enough...

The fix seems straightforward. @Dilvish-fo: Any other places that come to mind or should this be it?

### DIFF
--- a/default/python/AI/InvasionAI.py
+++ b/default/python/AI/InvasionAI.py
@@ -444,7 +444,7 @@ def evaluate_invasion_planet(planet_id, secure_fleet_missions, verbose=True):
         build_time = 4
         planned_troops = troops if system_secured else min(troops + max_jumps + build_time, max_troops)
         planned_troops += .01  # we must attack with more troops than there are defenders
-        troop_cost = math.ceil(planned_troops / 6.0) * 20 * FleetUtilsAI.get_fleet_upkeep()
+        troop_cost = math.ceil((planned_troops+_TROOPS_SAFETY_MARGIN) / 6.0) * 20 * FleetUtilsAI.get_fleet_upkeep()
     else:
         loc = locs[0]
         species_here = universe.getPlanet(loc).speciesName
@@ -455,7 +455,7 @@ def evaluate_invasion_planet(planet_id, secure_fleet_missions, verbose=True):
                                                                CombatRatingsAI.get_species_troops_grade(species_here))
         planned_troops = troops if system_secured else min(troops + max_jumps + build_time, max_troops)
         planned_troops += .01  # we must attack with more troops than there are defenders
-        ships_needed = math.ceil(planned_troops / float(troops_per_ship))
+        ships_needed = math.ceil((planned_troops+_TROOPS_SAFETY_MARGIN) / float(troops_per_ship))
         troop_cost = ships_needed * cost_per_ship  # fleet upkeep is already included in query from server
 
     # apply some bias to expensive operations
--- a/default/python/AI/InvasionAI.py
+++ b/default/python/AI/InvasionAI.py
@@ -493,8 +493,8 @@ def send_invasion_fleets(fleet_ids, evaluated_planets, mission_type):
         sys_id = planet.systemID
         found_fleets = []
         found_stats = {}
-        min_stats = {'rating': 0, 'troopCapacity': ptroops}
-        target_stats = {'rating': 10, 'troopCapacity': ptroops+1}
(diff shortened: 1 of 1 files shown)


Morlic-fo:
Thinking a bit more about it, we also seem to miss some check for ongoing invasion missions that would stop the actual invasion if we do not have sufficient troops (or I did not find that check and the same bug exists there as well).

That should probably adressed in a follow-up PR. I am not familiar enough with the FleetMission code to provide a quick fix for that. @Dilvish-fo could you look into this? (Issue is not really release-blocking though)

Dilvish-fo:
Weird, I had thought this was already taken into account, but if it was at some point then it seems to have been dropped somehow, so I agree some change like this seems necessary.

I think, though, that the better place to make this kind of change, or at least the primary point of change, would be at/around lines 444 and 454, in the calc for "planned_troops" -- this is where we take into account the effects of estimated construction and travel time.  That seems the best place to also add the required margin of victory, so that the troop_cost calc will include that. 

If we did that, then perhaps at the place where you currently proposed the change we could leave the min_stats as simply ptroops, but have target_stats be ptroops+_INVASION_SURPLUS 
where that surplus would be an additional preferred safety factor to account for the various uncertainties in the planning (and we could start it off at just 1).  And the cost estimate at the 444 and 454 vicinity could also take into account that preferred safety factor.

Dilvish-fo:
and while I would agree that this should not be "release blocking",  I strongly feel that at least the simple fix you currently have here should get into the release, if we wind up not agreeing about or getting to the changes I proposed above.

Also, regarding FleetMissions, there is currently a bit of a placeholder for invasions, in the _check_abort_mission() method.  I hadn't gotten to coding up anything there yet because I would want the abort to be a true last resort-- the AI should decide if it can simply augment the already sent troops.  I haven't gotten to that yet, and rather than have the AI abort invasions where they wouldn't immediately succeed I have left it where the AI will just continue with the invasion and then hopefully be keeping the system secured while a new invasion mission is made and sent.  (not ideal, but better than an overly sensitive abort I think).

Vezzra:
> and while I would agree that this should not be "release blocking", I strongly feel that at least the simple fix you currently have here should get into the release, if we wind up not agreeing about or getting to the changes I proposed above.

Definitely. Unless you guys don't get a better/more thorough fix in for 0.4.7, this PR should be merged as is at least.

Morlic-fo:
Yeah, I intended this fix to go in for release, hence the 0.4.7 tag. I meant the follow-up thought of not invading with a fleet that has insufficient troops to not be release-critical.

> Weird, I had thought this was already taken into account, but if it was at some point then it seems to have been dropped somehow, so I agree some change like this seems necessary.

It seems like I introduced the bug 2 years ago when we moved away from hardcoded ship parts to a dynamic look-up: https://github.com/freeorion/freeorion/commit/950b8cd3fd91609959b1f623c6ff357431d9084e

> I haven't gotten to that yet, and rather than have the AI abort invasions where they wouldn't immediately succeed I have left it where the AI will just continue with the invasion and then hopefully be keeping the system secured while a new invasion mission is made and sent. (not ideal, but better than an overly sensitive abort I think).

Maybe it should be checked when issuing the fleet order? That way, the mission would not be aborted and the fleet stays in orbit but the troop ships wouldn't be "wasted". Then again, using the troop ships immediately reduces fleet upkeep, so not sure what is better assuming we do not implement some "decent" logic to consider how likely it is that we keep that system secured...

> I think, though, that the better place to make this kind of change, or at least the primary point of change, would be at/around lines 444 and 454, in the calc for "planned_troops" -- this is where we take into account the effects of estimated construction and travel time. That seems the best place to also add the required margin of victory, so that the troop_cost calc will include that.

Good point, I moved the calculation to these lines. If that is what you had in mind, feel free to squash and merge.

Dilvish-fo:
Yes, this looks good to me.  Since the Travis build error appears to be due solely to some setup problem

```
error: There is no SDK with the name or path '/Users/travis/build/freeorion/freeorion/Xcode/macosx10.7'
```

(and I couldn't see it being affected by these python scripts anyways) I'll go ahead and squash and merge.

Vezzra:
> Since the Travis build error appears to be due solely to some setup problem

Erm, that was just me messing with the Xcode project. Already fixed (reverted the offending commit).

PR #1445, opened 2017-03-27
Allow build to proceed with unknown commit

dbenage-cx:
User on IRC needed to use downloaded source due to poor connection(download allows resume).
The downloaded source can not compile without setup/clone of git repository, as util/Version.cpp is not created.

PR allows build to proceed when `git show -s` fails, setting the commit portion of string to "Unknown"

### DIFF
--- a/cmake/make_versioncpp.py
+++ b/cmake/make_versioncpp.py
@@ -43,10 +43,9 @@ class Generator(object):
 
     def execute(self, version, branch, build_no, build_sys):
         if build_no == INVALID_BUILD_NO:
-            print "WARNING: Can't determine git commit, %s not updated!" % self.outfile
-            return
+            print "WARNING: Can't determine git commit!"
 
-        if os.path.isfile(self.outfile):
+        if os.path.isfile(self.outfile) and not build_no == INVALID_BUILD_NO:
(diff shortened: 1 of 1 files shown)


Vezzra:
Um, on [line 119](https://github.com/freeorion/freeorion/pull/1445/files#diff-c87dd5137c722392e346ca9a759e5ae0R119) `build_no` already gets initialized with a default value:
```
build_no = INVALID_BUILD_NO
```
...with `INVALID_BUILD_NO` being `"????"`. So the build should not fail if `git show -s` fails, the `try...except` clause you adjusted makes sure of that. Whatever causes the compile to fail (`util/Version.cpp` not being created?) must have a different reason.

Unless it's the `"????"` string that somehow messes things up, but in this case the line setting `INVALID_BUILD_NO` should be changed instead (`INVALID_BUILD_NO = "unknown"`).

dbenage-cx:
Changed to allow `INVALID_BUILD_NO` to proceed instead.

adrianbroher:
Tested with the regular build and with an `git export` and works as expected.

Cjkjvfnby:
PS. `git shallow clone` helped me in same situation.  I cloned last state of master and after what fetch other history by chunks.

PR #1446, opened 2017-03-28
Parsed Content Checksum

geoffthemedio:
Would be useful for verifying that clients and server have the same content scripts without having to send all of those scripts.

Work in progress. So far just calculates and output checksums for predefined ship designs and their manager when parsing / constructing.

Not entirely sure what I'm doing, but hopefully the used method will be somewhat implementation-independent.


universe/ShipDesign.cpp
```diff
@@ -84,6 +84,52 @@ namespace {
...
+        sum %= CHECKSUM_MODULUS;
+    }
+
+    template <>
+    void CheckSumCombine(unsigned int& sum, const double& t) {
+        // biggest and smallest possible float should be ~10^(+/-308)
+        // taking log gives a number in the range +/- 309
+        // adding 400 gives numbers in the range ~0 to 800
+        // multiplying by 10'000 gives numbers in the range ~0 to 8'000'000
+        sum += static_cast<unsigned int>((std::log(std::abs(t)) + 400.0) * 10'000);
```
geoffthemedio: should switch to base 10 log for this to make sense

universe/ShipDesign.cpp
```diff
@@ -84,6 +84,52 @@ namespace {
         // have an ID yet) is the same as one that has already been added
         // (which does have an ID)
     }
+
+    constexpr unsigned int CHECKSUM_MODULUS = static_cast<unsigned int>(10'000'000);
+
+    template <class C>
+    void CheckSumCombine(unsigned int& sum, const C& c) {
```
geoffthemedio: effectively implements a custom hash-like function, which I hope is more portable than boost::hash

universe/ShipDesign.cpp
o01eg: Will floating-point works equally on all platforms?

geoffthemedio: My hope is that this method should be relatively robust to platform differences in floating point representation, as it doesn't just use the bit values, but rather manipulates the numerical values.

If there are cases where the value of something gets parsed as values on the order of 10<sup>308</sup> and stored as such, and some platforms have higher allowed float exponents than others, then it could have inconsistent results. There are uses of `FLT_MAX` and `DBL_MAX` in the code, but seemingly only for display purposes, not for actually storing results of parsing, though, and what is parsed is what I want to get a checksum for.

But testing for such issues is partly why I've posted this pull request.

universe/ShipDesign.cpp
```diff
@@ -84,6 +84,52 @@ namespace {
         // have an ID yet) is the same as one that has already been added
         // (which does have an ID)
     }
+
+    constexpr unsigned int CHECKSUM_MODULUS = static_cast<unsigned int>(10'000'000);
```
geoffthemedio: Apparently the Travis CI compiler doesn't support the `'` separator in numeric constants from C++14.

universe/ShipDesign.cpp
o01eg: I wonder if someone play freeorion at the non-x86/x64 platform.

geoffthemedio: they might, but as above, hopefully it won't matter, since nothing in the FOCS should parse to such a high-valued float constant.

universe/ShipDesign.cpp
adrianbroher: Yeah, because we don't support C++14 for now.

adrianbroher:
Why don't you take the hash of the scripts but of the ship design object?

geoffthemedio:
I care if the parsed results are the same, not if the input text is the same.

LGM-Doyle:
Have you considered using [boost::hash](http://www.boost.org/doc/libs/1_63_0/doc/html/hash.html).  We are already using it in a few places, it is platform independent and curated by boost?

geoffthemedio:
Yes, and the design of my custom checksum combine function is based on the boost hash combine function. The boost hash is not, so far as I can tell, portable, which would make it useless for this purpose.

LGM-Doyle:
I'm not exactly sure what you mean by portable.  The page on [portability](http://www.boost.org/doc/libs/1_63_0/doc/html/hash/portability.html) is discussing argument dependent lookup on older compilers, without specifying how old is old.  Two things: 1) I don't think we are compliant with old compilers and 2) were you not intending on explicitly instantiating the hash function which avoids the lookup issue.

A second thing.  Why not serialize the parsed scripts and send them to the clients.  It can't take longer than the turn update.  It only needs to be done on startup.  It avoids all synchronisation issues.  We are already using `boost::serialization` extensively.  It would be the most convenient for multiplayer gamers who are playing on more than one server with differing content scripts.

We will need to checksum the binary content.  It is too large to transmit in a reasonable time.

geoffthemedio:
Portable in the sense that it returns the same value for the same input state / value on different systems, independent of things like the number of bits used to represent values or the order things are stored in containers.

For example, boost hash for floats appears to be based on interpreting each byte of the float representation as a char, or possibly 4 bytes at a time as a size_t, so the results will depend on how the float is represented on the system calculating the checksum.

As well, the method used to combine two values in boost hash into a new value involves evaluating `seed ^= value + (seed<<6) + (seed>>2);` which is probably more order-dependent than the simple adding + taking modulus that I'm using, making the order in which things are stored in containers more important. Possibly that doesn't matter now, but it could in future.

Sending all the parsed content could work, but most of the content manager classes aren't really set up to support that.

LGM-Doyle:
Thanks for the answers.

Vezzra:
@LGM-Doyle, if we don't/can't serialize the entire contents of the `default` folder, it probably won't help/simplify things much if we serialize parts of it. Unless we strictly separate binary contents and scripts with exactly defined rules what happens if a binary content referenced in the scripts can't be found.

LGM-Doyle:
@Vezzra if binary content can't be found the result will be the same as if the checksum doesn't match.  The game won't start with an explanation of what content is mismatched.

A checksum solves all the problems at hand, by forcing the server and clients to use the same data.  A checksum is a lot less work than serializing the rules.

Although it is a moot point, I will outline my thoughts.  

Separating text and binary content is simple.  The game engine already needs to know, so that it can parse text, play sounds and display images.

The `boost::phoenix` parsed content is unambiguously text content, straightforward to serialize in the same way as the gamestate, encompasses all of the game rules defined by scripting content and is small enough to send on startup.  The text content must be identical after parsing on the server and all clients for the game to work.

The binary content is referred to by a text tag indexed to a file.  The binary content does not need to be identical on the server and the clients.  If a client wants to use custom sound effects, fonts or ship icons, it will probably by ok.

Serialize the parsed content and checksum the binary content.  Send both.  The "rules" from the scripting will always be correct.  If any binary content is missing, refuse to start.  If any binary content is different, warn the client, but allow them to continue.

This ignores other issues like:
- does the universe generation code have to be identical? probably not
- does the AI client code need to be identical on the server and the clients? maybe not
- should the messaging protocol be negotiated/confirmed? in an ideal world

geoffthemedio:
By binary content, are you referring to images and sounds? I don't think it's necessary to serialize these in-engine. If they're missing, the player will just see a missing art texture or hear nothing. This wouldn't affect the gameplay mechanics.

Universe generation and AI turn playing is done by the server. Clients don't even need to know what language they are written in.

LGM-Doyle:
I agree.

I had assumed, the binary content would be checksummed. Now that I've checked that there is 150MB of binary content, that was a poor assumption.

o01eg:
I think server-only and client-only resources can be excluded.

geoffthemedio:
With the current version of this change, quickstarting a default game, in the server log from lines 102-117, I get:
````
2017-03-31 00:02:40.744577 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_KRAKEN_1 checksum: 7001
2017-03-31 00:02:40.744577 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_KRAKEN_2 checksum: 8516
2017-03-31 00:02:40.744577 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_KRAKEN_3 checksum: 8377
2017-03-31 00:02:40.745577 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_KRILL_1 checksum: 6140
2017-03-31 00:02:40.745577 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_KRILL_2 checksum: 7742
2017-03-31 00:02:40.745577 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_KRILL_3 checksum: 8706
2017-03-31 00:02:40.745577 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_KRILL_4 checksum: 9669
2017-03-31 00:02:40.745577 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_PSIONIC_SNOWFLAKE checksum: 12823
2017-03-31 00:02:40.746578 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_SNAIL checksum: 7009
2017-03-31 00:02:40.746578 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_SNOWFLAKE_1 checksum: 7487
2017-03-31 00:02:40.746578 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_SNOWFLAKE_2 checksum: 9461
2017-03-31 00:02:40.746578 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_SNOWFLAKE_3 checksum: 9145
2017-03-31 00:02:40.746578 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_TREE checksum: 6277
2017-03-31 00:02:40.746578 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_VOID checksum: 6157
2017-03-31 00:02:40.747579 [debug] Server : ShipDesign.cpp:655 : ShipDesign SM_WHITE_KRAKEN checksum: 8187
2017-03-31 00:02:40.747579 [debug] Server : ShipDesign.cpp:1164 : PredefinedShipDesignManager checksum: 300305
````
Equivalent log lines for other operating systems would be interesting to compare.

**Comment by dbenage-cx** [comment]
The first instance of these seem to line up(different order), there is a second afterwards which increments each by 1.
[freeoriond.log.txt](https://github.com/freeorion/freeorion/files/883993/freeoriond.log.txt)

**Comment by geoffthemedio** [comment]
It should be only the monster ship designs that have checksums +1 relative to the first set of log lines. Up to `SD_TROOP_SHIP` the first and second listings should be the same. From `SM_ACIREMA_GUARD` and after it should be all +1 due to the monster flag being set true when the predefined designs are duplicated and added to the universe.

**Comment by Vezzra** [comment]
> Equivalent log lines for other operating systems would be interesting to compare.

Hm, when I do a default quickstart game, I get the exact same lines, in the same order, but the line numbers are offset by 1: lines 103-118.

Complete server log: [freeoriond.log.zip](https://github.com/freeorion/freeorion/files/885054/freeoriond.log.zip)

**Comment by geoffthemedio** [comment]
Seems like it's working OK so far then.

Offset line numbers are probably related to an extra unrelated line earlier in the log.

**Comment by geoffthemedio** [comment]
Now partly checking the HullTypes, which involves processing some floating-point variables...
````
2017-03-31 12:50:52.781691 [debug] Server : ShipDesign.cpp:586 : HullTypeManager checksum: 6035456
2017-03-31 12:50:53.287062 [debug] Server : ShipDesign.cpp:1211 : PredefinedShipDesignManager checksum: 300305
````

**Comment by Vezzra** [comment]
Got the same checksums on OSX:
```
2017-03-31 15:17:00.108447 [debug] Server : ShipDesign.cpp:586 : HullTypeManager checksum: 6035456
2017-03-31 15:17:00.528571 [debug] Server : ShipDesign.cpp:1211 : PredefinedShipDesignManager checksum: 300305
```

**Comment by geoffthemedio** [comment]
Input from someone on Linux about why the template stuff is failing would be appreciated / helpful. Works fine in Windows / MSVC.

**Comment by LGM-Doyle** [comment]
`gcc` wants the order changed so that the `char` template is defined before it is defining `std::string` and looking for the char.

I've included a patch that compiles for me.

````patch
diff --git a/universe/ShipDesign.cpp b/universe/ShipDesign.cpp
index b2ebbf2..7102654 100644
--- a/universe/ShipDesign.cpp
+++ b/universe/ShipDesign.cpp
@@ -88,38 +88,6 @@ namespace {
 
     const unsigned int CHECKSUM_MODULUS = 10000000U;    // reasonably big number that should be well below UINT_MAX, which is ~4.29x10^9 for 32 bit unsigned int
 
-    // applies to pairs (including map value types)
-    template <class C, class D>
-    void CheckSumCombine(unsigned int& sum, const std::pair<C, D>& p)
-    {
-        std::cout << "CheckSumCombine(pair): " << typeid(p).name() << std::endl << std::endl;
-        CheckSumCombine(sum, p.first);
-        CheckSumCombine(sum, p.second);
-    }
-
-    // should apply to vectors, maps, and strings
-    template <class C>
-    void CheckSumCombine(unsigned int& sum, const C& c,
-                         const typename C::const_iterator* it = nullptr,
-                         const typename C::value_type* val = nullptr)
-    {
-        std::cout << "CheckSumCombine(C container): " << typeid(c).name() << std::endl << std::endl;
-        for (const typename C::value_type& t : c)
-            CheckSumCombine(sum, t);
-        sum += c.size();
-        sum %= CHECKSUM_MODULUS;
-    }
-
-    // applies to classes that have GetCheckSum methods
-    template <class C>
-    void CheckSumCombine(unsigned int& sum, C& c,
-                         decltype(((C*)nullptr)->GetCheckSum())* val = nullptr)
-    {
-        std::cout << "CheckSumCombine(C with GetCheckSum): " << typeid(c).name() << std::endl << std::endl;
-        sum += c.GetCheckSum();
-        sum %= CHECKSUM_MODULUS;
-    }
-
     // unsigned types (eg. bool, unsigned int, unsigned long int, unsigned char)
     template <class T, typename std::enable_if<std::is_unsigned<T>::value, T>::type* = nullptr>
     void CheckSumCombine(unsigned int& sum, const T& t) {
@@ -157,6 +125,38 @@ namespace {
     void CheckSumCombine(...)
     { std::cout << "CheckSumCombine(...)" << std::endl << std::endl; }
 
+    // applies to pairs (including map value types)
+    template <class C, class D>
+    void CheckSumCombine(unsigned int& sum, const std::pair<C, D>& p)
+    {
+        std::cout << "CheckSumCombine(pair): " << typeid(p).name() << std::endl << std::endl;
+        CheckSumCombine(sum, p.first);
+        CheckSumCombine(sum, p.second);
+    }
+
+    // should apply to vectors, maps, and strings
+    template <class C>
+    void CheckSumCombine(unsigned int& sum, const C& c,
+                         const typename C::const_iterator* it = nullptr,
+                         const typename C::value_type* val = nullptr)
+    {
+        std::cout << "CheckSumCombine(C container): " << typeid(c).name() << std::endl << std::endl;
+        for (const typename C::value_type& t : c)
+            CheckSumCombine(sum, t);
+        sum += c.size();
+        sum %= CHECKSUM_MODULUS;
+    }
+
+    // applies to classes that have GetCheckSum methods
+    template <class C>
+    void CheckSumCombine(unsigned int& sum, C& c,
+                         decltype(((C*)nullptr)->GetCheckSum())* val = nullptr)
+    {
+        std::cout << "CheckSumCombine(C with GetCheckSum): " << typeid(c).name() << std::endl << std::endl;
+        sum += c.GetCheckSum();
+        sum %= CHECKSUM_MODULUS;
+    }
+
     // doubles
     void CheckSumCombine(unsigned int& sum, const double& t) {
         std::cout << "CheckSumCombine(double): " << typeid(t).name() << std::endl << std::endl;
````

**Comment by geoffthemedio** [comment]
Is it enough to move the `// should apply to vectors, maps, and strings` case to immediately after the `// signed types (eg. int, char) but not float or double which are covered by specialized functions` case?

**Comment by geoffthemedio** [comment]
guess I'll find out in ~35 min...

edit: no

**Comment by geoffthemedio** [comment]
@Vezzra could you update the Xcode project for this branch?

**Comment by Vezzra** [comment]
I'm getting the following compile errors on OSX (unless these are addressed, I can't test the required changes to the Xcode project):
```
/Users/user/SoftwareProjekte/FO/src/universe/Species.cpp:52:40: error: cannot pass object of non-trivial type 'const std::shared_ptr<const Condition::ConditionBase>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
    CheckSums::CheckSumCombine(retval, m_location);
                                       ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Species.cpp:8:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:65:34: error: cannot pass object of non-trivial type 'const typename vector<FocusType, allocator<FocusType> >::value_type' (aka 'const FocusType') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
            CheckSumCombine(sum, t);
                                 ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Species.cpp:8:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:65:34: error: cannot pass object of non-trivial type 'const typename vector<shared_ptr<EffectsGroup>, allocator<shared_ptr<EffectsGroup> > >::value_type' (aka 'const std::__1::shared_ptr<Effect::EffectsGroup>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
            CheckSumCombine(sum, t);
                                 ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Species.cpp:8:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Building.cpp:9:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Building.cpp:9:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:65:34: error: cannot pass object of non-trivial type 'const typename vector<shared_ptr<EffectsGroup>, allocator<shared_ptr<EffectsGroup> > >::value_type' (aka 'const std::__1::shared_ptr<Effect::EffectsGroup>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
            CheckSumCombine(sum, t);
                                 ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Building.cpp:9:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ShipDesign.cpp:6:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ShipDesign.cpp:6:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ShipDesign.cpp:6:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:65:34: error: cannot pass object of non-trivial type 'const typename vector<shared_ptr<EffectsGroup>, allocator<shared_ptr<EffectsGroup> > >::value_type' (aka 'const std::__1::shared_ptr<Effect::EffectsGroup>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
            CheckSumCombine(sum, t);
                                 ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ShipDesign.cpp:6:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ShipDesign.cpp:6:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Special.cpp:6:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Special.cpp:6:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:65:34: error: cannot pass object of non-trivial type 'const typename vector<shared_ptr<EffectsGroup>, allocator<shared_ptr<EffectsGroup> > >::value_type' (aka 'const std::__1::shared_ptr<Effect::EffectsGroup>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
            CheckSumCombine(sum, t);
                                 ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Tech.cpp:10:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:65:34: error: cannot pass object of non-trivial type 'const typename vector<shared_ptr<EffectsGroup>, allocator<shared_ptr<EffectsGroup> > >::value_type' (aka 'const std::__1::shared_ptr<Effect::EffectsGroup>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
            CheckSumCombine(sum, t);
                                 ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Tech.cpp:10:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:65:34: error: cannot pass object of non-trivial type 'const typename vector<ItemSpec, allocator<ItemSpec> >::value_type' (aka 'const ItemSpec') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
            CheckSumCombine(sum, t);
                                 ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Tech.cpp:10:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
/Users/user/SoftwareProjekte/FO/src/util/CheckSums.cpp:12:16: error: use of undeclared identifier 'DBL_MAX_10_EXP'
        assert(DBL_MAX_10_EXP < 400);
               ^
/Users/user/SoftwareProjekte/FO/src/util/CheckSums.cpp:25:16: error: use of undeclared identifier 'FLT_MAX_10_EXP'
        assert(FLT_MAX_10_EXP < 40);
               ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Field.cpp:9:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:65:34: error: cannot pass object of non-trivial type 'const typename vector<shared_ptr<EffectsGroup>, allocator<shared_ptr<EffectsGroup> > >::value_type' (aka 'const std::__1::shared_ptr<Effect::EffectsGroup>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
            CheckSumCombine(sum, t);
                                 ^
In file included from /Users/user/SoftwareProjekte/FO/src/universe/Field.cpp:9:
In file included from /Users/user/SoftwareProjekte/FO/src/universe/ValueRef.h:8:
/Users/user/SoftwareProjekte/FO/src/universe/../util/CheckSums.h:53:32: error: cannot pass object of non-trivial type 'const std::__1::basic_string<char>' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        CheckSumCombine(sum, p.first);
                               ^
```

**Comment by Vezzra** [comment]
I just pushed a fix for a broken static library reference in the Xcode project to master, so in order to avoid merge conflicts between the changed Xcode project in master and the changes to it I'll add here, please rebase.

**Comment by geoffthemedio** [comment]
master build is currently broken, so am waiting for a fix for that...

**Comment by geoffthemedio** [comment]
rebased

**Comment by Vezzra** [comment]
I pushed a fix for the `use of undeclared identifier` compile errors to the branch (apparently you need to include `float.h` in `CheckSums.cpp`, otherwise the `DBL_MAX_10_EXP` and `FLT_MAX_10_EXP` don't get declared).

However, I'm also still getting these `cannot pass object of non-trivial type XYZ through variadic function; call will abort at runtime` compile errors. Apparently the clang compiler (at least the version on my system) can't handle your use of the `void CheckSumCombine(...)` variadic function.

See also the build log of the failed Travis CI builds.

**Comment by geoffthemedio** [comment]
Removed the ... form. Hopefully works now?

**Comment by Vezzra** [comment]
Trying to build 3604eac7440801eefe4c4746d77dce8dd10ff59d gives me tons of compile errors, on OSX as well as on Linux.

Build logs: [buildlogs.zip](https://github.com/freeorion/freeorion/files/992947/buildlogs.zip)

So, still can't test the updated Xcode project...

**Comment by geoffthemedio** [comment]
The travis build log suffices for testing the OSX build in this case.

Why it doesn't find the various declared template functions, I don't know, though.

**Comment by Vezzra** [comment]
I can push the adjustments to the Xcode project without testing, if that helps...

**Comment by geoffthemedio** [comment]
if it's adjusted for unrelated (to this PR) changes, then sure.

**Comment by Vezzra** [comment]
No, I was referring to the the updates required for this branch, as you requested above:

> @Vezzra could you update the Xcode project for this branch?

**Comment by geoffthemedio** [comment]
I thought you already did that...

Sure then, push those as well.

**Comment by geoffthemedio** [comment]
I'm about out of ideas for what else to try to get GCC to accept what seems like it should be correct template code...

File: universe/Effect.h [hunk 4]
```diff
@@ -1123,7 +1123,7 @@ class FO_COMMON_API GenerateSitRepMessage : public EffectBase {
     const std::string& Icon() const
     { return m_icon; }
 
-    const auto& MessageParameters() const
+    auto MessageParameters() -> const std::vector<std::pair<std::string, ValueRef::ValueRefBase<std::string>*>>&
     { return m_message_parameters; }
```
LGM-Doyle: @geoffthemedio `decltype(auto)` is a [C++14 feature](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3638.html).  Are you pushing us to migrate to C++14 ASAP 😉 

Does VC++ default to C++14?  We have explicitly restricted gcc to C++11 features in the CMakeList.txt.

Upgrading to C++14 requires no changes for our minimum required compiler version for VC++ (2015/ MSC_VER = 1900) and gcc (5.0). 

I can't tell what minimum version that we require for clang.  I think that upgrading to C++14 requires a change from clang 3.3 to 3.4. The release dates for those versions are 17 June 2013 and 2 January 2014 respectively.

geoffthemedio: I just wanted to be rid of the unnecessary and long return type, and my first attempt with just auto as the return type didnt work in GCC. The error message led to me trying this syntax.

geoffthemedio: I dont think MSVC defaults to any particular standard. Each release just had various features from the standards supported.

LGM-Doyle: That is difficult and unfortunate.  I'd assumed that with a simple setting fix, you could save yourself the wait on the build bots.

geoffthemedio: The bots are doing more than validating C++ standard versions... A lot of the issues in this pull request / branch have been, I think based on googling the issues or error messages, bugs or missing features in GCC, not standard mismatches. It also checks other inconsistencies like what needs API specifications.

Vezzra:
Builds and runs ok now on OSX.

universe/Effect.h
LGM-Doyle: My original comment was to point out that the single line of code was not C++11 and would not compile with the `--std=c++11` switch in either clang/gcc.  I was unaware that VC++ did not have a switch to toggle standards and assumed that you had forgotten to change back to C++11 mode.

I checked yesterday and the codebase compiles and runs under C++14 for clang and gcc on linux.  This PR did not successfully compile on either.

If we try the C++ version change on OSX and it works, then I suggest that we upgrade to C++14.  It will make the behavior of VC++ and the the other two compilers more similar.

geoffthemedio: Apparently the latest update to MSVC2015 adds such a switch for c++14 (but not 11 as far as I can see): https://blogs.msdn.microsoft.com/vcblog/2016/06/07/standards-version-switches-in-the-compiler/

However, the C++14 support in 2015 is incomplete, though possibly acceptably so:
https://blogs.msdn.microsoft.com/vcblog/2017/05/10/c17-features-in-vs-2017-3/

geoffthemedio:
Any problems running with this applied?

LGM-Doyle:
It compiled and ran correctly.  

It crashed later in the game when the `CombatLogManager` received an unrequested log, reported the error and then proceeded to crash.  I fixed the crash in master, but I have not determined what caused it to send an incorrect log in the first place.  

I thought that this PR only affected startup, so this is possibly an unrelated bug.

geoffthemedio:
It also sends some messages when joining games with checksum info. Don't think that would have any impact on combat logs being sent, though.

PR #1447, opened 2017-03-28
Do not issue next fleet order if previous one could not be issued

Morlic-fo:
Seems to partially fix an issue reported in http://freeorion.org/forum/viewtopic.php?f=28&t=10483#p88044

@Dilvish-fo: It seems to me that we would always want to break here if we can't issue an order but I can't tell if there was a reason not to introduce the ```break``` statement. If there are any negative side effects, we should discuss and adress those before merge.

### DIFF
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -393,6 +393,7 @@ class AIFleetMission(object):
                             self.clear_fleet_orders()
                             self.clear_target()
                             return
+                break  # do not order the next order until this one is finished.
         else:  # went through entire order list
             if order_completed:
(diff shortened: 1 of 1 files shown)


Dilvish-fo (approved):
This is probably best, at least for now, since I can't think of any current situations where we would want to try executing a later order after the first one failed, so I've approved it.

I think the code had been intended to allow for something vaguely along the lines of an invasion order followed by a bombard order (so if the invasion order can't be issued because shields are still up, the bombardment could happen).  I say vaguely like that because when this structure was put into place we didn't  have bombardment actually implemented.

Vezzra:
So, LGTM?

Issue #1177, opened 2016-12-16
Sidepanel stat icons flickers during turn updates


Vezzra:
### Environment

* **FreeOrion Version**: ~~eaf735e20cf50767a2ade2060543ca68d4d8affc~~ 853af0ea037ff01e0a63645984cebf0228f81a5a
* **Fetched as**
  * Weekly development build

### Description

~~When executing a turn, the stealth stat icon in the sidepanel does seem to flicker (go black) briefly (perhaps a frame), in the case where all the sub-panels are collapsed.~~

When executing a turn, any stat icon in the sidepanel that is too close to the arrow icons to the right does seem to flicker (go black) briefly (perhaps a frame). Once you resize the sidepanel so that there is a sufficient gap between the stat icon(s) and the arrow icon(s), the flickering is gone.
### Expected Result

Well, no flickering of course. 😃

LGM-Doyle:
I can't see the flicker anymore.  I compared eaf735e and master and the flicker seems gone.

Vezzra:
Nope, still there, however, I now discovered the real issue: it's not so much the stealth icon per se, it's any icon that is too close to the arrow icons to the right. Once you resize the sidepanel so that there is a sufficient gap between the stat icon(s) and the arrow icon(s), the flichering is gone. If a stat icon is too close to the arrow icon to the right of it, it flickers.

Edited title and OP accordingly.

Vezzra:
So maybe you don't notice it because your sidepanel is sized so wide the gap between the stealth icon and the arrow icon is wide enough?

LGM-Doyle:
@Vezzra thanks for triaging this issue.

Vezzra:
Fixed by #1448.

PR #1448, opened 2017-03-29
Fix bug #1177 Sidepanel stat icons flickers during turn updates.

LGM-Doyle:
This fixes issue #1177.

The underlying problem was that the `StatisticIcons` for the Military, Population and Resource panels were initialized with zeros instead of the correct value, and then immediately rendered.

This PR:
- uses the correct initial values for the panel icons.
- adds `StatisticIcon::MinUsableSize()` so that the icons can be clipped as tightly as possible
- adds `PreRender()` to the panels so that the icons are only layed out once.
- moves the rendering, positioning and layering of the `AccordionPanel` expand button into `AccordionPanel`

### DIFF
--- a/UI/MilitaryPanel.cpp
+++ b/UI/MilitaryPanel.cpp
@@ -238,8 +238,6 @@ void MilitaryPanel::DoLayout() {
         Resize(GG::Pt(Width(), m_multi_meter_status_bar->Bottom() + EDGE_PAD - top));
     }
 
-    m_expand_button->MoveTo(GG::Pt(Width() - m_expand_button->Width(), GG::Y0));
-
     SetCollapsed(!s_expanded_map[m_planet_id]);
 }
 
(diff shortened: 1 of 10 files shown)


Vezzra (approved):
Yep, fixed the problem, the flickering is gone (tested on OSX). Can't say much about the code, looks fine to me.

Vezzra:
Corresponding issue #1177 is assigned to "optional for 0.4.7" milestone, so should a PR addressing it.

UI/MilitaryPanel.cpp
```diff
@@ -41,16 +41,27 @@ MilitaryPanel::MilitaryPanel(GG::X w, int planet_id) :
...
 
+    const auto obj = GetUniverseObject(m_planet_id);
+    if (!obj) {
+        ErrorLogger() << "Invalid object id " << m_planet_id;
+        return;
+    }
+
     // small meter indicators - for use when panel is collapsed
-    m_meter_stats.push_back({METER_SHIELD, new StatisticIcon(ClientUI::MeterIcon(METER_SHIELD), 0, 3, false,
+    m_meter_stats.push_back({METER_SHIELD, new StatisticIcon(ClientUI::MeterIcon(METER_SHIELD),
```
geoffthemedio, line 51: are these changes related to the point of this pull request?

LGM-Doyle: The values that the icon was flickering between were the initial incorrect default value and the correct value.  These lines initialize the icons with the correct value.

geoffthemedio:
haven't tested myself, but no objections

Issue #996, opened 2016-09-27
Weapon refinement explanations and species effects unclear


MatGB:
### Description

Weapon refinement techs need to be clearer about the expected results and mention further modifiers from effect such as species.

Hull design Pedia articles need to clearly say stats are based on your current empire tech and may be further modified by species, etc.
### Expected Result

Improved explanations to prevent confusion for new players and allow further modifiers to be explained and integrated clearly. Discussion here: [Post](http://www.freeorion.org/forum/viewtopic.php?p=85701#p85701)

Vezzra:
Everyone, but especially @MatGB: status report? Any progress on this? Unless already done or close to be done I intend to reassign this to the "optional for 0.4.7" milestone, as I don't want to hold up the release for it.

Vezzra:
No objections so far, so I'm going to reassign.

memem359:
Started a [new discussion thread](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10484) about altering the Pedia weapon text.

I didn't touch the Hull entries, because it wasn't clear to me what needed adjustment.

PR #1449, opened 2017-03-29
Reword Pedia weapon text

memem359:
Aims to fix the [issue 996](https://github.com/freeorion/freeorion/issues/996).

Try to make clearer what happens with ship direct fire weapons, with the upgrading and the species pilot effect.

Only touched the en.txt file, so that can be dropped into an existing stringtables directory without recompiling the code.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -8000,7 +8000,7 @@ Direct Weapon
 PC_DIRECT_WEAPON_DESC
 '''Direct-fire weapons attack and can damage targets each combat round (unlike [[encyclopedia FIGHTER_TECHS]], which must first be launched before they can attack).
 
-Some weapons in this category are capable of firing multiple shots per combat round.'''
+Some weapons in this category are capable of firing multiple shots per combat round. Ships built by species with the piloting trait can affect the weapon damage.'''
 
 # Ship part classes
 PC_FIGHTER_HANGAR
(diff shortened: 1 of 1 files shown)


geoffthemedio:
proposed changes are overly verbose and sometimes misleading

might help to break unrelated sentences into separate paragraphs.

adrianbroher:
Also this PR contains a merge of master into the topic branch, which is unconditionally a reason for rejecting this PR.

memem359:
It is more verbose than what was there, but I don't agree with overly or the misleading.
Can you give some examples, so I can see what you mean?
Either here, or in the forum [http://www.freeorion.org/forum/viewtopic.php?f=28&t=10484](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10484)

It was my understanding that new players were getting confused with how species piloting worked and how ship upgrades worked. And at least one person said it was unwieldy to see "damage increased by X" and then try to find the link to what the original damage was. If the text seems repetitive, that is to make each entry more understandable on its own.

adrianbroher: I don't understand what you just said.
This is my first attempt at a FO pull request. Every other PR that I've seen also does the merge of master. What do you mean by topic branch?

geoffthemedio:
getting an upgrade within supply is unclear about whether the effect is only active while in supply and won't work when not in supply

referring to "one level" or similar of weapon damage, and referring to specific amounts of damage is confusing

referring to specific amounts of damage, and then referring to other effects in a separate sentence is confusing. instead, say "to 5 (before species effects)" or somesuch, or just accept that a single source of a bonus can't tell you what the total result of itself and other potential bonuses will be, and just say "+2" or similar.

"will increase (decrease) shot damage by 2 per bonus (malus) level" is very unclear

re: merging master, look again. (almost?) none of the freeorion pull requests include a commit like "Merge remote-tracking branch 'origin/master' into weapon-text"

adrianbroher:
> Every other PR that I've seen also does the merge of master.

Yes, we merge the pull request into master.  We reject pull requests which merged master (it's the opposite direction).

> What do you mean by topic branch?

A branch which does implement a topic like the branch your pull request is based on (here: update some pedia text).  Opposing to that there are the long lived development branches (e.g. master) and the release branches.

memem359:
Ahah... I was playing around with the GitHub repositories, and wanted to make sure I had the latest "master", so that is why it has the remote-merging. When I saw there were no differences other than the en.txt file, I figured it was safe... I'm guessing I should go back to the master fork, make the branch, and then copy the changes to there.

Will this require a brand new pull request, since I need to reorganize my branches/forks?

geoffthemedio: That helps a lot.
For the supply, I'll have to think about how to state that upgrade is permanent, once it happens. The upgrade doesn't happen unless the ship is in range, so I wanted to emphasize that.

Let me ask this... are there other modifiers that are in the game now (or in the near future) that affect weapon damage, other than piloting? A lot of that wording was to keep the door open for more than one modifier. (Originally, I had "will increase (decrease) shot damage by 2 per pilot level".)

memem359:
Since I need to start from a clean GitHub branch (adrianbroher) and have more work to do on the text itself (geoffthemedio), I'll delete the branch and close this PR. I'll submit a new one either later this week or early next week.

dbenage-cx:
(since closed, but will post in case it is of help in future.  If you have further questions on git/github, might post to http://www.freeorion.org/forum/viewtopic.php?f=9&t=9366 )

> Will this require a brand new pull request, since I need to reorganize my branches/forks?

Not necessarily, you can use `git rebase -i` for interactive rebase, and drop the merge commit.
If you need upstream changes: `git fetch upstream && git rebase upstream/master`.
(assuming you have `git remote add upstream https://github.com/freeorion/freeorion.git`)

> are there other modifiers that are in the game now (or in the near future) that affect weapon damage

Solar concentrator ship part is only one that comes to mind currently, but good possibility of other effects as development progresses.

Vezzra:
Superseded by #1451.

PR #1450, opened 2017-03-30
Switch reductions in tech costs to reductions in tech research time when conditions are met

geoffthemedio:
An experiment. I don't have a great motivational explanation, but it seems like reducing the minimum turns, instead of the cost of researching stuff, when conditions are met, makes sense. Letting the player focus on the "boosted" tech instead of spreading the RP over more techs seems more interesting.

Particular numbers are work in progress, documentation needs updating.

Also makes the scripts a bit simpler.

### DIFF
--- a/default/scripting/techs/learning/PSIONICS.focs.txt
+++ b/default/scripting/techs/learning/PSIONICS.focs.txt
@@ -3,11 +3,11 @@ Tech
     description = "LRN_PSIONICS_DESC"
     short_description = "THEORY_SHORT_DESC"
     category = "LEARNING_CATEGORY"
-    researchcost = 300 * [[TECH_COST_MULTIPLIER]] - (250 * [[TECH_COST_MULTIPLIER]] * Statistic If condition = AND [
+    researchcost = 300 * [[TECH_COST_MULTIPLIER]]
+    researchturns = 8 - (4 * Statistic If condition = AND [
         OwnedBy empire = Source.Owner
         HasTag name = "TELEPATHIC"
     ])
-    researchturns = 4
     tags = [ "PEDIA_LEARNING_CATEGORY" "THEORY" ]
     prerequisites = "LRN_TRANSLING_THT"
     graphic = "icons/tech/psionics.png"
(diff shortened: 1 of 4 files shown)


MatGB:
K, I think I like the idea of doing this for some stuff, but not for these specific techs, *especially* not Psionics unless we substantially cut the base cost as well.

For the stealth techs, I created them when I split up the parts from the planetary stealth and deliberately reduced the costs of one by the cost of the other but didn't otherwise change them: I did test giving similar reductions both ways and thought I'd committed that but it appears I eitehr scrapped that or messed up somewhere. It was a deliberate choice to do it this way to keep the total cost the same, I do think all the tech costs are in need of balancing but these aren't too bad: plus, I'd be loath to introduce a way to speed up planetary stealth improvements given they can already be a massive pain.

For Psionics as a specific, the current mechanic works well and the AI makes use of it, I'm uncertain of the current costs, it's probably too expensive base cost and too big a reduction but I do know that if an empire can afford 600 RP quickly it basically doesn't need Psionics, the main purpose of which is gating Distributed Thought Computing, which gives a deliberate cheap research speedup to some species and not others (doing this would *really* hurt Laenfa and George, the latter really does need this boost or something else similar).

adrianbroher:
Did this ever came to a conclusion? This branch is now over a year old.

hifron:
Human thoughts are slow and there would be some speedup, then could be maybe also tech turn reduction. But some thoughts are made by inspiration, but artificial thoughts the way the some human personalities think could speed research in some areas. Artificial thoughts may be bionic or some way learnable. Therefore some speedup with research time is maybe possible, but that doesn't necessary means there is also tech cost reduction.

If some species could not adapt or hurt them, why they don't have some advantage or could gain some advantage?

o01eg:
@geoffthemedio Is it going to be done and merged?

geoffthemedio:
I don't have any plans related to this... Probably needs a design discussion on the forums, and this specific pull request is not really needed / relevant.

PR #1451, opened 2017-03-30
Pedia text for direct weapons

memem359:
This is the second attempt to address Issue #996 
(First attempt was my botched PR #1449 , which was closed.)

Adjusted the en.txt file to add information about the effects of weapon refinement techs and species piloting, which are not clear in the current Pedia.

Forum discussion here: [http://www.freeorion.org/forum/viewtopic.php?f=28&t=10484](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10484)

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -8000,7 +8000,7 @@ Direct Weapon
 PC_DIRECT_WEAPON_DESC
 '''Direct-fire weapons attack and can damage targets each combat round (unlike [[encyclopedia FIGHTER_TECHS]], which must first be launched before they can attack).
 
-Some weapons in this category are capable of firing multiple shots per combat round.'''
+Some weapons in this category are capable of firing multiple shots per combat round. Ships built by species with the piloting trait can affect the weapon damage.'''
 
 # Ship part classes
 PC_FIGHTER_HANGAR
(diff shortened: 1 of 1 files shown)


default/stringtables/en.txt
```diff
@@ -10448,91 +10448,92 @@ SHP_WEAPON_1_2
 Mass Driver 2
 
 SHP_WEAPON_1_2_DESC
-Improves Mass Driver by one level, increasing strength by 1.
+Upgrades [[shippart SR_WEAPON_1_1]] parts on all ships within [[metertype METER_SUPPLY]], increasing shot [[encyclopedia DAMAGE_TITLE]] to 4.
```
dbenage-cx: Think `increasing by X` is preferrable to `increasing to X`, the later is not always correct.

default/stringtables/en.txt
```diff
@@ -10448,91 +10448,92 @@ SHP_WEAPON_1_2
...
 SHP_WEAPON_1_3_DESC
-Improves Mass Driver by two levels, increasing strength by 2.
+Upgrades [[shippart SR_WEAPON_1_1]] parts on all ships within [[metertype METER_SUPPLY]], increasing shot [[encyclopedia DAMAGE_TITLE]] to 5.
 
 SHP_WEAPON_1_4
 Mass Driver 4
 
 SHP_WEAPON_1_4_DESC
-Improves Mass Driver by three levels, increasing strength by 3.
+The ultimate refinement of the Mass Driver. Upgrades [[shippart SR_WEAPON_1_1]] parts on all ships within [[metertype METER_SUPPLY]], increasing shot [[encyclopedia DAMAGE_TITLE]] to 6.
```
dbenage-cx: Further improvements of (at least) mass drivers have been discussed, might drop the use of `ultimate` or any hints at a final refinement.

default/stringtables/en.txt
```diff
@@ -11743,31 +11744,43 @@ SR_WEAPON_0_1
...
+Each flak cannon on a ship built by a species with the piloting trait will have an additional 1 shot per piloting level (-1 shot for Bad Pilots).'''
 
 SR_WEAPON_1_1
 Mass Driver
 
 SR_WEAPON_1_1_DESC
-The Mass Driver, a basic weapon.
+'''The Mass Driver, a basic weapon. Upgrading the weapon technology will increase the shot damage.
+
+Each mass driver on a ship built by a species with the piloting trait will do an additional 1 shot damage per piloting level (-1 for Bad Pilots).'''
```
dbenage-cx: Comparing side by side with Flak Cannon, the difference between `1 shot` and `1 shot damage` is easily missed.
Suggest mentions of shots change to "number of shots" and damage to "damage per shot", or similar.

dbenage-cx (changes requested):

default/stringtables/en.txt
```diff
@@ -10448,91 +10448,92 @@ SHP_WEAPON_1_2
...
 SHP_WEAPON_3_4_DESC
-Improves Plasma Cannon three levels increasing base strength by 9.
+The ultimate refinement of the Plasma Cannon. Upgrades [[shippart SR_WEAPON_3_1]] parts on all ships within [[metertype METER_SUPPLY]], increasing shot [[encyclopedia DAMAGE_TITLE]] to 18.
 
 SHP_WEAPON_4_1
 Death Rays
 
 SHP_WEAPON_4_1_DESC
-Unlocks the Death Ray, the ultimate ship's weapon.
+Unlocks the [[shippart SR_WEAPON_4_1]], a more powerful ship's weapon than the [[shippart SR_WEAPON_3_1]]. The ultimate ship's weapon.
```
dbenage-cx: Realize this was from original, I'd prefer to drop use of ultimate describing this weapon, since it is surpassed by spinal cannon or flak cannon, depending on use case.

default/stringtables/en.txt
memem359: How about for FlakC: "will increase the number of shots by 1 per piloting level", and leave shot damage as is?

default/stringtables/en.txt
memem359: No problemo on the Death Ray not being on top.

default/stringtables/en.txt
memem359: I thought it might be worth it to let newbies know that they reached the current peak for that particular weapon, but no problem to drop it.

default/stringtables/en.txt
memem359: On second thought, I'll change "shot damage" to "damage" for the piloting effect. Since I state that it applies to each mass driver, we don't need to say "per shot". And since it is affecting "base" or "weapon" damage, don't want to get that confused with the "shot damage".

default/stringtables/en.txt
memem359: This is the only comment where I'd like a bit of a debate before making changes.

Can you give an example where the later is not correct?
If you are referring to the species piloting, the ship part description adds that information (that there may be additional modifiers so that "base" damage can be different than "shot" damage). If you think it would help, more text could be added to the ship part description, maybe something like "The weapon's damage is a combination of the shot damage and other modifiers" and then the sentence about the pilot trait.

The ship part lists "shot damage", which is always the value from level 1 of the weapon.
Saying that "increasing shot damage to X" matches up with the ship part description.

"Increasing by X" might be confusing in multiple ways.
One user said they couldn't always remember the initial number, so saying that it increases by X made him have to do extra work to get the number he wanted.
If Laser4 increases by 6, is that 6 more than Laser3, or 6 more than Laser1? Giving an absolute number removes that ambiguity.

dbenage-cx: Fair enough, possibly state the value is a base value, and may be altered by other effects.

Rereading this: `on all ships within Supply` might lead a player to think it is a one time effect.
Maybe add a sentence that a ship receives upgrades once it reaches supply range.

Piloting trait and solar concentrator are only effects I can think of that might effect damage per shot/number of shots currently.

dbenage-cx:
Might [squash/fixup](https://help.github.com/articles/using-git-rebase-on-the-command-line/) commits, but can be done with merge (which I'll do tonight if no other comments).

dbenage-cx (approved):

MatGB:
Thank you for doing this @memem359 , my last attempt was so awful I scrapped it entirely, for such a simple mechanic it's surprisingly hard to write clear, concise descriptions.

PR #1452, opened 2017-03-30
Deleted contents of default/data/art/combat 

geoffthemedio:
I suspect all these files are not used now, since the 3D combat system was removed.


dbenage-cx:
Do not see any errors after starting new game or code mentions, LGTM.

Vezzra (approved):
Builds and runs ok on OSX.

geoffthemedio:
guess we'll find out then...

geoffthemedio:
note that this cut out about 45 MB of data (~ 33% of default/data/art )

Vezzra:
> guess we'll find out then...

No risk, no fun... 😉

MatGB:
I don't think I've ever seen most of this stuff, probably ought to have delved into the unused folders a lot earlier. Should we put copies into the assets repo in case we find a use for any of it in future?

geoffthemedio:
It's all still in the git and svn histories...

TheSilentOne1:
There are a couple of currently unused planet textures which we could put in the game.
Also there are some 3D models by pd, but they're already in the asset repository.
There is a skybox which we probably won't use since there won't be 3D combat.
And lastly, there are some star graphics for the 3D combat star representation. They might be useful at some point, but as Geoff points out, they're still accessible.
I'll try out the planet textures and create a PR if they're good(looking on the sidepanel).

adrianbroher:
> Should we put copies into the assets repo in case we find a use for any of it in future?

The asset repo is for asset sources, so no.  I tried to reverse engineer some of the files into something editable but I never came around to clean those up.

PR #1455, opened 2017-04-01
Create Xcode project with CMake

adrianbroher:
This PR enables generating a Xcode project with CMake.  So far the project compiles and runs as far as I tested, but is not relocatable thanks to some nastiness regarding the RPATH for the FreeOrion binaries and several other minor issues but should be consider a Work In Progress.  Also TravisCI picks up the system pythonlib, while my local installation picks the pythonlib within the SDK, which needs to be fixed.

Feedback for the layout of the generated Xcode project would be appreciated (are all files you care about are listed in the project, are they at the right place, etc).

If someone wants to test this out please keep the following things in mind:
* You will need CMake 3.4 or later to generate the build system on MacOSX.
* Use the CMake `-G` parameter to select the `Xcode` generator, otherwise you will get a plain `Makefile` project.
* Have FreeOrionsSDK v7 installed.  The last versions contained various bug fixes that prevented the usage of CMake on MacOSX.

Currently this branch only builds on TravisCI with `CMake` and the `Xcode` generator only.  When all breaking bugs are fixed this will be undone.


adrianbroher:
The remaining issues were easier to fix than I had anticipated.  The build result should now be equivalent to the manual maintained Xcode project.  However the generated Xcode project file structure may need to be tweaked but I leave the assessment of that to @Vezzra.

Vezzra:
When I run `cmake -G Xcode` within the source dir and try to build the generated Xcode project, I get this link error for target `GiGi`:
```
Undefined symbols for architecture i386:
  "_fchmodat", referenced from:
      boost::filesystem::detail::permissions(boost::filesystem::path const&, boost::filesystem::perms, boost::system::error_code*) in libboost_filesystem.a(operations.o)
ld: symbol(s) not found for architecture i386
```

adrianbroher:
> When I run cmake -G Xcode within the source dir

Running CMake as a in-source build is usually considered bad style for various reasons.  I should add a guard and some build instructions for this.

> Undefined symbols for architecture i386:  "_fchmodat"

This should have cropped up some time ago already, because that seems to be a Boost 1.59 bug  [boost::filesystem #10591]?  Do you run a MacOSX < 10.10? Maybe the manual maintained Xcode project covers this by not targeting 10.7?

[boost::filesystem #10591]: https://svn.boost.org/trac/boost/ticket/10591

Vezzra:
> Running CMake as a in-source build is usually considered bad style for various reasons.

I know, and my first try has been with a out-of-source build. Then I got that error, and started to wonder if maybe in this case an in-source build might be necessary and tried again with that. Got the error (as was to be expected) in both cases.

>  I should add a guard and some build instructions for this.

That's probably a good idea.

> Do you run a MacOSX < 10.10?

No... I'm on 10.10.5.

> Maybe the manual maintained Xcode project covers this by not targeting 10.7?

Well, the manually maintained Xcode project has 10.7 set as deployment target (not that this does us anything good, as successfully building and running on any OSX prior to 10.9 doesn't work anyway anymore), so no difference here to the generated one.

However, there is a difference in the `Base SDK` setting: the manually maintained Xcode project has `No SDK (latest OSX)` here, and IIRC I changed it to that because having `10.7` there gave me problems. The generated project has `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.7.sdk` for that setting. Which for one is kind of weird, because `10.7` should be sufficient (without the entire path), and on my systems resolves to "not found" anyway (don't know what it's using instead).

`No SDK (latest OSX)` is the default setting and it's probably best to leave it at that - you can't make assumptions about the SDKs people have installed on their systems anyway, and apparently it's not necessary to set the base SDK to the minimum OSX version you want to support, the `OSX Deployment Target" setting seems to be sufficient for that.

Changing the `Base SDK` setting to the default `No SDK (Latest OSX)` solves the link error, but then I get a lot of compile errors for the human client. No idea what's going on here, and don't have the time to investigate right now. Most likely after the release.

adrianbroher:
> Changing the Base SDK setting to the default No SDK (Latest OSX) solves the link error.

Because you're circumventing the deployment target setting here by selecting *No SDK (Latest OSX)*.  This explains why older OSX versions can't run the game because it uses a 10.10+ exclusive API via boost::filesystem.

LGM-Doyle:
@adrianbroher will this mean that when adding files to the project, non Xcode devs will not have to manually update the Xcode/FreeOrion.xcodeproj/project.pbxproj file?

adrianbroher:
@LGM-Doyle, yes that is a favourable side effect and one I really anticipate.  The main point however is a consistent build across all platforms in terms of build configuration which is currently not the case unfortunately.

Vezzra:
> Because you're circumventing the deployment target setting here by selecting No SDK (Latest OSX). This explains why older OSX versions can't run the game because it uses a 10.10+ exclusive API via boost::filesystem.

Don't know about the "circumventing" - it's the default setting after all. Anyway, I'd expect *not* to have to set the Base SDK to the min OSX version I want to support, that's what I have the deployment target setting for, after all. Otherwise, what's the point of having two separate settings, if I have to set them both to the same anyway?

And until we started supporting C++11, this worked perfectly fine. Base SDK set to `No SDK (Latest OSX)` and deployment target set to `10.7` produced builds that actually worked even on 10.6.

I now just got curious and tried setting the Base SDK to `10.7` for the manually maintained Xcode project. To my big surprise FO built and ran successfully??? Can't check if the resulting build also runs on OSX versions prior to 10.9 right now, will have to do that.

I'm... confused. 😕

adrianbroher:
> Don't know about the "circumventing" - it's the default setting after all. Anyway, I'd expect not to have to set the Base SDK to the min OSX version I want to support, that's what I have the deployment target setting for, after all. Otherwise, what's the point of having two separate settings, if I have to set them both to the same anyway?

To my understanding and research you give a range of supported OSX API versions with that. See this article about the topic: http://www.fantageek.com/blog/2016/01/26/sdk-and-deployment-target/  The article focuses on iOS, but I believe the same mechanic applies to OSX too.

> I'm... confused.

To be honest I tried to combine various possible potential bug issues but I can't give a conclusion either.  Especially because I'm using the same SDK and the same CMake generated project, but with a different MacOSX version (10.11) and probably a different Xcode version.  However my local build compiles, links and runs just fine without any settings changed.  Same goes for the unmodified manual project.

You can try the latest master of freeorion-sdk.  I have added a patch that should avoid the use of the function mentioned in the previous commit.  It was added to boost 1.60 but before rebuilding the SDK with a new boost version I would prefer to check if this actually helps.

Vezzra:
> To my understanding and research you give a range of supported OSX API versions with that.

Yep, the article you linked is pretty clear about that. Which means I better revert 3096ef7d3dc9bd7d6f51b0879b14803b25d04332, as that potentially creates more problems than it might solve. Better not require people to install some ancient SDK version.

> ...but with a different MacOSX version (10.11) and probably a different Xcode version

Ah yes, that might very well be the cause. I'm still using Xcode 6.4 on OSX 10.10.

> You can try the latest master of freeorion-sdk.

Will do.

memem359:
When I was trying to build from Xcode yesterday, it wouldn't start, and popped up the message "My Mac is not a valid target destination".
The recent revert fixed that.
I can't even guess what the right/best settings are, but I figured I should mention the problem my system had with commit [3096ef7](https://github.com/freeorion/freeorion/commit/3096ef7d3dc9bd7d6f51b0879b14803b25d04332)
(I'm using Xcode 8.2 on OSX 10.11)

Vezzra:
@memem359, thanks for the feedback.

Vezzra:
> You can try the latest master of freeorion-sdk. I have added a patch that should avoid the use of the function mentioned in the previous commit.

Didn't change anything. I'm still getting exactly the same link error...

adrianbroher:
Are you sure that you don't use any stray old FreeOrionSDK files?

Vezzra:
I can't think of anything I might have done wrong... I built the SDK from latest master (freeorion/freeorion-sdk@72150b81649a7cb87a0adee1879d68c465caf1cc), copied the contents of the generated `dep` archive over into my working copy (replacing the old `dep`), regenerated the build tree and tried to build the resulting Xcode project.

I'll give it another try asap.

Vezzra:
Looking at the patch for boost, the strange thing is, AFAICT, that this patch apparently is for OSX versions <10.10. But I *am* on 10.10, so this shouldn't happen anyway...?

adrianbroher:
> so this shouldn't happen anyway...?

Quote from boost::filesystem-#10591:

> The problem is no only with usage of iPhoneOS8.1.sdk, it is also affecting MacOSX10.10.sdk as in both AT_SYMLINK_NOFOLLOW is nowadays defined

> This means that fchmodat will only be available when *targeting* OS X 10.10+ or iOS 8.0+, but the AT_* constants are always defined when building *on* 10.10, so they cannot be used for detection whether fchmodat is available.

Boost::filesystem assumes `fchmodat` is part of the OSX API, even when you're targeting an OSX older than 10.10.  I assume I can't reproduce this issue locally because Apple fixed this in later SDK versions and AT_SYMLINK_NOFOLLOW isn't defined when targeting an less than 10.10 OSX version.

Vezzra:
Ah, crap, of course, the Xcode project generated by the cmake scripts has `10.7` as Base SDK, my bad. Which is why the error goes away when I manually set that to `No SDK (Latest OSX)`. But then I get this bunch of other errors...

I still have no idea why building the generated Xcode project gives so different results than the manually maintained one. I guess I'll have to painstakingly compare all the settings, differences in layout etc. Ugh. Well, it has to be done, but it will take me some time to get around to it, so I have to bid your patience...

adrianbroher:
> I still have no idea why building the generated Xcode project gives so different results than the manually maintained one.

The build now mostly matches the outcome on Linux, except with the platform specific adaptions for OSX like creating an Application bundle or link some targets with "-undefined dynamic_lookup".  Which is obvious, because the CMake build rules which were used on Linux until now are now applied to the OSX build.  But how does this matter as long as the build artifact works as intended?

>  I guess I'll have to painstakingly compare all the settings, differences in layout etc. Ugh. Well, it has to be done, but it will take me some time to get around to it, so I have to bid your patience...

Here, have some easy ones I have seen so far:

* Manual maintained Xcode Project (MM) has a extra static library for common FreeOrion client code, the CMake generated Xcode Project (CG) does not.
* MM builds libGiGi and libGiGiSDL as static library (and maybe freeorioncommon and freeorionparse too), CG creates shared libraries for all of those.
* MM enforces the deployment target only on freeorion, freeorionca, freeoriond, CG does this on every build artifact, however I don't know if this makes a difference or not.

Vezzra:
> But how does this matter as long as the build artifact works as intended?

*If* the build artifact works as intended, it doesn't matter at all. However, we're not there yet.

What I've been referring to is why building the MM works while the CG produces all kind of different errors, apparently depending on what I set as Base SDK and deployment target.

> Manual maintained Xcode Project (MM) has a extra static library for common FreeOrion client code, the CMake generated Xcode Project (CG) does not.

Hm, I'd expect that to actually reduce the potential for issues - one less library which needs to be linked against.

> MM builds libGiGi and libGiGiSDL as static library (and maybe freeorioncommon and freeorionparse too), CG creates shared libraries for all of those.

Could be it... after all, I get the link error fo libGiGi.

> MM enforces the deployment target only on freeorion, freeorionca, freeoriond, CG does this on every build artifact, however I don't know if this makes a difference or not.

This is strange - I checked the settings of the MM, and from what I can tell it also enforces the deployment target on all these build targets... actually, that setting is defined on the project level and inherited by all build targets. None of the targets overrides the global project setting.

Anyway, my recent test runs produced new results: after changing the deployment target both for the SDK and FO itself to 10.9, I rebuilt the SDK, regenerated and rebuilt the Xcode project with the updated SDK. On that try (with both Base SDK and deployment target set to 10.9) I still got the same errors.

But when I manually set the Base SDK to `No SDK (Latest OSX)` building the `ALL_BUILD` target succeeded for the first time. Don't ask my why that combo works, while e.g. Base SDK = `10.7` and deployment target `No SDK (Latest OSX)` does not.

However, that success now leads me to my next questions: what should I expect as the final build artefact(s)? What I got is the binaries (the 3 executables and the libs), nothing else. How do I build the app bundle?

adrianbroher:
> However, that success now leads me to my next questions: what should I expect as the final build artefact(s)?

You should have a directory named after the configuration (Release, Debug) you selected.  This directory should contain:

* freeorionca
* freeoriond
* libfreeorioncommon.dylib
* libfreeorionparse.dylib
* libGiGi.dylib
* libGiGiSDL.dylib

Those files are just build results and copied into the bundle after the *freeorion* target was built successfully.  When building the *freeorion* target in the same configuration directory a `FreeOrion.app` bundle should appear.  This should contain the whole application (after a successful build of course):

* Contents/Info.plist
* Contents/MacOS/FreeOrion # (the freeorion binary)
* Contents/Frameworks/
* Contents/Resources/main.nib
* Contents/Resources/FreeOrion.icns
* Contents/Resources/README.md
* Contents/Resources/default/
* Contents/SharedSupport/libfreeorioncommon.dylib
* Contents/SharedSupport/libfreeorionparse.dylib
* Contents/SharedSupport/libGiGi.dylib
* Contents/SharedSupport/libGiGiSDL.dylib

The *packaging* target should pack this bundle into a dmg file.

adrianbroher:
I did a new local build with the patched FreeOrionSDK and things still work.  Checking `libs/libboost_filesystem.a` with `nm` doesn't show any reference to `_fchmodat` anymore.

Vezzra:
Finally got back to this. Using the patched SDK (freeorion/freeorion-sdk@eaeae973f94ac6307a318ee6bf412d7b016e2cb3) and trying to build the `osx-cmake` branch with the Xcode project generated by `cmake -G Xcode` in an out-of-source build directory still gives me build errors:
```
Build target GiGi

Ld /Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGi.dylib normal i386
    cd /Users/user/SoftwareProjekte/FO/test/src
    export MACOSX_DEPLOYMENT_TARGET=10.9
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -arch i386 -dynamiclib -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -L/Users/user/SoftwareProjekte/FO/test/build/Debug -F/Users/user/SoftwareProjekte/FO/test/build/Debug -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk/System/Library/Frameworks -filelist /Users/user/SoftwareProjekte/FO/test/build/GG/FreeOrion.build/Debug/GiGi.build/Objects-normal/i386/GiGi.LinkFileList -install_name @rpath/libGiGi.dylib -Xlinker -rpath -Xlinker @loader_path/../SharedSupport -Xlinker -rpath -Xlinker @loader_path/../Frameworks -mmacosx-version-min=10.9 -stdlib=libc++ -dynamiclib -Wl,-headerpad_max_install_names /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_date_time.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_filesystem.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_regex.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_signals.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_system.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_thread.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_log.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_chrono.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_atomic.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_log_setup.a -framework OpenGL -framework OpenGL /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libGLEW.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libfreetype.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libpng.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libz.a -single_module -Xlinker -dependency_info -Xlinker /Users/user/SoftwareProjekte/FO/test/build/GG/FreeOrion.build/Debug/GiGi.build/Objects-normal/i386/GiGi_dependency_info.dat -o /Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGi.dylib

Undefined symbols for architecture i386:
  "_fchmodat", referenced from:
      boost::filesystem::detail::permissions(boost::filesystem::path const&, boost::filesystem::perms, boost::system::error_code*) in libboost_filesystem.a(operations.o)
ld: symbol(s) not found for architecture i386
clang: error: linker command failed with exit code 1 (use -v to see invocation)


Build target GiGiSDL

Ld /Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGiSDL.dylib normal i386
    cd /Users/user/SoftwareProjekte/FO/test/src
    export MACOSX_DEPLOYMENT_TARGET=10.9
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -arch i386 -dynamiclib -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -L/Users/user/SoftwareProjekte/FO/test/build/Debug -F/Users/user/SoftwareProjekte/FO/test/build/Debug -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk/System/Library/Frameworks -F/Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/Frameworks -filelist /Users/user/SoftwareProjekte/FO/test/build/GG/FreeOrion.build/Debug/GiGiSDL.build/Objects-normal/i386/GiGiSDL.LinkFileList -install_name @rpath/libGiGiSDL.dylib -Xlinker -rpath -Xlinker @loader_path/../SharedSupport -Xlinker -rpath -Xlinker @loader_path/../Frameworks -mmacosx-version-min=10.9 -stdlib=libc++ -dynamiclib -Wl,-headerpad_max_install_names /Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGi.dylib -framework OpenGL -framework OpenGL /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libGLEW.a -framework SDL2 -framework Cocoa /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_date_time.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_filesystem.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_regex.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_signals.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_system.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_thread.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_log.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_chrono.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_atomic.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_log_setup.a -framework OpenGL /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libGLEW.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libfreetype.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libpng.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libz.a -single_module -Xlinker -dependency_info -Xlinker /Users/user/SoftwareProjekte/FO/test/build/GG/FreeOrion.build/Debug/GiGiSDL.build/Objects-normal/i386/GiGiSDL_dependency_info.dat -o /Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGiSDL.dylib

clang: error: no such file or directory: '/Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGi.dylib'


Build target freeorionca

CompileC /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/Objects-normal/i386/DependencyVersions.o util/DependencyVersions.cpp normal i386 c++ com.apple.compilers.llvm.clang.1_0.compiler
    cd /Users/user/SoftwareProjekte/FO/test/src
    export LANG=en_US.US-ASCII
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -x c++ -arch i386 -fmessage-length=0 -fdiagnostics-show-note-include-stack -fmacro-backtrace-limit=0 -Wno-trigraphs -fpascal-strings -O0 -Wno-missing-field-initializers -Wno-missing-prototypes -Wno-return-type -Wno-non-virtual-dtor -Wno-overloaded-virtual -Wno-exit-time-destructors -Wno-missing-braces -Wparentheses -Wswitch -Wno-unused-function -Wno-unused-label -Wno-unused-parameter -Wno-unused-variable -Wunused-value -Wno-empty-body -Wno-uninitialized -Wno-unknown-pragmas -Wno-shadow -Wno-four-char-constants -Wno-conversion -Wno-constant-conversion -Wno-int-conversion -Wno-bool-conversion -Wno-enum-conversion -Wno-shorten-64-to-32 -Wno-newline-eof -Wno-c++11-extensions -DCMAKE_INTDIR=\"Debug\" -DFREEORION_BUILD_AI -DFREEORION_MACOSX -DBOOST_ALL_NO_LINK -DBOOST_DATE_TIME_NO_LOCALE -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -fasm-blocks -fstrict-aliasing -Wdeprecated-declarations -Winvalid-offsetof -mmacosx-version-min=10.9 -g -Wno-sign-conversion -I/Users/user/SoftwareProjekte/FO/test/build/Debug/include -I/Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/include -I/Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/Frameworks/Python.framework/Versions/Current/include/python2.7 -I/Users/user/SoftwareProjekte/FO/test/src/GG -I/Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/DerivedSources/i386 -I/Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/DerivedSources -Wmost -Wno-four-char-constants -Wno-unknown-pragmas -F/Users/user/SoftwareProjekte/FO/test/build/Debug -stdlib=libc++ -fvisibility=hidden -std=c++11 -MMD -MT dependencies -MF /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/Objects-normal/i386/DependencyVersions.d --serialize-diagnostics /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/Objects-normal/i386/DependencyVersions.dia -c /Users/user/SoftwareProjekte/FO/test/src/util/DependencyVersions.cpp -o /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/Objects-normal/i386/DependencyVersions.o

/Users/user/SoftwareProjekte/FO/test/src/util/DependencyVersions.cpp:11:13: fatal error: 'python2.7/patchlevel.h' file not found
#   include <python2.7/patchlevel.h>
            ^
1 error generated.

Ld /Users/user/SoftwareProjekte/FO/test/build/Debug/freeorionca normal i386
    cd /Users/user/SoftwareProjekte/FO/test/src
    export MACOSX_DEPLOYMENT_TARGET=10.9
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -arch i386 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -L/Users/user/SoftwareProjekte/FO/test/build/Debug -F/Users/user/SoftwareProjekte/FO/test/build/Debug -filelist /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/Objects-normal/i386/freeorionca.LinkFileList -Xlinker -rpath -Xlinker @loader_path/../SharedSupport -Xlinker -rpath -Xlinker @loader_path/../Frameworks -mmacosx-version-min=10.9 -stdlib=libc++ -Wl,-search_paths_first -Wl,-headerpad_max_install_names /Users/user/SoftwareProjekte/FO/test/build/Debug/libfreeorioncommon.dylib /Users/user/SoftwareProjekte/FO/test/build/Debug/libfreeorionparse.dylib /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/Frameworks/Python.framework/Versions/Current/lib/libpython2.7.dylib /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_python.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_date_time.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_filesystem.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_iostreams.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_log.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_regex.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_serialization.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_signals.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_system.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_thread.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libz.a -framework CoreFoundation -Xlinker -dependency_info -Xlinker /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/Objects-normal/i386/freeorionca_dependency_info.dat -o /Users/user/SoftwareProjekte/FO/test/build/Debug/freeorionca

ld: file not found: /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorionca.build/Objects-normal/i386/DependencyVersions.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)


Build target freeoriond

CompileC /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/Objects-normal/i386/DependencyVersions.o util/DependencyVersions.cpp normal i386 c++ com.apple.compilers.llvm.clang.1_0.compiler
    cd /Users/user/SoftwareProjekte/FO/test/src
    export LANG=en_US.US-ASCII
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -x c++ -arch i386 -fmessage-length=0 -fdiagnostics-show-note-include-stack -fmacro-backtrace-limit=0 -Wno-trigraphs -fpascal-strings -O0 -Wno-missing-field-initializers -Wno-missing-prototypes -Wno-return-type -Wno-non-virtual-dtor -Wno-overloaded-virtual -Wno-exit-time-destructors -Wno-missing-braces -Wparentheses -Wswitch -Wno-unused-function -Wno-unused-label -Wno-unused-parameter -Wno-unused-variable -Wunused-value -Wno-empty-body -Wno-uninitialized -Wno-unknown-pragmas -Wno-shadow -Wno-four-char-constants -Wno-conversion -Wno-constant-conversion -Wno-int-conversion -Wno-bool-conversion -Wno-enum-conversion -Wno-shorten-64-to-32 -Wno-newline-eof -Wno-c++11-extensions -DCMAKE_INTDIR=\"Debug\" -DFREEORION_BUILD_SERVER -DFREEORION_MACOSX -DBOOST_ALL_NO_LINK -DBOOST_DATE_TIME_NO_LOCALE -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -fasm-blocks -fstrict-aliasing -Wdeprecated-declarations -Winvalid-offsetof -mmacosx-version-min=10.9 -g -Wno-sign-conversion -I/Users/user/SoftwareProjekte/FO/test/build/Debug/include -I/Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/include -I/Users/user/SoftwareProjekte/FO/test/src/GG -I/Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/Frameworks/Python.framework/Versions/Current/include/python2.7 -I/Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/DerivedSources/i386 -I/Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/DerivedSources -Wmost -Wno-four-char-constants -Wno-unknown-pragmas -F/Users/user/SoftwareProjekte/FO/test/build/Debug -stdlib=libc++ -fvisibility=hidden -std=c++11 -MMD -MT dependencies -MF /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/Objects-normal/i386/DependencyVersions.d --serialize-diagnostics /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/Objects-normal/i386/DependencyVersions.dia -c /Users/user/SoftwareProjekte/FO/test/src/util/DependencyVersions.cpp -o /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/Objects-normal/i386/DependencyVersions.o

/Users/user/SoftwareProjekte/FO/test/src/util/DependencyVersions.cpp:11:13: fatal error: 'python2.7/patchlevel.h' file not found
#   include <python2.7/patchlevel.h>
            ^
1 error generated.

Ld /Users/user/SoftwareProjekte/FO/test/build/Debug/freeoriond normal i386
    cd /Users/user/SoftwareProjekte/FO/test/src
    export MACOSX_DEPLOYMENT_TARGET=10.9
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -arch i386 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -L/Users/user/SoftwareProjekte/FO/test/build/Debug -F/Users/user/SoftwareProjekte/FO/test/build/Debug -filelist /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/Objects-normal/i386/freeoriond.LinkFileList -Xlinker -rpath -Xlinker @loader_path/../SharedSupport -Xlinker -rpath -Xlinker @loader_path/../Frameworks -mmacosx-version-min=10.9 -stdlib=libc++ -Wl,-search_paths_first -Wl,-headerpad_max_install_names /Users/user/SoftwareProjekte/FO/test/build/Debug/libfreeorioncommon.dylib /Users/user/SoftwareProjekte/FO/test/build/Debug/libfreeorionparse.dylib /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/Frameworks/Python.framework/Versions/Current/lib/libpython2.7.dylib /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_python.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_date_time.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_filesystem.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_iostreams.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_log.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_regex.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_serialization.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_signals.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_system.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_thread.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libz.a -framework CoreFoundation -Xlinker -dependency_info -Xlinker /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/Objects-normal/i386/freeoriond_dependency_info.dat -o /Users/user/SoftwareProjekte/FO/test/build/Debug/freeoriond

ld: file not found: /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeoriond.build/Objects-normal/i386/DependencyVersions.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)


Build target freeorion

Ld /Users/user/SoftwareProjekte/FO/test/build/Debug/FreeOrion.app/Contents/MacOS/FreeOrion normal i386
    cd /Users/user/SoftwareProjekte/FO/test/src
    export MACOSX_DEPLOYMENT_TARGET=10.9
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -arch i386 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk -L/Users/user/SoftwareProjekte/FO/test/build/Debug -F/Users/user/SoftwareProjekte/FO/test/build/Debug -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk/System/Library/Frameworks -F/Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/Frameworks -filelist /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorion.build/Objects-normal/i386/FreeOrion.LinkFileList -Xlinker -rpath -Xlinker @loader_path/../SharedSupport -Xlinker -rpath -Xlinker @loader_path/../Frameworks -mmacosx-version-min=10.9 -stdlib=libc++ -Wl,-search_paths_first -Wl,-headerpad_max_install_names /Users/user/SoftwareProjekte/FO/test/build/Debug/libfreeorioncommon.dylib /Users/user/SoftwareProjekte/FO/test/build/Debug/libfreeorionparse.dylib /Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGiSDL.dylib -framework OpenGL -framework OpenGL -framework OpenAL -framework SDL2 -framework Cocoa -framework ogg -framework vorbis -framework vorbis -framework vorbis /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libz.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_locale.a /usr/lib/libiconv.dylib /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_iostreams.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_serialization.a -framework CoreFoundation /Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGi.dylib /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_date_time.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_filesystem.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_log.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_regex.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_signals.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_system.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_thread.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_chrono.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_atomic.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libboost_log_setup.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libfreetype.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libpng.a -framework OpenGL /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libGLEW.a /Users/user/SoftwareProjekte/FO/test/src/Xcode/dep/lib/libz.a -fobjc-link-runtime -Xlinker -dependency_info -Xlinker /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorion.build/Objects-normal/i386/FreeOrion_dependency_info.dat -o /Users/user/SoftwareProjekte/FO/test/build/Debug/FreeOrion.app/Contents/MacOS/FreeOrion

clang: error: no such file or directory: '/Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGiSDL.dylib'
clang: error: no such file or directory: '/Users/user/SoftwareProjekte/FO/test/build/Debug/libGiGi.dylib'

PhaseScriptExecution CMake\ PostBuild\ Rules /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorion.build/Script-4A9594C5299C461E81E6E19E.sh
    cd /Users/user/SoftwareProjekte/FO/test/src
    /bin/sh -c /Users/user/SoftwareProjekte/FO/test/build/FreeOrion.build/Debug/freeorion.build/Script-4A9594C5299C461E81E6E19E.sh

/Applications/CMake.app/Contents/bin/cmake -E make_directory /Users/user/SoftwareProjekte/FO/test/build/Debug/FreeOrion.app/Contents/MacOS/../Executables
/Applications/CMake.app/Contents/bin/cmake -E copy_if_different /Users/user/SoftwareProjekte/FO/test/build/Debug/freeorionca /Users/user/SoftwareProjekte/FO/test/build/Debug/FreeOrion.app/Contents/MacOS/../Executables
Error copying file (if different) from "/Users/user/SoftwareProjekte/FO/test/build/Debug/freeorionca" to "/Users/user/SoftwareProjekte/FO/test/build/Debug/FreeOrion.app/Contents/MacOS/../Executables".
make: *** [freeorion_buildpart_0] Error 1
Command /bin/sh failed with exit code 2
```
Editing the generated Xcode project and resetting the value for `Base SDK` to the default `No SDK (Latest OS X)` or `OS X 10.10` (which is the same on my system, as I'm on 10.10) results in a successful build. I get a FO app bundle like you described, which seems to be fully functional (did a brief test).

AFAICT from all my different attempts so far is that apparently `Base SDK` needs to be set to `OS X 10.10` or later. I guess the best would be to not set `Base SDK` at all, which means that the generated Xcode project will fall back onto the default value `No SDK (Latest OS X)`, as this makes no assumptions about which legacy SDKs have to be installed on the system.

However, I don't know what changes have to be made to to the cmake scripts to achieve that.

Aside from that, so far so excellent. 👍 😃 

*However*, there is another issue, not a deal breaker, but annoying. In Xcode, for each build target so called "schemes" get created, which contain various settings for a fixed set of (what I call) build profiles (which are "Run", "Test", "Analyze", "Profile" and "Archive"). One of those settings is which build configuration ("Release", "Debug" or "Coverage") is used by a profile.

Now, the profile used when doing a standard build (keyboard shortcut CMD+B) apparently is "Run", which has "Debug" set as build configuration. Meaning, when doing a standard build in the generated Xcode project, you get a debug build.

But unless you actually want to step through the code with the debugger, a debug build is quite useless - it inflates the resulting app bundle to more than twice the size of a "Release" build, and the app is also *significantly* slower.

Which means, what you usually want when doing a standard build is using the "Release" build configuration, which is what the "Run" profile should use. Any idea how to tell the cmake scripts to do that?

The annoying thing is, you'd have to change that setting in the schemes of all build targets you usually use. If I e.g. only change it in the scheme for target "ALL_BUILD", and then build target "package" (to get a nice DMG package), Xcode will produce a debug build and package that.

And every time I regenerate the Xcode project via cmake, I have to remember to reapply these adjustments again. Ugh.

Vezzra:
Did some digging, and the issue with the base SDK can be solved by adding this line to the cmake script:
```
set(CMAKE_OSX_SYSROOT macosx CACHE STRING "Set the OSX base SDK version")
```
This will set `Base SDK` to `Latest OS X`. Not exactly the actual default, but good enough. If you prefer to explicitely set the base SDK to 10.10, the line looks like this:
```
set(CMAKE_OSX_SYSROOT macosx10.10 CACHE STRING "Set the OSX base SDK version")
```
But, as mentioned above, I'd prefer the first variant.

The issue with the build configuration is more tricky. Apparently the Xcode generator happily ignores whatever `CMAKE_BUILD_TYPE` is set to. The only way to force it to use the "Release" build configuration in the "Run" build profile is to set `CMAKE_CONFIGURATION_TYPES` *only* to `Release` (no `Debug`). Then all build profiles in all schemes will use "Release" as build configuration.

So I suggest to set `CMAKE_CONFIGURATION_TYPES` to `Release` for the OSX platform. If one needs to make a debug build, they just have to regenerate the Xcode project with `CMAKE_CONFIGURATION_TYPES` set to just `Debug` or `Debug Release` (whatever floats their boat).

Unless you find some way how to tell the Xcode generator to set the build configurations in the build profiles of the target schemes. I googled it, but failed to find something I could make sense of.

adrianbroher:
> [When I change the base SDK]

There is no point in that aside from helping locating the issue. The problem is that the FreeOrionSDK (or rather boost::filesystem) builds with some wrong assumptions about it's build environment and therefor is linking to a function that does not exist in the minimum OSX version requested.

> Any idea how to tell the cmake scripts to  [select another default build configuration]?

Reordering the `CMAKE_CONFIGURATION_TYPE` in CMakeLists.txt:13 in should do that.  Just place the `Release` entry at the front of the list.  This variable tells CMake what configurations are available for multiconfig generators like Xcode or VS and as far as I can see from the CMake code the first configuration is used as default one.

Vezzra:
> There is no point in that aside from helping locating the issue.

Ok, not wrt to fixing the issue, but I still think the base SDK should be set to `Latest OS X`, for the reasons I cited. Avoids "SDK not found" issues when the 10.9 SDK is not availabe on a system (expect this to happen on 10.13, if that isn't already the case with 10.12). Or is there a particular reason for setting the base SDK to 10.9? Judging by my experience so far setting the deployment target to the min OSX version you want to support seems sufficient...

> I did a new local build with the patched FreeOrionSDK and things still work.  Checking `libs/libboost_filesystem.a` with `nm` doesn't show any reference to `_fchmodat` anymore.

I just gave this another try and rebuilt the patched SDK on my local system. To be extra sure, I added:
```
set(CMAKE_OSX_SYSROOT macosx10.9 CACHE STRING "Set the OSX base SDK version")
```
to the cmake script. Then I checked `libboost_filesystem.a` with `nm`, and it showed a reference to `_fchmodat`. This is frustrating. How do I get rid of that??? :frowning_face:

> Reordering the CMAKE_CONFIGURATION_TYPE in CMakeLists.txt:13 in should do that.

I found that suggestion too when I googled the issue, and I guess that worked for some earlier Xcode version. However, not anymore. Apparently the order in which the build configurations are listed here doesn't matter anymore, Xcode just stubbornly sets "Debug" as build configuration for the "Run" profile.

Unless there is a cmake way to tell Xcode to use "Release" for "Run", only omitting "Debug" from `CMAKE_CONFIGURATION_TYPE` forces Xcode to use "Release" for all build profiles.

adrianbroher:
You could add a preprocessor message to the SDK patch and check if you can see the message.  I will set up a test branch on the SDK for that.

Vezzra:
> You could add a preprocessor message to the SDK patch and check if you can see the message.

What preprocessor message and where should it be added? What is that message supposed to tell me?

Vezzra:
The switch to shared boost libraries on OSX indeed solved the issue with that `_fchmodat` symbol. The good news are, building FO with cmake finally works on OSX. However, I had to tackle a few issues to get there:

* A compile error in [`DependencyVersions.cpp`, line 11](https://github.com/freeorion/freeorion/blob/621359c4cb9afa4a63c3e2ae4103c52e8861fdca/util/DependencyVersions.cpp#L11): `PYTHON_INCLUDE_PATH` gets set to `.../Python.framework/Versions/Current/include/python2.7`, which of course breaks `#include <python2.7/patchlevel.h>`. I provided a PR with a fix: #1850.

* The shared boost libraries didn't get copied to the app bundle, which of course caused the app to crash immediately. Fixed by 621359c4cb9afa4a63c3e2ae4103c52e8861fdca.

* The resulting app bundle was 30-40MB larger than the app bundle created by the manually maintained Xcode project. Upon investigation I discovered that cmakes copy_directory command (which is used to copy the frameworks to the app bundle) follows symlinks. As frameworks on OSX contain quite some symlinks, this resulted in the copies being up to twice as large as the originals. PR with a proposed fix: #1851.

* Another issue is something already mentioned in the last part of [this comment](https://github.com/freeorion/freeorion/pull/1455#issuecomment-299706377). Because of the (IMO confusing/convoluted) way Xcode handles build configurations, unless you remove 'Debug' from `CMAKE_CONFIGURATION_TYPES`, the generated Xcode project always produces debug builds when starting a build via CMD-B (which is what you normally do), regardless what `CMAKE_BUILD_TYPE` is set to. That's seriously annoying, so I provided a fix for that issue to: #1852.

With all the provided PRs applied, this branch (osx-cmake) builds nicely with cmake. There is only one issue left which I haven't been able to solve (yet), but isn't really a deal breaker: packaging the generated app bundle into a DMG by using the "package" target doesn't work. The DMG file gets created, but without the app bundle, it only contains the symlink to the OSX "Applications" folder (and therefore is only 72KB in size).

I have no idea what's going on here, as that definitely already has been working. The only thing which might be a possible cause for that change is that I upgraded my cmake installation to the current version. But after downgrading to the version I had installed before, the problem persisted, which indicates that the new version very likely isn't the culprit here.

However, I can work around that issue. Being able to retire the manually maintained Xcode project vastly outweighs this minor annoyance, so that shouldn't block switching to cmake on OSX.

Vezzra:
Found out why the FO app bundle didn't get copied to the generated DMG file when building the "package" target: apparently there is no proper `install` statement issued for the OSX platform. PR with a fix: #1854.

This solves the last remaining issue. Once all the fixes are applied/merged, I guess we're good to make the switch...?

adrianbroher:
From my side we can merge now.  I left the OSX specific files in the Xcode directory.  The packaging related files for all platforms  can be cleaned up in another PR.

adrianbroher:
@LGM-Doyle  I've added the `Fix compile error when building with Clang` commit to fix a bug related to `Universe::InsertID`.  I replaced `std::shared_ptr<T>(new T(std::forward<Args>(args)...));` with `std::make_shared<T>(args...);` which should be equivalent to my knowledge.  Is there a particular reason why you chose former?

Vezzra (approved):
Did a brief final test, everything works perfectly. As far as I am concerned, LGTM.

Once this is merged, I'll update the compile instructions for OSX on the wiki as soon as I can get around to it (unless you absolutely want to do that yourself...).

Vezzra:
> The packaging related files for all platforms can be cleaned up in another PR.

For OSX, that's just removing the `make_dmg.py` script in the `Xcode` folder, IMO no extra PR needed for that. If you don't add a respective commit, I'll do it after this PR has been merged.

LGM-Doyle:
@adrianbroher I had some internet troubles, otherwise I would have responded sooner.

I'm excited to see progress on this PR.  It will be a significant improvement for other developers.

The original line of code was doing two things:  
1) allowing the constructors to remain private.  This was to preserve the intent of the original code.  Changing it as you did does not impact performance.
2) perfectly forwarding the `Args..`.  Not forwarding the arguments, will mean that there is an unnecessary copy before the constructors of the universe objects are called.  It would be best if `std::forward<Args>(args)...` was restored, if it will compile on clang.  If forwarding won't compile, then I have no concerns about the changes.

p.s. Although it is not relevant to this discussion,  `std::make_shared<T>(args)` and `std::shared_ptr<T>(new T(args))` do slightly different things:
- `std::make_shared<T>(args)` makes a single memory allocation for both the pointer control block and the `T`.  The memory is not freed until all of the `shared_ptrs` **and** `weak_ptrs` are destroyed.
- `std::shared_ptr<T>(new T(args))` allocates separate memory for the `T` and the pointer control block.  The memory for the `T` is freed when the last `shared_ptr` is destroyed even if `weak_ptrs` still exist.  The memory for the control block is freed when all  of the `shared_ptrs` **and** `weak_ptrs` are destroyed.

LGM-Doyle:
Huzzah!

This is a great improvement.  I've been looking forward to seeing it land. 

Thanks for doing the work.

PR #1458, opened 2017-04-02
Fix FleetWnd doesn't update StatsIcon when ship with fighters is merged.

LGM-Doyle:
The problem is that when the FleetDataPanel is initialized it determines
which StatisticIcons are included based on its fleet's current
capabilities.  When a ship is transferred in with new capabilities, like
fighters, it does not have a StatisticIcon to update with the new
values.

This commit unconditionally, includes all StatisticIcons possible for
fleets and then conditionally displays those with non null values.

This problem was reported in the [forum](http://freeorion.org/forum/viewtopic.php?f=28&t=10488&sid=3057b40a92976e087d01437ed58e885e)

### DIFF
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -1058,6 +1058,7 @@ class FleetDataPanel : public GG::Control {
 public:
     FleetDataPanel(GG::X w, GG::Y h, int fleet_id);
     FleetDataPanel(GG::X w, GG::Y h, int system_id, bool new_fleet_drop_target);
+    ~FleetDataPanel();
 
     /** Upper left plus border insets. */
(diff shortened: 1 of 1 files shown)


dbenage-cx:
Thanks for looking into this.
iirc Wnd::Visible is changed when parent wnd is shown/hidden, e.g. with the fleet wnd open: open and close the research wnd and notice issues.

LGM-Doyle:
Thanks for reviewing this.  

I changed it to attaching and detaching the windows.
I had originally balked at doing the resource management in the destructor.

dbenage-cx:
As far as testing LGTM.  Have not found any issues with stat icons appearing when appropriate.

PR #1459, opened 2017-04-02
Move delete fleet responsiblity from fleetwnd to FleetTransferOrder.

LGM-Doyle:
When a ship/fleet is transferred into a fleet, the current `FleetWnd` UI code is repsonsible for checking if the source fleet is now empty and deleting the fleet.  

This PR moves this responsibility from the UI code into the FleetTransferOrder code.  The Orders code is part of the the game logic maintaining correct game state.  The UI code is not.

### DIFF
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -1842,7 +1842,6 @@ public:
                     ErrorLogger() << "FleetsListBox::AcceptDrops  unable to get dropped fleet?";
                     continue;
                 }
-                int dropped_fleet_id = dropped_fleet->ID();
 
                 // get fleet's ships in a vector, as this is what FleetTransferOrder takes
                 const std::set<int>& ship_ids_set = dropped_fleet->ShipIDs();
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -388,20 +388,6 @@ namespace {
         // order ships moved into target fleet
         HumanClientApp::GetApp()->Orders().IssueOrder(
             OrderPtr(new FleetTransferOrder(client_empire_id, target_fleet->ID(), empire_system_ship_ids)));
-
-
-        // delete empty fleets from which ships have been taken
-        GetUniverse().InhibitUniverseObjectSignals(true);
-        for (int fleet_id : empire_system_fleet_ids) {
-            HumanClientApp::GetApp()->Orders().IssueOrder(OrderPtr(
-               new DeleteFleetOrder(client_empire_id, fleet_id)));
-        }
-        GetUniverse().InhibitUniverseObjectSignals(false);
-
-        // signal to update UI
-        system->FleetsRemovedSignal(empire_system_fleets);
-        system->StateChangedSignal();
-        target_fleet->StateChangedSignal();
     }
 
    /** Returns map from object ID to issued colonize orders affecting it. */
@@ -1865,11 +1851,6 @@ public:
                 // order the transfer
                 HumanClientApp::GetApp()->Orders().IssueOrder(
                     OrderPtr(new FleetTransferOrder(empire_id, target_fleet_id, ship_ids_vec)));
-
-                // delete empty fleets
-                if (dropped_fleet->Empty())
-                    HumanClientApp::GetApp()->Orders().IssueOrder(
-                        OrderPtr(new DeleteFleetOrder(empire_id, dropped_fleet_id)));
             }
 
         } else if (!dropped_ships.empty()) {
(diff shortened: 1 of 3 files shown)


UI/FleetWnd.cpp
```diff
@@ -1865,11 +1851,6 @@ class FleetsListBox : public CUIListBox {
                 // order the transfer
                 HumanClientApp::GetApp()->Orders().IssueOrder(
                     OrderPtr(new FleetTransferOrder(empire_id, target_fleet_id, ship_ids_vec)));
-
-                // delete empty fleets
-                if (dropped_fleet->Empty())
-                    HumanClientApp::GetApp()->Orders().IssueOrder(
-                        OrderPtr(new DeleteFleetOrder(empire_id, dropped_fleet_id)));
```
dbenage-cx, line 1872: `dropped_fleet_id` now unused

Dilvish-fo:
and of course when the AI transfers ships it doesn't use the UI code at all, and so presumably it's been left with a bunch of empty fleets clogging things up? I see in the AI code that after a transfer it checks a few things, including whether the old fleet is now empty, and deletes its entry in AIstate in that case, but I don't see anywhere that it was issuing a DeleteFleetOrder like the human UI had been doing.

And in fact, in AIInterface.cpp I find
```
    int IssueDeleteFleetOrder() {
        return 0;
    }
```

And it looks like this nop function is not even exposed to the AI, and perhaps there is no need to.

I note that there is one surviving use of DeleteFleetOrder in FleetWnd.cpp (in CreateNewFleetsForShips, around lines 281-285), and it is seeming to me that the form of NewFleetOrder used there should probably be doing the same empty-fleet checking and deleting that you have FleetTransferOrder doing here, and there is probably no need at all for DeleteFleetOrder anywhere.  But that could just as well be a separate PR even if we all agree on it.

geoffthemedio:
The server cleans up empty fleets each turn, so there's no real need to do so with orders. It's done in the UI / human client just so there is no empty fleet shown in the FleetWnd after transferring all ships out of a fleet.

adrianbroher:
@geoffthemedio is that a point for @Dilvish-fo or for @LGM-Doyle ?  Because I agree with @LGM-Doyle here, the UI should not be in charge of maintaining the model consistency.  Having a potential server NoOP here is favorable to duplicated code.

geoffthemedio:
Just a neutral and hopefully informative or helpful observation...

LGM-Doyle:
I've removed the unused variable.

I think, we are mostly on the same page page.

Here is some more information.

When the AI calls [`FleetTransferOrder`](https://github.com/freeorion/freeorion/blob/master/default/python/AI/FleetUtilsAI.py#L236) in `FleetUtilsAI.py`, it also immediately [deletes](https://github.com/freeorion/freeorion/blob/master/default/python/AI/FleetUtilsAI.py#L246) the from fleet if it is empty.  So the AI client requires that empty fleets will be removed immediately.

This code at the beginning of `MapWnd::Render()` 
````
    // HACK! This is placed here so we can be sure it is executed frequently
    // (every time we render), and before we render any of the
    // FleetWnds.  It doesn't necessarily belong in MapWnd at all.
    FleetUIManager::GetFleetUIManager().CullEmptyWnds();
````
indicates that the human client requires that empty fleets will be removed immediately, but that is not always happening.

As @geoffthemedio pointed out the server removes the empty fleets, during the turn update.  The gamestate is correct in the next turn update.

However, both clients are allowed to remove ships from and destroy fleets that they own.  They both require that an empty fleet be destroyed immediately to maintain the expected invariant that a fleet is not empty.  

The client backends should delete the empty fleet and maintain the invariant immediately.  The server can check that the order was valid and duplicate the work.

If the client backends handle this correctly, then the clients can require and expect that there are no empty fleets.

I have not checked that there are no other cases of fleet losing ships during the human turns, so I don't think that removing the check in `MapWnd` is correct immediately before a release.

geoffthemedio:
There may be some history in play here, in that ships ordered to colonize arent immediately destroyed client side. That used to leave empty fleets that could reappear if the colonize order was cancelled, meaning deleting fleets during a flet transfer order could have caused problems or client-server inconsistencies.

Dilvish-fo:
@LGM-Doyle 
> it also immediately deletes the from fleet 

I think you are misreading that code; it looks to me like you are referring to the same code I had mentioned, "and deletes its entry in AIstate in that case" -- it is only deleting some AI info about the fleet, not doing anything to delete the Fleet UniverseObject, as far as I can see.

> both clients are allowed to remove ships from and destroy fleets that they own. 

Hopefully we are now on the same page that the AI currently has no way to actually delete fleets, although from what Geoff says about the server side cleanup it may not matter.  But I'd still like to push for a response from you guys about my suggestion that neither of the clients should actually need a standalone DeleteFleetOrder and it should probably be done away with.

LGM-Doyle:
@Dilvish-fo, you are correct I misread the AI python code in `FleetUtilsAI.py`.

I agree that neither client should need or be able to delete a fleet since a ship can't exist without a fleet and a fleet can't be empty.  A fleet being empty should either be a transient event that the clients never see, or an invariant that `Fleet` enforces on its own.

I will submit another PR after this one removing the DeleteFleetOrder if everyone is in agreement.

Issue #521, opened 2016-03-12
Debug logs of Production update printed to all AI logs


Cjkjvfnby:
Debug text related to single AI printed to all AI logs. 
`AI_1.log`:

```
2016-03-12 13:18:48.344096 [debug] AI : ========= Production Update for empire: 1 ========
2016-03-12 13:18:48.345097 [debug] AI : ========= Production Update for empire: 2 ========
2016-03-12 13:18:48.345097 [debug] AI : ========= Production Update for empire: 3 ========
2016-03-12 13:18:48.345097 [debug] AI : ========= Production Update for empire: 4 ========
2016-03-12 13:18:48.346097 [debug] AI : ========= Production Update for empire: 5 ========
2016-03-12 13:18:48.346097 [debug] AI : ========= Production Update for empire: 6 ========
2016-03-12 13:18:48.346097 [debug] AI : ========= Production Update for empire: 7 ========
2016-03-12 13:18:48.346097 [debug] AI : ========= Production Update for empire: 8 ========
2016-03-12 13:18:48.346097 [debug] AI : ========= Production Update for empire: 9 ========
2016-03-12 13:18:48.347097 [debug] AI : ========= Production Update for empire: 10 ========
2016-03-12 13:18:48.347097 [debug] AI : ========= Production Update for empire: 11 ========
2016-03-12 13:18:48.347097 [debug] AI : ========= Production Update for empire: 12 ========
2016-03-12 13:18:48.347097 [debug] AI : ========= Production Update for empire: 13 ========
2016-03-12 13:18:48.347097 [debug] AI : ========= Production Update for empire: 14 ========
2016-03-12 13:18:48.348097 [debug] AI : ========= Production Update for empire: 15 ========
2016-03-12 13:18:48.348097 [debug] AI : ========= Production Update for empire: 16 ========
2016-03-12 13:18:48.348097 [debug] AI : ========= Production Update for empire: 17 ========
2016-03-12 13:18:48.348097 [debug] AI : ========= Production Update for empire: 18 ========
2016-03-12 13:18:48.348097 [debug] AI : ========= Production Update for empire: 19 ========
2016-03-12 13:18:48.349097 [debug] AI : ========= Production Update for empire: 20 ========
```

If any other error occurred on update them printed to all log too.
This is example from my **development** branch with additional entries. 
`AI_1.log`:

```
2016-03-12 13:19:18.325811 [debug] AI : ========= Production Update for empire: 1 ========
[...]

PR #1460, opened 2017-04-02
Get rid of excess Production Update messages in log files

memem359:
This should address issue [#521 Debug logs of Production update printed to all AI logs](https://github.com/freeorion/freeorion/issues/521) (the oldest "optional for 0.4.7" issue).

Short Version: The function UpdateResourcePools loops over all empires, even though it only needs to run on the one matching its empire_id. Changed the selection code to be the same as UpdateResearchQueue and UpdateProductionQueue. The only change in the game should be a bit fewer wasted CPU cycles and unneeded print statements in the AI and freeorion logs. (In my test game, the AI behavior was the same.)

Additional Info: The poster for #521 suspected that the same messages were being sent to all the AI logs. What was actually happening was the same inefficient code was being run on every player client (human and AI). It was looping over all empires, sending messages to its Logger for each one, and after a bit of boondoggling would exit without doing anything if the loop empire was not the same as the client empire_id. (Non-matching ids are the sources of the message "not enough PP to be worth simulating future turns production"). To the best of my knowledge, looping over all empires and calling empire->UpdateResourcePools for each one is unnecessary.

### DIFF
--- a/AI/AIInterface.cpp
+++ b/AI/AIInterface.cpp
@@ -220,8 +220,15 @@ namespace AIInterface {
     }
 
     void UpdateResourcePools() {
-        for (std::map<int, Empire*>::value_type& entry : AIClientApp::GetApp()->Empires())
-            entry.second->UpdateResourcePools();
+//        for (std::map<int, Empire*>::value_type& entry : AIClientApp::GetApp()->Empires())
+//            entry.second->UpdateResourcePools();
+        int empire_id = AIClientApp::GetApp()->EmpireID();
+        Empire* empire = ::GetEmpire(empire_id);
+        if (!empire) {
+            ErrorLogger() << "UpdateResourcePools : couldn't get empire with id " << empire_id;
+            return;
+        }
+        empire->UpdateResourcePools();
     }
 
     void UpdateResearchQueue() {


Dilvish-fo:
Let me point out that I think the current state is more of an incomplete implementation rather than something actually broken (which seems to be the view of it in this PR and in #521).   However, 
after writing all this I wound up deciding that multiplayer considerations probably lean towards the cleanup proposed here or something like it rather than trying to get to a less broken-seeming intermediate point.  

Food for thought: some considerations that I think are pertinent but not yet mentioned here nor in #521, having to do with the expected future implementation of an espionage system.

Right now, each empire is given substantial knowledge about each of the other empires, at least including what techs they have researched (allowing you to get meaningful details in the meter tooltips for their planets and ships), and you get access to their production queue so that you can see when they have a building under construction on a planet and how far along it is.  One of the things that is ***not*** currently visible to other empires are the actual ResourcePools, and so this bit of code cited here sees no resources available and doesn't really accomplish anything currently.

It's my understanding of our intentions that much of the information about other empires that is currently freely provided, would become gated by technologies and/or espionage projects (with potentially varying degrees of accuracy).  I would expect that the currently missing ResourcePool information would be up for grabs in that system, so that a spy-intensive empire would not only be able to know what its enemies were building but also would be able to predict how long for such things to be completed (i.e., at least part of the code being considered nonfunctional here would actually be fully functional).

So, another option to consider here is just to go ahead and expose the ResourcePools so that the production estimator would have some info to work with, and tooltips for enemy under-construction buildings could indicate the projected remaining construction time (though in many cases those projections would currently probably be pretty unreliable).  That would seem to me to be more of a step forward towards our expected target that stripping the code out would.  But, it is an intermediate stopping point that our multiplayer friends might not like, and unreliable estimates would probably seem even more broken that some logging spam even if there was not really anything broken about them.

memem359:
That was a reason why I commented out the previous loop instead of erasing it, in case it was meant to be reused in the future.

I had turned on the Super-Tester in my test game. Right now, knowing the other empire's PP and RP still doesn't cause the UpdateResourcePools to do anything.

If espionage is implemented in the future, it should probably be a dedicated routine (which loops over all empires except for the home empire), with its own style of log messages, instead of being mixed in with the home empire updates.

geoffthemedio:
Having visibulity of other empires' planets is separate from whether the player's UI will attempt to populate resource pools or simzlate queues with that information.

memem359:
I suspected that was true.
That was a personal sanity check. The code is complicated enough, there may have been something buried deep that would activate when the client was aware of other empire's RP and PP. (For example, if this also helped to update the PP and RP numbers in the Empire window, I would have found out the hard way.)

This commit has a small (and easily reversed) change. I am trying to make the argument that it is worth it to simplify the AI logs a bit now, and mark one of the older Issues as Closed, rather than worry about whatever large-scale and wide-spread changes espionage will require in the future.

geoffthemedio:
seems reasonable to me as a temporary change. no objections to the code itself.

PR #1462, opened 2017-04-03
Strip whitespace from planet names

Morlic-fo:
Before this PR, there is a trailing whitespace if a planet has no cardinal suffix:
```
print planet
P_59<Russell Asteroids >
```
With this PR, the trailing whitespace is removed:
```
print planet
P_59<Russell Asteroids>
```


dbenage-cx:
See this more as a bug with `Planet::CardinalSuffix` (I will provide PR), but this seems ok at least as a fallback.

Vezzra (approved):
LGTM

Issue #1404, opened 2017-03-15
Add Tooltip for Focus Selection Dropdown


Morlic-fo:
### Environment
* **FreeOrion Version**: build 2017-03-14.853af0e
* **Operating System**: Windows 7
* **Graphic card used**: Nvidia GT520M
* **Fetched as** Weekly test build

### Description
When hovering over the (collapsed) dropdown to select a planetary focus in the sidepanel, a tooltip for the focus is displayed.
It would be nice to have the tooltip shown for each entry when the dropdown is expanded.

See screenshot for reference.

![dropdowntooltip](https://cloud.githubusercontent.com/assets/11706951/23972582/ac628164-09d2-11e7-8b92-e7338e4a4dac.png)

LGM-Doyle:
Implemented by #1466

PR #1466, opened 2017-04-04
Add tooltip to SidePanel focus drop down items.

LGM-Doyle:
Implements request #1404.


geoffthemedio (commented):
If something is this simple and probably harmless I think you can just commit it directly...

PR #1469, opened 2017-04-06
Make logistics facilitator effect use LocalCandidate within scope expression.

LGM-Doyle:
The scope statement was using Target within the scope expression, which
was generating error messages.

I am not certain that this is either definitely a bug or definitely the fix for the bug.

If someone could build a logistics facilitator with master and verify that they see this error.
````
FollowReference : top level object (Target) not defined in scripting context
````
, and then check that this PR fixes the problem.


dbenage-cx:
Nice find, can verify.

geoffthemedio:
There is no Target object defined when evaluating a scope condition. LocalCandidate is probably right.

LGM-Doyle:
Thanks for checking.

PR #1470, opened 2017-04-07
Fix client networking synchronous message hang

LGM-Doyle:
This PR fixes a bug reported on the [forum](http://freeorion.org/forum/viewtopic.php?f=28&t=10497&sid=dfee4de00afbbb552882a49ac6e2d965).

The problem as reported by Scara on the forum is that repeatedly loading games can cause a `freeoriona` AI client to hang.

This was difficult to replicate.  I only replicated it twice, and only attached the debugger once.  I was unable to observe the hang after applying this PR.  However, if someone has a more reliable way to replicate this hang, it would be good to test the PR that way.

The underlying problem is that the AI process in unresponsive and blocks on network requests when processing time consuming requests, like `GAME_START` and `TURN_UPDATE`.  This PR does not fix that problem, because I couldn't imagine a simple fix suitable for days before a release.

The specific problem I observed is that in the `MessageQueue` in `ClientNetworking` it is possible for `EraseFirstSynchronousResponse()` to hang after the AI client has shutdown the networking thread.  

First, In a few places `MessageQueue` was checking a condition under one mutex lock and then in a different function, under a different mutex lock using that result.

`MessageQueue` provided `MessageAvailable()` which was used to guard `GetMessage(Message)`.  I removed `MessageAvailable()` and change the signature of `GetMessage` to  `optional<Message> GetMessage()`.  The queue state is now checked and changed under the same mutex.

Second, in the synchronous message reception, `MessageQueue` never checked that `ClientNeworking` was still connected and receiving. 

 Now `optional<Message> GetSynchronouMessage(sent_message)` checks after each failed receive attempt to see if the rx network has shutdown.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -799,14 +799,12 @@ void HumanClientApp::HandleSystemEvents() {
     if (m_connected && !m_networking->IsConnected()) {
         m_connected = false;
         DisconnectedFromServer();
-    } else if (m_networking->MessageAvailable()) {
-        Message msg;
-        m_networking->GetMessage(msg);
+    } else if (auto msg = Networking().GetMessage()) {
         try {
-            HandleMessage(msg);
+            HandleMessage(*msg);
         } catch (const std::exception& e) {
             ErrorLogger() << "exception handing message: " << e.what();
-            ErrorLogger() << "message type: " << msg.Type() << " and text: " << msg.Text();
+            ErrorLogger() << "message type: " << msg->Type() << " and text: " << msg->Text();
         }
     }
 }
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -684,13 +684,12 @@ void HumanClientApp::RequestSavePreviews(const std::string& directory, PreviewIn
         }
     }
     DebugLogger() << "HumanClientApp::RequestSavePreviews Requesting previews for " << generic_directory;
-    Message response;
-    m_networking->SendSynchronousMessage(RequestSavePreviewsMessage(PlayerID(), generic_directory), response);
-    if (response.Type() == Message::DISPATCH_SAVE_PREVIEWS){
-        ExtractDispatchSavePreviewsMessageData(response, previews);
+    const auto response = m_networking->SendSynchronousMessage(RequestSavePreviewsMessage(PlayerID(), generic_directory));
+    if (response && response->Type() == Message::DISPATCH_SAVE_PREVIEWS){
+        ExtractDispatchSavePreviewsMessageData(*response, previews);
         DebugLogger() << "HumanClientApp::RequestSavePreviews Got " << previews.previews.size() << " previews.";
     }else{
-        ErrorLogger() << "HumanClientApp::RequestSavePreviews: Wrong response type from server: " << response.Type();
+        ErrorLogger() << "HumanClientApp::RequestSavePreviews: Wrong response type from server: " << response->Type();
     }
 }
 
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -685,10 +685,10 @@ void HumanClientApp::RequestSavePreviews(const std::string& directory, PreviewIn
     }
     DebugLogger() << "HumanClientApp::RequestSavePreviews Requesting previews for " << generic_directory;
     const auto response = m_networking->SendSynchronousMessage(RequestSavePreviewsMessage(PlayerID(), generic_directory));
-    if (response && response->Type() == Message::DISPATCH_SAVE_PREVIEWS){
+    if (response && response->Type() == Message::DISPATCH_SAVE_PREVIEWS) {
         ExtractDispatchSavePreviewsMessageData(*response, previews);
         DebugLogger() << "HumanClientApp::RequestSavePreviews Got " << previews.previews.size() << " previews.";
-    }else{
+    } else {
         ErrorLogger() << "HumanClientApp::RequestSavePreviews: Wrong response type from server: " << response->Type();
     }
 }
--- a/network/MessageQueue.cpp
+++ b/network/MessageQueue.cpp
@@ -12,7 +12,7 @@ namespace {
 }
 
 MessageQueue::MessageQueue(boost::mutex& monitor) :
-    m_monitor(monitor)
+    m_monitor(monitor), m_stopped_growing(false)
 {}
 
 bool MessageQueue::Empty() const {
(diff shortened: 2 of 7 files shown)


client/human/HumanClientApp.cpp
```diff
@@ -684,13 +684,12 @@ void HumanClientApp::RequestSavePreviews(const std::string& directory, PreviewIn
         }
     }
     DebugLogger() << "HumanClientApp::RequestSavePreviews Requesting previews for " << generic_directory;
-    Message response;
-    m_networking->SendSynchronousMessage(RequestSavePreviewsMessage(PlayerID(), generic_directory), response);
-    if (response.Type() == Message::DISPATCH_SAVE_PREVIEWS){
-        ExtractDispatchSavePreviewsMessageData(response, previews);
+    const auto response = m_networking->SendSynchronousMessage(RequestSavePreviewsMessage(PlayerID(), generic_directory));
+    if (response && response->Type() == Message::DISPATCH_SAVE_PREVIEWS){
```
geoffthemedio: space b4 {

client/human/HumanClientApp.cpp
```diff
@@ -684,13 +684,12 @@ void HumanClientApp::RequestSavePreviews(const std::string& directory, PreviewIn
...
+        ExtractDispatchSavePreviewsMessageData(*response, previews);
         DebugLogger() << "HumanClientApp::RequestSavePreviews Got " << previews.previews.size() << " previews.";
     }else{
```
geoffthemedio: no yours, but spaces around else

network/MessageQueue.cpp
```diff
@@ -10,8 +11,8 @@ namespace {
     };
 }
 
-MessageQueue::MessageQueue(boost::mutex& monitor) :
-    m_monitor(monitor)
+MessageQueue::MessageQueue(boost::mutex& monitor, const bool& rx_connected) :
+    m_monitor{monitor}, m_rx_connected{rx_connected}
```
geoffthemedio: prefer broken over lines for each item in initialization list

geoffthemedio:
appears to (still) work

LGM-Doyle:
Changes completed.

geoffthemedio:
seems like additional tests would be useful... at least to verify nothing was unexpectedly broken.

dbenage-cx:
Not certain if a valid test, to consistently get AI to hang, I tried:

```diff
diff --git a/default/python/AI/FreeOrionAI.py b/default/python/AI/FreeOrionAI.py
index d4435be..c38d13f 100644
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -3,6 +3,7 @@ these methods in turn activate other portions of the python AI code."""
 import pickle  # Python object serialization library
 import sys
 import random
+from time import sleep
 
 from common.configure_logging import redirect_logging_to_freeorion_logger
 redirect_logging_to_freeorion_logger()
@@ -165,6 +166,7 @@ def handleChatMessage(sender_id, message_text):  # pylint: disable=invalid-name
     # perhaps it is a debugging interaction
     if handle_debug_chat(sender_id, message_text):
         return
+    sleep(30)
     if not diplomatic_corp:
         DiplomaticCorp.handle_pregame_chat(sender_id, message_text)
     else:
```
On quickstart, typing text in message box and exiting before AI responds (via wm X) leaves 2 of 3 AI processes remaining: `AI_2` and `AI_3`
Restarting client with quickstart results in return to intro screen with error dialog `Connection to server has been lost`, along with console output noted in forum post.  After which game will successfully quickstart (no difference with master, at least on this platform).

geoffthemedio:
@dbenage-cx TLDR it works better, the same, or worse?

dbenage-cx:
@geoffthemedio the same

LGM-Doyle:
@dbenage-cx I tried the changes that you propose and they trigger a crash, but not the specific type of crash that this PR prevents.

The hang that I observed is that the AI client networking is shutdown by the server after a message is sent on a good TX connected channel, but before the synchronous response is received on the now closed RX channel.  The AI client then hangs waiting for the response.

I've had a difficult time thinking of a method to observe this PR "not hanging" because any message in the log file will be immediately overwritten by the newly loaded AI.

I made the following changes to write a sentinel file if the early exit portion of this PR is triggered.  It triggered after repeatedly and alternately loading two ~200 turn save games.  It does not trigger in regular game play.  So this PR works as intended.

p.s. Thanks for demonstrating how to colorize embedded "diff" code.

````diff
diff --git a/network/MessageQueue.cpp b/network/MessageQueue.cpp
index a11574e..8c9b95e 100644
--- a/network/MessageQueue.cpp
+++ b/network/MessageQueue.cpp
@@ -2,7 +2,11 @@
 
 #include "Message.h"
 
+#include "../util/Directories.h"
+
 #include <boost/optional/optional.hpp>
+#include <boost/filesystem/operations.hpp>
+#include <boost/filesystem/fstream.hpp>
 
 namespace {
     struct SynchronousResponseMessage {
@@ -58,8 +62,11 @@ boost::optional<Message> MessageQueue::GetFirstSynchronousMessage() {
     boost::mutex::scoped_lock lock(m_monitor);
     std::list<Message>::iterator it = std::find_if(m_queue.begin(), m_queue.end(), SynchronousResponseMessage());
     while (it == m_queue.end()) {
-        if (!m_rx_connected)
+        if (!m_rx_connected) {
+            boost::filesystem::ofstream os(GetUserDataDir() / "sparky_the_sentinel_file.txt");
+            os << "Would have hung.";
             return boost::none;
+        }
         m_have_synchronous_response.wait(lock);
         it = std::find_if(m_queue.begin(), m_queue.end(), SynchronousResponseMessage());
     }
````

LGM-Doyle:
I've marked this as cherry-pick for release.

This bug is triggered by reasonable and typical behavior, multiple re-loads, or an end turn followed by a quick load game.

The current workaround requires the user to find and kill the hung AI client.  The workaround may not be obvious to some players.  

The risk is that the problem and the fix are not obvious and may cause other problems.

Adding it to the release candidate will test it with a wider audience.

Vezzra:

LGM-Doyle:
Not to my knowledge.

Vezzra:
Ok, so unless anyone objects, I'm going to merge and cherry pick this before RC2, if no one else does it.

Question: has the bug being fixed by this been introduced after 0.4.6, or already been around before? So I can know if I need to add this to the changelog.

LGM-Doyle:
The bug has existed for a decade, but the changes to the shutdown procedure make it more likely to be triggered.

Vezzra:

PR #1471, opened 2017-04-07
Ai tech adjustments

Dilvish-fo:
adjustments for fighter techs as discussed in #633, as well as a few other tweaks.


Dilvish-fo:
heads-up-- basically right after submitting this PR I realized that the first commit in the group had overlooked  a couple more spots where corrections were needed, so I quickly added them to that commit to keep them altogether, and force-pushed the new set of commits to the branch.  Hopefully I got that in before anyone pulled, but if you did you'll probably want to just delete what you pulled and then re-pull.

Morlic-fo:
Looks good.

Vezzra:
Now *that* was fast 😄

PR #1472, opened 2017-04-08
Unused RP are automatically allocated to (unpaused) researchable techs

geoffthemedio:


Empire/Empire.cpp
```diff
@@ -166,6 +167,11 @@ namespace {
                 elem.allocated_rp = 0.0f;
             }
         }
+
+        if (total_RPs_spent >= RPs - EPSILON)
+            return; // don't further allocate if nothing left to spend
```
Dilvish-fo: The place these lines are located, after the end of the loop, it seems they are only avoiding the logging line that reports how much RP was spent, not actually stopping any allocation the way the comment indicates is intended.

If I move these lines up into the bottom of the loop so that they stop further allocation, then they cause a UI error:   start a new game, queue up a cheap tech like Planet Eco and then an expensive one like Radar.  Now shift Radar to first spot so that it soaks up all the available RP.  Planet Eco still shows as if it is getting RP allocated to it since the loop exited as soon as Radar used up the RP.  (The server prevents any actual overallocation during turn processing though.)

geoffthemedio: This early exit code is now pointless. It's left over from when I had the auto-allocation in SetTechQueueElementSpending, but it's been moved to CheckResearchProgress.

Dilvish-fo:
My first thought was that the idea seems fine/good to prevent folks, especially beginners, from losing out on RP, though a SitRep would be most helpful, both to bring the issue to the attention of the player and so that they could know which tech actually got the RP without having to add a bunch of techs to the queue to see which one had gotten some RP.

*But*, then I considered that we are talking about adding tech lines where researching one tech might lock you out of others, or at least make them extra expensive.  It would not be good to have this auto-allocation trigger any such result.  I imagine we could try adding ways to prevent that, but I would be concerned about their reliability.

geoffthemedio:
If such tech lines are added, this could be revisited, or restrictions could be added.

(I think there are better ways to make players choose between content options, however that's a forum discussion.)

Lack of a display of progress for techs not currently on the queue is arguably a separate issue.

I'm not so concerned about showing what techs are getting allocations automatically though, as if players care enough to know, they can just enqueue some techs.

geoffthemedio:
Objections to merging as is?

Vezzra:
As far as I'm concerned, go ahead.

Dilvish-fo:
Sounds fine to me.

PR #1475, opened 2017-04-08
Allow invasion through planet shields

geoffthemedio:
Allows invasion of planets that have shields, but with reduce effective troop numbers as the level of shielding increases. Could possibly be used as a feature or ability of a species, government policy, or type of troop pod.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -1431,7 +1431,9 @@ std::set<std::shared_ptr<const Ship>> AutomaticallyChosenInvasionShips(int targe
         if (!AvailableToInvade(ship, system_id, empire_id))
             continue;
 
-        invasion_troops += ship->TroopCapacity();
+        // how many troops are invading?
+        float effective_troops = ship->TroopCapacityAgainstShields(target_planet->CurrentMeterValue(METER_SHIELD));
+        invasion_troops += effective_troops;
 
         retval.insert(ship);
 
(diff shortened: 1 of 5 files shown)


MatGB:
This would need some serious balancing thought and probably significant AI work (they're having enough problems in this area until recently anyway): we deliberately added the 1 point shield amount to all colonies and outposts not that long ago and the basic "have shields = no invasion" is a nice simple mechanic.

UI/SidePanel.cpp
```diff
@@ -1643,7 +1645,7 @@ void SidePanel::PlanetPanel::Refresh() {
     bool at_war_with_me =   !mine && (populated || (has_owner && Empires().GetDiplomaticStatus(client_empire_id, planet->Owner()) == DIPLO_WAR));
 
     bool being_invaded =    planet->IsAboutToBeInvaded();
-    bool invadable =        at_war_with_me && !shielded && visible && !being_invaded && !invasion_ships.empty();
+    bool invadable =        at_war_with_me && /*!shielded && visible && */ !being_invaded && !invasion_ships.empty();
```
Dilvish-fo, line 1648: Did you really mean to take out the visibility requirement?

Vezzra:
This is an excellent example for a PR that should have an accompanying thread on one of the design forums, to avoid having the design discussion starting here in the comment section.

The idea definitely merits serious consideration and discussion, but I'll wait for the accompanying thread to be opened to add my 2c. 😃

Vezzra:
Link to design discussion on the forum:

http://www.freeorion.org/forum/viewtopic.php?f=6&t=10500

adrianbroher:
Did this ever came to a conclusion? This branch is now over a year old.

Vezzra:
No, but like #1091, we might at some point get back to this.

@geoffthemedio?

PR #1476, opened 2017-04-08
Adds Tests to Abort AI Functions when No Empire is Available

geoffthemedio:
I noticed this branch sitting without a pull request. I don't remember why I pushed it, but it was probably due to a forum thread or issue. I can't find which that was, though. Regardless, the safety checks added here might be useful, so I thought I'd rebase and make a new pull request.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -66,6 +66,10 @@ def startNewGame(aggression_input=fo.aggression.aggressive):  # pylint: disable=
     """Called by client when a new game is started (but not when a game is loaded).
     Should clear any pre-existing state and set up whatever is needed for AI to generate orders."""
     empire = fo.getEmpire()
+    if empire is None:
+        print "This client has no empire. Ignoring new game start message."
+        return
+    
     if empire.eliminated:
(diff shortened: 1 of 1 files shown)


default/python/AI/FreeOrionAI.py
```diff
@@ -202,6 +226,16 @@ def generateOrders():  # pylint: disable=invalid-name
     at end of this function, fo.doneTurn() should be called to indicate to the client that orders are finished
     and can be sent to the server for processing."""
     empire = fo.getEmpire()
+    if empire is None:
+        print "This client has no empire. Doing nothing to generate orders."
+        try:
+            # early abort if no empire. no need to do meter calculations
+            # on last-seen gamestate if nothing can be ordered anyway...
+            fo.doneTurn()
+        except Exception as e:
```
LGM-Doyle: This looks odd, but is correct.  I'd add a comment that `doneTurn()` is issued on behalf of the client network id, not the empire id, so not having a correct empire id does not invalidate `doneTurn()`

geoffthemedio: comment added

geoffthemedio:
rebased. futher comments / concerns?

LGM-Doyle:
LGTM

Issue #1400, opened 2017-03-15
Tone down the Solar Hull


Vezzra:
As noted in two threads on our forum, here:

http://www.freeorion.org/forum/viewtopic.php?f=15&t=10450

...and here:

http://www.freeorion.org/forum/viewtopic.php?p=87784#p87784

...the Solar Hull needs some toning down, as it's really too overpowered. For 0.4.7 only some quick and easy tweaking is possible, so I suggest:

* Reduce the internal slots to 5
* Reduce the external slots to 15
* Reduce speed to 60
* Reduce min build time to 8
* Increase buils costs to 160

Or something along those lines. These numbers are just some rough ideas, but should give a general idea what direction I think we should go.

Vezzra:
@freeorion/ai-team, please object if some of these changes would screw up the AIs ability to use the Solar Hull properly/effectively.

afwbkbc:
I suggest leaving them as they are but making them unrepairable. If it's damaged - it will stay damaged. Free drydocking them is insane.

geoffthemedio:
> Reduce min build time to 8

That doesn't sound like a nerf

Vezzra:
> That doesn't sound like a nerf

The reason for that change isn't so much nerfing, but enabling a player to counter another player with Solar Hulls a bit more quickly. See the discussions I linked to above, a major problem is once you learn another player is going to field Solar Hulls against you, you are potentially doomed already.

In that situation every turn counts, so being able to field your own Solar Hulls two turns earlier might help. Not much, but at least a little bit. Together with slowing the Solar Hull down this is supposed to make the whole issue less bad. Not good, not even close, but just less bad. That has to suffice for the release.

MatGB:
I did deliberately keep them longer to build than the others when I halved all the build times but am more than happy to see that reduced: another option I thought of was to unlock the solar hull building with an earlier tech (that's actually something I'd like to consider when we redo the tree, the buildings unlock earlier than the ships that use them).

I'm not too happy with reducing their slot totals too much, I am happy with increasing their costs, reducing their base structure and either reducing their speed a bit (they still need to be fast, they're the energy line flagship) or changing the way engines work/are costed.

I'm working through some options to test this week, but it does have to be done quickly: recosting the ships was something I had meant to do and completely forgot about, mea culpa.

afwbkbc:
2 turns will make no difference, maybe 5 can.
Adding min build time reduction tech branch is better solution.

[...]

PR #1477, opened 2017-04-08
Temp balance pass for Solar and other large hulls

MatGB:
A fix for #1400 for 0.4.7 release, increases costs for several larger hulls and removes several slots from the Solar.

The costs are probably about right, the Solar Hull changes aren't but will do for now: the big problem isn't the hull itself, but the insane speed is the big problem and that means reworking the engine parts to be better, more expensive but non-stacking.

The Solar Hull should have huge numbers of slots, but being able to get a base speed of 300 is a problem.

### DIFF
--- a/default/scripting/ship_hulls/asteroid/SH_SCATTERED_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_SCATTERED_ASTEROID.focs.txt
@@ -27,7 +27,7 @@ Hull
         Slot type = Internal position = (0.70, 0.50)
         Slot type = Core     position = (0.45, 0.50)
     ]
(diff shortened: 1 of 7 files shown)


MatGB:
@Vezzra as far as I'm concerned this can merge straight in but there's always a chance I've missed something, sorry for the delay.

dbenage-cx:
`SH_SOLAR_DESC` stringtable entry could use update for slot layout (and structure?).

MatGB:
Good point, thanks: I always leave stringtable updates till I'm done then sometimes forget to do them, it's a good reason to have the hull descriptions auto generated as much as possible but that's beyond me.

Vezzra (approved):
Yep, looks fine.

> sorry for the delay

No problem. Catching the flu is nasty after all... 😉

Issue #921, opened 2016-08-25
Loading ship designs from file doesn't work


l29ah:
I've created a design, used `Save Design File`, it created a file with the design in `~/.config/freeorion/shipdesigns/Дестроер.txt`, yet i don't see the design neither in `Finished`, nor in the `Saved` tabs in a new game.

Voker57:
works for me, but freeorion saved them in ~/.local/share/freeorion/shipdesigns/

dbenage-cx:
### Environment
- **FreeOrion Version**: 0.4.5+
- **Operating System**: Fedora 24
- **Graphic card used**: n/a
- **Fetched as** 
  - Compiled from source e017c3f

### Description

Ship designs saved during the current session are not available as saved designs without client restart.  These designs are also not loaded as pre-defined designs if a new game is started within the same session.
Closing the client and restarting resolves this issue.

### Expected Result

Saving a new ship design should result in immediate listing in the design window 'Saved' tab.
Saved designs should be available as pre-defined designs for new games in the current session (pending the option is set to the default state of True/1)

### Steps to reproduce
- Start new game with `--auto-add-saved-designs 1`
- Create and save a new ship design
- Design is not listed on the `Saved` tab
- Resign current game and start a new game.
- Previous design is not listed in `Saved` tab, nor available as pre-defined design.

Vezzra:
@geoffthemedio, why did you change the category from "bug" to "feature request"?

geoffthemedio:
Because it works as is when you restart the game. It's not a bug, it's a missing feature.

Vezzra:
Only if requiring a restart of the app is intended by the current implementation, which I guess it is judging by your statement. Otherwise it _would_ be a bug, as just staring a new game (without restarting the program) should be sufficient.

Although @Voker57's comment seems to indicate that for him this actually works - just starting a new game gives him the saved designs, so how does that work?

geoffthemedio:
Not "intended", but not completely broken.

I suspect @Voker57's comment is misleading.

Vezzra:
> Not "intended", but not completely broken.

That's still a bug and not a feature request then in my eyes. Either the current implementation is expected to require a program restart, or it's expected that just starting a new game should be sufficient. In the first case it's not a bug, in the latter case it is (IMO).

If we want to distinguish between "completely broken" things and just "impaired" things, we could introduce something like "category:glitch".

Voker57:
[...]

PR #1479, opened 2017-04-08
correct help message for auto-add-saved-designs option so that it matches actual behavior

Dilvish-fo:
As was pointed out below, I had gotten a bit mixed up about the option behavior; I have changed this PR to simply be for harmonizing the option help message with the actual behavior.

Previous (now stale) description:
It has long vexed me that this option did not actually accomplish its function (so instead, saved designs were required to be manually loaded each game), and I finally saw where to move things so that it does work.  

In the new location the attempted loading of saved designs can be triggered additional times during the first turn if new designs are added during that turn, but I think there is no harm from this-- it takes only a tiny fraction of a second and the SavedDesignsManager realizes that each one is already present and that nothing more need be done.

I am hoping this could make it into 0.4.7

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1348,7 +1348,7 @@ OPTIONS_DB_UI_AUTO_REPOSITION_WINDOWS
 Toggles whether to automatically reposition windows when the application size changes.
 
 OPTIONS_DB_AUTO_ADD_SAVED_DESIGNS
-At game start automatically add all user-saved ship designs to the player-empire known designs.
+Automatically add all user-saved ship designs to the player-empire known designs, if and only if the ShipDesign screen is visited on turn 1.
 
 OPTIONS_DB_DESIGN_PEDIA_DYNAMIC
 In the Design Window, dynamically update the pedia detail page while the design name is being edited.


geoffthemedio:
looks OK, though there's probably a better place for the reasons you noted.

dbenage-cx:
> this option did not actually accomplish its function (so instead, saved designs were required to be manually loaded each game)

Saved designs load consistent with option for me both with/without this PR.
Curious what portion does not work (other than #921 but not covered here)

Dilvish-fo:
> Saved designs load consistent with option for me both with/without this PR. 
Curious what portion does not work (other than #921 but not covered here.

bah, right, actually the aspect of it that is not working right is unchanged by this PR, turns out I had simply changed my own behavior while working on this which made it seem like this PR made a difference. 
 The aspect not really working right is that the autoloading will work if and only if you visit the ShipDesign Screen during the first turn (either with and without this PR ), which is inconsistent with the option description.   

Typically I don't visit the ShipDesign screen until after turn 1 and so was just in the habit of manually triggering the saved design loads via right click menu, and over a long period of time I was just left with the recollection that it wasn't working right and forgot that it actually could at least semi-auto work if I visited the ShipDesign screen on turn 1.  Even after I saw the code referencing turn 1, I still had it in my head that it wasn't working at all because I would check for the designs in my Build Selector rather than checking in the Design Screen.  Once I had moved the code, though, I knew I would have to visit the Design screen, so I got it in my head that the code move was what made it work.  Not feeling so bright right now.  *rueful grin*

Until we actually get it fixed correctly, I would propose changing the help message for this option from 

```At game start automatically add all user-saved ship designs to the player-empire known designs.```

to

```Automatically add all user-saved ship designs to the player-empire known designs, if and only if the ShipDesign screen is visited on turn 1.```

I suspect that to actually get it working at game start we would need to make a new public function in DesignWnd or something (my recollection is that in the initializer does not work because the player empire is not set at that point).

Vezzra:
> Until we actually get it fixed correctly, I would propose changing the help message for this option from 
>
>```At game start automatically add all user-saved ship designs to the player-empire known designs.```
>
>to
>
>```Automatically add all user-saved ship designs to the player-empire known designs, if and only if the ShipDesign screen is visited on turn 1.```

This sounds like a good quick fix for the release. It's very simple and straightforward. As far as I'm concerned, LGTM.

As a simple fix, if this doesn't get merged before the creation of the release branch, it is certainly eligible for being cherry picked if it goes in in time for RC1. Just make sure to either tag it for cherry-picking, or go ahead and do the cherry-picking yourself.

Or, if no one objects, I'm going to merge this myself before I create the release branch (well, I hope I don't forget about it 😉).

PR #1481, opened 2017-04-10
Add named loggers with individually controlled log levels.

LGM-Doyle:
### Merge Message 

Adds named loggers with individually controlled log levels.

### Motivation

There are currently several different ways to control which log messages are generated and printed in the log files
  * a command-line option --log-level that sets the log level for all log messages.
  * a verbose-combat-logging option in OptionsDB that gates logging of detailed combat mechanics and the combat log window.
  * a verbose-logging option in OptionsDB that gates combat and effects processing logging and other miscellaneous locations.
  * a compiled in option TRACE_EXECUTION used to gate logging in the networking and finite state machine (FSM) code.

  These can be configured/toggled from the following locations.

  Method | Command line | config.xml, persistent_config.xml | OptionsWnd | compiler only
  --| -- | -- | -- | --
  log-level | X | X | |
  verbose-logging | X | X | X |
  verbose-combat-logging | X | X | X |
  TRACE_EXECUTION |  |  | | X

### Goal
This PR creates a consistent system where it is easy to create/remove named loggers for specific sections of code that can be configured from the command-line, or OptionsDB or the OptionWnd at runtime.

### Usage
The default logger works the same as before.

Create a new named logger. 
```cpp
    CreateThreadedLogger(name_of_logger);
````

Create a log message.
```cpp
    ErrorLogger(name_of_logger) << "any streamable output, only computed "
                                << "at the error threshold or greater "
                                << "for the 'name_of_logger' logger.";
    WarnLogger(name_of_logger)  << "streamable output";
    InfoLogger(name_of_logger)  << "streamable output";
    DebugLogger(name_of_logger) << "streamable output";
    TraceLogger(name_of_logger) << "streamable output";
```

The `XLogger(name_of_logger) << ` is a macro that gates computation of everything to the right of the stream operator (`<<`) based on the runtime threshold of the specific logger.  So providing that all of the work assembling the log message is in functions to the right of the stream operator (`<<`) no work will be done for log messages that would not be printed in the log outputs.

(Optionallly) Set the log threshold in config.xml/persistent_config.xml
```xml
  <logging>
    <sources>
      <name_of_logger>warn</name_of_logger>
    </sources>
  </logging>
```

(Optionally) Set the log threshold from the OptionsWnd with the Other/Logging/Log threshold for individual sources/name_of_logger drop down list.

(Optionally) Override each executables defaut logger to specific log levels in  config.xml/persistent_config.xml.

```xml
  <logging>
    <execs>
      <ai>warn</ai>
      <client>debug</client>
      <server>trace</server>
    </execs>
  </logging>
```

(Optionally) Override each executables default logger to specific log levels from the OptionsWnd with the Other/Logging/Log threshold for default logger for process/<ai/client/server> drop down list.

(Optionally) Override all individual loggers of all executables log levels with the command-line/OptionsDB option

``` --log-level <error/warn/debug/info/debug/trace>```

### Changes in this PR.
The changes are in three steps: refactor `Logger` to allow names loggers, add ability to set logger thresholds from OptionsWnd, replace the various detailed loggers with the new named loggers.

### Additional Notes
- Adds a dependency on boost log_setup.
- Splits logger OptionsDB dependency into LoggerWOptionsDB so that logger is more suitable for a testing environment and doesn't need to pull in the entire OptionsDB framework.
- Moves year/month/day from each log line to the head of file to make space for more relevant information.
- OptionsDB now dumps messages about unrecognized options to the console because they are initialized during static initialization before the logger backend is configured.  I left those as is, so that this PR didn't change the intent of any log messages.  The messages should be changed as the current usage of OptionsDB no longer agrees with the expectation that all options be present and accounted for at static initialization.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1335,13 +1335,11 @@ namespace {
         }
 
         TraceLogger() << "Alignments:";
-        for (const Alignment& p : m_alignments) {
+        for (const auto& p : m_alignments)
             TraceLogger() << " ... " << p.Name();
-        }
         TraceLogger() << "Alignment Effects:";
-        for (std::shared_ptr<Effect::EffectsGroup> p : m_effects_groups) {
+        for (const auto& p : m_effects_groups)
             TraceLogger() << " ... " /*<< p.Dump()*/;
-        }
     }
 
     /** returns the singleton alignment manager */
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -2687,7 +2687,7 @@ void Empire::AddShipDesign(int ship_design_id, int next_design_id) {
 
             ShipDesignsChangedSignal();
 
-            TraceLogger() << "AddShipDesign::  " << ship_design->Name() << " ("<<ship_design_id
+            TraceLogger() << "AddShipDesign::  " << ship_design->Name() << " (" << ship_design_id
                           << ") to empire #" << EmpireID()
                           << (is_at_end_of_list ? " at end of list." : " in front of id ")
                           << next_design_id;
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1334,15 +1334,13 @@ namespace {
             throw e;
         }
 
-        if (GetOptionsDB().Get<bool>("verbose-logging")) {
-            DebugLogger() << "Alignments:";
-            for (const Alignment& p : m_alignments) {
-                DebugLogger() << " ... " << p.Name();
-            }
-            DebugLogger() << "Alignment Effects:";
-            for (std::shared_ptr<Effect::EffectsGroup> p : m_effects_groups) {
-                DebugLogger() << " ... " /*<< p.Dump()*/;
-            }
+        TraceLogger() << "Alignments:";
+        for (const Alignment& p : m_alignments) {
+            TraceLogger() << " ... " << p.Name();
+        }
+        TraceLogger() << "Alignment Effects:";
+        for (std::shared_ptr<Effect::EffectsGroup> p : m_effects_groups) {
+            TraceLogger() << " ... " /*<< p.Dump()*/;
         }
     }
 
@@ -2691,11 +2689,10 @@ void Empire::AddShipDesign(int ship_design_id, int next_design_id) {
 
             ShipDesignsChangedSignal();
 
-            if (GetOptionsDB().Get<bool>("verbose-logging"))
-                DebugLogger() << "AddShipDesign::  " << ship_design->Name() << " ("<<ship_design_id
-                              << ") to empire #" << EmpireID()
-                              << (is_at_end_of_list ? " at end of list." : " in front of id ")
-                              << next_design_id;
+            TraceLogger() << "AddShipDesign::  " << ship_design->Name() << " ("<<ship_design_id
+                          << ") to empire #" << EmpireID()
+                          << (is_at_end_of_list ? " at end of list." : " in front of id ")
+                          << next_design_id;
         }
     } else {
         // design in not valid
--- a/universe/Building.cpp
+++ b/universe/Building.cpp
@@ -330,7 +330,7 @@ BuildingTypeManager::BuildingTypeManager() {
     }
 
     TraceLogger() << "Building Types:";
-    for (const std::map<const std::string, BuildingType*>::value_type& entry : m_building_types) {
+    for (const auto& entry : m_building_types) {
         TraceLogger() << " ... " << entry.first;
     }
 
--- a/universe/Building.cpp
+++ b/universe/Building.cpp
@@ -320,9 +320,7 @@ BuildingTypeManager::BuildingTypeManager() {
     if (s_instance)
         throw std::runtime_error("Attempted to create more than one BuildingTypeManager.");
 
-    if (GetOptionsDB().Get<bool>("verbose-logging")) {
-        DebugLogger() << "BuildingTypeManager::BuildingTypeManager() about to parse buildings.";
-    }
+    TraceLogger() << "BuildingTypeManager::BuildingTypeManager() about to parse buildings.";
 
     try {
         parse::buildings(m_building_types);
@@ -331,11 +329,9 @@ BuildingTypeManager::BuildingTypeManager() {
         throw e;
     }
 
-    if (GetOptionsDB().Get<bool>("verbose-logging")) {
-        DebugLogger() << "Building Types:";
-        for (const std::map<const std::string, BuildingType*>::value_type& entry : m_building_types) {
-            DebugLogger() << " ... " << entry.first;
-        }
+    TraceLogger() << "Building Types:";
+    for (const std::map<const std::string, BuildingType*>::value_type& entry : m_building_types) {
+        TraceLogger() << " ... " << entry.first;
     }
 
     // Only update the global pointer on sucessful construction.
--- a/universe/Effect.cpp
+++ b/universe/Effect.cpp
@@ -367,10 +367,10 @@ void SetMeter::Execute(const TargetsCauses& targets_causes, AccountingMap* accou
         TargetSet                           targets               = targets_and_cause.target_set;
 
 
-        DebugLogger(effects) << "\n\nExecute SetMeter effect: \n" << Dump();
-        DebugLogger(effects) << "SetMeter execute targets before: ";
+        TraceLogger(effects) << "\n\nExecute SetMeter effect: \n" << Dump();
+        TraceLogger(effects) << "SetMeter execute targets before: ";
         for (std::shared_ptr<UniverseObject> target : targets)
-            DebugLogger(effects) << " ... " << target->Dump();
+            TraceLogger(effects) << " ... " << target->Dump();
 
         if (!accounting_map) {
             // without accounting, can do default batch execute
@@ -416,9 +416,9 @@ void SetMeter::Execute(const TargetsCauses& targets_causes, AccountingMap* accou
             }
         }
 
-        DebugLogger(effects) << "SetMeter execute targets after: ";
+        TraceLogger(effects) << "SetMeter execute targets after: ";
         for (std::shared_ptr<UniverseObject> target : targets)
-            DebugLogger(effects) << " ... " << target->Dump();
+            TraceLogger(effects) << " ... " << target->Dump();
     }
 }
 
@@ -582,16 +582,16 @@ void SetShipPartMeter::Execute(const TargetsCauses& targets_causes, AccountingMa
         const TargetsAndCause&              targets_and_cause     = targets_entry.second;
         TargetSet                           targets               = targets_and_cause.target_set;
 
-        DebugLogger(effects) << "\n\nExecute SetShipPartMeter effect: \n" << Dump();
-        DebugLogger(effects) << "SetShipPartMeter execute targets before: ";
+        TraceLogger(effects) << "\n\nExecute SetShipPartMeter effect: \n" << Dump();
+        TraceLogger(effects) << "SetShipPartMeter execute targets before: ";
         for (std::shared_ptr<UniverseObject> target : targets)
-            DebugLogger(effects) << " ... " << target->Dump();
+            TraceLogger(effects) << " ... " << target->Dump();
 
         Execute(source_context, targets);
 
-        DebugLogger(effects) << "SetShipPartMeter execute targets after: ";
+        TraceLogger(effects) << "SetShipPartMeter execute targets after: ";
         for (std::shared_ptr<UniverseObject> target : targets)
-            DebugLogger(effects) << " ... " << target->Dump();
+            TraceLogger(effects) << " ... " << target->Dump();
     }
 }
 
--- a/universe/Effect.cpp
+++ b/universe/Effect.cpp
@@ -30,7 +30,7 @@
 #include <iterator>
 
 namespace {
-    CreateThreadedLogger(effects);
+    DeclareThreadSafeLogger(effects);
 }
 
 using boost::io::str;
--- a/universe/Effect.cpp
+++ b/universe/Effect.cpp
@@ -29,6 +29,9 @@
 #include <cctype>
 #include <iterator>
 
+namespace {
+    CreateThreadedLogger(effects);
+}
 
 using boost::io::str;
 
@@ -351,8 +354,6 @@ void SetMeter::Execute(const TargetsCauses& targets_causes, AccountingMap* accou
                        bool only_meter_effects, bool only_appearance_effects,
                        bool include_empire_meter_effects, bool only_generate_sitrep_effects) const
 {
-    bool log_verbose = GetOptionsDB().Get<bool>("verbose-logging");
-
     if (only_appearance_effects || only_generate_sitrep_effects)
         return;
 
@@ -365,15 +366,17 @@ void SetMeter::Execute(const TargetsCauses& targets_causes, AccountingMap* accou
         const TargetsAndCause&              targets_and_cause     = targets_entry.second;
         TargetSet                           targets               = targets_and_cause.target_set;
 
-        if (log_verbose) {
-            DebugLogger() << "\n\nExecute SetMeter effect: \n" << Dump();
-            DebugLogger() << "SetMeter execute targets before: ";
+        DebugLogger(effects) << [&targets, this]() {
+            std::stringstream ss;
+            ss << "\n\nExecute SetMeter effect: \n" << Dump();
+            ss << "SetMeter execute targets before: ";
             for (std::shared_ptr<UniverseObject> target : targets)
-                DebugLogger() << " ... " << target->Dump();
-        }
+                ss << " ... " << target->Dump();
+            return ss.str();
+        }();
 
-        if (!accounting_map && !log_verbose) {
-            // without accounting and without verbose logging, can do default batch execute
+        if (!accounting_map) {
+            // without accounting, can do default batch execute
             Execute(source_context, targets);
 
         } else if (!accounting_map) {
@@ -416,12 +419,14 @@ void SetMeter::Execute(const TargetsCauses& targets_causes, AccountingMap* accou
             }
         }
 
-        if (log_verbose) {
-            DebugLogger() << "SetMeter execute targets after: ";
+        DebugLogger(effects) << [&targets]() {
+            std::stringstream ss;
+            ss << "SetMeter execute targets after: ";
             for (std::shared_ptr<UniverseObject> target : targets) {
-                DebugLogger() << " ... " << target->Dump();
+                ss << " ... " << target->Dump();
             }
-        }
+            return ss.str();
+        }();
     }
 }
 
@@ -573,8 +578,6 @@ void SetShipPartMeter::Execute(const TargetsCauses& targets_causes, AccountingMa
                                bool only_meter_effects, bool only_appearance_effects,
                                bool include_empire_meter_effects, bool only_generate_sitrep_effects) const
 {
-    bool log_verbose = GetOptionsDB().Get<bool>("verbose-logging");
-
     if (only_appearance_effects || only_generate_sitrep_effects)
         return;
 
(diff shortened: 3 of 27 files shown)

### CODE REVIEW

**Comment by LGM-Doyle** [comment]
It now compiles on OSX and Windows, as well as the original linux.

It needs testing on those OSs, because OSX and particularly Windows required additional coding.

Testing is as follows:
- quickstart a game
- check that a log file has generated some log messages
- go to `OptionsWnd/Other/Logging` and change some options to generate more/fewer log reports.
- check in the log file that the log reports you expected to be present/absent are present/absent.

Thanks in advance for your testing time.

**Comment by geoffthemedio** [comment]
This seems overly complicated to me... In particular, are 5 log levels really necessary?

I find the lambda function syntax for controlling whether to evaluate stuff to log particularly clunky / awkward. A simple if block seems preferable. Can this be easily integrated with the named loggers, perhaps to poll whether a particular named logger has a particular log level?

The pull request is another huge one, which is difficult to review. Can you break it into smaller chunks?

**Comment by LGM-Doyle** [comment]
I will start start breaking it into smaller chunks.  

> In particular, are 5 log levels really necessary?

@dbenage-cx provided two external discussions on levels of logging in the [forum](http://freeorion.org/forum/viewtopic.php?f=9&t=10507&sid=6643eebed8606a8c410ae3cb97ba64a0)

I think that at least 4 levels are necessary.

We should have 2 levels of problem log messages, error and warning.  Errors are a notice to the devs that something should be fixed in the code.  (Ideally) there should be no errors in released code. Warnings are a notice to the user that something is in a downgraded state, like a missing sound system.  

We need at least 2 levels of progress/state log messages for developers, debug and trace.  This distinction is already apparent in the existing code with `verbose-logging`, `verbose-combat-logging` and `TRACE_EXECUTION`.  Collapsing these two levels into one level would mean that when debugging the network code the options are no logs, or many logs messages per exchange between the client and the server.

> Can this be easily integrated with the named loggers, perhaps to poll whether a particular named logger has a particular log level?

No it can't be done easily.  It would involve replicating parts of the `boost::logging` API macros to extract that state from the backend.  It would be dependent on internals of `boost::logging` and consequently be brittle.

However, most of the lamda functions are in the trace logging of the effects code.  Most are only preventing unnecessary iteration over some set, when there will be no log generated.  They can be removed.

**Comment by dbenage-cx** [comment]
Briefly checked this PR earlier, really like some of the changes.

If nothing else changes, please set the default severity of `effects` to *Info* (or change many to trace).  I only played to ~turn 34 with default severities and freeorion.log was >655MB.

No config option for `combat_log` is present in config.xml unless set at command line/persistent config.
Would prefer that any options always have a config entry.

**Comment by adrianbroher** [comment]
> The XLogger(name_of_logger) << is a macro that gates computation of everything to the right of the stream operator (<<) based on the runtime threshold of the specific logger. 

Did you check if this is actually true?

http://www.boost.org/doc/libs/1_59_0/libs/log/doc/html/log/rationale/why_not_lazy_streaming.html

Further remarks after the PR was shrinked to a consumable size.

**Comment by dbenage-cx** [comment]
@adrianbroher I believe the wording there suggests that `packet.size()` would always be called only if "lazy streaming" was implemented and done so without additional work.

The first doc mention I see confirming this is the note at http://www.boost.org/doc/libs/1_59_0/libs/log/doc/html/log/tutorial/trivial_filtering.html

In practice, [this diff](https://gist.github.com/dbenage-cx/79b1ac68acc6b02425b4211539ad6774)(print to console when sound enabled) also appears to exhibit the functionality described by LGM-Doyle.

**Comment by LGM-Doyle** [comment]
Thanks everyone for the comments.  I'm working on the fixes and the split.

@dbenage-cx I changed the effects priority to `trace`.  It should have been `trace`, but I had removed the `info` level and bumped the effects logs up to debug while considering what the "right" number of levels was.

I'm working on the `combat_log` logger not appearing in the `config.xml` file.  

@adrianbroher I think @dbenage-cx's example is a perfect answer to your question.

I now plan to do some "playtesting" of the release candidate.

**Comment by LGM-Doyle** [comment]
I've now broken this up into multiple parts.

This PR is now the last part.

This PR replaces the various other verbose or trace logging methods with a named `TraceLogger(<a_name>)`

It follows the core implementation #1529. 

***Merge Message***
Replace other verbose or trace logging methods with a TraceLogger(a_name).

**Comment by LGM-Doyle** [comment]
I can see that the clang build is failing.  I will work on that.

**Comment by LGM-Doyle** [comment]
> @dbenage-cx wrote:
> No config option for combat_log is present in config.xml unless set at command line/persistent config.
Would prefer that any options always have a config entry.

The loggers are lazy, multi-threaded and designed to work during static initialization.  A logger does not exist until the first time it sends a log to a backend, which is why it doesn't appear in config.xml.

The loggers are also intentionally not tied to OptionsDB, so that they can be useful in testing frameworks without being dependent on local config.xml state, which would mean tests can't runs in parallel or independently.

However, I added some `RegisterLoggerWithOptionsDB()` statements to the `HumanClientApp` for the standard named loggers so that they will always appear in the config.xml.

This does not prevent adhoc loggers from being declared and used anywhere in the code.  Adhoc loggers will correctly register with OptionsDB automatically.

**Comment by geoffthemedio** [comment]
What does
> Adhoc loggers will correctly register with OptionsDB automatically.

mean?

**Comment by LGM-Doyle** [comment]
It means that you don't need to do anything more than ,
````
    CreateThreadedLogger(adhoc_name);
    DebugLogger(adhoc_name);
````
anywhere in the code, server, human or ai clients, to have the logger named "adhoc_name" show up in config.xml and in the `OptionsWnd`.  It also means that once you delete those lines the logger is gone and config.xml will clean itself up automatically.  I wanted to be clear that the registration in the human client was only required to force the registration before first use.

**Comment by geoffthemedio** [comment]
I think there's enough log stuff to warrant its own tab.

I don't think the tooltip needs to list the log levels in order explicitly. This is clear enough from the droplists and their names.

The subsection titles are too long. Also not clear... perhaps "Process Log Thresholds" and "Code Section Log Thresholds". Then have tooltips explaining what overrides what and when.

Why don't I have server and AI client options?

log_options

**Comment by LGM-Doyle** [comment]
I move the loggers to a new `OptionsWnd` section called `Logs`.
All 3 executables default loggers should always be visible in the `OptionsWnd`

**Comment by geoffthemedio** [comment]
If I change one debug option, I get a bunch of log level setting log entries unrelated to that change. Eg. FSM from debug to info, and the client log contains
````
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "ai" logger threshold to "debug".
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "client" logger threshold to "info".
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "server" logger threshold to "debug".
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "FSM" logger threshold to "info".
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "combat" logger threshold to "info".
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "combat_log" logger threshold to "info".
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "effects" logger threshold to "info".
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "log" logger threshold to "info".
10:43:19.481843 [info] log : Logger.cpp:280 : Setting "network" logger threshold to "info".
````

Might be some problems with string formats, as seen in some log entries:
`MapWnd::InitTurnRendering time:     1730 Âµs`

**Comment by dbenage-cx** [comment]
Intermittent with boost 1.64.0, when resigning from game:
```tty
terminate called after throwing an instance of 'terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::log::v2_mt_posix::system_error> >'
  what():  Failed to set TLS value: Invalid argument
```

Have not been able to reoccur under gdb, this message looks like it was added in 1.62.0
[src/thread_specific.cpp](https://fossies.org/diffs/boost/1_61_0_vs_1_62_0/libs/log/src/thread_specific.cpp-diff.html)

LGM-Doyle:
@geoffthemedio I've made it only set options that have changed in `LoggerWithOptionsDB`, which fixes the redundant messages.

@dbenage-cx I thought that it was possible that logging system would crash during static deinitialization, but for a different reason.  

I've included a possible fix. If you can reliably generate the crash, could you report if this fixes it.

The bug that I expected is that the file logger portion of `boost::log` crashes during static de-initialization because `boost::filesystem` uses a global static `std::locale`, which may/may not be destroyed before the logs are flushed during their destructors.  I was unable to intentionally cause this crash and I did not include a "fix" to a non-existent problem.

Your problem seems slightly different.  It appears that the thread local storage is being exhausted during static de-initialization.  

I've added the fix that I would have used for the former expected problem and I am hoping that it fixes your reported problem.

The fix is to shutdown the file sinks before the end of `main()`.  Hence there will be no problems with the order of static de-initialization of the file sink, filesystem and locale.  However, the down side is that log messages from destructors that run after the end of `main()` are dumped to the console.

Thanks for testing.

dbenage-cx:
@LGM-Doyle I could not reliably generate the message, I would estimate 1 instance in <10 attempts.
I was not successful in generating it with the changes.

To the `setting "x" logger threshold to logger "y"` messages, imo these would fail to meet criteria of `info`.  But that is a very minor concern in this case and not many have weighed in on such discussion.

Issue #1536 is more of a concern with this PR.

util/Logger.cpp
```diff
@@ -318,6 +325,19 @@ void InitLoggingSystem(const std::string& log_file, const std::string& _unnamed_
...
+    // The file sink may not be safe to use during static deinitialization, because of the
+    // following bug:
+
+    // http://www.boost.org/doc/libs/1_64_0/libs/log/doc/html/log/rationale/why_crash_on_term.html
+    // https://svn.boost.org/trac/boost/ticket/8642
+    // https://svn.boost.org/trac/boost/ticket/9119
+
+    // When either ticket is fixed the ShutdownLoggingSystem() function can be removed.
+
+    GetLoggersToSinkFrontEnds().ShutdownFileSinks();
```
dbenage-cx: Are all file sinks removed, or only those for the relevant process?
e.g. logging of the client stops when the server is stopped

LGM-Doyle: The file sink is actually, one backend coupled to a frontend per logger.  All of the frontends are shutdown, which shutsdown the single backend.  

The sinks then fall back to the `std::cout` default sink, which is why the logs that occur after the end of `main()` are dumped to the console.  Dumping the logs to the console is ugly, but better than burying potentially useful crash debugging log messages.

The human client does not also close the file sinks for the server or the AIs.  They do that on their own.

LGM-Doyle:
@dbenage-cx, I understand your point.

I had assumed that most people would dial the threshold down to warning for the `log` logger.

The following point is for discussion purposes only and I'm not suggesting changing this PR.

Perhaps the log levels, info, debug and trace should not have universal meaning, but meaning within the scope of code that the named logger is covering.  Otherwise, some of the named loggers will only be able to produce logs in a compressed range.  

For example, currently the `effects` logger only logs at the trace level.  This makes sense from a high level perspective, but it means that the `effects` logger is not as useful as it could be to a dev working on the effects code if they had access to the full range of non-error levels.

We could add a `player_info` named logger for player directed log message.  The messages could possibly be translated. This channel should be player directed, with no code references.  For the production release we could ship with warning and errors on for all loggers and the info level on for the `player_info` logger.

dbenage-cx:
@LGM-Doyle So it is clear we are not proposing changes in this PR (and someone does not come chasing after us 😉 ), I've replied in forum [Log Severity Levels thread](http://www.freeorion.org/forum/viewtopic.php?f=9&t=10507&p=88552#p88552)

LGM-Doyle:
Rebased.

LGM-Doyle:
Rebased.

Empire/Empire.cpp
```diff
@@ -2691,11 +2689,10 @@ void Empire::AddShipDesign(int ship_design_id, int next_design_id) {
 
             ShipDesignsChangedSignal();
 
-            if (GetOptionsDB().Get<bool>("verbose-logging"))
-                DebugLogger() << "AddShipDesign::  " << ship_design->Name() << " ("<<ship_design_id
-                              << ") to empire #" << EmpireID()
-                              << (is_at_end_of_list ? " at end of list." : " in front of id ")
-                              << next_design_id;
+            TraceLogger() << "AddShipDesign::  " << ship_design->Name() << " ("<<ship_design_id
```
geoffthemedio: spaces around last <<

Empire/Empire.cpp
```diff
@@ -1334,15 +1334,13 @@ namespace {
...
-            for (const Alignment& p : m_alignments) {
-                DebugLogger() << " ... " << p.Name();
-            }
-            DebugLogger() << "Alignment Effects:";
-            for (std::shared_ptr<Effect::EffectsGroup> p : m_effects_groups) {
-                DebugLogger() << " ... " /*<< p.Dump()*/;
-            }
+        TraceLogger() << "Alignments:";
+        for (const Alignment& p : m_alignments) {
+            TraceLogger() << " ... " << p.Name();
```
geoffthemedio: can put { } around single line of code on same line

universe/Building.cpp
```diff
@@ -333,11 +331,9 @@ BuildingTypeManager::BuildingTypeManager() {
         throw e;
     }
 
-    if (GetOptionsDB().Get<bool>("verbose-logging")) {
-        DebugLogger() << "Building Types:";
-        for (const std::map<const std::string, BuildingType*>::value_type& entry : m_building_types) {
-            DebugLogger() << " ... " << entry.first;
-        }
+    TraceLogger() << "Building Types:";
+    for (const std::map<const std::string, BuildingType*>::value_type& entry : m_building_types) {
```
geoffthemedio: looks like a good candidate for `const auto&`

universe/Encyclopedia.cpp
```diff
@@ -21,12 +21,10 @@ Encyclopedia::Encyclopedia() :
...
 
-    if (GetOptionsDB().Get<bool>("verbose-logging")) {
-        DebugLogger() << "(Category) Encyclopedia Articles:";
-        for (std::map<std::string, std::vector<EncyclopediaArticle>>::value_type& entry : articles) {
-            const std::string& category = entry.first;
-            for (const EncyclopediaArticle& article : entry.second)
-            { DebugLogger() << "(" << category << ") : " << article.name; }
-        }
+    TraceLogger() << "(Category) Encyclopedia Articles:";
+    for (std::map<std::string, std::vector<EncyclopediaArticle>>::value_type& entry : articles) {
```
geoffthemedio: `auto`?

universe/Field.cpp
```diff
@@ -237,11 +237,9 @@ FieldTypeManager::FieldTypeManager() {
...
     }
 
-    if (GetOptionsDB().Get<bool>("verbose-logging")) {
-        DebugLogger() << "Field Types:";
-        for (const std::map<std::string, FieldType*>::value_type& entry : *this) {
-            DebugLogger() << " ... " << entry.first;
-        }
+    TraceLogger() << "Field Types:";
+    for (const std::map<std::string, FieldType*>::value_type& entry : *this) {
+        TraceLogger() << " ... " << entry.first;
```
geoffthemedio: {} on one line?

universe/ShipDesign.cpp
```diff
@@ -124,12 +124,10 @@ PartTypeManager::PartTypeManager() {
...
     }
 
-    if (GetOptionsDB().Get<bool>("verbose-logging")) {
-        DebugLogger() << "Part Types:";
-        for (const std::map<std::string, PartType*>::value_type& entry : m_parts) {
-            const PartType* p = entry.second;
-            DebugLogger() << " ... " << p->Name() << " class: " << p->Class();
-        }
+    TraceLogger() << "Part Types:";
+    for (const std::map<std::string, PartType*>::value_type& entry : m_parts) {
```
geoffthemedio: `auto` ?

universe/ShipDesign.cpp
```diff
@@ -484,12 +482,10 @@ HullTypeManager::HullTypeManager() {
...
     }
 
-    if (GetOptionsDB().Get<bool>("verbose-logging")) {
-        DebugLogger() << "Hull Types:";
-        for (const std::map<std::string, HullType*>::value_type& entry : m_hulls) {
-            const HullType* h = entry.second;
-            DebugLogger() << " ... " << h->Name();
-        }
+    TraceLogger() << "Hull Types:";
+    for (const std::map<std::string, HullType*>::value_type& entry : m_hulls) {
```
geoffthemedio: `auto` ?

universe/ShipDesign.cpp
```diff
@@ -1075,17 +1071,15 @@ PredefinedShipDesignManager::PredefinedShipDesignManager() {
...
-            const ShipDesign* d = entry.second;
-            DebugLogger() << " ... " << d->Name();
-        }
-        DebugLogger() << "Monster Ship Designs:";
-        for (const std::map<std::string, ShipDesign*>::value_type& entry : m_monster_designs) {
-            const ShipDesign* d = entry.second;
-            DebugLogger() << " ... " << d->Name();
-        }
+    TraceLogger() << "Predefined Ship Designs:";
+    for (const std::map<std::string, ShipDesign*>::value_type& entry : m_ship_designs) {
```
geoffthemedio: `auto` ?

universe/ShipDesign.cpp
```diff
@@ -1075,17 +1071,15 @@ PredefinedShipDesignManager::PredefinedShipDesignManager() {
...
+        const ShipDesign* d = entry.second;
+        TraceLogger() << " ... " << d->Name();
+    }
+    TraceLogger() << "Monster Ship Designs:";
+    for (const std::map<std::string, ShipDesign*>::value_type& entry : m_monster_designs) {
```
geoffthemedio: `auto` ?

universe/Special.cpp
```diff
@@ -23,11 +23,9 @@ namespace {
                 throw e;
             }
 
-            if (GetOptionsDB().Get<bool>("verbose-logging")) {
-                DebugLogger() << "Specials:";
-                for (const std::map<std::string, Special*>::value_type& entry : m_specials) {
-                    DebugLogger() << " ... " << entry.first;
-                }
+            TraceLogger() << "Specials:";
+            for (const std::map<std::string, Special*>::value_type& entry : m_specials) {
```
geoffthemedio: `auto` ?

universe/Species.cpp
```diff
@@ -330,16 +330,14 @@ SpeciesManager::SpeciesManager() {
...
-        for (const std::map<std::string, Species*>::value_type& entry : m_species) {
-            const Species* s = entry.second;
-            DebugLogger() << " ... " << s->Name() << "  \t" <<
-                (s->Playable() ?        "Playable " : "         ") <<
-                (s->Native() ?          "Native " : "       ") <<
-                (s->CanProduceShips() ? "CanProduceShips " : "                ") <<
-                (s->CanColonize() ?     "CanColonize " : "            ");
-        }
+    TraceLogger() << "Species:";
+    for (const std::map<std::string, Species*>::value_type& entry : m_species) {
```
geoffthemedio: `auto` ?

universe/Universe.cpp
```diff
@@ -949,20 +944,25 @@ namespace {
...
 
+    std::string StoreTargetsAndCausesOfEffectsGroupsWorkItem::GenerateReport() const {
+        boost::unique_lock<boost::shared_mutex> guard(*m_global_mutex);
+        std::stringstream ss;
+        ss << "StoreTargetsAndCausesOfEffectsGroups: effects_group: "
+           << m_effects_group->AccountingLabel()
+           << "  specific_cause: " << m_specific_cause_name
+           << "  sources: ";
+        for (std::shared_ptr<const UniverseObject> obj : *m_sources) {
+            ss << obj->Name() << " (" << std::to_string(obj->ID()) << ")  ";
```
geoffthemedio: {} to same line

LGM-Doyle (commented):
Thanks for looking at this. 
I've made the changes.

geoffthemedio:
I feel like the categorization of the log level options is unnecessarily confusing... It's not at all clear in the options tab or tooltips how or why there are separate per-process and per-source options, and whether or how these interact or override eachother...

As far as I can tell, the per-process options work like each process type has a different default source, and the client/ai/server process settings don't otherwise interact with the more specialized sources at all. So, I would reorganize and rename the sources, so that there are just one group of source log levels, and call the first three "client general", "ai general", and "server general".

LGM-Doyle:
I've converted it as requested.  I left the labels in the log themselves as ai, server and client.

universe/Effect.cpp
```diff
@@ -365,15 +366,14 @@ void SetMeter::Execute(const TargetsCauses& targets_causes, AccountingMap* accou
...
         TargetSet                           targets               = targets_and_cause.target_set;
 
-        if (log_verbose) {
-            DebugLogger() << "\n\nExecute SetMeter effect: \n" << Dump();
-            DebugLogger() << "SetMeter execute targets before: ";
-            for (std::shared_ptr<UniverseObject> target : targets)
-                DebugLogger() << " ... " << target->Dump();
-        }
 
-        if (!accounting_map && !log_verbose) {
```
geoffthemedio, line 375: Changing `if (!accounting_map && !log_verbose)` to just `if (!accounting_map)` means that the following `else if (!accounting_map)` case will never execute. Something doesn't make sense here...

Issue #1483, opened 2017-04-11
stealthed monster-carriers can result in buggy-looking combat reports


Dilvish-fo:
Presumably the following is from a stealthed medium krill moving into a system with one of my colonies. 
 The summary of forces is just my planet and zero neutral forces; nevertheless the zero neutral forces manage to launch fighters (which then do nothing)
![buggy_combat_report](https://cloud.githubusercontent.com/assets/10995939/24925304/a9620ed2-1eac-11e7-8076-a8ca39a4757b.png)

I am pretty sure that we had decided a long while back that armed monsters should not be stealthed, and I think that should apply to monster-carriers also.  So I would propose that we make that change (which I think is a correction really) and cherry pick it for 0.4.7

I also have a recollection that we recently discussed having fighter-launch cause enemies present to gain visibility the same as if they had been attacked by direct fire weapons, but I can't seem to locate that discussion now.  My recollection is that we talked about making it be some-chance-of-losing stealth and punted it for some later stealth re-work.  I would be pretty inclined myself to have the interim position be that launching fighters grants visibility just like a direct attack.

Also just a note about my labels here: I am categorizing the more specific issue as a bug regarding content scripting for armed monsters, and the suggestion about combat visibility in general as a game mechanic tweak.

Vezzra:
> I also have a recollection that we recently discussed having fighter-launch cause enemies present to gain visibility the same as if they had been attacked by direct fire weapons, but I can't seem to locate that discussion now.

http://www.freeorion.org/forum/viewtopic.php?p=88085#p88085

I'd consider stealthy carriers being overpowered as a "known issue" for 0.4.7, and postpone a more thorough fix until after the release. I expect a major part of 0.4.8 being fixes and adjustments for the new fighter/carrier mechanic anyway.

The combat report does look a bit weird, but then, considering you were fighting an enemy that didn't break stealth, it's also somehow fitting. I mean, you don't know where the fighters came from, only that they have been there. So no estimates on the numbers of the mysterious thing(s) you couldn't see...

> I am pretty sure that we had decided a long while back that armed monsters should not be stealthed

We did? Can't remember right now... but that's probably my faulty memory. Anyway, we can't really prevent that, as monsters could get increased stealth e.g. from ion storms.

Dilvish-fo:
> We did? Can't remember right now.

[...]

PR #1485, opened 2017-04-13
adjust pedia descriptions of Dyson Forests and Floaters to link to…

Dilvish-fo:
…each other and (as a helpful tip for beginning players) note that Active Radar is generally required to detect Floaters, as I suggested in #1483.

(side note: I am figuring that Pedia entries call for the "component:user manual" label; someone please correct that if I am wrong.)


Vezzra (approved):
Looks fine. Will merge before RC1 if no one objects.

PR #1488, opened 2017-04-13
fix doubleclick on production queue item so that it completes the map…

### LINKED ISSUE #1486: double clicking a production queue item causes an incomplete UI update
most specifically, the MapWnd active system indicator and the color highlighting of queue item locations will not update even though the BuildDesignator system has changed.   This was [raised on our forums](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10510) by Oberlus: 
> When you have the production queue open and you click on a system in the map, the production entries corresponding to the planets of that system get the planet's name highlighted (plus you hear that soft sound for selecting sth and you see the planets of that system in the right panel). I love that feature, makes things easier.
If you single-click on an entry of the production queue the map focus on the corresponding system but the selected system/fleet does not change. Again lovable.
If you double-click on a production entry the map focus change, the planets of that system show up at your right and you have effectively selected that planet from that panel (you even hear that sound for changing selected object in the map). However (here comes the possible minor bug), the system doesn't show as selected (there is no white ring around it, it keeps circling on the previously selected object in the map) and the planets names in the production queue aren't highlighted. Then, if you try to select the system in the map (or the planet in the right panel, that is already selected) to get that helpful highlight you get nothing, you (double-)click on the planet to no avail. I realised that the planet is actually selected but the GUI is not showing the change. So when you click on the planet, as it is already selected, it does nothing. If you select another system and then back the one you want, you get the corresponding production entries highlighted as desired. The workaround is to not use the double-click on the queue list, just single-click (to locate the system in the map) and then select the system in the map.
Also, it seems I can't get screenshots from within the game (fullscreen), I just get a black box to paste. I'm on Windows 7 (I thought I was Windows 10, I'm so noob with M$ systems...).

I have confirmed this on Linux, and also found that this can lead to the MapWnd showing the system selection indicator circle active on two different systems simultaneously.   I think the problem stems from ProductionWnd::QueueItemDoubleClickedSlot() calling BuildDesignatorWnd::CenterOnBuild(), which among other things calls BuildDesignatorWnd::SelectSystem() which calls SidePanel::SetSystem() to select the system and which also selects the planet;  this all is a subset of what happens when clicking on a MapWnd system with the production window open.  The latter triggers MapWnd::SelectSystem(), which changes the system icon selection as well as triggering some other updating.

I will explore having BuildDesignatorWnd::CenterOnBuild() call to MapWnd::SelectSystem() rather than skipping over steps directly to BuildDesignatorWnd::SelectSystem().

Dilvish-fo:
… window system selection. Fixes #1486.

### DIFF
--- a/UI/BuildDesignatorWnd.cpp
+++ b/UI/BuildDesignatorWnd.cpp
@@ -1051,7 +1051,7 @@ void BuildDesignatorWnd::CenterOnBuild(int queue_idx, bool open) {
             MapWnd* map = ClientUI::GetClientUI()->GetMapWnd();
             map->CenterOnObject(system_id);
             if (open) {
-                SelectSystem(system_id);
+                HumanClientApp::GetApp()->GetClientUI().GetMapWnd()->SelectSystem(system_id);
                 SelectPlanet(location_id);
             }
         }


UI/BuildDesignatorWnd.cpp
```diff
@@ -1051,7 +1051,8 @@ void BuildDesignatorWnd::CenterOnBuild(int queue_idx, bool open) {
             MapWnd* map = ClientUI::GetClientUI()->GetMapWnd();
             map->CenterOnObject(system_id);
             if (open) {
-                SelectSystem(system_id);
+                HumanClientApp::GetApp()->GetClientUI().GetMapWnd()->SelectSystem(system_id);
+                //SelectSystem(system_id);
```
adrianbroher: Don't you trust Git?

Dilvish-fo: I'm not understanding your comment; I don't think I mistrust Git.

If you are asking about the commented-out line 1055, I simply forgot to remove it once I had line 1054 working right, and I guess I'll go ahead and add another commit here to remove it.

If you are asking why did I put this in as a PR rather than just pushing the commit to master and closing the related issue, it was to help make sure it was in the open long enough to get consideration as to whether this is enough of a bugfix to merit cherrypicking to 0.4.7 and therefor triggering an RC2 (I am slightly leaning towards thinking it is, particularly in light of the changelog edits Mat would also like to make).  It could be that just having the issue being flagged as optional for 0.4.7 was enough to ensure that consideration, but I wasn't sure.  Or perhaps there is some other mechanism it would have been better for me to rely on; pointers welcome.

If your question had some other focus, please clarify.

adrianbroher: The current source code is full of commented out lines, which are not maintained when refactoring and sometimes a reader can't even deduce the intent of the commented out code anymore because  the surrounding code evolved and the commit history is just convoluted or worse: Badly documented (hint @LGM-Doyle: Changed after PR Review #xxxx).

Consider now +2 years and you see this chunk of code:

> Yeah, probably the line was commented out because SelectSystem was broken, but fortunately I fixed this just a week ago.  Lets uncomment this.

Commented out code never adds any value, only confusion in the future.  Removing it in a different commit only adds additional ways to misinterpret the context and add even more confusion.

Dilvish-fo: Wow, that is quite a lecture for what I already said was accidental-- a simple "Yes, I was referring to the commented line" seems it would have been 100% sufficient.   Are you trying to say you deem it to be bad practice to comment out code while working on replacing it?  That the code must be fully deleted **before** the replacement code is ready and working?

Dilvish-fo: and I would not expect the extra commit to make it into master -- I would expect this to PR to be accepted with a squash commit.  Are you saying I should have immediately done a force push with a clean single commit replacing the existing one?

Dilvish-fo:
Hmm, I see that the reference I tried including in my last commit here wound up not referring back to the right commit; I'll be more careful in the future.  If someone else winds up accepting this PR please be sure to do a squash commit and just delete the message line from the second commit.  (I plan on taking care of that myself, but just in case...)

UI/BuildDesignatorWnd.cpp
adrianbroher: > Wow, that is quite a lecture for what I already said was accidental-- a simple "Yes, I was referring to the commented line" seems it would have been 100% sufficient. 

Yes this kind of lecture is unfortunately required, because things like that are getting committed all the time.  If it is because of carelessness, ignorance or malicious intent doesn't matter because the outcome, hard to understand code and programmers intent, is the same.

> Are you trying to say you deem it to be bad practice to comment out code while working on replacing it?  That the code must be fully deleted before the replacement code is ready and working?

Yes.  Try out commiting onto a local branch, rebasing and cleaning up before publishing instead.

> and I would not expect the extra commit to make it into master -- I would expect this to PR to be accepted with a squash commit. Are you saying I should have immediately done a force push with a clean single commit replacing the existing one?

Yes.  Someone will fuck it up in the future and do the wrong thing (like you already have seen in you last comment).  Having rules/processes in place is not a thing to bother people, but to avoid errors that can be avoided by following a process.

adrianbroher: BTW:  It's not a personal attack. I don't trust/lecturing myself the same way all the time. Before I push something I review my commits and I still fuck things up in regular fashion.

LGM-Doyle: @Dilvish-fo the fix works for me.

@adrianbroher, my references to the review in commit messages is intended
- to provide a link to the full discussion of the change.
- to acknowledge and credit the work of the reviewers, who are not committing the code so that it is clear in the commit log who found the bug/suggested the fix and did the work.

While we are using githib the references are obvious.

I realize the github will not last forever.  And when it is gone, if we choose not to archive these discussions are not archived then those references will be dangling.

Dilvish-fo:
ok, I've rebased and will go ahead and merge this to master now, and will check if there is a suitable forum thread for me to ask about cherrypicking this to 0.4.7

Vezzra:
Cherry picking to release branch done.

PR #1489, opened 2017-04-14
Add mention of star type restriction to energy shipyard descriptions.

LGM-Doyle:
There are star restrictions on building both the Energy Compressor and the Solar Containment Unit in the code, but no mention in the descriptions.  

This PR adds a mention to the description.

There are no code changes in this PR, so adding it to a second release candidate would be very low risk.

It does not fix a bug, so adding it to a second release candidate is of low utility.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -11128,6 +11128,8 @@ Energy Compressor
 BLD_SHIPYARD_ENRG_COMP_DESC
 '''This is an upgrade to the [[buildingtype BLD_SHIPYARD_BASE]] and is required for the formation of all energy hulls and all further energy hull shipyard upgrades.
 
+Requiring a high density star to siphon energy from, it can only be built at a system with a [[STAR_BLUE]], [[STAR_WHITE]] or [[STAR_BLACK]] star type.
+
 A highly specialized shipyard designed to create and mount parts on large bundles of densely compressed energy. This is a fragile process, and this shipyard will explode mightily if it takes more than a very small amount of [[encyclopedia DAMAGE_TITLE]].'''
 
 BLD_SHIPYARD_ENRG_SOLAR
@@ -11135,6 +11137,8 @@ Solar Containment Unit
 
 BLD_SHIPYARD_ENRG_SOLAR_DESC
 '''This is an upgrade to the [[buildingtype BLD_SHIPYARD_ENRG_COMP]] and allows formation of [[shiphull SH_SOLAR]]s.
+
+Requiring a very high density star to siphon energy from, it can only be built at a system with a [[STAR_BLACK]] star type.
 
 The task of creating a miniature sun is a gargantuan task in and of itself. The production of such a body on which ship parts can be safely mounted is even more so.'''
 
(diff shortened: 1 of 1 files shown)


Vezzra:
While it may not be a bug in the technical sense, I still consider missing/incomplete/wrong/misleading information in the pedia to be a "bug" documentation-wise, so to speak.

Which means, I see this as a fix, so I think it should be cherry picked to the release branch if it gets merged in time for RC2.

default/stringtables/en.txt
```diff
@@ -11128,6 +11128,8 @@ Energy Compressor
 BLD_SHIPYARD_ENRG_COMP_DESC
 '''This is an upgrade to the [[buildingtype BLD_SHIPYARD_BASE]] and is required for the formation of all energy hulls and all further energy hull shipyard upgrades.
 
+Requiring a high density star to siphon energy from, it can only be built at a system with a [[STAR_BLUE]], [[STAR_WHITE]] or [[STAR_BLACK]] star type.
```
Vezzra: Not a native speaker, so I'm not sure, but wondering if "high density" is the correct term here? Can't think of a better term/wording though...

default/stringtables/en.txt
```diff
@@ -11136,6 +11138,8 @@ Solar Containment Unit
 BLD_SHIPYARD_ENRG_SOLAR_DESC
 '''This is an upgrade to the [[buildingtype BLD_SHIPYARD_ENRG_COMP]] and allows formation of [[shiphull SH_SOLAR]]s.
 
+Requiring a very high density star to siphon energy from, it can only be built at a system with a [[STAR_BLACK]] star type.
```
Vezzra: Ditto.

Vezzra (commented):
Aside from the "high density" thing, looks fine to me.

default/stringtables/en.txt
MatGB: The hulls themselves have "high energy density" as a descriptor, I think high energy stars is better than high density but I'm not an astronomer (and my friend that is does Brown Dwarves so not much help there).

MatGB:
And agreeing with @Vezzra on what counts as a bug: bad, inaccurate or misleading documentation is as much a bug for a Release as other things, whether it's important enough on its own to block a release is always going to be a judgement.

default/stringtables/en.txt
LGM-Doyle: I agree "high density" is not quite right.

We could try "high gravity", "high spatial curvature",  "high space-time curvature" or "high mass" as other perhaps better descriptions of the imaginary technology.

"high energy"  works too.

Or ask your friend, a brown dwarf expert is qualified to write more correct future tech description of stellar mass/energy harvesting.

Vezzra: "High energy stars" sounds good to me to, so seems to be something we can all agree on. Unless @MatGB wants to ask his friend for better suggestions, @LGM-Doyle, can you put this change in so we can merge?

LGM-Doyle: Done and merged.

Vezzra:

PR #1491, opened 2017-04-15
Fix bugs in DesignWnd hulls (un)available scrolling.

LGM-Doyle:
This PR fixes the UI problem described below, which I found while play testing the release candidate.

Steps to replicate starting from the DesignWnd
- display the hulls
- show both available and unavailable hulls.
- scroll to some non-zero offset
- hide both available and unavailable hulls.
- show one type of hull.

Nothing will display.  It should show the hulls that have just been toggled to show.

This PR fixes the problem by starting the vscroll at the correct position when it is added and when it is removed, not adjusting the vscroll position when the first_row_shown is changed and adjusting the offset to the hulls shown if only the last item of a type of hulls would display.

I've marked this as optional for v0.4.7, since the bug is not a crash and clicking away from the hulls list and back again is a simple and obvious workaround.

### DIFF
--- a/GG/src/ListBox.cpp
+++ b/GG/src/ListBox.cpp
@@ -1251,23 +1251,14 @@ void ListBox::BringRowIntoView(iterator target)
 
 void ListBox::SetFirstRowShown(iterator it)
 {
-    if (it == m_rows.end())
+    if (it == m_rows.end() && !m_rows.empty())
         return;
 
     RequirePreRender();
 
+    m_first_row_shown = it;
+
     AdjustScrolls(false);
-
-    if (!m_vscroll && m_first_row_shown == m_rows.begin())
-        return;
-
-    m_first_row_shown = m_vscroll ? it : m_rows.begin();
-
-    Y acc(0);
-    for (iterator it2 = m_rows.begin(); it2 != m_first_row_shown; ++it2)
-        acc += (*it2)->Height();
-    m_vscroll->ScrollTo(Value(acc));
-    SignalScroll(*m_vscroll, true);
 }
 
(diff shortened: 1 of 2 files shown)


Vezzra:
As a fix this should be cherry picked into the release branch if merged in time for RC2.

Dilvish-fo:
It seems to me that the underlying problem is probably more generic to listboxes than just the DesignWnd hull scrolling.   @LGM-Doyle if you have some time perhaps you could take a look at the problem described [at our forums here](http://www.freeorion.org/forum/viewtopic.php?p=88332#p88332) and see if it seems to you to be related and whether your listbox changes here make any improvement for that problem?

LGM-Doyle:
@Dilvish-fo what specific issue from that forum post do you think is related to the ListBox code?

I've posted #1495 to fix the combat log offset, which was *not* due to ListBox changes, but was my breakage.

This PR does fix the initial offset of the ListBox scroll bars, which may have been causing list box scroll bars to not reset to zero when all list box elements were removed elsewhere.  The parts of the code particular to the DesignWnd, prevent the list of hulls from scrolling to show only the last hull after all the hulls are removed and then only one set of hulls is re-added.

Dilvish-fo:
> what specific issue from that forum post do you think is related to the ListBox code?

erm, blank space in something I had imagined might use a listbox?  hehe, I guess I am distracted with some AI issues I am working on and didn't really look at that issue closely enough.

dbenage-cx:
Working well here.
Just to note, have not found any related abnormal behavior in other listboxes.

LGM-Doyle:
Merged and cherry picked to v0.4.7 release branch.

Vezzra:
Thanks!

PR #1495, opened 2017-04-16
Fix bug combat log window layout initial position has incorrect offset.

LGM-Doyle:
The problem as described in the [forum](http://freeorion.org/forum/viewtopic.php?f=28&t=10512&sid=780912247a46d3c866de451b34b61eb9) and other places, is that the combat log is often draw with the incorrect offset from the upper left corner of the combat log tab wnd.

The problem is that the combat log layout is initialized with the absolute position of its parent window and it should be initialized with the relative position.

This PR fixes the problem.

This should be included in 0.4.7. RC2, since it is a small fix easily verified fix.

I marked it as cherry-pick for release so that it would be discussed.  The bug is a minor UI bug that has persisted for a while, so it is not essential for the release.  However, as the forum post indicates, it is an obvious and inconvenient bug for beginners who are more likely to use the combat log regularly.

### DIFF
--- a/UI/CombatReport/CombatLogWnd.cpp
+++ b/UI/CombatReport/CombatLogWnd.cpp
@@ -386,7 +386,7 @@ void CombatLogWnd::Impl::SetLog(int log_id) {
                            GetOptionsDB().Get<bool>("verbose-combat-logging");
 
     m_wnd.DeleteChildren();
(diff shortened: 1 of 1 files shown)


dbenage-cx:
Nice that this was a simple fix, assume this was from the other changes this cycle and does not need changelog amendment.

LGM-Doyle:
@dbenage-cx nope, it looks like I added this bug a year ago this hour d11e2daa96, so it was in the last release.

PR #1497, opened 2017-04-16
Ai fleet mission fix

### LINKED ISSUE #1494: AI prematurely ending system-secure missions
I just discovered that the AI is commonly prematurely ending its system-secure missions, such as when it wants to secure a system in order to invade a planet (and in many other cases as well), I have observed it only having fleet orders to move to the end system, but not having any other fleet orders, so that when it issues the final move order it is thinking all its orders are finished and it is clearing itself.

I suspect that this is an unintended side effect from recent changes that allowed the AI to pass through systems without always stopping, it turns out that [some of the fleet order creation code](https://github.com/freeorion/freeorion/blob/master/default/python/AI/AIFleetMission.py#L518-L527) was apparently dependent on the practice of always stopping at each system, and it appears that the normal final OrderMilitary() was no longer being generated, which in turn was allowing for premature termination of the secure mission as soon as the final move order was issued.

I am working on a fix and hope to submit a PR promptly.

As this is a fairly significant impediment for the AI, I am setting the Milestone as 0.4.7, not optional; requiring an RC2.

Dilvish-fo:
no longer have the AI fleet order generation process require the fleet to arrive at its destination before its final (generally mission critical) order is generated, in order to ensure the mission is not prematurely cleared simply because all of its move orders have been issued.  Fixes #1494

The first three commits fix bugs/weaknesses that were either made apparent or caused by the final commit (which was actually written first); putting them in this order ensures (I believe) that at no point in the sequence of commits does the AI become more buggy rather than less.

@Cjkjvfnby , @Morlic-fo  as highlighted by the fleetorder commit, it turns out that many of the uses of [fleet.get_system()](https://github.com/freeorion/freeorion/blob/master/default/python/AI/universe_object.py#L80) in various flavors of can_issue_order() in fleet_orders.py were using it to try verifying that the fleet is actually at the respective system, not merely that the fleet is on a starlane ending at that system.  It seems that this mismatch had been hidden because of how the fleets were only generating their final order once actually at their final destination.  Without this fleetorder change the errors would now pop up because the fleets will often check about issuing their final order while still on the final travel leg to the destination; for many this may not actually cause problems, but at least for the case of invasion fleets a failure in that final order can trigger an exploration request to get visibility on the system, and the way things are set up it was setting -1 as the target system for emergency exploration, which would actually wind up causing a crash later in AIstate.

I think the most straightforward short term fix for this immediate problem is the one I have done here, to change most of these checks like
```self.fleet.get_system() == self.target```

over to 
```self.fleet.get_object().systemID == self.target.id```

but longer term, to guard against similar usage troubles, I am inclined to think we should rename the current fleet.get_system() to be fleet.get__next_system(), and make a fleet.get_system() that returns just the actual current system, or in the event the fleet is on a starlane then it could return a pseudo-system object with ID -1.  I suppose we'll just need to look carefully at all the code to what will really make the most sense.  So long as you guys agree with the fix in the commits here there is no reason to rush a change to fleet.get_system(), but I'd like you guys to be mulling it over.

### DIFF
--- a/default/python/AI/fleet_orders.py
+++ b/default/python/AI/fleet_orders.py
@@ -304,7 +304,7 @@ class OrderOutpost(AIFleetOrder):
         if ship_id is None:
             ship_id = FleetUtilsAI.get_ship_id_with_role(self.fleet.id, ShipRoleType.BASE_OUTPOST)
         ship = universe.getShip(ship_id)
-        return ship is not None and self.fleet.get_system() == self.target.get_system() and ship.canColonize
+        return ship is not None and self.fleet.get_object().systemID == self.target.get_system().id and ship.canColonize
 
     def issue_order(self):
         if not super(OrderOutpost, self).issue_order():
@@ -369,7 +369,7 @@ class OrderColonize(AIFleetOrder):
         ship = universe.getShip(ship_id)
         if ship and not ship.canColonize:
             print >> sys.stderr, "colonization fleet %d has no colony ship" % self.fleet.id
-        return ship is not None and self.fleet.get_system() == self.target.get_system() and ship.canColonize
+        return ship is not None and self.fleet.get_object().systemID == self.target.get_system().id and ship.canColonize
 
 
 class OrderAttack(AIFleetOrder):
@@ -420,7 +420,7 @@ class OrderInvade(AIFleetOrder):
         planet = self.target.get_object()
         return all((
             ship is not None,
-            self.fleet.get_system().id == planet.systemID,
+            self.fleet.get_object().systemID == planet.systemID,
             ship.canInvade,
             not planet.currentMeterValue(fo.meterType.shield)
         ))
@@ -482,7 +482,7 @@ class OrderMilitary(AIFleetOrder):
         ship_id = FleetUtilsAI.get_ship_id_with_role(self.fleet.id, ShipRoleType.MILITARY)
         universe = fo.getUniverse()
         ship = universe.getShip(ship_id)
-        return ship is not None and self.fleet.get_system() == self.target and (ship.isArmed or ship.hasFighters)
+        return ship is not None and self.fleet.get_object().systemID == self.target.id and (ship.isArmed or ship.hasFighters)
 
     def issue_order(self):
         if not super(OrderMilitary, self).issue_order():
--- a/default/python/AI/fleet_orders.py
+++ b/default/python/AI/fleet_orders.py
@@ -490,7 +490,9 @@ class OrderMilitary(AIFleetOrder):
         target_sys_id = self.target.id
         fleet = self.target.get_object()
         system_status = foAI.foAIstate.systemStatus.get(target_sys_id, {})
-        if fleet and fleet.systemID == target_sys_id and not (system_status.get('fleetThreat', 0) + system_status.get('planetThreat', 0) + system_status.get('monsterThreat', 0)):
+        if all((fleet, fleet.systemID == target_sys_id, system_status.get('currently_visible', False),
+               not (system_status.get('fleetThreat', 0) + system_status.get('planetThreat', 0) +
+                        system_status.get('monsterThreat', 0)))):
             self.order_issued = True
 
 
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -507,24 +507,17 @@ class AIFleetMission(object):
                     self.orders.append(resupply_fleet_order)
             return  # no targets
 
-        # for some targets fleet has to visit systems and therefore fleet visit them
         if self.target:
+            # for some targets fleet has to visit systems and therefore fleet visit them
             system_to_visit = self.target.get_system()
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: default/python/AI/fleet_orders.py [hunk 1]
```diff
@@ -482,15 +482,17 @@ def can_issue_order(self, verbose=False):
...
+        return ship is not None and self.fleet.get_object().systemID == self.target.id and (ship.isArmed or ship.hasFighters)
 
     def issue_order(self):
         if not super(OrderMilitary, self).issue_order():
             return
         target_sys_id = self.target.id
         fleet = self.target.get_object()
         system_status = foAI.foAIstate.systemStatus.get(target_sys_id, {})
-        if fleet and fleet.systemID == target_sys_id and not (system_status.get('fleetThreat', 0) + system_status.get('planetThreat', 0) + system_status.get('monsterThreat', 0)):
+        if all(fleet, fleet.systemID == target_sys_id, system_status.get('currently_visible', False),
```
dbenage-cx: Guessing `all` needs an extra set of parenthesis here?
```
LoggingWrapper.cpp:23 : Chat Message to human: AI_Error: AI script error in "issue_fleet_orders_for_fleet_missions": "all() takes exactly one argument (4 given)"
: : Traceback (most recent call last):
: :   File "/opt/projects/freeorion/default/python/AI/FreeOrionAI.py", line 286, in generateOrders
: :     action()
: :   File "/opt/projects/freeorion/default/python/AI/FleetUtilsAI.py", line 532, in issue_fleet_orders_for_fleet_missions
: :     mission.issue_fleet_orders()
: :   File "/opt/projects/freeorion/default/python/AI/AIFleetMission.py", line 368, in issue_fleet_orders
: :     fleet_order.issue_order()
: :   File "/opt/projects/freeorion/default/python/AI/fleet_orders.py", line 495, in issue_order
: :     system_status.get('monsterThreat', 0))):
: : TypeError: all() takes exactly one argument (4 given)
```

Dilvish-fo: yes, right, that's corrected now

Dilvish-fo:
Anyone who had already grabbed the PR before this comment was posted, please get it again, I made a correction to one of the intermediate commits.

LGM-Doyle:
I play tested this PR and the AI is conquering planets, so it works for me.  I'm not sure what the buggy behavior was, so I can't say if it was fixed.

Dilvish-fo:
@LGM-Doyle The particular buggy behavior that prompted the issue and this PR was that the AI's secure missions would sometimes be prematurely cleared -- the securing fleet would head into the target system and the mission would be cleared during that final leg of the journey without waiting for the fleet to arrive, have combat, and be sure the system was actually secured (and I confirmed this was happening for this reason, not because the AI was deciding some more important defense need merited clearing the preexisting mission).  It might not always be obvious that's what happened, but sometimes the AI military fleet would immediately depart with high health and some not-quite-dead enemy forces still present, which had made me look into what was happening.  I had a test case save game with which I was able to verify the fix worked regarding that.

The fix did have some broader implications/complications, though, leading to the additional 3 commits (the first 3 here), one of which had a bug at first.  So playtesting is quite welcome.  So long as nothing looks wrong and you don't get warning messages, we'll count it as good.  Thanks for your testing.

dbenage-cx:
Do not see the frequent "ping-ponging" of AI fleets with this PR.

Do not think there are any additional warnings, though AI logs are pretty heavy with errors even with master (mostly "fleet *id* can't travel to system S_*id*<*name*>")

Morlic-fo (commented):
I currently do not have the time for a thorough review / testing and will probably not have time until the weekend.

On first glance, this looks ok and reasonable. So given the positive feedback from dbenage-cx and LGM-Doyle, I have no objections.

LGM-Doyle:
My AI logs were also not completely error free.  I don't think that I've ever had an AI log from a long game that was error free.

I saw "fleet X can't travel to system Y errors" and "IssueAggressionOrder : no fleet with passed id" errors.

Dilvish-fo:
> My AI logs were also not completely error free. I don't think that I've ever had an AI log from a long game that was error free.

right, when I said "don't get warning messages" I meant actual warning messages in the game message window, not just the many warnings/errors that make it into the logs. 

> "fleet X can't travel to system Y errors" and "IssueAggressionOrder : no fleet with passed id" errors.

the former probably shouldn't be getting written as errors, or perhaps not even logged at all, or not in the vast majority of cases when it is simply due to the AI inquiring about a travel path that wound up having too many legs out of Supply, or perhaps was blocked by a monster or whatnot.  As it stands now the message is not very helpful/informative, particularly if it is being spammed, I'll try to remember about looking at tightening up the requirements to trigger a warning.  

As for the aggression order, that does sound like an actual error, though it might wind up being an effectively trivial one; perhaps the AI is trying to set aggression on empty fleets or something (which should be about to get cleaned up by some pending PRs).  I'll try to look into that one a bit more also.

PR #1508, opened 2017-04-19
Added Carrier premades, tweaked some names

MatGB:
Added three carrier premades of designs I used regularly in the early game to better showcase the mechanic.

I had meant to do these some time ago and forgot while ill, as we're having another RC these should go in.


Dilvish-fo:
LGTM

Vezzra:

PR #1510, opened 2017-04-19
Fix Fleet::Copy() doesn't clear m_travel_route with partially visible subject.

LGM-Doyle:
If a fleet with a travel route is cloned from a fleet only partially
visible to an empire its `m_travel_route` is not cleared, which may result
in an inconsistent path in the resulting fleet.

This bug interacts with [Universe::UpdateEmpireLatestKnownObjectsAndVisibilityTurns()](https://github.com/freeorion/freeorion/blob/master/universe/Universe.cpp#L2397) which re-uses the existing fleet object to make traveling fleet clones with correct previous and next system ids but an empty path.  If `known_object` already has a path that does not contain the next system id then the whole path is pruned to an empty path.

This bug goes on the cause other bugs, such as fleet buttons not appearing because the fleet is enroute, but has no final destination.  

The fix is simply to clear the `m_travel_route`.

### DIFF
--- a/universe/Fleet.cpp
+++ b/universe/Fleet.cpp
@@ -154,7 +154,10 @@ void Fleet::Copy(std::shared_ptr<const UniverseObject> copied_object, int empire
                 if (!copied_fleet->m_travel_route.empty())
                     m_travel_route.push_back(moving_to);
                 ShortenRouteToEndAtSystem(travel_route, moving_to);
-                if (!travel_route.empty() && travel_route.front() != 0 && travel_route.size() != copied_fleet_route.size()) {
+                if (!travel_route.empty()
+                    && travel_route.front() != INVALID_OBJECT_ID
+                    && travel_route.size() != copied_fleet_route.size())
+                {
                     try {
                         travel_distance -= GetPathfinder()->ShortestPath(travel_route.back(),
                                                                       copied_fleet_route.back()).second;
--- a/universe/Fleet.cpp
+++ b/universe/Fleet.cpp
@@ -150,7 +150,8 @@ void Fleet::Copy(std::shared_ptr<const UniverseObject> copied_object, int empire
 
                 const std::list<int>& copied_fleet_route = copied_fleet->m_travel_route;
 
-                if (m_travel_route.empty() && !copied_fleet->m_travel_route.empty())
+                m_travel_route.clear();
+                if (!copied_fleet->m_travel_route.empty())
                     m_travel_route.push_back(moving_to);
                 ShortenRouteToEndAtSystem(travel_route, moving_to);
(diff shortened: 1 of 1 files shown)


LGM-Doyle:
I'd like this included in the release candidate, so I'm bumping it up to non optional.

It is a 2 line fix.

The problem causes foreign Fleets in clients to have corrupted travel routes.

geoffthemedio:
you appear to be modifying Fleet::Copy not Fleet::Clone

universe/Fleet.cpp
```diff
@@ -150,7 +150,8 @@ void Fleet::Copy(std::shared_ptr<const UniverseObject> copied_object, int empire
 
                 const std::list<int>& copied_fleet_route = copied_fleet->m_travel_route;
 
-                if (m_travel_route.empty() && !copied_fleet->m_travel_route.empty())
+                m_travel_route.clear();
+                if (!copied_fleet->m_travel_route.empty())
                     m_travel_route.push_back(moving_to);
```
geoffthemedio, line 155: given the `this->m_travel_route = travel_route;` below, does any of this do anything?

LGM-Doyle: I changed the references in the title and commit from `Clone()` to `Copy()`.

Yes it does something. `Copy()` works on an existing object, so `m_travel_route` may not be empty, particularly when called from [here](https://github.com/freeorion/freeorion/blob/master/universe/Universe.cpp#L2397).  `ShortenRouteToEndAtSystem()` uses `m_travel_route` to calculate `travel_route`.

I could refactor it to remove the use of `ShortenRouteToEndAtSystem()`, but this is a simpler change that just fixes the bug, the day before the release candidate is made.

universe/Fleet.cpp
```diff
@@ -150,7 +150,8 @@ void Fleet::Copy(std::shared_ptr<const UniverseObject> copied_object, int empire
...
                 ShortenRouteToEndAtSystem(travel_route, moving_to);
                 if (!travel_route.empty() && travel_route.front() != 0 && travel_route.size() != copied_fleet_route.size()) {
```
geoffthemedio: not this PR, but that `!= 0` should probably be `!= INVALID_OBJECT_ID` ...

LGM-Doyle: Fixed and the spacing immediately below, also not on topic.

universe/Fleet.cpp
geoffthemedio: I can't really tell how this is all supposed to work, or be sure that there's not some case this could break, particularly with what's passed into ShortenRouteToEndAtSystem. If you're sure it should work though, I suppose it's OK...

LGM-Doyle: I'll take that as tepid approval.  Merged and cherry picked.

Vezzra: 👍

PR #1514, opened 2017-04-20
allow AI (higher aggressions) to use History Analyser earlier

Dilvish-fo:
cautious through maniacal AIs, moves forward both when they study LRN_PHYS_BRAIN and can build their first history analyser and also when then could build additional ones (if they have conquered another capital)


Dilvish-fo:
I have been testing this a couple days; works fine.

Dilvish-fo:
since this is a tweak, not a bug fix, I am not really sure it is suitable for adding to 0.4.7 at this point, so I'll leave this as optional for 0.4.7 and let @Vezzra  decide

Vezzra:
I'd prefer to keep the cherry picking strictly to bugfixes. (I made an exception for @MatGB's premade carrier designs only because he obviously wanted to get them in earlier and just forgot, and because of the practically non-existent potential for breaking something.)

Dilvish-fo:
Turns out this commit did have a minor semi-bug which could, under some circumstances, trigger an AI warning message ingame on turn 1 (though no actual harm would result).  I've fixed that now in master with  f1de91a

PR #1519, opened 2017-04-22
Config file processing function for the "main" files

memem359:
I found this To-Do comment in the three "main" cpp files, so I made it (refactoring) happen.

`TODO Code combining config, persistent_config and commandline args is copy-pasted
slightly differently in chmain, dmain and camain.  Make it into a single function.`

I tried a single player game, and didn't notice any changes (as expected).

Since I am adding a cpp and header file to the repository, it is possible that I should have changed something else for compilation with Linux or Windows (??).
The Xcode project file is modified, for Mac OSX compilation.

### DIFF
--- a/server/dmain.cpp
+++ b/server/dmain.cpp
@@ -1,16 +1,10 @@
 #include "ServerApp.h"
 
+#include "../client/ConfigFileProcess.h"
 #include "../parse/Parse.h"
-#include "../util/OptionsDB.h"
-#include "../util/Directories.h"
-#include "../util/i18n.h"
 #include "../util/Logger.h"
-#include "../util/Version.h"
-#include "../util/XMLDoc.h"
 
 #include <GG/utf8/checked.h>
-
-#include <boost/filesystem/fstream.hpp>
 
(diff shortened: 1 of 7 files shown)


server/dmain.cpp
```diff
@@ -1,17 +1,11 @@
 #include "ServerApp.h"
 
+#include "../client/ConfigFileProcess.h"
```
geoffthemedio, line 3: doesn't make much sense to include a client file in the server code...

adrianbroher:
Breaks build and includes a merge from master. Closing as invalid.  Please read the `CONTRIBUTING.md`.

memem359:
I'll try a different PR later, but could use some advice to get it right.

adrianbroher - When you say "breaks build", is it the "merge from master", or some an additional issue? (Edit: I think I figured out what I was doing wrong with "rebase". Hopefully I won't keep making that mistake.)

geoffthemedio - Where do you suggest putting it? This is a function common to client/AI/camain, client/human/chmain, and server/dmain.

geoffthemedio:
maybe in util?

PR #1520, opened 2017-04-22
Code for processing config files put into one function, take 2

memem359:
I found this To-Do comment in the three "main" cpp files, so I made it (refactoring) happen.

`TODO Code combining config, persistent_config and commandline args is copy-pasted slightly differently in chmain, dmain and camain. Make it into a single function.`

I tried a single player game, and didn't notice any changes (as expected).

Since I am adding a cpp and header file to the repository, it is possible that I should have changed something else for compilation with Linux or Windows (??).
The Xcode project file is modified, for Mac OSX compilation.

**Repeat of PR #1519 (already closed by the mods), but the new function was moved to the util folder, and the commit errors were straightened out.

### DIFF
--- a/client/human/chmain.cpp
+++ b/client/human/chmain.cpp
@@ -1,24 +1,17 @@
-
+#include "chmain.h"
 
 #include "HumanClientApp.h"
 #include "../../parse/Parse.h"
-#include "../../util/OptionsDB.h"
-#include "../../util/Directories.h"
 #include "../../util/Logger.h"
-#include "../../util/Version.h"
-#include "../../util/XMLDoc.h"
-#include "../../util/i18n.h"
 #include "../../UI/Hotkeys.h"
 
 #include <GG/utf8/checked.h>
 
 #include <boost/format.hpp>
 #include <boost/filesystem/operations.hpp>
-#include <boost/filesystem/fstream.hpp>
 
 #include <thread>
 #include <chrono>
-#include <iostream>
 
 #if defined(FREEORION_LINUX)
 /* Freeorion aims to have exceptions handled and operation continue normally.
@@ -136,42 +129,7 @@ int mainConfigOptionsSetup(const std::vector<std::string>& args) {
         // Add the keyboard shortcuts
         Hotkey::AddOptions(GetOptionsDB());
 
-
-        // TODO Code combining config, persistent_config and commandline args is copy-pasted
-        // slightly differently in chmain, dmain and camain.  Make it into a single function.
-
-        // read config.xml and set options entries from it, if present
-        {
-            XMLDoc doc;
-            try {
-                boost::filesystem::ifstream ifs(GetConfigPath());
-                if (ifs) {
-                    doc.ReadDoc(ifs);
-                    // reject config files from out-of-date version
-                    if (doc.root_node.ContainsChild("version-string") &&
-                        doc.root_node.Child("version-string").Text() == FreeOrionVersionString())
-                    {
-                        GetOptionsDB().SetFromXML(doc);
-                    }
-                }
-            } catch (const std::exception&) {
-                std::cerr << UserString("UNABLE_TO_READ_CONFIG_XML") << std::endl;
-            }
-            try {
-                boost::filesystem::ifstream pifs(GetPersistentConfigPath());
-                if (pifs) {
-                    doc.ReadDoc(pifs);
-                    GetOptionsDB().SetFromXML(doc);
-                }
-            } catch (const std::exception&) {
-                std::cerr << UserString("UNABLE_TO_READ_PERSISTENT_CONFIG_XML")  << ": " 
-                          << GetPersistentConfigPath() << std::endl;
-            }
-        }
-
-
-        // override previously-saved and default options with command line parameters and flags
-        GetOptionsDB().SetFromCommandLine(args);
+        ConfigFileProcess(args);
 
         CompleteXDGMigration();
 
--- a/client/human/chmain.h
+++ b/client/human/chmain.h
@@ -1,8 +1,7 @@
 #ifndef _CHMAIN_H
 #define _CHMAIN_H
 
-#include <string>
-#include <vector>
+#include "../../util/ConfigFileProcess.h"
 
 int mainConfigOptionsSetup(const std::vector<std::string>& args);
 int mainSetupAndRun();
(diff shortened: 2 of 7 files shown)


util/ConfigFileProcess.h
```diff
@@ -0,0 +1,22 @@
+#ifndef _ConfigFileProcess_h_
+#define _ConfigFileProcess_h_
+
+#include <vector>
+#include <string>
+#include <map>
```
adrianbroher, line 6: Almost none of those headers is needed by the `ConfigFileProcess` declaration so don't pollute this header with them.

util/ConfigFileProcess.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include "Directories.h"
+#include "OptionsDB.h"
+#include "i18n.h"
+#include "Version.h"
+#include "XMLDoc.h"
+
+#include <boost/filesystem/fstream.hpp>
+
+// client/human/chmain, server/dmain and client/AI/camain have common code
```
adrianbroher, line 17: Not a docstring; does not contain any useful documentation.

util/ConfigFileProcess.cpp
```diff
@@ -0,0 +1,39 @@
+#include "ConfigFileProcess.h"
+
+//
+// combining config, persistent_config and commandline args
```
adrianbroher, line 4: Don't place documentation at the definition, only to the declaration unless the symbol is limited to a file.

adrianbroher (changes requested):
Doesn't contain the required build system changes for all platforms.

util/ConfigFileProcess.cpp
```diff
@@ -0,0 +1,39 @@
...
+//
+void ConfigFileProcess(const std::vector<std::string> args)
```
adrianbroher, line 6: Also the function name doesn't represent what the function is doing.  `ReadOptions` would be more fitting and also in line with the naming convention.

util/ConfigFileProcess.h
memem359: Okay, I was still operating from the old model where all includes are in the header. But I've read some blogs on the topic, so I understand now what I should be aiming for.

util/ConfigFileProcess.cpp
adrianbroher: You're copying the args parameter instead of passing a reference to it too.

client/human/chmain.h
```diff
@@ -1,8 +1,7 @@
 #ifndef _CHMAIN_H
 #define _CHMAIN_H
 
-#include <string>
-#include <vector>
+#include "../../util/ConfigFileProcess.h"
```
Dilvish-fo, line 4: this include is not actually needed by this file is it?  Seems to me that per the other comments from @adrianbroher  this include should just go directly into chmain.cpp.   And a question for Marcel: it seems that this chmain.h header is not actually needed (and currently does not seem to be used at all in master); the same declarations are currently simply [directly made in chmain.cpp](https://github.com/freeorion/freeorion/blob/master/client/human/chmain.cpp#L54-L55).   Seems to me we should just delete this file rather than adding it as an include to chmain.cpp (or else delete the duplicate declarations in chmain.cpp)

client/human/chmain.cpp
```diff
@@ -1,24 +1,17 @@
-
+#include "chmain.h"
```
Dilvish-fo, line 1: please note [my related comment](https://github.com/freeorion/freeorion/pull/1520/files/e4c3b86ebe35bcdb4d0266f15319199598b3f384#r112816905).

memem359:
> Doesn't contain the required build system changes for all platforms.

EDIT: (Found the Wiki page for adding and removing source-files)

Will adding the new file to util/CMakeLists.txt be sufficient, or is there something else that needs to be done?

client/human/chmain.h
adrianbroher: >  And a question for Marcel: it seems that this chmain.h header is not actually needed (and currently does not seem to be used at all in master)

It is used in `client/human/chmain.mm`.  However I'm not sure how this file interacts with the rest of the chmain files.  But I'm quite sure that OSX doesn't use the main method in chmain.cpp at all.

adrianbroher:
> Will adding the new file to util/CMakeLists.txt be sufficient, or is there something else that needs to be done?

You will need to add the files to the MSVC project too.  And when building the file as part of freeorioncommon the symbol needs to be exported with `FO_COMMON_API`.

util/ConfigFileProcess.h
dbenage-cx: @memem359 As the api docs are generated with doxygen, [Doxygen Manual - Documenting the code](http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html) might be helpful (use `@` style commands)

dbenage-cx:
The name `ConfigFileProcess` does not seem apt, as this is **_processing config settings_** and not concerning a **_config file process_** (e.g. system process).
Possibly confusing without reading the documentation/implementation.

(bold is not strong emphasis here, italic alone made the sentence difficult to read)

LGM-Doyle:
@memem359 the code that you are refactoring is basically
````
     doc.ReadDoc(GetConfigPath());
     GetOptionsDB().SetFromXML(doc);

     doc.ReadDoc(GetPersistentConfigPath());
     GetOptionsDB().SetFromXML(doc);
````
, after removing the error and file handling.

Have you considered moving the error and file handling into `GetOptionsDB()` so that at the top level you can write:
````
     GetOptionsDB().SetFromFile(GetConfigPath(), FreeOrionVersionString()); 
     GetOptionsDB().SetFromFile(GetPersistentConfigPath());
````

`OptionsDB` already has `SetFromXML()` and `SetFromCommandLine()`.  Adding `SetFromFile()` seems a good fit.

The flow in the `freeorion`, `freeoriond` and `freeoriona` is simple: set from `config.xml`, then override from `persistent_config.xml` and finally override from the command line.  

This has the added benefit of not requiring a new file added to the project.

memem359:
@LGM-Doyle That sounds like an elegant solution.
(And not just about avoiding the new file add).
I will close this PR, and try your suggestion.

Thanks for all the comments.

Vezzra:
@memem359, once you've created the new PR (which basically will replace this one), post a reference to it here.

memem359:
The 3rd PR attempt for this (using LGM-Doyle's suggestion, which avoids the need to add a new file) is #1522

PR #1521, opened 2017-04-23
Have the AI combat planning deem planetary defenses to be more…

Dilvish-fo:
…significant when they are bolstered by fleet forces (in addition to the normal ratings combination that is already done).

In the [Fighters forum thread](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10339&start=45) some of the feedback has been about the AI still seeming to have hard time doing battle projections with carriers, and it seems to me at least that is especially so when the AI is attacking a defensive force that includes both planetary defenses and fleets.   Some players seem to consider this to be a bug, but I think it's more accurately just one of the many areas we have with plenty of room for improvement, so I am labeling this a tweak.

It's just a week or two ago that the AI started at least partially considering that fighters will not attack planets, but but this new prong of the attack planning only considers the attack force versus the planet forces, independent of any local enemy fleets (the main prong of attack planning does consider the combined defenses, but using a more classic ratings method that does not consider the fighter attack limitations).  This PR attempts to strengthen that planetary prong of attack planning by up-weighting the planetary defenses when they are bolstered by fleet forces taking into account both the combat rating of those fleet forces and also the number of enemy ships present (which can provide some screening of the planet regardless of their combat rating).

This area deserves much more attention and improvement than what is provided here, but I feel this is at least a little bit of an improvement, and was wanting to try to get something up for review and testing in time for it to be at least considered for 0.4.7.  There is some potential for downside to the approach here, in that if the up-weighting were more severe than is really merited, then the AI would wind up playing over-conservatively.  In my testing so far this seems to be working out fine, but more testing is needed.

@Morlic-fo  @MatGB @LGM-Doyle @dbenage-cx I am particularly hoping that you all might have time both to playtest and also eyeball the calculation in the code and give any suggested alternatives you might have, and also opine on whether this seems like too big of a gamble to try for 0.4.7 or if it seems reasonable to you.  Thanks

*** as a minor note, [the additional bit of system filtering in AIstate](https://github.com/Dilvish-fo/freeorion/blob/aaa2a01fd040966ade424f9e7a3157ce3c3b4caf/default/python/AI/AIstate.py#L150) is to deal with a problem that seems to stem from seeing an enemy fleet enroute to a system which the human client has never observed and so cannot retrieve an UniverseObject for it.  If I entered the enemy_ship_count into into a standalone dictionary instead of into the main AIstate system_status dictionary, then we wouldn't have that trouble, but then I'd also need to tinker with __set_state to preserve backwards compatibility and it also just seemed more appropriate to me to have it in the main system status dictionary.

### DIFF
--- a/default/python/AI/MilitaryAI.py
+++ b/default/python/AI/MilitaryAI.py
@@ -339,6 +339,9 @@ class Allocator(object):
     def _planet_threat(self):
         return get_system_planetary_threat(self.sys_id)
 
+    def _enemy_ship_count(self):
+        return foAI.foAIstate.systemStatus.get(self.sys_id, {}).get('enemy_ship_count', 0.)
+
 
 class CapitalDefenseAllocator(Allocator):
 
@@ -420,6 +423,22 @@ class TargetAllocator(Allocator):
 
     def _take_any(self):
         return self.assigned_rating > 0
+
+    def _planet_threat_multiplier(self):
+        # to the extent that enemy planetary defenses are bolstered by fleet defenses, a smaller portion of our
+        # attacks will land on the planet and hence we need a greater portion of planet-effective attacks.  One
+        # desired characteristic of the following planet_threat_multiplier is that if the entire local threat is
+        # due to planetary threat then we want the multiplier to be unity.  Furthermore, the more enemy ships are
+        # present, the smaller proportion of our attacks would be directed against the enemy planet.  The following is
+        # just one of many forms of calculation that might work reasonably.
+        # TODO: assess and revamp the planet_threat_multiplier calculation
+        return  ((self._enemy_ship_count() + self._local_threat()/self._planet_threat())**0.5
+                 if self._planet_threat() > 0 else 1.0)
+
+    def _allocation_vs_planets(self):
+        return CombatRatingsAI.rating_needed(
+            self.safety_factor*self._planet_threat_multiplier()*self._planet_threat(),
+            self.assigned_rating_vs_planets)
 
 
 class TopTargetAllocator(TargetAllocator):
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -76,7 +76,7 @@ class AIstate(object):
         self.exploredSystemIDs = {}
         self.unexploredSystemIDs = {self.__origin_home_system_id: 1}
         self.fleetStatus = {}  # keys: 'sysID', 'nships', 'rating'
-        # systemStatus keys: 'name', 'neighbors' (sysIDs), '2jump_ring' (sysIDs), '3jump_ring', '4jump_ring'
+        # systemStatus keys: 'name', 'neighbors' (sysIDs), '2jump_ring' (sysIDs), '3jump_ring', '4jump_ring', 'enemy_ship_count'
         # 'fleetThreat', 'planetThreat', 'monsterThreat' (specifically, immobile nonplanet threat), 'totalThreat', 'localEnemyFleetIDs',
         # 'neighborThreat', 'max_neighbor_threat', 'jump2_threat' (up to 2 jumps away), 'jump3_threat', 'jump4_threat', 'regional_threat'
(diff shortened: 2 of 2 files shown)


default/python/AI/MilitaryAI.py
```diff
@@ -421,6 +424,22 @@ def _calculate_threat(self):
...
+        # present, the smaller proportion of our attacks would be directed against the enemy planet.  The following is
+        # just one of many forms of calculation that might work reasonably.
+        # TODO: assess and revamp the planet_threat_multiplier calculation
+        return  ((self._enemy_ship_count() + self._local_threat()/self._planet_threat())**0.5
+                 if self._planet_threat() > 0 else 1.0)
+
+    def _allocation_vs_planets(self):
+        return CombatRatingsAI.rating_needed(
+            self.safety_factor*self._planet_threat_multiplier()*self._planet_threat(),
+            self.assigned_rating_vs_planets)
```
Morlic-fo, line 441: Any reason you are introducing this in this subclass only instead of the ```Allocator``` class?

Dilvish-fo: It was intentional to only do it for the TargetAllocator (and its further subclasses), yes.  

For the two defense type allocators, the most important timing for planning is before the enemy forces actually move into the system, and the current form of calculation I have here just won't help then.  Also the AI will already also have at least one planet of their own there, generally contributing at least something to the battle to counterbalance the other side having a planet present.  It's just a more complicated scenario I wasn't anxious to jump into tinkering with (and didn't seem to get as much attention from players as being noticeably off for the AI). 

However, planning some more flexibility for the future, I suppose this could nearly all go into the base Allocator class except with _planet_threat_multiplier() being a flat unity there, and then this more complicated calc for _planet_threat_multiplier() would override that for the TargetAllocators.  Does that sound better to you?

Morlic-fo: We can always move it once it is actually required/useful. Your reasoning sounds fine. Thinking more about it, we do not even care if we can destroy planetary defenses while defending but we do care that the enemy ships are destroyed.

LGM-Doyle:
I played a game with this PR and it has not broken anything.

I'm not familiar with this part of the code, but it looks like it should work as intended.

The release deadline does not mean much to the developers, work happens before the release and will continue after the release.  

However, to players it is a commitment or promise.  If they notice an upcoming release and return on the day to find it delayed, it is small disappointment.  Once the release is announced, I lean against delays for any non-game breaking problems, hard crashes or commonly occurring unplayable states.

Vezzra:
I'm wary to incorporate something like this into the release branch at this stage. That should be restricted strictly to fixes.

However, considering these changes aren't very complicated and extensive, and close enough to being a fix for a definitely non-optimal behaviour of the AI, I guess I can make an exception, *if* I get some second opinions in favor of it, no objections against, *and* you guys get this merged in time.

I agree with @LGM-Doyle on that point, I don't want to delay RC2 again for something that isn't strictly a fix for a serious issue.

Dilvish-fo:
heads-up, I just pushed a revised version of the commit, which fixed a glitch in the AIstate system filtering (line 150) I mentioned above.

dbenage-cx:
Moderate play testing turned up no issues

Dilvish-fo:
>  if I get some second opinions in favor of it,

bump

LGM-Doyle:
I've played it.  It works.  Get it in before the deadline.

What does the asterisks change in the last commit do?

Vezzra:
Just wanted to remind that it's less than two hours until the deadline, so if you want this to go into RC2 please merge...

Dilvish-fo:
> What does the asterisks change in the last commit do?

the comprehension in that line makes a container holding either zero or one boost python set object sort of thing (I forget what the exact type was) that itself contains the planetIDs; the asterisk unpacks the outer container that so that we actually wind up with a set of planetIDs from the line instead of a set containing a boost python container of planetIDs.  This constructoin relies on the fact that we'll have either zero or one groups of planetIDs.  If the comprehension were producing a container that could have two or more groups of planetIDs then we'd need a different construction here, like set().union(sys.planetIDs for sys in blah if sys)

PR #1524, opened 2017-04-24
Fix travis cpp api generation

dbenage-cx:


Vezzra:
Issues and PRs dealing with stuff that doesn't directly affect builds or docs included in builds don't need to be assigned to milestones.

PR #1527, opened 2017-04-24
reapportions horizontal space in ObjectsWnd FilterDialog params…

Dilvish-fo:
…dropdown to improve legibility.  Depending on what font text size you are using, the second dropdown for conditions in the ObjectsWindow FIlterDialog will either be a little clipped or very clipped, enough so that I consider this a UI bug:

napa_headlight_visa_record1

At the font size shown above (16) none of the conditions need the full space allocated to their dropdown, but as you can see the params dropdown needed for some of the conditions can wind up so clipped that it requires a fair bit of knowledge to know what the entries really are.  A smaller font size does of course help with that, but even going to font 12 does not allow the params to be fully shown, and then of a smaller font size also would mean that the first condition dropdown would need even less space.

This PR makes a modest change to the apportioning of that horizontal space between the two dropdowns; at the shown font size (16) the second one still doesn't have enough space to show everything fully, but it is just barely enough to distinguish the two types of Ancient Ruins special: 
napa_headlight_visa_record1
And, not shown, with this adjusted apportioning even the longer conditions are still not at all clipped in the first dropdown at font 16.

I am currently planning on merging this and cherrypicking in time for the RC2 deadline unless someone objects.

### DIFF
--- a/UI/ObjectListWnd.cpp
+++ b/UI/ObjectListWnd.cpp
@@ -652,7 +652,7 @@ private:
     }
 
     GG::X   DropListWidth() const
-    { return GG::X(ClientUI::Pts()*18); }
+    { return GG::X(ClientUI::Pts()*15); }
 
     GG::X   ParamsDropListWidth() const
     { return CONDITION_WIDGET_WIDTH - DropListWidth(); }


Vezzra:
Yep, I admit this has been bothering me too.

@Dilvish-fo, I'm willing to wait a bit for you to merge and cherry pick this, but please don't make me wait too long, I don't want to go to bed too late today... 😉

Vezzra (approved):
Btw, LGTM.

Dilvish-fo:
OK, done, thanks

Vezzra:
👍

PR #1528, opened 2017-04-24
Add named loggers: preamble.

LGM-Doyle:
This is part of the broken up PR #1481.

These are 4 minor commits that precede the core portion of the named loggers implementation

The purpose of this PR is reduce the extraneous noise in the core PR.

This PR:
+ removes the `fatal` logging level
+ renames `InitLogger` to `InitLoggingSystem`
+ adds two includes which will be removed from `Logger.h` to files which depend on them.

### DIFF
--- a/util/Logger.h
+++ b/util/Logger.h
@@ -8,9 +8,11 @@
 #include "Export.h"
 
 
-/** Initializes the logging system. Log to the given file.
- * If the file already exists it will be deleted. */
-FO_COMMON_API void InitLogger(const std::string& logFile, const std::string& pattern);
+
+/** Initializes the logging system. Log to the given file.  If the file already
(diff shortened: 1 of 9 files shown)


util/Logger.h
```diff
@@ -8,9 +8,11 @@
 #include "Export.h"
 
 
-/** Initializes the logging system. Log to the given file.
- * If the file already exists it will be deleted. */
-FO_COMMON_API void InitLogger(const std::string& logFile, const std::string& pattern);
+
+/** Initializes the logging system. Log to the given file.  If the file already
+ * exists it will be deleted. \p root_logger_name is the name by which the
```
geoffthemedio, line 13: Can you rephrase the explanation of root_logger_name ?

PR #1529, opened 2017-04-24
Add named loggers: core.

LGM-Doyle:
This is part of the broken up PR #1481.

This is the core PR.  It follows the preamble #1528.

***Merge Message***
Adds named loggers with individually controlled log levels.

This PR implements all of the functionality of the named loggers, but does not change any of the currently used loggers, or provide the `OptionsWnd` UI interface to change the thresholds.

PR #1531, opened 2017-04-24
Add named loggers: move version string into app.

LGM-Doyle:
This is part of the broken up PR #1481.

This PR moves responsibility for adding the version string to the log file, out of the logger and into the application code.  The application has a version number, not the logger code.

It follows the core implementation #1529.

### DIFF
--- a/util/Logger.cpp
+++ b/util/Logger.cpp
@@ -1,7 +1,5 @@
 #include "Logger.h"
 
-#include "Version.h"
-
 #include <boost/log/core.hpp>
 #include <boost/log/trivial.hpp>
 #include <boost/log/expressions.hpp>
@@ -323,7 +321,6 @@ void InitLoggingSystem(const std::string& log_file, const std::string& _unnamed_
     // Print setup message.
     auto date_time = std::time(nullptr);
     InfoLogger(log) << "Logger initialized at " << std::ctime(&date_time);
-    InfoLogger() << FreeOrionVersionString();
 }
 
 void ShutdownLoggingSystemFileSink() {
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -12,6 +12,8 @@
 #include "../../network/Message.h"
 #include "../../network/ClientNetworking.h"
 #include "../util/Random.h"
+#include "../util/Version.h"
+
 
 #include "../../universe/System.h"
 #include "../../universe/Species.h"
@@ -95,6 +97,8 @@ AIClientApp::AIClientApp(const std::vector<std::string>& args) :
 
     InitLoggingSystem(AICLIENT_LOG_FILENAME, "AI");
     InitLoggingOptionsDBSystem();
+
+    InfoLogger() << FreeOrionVersionString();
     DebugLogger() << PlayerName() + " ai client initialized.";
 }
 
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -227,6 +227,8 @@ HumanClientApp::HumanClientApp(int width, int height, bool calculate_fps, const 
     RegisterLoggerWithOptionsDB("client", true);
     RegisterLoggerWithOptionsDB("server", true);
 
+    InfoLogger() << FreeOrionVersionString();
+
     try {
         InfoLogger() << "GL Version String: " << GetGLVersionString();
     } catch (...) {
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -147,6 +147,7 @@ ServerApp::ServerApp() :
     InitLoggingSystem(SERVER_LOG_FILENAME, "Server");
     InitLoggingOptionsDBSystem();
 
+    InfoLogger() << FreeOrionVersionString();
     LogDependencyVersions();
 
     m_fsm->initiate();


UI/OptionsWnd.cpp
```diff
@@ -375,6 +376,68 @@ namespace {
             LockColWidths();
         }
     };
+
+
+    /**Create UI controls for a logger level option.*/
+    void LoggerLevelOption(
```
geoffthemedio: if you have such a long list of parameters, you might as well start them on the line with the function name, and break them over two lines, but not going so far right on either line...

UI/OptionsWnd.cpp
```diff
@@ -375,6 +376,68 @@ namespace {
...
+            auto level_name = to_string(static_cast<LogLevel>(ii));
+            auto* priority_row = new CUISimpleDropDownListRow(level_name);
+            // use the row's name to store the option value.
+            priority_row->SetName(level_name);
+            drop_list->Insert(priority_row);
+        }
+
+        // Select the current filtering level in the list
+        auto selected_level = static_cast<std::size_t>(to_LogLevel(GetOptionsDB().GetValueString(option_name)));
+        // Select the current filter level
```
geoffthemedio: redundant comment...

UI/OptionsWnd.cpp
```diff
@@ -375,6 +376,68 @@ namespace {
...
+        if (drop_list->NumRows() >= 1)
+            drop_list->Select(selected_level);
+
+        // Make a layout with a row etc. for this option
+        GG::Layout* layout = new GG::Layout(GG::X0, GG::Y0, GG::X1, GG::Y1, 1, 2, 0, LAYOUT_MARGIN);
+        layout->Add(logger_label, 0, 0);
+        layout->Add(drop_list   , 0, 1, 1, 1, GG::ALIGN_VCENTER);
```
geoffthemedio: extra whitespace after, not before, the ,

UI/OptionsWnd.cpp
```diff
@@ -687,12 +773,21 @@ GG::ListBox* OptionsWnd::CreatePage(const std::string& name) {
     return page;
 }
 
-void OptionsWnd::CreateSectionHeader(GG::ListBox* page, int indentation_level, const std::string& name) {
+void OptionsWnd::CreateSectionHeader(
+    GG::ListBox* page, int indentation_level, const std::string& name, const std::string& tooltip)
```
geoffthemedio: move start of parameters back where they were, and break the line half way through

UI/OptionsWnd.h
```diff
@@ -70,7 +70,8 @@ class OptionsWnd : public CUIWnd {
     void                DoLayout();
 
     GG::ListBox*        CreatePage(const std::string& name);
-    void                CreateSectionHeader(GG::ListBox* page, int indentation_level, const std::string& name);
+    void                CreateSectionHeader(
+        GG::ListBox* page, int indentation_level, const std::string& name, const std::string& tooltip = "");
```
geoffthemedio: ... no

util/Logger.cpp
```diff
@@ -1,81 +1,292 @@
...
-    logging::trivial::severity_level IntToSeverity(int num) {
-        switch (num) {
-        case 5:   return logging::trivial::fatal;   break;
-        case 4:   return logging::trivial::error;   break;
-        case 3:   return logging::trivial::warning; break;
-        case 2:   return logging::trivial::info;    break;
-        case 1:   return logging::trivial::debug;   break;
-        default:  return logging::trivial::trace;
-        }
+    // Create the log logger for logging of logger and logging related events.
```
geoffthemedio: so does this code have something to do with loggers?

util/Logger.cpp
```diff
@@ -1,81 +1,292 @@
...
+    // Compile time constant pointers to constant char arrays.
+    constexpr const char* const log_level_names[] = {"trace", "debug", "info", "warn", "error"};
+}
+
+
+std::string to_string(const LogLevel level)
+{ return log_level_names[static_cast<std::size_t>(level)]; }
+
+LogLevel to_LogLevel(const std::string& text) {
+    if (text == "error")    return LogLevel::error;
```
geoffthemedio: string literals written out in two places seems questionable...

util/Logger.cpp
```diff
@@ -1,81 +1,292 @@
...
-    logging::core::get()->set_filter(logging::trivial::severity >= logging::trivial::trace);
+        // Set a filter to only format this channel
+        sink_frontend->set_filter(log_channel == channel_name);
+
+        // Remove any previous frontend for this channel
+        {
+            boost::mutex::scoped_lock(LoggersMutex());
+
+            // const std::unordered_map<std::string, boost::shared_ptr<TextFileSinkFrontend>>::iterator&
+            const auto&
```
geoffthemedio: odd extra linebreak...

util/Logger.cpp
```diff
@@ -1,81 +1,292 @@
...
+                logging::core::get()->remove_sink(name_and_old_frontend->second);
+                LoggersToSinkFrontEnds().erase(name_and_old_frontend);
+            }
+
+            LoggersToSinkFrontEnds().insert({channel_name, sink_frontend});
+        }
+
+        // Add the frontend.
+        logging::core::get()->add_sink(sink_frontend);
+
```
geoffthemedio: extra blank line?

util/Logger.cpp
```diff
@@ -1,81 +1,292 @@
...
+    // Create the frontend for formatting default records.
+    CreateFileSinkFrontEnd("");
+
+    // Add global attributes to all records
     logging::core::get()->add_global_attribute("TimeStamp", attr::local_clock());
 
-    DebugLogger() << "Logger initialized";
-    InfoLogger() << FreeOrionVersionString();
+    SetLoggerThresholdCore("", default_LogLevel);
```
geoffthemedio: would prefer name `default_log_level`

util/Logger.cpp
```diff
@@ -1,81 +1,292 @@
...
+    // Print setup message.
+    auto date_time = std::time(nullptr);
+    InfoLogger(log) << "Logger initialized at " << std::ctime(&date_time);
 }
 
-void SetLoggerPriority(int priority) {
-    logging::trivial::severity_level i_severity = IntToSeverity(priority);
-    logging::core::get()->set_filter(logging::trivial::severity >= i_severity);
+void OverrideLoggerThresholds(const LogLevel threshold) {
+    InfoLogger(log) << "Forcing all logger threshold to be " << to_string(threshold);
```
geoffthemedio: "all logger" threshold ?
all logger-thresholds ?

util/Logger.h
```diff
@@ -1,56 +1,245 @@
...
+    The intended uses of the levels are:
+    error - used for "major" unrecoverable errors which will affect game play.  Error level issues
+            need to be fixed.  Error level will probably not be turned off unless they are flooding
+            the logs.
+            Examples are: the game is about to crash, a string is missing from the stringtable, etc.
+    warn  - used for "minor", recoverable errors that will not affect game play but do indicate a
+            problem.
+            For example a missing id that can be ignored, an extra item in a container.
+    info  - used to report normal game state and progress.  This should be the default level of
+            logging.  The number of log should be low enought to not floow the logs.
```
geoffthemedio: "log" -> "log entries" ?
enought
floow?

util/Logger.h
```diff
@@ -1,56 +1,245 @@
...
+            For example reporting that the network connected/disconnected.
+    debug - used for low-level implementation or calculation details.  For a named logger this
+            level will probably only be turned on by devs working on that
+            section of code.  This will be detailed and perhaps voluminous.
+            For example reporting that the network disconnected via a client initiated shutdown
+            with a linger time of 30 ms before closing.
+    trace - used for the highest detail of logging.
```
geoffthemedio: example might be helpful here...

util/Logger.h
```diff
@@ -1,56 +1,245 @@
...
+
+    The intended uses of the levels are:
+    error - used for "major" unrecoverable errors which will affect game play.  Error level issues
+            need to be fixed.  Error level will probably not be turned off unless they are flooding
+            the logs.
+            Examples are: the game is about to crash, a string is missing from the stringtable, etc.
+    warn  - used for "minor", recoverable errors that will not affect game play but do indicate a
+            problem.
+            For example a missing id that can be ignored, an extra item in a container.
+    info  - used to report normal game state and progress.  This should be the default level of
```
geoffthemedio: example might be helpful here... perhaps to contrast with debug, "For example reporting that the network disconnected" (ie. without the detail in debug / trace) ?

util/Logger.h
```diff
@@ -1,56 +1,245 @@
...
+    And used with
+
+    ErrorLogger(name_of_logger) << "any streamable output, only computed "
+                                << "at the error threshold or greater "
+                                << "for the 'name_of_logger' logger.";
+    WarnLogger(name_of_logger)  << "streamable output";
+    InfoLogger(name_of_logger)  << "streamable output";
+    DebugLogger(name_of_logger) << "streamable output";
+    TraceLogger(name_of_logger) << "streamable output";
+*/
```
geoffthemedio: not clear how to set a minimum log level for a particular logger... which I thought was a thing?

util/Logger.h
```diff
@@ -1,56 +1,245 @@
...
+
+// The logging levels.
+enum class LogLevel {trace, debug, info, warn, error};
+
+constexpr LogLevel max_LogLevel = LogLevel::error;
```
geoffthemedio: as before, prefer `max_log_level`

util/Logger.h
```diff
@@ -1,56 +1,245 @@
...
+                FO_GLOBAL_LOGGER_NAME_ONE_ARG(__VA_ARGS__))
+
+#endif
+
+/** Initializes the logging system. Log to the given file.  If the file already
+ * exists it will be deleted. \p default_exec_logger_name is the name by which the
+ * default logger "" appears in the log file.*/
+FO_COMMON_API void InitLoggingSystem(const std::string& logFile, const std::string& default_exec_logger_name);
+
+/** Overrides all logger thresholds to \p threshold.*/
```
geoffthemedio: meaning of this comment is unclear... what does overriding a threshold do?

util/LoggerWOptionsDB.cpp
```diff
@@ -0,0 +1,178 @@
+#include "LoggerWOptionsDB.h"
```
geoffthemedio: What does `W` stand for?

util/LoggerWOptionsDB.cpp
```diff
@@ -0,0 +1,178 @@
...
+    constexpr auto source_option_name_prefix = "logging.sources.";
+
+    std::regex exec_name_regex("(?:logging\\.execs\\.)(\\S+)");
+    std::regex source_name_regex("(?:logging\\.sources\\.)(\\S+)");
+
+
+    /** Add the log threshold for a logger with \p full_option_name to OptionsDB and return the
+        threshold read from OptionsDB. */
+    LogLevel AddLoggerToOptionsDB(const std::string& full_option_name) {
+
```
geoffthemedio: extra blank line looks odd

util/LoggerWOptionsDB.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+
+void UpdateLoggerThresholdsFromOptionsDB() {
+    SetLoggerThresholds(LoggerExecutableOptionsLabelsAndLevels());
+    SetLoggerThresholds(LoggerSourceOptionsLabelsAndLevels());
+}
+
+namespace {
+
```
geoffthemedio: blank?

util/LoggerWOptionsDB.h
```diff
@@ -0,0 +1,74 @@
...
+
+        <ai>debug</ai>
+      </sources>
+    </logging>
+
+    The <execs> section controls the default log threshold of each client (freeorion.log), server
+    (freeoriond.log) and AI (AI_x.log) files.
+
+    The <sources> section controls the log threshold of the named loggers.
+
```
geoffthemedio: ...

util/LoggerWOptionsDB.h
```diff
@@ -0,0 +1,74 @@
...
+*/
+
+DiscreteValidator<std::string> LogLevelValidator();
+
+/** Initializes the logging system with settings from OptionsDB and starts capturing
+ * LoggerCreatedSignal to add new loggers to OptionsDB on the fly. */
+FO_COMMON_API void InitLoggingOptionsDBSystem();
+
+// Configure a logger and lookup and/or register the \p name logger in OptionsDB.  Set the initial threshold.
+FO_COMMON_API void RegisterLoggerWithOptionsDB(const std::string logger);
```
geoffthemedio: not by reference?
maybe call it `logger_name` to be clearer?

LGM-Doyle (commented):
@geoffthemedio thanks for reviewing this larg chunk of code.

I've made all of the requested changes.  

I added the following un-requested changes:
- changed the frontend mutex from `boost::scoped_lock` to `std::lock_guard` so that the PR will compile with `clang` and `VC`
- fixed override loggers so that it overrides existing loggers.
- fixed override loggers so that is is non-persistent and doesn't keep overriding loggers on next startup.
- fixed duplicated literals in `LogLevelValidator` which was the same error as in Logger.cpp.

UI/OptionsWnd.cpp
geoffthemedio: In cases where the function name takes up most of the line, sure, but in this case it won't save any space since you use up a full second line and the name is less than that width.

util/LoggerWOptionsDB.cpp
geoffthemedio: not a fan of initialization in general, and this one saves only a few characters...

LGM-Doyle: Renamed to `LoggerWithOptionsDB.<h/cpp>`.

util/Logger.h
```diff
@@ -1,56 +1,292 @@
...
+    InfoLogger()  << "Put any streamable output here.  It will only be
+                  << "computed at the info threshold or greater.";
+    DebugLogger() << "Put any streamable output here.  It will only be
+                  << "computed at the debug threshold or greater.";
+    TraceLogger() << "Put any streamable output here.  It will only be
+                  << "computed at the trace threshold or greater.";
+
+    The named loggers are created with:
+
+    CreateThreadedLogger(name_of_logger);
```
geoffthemedio: Is it necessary to have this function be called before using a named logger? It's not clear where to put it, or if it's a problem to call it twice... or why it's "Threaded". Can't I just use DebugLogger("effects") and have it create that logger if it didn't already exist the first time something is logged to that name?

util/Logger.h
```diff
@@ -1,56 +1,292 @@
...
+    to consume the generated log.
+
+
+    Setting Logger Thresholds:
+
+    Set the logger threshold with:
+
+    SetLoggerThreshold(name_of_logger, a_threshold);
+
+    a_threshold is written LogLevel::error, etc. Use "" for the name of the unnamed source.
```
geoffthemedio: No overload taking just a threadhold and no name that does the equivalent of passing "" as the name?

util/Logger.h
```diff
@@ -1,56 +1,292 @@
...
+    operator. If the "streamable output" is a function call it is only
+    executed if that given source is above its threshold and it has a sink
+    to consume the generated log.
+
+
+    Setting Logger Thresholds:
+
+    Set the logger threshold with:
+
+    SetLoggerThreshold(name_of_logger, a_threshold);
```
geoffthemedio: Not clear when / why I'd want to do this. Aren't there options for that? Can / would I set the threshold repeatedly while running some bit of code?

util/Logger.h
```diff
@@ -1,56 +1,292 @@
...
+
+    a_threshold is written LogLevel::error, etc. Use "" for the name of the unnamed source.
+
+
+    Force the threshold for all loggers to be the same with:
+
+    OverrideAllLoggerThresholds(a_threshold);
+
+    These functions are used to provide further options for setting logger thresholds using
```
geoffthemedio: So most coders shouldn't use them then, and they're really implementation details?

util/LoggerWithOptionsDB.cpp
```diff
@@ -0,0 +1,172 @@
...
+    constexpr auto exec_option_name_prefix = "logging.execs.";
+    constexpr auto source_option_name_prefix = "logging.sources.";
+
+    std::regex exec_name_regex("(?:logging\\.execs\\.)(\\S+)");
+    std::regex source_name_regex("(?:logging\\.sources\\.)(\\S+)");
+
+    /** Add the log threshold for a logger with \p full_option_name to OptionsDB and return the
+        threshold read from OptionsDB. */
+    LogLevel AddLoggerToOptionsDB(const std::string& full_option_name) {
+
```
geoffthemedio: odd blank line

util/LoggerWithOptionsDB.cpp
```diff
@@ -0,0 +1,172 @@
...
+        auto description = UserStringNop("OPTIONS_DB_LOGGER_SOURCE_LEVEL");
+
+        if (is_an_exec_root_logger) {
+            default_level = to_string(default_sink_level);
+            description =  UserStringNop("OPTIONS_DB_LOGGER_FILE_SINK_LEVEL");
+        }
+
+        // Create the new option if necessary.
+        if (!GetOptionsDB().OptionExists(full_option_name))
+            GetOptionsDB().Add<std::string>(
```
geoffthemedio: prefer parameters starting on same line as call

util/LoggerWithOptionsDB.cpp
```diff
@@ -0,0 +1,172 @@
...
+    std::set<std::string> valid_labels;
+    for (const auto& label_and_value: ValidNameToLogLevel())
+        valid_labels.emplace(label_and_value.first);
+    auto validator = DiscreteValidator<std::string>(valid_labels);
+    return validator;
+}
+
+void InitLoggingOptionsDBSystem() {
+    // Initialize the internal logger
+    RegisterLoggerWithOptionsDB("log");
```
geoffthemedio: so is this just intended for logging about the logging system? not clear what "internal" means.

LGM-Doyle (commented):
I think that I've covered all the issues.

dbenage-cx:
(Edit: this comment was intended for original PR (add_detailed_logging), leaving it here to not spam email)

With this branch, I am seeing config entries for 40 AIs from the following options (with default 6 AI game):
https://github.com/freeorion/freeorion/blob/master/client/AI/AIClientApp.cpp#L35

Noticed due to console message `[info]    OptionsDB::SetFromXMLRecursive() : while processing config.xml the following exception was caught when attempting to set option "AI.config.trait.empire-id.all": bad lexical cast: source type value could not be interpreted as target`

Possibly that is intended, but mentioning since with master there are none.
Unintended bug fix? (Or I overlooked some commit)

LGM-Doyle:
@dbenage-cx that looks like an unrelated bug, where the range of the validators for the AI traits did not cover the default values.  This PR forced the defaults to be evaluated because it called `OptionsDB().Commit()` during startup.

I've fixed the problem in `AIClientApp.cpp` in master.

I've also removed the calls to `OptionsDB().Commit()`, while fixing the passing of logger config information.

Thanks.

geoffthemedio:
not sure if rebasing is necessary... seems to be just one additional commit vs. master now?

LGM-Doyle:
rebased.

Issue #1499, opened 2017-04-16
Allied vision does not reveal ship type


Voker57:
They show up as "Ship" with unknown stats, as seen on that screenshot: 

![unseen ships](http://dump.bitcheese.net/files/pumucir/ship.png)

Save (load with Voker57 empire, look at system in screenshot): http://dump.bitcheese.net/files/lihusyz/hidden.mps

version: teamsmac-11.1, based on e45dedee03a215e5a7be8f

Dilvish-fo:
poking around with your save file, I see that the direct ship info actually all comes over fine, but the ship design does not (only the design ID).  I'm submitting a PR that fixes this.

PR #1532, opened 2017-04-25
when getting visibility of ships from allies, also get knowledge of …

Dilvish-fo:
…the ship designs

As described in #1499, currently when getting visibility of ship via an ally, such ships are simply displayed as "ship" without any design info, and their fleets show as having zero ships.

This PR fixes that by ensuring that when getting visibility of ships from allies, the check for giving knowledge of the ship design is also made.  This can be tested with the savegame posted at the above issue, just remember to load it as a multiplayer game and select empire volker57 for yourself, and then advance the game by one turn for visibilities to update.

### DIFF
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1682,14 +1682,8 @@ void Universe::SetEmpireObjectVisibility(int empire_id, int object_id, Visibilit
 
     // if object is a ship, empire also gets knowledge of its design
     if (vis >= VIS_PARTIAL_VISIBILITY) {
-        if (std::shared_ptr<const Ship> ship = GetShip(object_id)) {
-            int design_id = ship->DesignID();
-            if (design_id == ShipDesign::INVALID_DESIGN_ID) {
-                ErrorLogger() << "SetEmpireObjectVisibility got invalid design id for ship with id " << object_id;
-            } else {
-                m_empire_known_ship_design_ids[empire_id].insert(design_id);
-            }
-        }
+        if (std::shared_ptr<const Ship> ship = GetShip(object_id))
+            SetEmpireKnowledgeOfShipDesign(ship->DesignID(), empire_id);
     }
 }
 
@@ -2267,6 +2261,7 @@ namespace {
         // empire id)
         Universe::EmpireObjectVisibilityMap input_eov_copy = empire_object_visibility;
         Universe::EmpireObjectSpecialsMap input_eovs_copy = empire_object_visible_specials;
+        Universe& universe = GetUniverse();
 
         for (std::map<int, Empire*>::value_type& empire_entry : Empires()) {
             int empire_id = empire_entry.first;
@@ -2291,8 +2286,13 @@ namespace {
                     int obj_id = allied_obj_id_vis_pair.first;
                     Visibility allied_vis = allied_obj_id_vis_pair.second;
                     std::map<int, Visibility>::iterator it = obj_vis_map.find(obj_id);
-                    if (it == obj_vis_map.end() || it->second < allied_vis)
+                    if (it == obj_vis_map.end() || it->second < allied_vis) {
                         obj_vis_map[obj_id] = allied_vis;
+                        if (allied_vis < VIS_PARTIAL_VISIBILITY)
+                            continue;
+                        if (std::shared_ptr<const Ship> ship = GetShip(obj_id))
+                            universe.SetEmpireKnowledgeOfShipDesign(ship->DesignID(), empire_id);
+                    }
                 }
             }
         }


universe/Universe.cpp
```diff
@@ -2291,8 +2297,10 @@ namespace {
                     int obj_id = allied_obj_id_vis_pair.first;
                     Visibility allied_vis = allied_obj_id_vis_pair.second;
                     std::map<int, Visibility>::iterator it = obj_vis_map.find(obj_id);
-                    if (it == obj_vis_map.end() || it->second < allied_vis)
+                    if (it == obj_vis_map.end() || it->second < allied_vis) {
                         obj_vis_map[obj_id] = allied_vis;
+			universe.CheckShipDesignVisibility(empire_id, obj_id, allied_vis);
```
geoffthemedio: weird indenting / tab to spaces

universe/Universe.h
```diff
@@ -265,6 +265,9 @@ class FO_COMMON_API Universe {
      * \p object_id' object from the object map and the set of known objects. */
     void            ForgetKnownObject(int empire_id, int object_id);
 
+    /** Gives knowledge of ship designs for sufficiently visible ships */
+    void	    CheckShipDesignVisibility(int empire_id, int object_id, Visibility vis);
```
geoffthemedio: tab to spaces

universe/Universe.cpp
```diff
@@ -1660,6 +1660,21 @@ void Universe::ForgetKnownObject(int empire_id, int object_id) {
     objects.Remove(object_id);
 }
 
+void Universe::CheckShipDesignVisibility(int empire_id, int object_id, Visibility vis) {
```
geoffthemedio: could use a better name, to clarify that it's setting, not just "checking" the visibility...

universe/Universe.cpp
```diff
@@ -1660,6 +1660,21 @@ void Universe::ForgetKnownObject(int empire_id, int object_id) {
...
+    // if object is a ship, empire also gets knowledge of its design
+    if (vis >= VIS_PARTIAL_VISIBILITY) {
```
geoffthemedio: could switch to `if (vis < VIS_PARTIAL_VISIBILITY) return;` and de-indent the following code

universe/Universe.cpp
```diff
@@ -1660,6 +1660,21 @@ void Universe::ForgetKnownObject(int empire_id, int object_id) {
...
+        if (std::shared_ptr<const Ship> ship = GetShip(object_id)) {
```
geoffthemedio: same... switch to `if (!ship) return;` and de-indent

universe/Universe.cpp
Dilvish-fo: so, would you prefer SetShipDesignVisibility?

geoffthemedio: That would be even more misleading.

Maybe MakeDesignVisibileToEmpireIfShipVisible or somesuch?

Dilvish-fo: howabout GrantShipDesignKnowledge

geoffthemedio: basically the same as SetShipDesignVisibility isn't it?

Dilvish-fo: No, I certainly don't think it's the same.  I thought you were objecting to SetShipDesignVisibility because we are not really setting visibility like we do for objects.  This is the function that is responsible for granting knowledge of ship designs, and it seems to me that GrantShipDesignKnowledge is a very good name for it; that is its essence.

In trying to guess why you think my proposed names would be misleading and yours not, I can only see that your function name mentions the requirement of the ship being visible.  Now, if we have one function that handled this for when the ships are visible, and a different function that granted ship design knowledge in other situations, then I could see that the distinction should be part of the name.  But I don't see that in this case.

If the premise is that the function name needs to include all relevant conditions for operation regardless of whether they are needed to distinguish from some other function, then it seems we should really be requiring the name to mention that it is Partial visibility being required not just Basic.  But, mentioning the ship visibility in the name at all, let alone with the "Partial" reference, just seems pointlessly long to me.  

But perhaps I am misunderstanding your reasoning?

geoffthemedio: I don't like a name that suggests it unconditionally gives visibility for a particular ship design (of the specified ship ID) to a particular empire, when it actually only does so if the specified ship is visible above some threshold. It could be MakeDesignVisibileToEmpireIfShipAtLeastPartiallyVisible as well. I don't object to an awkwardly long function name if it clarifies what it does...

Although, looking at it again, the vis parameter isn't really needed... it just acts as a bool test for whether to do anything, which I would rather test outside the function, and give the function itself a simpler behaviour and name.

Dilvish-fo: It's not my understanding that a verb form function name carries a contract or even implication that there can be no constraints on the performance of the action, but I won't press that point.

A name like MakeDesignVisibileToEmpireIfShipAtLeastPartiallyVisible seems to me to put excessive implementation detail in the name.  As for separating the visibility test, that just seems like extra code duplication to me.  

I'll make another proposal: HandleShipDesignKnowledgeGrant

geoffthemedio: It's not a "contract", but I find the name misleading as proposed.

The use of the vis parameter is particularly confusing in terms of its function and how it should be represented in the name, because it doesn't do what I'd expect, which is a test based on the visibility of the object, but rather is an independent parameter that controls whether to do anything at all. That would be clearer if tested at the call location, so that it doesn't appear from inspect of the calling code that something will happen which isn't the case.

I don't have a sense of what HandleShipDesignKnowledgeGrant will do from the name, though that's partly just that it isn't conforming to the vague conventions I'm expecting / imagining.

Dilvish-fo:
@geoffthemedio   OK, I have pushed a replacement commit now which I believe addresses all your requests/suggestions, including taking the visibility check out of the new method, and it's my understanding from the previous discussion that in that context the name GrantShipDesignKnowledge is acceptable.

universe/Universe.h
```diff
@@ -265,6 +265,9 @@ class FO_COMMON_API Universe {
      * \p object_id' object from the object map and the set of known objects. */
     void            ForgetKnownObject(int empire_id, int object_id);
 
+    /** If \a object_id is a ship, gives knowledge of its design to the specified \a empire_id */
+    void	    GrantShipDesignKnowledge(int empire_id, int object_id);
```
geoffthemedio: tab to spaces

universe/Universe.cpp
```diff
@@ -1660,6 +1660,17 @@ void Universe::ForgetKnownObject(int empire_id, int object_id) {
...
 }
 
+void Universe::GrantShipDesignKnowledge(int empire_id, int object_id) {
+    std::shared_ptr<const Ship> ship = GetShip(object_id);
+    if (!ship) return;
+    int design_id = ship->DesignID();
+    if (design_id == ShipDesign::INVALID_DESIGN_ID) {
+        ErrorLogger() << "GrantShipDesignKnowledge got invalid design id for ship with id " << object_id;
+    } else {
+        m_empire_known_ship_design_ids[empire_id].insert(design_id);
```
geoffthemedio: Pondering if it's worth calling `Universe::SetEmpireKnowledgeOfShipDesign` instead of re-implementing some of the checks here... You'd also get some extra detailed error messages instead of silently returning if the ship doesn't exist, and would get a check that `empire_id` isn't `ALL_EMPIRES`, which probably should be checked for.

Also, the name is oddly inconsistent with that function's naming... Maybe instead it could be `SetEmpireKnowledgeOfDesignOfShip` ?

And perhaps change the parameter name to `ship_id` instead of `object_id` ?

Dilvish-fo: > Pondering if it's worth calling Universe::SetEmpireKnowledgeOfShipDesign instead of re-implementing some of the checks here.. You'd also get some extra detailed error messages instead of silently returning if the ship doesn't exist, 

Hmm, well, It looks to me like Universe::SetEmpireKnowledgeOfShipDesign  *does NOT* check existence of the ship (it never even gets the ship_id), and anyways it's not an error in the current case if there is no ship for the object_id; quite often the object would be a fleet or planet or whatnot.  That's why the comment for this method notes "If \a object_id is a ship, ..."

> and would get a check that empire_id isn't ALL_EMPIRES, which probably should be checked for.

if empire_id were ALL_EMPIRES, then in both places that this current method is called from it would have already caused some complications with the object visibility maps; if there were a chance in these constructions for the empire_id to be bad then waiting for Universe::SetEmpireKnowledgeOfShipDesign would be rather late to check it.  

Although relying on Universe::SetEmpireKnowledgeOfShipDesign  could let us save a few lines of code, it would be at a cost -- by doing the INVALID_SHIP_DESIGN check here we are able to give a more informative error message, so I think it's a little better the way it is now.

> Also, the name is oddly inconsistent with that function's naming... Maybe instead it could be SetEmpireKnowledgeOfDesignOfShip ?  And perhaps change the parameter name to ship_id instead of object_id ?

As for the parameter name, since we expect that it often will NOT be a ship, it seems to me better to leave it as object_id.  As for the method name, if the current one strikes you as discordant with SetEmpireKnowledgeOfShipDesign, I myself would prefer to use HandleShipDesignKnowledge.   But if you definitely would prefer SetEmpireKnowledgeOfDesignOfShip then just say so and we'll go with that.

geoffthemedio: OK, if you prefer to duplicate a check for a more localized error message, that's reasonable.

I'd not call it HandleShipDesignKnowledge though, as the "handle" implies a response to an event, or handling all ship design knowledge stuff together, rather than doing it for a specific empire and ship id. So go with SetEmpireKnowledgeOfDesignOfShip please.

Dilvish-fo:
Having let this percolate for a week, I am now thinking that the position I took about checking empire_id was a bit weak-- it's not seeming great to count on details of where the method is currently called from.   So, then it really does seem better to go ahead and rely on Universe::SetEmpireKnowledgeOfShipDesign, and then the code gets a bit simpler also.

The changes around line 1686 are now more really a somewhat extraneous refactoring, not actually at all needed for the purposes of the PR, so, would it be better practice for me to put that in as a separate commit, outside of the PR?

geoffthemedio:
I don't care so much about keeping pull requests on-topic... @adrianbroher might, though.

edit: those comment / close and comment buttons are too close together...

Dilvish-fo:
So, it's my understanding this is fine now, absent further comments I'll be planning to merge it one day fairly soon.

geoffthemedio:
looks fine...

Issue #1537, opened 2017-04-29
Research and Production Queue UI inconsistent


Morlic-fo:
Version: RC2 / 2017-04-24.60d06dc

Production Queue entries are removed by ctrl+left-click while Research Queue entries are removed by double-left-click; ctrl+left-click does not do anything.

It seems that the behaviour should be consistent between these two.

MatGB:
Been meaning to open a similar issue: we need to standardise the UI across different windows, it used to be that double click removed items everywhere but as that was incredibly laggy in some windows they got changed.

Now ctrl+click does it in some, right click/delte or remove does it in some and double click still does it in, IIRC, research.

Needs to be the same across all screens, including the design window.

dbenage-cx:
@MatGB Anything specific that is different on design window?  I only noticed the ship design layout using double click for part removal (only other option is to drag the part away or drag a different part on top of it).

SitReps also use double click to "completely ignore", but only on the icon portion.

I am favouring deletion on Control+LeftClick for easier use on trackpads/touchscreens and preventing accidental double clicks from being troublesome (partially complete queue item).

MatGB:
I'll check through, I think it was that research was different and I knew that ctrl+click had been added to design relatively recently.

If we're consistent I'm happy, if we're consistent based around what we do in production I'm even happier because I like that and agree with you on ctrl click being best.

dbenage-cx:
Any remaining inconsistencies?

Morlic-fo:
Seems consistent now. At the very least, the original point has been adressed.

PR #1539, opened 2017-05-02
Consistent research queue mouse click behaviour

dbenage-cx:
Addresses a few inconsistencies in research queue wrt left mouse clicks.

First two commits refer to #1537, changing to the following behaviour when clicking an item in queue:
* LeftClick: Set pedia article to this tech.  If tech is visible, select and zoom to the entry in tree or list view.
* Control+LeftClick: Remove item from queue
* DoubleLeftClick: Same as LeftClick, but will force visibility of tech category and status.

This is consistent behaviour with the production queue and the design window finished ship designs list.

Third commit addresses an issue with tech status control not changing state when centering on a tech, as done with DoubleLeftClick.

### DIFF
--- a/UI/TechTreeWnd.cpp
+++ b/UI/TechTreeWnd.cpp
@@ -233,6 +233,9 @@ public:
     void Render() override;
 
     void LDrag(const GG::Pt& pt, const GG::Pt& move, GG::Flags<GG::ModKey> mod_keys) override;
+
+    /** Set checked value of control for TechStatus @p status to @p state */
+    void SetTechStatus(TechStatus status, bool state);
     //@}
 
 private:
@@ -470,6 +473,16 @@ void TechTreeWnd::TechTreeControls::LDrag(const GG::Pt& pt, const GG::Pt& move, 
 
         GG::Wnd::LDrag(pt, final_move, mod_keys);
     }
+}
+
+void TechTreeWnd::TechTreeControls::SetTechStatus(TechStatus status, bool state) {
+    auto control_it = m_status_buttons.find(status);
+    if (control_it == m_status_buttons.end()) {
+        WarnLogger() << "No control for status " << status << " found";
+        return;
+    }
+
+    control_it->second->SetCheck(state);
 }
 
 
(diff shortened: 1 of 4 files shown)


UI/TechTreeWnd.cpp
```diff
@@ -233,6 +233,9 @@ class TechTreeWnd::TechTreeControls : public CUIWnd {
     void Render() override;
 
     void LDrag(const GG::Pt& pt, const GG::Pt& move, GG::Flags<GG::ModKey> mod_keys) override;
+
+    /** Set checked value of control for TechStatus @p status to @p state */
+    void SetStatusState(TechStatus status, bool state);
```
geoffthemedio: I'd call this `SetTechStatus` ... adding "state" to the name is confusing

UI/TechTreeWnd.cpp
```diff
@@ -472,6 +475,16 @@ void TechTreeWnd::TechTreeControls::LDrag(const GG::Pt& pt, const GG::Pt& move,
     }
 }
 
+void TechTreeWnd::TechTreeControls::SetStatusState(TechStatus status, bool state) {
+    auto control_it = m_status_buttons.find(status);
+    if (control_it == m_status_buttons.end()){
```
geoffthemedio: space b4 {

geoffthemedio: you don't use `control_it` so could do just `if (m_status_buttons.find(status) == m_status_buttons.end())`

UI/TechTreeWnd.cpp
```diff
@@ -2241,6 +2256,10 @@ void TechTreeWnd::TogglePedia() {
 bool TechTreeWnd::PediaVisible()
 { return m_enc_detail_panel->Visible(); }
 
+bool TechTreeWnd::TechIsVisible(const std::string& tech_name) const {
+    return TechVisible(tech_name, GetCategoriesShown(), GetTechStatusesShown());
```
geoffthemedio: prefer { and } on same line for single-line function definition

UI/TechTreeWnd.cpp
dbenage-cx: Fixed whitespace, `control_it` is used to set the checked state.

dbenage-cx:
TY for review, rebased.

PR #1541, opened 2017-05-03
Rename universe_generation/statistics.py to universe_statistics.py

Morlic-fo:
The previous name is in conflict with the Python 3 standard library:
https://docs.python.org/3/library/statistics.html

While we use use an older Python version at the moment, it still seems cleaner to avoid this potential future conflict.


Dilvish-fo (approved):
sounds & looks good to me

Vezzra (approved):
What @Dilvish-fo said.

Seems to work (did a brief test).

Issue #1246, opened 2017-01-26
Link Error: unresolved external symbol


geoffthemedio:
After making ~UniverseObject public to resolve https://github.com/freeorion/freeorion/issues/1245 I now get link errors:
````
2>       Link:
2>            Creating library ../../FreeOrionCA.lib and object ../../FreeOrionCA.exp
2>     1>
2>AIWrapper.obj : error LNK2001: unresolved external symbol "class Universe const volatile * __cdecl boost::get_pointer<class Universe const volatile >(class Universe const volatile *)" (??$get_pointer@$$CDVUniverse@@@boost@@YAPDVUniverse@@PDV1@@Z)
2>     1>
2>UniverseWrapper.obj : error LNK2001: unresolved external symbol "class UniverseObject const volatile * __cdecl boost::get_pointer<class UniverseObject const volatile >(class UniverseObject const volatile *)" (??$get_pointer@$$CDVUniverseObject@@@boost@@YAPDVUniverseObject@@PDV1@@Z)
2>     1>
2>UniverseWrapper.obj : error LNK2001: unresolved external symbol "class Fleet const volatile * __cdecl boost::get_pointer<class Fleet const volatile >(class Fleet const volatile *)" (??$get_pointer@$$CDVFleet@@@boost@@YAPDVFleet@@PDV1@@Z)
2>     1>
2>UniverseWrapper.obj : error LNK2001: unresolved external symbol "class Ship const volatile * __cdecl boost::get_pointer<class Ship const volatile >(class Ship const volatile *)" (??$get_pointer@$$CDVShip@@@boost@@YAPDVShip@@PDV1@@Z)
2>     1>
2>UniverseWrapper.obj : error LNK2001: unresolved external symbol "class Planet const volatile * __cdecl boost::get_pointer<class Planet const volatile >(class Planet const volatile *)" (??$get_pointer@$$CDVPlanet@@@boost@@YAPDVPlanet@@PDV1@@Z)
2>     1>
2>UniverseWrapper.obj : error LNK2001: unresolved external symbol "class System const volatile * __cdecl boost::get_pointer<class System const volatile >(class System const volatile *)" (??$get_pointer@$$CDVSystem@@@boost@@YAPDVSystem@@PDV1@@Z)
2>     1>
[...]

PR #1542, opened 2017-05-05
Fix MSVC link error

wheals:
Before this, I consistently ran into the following error:
```
4>UniverseWrapper.obj : error LNK2005: "class Universe const volatile * __cdecl boost::get_pointer<class Universe const volatile >(class Universe const volatile *)" (??$get_pointer@$$CDVUniverse@@@boost@@YAPDVUniverse@@PDV1@@Z) already defined in AIWrapper.obj
4>../../FreeOrionCA.exe : fatal error LNK1169: one or more multiply defined symbols found
```
MSVC version: Microsoft Visual Studio Community 2015 14.0.24720.00 Update 1

Someone else who uses MSVC should probably test this. For what it's worth, [this comment](https://github.com/freeorion/freeorion/issues/1246#issuecomment-275485658) suggests the functions are no longer necessary with the removal of TemporaryPtr.

I also added an update to my preferred crediting since this is my first PR in a while, didn't want to cause the bother of a separate PR.

### DIFF
--- a/default/credits.xml
+++ b/default/credits.xml
@@ -59,7 +59,7 @@
       <PERSON name="Vincent Legoll"           nick="vincele"          task="Programming"/>
       <PERSON name="Fabian Wendt"             nick="Morlic"           task="Programming"/>
       <PERSON name=""                         nick="mel-odious"       task="Programming"/>
-      <PERSON name="Mark Shmuale"             nick="wheals"           task="Programming"/>
+      <PERSON name="S. Mark"                  nick="wheals"           task="Programming"/>
       <PERSON name="Andrey Solomatin"         nick="Cjkjvfnby"        task="Programming"/>
       <PERSON name=""                         nick="LGM-Doyle"        task="Programming"/>
       <PERSON name="D. Benage"                nick="dbenage-cx"       task="Programming"/>
--- a/python/UniverseWrapper.cpp
+++ b/python/UniverseWrapper.cpp
@@ -39,8 +39,6 @@ namespace boost {
     const volatile Field*           get_pointer(const volatile Field* p) { return p; }
     template<>
     const volatile Building*        get_pointer(const volatile Building* p) { return p; }
-    template<>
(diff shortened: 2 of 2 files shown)


wheals:
Judging by the AppVeyor failure, seems at least some versions of MSVC need the definitions after all... :( No idea how to detect whether they're necessary, though.

geoffthemedio:
With this patch applied, I get a bunch of link errors, similar to the AppVeyor results.

wheals:
Seems that the function definitions are necessary [specifically for MSVC2015 Update 3](http://stackoverflow.com/questions/38261530/unresolved-external-symbols-since-visual-studio-2015-update-3-boost-python-link). I'll update and make sure that that fixes the link error without needing to change the code, then close the PR.

wheals:
OK, I upgraded to Update 3 (14.0.25431.01), and that fixed *most* of the errors, but one of them still remained. I'll try pushing the fix for it now, and see if it also works on other machines.

If not, maybe changing the `#if` to use [_MSC_FULL_VER](https://msdn.microsoft.com/en-us/library/b0084kay.aspx) will help.

adrianbroher:
> OK, I upgraded to Update 3 (14.0.25431.01), and that fixed most of the errors, but one of them still remained.

Have you tried ~~turning it off and on again~~ to do a clean build?  It's maybe possible that the template code already is instantiated somewhere in an object file causing the multiple symbol definition.

wheals:
I did do a clean build after upgrading, but I can try cleaning the FreeOrionCA and FreeOrionD projects again and trying yet another time (hesitant to clean out everything since my computer tends to choke while compiling the parser -- it actually crashed when I did the clean build earlier, and that might have messed it up).

geoffthemedio:
You can reduce the number of parallel builds in the options somewhere. Might help get the parser to build.

wheals:
Seems a clean and a rebuild fixed it. Sorry for the trouble.

PR #1543, opened 2017-05-05
Fix a crash on startup during log initalization

wheals:
For some reason, when [`AddLoggerToOptionsDB`](https://github.com/freeorion/freeorion/blob/master/util/LoggerWithOptionsDB.cpp#L33) gets called during initialization of the logger, `log_level_names` is full of garbage data. Removing the `constexpr` as here fixes it.

I'm not sure if this is MSVC screwing up and not following the C++ standard, or the other compilers/versions of MSVC being too lax.

@LGM-Doyle You're working on the logger, so I figure you might know more about this?

### DIFF
--- a/util/Logger.cpp
+++ b/util/Logger.cpp
@@ -47,8 +47,8 @@ namespace {
     }
 
 
-    // Compile time constant pointers to constant char arrays.
-    constexpr const char* const log_level_names[] = {"trace", "debug", "info", "warn", "error"};
+    // Constant pointers to constant char arrays.
+    const char* const log_level_names[] = {"trace", "debug", "info", "warn", "error"};
 }
 
 


LGM-Doyle:
@wheals thanks for reporting and providing a patch.

Some checking indicates that MSVC 2015 does not support `constexpr` in a variety of ways, specifically [stock MSVC 2015](https://blogs.msdn.microsoft.com/vcblog/2015/04/29/c11-constant-expressions-in-visual-studio-2015-rc/) did not support array initialization until [update 1](https://blogs.msdn.microsoft.com/vcblog/2015/12/02/constexpr-in-vs2015-update-1/).  There were further `constexpr` fixes in [update 2](https://blogs.msdn.microsoft.com/vcblog/2016/02/11/compiler-improvements-in-vs-2015-update-2/)

What version of MSVC are you using?
If you are not using the latest MSVC 2015 update (which I think is 3 but I am on linux) and you upgrade does that fix the problem?

@geoffthemedio, @adrianbroher when we say that the freeorion SDK requires MSVC 2015 do we mean a fully patched MSVC2015 or stock?

wheals:
I'm on update 1. Seems that was the cause of my problem in #1542 too; guess I'll just upgrade my version of MSVC and check that both of the issues are fixed.

geoffthemedio:
I'm not aware of specific update requirements for building, but if there are some, then probably it should be noted as such.

wheals:
I can confirm that the crash does not happen in Update 3.

wheals:
Upgrading to Update 3 does fix the crash. Should I submit a PR to the SDK repository clarifying that Update 3 is necessary? I can edit the corresponding wiki page myself, if necessary.

adrianbroher:
@LGM-Doyle 
> when we say that the freeorion SDK requires MSVC 2015 do we mean a fully patched MSVC2015 or stock?

I never gave this a second thought.  Both the SDK and FreeOrion itself were always build with whatever was part of the AppVeyor `Visual Studio 2015` image provided. As far as I can trace this in the AppVeyor documentation history we used `VS 2015 Update 3` since we did CI.

@wheals 
> Should I submit a PR to the SDK repository clarifying that Update 3 is necessary?

Feel free to do so.

PR #1544, opened 2017-05-05
Use unique_ptr for BuildingType parser

LGM-Doyle:
**Problem**
The BuildingTypeParser is a generator.  In C++11 is should return a `std::unique_ptr<>` to indicate that it has created something that requires memory management and that it is passing management responsibility to the calling function.  The calling function can chose to receive it as a `unique_ptr<>` or transfer it to a `shared_ptr<>` if it needs a shared ownership model.

**Solution**
Return `std::unique_ptr<>`s.

### DIFF
--- a/universe/Building.h
+++ b/universe/Building.h
@@ -194,7 +194,7 @@ private:
 /** Holds all FreeOrion building types.  Types may be looked up by name. */
 class BuildingTypeManager {
 public:
-    typedef std::map<std::string, BuildingType*>::const_iterator iterator;
+    typedef std::map<std::string, std::unique_ptr<BuildingType>>::const_iterator iterator;
 
     /** \name Accessors */ //@{
     /** returns the building type with the name \a name; you should use the
@@ -214,9 +214,8 @@ public:
 
 private:
     BuildingTypeManager();
-    ~BuildingTypeManager();
 
-    std::map<std::string, BuildingType*> m_building_types;
+    std::map<std::string, std::unique_ptr<BuildingType>> m_building_types;
 
     static BuildingTypeManager* s_instance;
 };
--- a/UI/BuildDesignatorWnd.cpp
+++ b/UI/BuildDesignatorWnd.cpp
@@ -788,7 +788,7 @@ void BuildDesignatorWnd::BuildSelector::PopulateList() {
         // craete and insert rows...
         std::vector<GG::ListBox::Row*> rows;
         rows.reserve(std::distance(manager.begin(), manager.end()));
-        for (const std::map<std::string, BuildingType*>::value_type& entry : manager) {
+        for (const auto& entry : manager) {
             const std::string name = entry.first;
             if (!BuildableItemVisible(BT_BUILDING, name))
                 continue;
--- a/UI/ChatWnd.cpp
+++ b/UI/ChatWnd.cpp
@@ -190,7 +190,7 @@ void MessageWndEdit::FindGameWords() {
             m_game_words.insert(UserString(tech_name));
     }
     // add building type names
-    for (const std::map<std::string, BuildingType*>::value_type& entry : GetBuildingTypeManager()) {
+    for (const auto& entry : GetBuildingTypeManager()) {
         if (entry.second->Name() != "")
             m_game_words.insert(UserString(entry.second->Name()));
     }
--- a/UI/ClientUI.cpp
+++ b/UI/ClientUI.cpp
@@ -795,7 +795,7 @@ bool ClientUI::ZoomToContent(const std::string& name, bool reverse_lookup/* = fa
                 return ZoomToTech(tech->Name());
         }
 
-        for (const std::map<std::string, BuildingType*>::value_type& entry : GetBuildingTypeManager())
+        for (const auto& entry : GetBuildingTypeManager())
             if (boost::iequals(name, UserString(entry.first)))
                 return ZoomToBuildingType(entry.first);
 
--- a/UI/EncyclopediaDetailPanel.cpp
+++ b/UI/EncyclopediaDetailPanel.cpp
@@ -242,7 +242,7 @@ namespace {
             }
 
         } else if (dir_name == "ENC_BUILDING_TYPE") {
-            for (const std::map<std::string, BuildingType*>::value_type& entry : GetBuildingTypeManager()) {
+            for (const auto& entry : GetBuildingTypeManager()) {
                 std::string custom_category = DetermineCustomCategory(entry.second->Tags());
                 if (custom_category.empty()) {
                     sorted_entries_list.insert({UserString(entry.first),
@@ -470,7 +470,7 @@ namespace {
                     dir_entries[UserString(tech_name)] = std::make_pair(VarText::TECH_TAG, tech_name);
 
             // building types
-            for (const std::map<std::string, BuildingType*>::value_type& entry : GetBuildingTypeManager())
+            for (const auto& entry : GetBuildingTypeManager())
                 if (DetermineCustomCategory(entry.second->Tags()) == dir_name)
                     dir_entries[UserString(entry.first)] = std::make_pair(VarText::BUILDING_TYPE_TAG, entry.first);
 
--- a/parse/BuildingsParser.cpp
+++ b/parse/BuildingsParser.cpp
@@ -15,15 +15,26 @@
 #if DEBUG_PARSERS
 namespace std {
     inline ostream& operator<<(ostream& os, const std::vector<std::shared_ptr<Effect::EffectsGroup>>&) { return os; }
-    inline ostream& operator<<(ostream& os, const std::map<std::string, BuildingType*>&) { return os; }
-    inline ostream& operator<<(ostream& os, const std::pair<const std::string, BuildingType*>&) { return os; }
+    inline ostream& operator<<(ostream& os, const std::map<std::string, std::unique_ptr<BuildingType>>&) { return os; }
+    inline ostream& operator<<(ostream& os, const std::pair<const std::string, std::unique_ptr<BuildingType>>&) { return os; }
 }
 #endif
 
 namespace {
     const boost::phoenix::function<parse::detail::is_unique> is_unique_;
 
-    const boost::phoenix::function<parse::detail::insert> insert_;
+    struct insert_BuildingType {
+        typedef void result_type;
+
+        void operator()(std::map<std::string, std::unique_ptr<BuildingType>>& building_types,
+                        const std::string& name,
+                        BuildingType* building_type) const
+        {
+            std::unique_ptr<BuildingType> building_typep(building_type);
+            building_types.insert(std::make_pair(name, std::move(building_typep)));
+        }
+    };
+    const boost::phoenix::function<insert_BuildingType> insert_;
 
     struct rules {
         rules() {
@@ -73,7 +84,7 @@ namespace {
         }
 
         typedef parse::detail::rule<
-            void (std::map<std::string, BuildingType*>&),
+            void (std::map<std::string, std::unique_ptr<BuildingType>>&),
             boost::spirit::qi::locals<
                 std::string,
                 std::string,
@@ -83,7 +94,7 @@ namespace {
         > building_type_rule;
 
         typedef parse::detail::rule<
-            void (std::map<std::string, BuildingType*>&)
+            void (std::map<std::string, std::unique_ptr<BuildingType>>&)
         > start_rule;
 
         building_type_rule          building_type;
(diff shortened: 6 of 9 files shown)


Vezzra:
The "component:content scripting" label is for issues/PRs which deal with the actual FOCS/Python scripts, not the parser/interface C++ code. Changed the label to "component:internal". Will do that for all the issues/PRs of this group which have already been created (#1544 - #1551).

LGM-Doyle:
Thanks.

I think I've made that mistake twice now.

universe/Building.h
```diff
@@ -214,9 +214,8 @@ class BuildingTypeManager {
 
 private:
     BuildingTypeManager();
-    ~BuildingTypeManager();
 
-    std::map<std::string, BuildingType*> m_building_types;
+    std::map<std::string, std::unique_ptr<BuildingType>> m_building_types;
```
geoffthemedio, line 218: This doesn't make sense to me. Shouldn't it be stored in the manager and returned from GetBuildingType as a shared_ptr? Storing as unique and then returning a raw pointer seems like a bad idea.

Returning from a parser as unique_ptr also seems pointless... why not just return a shared_ptr in the first place? It accomplishes the same thing (no leaks) without the awkwardness of having to convert from unique to shared for storage.

LGM-Doyle: Returning `unique_ptr` from the parser is correct.  It is explicit in the function signature that the parser does not retain any stake in the parsed object.  Returning `shared_ptr` implies otherwise.

This PR is just focused on fixing the parser.

The `BuildingTypeManager` never passes ownership of the building types to any other code.  It is statically allocated and never destroyed so its lifetime exceeds all of its calling functions.  It is correct for the building manager to return raw pointers.

Converting a received `unique_ptr` from a generator into a `shared_ptr` is as simple as constructing from the `unique_ptr`:
````
unique_ptr<T> generatorT() {
    return make_unique<T>();
}

main() {
    shared_ptr<T> receive_T = generatorT();
}
````  
It is one line of code, not more lines of code than not doing the conversion. It increases code clarity because the transfer of ownership is explicit. 

Here is further exposition to clarify what I think are common conventions for smart pointers in C++:
- `unique_ptr<T>` - I am the sole owner of this `T`.  If I move this `unique_ptr<T>` to you then you become the sole owner of `T`.  There is no overhead (relative to a raw pointer) to dereferencing, moving, copying or destroying this `unique_ptr<T>`.  There is possibly a memory allocation overhead to storing a custom destructor for this `T`
- `shared_ptr<T>` - I share ownership of this `T`. If I move/copy this `shared_ptr<T>` then you also share ownership. There is no overhead to dereferencing or moving this `shared_<T>`.  Copying or destroying this `shared_ptr<T>` requires an extra memory dereference to the control block and an atomic increment/decrement.  Constructing this `shared_ptr<T>` requires allocating additional memory for the control block. If constructed with `make_shared` then the control block may be contiguous with the `T`.
- `weak_ptr<T>`  - I can observe a `shared_ptr<T>`, but I do not own it nor can I prevent it from being destroyed.  "Dereferencing", constructing, copying or destroying `weak_ptr<T>` all have overhead. 
- raw `T*`  - Some other entity is managing the memory for this object.  I know, from program semantics, that the lifetime of this objects exceeds my use.  There is no overhead to any operations.

geoffthemedio: To me returning...

`raw T*` indicates nothing about ownership. Returning one indicates that smart pointers aren't be used, and careful consideration of whether and when to delete is required.

`shared_ptr` indicates that the pointed-to object can be used as needed and the pointer can be allowed to go out of scope safely, and can always be assumed to point to a valid object if it did so when initially acquired, and that there may be multiple pointers to the object in existence simultaneously.

`weak_ptr` indicates that the pointed-to object may stop existing after the pointer is acquired, so needs to be converted to a `shared_ptr` before use.

`unique_ptr` indicates that the calling code is receiving the only valid pointer and needs to save it if the object is to continue existing / remain valid.

Since for all the things like `BuildingType` in the manager classes, there can be multiple pointers to it at the same time, they should be held in `shared_ptr`. Using a `unique_ptr` appears very wrong, because it will always need to be possible to have at least two pointers to the object exist simultaneously: within the manager, and within code that asks the manager for a pointer to the object by name or id. (Holding it in a `unique_ptr` and then returning a raw pointer seems even worse, though is probably equivalent by necessity.)

So it's fine to return a `unique_ptr` from a generator class, but if the pointed-to object is meant to continue existing within a manager class and have multiple pointers to it existing at the same time, it should be immediately assigned to a `shared_ptr` to be kept within the manager.

Dilvish-fo: > To me returning... ```raw T*``` indicates nothing about ownership. 

I think the convention makes sense within (and only within) the 'no naked new' framework-- ***once you are confident your code employs no naked 'new's,*** and all heap allocations are required to be owned either via a unique_ptr or a shared_ptr, then by implication a raw pointer can never be used for ownership, and for improved efficiency you could allow this extra convention.  Here is another statement of what I think is essentially the same convention (or at least is an overlapping one): [Take smart pointers as parameters only to explicitly express lifetime semantics...A function that does not manipulate lifetime should take raw pointers or references instead.](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rr-smartptrparam)

Now, a quick grep of FreeOrion appears to show a great many naked 'new's still in our code, so perhaps it is premature to try jumping forward to this use of raw pointers.   Also, I still remember BigJoe's work going through replacing our uses of raw pointers with shared_ptr, and the idea of intentionally returning to using raw pointers in any situations evokes a bit of cognitive dissonance.

But overall this does seem like the right direction for us to be heading.

LGM-Doyle: @geoffthemedio thanks for writing back.  
@Dilvish-fo thanks for the link.  I will definitely read through that.  Isn't appeal to authority one of those logical fallacy I was warned about 😉 

I should have written, conventions for smart pointer use in C++11.  We are still making the transition to C++11 and wrt to pointer use there are two areas that are still using pre-C++11 conventions: the parser code and GG.  Those two areas of code are still using `new` and `delete` to manage the lifetime of heap based objects. 

Once those are updated then a raw `T*` will explicitly indicate non-ownership of a heap object.  Code can not safely store a raw pointer or allow control to return to the owner and that is also implied by the raw pointer type.  If code does not store raw pointers, then there will not be multiple pointers to the managed objects existing at the same time.

This PR is not about fixing `BuildingTypeManager`.

However, `BuildingTypeManager` is safe and correct to return raw pointer on two accounts.

First, it is static, global, parses the building types once and never de-allocates them.

Second, I checked ~52 uses of its returned pointers and they are correctly used and never stored.

There are two cases where `m_enqueue_location` and `m_location` are stored, but that is an argument for storing/passing those as `shared_ptrs` not the whole `BuildingType`.

geoffthemedio: > First, it is static, global, parses the building types once and never de-allocates them.
>
> Second, I checked ~52 uses of its returned pointers and they are correctly used and never stored.

I'd rather not build in assumptions that the contents of manager classes will stay valid. Particularly due to the potential for multirhreading in network messaging between the server and clients, and proposals to update the content when switching content directories from default and back. Even within a function's local scope, a raw pointer could be invalidated in another thread. Making the pointer point to a const type seems better to specify that the returned shared pointer shouldn't be used to modify the object.

LGM-Doyle: Rebased.

`shared_ptr` and `weak_ptr` are as thread safe as a raw pointer.  

Constructing, copying or destroying `shared_ptr` or `weak_ptr` guarantees that adjusting the counters are atomic operations.

Dereferencing and using the pointed to object is not thread-safe.

The pointers, `shared_ptr` and `weak_ptr` are thread-safe, but accessing the pointed to object without some other thread safety consideration is not.

Being passed any kind of pointer does not imply or enforce any thread safety.  If we want/expect objects to be thread safe, then they require additional protection beyond `shared_ptr`.

`const` is the correct way to indicate that an object should not be modified. This does not control or indicate anything about whether an object can/should be deleted.  Additionally, accessing a `const` object is not thread safe. 

The ownership that the type of pointer indicates is of resource management, in this case memory.  Here is another table:

Pointer type | modifiability | ownership (can I delete it)
---|---|---
 `unque_ptr<T>`              | I can        modify this `T`   |   I own/manage its resources (memory).
 `unque_ptr<const T>`    | I can  not modify this `T`   |   I own/manage its resources (memory).
 `shared_ptr<T>`             | I can        modify this `T`   |   I share ownership/management.
 `shared_ptr<const T>`   | I can not  modify this `T`   |   I share ownership/management.
 `T*`                                  | I can        modify this `T`   |   I do not own it.
 `const T*`                        | I can not  modify this `T`   |   I do not own it.

Regarding `BuildingManager` the pointed to building types are not thread safe and using `shared_ptr` will not make them thread safe.

geoffthemedio: Perhaps I don't understand how they work, but by thread safety, I meant that if some function's local scope includes a (valid) `shared_ptr` to an object, then that object can be assumed to exist as long as the `shared_ptr` doesn't go out of scope or get reset within that function. If the manager classes store their pointers as `shared_ptr`, and something in one thread causes them to reset all those pointers, any other copies of the `shared_ptr` in use elsewhere should keep the object from being deallocated until all copies are reset. A raw pointer wouldn't keep the object from being deallocated when the manager class resets all its pointers.

This isn't about making accessing objects thread safe, but just keeping the pointers to objects valid as long as there exists such a pointer in some scope where it could be dereferenced.

LGM-Doyle: Yes it won't be deleted, but it also can't be safely dereferenced in either thread.  So, it is effectively leaked.  It is as good as deleted, but a lot harder to debug because a dereference could cause a segfault, bad data or no problem at all.  These types of crashes will be more likely to happen in large games than small test cases, because more objects are being accessed.  They will also be architecture specific.

Passing any kind of pointer to another thread must involve consideration of managing access to the dereferenced object.  Using `shared_ptr` is not a way around that work.

Our default instinct should be to use `unique_ptr` for memory management.

If we default to `shared_ptr`, without consideration of why in each case, then we will exchange "technical" resource leaks for "effective" resource leaks.

In this case, we have single resource manager, called `BuildingManager`.  None of the clients store the pointers.  None of the clients are thread safe.   None of the clients' lifetimes exceed the manager, or any of the managed objects.  The pointers should be `unique_ptrs`.

geoffthemedio: > Yes it won't be deleted, but it also can't be safely dereferenced in either thread.

The `shared_ptr` can never be safely dereferenced, or the member functions of the pointed-to-object may not be threadsafe, depending on how they are implemented?

LGM-Doyle: The shared pointer can't be safely dereferenced.  It doesn't matter about the type of object.  I can't think of any exceptions, data members, function members, static functions and type info read/write could cause a crash.

Consider a simple case of `shared_ptr<int>`, with two different threads reading the `int`.  

Done correctly, with a mutex or some other technique thread safety, the compiler can guarantee at a hardware level, caches are coherent, reads are well ordered etc. and the two reads will not interfere with each other and will both return the same `int` value.  

Without some sort of thread safely mechanism dereferencing that pointer is undefined behavior, from a best case of two correct reads, to a worst case of a crash.

geoffthemedio: To be clear, I'm / we're talking about two distinct instances of `shared_ptr` that point to / share ownership of the same object, being dereferenced to call a const member function of the object, or to read a POD / built-in type like int, with no modification of the `shared_ptr` or the pointed-to object occurring.

According to http://en.cppreference.com/w/cpp/memory/shared_ptr all the member functions can be called on separate copies of a `shared_ptr` even if they point to the same object. Presumably this includes `get()`, so whatever syncronization issues you're referring to must also apply to raw pointers as well then... (?)

The scenario I'm asking about is two threads accessing the same object through two distinct pointers that point to that same object. The access is only using const member functions or simple reads of the value of the pointed-to data, which don't do anything internal with mutable state or otherwise alter the state of the object or data.

You're saying that's not thread safe and can crash?

LGM-Doyle: tl;dr; Yes, that is not thread safe and can crash.

It is unsafe at the hardware level.  It is described in the [Intel Guide for Developing Multithreaded Applications](https://software.intel.com/en-us/articles/intel-guide-for-developing-multithreaded-applications).  

At the hardware level, a read is not a passive observation.  In the middle of the read the data at the storage location may not exist and will be re-written at the end of the read.  If two threads try to read the same location without synchronization, undefined behavior happens.  

I will try and clarify what I wrote 3 posts ago,
> "shared_ptr and weak_ptr are as thread safe as a raw pointer."  

The smart pointers are designed to be as safe as a raw pointer, but not more.

For example, if you have two threads with a raw pointer in each respective thread pointing to a common object. 

Copying the pointers or reading the address from the pointers with no consideration for thread synchronization is safe because each pointer is in memory only accessed by their respective thread, so those operations don't interact with any shared memory.  

Dereferencing the pointers (any read or write of the pointed to object) without consideration for thread synchronization causes undefined behavior.  This is true even for primitive types, POD, const function or data members, or dynamic casts.

Operations on the pointers themselves are safe.  Any operations on the pointed to object are unsafe.

The std smart pointers, including the `shared_ptr` have the same guarantee.  Operations on the pointers themselves, all of the member functions constructing, copying, moving, etc are safe.  The smart pointers use synchronization techniques internally to guarantee that.  Any operations on the pointed too object is unsafe.

The statement you referenced is intended to clarify that operations on the pointers themselves, specifically copying which increments counters in memory shared between the threads is safe in multiple threads.  

*[edit by @Vezzra: fixed link to intel guide]*

Vezzra: @LGM-Doyle, the link to the intel guide was broken, I went ahead and fixed it.

LGM-Doyle: Thanks.

PR #1545, opened 2017-05-05
Use unique_ptr for the ShipDesign parser.

LGM-Doyle:
**Problem**
The ShipDesignParser is a generator.  In C++11 is should return a `std::unique_ptr<>` to indicate that it has created something that requires memory management and that it is passing management responsibility to the calling function.  The calling function can chose to receive it as a `unique_ptr<>` or transfer it to a `shared_ptr<>` if it needs a shared ownership model.

**Solution**
Return `std::unique_ptr<>`s.

### DIFF
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -124,7 +124,7 @@ namespace {
 
     class SavedDesignsManager {
     public:
-        const std::map<std::string, ShipDesign*>& GetDesigns() const
+        const std::map<std::string, std::unique_ptr<ShipDesign>>& GetDesigns() const
         { return m_saved_designs; }
 
         static SavedDesignsManager& GetSavedDesignsManager() {
@@ -132,10 +132,8 @@ namespace {
             return manager;
         }
 
-        void ClearDesigns() {
-            for (std::map<std::string, ShipDesign*>::value_type& entry : m_saved_designs)
-            { delete entry.second; }
-        }
+        void ClearDesigns()
+        { m_saved_designs.clear(); }
 
         void RefreshDesigns() {
             using namespace boost::filesystem;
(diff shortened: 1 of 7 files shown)


UI/DesignWnd.cpp
```diff
@@ -159,15 +157,12 @@ namespace {
 
             for (const path& design_path : design_files) {
                 try {
-                    std::map<std::string, ShipDesign*> file_designs;
+                    std::map<std::string, std::unique_ptr<ShipDesign>> file_designs;
                     parse::ship_designs(design_path, file_designs);
 
-                    for (std::map<std::string, ShipDesign*>::value_type& design_entry : file_designs) {
+                    for (auto& design_entry : file_designs) {
```
dbenage-cx: Would `auto&&` be preferable here?

LGM-Doyle: I have changed it.

Why do you think that it is significant?

dbenage-cx: After reading a bit more (using more accredited sources), it does not seem to be significant in this case.  I failed to find any adverse effects, possibly providing better support for emplace/forward (though I may be wrong about this as well).

universe/ShipDesign.h
```diff
@@ -614,10 +614,9 @@ class FO_COMMON_API PredefinedShipDesignManager {
 
 private:
     PredefinedShipDesignManager();
-    ~PredefinedShipDesignManager();
 
-    std::map<std::string, ShipDesign*>  m_ship_designs;
-    std::map<std::string, ShipDesign*>  m_monster_designs;
+    std::map<std::string, std::unique_ptr<ShipDesign>>  m_ship_designs;
```
geoffthemedio, line 618: if not lining up with other variable names, no need for the extra space before m_

Issue #1339, opened 2017-02-19
Use delegated constructors to remove code duplication


adrianbroher:
Currently we duplicate the constructor code used to create objects or use a dedicated Init class.  To avoid this unnecessary code duplication C++11 the concept of delegated constructors.  Delegated constructors allow the developer to call another constructor within the same class within the initialization list to remove said code duplication.  For example:

``` C++
class Foo
{
public:
    Foo() :
        Foo("", 0.0f, nullptr)
    {}

    Foo(const std::string& name, float weight, Foo* ancestor) :
       m_name(name),
       m_weight(m_weight),
       m_ancestor(ancestor)
   {
       if (ancestor && name.empty())
           throw std::invalid_argument("Foo::Foo(): foo instances with ancestor require a name");
   }

private:
    std::string m_name;
    float m_weight;
    Foo* ancestor;
};
```

This reduces possible code duplication in the constructor body, initializer lists and removes the need of `Init` member functions.

Pitfalls:

* When delegating the constructor the delegated constructor must be the only element within the initialization list.

Possible candidates for review were collected by calling:

```bash
for CLASS in $(grep --exclude=stutil.py --exclude-dir=cmake --exclude-dir default -rh '\<class\>' * | sed -e 's/^ *//;/^class/!d;/\\$/d;/Calling Visit/d;/cannot call/d;/streaming operator/d;/MultiEditStyle flags/d;s/: .*//;s/FO_COMMON_API //;s/GG_API //;s/;.*$//;s/ {}\?$//;s/ *$//;s/ \/\/.*//;s/ :$//;s/^class //;s/.*:://;/m_self/d;/m_app/d' | sort -u); do
     (echo ${CLASS} && grep -rn --exclude-dir default " \<${CLASS}(" * | sed '/\<new\>/d;/\<void\>/d') >> ../constructor-candidates ;
done
```

from within the repository root.  The list was reviewed after that and the current result can be viewed via:

https://gist.github.com/adrianbroher/2a6eb71a3fc45896b6a8f7dea39cfbf3

memem359:
Is the status still "help wanted"?
I could start working on that suggested list, at the end of the initial post.

But it would help to know if I'm about to duplicate work. (Annoying to spend the time, and then find that someone else already committed the change just before I was about to.)
Or if you want it handled a certain way. (Submit PR for one file at a time, or a bunch, or wait until almost all are ready?)

adrianbroher:
[...]

PR #1561, opened 2017-05-12
Default member initializers to shorten initialization lists

memem359:
Refactoring constructors to shorten the initialization lists, by assigning default member initializers. There should not be any change in game play. Most of these values are placeholders (like ALL_EMPIRES before the empire id is set).

This was inspired by issue #1339 
I was looking for easy ways to trim repetitive code, while trying to understand FO better.
(After these changes are in place, I can think harder about where to use delegated constructors.)


Empire/Empire.h
```diff
@@ -181,13 +177,13 @@ struct FO_COMMON_API ProductionQueue {
...
-        int             turns_left_to_completion;
-        int             rally_point_id;
+        float           allocated_pp = 0.0f;        ///< PP allocated to this ProductionQueue Element by Empire production update
+        float           progress = 0.0f;            ///< fraction of this item that is complete.
+        float           progress_memory = 0.0f;     ///< updated by server turn processing; aides in allowing blocksize changes to be undone in same turn w/o progress loss
+        int             blocksize_memory = 1;       ///< used along with progress_memory
+        int             turns_left_to_next_item = -1;
+        int             turns_left_to_completion = -1;
+        int             rally_point_id = INVALID_OBJECT_ID;
         bool            paused;
```
geoffthemedio, line 187: seems like an odd omission not to default `paused` to `false`

Empire/ResourcePool.h
```diff
@@ -69,8 +70,8 @@ class FO_COMMON_API ResourcePool {
...
     std::map<std::set<int>, float>          m_connected_object_groups_resource_target_output;   ///< cached map from connected group of objects that can share resources, to how much resource would, if all meters equaled their target meters, be output by ResourceCenters in the group.  regenerated during update from other state information.
-    int                                     m_stockpile_object_id;                              ///< object id where stockpile for this pool is located
-    float                                   m_stockpile;                                        ///< current stockpiled amount of resource
+    int                                     m_stockpile_object_id = INVALID_OBJECT_ID;          ///< object id where stockpile for this pool is located
+    float                                   m_stockpile = 0.0;                                  ///< current stockpiled amount of resource
```
geoffthemedio: append f to 0.0

UI/TechTreeLayout.h
```diff
@@ -134,18 +134,18 @@ class TechTreeLayout::Node {
...
     void    CreateEdges(double x_margin, double column_width, double row_height);
 
-    int                 m_depth;        // depth 1 available at beginning 2 one requisite etc
-    int                 m_row;          // layout row, every node is organized in a straight tabelle system
+    int                 m_depth = -1;   // depth 1 available at beginning 2 one requisite etc
+    int                 m_row = -1;     // layout row, every node is organized in a straight tabelle system
     std::string         m_tech;         // name
-    double              m_x;            // left border
-    double              m_y;            // top border
+    double              m_x = 0;        // left border
```
geoffthemedio: 0.0

combat/CombatLogManager.cpp
```diff
@@ -64,18 +64,10 @@ namespace {
...
-
+CombatParticipantState::CombatParticipantState()
+{}
 
-CombatParticipantState::CombatParticipantState(const UniverseObject& object):
-current_health(0.0f),
-max_health(0.0f)
-{
+CombatParticipantState::CombatParticipantState(const UniverseObject& object) {
     FillState(*this, object);
```
geoffthemedio: {} on one line

universe/Building.h
```diff
@@ -61,13 +61,7 @@ class FO_COMMON_API Building : public UniverseObject {
...
     friend class Universe;
     /** \name Structors */ //@{
-    Building() :
-        UniverseObject(),
-        m_building_type(),
-        m_planet_id(INVALID_OBJECT_ID),
-        m_ordered_scrapped(false),
-        m_produced_by_empire_id(ALL_EMPIRES)
-    {}
+    Building() {}
```
geoffthemedio, line 64: {} to newline to be consistent

universe/Fleet.h
```diff
@@ -140,17 +140,7 @@ class FO_COMMON_API Fleet : public UniverseObject {
...
-        m_prev_system(INVALID_OBJECT_ID),
-        m_next_system(INVALID_OBJECT_ID),
-        m_aggressive(true),
-        m_ordered_given_to_empire_id(ALL_EMPIRES),
-        m_travel_route(),
-        m_travel_distance(0.0),
-        m_arrived_this_turn(false),
-        m_arrival_starlane(INVALID_OBJECT_ID)
-    {}
+    Fleet() {}
```
geoffthemedio, line 143: {} 2 newline

universe/Meter.h
```diff
@@ -49,13 +49,13 @@ class FO_COMMON_API Meter {
     void        BackPropagate();                    ///< sets previous equal to initial, then sets initial equal to current
     //@}
 
-    static const float DEFAULT_VALUE;               ///< value assigned to current or initial when resetting or when no value is specified in a constructor
+    static constexpr float DEFAULT_VALUE = 0.0f;        ///< value assigned to current or initial when resetting or when no value is specified in a constructor
```
geoffthemedio: dedent comment start to match

universe/Planet.h
```diff
@@ -1,6 +1,7 @@
 #ifndef _Planet_h_
 #define _Planet_h_
 
+#include "Enums.h"
```
geoffthemedio: @adrianbroher thoughts on extra include vs. initializer lists in .cpp file?

universe/Ship.h
```diff
@@ -1,11 +1,15 @@
 #ifndef _Ship_h_
 #define _Ship_h_
 
+#include "ShipDesign.h"
```
geoffthemedio: I'd probably move `INVALID_DESIGN_ID` out of the `ShipDesign` class rather than require this include...

universe/ShipDesign.h
```diff
@@ -537,25 +537,25 @@ class FO_COMMON_API ShipDesign {
...
-    float   m_structure;
-    float   m_speed;
-    float   m_research_generation;
-    float   m_industry_generation;
-    float   m_trade_generation;
-    bool    m_is_production_location;
+    bool    m_is_armed = false;
+    bool    m_has_fighters = false;
+    bool    m_can_bombard = false;
+    float   m_detection = 0.0;
```
geoffthemedio: append f, and on subsequent lines

util/Order.cpp
```diff
@@ -106,27 +100,17 @@ void RenameOrder::ExecuteImpl() const {
...
-{
-    m_fleet_names.push_back(fleet_name);
-    m_fleet_ids.push_back(fleet_id);
-    m_ship_id_groups.push_back(ship_ids);
-    m_aggressives.push_back(aggressive);
-}
-
+    NewFleetOrder(empire, std::vector<std::string>(1,fleet_name),
+                  std::vector<int>(1,fleet_id),
+                  system_id, std::vector<std::vector<int>>(1,ship_ids),
```
geoffthemedio: spaces after , in parameter lists

util/Order.h
```diff
@@ -28,12 +28,11 @@ class ShipDesign;
 class FO_COMMON_API Order {
 public:
     /** \name Structors */ //@{
-    Order();
+    Order() {}
```
geoffthemedio: 2 newline

util/XMLDoc.h
```diff
@@ -181,10 +167,9 @@ class FO_COMMON_API XMLElement
      * @param[in] tag  The tag name of this XML element.
      * @param[in] text  The text assigned to this XML element.
      */
-    XMLElement(const std::string& tag, const std::string& text) :
+    XMLElement(const std::string& tag, const std::string& text = "") :
```
geoffthemedio: make `explicit`

universe/Ship.h
adrianbroher: That's okay for now.  Moving the `INVALID_DESIGN_ID` out of the `ShipDesign` class to allow making it `extern` would be a good future issue anyway.

universe/Planet.h
adrianbroher: > thoughts on extra include vs. initializer lists in .cpp file?

If it can be avoided (with `extern` eg) try to avoid it.  I would leave it as suggested in this case as `Enums.h` doesn't pull in everything as other headers would do.

adrianbroher:
Another problem in general.  The lines you touched may contain trailing comments, which are an eyesore.  I would prefer if you move those comments above the element being commented to avoid overlong lines (anything out of sight is out of mind) and the pointless `source code and spreadsheet tables are the same thing so everything is almost but not always aligned` game.  Same goes for the indentation between type and member variable name.

universe/Planet.h
geoffthemedio: I'm a bit more hesitant with including `Enums.h` in unnecessary places because it can be changed for some many unrelated reasons. I suppose there's a difference between "makes lots of stuff need a recompile" vs. "makes recompiling lots of stuff slower", though.

geoffthemedio:
Whereas I still find that horizontal alignment of names in declarations makes it easier to read and quickly parse.

Empire/Empire.h
memem359: I was looking for (easy) cases where I would obviously save some lines of code. In this case, I only noticed the one line in the Empire.cpp file, and didn't notice the 3 ctor definitions in the .h file that have default parameters for "paused".

If it it okay, I'll save that for a future PR-- check how ProductionQueue::Element is being used in the code, and see if refactoring makes sense.

Empire/ResourcePool.h
memem359: Yeah, I was copying over the existing values (which were in the initializer list, as you can see in the cpp file), and wasn't sure if I should make the change... I will make the change now.

universe/Planet.h
memem359: I will roll back the 3 lines that use PT_TERRAN and PT_SIZE, and then I won't need to move Enums.h

This was (accidentally) left over from an earlier attempt at this PR, when I tried being more aggressive in using default member inits (to get rid of as many init lists as possible, regardless of how many code lines it would save).

memem359:
As much as possible, I tried to leave the comments and spacing as they were, to match up with the style of the rest of the file. I'm not happy with some of the previous style choices (like a few really long trailing comments), but I figured it would be better to minimize the changes I made for this PR.

To really get rid of the "eyesores" would mean adjusting almost every other line of the header files, to make the style consistent. I have been told that it is better not to mix in stylistic changes for a PR being done for a different reason.

universe/Building.h
memem359: I don't see the style conflict. (What should I be looking at?)
The public destructor a few lines down has the {} on the same line.

universe/Fleet.h
memem359: Should I still do this?
I figured the (local) style was for 1 line.
The public destructor has {} on the same line, and a little further up there are functions like SetArrivalStarlane that have the body (with a return value) on the same line.

universe/Building.h
geoffthemedio: various other cases in this PR have the {} on the next line. If it's consistent within the file then nevermind though.

universe/Ship.h
memem359: I guess I know what I should work on next...

combat/CombatLogManager.cpp
memem359: done

memem359:
Thanks for the comments and corrections.
I've made the last commit in response to the review comments.
If I need to take further action (if a necessary change was missed, or the commits need squashing/rebasing), please let me know.

geoffthemedio:
Other than needing a rebase, looks OK. I'd err on the side of giving everything a default value rather than being selective based on whether they are given different values in different constructors though. Less trouble to keep track of or update that way.

memem359:
The rebase/fixes are finished, assuming that the automated checks clear.
(ShipDesign.h no longer needs to be included in Ship.h)

I expect to try a few more rounds of similar refactoring in the future, so I will look to "err on the side of giving everything a default value".

memem359:
Found a problem.
I'll write again after I find the fix and commit it.

When I add a ship to the production queue, the number of ships (instead of being 1x) is a very large number. I can then change it to 1x, but it shouldn't start that way. Adding a building has a nonsense number of PP required. (The cost listed in the table of available builds is normal; something is getting messed up when added to the queue.)

memem359:
Problem fixed, and I did another master-rebase.
As far as I can tell, the game plays normally with this branch now.

(I needed to give a default value for the "blocksize" variable.
I had done that when I was testing the code before the PR, but somehow it didn't make it into the recent commits.)

combat/CombatLogManager.cpp
geoffthemedio: really? I don't see the change...

memem359: I thought you meant line 67 for the call with the empty argument list.
That used to have {} on a new line (compared to before commit 	c36b834 ).

If you are referring to the function that uses "object" as an argument, then it shouldn't have {} on one line because the function body is not empty: `FillState(*this, object);`.

geoffthemedio: The { } around FillState

memem359: Oh, I see what you mean. Thanks for clarifying. Just committed the change.

PR #1562, opened 2017-05-12
Improve AI trooper movement

### LINKED ISSUE #1538: AI sends troop ships prematurely
Version: 2017-04-24.60d06dc (0.4.7 RC2)

When trying to invade a planet, the AI is supposed to wait out of supply until it has military support in the target system (if the enemy has shields on the planet).
Relevant LOC:
https://github.com/freeorion/freeorion/blob/master/default/python/AI/fleet_orders.py#L36

However, the militarySystemIDs are filled whenever military fleets are *allocated* to that system not when they arrive. This means that the troop ships will be sent several turns before any military support is close. It is thus both easy to predict and to stop the impending invasion.

A stopgap solution could be to wait in supply until the military ships are at / next to the target system. 

A more sophisticated and better solution could be to link the relevant fleets, calculate the ETA for the military fleet and time the invasion fleet accordingly. If their path can be combined such as the invasion fleet follows the military fleet through dangerous territory, that would also be a good addition.

@Dilvish-fo : Any further thoughts on this? IIRC, you implemented the current waiting condition.

Morlic-fo:
Currently, the AI will advance its troop ships beyond its supply as soon as it has allocated military resources to the target system regardless how long the military fleet will take to arrive.

This PR adresses the issue by calculating the estimated time it will take the military fleets to arrive at the target system and only advance the troop ships, if those would arrive later than the military fleets.

An additional commit adresses another flaw of the current logic: The AI would correctly wait at its borders but once it started moving into enemy territory, it would not stop despite not having military resources allocated.
In this PR, the behaviour is checked to always require a military fleet (with correct arrival time, see above) if trooper want to advance into unsupplied systems regardless of their current position.

The behaviour of the AI can be further improved by making sure that the troopers will stay at a safe distance, preferably out of enemy scout range. This is intended for a later PR.

Resolves #1538

### DIFF
--- a/default/python/AI/fleet_orders.py
+++ b/default/python/AI/fleet_orders.py
@@ -43,13 +43,19 @@ def trooper_move_reqs_met(main_fleet_mission, order, verbose):
                            "assigned to secure the target system.") % (order.fleet.id, invasion_planet)
                 return False
 
-            eta_this_fleet = FleetUtilsAI.calculate_estimated_time_of_arrival(order.fleet.id, invasion_system.id)
-            if all((eta_this_fleet - 1) < FleetUtilsAI.calculate_estimated_time_of_arrival(fid, invasion_system.id)
-                    for fid in military_support_fleets):
+            # if there is a threat in the enemy system, do give military ships at least 1 turn to clear it
+            delay_to_move_troops = 1 if MilitaryAI.get_system_local_threat(order.target.id) else 0
+
+            def eta(fleet_id):
+                return FleetUtilsAI.calculate_estimated_time_of_arrival(fleet_id, invasion_system.id)
+
+            eta_this_fleet = eta(order.fleet.id)
+            if all(((eta_this_fleet - delay_to_move_troops) <= eta(fid) and eta(fid)) for fid in military_support_fleets):
                 if verbose:
                     print ("trooper_move_reqs_met() holding Invasion fleet %d before leaving supply "
                            "because target (%s) has nonzero max shields and no assigned military fleet would arrive"
-                           "at least 1 turn earlier than the invasion fleet") % (order.fleet.id, invasion_planet)
+                           "at least %d turn earlier than the invasion fleet") % (
+                                order.fleet.id, invasion_planet, delay_to_move_troops)
                 return False
 
         if verbose:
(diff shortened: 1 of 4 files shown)


default/python/AI/fleet_orders.py
```diff
@@ -27,24 +28,35 @@ def trooper_move_reqs_met(main_fleet_mission, order, verbose):
...
+            military_support_fleets = MilitaryAI.get_military_fleets_with_target_system(invasion_system.id)
+            if not military_support_fleets:
+                if verbose:
+                    print ("trooper_move_reqs_met() holding Invasion fleet %d before leaving supply "
+                           "because target (%s) has nonzero max shields and there is not yet a military fleet "
+                           "assigned to secure the target system.") % (order.fleet.id, invasion_planet)
+                return False
+
+            eta_this_fleet = FleetUtilsAI.calculate_estimated_time_of_arrival(order.fleet.id, invasion_system.id)
+            if all((eta_this_fleet - 1) < FleetUtilsAI.calculate_estimated_time_of_arrival(fid, invasion_system.id)
```
Dilvish-fo: Seems to me like this should either drop the -1 or make the comparison be <=.

Also, just noting, I expect that at some point in the future we'd want to be able to allow, under certain circumstances, for the troopers and warfleet to arrive on the same turn.  Perhaps add a todo comment for that?

Dilvish-fo:
The logic sounds good, and other than the one comment the changes look good to me.  Did you have a test case to be able to test out the actual operation of the change?

Morlic-fo:
Thanks for the review. Inspired by your remarks, I added two more points to the logic:

1. The AI is now allowed to move with the military ships into the target system if there is no enemy firepower (i.e. threat) present. Further conditions could be applied in future PRs but this is simple and does not contradict the movement logic outside of this function which does not allow moving into a system with enemy threat unless secured.

2. The AI will now not wait at the supply borders but rather at least 4 jumps away from the target system. This means safer positioning for the troop ships as well as hiding the intentions of the AI a little bit better while still being relatively "soon" at the target location if a nearby military fleet is allocated.
That commit is based on a change to the universe.jumpDistance function which may or may not require some discussion and is extracted in #1564 which should be merged first.


Regarding the test case: With the new additional logic of having the AI wait inside the supply, it becomes rather easy to see and confirm the effect of the change. The AI seems to correctly wait for military support and is then able to invade its target.
There are the usual problems with the AI suiciding its military escort into the target system because it has no vision of and the enemy has sent ships there in the meantime but that seems unrelated to this PR...

PR #1563, opened 2017-05-13
Let the AI attack enemy fleets

Morlic-fo:
Currently, the AI won't bother attacking enemy scouts, troop fleets or military fleets (unless it is forced to in order to achieve another target).

This PR provides additional military allocations to attack enemy fleets.
A relatively easy test of the new behaviour is to send a scout ship into / next to enemy supply. Unless the AI is preoccupied (or at least thinks it is), it will try to hunt down the scout.

Before merge, several issues must be adressed (input is welcome @freeorion/ai-team):

1. The AI is currently still sending troop ships / scout ships next to enemy warships. With this PR, that behaviour is heavily punished and the AIs will cripple each other.

2. There should be additional logic to let the AI not be baited by, e.g. sending a fast scout to draw forces away from some strategic objective.
Possibly, a new mission could be introduced for attacking / hunting down fleets which will be reset every single turn in order to free resources for more urgent tasks.

3. The logic should be extended to predict enemy movement. At least, it should be able to intercept fleets that are travelling on starlanes by moving into the neighboring system.

### DIFF
--- a/default/python/AI/MilitaryAI.py
+++ b/default/python/AI/MilitaryAI.py
@@ -385,7 +385,7 @@ class PlanetDefenseAllocator(Allocator):
             super(PlanetDefenseAllocator, self).allocate()
             return
         if self._minimum_allocation(self._calculate_threat()):
-            pass  # raise ReleaseMilitaryException TODO
+            pass
 
     def _minimum_allocation(self, threat):
         super_call = super(PlanetDefenseAllocator, self)._minimum_allocation(threat)
@@ -513,6 +513,23 @@ class BorderSecurityAllocator(LocalThreatAllocator):
 
     def _maximum_allocation(self, threat):
         return self._max_alloc_factor * self.safety_factor * max(self._local_threat(), self._neighbor_threat())
+
+
+class AttackEnemyFleetsAllocator(Allocator):
+    _allocation_group = 'enemyFleets'
+    _min_alloc_factor = 4  # make sure we only attack enemy fleets if we expect low losses
+    _max_alloc_factor = 6
+    _potential_threat_factor = 0
+
+    def _calculate_threat(self):
+        if not self._enemy_ship_count():
+            return 0.
+        retval = self._enemy_ship_count()*5 + CombatRatingsAI.combine_ratings(
+                self._local_threat(), self._neighbor_threat())
+        return retval
+
+    def _take_any(self):
+        return False
 
 
 class ReleaseMilitaryException(Exception):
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
Yup this is one of those things that I'm sure has been nagging at us all for a long while, I'm glad to see some progress on it.  Regarding your noted points:

> 1. The AI is currently still sending troop ships / scout ships next to enemy warships. With this PR, that behaviour is heavily punished and the AIs will cripple each other.

Adding some retreat-from-threat code would do a lot to cover these situations and the broader class of similar situations created by enemy movement rather than self movement.  I think that would give most of what scouts need (though probably they should be assigned a new exploration target if their current one is blocked by a mobile empire enemy).  I agree that troopers (and outpost/colony ships) should have much more restrictive criteria for moving towards a system that has neighbor threat.

> 2. There should be additional logic to let the AI not be baited by, e.g. sending a fast scout to draw forces away from some strategic objective.  Possibly, a new mission could be introduced for attacking / hunting down fleets, which will be reset every single turn ...

A few thoughts related to this:  I think the specialized mission would be very helpful.  It would probably also be helpful to have some kind of fast picket scout hunter subclass of the MilitaryShipDesigner which would be preferentially used for these missions.  Perhaps only do the single-turn reset of the scout hunter missions which are being fulfilled by a ship (or ships) with a non-scout-hunter design.   (sidenote, also, it's been nagging me that I think our MilitaryShipDesigner and WarShipDesigner names should be swapped; the former seems like the much more generic name to me.  It would be a bit of a nuisance to do, but I think it should be done.)

On the baiting point, currently it looks to me like this would be assigning enough firepower to cover all neighbor threat, am I reading that right?  I don't think we always need to.  I think that, beyond the actual target threat there should be a bit of a stochastic determination of how much neighbor threat to provide coverage for, at least if there was some other task that we might want to use the remaining fleet strength on.  Also we should consider travel speeds here --  If we could send a relatively cheap fast picket that could arrive in one turn and kill a trooper, scout, outpost ship, or even a smaller warship, and all the neighbor threat would require 2 turns to reach the target system, then just send the fast picket.  Also if our fleets were not going to be able to arrive in a single turn, then we should be much more restrictive about when we would try that move.

> 3. The logic should be extended to predict enemy movement. At least, it should be able to intercept fleets that are travelling on starlanes by moving into the neighboring system.

I am pretty sure that as things stand, because of how the system status tracking of enemies works with moving fleets, this will already be able to intercept moving fleets if they stop at their current immediate destination.  Predicting whether or not they will actually stop, and what their next leg might be if they do not, will certainly be a helpful further development.

geoffthemedio:
@Dilvish-fo @Morlic-fo  what's the status on this now?

Vezzra:
@Morlic-fo, @Dilvish-fo, bump?

Morlic-fo:
Will be part of a more thorough military AI revision (probably for 0.4.9).

PR #1564, opened 2017-05-13
Use generalized JumpDistanceBetweenObjects in python universe wrapper

Morlic-fo:
As far as I can see, the behaviour of ```JumpDistanceBetweenObjects``` is the same as ```JumpDistanceBetweenSystems``` if system_ids are passed but also allows fleet_ids as arguments which is a useful feature for the AI.

Having a single interface is simpler and cleaner so I would rather replace the existing function than introduce another one. Any concerns about that?

### DIFF
--- a/python/UniverseWrapper.cpp
+++ b/python/UniverseWrapper.cpp
@@ -331,7 +331,13 @@ namespace FreeOrionPython {
                                                     JumpDistanceFunc,
                                                     return_value_policy<return_by_value>(),
                                                     boost::mpl::vector<int, const Universe&, int, int>()
-                                                ))
+                                                ),
+                                                "If two system ids are passed or both objects are within a system, "
+                                                "return the jump distance between the two systems. If one object "
+                                                "(e.g. a fleet) is on a starlane, then calculate the jump distance "
+                                                "from both ends of the starlane to the target system and "
+                                                "return the smaller one."
+                                                )
 
             .def("shortestPath",                make_function(
                                                     ShortestPathFunc,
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
The general principle sounds fine, and for the particular purpose you currently contemplate with #1562 I think this change would work pretty much fine, but there is a complication that I think should be noted with a docstring being provided for the python exposure of [jumpDistance](https://github.com/freeorion/freeorion/blob/master/python/UniverseWrapper.cpp#L330-L334).    The complication is    [For an object in transit return the distance to the closest system.](https://github.com/freeorion/freeorion/blob/master/universe/Pathfinder.cpp#L778-L779)  And to clarify that comment slightly, it is referring to whichever end of the starlane is closest to the target system, not to which end of the starlane is the ship/fleet closest to.  (I expect that the choice to code it this way was driven by FOCS scripting concerns for Effects scope conditions.)  

For your currently contemplated purpose, I think the only effect of the complication would be that if a troopship had its target changed while it was in-transit between systems, then it might wind up waiting an extra turn at its current destination when really it might be OK for it to move right away, but that's a rare, minor, and probably conservative variation that would be fine.   The difference though might be more important for any other uses we would start coming up with for this function, so I think it should be noted. 
 My recollection is that the docstring is just added as an extra parameter in the exposure, like was done for [validShipDesign](https://github.com/freeorion/freeorion/blob/master/python/UniverseWrapper.cpp#L511)

Morlic-fo:
Thanks for pointing that out. I added a docstring to the function definition.

Is the formatting ok? The only other long docstring I found (for validShipDesign) is written all in one line which results in a kind of awkward 428 characters line... My version seems better readable but can change if prefered...

Morlic-fo:
Any objections against merging this?

Dilvish-fo:
I finally got around to checking out the PR to confirm that the formatting looks fine, and yes, it looks fine to me.  I'd say go ahead and merge.

PR #1565, opened 2017-05-13
Swap names of WarShipDesigner and MilitaryShipDesigner

Morlic-fo:
As suggested by @Dilvish-fo

### DIFF
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -5,13 +5,13 @@ implement the rating function and some additional information for the optimizing
 
 Example usage of this module:
 import ShipDesignAI
-myDesign = ShipDesignAI.MilitaryShipDesigner()
+myDesign = ShipDesignAI.WarShipDesigner()
 myDesign.additional_specifications.enemy_mine_dmg = 10
 best_military_designs = myDesign.optimize_design()  # best designs per planet: (rating,planetID,design_id,cost) tuples
 
 
 Available ship classes:
-- MilitaryShipDesigner: basic military ship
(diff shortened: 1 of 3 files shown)


default/python/AI/ShipDesignAI.py
```diff
@@ -1581,18 +1581,18 @@ def _combat_rating(self):
         return combat_stats.get_rating(enemy_stats=self.additional_specifications.enemy)
 
 
-class WarShipDesigner(ShipDesigner):
+class MilitaryShipDesigner(ShipDesigner):
     basename = "Warship (do not build me)"
```
Dilvish-fo: Seems to me we should probably also change this 'Warship' to something like 'Military ship base'

Dilvish-fo (approved):

default/python/AI/ShipDesignAI.py
Morlic-fo: We have the ```TroopShipDesignerBaseClass``` and ```ColonisationShipDesignerBaseClass```

So, I would go with MilitaryShipDesignerBaseClass``` unless you have a better idea including the other base classes...

Dilvish-fo: Are you talking about line 1584?  For that either the current state of the PR or else MilitaryShipDesignerBaseClass would either be fine with me.  I was referring to the 'Warship' in line 1585, though (I am pretty sure that is the line I had selected when I made the comment, but I should have double checked how things appeared once I actually entered the comment)

Morlic-fo: Oh, you are talking about the ```basename``` variable. Good point, I overlooked that.

PR #1568, opened 2017-05-14
Move ShipDesign::INVALID_DESIGN_ID to extern

memem359:
While working on PR #1561 , there was the suggestion to make ShipDesign::INVALID_DESIGN_ID into a FO_COMMON_API extern variable. This will reduce the number of times that ShipDesign.h is "included".

In particular, this allows me to remove ShipDesign.h from Empire.h.
(Empire.h is used in 61 cpp files and Order.h. Many of those do not need ShipDesign.h.)

I also removed ShipDesign::MAX_ID.
It was only used in one line in the entire code, in Universe.cpp
In that file, I replaced ShipDesign::MAX_ID with the local variable MAX_ID, as they both have the same value.

(This commit will cause a conflict with 1561. I figured I would wait to see which one is merged with master first, then rebase/fix the other PR at that time.)

### DIFF
--- a/universe/ShipDesign.cpp
+++ b/universe/ShipDesign.cpp
@@ -19,6 +19,8 @@
 #include <cfloat>
 #include <boost/filesystem/fstream.hpp>
 
+extern FO_COMMON_API const int INVALID_DESIGN_ID = -1;
+
 using boost::io::str;
 
 namespace {
@@ -517,10 +519,6 @@ HullTypeManager::iterator HullTypeManager::end() const
 ////////////////////////////////////////////////
 // ShipDesign
 ////////////////////////////////////////////////
-// static(s)
-const int       ShipDesign::INVALID_DESIGN_ID = -1;
-const int       ShipDesign::MAX_ID            = 2000000000;
-
 ShipDesign::ShipDesign() :
     m_id(INVALID_OBJECT_ID),
     m_name(),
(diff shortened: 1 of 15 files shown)


universe/ShipDesign.cpp
```diff
@@ -19,6 +19,8 @@
 #include <cfloat>
 #include <boost/filesystem/fstream.hpp>
 
+extern FO_COMMON_API const int INVALID_DESIGN_ID = -1;
```
geoffthemedio, line 22: is the `extern FO_COMMON_API` part necessary / appropriate in the .cpp file?

geoffthemedio:
Otherwise looks OK to me.

universe/ShipDesign.cpp
memem359: I don't know if it is necessary or not.
BUT... I copied the ALL_EMPIRES assignment, which does have it. (Universe.cpp, line 73)
I figured it would be safer to copy what seems to working, rather than try to figure it out myself.

Edit: Most (all?) of the files in the Universe folder are part of the Common build target.
I figured that required the use of FO_COMMON_API ?

geoffthemedio: In the header, yes, but I doubt it's necessary in the .cpp.

memem359: I am going to make a guess about the situation.
In the cases where the variable is already declared in the header (like INVALID_GAME_TURN and INVALID_OBJECT_ID), the assignment in the .cpp does not have "extern FO_COMMON_API".

In this case, I have not declared INVALID_DESIGN_ID in ShipDesign.h. (That is also true for ALL_EMPIRES, which is assigned in Universe.cpp but not declared in Universe.h.)

If the suggestion is to make the "extern FO_COMMON_API" go away from the cpp file, I will likely need to add it to the ShipDesign.h file first. Let me know if you want this to happen, or to leave it in peace.

geoffthemedio:
builds, seems to run OK

PR #1576, opened 2017-05-17
Refactor use of OrderPtr alias.

LGM-Doyle:
This PR improves the use of the `OrderPtr` alias by doing two things.

This PR replaces 
````OrderPtr( new X())````
code with 
````std::make_shared<X>()````

which is more exception safe and may result in fewer allocations if the `shared_ptr` control block is allocated with the `X`

The PR uses `auto` in all `OrderSet` for loops, replacing `entry` with `id_and_order` to increase clarity

### DIFF
--- a/UI/ProductionWnd.cpp
+++ b/UI/ProductionWnd.cpp
@@ -930,9 +930,9 @@ void ProductionWnd::QueueItemMoved(GG::ListBox::Row* row, std::size_t position) 
         return;
 
     HumanClientApp::GetApp()->Orders().IssueOrder(
-        OrderPtr(new ProductionQueueOrder(client_empire_id,
-                                          boost::polymorphic_downcast<QueueRow*>(row)->queue_index,
-                                          position)));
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: UI/ProductionWnd.cpp [hunk 1]
```diff
@@ -1126,8 +1126,8 @@ void ProductionWnd::QueueItemRallied(GG::ListBox::iterator it, int object_id) {
         return;
 
     HumanClientApp::GetApp()->Orders().IssueOrder(
-        OrderPtr(new ProductionQueueOrder(client_empire_id, std::distance(m_queue_wnd->GetQueueListBox()->begin(), it),
-                                          rally_point_id, false, false)));
+        std::make_shared<ProductionQueueOrder>(client_empire_id, std::distance(m_queue_wnd->GetQueueListBox()->begin(), it),
+                                          rally_point_id, false, false));
```
geoffthemedio: indenting

geoffthemedio:
otherwise looks OK

Issue #1573, opened 2017-05-16
dropdown mousewheel scroll anomalies/bugs


Dilvish-fo:
### Environment
* **Operating System**: Ubuntu Linux
* **Fetched as** 
  * Compiled from source

### Description
There are two aspects to this, the first is nuisance behavior, the second is a bug
a) *Unintended dropdown scrolling of planetary focus and of production queue repeat/blocksize* -- Neither of these require clicking on the dropdown element to activate mousewheel scrolling, and so it can easily happen by accident when in the process of mousewheel-scrolling the larger listboxes (SidePanel, and ProductionQueue) which contain the panels on which these dropdowns appear.  This causes actual changes/problems for the planetary focii, though as noted below it may only cause a display glitch for the ProductionQueue.
b) *mismatch between display and actual setting for mousewheel-scrolled ProductionQueue dropdowns*
If using only mousewheel scroll to change a productionqueue item panel repeat or blocksize dropdown (without afterwards left-clicking on the dropdown), the numeric label on the dropdown will change, but it appears that the status of the item does not actually change -- the reported PP this turn and until completion will not change, and upon advancing to the next turn the dropdowns for that item will have reverted to their previous values.

### Expected Result
UI elements should not be excessively easy to accidentally change, and their display status should match their actual status.

### Proposed fixes
*  for accidental scrolling-- enable mousewheel scrolling of dropdowns only while either (i) they have been left-clicked to open up their dropdown listing, or (ii) the shift key is being held down
* for production panel display mismatch -- if the display label is changed via mousewheel scroll, ensure that the actual setting is changed also

geoffthemedio:
These are two almost unrelated issues.

The production batch size and repetitions one is longstanding, and I'm slightly surprised there isn't already a bug about it.

Dilvish-fo:
I'll agree the two aspects are distinct, but they are also related-- if the resolution for the first issue were to only enable the dropdown mousewheel scroll upon left-clicking the dropdown, then the second issue would already be fixed by that.

LGM-Doyle:
@Dilvish-fo I added a [PR](https://github.com/freeorion/freeorion/pull/1577) that restricts the mouse wheel scrolling of all `DropDownLists` as you suggested.

Dilvish-fo:
[...]

PR #1577, opened 2017-05-19
Only enable DropDownList mouse scrolling when dropped or shift keyed pressed.

LGM-Doyle:
Addresses the first part of issue #1573.

### DIFF
--- a/GG/src/DropDownList.cpp
+++ b/GG/src/DropDownList.cpp
@@ -482,6 +482,7 @@ boost::optional<DropDownList::iterator> ModalListPicker::MouseWheelCommon(
     }
     if (move != 0) {
         std::advance(cur_it, move);
+        LB()->BringRowIntoView(cur_it);
         return cur_it;
     }
     return boost::none;
--- a/GG/src/DropDownList.cpp
+++ b/GG/src/DropDownList.cpp
@@ -122,7 +122,7 @@ private:
     const DropDownList* m_relative_to_wnd;
     bool                m_dropped; ///< Is the drop down list open.
 
-    /** Should the list wnd scroll when dropped? */
+    /** Should the list wnd scroll only when dropped? */
     bool                m_only_mouse_scroll_when_dropped;
 };
 
--- a/GG/src/DropDownList.cpp
+++ b/GG/src/DropDownList.cpp
@@ -93,6 +93,9 @@ public:
     boost::optional<DropDownList::iterator> MouseWheelCommon(
         const Pt& pt, int move, Flags<ModKey> mod_keys);
 
+    /** Set the drop down list to only mouse scroll if it is dropped. */
+    void SetOnlyMouseScrollWhenDropped(bool enable);
+
 protected:
     /** ModalListPicker needs to process its own key press events because modal
         windows in GG can't have parents. */
@@ -118,6 +121,9 @@ private:
     const size_t        m_num_shown_rows;
     const DropDownList* m_relative_to_wnd;
     bool                m_dropped; ///< Is the drop down list open.
+
+    /** Should the list wnd scroll when dropped? */
+    bool                m_only_mouse_scroll_when_dropped;
 };
 
 namespace {
@@ -159,7 +165,8 @@ ModalListPicker::ModalListPicker(Clr color, const DropDownList* relative_to_wnd,
     m_lb_wnd(GetStyleFactory()->NewDropDownListListBox(color, color)),
     m_num_shown_rows(std::max<std::size_t>(1, num_rows)),
     m_relative_to_wnd(relative_to_wnd),
-    m_dropped(false)
+    m_dropped(false),
+    m_only_mouse_scroll_when_dropped(false)
 {
     m_lb_wnd->SelChangedSignal.connect(
         boost::bind(&ModalListPicker::LBSelChangedSlot, this, _1));
@@ -445,9 +452,15 @@ boost::optional<DropDownList::iterator> ModalListPicker::KeyPressCommon(
     return boost::none;
 }
 
+void ModalListPicker::SetOnlyMouseScrollWhenDropped(bool enable)
+{ m_only_mouse_scroll_when_dropped = enable; }
+
 boost::optional<DropDownList::iterator> ModalListPicker::MouseWheelCommon(
     const Pt& pt, int move, Flags<ModKey> mod_keys)
 {
+    if (m_only_mouse_scroll_when_dropped && !Dropped())
+        return boost::none;
+
     DropDownList::iterator cur_it = CurrentItem();
     if (cur_it == LB()->end())
         return boost::none;
@@ -882,6 +895,9 @@ void DropDownList::MouseWheel(const Pt& pt, int move, Flags<ModKey> mod_keys)
         Control::MouseWheel(pt, move, mod_keys);
     }
 }
+
+void DropDownList::SetOnlyMouseScrollWhenDropped(bool enable)
+{ m_modal_picker->SetOnlyMouseScrollWhenDropped(enable); }
 
 ListBox* DropDownList::LB()
 { return m_modal_picker->LB(); }
--- a/GG/src/DropDownList.cpp
+++ b/GG/src/DropDownList.cpp
@@ -525,9 +525,12 @@ void ModalListPicker::KeyPress(Key key, std::uint32_t key_code_point, Flags<ModK
 
 void ModalListPicker::MouseWheel(const Pt& pt, int move, Flags<ModKey> mod_keys)
 {
-    if (!LB()->InWindow(pt))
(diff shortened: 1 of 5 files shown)


geoffthemedio:
As a general rule, I'm opposed. There are plenty of cases where being able to switch what is selected with just the mousewheel are very convenient. Pressing shift (I'd have used CTRL) is not obvious and probably more awkward to do than opening the list, as it requires more than just using the mouse.

If implemented, this should be a flag, and only set in specific cases where it is a problem, such as when the containing Wnd is often scrolled.

Better would be some way to detect that the containing Wnd is being scrolled, and disable the scrolling of the list when the mouse is only over it due to a mousewheel scroll event, until the mouse leaves and reenters the control by movement.

LGM-Doyle:
I'd agree, I am mildy opposed.  I do find it useful to scroll through lists without dropping them if I know the contents.  I did not have problems with erroneous scrolling.  However, if others feel strongly I figured that I could push it and others will decide.

I used shift because CTRL implies a change in behavior of an action that will actively change something.  Shift just reditects/shifts the input from the top-level UI element to the drop down list.

Dilvish-fo:
> There are plenty of cases where being able to switch what is selected with just the mousewheel are very convenient

I suspect the cases you are referring to are in the Game Setup screens?  It hasn't been working in the Production Panels, and the only other case I can think of is the SidePanel.   I guess I wasn't quite clear enough in that issue but I was proposing the restriction only for the two cases that I had cited as problems (ProductionPanel and SidePanel)), where the dropdowns are a subcomponent contained withing a larger scrollable listbox.

I would propose that you leave the default behavior for dropdowns be that mousewheel scrolling can work, but that that it be an optional parameter at creation that makes mousewheel activity more restrictive, and then we only turn that on for dropdowns that appear in panels that themselves are in scrollable listboxes (of which ProductionPanel and SidePanel are the only ones that come to mind).

Geoff's suggestion about making it dependent on whether it was scrolling movement of the listbox bringing the dropdown under the cursor does sound like it would be a fine resolution, perhaps ideal from the end-user standpoint, but sounds quite a bit trickier to code up.

If you guys really wanted to preserver the existing behavior for the SidePanel, howabout making it controllable via an option.

geoffthemedio:
Also on the multiplayer lobby player list, though that's probably a case where it should require a click. Also the objects list filters dialog. Also on the options screen, which is the only place I've unintentionally scrolled the droplist while scrolling the containing page, though in that case I'd be more inclined to reorganize the options page so tthat scrollbars aren't ever needed to see them all.

Dilvish-fo:
@LGM-Doyle I just wanted to mention, for now at least I have added this commit to the handful of personal UI customizations that I normally play with, so thanks for making that easy for me.

Dilvish-fo:
@geoffthemedio  ~~what did you think of my suggestion~~ about making this extra restriction be something controllable via parameter at construction time, and it only be turned on for a handful of cases ~~?~~   It sounds like you would support this restriction for the multiplayer lobby player list, and possibly (at least until such time as they got reorganized to no longer require scrolling) the options pages, and that we would both agree there is no need for the restriction in the objects list filters dialog and in the galaxy setup droplists (since they do not appear on panels within a larger listbox).

I think nobody could be wedded to not having the restriction in the ProductionWnd repeat and blocksize dropdowns, since mouswheel scrolling currently does even not properly work for them, and they are also a particularly dangerous area to allow accidental scrolling, so I am hoping no one objects to this restriction for those dropdowns?

Re the sidepanels, any objection to it being controlled by an option, such as 
```guard-planet-foci-mousewheel-scrolling```

**Comment by geoffthemedio** [comment]
I thought you were agreeing with my above suggestion / insistence to have a flag to do that (which would be a constructor parameter / toggleable setting)...

**Comment by Dilvish-fo** [comment]
Oh, I guess perhaps your suggestion had passed into my subconscious while somehow not latching onto my conscious attention.  So yes, I am agreeing with your suggestion/insistence on that point.

What do you think of my list of particular places the flag would be set to on?  (it's now seeming like I was partly just relisting places you had already meant as proposals).

**Comment by geoffthemedio** [comment]
For the first pass, anywhere that has a containing window that scrolls is reasonable to disable the no-click mousewheeling on droplists.

**Comment by LGM-Doyle** [comment]
I think that this works as requested.

File: GG/src/DropDownList.cpp [hunk 1]
```diff
@@ -118,6 +121,9 @@ class ModalListPicker : public std::enable_shared_from_this<ModalListPicker>, pu
     const size_t        m_num_shown_rows;
     const DropDownList* m_relative_to_wnd;
     bool                m_dropped; ///< Is the drop down list open.
+
+    /** Should the list wnd scroll when dropped? */
```
Dilvish-fo: to be more clear (and consistent with the method comments), "when" --> "only when"

But also, I'll let myself mention a slight generalization here, which would not in any sense be needed but which I think would be kind of nice to do now, or at least I'll mention it now so it's kept in mind -- namely, if this variable and the related methods were renamed somewhat more generically along the lines of "m_restrict_mousewheel_scroll".

That would just be a preparatory change, looking forward to when we could add back the enablement condition you had before, ctrl bring pressed, so that mousewheel scrolling would be enabled if any of the three conditions were met-- the restriction bool is not set, or the dropdown is dropped, or ctrl is being pressed.  I would not propose actually adding back the ctrl enablement at this time, because I believe the ProductionWnd dropdown handling is right now dependent on actual clicks being made in order for changes to the dropdown to get processed properly, so that would need a bit of work done at the same time.

LGM-Doyle: I fixed the comment, but left the variable name as is since that reflects its current usage.

We can change the variable name when we change the code, after the production window bug(s) is fixed.

Keep the original issue open and perhaps copy your comment into the original issue as a reminder.

I'd prefer shift-mousewheel in this case since it modifies the UI behavior, while still doing the same action.

By analogy with a keyboard and the 'q' key: `q` types a lower case 'q', `shift-q`, types an uppercase 'Q' and `ctrl-q` quits the application.  Both `q` and `shift-q` enter a letter into the data stream, the same action.  `ctrl-q` does a different action.  

In our case both `mousewheel` and `shift-mousewheel` scroll something, the whole list or the embedded llist.  I would expect `ctrl-mousehweel` to do something different, like maybe a scrolling delete.

geoffthemedio (commented):
Should probably also do options wnd droplists until they it doesnt have scrollbars.

LGM-Doyle:
Added.

geoffthemedio:
The focus and production quantity droplists don't respond at all to mousewheel events now. They also don't in master when open, but if the idea here is to make it only work when dropped, then they'd need to also work when dropped, in addition to not working when not dropped.

LGM-Doyle:
As you observed this was not working in master before this PR.

I've now fixed it.

Code in GUI.cpp to redirect events, handles keyboard and mouse events for modal windows differently.  The keyboard events were reaching the dropped `ModalListPicker.  The mouse wheel events were being sent to the child `ListBox` which would ignore them if it did not have a scroll bar.  

I've added an event filter to always redirect the mouse wheel events to the `ModalListPicker`

GG/src/DropDownList.cpp
```diff
@@ -30,6 +30,7 @@
 #include <GG/Scroll.h>
 #include <GG/StyleFactory.h>
 #include <GG/WndEvent.h>
+#include "../util/Logger.h"
```
geoffthemedio: we've avoided putting any FreeOron-specific code / headers into GG code, historically because it was / is supposed to be an independent library that happens to only be used by FreeOrion...

regardless, is this even necessary?

LGM-Doyle: Fixed.

I meant to remove it before I pushed this PR.

Keeping GG separate with a distinct responsibilities is good for code quality.

However, GG has no useful file logger with timestamps to see sequencing of events.  `std::cout` is a poor substitute.

All of the FreeOrion dependence in `Logger` is in LoggerWithOptionsDB.  Whenever a logger is helpful for debugging I include it and I've always intended to remove it.

geoffthemedio:
With this pull request applied, the resolution drop list scrolls through the options, but doesn't scroll the list itself, even if the selected item goes out of the shown range. Previously, it would scroll the list, but not change the selected item.

Also, the lists don't respond to mousewheel events over the drop list itself, but rather only on the dropped part of the list that appears when the list is opened.

LGM-Doyle:
@geoffthemedio thanks for finding those bugs. 

I fixed the scrolling and added the ability to scroll while over the anchor portion of the drop list.

Vezzra:
@Dilvish-fo, @LGM-Doyle, has #1573 been sufficiently addressed with this PR merged (so it can be closed as resolved), or is there still more work necessary?

LGM-Doyle:
Yes, I think that it can be closed as resolved.

Vezzra:
@Dilvish-fo? #1573 has been opened by you, so do you think everything in that issue has been addressed?

Issue #1575, opened 2017-05-16
Logger system is currently broken


memem359:
### Environment
* **FreeOrion Version**: Running with the default settings. (Trashed old config files to let the game create one, and started a game with the initial settings.)
* **Operating System**: MacOSX
* **Fetched as**
  * Weekly development build (May 16), but I've been seeing the same behavior the past few weeks when building from (master) source.

### Description
More work needs to be done after the recent logger system PR changes, as it is broken enough to not be useful.

- Currently, the only logger behaving as it should is `InfoLogger(log)`. Those messages (being sent from Logger.cpp, mostly line 286) are the only things I see in the AI and freeorion logs, and it is only for the initial logger threshold settings.

- I tried running from the command line and redirecting standard output to a file. All of the other expected log messages are being sent there (from all the clients), and the file sender is not being converted to human readable format. As an example:
`[2017-05-16 13:59:42.399184] [0xa6f311c0] [info]    ProductionQueue::Update: Simulating future turns of production queue`

-  After loading the hotkey settings and various ship part names, there are 220 lines of messages complaining about config.xml options. As an example: `[2017-05-16 13:57:42.443918] [0xa6f311c0] [info]    Option "UI.windows.design.base-selector.left", was in config.xml but was not recognized.  It may not be registered yet or you may need to delete your config.xml if it is out of date.`
Since this is the default file created by the game, the default values should be updated. (I think a search for `db.Add` will find the necessary code.)
The unrecognized options are: UI.windows.* , app-* (height, left, top, width), logging.execs.* , and logging.sources.*

LGM-Doyle:
@memem359 thanks for reporting this issue.

Could you provide the following to help me debug this issue: the log files freeorion.log, freeoriond.log,  AI_X.logs and the file with the re-directed `std::out`.

What version of OSX are you on?
What version of clang did you compile with?
What version of boost::log did you link to?

thanks.

memem359:
MacOSX 10.12.5 ("Sierra", latest version)

I am using the weekly test build provided by Vezzra (build 2017-05-16.18e65e5).
I only mentioned that I had seen this kind of behavior in my local builds for a few weeks, in case there was concern that it was specifically this released test build.

Specific steps:

1) Make sure all old log and config files were deleted.
2) Start game from command line: `FreeOrion.app/Contents/MacOS/Freeorion >~/Desktop/std_out.txt 2>~/Desktop/std_err.txt`
[...]

PR #1578, opened 2017-05-19
(Mac OSX) Update the Xcode project file to be more like CMakeLists

memem359:
This is mostly due to #1575 . This gets back most of the functionality of the logger system for a Mac OSX build. (Edit: The logger system is broken for a Mac build without these changes.) I have also tweaked other settings to get the project file to look more like the settings that are in CMakeLists.txt. As far as I can tell, this hasn't changed the game behavior (other than restoring the logger).

These changes are easier to see from the Xcode Editor, but not as easy to compare with the state before the commit. Hopefully this list of the chages will make more sense:

- The visibility=hidden was removed from the Parse target compiler flags and from all the various linker flags. It was added to the compiler flags at the Project level (affects all targets unless overridden locally).

- FREEORION_BUILD_PARSE was removed from the Project settings, and added to the Parse target.

- FREEORION_BUILD_COMMON was added to the Common target (was entirely missing before)

- For the Common target, the "link binary with libraries" list was missing the Zlib and the boost date-time and regx libs. I've added them, and reordered the list to make it easier to compare with CMakeLists.


Vezzra:
> For the Common target, the "link binary with libraries" list was missing the Zlib and the boost date-time and regx libs. I've added them, and reordered the list to make it easier to compare with CMakeLists.

Considering that these missing libraries never caused any issues it's probably not necessary to link them to the Common target, but if it's done that way in the cmake scripts, I guess it's fine nonetheless.

Vezzra (approved):
Did only some very brief testing, but looks ok and builds and runs fine.

memem359:
@Vezzra 
I should have added the comment that I don't know how many of these were necessary, but it is less confusing to make it consistent with the CMakeLists.txt settings.

PR #1579, opened 2017-05-19
add blockade description to pedia

TheSilentOne1:
This adds a description for the blockade mechanic to pedia > game concepts.
Maybe "system domination", or "system (space) superiority", or "system space control" would be a more suitable name instead of "blockade"?
(Note: interruption  of ground troop regeneration is not correct as of now.)


default/stringtables/en.txt
```diff
@@ -4628,6 +4628,14 @@ Species Traits
 SPECIES_TRAITS_TEXT
 Non-exhaustive list of Species Traits:
 
+SYSTEM_SUPERIORITY_TITLE
+System Superiority
+
+SYSTEM_SUPERIORITY_TEXT
+'''When an armed fleet set to aggressive mode arrives at a system unoccupied by hostile fleets, it establishes system superiority. Once it has been established, supply is interrupted for hostile empires, industry of hostile planets can only be used on the planet it is generated on, and ground troop regeneration is stopped. Also, all starlane entries in the occupied system become guarded by the blockading force, so if hostile forces arrive, they can only exit by the starlane entry by which they arrived.
+If two hostile fleets arrive at an unoccupied system at the same time, no system superiority is established until only one side remains and the other one has either been destroyed or retreated. Until then, both empires will maintain supply for their empire.
```
geoffthemedio: What does "maintain supply" mean?

Empires maintain X for their empire could be rephrased.

TheSilentOne1: How about "Until then, supply will not be blocked for both empires"?

Dilvish-fo: I think "blockade" is a well known concept with a much more narrow and applicable meaning than the rather vague "system superiority".  I could much more readily consider the term "system control" rather than "system superiority", but still the term "blockade" seems much better to me.  Can you describe any particular problem you see with using the term "blockade"?

Also, I think the clause 
```industry of hostile planets can only be used on the planet it is generated on``` 
would be a little better to refer to the hostile planets specifically in the blockaded system, and maybe with a little more adjustment, like 
```industry generated on hostile planets within the blockaded system can only be used on the respective planet.```

Comment (Dilvish-fo) [reply default/stringtables/en.txt]: Also, the clause 
```so if hostile forces arrive, they can only exit by the starlane entry by which they arrived.```
is a bit incomplete/vague regarding what happens if multiple hostile fleets arrive along different starlanes, what if one gets destroyed in the meantime, etc.  I am pretty sure that at least one of those descriptions I linked had a few sentences about that.

Furthermore, the phrase 
```no system superiority is established``` 
should be clarified that it is with respect to each other -- they each can establish blockades against all other empires even while battling against each other.

Hmm, and the part about space monsters seems slightly incomplete also.    I really think that pretty much all of the information [in this explanation](http://www.freeorion.org/forum/viewtopic.php?p=76985#p76985) should be present, unless you think some of it has been superseded now?

Comment (geoffthemedio) [reply default/stringtables/en.txt]: All of these discussions seem like they are ignoring the way supply propagation was modified so that atmost one empire can supply each system (shared and connected through alliances excepted). So, I dont understand what the references to supply inerruption to (seemingly) multiple empires means...

Comment (Dilvish-fo) [reply default/stringtables/en.txt]: Right, yes, the previous supply mechanics are now superseded in that respect and so the clause
```both empires will maintain supply for their empire.```

should be changed to something like 
```neither such empire will establish a blockade against each other, although they will both establish blockades against all other empires.```

Also we should probably have at least have some acknowledgement of supply competition, at least a link to an article about supply in general which includes supply competition. 

I have not examined the supply competition code but my experience ingame leads me to believe that supply competition and blockades are two separate potential blocks, and the results I see seem to indicate to me that the blockade effect is taken into consideration first, before supply competition is assessed.  So it seems to me the rest of the supply blockade discussion here is still OK, and that it would be most clear to have a separate supply article which would talk about supply meters and how supply propagation can be blocked either by blockades or by supply competition.  And then either here or in the general article it should mention how a blockade can allow supply to propagate across an opposing supply source.  That seems very closely related to the supply competition idea, so I would probably leave it for the general supply article.

**Comment by TheSilentOne1** [comment]
@Dilvish-fo, @geoffthemedio: I've applied your suggestions, thanks.

> Can you describe any particular problem you see with using the term "blockade"?

For me, "blockade" conveys mostly that the planets within the system are shut off from the outside. My impression is that with starlane entries being blocked as well a term like "system control" or "superiority" may be more fitting. "System control" sounds to me like the ships spread out through the system. But I'm not a native speaker ofc.

> All of these discussions seem like they are ignoring the way supply propagation was modified so that atmost one empire can supply each system [...]

I've modified the description so the effect of a blockade/system control on supply propagation are more clear. It does seem there is no pedia article on the supply propagation mechanism yet (at least the supply pedia entry does not mention it), so that's something I'll put on my to-do list.

> Hmm, and the part about space monsters seems slightly incomplete also. I really think that pretty much all of the information in this explanation should be present, unless you think some of it has been superseded now?

Not sure what you're referring to, I think I covered the two different aspects (blockade breaks after one turn, blockade monster if arriving on the same turn)?

**Comment by TheSilentOne1** [comment]
(I'll change the system control / superiority bit if we decide to stick with the current name.)

**Comment by Dilvish-fo** [comment]
> For me, "blockade" conveys mostly that the planets within the system are shut off from the outside. My impression is that with starlane entries being blocked as well a term like "system control" or "superiority" may be more fitting. "System control" sounds to me like the ships spread out through the system. But I'm not a native speaker ofc.

Well, I am a native speaker and I don't think that "system superiority" or "system control" offers anything over "blockade".  When Iran threatens to blockade the Strait of Hormuz, it matters because that would affect ships trying to move through it to other destinations.  "Blockade" is a single word so more convenient, plus we have been using it for many many years, so I think it should not be changed without a pretty good reason.

```If all hostile fleets are destroyed before new fleets arrive, the previously opened exits will become closed again.```

I think that rather than just "hostile fleets" this needs to refer to "armed aggressive fleets owned by the blockaded empire", and I would drop the "hostile", because the opening and closing of starlanes we are talking about is with regards to the owner of the mentioned fleets, not with regards to the empire making the blockade.  Also, now that I suggest the phrase "owned by the blockaded empire" I find myself wondering about allies-- alliances were in a very nascent form when I had updated the blockade system, and I don't think I touched on them.  Taking a look at the Fleet::BlockadedAtSystem() and the Empire::UpdateSupplyUnobstructedSystems code, I still don't see any mention of alliances, so it might be helpful to explicitly mention that Allied fleets do not have a direct impact on supply blocks and starlane availability (though they could have an indirect impact by helping kill the blockader).

> Not sure what you're referring to, I think I covered the two different aspects (blockade breaks after one turn, blockade monster if arriving on the same turn)?

I guess it's a few smallish things.  My language had referred to "stays fighting/surviving" rather than just to "stays", and that distinction does matter, a group of unarmed troopships won't be able to force their way past a blockading monster just by surviving a turn.  Also, your phrase "will be able to establish" a blockade is actually slightly stronger than my phrasing "can still blockade" and I think the latter is more compatible with the implicit uncertainty about surviving the initial combat (but I'll acknowledged that is a very nuanced distinction not necessarily made by everyone).  Also, the clause "the monster's control of the system will be broken" seems to me to raise the question of whether it is broken with regards to all empires (it is not), and just referring to the blockade lifting with regards to the empire fighting it seems more clear to me.

**Comment by Dilvish-fo** [comment]
> , and ground troop regeneration is stopped

Oh, also, although we are perhaps still having some discussion in your other PR about the troop regen, it seems unlikely it would wind up being actually tied to the establishment of a blockade, which is what this language suggests.  I would propose that this blockade description refer the reader with a link to some other article about troop regen.

"Establishment of a blockade is not necessarily required in order to stop enemy troop regeneration; for more information please refer to <blah>."

**Comment by TheSilentOne1** [comment]
> Well, I am a native speaker and I don't think that "system superiority" or "system control" offers anything over "blockade".

How about "system blockade" instead of just "blockade" then? That would make clear it's not just the planets that are affected.

> Oh, also, although we are perhaps still having some discussion in your other PR about the troop regen, it seems unlikely it would wind up being actually tied to the establishment of a blockade, which is what this language suggests.

I think the troop regen interruption is definitely something that needs to be revised (and my impression is  there's others that share my view). I'd like to encourage you to think about it again as well, but we should better have this discussion on the forums. Anyway, until there's consensus on how to handle it, I'll remove the troop regen line.

> My language had referred to "stays fighting/surviving" rather than just to "stays", and that distinction does matter, a group of unarmed troopships won't be able to force their way past a blockading monster just by surviving a turn.

I don't think that's correct. I can sneak a colony ship past a sentry because it survives one turn and then continues moving.

Let me make clear that I do realize now that the blockade mechanism is something you put a lot of work and love into, and I find it absolutely great as it is. But I think the UI doesn't give enough information on what's happening, and that's something I'd like to improve.

**Comment by MatGB** [comment]
The discussion triggered my historian/international law training.

A blockade is/can be a legal term: the Union declared a blockade of CSA ports during the civil war and it was the eventual successful implementation of that that pushed GB closer to the Union side. A blockade in international law/military history is a military attempt to prevent access to a port, territory or country. I think in our terms you blockade planets within a system.

I think the term we need for blocking travel through a system/past a chokepoint is Interdiction: there's always going to be a crossover between the two and normally if you're interdicting an enemy system you're also blockading the planets within that system, but given the complications and confusions I think we need to use two different terms: note I'm not 100% sold interdiction is correct but it's the closest I've come up with.

(related: we use the term bombard sometimes to mean 'using a specifically scripted bombardment weapon' and sometimes 'shooting at a planet with a normal weapon', one of them needs a different name as well)

**Comment by Dilvish-fo** [comment]
> I don't think that's correct. I can sneak a colony ship past a sentry because it survives one turn and then continues moving.

Ah, ok, perhaps I had written it up wrong originally, or perhaps that was an aspect that changed with the monster blockade tweak that I think went in sometime this past year; it does sound familiar.

> I think the troop regen interruption is definitely something that needs to be revised (and my impression is there's others that share my view). I'd like to encourage you to think about it again as well

I'm fine with the proposal in that other discussion to shift it to requiring that the armed aggressive enemy fleets survive, but that still would not necessary establish a blockade.  I don't think there was any significant push there to tie regen only to blockades,  so I think it's best dealt with on its own page.

@MatGB 
> I think the term we need for blocking travel through a system/past a chokepoint is Interdiction: there's always going to be a crossover between the two and normally if you're interdicting an enemy system you're also blockading the planets within that system, but given the complications and confusions I think we need to use two different terms: note I'm not 100% sold interdiction is correct but it's the closest I've come up with.

It's certainly fun to hear about the more specific military term, and it could be nice to include the use of the term interdiction as part of our description of lane blocking caused by blockades, but since (apparently unlike real life) we do not ever have interdiction without a blockade I don't think we need any special emphasis about the interdiction term or independent use of it.  The normal use of the term blockade does commonly include the interdiction concept (as in my example re the Strait of Hormuz).

**Comment by MatGB** [comment]
>  since (apparently unlike real life) we do not ever have interdiction without a blockade

Except we do: to blockade a planet you need to be in the same system as that planet, to interdict a planet/system/supply line you merely need to cut supply to it. Blockades cannot happen without interdiction (I think), but interdiction can happen without blockades.

When a ship is prevented from passing through a system to go elsewhere it is interdicted, when a planet is cut off from all others, including others in system it is blockaded.

Oh, it was always the case that if you survived against a monster you could then move on, there was a very brief period recently where it changed due to a tweak to some code but it was fixed very quickly.

**Comment by Dilvish-fo** [comment]
> How about "system blockade" instead of just "blockade" then? That would make clear it's not just the planets that are affected.

I don't have any objection to the term "system blockade" being used at least some (and I imagine I have likely used it some myself), but I expect that in many cases the simpler "blockade" term would be best; a clause like "if a planet is blockaded" is preferable over "if a planet is system-blockaded".  Our blockades may have multiple different effects, but it's not like we have multiple kinds of blockades that need to be distinguished (well, unless you maybe want to consider a monster blockade a different kind, but still, then the distinction would be monster blockade versus empire blockade).

**Comment by dbenage-cx** [comment]
@TheSilentOne1 Would like to respond to some points here. Did you have an existing thread in mind or a new one?

**Comment by TheSilentOne1** [comment]
@dbenage-cx: I've opened a new thread here: http://www.freeorion.org/forum/viewtopic.php?f=6&t=10566

**Comment by TheSilentOne1** [comment]
I've updated the description and from my side it's GTG. If a native speaker could have a look that'd be good. When all corrections have been applied I'll squash & merge.

File: default/stringtables/en.txt [hunk 2]
```diff
@@ -4628,14 +4628,15 @@ Species Traits
...
-If fleets from two or more hostile empires arrive at an unoccupied system at the same time, neither such empire will establish system control over each other, although they will both establish control against all other empires.
-Space monsters can also establish system control. However, if an empire fleet arrives and stays at the system for one full turn, the monster's control of the system will be broken. If a space monster and an empire fleet arrive at the same time, the empire fleet will be able to establish system control despite the monsters presence.'''
+SYSTEM_BLOCKADE_TITLE
+System Blockade
+
+SYSTEM_BLOCKADE_TEXT
+'''When an armed fleet set to aggressive mode arrives at a system currently unoccupied by hostile fleets, it establishes a blockade. If this happens, [[metertype METER_SUPPLY]] propagation is interrupted for all other empires, and industry generated on hostile planets within the system can only be used on the respective planet. If hostile planets come under attack, their meter growth (e. g. ground troops) will be suspended.
```
geoffthemedio: all other empires? does it matter if they are at peace or allied? what constitutes a hostile fleet or planet? what is a respective planet? what does it mean to come under attack? That sentence seems like it needs an "Additionally, " at the start.

default/stringtables/en.txt
```diff
@@ -4628,14 +4628,15 @@ Species Traits
...
+Also, all starlane entries in the blockaded system become guarded by the blockading force, so if hostile forces enter the system, they can exit only by the starlane entry by which they arrived. However, if more fleets from the blockaded empire arrive by different starlane entries than the previous one, these entries will become available as an exit as well. If at one point all armed aggressive fleets owned by the blockaded empire are destroyed before new fleets arrive, the previously opened exits will become closed again.
```
geoffthemedio: "Also," is awkward at the start of a new paragraph. "become guarded" is a bit awkward... maybe "are" or "will be"? "by which" sounds weird.. maybe "on which" or "through which" when referring to an entry... ("by which" seems better if referring to the starlane, rather than the entry, which is also an alternative. "If at one point" -> "If". What if the armed aggressive fleets leave the system or go passive; does that count?

default/stringtables/en.txt
```diff
@@ -4628,14 +4628,15 @@ Species Traits
...
+Allied fleets do not have a direct impact on supply blocks and starlane availability apart from helping with eliminating the blockading empire's forces.
+If fleets from two or more hostile empires arrive at an unblockaded system at the same time, neither such empire will establish a system blockade over each other, although they will establish a blockade against all other hostile empires that arrive at a later time.
+Space monsters can also establish blockades. However, if an empire fleet becomes blockaded by a space monster, it will be able to evade the blockade after staying at the system for one full turn and leave the system by any starlane exit. Also, if a space monster and an empire fleet arrive at the same time, the empire fleet will be able to act first and establish a blockade against the monster.'''
```
geoffthemedio: Empire fleets can leave after one turn in a system with monsters, but the supply propagation restrictions remain as long as the monster is present?

TheSilentOne1:
@geoffthemedio: I've reworked the description and adressed your points. Probably still needs some fine-tuning, suggestions are welcome.

default/stringtables/en.txt
```diff
@@ -4628,6 +4628,16 @@ Species Traits
 SPECIES_TRAITS_TEXT
 Non-exhaustive list of Species Traits:
 
+SYSTEM_BLOCKADE_TITLE
+System Blockade
+
+SYSTEM_BLOCKADE_TEXT
+'''When an armed fleet set to aggressive mode arrives at a system, it will establish a system blockade against all empires it is at war with and that currently do not blockade the system. When a blockade is established, [[metertype METER_SUPPLY]] propagation is interrupted for the blockaded empire. Also, if the blockaded empire posesses colonies within the system, the production generated by these colonies can only be used on the planet it is generated on. Additionally, if the blockaded planets are attacked by the incoming fleet, their meter growth (e. g. ground troop regeneration) will be suspended for one turn.
```
MatGB: This last bit isn't accurate, currently if any combat occurs then some meters, specifically troops, don't grow, even if the planet isn't attacked (you can do this with a stealthed carrier), but others do, defence definitely does as planetary defences can whittle down a visible blockading fleet. Shields are gated so if the planet has no infrastructure they don't go up, but I might be misreading the code as [I don't see how it ever goes up](https://github.com/freeorion/freeorion/blob/master/default/scripting/techs/defense/BarrierShield.focs.txt#L19)—I'm terrible at embedded max/min statements, I don't see any restrictions on combat but I also don't see how it goes up for outposts.

Maybe try
> Additionally, as blockading normally initiates a combat in the system, effects that only work if no combat has occurred such as Orbital Drydocks and Troop meter growth are blocked for that turn.

default/stringtables/en.txt
```diff
@@ -4628,6 +4628,16 @@ Species Traits
...
+After a system blockade has been set up, all starlane entries in the blockaded system will be guarded by the blockading force, so if hostile forces enter the system, they can exit only through the starlane entry from which they arrived. However, if more fleets from the blockaded empire arrive by different starlane entries than the first one, these entries will become available as an exit as well. If all armed fleets owned by the blockaded empire are destroyed or leave the system before new fleets from this empire arrive, the previously opened exits will become closed again.
+Allied fleets do not have a direct impact on supply blocks and starlane availability apart from helping with eliminating the blockading empire's forces.
+If fleets from two or more hostile empires arrive at an unblockaded system at the same time, neither such empire will establish a system blockade over each other, although they will establish a blockade against all other hostile empires that arrive at a later time.
```
MatGB: I don't think this is correct, supply is blocked for one side or the other, I think it stays as it was unless superiority in system is achieved but am not sure.

MatGB:
Native speaker review, no problems with the language used but I think some bits are wrong, but I may not be correct on this.

default/stringtables/en.txt
TheSilentOne1: If the fleets arrive at the same time, there will be no system blockade, so other factors will decide who can propagate supply into the system (supply strength, colonies or outpost present, sum of supply sources, distance to nearest supply source). So that means who had supply before will keep it. So I think the description should be correct?

default/stringtables/en.txt
TheSilentOne1: Thanks for poiting that out. I'll change the description accordingly.

TheSilentOne1:
I checked nearly all the described situations in-game (which is not as easy as one may think, battles are often over really quick), so I _think_ they should be correct. Is there a specific part (apart from what you already mentioned) that doesn't seem to be correct to you?

TheSilentOne1:
Changed the description with regard to combat effects. If there shouldn't be any new comments I'll merge the PR tomorrow.

PR #1582, opened 2017-05-22
Prevent unowned monsters from attacking adequately stealthed ships

dbenage-cx:
Unowned monsters are currently able to attack stealthed ships.

Test save with empire ship(65 stealth) vs unowned juggernaut (system: Pawn ⍺)
and unowned juggernaut vs empire ship(65 stealth) and owned monster(5 stealth) (system: Thrax β)
[save-20170522_124712.zip](https://github.com/freeorion/freeorion/files/1019746/save-20170522_124712.zip)

### DIFF
--- a/combat/CombatSystem.cpp
+++ b/combat/CombatSystem.cpp
@@ -1090,10 +1090,9 @@ namespace {
             for (int object_id : valid_target_object_ids) {
                 std::shared_ptr<const UniverseObject> obj = combat_info.objects.Object(object_id);
                 for (int attacking_empire_id : combat_info.empire_ids) {
-                    if (attacking_empire_id == ALL_EMPIRES
-                        && ObjectAttackableByMonsters(obj, monster_detection))
-                    {
-                        empire_infos[ALL_EMPIRES].target_ids.insert(object_id);
+                    if (attacking_empire_id == ALL_EMPIRES) {
+                        if (ObjectAttackableByMonsters(obj, monster_detection))
+                            empire_infos[ALL_EMPIRES].target_ids.insert(object_id);
                     } else if (ObjectAttackableByEmpire(obj, attacking_empire_id)) {
                         empire_infos[attacking_empire_id].target_ids.insert(object_id);
                     }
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
From a quick look at the code, without testing, it looks like a good fix, but the commit message seems misleading to me.  From the message it sounds like you are making stealthed ships absolutely unattackable by monsters.  I would suggest changing the commit message to
``` Fix unowned monster detection check so that they are not always able to attack stealthed targets regardless of the monster's detection.```

MatGB:
Loaded the savegame and I can't see that the monster has ever attacked the stealthed ship, I get "Unknown empire cannot target Shelyak" each turn, which is how it should be, I've never noticed a monster attacking a ship it wasn't meant to be able to see and we did a lot of work rebalancing monstrous detection last year so that you could stealth past some monsters, I've definitely never had a guardian attack a stealthed ship and I'm fairly sure I'd have noticed a moving one.

Also, as I can't read the code properly, Dilvish is correct that the commit message is misleading, when I got the initial email I was very worried as some monsters (specifically the experimentor critters) need to be able to target high stealth ships and are balanced accordingly (the numbers aren't quite right but they're in the right ballpark).

dbenage-cx:
@Dilvish-fo TYVM, as the suggested line is a bit over the recommended 50, I've reworded the summary and provided a more detailed explanation.  Please let me know if it could be improved further.

@MatGB For me, this has been exceedingly common recently, not limited to this one case. Here are some logs portions of first bout to compare: [master](https://gist.github.com/dbenage-cx/adfccb201e6652d78d88bcc988ae6177) -  [PR](https://gist.github.com/dbenage-cx/1acad4a4a6fb47614ba885e5986e9eac)
With master (same save game, empire ship has 65 stealth):
screenshot_20170522_123025

geoffthemedio:
looks reasonable...

Vezzra:
@dbenage-cx, this PR has been proposed and accepted for inclusion into the 0.4.7.1 bugfix release, but when I try to cherry pick b95e34b5f35ec33f93ca98bede8b1b59fb1e2643 into the release branch, I get conflicts I can't resolve (because I can't figure out what's wrong).

Can you do the cherry picking? The release branch has already been (re)created, so you can go ahead.

Vezzra:
@dbenage-cx, nevermind, upon further investigation I realized this PR fixes a bug introduced *post* 0.4.7, so it does not apply for 0.4.7.1 anyway.

PR #1584, opened 2017-05-24
AI savegame compatibility framework

Morlic-fo:
Basic principle:

1. The AIstate class has a version number.
2. Whenever the AIstate class is modified in a way that would break savegame compatibility
  i.  The version number must be increased by 1
  ii. The convert_to_version() function must implement the required changes to get a playable AIstate instance compatible with the new AIstate version.
3. If the old state is several versions old, then convert it iteratively to the next version until the current version is reached (i.e. 1->2->3->4)
4. If conversion fails, inform the player about the problem and try to play with a fresh AIstate instance.

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -74,7 +74,12 @@ def convert_to_version(state, version):
 
     if version == 0:
         pass  # only version number added
-    # elif version == 1:
+    elif version == 1:
+        try:
+            state['_aggression'] = state['character'].get_trait(Aggression).key
+        except Exception as e:
+            raise ConversionError("Error when converting to compatibility version 1: "
+                                   "Can't find aggression in character module. Exception thrown was: " + e.message)
     #   state["some_new_member"] = some_default_value
     #   del state["some_removed_member"]
     #   state["list_changed_to_set"] = set(state["list_changed_to_set"])
@@ -93,7 +98,7 @@ class AIstate(object):
     is playable with this AIstate version, i.e. new members must be added
     and outdated members must be modified and / or deleted.
     """
-    version = 0
+    version = 1
 
     def __init__(self, aggression):
         # Do not allow to create AIstate instances with an invalid version number.
@@ -109,6 +114,8 @@ class AIstate(object):
         self.uid = self.generate_uid(first=True)
         # unique ids for turns.  {turn: uid}
         self.turn_uids = {}
+
+        self._aggression = aggression
 
         # 'global' (?) variables
         self.colonisablePlanetIDs = odict()
@@ -158,12 +165,15 @@ class AIstate(object):
                 convert_to_version(state, v+1)
             self.__dict__ = state
         except ConversionError:
-            try:
-                aggression = state['character'].get_trait(Aggression).key
-            except Exception as e:
-                print >> sys.stderr, "Could not find the aggression level of the AI, defaulting to typical."
-                print >> sys.stderr, e
-                aggression = fo.aggression.typical
+            if '_aggression' in state:
+                aggression = state['_aggression']
+            else:
+                try:
+                    aggression = state['character'].get_trait(Aggression).key
+                except Exception as e:
+                    print >> sys.stderr, "Could not find the aggression level of the AI, defaulting to typical."
+                    print >> sys.stderr, e
+                    aggression = fo.aggression.typical
             self.__init__(aggression)
 
     def generate_uid(self, first=False):
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -14,7 +14,7 @@ from EnumsAI import MissionType, ShipRoleType
 import CombatRatingsAI
 import MilitaryAI
 import PlanetUtilsAI
-from freeorion_tools import dict_from_map
+from freeorion_tools import dict_from_map, print_error
 from universe_object import System
 from AIDependencies import INVALID_ID
 from character.character_module import create_character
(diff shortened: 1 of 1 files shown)


default/python/AI/AIstate.py
```diff
@@ -43,10 +43,62 @@
...
+    :param state: savegame state, modified in function
+    :type state: dict
+    :param version: Version to convert to
+    :type version: int
+    """
+    print "Trying to convert savegame state to version %d..." % version
+    current_version = state.get("version", -1)
+    print "  Current version: %d" % current_version
+    if current_version == version:
+        raise ConversionError("Can't convert AI Savegame to the same version.")
```
Dilvish-fo: in these three ConversionErrors I think it would be better to refer more specifically to "compatibility version" rather than just "version"

default/python/AI/AIstate.py
```diff
@@ -43,10 +43,62 @@
...
+    version number must be increased by 1 and the convert_to_version()
+    function must be updated so a saved state from the previous version
+    is playable with this AIstate version, i.e. new members must be added
+    and outdated members must be modified and / or deleted.
+    """
+    version = 0
+
     def __init__(self, aggression):
+        # Do not allow to create AIstate instances with an invalid version number.
+        assert(hasattr(AIstate, 'version') and isinstance(AIstate.version, int) and AIstate.version >= 0)
```
Dilvish-fo: So far, the only way I have been able to think of this assertion failing is if someone were to customize their AI code but mistakenly set the version to something like "3b" or whatnot.  Testing that scenario out, I just get the rather unhelpful log error and corresponding chat message 
```Chat Message to human: AI script Error in "startNewGame"```
And then when trying to advance to turn 2 the game would not advance (server seems stuck waiting for AI orders that will never arrive).

So, for that scenario, and any other scenarios you are trying to cover here, we need a smoother handling.   I would be inclined to handle all such scenarios by raising a ConversionError from the convert_to_version() function rather than any assertion here.

File: default/python/AI/AIstate.py [hunk 3]
```diff
@@ -95,6 +147,14 @@ def __init__(self, aggression):
         self.empire_standard_enemy_rating = 0  # TODO: track on a per-empire basis
         self.character = create_character(aggression, self.empireID)
 
+    def __setstate__(self, state):
+        try:
+            for v in range(state.get("version", -1), AIstate.version):
+                convert_to_version(state, v+1)
+            self.__dict__ = state
+        except ConversionError:
+            self.__init__(aggression=fo.aggression.typical)
```
Dilvish-fo: An idea that just struck me here is that we could still double check if the provided state had an aggression attribute and then use that rather than necessarily defaulting to ```aggression.typical```.  That's probably actually the single most important piece of info from the savestate, and one of the least likely to be problematic.

Dilvish-fo (changes requested):
It does look like a better framework than merely relying on __setstate__().    I have some requested changes.

Morlic-fo:
I worked in your suggestions but am not too happy with the fact that the AI aggression level is currently not directly stored in the AIstate but only in the character. 

Relying on some complex object which is possibly subject to future changes seems unsafe, so I prefer to store it directly in the AIstate again...

default/python/AI/AIstate.py
```diff
@@ -158,12 +165,15 @@ def __setstate__(self, state):
...
-            try:
-                aggression = state['character'].get_trait(Aggression).key
-            except Exception as e:
-                print >> sys.stderr, "Could not find the aggression level of the AI, defaulting to typical."
-                print >> sys.stderr, e
-                aggression = fo.aggression.typical
+            if '_aggression' in state:
+                aggression = state['_aggression']
+            else:
+                try:
```
genestack-solomatin: I don't read total thread, but this looks strange for me.  We have handler for difference and have some separate code to handle difference.

default/python/AI/AIstate.py
```diff
@@ -74,7 +74,12 @@ def convert_to_version(state, version):
 
     if version == 0:
```
genestack-solomatin: May be make some code that find all function with special name like 'convert_1_2',  then detect current code state and current save state and call all functions one by one.   

Maybe move this  functions to separate module, so they will not be present in main module. And it will be easy to disable code check for them.

Morlic-fo: > May be make some code that find all function with special name like 'convert_1_2', then detect current code state and current save state and call all functions one by one.

Any particular advantage you see in that approach?
Sure, the if-chain used here may become rather long but it is sequential, self-documenting and logically coherent. It is unambigious and trivial to maneuver to the correct part of the code.
Splitting the body into multiple bodies each called only once seems not to add any kind of real code abstraction.

Particularly long or complex conversion-steps may still be extracted into functions if it seems necessary.

> Maybe move this functions to separate module, so they will not be present in main module. And it will be easy to disable code check for them.

Probably a good idea.

default/python/AI/AIstate.py
Morlic-fo: If an ConversionError was caught, then the savegame could not be converted (e.g. ill-defined conversion function, invalid version number, invalid pickle-data...).

Considering we couldn't "load" the AIstate, we need to create a new one. That requires an aggression level which we try to extract from the saved state. If that fails, we then default to the typical aggression level.

Probably needs an explanatory comment. Any other issues with the approach?

genestack-solomatin: I have joined two part of code.   Looks like you do almost same thing twice. 

```python
elif version == 1:
       try:
           state['_aggression'] = state['character'].get_trait(Aggression).key
       except Exception as e:
           raise ConversionError
except:
   try:
        aggression = state['character'].get_trait(Aggression).key
        except Exception
```

default/python/AI/AIstate.py
genestack-solomatin: > Any particular advantage you see in that approach?

It is common practice.   You write updates as separate parts: `a->b`, `b->c`, `c->d`

if you have version `a` you need apply `a->b`, `b->c`, `c->d`
If you have version `b` you need to apply `b->c`, `c->d`

As I see you want to write this by hand.
```python
if version == a:
    a_b()
    b_c()
    c_d()
elif version == b:
    b_c()
    c_d()
```

Morlic-fo: > As I see you want to write this by hand.

No. That was never the intention.

The two possible choices are:
1. A fall-through pseudo-"switch"-statement:
```
if state.version == a:
    a_b(state)  # state.version is now b
if state.version == b:  # also is true if initial version was a
    b_c(state)  # state.version is now c
```

2. elif-statements as in this PR and iteratively convert the state by calling the function with increasing version number in a for-loop, i.e. upgrade only by 1 version by function call (cleaner than the fall-through switch if additional error checks are implemented that should be done between each step).

In any case, the ```b_c()``` part is written only once here:
```
(el)if version == b():
    b_c()
```

default/python/AI/AIstate.py
Morlic-fo: Yes. And your point is ... ?

The idea is to cherry pick that very piece of information of the saved state if the provided state was not usable for any other reasons.

I load an AI version "738" that has no ```_aggression``` attribute anymore because it was removed again in version "435". It still does have the aggression stored as in
```        aggression = state['character'].get_trait(Aggression).key```
My current AIstate version is 740. The conversion from 738->739 throws a ```ConversionError``` for some reason. Now I can not use the saved state and fall back to creating a new AIstate instance. That requires the aggression level.
So I catch the excepion and use
```        aggression = state['character'].get_trait(Aggression).key```

Because I started with a version way beyond version 1, your first code sample was and would never be called. They are completely unrelated.

default/python/AI/AIstate.py
Cjkjvfnby: So this if-elif is just bad example.  Please add a docstring how to manage updates.

Dilvish-fo (approved):

Dilvish-fo:
This is looking fine to me, and CJ could always follow up more later on if he feels the need for that.  I probably will want to be following up on the general topic of serializing/unserializing the AIstate, but I don't want that to delay this since some other pending work is also dependent on this.  Morlic, I'll leave it to you in case you had any further tweaks, but I think this should go ahead and get merged.

PR #1586, opened 2017-05-26
Change ship design parser ignore UUID and ShipDesignManifest.focs.txt.

LGM-Doyle:
This PR changes the ship design parser to ignore a file called "ShipDesignManifest.focs.txt" and to ignore any UUIDs in ship designs.  

I am posting a PR which will allow editing of saved designs and arbitrary ordering of ship designs and monsters.  It uses "ShipDesignManifest.focs.txt" to provide an ordered list of UUIDs to store that ordering off-line.

Applying this PR before people start testing the other PR will mean that when the UUIDs are added to their saved designs, no error messages will be generated when they switch to another branch.

Issue #1145, opened 2016-11-27
Allow editing and deletion of saved ship designs within app


MatGB:
### Description
The saved designs feature is a godsend that is very useful, when I switch to a new machine or do an OS reinstall it always throws me when they're not there.

However, sometimes I mess a design up, and sometimes we introduce new features (eg: Fighters) or parts (eg: Flak Cannon) that require a rethink of designs. And then there are the inevitable typos.

It would be *really* helpful if the client allowed for editing of saved designs directly within it in a user friendly way.

LGM-Doyle:
I added this comment to #884, when I meant to comment here.

@MatGB do the newly merged `DesignWnd`changes to saved ship designs satisfy this feature request enough to close it and start opening bug issues and/or new feature requests based on the current state of the game?

MatGB:
Yup, this one is now resolved to my satisfaction, many thanks: it's been a particular godsend given I'm working on the engine parts fix at the moment and that means switching between radically different designs for testing, etc.

Issue #886, opened 2016-08-19
Producible Items list should be sortable via scripts and UI


MatGB:
### Description

It is now possible to reorder the ship designs list and this carries through to the producible items list. If something has not been reordered the list defaults to alphabetical by key name with buildings before ships, ships list by keyname then design time. It would be good if some method of sorting was available both to the player and for scripting, currently to move something within the list a scripter has to change its keyname, update all the stringtables and amend AI code.
### Expected Result

Some method in which preunlocked and tech unlocked items can be given a priority number or groupname so that they all appear together, allowing—especially for predesigned ships—scripters to group things by type within the display to improve the UI and improve player workflow.

If players can also drag/drop to rearrange the list in some way this would be an additional benefit.

MatGB:
Specifically, I've got several parts into a reordering of the premade ships several times now and gotten sidelined or delayed, it's such a longwinded task for a very simple set of changes, as we can move ships up and down within the UI it should be possible to script a display order for them.

geoffthemedio:
Some overlap with, but better written than, https://github.com/freeorion/freeorion/issues/784

LGM-Doyle:
I've moved this comment from #784 where it was accidentally added.

@MatGB do the changes in the `DesignWnd` satisfy this feature request?  In particular, the procedure that I have duplicated below to edit and/or re-order default ship designs and monsters.

And secondly, is there a good place on the website to place these instructions where scripters can use them as a reference?


------------------ Start of Instructions -------------------
****Editing default ship designs and monsters****
1. Move or backup the `shipdesigns` directory in your freeorion data directory.
2. Copy the `default/scripting/ship_designs` or `default/scripting/monsters` directory into your
data directory and replace your `shipdesigns` directory.  
3. Start a new game.
4. Your saved ship designs will now be the default ship designs or monsters.  Edit the ships.  
    i. Add new saved designs.  
    ii. Drag and drop to change ordering.
    iii. If you use a stringtable index for **both** the ship name **and** description then it will save the stringtable index and use the translated text for display.  You can check this in the pedia window.  When you add/remove parts the name and description in the pedia window will be pulled from the stringtable.
    iv. Note: You can not start a ship with a monster hull, but you can add/remove components to monster designs in the saved design directory.  
5. Exit the game.
6. Copy the `shipdesigns` directory from your data directory back into the `default/scripting/ship_designs` or `default/scripting/monsters` directory.

------------------ End of Instructions -------------------

[...]

PR #1587, opened 2017-05-26
Improve design window editing of ship designs.

LGM-Doyle:
~~Warning: Testing this PR will add UUIDs to your saved designs. This will
    cause parsing errors in other branches until/unless PR #1586 is merged.  Back
    them up.  These UUIDs will not be removed when you switch back to the master
    branch because they are not part of the repository.  PR #1586 makes the
    ShipDesign parser always ignore the UUID tags.  I will strikeout this
    warning if it is merged.~~ PR #1588 has been merged and ShipDesign UUIDs 
    are part of master, so the previous warning is not relevant.

.
.
.

I will cross [post](http://freeorion.org/forum/viewtopic.php?f=6&t=10572) the detailed description of what this PR does to the forum in a moment.
People can choose which issues should be discussed in which venue.

This PR implements the feature requested in issue #1145.

It allows saved designs to be manipulated, saved, deleted, drag dropped, added
to the finished designs individually or as a group at any time or automatically
at game start.

It allows the finished designs to be made obsolete or restored over the course
of the game.  Obsolete designs don't appear in the production window, but still
appear in the design window.

This PR shows the complete work. I will break it into smaller PRs.

#### Key Issues

The key technical issue directly related to this PR are about how and where
should the ordering information for saved and finished ship designs be stored.
The solution to the question of where to store information to allow the player
to unobsolete designs follows naturally from these decisions.

Saved designs are stored on disk and across multiple games. The identifier needs
to be unique across un-coordinated instances of the game.

The design name is overloaded with constraints.  It is used as the human visible
name, the index into the translation table, the sorting index and the default
file name.  Using the design name as the index forces uniqueness constraints on
players.

This PR adds a UUID to each design.  It an be used off-line to identify ship
designs between games without coordination.  The UUID is then used in the
ShipDesignManifest.focs.txt to store the off-line sort order.  The UUID is also
the index into the look up table in game.

Finished design order needs to be available to the DesignWnd and the
BuildDesignator and also stored in the save game file. The server does not need
to know about the UI ship design ordering.

The previous work in this area stored the human client design ordering on the
server in the empire data.  This was expedient, but not the best location.

This PR moves the client side data into the `SaveGameUIData` and creates
`ShipDesignManager` which resides in the `ClientUI` as a strictly data element
to manager the ship designs human client UI data and its serialization. This
provides an interface to the data for both the `BuildDesignator` and the
`DesignWnd`.

Once `ShipDesigManager` was created it is an good location to store an
additional bit to indicate that the player has obsoleted a design, but might
want to restore it.  This is displayed in the GUI by desaturating the ship
design. It is filtered by an additional toggle in the Availabilities filter.

#### Miscellaneous Issues

The PR fixed various other minor issues along the way.

- Fix for the memory leak in ShipDesignParser by converting it to unique_ptr,
  which is already submitted as #1545

- Moves uniqueness checking of names and now UUIDs out of the parser and into
  the ShipDesign code.

- Removes the file dialog for saving ship designs and replaces it with a name
  based on the UUID, in the ship design directory, which can be set in the
  options window.  The disk work to create/edit/reorder/delete saved designs now
  happens automatically.

- Fixes various list box and queue list box issues with dragging and dropping.
  Most importantly, queue list box `AcceptDrops` never actually accepted any
  drops and just triggered all of its clients to regenerate the entire list.
  Consequently, `ListBox::AcceptDrops` having never been used was buggy.

- `OrderSet`has no way to load an order that have already been executed on the
  client, but still needs to be transmitted to the server.

- `BasesListBox` is refactored to replace a long sequence of if(bool identifying
  type of base) then (do base action) with inheritance.  This refactoring also
  results in using a common UI element to display hulls with a name.

- Fixes the bug where saved designs were only added if the design window was opened on the first turn.

- Fixes the bug where if a design was renamed twice, it was added as a new design the second time.

- Updates the pedia to reflect the new UI elements.


dbenage-cx:
From brief test, `SD_COLONY_SHIP` is not added and looks like `SD_TROOP_SHIP` is missing from manifest.

    [warn] server : ShipDesign.cpp:1254 : UUID 08a58b08-0929-496d-84fc-faa91424ca01 is in ship design manifest for a ship design that does not exist.
    [warn] server : ShipDesign.cpp:1278 : Missing ship design 08a58b08-0929-496d-84fc-faa91424ca02 called Troop Ship added to the manifest.

[freeoriond.log.txt line 657-738](https://github.com/freeorion/freeorion/files/1033347/freeoriond.log.txt)

geoffthemedio:
177 commits affecting 124 files? Please separate into smaller parts.

MatGB:
> This PR shows the complete work. I will break it into smaller PRs.

default/scripting/monster_designs/ShipDesignManifest.focs.txt
```diff
@@ -0,0 +1,36 @@
+ShipDesignManifest
```
geoffthemedio: what is the purpose of this list and when and how is it modified?

geoffthemedio:
> > This PR shows the complete work. I will break it into smaller PRs.

ah, sorry; missed that in the lengthy description.

default/scripting/monster_designs/ShipDesignManifest.focs.txt
LGM-Doyle: `ShipDesignManifest,focs,txt` stores the order of the ship designs.  

Players modify it by dragging and dropping ships.

LGM-Doyle:
@dbenage-cx thanks for trying it out.  

I tried to replicate your issue by pulling this branch into a new workspace,but there were no errors in the logs.

Did you try editing the default designs and copying them back into the default directory?  

If so there is a problem with the handling of the string table lookup.   It gets converted into plain text, which is not a problem for players, but will break editing of default designs.   That might have caused your issue.  I will fix that issue.

MatGB:
Having looked it over a bit, on the face of it I **really* want this and will help test when I've finished my current. It not only addresses #1145 but also #886, something I've wanted to be able to do every single time I've gone to tweak the premade ship designs in some way.

So thanks for putting the work in, I've got a few other observations but they're at a tangent so I'll put them on  [the forum thread](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10572)

default/scripting/monster_designs/ShipDesignManifest.focs.txt
geoffthemedio: It should not be stored in the resources directory then. It belongs in the user data directory.

LGM-Doyle: Correct.

However, the copies in the resource directory provide the order for the default designs and the monsters.  The copy in the user data directory provides the order for the user's saved designs.

As is in this PR users can't reorder the default designs or the monsters.

dbenage-cx:
I did not try editing any definitions after pulling, will debug later tonight.

default/scripting/monster_designs/ShipDesignManifest.focs.txt
dbenage-cx: Just to be sure (I did not notice any cases doing so), I assume the `default` manifest is never modified, only the users(data dir)?

dbenage-cx: doh, previous response did not update...

dbenage-cx:
Was not able to determine how to catch a break at `universe/ShipDesign:1254` (never actually breaks).

With another box, the first design parsed is `SD_BASE_DECOY` (comsat), which is not added instead of colony ship.

Playing with logs, it looks like `phrase_parse`(ParseImpl.h:125) does not return from the first design until after parts and hulls are parsed, at which point distance(ParseImpl.h:129) is different.
Skipping that check allows the first design to be added (noting only as an initial workaround):
```diff
diff --git a/parse/ParseImpl.h b/parse/ParseImpl.h
index e58ff55..1208030 100644
--- a/parse/ParseImpl.h
+++ b/parse/ParseImpl.h
@@ -128,7 +128,7 @@ namespace parse { namespace detail {
 
         std::ptrdiff_t distance = std::distance(first, parse::detail::s_end);
 
-        return success && (!distance || (distance == 1 && *first == '\n'));
+        return success; // && (!distance || (distance == 1 && *first == '\n'));
     }
 } }
```

LGM-Doyle:
@dbenage-cx I've added commits to create a logger named `parse` which you can use for trace logging of the parse.  

I added additional debug messages around the code that you indicated to log the erroneous length and the entire file that generated the error.

On my end this generated a lot of noise.  Many of the species files are incompletely parsed, but work correctly in the game.  I think that the parse return code is being ignored and the semantic actions are working anyway.

If you could run this PR with the parse logger at trace threshold, then it will show exactly what Colony ship file is being parsed and it will hopefully be clear the cause of the failure.  

The changes also include the fix to not convert the stringtable tags with `UserString()` to preserve the internationalization in the default designs and the monsters.  I don't think that is the direct cause of your issue.

Thanks.

dbenage-cx:
Thank you for adding parse logger :)
[freeoriond.log.txt](https://github.com/freeorion/freeorion/files/1034305/freeoriond.log.txt)

LGM-Doyle:
Thanks for that log file.  Your previous observations are at the exact point of the problem.

`parse::detail::s_end` which is used to calculate `distance` is static and global.  It implicitly assumes that only one file will be parsed at a time and is over written whenever a new parse is started.  If a parse is potentially recursive like the ship designs, the return value from parse_file is meaningless.  

Until this PR all of the ship and monster design parsing ignored the return value.  I assumed that it was good and added it to the criteria for a good ship design parse.

I've fixed it by ignoring the suspect static.

dbenage-cx:
TY for providing a fix, those designs are now added.

Continuing with testing, an invalid design could be added and produced by manually editing the file.
(e.g.) Save a ship design with hull `SH_BASIC_MEDIUM` (Basic Medium Hull), open the file and change the parts (last entry is an internal slot):
```
    parts = [
        "SR_WEAPON_1_1"
        "SR_WEAPON_1_1"
        "SR_WEAPON_1_1"
    ]
```
That design can be added to empire and successfully produced at a shipyard.


Moving an obsolete design in the Finished tab logs an error (though no noticeable problem within client):
```
[error] client : Order.cpp:1006 : Tried to remove a ShipDesign that the empire wasn't remembering
```

LGM-Doyle:
I've fixed both of those errors.

In a few places the design window was removing obsolete designs from the server empire, which was correctly producing an error, because it didn't already know the design.  That is fixed.

I made it not possible to create an invalid `ShipDesign`.  `ShipDesign` now throws `std::runtime_error` if it is passed invalid parameters at run time. All of the locations that create new `ShipDesigns` now catch that.  The deserialization checks design validity and forces a valid design with the first hull available and no parts.  The only way to create an invalid ShipDesign is to deserialize an invalid design when the hull manager has no hulls.

MatGB:
Playing with this rebased onto master, just tried to 'delete' instead of 'obsolete' a premade and the game crashed, are you meant to be able to delete premades?

LGM-Doyle:
The crash from 'delete' is now fixed.  I made the the same mistake as with 'obsolete'.  I've refactored things a bit to hopefully prevent making the mistake a third time.

There is nothing special about the premades.  You should be able to delete them the same as other designs.

I added a message when the saved designs queue is empty to 'Select an existing Finished design, right click and select "Save Design" to create a saved design.'

Also not new, but not mentioned elsewhere, when you are adding/editing a saved design if you use a stringtable look up name in both the name and the description it will save the stringtable lookup codes, not the translated text.  This feature is not very discoverable, but I didn't want it to break game immersion for players.

@MatGB do you have any other comments about the PR?  Anything, not broken, but not right?  Wonky?

MatGB:
K, recompiled and the crash didn't reoccur, so I assume fixed.

Regarding overall comments, my old designs haven't loaded in at all so I've had to play through a game saving designs, then start a new game to see if it's really doing much, I like it overall but haven't had time to test everything (specifically the premades need to be given an order so I'll learn how to do that).

Change not sure of: new designs appear at the top of the list of ships, not the bottom. Not necessarily a bad thing but it's a change I didn't expect and keep forgetting.

Problem: before the 'saved' tab automatically showed all designs including ones not unlocked, now I have to click 'unavailable' and it then shows them greyed out, but I have to unclick that when I go back into the main empire designs list. I didn't notice this at first and on my second game was going to report a bug. I normally like to decide what techs I'm researching then add all the designs I'll likely need at the same time, if they don't display at all I have to mess around and I don't see the need for this change (if there is one please say).

I definitely like the save without filename popup, the easy editing of old designs (my 2nd game used the new engine parts so the editing designs feature very useful)

More will follow as I use it more, but overall it's a massive step forward, thank you.

LGM-Doyle:
> my old designs haven't loaded in at all 

Fixed.

During testing my directory of legacy ship designs was accidentally converted. I continued to verify that the conversion code worked with these already converted ships. I have made some new legacy designs and fixed the conversion code.

> specifically the premades need to be given an order so I'll learn how to do that

Here are the instructions.  If you know of the expected location in the website for these scripter instructions then you or I can copy them there.

This method is not discoverable, since it was a secondary objective and I did not want a way for players to accidentally start modifying their default directory.

------------------ Start of Instructions -------------------
****Editing default ship designs and monsters****
1. Move or backup the `shipdesigns` directory in your freeorion data directory.
2. Copy the `default/scripting/ship_designs` or `default/scripting/monsters` directory into your
data directory and replace your `shipdesigns` directory.  
3. Start a new game.
4. Your saved ship designs will now be the default ship designs or monsters.  Edit the ships.  
    i. Add new saved designs.  
    ii. Drag and drop to change ordering.
    iii. If you use a stringtable index for **both** the ship name **and** description then it will save the stringtable index and use the translated text for display.  You can check this in the pedia window.  When you add/remove parts the name and description in the pedia window will be pulled from the stringtable.
    iv. You can not start a ship with a monster hull, but you can add/remove components.  
5. Exit the game.
6. Copy the `shipdesigns` directory from your data directory back into the `default/scripting/ship_designs` or `default/scripting/monsters` directory.

------------------ End of Instructions -------------------

> new designs appear at the top of the list of ships

I preferred that the new designs appear at the top so that in the production window they are at the break between buildings and ships.  Usually when I make a new design, I build it and this avoids scrolling to the bottom of the list.   

> Problem: before the 'saved' tab automatically showed all designs including ones not unlocked, now I have to click 'unavailable'  ... I don't see the need for this change (if there is one please say).

There is no technical reason it is this way.

It meets my expectation that unavailable designs are unavailable.  If I'm cherry picking saved designs I expect the filtering options to be the same.  If I'm not using the saved designs auto load then I look at the saved designs to see what good designs from games past can I use now, with the tech I've got currently.

I'll have to think about your concerns and your play style.  Your suggestion from the forum of being able to add all the saved designs for a particular hull might help in this area.  

Here is a description of my play style with this PR if it helps.  I play with the default design auto load off and the saved design auto load turned on.  As the game progresses, and I research techs, ships become available and appear in the list.  I obsolete ships when they are superseded. In the production window there is only a small list of designs to choose from.      

I use the saved designs list to work on the "master list" of perfect designs.  It is kinda a meta game.

MatGB:
Still not getting my old designs, might I need to do something my end?

Just had a crash, I looked at saved designs, then looked at existing empire designs and revealed unavailable, then went to empty hulls and double clicked Titanic, that caused the crash.

Logs attached, let me know which area needs to be set to a higher level of logging if it's not obvious here.
[design-crash.zip](https://github.com/freeorion/freeorion/files/1052136/design-crash.zip)

LGM-Doyle:
@MatGB I was able to duplicate the crash only once.  I added some code to avoid crashing and generate an error message.  I've been unable to crash it or generate the error message. I don't think that the underlying problem is fixed, but without a more reliable way to crash I'm not making further progress.

On to loading your old designs.  I wrote the code to fix the problem.  I tested the code to fix the problem.  I wrote a note about fixing the problem.  I never pushed the code to fix the problem.

Sorry.

Be sure to copy your old designs back into the <data_dir>shipdesigns directory.

parse/Parse.cpp
```diff
@@ -441,25 +441,33 @@ namespace parse {
...
                 } else {
-                    TraceLogger() << "Parse: Skipping file " << file.string()
+                    TraceLogger(parse) << "Parse: Skipping file " << file.string()
                                   << " due to extension (" << fn_stem_ext << fn_ext << ")";
                 }
             }
+
+            // If in permissive mode and no scripts are found allow all files to be scripts.
+            if (scripts.empty() && !files.empty()) {
+                WarnLogger() << path << " does not contain scripts with the expected suffix .focs.txt. "
```
dbenage-cx: `path.string()`

LGM-Doyle:
The loading of legacy ships is still broken, if the suffix is not ".focs.txt"  I will work on a fix tomorrow.

parse/Parse.cpp
geoffthemedio: or probably better `PathString(path)`?

MatGB:
I can't get #1600 to play nicely with this through rebasing or cherry picking so can't test it on old designs but am mostly finding it works fine with updates and similar.

LGM-Doyle:
@MatGB as of about an hour ago, when I posted #1601, this PR is rebased onto master which includes #1600.  I should have posted a note to that effect.

It also includes #1601.  Now incorrect designs, wrong number of slots, mismatches between parts and slots etc. should produce a fairly explicit error message showing the erroneous FOCS and a version that is valid.

Thanks for wheeling around and re-testing.

MatGB:
Another crash, trying to call up a Symbiotic Hull. Am now getting *some* old designs on the screen but only for hulltypes I've played with since the change, the last crash was when I first researched a Titan, is it possible the parser isn't liking hulls it hasn't met before? That would be weird but that's what the pattern looks.
[design-crash-freeorion.log.zip](https://github.com/freeorion/freeorion/files/1062689/design-crash-freeorion.log.zip)

LGM-Doyle:
@MatGB two things.

I think that I have fixed the crash.  Double clicking on a hull was setting the hull and the parts, but updating the element from which it was getting the parts before it set the parts.  Most of the time it didn't crash, which is why it seemed various unrelated effects were causing the crash

The upgrade for legacy designs is a one time thing.  If *none* of the files in the 'shipdesigns' directory have the proper '.focs.txt' extension then it will try to use every file as a FOCS script and upgrade all of the designs.  If some of the files have the correct extension and some don't, then it will ignore the files without the correct suffix.  

That would explain the pattern you are seeing if one of the previous versions of this PR did not upgrade the designs in your directory, but did add some new files.  Designs that are upgraded should have the '.focs.txt' extension and have a uuid in the file.

A way to get the old ones upgraded would be to move the new ones, ones with the '.focs.txt' extension, out of the 'shipdesigns/' directory.  Then start a game, so that the old ones are upgraded.  Quit the game and move the previously upgraded designs back into 'shipdesigns/'



I think that this is the correct approach, because other than the small number of people testing this PR, nobody has a partially converted shipdesigns directory.

MatGB:
OK, that makes sense and having moved my new designs it's now converted all the old ones. It hasn't deleted the old designs which possibly isn't ideal but it's liveable, obviously can't tell about crashing yet, thanks. 

And agree that the moving files as a workaround is perfectly acceptable as it's likely only a handful that've tested this at all, and we're all capable of doing that.

MatGB:
More feedback: if you're in the saved designs column, the only way to add a design is to right click on the column.

If you want a saved design that, for example, has plasma shields but you've already researched blackshield then it's natural to want to load that design, edit it and add it to empire designs. Generally in that specific case I'm unlikely to want to edit the saved design because it'll be rare I want something new with a high RP shield at the end game. Regardless, I want a way to add the design to my buildable designs from the saved column after editing it, perhaps a third button alongside 'update design' and 'add saved design'? 

Specifically I think we need something there anyway as both those buttons now do something other than what they did before and that's still confusing me, definitely reword to 'update saved design' and 'save new design', with perhaps 'add design to empire'?

LGM-Doyle:
@MatGB 

I changed the labels on the buttons from "Add New Design" and "Add Saved Design" to "Add Finished Design" and "Add Saved Design" 
I changed the labels on the buttons from "Update Design" and "Update Saved Design" to "Update Finished Design" and "Update Saved Design" 

I changed the tracking that determines when to add a design to finished or saved designs.

Previously, both update and add new design decided where to make a change by they type of design being edited, finished or saved.  That is unchanged for updating.

Now adding a new design uses the last tab visited,  finished or saved to determine if it will "Add Finished Design" or "Add Saved Design".

Now a player can start editing a saved design and go to the finished design tab.  Both buttons will be active and say "Update Saved Design" and "Add Finished Design".  I think that this is the situation, you were describing where you have a saved design you do not want to change that is almost correct for the current empire's situation.

This change also means that you can directly add saved designs from an empty hull start, by selecting the hull, adding parts and then clicking on the saved designs tab, which will make new designs go to the saved designs.

LGM-Doyle:
@geoffthemedio I think that this PR is now pared down close to an indivisible core, which will be difficult to split further without producing a broken experience in master.  

Most of the changes are in DesignWnd.cpp and can be summed up as create and use the current and saved design managers.  It is possibly easiest to start reviewing in DesignWnd.cpp.  Top to bottom the changes are, adding the current and saved design managers, changing the `BasesListBox` to work with them, and changing the main panel to work with the `BasesListBoxes`.

Changes in other files are either for startup or the common code in ShipDesigns for loading monsters, predefined designs or saved designs.

There is a problem with the VC++ build.  I think it is in the project, files which I have modified by hand and am unable to debug.

UI/DesignWnd.h
```diff
@@ -2,8 +2,49 @@
...
+};
+
+/** ShipDesignManager tracks information known to the client about ShipDesigns.  This includes
+    the order of designs in the UI. The information is currently used in the DesignWnd and the
+    BuildDesignator. */
+class ShipDesignManager {
+public:
+
+    /** Designs provides ordered lists of designs for display in the UI. */
+    class Designs {
```
geoffthemedio: what is this class for? why not just store a vector in ShipDesignManager?

UI/DesignWnd.h
```diff
@@ -2,8 +2,49 @@
...
+        virtual std::vector<int> OrderedIDs() const
+            /* VC++ won't link ShipDesignManager if Designs is pure *///= 0;
+        { return std::vector<int>{}; }
+    };
+
+    ShipDesignManager();
+    ~ShipDesignManager();
+
+    void StartGame(int empire_id);
+    void Load(const SaveGameUIData& data);
```
geoffthemedio: seemingly needs to be `virtual`

UI/ClientUI.cpp
```diff
@@ -978,8 +983,11 @@ std::vector<std::shared_ptr<GG::Texture>> ClientUI::GetPrefixedTextures(const bo
     return prefixed_textures_and_dist.first;
 }
 
-void ClientUI::RestoreFromSaveData(const SaveGameUIData& ui_data)
-{ GetMapWnd()->RestoreFromSaveData(ui_data); }
+void ClientUI::RestoreFromSaveData(const SaveGameUIData& ui_data) {
+    GetMapWnd()->RestoreFromSaveData(ui_data);
+    m_ship_designs->Load(ui_data);
+
```
geoffthemedio: blank line

UI/DesignWnd.cpp
```diff
@@ -122,134 +124,629 @@ namespace {
...
+        void Load(const std::vector<std::pair<int, bool>>& ids_and_obsoletes);
+        std::vector<std::pair<int, bool>> Save() const;
+
+        private:
+        int m_empire_id;
+        std::list<int> m_ordered_ids;
+
+        // An index from the id to the obsolescence state and the location in the m_ordered_ids list.
+        std::unordered_map<int, std::pair<bool,  std::list<int>::const_iterator>> m_id_to_obsolete_and_loc;
+
```
geoffthemedio: blank line

UI/DesignWnd.cpp
```diff
@@ -122,134 +124,629 @@ namespace {
...
+        void Load(const std::vector<std::pair<int, bool>>& ids_and_obsoletes);
+        std::vector<std::pair<int, bool>> Save() const;
+
+        private:
+        int m_empire_id;
+        std::list<int> m_ordered_ids;
+
+        // An index from the id to the obsolescence state and the location in the m_ordered_ids list.
+        std::unordered_map<int, std::pair<bool,  std::list<int>::const_iterator>> m_id_to_obsolete_and_loc;
```
geoffthemedio: extra space after ,

UI/DesignWnd.h
```diff
@@ -2,8 +2,49 @@
...
+    };
+};
+
+/** ShipDesignManager tracks information known to the client about ShipDesigns.  This includes
+    the order of designs in the UI. The information is currently used in the DesignWnd and the
+    BuildDesignator. */
+class ShipDesignManager {
+public:
+
+    /** Designs provides ordered lists of designs for display in the UI. */
```
geoffthemedio: describe use / design somewhat, particularly that is it subclassed...

UI/DesignWnd.h
LGM-Doyle: All changes are done.

Thanks a lot for figuring this out.  I was looking at the msvc project file for causes of the error.

Do you understand why VC++ requires it to be `virtual`?  It is never sub-classed.

PR #1588, opened 2017-05-27
Add UUID to ship design parser.

LGM-Doyle:
This PR depends on #1545 to convert ship design parsing to use `unique_ptr`

It adds the ability to parse a UUID as part of the ship design.

It makes all ShipDesign creators, server for default designs and monsters, and clients AI and human correctly add either an existing from disk UUID or a new random UUID to ship designs.

The ShipDesignOrder correctly transfers UUIDs to the server.

It adds a UUID to each default design and monster.

### DIFF
--- a/AI/AIInterface.cpp
+++ b/AI/AIInterface.cpp
@@ -23,6 +23,7 @@
 
 #include <boost/timer.hpp>
 #include <boost/python/str.hpp>
+#include <boost/uuid/random_generator.hpp>
 
 
 #include <map>
@@ -847,9 +848,11 @@ namespace AIInterface {
         int empire_id = AIClientApp::GetApp()->EmpireID();
         int current_turn = CurrentTurn();
 
+        const auto uuid = boost::uuids::random_generator()();
+
         // create design from stuff chosen in UI
         ShipDesign* design = new ShipDesign(name, description, current_turn, ClientApp::GetApp()->EmpireID(),
-                                            hull, parts, icon, model, nameDescInStringTable);
+                                            hull, parts, icon, model, nameDescInStringTable, false, uuid);
 
         if (!design) {
             ErrorLogger() << "IssueCreateShipDesignOrderOrder failed to create a new ShipDesign object";
--- a/AI/AIInterface.cpp
+++ b/AI/AIInterface.cpp
@@ -834,7 +834,7 @@ namespace AIInterface {
 
     int IssueCreateShipDesignOrder(const std::string& name, const std::string& description,
                                    const std::string& hull, const std::vector<std::string> parts,
-                                   const std::string& icon, const std::string& model, bool nameDescInStringTable)
+                                   const std::string& icon, const std::string& model, bool name_desc_in_stringtable)
     {
         if (name.empty() || description.empty() || hull.empty()) {
             ErrorLogger() << "IssueCreateShipDesignOrderOrder : passed an empty name, description, or hull.";
@@ -852,7 +852,7 @@ namespace AIInterface {
 
         // create design from stuff chosen in UI
         ShipDesign* design = new ShipDesign(name, description, current_turn, ClientApp::GetApp()->EmpireID(),
-                                            hull, parts, icon, model, nameDescInStringTable, false, uuid);
+                                            hull, parts, icon, model, name_desc_in_stringtable, false, uuid);
 
         if (!design) {
             ErrorLogger() << "IssueCreateShipDesignOrderOrder failed to create a new ShipDesign object";
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -120,7 +120,7 @@ namespace {
 
     const std::string DESIGN_FILENAME_EXTENSION = ".txt";
     const std::string UNABLE_TO_OPEN_FILE = "Unable to open file";
-    boost::filesystem::path SavedDesignsDir() { return GetUserDataDir() / "shipdesigns"; }
+    boost::filesystem::path SavedDesignsDir() { return GetUserDataDir() / "shipdesigns/"; }
 
     class SavedDesignsManager {
     public:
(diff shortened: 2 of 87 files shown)


AI/AIInterface.cpp
```diff
@@ -847,9 +848,11 @@ namespace AIInterface {
         int empire_id = AIClientApp::GetApp()->EmpireID();
         int current_turn = CurrentTurn();
 
+        const auto uuid = boost::uuids::random_generator()();
+
         // create design from stuff chosen in UI
         ShipDesign* design = new ShipDesign(name, description, current_turn, ClientApp::GetApp()->EmpireID(),
-                                            hull, parts, icon, model, nameDescInStringTable);
+                                            hull, parts, icon, model, nameDescInStringTable, false, uuid);
```
geoffthemedio: use underscores not camel case for variable / parameter names.

UI/DesignWnd.cpp
```diff
@@ -120,22 +121,20 @@ namespace {
 
     const std::string DESIGN_FILENAME_EXTENSION = ".txt";
     const std::string UNABLE_TO_OPEN_FILE = "Unable to open file";
-    boost::filesystem::path SavedDesignsDir() { return GetUserDataDir() / "shipdesigns"; }
+    boost::filesystem::path SavedDesignsDir() { return GetUserDataDir() / "shipdesigns/"; }
```
geoffthemedio, line 124: why?

parse/ShipDesignsParser.cpp
```diff
@@ -5,24 +5,65 @@
...
+    BOOST_PHOENIX_ADAPT_FUNCTION(bool, is_valid_uuid_, is_valid_uuid, 1)
+
+    // A UUID parser
+    boost::uuids::uuid parse_uuid(const std::string& uuid_string) {
+        try {
+            return boost::lexical_cast<boost::uuids::uuid>(uuid_string);
+        } catch (boost::bad_lexical_cast&) {
+            // This should never happen because the previous is_valid should short circuit.
+            ErrorLogger() << uuid_string << " is not a valid UUID.  A valid UUID looks like 01234567-89ab-cdef-0123-456789abcdef";
+            return boost::uuids::uuid{{0}};
```
geoffthemedio: below you use `boost::uuids::nil_generator()()`... are these distinct?

parse/ShipDesignsParser.cpp
```diff
@@ -120,26 +169,22 @@ namespace {
...
-        for (const boost::filesystem::path& file : ListScripts("scripting/monster_designs")) {
-            result &= detail::parse_file<rules, std::map<std::string, ShipDesign*>>(file, designs);
-        }
+    bool ship_designs(std::map<std::string, std::unique_ptr<ShipDesign>>& designs)
+    { return ship_designs("scripting/ship_designs", designs); }
 
-        return result;
-    }
+    bool monster_designs(std::map<std::string, std::unique_ptr<ShipDesign>>& designs)
+    { return ship_designs("scripting/monster_designs", designs); }
```
geoffthemedio, line 189: doesn't originate with this pull request, but I'm wondering if passing a relative directory like this is portable...

universe/ShipDesign.h
```diff
@@ -111,6 +112,7 @@ class FO_COMMON_API PartType {
     /** \name Accessors */ //@{
     const std::string&      Name() const            { return m_name; };             ///< returns name of part
     const std::string&      Description() const     { return m_description; }       ///< returns description string, generally a UserString key.
+
```
geoffthemedio: ?

universe/ShipDesign.h
```diff
@@ -432,6 +443,11 @@ class FO_COMMON_API ShipDesign {
     const std::string&              Name(bool stringtable_lookup = true) const;
     void                            SetName(const std::string& name);
 
+    /** Return the UUID. */
+    boost::uuids::uuid              UUID() const { return m_uuid; }
+    /** Set the UUID. */
+    void                            SetUUID(const boost::uuids::uuid& uuid);
```
geoffthemedio: move into Mutators section of header

util/Order.h
```diff
@@ -613,6 +614,7 @@ class FO_COMMON_API ShipDesignOrder : public Order {
     void ExecuteImpl() const override;
 
     int m_design_id = INVALID_OBJECT_ID;
+    boost::uuids::uuid m_uuid;
```
geoffthemedio, line 617: is this related to this pull request? it's not mentioned in or seemingly relevant to the description.

parse/ShipDesignsParser.cpp
LGM-Doyle: They are not distinct.  

I should be using the `nil_generator()` or `nil_uuid()` and the test `is_nil()` everywhere, but I started out using the constructor `{{0}}` and an equality test.

Eventually, all of the nil generators are/should be replaced with random generators.  `{{0}}` is like the null pointer of UUID.  It indicates that something is not initialized.

I've removed the `{{0}}` from this PR.

parse/ShipDesignsParser.cpp
LGM-Doyle: It is portable.  Boost filesystem is on track to be part of standard C++17.

universe/ShipDesign.h
LGM-Doyle: removed.

universe/ShipDesign.h
LGM-Doyle: moved.

util/Order.h
LGM-Doyle: I've added it to the description.  It is part of make ShipDesignOrder copy the design UUID.  It is there so that when new designs are created by a client AI/human the UUID is copied along with the other ship design data.

LGM-Doyle (commented):
Updated.

parse/ShipDesignsParser.cpp
geoffthemedio: I mean the use of `/` baked into the path, as opposed to using `"scripting" / "monster_designs"`

geoffthemedio:
Xcode build fails for nontrivial reasons...

LGM-Doyle:
Thanks for the heads up.

`uuid` is POD for efficiencies sake.  That means that each compiler will generate copy constructors as it sees fit.  I think what is happening is that clang has generated a list initializer constructor that is grabbing only the first element of the underlying data array.  I've switched to parentheses, which will hopefully find the copy constructor.

parse/ShipDesignsParser.cpp
LGM-Doyle: I misinterpreted the question.

Yes it is portable.

The POSIX forward slash in pathnames '/'  is referred to as the generic format and must be accepted by all conforming implementations of filesystem for all systems.

parse/ShipDesignsParser.cpp
```diff
@@ -133,26 +169,22 @@ namespace {
...
-    bool ship_designs(const boost::filesystem::path& path, std::map<std::string, std::unique_ptr<ShipDesign>>& designs)
-    { return detail::parse_file<rules, std::map<std::string, std::unique_ptr<ShipDesign>>>(path, designs); }
-
-    bool ship_designs(std::map<std::string, std::unique_ptr<ShipDesign>>& designs) {
+    bool ship_designs(const boost::filesystem::path& path, std::map<std::string, std::unique_ptr<ShipDesign>>& designs) {
         bool result = true;
-
-        for(const boost::filesystem::path& file : ListScripts("scripting/ship_designs")) {
-            result &= detail::parse_file<rules, std::map<std::string, std::unique_ptr<ShipDesign>>>(file, designs);
+        for(const boost::filesystem::path& file : ListScripts(path)) {
```
geoffthemedio, line 174: should have a space before (, but predates your pull request

PR #1589, opened 2017-05-27
Add ShipDesignManfest to parser.

LGM-Doyle:
This PR depends on #1588

It adds a parser for ShipDesignManifest which stores the order of ship designs in a ship design directory such as the default designs, monsters or saved designs.

PR #1590, opened 2017-05-27
Remove design wnd save file dialog

LGM-Doyle:
This PR removes the save file dialog for saved designs.

It depends on #1589

The dialog's purpose was to get a valid file name.

Its problems were that it breaks game immersion and allows for incorrect file names.

Furthermore the ship name allows for embedded decorator code, which will also result in an invalid suggested file name.

This PR generates the name from the design UUID and the saved design directory option.  The name is always valid and the player is not bothered with details of individual ship names.

### DIFF
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -218,6 +218,7 @@ namespace {
             if (!ofs)
                 throw std::runtime_error(UNABLE_TO_OPEN_FILE);
             ofs << design.Dump();
+            TraceLogger() << "Wrote ship design to " << PathString(file);
 
         } catch (const std::exception& e) {
             ErrorLogger() << "Error writing design file.  Exception: " << ": " << e.what();
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -34,6 +34,7 @@
 #include <boost/filesystem/operations.hpp>
 #include <boost/filesystem/fstream.hpp>
 #include <boost/uuid/random_generator.hpp>
+#include <boost/uuid/uuid_io.hpp>
 
 #include <algorithm>
 #include <iterator>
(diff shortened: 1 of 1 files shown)


LGM-Doyle:
Rebased.

UI/DesignWnd.cpp
```diff
@@ -210,24 +212,20 @@ namespace {
     SavedDesignsManager& GetSavedDesignsManager()
     { return SavedDesignsManager::GetSavedDesignsManager(); }
 
-    void WriteDesignToFile(int design_id, boost::filesystem::path& file) {
-        const ShipDesign* design = GetShipDesign(design_id);
-        if (!design)
-            return;
-
+    void WriteDesignToFile(const ShipDesign& design, boost::filesystem::path& file) {
```
geoffthemedio, line 215: Seems like there should be some sent to InfoLogger when the design write succeeds, and possibly also to the messages window...

LGM-Doyle: I added a TraceLogger statement.

I did not add a messages window message.  The user confirmation that the save worked is that the saved design changed in the design window.

From an in game perspective these saves are not a user initiated disk action, and should not be confirmed in game.

Saving the game is user initiated and there is no other confirmation (spinner etc.) that the save succeeded.  The user initiates the save and may expect a confirmation that the save has happened.

Adjusting the tech status filter also changes the on disk "config.xml" file but does not have a message window confirmation of the change.  The confirmation that it worked is immediate change of the filter status.  The confirmation that it persisted is when it re-loads in the same state that it exited with.

During development I did not have any additional debug statements in this location that I subsequently removed.  So I think more logging here just adds noise.

geoffthemedio:
Right clicking save design on a design in the finished designs list doesn't cause it to appear in the saved designs list. Resigning and starting a new name also doesn't make it appear. Have to restart the client to see the new designs.

If I have a .txt design file, it is ignored if there are any of the new-style designs in the directory as well... If planning to support the older files, making them all disappear as soon as the player makes a new one is confusing. Is "converting" the old design files in a later pull request? (This was mentioned in the other pull request for the whole series together)

MatGB:
The latter is fixed now the conversion is working: remove any new designs you have in the folder and it'll convert the old designs into the new format, then put the new designs back. Given it's only you, me and possibly a small handful of others played with the setup it won't matter once we merge in and it converted all of mine fine once we figured out what the problem was. (the conversion's in the 'big' PR with everything, not sure where else it is in the split PRs)

geoffthemedio:
OK. Hopefully the saved designs will also appear in the saved tab immediately at some point.

PR #1591, opened 2017-05-27
Fix ListBox handling of dropped windows.

LGM-Doyle:
There are a number of issues with QueueListBox and ListBox handling of dropped windows:
- they use `string` to indicate if a drop type should be accepted.  However, they use opposite conventions for the empty string '""'  to indicate allow no types or allow all types.
- QueueListBox::AcceptsDrops never actually accepted the dropped window.  In all cases its derived classes  just regenerated everything.
- there was an unused plan to throw an exception to indicate not accepting a dropped window.  

This PR adds the `ListBox` functions `void AllowDrops(bool)`, `bool AllowingDrops()`, `void AllowAllDropTypes(bool)` and `bool AllowedDropType(string)` as clearer alternatives to passing empty strings to allowing all/no drops.

This PR improves `AcceptDrops` so that `QueueListBox` uses the `AcceptDrops` of `ListBox` and doesn't require derived classes to completely redraw themselves in response to a drop.

This PR removes the exception that was never caught and never thrown.

### DIFF
--- a/GG/GG/ListBox.h
+++ b/GG/GG/ListBox.h
@@ -97,13 +97,7 @@ extern GG_API const ListBoxStyle LIST_BROWSEUPDATES;  ///< Causes a signal to be
     <br>Note that drag-and-drop support is a key part of ListBox's
     functionality.  As such, special effort has been made to make its use as
     natural and flexible as possible.  This includes allowing arbitrary
-    reordering of ListBox rows when the LIST_NOSORT is in effect, and includes
-    the use of the DontAcceptDrop exception.  The DontAcceptDrop exception can
-    be thrown by any client of the ListBox in response to its
-    DropAcceptableSignal.  Such a throw will cause the drop to be refused.
-    Note that a DropAcceptableSignal is emitted for each row dropped into the
-    ListBox, so individual rows may be accepted or rejected from a single
-    multi-row drop. */
+    reordering of ListBox rows when the LIST_NOSORT is in effect.*/
 class GG_API ListBox : public Control
 {
 public:
@@ -495,13 +489,6 @@ public:
     };
 
     /** \name Exceptions */ ///@{
-    /** The base class for ListBox exceptions. */
-    GG_ABSTRACT_EXCEPTION(Exception);
-
-    /** Thrown by a ListBox that does not wish to accept a potential drop, for
-        whatever reason. This may be throw by anyone -- even in client code
-        activated by a DropAcceptableSignal. */
-    GG_CONCRETE_EXCEPTION(DontAcceptDrop, GG::ListBox, Exception);
     //@}
 
 protected:
--- a/GG/GG/ListBox.h
+++ b/GG/GG/ListBox.h
@@ -223,9 +223,9 @@ public:
 
     /** \name Signal Types */ ///@{
     /** emitted when the list box is cleared */
-    typedef boost::signals2::signal<void ()>                                                ClearedSignalType;
+    typedef boost::signals2::signal<void ()>                                                ClearedRowsSignalType;
     /** emitted when one or more rows are selected or deselected */
-    typedef boost::signals2::signal<void (const SelectionSet&)>                             SelChangedSignalType;
+    typedef boost::signals2::signal<void (const SelectionSet&)>                             SelRowsChangedSignalType;
     /** the signature of row-change-notification signals */
     typedef boost::signals2::signal<void (iterator)>                                        RowSignalType;
     /** the signature of const row-change-notification signals */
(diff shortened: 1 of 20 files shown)


GG/src/ListBox.cpp
```diff
@@ -1441,10 +1460,20 @@ void ListBox::AddPaddingAtEnd(bool enable)
...
+        m_allowed_drop_types = std::unordered_set<std::string>();
+    m_allowed_drop_types->insert(str);
+}
 
 void ListBox::DisallowDropType(const std::string& str)
-{ m_allowed_drop_types.erase(str); }
+{
+    // Create the set if necessary
+    if (!m_allowed_drop_types)
+        m_allowed_drop_types = std::unordered_set<std::string>();
```
geoffthemedio: If `DisallowDropType` is called when m_allowed_drop_types is `none`, that will presumably disallow all drop types? A bit awkward / potentially unexpected behaviour...

UI/DesignWnd.cpp
```diff
@@ -1183,7 +1183,7 @@ class BasesListBox : public QueueListBox {
     static const std::string BASES_LIST_BOX_DROP_TYPE;
 
     /** \name Structors */ //@{
-    BasesListBox(const std::string& drop_type = "");
+    BasesListBox(const boost::optional<std::string>& drop_type = boost::none);
```
geoffthemedio: not new to your code, but should probably be explicit

UI/DesignWnd.cpp
```diff
@@ -1448,7 +1448,7 @@ bool BasesListBox::SavedDesignListBoxRow::LookupInStringtable() const {
 
 const std::string BasesListBox::BASES_LIST_BOX_DROP_TYPE = "BasesListBoxRow";
 
-BasesListBox::BasesListBox(const std::string& drop_type) :
+BasesListBox::BasesListBox(const boost::optional<std::string>& drop_type) :
```
geoffthemedio, line 1451: explicit

GG/GG/ListBox.h
```diff
@@ -323,12 +320,17 @@ class GG_API ListBox : public Control
...
+    bool AllowingDrops();
+
+
     mutable ClearedSignalType        ClearedSignal;         /// the cleared signal object for this ListBox
     mutable BeforeInsertSignalType   BeforeInsertSignal;    ///< the before insert signal object for this ListBox
     mutable AfterInsertSignalType    AfterInsertSignal;     ///< the after insert signal object for this ListBox
     mutable SelChangedSignalType     SelChangedSignal;      ///< the selection change signal object for this ListBox
     mutable DroppedSignalType        DroppedSignal;         ///< the dropped signal object for this ListBox
     mutable DropAcceptableSignalType DropAcceptableSignal;  ///< the drop-acceptability signal object for this ListBox
+    mutable RowMovedSignalType       MovedSignal;           ///< the moved signal object for this ListBox
```
geoffthemedio: `MovedSignal` is ambiguous... does it refer to moving the whole control, or just a row?

UI/DesignWnd.cpp
```diff
@@ -1544,27 +1544,26 @@ void BasesListBox::ChildrenDraggedAway(const std::vector<GG::Wnd*>& wnds, const
     DetachChild(wnds.front());
 }
 
-void BasesListBox::QueueItemMoved(GG::ListBox::Row* row, std::size_t position) {
-    BasesListBox::CompletedDesignListBoxRow* control =
-        boost::polymorphic_downcast<BasesListBox::CompletedDesignListBoxRow*>(row);
-    Empire* empire = GetEmpire(m_empire_id_shown);
-    if (control && empire && position <= NumRows()) {
+void BasesListBox::QueueItemMoved(const GG::ListBox::iterator& row_it, const GG::ListBox::iterator& original_position_it) {
```
geoffthemedio: I'd break those parameters over two lines

UI/DesignWnd.cpp
```diff
@@ -1544,27 +1544,26 @@ void BasesListBox::ChildrenDraggedAway(const std::vector<GG::Wnd*>& wnds, const
...
+    if (BasesListBox::CompletedDesignListBoxRow* control =
+        dynamic_cast<BasesListBox::CompletedDesignListBoxRow*>(*row_it))
+    {
+        if(!GetEmpire(m_empire_id_shown))
```
geoffthemedio: space b4 (

UI/DesignWnd.cpp
```diff
@@ -1544,27 +1544,26 @@ void BasesListBox::ChildrenDraggedAway(const std::vector<GG::Wnd*>& wnds, const
...
         int insert_before_id = insert_before_control
             ? insert_before_control->DesignID() : INVALID_DESIGN_ID;
 
-        if (design_id != insert_before_id)
-            //insert_before_row may be end() to insert as last item
-            Insert(control, insert_before_row);
         control->Resize(ListRowSize());
         HumanClientApp::GetApp()->Orders()
             .IssueOrder(std::make_shared<ShipDesignOrder>(m_empire_id_shown, design_id, insert_before_id));
+        return;
```
geoffthemedio: not in a loop (I think?) ... what does this return do?

GG/GG/ListBox.h
geoffthemedio: The concern is for cases where the signature isn't apparent, such as when connecting the signal to a function by name.

The other signal names are mostly not ambiguous, though...  Perhaps `DroppedSignal` could refer to the `ListBox` being dropped, rather than being dropped on, but that wouldn't be my default assumption. With `MovedSignal`, it could easily mean the `ListBox` itself is being moved around.

LGM-Doyle: I've changed them all.  `LeftClickedSignal` was also ambiguous, but the new `LeftClickedRowSignal` is not.

PR #1593, opened 2017-05-28
Remove all Logger(X) declarations from anonymous namespaces.

LGM-Doyle:
This is intended to change the linker behavior on OSX which is only
linking some of the loggers.

See issue #1575.


LGM-Doyle:
@memem359 solved the linker issues on OSX, so I'm closing this PR.

PR #1594, opened 2017-05-30
Change OrderSet add suppress_ immediate_execution parameter.

LGM-Doyle:
This allows orders to be injected into the OrderSet at startup and load
without being executed immediately and then again in MapWnd::InitTurn()

This will allow saved ship designs to be added to its empire by the human client at game start (new game or loaded game).

### DIFF
--- a/util/OrderSet.cpp
+++ b/util/OrderSet.cpp
@@ -15,14 +15,15 @@ const OrderPtr OrderSet::ExamineOrder(int order) const {
     return retval;
 }
 
-int OrderSet::IssueOrder(const OrderPtr& order)
-{ return IssueOrder(OrderPtr(order)); }
+int OrderSet::IssueOrder(const OrderPtr& order, bool suppress_immediate_execution /*= false*/)
+{ return IssueOrder(OrderPtr(order), suppress_immediate_execution); }
 
-int OrderSet::IssueOrder(OrderPtr&& order) {
+int OrderSet::IssueOrder(OrderPtr&& order, bool suppress_immediate_execution /*= false*/) {
     int retval = ((m_orders.rbegin() != m_orders.rend()) ? m_orders.rbegin()->first + 1 : 0);
     auto inserted = m_orders.insert(std::make_pair(retval, std::forward<OrderPtr>(order)));
 
-    inserted.first->second->Execute();
+    if (!suppress_immediate_execution)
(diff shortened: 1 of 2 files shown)


util/OrderSet.h
```diff
@@ -52,9 +52,11 @@ class FO_COMMON_API OrderSet {
     //@}
 
     /** \name Mutators */ //@{
-    /** executes a given Order, then stores it in the OrderSet. Returns an index that can be used to reference the
-        order.*/
-    int            IssueOrder(const OrderPtr& order);
+    /** executes \p order unless \p suppress_immediate_execution is true, then
```
geoffthemedio: not clear whether order is stored if `suppress_immediate_execution` is false

util/OrderSet.cpp
```diff
@@ -15,11 +15,15 @@ const OrderPtr OrderSet::ExamineOrder(int order) const {
...
 }
 
-int OrderSet::IssueOrder(const OrderPtr& order) {
+int OrderSet::IssueOrder(const OrderPtr& order, bool suppress_immediate_execution /*= false*/)
+{ return IssueOrder(OrderPtr(order), suppress_immediate_execution); }
+
+int OrderSet::IssueOrder(OrderPtr&& order, bool suppress_immediate_execution /*= false*/) {
     int retval = ((m_orders.rbegin() != m_orders.rend()) ? m_orders.rbegin()->first + 1 : 0);
-    m_orders[retval] = order;
+    auto inserted = m_orders.insert(std::make_pair(retval, std::forward<OrderPtr>(order)));
```
geoffthemedio: comment explaining this would be helpful...

util/OrderSet.h
```diff
@@ -52,9 +52,11 @@ class FO_COMMON_API OrderSet {
     //@}
 
     /** \name Mutators */ //@{
-    /** executes a given Order, then stores it in the OrderSet. Returns an index that can be used to reference the
-        order.*/
-    int            IssueOrder(const OrderPtr& order);
+    /** If \p suppress_immediate_execution is true, then execute the \p order on the client.
```
geoffthemedio: do you mean if false then execute?

LGM-Doyle: That is what I meant.  I've fixed it.

PR #1595, opened 2017-05-30
Change ServerApp to store OrderSets with unique_ptr

LGM-Doyle:
In some locations m_turn_sequence was being cleared or set to a new heap
object without deleting the old order.

Changing to unique_ptr fixes the leaks.


geoffthemedio (commented):
Can't test but looks fine

PR #1596, opened 2017-05-30
Change Empire to add an overload ShipDesignAvailable(const ShipDesign&)

LGM-Doyle:
This will allow checking a saved design availability which has not been
added to the universe.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1672,22 +1672,22 @@ const std::vector<int>& Empire::OrderedShipDesigns() const
 { return m_ship_designs_ordered; }
 
 bool Empire::ShipDesignAvailable(int ship_design_id) const {
-    //// if design isn't kept by this empire, it can't be built.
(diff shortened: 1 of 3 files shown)


Dilvish-fo:
Looks like I managed to overlook that you already covered this, lol.  Nevermind.  
~~Hmm, I'd like to note one potential complication-- the python exposure of this method in EmpireWrapper.py.   This method is currently exposed but not used.  It is possible to still work with overloaded methods exposed to python, and I am pretty sure that we already have at least a couple, but I also vaguely recall running into some extra work getting the exposure to connect to the one we wanted.  Perhaps you'll be able to sort through it more readily/cleanly, but you can find one possible example handling with the UniverseWrapper exposure of ShipDesign::Parts() (the exposure of the second variant is commented out but I think it had worked).~~

LGM-Doyle:
Thanks for giving this a look over in any case.

PR #1599, opened 2017-06-06
Revert blocksize increase handling to be proportional reduction

### LINKED ISSUE #1574: increasing blocksize of partially completed ships destroys all current progress
### Environment
* **Operating System**: Ubuntu Linux
* **Fetched as** 
  * Compiled from source

### Description
increasing the blocksize of a partially completed ship-type build item will cause all current progress to be lost (unless the change is undone before advancing to the next turn).

### Expected Result
Doubling the blocksize (or increasing by some other factor N) should not cause all current progress to be lost, but instead should cause the current percentage completion to simply drop to 1/2 (or 1/N) of what it was before the change.

Dilvish-fo:
If the blocksize of a production queue build item is increased, reduce the progress in proportion to the increase. This essentially reverts a recent change (which seems like it might have accidentally slipped into c6cc240 since the commit message makes no mention of it), and fixes #1574

Per discussion at http://www.freeorion.org/forum/viewtopic.php?p=89017#p89017


Vezzra (approved):
Didn't test, but looks ok.

PR #1601, opened 2017-06-08
Improve ShipDesign constructor to (almost) always produce a valid design.

LGM-Doyle:
This PR depends on #1589.

It modifies the `ShipDesign` constructors so that they will
always produce a valid design except if the `HullTypeManager` has
no hulls at all.  In that case, the hull will be the empty string.

If provided with invalid input parameters the constructor will
remove parts or replace the hull with the first hull from the
hull manager, until it has a valid design.

When provided with invalid input parameters the constructor
produces an explicit error message showing the erroneous FOCS and
the corrected FOCS.  I've included an example at the bottom of
this post.

There is an option in the constructor to request it to throw on
bad input rather than produce the degraded design.  This enables
it to produce the detailed error message, because before the
constructor throws it has access to a complete `ShipDesign`.

Parsing and deserialization uses the other, non throwing, option
of generating a degraded but valid design.

***Sample error***:

````
11:17:15.918952 [debug] client : ShipDesign.cpp:932 : Invalid ShipDesign part "FT_HANGAR_1" can't be mounted in SL_EXTERNAL slot. Removing "FT_HANGAR_1"
11:17:15.918987 [debug] client : ShipDesign.cpp:932 : Invalid ShipDesign part "FT_BAY_1" can't be mounted in SL_INTERNAL slot. Removing "FT_BAY_1"
11:17:15.919029 [warn] client : ShipDesign.cpp:972 : Invalid ShipDesign:
ShipDesign
    name = "SD_CARRIER parts conflict with slot"
    uuid = "08a58b08-0929-496d-84fc-faa91424ca34"
    description = "SD_CARRIER_DESC"
    hull = "SH_STANDARD"
    parts = [
        "AR_STD_PLATE"
        "FT_HANGAR_1"
        "FT_BAY_1"
        "FT_BAY_1"
    ]
    model = "mark1"

ShipDesign was made valid as:
ShipDesign
    name = "SD_CARRIER parts conflict with slot"
    uuid = "08a58b08-0929-496d-84fc-faa91424ca34"
    description = "SD_CARRIER_DESC"
    hull = "SH_STANDARD"
    parts = [
        "AR_STD_PLATE"
        ""
        "FT_BAY_1"
        ""
    ]
    model = "mark1"
````

### DIFF
--- a/universe/ShipDesign.h
+++ b/universe/ShipDesign.h
@@ -417,7 +417,11 @@ FO_COMMON_API const HullType* GetHullType(const std::string& name);
 class FO_COMMON_API ShipDesign {
 public:
     /** \name Structors */ //@{
+private:
+    /** The ShipDesign constructor constructs invalid designs and is only used by boost
+        serialization. */
     ShipDesign();
+public:
     ShipDesign(const std::string& name, const std::string& description,
                int designed_on_turn, int designed_by_empire, const std::string& hull,
                const std::vector<std::string>& parts,
--- a/universe/ShipDesign.h
+++ b/universe/ShipDesign.h
@@ -9,6 +9,7 @@
 #include <set>
 #include <string>
 #include <vector>
+#include <stdexcept>
 
 #include <boost/algorithm/string/case_conv.hpp>
 #include <boost/filesystem/operations.hpp>
@@ -425,21 +426,38 @@ private:
         serialization. */
     ShipDesign();
 public:
-    /** The public ShipDesign constructor will only construct valid ship designs.  If the passed
-        in parameters (\p hull and \p parts) would result in an invalid design it throws
-        std::runtime_error. */
-    ShipDesign(const std::string& name, const std::string& description,
+    /** The public ShipDesign constructor will only construct valid ship
+        designs, as long as the HullTypeManager has at least one hull.
+
+        If \p should_throw is not boost::none and the passed in parameters (\p
+        hull and \p parts) would result in an invalid design it generates an
+        explicit log message showing the FOCS corresponding to the passed in
+        parameters and the FOCS corresponding to a corrected valid design and
+        then throws std::invalid_argument.  This can be used to test design
+        validity and provide an explcit log message.
+
+        should_throw is not used but it is a literal reminder that
+        std::invalid_argument should be caught.
+
+        If \p should_throw is boost::none it will correct the errors in
+        parameters, print a log message and return a valid ship design.  The
+        only exception is if there are no ship hull in the HullManager.  Then
+        the constructor will not throw, but will return an invalid ship design
+        with a empty "" hull.
+    */
+    ShipDesign(const boost::optional<std::invalid_argument>& should_throw,
+               const std::string& name, const std::string& description,
                int designed_on_turn, int designed_by_empire, const std::string& hull,
                const std::vector<std::string>& parts,
                const std::string& icon, const std::string& model,
                bool name_desc_in_stringtable = false, bool monster = false,
                const boost::uuids::uuid& uuid = boost::uuids::nil_uuid()
               );
-    /** The public ShipDesign constructor will only construct valid ship designs.  If the passed
-        in parameters (\p hull and \p parts) would result in an invalid design it throws
-        std::runtime_error. */
+
+    /**  The public ShipDesign constructor will only construct valid ship
+         designs, as long as the HullTypeManager has at least one hull. */
     ShipDesign(const std::string& name, const std::string& description,
-               const std::string& hull,
+               int designed_on_turn, int designed_by_empire, const std::string& hull,
                const std::vector<std::string>& parts,
                const std::string& icon, const std::string& model,
                bool name_desc_in_stringtable = false, bool monster = false,
@@ -542,8 +560,13 @@ private:
         MaybeInvalidDesign(const std::string& hull, const std::vector<std::string>& parts);
 
     /** Force design invariants to be true. If design invariants are not begin met provide an
-        explicit log message about how it was corrected. */
-    void MakeValid();
+        explicit log message about how it was corrected and throw
+        std::invalid_argument if \p should_throw is not none.
+
+        should_throw is not used but it is a literal reminder that
+        std::invalid_argument should be caught.
+    */
+    void ForceValidDesignOrThrow(const boost::optional<std::invalid_argument>& should_throw);
 
     void BuildStatCaches();
 
--- a/universe/ShipDesign.h
+++ b/universe/ShipDesign.h
@@ -11,6 +11,9 @@
 #include <vector>
 
 #include <boost/algorithm/string/case_conv.hpp>
+#include <boost/filesystem/operations.hpp>
+#include <boost/functional/hash.hpp>
+#include <boost/optional/optional.hpp>
 #include <boost/variant.hpp>
 #include <boost/serialization/access.hpp>
 #include <boost/serialization/nvp.hpp>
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Comment by LGM-Doyle** [comment]
Rebased.

**Comment by geoffthemedio** [comment]
What is the motivation for producing a valid design that is not what was intended instead of just giving an error or aborting in that situation?

**Comment by LGM-Doyle** [comment]
tl;dr;  The valid but not intended design means that modders can get graphical feedback in the design window of parsing errors, while at the same time guaranteeing all consumers a valid `ShipDesign`

The motivation is to split the difference between always having a correctly constructed `ShipDesign` and being as robust as possible to bad user (modder) input.  With a tertiary goal of providing the graphical feedback for mis-configured FOCS.

Currently, the `ShipDesign` constructor with incorrect input produces an invalid `ShipDesign` with a message in the log file.  There is the predicate `ValidDesign` that should be called before creating a `ShipDesign`, but there is no guarantee to consumers of `ShipDesigns` that they are valid.

One solution (the first that I pursued) is that the `ShipDesign` constructor could throw `std::invalid_argument` when the arguments are invalid.  This is correct C++ in that a constructor should not return an incorrectly configured object.  It also works fine if devs always remember to fairly locally catch `invalid_argument`.  However, it is likely this will result in non-local "`catch(...) //something weird happened here`" type code constructs.  

This PR's solution provides two constructors that only produce valid `ShipDesigns`.  One requires explicit mention of the exception as a reminder to catch it locally.  The other produces a valid, but not the intended design.

There is also the minor issue with loading and deserialization. If a `ShipDesign` fails to load/deserialize` then the entire load/deserialize needs to fail if any ships use(used?) that design.   This is only an issue loading old saves because this PR makes a short parts vectors an error and a few of the default designs had short parts vectors that are corrected in this PR.  The PR adding current and saved design managers will break loading old saves in any case, (unless you want that changed)

And lastly, there is graphical feedback in the design window of a bad parse.  For example, in the `ShipDesign` file format, there is no indication of which parts are external/internal/core.  By allowing a erroneous parse to succeed a modder can see in the design window which part was removed and hopefully surmise why.

I think it is important that the `ShipDesign` constructor only return valid constructed objects, which requires throwing on invalid input.  The non-throwing soft fail alternative I thought was an added benefit, but I don't think it essential.  If you want I can remove it.

**Comment by Dilvish-fo** [comment]
tl;dr for anyone other than LGM-Doyle is simply that I think the short-partslist log message should be of info type rather than error or warning type.

@LGM-Doyle you, however, I will ask to read my full set of comments as I wade through my thinking on the issues.  
My main (potential) concern is to be sure that the AI would not be getting fed back redacted designs if it tried out something that was not a valid design (it would be a nuisance to have to have the AI double check every design after creation).  It looks to me like the python exposure uses the throwing version and so the AI will get a failure code back if it tried an invalid design, but it would be appreciated if you could confirm that understanding for me.   Any redaction of preexisting scripted designs seems like it would not really be a problem for the AI since it will be assessing those designs based on the components they actually include.

I do have some concern (or at least I am not convinced) about at least one aspect of this PR from a player perspective, though.
>  this PR makes a short parts vectors an error 

Here you say "error", but as far as I can see in the code it is just a log warning.  I can see some significant value for the warning in the case of a truncated partslist, but in the case of the short parts list (which is not inherently any kind of error whatsoever and is quite often intentional) it seems to me that generating a warning creates more warning-clutter-problem than it creates any user benefit.  It seems to me that it should be of the "info" type log message, so that if someone adding hull slots has an easy way to know just which preexisting designs use that hull but haven't yet been adjusted, and otherwise they don't need to clutter up their log.  An empty partslot will be readily visible in the design window.

If I am misunderstanding the code and this is an actual error as per the comment:  I don't *see* any places where this would cause the AI trouble since it currently pads parts-lists with empty strings.  So although the possibility of trouble should it ever fail to do so is a bit bothersome, if there is any real player benefit then I guess that would be fine.  But...I am not understanding any significant player benefit from it being an error.  At most I come up with some starting portion of motivation about wanting to make sure that if someone adds a part slot to  a hull, then they are forced to necessarily go and update any and all designs that use that hull, because, um, ... well I seem to lose the line of reasoning before I reach an actual benefit, and so I am hoping I am correctly reading that a short partslist is not actually an error.

**Comment by LGM-Doyle** [comment]
I have made short part list back into a non-warning. (It was never an error.)  

@Dilvish-fo what do you mean by truncated vs. short?

When I said short I mean missing slots, not empty slots with no parts.

For example, the `Basic Large Hull`has four slots.  A warning is generated for a parsed ship design with only 1, 2, or 3 slots. ie ````parts = ["" "" ""]```` results in a warning.  Any design with 4 slots, even empty slots generates is good.  i.e. ````parts = ["" "" "" ""]```` is a good parts list.

This is the warning that I just removed, but I am unsure if you consider that a warning about a truncated or a short parts list. 


Info is not a lesser type of error/warning.  
Info/debug/trace are a continuum of verbosity logging regular operation.  
Error/warning indicates that something large/small is not right.  

Using info to indicate a little warning means that we need a third level of error condition.  We don't need a third level of error condition.  Using error to indicate, negatively impacts game performance and should not happen in released code; and warning won't impact game performance, and may be caused local conditions modded FOCS, disabled sound card etc. suffices.  

Having a clear distinction between error conditions and not error conditions means that testing with only errors and warning enabled highlights all problem issues.

There were two benefits to short parts list being a warning.

 For the modder if there are no warnings in the logs, then without knowing anything about a given hull the empty strings in the parts list represent all the empty slots.  Not a huge improvement, but when modding an unfamiliar syntax obvious placeholders are helpful.

For the parser and parts list handler it means an unambiguous parse/parts list length.  The code uses vector indexing so early correction of user input could possibly preempt bugs. 

This PR changes nothing in the AI creation of new designs.  In the existing code and this PR if the AI passes invalid design parameters it returns false and does not create the design.  

In addition there are no changes for players who are not modding files.

**Comment by LGM-Doyle** [comment]
@Dilvish-fo I changed the constructor so that when it throws it does not log.  There are no spurious log warnings when the computer uses the exceptions to determine if a design is viable.

All constructor variants still only produce valid designs unless there are no hulls in the `HullTypeManager`.

**Comment by Dilvish-fo** [comment]
Truncated vs short:  your code refers to truncating a provided too-long partslist, and padding a provided too-short partslist.

Thanks for your response, it sounds fine to me.

File: universe/ShipDesign.h [hunk 1]
```diff
@@ -520,15 +551,26 @@ class FO_COMMON_API ShipDesign {
...
-
+    /** Return true if \p hull and \p parts would make a valid design. */
+    static bool ValidDesign(const std::string& hull, const std::vector<std::string>& parts);
 private:
+    /** Return a valid hull and parts pair iff the \p hull and \p parts vectors
+        would not make a valid ShipDesign.
+        Also pad parts with "" if it is shorter than the \p hull number of slots.
+        Otherwise return none. If \p produce_log is true then produce log messages. */
+    static boost::optional<std::pair<std::string, std::vector<std::string>>>
+        MaybeInvalidDesign(const std::string& hull, std::vector<std::string>& parts, bool produce_log);
```
geoffthemedio, line 562: is parts intentionally not const& ?

LGM-Doyle: Yes.  That is what the "Also pad parts with \"\"" portion of the comment refers to.

Since we decided that a short parts vector is not a warning case, and if the return value is not none it indicates an error `MaybeInvalidDesign` needs a way to adjust parts (make it as long as the hull's number of slots).

LGM-Doyle:
I think that I have addressed all the concerns raised in this PR.  

Are there any further concerns with merging it?

PR #1602, opened 2017-06-08
Give Super Tester species perfect stealth

Morlic-fo:
For AI testing purposes, I often build super testers on turn 1, and keep passing turns while observing the AI behavior. When not playing the game, the AI will eventually conquer the planet. 

Giving the super testers perfect stealth means that the AI will not detect the human player and act as if it wasn't there.

### DIFF
--- a/default/scripting/species/SP_SUPERTEST.focs.txt
+++ b/default/scripting/species/SP_SUPERTEST.focs.txt
@@ -37,6 +37,10 @@ Species
             effects = SetDetection value = Value + 10000
 
         EffectsGroup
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Seems reasonable to me, but comment from AI dev would be appreciated...

LGM-Doyle:
👍 This is a good change.  I do the same thing and have had the same problem.

Dilvish-fo:
I agree totally.

Dilvish-fo (approved):

PR #1603, opened 2017-06-12
Fix bug in production queue projections

Dilvish-fo:
A bug introduced with the shift to percentage completion caused early termination of calculations relating to elements with a repeat count greater than 1, which would then often cause elements later in the queue to show substantially premature projected completion times (up to hundreds of turns difference).

A simple test case I used to compare results was starting up a single player game with mostly default stats, but no monsters, no AIs, and choosing Gysache.  I advanced for 5 turns with no other actions, to get the production meter to plateau, and then added 5x2 Scouts (5 repeats of a 2-unit block), 20x2 Outpost ships, 10x5 Outposts ships and another 10x5 Outpost ships.  Without this correction the projections for first completions show projected as happening after 2, 13, 73 and 117 turns; with this correction they are projected at 2, 15, 184 and 384 turns (which appear as accurate as I can reasonably confirm without deleting all ships as they are produced). 

Even if we are not too concerned actually about predictions of things 70+ turns in the future, a discrepancy of  ~15% already 13 turns out, which apparently can steadily increase up to 150% at ~75 turns out, seems like enough that I think if we do wind up doing a bugfix release for 0.4.7, that this should be included in that (should we have a label or milestone for that?)

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1118,7 +1118,11 @@ void ProductionQueue::Update() {
             std::tie(item_cost, build_turns) = queue_item_costs_and_times[key];
             float total_item_cost = item_cost * element.blocksize;
 
-            float additional_pp_to_complete_element = total_item_cost * (1.0f - element.progress); // additional PP, beyond already-accumulated PP to finish this element once
+            // calculate additional PP, beyond already-accumulated PP to finish all repeats of this element; 
+            // that is the total_item_cost times the number of repeats, less the current progress percentage times
+            // the total_item_cost, below reorganized as total_item_cost times the difference between the number
+            // of repeats and the percentage element.progress.
+            float additional_pp_to_complete_element = total_item_cost * (element.remaining - element.progress);
             //DebugLogger()  << " element total cost: " << element_total_cost << "; progress: " << element.progress;
             if (additional_pp_to_complete_element < EPSILON) {
                 //DebugLogger()  << "     will complete next turn";


Empire/Empire.cpp
```diff
@@ -1118,7 +1118,7 @@ void ProductionQueue::Update() {
             std::tie(item_cost, build_turns) = queue_item_costs_and_times[key];
             float total_item_cost = item_cost * element.blocksize;
 
-            float additional_pp_to_complete_element = total_item_cost * (1.0f - element.progress); // additional PP, beyond already-accumulated PP to finish this element once
+            float additional_pp_to_complete_element = total_item_cost * (element.remaining - element.progress); // additional PP, beyond already-accumulated PP to finish this element once
```
geoffthemedio: Can you add some comments before this line explaining why using `element.remaining` here makes sense? I think it does make sense, but conceptually the `remaining` integer count doesn't immediately seem to be in the same units as the float `progress`, so a note about how `progress` ranges from 0 to 1 and thus can be usefully subtracted from `remaining` would be helpful.

Dilvish-fo: I just noticed that the comment at the end of that line is also now mismatched to the calculation.  Howabout I just revert the whole line to its status prior to c6cc24013:
```
float additional_pp_to_complete_element = total_item_cost * element.remaining - element.progress*total_item_cost; // additional PP, beyond already-accumulated PP, to build all items in this element
```

Since you had changed both the calculation and the comment in that commit, it looks like perhaps you had been considering changing the way that additional_pp_to_complete_element would be put to use, but the change slipped into the grooming commit before you made the followup change to how the variable was actually being used.

One thing to keep in mind: I've been working on reorganizing this whole method, in prep for the Imperial Stockpile, and I've concluded that in this whole subsection section about figuring out the smallest possible max_turns for the following loop, and using additional_pp_to_complete_element to maybe avoid the loop altogether, I really just added a fair bit of code complexity for practically no real benefit.  Most of this whole subsection is eliminated in a cleanup PR that I will probably be submitting later today, but I wanted to get in a simple fix for this bug as a standalone commit so that it would be most suitable to include in any bugfix release we might wind up doing for 0.4.7.

Dilvish-fo: I just realized, perhaps part of the issue had been that the original comment reference to ```build all items``` was, by itself, a bit ambiguous about whether it meant build just the entire block of items, or build all repeats of the block (it was meant to be the latter, and I will double check that I am being more clear  in similar cases in the later PR I submit).   So I should also amend the comment to make that more clear.

geoffthemedio: I prefer it with `total_item_cost` factored out

Dilvish-fo:
OK, I've augmented the comment and pushed a replacement commit.

Vezzra:
> should we have a label or milestone for that?

I think a milestone (something like "0.4.7.1 bugfix release") should fit the bill. [I've opened a thread](http://freeorion.org/forum/viewtopic.php?f=4&t=10623) to discuss the suggestion of making a bugfix release and a proper procedure for that.

PR #1605, opened 2017-06-15
Production queue reorganization

Dilvish-fo:
This PR is for a  modest bit of cleanup and reorganization of ProductionQueue::Update() and some related code.  The primary intent is to better prepare this code for implementation of an Imperial Stockpile, for which I am also about to submit a PR, but the Imperial Stockpile may possibly linger for a while as a work in progress, whereas I think this reorganization is suitable to go ahead and put into master.

PR #1606, opened 2017-06-15
Production queue drop dynamic programming for projections

Dilvish-fo:
This PR is intermediate between #1605 and the Imperial Stockpile PR I am about to submit, and I am really just posting this for reference and to possibly facilitate some testing and a somewhat more specific discussion of the changes here separately from the other changes that appear with the Imperial Stockpile; I do not expect to merge this PR apart from the Imperial Stockpile PR.

The approach that we had been using for ProductionQueue projections for the past several years is one I considered to be a sort of dynamic programming approach, and it was very efficient.  Unfortunately, at least in its current form it relies upon some constraints that would be violated by an Imperial Stockpile.  So, here I reorganize the projections code to put it back more akin to what it had been prior to the dynamic programming approach, but with a greater emphasis on caching and reusing the results of all Evals of the current universe state.  The resulting code is much more simple and streamlined and can readily be adapted to work with an Imperial Stockpile.  By itself, even without the stockpile, this change does however cause a slowdown in the projections code by a factor of some 20-fold to 40-fold.

In my initial testing the total time taken was still well below the cap of half a second, but that was also with a fairly modest amount of PP being handled.  I am hoping that some other players who can compile would be willing to test this out; save games for current master (after the recent addition of production queue splitting/duplicating) should be fully compatible with this PR, so you have an advanced game you could compare the ProductionQueue processing time between master and this PR.

Besides simply comparing any noticeable delay, you can directly check the processing time by searching freeorion.log for the phrase 
```ProductionQueue::Update: Projections took ```

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -177,14 +177,15 @@ namespace {
       * system groups that are able to exchange resources with the build
       * location and the amount of minerals and industry produced in the group).
       * Elements will not receive funding if they cannot be produced by the
-      * empire with the indicated \a empire_id this turn at their build location. */
+      * empire with the indicated \a empire_id this turn at their build location. 
+      * Also checks if elements will be completed this turn. */
     void SetProdQueueElementSpending(std::map<std::set<int>, float> available_pp,
                                      const std::vector<std::set<int>>& queue_element_resource_sharing_object_groups,
                                      const std::map<std::pair<ProductionQueue::ProductionItem, int>, std::pair<float, int>>& queue_item_costs_and_times,
                                      const std::vector<bool>& is_producible,
                                      ProductionQueue::QueueType& queue,
                                      std::map<std::set<int>, float>& allocated_pp,
-                                     int& projects_in_progress)
+                                     int& projects_in_progress, bool simulating)
     {
         //DebugLogger() << "========SetProdQueueElementSpending========";
         //DebugLogger() << "production status: ";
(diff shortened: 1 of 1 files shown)


LGM-Doyle:
This makes me sad, 😢 .

The dynamic programming queue prediction was elegant code that made me happy, 😄, to have read it.

Dilvish-fo:
> The dynamic programming queue prediction was elegant code that made me happy, :smile:, to have read it.

Thanks, that is really great to hear; from a coding perspective that's my code I am most proud of here.   I think it is almost definite that at the very least I'll at least keep the dynamic programming approach in the mixed mode that I mention in the Imperial Stockpile PR.  I mostly just really wanted to get a working version of the Imperial Stockpile up and running and figured I'd go back and work on efficiency afterwards.

I also was a bit caught by surprise at just how much difference this dynamic programming approach still made.  As I was working through things for the stockpile I started thinking that perhaps a significant part of the time savings from the DP approach had been that it allowed the time/cost determinations to be made only once for the entire 500 turns of projections, and so that perhaps caching that info might let the non-DP approach do nearly as well. Although it's currently looking like the core DP approach itself accounted for the great bulk of time savings, I'm going to continue noodling through this to see what else can possibly be worked out to further reduce duplicating data structures etc.  I am a bit concerned that in larger games with many well developed empires, that the time consumed here could get to be really noticeable.

Dilvish-fo:
since I had cleaned-up, rebased and merged #1605 on which this depended, I've now rebased this to master.  (Keep in mind, though, this PR is really just for reference & potentially more focused comment, as a stepping stone towards the Imperial Stockpile PR #1607, I do not expect to merge this PR apart from the Imperial Stockpile PR.)

Vezzra:
@Dilvish-fo, can you post a reference to the superseding PR here?

PR #1609, opened 2017-06-18
Store the AIstate version as instance attribute

Morlic-fo:
Hopefully resolves the bug reported in this thread:
http://freeorion.org/forum/viewtopic.php?f=25&t=10618#p89192


MatGB:
Working for me as mentioned in forum, recommend expediting merge as master is broken without it.

PR #1611, opened 2017-06-20
Game Rules

geoffthemedio:
Adds game rules manager and UI support.

Rules are stuff that can be configured about how the game works, eg. number of combat rounds, if experimentors are generated, if all techs should cost 50% more than default, etc.


Vezzra:
Excellent idea! 😃

But can you add a brief description of what these game rules are supposed to be to the OP? Although looking over the changes gives a pretty good idea, that shouldn't be necessary.

geoffthemedio:
The particular rules included here are arbitrary... not the main point of the pull request. I will also add more, and additional mechanisms for specifying rules, including from within Python or FOCS.

Vezzra:
I understand, but still... some brief explanation of the basic idea/concept behind these "game rules" would be nice. When I first saw that PR my immediate reaction has been: what's that? The "achievements" [you've proposed on the forum](http://freeorion.org/forum/viewtopic.php?f=6&t=10597)? Or something completely different?

Two or three examples of what kind of game rules you have in mind would probably be sufficient to give a good basic idea.

Just my 2c, it's up to you.

util/MultiplayerCommon.h
```diff
@@ -20,6 +21,94 @@ FO_COMMON_API extern const std::string SP_SAVE_FILE_EXTENSION;
...
+    void    ClearExternalRules();
+
+    /** Resets all rules to default values. */
+    void    ResetToDefaults();
+    //@}
+
+private:
+    GameRules();
+
+    std::map<std::string, Rule> m_game_rules;
```
LGM-Doyle: Use `std::unordered_map`.  

There is no reason to use `std::map`, it is slower and the primary use of this table is not to sort the rules by name.  If an alphabetically sorted list is needed for UI purposes, then do it in-place.

LGM-Doyle (commented):
Great addition.

It works for me.  All techs, buildings and ships can be researched for 1 point and in 1 turn.

Dilvish-fo:
a minor point, but the python interface is organized with a single file for all the enum exposures-- EnumWrapper.cpp

Dilvish-fo:
Also, more broadly, I'd like to note that I share Vezzra's and LGM-Doyle's reaction that this is a great idea.

geoffthemedio:
@Dilvish-fo I'm aware, and debated putting the enum in https://github.com/freeorion/freeorion/blob/master/python/EnumWrapper.cpp though noted that that has just a `WrapGameStateEnums` function, and wasn't sure if a game rule type is really part of the gamestate.

geoffthemedio:
This seems to be working more or less as intended now. Rules can be scripted in C++ or parsed definition files. They appear in the UI and can be set. C++, FOCS, or the Python API can access them.

Further comments?

util/MultiplayerCommon.h
```diff
@@ -20,6 +21,132 @@ FO_COMMON_API extern const std::string SP_SAVE_FILE_EXTENSION;
...
+  * "bool unused_bool = RegisterGameRules(&foo)"). */
+FO_COMMON_API bool RegisterGameRules(GameRulesFn function);
+
+/** returns the single instance of the GameRules class */
+FO_COMMON_API GameRules& GetGameRules();
+
+/** Database of values that control how the game mechanics function. */
+class FO_COMMON_API GameRules {
+public:
+    enum RuleType : int {
```
LGM-Doyle, line 44: What is the purpose of this enumeration?

LGM-Doyle (commented):
Thanks for making the previous changes to use `unordered_map`.

As usual the code looks good.  It compiles, changes production, research times and duration of combat as expected on linux.

I've added some questions about the usage of `engine_internal` and the enumeration of the rule types.

util/MultiplayerCommon.h
geoffthemedio: it tracks the type of each rule... I thought the name and enum values were self explanatory...

UI/GalaxySetupWnd.h
```diff
@@ -100,18 +144,18 @@ class GalaxySetupWnd : public CUIWnd {
     bool                    EndedWithOk() const {return m_done && m_ended_with_ok;}
 
     /** returns the panel containing all the user-chosen options. */
-    const GalaxySetupPanel& Panel()      const  {return *m_galaxy_setup_panel;}
+    //const GalaxySetupPanel& Panel()      const  {return *m_galaxy_setup_panel;}
```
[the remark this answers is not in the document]
geoffthemedio: forgot

util/MultiplayerCommon.h
```diff
@@ -20,6 +21,132 @@ FO_COMMON_API extern const std::string SP_SAVE_FILE_EXTENSION;
...
+    RuleType RuleTypeForType(double dummy)
+    { return DOUBLE; }
+    RuleType RuleTypeForType(std::string dummy)
+    { return STRING; }
+
+    struct FO_COMMON_API Rule : public OptionsDB::Option {
+        Rule();
+        Rule(RuleType rule_type_, const std::string& name_, const boost::any& value_,
+             const boost::any& default_value_, const std::string& description_,
+             const ValidatorBase *validator_, bool engine_internal_,
```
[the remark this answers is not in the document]
geoffthemedio: It tracks whether a rule is defined internally, ie. in C++ code, or externally, from a parsed rules definition. The C++ code knows that combat rounds is defined in itself (C++ code) so doesn't need to check that this is the case. Original purpose was to put the internal vs. scripted rules in separate tabs, but that has been replaced with an explicit category field. Now it's just tracking where they are defined, which isn't used anywhere, but probably will be incorporated into pedia listing of rules.

util/MultiplayerCommon.h
```diff
@@ -20,6 +21,132 @@ FO_COMMON_API extern const std::string SP_SAVE_FILE_EXTENSION;
...
+        if (it == m_game_rules.end())
+            throw std::runtime_error("GameRules::Set<>() : Attempted to set nonexistent rule \"" + name + "\".");
+        it->second.SetFromValue(value);
+    }
+
+    void    SetFromStrings(const std::vector<std::pair<std::string, std::string>>& names_values);
+
+    /** Removes game rules that were added without being specified as
+        engine internal. */
+    void    ClearExternalRules();
```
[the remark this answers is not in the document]
geoffthemedio: No, but it will be helpful when it becomes possible to change what rules are supported by changing the selected content directory.

util/MultiplayerCommon.h
geoffthemedio: to elaborate, `Rule` is derived from `OptionsDB::Option` which internally uses `boost::any` to store the values. Options are all defined / registered in C++ code, so the type of the options are all known in the same C++ that gets their value as needed by calling `OptionsDB::Get<T>`, but `Rule` can also be created from definitions in script files, and thus can store one of several types, with no (other) way to know what the type is by inspecting a `Rule`. The enum is thus used to track what type is stored in each `Rule` so it's known what type to `Get`. This is useful within C++ code that otherwise wouldn't know the type stored in a `Rule`, and within the Python API to provide some type-checks when rule values are requested. It's not actually used in the FOCS API, which just does try-catch around the `Get` calls, though in principle could be as well.

Vezzra (approved):
> Further comments?

I've pushed an update of the Xcode project, builds and runs fine with that on OSX.

As far as I'm concerned, LGTM.

PR #1612, opened 2017-06-21
Return file path along with parsed ship design in ShipDesignsParser.

LGM-Doyle:
The file path will allow DesignWnd to update saved designs when players
change their designs.


geoffthemedio:
didn't test, but no objections

PR #1614, opened 2017-06-23
Monster nest eradication

TualhaFO:
New building that removes a monster nest. See [discussion on forums](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10627). Please review and test.


MatGB:
Apart from the way too high costs as mentioned on the forum, it needs a sitrep priority set for the sitreps, I'd suggest at or about line 60
[freeorion/common_user_customizations.txt at master · freeorion/freeorion](https://github.com/freeorion/freeorion/blob/master/default/customizations/common_user_customizations.txt#L60)

TualhaFO:
Hmm. Experimentation shows that a monster can hatch just _before_ the nest is removed. Guess it depends on the order in which effects are applied — don’t know where that’s controlled. To mirror that order, perhaps this effect message should come right after tame monster messages, say line 87?

I have committed and pushed that.

dbenage-cx:
@TualhaFO Could adjust the `priority` of removal effect in `NEST_ERADICATOR` and/or the creation effects in [specials/monster_nest.macros](https://github.com/freeorion/freeorion/blob/master/default/scripting/specials/monster_nest.macros).

TualhaFO:
This is weird … I tried various things, including setting priority to 1 for nest removal and 200 for monster creation, but I still got monsters popping before the nests were removed (on the same turn).

This was in 0.4.6; I haven’t tried to build the latest yet, as I lack gcc 5.0 and whatever version of Boost is needed. Were priorities working correctly in 0.4.6?

Dilvish-fo:
I don't think that priorities can help with the issue of a monster appearing on the same turn as the nest is removed -- priorities have to do with the order that effects are applied, but my recollection prompts me that all scope and activation determinations are made prior to any Effects being applied.  This seems to be an area where our Scripting/Effects writeup is still incomplete, but I do find apparent confirmation in [the code comment](https://github.com/freeorion/freeorion/blob/f0d71bd8f82e6ec568de821c012b633409b26bda/universe/Universe.cpp#L445) 
```
    // cache all activation and scoping condition results before applying Effects, since the application of
    // these Effects may affect the activation and scoping evaluations
```
So, the option that comes to mind, other than just accepting the current behavior, would be to have the nest special check for the presence of the building.

TualhaFO:
Aha! That makes sense. Thank you, I will investigate that approach.

I’ve built gcc 5.4 now, but haven’t had time to build Boost yet. So much to do, so little time …

geoffthemedio:
@MatGB verdict on this, despite the apparent effects infrastructure issue?

TualhaFO:
Hi folks, sorry for not writing earlier … my life has gotten crazy busy, I really can’t work on this patch for the foreseeable future. Hope someone likes it enough to take it over.

dbenage-cx:
@TualhaFO Thanks for posting a status update.

Posted a version [here](https://github.com/dbenage-cx/freeorion/tree/ref_tualhafo-1614_nest-eradication) with an [additional commit](https://github.com/dbenage-cx/freeorion/commit/4b00cbad3ea565bea34717b4ba5f041cc0ad68b6) for proposed fix (branch is rebased to master).

In minor testing setting nest spawn probability to 1.0, not seeing any spawns on same turn as removal.

Vezzra:
@dbenage-cx, can you create a PR with your fixed version of this PR? Then we can close this one here and continue with yours (seeing as @TualhaFO won't have time to work on theirs).

dbenage-cx:
Superceeded by #1799

PR #1615, opened 2017-06-25
Move Phototrophic effects evaluation after all other effects

Morlic-fo:
Because the phototrophic effect activates only with a target population
greater than zero, it depends on the order of effect evaluations that
modify the target population.

Prior to this PR, there were effects that modify the target population
with the same priority as the phototrophic trait which is ill-defined
behavior.

This PR adresses the issue by ensuring that the Phototrophic trait is always evaluated after all other effects where applied.


MatGB:
OK, appears I must've messed this up a bit when I set the priorities, glad we've caught it. Very Late was supposed to be just growth specials or self-sustaining/phototroph stuff so in theory it wouldn't clash.

Also, I had meant and thought I had removed all the population > 0 effects, they never worked properly anyway, but if this fixes it we're good.

Morlic-fo:
I am not sure if this actually fixes anything or if the previous implementation happened to get the same results by relying on some implicit order of effect evaluation.

This is mostly an observation from reading the scripts and without test cases.

default/scripting/common/priorities.macros
```diff
@@ -15,3 +15,5 @@ LATE_PRIORITY
 VERY_LATE_PRIORITY
 '''120'''
 
+LATEST_GROWTH_PRIORITY
+'''99999'''
```
dbenage-cx: add newline

dbenage-cx (changes requested):

dbenage-cx (approved):

default/scripting/common/priorities.macros
Morlic-fo: Fixed.

dbenage-cx: Was too hasty clicking approved, looks like other file has the added EOL but not this one (or browser is not updating)

geoffthemedio:
@MatGB @dbenage-cx @Morlic-fo any further concerns / comments?

MatGB:
Nope, LGTM

dbenage-cx:
Not had time to test, otherwise LGTM

PR #1616, opened 2017-06-25
add icon to indicate if a fleet is blockaded

TheSilentOne1:
This PR adds an icon that indicates if a fleet is being blockaded, as discussed in this thread: http://www.freeorion.org/forum/viewtopic.php?f=6&t=10566

For now, I chose an icon instead of a perimeter border. In the screenshot below the monster is blockaded.

block


UI/FleetButton.cpp
```diff
@@ -172,13 +172,28 @@ FleetButton::FleetButton(const std::vector<int>& fleet_IDs, SizeType size_type)
 
     // select icon(s) for fleet(s)
     int num_ships = 0;
+    m_fleet_blockaded = false;
     for (std::shared_ptr<const Fleet> fleet : fleets) {
-        if (fleet)
+        if (fleet) {
             num_ships += fleet->NumShips();
+            if (fleet->Blockaded())
+                m_fleet_blockaded = true;
```
geoffthemedio: can presumably break out of the loop in the event a fleet is found, as `m_fleet_blockaded` can't get any more or less `true` than it is made here...

UI/FleetButton.cpp
```diff
@@ -172,13 +172,28 @@ FleetButton::FleetButton(const std::vector<int>& fleet_IDs, SizeType size_type)
...
+        }
+    }  
+
+    // add RotatingGraphics for all icons needed
+    if (m_fleet_blockaded) {
+        std::shared_ptr<GG::Texture> blockaded_texture = FleetBlockadedIcon(size_type);
+        RotatingGraphic* icon = new RotatingGraphic(blockaded_texture, GG::GRAPHIC_FITGRAPHIC | GG::GRAPHIC_PROPSCALE);
+        icon->SetRPM(0.0f);
```
geoffthemedio: if it's not rotating, why do you use a `RotatingGraphic` instead of a `StaticGraphic` ?

UI/FleetButton.cpp
```diff
@@ -436,6 +456,24 @@ std::shared_ptr<GG::Texture> FleetSizeIcon(unsigned int fleet_size, FleetButton:
...
 
+std::shared_ptr<GG::Texture> FleetBlockadedIcon(FleetButton::SizeType size_type) {
+    if (size_type == FleetButton::FLEET_BUTTON_NONE)
+        return nullptr;
+
+    std::string size_prefix = FleetIconSizePrefix(size_type);
+    if (size_type == FleetButton::FLEET_BUTTON_TINY)
+        size_prefix = "tiny-";
+
+    static std::shared_ptr<GG::Texture> retval;
```
geoffthemedio: why `static` ?

UI/FleetButton.cpp
TheSilentOne1: Num_ships is still being calculated.

universe/Fleet.cpp
```diff
@@ -1156,6 +1156,23 @@ void Fleet::CalculateRouteTo(int target_system_id) {
     }
 }
 
+bool Fleet::Blockaded() const {
+    /** returns true if fleet is in a system and any exit is blocked */
```
geoffthemedio: explain what being blockaded means in this context? presumably is that it is prevented from using at least one starlane entry to leave to leave its current system?

universe/Fleet.cpp
```diff
@@ -1156,6 +1156,23 @@ void Fleet::CalculateRouteTo(int target_system_id) {
...
+    std::shared_ptr<System> system = GetSystem(SystemID());
+
+    if (!system)
+        return false;
+
+    const std::map<int, bool> adjacent_system_IDs = system->VisibleStarlanesWormholes(Owner());
+
+    for (std::pair<int, bool> target_system : adjacent_system_IDs) {
```
geoffthemedio: `const auto&`

geoffthemedio:
If you're going to indicate this with an icon, then an explanation of what it means with a tooltip, perhaps specifying which destinations it can move towards or can't move towards would be preferable.

UI/FleetButton.cpp
TheSilentOne1: m_icons is a std::vector<RotatingGraphic*>.

UI/FleetButton.cpp
geoffthemedio: maybe make it `if (!m_fleet_blockaded && fleet->Blockaded())` then, since `Fleet::Blockaded` is a bit expensive looking...

TheSilentOne1:
Thanks for reviewing, I've added an explanatory tooltip and applied your other suggestions & corrections.
There's still a bug which I haven't been able to figure out: the blockade indicator seems to show for monsters and the player fleets as intended, but not for AI players. Fleet->Blockaded() seems to always return false for them. Any idea what might cause this?

geoffthemedio:
Is information about other empires' fleets being bloackaded sent to players?

UI/FleetButton.cpp
```diff
@@ -291,6 +313,40 @@ void FleetButton::LayoutIcons() {
...
+    if (m_fleet_blockaded) {
+        std::shared_ptr<Fleet> fleet;
+        std::shared_ptr<System> current_system;
+        std::map<int, bool> adjacent_system_IDs;
+        std::string available_exits = "";
+        int available_exits_count = 0;
+
+        if (!m_fleets.empty())
+            // can just pick first fleet because all fleets in system should have same exits
+            fleet = GetFleet(*m_fleets.begin());
```
LGM-Doyle, line 327: Is this true?
I thought each blockaded non-monster fleet could only exit along its entry path?

LGM-Doyle (commented):
I came to see what the icon looks like.  I like the icon. 

I left some review comments.

UI/FleetButton.cpp
TheSilentOne1: All fleets grouped under a fleet button should have the same exits available. If an armed fleet arrives from a different starlane, it will be opened as an exit for all fleets.

TheSilentOne1: On GG::Rotating/StaticGraphic: as there doesn't seem to be a good reason why the fb shouldn't be a StaticGraphic, I will change it as Geoff and you suggested.

TheSilentOne1:
> Is information about other empires' fleets being bloackaded sent to players?

Hm, that might be it; it's curious that it does seem to work for monsters though.

UI/FleetButton.cpp
TheSilentOne1: Okay will change.

universe/Fleet.cpp
```diff
@@ -1156,6 +1156,24 @@ void Fleet::CalculateRouteTo(int target_system_id) {
...
 
+bool Fleet::Blockaded() const {
+    /** returns true if fleet is in a system and is prevented from using at least
+        one starlane entry to leave its current system */
+    std::shared_ptr<System> system = GetSystem(this->SystemID());
+
+    if (!system)
+        return false;
+
+    const std::map<int, bool> adjacent_system_IDs = system->VisibleStarlanesWormholes(this->Owner());
```
Dilvish-fo: > There's still a bug which I haven't been able to figure out: the blockade indicator seems to show for monsters and the player fleets as intended, but not for AI players. Fleet->Blockaded() seems to always return false for them. Any idea what might cause this?

I am pretty sure that part, perhaps all, of the problem is from this loop based on VisibleStarlanesWormholes() winding up empty for any empire other than your own or ALL_Empires, because VisibleStarlanesWormholes() relies on Universe::GetObjectVisibilityByEmpire, which for empireID ALL_EMPIRES will always return VIS_FULL_VISIBILITY, but for any other empire it checks visibility via the m_empire_object_visibility maps, which in the serialization process for an empire update only populates it via Universe::GetEmpireObjectVisibilityMap() with that empire's info (i.e., you don't get visibility info for other empires).

I think the solution is to not use VisibleStarlanesWormholes() at all.  The solution I would propose is to just base the loop on System::StarlanesWormholes() instead.  If you cant see the enemy fleet being blockaded along any starlane of which ___you___ are aware (and currently you would normally be aware of all of them if you can see a fleet there anyways), then consider it not blockaded.

TheSilentOne1: Thank you, that fixed it, and also thank you that you took the time. I will update the PR soon.

TheSilentOne1:
Updated and squashed the commits. Light testing suggests everything works now as intended.

UI/FleetButton.cpp
Dilvish-fo: **edit** just realized Doyle said "non-monster", I previously thought he was talking about monster fleets.  I think I need more coffee, hehe.  So, I'm not sure how much the rest of my comment here matters, but I'll leave it for @TheSilentOne1 to consider

 The thing about opening up extra lanes as exits is tracked at an empire level, within the empire object itself, and it's my recollection that we left it that monsters would not share exits.

UI/FleetButton.cpp
```diff
@@ -172,13 +172,29 @@ FleetButton::FleetButton(const std::vector<int>& fleet_IDs, SizeType size_type)
...
+                m_fleet_blockaded = true;
+        }
+    }  
+
+    // add graphics for all icons needed
+    if (m_fleet_blockaded) {
+        std::shared_ptr<GG::Texture> blockaded_texture = FleetBlockadedIcon(size_type);
+        if (blockaded_texture) {
+            GG::StaticGraphic* icon = new GG::StaticGraphic(blockaded_texture, GG::GRAPHIC_FITGRAPHIC | GG::GRAPHIC_PROPSCALE);
+            GG::Clr opposite_clr(255 - this->Color().r, 255 - this->Color().g, 255 - this->Color().b, this->Color().a);
```
geoffthemedio, line 189: there isn't some function to find a reverse colour being used for starlane wasted PP indication that could be used here as well? (haven't checked... am just wondering)

UI/FleetButton.cpp
```diff
@@ -291,6 +314,40 @@ void FleetButton::LayoutIcons() {
...
+        if (!m_fleets.empty())
+            // can just pick first fleet because all fleets in system should have same exits
+            fleet = GetFleet(*m_fleets.begin());
+        else return;
+
+        current_system = GetSystem(fleet->SystemID());
+        adjacent_system_IDs = current_system->StarlanesWormholes();
+
+        for (const auto& target_system_id : adjacent_system_IDs) {
+            if (!fleet->BlockadedAtSystem(fleet->SystemID(), target_system_id.first)) {
```
geoffthemedio: prefer `if (fleet->Blockaded(...)) continue;` so the following can be de-indented

UI/FleetButton.cpp
```diff
@@ -436,6 +493,23 @@ std::shared_ptr<GG::Texture> FleetSizeIcon(unsigned int fleet_size, FleetButton:
...
+std::shared_ptr<GG::Texture> FleetBlockadedIcon(FleetButton::SizeType size_type) {
+    if (size_type == FleetButton::FLEET_BUTTON_NONE)
+        return nullptr;
+
+    std::string size_prefix = FleetIconSizePrefix(size_type);
+    if (size_type == FleetButton::FLEET_BUTTON_TINY)
+        size_prefix = "tiny-";
+
+    std::shared_ptr<GG::Texture> retval = ClientUI::GetClientUI()->GetTexture(ClientUI::ArtDir() / "icons" / "fleet" / (size_prefix + "blockade.png"), false);
+    glBindTexture(GL_TEXTURE_2D, retval->OpenGLId());
```
geoffthemedio, line 506: is this really necessary to do every time the texture is gotten? not just once? or better done by the calling code that uses the returned texture?

geoffthemedio:
@TheSilentOne1 plan to return to this?

TheSilentOne1:
Sorry, I hope to pick this up sometime next week. And after that I'll be active on a more regular basis again.

UI/FleetButton.cpp
TheSilentOne1: I don't think so. The starlane color is also computed by (255 - old_color) when PP are wasted.

UI/FleetButton.cpp
TheSilentOne1: I can't really answer that. I reproduced how the textures for the other fleet icons are loaded (FleetHeadIcons(), FleetSizeIcon(), FleetSelectionIndicatorIcon()).

UI/FleetButton.cpp
TheSilentOne1: Sorry, I've tried to find the right place in the code, but I'm afraid it is beyond me. However, I have checked it in game, and non-monster agressive fleets do indeed share exits. I moved in on a hostile homeworld from to sides, and was able to retreat to either systems I was coming from, but not to a third adjacent system.

UI/FleetButton.cpp
```diff
@@ -291,6 +314,41 @@ void FleetButton::LayoutIcons() {
...
+                available_exits_count++;
+            }
+        }
+
+        if (available_exits_count >= 1)
+            SetBrowseText(UserString("FB_TOOLTIP_BLOCKADE_WITH_EXIT") + available_exits);
+        else
+            SetBrowseText(UserString("FB_TOOLTIP_BLOCKADE_NO_EXIT"));
+    } else {
+        SetBrowseText("");
```
geoffthemedio: does this do something different from `ClearBrowseInfoWnd`?

TheSilentOne1: Guess not. I've replaced it.

geoffthemedio:
I don't think I understand what the blockade indicator is showing. See these screenshots...

Initially, my fleet (argressive, armed), is moving past Lemaitre, where a maintenance ship is located.
before
If I let the move go through, with the ship moving to the following system, then it stops in the system with the ship, but is not blockaded. I would have expected it to be.
after_continuing
If I instead order the ship to move to Lemaitre, rather than the following system, it ends up blockading the maintenance ship... which doesn't make sense to me at all... why would the stationary ship be blockaded by an arriving ship?
after_stopped

Also, the tooltip for a blockaded fleet seems misleading... Can't a blockaded fleet move back on the lane it arrive from, if any? And could the tooltip indicate which lanes a blockaded fleet can or can't move on?
blockaded_fleet_tooltip

[blockade.zip](https://github.com/freeorion/freeorion/files/1172867/blockade.zip)

TheSilentOne1:
I _think_ everything is working as it should; it's the pedia description that doesn't have the special case "blockade with monsters" right yet.
When your fleet arrives at Lemaitre, the feet is blockaded / stopped and an initial combat occurs. After the combat, the monster blockade is lifted, and, if your fleet stays, it will be able to blockade the monster. However, if it leaves no blockade will be created, so no marker. (Also, as a special case, blockaded monsters are unable to move.)
The behaviour is different for empire fleets. @Dilvish-fo, please correct me if I'm wrong.

geoffthemedio:
That sounds reasonable, though the special rules for monsters should be addressed in the tooltip.

MatGB:
Yeah, monsters are intentionally easy to pin down to enable a player to eliminate them, especially important for the very slow ones. The engine doesn't really know the difference between moving and fixed monsters, perhaps it should?

Regardless, apart from the tool tip everything in the pics is as it should be.

TheSilentOne1:
I've modified the tooltip for monsters so that it now reads: 

> This monster fleet is blockaded by an empire fleet. When monsters are blockaded by empire fleets, they are prevented from moving.

UI/FleetButton.cpp
```diff
@@ -342,10 +342,12 @@ void FleetButton::LayoutIcons() {
...
         }
 
-        if (available_exits_count >= 1)
+        if (fleet->Owner() == ALL_EMPIRES)  // as above, if first fleet of fleet-button is "monster-owned", all fleets are
+            SetBrowseText(UserString("FB_TOOLTIP_BLOCKADE_MONSTER"));
+        else if (available_exits_count >= 1)
             SetBrowseText(UserString("FB_TOOLTIP_BLOCKADE_WITH_EXIT") + available_exits);
         else
-            SetBrowseText(UserString("FB_TOOLTIP_BLOCKADE_NO_EXIT"));
+            SetBrowseText(UserString("FB_TOOLTIP_BLOCKADE_NO_EXIT")); // this currently shouldn't happen, empire ships should always have an exit
```
geoffthemedio: What if a ship is produced (or moved somewhere by an effect, not using a starlane) into a system that already contains a hostile empire fleet?

TheSilentOne1: Hm, that may indeed cause the ship to have no available exit, though I'm not 100% sure. Okay if I just remove the comment?

geoffthemedio: I suppose? Was just wondering...

TheSilentOne1: For now, I've removed the comment, but I've made a note of it and I'll report when I was able to test if a fleet can indeed have no exits if they are newly produced or arrive by star gate etc..

universe/Fleet.cpp
```diff
@@ -1156,6 +1156,24 @@ void Fleet::CalculateRouteTo(int target_system_id) {
...
 
+bool Fleet::Blockaded() const {
+    /** returns true if fleet is in a system and is prevented from using at least
+        one starlane entry to leave its current system */
+    std::shared_ptr<System> system = GetSystem(this->SystemID());
+
+    if (!system)
+        return false;
+
+    const std::map<int, bool> adjacent_system_IDs = system->StarlanesWormholes();
```
geoffthemedio: should store as a const reference... or better yet, just do
`for (const auto& target_system : system->StarlanesWormholes())`

UI/FleetButton.cpp
```diff
@@ -291,6 +314,43 @@ void FleetButton::LayoutIcons() {
...
+        std::string available_exits = "";
+        int available_exits_count = 0;
+
+        if (!m_fleets.empty())
+            // can just pick first fleet because all fleets in system should have same exits
+            fleet = GetFleet(*m_fleets.begin());
+        else return;
+
+        current_system = GetSystem(fleet->SystemID());
+        adjacent_system_IDs = current_system->StarlanesWormholes();
```
geoffthemedio: same here

MatGB:
Fleets that are newly built into a system that's blockaded are meant to be themselves blockaded as per last discussion on the subject, but there may be a bug that's allowing them to escape as I'm fairly sure but not certain I've seen it—there was some refactoring of the code here at some point relatively recently that changed some behaviour and we never did narrow down whether everything worked in all cases.

I have zero clue about stargates, if I'm porting a fleet into a location of mine blockaded it's to fight off the blockade, but I'll keep an eye out for it, I think they *should* be stuck

TheSilentOne1:
That's good to know, and with regard to this discussion that would mean fleets can be trapped, and that there's a reason for a special tooltip for this case. So I'll leave it there.

MatGB:
Aside: have just confirmed that newly built ships can escape, and this counts as a bug, if you're able to track it down while working on this cool otherwise we'll need an issue opened.

TheSilentOne1:
I'd rather not meddle with the blockade code, so I've created an issue here: https://github.com/freeorion/freeorion/issues/1648

PR #1621, opened 2017-06-29
Limit engine parts to one per ship, increase stats

dbenage-cx:
Restricts existing engine parts to one per design.
Increases capacity(speed) and adjusts build cost and time.
[Forum discussion](http://freeorion.org/forum/viewtopic.php?f=15&t=10540)

### DIFF
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -850,7 +850,7 @@ class ShipDesigner(object):
         self.production_time = local_time_cache.get(self.hull.name, self.hull.productionTime(fo.empireID(), self.pid))
 
         # read out part stats
-        shield_counter = cloak_counter = detection_counter = colonization_counter = 0  # to deal with Non-stacking parts
+        shield_counter = cloak_counter = detection_counter = colonization_counter = engine_counter = 0  # to deal with Non-stacking parts
         hangar_parts = set()
         for part in self.parts:
             self.production_cost += local_cost_cache.get(part.name, part.productionCost(fo.empireID(), self.pid))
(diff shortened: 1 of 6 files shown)


dbenage-cx:
Initial stat increases:

Part | Slots | Speed | Build Cost | Build Time
----- | ----- | ----- | ----- | -----
`Improved Engine Couplings` | internal | 10->*__20__* | 10->*__15__* | 3
`N-Dimensional Engine Matrix` | internal, *__core__* | 20->*__40__* | 20->*__25__* | 4->*__5__* 
`Transpatial Drive` | core | 40->*__60__* | 80 | 8
`Singularity Engine Core` | core | 30->*__80__* | 30->*__40__* | 5->*__6__*

Costs were increased starting at 15 (more than most fuel parts, cheaper than shields) and slightly less than double cost for doubled speed.  Cost was not changed for `Transpatial Drive` as it potentially saves a slot for engine + stealth.

*(edit by Vezzra: changed Singularity Core Engine to Transpatial Drive in last sentence)*

Vezzra:
>  Cost was not changed for Singularity Engine Core as it potentially saves a slot for engine + stealth.

Um, I guess you mean the Transpatial Drive, not the Singularity Engine Core...?

MatGB:
He did, yes: and I wonder why I made it that expensive when comparing it to other stuff now I look through it, the stealth affect isn't good enough on its own. Ah well, the balance job will never be fully done.

For what it's worth, been playing with this for weeks now, vastly prefer it.

Vezzra:
Ok, made the edit to correct the post.

Morlic-fo (approved):
The AI changes are sufficient, there should be no issues with this from AI side.

MatGB:
Yup, I think the costs will need tweaking but they're fine and playable as is, better to have this in and do further work when we've time, I haven't tested yet with the unnerfed Solar Hullbut I suspect this will allow us to remove most of that nerfing.

dbenage-cx:
TY for testing and reviewing

PR #1624, opened 2017-07-07
add planet status marker

TheSilentOne1:
This PR adds an icon to indicate blockaded planets. For this, it adds PlanetHasSupply() to SupplyManager.
Corresponding forum thread: http://www.freeorion.org/forum/viewtopic.php?f=6&t=10566


Empire/Supply.cpp
```diff
@@ -163,6 +163,13 @@ bool SupplyManager::SystemHasFleetSupply(int system_id, int empire_id, bool incl
     return false;
 }
 
+bool SupplyManager::PlanetHasSupply(int planet_id) const {
```
Dilvish-fo: Since there are already a couple "SystemHasFleetSupply" methods here, and this is covering the same thing except at the planet level, it seems to me this should probably be named PlanetHasFleetSupply to be consistent.

Empire/Supply.cpp
```diff
@@ -163,6 +163,13 @@ bool SupplyManager::SystemHasFleetSupply(int system_id, int empire_id, bool incl
...
+    std::shared_ptr<const Planet> planet = GetPlanet(planet_id);
+    if (!planet)
+        return false;
+    return (planet->Owner() == EmpireThatCanSupplyAt(planet->SystemID()));
```
Dilvish-fo: Although we do not seem likely to revert back to allowing multiple empires to supply a system, I hate to see that assumption be woven into code where it is not needed, especially since it looks like it would be a bit more efficient as
```    return SystemHasFleetSupply(planet->SystemID(), planet->Owner); ```

Or, since in the only place you are using this you already have the planet object available, perhaps instead of adding this new method at all you simply use ```SystemHasFleetSupply(planet->SystemID(), planet->Owner)``` in the SidePanel code

**Comment by TheSilentOne1** [comment]
@Dilvish-fo: sounds reasonable. I've removed PlanetHasSupply() and used SystemHasFleetSupply() instead as you suggested.

Suggestions for the tooltip text from a native speaker would be welcome, other than that gtg from my side.

File: UI/SidePanel.cpp [hunk 3]
```diff
@@ -1136,12 +1145,25 @@ void SidePanel::PlanetPanel::RefreshPlanetGraphic() {
...
+
+    // add blockade graphic
+    if (!GetSupplyManager().SystemHasFleetSupply(planet->SystemID(), planet->Owner()) && (!planet->Unowned())) {
+        const std::shared_ptr<GG::Texture>& blockaded_texture = ClientUI::GetTexture(ClientUI::ArtDir() / "icons" / "blockade.png", true);
+        int texture_size = GetOptionsDB().Get<int>("UI.sidepanel-planet-blockaded-icon-size");
+
+        m_planet_blockaded_graphic = new GG::StaticGraphic(blockaded_texture, GG::GRAPHIC_FITGRAPHIC | GG::GRAPHIC_PROPSCALE);
+        m_planet_blockaded_graphic->Resize(GG::Pt(GG::X(texture_size), GG::Y(texture_size)));
+        m_planet_blockaded_graphic->MoveTo(GG::Pt(GG::X(MaxPlanetDiameter() / 2 - texture_size / 2 + 3),
+                                           GG::Y(planet_graphic_height / 2 - texture_size / 2)));
```
geoffthemedio: line up GG::Y with GG::X

UI/SidePanel.cpp
```diff
@@ -1136,12 +1145,25 @@ void SidePanel::PlanetPanel::RefreshPlanetGraphic() {
...
+        planet_graphic_height = GG::Y(MaxPlanetDiameter());
         GG::Pt planet_image_pos(GG::X(MaxPlanetDiameter() / 2 - planet_image_sz / 2 + 3),
                                 GG::Y(MaxPlanetDiameter() / 2 - planet_image_sz / 2));
         m_rotating_planet_graphic = new RotatingPlanetControl(planet_image_pos.x, planet_image_pos.y,
                                                               m_planet_id, m_star_type);
         AttachChild(m_rotating_planet_graphic);
     }
+
+    // add blockade graphic
+    if (!GetSupplyManager().SystemHasFleetSupply(planet->SystemID(), planet->Owner()) && (!planet->Unowned())) {
```
geoffthemedio: not having fleet supply doesn't necessarily mean a planet is blockaded, such as if there is no supply source that could read it in any case...

UI/SidePanel.cpp
```diff
@@ -1136,12 +1145,25 @@ void SidePanel::PlanetPanel::RefreshPlanetGraphic() {
...
+        const std::shared_ptr<GG::Texture>& blockaded_texture = ClientUI::GetTexture(ClientUI::ArtDir() / "icons" / "blockade.png", true);
+        int texture_size = GetOptionsDB().Get<int>("UI.sidepanel-planet-blockaded-icon-size");
+
+        m_planet_blockaded_graphic = new GG::StaticGraphic(blockaded_texture, GG::GRAPHIC_FITGRAPHIC | GG::GRAPHIC_PROPSCALE);
+        m_planet_blockaded_graphic->Resize(GG::Pt(GG::X(texture_size), GG::Y(texture_size)));
```
geoffthemedio: better to merge the Resize and MoveTo calls into a single SizeMove call

UI/SidePanel.cpp
```diff
@@ -1870,9 +1892,14 @@ void SidePanel::PlanetPanel::Refresh() {
     if (m_specials_panel)
         m_specials_panel->Update();
 
-    // set stealth browse text
     ClearBrowseInfoWnd();
 
+    // set planetpanel blockade browse text
+    if (!GetSupplyManager().SystemHasFleetSupply(planet->SystemID(), planet->Owner()) && (!planet->Unowned()))
+        SetBrowseInfoWnd(std::make_shared<TextBrowseWnd>(UserString("PL_BLOCKADED"),
```
geoffthemedio: I would set this browse wnd for the blockaded icon, not for the whole planet

UI/SidePanel.cpp
TheSilentOne1: Isn't a planet always a source for fleet supply as long as it isn't blockaded? If not, how else would I determine if a planet is blockaded?

UI/SidePanel.cpp
TheSilentOne1: If I use m_planet_blockaded_graphic->SetBrowseInfoWnd(...), the tooltip doesn't show. Any idea why it doesn't?

geoffthemedio: might not be enabled or might not be interactive

UI/SidePanel.cpp
geoffthemedio: At most one empire supplies each system. Assuming no fleets are present that can obstruct supply propagation, if empire B's supply range is bigger in a system, it will overpower empire A's supply range, and A won't be able to supply fleets there (unless allied). So far as I'm aware, this is not considered a "blockade"... although perhaps it should be?

I'm also not sure what happens when a planet just has 0 or less than 1 supply range. Can it supply fleets in system in that case? If not, it probably shouldn't be considered a blockade, as nothing is being blocked.

Dilvish-fo: > So far as I'm aware, this is not considered a "blockade"... although perhaps it should be?

Although I could see some validity to the idea of splitting the "blockade" concept into "fleet blockade" and "supply blockade", where the former is always accompanied by the latter but the latter might happen without the former, that seems to me to get a bit over-complicated.  I think I'd rather just refer to this situation as a "supply-contest-tie" or somesuch.  But I wouldn't object to that situation still leading to the planet-blockade icon being shown, it's just that it's tooltip then should be a bit more specific/clear.

> I'm also not sure what happens when a planet just has 0 or less than 1 supply range. Can it supply fleets in system in that case? If not, it probably shouldn't be considered a blockade, as nothing is being blocked.

It's not so common to even encounter a clear situation where this can be tested.  I am pretty sure, though, that  few weeks ago I did see a situation where I could test this, at either a Huge planet colony or a colony planet with elliptical orbit special or accretion disk system special, and my ships **did** still resupply there, at least if I stopped.  (I'll acknowledge, the scope of either my testing or memory is leaving something to be desired.)  However it's also my recollection that at GasGiant outposts they would not-- if I am not mis-remembering that then perhaps there is some difference between colonies and outposts?  I will try to do some more thorough testing sometime.

TheSilentOne1:
A quick note that I plan to pick this up again in the not-too-distant future. I needed some time to mull this over, and I think that we need to think about how to display, in general, the status of a planet (disconnected from supply, incapacitated due to attack or bombardment on the previous turn, disabled by rebels/riots...). I have some other things I want to do first and will make a post on the forums about the subject before resuming the work here.

geoffthemedio:
@TheSilentOne1 reminder

TheSilentOne1:
No need to, I have this on my todo list. I basically know where I'd like to take this (see above), I've started with a little concept work, but I'd like to do some revision work on bombardment first before I get back to this if it's okay.

geoffthemedio:
@TheSilentOne1 reminder anyway...

TheSilentOne1:
Despite my original plans, I probably won't be able to actively participate for about (at least) half a year.

TheSilentOne1:
I've resumed work on this. Instead of just a "blockade/no supply marker", this PR now introduces a more general "planet status marker" which will appear e. g. if a planet is disconnected from supply, or was attacked or conquered on the previous turn. More uses for the marker could be: planet was recently bombarded, happiness is very low (open for suggestions). If there is only one active status the icon will change to a specific marker (e. g. a supply crate for supply disconnection), otherwise it will show a generic one and the tooltip will give more in-depth information about active statusses.
Speaking of the tooltip, I'm still clueless why planet_status_graphic will not show one. Helpful hints, as usual, are much appreciated.

example
_Mean ol' Kraken attacked Canon Beta I._

TheSilentOne1:
Fixed the tooltip. Open for review.

geoffthemedio:
Could you add a blank line between multiple blocks of notification text in the tooltip? And perhaps also add the icons for each notification at the start of the text in the tooltip?

UI/SidePanel.cpp
```diff
@@ -1835,7 +1840,66 @@ void SidePanel::PlanetPanel::Refresh() {
...
+        if ((planet->GetMeter(METER_HAPPINESS)->Current() <= 5) && (planet->GetMeter(METER_POPULATION)->Current() > 0)) {
+            planet_status_messages.emplace_back(boost::io::str(FlexibleFormat(
+                UserString("OPTIONS_DB_UI_PLANET_STATUS_UNHAPPY")) % planet->Name()));
+            planet_status_texture = ClientUI::GetTexture(ClientUI::ArtDir() / "icons" / "planet_status_unhappy.png", true);
+        }
+
+        // status: bombarded (TBD)
+
+        // status: several
+        if (planet_status_messages.size() > 1) {
```
geoffthemedio, line 1880: Any way to arrange multiple buttons to be shown simultaneously instead of switching to a generic icon in that case?

UI/SidePanel.cpp
```diff
@@ -1835,7 +1840,66 @@ void SidePanel::PlanetPanel::Refresh() {
...
+            planet_status_texture = ClientUI::GetTexture(ClientUI::ArtDir() / "icons" / "planet_status_warning.png", true);
+        }
+
+        if (planet_status_messages.size() > 0 && planet_status_texture) {
+            int texture_size = GetOptionsDB().Get<int>("UI.sidepanel-planet-status-icon-size");
+            m_planet_status_btn = Wnd::Create<CUIButton>(
```
geoffthemedio: need to disable this button so that clicking it doesn't produce the button click sound

default/stringtables/en.txt
```diff
@@ -1582,6 +1582,24 @@ Sets the color of the scan-line shading over field icons.
...
+Sets the size of the planet status icon.
+
+OPTIONS_DB_UI_PLANET_STATUS_ICON_TITLE
+Planet status notification
+
+OPTIONS_DB_UI_PLANET_STATUS_ATTACKED
+%1% was attacked on the previous turn.
+
+OPTIONS_DB_UI_PLANET_STATUS_CONQUERED
+%1% was just conquered by our empire. Happiness will probably be reduced.
```
geoffthemedio: don't refer to "our" anything. (or the player in the second person)

UI/SidePanel.cpp
TheSilentOne1: I was passing a GG::Texture instead of a GG::SubTexture, so the wrong StaticGraphic constructor was called. This is fixed now.

PR #1625, opened 2017-07-13
Remove suppress_immediate_execution in OrderSet::IssueOrder.

LGM-Doyle:
Replace it with single execution on the client and server of each order.

This means that if an order needs to be issued during game load it can
take immediate effect on the client, and still not be re-executed by the
MapWnd initialization.

This does not need a special bool parameter.  It just works regardless of whether the order is coming from a loaded order or an order issued before or after the `MapInit()` function.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -1020,13 +1020,13 @@ bool HumanClientApp::ToggleFullscreen() {
     return true;
 }
 
-void HumanClientApp::StartGame() {
+void HumanClientApp::StartGame(bool is_new_game) {
     m_game_started = true;
 
     if (auto&& map_wnd = ClientUI::GetClientUI()->GetMapWnd())
         map_wnd->ResetEmpireShown();
 
-    ClientUI::GetClientUI()->GetShipDesignManager()->StartGame(EmpireID());
+    ClientUI::GetClientUI()->GetShipDesignManager()->StartGame(EmpireID(), is_new_game);
 
     UpdateCombatLogManager();
 }
--- a/util/Order.cpp
+++ b/util/Order.cpp
@@ -34,12 +34,19 @@ Empire* Order::GetValidatedEmpire() const {
 }
 
 void Order::Execute() const {
+    if (m_executed)
+        return;
+
     ExecuteImpl();
     m_executed = true;
 }
 
-bool Order::Undo() const
-{ return UndoImpl(); }
+bool Order::Undo() const {
+    auto undone =  UndoImpl();
+    if (undone)
+        m_executed = false;
+    return undone;
+}
 
 bool Order::Executed() const
 { return m_executed; }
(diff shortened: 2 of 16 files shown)


util/Order.cpp
```diff
@@ -34,12 +34,19 @@ Empire* Order::GetValidatedEmpire() const {
...
     ExecuteImpl();
     m_executed = true;
 }
 
-bool Order::Undo() const
-{ return UndoImpl(); }
+bool Order::Undo() const {
+    auto undone =  UndoImpl();
+    if (undone)
+        m_executed = false;
```
geoffthemedio, line 47: Is this a good idea? If rescinding an order, it probably shouldn't be re-executed when loading...

util/SerializeOrderSet.cpp
```diff
@@ -34,8 +36,15 @@ BOOST_CLASS_EXPORT(ForgetOrder)
 template <class Archive>
 void Order::serialize(Archive& ar, const unsigned int version)
 {
-    ar  & BOOST_SERIALIZATION_NVP(m_empire)
-        & BOOST_SERIALIZATION_NVP(m_executed);
+    ar  & BOOST_SERIALIZATION_NVP(m_empire);
+    /* m_executed is intentionally not serialized so that an order is run once
```
geoffthemedio: discuss how this works when loading a game? ie. all orders re-executed when a game is loaded, assuming the saved gamestate is that before any orders were executed by the client?

client/human/HumanClientApp.cpp
```diff
@@ -1016,13 +1016,13 @@ bool HumanClientApp::ToggleFullscreen() {
     return true;
 }
 
-void HumanClientApp::StartGame() {
+void HumanClientApp::StartGame(bool is_new_game) {
```
geoffthemedio, line 1023: seems unrelated, though I suppose isn't. a comment indicating the relevance might be helpful.

geoffthemedio:
Seems to work OK. Is there a particular test case, or are you fairly sure this will work as intended?

client/human/HumanClientApp.cpp
LGM-Doyle: I added a comment that `is_new_game` is false for loaded games.

LGM-Doyle:
There is a simple test case, which demonstrates the problem and exercises the new code.
- Quickstart a game.
- On turn 1 create a new ship design.
- save the game before you hit next turn
- load the just saved game.
- the log will have several error messages of the form `Tried to remove a ShipDesign id = 0that the empire wasn't remembering`.  Other than the error messages there is no other problem.  This PR results in no error message, because the code "knows" that the game is being reloaded on turn 1 and does not add a second set of orders to create the saved ship designs for the human empire.  (The formatting of the error message is already fixed in https://github.com/freeorion/freeorion/commit/a1bb05e438350af16a7a9581c868bae4a4de5620)

client/human/HumanClientApp.cpp
geoffthemedio: "is" or "should be" ?

LGM-Doyle: Changed to should and rebased,

LGM-Doyle: Changed to should and rebased.

PR #1626, opened 2017-07-13
Is the AlignmentManager a work in progress or should it be commented out?

LGM-Doyle:
The `AlignmentManager` code generates some spurious warnings.  

Is it a work in progress that should be left alone or can it be commented out for now until it is finished?


geoffthemedio:
It's certainly not "in progress". Whether it ever will be is unclear, but the implementation as is there now was probably premature.

Vezzra:
That thing is a truly ancient piece of code, it's been there as long as I can remember. IMO it does not make much sense to keep it, there hasn't even been something designed regarding a "Alignment" mechanic. Maybe there have some ideas been discussed ***long*** ago, but even if, that's probably buried in some 10 or more years old threads. So, lets just get rid of it.

But please, no commenting out (we already have enough commented out code lying around), just remove the entire thing from the codebase. It's preserved anyway in the repo, so we can dig it up if we really want to.

Vezzra:
Tagging as "refactoring". Although that doesn't really apply, it's probably the the closest to "code maintenance"...

LGM-Doyle:
I have changed this to simply remove the `AlignmentManager` class.

geoffthemedio:
Should probably remove https://github.com/freeorion/freeorion/tree/master/default/scripting/alignments at the same time...?

Vezzra (changes requested):
In addition to what @geoffthemedio mentioned, there are several more pieces of code related to the rudimentary implementation of an "Alignment" mechanic:

* The [`Alignment` class in `Empire.h`](https://github.com/freeorion/freeorion/blob/master/Empire/Empire.h#L22)
* In `Empire.cpp`: [three global functions](https://github.com/freeorion/freeorion/blob/master/Empire/Empire.cpp#L1274) right above the namespace containing the `AlignmentManager` implementation
* The alignments parser

All these must be removed also, and the cmake scripts, the MSVC and Xcode projects and [default/scripting/README.md](https://github.com/freeorion/freeorion/blob/master/default/scripting/README.md) need to be updated accordingly.

LGM-Doyle:
All aspects of `Alignment` are now removed by this PR.

PR #1627, opened 2017-07-13
Fix UniverseObject id allocation.

LGM-Doyle:
This PR depends on #1625.

***The Symptom***
As observed by @MatGB, occasionally splitting fleets in the `FleetWnd` causes the UI to hang.  The problem can be replicated as follows:
1. Create a large fleet and merge it.  
2. Hit next turn.
3. As soon as the UI is active again, start splitting and merging the fleet.  If you hit the bug the UI will hang.  On my machine with a fleet of around 100 ships the split time is typically 24 ms with hangs of at least 6 seconds.

***The Problem***
The server coordinates all of the object ids in the game.  The fleet wnd calls `GetNewDesignID()`, which in turn calls `ClientNetworking::SendSynchronousMessage()`.  `SendSynchronousMessage()` suspends the thread until the server responds with a new id.  Suspending the UI thread, due to a busy server or network congestion due to other clients, causes the glitch.

If the server crashes during this glitch it may result in the hung AI processes, that prevent the game from starting up.

***This Solution***
This PR partitions the id space modulo the number of clients plus one for the server.  Each client's backend can then immediately allocate an id from its assigned space.

Benefits of this solution include: continuing to use `int` for the id and continuing to have sequential small integer ids, which display well in the Objects window.

On my machine for the same test case, the times to split a fleet were all around 18ms.

***The Details***
This PR creates a class `IDAllocator`, which allocates ids and checks that ids are appropriate for the client that is trying to use them.  `IDAllocator` is used to allocate ids for both universe objects and ship designs.

It pulls all id allocation into the `Universe` code.  This stops using the idiom where an id is sent to the client, which then sends the same id back to the server.  This is true everywhere except the new ShipDesign order which uses the ship design id from the new ship design on the client when re-running the order on the server.

This changes all of the per type universe object construction code like `universe.CreateBuilding(assorted parameters)` to a templated `universe.InsertNew<Building>(assorted parameters`.  The template version works for a new types of universe object `T` or new constructors of types automatically.

There is also a `Universe::InsertTemp<T>()`, which creates an object with the temp id.

Finally, there is code to allow current saved games to load and ignore incorrectly assigned ids, that are stored in orders already executed on the client and still pending execution on the server.  This code can be removed once enough saved games are using the `IDAllocator` and then all orders will have their id verified on the server.


geoffthemedio:
Can one then tell from the id which empire created an object and how many objects that empire has created?

Dilvish-fo:
> Can one then tell from the id which empire created an object and how many objects that empire has created?

I'm sure it would be possible.  You could make it more difficult in various ways, with a complex partition scheme that would vary from game to game, but that would still be just more difficult to a limited extent since the partition scheme would be open source.  With some finite (and probably not large) number of object ID samples for their own empire and for an empire of interest then I expect a client could figure out the partition scheme details for that game.

So for multiplayer games we should at least plan or note the possibility  (I'd be content with a TODO comment) for an option for turning this approach on or off as part of game setup.  If the partition approach is just-enough complex that a player couldn't easily/automatically figure it out for that game just by eyeballing a few samples then that's probably good enough for the very great majority of multiplayer games.  The really distrustful ones could choose to deal with the periodic hangs if they want (if/when the controlling option were implemented).  We could let the Volker et al. multiplayer gang let us know how much of an issue they think the control option is.

mmoderau:
Well, you could always encrypt the ID generated on client and use the encrypted ID as global ID, decrypting only on server in order to verify that the ID comes from the right client. 

This would require asymmetric encryption with server holding the private key and clients using the public key. 

As encryption is reversible, each encrypted ID would be unique.

Dilvish-fo:
Hmm, certainly an interesting idea, worth thinking about some more.  To use the encrypted ID as the global ID, that would mean an int/string/block/whatever the same size as the public encryption key, correct?  The biggest that would be easily manageable would be a long long int at 64 bits, but I suppose that would be enough to provide a reasonable level of security for our purposes.

LGM-Doyle:
Yes, you could tell the client from the modulus of the ids.  I've added an obfuscation function to help with that as described below.

Note that only fleet ids and ship design ids are issued on the client.  All empire production is handled on the server.  If the leaked information is being exploited by human players, it is also easy for their adversaries to spoof the number of ids allocated by joining and merging fleets, or creating and not using ship designs.

Replacing the ids with UUIDs is a simple solution that neither leaks any information nor requires any coordination. I did not use that solution for two reasons.  I was unsure of how culturally committed FreeOrion is to using small integers as ids.  UUIDs are not as easy to type or display in the UI.  Secondly, conversion and verification of the type of the id is a larger task than necessary to fix this bug.  However, I have added a note in the code that using UUID would be a better solution that the `IDAllocator`.

I added `IDAllocator::ObfuscateBeforeSerialization()`, which is run once per turn and does the following:

- randomizes the moduli 
-  adds a random offset to each allocator, so that they all appear to have allocated between 1 and 1.5 times as many ids as the client that allocated the greatest number of ids that turn.

I hope that sufficiently limits the information leaked via the allocated ids.

geoffthemedio:
needs rebase

LGM-Doyle:
rebased.

universe/IDAllocator.cpp
```diff
@@ -0,0 +1,342 @@
...
+        & BOOST_SERIALIZATION_NVP(m_warn_threshold)
+        & BOOST_SERIALIZATION_NVP(m_exhausted_threshold);
+
+    if (Archive::is_loading::value) {
+        // Always load whatever is there.
+        ar & BOOST_SERIALIZATION_NVP(m_empire_id)
+            & BOOST_SERIALIZATION_NVP(m_empire_id_to_next_assigned_object_id)
+            & BOOST_SERIALIZATION_NVP(m_offset_to_empire_id);
+
+        DebugLogger(IDallocator) << "Deserialized [" << [this](){
```
geoffthemedio, line 288: space between `()` and `{`

universe/IDAllocator.cpp
```diff
@@ -0,0 +1,342 @@
...
+}
+
+template <class Archive>
+void IDAllocator::SerializeForEmpire(Archive& ar, const unsigned int version, const int empire_id) {
+    DebugLogger(IDallocator) << (Archive::is_loading::value ? "Deserialize " : "Serialize ")
+                             << "IDAllocator()  server id = "
+                             << m_server_id << " empire id = " << empire_id;
+
+    ar & BOOST_SERIALIZATION_NVP(m_invalid_id)
+        & BOOST_SERIALIZATION_NVP(m_temp_id)
```
geoffthemedio, line 276: ideally line up the `&`

universe/Universe.cpp
```diff
@@ -153,6 +154,29 @@ void Universe::Clear() {
     m_marked_destroyed.clear();
 }
 
+
+void Universe::ResetAllIDAllocation(const std::vector<int>& empire_ids) {
+
```
geoffthemedio, line 159: unnecessary blank line?

PR #1629, opened 2017-07-16
Update engine tech descriptions

luiges90:
...For some reasons the engine techs still say the engine parts are stackable and have an old increase value.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -10842,7 +10842,7 @@ Improved Engine Couplings
 SHP_IMPROVED_ENGINE_COUPLINGS_DESC
 '''[[COLONY_BUILDING_TIME_DECREASE]]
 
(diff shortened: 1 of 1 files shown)


Vezzra (approved):
Looks fine. Anyone any objections against merging this? Any corrections? @MatGB?

Ouaz (approved):
It's OK; Related to #1621 : 

"+FU_ENGINE_PART_LIMIT
+Only one engine part may be equipped on a ship."

Ok too for the value increase (10 ; 20 ; 30 to 20 ; 40 ; 80)

MatGB:
Ah, we did the part descriptions but missed the tech descriptions, another reason to either have the stats be automatically in or to remove details from techs and have them link to the part.

Also, I misread the first email and htought this was an Issue not a PR, sorry to not review it earlier and thanks for the catch.

PR #1630, opened 2017-07-17
Simplify part stacking rules by using the FOCS "exclusions" keyword

Morlic-fo:
The previous effectgroups were a workaround from a time where we did not have the ```exclusions``` keyword in FOCS which provides a simpler and more user friendly functionality by actually preventing the user from building invalid designs.

This also reduces the number of effects in the game, possibly leading to some minor performance improvements in the lategame.

Note that the BEST_SHIELD_EFFECT and BEST_CLOAK_EFFECT macros are kept for now because FU_TRANSPATIAL_DRIVE and SH_ROBOTIC_INTERFACE_SHIELDS rely on these.

### DIFF
--- a/default/scripting/ship_parts/Cloaks.focs.txt
+++ b/default/scripting/ship_parts/Cloaks.focs.txt
@@ -1,6 +1,7 @@
 Part
     name = "ST_CLOAK_4"
     description = "ST_CLOAK_4_DESC"
+    exclusions = [[ALL_CLOAKS]]
     class = Stealth
     capacity = 80
     mountableSlotTypes = Internal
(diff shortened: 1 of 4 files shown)


MatGB:
Good call: I'd been thinking we ought to try to simplify using priorities, would that work for the remaining effects, a stacking group  for both with the one we don't want with a later priority should work.

Morlic-fo:
> I'd been thinking we ought to try to simplify using priorities, would that work for the remaining effects

If you are refering to the robotics shields and transpatial drive, then that approach seems to be viable. The priorities would need to scale with the effect/part capacity so that the current "the strongest effect comes into play" rule holds.

geoffthemedio:
@MatGB can you check that this doesn't break anything, then merge if OK?

MatGB:
Appears to work fine, exclusions apply correctly and parts still work.

PR #1631, opened 2017-07-17
Reorganize FOCS files for ship parts

Morlic-fo:
Currently, there is some mix between grouping up similar parts into a single file (e.g. weapons) or defining only 1 part per focs file (e.g. engines).
This PR moves more parts into a common file to improve file organization.

The Bombard file becomes rather large but most of the definitions are copy-pasted and can be replaced/shortened by macros which will be subject to a later PR.

Based and depending on #1630.


geoffthemedio:
Needs rebase.

Some comment from scripters about whether this reorganization is preferred would be appreciated.

MatGB:
I don't have a strong opinion but I've got used to them being split up, the idea had been to make it easier to add/find/amend individual items, I was never clear why a remnant were in shared files but it was a small number of generic parts.

@dbenage-cx it was your plan to split them up in the first place, you're probably best to comment.

Morlic-fo:
I think we should be consistent. Having both ways seems odd. 

I can see the merits of having one part per file. In that case though, I would still suggest to organize the different part types using subdirectories similar to the hulls directory.

dbenage-cx:
@MatGB Lack of foresight on my part 😉 I kept them grouped if the first and last entries were easily compared.

@Morlic-fo Agree with your last post and would prefer the single definition per file (organized through sub-directories) for default content, with macro files following scope.

dbenage-cx:
@Morlic-fo Any work currently planned/in-progress, or would you prefer I take this on?

Morlic-fo:
I'm not actively working on this at the moment. If you want to take over, feel free to do so.

Vezzra:
Superseded by #1718.

PR #1636, opened 2017-07-18
Derive GG::Wnd from std::enable_shared_from_this

LGM-Doyle:
This is the a step in converting GG to use the std smart pointers to
manage windows.

Calling shared_from_this() within a constructor will throw bad_weak_ptr
because construction of its internal weak_ptr is incomplete.  This will
have the consequence that AttachChild() can not be called from within
either the parent or the child window's constructor.  Any constructors
that need to call AttachChild() need to be converted to factory
functions.

### DIFF
--- a/GG/GG/Wnd.h
+++ b/GG/GG/Wnd.h
@@ -38,6 +38,7 @@
 #include <list>
 #include <set>
 #include <vector>
+#include <memory>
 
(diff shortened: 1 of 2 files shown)


geoffthemedio:
haven't tested, but simple enough...

PR #1637, opened 2017-07-18
Add GG::Wnd::Create<T>() and GG::Wnd::CompleteConstruction().

LGM-Doyle:
After the conversion to using shared_ptr to track manage Wnd, AttachChild() will
not be callable within a constructor, because it requires a completely
constructed shared_from_this().

Wnd::Create<T> is a generic factory function for Wnd derived types T.  It
constructs the shared pointer and then calls CompleteConstruction() to handle
operations that require a shared_from_this().

T can override CompleteConstruction() to provide any post
constructor work needed to complete the class T.

### DIFF
--- a/GG/GG/Wnd.h
+++ b/GG/GG/Wnd.h
@@ -338,8 +338,8 @@ public:
     }
 
     /** CompleteConstruction() should be overriden to complete construction of derived classes that
-        need a fully formed weak_from_this() (i.e. to call AttachChild()) to be correctly constructed. */
-    virtual void CompleteConstruction(){};
+        need a fully formed weak_from_this() (e.g. to call AttachChild()) to be correctly constructed. */
+    virtual void CompleteConstruction() {};
 
     /** \name Accessors */ ///@{
     /** Returns true iff a click over this window does not pass through.  Note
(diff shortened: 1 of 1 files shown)


GG/GG/Wnd.h
```diff
@@ -309,6 +326,21 @@ class GG_API Wnd : public boost::signals2::trackable
...
+    template <typename T, typename... Args>
+    static T* Create(Args&&... args)
+    {
+        auto wnd(new T(std::forward<Args>(args)...));
+        wnd->CompleteConstruction();
+        return wnd;
+    }
+
+    /** CompleteConstruction() should be overriden to complete construction of derived classes that
+        need a fully formed weak_from_this() (i.e. to call AttachChild()) to be correctly constructed. */
```
geoffthemedio: eg. not ie.

GG/GG/Wnd.h
```diff
@@ -309,6 +326,21 @@ class GG_API Wnd : public boost::signals2::trackable
...
+    virtual void CompleteConstruction(){};
```
geoffthemedio: prefer space b4 {}

GG/GG/Wnd.h
LGM-Doyle: fixed.

GG/GG/Wnd.h
LGM-Doyle: fixed.

Issue #1639, opened 2017-07-20
Ship design can crash when right-clicking slots while dragging parts


rriegs:
### Environment

* **FreeOrion Version**: FreeOrionCH v0.4.7+ [build 2017-07-19.79efb88] CMake
* **Operating System**: Ubuntu 16.04.2 LTS (Linux 4.4.0-83 x86_64)
* **Graphic card used**: Intel Corporation 3rd Gen Core processor Graphics Controller (rev 09)
* **Fetched as**
  * Compiled from source

### Description

In the Design screen, left-click-dragging parts from the Ship Parts window over ship slots but then right-clicking (while still dragging) results in strange behavior: the part is placed, but its listing in the Ship Parts window vanishes and can no longer be selected.  Closing and reopening the Design screen restores the Ship Parts window's contents.

Further, left-click-dragging parts from *other ship slots* over ship slots and then right-clicking sometimes results in a segfault.  Steps to reproduce are as follows:

1. Place a ship part as normal.
2. Left-click-drag the placed ship part over another ship slot (empty or otherwise).
3. Right-click while still dragging.

The above procedure sometimes does not result in a crash, especially if ship parts are initially placed via right-click-while-dragging instead of as normal (as per Step 1 above).  Regardless, even if a crash does not occur, subsequent normal modifications to ship parts (left-click drag-and-dropping only) can later result in a crash.

I hypothesize that the ship part is somehow being placed in two ship slots at the same time, and that modifications that would normally delete the part from one slot then result in a dead pointer in the other slot, causing a crash.  Nothing useful seems to end up in the logs.

### Expected Result

At minimum, right-clicking while left-click-dragging a ship part should have no effect.

*Feature*: On the other hand, I think it would be neat if it placed a copy of the part while *still* dragging the part, thereby allowing you to place many copies of the same part in adjacent slots by left-click-dragging a part (from the Ship Parts window *or* from another slot) and then right-clicking multiple slots in sequence.

dbenage-cx:
For the second portion, could only replicate exact steps without gdb. [abrt stacktrace](https://github.com/freeorion/freeorion/files/1164106/designwnd-part-drag-rclick.stacktrace.txt)

With gdb, added a step of double clicking the new part location.
Thus:
* Double click medium basic hull
* Double click standard plate armor
* Left button drag the armor part from top external slot to bottom external slot
* Still holding left button, right click
* Release left button
* Double left click bottom part slot

[...]

PR #1632, opened 2017-07-18
Convert GG  to use smart pointers.

LGM-Doyle:
This PR converts GG to use smart_pointers to manage `GG::Wnd`.

The GG code was one of the areas still using naked new to manage heap resources. It moves towards the 'no naked new' framework @Dilvish [described](https://github.com/freeorion/freeorion/pull/1544#discussion_r115542780).  

The PR has the benefit of fewer leaks and making it more difficult to leak memory.  It will allow FreeOrion to treat pointers throughout the codebase in a uniform and standard manner.  By using shared_ptr it allows the removal of the window dying code, which pre-invented the `shared_ptr` reference counting behavior.

It uses `shared_ptr` for 'owning' pointers in the `ZList`, `m_children` and dragged windows in `GUI`.  'Non owning' pointers are `weak_ptr`, for example the parent pointer.  This means that there are no reference cycles, and when the child `shared_ptr` is destroyed, its entire hierarchy is recovered.

***Breakup***
The PR will be broken into smaller PRs.

However, all pointers derived from `GG::Wnd` must be converted at the same time and there will be one large commit, which is the smallest commit that changes all of the pointers and still compiles.  There may also be one large PR, the smallest set of commits following changing the children to `shared_ptr` that does not have a known crash.

***Technical Detail***
Most of the conversion is straightforward conversion of pointers to either `shared_ptr` or `weak_ptr`.

Conversion of the constructors ran into difficulty, when constructors are calling `AttachChild()` or `SetLayout()`.

`AttachChild()` needs to call `shared_from_this()` in order to get the parent pointer.  The function`shared_from_this<Wnd>()` can't be called from within a constructor or a destructor, because it needs a completely constructed `Wnd` that has been assigned to a `shared_ptr` and is not expired.  

The solution in this PR is to create a generic factory function, `static shared_ptr<T> Wnd::Create<T>(Args...)` for all types `T` derived from `Wnd`.  The factory function `, Wnd::Create<T>(Args...)` constructs a `new T(Args)` and then calls `virtual void Wnd::CompleteConstruction()`.  Any class that calls `AttachChild()` or `SetLayout()` overrides `Wnd::CompleteConstruction()` to complete its construction after it has a valid `shared_from_this()` pointer.

All of the commits creating the `CompleteConstruction()` functions precede the commit changing over to `shared_ptr` which only needs to change the return type of `Wnd::Create<T>(Args...)` from `T*` to `shared_ptr<T>`.

PR #1645, opened 2017-07-28
Refactor GUI before converting to `shared ptr`

LGM-Doyle:
This PR makes the following changes to GUI.cpp that can be made before the big changes to use `shared_ptr`:
- converts `GG::s_impl` from a `shared_ptr` to a `unique_ptr` which more accurately conveys its purpose.
- uses `m_curr_wnd_under_cursor` instead of `m_drag_wnds[x]` to be clear that the code cares about the window under the cursor.

### DIFF
--- a/GG/src/GUI.cpp
+++ b/GG/src/GUI.cpp
@@ -1217,12 +1217,13 @@ void GUI::RegisterModal(Wnd* wnd)
 
 void GUI::Remove(Wnd* wnd)
 {
-    if (wnd) {
-        if (!m_impl->m_modal_wnds.empty() && m_impl->m_modal_wnds.back().first == wnd) // if it's the current modal window, remove it from the modal list
-            m_impl->m_modal_wnds.pop_back();
-        else // if it's not a modal window, remove it from the z-order
-            m_impl->m_zlist.Remove(wnd);
-    }
+    if (!wnd)
+        return;
+
+    if (!m_impl->m_modal_wnds.empty() && m_impl->m_modal_wnds.back().first == wnd) // if it's the current modal window, remove it from the modal list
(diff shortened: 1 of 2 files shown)


GG/src/GUI.cpp
```diff
@@ -316,28 +316,28 @@ void GUIImpl::HandleMouseButtonPress(unsigned int mouse_button, const Pt& pos, i
     m_mouse_button_state[mouse_button] = true;
     m_drag_wnds[mouse_button] = m_curr_wnd_under_cursor; // track this window as the one being dragged by this mouse button
     if (m_curr_wnd_under_cursor) {
-        m_prev_wnd_drag_position = m_drag_wnds[mouse_button]->UpperLeft();
+        m_prev_wnd_drag_position = m_curr_wnd_under_cursor->UpperLeft();
```
geoffthemedio, line 319: This doesn't seem like a clarification, but rather a change in logic. Please elaborate on why this should be changed...?

LGM-Doyle: It is not a change in logic. Line 317 is `m_drag_wnds[mouse_button] = m_curr_wnd_under_cursor;`, so they point to the same window.  `m_curr_wnd_under_cursor` is set from `CheckedGetWindowUnder` at the beginning of this function.  It seemed clearer to continue referring to it by its first name.  

It anticipated the bug fix for https://github.com/freeorion/freeorion/issues/1639.  THe commit, https://github.com/freeorion/freeorion/pull/1632/commits/d8f60387906bacd1e8184c65e2db40b53b16d4d8, changes the code so that an additional mouse button click while a window is already being dragged does not start a new drag.  `curr_wnd_under_cursor` and  are not the same `m_drag_wnds[mouse_button]`.

PR #1652, opened 2017-08-03
Fix Bug #1639 Dragging parts in DesignWnd incorrectly hides parts.

LGM-Doyle:
This PR depends on #1632 and fixes issues raised in #1639.

The PR uses `GUI::DragDropWnd()` to check if a `PartControl` that is dragged over a slot should hide to indicate that the part can be replaced or not hide because it is being dragged over itself.

The PR also refactors some code associated with the dragging and dropping of `PartControls` in order to reduce repeated code and level of nesting.


geoffthemedio:
rebase...

LGM-Doyle:
rebased.

Issue #1405, opened 2017-03-16
The --auto-quit option pauses but doesn't exit the application.


LGM-Doyle:
This issue arose in the review of PR [#1375 Improve app termination correctness.](https://github.com/freeorion/freeorion/pull/1375#issuecomment-286592239)

### Environment

* **FreeOrion Version**: master and [PR #1375 ](https://github.com/freeorion/freeorion/pull/1375/commits/f4447598c7983b98ec58a876c28129c1fe3e0c98)
* **Operating System**: Ubuntu

### Expected Result
The expected behavior of auto quit is to run some number of turns and then exit the application.  For example:
`./freeorion --auto-advance-n-turns 10 --auto-quit --load ~/.local/share/freeorion/save/test.sav`

should load `text.sav`, run 10 turns and then exit.

### Description of Problem
The buggy behavior as described in more detail [here](https://github.com/freeorion/freeorion/pull/1375#issuecomment-286922496), is that the application runs and then pauses, but does not exit.

LGM-Doyle:
I forgot to mention that this works for me.  I do not have a non-working example of this issue to fix.

This command line ````./freeorion --auto-advance-n-turns 10 --auto-quit --quickstart```` also works correctly for me.

geoffthemedio:
I get a save game in progress popup after 10 turns. Clicking to exit dismisses it, but doesn't exit. Clicking the top-right gives me the popup again. Clicking to exit again actually exits.

[logs.zip](https://github.com/freeorion/freeorion/files/1195190/logs.zip)

LGM-Doyle:
Thanks for the logs.  I'll look into it.

PR #1653, opened 2017-08-04
Fix bug #1405 auto quit pauses but doesn't exit.

LGM-Doyle:
This PR refactors and changes the following auto-save and auto-quit functionality:
- always create a final turn auto-save when auto-quitting
- do not catch all exceptions in HumanClientApp message queue.  Catching all exceptions creates a race state with `CleanQuit` exception which may be caught in the message queue catch statement depending on timing.  It is unclear from the code and the commit that added the `try ... catch` what exceptions were meant to be caught.
- move auto-quit to the completion of the (now automatic) auto-save.  Previously, the auto-quit was in a race with the auto-save which would prevent the auto-quit from happening on some machines.

Based on information in @geoffthemedio 's logs posted in #1405 I think that there were multiple race states preventing auto-quit from working consistently.  The auto-quit needed to be initiated after the auto-save completed and outside of the window where the message queue was polled.  The auto-save being in progress prevents the server from exiting and the `CleanQuit` exception being caught prevents the client from exiting.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -1138,20 +1138,25 @@ namespace {
 }
 
 void HumanClientApp::Autosave() {
+    // Create an auto save for 1) new games on turn 1, 2) if auto save is
+    // requested on turn number modulo autosave.turns or 3) on the last turn of
+    // an auto run.
+
     // autosave only on appropriate turn numbers, and when enabled for current
     // game type (single vs. multiplayer)
     int autosave_turns = GetOptionsDB().Get<int>("autosave.turns");
-    if (autosave_turns < 1)
-        return;     // avoid divide by zero
-    if (CurrentTurn() % autosave_turns != 0 && CurrentTurn() != 1)
-        return;     // turns divisible by autosave_turns, and first turn, have autosaves done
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Seems to work.

What was the previous behaviour for auto-saving in auto-advance and auto-quit games? I'm a bit skeptical about forcing auto-saves in cases when it's not enabled, just because auto-quit is enabled.

Dilvish-fo:
This fixes auto-quit for me, which is great.  

Although I would most often not really need an autosave just prior to the autoquit (I generally just parse log files for AI performance data when using auto-advance and auto-quit), I could see it being handy sometimes, it seems no stranger than the preexisting always-forced autosave on turn 1, and the cost seems pretty modest.

If the choice is between adding an extra option to be able to force an autosave before auto-quite, versus just rolling that into auto-quit as herein, I would have no major objection to the former but the latter also seems reasonable to me (though I suppose the auto-quit help message should then be edited to mention the autosave).

Again, thanks much for this fix.

geoffthemedio:
Based on @Dilvish-fo 's comment, will merge now. Tweaks to the autosave timing can be implemented later...

PR #1656, opened 2017-08-05
Use Wnd::Create to move some modal wnds from stack to heap.

### LINKED ISSUE #1655: Popup Windows Blank
While testing a pull request, a dialog popup was blank. I tried again with master by clicking the top-right X before an autosave completed, and got the same, see attached.
blank_error

LGM-Doyle:
This PR fixes #1655, which is caused because some of the modal windows were still being created on the stack and only calling their constructors and not their `CompleteConstruction()` functions.

### DIFF
--- a/UI/ClientUI.cpp
+++ b/UI/ClientUI.cpp
@@ -994,12 +994,12 @@ ClientUI* ClientUI::GetClientUI()
 { return s_the_UI; }
 
 void ClientUI::MessageBox(const std::string& message, bool play_alert_sound/* = false*/) {
-    GG::ThreeButtonDlg dlg(GG::X(320), GG::Y(200), message, GetFont(Pts()+2),
-                           WndColor(), WndOuterBorderColor(), CtrlColor(), TextColor(), 1,
-                           UserString("OK"));
+    auto dlg = GG::Wnd::Create<GG::ThreeButtonDlg>(GG::X(320), GG::Y(200), message, GetFont(Pts()+2),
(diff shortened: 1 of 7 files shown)


UI/ClientUI.cpp
```diff
@@ -994,12 +994,12 @@ ClientUI* ClientUI::GetClientUI()
 { return s_the_UI; }
 
 void ClientUI::MessageBox(const std::string& message, bool play_alert_sound/* = false*/) {
-    GG::ThreeButtonDlg dlg(GG::X(320), GG::Y(200), message, GetFont(Pts()+2),
-                           WndColor(), WndOuterBorderColor(), CtrlColor(), TextColor(), 1,
-                           UserString("OK"));
+    auto dlg = /*TODO: Remove extra shared_ptr wrap after Wnd::Create converted to return shared_ptr*/std::shared_ptr<GG::ThreeButtonDlg>(GG::Wnd::Create<GG::ThreeButtonDlg>(GG::X(320), GG::Y(200), message, GetFont(Pts()+2),
+                                                   WndColor(), WndOuterBorderColor(), CtrlColor(), TextColor(), 1,
+                                                                                                                                                                              UserString("OK")));
```
geoffthemedio: ... indenting?

LGM-Doyle: Indenting fixed.

It is reverted in the first few commits of https://github.com/freeorion/freeorion/pull/1632.

PR #1657, opened 2017-08-05
Added an appdata.xml file for Linux software gallery integration

Mailaender:
From http://linkdupont.fedorapeople.org/sources/freeorion.appdata.xml which should be included upstream. See https://fedoraproject.org/wiki/Packaging:AppData for more information about what this does.


Mailaender:
See https://build.opensuse.org/request/show/514718 for an integration test.

geoffthemedio:
@adrianbroher do you have any comment on this?

Dilvish-fo:
I'll certainly defer to adrianbroher on this, but in the meantime I'll also note that while [the related guide from debian](https://wiki.debian.org/AppStream/Guidelines) appears to say that the .desktop file is already minimally sufficient for the main purpose of providing basic info for a software center, the extra appdata here does look helpful.

CMakeLists.txt
```diff
@@ -581,6 +581,12 @@ install(
     DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/applications
 )
 
+install(
+    FILES
+    ${CMAKE_SOURCE_DIR}/freeorion.appdata.xml
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/appdata
```
Dilvish-fo: While the Fedora page that you linked to does appear to  (rather indirectly to my eyes) refer to this file as being installed in this appdata directory,  both the freedesktop.org page that you linked (and which the Fedora page itself links to) and the debian page that I linked both appear to me to explicitly direct that the file be installed to the /usr/share/metainfo/ directory.

Mailaender: In fact I am not sure. I thought metainfo was for plugins/addons. Something must have changed there. Maybe @hughsie can explain current best practice.

adrianbroher:
I will take a look at this PR after reading the spec.

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>freeorion.desktop</id>
+  <name>FreeOrion</name>
+  <summary>Turn-based space empire and galactic conquest (4X) computer game</summary>
+  <summary xml:lang="de">Rundenbasierendes Weltraum-Strategiespiel</summary>
+  <developer_name>FreeOrion Project</developer_name>
+  <description>
+  <p>
```
Cjkjvfnby: Missed indent.

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+  <project_license>GPL-2.0</project_license>
+  <url type="bugtracker">https://github.com/freeorion/freeorion/issues</url>
+  <url type="donation">http://www.freeorion.org/index.php/Donations</url>
+  <url type="faq">http://www.freeorion.org/index.php/FAQ</url>
+  <url type="help">http://www.freeorion.org/forum/viewforum.php?f=25</url>
+  <url type="homepage">http://www.freeorion.org</url>
+  <screenshots>
+    <screenshot type="default">
+      <caption>Galaxy setup</caption>
+      <image height="600" width="786">http://www.freeorion.org/images/thumb/1/1a/Freeorion_0.4.5_b2015-09-01.f203162_singleplayer.JPG/786px-Freeorion_0.4.5_b2015-09-01.f203162_singleplayer.JPG</image>
```
Cjkjvfnby: I am not sure but screenshot might be little outdated. `31-Dec-2015`

Mailaender: Indeed. It is the most recent from http://www.freeorion.org/index.php/Screenshots though.

CMakeLists.txt
hughsie: You can use either; `/usr/share/metainfo` is the new location but appdata will be scanned basically forever and the old location works on older distros like Fedora 23 -- it's up to you. If this is a new file, I'd go with the new location.

Vezzra:
Question: is this `freeorion.appdata.xml` something specific for the Fedora distro, or something that's used by other distros as well? If it's Fedora specific, I'm a bit wary to include that into upstream. I'm skeptical about the usefulness of maintaining distro-specific bits and pieces... but let's hear what @adrianbroher has to say.

hughsie:
> is this freeorion.appdata.xml something specific for the Fedora distro

It's used by every major distribution now. Over 1100 different upstreams have now added AppData files. :)

Vezzra:
Excellent. In that case I'm very much in favor of accepting this PR.

Mailaender:
Updated.

geoffthemedio:
Any further objections or comments?

Vezzra:
@adrianbroher? You've assigned this to yourself...

adrianbroher:
@Vezzra see https://github.com/freeorion/freeorion/pull/1657#issuecomment-320508164

Vezzra:
@adrianbroher, yes, I'm aware, just wanted to make sure @geoffthemedio's comment didn't escape your notice (got the impression that he wants to merge this).

geoffthemedio:
@adrianbroher reminder / your input would be appreciated.

geoffthemedio:
@adrianbroher Another reminder, since you've been active again recently, in case you have some time too look at this...

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
```
adrianbroher: Wrong type `desktop`, according to https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html#spec-appdata-filespec this is legacy and should be replaced with `desktop-application`.

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+  <id>freeorion.desktop</id>
```
adrianbroher: Dubious `id`, according to https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-filespec it should be a reverse domain notation e.g. `org.freeorion`.

Probably need to cross check with `freeorion.desktop` file.

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+      <p>
+       FreeOrion is a free, open source, turn-based space empire and galactic conquest (4X) computer game being designed and built by the FreeOrion project. FreeOrion is inspired by the tradition of the Master of Orion games, but is not a clone or remake of that series or any other game.
+      </p>
+  </description>
+  <keywords>
+    <keyword>game</keyword>
+    <keyword>strategy</keyword>
+  </keywords>
+  <metadata_license>CC-BY-SA-3.0</metadata_license>
+  <project_license>GPL-2.0</project_license>
```
adrianbroher: How do we display the actual "dual" licensing of game assets?

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>freeorion.desktop</id>
+  <name>FreeOrion</name>
+  <summary>Turn-based space empire and galactic conquest (4X) computer game</summary>
+  <summary xml:lang="de">Rundenbasierendes Weltraum-Strategiespiel</summary>
+  <developer_name>FreeOrion Project</developer_name>
+  <description>
+      <p>
+       FreeOrion is a free, open source, turn-based space empire and galactic conquest (4X) computer game being designed and built by the FreeOrion project. FreeOrion is inspired by the tradition of the Master of Orion games, but is not a clone or remake of that series or any other game.
```
adrianbroher: Probably should be synced with or created from the first section of the README.

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+      </p>
+  </description>
+  <keywords>
+    <keyword>game</keyword>
```
adrianbroher: Probably wrong lower case:

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-filespec

> A list of valid category names can be found in the Freedesktop menu spec.

https://specifications.freedesktop.org/menu-spec/latest/apa.html#main-category-registry

> Note that category names are case-sensitive.

So it's `Game`, not `game`.

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+    <keyword>strategy</keyword>
```
adrianbroher: There is not additional category `strategy` but `StrategyGame`; see https://specifications.freedesktop.org/menu-spec/latest/apas02.html

Also crosscheck with `freeorion.desktop` required.

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+  </keywords>
+  <metadata_license>CC-BY-SA-3.0</metadata_license>
+  <project_license>GPL-2.0</project_license>
+  <url type="bugtracker">https://github.com/freeorion/freeorion/issues</url>
+  <url type="donation">http://www.freeorion.org/index.php/Donations</url>
+  <url type="faq">http://www.freeorion.org/index.php/FAQ</url>
+  <url type="help">http://www.freeorion.org/forum/viewforum.php?f=25</url>
```
adrianbroher: The linked forum is not an `online user's reference, a software manual or help page.`, so this would probably the wiki I guess?

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-filespec

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>freeorion.desktop</id>
+  <name>FreeOrion</name>
+  <summary>Turn-based space empire and galactic conquest (4X) computer game</summary>
+  <summary xml:lang="de">Rundenbasierendes Weltraum-Strategiespiel</summary>
+  <developer_name>FreeOrion Project</developer_name>
```
adrianbroher: Do we have a formal developer name? @freeorion/developers

freeorion.appdata.xml
```diff
@@ -0,0 +1,46 @@
...
+  <url type="help">http://www.freeorion.org/forum/viewforum.php?f=25</url>
+  <url type="homepage">http://www.freeorion.org</url>
+  <screenshots>
+    <screenshot type="default">
+      <caption>Galaxy setup</caption>
+      <image height="600" width="786">http://www.freeorion.org/images/thumb/1/1a/Freeorion_0.4.5_b2015-09-01.f203162_singleplayer.JPG/786px-Freeorion_0.4.5_b2015-09-01.f203162_singleplayer.JPG</image>
+    </screenshot>
+    <screenshot>
+      <caption>Ship design</caption>
+      <image height="599" width="787">http://www.freeorion.org/images/thumb/c/c7/Freeorion_0.4.5_b2015-09-01.f203162_ingame-windows_8.JPG/787px-Freeorion_0.4.5_b2015-09-01.f203162_ingame-windows_8.JPG</image>
```
adrianbroher: Are those reliable permanent URLs?  Do we need to modify the infrastructure?

adrianbroher (changes requested):
Most of those notes are reminders for myself @Mailaender, no need to act unless you want to.

freeorion.appdata.xml
adrianbroher: `freeorion.desktop` is already correct.

freeorion.appdata.xml
adrianbroher: Fixed e7cdc82c103aa6d890189fa1e95a6bd8dec53aa5

freeorion.appdata.xml
adrianbroher: Fixed bf4bdf49ae5354c415621a9657ae63c0602bb493

freeorion.appdata.xml
adrianbroher: Fixed b3c99b8b3c6e75e8917cb6f28ef3aa72fcfde5d4

freeorion.appdata.xml
adrianbroher: Fixed 4675bd22738d693b9c59cb3f79a8b5f768232d7f

freeorion.appdata.xml
adrianbroher: Fixed 4675bd22738d693b9c59cb3f79a8b5f768232d7f

freeorion.appdata.xml
geoffthemedio: FreeOrion Project is fine, I think...

freeorion.appdata.xml
geoffthemedio: I have / am adding newer screenshots from @dbenage-cx on the forums...

http://freeorion.org/index.php/Screenshots

freeorion.appdata.xml
dbenage-cx: Per [MediaWiki manual - thumb.php](https://www.mediawiki.org/wiki/Manual:Thumb.php) looks like this specific thumbnail should be:
`http://freeorion.org/thumb.php?f=Freeorion_0.4.5_b2015-09-01.f203162_ingame-windows_8.JPG&w=787`
If a generated thumbnail is saved may depend on configuration of LocalSettings.php

Infrastructure change does not look required, but may be desirable to change filenames. I've posted a response in forum [thread](http://freeorion.org/forum/viewtopic.php?f=4&t=10676)

geoffthemedio: Can host a file directly instead of relying on wiki upload, if necessary.

freeorion.appdata.xml
Mailaender: This isn't `⁠<categories>` though.

freeorion.appdata.xml
Mailaender: You can't just change this without renaming `freeorion.desktop` as well.

freeorion.appdata.xml
Mailaender: Compare: https://github.com/freeorion/freeorion/blob/master/freeorion.desktop#L14 This may be redundant here.

freeorion.appdata.xml
Vezzra: Yep, FreeOrion Project should be ok. That's kind of our "formal developer name"...

freeorion.appdata.xml
hughsie: The idea here is that apps can specify software-center-only categories or keywords in the appdata, and the ones shared with users shell get put in the existing desktop file. Most apps don't need extra categories or keywords.

freeorion.appdata.xml
adrianbroher: @Mailaender thanks for pointing out, fixed in b52d6bc303eac68ef1c69d4adf976b55603c4602

freeorion.appdata.xml
adrianbroher: > Can host a file directly instead of relying on wiki upload, if necessary.

That's probably a better idea than to rely on some wikimedia internal URL, that may change in the next version.

freeorion.appdata.xml
adrianbroher: Fixed with b6612540af8c861d4ab5fabbe1b82fbbcfbf748f

freeorion.appdata.xml
adrianbroher: @freeorion/developers Any feedback on this?

freeorion.appdata.xml
adrianbroher: @geoffthemedio Any progress here?

freeorion.appdata.xml
dbenage-cx: Not the wiki, since there was a decision to remove anything duplicated by in-game pedia.
I'd either opt for a looser definition which allows forums/IRC/etc and change to the top level of forum, or remove the entry.

adrianbroher:
Superseded by #2826 

@geoffthemedio or @Vezzra this PR can be closed.

PR #1660, opened 2017-08-08
Hostless server

o01eg:
This PR adds server command-line options `--hostless` to launch server in hostless mode:

- Server works without host player.
- Server accepts players in multiplayer lobby and reject them when game is started.
- Players allowed to manage galaxy settings and AI players settings.
- When the game ends server with all connected players returns to multiplayer lobby with last galaxy setup and AI count.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -1200,9 +1200,11 @@ std::string HumanClientApp::SelectLoadFile() {
     return sfd->Result();
 }
 
-void HumanClientApp::ResetClientData() {
-    m_networking->SetPlayerID(Networking::INVALID_PLAYER_ID);
-    m_networking->SetHostPlayerID(Networking::INVALID_PLAYER_ID);
+void HumanClientApp::ResetClientData(bool save_connection) {
+    if (! save_connection) {
+        m_networking->SetPlayerID(Networking::INVALID_PLAYER_ID);
+        m_networking->SetHostPlayerID(Networking::INVALID_PLAYER_ID);
+    }
     SetEmpireID(ALL_EMPIRES);
     m_ui->GetMapWnd()->Sanitize();
 
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -1201,7 +1201,7 @@ std::string HumanClientApp::SelectLoadFile() {
 }
 
 void HumanClientApp::ResetClientData(bool save_connection) {
-    if (! save_connection) {
+    if (!save_connection) {
         m_networking->SetPlayerID(Networking::INVALID_PLAYER_ID);
         m_networking->SetHostPlayerID(Networking::INVALID_PLAYER_ID);
     }
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1078,7 +1078,7 @@ OPTIONS_DB_SINGLEPLAYER
 Start server in single player host mode. This only allows clients on localhost to connect.
 
 OPTIONS_DB_HOSTLESS
-Start server in hostless mode. Server doesn't require host.
+Start server in hostless mode. Server accepts players in the multiplayer lobby where returns after the game session ends.
 
 OPTIONS_DB_GENERATE_CONFIG_XML
 Uses default settings, settings from any existing config.xml file, and settings given on the command line to generate a config.xml file. This will overwrite the current config.xml file, if it exists.
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -959,7 +959,7 @@ SERVER_TURN_EVENTS_ERRORS
 Python scripted turn events executed with errors. See log files for detailed error messages. Game can continue, but gameplay will probably be impaired.
 
 SERVER_ALREADY_PLAYING_GAME
-Server doesn't accept players. The game is already started.
+The server is not currently accepting players because a game is being played.
 
 ERROR_PYTHON_AI_CRASHED
 Python AI for %1% crashed.
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1078,7 +1078,7 @@ OPTIONS_DB_SINGLEPLAYER
 Start server in single player host mode. This only allows clients on localhost to connect.
 
 OPTIONS_DB_HOSTLESS
-Start server in hostless mode. Server accepts players in the multiplayer lobby where returns after the game session ends.
+Start server in hostless mode. Server accepts players in the multiplayer lobby, and returns to the lobby after the game session ends.
 
 OPTIONS_DB_GENERATE_CONFIG_XML
 Uses default settings, settings from any existing config.xml file, and settings given on the command line to generate a config.xml file. This will overwrite the current config.xml file, if it exists.
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -958,6 +958,9 @@ Universe generation completed with errors. See log files for detailed error mess
 SERVER_TURN_EVENTS_ERRORS
 Python scripted turn events executed with errors. See log files for detailed error messages. Game can continue, but gameplay will probably be impaired.
 
+SERVER_ALREADY_PLAYING_GAME
+Server doesn't accept players. The game is already started.
+
 ERROR_PYTHON_AI_CRASHED
 Python AI for %1% crashed.
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1077,6 +1077,9 @@ Print version and exit.
 OPTIONS_DB_SINGLEPLAYER
 Start server in single player host mode. This only allows clients on localhost to connect.
 
+OPTIONS_DB_HOSTLESS
+Start server in hostless mode. Server doesn't require host.
+
 OPTIONS_DB_GENERATE_CONFIG_XML
 Uses default settings, settings from any existing config.xml file, and settings given on the command line to generate a config.xml file. This will overwrite the current config.xml file, if it exists.
 
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -279,7 +279,12 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
 ////////////////////////////////////////////////////////////
 Idle::Idle(my_context c) :
     my_base(c)
-{ TraceLogger(FSM) << "(ServerFSM) Idle"; }
+{
+    TraceLogger(FSM) << "(ServerFSM) Idle";
+    if (Server().IsHostless()) {
+        post_event(Hostless());
+    }
+}
 
 Idle::~Idle()
 { TraceLogger(FSM) << "(ServerFSM) ~Idle"; }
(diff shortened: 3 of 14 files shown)


geoffthemedio:
The concept of a dedicated server, and adjusting the CMake options to be able to build the server without requiring GUI dependencies are separate. The term for the former shouldn't be used to enable the latter.

geoffthemedio:
The concept of a dedicated server is also not the same as having a server run and games play without an identified host player.

o01eg:
@geoffthemedio I wonder what correct terminology for it.

Will it be better to split this PR?

Dilvish-fo:
I would have expected that even with a separate dedicated server, each game would still have a host, who determines galaxy setup, accepts or rejects players, and finally starts the game once the players have joined and any AI's have been set up.  Could you please explain your vision for how a game would be started here?

And I'm not sure that the bit about keeping the same server running all the time even after the players leave is really the best route.  In the bigger picture / longer term, in order to be able to handle multiple games in parallel I expect that you'd want to have a meta-server, where people go to to have a server instance launched for them, and then it seems that those instances should just terminate when their game is done, or it seems it would make a mess for the meta-server.  In the meantime, you could just have the server get relaunched after each game by a script with a loop.

o01eg:
> I would have expected that even with a separate dedicated server, each game would still have a host, who determines galaxy setup, accepts or rejects players, and finally starts the game once the players have joined and any AI's have been set up. Could you please explain your vision for how a game would be started here?

Each player have an access to those setting. As any setting change leads to reset player's ready status the game won't start until all player confirm settings.

> And I'm not sure that the bit about keeping the same server running all the time even after the players leave is really the best route. In the bigger picture / longer term, in order to be able to handle multiple games in parallel I expect that you'd want to have a meta-server, where people go to to have a server instance launched for them, and then it seems that those instances should just terminate when their game is done, or it seems it would make a mess for the meta-server. In the meantime, you could just have the server get relaunched after each game by a script with a loop.

I mostly get concept from Freeciv which, I think, have most mature multiplayer. There meta-server is a site http://meta.freeciv.org/ which only collects information about other running servers and doesn't launch or stop them.

The servers in a `Pregame` state aren't running game but awaiting for players to connect and start game.

o01eg:
Rebased and moved out build options to separate PR #1667.

o01eg:
Rebased. Fix conflicts.
Clean up AI when leave PlayingGame. Connected players goes to multiplayer lobby.
I suppose this PR finished for features.

geoffthemedio:
Update the summary with what this actually does now?

o01eg:
@geoffthemedio Updated.

geoffthemedio:
I think this concept still needs to be renamed, and perhaps split into separate "hostless" (any human player can modify game settings) and "persistent" (server reverts to the Idle state if all human players leave the game. Probably not MPLobby, as I don't see a reason to not allow it to support a single player game as well.

o01eg:
> I think this concept still needs to be renamed, and perhaps split into separate "hostless" (any human player can modify game settings)

If it's going to be a separate PR then it should work with "hostness" game when host allows other human player modify game settings. So I think "hostless" is a wrong term for it.

>  and "persistent" (server reverts to the Idle state if all human players leave the game.

I've never met "persistent" term with same meaning.  May be better "hostless" or even "headless" (I met last term in freeciv and factorio).

> Probably not MPLobby, as I don't see a reason to not allow it to support a single player game as well.
MPLobby accepts Join message while Idle.

If someone wants to play single player game at dedicated server he can just press "Start button" while he only human at this server.

Voker57:
"headless" or "dedicated" are terms related to detaching server from game process. "leaderless" and "persistent" for suggested modes sound good to me.

o01eg:
I think "persistent" term is better to use for some state storages between the server's launchs.

o01eg:
> split into separate "hostless" (any human player can modify game settings)

Also I think it will be temporary solution before complete role-based authorization will be implemented.

geoffthemedio:
"persistent" can be applied in more than one context.

I'm unconvinced that anything is necessary or particularly useful beyond having either just one (the host) or all human players be allowed to modify the setting during game setup. (Although the host doesn't necessarily need to be connected to a server for the other players to start a game.)

o01eg:
I'll split galaxy settings editable by everyone into separate PR which will provide an option for host to allow/disallow access to galaxy setting for other. In normal mode it will be default-disable and host decides what to do, in dedicated mode it will be always enabled.

geoffthemedio:
as above, don't call it dedicated.

o01eg:
Rebased over https://github.com/freeorion/freeorion/pull/1687
Rename it into "hostless" term.

geoffthemedio:
If there is no host, can the anyone-can-edit option be either on or off?

If it's off, is there a way to set them by the person starting up the server? Does the server remember these settings when returning to MP lobby after a game?

o01eg:
> If there is no host, can the anyone-can-edit option be either on or off?

Now it always on in hostless mode. And no one can switch it off.

> If it's off, is there a way to set them by the person starting up the server?

It's good idea to storage them somewhere and have predefined settings but I think it can be implemented later.

> Does the server remember these settings when returning to MP lobby after a game?

Oh, I look forward to reuse GalaxySetupData from PlayingGame state.

default/stringtables/en.txt
```diff
@@ -1077,6 +1077,9 @@ Print version and exit.
 OPTIONS_DB_SINGLEPLAYER
 Start server in single player host mode. This only allows clients on localhost to connect.
 
+OPTIONS_DB_HOSTLESS
+Start server in hostless mode. Server doesn't require host.
```
geoffthemedio: Should elaborate on this... particularly that it will make the server revert to MPLobby after a game, rather than exiting and requiring a restart.

o01eg: Detailed description.

o01eg:
Rebased over https://github.com/freeorion/freeorion/pull/1687.
Use `GalaxySetupData` from played game in multiplayer lobby.
Introduces non-default move-constructor for `GalaxySetupData` and `= default` for assign and copy-constructor.

default/stringtables/en.txt
```diff
@@ -1078,7 +1078,7 @@ OPTIONS_DB_SINGLEPLAYER
 Start server in single player host mode. This only allows clients on localhost to connect.
 
 OPTIONS_DB_HOSTLESS
-Start server in hostless mode. Server doesn't require host.
+Start server in hostless mode. Server accepts players in the multiplayer lobby where returns after the game session ends.
```
geoffthemedio: ...lobby, and returns to the lobby after the game session ends.

o01eg: Rephrased description.

geoffthemedio:
In a 2 player multiplayer game, if a player resigns, that player is sent back to the intro screen, whereas the other is sent to the lobby. Seems a bit odd.

The layout still needs tweaking... Extra padding space between the file button and the check box.
check_position_layout

The tooltip issue is possibly not specific to the checkbox. None of the widgets on the multiplayer setup screen seem to have them, even when the equivalents on the single player galaxy setup screen do.

o01eg:
> In a 2 player multiplayer game, if a player resigns, that player is sent back to the intro screen, whereas the other is sent to the lobby. Seems a bit odd.

`Resign` disconnects from the server so only other players remain on the server.

> The layout still needs tweaking... Extra padding space between the file button and the check box.

I'll add them in appropriate PR.

o01eg:
Add error for someone trying to join to playing game.

default/stringtables/en.txt
```diff
@@ -958,6 +958,9 @@ Universe generation completed with errors. See log files for detailed error mess
 SERVER_TURN_EVENTS_ERRORS
 Python scripted turn events executed with errors. See log files for detailed error messages. Game can continue, but gameplay will probably be impaired.
 
+SERVER_ALREADY_PLAYING_GAME
+Server doesn't accept players. The game is already started.
```
geoffthemedio: The server is not currently accepting players because a game is being played. Players can only join when the server is in the multiplayer lobby.

server/ServerFSM.cpp
```diff
@@ -234,7 +234,7 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
                      << ", named \"" << player_connection->PlayerName() << "\".";
 
     bool must_quit = false;
-    if (id == ALL_EMPIRES) {
+    if (id == ALL_EMPIRES && (! m_server.IsHostless())) {
```
geoffthemedio: remove space after `!`

default/stringtables/en.txt
o01eg: I suppose second part can be skipped.

server/ServerFSM.cpp
```diff
@@ -371,36 +380,110 @@ sc::result Idle::react(const Error& msg) {
...
+    }
+}
+
 MPLobby::MPLobby(my_context c) :
     my_base(c),
-    m_lobby_data(new MultiplayerLobbyData()),
-    m_server_save_game_data(new ServerSaveGameData())
+    m_lobby_data(new MultiplayerLobbyData(std::move(Server().m_galaxy_setup_data))),
+    m_server_save_game_data(new ServerSaveGameData()),
+    m_ai_count(1)
```
geoffthemedio: this line doesn't seem to do what it's supposed to... I get default MP lobbies with no AIs in them

o01eg: AI players doesn't stored in `GalaxySetupData` so they cann't be reused. Only galaxy options and game rules are saved,

client/human/HumanClientApp.cpp
```diff
@@ -1176,9 +1176,11 @@ std::string HumanClientApp::SelectLoadFile() {
     return sfd->Result();
 }
 
-void HumanClientApp::ResetClientData() {
-    m_networking->SetPlayerID(Networking::INVALID_PLAYER_ID);
-    m_networking->SetHostPlayerID(Networking::INVALID_PLAYER_ID);
+void HumanClientApp::ResetClientData(bool save_connection) {
+    if (! save_connection) {
```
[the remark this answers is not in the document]
o01eg: Fixed.

server/ServerFSM.cpp
o01eg: Only AI count could be reused from connection information.

geoffthemedio: The AI count is reproduced when going back to the lobby, but the issue is that the initial lobby state has 0 AIs, even though it looks like it's supposed to be 1 AI by default based on this line...

o01eg: `m_ai_count` is supposed to replace static `AI_count` which uses for AI name's postfix. Maybe it better to rename this variable.

server/ServerFSM.cpp
```diff
@@ -455,6 +459,21 @@ MPLobby::MPLobby(my_context c) :
                     player_setup_data.m_starting_species_name = sm.SequentialPlayableSpeciesName(player_id);
 
                 m_lobby_data->m_players.push_back(std::make_pair(player_id, player_setup_data));
+            } else if (player_id != Networking::INVALID_PLAYER_ID && player_connection->GetClientType() == Networking::CLIENT_TYPE_AI_PLAYER) {
+                PlayerSetupData player_setup_data;
+                player_setup_data.m_player_id =     -1;
```
geoffthemedio: why the hard-coded / magic `-1` and not `Networking::INVALID_PLAYER_ID`?

PR #1664, opened 2017-08-09
Rules and ValeRefs

geoffthemedio:
A collection of commits I want to test build before pushing to the GitHub master.

Adds some rules, and some ValueRef features.


Dilvish-fo (commented):
Edited to summarize/simplify after Geoff's clarification:
I had misunderstood [this commit](https://github.com/freeorion/freeorion/pull/1664/commits/3863c33875b735d11a9cdaf086066008e847441a) to be proposing an additional/new cost multiplier on colony pods, which I objected to and so asked for a forum discussion on it.

geoffthemedio:
Are you assuming `SHIP_PART_COST_MULTIPLIER` contains something it doesn't?
```
SHIP_PART_COST_MULTIPLIER
'''(GameRule name = "RULE_SHIP_PART_COST_FACTOR")'''
````
As this is a ship part's cost being defined, and the rule being added is for the cost of ship parts, I don't see why adding this multiplier would be at all controversial, or why there would ever have been a discussion about it. Perhaps you're thinking of
````
FLEET_UPKEEP_MULTIPLICATOR
'''(1 + 0.01 * ShipDesignsOwned empire = Source.Owner)'''
````

Dilvish-fo:
> Are you assuming SHIP_PART_COST_MULTIPLIER contains something it doesn't?

well, I would say that I made an inference that turned out to be incorrect, though I'll admit a bad inference doesn't work out much better than a bad assumption.  I saw a commit message that talked about adding "missing" cost multipliers which were themselves not defined in that commit, I checked the other commit messages for any mention of parts cost, and then I made the bad inference that it was referring to the parts cost multiplier I had some knowledge of (but had apparently forgotten the actual name of).  Perhaps I was still sensitized to the possibility of hidden/semi-hidden cost changes; I'll try to finish getting over that.

Do I understand correctly now that this is simply a boolean on/off rule about whether ship parts have a cost multiplier at all?

> I don't see... why there would ever have been a discussion about it.

If I am now understanding correctly, then it may not need an actual discussion, right.  But since I think that we would expect to have a general plan to at some point add AI support for this and similar rules, it would help greatly if when people submit game rules/changes like this (and anything else that would affect the AI) that they also provide a summary of the rule for the AI team to refer to.  In the past that used to generally take place via discussions on the forums, but I'll take a look at setting up a task list in an AI Gameplay Update Issue about updating the AI specifically with respect to gameplay changes, not every 'make the AI better about X' task, which you or anyone else with write privileges could add directly to (and anyone else could add a comment with a notice/task that one of us could then move up into the main task list).  I'll set up an initial task referring to these rules, but it would be most helpful if you could comment here in this PR with a list/summary of the rules so far (ones in this PR and any you've already put into master).

Hmm, or perhaps that tasklist would be better done as a GitHub 'project', I'll read up on them a bit too.

geoffthemedio:
There is an interface in the game that provides basic info about all the available rules, on the single player galaxy setup tab. The name is shown, and there is a tooltip with a short description for each rule.

The `SHIP_PART_COST_MULTIPLIER` is a multipicative factor that increases or decreases the cost of ship parts. It can be any number from 0.1 to 10.0, in steps of 0.1. It only affect ship parts, not hull costs, and would plausibly be used to tweak the balance of an actual game being played. It is separate from the toggle that makes ships take exactly 1 turn and 1 PP to produce, which overrides the scripted costs entirely and is mainly intended for testing.

Vezzra:
@Dilvish-fo:

> I'll take a look at setting up a task list in an AI Gameplay Update Issue about updating the AI specifically with respect to gameplay changes

Maybe use a github project instead? Might be better for this purpose. Just an idea...

geoffthemedio:
If there are no objections, I'll merge this set soon...

Issue #1569, opened 2017-05-15
UI shows false mandatory stops on areas discovered from derelict scout


Oberlus:
### Environment

* **FreeOrion Version**: 2017-05-09.ab336ed_Test_Win32
* **Operating System**: Windows 7
* **Graphic card used**: Nvidia 660 GTX
* **Fetched as**
  * Binary release
  * Weekly development build

### Description
On an area discovered from a derelict scout, the UI shows mandatory stops on star systems as if they where still unexplored:
![mandatory_stops](https://cloud.githubusercontent.com/assets/28650914/26055575/c89517a6-3971-11e7-9c2e-c643314c36d0.jpg)
The correct stops for that route are shown correctly (the unnecessary stops are cleared out) after next turn.

### Expected Result
The stops labelled 3 and 5 in the image should not be there. That would be the expected stops if the area wasn't discovered. Quoting MatGB: "For some reason the pathfinding algorithm doesn't know that those systems are now known so it wants you to stop at all these 'unknown space' sections".

MatGB:
Related/part of.
![screenshot_20170602_140441](https://cloud.githubusercontent.com/assets/11533206/26726835/771e8d42-479c-11e7-84ee-1477585090b3.png)
The system at Vanant y was hidden and "unexplored region" due to the ion cloud, it moved away so I know I can safely move my destroyer up to take out a dyson forest, but the destroyer is marked as stopping (and will stop at) a now visible system.

Suspect that the timing of effects is after the routefinder mapping, beyond that no clue.

PR #1666, opened 2017-08-12
Add a post-effect, pre-obstruction, vis update

Dilvish-fo:
Adds a visibility update
  after combat resolution and effect application, but
  before determination of system obstructions for empires,
  to prevent certain visibility-related obstructions from
  persisting for an extra turn past the point at which the
  visibility issue was resolved.

This PR addresses the problems cited in Issue #1569 .  

The current sequence of relevant actions is that we have 
  - a fleet movement phase followed by 
  - a visibility update and 
  - combat resolution, and then 
  - another visibility update 
(all the above still using meter values from the prior turn), 
and then finally we have 
  - effects application and meter updates
followed by
  - system obstruction determination
and various more steps, eventually including advancing the turn number and a final visibility update

The fact that system obstruction is currently determined based on pre-effect visibility status means that any obstructions due entirely to the system not having been previously sufficiently visible will currently be cleared on the ***same turn*** the system becomes sufficiently visible if that is due purely to fleet movement and previous-turn meter values, but currently require an ***extra turn*** to clear if the visibility is gained through newly processed effects like Derelict Scout or through meter changes like the fading of ion cloud obscuration.

Besides affecting fleet travel as noted in the cited issue, this also affects supply propagation (though I think it is only in a much more restrictive set of circumstances that the supply propagation issue could ever matter).

I think a reasonable resolution is to add a visibility update after effects application and meter updates but before system obstruction determination.   I have confirmed that this fixes the extra turn of fleet movement blockage in the Derelict Scout case (and believe that it would also resolve it for the moving ion cloud case).

This does have at least the possibility of changing behavior other than that cited in the issue, due to the visibility being recorded as changed one turn sooner in these cases, but besides the supply issue noted above which I don't think is a problem I cannot think of any such cases in current content, and regardless, in all such cases the new behavior would be harmonizing the timing of results obtained from visibility-changes-due-purely-to-movement with those from visibility-changes-due-to-effects-and-meter-changes, which seems appropriate to me.

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -2943,6 +2943,10 @@ void ServerApp::PostCombatProcessTurns() {
 
     DebugLogger() << "ServerApp::PostCombatProcessTurns empire resources updates";
 
+    // now that we've had combat and applied Effects, update visibilities again, prior
+    //  to updating system obstructions below.
+    m_universe.UpdateEmpireObjectVisibilities();
+    m_universe.UpdateEmpireLatestKnownObjectsAndVisibilityTurns();
 
     // Determine how much of each resource is available, and determine how to
     // distribute it to planets or on queues


geoffthemedio:
sounds reasonable...

Dilvish-fo:
OK, well that's good enough for me, I'm going to merge this now to be sure it's in this week's test build so it will get a bit broader testing before we finalize the 4.7.1 bugfix release

PR #1667, opened 2017-08-12
Headless compilation.

o01eg:
Splitted part of #1660. Adds configure option to build only server and AI binaries and some fixes for GUI-less environment.
Looks like it usable with `--external-server-address` client's option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,9 +30,10 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+option(BUILD_SHARED_LIBS "Build project as shared libraries." ON)
 option(BUILD_TESTING "Build the testing tree." OFF)
 option(FO_LINK_STATIC_BOOST_LIBS "Link static boost libraries." ${APPLE})
-
+option(BUILD_HEADLESS "Build only headless components: server and AI." OFF)
 
 if(BUILD_TESTING)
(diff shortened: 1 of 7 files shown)


geoffthemedio:
I suspect @adrianbroher would have some comments / concerns?

What are the "fixes for GUI-less environment"?

o01eg:
It's for those systems which haven't any graphic libraries installed.

geoffthemedio:
What, not for what.

o01eg:
There is a second commit which removes `"../UI/LinkText.h"` from server dependencies.

geoffthemedio:
Is that just removing an unnecessary header include? Presumably that's not specific for a GUI-less environment.

o01eg:
`UI/LinkText.h` requires `glew` include files.

geoffthemedio:
Sure, but does removing that include break anything, or is it entirely unnecessary regardless of whether the human client is also being built? If removing it means the server will not be fully functional if a human graphical client connects to it, that's also a problem.

adrianbroher:
TL;DR: LGTM

> I suspect @adrianbroher would have some comments / concerns?

No objections,  the commit looks ok, albeit ugly.  But that's because noone is willing break up into the source into the components the software actually consists of.

> Sure, but does removing that include break anything, or is it entirely unnecessary regardless of whether the human client is also being built?

Why should it?  Any symbols provided by this include are not used.  If there were symbols used the compiler would complain about the missing declaration of those symbols.

geoffthemedio:
I was just guessing there might have been a reason for the include being there. If it'd really unnecessary then there's no reason to keep it seemingly.

PR #1665, opened 2017-08-10
Improve python error handling during game shutdown and crashes.

LGM-Doyle:
This PR fixes issue [#1659 insufficient exception logging in AI scripts.](https://github.com/freeorion/freeorion/issues/1659)

It fixes the following issues with startup, shutdown and error reporting related to the python engine.
- adds a parameter `is_error` to `ResetToIntro()` so that in progress saves can be short circuited when the reset is caused by an error.
- refactors python initialization in the server, so that failure to initialize will cause the server to shutdown. The server shutdown is delayed until after the AI client have connected to allow for a orderly shutdown of the AI clients.
- detects disconnection of the AI clients before they are assigned an ID.
- in `_stdXLikeStream.write()` from `configure_logging.py`  it detects when streamed output is not associated with a python execution stack and does not print a filename and location for the output.  This happens with errors trapped in the C++ code and then printed with `PyErr_Print()`.

I've included the patch used to test various methods of generating errors/exiting the AI/server python processes, to document what is expected to work.

Each clause in the patch exercises a different method of causing an exit/exception in the python code, traversing boost python, the AI, server and human client executables.  

All methods, except those that use python `os` to kill the process, log the error.   Methods that are not within a structure (like `try ..catch`) that recovers from the error, shutdown gracefully if fatal.  Using python `os` to kill the AI process shutsdown quietly, but correctly.  Using python `os` to kill the server process, strands the AI processes.  

````diff
diff --git a/default/python/AI/FreeOrionAI.py b/default/python/AI/FreeOrionAI.py
index fe45735..72cc31d 100644
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -12,6 +12,25 @@ import random
 
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client  # pylint: disable=import-error
 
+# debug("Testing exit handling messaging in FreeOrionAI imports")
+# debug("About to RaiseException(\"test\")")
+# raise Exception('test')
+# debug("About to exit(99)")
+# exit(99)
+# debug("About to exit(0)")
+# exit(0)
+# debug("About to sys.exit(0)")
+# sys.exit(0)
+# debug("About to SystemExit()")
+# raise SystemExit()
+# debug("About to quit()")
+# quit()
+# debug("About to os.abort(). Note: The log will end abruptly.")
+# import os; os.abort()
+# debug("About to os.system('exit'). Note: This should do nothing.")
+# import os; os.system('exit')
[...]

PR #1670, opened 2017-08-13
Gate initial and last turn autosaves with OptionsDB options.

LGM-Doyle:
This addresses the concern that the autoquit autosave was not gated by an option, raised here https://github.com/freeorion/freeorion/pull/1653#issuecomment-320354748.

It also adds an option to gate the initial turn autosave.

The options are named:
- autosave.initial-turn
- autosave.last-turn

This PR depends on #1665, because they both make changes to the shutdown sequence.

### DIFF
--- a/UI/OptionsWnd.cpp
+++ b/UI/OptionsWnd.cpp
@@ -692,6 +692,8 @@ void OptionsWnd::CompleteConstruction() {
     BoolOption(current_page, 0, "autosave.multiplayer",   UserString("OPTIONS_MULTIPLAYER"));
     IntOption(current_page,  0, "autosave.turns",         UserString("OPTIONS_AUTOSAVE_TURNS_BETWEEN"));
     IntOption(current_page,  0, "autosave.limit",         UserString("OPTIONS_AUTOSAVE_LIMIT"));
+    BoolOption(current_page, 0, "autosave.initial-turn",  UserString("OPTIONS_AUTOSAVE_INITIAL_TURN"));
+    BoolOption(current_page, 0, "autosave.last-turn",     UserString("OPTIONS_AUTOSAVE_LAST_TURN"));
     m_tabs->SetCurrentWnd(0);
 
     // Keyboard shortcuts tab
--- a/UI/OptionsWnd.cpp
+++ b/UI/OptionsWnd.cpp
@@ -688,12 +688,12 @@ void OptionsWnd::CompleteConstruction() {
 
     // Ausosave settings tab
     current_page = CreatePage(UserString("OPTIONS_PAGE_AUTOSAVE"));
-    BoolOption(current_page, 0, "autosave.single-player", UserString("OPTIONS_SINGLEPLAYER"));
-    BoolOption(current_page, 0, "autosave.multiplayer",   UserString("OPTIONS_MULTIPLAYER"));
-    IntOption(current_page,  0, "autosave.turns",         UserString("OPTIONS_AUTOSAVE_TURNS_BETWEEN"));
-    IntOption(current_page,  0, "autosave.limit",         UserString("OPTIONS_AUTOSAVE_LIMIT"));
-    BoolOption(current_page, 0, "autosave.initial-turn",  UserString("OPTIONS_AUTOSAVE_INITIAL_TURN"));
-    BoolOption(current_page, 0, "autosave.last-turn",     UserString("OPTIONS_AUTOSAVE_LAST_TURN"));
+    BoolOption(current_page, 0, "autosave.single-player",   UserString("OPTIONS_SINGLEPLAYER"));
+    BoolOption(current_page, 0, "autosave.multiplayer",     UserString("OPTIONS_MULTIPLAYER"));
+    IntOption(current_page,  0, "autosave.turns",           UserString("OPTIONS_AUTOSAVE_TURNS_BETWEEN"));
+    IntOption(current_page,  0, "autosave.limit",           UserString("OPTIONS_AUTOSAVE_LIMIT"));
+    BoolOption(current_page, 0, "autosave.galaxy-creation", UserString("OPTIONS_DB_AUTOSAVE_GALAXY_CREATION"));
+    BoolOption(current_page, 0, "autosave.game-close",      UserString("OPTIONS_DB_AUTOSAVE_GAME_CLOSE"));
     m_tabs->SetCurrentWnd(0);
 
     // Keyboard shortcuts tab
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -642,7 +642,8 @@ void HumanClientApp::SaveGame(const std::string& filename) {
 void HumanClientApp::SaveGameCompleted() {
     m_game_saves_in_progress.pop();
 
-    // Either indicate that all save are completed or start the next save.
+    // Either indicate that all saves are completed or start the next save.
+    // Autosaves and player saves can be concurrent.
     if (m_game_saves_in_progress.empty()) {
         DebugLogger() << "Save games completed.";
         SaveGamesCompletedSignal();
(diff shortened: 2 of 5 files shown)


client/human/HumanClientApp.cpp
```diff
@@ -625,16 +627,29 @@ void HumanClientApp::CancelMultiplayerGameFromLobby()
...
+    DebugLogger() << "Sent save initiate message to server.";
 }
 
 void HumanClientApp::SaveGameCompleted() {
-    DebugLogger() << "HumanClientApp::SaveGameCompleted by server.";
-    m_save_game_in_progress = false;
-    SaveGameCompletedSignal();
+    m_game_saves_in_progress.pop();
+
+    // Either indicate that all save are completed or start the next save.
```
geoffthemedio: `save` -> `saves`

When would there be multiple saves in progress simultaneously?

LGM-Doyle: In the current game if the user saves while the autosave is in progress.

It only matters if the user saves and then immediately exits.  In the current game that means only the first save that completes is kept, the autosave.  The user save is always aborted.  

With PR #1671, the scenario would happen whenever the user saves and then immediately quits.  Although in that case the last turn autosave would be lost.

geoffthemedio:
autosave_options

Seems to work as intended.

The first turn autosave is labelled as being from turn 1, so the in-game description should match that.

There is a missing string for the description in the tooltip.

I'd rename the options "Autosave on first turn" and "Autosave when resigning or closing".

The descriptions should also not just reiterate the name label... so "On first turn of a new game, generate an autosave of the initial game state, before any player actions are taken" and "When resigning a game or closing the application window, generate an autosave" or similar.

UI/OptionsWnd.cpp
```diff
@@ -692,6 +692,8 @@ void OptionsWnd::CompleteConstruction() {
     BoolOption(current_page, 0, "autosave.multiplayer",   UserString("OPTIONS_MULTIPLAYER"));
     IntOption(current_page,  0, "autosave.turns",         UserString("OPTIONS_AUTOSAVE_TURNS_BETWEEN"));
     IntOption(current_page,  0, "autosave.limit",         UserString("OPTIONS_AUTOSAVE_LIMIT"));
+    BoolOption(current_page, 0, "autosave.initial-turn",  UserString("OPTIONS_AUTOSAVE_INITIAL_TURN"));
+    BoolOption(current_page, 0, "autosave.last-turn",     UserString("OPTIONS_AUTOSAVE_LAST_TURN"));
```
geoffthemedio: maybe rename these? it's not just an initial turn autosave, it's a before anything happens initial gamestate autosave. It's also not a last turn autosave, it's an autosave when closing or resigning.

default/stringtables/en.txt
```diff
@@ -1638,6 +1638,12 @@ Sets the number of turns between autosaves.
 OPTIONS_DB_AUTOSAVE_LIMIT
 Sets the maximum number of autosave files to keep.
 
+OPTIONS_DB_AUTOSAVE_INTIAL_TURN
+If true, autosave the initial turn before any game play.
```
geoffthemedio: Prefer "Toggles whether to" rather than "If true, "

default/stringtables/en.txt
```diff
@@ -2267,6 +2273,12 @@ Autosaves limit
 OPTIONS_AUTOSAVE_TURNS_BETWEEN
 Turns between autosaves
 
+OPTIONS_AUTOSAVE_INITIAL_TURN
+Autosave on turn 0 before any game play
```
geoffthemedio: turn 1 I think?

UI/OptionsWnd.cpp
geoffthemedio: should there be a distinction between galaxy creation and everything that happens before the start of the first turn?

LGM-Doyle: There is a distinction in the server side code, but the programmer can see that the save is generated in the human client FSM `WaitingForGameStart` when the first turn game data is extracted.

I don't think there is a distinction to a player.

I forgot to change the variable names in the non-UI code in `HumanClientApp`, so I have re-written the last commit.

geoffthemedio: But the player might see the distinction if/when we make it possible to save and edit galaxy maps (at various stages of being populated with stuff) independently from saved games...

geoffthemedio:
I consistently get crashes from a null pointer exception when starting a multiplayer game with two players. The stack:
````
>	FreeOrion.exe!HumanClientApp::SaveGameCompleted() Line 645	C++
 	FreeOrion.exe!PlayingTurn::react(const SaveGameComplete & msg) Line 898	C++
 	FreeOrion.exe!boost::statechart::simple_state<PlayingTurn,PlayingGame,boost::mpl::list<boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na>,0>::local_react_impl_non_empty::local_react_impl<boost::mpl::list<boost::statechart::custom_reaction<SaveGameDataRequest>,boost::statechart::custom_reaction<SaveGameComplete>,boost::statechart::custom_reaction<AdvanceTurn>,boost::statechart::custom_reaction<TurnUpdate>,boost::statechart::custom_reaction<TurnEnded>,boost::statechart::custom_reaction<PlayerStatus>,boost::statechart::custom_reaction<DispatchCombatLogs>,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na>,boost::statechart::simple_state<PlayingTurn,PlayingGame,boost::mpl::list<boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na>,0> >(boost::statechart::simple_state<PlayingTurn,PlayingGame,boost::mpl::list<boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na>,0> & stt, const boost::statechart::event_base & evt, const void * eventType) Line 820	C++
 	FreeOrion.exe!boost::statechart::simple_state<PlayingTurn,PlayingGame,boost::mpl::list<boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na,boost::mpl::na>,0>::react_impl(const boost::statechart::event_base & evt, const void * eventType) Line 489	C++
 	FreeOrion.exe!boost::statechart::state_machine<HumanClientFSM,IntroMenu,std::allocator<void>,boost::statechart::null_exception_translator>::send_event(const boost::statechart::event_base & evt) Line 881	C++
 	FreeOrion.exe!boost::statechart::state_machine<HumanClientFSM,IntroMenu,std::allocator<void>,boost::statechart::null_exception_translator>::process_event(const boost::statechart::event_base & evt) Line 275	C++
 	FreeOrion.exe!HumanClientApp::HandleMessage(Message & msg) Line 886	C++
 	FreeOrion.exe!HumanClientApp::HandleSystemEvents() Line 865	C++
 	GiGi.dll!GG::ModalEventPump::operator()() Line 114	C++
 	GiGiSDL.dll!GG::SDLGUI::Run() Line 883	C++
 	FreeOrion.exe!mainSetupAndRun() Line 258	C++
 	FreeOrion.exe!wmain(int argc, wchar_t * * argv, wchar_t * * envp) Line 104	C++
````
Seems like the non-host player is getting the `SaveGameComplete` message after not having initiated a save and trying to pop an empty stack?

Just putting `if (!m_game_saves_in_progress.empty())` before the stack pop in `HumanClientApp::SaveGameCompleted()` avoids the crash.

geoffthemedio:
I also get an error message in game (no crash) when resigning from a game with no hosts using the hostless branch. It tries to save on exit, but can't because it's not the host. I assume this would also happen if the previous crash didn't in master.

PR #1671, opened 2017-08-13
Add --continue game startup option.

LGM-Doyle:
I've cross posted this to the forum, as http://freeorion.org/forum/viewtopic.php?f=6&t=10674

This PR adds a "--continue" command line option and a "Continue" button to the intro menu.

It depends on #1670.  I made it a separate PR so that it could be discussed separately on the forum.

To continue means to find the newest single player autosave, load it and continue play.  The "--continue" option skips the intro menu the same as the "--quickstart" option.  

The "Continue" button on the main menu is enabled by the autosave.last-turn option added in #1670.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -671,6 +671,10 @@ void HumanClientApp::LoadSinglePlayerGame(std::string filename/* = ""*/) {
         try {
             auto sfd = GG::Wnd::Create<SaveFileDialog>(SP_SAVE_FILE_EXTENSION, true);
             sfd->Run();
+
+            // Update intro screen Load & Continue buttons if all savegames are deleted.
+            m_ui->GetIntroScreen()->RequirePreRender();
+
             if (!sfd->Result().empty())
                 filename = sfd->Result();
         } catch (const std::exception& e) {
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -1206,9 +1206,10 @@ void HumanClientApp::Autosave() {
 
     auto autosave_file_path = CreateNewAutosaveFilePath(EmpireID(), m_single_player_game);
 
-    // check for and remove excess oldest autosaves
+    // check for and remove excess oldest autosaves.  The minimum is 1 for the
+    // initial or final save.
     boost::filesystem::path autosave_dir_path(GetSaveDir() / "auto");
-    int max_autosaves = GetOptionsDB().Get<int>("autosave.limit");
+    int max_autosaves = std::max(1, GetOptionsDB().Get<int>("autosave.limit"));
     RemoveOldestFiles(max_autosaves, autosave_dir_path);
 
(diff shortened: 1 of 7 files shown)


LGM-Doyle:
I made the following changes in response to the discussion on the forum:
- the continue function now loads the newest single player savegame of either type autosave or player save.
- the "Continue" and "Load" buttons on the `IntroScreen` are only present when there is at least one available save game.  The buttons will disappear if the user deletes all save games from within the game, but not if the user deletes all save games from another app.

default/stringtables/en.txt
```diff
@@ -1638,6 +1638,12 @@ Sets the number of turns between autosaves.
 OPTIONS_DB_AUTOSAVE_LIMIT
 Sets the maximum number of autosave files to keep.
 
+OPTIONS_DB_AUTOSAVE_INTIAL_TURN
+If true, autosave the initial turn before any game play.
```
geoffthemedio: It is slightly clearer than "toggles", but the "if true, stuff" idiom is just too awkward to actually use, in my opinion...  At least switch to "Enables an autosave on the initial turn..." or similar.

LGM-Doyle: I changed the wording to "Enable single-player autosaves." etc.

LGM-Doyle (commented):
I've fixed both of the bugs reported in #1670:
- the pop on empty stack bug described [here](https://github.com/freeorion/freeorion/pull/1670#issuecomment-324844247)
- non-host players initiating autosaves in multiplayer, described [here](https://github.com/freeorion/freeorion/pull/1670#issuecomment-324847139)

I also changed it so that it does **not** autosave when the player loads another game.  This prevents it from generating new autosaves, when the player is searching through the existing autosaves for a particular game.

geoffthemedio:
Need another rebase.

Also, given that https://github.com/freeorion/freeorion/pull/1660 is merged, perhaps there should be an autosave initiated automatically when there is no host player?

LGM-Doyle:
Rebased.

Currently, the save piggybacks on the infrastructure for the single player/host save game.  Adding autosave to a headless game requires refactoring of the clients and the server code.

client/human/HumanClientApp.cpp
```diff
@@ -1054,6 +1058,45 @@ void HumanClientApp::UpdateCombatLogManager() {
...
+                        continue;
+
+                    std::time_t t = last_write_time(file_path);
+                    files_by_write_time.insert({t, file_path});
+                }
+            };
+
+            // Find all save games in either player or autosaves
+            add_all_savegames_in(GetSaveDir());
+            add_all_savegames_in(GetSaveDir() / "auto");
```
geoffthemedio, line 1086: An issue with this bit is that it only looks in the root save dir and root/auto, not any other save directory subdirectories. This means that if a player wants to use directories to organize their saves, they can't. Previously one could open the load dialog and look into the subdirectories (or random other places, but that's less important).

PR #1675, opened 2017-08-15
Fix Compile Errors

geoffthemedio:
Hopefully fix GCC compile error: invalid initialization of non-const reference of type 'std::shared_ptr<UniverseObject>&' from an rvalue of type 'std::shared_ptr<UniverseObject>'


geoffthemedio:
Dealt with by @LGM-Doyle in https://github.com/freeorion/freeorion/commit/ef3720646dfdde65d1b378ff6e6532cda26048a9

LGM-Doyle:
I pulled master it was broken but there was no issue.  I hoped that I could push the fix before anyone noticed that it was broken (en route to fixing the missing `MultiEdit::CompleteConstruction()` which no one has yet reported was broken).

PR #1676, opened 2017-08-15
Species adjustments for balance

Voker57:
During lengthy multiplayer sessions we figured these ways to make Egassem & Gysache less steamrolling and Laenfa somewhat playable.


Dilvish-fo:
That really only makes Laenfa "somewhat playable"?  Is that only in comparison to Egassem and Gysache or in comparison to the other playable races as well?

MatGB:
Laenfa in single player against the AI are overpowered due to stealth protection but otherwise a bit meh. In multiplayer they're awful, humans know to research detection and how to get around stealth, sorting them out to be a better overall experience has been a longterm goal.

In single player, an easy win is Laenfa in a young galaxy, in multiplayer I don't think I'd want to play them at all.

Voker57:
@Dilvish-fo It's a challenging species to play but has an unique bonus. 

Jury's still out on precise overall balance but with these patches getting these species does not mean game over/easy win.

Vezzra:
Sounds reasonable. Also take into account, while Leanfa might already be overpowered in single player, being so underwhelming in multiplayer means that in single player AIs starting with Laenfa will severely suck too.

So improving the Laenfa might make them even more overpowered in single player for the human player, but it also makes them better for the AI players.

Not so sure about reducing the industry bonus of the Egassem, because their ultimate industry is pretty much what sets them apart (is their unique characteristic). Taking that away makes them kind of boring.

Although I understand the issue with them being extremely overpowered once they acquire other species which cancel out their weaknesses. So as a stop-gap solution that's probably fine. However, once Influence gets implemented and mechanics based on that which make assimilation of other species much more difficult, we should change that back.

geoffthemedio:
@Dilvish-fo @Vezzra @MatGB Are there some conclusions about what to do with this? OK as is? Request changes?

Vezzra:
I'd say merge as is, but keeping a reminder somewhere (I guess by opening an issue) that Egassem should be put back to ultimate industry once Influence is implemented and acquiring species into your empire has been made more difficult.

@MatGB, @Dilvish-fo? Any further comments?

MatGB:
Yeah, it's not an ideal set of solutions, I'm fairly convinced the big problem with Egassem is the ground troops bonus not the industry bonus, but that whole area needs a balance and it's below research, industry and growth techs on my list to fiddle with.

So merge in and we'll reassess later. I'd been planning on opening a Project up, linking this and working through a lot more stuff but time keeps getting in the way.

Vezzra:
> but time keeps getting in the way.

Tell me about it... 😓

PR #1677, opened 2017-08-15
Replace iterator types with auto

geoffthemedio:
Replace iterator types with auto


Vezzra:
@geoffthemedio, don't know why you put this PR up and then closed it, so I just assumed "invalid" as reason for closing. If you think another label fits better, feel free to change it.

geoffthemedio:
The point was just to get the test builds to run. After they ran, it wasn't needed anymore.

PR #1678, opened 2017-08-15
Add `--log-file` option to redirect logs.

o01eg:
Add option `--log-file` to the server, client and AI to change default log file.

### DIFF
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -87,17 +87,21 @@ AIClientApp::AIClientApp(const std::vector<std::string>& args) :
     // read command line args
 
     m_player_name = args.at(1);
-    const std::string AICLIENT_LOG_FILENAME((GetUserDataDir() / (m_player_name + ".log")).string());
     if (args.size() >=3) {
         m_max_aggression = boost::lexical_cast<int>(args.at(2));
     }
 
+    // Force the log file if requested.
+    if (GetOptionsDB().Get<std::string>("log-file").empty()) {
+        const std::string AICLIENT_LOG_FILENAME((GetUserDataDir() / (m_player_name + ".log")).string());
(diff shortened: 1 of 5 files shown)


geoffthemedio:
When is this needed or useful?

Dilvish-fo:
So, what's the use-case for this?  Is it part of your standalone server work?
I'm currently not really seeing the the benefit outweigh the clutter here.   Also, it looks like by itself this will only cause the human client log to be redirected, that to redirect the server and AI logs you would need to start those separately (and so for the AI would only work for multiplayer games).  So the help message should be more clear about that.

Also, if someone specifies the same file for them all I suppose the result might be OS dependent, but on mine (Ubuntu Linux, just setting the server log and human log the same) it tries to put them into the same file but does a poor job of it, overwriting each other in strange ways.  So you'd probably need to check if the specified logfile is already open for writing and if it is then instead just use the default logfile, with an error message written to the console, or something.

o01eg:
It mostly done to separate logs from two or more client which I run on the same system to test multiplayer but I made it same way as `log-level` so it works for server and AI as well.
About setting same log file I suppose it user's error.
However I look up to `text_file_backend` documentation and it accept only `std::ios_base::openmode` open flag which doesn't support any locks.

geoffthemedio:
Maybe it should automatically change the log file name if there is already a server or client/ui process running, to avoid mixing two process' logs into one file? I don't think there's presently any way to check for that, though.

o01eg:
@geoffthemedio Then it need a way to get which log file is used by running processes. And it requires many platform-depended code.
I think it will be good to add process id as prefix to log file. The only trouble then is to clean up old logs.

o01eg:
It sad but it looks like boost::process appeared at 1.64 version.

o01eg:
Rebased.
So will it be responsibility of those who set `log-file` to prevent collisions?

PR #1688, opened 2017-08-23
AI logger related fixes

Dilvish-fo:
These changes appear necessary to me to fix an AI logging problem from 2f4e85ed, which would potentially cause the AI to semi-silently fail when it had an error rather than reporting it to the user via chat (and in conjunction with a recent naming error resulting from a942bf8d was in fact causing the AI to semi-silently fail).

@LGM-Doyle  All or most of these logging adjustments are to changes that you very recently made, and since you're the architect of the current AI logging structure I am a bit apprehensive that I am misunderstanding something here or else I would just directly commit it since it appears to me necessary to fix the AI.  But perhaps you have some other way you'd prefer it resolved...

@geoffthemedio since they are already tangentially involved I'd like to go ahead and ask you here about the changes in a942bf8d which then in turn exposed this logging error-- in one of your "grooming" commits you changed the naming convention / names under which a few empire methods were exposed to the AI, without making corresponding changes to the AI, and thus breaking it (but because of this logging error that wouldn't have immediately shown up in testing unless you manually checked the AI logs and even then the errors would not have been clearly identified).   We can fix those naming errors either by reverting those three names or by changing the places they are used by the AI.  Since seemingly all of the other names in the EmpireWrapper are still following the other naming convention, I would be more inclined to revert the three names -- if we are going to change the naming convention we can do them all at once.  If you have a preference please let me know ASAP.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -135,7 +135,7 @@ def resumeLoadedGame(saved_state_string):  # pylint: disable=invalid-name
             # assigning new state
             foAIstate = AIstate.AIstate(fo.aggression.aggressive)
             foAIstate.session_start_cleanup()
-            error("Fail to load aiState from saved game: %s", e, exc_info=True)
+            error("Fail to load aiState from saved game: %s" % e, exc_info=True)
 
     aggression_trait = foAIstate.character.get_trait(Aggression)
     diplomatic_corp_configs = {fo.aggression.beginner: DiplomaticCorp.BeginnerDiplomaticCorp,
@@ -250,7 +250,7 @@ def generateOrders():  # pylint: disable=invalid-name
             # not invalidate doneTurn()
             fo.doneTurn()
         except Exception as e:
-            error("Exception in doneTurn() on non-existent empire", e, exc_info=True)
+            error("Exception %s in doneTurn() on non-existent empire" % e, exc_info=True)
         return
 
     if empire.eliminated:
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
@LGM-Doyle one more thing I just noticed, even with this fix/adjustment, I am now getting the full error message (including traceback) I would expect being sent to chat, but the actual log file error message is missing the traceback portion.  So it seems there is something more needing to be fixed in there somewhere.

geoffthemedio:
These changes https://github.com/freeorion/freeorion/commit/a942bf8dc78626010f6a22c7215625ee9c83115a#diff-aad399b61003043cf9a02e9b18768456L293 were unintentional when find-replacing. I meant to adjust the naming convention in C++, and didn't notice it had altered what was exposed to Python.

Please revert or adjust what things are called when exposed to Python to suit your preference or Python style guides.

LGM-Doyle:
@Dilvish-fo your changes are correct.  It is not necessary to add the exception `e` as a separate parameter.  The python logging module will fetch the exception information on its own with `sys.exc_info()`, if  `exc_info=True` is passed as a parameter.  

However, the python logger documentation implies that both lines of code should produce the same result, and it does on my machine.  This is clearly not true to all versions of python2.

Dilvish-fo:
> your changes are correct.

OK, I'll go ahead and merge this, but please note the Issue I just created, cited above, for me at least there is still a lingering problem (though at least it is non-breaking).

Issue #1680, opened 2017-08-16
Pedia Crashes


geoffthemedio:
Starting with https://github.com/freeorion/freeorion/commit/aeb1df1695c97524ee95dd5c35222b303b71808a opening the pedia immediately crashes the client. Occurs from the intro screen, or elsewhere.

Debugger's call stack:
````
>	FreeOrion.exe!std::_Copy_unchecked1<char const *,std::back_insert_iterator<std::basic_string<char,std::char_traits<char>,std::allocator<char> > > >(const char * _First, const char * _Last, std::back_insert_iterator<std::basic_string<char,std::char_traits<char>,std::allocator<char> > > _Dest, std::_General_ptr_iterator_tag __formal) Line 2316	C++
 	FreeOrion.exe!TextLinker::MarkLinks() Line 475	C++
 	FreeOrion.exe!CUILinkTextMultiEdit::CompleteConstruction() Line 1017	C++
 	FreeOrion.exe!GG::Wnd::Create<CUILinkTextMultiEdit,std::basic_string<char,std::char_traits<char>,std::allocator<char> > const &,GG::Flags<GG::MultiEditStyle> >(const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & <args_0>, GG::Flags<GG::MultiEditStyle> && <args_1>) Line 366	C++
 	FreeOrion.exe!CUILinkTextBlock::CUILinkTextBlock(const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & str, const std::shared_ptr<GG::Font> & font, GG::Flags<GG::TextFormat> format, const GG::Clr & color, GG::Flags<GG::WndFlag> flags) Line 10	C++
 	FreeOrion.exe!GG::Wnd::Create<CUILinkTextBlock,std::basic_string<char,std::char_traits<char>,std::allocator<char> > const &,std::shared_ptr<GG::Font> const &,GG::Flags<GG::TextFormat> &,GG::Clr const &,GG::WndFlag const &>(const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & <args_0>, const std::shared_ptr<GG::Font> & <args_1>, GG::Flags<GG::TextFormat> & <args_2>, const GG::Clr & <args_3>, const GG::WndFlag & <args_4>) Line 364	C++
[...]

PR #1689, opened 2017-08-23
Fix bug #1680, add `RichText::CompleteConstruction()`, make `TextLinker::FindLinks()` stable.

LGM-Doyle:
I hope that this PR fixes bug #1680.  I was not able to replicate the crash from the Pedia.  I was able to replicate the intermittent crash from the environment assessment.

This PR fixes the following two problems:
- `RichText` did not have a `CompleteConstruction()` function, so `PopulateBlocks()` was attaching blocks (children) during its construction.  
- `FindLinks()` returned different, incorrect results when run after a call to `MarkText()`.  `FindLinks()` uses the line data to generate the links, that `MarkText()` uses to set the text, which generates new line data that yields incorrect links when processed with `FindLinks()`.  Eventually, the incorrect links index out of bounds characters and cause a crash.

`MultiEdit` not having a `CompleteConstruction()` hid these problems.  That is why https://github.com/freeorion/freeorion/commit/aeb1df1695c97524ee95dd5c35222b303b71808a revealed them.

The solutions in this PR are:
- Add `RichText::CompleteConstruction()`, which involved refactoring `PopulateBlocks()` into `CreateBlocks()` and `AttachBlocks()` to be run in the constructor and `CompleteConstruction()` functions respectively.
- Call `SetLinkedText(RawText())` at the start of each call to `FindLinks()`.  The solution is not ideal since the underlying `TextControl` will make repeated calls to `ExpensiveParseFromTextToTextElements()`.  

This PR will need testing since I was not able to duplicate the Pedia crash.

### DIFF
--- a/GG/GG/RichText/RichText.h
+++ b/GG/GG/RichText/RichText.h
@@ -71,6 +71,8 @@ public:
              Clr color = CLR_BLACK, Flags<TextFormat> format = FORMAT_NONE,
              Flags<WndFlag> flags = NO_WND_FLAGS);
 
+    void CompleteConstruction() override;
+
     ~RichText();
     //@}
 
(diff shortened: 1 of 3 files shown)


geoffthemedio:
seems to work for me

Issue #572, opened 2016-03-28
Autosaving at end of turn


cpeosphoros:
Could an option be given that autosaving would occur at the end of a turn, instead or in addition of its beginning?

Mild dislexy sometimes makes me mistakenly hit end-turn instead of production, with a lot of orders already issued, and late in the game having to reload an autosave at the start of the turn and reissueing everything may be a pain.

geoffthemedio:
@LGM-Doyle You might consider this since you've been working on related features: https://github.com/freeorion/freeorion/pull/1670

Vezzra:
Implemented by #1696.

PR #1696, opened 2017-08-26
Implement feature request #572 Autosaving at end of turn

LGM-Doyle:
This PR implements feature request #572, autosaving at the beginning and end of turn for single player games.

It should follow PR #1671.

It changes when orders are deleted on the AI and human clients, from immediately when sent to the server, to after the server has confirmed that it has received the orders.  This allows orders that have been sent to the server, but not yet acknowledged to be included in the end of turn save game.

The feature is not implemented for multiplayer, because of the relative timing issues with multiple players sending orders at different times.

### DIFF
--- a/client/human/chmain.cpp
+++ b/client/human/chmain.cpp
@@ -248,7 +248,7 @@ int mainSetupAndRun() {
         if (GetOptionsDB().Get<bool>("continue")) {
             // immediately start the server, establish network connections, and
             // go into a single player game, continuing from the newest
-            // autosave game.
+            // save game.
             app.ContinueSinglePlayerGame();  // acceptable to call before app()
         }
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1640,11 +1640,17 @@ Boolean for AI testing which indicates if all AIs are forced to have the same Em
 OPTIONS_DB_AI_CONFIG_TRAIT_EMPIREID_FORCED_VALUE
 Forced value of the EmpireID Trait.  A value from 0 to 39.
 
-OPTIONS_DB_AUTOSAVE_SINGLE_PLAYER
-Enable single-player autosaves.
+OPTIONS_DB_AUTOSAVE_SINGLE_PLAYER_TURN_START
+Enable single-player autosaves at turn start.
 
-OPTIONS_DB_AUTOSAVE_MULTIPLAYER
-Enable multi-player autosaves.
+OPTIONS_DB_AUTOSAVE_SINGLE_PLAYER_TURN_END
+Enable single-player autosaves at turn end.
+
+OPTIONS_DB_AUTOSAVE_MULTIPLAYER_TURN_START
+Enable multi-player autosaves at turn start.
+
+OPTIONS_DB_AUTOSAVE_MULTIPLAYER_TURN_END
+Enable multi-player autosaves at turn end.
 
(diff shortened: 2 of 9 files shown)


client/human/chmain.cpp
```diff
@@ -244,6 +245,13 @@ int mainSetupAndRun() {
             app.NewSinglePlayerGame(true);  // acceptable to call before app()
         }
 
+        if (GetOptionsDB().Get<bool>("continue")) {
+            // immediately start the server, establish network connections, and
+            // go into a single player game, continuing from the newest
+            // autosave game.
```
geoffthemedio: misleading comment, I think? ("auto")

default/stringtables/en.txt
```diff
@@ -1667,6 +1673,9 @@ The volume (0 to 255) at which music should be played.
 OPTIONS_DB_QUICKSTART
 Starts a new quick-start game, bypassing the main menu.
 
+OPTIONS_DB_CONTINUE
+Continues play from lastest autosave, bypassing the main menu.
```
geoffthemedio: same re: "auto" ?

geoffthemedio:
What does this do with orders that have already been issued when loading a game? Is the gamestate written to the save file, and sent to players when loading the game, the same as at the start of the turn? Are players sent their already-issued orders when loading the save? Are those orders re-issued / executed by the clients, then sent back to the server again for a subsequent autosave or when ending turn? Does the server erase orders loaded from the save after sending to the players when loading, so that orders sent back to the server when ending turn don't conflict? Do AIs also get a chance to re-issue orders when loading an end-of-turn autosave, and do they understand / handle this situation correctly?

LGM-Doyle:
> What does this do with orders that have already been  issued when loading a game? 

Orders that need to be re-issued on load are re-issued in `MapWnd::InitTurn()`.  This is not changed in this PR.

> Is the gamestate written to the save file,  and sent to players when loading the game, the same as at the start of  the turn? 

Yes the server side game state is the same.  The client information may be different.

> Are players sent their already-issued orders when loading the  save? 

Yes.

> Are those orders re-issued / executed by the clients, then sent  back to the server again for a subsequent autosave or when ending turn?  

Yes.  They are re-issued in `MapWnd::InitTurn()`.

> Does the server erase orders loaded from the save after sending to the  players when loading, so that orders sent back to the server when ending  turn don't conflict? 

The server side orders are not part of the save state.

> Do AIs also get a chance to re-issue orders when  loading an end-of-turn autosave, and do they understand / handle this  situation correctly?

The loaded orders are not saved on the server, only sent to the clients.
The AIs always recompute and send new orders.
Humans always hit "Turn X" to end their turn and send their orders.

geoffthemedio:
@Dilvish-fo Is there any potential issue if AIs are asked to play the same turn twice? I'm wondering if the AI state information will be inconsistent (or even broken?) if reloading a game from after the AI had played its turn, when the AI state info would be from after that turn way played, but then asking the AI to generate orders for that same turn again.

LGM-Doyle:
@geoffthemedio this PR does not change how saves are handled if a human manually saves the game at the end of their turn.

Vezzra:
@LGM-Doyle, with this PR merged, can we close #572 as resolved?

LGM-Doyle:
It resolves it for single player.  

The problem does not make the same sense for multiplayer, since there are multiple players clicking end turn.

PR #1698, opened 2017-08-27
Introduce helper function for partial visibility turn lookup

Morlic-fo:
Cleanup part2, based on #1697

PR #1699, opened 2017-08-27
Cleanup update_system_status() pt.2

Morlic-fo:
Based on #1698

This PR should not change any logic.

--> Squash & Merge

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -451,9 +451,7 @@ class AIstate(object):
             mob_ratings = []  # mobile & unowned
             mobile_fleets = []  # mobile and either owned or unowned
             for fid in local_enemy_fleet_ids:
-                fleet = universe.getFleet(fid)
-                if not fleet:
-                    continue
+                fleet = universe.getFleet(fid)  # ensured to exist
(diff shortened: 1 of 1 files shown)


default/python/AI/AIstate.py
```diff
@@ -399,8 +399,8 @@ def update_system_status(self):
                 # track old/dead enemy fighters for rating assessments in case not enough current info
                 e_f_dict = old_e_fighters if dead_fleet else cur_e_fighters
                 for stats in ship_stats:
-                    attacks = stats[0]
-                    if attacks:
+                    # log only ships that are armed
+                    if stats[0]:
```
Dilvish-fo, line 400: it's not  a failing of this commit/PR, but I wanted to note something that comes to mind when viewing this line (but not make a commit now that would conflict with this PR)-- it seems we have not updated this line yet for fighters and the possibility of a pure carrier ship with no DF weapons.  I think this line should probably read
```
  if stats[0] or stats[4]:
```
that further, it would probably be best to somewhere have an ATTACKS_INDEX and FIGHTER_CAPACITY_INDEX variable to refer to rather than a hardcoded 0 and 4.

Dilvish-fo (approved):

PR #1700, opened 2017-08-27
Remove int() conversion for systemStatus entries

Morlic-fo:
There is no obvious reason to convert/round the calculated values, so I'd prefer to keep the code both more consistent and simple by simply working with floats as we do everywhere else.


Dilvish-fo (approved):
looks fine

PR #1702, opened 2017-08-27
Refactor resetting of AIstate.systemStatus

Morlic-fo:
Based on #1701 

By resetting values that are calculated each turn from scratch at the beginning of update_system_status() we can avoid working with additional variables and get rid of multiple .get() and .setdefault() calls which overall improves readability.

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -430,7 +430,6 @@ class AIstate(object):
             if verbose:
                 print "AIState threat evaluation for %s" % system
             # update fleets
-            local_enemy_fleet_ids = sys_status['localEnemyFleetIDs']
             if system:
                 sys_status['name'] = system.name
                 # TODO: double check are these checks/deletes necessary?
@@ -444,7 +443,7 @@ class AIstate(object):
             enemy_ratings = []  # owned & mobile
             mob_ratings = []  # mobile & unowned
             mobile_fleets = []  # mobile and either owned or unowned
-            for fid in local_enemy_fleet_ids:
+            for fid in sys_status['localEnemyFleetIDs']:
                 fleet = universe.getFleet(fid)  # ensured to exist
                 fleet_rating = CombatRatingsAI.get_fleet_rating(
                     fid, enemy_stats=CombatRatingsAI.get_empire_standard_fighter())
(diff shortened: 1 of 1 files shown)


default/python/AI/AIstate.py
```diff
@@ -386,8 +389,8 @@ def update_system_status(self):
 
             if fleet.ownedBy(empire_id):
                 if not dead_fleet:
-                    my_fleets_by_system.setdefault(this_system_id, []).append(fleet_id)
-                    fleet_spot_position.setdefault(fleet.systemID, []).append(fleet_id)
+                    self.systemStatus[this_system_id]['myfleets'].append(fleet_id)
+                    self.systemStatus[fleet.systemID]['myFleetsAccessible'].append(fleet_id)
```
Dilvish-fo, line 393: This causes an exception if, for example, the fleet is mid-starlane because it tries to access info for systemID = -1, which is not within universe.systemIDs that you initialized the dictionaries from.  Also, it is possible to observe an enemy fleet mid-starlane between systems which one has no knowledge at all about (not even basic vis) and so ```this_system_id``` might be a system ID which likewise is not within universe.systemIDs (which is only for systems known to that player)

Dilvish-fo (commented):
Although I can see the appeal of simplifying the code to avoid the cited .get() and .setdefault() calls, I am skeptical of our ability to reliably avoid them, particularly since I am not confident that the sources of trouble I mention in the code comment are the only sources of trouble that could possibly arise.

Dilvish-fo:
Be sure to read my code comment even though github decided to hide it as 'outdated' -- it's not clear to me why github would do that, and regardless, the comment still applies.

Morlic-fo:
Right, I didn't think about those edge cases.

Any concerns about using a defaultdict instead (possibly converting it to a standard dict at the end of this function to not interfere with the rest of the codebase)
```
>>> def default_systemstatus():
>>>     return {'myfleets': [], 'myFleetsAccessible': [], ... }
>>>
>>> A = defaultdict(default_systemstatus)
>>> A[1]['myfleets'].append(10)

>>> print A
defaultdict(<function default_systemstatus at 0x023CCD70>, {1: {'myfleets': [10], myFleetsAccessible': []}})

>>> print A[2]
{'myfleets': [], 'myFleetsAccessible': []}
```

Dilvish-fo:
> Any concerns about using a defaultdict instead (possibly converting it to a standard dict at the end of this function to not interfere with the rest of the codebase)

Sounds like a good idea.

Morlic-fo:
Not as trivial as hoped, closing this for now.

PR #1703, opened 2017-08-27
Merge fleet related loops in update_system_status()

Morlic-fo:
I am not exactly sure why the loop was over all systems and ```system.fleetIDs``` rather than all ```universe.fleetIDs```.  If it was for performance considerations to reduce number of dict lookups, saving some int-key lookups in dicts seems rather minor.

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -379,10 +379,13 @@ class AIstate(object):
         for fleet_id in universe.fleetIDs:
             fleet = universe.getFleet(fleet_id)
             if not fleet or fleet.empty:
+                self.delete_fleet_info(fleet_id)  # this is safe even if fleet wasn't mine
                 continue
             # TODO: check if currently in system and blockaded before accepting destination as location
             this_system_id = fleet.nextSystemID if fleet.nextSystemID != INVALID_ID else fleet.systemID
             dead_fleet = fleet_id in destroyed_object_ids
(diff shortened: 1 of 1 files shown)


Vezzra:
@Morlic-fo, any reason why you didn't set a milestone for this issue and #1704 (like you did with all the other PRs in that series), or did you just forget?

Morlic-fo:
Just forgot.

Dilvish-fo:
> I am not exactly sure why the loop was over all systems and system.fleetIDs rather than all universe.fleetIDs

I don't recall for sure, but it was probably related to the tremendously greater need to sanitize data back then.  The server's system for managing universe objects was quite different and rather buggy, to put it mildly.

Dilvish-fo (commented):
Looking at the change in the one commit specific to this PR, it does look to me like it should be OK, because System::Copy() will not add fleets to its fleetID list if the player has never had visibility of those fleets.

But since this PR depends on 1702, which I believe has problems, it seems I should not approve this, at least not at this time.

Morlic-fo:
I rebased the PR to no longer depend on #1702

Morlic-fo:
> But since this PR depends on 1702, which I believe has problems, it seems I should not approve this, at least not at this time.

@Dilvish-fo with this issue addressed is this good to merge now or do you have worries about the correctness of the code?

Dilvish-fo (approved):

PR #1705, opened 2017-08-27
Partial progress to new logging system

Morlic-fo:
~~* Renaming all the previously existing ```info``` variables to allow the respective logger usage~~
* Use ```warn``` level logging instead of ```print >> sys.stderr``` for caught exceptions/inconsistencies which let the AI continue with its gameplay,

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -1,5 +1,4 @@
 import copy
-import sys
 from collections import Counter, OrderedDict as odict
 from time import time
 
@@ -182,9 +181,8 @@ class AIstate(object):
             else:
                 try:
                     aggression = state['character'].get_trait(Aggression).key
-                except Exception as e:
-                    print >> sys.stderr, "Could not find the aggression level of the AI, defaulting to typical."
-                    print >> sys.stderr, e
+                except Exception:
+                    warn("Could not find the aggression level of the AI, defaulting to typical.", exc_info=True)
                     aggression = fo.aggression.typical
             self.__init__(aggression)
 
@@ -945,7 +943,7 @@ class AIstate(object):
         for fleet_id in fleets_to_split:
             fleet = universe.getFleet(fleet_id)
             if not fleet:
-                print >> sys.stderr, "Trying to split fleet %d but seemingly does not exist" % fleet_id
+                warn("Trying to split fleet %d but seemingly does not exist" % fleet_id)
                 continue
             fleet_len = len(fleet.shipIDs)
             if fleet_len == 1:
--- a/default/python/AI/PlanetUtilsAI.py
+++ b/default/python/AI/PlanetUtilsAI.py
@@ -1,5 +1,3 @@
-import sys
-
 import freeOrionAIInterface as fo  # pylint: disable=import-error
 import ColonisationAI
 from AIDependencies import INVALID_ID
@@ -43,7 +41,7 @@ def get_capital():
     universe = fo.getUniverse()
     empire = fo.getEmpire()
     if empire is None:
-        print >> sys.stderr, "Danger Danger! FO can't find an empire for me!!!!"
+        warn("Danger Danger! FO can't find an empire for me!!!!")
         return INVALID_ID
     empire_id = empire.empireID
     capital_id = empire.capitalID
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -44,8 +44,6 @@ global variables:
 #       - Filtering the weapon parts must be updated: current cache does not consider tech upgrades, weapons are ignored
 import copy
 import math
-import sys
-import traceback
 from collections import Counter, defaultdict
 
 import freeOrionAIInterface as fo
@@ -189,7 +187,7 @@ class ShipDesignCache(object):
         elif isinstance(pid, list):
             planets = pid
         else:
-            print >> sys.stderr, "Invalid parameter 'pid' for 'print_hulls_for_planets'. Expected int, list or None."
+            warn("Invalid parameter 'pid' for 'print_hulls_for_planets'. Expected int, list or None.")
             return
         print "Hull-cache:"
         get_planet = fo.getUniverse().getPlanet
@@ -208,7 +206,7 @@ class ShipDesignCache(object):
         elif isinstance(pid, list):
             planets = pid
         else:
-            print >> sys.stderr, "Invalid parameter 'pid' for 'print_parts_for_planets'. Expected int, list or None."
+            warn("Invalid parameter 'pid' for 'print_parts_for_planets'. Expected int, list or None.")
             return
         print "Available parts per planet:"
         get_planet = fo.getUniverse().getPlanet
@@ -335,7 +333,7 @@ class ShipDesignCache(object):
         - design_id_by_name
         """
         if self.map_reference_design_name or self.design_id_by_name:
-            print >> sys.stderr, "ShipDesignAI.Cache._build_cache_after_load() called but cache is not empty."
+            warn("ShipDesignAI.Cache._build_cache_after_load() called but cache is not empty.")
         for design_id in fo.getEmpire().allShipDesigns:
             design = fo.getShipDesign(design_id)
             if TESTDESIGN_NAME_BASE in design.name:
@@ -372,8 +370,7 @@ class ShipDesignCache(object):
             try:
                 cached_name = fo.getShipDesign(self.design_id_by_name[designname]).name
                 if cached_name != designname:
-                    print "  WARNING: ShipID cache corrupted."
-                    print "    Expected: %s, got: %s. Repairing cache." % (designname, cached_name)
+                    warn("ShipID cache corrupted. Expected: %s, got: %s." % (designname, cached_name))
                     design_id = next(iter([shipDesignID for shipDesignID in fo.getEmpire().allShipDesigns
                                           if designname == fo.getShipDesign(shipDesignID).name]), None)
                     if design_id is not None:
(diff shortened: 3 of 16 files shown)


Morlic-fo:
Rebased.

Dilvish-fo (approved):
Your proposed significance for warn() sounds fine to me; I think there are some existing uses of error() that should be swapped over to warn() then.

Morlic-fo:
I added another 2 commits to convert current ```error``` level logs to ```warn``` level. I found only three existing use-cases that seem worthy to switch. Most of the ```error``` calls are in ```FreeorionAI.py``` and are done at fairly high-level function calls / exception handling. To make my previous statement more precise, ```warn``` level should probably only apply to fairly local/function level stuff where errors do not affect how the rest of the AI works.

For example,
```
    if not saved_state_string:
        error("AI given empty state-string to resume from; this is expected if the AI is assigned to an empire "
              "previously run by a human, but is otherwise an error. AI will be set to Aggressive.")
```
is technically handled so that the AI can play perfectly well but still seems significant enough to log at ```error``` level.

default/python/AI/AIstate.py
```diff
@@ -182,9 +181,8 @@ def __setstate__(self, state):
             else:
                 try:
                     aggression = state['character'].get_trait(Aggression).key
-                except Exception as e:
-                    print >> sys.stderr, "Could not find the aggression level of the AI, defaulting to typical."
-                    print >> sys.stderr, e
+                except Exception:
+                    warn("Could not find the aggression level of the AI, defaulting to typical.", exc_info=True)
```
LGM-Doyle, line 185: This should be an error.  `character` and `aggression` are required and expected aspects of the AI model.  If `character` is missing from `state`, or `Aggression` is missing from `character` or `Aggression` has no key, these are all errors.  Furthermore, this is an unexpected `Exception`, so there will be traceback information suggesting it is more important than a warning.

default/python/AI/PlanetUtilsAI.py
```diff
@@ -43,7 +41,7 @@ def get_capital():
     universe = fo.getUniverse()
     empire = fo.getEmpire()
     if empire is None:
-        print >> sys.stderr, "Danger Danger! FO can't find an empire for me!!!!"
+        warn("Danger Danger! FO can't find an empire for me!!!!")
```
LGM-Doyle, line 44: Danger Danger, but not an error?

default/python/AI/ShipDesignAI.py
```diff
@@ -189,7 +187,7 @@ def print_hulls_for_planets(self, pid=None):
         elif isinstance(pid, list):
             planets = pid
         else:
-            print >> sys.stderr, "Invalid parameter 'pid' for 'print_hulls_for_planets'. Expected int, list or None."
+            warn("Invalid parameter 'pid' for 'print_hulls_for_planets'. Expected int, list or None.")
```
LGM-Doyle, line 190: This should be an error.  The type of `pid `is wrong, which should never happen.

default/python/AI/ShipDesignAI.py
```diff
@@ -208,7 +206,7 @@ def print_parts_for_planets(self, pid=None):
         elif isinstance(pid, list):
             planets = pid
         else:
-            print >> sys.stderr, "Invalid parameter 'pid' for 'print_parts_for_planets'. Expected int, list or None."
+            warn("Invalid parameter 'pid' for 'print_parts_for_planets'. Expected int, list or None.")
```
LGM-Doyle, line 209: This should be an error.  The type of `pid `is wrong, which should never happen.

default/python/AI/ShipDesignAI.py
```diff
@@ -372,17 +370,15 @@ def _check_cache_for_consistency(self):
             try:
                 cached_name = fo.getShipDesign(self.design_id_by_name[designname]).name
                 if cached_name != designname:
-                    print "  WARNING: ShipID cache corrupted."
-                    print "    Expected: %s, got: %s. Repairing cache." % (designname, cached_name)
+                    warn("ShipID cache corrupted. Expected: %s, got: %s." % (designname, cached_name))
```
LGM-Doyle, line 373: This should be an error.  The cache is broken, which should never happen, even with bad data.

default/python/AI/ShipDesignAI.py
```diff
@@ -372,17 +370,15 @@ def _check_cache_for_consistency(self):
...
                     design_id = next(iter([shipDesignID for shipDesignID in fo.getEmpire().allShipDesigns
                                           if designname == fo.getShipDesign(shipDesignID).name]), None)
                     if design_id is not None:
                         self.design_id_by_name[designname] = design_id
                     else:
                         corrupted.append(designname)
             except AttributeError:
-                print "  WARNING: ShipID cache corrupted. Could not get cached shipdesign. Repairing Cache."
-                print traceback.format_exc()  # do not print to stderr as this is an "expected" exception.
+                warn("ShipID cache corrupted. Could not get cached shipdesign. Repairing Cache.", exc_info=True)
```
LGM-Doyle, line 381: Should this be an error?  Is it possible for the cache to be corrupted and still operating correctly?

default/python/AI/ShipDesignAI.py
```diff
@@ -567,9 +562,8 @@ def _update_buildable_items_this_turn(self, verbose=False):
                 if i != idx:
                     available_hulls[i], available_hulls[idx] = available_hulls[idx], available_hulls[i]
             except ValueError:
-                print >> sys.stderr, ("hull in testhull cache not in available_hulls"
-                                      "even though it is supposed to be a proper subset.")
-                traceback.print_exc()
+                warn("hull in testhull cache not in available_hulls even though it is supposed to be a proper subset.",
```
LGM-Doyle, line 565: This should be an error.

default/python/AI/ShipDesignAI.py
```diff
@@ -1043,9 +1037,8 @@ def add_design(self, verbose=False):
                 return _get_design_by_name(Cache.map_reference_design_name[reference_name]).id
             except AttributeError:
                 cached_name = Cache.map_reference_design_name[reference_name]
-                print >> sys.stderr, "%s maps to %s in Cache.map_reference_design_name." % (reference_name, cached_name),
-                print >> sys.stderr, "But the design seems not to exist..."
-                traceback.print_exc()
+                warn("%s maps to %s in Cache.map_reference_design_name." % (reference_name, cached_name))
+                warn("But the design seems not to exist...", exc_info=True)
```
LGM-Doyle, line 1041: This should be an error, as it is an event that should never happen.

default/python/AI/ShipDesignAI.py
```diff
@@ -1093,7 +1086,7 @@ def optimize_design(self, additional_parts=(), additional_hulls=(),
         elif isinstance(loc, list):
             planets = loc
         else:
-            print >> sys.stderr, "Invalid loc parameter for optimize_design(). Expected int or list but got", loc
+            warn("Invalid loc parameter for optimize_design(). Expected int or list but got %s" % loc)
```
LGM-Doyle, line 1089: The data type is invalid and this should never happen, so it should be an error.

LGM-Doyle (commented):
:smiley: you are using the new logging functions.  

I have quibbled with some of the levels as warnings.  Using error only to indicate unrecoverable errors is too high a threshold for the AI, since if the AI traps an exception and sends empty orders it can be considered as barely operating as intended.

Cases where the event should never happen should be errors and not warnings.  For example a data structure entirely within the python control that is missing a required element, should be an error.  There is no expected/valid reason for the data structure to be missing.  On the other hand a look up table within `fo` missing an id, should only be a warning.  

Another way to consider it is, if the event indicates the algorithm is broken, then it is an error.  If the event indicates a good algorithm has been fed bad data, then it is a warning.

And lastly, an error should indicate something that must be fixed by a developer before release, versus a warning which may be due to bad data/FOCS content.

These are minor concerns, and if you think it better to use `warn()`, this PR is an improvement over the existing code as is.

Dilvish-fo:
@Morlic-fo @Cjkjvfnby Although it would have been much more convenient if LGM-Doyle had gotten his comments in sooner, I think he makes a persuasive argument and this kind of thing is easy enough to change, so let's talk this through a bit more.   I think we should also note a key distinction being that (to my understanding) the error logger will now also always report the error message to the user via chat window.   That seems consistent with his proposed significance for error(), and in any grey area cases it may help decide which type of logger to use.

So, what do you two think about Doyle's proposed distinction between warn and error?

default/python/AI/PlanetUtilsAI.py
Morlic-fo: Not even sure why this part of the code exists, this case is handled as soon as the AI code is called. Even if not, we would have a gazillion of errors thrown if an empire does not exist as it is nowhere else in the code.

Probably best to remove that logging completely...

Morlic-fo:
Seems reasonable.

Cjkjvfnby:
> So, what do you two think about Doyle's proposed distinction between warn and error?

Looks ok.

default/python/AI/ShipDesignAI.py
Morlic-fo: This indicates a mismatch between AI cache and server ids. This is as likely an error of the server as it is of the AI code. In my opinion, this should stay on ```warn``` level.

default/python/AI/ShipDesignAI.py
Morlic-fo: The history of this is that some design ids somehow magically changed during turns once (probably related to a moderator ingame). Within a single turn, the cache worked fine. Next turn, there was a mismatch between cached and server ids. So, yes, the cache may very well work fine (though inefficient) if such a mismatch happens.
If there is intra-turn mismatch between the data, there is no way the AI can play anyway as it would get "random" data from the server and pretty much all the code is unlikely to work at that point.

Morlic-fo:
Posted #1746 for the proposed changes that I agree with.

default/python/AI/ShipDesignAI.py
LGM-Doyle: I'd still make it an error, because it is something that a dev needs to fix.  Although maybe not the dev who wrote the python code.

Cjkjvfnby:
@Dilvish-fo, can you summarize it to http://www.freeorion.org/index.php/Python_Development?

PR #1706, opened 2017-08-27
Fix compiler warnings in gcc clang

LGM-Doyle:


universe/UniverseGenerator.cpp
```diff
@@ -172,7 +174,7 @@ namespace Delauney {
 
         // loop through points generated from systems, excluding the final
         // 3 points added for the covering triangle
-        for (int n = 0; n < points_vec.size() - 3; n++) {
+        for (std::size_t n = 0; n < points_vec.size() - 3 && n < std::numeric_limits<int>::max() ; n++) {
```
geoffthemedio: Just `static_cast<int>(points_vec.size())` to avoid the extra casts later?

Is testing against `std::numeric_limits` really necessary?

universe/UniverseGenerator.cpp
```diff
@@ -249,7 +251,7 @@ namespace Delauney {
 
             // add triangle for last and first points and n
             triangle_list.push_front(
-                {n, (point_idx_list.front()).num, (point_idx_list.back()).num, points_vec});
+                {int(n), (point_idx_list.front()).num, (point_idx_list.back()).num, points_vec});
```
geoffthemedio: prefer `static_cast` to c-style casts.

universe/UniverseGenerator.cpp
LGM-Doyle: You can cast `size_t` to `int`, but not safely.  `int` is signed and `size_t` is unsigned and the spec places no restrictions on their relative sizes.

An alternative is to put the test before the loop and short circuit the loop if  `points_vec.size() > std::numeric_limits<int>::max()`.  However, the method in this PR allows the most possible amount of work to be done before it fails.

Realistically, something else will fail before it hits this limit.

geoffthemedio: I would rather have tests earlier in the function to ensure that the input data is acceptable (eg. not more than 10^5 systems to triangulate) than have the extra tests cluttering the important bits of code. There is no benefit to looping over just some of the points.

LGM-Doyle:
The test is moved to before the loop and the casts have been converted to `static_cast`.

Issue #1618, opened 2017-06-28
Combat not always initiated between planets and aggressive fleets


MatGB:
### Environment

* **FreeOrion Version**:  Compiled from #1587  
* **Operating System**: Linux
  * Compiled from source

### Description
I've thought this was occasionally happening for a bit but hadn't found an example that was clear. Screenshot shows the problem, the previous turn two Yellow ships arrived and died, the next turn this ship arrived, debug data dump says it's on aggressive, but no combat or bombard is reported. That ship should be dead, but isn't. However, my system is still blockaded.

Autosaves for the previous turn and this, to see this happen it should be OK to simply hit end turn, can't guarantee combat won't happen.

### Expected Result
Combat should occur if there's an aggressive ship in system owned by a player other than planet owners.
[combat-failure.zip](https://github.com/freeorion/freeorion/files/1109671/combat-failure.zip)
[no-planet-combat.zip](https://github.com/freeorion/freeorion/files/1109670/no-planet-combat.zip)
![screenshot_20170628_183720](https://user-images.githubusercontent.com/11533206/27652081-54f6c9a0-5c32-11e7-8f6d-1590f9874bd0.png)

LGM-Doyle:
@MatGB I was unable to load the game.  Do you know specifically which commit you were playing from?  There have been changes in PR #1587 and also in `master` to the various parts of the save file format, so I'll need that in order to replicate the problem.  Also if you merged it with other branches, then I will probably need that branch posted.

Or if it is happening regularly, then just wait until it shows up in a game from `master`.

MatGB:
Probably best to wait till I find another instance but if you do want to try a complete mess of a branch, I put it up here:
[Commits · MatGB/freeorion](https://github.com/MatGB/freeorion/commits/add_design_wnd_library)

Several WIPs including yours mixed in, not at all surprising it didn't load in hindsight. Still, quitting that game now so I can start a fresh branch from master now we've got the design stuff merged in.

LGM-Doyle:
I'll wait.

I'll keep it in mind and if I find it while testing other things. I'll link the fix back here.

Does correct bombardment resume on the next turn?  Or is it a transient event that fixes itself on the next turn?

MatGB:
In this case the following turn combat happened as usual and the ships died, my initial reaction was that it wasn't set to aggressive hence the data dump which confirmed it was, combined with the established blockade.

Dilvish-fo:
[...]

PR #1708, opened 2017-08-28
Clear effect generated visibilites after applying.

LGM-Doyle:
This fixes bug #1618, where visibility effects were erroneously applied
repeatedly.

### DIFF
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1662,6 +1662,8 @@ void Universe::ApplyEffectDerivedVisibilities() {
             m_empire_object_visibility[empire_entry.first][object_entry.first] = object_entry.second;
         }
     }
+
+    m_effect_specified_empire_object_visibilities.clear();
 }
 
 void Universe::ForgetKnownObject(int empire_id, int object_id) {


Dilvish-fo:
Although this does fix 1618 for me, I have some concerns about this precise fix, which I think will take me a bit of time to write up carefully, so I wanted to go ahead and get this note in here.  I do expect to get my concerns written up sometime today.

Dilvish-fo:
OK, I'll try to be clear here but it's a bit complicated so bear with me and let me know any areas I need to improve clarity:

In the motivating case here, we have a detection effect which was operating as a limit on vision even though it wasn't really intended to do so.  The solution of this PR is to make visibility/detection effects be nonrestricting boons, specifically by causing them to only be processed once, in the first of the several post-combat calls to UpdateEmpireObjectVisibilities() that occur each turn, so that it can be superseded by other visibility sources in the later calls to UpdateEmpireObjectVisibilities().

What looks to me like the most potentially-problematic aspect of this proposed implementation is that the final one of those UpdateEmpireObjectVisibilities() calls, in which the Effects-granted visibility would no longer apply, is in the post-turn-increment portion of PostCombatProcessTurns, which I think could interfere with (prevent) combat and possibly blockades in any situations where effect-generated visibility could make a difference during the next cycle of PreCombat/Combat/PostCombat.  This could happen in current content if a George colony were sufficiently stealthed to normally avoid detection/combat by an armed enemy in that system, but if that enemy also possessed a different George colony then the Communal Vision effect should give the enemy visibility to attack.

Currently (without this PR), Universe.m_effect_specified_empire_object_visibilities is cleared and then rebuilt once during each PreCombat/Combat/PostCombat cycle, specifically in the PostCombat call to ApplyAllEffectsAndUpdateMeters(), and which then currently lasts until the same point in the next cycle.  This PR adds an extra, much sooner, clearing of m_effect_specified_empire_object_visibilities, which I don't think is the right approach because of the problem noted above.

This PR effectively makes the SetVisibility FOCS Effect to be a boon-only effect (**edit:** on further thought, that may not really be fully accurate in all situations, but I think the rest of the discussion still stands OK), which by itself I think is consistent with all current uses of it and could be a generally OK idea, but then I think the actual solution implementation should be different than that of this PR; I think that the minimal good fix would be as shown in the diff at the bottom of this post, which is sort of a subset of what I am about to describe for my idea of the best fix. 

Really I think we should split this Effect into three different ones--  GrantVisibility being boon-only and would replace all the current FOCS instances of SetVisibility,  SetVisibility being boon-malus-neutral and would establish an absolute overriding visibility level, and a LimitVisibility which is malus-only restriction.   

Each of these three would work with their data structure corresponding to the current m_effect_specified_empire_object_visibilities, which would be cleared and rebuilt in Universe::ApplyAllEffectsAndUpdateMeters just like m_effect_specified_empire_object_visibilities is currently cleared and rebuilt.  We could leave the current m_effect_specified_empire_object_visibilities be fore SetVisibility and then add m_effect_specified_empire_object_visibility_boons for GrantVisibility and m_effect_specified_empire_object_visibility_limits for LimitVisibility.  The only real difference in how they are applied would be that a boon would be applied with a max-type comparison to the otherwise-current visibility (akin to as in PropagateVisibilityToContainerObjects()), a limit would be applied with a min type comparison, and the flat value would be a plain assignment (like SetVisibility currently does).    All three could be handled within ApplyEffectDerivedVisibilities(), but to have the limits and flat values work properly I think we'd need to call ApplyEffectDerivedVisibilities() twice, both before and after the PropagateVisibilityToContainerObjects() call.  We'd also need to decide an order of precedence in case of overlapping/conflicting Effects, I'd propose that we first do limits, then flat sets, then boons (so the latter have priority).

To further illustrate, the following diff shows an alternate fix for #1618, by changing the significance of SetVisibility to be a boon/minimum:
```diff
diff --git a/universe/Universe.cpp b/universe/Universe.cpp
index 85a54b8..843f092 100644
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1659,7 +1659,8 @@ void Universe::ApplyEffectDerivedVisibilities() {
         for (const auto& object_entry : empire_entry.second) {
             if (object_entry.first <= INVALID_OBJECT_ID)
                 continue;
-            m_empire_object_visibility[empire_entry.first][object_entry.first] = object_entry.second;
+            m_empire_object_visibility[empire_entry.first][object_entry.first] = 
+                std::max(object_entry.second, m_empire_object_visibility[empire_entry.first][object_entry.first]);
         }
     }
 }
```

geoffthemedio:
Would it help to move the clear-effect-visibilities call from `ApplyEffectDerivedVisibilities()` into the code that calls it, in carefully selected (but not all) locations where it is called (indirectly through `UpdateEmpireObjectVisibilities()`, so that multiple updates can occur using the same effect results?

Dilvish-fo:
> Would it help to move the clear-effect-visibilities call from ApplyEffectDerivedVisibilities() into the code that calls it, in carefully selected (but not all) locations where it is called (indirectly through UpdateEmpireObjectVisibilities(), so that multiple updates can occur using the same effect results?

I think maybe you are looking at the code *with* this PR-- without this PR there is no clear-effect-visibilities call from ApplyEffectDerivedVisibilities() and multip[le updates are indeed already made using the same efffect results.

Dilvish-fo:
Or if you were just proposing that as a further modification to the present PR, well, I still don't think it would help-- there is only one place (that I can find, at least) during each PreCOmbat/Combat/PostCompat cycle that the effect-visibilities structure is populated so adding extra clearings of it looks like a not-great way to try improving things to me, and I really don't think there is any need to move it, as discussed more above.

Dilvish-fo:
In case my long write-up above was just simply too-long, let me summarize: I am pretty confident that the actual problem with the current code has nothing to do with where the effect-visibilities structure is currently populated or cleared-- the current code is working properly**(edit, see next post)  if you think of SetVisibility as acting as an absolute setting of the visibilty level which overrides most other sources of visibility.  But we use it in some places in the scripts like it is a get-at-least-this-visibility sort of thing.  Which is easy to accomplish with the diff I note above (or to support with a new GrantVisbility effect I propose above).

Dilvish-fo:
well, mostly properly anyways-- if we consider that the current expectations for SetVisibility are really more of a scripting bug than actual code bug (even if we might consider changing the code to conform to the scripting expectations), the actual code bug that appears to be revealed in my pairs of save files in #1618 appears to me to be that the T-1 file **does** result in combat, not that the T-2 file does **not** result in combat.  I think the underlying bug is that m_effect_specified_empire_object_visibilities should be getting serialized  with the universe, but is not.

LGM-Doyle:
The problem is that the time when the state of the world is analyzed to determine what effect to apply is decoupled from the time when the effect is applied.

In this specific case, when the telepathy effect is processed, the decision to grant basic visibility is correct.  Later in the turn processing, when ship arrival or combat has happened, then the applying the result based on the previous decision is incorrect.

The analysis and the application of the result have to happen at the same time, otherwise the bug will just be pushed to a new location.

Dilvish-fo:
> The problem is that the time when the state of the world is analyzed to determine what effect to apply is decoupled from the time when the effect is applied... then the applying the result based on the previous decision is incorrect.... otherwise the bug will just be pushed to a new location.

@LGM-Doyle I disagree.  You seem to be totally glossing over the issue of "what is the right meaning for SetVisibility", which is critical here. (and also the issue with seralization).  If we add serialization for it and then either (i) take it as setting the absolute visibility level (except for the very limited circumstances that it could be overriden by PropagateVisibilityToContainerObjects, PropagateVisibilityToSystemsAlongStarlanesm and SetTravelledStarlaneEndpointsVisible) then it is entirely proper that there be no combat and there would be no combat (although that would be a strange telepathy effect, it would be valid), or (ii) make SetVisibility be boon-only with the diff I showed above, then the combat should, and would, occur.  In either of those approaches, if you think the bug has been "pushed to a new location" then please explain just what bug you think would be remaining/pushed.

Dilvish-fo:
And also let me clarify that I am not trying to push any new work on you for this, @LGM-Doyle, if we just decide whether we want to (i) change the actual meaning of SetVisbility to be boon-only, or (ii) add GrantVisbility and LimitVisibility as I describe above, then I will be happy to put together a corresponding PR and you could review and test it for bugs.

geoffthemedio:
> I think maybe you are looking at the code with this PR--

Yes... I'm suggesting another way to add the change from this PR without requiring that the visibilities determined by effects be immediately cleared upon apply them.

> without this PR there is no clear-effect-visibilities call from ApplyEffectDerivedVisibilities() and multip[le updates are indeed already made using the same efffect results.

Yes, I'm suggesting a way to retain such multiple updating while also clearing the old visibilities to set after they no longer should be retained.

This case is quite like how the Destroy effect works. It doesn't directly destroy something, but rather adds the thing-to-be-destoyed to a list, which is cleared immediately before effects are executed, along with when the universe is reset or cleared. Basically, after each `m_marked_destroyed.clear();` there should be a `m_effect_specified_empire_object_visibilities.clear();`.

https://github.com/freeorion/freeorion/blob/master/universe/Universe.cpp#L1447

I'm not clear on why sometimes there is a call to one of those, but not the other...

https://github.com/freeorion/freeorion/blob/master/universe/Universe.cpp#L461

>Or if you were just proposing that as a further modification to the present PR, well, I still don't think it would help-- there is only one place (that I can find, at least) during each PreCOmbat/Combat/PostCompat cycle that the effect-visibilities structure is populated so adding extra clearings of it looks like a not-great way to try improving things to me, and I really don't think there is any need to move it, as discussed more above.

Relative to this PR, I suggesting less clearing; only clear when appropriate, not every time when setting visibilities from what the effects determined should used.

>  You seem to be totally glossing over the issue of "what is the right meaning for SetVisibility"...

The issue of how SetVisiblity should work is not really relevant to the bug this PR is attempting to resolve, which is that previous turn's effect-derived visibility is being re-applied on later turns.

Dilvish-fo:
> The issue of how SetVisiblity should work is not really relevant to the bug this PR is attempting to resolve, which is that previous turn's effect-derived visibility is being re-applied on later turns.

No, it is not being reapplied on later turns (unless you mean the post-turn-increment portion of PostCombatProcessing, or the first stages of the immediately following cycle, which are still  proper places for it to be reapplied).  That is not why we got the different results from the two save files in #1618.  

The noncombat bug from using the T-2 file is exactly because of the issue of how SetVisibility should work-- the script treated it like a boon-only thing whereas SetVisibility actually is hard-setting the visibility to Basic so for the currently implemented meaning of SetVisibility it was proper that there was no combat (it is most accurate to consider the lack of combat a scripting mistake, not an engine bug).

The reason that the T-1 file gave a different result is because m_effect_specified_empire_object_visibilities does not get serialized--  the savefile corresponds to the state in between Precombat/Compat/PostCombat cycles, which in "normal" play has m_effect_specified_empire_object_visibilities carrying over from the end of one cycle and into the beginning of the next.  But when the file is loaded without m_effect_specified_empire_object_visibilities having been serialized, then the turn T PreCombat call to UpdateEmpireObjectVisibilities() was thinking there were not any Visibilty Effects to handle, when really there were.  The only bug there is the lack of serialization.

geoffthemedio:
The relevant script doesn't assume that the visibility setting effect will only give a boost...

https://github.com/freeorion/freeorion/blob/master/default/scripting/species/common/telepathic.macros#L6

It specifically checks that the target is not visible in the scope.

The potential problem is (as I think @LGM-Doyle described above), if the gamestate changes between when it's decided to set visibility and when it is actually set. This is likely to occur because there are multiple visibility updates, at which point the visibilities are actually set, in part based on what was determined by the last effects execution.

LGM-Doyle:
No worries.  I should have stopped at identifying the problem.  If you want to write the patch, that is great :smile:.

We are more in agreement than not.   Both of your suggestions are improvements.

Adding `GrantVisibility()` and `LimitVisibility()`, option ii) is the better alternative.  `SetVisibility()` should simply mean set visibility, otherwise code in Universe.cpp will resort to directly setting the visibility variables.  I would change the names to `SetMinVisibility()` and `SetMaxVisibility()` which is more functional, although less prosaic.  

Using `SetMinVisibility()` or `SetMaxVisibility()` will move some of the analysis out of `ApplyAllEffectsAndUpdateMeters()` and into `UpdateEmpireObjectVisibilities()`.  This will reduce some of the oddities due to the separation of analysis and effect.

While neither clearing `m_effect_specified_empire_object_visibilities` as proposed in this PR, nor persisting it into the next turn are entirely satisfactory, persisting is better.  Clearing it means that effect generated visibility changes will not be applied pre-combat.  Persisting it means that effect generated visibility changes will not account for empire orders, fleet movement, colonization etc.  That will just result in the regular one turn lag of effects, which players are accustomed to.

A better solution would embed the analysis in `UpdateEmpireObjectVisibilities()`, the same as the other lines in that function.  I don't have a suggestion of how to implement that better solution.

Dilvish-fo:
> It specifically checks that the target is not visible in the scope.  The potential problem is (as I think @LGM-Doyle described above), if the gamestate changes between when it's decided to set visibility and when it is actually set.

I did not read his description that way, but if you consider the issue to be like that, how is this any different than the other effects we have previously discussed with respect to movement?  Like shields still being degraded for combat during the first turn you move out of a molecular cloud.  I feel pretty strongly that my take here is most consistent with those previous discussions.  That said, I guess we could revisit the entire issue of re-evaluating effects after movement and before combat.

LGM-Doyle:
Superceded by #1712.

PR #1712, opened 2017-08-31
Fix telepathic detection

### LINKED ISSUE #1618: Combat not always initiated between planets and aggressive fleets

Dilvish-fo:
This follows on the discussion in #1708 and on our forums, it fixes #1618 

The first commit causes m_empire_object_visibility to be serialized with the universe to prevent
visibility-effect-related errors/discrepancies upon savefile loading as was noted in #1618.  Since this changes serialization and breaks savefile compatibility, the examples from #1618 won't work, but the following two files are made with this and demonstrate the fix, with the desired result being that both files will have the same combat/no-combat status on turn T(17).  With just the first commit from this PR the result is no combat.
[T_minus_1.sav.zip](https://github.com/freeorion/freeorion/files/1267713/T_minus_1.sav.zip)
[T_minus_2.sav.zip](https://github.com/freeorion/freeorion/files/1267714/T_minus_2.sav.zip)

The second commit adjusts the backend processing of effects from SetVisibility so that they act as only a boon, never decreasing visibility.  This is compatible with all current uses in the scripts, and fixes #1618.  

As noted in the discussions, there is another more involved change being contemplated, which would change the parsing of SetVisibility to allow ValueRefs including min/max, and then the datatype for m_empire_object_visibility would be also planned to change.  In that event the second commit here should be reverted, but the material from the first commit would only need a small adjustment to datatype.  So even though this PR might be considered to be an interim fix, I think it is worthwhile.

### DIFF
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1659,7 +1659,11 @@ void Universe::ApplyEffectDerivedVisibilities() {
         for (const auto& object_entry : empire_entry.second) {
             if (object_entry.first <= INVALID_OBJECT_ID)
                 continue;
-            m_empire_object_visibility[empire_entry.first][object_entry.first] = object_entry.second;
+            // TODO: If/when the parser for SetVisibility is expanded to allow ValueRefs like
+            // max(Target.visibility, Basic)
+            // then the folllowing should be reverted to a simple assignment.
+            m_empire_object_visibility[empire_entry.first][object_entry.first] = 
+                std::max(object_entry.second, m_empire_object_visibility[empire_entry.first][object_entry.first]);
         }
     }
 }
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -2977,6 +2977,19 @@ void Universe::GetEmpireObjectVisibilityTurnMap(EmpireObjectVisibilityTurnMap& e
         empire_object_visibility_turns[encoding_empire] = it->second;
 }
 
+void Universe::GetEffectSpecifiedVisibilities(EmpireObjectVisibilityMap& effect_specified_empire_object_visibilities, int encoding_empire) const {
+    if (encoding_empire == ALL_EMPIRES) {
+        effect_specified_empire_object_visibilities = m_effect_specified_empire_object_visibilities;
+        return;
+    }
+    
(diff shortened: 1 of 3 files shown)


geoffthemedio:
looks fine

universe/Universe.cpp
```diff
@@ -1659,7 +1659,11 @@ void Universe::ApplyEffectDerivedVisibilities() {
         for (const auto& object_entry : empire_entry.second) {
             if (object_entry.first <= INVALID_OBJECT_ID)
                 continue;
-            m_empire_object_visibility[empire_entry.first][object_entry.first] = object_entry.second;
+            // TODO: If/when the parser for SetVisibility is expanded to allow ValueRefs like
+            // max(Target.visibility, Basic)
+            // then the folllowing should be reverted to a simple assignment.
+            m_empire_object_visibility[empire_entry.first][object_entry.first] = 
+                std::max(object_entry.second, m_empire_object_visibility[empire_entry.first][object_entry.first]);
```
LGM-Doyle, line 1666: As mentioned in #1708, I think that `SetMinVisibility()` and `SetMaxVisibility()` are clearer and more functional, (although not as functional as something evaluated like a ValueReef).  With this change setting visibility from an effect can only act as a set minimum visibility.

I don't object to merging this PR, which is a step forward and fixes the bug.

Issue #1707, opened 2017-08-28
Pedia links do not highlight on rollover


dbenage-cx:
### Environment
* **FreeOrion Version**: 0.4.7+
* **Operating System**: Fedora 26
* **Fetched as** Compiled from source

### Description
Links within the pedia window fail to highlight/change color when mouse hovers over them.

Text appears to get decorated twice, giving overlapping rgba tags (in this case, with different values).
Initially [reported](https://github.com/freeorion/freeorion/pull/1580#issuecomment-325248642) by @MatGB, which notes an issue with owned colors.

Presumably related to PR #1689 (not verified)

dbenage-cx:
@LGM-Doyle TY for researching and providing a fix.

Resolved by #1713

PR #1713, opened 2017-08-31
Fix bug #1707 pedia links not highlighting on rollover.

LGM-Doyle:
CUILinkTextMultiEdit uses the underlying TextControl to store the text
processed to include the tags and overrides virtual SetText() to assign
the processed text.  The problem is that the underlying TextControl
calls SetText() internally when adjusting the scroll bars in its
CompleteConstruction().

This commit brackets CUIMultiEdit::CompleteConstruction() in
CUILinkTextMultiEdit::CompleteConstruction() in a
m_already_setting_text_so_dont_link guard to prevent the double
wrapping.


geoffthemedio:
if it fixes it for you, go ahead...

PR #1714, opened 2017-08-31
[AI] Cleanup

Cjkjvfnby:
Fix long lines
Rename variable
Except more specific exception

### DIFF
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -13,23 +13,20 @@ from common.configure_logging import convenience_function_references_for_logger
 __designStats = {}
 
 
-def stats_meet_reqs(stats, reqs):
+def stats_meet_reqs(stats, requirements):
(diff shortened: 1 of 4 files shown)


default/python/AI/FleetUtilsAI.py
```diff
@@ -14,22 +14,22 @@
...
     try:
-        for key in reqs:
-            if stats.get(key, 0) < reqs[key]:
+        for key in requirements:
+            if stats.get(key, 0) < requirements[key]:
                 return False
         return True
-    except:
+    except KeyError:
         return False
```
Morlic-fo: That try-except seems redundant given we only look up keys that we get from the keys() method in one dict and use ```get``` to look those up in the other dict.

Morlic-fo:
Needs rebase. The except-clause seems redundant but looks fine otherwise.

Dilvish-fo (approved):
looks fine to me

Issue #1690, opened 2017-08-23
Traceback information missing from AI log files


Dilvish-fo:
### Environment
Ubuntu Linux 16.04
apparently not necessarily happening in (Fedora?) but I am asking for confirmation of that in light of the more specific information below

### Description
AI error logging explicitly triggered by calls to the error() logger , at least as described in #1688 and even after the partial fix there, will result in traceback information being omitted from the log file even though it is properly sent to the human player in chat.

This is not a *huge* problem in that the AI is not broken by it, and the desired log info does get sent to the message window, but in order to be able to handle issues encountered by playtesters who submit logfiles, the traceback info really should appear in the logfile as well.

### Steps to reproduce
 - reintroduce python interface naming errors by reverting 5681081a
 - test with and without the fix from #1688, recompiling
 - start a game with 1 AI (best only 1 to avoid chat window overflow)
 - check chat window and logs

(Quote from #1688)
> It is not necessary to add the exception e as a separate parameter. The python logging module will fetch the exception information on its own with sys.exc_info(), if exc_info=True is passed as a parameter. 

@LGM-Doyle I think that is not fully accurate; there is still a remaining problem in that the traceback is not properly appearing in the log files in these cases.   Please review the 3 snippets below and doublecheck in just which of these cases your results differ.    I am running Python 2.7.12, how about yourself?  I see that there was a 2.7.13 released a couple months ago, but its release notes do not mention logging.

Here is a representative logfile snippet resulting from the python interface name problem *without* the partial fix of 1688 (note, it looks to me like the traceback here is actually from a second exception encountered while trying to handle the logging for the originally-caught exception that was being logged).   

```
00:53:29.176368 [debug] python : ColonisationAI.py:476 - Outpost Fleets Without Missions: 0
00:53:29.176643 [debug] python : ColonisationAI.py:477 - 
00:53:29.176759 [error] root : FreeOrionAI.py:340 - Exception while trying to calculate_priorities
00:53:29.177636 [error] python : traceback.py:13 - Traceback (most recent call last):
00:53:29.178267 [error] python : traceback.py:13 -   File "/home/FO/programs/fo_recov/default/python/AI/freeorion_tools/_freeorion_tools.py", line 111, in emit
00:53:29.178871 [error] python : traceback.py:13 -     msg = self.format(record)
00:53:29.179472 [error] python : traceback.py:13 -   File "/usr/lib/python2.7/logging/__init__.py", line 734, in format
[...]

PR #1716, opened 2017-09-01
Improve performance of python logger

### LINKED ISSUE #1715: New python logging has horrible performance
### Environment
* **FreeOrion Version**:  Test build 2017-08-24.4e12b03
* **Operating System**: Win7, 64-bit
* **Python Version**: FreeOrion-distributed

### Description
FreeOrion has become extremely slow for me. Starting a new game takes ages and the AI takes a long time to finish their turns. As it turns out, this is caused by the new python logging framework introduced by @LGM-Doyle in https://github.com/freeorion/freeorion/pull/669.

I wrote a simple test scenario for profiling: 1000 print statements and another 1000 arbitrary list extension to make sure it is not the python interpreter itself hanging.

### Result
 It takes a whole 7 seconds on my system to finish 
the 1000 print statements. Only 274ms are spent doing the actual write() call. The rest of time is spent inspecting the stack: [profiling_log.txt](https://github.com/freeorion/freeorion/files/1271262/profiling_log.txt)

### Expected Result
The overhead of the logger needs to be significantly reduced to ensure acceptable performance.

### Steps to reproduce

1. Pull https://github.com/Morlic-fo/freeorion/commit/9a98c6b6a956e7dcf24eeafa3c7e7aa2f0d48b1a
2. Start a game of FreeOrion with a single AI
3. Wait for the game to crash
4. Check end of logfile for profiling results

Morlic-fo:
Aims to fix #1715 

Problem: The ```inspect.stack()``` function walks through the stack which takes a long time.

Solution: We only require information about a single frame which we can find using the ```sys._getframe(1)``` call.

Note that ```sys._getframe``` does not exist for all python versions. Official documentation states
```
This function should be used for internal and specialized purposes only. It is not guaranteed to exist in all implementations of Python.
```
At least my version of ```inspect.stack()``` relies on that function call anyway so I expect it to matter only on systems which did not support the current implementation either. However, feedback from different OS / python versions would be useful.

Performance for the test case described in #1715 improved from 7 seconds to 350ms for me which makes FreeOrion playable again for me.

### DIFF
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -82,7 +82,6 @@ log.
 import sys
 import string
 import logging
-import inspect
 import os
 
 try:
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -128,14 +128,20 @@ class _stdXLikeStream(object):
 
     def write(self, msg):
         # Grab the caller's call frame info
-        try:
+        if hasattr(sys, '_getframe'):
             frame = sys._getframe(1)
-            line_number = frame.f_lineno
-            function_name = frame.f_code.co_name
-            filename = frame.f_code.co_filename
-            if string.lower(filename[-4:]) in ('.pyc', '.pyo'):
-                filename = filename[:-4] + '.py'
-        except:
+            try:
+                line_number = frame.f_lineno
+                function_name = frame.f_code.co_name
+                filename = frame.f_code.co_filename
+                if string.lower(filename[-4:]) in ('.pyc', '.pyo'):
+                    filename = filename[:-4] + '.py'
+            except:
+                (filename, line_number, function_name) = ("", "", "")
+            finally:
+                # Explicitly del references to the caller's frame to avoid persistent reference cycles
+                del frame
+        else:
             (filename, line_number, function_name) = ("", "", "")
 
         try:
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -80,6 +80,7 @@ log.
 
 """
 import sys
+import string
 import logging
 import inspect
 import os
@@ -129,7 +130,12 @@ class _stdXLikeStream(object):
     def write(self, msg):
         # Grab the caller's call frame info
         try:
-            (filename, line_number, function_name, _, _) = inspect.getframeinfo(sys._getframe(1), context=0)
+            frame = sys._getframe(1)
+            line_number = frame.f_lineno
+            function_name = frame.f_code.co_name
+            filename = frame.f_code.co_filename
+            if string.lower(filename[-4:]) in ('.pyc', '.pyo'):
+                filename = filename[:-4] + '.py'
         except:
             (filename, line_number, function_name) = ("", "", "")
 
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -128,11 +128,9 @@ class _stdXLikeStream(object):
 
     def write(self, msg):
         # Grab the caller's call frame info
-        stack = inspect.stack()
-        if len(stack) > 1:
-            (_, filename, line_number, function_name, _, _) = inspect.stack()[1]
-        else:
-            # No available call frame info
+        try:
+            (filename, line_number, function_name, _, _) = inspect.getframeinfo(sys._getframe(1), context=0)
+        except:
             (filename, line_number, function_name) = ("", "", "")
 
         try:
@@ -140,7 +138,6 @@ class _stdXLikeStream(object):
 
         finally:
             # Explicitly del references to the caller's frame to avoid persistent reference cycles
-            del stack
             del filename
             del line_number
             del function_name
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -80,7 +80,6 @@ log.
 
 """
 import sys
-import string
 import logging
 import os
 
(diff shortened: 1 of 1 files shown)


Morlic-fo:
I added another commit which directly parses the obtained stack frame. This avoids several seemingly not very helpful checks in the ```inspect``` module.
With that second commit, all overhead regarding the filename lookup is removed and the performance is pretty much that of a "native" ```print``` statement (further speedup by x3 for me. The test case now takes 110ms from me rather than the initial 7 seconds).

memem359:
The logger for the Mac OSX build hasn't worked since last winter. So until that PR fix goes through, you'll want someone using Linux for your cross-check.

Dilvish-fo:
> The logger for the Mac OSX build hasn't worked since last winter. So until that PR fix goes through...

@memem359 um...?  I tried searching for both PRs and Issues with the word log or logger, and couldn't find a match, just what are you referring to?

dbenage-cx:
@Dilvish-fo https://github.com/freeorion/freeorion/issues/1575#issuecomment-313884360

Dilvish-fo:
@dbenage-cx  ah, thanks much.  I am a bit puzzled now at least about how I missed that issue, surely someone in my household must have distracted me.  I've commented more there.  

@memem359 if I was understanding you correctly in that issue and the PRs linked from it, it sounded like you said those PRs fixed the logging issue for you, and so it seems that you should still be able to do a cross check here...?

Dilvish-fo:
This PR does speed up my results by more than 30-fold.
without PR:
```
22:23:08.266484 [debug] python : FreeOrionAI.py:19 -          1910090 function calls in 0.940 seconds
22:23:08.266498 [debug] python : FreeOrionAI.py:19 - 
22:23:08.266503 [debug] python : FreeOrionAI.py:19 -    Ordered by: cumulative time
22:23:08.266507 [debug] python : FreeOrionAI.py:19 - 
22:23:08.266511 [debug] python : FreeOrionAI.py:19 -    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
22:23:08.266515 [debug] python : FreeOrionAI.py:19 -         1    0.007    0.007    0.940    0.940 /home/FO/programs/fo_recov/default/python/AI/FreeOrionAI.py:24(test_logger)
22:23:08.266520 [debug] python : FreeOrionAI.py:19 -      1998    0.041    0.000    0.932    0.000 /home/FO/programs/freeorion_builds/testing3/default/python/common/configure_logging.py:129(write)
22:23:08.266523 [debug] python : FreeOrionAI.py:19 -      3996    0.004    0.000    0.885    0.000 /usr/lib/python2.7/inspect.py:1058(stack)
```

with the PR:
```
22:57:26.909562 [debug] python : FreeOrionAI.py:19 -          17984 function calls in 0.027 seconds
22:57:26.909577 [debug] python : FreeOrionAI.py:19 - 
22:57:26.909588 [debug] python : FreeOrionAI.py:19 -    Ordered by: cumulative time
22:57:26.909599 [debug] python : FreeOrionAI.py:19 - 
22:57:26.909610 [debug] python : FreeOrionAI.py:19 -    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
22:57:26.909621 [debug] python : FreeOrionAI.py:19 -         1    0.003    0.003    0.027    0.027 /home/FO/programs/fo_recov/default/python/AI/FreeOrionAI.py:24(test_logger)
22:57:26.909632 [debug] python : FreeOrionAI.py:19 -      1998    0.019    0.000    0.024    0.000 /home/FO/programs/freeorion_builds/testing3/default/python/common/configure_logging.py:129(write)
22:57:26.909642 [debug] python : FreeOrionAI.py:19 -      1998    0.002    0.000    0.003    0.000 /usr/lib/python2.7/posixpath.py:82(split)
```
looks good to me

default/python/common/configure_logging.py
```diff
@@ -128,19 +128,27 @@ def __init__(self, level):
...
-        if len(stack) > 1:
-            (_, filename, line_number, function_name, _, _) = inspect.stack()[1]
+        if hasattr(sys, '_getframe'):
+            frame = sys._getframe(1)
+            try:
+                line_number = frame.f_lineno
+                function_name = frame.f_code.co_name
+                filename = frame.f_code.co_filename
+                if string.lower(filename[-4:]) in ('.pyc', '.pyo'):
+                    filename = filename[:-4] + '.py'
```
LGM-Doyle: I would remove these two lines changing the suffix of the compiled/optimized byte code to indicate that it is running from the text file.  It is not true.  It makes it more difficult for people to determine if they are running from stale compiled (.pyc) files if the log incorrectly indicates that they are running the current python (.py) files.

LGM-Doyle (commented):
It looks good to me, 👍, except as noted about the .pyc filenames. 

I see a performance improvement with this PR for the test case from 750 ms to 60 ms on Fedora.

Oddly, this PR makes bug [#1690 Traceback information missing from AI log files](https://github.com/freeorion/freeorion/issues/1690)  start occurring on my machine.  This is good. I will be able to fix it, now that I can replicate it.

Thanks for reporting and fixing this.  I was not aware of the slow down on the Windows machines.

Morlic-fo:
I did remove the *.pyc to *.py conversion. It was replicating the behavior from ```inspect.stack()``` and seemed cleaner but if you think the information might help you debugging some issues, then there is merit in keeping it.

LGM-Doyle:
Thanks.  I did not known that `inspect.stack()` hid the compiled file type suffixes.

PR #1709, opened 2017-08-29
Fuel parts auto upgrade

agrrr3:
This is an implementation of auto-upgrading fuel tanks when researching deuterium or antimatter tank technology. 

The implementation adds as fuel as one effect per tech. It reads like:

> Medium Basic Ship Hull:  3
> Ship Part  Basic Tank:     1
>Deuterium Tank Tech:       1
>Antimatter Tank Tech:      3

> Medium Basic Ship Hull..3
> Ship Part  Basic Tank...1
> Deuterium Tank Tech.....1
> Antimatter Tank Tech....3

Also note that there is no check for supply connection. Fuel max gets upgraded instantly when tech is there. But to get fuel, you actually need a refuel source (which in 99% means supply-line connection)

Discussion is in [freeorion forum thread](http://www.freeorion.org/forum/viewtopic.php?f=15&t=10681)

PR #1718, opened 2017-09-02
Reorganize FOCS files

dbenage-cx:
Reorganizes FOCS files in a consistent manner using sub directories and limiting to one definition per file.

Superceeds #1631 

Open PRs with conflicts: #1709 #1343


o01eg:
Could you also move FOCS read only by the server or the AIs move to separate directory like `default/server/`? As example there are scripts for empire statistic.

dbenage-cx:
@o01eg I fail to see what would establish statistics to the domain of server?  I also do not see any definitions (or other files within `default/scripting`) that are specific to the AI.

o01eg:
> I fail to see what would establish statistics to the domain of server?

As I've just rechecked only server calls code `Universe::UpdateStatRecords` to parse statistics scripts.

>  I also do not see any definitions (or other files within default/scripting) that are specific to the AI.

Ok.

geoffthemedio:
I don't think categorizing stuff as "server only" is useful conceptually. AI stuff is doing something conceptually different. Statistics, even if only actually parsed by the server, are just another part of the game content, and they do appear in the client.

LGM-Doyle:
Thanks for the notice about the conflict with #1343.

o01eg:
@geoffthemedio It's about changes could be made without breaking compatibility with the client.

geoffthemedio:
I think the client would require matching stringtable updates for changes to statistics, so that could be potentially misleading...

MatGB:
Unlike last time this doesn't conflict with any of my current work so that's a relief.

I would rather keep the stats where they are, they're written using FOCS and I sometimes use them for testing purposes, plus at some point we want to gate them in some way based on intelligence/spying/diplomacy which'll be part of scripting.

But it's a soft preference, if there's a strong reason to change I don't object, I'm just not sure I understand why you'd want to.

Regarding the PR overall, it looks fine to me, haven't got time to test it currently but the principle is sound.

dbenage-cx:
@o01eg Thank you for the info on statistics, I had not looked into those in the past.

PR #1719, opened 2017-09-02
Fix attempting to start 2nd server causing the human client to hang.

LGM-Doyle:
**Problem**
A server is already running and the human client attempts to start a new game or load a game.  The screen goes black and provides no additional feedback to the player, nor any way to recover.

**This Solution**
This PR informs the player with a popup, "Can't start server.  A local server is already running." if the server can't start.  The player can then either quit this client or resign/quit from the other client/server and start a game from this client.

This PR attempts to connect to a server for 100ms before starting a local server.  If it successfully connects it will not start another server.  Otherwise it starts the server as normal.  This results in a 100ms delay in starting up a local server.

There are changes to the server code to prevent the 2nd client's disconnection (which is an unestablished player connection on the server) from causing the game to shutdown.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -411,7 +411,20 @@ namespace {
 #include <stdlib.h>
 #endif
 
+namespace {
+    class LocalServerAlreadyRunningException : public std::runtime_error {
+    public:
+        explicit LocalServerAlreadyRunningException(const std::string& s) : std::runtime_error(s) {}
+    };
+}
+
 void HumanClientApp::StartServer() {
+    auto connected = m_networking->ConnectToLocalHostServer(std::chrono::milliseconds(100));
+    if (connected) {
+        ErrorLogger() << "Can't start local server because a server is already connecting at 127.0.0.0.";
+        throw LocalServerAlreadyRunningException("LOCAL_SERVER_ALREADY_RUNNING_ERROR");
+    }
+
     std::string SERVER_CLIENT_EXE = ServerClientExe();
     DebugLogger() << "HumanClientApp::StartServer: " << SERVER_CLIENT_EXE;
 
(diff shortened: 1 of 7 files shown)


network/ServerNetworking.cpp
```diff
@@ -112,6 +112,10 @@ bool PlayerConnection::SendMessage(const Message& message) {
     return SyncWriteMessage(message);
 }
 
+bool PlayerConnection::IsEstablished() const {
+    return (m_ID != INVALID_PLAYER_ID || !m_player_name.empty() || m_client_type != Networking::INVALID_CLIENT_TYPE);
+}
+
```
o01eg: Should be client type tested here? I'm going to set it before establishing between authentication request and response.

client/human/HumanClientApp.cpp
```diff
@@ -411,7 +411,20 @@ namespace {
 #include <stdlib.h>
 #endif
 
+namespace {
+    class LocalServerAlreadyRunningException : public std::runtime_error {
+    public:
+        explicit LocalServerAlreadyRunningException(const std::string& s) : std::runtime_error(s) {}
```
geoffthemedio: Is this ever created with another string parameter? If not, can remove the parameter and set it in the constructor, rather than needing to pass it each time...

client/human/HumanClientApp.h
```diff
@@ -53,7 +53,9 @@ class HumanClientApp :
 
     void                SetSinglePlayerGame(bool sp = true);
 
-    void                StartServer();                  ///< starts a server process on localhost
+    /** Starts a server process on localhost.  Throws a runtime_error if
```
geoffthemedio: shouldn't it be `throws a LocalServerAlreadyRunningException if` ...?

network/ServerNetworking.cpp
```diff
@@ -112,6 +112,10 @@ bool PlayerConnection::SendMessage(const Message& message) {
     return SyncWriteMessage(message);
 }
 
+bool PlayerConnection::IsEstablished() const {
+    return (m_ID != INVALID_PLAYER_ID || !m_player_name.empty() || m_client_type != Networking::INVALID_CLIENT_TYPE);
```
geoffthemedio: based on the descriptive comment, these should be `&&` instead of `||`

geoffthemedio:
haven't tested myself, but if it works, OK

LGM-Doyle:
Testing is straightforward.

Quickstart a game and then quickstart another game.  The second instance should not make it past the intro screen, before it pops up the error message.

PR #1720, opened 2017-09-03
Fix bug #1690 python traceback info missing from log.

LGM-Doyle:
This might fix bug #1690.

I was initially unable to the behavior described in #1690.  However, while testing #1716, I saw similar behavior, with traceback information of uncaught python exceptions not appearing in either the log or the console window.

The problem is that uncaught python exceptions are handled by `sys.excepthook` and it was not overridden.  

This PR overrides `sys.excepthook` with  `_unhandled_exception_hook()` which sends the
traceback information to the log.

This will need testing from someone who can replicate the bug.

### DIFF
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -83,6 +83,7 @@ import sys
 import logging
 import inspect
 import os
+import traceback
 
 try:
     import freeorion_logger  # FreeOrion logger interface pylint: disable=import-error
@@ -183,6 +184,12 @@ def _create_narrow_handler(level):
     return h
 
 
+def _unhandled_exception_hook(*exc_info):
+    traceback_msg = "Uncaught exception: {0}".format(
+        "".join(traceback.format_exception(*exc_info)))
+    logging.getLogger().error(traceback_msg)
+
+
 def redirect_logging_to_freeorion_logger(initial_log_level=logging.DEBUG):
     """Redirect stdout, stderr and the logging.logger to hosting process' freeorion_logger."""
 
@@ -198,6 +205,9 @@ def redirect_logging_to_freeorion_logger(initial_log_level=logging.DEBUG):
         logger.addHandler(_create_narrow_handler(logging.ERROR))
         logger.addHandler(_create_narrow_handler(logging.FATAL))
         logger.addHandler(_create_narrow_handler(logging.NOTSET))
+
+        # Replace the system unhandled exception handler
+        sys.excepthook = _unhandled_exception_hook
 
         logger.setLevel(initial_log_level)
         logger.info("The python logger is initialized with a log level of %s" %
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -165,6 +165,8 @@ class _LoggerHandler(logging.Handler):
         msg = str(record.message) + "\n"
         if record.exc_info:
             if not isinstance(record.exc_info, tuple):
+                # record.exc_info is not a local variable and will be garbage collected when record
+                # is garbage collected by the logging library
                 record.exc_info = sys.exc_info()
             traceback_msg = "".join(traceback.format_exception(*record.exc_info))
             msg += traceback_msg
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -161,7 +161,14 @@ class _LoggerHandler(logging.Handler):
         }[level]
 
     def emit(self, record):
-        self.logger(str(record.msg) + "\n", str(record.name), str(record.filename),
+        record.message = record.getMessage()
+        msg = str(record.message) + "\n"
+        if record.exc_info:
+            if not isinstance(record.exc_info, tuple):
+                record.exc_info = sys.exc_info()
+            traceback_msg = "".join(traceback.format_exception(*record.exc_info))
+            msg += traceback_msg
+        self.logger(msg, str(record.name), str(record.filename),
                     str(record.funcName), str(record.lineno))
 
 
--- a/default/python/AI/DiplomaticCorp.py
+++ b/default/python/AI/DiplomaticCorp.py
@@ -5,7 +5,7 @@ import freeOrionAIInterface as fo
 import FreeOrionAI as foAI
 from character.character_module import Aggression
 from character.character_strings_module import possible_greetings
-from freeorion_tools import UserStringList, chat_on_error
+from freeorion_tools import UserStringList
 
 from common.configure_logging import convenience_function_references_for_logger
 (debug, info, warn, error, fatal) = convenience_function_references_for_logger(__name__)
@@ -25,7 +25,6 @@ class DiplomaticCorp(object):
     def __init__(self):
         self.be_chatty = True
 
-    @chat_on_error
     def handle_diplomatic_message(self, message):
         """Handle a diplomatic message update from the server,
         such as if another player declares war, accepts peace, or cancels a proposed peace treaty.
@@ -82,7 +81,6 @@ class DiplomaticCorp(object):
             greets = UserStringList("AI_FIRST_TURN_GREETING_BEGINNER")
         return random.choice(greets)
 
-    @chat_on_error
     def handle_diplomatic_status_update(self, status_update):
         """Handle an update about the diplomatic status between players, which may
         or may not include this player."""
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -33,7 +33,7 @@ import ResearchAI
 import ResourcesAI
 import TechsListsAI
 from AIDependencies import INVALID_ID
-from freeorion_tools import chat_on_error, handle_debug_chat, AITimer, init_handlers
+from freeorion_tools import handle_debug_chat, AITimer, init_handlers
 from common.listeners import listener
 from character.character_module import Aggression
 from character.character_strings_module import get_trait_name_aggression, possible_capitals
@@ -66,7 +66,6 @@ foAIstate = AIStateMock()
 diplomatic_corp = None
 
 
-@chat_on_error
 def startNewGame(aggression_input=fo.aggression.aggressive):  # pylint: disable=invalid-name
     """Called by client when a new game is started (but not when a game is loaded).
     Should clear any pre-existing state and set up whatever is needed for AI to generate orders."""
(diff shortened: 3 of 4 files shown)


Dilvish-fo:
> The problem is that uncaught python exceptions are handled by sys.excepthook and it was not overridden.

I added a comment to the cited issue to emphasize/clarify that my problems described in the issue were stemming from explicit calls to the error logger (i.e. generally from explicitly caught exceptions), and that uncaught exceptions have been producing tracebacks just fine.   Also, when looking at this more, it seems to me we need to distinguish exceptions that are caught in a regular try:except structure that then makes a call to the error logger, versus exceptions that are caught via the @chat_on_error decorator having been applied to the function they appear in.

So, unfortunately, this PR seems to make no difference to me for the problems I report in #1690.  Either with or without this PR, attempting to log a traceback via a call to error(), like from around line 340 of FreeOrionAI.py, just gives me a single error line without a traceback.  Specifically, either or without this PR, but with the following changes that trigger a caught exception
```diff
diff --git a/default/python/AI/ColonisationAI.py b/default/python/AI/ColonisationAI.py
index 954dc28..6cead09 100644
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -253,6 +253,7 @@ def check_supply():
 
 
 def survey_universe():
+    I_am_not_defined
     check_supply()
     colonization_timer.start("Categorizing Visible Planets")
     universe = fo.getUniverse()
```
I simply get the following in the log (but should also get a traceback)
```
19:37:49.033198 [debug] root : FreeOrionAI.py:318 - Calling AI Modules
19:37:49.033273 [error] root : FreeOrionAI.py:340 - Exception global name 'I_am_not_defined' is not defined while trying to survey_universe
19:37:49.034455 [debug] python : ShipDesignAI.py:150 - 
```

Dilvish-fo:
Regarding the uncaught exceptions that this PR was more specifically directed to, it makes only a very small bit of difference in the logging I get.  Without this PR, but with the following change to cause an exception:
```diff
diff --git a/default/python/AI/FreeOrionAI.py b/default/python/AI/FreeOrionAI.py
index e28f596..f362175 100644
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -225,13 +225,14 @@ def handleDiplomaticStatusUpdate(status_update):  # pylint: disable=invalid-name
     diplomatic_corp.handle_diplomatic_status_update(status_update)
 
 
-@chat_on_error
-@listener
+#@chat_on_error
+#@listener
 def generateOrders():  # pylint: disable=invalid-name
     """Called once per turn to tell the Python AI to generate and issue orders to control its empire.
     at end of this function, fo.doneTurn() should be called to indicate to the client that orders are finished
     and can be sent to the server for processing."""
 
+    I_am_not_defined
     rules = fo.getGameRules()
     print "Defined game rules:"
     for rule in rules.getRulesAsStrings:
```
then I get the following in my log:
```
20:17:16.974774 [debug] ai : AIFramework.cpp:136 : PythonAI::GenerateOrders : initializing turn
20:17:16.974853 [error] python : : - Traceback (most recent call last):
20:17:16.974921 [error] python : : -   File "/home/FO/programs/fo_recov/default/python/AI/FreeOrionAI.py", line 235, in generateOrders
20:17:16.975036 [error] python : : -     I_am_not_defined
20:17:16.975178 [error] python : : - NameError: global name 'I_am_not_defined' is not defined
```

If I run with this PR and the above change, then the logging is
```
20:20:01.200286 [debug] ai : AIFramework.cpp:136 : PythonAI::GenerateOrders : initializing turn
20:20:01.200414 [error] root : configure_logging.py:190 - Uncaught exception: Traceback (most recent call last):
20:20:01.200426 [error] root : configure_logging.py:190 -   File "/home/FO/programs/fo_recov/default/python/AI/FreeOrionAI.py", line 235, in generateOrders
20:20:01.200441 [error] root : configure_logging.py:190 -     I_am_not_defined
20:20:01.200445 [error] root : configure_logging.py:190 - NameError: global name 'I_am_not_defined' is not defined
20:20:01.200459 [error] root : configure_logging.py:190 - 
```
The only difference I notice is that with this PR it is adding references to specific lines within the logging code itself.  

Anyways, while this PR does not seem to help me, it also does not seem to hurt me, so if it helps anyone else I am fine with it.

LGM-Doyle:
Thanks for clarifying that the bug was with `error("blah", exc_info=True)` and that the uncaught exceptions  in unwrapped functions had correct traceback information.  

I added better formatting, including traceback information to the `_LoggingHandler()`.  I removed `chat_on_error`, which is now duplicating the behavior of `sys.excepthook`.

default/python/common/configure_logging.py
```diff
@@ -161,7 +161,14 @@ def __init__(self, level):
         }[level]
 
     def emit(self, record):
-        self.logger(str(record.msg) + "\n", str(record.name), str(record.filename),
+        record.message = record.getMessage()
+        msg = str(record.message) + "\n"
+        if record.exc_info:
+            if not isinstance(record.exc_info, tuple):
+                record.exc_info = sys.exc_info()
```
Morlic-fo: Is this safe?
On first sight, this looks like a potential memory leak. The ```sys.exc_info()``` call returns a reference to a stack frame as third tuple element and thus introduces circular references that generally should be resolved by explicit deletion.

See https://docs.python.org/2/library/sys.html#sys.exc_info

If you are confident an explicit deletion is not required here, please add some comment here why this is safe.

Dilvish-fo: > On first sight, this looks like a potential memory leak. 

from the docs you cite, the issue sounds more like a garbage-collection efficiency issue for us rather than an actual memory leak.  But still worth avoiding if possible, certainly.

> Note Beginning with Python 2.2, such cycles are automatically reclaimed when garbage collection is enabled and they become unreachable, but it remains more efficient to avoid creating cycles.

Morlic-fo: I don't want to argue semantics here. It may technically not be a memory leak but the effect I think is very much the same.

Reference cycles are supposed to not be completely freed if objects defining a ```__del__``` method are involved (seemingly this was resolved in Python 3.4). Instead, it is put into the garbage which must be manually cleared: https://docs.python.org/2/library/gc.html#gc.garbage
To my knowledge we never manually clear the garbage so the memory would never be freed.

It may of course be that there are additionally measures taken in case of tracebacks and it is actually safe to rely on the garbage collector.

Dilvish-fo (commented):
> I added better formatting, including traceback information to the _LoggingHandler(). I removed chat_on_error, which is now duplicating the behavior of sys.excepthook.

@LGM-Doyle  Wunderbar!  This has now fixed #1690 for me.  Besides the issue Morlic raised about sys.exc_info (which looks to me like it can be readily dealt with and should be) this all looks good to me.

Dilvish-fo:
@Morlic-fo the following changes work fine for me and seem to me they would avoid the circular reference, since the unpacked traceback from sys.exec_info() doesn't get into a local variable until the traceback.format_exception() level, which is not itself referenced in the traceback, or am I misunderstanding how garbage collection considers that unpacking?
```diff
diff --git a/default/python/common/configure_logging.py b/default/python/common/configure_logging.py
index 651c3d3..c7fde33 100644
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -164,10 +164,10 @@ class _LoggerHandler(logging.Handler):
         record.message = record.getMessage()
         msg = str(record.message) + "\n"
         if record.exc_info:
-            if not isinstance(record.exc_info, tuple):
-                record.exc_info = sys.exc_info()
-            traceback_msg = "".join(traceback.format_exception(*record.exc_info))
-            msg += traceback_msg
+            if isinstance(record.exc_info, tuple):
+                msg += "".join(traceback.format_exception(*record.exc_info))
+            else:
+                msg += "".join(traceback.format_exception(*sys.exc_info()))
         self.logger(msg, str(record.name), str(record.filename),
                     str(record.funcName), str(record.lineno))
```

Morlic-fo:
I did some further research and learnt that reference cycles in tracebacks are indeed resolved by garbage collection (eventually) as they do not define a ```__del__``` method so Python has no problems resolving them.

Regardless, the diff you posted does resolve the issue anyway:
```
import sys
import traceback
import gc

def foo():
    try:
        raise Exception
    except:
        e = sys.exc_info()
        print traceback.format_exception(*e)

def bar():
    try:
        raise Exception
    except:
        print traceback.format_exception(*sys.exc_info())    
        
print gc.collect()
print gc.garbage

foo()
print gc.collect()  # finds 4 unreachable objects
print gc.garbage  # empty

bar()
print gc.collect()  # finds 0 unreeachable objects
print gc.garbage  # empty
```


Showcase of actual memory-leaks related with ```__del__``` functions:
```
class Collectable():
    def __init__(self):
        self._A = self

class NotCollectable():
    def __init__(self):
        self._A = self
    
    def __del__(self):
        pass

def memory_leak():
    a = Collectable()
    b = NotCollectable()

def no_memory_leak():
    a = Collectable()
    b = NotCollectable()
    del b._A  # manually resolve reference cycle

no_memory_leak()
print gc.collect()
print gc.garbage  # empty

memory_leak()
print gc.collect()
print gc.garbage  # contains an instance of NotCollectable
```

LGM-Doyle:
I added a clarifying comment.

Two things:
- ~~`record.inc_info` is not a local variable.~~
- the code setting `record.exc_info` is the same as the logging library `Handler` code replaced by `_LoggerHandler`.  If the `record` is correctly garbage collected in the library code then it is also correctly collected here.  Any additional handlers added in the future can expect/assume that the first handler behaves like the default handler in upgrading `record.exc_info` from a boolean to the tuple.

Morlic-fo:
> record.inc_info is not a local variable.

```record``` is a/the local variable in this context (eventhough it references an already existing object) as it is defined in the function local namespace.

Anyway, I understand now that the ```record``` object is created not within the stack frame where the Exception is thrown (or any previous) but later and thus no reference to the ```record``` object is stored in the traceback obtained by ```sys.exc_info()```.

Sorry for the troubles, should be fine :)
```
import sys
import traceback
import gc

class Record(object):
    def __init__(self):
        self.exc_info = None

def test(func):
    try:
        raise Exception
    except:
        foo = Record()  # could also be defined outside the try-except clause, same result
        func(foo)
        
def clean(record):
    local_variable = Record()
    local_variable.exc_info = sys.exc_info()
    traceback.format_exception(*local_variable.exc_info)

def not_clean(record):
    record.exc_info = sys.exc_info()
    traceback.format_exception(*record.exc_info)

def test_not_clean2():
    try:
        raise Exception
    except:
        foo = sys.exc_info()
        traceback.format_exception(*foo)

gc.collect()  # just to make sure nothing lingers
    
test(clean)
print gc.collect()

test(not_clean)
print gc.collect()

test_not_clean2()
print gc.collect()
```

LGM-Doyle:
`record` is local.  I misunderstood that the python definition of local includes passed in parameters.  I've fixed (~~struckout~~) my previous comment.

Thanks for the review.

PR #1723, opened 2017-09-04
Cleanup AI turn start functions

Morlic-fo:


default/python/AI/AIstate.py
```diff
@@ -368,6 +366,13 @@ def update_system_status(self):
         min_hidden_attack = 4
         min_hidden_health = 8
 
+        # TODO: Variables that are recalculated each turn from scratch should not be stored in AIstate
```
Cjkjvfnby, line 369: `should be stored in turn_state.State`

https://github.com/freeorion/freeorion/blob/master/default/python/AI/turn_state.py#L10

Cjkjvfnby (approved):

Cjkjvfnby:
Looks ok.  

> Remove redundant local variable  

I prefer term `inline`  like action in PyCharm that do this refactoring.

Morlic-fo:
> I prefer term inline

Thanks for the review, I will remember that for the next time.

PR #1724, opened 2017-09-05
Avoid paths with hostile ships during auto-explore

dbenage-cx:
Did not find an open issue, at least one request is posted in [this thread](http://freeorion.org/forum/viewtopic.php?f=28&t=10651)

Only the shortest route to a system is checked with no attempt to find an alternate route.

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -6904,15 +6904,50 @@ bool MapWnd::IsFleetExploring(const int fleet_id){
     return it != m_fleets_exploring.end();
 }
 
-namespace { //helper function for DispatchFleetsExploring
+namespace {
+    typedef std::unordered_set<int> SystemIDListType;
+    typedef std::unordered_set<int> FleetIDListType;
+    typedef std::vector<int> RouteListType;
+    typedef std::pair<double, RouteListType> OrderedRouteType;
+    typedef std::pair<int, RouteListType> FleetRouteType;
+    typedef std::pair<double, FleetRouteType> OrderedFleetRouteType;
+    typedef std::unordered_map<int, int> SystemFleetMap;
+
+    /** Increases the distance stored as priority for a route when the destination has been previously seen */
+    const double PRIORITY_DEFER_DISTANCE_MULTIPLIER = 10.0;
+
+    /** Arbitrary short-circuit for max routes per destination for each fleet */
+    const int MAX_ROUTES_PER_FLEET = 100;
+
+    /** Number of jumps in a given route */
+    int JumpsForRoute(const RouteListType& route) {
+        int count = static_cast<int>(route.size());
+        if (count > 0) // dont count source system
+            -- count;
+        return count;
+    }
+
+    /** If @p fleet can determine an eta for @p route */
+    bool FleetRouteInRange(const std::shared_ptr<Fleet>& fleet, const RouteListType& route) {
+        std::list<int> route_list;
+        std::copy(route.begin(), route.end(), std::back_inserter(route_list));
+
+        auto eta = fleet->ETA(fleet->MovePath(route_list));
+        if (eta.first == Fleet::ETA_NEVER || eta.first == Fleet::ETA_UNKNOWN || eta.first == Fleet::ETA_OUT_OF_RANGE)
+            return false;
+
+        return true;
+    }
+
+    //helper function for DispatchFleetsExploring
     //return the set of all systems ID with a starlane connecting them to a system in set
-    std::set<int> AddNeighboorsToSet(const Empire *empire, const std::set<int> system_ids){
-        std::set<int> retval;
+    SystemIDListType AddNeighboorsToSet(const Empire *empire, const SystemIDListType& system_ids){
+        SystemIDListType retval;
         auto starlanes = empire->KnownStarlanes();
-        for (int system_id : system_ids) {
+        for (auto system_id : system_ids) {
             auto new_neighboors_it = starlanes.find(system_id);
             if (new_neighboors_it != starlanes.end()){
-                for (int neighbor_id : new_neighboors_it->second) {
+                for (auto neighbor_id : new_neighboors_it->second) {
                     retval.insert(neighbor_id);
                 }
             }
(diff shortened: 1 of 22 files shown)


UI/MapWnd.cpp
```diff
@@ -6985,9 +6988,51 @@ void MapWnd::DispatchFleetsExploring() {
...
+                            auto object_type = object->ObjectType();
+                            if (object_type != OBJ_SHIP && object_type != OBJ_FLEET)
+                                continue;
+
+                            // Owned by non-hostile empire
+                            if (object->Owner() != ALL_EMPIRES &&
+                                Empires().GetDiplomaticStatus(empire_id, object->Owner()) != DIPLO_WAR)
+                            { continue; }
+
+                            auto object_vis_map = GetUniverse().GetObjectVisibilityTurnMapByEmpire(object_id, empire_id);
```
geoffthemedio: What's the purpose of this check?

UI/MapWnd.cpp
```diff
@@ -6985,9 +6988,51 @@ void MapWnd::DispatchFleetsExploring() {
...
+                                is_unobstructed = false;
+                                break;
+                            }
+                        }
+                        if (!is_unobstructed)
+                            break;
+                    }
+                }
+                if (!is_unobstructed) {
+                    std::string obstructed_route_msg("Exploring ship not using obstructed route:");
```
geoffthemedio: purpose of this variable instead of putting as a literal in the loop?

UI/MapWnd.cpp
```diff
@@ -6985,9 +6988,51 @@ void MapWnd::DispatchFleetsExploring() {
...
+                    for (auto route_system_id : route)
+                        obstructed_route_msg += " " + std::to_string(route_system_id);
```
geoffthemedio: could just output "Route: 2 24 82 58 173" or similar, rather than the "Exploring ship..." then a single number repeatedly over several lines

UI/MapWnd.cpp
```diff
@@ -6985,9 +6988,51 @@ void MapWnd::DispatchFleetsExploring() {
                 auto pair = GetPathfinder()->ShortestPath(
                     fleet->SystemID(), system_supply.first, empire_id);
 
+                auto route = pair.first;
+
+                // check if path contains hostile ships/fleets
+                bool is_unobstructed = true;
```
geoffthemedio: I think you're checking if the whole route is unobstructed, not just a single system, so should clarify that in the variable name...

geoffthemedio:
This would be much more useful if it would do the pathing by avoiding pathing through flagged systems, rather than rejecting any exploration destination if the shortest path to it has any threats.

dbenage-cx:
Thinking such functionality would replace use of `ShortestPath` (or the config option would toggle which method to use)
(Edit: osx ci fail is just a timeout)

UI/MapWnd.cpp
dbenage-cx: Checks if empire has observed this object, added comment to clarify

UI/MapWnd.cpp
dbenage-cx: Fixed

UI/MapWnd.cpp
geoffthemedio: Since this is running on the client, when would an empire's client know about an object to run that check if it hadn't observed it?

dbenage-cx: TY, I did not realize they were limited to client info.  I've removed this check.

geoffthemedio:
You could possibly continue to use ShortestPath, or a similar function, by implementing a filtered system graph for the pathfinding, similar to EdgeVisibilityFilter
https://github.com/freeorion/freeorion/blob/master/universe/Pathfinder.cpp#L490

But instead, exclude systems that are marked as dangerous / hostile, or those flagged by the player to avoid pathing through, but always include the destination system

dbenage-cx:
Moved original branch to [here](https://github.com/dbenage-cx/freeorion/tree/defunct_avoid_hostile_during_explore).

Probably not the most ideal solution, I've quite some studying to do on boost graph to know if a property is more beneficial or how to toggle imposing one view on another.
A form of UniverseObject::Contains(predicate) might be worthwhile to replace the hostile fleet check added here and any future needs.

geoffthemedio:
Doesn't seem to work for monsters...

about_to_path
pathed_to_death

[about_to_path_to_death.zip](https://github.com/freeorion/freeorion/files/1362682/about_to_path_to_death.zip)

dbenage-cx:
Ty for testing, will look into it this weekend.

Dilvish-fo:
Perhaps I'm overlooking something, but it looks to me like the code logic is like: if there is nonhostile path to the target system (far_system_id)  that meets fuel requirements, then give a MoveOrder to go to that target system (throwing away the nonhostile route).  So it looks to me like you just need to specify each leg of the nonhostile route, using the append form of MoveOrder.

universe/Pathfinder.cpp
```diff
@@ -889,6 +947,34 @@ std::pair<std::list<int>, double> Pathfinder::PathfinderImpl::ShortestPath(int s
...
 
+std::pair<std::list<int>, double> Pathfinder::ShortestNonHostilePath(int system1_id, int system2_id,
+                                                                     int empire_id) const
+{ return pimpl->ShortestNonHostilePath(system1_id, system2_id, empire_id); }
+
+std::pair<std::list<int>, double> Pathfinder::PathfinderImpl::ShortestNonHostilePath(int system1_id, int system2_id,
+                                                                                     int empire_id) const
+{
+    if (empire_id == ALL_EMPIRES) {
+        ErrorLogger() << "Invalid empire " << empire_id;
```
LGM-Doyle: Why disallow `ALL_EMPIRES` here?  Why not allow weak monsters etc the ability to try and avoid empire fleets?

See also the comment in `System::CointainsHostileFleet()` code.

universe/System.cpp
```diff
@@ -310,6 +310,32 @@ bool System::Contains(int object_id) const {
     return m_objects.find(object_id) != m_objects.end();
 }
 
+bool System::ContainsHostileFleet(int empire_id) const {
+    for (auto obj_id : m_objects) {
+        auto object = GetUniverseObject(obj_id);
```
LGM-Doyle: Change this to 
````
auto fleet = GetFleet(id)
if (!fleet)
    continue;
````
and remove the next 12 lines.
The intent is clearer and the code is shorter.

File: universe/System.cpp [hunk 7]
```diff
@@ -310,6 +310,32 @@ bool System::Contains(int object_id) const {
...
+        auto obj_type = object->ObjectType();
+        if (obj_type <= INVALID_UNIVERSE_OBJECT_TYPE || obj_type >= NUM_OBJ_TYPES) {
+            ErrorLogger() << "Invalid type for object " << obj_id;
+            continue;
+        } else if (obj_type != OBJ_FLEET) {
+            continue;
+        }
+
+        if (object->Owner() != ALL_EMPIRES &&
+            Empires().GetDiplomaticStatus(empire_id, object->Owner()) != DIPLO_WAR)
```
LGM-Doyle: Why not allow `ALL_EMPIRES`?  They are at war with everybody, but it is still a valid state.

I think this will also fix the issue @geoffthemedio observed with the monsters.

LGM-Doyle (commented):
It looks like a good change.

universe/System.cpp
Dilvish-fo: No, I don't think so, for two reasons-- 1) this code already works to treat any ```ALL_EMPIRES``` fleet as at-war, and is essentially how all similar code works because (2) GetDiplomaticStatus() returns  INVALID_DIPLOMATIC_STATUS if either specified empire is ALL_EMPIRES.  (I can only guess that the design decision there was to emphasize that everything ALL_EMPIRES is always at war with everyone else).

universe/Pathfinder.cpp
Dilvish-fo: Although it's an interesting idea, perhaps worth opening a design discussion for, it is certainly consistent with our blockades approach (monsters are very poor at enforcing or evading blockades) that they would likewise be poor (or even worse, if that were possible) at avoiding enemies not even in the same system as them.

dbenage-cx: My initial impression:
When `ALL_EMPIRES` is in the context of Unowned empire, may need a special graph map for Unowned (Unowned ships possibly share visibility).
When `ALL_EMPIRES` refers to multiple empires, there is no clear determination for visibility or hostility.
I will review this and see if an alternative seems feasible.

universe/System.cpp
dbenage-cx: if fleet->Owner() == ALL_EMPIRES, the function proceeds to return True(has a hostile ship), instead of checking for diplo status and potentially continuing to next fleet.
I styled a negate here, so that an additional check could easily be added if later desired, possibly for Fleet::HasArmedShips/HasFighterShips.

dbenage-cx: @Dilvish-fo Opps, I did not see your comments prior to posting mine. :/

geoffthemedio:
Still pathes to death in my test save. Not sure whether the pathing decision gets made after or if it was already made before the load, though...

geoffthemedio:
In my test saved posted above, with the latest version, it still pathes the exploring ship through the monster system.

dbenage-cx:
@geoffthemedio Ah, it is returning to resupply

geoffthemedio:
Arguably part of "auto-explore" that shouldn't result in ships suiciding themselves...

dbenage-cx:
Looking into a more versatile use of `UniverseObjectVisitor`

geoffthemedio:
Now exploring ships don't seem to go back to get refueled at all...

It didn't path into the monster, though.

geoffthemedio:
But it does path to death in this save...
[still_pathing_to_death.zip](https://github.com/freeorion/freeorion/files/1384627/still_pathing_to_death.zip)

dbenage-cx:
Not found any occurrences of pathing into a "known" hostile now.
Improved the determination for refuelling (note in your first savegame, the scout will now path to refuel if stopped and then set to explore again the next turn)

default/stringtables/en.txt
```diff
@@ -1187,6 +1187,12 @@ Sets how fast to render dots for fleet supply lines.
 OPTIONS_DB_FLEET_EXPLORE_IGNORE_HOSTILE
 Ignore hostile ships when a fleet is set to explore
 
+OPTIONS_DB_FLEET_EXPLORE_SYSTEM_ROUTE_LIMIT
+Max number of fleet routes used to determine favored route to an unexplored system. A value of 0 or less indicates no limit.
```
geoffthemedio: This explanation lacks context. If it's supposed to ever be read by a player, it needs some elaboration...

geoffthemedio: please rewrite this text in a later commit...

PR #1725, opened 2017-09-05
Fix directories initialization for headless build.

o01eg:
In headless configuration freeorion client binary doesn't installed so directories initialization fails to determine correct path.
Use `freeoriond` instead `freeorion` as it presents in all configurations.

### DIFF
--- a/util/Directories.cpp
+++ b/util/Directories.cpp
@@ -323,9 +323,9 @@ void InitBinDir(const std::string& argv0) {
             fs::path binary_file = fs::system_complete(fs::path(path_text));
             bin_dir = binary_file.branch_path();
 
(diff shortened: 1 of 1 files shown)


geoffthemedio:
A test report from someone on Linux would be good.

Is the same change not needed for Windows / OSX?

o01eg:
Windows always uses current directory and OSX uses their directories layout so they doesn't test if binary exists.

dbenage-cx:
I tried:
```
$build/freeorion -q
```
and (with `$build-headless/cmake -DBUILD_HEADLESS=ON .. && make`)
```
$build-headless/freeoriond &
$build/freeorion --force-external-server -q
```
(and again, replacing quickstart with hosting multiplayer)

All games started without issue.

PR #1726, opened 2017-09-05
Fix hostless server reaction if on disconnection in WaitingForMPGameJoiners

o01eg:
Fix `Hostless` event processing if unrecoverable disconnection happened during server state `WaitingForMPGameJoiners`.

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -1529,6 +1529,17 @@ sc::result WaitingForMPGameJoiners::react(const CheckStartConditions& u) {
     return discard_event();
 }
 
+sc::result WaitingForMPGameJoiners::react(const Hostless& msg) {
+    TraceLogger(FSM) << "(ServerFSM) WaitingForMPGameJoiners.Hostless";
+    ServerApp& server = Server();
+
+    // Remove the ai processes.  They either all acknowledged the shutdown and are free or were all killed.
+    server.m_ai_client_processes.clear();
+
+    // Don't DisconnectAll. It cause segfault here.
+    return transit<MPLobby>();
+}
+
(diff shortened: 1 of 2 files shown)


server/ServerFSM.cpp
```diff
@@ -1529,6 +1529,17 @@ sc::result WaitingForMPGameJoiners::react(const CheckStartConditions& u) {
     return discard_event();
 }
 
+sc::result WaitingForMPGameJoiners::react(const Hostless& msg) {
```
geoffthemedio, line 1532: when does this occur such that handling it is needed?

o01eg: It possible when AI process fails at start.

geoffthemedio: Can you outline the sequence of events here? Like the server is in the MP lobby, the players setup and start a game with AIs, those AIs fail to start up, the server notices, and somewhere posts itself a `Hostless` event to revert to the MP lobby?

Also, should `Idle` have two `Hostless` reactions declared here:

https://github.com/freeorion/freeorion/blob/master/server/ServerFSM.h#L147

o01eg: `Hostless` event is generated here https://github.com/freeorion/freeorion/blob/master/server/ServerFSM.cpp#L274 instead `ShutdownServer`.
I'll delete doubled `Hostless` reacton. It's a mistake.

geoffthemedio: That's where it is generated in the code, but in what situation / series of events would that occur?

o01eg: Server goes from `MPLobby` to `WaitingForMPGameJoiners`, launches AIs. They connects but quickly fails so the server gets disconnection. They server decides to shutdown but for hostless mode it replaced by `Hostless` event.

PR #1730, opened 2017-09-06
Fix fuel projection for fleet movement

Morlic-fo:
This is supposed to fix what I think is an error in the current fleet movement code.

Please double-check the math / my understanding of game rules:
```
Let f be the fleet's fuel
Let n be the number of unsupplied stops
Let d be the distance from target system to supply

After reaching our target, we have f_new = f_current - n
We then need another d fuel to resupply, i.e. 
        f_new >= d
   <=>  f_current >= n + d

The old condition was
       n < (f_current - (d -1))
   <=> f_current > n + d - 1
```
Now, consider that f_current = 5.1, n=d=3

We do not have enough fuel to return from the target but the second condition is true regardless, i.e. the fleet would possibly be stranded at the target location.

### DIFF
--- a/default/python/AI/MoveUtilsAI.py
+++ b/default/python/AI/MoveUtilsAI.py
@@ -86,11 +86,26 @@ def can_travel_to_system(fleet_id, from_system_target, to_system_target, ensure_
             print "target in Ring 2, has enough aggression"
         elif target_sys_id in ColonisationAI.annexable_ring3 and foAI.foAIstate.character.may_travel_beyond_supply(3):
             print "target in Ring 2, has enough aggression"
-    if (not unsupplied_stops or not ensure_return or
-        target_sys_id in fleet_supplyable_system_ids and len(unsupplied_stops) <= fuel
-        or target_sys_id in ColonisationAI.annexable_ring1 and len(unsupplied_stops) < fuel
-        or target_sys_id in ColonisationAI.annexable_ring2 and foAI.foAIstate.character.may_travel_beyond_supply(2) and len(unsupplied_stops) < fuel - 1
-        or target_sys_id in ColonisationAI.annexable_ring3 and foAI.foAIstate.character.may_travel_beyond_supply(3) and len(unsupplied_stops) < fuel - 2):
+
+    def has_enough_fuel(distance_from_supply):
+        # Let f be the ships fuel
+        # Let n be the number of unsupplied stops
+        # Let d be the distance from target system to supply
+        #
+        # After reaching our target, we have f_new = f_current - n
+        # We then need another d fuel to resupply,
+        # i.e. f_new >= d or f_current >= (n + d)
+        return fuel > (len(unsupplied_stops) + distance_from_supply)
+
+    def movement_allowed(x):
+        return foAI.foAIstate.character.may_travel_beyond_supply(x) and has_enough_fuel(x)
+
+    # TODO: use any-iterator to simplify conditions
+    if (not unsupplied_stops or not ensure_return
+            or target_sys_id in fleet_supplyable_system_ids and movement_allowed(0)
+            or target_sys_id in ColonisationAI.annexable_ring1 and movement_allowed(1)
+            or target_sys_id in ColonisationAI.annexable_ring2 and movement_allowed(2)
+            or target_sys_id in ColonisationAI.annexable_ring3 and movement_allowed(3)):
         return [universe_object.System(sid) for sid in short_path]
     else:
         # print " getting path from 'can_travel_to_system_and_return_to_resupply' ",


default/python/AI/MoveUtilsAI.py
```diff
@@ -86,11 +86,26 @@ def can_travel_to_system(fleet_id, from_system_target, to_system_target, ensure_
...
+        # Let f be the ships fuel
+        # Let n be the number of unsupplied stops
+        # Let d be the distance from target system to supply
+        #
+        # After reaching our target, we have f_new = f_current - n
+        # We then need another d fuel to resupply,
+        # i.e. f_new >= d or f_current >= (n + d)
+        return fuel > (len(unsupplied_stops) + distance_from_supply)
+
+    def movement_allowed(x):
```
Dilvish-fo, line 100: It's a pretty minor thing, but it sort of grates on me to see a variable name x for something that's always expected to be an integer-- howabout 'n' or 'i' or even 'jumps'?  Or, well actually I am going to propose a more complicated calc so I think we would need two such vars, perhaps m and n, though really more descriptive names would be better.

Dilvish-fo (changes requested):
Yes, that was stale these days.  It's my recollection that at least my understanding back then was that there was no fuel cost for passing along a starlane if either end of the starlane was supplied.  So in addition to that making a 1 difference in the final calc, the tally of unsupplied stops was based on that too.  There is another way that tally of unsupplied stops is stale also-- back when that calc was first structured I believe that fleets had to actually stop in a supplied system to get refueled, not just pass through them, so the total number of unsupplied systems passed through is what mattered.  That's not really the case anymore.

These days we need to consider 2 numbers for a flightpath-- the maximum number of sequential unsupplied departures anywhere along the movepath (called perhaps 'most_unsupplied'), and then if the target is unsupplied we need to count the final number of sequential unsupplied departures immediately prior to reaching the target (called perhaps 'final_unsupplied', which may or may not be a different set of jumps, and if the target is supplied or ensure_return is False then final_unsupplied would be zero).  And this fuel helper function would be like
```
def has_enough_fuel():
    return fuel >= max(most_unsupplied, 2*final_unsupplied + 1)
```

For example, let 'S' designate a supplied system, 'U' an unsupplied system, and 'T' our target destination, and consider a movepath SUUUUSUUT.  Since we are starting at a supplied system we need to consider our fuel to be at max even if our last meter reading was below (but otherwise we'd just use current meter as our fuel reading).  Then our longest sequence of unsupplied departures is 4.  Then we look at the final set of unsupplied jumps, which is length 2, but to get back into supply from there would be 3 jumps.  So our fuel has to be at least max(4, 5).

At least, that's how this is looking to me right now, please doublecheck this reasoning.

Dilvish-fo:
I just realized I needed to make an adjustment to my proposed algorithm above.

>Since we are starting at a supplied system we need to consider our fuel to be at max even if our last meter reading was below (but otherwise we'd just use current meter as our fuel reading). 

Even if not starting at a supplied system, we'd need to consider refueling along the way.  One way that comes to mind to handle this is to actually convert the movepath into a string of 'S' and 'U' like in my example, let's call this string the 'movepath_string'.   Working with the sample above, our final T sould be a U in this case, so we'd have 'SUUUUSUUU', and 'SUUUUSUUU'.split('S') is ['', 'UUUU', 'UUU'], and our code could be something like (well, I'm sure the following could still be improved, but here it is off the cuff):
```python
fuel = int(FleetUtilsAI.get_fuel(fleet_id))
max_fuel = int(FleetUtilsAI.get_max_fuel(fleet_id))
movepath_string = get_movepath_string(short_path)
unsupplied_subpaths_a = movepath_string[:-1].split('S')
unsupplied_subpaths_b = movepath_string.split('S')
last_leg_fuel = fuel if len(unsupplied_subpaths_b) == 1 else max_fuel
return_needs = 0 if not ensure_return else 2*len(unsupplied_subpaths_b[-1]) - 1
enough_fuel = (fuel >= len(unsupplied_subpaths_a[0]) and
                         maxfuel >= max(len(p) for p in unsupplied_subpaths_a) and
                         last_leg_fuel >= return_needs)
```
Note: if a change in path were done while the fleet was already out of supply, and the path did not bring it back through supply at any point, the above calc would get it back to its starting point but not necessarily all the way back into supply, so that situation still needs to be dealt with.

Morlic-fo:
You raise some valid points. I currently don't have the time to double-check all of it. 

However, assuming that the target destination is ```final_unsupplied``` steps away from supply is incorrect. It is an upper bound as we walked through a path that took that many steps but there may very well exist shorter path to supply.
So that part should be replaced with the supply-tier lookup of the target system instead.

Dilvish-fo:
If you're not sure you'll have time available anytime soon to tackle the more complete fix, I suppose that what you have here is still a minor improvement over what's in master-- if you'd add a TODO comment flagging the need for a more thorough overhaul then I'd be OK with this small improvement going in now.

Morlic-fo:
Superseded by more rigorous changes in #1761

Issue #672, opened 2016-05-18
Bombardment display UI


MatGB:
I'm finally playtesting all the new bombardment weapons to see how valid the strategy is. It's probably going to be a good, solid strategy, but the current UI is terrible and confusing.

Problem 1): If you select a fleet within a system entirely made up of bombard ships of the same weapon fit, you also need to select each individual ship to initiate a bombard. That bombard will last for one turn only then you need to do the whole thing again.

Problem 2): If you have selected and initiated a bombard, go and do something else within the game, then reload the system (mostly, to check you've ordered a repeat bombard), then the planet you had ordered the bombard for _will display population as if the bombard is already complete_, the only way to check if the planet is actually population zero is to cancel the bombard, select a different system, then come back to this system.

There are other, more minor, issues, but these two are the main ones, the top is a micromanagement problem writ large, if your strategy is to denude the galaxy of life you're going to be clicking a lot of ships _every single turn_. The latter is confusing me constantly.

Ideally selecting a fleet allows you to bombard for that entire fleet, but selecting individual ships allows you to choose which goes where as now. I'm not sure how or if we should enable repeat orders but some method of reminding players is essential, a Sitrep needs to be added telling you what your ships have done as well as your opponent.  The population display issue is a major one that needs dealing with, it's especially a problem when you reload from save and check through your fleets before ending turn.

ghost:
A few more glitches I have observed:

If you have two bombard fleets in a system, they can't attack the same planet.

You can't cancel the bombard order if you move the fleet to a different system.

You can't cancel the bombard order once a planet shows 0 population, even if it isn't actually dead.

When you bombard, the next turn the ship(s) show as bombarding, but no action will be taken.  What you have to do is select the ships, order a bombard, then cancel it, then order it again.  Randomly, you may have to reselect all the ships at each step.

If your bombard strength is more than the current population, but less than pop+growth, the planet isn't wiped, but you can't cancel the glitched order due to the last two glitches above.

Vezzra:
Hm, looks like there are some serious issues with the bombard UI. I wonder if we should try to address those for the release, or keep it flagged for post release (as it's now)?

Comments, opinions?

MatGB:
Depends on how much work it'd be. Currently bombardment isn't there as a strategic help, it needs some fairly major balancing and possibly some effects tweaks as well, so having it a bit buggy for this release (given none of them are new bugs) would be fine.

[...]

PR #1731, opened 2017-09-07
add planet destroyer cannon

TheSilentOne1:
Have you destroyed a planet today? If not, try this PR.

_"Great cannon, great value for your PPs. Finally the quiet I deserve. Highly recommend this."_
Xarix IVth, Trith High Counsellor

Adds planet destroyer cannon ship part, which can be used to destroy enemy planets regardless of shields or defending fleets. Planet destruction can only be prevented if the ship with the pdc is destroyed in combat before the new turn starts.

Pretty much done, tested and should be working, except I haven't taught the AI to utilize this. If someone from the AI team finds the time to take this on, that'd be great.

@V-ktor: if you feel like providing a 3D image for the cannon I'd very much appreciate it.


Dilvish-fo:
Hmmm, seems to me like this is involves some nontrivial design discussion, so [I opened a forum thread](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10690).

Dilvish-fo:
I think it was just a little typo-- when debugging always remember to check your logs; freeorion.log (and console output for me) identified parsing errors.  A typo in the path for your macros were preventing them from being read.  Although it seems technically sufficient to just have the import done at one spot, it makes sense to me to leave it in (and fix) at both.  I haven't actually tested the ingame operation of the part after this fix, but it makes sense to me that this would cause the trouble you see.  Also note the missing newline that should be added.
```diff
diff --git a/default/scripting/ship_parts/General/SP_PLANET_DESTROYER.focs.txt b/default/scripting/ship_parts/General/SP_PLANET_DESTROYER.focs.txt
index 80818e5..e0c0876 100644
--- a/default/scripting/ship_parts/General/SP_PLANET_DESTROYER.focs.txt
+++ b/default/scripting/ship_parts/General/SP_PLANET_DESTROYER.focs.txt
@@ -56,5 +56,5 @@ Part
 
 #include "/scripting/common/upkeep.macros"
 
-#include "/scripting/ship_parts/general/planet_destruction.macros"
+#include "/scripting/ship_parts/General/planet_destruction.macros"
 
diff --git a/default/scripting/specials/DESTRUCTION_COUNTDOWN.focs.txt b/default/scripting/specials/DESTRUCTION_COUNTDOWN.focs.txt
index 747aa71..5f0f8b7 100644
--- a/default/scripting/specials/DESTRUCTION_COUNTDOWN.focs.txt
+++ b/default/scripting/specials/DESTRUCTION_COUNTDOWN.focs.txt
@@ -18,4 +18,4 @@ Special
 
 
 
-#include "/scripting/ship_parts/general/planet_destruction.macros"
\ No newline at end of file
+#include "/scripting/ship_parts/General/planet_destruction.macros"
\ No newline at end of file```

**Comment by TheSilentOne1** [comment]
Unfortunately, that doesn't fix it. I was pretty sure the macro was being be called, because destruction  without a countdown is working, and that uses the macro as well.

**Comment by Dilvish-fo** [comment]
> I was pretty sure the macro was being be called, because destruction without a countdown is working, and that uses the macro as well.

Parse errors for a piece of content will generally at least partly break that piece of content, even if it may not be entirely broken.  

Things that come to mind to further try to help with debugging are to (i) set autogenerated effect descriptions on (at bottom of UI options) and then check the Pedia entries for your part and specials and make sure they match your script the way you expect, (ii) perhaps set effects processing threads to 1, (iii) perhaps set the parser logging to info or trace and review those extra logs carefully, and/or (iv) take out parts of your script until you get to a subset that is working 100 percent right, and then slowly add bits making sure each bit is working right before adding another.

File: default/scripting/ship_parts/General/SP_PLANET_DESTROYER.focs.txt [hunk 1]
```diff
@@ -7,7 +7,23 @@ Part
     buildtime = 8
     tags = [ "CTRL_ALWAYS_BOMBARD" ]
     location = All
-    effectsgroups = [    
+    effectsgroups = [
+
+        // pdc gives speed malus increasing with refinement
+        EffectsGroup
```
geoffthemedio: refinements should never give a penalty when researched

TheSilentOne1: Is that a set in stone rule? The idea is that the countdown decrease with each refinement is very powerful, and the increasing speed malus is supposed to even that out somewhat.

geoffthemedio: > Is that a set in stone rule?

yes

Things players can pick to turn on / off can have penalties, not techs that can't be turned off.

TheSilentOne1: ok

TheSilentOne1:
I currently don't have any plans to add more features, so from my side open for reviewing.

UI/SidePanel.cpp
```diff
@@ -1476,6 +1560,45 @@ std::set<std::shared_ptr<const Ship>> AutomaticallyChosenBombardShips(int target
...
+    if (target_planet->OwnedBy(empire_id))
+        return retval;
+
+    for (auto& ship : Objects().FindObjects<Ship>()) {
+        // owned ship is capable of bombarding a planet in this system
+        if (!AvailableToDestroy(ship, system_id, empire_id))
+            continue;
+
+        // Select ship if the planet contains a content tag specified by the ship, or ship is tagged to always be selected
+        for (const std::string& tag : BombardTagsForShip(ship)) {
```
dbenage-cx: Should these check for destroy tags instead?

dbenage-cx:
When specific to just destroying planet(s), Destroy seems too generic (CanDestroy/DestroyOrder/etc), suggest DestroyPlanet.

As it is of limited scope, curious if the new part class would be better suited as a tag?
A tag might be less efficient to check for, but could allow use in a hull or possibly ~~a special~~ with a species.

If the part class is retained, looks like unit tests need update.  And as default/scripting/ship_parts are organized by class, please move those definitions to a class specific sub-directory.

TheSilentOne1:
>When specific to just destroying planet(s), Destroy seems too generic (CanDestroy/DestroyOrder/etc), suggest DestroyPlanet.

Agreed.

>As it is of limited scope, curious if the new part class would be better suited as a tag?
A tag might be less efficient to check for, but could allow use in a hull or possibly a special.

Not sure I completely understand what you mean, do you suggest to make the pdc PC_BOMBARD or similar, with a tag like "planet_destruction" setting it apart from other bombard weapons?

dbenage-cx:
Yes, but maybe for pdc PC_GENERAL, unless it can also enable bombard orders.
Ship::CanDestroy could return HasTag(ShipDesign::TAG_DESTROY_PLANET), and a similar replacement for ShipDesign::CanDestroy

TheSilentOne1:
>Yes, but maybe for pdc PC_GENERAL

Sounds good, will do. (Not sure when I will get around to do it though, hopefully later this week.)

TheSilentOne1:
I've implemented the changes you suggested.

server/ServerApp.cpp
```diff
@@ -30,6 +30,7 @@
 #include "../util/SitRepEntry.h"
 #include "../util/ScopedTimer.h"
 #include "../util/Version.h"
+//#include "../client/human/HumanClientApp.h"
```
o01eg: I suppose it's dangerous change.

UI/SidePanel.cpp
```diff
@@ -1476,6 +1560,45 @@ std::set<std::shared_ptr<const Ship>> AutomaticallyChosenBombardShips(int target
...
+* @param target_planet_id ID of Planet to potentially destroy
+*/
+std::set<std::shared_ptr<const Ship>> AutomaticallyChosenDestroyShips(int target_planet_id) {
+    std::set<std::shared_ptr<const Ship>> retval;
+
+    int empire_id = HumanClientApp::GetApp()->EmpireID();
+    if (empire_id == ALL_EMPIRES)
+        return retval;
+
+    std::shared_ptr<const Planet> target_planet = GetPlanet(target_planet_id);
```
geoffthemedio: prefer auto to explicit verbose type in cases like this

UI/SidePanel.cpp
```diff
@@ -1577,6 +1701,12 @@ void SidePanel::PlanetPanel::Refresh() {
         bombard_ships.insert(autoselected_bombard_ships.begin(), autoselected_bombard_ships.end());
     }
 
+    std::set<std::shared_ptr<const Ship>> destroy_ships = ValidSelectedDestroyShips(SidePanel::SystemID());
```
geoffthemedio: auto?

UI/SidePanel.cpp
```diff
@@ -1577,6 +1701,12 @@ void SidePanel::PlanetPanel::Refresh() {
...
+    if (destroy_ships.empty()) {
+        std::set<std::shared_ptr<const Ship>> autoselected_destroy_ships = AutomaticallyChosenDestroyShips(m_planet_id);
```
geoffthemedio: auto?

UI/SidePanel.cpp
```diff
@@ -2223,6 +2369,19 @@ namespace {
                 }
             }
         }
+
+        // is selected ship ordered to destroy a planet?  If so, recind that order
+        if (ship->OrderedDestroyPlanet() != INVALID_OBJECT_ID) {
+            for (const auto& id_and_order : orders) {
+                if (std::shared_ptr<DestroyPlanetOrder> order = std::dynamic_pointer_cast<DestroyPlanetOrder>(id_and_order.second)) {
```
geoffthemedio: auto

UI/SidePanel.cpp
```diff
@@ -2358,6 +2517,51 @@ void SidePanel::PlanetPanel::ClickBombard() {
     OrderButtonChangedSignal(m_planet_id);
 }
 
+void SidePanel::PlanetPanel::ClickDestroy() {
+    // order or cancel planet destruction, depending on whether it has previously
+    // been ordered
+    std::shared_ptr<const Planet> planet = GetPlanet(m_planet_id);
```
geoffthemedio: auto

UI/SidePanel.cpp
```diff
@@ -2358,6 +2517,51 @@ void SidePanel::PlanetPanel::ClickBombard() {
...
+    if (!planet ||
+        !m_order_issuing_enabled ||
+        (planet->CurrentMeterValue(METER_POPULATION) <= 0.0 && planet->Unowned()))
+    { return; }
+
+    int empire_id = HumanClientApp::GetApp()->EmpireID();
+    if (empire_id == ALL_EMPIRES)
+        return;
+
+    std::map<int, std::set<int>> pending_destroy_orders = PendingDestroyPlanetOrders();
```
geoffthemedio: auto, auto

UI/SidePanel.cpp
```diff
@@ -2358,6 +2517,51 @@ void SidePanel::PlanetPanel::ClickBombard() {
...
+    std::map<int, std::set<int>>::const_iterator it = pending_destroy_orders.find(m_planet_id);
+
+    if (it != pending_destroy_orders.end()) {
+        const std::set<int>& planet_destroy_orders = it->second;
```
geoffthemedio: auto

UI/SidePanel.cpp
```diff
@@ -2358,6 +2517,51 @@ void SidePanel::PlanetPanel::ClickBombard() {
...
+        // cancel previous destroy orders for this planet
+        for (int order_id : planet_destroy_orders)
+        { HumanClientApp::GetApp()->Orders().RescindOrder(order_id); }
+
+    } else {
+        // order selected destroy ship to destroy planet
+        std::set<std::shared_ptr<const Ship>> destroy_ships = ValidSelectedDestroyShips(planet->SystemID());
```
geoffthemedio: auto

UI/SidePanel.cpp
```diff
@@ -2358,6 +2517,51 @@ void SidePanel::PlanetPanel::ClickBombard() {
...
+
+        if (destroy_ships.empty()) {
+            std::set<std::shared_ptr<const Ship>> autoselected_destroy_ships = AutomaticallyChosenDestroyShips(m_planet_id);
```
geoffthemedio: auto, etc. you get the point

universe/Effect.h
```diff
@@ -765,6 +765,50 @@ class FO_COMMON_API AddSpecial : public EffectBase {
     void serialize(Archive& ar, const unsigned int version);
 };
 
+/** Removes all buildings from the target object. */
+class FO_COMMON_API RemoveAllBuildings : public EffectBase {
```
geoffthemedio: This seems unnecessary. There is already a destroy effect that can be targeted to all buildings on a planet.

default/stringtables/en.txt
```diff
@@ -6283,6 +6295,21 @@ EFFECT_BIOWEAPON
 EFFECT_BIOWEAPON_LABEL
 Bioweapon Attack
 
+EFFECT_PLANET_TARGETED
+Our planet % planet is being targeted by a planet destroyer cannon!
```
geoffthemedio: this looks malformed...

universe/Condition.cpp
```diff
@@ -8479,6 +8479,144 @@ unsigned int OrderedBombarded::GetCheckSum() const {
...
+        EvalImpl(matches, non_matches, search_domain, OrderedDestroyedSimpleMatch(subcondition_matches));
+    }
+    else {
+        // re-evaluate contained objects for each candidate object
+        ConditionBase::Eval(parent_context, matches, non_matches, search_domain);
+    }
+}
+
+bool OrderedDestroyed::RootCandidateInvariant() const
+{
```
geoffthemedio: move to end of prev line, or move actual code and closing `}` to same line as this

util/Order.cpp
```diff
@@ -804,6 +804,103 @@ bool BombardOrder::UndoImpl() const {
 }
 
 ////////////////////////////////////////////////
+// DestroyPlanetOrder
+////////////////////////////////////////////////
+DestroyPlanetOrder::DestroyPlanetOrder() :
```
geoffthemedio: why is this needed, as opposed to making it a bombard effect...?

default/stringtables/en.txt
MatGB: Style is 
Planetname: stuff happens 

And as many pedia links as you can, and ideally a link to the attacking ship. 

Also remember to put the sitrep into the priority order list so it appears near the top.

geoffthemedio: ... ?? what? I have no idea what this comment is referring to.

universe/Effect.h
TheSilentOne1: If I'm not mistaken, this requires a new scope for the buildings on the planet. With this solution, the script is shorter and easier to read (all effects of planet destruction are kept in one place).

util/Order.cpp
TheSilentOne1: I don't understand. Can you elaborate how that might work?

default/stringtables/en.txt
MatGB: Sorry, was on phone.
default/customizations/common_user_customizations.txt

I don't know why it's in that file, it used to be in the stringtable, it's right that it's not there but it's not really a customisation and it's important that it's updated for new effects.

TheSilentOne1: Done, thanks for pointing out.

universe/Effect.h
geoffthemedio: Use `Effect::Conditional` to narrow the target set when executing an effectsgroup.

util/Order.cpp
geoffthemedio: Make the bombard effect of a part destroy a planet. When told to bombard a planet, that ship will destroy it.

default/stringtables/en.txt
geoffthemedio: I still have no idea what you're talking about.

The point is that the stringtable entry this comment thread is pointing to looks malformed. The lone % doesn't make sense.

default/stringtables/en.txt
```diff
@@ -6283,6 +6295,21 @@ EFFECT_BIOWEAPON
...
+At %planet%: our planet is being targeted by a planet destroyer cannon from the ship %ship%!
+
+EFFECT_PLANET_DESTRUCTION
+We have destroyed the %empire% planet %planet%!
+
+EFFECT_PLANET_DESTRUCTION_LABEL
+Destruction of hostile planet
+
+EFFECT_PLANET_DESTRUCTION_OWN
+Our planet %planet% was destroyed by %empire%!
```
MatGB: Superfluous 'Our planet', the link indicates it is a planet already.

default/stringtables/en.txt
```diff
@@ -6283,6 +6295,21 @@ EFFECT_BIOWEAPON
 EFFECT_BIOWEAPON_LABEL
 Bioweapon Attack
 
+EFFECT_PLANET_TARGETED
+At %planet%: our planet is being targeted by a planet destroyer cannon from the ship %ship%!
+
+EFFECT_PLANET_DESTRUCTION
+We have destroyed the %empire% planet %planet%!
```
MatGB: Again, planet first, or possibly system (will the link still work after it's destroyed?)
At %system%: we destroyed %empire% %planet%

util/Order.cpp
TheSilentOne1: There might be cases where a ship is capable of destroying as well as bombarding a planet.

default/stringtables/en.txt
```diff
@@ -6283,6 +6295,21 @@ EFFECT_BIOWEAPON
...
+
+EFFECT_PLANET_DESTRUCTION_LABEL
+Destruction of hostile planet
+
+EFFECT_PLANET_DESTRUCTION_OWN
+Our planet %planet% was destroyed by %empire%!
+
+EFFECT_PLANET_DESTRUCTION_OWN_LABEL
+Destruction of own planet
```
MatGB: When is this used? If it's a sitrep it needs to say where.

default/stringtables/en.txt
MatGB: Ah, I was commenting on the actual text which was badly formatted and not per current style, both have subsequently been fixed.

default/stringtables/en.txt
TheSilentOne1: It will point to the remaining asteroids. Name stays the same.

default/stringtables/en.txt
TheSilentOne1: It's the label/category, as opposed to "Destruction of hostile planet".

MatGB: Of course it is, sorry, it even says _LABEL. I need more coffee.

universe/Effect.h
TheSilentOne1: Can you give an example how to do that?

default/stringtables/en.txt
Vezzra: That depends. Maybe the problem is that you actually had too much... 😉

util/Order.cpp
geoffthemedio: That can be relevant for a variety of bombard effects that a player .ay notalways eant to use all together or not at all. The solution should probably be more general and adaptable than this whixh is specific to a particular part.

TheSilentOne1: Still not sure if I completely understand what you mean. Do you mean I should merge the DestroyPlanet Order into the BombardOrder, e. g. by adding a bool is_destroy_order, which is checked to determine if it is actually a bombard or a planet destruction order?

dbenage-cx: Not looked into the idea fully, could possibly have a base order specialized by script definitions.  That concept may not suitable for the scope of this PR though.

geoffthemedio: If it's important to be able to use only particular bombard weapons when bombarding, then there should be a general way to do this. Your particular new bombard weapon probably doesn't warrant its own separate order class, when there could be various other similar bombard weapons as well. Not just a single bool flag, but a way to control it in general. Or we can accept that it's all or nothing.

MatGB:
Perhaps exclusions? You can only have one type of bombard weapon per ship would be simple to do and work well enough.

geoffthemedio:
@MatGB Sounds reasonable.

TheSilentOne1:
Agreed.

TheSilentOne1:
Almost done with removing the DestroyPlanetOrder and using BombardOrder instead. There's one thing though which I haven't been able to figure out so far. For some reason planet->AboutToBeBombarded() and ship->OrderedBombardPlanet are not being reset during turn processing (although the bombard order is being removed as it should). I'll keep looking for the culprit but if someone has an idea what may be the issue I'm, as usual, thankful for advice.

dbenage-cx:
IIRC (I was stuck as well) the previous turn order is not rescinded (the related code in sidepanel only deals with the orders for the current turn).  If that is the case, after the objects status is reset, it might be later updated from a previous order.

MatGB:
There are several known significant UI bugs for bombarding #672 some of which have been tweaked but never fully solved.

TheSilentOne1:
@dbenage-cx: Unfortunately, I don't think that's it. The orders are cleared as they are supposed to. Maybe something is amiss with resetting the object status. Do you happen to know where that's done?

@MatGB: I'm aware of those bugs, and the ones who are not resolved here I plan to deal with when revising bombardement. That's next on my list, and I've started with it, but I want to finish the pdc first.

dbenage-cx:
[CleanUpBombardStateInfo](https://github.com/freeorion/freeorion/blob/master/server/ServerApp.cpp#L2766) and [BombardOrder::UndoImpl](https://github.com/freeorion/freeorion/blob/master/util/Order.cpp#L780).
Might note the [documentation of OrderSet](https://github.com/freeorion/freeorion/blob/master/util/OrderSet.h#L20), which implies a previous turns orders are executed (can not invoke Undo, as it is deleted for the current turn).
When the client updates/refreshes from the server data, that previous order could override some client change not sent to server (e.g. a call to `Planet::ResetIsAboutToBeBombarded`).

TheSilentOne1:
> When the client updates/refreshes from the server data, that previous order could override some client change not sent to server (e.g. a call to Planet::ResetIsAboutToBeBombarded).

Yes, that should be it. The server sends its gamestate and overrides the client's gamestate AFTER orders have already been rescinded on the client side. So the solution should be to rescind the orders on the client side after the server has sent the complete gamestate. But I'm not sure how the client knows it has received the final universe data. Can I somehow check for it in ClientApp::HandleTurnPhaseUpdate()?

dbenage-cx:
From `Order::Undo` (Which is used by `OrderSet::RescindOrder`):
> If this function returns true, it reverts the game state to what it was before this order was executed

If it could be managed to somehow undo an order from a previous turn, it probably would not have the desired effect.
This seems like the purview of the persistent order flag, where the order has a special casing between reverting to previous state(undo) and preventing the effects from further applying on a given turn(cancel?).   (I've not looked at it fully yet, but seems to be missing at least the suggestion of changing `m_orders.Reset()`->`m_orders.ResetNonPersistentOrders()` ?)

TheSilentOne1:
Thanks for your help, @dbenage-cx, I was able to fix it now. Largest problem was my ignorance of the client/server interaction. In HumanClientApp.cpp, HandleTurnUpdate() now resets the universe's bombardment state, and after that cleans up old orders and re-execute the persisting ones.

TheSilentOne1:
> Use Effect::Conditional to narrow the target set when executing an effectsgroup.

I'd like to use this and remove "RemoveAllSpecials" and "RemoveAllBuildings", but I can't find an example where "If..." is used to narrow down the scope. If someone does know how to do that, kindly give me a hint. Alternatively, if we don't want to keep the two functions, I can get rid of them by declaring additional scopes for buildings and specials in the pdc script.
Apart from that, open for (further) reviewing.

geoffthemedio:
https://github.com/freeorion/freeorion/blob/master/default/scripting/fields/fields.macros#L17

Won't help with RemoveAllSpecials, though, I think...

TheSilentOne1:
Another way to handle it would be something like `RemoveSpecial name = "ALL"`. Or is that too hacky?

geoffthemedio:
Should there even be a RemoveAllSpecials effect? This assumes quite a bit about how every special is implemented... Perhaps some are reasonable to leave on planets that are destroyed?

MatGB:
Given that even using the cannon is going to be fairly unusual and I can't think of a current special that wouldn't get destroyed if the planet it was on went up (scratch that, can't think of one that shouldn't be), I don't think it's worth worrying about that much, if a planet blows up specials are lost is fine to get this working surely?

TheSilentOne1:
> Should there even be a RemoveAllSpecials effect?

Would you rather have `[[REMOVE_MOST_SPECIALS]]`? =
```
RemoveSpecial name = "name_1"
RemoveSpecial name = "name_2"
RemoveSpecial name = "name_n"
```
One disadvantage of this method would be that most new specials (and buildings) would need to be added to the list.

agrrr3:
General scripting-wise I think it would be better to add a glob option than (only) remove all. 
The remove-all could look like this

RemoveSpecial matching = "*"

With a glob feature one could handle families of specials like e.g. PLANET_MORPHING_TO_DESERT, PLANET_MORPHING_TO_BARREN... 

RemoveSpecial matching = "PLANET_MORPHING_TO_*"

Is there a reason not to provide globs? Should I add a feature request?

Vezzra:
> General scripting-wise I think it would be better to add a glob option

If we add something like this, it wouldn't make sense to provide that for `RemoveSpecial` only, but as a general feature wherever universe objects, specials, species etc. are referenced by name (and where it makes sense to reference more than one specific element of course).

dbenage-cx:
Curious if multiple entries could be allowed for `name`, even if it is just shorthand for multiple effects, something akin to:
```
RemoveSpecial name = [ [[ALL_SPECIALS]] ]
RemoveBuilding name = [ [[ALL_BUILDINGS]] ]
```

geoffthemedio:
What could be done is to have, in the parser, a `RemoveAllSpecials` case, which will set up the resulting `Effect::RemoveSpecial` object so that it removes all specials. `UniverseObject::RemoveAllSpecials` can be added, and if `Effect::RemoveSpecial` is constructed and executed with `nullptr` as its special name `ValueRef`, then it would remove all specials. This is distinct from passing a constant empty string as the `ValueRef`, which would maintain the current behaviour of trying to remove the special named `""`, which does nothing.

TheSilentOne1:
@geoffthemedio: changed RemoveAllSpecials as you suggested.
Rebased.

TheSilentOne1:
Adjusted to recent FleetWnd changes and rebaesd.

geoffthemedio:
My objection to RemoveAllBuildings was stronger than RemoveAllSpecials...

TheSilentOne1:
> Seems that there is still something not right with cleaning up orders or the game state on the server side. Need to look into it again.

A quick note that unfortunately, I'm still stuck. Problem is: during turn processing, the server is supposed to reset the bombardment state of planets and ships. However, when it sends the final gamestate to the client, planets and ships still have the bombardment info set - although I even reset it right before the server sends the gamestate.
This issue shouldn't be related to persisting orders, which I have disabled until this can be sorted out.

geoffthemedio:
Is that info copied and serialized in the appropriate functions?

TheSilentOne1:
> Is that info copied and serialized in the appropriate functions?

From SerializeUniverse.cpp:
```
void Planet::serialize(Archive& ar, const unsigned int version) {
...
        & BOOST_SERIALIZATION_NVP(m_is_about_to_be_bombarded)
}
```
```
void Ship::serialize(Archive& ar, const unsigned int version)
{
...
        & BOOST_SERIALIZATION_NVP(m_ordered_bombard_planet_id)
}
```

So yes, I think so.

geoffthemedio:
And in the copy function(s)?

TheSilentOne1:
Where do I find them?

geoffthemedio:
Search for `::Copy`

TheSilentOne1:
From Planet::Copy():
```
if (vis >= VIS_FULL_VISIBILITY) {
                this->m_is_about_to_be_bombarded =  copied_planet->m_is_about_to_be_bombarded;
[...]
}
```

And from Ship::Copy():
```
if (vis >= VIS_FULL_VISIBILITY) {
                this->m_ordered_bombard_planet_id = copied_ship->m_ordered_bombard_planet_id;
[...]
}
```

TheSilentOne1:
Moving `this->m_ordered_bombard_planet_id = copied_ship->m_ordered_bombard_planet_id;` into `if (vis >= VIS_BASIC_VISIBILITY)` seems to solve the ship status being reset, doesn't seem to work for `this->m_is_about_to_be_bombarded` so far though.

TheSilentOne1:
I think it is fixed now. I will test it some more later today and re-enable persistent orders. Thanks for the pointer to ::Copy, @geoffthemedio.

geoffthemedio:
It's not a matter of "knowing" so much as the UI updating properly. If `m_is_about_to_be_bombarded` isn't passed or updated, then the cancel bombardment button might not show on the planet panel.

They shouldn't need to be cleared client-side, as long as the default values are false. They do need to be re-set client-side during / after unpacking the gamestate though. But not serializing the flags at all is probably reasonable.

TheSilentOne1:
> But not serializing the flags at all is probably reasonable.

Should I remove the flags only from `Planet::Copy()` and `Ship::Copy()`, or from SerializeUniverse as well? Not sure if the latter would cause problems with saving/loading.

geoffthemedio:
Might affect save compatibility. The server might need to re-determine the flags if not saved... not sure how a mid turn save works for that...

TheSilentOne1:
Leaving the flags in SerializeUniverse.cpp, mid-turn saving works as expected to. Removing them breaks save compatibility and seems to break some more stuff as well. Leave as is?
Squashed recent commits.

geoffthemedio:
Perhaps... what's the current status of this? Still appears to have some separate destroy vs. bombard stuff...

TheSilentOne1:
Instead of `ordered_destroyed_by`, maybe the `ordered_bombarded_by` condition could be used. It would require a way to check if a ship has the pdc part; could be maybe achieved with `DesignHasPart`? I'll look into it.
Apart from that, I can only think of the destroy / bombard button. Probably not a good idea to merge them. Fleets may contain bombers and destroyers, and the player will want to choose which one to use.

TheSilentOne1:
> could be maybe achieved with DesignHasPart? 

Might not even need that, because effect is caused by the pdc part?
... will try.

geoffthemedio:
There are lots of potential effects that could be implemented as bombardment actions... I don't think it's necessary to have a separate button for each of them. Maybe a "include in auto bombard selection" as a flag for a part, so that it won't automatically be chosen to bombard a planet, and needs to be selected to be ordered to use its effect...

TheSilentOne1:
> I don't think it's necessary to have a separate button for each of them.

I see that a separate button, at least with the current implementation, means a lot of additional code. On the other hand, a button reading "bombard", that only appears when a ship with planet destruction capabilities is selected, and causes planet destruction instead of bombardment, as opposed to what it reads, makes the UI less intuitive; so I'd rather hang on to it.

geoffthemedio:
It's not intuitive if one considers planet destruction to be just another bombardment effect...  But I can see it not being.

Perhaps there should be a flag set for each bombardment part, which marks it as any combination of things like population destroying, infrastructure destroying, or planet destroying. Likely populations would have different opinions of empires that use parts labelled with each. This could also be used to create separate buttons in the GUI, without being overly specific to a single part.

TheSilentOne1:
> Perhaps there should be a flag set for each bombardment part, which marks it as any combination of things like population destroying, infrastructure destroying, or planet destroying.

That sounds fine to me. I should have time to resume work on this soon.

PR #1734, opened 2017-09-08
add "dismiss sensor ghost" also to fleet btn / map wnd

TheSilentOne1:
I suggest to move dismissing sensor ghosts from the fleet wnd to the map wnd.
To dismiss a sensor ghost, the player currently needs to first l-click the fleet, then move the mouse to the fleet wnd, and dismiss the fleet per r-click popup.
With this PR, the player only need to r-click the fleet button on the map to get to the popup, which saves one l-click and moving the mouse.

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -5495,6 +5495,44 @@ void MapWnd::FleetButtonRightClicked(const FleetButton* fleet_btn) {
         return;
     }
 
+    // if fleetbutton holds currently not visible fleets, offer to dismiss them
+    int empire_id = HumanClientApp::GetApp()->EmpireID();
+    std::vector<int> sensor_ghosts;
+
+    // find sensor ghosts
+    for (int fleet_id : btn_fleets) {
+        auto fleet = GetFleet(fleet_id);
+        if (!fleet)
+            continue;
+        if (fleet->OwnedBy(empire_id))
+            continue;
+        if (GetUniverse().GetObjectVisibilityByEmpire(fleet_id, empire_id) >= VIS_BASIC_VISIBILITY)
+            continue;
+        sensor_ghosts.push_back(fleet_id);
+    }
+
+    // should there be sensor ghosts, offer to dismiss them
+    if (sensor_ghosts.size() > 0) {
+        auto popup = GG::Wnd::Create<CUIPopupMenu>(fleet_btn->LowerRight().x, fleet_btn->LowerRight().y);
+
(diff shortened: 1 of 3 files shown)


geoffthemedio:
Why move and not duplicate?

TheSilentOne1:
I thought about that as well, and that would ofc also be fine; but why even offer a complicated way when it can be done simple(r)?

geoffthemedio:
the fleets window might already be open

TheSilentOne1:
I've restored the fleet wnd "dismiss sensor ghost" functionality.

LGM-Doyle:
I like this idea.

Reasons to keep the fleet wnd option are when the player will not want to hide all fleets associated with a fleet button:
- if there are multiple fleets with a fleet button and not all of the fleet are sensor ghosts
- if the player knows that some of the ghosts are harmless, mobile, or are known quantities and others are hostile, stationary or unknown marauding stealthy fleets.  The player may only want to dismiss the harmless fleets and keep the others.

When I tested this and dismissed ghost fleets from the map window, it did not update the open fleet window.

A change you might want to consider, is to have the dismiss ghost option if **any** fleets are ghosts.  Then the action would dismiss ghost fleets, but not non-ghost fleets.

TheSilentOne1:
Sounds good to me, I'll implement your suggestions.

geoffthemedio:
Looks fine, if it works for you.

TheSilentOne1:
I've implemented LGM-Doyles suggestion and squashed the commits. Light testing suggests it works as intended.

UI/MapWnd.cpp
```diff
@@ -5495,6 +5495,41 @@ void MapWnd::FleetButtonRightClicked(const FleetButton* fleet_btn) {
...
+    // if fleetbutton holds currently not visible fleets, offer to dismiss them
+    int empire_id = HumanClientApp::GetApp()->EmpireID();
+    std::vector<int> sensor_ghosts;
+
+    // find sensor ghosts
+    for (int fleet_id : btn_fleets) {
+        auto fleet = GetFleet(fleet_id);
+        if (!fleet) continue;
+        if (fleet->OwnedBy(empire_id)) continue;
+        if (GetUniverse().GetObjectVisibilityByEmpire(fleet_id, empire_id) >= VIS_BASIC_VISIBILITY) continue;
```
geoffthemedio: put the continues on their own lines

geoffthemedio:
otherwise looks OK

PR #1737, opened 2017-09-09
Promote usage of turn_state.State

Morlic-fo:
This PR aims to further reduce the clutter that exists in the AI code.

1) Move various functionality from ```ColonizationAI.survey_universe()``` to the ```turn_state``` module. The responsibility of updating the turn_state is given to ```FreeOrionAI.generateOrders()```. Further work is needed to eventually completely remove ```survey_universe()```.

2) Unify the usage of various module level variables by the function ```turn_state.State.get_empire_planets_by_system()```. By default, it returns a map from systems where the AI owns a planet to a tuple of planets in that system. The parameter ```include_outposts``` defaults to True but can be set to False to only consider inhabited planets.
Optionally, a system_id can be passed to the function. If specified, return the empire owned planets in the system instead of the whole map.
As the function gets called at various places, the map is cached with (mostly) only read-access. The data structure is safe enough to return without having to worry about outside modifications which are still possible but require special effort so are irrelevant for our work.

3) Inline or strictly remove multiple (cross-)module level variables that serve no purpose in today's code.

### DIFF
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -709,7 +709,7 @@ def generate_production_orders():
                     use_sys = ([(-1, INVALID_ID)] + sorted([(dist, sys_id) for sys_id, dist in distance_map.items()]))[:2][-1][-1]  # kinda messy, but ensures a value
                 if use_sys != INVALID_ID:
                     try:
-                        use_loc = state.get_empire_planets_by_system(use_sys, include_outposts=True)[0]
+                        use_loc = state.get_empire_planets_by_system(use_sys)[0]
                         res = fo.issueEnqueueBuildingProductionOrder(building_name, use_loc)
                         print "Enqueueing %s at planet %d (%s) , with result %d" % (building_name, use_loc, universe.getPlanet(use_loc).name, res)
                         if res:
@@ -736,8 +736,7 @@ def generate_production_orders():
         if not bh_pilots and len(queued_building_locs) == 0 and (red_pilots or not already_got_one):
             use_loc = None
             nominal_home = homeworld or universe.getPlanet(
-                (red_pilots + state.get_empire_planets_by_system(AIstate.empireStars[fo.starType.red][0],
-                                                                 include_outposts=True))[0])
+                (red_pilots + state.get_empire_planets_by_system(AIstate.empireStars[fo.starType.red][0]))[0])
             distance_map = {}
             for sys_id in AIstate.empireStars.get(fo.starType.red, []):
                 if sys_id == INVALID_ID:
@@ -748,7 +747,7 @@ def generate_production_orders():
                     pass
             red_sys_list = sorted([(dist, sys_id) for sys_id, dist in distance_map.items()])
             for dist, sys_id in red_sys_list:
-                for loc in state.get_empire_planets_by_system(sys_id, include_outposts=True):
+                for loc in state.get_empire_planets_by_system(sys_id):
                     planet = universe.getPlanet(loc)
                     if planet and planet.speciesName not in ["", None]:
                         species = fo.getSpecies(planet.speciesName)
@@ -797,7 +796,7 @@ def generate_production_orders():
                 use_sys = ([(-1, INVALID_ID)] + sorted([(dist, sys_id) for sys_id, dist in distance_map.items()]))[:2][-1][-1]  # kinda messy, but ensures a value
             if use_sys != INVALID_ID:
                 try:
-                    use_loc = state.get_empire_planets_by_system(use_sys, include_outposts=True)[0]
+                    use_loc = state.get_empire_planets_by_system(use_sys)[0]
                     res = fo.issueEnqueueBuildingProductionOrder(building_name, use_loc)
                     print "Enqueueing %s at planet %d (%s) , with result %d" % (building_name, use_loc, universe.getPlanet(use_loc).name, res)
                     if res:
@@ -876,7 +875,7 @@ def generate_production_orders():
                 use_sys = ([(-1, INVALID_ID)] + sorted([(dist, sys_id) for sys_id, dist in distance_map.items()]))[:2][-1][-1]  # kinda messy, but ensures a value
             if use_sys != INVALID_ID:
                 try:
-                    use_loc = state.get_empire_planets_by_system(use_sys, include_outposts=True)[0]
+                    use_loc = state.get_empire_planets_by_system(use_sys)[0]
                     res = fo.issueEnqueueBuildingProductionOrder(building_name, use_loc)
                     print "Enqueueing %s at planet %d (%s) , with result %d" % (building_name, use_loc, universe.getPlanet(use_loc).name, res)
                     if res:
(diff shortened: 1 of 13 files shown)


default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -218,3 +219,54 @@ def get_partial_visibility_turn(obj_id):
     """
     visibility_turns_map = fo.getUniverse().getVisibilityTurnsMap(obj_id, fo.empireID())
     return visibility_turns_map.get(fo.visibility.partial, -9999)
+
+
+class ReadOnlyDict(Mapping):
```
Cjkjvfnby, line 224: Looks like you don't use Mapping and it can be replaced to object.

default/python/AI/ProductionAI.py
```diff
@@ -710,7 +710,7 @@ def generate_production_orders():
                     use_sys = ([(-1, INVALID_ID)] + sorted([(dist, sys_id) for sys_id, dist in distance_map.items()]))[:2][-1][-1]  # kinda messy, but ensures a value
                 if use_sys != INVALID_ID:
                     try:
-                        use_loc = state.get_empire_planets_by_system(include_outposts=True)[use_sys][0]
+                        use_loc = state.get_empire_planets_by_system(use_sys, include_outposts=True)[0]
```
Cjkjvfnby: `use_sys, include_outposts=True` Please call arguments in same way as them declared - > `sys_id=use_sys, include_outposts=True`  here and in other places.

default/python/AI/ProductionAI.py
```diff
@@ -665,7 +665,7 @@ def generate_production_orders():
     building_name = "BLD_SOL_ORB_GEN"
     if empire.buildingTypeAvailable(building_name) and foAI.foAIstate.character.may_build_building(building_name):
         already_got_one = 99
-        for pid in list(AIstate.popCtrIDs) + list(AIstate.outpostIDs):
+        for pid in list(state.get_inhabited_planets()) + list(AIstate.outpostIDs):
```
Cjkjvfnby: https://docs.python.org/2/library/itertools.html#itertools.chain ?

default/python/AI/turn_state.py
```diff
@@ -45,6 +50,51 @@ def __update_planets(self):
...
+        """Get the supply level of a system.
+
+        Negative values indicate jumps away from supply.
+
+        :type sys_id: int
+        :return: Supply value of a system or -99 if system is not connected
+        :rtype: int
+        """
+        retval = self.__system_supply.get(sys_id, -99)
+        if retval == -99:
```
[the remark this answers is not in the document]
Morlic-fo: fixed.

default/python/AI/freeorion_tools/_freeorion_tools.py
Morlic-fo: Mapping is used to provide dict-like utilities such as ```__contains__```, ```keys```, ```values```, ```items```, ```get```. It also allows copying like ```dict(ReadOnlyDict)```.

default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -218,3 +219,54 @@ def get_partial_visibility_turn(obj_id):
...
+      Example usage:
+      my_dict = ReadOnlyDict({1:2, 3:4})
+      print my_dict[1]
+      for k in my_dict:
+          print my_dict.get(k, -1)
+      for k in my_dict.keys():
+          print my_dict[k]
+      for k, v in my_dict.iteritems():
+          print k, v
+      my_dict['a'].append(3)  # this actually modifies the dict!
```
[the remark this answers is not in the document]
Morlic-fo: Fixed. Forgot to modify this line of the docstring after adding the hash-check.

default/python/AI/ProductionAI.py
Morlic-fo: I understand where you are coming from but I don't agree with you here.

The ```sys_id=``` adds no information to the call as the meaning of the parameter is pretty clear from context. Additionally, it makes the code less readable because it isn't unusual to have some local ```sys_id``` variable that is not supposed to be passed here:
```
sys_id = 1
use_sys = 2
state.get_empire_planets_by_system(sys_id=use_sys)
state.get_empire_planets_by_system(use_sys)
```
Second call is much clearer for me and makes navigating the code easier as well.

I am open to change it if you have a good argument for your case.

default/python/AI/freeorion_tools/_freeorion_tools.py
Cjkjvfnby: > Mapping is used to provide dict-like utilities such as __contains__

You have own `__contains__`. 

> It also allows copying like dict(ReadOnlyDict).

No, you don't use `Mapping.__iter__`


Mapping docs say:
```
"""A Mapping is a generic container for associating key/value
    pairs.

    This class provides concrete generic implementations of all
    methods except for __getitem__, __iter__, and __len__.
"""
```
And I see much more methods redefined.

default/python/AI/ProductionAI.py
Cjkjvfnby: One of them looks like error:
```
state.get_empire_planets_by_system(use_sys)
state.get_empire_planets_by_system()
```

default/python/AI/freeorion_tools/_freeorion_tools.py
Dilvish-fo: > You have own __contains__.

Morlic may not have chosen the best examples to mention, but Mapping does still provide other methods that seem worthwhile.  

> And I see much more methods redefined.

__contains__ is the only one I noticed.   @Morlic-fo  I'll admit I'm left wondering why it is redefined-- is there a good reason, to put into a comment, or did you just get a bit carried away?

Dilvish-fo (commented):
Seems it may be worthwhile to have a small bit more discussion on the Mapping issue, but otherwise looks good to me.

default/python/AI/freeorion_tools/_freeorion_tools.py
Morlic-fo: You are right that ```__contains__``` is already defined in mapping, no reason to redefine. ```__nonzero__``` was redundant as well as python standards dictate that, if no ```__nonzero__``` function is defined, ```__len__``` is called instead.
I removed those function definitions.

Cjkjvfnby (approved):

Dilvish-fo:
@Morlic-fo  ~~As I had indicated in my message a few hours ago about the resourceAI bug, I was still doing some playtesting of this, which I wanted to get done before approving it for merger.  I hope you have tested it thoroughly enough to have confidence both about bugs that would have gotten reported/logged, but also that might slip by merely as a hinderance to the AI.~~ 
nevermind, oops

Dilvish-fo:
bah, nevermind that last comment of mine, I'll continue my testing, somehow I thought this had already been merged just now, though now I see it has not and is still open.

Dilvish-fo (approved):
OK, looks fine and seems to play fine for me.  Looks like there is now a conflict around some print statements that already got changed to logger in master, but once that is resolved  this looks good to merge.

PR #1738, opened 2017-09-09
Convert savefile previews to use aynchronous client messaging.

LGM-Doyle:
This PR converts retrieving save file preview from the server from using the synchronous client messaging to asynchronous.

The problems are the same as the problems with synchronous requests of object IDs, client glitches up to and including crashes.

PR #1739, opened 2017-09-09
Remove client network synchronous messaging

LGM-Doyle:
This PR should follow #1738 which removes the last use of client synchronous messaging.

Synchronous messaging is broken in ways that can not be fixed.

If the client makes a synchronous request and the server FSM is in a
state that will not respond, the client hangs.

The implementation assumes that the client is single threaded.  If the
client is not single threaded and more than one thread makes a
synchronous request and the wrong thread receives any server response,
the client hangs.

If the server attempts to change the client FSM to a new state (like
shutdown) after a synchronous request is made, the client will ignore
the request.  Even if the server satisfies the synchronous request and
then sends the state change message, the client may send another
synchronous message before it handles the state change message.  Both
client and server hang.

The solution is to use TCP as an asynchronous protocol.  It works well 😉

### DIFF
--- a/network/ClientNetworking.cpp
+++ b/network/ClientNetworking.cpp
@@ -191,10 +191,6 @@ public:
         Remove the message from the incoming message queue. */
     boost::optional<Message> GetMessage();
 
-    /** Sends \a message to the server, then blocks until it sees the first
-        synchronous response from the server. */
-    boost::optional<Message> SendSynchronousMessage(const Message& message);
-
     /** Disconnects the client from the server. */
     void DisconnectFromServer();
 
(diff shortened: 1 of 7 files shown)


geoffthemedio:
So does the save dialog now (with this applied) initially appear without contents, and then update itself when the message arrives? And if it never arrives, or takes too long, the player can just cancel the dialog?

LGM-Doyle:
Yes.

Although it is the previous PR #1738 that implements that functionality.

This PR is the single commit removing the now unused synchronous client networking interface.

geoffthemedio:
rebase?

LGM-Doyle:
Rebased.

PR #1742, opened 2017-09-11
Fix bug that prevents the AI from using growth focus

Morlic-fo:
The AI currently considers setting the growth focus only for planets that are already at growth focus. This PR allows it to use growth focus correctly for growth specials.


Dilvish-fo (approved):
good catch; indentation bugs can be tricky to notice.

PR #1743, opened 2017-09-12
Fix ResourceAI bugs regarding species without research focus

Morlic-fo:
The first two commits fix an issue for species that can not set research focus.´

The third commit is unlikely to have any real impact but offers recover from potential mistakes in the preceeding code (for example if a focus could not be set but is assumed to have been).

### DIFF
--- a/default/python/AI/ResourcesAI.py
+++ b/default/python/AI/ResourcesAI.py
@@ -528,7 +528,7 @@ def set_planet_industry_and_research_foci(focus_manager, priority_ratio):
         cumulative_pp += i_pp
         cumulative_rp += i_rp
         if RESEARCH not in pinfo.planet.availableFoci:
-            if focus_manager.bake_future_focus(pid, pinfo.current_focus, False):
+            if focus_manager.bake_future_focus(pid, INDUSTRY, False):
                 target_pp += i_pp
                 target_rp += i_rp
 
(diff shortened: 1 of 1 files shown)


Dilvish-fo (commented):
The first two commits jointly look good, yes.  The third one leaves me puzzled because I am not seeing a path by which ratios could be missing a pid that is in focus_manager.raw_planet_info.  Did you actually observe such a  thing happening?  Your PR description and commit message seem to suggest that this is a bugfix, but perhaps you really just meant this one as a code simplification?   As a simplification, it looks OK to me.

As a bugfix, I guess I'd prefer to know how the mismatch was occurring, and fix that, rather than just necessarily setting that planet to Industry focus.

Dilvish-fo:
And now I am realizing that apparently I had glossed over the final sentence of your PR description, regarding the third commit.   I'm still not seeing how bake_future_focus could wind up being mistaken about setting a focus, but I'm game to simply accept the change as a simplification.

Dilvish-fo (approved):

Dilvish-fo:
@Morlic-fo gah, lesson for us both-- I was doing a test run on your more involved turn_state PR, and discovered I should have done a bigger one on this also,  [a loop error from this third commit](https://github.com/freeorion/freeorion/commit/b3dc5f5018acb64a478afe437f34dd1bb4c74cbc) takes a while to reveal itself.

Morlic-fo:
Thanks for catching and fixing the error I introduced. Sorry for that.

PR #1746, opened 2017-09-12
Switch some logging level from warn to error

Morlic-fo:
Follow-up for #1705 post-merge discussion.


Dilvish-fo (approved):

LGM-Doyle (approved):
Reviewed and approved in a timely manner.

PR #1748, opened 2017-09-13
Let the AI surrender if it has no more populated planets

Morlic-fo:
The AI basically lost if is has no more populated planets. There are no comeback mechanics that make it worth trying to play on. The AI will achieve nothing at that point.

Surrendering for now means offering humans peace/alliance (and accepting any such proposal) and from then on basically just passing the turn. This avoids all these random and often hard to track down errors and chat message spam that occurs if some AI code relies on the AI having some planets.

It may also give the human player a somewhat satisfying end so that he does not necessarily feel pushed to clean up all the remaining AI ships just to see the win SitRep (which of course still is possible).

For easier testing, best use ingame console:
```
start 2
ai.surrender()
```

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -994,6 +994,7 @@ class AIstate(object):
 
     def surrender(self):
         chat_human("Well played. I surrender.")
+        chat_human("NOTE: The AI will accept any peace or alliance proposal at this point.")
         self.surrendered = True
         my_empire_id = fo.empireID()
         for empire_id in fo.allEmpireIDs():
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -13,7 +13,7 @@ from EnumsAI import MissionType, ShipRoleType
 import CombatRatingsAI
 import MilitaryAI
 import PlanetUtilsAI
-from freeorion_tools import dict_from_map, get_partial_visibility_turn
+from freeorion_tools import dict_from_map, get_partial_visibility_turn, chat_human
 from universe_object import System
 from AIDependencies import INVALID_ID
 from character.character_module import create_character, Aggression
@@ -87,6 +87,10 @@ def convert_to_version(state, version):
         for var_name in ['visInteriorSystemIDs', 'exploredSystemIDs', 'visBorderSystemIDs', 'unexploredSystemIDs']:
             state[var_name] = set(state.get(var_name, {}).keys())
 
+    elif version == 3:
+        # added functionality for AI to surrender
+        state['surrendered'] = False
+
     #   state["some_new_member"] = some_default_value
     #   del state["some_removed_member"]
     #   state["list_changed_to_set"] = set(state["list_changed_to_set"])
@@ -105,7 +109,7 @@ class AIstate(object):
     is playable with this AIstate version, i.e. new members must be added
     and outdated members must be modified and / or deleted.
     """
-    version = 2
+    version = 3
 
     def __init__(self, aggression):
         # Do not allow to create AIstate instances with an invalid version number.
@@ -169,6 +173,7 @@ class AIstate(object):
         self.__empire_standard_enemy = CombatRatingsAI.default_ship_stats().get_stats(hashable=True)
         self.empire_standard_enemy_rating = 0  # TODO: track on a per-empire basis
         self.character = create_character(aggression, self.empireID)
+        self.surrendered = False
 
     def __setstate__(self, state):
         try:
@@ -986,3 +991,7 @@ class AIstate(object):
 
     def get_standard_enemy(self):
         return CombatRatingsAI.ShipCombatStats(stats=self.__empire_standard_enemy)
+
+    def surrender(self):
+        chat_human("Well played. I surrender.")
+        self.surrendered = True
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -995,3 +995,21 @@ class AIstate(object):
     def surrender(self):
         chat_human("Well played. I surrender.")
         self.surrendered = True
+        my_empire_id = fo.empireID()
+        for empire_id in fo.allEmpireIDs():
+            if empire_id == my_empire_id or fo.playerIsAI(empire_id):
+                continue
+            try:
+                msg = fo.diplomaticMessage(my_empire_id, empire_id,
+                                           fo.diplomaticMessageType.peaceProposal)
+                debug("Sending peace proposal")
+                fo.sendDiplomaticMessage(msg)
+            except:
+                pass
+            try:
+                msg = fo.diplomaticMessage(my_empire_id, empire_id,
+                                           fo.diplomaticMessageType.alliesProposal)
+                debug("Sending allies proposal")
+                fo.sendDiplomaticMessage(msg)
+            except:
+                pass
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -257,6 +257,21 @@ def generateOrders():  # pylint: disable=invalid-name
             error("Exception %s while trying doneTurn() on eliminated empire" % e, exc_info=True)
         return
 
+    universe = fo.getUniverse()
+    if not any(planet and planet.speciesName and planet.ownedBy(empire.empireID)
+               for planet in map(universe.getPlanet, universe.planetIDs)):
+        debug("This empire has no colonies left. Surrendering.")
+        if not foAIstate.surrendered:
+            foAIstate.surrender()
+
+    if foAIstate.surrendered:
+        try:
+            debug("The AI has surrendered. Aborting order calculation.")
+            fo.doneTurn()
+        except Exception as e:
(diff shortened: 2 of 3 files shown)


Morlic-fo:
There should probably be some stringtable for the chat messages but I wanted to get the general idea out first.

default/python/AI/AIstate.py
```diff
@@ -87,6 +87,10 @@ def convert_to_version(state, version):
         for var_name in ['visInteriorSystemIDs', 'exploredSystemIDs', 'visBorderSystemIDs', 'unexploredSystemIDs']:
             state[var_name] = set(state.get(var_name, {}).keys())
 
+    elif version == 3:
```
Cjkjvfnby, line 90: Make a function `state.should_surrender` at the and call it each turn.   It will not hit performance, but will make code more clear.

default/python/AI/AIstate.py
```diff
@@ -986,3 +991,26 @@ def log_war_declaration(self, initiating_empire_id, recipient_empire_id):
...
+        my_empire_id = fo.empireID()
+        for empire_id in fo.allEmpireIDs():
+            if empire_id == my_empire_id or fo.playerIsAI(empire_id):
+                continue
+            try:
+                msg = fo.diplomaticMessage(my_empire_id, empire_id,
+                                           fo.diplomaticMessageType.peaceProposal)
+                debug("Sending peace proposal")
+                fo.sendDiplomaticMessage(msg)
+            except:
```
Cjkjvfnby, line 1008: Bare except.   May be let it fail? It will be handled later.

Cjkjvfnby (changes requested):

default/python/AI/AIstate.py
Morlic-fo: I am not sure I understand what you want to say or how it is related to this particular part of the code. Maybe you can clarify what you mean?

As far as this part of the code is concerned:
The AI should keep track of the fact that it surrendered in savegames. Thus this information needs to be stored in the AIstate instance which it is with the boolean ```AIstate.surrendered```. This part of the code is supposed to provide backwards compatibility with previous savegames.

Cjkjvfnby: > The AI should keep track of the fact that it surrendered in savegames. 

I suggest to calculate it each game session and don't change AIState.

Dilvish-fo: > I suggest to calculate it each game session and don't change AIState.

if you don't save it in AIstate and instead just keep it in some session-specific variable, then it will spam out its surrender messages and peace/alliance proposals anytime you load a game where one or more AIs have previously surrendered.  I don't think the player really needs a reminder of the surrender, it would just seem a little broken.  

Why do you *not* want to keep track that the AI has surrendered?

geoffthemedio:
Does / should anything extra happen in a multiplayer game? The AI just accepts alliances with any human player, presumably? Could have some unexpected consequences if those humans aren't also allied and if there are other game mechanics that change when in an allied state... eg. tech sharing, visibility info sharing...

o01eg:
@geoffthemedio In the teamsmac edition alliances with AI are forbidden.

default/python/AI/AIstate.py
Cjkjvfnby: > if you don't save it in AIstate and instead just keep it in some session-specific variable, then it will spam out its surrender messages and peace/alliance proposals anytime you load a game where one or more AIs have previously surrendered. I don't think the player really needs a reminder of the surrender, it would just seem a little broken.

Good point.  

> Why do you not want to keep track that the AI has surrendered?

I want to keep AI state as slim as is possible. 

PS. Can AI just kill all his fleets and outposts and go to terminated stage?

default/python/AI/FreeOrionAI.py
```diff
@@ -257,6 +257,21 @@ def generateOrders():  # pylint: disable=invalid-name
             error("Exception %s while trying doneTurn() on eliminated empire" % e, exc_info=True)
         return
 
+    universe = fo.getUniverse()
+    if not any(planet and planet.speciesName and planet.ownedBy(empire.empireID)
+               for planet in map(universe.getPlanet, universe.planetIDs)):
+        debug("This empire has no colonies left. Surrendering.")
+        if not foAIstate.surrendered:
+            foAIstate.surrender()
```
Cjkjvfnby, line 265: What about colony ships?

Morlic-fo: Could in principle be added but I think the AI would already be so weak at that point, that it wouldn't matter in terms of posing a challenge.

Dragging the game out by having it surrender too late can make the last part of the game feel like a chore and rather unfun. Making the AI surrender too quickly after the first losses, however, is unsatisfying for the player as well.

Considering only the colonized planets as surrender condition feels like a good starting point with trivial implementation compared to keeping track of how quickly it loses planets etc and trying to deduce from that if it should surrender) which can be adjusted in the future after some playtesting feedback indicates that this is too early/late.

default/python/AI/AIstate.py
Morlic-fo: > Can AI just kill all his fleets and outposts and go to terminated stage?

I don't think this is actually desirable from a game design perspective:

For fleets, this potentially robs the player of the chance to engage into that last fleet battle that they want.

For outposts, there are more severe implications for the ongoing game: The player may feel forced to conquer the outposts before the last colony in order to "optimize" because troops may be cheaper or more readily available than outposters.
It also impacts other players that may currently be in the process of invading such outposts which suddenly disappear.

MatGB:
Couldn't it just gift remaining assets to the conquering player? 

I think I'd prefer that than a rump ally with unusable outposts. 

Regardless, something of this nature definitely needed, the cleanup phase of a conquest is always dull.

Dilvish-fo:
>> Can AI just kill all his fleets and outposts and go to terminated stage?

>For fleets, this potentially robs the player of the chance to engage into that last fleet battle that they want.

I think the scenarios in which the AI lost all of its colonies but still had some significant fleet that the human player really wanted to battle against are so rare that we can just ignore them.  (Or else just don't have the AI surrender if they have a 'strong' fleet, and figure out other ways to deal with the code situations you hoped to simplify by allowing surrender).

> Couldn't it just gift remaining assets to the conquering player?  I think I'd prefer that than a rump ally with unusable outposts.

I agree with MatGB on this one, seems it should be possible.

Dilvish-fo (commented):
And just to clarify/expand a bit more on the gifting idea, there are at least a few ways I can think of for the AI to decide which player to decide which player to surrender to.  My preference would probably be that when the AI realizes it lost its last colony it checks its sitreps to see which players conquered colonies from it that last turn and then if there were multiple it would choose among them (if it had a Capital colony that was conquered that last turn then whoever got it should probable be the chosen one, otherwise it could be a random choice or eventually based on influence).

Other possible ways could be the the AI to keep track of its major colony (independent of the empire.CapitalID property since that depends on a palace) and then surrender to whoever took that.  Or the selection could be based on influence or even totally random regardless of whether that empire even captured a colony from the surrendering one could be OK.

Vezzra:
IMO surrendering to the enemy that conquered the last colony is somewhat problematic. This might, in situations where several empires conquer another one, encourage tactics where everyone tries to isolate one small colony within the portion they conquered, and only finally conquers it once they think all the other empires have wiped out the rest.

Leading to a situation where everyone waits on everyone else. Doesn't sound like fun.

The best approach IMO is the one taking influence into account. However, until that's implemented, how about surrendering to the enemy that took the most colonies (maybe within the last 10-20 turns)?

Vezzra:
@freeorion/ai-team, status report on this?

Morlic-fo:
Should be discussed in more details on the forum and then implemented more thoroughly with server-side support for surrendering, I suppose.

Dilvish-fo:
Let's make it a point to wrap this up and get it in for 0.4.8. : [Forum Discussion](http://www.freeorion.org/forum/viewtopic.php?f=2&t=10890)

p.s., let's not try to take on server-side support for surrendering, I think, at least not now.  Let's just stick with gifting the assets for now.

Dilvish-fo:
Apparently the forum discussion probably didn't reach a conclusion yet, but I can't check what the actual status is, nor try to advance the discussion, since the forums are not working for me.  I didn't think they are working for anyone yet, are they?

Vezzra:
Forums are still down. @geoffthemedio told me that Tyreth is currently working on fixing/upgrading the forum.

geoffthemedio:
If you just want to view the boards, you can at http://freeorion.org/forum321/

Posting doesn't work, though.

Dilvish-fo:
> If you just want to view the boards, you can at http://freeorion.org/forum321/

Ah, thanks, that is at least a bit of help.  And it seems that the thread and post numbers are the same, so it works fine to take a forum thread/post address, like the discussion link towards the top here: ```http://www.freeorion.org/forum/viewtopic.php?f=2&t=10890``` and just swap in ```forum321``` in places of ```forum```

so to review the forum discussion for this PR, we can go to http://www.freeorion.org/forum321/viewtopic.php?f=2&t=10890

PR #1755, opened 2017-09-17
Add game rule to enable/disable Super Tester Takeover.

o01eg:


dbenage-cx:
Works as intended.

@freeorion/developers  I failed to find mention of using [col_bld_gen.py](https://github.com/freeorion/freeorion/blob/master/default/scripting/buildings/colonies/col_bld_gen.py) and posting generated results instead of editing these files directly.
If that is still the policy, we should probably add a comment for generation to these files.

o01eg:
Added fix for `col_bld_gen.py`

PR #1757, opened 2017-09-19
Add policy comment to colony building definitions

dbenage-cx:
Inserts a comment at top of colony building definitions to remind of policy to update script.
Suggestions welcome for alternative to `Upstream Changes`, since that may not always be accurate.

Regenerated definitions catches an unrelated change for Replicon colony buildings (building cost modifier).

### DIFF
--- a/default/scripting/buildings/colonies/SP_ABADDONI.focs.txt
+++ b/default/scripting/buildings/colonies/SP_ABADDONI.focs.txt
@@ -1,3 +1,5 @@
+// For long term changes - Do not modify this definition directly
+//                     Instead modify and execute col_bld_gen.py and use the result.
 BuildingType
     name = "BLD_COL_ABADDONI"
     description = "BLD_COL_ABADDONI_DESC"
--- a/default/scripting/buildings/colonies/SP_BANFORO.focs.txt
+++ b/default/scripting/buildings/colonies/SP_BANFORO.focs.txt
@@ -1,3 +1,5 @@
+// For long term changes - Do not modify this definition directly
+//                     Instead modify and execute col_bld_gen.py and use the result.
 BuildingType
     name = "BLD_COL_BANFORO"
     description = "BLD_COL_BANFORO_DESC"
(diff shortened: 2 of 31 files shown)


Dilvish-fo:
It's a good idea overall, instead of "For Upstream Changes" perhaps simply drop the phrase entirely and simply let the comment lead with "Do not modify..."

dbenage-cx:
Maybe `Note: When creating a Pull Request`... ?
Since the script is not used by the game, would not want to scare off some user from experimenting with local changes.

Dilvish-fo:
> Maybe  ```When creating a Pull Request``` ... would not want to scare off some user from experimenting

Good point, though the issue seems not really specific to pull requests to me (though admittedly that's the most important time for the reminder).  How about either ```For permanent changes ...``` or ```For long term changes ...```

dbenage-cx:
Of the two, prefer `long term changes`, but will leave unaltered for a few days to see if any other opinions surface ;)

o01eg:
What if test in CI that those script files same as generated by `col_bld_gen.py`script?

Vezzra:
Maybe "Non-temporary changes"?

Vezzra (approved):

Vezzra:
Btw, these are definitely changes to content scripts, not art/music/text assets. Changed tag accordingly.

dbenage-cx:
Changed to `For Pull Requests` with the understanding it implies changes to this repository, even if not in the form of a PR.
With permanent/long-term/non-temporary a user could infer the game utilizes col_bld_gen.py

@o01eg Suppose some pre-build step could be done with a webhook to start full build of both travis and appveyor.  The comment added here could still serve as a useful reminder.

geoffthemedio:
Should code comments refer to a particular version control system term (eg. "pull requests") ?

Vezzra:
Which brings us back to "For upstream changes" (which should be a VCS independent term)?

geoffthemedio:
"For upstream changes" seems overly jargony.

dbenage-cx:
Could use a more verbose form, unless there is a term other than `Upstream` suitable to replace uri.
```
// When making changes for submission to https://github.com/freeorion/freeorion:master:
// Do not modify this definition directly, instead modify and execute col_bld_gen.py and use the result.
```

Dilvish-fo:
> With permanent/long-term/non-temporary a user could infer the game utilizes col_bld_gen.py

Was this the only objection to "Permanent" or "Long term"?  It seems this concern could well be dealt with by adding some directions for use at the start of col_bld_gen.py (which frankly probably should be there anyways), and it seems to me that either of these terms much better avoids the awkwardness  that it seems the other phrasings keep stumbling into.

dbenage-cx:
> Was this the only objection to "Permanent" or "Long term"?

Roughly yes, I failed to see how changing the script results in a longer term solution unless it is committed.  That said, if others are in agreement one is a better term, will change accordingly.

I'll look at adding instructions to script unless a proposal is already under-way. sp_desc_name (in species_list) looks unused, and might be good to have default values as a variable.

Searching for other references to this file, I noticed I used `For official changes` in wiki.

Dilvish-fo:
> Roughly yes, I failed to see how changing the script results in a longer term solution unless it is committed... Searching for other references to this file, I noticed I used For official changes in wiki.

To me, the "official changes" / commitment-to-repository aspect is secondary to the main point, which is instruction on the organization of our materials and the best practices for changing them.  For someone just looking at a colony building file without the advisory there would not be an obvious indicator that long term changes should be made elsewhere, hence the advisory comment.  The fact that we would require those best practices be followed in order to accept a change into the repository seems largely implicit or at least sufficiently suggested by the advisory, and trying to explicitly address the repository issue in the advisory seems to just add extra details to deal with or dance around, like "which repository" or "what commitment process".

dbenage-cx:
Added slight rework of colony script with some general comments
Split generated results to separate commit

geoffthemedio:
Looks better

PR #1758, opened 2017-09-20
Improve MapWnd::PlotFleetMovement() performance.

LGM-Doyle:
This is a collection of commits improving the performance of `MapWnd::PlotFleetMovement()`.  Performance improved from 10 ms to 2.5 ms for a specific test case.

The improvements are:
- don't delete the last path member and append an empty path.  This is a bug fix.
- stop processing `StopFleetExploring(fleet)` if fleet was not exploring
- In `Fleet::HasArmed/FighterShips()` check each ship for arms/fighters before finding the next ship in the objects map.  Previously, this found all the ships first, which is a waste if the first ship has arms/fighters.
- exit `Fleet::BlockadeAtSystem` as early as possible.
- fix a bug in `Fleet::SetRoute()` where `m_next_system` could end up being a null reference in a one hop path.
- refactor `GG::Wnd::DeleteChildren()` to delete `m_children` using `clear()` instead of `erase()` one at a time.

### DIFF
--- a/GG/GG/Wnd.h
+++ b/GG/GG/Wnd.h
@@ -1015,6 +1015,9 @@ private:
     virtual void BeginNonclientClippingImpl();
     virtual void EndNonclientClippingImpl();
 
+    /** Code common to DetachChild and DetachChildren. */
+    void DetachChildCore(Wnd* wnd);
+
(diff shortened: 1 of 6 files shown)


geoffthemedio:
After very minimal testing, no immediately obvious problems. No code objections.

PR #1760, opened 2017-09-21
Fix bug #963 two mouse pointers shown when focus is lost.

### LINKED ISSUE #963: Ghost cursor remains when moving mouse pointer out of window
### Environment
- **FreeOrion Version**: 2016-09-09.3058052 (RC1 0.4.6)
- **Operating System**: Windows 7
- **Graphic card used**: GT520M
- **Fetched as**   \* Weekly development build
### Description

When moving the mouse out of the Freeorion window, a ghost mouse pointer keeps being drawn where the window was left:

test

Notice the two mouse pointers.

The ghost pointer remains until the window loses focus or the mouse is moved into the window again. If the mouse pointer is moved through the window while out of focus, there will be a ghost pointer left behind again.
### Expected Result

Mouse pointer isn't drawn when mouse pointer is out of the window.
### Steps to reproduce
- Start Freeorion in Windowed mode
- Move Mouse pointer out of game screen

LGM-Doyle:
This PR fixes #963.

It only renders the mouse when SDL says that the app has focus.

I've tested it on linux, but since the OSX and Windows windowing systems that SDL proxies are considerably different, additional testing is a good idea.

### DIFF
--- a/GG/GG/GUI.h
+++ b/GG/GG/GUI.h
@@ -468,9 +468,13 @@ protected:
     // EventPumpBase interface
     void SetFPS(double FPS);               ///< sets the FPS value based on the most recent calculation
     void SetDeltaT(unsigned int delta_t);  ///< sets the time between the most recent frame and the one before it, in ms
+
     //@}
 
     virtual void   Run() = 0;              ///< initializes GUI state, then executes main event handler/render loop (PollAndRender())
+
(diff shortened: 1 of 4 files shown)


geoffthemedio:
works for me

Vezzra:
Brief test indicates that this works as intended on OSX.

PR #1761, opened 2017-09-21
Improve AI pathfinding

Morlic-fo:
This PR replaces the current AI pathfinding with a new pathfinding algorithm that considers (re)fuel mechanics and allows to find paths that require detours to resupply (and planned for the future around systems blocked by e.g. a stationary monster).

The core algorithm is based on A* with the ```universe.shortestPathDistance```as heuristic. However, this implementation keeps track of the fleet's fuel during pathfinding. Reached nodes are only closed for a certain fuel level at that node, i.e. longer paths that leave the fleet with more fuel at a system will be considered in pathfinding. This is how detours for refueling are found.

The previous algorithms used jump distance rather than starlane distance, consider that fact when comparing results. Nowadays, the AI won't stop at each system anymore so using ```universe.shortestPathDistance``` is the prefered metric as long as fuel is taken into account.

With regard to performance:
1) The ```universe.shortestPathDistance``` as A* heuristic is exact if no detour is required. This means the shortest path will be found without traversing other branches in that case.
2) The algorithm ignores paths that would be "much" longer than the ```universe.shortestPathDistance``` to lessen the impact of worst-case scenarios (i.e. no valid path available). "Much longer" in this PR is defined as either twice as long as the ```universe.shortestPathDistance``` or at least 2000uu longer.

If it still turns out to be a performance hazard in some specific situations, there are some more optimizations that could be implemented (mentioned in ToDos) but performance seemed fine in my tests so far, so I didn't bother yet. As last measure, reimplementation in C++ is another option.

### DIFF
--- a/default/python/AI/pathfinding.py
+++ b/default/python/AI/pathfinding.py
@@ -90,7 +90,8 @@ def find_path_with_resupply(start, target, fleet_id):
         # which we already found to those systems
         for neighbor in universe.getImmediateNeighbors(current, empire_id):
             new_dist = path_info.distance + universe.linearDistance(current, neighbor)
-            new_fuel = fleet.maxFuel if neighbor in supplied_systems else path_info.fuel - 1
+            new_fuel = (fleet.maxFuel if (neighbor in supplied_systems or current in supplied_systems) else
+                        path_info.fuel - 1)
             if all((new_dist < dist or new_fuel > fuel) for dist, fuel, _ in path_cache.get(neighbor, [])):
                 predicted_distance = new_dist + universe.shortestPathDistance(neighbor, target)
                 if predicted_distance > max(2*shortest_possible_path_distance, shortest_possible_path_distance+5):
--- a/default/python/AI/pathfinding.py
+++ b/default/python/AI/pathfinding.py
@@ -71,7 +71,7 @@ def find_path_with_resupply(start, target, fleet_id, minimum_fuel_at_target=0):
     # supply graph suggests. For example, we could have recently refueled
     # in a system that is now blockaded by an enemy or we have found
     # the refuel special.
-    target_distance_from_supply = -min(state.get_systems_by_supply_tier(target), 0)
+    target_distance_from_supply = -min(state.get_system_supply(target), 0)
     if (fleet.maxFuel + 1 < (target_distance_from_supply + minimum_fuel_at_target) and
             universe.jumpDistance(start, target) > (start_fuel - minimum_fuel_at_target)):
         # can't possibly reach this system with the required fuel
--- a/default/python/AI/pathfinding.py
+++ b/default/python/AI/pathfinding.py
@@ -14,6 +14,28 @@ from turn_state import state
 #    - For large graphs, check if there are any must-visit nodes (e.g. only possible resupplying system),
 #      then try to find the shortest path between those and start/target.
 def find_path_with_resupply(start, target, fleet_id):
+    """Find the shortest possible path between two systems that complies with FreeOrion fuel mechanics.
+
+     If the fleet can travel the shortest possible path between start and target system, then return that path.
+     Otherwise, find the shortest possible detour including refueling.
+
+
+     The core algorithm is a modified A* with the universe.shortestPathDistance as heuristic.
+     While searching for a path, keep track of the fleet's fuel. Compared to standard A*/dijkstra,
+     nodes are locked only for a certain minimum level of fuel. If a longer path yields a higher fuel
+     level at a given system, then that path considered for pathfinding and added to the queue.
+
+
+    :param start: start system id
+    :type start: int
+    :param target:  target system id
+    :type target: int
+    :param fleet_id: fleet to find the path for
+    :type fleet_id: int
+    :return: shortest possible path including resupply-deroutes in the form of system ids
+             including both start and target system
+    :rtype: tuple[int]
+    """
     path_information = namedtuple('path_information', ['distance', 'fuel', 'path'])
     universe = fo.getUniverse()
     empire_id = fo.empireID()
--- a/default/python/AI/pathfinding.py
+++ b/default/python/AI/pathfinding.py
@@ -2,8 +2,17 @@ from heapq import heappush, heappop
 from collections import namedtuple
 
 import freeOrionAIInterface as fo
+from turn_state import state
 
 
+# TODO Allow short idling in system to refuel
+# TODO Consider additional optimizations:
+#    - Check if universe.shortestPath is valid before running pathfinder
+#      (seemingly not worth it - we have exact heuristic in that case)
+#    - Cut off branches that can't reach target due to supply distance
+#    - For large graphs, check existance on a simplified graph (use single node to represent supply clusters)
+#    - For large graphs, check if there are any must-visit nodes (e.g. only possible resupplying system),
+#      then try to find the shortest path between those and start/target.
 def find_path_with_resupply(start, target, fleet_id):
     path_information = namedtuple('path_information', ['distance', 'fuel', 'path'])
     universe = fo.getUniverse()
@@ -11,8 +20,22 @@ def find_path_with_resupply(start, target, fleet_id):
     fleet = universe.getFleet(fleet_id)
     supplied_systems = set(fo.getEmpire().fleetSupplyableSystemIDs)
 
-    # initialize data structures
     start_fuel = fleet.maxFuel if start in supplied_systems else fleet.fuel
+
+    # We have 1 free jump from supplied system into unsupplied systems.
+    # Thus, the target system must be at most maxFuel + 1 jumps away
+    # in order to reach the system under standard conditions.
+    # In some edge cases, we may have more supply here than what the
+    # supply graph suggests. For example, we could have recently refueled
+    # in a system that is now blockaded by an enemy or we have found
+    # the refuel special.
+    target_distance_from_supply = -state.get_systems_by_supply_tier(target)
+    if (fleet.maxFuel + 1 < target_distance_from_supply and
+            universe.jumpDistance(start, target) > start_fuel):
+        # can't possibly reach this system
+        return None
+
+    # initialize data structures
     path_cache = {}
     shortest_possible_path = universe.shortestPathDistance(start, target)
     queue = [(shortest_possible_path, path_information(distance=0, fuel=start_fuel, path=(start,)), start)]
--- a/default/python/AI/pathfinding.py
+++ b/default/python/AI/pathfinding.py
@@ -103,7 +103,7 @@ def find_path_with_resupply(start, target, fleet_id):
             predicted_distance = new_dist + universe.shortestPathDistance(neighbor, target)
 
             # Ignore paths that are much longer than the shortest possible path
-            if predicted_distance > max(2*shortest_possible_path_distance, shortest_possible_path_distance+5):
+            if predicted_distance > max(2*shortest_possible_path_distance, shortest_possible_path_distance+2000):
                 continue
 
             # All checks passed, consider this path for further pathfinding
--- a/default/python/AI/pathfinding.py
+++ b/default/python/AI/pathfinding.py
@@ -68,14 +68,14 @@ def find_path_with_resupply(start, target, fleet_id):
 
     while queue:
         # get next system u with path information
-        (_, path_info, u) = heappop(queue)
+        (_, path_info, current) = heappop(queue)
 
         # did we reach the target?
-        if u == target:
+        if current == target:
             return path_info
 
         # add information about how we reached here to the cache
-        path_cache.setdefault(u, []).append(path_info)
+        path_cache.setdefault(current, []).append(path_info)
 
         # check if we have enough fuel to travel to neighbors
         if path_info.fuel < 1:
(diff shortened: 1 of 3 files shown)


default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,134 @@
...
+    path_cache = {}
+    queue = []
+
+    # add starting system to queue
+    heappush(queue, (shortest_possible_path_distance, path_information(distance=0, fuel=start_fuel, path=(start,)),
+                     start))
+
+    while queue:
+        # get next system u with path information
+        (_, path_info, current) = heappop(queue)
```
Dilvish-fo: If there is no reason to ever work with empty paths here (and I don't see any), then the final item in ```path_info.path``` will always be the same as ```current``` and I would like to suggest we keep the heap structure simpler and then here say
```python
(_, path_info) = heappop(queue)
current = path_info.path[-1]
```

Although it adds an extra line right here, in several other places (wherever we construct new entries for the queue) it keeps things simpler.   And to me it seems a bit helpful that the start of the loop has this explicit reminder of the relationship between ```path_info.path``` and ```current```.

Dilvish-fo (commented):
This is looking very good.  I just thought of a couple minor-ish things that hadn't come to mind in our previous review.

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,134 @@
...
+from collections import namedtuple
+
+import freeOrionAIInterface as fo
+from turn_state import state
+
+from common.configure_logging import convenience_function_references_for_logger
+(debug, info, warn, error, fatal) = convenience_function_references_for_logger(__name__)
+
+
+# TODO Allow short idling in system to refuel
```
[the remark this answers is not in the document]
Morlic-fo: I was talking about the refueling mechanics if the fleet is stationary in an unsupplied system. I clarified that comment.

If our path reaches the target system but is e.g. 0.1 fuel off, then it could be prefered to let it wait there for one turn rather than having to take a detour. The logic to find suitable circumstances isn't trivial and needs some thought but it would be useful in (various) edge cases especially for those living organic ships.

default/python/AI/pathfinding.py
Morlic-fo: Good call. I changed it.

Dilvish-fo:
I've been holding off from merging this to give CJ a chance to review this final version, but he had already given comments on most of it before, and so unless you have some plans to add more for review, @Morlic-fo , I'd suggest it's waited long enough and we go ahead and merge.

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+        return None
+
+    if fleet.maxFuel < minimum_fuel_at_target:
+        return None
+
+    start_fuel = fleet.maxFuel if start in supplied_systems else fleet.fuel
+
+    # We have 1 free jump from supplied system into unsupplied systems.
+    # Thus, the target system must be at most maxFuel + 1 jumps away
+    # in order to reach the system under standard conditions.
```
Cjkjvfnby, line 77: > Thus, the target system must be at most maxFuel + 1 jumps away in order to reach the system under standard conditions.

`standard conditions` need definition.  `maxFuel + 1 jumps `  is distance that can be reached on worst case  (we start on the edge of the our borders and go to outer space.)

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # initialize data structures
+    path_cache = {}
+    queue = []
+
+    # add starting system to queue
+    heappush(queue, (shortest_possible_path_distance, path_information(distance=0, fuel=start_fuel, path=(start,))))
+
+    while queue:
+        # get next system u with path information
```
Cjkjvfnby: `u` ???

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+        (_, path_info) = heappop(queue)
```
Cjkjvfnby: No need parenthesis here.

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+        # check if we have enough fuel to travel to neighbors
+        if path_info.fuel < 1:
+            continue
+
+        # add neighboring systems to the queue if the resulting path
+        # is either shorter or offers more fuel than the other paths
+        # which we already found to those systems
+        for neighbor in universe.getImmediateNeighbors(current, empire_id):
+            new_dist = path_info.distance + universe.linearDistance(current, neighbor)
+            new_fuel = (fleet.maxFuel if (neighbor in supplied_systems or current in supplied_systems) else
```
Cjkjvfnby, line 121: `current in supplied_systems` means `fleet.maxFuel - 1`

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+#      are very close the target system but can't reach it due to fuel constraints
+# TODO Consider additional optimizations:
+#    - Check if universe.shortestPath complies with fuel mechanics before running
+#      pathfinder (seemingly not worth it - we have exact heuristic in that case)
+#    - Cut off branches that can't reach target due to supply distance
+#    - For large graphs, check existance on a simplified graph (use single node to represent supply clusters)
+#    - For large graphs, check if there are any must-visit nodes (e.g. only possible resupplying system),
+#      then try to find the shortest path between those and start/target.
+def find_path_with_resupply(start, target, fleet_id, minimum_fuel_at_target=0):
+    """Find the shortest possible path between two systems that complies with FreeOrion fuel mechanics.
```
Cjkjvfnby: Please add comment how enemy fleets and monsters is counted during pathfinding.

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+            # more fuel. Priority queueing should ensure that all previously found paths here are shorter but check
+            # it anyway to be sure...
+            if any((dist <= new_dist and fuel >= new_fuel) for dist, fuel, _ in path_cache.get(neighbor, [])):
+                continue
+
+            # calculate the new distance prediction, i.e. the A* heuristic.
+            predicted_distance = new_dist + universe.shortestPathDistance(neighbor, target)
+
+            # Ignore paths that are much longer than the shortest possible path
+            if predicted_distance > max(2*shortest_possible_path_distance, shortest_possible_path_distance+2000):
```
Cjkjvfnby: 2000 is magic constant, pleas give it meaningful name, and may be some comments.

Cjkjvfnby (changes requested):
I have checked comments and docstrings.   

I am not common with A* need time to understand it enough to make code review for algorithm implementation.   I am a little busy with my study, so please don't wait me.

default/python/AI/pathfinding.py
Dilvish-fo: Currently this doesn't do anything about enemy fleets and monsters (nor does the function this is replacing).  I've been experimenting, though, with some augmentation for this that does take them into account, I plan to post a PR for that shortly after this gets merged.

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+    if shortest_possible_path_distance == -1:
+        warn("Requested path between disconnected systems, doing nothing.")
+        return None
+
+    # make sure the minimum fuel at target is realistic
+    if minimum_fuel_at_target < 0:
+        error("Requested negative fuel at target.")
+        return None
+
+    if fleet.maxFuel < minimum_fuel_at_target:
```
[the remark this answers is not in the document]
Morlic-fo: It is a relatively stupid query but implies no strict logic error. For example, function ```foo``` calls this function to determine whether a scout can reach the target system and then return. A query that triggers this check and the answer to it are both perfectly fine and valid in this context and definitely not an ```error```.

Should it already be caught by the caller? Probably so. That's why it is logged as a warning.

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+def find_path_with_resupply(start, target, fleet_id, minimum_fuel_at_target=0):
+    """Find the shortest possible path between two systems that complies with FreeOrion fuel mechanics.
+
+     If the fleet can travel the shortest possible path between start and target system, then return that path.
+     Otherwise, find the shortest possible detour including refueling.
+
+
+     The core algorithm is a modified A* with the universe.shortestPathDistance as heuristic.
+     While searching for a path, keep track of the fleet's fuel. Compared to standard A*/dijkstra,
+     nodes are locked only for a certain minimum level of fuel. If a longer path yields a higher fuel
```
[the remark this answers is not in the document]
Dilvish-fo: The comment you quote is about the process of building the set of intermediates explored while determining the final result, which is clearly stated to be the shorted path meeting the fuel requirements.  The two statements are not in conflict.   Though I suppose some more words could be added to help clarify the distinction.

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # make sure the minimum fuel at target is realistic
+    if minimum_fuel_at_target < 0:
+        error("Requested negative fuel at target.")
+        return None
+
+    if fleet.maxFuel < minimum_fuel_at_target:
+        return None
+
+    start_fuel = fleet.maxFuel if start in supplied_systems else fleet.fuel
```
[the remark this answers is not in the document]
Morlic-fo: It avoids a potential issue and the additional code is trivial.

default/python/AI/pathfinding.py
```diff
@@ -0,0 +1,139 @@
...
+    :param start: start system id
+    :type start: int
+    :param target:  target system id
+    :type target: int
+    :param fleet_id: fleet to find the path for
+    :type fleet_id: int
+    :param minimum_fuel_at_target: optional - if specified, only accept paths that leave the
+                                   fleet with at least this much fuel left at the target system
+    :type minimum_fuel_at_target: int
+    :return: shortest possible path including resupply-deroutes in the form of system ids
```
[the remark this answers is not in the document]
Dilvish-fo: Right, Morlic, if this wasn't simply a typo slip from "resupply-sideroutes" then please consider it as a clarification.

default/python/AI/pathfinding.py
Morlic-fo: Open for discussion but for me:
```a, b = foo()``` implies ```foo()``` returns 2 values by default (i.e. an iterable of exactly 2 items)
```(a, b) = foo()``` implies ```foo()``` returns 1 value by default but that value is a tuple in this particular use case.

default/python/AI/pathfinding.py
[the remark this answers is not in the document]
Dilvish-fo: Not necessarily stupid, though it could be valid to require the fuel check be done prior to calling this function.

Suppose the AI is checking several fleets to see which one, if any, can make it to the target with the desired remaining fuel, do we necessarily want to force the AI to separately check the maxfuel levels before calling this method, or can we simply let this method handle that aspect?    

Seems to me to be fine, slightly simpler, to let it be handled here.

default/python/AI/pathfinding.py
Morlic-fo: Added a TODO how I have envisioned to handle this - an additional argument for systems that are not to be considered for pathfinding. The logic what those systems are (e.g. those blocked by monsters) is supposed to be figured out by the calling function in advance.

default/python/AI/pathfinding.py
Morlic-fo: One would intuitively think so. I initially missed this as well.

However, as Dilvish correctly pointed out and as you can easily verify ingame by moving a fleet out of supply, the first jump moving out of supply does not consume any fuel (or is immediately refueled).

default/python/AI/pathfinding.py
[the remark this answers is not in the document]
Dilvish-fo: My recollection of the code is that fuel-refilling occurs during the movement stage, prior to any meter changes.  If a system becomes newly-supplied on a given turn (whether from meter change or from a blockade being lifted that same turn), then yes, a fleet in such a supplied system be at fuel less than max (and this is consistent with my recollection from experience).  So no, I don't think that simplification works.

default/python/AI/pathfinding.py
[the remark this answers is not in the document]
Morlic-fo: I rephrased the comment. Is this clearer now or does it need more work?

default/python/AI/pathfinding.py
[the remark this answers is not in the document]
Dilvish-fo: p.s., I recently added [an Order of Turn Events section to the Scripting Tutorial](http://www.freeorion.org/index.php/FOCS_Scripting_Tutorial#Order_of_Events_During_the_Turn) to help clarify such things for everyone.

default/python/AI/pathfinding.py
Dilvish-fo: No, there is no fuel cost to depart from a supplied system.

edit: whoops, Morlic's comment didn't show up before I submitted my comment, to clarify, my "no' was to cj, not to Morlic.

Cjkjvfnby: I suggest to add comment about this into code.

default/python/AI/pathfinding.py
Cjkjvfnby: > returns 2 values by default

There is no 2 values. It is always one value  `return a, b` is `return (a, b)`.  Tuple is `iterable of exactly 2 items`  in this particular use case.

Cjkjvfnby (approved):
There is one comment left  about docstring.

Dilvish-fo (approved):

PR #1763, opened 2017-09-22
Python exposure of effect accounting 

Dilvish-fo:
This is Geoff's branch, plus a fix from me.  This allows the AI to get the more detailed meter accounting breakdowns that the human player can see in tooltips.   

A motivating application is to let the AI better figure out if a stealth effect on a planet is a permanent thing that should forbid invasion attempts (until detection improves sufficiently) or a temporary thing that could be planned around but still be expected to allow an invasion in the near term.

Since the initial commit has compile problems on Linux for me, and [apparently runtime problems on Windows for Geoff](http://www.freeorion.org/forum/viewtopic.php?p=90113#p90113), I'd propose this be squash-merged for a clean commit to master.

### DIFF
--- a/python/EnumWrapper.cpp
+++ b/python/EnumWrapper.cpp
@@ -136,6 +136,18 @@ namespace FreeOrionPython {
             .value("destroy",       CR_DESTROY)
             .value("retain",        CR_RETAIN)
         ;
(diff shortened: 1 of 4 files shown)


geoffthemedio:
If it works for you after the fixes, go ahead...

PR #1764, opened 2017-09-23
Visibility ValueRefs

geoffthemedio:
Adds a Visibility ValueRef type, including an EmpireObjectVisibility lookup. Includes parsing. Also some related Dump function tweaks.

### DIFF
--- a/default/scripting/buildings/SCRYING_SPHERE.focs.txt
+++ b/default/scripting/buildings/SCRYING_SPHERE.focs.txt
@@ -26,7 +26,7 @@ BuildingType
                     Not Unowned
                 ]
             ]
-            effects = SetVisibility visibility = Partial empire = Source.Planet.Owner
+            effects = SetVisibility empire = Source.Planet.Owner visibility = Partial
     icon = "icons/specials_huge/ancient_ruins_excavated.png"
 
 #include "/scripting/common/base_prod.macros"
--- a/default/scripting/buildings/SCRYING_SPHERE.focs.txt
+++ b/default/scripting/buildings/SCRYING_SPHERE.focs.txt
@@ -26,7 +26,9 @@ BuildingType
                     Not Unowned
                 ]
             ]
-            effects = SetVisibility empire = Source.Planet.Owner visibility = Partial
+            effects = SetVisibility empire = Source.Planet.Owner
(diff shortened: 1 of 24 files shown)


geoffthemedio:
One potential issue is that order of effects execution may be partly lost, in that only the last-executed effect will be remembered, which sets which ValueRef to evaluate later to get the final visibility for the object. Will that be a problem in practice?

Dilvish-fo:
> ...only the last-executed effect will be remembered...Will that be a problem in practice?

It seems to me it could be a problem if an effect granting max(Partial, Value) were being overwritten by one granting just max(Basic, Value), and I think such situations could be encountered with current content.  Instead of having ```EmpireObjectVisValueRefMap``` record a single entry for each target object, how about having it keep a vector of entries for each target object?

(edited to correct which visibilities map I was referring to)

geoffthemedio:
Just adding a vector of ValueRef to evaluate in order for each object would be OK, and would automatically keep track of the order (and thus effect priorities) of effects applied to each object. But is it expected that there would be cases where the Visibility would be determined using a statistic, which needs to be evaluated at a particular time with respect to other effects setting visibility? This whole system already complicates things a bit with doing that, as the gamestate at the time the effect is run and the time when the visibilities are evaluated will be potentially different, but I think that just has to be accepted... But it could be set up to to run through all the saved ValueRefs in a particular oder, to ensure that effect priorities are respected between different objects, not just per-object. But perhaps that's unnecessary complication for now?

Dilvish-fo:
> But is it expected that there would be cases where the Visibility would be determined using a statistic, which needs to be evaluated at a particular time with respect to other effects setting visibility? 

Well, it's not something that I myself expect.  I do expect that it would be sufficient to simply note these kinds of constraints/limitations/complications in the [SetVisibility subsection](http://www.freeorion.org/index.php/FOCS_Scripting_Details#SetVisibility) I just added to the FOCS Scripting Details wiki page.  For now I also added a link back to here.  Please double check it to see I got the new parsing description right.

> But perhaps that's unnecessary complication for now?

Seems so to me.

geoffthemedio:
Storing and evaluating the vector of ValueRefs for visibility should be implemented now. Quick test with some simple visibility-granting effectsgroups on humans seems to work.

PR #1765, opened 2017-09-23
limit AI to add max 10 items to research queue - Speedup for AI-Turn

banduri:
in turn 2 the AI adds all techs to the research queue and then never again adds anything. Boost seams to be slow on call to  fo.getEmpire().researchQueue when there are 100+ elements in the queue, so i reduced it to max 10 elements that can be added to the research queue per turn.

### DIFF
--- a/default/python/AI/ResearchAI.py
+++ b/default/python/AI/ResearchAI.py
@@ -683,7 +683,7 @@ def generate_classic_research_orders():
                 tech_base.update(missing_prereqs + [tech])
         cum_cost = 0
         print "  Enqueued Tech: %20s \t\t %8s \t %s" % ("Name", "Cost", "CumulativeCost")
-        for name in techs_to_add:
+        for name in techs_to_add[:10]:
             try:
                 enqueue_res = fo.issueEnqueueTechOrder(name, -1)
                 if enqueue_res == 1:


geoffthemedio:
Does the AI know how to handle this change? ie. will it add the remaining techs later? @Dilvish-fo @Morlic-fo does the AI ever reorder techs based on the strategic situation?

Dilvish-fo:
> Boost seams to be slow on call to fo.getEmpire().researchQueue when there are 100+ elements in the queue, so i reduced it...

Could you quantify both the "seems to be slow" and the timing improvement from your change?  

There are some complications related to your approach which would make me more inclined to try some other approaches to improving the efficiency of the ResearchAI before going this route.  They aren't necessarily big complications, but it would still be an extra task to make sure they were all identified and resolved in some fashion.   One complication that I think would definitely need to be addressed is the acceleration of Death Ray 2 that can take place if Death Ray 1 is awarded by an Ancient ruins; the code for that currently assumes, and is dependent on, DR2 already being in the queue.  A much more minor complication that it would probably be basically OK to ignore is that the ColonisationAI at one point checks to see if SHP_ASTEROID_HULLS is in the first 11 items on the research queue.   It's not that much difference if there are only 10 items on the queue, but at least we'd probably want to harmonize the Colonisation code with the ResearchAI approach.  And then there are other shiftings around that would need to be looked at also.  So I'd rather not get into chasing down such complications if there is not a compelling reason to.

I think there are some alternative ways we could consider making the ResearchAI more efficient (or at least more streamlined albeit at the cost of some responsiveness), but I'd like to know a bit more about how much of an issue it really is as things stand.

Morlic-fo:
~~Unless I am missing something, this limits the techs that are enqueued per turn to 10. It does not limit the number of entries in the research queue which would still grow by 10 per turn (minus what was completed) unless all techs are enqueued.~~

Calls to the research queue never showed up in my profiling of the AI code as taking significant amount of time. What is this supposed to fix?

banduri:
i just did the profiling again, and could not get any improvements this time, which is sad. So i will turn down that pull request since it must be some problem with my pc i could not figure out. It also tries to fix that nearly all techs are added to the research at turn 4. The logfile says something like
```
ResearchAI.py:686 -   Enqueued Tech:                 Name                   Cost        CumulativeCost
...
ResearchAI.py:696 -     Enqueued Tech:      SHP_BLACKSHIELD                 7500           52186
```
and does not add anything again till turn 100 which is when i stopped the game . I think that reduces the flexibility of the AI.

Morlic-fo:
> I think that reduces the flexibility of the AI.

Obviously hardcoded tech paths reduce flexibility. However, if you do not change the actual logic by which order techs are enqueued, simply enqueuing them one (ten) at a time does not change anything at all.

Note that the techs in the queue can always be reordered so in practice the flexibility isn't impacted negatively by enqueueing all the techs at once. It doesn't necessarily produce nice code though so if a more dynamic research approach is implemented in the future, it may very well make sense to not enqueue all techs at turn one.

PR #1766, opened 2017-09-24
Remove ill-performing AI focus "dithering"

Morlic-fo:
While the original idea may have been good and correct, nowadays the code logic/math is simply wrong and we pay much higher amount of PP than what is documented in that cryptic part of ancient code.
I didn't bother to check the math, but you can clearly see ingame that this isn't working at all as intended.

Besides wrong math, there is another problem: Temporarily dithering the focus means that other planets are indirectly pushed to change focus as well because the ratios and projected target research / production change. This implies even more focus change penalties and complete instability of the AI resource code producing all these wiggly lines in the resource output.

### DIFF
--- a/default/python/AI/ResourcesAI.py
+++ b/default/python/AI/ResourcesAI.py
@@ -535,38 +535,21 @@ def set_planet_industry_and_research_foci(focus_manager, priority_ratio):
     # smallest possible ratio of research to industry with an all industry focus
     maxi_ratio = cumulative_rp / max(0.01, cumulative_pp)
 
-    for adj_round in [2, 3, 4]:
-        for pid, pinfo in focus_manager.raw_planet_info.items():
-            ii, tr = pinfo.possible_output[INDUSTRY]
-            ri, rr = pinfo.possible_output[RESEARCH]
-            ci, cr = pinfo.current_output
-            research_penalty = (pinfo.current_focus != RESEARCH)
-            # calculate factor F at which ii + F * tr == ri + F * rr =====> F = ( ii-ri ) / (rr-tr)
-            factor = (ii - ri) / max(0.01, rr - tr)  # don't let denominator be zero for planets where focus doesn't change RP
-            planet = pinfo.planet
-            if adj_round == 2:  # take research at planets with very cheap research
-                if (maxi_ratio < priority_ratio) and (target_rp < priority_ratio * cumulative_pp) and (factor <= 1.0):
-                    target_pp += ri
-                    target_rp += rr
-                    focus_manager.bake_future_focus(pid, RESEARCH, False)
-                continue
-            if adj_round == 3:  # take research at planets where can do reasonable balance
-                if has_force or foAI.foAIstate.character.may_dither_focus_to_gain_research() or (target_rp >= priority_ratio * cumulative_pp):
-                    continue
-                pop = planet.currentMeterValue(fo.meterType.population)
-                t_pop = planet.currentMeterValue(fo.meterType.targetPopulation)
-                # if AI is aggressive+, and this planet in range where temporary Research focus can get an additional RP at cost of 1 PP, and still need some RP, then do it
-                if pop < t_pop - 5:
-                    continue
-                if (ci > ii + 8) or (((rr > ii) or ((rr - cr) >= 1 + 2 * research_penalty)) and ((rr - tr) >= 3) and ((cr - tr) >= 0.7 * ((ii - ci) * (1 + 0.1 * research_penalty)))):
-                    target_pp += ci - 1 - research_penalty
-                    target_rp += cr + 1
-                    focus_manager.bake_future_focus(pid, RESEARCH, False)
-                continue
-            if adj_round == 4:  # assume default IFocus
-                target_pp += ii  # icurTargets initially calculated by Industry focus, which will be our default focus
-                target_rp += tr
-                ratios.append((factor, pid, pinfo))
+    for pid, pinfo in focus_manager.raw_planet_info.items():
+        ii, tr = pinfo.possible_output[INDUSTRY]
+        ri, rr = pinfo.possible_output[RESEARCH]
+        # calculate factor F at which ii + F * tr == ri + F * rr =====> F = ( ii-ri ) / (rr-tr)
+        factor = (ii - ri) / max(0.01, rr - tr)
+        # take research focus if cheap, otherwise assume industry focus for now
+        if maxi_ratio < priority_ratio and target_rp < priority_ratio * cumulative_pp and factor <= 1.0:
+            target_pp += ri
+            target_rp += rr
+            print "Taking research focus at %s as it is cheap here." % pinfo.planet
+            focus_manager.bake_future_focus(pid, RESEARCH, False)
+        else:
+            target_pp += ii
+            target_rp += tr
+            ratios.append((factor, pid, pinfo))
 
     ratios.sort()
     printed_header = False
@@ -598,6 +581,7 @@ def set_planet_industry_and_research_foci(focus_manager, priority_ratio):
             print "pID (%3d) %22s | c: %5.1f / %5.1f | cT: %5.1f / %5.1f |  cF: %8s | nF: %8s | cT: %5.1f / %5.1f | %.2f" % (pid, pinfo.planet.name, c_rp, c_pp, ot_rp, ot_pp, _focus_names.get(old_focus, 'unknown'), _focus_names[RESEARCH], nt_rp, nt_pp, ratio)
             continue  # RP is getting too expensive, but might be willing to still allocate from a planet with less PP to lose
         # if focus_manager.planet_map[pid].currentMeterValue(fo.meterType.targetPopulation) >0: #only set to research if pop won't die out
+        print "Taking research focus at %s despite being costly." % pinfo.planet
         focus_manager.bake_future_focus(pid, RESEARCH, False)
         target_rp += (rr - tr)
         target_pp -= (ii - ri)
--- a/default/python/AI/character/character_module.py
+++ b/default/python/AI/character/character_module.py
@@ -230,10 +230,6 @@ class Trait(object):
         """
         return 0.0
 
-    def may_dither_focus_to_gain_research(self):  # pylint: disable=no-self-use,unused-argument
-        """Return True if permitted to trade production at a loss for research"""
-        return True
-
     def may_research_heavily(self):  # pylint: disable=no-self-use,unused-argument
         """Return true if allowed to target research/industry > 1.5"""
         return True
@@ -372,9 +368,6 @@ class Aggression(Trait):
 
     def military_safety_factor(self):
         return [4.0, 3.0, 2.0, 1.5, 1.2, 1.0][self.aggression]
-
-    def may_dither_focus_to_gain_research(self):
-        return self.aggression < fo.aggression.aggressive
 
     def may_research_heavily(self):
         return self.aggression > fo.aggression.cautious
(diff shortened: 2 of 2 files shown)


Dilvish-fo:
> Remove ill-performing AI focus "dithering"... I didn't bother to check the math, but you can clearly see ingame that this isn't working at all as intended....instability...producing all these wiggly lines in the resource output.

### Summary
While I think that the current dithering could be improved, it looks far from broken to me and my small bit of current testing, and I'm not at all convinced it should just be thrown out.  We also need to agree on the framework for making such assessments.

####  math...instability...wiggly lines in the resource output.
Although stability of output does have some have planning value, the magnitude of output seems far more important to me, especially over time, and I don't think an approach should be rejected merely because it produces a squiggle in plotted output.  I will try to clarify the calculations employed here.

#### "Ill Performing"
The limited tests I ran just now, while having somewhat mixed results, appear to me to show more support for dithering performing OK-to-well rather than "ill".   

We need some agreement on how to assess performance; my strong preference is that the primary foundation be the AI charting script (and we could make further adjustments if some more specific outputs are desired).  Of course, actual gameplay results against a human player is the paramount characteristic, but in the interests of repeatability and thorough testing I think we need to assess the results of AI vs AI play unless we have some strong reason to think that the assessment against a human would be noticeably different.  In case Marcel's last edits to the script had broken it for you as they had for me (perhaps he had tested it in ipython with the automatic pylab import option active), please note I've now pushed a fix for that.  For this assessment I used the combined PP + 2*RP chart.  

That 2:1 valuation weighting of RP versus PP is itself a key point we need to reach agreement about at least as a ballpark reference point, since the whole point of dithering is a tradeoff between PP and RP.  The current dithering calculations implicitly relied on a heavier weighting for RP in that ballpark of 2:1.  (Assuming we do stick with dithering, I can take it on to rework those calculations to explicitly reference the same RESEARCH_WEIGHTING value that is used for the Protection Focus assessment, and provide more commentary explaining the reasoning behind the decision calculations, as well as try to refactor them some to make them more clear). 

The primary metric I looked at was the cumulative PP+2*RP value provided by the charting script, but I also looked at the spot-values for the final turn.  My basic setup was 10 AIs in a 250 system galaxy, with no monsters and everything else medium, random shape.  For each test I would save a game at the very beginning to have a consistent starting point, then quit, and then reload and run for N turns, one reload/run with dithering and once without.   The auto-advance-n-turns option seems broken for me, so I out on auto-advance for most of the game until close to the desired endpoint and then manually advanced the last couple turns to ensure no overshoot.  I ran the first galaxy setup for 150 turns and the second galaxy setup for 210 turns.  

  I did that for two different starting galaxy setups (still a fairly small sample, just all I had time for right now).  In each case I compared the aggregate output for all 10 AIs, as well as the aggregate for the top 5 and bottom 5 when tiered according to the sum of their outputs with and without dithering.  (And then as noted, I eyeballed the final turn values.)

In both my test cases the overall group  of AI's did better with dithering than without, in the first case (150 turns) 5.9% better and in the second case (different RNG key, 210 turns) 11.6% better.   Looking at the AIs tiered by top-5 versus bottom-5, in both galaxies the bottom 5 did 20-25% better with dithering than without, in the 150 turn galaxy the top-5 AIs did 1.4% better *without* dithering, and in the 210 turn galaxy the top 5 AIs did 3.7% better with dithering.   

In a couple cases there were big swings on how a given AI would do in the no-dithering versus dithering (probably due the shifting of a balance point relative to a neighbor), but mostly the top-5/bottom-5 tiering stayed the same whether with dithering, without, or based on the aggregate.

In the 150 turn galaxy, with dithering the top final AI had about 15% lower PP+2RP score on turn 150 than did the top AI in the no-dithering run, but in the 210 turn galaxy that situation was reversed.

So, these results are somewhat of a mixed bag, and indicate a need for further testing.  The final assessment noted above, how well was the best performing AI doing at the very end of the test period, is probably the single most important indicator of how a human player would ultimately perceive the toughness of FO, but it's also a metric that probably requires the most number of test games to assess (and gave mixed results here).  How well the top tier of AIs in a given game have performed is probably a metric with close to the same significance and probably (being an aggregate measure) requires fewer games to reliably asses, but still had mixed results in these two test cases.  How well the body of AI's did as an aggregate seems not quite as important, but still a useful indicator of how tough FO would feel to a human player, and both of these initial tests came out in favor of dithering in that respect.  

Overall conclusion-- seems unclear, perhaps a slight indicator in favor of dithering, more testing needed.

Dilvish-fo:
One bit of followup re your concern about a cascade of focus changes-- please note that the code [already has some protection](https://github.com/freeorion/freeorion/blob/master/default/python/AI/ResourcesAI.py#L562-L563) against causing a focus change cascade, in that a planet in dithering mode doesn't add it's long term nominal target rp and pp to the target tracking tallies, but instead adds a dithering derivation from its current rp and pp.   I think that goes a long way to preventing focus change cascades, but I think we could do more along those lines as well.

Morlic-fo:
Re your testing:
The relative impact of the dithering is comparatively small especially if none of the two big but rare issues mentioned above happens (i.e. meter growth and focus switching) even within the turn where the dithering happens and for that particular planet. To get any statistical significant difference, I'd expect you would need a rather large number of sample games.

>  I can take it on to rework those calculations to explicitly reference the same RESEARCH_WEIGHTING value that is used for the Protection Focus assessment, and provide more commentary explaining the reasoning behind the decision calculations, as well as try to refactor them some to make them more clear

Please do. I can't follow where all the magic numbers come from at all in the current state.

For example:
``` # if AI is aggressive+, and this planet in range where temporary Research focus can get an additional RP at cost of 1 PP, and still need some RP, then do it```

1) Turn 0 - Planet has maxed meters at industry focus (10 PP, 2 RP)
2) Planet switches to research focus
3) Turn 1 output:  8 / 3
4) Planet switches to industry focus again
5) Turn 2 output: 9 / 2
6) Turn 3 output: 10 / 2

--> You trade 3 PP for 1 RP even in the best case scenario which implied to me the code was effectively broken.
> The current dithering calculations implicitly relied on a heavier weighting for RP in that ballpark of 2:1.
I suppose when relying on a 2:1 conversion that means you pay "1 additional PP". However, this is still a worse trade-off than the suggested 2:1 conversion.
If the max. PP meter is increased during any of the three turns, then we pay a higher price than 3 PP.

> a planet in dithering mode doesn't add it's long term nominal target rp and pp to the target tracking tallies, but instead adds a dithering derivation from its current rp and pp
The very next turn, the values change as the planet is out of dithering mode, so the very next turn the AI may get different results for a focus evaluation. Then, after meters regrew to maximum, the planet switches into dithering mode again. Arguably, this will only have impact in rare circumstances where the 2-4 RP/PP difference will move the AI over one of these thresholds https://github.com/freeorion/freeorion/blob/master/default/python/AI/ResourcesAI.py#L585


Another issue that does not make sense to me yet is that we pick planets with a 3:1 dithering ratio (see above), when we previously only selected planets with a ```factor``` of smaller than 1. A 3:1 dithering ratio would imply from my understanding, a ```factor``` of 3 (for the dithered part of the output). Wouldn't we want to use planets with lower ```factor``` first to get our research goal?
By that logic, it would make more sense to move the dithering part to the end of the code. That would also avoid the potential focus-instability issue.

Dilvish-fo:
> You trade 3 PP for 1 RP even in the best case scenario which implied to me the code was effectively broken.

Hah, ok, well, now I am really kicking myself for overlooking the mistake with the focus-change penalty handling that makes dithering always be wastefully cut short like that, you certainly have exposed a significant flaw, which makes the current dithering rarely enough beneficial, and often enough neutral-to-mildly-harmful, that I can understand you calling it ill-performing and at least somewhat broken.

Originally there was no focus change penalty, and dithering could easily get a 1:1 tradeoff.   When the focus change penalty was added I changed the the dither decision calc in a few ways, but it's hard to see now why I ever thought that change to the last the last clause (which is what is terminating dither early) would work.  **Luckily, it's easy enough to fix.**

I have a solid basic fix for it now, and am working on a more comprehensive reworking of the dither code to make it both much more clear, and also stronger in a few ways.

Since your current view of the while dithering idea seems strongly colored by the under-performance of the current code, let me explain a now-working counter-example that highlights when it can be most valuable.

Consider an Empire that only has their capital, which is Egassem.  They have a some Growth tech (PlanetEco, SubHab), some Learning tech (AlgoEssence, NascentAI) and some Production tech (Robotics, Fusion).  Their planet has stabilized at 22 pop.  With Industry focus, their I;R targets are 30.8 PP ; 7 RP and with Research Focus their I;R targets are  11 PP ; 11.95 RP.  So, for their regular focus decision based on steady-state levels, they are looking at a research cost factor of (30.8-11)/(11.95-7) = 4.0  PP per point of RP,  Very Expensive Research.  Dithering, on the other hand, lets them get a decent extra chunk of research at a final average cost of only 1.7 PP per point of RP, shown in the table below, which starts at steady-state Industry Focus and shows one complete 11-turn Dithering cycle, using the following headers:
Foc : Focus (I or R)
C I : Current Industry
C R : Current Research
Tot Delta I : running tally of Industry change (loss) relative to steady-state Industry Focus
Tot Delta R : running tally of Research change (gain) relative to steady-state Industry Focus
Avg R Cost : cumulative average cost of RP in this dithering cycle

Foc | C I  | C R | Tot Delta I | Tot Delta R | Avg R Cost
-- | -- | -- | -- | -- | --
I | 30.8 | 7.0 | 0 | 0 | -
R | 28.8 | 8.0 | -2 | 1 | 2.0
R | 27.8 | 9.0 | -5 | 3 | 1.7
R | 26.8 | 10.0 | -9 | 6 | 1.5
R | 25.8 | 11.0 | -14 | 10 | 1.4
R | 24.8 | 11.95 | -20 | 14.95 | 1.3
I | 25.8 | 9.95 | -25 | 17.9 | 1.4
I | 26.8 | 8.95 | -29 | 19.85 | 1.5
I | 27.8 | 7.95 | -32 | 20.8 | 1.5
I | 28.8 | 7.0 | -34 | 20.8 | 1.6
I | 29.8 | 7.0 | -35 | 20.8 | 1.7

And then the cycle repeats.  If we consider that using this AI's research cost factor of 4, both steady-state Industry focus and steady-state Research focus are worth a combined value of 58.8/turn, whereas dithering yields an average value of 67.3/turn, nearly 15% more.  Even if we value the RP at a more generic 2PP, the dithering approach is still more effective than either steady-state approach.

If you wonder, "Instead of shifting to Research for a short while like with dithering, how about shifting to Research for a medium long while before going back to Industry?  Even with a high valuation for RP that is actually slightly *less* productive in combined value than either of the steady-state approaches, and if you consider the more generic RP valuation of 2PP, then it is *significantly less* productive.  

This is making me think we should try to do even more than we do now to avoid unneeded focus changing.  And on that point, you expressed some dissatisfaction about the existing code to reduce focus change cascade from dithering, but like I said earlier, I have ideas on how to further improve that-- I think that by working with the calculated expected average values from dithering, rather than the dithering "spot-values", we can avoid having dithering trigger any non-dither focus changing except perhaps at the very start of a dithering (which can happen with any new focus decision).

Anyways, hopefully that all clarifies for you why I steer towards making this work better rather than throw it out.

Of course, perhaps sometimes the AI might not even want to pay 1.7 PP per RP, in my rework of this code I have some ideas on how to address that and and some other potential considerations that come to mind.

I'll try to get a PR out on this later today.

Morlic-fo:
> Anyways, hopefully that all clarifies for you why I steer towards making this work better rather than throw it out.

Sounds great. I'll keep this open until your PR is up.

Morlic-fo:
superseded by #1773

PR #1767, opened 2017-09-24
replace "planet bombarded" sitreps with "planet attacked"

TheSilentOne1:
Currently when a planet is damaged by a ship a sitrep is generated stating the planet "was bombarded". That can be confused with when actual bombardement weapons are used; so I suggest to rename the sitrep messages to "planet attacked".

### DIFF
--- a/default/stringtables/de.txt
+++ b/default/stringtables/de.txt
@@ -3329,17 +3329,17 @@ In %system%: Ein %shipdesign% wurde beschädigt.
 SITREP_UNOWNED_SHIP_DAMAGED_AT_SYSTEM_LABEL
 Fremdes Schiff beschädigt
 
-SITREP_PLANET_BOMBARDED_AT_SYSTEM
-Der Planet %planet% im System %system% vom Imperium %empire% ist bombardiert worden.
+SITREP_PLANET_ATTACKED_AT_SYSTEM
+Der Planet %planet% im System %system% vom Imperium %empire% ist angegriffen worden.
 
-SITREP_PLANET_BOMBARDED_AT_SYSTEM_LABEL
-Imperiumsplanet bombardiert
+SITREP_PLANET_ATTACKED_AT_SYSTEM_LABEL
+Imperiumsplanet angegriffen
 
-SITREP_UNOWNED_PLANET_BOMBARDED_AT_SYSTEM
-Der Planet %planet% von %system% ist bombardiert worden.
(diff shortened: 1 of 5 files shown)


MatGB:
I agree with this, had been thinking similarly. Is it necessary to change the backend? If so a find /replace in all the stringtables ought to be done.

TheSilentOne1:
Right, I've replaced all occurences in the stringtables. Ru+Fr translation need to be revised,
@Ouaz @Cjkjvfnby would you post a translation here?

Ouaz:
@TheSilentOne1 

For the FR stringtable:

5975 `Dans %system% : la planète %planet% de %empire% a été` **attaquée**.

5978 `Planète impériale`  **attaquée**

5981 `Dans %system% : la planète %planet% a été`  **attaquée**.

5984 `Planète libre` **attaquée**

Cjkjvfnby:
Added myself as review to not forget provide run translation.

Cjkjvfnby:
`UNOWNED_PLANET_ATTACKED`  is confusing. If nobody owns it why it is attacked?  May be it is planet with unknown owner?  

```
SITREP_PLANET_ATTACKED_AT_SYSTEM
В %system%: планета %planet% была атакована %empire%.  

SITREP_PLANET_ATTACKED_AT_SYSTEM_LABEL
Наша планета атакована
 
SITREP_UNOWNED_PLANET_ATTACKED_AT_SYSTEM
В %system%: планета %planet% была атакована.
 
SITREP_UNOWNED_PLANET_ATTACKED_AT_SYSTEM_LABEL
Ничья планета была атакована
```

Cjkjvfnby (approved):

MatGB:
Unowned should probably be neutral or unaligned, either is more grammatical. We decided not to have uninhabited worlds with defence stats awhileback but it wasn't final. 

I don't like unowned for an inhabited worlds, the inhabitants own it. Good catch @Cjkjvfnby

TheSilentOne1:
"Unowned" is used in many different contexts of the game and if we think about changing the term, we'll need to consider the other occurences (unowned fleet etc.) as well. "Neutral" does sound like the right term to me, though. But I think the discussion should rather be on the forums than here.
@Ouaz @Cjkjvfnby Thank you for providing the translations.

Vezzra:
> But I think the discussion should rather be on the forums than here.

Yep, please.

PR #1768, opened 2017-09-25
Battle Detection and Part Noisiness

geoffthemedio:
Adds a noisiness stat to ship parts. Adds battle scanner parts and a battle detection meter.

Work in progress.


Vezzra:
Hm, strictly speaking the tertiary stats/noisiness feature and the new battle detection meter & scanner parts are two different things (the latter requiring the former though, if I understand correctly) which could probably be broken up into two different PRs.

Regarding the battle detection meter & scanner parts, can you open a forum thread (on the design subforum) and explain your idea there? The difference to the "normal" detection mechanic isn't immediately obvious to me...

dbenage-cx:
Got excited to see how this was progressing and did a sneak peak earlier.

An odd issue is ships with troop pods, outposts, and colonies are given max pop(65k) for each part. Did not see what the exact cause was, but was a fun twist ;)

geoffthemedio:
> given max pop(65k) for each part.

Should be resolved.

LGM-Doyle:
As per @Dilvish-fo's suggestion I have rebased the Noisiness branch with the parser changes in [here](https://github.com/LGM-Doyle/freeorion/commits/Noisiness). The only changed commits were the following:
- Added enums for a tertiarty part stat (intended for weapon noisiness).
-- made a separate parser for WeaponStats
-- Fixed some rebasing issues already in the commit.  As it was the commit did not compile.
- Fix rebasing issues.
-- preserve the separate parser created in the previous change
- Fix rebase issues. (II)
-- Removed this commit, since it was empty as the rebasing issues were already fixed.
- ValueRef handling and parsing for accessing and setting tertiary part stats.
-- added the Tertiary stat
- Renamed tertiary stat to noisiness.
-- Did the rename in the new code

PR #1552, opened 2017-05-07
Add UI button to create persistent config

dbenage-cx:
Allows creation of persistent config within game.

Adds a button to the misc option page which stores any registered options that differ from their default value (excluding version string).

~~Blocked by #1580~~

PR #1769, opened 2017-09-26
Override default config options after initial adjustments

dbenage-cx:
Updates the default value in OptionsDB to reflect the resulting default after initialization.
Majority of these cases are in windows inheriting CUIWnd.

With #1552 creating a persistent config results in only two entries on my system for `app-left-windowed` and `app-top-windowed` from default config.

### DIFF
--- a/UI/CUIWnd.cpp
+++ b/UI/CUIWnd.cpp
@@ -622,7 +622,6 @@ GG::Rect CUIWnd::CalculatePosition() const
 { return GG::Rect(INVALID_X, INVALID_Y, INVALID_X, INVALID_Y); }
 
 void CUIWnd::SetDefaultedOptions() {
-    std::unordered_set<std::string> retval;
     OptionsDB& db = GetOptionsDB();
     std::set<std::string> window_options;
     db.FindOptions(window_options, "UI.windows." + m_config_name);
--- a/UI/CUIWnd.cpp
+++ b/UI/CUIWnd.cpp
@@ -165,6 +165,7 @@ void CUIWnd::CompleteConstruction() {
     GG::Wnd::CompleteConstruction();
     Init();
     ValidatePosition();
+    SetDefaultedOptions();
 }
 
 void CUIWnd::Init() {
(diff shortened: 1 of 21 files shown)


util/OptionsDB.h
```diff
@@ -309,6 +309,14 @@ class FO_COMMON_API OptionsDB {
         m_dirty |= it->second.SetFromValue(value);
     }
 
+    /** Sets the default value of option @p name to the current value */
+    void        OverrideDefault(const std::string& name) {
```
geoffthemedio: Could this have a clearer name? SetCurrentAsDefault or similar?

dbenage-cx: Snuck in on me, no longer used so will remove

geoffthemedio:
@LGM-Doyle could you comment?

UI/CUIWnd.cpp
```diff
@@ -616,6 +621,67 @@ void CUIWnd::ResetDefaultPosition() {
 GG::Rect CUIWnd::CalculatePosition() const
 { return GG::Rect(INVALID_X, INVALID_Y, INVALID_X, INVALID_Y); }
 
+void CUIWnd::SetDefaultedOptions() {
+    std::unordered_set<std::string> retval;
```
LGM-Doyle: Remove `retval` it is unused.

LGM-Doyle (commented):
@geoffthemedio This looks fine to me other than the unused `retval`.  Was there something specific that you were concerned about?

@dbenage-cx BTW the add persistent config button works well 👍

UI/CUIWnd.cpp
geoffthemedio: @LGM-Doyle You've done a bunch of work on Wnd setup and related stuff, so I thought you might have some perspective on the timing of option default setting and layout happening...

LGM-Doyle: It looks fine .  If there are non-obvious bugs they will be fixed in due course.

dbenage-cx: Removed, ty for reviewing

PR #1771, opened 2017-09-30
Ship part upgrade revamp

### LINKED ISSUE #1770: ship part auto-upgrades somewhat violate the 'combat takes place with end-of-turn meter values' rule
Ship part auto-upgrades take place via a two-step process.  When an empire has researched a part upgrade tech, then that tech will increase the part's max-capacity meter during the regular meter updating process, wherever the ship is.  The actual part capacity meter does not get increased until the next time that the ship is resupplied, which is something that takes place during the turn's movement phase, prior to combat.  The result is that one can head into combat thinking that one's enemy's ships have a certain level of firepower, but instead they get upgraded just prior to combat.  That can sometimes result in a surprise loss that had not looked possible. This is contrary to how we have generally aimed to handle the timing issues around meter updates and combat.

One possibility to deal with this could be to change this upgrade process.  I think the current system was settled on both to constrain the upgrade to Supplied space, and to allow the upgrade to take place in one swoop, rather than having the part meter value slowly climb towards the max over the span of a few turns.  An alternative could be to make them behave more like other max-meter increases, but with an increased regen rate akin to the boosted regen from Force-Energy Structures, high enough to let the increase take place in one turn.  The trickier part would be constraining it to only take place in Supplied space, perhaps we'd need to add a last_turn_resupplied attribute for ships (and a corresponding ValueRef).  And then we'd also need to figure out how to block the min 1-per-turn normal meter regen if not supplied. 

Another possibility that comes to mind is to adjust the UI for the human player to show the upgraded values when it can reasonably predict that they would be about to be upgraded.  I think it would be able to do this just fine for the majority of cases, and especially for the most important case which is when the enemy fleets are sitting still (clearly either supplied or not) and the player is making a decision about whether to attack them.  We could also, at some point, code up some similar predictions for the AI to use.

Those are the ideas that come to mind for me-- any others?

Dilvish-fo:
- works with new valuerefs to directly upgrade part capacity via
   normal meter effects, so that 'surprise' upgrades do not take place during the (pre-combat) movement phase

As discussed somewhat, in the [Fuel Parts Upgrade thread](http://www.freeorion.org/forum/viewtopic.php?p=90144#p90144) (although, due to immediate-upgrade regardless of supply, having been decided on for fuel parts, this currently is only applied to weapons).

This resolves #1770.

For a clear example of the differing result in the combat situations contemplated in #1770, a test case can be staged with a starting Frigate, without SR 1_2 having been researched, sitting in supply one short jump from a monster.  So, with standard pilots damage 3.  (On my system, 50 stars, medium everything, zero AIs, RNG key 0, comes up like this with a suitable monster a jump right then up.)  Save the game with the frigate one jump from the monster.  With current handling, if you then research SR_1_2, and on the same turn that you research it (i.e., just after you get the research sitrep, but before the change shows in the regular ship meters), you launch the ship at the monster, and then check the log, you will see that the upgrade became effective during the movement phase and the ship was attacking with strength 4.  With this PR the upgrade does not become effective until the normal meter update, after that combat round, and the log shows the ship attacking with strength 3.

(edited to clean up references to the now-abandoned idea of the Max meters being superfluous)

### DIFF
--- a/default/scripting/techs/ship_weapons/ship_weapons.macros
+++ b/default/scripting/techs/ship_weapons/ship_weapons.macros
@@ -7,6 +7,17 @@ WEAPON_BASE_EFFECTS
                 SetMaxCapacity partname = "@1@" value = Value + PartCapacity name = "@1@"
                 SetMaxSecondaryStat partname = "@1@" value = Value + PartSecondaryStat name = "@1@"
             ]
+
+// The following is really only needed on the first resupplied turn after an upgrade is researched, since the resupply currently
+// takes place in a portion of the turn before meters are updated, but currently there is no good way to restrict this to
+// only that first resupply (and it is simply mildly inefficient to repeat the topup later).
+           EffectsGroup
+            scope = And [
+                        [[EMPIRE_OWNED_SHIP_WITH_PART]]
+                        Turn high = LocalCandidate.LastTurnResupplied 
+            ]
+            accountinglabel = "@1@"
+            effects = SetCapacity partname = "@1@" value = Value + [[ARBITRARY_BIG_NUMBER_FOR_METER_TOPUP]]
 '''
 
 // @1@ part name
@@ -19,10 +30,7 @@ WEAPON_UPGRADE_CAPACITY_EFFECTS
                         [[SHIP_PART_UPGRADE_RESUPPLY_CHECK(CurrentContent)]]
             ]
             accountinglabel = "@1@"
-            effects = [
-                SetCapacity partname = "@1@" value = Value + @2@
-                SetMaxCapacity partname = "@1@" value = Value + @2@
-            ]
-'''
+            effects = SetMaxCapacity partname = "@1@" value = Value + @2@
+            '''
 
 #include "../techs.macros"
--- a/default/scripting/techs/ship_weapons/ship_weapons.macros
+++ b/default/scripting/techs/ship_weapons/ship_weapons.macros
@@ -1,11 +1,7 @@
 // @1@ part name
 WEAPON_BASE_EFFECTS
 '''        EffectsGroup
-            scope = And [
-                Ship
-                OwnedBy empire = Source.Owner
-                DesignHasPart  name = "@1@"
-            ]
+            scope = [[EMPIRE_OWNED_SHIP_WITH_PART]]
             accountinglabel = "@1@"
             effects = [
                 SetMaxCapacity partname = "@1@" value = Value + PartCapacity name = "@1@"
@@ -19,10 +15,14 @@ WEAPON_UPGRADE_CAPACITY_EFFECTS
 '''
         EffectsGroup
             scope = And [
-                Ship
-                OwnedBy empire = Source.Owner
-                DesignHasPart  name = "@1@"
+                        [[EMPIRE_OWNED_SHIP_WITH_PART]]
+                        [[SHIP_PART_UPGRADE_RESUPPLY_CHECK(CurrentContent)]]
             ]
             accountinglabel = "@1@"
-            effects = SetMaxCapacity partname = "@1@" value = Value + @2@
+            effects = [
+                SetCapacity partname = "@1@" value = Value + @2@
+                SetMaxCapacity partname = "@1@" value = Value + @2@
+            ]
 '''
+
+#include "../techs.macros"
--- a/default/scripting/techs/techs.macros
+++ b/default/scripting/techs/techs.macros
@@ -11,7 +11,8 @@ EMPIRE_OWNED_SHIP_WITH_PART
             ]
 '''
 
-// For inclusion in a scope macro, which should have already limited scope to ships owned by an empire
+// For inclusion in a scope macro for an upgrade increase to a Max type meter (MaxCapacity or MaxSecondaryStat), and that scope 
+// should have already limited scope to ships owned by an empire
 // In most cases the substitution provided for @1@ may simply be "CurrentContent" (without the quote marks) but
 // if providing an explicit tech name then include quotes with the substitution, 
 // i.e., either use
@@ -22,3 +23,6 @@ SHIP_PART_UPGRADE_RESUPPLY_CHECK
 '''
            ( LocalCandidate.LastTurnResupplied > TurnTechResearched empire = LocalCandidate.Owner name = @1@ )
 '''
+
+ARBITRARY_BIG_NUMBER_FOR_METER_TOPUP
+''' 1000000 '''
--- a/default/scripting/techs/techs.macros
+++ b/default/scripting/techs/techs.macros
@@ -1,2 +1,24 @@
 CANDIDATE_BATTLE_CHECK
 '''Turn low = LocalCandidate.System.LastTurnBattleHere + 1'''
+
+// In typical usage this is used as a plain macro, not a substitution macro; the substitution is instead done within the calling macro
+EMPIRE_OWNED_SHIP_WITH_PART
+'''
+            And [
+                Ship
+                OwnedBy empire = Source.Owner
+                DesignHasPart  name = "@1@"
+            ]
+'''
+
+// For inclusion in a scope macro, which should have already limited scope to ships owned by an empire
+// In most cases the substitution provided for @1@ may simply be "CurrentContent" (without the quote marks) but
+// if providing an explicit tech name then include quotes with the substitution, 
+// i.e., either use
+// SHIP_PART_UPGRADE_RESUPPLY_CHECK("SOME_TECH")
+// or
+// SHIP_PART_UPGRADE_RESUPPLY_CHECK(CurrentContent)
+SHIP_PART_UPGRADE_RESUPPLY_CHECK
+'''
+           ( LocalCandidate.LastTurnResupplied > TurnTechResearched empire = LocalCandidate.Owner name = @1@ )
+'''


geoffthemedio:
I don't think you should be adding `SetCapacity` or similar effects as part of upgrading at all; just gate whether to apply various `SetMaxCapacity` effects on whether a ship has been in supply. The regular resupply mechanics will kick in when the ship is in supply and do the actual capacity setting.

This is preferable because
-Capacity meters are persistent, so if you add +1 to them every turn, they'll increase by 1 each turn; they aren't reset like the max/target meters, and continue doing so even when away from supply, so that any subsequent upgrades will effectively kick in immediately, even when away from supply.
-Currently, ships could be affected by a storm or enemy part or an event of some sort that reduces their weapon capacity meters until they can get back into resupply. This would prevent that.

If you want to move all the resupply effects on these meters into FOCS, that can be done, but it would be a separate "resupply" effect, not added to each upgrade tech.

Dilvish-fo:
> I don't think you should be adding SetCapacity or similar effects as part of upgrading at all; just gate whether to apply various SetMaxCapacity effects on whether a ship has been in supply. The regular resupply mechanics will kick in when the ship is in supply and do the actual capacity setting.

Well then this wouldn't do anything to resolve #1770, it seems to me, it would just delay by a turn the situational timing that results in a 'surprise' weapon upgrade during the movement phase (immediately prior to combat), and so would make this PR pretty pointless, it seems.

When you implemented these new ValueRefs and quoted me when posting about them, were you just hoping that I (or someone) would use them to reimplement fuel upgrades to require being in Supply (but without requiring a MaxMaxFuel meter)?  I had thought that whole issue was already resolved, and I guess I read too much into your implementing these shortly after I posted #1770 .

> This is preferable because
-Capacity meters are persistent, so if you add +1 to them every turn, they'll increase by 1 each turn; they aren't reset like the max/target meters, and continue doing so even when away from supply, so that any subsequent upgrades will effectively kick in immediately, even when away from supply.

Ah, well, then that would be a reason to not take the next step I had proposed, of no longer bothering with the Max meters, but does not seem to be an objection to this PR itself.  As it stands now, with this PR, the upgrading still works fine and the Max meters keep the basic capacity meters from getting over-filled, and in my observation and thinking this PR does not cause 'any subsequent upgrades [to] effectively kick in immediately' (because none of the meters get upgraded if not in supply).

> -Currently, ships could be affected by a storm or enemy part or an event of some sort that reduces their weapon capacity meters until they can get back into resupply. This would prevent that.

Hmm, OK, well so currently they *could be*, but currently they actually *don't*, because we don't currently have any such effects, correct?  The existing storm effects like from Molecular Cloud do not require getting back into supply to recover from because they are implemented at the ship level and not at the part level, is my understanding.  I agree there would be some value to preserving the availability of this line of potential future content, and if we can come up with some other solution for #1770 then I'd agree this is enough reason to drop this PR.  If we don't come up with another solution for #1770 that we are sufficiently happy with then I think we'd need to put some more thought into whether solving that is perhaps more important than preserving the availability of persistent part capacity meter degradation  (or perhaps we explore alternative ways to accomplish the degradation, such as via Specials applied to the ship).

So, it seems to me we should let this PR sit for a now, while we brainstorm other solutions for #1770.

(edited to correct typo references to issue 1170 rather than 1770)

geoffthemedio:
I figured that if you make the minimum turn for an upgrade-related max meter increase be the turn after the last resupply for a ship, then that ship, when moving into supply, won't get an upgrade that turn, so should have the same meter values as seen on the turn before.

The resupply could still kick in in the event a ship had a lower current than max meter for reasons unrelated to upgrades, though.

To avoid that, resupply can perhaps be implemented by effects as well, instead of by C++, and appropriately gated, or perhaps added (in C++) as a separate step from Fleet::MovementPhase where is is currently, so that it happens after combat. Either of those cases might be tricky to get to work with mid-movement resupply as is currently done at each system a ship passes through during a turn's movement.

If it's preferable to not have separate tracking of max and current meters for weapon parts, that will complicate some things like fighters, where the max capacity and current capacity are really both needed, but is perhaps possible. The paired capacity meters could be changed to a single unpaired meter which is reset each turn like a max or target meter, similar to detection or stealth. I do like having the potential for effects to persistently modify ships' meters, though...

I implemented stuff that had been requested a couple times, without much thought about the details of how it would / could be used.

Dilvish-fo:
> I figured that if you make the minimum turn for an upgrade-related max meter increase be the turn after the last resupply for a ship, then that ship, when moving into supply, won't get an upgrade that turn, so should have the same meter values as seen on the turn before. The resupply could still kick in in the event a ship had a lower current than max meter for reasons unrelated to upgrades, though.

Ah, I guess that moving into supply after an upgrade has been researched is one way that this problem could be triggered, although the situations in which a fleet could both be moving into supply and on that same turn getting into combat seem more rare than the primary scenario I was concerned about.  As noted in #1770, a ship can be sitting in supply, and then on the turn that an upgrade has just been researched, move into combat (or, a ship could move into combat against an enemy ship sitting in it's supply, on the same turn that the enemy just researched an upgrade).  The upgraded maxCapacity meter will not be normally visible to the player, they will just see the non-upgraded stats when they order the move, but the upgrade will be made effective during the movement phase, just prior to combat.

> To avoid that, resupply can perhaps be implemented by effects as well, instead of by C++, and appropriately gated, or perhaps added (in C++) as a separate step from Fleet::MovementPhase where is is currently, so that it happens after combat. Either of those cases might be tricky to get to work with mid-movement resupply as is currently done at each system a ship passes through during a turn's movement.

Ack, I agree those solutions sound tricky to get to work with midmovement resupply.  Please note, the current PR "avoids that", for both sets of scenarios.

> If it's preferable to not have separate tracking of max and current meters for weapon parts...

No, no, I had simply mentioned what had appeared to me as a possible simplification of things, I don't have any special desire to remove the max meters and you have noted a number of reasons for them to continue. 

> I do like having the potential for effects to persistently modify ships' meters, though...

OK, no worries.  I have now simplified this PR, all of the out-of-supply setting of Capacity meters was not really needed for the current purpose. (I think I may have been overzealous with that as sort of preparation for the possibility of doing away with the Max meters, which idea I have now abandoned.)

This PR now both solves the surprise upgrade issue of #1770, while still allowing for out-of-supply persistent degradation of part meters as you wish (I believe).

Dilvish-fo:
Please note, I have pushed a simpler commit replacing the original.

geoffthemedio:
So does this not apply anymore?
>> I don't think you should be adding SetCapacity or similar effects as part of upgrading at all; just gate whether to apply various SetMaxCapacity effects on whether a ship has been in supply. The regular resupply mechanics will kick in when the ship is in supply and do the actual capacity setting.
>
>Well then this wouldn't do anything to resolve #1770, it seems to me, it would just delay by a turn the situational timing that results in a 'surprise' weapon upgrade during the movement phase (immediately prior to combat), and so would make this PR pretty pointless, it seems.

You seem to have done what I was suggesting...

Dilvish-fo:
> So does this not apply anymore? 
>>I don't think you should be adding SetCapacity or similar effects as part of upgrading at all...

>You seem to have done what I was suggesting...

No, notice that SetCapacity is still part of the upgrade macro, and that is important for this to work;  I simply removed it from the two base macros (i.e., the two supply-independent macros).

geoffthemedio:
> This PR now [...] still allow[s] for out-of-supply persistent degradation of part meters as you wish (I believe).

It does not, because the SetCapacity effect will be active every turn after a ship has been in supply after at least one upgrade tech for that weapon was researched. The Capacity meter is still persistent, so the effect will re-apply each turn, incrementing the Capacity up until it hits the max. That might take several tuns depending where it starts, how many upgrade effects for how much each turn apply, and what the max is, but it won't let the capacity be unchanged even when out of supply, if it's ever been in a situation to start getting the effects.

There's also no really much point in having the SetCapacity effect be as complicated as you've made it... Presumably the point is to fully upgrade whenever in supply, so you can just set the meter to a fixed arbitrary big number, eg.1000000, and let the clamping (which limits current to max) do its work. This also only needs to be done once per turn, not once per each upgrade tech.

Perhaps the point at which the turn number is incremented needs to be moved earlier in `ServerApp::PostCombatProcessTurns`, so that the production, growth, and such are the first things that occur during a turn, rather than the last. Combat would be (almost) the last thing, and then right after `// check for loss of empire capitals`, the turn number would be incremented. Then, I would expect that gating upgrade effects on the max meter, according to whether a ship has been in supply the turn before, would fix things. The problem is that meters are increased before combat unexpectedly, and this would make them increase even earlier during the turn cycle, so that the values seen by the client would still be the values used on the server for combat, but then adding a delay to the effect script would make the effect actually occur on the next turn after the combat, compared with now...

Dilvish-fo:
> It does not, because the SetCapacity effect will be active every turn after a ship has been in supply after at least one upgrade tech for that weapon was researched. 

Doh! *Sigh, right.* 

> There's also no really much point in having the SetCapacity effect be as complicated as you've made it... This also only needs to be done once per turn, not once per each upgrade tech.

OK, but it would need to be done for each individual part type, right?  So it seems that the WEAPON_BASE_EFFECTS macro would be the right place for it (for weapon parts anyways).  But not WEAPON_BASE_DEFAULT_EFFECTS since that only applies to ships that could never not have the upgrade tech.  While we think more about your idea for changing the point of turn number increase, I'll go ahead and propose a revision for this PR.

> Perhaps the point at which the turn number is incremented needs to be moved earlier...

Sounds promising.  I've added "loss of empire capitals checked" to the [Order of Events During the Turn summary](http://www.freeorion.org/index.php/FOCS_Scripting_Tutorial#Order_of_Events_During_the_Turn) in the Scripting Tutorial, to facilitate thinking this change through.  I don't currently see any prohibitive problems with the idea, though we'd probably want to double check all scripts that reference turn number to see if they might be impacted and require a corresponding tweak (likewise for the AI).  This would also harmonize the timing of these upgrades to be the same as with other types of upgrades like Growth tech max pop increases (to be same turn as the tech is researched), which would be nice and as Mat has requested in the [design thread](http://www.freeorion.org/forum/viewtopic.php?p=90175#p90175).

In the meantime, I have revised this PR so that I think it resolves the concerns you noted.   If we do proceed with the timing change you suggest considering above, the first thing in that PR could be to revert this one, but if you'd rather just let this sit while we think through the timing change possibility I don't have any real argument with that either.

Vezzra:
@Dilvish-fo, status update on this?

Dilvish-fo:
Hmm, I guess I was waiting to see if @geoffthemedio agreed that I had resolved his concerns, and then I got very busy with other matters and this slipped from attention (and I actually had it in my head that this fix had already gone in).   I'll do a little more testing with this to be sure that some other change in the past couple months hasn't interfered with it, and then if there is no further comment I'll plan to go ahead and merge it.

Vezzra:
Ok. @geoffthemedio, any more comments/input from your side?

PR #1773, opened 2017-09-30
Ai focus dithering fix

Dilvish-fo:
As discussed in #1766, AI Focus dithering (short term swaps between Research and Industry Focus), is currently underperforming because of a bug that causes the research phase of focus dithering to be cut short too soon.  In addition to fixing that bug, this PR restructures and clarifies the dither code.  I am also working on some further improvements I had mentioned in #1766, but those are taking some more time and I wanted to at least get this primary fix and clarification in for review right away.

### DIFF
--- a/default/python/AI/ResourcesAI.py
+++ b/default/python/AI/ResourcesAI.py
@@ -575,7 +575,7 @@ def set_planet_industry_and_research_foci(focus_manager, priority_ratio):
             ii, tr = pinfo.possible_output[INDUSTRY]
             ri, rr = pinfo.possible_output[RESEARCH]
             ci, cr = pinfo.current_output
-            research_penalty = (pinfo.current_focus != RESEARCH)
+            research_penalty = AIDependencies.FOCUS_CHANGE_PENALTY if (pinfo.current_focus != RESEARCH) else 0
             # calculate factor F at which ii + F * tr == ri + F * rr =====> F = ( ii-ri ) / (rr-tr)
             factor = (ii - ri) / max(0.01, rr - tr)  # don't let denominator be zero for planets where focus doesn't change RP
             planet = pinfo.planet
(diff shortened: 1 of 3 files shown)


default/python/AI/ResourcesAI.py
```diff
@@ -551,14 +553,48 @@ def set_planet_industry_and_research_foci(focus_manager, priority_ratio):
                     focus_manager.bake_future_focus(pid, RESEARCH, False)
                 continue
             if adj_round == 3:  # take research at planets where can do reasonable balance
-                if has_force or foAI.foAIstate.character.may_dither_focus_to_gain_research() or (target_rp >= priority_ratio * cumulative_pp):
+                # if this planet in range where temporary Research focus ("research dithering") can get some additional
+                # RP at a good PP cost, and still need some RP, then consider doing it.
```
Morlic-fo, line 590: I would really like an introductory comment that explains what "research dithering" actually is supposed to do: Abuse the fact that meters fall only by the same rate that they grow. So the fundamental idea is to grow to the research target to maximum while the PP meter is still above its target value at research focus target. 

Also, some explicit math wouldn't hurt. Something like
```
Assuming we switch to a research focus for n turns with a focus penalty of p and assuming integer values of Research(R) and Industry(I) only:

R = {sum(i), i=1..n}  = n(n+1)/2
I = {-sum(i+p), i=1..n} = -n(n+1+2p)/2

Growing up again until industry meter maximum, we need (n+p) turns.
R = {sum(i), i = 0..n-p-1} = (n-p-1)(n-p)/2
I = {-sum(i), i = 1..n+p-1} = (n+p-1)(n+p)/2

Adding both, we have
R = n^2 - np + p^2 / 2 + p/2
I = - (n^2 + 2np +p^2/2 -p/2)

With current p = 1 this yields
R = 1 + n(n-1)
I = - n(n+2)
or a cost ratio of -I/R = n(n+2) / (1+n(n-1))

For example, a cycle consisting of switching 5 turns to research yields (n=5)
R = 21
I = -35
or a cost of ~1.67 PP per RP
```

default/python/AI/ResourcesAI.py
```diff
@@ -551,14 +553,48 @@ def set_planet_industry_and_research_foci(focus_manager, priority_ratio):
...
+                MIN_DITHER_TARGET_RESEARCH_GAP = 3
+                if (rr - tr) < MIN_DITHER_TARGET_RESEARCH_GAP:
+                    continue
+
+                # could double check that planet even has Industry Focus available, but no harm even if not
+
+                # So at this point we have determined the planet has research targets compatible with employing focus
+                # dither.  The research focus phase will last until current research reaches the Research-Focus
+                # research target, determined by the 'research_capped' indicator, at which point the focus is
+                # changed to Industry (currently allowed to be handled by standed focus code later).  The research_caped
```
Morlic-fo: typos: ```standard```, ```capped```

Morlic-fo (changes requested):
Thanks for the clarifying comments.

A top-level comment explaining what research dithering actually is and how it works is still missing. There are some comments about it spread throughout the other comments but a clear introductory comment would be good.

Morlic-fo:
Somehow didn't get into my review message but the actual fix seems to work on first sight. In my previous test case where the AI did those 1-turn dithers, it now performs a full 5 turn cycle.

Vezzra:
@Dilvish-fo, @Morlic-fo, progress report on this? Anything preventing it from being merged?

Morlic-fo:
The actual fix works so if @Dilvish-fo rebases I am fine with merging this as is and fix the comment/style post-merge if time does not allow more thorough revision.

Needs a rebase though.

Dilvish-fo:
Heyas, I've just been very very busy for a while now, but it certainly has been on my mind to get back to this.  There was still a fair bit more I had wanted to do with it, but perhaps I'll just make a reminder for myself to cover the further improvements so that this fix can get in more promptly.  Consider me nudged  :smile:

Dilvish-fo:
OK, I've rebased this, and also addressed a few more of the suggestions (some additional/clearer comments, AIDependencies, character trait).  I still have some plans to do some more substantive work on this, but I'll see how the time goes, perhaps I'll just leave that for a later PR.

Morlic-fo (approved):

Vezzra:
> I still have some plans to do some more substantive work on this, but I'll see how the time goes, perhaps I'll just leave that for a later PR.

Seeing as this is flagged as mandatory for the release, preferrably it should get merged rather sooner than later. So, leaving more substantive work for a later PR and merging this now sounds good to me.

PR #1774, opened 2017-09-30
Change titles of top level ship design articles

### LINKED ISSUE #1553: Pedia Categories Monster Types and Ship Design Empty From Intro Screen
When starting FreeOrion, the pedia from the intro screen has Monster Types and Ship Design categories listed in the index. These are empty until a game starts and the server sends the relevant info. They should thus probably be labelled as "[Game -", like Buildings or Ships, which are populated with game-specific info.

Alternative, these articles could be populated with info read from the default scripted stuff.

dbenage-cx:
Fixes #1553 

Russian translation changed as the entries have a consistent style with default and the article link may be easily overlooked.


dbenage-cx:
This PR is pending some resolution to [discussion thread](http://freeorion.org/forum/viewtopic.php?f=6&t=10714).

geoffthemedio:
Just saw the link to the discussion thread after merging, but I think it doesn't matter, as these changes are / were worthwhile regardless of any reorganization / moving stuff out of the default index.

PR #1775, opened 2017-10-01
Parse and store specials with unique ptr

### LINKED ISSUE #1546: SpecialsParser should return unique_ptr and not leak memory.
**Environment: all**

**Description**
The `parse/SpecialsParser.cpp` is a generator.  In C++11 a generator should return a `std::unique_ptr<>` to indicate that it has created something that requires memory management and that it is passing management responsibility to the calling function.  The calling function can chose to receive the pointer as a `unique_ptr<>` or transfer it to a `shared_ptr<>` if it needs a shared ownership model.

**Expected Result**
Return `std::unique_ptr<>`s.

**Steps to reproduce**
Play game.

**Suggestions**

dbenage-cx:
On local system, compiling with boost v1.62 or lower fails(1.63+ is ok), opening PR for CI results to help troubleshoot

Resolves #1546

### DIFF
--- a/parse/SpecialsParser.cpp
+++ b/parse/SpecialsParser.cpp
@@ -8,28 +8,66 @@
 
 #include <boost/spirit/include/phoenix.hpp>
 
-
 #define DEBUG_PARSERS 0
 
 #if DEBUG_PARSERS
 namespace std {
     inline ostream& operator<<(ostream& os, const std::vector<std::shared_ptr<Effect::EffectsGroup>>&) { return os; }
-    inline ostream& operator<<(ostream& os, const std::map<std::string, Special*>&) { return os; }
-    inline ostream& operator<<(ostream& os, const std::pair<const std::string, Special*>&) { return os; }
+    inline ostream& operator<<(ostream& os, const std::map<std::string, std::unique_ptr<Special>&) { return os; }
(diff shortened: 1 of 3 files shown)


LGM-Doyle:
@dbenage-cx I can confirm that this works on Fedora with Boost v1.63.

dbenage-cx:
1.63 introduces variadic templates, the same results can be seen by defining `BOOST_NO_CXX11_VARIADIC_TEMPLATES`.

I was not able to determine the correct macros and order to increase number of args.
Instead, using a struct to handle transition.

parse/SpecialsParser.cpp
```diff
@@ -8,29 +8,67 @@
...
+    inline ostream& operator<<(ostream& os, const std::pair<const std::string, std::unique_ptr<Special>&) { return os; }
 }
 #endif
 
 namespace {
     const boost::phoenix::function<parse::detail::is_unique> is_unique_;
 
-    const boost::phoenix::function<parse::detail::insert> insert_;
+    struct special_pod {
+        special_pod (std::string name_,
```
geoffthemedio: remove space after `pod`

geoffthemedio:
otherwise looks fine

PR #1776, opened 2017-10-03
Fix AI usage of growth focus

Morlic-fo:
The first commit of this PR is some general refactoring of the growth focus evaluation code to improve readability.

The second commit fixes the calculation of the "usable"  population gained by using growth focus. The AI underestimated the benefit of the growth focus which made it not using the focus in some situations where it was already beneficial to do so.

### DIFF
--- a/default/python/AI/ResourcesAI.py
+++ b/default/python/AI/ResourcesAI.py
@@ -435,8 +435,9 @@ def set_planet_growth_specials(focus_manager):
             # the increased population on the planet using this growth focus
             # is mostly wasted, so ignore it for now.
             pop = planet.currentMeterValue(fo.meterType.population)
-            if pop > potential_pop_increase - 2 * planet.size:
-                _print_evaluation("would lose more population (%.1f) than total gain everywhere else." % pop)
+            pop_gain = potential_pop_increase - planet.size
+            if pop > pop_gain:
+                _print_evaluation("would lose more pop (%.1f) than gain everywhere else (%.1f)." % (pop, pop_gain))
                 continue
 
             # If we have a computronium special here, then research focus will have higher priority.
@@ -445,7 +446,7 @@ def set_planet_growth_specials(focus_manager):
                 continue
 
             _print_evaluation("considered (pop %.1f, growth gain %.1f, current focus %s)" % (
-                pop, potential_pop_increase - 2*planet.size, pinfo.current_focus))
+                pop, pop_gain, pinfo.current_focus))
 
             # add a bias to discourage switching out growth focus to avoid focus change penalties
             if pinfo.current_focus == GROWTH:
--- a/default/python/AI/ResourcesAI.py
+++ b/default/python/AI/ResourcesAI.py
@@ -403,7 +403,7 @@ def set_planet_growth_specials(focus_manager):
     if not useGrowth:
         return
 
-    # TODO: also consider potential future benefit re currently unpopulated planets
+    # TODO Consider actual resource output of the candidate locations rather than only population
     for special, locations in ColonisationAI.available_growth_specials.iteritems():
         # Find which metabolism is boosted by this special
         metabolism = AIDependencies.metabolismBoosts.get(special)
--- a/default/python/AI/ResourcesAI.py
+++ b/default/python/AI/ResourcesAI.py
@@ -400,37 +400,72 @@ def assess_protection_focus(pid, pinfo):
 
 def set_planet_growth_specials(focus_manager):
     """set resource foci of planets with potentially useful growth factors. Remove planets from list of candidates."""
-    if useGrowth:
-        # TODO: also consider potential future benefit re currently unpopulated planets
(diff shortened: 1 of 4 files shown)


default/python/AI/ResourcesAI.py
```diff
@@ -400,37 +400,73 @@ def assess_protection_focus(pid, pinfo):
...
-                    planet = focus_manager.all_planet_info[spPID].planet
-                    if focus_manager.bake_future_focus(spPID, GROWTH):
-                        print "%s focus of planet %s (%d) at Growth Focus" % (["set", "left"][cur_focus == GROWTH], planet.name, spPID)
-                        break
-                    else:
-                        print "failed setting focus of planet %s (%d) at Growth Focus; focus left at %s" % (planet.name, spPID, planet.focus)
+    if not useGrowth:
+        return
+
+    # TODO: also consider potential future benefit re currently unpopulated planets
```
Dilvish-fo: I suspect this TODO was added in a moment of exuberance, it no longer seems that sensible to me (having the colonization valuation calculations take into account the availability of a growith special seems to make a lot more sense than the other way around).  I'd suggest we go ahead and delete this TODO now.

Dilvish-fo (approved):
Looks good to me.

Dilvish-fo:
@Morlic-fo , heads-up that I am having some second thoughts about the actual 'fix' part of this-- at first I was thinking, hmm, I can't remember when there was actually ever a 2x penalty that would have applied there, but it certainly seemed intentional, how strange.  Then I realized, a bit too late, that it was probably an intentional overestimation of the penalty-- there is generally no point, it's even slightly disadvantageous, to change to growth focus if the planet loss is only evenly matched by the potential gain.  

I think the real bug there is that the comments and log message had not gotten properly updated when the penalty was increased.

I'm thinking through some other related changes, and am likely to add at least some of that penalty back, but if I do I'll be sure to make the intention clear in the comments and log message

Morlic-fo:
> there is generally no point, it's even slightly disadvantageous, to change to growth focus if the planet loss is only evenly matched by the potential gain.

- You increase population growth on all affected planets as it depends on the max. pop meter
- You possibly gain population on more important planets (such as the capital which has cultural archives)
- You split the RP/PP meter growth to multiple planets rather than having the single planet slowly grow its meters (I suppose a growth dithering would be worthwhile for that reason in combination with increased pop growth).

I noticed improvements in the order of 10% for the RP+2*PP metric in the early game where the AI was able to grab a planet with growth special early on (1st or 2nd expansion planet). I did not find negative impact in other games. Maybe you are thinking of (or found) scenarios that did not occur in my testgames.

Dilvish-fo:
> You increase population growth on all affected planets as it depends on the max. pop meter

Yes, although depending on what the current pop is on those planets, that additional population growth might be minuscule.  That's part of what I plan to change with this calculation, for it to take into account population growth sensitivity, not just max pop levels.

> You possibly gain population on more important planets (such as the capital which has cultural archives)

Yes, and perhaps the pop on Capital (or any planets with similar bonuses, though none come to mind) should be overweighted to account for that..

> You split the RP/PP meter growth to multiple planets rather than having the single planet slowly grow its meters (I suppose a growth dithering would be worthwhile for that reason in combination with increased pop growth).

I think once you separate out the pop growth sensitivity issue there is not much remaining value for this point-- it takes a fair of both research/industry tech and current pop to be in place for the growth rate differential to impact whether the RP/PP meter growth of a planet will be capped.  And the pop meter change calculations are rather asymmetrical, the situations where it could make sense to dither Growth focus would be rather rare if ever-- a planet has to have population of at least 10 for its pop growth rate to be as fast on the upswing as it would be on the downswing, and for smaller planets a single turn of pop loss could take several-to-many turns of pop growth to make up for.
```c++
// upswing
    pop_change = cur_pop * (target_pop + 1 - cur_pop) / 100;

// downswing
    pop_change = -(cur_pop - target_pop) / 10;
```

> I noticed improvements in the order of 10% for the RP+2*PP metric in the early game where the AI was able to grab a planet with growth special early on (1st or 2nd expansion planet). I did not find negative impact in other games. Maybe you are thinking of (or found) scenarios that did not occur in my testgames.

Yes, a newly founded planet of the same metab as home planet is an ideal place for growth focus since an industry or research focus would produce so little until the planet pop has had time to grow.  I'll make sure that what I do would not interfere with that.

Morlic-fo:
The idea is not to dither for population growth but abuse the fact that you lose resource meters on only 1 planet (which previously had industry/research focus) by 1 per turn while the population grows on N planets.

So even if the steady-state growth focus may be a bad idea due to excessive resource losses (for example because the growth special is on a huge planet with gaia special and mining boosts and gas giant and asteroids), it could be worth it to dither temporarily.

Dilvish-fo:
Hmm, yes, I suppose there are at least prospects for that being worthwhile in some cases.  When I take into account the growth sensitivity as I mentioned above, I'll look at putting in a short-term decision (which could result in dithering), or at least make a TODO comment about it.

PR #1777, opened 2017-10-03
Add PingServer() and PingLocalHostServer() to client networking.

LGM-Doyle:
This allows attempting to connect to the server to time out without
reporting an error.


geoffthemedio:
didn't test, but no objections

Issue #1780, opened 2017-10-05
Mismatch between the description and actual operation of  Empire::UnrestrictedLaneTravel()


Dilvish-fo:
The comments/descriptions for ```Empire::UnrestrictedLaneTravel()``` and the underlying ```m_available_system_exit_lanes``` are written up with their primary (only current) use case pattern in mind, but their implementation wound up being somewhat more specific and relied on the manner in which ```UnrestrictedLaneTravel()``` is currently always paired (either directly or indirectly) with ```Empire::SupplyUnobstructedSystems()```.  This leaves the comment/description for ```m_available_system_exit_lanes``` being perhaps a bit misleading, and the comments/description for  ```Empire::UnrestrictedLaneTravel()``` being yet more misleading.    

The point of deviation is that ```m_available_system_exit_lanes``` in fact only tracks lanes that are actually being preserved against blockade (by the presence of an empire fleet, as handled by ```Empire::UpdateSupplyUnobstructedSystems()```), ```m_available_system_exit_lanes``` does not necessarily include all the various lanes which are available simply because they are not blockaded.   The discrepancy became apparent to me as I was debugging some new AI pathfinding behavior that was at first trying to use ```UnrestrictedLaneTravel()``` without ```Empire::SupplyUnobstructedSystems()```.

There are two main approaches to rectifying this-- change the comments or change the behavior (and in the latter case there can be some notable variance in what behavior is changed).  Concurrent with this submission I am proposing two alternative PRs for discussion, one clarifying comments/descriptions and the other mostly changing behavior.  I thought the individual PRs could be focused on the best way to do their particular approach, and this Issue could be place to discuss the choice between the two.   I propose that in either case we leave the derivation of ```m_available_system_exit_lanes``` be as it is now, and simply clarify the description for it.   

Then in the first PR #1781, we simply also make similar corrections to the description for ```UnrestrictedLaneTravel()```.  This PR is my strong preference because it retains some more granularity of information that the AI could use in its pathfinding code without having to redetermine the info from ```m_available_system_exit_lanes``` (which is possible but seems to me it would clutter and bog down the AI code).  To make things even more clear, perhaps we should also consider changing the names to be ```m_preserved_system_exit_lanes``` and ```PreservedLaneTravel()```

[...]

PR #1781, opened 2017-10-05
Unrestricted lane travel conform comments

Dilvish-fo:
As noted in Issue #1780, there is currently a mismatch between the description and actual operation of  ```UnrestrictedLaneTravel()```; this PR proposes to correct that by correcting the comments/description to conform to the actual operation.  

(This is presented along with the main alternative correction for consideration in #1782, which instead proposes to conform the operation to the current description, which essentially causes it to merge in the results from ```Empire::SupplyUnobstructedSystems()```.)

As noted in the above Issue, for the main code base either approach would not make a current difference in operation, because of the manner in which all current uses of ```UnrestrictedLaneTravel()``` are also accompanied (directly or indirectly) by references to ```Empire::SupplyUnobstructedSystems```.

However, the AI could beneficially use the more specific information as retained in this PR (and that information could be separately recreated if necessary, but that would seem cluttered and wasteful).

A further clarification that might be helpful as part of this PR could be to change the attribute and method names to be ```m_preserved_system_exit_lanes``` and ```PreservedLaneTravel()```

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -2005,11 +2005,11 @@ void Empire::UpdateSupplyUnobstructedSystems(const std::set<int>& known_systems)
             // so it is obstructed, so isn't included in the unobstructed
             // systems set.  Furthermore, this empire's available system exit
             // lanes for this system are cleared
-            if (!m_available_system_exit_lanes[sys_id].empty()) {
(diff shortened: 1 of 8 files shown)


Dilvish-fo:
I've now renamed the related Empire attribute and its getter, as discussed, and rebased.  This compiles fine for me and appears to run fine, so I am going to go ahead and merge now.

PR #1782, opened 2017-10-05
Unrestricted lane travel functional correction

Dilvish-fo:
As noted in Issue #1780, there is currently a mismatch between the description and actual operation of  ```UnrestrictedLaneTravel()```; this PR proposes to correct that by conforming the operation to the description.  This is presented along with the alternative correction in PR #1781, which proposes to simply conform the description to the actual operation.

As noted in that other PR, I think that other approach would be more convenient/helpful for the AI and so I prefer that other PR.  But I thought it would help clarify the discussion to put them both up for review.


Dilvish-fo:
Per discussion in the underlying Issue, I'm closing this in favor of #1781

Issue #1548, opened 2017-05-05
ShipHullParser should return unique_ptr and not leak memory.


LGM-Doyle:
**Environment: all**

**Description**
The `parse/ShipHullParser.cpp` is a generator.  In C++11 a generator should return a `std::unique_ptr<>` to indicate that it has created something that requires memory management and that it is passing management responsibility to the calling function.  The calling function can chose to receive the pointer as a `unique_ptr<>` or transfer it to a `shared_ptr<>` if it needs a shared ownership model.

Furthermore the error condition near line 38 leaks the passed in pointer.  This is a static leak, it only leaks objects that have failed to parse once per game.

**Expected Result**
Return `std::unique_ptr<>`s.

**Steps to reproduce**
Play game.

**Suggestions**
A solution of the same issue in the BuildingType parser is in #1544, is anyone would like to assign themselves and help out.

PR #1783, opened 2017-10-06
Parse and store PartType with unique_ptr

### LINKED ISSUE #1551: ShipPartsParser should return unique_ptr and not leak memory.
**Environment: all**

**Description**
The `parse/ShipPartsParser.cpp` is a generator.  In C++11 a generator should return a `std::unique_ptr<>` to indicate that it has created something that requires memory management and that it is passing management responsibility to the calling function.  The calling function can chose to receive the pointer as a `unique_ptr<>` or transfer it to a `shared_ptr<>` if it needs a shared ownership model.

Furthermore the error condition near line 36 leaks the passed in pointer.  This is a static leak, it only leaks objects that have failed to parse once per game.

**Expected Result**
Return `std::unique_ptr<>`s.

**Steps to reproduce**
Play game.

**Suggestions**

dbenage-cx:
Resolves #1551

### DIFF
--- a/universe/ShipDesign.cpp
+++ b/universe/ShipDesign.cpp
@@ -197,11 +197,14 @@ PartTypeManager::PartTypeManager() {
         throw;
     }
 
-    TraceLogger() << "Part Types:";
-    for (const auto& entry : m_parts) {
-        const PartType* p = entry.second;
-        TraceLogger() << " ... " << p->Name() << " class: " << p->Class();
-    }
+    TraceLogger() << [&]() {
+            std::string retval("Part Types:");
+            for (const auto& pair : m_parts) {
+                const auto& part = pair.second;
+                retval.append("\n\t" + part->Name() + " class: " + boost::lexical_cast<std::string>(part->Class()));
+            }
+            return retval;
+        }();
 
     // Only update the global pointer on sucessful construction.
     s_instance = this;
@@ -209,15 +212,9 @@ PartTypeManager::PartTypeManager() {
     DebugLogger() << "PartTypeManager checksum: " << GetCheckSum();
 }
 
-PartTypeManager::~PartTypeManager() {
-    for (auto& entry : m_parts) {
-        delete entry.second;
-    }
-}
-
 const PartType* PartTypeManager::GetPartType(const std::string& name) const {
     auto it = m_parts.find(name);
-    return it != m_parts.end() ? it->second : nullptr;
+    return it != m_parts.end() ? it->second.get() : nullptr;
 }
 
 const PartTypeManager& PartTypeManager::GetPartTypeManager() {
--- a/universe/ShipDesign.cpp
+++ b/universe/ShipDesign.cpp
@@ -197,7 +197,7 @@ PartTypeManager::PartTypeManager() {
         throw;
     }
 
-    TraceLogger() << [&]() {
+    TraceLogger() << [this]() {
             std::string retval("Part Types:");
             for (const auto& pair : m_parts) {
                 const auto& part = pair.second;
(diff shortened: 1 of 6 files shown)


geoffthemedio:
looks OK

universe/ShipDesign.cpp
```diff
@@ -197,27 +197,24 @@ PartTypeManager::PartTypeManager() {
         throw;
     }
 
-    TraceLogger() << "Part Types:";
-    for (const auto& entry : m_parts) {
-        const PartType* p = entry.second;
-        TraceLogger() << " ... " << p->Name() << " class: " << p->Class();
-    }
+    TraceLogger() << [&]() {
+            std::string retval("Part Types:");
```
LGM-Doyle: Same as the other PR.  I would prefer an explicit capture `[this]`

Dilvish-fo:
Just to confirm something about this whole set of PRs for unique_ptr, while the original version of the one for Specials would (as expected) not compile for me with Boost 1.58, the version of it that was merged into master, and this version, do compile and run just fine with Boost 1.58.

I was wanting to try testing compilation with Boost 1.56, but it seems that the cmake defines I used to use to specify a nonstandard location for boost are no longer working for me.  So, has anyone compiled this against boost 1.56 so we can be sure our reqs have not changed?    And if anyone can tell me the right cmake variables to define to specify an alternate boost location/version that would be appreciated.  I used to (perhaps a couple years ago now) do it just by defining ```CMAKE_INCLUDE_PATH``` and ```CMAKE_LIBRARY_PATH``` to point to the include and lib folders of the alternate boost installation.

dbenage-cx:
@Dilvish-fo For some reason, I had 1.58 in my head as the minimum due and is the earliest I have compiled.  1.59 may be the minimum for windows though, per [sdk commit](https://github.com/freeorion/freeorion-sdk/commit/dca3656e327f1c2939cd41c144f9e8abe39d1a65).

I use different versions via `cmake -DBOOST_ROOT=/boost/dir ..`, there was some minor discussion on #1385 when I was using a different approach.  If you like I can make a forum post on my current setup.

dbenage-cx:
Marking `WIP` pending resolution of some existing memory leaks per [other PR comment](https://github.com/freeorion/freeorion/pull/1784#issuecomment-334821683).

LGM-Doyle:
@dbenage-cx I think that there has been some confusion.  Both #1784 and this PR are good to merge as is.

These PRs fix the leaks that already exist in the current code.

The leak in the existing code is that after `raw_pointer = new X`, (or `new_<X>()`) is called any error condition will drop the parse result and leak the `X`.  Specifically in the case of #1547, #1548, #1549, #1550, and #1551 a duplicate key would mean that the insert into the `map` fails and the pointed too object is dropped.

This PR fixes the leak by immediately calling `std::unique_ptr<X>(new X)`, which means that any error in the parser or a duplicate key will eventually call the `std::unique_ptr` destructor.  Hence no leak.

dbenage-cx:
Ah, I mistook your comment to mean there were 3 of 6 memory leaks fixed in that one PR, which surprised me from both ends.
In that case, I will merge these 2 when I return tonight (pending no other issues noted) and work on PRs for the other related issues.

Dilvish-fo:
>  there was some minor discussion on # 1385 when I was using a different approach. If you like I can make a forum post on my current setup.

@dbenage-cx that would be great if you could make that post-- just just using BOOST_ROOT as recommended in that 1385 discussion is not working for me, not even if I also specified BOOST_INCLUDEDIR and BOOST_LIBRARYDIR.  Perhaps my old boost 1.56 installation does not have all libraries currently needed & so it just rejects that installation.  I tried it with Boost_DEBUG=ON to make FindBoost print out some debugging info, but although it confirmed the above vars were set, it still did not give any indication of why it wound up ignoring them and going for the main/standard boost installation.

dbenage-cx:
@Dilvish-fo [forum thread](http://freeorion.org/forum/viewtopic.php?f=24&t=10725&sid=115a7df3ad6b91aa9dd3742dfea164ce)

PR #1784, opened 2017-10-06
Parse and store HullType with unique_ptr

### LINKED ISSUE #1548: ShipHullParser should return unique_ptr and not leak memory.

dbenage-cx:
Resolves #1548

### DIFF
--- a/universe/ShipDesign.cpp
+++ b/universe/ShipDesign.cpp
@@ -650,11 +650,13 @@ HullTypeManager::HullTypeManager() {
         throw;
     }
 
-    TraceLogger() << "Hull Types:";
-    for (const auto& entry : m_hulls) {
-        const HullType* h = entry.second;
-        TraceLogger() << " ... " << h->Name();
-    }
+    TraceLogger() << [&]() {
+            std::string retval("Hull Types:");
+            for (const auto& entry : m_hulls) {
+                retval.append("\n\t" + entry.second->Name());
+            }
+            return retval;
+        }();
 
     if (m_hulls.empty())
         ErrorLogger() << "HullTypeManager expects at least one hull type.  All ship design construction will fail.";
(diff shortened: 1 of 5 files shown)


universe/ShipDesign.cpp
```diff
@@ -650,11 +650,13 @@ HullTypeManager::HullTypeManager() {
         throw;
     }
 
-    TraceLogger() << "Hull Types:";
-    for (const auto& entry : m_hulls) {
-        const HullType* h = entry.second;
-        TraceLogger() << " ... " << h->Name();
-    }
+    TraceLogger() << [&]() {
+            std::string retval("Hull Types:");
```
LGM-Doyle: I prefer explicit capture of `[this]` here.

In this location it make no difference.

Implicit capture as a practice throughout the code base, may easily lead to capturing a variable with shorter scope than the lambda.  Given that the code  is written by many hands, there are many differing uses of scope and commonly using unnamed capture will make it harder to debug crashes due to a lambda referencing an out of scope object.

LGM-Doyle (commented):
Since there will be minor conflicts, renaming `X*` to `uniqueptr<X>` with #1785, I can wait for you to finish all of these bug fixes before merging any of #1785, so that I do the rebasing work.

dbenage-cx:
> all of these bug fixes

@LGM-Doyle Do you refer to both PRs, or is there more than one in this PR?
Lately, github notices are not always showing, so want to make sure that is not extending to review comments.

LGM-Doyle:
You have changed everything I requested.

You have fixed 3 of the 6 memory leaks in the parser.  I was hoping that you would fix the other 3 as well.  I was suggesting that if you plan to fix the other 3, that I can wait to merge the parts of #1785 that would conflict with the fixes until after you have completed the fixes.  The conflicts would  be minor, but the offer is there.

dbenage-cx:
I need to familiarize myself with valgrind and identify those leaks.
Please do not feel a need to wait on me, especially as it may be easier to rebase these smaller PRs.

PR #1785, opened 2017-10-06
Parse on background thread

LGM-Doyle:
I will break this PR into smaller PRs that are easier to review.  

This PR moves parsing from the main UI thread to another thread.  I describe specific improvements below.

Most of the commits fall into two categories related to fixing bugs in the lexer or making the parsing multi-threaded, as described in the sections below.

**Improvements**
- This PR improves startup performance by removing parsing from the foreground thread.  A simple example is to start the Intro screen and then select the buildings section of the Pedia.  It will open faster, because parsing starts and possibly completes before the window is opened.

- It fixes the bug where the singleton lexer prevented concurrent parsing, and caused errors in file length checking and error reporting.  This PR restores checking for and reporting on parsers failing because of garbage in the file after the parsed content.

- It fixes the leak due to parsers being static initialized, used at startup, never used again and only recovered when the app is closed.

- It decouples the universe object managers from the parser the directory structure.  This is a step towards enabling testing universe and game engine code without an existing scripting directory.

- It moves the parsed file paths out of the parser and into the app code.  This make the scripting directory structure a property of the app, not the parser.

- It enables the resource directory to be changed without requiring a restart.

**Bug fix overview**
The lexer is currently a statically initialized singleton.  It is bound to a single specific file.  Consequently, only one file can be parsed at a time.  This is true even on a single thread.

This causes problems with ship designs containing parts and species using macros.  The check for correctly parsed file length was disabled to hide this bug.  The parser can crash the application if an error iterator from one file points beyond the end of the file being parsed, when the error is presented to the UI.

The fix is to create one lexer per file being concurrently parsed.  After the fix, the `parse::THING` functions are safe to call concurrently and multi-threaded.

The problem is replicated for every parser that is bound to the lexer.

Many of the parsers are statically initialized.  The order of initialization is at the discretion of the linker and the loader and difficult to infer, as noted by several comments in the code.  This PR changes all the parsers to be on the stack and makes initialization order explicit and obvious.

Secondly, not a bug, but many of the parsers are rules with a distinguished rule often called `start`.  This is exactly the purpose of `spirit::grammar`.  This PR replaces rules with a single distinguished rule with a grammar.

The fix portion of this PR is repetitive, but straightforward.  Find a singleton parser.  Convert it to a grammar.  Pass it as a parameter to its dependents.  Remove the singleton.  Repeat.

**Multithreaded parsing overview**
The parsing is made multi-threaded by converting the universe object managers to take a `std::future<>` with the results of the parse.  Parsing can be started early on separate threads in the clients and the server before it is needed in the `MapWnd`, pedia or universe generation.

### DIFF
--- a/universe/Special.h
+++ b/universe/Special.h
@@ -8,12 +8,12 @@
 #include <boost/optional/optional.hpp>
 
 #include "../util/Export.h"
+#include "../util/Pending.h"
 
 #include <memory>
 #include <string>
 #include <vector>
 #include <map>
-#include <future>
 
 
 namespace Effect {
@@ -134,7 +134,7 @@ class FO_COMMON_API SpecialsManager {
     unsigned int GetCheckSum() const;
 
     /** Sets types to the value of \p future. */
-    FO_COMMON_API void SetSpecialsTypes(std::future<SpecialsTypeMap>&& future);
+    FO_COMMON_API void SetSpecialsTypes(Pending::Pending<SpecialsTypeMap>&& future);
 
     private:
     /** Assigns any m_pending_types to m_specials. */
@@ -142,7 +142,7 @@ class FO_COMMON_API SpecialsManager {
 
     /** Future types being parsed by parser.  mutable so that it can
         be assigned to m_species_types when completed.*/
-    mutable boost::optional<std::future<SpecialsTypeMap>> m_pending_types = boost::none;
+    mutable boost::optional<Pending::Pending<SpecialsTypeMap>> m_pending_types = boost::none;
 
     mutable SpecialsTypeMap m_specials;
 };
--- a/universe/Special.h
+++ b/universe/Special.h
@@ -5,6 +5,7 @@
 #include "ValueRefFwd.h"
 
 #include <boost/serialization/nvp.hpp>
+#include <boost/optional/optional.hpp>
 
 #include "../util/Export.h"
 
@@ -12,6 +13,7 @@
 #include <string>
 #include <vector>
 #include <map>
+#include <future>
 
 
 namespace Effect {
@@ -120,8 +122,10 @@ void Special::serialize(Archive& ar, const unsigned int version)
 }
 
 /** Look up table for specials.*/
-class SpecialsManager {
+class FO_COMMON_API SpecialsManager {
     public:
+    using SpecialsTypeMap = std::map<std::string, std::unique_ptr<Special>>;
+
     SpecialsManager();
     ~SpecialsManager();
 
@@ -129,10 +133,20 @@ class SpecialsManager {
     const Special* GetSpecial(const std::string& name) const;
     unsigned int GetCheckSum() const;
 
+    /** Sets types to the value of \p future. */
+    FO_COMMON_API void SetSpecialsTypes(std::future<SpecialsTypeMap>&& future);
+
     private:
-    std::map<std::string, std::unique_ptr<Special>> m_specials;
+    /** Assigns any m_pending_types to m_specials. */
+    void CheckPendingSpecialsTypes() const;
+
+    /** Future types being parsed by parser.  mutable so that it can
+        be assigned to m_species_types when completed.*/
+    mutable boost::optional<std::future<SpecialsTypeMap>> m_pending_types = boost::none;
+
+    mutable SpecialsTypeMap m_specials;
 };
 
-const SpecialsManager& GetSpecialsManager();
+FO_COMMON_API SpecialsManager& GetSpecialsManager();
 
 #endif // _Special_h_
--- a/universe/Special.h
+++ b/universe/Special.h
@@ -11,6 +11,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <map>
 
 
 namespace Effect {
@@ -117,5 +118,21 @@ void Special::serialize(Archive& ar, const unsigned int version)
         & BOOST_SERIALIZATION_NVP(m_location)
         & BOOST_SERIALIZATION_NVP(m_graphic);
 }
+
+/** Look up table for specials.*/
+class SpecialsManager {
+    public:
+    SpecialsManager();
+    ~SpecialsManager();
+
+    std::vector<std::string> SpecialNames() const;
+    const Special* GetSpecial(const std::string& name) const;
+    unsigned int GetCheckSum() const;
+
+    private:
+    std::map<std::string, std::unique_ptr<Special>> m_specials;
+};
+
+const SpecialsManager& GetSpecialsManager();
 
 #endif // _Special_h_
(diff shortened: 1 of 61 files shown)


Dilvish-fo:
Sounds like a very good advance.

> Parsing can be started early on separate threads in the clients and the server before it is needed in the `MapWnd`, pedia or universe generation. 

It seems to me it could also be helpful to have the AI Clients specify a slightly lower priority for their parsers than the human client or server do.

dbenage-cx:
@LGM-Doyle [Return std::unique_ptr from parsers Project](https://github.com/freeorion/freeorion/projects/3) should be complete (No link method for github projects?)

LGM-Doyle:
@dbenage-cx thanks for doing the work.  I'm in the process of rebasing this PR.

It is strange that there is no conversion as part of a github project.

LGM-Doyle:
This is rebased and no longer a work in progress.

I added a single commit so that pending saved designs always show up when expected.

LGM-Doyle:
With the parser bug fix merged, this PR only concerns non-main thread parsing.  This description should make reviewing easier.

The PR basically for each type of universe object creates a `std::future` to hold pending parse results (from another thread) and passes those to the universe object type manager.  The manager checks if the results are complete, when it first needs the results.

In "util/AppInterface.cpp" the PR creates `IApp::ParseUniverseObjectTypes()`, which calls each of the universe type managers and to parsing content on a background thread like this:
````GetBuildingTypeManager().SetBuildingTypes(Pending::StartParsing(parse::buildings, rdir / "scripting/buildings"));````

The new file "util/Pending.h" defines `Pending::StartParsing()` which creates the `std::future`.  This file also defines `WaitForPending()` and `SwapPending()` which wait for a pending future result and then swap that result with the current set of things.

Each type of universe object defines a function like `CheckPendingBuildingTypes()` which checks if a `std::future` has results and swaps it in using the functions in "Pending.h"

**Comment by geoffthemedio** [comment]
The name `ParseUniverseObjectTypes` is misleading, as it also parses non-object stuff like articles, or not-really-object stuff like species or specials or part types.

File: universe/Special.h [hunk 1]
```diff
@@ -118,4 +121,32 @@ void Special::serialize(Archive& ar, const unsigned int version)
...
+    ~SpecialsManager();
+
+    std::vector<std::string> SpecialNames() const;
+    const Special* GetSpecial(const std::string& name) const;
+    unsigned int GetCheckSum() const;
+
+    /** Sets types to the value of \p future. */
+    FO_COMMON_API void SetSpecialsTypes(Pending::Pending<SpecialsTypeMap>&& future);
+
+    private:
```
geoffthemedio: dedent

universe/Species.cpp
```diff
@@ -356,85 +355,76 @@ SpeciesManager::SpeciesManager() {
...
-
     // Only update the global pointer on sucessful construction.
     s_instance = this;
 }
 
 const Species* SpeciesManager::GetSpecies(const std::string& name) const {
+    CheckPendingSpeciesTypes();
     auto it = m_species.find(name);
-    return it != m_species.end() ? it->second.get() : nullptr;
+    return it != end() ? it->second.get() : nullptr;
```
geoffthemedio: why `end()` instead of `m_species.end()`? The extra call to `CheckPendingSpeciesTypes()` seems unnecessary.

universe/Tech.h
```diff
@@ -298,28 +301,44 @@ class FO_COMMON_API TechManager {
     unsigned int                    GetCheckSum() const;
     //@}
 
+    using TechParseTuple = std::tuple<
+        TechManager::TechContainer, // techs_
+        std::map<std::string, std::unique_ptr<TechCategory>>, // tech_categories,
+        std::set<std::string> // categories_seen
+        >;
+        /** Sets types to the value of \p future. */
```
geoffthemedio: dedent

universe/Universe.cpp
```diff
@@ -175,6 +167,37 @@ void Universe::ResetAllIDAllocation(const std::vector<int>& empire_ids) {
...
+void Universe::SetInitiallyUnlockedFleetPlans(Pending::Pending<std::vector<FleetPlan*>>&& future)
+{ m_pending_fleet_plans = std::move(future);}
+
+const std::vector<FleetPlan*>& Universe::InitiallyUnlockedFleetPlans() const
+{ return Pending::SwapPending(m_pending_fleet_plans, m_unlocked_fleet_plans); }
+
+void Universe::SetMonsterFleetPlans(Pending::Pending<std::vector<MonsterFleetPlan*>>&& future)
+{ m_pending_monster_fleet_plans = std::move(future); }
+
+const std::vector<MonsterFleetPlan*>&  Universe::MonsterFleetPlans() const
```
geoffthemedio: xtra space

universe/Universe.h
```diff
@@ -400,6 +405,33 @@ class FO_COMMON_API Universe {
...
 
+    /** Set items unlocked before turn 1 from \p future.*/
+    void SetInitiallyUnlockedItems(Pending::Pending<std::vector<ItemSpec>>&& future);
+    /** Items unlocked before turn 1.*/
+    const std::vector<ItemSpec>&  InitiallyUnlockedItems() const;
+
+    /** Set buildings unlocked before turn 1 from \p future.*/
+    void SetInitiallyUnlockedBuildings(Pending::Pending<std::vector<ItemSpec>>&& future);
+    /** Buildings unlocked before turn 1.*/
+    const std::vector<ItemSpec>&  InitiallyUnlockedBuildings() const;
```
geoffthemedio: bunch of extra spaces in all these...

util/Pending.h
```diff
@@ -0,0 +1,106 @@
...
+/** namespace Pending collection classes and functions used for
+    asynchronously parsing universe types, statistics etc.*/
+namespace Pending {
+    /** Pending holds the std::future results of a T.*/
+    template <typename T>
+    struct FO_COMMON_API Pending {
+        Pending (boost::optional<std::future<T>>&& pending_, const std::string& name_) :
+            pending(std::move(pending_)),
+            filename(name_)
+            {}
```
geoffthemedio: dedent to match function name

util/Pending.h
```diff
@@ -0,0 +1,106 @@
...
+#include <future>
+#include <string>
+
+/** namespace Pending collection classes and functions used for
+    asynchronously parsing universe types, statistics etc.*/
+namespace Pending {
+    /** Pending holds the std::future results of a T.*/
+    template <typename T>
+    struct FO_COMMON_API Pending {
+        Pending (boost::optional<std::future<T>>&& pending_, const std::string& name_) :
```
geoffthemedio: prefer no space after Pending before (

geoffthemedio:
seems to work...

util/Pending.h
LGM-Doyle: Done.

LGM-Doyle (commented):
Changes are done.  

I also changed ParseUniverseObjectTypes() to StartBackgroundParsing().

geoffthemedio:
I assume the Travis build timeout can be ignored, so OK to merge.

PR #1786, opened 2017-10-06
Remove filename singletons from parser.

LGM-Doyle:
The use of s_filename, s_text_it, s_begin, s_end and Rule as static
singletons requires parsing to be single threaded, non-recursive and
non-concurrent, or generate_error_string() will fail.

Any recursive or concurrent parse accessing two or more files (ships
with parts, species with macros, etc) would not have correct error
messages and might crash on a bad parse.  The second file opened resets
all of the singleton variables, which if there is an error in the first
file will be pointing to the wrong variables.  At best this indexes into
the wrong string.  At worst it indexes into a freed string.

This commit removes the singletons and threads the filename, begin and
end through the Rule object to generate_error_string().

### DIFF
--- a/parse/BuildingsParser.cpp
+++ b/parse/BuildingsParser.cpp
@@ -41,7 +41,9 @@ namespace {
     BOOST_PHOENIX_ADAPT_FUNCTION(void, insert_building_, insert_building, 6)
 
     struct rules {
-        rules() {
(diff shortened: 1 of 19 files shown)


LGM-Doyle:
Merged along with #1795.

PR #1787, opened 2017-10-06
Research tree balancing—Learning focused

MatGB:
Work in progress
- Moves Stellar Tomography earlier in tree, reduces costs and effect
- Changes Nascent AI to give an infrastructure based bonus
- Changes Adaptive Automation to give Infrastructure based bonus
- Reduces cost and power of Quantum Networking
- Reduces stat change bonus from Force Energy Structures

Follows from various forum discussions, objective is to make the research tree more interesting and less powerful, with more early/mid game bonuses but reduced overall bonuses. Tones down Force Energy because that's been way overpowered for a long time, and if the total bonuses available are to be reduced you don't need the substantial speed bump it gives.

I tried the Nasent AI infrastructure change as an experiment and liked it so much I did the same to Adaptive Automation, on balance it's a much more effective change than I was expecting and reduces the incentive to micromanage colony focus settings in the turns after conquest/colonisation.

I've been playing with this for a few weeks now and while the AI hasn't been adjusted I much prefer the tempo it gives the game: further work to follow but we said incremental changes were better and easier for the @freeorion/ai-team to adapt to.

### DIFF
--- a/default/scripting/techs/construction/FORCE_ENERGY_STRC.focs.txt
+++ b/default/scripting/techs/construction/FORCE_ENERGY_STRC.focs.txt
@@ -15,31 +15,31 @@ Tech
             scope = And [
                 ProductionCenter
                 OwnedBy empire = Source.Owner
-                Industry high = RootCandidate.TargetIndustry - 3
+                Industry high = RootCandidate.TargetIndustry - 2
             ]
-            effects = SetIndustry value = Value + 2
+            effects = SetIndustry value = Value + 1
 
         EffectsGroup
             scope = And [
                 ProductionCenter
                 OwnedBy empire = Source.Owner
-                Research high = RootCandidate.TargetResearch - 3
+                Research high = RootCandidate.TargetResearch - 2
             ]
-            effects = SetResearch value = Value + 2
+            effects = SetResearch value = Value + 1
 
         EffectsGroup
             scope = And [
                 ProductionCenter
                 OwnedBy empire = Source.Owner
-                Construction high = RootCandidate.TargetConstruction - 3
+                Construction high = RootCandidate.TargetConstruction - 2
             ]
-            effects = SetConstruction value = Value + 2
+            effects = SetConstruction value = Value + 1
 
         EffectsGroup
             scope = And [
                 ProductionCenter
                 OwnedBy empire = Source.Owner
-                Industry low = RootCandidate.TargetIndustry - 3 high = RootCandidate.TargetIndustry
+                Industry low = RootCandidate.TargetIndustry - 2 high = RootCandidate.TargetIndustry
             ]
             effects = SetIndustry value = Target.TargetIndustry
 
@@ -47,7 +47,7 @@ Tech
             scope = And [
                 ProductionCenter
                 OwnedBy empire = Source.Owner
-                Research low = RootCandidate.TargetResearch - 3 high = RootCandidate.TargetResearch
+                Research low = RootCandidate.TargetResearch - 2 high = RootCandidate.TargetResearch
             ]
             effects = SetResearch value = Target.TargetResearch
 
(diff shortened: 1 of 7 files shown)


Vezzra:
I think the changes to the Stellar Tomography, Quantum Networking and Force Energy Structures sound reasonable. Although, if you are already at it, Pure Energy Metabolism (I hope I git that right) also is in desperate need of a nerf too.

However, I *strongly* object to the changes to Nascent AI and Adaptive Automation. I realize you have a penchant for basing some of the resource boni on infrastructure instead of population, but AFAIK the (long standing) design decision to base (non-flat) resource boni solely on pop hasn't been revised, so we need to stick to that.

Deviating from that will introduce a quite fundamental change to the intended purpose of population, infrastructure and the basic resource types and how the depend on each other, with far reaching consequences to game dynamics and future game design.

Don't get me wrong, I don't say that this design decision isn't open to discussion, but before we introduce/merge an actual implementation of such a change, we need a proper design discussion with a solid resolve (which should happen on the forum of course).

Given that you've suggested that repeatedly in the past, I suggest to open a thread on the design subforums (preferably on the Top Priority Game Design subforum, as I think that topic has some priority - if you can't open new threads there, AFAIK only moderators can, open it in the Other Game Design subforum and I'll move the thread).

That said, a change to Adaptive Automation you could do in this PR already is nerfing it. 5 is ***way*** too much, I suggest decreasing that to 2 (like NAI), or 3 at least.

Vezzra:
Side note: this is a tweak, not a refactoring. 😉

MatGB:
Forum thread is planned, but I had to push my branch so we could bughunt and it made sense to do the PR.

Changing PEM is also very high on the list but I realised I'd gotten ahead in the 'small incremental changes' part and wanted to push this first so people could look at it: PEM and Enclave To the Void are next.

Vezzra:
Ah, ok. Yep, makes sense.

Regarding forum thread: Please separate a discussion about the changes this PR introduces from the more fundamental discussion about basing resource boni on other meters than pop. Each of these should get their own thread.

dbenage-cx:
Just FYI: Looks like the addition of Stellar Tomography as prereq for Gravitonics causes an error for redundant tech (at least with rebase to master):
`ERROR: Redundant tech dependency found (A <-- B means A is a prerequisite of B): PRO_SINGULAR_GEN <-- PRO_ZERO_GEN, PRO_ZERO_GEN <-- PRO_NEUTRONIUM_EXTRACTION, PRO_SINGULAR_GEN <-- PRO_NEUTRONIUM_EXTRACTION; remove the PRO_SINGULAR_GEN <-- PRO_NEUTRONIUM_EXTRACTION dependency.`

Vezzra:
@freeorion/ai-team, how much work will it be to adjust the AI to the changes in this PR? I've set it to the "v0.4.8 (optional)" milestone for now, assuming that the required AI adjustments wouldn't be too difficult.

Otherwise I'd suggest to postpone this PR until after the release, as making the AI being able to handle the new Imperial Stockpile feature has top priority, and I don't want to delay 0.4.8 too much.

Dilvish-fo:
This had slipped past me.  I share your expectation that this shouldn't be too bad to adjust the AI for, should be fine for 0.4.8.

I note this still appears to have the NAI bonus based off of construction, I assume there is still some more changes and/or discussion expected?

Vezzra:
> I note this still appears to have the NAI bonus based off of construction, I assume there is still some more changes and/or discussion expected?

Yes, definitely. I strictly against merging something that introduces a change which is in conflict with a design decision that hasn't been revised. So either remove the dependence of a resource bonus on infrastructure from this PR (which I recommend, so the other changes can go in), or keep it on hold unless we can have a proper design discussion and resolution on the issue.

PR #1788, opened 2017-10-06
Automatically verify checksums.

LGM-Doyle:
This PR automatically verifies checksums and reports a warning in the log if there is a mismatch.
It does not do checksums for the single player client, since the server and client are started from a common directory so there can be no mismatch.

### DIFF
--- a/client/AI/AIClientApp.cpp
+++ b/client/AI/AIClientApp.cpp
@@ -395,16 +395,7 @@ void AIClientApp::HandleMessage(const Message& msg) {
 
     case Message::CHECKSUM: {
         TraceLogger() << "(AIClientApp) CheckSum.";
-        std::map<std::string, unsigned int> checksums;
-        ExtractContentCheckSumMessageData(msg, checksums);
-
-        auto log_msg = [checksums]() {
-                std::string retval("");
-                for (const auto& cs : checksums)
-                    retval.append("\n\t" + cs.first + " : " + std::to_string(cs.second));
-                return retval;
-            };
-        DebugLogger() << "Got checksum message from server:" << log_msg();
+        VerifyCheckSum(msg);
         break;
     }
 
--- a/client/ClientApp.cpp
+++ b/client/ClientApp.cpp
@@ -167,3 +167,24 @@ void ClientApp::SetEmpireID(int empire_id)
 
 void ClientApp::SetCurrentTurn(int turn)
 { m_current_turn = turn; }
+
+void ClientApp::VerifyCheckSum(const Message& msg) {
+    std::map<std::string, unsigned int> server_checksums;
+    ExtractContentCheckSumMessageData(msg, server_checksums);
+
+    auto client_checksums = CheckSumContent();
+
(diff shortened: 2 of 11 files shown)


geoffthemedio:
The SP server can be run from a different location when desired...

geoffthemedio:
`--force-external-server`

LGM-Doyle:
Thanks for the additional information.

I've added a server startup flag `--skip-checksum`, which is added when a single player game is started and `--force-external-server` is false.

geoffthemedio:
Why skip it ever?

LGM-Doyle:
The checksum takes a noticeable amount of time and delays startup.

It is not skipping the check.  It is determining that it trivially passes without doing the work.

This method tells the server, "I know that the checksum will pass, so don't waste time computing it"

Another method would be for a local connection to checksum the directory first and then if that is not the same, checksum the contents.

LGM-Doyle:
Rebased.

geoffthemedio:
Any idea what's up with the OSC build error...?
````
/Users/travis/build/freeorion/freeorion/python/EnumWrapper.cpp:215:13: error: expected expression

            .value("toggle",        GameRules::TOGGLE)
            ^
````

o01eg:
Need to rebase over latest commit in `master`:

https://github.com/freeorion/freeorion/commit/98625c0743266d7fcf9cb70bcb23cef0e2822fa8 intoduced this typo
https://github.com/freeorion/freeorion/commit/aaee5e34bd79917e200d8cd7bcd275f57260db9f fixed it

LGM-Doyle:
Thanks @o01eg I did not realize that I rebased onto a broken master.

This is now rebased.

geoffthemedio:
Started a hostless server, and the client with --force-external-server. Joined the server, added an AI, and started the game. In the client log, I see
````
12:57:03.680965 [error] FSM : HumanClientFSM.cpp:115 : A CheckSum event was not handled by any of these states : [struct WaitingForMPJoinAck].  It is being ignored.
````

LGM-Doyle:
The master branch has the same error, so it is **not** caused by this PR.

The issue is that the `WaitingForJoinAck` state on the client does not handle the checksum response.

I've added a commit to **this** PR to sawp the order of the join ACK and checksum in the server, which fixes the issue.

o01eg:
@geoffthemedio If you start hostless server you shouldn't try to become its host with `--force-external-server`. Did you successfully gain control on the server?

geoffthemedio:
I joined the server from the multiplayer join dialog, so I assume `--force-external-server` was actually irrelevant.

geoffthemedio:
After joining a multiplayer server (running hostless on same machine), I don't get any checksum check messages in the log. What makes it happen?

LGM-Doyle:
For some reason, I did not change the MPLobby checksum to use VerifyCheckSum().  I have fixed that.

I've also generalized the checksum so that the AI and human clients use the same code.

The message received when the checksum is good  is:
````Checksum received from server matches client checksum.````

PR #1789, opened 2017-10-07
Store and parse Species with unique_ptr

### LINKED ISSUE #1547: SpeciesParser should return unique_ptr and not leak memory.
**Environment: all**

**Description**
The `parse/SpeciesParser.cpp` is a generator.  In C++11 a generator should return a `std::unique_ptr<>` to indicate that it has created something that requires memory management and that it is passing management responsibility to the calling function.  The calling function can chose to receive the pointer as a `unique_ptr<>` or transfer it to a `shared_ptr<>` if it needs a shared ownership model.

Furthermore the error condition near line 35 leaks the passed in pointer.  This is a static leak, it only leaks objects that have failed to parse once per game.

**Expected Result**
Return `std::unique_ptr<>`s.

**Steps to reproduce**
Play game.

**Suggestions**

dbenage-cx:
Resolves #1547


LGM-Doyle:
Looks good.

PR #1792, opened 2017-10-07
Expand turn state usage

Morlic-fo:
Some more progress on eventually moving all turn-state related variables into the relevant module rather than having it spread around unrelated modules.

### DIFF
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -1023,7 +1023,7 @@ def generate_production_orders():
             dd_planet = universe.getPlanet(pid)
             if dd_planet:
                 queued_sys.add(dd_planet.systemID)
-        cur_drydoc_sys = set(ColonisationAI.empire_dry_docks.keys()).union(queued_sys)
+        cur_drydoc_sys = set(state.get_empire_drydocks().keys()).union(queued_sys)
         covered_drydoc_locs = set()
         for start_set, dest_set in [(cur_drydoc_sys, covered_drydoc_locs),
                                     (covered_drydoc_locs, covered_drydoc_locs)]:  # coverage of neighbors up to 2 jumps away from a drydock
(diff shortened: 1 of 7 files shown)


default/python/AI/ProductionAI.py
```diff
@@ -315,7 +315,7 @@ def generate_production_orders():
                         cost, time = empire.productionCostAndTime(production_queue[production_queue.size - 1])
                         building_expense += cost / time
 
-            if (total_pp > 40 or ((current_turn > 40) and (ColonisationAI.empire_status.get('industrialists', 0) >= 20))) and ("BLD_INDUSTRY_CENTER" in possible_building_types) and ("BLD_INDUSTRY_CENTER" not in (capital_buildings+queued_building_names)) and (building_expense < building_ratio*total_pp):
+            if (total_pp > 40 or ((current_turn > 40) and (state.population_with_industry_focus() >= 20))) and ("BLD_INDUSTRY_CENTER" in possible_building_types) and ("BLD_INDUSTRY_CENTER" not in (capital_buildings+queued_building_names)) and (building_expense < building_ratio*total_pp):
```
Dilvish-fo, line 318: Seeing these lines makes me think I'd like to add a TODO here, to broaden this somewhat so that it could take into account any captured Industrial Centers (i.e., for completed buildings look not only at the Capital).  It's not really related to the subject of this PR, but if I just go directly add that TODO now it would create a conflict with the merger of this, so I'm noting this comment and you can decide if you want to go ahead and add a commit here for the TODO or else we can just try to remember this later.

Dilvish-fo (approved):
Looks fine to me, my only comment is not really about these changes, but simply about some of the code that is touched.

default/python/AI/ProductionAI.py
Morlic-fo: I pushed a separate commit with the TODO comment

PR #1796, opened 2017-10-07
Let AI character gate usage of growth focus

Morlic-fo:
Seems like something that should be controlled by the character module than some bland module level switch ```useGrowth=True```. 

Typical aggression seems reasonable for using a (somewhat) advanced focus but obviously open for discussion.

### DIFF
--- a/default/python/AI/character/character_module.py
+++ b/default/python/AI/character/character_module.py
@@ -238,6 +238,10 @@ class Trait(object):
         """Return true if allowed to target research/industry > 1.5"""
         return True
 
+    def may_use_growth_focus(self):
+        """Return True if permitted to use growth focus."""
+        return True
+
     def may_travel_beyond_supply(self, distance):  # pylint: disable=no-self-use,unused-argument
         """Return True if able to travel distance hops beyond empire supply"""
         # TODO Remove this tap it is one of the empire id dependent taps. See EmpireIDTrait.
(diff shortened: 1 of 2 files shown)


MatGB:
I think I disagree, it's a relatively basic part of the game that's normally advantageous, players do learn stuff from the AI and seeing it being used and sometimes deciding whether to keep it on a captured planet is useful for less experienced players.

default/python/AI/character/character_module.py
```diff
@@ -379,6 +383,9 @@ def may_dither_focus_to_gain_research(self):
     def may_research_heavily(self):
         return self.aggression > fo.aggression.cautious
 
+    def may_use_growth_focus(self):
+        return self.aggression >= fo.aggression.typical
```
Dilvish-fo: So, probably only changing this occurrence of ```fo.aggression.typical``` to ```fo.aggression.beginner```, and perhaps adding a comment confirming the current intent is to allow for all but make it easy to customize.

Dilvish-fo (commented):
I agree that the  module level switch ```useGrowth=True```  is either better off being handled in the character module, or perhaps is altogether pointless now if we just want to stick with letting all AIs use growth focus.   So, howabout we move it to the character module, but let it stay at default True for all aggression levels, at least for now.  If someone wants to experiment with having Beginner AIs not be able to use growth then it would be easy for them to do that.

Morlic-fo:
Ok, I changed it to allow any aggression level to use it for now.

PR #1799, opened 2017-10-08
Monsert nest eradication

dbenage-cx:
Adjustments and fixes to original monster nest eradicator building.

[forum thread](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10627)


MatGB:
Looks fine to me, untested: it'll need a balance pass when we get onto hulls and monster techs but it'll be easier to balance it when it's been played with a bit.

Vezzra:
Haven't tested, but LGTM. Agree with @MatGB, balancing will be easier once this goes in and players can test it.

Issue #1804, opened 2017-10-09
pycompile returns error for freeOrionAIInterface.py


o01eg:
### Environment
* **FreeOrion Version**: d68433e
* **Operating System**: Ubuntu Server 16.04 LTS
* **Graphic card used**: -
* **Fetched as** 
  * Compiled from source

### Description
`pycompile` couldn't process `freeOrionAIInterface.py` file and returns error:
```
  File "/usr/share/games/freeorion/default/python/interface_mock/result/freeOrionAIInterface.py", line 3373
    None = None  # ruleType(1, "None")
SyntaxError: cannot assign to None
```

### Expected Result
Successfull installation.

### Steps to reproduce
* Enter to `default` directory.
* Execute `python2.7 -m compileall .`
* Issue occurs.

o01eg:
Add test for same issues https://github.com/freeorion/freeorion/pull/1805

Morlic-fo:
Fixed by
https://github.com/freeorion/freeorion/commit/98625c0743266d7fcf9cb70bcb23cef0e2822fa8
https://github.com/freeorion/freeorion/commit/3a0dcf40afe43fbd45d2140c6619c1fcf870f1c9

PR #1806, opened 2017-10-10
Fix "SyntaxError: cannot assign to None" in freeOrionAIInterface.py

geoffthemedio:
Not sure if this is the correct fix for https://github.com/freeorion/freeorion/issues/1804


Morlic-fo:
No.

The relevant file is an autogenerated mock for the interface. The fix is to fix that interface not to manually edit this file.

Interface fix: https://github.com/freeorion/freeorion/commit/98625c0743266d7fcf9cb70bcb23cef0e2822fa8
Mock fix: https://github.com/freeorion/freeorion/commit/98625c0743266d7fcf9cb70bcb23cef0e2822fa8

PR #1802, opened 2017-10-08
Remove int_complex and double_complex parser singletons

LGM-Doyle:

PR #1801, opened 2017-10-08
Remove CommonParamParser singleton.

LGM-Doyle:
This PR starts removing the lexer singleton. It changes references to the lexer singleton to parameters in the `IntValueRefParser` and `DoubleValueRefParser`.  It changes some references to the label singleton to parameters.  It removes the CommonParams singleton parser.

PR #1797, opened 2017-10-07
Change various enum parser singletons into passed parameters.

LGM-Doyle:

PR #1808, opened 2017-10-10
Fix parser bug that prevents concurrent parsing.

LGM-Doyle:
PR #1785 depends on this PR.

This PR contains all of the commits for PRs #1797, #1801, #1802, and #1803 which were created to track down a pair of declarations that causes the MSVC Appveyor CI to fail.

This PR contains all of the fixes to the parser that allow it to correctly concurrently parse files.  It will also enable multi-threaded parsing, which is the goal of #1785. 

It parses the same as the current master branch.

The problem as described in more detail in #1785 is that the lexer is a static singleton that can only be bound to a single file/string for parsing at one time.  

The fix is to make lexers that are non-static, and non-singleton and bind one to each file.

In order to do that this PR finds each singleton parser that is bound to the lexer,  converts it to a grammar, passes it as a parameter to its dependents and then removes the singleton.


parse/ValueRefParser.h
```diff
@@ -1,38 +1,329 @@
...
+    using value_ref_grammar = detail::grammar<value_ref_signature<T>>;
+}
+
+namespace parse { namespace detail {
+
+    using condition_signature      = Condition::ConditionBase* ();
+    using condition_parser_rule    = rule<condition_signature>;
+    using condition_parser_grammar = grammar<condition_signature>;
+
+using name_token_rule = rule<const char* ()>;
```
dbenage-cx: The inconsistent indentation in this file is mildly confusing.

parse/EffectParser.cpp
```diff
@@ -1,28 +1,88 @@
...
-                |    detail::effect_parser_5()
+    effects_group_grammar::effects_group_grammar(
+        const lexer& tok,
+        detail::Labeller& labeller,
+        const detail::condition_parser_grammar& condition_parser,
+        const parse::value_ref_grammar<std::string>& string_grammar
+    ) :
+        effects_group_grammar::base_type(start, "effects_group_grammar"),
+        effects_grammar(tok, labeller, condition_parser, string_grammar)
+        {
```
geoffthemedio: dedent

dbenage-cx (commented):
Couple of formatting notes.

parse/ConditionParser3.cpp
```diff
@@ -10,176 +8,182 @@
...
-                ;
-
-            value_test_1
-                = ('('
-                >> parse::double_value_ref() [ _a = _1 ]
-                >> (    lit("==")   [ _d = Condition::EQUAL ]
+namespace parse { namespace detail {
+    condition_parser_rules_3::condition_parser_rules_3(
+        const parse::lexer& tok,
+        parse::detail::Labeller& labeller,
```
geoffthemedio: Is the `parse::detail::` necessary? This is already in that namespace.

parse/ConditionParser6.cpp
```diff
@@ -24,176 +22,139 @@ namespace std {
...
-
-            focus_type
-                =   tok.Focus_
-                >   (
-                        (parse::detail::label(Type_token) > string_ref_vec [ _val = new_<Condition::FocusType>(_1) ])
+namespace parse { namespace detail {
+    condition_parser_rules_6::condition_parser_rules_6(
+        const parse::lexer& tok,
+        Labeller& labeller,
+        const parse::detail::condition_parser_grammar& condition_parser,
```
geoffthemedio: `parse::detail::` seems to randomly appear or not before `condition_parser_grammar`

parse/EffectParser.h
```diff
@@ -1,23 +1,316 @@
...
+        set_stockpile_or_vis_rule         set_visibility;
+        parse::detail::effect_parser_rule start;
+        planet_type_parser_rules planet_type_rules;
+        planet_size_parser_rules planet_size_rules;
+        parse::empire_affiliation_enum_grammar empire_affiliation_type_enum;
+        parse::set_non_ship_part_meter_enum_grammar set_non_ship_part_meter_type_enum;
+        parse::set_ship_part_meter_enum_grammar set_ship_part_meter_type_enum;
+    };
+
+    struct effect_parser_rules_3 : public effect_parser_grammar {
```
geoffthemedio: these all really need to be in the general effect parser header...?

parse/Parse.cpp
```diff
@@ -581,48 +359,181 @@ namespace parse {
...
-                const lexer& lexer = lexer::instance();
-                label_rule& retval = rules[name];
-                if (PARSING_LABELS_OPTIONAL) {
-                    retval = -(lexer.name_token(name) >> '=');
-                } else {
-                    retval =  (lexer.name_token(name) >> '=');
-                }
-                retval.name(std::string(name) + " =");
-                return retval;
+    double_grammar::double_grammar(const parse::lexer& tok) :
```
geoffthemedio, line 362: what's the distinction in naming between grammars and rules?

LGM-Doyle: A `rule` is a [boost spirit rule](http://www.boost.org/doc/libs/1_65_1/libs/spirit/doc/html/spirit/qi/reference/nonterminal/rule.html)

`rules` are a struct with a bunch of rules as members.  I did not name anything rules, nor did I fix the consistency of naming some things rules and others rule, when they are actually structs containing multiple rules.

A `grammar` is a [boost spirit grammar](http://www.boost.org/doc/libs/1_65_1/libs/spirit/doc/html/spirit/qi/reference/nonterminal/grammar.html), which is a set of rules with one rule designated as the start point.  The rule is designated in the constructor with `X::base_type(designated_start_rule_name)`.  The grammar otherwise, behaves like a simple rule.  Instead of typing `some_rules.special_rule_name`, type `some_grammar`.  Grammars are the built in `boost::spirit` way for composing more sophisticated rules.

The existing parser re-invented the grammar functionality in two places:
- All of the `parse::X()` functions were required to supply a `Rule` which was a structure containing at least one rule called `start`.  I replaced these with grammars, which can designate their own start point. 
- Some of the other rules were sets of with a single starting rule usually called `expr`, which was then referenced by the singleton that exposed those rules as a single rule.  I changed those to a grammar.

Additionally, in a few places there are structs that expose two or more rules as starting points.  I only fixed those that  were exposed as singletons in a way that prevents concurrent parsing.

parse/ValueRefParser.h
LGM-Doyle: I fixed the indentation.

LGM-Doyle (commented):
Thanks for reviewing.

I've made the changes.

I moved two commits that restore the end of file character counting from #1785 to this PR, since I had implied they were already included.  The changes are only in "ParseImpl.h"  `parse_file()`.

parse/EmpireStatsParser.cpp
```diff
@@ -15,9 +16,18 @@ namespace std {
...
 namespace {
-    struct rules {
-        rules(const std::string& filename,
-              const parse::text_iterator& first, const parse::text_iterator& last)
+    using start_rule_payload = std::map<std::string, ValueRef::ValueRefBase<double>*>;
+    using start_rule_signature = void(start_rule_payload&);
+
+    struct grammar : public parse::detail::grammar<start_rule_signature> {
+        grammar(const parse::lexer& tok,
+              const std::string& filename,
```
geoffthemedio: indent params to match

parse/FieldsParser.cpp
```diff
@@ -36,9 +37,20 @@ namespace {
...
 
-    struct rules {
-        rules(const std::string& filename,
-              const parse::text_iterator& first, const parse::text_iterator& last)
+    using start_rule_payload = std::map<std::string, std::unique_ptr<FieldType>>;
+    using start_rule_signature = void(start_rule_payload&);
+
+    struct grammar : public parse::detail::grammar<start_rule_signature> {
+        grammar(const parse::lexer& tok,
+              const std::string& filename,
```
geoffthemedio: indent

parse/FleetPlansParser.cpp
```diff
@@ -18,9 +18,15 @@ namespace std {
...
 namespace {
-    struct rules {
-        rules(const std::string& filename,
-              const parse::text_iterator& first, const parse::text_iterator& last)
+    using start_rule_payload = std::vector<FleetPlan*>;
+    using start_rule_signature = void(start_rule_payload&);
+
+    struct grammar : public parse::detail::grammar<start_rule_signature> {
+        grammar(const parse::lexer& tok,
+              const std::string& filename,
```
geoffthemedio: ...

parse/GameRulesParser.cpp
```diff
@@ -74,9 +74,17 @@ namespace {
...
-    struct rules {
-        rules(const std::string& filename,
-              const parse::text_iterator& first, const parse::text_iterator& last)
+    using start_rule_payload = GameRules;
+    using start_rule_signature = void(start_rule_payload&);
+
+    struct grammar : public parse::detail::grammar<start_rule_signature> {
+        grammar(const parse::lexer& tok,
+                const std::string& filename,
+                const parse::text_iterator& first, const parse::text_iterator& last):
```
geoffthemedio: space b4 `:`

parse/ItemsParser.cpp
```diff
@@ -16,9 +17,16 @@ namespace std {
...
 namespace {
-    struct rules {
-        rules(const std::string& filename,
-              const parse::text_iterator& first, const parse::text_iterator& last)
+    using start_rule_payload = std::vector<ItemSpec>;
+    using start_rule_signature = void(start_rule_payload&);
+
+    struct grammar : public parse::detail::grammar<start_rule_signature> {
+        grammar(const parse::lexer& tok,
+              const std::string& filename,
```
geoffthemedio: ...

parse/ParseImpl.h
```diff
@@ -48,52 +57,71 @@ namespace parse { namespace detail {
...
-    effects_group_rule& effects_group_parser();
+    struct int_grammar : public grammar<int()> {
+        int_grammar(const parse::lexer& tok);
+        rule<int()> int_;
+    };
 
+    using label_rule = rule<>;
+    /** Store label_rules. */
+    class Labeller {
+        public:
```
geoffthemedio: dedent `public:` and similar

geoffthemedio:
Seems to work. Code layout is rather ugly in various places, but I suspect you're not inclined or able to fix it on your own, and it's probably not worth holding things up over.

LGM-Doyle (commented):
The project could adopt a C++ style checker. I've configured and run a few on the code base.

The freeorion C++ style is idiosyncratic.  None of the style checkers I've tried can be configured match its unique style.

The better maintained style checkers run on many editors on all 3 of our supported platforms and/or have easily configured git pre-commit and pre-push hooks.   However, they are less configurable.  The consensus is that with fewer options it is easier to guarantee the source code transformation does not introduce bugs.

Adopting a style checker, would mean there would no longer be a style component to the review process.

PR #1812, opened 2017-10-13
Document FOCS content tag use

dbenage-cx:
The tag use in various species macros are absent, to be done at later date in separate PR/commits.

### DIFF
--- a/default/scripting/ship_parts/Shield/SH_ROBOTIC_INTERFACE_SHIELDS.focs.txt
+++ b/default/scripting/ship_parts/Shield/SH_ROBOTIC_INTERFACE_SHIELDS.focs.txt
@@ -9,13 +9,17 @@ Part
     tags = [ "PEDIA_PC_SHIELD" ]
     location = And [
         OwnedBy empire = Source.Owner
+        /// @content_tag{ROBOTIC} Allows a location to produce this PartType
         HasTag name = "ROBOTIC"
     ]
     effectsgroups =
+        /// @content_tag{ROBOTIC} Ships with this part and ROBOTIC_HULL tag contribute towards increased shields of other ships with this part within 0 starlane jumps
+        /// @content_tag{ROBOTIC_HULL} Ships with this part and ROBOTIC tag contribute towards increased shields of other ships with this part within 0 starlane jumps
         EffectsGroup
             scope = And [
                 Source
                 HasTag name = "ROBOTIC"
+                /// @content_tag{ROBOTIC_HULL} Combined with ROBOTIC tag, allows shield increase from this PartType
                 HasTag name = "ROBOTIC_HULL"
             ]
             activation = ([[ROBOTIC_SHIELD_EFFECT]] >= [[BEST_SHIELD_EFFECT]])
--- a/default/scripting/ship_parts/Troops/GT_TROOP_POD_2.focs.txt
+++ b/default/scripting/ship_parts/Troops/GT_TROOP_POD_2.focs.txt
@@ -9,6 +9,7 @@ Part
     tags = [ "PEDIA_PC_TROOPS" ]
     location = And [
         OwnedBy empire = Source.Owner
+        /// @content_tag{NO_ATTACKTROOPS} Prevents production of ships with this part in their design at locations with this tag
         Not HasTag name = "NO_ATTACKTROOPS"
         Troops low = 4
     ]
--- a/default/scripting/specials/planet/MODERATE_TECH_NATIVES.focs.txt
+++ b/default/scripting/specials/planet/MODERATE_TECH_NATIVES.focs.txt
@@ -8,6 +8,7 @@ Special
         Planet
         Unowned
         Species
+        /// @content_tag{PRIMITIVE} Prevents this special from being added to a location with this tag during universe generation
         Not HasTag name = "PRIMITIVE"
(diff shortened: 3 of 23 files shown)


parse/Parse.cpp
```diff
@@ -29,6 +29,13 @@ namespace std {
 
 namespace parse {
     using namespace boost::xpressive;
+    const sregex CONTENT_TAG_REGEX = bol >> *space >> "#" >> space >> "@content_tag{" >> -*_ >> _n;
```
geoffthemedio: The extra `#` symbol seems unnecessary... why not have the Doxygen in regular comments? That way it can be left in the text during parsing, so that if there is any error that outputs a group of lines surrounding the parse error location, it will include the Doxygen-related text and so be easier to recognize.

default/scripting/techs/learning/PSIONICS.focs.txt
```diff
@@ -5,6 +5,7 @@ Tech
     category = "LEARNING_CATEGORY"
     researchcost = 300 * [[TECH_COST_MULTIPLIER]] - (250 * [[TECH_COST_MULTIPLIER]] * Statistic If condition = AND [
         OwnedBy empire = Source.Owner
+        # @content_tag{TELEPATHIC} Decreases research cost this tech for empires owning any object with this tag
```
geoffthemedio: cost of this tech
for empires that own

default/scripting/specials/planet/MODERATE_TECH_NATIVES.focs.txt
```diff
@@ -8,6 +8,7 @@ Special
         Planet
         Unowned
         Species
+        # @content_tag{PRIMITIVE} Prevents this special at a location with tag
```
geoffthemedio: clarify: prevents this special from being added to a location with this tag during universe generation

default/scripting/ship_parts/Troops/GT_TROOP_POD_2.focs.txt
```diff
@@ -9,6 +9,7 @@ Part
     tags = [ "PEDIA_PC_TROOPS" ]
     location = And [
         OwnedBy empire = Source.Owner
+        # @content_tag{NO_ATTACKTROOPS} Prevents construction for this part at a location with tag
```
geoffthemedio: Prevents production of ships with this part in their design at locations with this tag

default/scripting/ship_parts/Shield/SH_ROBOTIC_INTERFACE_SHIELDS.focs.txt
```diff
@@ -9,13 +9,17 @@ Part
     tags = [ "PEDIA_PC_SHIELD" ]
     location = And [
         OwnedBy empire = Source.Owner
+        # @content_tag{ROBOTIC} Allows a location to construct this PartType
         HasTag name = "ROBOTIC"
     ]
     effectsgroups =
+        # @content_tag{ROBOTIC} Owned Ships with this part and ROBOTIC_HULL tag contribute towards increased shields of other owned ships with this part within 0 starlane jumps
```
geoffthemedio: "owned" ambiguous

default/scripting/ship_parts/Shield/SH_ROBOTIC_INTERFACE_SHIELDS.focs.txt
```diff
@@ -9,13 +9,17 @@ Part
     tags = [ "PEDIA_PC_SHIELD" ]
     location = And [
         OwnedBy empire = Source.Owner
+        # @content_tag{ROBOTIC} Allows a location to construct this PartType
```
geoffthemedio: all "construct" -> "produce"

parse/Parse.cpp
dbenage-cx: Ok, I will see if tag_parser.py can be reworked instead (Doxygen does not parse `//`, and for text files possibly not `/** */`, but I will need to refresh the details)

dbenage-cx:
Ty for reviewing, updated with corrections and changed to using tag parser script.
Rebased

parse/Parse.cpp
geoffthemedio: meaning Doxygen requires the first character on a documentation line to be `#` ...? It can being with `// #` ?

dbenage-cx: `content_tag` is an [aliased](https://github.com/freeorion/freeorion/blob/master/doc/Doxyfile.in#L30) special command ([xrefitem](http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdxrefitem)).
A [docblock](http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html) style that supports special commands is desired.
I chose the triple slash as it is supported by the FOCS parser and does not imply a multi-line style comment as much as `/**  */`.

Doxygen may produce bad links/results if the parsed content does not align favourably with the unmodified source.  Usage of `all_tags.dox` and linking to github source helps alleviate that, but the issue has been temperamental in the past.  It may be possible to further shoehorn that form of comment in, I would rather stick with a documented form.

geoffthemedio: I didn't understand most of that, but if the result is that you put the Doxygen stuff in a standard-ish comment, it's fine, I think.

PR #1813, opened 2017-10-13
Display message time.

o01eg:
This PR adds UTC timestamp to server-side chat messages and display them in the local timezone on client.

### DIFF
--- a/UI/ClientUI.cpp
+++ b/UI/ClientUI.cpp
@@ -52,6 +52,8 @@
 #include <boost/filesystem/fstream.hpp>
 #include <boost/filesystem/operations.hpp>
 #include <boost/system/system_error.hpp>
+#include <boost/date_time/posix_time/posix_time_io.hpp>
+#include <boost/date_time/c_local_time_adjustor.hpp>
 
 #include <string>
 #include <algorithm>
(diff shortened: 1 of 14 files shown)


geoffthemedio:
Should probably have an option to enable / disable this...

o01eg:
Add options `UI.display-timestamp` and `UI.display-timestamp-format`.

UI/ClientUI.cpp
```diff
@@ -651,6 +656,11 @@ ClientUI::ClientUI() :
 
     // Set the root path for image tags in rich text.
     GG::ImageBlock::SetDefaultImagePath(ArtDir().string());
+
+    // Set facet to format timestamp in chat.
+    boost::posix_time::time_facet* facet = new boost::posix_time::time_facet();
+    facet->format(ClientUI::DisplayTimestampFormat().c_str());
+    m_date_format_sstream.imbue(std::locale(std::locale(""), facet));
```
geoffthemedio: does this call to `imbue` take ownership of `facet`?

o01eg: Yes, it takes:
http://www.cplusplus.com/reference/locale/locale/locale/
> The constructed locale object takes over responsibility for deleting this facet object.

o01eg:
Rebased, allowed UI to show messages without sender.

o01eg:
Rebased. Removed `UI.display-timestamp-format`.

geoffthemedio:
MPLobby chat layout is a bit awkward with the lengthy timestamps before the chat messages...
mplobby_chat_with_timestamps

o01eg:
I increased width of window and chat box.

geoffthemedio:
Can you add an OptionsWnd checkbox to turn the timestamp on / off?

o01eg:
Added in UI section.

UI/OptionsWnd.cpp
```diff
@@ -522,6 +522,7 @@ void OptionsWnd::CompleteConstruction() {
     BoolOption(current_page, 0, "UI.window-quickclose",          UserString("OPTIONS_QUICK_CLOSE_WNDS"));
     BoolOption(current_page, 0, "UI.sidepanel-planet-shown",     UserString("OPTIONS_SHOW_SIDEPANEL_PLANETS"));
     BoolOption(current_page, 0, "UI.auto-reposition-windows",    UserString("OPTIONS_AUTO_REPOSITION_WINDOWS"));
+    BoolOption(current_page, 0, "UI.display-timestamp",          UserString("OPTIONS_DB_DISPLAY_TIMESTAMP"));
```
geoffthemedio: The options DB entry is too verbose for the option label. It works as the tooltip, but the options UI label should be short, like just "Chat Timestamps". Have a separate stringtable entry in the game options section and use that for the extra label...

https://github.com/freeorion/freeorion/blob/master/default/stringtables/en.txt#L2325

o01eg: Fixed.

default/stringtables/en.txt
```diff
@@ -2340,6 +2343,9 @@ Show side panel planets
 OPTIONS_AUTO_REPOSITION_WINDOWS
 Automatically reposition windows
 
+OPTIONS_DISPLAY_TIMESTAMP
+Show chat timestamp.
```
geoffthemedio: no `.` (as with similar strings)

o01eg: Removed.

PR #1818, opened 2017-10-17
Remove level of if nesting in  GUI::RenderWindow()

LGM-Doyle:
The reason I pushed this for review is that the change prevents WriteWndtoPNG from writing a non-visible window.  I could not determine if writing a non-visible window was intentional.

### DIFF
--- a/GG/src/GUI.cpp
+++ b/GG/src/GUI.cpp
@@ -1615,47 +1615,45 @@ void GUI::RenderWindow(const std::shared_ptr<Wnd>& wnd)
 
 void GUI::RenderWindow(Wnd* wnd)
 {
-    if (!wnd)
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I don't see why you'd want to prevent a non-visible window from being written... might be useful.

LGM-Doyle:
The changes make the code agree with the existing description.  So as coded it was a bug.

 ````
    /** Saves \a wnd to file \a filename during the next render cycle.  If \a
        wnd is not rendered during the render cycle, or PNG support is not
        enabled, this is a no-op. */
    void SaveWndAsPNG(const Wnd* wnd, const std::string& filename) const;
 ````

geoffthemedio:
OK then.

PR #1819, opened 2017-10-17
Improve FleetWnd ship panel first render.

LGM-Doyle:
This PR improves the rendering performance of `FleetWnd`

This PR split `RequirePreRender()` in the `FleetWnd` into `RequirePreRender()` for layout updates and `RequireRefresh()` for data updates.  

Secondly, it fixes the initial render of the ship data panel when the fleet id id known to render correctly on the first frame after construction.

PR #1820, opened 2017-10-17
Generalize dismiss ghost.

LGM-Doyle:
This PR depends on the changes in #1819.

It generalizes dismiss ghost.

- It adds `MapWnd::ForgetObject()` as a general purpose forget object function.  
- It then replaces the exising uses on the `MapWnd` and the fleets section of the `FleetWnd` with the call to `MapWnd::ForgetObject()`.  
- It adds a dismiss ghost option to the ship portion of the `FleetWnd` so that all locations where a player can see a ghost they can also dismiss it.

### DIFF
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1746,7 +1746,7 @@ void Universe::ForgetKnownObject(int empire_id, int object_id) {
         ForgetKnownObject(empire_id, child_id);
 
     if (int container_id = obj->ContainerObjectID() != INVALID_OBJECT_ID) {
-        if (std::shared_ptr<UniverseObject> container = objects.Object(container_id)) {
+        if (auto container = objects.Object(container_id)) {
             if (auto system = std::dynamic_pointer_cast<System>(container))
                 system->Remove(object_id);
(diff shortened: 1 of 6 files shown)


LGM-Doyle:
Rebased.

I also added a fix to a bug existing in current master, where it was possible for a fleet to be removed but not its ship(s) if after it was detected as partially visible it moved completely outside of the empire's detection range.

universe/Universe.cpp
```diff
@@ -1740,19 +1740,22 @@ void Universe::ForgetKnownObject(int empire_id, int object_id) {
         return;
     }
 
-    for (int child_id : obj->VisibleContainedObjectIDs(empire_id)) {
-        if (std::shared_ptr<UniverseObject> child = objects.Object(child_id))
-            ForgetKnownObject(empire_id, child->ID());
-    }
+    // Remove all contain objects to avoid breaking fleet+ship, system+planet invariants
```
geoffthemedio: contain -> containing or contained ?

geoffthemedio:
otherwise OK

Issue #1824, opened 2017-10-23
FOCS Parser becomes very slow for nested parentheses


Morlic-fo:
### Environment
* **FreeOrion Version**:  test build 2017-10-23.0ac4a07
* **Operating System**: Win 7 / 64 bit
* **Fetched as**
  * Weekly development build

### Description
The FOCS parser takes a *very* long time to parse files that have a high level of nested parentheses. In current content, this is most noticeable for the experimentor outpost.

### Steps to reproduce
* Select an arbitrary focs file (e.g. default/scripting/buildings/EXPERIMENTOR_OUTPOST.focs.txt)
* Add parentheses around existing parentheses to increase number of nesting. For example,
```
EXPERIMENTOR_SPAWN_START_TURN
'''(((((((((([[EXPERIMENTOR_SPAWN_CALC_A]] * (1.2 - ( GalaxyPlanetDensity / 10 )) * ((Max(1, GalaxySize / 200))^0.4)))))))))))'''
```
* Start a game
* Check logs and compare time to parse the file.
* If the game still started in an acceptable amount of time, add another pair of parentheses to increase the nesting level and compare the timing to parse the file.

Result of above code:
```
23:40:00.742046 [debug] server : Pending.h:53 : Waiting for parse of "buildings" to complete.
23:40:01.742103 [debug] server : Pending.h:53 : Waiting for parse of "buildings" to complete.
(... some hundred lines of this...)
23:48:50.012319 [debug] server : Pending.h:53 : Waiting for parse of "buildings" to complete.
23:48:51.022376 [debug] server : Pending.h:53 : Waiting for parse of "buildings" to complete.
```
and still going strong after 9 minutes

geoffthemedio:
Can you find an earlier version where this was not a problem?

LGM-Doyle:
@Morlic-fo thanks for reporting this.

@geoffthemedio I checked back to at least May 2017 and the same problem exists.

Part of the problem is that when the spirit parser backtracks it leaks memory.  Adding parentheses, increases the backtracking, more memory leaks and more time is consumed.  You can verify this by checking the amount of memory used by the server, client and AI processes as you increase the number of redundant parentheses.

I had a work in progress branch fixing the memory leaks in the parser, but this bug allowed me to see that my approach was wrong.  I've started over and it may take some time to fix the leaks completely.

I can't tell if the leaks are masking an additional problem with the parentheses nesting.

Morlic-fo:
Fixed by #1844

PR #1825, opened 2017-10-23
Speed up parsing of EXPERIMENTOR_OUTPOST.focs.txt

Morlic-fo:
In current master, parsing the experimentor outpost takes around 24 seconds on my system.

This PR removes (hopefully) redundant parentheses to work around the performance issue described in #1824. The effect is that the file parses in only 1.7 seconds. The game starts significantly faster after this bottleneck is removed.

### DIFF
--- a/default/scripting/buildings/EXPERIMENTOR_OUTPOST.focs.txt
+++ b/default/scripting/buildings/EXPERIMENTOR_OUTPOST.focs.txt
@@ -78,7 +78,7 @@ BuildingType
                 Not OwnedBy affiliation = AnyEmpire
             ]
             activation = Or [
-                Turn high = [[EXPERIMENTOR_SPAWN_START_TURN]]      // always remove lanes before spawn start turn
+                Turn high = ([[EXPERIMENTOR_SPAWN_START_TURN]])      // always remove lanes before spawn start turn
                 Not ContainedBy Contains And [ Monster Unowned Not Design name = "SM_EXP_OUTPOST"]
             ]
             effects = RemoveStarlanes endpoint = WithinStarlaneJumps jumps = 1 condition = Source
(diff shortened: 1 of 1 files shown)


default/scripting/buildings/EXPERIMENTOR_OUTPOST.focs.txt
```diff
@@ -78,15 +78,15 @@ BuildingType
                 Not OwnedBy affiliation = AnyEmpire
             ]
             activation = Or [
-                Turn high = [[EXPERIMENTOR_SPAWN_START_TURN]]      // always remove lanes before spawn start turn
+                Turn high = ([[EXPERIMENTOR_SPAWN_START_TURN]])      // always remove lanes before spawn start turn
```
geoffthemedio, line 81: This seems to be doing the opposite of what you say in the summary...

Morlic-fo: The same commit removes the parentheses from the macro itself so no nesting level is added here.
I wasn't completely sure if the parentheses are needed in places like this. If they aren't, they should be removed as well.

geoffthemedio: I don't see why they would be... the macro should be a complete expression without them. If it parses and runs without error without them, then leave them off.

geoffthemedio:
Seems to work for me...
````
00:41:08.194450 [debug] timer : ScopedTimer.cpp:34 : parse_file "EXPERIMENTOR_OUTPOST.focs.txt" time:    10662 ms
[...]
00:45:02.163627 [debug] timer : ScopedTimer.cpp:34 : parse_file "EXPERIMENTOR_OUTPOST.focs.txt" time:      664 ms
````

default/scripting/buildings/EXPERIMENTOR_OUTPOST.focs.txt
geoffthemedio: In my tests, it doesn't parse properly without the brackets. Not sure why, but apparently they are needed.

PR #1830, opened 2017-10-29
Fix compiler warning on MSVC.

wheals:
See https://ci.appveyor.com/project/freeorion/freeorion/build/ci-2749#L1100 for example of the warning. 

Ironically this was introduced in #1706, while fixing other warnings.

### DIFF
--- a/universe/UniverseGenerator.cpp
+++ b/universe/UniverseGenerator.cpp
@@ -170,7 +170,7 @@ namespace Delauney {
         triangle_list.push_front({num_points_in_vec - 1, num_points_in_vec - 2,
                                   num_points_in_vec - 3, points_vec});
 
-        if (points_vec.size() > std::numeric_limits<int>::max()) {
+        if (points_vec.size() > static_cast<size_t>(std::numeric_limits<int>::max())) {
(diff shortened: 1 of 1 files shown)


universe/UniverseGenerator.cpp
```diff
@@ -170,7 +170,7 @@ namespace Delauney {
         triangle_list.push_front({num_points_in_vec - 1, num_points_in_vec - 2,
                                   num_points_in_vec - 3, points_vec});
 
-        if (points_vec.size() > std::numeric_limits<int>::max()) {
+        if (points_vec.size() > static_cast<size_t>(std::numeric_limits<int>::max())) {
```
geoffthemedio, line 173: is it safe to assume int max is always equal to or smaller than the largest possible size_t value?

wheals: Even if it isn't, this cast will not change the result of this comparison.
If INT_MAX is greater than the largest possible size_t, then, before adding the cast, the comparison will always return false, no matter how big `points_vec` is. After adding the cast, it's checking whether a particular size_t is bigger than the biggest possible size_t, which will also always be false.

PR #1834, opened 2017-10-31
Role-Based Authorization

o01eg:
Start of implementation role-based authorization.

Roles are property of connections.
Multiplayer lobby window use it to enable/disable stuff.

Forum thread: http://freeorion.org/forum/viewtopic.php?f=9&t=10737

### DIFF
--- a/UI/MultiplayerLobbyWnd.cpp
+++ b/UI/MultiplayerLobbyWnd.cpp
@@ -164,21 +164,21 @@ namespace {
                     int row_moderator_type = -1;
                     int row_number = 0;
                     if (client_type == Networking::CLIENT_TYPE_HUMAN_PLAYER ||
-                        ClientApp::GetApp()->Networking().HasAuthRole(Networking::ROLE_PLAYER))
+                        ClientApp::GetApp()->Networking().HasAuthRole(Networking::ROLE_CLIENT_TYPE_PLAYER))
                     {
                         Insert(GG::Wnd::Create<TypeRow>(w, type_row_height, Networking::CLIENT_TYPE_HUMAN_PLAYER));   // "Human" display / option
                         row_player_type = (row_number++);
                     }
 
                     if (client_type == Networking::CLIENT_TYPE_HUMAN_OBSERVER ||
-                        ClientApp::GetApp()->Networking().HasAuthRole(Networking::ROLE_OBSERVER))
+                        ClientApp::GetApp()->Networking().HasAuthRole(Networking::ROLE_CLIENT_TYPE_OBSERVER))
                     {
                         Insert(GG::Wnd::Create<TypeRow>(w, type_row_height, Networking::CLIENT_TYPE_HUMAN_OBSERVER)); // "Observer" display / option
                         row_observer_type = (row_number++);
                     }
 
                     if (client_type == Networking::CLIENT_TYPE_HUMAN_MODERATOR ||
-                        ClientApp::GetApp()->Networking().HasAuthRole(Networking::ROLE_MODERATOR))
+                        ClientApp::GetApp()->Networking().HasAuthRole(Networking::ROLE_CLIENT_TYPE_MODERATOR))
                     {
                         Insert(GG::Wnd::Create<TypeRow>(w, type_row_height, Networking::CLIENT_TYPE_HUMAN_MODERATOR));// "Moderator" display / option
                         row_moderator_type = (row_number++);
--- a/UI/MultiplayerLobbyWnd.cpp
+++ b/UI/MultiplayerLobbyWnd.cpp
@@ -761,16 +761,16 @@ void MultiPlayerLobbyWnd::LobbyUpdate(const MultiplayerLobbyData& lobby_data) {
 
     bool send_update_back = PopulatePlayerList();
 
-    if (send_update_back && (ThisClientIsHost() || m_lobby_data.m_any_can_edit))
+    if (send_update_back && HasAuthRole(Networking::ROLE_GALAXY_SETUP))
         SendUpdate();
 
     LogPlayerSetupData(m_lobby_data.m_players);
 }
 
 void MultiPlayerLobbyWnd::Refresh() {
-    m_any_can_edit->Disable(!ThisClientIsHost());
+    m_any_can_edit->Disable(!HasAuthRole(Networking::ROLE_HOST));
 
-    if (ThisClientIsHost() || m_lobby_data.m_any_can_edit) {
+    if (HasAuthRole(Networking::ROLE_GALAXY_SETUP)) {
         for (std::size_t i = 0; i < m_new_load_game_buttons->NumButtons(); ++i)
             m_new_load_game_buttons->DisableButton(i, false);
         m_galaxy_setup_panel->Disable(false);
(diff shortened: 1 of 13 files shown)


o01eg:
I suppose I finished functionality for this PR.

UI/MultiplayerLobbyWnd.cpp
```diff
@@ -159,19 +159,40 @@ namespace {
...
                 } else {
                     // For human players on host, have "Player", "Observer", and "Drop" options.  TODO: have "Ban" option.
-                    Insert(GG::Wnd::Create<TypeRow>(w, type_row_height, Networking::CLIENT_TYPE_HUMAN_PLAYER));   // "Human" display / option
-                    Insert(GG::Wnd::Create<TypeRow>(w, type_row_height, Networking::CLIENT_TYPE_HUMAN_OBSERVER)); // "Observer" display / option
-                    Insert(GG::Wnd::Create<TypeRow>(w, type_row_height, Networking::CLIENT_TYPE_HUMAN_MODERATOR));// "Moderator" display / option
+                    int row_player_type = -1;
+                    int row_observer_type = -1;
+                    int row_moderator_type = -1;
+                    int row_number = 0;
+                    if (client_type == Networking::CLIENT_TYPE_HUMAN_PLAYER ||
```
geoffthemedio, line 166: What's the logic with the `||` here? Incomplete transition, or will there be player clients that don't have the player role...?

network/ClientNetworking.cpp
```diff
@@ -285,6 +292,10 @@ bool ClientNetworking::Impl::PlayerIsHost(int player_id) const {
     return player_id == m_host_player_id;
 }
 
+bool ClientNetworking::Impl::HasAuthRole(Networking::RoleType role) const {
+    return m_roles.HasRole(role);
```
geoffthemedio: for single line functions, prefer `{ /* stuff */ }` all on one line below signature line.

UI/MultiplayerLobbyWnd.cpp
o01eg: Yes. It's a flexible system for any type of authorization configurations. Something like service account only to moderate games or where only authenticated players could play game and other could be only observers...

geoffthemedio: Doesn't really address the question... why do you check both client type and roles set for it? When/why would an observer connection have the player role, and why wouldn't an observer connection have the observer role, for example? Presumably you want to get rid of the observer and moderator connection types and just have human connections that have various roles assigned...?

server/ServerFSM.cpp
```diff
@@ -360,6 +367,13 @@ sc::result Idle::react(const HostSPGame& msg) {
     server.m_networking.SetHostPlayerID(host_player_id);
     if (!GetOptionsDB().Get<bool>("skip-checksum"))
         player_connection->SendMessage(ContentCheckSumMessage());
+    player_connection->SetAuthRoles({
+                    Networking::ROLE_HOST,
+                    Networking::ROLE_MODERATOR,
```
geoffthemedio: single player players aren't moderators or observers

server/ServerFSM.cpp
```diff
@@ -468,6 +482,13 @@ MPLobby::MPLobby(my_context c) :
                     player_setup_data.m_starting_species_name = sm.SequentialPlayableSpeciesName(player_id);
 
                 m_lobby_data->m_players.push_back(std::make_pair(player_id, player_setup_data));
+
+                player_connection->SetAuthRoles({
+                                Networking::ROLE_MODERATOR,
```
geoffthemedio: most players aren't moderators or observers

UI/MultiplayerLobbyWnd.cpp
o01eg: I check both because it possible when host sets client type to player which doesn't have a role for it.
I don't plan to get rid of client types. It is used for current type of connection while roles determine possible set of types.

server/ServerFSM.cpp
o01eg: Currently they can set yourself to Observer or Moderator so I just leave it.

server/ServerFSM.cpp

network/ClientNetworking.cpp
o01eg: Fixed.

UI/MultiplayerLobbyWnd.cpp
geoffthemedio: I don't understand. How would a player not have a role "for" a client type...?

Roles determine possible set(s) of types...? I thought the roles determined what a connection could do in game / lobby?

server/ServerFSM.cpp
geoffthemedio: So the `SetAuthRoles` function is setting the allowed roles, not the current roles? If so, rename it.

And perhaps call them permissions, not roles?

UI/MultiplayerLobbyWnd.cpp
o01eg: It could be if player get his client type from host. He could not do so itself because he doesn't have a role allows it but host have a role allows set client type for everyone.
Roles are what is allowed to do so roles for client types determines which client types player allowed to set.

server/ServerFSM.cpp
o01eg: They are current roles. Player can set itself to Moderator, Observer or Player in any time so he has ROLE_MODERATOR, ROLE_OBSERVER and ROLE_PLAYER.

UI/MultiplayerLobbyWnd.cpp
geoffthemedio: Then perhaps rename some roles to CAN_SET_SELF_TO_OBSERVER or similar.

o01eg: I could also add check before starting game and prevent it if some players doesn't have a roles to be allowed to use their client type.

server/ServerFSM.cpp
geoffthemedio: If that's how it works, then the term "role" is confusing.

o01eg:
Add restrictions so even host cann't set client type for player if he doesn't have a appropriate role.

geoffthemedio:
Please call them "permissions" or "allowed client types" or similar. "role" is confusing / misleading.

o01eg:
In terms of RBAC they are called "roles".

Maybe "groups" because it's not about client types only and "permissions" I've heard about objects, files and so on.

geoffthemedio:
I don't know what RBAC is. If this is a standard name for this concept, can you point me to some documentation of that?

o01eg:
@geoffthemedio Yep, there is basic: https://en.wikipedia.org/wiki/Role-based_access_control

geoffthemedio:
Looking at that page, what is currently a "client type" is the "active role" of a user, and those give permissions to the user such as moderator powers, editing galaxy settings, or control of an empire. What you are calling "roles" are "authorized roles".

So, I still think you should call them "allowed client types".

o01eg:
@geoffthemedio It have a poor definition what "active role" is so I suppose it which role test against current permission.

> So, I still think you should call them "allowed client types".

Ok, should I just rename appropriate roles or change something else?

geoffthemedio:
If every role is allowing the player to have a single client type, then calling them "allowed client types" in the code makes sense to me.

o01eg:
Rebased and renamed roles from `ROLE_PLAYER` to `ROLE_CLIENT_TYPE_PLAYER` and so on.

network/Networking.cpp
```diff
@@ -20,4 +20,26 @@ namespace Networking {
...
     { return GetOptionsDB().Get<int>("network.message-port"); }
+
+    AuthRoles::AuthRoles(const std::initializer_list<RoleType>& roles) {
+       for (RoleType r : roles) {
+           m_roles.set(r, true);
+       }
+    }
+
+    void AuthRoles::SetRole(RoleType role, bool value) {
+        m_roles.set(role, value);
```
geoffthemedio: for these single-line functions, put the `{}` on the same line as the single line of code.

network/Message.cpp
```diff
@@ -665,6 +665,10 @@ Message AuthResponseMessage(const std::string& player_name, const std::string& a
     return Message(Message::AUTH_RESPONSE, os.str());
 }
 
+Message SetAuthorizationRolesMessage(const Networking::AuthRoles& roles) {
+    return Message(Message::SET_AUTH_ROLES, roles.Text());
```
geoffthemedio: {} to same line as line of code...

geoffthemedio:
Otherwise seems OK

network/Networking.cpp
o01eg: Fixed.

network/Message.cpp
o01eg: Fixed.

o01eg:
Rebased to fix conflicts.

Issue #792, opened 2016-07-24
Missing support for alternate planet landscapes


Deepsloth:
The commit https://github.com/freeorion/freeorion/commit/2b3c363139f95df7e66fe89ea7af2e82dc5e201e added alternate planet landscapes to be used in the planet suitability pedia entries. Support for alternate images hasen't been implemented yet.

Discussion: http://www.freeorion.org/forum/viewtopic.php?f=10&t=9730&start=21

wheals:
Can be closed now.

Vezzra:
Implemented by #1835.

PR #1835, opened 2017-10-31
Implement support for alternate landscapes (#792).

wheals:
Tested in-game and was able to get both Inferno images, and it stayed the same between runs for the same planet.

### DIFF
--- a/UI/EncyclopediaDetailPanel.cpp
+++ b/UI/EncyclopediaDetailPanel.cpp
@@ -2426,28 +2426,34 @@ namespace {
         auto planet = GetPlanet(planet_id);
 
         // show image of planet environment at the top of the suitability report
-        std::string planet_type = boost::lexical_cast<std::string>(planet->Type());
+        const auto type = planet->Type();
+        std::string planet_type = boost::lexical_cast<std::string>(type);
         boost::algorithm::to_lower(planet_type);
 
         namespace fs = boost::filesystem;
-        std::vector<std::string> filenames;
-        fs::directory_iterator end_it;
-        for (fs::directory_iterator it(ClientUI::ArtDir() / "encyclopedia" / "planet_environments"); it != end_it; ++it) {
-            try {
-                if (fs::exists(*it) && fs::is_regular_file(it->status())) {
-                    auto filename = it->path().filename().string();
-                    if (boost::algorithm::starts_with(filename, planet_type)) {
-                        filenames.push_back(filename);
+        // Cache searches through the planet_environments directory
+        static std::unordered_map<PlanetType, std::vector<std::string>> filenames_by_type;
+        // Only search once per execution
+        if (!filenames_by_type.count(type)) {
+            fs::directory_iterator end_it;
+            for (fs::directory_iterator it(ClientUI::ArtDir() / "encyclopedia" / "planet_environments"); it != end_it; ++it) {
+                try {
+                    if (fs::exists(*it) && fs::is_regular_file(it->status())) {
+                        auto filename = it->path().filename().string();
+                        if (boost::algorithm::starts_with(filename, planet_type)) {
+                            filenames_by_type[type].push_back(filename);
+                        }
                     }
                 }
-            }
-            catch (const fs::filesystem_error& e) {
-                // ignore files for which permission is denied, and rethrow other exceptions
-                if (e.code() != boost::system::posix_error::permission_denied)
-                    throw;
+                catch (const fs::filesystem_error& e) {
+                    // ignore files for which permission is denied, and rethrow other exceptions
+                    if (e.code() != boost::system::posix_error::permission_denied)
+                        throw;
+                }
             }
         }
-        if (!filenames.empty()) {
+        if (filenames_by_type.count(type)) {
+            const auto& filenames = filenames_by_type[type];
             detailed_description += "<img src=\"encyclopedia/planet_environments/" + filenames[planet_id % filenames.size()] + "\"></img>";
         }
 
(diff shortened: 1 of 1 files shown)


UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2428,7 +2428,27 @@ namespace {
...
         boost::algorithm::to_lower(planet_type);
-        detailed_description += "<img src=\"encyclopedia/planet_environments/" + planet_type + ".png\"></img>";
+
+        namespace fs = boost::filesystem;
+        std::string filename;
+        int matches = 0;
+        fs::directory_iterator end_it;
+        for (fs::directory_iterator it(ClientUI::ArtDir() / "encyclopedia" / "planet_environments"); it != end_it; ++it) {
+            try {
+                if (fs::exists(*it) && !fs::is_directory(*it) && boost::algorithm::starts_with(it->path().filename().string(), planet_type)) {
```
geoffthemedio: is this safe for Windows wide character filenames?

wheals: This is pretty much the same code as [PrefixedTexturesAndDist](https://github.com/freeorion/freeorion/blob/d8164191fb4a800a0ed206393ebb300f36d29ba0/UI/ClientUI.cpp#L1121), so if that works, this should work too.

Note that it's only comparing the filename portion of the path, so unless there's a planet type with non-ASCII in its name, the problem should never even come up (it doesn't matter if the directory is located in a user directory with Cyrillic characters, for example).

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2428,7 +2428,27 @@ namespace {
...
+                    ++matches;
+                    if (planet_id % matches == 0) {
```
geoffthemedio: Could you add some comments explaining what this code is doing? This part in particular I don't understand... you're incrementing the number of matches and then checking if the planet ID is divisible by that number... then doing so again for any additional files found? Seems like this could be simplified by looking up the files once per execution.

Is this intended to achieve a particular distribution of the possible images based on the number of factors each planet ID number has...?

wheals: The idea is to pick a random file, by hashing the planet id, using reservoir sampling. That said, it would be simpler and clearer to just use a vector and then pick a random entry, so I'll change it to do that.

wheals:
Think the code is clearer now.

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2428,7 +2428,28 @@ namespace {
...
         std::string planet_type = boost::lexical_cast<std::string>(planet->Type());
         boost::algorithm::to_lower(planet_type);
-        detailed_description += "<img src=\"encyclopedia/planet_environments/" + planet_type + ".png\"></img>";
+
+        namespace fs = boost::filesystem;
+        std::vector<std::string> filenames;
+        fs::directory_iterator end_it;
+        for (fs::directory_iterator it(ClientUI::ArtDir() / "encyclopedia" / "planet_environments"); it != end_it; ++it) {
+            try {
+                if (fs::exists(*it) && !fs::is_directory(*it)) {
```
dbenage-cx: `!is_directory(*it)` -> `is_regular_file(it->status())`
(other types than just directory and file, iteration may provide stat as by-product)

The entire iteration should probably be enclosed in try/catch

wheals: If the entire for loop is enclosed in try/catch, then permission denied errors would result in all files after it being skipped, instead of just the file for which permission was denied

dbenage-cx: Guessing that is already an issue during iteration?
There are various other instances of directory iterations without a try/catch currently, so probably not a huge concern with an additional case (hope to eventually consolidate filesystem operations, but slow going).

wheals: I don't believe it's a big issue either way, but I'm not too knowledgable here. Like I said this is basically the same code as `ClientUI::PrefixedTexturesAndDist`, so I figure if it's good enough for that, it's good enough here too. Or at least, pending a consolidation of filesystem operations like you're hoping for.

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -2428,7 +2428,28 @@ namespace {
         // show image of planet environment at the top of the suitability report
         std::string planet_type = boost::lexical_cast<std::string>(planet->Type());
         boost::algorithm::to_lower(planet_type);
-        detailed_description += "<img src=\"encyclopedia/planet_environments/" + planet_type + ".png\"></img>";
+
+        namespace fs = boost::filesystem;
+        std::vector<std::string> filenames;
+        fs::directory_iterator end_it;
```
geoffthemedio: make `filenames` static, and wrap this all in `if (filenames.empty()) {` so that it only searches once per execution?

wheals: Done (though it was a bit more complicated than just making it static, since it has to cache the data for every planet type).

geoffthemedio:
Appears to have legit build errors.

Otherwise works for me on Windows.

wheals:
Hopefully this fixes it. It would be nice if the GG_ENUM macro made a std::hash overload in such a way that the enum could be used as a key directly in an unordered_map, but this should work as well.

wheals:
Looks like it worked.

geoffthemedio:
Could you implement such a hash function in the macro?

wheals:
I didn't want to try it because I don't have a way to compile on g++ or clang (which is the obvious way to test if it works), but I could just use Travis builds to make sure it's working.

geoffthemedio:
Can do as a separate pull request if you prefer...

wheals:
I think that'd be a better idea; implementing the change looks like it might be beyond my level of C++ expertise, and it's not necessary for the functionality here.

PR #1836, opened 2017-11-02
Fix AI enqueuing too many scouts when loading a game on turn 1

Morlic-fo:
The AI would enqueue a 3 scouts at game start and, once loading a game on turn 1, would add another 3 to the queue.

This PR fixes that bug by making sure the production queue is empty before enqueuing the scouts.


Dilvish-fo (approved):
sure, looks good to me.

PR #1838, opened 2017-11-04
Fix parser memory leaks

LGM-Doyle:
This PR fixes memory leaks in the parser.

***Description of Problem***
There are two problems which interact to cause these memory leaks:  `boost::spirit` is not designed to work well with dynamic polymorphism and `boost::phoenix` actors expect value semantics and do not handle move semantics.

The `boost::spirit` parsing system is not designed to handle dynamic polymorphism.  When semantic actions allocate memory on the heap and store it in actors, it leaks the memory allocated in any branch that backtracks.  This occurs even with a successful parse. For example, the following parser will always leak the memory allocated in the first unsuccessful branch.

````
a_leaky_parser = 
    unsuccessful_parser [ _val = new_<LeakedObject>()]
    | successful_parser [ _val = new_<GoodObject>()];
````

The simple fix of making `_val` a `unique_ptr` fails because the `boost::phoenix` actors do not handle C++11 move semantics.  rvalue references will eventually be supported in `spirit`, but that will at least require upgrading to spirit V3 and that requires C++14.

***This Solution***
This PR defines `MovableEnvelope<T>` which wraps a `unique_ptr<T>` with copy mechanics that move the pointer each time it is copied.  This fixes the leaks.  This adds the constraint that the wrapped object once created must move monotonically through the parser.

When/if we adopted a version of `spirit` that supports move mechanics, then `MovableEnvelope` can be replaced with `unique_ptr`.

***Testing***
Testing is straightforward.  Start a game and compare the memory use of this PR with current master.

My results are about 20MB of memory saved per executable:

executable | master | this PR | memory saved
------------- | -------- | -- | --
freeorion | 178 MB | 157 MB | 21 MB
freeoriond | 98 MB | 72 MB | 26 MB
freeorionca | 79 MB | 53 MB | 26 MB

An additional test is to replicate issue #1824 and add the redundant parentheses.  In current master, each set of redundant parentheses adds additional leaked memory.  With this PR there is no increase in memory used.  This PR does not improve the time taken to resolve the redundant parentheses, since the memory thrashing still happens.  It just does not leak.

***Overview of the PR***
The PR is most easily understood by looking at "parse/MovableEnvelope.h" which creates 4 things: `MovableEnvelope<T>`, `OpenEnvelope(s)()`, `construct_movable` and `deconstruct_movable`.

`MovableEnvelope<T>` wraps a `unique_ptr<T>` replacing the copy constructors and assignment operators with variants that move the wrapped `unique_ptr<T>` from the source to the destination instead of copying it.

`OpenEnvelope(s)()` performs a one-time extraction of the wrapped `unique_ptr<T>` so that it can be passed to the client that called the parsing sub-system.

`construct_movable` and `deconstruct_movable` are functors so that `boost::phoenix` actors can be created that work in a straightforward manner in the `spirit` parser framework.


Dilvish-fo:
Sounds great.

> This PR does not improve the time taken to resolve the redundant parentheses, since the memory thrashing still happens

Hmm, I do think that there is something going on though besides just the "same-old" memory thrashing the parser has had forever.  I haven't examined it closely enough to be absolutely sure or figure out when it started, but it certainly does feel to me that lately it takes the AI's a lot longer than it used to for them to get their War declarations off, and it seems like the parallel parsing changes you put in should have helped that rather than hurt it.

LGM-Doyle:
One specific thing that was slowing startup was the recent addition of CheckSums.  Computing these on the server and then the client adds delay.  I added [this](https://github.com/freeorion/freeorion/pull/1788) logic to not do the check sums for single player games.  That was only merged two weeks ago and only improves single player games.

LGM-Doyle:
This PR is now passing checks on all test platforms.

The code compiles successfully for `deconstruct_movable` is different on all of the older supported compiler versions on the test machines than on my machine.  This is because of their less robust `decltype` handling.

The PR should be tested on all platforms.  To test, quickstart a game.  If the game starts and there are no unusual errors in the console or logs, then the parser succeeded and the test is successful.

geoffthemedio:
I can't test this due to merge conflicts with the Xcode/FreeOrion.xcodeproj/project.pbxproj

Doesn't show up on github, but I get the errors when trying to pull using the command line instructions.

LGM-Doyle:
rebased

o01eg:
Works for me on Ubuntu 16.04 LTS and Gentoo. Looks very good. I got 257 MB RAM for server + 3 AI compared to 339 MB before.

geoffthemedio:
I can't meaningfully review all the code changes, but it seems to work.

dbenage-cx:
Looks great here.

Different point of view using heaptrack for both freeorion (normal) and freeoriond (external) : ~27.5MB leaked on each with master vs ~2MB with PR.
(Fedora 26, GCC7.2.1, no noted differences between boost 1.59 and 1.63)

Just for future reference, the largest observed remaining base portions:
* ~1MB within StringTable::Load on client and server
* ~500KB within OptionsWnd::CompleteConstruction on client
* ~700KB within PythonServer::Initialize() on server

This is from opening each menu window(top right), fleet window, load dialog, and the options window.

LGM-Doyle:
Thanks everyone for taking the time to test this PR.

@o01eg  and @dbenage-cx thanks for the additional data points.  There are still more leaks to be fixed.

geoffthemedio:
http://www.freeorion.org/forum/viewtopic.php?p=90485#p90485

PR #1839, opened 2017-11-04
Add missing check for invalid screen coords in MovementLineData

LGM-Doyle:
This also changes the return type of ScreenPosOnStarane to
boost::optional to depend less on the magic constant INVALID_POSITION


geoffthemedio:
seems fine

PR #1840, opened 2017-11-04
Add a population statistic.

LGM-Doyle:


Vezzra:
@LGM-Doyle, good idea, but why set this to mandatory for 0.4.8? Not that I expect any issues with it, but still, if there were, something like that shouldn't be release blocking...?

LGM-Doyle:
Changed.

I did not mean it to be release blocking.  

I had spliced it off from the leak fixing PR #1838, which I think should be release blocking.  The leaks due to valid FOCS can cause memory exhaustion. For example, adding redundant parentheses.

default/scripting/empire_statistics/Empire.focs.txt
```diff
@@ -24,3 +24,11 @@ Statistic name = "PLANETS_DEPOPULATED" value =
 
 Statistic name = "PLANETS_INVADED" value =
     SpeciesPlanetsInvaded empire = Source.Owner
+
+Statistic name = "TOTAL_POPULATION_STAT"
+    value = Statistic Sum
+        value = Source.Population
```
geoffthemedio: I doubt you really mean / want `Source.Population`. Probably `LocalCandidate.Population`.

See: https://github.com/freeorion/freeorion/blob/master/universe/ValueRef.h#L903

LGM-Doyle:
@geoffthemedio thanks for reviewing.

PR #1842, opened 2017-11-06
Don't save AI messages in the chat history.

o01eg:
Disable saving AI chat message in history buffer so they won't appear in multiplayer lobby.

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -1183,7 +1183,9 @@ sc::result MPLobby::react(const PlayerChat& msg) {
 
     boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
 
-    server.PushChatMessage(timestamp, sender->PlayerName(), data);
+    if (sender->GetClientType() != Networking::CLIENT_TYPE_AI_PLAYER) {
+        server.PushChatMessage(timestamp, sender->PlayerName(), data);
+    }
 
     if (receiver == Networking::INVALID_PLAYER_ID) { // the receiver is everyone
         for (auto it = server.m_networking.established_begin(); it != server.m_networking.established_end(); ++it) {
@@ -1812,7 +1814,9 @@ sc::result PlayingGame::react(const PlayerChat& msg) {
 
     boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
 
-    server.PushChatMessage(timestamp, sender->PlayerName(), data);
+    if (sender->GetClientType() != Networking::CLIENT_TYPE_AI_PLAYER) {
+        server.PushChatMessage(timestamp, sender->PlayerName(), data);
+    }
 
     for (auto it = server.m_networking.established_begin();
          it != server.m_networking.established_end(); ++it)


Dilvish-fo:
We get useful debugging info via AI messages, and need to be able to scroll back for them, so as-is this seems like a no-go to me.   Perhaps if you wanted to make it controllable by an Option?

geoffthemedio:
Is there so much AI chat spam that this is a problem...?

o01eg:
@Dilvish-fo It doesn't disable displaying them. They just doesn't re-appear in the next game session.

@geoffthemedio Looks like they are not usable for multiplayer lobby chat window as only game chat window localizes them.

geoffthemedio:
Please update the description to clarify this is for the lobby chat then... I assumed in-game from the title.

geoffthemedio:
Do AIs ever chat in the lobby?

Dilvish-fo:
@o01eg hmm your answers above have rather confused me.  I initially had the same impression as Geoff, that this was referring to the game chat window.  You have now clarified that this is only intended to affect the multiplayer lobby chat, but then in your reply to Geoff you state (I believe correctly) that the AIs cannot use the multiplayer lobby chat.  So that makes this whole thing seem pointless.  

I just double checked if returning to the MP Lobby from a game causes any MP game chat messages to be repeated in the lobby chat window, and it did not.  So I am not understanding the point of this PR.

> Do AIs ever chat in the lobby?

@geoffthemedio  I don't think that's currently possible.    It looks to me that in a multiplayer game the AI's are not even initiated until the game is started, because assignments of empires to AI or player or mod/obs can be swapped around at any time until then.    I just dug through the code a bit to confirm, and it does still look to me like the server does not try to start the AI's until the WaitingForMPGameJoiners state, which it initiates at the end of sc::result MPLobby::react(const LobbyUpdate& msg) only if the game is ready to start.

o01eg:
AI doesn't send messages in multiplayer lobby but they send message in the game. In hostless mode when the game ends their messages from game chat resend for those who joined to the multiplayer lobby.

> I just double checked if returning to the MP Lobby from a game causes any MP game chat messages to be repeated in the lobby chat window, and it did not.

Chat history message sends when player established so new players will see this messages. Maybe I should also add sending it even after returns to multiplayer lobby.

Dilvish-fo:
OK, thanks for the further explanation, I think I am understanding better now.  I am not sure this is really the best way to handle the issue, though.   Or at least what I suspect is really the primary issue, namely that with a chat history buffer of only 10 messages, if a player has any reason to want to review that chat history after the game is over, then having any of those 10 slots wasted by one of the trivial AI messages means that a meaningful message might get omitted.

It is seeming to me that a better solution is just to increase the size of the chat history buffer.  Although currently the AI does not currently engage in significant diplomacy, it is expected that in the not-too-distant future it will do more diplomacy than it does now, and any player wanting to review diplomatic exchanges after the game should also be able to see the AIs messages then.   It may very well be that the player would often be *more* interested in the various informal chat messages exchanged with other human players, but not necessarily always or entirely.

So, how about we simply increase the size of the chat history buffer from 10 to 100 or so?

o01eg:
I'll prepare PR to add options for history size. But there also one trouble with AI messages. They are send not as plain readable text but as `[<TEXT_IDENTIFIER_000>]` or so then it won't be any sense to see them in history.

Dilvish-fo:
> I'll prepare PR to add options for history size.

I'm not seeing a need for that to be an option-- seems we ought to be able to just make it reasonably large.  Whether we go from 10 to 100, 200 or even a thousand, I'm not seeing a reason for anyone to need an option to make it smaller.  The amount of data involved is fairly trivial, and people can just not scroll up for more messages than they want to see.

> But there also one trouble with AI messages. They are send not as plain readable text but as [<TEXT_IDENTIFIER_000>] or so then it won't be any sense to see them in history.

Please open an Issue for this (or just fix it and submit a PR, hehe).  Sounds like the ingame chat window is handling a layer of stringtable translation that the lobby chat window does not (but should).

o01eg:
What if remove chat controls from `MultiplayerLobbyWnd` and show `ChatWnd` on multiplayer lobby?

geoffthemedio:
If that works for doing to stringtable substitutions, then sure, sounds reasonable.

In future, consider attaching a screenshot demonstrating issues such as this...

geoffthemedio:
I don't really understand the point of either of these pull requests...

As far as I can tell, the chat history is fairly useless as is... it only seems to re-send chat messages from the lobby or in-game when a player joins the lobby. Saved games in multiplayer or single player don't re-send in-game chat messages when a loaded game resumes. It's just the latest chat messages on the server, regardless of whether they were in-game (AI or human presumably) or in-lobby that are sent, and only when someone joins the MP lobby.

It doesn't make any sense to me to send the latest in-game chat messages when someone joins the lobby, so what I'd do is just stop saving in-game chat messages in the history. I don't think any debugging will happen by playing an MP game then quitting and rejoining the lobby just to see the last few AI chat messages that might have been relevant to the AI code.

If chat history was stored in a save file and could be reviewed in the chat in-game, that would be useful, but it's not what's happening now.

Having lobby-only chat be saved on the server while it's running is questionably useful... and probably not useful enough to send up to 1000 lines of it when someone joins. 10 seems reasonable actually... maybe 30ish...?

Dilvish-fo:
> It doesn't make any sense to me to send the latest in-game chat messages when someone joins the lobby, 

Well, what I had \*thought* the point was, was that when the game ended folks would return to the game lobby and want to still see that chat history from the game, and also sometimes people would have a hiccup in their internet connection and rejoin the lobby and want to see the chat history.   But I have no experience with actual multiplayer FO games, perhaps I simply misunderstood.

geoffthemedio:
If someone else says it is or would be useful, that's enough for me... I just wouldn't have thought that.

Vezzra:
Well, it seems that to @o01eg at least it is useful enough to go through the trouble of implementing this. So, I'd say, as long as this doesn't introduce any issue and doesn't get in the way of something else, why not merge it?

If it turns out to be annoying, we can always revert it...

Dilvish-fo:
@geoffthemedio @Vezzra I didn't think that this was the PR that should have been merged-- I noted my objections above to this one.  It was the other one he recently opened pursuant to the discussions here,  #1873,  that should have been merged, to my understanding.  I thought that this one was just remaining open until @o01eg either opened up an Issue or PR regarding stringtable translation in the multiplayer lobby.  I should have been more clear with my understanding on that point, I realize Geoff is quite busy with reviewing many PRs.

So, can we revert this one, and merge #1873 instead?

Vezzra:
Oh, oops, I guess I got confused about what PR we've been tallking about. I've to admit I've obviously lost track which PR is which here, so sure, revert this and merge the other PR, if that's the one that should have been merged...

Dilvish-fo:
OK well, just for clarity of the record, I'll note that Geoff has merged the other one now, without reverting this one.  Whether or not that was intentional or just him being busy, I've decided there is not really a significant need for more discussion; I simply acquiesce to these changes and if there gets to be a time that the multiplayer guys start missing being able to see AI comments in the lobby window then they can just let us know that at such time...

PR #1847, opened 2017-11-08
System totals for Shields/Defense/Troops/Supply on the SidePanel

### LINKED ISSUE #596: Show system totals for Shields, Defense and Supply
Right up at a system's panel, where it's shown the Total Industry and Research for that system, I suggest that it should also be shown the total Shields and Defense values, along with the best planet's Supply value in that system.

Also, those values should be also displayed for systems not owned by the player.

wheals:
As proposed in the issue that had the design discussion, these show only if the same empire controls all planets (or if all planets belong to natives).

Fixes #596

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -2741,6 +2741,80 @@ void SidePanel::PlanetPanelContainer::EnableOrderIssuing(bool enable/* = true*/)
     }
 }
 
+namespace {
+    GG::X LabelWidth()
+    { return GG::X(ClientUI::Pts() * 9); }
+
+    GG::X ValueWidth()
+    { return GG::X(ClientUI::Pts() * 4); }
+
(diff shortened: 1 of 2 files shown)


UI/SidePanel.cpp
```diff
@@ -3171,12 +3274,18 @@ void SidePanel::RefreshImpl() {
     if (m_system_resource_summary->Empty()) {
         DetachChild(m_system_resource_summary);
     } else {
-        // add tooltips to the system resource summary
+        // add tooltips to the system resource/meter summary
         for (const auto& entry : meter_types) {
             MeterType type = entry.first;
+            ResourceType resource = MeterToResource(type);
```
dbenage-cx: [asserts](https://github.com/freeorion/freeorion/blob/master/universe/Enums.cpp#L48) on debug builds when `all_planets_share_owner`

geoffthemedio:
The tooltips need title bars.

wheals:
Added them.

PR #1849, opened 2017-11-08
Generate a sitrep when a ship naturally goes from 0.x to 1.x fuel.

### LINKED ISSUE #835: Refueling a ship with fuel < 1 to >= 1 should generate a sitrep
I have to poll all this horde of scouting ships stuck in the galaxy every turn otherwise.

wheals:
This is handy whenever you're in the early to mid game and don't want to just set your ships on autoexplore.

Fixes #835


MatGB:
Haven't tested but looks fine apart from minor quibble, please add the label to [freeorion/common_user_customizations.txt at master · freeorion/freeorion](https://github.com/freeorion/freeorion/blob/master/default/customizations/common_user_customizations.txt), probably best above SITREP_PLANET_DRIVE_FAILURE_LABEL (ln 96).

wheals:
I moved it a line farther up than that, to avoid splitting up SITREP_PLANET_DRIVE_FAILURE_LABEL and SITREP_PLANET_DRIVE_LABEL (not that you're ever likely to get them on the same turn unless you're really busy, but it looks better this way).

geoffthemedio:
seems to work...

PR #1850, opened 2017-11-09
OSX CMake branch: fix Python patchlevel.h path

Vezzra:
*Attention:* This PR is for the `cmake-osx` branch!

Apparently, on OSX, the cmake script variable `PYTHON_INCLUDE_PATH` gets set to `.../Python.framework/Versions/Current/include/python2.7`, which causes a compile error in [`DependencyVersions.cpp`, line 11](https://github.com/freeorion/freeorion/blob/621359c4cb9afa4a63c3e2ae4103c52e8861fdca/util/DependencyVersions.cpp#L11).

The fix proposed in this PR removes the "python2.7" part from the `include` statement.

However, to me having all the Python API includes directly in the header file search paths without having to specify the "python2.7" path in include statements looks like asking for trouble. So a better solution should be to adjust the cmake scripts so that `PYTHON_INCLUDE_PATH` gets set to `.../Python.framework/Versions/Current/include`.

However, as I don't know how to do that (at least in a non-hacky way), I decided to provide this PR as a stop-gap solution until someone can provide a better one.


adrianbroher:
> However, to me having all the Python API includes directly in the header file search paths without having to specify the "python2.7" path in include statements looks like asking for trouble.

It's the right way to fix this.  Our current implementation is faulty and warned by Python upstream:

https://docs.python.org/2/c-api/intro.html#include-files

> The header files are typically installed with Python. On Unix, these are located in the directories `prefix/include/pythonversion/` and `exec_prefix/include/pythonversion/`, where **prefix** and **exec_prefix** are defined by the corresponding parameters to Python’s **configure** script and version is `sys.version[:3]`. On Windows, the headers are installed in` prefix/include`, where **prefix** is the installation directory specified to the installer.
>
> To include the headers, place both directories (if different) on your compiler’s search path for includes. Do **not** place the parent directories on the search path and then use `#include <pythonX.Y/Python.h>`; this will break on multi-platform builds since the platform independent headers under **prefix** include the platform specific headers from **exec_prefix**.

Vezzra:
Rebased.

Vezzra:
> It's the right way to fix this. Our current implementation is faulty and warned by Python upstream

Ok, I see... well, in that case, we can merge this as it is, right?

However, what I'm wondering: isn't that awfully prone to overlaps/conflicts between header files with the same name? I mean, the Python header files include things like "code.h", "datetime.h", "eval.h", "import.h", "node.h", "object.h", "token.h" etc., header files with the same name can easily occur with other frameworks, libraries, or the source of an application. With just e.g. `#include <object.h>`, how can you make sure the right `object.h` gets included if there are more than just the Python one?

adrianbroher:
> Ok, I see... well, in that case, we can merge this as it is, right?

I would replace it with a plain

```
#include <patchlevel.h>
```
And test if the existing builds still work.

> With just e.g. #include <object.h>, how can you make sure the right object.h gets included if there are more than just the Python one?

You don't use the API that way.  When using Python you should only `#include <Python.h>`. The `Python.h` includes the details headers by `#include ""`, which searches relative to the current file (here: Python.h), whereas `#include <>` searches relative to a list of search pathes passed to the compiler and after that falling back to a built-in list of search pathes, taking the first matching file.

In case a hypothetic library uses `object.h` as API include name as public part of their API (which Python doesn't) the developer should reconsider their API design.  When a library uses `#include <>` to include other headers of their API they should fix their implementation.  When you use `include <>` to include headers within your poject you should fix your implementation.

The emphasis is here on `should`.  You have to work with tools at hand but I would assume that the respective upstream has some reasoning behind the usage instruction the published.

adrianbroher:
Cherry-picked into #1455

Vezzra:
> And test if the existing builds still work.

Yep, still works on all platforms. Well, looks like you beat me to it, just wanted to confirm that your solution worked for me.

PR #1851, opened 2017-11-10
OSX CMake branch: fix copying of frameworks to OSX app bundle

Vezzra:

On OSX, cmake’s copy_directory command follows symlinks, which does not work well when copying the frameworks to the app bundle (frameworks on OSX contain a lot of symlinks, so copying them while following symlinks results in the copy being up to twice the size of the original).

Fixed this by using the ‘cp’ command instead of ‘cmake -E copy_directory’.


adrianbroher:
Cherry-picked into #1455

PR #1852, opened 2017-11-10
OSX CMake branch: force ‘Release’ build config for Xcode generator

Vezzra:

When using the Xcode generator, including ‘Debug’ into CMAKE_CONFIGURATION_TYPES results in the Debug build configuration always used when building the generated Xcode project via CMD-B, regardless what CMAKE_BUILD_TYPE is set to. The only way to prevent this and force Xcode to actually use the Release build configuration is to set CMAKE_CONFIGURATION_TYPES to ‘Release’ only.

As the resulting app of a debug build on OSX is more than twice the size as that of a non-debug build, also significantly slower, and you won’t need debug builds with Xcode most of the time, set CMAKE_CONFIGURATION_TYPES to ‘Release’ only when the Xcode has been specified.


adrianbroher:
I doubt it is useful for branch here for Xcode.  The default `CMAKE_CONFIGURATION_TYPES` is `[ Debug, Release, MinSizeRel, RelWithDebInfo ]`. I only removed the last two configurations as they are not used as much but kept the order intact.  If you think that the build should be `Release` by default for all multi-config generators we could do that by changing the order.

Vezzra:
> If you think that the build should be Release by default for all multi-config generators we could do that by changing the order.

Unfortunately that doesn't work. We already ran into that issue, see our discussion in #1455, starting with the last part of [this comment](https://github.com/freeorion/freeorion/pull/1455#issuecomment-299706377). I guess that approach worked with earlier versions of Xcode that just used build configuration like most other similar IDEs do, but the introduction of these "schemes" probably broke that.

Short of removing "Debug" from `CMAKE_CONFIGURATION_TYPES`, nothing makes a cmake generated Xcode project use "Release" as default build configuration. At least I haven't found out any other way to fix this. So unless you've a better solution, I'd really prefer to include this patch, as having to manually go through all the schemes and correct the build configuration settings whenever the Xcode project gets regenerated is a *really* annoying chore.

adrianbroher:
I swapped the order of entries and regenerated the project it worked as I would expect.

Maybe this is a caching + CMake version issue.

You can inspect the current cache value with `grep CMAKE_CONFIGURATION_TYPES CMakeCache.txt`.

Can you look up the current build configuration with `xcodebuild -list` when being inside the project.

Edit: Have you tried shutting down Xcode in between the rebuilds?

Vezzra:
> Have you tried shutting down Xcode in between the rebuilds?

I had Xcode closed, completely wiped and regenerated the build tree, opened the freshly generated Xcode project and hit CMD-B.

Anyway, I've digged/googled a bit more into this, and apparently the issue is with Xcode here. CMake doesn't support generating these build schemes at all, at least not up until version 3.8. Only starting with version 3.9 there is experimental support for generating the `xcscheme` files.

Which means when cmake generates the Xcode project, it only creates the `<projectname>.xcodeproj` folder and the main `project.pbxproj` file within. And as far as that is concerned, you're right, the order of the build configurations in `CMAKE_CONFIGURATION_TYPES` does matter. The first build configuration in that list is set as the default build configuration for all targets, and also as the default build configuration to be used when building from the command line.

However, as soon as you open the Xcode project file with Xcode, `xcscheme` files are auto-generated for each target that doesn't already have one, and there Xcode apparently doesn't give a crap about default build configurations anymore. If there is a 'Debug' build configuration, it gets assigned to the 'Run', 'Test' and 'Analyze' build actions. And apparently the build configuration assigned to the 'Run' build action is what is used when you trigger a build via CMD-B.

I've found no solution for this. Looks like there is simply no way to generate those pesky `xcscheme` files from the command line. Xcode simply doesn't provide any means to automate the process of generating these files (short of opening a project in Xcode that is). All the solutions I came across involve maintaining the `xcscheme` files in the repo and copying them to the generated Xcode project in a post build step. Ugh, ugh, ugh. I intensly dislike that aproach, and I suspect you're not exactly enthusiastic about that idea either. :wink:

I don't know what exactly you checked after generating the Xcode project, I guess you looked into the `project.pbxproj` file? Otherwise, if you actually run your test on a Mac and building via CMD-B gave you a 'Release' build out of the box, then it has to be a Xcode version issue. I still use Xcode 6.4, maybe later versions consider the default build configuration when autogenerating the `xcscheme` files.

Anyway, if you can live with it, I'd really prefer if we could merge this patch, as it would make my life easier. If you really think that's a bad idea, I guess I can patch together some local workaround to alleviate the problem on my system.

adrianbroher:
My bad, I was using the manual project after all when testing.  The Xcode UI is confusing.  But yes, even with Xcode 9.1 and CMake 3.9.4 the IDE doesn't pick up `Release` as default configuration.

adrianbroher:
Cherry-picked into #1455

Vezzra:
>  The Xcode UI is confusing.

Tell me about it...

Issue #540, opened 2016-03-19
Damage Control techs need further balancing


cpeosphoros:
Refer to: http://www.freeorion.org/forum/viewtopic.php?f=15&t=10006&view=unread&sid=96ccf93875079e58510e2de75d3978b7#unread

MatGB said:

> this includes both the techs and the auto-repair hulls, ideally no repairs should happen immediately after combat except for Logistics, Nanobot and BioAdaptive, and none of them should be complete. Plus, the techs should repair based on current structure not max structure, forcing heavily damaged big ships to limp home for repairs not simply hide for a couple turns

Vezzra:
@MatGB, do you want to do any work on this front for 0.4.6?

MatGB:
Maybe. It's a low priority compared to other stuff, but if I've time _and_ we get the Drydock stuff agreed/completed then it'd be a good companion to it.

Alternatively, if someone else wants to give it a go holler and I'll be more clear about what I've in mind, it's spread around multiple forum threads at the moment.

Vezzra:
Hm, ok, so I'll leave that as undecided/optional.

Vezzra:
@MatGB: do you consider this issue sufficiently addressed by all the changes/adjustments that went in 0.4.6 to close this issue, or should we keep it open because there is still enough that needs to be done on that front?

MatGB:
This bit

>  the techs should repair based on current structure not max structure, forcing heavily damaged big ships to limp home for repairs not simply hide for a couple turns

still isn't done, need to change fleet field repair and advanced damage control to be %age current structure, easy to code, not so easy to balance or get the AI updated for.

It was a low priority before Release thing, but it can get done now.

Vezzra:
Ok, so it needs to stay open.

Vezzra:
@MatGB: any updates for this one, or does what you said last September still apply?

Vezzra:
@MatGB, status update on this?

dbenage-cx:
I've a [potential branch](https://github.com/freeorion/freeorion/compare/master...dbenage-cx:fix-540_damage_control_tech-current_structure) covering only the techs (as that was the last mention).
If no one else is currently working on it, I can take assignment.

If the scope should still include hulls, will open discussion thread to work out details.

Vezzra:
> If no one else is currently working on it, I can take assignment.

Fine with me - @MatGB? You working on this, or can @dbenage-cx take it up?

MatGB:
[...]

PR #1853, opened 2017-11-11
Balance adjustments for damage control techs

dbenage-cx:
Addresses some considerations in #540 for ship repair techs.

Damage Control: requires non-combat
*  Basic Damage Control - Requires Asymptotic Materials
    - 1 structure per turn
*  Advanced Damage Control - Requires Basic Damage Control and Reinforced Hull
    - 10% of current structure
    - minimum 3

Field Repair: requires non-combat, stationary and in supply, repairs based on current structure (or remaining damage), not cumulative with drydock.
*  Fleet Field Repair - Requires Interstellar Logistics
    - 15% of (lesser of) current structure or max - current
    - minimum 1
*  Advanced Field Repair - Requires Fleet Field Repair and Gravitic Architecture
    - 25% of (lesser of) current structure or max - current
    - minimum 5


Vezzra (changes requested):
The 50%/75% max structure caps for the Damage Control techs don't make much sense to me. After all, these are supposed to be a ships capabilities to repair itself, without outer help. You'd expect minor damages to be repaired most easily... why would a ship be able to repair almost critical damage, but not be able to repair some minor scratches? Not going for realism here, it's just very counter-intuitive.

If anything, I'd do it just the other way round: no repair if damage *exceeds* 50%/75%. That makes much more sense to me...

> Advanced Field Repair [...] allows repair outside of supply

Have some strong reservations about that. Having to be in supply is what sets the Fleet Repair techs apart from the Damage Control techs. It's their unique characteristic.

Dropping that requirement makes Advanced Fleet Repair more like another level of Advanced Damage Control, and IMO much too powerful, even more so considering that Advanced Fleet Repair and Advanced Damage Control stack.

Vezzra:
@freeorion/ai-team, can you give an assessment about how much work it will be to adjust the AI to these changes?

MatGB:
I have zero problem with the idea of advanced field repair being used out of supply, I completely reworked that tech and absorbed it into the main damage control tree ages ago and it was fairly useless before.

But Vezzra's put his finger on my problem with the %age thing, from a pure balance perspective it sort of makes sense but from any sort of actually playing/realistic expectation sense it's confusing. I can justify it: you can't fully repair a ship unless you go to a dock but you can bolt on extra plates and cover holes, but I'm not sure I get the why. If we're fairly strongly limiting repair of badly damaged ships then it's encouraging a return to drydock anyway.

Dilvish-fo:
> why would a ship be able to repair almost critical damage, but not be able to repair some minor scratches

Hmm, I have mixed thoughts on that-- you could think of it as dropping in a new engine or welding on new hull plates, but the welds won't be as strong as the original, and any non-replaced structures will still have microfractures accumulating, etc.   Have you ever crashed a car and then had it repaired back to 100%?  I think not :laughing:   In at least some ways I like the idea of having a cap on how much a ship can be repaired, but I could also imagine it might just wind up seeming more like a complicated nuisance.

> freeorion/ai-team, can you give an assessment about how much work it will be to adjust the AI to these changes?

Hmm, can I just say "No"? :smiley:   

Short answer is that it probably would not be hard to make sure this doesn't break the AI, beyond that it really depends on how much we try to accomplish and it's hard to say without seeing where this winds up and then working on the AI a little.

Keep in mind that the AI's repair planning is still on the rudimentary side as it is, with plenty of TODO comments in the code about ways it still should be improved.  If ships are badly damaged it will put them in passive mode and send them back to a drydocks for repair, and that's about the sum of it.  There are many things it doesn't yet take into account, including the advanced repair techs that might make such a trip somewhat or entirely unneeded.  So there is already plenty of room for more highly nuanced decisions, even without this PR.

I think the most important thing to keep in mind is that caps on how much can be repaired would absolutely require AI adjustments, or AI fleets would get stuck sitting at drydocks.   Also pretty importantly, putting in a floor below which repairs are not possible should also be accompanied by an AI adjustment. 

Both of those things, and adding a new tech into the AI research plans, would be fairly easy to handle in at least a reasonable way.   Chances are we'd probably get around to implementing at least some of our current 'TODO' notes also.

Vezzra:
> you could think of it as dropping in a new engine or welding on new hull plates, but the welds won't be as strong as the original, and any non-replaced structures will still have microfractures accumulating, etc.

I'm well aware that finding a proper fluff explanation isn't the problem here, although, if you want to go down that specific road to justify it, I could probably think of a couple of things why that still doesn't make sense (I mean, come on, we can, however slowly and painstakingly, do repairs on a ship that is at 1% structure, meaning, at the point of falling apart, and, given enough time, get back to 50 or even 75% structure, but if just a single armor plate gets blown off in the last combat, we can't replace it???).

My main issue is it's counter-intuitive. As a player I expect a basic damage control tech to be able to deal with minor damage best/fastest, and becoming increasingly ineffective/useless with more serious damage.

Or, to phrase it the other way round, having to resort to repair methods that are more difficult to provide (be it by having to return to a drydock, or researching more advanced repair techs) when having to deal with more serious damage.

That's simple, straightforward and intuitive. Probably not realistic, I certainly give you that - in reality, for a really complete repair, especially after suffering severe damage, you'd have to go to a drydock.

But from a gameplay perspective, I don't see the fun in this. As a player I'd find that mostly annoying, having to send my ships around to get minor damage repaired. Ugh... 👎 😉 

> In at least some ways I like the idea of having a cap on how much a ship can be repaired

The only way how that could be interesting in my eyes would be if making that cap dependent on how much damage a ship has suffered since the last visit to a drydock. E.g. you could say Damage Control techs can only repair up to double the amount of the lowest structure the ship has had since it has last been at a drydock.

Of course, that would make the whole repair mechanics more complicated. You'd have to keep track of that "structural low point", which would only get reset to max structure at a drydock.

I have to admit I actually like that idea, but I really wonder if that wouldn't make repair too detailed and complicated. Not really KISS anymore.

> but I could also imagine it might just wind up seeming more like a complicated nuisance.

Yeah, pretty much this.

geoffthemedio:
To me, whether a ship has a limit to how much it can be repaired in various situations seems like it would be a hull-line-dependent feature.

Vezzra:
> Hmm, can I just say "No"? 😃

Of course. However, if you had also told me that those changes had the potential to break the AI, I'd have to flag this PR for post release as a consequence... 😉 

> Short answer is that it probably would not be hard to make sure this doesn't break the AI

That's pretty much what I need to know. As long as we can make sure with a reasonable amount of effort that the AI wouldn't get worse at repairs with these changes going in, we're good. Otherwise we'd have to consider postponing merging this until after the release.

Vezzra:
> To me, whether a ship has a limit to how much it can be repaired in various situations seems like it would be a hull-line-dependent feature.

That's certainly an interesting idea, I like it. Would add another thing by which we can distinguish the hull lines. Maybe even have hull line specific repair techs (I mean, it would make sense that there are certain repair techniques that work particularly well on asteroid hulls, but not at all on organic hulls, and vice versa).

dbenage-cx:
I pushed some changes I had been testing along with consistent naming. I am open to any proposed changes (still parsing the discussion), but the improved descriptions may help clarify the division.

My intention towards the normal damage control techs were for those temporary repairs done to keep a ship still functional, with the objective of those repairs to return for full repair. e.g. the ship crew does not include dedicated maintenance staff. Such definition might couple better if there were side-effects for ships with low structure (reduced movement / erratic behavior).
I will remove the cap on these techs, they were intended to satisfy 'do not fully repair'.

The self repair techs were for more proper repairs when a drydock is not available, and requiring the ship be stationary.  I added supply requirement to advanced field repair. Out of supply repairs seem like a good candidate for a part effect.

Vezzra:
We've once again managed to start a design discussion in a PR comment section. I've created a corresponding thread on the forum and cited the entire above design discussion:

http://freeorion.org/forum/viewtopic.php?f=6&t=10775

Please continue all design related discussion there.

Vezzra (approved):

Vezzra:
@dbenage-cx, closed as what? Superseded (do you want to follow up with another PR) or invalid (do you just want to discard the proposed changes)?

dbenage-cx:
Hmm, was cleaning up old branches and missed this had pending PR.
Does not look like this proposal has much traction, so I will leave as closed/invalid.

PR #1854, opened 2017-11-12
OSX CMake: Fix DMG package generation on OSX

Vezzra:

In order for the FO app bundle to actually get copied into the DMG file CPack generates, a proper install statement needs to be issued for the OSX platform.


adrianbroher:
Cherry-picked into #1455

PR #1855, opened 2017-11-14
 Add a section to the fleet browse window with ship designs. 

wheals:


geoffthemedio:
minor concern about what happens if there are a lot of designs... is there a cap on how many rows it will show?

wheals:
Didn't implement a cap, but note that it only shows for designs that actually have a ship existing, and since ships from designs that you've stopped using in a game are probably going to be destroyed in combat anyway since they're weaker, it sort of tends to balance out.
Might be more of a problem on a smaller monitor than mine (1080p), or if you use more varied designs than I tend to, but my test save is ~350 turns in and looks fine.

PR #1856, opened 2017-11-14
Move the x axis label to be above the bars in the combat report summary.

wheals:

It looks like this now:
image

This is just a personal preference on my part, so if other people think it looks better as is I'm ok with keeping the status quo.


Vezzra:
I very rarely look at the combat summary, so I'm probably not the best person to judge this, but the proposed changes look fine to me.

MatGB:
I use it a fair bit especially when testing, and it did confuse me sometimes for multi-side fights, I definitely approve of this.

PR #1857, opened 2017-11-14
Make the escape key close windows on the MapWnd in LIFO order.

wheals:

I tested this on a variety of combinations of windows and it always seemed to work, but there might be edge cases I missed still.

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -5837,9 +5837,8 @@ void MapWnd::Sanitize() {
 }
 
 void MapWnd::PushWndStack(std::shared_ptr<GG::Wnd> wnd) {
-    if (!wnd) {
+    if (!wnd)
         return;
-    }
     // First remove it from its current location in the stack (if any), to prevent it from being 
     // present in two locations at once.
     RemoveFromWndStack(wnd);
@@ -5855,8 +5854,11 @@ void MapWnd::RemoveFromWndStack(std::shared_ptr<GG::Wnd> wnd) {
 
 bool MapWnd::ReturnToMap() {
     std::shared_ptr<GG::Wnd> wnd;
-    // Find the top non-null Wnd in the stack (check if it's visible, in case it was somehow closed
-    // without being removed from the stack).
+    // Pop the top Wnd from the stack, and repeat until we find a non-null and visible one (or
+    // until the stack runs out).
+    // Need to check that it's visible, in case it was closed without being removed from the stack;
+    // if we didn't reject such a Wnd, we might close no window, or even open a window.
+    // Either way, the Wnd is removed from the stack, since it is no longer of any use.
     while (!m_wnd_stack.empty() && !(wnd && wnd->Visible())) {
         wnd = m_wnd_stack.back();
         m_wnd_stack.pop_back();
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -1460,6 +1460,7 @@ void MapWnd::CompleteConstruction() {
     m_sitrep_panel->ClosingSignal.connect(
         boost::bind(&MapWnd::HideSitRep, this));
     if (m_sitrep_panel->Visible()) {
+        PushWndStack(m_sitrep_panel);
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Default open windows seemingly don't get added to the list, so won't close with escape until closed and reopened. Eg. messages, empires.

wheals:
Good catch, I missed that because my config options don't seem to open any windows by default.

UI/MapWnd.cpp
```diff
@@ -5824,70 +5836,60 @@ void MapWnd::Sanitize() {
...
+        m_wnd_stack.erase(it);
     }
+}
 
-    if (m_production_wnd->Visible()) {
-        ToggleProduction();
-        some_subscreen_was_visible = true;
+bool MapWnd::ReturnToMap() {
+    std::shared_ptr<GG::Wnd> wnd;
+    while (!m_wnd_stack.empty() && !(wnd && wnd->Visible())) {
```
geoffthemedio: add comment explaining what this is doing and/or why

UI/MapWnd.cpp
```diff
@@ -5824,70 +5836,60 @@ void MapWnd::Sanitize() {
...
-    // close sidepanel if open
-    if (SidePanel::SystemID() != INVALID_OBJECT_ID) {
-       SelectSystem(INVALID_OBJECT_ID);
-       return true;
+    } else if (cui && wnd == cui->GetPlayerListWnd()) {
+        HideEmpires();
+    } else if (cui && wnd == cui->GetMessageWnd()) {
+        HideMessages();
+    } else {
+        ErrorLogger() << "Unknown GG::Wnd " << wnd->Name() << " found in MapWnd::m_wnd_stack";
```
geoffthemedio, line 5900: are you sure `wnd` is not null here?

UI/MapWnd.cpp
```diff
@@ -5824,70 +5836,60 @@ void MapWnd::Sanitize() {
...
-        ToggleSitRep();
-        some_subscreen_was_visible = true;
-    }
-
-    if (m_research_wnd->Visible()) {
-        ToggleResearch();
-        some_subscreen_was_visible = true;
-    }
+void MapWnd::PushWndStack(std::shared_ptr<GG::Wnd> wnd) {
+    RemoveFromWndStack(wnd);
```
geoffthemedio: add safety check for null `wnd`

UI/MapWnd.cpp
```diff
@@ -5824,70 +5836,60 @@ void MapWnd::Sanitize() {
...
-
-    // close fleets window if open
-    FleetUIManager& fm = FleetUIManager::GetFleetUIManager();
-    GG::Wnd* active_fleet_wnd = fm.ActiveFleetWnd();
-    if (active_fleet_wnd) {
+    } else if (wnd == m_combat_report_wnd) {
+        m_combat_report_wnd->Hide();
+    } else if (wnd == m_side_panel) {
+        SelectSystem(INVALID_OBJECT_ID);
+    } else if (wnd.get() == fm.ActiveFleetWnd()) {
```
geoffthemedio, line 5893: also here

UI/MapWnd.cpp
wheals: Yes, the check on line 5865 will return from the function if `!wnd`

UI/MapWnd.cpp
wheals: Same as the other comment

UI/MapWnd.cpp
```diff
@@ -5837,6 +5837,11 @@ void MapWnd::Sanitize() {
 }
 
 void MapWnd::PushWndStack(std::shared_ptr<GG::Wnd> wnd) {
+    if (!wnd) {
+        return;
```
geoffthemedio: remove the `{}` or at least put on the same line as `return`

UI/MapWnd.cpp
```diff
@@ -5850,10 +5855,13 @@ void MapWnd::RemoveFromWndStack(std::shared_ptr<GG::Wnd> wnd) {
 
 bool MapWnd::ReturnToMap() {
     std::shared_ptr<GG::Wnd> wnd;
+    // Find the top non-null Wnd in the stack (check if it's visible, in case it was somehow closed
+    // without being removed from the stack).
```
geoffthemedio: are any such null Wnds removed at some point?

PR #1861, opened 2017-11-17
Set stringtable filename from system locale

### LINKED ISSUE #1811: Honor locale language env if no --stringtable-filename is given in parameters or set in the UI - linux
### Environment
linux FO 0.4.7.1

### Description
when my locale is fr 
(ex : LANG=fr_FR.UTF-8
LANGUAGE=fr_FR:fr
LC_CTYPE="fr_FR.UTF-8"
LC_NUMERIC="fr_FR.UTF-8"
LC_TIME="fr_FR.UTF-8"
LC_COLLATE="fr_FR.UTF-8"
LC_MONETARY="fr_FR.UTF-8"
LC_MESSAGES="fr_FR.UTF-8"
LC_PAPER="fr_FR.UTF-8"
LC_NAME="fr_FR.UTF-8"
LC_ADDRESS="fr_FR.UTF-8"
LC_TELEPHONE="fr_FR.UTF-8"
LC_MEASUREMENT="fr_FR.UTF-8"
LC_IDENTIFICATION="fr_FR.UTF-8"
LC_ALL=fr_FR.UTF-8
)

it would be nice for the UI to use the fr string table automatically.

keep the --strintable-filename options and the option in the UI, thoses are great
for override current locale.

thanks.

I understand this is a low-prio feature and it's ok. just consider it.

### Expected Result
if the locale is french, then start FO with the fr stringtable

### Steps to reproduce

export LC_ALL=fr_FR.UTF-8;export LANG=fr_FR.UTF-8;export LANGUAGE=fr_FR:fr
then : freeorion

dbenage-cx:
Attempts to determine preferred language from system locale.
Check is only performed when the `stringtable-filename` option is at default value.

Resolves #1811

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,6 +321,7 @@ target_link_libraries(freeorioncommon
     ${Boost_DATE_TIME_LIBRARY}
     ${Boost_FILESYSTEM_LIBRARY}
     ${Boost_IOSTREAMS_LIBRARY}
+    ${Boost_LOCALE_LIBRARY}
     ${Boost_LOG_LIBRARY}
     ${Boost_LOG_SETUP_LIBRARY}
     ${Boost_REGEX_LIBRARY}
--- a/util/Directories.cpp
+++ b/util/Directories.cpp
@@ -635,3 +635,14 @@ fs::path GetPath(const std::string& path_string) {
     }
     return GetPath(path_type);
 }
+
+bool IsExistingFile(const fs::path& path) {
+    try {
+        auto stat = fs::status(path);
+        return fs::exists(stat) && fs::is_regular_file(stat);
+    } catch(boost::filesystem::filesystem_error& ec) {
+        ErrorLogger() << "Filesystem error during stat of " << PathToString(path) << " : " << ec.what();
+    }
+
+    return false;
+}
--- a/util/Directories.h
+++ b/util/Directories.h
@@ -108,5 +108,8 @@ FO_COMMON_API boost::filesystem::path GetPath(PathType path_type);
 /** Returns path for path type cast from @p path_string */
 FO_COMMON_API boost::filesystem::path GetPath(const std::string& path_string);
 
+/** Returns if path exists and is a regular file */
+FO_COMMON_API bool IsExistingFile(const boost::filesystem::path& path);
+
 
 #endif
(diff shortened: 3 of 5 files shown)


geoffthemedio:
Seems to work for me. If someone with a system language that doesn't match the default stringtable could report, that would be useful...

Vezzra:
Doesn't seem to work on OSX. When I launch FO, the English stringtable is used, although the system language on my Mac is German, and to make sure I also deleted config.xml...

dbenage-cx:
TY for testing, expanded to always log prior to return.

If testing different language settings, it may be easier to set an empty value in persistent_config.xml (`<stringtable-filename></stringtable-filename>`), rather than delete config.xml between runs.

@Vezzra Possibly the locale differs when run as an Application.
(So far that is the only relevant find from search results, from an issue with VSCode.)
If you find time to test again, does the log differ when run from terminal?
Also be curious to know what `LANG` is set to (e.g. by executing `locale`), I assume that might be `de_DE.UTF-8`

dbenage-cx:
Once I can loan time on a mac, I will look to changing Info.plist to execute a script setting `LANG` from `defaults read -g AppleLanguages` prior to launch.

Vezzra:
@dbenage-cx, yep, starting the app normally vs. starting from the command line definitely gives different results. When starting from the command line, detecting and setting the stringtable according to the system locale works.

When starting the app normally, the system locale does not get detected correctly. This is the corresponding log output:
```
11:48:40.794874 [debug] client : i18n.cpp:53 : Detected language: c
11:48:40.794951 [warn] client : i18n.cpp:61 : Stringtable file /Users/user/SoftwareProjekte/FO/src/Xcode/build/Test/FreeOrion.app/Contents/Resources/default/stringtables/c.txt not found, falling back to default
```
When starting from the command line, the language is correctly detected as `de`.

> Also be curious to know what LANG is set to (e.g. by executing locale), I assume that might be de_DE.UTF-8

`locale` gives `LANG="de_AT.UTF-8"` (well, my system locale is set to Austria of course... :wink:).

dbenage-cx:
Tyvm for checking, will change to always compare by lower case.

Would prefer to set environment prior to launch, but still struggling to have the resulting app bundle to execute a script.
If it does not pan out in next few days I can add osx specific code to InitStringtableFileName or possibly some objc target.

geoffthemedio:
This seemingly doesn't break anything on OSX, so is there a downside the merging as is?

dbenage-cx:
No downside afaict, support for osx app can follow in later PR.
Changed to conversion of detected language to lower case, rebased.

PR #1863, opened 2017-11-17
Close fleet windows properly when clicking on the map.

### LINKED ISSUE #1859: Fleet Windows not showing
### Environment
```
$ inxi -GSx
System:    Host: chronos-18 Kernel: 4.4.0-98-generic x86_64 (64 bit gcc: 5.4.0)
           Desktop: Cinnamon 3.2.7 (Gtk 3.18.9) Distro: Linux Mint 18.1 Serena
Graphics:  Card: NVIDIA GM200 [GeForce GTX 980 Ti] bus-ID: 01:00.0
           Display Server: X.Org 1.18.4 drivers: nvidia (unloaded: fbdev,vesa,nouveau)
           Resolution: 1920x1080@60.00hz, 1680x1050@59.88hz
           GLX Renderer: GeForce GTX 980 Ti/PCIe/SSE2 GLX Version: 4.5.0 NVIDIA 384.98 Direct Rendering: Yes
```
* **FreeOrion Version**: 5f3d5b0  
* **Fetched as**
  * Compiled from source
### Description
Clicking on a fleet (or otherwise selecting it, e.g., via <CTRL>-<G>) only opens its Fleet Window once. After that, selecting a fleet doesn't open its fleet window, effectively rendering the game unplayable. 
### Expected Result
Selecting a fleet should open its fleet window.
### Steps to reproduce
* Compile 5f3d5b0  
* Play game
* Select a fleet. Fleet window shows.
* Unselect it.
* Select it again. Fleet window doesn't show.
### Notes
* Happens both with player's fleets and with monsters.
* Closing the game and reopening resets the behavior (i.e. each fleet can be selected once and have its window shown), which is effectively a workaround, albeit a very clumsy one.
* Moving/resizing a Fleet Window and closing it doesn't save its position/size (another fleet's window will open at the default position and size).
* I've tried reverting recent commits which affect FleetWnd.cpp or FleetWnd.h and recompiling, to no avail.

wheals:
The problem was that the shared_ptr in m_wnd_stack was keeping the window alive for too long. Possibly it should be changed to hold weak_ptrs instead?

Fixes #1859

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -2634,12 +2634,22 @@ void MapWnd::LButtonUp(const GG::Pt &pt, GG::Flags<GG::ModKey> mod_keys) {
 void MapWnd::LClick(const GG::Pt &pt, GG::Flags<GG::ModKey> mod_keys) {
     m_drag_offset = GG::Pt(-GG::X1, -GG::Y1);
     FleetUIManager& manager = FleetUIManager::GetFleetUIManager();
+    const auto fleet_wnd = manager.ActiveFleetWnd();
     bool quick_close_wnds = GetOptionsDB().Get<bool>("UI.window-quickclose");
 
     // if a fleet window is visible, hide it and deselect fleet; if not, hide sidepanel
-    if (!m_dragged && !m_in_production_view_mode && manager.ActiveFleetWnd() && quick_close_wnds) {
-        manager.CloseAll(); }
-    else if (!m_dragged && !m_in_production_view_mode) {
+    if (!m_dragged && !m_in_production_view_mode && fleet_wnd && quick_close_wnds) {
+        // Remove the reference to the active fleet window from the stack, to avoid it keeping the window
+        // alive for longer than it is supposed to.
+        auto it = std::find_if(m_wnd_stack.begin(), m_wnd_stack.end(),
+            [=](std::shared_ptr<GG::Wnd> wnd) {
+                return wnd.get() == fleet_wnd;
+            });
+        if (it != m_wnd_stack.end()) {
+            m_wnd_stack.erase(it);
+        }
+        manager.CloseAll();
+    } else if (!m_dragged && !m_in_production_view_mode) {
         SelectSystem(INVALID_OBJECT_ID);
         m_side_panel->Hide();
     }
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -2639,15 +2639,6 @@ void MapWnd::LClick(const GG::Pt &pt, GG::Flags<GG::ModKey> mod_keys) {
 
     // if a fleet window is visible, hide it and deselect fleet; if not, hide sidepanel
     if (!m_dragged && !m_in_production_view_mode && fleet_wnd && quick_close_wnds) {
-        // Remove the reference to the active fleet window from the stack, to avoid it keeping the window
-        // alive for longer than it is supposed to.
-        auto it = std::find_if(m_wnd_stack.begin(), m_wnd_stack.end(),
-            [=](std::shared_ptr<GG::Wnd> wnd) {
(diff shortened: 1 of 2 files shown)


Vezzra (dismissed):
Works on OSX.

Vezzra:
At a first glance it looked like the issue had been fixed, selecting a fleet, deselecting, selecting again worked, the fleet window popped up, closed, popped up again.

However, when I started a new test game and began switching between fleets, new problems turned up: when selecting another fleet with the fleet window open, either the fleet windows position and size got reset, or the fleet window didn't switch to the selected fleet(s).

Then, after closing the fleet window, it wouldn't open again, until I pressed ESC (sometimes had do that multiple times). So, there is still something seriously amiss...

dbenage-cx:
As it has not been mentioned prior, probably warrants testing with `UI.multiple-fleet-windows`

wheals:
Working on a further fix, hoping I can catch the time today to get it done.

wheals:
Changing it to use `weak_ptr` was simpler than I expected, and seems like the better choice as far as the code goes anyway. I tested Vezzra's issues and couldn't reproduce them after adding this commit, and `UI.multiple-fleet-windows` seems to work as well, though I've never used it before so it may have changed.

Vezzra (approved):
Just played a test game for 90 turns, seems to be fixed. Had no issues with the fleet window anymore, everything works as intended.

UI/MapWnd.cpp
```diff
@@ -5922,7 +5923,11 @@ void MapWnd::PushWndStack(std::shared_ptr<GG::Wnd> wnd) {
 }
 
 void MapWnd::RemoveFromWndStack(std::shared_ptr<GG::Wnd> wnd) {
-    auto it = std::find(m_wnd_stack.begin(), m_wnd_stack.end(), wnd);
+    auto it = std::find_if(m_wnd_stack.begin(), m_wnd_stack.end(),
+        [&wnd](std::weak_ptr<GG::Wnd> weak_wnd) {
+            return weak_wnd.lock() == wnd;
```
geoffthemedio: what happens if an empty pointer is passed in as `wnd` ?

wheals: The first `weak_ptr` that compares equal to an empty `shared_ptr` will be removed; i.e., any null ones or ones whose `shared_ptr` has already been destructed. If they are null (or have had their `shared_ptr` destructed), they don't contribute anything to the stack anymore, so no harm done.

LGM-Doyle: @wheals you could try `GG::ProcessThenRemoveExpiredPtrs()`, which will process a container of weak pointers with a supplied function, while removing any null ones.  There are examples of its use in "FleetWnd.cpp", including [here](https://github.com/freeorion/freeorion/blob/master/UI/FleetWnd.cpp#L438)

wheals: I think the problem with that function for this use case is that it doesn't give fine-grained enough control over the vector. In specific, both times I iterate over `m_wnd_stack` (`RemoveFromWndStack` and `ReturnToMap`), I remove items from the vector, and doing so during the iteration would invalidate all the iterators.

LGM-Doyle: I see.  I did not look at your code closely enough.

You could process the vector afterwards to remove the null `weak_ptrs`.  Otherwise, they just leak.

wheals: Good idea; they wouldn't have leaked generally because they are removed in ReturnToMap if one presses escape enough, but if someone happens to not use the escape shortcut it could be a problem.

Vezzra:
Did another brief test with the updated PR, still seems to work fine on OSX.

Vezzra:
Would be good if this could be merged before coming Monday, so the next test builds already contain the fix. Judging by the complaints we got on the forum, that issue is quite confusing for players.

PR #1867, opened 2017-11-18
Change SitRepPanel to only validate sitreps that might be shown.

LGM-Doyle:
Previously, each turn it validated all sitreps since the start of time.

This saves an increasing amount of time during turn init every turn of the game.

### DIFF
--- a/UI/SitRepPanel.cpp
+++ b/UI/SitRepPanel.cpp
@@ -480,10 +480,10 @@ void SitRepPanel::SizeMove(const GG::Pt& ul, const GG::Pt& lr) {
 }
 
 namespace {
-    /* Sort empire's sitreps for each turn */
-    std::map<int, std::list<SitRepEntry>> GetSitRepsSortedByTurn(int empire_id, std::set<std::string> hidden_sitrep_templates) {
-        std::map<int, std::list<SitRepEntry>> turns;
-        bool verbose_sitrep = GetOptionsDB().Get<bool>("verbose-sitrep");
+    /* Sort sitreps for each turn.
+     Note: Validating requires substituting all of the variables which is time
+     consuming for sitreps that the player will never view. */
+    std::map<int, std::list<SitRepEntry>> GetUnvalidatedSitRepsSortedByTurn(int empire_id) {
         std::set<Empire*> sr_empires;
         Empire* empire = GetEmpire(empire_id);
         if (empire) {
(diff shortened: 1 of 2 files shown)


UI/SitRepPanel.cpp
```diff
@@ -494,68 +494,107 @@ namespace {
...
+/** Return true if the sitrep is not hidden, validates and is not snoozed. */
+bool SitRepPanel::IsSitRepInvalid(SitRepEntry& sitrep) const {
+    if (m_hidden_sitrep_templates.count(
+            sitrep.GetLabelString().empty() ? sitrep.GetTemplateString() : sitrep.GetLabelString()))
+    { return true; }
+
+    // Validation is time consuming because all variables are substituted
+    bool validated = sitrep.Validate();
+
+    // verbose-sitrep will hide all sitreps that do not validate
```
geoffthemedio: seems misleading... I'd say having verbose-sitrep off / disabled will hide sitreps that do not validate

UI/SitRepPanel.cpp
```diff
@@ -494,68 +494,107 @@ namespace {
...
 
-int SitRepPanel::GetNextNonEmptySitrepsTurn(const std::map<int, std::list<SitRepEntry>>& turns, int turn, bool forward) const {
+/** Return true if the sitrep is not hidden, validates and is not snoozed. */
+bool SitRepPanel::IsSitRepInvalid(SitRepEntry& sitrep) const {
+    if (m_hidden_sitrep_templates.count(
+            sitrep.GetLabelString().empty() ? sitrep.GetTemplateString() : sitrep.GetLabelString()))
+    { return true; }
+
+    // Validation is time consuming because all variables are substituted
+    bool validated = sitrep.Validate();
```
geoffthemedio, line 515: is validation faster than the other checks done below? otherwise, why not move it to the end of this function?

LGM-Doyle: The snoozed sitrep filters work with the interpolated text, which validation generates.  

It would be best to change the snooze code to work with the data and then only validate text that will be shown, but I did not take the time to improve that.

This order is the best that can be achieved without improving the snooze code.

UI/SitRepPanel.cpp
LGM-Doyle: fixed.

LGM-Doyle (commented):
Both comments are addressed.

PR #1868, opened 2017-11-18
Make GeneralizedLocation report no location for fields and temp objs

LGM-Doyle:
`GeneralizedLocation` always reported no location (system id) with an error message for types of objects that it does not know about.  

This PR make fields a known type that has no location, because it is not centered on a system.  
This fixes error messages that started up later in the game when the Experimentors start adding/removing starlanes.

It also make temporary objects that are not located in a system, or fleets return no location and not produce an error message.  The error message was being triggered by temporary objects created by the pedia.

### DIFF
--- a/universe/Pathfinder.cpp
+++ b/universe/Pathfinder.cpp
@@ -839,7 +839,7 @@ namespace {
         if (fleet)
             return std::make_pair(fleet->PreviousSystemID(), fleet->NextSystemID());
 
-        if (auto field = std::dynamic_pointer_cast<const Field>(obj))
+        if (std::dynamic_pointer_cast<const Field>(obj))
             return nullptr;
 
         // Don't generate an error message for temporary objects.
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Some fields are located "in" or "at" a system.

LGM-Doyle:
If a `field->SystemID()` returns a location, then that will be the location used for the field.

universe/Pathfinder.cpp
```diff
@@ -838,6 +839,13 @@ namespace {
         if (fleet)
             return std::make_pair(fleet->PreviousSystemID(), fleet->NextSystemID());
 
+        if (auto field = std::dynamic_pointer_cast<const Field>(obj))
```
geoffthemedio: the `auto field = ` is probably unnecessary and likely to produce an unused variable warning.

LGM-Doyle: fixed.

PR #1869, opened 2017-11-18
Move value ref parsers into their own compiled file.

LGM-Doyle:
This change reduces compile time by  having the value ref parsers only compile once.  It reduces size of libfreeorionparse.so for the DEBUG build with `gcc` by 100MB.  It did not make a difference for `clang`


geoffthemedio:
Trying to build this, I get
`1>..\..\parse\EnumValueRefRules.cpp(163): fatal error C1060: compiler is out of heap space`

LGM-Doyle:
1. rebased
2. I refactored the locations where `enum_value_ref_rules` are instantiated, which hopefully will reduce the memory requirements and fix the crash.
3. Switching MSVC to use the 64-bit toolset might have fixed this issue.  I was unable to do this myself.  Here is additional information in case someone else wants to make this change.

    I think the underlying problem is that MSVC defaults to using the 32-bit toolset to compile even 64-bit executables on 64-bit machines.  The 32-bit toolset has a lower memory limit than may be available on modern machines. Our parsers bump into these limits and trigger C1060.

    Last week, I tried changing the XML in the MSVC project by hand to enable the 64bit toolset to make the AppVeyor continuous integration (CI) work without changing the code.  I was  unable to fix the crash, but that may be because I failed to turn on the 64-bit toolset.

    Switching the MSVC to use the 64-bit toolset, may provide other benefits.  It is recommended with whole program optimization, which we have enabled.

    It may be easier setup the 64-bit toolset from within the MSVC environment, so I've included some links in case anyone wants to pursue this further.

    [Fatal Error C1060](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1060)
    [How to: Enable a 64-Bit, x64 hosted Visual C++ toolset](https://docs.microsoft.com/en-us/cpp/build/how-to-enable-a-64-bit-visual-cpp-toolset-on-the-command-line)

LGM-Doyle:
@geoffthemedio does this PR build correctly for you now?

geoffthemedio:
yes

geoffthemedio:
I can set a project, eg. GiGi, to build in x64 mode in the IDE. It compiles, but when it tries to link, I get a bunch of unresolved external symbol errors from Boost and libpng and various others, likely because they're not also all built in 64 bit mode.

geoffthemedio:
I can set a project, eg. GiGi, to build in x64 mode in the IDE. It compiles, but when it tries to link, I get a bunch of unresolved external symbol errors from Boost and libpng and various others, likely because they're not also all built in 64 bit mode.

Dilvish-fo:
It appears to me that the solution may be to make it use the 64-bit toolset, but still target a 32-bit x86 platform in order to be compatible with the 32-bit libraries.   (At least, it is sounding like it went into full x64 bit mode rather than just using the 64-bit toolchain).  Various articles talk about doing that specifically to overcome memory limitations of the 32-bit toolchain.

Here is [a description of how to do specifically use the 64-bit toolchain via environment variable](https://stackoverflow.com/questions/19820718/how-to-make-visual-studio-use-the-native-amd64-toolchain#comment75409933_23793055), and [here is a description of doing it on a per-project basis](https://stackoverflow.com/a/25626630).  If the project doesn't already explicitly specify targeting x86 cpus, then I guess you'd need to specify that with the Configuration Manager. 

I hope that helps.

Dilvish-fo:
@LGM-Doyle could you rebase and clear up the conflict from the planet size changes?

Dilvish-fo:
For reference, the by-project approach for using the x64 toolset seems to work fine for me
```diff
diff --git a/msvc2015/Parsers/Parsers.vcxproj b/msvc2015/Parsers/Parsers.vcxproj
index e9d7e84..78d060c 100644
--- a/msvc2015/Parsers/Parsers.vcxproj
+++ b/msvc2015/Parsers/Parsers.vcxproj
@@ -20,6 +20,9 @@
     <RootNamespace>Parsers</RootNamespace>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup>
+    <PreferredToolArchitecture>x64</PreferredToolArchitecture>
+  </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
```

And anyways, it sounds like after the refactoring this is compiling & linking fine for Geoff?  It seems to run fine for me, and seems like a good idea.  Time to merge?

LGM-Doyle:
@geoffthemedio has requested in #1911 that this PR wait until after https://github.com/freeorion/freeorion/pull/1768 is merged.

geoffthemedio:
It was that and meter modifications into effects. Probably I should just say go ahead though, as it's unreasonable to make your changes wait more multiple sets of mine.

LGM-Doyle:
Closed. Changes were merged as part of https://github.com/freeorion/freeorion/pull/1911

Vezzra:
Set the "status:merged" label to indicate the changes here have been merged although the PR itself has been closed.

PR #1871, opened 2017-11-18
Remove second exit path for HumanClientApp.

LGM-Doyle:
The HumanClientApp used HummanClientApp::CleanQuit to exit sometimes and
SDLGUI::QuitSignal at other times.

In this commit HumanClientApp now overrides SDLGUI::Exit and has a
single exit mechanic.

### DIFF
--- a/client/human/HumanClientApp.h
+++ b/client/human/HumanClientApp.h
@@ -72,7 +72,7 @@ public:
     void                ResetToIntro(bool skip_savegame);
 
     /** Exit the App with \p exit_code. */
-    void                Exit(int exit_code) override;
+    void                ExitApp(int exit_code = 0) override;
     /** Exit SDL. (Called by FSM) */
     void                ExitSDL(int exit_code);
 
--- a/client/human/HumanClientApp.h
+++ b/client/human/HumanClientApp.h
@@ -22,8 +22,6 @@ class HumanClientApp :
     public GG::SDLGUI
 {
 public:
-    class CleanQuit : public std::exception {};
-
     typedef boost::signals2::signal<void (bool)> FullscreenSwitchSignalType;
     typedef boost::signals2::signal<void ()>     RepositionWindowsSignalType;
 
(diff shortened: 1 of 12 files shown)


client/human/HumanClientApp.h
```diff
@@ -72,7 +70,12 @@ class HumanClientApp :
     void                ChangeLoggerThreshold(const std::string& option_name, LogLevel option_value);
 
     void                ResetToIntro(bool skip_savegame);
-    void                ExitApp();
+
+    /** Exit the App with \p exit_code. */
+    void                Exit(int exit_code) override;
+    /** Exit SDL. (Called by FSM) */
```
geoffthemedio: some explanations here about why these are separate functions and/or why you would use one or the other in particular situations would be helpful...

LGM-Doyle: I have refactored the code so that `GG::GUI::Exit()` is called `GG::GUI::ExitApp()`, `ExitSDL()` is private and the FSM is passed a `std::function<void()>` as an action to take after the server has exited.  Hopefully, being private makes it clear that users of `HumanClientApp` should not call `ExitSDL()` 

A better solution would have been to directly bind `SDLGUI::ExitApp()` in `ResetOrExitApp()`, so that there is no need to expose the private one liner `ExitSDL()`.  However, `SDLGUI` is a pure virtual class and all forms of intentional slicing result in uncompileable code.

client/human/HumanClientFSM.h
```diff
@@ -372,10 +375,9 @@ struct QuittingGame : boost::statechart::state<QuittingGame, HumanClientFSM> {
 
     std::chrono::steady_clock::time_point m_start_time;
 
-    // if m_reset_to_intro is true QuittingGame transits to the Intro menu, otherwise it
-    // exits the app.
-    bool m_reset_to_intro;
-    Process* m_server_process;
+    Process* m_server_process = nullptr;
+    /// An action to be completed after disconnecting from the server
+    std::function<void()> m_after_server_shutdown_action = std::function<void()>();
```
geoffthemedio, line 380: for clarity, or is explicitly calling the default construction doing something special here...?

LGM-Doyle: It was for clarity.  `m_server_process` needs a default value. `m_start_time` can't have a default value, it needs to be initialized to `Clock::now()`.  It was to be clear that `m_after_server_shutdown_action` was properly considered.

Issue #1865, opened 2017-11-18
Splitting ships for each design causing the entire fleet to disappear


cpeosphoros:
### Environment
```
$ inxi -GSx
System:    Host: chronos-18 Kernel: 4.4.0-98-generic x86_64 (64 bit gcc: 5.4.0)
           Desktop: Cinnamon 3.2.7 (Gtk 3.18.9) Distro: Linux Mint 18.1 Serena
Graphics:  Card: NVIDIA GM200 [GeForce GTX 980 Ti] bus-ID: 01:00.0
           Display Server: X.Org 1.18.4 drivers: nvidia (unloaded: fbdev,vesa,nouveau)
           Resolution: 1920x1080@60.00hz, 1680x1050@59.88hz
           GLX Renderer: GeForce GTX 980 Ti/PCIe/SSE2 GLX Version: 4.5.0 NVIDIA 384.98 Direct Rendering: Yes
```
* **FreeOrion Version**:  80e7841 
* **Fetched as**
  * Compiled from source

### Description
At turn 253, splitting ships for design causes the fleet to disappear.

It affects any fleet, no matter if it has or has not different design ships.

This is the save game (without XML compression)
[disappearing ships.sav.zip](https://github.com/freeorion/freeorion/files/1484672/disappearing.ships.sav.zip)
.

### Expected Result
Ships should not disappear...

### Steps to reproduce
* Load the game.
* Try to split any fleet.
* Issue occurs.

wheals:
Might be worth checking if this still happens after applying the fix in #1863

dbenage-cx:
@wheals I tried with your PR and with b6940d0 with no change.

The save file states a build from 80e7841, curious if the game is a continuation from an earlier build?

Loading the save generates a number of error logs of form:
[error] client : ObjectMap.cpp:108 : ObjectMap::FindObjects couldn't find object with id 169352713

When splitting the fleet, none of the passed ships ids are found, thus no fleets are created.

Moreover, clicking next turn for me ends up segfaulting the server, stemming from IDAllocator.cpp:312 (during deallocation? posting screenshot as the info may be easier to parse)
![fo_debug_181117_01](https://user-images.githubusercontent.com/17205519/32986903-e0745454-cca1-11e7-8964-b3aeabfd4154.png)

cpeosphoros:
@dbenage-cx You are right, this is a build from 80e7841. Not a continuation. I always start fresh games after compiling.

[...]

PR #1874, opened 2017-11-21
Remove irrelevant warning in UniverseObject::MoveTo()  of invalid position.

LGM-Doyle:
This changes the warning about moving to an invalid position to a clamp to the maximum size of the universe on the x and/or y dimension that exceeds the bounds of the universe.

### DIFF
--- a/universe/UniverseObject.cpp
+++ b/universe/UniverseObject.cpp
@@ -13,8 +13,6 @@
 
 #include <stdexcept>
 #include <boost/lexical_cast.hpp>
-// TODO replace with std::clamp when converting to C++17
-#include <boost/algorithm/clamp.hpp>
 
 // static(s)
 const int INVALID_OBJECT_ID      = -1;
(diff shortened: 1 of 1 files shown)


geoffthemedio:
why?

LGM-Doyle:
I was looking into #1865 and some of the fields in the saved game are outside of the boundary of the universe.  Those fields are in an invalid state and the only feedback was a warning in the log file when they were moved.  The universe objects should always be in a valid state.  So the move should do one of the following after generating the error message:
- not move the object from its current valid location
- move the object to (INVALID_POSITION,INVALID_POSITION)
- throw
- go to the closest valid location.  

I chose the latter because I was thinking that the field had just drifted out of the universe.

geoffthemedio:
I'd guess it's more likely they were spawned there and moved a short distance that left them still outside the nominal universe size.

I'd just remove the warning... there's no real reason to clamp objects within a square of length nominal universe width to a side...

LGM-Doyle:
I've removed the warning.

geoffthemedio:
looks reasonable, merge if you'd like, perhaps after test builds complete

Vezzra:
@LGM-Doyle, title and description of the PR should probably be changed to reflect the actual change that went in (so someone browsing the closed PRs doesn't get the wrong idea of what this PR did).

LGM-Doyle:
Done.

Vezzra:
Thx!

PR #1864, opened 2017-11-18
Remove Phototrophic min pop requirement on blue and white stars

dbenage-cx:
Removes the minimum pop restriction for phototrophic pop bonus on white or blue stars.

Workaround for issue #1619, in event a suitable fix is not implemented before 0.4.8rc.

PR #1875, opened 2017-11-21
Reapply effects when determining planet suitability

### LINKED ISSUE #1619: Incorrect Planet Suitability Report
Freeorion Version: 2017-06-25.ec064ec

As reported [in the forum](http://freeorion.org/forum/viewtopic.php?f=28&t=10630#p89326), the planet suitability report sometimes report an inaccurate value (see screenshot). On first glance, it might be that the phototrophic bonus is incorrectly included in the suitability report (which would be +9 for medium blue star).

planetsuitabilitybug

Steps to reproduce:
1. Download the [savegame in this forum post](http://freeorion.org/forum/viewtopic.php?f=28&t=10630#p89308).
2. Load as multiplayer game and choose the Union empire.
3. Check suitability report for the planet Maretta II
4. Optionally wait till the Laenfa Colony is finished.

dbenage-cx:
[Test Save](https://github.com/freeorion/freeorion/files/1492764/PhotoBlueWhite.zip)
Correct results for key planets:

| Planet | Target Pop |
| ----- | ----- |
| Kraz ɑ I | +10 |
| Kraz ɑ II | -5.5 |
| Kraz γ III | +2.0 |
| Rum II | -0.5 |
| York β I | +12 |

Resolves #1619

### DIFF
--- a/UI/EncyclopediaDetailPanel.cpp
+++ b/UI/EncyclopediaDetailPanel.cpp
@@ -2525,6 +2525,7 @@ namespace {
 
         auto font = ClientUI::GetFont();
         GG::X max_species_name_column1_width(0);
+        std::vector<int> planet_id_vec { planet_id };
 
         GetUniverse().InhibitUniverseObjectSignals(true);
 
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
Nice catch.  I expect this has also been responsible for some other discrepancies that would sometimes appear with the population estimates.

dbenage-cx:
I need to clean up the comments, especially to note not to override meter values (easy to achieve opposite behavior from the bug, Rum II reports correctly but not York β I).

This PR is minimal for the issue, I've not looked closely at `UpdateMeterEstimates` to determine if there is a potential issue with other uses.

dbenage-cx:
And now can not reproduce when reverting to `UpdateMeterEstimates`. likely just missed testing for this case.

Vezzra:
Did a quick test with this PR and the test save you provided. Kraz ɑ I, Kraz γ III and York β I showed the correct results according to your table, but for the other two (which are those with the Temporal Anomaly special) I get the following values:

| Planet | Target Pop |
| ----- | ----- |
| Kraz ɑ II | +0.5 |
| Rum II | +2.5 |

dbenage-cx:
@Vezzra ty for testing, restored use of `ApplyMeterEffectsAndUpdateMeters`

Vezzra:
@dbenage-cx, oops, I think I confused some branches when I did my quick testing. Just noticed that instead of testing with the branch of this PR I used the one of PR #1864 (apparently I only looked at the "fix-1619" part of the branch name). So forget what I said above, sorry... 😳 

I did however another test with the latest version of the PR, and now all values in the suitability report are the same as in the table you provided.

dbenage-cx:
I've removed the relevant comment portion, but retained use of `ApplyMeterEffectsAndUpdateMeters` as it may be slightly cheaper in some instances.

PR #1876, opened 2017-11-24
Filesystem operation common utility functions - IsExistingFile PathsInDir

dbenage-cx:


util/Directories.cpp
```diff
@@ -635,3 +635,55 @@ fs::path GetPath(const std::string& path_string) {
     }
     return GetPath(path_type);
 }
+
+bool IsExistingFile(const fs::path& path) {
```
geoffthemedio, line 639: what does this do differently from just calling `fs::is_regular_file(path)` ?

geoffthemedio:
Without being used in specific places, it's unclear why these are needed...

util/Directories.cpp
dbenage-cx: The main use of such functions is work towards consolidating filesystem operations.
The are declaration compatible excerpts from larger work towards VFS.

Without considering the above (as it has not been discussed in depth), for this one function:
* Eliminates need to include boost/filesystem/operations.hpp at call site
* Short-circuits if `fs:exists` fails (with single stat call)
* Ensures filesystem errors are handled

dbenage-cx:
`IsExistingPath` is picked from (EDIT, wrong pr) #1861
Both are used in [refactor of suitability report](https://github.com/freeorion/freeorion/compare/master...dbenage-cx:refactor_enc_suitability_report), which I hoped to rebase prior to posting as PR.

dbenage-cx:
@geoffthemedio Prefer I rebase this branch (leaving just `PathsInDir`), or close this PR as invalid and keep the commit within other PR?

geoffthemedio:
Within the other(s) seems preferable.

PR #1881, opened 2017-11-26
Fix make Pending explicitly non copyable

LGM-Doyle:
This PR is an extension to #1416 to check if making Pending explicitly non-copyable will force MSVC to chose the move constructor in the cmake build.


adrianbroher:
Cherry picked your commit into #1416 because I wanted to rebase it.

PR #1883, opened 2017-11-27
AI basic stockpile use

### LINKED ISSUE #1879: Adjust AI for new Imperial Stockpile mechanic
The AI needs some (at least) basic adjustment to be able to handle the new Imperial Stockpile mechanic before the 0.4.8 release. I've created this issue and assigned it to the "0.4.8 (mandatory)" milestone so this shows up on the list of things which need to be done before we can get 0.4.8 out.

Dilvish-fo:
Enables the AI to make basic use of the imperial stockpile.

Resolves #1879

### DIFF
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -1402,8 +1402,34 @@ def generate_production_orders():
                 if (not prioritized) and (this_priority == PriorityType.PRODUCTION_INVASION):
                     fo.issueRequeueProductionOrder(production_queue.size - 1, 0)  # move to front
         print
+    update_stockpile_use()
     fo.updateProductionQueue()
     _print_production_queue(after_turn=True)
+
+
+def update_stockpile_use():
+    """Decide which elements in the production_queue will be enabled for drawing from the imperial stockpile.  This
+    initial version simply ensures that every resource group with at least one item on the queue has its highest
+    priority item be stockpile-enabled.
+
+    :return: None
+    """
+    # TODO: Do a priority and risk evaluation to decide on enabling stockpile draws
+    empire = fo.getEmpire()
+    production_queue = empire.productionQueue
+    resource_groups = set(tuple(el.key()) for el in empire.planetsWithAvailablePP)
+    planets_in_stockpile_enabled_group = set()
+    for queue_index, element in enumerate(production_queue):
+        if element.locationID in planets_in_stockpile_enabled_group:
+            # TODO: evaluate possibly disabling stockpile for current element if was previously enabled, perhaps
+            # only allowing multiple stockpile enabled items in empire-capital-resource-group, or considering some
+            # priority analysis
+            continue
+        group = next((_g for _g in resource_groups if element.locationID in _g), None)
+        if group is None:
+            continue # we don't appear to own the location any more
+        if fo.issueAllowStockpileProductionOrder(queue_index, True):
+            planets_in_stockpile_enabled_group.update(group)
 
 
 def build_ship_facilities(bld_name, best_pilot_facilities, top_locs=None):
--- a/default/python/AI/TechsListsAI.py
+++ b/default/python/AI/TechsListsAI.py
@@ -223,6 +223,7 @@ class TechGroup1SparseC(TechGroup1):
             "LRN_FORCE_FIELD",
             "SHP_WEAPON_2_2",
             "PRO_ORBITAL_GEN",
+            "PRO_IMPERIAL_STOCKPILE_2",
             "SPY_DETECT_2",
             "SHP_SPACE_FLUX_DRIVE",
         )
@@ -237,6 +238,7 @@ class TechGroup2(TechGroup):
             "PRO_FUSION_GEN",
             "PRO_SENTIENT_AUTOMATION",
             "PRO_EXOBOTS",
+            "PRO_IMPERIAL_STOCKPILE_2",
             # "PRO_MICROGRAV_MAN",  # handled by fast-forwarding when we have asteroids
             # "PRO_ORBITAL_GEN",    # handled by fast-forwarding when we have a GG
 
@@ -282,7 +284,8 @@ class TechGroup2A(TechGroup2):
             self.weapon,
             self.weapon,
             self.weapon,
-            self.economy
+            self.economy,
+            self.economy,
         )
 
 
@@ -304,7 +307,8 @@ class TechGroup2B(TechGroup2):
             self.weapon,
             self.weapon,
             self.economy,
-            self.economy
+            self.economy,
+            self.economy,
         )
 
 
@@ -325,6 +329,7 @@ class TechGroup2SparseA(TechGroup2):
             self.weapon,
             self.defense,
             self.weapon,
+            self.economy,
             self.weapon,
             self.weapon
         )
@@ -347,6 +352,7 @@ class TechGroup2SparseB(TechGroup2):
             self.defense,
             self.armor,
             self.weapon,
+            self.economy,
             self.weapon,
             self.weapon
         )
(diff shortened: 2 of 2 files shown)


default/python/AI/ProductionAI.py
```diff
@@ -1402,10 +1402,37 @@ def generate_production_orders():
...
+    priority item be stockpile-enabled.
+
+    :return: None
+    """
+    # TODO: Do a priority and risk evaluation to decide on enabling stockpile draws
+    empire = fo.getEmpire()
+    production_queue = empire.productionQueue
+    resource_groups = set(tuple(el.key()) for el in empire.planetsWithAvailablePP)
+    planets_with_group_stockpile_elem = set()
+    for queue_index in range(len(production_queue)):
```
Morlic-fo: ```enumerate(production_queue)``` seems clearer.

default/python/AI/ProductionAI.py
```diff
@@ -1402,10 +1402,37 @@ def generate_production_orders():
...
+        element = production_queue[queue_index]
+        if element.locationID in planets_with_group_stockpile_elem:
+            continue
+        group = None
+        for group in resource_groups:
+            if element.locationID in group:
+                break
+        else:
+            continue # we don't appear to own the location any more
```
Morlic-fo: Use built-in ```next```:
```
group = next(x for x in resource_groups if element.locationID in x, None)
if group is None:
    continue
```

default/python/AI/ProductionAI.py
```diff
@@ -1402,10 +1402,37 @@ def generate_production_orders():
...
+    """
+    # TODO: Do a priority and risk evaluation to decide on enabling stockpile draws
+    empire = fo.getEmpire()
+    production_queue = empire.productionQueue
+    resource_groups = set(tuple(el.key()) for el in empire.planetsWithAvailablePP)
+    planets_with_group_stockpile_elem = set()
+    for queue_index in range(len(production_queue)):
+        element = production_queue[queue_index]
+        if element.locationID in planets_with_group_stockpile_elem:
+            continue
```
Morlic-fo: Prefer a comment for clarity here: ```# For now, use stockpile only for the highest priority item in a resource group```.

default/python/AI/ProductionAI.py
```diff
@@ -1402,10 +1402,37 @@ def generate_production_orders():
...
+    initial version simply ensures that every resource group with at least one item on the queue has its highest
+    priority item be stockpile-enabled.
+
+    :return: None
+    """
+    # TODO: Do a priority and risk evaluation to decide on enabling stockpile draws
+    empire = fo.getEmpire()
+    production_queue = empire.productionQueue
+    resource_groups = set(tuple(el.key()) for el in empire.planetsWithAvailablePP)
+    planets_with_group_stockpile_elem = set()
```
Morlic-fo: The variable name confused me when reading it for the first time.

Maybe ```planets_in_stockpile_enabled_group```?

default/python/AI/ProductionAI.py
```diff
@@ -1402,10 +1402,37 @@ def generate_production_orders():
...
     _print_production_queue(after_turn=True)
 
 
+def update_stockpile_use():
+    """Decide which elements in the production_queue will be enabled for drawing from the imperial stockpile.  This
+    initial version simply ensures that every resource group with at least one item on the queue has its highest
+    priority item be stockpile-enabled.
+
+    :return: None
+    """
```
Morlic-fo, line 1416: The current implementation allows to have a low-priority item in the queue that is stockpile-enabled:

Consider there is a short disruption in the supply lines. The disconnected system enqueues some filler item to get rid of excess PP in that system but that item is not really needed. Because it is disconnected, it will draw from stockpile.
After supply connection is restored, the item is still enabled for stockpile-usage. 

So, we may want to add a loop prior to this implementation that stockpile-disables all the queue items.

default/python/AI/ProductionAI.py
```diff
@@ -1402,10 +1402,37 @@ def generate_production_orders():
...
+    # TODO: Do a priority and risk evaluation to decide on enabling stockpile draws
+    empire = fo.getEmpire()
+    production_queue = empire.productionQueue
+    resource_groups = set(tuple(el.key()) for el in empire.planetsWithAvailablePP)
```
Morlic-fo, line 1420: Does ```empire.planetsWithAvailablePP``` consider supply groups without any PP?

AI/AIInterface.h
```diff
@@ -345,6 +345,9 @@ namespace AIInterface {
     int IssueChangeProductionQuantityOrder(int queue_index, int new_quantity, int new_blocksize);
     int IssueRequeueProductionOrder(int old_queue_index, int new_queue_index);
     int IssueDequeueProductionOrder(int queue_index);
+    int IssueChangeProductionQuantityOrder(int queue_index, int new_quantity, int new_blocksize);
```
Morlic-fo: ?

Dilvish-fo: that's a whoops! copy-paste error. will be fixed

default/python/AI/ProductionAI.py
Dilvish-fo: If we were going to do that I think that a simpler way to get the same treatment would be to handle it within the same loop, checking and turning off the stockpile as part of the ```if element.locationID in planets_with_group_stockpile_elem:``` check

But, I had considered that route and chose not to take it, I'd be more open to simply not enabling the stockpile for items outside the group that includes the imperial capital.  Especially as we expand this code, but even now, I think it's fine and good that on turn N the top priority item would get marked as stockpile enabled, and then if on turn N+1 another item is made top priority and the former top slips to # 2 priority, it is still stockpile enabled.

I think that the real problem in your scenario is that an unwanted filler item is being built at all-- under both your proposal and the current PR the "filler" item will still be getting stockpile enabled while the side system is disconnected, and also under both approaches it will continue getting built and drawing PP once the system reconnects to the main group, the only difference is the _potential_ for some extra stockpile drain, and I don't see avoiding that being worth the cost of restricting the main group to having just its top item be stockpile enabled.

Let's look a little more closely at the "filler" item.  My recollection is that such can only be ships.  Outpost ships still be valuable no matter where they are built (at least currently), and to a fair extent I would expect the same to be true for colony ships (which the AI would be very unlikely to build anyways).   In an isolated system either of those two ship types might not be the most advanced available to the empire, but should still be fine to build.  The real potential problem right now is with military ships and to a lesser extent troop ships.  

Right now there is a check for military ships, so that if we start building one with relatively poor pilots it is logged, but it still gets built.  I think that's the bit of could that should be augmented for this problem-- no need to directly look at pilot, but for military ships and troop ships if the rating/cost efficiency number there is too low then we just don't build such things there. That threshold could be at 50% or 70%, somewhere in there, barring exceptional circumstances.

default/python/AI/ProductionAI.py
Dilvish-fo: My recollection is that I had checked that when writing this code, and the answer was yes.

default/python/AI/ProductionAI.py
Dilvish-fo: ok sure

default/python/AI/ProductionAI.py
Dilvish-fo: ok fine

default/python/AI/ProductionAI.py
Dilvish-fo: Seems to me that a more accurate (at least for the current code) version of that comment is already only a few (ok, 22) lines higher up, in the routine's docstring, where it seems more important.  A repeat isn't seeming warranted to me.

default/python/AI/ProductionAI.py
Dilvish-fo: with some extra parentheses that will work, ok

Dilvish-fo:
I amended the interface commits somewhat and pushed them to master so that after this next weekly build anyone (particularly Morlic and CJ) would be able to test and further augment the related AI python code.  

I also adjusted the AI python code as discussed above, and rebased.  @Morlic-fo if you are content with this as a base for further development, feel free to go ahead and merge it; it might easily be another week before I am able to get back here.

PR #1884, opened 2017-11-27
Use json-based encoding for the AI savegame state

Morlic-fo:
The previous implementation of the AI savegames used the ```pickle``` module which trivially allows arbitrary code execution and must be considered a serious security threat.

The proposed new implementation is based on json encoding and the python standard library json decoder (https://docs.python.org/2/library/json.html). The standard json protocol is extended to allow a greater variety of (trusted) classes and non-string dictionary keys.

The standard python json decoder is to my best knowledge considered safe and will only (de)construct the following objects:
```dict, list, unicode, int, float, True, False, None```

The extended en-/decoder allows sets, tuples, Enums defined in EnumsAI.py and a set of trusted classes. It will use the local source code to determine which classes and modules are trusted and use the local source code to reconstruct those classes.

For obvious reasons, this breaks compatibility with older savegames as far as the AI is concerned (i.e. it will spam errors when loading an older savegame but play from a fresh state as usual).

Link to some discussion preceding this PR:
https://github.com/Morlic-fo/freeorion/pull/15

### CODE REVIEW

File: default/python/AI/AIstate.py [hunk 1]
```diff
@@ -62,18 +62,10 @@ def convert_to_version(state, version):
...
-            raise ConversionError("Error when converting to compatibility version 1: "
-                                  "Can't find aggression in character module. Exception thrown was: " + e.message)
-    elif version == 2:
-        # some dicts which used only the keys were changed to sets
-        for var_name in ['visInteriorSystemIDs', 'exploredSystemIDs', 'visBorderSystemIDs', 'unexploredSystemIDs']:
-            state[var_name] = set(state.get(var_name, {}).keys())
+    # Starting with version 3, we switched from pickle to json-style encoding
+    # Do not try to load an older savegame even if it magically passed the encoder.
+    if version <= 3:
+        raise ConversionError("The AI savegame version is no longer supported.")
```
Cjkjvfnby, line 69: It is not clear for me at this point if we will just cleanup AI state (Is message like "AI state will lost memory and continue game" show to user) or stop game (in that case it will be useful to add game commit/ version that should be rollback to finish game.)

default/python/AI/AIstate.py
```diff
@@ -173,6 +164,16 @@ def __setstate__(self, state):
                     error("Could not find the aggression level of the AI, defaulting to typical.", exc_info=True)
                     aggression = fo.aggression.typical
             self.__init__(aggression)
+            return
+
+        for content in ("colonisablePlanetIDs", "colonisableOutpostIDs"):
+            # could be that there is only a different species definition file
+            # but could also be some attempt to get in some nasty string.
+            sorted_planets = state[content].items()
+            sorted_planets.sort(lambda x, y: cmp(x[1], y[1]), reverse=True)
```
Cjkjvfnby: Use `key=itemgetter(1)`   https://docs.python.org/2/library/operator.html#operator.itemgetter

default/python/AI/AIstate.py
```diff
@@ -173,6 +164,16 @@ def __setstate__(self, state):
                     error("Could not find the aggression level of the AI, defaulting to typical.", exc_info=True)
                     aggression = fo.aggression.typical
             self.__init__(aggression)
+            return
+
+        for content in ("colonisablePlanetIDs", "colonisableOutpostIDs"):
+            # could be that there is only a different species definition file
+            # but could also be some attempt to get in some nasty string.
+            sorted_planets = state[content].items()
```
Cjkjvfnby: `state[content].items().sorted(...)`

default/python/AI/SaveGameManager/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+
+Public functions:
+    encode  -  encode a python object to an unambigious string representation
+    decode  -  decode and interpret a json string to retreive python objects
+    build_savegame_string  - encode the AIstate and return the zlib-compressed result
+    load_savegame_string  - decode a savegame string generated by build_savegame_string
+    assert_content  - helper function to verify the decoded content of a class matches the expectation
+
+Defined Exceptions:
+    CanNotSaveGameException   -  raised when an object could not be encoded
```
Cjkjvfnby: Exception name `Can Not Save` description  `Can not encode`, this is confusing.

default/python/AI/SaveGameManager/_decoder.py
```diff
@@ -0,0 +1,182 @@
...
+import json
+
+import EnumsAI
+from freeorion_tools import profile
+
+from _definitions import *
+
+
+@profile
+def load_savegame_string(string):
```
Cjkjvfnby: It is good place to add type hint for returning value.  PyCharm will have autocomplete on aistate.

"""
rtype: AIState
"""

default/python/AI/SaveGameManager/_decoder.py
```diff
@@ -0,0 +1,182 @@
...
+
+The decoder will only load trusted classes as defined in _definitions.py,
+if an unknown/untrusted object is encountered, it will raise a InvalidSaveGameException.
+
+When classes are loaded, their __setstate__ method will be invoked if available or
+the __dict__ content will be set directly. It is the responsiblity of the trusted classes
+to provide a __setstate__ method to verify and possibly sanitize the content of the passed state.
+
+Example usage:
+    Importing this module directly is not advised.
```
Cjkjvfnby: Importing `_*` is not advised by default, other methods are public.   Importing from __init__ instead of certain file is normal practice.  Is it special case here? Then describe it,  otherwise just remove this.

default/python/AI/SaveGameManager/_decoder.py
```diff
@@ -0,0 +1,182 @@
...
+        # primitive types do not have to be interpreted
+        if x is None or isinstance(x, (int, float, bool)):
+            return x
+
+        # special handling for dicts as they could encode our classes
+        if isinstance(x, dict):
+            return self.__interpret_dict(x)
+
+        # for standard containers, interpret each element
+        if isinstance(x, (list, tuple, set)):
```
Cjkjvfnby: Need some clarification about method argument.  `tuple` and `set` cannot be in `json`.

default/python/AI/SaveGameManager/_decoder.py
```diff
@@ -0,0 +1,182 @@
...
+
+            if x == NONE:
+                return None
+
+        # no special cases apply at this point, just return the value
+        return x
+
+
+def _replace_quote_placeholders(s):
+    """Replace PLACEHOLDER with quotes if not nested within another encoded container"""
```
Cjkjvfnby: Please add examples of quoted strings.

default/python/AI/universe_object.py
```diff
@@ -4,11 +4,13 @@
 from common.configure_logging import convenience_function_references_for_logger
 (debug, info, warn, error, fatal) = convenience_function_references_for_logger(__name__)
 
+
 class UniverseObject(object):
     """
     Stores information about AI target - its id and type.
     :type id: int
     """
+    object_name = 'universe_object'
```
Cjkjvfnby, line 13: It is not part of API so lets it be `_object_name`

default/python/tests/AI/test_savegame_manager.py
```diff
@@ -0,0 +1,144 @@
...
+    try:
+        loaded_obj = SaveGameManager.decode(retval)
+        raise AssertionError("Could load object from untrusted module")
+    except SaveGameManager.InvalidSaveGameException:
+        pass
+
+
+def test_getstate_call():
+    with TrustedScope():
+        try:
```
Cjkjvfnby: https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

Cjkjvfnby (commented):
Looks good.

I did not set any approve/request status, because don't know if I can check your responses before this will be merged.  Non of my comment prevent merging.

default/python/AI/AIstate.py
geoffthemedio: Do both...? Let the player continue with the current program version if they want, but add a note that the loaded save file was created with version X so that the player can revert if they want.

Dilvish-fo: The game will automatically continue, simply recreating a new AIstate rather than using the old stale-versioned (and potentially unsafe) one, and I think there is no need (and no ready way) to pause the game loading and query the user about what they want to do about it.   (As Morlic hinted with the 'magically passed' reference, I don't see how any such older state could even pass through the new encoder, and if it somehow did it would be hopelessly garbled).

The comment line could be somewhat clearer though, rather than referring to no loading an "older savegame" I think it would be more clear to refer to an "older version aistate", and the ConversionError message, which will be presented to the player in the chat window, could more clearly explain that the AIstate will be recreated fresh.

Dilvish-fo (approved):
Looks good to me.  I did suggest a minor additional change related to one of CJs requests, but like him I'm not sure when I'd be able to check back and my comment need not hold anything up.

geoffthemedio:
@Morlic-fo are there any further tweaks coming soon, or should this be merged now?

Morlic-fo:
I pushed some last tweaks/fixes. If noone objects, this is ready for merge as far as I am concerned.

@Vezzra : When you get this into the next test build, please make sure to state that it does break (AI) savegame compatibility so folks should finish their games before updating.

Cjkjvfnby:
Can we just make loading from both versions and save as new version?  This hack can be present for some time (about 2-4 weeks) to make transition smooth.

Cjkjvfnby:
Maybe, but not in this PR.
https://twitter.com/raymondh/status/842777864193769472

Morlic-fo:
In my tests using the default zlib compression from this PR, the length of the string was reduced by about 1% when omitting the whitespace in the list/dict encoder.
I'd rather have more readable strings during debugging.

Cjkjvfnby:
@Morlic-fo, If you don't think https://github.com/freeorion/freeorion/pull/1884#issuecomment-350833878 will work when go ahead and merge it.

Vezzra:
> @Vezzra : When you get this into the next test build, please make sure to state that it does break (AI) savegame compatibility so folks should finish their games before updating.

As I already mentioned elsewhere, I don't think we need to be overly concerned with breaking savegame compatibility, as we don't guarantee savegame compatibility between test builds (not to speak of releases) anyway.

Of course, if you think it necessary, I can add a brief notice about the new test build not being savegame compatible. I'm just a bit wary, because if we give an explicit statement now, players might expect it in future, and start to complain if we don't mention it every time savegame compatibility is broken. Which is something I don't want to have to worry about each time I provide test builds...

Vezzra:
Anything preventing this from being merged? @Morlic-fo, @Dilvish-fo?

dbenage-cx:
Possibly due to auto-resolution during rebase, figured I would note in case it is issue with merge as well.

When checking out this PR, if I rebase to master (as of 965171a):
*  With xml/non-binary saves, there is a (seemingly) infinite wait for AI to send save message.
```
FreeOrionAI.py:163 - Preparing for game save by serializing state
configure_logging.py:204 - Uncaught exception: Traceback (most recent call last):
configure_logging.py:204 -   File "/opt/projects/freeorion/default/python/AI/FreeOrionAI.py", line 166, in prepareForSave
configure_logging.py:204 -     import savegame_codec
configure_logging.py:204 -   File "/opt/projects/freeorion/default/python/AI/savegame_codec/__init__.py", line 45, in <module>
configure_logging.py:204 -     from _decoder import decode, load_savegame_string
configure_logging.py:204 -   File "/opt/projects/freeorion/default/python/AI/savegame_codec/_decoder.py", line 19, in <module>
configure_logging.py:204 -     from _definitions import *
configure_logging.py:204 -   File "/opt/projects/freeorion/default/python/AI/savegame_codec/_definitions.py", line 11, in <module>
configure_logging.py:204 -     fleet_orders.OrderAttack,
configure_logging.py:204 - AttributeError: 'module' object has no attribute 'OrderAttack'
configure_logging.py:204 - 
```
*  Neither manual or auto saves are generated when binary saves are set, no AI log message relating to save game.

Previous *.pyc files were deleted and a new game started in each instance.

Directly checking out this branch without rebasing, saving appears ok.

Morlic-fo:
> Possibly due to auto-resolution during rebase, figured I would note in case it is issue with merge as well.

Thanks for the report. It was a simple issue with classes which were referenced in this PR but already removed in master. Should be fixed.

dbenage-cx:
@Morlic-fo works for me, ty

PR #1891, opened 2017-11-30
Boost 1.61.0, 1.62.0 - Resolve logging of const char ternary result

### LINKED ISSUE #1889: FTBFS: gcc/boost error in universe/ShipDesign.cpp
Compiling with gcc 6.3.0, boost 1.62.0.1 on Debian Linux

universe/ShipDesign.cpp:1608:23: error: invalid initialization of non-const reference of type ‘const char*&’ from an rvalue of type ‘const char*’
        DebugLogger() << "Populating pre-defined ships with "
                      << (are_monsters ? "monster":"ship") << " designs.";
It appears boost with gcc does not like alternates.
Commenting out this statement allows compilation to complete.

dbenage-cx:
Workaround for boost versions 1.61.0 and 1.62.0 (and possibly 1.60.0).
Not an issue with boost versions 1.59.0, 1.63.0, 1.64.0

Assume this is due to use within template, as other similar (non-template) uses work as expected.

Resolves #1889

### DIFF
--- a/universe/ShipDesign.cpp
+++ b/universe/ShipDesign.cpp
@@ -1605,7 +1605,7 @@ namespace {
             return;
 
         DebugLogger() << "Populating pre-defined ships with "
-                      << (are_monsters ? "monster":"ship") << " designs.";
+                      << std::string(are_monsters ? "monster" : "ship") << " designs.";
 
         FillDesignsOrderingAndNameTables(
             *parsed, designs, ordering, name_to_uuid);


LGM-Doyle:
@dbenage-cx thanks for the fix.

dbenage-cx:
@LGM-Doyle np, was possibly fixed with [boost ticket #11998](https://svn.boost.org/trac10/ticket/11998). Testing for resolution (edit: rather, the exact commit that resolved) seemed like needless work ;)

Issue #1732, opened 2017-09-07
Heap-buffer-overflow in creating universe in the second game session.


o01eg:
### Environment

* **FreeOrion Version**: 5f7fd1689d8a92c35af801e80c49ca3b97e4a073
* **Operating System**: Gentoo Linux AMD64
* **Graphic card used**: NVIDIA GeForce 750 Ti
* **Fetched as** 
  * Compiled from source

### Description
The server runs on the same OS. After some turns in the first game session client was disconnected and then reconnected. While starting next game server gets heap-buffer-overflow.

```
==2372==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61100034dc48 at pc 0x7f7ef486290e bp 0x7fffffff8ea0 sp 0x7fffffff8e90
READ of size 2 at 0x61100034dc48 thread T0
    #0 0x7f7ef486290d in Pathfinder::PathfinderImpl::WithinJumpsCacheHit(std::unordered_set<int, std::hash<int>, std::equal_to<int>, std::allocator<int> >*, unsigned long, unsigned long, std::vector<short, std::allocator<short> >&) const /mnt/another/srcs/GIT/freeorion/universe/Pathfinder.cpp:1025
    #1 0x7f7ef487747e in void boost::_mfi::cmf4<void, Pathfinder::PathfinderImpl, std::unordered_set<int, std::hash<int>, std::equal_to<int>, std::allocator<int> >*, unsigned long, unsigned long, std::vector<short, std::allocator<short> >&>::call<Pathfinder::PathfinderImpl const* const, std::unordered_set<int, std::hash<int>, std::equal_to<int>, std::allocator<int> >*, unsigned long, unsigned long, std::vector<short, std::allocator<short> > >(Pathfinder::PathfinderImpl const* const&, void const*, std::unordered_set<int, std::hash<int>, std::equal_to<int>, std::allocator<int> >*&, unsigned long&, unsigned long&, std::vector<short, std::allocator<short> >&) const /usr/include/boost/bind/mem_fn_template.hpp:561
[...]

PR #1893, opened 2017-11-30
Fix bug #1732 pathfinder not cleared in universe clear

LGM-Doyle:
This PR may fix bug #1732.  I was not able to replicate the bug, but I think that I understand it from the AddressSantizer dump in the bug report.

The problem is that the hostless improvement violates assumptions in the server code about the persistence of objects, in this specific case the `Pathfinder`.  While the new universe is being created the pathfinder is used with references to the old galaxy topology.

This PR:
- merges `Universe::ResetUniverse()` into `Universe::Clear()`
- adds creation of a new `Pathfinder` to `Clear()`

This PR will need to be tested by someone who can replicate the problem, preferably with a back trace if it crashes. I suspect that there are other bugs with the same symptom, crashing when starting a second hostless game, because a universe sub object is assuming a persistence that the hostless modification has removed.  The backtrace will determine if it is causes by the `Pathfinder` or some other component.

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -1391,7 +1391,7 @@ void ServerApp::GenerateUniverse(std::map<int, PlayerSetupData>& player_setup_da
     DebugLogger() << "GenerateUniverse with seed: " << seed;
 
     // Reset the universe object for a new universe
-    universe.ResetUniverse();
+    universe.Clear();
 
     // Reset the object id manager for the new empires.
     std::vector<int> empire_ids(player_setup_data.size());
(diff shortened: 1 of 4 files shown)


Vezzra:
@o01eg, you reported #1732, can you check if this PR fixes the issue for you?

o01eg:
I couldn't reproduce old issue but somehow found new another.

Another issue I cann't catch on `master` branch but it happens on this PR while backtrace unrelated with changed files.

geoffthemedio:
If it doesn't break anything, and is considered better, then go ahead anyway...

o01eg:
Looks like #1732 will be fixed with it as that issue happened at game start.

PR #1896, opened 2017-12-01
Add missing line to laenfa file

alleryn:
A line was missing, which caused the pedia not to display Laenfa's broad tolerance.


Vezzra:
Didn't test, but looks reasonable.

PR #1900, opened 2017-12-03
Clarify need for casting certain AIstate attributes

Dilvish-fo:
I was about to simply push these, but noted they would make a minor conflict with #1884, but would have no conflict if this follows that PR.

@Morlic-fo , please merge this after you've merged #1884.

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -87,11 +87,20 @@ class AIstate(object):
     """Stores AI game state.
 
     IMPORTANT:
-    If class members are redefined, added or deleted, then the
+    (i) If class members are redefined, added or deleted, then the
     version number must be increased by 1 and the convert_to_version()
     function must be updated so a saved state from the previous version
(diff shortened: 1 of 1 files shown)


default/python/AI/AIstate.py
```diff
@@ -114,6 +123,7 @@ def __init__(self, aggression):
         # unique ids for turns.  {turn: uid}
         self.turn_uids = {}
 
+        # see AIstate docstring re importance of int cast for aggression
```
Cjkjvfnby, line 126: ` re ` ?

Dilvish-fo: @Cjkjvfnby 

> about; concerning.
"I saw the deputy re the incident"

the above example is [from here](https://www.google.com/search?q=dictionary%3A+re&oq=dictionary%3A+re&aqs=chrome..69i57j69i58.3087j0j7&sourceid=chrome&ie=UTF-8)

you can think of it as akin to an abbreviation for "regarding", although I don't think that's what it actually is.

PR #1902, opened 2017-12-04
Add ccache to Travis CI to improve build times.

LGM-Doyle:
***Problem:*** 
Many of the Travis CI builds are failing due to timeouts, when the code is correct.

***This Solution:*** 
Add ccache to the Travis, ubuntu and OSX builds.

In testing this improved the build times for a warm cache from 45 minutes to 3-4 minutes for ubuntu and 3-6 minutes for OSX.  

The ccache for new branches is pulled from master.

New branches that are  not C++, python and/or FOCS, and smaller C++ changes will build in 3-6 minutes.  Very large changes that touch almost all C++ files may still timeout, but if any files succeed in compiling will still add to the cache.

@adrianbroher, I would appreciate your insight into this.

Thanks.


adrianbroher:
> @adrianbroher, I would appreciate your insight into this.

I like the general idea and the current outcome, but not the implementation.

Currently you place the detection logic within the project CMakeLists.txt, whereas it would be a prime candidate for a CMake module.  In fact such a Module was proposed to CMake upstream (https://gitlab.kitware.com/cmake/cmake/issues/17407) just recently, so we can grab that instead of using our own implementation.

The Dockerfile should not be a part of this PR and instead the SDK one should be extended, but I assume this was just a necessity for testing this PR.

LGM-Doyle:
I'm glad that you like the outcome and I will work to improve the implementation.

I added an [SDK PR](https://github.com/freeorion/freeorion-sdk/pull/36) with the changes to the Dockerfile to add `ccache` and `timeout`.  When it is merged I'll remove the Dockerfile components from this PR.

I've included upstream implementation of the `ccache` module.  It needed some bug fixes that I think are now complete.

adrianbroher:
The `latest` Docker image is now version 7, which incorporates `ccache` and `coreutils` as suggested in this PR and freeorion/freeorion-sdk#36.

Any idea why this PR was failing the last three builds?

LGM-Doyle:
I've updated this PR to use the new Docker image and it improves the linux warm cache build time to 2 minutes.

The last three jobs failed for the following reasons.

The CMake module that I pulled is a WIP and progress and I had to fix a few bugs which prevented the ubuntu builds from finding and using ccache.  I killed two jobs while fixing these.

The third one looks like the OSX build ran until it timed out without finding the cache.  I've tried to fix this, but I won't know until the OSX job works its way through the queue of backlogged jobs.

LGM-Doyle:
The CMake module code is now working for both linux and OSX.

CMakeLists.txt
```diff
@@ -24,13 +24,17 @@ ENDIF(NOT CMAKE_BUILD_TYPE)
 
 message(STATUS "Build type CMAKE_BUILD_TYPE set to ${CMAKE_BUILD_TYPE}")
 
+include(UseCompilerCache)
+FIND_COMPILER_CACHE(PROGRAM ccache)
```
adrianbroher: Function call is upper case, upstream mandates lowercase for most kind of symbols and keywords, except variables.

CMakeLists.txt
```diff
@@ -159,7 +163,6 @@ if(UNIX AND NOT APPLE)
     add_definitions(-DENABLE_BINRELOC)
 endif()
 
-
```
adrianbroher: ?

CMakeLists.txt
```diff
@@ -24,13 +24,17 @@ ENDIF(NOT CMAKE_BUILD_TYPE)
...
+include(UseCompilerCache)
+FIND_COMPILER_CACHE(PROGRAM ccache)
 
 ##
 ## Global project configuration
 ##
 
 project(FreeOrion)
 
+USE_COMPILER_CACHE_WITH_XCODE()
```
adrianbroher: Upper case, should be lower case.

cmake/UseCompilerCache.cmake
```diff
@@ -0,0 +1,144 @@
...
+    endif()
+
+    find_program(CCACHE_PROGRAM ${ARGS_PROGRAM})
+
+    # Quit if required and not found
+    if(REQUIRED AND NOT CCACHE_PROGRAM)
+        message(FATAL_ERROR "Failed to find ${CCACHE_PROGRAM} (REQUIRED)")
+    endif()
+
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
```
adrianbroher: You can't assume that `CCACHE_PROGRAM` is set here.

adrianbroher (changes requested):

LGM-Doyle (commented):
Thank you for the prompt review.

I've made all of the requested changes.

adrianbroher (approved):

PR #1904, opened 2017-12-07
Fix bug in Empire.cpp location condition evaluated without context.

LGM-Doyle:
This was causing errors of the form:
"FollowReference : top level object Source not defined in scripting context",
when right clicking on some items in the ProductionQueue when trying to evaluate the location condition.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -704,7 +704,8 @@ bool ProductionQueue::ProductionItem::EnqueueConditionPassedAt(int location_id) 
             const Condition::ConditionBase* c = bt->EnqueueLocation();
             if (!c)
                 return true;
-            return c->Eval(location_obj);
+            ScriptingContext context(location_obj);
+            return c->Eval(context, location_obj);
         }
         return true;
         break;


geoffthemedio:
This may indicate a bug in the script of the condition. What is specified as a source in other cases where location or enqueue conditions are evaluated?

LGM-Doyle:
Another case taken from "Planet.cpp"  is:
````
std::vector<std::string> Planet::AvailableFoci() const {
    std::vector<std::string> retval;
    auto this_planet = std::dynamic_pointer_cast<const Planet>(UniverseObject::shared_from_this());
    if (!this_planet)
        return retval;
    ScriptingContext context(this_planet);
    if (const auto* species = GetSpecies(this_planet->SpeciesName())) {
        for (const auto& focus_type : species->Foci()) {
            if (const auto* location = focus_type.Location()) {
                if (location->Eval(context, this_planet))
                    retval.push_back(focus_type.Name());
            }
        }
    }
    return retval;
}
````

In that case the `Location` condition is evaluated with a `ScriptingContext` created from the planet object as the first parameter and the planet as the second parameter.

In this PR, the `EnqueueLocation` condition is also evaluated with a `ScriptingContext` created from the location object, (which is a planet) as the first parameter and the location object as the second parameter

PR #1905, opened 2017-12-07
Fix AI attempts to colonize uninhabitable environments

LGM-Doyle:
The AI was attempting to colonize uninhabitable locations, because it was adding in the self-sustaining bonus without considering if the location was habitable.

This generated the error message:
"ColonizeOrder::ExecuteImpl nonzero colonist capacity and planet that ship's species can't colonize"

This PR add details to the error message and adds a check for habitability before applying the self-sustaining or phototroph boni.

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -80,6 +80,7 @@ def outpod_pod_cost():
 def calc_max_pop(planet, species, detail):
     planet_size = _get_planet_size(planet)
     planet_env = species.getPlanetEnvironment(planet.type)
+    uninhabitable = planet_env == fo.planetEnvironment.uninhabitable
     tag_list = list(species.tags) if species else []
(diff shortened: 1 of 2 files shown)


default/python/AI/ColonisationAI.py
```diff
@@ -80,6 +80,7 @@ def outpod_pod_cost():
 def calc_max_pop(planet, species, detail):
     planet_size = _get_planet_size(planet)
     planet_env = species.getPlanetEnvironment(planet.type)
+    uninhabitable = planet_env == fo.planetEnvironment.uninhabitable
```
Morlic-fo: Unless I am overlooking some game mechanic, I think we can just do
```
if planet_env == fo.planetEnvironment.uninhabitable:
    detail.append("Uninhabitable.")
    return 0
```

Dilvish-fo:
Looks good, nice  catch

LGM-Doyle:
@Morlic-fo thanks for the simplification.

PR #1906, opened 2017-12-07
Increase max autosaved turns to 10000

LGM-Doyle:
The current value of 100 is too small for a full game.

Since the user must voluntarily set it, I don't think there is a need for an upper limit.


geoffthemedio:
shrug... ok...

PR #1908, opened 2017-12-09
Remove duplicate supply considerations in InvasionAI

### LINKED ISSUE #1793: AI supply_val calculation for invasion score seems wrong
Relevant piece of code:
```
    if p_sys_id in ColonisationAI.annexable_system_ids:  # TODO: extend to rings
        supply_val = 100
    elif p_sys_id in ColonisationAI.annexable_ring1:
        supply_val = 200
    elif p_sys_id in ColonisationAI.annexable_ring2:
        supply_val = 300
    elif p_sys_id in ColonisationAI.annexable_ring3:
        supply_val = 400
```
 https://github.com/freeorion/freeorion/blob/master/default/python/AI/InvasionAI.py#L437

What seems wrong to me is:
1) ```ColonisationAI.annexable_system_ids``` contains those planets that are hypothetically, in the very best scenario (best specials, tiny planet etc) able to connect to supply.
2) The supply_val gets greater the further we are from supply (but to which we can not possibly connect to).
3) The ```ColonisationAI.evaluate_planet``` already considers the supply (or is this supposed to compensate for a supply penalty there? If so, this needs to be corrected accordingly)

@Dilvish-fo : Thoughts on what is happening and what is supposed to happen?

Morlic-fo:
Resolves #1793.

The supply considerations in InvasionAI are meaningless: Pretty much any system will be in annexable systems - every planet will get the same bonus score.

As far as the general thought is concerned: The relevant ColonizationAI functions already deal with the actual supply rather than some arbitrary guesstimation which happens in annexable_system_ids.

In consequence, the outdated code may be dropped.

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -29,7 +29,6 @@ empire_shipyards = {}
 available_growth_specials = {}
 active_growth_specials = {}
 empire_metabolisms = {}
-annexable_system_ids = set()
 planet_supply_cache = {}  # includes system supply
 all_colony_opportunities = {}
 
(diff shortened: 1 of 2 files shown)


Cjkjvfnby (commented):
Looks ok

Morlic-fo:
I cleaned up the PR by pushing the preceding grooming commit to master.

@Cjkjvfnby : I didn't change the existing code just reordered it for better structure. Anyway, currently we mostly use ```print``` throughout the codebase for historical reasons. I guess the long-term strategy is to convert all print-statements to the correct logging statements eventually.

Dilvish-fo:
> Morlic-fo requested a review from Dilvish-fo 12 days ago

Heya, just got free again today, should be able to catch up on this and other things over the next few days.

>  I guess the long-term strategy is to convert all print-statements to the correct logging statements eventually.

Yes.  Updating this kind of thing anytime you touch a line, in the same commit, makes the more substantive change a bit more work to review, but then saving them all up for some big mass of print-->logger changes would, I think, complicate things for git bisect  (though I suppose it may not be as important for the AI code).  I think I like better the idea that when we otherwise touch such a line, we generally try to immediately/promptly follow up with another commit that makes the print--> logger changes.  How does that sound?

Dilvish-fo:
I agree that this InvasionAI assessment of supply was superfluous (and inferior) to that done by the ColonisationAI evaluation scoring of planets (though it still as plenty of room for further improvement), and this can be discarded.

PR #1910, opened 2017-12-09
Ignore EndGame and Disconnection events in IntroMenu

LGM-Doyle:
The events are sent by the server to all clients after the human client resigns to instruct non-host clients to return to the intro menu.  The host client will already be in the intro menu and should ignore the events.

This fixes these errors:
A EndGame event was not handled by any of these states : [9IntroMenu].  It is being ignored.
A Disconnection event was not handled by any of these states : [9IntroMenu].  It is being ignored.


geoffthemedio:
ok

PR #1866, opened 2017-11-18
Change reduce qi::locals and semantic actions in parser

LGM-Doyle:
This should be considered after PRs #1869 and #1870, which move the value ref parsers into their own compiled unit and generalize the repeated object parser.

This PR reduces the unnecessary use of `qi::locals` and semantic actions in the parser.  

It allows this 
````
rule<some_type (), qi::locals<int, double, string>> some_rule

qi::_1_type _1;
qi::_a_type _a;
qi::_b_type _b;
qi::_c_type _c;

some_rule =      
    int [_a = _1] 
    >> double [ _b = _1] 
    >> string [ _val = construct<some_type>(_a, _b, _1) ]
````
to be written
````
rule<some_type ()> some_rule

qi::_1_type _1;
qi::_2_type _2;
qi::_3_type _3;

some_rule = 
    (int  >> double >> string) [ _val = construct<some_type>(_1, _2, _3) ]
````

PR #1911, opened 2017-12-09
Change remove Labeller double lookup

LGM-Doyle:
This PR should follow #1866.

The current labeller performs a unnecessary double lookup to find the label rule.  It first uses an `extern const char*` to look up the lexer tokenizer for the particular label and then looks up a cached rule.  

Since the tokenizer is visible from all locations where the first lookup table is used, this PR directly uses the tokenizer to find the rule.

It then removes the unused look up table, the `char*` duplicates of the tokenizer names, and the `char*` attribute from the lexer.


LGM-Doyle:
rebased

geoffthemedio:
sorry for the delay in addressing these... I'm doing some actual substantial coding in the meter effects and noisiness branches, which involve a substantial amount of parser changes, which I'd rather not need to completely rework the changes for if these changes go in first...

LGM-Doyle:
Not the news, that I was hoping to hear, but thanks for the update.

The changes may not be as difficult to merge as you imagine.  The intent of these changes is to make the boost spirit parsing code closer to a one to one correspondence with the FOCS code so that devs can more easily add/modify FOCS without fully understanding the boost spirit code.

Dilvish-fo:
> [Geoff]  I'm doing some actual substantial coding in the meter effects and noisiness branches, which involve a substantial amount of parser changes

> [LGM-Doyle] The changes may not be as difficult to merge as you imagine.   The intent of these changes is to make the boost spirit parsing code closer to a one to one correspondence with the FOCS code...

@geoffthemedio  @LGM-Doyle  Perhaps I'm just being a busybody about other people's work, but this is a logjam I'd like to try help clearing (with suggestions about stuff other people could do 😄  )

I can certainly sympathize with Geoff's concern, because if there is one person here with more on their FO plate than LGM-Doyle it would have to be Geoff.

It looks to me like most of the parser changes Geoff referenced were in the already-merged Meter Growth PR, and that only a modest bit remains in the Noisiness PR.   That bit though does include a Labeller reference which appears to me would be a minor conflict.

Given that LGM-Doyle both believes that his changes should be easy to adapt to, and is of course most familiar with them, I'd like to suggest that LGM-Doyle rebase his ```change_remove_labeller_double_lookup``` branch and add that as a branch directly here in the freeorion repo, and then make second branch ```Noisiness2```  here in the freeorion repo and corresponding PR from that second branch to his first branch, in which he adds all of Geoff's commits in the ```Noisiness``` branch together with the required parser reconciliation (which appears to me would likely be quite modest).  Then Geoff could review that and get comfortable with merging this PR and switch over to working with the Noisiness2 branch.  And if there were similar concerns/conflicts with the other parser PRs then perhaps the same approach could be taken with them, one by one.

Ideally, these parser changes might even be able to mildly facilitate Geoff's further progress with the Noisiness(2) branch, it seems to me.

PR #1912, opened 2017-12-10
Tune up the stockpile 

agrrr3:
In the forum is the consensus that the imperial stockpile is basically worthless with the current settings.
Discussion here [http://www.freeorion.org/forum/viewtopic.php?f=28&t=10810&p=90852#p90852](url)

This PR adjusts the basic values to something actually useful. I'd  like to merge this and ask for the second round of play test feedback.

* make the currently useless imperial stockpile more useful
** better transfer ratio with early techs
** higher extraction limit with later techs
* fix tabs to spaces
```
                           cost Tu  Transfer Ratio  Use Limit
Predictive Stockpiling      0RP  1   +0.2 -> +0.4   +2 ->  +3
Generic Supplies           40RP  4       +0.2       +3 ->  +7
Transcendent Design       150RP  5      +0.00          +0
Interspecies Academy (6x)  30PP  5      +0.05       +2 -> +15
Interstellar Entangle Fac 200RP  5      +0.20       +5 -> +15
```

### DIFF
--- a/default/scripting/buildings/INTERSPECIES_ACADEMY.focs.txt
+++ b/default/scripting/buildings/INTERSPECIES_ACADEMY.focs.txt
@@ -78,7 +78,7 @@ BuildingType
             priority = [[LATE_PRIORITY]]
             effects = [
                 SetEmpireMeter empire = Source.Owner meter = "METER_IMPERIAL_PP_TRANSFER_EFFICIENCY" value = min(1.0, Value + 0.05)
-                SetEmpireMeter empire = Source.Owner meter = "METER_IMPERIAL_PP_USE_LIMIT"           value = Value + 2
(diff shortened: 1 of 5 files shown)


agrrr3:
Hi I would like to merge this ASAP so we get it in the test build (and change the settings in the next iteration after feedback). 

Note that i think we will have to nerf Transcendent Design and the Entanglement Factory (probably toning them a bit down and making more expensive; maybe shifting). 
PP wise actually the Academies are cheap to build but expensive,in the sense that you need to put 600PP in the stockpile to reach break even with the 30 PP (so 3600PP if you build the six of them); and only if the bonus doesnt get capped at 100%.

We need to get playtesting reports from the players. 
As we can do more tweaking next iteration i dont see a reason for this not to merge.

PR #1916, opened 2017-12-11
Prevent Fleet::Copy() from referencing unknown systems

LGM-Doyle:
This PR prevents `Fleet::Copy()` from copying `m_next_system` and `m_prev_system` when the system is not visible to the referenced empire.

This was creating an error in the human client when it tried to follow a path to an invisible `m_next_system`

### DIFF
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -1535,7 +1535,10 @@ void FleetDataPanel::Refresh() {
             add_overlay("bombarding.png");
         }
 
-        // Moving fleets can't be gifted.
+        // Moving fleets can't be gifted.  The order will be automatically
+        // cancelled on the server.  This make the UI appear to cancel the
+        // order when the ship is moved without requiring the player to
+        // re-order the gifting if the ship is stopped.
         if (fleet->OrderedGivenToEmpire() != ALL_EMPIRES && fleet->TravelRoute().empty())
             add_overlay("gifting.png");
 
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -1534,7 +1534,9 @@ void FleetDataPanel::Refresh() {
         {
             add_overlay("bombarding.png");
         }
-        if (fleet->OrderedGivenToEmpire() != ALL_EMPIRES)
+
+        // Moving fleets can't be gifted.
+        if (fleet->OrderedGivenToEmpire() != ALL_EMPIRES && fleet->TravelRoute().empty())
             add_overlay("gifting.png");
 
         if ((fleet->GetVisibility(client_empire_id) < VIS_BASIC_VISIBILITY)
@@ -3587,6 +3589,7 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
     }
 
     if (fleet->OwnedBy(client_empire_id)
+        && fleet->TravelRoute().empty()
         && !peaceful_empires_in_system.empty()
         && !ClientPlayerIsModerator())
     {
--- a/universe/Fleet.cpp
+++ b/universe/Fleet.cpp
@@ -84,7 +84,7 @@ namespace {
         system not known to the \p empire_id. */
     std::list<int> TruncateRouteToEndAtSystem(const std::list<int>& full_route, int empire_id, int last_system) {
 
-        if (full_route.empty())
+        if (full_route.empty() || (last_system == INVALID_OBJECT_ID))
             return std::list<int>();
 
         auto visible_end_it = full_route.cend();
--- a/universe/Fleet.cpp
+++ b/universe/Fleet.cpp
@@ -162,20 +162,18 @@ void Fleet::Copy(std::shared_ptr<const UniverseObject> copied_object, int empire
             if (Unowned())
                 m_name =                  copied_fleet->m_name;
 
-            if (vis >= VIS_FULL_VISIBILITY) {
-                m_travel_route =              copied_fleet->m_travel_route;
+            // Truncate the travel route to only systems known to empire_id
+            int moving_to = (vis >= VIS_FULL_VISIBILITY
+                             ? (!copied_fleet->m_travel_route.empty()
+                                ? copied_fleet->m_travel_route.back()
+                                : INVALID_OBJECT_ID)
+                             : m_next_system);
+
+            m_travel_route = TruncateRouteToEndAtSystem(copied_fleet->m_travel_route, empire_id, moving_to);
+
+
+            if (vis >= VIS_FULL_VISIBILITY)
                 m_ordered_given_to_empire_id =copied_fleet->m_ordered_given_to_empire_id;
-
-            } else {
-                int             moving_to =         m_next_system;
-                std::list<int>  travel_route;
-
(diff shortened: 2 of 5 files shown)


geoffthemedio:
What happens if a fleet is transferred to a new owner? Could an empire then own a ship but not be informed of where it's actually moving to before the final destination is its next system?

LGM-Doyle:
Currently, gifting requires that the gifted item be in a system where the receiving empire has at least one  other object.  It is possible to gift a moving fleet.  However, since there is already an object in the system it is not possible, for the receiving empire to not know of the star lanes, and I think therefore the terminating systems.

I have added the restriction that it is not possible to gift a moving fleet.  This prevents empires gifting a fleet that is moving into danger, or to attack, since the receiving empire does not have an opportunity to change this order before it is executed.

geoffthemedio:
I was thinking more of effect-mediated ownership changes, like psychogenic domination.

LGM-Doyle:
I don't think that this is a problem for the following reasons.

Effect mediated ownership changes will happen on the server in `PostCombatProcessTurns()` and be followed by a visibility update.
````
void ServerApp::PostCombatProcessTurns() {
...
    m_universe.ApplyAllEffectsAndUpdateMeters(false);
...
    m_universe.UpdateEmpireObjectVisibilities();
...
````

`UpdateEmpireObjectVisibilities()` calls `SetTravelledStarlanEndpointsVisible()` which grants basic visibility of systems from/to which owned fleets are traveling.
````
void Universe::UpdateEmpireObjectVisibilities() {
...
    SetTravelledStarlaneEndpointsVisible(Objects(), m_empire_object_visibility);
...
````

It is only after this that the fleets are copied out to the individual empires/clients. A copied fleet ***owned by a different empire*** would lose knowledge of where it is going to/coming from if the empire/client does not already know of the system.

**Comment by geoffthemedio** [comment]
I'm thinking of the final destination / parts of the move route past the next system a fleet is moving to.

**Comment by LGM-Doyle** [comment]
I think that I understand your concern(, but if not I will try again).  Now in the event of a transfer of ownership, the new owner will gain knowledge of the systems to/from which it is going, but not gain knowledge or dangling references to any further systems on the travel route.

While looking at the code I found (and fixed) more bugs:
- `Fleet::ShortenRouteToEndAtSystem()` has a direct reference to `this->EmpireID(), which is the wrong empire id during a copy
- Some locations calling `Fleet::ShortenRouteToEndAtSystem()` assumed incorrectly that it shortened the passed in route, when it actually shortens `m_travel_route` and returns it via the passed in parameter.
- `m_travel_distance` is incorrectly updated assuming that the portions of the route are always the shortest path.  This is no longer true, because both human and AI clients can create arbitrary routes that are not the shortest path between the end points.

In addition, `m_travel_distance` is never accessed, nor is it accessible via any public function.  I removed `m_travel_distance`

I refactored  `void Fleet::ShortenRouteToEndAtSystem(route&, final_system)` into `route TruncateRouteToEndAtSystem(const& route, empire_id, final_system)`, which passes the candidate route in and returns a truncated route.

I refactored `Fleet::Copy()` so that `m_travel_route` is always filtered for systems known to the receiving empire, regardless of whether it is the owner.

This means for an effect mediated transfer of fleet ownership, the new owner will always gain knowledge of the next system on the route, but further segments of the travel route will be truncated at the first system that the new owner does not have prior knowledge of.

File: universe/Fleet.cpp [hunk 1]
```diff
@@ -79,6 +79,34 @@ namespace {
...
+            // if requested last system not in route, do nothing
+            if (visible_end_it == full_route.end())
+                return std::list<int>();
+
+            ++visible_end_it;
+        }
+
+        // remove any extra systems from the route after the apparent destination
+        auto end_it = std::find_if(full_route.begin(), visible_end_it,
+                                   boost::bind(&SystemHasNoVisibleStarlanes, _1, empire_id));
```
geoffthemedio: this is a bit misleading due to that function's name... maybe add a comment that this also excludes systems not known to the specified empire?

geoffthemedio: also maybe a comment noting that if a ship is in a system, or is moving between two systems, then those systems should always be visible to the empire that owns the ship...

util/SerializeUniverse.cpp
```diff
@@ -250,9 +251,12 @@ void Fleet::serialize(Archive& ar, const unsigned int version)
         & BOOST_SERIALIZATION_NVP(m_next_system)
         & BOOST_SERIALIZATION_NVP(m_aggressive)
         & BOOST_SERIALIZATION_NVP(m_ordered_given_to_empire_id)
-        & BOOST_SERIALIZATION_NVP(m_travel_route)
-        & BOOST_SERIALIZATION_NVP(m_travel_distance)
-        & BOOST_SERIALIZATION_NVP(m_arrived_this_turn)
+        & BOOST_SERIALIZATION_NVP(m_travel_route);
+    if (version < 3) {
+        double dummy_travel_distance;
+        ar & boost::serialization::make_nvp("m_travel_distance", dummy_travel_distance);
```
geoffthemedio, line 257: could just call the dummy variable `m_travel_distance` and use `BOOST_SERIALIZATION_NVP`

UI/FleetWnd.cpp
```diff
@@ -1534,7 +1534,9 @@ void FleetDataPanel::Refresh() {
         {
             add_overlay("bombarding.png");
         }
-        if (fleet->OrderedGivenToEmpire() != ALL_EMPIRES)
+
+        // Moving fleets can't be gifted.
+        if (fleet->OrderedGivenToEmpire() != ALL_EMPIRES && fleet->TravelRoute().empty())
```
geoffthemedio: this doesn't make sense... if the situation somehow arises that a moving fleet is being gifted, it should be displayed. enforcing such a rule should be done where the gifting order is issued and where it is executed, not just when displaying the indicator.

geoffthemedio:
otherwise looks OK...

UI/FleetWnd.cpp
LGM-Doyle: The rule is enforced on the server in the gifting code.

However, in the UI without this line of code, gifting a ship and then moving it still shows the gifting icon, even though the gift will not happen on the server.  This code makes the result of the order set be reflected in the UI.

 I added a comment to clarify this.

Another approach is the cancel the gifting when creating the move order in the UI, but then the order needs to be manually restored when a fleet is stopped.  I found that the proposed  approach takes fewer clicks to cancel a gifting order by moving a fleet and then restore the order by stopping the fleet from moving.  In addition, it is clear that the moving is cancelling the order because stopping the fleet restores the gifting icon.

universe/Fleet.cpp
LGM-Doyle: I expanded the comment.

util/SerializeUniverse.cpp
LGM-Doyle: I did not change this.

Calling the variable `m_travel_distance`, implies for a casual reader that there is a member variable due to our naming conventions.  There are reasonable scenarios where a developer code remove/modify the line of code because they know the `m_travel_distance` is not a member variable and break the deserialization.

The proposed line of code can only be read as serialize/deserialize a double in the archive named `m_travel_distance` into a variable called `dummy_travel_distance`.  The name `dummy_X` implies that it is not used.

PR #1917, opened 2017-12-13
Change g_indent into ntabs parameter of Dump(unsigned int ntabs)

LGM-Doyle:
#### Problem ####
Multiple threads increment/decrement `g_indent` without any synchonization.  

I think that this leads to rare buggy behavior every few thousand turns.  It causes data corruption, log corruption and/or crashes.

#### This Solution ####
Convert `g_indent` from a global that requires synchronization to a parameter of `Dump()`

### DIFF
--- a/universe/Building.cpp
+++ b/universe/Building.cpp
@@ -204,58 +204,46 @@ void BuildingType::Init() {
 
 std::string BuildingType::Dump(unsigned short ntabs) const {
     std::string retval = DumpIndent(ntabs) + "BuildingType\n";
-    ++ntabs;
-    retval += DumpIndent(ntabs) + "name = \"" + m_name + "\"\n";
-    retval += DumpIndent(ntabs) + "description = \"" + m_description + "\"\n";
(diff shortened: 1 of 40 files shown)


universe/Building.cpp
```diff
@@ -228,34 +228,34 @@ std::string BuildingType::Dump() const {
...
 
     if (m_location) {
-        retval += DumpIndent() + "location = \n";
-        ++g_indent;
-            retval += m_location->Dump();
-        --g_indent;
+        retval += DumpIndent(ntabs) + "location = \n";
+        ++ntabs;
+            retval += m_location->Dump(ntabs);
+        --ntabs;
```
dbenage-cx: These simple instances of increment, pass, decrement could now be reduced to `->Dump(ntabs + 1)`

dbenage-cx (approved):
LGTM, consider comment optional.

universe/Building.cpp
LGM-Doyle: Good point.  I made the changes.

geoffthemedio:
Looks reasonable after skimming. OK assuming it's been tested at least a bit.

Issue #1920, opened 2017-12-16
Homeworld population bonus appears as Unknown


alleryn:
Forum thread: http://www.freeorion.org/forum/viewtopic.php?f=25&t=10819&p=91005#p91005

### Environment
* **FreeOrion Version**: 110ba9f - 33c6468
* **Operating System**: Windows 10 Home
* **Fetched as**
  * Compiled from source

### Description
Starting in 110ba9f (Make Universe:: ResetUniverse() and Clear() into single function.), hovering over the population on a planet now displays the "Homeworld" bonus as "Unknown" and the bonus disappears entirely upon loading a saved game.  Was working in the previous commit b1ab67a.

Logs, save, and screenshot available in forum thread (link at top of this post).

### Expected Result
Hovering should list the HOMEWORLD_BONUS_POPULATION as "Homeworld" and not as "Unkown".  The bonus should remain present after loading a saved game.

### Steps to reproduce
* Start a single player game.
* Hover over the human player's homeworld's population.
* Note that the last entry is "Unknown +<(2*Planetsize)>" (e.g. "Unknown +8.00 for a large homeworld")
* Save, or allow the game to autosave by advancing to the next turn.
* Load the save.
* Advance a turn.
* Hover over the human player's homeworld's population.
* Note that the homeworld population bonus has disappeared.

dbenage-cx:
@alleryn TY for finding the related commit!

Confirmed on linux with recent master (may require next turn after load)

alleryn:
> may require next turn after load

Oops, thanks i skipped a step.  Edited that in above.

Vezzra:
@LGM-Doyle, that's one of your changes. I know you'll be gone tomorrow already, so you most likely won't have time to look into or even fix this before you go. But if there is any chance you could at least give it a quick glance and some pointers as to what might be the cause, that could help someone else working on it in the meantime...

alleryn:
Fixed in ba69006

PR #1921, opened 2017-12-17
Only ClearSpeciesHomeworlds in universe generation.

LGM-Doyle:
This fixes issue #1920.

Commit 110ba9fbb3b5ce016b9938649b423f9e91736941 changed
GetSpeciesManager().ClearSpeciesHomeworlds() from happening only in
GenerateUniverse() into Universe::Clear() which clears the homeworld
bonus incorrectly.

This commit restores clearing the species homeworlds only in generate
universe.

Thanks @alleryn for reporting and doing the bisection for this bug.
Thanks @Vezzra for shouting me down.

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -1392,6 +1392,7 @@ void ServerApp::GenerateUniverse(std::map<int, PlayerSetupData>& player_setup_da
 
     // Reset the universe object for a new universe
     universe.Clear();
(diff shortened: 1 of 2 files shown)


Vezzra:
> Thanks @Vezzra for shouting me down.

I sincerely hope that my mild request for assistance didn't really come across as "shouting"... 😉

geoffthemedio:
I think a better choice of words would be "flagging down".

PR #1930, opened 2017-12-31
Stockpile Limit Meter

geoffthemedio:
Stockpile input and output are limited by a new planetary stockpile meter. Species can be good or bad at stockpiling. There is a stockpile focus, currently only available to Laefna.


Empire/Empire.cpp
```diff
@@ -142,7 +142,7 @@ namespace {
             if (excess_here < EPSILON)
                 continue;
             // Transfer excess to stockpile
-            new_contributions += excess_here * stockpile_transfer_efficiency;
+            new_contributions = std::max(new_contributions + excess_here, stockpile_limit);
```
Dilvish-fo: is this revised calc right, particularly in light of how it gets used in the following line?
```return starting_stockpile + new_contributions - stockpile_used;```

  For one, it seems to me it should be an std::min operation rather than max, and second, suppose that ```starting_stockpile``` was already at the ```stockpile_limit```, and that none is used, and ```new_contributions``` cap out at the stockpile limit then that return value could be up to twice the ```stockpile_limit``` 

I think I'd suggest letting this line 145 be simply something like 
``` new_contributions += excess_here```

and then putting the limit in the return line
```return std::min(starting_stockpile + new_contributions - stockpile_used, stockpile_limit);```

File: Empire/Empire.cpp [hunk 2]
```diff
@@ -1024,8 +1040,8 @@ void ProductionQueue::Update() {
     auto available_pp = AvailablePP(industry_resource_pool);
     float pp_in_stockpile = industry_resource_pool->Stockpile();
     TraceLogger() << "========= pp_in_stockpile:     " << pp_in_stockpile << " ========";
-    const Meter* pp_use_limit = empire->GetMeter("METER_IMPERIAL_PP_USE_LIMIT");
-    float available_stockpile = pp_use_limit? std::min(pp_in_stockpile, pp_use_limit->Current()) : pp_in_stockpile;
+    float stockpile_limit = StockpileCapacity();
+    float available_stockpile = std::min(pp_in_stockpile, stockpile_limit);
```
Dilvish-fo: It seems like there are a few places, like here, and the next change below, where the previous reference to ```pp_use_limit->Current()``` has been changed to ```stockpile_limit```, but given the new meaning for the limit it seems to me the reference isn't needed here at all-- I didn't track down where the actual stockpile value is set, but I expect that you subject it to the limit at that time, and so here we perhaps don't need the new variable anymore, or at least could simply say
``` float available_stockpile = pp_in_stockpile```

File: Empire/Empire.cpp [hunk 3]
```diff
@@ -1182,7 +1198,7 @@ void ProductionQueue::Update() {
         sim_pp_in_stockpile = CalculateNewStockpile(m_empire_id, sim_pp_in_stockpile,
                                                     available_pp, allocated_pp,
                                                     allocated_stockpile_pp);
-        sim_available_stockpile = std::min(sim_pp_in_stockpile, pp_use_limit->Current());
+        sim_available_stockpile = std::min(sim_pp_in_stockpile, stockpile_limit);
```
Dilvish-fo: It may have seemed like this min operation was still needed here simply because of the above-mentioned bug with ```CalculateNewStockpile```, but I don't think it is.

Dilvish-fo:
Hmm, looks like my kids distracted me long enough while I was in the middle of reviewing that I forgot I was looking at an individual commit, not the final combined code, but I think the comments are still largely applicable....

Empire/Empire.cpp
geoffthemedio: Current version is (w/o comments and log line):
````
            new_contributions += excess_here;
        }
        if (new_contributions > stockpile_limit)
            new_contributions = stockpile_limit;
        return starting_stockpile + new_contributions - stockpile_used;
````
Doing an `std::min` at the end with `stockpile_used` isn't needed, as it's supposed to limit the inputs, not the inputs after the withdrawals are deducted, and the above `if` lines limit the contributions to `stockpile_limit`.

File: Empire/Empire.cpp (hunk 3 above)
Comment (geoffthemedio) [reply Empire/Empire.cpp]: I think it makes sense... What is available for use is the lesser of what's in the stockpile and the per-turn withdrawal limit.

File: Empire/Empire.cpp (hunk 2 above)
Comment (geoffthemedio) [reply Empire/Empire.cpp]: Here it needs to know what amount of stockpile PP is available (the lesser of what is in the stockpile and how much can be withdrawn per turn) when setting the spending on each item on the production queue, not just when calculating the new stockpile amount for the next turn based on the results of those allocations.

File: Empire/Empire.cpp [hunk 4]
```diff
@@ -143,13 +144,13 @@ namespace {
...
-            TraceLogger() << "allocated here: " << allocated_here
-                          << "  excess here: " << excess_here
+            new_contributions += excess_here;
+            TraceLogger() << "...allocated in group: " << allocated_here
+                          << "  excess in group: " << excess_here
                           << "  to stockpile: " << new_contributions;
         }
-        TraceLogger() << "starting_stockpile: " << starting_stockpile 
-                      << "  stockpile_used: " << stockpile_used;
+        if (new_contributions > stockpile_limit)
```
agrrr3, line 152: Instead of this the cut off should happen before the trace logger in line 148 or the trace statement " to stockpile:" in line 150 should move behind or into the if block.

agrrr3:
Great branch :)
Some notes from experience. 

Gating tech effects by focus is a good idea, it captures the idea of dedication. I guess there should be a bonus based on population as well, the +3 boost is too low given the dedication of a bigger colony. But its really hard to establish numbers if we dont define what should/shouldnt easily be buildable.

Ppl in the forums say they dont really want stockpiling to be an option in a nondedicated empire, the difference of factor 2 between average species and the distributed empire species is probably too low. The current value for the laenfa as distributed empire species of 0.04 is way to low IMHO. In my tests it worked with a 0.1 species and technology having a single disconnected production planet. And with 0.3 the distributed empire worked.
Not sure if one should strive for keeping the general multipliers (i.e. either make Laefna ultimate_ or most of the species bad_ ). Usually you want the most of the species to be average_ but ultimate sounds extreme.

Its not clear yet if the stockpile ratio is not needed as there are micromanagement issues with input limits. Round 3 should establish what the issues are exactly and if there are ways to amend this. (making the input limit 5 times higher than the output limit; or switching to a stockpile limit or completely removing the input limits relying only on the extraction limit).

So i dont know how to proceed for round three. Should we merge (e.g. the other branch contains a bugfix for the interspecies academy effect activation) the two branches? Should we add the SP_SLY so laenfa is the stealthy empire species and the sly is the distributed empire species?
Again I want to get community playtesting feedback as fast as possible, so i want round three to go into the next test builds (next monday?)

Empire/Empire.cpp
geoffthemedio: I made it the way it is so that the excess in each group can be seen independently. A separate log line could be added for the limit-restricted contributions if needed.

default/scripting/buildings/INTERSPECIES_ACADEMY.focs.txt
```diff
@@ -76,32 +77,8 @@ BuildingType
             activation = Source
             accountinglabel = "INTERSPECIES_ACADEMY_LABEL"
             priority = [[LATE_PRIORITY]]
-            effects = [
```
geoffthemedio, line 79: @agrrr3 not really sure what this building is supposed to do or should do with this setup... alternatives requested.

agrrr3: In this branch each academy boosts the local max contribution to the global limit like the other tech but without needing the focus set to stockpile. I think this as okayish.

Compared to tech the idea of the building is that is a more natural fit for distributed empires (or large empires) because you need space between the buildings. I would like to have such a tech in the game.

One could gate the effect also by focus, but this would make it less distinguished from the other techs. With the other techs you can make any number of planets contribute, that is not possible with the academy. Not sure on this one.

Another use would be to make the species on the planet a better stockpiling species (so that the whole population of that in the species would contribute more), then we would need to make the building a lot more expensive probably. And to make it more strategic, we could make the effect interruptible. Only boost the the stockpiling if there were no enemy ships in orbit the last turn.

Or we could make it a research bonus instead of stockpiling bonus, because exchange between people and ideas is how research flourishes. Then we need to remove the "design" part from the fluff, because design is not about research but about application.

geoffthemedio:
The average / great / ultimate stockpiling currently just affects the base population-derived stockpiling effects, but not the focus-gated tech-related effects... Those depend on the focus, which currently only Laefna have access to.

As noted, the academy had a bunch of effects removed in this pull request as I wasn't sure what to do with it with this mechanism for stockpile limits. It needs rewriting anyway, so if you can provide a diff or patch or additional pull request for that, please do...

Empire/Empire.cpp
agrrr3: Ah, i missed the accumulation in  new_contributions.  Forget my comment.

agrrr3:
Oh so you made the tech the bigger help in stockpiling and not the species bonus (i made it the other way round) and made the tech available only for the stockpiling species. Not sure I agree with this. A tech which is only usable by one species is a bit of a downer. Guess I need to play first and give more informed feedback then.

geoffthemedio:
I didn't put much thought into balance... just got the mechanics working. Adjust as you see fit...

geoffthemedio:
Superseded by updated https://github.com/freeorion/freeorion/pull/1928

default/scripting/buildings/INTERSPECIES_ACADEMY.focs.txt
Vezzra: > Or we could make it a research bonus instead of stockpiling bonus, because exchange between people and ideas is how research flourishes. Then we need to remove the "design" part from the fluff, because design is not about research but about application.

Sorry for chiming in only now, but I really want to voice my support for this idea. It just makes so much more sense for a building like this...

PR #1931, opened 2018-01-01
Adding OpenBSD support, fallback paths related to cmake

devnexen:
settings rather than hardcoded.


network/Networking.cpp
```diff
@@ -13,7 +13,11 @@ namespace {
 namespace Networking {
     const std::string DISCOVERY_QUESTION = "Yo, can I play Free-O here, dog?";
     const std::string DISCOVERY_ANSWER = "Word!";
+#ifdef FREEORION_OPENBSD
```
adrianbroher, line 16: What is the point of this condition?  Why does OpenBSD need a different linger time?

adrianbroher: Understood. Can you add a comment for this on the change?

adrianbroher:
Looks good so far.

PR #1934, opened 2018-01-06
Sitreport We got incoming enemy ships

agrrr3:
* Defined via a tech spec via custom_sitreps.txt
* I need to test this a bit
  
[Forum discussion](http://freeorion.org/forum/viewtopic.php?f=6&t=10839)

### DIFF
--- a/default/customizations/custom_sitreps.txt
+++ b/default/customizations/custom_sitreps.txt
@@ -105,6 +105,53 @@ EffectsGroup
             empire = Source.Owner
 
 
+EffectsGroup
+    scope = And [
+        System
+        // The following assets need protection
+        Contains And [
+            OwnedBy empire = Source.Owner
+            Or [
+                Planet
+                Ship
+            ]
+        ]
+        // Dont warn about system where there are already enemies this turn
+        Not Contains And [
+            Ship
+            Or [
+                OwnedBy affiliation = EnemyOf empire = Source.Owner
+                UnOwned
+            ]
+	    VisibleToEmpire empire = Source.Owner
+        ]
+        // Only warn if enemies are incoming
+        WithinDistance distance = 120 condition = And [
+            Fleet
+            Not Stationary
+            Or [
+                OwnedBy affiliation = EnemyOf empire = Source.Owner
+                UnOwned
+            ]
+	    VisibleToEmpire empire = Source.Owner
+            Turn low = [[ETA_NEXT_TURN]] high = [[ETA_NEXT_TURN]]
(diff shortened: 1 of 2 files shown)


default/customizations/custom_sitreps.txt
```diff
@@ -105,6 +105,51 @@ EffectsGroup
...
+        System
+        // The following assets need protection
+        Contains And [
+            OwnedBy empire = Source.Owner
+            Or [
+                Planet
+                Ship
+            ]
+        ]
+        // Dont warn about system where there are already enemies this turn
```
geoffthemedio, line 119: Do you want to never warn if there are any enemies in the system... or:
-Enemies with the same owner as those arriving
-Enemies visible to the owner the recipient of the message

default/customizations/custom_sitreps.txt
```diff
@@ -105,6 +105,51 @@ EffectsGroup
...
+        Not Contains And [
+            Ship
+            Or [
+                OwnedBy affiliation = EnemyOf empire = Source.Owner
+                UnOwned
+            ]
+        ]
+        // Only warn if enemies are incoming
```
geoffthemedio: Should probably check that the incoming enemies are visible to the recipient.

default/customizations/custom_sitreps.txt
agrrr3: The warning is to draw attention to the system. If there was already combat, you have already a message for doing that.

I didnt know that i have to script a visibility check. If I remember correctly, without OwnedBy empire = Source.Owner i did not get matches outside my known empire without checking visibility, is that the way it works? Does not sound right.

geoffthemedio: I don't understand your question.

Effects run on the server based on the actual universe, not the universe as seen by a particular player. If you don't add a visibility test to some effect, it won't add it automatically.

agrrr3: That makes more sense to me. So probably I had some other condition which limited the results. If not there should have been a lot of sitreps about system i did not know of yet.

geoffthemedio: Not about systems your empire can't see, but about ships that aren't visible to the player's empire that are arriving in systems you can see. The scope already limits it to systems containing objects owned by the empire receiving the sitrep messages.

agrrr3:
Fixed the visibility checks, sitrep label. Rebased.

geoffthemedio:
Looks ok. Works as intended for you?

agrrr3:
Works fine :)

PR #1939, opened 2018-01-10
Fix CI build for OSX

dbenage-cx:
Minor attempt to fix travis errors, will look into this more tonight

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
       env:
         - BUILD_UNITTEST=true
     - os: osx
-      osx_image: xcode8.1
+      osx_image: xcode8.3
       compiler: clang
   allow_failures:
     - env:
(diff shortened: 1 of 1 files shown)


geoffthemedio:
It's failing with a time limit now, rather than immediately, so that's an improvement.

adrianbroher:
The `brew update` (line 35) statement should be obsolete too when raising the OSX image to version 8.3 (line 21), which would shave off another 2 to 3 minutes of build time.  The previous 8.1 version was retired some time ago by TravisCI anyway.

https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce

dbenage-cx:
@adrianbroher TY for link, I removed the request for a specific `osx_image` to lower a need to update the config each time the image is deprecated (but maybe you prefer we not implicitly follow the default?).

I did not find anything around the 28 December time-frame (roughly when builds stopped working), so assume this was when the deprecated 8.1 image was removed.

adrianbroher:
@dbenage-cx Yes, I would prefer to keep an explicit version instead of praying for the current default to work.

PR #1941, opened 2018-01-11
Move Meter Growth Into Effects

geoffthemedio:
Move current meter growth from C++ into scripted effects.


agrrr3 (changes requested):
Cant comment directly on the relevant source snippets it seems....

Decreasing stockpile limit is missing
STANDARD_STOCKPILE_GROWTH

I also find the formula for in/decrease confusing and not consistently used (e.g. different in construction).
Also dont see how force-energy structures could work with the current code.
Maybe a meters file which provides a macro for in/decrease for consistency and easier changeability?

geoffthemedio:
> Decreasing stockpile limit is missing
> STANDARD_STOCKPILE_GROWTH

Not sure what you mean by "Decreasing stockpile limit", but yes, I've forgotten to add the `STANDARD_STOCKPILE_GROWTH` macro to any of the general stockpile macros. Edit: Added. /Edit

> I also find the formula for in/decrease confusing

It is a complicated formula, but writing the "move towards target by at most one point" like https://github.com/freeorion/freeorion/pull/1941/files#diff-979b9482dcd97f45895a4b5370c1b593R31 doesn't require pre-screening with a scope or activation condition, which may not be evaluated using the same meter values as will be present when the effect actually executes, so wouldn't work for during-a-turn focus changes and meter changes that depend on them.

> not consistently used (e.g. different in construction)

Thought I'd updated the construction script, but apparently I haven't, or haven't pushed it yet. Edit: Pushed /Edit

> Also dont see how force-energy structures could work with the current code.

It should still work fine, but may not have its effects properly registered in the predicted meter changes on the turn of a focus change. Should be easily updated to use `Value(...)` expressions, though. Edit: Done /Edit

> Maybe a meters file which provides a macro for in/decrease for consistency and easier changeability?

Could probably be done.

geoffthemedio:
Seems to be working for me. Any objections?

default/scripting/species/common/general.macros
```diff
@@ -13,15 +24,19 @@ FOCUS_CHANGE_PENALTY
             activation = And [
                 Planet
                 Not Focus type = "FOCUS_INDUSTRY"
+                (0 <= Source.TurnsSinceFocusChange < 1)
```
dbenage-cx: Significance of writing it this way?
Seems to activate only on a value of 0, so the effect would always evaluate to `Value - 1`

UI/SidePanel.cpp
```diff
@@ -957,8 +955,7 @@ void SidePanel::PlanetPanel::CompleteConstruction() {
     // focus-selection droplist
     m_focus_drop = GG::Wnd::Create<CUIDropDownList>(6);
     AttachChild(m_focus_drop);
-    m_focus_drop->DropDownOpenedSignal.connect(
-        boost::bind(&SidePanel::PlanetPanel::FocusDropListOpened, this, _1));
+        [this](bool open){ FocusDropListOpened(open); });
```
dbenage-cx: guessing errant line added from rebase

dbenage-cx (commented):
Seem to be getting slightly more output than target output when max is steady. e.g. Target Industry 10.0 but Industry Output is 10.2.

Planet shields/defense do not drop below 1.0 when attacked, and do not increase from 0.0 on outposts.

Troops continue to increase on a blockaded/bombarded planet.

geoffthemedio:
The shields aren't supposed to increase if at zero, based on the description of planetary barrier shield... Probably that needs to change, though.

Dilvish-fo:
> The shields aren't supposed to increase if at zero, based on the description of planetary barrier shield... Probably that needs to change, though.

I'm not sure if you are saying the tech should change or the description.  The description is more specific than just being at zero, but I suppose it could still be a bit more clear:
```
Increases max Shields on planets by 30, cumulative with Force-Field Harmonics, and regenerates each
turn by one plus the greater of (i) 25% of the value of the planets' Infrastructure or (ii) 1 (so long as the
shields are not already knocked to zero). 
```
The "(so long as the are not already knocked to zero)" only applies to that last part of the clause, the fixed value 1, and the reference to "knocked to zero" is meant to clarify the context of them being knocked down to zero during combat or whatnot.  Once either the infrastructure has climbed back above zero, or the regular (combat gated) shield regen brings shields back above zero, then the additional portion from planetary barrier shield also applies.  This was to fix a bug with the earliest implementation where the regen would apply even during combat, preventing the planet from ever being invadable.

Dilvish-fo:
I suppose that just swapping the order of clauses (i) and (ii) of that description would eliminate the ambiguity...

dbenage-cx:
What I mean is that on outposts (with no hostiles in system), current meter of shield/defense is fixed at 0.0 regardless of the target meter value.  If that is not an intended change, maybe move the base meter gain effects from species to a starting tech (or other suitable place).

default/scripting/species/common/general.macros
geoffthemedio: The original idea of the focus change penalty was that it would start out large and then decrease in magnitude over several turns. For now just `(0 == Source.TurnsSinceFocusChange)` is probably OK.

geoffthemedio:
> Seem to be getting slightly more output than target output when max is steady.
> e.g. Target Industry 10.0 but Industry Output is 10.2.

I think this is actually a problem with the target meter value being a turn out of date. When the values are actually stable, I get consistent numbers.

> move the [shield/defense ]base meter gain effects from species to a starting tech (or other suitable place).

Done for shields and defense. Natives also get it separately.

universe/ValueRef.cpp
```diff
@@ -707,11 +710,10 @@ double Variable<double>::Eval(const ScriptingContext& context) const
...
     if (object && meter_type != INVALID_METER_TYPE) {
         if (object->GetMeter(meter_type))
-            return object->InitialMeterValue(meter_type);
-
-    } else if (property_name == "TradeStockpile") {
-        if (const Empire* empire = GetEmpire(object->Owner()))
-            return empire->ResourceStockpile(RE_TRADE);
+            if (m_return_immediate_value)
+                return object->CurrentMeterValue(meter_type);
+            else
```
dbenage-cx: dangling else

default/scripting/species/common/troops.macros
```diff
@@ -39,10 +39,19 @@ EffectsGroup      // gives human bonuses when AI Aggression set to Beginner
             accountinglabel = "DIFFICULTY"
             priority = [[DEFAULT_PRIORITY]]
             effects = SetMaxTroops value = max(6, Value * 2)
+
+        EffectsGroup // increase 1 per turn, up to max
+            scope = Source
+            activation = And [
+                Planet
+                (Source.LastTurnConquered < CurrentTurn)
```
dbenage-cx, line 47: LastTurnAttackedByShip?
Probably no gain for outposts (did not verify).

default/scripting/species/common/supply.macros
```diff
@@ -35,4 +41,17 @@ ULTIMATE_SUPPLY
...
+
+STANDARD_SUPPLY_GROWTH
+'''        EffectsGroup // increase 1 per turn, up to max
+            scope = Source
+            activation = And [
+                Planet
+                (Source.LastTurnConquered < CurrentTurn)
+            ]
+            priority = [[AFTER_ALL_TARGET_MAX_METERS_PRIORITY]]
+            effects = SetSupply value = min(Value(Target.MaxSupply), Value + 1)
```
dbenage-cx: No gain for outposts, could move to techs/ship_hulls/SHP_GAL_EXPLO.focs.txt

default/scripting/species/common/planet_shields.macros
```diff
@@ -9,12 +9,24 @@ NATIVE_PLANETARY_SHIELDS
...
 
+STANDARD_SHIELD_GROWTH
+'''        EffectsGroup // increase 1 per turn, up to max
+            scope = Source
+            activation = And [
+                Planet
+                (Source.LastTurnConquered < CurrentTurn)
+            ]
+            priority = [[AFTER_ALL_TARGET_MAX_METERS_PRIORITY]]
+            effects = SetShield value = min(Value(Target.MaxShield), Value + 1)
```
dbenage-cx: Not gated by combat, keeping populated planets at 1.0 shields.
When not attacked, populated planets gain from both this and the effect added to defense.
Might consider moving the macros with multiple entry points to a file in scripting/common, with stacking groups.

universe/ValueRef.cpp
geoffthemedio: fixed

default/scripting/species/common/troops.macros
geoffthemedio: added the condition. not sure what you mean about outposts ... should be or shouldn't be?

default/scripting/species/common/supply.macros
geoffthemedio: You mean there shouldn't be, or should be an effect for outposts?

default/scripting/species/common/troops.macros
dbenage-cx: There is currently no effect to increase troop current meter for outposts and there should be one.

Same for supply, could do so for all relevant meters, I suspect stealth and stockpile are likely affected.
I can look into this late tonight if you would rather.

geoffthemedio: Stealth does not need an effect to increase it turn by turn. It just has a single meter not a separate max/target meter.

geoffthemedio: Added some additional effects for outpost troops and supply.

default/scripting/techs/construction/ORBITAL_CON.focs.txt
```diff
@@ -6,13 +6,25 @@ Tech
...
-    effectsgroups =
+    effectsgroups = [
         EffectsGroup
             scope = And [
                 Planet
                 OwnedBy empire = Source.Owner
             ]
             effects = SetMaxSupply value = Value + 1
+            
+        EffectsGroup    // outpost supply increases 1 per turn up to max
```
dbenage-cx: Tech is not a starting unlock, so outposts still will not grow towards max supply until this tech is researched.

I've not noticed if supply meter growth supposed to be affected by combat. This effectsgroup has no check, but STANDARD_SUPPLY_GROWTH does.

geoffthemedio: I don't think that's a problem, as outposts have 0 max supply before this tech anyway.

geoffthemedio:
I think the turn out of date meter values are mostly resolved now.

default/scripting/techs/construction/ORBITAL_CON.focs.txt
dbenage-cx: Tiny/Small planets?

dbenage-cx:
Posting as comment since some of the changed files are not showing up on html review for me.

1) Few parse errors from include resolution in species definitions, add as absolute includes to the macro file.
```diff
diff --git a/default/scripting/species/common/population.macros b/default/scripting/species/common/population.macros
index 75aa056e1..7dc6b241a 100644
--- a/default/scripting/species/common/population.macros
+++ b/default/scripting/species/common/population.macros
@@ -240,3 +240,7 @@ LIFECYCLE_MANIP_POPULATION_EFFECTS
                 SetPopulation value = [[MIN_RECOLONIZING_SIZE]]
             ]
 '''
+
+#include "/scripting/species/common/advanced_focus.macros"
+#include "/scripting/species/common/focus.macros"
+#include "/scripting/species/common/general.macros"
```

2) `SetSpecies name = ThisSpecies` is passing building name:
```
[error] client : Universe.cpp:1217 : GetEffectsAndTargets couldn't get Species BLD_COL_REPLICON
[error] client : Planet.cpp:242 : Planet::EnvironmentForSpecies couldn't get species with name ""
[error] client : Planet.cpp:242 : Planet::EnvironmentForSpecies couldn't get species with name ""
```

An option could be to pass the species name:
```diff
diff --git a/default/scripting/buildings/colonies/col_bld_gen.py b/default/scripting/buildings/colonies/col_bld_gen.py
index fab220ef2..c5857f5d5 100644
--- a/default/scripting/buildings/colonies/col_bld_gen.py
+++ b/default/scripting/buildings/colonies/col_bld_gen.py
@@ -106,8 +106,8 @@ BuildingType
         ${species_condition}
     ]
     effectsgroups = [
-        [[LIFECYCLE_MANIP_POPULATION_EFFECTS]]
-            
+        [[LIFECYCLE_MANIP_POPULATION_EFFECTS("${id}")]]
+
         EffectsGroup
             scope = And [
                 Object id = Source.PlanetID
@@ -125,7 +125,7 @@ BuildingType
                     ]
                     empire = Source.Owner
             ]
-            
+
         EffectsGroup
             scope = Source
             activation = Turn low = Source.CreationTurn + 2
diff --git a/default/scripting/species/common/population.macros b/default/scripting/species/common/population.macros
index 75aa056e1..c734587af 100644
--- a/default/scripting/species/common/population.macros
+++ b/default/scripting/species/common/population.macros
@@ -210,6 +210,7 @@ PHOTOTROPHIC_BONUS
             effects = SetTargetPopulation value = Value - 10 * Source.SizeAsDouble
 '''
 
+// @1@ Species key
 LIFECYCLE_MANIP_POPULATION_EFFECTS
 '''
         EffectsGroup
@@ -222,7 +223,7 @@ LIFECYCLE_MANIP_POPULATION_EFFECTS
                 Turn low = Source.CreationTurn + 1 high = Source.CreationTurn + 1
             ]
             effects = [
-                SetSpecies name = ThisSpecies
+                SetSpecies name = @1@
                 SetPopulation value = 1
             ]
             
@@ -236,7 +237,7 @@ LIFECYCLE_MANIP_POPULATION_EFFECTS
                 Turn low = Source.CreationTurn + 1 high = Source.CreationTurn + 1
             ]
             effects = [
-                SetSpecies name = ThisSpecies
+                SetSpecies name = @1@
                 SetPopulation value = [[MIN_RECOLONIZING_SIZE]]
             ]
 '''
```

3) Similar error for `CanColonize`:
`[error] client : Condition.cpp:8516 : CanColonize couldn't get species: BLD_COL_REPLICON`
This one I believe to be from the scope condition in SHP_GAL_EXPLO.focs.txt, possible fix:
```diff
diff --git a/default/scripting/techs/ship_hulls/SHP_GAL_EXPLO.focs.txt b/default/scripting/techs/ship_hulls/SHP_GAL_EXPLO.focs.txt
index 5c58cc138..fee90e7c1 100644
--- a/default/scripting/techs/ship_hulls/SHP_GAL_EXPLO.focs.txt
+++ b/default/scripting/techs/ship_hulls/SHP_GAL_EXPLO.focs.txt
@@ -59,7 +59,6 @@ Tech
             scope = And [
                 Planet
                 OwnedBy empire = Source.Owner
-                CanColonize
                 
                 (LocalCandidate.Population              <  LocalCandidate.TargetPopulation)
                 (LocalCandidate.Population +
@@ -73,7 +72,10 @@ Tech
                     (0.1 <= LocalCandidate.Happiness    <= ([[MIN_RECOLONIZING_HAPPINESS]] - 0.001) )
                 ]
             ]
-            activation = Source
+            activation = And [
+                Species
+                CanColonize
+            ]
             effects = [
                 GenerateSitRepMessage
                     message = "SITREP_POP_THRESHOLD"
```

geoffthemedio:
1 - The `#include "common/*.macros"` line in most / all species definitions should already be including those files, but don't seem to, for unknown reasons.

3 - was a consequence of the preceeding, which was setting the species of a planet to the name of a building type, and is not something that requires separate changes to fix.

Issues should be resolved.

dbenage-cx:
Troops still have base regen at a planet when bombarded.
This makes invasion frustrating as you need to calculate how many troops will be available on the turn you expect to invade.
e.g. On a bombarded planet with 3.0 current troops (max 4.0+), clicking invade with 4.0 troops will result in a tie.

There is some strangeness with ship meter icons sometimes showing 5.0 more than listed in accounting details/tooltip.
e.g. Ships with 8xdeath rays @ 30.0 dmg might show 245 next to direct weapon icon, but list with a total of 240 in tooltip (with no changes occurring next/previous turn, also noticed wrt structure).

geoffthemedio:
Can you be more specific than "sometimes"? I don't see anything unusual with 8x Death Rays.
8xdeathray

dbenage-cx:
Loading some previous games, not always +5.0 and could not find an instance from my previous description of 8xdeath rays :/
A seemingly consistent occurrence on predesign is the structure for Large Snowflake (+reinforced hull), displaying 130 on icon, 125 on tooltip.

(In first screenshot, the tooltip for damage aligns with icon for 100 damage)

fo_1941_a
fo_1941_b
fo_1941_c

geoffthemedio:
I suspect it's a rounding issue when just 2 digits are requested but it overrides to display 3. ie. it's rounding to the nearest 10 then displaying to the 1's digit.

geoffthemedio:
Made the minimum digits for DoubleToString 3 to hopfully avoid the rounding issue.

Switched to using initial instead of current troops for display and calculation of ground combat.

default/scripting/common/priorities.macros
```diff
@@ -15,5 +36,17 @@ LATE_PRIORITY
...
+FOCUS_CHANGE_PENALTY_PRIORITY
+'''1100'''
+
+
+METER_OVERRIDE_PRIORITY
+'''2000'''
+
+
+END_CLEANUP_PRIORITY
+'''5000'''
```
dbenage-cx: EOL

default/scripting/buildings/GATEWAY_VOID.focs.txt
```diff
@@ -27,19 +27,20 @@ BuildingType
                     ]
                     empire = Source.Owner
             ]
+
```
dbenage-cx, line 30: Nitpick: many of the added "blank" lines in ~.focs.txt~ FOCS files contain excess whitespace

default/scripting/species/common/happiness.macros
```diff
@@ -1,5 +1,17 @@
 AVERAGE_HAPPINESS
 '''
+        EffectsGroup    // increase or decrease 1 per turn towards target
+            scope = Source
+            activation = And [
+                Planet
+                (Source.Population != Source.TargetPopulation)
```
dbenage-cx: `Source.Happiness != Source.TargetHappiness`

default/scripting/species/common/population.macros
```diff
@@ -191,7 +206,42 @@ PHOTOTROPHIC_BONUS
...
+            ]
+            effects = [
+                SetSpecies name = @1@
+                SetPopulation value = [[MIN_RECOLONIZING_SIZE]]
+            ]
+'''
+
+#include "/scripting/species/common/advanced_focus.macros"
+#include "/scripting/species/common/focus.macros"
+#include "/scripting/species/common/general.macros"
```
dbenage-cx: EOL

default/scripting/species/common/stealth.macros
```diff
@@ -57,3 +57,5 @@ ULTIMATE_STEALTH
             stackinggroup = "SPECIES_STEALTH_STACK"
             effects = SetStealth value = Value + [[HIGH_STEALTH]]
 '''
+
+#include "/scripting/common/stealth.macros"
```
dbenage-cx: EOL

util/i18n.cpp
```diff
@@ -209,14 +209,58 @@ namespace {
...
+        rounding_factor = pow(10.0, static_cast<double>(pow10 - digits + 1));
+        mag = mag / rounding_factor;
+        mag = round(mag);
+        mag *= rounding_factor;
+        //std::cout << "magnitude after second rounding to " << digits << " digits: " << mag << std::endl;
+
+        return mag;
+    }
+
+    int TestRounding() {
```
dbenage-cx, line 247: Unused.

dbenage-cx (commented):
Otherwise LGTM

util/i18n.cpp
geoffthemedio: used, just not currently active.

default/scripting/common/priorities.macros
geoffthemedio: not sure why / if having endlines matters...

dbenage-cx: Some of the parsing has issue with not reading/matching last line in file when it does not end in newline.

Dilvish-fo:
> Made the minimum digits for DoubleToString 3 to hopfully avoid the rounding issue.

@geoffthemedio I am finding the display like this to be really cluttered with extra zeroes, noticeably less legible.  I think we need to find a way to handle the rounding issue better.

geoffthemedio:
> I think we need to find a way to handle the rounding issue better.

Should be fixed in master.

PR #1943, opened 2018-01-12
Add minimum AI, maximum human player options (mp)

dbenage-cx:
Compliments the existing max AI/min human options.

Renames existing options consistent with recent option changes.


geoffthemedio:
haven't tested but looks ok

PR #1945, opened 2018-01-15
Clear human client pending save game counter

### LINKED ISSUE #1944: Saving fails after resigning
### Environment
* **FreeOrion Version**: 0.4.7+ 0e7f977
* **Operating System**: Fedora 26

### Description
Both manual and autosaves are no longer processed when (sometimes) resigning and starting a new game.

For autosaves: no message relating to saves appears on console, nor do save files get created.
For manual saves: Dialog appears and exits as normal, but no further messages or files are generated.

Likely first reported in [this forum thread](http://freeorion.org/forum/viewtopic.php?f=28&t=10857) (suspect this is same issue).
[Logs (only first AI of 20)](https://github.com/freeorion/freeorion/files/1630058/freeorion.zip)

### Expected Result
Game saves as normal

### Steps to reproduce
Likely wish to configure game so that the saving takes a longer time, for me generally 20+ AI is enough, not 100% reproducable:
* Start a new game with new galaxy auto saves enabled
* Resign the game after save starts
* Exit the pending save dialog
* Start a new game, issue reproduced if saving did not initiate.

dbenage-cx:
Not certain this catches all cases (and we may wish to handle the remaining queue differently).

Fixes #1944

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -429,6 +429,13 @@ namespace {
             std::runtime_error("LOCAL_SERVER_ALREADY_RUNNING_ERROR")
         {}
     };
+
+    void ClearPreviousPendingSaves(std::queue<std::string>& pending_saves) {
+        if (pending_saves.empty())
+            return;
+        WarnLogger() << "Clearing " << std::to_string(pending_saves.size()) << " pending save game request(s)";
+        std::queue<std::string>().swap(pending_saves);
+    }
 }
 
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Do automatic saves when resigning / quitting still resolve normally with this?

dbenage-cx:
I just re-verified with only save.auto.exit.enabled checked, the game is correctly saved on resign or exit unless the player cancels the dialog.

PR #1946, opened 2018-01-15
Allow random as setup seed option

dbenage-cx:
Supports setting setup.seed config option to RANDOM, to generate different seeds on new game.

### DIFF
--- a/UI/GalaxySetupWnd.cpp
+++ b/UI/GalaxySetupWnd.cpp
@@ -764,18 +764,8 @@ void GalaxySetupPanel::CompleteConstruction() {
     SettingsChangedSignal();
 }
 
-namespace {
-    // set of characters from which to generate random seed that excludes some ambiguous letter/number pairs
-    static char alphanum[] = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
-}
-
 void GalaxySetupPanel::RandomClicked() {
-    std::string s;
-    ClockSeed(); // to ensure we don't always get the same sequence of seeds
-    for (int i = 0; i < 8; ++i)
-        s += alphanum[ RandSmallInt(0, (sizeof(alphanum) - 2))];
-    m_seed_edit->SetText(s);
-    //std::cout << "GalaxySetupPanel::RandomClicked() new seed: " << s << std::endl;
+    m_seed_edit->SetText("");
     SettingChanged();
 }
 
@@ -929,7 +919,7 @@ void GalaxySetupPanel::SetFromSetupData(const GalaxySetupData& setup_data) {
 }
 
 void GalaxySetupPanel::GetSetupData(GalaxySetupData& setup_data) const {
-    setup_data.m_seed =             GetSeed();
+    setup_data.SetSeed(GetSeed());
     setup_data.m_size =             Systems();
     setup_data.m_shape =            GetShape();
     setup_data.m_age =              GetAge();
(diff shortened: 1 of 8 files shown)


UI/GalaxySetupWnd.cpp
```diff
@@ -764,18 +764,8 @@ void GalaxySetupPanel::CompleteConstruction() {
...
-}
-
 void GalaxySetupPanel::RandomClicked() {
-    std::string s;
-    ClockSeed(); // to ensure we don't always get the same sequence of seeds
-    for (int i = 0; i < 8; ++i)
-        s += alphanum[ RandSmallInt(0, (sizeof(alphanum) - 2))];
-    m_seed_edit->SetText(s);
-    //std::cout << "GalaxySetupPanel::RandomClicked() new seed: " << s << std::endl;
+    m_seed_edit->SetText("RANDOM");
```
geoffthemedio: doesn't work in languages other than english

dbenage-cx: TY, addressed by storing it separate.
Note an actual seed is always displayed in mp lobby.  I did not address this so players can compare if they suspect different seeds (or want to argue about why one char should change ;) ).

Dilvish-fo:
So, is this essentially a more explicit equivalent to leaving the seed empty (which I would agree is by itself a useful improvement), or is there some other difference as well?

dbenage-cx:
There are a few other differences:
The current behavior when clicking random on single player setup is to determine a seed on each click, the determined seed is the value saved to setup.seed
This changes to setting config value of setup.seed to RANDOM when clicking random, so each new game is a different seed.
That determination is made after clicking ok on single player setup.
In multiplayer lobby the seed is determined and sent to clients (resetting text to the new seed).

The seed is now explicitly determined before universe generation.
One effect of this is that the currently used seed is now shown in Pedia->[Game - Galaxy] and save previews. An empty seed previously showed an empty value.
I was not aware of the behavior of an empty setup.seed value prior to this, so had not extensively tested such.
It is possible a different seed was used at different points when set to an empty value.

server/ServerApp.cpp
```diff
@@ -1380,7 +1380,7 @@ void ServerApp::GenerateUniverse(std::map<int, PlayerSetupData>& player_setup_da
             seed = static_cast<unsigned int>(h);
         } catch (...) {}
     }
-    if (GetGalaxySetupData().m_seed.empty()) {
+    if (GetGalaxySetupData().GetSeed().empty() || GetGalaxySetupData().GetSeed() == "RANDOM") {
```
geoffthemedio, line 1383: What happens if you click the button to randomly generate a seed, then put the cursor in the text box to type a seed and press enter. Presumably it will then set the seed to whatever the localized version of "Random" is, and not actually generate a random seed.

dbenage-cx: Good point, I've changed to disable the edit control when set to random, and convert an empty seed value to the random toggle.

I mildly prefer the dice graphic so left it unchanged, would you prefer a checkbox or an alternate rendering when toggled?

dbenage-cx:
Rebased to master (5266024)

server/ServerApp.cpp
geoffthemedio: Different graphic is OK as long as it's clear when it's enabled or disabled... (eg. not like the tech status display toggles)

dbenage-cx: Changing to checkbox does not look appropriate and not sure what the usability issue is with tech controls, looks best to leave it for now.

geoffthemedio: The filters for techs look almost or exactly the same when on or off. Only when moused over do they look different.

geoffthemedio:
haven't tested, but no further comments

Vezzra:

PR #1947, opened 2018-01-17
Customization authorization roles.

o01eg:
This PR allows python script to return authorization roles for players.

- `AuthProvider.is_require_auth_or_return_roles` returns list of `freeoion.roleType` with default roles for anonymous login or `True` if player name requires authorization.
- `AuthProvider.is_success_auth_and_return_roles` returns list of roles for authenticated player or `False` if authentication failed.

Example `auth.py` script loads roles information from `auth.txt` files. First line is a list of default roles for unauthenticated users, next lines are in format `<login>:<roles characters>:<password>`

Connect windows allows to choose client type. It useful if user doesn't have a player role.
fo


server/ServerFSM.cpp
```diff
@@ -810,7 +827,9 @@ sc::result MPLobby::react(const AuthResponse& msg) {
     std::string auth;
     ExtractAuthResponseMessageData(message, player_name, auth);
 
-    if (!server.IsAuthSuccess(player_name, auth)) {
+    Networking::AuthRoles roles;
+
+    if (!server.IsAuthSuccess(player_name, auth, roles)) {
         // wrong password
         player_connection->SendMessage(ErrorMessage(str(FlexibleFormat(UserString("ERROR_WRONG_PASSWORD")) % player_name),
```
geoffthemedio: No distinction in the error message between wrong password and right password with a disallowed role?

o01eg: Roles tests in `MPLobby::EstablishPlayer` and returns `ERROR_CLIENT_TYPE_NOT_ALLOWED`. `IsAuthSuccess` only tests password and returns roles without testing them.

geoffthemedio: Don't understand what you mean. `IsAuthSuccess` returns a boolean, not "roles".

geoffthemedio: Should probably return roles rather than fill the passed-in parameter... would be clearer.

geoffthemedio: Need to update the function name then... or have two functions, one of which returns a connection's roles.

o01eg: It is a wrapper for `AuthProvider.is_auth_success`. So it better to rename them.

o01eg: Renamed them.

o01eg:
Some fixes in authorization.
Allow to choose role in connection window.

UI/ServerConnectWnd.cpp
```diff
@@ -183,13 +226,20 @@ void ServerConnectWnd::OkClicked() {
     GetOptionsDB().Set("setup.multiplayer.host.address", m_IP_address_edit->Text());
     GetOptionsDB().Commit();
 
-    m_result.first = *m_player_name_edit;
+    m_result.player_name = *m_player_name_edit;
+    const auto& it = m_client_type_list->CurrentItem();
```
geoffthemedio: Normally store itetators by vale, I think...

UI/ServerConnectWnd.h
```diff
@@ -6,12 +6,19 @@
...
+#include "../network/Networking.h"
 
 /** server connections window */
 class ServerConnectWnd : public CUIWnd
 {
 public:
+    /** Connection parameters */
+    struct Result {
+        std::string player_name;
+        std::string server_name;
```
geoffthemedio: Is it always a name or sometimes intead an IP address?

o01eg: It's named after variable `server_name` in `HumanClientApp::MultiPlayerGame`. Should them be renamed?

UI/ServerConnectWnd.cpp
o01eg: Fixed.

client/human/HumanClientApp.cpp
```diff
@@ -653,16 +652,16 @@ void HumanClientApp::MultiPlayerGame() {
     m_connected = m_networking->ConnectToServer(server_name);
     if (!m_connected) {
         ClientUI::MessageBox(UserString("ERR_CONNECT_TIMED_OUT"), true);
-        if (server_connect_wnd->Result().second == "HOST GAME SELECTED")
+        if (server_connect_wnd->GetResult().server_name == "HOST GAME SELECTED")
```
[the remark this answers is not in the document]
o01eg: May be something with "destination"?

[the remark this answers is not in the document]
o01eg: Renamed to `server_dst`.

client/human/HumanClientApp.cpp
```diff
@@ -626,12 +625,12 @@ void HumanClientApp::MultiPlayerGame() {
     auto server_connect_wnd = GG::Wnd::Create<ServerConnectWnd>();
     server_connect_wnd->Run();
 
-    std::string server_name = server_connect_wnd->Result().second;
+    std::string server_dst = server_connect_wnd->GetResult().server_dst;
```
geoffthemedio: I don't know what `dst` refers to.

o01eg: Server destination

UI/ServerConnectWnd.cpp
```diff
@@ -183,13 +226,20 @@ void ServerConnectWnd::OkClicked() {
...
+    }
     if (m_host_or_join_radio_group->CheckedButton() == 0) {
-        m_result.second = "HOST GAME SELECTED";
+        m_result.server_dst = "HOST GAME SELECTED";
     } else {
-        m_result.second = *m_IP_address_edit;
-        if (m_result.second.empty() && !(***m_servers_lb->Selections().begin()).empty()) {
-            m_result.second = boost::polymorphic_downcast<GG::Label*>(
+        m_result.server_dst = *m_IP_address_edit;
+        if (m_result.server_dst.empty() && !(***m_servers_lb->Selections().begin()).empty()) {
```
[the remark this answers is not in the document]
o01eg: I wonder. I've never been have a empty `server_name` to check.
I suppose `m_servers_lb->Selections().begin()` is a `std::unordered_set<ListBox::iterator>::iterator` which derefs to `std::list<std::shared_ptr<Row>>::iterator` derefs to `std::shared_ptr<Row>` and next to `Row`. 
It should not be safe for empty selection but is looks like button "Ok" is blocked in this case.

client/human/HumanClientApp.cpp
geoffthemedio: `dest` would be better than `dst`, which looks like an initialism.

o01eg: Changed.

geoffthemedio:
Need an "as" before the what to join as droplist.
Joining should be as a "Player" not as a "Human".
Why can one only pick a role when joining, not when hosting?

o01eg:
> Need an "as" before the what to join as droplist.

Fixed.

> Joining should be as a "Player" not as a "Human".

Strings id `HUMAN_PLAYER`, `OBSERVER` and `MODERATOR` are also used in multiplayer lobby and game rules script. Should it be changed everywhere to `PLAYER`, `OBSERVER`, `MODERATOR`?

> Why can one only pick a role when joining, not when hosting?

Host never tests and gets all roles so he always can change client type in lobby.

geoffthemedio:
> Strings id HUMAN_PLAYER, OBSERVER and MODERATOR are also used in multiplayer lobby and game rules script. Should it be changed everywhere to PLAYER, OBSERVER, MODERATOR?

Perhaps. In that context it seems clearer what is meant, as "HUMAN" is contrasted mainly with "AI". But in the case of the what-to-join-as droplist, all the options are for the human who's joining, and one can't join as an AI, so it's redundant. Maybe "Human Player" in both cases? Though that will take up space and possibly not fit in the droplist on in the multiplayer lobby list.

> Host never tests and gets all roles so he always can change client type in lobby.

That's fine, but it seems odd as a user that one can pick a role before joining but not before hosting.

o01eg:
Fixed conflicts. Use Player instead of Human in both server connection window and multiplayer lobby window.

> In that context it seems clearer what is meant, as "HUMAN" is contrasted mainly with "AI". But in the case of the what-to-join-as droplist, all the options are for the human who's joining, and one can't join as an AI, so it's redundant. Maybe "Human Player" in both cases? Though that will take up space and possibly not fit in the droplist on in the multiplayer lobby list.

Even in multiplayer lobby "AI" and "Human"/"Player" never appears in the one dropdown list as each row either human or AI.

geoffthemedio:
default should be player, not moderator.

o01eg:
Fixed. I missed sorting enabled by default.

PR #1948, opened 2018-01-17
Hostless autosaves

o01eg:
Adds autosaves to hostless mode and options to enable/disable it.

Fix issues from http://freeorion.org/forum/viewtopic.php?f=25&t=10830

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1795,6 +1795,9 @@ Use Binary serialization for saving games. Binary serialization is faster to sav
 OPTIONS_DB_XML_ZLIB_SERIALIZATION
 When saving games with XML serialization, compress most of the XML before writing the file. Compression substantially reduces save file sizes, but may make saves unloadable due to memory requirements to decompress the save data.
 
+OPTIONS_DB_HOSTLESS_SAVE
+Enable autosave in hostless mode.
+
 OPTIONS_DB_UI_WINDOWS_EXISTS
 True if the window with a given config name currently exists, false if one was created and deleted, doesn't exist if no window has yet been created with that name.
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1741,6 +1741,9 @@ Enable autosave after the galaxy is created before any game play.
 OPTIONS_DB_AUTOSAVE_GAME_CLOSE
 Enable autosave when resigning or closing the game.
 
+OPTIONS_DB_AUTOSAVE_HOSTLESS
+Enable autosave in hostless mode.
+
 OPTIONS_DB_UI_MOUSE_LR_SWAP
 Swaps results of clicking left and right mouse buttons.
 
@@ -1794,9 +1797,6 @@ Use Binary serialization for saving games. Binary serialization is faster to sav
 
 OPTIONS_DB_XML_ZLIB_SERIALIZATION
 When saving games with XML serialization, compress most of the XML before writing the file. Compression substantially reduces save file sizes, but may make saves unloadable due to memory requirements to decompress the save data.
-
-OPTIONS_DB_HOSTLESS_SAVE
-Enable autosave in hostless mode.
 
 OPTIONS_DB_UI_WINDOWS_EXISTS
 True if the window with a given config name currently exists, false if one was created and deleted, doesn't exist if no window has yet been created with that name.
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -2325,7 +2325,7 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
     }
 
     if (server.IsHostless() && GetOptionsDB().Get<bool>("save.auto.hostless.enabled")) {
-        boost::filesystem::path autosave_dir_path = GetServerSaveDir();
+        boost::filesystem::path autosave_dir_path = GetServerSaveDir() / "auto";
         const auto& extension = MP_SAVE_FILE_EXTENSION;
         // Add timestamp to autosave generated files
         std::string datetime_str = FilenameTimestamp();
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -2166,6 +2166,24 @@ sc::result WaitingForTurnEndIdle::react(const SaveGameRequest& msg) {
     return transit<WaitingForSaveData>();
 }
 
+sc::result WaitingForTurnEndIdle::react(const HostlessSave& e) {
+    TraceLogger(FSM) << "(ServerFSM) WaitingForTurnEndIdle.HostlessSave";
+
+    ServerApp& server = Server();
+
+    boost::filesystem::path autosave_dir_path = GetServerSaveDir() / "auto";
+    const auto& extension = MP_SAVE_FILE_EXTENSION;
+    // Add timestamp to autosave generated files
+    std::string datetime_str = FilenameTimestamp();
+
+    std::string save_filename = boost::io::str(boost::format("FreeOrion_%04d_%s%s") % server.CurrentTurn() % datetime_str % extension);
+
+    boost::filesystem::path save_path(autosave_dir_path / save_filename);
+
+    context<WaitingForTurnEnd>().m_save_filename = save_path.string();
+    return transit<WaitingForSaveData>();
+}
+
 
 ////////////////////////////////////////////////////////////
 // WaitingForSaveData
@@ -2324,6 +2342,7 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
         }
     }
 
+    post_event(HostlessSave());
     return transit<WaitingForTurnEnd>();
 }
 
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -2324,8 +2324,8 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
         }
     }
 
-    if (server.IsHostless() && GetOptionsDB().Get<bool>("save.server.hostless")) {
-        boost::filesystem::path autosave_dir_path = GetServerSaveDir() / "auto";
+    if (server.IsHostless() && GetOptionsDB().Get<bool>("save.auto.hostless.enabled")) {
+        boost::filesystem::path autosave_dir_path = GetServerSaveDir();
         const auto& extension = MP_SAVE_FILE_EXTENSION;
         // Add timestamp to autosave generated files
         std::string datetime_str = FilenameTimestamp();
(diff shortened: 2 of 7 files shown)


server/ServerFSM.cpp
```diff
@@ -2324,6 +2342,9 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
         }
     }
 
+    if(server.IsHostless() && GetOptionsDB().Get<bool>("save.server.hostless")) {
```
geoffthemedio: space between `if` and `(`.

server/ServerFSM.cpp
```diff
@@ -2166,6 +2166,24 @@ sc::result WaitingForTurnEndIdle::react(const SaveGameRequest& msg) {
     return transit<WaitingForSaveData>();
 }
 
+sc::result WaitingForTurnEndIdle::react(const HostlessSave& e) {
```
geoffthemedio: Is a separate event needed for this? Could you just check in `react(const SaveGameRequest& msg)` if the server is in hostless mode, and if so, don't check for a sender connection or if its player is the host?

o01eg: I've tried but `SaveGameRequest` requires `PlayerConnectionPtr& player_connection` which cann't be `nullptr`.

o01eg: Yes. I suppose if `SaveGameRequest` doesn't have a connection then it was send in hostless mode.

util/MultiplayerCommon.cpp
```diff
@@ -33,6 +33,7 @@ namespace {
         db.Add<std::string>("resource.stringtable.path",    UserStringNop("OPTIONS_DB_STRINGTABLE_FILENAME"),   "");
         db.Add("save.format.binary.enabled",                UserStringNop("OPTIONS_DB_BINARY_SERIALIZATION"),   false);
         db.Add("save.format.xml.zlib.enabled",              UserStringNop("OPTIONS_DB_XML_ZLIB_SERIALIZATION"), true);
+        db.Add("save.server.hostless",                      UserStringNop("OPTIONS_DB_HOSTLESS_SAVE"),          false);
```
dbenage-cx: Perhaps save.auto.hostless.enabled, grouping with other autosave options.

o01eg: Fixed

server/ServerFSM.cpp
```diff
@@ -2324,8 +2324,8 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
         }
     }
 
-    if (server.IsHostless() && GetOptionsDB().Get<bool>("save.server.hostless")) {
-        boost::filesystem::path autosave_dir_path = GetServerSaveDir() / "auto";
```
geoffthemedio: why do you say the auto save subdirectory is unused...?

geoffthemedio: It looks like it checks that the save path is in the save directory, not that it is the save directory.

o01eg: It also checks if `auto` directory exists but don't create it.

geoffthemedio: ok... so create it then? autosaves are supposed to be segregated into the auto directory

o01eg: Ok, I'll look into it.

o01eg: Looks like `boost::filesystem::canonical` in `IsInDir` requires directories to exist. There is a simple solution with `boost::filesystem::weakly_canonical` but it was introduced in boost 1.60

geoffthemedio: > Looks like boost::filesystem::canonical in IsInDir requires directories to exist.

What is the importance / consequence of this?

o01eg: > What is the importance / consequence of this?

It throws error if path not exists: http://www.boost.org/doc/libs/1_58_0/libs/filesystem/doc/reference.html#canonical
Yet I get a little  workaround around it.

server/ServerFSM.cpp
```diff
@@ -2324,6 +2324,20 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
         }
     }
 
+    if (server.IsHostless() && GetOptionsDB().Get<bool>("save.auto.hostless.enabled")) {
```
geoffthemedio, line 2327: I think autosaves are supposed to be done at the end of a turn, not at the start.

server/ServerFSM.cpp
```diff
@@ -2324,6 +2324,20 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
...
+        boost::filesystem::path autosave_dir_path = GetServerSaveDir() / "auto";
+        const auto& extension = MP_SAVE_FILE_EXTENSION;
+        // Add timestamp to autosave generated files
+        std::string datetime_str = FilenameTimestamp();
+
+        std::string save_filename = boost::io::str(boost::format("FreeOrion_%04d_%s%s") % server.CurrentTurn() % datetime_str % extension);
+
+        boost::filesystem::path save_path(autosave_dir_path / save_filename);
+
+        PlayerConnectionPtr dummy_connection = nullptr;
```
geoffthemedio, line 2337: is this temporary local necessary vs. just passing `nullptr` directly?

util/MultiplayerCommon.cpp
```diff
@@ -33,6 +33,7 @@ namespace {
         db.Add<std::string>("resource.stringtable.path",    UserStringNop("OPTIONS_DB_STRINGTABLE_FILENAME"),   "");
         db.Add("save.format.binary.enabled",                UserStringNop("OPTIONS_DB_BINARY_SERIALIZATION"),   false);
         db.Add("save.format.xml.zlib.enabled",              UserStringNop("OPTIONS_DB_XML_ZLIB_SERIALIZATION"), true);
+        db.Add("save.auto.hostless.enabled",                UserStringNop("OPTIONS_DB_HOSTLESS_SAVE"),          false);
```
geoffthemedio: why not true by default?

server/ServerFSM.cpp
o01eg: It's necessary as message constructor accepts non-const reference to connection.

util/MultiplayerCommon.cpp
o01eg: Made it enabled.

server/ServerFSM.cpp
geoffthemedio: How long does it expect that reference to be valid? Beyond the lifetime of `dummy_connection`?

o01eg: It's need to be valid until `MessageEventBase` constructor so it don't have to be valid outside its scope.

server/ServerFSM.cpp
o01eg: Doesn't turn end in the end of `ProcessingTurn::react(const ProcessTurn&)`?

geoffthemedio: I mean after all players have given orders, it should autosave, then process those orders and the between-turn events and updates. If the save is loaded, players should be able to issue more orders and re-end the turn. At least that's how it should work in single player.

o01eg: I'll try to change it but I wonder which correct state and event sequence to save a game now.

o01eg: I don't see how the game saved after getting all orders and before turn processing. We waits for orders in `TurnOrder` and raise `CheckTurnEndConditions` where the server goes to `ProcessingTurn` state when all orders are received.

geoffthemedio: Singleplayer autosaves are initiated by the player's client when the player clicks end turn, before the actual end turn message is sent, in `HumanClientApp::StartTurn`

server/ServerFSM.cpp
geoffthemedio: Should probably change `MessageEventBase` to take a `const PlayerConnectionPtr&` then... doesn't seem to need to be a non-const pointer (but does need to be a pointer-to-non-const).

server/ServerFSM.cpp
o01eg: The client sends `SAVE_GAME_INITIATE` which translates to `SaveGameRequest` event, but only `WaitingForTurnEndIdle` is a state processes `SaveGameRequest` where the server goes from `WaitingForSaveData` after game saved or from `ProcessingTurn` after `ProcessTurn` has been handled.

I don't see a way how the game can save something between acquiring orders and turn processing.

server/ServerFSM.cpp
o01eg: Should it be done in this Pull-Request?

server/ServerFSM.cpp
geoffthemedio: When the client sends the request, the server is still in `WaitingForTurnEndIdle`.

server/ServerFSM.cpp
geoffthemedio: I guess it's not necessary.

server/ServerFSM.cpp
o01eg: Yes, but it before all players sent their orders.

o01eg: Yes, but other players can send their orders after host.

default/stringtables/en.txt
```diff
@@ -1795,6 +1795,9 @@ Use Binary serialization for saving games. Binary serialization is faster to sav
 OPTIONS_DB_XML_ZLIB_SERIALIZATION
 When saving games with XML serialization, compress most of the XML before writing the file. Compression substantially reduces save file sizes, but may make saves unloadable due to memory requirements to decompress the save data.
 
+OPTIONS_DB_HOSTLESS_SAVE
```
geoffthemedio: should call this something with `AUTOSAVE` in the name, and put it with the other autosave strings.

server/ServerFSM.cpp
geoffthemedio: not in single player.

in multiplayer, I think the autosaves are done at the start of a turn. so I guess this timing is OK.

o01eg: Yes, after all orders have been received and processed there will be start of next turn where I push `SaveGameRequest` for hostless autosave.

default/stringtables/en.txt
o01eg: Renamed and moved.

geoffthemedio:
Does this respect the max autosaves setting?

https://github.com/freeorion/freeorion/blob/master/client/human/HumanClientApp.cpp#L1325

o01eg:
@geoffthemedio No, because it's a client's options. Should it also ported?

geoffthemedio:
Probably should also be an option that can be set on the server. Doesn't need to be in this pull request, though.

These changes work for you as intended?

o01eg:
Yes, they work.

PR #1950, opened 2018-01-18
Improve UpdateMeterEstimates() invalid id error message.

LGM-Doyle:
Refactor `UpdateMeterEstimates()` so that ~~large numbers of contained object
ids can't resize `objects_list` vector while being iterated and cause a
crash.~~ the error message includes the invalid id and its valid container, rather than the less helpful "Universe::UpdateMeterEstimates tried to get an invalid object..."

### DIFF
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -690,19 +690,18 @@ void Universe::UpdateMeterEstimates() {
     UpdateMeterEstimatesImpl(std::vector<int>());
 }
 
-void Universe::UpdateMeterEstimates(int object_id, bool is_update_contained_objects) {
+void Universe::UpdateMeterEstimates(int object_id, bool update_contained_objects) {
     // ids of the object and all valid contained objects
-    std::unordered_set<int> all_ids;
+    std::unordered_set<int> collected_ids;
 
-    // Collect objects to update meter for.  this may be a single object, a group of related
-    // objects, or all objects in the (known) universe.  also clear effect accounting for meters
-    // that are to be updated. Return true on success (all_ids is valid).
-    std::function<bool (int, int)> update_obj_and_contained_objs =
-        [this, &all_ids, is_update_contained_objects, &update_obj_and_contained_objs]
+    // Collect objects ids to update meter for.  This may be a single object, a
+    // group of related objects. Return true if all collected ids are valid.
+    std::function<bool (int, int)> collect_ids =
+        [this, &collected_ids, update_contained_objects, &collect_ids]
         (int cur_id, int container_id)
     {
         // Ignore if already in the set
-        if (all_ids.count(cur_id))
+        if (collected_ids.count(cur_id))
             return true;
 
         auto cur_object = m_objects.Object(cur_id);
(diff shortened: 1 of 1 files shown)


universe/Universe.cpp
```diff
@@ -686,42 +686,57 @@ void Universe::InitMeterEstimatesAndDiscrepancies() {
...
 
-void Universe::UpdateMeterEstimates(int object_id, bool update_contained_objects) {
-    if (object_id == INVALID_OBJECT_ID) {
+void Universe::UpdateMeterEstimates(int object_id, bool is_update_contained_objects) {
+
+    auto panic_update_all_estimates = [this](){
         for (int obj_id : m_objects.FindExistingObjectIDs())
             m_effect_accounting_map[obj_id].clear();
-        // update meters for all objects.  Value of updated_contained_objects is irrelivant and is ignored in this case.
+        // update meters for all objects.  Value of updated_contained_objects is irrelevant and is ignored in this case.
```
geoffthemedio: Comment refers to old parameter name

universe/Universe.cpp
```diff
@@ -686,42 +686,57 @@ void Universe::InitMeterEstimatesAndDiscrepancies() {
 void Universe::UpdateMeterEstimates()
 { UpdateMeterEstimates(INVALID_OBJECT_ID, false); }
 
-void Universe::UpdateMeterEstimates(int object_id, bool update_contained_objects) {
-    if (object_id == INVALID_OBJECT_ID) {
+void Universe::UpdateMeterEstimates(int object_id, bool is_update_contained_objects) {
+
+    auto panic_update_all_estimates = [this](){
```
geoffthemedio: Meaning of function name? Why "panic"?

universe/Universe.cpp
```diff
@@ -686,42 +686,57 @@ void Universe::InitMeterEstimatesAndDiscrepancies() {
...
     // collect objects to update meter for.  this may be a single object, a group of related objects, or all objects
     // in the (known) universe.  also clear effect accounting for meters that are to be updated.
-    std::set<int> objects_set;
-    std::list<int> objects_list;
-    objects_list.push_back(object_id);
+    std::function<void (int, std::unordered_set<int>&, int)> update_obj_and_contained_objs =
+        [this, &all_ids, is_update_contained_objects, &panic_update_all_estimates, &update_obj_and_contained_objs]
+        (int cur_id, std::unordered_set<int>& all_ids_, int container_id)
+    {
+        if (all_ids_.count(cur_id))
```
geoffthemedio: Comment what this doing / why?

geoffthemedio:
`objects_list` is a list, not a vector, soshould be safe to append to while being iterated through.

LGM-Doyle (commented):
You are correct the list would not have re-sized itself.  I mis-read that.  However, the original intent of this code change was to add the container id to the error message instead of just the less useful, "This id, X, does not exist" message.

geoffthemedio:
Update the description? And title?

universe/Universe.cpp
```diff
@@ -683,47 +683,60 @@ void Universe::InitMeterEstimatesAndDiscrepancies() {
...
-    std::list<int> objects_list;
-    objects_list.push_back(object_id);
+void Universe::UpdateMeterEstimates() {
+    for (int obj_id : m_objects.FindExistingObjectIDs())
+        m_effect_accounting_map[obj_id].clear();
+    // update meters for all objects.
+    UpdateMeterEstimatesImpl(std::vector<int>());
+}
+
+void Universe::UpdateMeterEstimates(int object_id, bool is_update_contained_objects) {
```
geoffthemedio: Why the "is_" in the parameter name?

universe/Universe.cpp
```diff
@@ -683,47 +683,60 @@ void Universe::InitMeterEstimatesAndDiscrepancies() {
...
+    // ids of the object and all valid contained objects
+    std::unordered_set<int> all_ids;
+
+    // Collect objects to update meter for.  this may be a single object, a group of related
+    // objects, or all objects in the (known) universe.  also clear effect accounting for meters
+    // that are to be updated. Return true on success (all_ids is valid).
+    std::function<bool (int, int)> update_obj_and_contained_objs =
```
geoffthemedio: Is this updating meters, clearing effect accounting, or collecting ids to update later? Naming seems inconsistent or incomplete.

universe/Universe.cpp
LGM-Doyle: Removed.

The "is_" is there so that the expression `if (is_X)` reads correctly.  However, `if (update_contained_objects)` is better.

universe/Universe.cpp
LGM-Doyle: I refactored it so that it only collects ids and is called `collect_ids()`.

geoffthemedio:
no further comments

PR #1952, opened 2018-01-21
Don't stop AI if it can't write config file.

o01eg:
Allows AI processes to start if config dir is read-only.


Dilvish-fo (approved):
Looks good to me

PR #1953, opened 2018-01-21
Early ai economy adjustments

Dilvish-fo:
Some adjustments to the AI's first techs and construction of the History Analyzer.  I've done my analysis relative to a species with standard research, industry and population.  I think this holds well enough still for the species with other variants.

For the backdrop on both commits, consider the standard starting tech lineup
(completion times are relative to turn 1)
Planet Eco, completing in 2 turns (turn 3)
Sub Hab, completing in 5 turns (turn 6)
Algo Elegance, completing in 7 turns (turn 8)
Phys Brain, completing in 9 turns. (turn 10)

Commit 1:
Moving Phys Brain forward into any of the first 3 spots advances its completion time to 2 turns (turn 3), leaves the completion times for Planet Eco and Sub Hab unchanged, and delays algo elegance by 1 turn.  Because of the growing population the completed construction of the History Analyzer winds up being just 5 turns sooner rather than the full 7 turns advancement of the tech, but it is still well worth the slight delay to Algo Elegance. (& I consider the benefit also much greater than the associated delay in completing any other scouts or outpost ships, etc).

Commit 2:
Completes Algo Elegance 3 turns sooner, delays Sub Hab by 3 turns.
Valuing RP at 2 PP, this adjustment is a modest improvement (and in my tests does not delay the completed construction of the History Analyzer).  At max benefit on a Large Homeworld on research focus, Sub Hab is worth 1.5 PP and ~0.15 RP (without Algo Elegance), or a total of 1.8 PP per turn, compared to the max value of Algo Elegance being ~2.1 RP (equiv to ~4.2 PP).  If there are any situations where Sub Hab would be needed for successful colonization by the initial Colony ship during those first few turns, they are fairly rare.   

I think that moving Algo Elegance before Planet Eco could also usually have a very slight benefit, but then that would also potentially restrict early colonization in a noticeable number of cases, so I choose in favor of the more flexible early colonization on that point.

### DIFF
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -1544,20 +1544,6 @@ def find_automatic_historic_analyzer_candidates():
 
     min_pp, turn_trigger, min_pp_per_additional = conditions.get(foAI.foAIstate.character.get_trait(Aggression).key,
                                                                  (ARB_LARGE_NUMBER, ARB_LARGE_NUMBER, ARB_LARGE_NUMBER))
-    # If we can colonize good planets instead, do not build this.
-    num_colony_targets = 0
-    for pid in ColonisationAI.all_colony_opportunities:
-        try:
-            best_species_score = ColonisationAI.all_colony_opportunities[pid][0][0]
-        except IndexError:
-            continue
-        if best_species_score > 500:
-            num_colony_targets += 1
-
-    num_covered = get_number_of_existing_outpost_and_colony_ships() + get_number_of_queued_outpost_and_colony_ships()
-    remaining_targets = num_colony_targets - num_covered
-    min_pp *= remaining_targets
-
     max_enqueued = 1 if total_pp > min_pp or fo.currentTurn() > turn_trigger else 0
     max_enqueued += int(total_pp / min_pp_per_additional)
 
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -314,6 +314,8 @@ def generate_production_orders():
                     if res:
                         cost, time = empire.productionCostAndTime(production_queue[production_queue.size - 1])
                         building_expense += cost / time
+                        res = fo.issueRequeueProductionOrder(production_queue.size - 1, 0)  # move to front
+                        print "Requeueing %s to front of build queue, with result %d" % ("BLD_AUTO_HISTORY_ANALYSER", res)
 
             # TODO: check existence of BLD_INDUSTRY_CENTER (and other buildings) in other locations in case we captured it
             if (total_pp > 40 or ((current_turn > 40) and (state.population_with_industry_focus() >= 20))) and ("BLD_INDUSTRY_CENTER" in possible_building_types) and ("BLD_INDUSTRY_CENTER" not in (capital_buildings+queued_building_names)) and (building_expense < building_ratio*total_pp):
--- a/default/python/AI/ResearchAI.py
+++ b/default/python/AI/ResearchAI.py
@@ -659,7 +659,7 @@ def generate_classic_research_orders():
     #
     # set starting techs, or after turn 100 add any additional default techs
     #
-    if (fo.currentTurn() == 1) or ((total_rp - research_queue.totalSpent) > 0):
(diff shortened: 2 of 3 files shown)


Morlic-fo:
There is existing fast-forward code for the History Analyzer tech:
https://github.com/freeorion/freeorion/blob/master/default/python/AI/ResearchAI.py#L762

The logic there is currently (logically) incompatible with your suggested TechListsAI change. There are two ways to resolve this (your call):

1) Tweak the ResearchAI (and/or related ProductionAI) logic and drop the changes here
2) Keep the changes here and tweak/drop the ResearchAI logic

Dilvish-fo:
Ah, thanks for pointing that out.  That existing 'fast-forward' code for the History Analyzer already seems to me to be effectively 'delay the History Analyzer' code, although that is totally my fault because I had not noticed that code the last time I increased priority (at least partially, or tried at least) for the History Analyzer, in a rushed PR just prior to 0.4.7.

I prefer the placement I propose here over the later positioning done by the ResearchAI logic.  Looking at the clock, though, I don't have time to write my reasoning up well at the moment, but I'll try to follow up very soon. 

When looking into that, I also noted that the ProductionAI  find_automatic_historic_analyzer_candidates() method had been doing more gating than I had recalled for it.  For clarity I think we should break it into two functions, one which decides whether the AI generally wants to build a History Analyzer at all, and then another which simply determines the candidate locations.  And I have  a couple tweaks I'd propose for those things, too, so I'll plan to work that up and add some more commits here.

Morlic-fo:
> That existing 'fast-forward' code for the History Analyzer already seems to me to be effectively 'delay the History Analyzer' code

IIRC the initial idea was to not cripple the AI's early game fleet development by "wasting" 20 turns or so building the history analyzer with low industry output, hence the gating in the production code. 

At that time, I think the tech wasn't added to the TechListsAI at all but rather be only enqueued by ResearchAI "fast-forwarding" as soon as it would actually make use of the tech. I still like the general idea of that approach - only research a tech if it is actually useful rather than at a static point.

However, if you think that it would be a better strategy for the AI to always rush the history analyzer, then it makes perfect sense to drop the fast-forward code in ResearchAI and lower or reduce the gating in the ProductionAI.

Dilvish-fo:
When it's a 'sometimes useful' thing, then I agree about having a thing default to being delayed, and having code to accelerate it when appropriate.  When it's a 'nearly-always-the-most-useful' thing, then I think the code should default to it being done fast, and the special-handling code can instead try to identify the rare times when it should be delayed.  I think the history Analyzer is one of those 'nearly-always-the-most-useful' things.  So much so that I don't currently see the need to try writing up code for when it should be delayed.

Consider that the History Analyzer costs about the same as colonizing another planet at that stage of the game, or else about the same as 3 troop ships or 2 Frigates.

The situations where colonizing a planet can get you as much return that quickly are if you can colonize into a system with the Philosopher planet, or colonize a good planet with Eccentric Orbit.  Both of those would give very slightly better return (and growing over time) but colonizing a planet always carries risk of getting there and/or being invaded.  And at that stage you probably still have your starter Colony Ship to use for that opportunity anyways.  If not,  I would always choose to build the Analyzer first even compared to those opportunities, for the greater safety/reliability.
 
Sometimes I am tempted to build troop ships first, but nearly always I wind up deciding to build the analyzer first, and I wouldn't worry about having the AI sometimes try to do the invasion first.   Similarly, I suppose there might be rare conditions where you could prefer to have two extra frigates rather than 5 RP per turn-- like maybe if the galaxy is super crowded or something?  Although even then I would think that trying to fight off attackers with your starting frigate while you get a big research lead would be just as viable.  But perhaps someone with more experience than I with super crowded start conditions could describe a scenario where the analyzer should be delayed.

Dilvish-fo:
p.s.
As far as I can see, the situations where ProductionAI.find_automatic_historic_analyzer_candidates() currently has the AI delaying the analyzer are for colonization opportunities (which I disagree with as discussed above and am changing), and in order to intentionally handicap the lower aggression AIs.  So even if they spend 8 RP on LRN_PHYS_BRAIN when we are not yet ready to let them build the analyzer, they are just a tiny bit more handicapped, *shrug*.

PR #1954, opened 2018-01-21
Adjustments to early ai shipbuilding

Dilvish-fo:
Generalizes the code for the AI's initial automatic ship builds, reduces number of initial scouts from 3 to 2, adds an outpost ship.

I think that most often simply 2 additional scouts are sufficient, and any more the AI can use its regular code to decide on later on.  The outpost ship is handy enough I want to make sure it gets onto the initial queue (even though it's likely to be delayed by other things such as the history analyzer).


Morlic-fo (approved):
Looks like a good improvement

PR #1956, opened 2018-01-21
Improve certain ai colony valuations

Dilvish-fo:
When valuing a primary planet, the AI takes into account whether or not some other planet can add value, like in systems with asteroids or gas giants, and whether or not it already owns that other planet.  

This PR adds an aggression based character trait for what portion of that potential added value the AI will use if it would have to invade that planet, and makes some other slight adjustments to the calculations.

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -882,7 +882,7 @@ def evaluate_planet(planet_id, mission_type, spec_name, detail=None):
                             ast_val = 0
                             break
                     elif other_planet.speciesName:
-                        if other_planet.owner==empire.empireID:
+                        if other_planet.owner == empire.empireID:
                             ownership_factor = 1.0
                         elif pid in prospective_invasion_targets:
                             ownership_factor = character.secondary_valuation_factor_for_invasion_targets()
@@ -911,7 +911,7 @@ def evaluate_planet(planet_id, mission_type, spec_name, detail=None):
                     elif other_planet.speciesName:
                         other_species = fo.getSpecies(other_planet.speciesName)
                         if other_species and other_species.canProduceShips:
-                            if other_planet.owner==empire.empireID:
+                            if other_planet.owner == empire.empireID:
                                 ownership_factor = 1.0
                             elif pid in prospective_invasion_targets:
                                 ownership_factor = character.secondary_valuation_factor_for_invasion_targets()
@@ -1007,7 +1007,7 @@ def evaluate_planet(planet_id, mission_type, spec_name, detail=None):
                 p2 = universe.getPlanet(pid)
                 if p2:
                     if p2.size == fo.planetSize.asteroids:
-                        if p2.owner==empire.empireID:
+                        if p2.owner == empire.empireID:
                             asteroid_factor = 1.0
                         elif p2.unowned:
                             asteroid_factor = 0.5
@@ -1017,7 +1017,7 @@ def evaluate_planet(planet_id, mission_type, spec_name, detail=None):
                             asteroid_factor = 0.0
                         ast_shipyard_name = p2.name
                     if p2.size == fo.planetSize.gasGiant:
-                        if p2.owner==empire.empireID:
+                        if p2.owner == empire.empireID:
                             gg_factor = 1.0
                         elif p2.unowned:
                             gg_factor = 0.5
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -8,6 +8,7 @@ import AIstate
 import FleetUtilsAI
 import FreeOrionAI as foAI
 import PlanetUtilsAI
+import PriorityAI
 import ProductionAI
 import MilitaryAI
 from turn_state import state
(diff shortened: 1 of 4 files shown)


LGM-Doyle:
It looks good to me, in particular the new code in the character module.

default/python/AI/ColonisationAI.py
```diff
@@ -654,6 +660,8 @@ def evaluate_planet(planet_id, mission_type, spec_name, detail=None):
     capital_id = PlanetUtilsAI.get_capital()
     homeworld = universe.getPlanet(capital_id)
     planet = universe.getPlanet(planet_id)
+    prospective_invasion_targets = [pid for pid, pscore, trp in
+                                    AIstate.invasionTargets[:PriorityAI.allottedInvasionTargets] if pscore > 20]
```
Morlic-fo: The value of ```allottedInvasionTargets``` is currently updated after colonization planning so it would always use last turn's value which would be zero in case of a loaded game...

It is defined as ```allottedInvasionTargets = 1 + int(fo.currentTurn() / 25)``` which is independent from other calculations, so it should probably be redefined as a function in an appropriate module which solves the timing issue (and possibly resolves some minor bugs in existing code as well...)
```
def allotted_invasion_targets():
    return 1 + int(fo.currentTurn() / 25)
```

default/python/AI/ColonisationAI.py
```diff
@@ -983,46 +999,46 @@ def evaluate_planet(planet_id, mission_type, spec_name, detail=None):
...
+                            gg_factor = 0.0
+        if asteroid_factor > 0.0:
             if tech_is_complete("PRO_MICROGRAV_MAN") or "PRO_MICROGRAV_MAN" in empire_research_list[:10]:
-                if got_owned_asteroids:  # can be quickly captured
-                    flat_industry += 5  # will go into detailed industry projection
-                    detail.append("Asteroid mining ~ %.1f" % (5 * discount_multiplier))
-                else:  # uncertain when can be outposted
-                    asteroid_bonus = 2.5 * discount_multiplier  # give partial value
-                    detail.append("Asteroid mining %.1f" % (5 * discount_multiplier))
+                flat_industry += 5 * asteroid_factor  # will go into detailed industry projection
```
Morlic-fo, line 1029: I figured the existing implementation updated the ```asteroid_bonus``` rather than the ```flat_industry``` to explicitly account for the fact that a non-owned asteroid would not add the industry bonus directly so it shouldn't be treated as a real bonus here (and in the projection).

default/python/AI/ColonisationAI.py
```diff
@@ -983,46 +999,46 @@ def evaluate_planet(planet_id, mission_type, spec_name, detail=None):
...
+                            ast_shipyard_name, discount_multiplier * 30 * pilot_val))
+        if gg_factor > 0.0:
             if tech_is_complete("PRO_ORBITAL_GEN") or "PRO_ORBITAL_GEN" in empire_research_list[:5]:
-                if got_owned_gg:
-                    flat_industry += per_ggg  # will go into detailed industry projection
-                    detail.append("GGG ~ %.1f" % (per_ggg * discount_multiplier))
-                else:
-                    gas_giant_bonus = 0.5 * per_ggg * discount_multiplier
-                    detail.append("GGG %.1f" % (0.5 * per_ggg * discount_multiplier))
+                flat_industry += per_ggg * gg_factor  # will go into detailed industry projection
```
Morlic-fo, line 1039: See comment above re ```flat_industry```

Morlic-fo (changes requested):
General idea seems good but the implementation requires some tweaks

default/python/AI/ColonisationAI.py
```diff
@@ -983,46 +999,46 @@ def evaluate_planet(planet_id, mission_type, spec_name, detail=None):
...
-                            got_owned_gg = True
-        if got_asteroids:
+                        if p2.owner==empire.empireID:
+                            gg_factor = 1.0
+                        elif p2.unowned:
+                            gg_factor = 0.5
+                        elif pid in prospective_invasion_targets:
+                            gg_factor = character.secondary_valuation_factor_for_invasion_targets()
+                        else:
+                            gg_factor = 0.0
```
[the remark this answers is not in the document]
Dilvish-fo: Right, yes I agree taking a max there is the right approach.

default/python/AI/ColonisationAI.py
Dilvish-fo: The previous code had given a half-value when the asteroids were not owned by the current empire, because it's "uncertain when can be outposted [or invaded]".  It seems to me that the treatment is mostly just generalized some now-- the main differences are that different aggression AIs can now weight that uncertainty/delay differently, and the calculation is I think improved by changing the buckets.  

As far was which bucket it goes into, it seemed best to me to make the 'asteroid_bonus' be focused on the non-industrial (i.e., currently shipbuilding) benefits from asteroids, and let the risk-adjusted industry bonus have its expected value be calculated along with other industry bonuses, which takes into account the time needed for meters to grow, rather than just leave it at the simplistic base*discount_multiplier treatment (which was essentially just treating it as if the meter instantly gets the full (or in this case, half) benefit of the bonus.  The difference becomes more significant the more other bonuses to industry there are.

The only way I really see to improve that treatment, would be rather than just risk-scaling the base bonus, we did one calc without the contingent bonuses (i.e. stemming from planets not owned by the present empire) at all, and another calc with all the contingent bonuses being given full value, and then we do the risk scaling on the difference between those two values.  I suppose we would also need to keep track of different sources of contingency maybe having different risk scales associated with them, and then wind up combining multiple different contingent projections.  That sounds to me like a good treatment, but a bit more than I can take on right now.

default/python/AI/ColonisationAI.py
Dilvish-fo: For the same reasons as above, I think it better to have this go into ```flat_industry```; I think the ```gas_giant_bonus``` is now unused/unneeded and I should just take it out entirely.

default/python/AI/ColonisationAI.py
Dilvish-fo: yes, good suggestion

default/python/AI/ColonisationAI.py
Morlic-fo: Thinking some more about it, I have no more objections to "mess up" the projected industry value. The AI will generally pretty greedily outpost asteroids in that system anyway. 

The industry projection should probably even be extended at some point in the (distant) future to also account for (speculatively) to-be-finished techs in the time period of projection.

PR #1957, opened 2018-01-25
Combine overlapping moving fleet buttons into a single FleetWnd

LGM-Doyle:
***Overview***
This PR changes the selection behavior of fleet buttons for moving fleets, so that when a fleet button for a moving fleet is selected it will open a `FleetWnd` with the fleets in the selected fleet button and any fleet buttons it touches.

This addresses the issue from the [forum](http://freeorion.org/forum/viewtopic.php?f=28&t=10853&sid=93a393c8b47910777b361103409bed10.), when fleet buttons are too close then it is not possible to select the underlying fleet button at any zoom level.

This PR also adds `FleetWnd` tracking behavior for moving fleets on subequent turns, when the fleets move.  If all the fleets, or all of the selected fleets remain "close" together, the the `FleetWnd` will remain open and track the pack of fleets movement. "close" is defined as about the size of the fleet button at the original zoom level that the `FleetWnd` was opened at.  

***Details to Aid reviewing***
The code make changes to `FleetWnd` and `MapWnd`.

The `FleetWnd` changes are to  add `m_bounding_box` which determines during each `Refresh()` if moving fleets would still fit within `m_bounding_box` if it were translated to their new location.  Otherwise, if closes the `FleetWnd`.

The `MapWnd` changes are in two parts:
- When the moving fleets buttons are created they are sorted and stored into two types: on starlanes and offroad.  This facilitates efficiently comparing for overlap only fleet buttons on the same starlane.
- When fleet buttons are clicked a new function `MapWnd::FleetIDsOfFleetButtonsOverlapping(fleet)`  returns fleet ids for overlapping buttons, with the categories of in system, moving on a starlane and off-road. These ids are used for opening `FleetWnds` and other actions.

### DIFF
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -461,9 +461,11 @@ std::set<int> FleetUIManager::SelectedShipIDs() const {
     return active_wnd->SelectedShipIDs();
 }
 
-std::shared_ptr<FleetWnd> FleetUIManager::NewFleetWnd(const std::vector<int>& fleet_ids,
-                                      int selected_fleet_id/* = INVALID_OBJECT_ID*/,
-                                      GG::Flags<GG::WndFlag> flags/* = GG::INTERACTIVE | GG::DRAGABLE | GG::ONTOP | CLOSABLE | GG::RESIZABLE*/)
+std::shared_ptr<FleetWnd> FleetUIManager::NewFleetWnd(
+    const std::vector<int>& fleet_ids,
+    double allowed_bounding_box_leeway /*= 0*/,
+    int selected_fleet_id/* = INVALID_OBJECT_ID*/,
+    GG::Flags<GG::WndFlag> flags/* = GG::INTERACTIVE | GG::DRAGABLE | GG::ONTOP | CLOSABLE | GG::RESIZABLE*/)
 {
     std::string config_name = "";
     if (!GetOptionsDB().Get<bool>("ui.fleet.multiple.enabled")) {
@@ -471,7 +473,7 @@ std::shared_ptr<FleetWnd> FleetUIManager::NewFleetWnd(const std::vector<int>& fl
         // Only write to OptionsDB if in single fleet window mode.
         config_name = FLEET_WND_NAME;
     }
-    auto retval = GG::Wnd::Create<FleetWnd>(fleet_ids, m_order_issuing_enabled, selected_fleet_id, flags, config_name);
+    auto retval = GG::Wnd::Create<FleetWnd>(fleet_ids, m_order_issuing_enabled, allowed_bounding_box_leeway, selected_fleet_id, flags, config_name);
 
     m_fleet_wnds.insert(std::weak_ptr<FleetWnd>(retval));
     retval->ClosingSignal.connect(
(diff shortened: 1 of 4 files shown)


LGM-Doyle:
@dbenage-cx  and @Dilvish-fo  you may wish to review this PR, since you both made suggestions in the forum.

UI/FleetWnd.cpp
```diff
@@ -2769,10 +2771,33 @@ int FleetDetailPanel::ShipInRow(GG::ListBox::iterator it) const {
...
+        else
+            return GG::Rect(
+                std::min(box.Left(),    pt.x),
+                std::min(box.Top(),     pt.y),
+                std::max(box.Right(),   pt.x),
+                std::max(box.Bottom(),  pt.y));
+    }
+
+    /** Can \p bounds encompass \p candidate? */
+    bool CouldContain(GG::Rect bounds, GG::Rect candidate) {
```
geoffthemedio: somewhat misleading function name... doesn't check if `candidate` is or could be within, just that it is smaller or equal in size.

UI/FleetWnd.cpp
```diff
@@ -3003,13 +3047,18 @@ void FleetWnd::Refresh() {
...
+        // Grow the fleets bounding box
+        fleets_bounding_box = CreateOrGrowBox(fleets_that_exist.empty(), fleets_bounding_box, fleet_loc);
+
         fleets_that_exist.insert(fleet_id);
-        fleet_locations_ids.insert({{fleet->SystemID(), GG::Pt(GG::X(fleet->X()), GG::Y(fleet->Y()))},
-                                    fleet_id});
+        fleet_locations_ids.insert({{fleet->SystemID(), fleet_loc}, fleet_id});
 
     }
+    GG::Pt bounding_box_center{fleets_bounding_box.MidX(), fleets_bounding_box.MidY()};
```
geoffthemedio: might make your initialization format a bit more consistent... here `GG::Pt b_b_c{}`, above `auto f_l = GG::Pt()`.

UI/FleetWnd.cpp
Dilvish-fo: If I am understanding right, this is being used to check if fleets sharing a FleetWnd have dispersed too far, from one turn to the next, to still share the FleetWnd, and that suggests a names to me like ```FleetsHaveNotDispersed``` or ```FleetsStillClustered```.

geoffthemedio: The function itself has nothing to do with fleets, so I'd have its name describe what it actually does, not how it's used in one case.

dbenage-cx:
Fleets near systems seem a more common case for me, but also generally easier to zoom in to distinctively select (not necessary for this PR imo).

From brief testing, looks great for occurrences away from systems. 👍

UI/FleetWnd.cpp
dbenage-cx: Name suggestions seem easier to come up with if args are swapped: `CouldRelocateWithin`

Dilvish-fo: > so I'd have its name describe what it actually does, not how it's used in one case.

It seems to me that I am proposing to have the name describe what it actually does in the most significant sense-- the determination that it is making, rather than the details of the logic used to make that determination.  And I am puzzled about the "not how it's used in one case" -- it appears to me that this function is used in exactly two places, and in both places it is used the same general way with the same the significance.

The name ```CouldRelocateWithin``` could be technically accurate as suggested, but it would seem to me to imply a different significance and purpose for the function than what it actually has, which would make code harder to understand rather than easier.

dbenage-cx: In this sense, I'd expect `FleetsX(non-Fleet args)` to derive fleets from args.
Would view this function with a potential to move into GG/PtRect as non-/member function if later desired for other cases.

UI/FleetWnd.cpp
LGM-Doyle: Fixed in two locations.

LGM-Doyle (commented):
Thanks everyone for the quick reviews.

UI/FleetWnd.cpp
geoffthemedio: I'd prefer `SmallerThan` and `BiggerThan` for this case.

LGM-Doyle: I changed it to `bool SmallerOrEqual(GG::Rect ll, GG::Rect rr)`.

geoffthemedio:
A quirk / bug: I've got two fleets in a single window, both moving to the same place. When I right click on a destination, the other is removed from the fleets window. Clicking the fleet icon again brings them back, but by opening a new fleet window of a different size than the one I started with. If I close that and click on the fleet button again, I get back the original sized fleet window with both fleets in it.
before
after_move_order_click
after_fleet_button_click

PR #1960, opened 2018-01-29
AI support for playing SLY

### LINKED ISSUE #1958: Missing AI support for species Sly
Introduced in #1928 

* Possibly some wrong values for gas giants as previously assumed uninhabitable
* AI does not know about the different gas giant generator value on gas giants

Dilvish-fo:
Informs the AI about the new species The Sly and related changes.  Tries to keep a starting-Sly empire in stealthy mode during the initial growth stages by reducing outposting and military priorities, currently until the empire has acquired another colonizer.  

More can be done, but this seems to be enough to make the AI fairly competent with the Sly.

Resolves #1958

### DIFF
--- a/default/python/AI/PriorityAI.py
+++ b/default/python/AI/PriorityAI.py
@@ -240,9 +240,8 @@ def _calculate_colonisation_priority():
     allottedColonyTargets = 1 + int(total_pp * turns_to_build * allotted_portion / colony_cost)
     outpost_prio = foAI.foAIstate.get_priority(PriorityType.PRODUCTION_OUTPOST)
     # if have any outposts to build, don't build colony ships TODO: make more complex assessment
-    if outpost_prio > 0 or num_colonies > colony_growth_barrier:
+    if outpost_prio > 0:
         return 0.0
-
     if num_colonies > colony_growth_barrier:
         return 0.0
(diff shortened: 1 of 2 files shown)


Morlic-fo:
Looks ok. Did you test and verify that the AI successfully builds and uses colony ships as sly? I imagine there might be issues regarding the supply evaluation in ColonisationAI.evaluate_planet(). Perhaps the supply penalty there can/should be reduced for sly colonization?

> reducing outposting and military priorities

I could not find an explicitly reduced military priority in the diff. Is this a missing commit, outdated description or some indirectly implied change?

Dilvish-fo:
> I could not find an explicitly reduced military priority in the diff

Ah, I think I was a bit too tired when I pushed this and made the PR, lost track a bit of implemented vs planned.  

I've added the work in progress label now, will try to get around to the military aspect also.

> Did you test and verify that the AI successfully builds and uses colony ships as sly?

Yes, although it could be they are still making more outposts than would really be best, I'll continue trying to fine tune this a bit more.

agrrr3:
Should we merge this and you put up another PR with more polishing? So you can get more playtesting feedback from people

Vezzra:
I guess it would be beneficial for playtesting the Sly if basic AI support gets merged into master, even if more work on it is planned for 0.4.8...

PR #1961, opened 2018-01-29
Auto-upgrade of attack troops on GRO_NANO_CYBERNET

agrrr3:
Instead of unlock GT_TROOPS_POD_2 researching GRO_NANO_CYBERNET doubles
the troop capacity of GT_TROOPS_POD if these get built later on.

[Discussion](http://www.freeorion.org/forum/viewtopic.php?f=15&t=10865) in the forum

Differences to before (some may be considered bugs):
* lot less of clutter of troop ship designs
* advanced troops dont have increased cost (was double the cost before)
* advanced troops need only 2 instead of 4 defensive troops at build location
* Pedia information lists troop capacity of the part always as 2


agrrr3:
Didnt check for runtime delays (having one effect more).

And probably, if we want to go this route, AI also needs to be adjusted. So could one please mark this as work in progress until the discussion is more advanced?

agrrr3:
According to the discussion, we would like to have the parts separated for flexibility.

The basic UI problem got also already addressed by a better UI for deprecating parts.

PR #1962, opened 2018-01-29
[tests] Fix class name

Cjkjvfnby:
I just fix module name.
Since we have continuous integration, we should setup it to run this tests.


Morlic-fo:
Good catch, the tests path weren't in my IDE project settings so these were missed when renaming.

PR #1963, opened 2018-02-01
Imperial Stockpile (4th revision)

agrrr3:
So this became the fifth iteration [implementation design for fourth iteration](http://www.freeorion.org/forum/viewtopic.php?f=5&t=10810&p=91404#p91404):

- no input limits and stockpile use enabled by default (geoff already in master)
-  **sly growth nerf; size 3 equivalent adjustment for gas giants**
- ~~also **apply _POPULATION to self-sustaining** (also nerfs Sly population)~~
- **yet another sly stockpile boost**- 1000%
- **flux bubble nerf** (remove auto-unlock, more expensive/hitpoints)
- **special refueling for sly on gas giants** 0.1 fuel extra per turn
- **gas giant + space elevator nerf** (only +3 supply instead +4; sly minus one supply)
- populated gas giant nerf (**planets get only +5 for populated generator in the same system**
; @freeorion/ai-team AI probably needs adjustments???)
- **no terraforming for sly**

IMHO its interesting to play sly now, maybe they are underpowered compared to Laenfa
So this became the fifth iteration [implementation design for fourth iteration](http://www.freeorion.org/forum/viewtopic.php?f=5&t=10810&p=91404#p91404):

- no input limits and stockpile use enabled by default (geoff already in master)
-  **sly growth nerf; size 3 equivalent adjustment for gas giants**
- ~~also **apply _POPULATION to self-sustaining** (also nerfs Sly population)~~
- **yet another sly stockpile boost**- 1000%
- **flux bubble nerf** (remove auto-unlock, more expensive/hitpoints)
- **special refueling for sly on gas giants** 0.1 fuel extra per turn
- **gas giant + space elevator nerf** (only +3 supply instead +4; sly minus one supply)
- populated gas giant nerf (**planets get only +5 for populated generator in the same system**
; @freeorion/ai-team AI probably needs adjustments???)
- **no terraforming for sly**

IMHO its interesting to play sly now, maybe they are underpowered compared to Laenfa


default/scripting/buildings/GAS_GIANT_GEN.focs.txt
```diff
@@ -16,29 +16,37 @@ BuildingType
...
+            activation = ContainedBy And [
+                Planet
+                Not Population high = 0
+            ]
             stackinggroup = "GAS_GIANT_GEN_STACK"
-            priority = [[VERY_LATE_PRIORITY]]
-            effects = SetTargetIndustry value = Value + 10
+            // if there is an unpopulated gas giant generator planet then that bonus applies
+            // VERY_LATE_PRIORITY__PLUS_ONE
+            priority = 121
```
Dilvish-fo: Making the priority for the populated case be different than the priority for the non-populated case seems to me to invite trouble

default/scripting/species/common/population.macros
```diff
@@ -129,7 +129,7 @@ SELF_SUSTAINING_BONUS
                 //TargetPopulation low = 0
             ]
             accountinglabel = "SELF_SUSTAINING_LABEL"
-            priority = [[VERY_LATE_PRIORITY]]
+            priority = [[EARLY_PRIORITY]]
```
Dilvish-fo: To be consistent with the idea (noted in the script comment) that self-sustaining is meant to be equivalent to having 3 growth specials, then if you change this then you'd also need to change growth specials to have them apply early also, or you wind up nerfing the self-sustaining max pop relative to the other metabolisms.  And I think we'd have to look really closely at things if we considered changing the timing of the growth specials.  

A lot of time went into a previous pass at getting the various growth techs & specials to work out right; since it seeems you are really just looking for another way to nerf the Sly here, I think it would probably be better to look a little farther and leave self-sustaining alone.

agrrr3: Ah because nobody objected to the idea i thought i take it in. I take the commit out then.

Dilvish-fo: Well, let's see about getting some other feedback-- as I think about it more, this would not make self-sustaining *always* worse than the other metabolisms, in the case of a Good Pop species this would give self-sustaining a bigger boost than other metabolisms.  So perhaps we just change the code comment to say self-sustaining is meant to be "roughly equivalent" to having three growth specials. 
@Vezzra @MatGB I recall you guys putting a fair bit of thought into Population scripting, do you guys have an opinion on this?

default/scripting/species/common/supply.macros
```diff
@@ -2,6 +2,15 @@ NO_SUPPLY
 '''
 '''
 
+VERY_BAD_SUPPLY
+'''     EffectsGroup
+            description = "VERY_BAD_SUPPLY_DESC"
+            scope = Source
+            activation = Planet
+            accountinglabel = "VERY_BAD_SUPPLY_LABEL"
```
Dilvish-fo, line 10: ```VERY_BAD_SUPPLY_DESC``` and ```VERY_BAD_SUPPLY_LABEL``` are still lacking stringtable entries, it seems.

default/scripting/species/common/population.macros
Morlic-fo: Regardless of whether or not the change to self-sustaining priority makes sense in general, I really don't think it should be a part of this PR.

Please also note that effect timing may (and in this case does) need tweaks in the AI code.

Vezzra: > @Vezzra @MatGB I recall you guys putting a fair bit of thought into Population scripting, do you guys have an opinion on this?

Actually that has been @MatGB (I haven't fiddled with the pop scripting), so I guess he might want to say something on the subject.

But of course I have an opinion on this as well, I agree with you, @Dilvish-fo, and @Morlic-fo: we should leave the self-sustaining metabolism scripts alone. If the Sly need some nerfing wrt to their pop levels, I suggest to just do that by giving them the bad or very bad pop trait.

If that isn't sufficient, add bad industry/research. There are plenty of species traits we can use to fine-tune them. But please refrain from changing things that will also change other things (by that I also refer to the approach of changing the planet size of gas giants!).

Trying to make a species balanced by fiddling around with settings/scripting that affects/changes other species or game elements as well carries a great potential of introducing a lot of issues. Which, IMO, should be avoided.

agrrr3: Regarding changing the gas giants size - as @Dilvish-fo is on board for the AI team this is Ok, right? 

@Vezzra probably you dont have much time at your hands right now for the project. I and others spend a lot of time on discussions, and then there is a conclusion/decision. And then work gets done to have it implemented. If you rightly object in this stage, a lot of effort is wasted. If you can chime in earlier it would be greatly appreciated.

dbenage-cx: >  (by that I also refer to the approach of changing the planet size of gas giants!)

As I understand it, the planet size is not changed, only the value returned from [`SizeAsInt`](https://github.com/freeorion/freeorion/blob/master/universe/Planet.cpp#L203), which is not the same as [`PlanetSize`](https://github.com/freeorion/freeorion/blob/master/universe/Enums.h#L59) cast to int.

Propose `SizeAsInt` function/keywords should be renamed to avoid confusion (HabitableSize?), such need not be done in this PR imo.

Vezzra: @agrrr3, I understand the frustration, but as everyone else I'm doing this in my free time for fun, and life tends to get in the way much more than I'd like. I'm trying, but there is only so much priority I'm willing to give something that's just a hobby, after all.

Which is why I don't expect anyone to receive my input as something they must abide, but as mere suggestions/proposals. 😉 

That said, in this particular case I responded to a request to comment that had only been made a day ago, so that couldn't have been too much of a problem - unless you refer to my objection to the change of the planet size of gas giants. In this case I'm sorry, I didn't realize that reverting that decision (and trying a different approach) would undo so much work.

It certainly can remain, at least for the time being, as a stop-gap solution until we can come up with better ideas how to balance the Sly. However, regardless if I mentioned it or not, the issue I pointed out is still true, and sooner or later we'd have to put in the work to solve this properly anyway. It's not my fault no one else noticed the problem with that approach... 😉

Vezzra: @dbenage-cx:

> As I understand it, the planet size is not changed, only the value returned from [`SizeAsInt`](https://github.com/freeorion/freeorion/blob/master/universe/Planet.cpp#L203), which is not the same as [`PlanetSize`](https://github.com/freeorion/freeorion/blob/master/universe/Enums.h#L59) cast to int.

Well, but shouldn't it be? I always thought `SizeAsInt` has been added to provide content scripts with the means to use the planet size in formulas. Having `SizeAsInt` returning the same value for gas giants and large planets is asking for (future) trouble IMO. Right now the planet size of gas giants isn't used anywhere else (IIRC), so it won't cause any immediate side effects, but that can very well change in the future.

> Propose `SizeAsInt` function/keywords should be renamed to avoid confusion (HabitableSize?)

I wouldn't change `SizeAsInt`, but *add* a function/keyword `HabitableSize`, so you can still refer to the actual planet size in a formula, as well as the "population capacity". That should go a long way to mitigate a lot of potential issues.

Nevertheless, even with that distinction available, I'm still hesitant to make gas giants like large planets wrt to pop capacity just because we have trouble balancing one species. Not a big issue though, and certainly not something we need to address right now. Plenty of time for that should there ever arise any problems.

> such need not be done in this PR imo

Agreed. But it definitely should be done at some point, I'd even suggest to open an issue for that (otherwise it will surely be forgotten 😉).

default/scripting/buildings/SPACE_ELEVATOR.focs.txt
```diff
@@ -12,7 +12,7 @@ BuildingType
     ]
     EnqueueLocation = [[ENQUEUE_BUILD_ONE_PER_PLANET]]
     effectsgroups = [
-        // check GasGiant first, so not another match
+        // gas giant is first in the stacking group in order not to trigger another effect by planet size
```
geoffthemedio: GasGiant is also a planet size.

https://github.com/freeorion/freeorion/blob/master/parse/PlanetSizeValueRefParser.cpp

agrrr3: I couldnt figure out where it does the equality check using the enum. is it already specified in [ConditionParser6.cpp](https://github.com/freeorion/freeorion/blob/57425eb8a95d672b90976a46539b62ebbaf234ee/parse/ConditionParser6.cpp#L112)?

Anyway so should I remove the comment and the stackinggroup i introduced?

geoffthemedio:
You need to rebase, not merge master.

https://github.com/freeorion/freeorion/blob/master/CONTRIBUTING.md

default/scripting/buildings/SPACE_ELEVATOR.focs.txt
geoffthemedio: I don't understand the first question.

I don't know what the point of the stacking group is, so I don't know.

agrrr3:
@geoffthemedio yeah i tried to change the space elevator using the github webpage (because i didnt have access to my development box) without realising that it would perform a merge

default/scripting/buildings/SPACE_ELEVATOR.focs.txt
agrrr3: I thought that the size conditions would check for the numeric size. Now the GG numeric size is 4, same as large, so I thought both the effect for large would kick in for a GG. In order to prevent it i introduced the priority/stackinggroup.

(and my first question was related - where in the code is the comparison for the Planet size = xxx condition? so i could have had a look there and understand for myself)

agrrr3:
I rebased, removed the unnessecary stuff from the space elevator and tested the change. 
So are we good for merge? 

@freeorion/ai-team any concerns from you? should i open issues for some topics?

default/scripting/buildings/GAS_GIANT_GEN.focs.txt
```diff
@@ -16,29 +16,36 @@ BuildingType
             scope = And [
                 Planet
                 InSystem id = Source.SystemID
-                Planet type = GasGiant
```
geoffthemedio, line 19: What's the point of these changes? You're removing the scope conditions that limit the effects to gas giants or non-gas giants, and adding conditions that the building only functions when it is contained by a planet with no population or with population. The building should always be on a planet, so that part of the activation is redundant... But I'm confused by changing the scope discrimination between gas giant and other, to an activation condition based on population or not.

agrrr3: I reversed the application:
Before the scope for the +10/+5 effect was differentiating between gas giants and other planets. 
Now the source makes the difference for all planets.

So in effect, if there is no unpopulated gas giant generator (+10 bonus), a populated gas giant generator adds +5 
Before only the populated gas giant had only +5, the other planets had the +10 bonus,.

About Planet condition in the containedBy in the activation: the building is also contained by a system, isnt it?

geoffthemedio: Yes; systems contain fleets, ships, planets, buildings, and possibly fields. Fleets contain ships, and planets contain buildings.

agrrr3: I was afraid that without the Planet condition, the ContainedBy would put the system and the planet in the scope. Not sure what the Population condition does with Systems, but if it isnt ignored it would do the wrong thing in one of the EffectGroups.

geoffthemedio: `ContainedBy Planet` will match things contained by planets, not things contained by some system... `Contains Planet` would match things that contain a planet, such as a system.

agrrr3: exactly yes, thats why i added the Planet condition to the ContainedBy, because you asked

quoting you: "adding conditions that the building only functions when it is contained by a planet with no population or with population. The building should always be on a planet, so that part of the activation is redundant..."
(i dont think its redundant)

geoffthemedio: Buildings are always on a planet. Therefore, checking in the activation condition if the building (which is what is being tested by an activation condition) is on a planet is pointless.

This has nothing to do with the scope.

agrrr3: Oh right

agrrr3: Got confused ... activation not scope..
..but...
the activation does not check if the building is contained by a planet

the activation checks if the planet containing the building  is populated or not.

i would think that one the following would evaluate to true for the containing system regardless of the containing planet

> activation = ContainedBy Not Population high = 0

> activation = ContainedBy Population high = 0

geoffthemedio: Systems don't have population. But OK, if you want to filter out the system before testing the population, fine.

geoffthemedio:
@freeorion/ai-team comments or objections?

default/scripting/species/SP_SLY.focs.txt
```diff
@@ -24,7 +24,7 @@ Species
 
         [[BAD_POPULATION]]
```
geoffthemedio: Instead of using the standard population effectsgroup macros and needing to adjust gas giant size to balance just this species, could give them a custom alternative / lower population mechanic...

agrrr3: Would like to do that in the next iteration. I'd prefer more playtesting results first.

default/scripting/buildings/GAS_GIANT_GEN.focs.txt
agrrr3: Thanks for you patience geoff, I figured out why the problem i was referring to does not exist in this case (The +10 bonus has priority, so I dont have to constrain the activation condition for the +5), so I could simplify the activation condition even more (Source instead of ContainedBy Not Population high = 0).

geoffthemedio: I'd actually keep the activation conditions to clarify the intent better than just having the stacking groups and relying on the ordering of effects in the script.

default/scripting/species/SP_SLY.focs.txt
Dilvish-fo: > Would like to do that in the next iteration. 

Both Vezzra and Geoff have now raised concerns (which make sense to me) about the whole approach of trying to balance the Sly by changing the more general characteristic of gas giant size.   That doesn't seem like a suitable situation for making a change while planning to undo it in the near future. 

So, it seems to me that shifting the Sly from Bad Population to a new Very Bad Population with factor -0.5 would accomplish very nearly the same thing (subject perhaps to what priority it got), and that would be a better thing to try here.

agrrr3:
No prob. Ill whip something up.

Am 08.02.2018 00:35 schrieb "Dilvish-fo" <notifications@github.com>:

agrrr3:
But jeep in mind that this is only a sly issue because it is the only GG
species. The same issue of extreme pop growth is there for any GG species.

So probably i will do sth like a minor metabolism (so not full blown like
the 5 existing ones)

Am 08.02.2018 11:27 schrieb "Axel Groß" <axel.gross@iaeste.at>:

default/scripting/species/SP_SLY.focs.txt
geoffthemedio: I was thinking they could be giving a completely separate population mechanic... like fixed 5 max pop, and not fit into the good/average/bad system.

agrrr3: with your suggestion none of the growth techs would mean anything. that doesnt seem right.
hm i think for such things we'll need a design discussion in the forum. for this iteration i will do something analog to the habitable size 4

agrrr3:
So i wasnt really able to balance sly with an planet-equivalent size of 4, so now gasgiants are like medium sized planets. Self-sustaining is gone, but gas giants are now good environments for sly.
The whole thing is a bit more like the other species, so its easier to compare/balance but less distinct. 
It would be nice to add a gas giant native species without colonization but with planet-equivalent size of 6. Also was thinking about differentiating types of gas giants, but that will be a new forum discussion.

default/stringtables/en.txt
```diff
@@ -7164,14 +7177,16 @@ Sly
...
+Greatly prefer [[PT_GASGIANT]]s.
+
+[[encyclopedia GASEOUS_SPECIES_TITLE]]
 '''
 
 SP_SLY_DESC
 '''Description:
 The Sly live in gas giants. The gas giants are so foreign to most species in the universe that, for most of their history, they have been overlooked.
 Not being able to build devices, they couldn't leave their homeworld. Recently, they were able to obtain necessary equipment and technology for spacefaring from traders in exchange for their knowledge of history.
+Keeping a strict diet a Sly crew on gas giants actually can help in refinining fuel for their ship.
```
geoffthemedio: this sentence is borderline incoherent without a comma or rewording / reordering

default/stringtables/en.txt
```diff
@@ -14093,6 +14108,10 @@ GREAT_WEAPONS_DESC
 ULTIMATE_WEAPONS_DESC
 +++	Ultimate Pilots: Base damage per ship weapon increased by three levels.
 
+GASEOUS_DESC
+'''+	Lives on Gas Giants
```
geoffthemedio: Probably `Lives` -> `Live`

default/stringtables/en.txt
```diff
@@ -7186,7 +7186,7 @@ SP_SLY_DESC
 '''Description:
 The Sly live in gas giants. The gas giants are so foreign to most species in the universe that, for most of their history, they have been overlooked.
 Not being able to build devices, they couldn't leave their homeworld. Recently, they were able to obtain necessary equipment and technology for spacefaring from traders in exchange for their knowledge of history.
-Keeping a strict diet a Sly crew on gas giants actually can help in refinining fuel for their ship.
+On the right diet Sly excrement refined fuel. When staying on a gas giant, a Sly crew will create some fuel for the ship.
```
geoffthemedio: , after diet
excrement -> excrete
remove "some"
When in a system with a gas giant, a Sly-crewed ship will regenerate fuel.

PR #1970, opened 2018-02-03
Update tests, setup travis

Cjkjvfnby:
Need to remove failing test before merge.  I leave it as example, to allow check how error output looks like.

Suppress steps defined in global scope
```
before_script: echo ""
before_cache: echo ""
```

This allow to have "name" in test output
```
env: NAME="Pytest"
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,16 @@ matrix:
         # Auth token to push API documentation
         - secure: "JKeXk8p65hodb12PVRST6A90swsNubc+46EbSJGSghldIxbFWLBAlwU+KLeOMO4V0veu6k4lnMa50V0UYFZmoUsS6W0aL5Ybo98SpzXHiNLOmOluoqJoF9TBsOTCCRFbWbccgJyVEtulgRcdml96naS51lq9Sw/VO/N3Z472304="
     - os: linux
+      language: python
+      python: 2.7.14
+      install:
+        - pip install pytest==3.4.0
+      script:
+        - pytest
+      before_script: echo ""
+      before_cache: echo ""
+      env: NAME="Pytest"
+    - os: linux
       services:
         - docker
       env:
@@ -56,14 +66,14 @@ before_script:
   - |
     # Make certain ~/.ccache exists
     mkdir -p $HOME/.ccache;
-    
+
     if [ -e /Users/travis/.ccache/ccache.conf ]; then
        echo "Show global /Users/travis/.ccache/ccache.conf"
        more /Users/travis/.ccache/ccache.conf
     else
        echo "Missing global /Users/travis/.ccache/ccache.conf"
     fi;
-    
+
     if [ -e $HOME/.ccache/ccache.conf ]; then
        echo "Show local $HOME/.ccache/ccache.conf"
        more $HOME/.ccache/ccache.conf


.travis.yml
```diff
@@ -12,6 +12,16 @@ matrix:
         - BUILD_APIDOC=true
         # Auth token to push API documentation
         - secure: "JKeXk8p65hodb12PVRST6A90swsNubc+46EbSJGSghldIxbFWLBAlwU+KLeOMO4V0veu6k4lnMa50V0UYFZmoUsS6W0aL5Ybo98SpzXHiNLOmOluoqJoF9TBsOTCCRFbWbccgJyVEtulgRcdml96naS51lq9Sw/VO/N3Z472304="
+    - os: linix
```
adrianbroher: `linux`, not `linix`.

.travis.yml
```diff
@@ -12,6 +12,16 @@ matrix:
...
+      language: python
+      python: 2.7.14
+      install:
```
adrianbroher: I would rather prefer to move the `install`, `script` and `before_*` sections out of the build matrix and add them rather as conditional steps below.

But of now it can stay as it is.

.travis.yml
```diff
@@ -12,6 +12,16 @@ matrix:
...
+        - pip install pytest
+      script:
+        - pytest
+      before_script: echo ""
```
adrianbroher: Why do you want to suppress the global build steps anyway?

.travis.yml
```diff
@@ -12,6 +12,16 @@ matrix:
...
+      before_cache: echo ""
+      env: NAME="Pytest"
```
adrianbroher, line 24: Please go with `BUILD_<stage name>` (here: `BUILD_PYTEST`) as environment variable to match the logic in the other build matrix selectors.

.travis.yml
```diff
@@ -12,6 +12,16 @@ matrix:
         - BUILD_APIDOC=true
         # Auth token to push API documentation
         - secure: "JKeXk8p65hodb12PVRST6A90swsNubc+46EbSJGSghldIxbFWLBAlwU+KLeOMO4V0veu6k4lnMa50V0UYFZmoUsS6W0aL5Ybo98SpzXHiNLOmOluoqJoF9TBsOTCCRFbWbccgJyVEtulgRcdml96naS51lq9Sw/VO/N3Z472304="
+    - os: linix
+      language: python
+      python: 2.7.14
```
adrianbroher: Need to look up if this matches with the docker image, but it's okay for now.

.travis.yml
```diff
@@ -12,6 +12,16 @@ matrix:
...
+      install:
+        - pip install pytest
```
adrianbroher: I doubt any `pytest` version is fine. Please use an explicit version marker like `pip install pytest==2.7.3`.

adrianbroher (changes requested):

adrianbroher:
The current test configuration doesn't seem to do a lot of test, especially style related ones. Only a few selected test scripts.  Is this intentional?

.travis.yml
Cjkjvfnby: good catch, strange that it works

.travis.yml
Cjkjvfnby: Matrix `include` directive create separate job. For that job `before_script` is not required and it breaks script part, because it ends with `cd ...`

.travis.yml
Cjkjvfnby: My IMHO that `install`, `scripts` and `before_*` should not be global and present in matrix for each job, but I did not know how to reuse code properly, may be put `*.sh` script to repository.

.travis.yml
Cjkjvfnby: This variable is used only for user interface of travis.  I plan to add same variable to compile stage too.
name

Cjkjvfnby:
> The current test configuration doesn't seem to do a lot of test, especially style related ones. Only a few selected test scripts. Is this intentional?

It is all tests we have for python (unittests-integration). Style checks will be run as separate job, I will add them in next PR.

Cjkjvfnby:
@adrianbroher?

Cjkjvfnby:
Pytest changes are already merged.

Cjkjvfnby:
@adrianbroher, I fixed/responded for your comments.

adrianbroher:
Looks okay so far.  Can you remove the intentionally failing unit test before merging?

adrianbroher (approved):

PR #1972, opened 2018-02-04
Improve DesignWnd (obsolete parts and hulls, mass replace) 

LGM-Doyle:
This PR makes the following changes to the DesignWnd
- It allows drag and drop re-ordering of hulls, which is persisted in the save games.
- It allows ctrl-click (and a right click menu option) to obsolete hulls.  An obsolete hull will transitively obsolete any designs using that hull. But the design itself is not obsolete, so if the hull is un-obsoleted the design comes back.
- It allows ctrl-click (and a right click menu option) to obsolete parts, which similarly to hulls, transitively obsoletes designs.
- It allows ctrl-click-drag to remove all parts of the same type from a design on the main panel
- It allows a part placed while holding down ctrl to replace all parts (including the empty slot) of the same type, subject to slot type requirements.
- It updates the design window pedia entry to reflect these changes.

This PR was in response to the [forum discussion](http://freeorion.org/forum/viewtopic.php?f=15&t=10865&sid=be022186e2d62ab750099133bd9dca08) about making it easier to obsolete parts.


agrrr3:
Tested the obsoletion stuff and think its already great 🥇 :)

What I would love though - at the moment its not possible to un-obsolete a single design if one of its parts is obsolete (probably same if the hull is obsolete).
I would love to say make all these obsolete, buuut I still need this single design for some time.
If that is not easy, the menu should show that its not possible to un-obsolete the design.

geoffthemedio:
Alternatively, could add a filter to the producible items list that shows obsolete designs.

LGM-Doyle:
@agrrr3, thanks for taking the time to review this.  Your idea is good and I think I have implemented it.

I've changed the behavior so that obsoleting a design is recorded only if the change would be different from the current obsolete state based on hull and part obsoleting alone.

Your example now works.  It is possible to obsolete a hull/part, then show the obsolete designs and un-obsolete one or more designs.

@geoffthemedio, I'd be happy to extend the obsolete filtering to the production window.  I'd like to do it in a separate PR also allowing players to obsolete buildings, so there is consistency in the production window UI and all the different types of producible items can be obsoleted with the same interface.

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+
+        public:
+        AvailabilityManager(bool obsolete, bool available, bool unavailable);
+
+        bool GetAvailability(const Availability::Enum type) const;
+        void SetAvailability(const Availability::Enum type, const bool state);
+        void ToggleAvailability(const Availability::Enum type);
+
+        /** Return none or the shown state of the design \p id */
+        boost::optional<Availability::Enum> ShowDesignAs(int id) const;
```
geoffthemedio: Name sounds like a verb (mutator) but is actually a getter... Maybe better like `GetDesignAvailability(int id)` or similar?

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+        obsolete_parts = m_obsolete_parts;
+    }
+
+    //////////////////////////////////////////////////
+    //  AvailabilityManager                         //
+    //////////////////////////////////////////////////
+
+    /** A class to allow the storage of availabilty in a common location. */
+    class AvailabilityManager {
+        private:
```
geoffthemedio: dedent

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+        if (hull.empty()) {
+            ErrorLogger() << "Hull name is empty()";
+            return ;
+        }
+
+        if (hull == next_hull)
+            return ;
+
+        // if hull already exists, this is a move.  Remove the old location
+        auto existing_it =  m_hull_to_obsolete_and_loc.find(hull);
```
geoffthemedio: extra space

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+            std::get<Availability::Available>(m_availabilities) = state;
+            break;
+        case Availability::Future:
+            std::get<Availability::Future>(m_availabilities) = state;
+            break;
+        }
+    }
+
+    void AvailabilityManager::ToggleAvailability(const Availability::Enum type) {
+        SetAvailability(type, !GetAvailability(type));
```
geoffthemedio: prefer `{}` on same line for single line function

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+        const Empire* empire = GetEmpire(empire_id);  // may be 0
+        bool available = empire ? empire->ShipPartAvailable(id) : true;
+
+        const auto& manager = GetCurrentDesignsManager();
+        const auto obsolete = manager.IsPartObsolete(id);
+
+        return ShowAs(available, obsolete);
+    }
+
+    boost::optional<Availability::Enum> AvailabilityManager::ShowAs(bool available, bool obsolete) const {
```
geoffthemedio: declaration line a bit line... would move params to new line

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
     }
 
+    void CurrentShipDesignManager::InsertHullBefore(const std::string& hull, const std::string& next_hull) {
+        if (hull.empty()) {
+            ErrorLogger() << "Hull name is empty()";
+            return ;
+        }
+
+        if (hull == next_hull)
+            return ;
```
geoffthemedio: extra space

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+        return false;
+    }
+
+    bool CurrentShipDesignManager::IsHullObsolete(const std::string& hull) const {
+        auto it_hull = m_hull_to_obsolete_and_loc.find(hull);
+        return (it_hull != m_hull_to_obsolete_and_loc.end() && it_hull->second.first);
+    }
+
+    bool CurrentShipDesignManager::IsPartObsolete(const std::string& part) const {
+        return m_obsolete_parts.count(part);
```
geoffthemedio: prefer `{}` on one line

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+            if (IsPartObsolete(part) != obsolete) {
+                it->second.first = obsolete;
+                return;
+            }
+        }
+
+        it->second.first = boost::none;
+    }
+
+    void CurrentShipDesignManager::SetHullObsolete(const std::string& id, const bool obsolete) {
```
geoffthemedio: I think `id` should be `name` ?

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+
+        bool GetAvailability(const Availability::Enum type) const;
+        void SetAvailability(const Availability::Enum type, const bool state);
+        void ToggleAvailability(const Availability::Enum type);
+
+        /** Return none or the shown state of the design \p id */
+        boost::optional<Availability::Enum> ShowDesignAs(int id) const;
+        /** Return none or the shown state of the hull \p id */
+        boost::optional<Availability::Enum> ShowHullAs(const std::string& id) const;
+        /** Return none or the shown state of the part \p id */
```
geoffthemedio: meaning of "the shown state of the hull/part" isn't clear to me.

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+    //////////////////////////////////////////////////
+    //  AvailabilityManager                         //
+    //////////////////////////////////////////////////
+
+    /** A class to allow the storage of availabilty in a common location. */
+    class AvailabilityManager {
+        private:
+        // A tuple of bools for the state of the 3 availability filters:
+        // Obsolete, Available and Unavailable
+        std::tuple<bool, bool, bool> m_availabilities;
```
geoffthemedio: Just having three bool members seems to me like it would be a lot simpler than dealing with the tuple in all the code below.

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+        }
+    }
+
+    void AvailabilityManager::ToggleAvailability(const Availability::Enum type) {
+        SetAvailability(type, !GetAvailability(type));
+    }
+
+    boost::optional<Availability::Enum> AvailabilityManager::ShowDesignAs(int id) const {
+        int empire_id = HumanClientApp::GetApp()->EmpireID();
+        const Empire* empire = GetEmpire(empire_id);  // may be 0
```
geoffthemedio: may be `nullptr` ?

UI/DesignWnd.cpp
```diff
@@ -676,53 +708,320 @@ namespace {
...
+        if (showing_obsolete && obsolete && showing_available && available)
+            return Availability::Obsolete;
+
+        if (showing_obsolete && obsolete && showing_future && !available)
+            return Availability::Obsolete;
+
+        if (showing_obsolete && obsolete && !showing_available && !showing_future)
+            return Availability::Obsolete;
+
+
```
geoffthemedio: intentional extra blank line?

UI/DesignWnd.cpp
```diff
@@ -2532,6 +2824,36 @@ void SavedDesignsListBox::BaseLeftClicked(GG::ListBox::iterator it, const GG::Pt
 }
 
 
+void EmptyHullsListBox::BaseRightClicked(GG::ListBox::iterator it, const GG::Pt& pt,
+                                               const GG::Flags<GG::ModKey>& modkeys)
```
geoffthemedio: dedent to match first character of previous line's first parameter

UI/DesignWnd.cpp
```diff
@@ -2532,6 +2824,36 @@ void SavedDesignsListBox::BaseLeftClicked(GG::ListBox::iterator it, const GG::Pt
...
+{
+    HullAndPartsListBoxRow* hull_parts_row = dynamic_cast<HullAndPartsListBoxRow*>(it->get());
```
geoffthemedio: I'd have used `auto` in this case

geoffthemedio:
otherwise OK

LGM-Doyle (commented):
Once again, thanks for the thorough review.

dbenage-cx:
Very nice changes here!

>  It is possible to obsolete a hull/part, then show the obsolete designs and un-obsolete one or more designs.

Might make some mention in pedia that un-obsoleting a design will override hull/part obsolete conditions.
e.g. (Tested via #1982) Given a design with Large Basic Hull and Mass Driver:
* Obsolete Large Basic Hull -> design obsolete
* Un-obsolete design
* Obsolete Mass Driver -> design remains non-obsolete

LGM-Doyle:
@dbenage-cx, thanks for testing and finding the bug that you reported.  I can replicate it.  I will work on a solution that does not "forget" that the design should be un-obsolete if a player flickers (toggles twice) the part (Mass Driver).

PR #1976, opened 2018-02-06
Cleanup AI code

Cjkjvfnby:
Fix/suppress style warnings

### DIFF
--- a/default/python/tox.ini
+++ b/default/python/tox.ini
@@ -7,6 +7,7 @@ putty-ignore =
   *.py : +E501 # temp ignore of the line too long
   AI/AIDependencies.py : +E241
   AI/FreeOrionAI.py : +E402,E722
+  AI/savegame_codec/__init__.py : +F401
   auth/auth.py : +E402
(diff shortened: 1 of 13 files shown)


default/python/tox.ini
```diff
@@ -4,6 +4,7 @@ exclude =
   AI/
 putty-ignore =
   *.py : +E121,E123,E126,E226,E24,E704,W503 # default ignore list
+  *.py : +F812 # list comprehension redefines '<variable>' from line <line_no>
```
Morlic-fo, line 7: I'd strongly prefer to keep that rule active. This is more than just a code-style issue and may introduce hard to find bugs.
```
foo = 'test'
bar = [foo for foo in range(3)]
print foo  # oops it's 2 now!
```

Cjkjvfnby: I plan to fix it later, after this check will be included to the CI  run.

Dilvish-fo (approved):
Subject to later stopping ignoring the list comprehension redefine issue as you discuss, sure, looks fine to me.

PR #1977, opened 2018-02-06
Planet habitable size

### LINKED ISSUE #1975: Planet size vs habitable size leads to confusion
In order to make it feasible that species live on gas giants without being overpowered, we chang(ed) the Planet::SizeAsInt() for gas giants from 4 to 6. 

Usually gas giants are thought to a lot bigger than large planets (SizeAsInt -> 4).

ATM the size is used for display and ordering in the objects view and calculating target population boni.
But it is also planned to use the size for other things in scripting.

We should do something about this.

- either add another method like HabitableSizeAsInt which gets used for population
- or introduce extra population malus for gas giant species (e.g. an extra minus 25% which applies to population boni as well); this maybe is fiddly for balancing and it is shown as malus in the effects

dbenage-cx:
Provides `Planet::HabitableSize` defined from game rules.

Changes `SizeAsDouble` value ref to return `PlanetSize` implicitly cast to double.

Replaces existing uses of `SizeAsDouble` in FOCS definitions with new `HabitableSize`, for consistency with previous values.

Initial discussion https://github.com/freeorion/freeorion/pull/1963#discussion_r165982968

Resolves #1975 

Minor conflict with #1869 #1963

@freeorion/ai-team I've only changed the comment of planet_size_map and exposed `habitableSize`, might be worth reviewing uses of `planetSize`.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1102,6 +1102,30 @@ Starlanes Everywhere
 RULE_STARLANES_EVERYWHERE_DESC
 Starlanes are generated between every pair of systems, regardless of geometry. Lanes may still be removed during a game, however. Lanes are also not rendered on the galaxy map.
 
+RULE_HABITABLE_SIZE_TINY
+Tiny planets Habitable size
+
+RULE_HABITABLE_SIZE_SMALL
+Small planets Habitable size
+
+RULE_HABITABLE_SIZE_MEDIUM
+Medium planets Habitable size
+
+RULE_HABITABLE_SIZE_LARGE
+Large planets Habitable size
+
+RULE_HABITABLE_SIZE_HUGE
+Huge planets Habitable size
+
+RULE_HABITABLE_SIZE_ASTEROIDS
+Asteroids Habitable size
+
+RULE_HABITABLE_SIZE_GASGIANT
+Gas Giants Habitable size
+
+RULE_HABITABLE_SIZE_DESC
+Value used to adjust for [[encyclopedia POPULATION_TITLE_SHORT_DESC]] changes on a planet of this size.
+
 
 ##
 ## Command-line and options database entries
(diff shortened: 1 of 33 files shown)


Dilvish-fo:
Hmm, yes, in looking at the AI code a bit more, I saw there were still a few places where the AI code had still been misusing planet.size as the habitable size instead of using the planet_size_as_int function from AIDependencies.py.  I have now pushed a fix for that to master now, having them refer to AIDependencies.planet_size_as_int.

That leaves three uses of AIDependencies.planet_size_as_int, which could be cleanly dealt with as an extra commit in this PR-- removing ```planet_size_map``` and  ```planet_size_as_int``` from AIDependencies.py, and replacing the following calls to ```AIDependencies.planet_size_as_int(planet.size)``` with ```planet.habitableSize()```

```
default/python/AI/ColonisationAI.py:62:    planet_size = AIDependencies.planet_size_as_int(planet.size)
default/python/AI/ColonisationAI.py:245:                                                     AIDependencies.planet_size_as_int(planet.size))
default/python/AI/ResourcesAI.py:437:            pop_gain = potential_pop_increase - AIDependencies.planet_size_as_int(planet.size)
```

dbenage-cx:
@Dilvish-fo Thank you for addressing those, updated as suggested (sans call).

Rebased

Dilvish-fo:
I am pretty sure that the cause of the new conflict is just that CJ's python codestyle cleanup added a couple blank lines between the planet_size_map dict and the planet_size_as_int() function.

geoffthemedio:
@Dilvish-fo any further concerns for this?

Dilvish-fo (approved):
Looks good to go, to me

default/stringtables/en.txt
```diff
@@ -1102,6 +1102,30 @@ Starlanes Everywhere
 RULE_STARLANES_EVERYWHERE_DESC
 Starlanes are generated between every pair of systems, regardless of geometry. Lanes may still be removed during a game, however. Lanes are also not rendered on the galaxy map.
 
+RULE_HABITABLE_SIZE_TINY
+Tiny planets Habitable size
```
geoffthemedio, line 1106: This names are too long to fit in the rules list, but that can be fixed later...

PR #1978, opened 2018-02-07
Fix wrong de-stringtable of cruiser/destroyer and laser/md

agrrr3:
When checking the obsoletion of ship designs I saw that the german names for the large mark designs were off.

### DIFF
--- a/default/stringtables/de.txt
+++ b/default/stringtables/de.txt
@@ -320,16 +320,16 @@ SD_MARK1_DESC
 Einfache Fregatte mit Massenkatapult. Kann nur auf Kolonien mit einem [[buildingtype BLD_SHIPYARD_BASE]] gebaut werden.
 
 SD_LARGE_MARK_1
-Zerstörer (M1)
+Kreuzer (M)
 
 SD_LARGE_MARK_2
-Zerstörer (M2)
+Kreuzer (L)
 
 SD_LARGE_MARK_3
-Kreuzer (L1)
+Zerstörer (M)
 
 SD_LARGE_MARK_4
-Kreuzer (L2)
+Zerstörer (L)
 
 SD_GRAVITATING1
 Gravitierend I


default/stringtables/de.txt
```diff
@@ -320,16 +320,16 @@ SD_MARK1_DESC
 Einfache Fregatte mit Massenkatapult. Kann nur auf Kolonien mit einem [[buildingtype BLD_SHIPYARD_BASE]] gebaut werden.
 
 SD_LARGE_MARK_1
-Zerstörer (M1)
+Kreuzer (M)
```
geoffthemedio, line 323: Why `Kreuzer (M)` and not `Kreuzer Ms` or more similar to the english equivalent? https://github.com/freeorion/freeorion/blob/master/default/stringtables/en.txt#L390

adrianbroher: Probably because it is not clear what `Ms` or `Lz` stands for.  I just recently saw the same translation, but didn't fix this because the meaning of the suffix was unclear.

Someone should add a translators note in the en.txt file.

adrianbroher: @MatGB *poke* You made those translation changes.

geoffthemedio: Probably those designs should be renamed in english as well then...

Dilvish-fo: I would vote for adding an explanation of the naming scheme somewhere, rather than just doing away with it.

Dilvish-fo: The naming scheme does seem like it probably works well enough in German, yes?  's' still for Standard [Armor], z still for Zortrium, etc.?   I expect that for some parts the naming scheme would have to use different initials between English and German, but that still doesn't really seem like a problem to me.

geoffthemedio:
Regardless of discussion, seems to be an improvement, so will merge. Other adjusted can be done separately.

PR #1980, opened 2018-02-08
Increase colony base hull cost to 3

Dilvish-fo:
As discussed [on the forums](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10854), tripling base hull cost to further reduce the viability of mass-producing comsats as a defensive measure.

We seem to have had concurrence there for us to take this step, but I'm posting as a PR to give anyone else a chance to notice and comment.


geoffthemedio:
I'm inclined to merge this immediately so it can be in the next test build...

PR #1984, opened 2018-02-08
[cleanup] cleanup ai

Cjkjvfnby:


default/python/AI/ResearchAI.py
```diff
@@ -1059,8 +1059,8 @@ def generate_classic_research_orders():
             continue
         old_designs = ShipDesignAI.WarShipDesigner().optimize_design(consider_fleet_count=False)
         new_designs = ShipDesignAI.WarShipDesigner().optimize_design(additional_hulls=unlocked_hulls,
-                                                                          additional_parts=unlocked_parts,
-                                                                          consider_fleet_count=False)
+                                                                     additional_parts=unlocked_parts,
+                                                                     consider_fleet_count=False)
```
Dilvish-fo, line 1063: Both from appearance here, and trying to count characters, these two lines do not seem to be getting aligned with the start of "additional_hulls", like it seems to me they should.

Cjkjvfnby: Looks perfectly aligned for me.  How do you see it?
align

Dilvish-fo: Ah, I had github showing the diff in split view, and then (I guess because of some strange wraparound issue) both of the changed lines were showing up as differently indented than additional_hulls and different from each other (and I just made a mistake when I tried counting characters).  Changing the diff to unified view shows them lining up fine.

Dilvish-fo (approved):

PR #1985, opened 2018-02-09
Stockpile auto-use update

Ouaz:
As usual, let me know if my written english needs rewording.


default/stringtables/en.txt
```diff
@@ -3992,7 +3992,7 @@ It also allows for some interaction with the [[encyclopedia MAP_WINDOW_ARTICLE_T
 The Production Summary panel shows available Production Points.
 If a system is selected, there are two columns shown. The left column shows production points across your entire empire. The right column gives production points for the selected system (but is absent if no system is selected).
 Available Points relates to the PP available for production. The right value are the maximum available production points for the selected system. This may be less than the total available when all of your empire's systems are not [[metertype METER_SUPPLY]] connected.
-Stockpile relates to the PP available for production using the imperial stockpile. Production items specifically need to be enabled to draw from the stockpile.
+Stockpile relates to the PP available for production using the imperial stockpile. Production items are enabled to draw from the Imperial Stockpile by default and have to be explicitly denied for doing so.
```
Dilvish-fo: It's definitely good to get this fixed, thanks.

I would suggest "disabled from" rather than "denied for", both here and below.

Ouaz: Thanks for the feedback. Commit force-updated with "disabled from".

Issue #1986, opened 2018-02-11
Not all species getting meter regen


Dilvish-fo:
After attacking an Acirema native planet and departing, its shields and defenses did not regenerate.  Looking over  their file, and #1941, it appears to me that they slipped through unadjusted for the new meter growth mechanic.  Counting the number of species files modified in 1941 (31 by my count) and comparing to the number of species files (44 active it looks to me), it seems a number of others must have slipped through also.

geoffthemedio:
Whether an individual species script is modified doesn't indicate much, since most of these sorts of effects are scripted through macros. Try with that pull request: https://github.com/freeorion/freeorion/pull/1987

Dilvish-fo:
Yes, that the basically solve the problem for the Acirema, but it seems to still leave things not quite 100% right.  Their meters began to regen ***on the second turn*** after my frigate left their system  (and to be very clear, I mean that on turn N+0 I ordered my Frigate to leave their system, on turn N+1 my Frigate had left but they had no regen, and on turn N+2 their shields and defenses regened by 1 point each), whereas in the past planetary meters would begin to regen on the first turn that there was no longer combat. 

This remaining problem seems like a different/broader broader than what I opened this Issue for; if you don't immediately see a fix for this remaining problem we should probably close this Issue as resolved (once we merge the PR) and open a new Issue for the remaining bit.

In case you want to test things out a bit, the galaxy settings are below, and the Acirema system was 3 jumps to the SW.
```
Seed: yedzhU9k
Number of Systems: 300
Shape: Random
Age: Medium
Starlanes: Medium
Planets: Medium
Specials: Medium
Monsters: Medium
Natives: Medium
AI Max Aggression: Maniacal
```

geoffthemedio:
Removed some - 1 from the tests on turns when planets have been attacked that gate when meters grow... might help. https://github.com/freeorion/freeorion/commit/9f7a1cfe0267bc6cf1fff2d8f4c39023aaa06d72

Dilvish-fo:
Yes, those (and the additional supply one) look right and fixed things for my Acirema test case.

PR #1987, opened 2018-02-11
Fix native planet shield and defense growth macros. Hopefully resolve…

geoffthemedio:
…s issue #1986


geoffthemedio:
Wasn't itself merged, but the equivalent commits were pushed directly to master. This doesn't fit with the superceded or duplicate status descriptions, so I went with merged.

PR #1988, opened 2018-02-11
Convert Terraforming building to Terraform Center

dbenage-cx:
Changes Terraforming to a once per planet building, changing environment over time.

Initial discussion in [Forum thread](http://www.freeorion.org/forum/viewtopic.php?f=15&t=10363).

WIP pending forum feedback


default/scripting/buildings/TERRAFORM.focs.txt
```diff
@@ -16,20 +15,13 @@ BuildingType
...
     ]
-    EnqueueLocation = And [
-        Planet
-        OwnedBy empire = Source.Owner
-        Not Contains Building name = "BLD_TERRAFORM_REVERT"
-        Not Contains Building name = "BLD_GAIA_TRANS"
-        Not HasSpecial name = "GAIA_SPECIAL"
-        Not Enqueued type = Building name = "BLD_TERRAFORM_REVERT"
-        Not Enqueued type = Building name = "BLD_GAIA_TRANS"
-    ]
```
Vezzra, line 27: I'm probably missing something important here, but why remove the `EnqueueLocation` clause entirely? I can see where you have to remove the `Not ... Building name = "BLD_TERRAFORM_REVERT"` parts, but shouldn't enqueuing a Terraformer building be prevented on planets with the Gaia special/building (present or enqueued)?

Moreover, with the proposed changes to the Terraformer building (one building doing all the terraforming required to achieve "good" environment), shouldn't enqueuing more than one of them at the same location be prevented?

default/scripting/buildings/TERRAFORM.focs.txt
```diff
@@ -16,20 +15,13 @@ BuildingType
...
     effectsgroups = [
         EffectsGroup
             scope = And [
                 Object id = Source.PlanetID
                 Planet
+                Not HasSpecial name = "UNSTABLE_SPECIAL"
+                Not Planet type = LocalCandidate.NextBetterPlanetType
```
Vezzra, line 24: Maybe I'm not understanding how that condition works right, but wouldn't that stop terraforming one step prior to reaching the optimal environment?

Vezzra (commented):
As far as I can see, with the implementation as proposed, the Terraforming building would not remove itself after completing the terraforming process. Is that intended?

default/scripting/buildings/TERRAFORM.focs.txt
dbenage-cx: IIRC, `EnqueueLocation` inherits from `location` when omitted. This does need re-added though as you are correct, this building should not be enqueue-able a second time.

default/scripting/buildings/TERRAFORM.focs.txt
dbenage-cx: Prevents the effect if the resulting type would be the same as current type.

o01eg:
Do we need it after https://github.com/freeorion/freeorion/pull/3634 ?

geoffthemedio:
probably can be considered superceded

PR #1989, opened 2018-02-12
Fix AI being kicked out from multiplayer lobby.

o01eg:
When multiplayer lobby kick out a player before he will be established e.g. for insufficient roles the server also kick out all AI from lobby because them have same player id (-1).

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -609,6 +609,10 @@ sc::result MPLobby::react(const Disconnection& d) {
     // if the disconnected player wasn't in the lobby, don't need to do anything more.
     // if player is in lobby, need to remove it
     int id = player_connection->PlayerID();
+    if (id == Networking::INVALID_PLAYER_ID) {
+        DebugLogger(FSM) << "MPLobby.Disconnection : Disconnecting player (" << id << ") was not established";
+        return discard_event();
+    }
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Can you add some comments explaining this / what is going on in that bit if code?

o01eg:
Yep. Added comment here.

server/ServerFSM.cpp
```diff
@@ -609,6 +609,11 @@ sc::result MPLobby::react(const Disconnection& d) {
     // if the disconnected player wasn't in the lobby, don't need to do anything more.
     // if player is in lobby, need to remove it
     int id = player_connection->PlayerID();
+    // if player was not etablished disconnect him as AI also have player_id is Networking::INVALID_PLAYER_ID
```
geoffthemedio: What this means isn't clear. Please elaborate.

olegshtch: Changed. It's mean that non-stablished player have an id (`INVALID_PLAYER_ID`) and AI entries have same id so if it will be processed it kick AIs from lobby.

o01eg: It's mean non-established player have an id `INVALID_PLAYER_ID` so if it will be processed it kick out AIs from lobby because they have a same id.

PR #1990, opened 2018-02-13
Move mock location

Cjkjvfnby:
I move generated mock to python folder.   Now it located in same subfolder where commons located. 

This changes will allow to have autocomplete support for mock and commons by adding single PYTHONPATH to IDE.


default/python/freeOrionAIInterface.py
```diff
@@ -1,3 +1,6 @@
+# Autogenerated do not modify manually!
```
Dilvish-fo: If we left this as a 'mock' freeOrionAIInterface.py, this comment would need to be augmented with an explanation of what the file is.   But it's one thing to have a mock file tucked away into a directory labeled as such, it's another to move it into one of our regular directories.

What I really I think we need to do is make this into a [Pep 484 stub file](https://www.python.org/dev/peps/pep-0484/#stub-files), which don't seem to often carry such comments, though I think it would still be helpful to lead with some lines like
```
# Autogenerated do not modify manually!
# This is a type-hinting python stub file, used by python IDEs to provide type hints.  For more information
# about stub files, see https://www.python.org/dev/peps/pep-0484/#stub-files
# During execution, the actual freeOrionAIInterface module is made available to the AI via
# a C++ Boost-python process as part of the AI launch.
```

Although the current form of the file has a lot of statements that don't seem to me like they would normally go into a stub file, a review of the PEP indicates that they don't seem to be barred, either, and in my brief tests they did not seem to cause any trouble.  Rather than moving this file to ```default/python/freeOrionAIInterface.py```, we can simply rename and move it to ```default/python/AI/freeOrionAIInterface.pyi``` and ouila! We have a working stub file (and no need for the user to twiddle with their PYTHONPATH).

Changes to the format/contents of the file, to make it more in line with standard stub files, could come later, it seems to me.

@Morlic-fo @LGM-Doyle Do either of you have much experience with python stub files yet?  (I'm not sure which IDEs beyond Pycharm and Wing actually support them yet.)  I invite discussion about the approach I suggest here.

default/python/interface_mock/interface_inspector.py
```diff
@@ -141,7 +141,7 @@ def inspect(obj, instances, classes_to_ignore=tuple()):
                               but have no methods, to get their instances.
     """
     debug("\n\nStart generating skeleton for %s\n\n" % obj.__name__)
-    folder_name = os.path.join(os.path.dirname(__file__), 'result')
+    folder_name = os.path.normpath(os.path.join(os.path.dirname(__file__), '..'))  # python/ folder
```
Dilvish-fo: So this would change

default/python/interface_mock/interface_inspector.py
```diff
@@ -141,7 +141,7 @@ def inspect(obj, instances, classes_to_ignore=tuple()):
...
     result_path = os.path.join(folder_name, '%s.py' % obj.__name__)
```
Dilvish-fo: and this extension would change to .pyi

Dilvish-fo (changes requested):

Cjkjvfnby:
> and no need for the user to twiddle with their PYTHONPATH

And we have `commons` and users still need to twiddle with their PYTHONPATH

Dilvish-fo:
>>and no need for the user to twiddle with their PYTHONPATH

>And we have commons and users still need to twiddle with their PYTHONPATH

I suppose that portion of my comment was especially superficial since if you just set up the project root as default/python then I think it all works out fine (for the thing in common and whatnot, but not the mock interface in its current location) without any special adjustments to PYTHONPATH (and would then handle your the moved mock interface fine also).  But that doesn't change my main point.

From your other PR I get the impression you are possibly wanting to have more functionality in the mock interface than what a typehinting stub file would have, and I guess that could be fine for that testing purpose, keeping it in its current location.  But I would still want to be enabling the typehinting via a standard stub file rather than via putting some semi-functional mock interface of the same name into the main python directories.

Cjkjvfnby:
I have moved stub location and renamed it to pyi. 

Next two step I want to do: 
  - renmae `mock` to `stub` for consistency
  - rework generation to have stub according pep-484

@Dilvish-fo, should I do all this in this PR or make it in separate?

Dilvish-fo:
If you need/want to move ```freeorion.py``` up to the ```default/python``` directory, please add an introductory docstring or comment to the file,explaining what it is there for, akin to what you've added to freeOrionAIInterface.pyi

Regarding that latter file, I had thought from your other PR that we already merged, that we were planning to have two files going forward, one this .pyi stub file for IDE typehinting, and the other being the mock interface file that would facilitate testing.  The files would start out with identical content, just different names, but would be expected to diverge in content as they each get more tuned to their particular purpose.  This seems to entirely move & rename the mock interface file, and the generator script does not seem to be modified to create two files.  Am I misunderstanding your plan?

As for all-in-this PR or via multiple PRs, I think it would probably be clearest if the reworking the stub file content were in its own PR.  As for renaming a directory and a file or two, that seems fine to just add here, or it's fine as a separate PR if you think that would be more clear.

Cjkjvfnby:
I have finish renaming and moving stubs.

Also I have removed duplicated entries in `tox.ini`

Dilvish-fo:
Perhaps I am overlooking something that should be obvious, but I find myself still having my previous question:
>  I had thought from your other PR that we already merged, that we were planning to have two files going forward, one this .pyi stub file for IDE typehinting, and the other being the mock interface file that would facilitate testing. The files would start out with identical content, just different names, but would be expected to diverge in content as they each get more tuned to their particular purpose. This seems to entirely move & rename the mock interface file, and the generator script does not seem to be modified to create two files. Am I misunderstanding your plan?

Dilvish-fo:
Is the file that you have now moved and renamed 'freeorion.pyi' the one you were planning on using for providing the FoAI constants for testing?  If so, I don't think it should (or even can) be a '.pyi' stub file.  If it has some other purpose, please explain for me, thanks.

Cjkjvfnby:
> Perhaps I am overlooking something that should be obvious, but I find myself still having my previous question:

Sorry, I answered you in my head.

This is two different files for different purposes.  

- Stub is auto generated and used for type hinting in IDE
- Mock is manually written, it contains only code that required for tests, nothing more.  Stub file without implementation is useless mocking.

Dilvish-fo (approved):

PR #1992, opened 2018-02-13
Remove stringtable includes

adrianbroher:
This PR partially removes the misuse of the translation API to get content specific configuration values.

Addresses parts of #907


parse/SpeciesParser.cpp
```diff
@@ -276,17 +279,82 @@ namespace {
...
         const lexer lexer;
-        start_rule_payload species_;
+        start_rule_payload::first_type species_;
+        start_rule_payload::second_type ordering;
+
+        boost::filesystem::path manifest_file;
 
         for (const boost::filesystem::path& file : ListScripts(path)) {
-            /*auto success =*/ detail::parse_file<grammar, start_rule_payload>(lexer, file, species_);
+            if (file.filename() == "SpeciesCensusOrdering.focs.txt" ) {
```
dbenage-cx, line 339: Can we use .inf for such files, or should we rename the existing .inf files?

universe/Species.h
```diff
@@ -218,8 +218,11 @@ class FO_COMMON_API SpeciesManager {
     { bool operator()(const std::map<std::string, std::unique_ptr<Species>>::value_type& species_entry) const; };
 
 public:
-    using SpeciesTypeMap = std::map<std::string, std::unique_ptr<Species>>;
-    using iterator = SpeciesTypeMap::const_iterator;
+    using SpeciesTypeMap = std::pair<
```
geoffthemedio: I'd have two variables, rather than making things complicated by wrapping two containers into a pair and then having `.first` and `.second` and `::second_type` everywhere...

parse/SpeciesParser.cpp
geoffthemedio: I think the distinction is that .inf files shouldn't contain scripted content... though in that case, monster fleets and categories should be in .focs.txt rather than .inf files. But the list in this pull request would fit in a .inf.

Regardless, lists like this, such as https://github.com/freeorion/freeorion/blob/master/default/scripting/starting_unlocks/items.inf

have no initial header line. The header is unnecessary if the file can only contain one top-level entry, so I think there is no need for `SpeciesCensusOrdering` in the file or parser.

Empire/Empire.cpp
```diff
@@ -42,53 +42,31 @@ namespace {
         rules.Add<bool>("RULE_STOCKPILE_IMPORT_LIMITED",
                         "RULE_STOCKPILE_IMPORT_LIMITED_DESC",
                         "", false, true);
+
+        rules.Add<double>("RULE_PRODUCTION_QUEUE_FRONTLOAD_FACTOR",
+                          "RULE_PRODUCTION_QUEUE_FRONTLOAD_FACTOR_DESC",
+                          "", 0.0, true, RangedValidator<double>(0.0, 30.0));
+        rules.Add<double>("RULE_PRODUCTION_QUEUE_TOPPING_UP_FACTOR",
+                          "RULE_PRODUCTION_QUEUE_TOPPING_UP_FACTOR_DESC",
+                          "", 20.0, true, RangedValidator<double>(0.0, 30.0));
```
geoffthemedio: these as rules seems to make a lot of sense.

parse/SpeciesParser.cpp
adrianbroher: @geoffthemedio 

Noted.  The implementation in this PR is pretty much a copy&paste job from the current ShipDesignParser, which uses a similar manifest file introduced with commit 36d659be52ba5e494a4c206b67e58b9ae1398c50

universe/Species.h
adrianbroher: I will try to change that.

adrianbroher:
@geoffthemedio Can you take a look at the new english string table entries?   I doubt those convey the purpose of the new `RULE_PRODUCTION_QUEUE_*` rules properly.

geoffthemedio:
The stringtable entries don't really explain what the rules do. I don't fully understand what they do either, though, so perhaps @Dilvish-fo would be better to ask for a concise description...

parse/SpeciesParser.cpp
geoffthemedio: Fair enough. My interpretation of .inf vs. .focs.txt is a bit speculative anyway...

universe/Species.h
adrianbroher: Fixed with e59ea1e066896434c8b4b3c6da1b79d409219edd

parse/SpeciesParser.cpp
adrianbroher: So should this be rewritten to use `*.inf` semantics or not?

adrianbroher:
>  I don't fully understand what they do either, though, so perhaps @Dilvish-fo would be better to ask for a concise description...

When building some item the game predicts how many PP should be spent per turn to build that item.  However during changes in the current production capabilities it is possible that the item doesn't collect enough PP in the predicted build turns and stays in an "almost complete" state, requiring one more turn than necessary, even when there are more than enough PP available in the first place.  The variables allow to tune how much overallocation of PP in percent the game is allowed to assign to the item construction during every build turn or/and at the final build turn to build the item in the predicted amount of turns.

geoffthemedio:
I understand that, but the details of the tuning are a bit vague for me, and those details are what needs to be described.

parse/SpeciesParser.cpp
geoffthemedio: I'm not sure. There are examples using both formats. Since I'm not sure, and nobody else who has a strong opinion has commented, I suppose it's fine to leave as you have it.

Dilvish-fo:
I believe that the shift to tracking progress by percentage completion that was done something over a year ago really did away with the need for these, both by my recollection of the issues, and the code comments
> //...Sometimes complications
// arise, though, which unexpectedly delay the completion even if the item
// had been fully-funded every turn, because costs have risen partway
// through (such as due to increasing ship costs resulting from recent ship
 // construction completion and ensuing increase of Fleet Maintenance costs.

I think that really we probably just want to yank these, probably after a bit of testing to be confident they are really unneeded now.  The Front load factor already defaults to zero, let's set the topping-up default to zero also, leave the descriptions as adrianbroher has them for now, and let that get play-tested a bit.

geoffthemedio:
> set the topping-up default to zero

I suppose this is OK then, once that is changed.

adrianbroher:
Thanks for the feedback.

PR #1994, opened 2018-02-14
Dont show GASEOUS_DESC for all species

agrrr3:
This PR moves the GASEOUS_DESC to an effect which is in the SP_SLY, before the sly boni were shown for all species

### DIFF
--- a/default/scripting/species/SP_SLY.focs.txt
+++ b/default/scripting/species/SP_SLY.focs.txt
@@ -40,6 +40,7 @@ Species
 
         // Extra refueling on gas giants
         EffectsGroup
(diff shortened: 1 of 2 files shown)


default/scripting/species/SP_SLY.focs.txt
```diff
@@ -40,6 +40,7 @@ Species
 
         // Extra refueling on gas giants
         EffectsGroup
+            description="GASEOUS_DESC"
```
geoffthemedio: spaces around `=`

geoffthemedio:
ok

PR #1997, opened 2018-02-14
Add Game UID universe property.

o01eg:
This PR adds gameUID property to `Universe` class.
It allows to follow the game session in saved games and separate different game sessions.

Hostless mode uses subfolder based on gameUID instead of "auto" subfolder so it simplify find saves.

Current implementations simply uses random UUID from python on creating universe or from boost on loading legacy save files.

### DIFF
--- a/default/python/universe_generation/universe_generator.py
+++ b/default/python/universe_generation/universe_generator.py
@@ -144,6 +144,19 @@ def create_universe(psd_map):
     seed_rng(seed_pool.pop())
     distribute_specials(gsd.specials_frequency, fo.get_all_objects())
 
+    # set game uid
+    empire_names = []
+    for psd in psd_map.values():
+        empire_names.append(psd.empire_name)
+
+    empire_names.sort()
+    for i, v in enumerate(empire_names):
+        empire_names[i] = v.capitalize()[:2]
+
+    fo.get_galaxy_setup_data().gameUID = "".join(empire_names) + str(random.randint(0,999)).zfill(3)
+
+    print "Game UID %s" % fo.get_galaxy_setup_data().gameUID
+
     # finally, write some statistics to the log file
     print "############################################################"
     print "##             Universe generation statistics             ##"
--- a/default/stringtables/ru.txt
+++ b/default/stringtables/ru.txt
@@ -3782,6 +3782,7 @@ ENC_GALAXY_SETUP
 # %8% monster frequency in the current galaxy.
 # %9% natives frequency in the current galaxy.
 # %10% upper AI aggression level in this galaxy.
+# %11% game UID
 ENC_GALAXY_SETUP_SETTINGS
 '''Семя: %1%
 Количество систем: %2%
@@ -3793,6 +3794,7 @@ ENC_GALAXY_SETUP_SETTINGS
 Монстры: %8%
 Аборигены: %9%
 Макс. агрессивность ИИ: %10%
+Идентификатор игры: %11%
 '''
 
 ENC_SHIP_PART
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -556,7 +556,7 @@ void ServerApp::NewSPGameInit(const SinglePlayerSetupData& single_player_setup_d
 
 bool ServerApp::VerifySPGameAIs(const SinglePlayerSetupData& single_player_setup_data) {
     const auto& player_setup_data = single_player_setup_data.m_players;
-    return NewGameInitVerifyJoiners(single_player_setup_data, player_setup_data);
+    return NewGameInitVerifyJoiners(player_setup_data);
 }
 
 void ServerApp::NewMPGameInit(const MultiplayerLobbyData& multiplayer_lobby_data) {
@@ -633,7 +633,7 @@ void ServerApp::NewMPGameInit(const MultiplayerLobbyData& multiplayer_lobby_data
     }
 
     NewGameInitConcurrentWithJoiners(multiplayer_lobby_data, psds);
-    if (NewGameInitVerifyJoiners(multiplayer_lobby_data, psds))
+    if (NewGameInitVerifyJoiners(psds))
         SendNewGameStartMessages();
 }
 
@@ -737,13 +737,8 @@ void ServerApp::NewGameInitConcurrentWithJoiners(
     m_universe.UpdateStatRecords();
 }
 
-bool ServerApp::NewGameInitVerifyJoiners(
-    const GalaxySetupData& galaxy_setup_data,
-    const std::vector<PlayerSetupData>& player_setup_data)
-{
+bool ServerApp::NewGameInitVerifyJoiners(const std::vector<PlayerSetupData>& player_setup_data) {
     DebugLogger() << "ServerApp::NewGameInitVerifyJoiners";
-
-    m_galaxy_setup_data = galaxy_setup_data;
 
     // associate player IDs with player setup data.  the player connection with
     // id == m_networking.HostPlayerID() should be the human player in
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -2531,7 +2531,8 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
     }
 
     if (server.IsHostless() && GetOptionsDB().Get<bool>("save.auto.hostless.enabled")) {
-        boost::filesystem::path autosave_dir_path = GetServerSaveDir() / "auto";
+        std::string subdir = GetGalaxySetupData().GetGameUID();
+        boost::filesystem::path autosave_dir_path = GetServerSaveDir() / (subdir.empty() ? "auto" : subdir);
         const auto& extension = MP_SAVE_FILE_EXTENSION;
         // Add timestamp to autosave generated files
         std::string datetime_str = FilenameTimestamp();
--- a/util/MultiplayerCommon.cpp
+++ b/util/MultiplayerCommon.cpp
@@ -282,7 +282,8 @@ GalaxySetupData::GalaxySetupData(GalaxySetupData&& base) :
     m_monster_freq(base.m_monster_freq),
     m_native_freq(base.m_native_freq),
     m_ai_aggr(base.m_ai_aggr),
-    m_game_rules(std::move(base.m_game_rules))
+    m_game_rules(std::move(base.m_game_rules)),
+    m_game_uid(std::move(base.m_game_uid))
 { SetSeed(m_seed); }
 
 const std::string& GalaxySetupData::GetSeed() const
@@ -340,6 +341,9 @@ Aggression GalaxySetupData::GetAggression() const
 const std::vector<std::pair<std::string, std::string>>& GalaxySetupData::GetGameRules() const
 { return m_game_rules; }
 
+const std::string& GalaxySetupData::GetGameUID() const
+{ return m_game_uid; }
+
 void GalaxySetupData::SetSeed(const std::string& seed) {
     std::string new_seed = seed;
     if (new_seed.empty() || new_seed == "RANDOM") {
@@ -350,6 +354,10 @@ void GalaxySetupData::SetSeed(const std::string& seed) {
         DebugLogger() << "Set empty or requested random seed to " << new_seed;
     }
     m_seed = new_seed;
+}
+
+void GalaxySetupData::SetGameUID(const std::string& game_uid) {
+    m_game_uid = game_uid;
 }
 
 /////////////////////////////////////////////////////
--- a/util/MultiplayerCommon.cpp
+++ b/util/MultiplayerCommon.cpp
@@ -356,9 +356,8 @@ void GalaxySetupData::SetSeed(const std::string& seed) {
     m_seed = new_seed;
 }
 
-void GalaxySetupData::SetGameUID(const std::string& game_uid) {
-    m_game_uid = game_uid;
-}
+void GalaxySetupData::SetGameUID(const std::string& game_uid)
+{ m_game_uid = game_uid; }
 
 /////////////////////////////////////////////////////
 // PlayerSetupData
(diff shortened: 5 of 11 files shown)


universe/Universe.cpp
```diff
@@ -2947,6 +2947,10 @@ void Universe::UpdateStatRecords() {
     }
 }
 
+void Universe::SetGameUID(const std::string& game_uid) {
+    m_game_uid = game_uid;
```
geoffthemedio: Prefer `{ ... }` on one line for single-line functions.

server/ServerFSM.cpp
```diff
@@ -2531,7 +2531,8 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
     }
 
     if (server.IsHostless() && GetOptionsDB().Get<bool>("save.auto.hostless.enabled")) {
-        boost::filesystem::path autosave_dir_path = GetServerSaveDir() / "auto";
+        std::string subdir = GetUniverse().GameUID();
+        boost::filesystem::path autosave_dir_path = GetServerSaveDir() / (subdir.empty() ? "auto" : subdir);
```
geoffthemedio: Does this need to be in a separate directory, rather than putting the (u)uid in the filename?

o01eg: It think it's simpler to search files organized in folders than search them by name component in the one big folder.

geoffthemedio: Are restrictions on numbers of autosaves yet supported for hostless? Does putting them in separate subdirectories complicate that?

o01eg: No, hostless mode doesn't support restrictions on numbers of autosaves.

universe/Universe.cpp
o01eg: Fixed.

Vezzra:
@geoffthemedio, I'm a bit puzzled, why the classification as "component:game mechanic"? I'd thought that should be "component:internal"...

geoffthemedio:
then change it...

geoffthemedio:
> It allows to follow the game session in saved games and separate different game sessions.

Could you explain what this means or elaborate / give examples? Other than autosave sorting in hostless mode, I don't understand the point of adding the UID tracking.

o01eg:
It also could be used to save game's statistics in DBMS.

geoffthemedio:
That doesn't really answer the question or entirely make sense.

o01eg:
I mean, if we have multiple games played on one server `game1`, `game2`, etc it simplifies management of these games and their turns if there will be an entity which allows to distinguish turn 10 of `game1` and turn 10 of `game2` but allow to group all turns of `game2`.

geoffthemedio:
For that purpose, a human-readable and perhaps settable (from the lobby during game config) "game name" would be more useful and memorable?

o01eg:
That's why I choose `std::string`type instead of `boost::uuid`. UUID values were used for example and it is filled in `generate_universe.py` as script could be simpler customized.

geoffthemedio:
So there are future plans to make it user-customizable?

o01eg:
Mostly host/admin-customizable, not for players. I could change algorithm in PR to something more readable like first 2 symbols from players' names and random numeric postfix.

geoffthemedio:
Might be worthwhile. Having it be human readable and (in future) customizable when players are setting up games seems useful.

Within `Universe` doesn't seem like an ideal place for this bit of info, though... Maybe `GalaxySetupData` instead? Also not ideal, but adding in separately to SP setup and MP lobby data seems awkward...

o01eg:
I've tried to move it to `GalaxySetupData` but it doesn't work. I exposes it to python with:
```cpp
            .add_property("gameUID",            make_function(&GalaxySetupData::GetGameUID,         return_value_policy<return_by_value>()),
                                                &GalaxySetupData::SetGameUID);
```
and use it in `universe_generator.py`:
```python
    fo.get_galaxy_setup_data().gameUID = ...
    print "Game UID %s" % fo.get_galaxy_setup_data().gameUID
```
It prints value from python but empty when I access to it from c++.

o01eg:
Looks like `GalaxySetupData` is setted twice. Before `GenerateUniverse` in `NewGameInitConcurrentWithJoiners` and after in `NewGameInitVerifyJoiners`. Will it possible to remove second? It uses in `VerifySPGameAIs` so I'm not sure about how to do it.

o01eg:
Rebased and reworked. Removed second assign in `NewGameInitVerifyJoiners`. Moves `gameUID` field to `GalaxySetupData`.

util/MultiplayerCommon.cpp
```diff
@@ -352,6 +356,10 @@ void GalaxySetupData::SetSeed(const std::string& seed) {
     m_seed = new_seed;
 }
 
+void GalaxySetupData::SetGameUID(const std::string& game_uid) {
+    m_game_uid = game_uid;
```
geoffthemedio: `{ ... }` one line

geoffthemedio:
Can / should this be displayed on the pedia page about the galaxy setup data?

server/ServerApp.cpp
```diff
@@ -737,14 +737,9 @@ void ServerApp::NewGameInitConcurrentWithJoiners(
...
 }
 
-bool ServerApp::NewGameInitVerifyJoiners(
-    const GalaxySetupData& galaxy_setup_data,
-    const std::vector<PlayerSetupData>& player_setup_data)
-{
+bool ServerApp::NewGameInitVerifyJoiners(const std::vector<PlayerSetupData>& player_setup_data) {
     DebugLogger() << "ServerApp::NewGameInitVerifyJoiners";
 
-    m_galaxy_setup_data = galaxy_setup_data;
```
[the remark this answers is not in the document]
o01eg: Yep.
In multiplayer game in `NewMPGameInit` before `NewGameInitVerifyJoiners` the server always calls `NewGameInitConcurrentWithJoiners` where galaxy setup data sets.

In singleplayer game when the server enters `WaitingForSPGameJoiners` state where calls `NewSPGameInit` which calls `NewGameInitConcurrentWithJoiners` where galaxy setup data sets; then in `WaitingForSPGameJoiners.CheckStartConditions` the server calls `VerifySPGameAIs` which calls `NewGameInitVerifyJoiners`.

util/MultiplayerCommon.cpp
o01eg: Fixed.

o01eg:
@geoffthemedio 
> Can / should this be displayed on the pedia page about the galaxy setup data?

Yes, I've added this to English and Russian encyclopedia pages. How could I add it to other languages?

geoffthemedio:
> How could I add it to other languages?

Don't. Someone who knows them may do so.

Dilvish-fo:
>> How could I add it to other languages?

> Don't. Someone who knows them may do so.

I agree.  This issue is somewhat compounded by the structure of ENC_GALAXY_SETUP_SETTINGS, which right now requires translators to make a new version of the entire string, including all sub-entries. 
```
ENC_GALAXY_SETUP_SETTINGS
  '''Seed: %1%
  Number of Systems: %2%
...
AI Max Aggression: %10%
Game UID: %11%
```
 If the structure were instead something like
```
ENC_GALAXY_SETUP_SETTINGS
  '''[[GSETUP_SEED]]: %1%
  [[GSETUP_STARS]]: %2%
...
[[GSETUP_AI_AGGR]]: %10%
[[GSETUP_UID]]: %11%
```
Where all those labels already match the labels used in the galaxy setup screen (except there is not yet such a label for GSETUP_UID), then the translator's job would be simpler and any untranslated portions would automatically fall back to English.

default/python/universe_generation/universe_generator.py
```diff
@@ -144,6 +144,19 @@ def create_universe(psd_map):
...
+    # set game uid
+    empire_names = []
+    for psd in psd_map.values():
+        empire_names.append(psd.empire_name)
+
+    empire_names.sort()
+    for i, v in enumerate(empire_names):
+        empire_names[i] = v.capitalize()[:2]
+
+    fo.get_galaxy_setup_data().gameUID = "".join(empire_names) + str(random.randint(0,999)).zfill(3)
```
Cjkjvfnby, line 156: You use pseudo-random number generator that initialized outside this code. It may produce always the same number for same universe seed (now or after some changes). It looks dangerous to me.

Maybe you should create own generator and initialize it with seed that uses current timestamp. IMHO names + time with milliseconds will be unique enough.

default/python/universe_generation/universe_generator.py
```diff
@@ -144,6 +144,19 @@ def create_universe(psd_map):
     seed_rng(seed_pool.pop())
     distribute_specials(gsd.specials_frequency, fo.get_all_objects())
 
+    # set game uid
+    empire_names = []
```
Cjkjvfnby, line 148: You can do all operations in single line:
`empire_names  = sorted(x.capitalize()[:2] for x in psd_map.values())`

default/python/universe_generation/universe_generator.py
Cjkjvfnby: OR just use https://docs.python.org/2/library/uuid.html#uuid.uuid3

Cjkjvfnby (commented):
> This PR adds gameUID property to Universe class.
> It allows to follow the game session in saved games and separate different game sessions.

For single player, it is possible to have two different saves that have the same universe.UUID like commits in the different branches in git.  

Should we add UUID to game turn too?

default/stringtables/ru.txt
```diff
@@ -3793,6 +3794,7 @@ ENC_GALAXY_SETUP_SETTINGS
 Монстры: %8%
 Аборигены: %9%
 Макс. агрессивность ИИ: %10%
+Идентификатор игры: %11%
```
Cjkjvfnby, line 3797: Уникальный?  Игры или мира?  Будут ли одинаковыми идентификаторы для одинаковых миров? (Я ожидаю, что галактики сделанные из одного семени будут одинаковыми.)

default/python/universe_generation/universe_generator.py
o01eg: I'm not Python coder.

default/python/universe_generation/universe_generator.py
o01eg: UUID is hard to memorize. But thanks for info about that `random`. I'll try to change it to timestamp seeded random.

default/stringtables/ru.txt
Cjkjvfnby: Если смотреть в контексте сбора всей статистики, то уникальность текущего UUID не очень.  Я бы лучше смотрел на дату вместо трёх значеного числа.

Dilvish-fo:
> If the structure were instead something like

I agree that would be a helpful change

default/python/universe_generation/universe_generator.py
Cjkjvfnby: You write python code, so you are.  :) 

Nobody expects a single person to have all the skills,  but as the team, we have all the skills :)

default/stringtables/ru.txt
o01eg: Ну, так как это питон, его легко поправить для конкретного сервера. В случае моего сервера даже три знака, увы, избыточно.

default/python/universe_generation/universe_generator.py
Vezzra: I don't think the universe generation Python scripts are the right place to generate the UUID, that should be done in the C++ backend code.

The universe generation scripts are, like the FOCS scripts, user customizable content, and should only contain stuff that needs to be user customizable. IMO that's not the case with the UUID, so no reason to rely on user customizable scripts to properly initialize it.

What's done in the C++ code can't be broken by users tinkering with the Pytho scripts.

Vezzra: Ant it would also avoid issues caused by having to use an independent PRNG. If you reseed the Python PRNG with a timestamp derived seed here, that can/will mess up reproducible universe generation.

o01eg: I'm going to get UID from DBMS to ensure uniqueness and use it as a primary and foreign key when store games' statistics.
So I prefer to don't move DBMS support into C++ backend code.

Vezzra: Why do you want to get the UID from a DBMS? Uniqueness can be ensured by a properly implemented C++ function too, that's not something which can only be done via a DBMS.

That said, if you absolutely want to, you can still generate a UID in the universe generation Python code, and overwrite an UID that has been generated in the C++ code. But I would still have a UID generated in the C++ code, so a game gets a proper UID even if the universe generation scripts fail to do generate one.

Unless the UID only gets used in the Python scripts, and is never used/referenced in the C++ backend code. In that case generating the UID only in the Python scripts is fine of course.

o01eg: In C++ it used to determine subfolder for save game in hostless mode and also shows in galaxy setup encyclopedia for the players.

Vezzra: In that case the UID really needs to be generated in the C++ code, and should not rely on user customizable scripts to do that correctly.

We can make it so that the C++ generated UID can be overwritten in the Python scripts if so desired. Of course we have to make sure the overwritten value is written back to the C++ data structures.

o01eg: It already has getter and setter Python API so the only left is the C++ initialization.

o01eg: There is one for loading old saves: https://github.com/freeorion/freeorion/blob/master/util/SerializeMultiplayerCommon.cpp#L40

Vezzra: Provided PR moving game UID generation to C++: #2269.

Vezzra: (@o01eg, please review)

Dilvish-fo:
> I agree that would be a helpful change

I was running a bit of a fever a couple days ago, and apparently failed to notice that I was approving of my own suggestion, LOL.

PR #1887, opened 2017-11-28
Orbital outposting fixes

Morlic-fo:
There are 3 (potential, probably interacting) bugs in the current code concerning orbital expansion:

1) The AI would still have planets listed as qualifying for orbital outposting that were already colonized by another mission. The first commit in this PR aims to clean up those lingering entries at turn start.

2) The AI would consider planets for orbital outposting that were already colonization-targeted.

3) If a target that was envisioned for orbital outposting was already claimed by another mission or another player, then that base would be (silently) reassigned to another planet in the system. However, the qualifyingBaseTargets would not be updated and the new target was still "open". The entry would eventually get cleaned up but not before colonization of the new target in the next turn - there was a 1 turn gap where an additional, unneeded base would be enqueued.

The PR refactors the current code and introduces helper classes that interface the qualifyingXBaseTargets dicts in the AIstate for convenience and easier usage. In the future, all orbital colonization logic will be handled by those classes.

Fixes #1886.

PR #1998, opened 2018-02-14
Rework AI orbital colonization

Morlic-fo:
Supersedes #1887 

This PR refactors the orbital colonization from the AI and includes additional logic to make sure to stop production of bases which aren't needed anymore (same idea as in #1887).

Two new classes are introduced. The OrbitalColonizationPlan defines how a given target planet is colonized using outpost bases. The OrbitalColonizationManager manages the existing plans and decides which of the plans are executed in a given turn. The execution itself is defined in the plans.

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -7,6 +7,7 @@ import freeOrionAIInterface as fo  # pylint: disable=import-error
 from common.print_utils import Table, Text, Float
 
 import AIFleetMission
+import ColonisationAI
 import ExplorationAI
 import FleetUtilsAI
 from EnumsAI import MissionType, ShipRoleType
@@ -68,6 +69,11 @@ def convert_to_version(state, version):
     if version <= 3:
         raise ConversionError("The AI savegame version is no longer supported.")
 
+    if version == 4:
+        del state['qualifyingOutpostBaseTargets']
+        del state['qualifyingColonyBaseTargets']
+        state['orbital_colonization_manager'] = ColonisationAI.OrbitalColonizationManager()
+
     #   state["some_new_member"] = some_default_value
     #   del state["some_removed_member"]
     #   state["list_changed_to_set"] = set(state["list_changed_to_set"])
@@ -95,7 +101,7 @@ class AIstate(object):
     via boost. If desiring to store a reference to a UniverseObject store its
     object id instead; for enum values store their int conversion value.
     """
-    version = 3
+    version = 4
 
     def __init__(self, aggression):
         # Do not allow to create AIstate instances with an invalid version number.
@@ -153,8 +159,7 @@ class AIstate(object):
         self.militaryRating = 0
         self.shipCount = 4
         self.misc = {}
-        self.qualifyingColonyBaseTargets = {}
-        self.qualifyingOutpostBaseTargets = {}
+        self.orbital_colonization_manager = ColonisationAI.OrbitalColonizationManager()
         self.qualifyingTroopBaseTargets = {}
         # TODO: track on a per-empire basis
         self.__empire_standard_enemy = CombatRatingsAI.default_ship_stats().get_stats(hashable=True)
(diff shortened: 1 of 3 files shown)


default/python/AI/AIstate.py
```diff
@@ -95,7 +101,7 @@ class AIstate(object):
     via boost. If desiring to store a reference to a UniverseObject store its
     object id instead; for enum values store their int conversion value.
     """
-    version = 3
+    version = 4
```
Dilvish-fo, line 104: It looks like a version update got missed at some point in the past, and the previous one should have been 4, this one 5.  (Note that the previous one was version 3, but line 68 was at the same time rejected version 3 as no longer supported.  If this one is left as version 4, then when trying to load a savegame from this version then I expect that lines 73-74 would cause an exception.

Dilvish-fo (commented):
Thanks for redoing this.  That one comment on the version is all I have at the moment, but I'll need to come back to look at it again more carefully than I can do now, hopefully in the next day or so.

default/python/AI/ColonisationAI.py
```diff
@@ -1310,3 +1224,265 @@ def get_first_column_value(x):
                 planet.specials,
             ])
     info(candidate_table)
+
+
+class OrbitalColonizationPlan(object):
+    def __init__(self, target_id, source_id):
+        self.target = target_id
```
Cjkjvfnby: please add docstring, about target and source types  (id of which object type)

default/python/AI/ColonisationAI.py
```diff
@@ -1310,3 +1224,265 @@ def get_first_column_value(x):
...
+
+    def get_targets(self):
+        """
+        Return all planets for which an orbital colonization plan exists.
+
+        :rtype: list[int]
+        """
+        return self._colonization_plans.keys()
+
+    def delete_plan(self, target_id):
```
Cjkjvfnby: Unused method should be removed.   Please check usage of the other new methods.

Cjkjvfnby (changes requested):
I did not review whole code.

default/python/AI/ColonisationAI.py
Dilvish-fo: > Unused method should be removed. 

While lack of current use is certainly reason to think twice about whether a certain piece of code might be essentially unneeded distracting complexity, I don't see lack of current use as an automatically sufficient reason to always delete the code, not in our environment at least.  Morlic is presenting a toolset here which by itself accomplishes certain things, and he apparently contemplates the likelihood of one of us or someone else further developing more complex approaches that could include deleting an orbital construction plan.  That capability seems like a good one to be able to support.  Making a rule that we couldn't ever provide any support for more complex planning than what is currently implemented seems guaranteed to slow down development.

Cjkjvfnby: This code is never executed, so it is not tested.   
Writing unused code is waste of time. 
Reviewing unused code is waste of time.
Supporting unused code is waste of time.

When you will need new public method, you will add public method that you need.  Using method that was written 10 years ago, and did not executed since may give you a surprise.

This method is simple code and it have small chance to be broken.   But it is easy to implement it when needed.  For complex code chance that it will require changes when it will be finally used is about 100%. This number is based on my personal experience as full day developer in big project.

In terms of master, we don't remove any code, we just don't add things that not needed now.

default/python/AI/AIstate.py
Morlic-fo: The versioning should be correct but the code probably isn't as clear as it could/should be.
You seem to worry about the line
```
    if version <= 3:
        raise ConversionError("The AI savegame version is no longer supported.")
```
The variable ```version``` here is the *target* version to convert the state into (the docstring states that but the variable could be renamed for further clarity), so if the ```current_version``` of the savegame state is already 3 or higher, that line would never be executed.

I'll add revising that code to make it a little clearer to my todo list.

default/python/AI/ColonisationAI.py
Morlic-fo: The function was in use in a previous implementation stage but I removed it for now. The reimplementation would be trivial.

default/python/AI/AIstate.py
Dilvish-fo: Ah, as I had mentioned, I didn't have enough time to do a thorough review, but I should have held off from commenting on that snippet without looking at the fuller code.  Changing ```version``` in that function to ```target_version``` would be more clear, but that's pretty low priority, the biggest issue was my rushed review of just a code fragment.

PR #2001, opened 2018-02-15
Split Resign and Concede behaviour

o01eg:
Allows player to actually concede with elimination his empire but without stopping entire game:

(edit by Dilvish: [Here is the corresponding forum thread.](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10912) )


geoffthemedio:
Exit implies quitting the program. Perhaps call them "Concede" and "Resign", so the latter has the same effect as it does now.

o01eg:
@geoffthemedio Renamed button names and appropriate variables and functions names.

Dilvish-fo:
Resign and Concede are very similar (they don't seem to be listed as direct synonyms, but they share a number of synonyms).  Perhaps the buttons offered should depend on whether it is a single player game or a multiplayer game being played.  'Concede' seems meaningless for a single player game, and the significant of 'Resign' for a multiplayer game is not clear to me if not conceding (it seems that in that case it probably really is meant as an 'Exit').

geoffthemedio:
It's not "Exit" because it won't immediately quit the program.

o01eg:
May be call it "Disconnect"?

geoffthemedio:
Not having concede in a single player game makes sense... I assume this was already the case after these changes.

Resign would consistently cause the client to leave a game, single or multiplayer. Concede would only be available in multiplayer and would convert a player to an observer without disconnecting.

o01eg:
> Concede would only be available in multiplayer and would convert a player to an observer without disconnecting.

It could be undesirable disclosure of the game. I think it better to disconnect player. He could reconnect as an observer if he wants AND most important he has rights to do it.

geoffthemedio:
Just check if a player has rights when providing the concede button. No need to dis and reconnect if a player is allowed to do so...

Dilvish-fo:
> It's not "Exit" because it won't immediately quit the program.

But it could.  Why else is a person in a multiplayer game leaving the game but not conceding?

o01eg:
> If possible, would spawning a new AI to take over the empire be preferable to passing object ownership to universe?

Is it stable now? I remember different errors when trying make AI load Player empire. Also it has to check AI limits. Better to implement it in other PR.

geoffthemedio:
> Why else is a person in a multiplayer game leaving the game but not conceding?

Just leaving the game because the game session is finished...

o01eg:
Disabled Concede button in singleplayer. Changing to observer without disconnect causes some strange issues.

UI/InGameMenu.cpp
```diff
@@ -50,12 +53,17 @@ void InGameMenu::CompleteConstruction() {
...
+    m_concede_btn->LeftClickedSignal.connect(
+        boost::bind(&InGameMenu::Concede, this));
+    m_resign_btn->LeftClickedSignal.connect(
+        boost::bind(&InGameMenu::Resign, this));
     m_done_btn->LeftClickedSignal.connect(
         boost::bind(&InGameMenu::Done, this));
 
-    if (!HumanClientApp::GetApp()->SinglePlayerGame()) {
+    if (HumanClientApp::GetApp()->SinglePlayerGame()) {
+        // no sense to concede in a singlesplayer game
```
LGM-Doyle: minor spelling 
singlesplayer -> singleplayer

UI/InGameMenu.h
```diff
@@ -34,8 +35,9 @@ class InGameMenu : public CUIWnd {
     std::shared_ptr<GG::Button> m_save_btn;         //!< Save game button;
     std::shared_ptr<GG::Button> m_load_btn;         //!< Load game button;
     std::shared_ptr<GG::Button> m_options_btn;      //!< Options button;
+    std::shared_ptr<GG::Button> m_concede_btn;      //!< Concede button;
     std::shared_ptr<GG::Button> m_done_btn;         //!< Done button;
-    std::shared_ptr<GG::Button> m_exit_btn;         //!< Quit game button;
+    std::shared_ptr<GG::Button> m_resign_btn;       //!< Quit game button;
```
LGM-Doyle: Comment disagrees with the code.  Remove the comment it just repeats the variable name.

o01eg:
UI freezes when I try to go `PlayingGame` -> `WaitingForGameStart` with `GameStart` event.

default/stringtables/en.txt
```diff
@@ -995,6 +995,8 @@ Too many human players.
 ERROR_CONNECTION_WAS_REPLACED
 Your connection was replaced.
 
+ERROR_CANNOT_ELIMINATE_NO_HUMAN_PLAYER
+Cannot eliminate other than Player
```
LGM-Doyle: This error message is confusing.  Change to something like.
Only human clients can concede.

LGM-Doyle (commented):
Very good.  I added some minor comments.

LGM-Doyle:
How about replace "Resign" with "Return to Main Menu", which is unambiguous?

Why not remove the "Concede" button from the single player UI if it is meaningless?

default/stringtables/en.txt
Dilvish-fo: Actually, we were just talking about letting the AI's concede this way also when they've lost their last colony (as an interim measure while we explore more interesting ways for them to surrender).  If the point of the section of code that uses this string was to prevent AI's from conceding, then please just delete that whole section of code.

I suspect, though that the point was more to prevent an Observer or Moderator from accidentally hitting the concede button (which should probably be disabled for them anyways),

I'd suggest changing the label to "ERROR_NONPLAYER_CANNOT_CONCEDE", change the message to "Only Players can concede" and change the referencing code to accept the concession from both human clients and AI clients.

UI/InGameMenu.cpp
o01eg: Fixed.

UI/InGameMenu.h
o01eg: Removed comments.

default/stringtables/en.txt
o01eg: Changed label and allowed an AI to eliminate itself.

o01eg:
I've published commits for changing to Observer to a `resign-or-disconnect-wip` branch if someone could look why UI freezes.

o01eg:
@LGM-Doyle 
> How about replace "Resign" with "Return to Main Menu", which is unambiguous?

It looks wider that other buttons

> Why not remove the "Concede" button from the single player UI if it is meaningless?

In other cases other buttons disabled too.

wheals:
>>How about replace "Resign" with "Return to Main Menu", which is unambiguous?

>It looks wider that other buttons

Maybe "Leave Game", then? If that might be confused with exiting the program, there could be a "Close FreeOrion" (or just "Exit") button as well, I've frequently wished there was one anyway.

Dilvish-fo:
>>How about replace "Resign" with "Return to Main Menu"

>Maybe "Leave Game", then?

Perhaps I am simply in too good of a mood right now, but I would like to suggest "Transcend Galaxy" 😃 (It seems quite appropriate to me in multiple ways.)

o01eg:
So what was decided about buttons' labels?

dbenage-cx:
How about adding a 3 option dialog to Resign, instead of adding a new Concede button?
The dialog might read "Concede empire before leaving?" with options Yes, No, Cancel.
Discussion for button label can then move to forum and be renamed later once a consensus is reached.

If conceding without leaving the game is worked out, I'd suggest an option on empires/players window.
(Tried that branch, but did not determine why UI is frozen)

o01eg:
> How about adding a 3 option dialog to Resign, instead of adding a new Concede button?
> The dialog might read "Concede empire before leaving?" with options Yes, No, Cancel.

Doesn't it increase risk of accidentally eliminate own empire?

> If conceding without leaving the game is worked out, I'd suggest an option on empires/players window.
(Tried that branch, but did not determine why UI is frozen)

I suppose it better to leave it for other PR if someone could fix this issue.

o01eg:
Rebased and changed dialog for Resign.

geoffthemedio:
Conceding should have a confirmation dialog, not be an option in a confirmation dialog.

o01eg:
Removed latest changes.

geoffthemedio:
So what does this pull request do now?

o01eg:
It's look finished for me and I'm waiting for other feedbacks. At least this PR was mentioned at forum http://freeorion.org/forum/viewtopic.php?p=91836#p91836

geoffthemedio:
That's nice, but doesn't answer my question...

o01eg:
Currently it adds `Concede` button and confirmation as displayed on picture. In the game it resets empire's planets and destroys empire's ships.

geoffthemedio:
Might be better added separately after a forum discussion, but I'm wondering if allowing concession should be a rule... Having an empire's stuff just disappear is more unbalancing than having them sit without being given orders but still defending themselves and needing to be defeated in combat.

o01eg:
@geoffthemedio I thought about it but there will be issue with diplomacy. If we don't drop diplomacy states it will be possible allies couldn't capture remaining planets. If we drop diplomacy it will be possible remaining fleet in allied system will attack fleet and planets.

Dilvish-fo:
I think both points are compatible-- although there may be a variety of (existing) gameplay balance complications around not letting someone concede, it also seems there would also be gameplay balance complications around allowing concession, particularly for Human Players who might have cause to quit a game while they still have quite a few assets around, and the idea of making concession subject to a GameRule seems pretty reasonable.

An extra consideration to keep in mind is that we were really hoping to be able to use this to allow essentially-but-not-quite-fully defeated AI's to concede.  They would (at least as currently planned) only be doing so after losing all their colonies, so rather than a totally simple On/Off rule, I would propose that the Rule gate whether or not a player can concede while they still own any colonies, using that as a proxy for their likely total assets including fleets.  If they don't have much then allowing them to concede cannot make much imbalance.

o01eg:
Created discussion on forum http://freeorion.org/forum/viewtopic.php?f=6&t=10912

o01eg:
@Dilvish-fo What if I'll create two game rules? One to choose is concede allowed or not (default Off) and second to choose concede strategy.

This PR will implement current elimination strategy. If someone get with other ideas it could be implemented in other PRs as other strategy.

Dilvish-fo:
> What if I'll create two game rules? One to choose is concede allowed or not (default Off) and second to choose concede strategy.

If by "concede strategy" you mean the particular consequences of conceding, then I suppose that could fit as a game rule (but does not really match the word "strategy"), and depending on the consequence choices that might still suit our plans for the AI.

If instead, you mean the process by which an AI actually decides if it would concede (which would match the word "strategy"), then I don't think it would really be suitable as a game rule, it should just be left to AI coding.

I guess a third possibility just crossed my mind, that you might have been intending to propose a generalization of the rule I proposed above, more of a rule for "concede threshold", i.e., the maximum assets that an empire could own and be allowed to concede (beyond which it would be deemed too disruptive to allow a concession).  That would probably be ok for the AI.

Keep in mind that with respect to the AI conceding, there are two main motivations-- one is a player desire to not have to chase down every errant fleet/outpost when the AI is already soundly defeated, and another is to simplify certain parts of the AI coding that get a bit messy (and are currently still slightly buggy) when the AI no longer owns any colonies.   So we would somewhat prefer this concession to be available anytime the AI has no colonies; if we can't count on this always being available in that circumstance then we need to just go ahead and code up a special AI version of it.

geoffthemedio:
For purposes of players not needing to chase down AI stuff, that should probably be sealt with by victory conditions or achievements, not making AIs concede. The AI should just keep persuing whatever goal it has as long as a player keeps cycling turns for it.

Vezzra:
I think we should continue this discussion in the thread @o01eg created on the forum. I made my reply there.

Dilvish-fo:
>  ...that should probably be [dealt] with by victory conditions or achievements, not making AIs concede...

You seem to be missing a significant part of the prior conversation.  A victory condition might help with respect to the last AI out of 6, but not for the first 5.  (And Achievements are currently nonexistent?)  [Here is the forum thread where we have been discussing the basic design choices](http://www.freeorion.org/forum/viewtopic.php?f=2&t=10890&p=91687&sid=e2974087f7afe30c7a7ef88690b2f9ad#p91687) in case you would like to contribute there.

o01eg:
@geoffthemedio 
> For purposes of players not needing to chase down AI stuff, that should probably be sealt with by victory conditions or achievements, not making AIs concede.

Agreed, we could consider empire eliminated if it has no colonies and cann't create one so we don't need to chase scouts and warships. But it isn't for this PR.

o01eg:
Rebased. Added rules to allow/disallow concede and to limit number of colonies when concede allowed.

Changed disconnection side. Before UI disconnects itself because it could be ensured concede will be successful. Now UI only closes menu and the server tries to disconnect client if it was eliminated. This cause very strange issue and client still connected to server but cann't make turns.

o01eg:
Rebased. Fixed conflicts.

default/stringtables/en.txt
```diff
@@ -1147,6 +1152,18 @@ Gas Giants Habitable size
 RULE_HABITABLE_SIZE_DESC
 Value used to adjust for [[encyclopedia POPULATION_TITLE_SHORT_DESC]] changes on a planet of this size.
 
+RULE_ALLOW_CONCEDE
+Allow concede
```
geoffthemedio: `Allow Conceding` or `Allow Player Conceding` ?

default/stringtables/en.txt
```diff
@@ -1147,6 +1152,18 @@ Gas Giants Habitable size
...
+
+RULE_ALLOW_CONCEDE_DESC
+Allows empire to concede with elimination its assets.
```
geoffthemedio: Allows empires to concede from a game. This eliminates the empire's assets from the universe. The game continues with the remaining players.

default/stringtables/en.txt
```diff
@@ -1147,6 +1152,18 @@ Gas Giants Habitable size
...
+
+RULE_CONCEDE_COLONIES_THRESHOLD
+Maximum colonies allowed to concede
```
geoffthemedio: Maximum Colonies When Conceding

default/stringtables/en.txt
```diff
@@ -1147,6 +1152,18 @@ Gas Giants Habitable size
...
+
+RULE_CONCEDE_COLONIES_THRESHOLD_DESC
+Disallow empire with more coloies to concede.
```
geoffthemedio: Empires with more than this number of colonies may not concede.

default/stringtables/en.txt
```diff
@@ -2411,6 +2431,8 @@ Button Disabled
 SAVE_DISABLED_BROWSE_TEXT
 The Save Game button is disabled, either because this is a multiplayer game in which you are neither Host nor Moderator, or because at the time this menu was opened at least one of the AIs had not yet finished submitting its moves for the turn (indicated by a green triangle status icon just to the left of its player-type icon in the Empires window).
 
+GAME_MENU_REALLY_CONCEDE
+Are you sure you want to concede the game?
```
geoffthemedio: Confirm conceding? This empire will be removed from the game.

geoffthemedio:
Having the greyed-out Concede button in single player seems awkward. Would prefer to not have it there unless it's at least plausible it could be used in the current game.

o01eg:
Load button is greyed-out in multi player. Would it better to hide both?

o01eg:
Fixed.

geoffthemedio:
The other button didn't bother me, but you can handle it as you prefer.

geoffthemedio:
Now there is a clickable concede button in a multiplayer game with a single player. It pops up a confirm dialog, but then it says the concede is not allowed. Could it instead by disabled when not allowed, but with a tooltip indicating why?

o01eg:
@geoffthemedio 
> Could it instead by disabled when not allowed, but with a tooltip indicating why?

I'm not sure. It's server decides if it's allowed to concede. Maybe use other error text if it wasn't allowed due colony number?

geoffthemedio:
At a minimum, some explanation in the error message about why the player couldn't concede should be given...

o01eg:
@geoffthemedio Implemented.

geoffthemedio:
I don't get any explanation when conceding from a one human MP game.
one_player_mp_concede

o01eg:
This error is shown when game rule disables concede. I renamed it. Now this situation is shown as "Concede is disabled."
Also I've added "Should be other human players to concede." error if there was last human player.

default/stringtables/en.txt
```diff
@@ -1015,8 +1015,14 @@ Your connection was replaced.
 ERROR_NONPLAYER_CANNOT_CONCEDE
 Only Players can concede.
 
-ERROR_CONCEDE_NOT_ALLOWED
-Concede is not allowed.
+ERROR_CONCEDE_DISABLED
+Concede is disabled.
+
+ERROR_CONCEDE_EXCEED_COLONIES
+Your colonies number exceeds limits to concede.
```
geoffthemedio: Don't refer to the player as "you". Instead maybe "Cannot concede; empire controls too many colonies."

default/stringtables/en.txt
```diff
@@ -1015,8 +1015,14 @@ Your connection was replaced.
...
+
+ERROR_CONCEDE_LAST_HUMAN_PLAYER
+Should be other human players to concede.
```
geoffthemedio: Maybe "Cannot concede from a game with only one human player."

default/stringtables/en.txt

geoffthemedio:
If concede is disabled due to a game rule, wouldn't the human client know this, and thus be able to disable the concede button when the rules don't allow it? Still seems awkward to have the button there, attempt to use it, then get back the message saying it's disabled.

geoffthemedio:
@o01eg Not sure if this is the right pull request, but need something relevant... The options for conceding and allied winning are somewhat random scattered through the game rules tabs. Could you put them together on a separate page?

geoffthemedio:
Also, the names don't fit; make them shorter?

concedish_options

o01eg:
@geoffthemedio I propose to add tab and category named "Multiplayer".

Vezzra:
To prevent that from being forgotten in the comment section of an already merged PR I suggest to open an issue for this.

geoffthemedio:
@o01eg Sounds reasonable. Can you do it, or make an issue describing what you want to do?

o01eg:
I've add issue #2148

PR #2002, opened 2018-02-16
Add mock module for tests

Cjkjvfnby:
Don't relay on mock that generated for autocomplete for tests

### DIFF
--- a/default/python/tests/conftest.py
+++ b/default/python/tests/conftest.py
@@ -8,6 +8,11 @@ import sys
 
 this_dir = os.path.dirname(__file__)
 
+sys.path.append(this_dir)
 sys.path.append(os.path.join(this_dir, '..', '..', 'python', 'AI/'))
-sys.path.append(os.path.join(this_dir, '..', '..', 'python', 'interface_mock', 'result'))
 sys.path.append(os.path.join(this_dir, '..', '..', 'python'))
+
+# This import put module to memory, after that we can import code form AI module without problem,
+#  because module named freeOrionAIInterface is already in memory and imports in AI just use it.
+# noinspection PyUnresolvedReferences
+import freeOrionAIInterface
--- a/default/python/tests/conftest.py
+++ b/default/python/tests/conftest.py
@@ -12,7 +12,11 @@ sys.path.append(this_dir)
 sys.path.append(os.path.join(this_dir, '..', '..', 'python', 'AI/'))
 sys.path.append(os.path.join(this_dir, '..', '..', 'python'))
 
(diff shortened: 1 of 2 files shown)


Dilvish-fo:
> Don't relay on mock that generated for autocomplete for tests

Even after reading the code changes, I'm not really understanding this PR description.

As far as the code changes go, it sounds like you are thinking that the current freeOrionAIInterface mock interface provides a sufficient framework to enable it being used in place of the actual freeOrionAIInterface for testing a nontrivial portion of the various AI modules, is that right?  I am a bit skeptical that it would really enable testing much.

Could you explain your vision for this a bit more, with some specific examples of what you expect it to accomplish?

Cjkjvfnby:
To test characters we need to import module with it. 

This module have import of the `freeOrionAIInterface` https://github.com/freeorion/freeorion/blob/master/default/python/AI/character/character_module.py#L91 and use couple of the constant for it. 

Tests run in operation system, and don't have access to real module provided by the game. So I put module with same name to module cache via `import` directive. 

Before this PR this was same module that used for autocomplete and monkey patching it before use in tests.
After this PR it is separate file for tests only all monkey patching is moved to it . Dependency between tests and autocomplete code is removed, that is main goal of this PR.

default/python/tests/conftest.py
```diff
@@ -8,6 +8,11 @@
 
 this_dir = os.path.dirname(__file__)
 
+sys.path.append(this_dir)
 sys.path.append(os.path.join(this_dir, '..', '..', 'python', 'AI/'))
-sys.path.append(os.path.join(this_dir, '..', '..', 'python', 'interface_mock', 'result'))
 sys.path.append(os.path.join(this_dir, '..', '..', 'python'))
+
+# This import put module to memory, after that we can import code form AI module without problem,
```
Dilvish-fo: For better grammar and more thorough guidance I would suggest replacing these two lines with 
```
# Since the 'true' freeOrionAIInterface is not available during testing, this import loads the 'mock' freeOrionInterface
# module into memory.  For testing purposes the mock interface provides values for certain constants which various
# AI modules rely upon.  Since any later imports of the freeOrionInterface module will simply refer back the first-
# loaded module of the same name, code from the AI modules can then be imported and tested, provided that such
# AI code only relies upon constants from the freeOrionAIInterface (and specifically does not depend on any of the
# various dynamic galaxy-state queries provided by the 'true' freeOrionAIInterface).
```

At least, that better communicates my current understanding of this.

Dilvish-fo (changes requested):

Dilvish-fo (approved):

PR #2003, opened 2018-02-18
Display current value for troop parts in ship articles

dbenage-cx:
Corrects display of troops in pedia articles.
master:
fo_troop-master

PR:
fo_troop-pr


geoffthemedio:
When / why is this distinction necessary?

dbenage-cx:
With current master, viewing the pedia page for a ship containing troop parts displays `Troops: 0` on each entry.  Other code expects `Ship::TroopCapacity` to return sum of `InitialPartMeterValue`. (bit short on time atm, will update with screenshots later if needed)

dbenage-cx:
Updated with simpler change.

The "initial" value for parts meters of the temporary ships in this listing will be 0, as the meter has not gone through a turn update.

geoffthemedio:
should do the same for colony capacity

dbenage-cx:
Added

PR #2007, opened 2018-02-19
Dont describe average stockpile in species description

agrrr3:
Removes the AVERAGE_STOCKPILE_DESC from the AVERAGE_STOCKPILE species' macro

### DIFF
--- a/default/scripting/species/common/stockpile.macros
+++ b/default/scripting/species/common/stockpile.macros
@@ -8,7 +8,9 @@ BAD_STOCKPILE
 '''
 
 AVERAGE_STOCKPILE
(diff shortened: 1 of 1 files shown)


geoffthemedio:
This seems a bit overcomplicated compare to all the other species traits that only show up as average for humans...

Dilvish-fo:
> This seems a bit overcomplicated compare to all the other species traits that only show up as average for humans...

I think that's because the other traits specifications are more repetitive, writing out the effects in full form for each variant (or at least with relatively less reliance on macros), just leaving out the description for the AVERAGE variant.  agrrr3 had written the stockpile trait more concisely with a bit greater reliance on macros, which makes this correction seem complex, but I think the overall file still more clear, concise and reliably maintainable.  And I don't really see a less complex way for him to solve this, without just getting into more code/script repetition.

PR #2008, opened 2018-02-19
Fix some stat and label strings

Ouaz:
See commit message for change list.

Discussion here: http://freeorion.org/forum/viewtopic.php?f=28&t=10870&p=91663#p91663

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -7020,7 +7020,6 @@ Self-Sustaining
 SELF_SUSTAINING_SPECIES_CLASS_DESC
 List of [[encyclopedia SELF_SUSTAINING_SPECIES_TITLE]] species:
 
-
 GASEOUS_SPECIES_CLASS
 Gaseous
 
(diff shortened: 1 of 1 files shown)


default/stringtables/en.txt
```diff
@@ -14194,16 +14192,16 @@ ULTIMATE_SUPPLY_DESC
...
---	No [[metertype METER_STOCKPILE]]
+−−	No [[metertype METER_STOCKPILE]]
 
 BAD_STOCKPILE_DESC
-	Bad [[metertype METER_STOCKPILE]]: 50%
+−	Bad [[metertype METER_STOCKPILE]]: 50%
 
 AVERAGE_STOCKPILE_DESC
-	Average [[metertype METER_STOCKPILE]]: 100%
+'''	Average [[metertype METER_STOCKPILE]]: 100%'''
```
geoffthemedio, line 14201: ?

Ouaz: To make it clear there is a tab space at the beginning. The three quotes are used for each "Average [meter]" strings (e.g. AVERAGE_SUPPLY_DESC a few lines before), except this one. So I added it.

PR #2010, opened 2018-02-19
Fix DesignWnd  does not display available SavedDesigns

LGM-Doyle:
This PR fixes a bug (which will be in this week's test builds) in the the DesignWnd, which is using ids to index the saved designs and consequently not finding any available designs.


geoffthemedio:
Looks fine...

PR #2011, opened 2018-02-19
Improve DesignWnd obsolete part flicker behavior

LGM-Doyle:
In #1972 @dbenage-cx reported that un-obsoleting a design will override hull/part obsolete conditions.

The problem is that the DesignWnd does not remember the sequence of UI events involved in obsoleting a design and its hull and parts.  

This PR gives each UI obsoleting event a number so that the order can be checked and the last obsoleting instruction given by the player takes precedence.

### DIFF
--- a/util/SerializeMultiplayerCommon.cpp
+++ b/util/SerializeMultiplayerCommon.cpp
@@ -53,22 +53,72 @@ void SaveGameUIData::serialize(Archive& ar, const unsigned int version)
         & BOOST_SERIALIZATION_NVP(map_zoom_steps_in)
         & BOOST_SERIALIZATION_NVP(fleets_exploring);
 
-    if (version >= 1) {
+    if (version >= 2) {
+        ar & BOOST_SERIALIZATION_NVP(obsolete_ui_event_count);
         ar & BOOST_SERIALIZATION_NVP(ordered_ship_design_ids_and_obsolete);
         ar & BOOST_SERIALIZATION_NVP(ordered_ship_hull_and_obsolete);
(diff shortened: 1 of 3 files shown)


util/SerializeMultiplayerCommon.cpp
```diff
@@ -53,22 +53,72 @@ void SaveGameUIData::serialize(Archive& ar, const unsigned int version)
...
+
+    // Only workarounds for old versions follow
+    if (version >= 2)
+        return;
+
+    if (!Archive::is_loading::value)
+        return;
+
+    if (version == 1) {
+        obsolete_ui_event_count = 1;
```
geoffthemedio, line 77: this is pretty long to keep in the serialize function... maybe move to a separate function?

geoffthemedio:
I find this whole system / concept overly complicated, if I understand vaguely correctly, with parts being obsolete separately from designs and one overriding the other... It would be sufficient to just have a given design be obsolete or not, and a way to mark all designs that use a part as obsolete, and unobsolete individual designs as well. But if people actually find this useful... then OK, as it can all just be ignored by the player if not wanted.

LGM-Doyle:
It is a set of features that are invisible and unnecessary if a player is not looking for the features.  However, it is close to features requested  in the forum.

util/SerializeMultiplayerCommon.cpp
LGM-Doyle: I moved the legacy changes into their own function.

geoffthemedio:
no further comments...

PR #2012, opened 2018-02-19
Improve DesignWnd, allow rename in MainPanel

LGM-Doyle:
This PR should be applied after #2011 

This PR allows renaming finished and saved designs directly from the MainPanel.

This partially address inconsistencies with renaming in the DesignWnd.  It does not fix the problem that renaming from the context menu only works on the turn that a design is created.

### DIFF
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -3793,8 +3793,8 @@ public:
     boost::optional<int>                GetReplacedDesignID() const;//!< returns ID of completed design selected to be replaced.
 
     /** If a design with the same hull and parts is registered with the empire then return the
-        design name, otherwise return boost::none. */
-    boost::optional<std::string>        CurrentDesignIsRegistered();
+        design, otherwise return boost::none. */
+    boost::optional<const ShipDesign*>        CurrentDesignIsRegistered();
     //@}
 
     /** \name Mutators */ //@{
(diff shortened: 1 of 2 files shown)


LGM-Doyle:
rebased

geoffthemedio:
haven't tested, but no objections to code after a skim...

geoffthemedio:
@LGM-Doyle If you're still around, will you have time to rebase this again soon?

geoffthemedio:
Rebased, fixed conflicts and committed: https://github.com/freeorion/freeorion/commit/cc22ba441c27ad2d1ea1ef18ec1665af4686bf4d

Vezzra:
Ugh, I'm a bit at a loss here how to tag/label/categorize this PR now. Strictly technically it has been superseded by cc22ba441c27ad2d1ea1ef18ec1665af4686bf4d. However, superseded PRs get assigned to the "Gateway to the Void" milestone to imply that they themselves didn't get merged, but something else in their place.

On the other hand, cc22ba441c27ad2d1ea1ef18ec1665af4686bf4d is just a rebased version of this PR (with conflicts resolved), so *actually* it would probably be more correct to say this PR has been merged. And in that case it should be (or stay) assigned to the next release milestone (which would be advisable since we can't assign the superseding commit).

We can of course keep the label "superseded", and still leave the PR assigned to the release milestone, but that is a potential source of confusion if, several months from now, someone checks the history, wonders why a superseded PR isn't assigned to "Gateway to the Void", and decides to rectify that "error". Making such exceptions to rules is always a problematic thing...

Any suggestions how to handle this?

geoffthemedio:
If another PR had been made with the rebased commits, and that one merged, what would you have labelled this one? Do the same, I suggest.

Vezzra:
Well, in that case I'd have another PR which I can assign to the relevant milestone. I can't do that with a superseding commit, so the relation of what has been merged to the proper milestone is lost (unless I keep the superseded PR assigned to the relevant milestone, which has it's own issues as explained above).

Not exactly a big deal, I just wondered if there was a way to handle it better...

Dilvish-fo:
It seems to me that the current status is most similar to it being as if this had been squash-merged, and that the most fitting label and milestone would be "merged" and "Next Release".

Vezzra:
Sounds reasonable.

PR #2013, opened 2018-02-19
Include m_design_by_empire in ShipDesign serialization.

LGM-Doyle:
The missing m_designed_by_empire caused the bug that designs could only
be renamed on the same turn that they were designed, since their
ownership was lost on serialization.


geoffthemedio:
don't really need a pull request for something like this...

LGM-Doyle:
I made the PR in case you wanted to comment on the clutter of versioning the ShipDesign serialization for a single line bug-fix.

PR #2014, opened 2018-02-20
Show PP waste warning if stockpile input is unusually high

agrrr3:
[forum discussion](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10870&p=91705#p91702)

* Show industry waste warning in MapWnd if stockpile input is 3 times as high as capacity
** i.e. the warning is shown if it takes at least 4 turns to extract the input


geoffthemedio:
Seems to work. Also seems like there needs to be an indication in one or more of the production or stockpile tooltips that indicate the max stockpile per-turn usage.

Dilvish-fo:
> Also seems like there needs to be an indication in one or more of the production or stockpile tooltips that indicate the max stockpile per-turn usage.

I agree, having to open up the ProductionWnd to see that info is a little bit of a nuisance.

Issue #2015, opened 2018-02-20
planet troops regen even during combat


Dilvish-fo:
[As noted on the forums](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10878), for over a week now planetary troops have been getting regen even when the planet is under attack.  I thought we had likely sorted out the issue with some corrections to the scripts, but that does not appear to be the case.  I have seen various levels of regen (and also with planets using Defense Focus or not); it does appear to coincide with the "regular" non-combat regen level.  The various Garrison tech scripts themselves all are gated by 
```
(LocalCandidate.System.LastTurnBattleHere < CurrentTurn)
```

Which does appear correct.  Just to check I tried swapping in ```LastTurnAttackedByShip``` in place of  ```LastTurnBattleHere``` but it made no difference.

geoffthemedio:
Can you provide a save demonstrating the issue?

Dilvish-fo:
Happening both at Nidor and at Gienah Beta.
[wrong_troop_regen.zip](https://github.com/freeorion/freeorion/files/1742067/wrong_troop_regen.zip)

Dilvish-fo:
minor update: using Debug Data Dumps, I verified that LastTurnAttackedByShip (for the planet) and LastTurnBattleHere (for the system) are being updated properly.

geoffthemedio:
That should be available in the objects list columns as well...

https://github.com/freeorion/freeorion/blob/master/UI/ObjectListWnd.cpp#L190

geoffthemedio:
Possibly `LocalCandidate` isn't being populated as expected when evaluating conditions that include `(LocalCandidate.System.LastTurnBattleHere < CurrentTurn)`. Then when it tries to evaluate the left part of that, it returns -65535, or the "invalid game turn" value.

Dilvish-fo:
>  Then when it tries to evaluate the left part of that, it returns -65535, or the "invalid game turn" value.

I had already done a test on that yesterday, with augmented logging for the path where that returns the invalid game turn value, and didn't get any hits from it...

Dilvish-fo:
hmm, although it was also yesterday that I discovered that sometimes, not always, when I rebuild FO my logging options return to the default "Debug" on everything, I don't recall the relative timing of that discovery with my check for the invalid turn, so I'll repeat that check.

geoffthemedio:
[...]

PR #2018, opened 2018-02-21
Augment effect tracing

Dilvish-fo:
As noted in #2015 I am finding this augmented logging to be pretty helpful in trying to debug Effects

### DIFF
--- a/universe/Effect.h
+++ b/universe/Effect.h
@@ -44,7 +44,8 @@ public:
                  std::vector<std::unique_ptr<EffectBase>>&& effects,
                  const std::string& accounting_label = "",
                  const std::string& stacking_group = "", int priority = 0,
-                 const std::string& description = "");
+                 const std::string& description = "",
+                 const std::string& content_name = "");
     virtual ~EffectsGroup();
 
     /** execute all effects in group */
--- a/universe/Effect.h
+++ b/universe/Effect.h
@@ -80,7 +80,7 @@ protected:
     std::string                 m_accounting_label;
     int                         m_priority;
     std::string                 m_description;
-    std::string	                m_content_name;
+    std::string                 m_content_name;
 
 private:
     friend class boost::serialization::access;
--- a/universe/Effect.h
+++ b/universe/Effect.h
@@ -68,6 +68,7 @@ public:
     bool                            HasSitrepEffects() const;
 
     void                            SetTopLevelContent(const std::string& content_name);
+    const std::string&              TopLevelContent() const { return m_content_name; }
 
     virtual unsigned int            GetCheckSum() const;
 
@@ -79,6 +80,7 @@ protected:
     std::string                 m_accounting_label;
     int                         m_priority;
     std::string                 m_description;
+    std::string	                m_content_name;
 
 private:
     friend class boost::serialization::access;
@@ -1297,7 +1299,8 @@ void EffectsGroup::serialize(Archive& ar, const unsigned int version)
         & BOOST_SERIALIZATION_NVP(m_activation)
         & BOOST_SERIALIZATION_NVP(m_stacking_group)
         & BOOST_SERIALIZATION_NVP(m_effects)
-        & BOOST_SERIALIZATION_NVP(m_description);
+        & BOOST_SERIALIZATION_NVP(m_description)
+        & BOOST_SERIALIZATION_NVP(m_content_name);
 }
 
 template <class Archive>
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1599,7 +1599,8 @@ void Universe::ExecuteEffects(const Effect::TargetsCauses& targets_causes,
             if (group_targets_causes.empty())
                 continue;
 
-            TraceLogger(effects) << "\n\n * * * * * * * * * * * (new effects group log entry)";
+            TraceLogger(effects) << "\n\n * * * * * * * * * * * (new effects group log entry)(" 
+                                 << effects_group->AccountingLabel() << ")";
 
             // execute Effects in the EffectsGroup
             effects_group->Execute(group_targets_causes,
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -945,7 +945,8 @@ namespace {
             ConditionCache&                    cached_condition_matches,
             std::shared_ptr<const UniverseObject> source,
             const ScriptingContext&            source_context,
-            Effect::TargetSet&                 target_objects);
+            Effect::TargetSet&                 target_objects,
+			std::string&					   match_log);
     };
 
     StoreTargetsAndCausesOfEffectsGroupsWorkItem::StoreTargetsAndCausesOfEffectsGroupsWorkItem(
@@ -1031,7 +1032,8 @@ namespace {
         StoreTargetsAndCausesOfEffectsGroupsWorkItem::ConditionCache&   cached_condition_matches,
         std::shared_ptr<const UniverseObject>                           source,
         const ScriptingContext&                                         source_context,
-        Effect::TargetSet&                                              target_objects)
+        Effect::TargetSet&                                              target_objects,
+		std::string&													match_log)
     {
         std::pair<bool, Effect::TargetSet>* cache_entry = nullptr;
 
@@ -1068,18 +1070,19 @@ namespace {
             target_objects.insert(target_objects.end(), target_set->begin(), target_set->end());
         }
 
+        // log condition scope matches, except for Source
+        if (cond->Description() != " that is the source object") {  // TODO: derive this comparison text
+            std::stringstream ss;
+            ss << "\nGenerated new target set, for Condition: " << cond->Description() << "    targets: (";
+            for (const auto& obj : *target_set)
+                ss << obj->Name() << " (" << std::to_string(obj->ID()) << ")  ";
+            ss << ")";
+            match_log.append(ss.str());
+        }
+
         cached_condition_matches.MarkComplete(cache_entry);
 
-        // log condition scope matches
-        TraceLogger(effects) << "Generated new target set, for Condition: " << cond->Description();
-        std::stringstream ss;
-        ss << "    targets: (";
-        for (const auto& obj : *target_set)
-            ss << obj->Name() << " (" << std::to_string(obj->ID()) << ")  ";
-        ss << ")";
-        TraceLogger(effects) << ss.str();
-
-        return *target_set; 
+        return *target_set;
     }
 
     std::string StoreTargetsAndCausesOfEffectsGroupsWorkItem::GenerateReport() const {
@@ -1099,15 +1102,17 @@ namespace {
     {
         ScopedTimer timer("StoreTargetsAndCausesOfEffectsGroups");
 
-        TraceLogger(effects) << GenerateReport();
 
         // get objects matched by scope
         const Condition::ConditionBase* scope = m_effects_group->Scope();
-        if (!scope)
+        if (!scope) {
+            TraceLogger(effects) << GenerateReport();
             return;
+        }
 
         // create temporary container for concurrent work
         Effect::TargetSet target_objects(*m_target_objects);
+        std::string match_log;
         // process all sources in set provided
         for (auto& source : *m_sources) {
             ScriptingContext source_context(source);
@@ -1132,7 +1137,8 @@ namespace {
                                                                 *condition_cache,
                                                                 source,
                                                                 source_context,
-                                                                target_objects);
+                                                                target_objects,
+                                                                match_log);
             {
                 boost::shared_lock<boost::shared_mutex> cache_guard;
 
@@ -1142,6 +1148,8 @@ namespace {
             }
 
             {
+                TraceLogger(effects) << GenerateReport() << match_log;
+
                 // NOTE: std::shared_ptr copying is not thread-safe.
                 // FIXME: use std::shared_ptr here, or a dedicated lock
                 boost::unique_lock<boost::shared_mutex> guard(*m_global_mutex);
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1073,7 +1073,7 @@ namespace {
         // log condition scope matches, except for Source
         if (!(dynamic_cast<const Condition::Source*>(cond))) {
             std::stringstream ss;
-            ss << "\nGenerated new target set, for Condition: " << cond->Description() << "    targets: (";
+            ss << "\nGenerated new target set, for Condition: " << cond->Dump() << "\n    targets: (";
             for (const auto& obj : *target_set)
                 ss << obj->Name() << " (" << std::to_string(obj->ID()) << ")  ";
             ss << ")";
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1070,7 +1070,15 @@ namespace {
 
         cached_condition_matches.MarkComplete(cache_entry);
 
-        //DebugLogger() << "Generated new target set!";
+        // log condition scope matches
+        TraceLogger(effects) << "Generated new target set, for Condition: " << cond->Description();
+        std::stringstream ss;
+        ss << "    targets: (";
+        for (const auto& obj : *target_set)
+            ss << obj->Name() << " (" << std::to_string(obj->ID()) << ")  ";
+        ss << ")";
+        TraceLogger(effects) << ss.str();
+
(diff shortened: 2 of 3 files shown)


universe/Universe.cpp
```diff
@@ -945,7 +945,8 @@ namespace {
             ConditionCache&                    cached_condition_matches,
             std::shared_ptr<const UniverseObject> source,
             const ScriptingContext&            source_context,
-            Effect::TargetSet&                 target_objects);
+            Effect::TargetSet&                 target_objects,
+			std::string&					   match_log);
```
geoffthemedio: replace tabs with 4 spaces

dbenage-cx:
Not seeing it right off ~, so will update after `make clean`.~ (still occurs)
On new/load game start: `Error in "/opt/projects/freeorion/build/freeorionca": double free or corruption (fasttop): 0x00007fc124314630`
[stacktrace](https://gist.github.com/dbenage-cx/67ddbdce645f270a67e5a3c7102d920c)

Dilvish-fo:
> replace tabs with 4 spaces

OK, in Visual Studio I've got tab size set to 4 now and "Insert Spaces" (sorry I didn't think to check that when I started using this).  What do you have for the setting right above that, about indenting "None", "Block", or "Smart"?  Any other Visual Studio settings that would be particularly good for me to adjust?

I've cleaned up the tabs for spaces in the commits & force pushed a new set.

Dilvish-fo:
I fixed a missing space in that last commit and force pushed it again....

geoffthemedio:
I have indenting on smart, but I don't recall specifically choosing that setting.

universe/Effect.h
```diff
@@ -79,6 +80,7 @@ class FO_COMMON_API EffectsGroup {
     std::string                 m_accounting_label;
     int                         m_priority;
     std::string                 m_description;
+    std::string					m_content_name;
```
geoffthemedio: tabs

universe/Effect.h
```diff
@@ -1298,6 +1300,7 @@ void EffectsGroup::serialize(Archive& ar, const unsigned int version)
         & BOOST_SERIALIZATION_NVP(m_stacking_group)
         & BOOST_SERIALIZATION_NVP(m_effects)
         & BOOST_SERIALIZATION_NVP(m_description);
+        & BOOST_SERIALIZATION_NVP(m_content_name);
```
geoffthemedio: remove `;` on previous line

universe/Universe.cpp
geoffthemedio: still tabs. whatever pushing you've done didn't work

universe/Universe.cpp
```diff
@@ -1068,10 +1070,19 @@ namespace {
             target_objects.insert(target_objects.end(), target_set->begin(), target_set->end());
         }
 
+        // log condition scope matches, except for Source
+        if (cond->Description() != " that is the source object") {  // TODO: derive this comparison text
```
geoffthemedio: could `dynamic_cast` to a pointer to a `Source` condition, and check that the cast worked, rather than testing text

Dilvish-fo:
I replaced (force push) the commit that still had the bad tabs & the bad semicolon, and also changed that Source check to use a dynamic cast as you suggested.  The polarity of that check is inverted from what we normally do, but it seems to work fine.

universe/Effect.h
```diff
@@ -79,6 +80,7 @@ class FO_COMMON_API EffectsGroup {
     std::string                 m_accounting_label;
     int                         m_priority;
     std::string                 m_description;
+    std::string	                m_content_name;
```
geoffthemedio: tab

geoffthemedio: Just do a find and replace... don't care if an intermediate commit has a tab if the final pull request changes are clean.

universe/Universe.cpp
geoffthemedio: still tab

universe/Effect.h
Dilvish-fo: > Just do a find and replace

Ah, OK, I just found the control to let me use regular expressions so I could enter a tab into the find box.   Using left and right arrows to navigate across that line they had all seemed like spaces. Still learning my way around VS, sorry...

geoffthemedio: I think you can just copy paste a tab character into the find box as well.

Dilvish-fo:
@dbenage-cx please double check that execution problem again now.

dbenage-cx:
Looks like this may be due to some missing stringtable keys, following prevents crash for me:
```diff
diff --git a/universe/ValueRef.cpp b/universe/ValueRef.cpp
index b78b6b07e..2a10cf48e 100644
--- a/universe/ValueRef.cpp
+++ b/universe/ValueRef.cpp
@@ -468,7 +468,10 @@ std::string FormatedDescriptionPropertyNames(
         if (property_name_part.empty())  // apparently is empty for a EFFECT_TARGET_VALUE_REFERENCE
             continue;
         std::string stringtable_key("DESC_VAR_" + boost::to_upper_copy(property_name_part));
-        formatter % UserString(stringtable_key);
+        if (UserStringExists(stringtable_key))
+            formatter % UserString(stringtable_key);
+        else
+            ErrorLogger() << "Missing stringtable key: " << stringtable_key;
     }
 
     std::string retval = boost::io::str(formatter);
```

geoffthemedio:
Are you getting this after a full clean rebuild? FlexibleFormat shouldn't be crashing regardless of what is passed to it...

dbenage-cx:
@geoffthemedio Not tried clean build since changes, starting one now but may not be back to check till late.

dbenage-cx:
No change from clean build without above diff (still crashes)

universe/Universe.cpp
```diff
@@ -1068,10 +1070,19 @@ namespace {
             target_objects.insert(target_objects.end(), target_set->begin(), target_set->end());
         }
 
+        // log condition scope matches, except for Source
+        if (!(dynamic_cast<const Condition::Source*>(cond))) {
+            std::stringstream ss;
+            ss << "\nGenerated new target set, for Condition: " << cond->Description() << "    targets: (";
```
geoffthemedio: Since this is being logged, rather than shown to the player, this should probably call `cond->Dump(...)` rather than `Description`. Also, why skip the output for `Source` conditions? They're probably almost the shortest, and always having it there makes it easier to serach / visually parse.

Dilvish-fo:
> Also, why skip the output for Source conditions? They're probably almost the shortest, and always having it there makes it easier to serach / visually parse.

```
17:08:41.432064 [trace] effects : Universe.cpp:1152 : StoreTargetsAndCausesOfEffectsGroups: effects_group:   specific_cause: DEF_GARRISON_2  sources: Glorious Sceptrum I (9086)  )
Generated new target set, for Condition:  that is a Planet and that belongs to source's owner empire and that has Population of at least 0.010000    targets: (Minkar γ I (868)  Minkar γ II (882)  Minkar γ Asteroids (896)  Gienah β I (3150)  Bacabs α III (6272)  Tyr I (8358)  Bacabs α V (6300)  Bacabs α I (6314)  Marfik α I (7980)  Phaet α I (6384)  Aten β I (7196)  Glorious Sceptrum I (9086)  )
```
Note that the source(s) is/are already reported on the first line of each EffectsGroup entry, so I don't think the extra reporting of Source matches would actually add any info.  (sidenote: I can't remember for sure under what circumstances it might group multiple sources there, but it is my recollection that there are some.)  Also, when I initially had the Source matches being logged as extra lines it seemed to have a large enough volume (or perhaps it was specifically the increase in number of lines) to noticeably slow down my computer's ability to search through the log.   

As for searching, initially I had tried searching by source and that worked passably well and can still be done, but what really seemed to make a huge improvement was having the content name added to the output and being able to search with that, then look at the source in that line, then the next line after each match in that logging section contains the description of the scope for the particular EffectsGroup followed by the matching target set.

> Since this is being logged, rather than shown to the player, this should probably call ```cond->Dump(...)``` rather than ```Description```

I certainly grant that would follow the general guideline we have about the respective primary uses of the two methods.  Keep in mind though that ```Dump``` output generally contains multiple newline characters which break up the logging output greatly, whereas Description output generally does not.   I personally find this streamlined grouping easy/nice to review, and would be concerned about the extra lines possibly bogging down my file editor.  But that all might just be a combination of pure subjective preference and speculative worry; if you still think you have a strong preference for the consistency of using ```Dump()``` here then I'll certainly change it.  (I could always just change it back for my own debugging work).

Dilvish-fo:
@dbenage-cx  I suppose changing from ```Description()``` to ```Dump()``` would also solve your crash, but regardless it is seeming worthwhile to better figure out just why you are getting that crash.  Could you perhaps reorganize your change to simply add 
```
+        if (!UserStringExists(stringtable_key))
+            ErrorLogger() << "Missing stringtable key: " << stringtable_key;
```
before the call to ```formatter % UserString(stringtable_key);``` (and perhaps needing something to flush the logger?), so that we could see for just what key X the call to ```formatter % ("ERROR: " + X)``` is causing a crash?

dbenage-cx:
```diff
diff --git a/universe/ValueRef.cpp b/universe/ValueRef.cpp
index b78b6b07e..42d637d60 100644
--- a/universe/ValueRef.cpp
+++ b/universe/ValueRef.cpp
@@ -468,6 +468,8 @@ std::string FormatedDescriptionPropertyNames(
         if (property_name_part.empty())  // apparently is empty for a EFFECT_TARGET_VALUE_REFERENCE
             continue;
         std::string stringtable_key("DESC_VAR_" + boost::to_upper_copy(property_name_part));
+        if (!UserStringExists(stringtable_key))
+            ErrorLogger() << "Missing key: " << stringtable_key;
         formatter % UserString(stringtable_key);
     }
```
[log files](https://github.com/freeorion/freeorion/files/1783194/freeorion_2018.zip) (ai/python @ WARN)
I do not see any additional keys noted with the altered diff.

Just had a similar crash with freeoriond instead, will build another client and try to replicate with headless since I can attach debugger there.

Dilvish-fo:
@dbenage-cx hmm, and in the run that those log files are from, the client crashed?  freeorion.log ends abruptly after one of the ```ValueRef.cpp:472 : Missing key: DESC_VAR_HAPPINESS``` lines, but that same key was reported multiple times before without a crash.  Does the logger maybe need to get flushed to make sure we're not missing some important lines there, or does it flush itself after each line logged?

dbenage-cx:
In the last posted log files, the client crashed with the stacktrace noting freeorionca.
Playing with this further I have had the same crash occur in different clients, in some instances a single AI client crashes and the human client returns to intro screen. Rarely it does not even occur on game start, but may on a subsequent turn.

In this post, crash is in freeorion (human client). The setup:
* $`build-host/freeoriond`
* $`gdb --args build-mp-1/freeorion --network.server.external.force 1`
* $`r`
* select multiplayer -> host -> add AI (x1)
* start

Excerpts from [backtrace](https://gist.github.com/dbenage-cx/2d9059cbb1a7bab8b78bc5763c89be68)
```
#5 0x00007fcd87717ce0 in StringTable_::operator[] (this=0x1ebb280, index="DESC_VAR_HAPPINESS") at /opt/projects/freeorion/util/StringTable.cpp:39
error_retval = "ERROR: DESC_VAR_HAPPINESS"
it = {first = <error: Cannot access memory at address 0xfdc>, second = ""}

#8 0x00007fcd8741c510 in ValueRef::FormatedDescriptionPropertyNames (ref_type=ValueRef::SOURCE_REFERENCE, property_names=std::vector of length 2, capacity 2 = {...}, return_immediate_value=false) at /opt/projects/freeorion/universe/ValueRef.cpp:473
stringtable_key = "DESC_VAR_HAPPINESS"
property_name_part = "Happiness"
__for_range = std::vector of length 2, capacity 2 = {"Planet", "Happiness"}
__for_begin = "Happiness"
__for_end = "\240\262\067d\315\177\000\000\065\000\000\000\000\000\000\000\000\257\067d\315\177\000\000\340\252\067d\315\177\000\000\020\260\067d\315\177\000\000\060\262\067d\315\177\000\000X\262\067", <incomplete sequence \315>
retval = "DESC_VAR_HAPPINESS"
num_references = 2
format_string = "%1%'s %2%'s %3%"
formatter = {items_ = std::vector of length 3, capacity 3 = {{argN_ = 0, res_ = "source", appendix_ = "'s ", fmtstate_ = {width_ = 0, precision_ = 6, fill_ = 32 ' ', flags_ = 4098, rdstate_ = std::_S_goodbit, exceptions_ = std::_S_goodbit, loc_ = {<boost::optional_detail::optional_base<std::locale>> = {<boost::optional_detail::optional_tag> = {<No data fields>}, m_initialized = false, m_storage = {dummy_ = {data = "h\220\377;\315\177\000", aligner_ = {<No data fields>}}}}, <No data fields>}}, truncate_ = 9223372036854775807, pad_scheme_ = 0}, {argN_ = 1, res_ = "planet", appendix_ = "'s ", fmtstate_ = {width_ = 0, precision_ = 6, fill_ = 32 ' ', flags_ = 4098, rdstate_ = std::_S_goodbit, exceptions_ = std::_S_goodbit, loc_ = {<boost::optional_detail::optional_base<std::locale>> = {<boost::optional_detail::optional_tag> = {<No data fields>}, m_initialized = false, m_storage = {dummy_ = {data = "P\347\377;\315\177\000", aligner_ = {<No data fields>}}}}, <No data fields>}}, truncate_ = 9223372036854775807, pad_scheme_ = 0}, {argN_ = 2, res_ = "", appendix_ = "", fmtstate_ = {width_ = 0, precision_ = 6, fill_ = 32 ' ', flags_ = 4098, rdstate_ = std::_S_goodbit, exceptions_ = std::_S_goodbit, loc_ = {<boost::optional_detail::optional_base<std::locale>> = {<boost::optional_detail::optional_tag> = {<No data fields>}, m_initialized = false, m_storage = {dummy_ = {data = "er_gramm", aligner_ = {<No data fields>}}}}, <No data fields>}}, truncate_ = 9223372036854775807, pad_scheme_ = 0}}, bound_ = std::vector<bool> of length 0, capacity 0, style_ = 1, cur_arg_ = 2, num_args_ = 3, dumped_ = false, prefix_ = "", exceptions_ = 0 '\000', buf_ = {<std::basic_streambuf<char, std::char_traits<char> >> = {<No data fields>}, putend_ = 0x7fcd3430bba6 "", is_allocated_ = true, mode_ = 24, alloc_ = {<__gnu_cxx::new_allocator<char>> = {<No data fields>}, <No data fields>}}, loc_ = {<boost::optional_detail::optional_base<std::locale>> = {<boost::optional_detail::optional_tag> = {<No data fields>}, m_initialized = false, m_storage = {dummy_ = {data = "\000�\2009T\355", <incomplete sequence \306>, aligner_ = {<No data fields>}}}}, <No data fields>}}
```

seems some output issues, __for_end should probably read:
```__for_end = <error reading variable: Cannot create a lazy string with address 0x0, and a non-zero length.>```

Dilvish-fo:
@dbenage-cx  Hmm, just to try getting another possible clue, would you mind trying that again with all the threading options set to single-threaded?

dbenage-cx:
@Dilvish-fo Good call, have not been able to reproduce when adding `--effects.ai.threads 1 --effects.server.threads 1 --effects.ui.threads 1` to both execs.  Will experiment with these tomorrow to see which narrows it down.

geoffthemedio:
Might be necessary to add threaded access guards / mutex locks to the stringtable...

Edit: actually, given how https://github.com/freeorion/freeorion/blob/master/util/StringTable.cpp#L36 is implemented, it may not be possible to make the stringtable threadsafe. It's returning by-reference an `std::string` that can be modified by another stringtable lookup, which would invalidate previously-returned such references held by other threads.

That said, how it works now is surprisingly fragile even a non-threaded context, so should probably be changed. That said, maybe don't use stringtable lookups in logging code? If it needs to be single-line, you could remove the newlines before outputting.

Dilvish-fo:
@dbenage-cx perhaps check if #2039 lets you now safely run this multi-threaded?

Even if it does, I think it would probably be best for me to shift these Description() calls to Dump() calls to avoid getting bogged down by the stringtable access, but in the meantime this seems like a useful test for 2039.

dbenage-cx:
After rebasing this PR with 2039 (9b5a070), I've not been able to reproduce any of the previous crashes.

PR #2019, opened 2018-02-22
Ai invasion adjustments

Dilvish-fo:
Since the AI currently has a harder time than a human player dealing with the server glitch that is allowing troop regen even for a planet under blockade, I thought I'd put this up for active testers to be able to have their AIs put up a little better fight.  

The first commit is one actually suitable to commit to master, so after discussion and any changes I may push it directly to master.  

The second is just a stopgap to help the AI deal with the current troop regen glitch, and is not something I'd really want to push to master even temporarily.  But any active testers would probably (I hope) have more enjoyable games for the duration of this server glitch if they cherrypick both of these commits.


Morlic-fo (approved):
First commit looks good.

Dilvish-fo:
I've removed the second commit now since the troop regen issue seems to be fixed.  I guess I'll go ahead and merge this, will likely make a superficial conflict for the initial/current meters PR, but that still needs a lot of file editing anyways and this conflict should be  easy to resolve.

PR #2022, opened 2018-02-25
Run python linter on travis

Cjkjvfnby:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,18 @@ matrix:
       before_cache: echo ""
       env: NAME="Pytest"
     - os: linux
+      language: python
+      python: 2.7.14
+      install:
+        - pip install flake8-putty==0.4.0
+      script:
+        - |
+          cd default/python
+          flake8
+      before_script: echo "Code style check https://github.com/freeorion/freeorion/tree/master/default/python#code-style-check"
+      before_cache: echo ""
+      env: NAME="Python linter"
+    - os: linux
       services:
         - docker
       env:
--- a/default/python/README.md
+++ b/default/python/README.md
@@ -24,6 +24,9 @@ planet types, planet sizes, and also other content get placed.
 
 # Code style check
 
+Each PR will be checked automatically, but you still can run checks manually.
+Fixing all issues are mandatory before merging PR.
+
 We use [flake8-putty](https://pypi.python.org/pypi/flake8-putty)
 which include [flake8](https://pypi.python.org/pypi/flake8)
 which include [pycodestyle](https://pypi.python.org/pypi/pycodestyle) and other tools.


Dilvish-fo:
Well, it seems to me that this shows substantial worth right away, in that it identified a substantive error from a recent commit (when cap was renamed).

I assume the idea is that going forward we would generally not merge PR's with errors reported in this Travis testing, and in that case it seems to me that this PR should go ahead and provide the fixes for the currently reported errors/warnings.

adrianbroher (approved):
No complaints from my side.

Vezzra:
Note: "Infrastructure" PRs don't require a milestone.

Cjkjvfnby:
> I assume the idea is that going forward we would generally not merge PR's with errors reported in this Travis testing, and in that case it seems to me that this PR should go ahead and provide the fixes for the currently reported errors/warnings.

Yes, I will fix warnings in separate PR. 

Can you create topic on forum with explanation?   I will add `echo <forum link>` to build, so user will know what to do if build failed.

Dilvish-fo:
> Can you create topic on forum with explanation? 

OK, here it is:
http://www.freeorion.org/forum/viewtopic.php?f=9&t=10900

As noted there, suggestions for improving that initial message are invited.

adrianbroher:
There is the CONTRIBUTE.md file. Stop relying on external infrastructure.

Dilvish-fo:
> There is the CONTRIBUTE.md file. Stop relying on external infrastructure.

@adrianbroher Your disagreement is clear but your guidance is not.   Are you objecting to the whole forum post about this new aspect of our process?  Are you meaning "Please add most of that forum post into a new section of [CONTRIBUTING.md](https://github.com/freeorion/freeorion/blob/master/CONTRIBUTING.md)" and then have the Travis error process give a link to that instead of to the forum post?  Or are you thinking only the latter bit about the link, and that the existing CONTRIBUTING.md reference to PEP8 is fully sufficient, no need or benefit to giving tips about IDE etc?  Or the tips about the IDE etc., are fine and good in the forum post, but not in CONTRIBUTING.md?

Dilvish-fo:
>  and some advertisement for an IDE a contributor doesn't care about because he already made a choice about his preferred tools.

Please get your inner grouch in check, hehe.  Many folks are open to learning about which free tools provide the best feature set and can objectively assess the information despite a previous decision they had made.  Pycharm has been the clear leader in that front for a while now, though it sounds like they have a serious contender in Wing (though I don't yet know enough about the free Wing version to say very much about them).  We would be doing contributors and the project a disservice if we simply let everyone bask in their ignorance.

Cjkjvfnby:
I have echoed link to code-style document into Travis run.  I don't think that link to `contribute.md` that links to code style document will be more useful.

I add notice about CI `contribute.md`, feel free extract this to separate PR.

Can I include forum link to code style document?

geoffthemedio:
Are these contributing guidelines discussions relevant to merging this pull request? Seemed to have no specific objections...

Dilvish-fo:
> Are these contributing guidelines discussions relevant to merging this pull request? Seemed to have no specific objections...

I think there is at least a small bit more to be resolved, and I have one more separate comment to make.

> Can I include forum link to code style document?

perhaps I am just not thinking hard enough at the moment to properly interpret this, but could you clarify what specific link you are asking about adding to what specific document?

CONTRIBUTING.md
```diff
@@ -33,6 +33,7 @@ wait for feedback from the developers.
 
 To ensure your PR won't be rejected please make sure that:
 
+* All CI tests for PR are passed, do not assign reviewers before got test results.
```
Dilvish-fo: If this was here before, it hadn't really sunk in for me-- what is the reasoning for saying "do not assign reviewers before got test results"?  (and at the least the grammar should be corrected to "...before getting...").

Dilvish-fo (changes requested):
This will probably happen if/when you make a text change in the area I asked about, but it seems we should see confirmation that this is passing the CI/travis check before it actually is merged.

Cjkjvfnby:
> but could you clarify what specific link you are asking about adding to what specific document?

Should I add link to forum [thread](http://www.freeorion.org/forum/viewtopic.php?f=9&t=10900) to the [default/python/README.md](https://github.com/freeorion/freeorion/blob/master/default/python/README.md)

CONTRIBUTING.md
Cjkjvfnby: Should I remove this line or only improve message?

Vezzra:
> Should I add link to forum [thread](http://www.freeorion.org/forum/viewtopic.php?f=9&t=10900) to the [default/python/README.md](https://github.com/freeorion/freeorion/blob/master/default/python/README.md)

Um, maybe I missed something, but wasn't the point to avoid exactly that? To have no links within the repo to external resources? To quote @adrianbroher:

> If you want document this properly expand the `CONTRIBUTING.md` with a link to the `default/python/README.md` and explain how to do the actual testing with the required prerequisites and dependencies and how tool invoke the test tool.

As I understand this statement, put a proper explanation into `default/python/README.md` and a link to `default/python/README.md` into  `CONTRIBUTING.md`. No links to forum threads/posts required - unless I misunderstood something, in that case, sorry. I admit I didn't completely get the discussion here... 😔

CONTRIBUTING.md
Dilvish-fo: > Should I remove this line or only improve message?

Well, I was wanting to actually talk out the idea, not necessarily just remove the line.  I suppose the thinking is that until it passes the CI tests, it could be considered slightly premature to ask for a review, since the CI tests might mandate changes.  But I am inclined to think that there are already plenty of things besides CI tests that could cause mid-PR changes, like reviews from other people, or just the OP decides to change something or add more commits.   So it seems to me that just like now, it should be fine for people to assign reviewers when they make the PR.

So if you feel strongly about it, let's continue the discussion, otherwise, yes, let's take that line out.

Cjkjvfnby: Removed, rebased.

Dilvish-fo (approved):

PR #2032, opened 2018-03-05
Fix AI mock interface typehinting for getVisibilityTurnMap

Morlic-fo:
While the mock function itesel was coded to return a dict instead of a VisibilityTurnMap, the dosctring specified the VisibilityTurnMap as :rtype:. The :rtype: is subsequently used by at least some IDEs as the inferred return type.

Specific warning fixed is ```unresolved attribute reference 'get' for class VisibilityIntMap```.

### DIFF
--- a/default/python/AI/freeOrionAIInterface.pyi
+++ b/default/python/AI/freeOrionAIInterface.pyi
@@ -1566,9 +1566,9 @@ class empire(object):
 
     def supplyProjections(self):
         """
-        :rtype: IntIntMap
+        :rtype: dict[int, int]
         """
-        return IntIntMap()
+        return dict()
 
     def techResearched(self, string):
         """
--- a/default/python/AI/freeOrionAIInterface.pyi
+++ b/default/python/AI/freeOrionAIInterface.pyi
@@ -2450,7 +2450,7 @@ class universe(object):
         :type number1: int
         :param number2:
         :type number2: int
-        :rtype: VisibilityIntMap
+        :rtype: dict[int, int]
         """
         return dict()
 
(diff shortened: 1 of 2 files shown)


Dilvish-fo:
This appears to me to be an AI infrastructure type issue, [which Vezzra has asked us to not assign milstones to](https://github.com/freeorion/freeorion/pull/2022#issuecomment-370248528).

On the substance of it, it looks fine to me, but since there is no real rush let's let it wait for CJ to have a chance to review, since the whole stub/interface is primarily his.

Morlic-fo:
Noted. I won't assign milestones for things like these in the future.

Cjkjvfnby (approved):
Please add comment:

```
TODO:  Change map like class parent from object to dict.  Or create special class that will be used as parent, this will greatly reduce number of lines in stub, and we can make good method docstrings. 
```

Vezzra:
Just a brief note on the "infrastructure" and "no milestone" thing: I noticed a couple of PRs that have been tagged as "infrastructure", of which at least some seem to be kind of edge cases.

"Infrastructure" usually refers to things like e.g. the Travis and AppVeyor CI, tools which check code style etc. It does *not* apply e.g. to logging code or any changes to the actual codebase (be it C++ or Python), IMO.

That's why "infrastructure" issues/PRs don't get assigned milestones - they don't affect the codebase, and therefore the builds, hence it doesn't really matter when such changes go in, hence no milestone.

On the other hand, everything that *does* change the the actual codebase *must* get a milestone assigned (because when such changes get in *does* affect the builds). Let this be your guideline when you decide on what to tag as "infrastructure" and assigning milestones.

Dilvish-fo:
Doh!

Cjkjvfnby:
> Just a brief note on the 

I think part of this comment should be added here https://github.com/freeorion/freeorion/blob/master/.github/labels.md#label-componentinfrastructure

Vezzra:
Done: #2043.

PR #2036, opened 2018-03-07
Replace print statements with logger calls in ShipDesignAI.py

Morlic-fo:
Chosen log levels seemed reasonable but are open for discussion...


Dilvish-fo (approved):
The the log level choices look fine to me.

Dilvish-fo (changes requested):
Could you rebase this now?  I think after the merge of CJs default logger PR you'll need to add ```debug``` and ```info``` to the import line.

PR #2038, opened 2018-03-09
Update readme

Cjkjvfnby:


default/python/README.md
```diff
@@ -28,15 +28,15 @@ We use [flake8-putty](https://pypi.python.org/pypi/flake8-putty)
 which include [flake8](https://pypi.python.org/pypi/flake8)
 which include [pycodestyle](https://pypi.python.org/pypi/pycodestyle) and other tools.
 
-`flake8-putty` is `flake8` plugin that allows one to disable rule for certain file, see `putty-ignore` section in `tox.ini`
+`flake8-putty` is `flake8` plugin that allows one to disable rule for certain file,
+see `putty-ignore` section in `tox.ini`. This allow to ignore certain
+warnings only for specified files, like bare excepts, that hard to fie,
+or special formatting in table like files.
```
Dilvish-fo: perhaps "files with tables"?
or "table-like files"?
We do, or did, have one of the latter for universe generation, though it's my recollection that's changed over now to being a python file with table-like entries, so probably "files with tables" is best.

default/python/tox.ini
```diff
@@ -13,3 +11,6 @@ putty-ignore =
   universe_generation\universe_generator.py : +E402
   universe_generation\universe_tables.py : +E201,E122,E231
   interface_stub\__init__.py : +F401
+  # TODO: fix this warnings
```
Dilvish-fo: "_these_ warnings"

default/python/tox.ini
```diff
@@ -13,3 +11,6 @@ putty-ignore =
...
+  *.py : +F812 # list comprehension redefines '<variable>' from line <line_no>
```
Dilvish-fo: I'm glad you are highlighting these, I think it would be especially good to get any F812 errors fixed right away.

Dilvish-fo (commented):
I think you were looking for a grammar review, yes?  So I have a few corrections/suggestions.

default/python/tox.ini
Cjkjvfnby: I will made it in one of the next PRs.

default/python/README.md
```diff
@@ -28,10 +28,10 @@ We use [flake8-putty](https://pypi.python.org/pypi/flake8-putty)
 which include [flake8](https://pypi.python.org/pypi/flake8)
 which include [pycodestyle](https://pypi.python.org/pypi/pycodestyle) and other tools.
 
-`flake8-putty` is `flake8` plugin that allows one to disable rule for certain file,
-see `putty-ignore` section in `tox.ini`. This allow to ignore certain
-warnings only for specified files, like bare excepts, that hard to fie,
-or special formatting in table like files.
+`flake8-putty` is `flake8` plugin that allows one to disable one or more rules(s) for a certain file,
```
Dilvish-fo: I'm sorry, I had a typo in my previous suggestion.  This should be simply "one or more rules"

An alternate form would be "one or more rule(s)", but it is never proper to write "rules(s)"

Dilvish-fo (approved):

PR #2039, opened 2018-03-10
String table thread safety

geoffthemedio:
Hopefully makes stringtable use more (but not completely and foolproof) thread safe. Makes error strings, which are returned by reference, persistent, so that another thread requesting a missing string won't invalidate previously returned references to error strings. Also adds mutexes and locks for accessing a stringtable. Stores loaded strings in shared pointers instead of raw pointers, which may avoid leaking memory when flushing stringtables. Prevents uninitialized stringtables from being used as fallback stringtables for a second stringtable, which could prevent weird results when the second tries to use the uninitialized contents on the first to set its backup info.

### DIFF
--- a/util/StringTable.cpp
+++ b/util/StringTable.cpp
@@ -11,41 +11,61 @@
 #include <iostream>
 
 
-// static(s)
-const std::string StringTable_::S_DEFAULT_FILENAME = "en.txt";
-const std::string StringTable_::S_ERROR_STRING = "ERROR: ";
+namespace {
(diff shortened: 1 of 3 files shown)


Dilvish-fo:
Just to note here where it is most important: [dbenage-cx reports that this PR appears to clear up the thread-related stringtable crashes that the Description() version of 2018 was causing for him](https://github.com/freeorion/freeorion/pull/2018#issuecomment-372070599), so this PR does appear to have at least made substantial improvements to the stringtable robustness.

PR #2023, opened 2018-02-26
BT_STOCKPILE

agrrr3:
The aim of this branch to make the [stockpile "project" build item](http://www.freeorion.org/forum/viewtopic.php?f=9&t=10892) work.

State: functional. So "make it work" is done

- stockpiling projects have a quantity and a repetition count
- basic cost of a stockpiling project is 1PP and 1 turn
- stockpiling projects are not allowed to draw from the imperial stockpile and the menu option for enabling is only shown for ships and buildings
- PP spent on a stockpiling project are added to the next turns imperial stockpile
- there is nothing special happening when a stockpiling project is finished

General next steps:
* Fix issue: encyclopedia entry for stockpiling (not sure what i did wrong, help wanted); probably open a issue after merging
* code review, cleanup, squash, merge (discussion on github)
* playtesting (help wanted)
* maybe refactor build item structure (suggestions welcome)

For "make it right" - I am not completely happy with having another overloaded version of build item type on so many different levels. Open for suggestions.

PR #2041, opened 2018-03-10
BT stockpile review changes

agrrr3:
These are the review changes post commit for the bt_stockpile branch.

Note that I am not completely sure if I have done the-right-thing(TM) with the switch statements in the ProductionWnd. I hoped to get rid of having to specify the default case by early check for the build types and return but the compiler does not seem to be able to infer this.

Original PR: https://github.com/freeorion/freeorion/pull/2023

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -447,21 +447,17 @@ bool Empire::ProducibleItem(BuildType build_type, int location_id) const {
     if (build_type == BT_BUILDING)
         throw std::invalid_argument("Empire::ProducibleItem was passed BuildType BT_BUILDING with no further parameters, but buildings are tracked by name");
 
-    auto build_location = GetUniverseObject(location_id);
-    if (!build_location)
-        return false;
-
     // must own the production location...
     auto location = GetUniverseObject(location_id);
     if (!location) {
-        WarnLogger() << "ShipDesign::ProductionLocation unable to get location object with id " << location_id;
+        WarnLogger() << "Empire::ProducibleItem for BT_STOCKPILE unable to get location object with id " << location_id;
         return false;
     }
 
     if (!location->OwnedBy(m_id))
         return false;
 
-    if (!std::dynamic_pointer_cast<const Planet>(location))
+    if (!std::dynamic_pointer_cast<const ResourceCenter>(location))
         return false;
 
     if (build_type == BT_STOCKPILE) {
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -451,28 +451,22 @@ bool Empire::ProducibleItem(BuildType build_type, int location_id) const {
     if (!build_location)
         return false;
 
+    // must own the production location...
+    auto location = GetUniverseObject(location_id);
+    if (!location) {
+        WarnLogger() << "ShipDesign::ProductionLocation unable to get location object with id " << location_id;
+        return false;
+    }
+
+    if (!location->OwnedBy(m_id))
+        return false;
+
+    if (!std::dynamic_pointer_cast<const Planet>(location))
+        return false;
+
     if (build_type == BT_STOCKPILE) {
-        Empire* empire = GetEmpire(m_id);
-        if (!empire) {
-            DebugLogger() << "ShipDesign::ProductionLocation: Unable to get pointer to empire " << m_id;
-            return false;
-        }
-
-        // must own the production location...
-        auto location = GetUniverseObject(location_id);
-        if (!location) {
-            WarnLogger() << "ShipDesign::ProductionLocation unable to get location object with id " << location_id;
-            return false;
-        }
-        if (!location->OwnedBy(m_id))
-            return false;
-
-        auto planet = std::dynamic_pointer_cast<const Planet>(location);
-        std::shared_ptr<const Ship> ship;
-        if (!planet)
-             return false;
-
         return true;
+
     } else {
         ErrorLogger() << "Empire::ProducibleItem was passed an invalid BuildType";
         return false;
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -445,7 +445,7 @@ bool Empire::ProducibleItem(BuildType build_type, int location_id) const {
         throw std::invalid_argument("Empire::ProducibleItem was passed BuildType BT_SHIP with no further parameters, but ship designs are tracked by number");
     
     if (build_type == BT_BUILDING)
-        throw std::invalid_argument("Empire::ProducibleItem was passed BuildType BT_BUILDING with no further parameters, but these types are tracked by name");
+        throw std::invalid_argument("Empire::ProducibleItem was passed BuildType BT_BUILDING with no further parameters, but buildings are tracked by name");
 
     auto build_location = GetUniverseObject(location_id);
     if (!build_location)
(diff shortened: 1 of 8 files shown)


Empire/ProductionQueue.cpp
```diff
@@ -145,9 +145,8 @@ namespace {
       * Elements will not receive funding if they cannot be produced by the
       * empire with the indicated \a empire_id this turn at their build location. 
       * Also checks if elements will be completed this turn. */
-    void SetProdQueueElementSpending(
+    float SetProdQueueElementSpending(
```
geoffthemedio: indicate meaning of return value in comments?

Empire/ProductionQueue.cpp
```diff
@@ -179,7 +178,7 @@ namespace {
         allocated_pp.clear();
         allocated_stockpile_pp.clear();
         float dummy_pp_source = 0;
-
+        float stockpile_transfer = 0;
```
geoffthemedio: `0.0f` for `float`

Empire/Empire.cpp
```diff
@@ -445,34 +445,28 @@ bool Empire::ProducibleItem(BuildType build_type, int location_id) const {
...
+        return false;
 
-        auto planet = std::dynamic_pointer_cast<const Planet>(location);
-        std::shared_ptr<const Ship> ship;
-        if (!planet)
-             return false;
+    if (!std::dynamic_pointer_cast<const Planet>(location))
+        return false;
 
+    if (build_type == BT_STOCKPILE) {
```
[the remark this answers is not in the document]
agrrr3: The code is the way it is because of keeping the pattern in the other ProducibleItem functions. 
So if I change this - should I change the other Empire::ProducibleItem functions as well?

Empire/Empire.cpp
```diff
@@ -445,34 +445,28 @@ bool Empire::ProducibleItem(BuildType build_type, int location_id) const {
...
         return false;
 
-    if (build_type == BT_STOCKPILE) {
-        Empire* empire = GetEmpire(m_id);
-        if (!empire) {
-            DebugLogger() << "ShipDesign::ProductionLocation: Unable to get pointer to empire " << m_id;
-            return false;
-        }
+    // must own the production location...
+    auto location = GetUniverseObject(location_id);
```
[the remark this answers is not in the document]
agrrr3: That was for getting a warning if bad object id like in Ship::ProductionLocation. As preparation for ship-based production i cast rather to ResourceCenter.

UI/ProductionWnd.cpp
```diff
@@ -742,23 +749,26 @@ namespace {
                 } else {
                     popup->AddMenuItem(GG::MenuItem(UserString("ALLOW_IMPERIAL_PP_STOCKPILE_USE"), false, false, allow_stockpile_action));
                 }
-                break;
-            case BT_STOCKPILE:
+	    default:
```
geoffthemedio: tab.

find and replace any / all such

UI/ProductionWnd.cpp
```diff
@@ -742,23 +749,26 @@ namespace {
...
-            } else if (build_type == BT_STOCKPILE) {
+                break;
+            case BT_STOCKPILE:
                 item_name = "PROJECT_BT_STOCKPILE";
-            } else {
-                ErrorLogger() << "Invalid build type (" << build_type << ") for row item";
-                return;
+                break;
+            default:
+	        break;
```
geoffthemedio: ...

Empire/ProductionQueue.cpp
```diff
@@ -307,9 +309,11 @@ ProductionQueue::ProductionItem::ProductionItem() :
 {}
 
 ProductionQueue::ProductionItem::ProductionItem(BuildType build_type_) :
-    build_type(build_type_),
-    name("PROJECT_BT_STOCKPILE")
-{}
+    build_type(build_type_)
+{
+    if (build_type_ == BT_STOCKPILE)
+        name = "PROJECT_BT_STOCKPILE";
```
geoffthemedio: Can this be / is this ever shown to the user as a stringtable lookup? If so, should wrap in `UserStringNoOp`

Empire/Empire.cpp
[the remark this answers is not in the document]
geoffthemedio: Looking at it now / again, it seems fine as is

Empire/ProductionQueue.cpp
agrrr3: Is it  a problem to nest a UserStringNoOp like this: UserString(UserStringNoOp("MSG_BLABLA"))?

Because ProductionWnd wraps the item names like this: UserString(elem.item.name);

geoffthemedio: If a string literal is already written in a `UserString` call then it doesn't also need a `NoOp` wrapper.

agrrr3: So this stays the way it was. The buildings also store their stringtable keys in the name as simple text. UserStringification happens on use site.

UI/BuildDesignatorWnd.cpp
```diff
@@ -331,7 +331,6 @@ namespace {
 
         // production item is a stockpiling project
         if (item.build_type == BT_STOCKPILE) {
-
```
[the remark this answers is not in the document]
agrrr3: Done. I think all of the original concerns are dealt with.

Empire/ProductionQueue.cpp
geoffthemedio: If it is written in a `UserString` call, it doesn't need `UserStringNoOp`. If it is assigned to a variable that is later passed into `UserString`, then the string literal it should be wrapped in `UserStringNoOp`.

agrrr3:
Note that d1a14be is no review outcome but rather a bugfix.

Empire/ProductionQueue.cpp
agrrr3: Why should i wrap a string in a UserStringNop at all? For documentation of intent?

Hm. I think i dont quite grok this. As far as i understand the intention, UserStringNop should be used instead of UserString if you do not want any stringtable lookup to happen (e.g. to preserve a key). Implementation-wise it does get stripped away anyway - so doesnt have any effect at all?

As far as I can see, multiple nested UserString calls like UserString(UserString(key)) would usually do the wrong thing (using the value of key as key for another lookup). In this case as the UserString call happens on the use site of ProductionItem, I should return the key. 

I think your reply means I should use UserStringNop in this case. I do that change, even if i do not understand why its needed.

geoffthemedio: It's there so automated analysis can easily find stringtable keys in code. If a string literal in code will ever be looked up explicitly as a stringtable key, then it needs to appear in a `UserString` or a `UserStringNoOp` wrapper (but not both).

Empire/Empire.cpp
```diff
@@ -1094,14 +1083,15 @@ ErrorLogger() << "Empire::PlaceProductionOnQueue() : BT_STOCKPILE";
...
-            << boost::lexical_cast<std::string>(build_type) << "  location: " << location;
+                      << boost::lexical_cast<std::string>(build_type) << "  location: " << location;
         return;
     }
 
     const bool paused = false;
     const bool allowed_imperial_stockpile_use = false;
-    const std::string dummy_str = "BT_STOCKPILE_DUMMY_BLD_NAME";
-    ProductionQueue::Element build(build_type, dummy_str, m_id, number, number, blocksize, location, paused, allowed_imperial_stockpile_use);
+    const std::string name = "PROJECT_BT_STOCKPILE";
```
geoffthemedio, line 1092: `PROJECT_BT_STOCKPILE` is in the stringtable. Does this need a wrapper?

PR #2043, opened 2018-03-11
Enhance description for label "component:infrastructure"

Vezzra:
As suggested [here](https://github.com/freeorion/freeorion/pull/2032#issuecomment-372098748) by @Cjkjvfnby in the comment section of #2032, my proposal for the changes/additions to the description of the "component:infrastructure" label.

### DIFF
--- a/.github/labels.md
+++ b/.github/labels.md
@@ -112,7 +112,16 @@ content of the PR may require extensive gameplay balance testing.
 
 The Issue/PR deals with non-game related infrastructure that aids development.
 For example this includes issues with the project homepage, used external
-services like continuous integration or hosting of release binaries.
+services like continuous integration, hosting of release binaries, tools for
+checking code style, etc.
+
+Infrastructure Issues/PRs should not entail/contain any changes to the actual
+codebase unless such changes are essentially required by the Issue/PR and would
+not make sense outside of its context. Any such non-codebase Infrastructure
+Issues/PRs should not be assigned a milestone unless they have particular
+relevance to the milestone, such as by establishing some type of
+testing/verification of the codebase which is desired to be completed for the
+milestone. 
 
 #### Label `component:internal`
 


.github/labels.md
```diff
@@ -112,7 +112,13 @@ content of the PR may require extensive gameplay balance testing.
 
 The Issue/PR deals with non-game related infrastructure that aids development.
 For example this includes issues with the project homepage, used external
-services like continuous integration or hosting of release binaries.
+services like continuous integration, hosting of release binaries, tools for
+checking code style, etc.
+
+Infrastructure Issues/PRs usually do *not* contain any changes to the actual
```
adrianbroher: Isn't this whole paragraph (excluding the "don't assign milestones" part) a reiteration of the previous one?

Vezzra: I wouldn't say "reiteration", but spelling out something the previous paragraph clearly implies. *IMO* this spelling out isn't really necessary (except maybe the part clarifying that no milestones need to be assigned, as you pointed out).

I'm just trying to honor @Cjkjvfnby's request in #2032. He wasn't specific when he said that part of my comment should be added to the description of the "component:infrastructure" label, so maybe he only meant the "no milestone" thing anyway.

Let's wait and see what @Cjkjvfnby and @Dilvish-fo have to say.

Cjkjvfnby: > He wasn't specific when he said that part of my comment should be added to the description of the "component:infrastructure" 

Texts are not my strong part :)  Current state is OK for me.

Dilvish-fo: Hmm, it looks to me like the second paragraph partly reiterates the first while trying, and at least partially succeeding, to clarify the first, and includes some new info (about the no milestones).  The second paragraph could easily be incorporated into the first, preference between one or two paragraphs seems like a style choice but also would be partly dependent on just how much of this winds up added.

Right now this second paragraph only partly succeeds in clarifying the first because it is partly self contradictory.  The main clause of the first sentence states that infrastructure PRs "_usually_" do not contain any changes to actual code, which implies that sometimes they can.  But the parenthetical says they categorically should not.  If the possibility of exceptions is truly contemplated, but only extremely rarely, then I would suggest something along the lines of 
```
Absent exceptional circumstances, Infrastructure Issues/PRs should not contain any changes to the actual codebase.
```

If you intend it to be a firm rule that there truly never be actual code changes in an infrastructure PR, then the "Absent exceptional circumstances" would be tossed and "should not" changed to "cannot".  But just consider then what if there are some code changes that go along with some infrastructure change, you could always simply break them out as a separate PR, but then the "code" PR would get a milestone while the "infrastructure" PR it goes hand-in-hand with would not-- seems like a strange/wrong outcome.  Which segues to my next paragraph...

Regarding the sentence about milestones, well, if "no milestones for infrastructure" stays the policy then it certainly should be so stated in this/these paragraph(s) somewhere. I think I was rather inactive during the period when these labels and such policy was worked out, but I'm not finding anything here to clarify it and I feel compelled to note both that it doesn't really make a lot of sense to me, and  when searching around about milestone policies I have not yet found any support for it.  What is wrong with the idea that we might want to have certain infrastructure/testing in place for 0.4.8 to improve the code-style and/or reliability of that release, in which case related infrastructure PRs should bear that milestone?

I certainly don't see any support for the "For this reason" phrase of the milestone sentence.

Vezzra: This has developed into an important discussion about policies, which should take place on the forum, not here in the comment section of a PR. I've [opened a corresponding thread](http://www.freeorion.org/forum/viewtopic.php?f=4&t=10913) on the forum, please continue the discussion there.

Vezzra:
Depending on the outcome of the [policy discussion about infrastructure Issues/PRs and their milestones](http://www.freeorion.org/forum/viewtopic.php?f=4&t=10913) on the forum, the additions proposed in this PR are subject to change. I've therefore tagged this as WIP.

Vezzra:
Ok, I've updated this PR and incorporated what @Dilvish-fo [suggested on the forum](http://www.freeorion.org/forum/viewtopic.php?p=91941#p91941).

Further comment, opinions?

adrianbroher:
FYI: It seems like the labels now support descriptions. I have added a copy to the `status:work in progres` one, which is just 3 characters short of the maximum size. Maybe we should convert the `.github/labels` document into those descriptions.

Dilvish-fo (approved):
Looks good to me.

And Marcel's suggestion about leveraging GitHub's new support for label descriptions sounds very good also (though rather than truly "convert", which seems would imply doing away with the labels.md document, I would suggest we simply 'add' the descriptions.  That would leave us with some 'description duplication', but certainly this infrastructure description would (to my understanding) be too long for Github's support, and I don't think we should so drastically abbreviate it as seems would be necessary (at least not without leaving this longer version for a fuller explanation).

Vezzra:
Adding label descriptions based on our already existing "label documentation" sounds like a good idea to me too. I also agree with @Dilvish-fo that we probably shouldn't replace the existing "label documentation", as the limited descriptions supported by the new github feature can't completely replace those more elaborate explanations.

PR #2051, opened 2018-03-13
Don't let new ships/monsters blockade

### LINKED ISSUE #1968: Newly created ships (especially monsters) should not immediately create blockades
When monsters develop from an unarmed form to an armed one, like Snowflakes or Krill, that transformation occurs as an effect which takes place after combat, but before new supply blockade determinations are made.  If this transformation takes place over an empire's planet, a blockade is created for at least one turn, without the planet having a chance to defend itself against the blockade on that initial turn.  I find this quite annoying when it happens to me, and expect most other players do also.

~~[I suppose that the same thing could happen with a newly created ship in a system with planets owned by multiple empires.]~~ edit:  on checking, ships are created after supply/blockade checks, so this wouldn't happen

I propose that monsters would not create supply blockades on the turn of their creation.

Dilvish-fo:
As noted in #1968 newly created ships/monsters should not create blockades, since they have not gone through even a potential combat cycle and thus may not be possible to defend against yet.  The most common culprits are Floaters transforming into Dyson Trees, and various other maturing monsters.

The first commit focuses on Supply blocks.  That generally also resolves the fleet movement blockade issue, but I think there were still some potential cases that needed to be dealt with, such as when supply is rightly being blocked by a long-existing but relatively blind ship (so blind it couldn't block the fleet movement of concern) in the same system as in which a differently owned ship is newly created. 

I have done some limited testing and this PR appears to work fine, although my testing was really all directed to the subject of the first commit, the situations the second commit are be directed to would be quite a bit trickier to get set up.

Resolves #1968

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -782,7 +782,15 @@ void Empire::UpdateSupplyUnobstructedSystems(const std::set<int>& known_systems)
                 }
             } else if (fleet->NextSystemID() == INVALID_OBJECT_ID || fleet->NextSystemID() == fleet->SystemID()) {
                 int fleet_owner = fleet->Owner();
-                if (fleet_owner == ALL_EMPIRES || Empires().GetDiplomaticStatus(m_id, fleet_owner) == DIPLO_WAR) {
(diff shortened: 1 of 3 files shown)


geoffthemedio (commented):
If it works, no objections to C++ changes...

PR #2054, opened 2018-03-16
Allow authenticated player to replace his connection without stopping the game.

o01eg:
Currently when authenticated player replaces his frozen connection game stops.
This PR change links between empire and player connections so old connection detaches from empire and new connection attaches to it and continue to play without requirement to restart the game.

Player turns out to the end in the empire window. Should we consider other sorting than by player id? I think it better to change sorting order in other independent PR.

This PR doesn't affect neither single player game nor multiplayer game without authentication.

P.S. Player resets his waiting state so he could send orders again. This works for loaded game, is it expected behaviour? It will allow authenticated player to ignore waiting state and fix his orders after reconnection. Also other players doesn't get he resets his waiting state. Discussion about it: http://freeorion.org/forum/viewtopic.php?f=6&t=10915

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -1638,6 +1638,44 @@ void ServerApp::AddObserverPlayerIntoGame(const PlayerConnectionPtr& player_conn
     // TODO: notify other players
 }
 
+void ServerApp::DropPlayerEmpireLink(int player_id)
+{ m_player_empire_ids.erase(player_id); }
+
+bool ServerApp::AddPlayerIntoGame(const PlayerConnectionPtr& player_connection) {
+    // search empire by player name
+    for (auto empire : Empires()) {
+        if (empire.second->PlayerName() == player_connection->PlayerName()) {
+            auto orders_it = m_turn_sequence.find(empire.first);
+            if (orders_it == m_turn_sequence.end()) {
+                WarnLogger() << "ServerApp::AddPlayerIntoGame empire " << empire.first
+                             << " for \"" << player_connection->PlayerName()
+                             << "\" doesn't wait for orders";
+                return false;
+            }
+            // make a link to new connection
+            m_player_empire_ids[player_connection->PlayerID()] = empire.first;
+
+            const OrderSet dummy;
+            const OrderSet& orders = orders_it->second ? *orders_it->second : dummy;
+
+            auto player_info_map = GetPlayerInfoMap();
+            bool use_binary_serialization = player_connection->ClientVersionStringMatchesThisServer();
+
+            player_connection->SendMessage(GameStartMessage(
+                m_single_player_game, empire.first,
+                m_current_turn, m_empires, m_universe,
+                GetSpeciesManager(), GetCombatLogManager(),
+                GetSupplyManager(),  player_info_map, orders,
+                static_cast<const SaveGameUIData*>(nullptr),
+                m_galaxy_setup_data,
+                use_binary_serialization));
+
+            return true;
+        }
+    }
+    return false;
+}
+
 bool ServerApp::IsHostless() const
 { return GetOptionsDB().Get<bool>("hostless"); }
 
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -288,18 +288,6 @@ bool ServerFSM::EstablishPlayer(const PlayerConnectionPtr& player_connection,
 {
     std::list<PlayerConnectionPtr> to_disconnect;
 
-    if (player_connection->IsAuthenticated()) {
-        // drop other connection with same name
-        for (auto it = m_server.m_networking.established_begin();
-             it != m_server.m_networking.established_end(); ++it)
-        {
-            if ((*it)->PlayerName() == player_name && player_connection != (*it)) {
-                (*it)->SendMessage(ErrorMessage(UserString("ERROR_CONNECTION_WAS_REPLACED"), true));
-                to_disconnect.push_back(*it);
-            }
-        }
-    }
-
     // set and test roles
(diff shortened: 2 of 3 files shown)


Vezzra:
@o01eg:

> WIP: ...

Does that mean the PR itself is still work in progress, or is that something which should be worked on once the PR got merged? Just to know if we should tag this as "work in progress"...

o01eg:
PR is finished IMHO but it something should be discussed before merge.

server/ServerApp.cpp
```diff
@@ -1638,6 +1638,42 @@ void ServerApp::AddObserverPlayerIntoGame(const PlayerConnectionPtr& player_conn
     // TODO: notify other players
 }
 
+void ServerApp::DropPlayersEmpire(int player_id) {
+    m_player_empire_ids.erase(player_id);
```
geoffthemedio: prefer `{}` on single line for single line function

o01eg:
Maybe it not any critical as connection replacement available only for authenticated user. Most users doesn't see changes even in multiplayer.

server/ServerApp.cpp
geoffthemedio: Rename -> `DropPlayerEmpire` ?

o01eg: May be even `DropPlayerEmpireLink` as it drops link between player and empire not empire itself.

geoffthemedio: sure

server/ServerApp.cpp
```diff
@@ -1638,6 +1638,41 @@ void ServerApp::AddObserverPlayerIntoGame(const PlayerConnectionPtr& player_conn
...
+            if (orders_it == m_turn_sequence.end()) {
+                WarnLogger() << "ServerApp::AddPlayerIntoGame empire " << empire.first
+                             << " for \"" << player_connection->PlayerName()
+                             << "\" doesn't wait for orders";
+                return false;
+            }
+            // make a link to new connection
+            m_player_empire_ids[player_connection->PlayerID()] = empire.first;
+
+            const OrderSet& orders = orders_it->second ? *orders_it->second : OrderSet();
```
geoffthemedio: I think you're binding a reference to a temporary here...

server/ServerApp.cpp
```diff
@@ -1638,6 +1638,42 @@ void ServerApp::AddObserverPlayerIntoGame(const PlayerConnectionPtr& player_conn
...
+            // make a link to new connection
+            m_player_empire_ids[player_connection->PlayerID()] = empire.first;
+
+            const OrderSet dummy;
+            const OrderSet& orders = orders_it->second ? *orders_it->second : dummy;
+
+            std::map<int, PlayerInfo> player_info_map = GetPlayerInfoMap();
+            bool use_binary_serialization = player_connection->ClientVersionStringMatchesThisServer();
+
+            player_connection->SendMessage(GameStartMessage(m_single_player_game, empire.first,
```
geoffthemedio: bump the inner parameters to a new line... these are all a bit too indented to be easily readable

server/ServerApp.cpp
```diff
@@ -1638,6 +1638,42 @@ void ServerApp::AddObserverPlayerIntoGame(const PlayerConnectionPtr& player_conn
...
+                             << "\" doesn't wait for orders";
+                return false;
+            }
+            // make a link to new connection
+            m_player_empire_ids[player_connection->PlayerID()] = empire.first;
+
+            const OrderSet dummy;
+            const OrderSet& orders = orders_it->second ? *orders_it->second : dummy;
+
+            std::map<int, PlayerInfo> player_info_map = GetPlayerInfoMap();
```
geoffthemedio: could just use `auto`

geoffthemedio:
otherwise, if it works, ok

server/ServerApp.cpp

server/ServerApp.cpp

server/ServerApp.cpp
```diff
@@ -1638,6 +1638,44 @@ void ServerApp::AddObserverPlayerIntoGame(const PlayerConnectionPtr& player_conn
...
+
+            auto player_info_map = GetPlayerInfoMap();
+            bool use_binary_serialization = player_connection->ClientVersionStringMatchesThisServer();
+
+            player_connection->SendMessage(GameStartMessage(
+                m_single_player_game, empire.first,
+                m_current_turn, m_empires, m_universe,
+                GetSpeciesManager(), GetCombatLogManager(),
+                GetSupplyManager(),  player_info_map, orders,
+                static_cast<const SaveGameUIData*>(nullptr),
```
geoffthemedio, line 1669: is this `static_cast` necessary?

server/ServerFSM.cpp
```diff
@@ -2095,6 +2103,14 @@ void PlayingGame::EstablishPlayer(const PlayerConnectionPtr& player_connection,
             {
                 // send playing game
                 server.AddObserverPlayerIntoGame(player_connection);
+            } else if (client_type == Networking::CLIENT_TYPE_HUMAN_PLAYER) {
+                // previous connection was dropped
+                // set empire link to new connection by name
+                // send plaing game
```
geoffthemedio: plaing ?

server/ServerApp.cpp
o01eg: Yes, there is a `GameStartMessage` function accepts same arguments but `const char *` instead `const SaveGameUIData*`.

server/ServerFSM.cpp

geoffthemedio:
Could you combine into one commit and re-push?

o01eg:
Yes. Rebased and squashed.

PR #2058, opened 2018-03-17
appveyor: Setup skip options

Cjkjvfnby:
It is reworked example of changes suggested by @dbenage-cx  on [forum](http://www.freeorion.org/forum/viewtopic.php?f=9&p=88852#p88852)   

[AppVeyor docs](https://www.appveyor.com/docs/how-to/filtering-commits/#skip-commits-2)

I have tested this PR over different cases, it works.
I think we can setup more precise exclusion schema, but I don know a lot about project compiling.

I have added `[skip appveyor]` and `[appveyor skip]` to match `[ci skip]` and `[skip ci]`  (that skips both Travis and Appveyor)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,7 +14,7 @@ skip_commits:
       - '**/*.md'
       - '**/*.xml'
       - '**/COPYING'
-  message: /.*\[skip appveyor\].*/
+  message: /.*\[(skip appveyor)|(appveyor skip)\].*/
 
 install:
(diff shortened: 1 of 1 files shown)


geoffthemedio:
could probably squash when done testing...

adrianbroher:
Could you please drop the `**/*.txt` entry, because it conflicts with #1416.

adrianbroher (approved):
Looks good to me.

PR #2059, opened 2018-03-19
Indicate rally destinations in ProductionWnd

### LINKED ISSUE #1862: Production rally color indicator
When the Production window is open, indicate by color which ships are rallied to the currently selected system.

Here is a mock up:

rally

Dilvish-fo:
If the (sidepanel) selected system is the rally destination for any ships
being built, have the ProductionQueue panel for such ships adjust their
location text to indicate such.  In such case "At Planet X" becomes "Rallying
from Planet X", and the text color changes to the darkened empire color complement.

rally

Resolves #1862

### DIFF
--- a/UI/ProductionWnd.cpp
+++ b/UI/ProductionWnd.cpp
@@ -467,10 +467,17 @@ namespace {
         // get location indicator text
         std::string location_text;
         bool system_selected = false;
+        bool rally_dest_selected = (elem.rally_point_id != INVALID_OBJECT_ID && elem.rally_point_id == SidePanel::SystemID());
         if (std::shared_ptr<const UniverseObject> location = GetUniverseObject(elem.location)) {
-            system_selected = (location->SystemID() != -1 && location ->SystemID() == SidePanel::SystemID());
-            if (GetOptionsDB().Get<bool>("ui.queue.production_location.shown"))
-                location_text = boost::io::str(FlexibleFormat(UserString("PRODUCTION_QUEUE_ITEM_LOCATION")) % location->Name()) + " ";
+            system_selected = (location->SystemID() != INVALID_OBJECT_ID && location ->SystemID() == SidePanel::SystemID());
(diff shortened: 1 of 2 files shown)


UI/ProductionWnd.cpp
```diff
@@ -467,10 +467,17 @@ namespace {
         // get location indicator text
         std::string location_text;
         bool system_selected = false;
+        bool rally_dest_selected = (elem.rally_point_id != -1 && elem.rally_point_id == SidePanel::SystemID());
```
geoffthemedio: probably `-1` -> `INVALID_OBJECT_ID`

UI/ProductionWnd.cpp
```diff
@@ -467,10 +467,17 @@ namespace {
...
         if (std::shared_ptr<const UniverseObject> location = GetUniverseObject(elem.location)) {
             system_selected = (location->SystemID() != -1 && location ->SystemID() == SidePanel::SystemID());
-            if (GetOptionsDB().Get<bool>("ui.queue.production_location.shown"))
-                location_text = boost::io::str(FlexibleFormat(UserString("PRODUCTION_QUEUE_ITEM_LOCATION")) % location->Name()) + " ";
+            if (GetOptionsDB().Get<bool>("ui.queue.production_location.shown")) {
+                if (rally_dest_selected && !system_selected) {
+                    location_text = boost::io::str(FlexibleFormat(UserString(
+                                    "PRODUCTION_QUEUE_ITEM_RALLIED_FROM_LOCATION")) % location->Name()) + " ";
```
geoffthemedio: indent additional 4 spaces or to match the previous line's parameters (which there are none of in this case)

Dilvish-fo: So, lining up under the "t" in location, correct?

geoffthemedio: Yep

PR #2061, opened 2018-03-22
Enclose oarchive scope when used with stringstream

dbenage-cx:
Resolves issue related to change in boost v1.66.0, see [boost trac #13354](https://svn.boost.org/trac10/ticket/13354)

On game start, server would abort with:
```
terminate called after throwing an instance of 'boost::archive::archive_exception'
  what():  input stream error-Resource temporarily unavailable
```
Stemming from [util/Message.cpp:1190](https://github.com/freeorion/freeorion/blob/master/network/Message.cpp#L1190)


geoffthemedio:
seems odd, but if it works then OK

PR #2063, opened 2018-03-25
Add build instructions manual

adrianbroher:
This PR adds a manual for building FreeOrion on most platforms.

The main advantage of this document compared to the Wiki pages should be, that this file is tracked with the source code version and future build changes, which isn't true for the Wiki pages.

It contains a condensed and updated information from the Wiki pages.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,0 +1,175 @@
+# Build FreeOrion
+
+
+## Hardware and OS Requirements
+
+FreeOrion will compile on Windows 7 (or later), Mac OS X 10.10 (or later) and
+Linux operating systems. Other operating systems have reported to work, but
+support is not actively maintained by the FreeOrion developers. FreeOrion is
+developed for x86 compatible processor architectures, other architectures
+haven't been tested.
+
+The FreeOrion source code makes heavy use of templates and requires much memory
+to compile ; 6 GiB RAM or more is recommended. To build FreeOrion, expect up to
+45 minutes on a Intel Core i5 system.
+
+
+## Required Software Dependencies
+
+FreeOrion depends on the following software to build:
+
+  * [Visual Studio] - 2015 for Windows Desktop Update 3 ; Windows only
+  * [Xcode] - 6.4 or later ; Mac OS X only
+  * [CMake] - 3.4 (Mac OS X) ; 3.1 or later (Other non-Windows)
+  * A C++11 compliant compiler - Other Operating Systems
+    * [GNU GCC] - 5.0 or later
+    * [Clang] - 3.9 or later
+  * [Python] - 2.7.*
+  * [Git]
+
+FreeOrion depends on the following libraries or APIs to run:
+
+  * OpenGL - 2.1 or later ; usually provided by the graphic card driver or
+    Operating System
+  * OpenGL Utilities (GLU) ; usually provided by the graphic card driver or
+    Operating System
+  * OpenAL - It's recommended to use the [OpenAL Soft] implementation
+  * [Boost] - 1.56 or later
+  * [zlib]
+  * [libpython] - 2.7.*
+  * [FreeType2]
+  * [libpng]
+  * [libogg]
+  * [libvorbis]
+  * [SDL2]
+
+
+## Obtaining FreeOrion Source Code and Software Dependencies
+
+For Windows and Mac OS X a [FreeOrion Software Development Kit] is provided as
+download to compile FreeOrion from source. It contains the preconfigured and
+-compiled build and runtime dependencies for the Visual Studio v140 toolchain on
+Windows and Mac OS X 10.9 SDK with Xcode 6.4 or later on Mac OS X.
+
+For Linux or other Operating Systems the build and runtime dependencies should
+be installed by the preferred way for the respective OS (e.g. via Package
+manager or compiling from source).
+
+Step by step procedure:
+
+ * On Windows or Mac OS X:
+   * Download the [FreeOrionSDK v8] from the FreeOrionSDK respository releases.
+ * Linux and other Operating Systems:
+   * Install build and runtime dependencies by the preferred way for the
+     respective OS.
+ * Create a project directory `freeorion-project`.
+ * On Windows:
+   * Unzip the SDK archive contents into the project directory.
+   * Execute the `bootstrap.bat` within the project directory. This will clone
+     the FreeOrion repository and place the dependencies at the correct place.
+ * On Mac OS X:
+   * Mount the SDK DMG image.
+   * Copy the contents of the mounted image into the project directory.
+   * Execute the `bootstrap.command` within the project directory. This will
+     clone the FreeOrion repository and place the dependencies at the correct
+     place.
+ * On Linux and other Operating Systems:
+   * Navigate into the project directory.
+   * Clone the project via Git:
+     ```
+     git clone https://github.com/freeorion/freeorion.git
+     ```
+
+This will leave you with the latest development branch `master` and the
+FreeOrion source code in:
+
+ * `freeorion-project/FreeOrion/` on Windows.
+ * `freeorion-project/` on Mac OS X.
+ * `freeorion-project/freeorion/` on Linux and other Operating Systems.
+
+This directory will be referred to as _source directory_ in the rest of the
+document.
+
+To build a specific release check out the desired version via `git checkout`:
+
+```
+git checkout vxxx
+```
+
+Where `vxxx` indicates the desired version.  A list of all available version
+can be listet by invoking `git tag -l`, where releases are indicated by a
+leading `v` followed by the release version number.
+
+
+# Compiling FreeOrion
+
+FreeOrion uses CMake as build system on most platforms. On Windows a manually
+maintained Visual Studio Project is used for building.
+
+Step by step procedure:
+
+ * Enter the _source directory_.
+ * On Windows:
+   * Open `msvc2015\FreeOrion.sln` with Visual Studio.
+   * Compile the whole project by selecting the `Build` -> `Build Solution`
+     menu entry.
+
+ * On Mac OS X:
+   * Create a `build` directory, which will contain all compile FreeOrion
+     build artifacs.
+   * Change into the `build` directory on the command line.
+   * Execute cmake to generate a Xcode project file:
+
+     ```
+     cmake -GXcode ..
+     ```
+   * Open `FreeOrion.xcodeproj` with Visual Studio
+   * Compile the whole project by selecting the `ALL_BUILD` scheme and
+     pressing 'Command' + 'B'.
+
+ * On Linux and other Operating Systems
+   * Create a `build` directory, which will contain all compile FreeOrion
+     build artifacs.
+   * Change into the `build` directory on the command line.
+   * Execute cmake to generate Makefiles:
+
+     ```
+     cmake ..
+     ```
+   * Compile the whole project by calling `make` within the build directory.
+     In case you want to utilize multiple CPU cores by running parallel compile
+     jobs check out the the [make jobs](`--jobs`) parameter of `make`.
+   * Create a symbolic link to the data directory inside the build directoy
+     by invoking:
+
+     ```
+     ln -s ../freeorion/default .
+     ```
+
+This will leave you with a build of FreeOrion executables.
+
+ * `freeorion-project/FreeOrion` on Windows.
+ * `freeorion-project/build/Release` on Mac OS X.
+ * `freeorion-project/freeorion/build` on Linux and other Operating Systems.
+
+
+[Visual Studio]: https://www.visualstudio.com/de/vs/older-downloads/
+[Xcode]: https://itunes.apple.com/de/app/xcode/id497799835?mt=12
+[CMake]: https://cmake.org/download/
+[GNU GCC]: https://gcc.gnu.org/releases.html
+[Clang]: http://releases.llvm.org/download.html
+[Python]: https://www.python.org/downloads/
+[Git]: https://git-scm.com/downloads
+[Boost]: http://www.boost.org/users/download/
+[zlib]: https://zlib.net/
+[libpython]: https://www.python.org/downloads/
+[FreeType2]: https://www.freetype.org/download.html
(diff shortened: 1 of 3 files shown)


BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+
+FreeOrion depends on the following software to build:
+
+  * [Visual Studio] - 2015 for Windows Desktop Update 3 ; Windows only
+  * [Xcode] - 6.4 or later ; Mac OS X only
+  * [CMake] - 3.4 (Mac OS X) ; 3.1 or later (Other non-Windows)
+  * A C++11 compliant compiler - Other Operating Systems
+    * [GNU GCC] - 5.0 or later
+    * [Clang] - 3.9 or later
+  * [Python] - 2 or later, but not 3 or later
```
geoffthemedio: In boost 1.56, the docs say Python 2.2 or later. I suspect FreeOrion requires something more like 2.7, which is the version included with the SDKs / installers.

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+  * [Git]
+
+FreeOrion depends on the following libraries or APIs to run:
+
+  * OpenGL - 2.1 or later ; usually provided by the graphic card driver or
+    Operating System
+  * OpenGL Utilities (GLU) ; usually provided by the graphic card driver or
+    Operating System
+  * OpenAL - It's recommended to use the [OpenAL Soft] implementation
+  * [Boost] - 1.56 or later
```
geoffthemedio, line 37: Probably 1.59

adrianbroher: The SDK contains 1.55, CMake required 1.56 and I'm not aware of any feature that we use and isn't provided by 1.56.

BUILD.md
adrianbroher: Good point, 2.0 is probably wrong. @Dilvish-fo are you aware of any minimum requirements in terms of the Python API (except the mentioned minimum requirements of Boost Python?)

BUILD.md
adrianbroher: Mh, seems like the whole version and documentation juggling made me dizzy.  The SDK contains 1.59, the Ubuntu CI uses 1.58.

BUILD.md
Morlic-fo: The AI scripts make use of python 2.7 features (e.g. dict comprehensions).

adrianbroher: Thanks for pointing out @Morlic-fo .

BUILD.md
```diff
@@ -0,0 +1,170 @@
+# Build FreeOrion
+
+
+## Hardware and OS Requirements
+
+FreeOrion will compile on most recent Windows, Mac OS X and Linux operating
```
Vezzra: I think with Windows and OSX we can be a bit more specific: Windows: Windows 7 (64bit) or later, Mac: OSX 10.10 or later.

Vezzra: I think with Windows and OSX we can be a bit more specific: Windows: Windows 7 (64bit) or later, Mac: OSX 10.10 or later.

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+systems. Other operating systems have reported to work, but support is not
+actively maintained by the FreeOrion developers. FreeOrion is developed for x86
+compatible processor architectures, other architectures haven't been tested.
+
+The FreeOrion source code makes heavy use of templates and requires much memory
+to compile ; 4GB RAM or more is recommended. To build FreeOrion, expect up to
```
Vezzra: That information probably should have been updated on the wiki long ago. Building FO on a system with merely 4GB RAM has become extremely difficult with the reimplementation of the parser (switching to boost spirit), which happened around the time when I joined the project (autumn 2011). I'm not sure if you can build current versions of FO even with 6GB RAM anymore.

To avoid frustration and make things easier for us, I'd recommend to raise the minimum RAM requirement for building FO to 8GB.

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+  * [libvorbis]
+  * [SDL2]
+
+
+## Obtaining FreeOrion Source Code and Software Dependencies
+
+For Windows and Mac OS X a [FreeOrion Software Development Kit] is provided as
+download to compile FreeOrion from source. It contains the preconfigured and
+-compiled build and runtime dependencies for Visual Studio v140 toolchain on
+Windows and Mac OS X 10.9 SDK with Xcode or later on Mac OS X.
```
Vezzra: "...for *the* Visual Studio v140 toolchain on Windows and Xcode 6.4 (and later) on Mac OS X."

Or, if you want to mention the required OS versions here too: "...for *the* Visual Studio v140 toolchain on Windows 7 (or later) and Xcode 6.4 (or later) on Mac OS X 10.10 (or later)."

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+
+For Linux or other Operating Systems the build and runtime dependencies should
+be installed by the preferred native way (e.g. via Package manager or compiling
```
Vezzra: I'm not a native speaker, but "preferred native way" sounds a bit odd. Maybe "preferred way of your distro", or "preferred way of the distro in question"?

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+from source).
+
+Step by step procedure:
+
+ * On Windows or Mac OS X:
+   * Download the [FreeOrionSDK v8] from the FreeOrionSDK respository releases.
+ * Linux and other Operating Systems:
+   * Install build and runtime dependencies by the preferdded native way.
```
Vezzra: See above.

Otherwise, if you want to stick with the "preferred native way" phrase, typo: preferdded -> preferred.

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+     ```
+
+This will leave you with the latest development branch `master` and the
+FreeOrion source code in:
+
+ * `freeorion-project/FreeOrion/` on Windows.
+ * `freeorion-project/` on Mac OS X.
+ * `freeorion-project/freeorion/` on Linux and other Operating Systems.
+
+This directory will be referred to _source directory_ in the rest of the
```
Vezzra: "...will be referred to *as*..."

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+document.
+
+To build a specific release check out the desired version with `git checkout`:
```
Vezzra: I think: "with" -> "via".

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+
+```
+git checkout vxxx
+```
+
+Where `vxxx` indicates the desired version.  A list of all available version
+can be listet by invoking `git tagi -l`, where release indicated by a leading
```
Vezzra: typo: `tagi` -> `tag`

"...where release**s** **are** indicated..."

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+     by invoking:
+
+     ```
+     ln -s ../freeorion/default .
+     ```
+
+This will leave you with a build of FreeOrion executables.
+
+ * `freeorion-project/FreeOrion/` on Windows.
+ * `freeorion-project/build` on Mac OS X.
```
Vezzra: `freeorion-project/build/Release` on Mac OS X.

BUILD.md
Dilvish-fo: Although 'preferred native way' does sound reasonable to me, the other variations do sound slightly better, I'll suggest "preferred way for the respective distro"

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+
+FreeOrion depends on the following software to build:
+
+  * [Visual Studio] - 2015 for Windows Desktop Update 3 ; Windows only
+  * [Xcode] - 6.4 or later ; Mac OS X only
+  * [CMake] - 3.4 (Mac OS X) ; 3.1 or later (Other non-Windows)
+  * A C++11 compliant compiler - Other Operating Systems
+    * [GNU GCC] - 5.0 or later
+    * [Clang] - 3.9 or later
+  * [Python] - 2.7 or later, but not 3 or later
```
Dilvish-fo: I personally think this line would read better as 
```
[Python] - 2.7.x (not 3.x)
```

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+ * On Linux and other Operating Systems
+   * Create a `build` directory, which will contain all compile FreeOrion
+     build artifacs.
+   * Change into the `build` directory on the command line.
+   * Execute cmake to generate Makefiles:
+
+     ```
+     cmake ..
+     ```
+   * Compile the whole project by calling `make` within the build directory.
```
Dilvish-fo, line 139: I get the impression from forum posts that for many people FO has been their first motivator to learn to compile, and it might be helpful to mention the ```-j n``` option for make.

BUILD.md
```diff
@@ -0,0 +1,170 @@
...
+FreeOrion depends on the following libraries or APIs to run:
+
+  * OpenGL - 2.1 or later ; usually provided by the graphic card driver or
+    Operating System
+  * OpenGL Utilities (GLU) ; usually provided by the graphic card driver or
+    Operating System
+  * OpenAL - It's recommended to use the [OpenAL Soft] implementation
+  * [Boost] - 1.56 or later
+  * [zlib]
+  * [libpython] - 2.7 or later, but not 3 or later
```
Dilvish-fo: same comment as above re 2.7.x

BUILD.md
Dilvish-fo: Also, while I can understand that you probably want to avoid having to mention too many different Linux flavors here with different details for each, it does seem a shame to not mention the sample one-liner command line for installing the dependencies via apt.

BUILD.md
adrianbroher: > That information probably should have been updated on the wiki long ago. Building FO on a system with merely 4GB RAM has become extremely difficult with the reimplementation of the parser (switching to boost spirit), which happened around the time when I joined the project (autumn 2011). I'm not sure if you can build current versions of FO even with 6GB RAM anymore.

YMMV, I'm building on a 6GB memory 6 core machine with almost full utilization via `-j 5` and it works just fine.  I would rather drop the specific memory size statement or remove the section completely.

BUILD.md
adrianbroher: Is the 64 bit limitation for Windows in place?

adrianbroher: Left out the 64 bit remark until clarified.

BUILD.md
adrianbroher: I wonder why this comment is now marked as outdated. I haven't touched that line and I would still like to discuss this.

BUILD.md
adrianbroher: Fixed, but I took the first suggestion without the OS version reiteration.

BUILD.md
adrianbroher: Your suggestion looks out of place compare to the pattern used in the previous entries, what about:

```
  * [Python] - 2.7 or later (not 3)
```

BUILD.md
adrianbroher: Fixed.

BUILD.md
adrianbroher: I adopted @Dilvish-fo suggestion, replacing `distro` with `OS`.

adrianbroher: @Dilvish-fo 
Regarding the _apt one-liner_: I avoided this on purpose, once for the reason you mentioned but also because we don't know on which distro version this `apt one-liner` is executed on.  Packages may be renamed between distro versions or the default packages are too stale for our requirements on LTS versions (eg. we need boost 1.58 on Ubuntu 16.04 LTS, which has 1.54(?) as default) so the one line is not really portable even between versions.

Which distributions and versions should be mentioned in the build instructions with their corresponding _package manager oneliner_?

BUILD.md
adrianbroher: I will raise the minimum requirements in a separate PR to boost 1.58.

BUILD.md
Morlic-fo: [There will be no Python 2.8](https://www.python.org/dev/peps/pep-0404/), so requiring specificically 2.7.x seems most precise.

Dilvish-fo: > Your suggestion looks out of place compare to the pattern used in the previous entries

It is indeed different than the previous patterns, precisely because of the basic difference in the requirement.  It is the first "or later", the one about 2.7, which is incomplete/misleading, so removing the second one doesn't really help.  Taken as a whole, either of your formulations do wind up accurately expressing the requirement, but it seems to me pointless and detrimental to leave an inconsistency/inaccuracy in the middle of them.

BUILD.md
Vezzra: > Is the 64 bit limitation for Windows in place?

I'm not sure what you mean by that. There are no settings in the VS project that would specifically require the Windows OS of machine on which you build FO to be of the 64bit variant. The 64bit requirement is just a consequence of building FO requiring more than 4GB RAM.

You could probably argue that having a min RAM requirement of >4GB and requiring a 64bit OS is basically redundant...

BUILD.md
Vezzra: > I'm building on a 6GB memory 6 core machine with almost full utilization via -j 5 and it works just fine.

Yeah, on your Linux system (which, besides being Linux, I guess is excellently maintained). Don't know how Windows 8 and higher do, but in my experience Windows isn't exactly the most memory efficient OS around.

But what I *do* know is that, with only 4GB RAM, I never got FO to successfully build on Windows and only with difficulties by employing some tricks on OSX. So 6GB is probably the minimum, at least on these platforms.

> I would rather drop the specific memory size statement or remove the section completely.

Why? RAM is a significant limiting factor, exactly the thing I'd expect to see in sections dealing with hardware requirements. When it comes to building FO, it's actually the only one which you can possibly not have enough of.

adrianbroher: Fixed.

BUILD.md
adrianbroher: Fixed.

BUILD.md
adrianbroher: Fixed

BUILD.md
adrianbroher: Added a note for the `--jobs` parameter.  Please check out if this fits with what you expect.

Issue #2029, opened 2018-03-05
Map Window "Wasted Production" Wasted Calculation is Wrong


agrrr3:
The window currently shows:
- Output
- Used
- Excess
- To Stockpile
- Wasted

With the stockpile transfer project the PP for the project are not counted as Excess, but are counted for To Stockpile, so Wasted can be a negative value.

Expected:
The wasted value should be correctly calculated.

agrrr3:
Actually this problem is different from what i thought.

Playing with unlimited input waste is always zero. If we allow limited input then waste makes sense, but you also need to restrict stockpile transfer projects. So it would be probably more sensible to drop the limited input rule.

The again the question must be what should be shown and what not, so I give this to a design discussion at [honing the imperial stockpile design](http://www.freeorion.org/forum/viewtopic.php?f=5&t=10579&p=90999#p90999).

PR #2066, opened 2018-04-06
No negative waste map wnd

agrrr3:
1. Fixes calculating resource allocation for stockpile transfer project when input limits are enabled.
2. fixes waste calculation for stockpile transfer projects in waste resource browse window 

See https://github.com/freeorion/freeorion/issues/2029

### DIFF
--- a/Empire/ProductionQueue.cpp
+++ b/Empire/ProductionQueue.cpp
@@ -577,6 +577,7 @@ std::string ProductionQueue::Element::Dump() const {
 ProductionQueue::ProductionQueue(int empire_id) :
     m_projects_in_progress(0),
     m_expected_new_stockpile_amount(0),
+    m_expected_transfer_to_stockpile(0),
     m_empire_id(empire_id)
 {}
 
@@ -763,6 +764,7 @@ void ProductionQueue::Update() {
         m_empire_id, pp_in_stockpile, available_pp, m_object_group_allocated_pp,
         m_object_group_allocated_stockpile_pp);
     m_expected_new_stockpile_amount += transfer_to_stockpile;
+    m_expected_transfer_to_stockpile = transfer_to_stockpile;
 
     // if at least one resource-sharing system group have available PP, simulate
     // future turns to predict when build items will be finished
--- a/Empire/ProductionQueue.cpp
+++ b/Empire/ProductionQueue.cpp
@@ -100,7 +100,7 @@ namespace {
         return retval;
     }
 
-    float CalculateNewStockpile(int empire_id, float starting_stockpile, float stockpile_transfer,
+    float CalculateNewStockpile(int empire_id, float starting_stockpile, float project_transfer_to_stockpile,
                                 const std::map<std::set<int>, float>& available_pp,
                                 const std::map<std::set<int>, float>& allocated_pp,
                                 const std::map<std::set<int>, float>& allocated_stockpile_pp)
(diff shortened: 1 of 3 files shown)


agrrr3:
One question: does the version somewhere need to be bumped for serialization?

Dilvish-fo:
I'd first ask, does the expected transfer to stockpile really need to be serialized at all?  Can't the client just calculate that as the difference between the actual current stockpile and the expected new stockpile amount?

agrrr3:
I think the client needs to know the difference between PP transferred to
stockpile via excess vs stockpile transfer projects.

Maybe the client could recalculate that transferred amount by adding up all
PP allocated to transfer projects. But I don't know where you would do that
code-wise.

Am I correct that the serialization only necessary for client/server
transfer and when loading saved games, the PP allocation gets recalculated
automatically?

2018-04-09 18:47 GMT+02:00 Dilvish-fo <notifications@github.com>:

Dilvish-fo:
Ah, having looked at the code here more now, it looks like that by "transfer to stockpile" you are only referring to the transfer from stockpile projects, is that correct?  If so, I can see that it couldn't be calculated exactly the way I proposed, but instead would need to be tallied they way you mention above (as is done in the code here).  But I think the respective variables should be renamed to be more clearly specific to stockpile transfer projects and not for all transfer into the stockpile.  

The client recalculates (its estimate of) the allocations at the start of the turn (via MapWnd InitTurn which calls Empire::UpdateResourcePools() which calls Empire::UpdateProductionQueue()) and anytime the production queue is modified.   The server just does it once, shortly before actually applying everything.  The part that only the server does, to actually spend the PP, is Empire::CheckProductionProgress().   

Calls to  Empire::UpdateProductionQueue(), which in turn calls m_production_queue.Update(), are done both by the server and by the client, and so that is a fine place to have the transfer calculations being done, which seems consistent with the approach you've taken here.   So I would also still think there is no need for serialization.

agrrr3:
I reverted the serialization. Load/Save and everything else still fine.

Dilvish-fo:
>  and everything else still fine.

Well, there is still my concern/comment about " the respective variables should be renamed to be more clearly specific to stockpile transfer projects and not for all transfer into the stockpile."   It seems to me that the over-general variable names is part of what led to my initial confusion about what was going on.

It looks to me that at least for the variables referencing "transfer_to_stockpile" they are actually specific about being for "project_transfer_to_stockpile" (or somesuch)-- do you have an objection to a more specific name like that?

agrrr3:
@Dilvish-fo  any more suggestions? i think this is done

Empire/ProductionQueue.cpp
```diff
@@ -100,7 +100,7 @@ namespace {
         return retval;
     }
 
-    float CalculateNewStockpile(int empire_id, float starting_stockpile,
+    float CalculateNewStockpile(int empire_id, float starting_stockpile, float stockpile_transfer,
```
Dilvish-fo: please also rename this ```stockpile_transfer``` to ```project_transfer_to_stockpile```

Empire/ProductionQueue.cpp
```diff
@@ -129,9 +129,9 @@ namespace {
                           << "  to stockpile: " << new_contributions;
         }
 
-        if (new_contributions > stockpile_limit &&
+        if ((new_contributions + stockpile_transfer) > stockpile_limit &&
             GetGameRules().Get<bool>("RULE_STOCKPILE_IMPORT_LIMITED"))
-        { new_contributions = stockpile_limit; }
+        { new_contributions = stockpile_limit - stockpile_transfer; }
 
         return starting_stockpile + new_contributions - stockpile_used;
```
Dilvish-fo: I think it would be slightly cleaner (and would avoid having this function seem to contain inconsistencies) if we had this return line be 
```
return starting_stockpile + new_contributions + stockpile_transfer - stockpile_used;
```
or after my requested variable rename, 
```
return starting_stockpile + new_contributions + project_transfer_to_stockpile - stockpile_used;
```

and simply did away with line 775
```
 m_expected_new_stockpile_amount += project_transfer_to_stockpile;
```

Or am I overlooking some trouble that would cause somewhere?

agrrr3: Sounds  good

PR #2070, opened 2018-04-08
Ai military and invasion stealth adjustments

Dilvish-fo:
This third PR primarily builds on the stealth/detection code added in #2069 and applies that in some military and invasion contexts-- having the AI abort Invasion missions (and also colony and outpost missions) if it no longer expects it can detect the target, and then some adjustments aimed at releasing military ships that currently can have a tendency to get stuck over a stealthed target that had at one point been visible to them.

### DIFF
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -2,7 +2,9 @@ from logging import debug, info, warn
 
 import freeOrionAIInterface as fo  # pylint: disable=import-error
 
-from fleet_orders import AIFleetOrder, OrderMove, OrderOutpost, OrderColonize, OrderMilitary, OrderInvade, OrderDefend
+# the following import is used for type hinting, which pylint seems not to recognize
+from fleet_orders import AIFleetOrder  # pylint: disable=unused-import
+from fleet_orders import OrderMove, OrderOutpost, OrderColonize, OrderMilitary, OrderInvade, OrderDefend
 import AIstate
 import EspionageAI
 import FleetUtilsAI
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -1,9 +1,10 @@
-from logging import info, warn
+from logging import debug, info, warn
 
 import freeOrionAIInterface as fo  # pylint: disable=import-error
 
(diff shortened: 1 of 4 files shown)


Dilvish-fo:
(rebased and force-pushed after fixing the PEP8 issue in the underlying first PR)

Morlic-fo:
That error is not an artifact. It happens during the unit tests. The unit tests currently import python/tests/freeOrionAIInterface.py which do not define planetSize.

Could be fixed by using and patching the autogenerated stub file (in tests/conftest.py) instead.

Dilvish-fo:
Rebased.

It's the commits starting with "Use CombatRatingsAI" that really make up this PR as opposed to 2069.  In playtesting I note that the commits about helping end combat missions over unowned stealthed planets and release ships for repair do seem to be doing at least their portion of the task properly, but I have seen the ships still sometimes seeming to be stuck there because the move routing code was apparently having more trouble than I think it should have in finding them a safe path back to a drydock.  I'll try to continue looking at that some more, but I do think it is a distinct issue that doesn't need to hold up these commits.

Dilvish-fo:
I still don't get why those unit tests are only seeming to run (and fail) for this PR and not the previous ones.  But whatever, I've added planetSize to the mock interface now, pushed that to master and also added it to this PR to clear the error here because it aggravates me 😃

Dilvish-fo:
OK, well although it looks like starType was the only remaining enumeration used in AIDependencies.py that was not yet added to the mock interface file, I've gone ahead and added all the remaining exposed enumerations to it, to hopefully clear up this issue for good.

Dilvish-fo:
rebased, to hopefully clear this unit test failure

Morlic-fo:
#2071 Should resolve the issue more cleanly without requiring to manually patch the interface.

The band-aid hackfix that made the module work during unit tests didn't work anymore as you introduced an import dependency to AIDependencies by adding EspionageAI to fleet_orders (which is why prior unit tests did not fail).

Dilvish-fo:
Good to know why this popped up now, thanks for the explanation.  I've already manually patched the mock interface, but 2071 still looks worthwhile.

Dilvish-fo:
rebased to resolve conflicts

Dilvish-fo:
I'm going to proceed with merging, but I'll still be happy to address any further comments that arise.  (And plan to continue some related work anyway.)

PR #2071, opened 2018-04-11
Improve savegame_codec behavior during unit test

Morlic-fo:
With #2070, the hacky fix for the import errors during unit tests showed its limits.

The revised version detects if the module is imported during a unit test and does not try to import AI modules in that case.

@Vezzra This is a change in the AI code base but strictly related to and only impacting unit tests (which would seem to imply infrastructure label). Not sure how to label stuff like this, so leaving it blank for now...

### DIFF
--- a/default/python/AI/savegame_codec/_definitions.py
+++ b/default/python/AI/savegame_codec/_definitions.py
@@ -1,5 +1,10 @@
 # a list of trusted classes - other classes will not be loaded
-try:
+import sys
+if 'pytest' in sys.modules:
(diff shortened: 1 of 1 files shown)


Dilvish-fo (approved):

Vezzra:
I guess we can probably give good reasons pro/contra the one or the other approach. Unless anyone objects, with changes primarily or even solely to the codebase, I'll go ahead and label this as "component:AI". The actual "infrastructure" components are the unit tests itself, after all.

PR #2072, opened 2018-04-13
Rework stockpiling techs

agrrr3:
Design discussion [What to do with the stockpiling techs](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10932).

* **add population boni which dont need focus**
* **restrict setting the focus to capital and homeworlds**
* **added stockpiling center building to entanglement tech (10% of industry)**
* fix some bugs

Did revert the following as it was controversial
* slow down stockpiling meter growth if stockpile focus is not set (so growth focus has slow growth)
* boost stockpiling meter growth if stockpile focus is set 

So instead 
* **de-/increase of stockpile meters is exactly like for research/industry**

todo
* prune this after playtesting feedback
* documentation/stringtables
* add a better icon for stockpiling center


Dilvish-fo:
In a programming context, I think that "refactor" has a rather specific meaning, different than how you are using it here in the PR title and commit message-- I think there here the more generic "restructure" or something like that would be much more appropriate.

geoffthemedio:
Redesign or rework seem more suitable.

Dilvish-fo:
I haven't yet done any testing, nor done a detailed comparison between all the corresponding Industry/Research scripting and this, but just from reviewing this, it is looking OK, with one possible exception:  it looks to me like one or more additional BLD_STOCKPILING_CENTERs could be built at a former Capital that has been captured with one or more already there, built by previous empires, by then redesignating the captured planet as the new Capital (i.e., building new Palace), after which you could then build one additional BLD_STOCKPILING_CENTER (and it appears the effects would currently stack).  That's because you have both ownedBy and ProducedByEmpire as requirements in the location/enqueue exclusions.   I'm open to discussing that as an intentional benefit from capturing a Capital, but in that case it should be commented as intended and probably mentioned in the Pedia as well.

If that's not the intention, then I'd suggest you drop the ProducedByEmpire clause

Dilvish-fo:
Ah, yes, I had overlooked that CaptureResult clause, but I do think it's is still better with this last change.

agrrr3:
@geoffthemedio As it seems it is impossible to get people to test branches and give feedback. Can we fasttrack this (and other 4.8 PRs) please to get people to use it via the test builds?

geoffthemedio:
I'm find with merging such PRs, but often @Dilvish-fo has AI-related objections to doing so... so up to him.

agrrr3:
Always forget about AI, seems to be my blind spot...

Dilvish-fo:
I'm fine with this one going in right away, and following up with AI adjustments sometime a bit later (and I'll try to get at least some rudimentary AI adjustments in before the next weekly build)

PR #2073, opened 2018-04-17
Add pedia tag to replicon

matt474:
Added a tag on the Replicon species so that it will now be placed in the robotic category in the pedia.


geoffthemedio:
looks reasonable

PR #2077, opened 2018-04-24
AI detection using empire meter

Dilvish-fo:
This PR has the AI determine empire detection strengths from the actual empire meter rather than by estimating based on empire techs & observed specials.  Note that empire meters were just recently exposed to python, if you do not compile FO you will need to use the latest weekly build in order to test this PR.

This PR also then has the AI focus on the detection strengths of only the empires that are locally active near a respective target colony site rather than having the AI always consider all empires.  This area could probably benefit from some further work (like keeping a longer memory of enemy fleets active in area, and slightly extending the range of consideration for enemy supply), but seems to work well enough for now.

@Cjkjvfnby @Morlic-fo my recollection about how to trigger the inspection handler to automatically update the interface stub file was a bit fuzzy, and I didn't find a writeup on it, and rather than spend time experimenting with that I wound up just manually adding the new empire.getMeter exposure to the interface stub file. If either of you could point me to instructions for that (or write some) I'd appreciate that.


default/python/AI/AIstate.py
```diff
@@ -154,7 +154,7 @@ def __init__(self, aggression):
         self.newlySplitFleets = {}
         self.militaryRating = 0
         self.shipCount = 4
-        self.misc = {}
+        self.misc = {}  # Keys: "observed_enemies"
```
Morlic-fo: At the very least, there is the 'enemies_sighted' key as well.

default/python/AI/EspionageAI.py
```diff
@@ -3,38 +3,46 @@
...
-                        if fo.currentTurn() < 40 else AIDependencies.DETECTION_TECH_STRENGTHS["SPY_DETECT_2"])
-    empire = None
-    if empire_id != ALL_EMPIRES:
-        empire = fo.getEmpire(empire_id)
+    return (AIDependencies.DETECTION_TECH_STRENGTHS["SPY_DETECT_1"] if fo.currentTurn() < 40
+            else AIDependencies.DETECTION_TECH_STRENGTHS["SPY_DETECT_2"])
+
+
+@cache_by_session_with_turnwise_update
+def get_empire_detection(empire_id):
```
Morlic-fo, line 17: Please add short docstring re behavior if passed ALL_EMPIRES

default/python/AI/EspionageAI.py
```diff
@@ -3,38 +3,46 @@
...
-                break
-        if empire_id == fo.empireID() and state.have_panopticon:
-            empire_detection += AIDependencies.PANOPTICON_DETECTION_BONUS
-    return empire_detection
-
-
-def colony_detectable_by_empire(planet_id, species_name=None, empire_id=ALL_EMPIRES,
-                                future_stealth_bonus=0, default_result=True):
+        return empire.getMeter(EmpireMeters.DETECTION_STRENGTH).initial
+    else:
```
Morlic-fo, line 32: Should probably log a warning here?

default/python/AI/EspionageAI.py
```diff
@@ -3,38 +3,46 @@
...
+        return default_empire_detection_strength()
+
+
+def get_max_empire_detection(empire_list):
```
Morlic-fo: Please add docstring which states that the own empire is ignored here.

default/python/AI/EspionageAI.py
```diff
@@ -3,38 +3,46 @@
...
+                                default_result=True):
     """
     Predicts if a planet/colony is/will-be detectable by an empire
     :param planet_id: required, the planet of concern
     :type planet_id: int
     :param species_name: will override the existing planet species if provided
     :type species_name: str
-    :param empire_id: empire whose detection ability is of concern
-    :type empire_id: int
+    :param empire: empire (or list of empires) whose detection ability is of concern
```
Morlic-fo: Should probably more precisely state empire_id in the comment as parameter name is now somewhat ambiguous and could refer to an actual empire object.

Morlic-fo (approved):
Untested but the code changes look fine. I added a few remarks where documentation is currently lacking.

Morlic-fo:
> If either of you could point me to instructions for that (or write some) I'd appreciate that.

There used to be some instructions for that. Not sure where those are now or what happened to them.

Anyway:
1) Start Freeorion with --ai-config pointing to https://github.com/freeorion/freeorion/blob/master/default/python/handlers/inspect_interface_config.ini

2) Start single player game with a single AI

3) Game session will crash once the interface file was created

default/python/AI/AIstate.py
Cjkjvfnby: I prefer to store them as separate variables:  `observed_enemies=set()` . It is simple and clear when you read the code.

default/python/AI/AIstate.py
```diff
@@ -400,6 +400,7 @@ def __update_system_status(self):
         supply_unobstructed_systems = set(empire.supplyUnobstructedSystems)
         min_hidden_attack = 4
         min_hidden_health = 8
+        observed_enemies = self.misc.setdefault("observed_enemies", set())
```
Cjkjvfnby: When you introduce new persistent collection, you can add some documentation about it. 

- what object does it contain
- when object added to it
- when object removed from it

default/python/AI/EspionageAI.py
```diff
@@ -45,7 +53,10 @@ def colony_detectable_by_empire(planet_id, species_name=None, empire_id=ALL_EMPI
     # The future_stealth_bonus can be used if the AI knows it has researched techs that would grant a stealth bonus to
     # the planet once it was colonized/captured
 
-    empire_detection = get_empire_detection(empire_id)
+    if isinstance(empire, int):
+        empire_detection = get_empire_detection(empire)
```
Cjkjvfnby: you can use `get_max_empire_detection([impire])` and make `get_empire_detection` private

Cjkjvfnby: Please move this code close to its usage.

default/python/AI/EspionageAI.py
```diff
@@ -54,7 +65,7 @@ def colony_detectable_by_empire(planet_id, species_name=None, empire_id=ALL_EMPI
         species_name = planet.speciesName
 
     # in case we are checking about aborting a colonization mission
-    if empire_id == fo.empireID() and planet.ownedBy(empire_id):
+    if empire == fo.empireID() and planet.ownedBy(empire):
```
Cjkjvfnby, line 91: This is an issue:
`1 == fo.empireID()` -> True
`[1] == fo.empireID()` -> False

Cjkjvfnby (changes requested):

default/python/AI/AIstate.py
Dilvish-fo: > At the very least, there is the 'enemies_sighted' key as well.

I've now added the other keys, and also renamed this new one to observed_empires, to clarify the distinction from 'enemies_sighted'

> I prefer to store them as separate variables: observed_enemies=set() . It is simple and clear when you read the code.

Isn't that just a tradeoff of complexity in different places, at least to avoid making savegames incompatible, because if you make it an attribute like that then you need to update the version and ```__setstate__``` method?  I think the current treatment is quite clear, and actually simpler.

genestack-solomatin: > at least to avoid making savegames incompatible

How many lines of code you need to add to __setstate__ method to remove redundant complexity?

- `setdefault` does not make code more readable.  You do some lazy initialization inside business logic, instead of putting it to logicless initialization code. 
- IDE type hinting does not work well with dict of type `str: object`

default/python/AI/EspionageAI.py
Dilvish-fo: > Please move this code close to its usage.

OK, moved

> you can use get_max_empire_detection([impire]) and make get_empire_detection private

While that could be possible with some reworking of things, as has now been clarified in the docstrings, ```get_max_empire_detection()``` excludes the current AI empire-- you could think of it as ```get_max_non_self_empire_detection()``` but that seems to me an unnecessarily cumbersome name.   So things seem simpler/better to me as-is.

default/python/AI/EspionageAI.py
Dilvish-fo: As has now been clarified in the docstrings, there isn't any use-case for intentionally passing a list containing only the current empire ID.  The only way that could even come up accidentally is in some sort of test multiplayer game in which the only empire is an AI empire, and in that case we would want the test to fall through as it currently does.

default/python/AI/AIstate.py
Morlic-fo: I have to agree with Cj here. In general, I would prefer storing variables directly. If there are things which need to be grouped, a class or a namedtuple helps both with type hinting and error checks. 

Using dicts with hardcoded string keys is kind of dangerous especially when missing those keys is somewhat expected: It is extremely hard to track down a bug based on a wrong ```get``` call because we just use the default value and no error/warning will ever be thrown. 

For example ```observed_enemies``` vs ```enemies_sighted``` - who would notice the error if I used ```sighted_enemies``` instead which is grammatically correct and in line with the other naming scheme.

If we want to use dicts, setter/getter functions should be used to avoid these problems (and provide type hinting).

Also, the ```__setstate__``` method has to be changed once and will never have to be looked at again (and even if we notice a few versions later that there was some hard to fix savegame compatibility bug, we could simply raise the minimum required version). The attribute itself is worked with as long as it exists.

default/python/AI/EspionageAI.py
Cjkjvfnby: > As has now been clarified in the docstrings, there isn't any use-case for intentionally passing a list containing only the current empire ID.

And passing int as current empire id is valid use case?  Docstring is invalid, since it was not updated after argument type was changed.

My complain is that int as argument and list with single argument with same int have different behavior.  It is not clear and it is not described in docstring.   When I see `int | list[int]` type, I assume that it is shortcut to remove unnecessary list wrapping:   `x(n) == x([n]) `.

Dilvish-fo: > And passing int as current empire id is valid use case? 

There is a fair bit specifically about that in both the docstring and some in a code comment.  I had added a fair bit of that specifically in response to your and Morlic's earlier comments, as I pretty clearly indicated in my comment above.  Are you sure you have read the current code?

> Docstring is invalid, since it was not updated after argument type was changed.

While some sort of relative-timing concern like that might be good reason to examine something closely, it's not much of an argument for asserting something is invalid.   I rather suspect that you simply misread/misinterpreted something.  

The docstring does in fact look valid to me.  If you wish to assert it is invalid, please be quite specific about what you think is invalid and why it would actually be invalid (not simply why you are suspicious).

> My complain is that int as argument and list with single argument with same int have different behavior. It is not clear and it is not described in docstring. 

To the contrary, it is described in the docstring, quite explicitly (and to my eyes, clearly).  Were you somehow still looking at an old page of code even after I said I explained all this in the docstring?

>When I see int | list[int] type, I assume that it is shortcut to remove unnecessary list wrapping: x(n) == x([n]).

I think I must suggest you reread the docstring, and beware of over-reliance on assumptions.

PR #2078, opened 2018-04-24
Fleet order adjustments part 1

Dilvish-fo:
This PR includes some general cleanup of AI treatment of order issuance and execution tracking, along with a number of changes directed at improving AI exploration and a bit for AI repair.

I am thinking it might be helpful to restructure some of the existing order types to subclass from this new OrderPause, and also plan some other changes to hopefully help facilitate maneuvers like a military fleet Pounce or somesuch, but that is still rather nascent and would follow in a separate PR.  As it stands I think this PR makes some minor improvements, and at least my testing has not indicated it causing any trouble.

### DIFF
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -5,7 +5,7 @@ import freeOrionAIInterface as fo  # pylint: disable=import-error
 import FreeOrionAI as foAI
 from EnumsAI import MissionType, ShipRoleType
 import CombatRatingsAI
-from universe_object import Planet
+from universe_object import Planet, Fleet
 from ShipDesignAI import get_part_type
 from AIDependencies import INVALID_ID
 import AIDependencies
@@ -621,3 +621,33 @@ def get_fleet_system(fleet):
     if isinstance(fleet, int):
         fleet = fo.getUniverse().getFleet(fleet)
     return fleet.systemID if fleet.systemID != INVALID_ID else fleet.nextSystemID
+
+
+def get_current_and_max_structure(fleet):
+    """Return a 2-tuple of the sums of structure and maxStructure meters of all ships in the fleet
+
+    :param fleet:
+    :type fleet: int | universe_object.Fleet | fo.Fleet
+    :return: tuple of sums of structure and maxStructure meters of all ships in the fleet
+    :rtype: (float, float)
+    """
+
+    universe = fo.getUniverse()
+    destroyed_ids = universe.destroyedObjectIDs(fo.empireID())
+    if isinstance(fleet, int):
+        fleet = universe.getFleet(fleet)
+    elif isinstance(fleet, Fleet):
+        fleet = fleet.get_object()
+    if not fleet:
+        return (0.0, 0.0)
+    ships_cur_health = 0
+    ships_max_health = 0
+    for ship_id in fleet.shipIDs:
+        # Check if we have see this ship get destroyed in a different fleet since the last time we saw the subject fleet
+        if ship_id in destroyed_ids:
+            continue
+        this_ship = universe.getShip(ship_id)
+        ships_cur_health += this_ship.initialMeterValue(fo.meterType.structure)
+        ships_max_health += this_ship.initialMeterValue(fo.meterType.maxStructure)
+
+    return ships_cur_health, ships_max_health
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -644,9 +644,14 @@ def get_current_and_max_structure(fleet):
     ships_max_health = 0
     for ship_id in fleet.shipIDs:
         # Check if we have see this ship get destroyed in a different fleet since the last time we saw the subject fleet
+        # this may be redundant with the new fleet assignment check made below, but for its limited scope it may be more
+        # reliable, in that it does not rely on any particular handling of post-destruction stats
         if ship_id in destroyed_ids:
             continue
         this_ship = universe.getShip(ship_id)
+        # check that the ship has not been seen in a new fleet since this current fleet was last observed
+        if not (this_ship and this_ship.fleetID == fleet.id):
+            continue
         ships_cur_health += this_ship.initialMeterValue(fo.meterType.structure)
         ships_max_health += this_ship.initialMeterValue(fo.meterType.maxStructure)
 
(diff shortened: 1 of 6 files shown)


default/python/AI/FleetUtilsAI.py
```diff
@@ -621,3 +621,33 @@ def get_fleet_system(fleet):
...
+    if isinstance(fleet, int):
+        fleet = universe.getFleet(fleet)
+    elif isinstance(fleet, Fleet):
+        fleet = fleet.get_object()
+    if not fleet:
+        return (0.0, 0.0)
+    ships_cur_health = 0
+    ships_max_health = 0
+    for ship_id in fleet.shipIDs:
+        # Check if we have see this ship get destroyed in a different fleet since the last time we saw the subject fleet
```
Morlic-fo, line 646: What happens if we saw the ship in another fleet (but it was not destroyed yet)?

default/python/AI/fleet_orders.py
```diff
@@ -540,9 +574,6 @@ def issue_order(self):
...
             print "Order issued: %s fleet: %s target: %s" % (self.ORDER_NAME, self.fleet, self.target)
-
-        if system_id == fleet.systemID:
-            if foAI.foAIstate.get_fleet_role(fleet_id) == MissionType.EXPLORATION:
-                if system_id in foAI.foAIstate.needsEmergencyExploration:
-                    foAI.foAIstate.needsEmergencyExploration.remove(system_id)
-            self.order_issued = True
+        ships_cur_health, ships_max_health = FleetUtilsAI.get_current_and_max_structure(fleet_id)
+        self.executed = (ships_cur_health == ships_max_health)
+        return True
```
Morlic-fo, line 578: Do we rely on the fact that we end up in the target system after an OrderRepair? Otherwise, we could consider the mission finished as soon as we reach full health even if we are not at the target system (e.g. due to innate repair ability of the hull)

Morlic-fo (approved):
Looks like a good improvement and clears things up that were previously rather weird re the usage of executed/issued in fleet orders.

default/python/AI/fleet_orders.py
```diff
@@ -264,29 +285,25 @@ def is_valid(self):
...
         system_id = self.target.get_system().id
         fleet = self.fleet.get_object()
-        if system_id not in [fleet.systemID, fleet.nextSystemID]:
-            # if self.order_type == AIFleetOrderType.ORDER_MOVE:
-            #     dest_id = system_id
-            # else:
-            # if self.order_type == AIFleetOrderType.ORDER_REPAIR:
-            #     fo.issueAggressionOrder(fleet_id, False)
+        if system_id == fleet.systemID:
+            if foAI.foAIstate.get_fleet_role(fleet_id) == MissionType.EXPLORATION:
```
Cjkjvfnby, line 293: Can you give an example of role when target system  will not be explored?

default/python/AI/FleetUtilsAI.py
Dilvish-fo: Good point.   I think that (ages ago) when I first wrote that destruction check, there was still some bugginess with contained-object updates.  I think that's all resolved now, and so a couple lines below I've added a fleet-assignment check, which makes this destruction check *mostly* redundant (or perhaps entirely redundant).  Since the destruction check isn't dependent on any particular handling of post-destruction stats, though, I think it is a bit more reliable, plus  we still have a small bit of quirkiness with combat-granted visibility, so I'd rather still leave in this destruction check.

There are probably also a number of places where we are evaluating enemy fleets that may not be currently visible, and in any such situation a similar screening of the ships should be made, we'll have to keep our eyes out for such.

default/python/AI/fleet_orders.py
Dilvish-fo: It's not clear to me if you are trying to raise a concern, or make a suggestion.  As of this PR we *do* consider the mission finished as soon as we reach full health even if we are not yet in the target system, and that is intentional.  Do you have some concern about that?

default/python/AI/fleet_orders.py
Dilvish-fo: Although any fleet type being in (or merely passing through) the system is sufficient to toggle the 'explored' flag, the 'needsEmergencyExploration' list is reserved for more particular scouting needs, generally related to unexpected visibility trouble.  Since a common source of unexpected visibility trouble is reduced visibility range, which is more likely to be an issue, or be a bigger issue, for a non-scouting fleet.  It's my recollection that this was especially common back when ion storms reduced visibility range but did not affect stealth, and I think that for a while (perhaps even still?) reducing detection range to zero would block visibility even at the same location.  It's possible that a more complex assessment could be done instead, but given as the whole 'needsEmergencyExploration' tactic is a bit more of a "Hail Mary" attempt in the face of unexpected failure rather than a precisely calculated tactic, I think the simple requirement of a scouting fleet being present is still the right rule.

default/python/AI/fleet_orders.py
Morlic-fo: No, I wanted to suggest exactly that behavior and seemingly misunderstood the code. 

What threw me off is that we will call ```fo.issueFleetMoveOrder``` even if we are already at full health. I suppose that if we were to queue another order, it would overwrite the move order. And we would consider the next queued order because the order is marked as both issued and executed.

default/python/AI/fleet_orders.py
```diff
@@ -527,11 +556,16 @@ def is_valid(self):
...
         system_id = self.target.get_system().id
-        fleet = self.fleet.get_object()
-        if system_id not in [fleet.systemID, fleet.nextSystemID]:
+        fleet = self.fleet.get_object()  # type: fo.fleet
+        if system_id == fleet.systemID:
+            if foAI.foAIstate.get_fleet_role(fleet_id) == MissionType.EXPLORATION:
+                if system_id in foAI.foAIstate.needsEmergencyExploration:
+                    foAI.foAIstate.needsEmergencyExploration.remove(system_id)
+        elif system_id != fleet.nextSystemID:
+            self.executed = False
```
Morlic-fo: This assignment is what made me confused about the behavior of the function. It suggests that the order is only executed after moving to the target system. However, it will always be overwritten at the end of function.

Can we remove this or is this necessary for logging purposes (please add a short comment in that case)?

Dilvish-fo: > Can we remove this

Done.  I overlooked cleaning that up when I added the superseding health check.

PR #2082, opened 2018-05-02
Fix AI exploration

Morlic-fo:
This PR fixes a bug where the AI would (sometimes?) not assign new exploration targets while it has existing exploration targets but idle/available scout fleets.

While the fix is straightforward (a list of universe_object.System objects was returned rather than a list of system ids), I don't understand how the relevant list comprehension failed in the way it does:
```
needs_coverage = [sys_id for sys_id in check_list if sys_id not in already_covered and sys_id != INVALID_ID]
```

Before the fix:
22:02:04.479107 [debug] python : ExplorationAI.py:52 - Explorable system IDs: [210, 237, 192, 174]
22:02:04.479107 [debug] python : ExplorationAI.py:53 - Already targeted: [<universe_object.System object at 0x03D952D0>]
22:02:04.479107 [debug] python : ExplorationAI.py:62 - Check list: [210, 237, 192, 174]
22:02:04.479107 [debug] python : ExplorationAI.py:64 - Needs coverage: [] 

After the fix:
22:09:12.186570 [debug] python : ExplorationAI.py:52 - Explorable system IDs: [210, 237, 192, 174]
22:09:12.186570 [debug] python : ExplorationAI.py:53 - Already targeted: [237]
22:09:12.186570 [debug] python : ExplorationAI.py:62 - Check list: [210, 237, 192, 174]
22:09:12.186570 [debug] python : ExplorationAI.py:64 - Needs coverage: [210, 192, 174]

### DIFF
--- a/default/python/AI/ExplorationAI.py
+++ b/default/python/AI/ExplorationAI.py
@@ -1,4 +1,4 @@
-from logging import error
+from logging import debug, error
 
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client # pylint: disable=import-error
 import FreeOrionAI as foAI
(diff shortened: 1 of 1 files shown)


Dilvish-fo (approved):
Good catch.  I too am failing to see just how that list interpretation had been producing the results you post here, but the fix certainly looks correct.

PR #2083, opened 2018-05-02
Improve AI scout assignment logic

Morlic-fo:
Previously, the scout - exploration target was pretty much random. This PR improves the algorithm by a simple heuristic to match scouts to nearby exploration targets. Basic algorithm:

1) Compute and store all possible scout-system pairings and their (jump) distance
2) As long as there are possible assignments left, choose the one with shortest jump distance

This is by no means optimal but it is an improvement over the existing logic. A more sophisticated algorithm could certainly lead to better results (e.g. use better heuristic or exact optimization, consider future assignments for scouts which are available in the next few turns, ...). However, this simple approach already seems to work "sufficiently well" in practice.

~The PR is based on bug fix #2082.~

### DIFF
--- a/default/python/AI/ExplorationAI.py
+++ b/default/python/AI/ExplorationAI.py
@@ -71,69 +71,55 @@ def assign_scouts_to_explore_systems():
 
     available_scouts = set(available_scouts)
     sent_list = []
-    while available_scouts and needs_coverage:
-        this_sys_id = needs_coverage.pop(0)
-        sys_status = foAI.foAIstate.systemStatus.setdefault(this_sys_id, {})
-        if this_sys_id not in explore_list:  # doesn't necessarily need direct visit
-            if universe.getVisibility(this_sys_id, fo.empireID()) >= fo.visibility.partial:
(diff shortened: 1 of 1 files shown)


Morlic-fo:
Rebased.

Dilvish-fo (approved):
Looks good.

FYI, I had planning on doing some more work in this area myself, to maximize scout safety and the benefits from Derelict Scout specials by prioritizing visiting systems that have been partially visible at any time (just not actually visited) and further prioritizing systems that do not have any asteroids.  But I'll be able to leverage the work you've done here via some refactoring.  Or if you want to go ahead and do that, please be my guest (just let me know).

Morlic-fo:
I have no immediate plans for further work on ExplorationAI. My main motivation was the horrible early game movement by the AI. It actually looks slightly competent now at that stage so I am happy with the result and will leave it at that. 

Your suggestions sound like good further improvements though (and possibly candidates for AI the character module - prefering safe or risky but possibly faster exploration).

PR #2084, opened 2018-05-03
Improve EspionageAI planetary stealth logic

Morlic-fo:
This PR fixes some issues with the current planetary stealth prediction used by the AI:
1) Log a warning if the predicted planet stealth does not match the actual meter value to find unsupported stealth effects and to verify the calculation
~2) Bugfix: base stealth stacks with stealth specials~
~3) Make the AI aware of supertester stealth - now it won't try (and fail) to invade their planets~
4) Catch some of the discrepencies between predicted planet stealth and meter value due to ion storms by checking if the planet is in a (known) ion storm. However, the code does not work if the AI can not see the ion storm itself (because the center is not in detection range?) - looking for suggestions for a better approach here so it may actually be used in the AI logic.

### DIFF
--- a/default/python/AI/EspionageAI.py
+++ b/default/python/AI/EspionageAI.py
@@ -86,6 +86,9 @@ def colony_detectable_by_empire(planet_id, species_name=None, empire=ALL_EMPIRES
         return default_result
     if species_name is None:
         species_name = planet.speciesName
+        verify_prediction = True
+    else:
+        verify_prediction = False
 
     # in case we are checking about aborting a colonization mission
     if empire == fo.empireID() and planet.ownedBy(empire):
@@ -113,6 +116,12 @@ def colony_detectable_by_empire(planet_id, species_name=None, empire=ALL_EMPIRES
 
     total_stealth = planet_stealth + sum([AIDependencies.STEALTH_STRENGTHS_BY_SPECIES_TAG.get(tag, 0)
                                           for tag in species_tags])
+
+    if verify_prediction:
+        meter_stealth = planet.currentMeterValue(fo.meterType.stealth)
+        if meter_stealth != total_stealth:
+            warn("Predicted stealth value for planet %s of %.1f but got %.1f" % (planet, total_stealth, meter_stealth))
+
     if isinstance(empire, int):
         empire_detection = get_empire_detection(empire)
     else:
--- a/default/python/AI/EspionageAI.py
+++ b/default/python/AI/EspionageAI.py
@@ -4,7 +4,7 @@ import freeOrionAIInterface as fo  # pylint: disable=import-error
 import AIDependencies
 from AIDependencies import ALL_EMPIRES
 from EnumsAI import EmpireMeters
-from freeorion_tools import cache_by_session_with_turnwise_update
+from freeorion_tools import cache_by_session_with_turnwise_update, object_is_in_ion_storm
 
 
 def default_empire_detection_strength():
@@ -119,6 +119,9 @@ def colony_detectable_by_empire(planet_id, species_name=None, empire=ALL_EMPIRES
 
     if verify_prediction:
         meter_stealth = planet.currentMeterValue(fo.meterType.stealth)
+        predicted_stealth = total_stealth
+        if object_is_in_ion_storm(planet_id):
+            predicted_stealth += AIDependencies.ION_STORM_STEALTH
         if meter_stealth != total_stealth:
             warn("Predicted stealth value for planet %s of %.1f but got %.1f" % (planet, total_stealth, meter_stealth))
 
--- a/default/python/AI/EspionageAI.py
+++ b/default/python/AI/EspionageAI.py
@@ -122,8 +122,8 @@ def colony_detectable_by_empire(planet_id, species_name=None, empire=ALL_EMPIRES
         predicted_stealth = total_stealth
         if object_is_in_ion_storm(planet_id):
             predicted_stealth += AIDependencies.ION_STORM_STEALTH
-        if meter_stealth != total_stealth:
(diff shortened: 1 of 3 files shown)


default/python/AI/EspionageAI.py
```diff
@@ -112,6 +116,15 @@ def colony_detectable_by_empire(planet_id, species_name=None, empire=ALL_EMPIRES
 
     total_stealth = planet_stealth + sum([AIDependencies.STEALTH_STRENGTHS_BY_SPECIES_TAG.get(tag, 0)
                                           for tag in species_tags])
+
+    if verify_prediction:
```
Dilvish-fo, line 120: This whole verify_prediction section would need a number of additional adjustments and constraints on when it would consider logging a discrepancy, and even with those it's not really clear to me just when we could actually get a useful signal-to-noise ratio from this.

Although we are not currently actually using the ```future_stealth_bonus``` parameter, I plan to add support for AIs to pursue planetary stealth techs in the relatively near future, and then would use ```future_stealth_bonus``` to take such expected stealth bonus into account.   So before updating the planet_stealth (above) with ```future_stealth_bonus```, we could save ```prediction_assessment_value = planet_stealth``` and then do our tests here with respect to ```prediction_assessment_value``` rather than ```planet_stealth```

Also, anytime we have just Basic Visibility of an object, its stealth meter will be set to Meter::LARGE_VALUE ( 2^15 ), and would only be updated thereafter if we get Partial Visibility of it, so that established a large range of cases in which we would expect a mismatch between the predicted stealth and the meter reading and should not warn for.

But also if we had momentarily gotten Partial Visibility from an Effect like from a Derelict Scout (or on the turn that a stealthy species colonized the planet), and have not after that moment had any visibility of the object at all, then our current meter reading would still be that from the momentary glimpse of Partial Visibility (even if that is now stale and even if our current detection strength is below that level of stealth).

default/python/AI/EspionageAI.py
```diff
@@ -112,6 +116,15 @@ def colony_detectable_by_empire(planet_id, species_name=None, empire=ALL_EMPIRES
...
+        meter_stealth = planet.currentMeterValue(fo.meterType.stealth)
+        predicted_stealth = total_stealth
+        if object_is_in_ion_storm(planet_id):
```
Dilvish-fo, line 123: The ion storm prediction is a rather complex issue-- we could know that an ion storm currently covers a planet but our last sighting of the planet (and corresponding meter readings) might or might not predate such.  Also, as a general matter we'll want to eventually do a really good job of predicting ion storm coverage (i.e. projecting trajectory from last observation), increasing the number of times when we might predict a planet to be currently covered by storm even if it was not at the last time we saw it.

So until such time as we got very confident of our ion storm predictions, I would probably suggest only using them to eliminate cases where we might otherwise warn about a prediction mismatch, not making a construction where a false prediction of ion storm coverage would trigger a warning (in that case, a debug statement at most, but I would even suggest against that for now, I think it would be too much noise).

default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -347,3 +348,22 @@ def wrapper(*args, **kwargs):
 
         return wrapper
     return decorator
+
+
+def object_is_in_ion_storm(object_id):
+    # FIXME: This isn't working if we can not detect the ion storm
+    #        as is the case if the center of the storm is out of detection range
+    universe = fo.getUniverse()
+    universe_object = universe.getPlanet(object_id)
```
Dilvish-fo, line 357: If we want to restrict the function to only dealing with planets, then we should rename it.  But it seems better to me to change this to  
```
universe_object =  universe.getObject(object_id)
```

default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -347,3 +348,22 @@ def wrapper(*args, **kwargs):
...
+    if not universe_object:
+        return False
+
+    for field_id in universe.fieldIDs:
```
Dilvish-fo, line 361: Before checking the fields, I would check if we own an object in the subject-object's system, and if so then return an early answer based entirely on whether that self-owned-object is in a field (see if it has an ion field contribution to its stealth meter-- Effect Accounting for meters is directly exposed to python now although I can't recall having used it yet, and worst case we could do some reliable inference about our owned object).  Otherwise, we could make a wrong prediction (of either True or False) needlessly based on stale field object info.

default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -347,3 +348,22 @@ def wrapper(*args, **kwargs):
 
         return wrapper
     return decorator
+
+
+def object_is_in_ion_storm(object_id):
+    # FIXME: This isn't working if we can not detect the ion storm
```
Dilvish-fo, line 354: Although some warning  of uncertainty is suitable here, I think the situation is not as dire as the comment currently describes.  ```SetEmpireFieldVisibilitiesFromRanges()``` (in universe.cpp) grants Partial Visibility of a field if any of the empire's detection range simply touches the perimeter of the field, no need to directly see its center.

Perhaps also add a TODO about predicting field trajectories and size-- with a bit of work the AI should be able to make nearly perfect projections of such after a single sighting.

Dilvish-fo (commented):
Good catch on the stacking.  And I like the idea of starting to try to track and deal with ion storms.  The "verify" bit needs a fair bit of work though.

Morlic-fo:
Thanks for the review, Dilvish. I am not sure when I will find the time to fix the issues you raised.

I cherry-picked the bugfixing commits and pushed them to master directly. Setting status to Work in progress for the remaining part (ion storms and verification).

geoffthemedio:
@Morlic-fo reminder...

geoffthemedio:
@Morlic-fo Any chance you'll get back to this?

PR #2088, opened 2018-05-09
Allow player to revoke and change his order

o01eg:
This PR allows player to prevent turn advance and change his orders.
It introduces `UNREADY` message client sends to revoke his orders and server sends to acknowledge revoking.
In multiplayer game "Turn" button renames to "Revise Orders" and revokes turn orders.
The server stores ready status and unset it when this message accepted. Turn orders remains and could be saved.

Forum thread: http://www.freeorion.org/forum/viewtopic.php?f=6&t=10915

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -1041,7 +1041,10 @@ void MapWnd::CompleteConstruction() {
     // turn button
     // determine size from the text that will go into the button, using a test year string
     std::string turn_button_longest_reasonable_text =  boost::io::str(FlexibleFormat(UserString("MAP_BTN_TURN_UPDATE")) % "99999"); // it is unlikely a game will go over 100000 turns
-    m_btn_turn = Wnd::Create<CUIButton>(turn_button_longest_reasonable_text);
+    std::string unready_button_longest_reasonable_text =  boost::io::str(FlexibleFormat(UserString("MAP_BTN_TURN_UNREADY")) % "99999");
+    m_btn_turn = Wnd::Create<CUIButton>(turn_button_longest_reasonable_text.size() > unready_button_longest_reasonable_text.size() ?
+                                        turn_button_longest_reasonable_text :
+                                        unready_button_longest_reasonable_text);
     m_btn_turn->Resize(m_btn_turn->MinUsableSize());
     m_btn_turn->LeftClickedSignal.connect(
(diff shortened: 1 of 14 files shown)


geoffthemedio:
"Unready" as a button label is ambiguous. Does it mean that is the player's current state, or is it an action verb that clicking would enact? Maybe rename to "Revise Orders" or similar so it's clearly a verb?

o01eg:
@geoffthemedio Renamed the button.

o01eg:
Fixed issue with unneeded player list refreshing.

Vezzra:
While this is certainly a very nice feature to have, I'm a bit reluctant to make it optional for the release. For the sake of finally making progress with 0.4.8, I'd recommend to postpone merging this until after the release (that is, after the release branch has been created of course), if that's ok?

@o01eg, @geoffthemedio?

o01eg:
@geoffthemedio I prefer to get release sooner.

o01eg:
Rebased. Fixed conflicts.

Vezzra:
@geoffthemedio, as the release branch has been created, this (and "post 0.4.8" PRs in general) can be merged now...

Vezzra:
I understand, however, isn't the whole point of having a release branch so working on stuff in master can resume unhindered...?

Or is it just the changes in this particular PR that you are cautious about?

Dilvish-fo:
>  isn't the whole point of having a release branch so working on stuff in master can resume unhindered...?

I would suggest adding a "relatively" before "unhindered", although that still might not clear up the issue here 100% it seems.

> I'm hesitant to do that immediately, as it will make cherry picking more complicated.

Correct me if I'm wrong, but (outside of the very limited areas where we've made commits to 0.4.8 that are specific to it) cherry picking into 0.4.8 only gets more complicated if we have multiple commits in 0.4.8+ master that touch the same code in some fashion, but then for some reason choose to just cherry pick a later one of that set of interacting commits into 0.4.8, skipping over an earlier one of the interacting commits, which could make the cherrypicking need to resolve a conflict.  The potential for that kind of situation seems *mostly* predictable to me.

It is still possible that we could merge this PR, then discover that the underlying (but now partly modified) code had a bug that needs fixing both here and in 0.4.8, and the fix here might then need some adjustment during the cherrypicking process.  It doesn't really seem to me, though, that the risk/cost there is so high that we need to freeze all work (or this particular work) to avoid it.

PR #2090, opened 2018-05-09
Fix ShipPartsOwned

### LINKED ISSUE #2089: ShipPartsOwned class filter broken
* **FreeOrion Version**: ae02c26b15018c63b1d59977e838ae5d626ac43f
* **Operating System**: Debian sid
* **Graphic card used**: ASUS Nvidia GTX1080
* **Fetched as** 
  * Compiled from source

### Description
This upkeep macro fails:

```
FLEET_UPKEEP_MULTIPLICATOR
'''(ShipPartsOwned empire = Source.Owner class = ShortRange)'''
```

With approximately this error:
```
/opt/freeorion/share/freeorion/default/scripting/ship_hulls/SH_BASIC_MEDIUM.focs.txt:13:72: Parse error.  Expected ) here:
```

### Expected Result
macro works

### Steps to reproduce
replace FLEET_UPKEEP_MULTIPLICATOR in default/scripting/common/upkeep.macros with above, launch freeoriond

Voker57:
Switch order of by_name and by_class to allow by_class to be parsed

My attempt to fix #2089

### DIFF
--- a/parse/IntComplexValueRefParser.cpp
+++ b/parse/IntComplexValueRefParser.cpp
@@ -182,8 +182,8 @@ namespace parse {
             |   outposts_owned
             |   parts_in_ship_design
             |   part_class_in_ship_design
-            |   ship_parts_owned_by_name
             |   ship_parts_owned_by_class
+            |   ship_parts_owned_by_name
             |   empire_design_ref
             |   slots_in_hull
             |   slots_in_ship_design


Dilvish-fo:
It does make sense to me that this would work, because the ``` ship_parts_owned_by_class``` absolutely requires the Class token, whereas the ```ship_parts_owned_by_name``` leaves the name token as optional (presumably so that you could have a way to count all parts owned regardless of name or class).

As it stands, though, this seems like a rather fragile fix/workaround.

I would think that at a minimum we would also need to add comments before the definitions for those parsers and before the lines you swapped here, to make the significance of the order of processing very clear.  And to be more consistent with that it seems to me it would be slightly better if we also swapped the order they are defined, so that ``` ship_parts_owned_by_class``` is defined before ```ship_parts_owned_by_name```, and maybe even swap the order they are listed in ValueRefParser.h.  But perhaps this extra order swapping is just cosmetic overkill -- @geoffthemedio do you have a preference?

geoffthemedio:
Definition order doesn't matter.

What does is that this change would probably just break `ship_parts_owned_by_name` instead of `ship_parts_owned_by_class`.

The `>` indicates expectation, not requirement, so the first tested parser will always match `ShipPartsOwned` as the "by class" variant, and the "by name" variant won't every be matched. If `name =` is specified, or `class =` omitted, it would just result in a failed parse.

Disambiguation would require use of the sequence operator `>>`, and probably some grouping with `( ... )`.

geoffthemedio:
I'm not sure if or when I can fix this. With the changes @LGM-Doyle made to parsing, so there are no local variables used, but instead just series of results from the parts of the parser in series, any changes I make, that previously would have been fine, just produce massive Boost-error template spam for unclear reasons. When I try to put in a sequence operator `>>`, it just won't compile, and I haven't been able to find a workaround.

Vezzra:
> I'm not sure if or when I can fix this.

Well, ouch. And I guess trying to revert @LGM-Doyle's changes to the parser code now (several months after these changes went in) is even more messy/impossible than trying to fix them...?

Voker57:
Better solution implemented.

PR #2095, opened 2018-05-11
Use submatch length in Vartext for identify label existence

### LINKED ISSUE #2087: Sitreps full of errors or not shown
(depending on whether you have the UI setting enabled for showing sitreps with errors).

Occurs on current clean master, appears likely to stem from bb3033a800

adrianbroher:
Fixes #2087

### DIFF
--- a/util/VarText.cpp
+++ b/util/VarText.cpp
@@ -170,12 +170,13 @@ namespace {
         std::string operator()(xpr::smatch const& match) const {
             // Labelled variables have the form %tag:label%,  unlabelled are just %tag%
             std::string tag = match[1];
-            // For unlabelled variables use the tag value as label.
-            std::string label = match[(match.size() == 1) ? 1 : 2];
+            // Use the label value. When missing, use the tag submatch as label instead.
+            std::string label = match[match[2].matched ? 2 : 1];
 
             // look up child
             auto elem = m_variables.find(label);
             if (m_variables.end() == elem) {
+                ErrorLogger() << "Substitute::operator(): No value found for label: " << label << " from token: " << match.str();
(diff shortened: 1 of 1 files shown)


Vezzra (approved):
Fixed the issue for me (tested on OSX).

Dilvish-fo (commented):
The fix works for me (tested on Win10).

I am tempted to go ahead and merge, but a typo in the commit message is making me exercise patience.

"existance" --> "existence"

also, "for identify" --> "to identify"

If you do wind up modifying the commit message, it might also be worthwhile/educational to mention that the size_>length change was because match.size() does not yield the intuitively expected result.

adrianbroher:
Updated, also changed the code to use `sub_match::matched` instead of querying the match string length.

Dilvish-fo:
I think you might have overlooked an extra +1 twist to  match_results::size() that makes it particularly non-intuitive:

```
    /// Returns one plus the number of marked sub-expressions in the regular
    /// expression that was matched if *this represents the result of a
    /// successful match. Otherwise returns 0.
```

adrianbroher:
I didn't overlook that.  Every RegEx implementation I encountered so far references the full match with capture group 0. I won't document default behavior.

If you disagree you have to add the docs yourself, as I'm off for the next days.

Dilvish-fo:
OK.  Since you indicate you'll be off the next few days I'll go ahead and merge this.

geoffthemedio:
Even if it's default behaviour, if it's potentially confusing or unexpected to someone reading the FreeOrion code but not the RegEx documentation, a note in a comment with some explanation is reasonable, I think.

PR #2098, opened 2018-05-12
Productionqueue quantityselector fix

### LINKED ISSUE #2097: subclassed dropdown lists broken
I first noticed that the ProductionQueue blocksize/repeat quantity selectors were broken.  They'll appear to change values, but other than their label the production queue does not respond to the change, and upon advancing the turn the numbers revert, confirming that the attempted change was ineffective.

This was caused by 26d5b0fa924aabf8be3d26e6c21e56bb7b70c9a4, which made droplists respond to left-mouse press instead of click, but didn't make the corresponding change to the subclasses, such as  ```QuantitySelector : public CUIDropDownList``` in ProductionWnd.cpp.

The fix for the ProductionQueue (and probably the others as well) appears fairly straightforward, although the QuantitySelectors had been overriding both ```DropDownList::LClick```  and ```DropDownList::LButtonDown```; it seems to me that the former  QuantitySelector::LButtonDown code can just be jettisoned and essentially replaced with the former LClick code.  I have a basic fix for the ProductionQueue ready that does that, and most/all of the small bit of jettisoned code seems to be a useless remnant from some older code anyways, but it's been long enough since I've done anything with that that I'll put it up as a PR for some review/testing rather than just commit it directly.

Searching for other overrides via Git Bash, I found 
```
$ grep "public CUIDropDownList" */*.cpp */*/*.cpp
UI/MultiplayerLobbyWnd.cpp:    class TypeSelector : public CUIDropDownList {
UI/ProductionWnd.cpp:    class QuantitySelector : public CUIDropDownList {
UI/SidePanel.cpp:class SidePanel::SystemNameDropDownList : public CUIDropDownList {
```
And I confirmed that the SidePanel SystemNameDropDownList no longer seems to be working right (it won't show the dropdown list, although it will allow mousescrolling).  I haven't checked what the TypeSelector was yet and whether or not it was affected.  I don't have much more time available just now but will try to come back to it before Monday if no one else gets around to it.

Dilvish-fo:
As noted in #2097, some recent changes to DropDownList broke the ProductionQueue QuantitySelector used for blocksize & repeat.

Although the QuantitySelectors had been overriding both DropDownList::LClick and DropDownList::LButtonDown; it seems to me that the former QuantitySelector::LButtonDown code can just be jettisoned and essentially replaced with the former LClick code. Most/all of the small bit of jettisoned code seems to be a useless remnant from some older code anyways, but it's been long enough since I've done anything with that that I thought I should put it up as a PR for some review/testing rather than just commit it directly.

Resolves #2097

### DIFF
--- a/UI/ProductionWnd.cpp
+++ b/UI/ProductionWnd.cpp
@@ -115,7 +115,6 @@ namespace {
             blocksize(build.blocksize),
             prevBlocksize(build.blocksize),
             amBlockType(amBlockType_),
-            amOn(false),
             h(h_)
         {
             MoveTo(GG::Pt(xoffset, yoffset));
@@ -155,8 +154,6 @@ namespace {
         }
 
         void SelectionChanged(GG::DropDownList::iterator it) {
-            amOn = false;
-
             DebugLogger() << "QuantSelector:  selection made ";
             if (it != end()) {
                 int quant = boost::polymorphic_downcast<const QuantRow*>(it->get())->Quant();
@@ -177,16 +174,9 @@ namespace {
         int     blocksize;
         int     prevBlocksize;
         bool    amBlockType;
-        bool    amOn;
         GG::Y   h;
 
-        void LosingFocus() override {
-            amOn = false;
-            DropDownList::LosingFocus();
-        }
-
         void LButtonDown(const GG::Pt&  pt, GG::Flags<GG::ModKey> mod_keys) override {
-            amOn = !amOn;
             if (this->Disabled())
                 return;
 
--- a/UI/ProductionWnd.cpp
+++ b/UI/ProductionWnd.cpp
@@ -187,14 +187,10 @@ namespace {
 
         void LButtonDown(const GG::Pt&  pt, GG::Flags<GG::ModKey> mod_keys) override {
             amOn = !amOn;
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Seems to work fine with your changes. The check for if it's disabled is worth keeping might be worth keeping, though.

Dilvish-fo:
>  The check for if it's disabled is worth keeping might be worth keeping, though.

I thought about that, but DropDownList::LButtonDown also has the same ```if Disabled() return```, so we are simply left then with a (in that case) probably superfluous check that the quantities hadn't somehow changed, which seems computationally cheap and if there somehow was some mechanism in place where they nevertheless could have changed, then i decided the code could honor that.  That was my thought on it, anyways.

geoffthemedio:
I'm not sure what gets called when exactly, but the concern was that the issuing of orders, and thus enabled status of the droplist, could be changed between when it's dropped and when something in the dropped list is clicked. If it's disabled at the later time but remains open from before, it may or may not be appropriate to emit a signal... Maybe the code the signal is connected to handles this case by checking whether order issuing is currently enabled?

Dilvish-fo:
OK, I suppose it's a bit safer to keep that extra check for Disabled.

Issue #1794, opened 2017-10-07
AI exception when loading save game - assign_military_fleets_to_systems


dbenage-cx:
### Environment

* **FreeOrion Version**: 0.4.7+ f5e26bc
* **Operating System**: Fedora
* **Fetched as**  Compiled from source

### Description
On load of save game posted in #1778 (from turn 129):
```
AI_4: AI_Error: AI script error : FreeOrionAI.py:generateOrders():333  - Exception 114698 while trying to assign_military_fleets_to_systems
Traceback (most recent call last):
  File "/opt/projects/freeorion/default/python/AI/FreeOrionAI.py", line 330, in generateOrders
    action()
  File "/opt/projects/freeorion/default/python/AI/MilitaryAI.py", line 803, in assign_military_fleets_to_systems
    fleet_list=found_fleets)
  File "/opt/projects/freeorion/default/python/AI/FleetUtilsAI.py", line 150, in get_fleets_for_mission
    fleet_pool_set.remove(fleet_id)
KeyError: 114698
```

### Expected Result
No exception

dbenage-cx:
@freeorion/ai-team Following diff allows load of saves games without an AI error for me. Not certain if potential fix or temporary workaround.
```diff
diff --git a/default/python/AI/FleetUtilsAI.py b/default/python/AI/FleetUtilsAI.py
index 59b4462da..7b541d101 100644
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -147,7 +147,8 @@ def get_fleets_for_mission(target_stats, min_stats, cur_stats, starting_system,
 
             # all checks passed, add ship to selected fleets and update the stats
             fleet_list.append(fleet_id)
-            fleet_pool_set.remove(fleet_id)
+            if fleet_id in fleet_pool_set:
+                fleet_pool_set.remove(fleet_id)
             this_rating = foAI.foAIstate.get_rating(fleet_id)
             cur_stats['rating'] = CombatRatingsAI.combine_ratings(cur_stats.get('rating', 0), this_rating)
             if 'ratingVsPlanets' in target_stats:
```

Morlic-fo:
Your diff does silence the error but I do not think that is a very good idea to do that. The error is a hint that there is a significant problem in the underlying code either on server side or within the AI scripts assumptions.

I pushed a commit to catch the KeyError and hopefully let the AI more gracefully end its turn but it will still show an error message in the chat for now.

[...]

PR #2102, opened 2018-05-15
Do not let AI issue new orders after loading a game

### LINKED ISSUE #1837: When loading a game, the AI will replay its turn with different results
Usually, the AI will begin turn X with a given AIstate ```pre_turn_X = post_turn_(X-1)```. After it has played the turn, the AIstate will be altered (```post_turn_X```).

When saving the game, AIstate ```post_turn_X``` will be saved. When loading the game, the AI will subsequently play the turn again - however with AIstate ```post_turn_X``` rather than ```pre_turn_X```. This may lead to different decision-making.

Furthermore, some orders may be issued a second time - for an example, see #1836.

The two straightforward approaches to fix this are both unsatisfactory.

A) In ```generateOrders```, check if the AI already played this turn (check an entry in the AIstate). If so, do not replay it. 
Problem: This makes debugging work more complicated as we can not simply modify some code and reload the game to check if it fixed the behavior. We would need a savegame one turn before the issue occurs which may very well not exist.

B) Always save the AI savegame state before ordering issues (e.g. keep a deepcopy of the pre-turn state or better: keep a savegame string from last turn even if no savegame was requested).
Problem: Issued orders still exist in the savegame. When code is modified and other issues will be ordered, there may be a mismatch between the AIstate and the actual orders issued. This is unacceptable.

A possible solution could be
C) Do not save the AI orders in the savegame (i.e. only save human orders). Always let the AI replay its turn on load. Save the pre-turn AIstate.

Morlic-fo:
This PR suggests one way to resolve #1837.

By storing the last played turn in the AIstate, we can detect if we already issued orders for a given turn (i.e. after loading the game). By default, the AI will not try to issue orders a second time in that case but use the orders in the savegame instead.

As discussed in #1837, for debugging work it is often useful to be able to load a savegame and replay the turn to check the effect of code changes. For this purpose, the flag ```replay_turn_after_load``` can be set in the AI config file.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -13,7 +13,7 @@ import random
 import freeOrionAIInterface as fo  # interface used to interact with FreeOrion AI client  # pylint: disable=import-error
 
 
-from common.option_tools import parse_config, get_option_dict
+from common.option_tools import parse_config, get_option_dict, check_bool
 parse_config(fo.getOptionsDBOptionStr("ai-config"), fo.getUserConfigDir())
 
 from freeorion_tools import patch_interface
@@ -313,8 +313,8 @@ def generateOrders():  # pylint: disable=invalid-name
     # TODO: Consider adding an option to clear AI orders after load (must save AIstate at turn start then)
     if fo.currentTurn() == foAIstate.last_turn_played:
         info("The AIstate indicates that this turn was already played.")
-        if not get_option_dict().get('replay_turn_after_load', False):
-            info("Aborting order generation. Orders from savegame will still be issued.")
+        if not check_bool(get_option_dict().get('replay_turn_after_load', 'False')):
+            info("Aborting new order generation. Orders from savegame will still be issued.")
             try:
                 fo.doneTurn()
             except Exception as e:
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -232,7 +232,6 @@ def generateOrders():  # pylint: disable=invalid-name
     """Called once per turn to tell the Python AI to generate and issue orders to control its empire.
     at end of this function, fo.doneTurn() should be called to indicate to the client that orders are finished
     and can be sent to the server for processing."""
-
     rules = fo.getGameRules()
     print "Defined game rules:"
     for rule in rules.getRulesAsStrings:
@@ -305,6 +304,25 @@ def generateOrders():  # pylint: disable=invalid-name
     print "***************************************************************************"
     print "***************************************************************************"
 
+    # When loading a savegame, the AI will already have issued orders for this turn.
+    # To avoid duplicate orders, generally try not to replay turns. However, for debugging
+    # purposes it is often useful to replay the turn and observe varying results after
+    # code changes. Set the replay_after_load flag in the AI config to let the AI issue
+    # new orders after a game load. Note that the orders from the original savegame are
+    # still being issued and the AIstate was saved after those orders were issued.
+    # TODO: Consider adding an option to clear AI orders after load (must save AIstate at turn start then)
+    if fo.currentTurn() == foAIstate.last_turn_played:
+        info("The AIstate indicates that this turn was already played.")
+        if not get_option_dict().get('replay_turn_after_load', False):
+            info("Aborting order generation. Orders from savegame will still be issued.")
+            try:
+                fo.doneTurn()
+            except Exception as e:
+                error("Exception %s while trying doneTurn()" % e, exc_info=True)
+            return
+        else:
+            info("Issuing new orders anyway.")
+
     if turn == 1:
         declare_war_on_all()
         human_player = fo.empirePlayerID(1)
@@ -345,6 +363,8 @@ def generateOrders():  # pylint: disable=invalid-name
         fo.doneTurn()
     except Exception as e:
         error("Exception %s while trying doneTurn()" % e, exc_info=True)  # TODO move it to cycle above
+    finally:
+        foAIstate.last_turn_played = fo.currentTurn()
 
     if using_statprof:
         try:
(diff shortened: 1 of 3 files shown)


Dilvish-fo:
Here's my initial thoughts--  did you get around to looking into  my request from that Issue, 
> Would you doublecheck if the use of the turn-begin autosave avoids this [primary] problem?

(I'll reiterate something I stated there, though, that even if the turn-begin autosave can solve the primary problem, there can still be some value to other handling like this.)

If the answer to that question is yes, then I would think that to add some real value this PR would need to include the discard-previous-orders aspect mentioned in the comments (either done automatically when replaying, or possibly subject to an additional option).

If the answer to that question was no (which would somewhat surprise me), then I suppose I'd see this add value even without the discard-previous-orders treatment, although I think I'd still prefer the discard to at least be an an option (and I'd probably want to understand better just why the turn-begin autosave did not actually solve the primary problem).

Dilvish-fo:
Hmm, and it just struck me, I guess we don't currently have client-side support for just discarding a client's current orders (at least not that I saw).  Although orders, at least some, have some ability to be undone. I'm not sure they all do, and that would probably be a messy process.  We'd probably need to be able to split out the ```Orders().ApplyOrders();``` step from AIClientApp.cpp to be separately triggered by the AI upon a savegame load (so the AI could simply not trigger them on its universe copy if it was going to discard them).  @geoffthemedio  any objections to that?

Also,  the AI changes its own state about fleets and ships as it joins and splits those, so it would need to keep a copy of its own 'turn-start-state' to revert to in the even that it discarded orders. 

It seems to me that during the steps of splitting fleets and whatnot, the client might request new object IDs from the server, and if those respective orders get thrown out then those object IDs would be wasted, but no actual harm should come from that.  @geoffthemedio does that sound right to you? 

that's all the extra issues that has come to mind so far...

Dilvish-fo:
Hmm, looking into the bit about getting new object IDs for splitting fleets, it is looking to me (and seeming familiar) that we have changed that process so that there is no longer an advance request for the server to generate the object ID -- the client and server now each generate IDs as they process orders, done in a way so that they stay in sync but without letting any client know which other client is generating which other IDs.  So just more reason not to need to worry about throwing away orders that had involved generating object IDs.

Morlic-fo:
> Would you doublecheck if the use of the turn-begin autosave avoids this [primary] problem?

Sorry for not mentioning this. Autosaves at beginning of the turn still seem to be *after* the AI python scripts. When I load a beginning of the turn autosave with this PR, I will see at the logs that the AI already played its turn.

Now that you mention it again, I forgot to doublecheck if the AI orders are actually also in that savegame or if there is some weird behavior where the AIstate is from after the turn but the orders are empty. I should probably also double-check what happens later in the game when AI turn take in the order of some seconds and if that influences the behavior (Is the AIClient even multithreaded?)

> Also, the AI changes its own state about fleets and ships as it joins and splits those, so it would need to keep a copy of its own 'turn-start-state' to revert to in the even that it discarded orders.

Yes,  we discussed this as option and consequence in the original issue. I figured that while the preferably solution would be clearing the issued orders (possibly optionally), it seems quite more complex and error-prone. What I came up with in this PR seemed simple and efficient enough as a stop-gap solution which could maybe even be included in the next release. Note I added a TODO in the relevant code pointing out the cleaner solution.

Unless you think we can get up the cleaner solution soon, I think the solution in this PR is better than the current state and easily reverted if something better comes along (but as mentioned above, I'll need to verify the correct behavior with autosaves at turn start more thoroughly).

Dilvish-fo (changes requested):
Looks pretty good to me, just a couple minor-ish requested changes

Dilvish-fo:
> Autosaves at beginning of the turn still seem to be after the AI python scripts. When I load a beginning of the turn autosave with this PR, I will see at the logs that the AI already played its turn.

OK, thanks.  Looking into this, it appears to me that the 'turn start' autosave is triggered by the Human Client when it receives the TurnUpdate, which will have also at the same time triggered the AIs to ```generateOrders()```.  I think I have an idea on a modest change to this process which would allow the turn-start savegame to use the AIState prior to the AI order generation, which would be helpful for our development/debugging process.  I've opened Issue #2108 for that and will try to get on that soon.

default/python/AI/FreeOrionAI.py
```diff
@@ -305,6 +304,25 @@ def generateOrders():  # pylint: disable=invalid-name
...
+    # To avoid duplicate orders, generally try not to replay turns. However, for debugging
+    # purposes it is often useful to replay the turn and observe varying results after
+    # code changes. Set the replay_after_load flag in the AI config to let the AI issue
+    # new orders after a game load. Note that the orders from the original savegame are
+    # still being issued and the AIstate was saved after those orders were issued.
+    # TODO: Consider adding an option to clear AI orders after load (must save AIstate at turn start then)
+    if fo.currentTurn() == foAIstate.last_turn_played:
+        info("The AIstate indicates that this turn was already played.")
+        if not get_option_dict().get('replay_turn_after_load', False):
+            info("Aborting order generation. Orders from savegame will still be issued.")
```
Dilvish-fo: Please add 'new' after 'Aborting'
Somehow this got left out of my review.

Morlic-fo:
Fixed the missing check_bool call and missing "new" in logging statement.

Dilvish-fo (approved):

Dilvish-fo:
I think we should cherry pick this for 0.4.8.  Although we've downgraded #1794 to optional after changing the related logging to warn level for the 0.4.8 branch, there is still some underlying bugginess going on, and the changes of this PR should make it so that by default those are at least reduced (and perhaps avoided entirely).

Vezzra:
> I think we should cherry pick this for 0.4.8.

Yep, it's a bugfix, as such it should go into 0.4.8. Done.

PR #2103, opened 2018-05-15
Convert config values read by python to correct types

Morlic-fo:
The standard library (Safe)ConfigParser will return strings. This leads to unintuitive (buggy) behavior as bool('0') or bool('False') both return True (as both are non-empty strings). So a config like ```some_option=False``` would not actually disable that option.

The simple solution is to convert the values to the "correct" python types using ```ast.literal_eval```. Note that the function 
> can be used for safely evaluating strings containing Python values from untrusted sources 
(https://docs.python.org/2/library/ast.html#ast.literal_eval)

### DIFF
--- a/default/python/common/option_tools.py
+++ b/default/python/common/option_tools.py
@@ -1,3 +1,4 @@
+import ast
 import os
 import sys
 from ConfigParser import SafeConfigParser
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
So, looking into this a bit, it seems to me like our handling is fairly consistent with the standards for config files; and for options that we expect to be booleans then ```some_option=true``` and ```some_option=false``` would both be fine.  And actually  ```some_option=False``` would happen to work also, but  ```some_option=True``` would not.   But accepting ```True``` and ```False``` does not appear to be standard for python config files even when expecting to deal with bools, and so it does not immediately seem like a bug that we would not support those.  If there was someplace that we were automatically generating config files that we could not properly read back in, then I would agree that's a bug.

We had implemented a helper function option_tools.check_bool() for handling known (at least expected) boolean option values.  It uses the same string values for ```True``` as does [RawConfigParser.getboolean()](https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.getboolean), but it is permissive on the ```False``` side rather than generating ValueErrors.

So, @Morlic-fo  if there is someplace we were autogenerating config files that use ```True``` and ```False``` or are otherwise failing to deal with bools properly, then I could see that issue being a bug with the mandatory milestone, otherwise it would be an optional grade feature/tweak proposal.  To my current understanding we fall in the latter category, and I would plan to change this to the optional milestone unless details are cited placing us in the first category. 

And regardless, it would certainly be fine to add some more documentation 

Now, even if there is some bug with the existing handling, we should have some discussion about whether this is really the best way to handle it, and conversely, even if there is no existing bug we could still talk about whether we'd want to make this extension.

Morlic-fo:
> I would plan to change this to the optional milestone

Sure.

Dilvish-fo:
Actually, taking a look at the code, it appears to me that ```check_bool``` already does a lower() operation before checking the values, and so it appears to me that no changes should really be necessary.  I think it's simply a matter of knowing to use ```check_bool()```.

Morlic-fo:
> I think it's simply a matter of knowing to use check_bool()

Yeah, I think I shouldn't open and fix issues at or after midnight.

PR #2104, opened 2018-05-15
Download MacOSX FreeOrionSDK during build environment setup

adrianbroher:
Download and unpack the FreeOrionSDK on MacOSX when creating a build
environment via CMake.  The FreeOrionSDK is downloaded only when the
archive does not exist and unpacked only when the unpacked version
doesn't match the configured version in the CMakeLists.txt


Vezzra (approved):
Works as intended on my system.

Issue #2111, opened 2018-05-18
Release 0.4.8 preparation


Vezzra:
This issue is used to organize the version 0.4.8 release of FreeOrion.

### Milestones

Important milestones for the release.  Changes will be announced by the release
manager.

| Date             | Milestone                         |
|:---------------- | --------------------------------- |
| May 20th 2018, 6pm | Create release branch from master |
| June 14th, 1pm | Release Candidate 1               |
| July 13th, 1pm | Release Candidate 2               |
| ~~July 19th, 1pm~~ | ~~Release Candidate 3~~               |
| August 23rd 2018, 1pm | Release Candidate 3               |
| ~~June 28th, 1pm~~ | ~~Declaring Release~~                 |
| ~~July 22nd, 1pm~~ | ~~Declaring Release~~                 |
| ~~July 23rd, 1pm~~ | ~~Declaring Release~~                 |
| August 30th 2018, 1pm | Declaring Release                 |

*All times are UTC.*

### Checklist

This is a checklist for the release manager to ensure all steps are done before
doing the actual release.

#### Preparations

* [x] Create a new GitHub Issue from `.github/RELEASE_TEMPLATE.md` and follow
      instructions.
* [x] Determine initial milestones and dates for release.
* [x] Create release branch from development branch master by using:
```
git checkout -b release-v0.4.8 master
```
* [x] *On release branch HEAD* - Update version number in `CMakeLists.txt` and
      `cmake/make_versioncpp.py` to `v0.4.8`, remove trailing `+` if needed.
* [x] *On release branch HEAD* - Disable `Super Testers` species by default.
* [x] *On master branch HEAD* - Update version number in `CMakeLists.txt` and
      `cmake/make_versioncpp.py` to `v0.4.8+`.
* [x] Push updated release and master branch to GitHub.
* [x] Update `ChangeLog.md`.

#### Release candidate 1

* [x] Integrate incoming bugfixes.
* [x] Update `ChangeLog.md` again if needed.
* [x] *On release branch HEAD* - Tag the release candidate by using:
```
git tag --annotate --message="0.4.8 Release Candidate 1" v0.4.8-rc1
```
* [x] Push release candidate tag to GitHub.
* [x] Build release candidate artifacts.
* [x] Upload release candidate artifacts to SourceForge.
* [x] Announce availability of release candidates in `News` section
       of FreeOrion Wiki.
* [x] Announce release candidate on Twitter as @FreeOrion
* [x] Test release candidate.

[...]

PR #2113, opened 2018-05-18
Streamline release procedure slightly to remove duplicative step

Dilvish-fo:
When reviewing our new release issue #2111, It appeared to me that since we do the 'create release branch' step before the 'Update version number' step, we need to duplicate the 'Update version number' step, which we wouldn't need to do if we just reorganized the steps slightly.

I think this also reduces the risk of someone else pushing changes to master while the person creating the release branch is in the middle of steps that would be thrown off by other commits to master.

It's a minor thing, but this seems a minor improvement to me which I couldn't resist proposing...

### DIFF
--- a/.github/RELEASE_TEMPLATE.md
+++ b/.github/RELEASE_TEMPLATE.md
@@ -45,16 +45,15 @@ doing the actual release.
 * [ ] Create a new GitHub Issue from `.github/RELEASE_TEMPLATE.md` and follow
       instructions. :wink:
 * [ ] Determine initial milestones and dates for release.
+* [ ] *On master branch HEAD* - Update version number in `CMakeLists.txt` and
+      `cmake/make_versioncpp.py` to `vX.Y.Z+`.
+* [ ] Push updated master branch to GitHub.
 * [ ] Create release branch from development branch master by using:
 ```
 git checkout -b release-vX.Y.Z master
 ```
-* [ ] *On release branch HEAD* - Update version number in `CMakeLists.txt` and
-      `cmake/make_versioncpp.py` to `vX.Y.Z`, remove trailing `+` if needed.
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
Hmmph. From the "Details":

> ** BUILD INTERRUPTED **

I presume that is simply a time-out?  It's making me wish we could restrict the build stages so that if all of the current changes were just to certain folders (like the scripting folders, or to .github) that the compile stages would be skipped.  But I suppose that would require maintaining a degree of history knowledge that might not really be supported.

I also note that even though we are using ccache, the compile time here seems rather tremendously long for something that ccache should really have been able to do in under a minute.  Is the problem that we have lots of disparate PRs that both tie down to significantly different points in master and also have enough header file changes to force recompilation of nearly all of FO, and then we throw out all of the other info each time because we are only allowing our ccache 200M of space?

I don't see anything in the Travis instructions for ccache saying that we need to pare down the size at all from the default, let alone so severely.  @adrianbroher I presume you have some reason, but is paring it all the way down to 200M really necessary?

Also, I found [some indication that Travis' instructions for enabling ccache for XCode are actually incomplete/wrong, but that a fix is possible](http://www.brianlheim.com/2017/08/20/supercollider-travisci.html).

Vezzra (commented):
> When reviewing our new release issue #2111, It appeared to me that since we do the 'create release branch' step before the 'Update version number' step, we need to duplicate the 'Update version number' step, which we wouldn't need to do if we just reorganized the steps slightly.

Actually, the update version number step is done deliberately after the create release branch step, because the version numbers on the master and release branches after the creation of the latter are different: on master, it is `vX.Y.Z+`, on the release branch it's `vX.Y.Z` (without the trailing '+').

So things should stay as they are.

PR #2115, opened 2018-05-20
Fix stockpile transfer description

agrrr3:
* fixed some bad information
** Wrong/misleading: Anything enqueued below the transfer project at a supply-connected planet will not be allocated PP.
* added more (hopefully) useful info

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -339,14 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
 Transfers PP into the imperial stockpile
 
 PROJECT_BT_STOCKPILE_DESC
-'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. The amount to transfer and how often the transfer should occur can be specified. PP spent on the transfer are added to the imperial stockpile.
+'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. The amount to transfer and how many times the transfer should repeat can be specified. PP allocated to the transfer are added to the imperial stockpile.
 
-Only uses PP that are produced in the same supply-group. No PP from the imperial stockpile are used.
+Only PP that are produced in the same supply-group are allocated to the stockpile transfer project. PP already in the imperial stockpile are not allocated to the stockpile transfer project.
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP used for stockpile transfer are not available for items on the build queue below it.
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP allocated to stockpile transfer are transferred to the stockpile, and not allocated to items lower in the production queue.
 
-For example: An empire wants to draw PP from the stockpile in a secondary supply-group. For this the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group and its stockpile use limit is 10 PP. To make sure that 10 PP are always available in the stockpile, one adds to the top of build queue a 10x Stockpile Transfer project on a planet in the primary supply-group.
-The Stockpile Transfer project is above other primary supply-group projects in the build queue. So it is fully funded and will increase the stockpile by 10 PP next turn. In order for a lasting effect one sets the repetition to 99 times.'''
+For example: An empire wants to draw PP from the stockpile in a secondary supply-group. For this the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group, its stockpile use limit is 10 PP and wants that 10 PP are always available in the stockpile. For this, the empire adds to the top of the queue a 10x Stockpile Transfer project on a planet in the primary supply-group.
+Because the Stockpile Transfer project is above other primary supply-group projects in the build queue, it is fully funded and will increase the stockpile by 10 PP next turn. To ensure a continuing transfer each turn, the repetition of the project is set to 99 times.'''
 
 
 ##
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -343,11 +343,11 @@ PROJECT_BT_STOCKPILE_DESC
 
 Only PP that are produced in the same supply-group (and none from the imperial stockpile) are allocated to the stockpile transfer project.
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP allocated to a stockpile transfer are not allocated to items lower in the production queue.
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP allocated to a stockpile transfer are sent to the stockpile, even if there are additional items later in the production queue.
 
-For example: An empire wants to draw PP from the stockpile in a secondary supply-group, in which there is insufficient PP being generated. To do this the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group, its stockpile use limit is 10 PP and wants that 10 PP are always available in the stockpile. The build queue contains items on planets in the primary supply-group which need 120 PP per turn, so all available PP would be used.
+For example: An empire wants to draw PP from the stockpile in a secondary supply-group, in which there is insufficient PP being generated. To do this, the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group, its stockpile use limit is 10 PP and it wants to have 10 PP always available in the stockpile. The build queue contains items on planets in the primary supply-group which need 120 PP per turn, so all available PP would be used.
 To be able to use the PP in the secondary supply-group, the empire adds to the top of the queue a 10x Stockpile Transfer project on a planet in the primary supply-group.
-Because the Stockpile Transfer project is above other primary supply-group projects in the build queue, it is fully funded and will increase the stockpile by 10 PP next turn. To ensure a continuing transfer each turn, the repetition of the project is set to 99 times.'''
+Because the Stockpile Transfer project is above other primary supply-group projects in the build queue, it is fully funded and will increase the stockpile by 10 PP on the next turn. To ensure a continuing transfer each turn, the repetition of the project is set to 99 times.'''
 
 
 ##
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -339,13 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
 Transfers PP into the imperial stockpile
 
 PROJECT_BT_STOCKPILE_DESC
-'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. The amount to transfer and how many times the transfer should repeat can be specified. PP allocated to the transfer are added to the imperial stockpile.
+'''PP allocated to the transfer project are added to the imperial stockpile. The amount to transfer and how many times the transfer should repeat can be specified.
 
-Only PP that are produced in the same supply-group are allocated to the stockpile transfer project. PP already in the imperial stockpile are not allocated to the stockpile transfer project.
+Only PP that are produced in the same supply-group (and none from the imperial stockpile) are allocated to the stockpile transfer project.
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP allocated to stockpile transfer are transferred to the stockpile, and not allocated to items lower in the production queue.
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP allocated to a stockpile transfer are not allocated to items lower in the production queue.
 
-For example: An empire wants to draw PP from the stockpile in a secondary supply-group. For this the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group, its stockpile use limit is 10 PP and wants that 10 PP are always available in the stockpile. For this, the empire adds to the top of the queue a 10x Stockpile Transfer project on a planet in the primary supply-group.
+For example: An empire wants to draw PP from the stockpile in a secondary supply-group, in which there is insufficient PP being generated. To do this the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group, its stockpile use limit is 10 PP and wants that 10 PP are always available in the stockpile. The build queue contains items on planets in the primary supply-group which need 120 PP per turn, so all available PP would be used.
+To be able to use the PP in the secondary supply-group, the empire adds to the top of the queue a 10x Stockpile Transfer project on a planet in the primary supply-group.
 Because the Stockpile Transfer project is above other primary supply-group projects in the build queue, it is fully funded and will increase the stockpile by 10 PP next turn. To ensure a continuing transfer each turn, the repetition of the project is set to 99 times.'''
 
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -339,14 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
 Transfers PP into the imperial stockpile
 
 PROJECT_BT_STOCKPILE_DESC
-'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. One can specify the amount and repetition of the transfer. PP spent on the transfer are added to the imperial stockpile.
+'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. The amount to transfer and how often the transfer should occur can be specified. PP spent on the transfer are added to the imperial stockpile.
 
 Only uses PP that are produced in the same supply-group. No PP from the imperial stockpile are used.
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used for controlling PP allocation to the stockpile. PP used for stockpile transfer are not available for items on the build queue below it.
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP used for stockpile transfer are not available for items on the build queue below it.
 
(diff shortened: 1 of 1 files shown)


default/stringtables/en.txt
```diff
@@ -339,11 +339,13 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
 Transfers PP into the imperial stockpile
 
 PROJECT_BT_STOCKPILE_DESC
-'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. Anything enqueued below the transfer project at a supply-connected planet will not be allocated PP.
+'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. You can specify the amount and repetition of the transfer. PP spent on the transfer are added to the imperial stockpile.
```
geoffthemedio: Do not refer to the player in the second person, eg. "you"

default/stringtables/en.txt
```diff
@@ -339,11 +339,13 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
 
-Only transfers PP that are available at this planet; PP allocation on planets that are not supply-connected to here will not be affected by this project.
+Only transfers PP that are available at this planet so only PP produced in the same supply-group are used and no PP from the imperial stockpile.
```
geoffthemedio: I don't see why your version is better...?

default/stringtables/en.txt
```diff
@@ -339,11 +339,13 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used to transfer PP to the stockpile even if there are things on the queue below it that could have been allocated available PP.'''
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue and may trigger a warning.  This project is used for fine-tuning allocation of PP to the stockpile even if there are items on the build queue below it that could have been allocated available PP.
```
geoffthemedio: mentioning a warning seems unnecessary, and is awkwardly phrased / placed at the end of this sentence.

"for fine-tuning" -> "controlling"

comma after "stockpile" before "even"

default/stringtables/en.txt
```diff
@@ -339,11 +339,13 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
+
+A fully funded transfer will be finished in a single turn so if you want to make sure you always have at least 10PP available in the stockpile, you can add a 10x Stockpile Transfer and set the repetition to 99 times.'''
```
geoffthemedio: This sentence is unclear. I suggests that having a 10x Stockpile Transfer will make at least 10 PP available, which is not clearly different from just always transferring up to 10 PP, and it doesn't clarify that it would only work if there is at least 10 PP available on the queue at that position. If you want to give some examples, clearly mark it as such, and give more details about the whole queue.

default/stringtables/en.txt
Dilvish-fo: The two versions seem like they mostly differ simply in emphasis, or what aspects are made most explicit.  After reading them multiple times I think I'd have a slightly preference for the newer one, which clarifies that the IS cannot itself be used to give PP to a transfer project.

geoffthemedio: phrasing needs to be improved then.

Dilvish-fo: @agrrr3 I'd suggest changing "and no PP from the imperial stockpile" to "(and transfer projects cannot be enabled to themselves draw on the Imperial Stockpile, since that would be a pointless cycle.)"

geoffthemedio: The "so" and the "and" before "no PP" are awkward. Split into 3 sentences? @Dilvish-fo's version is overly wordy.

default/stringtables/en.txt
agrrr3: I tried to make the example more clear. Probably the entry should be moved to another pedia article as there is some overflow even though only added one example. And i would like to add another example:

Another example: One has only one supply-group but wants to make sure that 5PP are always available in the stockpile for future use. Lets say one can draw more than 5PP from the stockpile. One can either disable stockpile use for all build items after adding once 5PP to the stockpile. Or one can add a 5x transfer project to the top of the build queue with a 99 times repetition.
This way always 5PP are added to the next turn's imperial stockpile while other projects use up the current stockpile.

Ouaz:
Maybe I'm wrong, but:

As now "_Unused PP will automatically be transferred to the Imperial Production Stockpile_", the Production Window article (PRODUCTION_WINDOW_ARTICLE_TEXT) should be updated accordingly:

`Generally, most excess PP will be wasted and lost, but a part of the excess will be stored in the imperial stockpile.`

to

`Excess PP will be stored in the imperial stockpile. `

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used to transfer PP to the stockpile even if there are things on the queue below it that could have been allocated available PP.'''
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used for controlling PP allocation to the stockpile. PP used for stockpile transfer are not available for items on the build queue below it.
+
+An example: One wants to draw PP from the stockpile in a secondary supply-group. For this the stockpile needs to be filled from the primary supply-group. Lets say one has 100PP industry in the primary supply-group and is able to draw 10PP from the stockpile. For making sure that 10PP are always available in the stockpile one adds to the top of build queue a 10x Stockpile Transfer project on a planet in the primary supply-group.
```
geoffthemedio: space between number and PP
don't use hokey "let's say" (and if you do, use an apostrophe for the contraction). Instead "If".
100 PP industry is redundant, maybe...
and what is able to draw 10 PP from the stockpile?
"For making sure" -> "To ensure"
comma after "in the stockpile"
Probably "An example" -> "For example"

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
+The Stockpile Transfer project is above other primary supply-group projects in the build queue. So it is fully funded and will increase the stockpile by 10PP next turn. In order for a lasting effect one sets the repetition to 99 times.'''
```
geoffthemedio: add commas

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
 Transfers PP into the imperial stockpile
 
 PROJECT_BT_STOCKPILE_DESC
-'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. Anything enqueued below the transfer project at a supply-connected planet will not be allocated PP.
+'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. One can specify the amount and repetition of the transfer. PP spent on the transfer are added to the imperial stockpile.
```
geoffthemedio: "One can specify the amount and repetition of the transfer" -> "The amount to transfer and for how many turns the transfer should occur can be specified"

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
 
-Only transfers PP that are available at this planet; PP allocation on planets that are not supply-connected to here will not be affected by this project.
+Only uses PP that are produced in the same supply-group. No PP from the imperial stockpile are used.
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used to transfer PP to the stockpile even if there are things on the queue below it that could have been allocated available PP.'''
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used for controlling PP allocation to the stockpile. PP used for stockpile transfer are not available for items on the build queue below it.
```
geoffthemedio: ... allocation to the stockpile when (something that explains what is meant).

default/stringtables/en.txt
agrrr3: I wouldnt want to use "turns" here. If the stockpile project is not fully funded you dont specify a number of turns.

default/stringtables/en.txt
agrrr3: Not sure i understand what you mean. "..allocation to the stockpile when you need more control.".?? 
I intended "PP used for stockpile transfer are not available for items on the build queue below it." to cover the part that PP spent on stockpile are not available for building in the same turn.

Would something like "This project is used for prioritising stockpile transfer in the build queue." be better?

Or "This project is used to allocate PP to the stockpile when the PP should transfer should happen before another build item."

default/stringtables/en.txt
geoffthemedio: OK, but the stylistic changes remain.

default/stringtables/en.txt
geoffthemedio: "This project is used for controlling PP allocation to the stockpile" is too simple and obvious a statement for where it is located in the text. It could be fine as the first sentence in the whole explanation, but not here in the midst of other text that is explaining the details.

If kept here, it needs to provide some additional information about what is meant, like that it's useful for controlling transfer to the stockpile when there is other stuff on the queue below it that could be allocated PP, or that it's useful to adjust the priority of stockpile allocation when more control is needed, without having to manipulate the stuff later on the queue (eg. by pausing or removing from the queue).

agrrr3: ok, i'll just remove it

default/stringtables/en.txt
agrrr3: Not really able to do so. My english grammar knowledge sucks comma-wise. Please somebody else step in.

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
 Transfers PP into the imperial stockpile
 
 PROJECT_BT_STOCKPILE_DESC
-'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. Anything enqueued below the transfer project at a supply-connected planet will not be allocated PP.
+'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. The amount to transfer and how often the transfer should occur can be specified. PP spent on the transfer are added to the imperial stockpile.
```
geoffthemedio: how often ... should occur -> how many times ... should repeat
PP spent on -> PP allocated to

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
 
-Only transfers PP that are available at this planet; PP allocation on planets that are not supply-connected to here will not be affected by this project.
+Only uses PP that are produced in the same supply-group. No PP from the imperial stockpile are used.
```
geoffthemedio: "No PP from the imperial stockpile are used" is potentially confusing. Maybe "PP already in the imperial stockpile are not allocated to to the stockpile transfer project."

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used to transfer PP to the stockpile even if there are things on the queue below it that could have been allocated available PP.'''
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP used for stockpile transfer are not available for items on the build queue below it.
+
```
geoffthemedio: used for -> allocated to the
... are transferred to the stockpile, and not allocated to items on the production queue.

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
+For example: An empire wants to draw PP from the stockpile in a secondary supply-group. For this the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group and its stockpile use limit is 10 PP. To make sure that 10 PP are always available in the stockpile, one adds to the top of build queue a 10x Stockpile Transfer project on a planet in the primary supply-group.
```
geoffthemedio: For this, ...
one adds ... a ... project -> a 10x Stockpile Transfer project is added to the top of the queue...

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
+The Stockpile Transfer project is above other primary supply-group projects in the build queue. So it is fully funded and will increase the stockpile by 10 PP next turn. In order for a lasting effect one sets the repetition to 99 times.'''
```
geoffthemedio: Because the ... project is above ... the queue, it is fully funded ...
To ensure a continuing transfer each turn, the repetition of the project is set to 99 times.

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
 Transfers PP into the imperial stockpile
 
 PROJECT_BT_STOCKPILE_DESC
-'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. Anything enqueued below the transfer project at a supply-connected planet will not be allocated PP.
+'''Transfers PP at the planet where this project is enqueued to the imperial stockpile. The amount to transfer and how many times the transfer should repeat can be specified. PP allocated to the transfer are added to the imperial stockpile.
```
geoffthemedio: 1st and 3rd sentences seem redundant. Could replace the 1st with the 3rd:

Production points (PP) allocated to the Stockpile Transfer project are transferred to the imperial stockpile. The amount to transfer and how many times the transfer should repeat can be specified.

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
 
-Only transfers PP that are available at this planet; PP allocation on planets that are not supply-connected to here will not be affected by this project.
+Only PP that are produced in the same supply-group are allocated to the stockpile transfer project. PP already in the imperial stockpile are not allocated to the stockpile transfer project.
```
geoffthemedio: ending two sentences right after eachother with "allocated to the stockpile transfer project" is awkward. Maybe:

Only PP that are generated in the same supply group (and none from the imperial stockpile) can be allocated to an Stockpile Transfer project.

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used to transfer PP to the stockpile even if there are things on the queue below it that could have been allocated available PP.'''
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP allocated to stockpile transfer are transferred to the stockpile, and not allocated to items lower in the production queue.
```
geoffthemedio: Second sentence is redundant with previously-given info.

default/stringtables/en.txt
```diff
@@ -339,11 +339,14 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
+
+For example: An empire wants to draw PP from the stockpile in a secondary supply-group. For this the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group, its stockpile use limit is 10 PP and wants that 10 PP are always available in the stockpile. For this, the empire adds to the top of the queue a 10x Stockpile Transfer project on a planet in the primary supply-group.
```
geoffthemedio: ... a secondary supply group, in which there is insufficient PP being generated.

For this ... -> To allow this, ...

... from the primary supply group, in which there are enough items on the production queue to use up all the available PP.

... and wants that 10 PP are always available... -> ... and wants there to always be 10 PP available ...

For this, the empire ... -> To accomplish this, the empire ...

default/stringtables/en.txt
```diff
@@ -339,11 +339,15 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
+'''PP allocated to the transfer project are added to the imperial stockpile. The amount to transfer and how many times the transfer should repeat can be specified.
 
-Only transfers PP that are available at this planet; PP allocation on planets that are not supply-connected to here will not be affected by this project.
+Only PP that are produced in the same supply-group (and none from the imperial stockpile) are allocated to the stockpile transfer project.
 
-Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue.  This project is used to transfer PP to the stockpile even if there are things on the queue below it that could have been allocated available PP.'''
+Surplus PP (PP not allocated to anything on the queue) at this or any other location will be transferred to the stockpile with or without a stockpile transfer project on the queue. PP allocated to a stockpile transfer are not allocated to items lower in the production queue.
```
geoffthemedio: Second sentence reads / sounds unrelated to the first. Needs to connect to it... so maybe `PP allocated to a stockpile transfer are sent to the stockpile, even if there are additional items on the queue to which the PP could have been allocated.`

default/stringtables/en.txt
```diff
@@ -339,11 +339,15 @@ PROJECT_BT_STOCKPILE_SHORT_DESC
...
+
+For example: An empire wants to draw PP from the stockpile in a secondary supply-group, in which there is insufficient PP being generated. To do this the stockpile needs to be filled from the primary supply-group. The empire has 100 PP industry in the primary supply-group, its stockpile use limit is 10 PP and wants that 10 PP are always available in the stockpile. The build queue contains items on planets in the primary supply-group which need 120 PP per turn, so all available PP would be used.
```
geoffthemedio: `To do this, the stockpile...`
`..., its stockpile use limit is 10 PP, and it wants there to always be 10 PP available...`
`... by 10 PP next turn.` -> `... by 10 PP on the next turn.`

default/stringtables/en.txt
agrrr3: Order of build items matters, so there should be something like "below" "lower" "with less priority" "further down" "later"

agrrr3: (and i think we are going in circles here, i will make a final suggestion and then leave it to you guys)

Issue #1846, opened 2017-11-08
Continue option tries to load corrupt saves


MatGB:
So, just had a power cut. On reboot and reloading client, I (naturally) hit the 'Continue' button. got an invalid save error message.

I went in and manually loaded the most recent autosave, then noticed after issuing a few orders it was for the beginning of the previous turn. My autosave folder contains two corrupt files that would be the autosaves for the end of that turn and the start of the next.

The 'Continue' option should only try to load the most recent valid save in the same way the manual load screen only displays valid saves.

(NB: marking this optional, it's such an edge case I nearly marked it no due date, also not sure if it should be UI or internal)

geoffthemedio:
Probably a UI issue, but somewhat complicated and a bit internal in practice. Problem is that the human client doesn't know how to load a full save file... it just reads the initial XML-encoded preview data, and ignores the rest. The rest includes some additional XML-encoded player, empire, galaxy, and other server information. The functions to read that data, and the structures into which it would be stored are currently server-only, and not available on the client. Additionally, after the headers, there is a compressed gamestate string that is also currently only readable by the server.

So, it depends where and how the save file was corrupted. If the initial headers are broken, then checking that when loading the file could let that broken save be skipped, but if it's something later in the file, then there's no easy way to check that it's a valid save without fully loading the save, which is done in the server, which at that point doesn't know whether it's loading an autosave or something else. There'd need to be a bunch of extra code in the client to try the next save in the list if the first didn't load...

PR #2122, opened 2018-05-23
Don't Continue Broken Saves

geoffthemedio:
This moves around a bunch of code so that the client can test-load the headers of save files before attempting to load them with the "continue" option from the menu. If the headers can't be parsed, then it skips the save file and tries the next, instead of attempting and failing to load the newest save file and then stopping.

Addresses https://github.com/freeorion/freeorion/issues/1846

### DIFF
--- a/UI/SaveFileDialog.cpp
+++ b/UI/SaveFileDialog.cpp
@@ -461,7 +461,6 @@ public:
     const FullPreview m_full_preview;
 };
 
-
 class SaveFileListBox : public CUIListBox {
 public:
     SaveFileListBox() :
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -20,7 +20,7 @@
 #include "../../network/ClientNetworking.h"
 #include "../../util/i18n.h"
 #include "../../util/LoggerWithOptionsDB.h"
-#include "../util/GameRules.h"
+#include "../../util/GameRules.h"
 #include "../../util/OptionsDB.h"
 #include "../../util/Process.h"
 #include "../../util/SaveGamePreviewUtils.h"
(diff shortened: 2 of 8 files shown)


Dilvish-fo:
OK, after a bit of fiddling I was able to test and confirm this works as intended on Linux.  I do have a suggestion though-- from the issue, 
> If the initial headers are broken, then checking that when loading the file could let that broken save be skipped, but if it's something later in the file, then there's no easy way to check that it's a valid save without fully loading the save, which is done in the server, 

I'd suggest that if the preview data also contained a checksum for everything that comes afterward, then it could be fairly easy for the client to double check that before actually telling the server to load a save file.  It might very well be that we wouldn't want to always take the time for that just when generating the preview file listing, but it seems it could be helpful to check before telling the server to actually load the game (whether from the Continue operation or if manually selected).  I also understand that might be more than you want to get into for this PR, but wanted to mention the idea.

Also, until we get the AI paying attention to the relevant option, for anyone else who wants to be able to more controllably edit a nominally uncompressed XML savefile, so that it could be possible to 'corrupt' the main body, or the preview, or both, without necessarily always corrupting the main body:
``` diff
index fe2e701..ef8203f 100644
@@ -166,7 +166,7 @@ def prepareForSave():  # pylint: disable=invalid-name
     # serialize (convert to string) global state dictionary and send to AI client to be stored in save file
     import savegame_codec
     try:
-        dump_string = savegame_codec.build_savegame_string()
+        dump_string = savegame_codec.build_savegame_string(False)
         fo.setSaveStateString(dump_string)
     except Exception as e:
         error("Failed to encode the AIstate as save-state string. "
```

geoffthemedio:
> I'd suggest that if the preview data also contained a checksum for everything that comes afterward

How would this checksum be calculated and validated?

Dilvish-fo:
> How would this checksum be calculated and validated?

Hmm, I had the idea for that suggestion because of the comment I quoted from the underlying issue, and the summary above had still seemed to focus entirely on the header.  But looking more closely at the actual code of this PR now,  it's looking to me like the try:except deserialization of the entire file in ```SaveFileWithValidHeader``` probably already accomplishes close enough to the same thing (at least in the Continue situation where it is used).

geoffthemedio:
> the the try:except deserialization of the entire file in SaveFileWithValidHeader

Re here: https://github.com/freeorion/freeorion/pull/2122/files#diff-3254dc27616ea26a3a8c106882f66fcbR203

It doesn't try to deserialize the whole file; just the header text parts. I specifically avoided trying to deserialize the compressed string part, as that could be quite slow and require large memory allocations, and would require moving even more of the save/load code from the server into common (so that the client can use it).

geoffthemedio:
Additional features / tweaks can be a separate PR, I think...

Vezzra:
So, can #1846 be closed as resolved?

~~Should this fix be cherry picked to the release branch?~~ Nevermind, has been assigned to the "post 0.4.8" milestone, so that answers this question.

adrianbroher:
Still, the question if #1846 is now fixed is still unanswered. @geoffthemedio ?

geoffthemedio:
Depends how a save is corrupt. If its header is intact, these changes won't detect the corruption.

PR #2126, opened 2018-05-25
Add dedicated module to interface the singleton aistate

Morlic-fo:
Currently, AIstate is stored in FreeOrionAI which is pretty bad from a design perspective considering it is required by most AI modules:

1) It introduces hard to follow circular imports (eventhough these are currently resolved)
2) FreeOrionAI.py has various side-effects on module load (parsing config, configuring logging...) which require a running AI client and may cause unit tests to fail if they indirectly import it.
3) FreeOrionAI.py is generally only imported for the AIstate

The PR proposes a cleaner design pattern by introducing a dedicated module through which the AIstate quasi-singleton is interfaced. It has no baseline dependencies on other modules but manages the creation of the AIstate from scratch or based on a savegame string by using lazy imports.

The resulting code changes are trivial:
```
- import FreeOrionAI as foAI
+ from aistate_interface import get_aistate()

- *foAI.foAIstate*
+ *get_aistate()*
```
and some different trivial stuff in FreeOrionAI considering its previous ownership of the AI state.

Finally, the workaround to make unit tests work with the savegame codec package is no longer required.

### DIFF
--- a/default/python/AI/aistate_interface.py
+++ b/default/python/AI/aistate_interface.py
@@ -1,0 +1,37 @@
+"""
+This module provides an interface for a quasi-singleton AIstate instance.
+
+The AIstate must be initialized after module imports using either
+   * create_new_aistate
+   * load_aistate
+This should happen only once per session and be triggered by the game client.
+
+Access to the AIstate is provided via get_aistate().
+
+Example usage:
+    from aistate_interface import get_aistate()
+    get_aistate().get_all_fleet_missions()
+
+"""
+_aistate = None
+
+
+def create_new_aistate(*args, **kwargs):
+    import AIstate
+    global _aistate
+    _aistate = AIstate.AIstate(*args, **kwargs)
+    return _aistate
+
+
+def load_aistate(savegame_string):
+    import savegame_codec
+    global _aistate
+    _aistate = savegame_codec.load_savegame_string(savegame_string)
+    return _aistate
+
+
+def get_aistate():
+    """
+    :rtype: AIstate.AIstate
+    """
+    return _aistate
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -8,11 +8,11 @@ from fleet_orders import OrderMove, OrderPause, OrderOutpost, OrderColonize, Ord
 import AIstate
 import EspionageAI
 import FleetUtilsAI
-import FreeOrionAI as foAI
 import MoveUtilsAI
 import MilitaryAI
 import InvasionAI
 import CombatRatingsAI
+from aistate_interface import get_aistate
 from universe_object import System, Fleet, Planet
 from EnumsAI import MissionType
 from AIDependencies import INVALID_ID
@@ -153,7 +153,7 @@ class AIFleetMission(object):
         system_id = main_fleet.systemID
         if system_id == INVALID_ID:
             return  # can't merge fleets in middle of starlane
-        system_status = foAI.foAIstate.systemStatus[system_id]
+        system_status = get_aistate().systemStatus[system_id]
 
         # if a combat mission, and only have final order (so must be at final target), don't try
         # merging if there is no local threat (it tends to lead to fleet object churn)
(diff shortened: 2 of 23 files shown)


Dilvish-fo (commented):
Sure, sounds good, looks good, just needs a bit more grooming for the flake8 check

default/python/AI/aistate_interface.py
```diff
@@ -0,0 +1,22 @@
+aistate = None
```
Cjkjvfnby: I prefer to access `aistate` directly without getter, this is one of the most usable object in a game.

May be we can rename it to shorter but steal meaningless name: `ai`. This object not only contains persistent state, but also some business logic.

Cjkjvfnby (dismissed):

default/python/AI/aistate_interface.py
Dilvish-fo: > I prefer to access aistate directly without getter, this is one of the most usable object in a game.

Are you saying that you think the computational cost of using a getter for this everywhere outweighs the benefits listed in the PR summary? If that's your concern then we could simply get it once at the beginning of every function where it is used ```aistate = get_aistate()``` instead of calling the getter every time we want to reference it. 

I have to note, though, I would not be so crazy about the clarity of distinction between 'aistate' and 'state' (from turn_state)-- I suppose your suggestion of shortening 'aistate' to 'ai' would be one way to deal with that, though I think I'd be a little more inclined to lengthen 'state' to 'turn_state' to strengthen the visual distinction

Morlic-fo: ```from aistate_interface import aistate``` is not possible without more sophisticated changes as the initialization currently happens after imports are finished.

So the choice is between 
```
from aistate_interface import get_aistate
get_aistate().do_stuff()
```

and something like
```
import aistate_interface as ai
ai.state.do_stuff()
```

The latter allows making stupid mistakes such as ```ai.state = 0``` or trying to import the state via the non-working ```from x import y``` statement. Should probably rename to ```_aistate``` when using the getter-approach to make that intent more clear.

Anyway, I have no strong preference.

Dilvish-fo: > So the choice is between

*cough* I will at least insist on comment 😄  on my suggestion of 
```
from aistate_interface import get_aistate

def some_func():
    aistate = get_aistate()
    aistate.do_stuff()
    aistate.do_other_stuff()
```

Morlic-fo: Sorry if I gave you the impression I ignored you. I just focussed on the two possibilities as far as the interface itself is concerned. Usage in functions can of course happen as you suggested (eventhough it won't have any relevant performance impact except for very long loops).

genestack-solomatin: > from aistate_interface import aistate is not possible without more sophisticated changes as the initialization currently happens after imports are finished.

I missed that.

Please add this to getter docstring and prefix module variable with '_'.  This will be enough for this PR.

Everything possible, even code generation upon import.
https://plumbum.readthedocs.io/en/latest/local_commands.html#import-hack

genestack-solomatin: > though I think I'd be a little more inclined to lengthen 'state' to 'turn_state' to strengthen the visual distinction

Freeorion code are full of long lines.  `ai` and `state` may reduce this number, 'aistate' and `turn_state` may enlarge.   

Both pairs are meaningful enough fo me.  Long lines or lines with linebreaks impact readability.

genestack-solomatin: > computational cost of using a getter

Lets consider cost as zero and work with it in a way, that make code more readable in each context. 

`aistate` looks better as `ai_state`

Cjkjvfnby:
I raise my concerns about [naming](https://github.com/freeorion/freeorion/pull/2126#discussion_r191065027), but it should not block this PR.

geoffthemedio:
@Morlic-fo @Dilvish-fo bump?

Morlic-fo:
Will come back to this and remaining open stuff probably in first week of September.

Morlic-fo:
Updating as in reflecting changes in source branch and triggering CI builds.

PR #2128, opened 2018-05-27
Sets launch capacity of FT_BAY_1 to 4 if FT_HANGAR_1 is aboard.

agrrr3:
To make interceptors a bit more useful, let them launch faster

Forum discussion: [Let 4 interceptors launch per bay (stopgap)](http://www.freeorion.org/forum/viewtopic.php?f=15&t=10968)

Implementation: The interceptor hangar sets the capacity of the launch bay in the ship to 4.

* @Dilvish-fo  probably will need AI adjustments? (forgot again which tag to use for AI)
* TODO: also in need of stringtable change

### DIFF
--- a/default/python/AI/AIDependencies.py
+++ b/default/python/AI/AIDependencies.py
@@ -646,9 +646,9 @@ PILOT_FIGHTER_CAPACITY_MODIFIER_DICT = {
     "ULTIMATE": {},
 }
 
-LAUNCH_BAY_BASE_CAPACITY = 2
 HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT = {
-    "FT_HANGAR_1": 2,
+    # hangar_name: {bay_name: effect, bay_name2: effect, ...}
+    "FT_HANGAR_1": {"FT_BAY_1": 2},
 }
 # </editor-fold>
 
--- a/default/python/AI/AIDependencies.py
+++ b/default/python/AI/AIDependencies.py
@@ -645,6 +645,11 @@ PILOT_FIGHTER_CAPACITY_MODIFIER_DICT = {
     "GREAT":    {},
     "ULTIMATE": {},
 }
+
+LAUNCH_BAY_BASE_CAPACITY = 2
+HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT = {
+    "FT_HANGAR_1": 2,
+}
 # </editor-fold>
 
 # <editor-fold desc="Extraordinary Species Rules">
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -694,7 +694,7 @@ class DesignStats(object):
         self.solar_stealth = 0
         self.fighter_capacity = 0
         self.fighter_launch_bays = 0
-        self.fighter_launch_bay_rate = 0
+        self.fighter_launch_bay_rate = AIDependencies.LAUNCH_BAY_BASE_CAPACITY
         self.fighter_damage = 0
 
     @property
@@ -918,7 +918,7 @@ class ShipDesigner(object):
                 else:
                     self.design_stats.fighter_capacity += self._calculate_hangar_capacity(part)
                     self.design_stats.fighter_damage = self._calculate_hangar_damage(part)
-                    self.design_stats.fighter_launch_bay_rate += self._calculate_fighter_launch_bay_rate(part)
+                    self.design_stats.fighter_launch_bay_rate = self._calculate_fighter_launch_bay_rate(part)
 
         self._apply_hardcoded_effects()
 
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -1567,11 +1567,12 @@ class ShipDesigner(object):
         return base + species_modifier + tech_bonus
 
     def _calculate_fighter_launch_rate(self, bay_parts, hangar_part_name):
-        hangar_bonus = AIDependencies.HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT.get(hangar_part_name, 0)
-        bay_parts_capacity = 0
+        launch_rate = 0
+        bays_bonus = AIDependencies.HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT.get(hangar_part_name, {})
         for bay_part in bay_parts:
-            bay_parts_capacity += bay_part.capacity
-        return bay_parts_capacity + ( len(bay_parts) * hangar_bonus )
+            launch_rate += bay_part.capacity
+            launch_rate += bays_bonus.get(bay_part.name, 0)
+        return launch_rate
 
     def _calculate_hangar_damage(self, hangar_part, ignore_species=False):
         hangar_name = hangar_part.name
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -919,7 +919,7 @@ class ShipDesigner(object):
         if len(bay_parts) > 0:
             hangar_part_name = None
             for hangar_part_name in hangar_part_names:
-                break;
+                break
             self.design_stats.fighter_launch_rate = self._calculate_fighter_launch_rate(bay_parts, hangar_part_name)
 
         self._apply_hardcoded_effects()
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -693,13 +693,8 @@ class DesignStats(object):
         self.asteroid_stealth = 0
         self.solar_stealth = 0
         self.fighter_capacity = 0
-        self.fighter_launch_bays = 0
-        self.fighter_launch_bay_rate = AIDependencies.LAUNCH_BAY_BASE_CAPACITY
+        self.fighter_launch_rate = 0
         self.fighter_damage = 0
-
-    @property
-    def fighter_launch_rate(self):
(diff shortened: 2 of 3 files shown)


default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
```diff
@@ -10,6 +10,16 @@ Part
     buildtime = 1
     tags = [ "PEDIA_PC_FIGHTER_HANGAR" ]
     location = OwnedBy empire = Source.Owner
+    effectsgroups = [
+        EffectsGroup
+            scope = [[EMPIRE_OWNED_SHIP_WITH_PART(FT_BAY_1)]]
+            activation = Source
+            effects = [
+                SetCapacity partname = "FT_BAY_1" value = 4
```
geoffthemedio: probably don't want to have this effect. would interfere with anything else that modifies the current capacity, and would make upgrade like mechanics not work.

default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
```diff
@@ -10,6 +10,17 @@ Part
     buildtime = 1
     tags = [ "PEDIA_PC_FIGHTER_HANGAR" ]
     location = OwnedBy empire = Source.Owner
+    effectsgroups = [
+        EffectsGroup
+            scope = [[EMPIRE_OWNED_SHIP_WITH_PART(FT_BAY_1)]]
+            activation = Source
+            stackinggroup = "INTERCEPTOR_FAST_LAUNCH_EFFECT"
+            effects = [
+                SetCapacity partname = "FT_BAY_1" value = Value + 2
```
geoffthemedio: also questionable, as it results in automatic repair / upgrade

Morlic-fo:
> probably will need AI adjustments?

yes.

agrrr3:
Currently we don't have a difference between max value and capacity for
launch bays and i never heard of anybody planning something like this.

So i suggest YAGNI and let's refactor when the need arises.

Geoff <notifications@github.com> schrieb am Mi., 30. Mai 2018, 01:03:

agrrr3:
Sorry, i mistakenly thought it didn't work without explicitly setting the capacity. 
But it i checked again and the launch capacity is already at max capacity in the turn the ships get build, so removed the capacity effect.

agrrr3:
Can anybody add the AI tag(?)

agrrr3:
Ok, i implemented the AI part. The AI should now know about the higher launch rate when doing ship design. And the combat AI was always using the launch bays current capacity, so it does not need any adjustment.

I dont know though where the python compilation happens. Else I'd say this is finished.

agrrr3:
Hm, after playtesting: the AI does not build carriers at all, something wrong i guess

default/python/AI/ShipDesignAI.py
```diff
@@ -907,6 +912,7 @@ def update_stats(self, ignore_species=False):
                 else:
                     self.design_stats.fighter_capacity += self._calculate_hangar_capacity(part)
                     self.design_stats.fighter_damage = self._calculate_hangar_damage(part)
+                    self.design_stats.fighter_launch_bay_rate += self._calculate_fighter_launch_bay_rate(part)
```
Morlic-fo: This seems wrong: You increase the launch bay rate whenever you add another hangar. Should probably just read = rather than +=.

default/python/AI/ShipDesignAI.py
```diff
@@ -1552,6 +1558,12 @@ def _calculate_weapon_shots(self, weapon_part, ignore_species=False):
             species_modifier = 0
         return base + species_modifier + tech_bonus
 
+    def _calculate_fighter_launch_bay_rate(self, hangar_part):
+        hangar_name = hangar_part.name
+        base = AIDependencies.LAUNCH_BAY_BASE_CAPACITY
+        hangar_bonus = AIDependencies.HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT.get(hangar_name, 0)
+        return base + hangar_bonus
```
Morlic-fo: We want the ShipDesignAI to be as generic as possible and handle trivial content changes without modification. This function breaks if a scripter adds another bay part.

default/python/AI/ShipDesignAI.py
```diff
@@ -897,7 +902,7 @@ def update_stats(self, ignore_species=False):
                 else:
                     self.design_stats.stealth = 0
             elif partclass in FIGHTER_BAY:
-                self.design_stats.fighter_launch_rate += capacity
+                self.design_stats.fighter_launch_bays += 1
```
Morlic-fo: Should probably keep a list/counter of bays instead and calculate the launch rate after all parts are processed. That would allow handling of future different bay parts, see comment below.

Morlic-fo (changes requested):

default/python/AI/ShipDesignAI.py
agrrr3: Not sure how calculating the launch rate at a different moment helps with handling content changes.

default/python/AI/ShipDesignAI.py
agrrr3: Do you have an idea how to do that without adding specific launch bay part types?

default/python/AI/ShipDesignAI.py
Morlic-fo: You want to query and use the ```part.capacity``` in case someone adds another bay part (say, with capacity / launch rate 3). That way, trivial content (i.e. stuff without an effects group) is automatically handled by the AI.

As the launch rate of at least 1 type of bay depends on the hangar type (i.e. the changes in your PR), we can not calculate the launch rate before we processed all other parts (and found out which hangar types there are in this design). So we need to defer that calculation. Keeping track of the number of launch bays is insufficient if there are different types of bays, so we want a list / counter of bays.

default/python/AI/ShipDesignAI.py
Morlic-fo: With something like 
```
def _calculate_fighter_launch_bay_rate(self, bay_part_name, hangar_name)
    base = get_part_type(bay_part_name).capacity
    hangar_bonus = AIDependencies.HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT.get(hangar_name, 0)
    return base + hangar_bonus
```
You can get the ```hangar_name``` after all parts are processed from the set ```hangar_parts``` (but should verify that we actually have one and only one hangar part first).

default/python/AI/AIDependencies.py
```diff
@@ -645,6 +645,11 @@
     "GREAT":    {},
     "ULTIMATE": {},
 }
+
+LAUNCH_BAY_BASE_CAPACITY = 2
+HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT = {
+    "FT_HANGAR_1": 2,
```
Morlic-fo: The hangar FOCS script would have no effect for a newly added launch bay. To make the AI aware of this, use a nested dict like
```
HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT = {
    # hangar_name: {bay_name: effect, bay_name2: effect, ...}
    "FT_HANGAR_1": {"FT_BAY_1": 2}, 
}
```

and query via ```HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT.get(hangar_name, {}).get(bay_name, 0)```

agrrr3:
Hm. You mean if somebody adds a new launch bay it shouldnt get the +2 bonus
for interceptors?

I guess i would prefer it gets a *2 bonus instead of a +2

Morlic-fo <notifications@github.com> schrieb am Do., 14. Juni 2018, 20:42:

Morlic-fo:
> Hm. You mean if somebody adds a new launch bay it shouldnt get the +2 bonus
> for interceptors?

The FOCS script you added only applies to "FT_BAY_1" in particular. By default, adding a "FT_BAY_2" would not get the bonus. So the AI shouldn't assume it would get the bonus. 

If a "FT_BAY_2" is added and it should also get that effect, then that should be a separate entry in the AIDependencies.py, i.e. something like

``` 
HANGAR_LAUNCH_CAPACITY_MODIFIER_DICT = {
    # hangar_name: {bay_name: effect, bay_name2: effect, ...}
    "FT_HANGAR_1": {"FT_BAY_1": 2, "FT_BAY_2": 3, "FT_BAY_3": 2}, 
}
```

default/python/AI/AIDependencies.py
agrrr3: Ok, finally i got you. I'd prefer giving an added part automatically the interceptor bonus, but as i see no easy way to do that, I'll take your code snippets, thank you

agrrr3:
Hm. That other lint error is not one of my changes...

Dilvish-fo:
> Hm. That other lint error is not one of my changes...

You didn't even touch the file that it is complaining about, lol.  The travis version of lint must have just gotten updated or something.  I also have a PR up, in which I've fixed those same lint complaints, so you can just ignore them here (or we'd wind up with pointless conflicts between our PRs).

Morlic-fo:
> The travis version of lint must have just gotten updated or something.

No, this happens if people push non-PEP8 compliant stuff directly to master: https://github.com/freeorion/freeorion/commit/e3821267123ba81c3e64ce1f354e2953a81f65b5

Morlic-fo (approved):

agrrr3:
Bump, this is not work-in-progress anymore since two weeks.
 @geoffthemedio can this go to master or are there review changes necessary?

PR #2129, opened 2018-05-27
Tactical drones as core part

agrrr3:
Design discussion started in [Megaflak](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10850#p91253).

Adds a tech and hangar core part for unmanned anti-fighter fighters.

Drones are even worse than interceptors (they always have damage 1) but are fielded really fast (6/turn/launch bay) and can be stored efficiently.


Dilvish-fo:
I just added a comment to that design discussion about the HUGE defensive value that something like this would have.  I think that until that aspect gets a lot of review & testing for balance concerns this would need to remain work-in-progress.

Oberlus:
I think all fighters are drones (fluff: they are too small and accelerate too fast to allow "living" things on it; Egassem certainly do not pilot their fighters), so there is no need for a different kind of interceptor in this MegaHangar part.
And thus, I would allow for three versions of it, one for each type of fighter (with different capacities for each type).

agrrr3:
I guess we will give the ShortRange weapons a default combatTargets condition which will much more likely attack enemy ships than fighters so I think we can include this afterwards.

The drones are very similar to Fighter fighters (FT_HANGAR_2) from targetting, maybe twice as effective (against fighters) and almost no use against ships (but still attacking those).

PR #2130, opened 2018-05-27
Stockpile focus simplification

agrrr3:
Forum discussion: [What to do with the stockpiling techs](http://www.freeorion.org/forum/viewtopic.php?f=6&t=10932&p=92684#p92674)

* decouple GROWTH and STOCKPILE
* allow STOCKPILE FOCUS everywhere
* stockpile focus on homeworld gives species population based stockpile boost (similar to homeworld growth bonus); amount is 0.04; this makes an AVERAGE_STOCKPILE species a GOOD_STOCKPILE species.

### DIFF
--- a/default/scripting/species/common/focus.macros
+++ b/default/scripting/species/common/focus.macros
@@ -52,13 +52,7 @@ HAS_ADVANCED_FOCI
         Focus
             name = "FOCUS_STOCKPILE"
             description = "FOCUS_STOCKPILE_DESC"
-            location = And [
-                OwnerHasTech name = "PRO_GENERIC_SUPPLIES"
-                Or [
-                    Homeworld name = Source.Species
-                    Capital
-                ]
-            ]
+            location = OwnerHasTech name = "PRO_GENERIC_SUPPLIES"
             graphic = "icons/focus/stockpile.png"
 
         Focus
--- a/default/scripting/species/common/stockpile.macros
+++ b/default/scripting/species/common/stockpile.macros
@@ -56,6 +56,24 @@ STANDARD_STOCKPILE
                 min(abs(Value(Target.MaxStockpile) - Value), 1) *
                     (1 - 2*(Statistic If condition = And [Target (Value > Value(Target.MaxStockpile))]))
 
+        // increase stockpile for species if Homeworld is set to stockpile focus
+        EffectsGroup
+            scope = And [
+                ProductionCenter
+                OwnedBy empire = Source.Owner
+                Species name = Source.Species
+                Not Homeworld name = Source.Species
+            ]
+            activation = And [
+                Planet
+                Focus type = "FOCUS_STOCKPILE"
+                Homeworld
+            ]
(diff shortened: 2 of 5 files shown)


Oberlus:
Up (It would be nice to have this merged for 0.4.9)

Vezzra:
@agrrr3, needs rebase.

I don't really know what this does, so I can't comment on it myself, hence tossing the question out to the others: is there anything preventing this from being merged?

geoffthemedio:
Is there some consensus that this change should be implemented, or is it still a suggestion from one person? There is some year old discussion in the linked thread about including it in 0.4.8. That thread also links to another at a point where it mostly a discussion of the basics of srätockpile mechanics...

Vezzra:
After glancing over the linked thread it seems to me that we didn't get into 0.4.8 in time, so postponed it - at least, as far as decoupling of stockpile and growth focus is concerned, first item on the list.

The "allow stockpile everywhere" and stockpile focus on homeworld things seem unrelated, but maybe I just missed it glancing over the thread.

The decoupling of stockpile and growth focus could probably be merged, although the objections brought up in the thread had been that this needs some playtesting and therefore shouldn't be incorporated so close to a release again apply. That should have gone in after the 0.4.8 release, so it could have received proper playtesting since then.

Which obviously didn't happen, so the question again is merge or postpone.

Furthermore, with this PR there is the issue of having two additional changes which probably should receive some prior playtesting.

On the other hand, given the current state of affairs we won't get 0.4.9 out before October. So if we merge now, the changes of this PR would hopefully receive suffcient playtesting until the release.

Bottom line: if we merge this, we should merge it *now*, otherwise we better postpone it again. I'm actually undecided (maybe with a very slight preference for merging, so we can move on with it), because I just have no idea how these changes will actually impact gameplay.

@geoffthemedio, any preference? As to your question, as far as I can tell, this PR is a suggestion of one person, @agrrr3, but at least supported by @Oberlus.

Vezzra:
@agrrr3, you need to address the current confilct in `default/stringtables/en.txt`, otherwise we can't merge even if we decide to do so...

agrrr3:
@Vezzra this proposal is rather consensus based and also requested by yourself. Look at your quote. The only liberty i took is a bonus to balance the higher cost of setting stockpile focus on a homeworld.

> Therefore, my proposal is as follows: For 0.4.8, remove stockpile boni from the growth focus, and remove the restriction of the stockpile bonus to certain colonies.

geoffthemedio:
Never merge master into a branch. Rebase the branch on master.

Vezzra:
> @Vezzra this proposal is rather consensus based...

Correct, as far as the first point ("decouple GROWTH and STOCKPILE") is concerned, the other two (as far as I can see) haven't been discussed in the thread you linked (unless I missed something, which is entirely possible since I only glanced over that thread).

Nevermind though, as there haven't been any comments by others (except @geoffthemedio and mine), @geoffthemedio didn't object, and you solved the conflict and rebased, I guess I go ahead and just merge this. We have to make some progress here... 😉

PR #2135, opened 2018-06-01
Fix AI supply calculation for outposts in evaluate_planet()

Morlic-fo:
For outposts, the ```tag_list``` is empty. The AI would then incorrectly assume the supply bonus of +1 as species without supply tag would have average supply (i.e. +1). Outposts, however, do not get any such bonus.


Dilvish-fo (approved):
Great catch

Dilvish-fo:
@Morlic-fo I just wanted to note that I've seen an even more significant improvement in AI performance from this fix than than the simply-good improvement I had expected-- this was a really great catch to make.

PR #2145, opened 2018-06-10
Have AI use base64 encoding/decoding for compressed save state strings

### LINKED ISSUE #2142: Have AI use base64 encoding for its compressed savestate string
So that, for example, when the game savefile options are set for uncompressed xml, the AI state won't introduce noncompliant character values, [as noted here](https://github.com/freeorion/freeorion/pull/2122#issuecomment-392583446).

Although the use of base64 encoding or not is not itself a bug, the underlying issue of introducing noncompliant characters into the uncompressed xml save does seem like a bug to me, so I am categorizing this as a bug.   I am setting it as optional for 0.4.8 because based on an initial review I think the fix can be quite simple and so possibly suitable to still include.

Dilvish-fo:
- so that the final save-state string will be UTF-8 compatible
- on save, if the save-state is not to be zlib-compressed then base64
  encoding is not applied, so that the savestate will still be human-readable

Somewhat perplexingly, it seems that the base64 decode attempt will only generate an exception
for an improperly padded but otherwise validly-encoded base64 string; a general failure there is not
necessarily detected immediately.

I am categorizing this as a bugfix rather than a tweak because the zlib compression was leading to xml save files being not fully UTF-8 compatible.

Resolves #2142

### DIFF
--- a/default/python/AI/savegame_codec/_decoder.py
+++ b/default/python/AI/savegame_codec/_decoder.py
@@ -15,6 +15,7 @@ import json
 
 import EnumsAI
 from freeorion_tools import profile
+from logging import debug
 
 from _definitions import (ENUM_PREFIX, FALSE, FLOAT_PREFIX, INT_PREFIX, InvalidSaveGameException, NONE, PLACEHOLDER,
                           SET_PREFIX, TRUE, TUPLE_PREFIX, trusted_classes, )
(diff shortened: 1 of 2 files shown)


Morlic-fo (approved):
Functionality seems correct. Should probably be refactored to read something like
```
try:
    return load_string(base64.b64decode(string))
except ... :
    return load_string(string)
```
at some point where load_string is the old functionality but given you want to include this in the release candidate, this can happen later.

PR #2147, opened 2018-06-10
Fix missing sensor ghosts

### LINKED ISSUE #2033: Missing sensor ghost after combat
### Environment
* **FreeOrion Version**: compiled from master (several versions between 2018-02-28 and 2018-03-06).
* **Operating System**: Ubuntu 16.04
* **Graphic card used**: GeForce GTX 860M
* **Fetched as** 
  * Compiled from source

### Description
A ship (scout) arrives to a unexplored system, combat ensues, the scout is killed, in the combat report you can see it was a Frigate Ms (standard initial ship) who killed the scout, but there is no sensor ghost in the system after the combat.

### Expected Result
There should be a sensor ghost for the enemy Frigate Ms ship.
 
### Steps to reproduce
* First step: load saved game "FreeOrion_Human_Player_Terran_0021_2018-Mar-02 19:17:59.107408.sav"
* Second step: do next turn.
* Issue occurs: scout Wezen arrives to system Ankh, combat ensues and no sensor ghost appears after combat.

[miss_ghost_sensor_after_combat_at_end_turn_21.zip](https://github.com/freeorion/freeorion/files/1784130/miss_ghost_sensor_after_combat_at_end_turn_21.zip)

Dilvish-fo:
This PR addresses #2033 re combat-granted sensor ghosts being missing.  

Resolves #2033 

(edited to clean up interim work-in-progress description which is no longer applicable)

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -5592,7 +5592,9 @@ std::vector<int> MapWnd::FleetIDsOfFleetButtonsOverlapping(int fleet_id) const {
 
     const auto& it = m_fleet_buttons.find(fleet_id);
     if (it == m_fleet_buttons.end()) {
-        auto vis_turn_map = GetUniverse().GetObjectVisibilityTurnMapByEmpire(fleet_id, HumanClientApp::GetApp()->EmpireID());
+        // Log that a FleetButton could not be found for the requested fleet, and include when the fleet was last seen
+        int empire_id = HumanClientApp::GetApp()->EmpireID();
+        auto vis_turn_map = GetUniverse().GetObjectVisibilityTurnMapByEmpire(fleet_id, empire_id);
         int vis_turn = -1;
         if (vis_turn_map.find(VIS_BASIC_VISIBILITY) != vis_turn_map.end())
             vis_turn = vis_turn_map[VIS_BASIC_VISIBILITY];
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -5592,7 +5592,12 @@ std::vector<int> MapWnd::FleetIDsOfFleetButtonsOverlapping(int fleet_id) const {
 
     const auto& it = m_fleet_buttons.find(fleet_id);
     if (it == m_fleet_buttons.end()) {
-        ErrorLogger() << "Couldn't find a FleetButton for fleet " << fleet_id;
+        auto vis_turn_map = GetUniverse().GetObjectVisibilityTurnMapByEmpire(fleet_id, HumanClientApp::GetApp()->EmpireID());
+        int vis_turn = -1;
+        if (vis_turn_map.find(VIS_BASIC_VISIBILITY) != vis_turn_map.end())
+            vis_turn = vis_turn_map[VIS_BASIC_VISIBILITY];
+        ErrorLogger() << "Couldn't find a FleetButton for fleet " << fleet_id 
+                      << " with last basic vis turn " << vis_turn;
         return fleet_ids;
     }
     const auto& fleet_btn = it->second;
--- a/combat/CombatSystem.cpp
+++ b/combat/CombatSystem.cpp
@@ -238,9 +238,9 @@ void CombatInfo::InitializeObjectVisibility()
         if (empire_id == ALL_EMPIRES)
             continue;
         empire_known_objects[empire_id].Insert(GetEmpireKnownSystem(system->ID(), empire_id));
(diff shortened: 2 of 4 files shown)


Dilvish-fo:
just in case the save file from the underlying issue is a bit stale in any way, here is a more current test file, on the following turn there should be a working sensor ghost at Ares.
[combat_vis_ghost_test.sav.zip](https://github.com/freeorion/freeorion/files/2088182/combat_vis_ghost_test.sav.zip)

Dilvish-fo:
I am happy to report I believe this issue is fixed.  I have removed the commit that temporarily displayed all stale fleets, and force pushed an additional commit with the key fix.

The basic issue was that the stale object determination process was trying to follow an incremental approach, each turn after combat, removing known destroyed objects or ones that became visible, and adding any newly-stale objects.  The problem was that if an object had been marked stale on a previous turn, the post-combat removal of non-stale objects wouldn't really be complete.  

Consider the case of an enemy ship  properly marked stale on turn 10.  On turn 20, it appears in a system where the player has a fleet but no other visibility, if the player's fleet is destroyed in that combat, then the enemy fleet is not in detection range after combat, nor destroyed, so it was not being removed from the stale list.  A post-move-pre-combat stalesness update would probably catch that particular example, but would not cover the situation of a previously-stale fleet being glimpsed during combat (in which again, the local detector fleet was destroyed).

Also, besides the difficulty in doing an accurate incremental determination, I think that the incremental process was actually not really more efficient than a per-turn de novo determination, and so the latter is what I chose as a fix.

UI/MapWnd.cpp
```diff
@@ -5592,7 +5592,12 @@ std::vector<int> MapWnd::FleetIDsOfFleetButtonsOverlapping(int fleet_id) const {
 
     const auto& it = m_fleet_buttons.find(fleet_id);
     if (it == m_fleet_buttons.end()) {
-        ErrorLogger() << "Couldn't find a FleetButton for fleet " << fleet_id;
+        auto vis_turn_map = GetUniverse().GetObjectVisibilityTurnMapByEmpire(fleet_id, HumanClientApp::GetApp()->EmpireID());
```
geoffthemedio: comment this block? why / what is it checking?

combat/CombatSystem.cpp
```diff
@@ -238,9 +238,9 @@ void CombatInfo::InitializeObjectVisibility()
         if (empire_id == ALL_EMPIRES)
             continue;
         empire_known_objects[empire_id].Insert(GetEmpireKnownSystem(system->ID(), empire_id));
-        empire_object_visibility[empire_id][system->ID()] = GetUniverse().GetObjectVisibilityByEmpire(empire_id, system->ID());
+        empire_object_visibility[empire_id][system->ID()] = GetUniverse().GetObjectVisibilityByEmpire(system->ID(), empire_id);
         for (int object_id : local_object_ids) {
-            Visibility obj_vis = GetUniverse().GetObjectVisibilityByEmpire(empire_id, object_id);
+            Visibility obj_vis = GetUniverse().GetObjectVisibilityByEmpire(object_id, empire_id);
```
geoffthemedio, line 243: this wasn't a major problem worth noting...?

UI/MapWnd.cpp
Dilvish-fo: The change is just augmenting the preexisting logging that happens when the MapWnd has when it is told to select a fleet (such as if a fleet reference in a SitRep is clicked) but it can't find a corresponding FleetButton.  The newly added info is simply the information about when the fleet had last been seen, if at all. This was helpful to me as part of debugging the problem here.

So, a comment could be 
```
// log that a FleetButton could not be found for the requested fleet, and include when the fleet was last seen
```

combat/CombatSystem.cpp
Dilvish-fo: It does certainly look like the kind of error that could have really significant consequences, I'll grant you that.

I am pretty sure that my initial description for the PR (which was up for a day or 2 while it was work in progress)  talked about it, and the corresponding commit message still does  (sufficiently I think).  
Not all of the related calls in this section of code had the problem.  Because of the particular use for these faulty cases, it looked to me like the only time they would potentially cause a real problem was when the object ID was no larger than the number of empires, which in the vast majority of galaxy setups is going to mean the issue could only apply to a System, which is never going to be a combat target, so as a practical matter the error was not causing problems (or doing so extremely rarely).

Is there something about the commit message that you'd like me to change?

geoffthemedio: > Is there something about the commit message that you'd like me to change?

no

UI/MapWnd.cpp
geoffthemedio: would be helpful

PR #2151, opened 2018-06-14
Fix misc ai invasion bugs

Dilvish-fo:
Each of these 4 commits is directed to a separate (albeit somewhat interacting) bug.

The second one, about having AI invasion planning to also consider a Military type mission to be sufficient to count as securing a system (whereas before it was only counting Secure type missions for that purpose, because they will more reliably remain at the target system until the invasion is complete) is responsive to situations where the AI was thinking it needed to build substantially more troopships than it really did, which can significantly delay invasion operations.  This current adjustment might possibly be on the more permissive side of what is really optimal, and further changes might be warranted, but I feel that this change is still an improvement over the previous handling (and made a significant improvement in a couple of test cases I had).

@Morlic-fo, I am feeling confident enough of these fixes being worthwhile that I might go ahead and merge this to get this into RC1, but I'd still like your review and comments

The one about the move path requirement is that without it, troop ships could be assigned to a mission but then be stuck in place, leaving the mission incomplete when other troop ships could have instead been chosen and would have been able to complete the invasion.

### DIFF
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -9,6 +9,7 @@ from universe_object import Planet, Fleet
 from ShipDesignAI import get_part_type
 from AIDependencies import INVALID_ID
 import AIDependencies
+import universe_object
 
 
 def stats_meet_reqs(stats, requirements):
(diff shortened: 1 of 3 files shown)


Morlic-fo:
Good catches on the bugs. The functional changes adressing them look fine. 

Regarding your "creative workaround" for the circular import issue, I opened #2152 to fix the underlying issue.

PR #2152, opened 2018-06-14
Clean up AI imports to avoid circular import issues

Morlic-fo:
Fixes the circular import issue pointed out in #2151 and gets rid of the hack introduced to work around it before it becomes too comfortable in the code base and starts laying eggs...


Dilvish-fo (approved):
Looks good, tests fine.  Thanks for getting that straightened out.

Cjkjvfnby:
@Morlic-fo , ping.

Issue #2140, opened 2018-06-08
No battle, blocking or mine damage


Piwoslaw:
### Environment

* **FreeOrion Version**: FreeOrion_2018-06-04.41c7665_Test_Win32
* **Operating System**: Windows 10, 64-bit
* **Graphic card used**:
* **Fetched as** 
Weekly development build

[save-20180608_065759.zip](https://github.com/freeorion/freeorion/files/2083008/save-20180608_065759.zip)

![fo08062018](https://user-images.githubusercontent.com/19825312/41140590-a7b469e2-6aee-11e8-987e-f53f5f31eb8b.jpg)

### Description
- No battle between my and enemy fleets, even when both parked in the same system for multiple turns.
- My fleet does not damage even scouts nor troop ships.
- Enemy fleet can pass right through systems where my fleet is stationed.
- My fleet in a system does not block enemy supply, enemy fleet does not block my supply.
- No damage to enemy ships in my system with mine tech.

### Steps to reproduce
Previous game (no save files) was OK, but new game has above symptoms. I tried to save, close FO, open and load game, but bug persists.

Vezzra:
@Piwoslaw, I just checked your saved game, and the explanation for your problem is very simple: you are at peace with all other empires. Combat, mine damage etc. only happens if you are at war.

So no bug at all here, everything works just as intended. 😉 

Closing this as invalid.

Piwoslaw:
Hi Vezzra,I checked and yes - I was at peace with everyone. After declaring war the AIs now have no chance:)I was not aware that this is on by default? In the previous game I played yesterday there was not issue with peace/war.Anyhow, thank you for the pointer and sorry for troubling you.Cheers,Piwoslaw

Vezzra:
> I was not aware that this is on by default?

It's not, or it shouldn't. Are you sure you didn't propose peace to them? Because if you didn't, that would mean it's a bug.

So, if you can replicate this situation (being at peace with the other empires despite never having proposed peace to them), please open another issue.

Piwoslaw:
[...]

PR #2156, opened 2018-06-15
Hack for game-start autosaves leaving all empires at peace.

Dilvish-fo:
Is a hack for #2140, should be OK for 0.4.8 but not OK for master

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -313,7 +313,7 @@ def generateOrders():  # pylint: disable=invalid-name
     # TODO: Consider adding an option to clear AI orders after load (must save AIstate at turn start then)
     if fo.currentTurn() == foAIstate.last_turn_played:
         info("The AIstate indicates that this turn was already played.")
-        if not check_bool(get_option_dict().get('replay_turn_after_load', 'False')):
+        if not check_bool(get_option_dict().get('replay_turn_after_load', 'False')) and fo.currentTurn() > 1:
             info("Aborting new order generation. Orders from savegame will still be issued.")
             try:
                 fo.doneTurn()


Morlic-fo:
I don't understand the fix: Why does the save contain the AIstate (otherwise, why isn't the turn replayed?) but not the orders?

Anyway, if it works, fine with me...

Dilvish-fo:
> I don't understand the fix: Why does the save contain the AIstate (otherwise, why isn't the turn replayed?) but not the orders?

Looking into it more, it probably does contain a nominal OrderSet from the AI, although the game start the save is triggered at a different node of the State Machine, and so I suspect it is likely this OrderSet is still empty at that point.  

As an extra little twist, Declaring War is done via Sending  a DiplomaticMessage rather than via an Order, but I think the main point for the purposes of this issue and PR is simply that the save is apparently done prior to the AI doing anything (before generating any orders and before sending the war declarations).That would be an ideal kind of spot for the turn-start autosave in general, but this one is triggered at a point that is quite specific to the start of game.

default/python/AI/FreeOrionAI.py
```diff
@@ -313,7 +313,7 @@ def generateOrders():  # pylint: disable=invalid-name
     # TODO: Consider adding an option to clear AI orders after load (must save AIstate at turn start then)
     if fo.currentTurn() == foAIstate.last_turn_played:
         info("The AIstate indicates that this turn was already played.")
-        if not check_bool(get_option_dict().get('replay_turn_after_load', 'False')):
+        if not check_bool(get_option_dict().get('replay_turn_after_load', 'False')) and fo.currentTurn() > 1:
```
geoffthemedio, line 316: Not sure what review you want from me, but `if (not((something is false) and (something else is true))` is bit hard to parse. Maybe `if ((something is true) or (something else is false))` could be a bit easier?

Or maybe it's already `if (not(something is false) and (something else is true))` in which case just get rid of the double negative?

Morlic-fo:
I can not reproduce the issue.

I tried having autosaves active after galaxy generation and at turn start. In either case, without having the AI replay the turn, I am at war with the AI.

Also, with the galaxy generation save, Production and Colonization orders are correctly stored in the savegame with the AI.

Dilvish-fo:
> I tried having autosaves active after galaxy generation and at turn start. In either case, without having the AI replay the turn, I am at war with the AI.

I haven't tried this on my Windows machine yet, but it is perfectly repeatable for me on Linux.  It is specifically only the very first autosave of the game, made pursuant to the "Enable autosave after galaxy is created before any gameplay' which causes this issue.  Enable that option, start a new game, to make things clear wait a few seconds for the AIs to declare war, check your freeorion.log for the phrase "Turn 0 autosave to" to make sure it really created this particular autosave, and then, without the changes of this PR and with the default False value for the AI replaying turns after load, load that galaxy creation autosave.

> Also, with the galaxy generation save, Production and Colonization orders are correctly stored in the savegame with the AI.

Yes, I double checked mine now, and that is the case for me also for the turn 0 autosave, and examining the code that seems to make tenative sense to me.  After galaxy creation, with this gamestart save option the Server server requests the client save state info and goes into WaitingForSaveData state, in this state it defers handling of all Diplomatic messages, so even if the AIs generate their war declarations before returning their orders and savestate, handling of those declarations would get deferred until after the server finished saving the game.

default/python/AI/FreeOrionAI.py
geoffthemedio: if you think it's safe, and since we probably need an RC2 anyway, sure...

Dilvish-fo:
@Morlic-fo  Hmm, I just thought of another simple approach that I would probably count as an actual fix for the underlying problem here, which I think would be much better than the current approach of this PR, and probably even fine for master-- simply move the block for declaring war to be before before the check for this 'no replay' option, problem solved.  

If the player reloads a turn 0/1 savegame, there is no harm in letting the AI's just re-declare war against everyone.  Even if the player had made peace with an AI on turn 1 and immediately saved the game, and then later reloaded that turn 1 save file, I think it would be reasonable even in that case for the AI to re-declare war.

Vezzra (commented):
Personally I'd [prefer the fix I suggested](https://github.com/freeorion/freeorion/issues/2140#issuecomment-397870234) in the comment section of #2140 (assuming it's as simple as I think it is).

But if you guys prefer this one here, I guess I'm okay with that too - it looks simple enough.

Vezzra:
@Dilvish-fo, as this PR is for the release branch, you don't need to add that to the project managing the PRs/commits that should be cherry picked. No cherry picking required, after all... :wink:

Dilvish-fo:
Superseded by #2164

PR #2159, opened 2018-06-16
Fix GOOD_POPULATION priority

### LINKED ISSUE #2153: Population effect priorities are broken
https://github.com/freeorion/freeorion/commit/9f02a55212b3f6f7193228c5fd91bc116d9102f9 by @geoffthemedio changed the order of growth effects. At the very least, species modifier priorities are changed.

There is a comment by @Dilvish-fo pointing that out but I can't see the issue resolved...(https://github.com/freeorion/freeorion/commit/9f02a55212b3f6f7193228c5fd91bc116d9102f9#r27525956)

Even if the changes are intended for whatever reason, this means that at the very least, the AI calculations are currently significantly broken and must be updated.

Dilvish-fo:
- had been changed without comment during the transition to effects based meter growth
- this reverts the priority back to what it had been before that, putting it back into
  the same priority tier as BAD_POPULATION, which is where the AI expects it.

Fixes #2153

### DIFF
--- a/default/scripting/species/common/population.macros
+++ b/default/scripting/species/common/population.macros
@@ -52,7 +52,7 @@ GOOD_POPULATION
             description = "GOOD_POPULATION_DESC"
             scope = Source
             activation = Planet
-            priority = [[EARLY_TARGET_POPULATION_SCALING_PRIORITY]]
+            priority = [[DEFAULT_PRIORITY]]
             effects = SetTargetPopulation value = Value +0.25*abs(Value) accountinglabel = "GOOD_POPULATION_LABEL"
 '''
 


Morlic-fo:
As discussed in #2153, this does not fix the issue.

geoffthemedio:
The way meter values are used and set was changed with the transition to effect based growth. Target population meters are set first, in a separate step (and thus priority), before the curent population meters, which are before the other meter effects that depend on the current population. Setting target pop meters after the current is probably a bad idea.

I'm not sure why I didn't also change the bad population priority but it should probably also be adjusted similarly.

This wasnt wirhout comment... it was noted with commit messages.

https://github.com/Dilvish-fo/freeorion/commit/9f02a55212b3f6f7193228c5fd91bc116d9102f9

Vezzra:
Ok, never mind, found a workaround: do *not* add the PR via the "Projects" widget, just apply a proper filter at the "Add cards" panel of the project and drag the card into the desired column.

Don't know why the "triage" feature doesn't work anymore... 😕

Dilvish-fo:
> This wasnt wirhout comment... it was noted with commit messages.

Ah, I had looked at that, but I got a bit over-focused on trying to understand the changes in light of the body/detail line and didn't parse out your subject line properly

OK, well it sounds like this PR is both insufficient in scope, and probably not even the right approach for a fix, from what Geoff has indicated.  I'll leave it open for now still, but I am expecting we'll get a better fix put up, at which we'll close this as superseded and reject it on the 0.4.8 project.

Morlic-fo:
Superseded by #2168.

PR #2160, opened 2018-06-17
Adjust timing of AI war declarations

### LINKED ISSUE #2140: No battle, blocking or mine damage

Dilvish-fo:
- So that a loading a galaxy-creation initial autosave, when the AI has its 'replay_turn_after_load' option at the default False value,  will not leave all the empires at peace

This is an alternatate fix, which I consider to be superior, to #2156 

Fixes #2140

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -304,6 +304,15 @@ def generateOrders():  # pylint: disable=invalid-name
     print "***************************************************************************"
     print "***************************************************************************"
 
+    # Do the war declarations before checking the 'replay_turn_after_load' option, so that if someone loads the
+    # initial galaxy creation autosave the AI's will still declare war
+    if turn == 1:
+        declare_war_on_all()
+        human_player = fo.empirePlayerID(1)
+        greet = diplomatic_corp.get_first_turn_greet_message()
+        fo.sendChatMessage(human_player,
+                           '%s (%s): [[%s]]' % (empire.name, get_trait_name_aggression(foAIstate.character), greet))
+
     # When loading a savegame, the AI will already have issued orders for this turn.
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Seems a bit odd... as it's partly replaying its turn by declaring war.

Vezzra:
> This is an alternatate fix, which I consider to be superior, to #2156

While I still prefer [my own suggestion](https://github.com/freeorion/freeorion/issues/2140#issuecomment-397870234), I agree that the fix proposed here is better than #2156.

> Seems a bit odd... as it's partly replaying its turn by declaring war.

Well, it *is* a release branch only stop-gap solution/hack...

geoffthemedio:
On that basis just starting at war would be preferable.

Dilvish-fo:
> While I still prefer my own suggestion...

>On that basis just starting at war would be preferable.

That sounds fine to me as well.

Dilvish-fo:
Superseded by #2164

PR #2161, opened 2018-06-17
Refix stale objects and sensor ghosts

Dilvish-fo:
#2147  fixed a problem with missing sensor ghosts.  In it I had discussed two possible solution approaches; unfortunately, the approach I chose to use there (de novo determination each turn) introduced a new problem with sensor ghosts failing to be marked stale when they should.  The alternative approach, which uses multiple updates to stale object info each turn, is implemented here.  I have two test cases below.  

The first one failed pre-#2147 but worked with it (and works here).  Advance one turn and a scout arrives at Ares and is killed, and there should be a sensor ghost remaining of the fleet that killed it (and of a nearby fleet). 
 [combat_vis_ghost_test.sav.zip](https://github.com/freeorion/freeorion/files/2108766/combat_vis_ghost_test.sav.zip)

The second one is a situation that had worked fine pre-#2147, but then was broken by #2147 (and again works fine here).  In the system to the SW of Stern Beta the client has knowledge an enemy scout was previously seen, but it is currently (and rightly) marked stale and is suppressed.  Advance one turn for the Frigate to move to Stern Beta.  After #2147 but before this PR, the enemy scout would no longer be recognized as stale and would appear as a sensor ghost.
 [let_scout_SW_from_Stern_stay_stale.sav.zip](https://github.com/freeorion/freeorion/files/2108772/let_scout_SW_from_Stern_stay_stale.sav.zip)

### DIFF
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -2562,8 +2562,15 @@ void Universe::UpdateEmpireStaleObjectKnowledge() {
         std::set<int>& stale_set = m_empire_stale_knowledge_object_ids[empire_id];
         const std::set<int>& destroyed_set = m_empire_known_destroyed_object_ids[empire_id];
 
-        // clear previous staleness determinations
-        stale_set.clear();
+        // remove stale marking for any known destroyed or currently visible objects
+        for (auto stale_it = stale_set.begin(); stale_it != stale_set.end();) {
+            int object_id = *stale_it;
+            if (vis_map.count(object_id) || destroyed_set.count(object_id))
+                stale_set.erase(stale_it++);
+            else
+                ++stale_it;
+        }
+
 
         // get empire latest known objects that are potentially detectable
         auto empires_latest_known_objects_that_should_be_detectable =
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -2566,7 +2566,7 @@ void Universe::UpdateEmpireStaleObjectKnowledge() {
         for (auto stale_it = stale_set.begin(); stale_it != stale_set.end();) {
             int object_id = *stale_it;
             if (vis_map.count(object_id) || destroyed_set.count(object_id))
-                stale_set.erase(stale_it++);
+                stale_it = stale_set.erase(stale_it);
             else
                 ++stale_it;
         }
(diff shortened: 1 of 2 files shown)


Vezzra:
I've tried to add this PR to the [project managing the issues/PRs which should be cherry picked](https://github.com/freeorion/freeorion/projects/8), by adding the PR to the project via the "Projects" widget above (like I've done before countless times).

However, with this PR it seems I can't drag the card that has beend created in the project from the "Triage" panel to any of the columns. That is, the *dragging* works just fine, but the change isn't saved. Once I reload the page, the card is back on the Triage panel.

Can someone else try to do that for me (drag the card into the "Proposed" column)? And see if they encounter the same odd issue...

Vezzra:
Ok, never mind, found a workaround: do *not* add the PR via the "Projects" widget, just apply a proper filter at the "Add cards" panel of the project and drag the card into the desired column.

Don't know why the "triage" feature doesn't work anymore... 😕

universe/Universe.cpp
```diff
@@ -2562,8 +2562,15 @@ void Universe::UpdateEmpireStaleObjectKnowledge() {
         std::set<int>& stale_set = m_empire_stale_knowledge_object_ids[empire_id];
         const std::set<int>& destroyed_set = m_empire_known_destroyed_object_ids[empire_id];
 
-        // clear previous staleness determinations
-        stale_set.clear();
+        // remove stale marking for any known destroyed or currently visible objects
+        for (auto stale_it = stale_set.begin(); stale_it != stale_set.end();) {
+            int object_id = *stale_it;
+            if (vis_map.count(object_id) || destroyed_set.count(object_id))
+                stale_set.erase(stale_it++);
```
geoffthemedio: I think calling erase on this iterator will invalidate it, which may be before the increment operator is applied. Need to store a copy, iterate the original, then call erase with the copy. Or use a filtering algorithm, or make a copy and iterate through that rather than the one being modified.

Dilvish-fo: Sure, good catch.  I had simply reverted that without double checking how good the quality of the original code was.

Since we are now compiling as C++11, a simpler/cleaner fix is available.

Dilvish-fo:
rebased to clear that lingering unrelated Travis CI fail.

PR #2164, opened 2018-06-17
Fix initial peace

### LINKED ISSUE #2140: No battle, blocking or mine damage

Dilvish-fo:
Preferred Alternative to #2156 and #2160 
Fixes #2140

### DIFF
--- a/Empire/EmpireManager.cpp
+++ b/Empire/EmpireManager.cpp
@@ -322,14 +322,14 @@ void EmpireManager::ResetDiplomacy() {
     // remove messages
     m_diplomatic_messages.clear();
 
-    // set all empires at peace with eachother (but not themselves)
+    // set all empires at war with each other (but not themselves)
     m_empire_diplomatic_statuses.clear();
     for (auto id_empire_1 : m_empire_map) {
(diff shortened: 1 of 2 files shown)


geoffthemedio:
If it works, I don't immediately see a problem...

PR #2165, opened 2018-06-18
Re-align SetTargetPopulation priorities

Dilvish-fo:
-Per discussion of #2153

Goes with the 4 tier proposal from Morlic, although I did not rename the existing priorities, I simply added one new one.

-Fixes #2153

PR #2168, opened 2018-06-18
Fix priorities for target population effects

### LINKED ISSUE #2153: Population effect priorities are broken

Morlic-fo:
Fixes #2153 and supersedes #2165.

See discussion in #2153.


Dilvish-fo (approved):

default/scripting/buildings/GATEWAY_VOID.focs.txt
```diff
@@ -34,7 +34,7 @@ BuildingType
                 InSystem id = Source.SystemID
             ]
             activation = Turn low = Source.CreationTurn + 1
-            priority = [[EARLY_POPULATION_OVERRIDE_PRIORITY]]
+            priority = [[EARLY_POPULATION_OVERRIDE_PRIORITY]]  // implies overriding target pop effects
```
geoffthemedio: this comment is unclear... or why is it not EARLY_TARGET_POPULATION_OVERRIDE_PRIOIRITY ...?

Morlic-fo: Would an extra "also" help?
```// also implies overriding target pop effects```

The effect should override both target pop and population. Not sure how to word it clearer.

That being said, it seems this new priority means that supply isn't overridden as most supply effects will/may happen later than population effects. Maybe we need to split the effect with the supply effect happening at a later priority so that different supply effects are correctly overridden?

Comment (Dilvish-fo) [reply default/scripting/buildings/GATEWAY_VOID.focs.txt]: Seems to me that anything more here is really just about pre-existing issues with BLD_GATEWAY_VOID, really independent of the target population effects.  A new issue could be opened up regarding possibly splitting the effects of BLD_GATEWAY_VOID, but I don't think we should be doing more with it here beyond the possible clarification to the comment that Morlic has proposed.

File: default/scripting/buildings/GATEWAY_VOID.focs.txt [hunk 2]
```diff
@@ -34,7 +34,7 @@ BuildingType
                 InSystem id = Source.SystemID
             ]
             activation = Turn low = Source.CreationTurn + 1
-            priority = [[EARLY_POPULATION_OVERRIDE_PRIORITY]]  // implies overriding target pop effects
+            priority = [[EARLY_POPULATION_OVERRIDE_PRIORITY]]  // also implies overriding target pop effects
```
geoffthemedio: still not clear to me what this means. is it that this priority will be after both target and current population, and thus override both? I think the use of "implies" is making it unclear.

Morlic-fo:
Comment now reads ``` // Overrides both target and current population effects```.

default/scripting/buildings/GATEWAY_VOID.focs.txt
Oberlus: Maybe this?
// overrides both Population and TargetPopulation

geoffthemedio:
better

PR #2173, opened 2018-06-23
python: Fix inspection warning F812

Cjkjvfnby:
- Fix list comprehension redefines '<variable>' from line <line_no>
- Turn on inspection in flake8 settings

### DIFF
--- a/default/python/AI/InvasionAI.py
+++ b/default/python/AI/InvasionAI.py
@@ -233,7 +233,7 @@ def get_invasion_fleets():
 
     sorted_planets = filter(lambda x: x[1] > 0, sorted_planets)
     # export opponent planets for other AI modules
-    AIstate.opponentPlanetIDs = [pid for pid, _, _ in sorted_planets]
+    AIstate.opponentPlanetIDs = [pid for pid, __, __ in sorted_planets]
     AIstate.invasionTargets = sorted_planets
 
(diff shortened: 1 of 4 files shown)


default/python/AI/InvasionAI.py
```diff
@@ -233,7 +233,7 @@ def get_invasion_fleets():
 
     sorted_planets = filter(lambda x: x[1] > 0, sorted_planets)
     # export opponent planets for other AI modules
-    AIstate.opponentPlanetIDs = [pid for pid, _, _ in sorted_planets]
+    AIstate.opponentPlanetIDs = [pid for pid, __, __ in sorted_planets]
```
Dilvish-fo, line 236: It seems wrong to me that this line would need to be changed, does the inspection really not know to ignore _ being overwritten?

Dilvish-fo (commented):
I'm surprised by one of the changes apparently being needed, but overall, sure looks fine.

default/python/AI/InvasionAI.py
Cjkjvfnby: Agree, but the tool does not care about it.
```
.\AI\InvasionAI.py:236:47: F812 list comprehension redefines '_' from line 181
```

Using underscore as unused is common practice, but not global practice. Another common use case is `gettext`.

PR #2174, opened 2018-06-23
Add missing Generic Supplies tooltips labels

Dilvish-fo:
[Apparently we overlooked these being missing before.](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10994&p=93024&sid=23f9271a09b2626133b439a19e79b40a#p93024) (We had contemplated the possibility there were some more missing labels we hadn't noticed)  And I have also confirmed this particular problem myself as well.


Vezzra:
Ah yes, good catch. That's what release candidates are for... :wink:

@Ouaz, be prepared to provide the French translation for these as quickly as possible after this goes in, as that *might* happen quite close to RC2.

Dilvish-fo:
@Ouaz,  what might work best, particularly given the potentially tight timeframe, is for you to submit the translation changes as a PR to ```fix_generic_supplies_labels``` branch of my fork; I could then accept them there and they would automatically become part of this PR.  Or at least, that might often work best.  Right now I am thinking it would probably be best for me to just merge this PR here right away.

Ouaz:
Sorry for not having answered, I didn't check my messages for a few days... It seems the RC2 release is still TBD, so I updated the FR file in master and created a note as usual in the Release project.

Vezzra:
No problem, responding within two days is more than good enough :wink:

Issue #2157, opened 2018-06-15
VarText Error for Ship seen with only Basic Vis


Dilvish-fo:
If the ship has never been observed with better than BASIC_VISIBILITY, then the SitRep can display 'ERROR' for the ShipDesign.

![image](https://user-images.githubusercontent.com/10995939/41492197-483e40ea-70b2-11e8-8f63-58c53c2ee0f7.png)

My initial review of the code in VarText, Ship, and ShipDesign all looks fine.  It appears to me that the source of the trouble may involve the ShipDesign attribute ```m_name_desc_in_stringtable``` wrongly being set to true.

[SitRepShipError.sav.zip](https://github.com/freeorion/freeorion/files/2107680/SitRepShipError.sav.zip)

Vezzra:
I consider this at least optional for 0.4.8, because if a fix can be provided in time, it should go into the release.

Dilvish-fo:
The problem is more complicated than I thought.

Although I was most recently thinking that this Issue was caused by a situation of ID== -1 (which is what I would expect for a ship that had never been seen with anything better than Basic Visibility), that was not the problem here.  My client knew that the ship had DesignID 4623, but my client had no knowledge of that design.  And the particularly weird thing is that I reviewed the design in the saved game and see no reason why it would be stealthed to me.

For that first combat it could have been possible that the enemy 'Ship' had been passing through an ion storm on its way to the system combat was at, and was still stealthed at the time of combat.  But I can send another ship to that system, and even though my ship survived the first turn of combat, I still only had Basic Vis of that enemy 'Ship', even though there was no ion storm active in the system.

Further pertinent details are that the combat-granted visibility seems to include the values of its structure and stealth meters (which makes sense to me though I didn't track down the exact code), but in this case the stealth meter is only showing a stealth of 15.  So I do not seem to be on the cusp of a full diagnosis and fix.

But perhaps PR 2163 is still a reasonable workaround for 0.4.8, even if not for master.

Vezzra:
As apparently there hasn't been anything done about this issue since 0.4.8, and the temporary fix was only applied to the 0.4.8 release branch, we now either (until a proper fix is provided):

* need to reapply the temporary fix to the 0.4.9 release branch once it has been created, which means we might have the same problem with whatever release comes after 0.4.9.
* just go ahead and apply the temporary fix to master, to avoid running into the issue every time we want to get a release out

I suggest the latter - @geoffthemedio?

[...]

PR #2176, opened 2018-06-24
in 0.4.8 Use 'Unknown Design' for unknown shipdesigns in Sitreps

Dilvish-fo:
- rather than having it display 'Error'
- it is expected that this will happen for ships that have never been glimpsed with
  better than Basic Visibility

This Pull Request is a somewhat more focused version of #2163,  specifically for release-v0.4.8
to provide a reasonable handling for #2157 in 0.4.8 while we continue to determine the more 
precise cause of the problem behavior in #2157 so as to possibly provide an even more finely tailored fix for master.

### DIFF
--- a/util/VarText.cpp
+++ b/util/VarText.cpp
@@ -93,8 +93,12 @@ namespace {
             return boost::none;
         }
         T* object = GetByID(id);
-        if (!object)
-            return boost::none;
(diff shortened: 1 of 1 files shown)


Vezzra (approved):
Sounds reasonable.

PR #2178, opened 2018-06-24
fix some MilitaryAI calculations

Dilvish-fo:
There are still a few places where the MilitaryAI has been using simple addition and subtraction for assessing fleet rating combinations and differences, even though we have had specialized (more accurate) calculations for those in the CombatRatingsAI for some time now.

This PR cleans up those lingering inaccurate uses of plain addition and subtraction.

@Morlic-fo in addition to your substantive review could you give your assessment of suitability to include in 0.4.8?

### DIFF
--- a/default/python/AI/CombatRatingsAI.py
+++ b/default/python/AI/CombatRatingsAI.py
@@ -412,16 +412,16 @@ def rating_needed(target, current=0):
     else:
         return target + current - 2 * (target * current)**0.5
 
+
 def rating_difference(first_rating, second_rating):
 
-    """Estimate the rating of an additional force rating needed to be added to the lesser of the given ratings
-     in order for it to match the the larger of the given ratings.
+    """Return the absolute nonlinear difference between ratings.
 
     :param first_rating: rating of a first force
     :type first_rating: float
     :param second_rating: rating of a second force
     :type second_rating: float
-    :return: Estimated missing rating for the lesser force to match the greater
+    :return: Estimated rating by which the greater force (nonlinearly) exceeds the lesser
     :rtype: float
     """
 
--- a/default/python/AI/CombatRatingsAI.py
+++ b/default/python/AI/CombatRatingsAI.py
@@ -411,3 +411,18 @@ def rating_needed(target, current=0):
         return 0
     else:
         return target + current - 2 * (target * current)**0.5
+
+def rating_difference(first_rating, second_rating):
+
+    """Estimate the rating of an additional force rating needed to be added to the lesser of the given ratings
+     in order for it to match the the larger of the given ratings.
+
+    :param first_rating: rating of a first force
+    :type first_rating: float
+    :param second_rating: rating of a second force
+    :type second_rating: float
+    :return: Estimated missing rating for the lesser force to match the greater
+    :rtype: float
+    """
+
+    return rating_needed(max(first_rating, second_rating), min(first_rating, second_rating))
--- a/default/python/AI/MilitaryAI.py
+++ b/default/python/AI/MilitaryAI.py
@@ -5,7 +5,7 @@ import universe_object
 from EnumsAI import MissionType
 import EspionageAI
 import FleetUtilsAI
-from CombatRatingsAI import combine_ratings
+from CombatRatingsAI import combine_ratings, combine_ratings_list, rating_difference
 import PlanetUtilsAI
 import PriorityAI
 import ProductionAI
@@ -183,7 +183,7 @@ class AllocationHelper(object):
         tup = (sys_id, min_rating, min_rating_vs_planets, take_any, max_rating)
         self.allocations.append(tup)
         self.allocation_by_groups.setdefault(group, []).append(tup)
-        self._remaining_rating -= min_rating
+        self._remaining_rating = rating_difference(self._remaining_rating,  min_rating)
 
 
 class Allocator(object):
@@ -629,7 +629,7 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
 
     mil_fleets_ids = list(FleetUtilsAI.extract_fleet_ids_without_mission_types(all_military_fleet_ids))
     mil_needing_repair_ids, mil_fleets_ids = avail_mil_needing_repair(mil_fleets_ids, split_ships=True)
-    avail_mil_rating = sum(map(CombatRatingsAI.get_fleet_rating, mil_fleets_ids))
+    avail_mil_rating = combine_ratings_list(map(CombatRatingsAI.get_fleet_rating, mil_fleets_ids))
 
     if not mil_fleets_ids:
         if "Main" in thisround:
@@ -774,7 +774,7 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
                 break
             this_alloc = min(remaining_mil_rating, max_alloc)
             new_allocations.append((sid, this_alloc, alloc, rvp, take_any))
-            remaining_mil_rating -= this_alloc
+            remaining_mil_rating = rating_difference(remaining_mil_rating,  this_alloc)
 
     base_allocs = set()
     # for lower priority categories, first assign base_alloc around to all, then top up as available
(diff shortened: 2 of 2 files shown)


default/python/AI/CombatRatingsAI.py
```diff
@@ -411,3 +411,18 @@ def rating_needed(target, current=0):
         return 0
     else:
         return target + current - 2 * (target * current)**0.5
+
+def rating_difference(first_rating, second_rating):
+
+    """Estimate the rating of an additional force rating needed to be added to the lesser of the given ratings
```
Cjkjvfnby: It takes me some time and checks other code, to understand what this function does.  Docstring is not helpful.

maybe something like that: 

`Return absolute difference between ratings. Keep in mind that rating is not the linear function.`

Cjkjvfnby (dismissed):

Morlic-fo:
I remember us having some discussion about this some time ago where the conclusion was not to use the CombatRatings functions here ( I think it was you coming up with a good reason for that). I am not exactly sure when that discussion took place and couldn't find it during a quick search.

I think the argument was based on the fact that the quadratic formula for adding combat ratings is only valid if the fleets are actually all in the same system/mission. However, as we (might) assign single ships here to many different missions/systems, the linear approach would be more suitable to estimate how much rating is left... Not sure if I am missing some other aspect, this is rather wishy-washy and not convincing. Maybe you remember? Or maybe I am just imagining things and the discussion never happened which would explain why I can not find it now...

Dilvish-fo:
I am not really recalling those conversations, but I am still a bit fuzzy-headed from this sinus infection or bad allergy flare-up that I've been experiencing this past week.  It's most likely you are remembering correctly, but if so, my thinking has changed.   And perhaps we had that conversation at a time when we had the combat ratings formula for combining ratings, but had not yet worked out a satisfactory formula for calculating ratings differences.  Using the CombatRatings formula for aggregation but then using simple subtraction for splitting off smaller forces would lead to a serious overestimation of remaining available forces.

The CombatRatings assessment gives the best assessment of the total rating the AI could summon to a single battle, which I feel pretty confident is the best number for the AI to start off considering (especially as it now uses that same max assessment to decide if conquest options are perhaps bigger than it wants to bite off at the moment), and then as it splits off forces for smaller battles it would need to adjust that tally down by also using the CombatRatings calculations.

Morlic-fo:
I do not understand why you define and use the ```rating_difference``` function here instead of using the ```rating_needed``` function.

If we start with a given remaining_rating which is smaller than the allocation performed, then after the allocation, the remaining rating should be zero. Which is exactly what ```rating_needed``` calculates. However, your function basically calculates the equivalent of abs(x - y) - why is that?

Dilvish-fo:
> I do not understand why you define and use the rating_difference function here instead of using the rating_needed function.

Well, I tried to explain it in the commit message-- the way we are using this is to determine a 'rating remaining', not really a 'rating needed', so although the math is the same I thought it would be more clear to access this calculation via a function which which explicitly determines a rating difference, and whose tooltips will be more applicable.

The choice of making this more general, allowing the two forces to be specified in either order, versus a more specific one requiring that the larger force be specified first, is not one that matters a lot to me.  It seemed to make more sense to me like this, but if you would prefer to require that the larger force be specified first, that would be ok with me.  

Then we have less of a reason to have both functions, and perhaps simply rename rating_needed to be rating_difference.  Or we could leave the additional function to have more specific tooltips.   This week though, in particular, I am both pressed for time and not feeling well, and not much wanting to try to decide an appropriate error handling to go with a a revised structure.  I'd rather put the structure through as is, and just agree that you are free to restructure it to specifically require a particular order of the arguments (or just eliminate the additional method) if that continues to seem best to you.  Or you could just put in a different PR to supersede this, is OK with me also.  But I'd like to have something equivalent to this get merged in time for RC2.

Morlic-fo:
My issue with your new function is not based on code style but seemingly incorrect maths (in context of this PR). 

Consider the following extreme case:
```
remaining_rating = 1e-9  # so >0 and not breaking out of loop
min_rating = 100
remaining_rating = rating_difference(remaining_rating, min_rating)
print remaining_rating  # pretty much 100
```
That is certainly a very different result from the previous ```remaining_rating -= min_rating```.

default/python/AI/MilitaryAI.py
```diff
@@ -183,7 +183,7 @@ def allocate(self, group, sys_id, min_rating, min_rating_vs_planets, take_any, m
         tup = (sys_id, min_rating, min_rating_vs_planets, take_any, max_rating)
         self.allocations.append(tup)
         self.allocation_by_groups.setdefault(group, []).append(tup)
-        self._remaining_rating -= min_rating
+        self._remaining_rating = rating_difference(self._remaining_rating,  min_rating)
```
Morlic-fo: Maths changed here from ```x = x - y``` to what is basically ``` x = abs(x - y)```

default/python/AI/MilitaryAI.py
```diff
@@ -774,7 +774,7 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
                 break
             this_alloc = min(remaining_mil_rating, max_alloc)
             new_allocations.append((sid, this_alloc, alloc, rvp, take_any))
-            remaining_mil_rating -= this_alloc
+            remaining_mil_rating = rating_difference(remaining_mil_rating,  this_alloc)
```
Morlic-fo, line 780: Maths changed here from ```x = x - y``` to what is basically ``` x = abs(x - y)```

default/python/AI/MilitaryAI.py
```diff
@@ -783,17 +783,18 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
             if remaining_mil_rating <= 0:
                 break
             base_allocs.add(sid)
-            remaining_mil_rating -= alloc
+            remaining_mil_rating = rating_difference(remaining_mil_rating,  alloc)
```
Morlic-fo: Maths changed here from ```x = x - y``` to what is basically ``` x = abs(x - y)```

default/python/AI/MilitaryAI.py
```diff
@@ -783,17 +783,18 @@ def get_military_fleets(mil_fleets_ids=None, try_reset=True, thisround="Main"):
...
                 break
             if remaining_mil_rating <= 0:
                 new_allocations.append((sid, alloc, alloc, rvp, take_any))
             else:
-                new_rating = min(remaining_mil_rating + alloc, max_alloc)
+                local_max_avail = combine_ratings(remaining_mil_rating, alloc)
+                new_rating = min(local_max_avail, max_alloc)
                 new_allocations.append((sid, new_rating, alloc, rvp, take_any))
-                remaining_mil_rating -= (new_rating - alloc)
+                remaining_mil_rating = rating_difference(local_max_avail, new_rating)
```
Morlic-fo, line 801: Maths changed here from ```x = x - y``` to what is basically ``` x = abs(x - y)```

Morlic-fo (dismissed):
For emphasis, see previous discussion

Dilvish-fo:
> That is certainly a very different result from the previous remaining_rating -= min_rating.

?? Actually, it looks to me like you are just citing an issue about the preexisting code structure, the essence of which which is not changed by this PR whatsoever.  And it furthermore looks to me like in two out of the three cases you are ignoring the lines like 
```
this_alloc = min(remaining_mil_rating, max_alloc)
```
which prevents any such problem.  

So I'll add a similar line for the middle case.

default/python/AI/MilitaryAI.py
Morlic-fo: In this case, there is no check that remaining_rating > min_rating. You rely on external code / calling convention to not break this function which seems like a ticking time bomb.

Morlic-fo (changes requested):
> in two out of the three cases

Right, the other case was bugged though as the loop break condition would have never been met. It is a difference if you once allocate a little more than available or do it 100 times more.

Anyway, still one potential bug introduced by your code.

Dilvish-fo:
> Right, the other case was bugged though as the loop break condition would have never been met.

  I suppose it can't be met now and is superfluous, but it was possible to have been met before

> Anyway, still one potential bug introduced by your code.

Ah, at first I had misunderstood what error you were thinking of here, but OK, so I'll add a similar check here as well.

Morlic-fo (approved):

PR #2192, opened 2018-06-28
Address vector element access wrt _GLIBCXX_ASSERTIONS

### LINKED ISSUE #2182: Crashes when running Fedora package
### Environment
* **FreeOrion Version**: 0.4.7.1
* **Operating System**: Linux (Fedora 28)
* **Graphic card used**: Intel HD 520
* **Fetched as** 
  * Compiled from source

### Description
I am maintaining the FreeOrion in Fedora. I recently started receiving crash reports of a crash occurring immediately when running `/usr/bin/freeorion`. The message when crashing is:

```
/usr/include/c++/8/bits/stl_vector.h:950: std::vector<_Tp, _Alloc>::const_reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) const [with _Tp = float; _Alloc = std::allocator<float>; std::vector<_Tp, _Alloc>::const_reference = const float&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
```

From the stack trace reported on retrace.fedoraproject.org, it's passing through some font render related function calls. I checked my installed versions of DejaVu and Roboto fonts. I also ran `/usr/bin/freeorion --resource-dir ~/git/freeorion/data/default` and it still crashes, so it's not a problem with the system version of the fonts. Is there something else I should check? Freetype versions or Boost versions?

### Expected Result
I expect the game to simply launch. :)

### Steps to reproduce
* Install `freeorion` from the Fedora repository: `dnf install freeorion`
* Run `freeorion`

### Links
 * The spec file and related sources can be found on src.fedoraproject.org: https://src.fedoraproject.org/rpms/freeorion
* The stack trace for the crash has been captured by a crash tracer: https://retrace.fedoraproject.org/faf/reports/2087972/
* The bug report contains additional data about the system that reported the crash: https://bugzilla.redhat.com/show_bug.cgi?id=1575292

dbenage-cx:
Resolves #2182 for current master

### DIFF
--- a/GG/src/GLClientAndServerBuffer.cpp
+++ b/GG/src/GLClientAndServerBuffer.cpp
@@ -105,7 +105,7 @@ void GLClientAndServerBufferBase<vtype>::createServerBuffer()
     glBindBuffer(GL_ARRAY_BUFFER, b_name);
     glBufferData(GL_ARRAY_BUFFER,
                  b_data.size() * sizeof(vtype),
-                 &b_data[0],
+                 b_data.empty() ? nullptr : &b_data[0],
                  GL_STATIC_DRAW);
     glBindBuffer(GL_ARRAY_BUFFER, 0);
 }
@@ -136,7 +136,7 @@ void GLRGBAColorBuffer::activate() const
         glColorPointer(4, GL_UNSIGNED_BYTE, 0, nullptr);
         glBindBuffer(GL_ARRAY_BUFFER, 0);
     } else {
-        glColorPointer(4, GL_UNSIGNED_BYTE, 0, &b_data[0]);
+        glColorPointer(4, GL_UNSIGNED_BYTE, 0, b_data.empty() ? nullptr: &b_data[0]);
     }
 }
 
@@ -171,7 +171,7 @@ void GL2DVertexBuffer::activate() const
         glVertexPointer(2, GL_FLOAT, 0, nullptr);
         glBindBuffer(GL_ARRAY_BUFFER, 0);
     } else {
-        glVertexPointer(2, GL_FLOAT, 0, &b_data[0]);
+        glVertexPointer(2, GL_FLOAT, 0, b_data.empty() ? nullptr: &b_data[0]);
     }
 }
 
@@ -190,7 +190,7 @@ void GLTexCoordBuffer::activate() const
         glTexCoordPointer(2, GL_FLOAT, 0, nullptr);
         glBindBuffer(GL_ARRAY_BUFFER, 0);
     } else {
-        glTexCoordPointer(2, GL_FLOAT, 0, &b_data[0]);
+        glTexCoordPointer(2, GL_FLOAT, 0, b_data.empty() ? nullptr: &b_data[0]);
     }
 }
 
@@ -212,7 +212,7 @@ void GL3DVertexBuffer::activate() const
         glVertexPointer(3, GL_FLOAT, 0, nullptr);
         glBindBuffer(GL_ARRAY_BUFFER, 0);
     } else {
-        glVertexPointer(3, GL_FLOAT, 0, &b_data[0]);
+        glVertexPointer(3, GL_FLOAT, 0, b_data.empty() ? nullptr: &b_data[0]);
     }
 }
 
@@ -231,7 +231,7 @@ void GLNormalBuffer::activate() const
         glNormalPointer(GL_FLOAT, 0, nullptr);
         glBindBuffer(GL_ARRAY_BUFFER, 0);
     } else {
-        glNormalPointer(GL_FLOAT, 0, &b_data[0]);
+        glNormalPointer(GL_FLOAT, 0, b_data.empty() ? nullptr: &b_data[0]);
     }
 }
 
--- a/GG/src/ListBox.cpp
+++ b/GG/src/ListBox.cpp
@@ -432,6 +432,7 @@ void ListBox::Row::SetColAlignments(const std::vector<Alignment>& aligns)
         return;
 
     m_col_alignments = aligns;
+    m_col_alignments.resize(m_cells.size(), ALIGN_NONE);
     auto&& layout = GetLayout();
     ValidateLayoutSize(layout.get(), aligns.size());
     for (std::size_t ii = 0; ii < m_cells.size(); ++ii) {
@@ -459,6 +460,7 @@ void ListBox::Row::SetColWidths(const std::vector<X>& widths)
         return;
 
     m_col_widths = widths;
+    m_col_widths.resize(m_cells.size(), GG::X(5));
     auto&& layout = GetLayout();
     ValidateLayoutSize(layout.get(), widths.size());
     for (std::size_t ii = 0; ii < m_cells.size(); ++ii) {
@@ -485,6 +487,7 @@ void ListBox::Row::SetColStretches(const std::vector<double>& stretches)
         return;
 
     m_col_stretches = stretches;
+    m_col_stretches.resize(m_cells.size(), 0.0);
     auto&& layout = GetLayout();
     ValidateLayoutSize(layout.get(), m_col_stretches.size());
     for (std::size_t ii = 0; ii < m_cells.size(); ++ii) {


GG/src/ListBox.cpp
```diff
@@ -431,12 +426,18 @@ void ListBox::Row::SetColAlignments(const std::vector<Alignment>& aligns)
     if (aligns == m_col_alignments)
         return;
 
-    m_col_alignments = aligns;
+    m_col_alignments.resize(m_cells.size(), ALIGN_NONE);
+
+    for (std::size_t i = 0; i < m_cells.size(); ++i) {
```
Dilvish-fo: although the resize() above makes m_col_alignments.size() the same as m_cells,size(), it sure seems to me like this would read better by using m_col_alignments.size() here.

GG/src/ListBox.cpp
```diff
@@ -458,11 +459,17 @@ void ListBox::Row::SetColWidths(const std::vector<X>& widths)
     if (widths == m_col_widths)
         return;
 
-    m_col_widths = widths;
+    m_col_widths.resize(m_cells.size(), X(5));
+
+    for (std::size_t i = 0; i < m_cells.size(); ++i) {
```
Dilvish-fo: same comment about m_cells.size()

GG/src/ListBox.cpp
```diff
@@ -484,11 +491,17 @@ void ListBox::Row::SetColStretches(const std::vector<double>& stretches)
     if (stretches == m_col_stretches)
         return;
 
-    m_col_stretches = stretches;
+    m_col_stretches.resize(m_cells.size(), 0.0);
+
+    for (std::size_t i = 0; i < m_cells.size(); ++i) {
```
Dilvish-fo: and here too

GG/src/ListBox.cpp
dbenage-cx: ty, fixed

GG/src/GLClientAndServerBuffer.cpp
```diff
@@ -170,7 +170,7 @@ void GL2DVertexBuffer::activate() const
         glBindBuffer(GL_ARRAY_BUFFER, b_name);
         glVertexPointer(2, GL_FLOAT, 0, nullptr);
         glBindBuffer(GL_ARRAY_BUFFER, 0);
-    } else {
+    } else if (!b_data.empty()) {
```
geoffthemedio: Could this leave OpenGL in an unexpected state depending on what was preivously bound?

GG/src/ListBox.cpp
```diff
@@ -311,17 +311,12 @@ void ListBox::Row::resize(std::size_t n)
     std::size_t old_size = m_cells.size();
 
     for (std::size_t ii = n; ii < old_size; ++ii) {
-        m_cells[ii].reset();
+        m_cells.at(ii).reset();
```
geoffthemedio: probably need to handle the exceptions these uses of .at(...) could throw

GG/src/ListBox.cpp
```diff
@@ -431,12 +426,18 @@ void ListBox::Row::SetColAlignments(const std::vector<Alignment>& aligns)
     if (aligns == m_col_alignments)
         return;
 
-    m_col_alignments = aligns;
+    m_col_alignments.resize(m_cells.size(), ALIGN_NONE);
+
+    for (std::size_t i = 0; i < m_col_alignments.size(); ++i) {
```
geoffthemedio: why?

GG/src/ListBox.cpp
```diff
@@ -458,11 +459,17 @@ void ListBox::Row::SetColWidths(const std::vector<X>& widths)
     if (widths == m_col_widths)
         return;
 
-    m_col_widths = widths;
+    m_col_widths.resize(m_cells.size(), X(5));
```
geoffthemedio: again, this seems pointless...

GG/src/GLClientAndServerBuffer.cpp
dbenage-cx: Can't speak to OpenGL state, the previous code could modify `b_data` and pass an uninitialized value.

GG/src/ListBox.cpp
dbenage-cx: For any elements missing from `aligns`, the value of `m_col_alignments` is either unchanged or initialized to `ALIGN_NONE`.
If the size of `aligns` exceeds `m_cells`, the excess elements are ignored.

GG/src/ListBox.cpp
```diff
@@ -431,12 +443,18 @@ void ListBox::Row::SetColAlignments(const std::vector<Alignment>& aligns)
...
+        if (i < aligns.size())
+            m_col_alignments.at(i) = aligns.at(i);
+    }
+
     auto&& layout = GetLayout();
     ValidateLayoutSize(layout.get(), aligns.size());
     for (std::size_t ii = 0; ii < m_cells.size(); ++ii) {
-        if (m_cells[ii])
-            layout->SetChildAlignment(m_cells[ii].get(), m_row_alignment | m_col_alignments[ii]);
+        if (m_cells.at(ii))
```
geoffthemedio: `ii` ranges from 0 to `m_cells.size() - 1`, so using `m_cells.at(ii)` here instead of `m_cells[ii]` seems pointless.

GG/src/ListBox.cpp
geoffthemedio: would be simpler if you kept `m_col_alignments = aligns;` and then shrank or expanded `m_col_alignments` to be the same size as `m_cells`. The extra uses of `.at(i)` and different varieties of bounds checking are confusing to read.

GG/src/GLClientAndServerBuffer.cpp
geoffthemedio: where is/was `b_data` being modified? `operator[]` doesn't add elements to a vector.

It's passing a pointer to where the data array of `b_data` would be if there were some. Doesn't necessarily mean there's a problem, since the `count` parameter to the subsequent `glDrawArrays(...)` call should be `0` if the appropriate vector size is passed and that vector is empty.

that said, I think out of bounds use of `operator[]` is undefined behaviour, but possibly so is (or effectively is) not specifying any vectex pointer before calling `glDrawArrays(...)`.

GG/src/GLClientAndServerBuffer.cpp
```diff
@@ -135,8 +135,8 @@ void GLRGBAColorBuffer::activate() const
         glBindBuffer(GL_ARRAY_BUFFER, b_name);
         glColorPointer(4, GL_UNSIGNED_BYTE, 0, nullptr);
         glBindBuffer(GL_ARRAY_BUFFER, 0);
-    } else {
-        glColorPointer(4, GL_UNSIGNED_BYTE, 0, &b_data[0]);
+    } else if (!b_data.empty()) {
+        glColorPointer(4, GL_UNSIGNED_BYTE, 0, &b_data.at(0));
     }
```
geoffthemedio: maybe add
````
} else {
    glColorPointer(4, GL_UNSIGNED_BYTE, 0, nullptr);
}
````
(untested)

GG/src/GLClientAndServerBuffer.cpp
dbenage-cx: Somehow I forgot about the differing behavior of `operator[]` :/
I've minimized the changes to ListBox.

Might look at defining a refactoring cells and cell properties into a new Row::Cell struct after 0.4.8 is finalized.

GG/src/ListBox.cpp
dbenage-cx: Ok

GG/src/GLClientAndServerBuffer.cpp
dbenage-cx: changed slightly to ternary

dbenage-cx:
Rebased

Dilvish-fo:
@Vezzra you had marked the underlying issue as cherry pick for release; I didn't notice any other PR or commits corresponding to this one but specific to 0.4.8, so I am thinking you probably were intending this PR?

Vezzra:
@Dilvish-fo, ah yes, of course, apparently I mixed up the issue and the corresponding PR. I meant to tag the PR for cherry picking.

Going to do that right now.

Issue #2190, opened 2018-06-27
Disable option that turns off the predefined designs (0.4.8 only, temporary workaround)


Vezzra:
As discussed in #2107, the option to turn off predefined ship designs is currently broken and, as a quick fix doesn't seem possible, should be disabled for 0.4.8.

geoffthemedio:
Option is hidden by that commit. Consider cherry picking?

geoffthemedio:
Feel free to revert and implement as you prefer...

Edit: Actually, just hiding that option isn't enough in any case, as there are popup menus to adjust the setting.

geoffthemedio:
Better PR added, probably just for the release branch, although made against master.

Vezzra:
> Consider cherry picking?

Done: a792292c753c8bd2ee6e6a8b475ac702ffe636fe

PR #2194, opened 2018-06-30
Remove references to option to add default designs

geoffthemedio:
Hopefully addresses https://github.com/freeorion/freeorion/issues/2190

Intended for the release branch, not master, in addition to https://github.com/freeorion/freeorion/commit/07db247a4c54b8c7f7cde67e8b0ed25bfb73c4bf

### DIFF
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -1128,8 +1128,8 @@ void ShipDesignManager::StartGame(int empire_id, bool is_new_game) {
         displayed_designs->InsertHullBefore(hull_name);
     }
 
-    // If requested initialize the current designs to all designs known by the empire
-    if (GetOptionsDB().Get<bool>("resource.shipdesign.default.enabled")) {
+    // If requested, initialize the current designs to all designs known by the empire
+    if (true) { //GetOptionsDB().Get<bool>("resource.shipdesign.default.enabled")) {
         // While initializing a new game, before sending info to players, the
         // server should have added the default design ids to an empire's known
         // designs. Loop over these, and add them to "current" designs.
@@ -2736,7 +2736,7 @@ void BasesListBox::ResetEmptyListPrompt()
 
 void CompletedDesignsListBox::ResetEmptyListPrompt() {
     if (!GetOptionsDB().Get<bool>("resource.shipdesign.saved.enabled")
-        && !GetOptionsDB().Get<bool>("resource.shipdesign.default.enabled"))
+        /*&& !GetOptionsDB().Get<bool>("resource.shipdesign.default.enabled")*/)
     {
         SetEmptyPromptText(UserString("NO_SAVED_OR_DEFAULT_DESIGNS_ADDED_PROMPT"));
     } else {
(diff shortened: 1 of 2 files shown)


UI/DesignWnd.cpp
```diff
@@ -2736,7 +2736,7 @@ void BasesListBox::ResetEmptyListPrompt()
 
 void CompletedDesignsListBox::ResetEmptyListPrompt() {
     if (!GetOptionsDB().Get<bool>("resource.shipdesign.saved.enabled")
-        && !GetOptionsDB().Get<bool>("resource.shipdesign.default.enabled"))
+        /*&& !GetOptionsDB().Get<bool>("resource.shipdesign.default.enabled")*/)
```
Dilvish-fo, line 2739: I think this change isn't quite right.  In all the the other places you are explicitly or implicitly replacing ```GetOptionsDB().Get<bool>("resource.shipdesign.default.enabled")``` with ```true``` whereas  here it looks like you may have overlooked the "!", since just commenting out this line winds up implicitly treating the logical clause like it was true, meaning treating the option like it was false.

I think the most textually simple change to make here, in keeping with the changes you made elsewhere, would be to not comment out the whole line, but change it to
```
        && false /*!GetOptionsDB().Get<bool>("resource.shipdesign.default.enabled")*/)
```

Dilvish-fo (commented):
I think one of the changes merits some adjustment (though it might not cause any actual harm the way it is).

I am also inclined to still recommend comment out the registration of the option, 
```
UI/ClientUI.cpp:573:        db.Add("resource.shipdesign.default.enabled", UserStringNop("OPTIONS_DB_ADD_DEFAULT_DESIGNS"),            true);
```
so that it won't appear as if it should be a working option if someone runs FO with the --help option.  But I guess I don't feel super strongly about that, perhaps it could just go into some release notes caveat list or something that the option is disabled?

Dilvish-fo:
Also, I suppose you might be reluctant to check out the release branch to be able to make this branch & PR actually based on it, since that (and then when you switch back to master) would likely trigger a painfully slow recompilation on Windows.  (Hopefully someday soon there will be ccache type functionality available for MSVC, I saw someone was at least working on similar functionality that would be compatible with the MSVC command line tools.)

If you'd like me to cherry pick these commits, plus make the changes I mention above, and put into a new branch based on 0.4.8 which could then be a PR specifically into 0.4.8, let me know.

geoffthemedio:
Go ahead. I can't make edits or compile for a few hours at least.

UI/DesignWnd.cpp
geoffthemedio: If I recall correctly, I tried to make things as though the extra option didn't exist. If you make it `&& false` the prompt will never display.

Dilvish-fo: >  If you make it && false the prompt will never display.

Correct; the prompt doesn't really make sense if the option is not available.   A more complete removal would be to remove the whole section of the if statement, but it seems your goal with the other changes was just to disable rather than completely remove, and simple disablement seems fine to me.  But in this particular case I think that means not showing this prompt at all.


```
NO_SAVED_OR_DEFAULT_DESIGNS_ADDED_PROMPT
'''Select a hull, add parts and then click "[[DESIGN_WND_ADD_FINISHED]]" to create an available design.

Neither the Saved Designs nor the Default Designs were added to the empire at game start.  Options to add these designs at game start can be changed in Options->Other.  Add the Saved Designs immediately with the context menu of any design under the Saved tab.'''
```

geoffthemedio: as you prefer...

Vezzra:
Superseded by #2199.

UI/DesignWnd.cpp
geoffthemedio: > Correct; the prompt doesn't really make sense if the option is not available.

There is still the option for the saved designs in this case though...

PR #2195, opened 2018-07-01
[logger] use same format for python and cpp loggers

Cjkjvfnby:
Logging example where both C++ and python logs
```
03:20:08.721918 [debug] ai : AIClientApp.cpp:318 : Message::GAME_START setting AI aggression as 5 (from rnd 51; max aggression 5)
03:20:08.721918 [debug] python : FreeOrionAI.py:86 - Initializing foAIstate...
```

### DIFF
--- a/python/LoggingWrapper.cpp
+++ b/python/LoggingWrapper.cpp
@@ -101,19 +101,19 @@ namespace {
         // if that log level is disabled.
         switch (log_level) {
         case LogLevel::trace:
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
Even if LGM-Doyle weren't apparently taking a break from FO work now, I think there wouldn't be much need to wait on him to review such a tiny/simple change.

PR #2196, opened 2018-07-01
[tests] Fix incorrect sorting

Cjkjvfnby:
`list.sort()` return `None`, next checks was always:  `None == None`

Anyway, these checks were useless because you compare dicts afterward. 

PS. squash and merge.


Morlic-fo:
Looks reasonable.

Dilvish-fo:
I temporarily restored the branch, so it could be available to double check what commit in master it had been based on,  in case it's not clear why Morlic's slightly more recent PR is failing Travis whereas this one passed Travis.

Cjkjvfnby:
One of my PR failing Travis too, but it does not look like it is changes introduced in PR.  https://github.com/freeorion/freeorion/pull/2195

PR #2197, opened 2018-07-01
Fix industry/research meter growth (#2183 - #2185)

### LINKED ISSUE #2183: NO_RESEARCH and NO_INDUSTRY doesn't update current meters
### Description
The current research meter of NO_RESEARCH species does not update, and thus is fixed to zero, ignoring fixed bonuses from techs like Nascent AI or specials like Eccentric Orbit.
Presumably, the same applies for NO_INDUSTRY species and fixed production bonuses like Adaptive Automation or Gas Giant Generator.

Dilvish wrote:
> I am pretty sure it is an accidental glitch from the move of meter regen into Effects. The basic Research and Industry meter regen is currently part of BASIC_RESEARCH and BASIC_INDUSTRY, and so gets left out of NO_RESEARCH and NO_INDUSTRY.
> Certainly a tech like Nascent AI tech is pretty explicitly supposed to be independent of the racial research abilities: "Each planet can contain such structures regardless of the population's occupations, giving a bonus to research on all worlds." And regardless of whether other possible fixed sources would necessarily be to explicitly applicable, if the effect sets the target meter as shown in the screenshot, then the target meter should be allowed to grow

### Expected Result
The current meter of every species, regardless of research/production trait, should update every turn.

Dilvish wrote:
> I think the most expeditious fix is simply to extract the respective meter regen clauses out of the BASIC_X macros, into a separate X_REGEN macro which could then be referenced by both the BASIC_X macro and the NO_X macro. Or probably better yet, just **move all the regen clauses into a single METER_REGEN macro which could then be incorporated into BASIC_POPULATION**.

### LINKED ISSUE #2184: Target meters not showing actual target when current meters are hold to zero
### Description
Effects that hold to zero a current Meter do not show it properly in the visible effects when there are other (previous) bonuses applying to TargetMeter.

Dilvish wrote [in the forum](http://www.freeorion.org/forum/viewtopic.php?f=28&t=10999#p93086):
> [...] the current implementation of that seems a bit buggy to me since if it is going to hold the primary meter to zero it should also hold the target meter to zero.

### Expected Result
Effects that hold current Meter to zero should also hold TargetMeter to zero.

### LINKED ISSUE #2185: Handling of Dead versus Dying colonies
### Description
The handling of dying colonies my be buggy. If I understood it correctly, the effect that handles it does treat the same dying (diminishing populations) and dead (already zeroed) colonies, why it probably should only apply to dead colonies.

Dilvish wrote:
> the comment description is "// removes residual production from a dead planet" whereas the implementation applies to any dying planet

Morlic-fo:
1st commit fixes #2183  where NO_INDUSTRY/NO_RESEARCH species didn't get the meter-modifying effectgroup.

2nd/3rd commit resolve #2185.

6th commit resolves #2184


Dilvish-fo:
The Travis failure wasn't just a simple timeout problem, it's an actual failure, regarding AI code.  At first glance I am not immediately seeing the connection to this PR, but just a few hours ago #2196 passed fine.  It could be that the underlying branch for #2196 was based on an older commit in master, so that might be something to look into.  The scripting of this PR looks fine to me.

```
18:18:02.755184 [debug] python : FreeOrionAI.py:55 - Path to folder for user specific data: /root/.local/share/freeorion
18:18:02.755190 [debug] python : FreeOrionAI.py:55 - Path to folder for user specific data: /root/.local/share/freeorion
18:18:02.755307 [debug] python : FreeOrionAI.py:56 - Python paths ['/usr/lib/python2.7', '/usr/lib/python2.7/plat-x86_64-linux-gnu', '/usr/lib/python2.7/lib-tk', '/usr/lib/python2.7/lib-old', '/usr/lib/python2.7/lib-dynload', '/usr/local/lib/python2.7/dist-packages', '/usr/lib/python2.7/dist-packages', '/freeorion/build/default/python', '/freeorion/default/python/AI']
18:18:02.755390 [debug] python : FreeOrionAI.py:56 - Python paths ['/usr/lib/python2.7', '/usr/lib/python2.7/plat-x86_64-linux-gnu', '/usr/lib/python2.7/lib-tk', '/usr/lib/python2.7/lib-old', '/usr/lib/python2.7/lib-dynload', '/usr/local/lib/python2.7/dist-packages', '/usr/lib/python2.7/dist-packages', '/freeorion/build/default/python', '/freeorion/default/python/AI']
18:18:02.755576 [debug] ai : AIFramework.cpp:129 : AI Python modules successfully initialized!
18:18:02.755614 [debug] ai : CommonFramework.cpp:100 : FreeOrion Python interface successfully initialized!
18:18:02.755897 [error] python : configure_logging.py:197 - Uncaught exception: Traceback (most recent call last):
18:18:02.755932 [error] python : configure_logging.py:197 -   File "/freeorion/default/python/AI/FreeOrionAI.py", line 382, in <module>
18:18:02.755940 [error] python : configure_logging.py:197 -     init_handlers(fo.getOptionsDBOptionStr("ai-config"), fo.getAIDir())
18:18:02.755947 [error] python : configure_logging.py:197 -   File "/freeorion/default/python/AI/freeorion_tools/handlers.py", line 10, in init_handlers
18:18:02.755956 [error] python : configure_logging.py:197 -     handlers = split(get_option_dict()[HANDLERS])
18:18:02.755963 [error] python : configure_logging.py:197 - KeyError: 'handlers'
18:18:02.755972 [error] python : configure_logging.py:197 - 
18:18:02.756177 [error] ai : AIClientApp.cpp:196 : AIClientApp failed due to error in python AI code for AI_1.  Exiting Soon. id = -1
terminate called after throwing an instance of 'std::runtime_error'
  what():  PythonAI failed to initialize.

```

Morlic-fo:
This looks more like a random travis fail rather than strictly related to any recent PR.

Looking at the history, the same error happened after merging #2172 but not afterwards for some while: https://travis-ci.org/freeorion/freeorion/builds/396101223?utm_source=github_status&utm_medium=notification

Dilvish-fo:
Perhaps rebase, just to get that error cleared (no big deal, but the red x's are a distraction).

Dilvish-fo (approved):

default/scripting/species/common/general.macros
```diff
@@ -23,6 +23,7 @@ STANDARD_METER_GROWTH
             activation = And [
                 Planet
                 TargetPopulation high = 0
+                Population high = 0.1
             ]
             effects = [
                 SetResearch value = 0
```
Dilvish-fo: I just remembered about #2184.  If you have this also zero out the target meters here, then I believe that #2184 would also be resolved (I am not positive that there are no other sources of such zeroing, but this is the only one I recall).

Dilvish-fo:
Hmm, display of my [previous comment](https://github.com/freeorion/freeorion/pull/2197#pullrequestreview-133477652) seems to be getting suppressed as being applicable to outdated lines, though I don't see why, so I'll just quote it here.

> I just remembered about #2184. If you have this also zero out the target meters here, then I believe that #2184 would also be resolved (I am not positive that there are no other sources of such zeroing, but this is the only one I recall).

Morlic-fo:
Besides your suggested change to resolve #2184, I also added a missing accountinglabel and assigned the END_CLEANUP_PRIORITY to the effect.

Dilvish-fo (approved):
Looks good to me

PR #2198, opened 2018-07-01
Improve AI tests

Cjkjvfnby:
I have added checks to error message too, not only error type.

I rework manager tests to make a single test to the sinlge feature. 

Before:
```
test_savegame_manager.py::test_simple_object_encoding PASSED             [ 20%]
test_savegame_manager.py::test_class_encoding PASSED                     [ 40%]
test_savegame_manager.py::test_getstate_call PASSED                      [ 60%]
test_savegame_manager.py::test_setstate_call PASSED                      [ 80%]
test_savegame_manager.py::test_enums PASSED                              [100%]

========================== 5 passed in 0.08 seconds ===========================
```

After:
```
test_savegame_manager.py::test_encoding_simple_object[0] PASSED          [  3%]
test_savegame_manager.py::test_encoding_simple_object[0.0] PASSED        [  7%]
test_savegame_manager.py::test_encoding_simple_object[] PASSED           [ 11%]
test_savegame_manager.py::test_encoding_simple_object[False0] PASSED     [ 15%]
test_savegame_manager.py::test_encoding_simple_object[1] PASSED          [ 19%]
test_savegame_manager.py::test_encoding_simple_object[1.2] PASSED        [ 23%]
test_savegame_manager.py::test_encoding_simple_object[12000.0] PASSED    [ 26%]
test_savegame_manager.py::test_encoding_simple_object[TestString] PASSED [ 30%]
test_savegame_manager.py::test_encoding_simple_object[True] PASSED       [ 34%]
test_savegame_manager.py::test_encoding_simple_object[False1] PASSED     [ 38%]
test_savegame_manager.py::test_encoding_simple_object[None] PASSED       [ 42%]
test_savegame_manager.py::test_encoding_simple_object[()] PASSED         [ 46%]
test_savegame_manager.py::test_encoding_simple_object[set([])] PASSED    [ 50%]
test_savegame_manager.py::test_encoding_simple_object[[]] PASSED         [ 53%]
test_savegame_manager.py::test_encoding_simple_object[{}] PASSED         [ 57%]
test_savegame_manager.py::test_encoding_simple_object[(1, 2, 3)] PASSED  [ 61%]
test_savegame_manager.py::test_encoding_simple_object[[1, 2, 3]] PASSED  [ 65%]
test_savegame_manager.py::test_encoding_simple_object[set([1, 2, 3])] PASSED [ 69%]
test_savegame_manager.py::test_encoding_simple_object[{'a': 1, True: False, (1, 2): set([1, 2, 3]), (1, 2, (3, (4,))): [1, 2, (3, 4), set([1, 2, 3])]}] PASSED [ 73%]
test_savegame_manager.py::test_ecoding_function PASSED                   [ 76%]
test_savegame_manager.py::test_ecoding_old_style_class PASSED            [ 80%]
test_savegame_manager.py::test_ecoding_type PASSED                       [ 84%]
test_savegame_manager.py::test_class_encoding PASSED                     [ 88%]
test_savegame_manager.py::test_getstate_call PASSED                      [ 92%]
test_savegame_manager.py::test_setstate_call PASSED                      [ 96%]
test_savegame_manager.py::test_enums PASSED                              [100%]

========================== 26 passed in 0.22 seconds ==========================

```

### DIFF
--- a/default/python/tests/AI/test_savegame_manager.py
+++ b/default/python/tests/AI/test_savegame_manager.py
@@ -135,13 +135,13 @@ def test_class_encoding():
 
 def test_getstate_call():
     with TrustedScope():
-        with pytest.raises(Success, message="__getstate__ was not called during encoding.", match=""):
+        with pytest.raises(Success, message="__getstate__ was not called during encoding."):
             savegame_codec.encode(GetStateTester())
 
 
 def test_setstate_call():
     with TrustedScope():
-        with pytest.raises(Success, message="__setstate__ was not called during decoding.", match=""):
+        with pytest.raises(Success, message="__setstate__ was not called during decoding."):
             savegame_codec.decode(savegame_codec.encode(SetStateTester()))
 
 
--- a/default/python/tests/AI/test_savegame_manager.py
+++ b/default/python/tests/AI/test_savegame_manager.py
@@ -84,21 +84,21 @@ def test_encoding_simple_object(simple_object):
     check_encoding(simple_object)
 
 
-def test_ecoding_function():
+def test_encoding_function():
     with pytest.raises(savegame_codec.CanNotSaveGameException,
                        message="Could save function",
                        match="Class __builtin__.function is not trusted"):
         check_encoding(lambda: 0)
 
 
-def test_ecoding_old_style_class():
+def test_encoding_old_style_class():
     with pytest.raises(savegame_codec.CanNotSaveGameException,
-                       message="Could Old style class",
+                       message="Could save Old style class",
                        match="Encountered unsupported object test_savegame_manager.OldStyleClass \(<type 'classobj'>\)"):
         check_encoding(OldStyleClass)
 
 
-def test_ecoding_type():
+def test_encoding_type():
     with pytest.raises(savegame_codec.CanNotSaveGameException,
                        message="Could save untrusted class",
                        match="Class __builtin__.type is not trusted"):
--- a/default/python/tests/AI/test_savegame_manager.py
+++ b/default/python/tests/AI/test_savegame_manager.py
@@ -1,6 +1,7 @@
 import pytest
 
 import savegame_codec
+from pytest import fixture
 
 
 class DummyTestClass(object):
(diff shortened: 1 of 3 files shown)


Morlic-fo (changes requested):
I don't like your suggested checks for the error messages for the following reasons:

1) The ```match``` parameter is checked against in a ```re.search``` call. ```match=""``` is basically equivalent to ```assert 1==1``` - don't do that.
2) If someone changes the text of an exception slightly, the tests will fail.
3) In case of 3, the incorrect reason is given. E.g.
```
with pytest.raises(savegame_codec.InvalidSaveGameException,
message="Could load object from untrusted module",
match="DANGER DANGER - test_savegame_manager.DummyTestClass not trusted"):
```
- The message says that the object could be loaded even if an InvalidSaveGameException has been thrown (but with a slightly different non-matching string).
4) The tests are designed to test the behavior of the function, not to test the exact error message.

As an alternative, I suggest the following:
Uses of match should generally be the minimum required human readable regexp or substring which indicates that a specific "subclass" of an exception was thrown, not the literal and entire error message (unless short and descriptive). This makes both the code and the tests more maintainable.

If you really want to test the error message, add a separate test. E.g.
```
with pytest.raises(TypeError) as e:
    raise TypeError("Expected Error message")
assert e.message == "Expected Error message"  # in general: use some regular expression here without refering to specific types or names of variables
```
This resolves 3) and 4) and, with appropriate regular expressions / substrings, minimizes the impact of 2)

Cjkjvfnby:
> The match parameter is checked against in a re.search call. match="" is basically equivalent to assert 1==1 - don't do that.

Missed that, removed.

> 2. If someone changes the text of an exception slightly, the tests will fail.

It is not common practice to change exception frequency. It is easy to find it and fix it. 

Message check is required when code can drop the same exception class in different locations.  For example, if someone add  `.foo` to `class_name` in [this line](https://github.com/freeorion/freeorion/blob/master/default/python/AI/savegame_codec/_encoder.py#L93) you will get `CanNotSaveGameException` exception.

> 3. In case of 3, the incorrect reason is given. E.g.
In the case of 2?

Looks good for me:
```
AssertionError: Pattern 'DANGER DANGER - test_savegame_manager.DummyTestClass not trusted!' not found in 'DANGER DANGER - test_savegame_manager.DummyTestClass not trusted'
```

> 4. The tests are designed to test the behavior of the function, not to test the exact error message.

As I mentioned in 2, the message is part of function behavior.

> As an alternative, I suggest the following:
Uses of match should generally be the minimum required human readable regexp or substring which indicates that a specific "subclass" of an exception was thrown, not the literal and entire error message (unless short and descriptive). This makes both the code and the tests more maintainable

Using the whole message make the test more readable.  Using some magic regexp has some chance to be wrong in case of little changes and be too general in case of real error.

default/python/tests/AI/test_savegame_manager.py
```diff
@@ -56,35 +57,59 @@ class OldStyleClass():
...
-            (1, 2, 3), [1, 2, 3], {1, 2, 3},
-            {'a': 1, True: False, (1, 2): {1, 2, 3}, (1, 2, (3, (4,))): [1, 2, (3, 4), {1, 2, 3}]},
-            ):
-        check_encoding(obj)
 
-    for obj in (lambda x: 1, check_encoding, OldStyleClass, DummyTestClass):
-        with pytest.raises(savegame_codec.CanNotSaveGameException,
-                           message="Could save unsupported, potentially dangerous object."):
-            check_encoding(obj)
+def test_ecoding_function():
```
Morlic-fo: Typo: encoding

default/python/tests/AI/test_savegame_manager.py
```diff
@@ -56,35 +57,59 @@ class OldStyleClass():
...
+    with pytest.raises(savegame_codec.CanNotSaveGameException,
+                       message="Could save function",
+                       match="Class __builtin__.function is not trusted"):
+        check_encoding(lambda: 0)
+
+
+def test_ecoding_old_style_class():
```
Morlic-fo: Typo: encoding

default/python/tests/AI/test_savegame_manager.py
```diff
@@ -56,35 +57,59 @@ class OldStyleClass():
...
+    with pytest.raises(savegame_codec.CanNotSaveGameException,
+                       message="Could Old style class",
```
Morlic-fo: missing "save"

Morlic-fo (commented):
> Looks good for me:

Right, I misremembered what quirks there used to be with ```match```.

Still, in 3.4.0, there were some issues with test output:
```
        with pytest.raises(KeyError, message="Could save stuff", match="Test"):
>           pass
E           Failed: Could save stuff matching 'Test'

        with pytest.raises(KeyError, message="Could save stuff", match="Test"):
>           raise ValueError
E           AssertionError: Pattern 'Test' not found in ''
```

Upgrading pytest to a more recent version (3.4.1 would be enough) would give cleaner output though.
```
        with pytest.raises(KeyError, message="Could save stuff", match="Test"):
>           pass
E           Failed: Could save stuff

        with pytest.raises(KeyError, message="Could save stuff", match="Test"):
>           raise ValueError
E           ValueError
```

default/python/tests/AI/test_savegame_manager.py
Cjkjvfnby: I usually go to previous message and add something like `, + check other cases`

Morlic-fo (approved):

Issue #2205, opened 2018-07-05
Wrong delimiter for cross-platform multiplayer.


o01eg:
### Environment
* **FreeOrion Version**: 2018-06-24.4983374 0.4.8 pre-RC2
* **Operating System**: Ubuntu Linux 18.04.0 x64
* **Fetched as** 
  * Compiled from source

### Description
I'm not sure but it looks like UI for loading save game use local delimiter not remote one. This cause wrong filename sent to the server. These are logs from multiplayer server. I suppose `Rudy` at 18:03 use Windows. I use Linux at 18:33:

```
18:03:28.630091 [debug] server : SaveGamePreviewUtils.cpp:93 : Loaded preview with: 2 human players
18:03:28.630104 [debug] server : SaveGamePreviewUtils.cpp:101 : LoadSaveGamePreviewData: Successfully loaded preview from: /var/lib/freeorion/save/./MeRuSeSp212/FreeOrion_0056_20180705_171854.mps
18:03:28.630136 [debug] server : ServerApp.cpp:971 : ServerApp::UpdateSavePreviews: Sending 87 previews in response.
18:03:30.124056 [debug] FSM : ServerFSM.cpp:977 : Get message from host or allowed player.
18:03:30.124115 [debug] FSM : ServerFSM.cpp:104 : PlayerSetupData:
18:03:30.124134 [debug] FSM : ServerFSM.cpp:119 :  ... -1 : AI_1, AI_PLAYER, SP_CHATO
18:03:30.124145 [debug] FSM : ServerFSM.cpp:119 :  ... -1 : AI_2, AI_PLAYER, SP_ETTY
18:03:30.124155 [debug] FSM : ServerFSM.cpp:119 :  ... 1 : Rudy, PLAYER, SP_EAXAW
18:03:30.124234 [debug] server : SaveLoad.cpp:542 : LoadEmpireSaveGameData: filename: /var/lib/freeorion/save/./MeRuSeSp212\FreeOrion_0088_20180705_174533.mps path:"/var/lib/freeorion/save/./MeRuSeSp212\FreeOrion_0088_20180705_174533.mps"
18:03:30.124284 [error] server : SaveLoad.cpp:574 : Error when reading save file. LoadEmpireSaveGameData exception: : Unable to open file
...
[...]

PR #2206, opened 2018-07-05
Use generic_wstring on Windows in PathToString

geoffthemedio:
Hopefully addresses https://github.com/freeorion/freeorion/issues/2205

### DIFF
--- a/util/Directories.cpp
+++ b/util/Directories.cpp
@@ -502,7 +502,7 @@ fs::path RelativePath(const fs::path& from, const fs::path& to) {
 #if defined(FREEORION_WIN32)
 
 std::string PathToString(const fs::path& path) {
-    fs::path::string_type native_string = path.native();
+    fs::path::string_type native_string = path.generic_wstring();// path.native();
     std::string retval;
     utf8::utf16to8(native_string.begin(), native_string.end(), std::back_inserter(retval));
     return retval;
--- a/util/Directories.cpp
+++ b/util/Directories.cpp
@@ -502,7 +502,7 @@ fs::path RelativePath(const fs::path& from, const fs::path& to) {
 #if defined(FREEORION_WIN32)
 
 std::string PathToString(const fs::path& path) {
-    fs::path::string_type native_string = path.generic_wstring();// path.native();
+    fs::path::string_type native_string = path.generic_wstring();
     std::string retval;
     utf8::utf16to8(native_string.begin(), native_string.end(), std::back_inserter(retval));
     return retval;
@@ -512,9 +512,11 @@ fs::path FilenameToPath(const std::string& path_str) {
     // convert UTF-8 directory string to UTF-16
     boost::filesystem::path::string_type directory_native;
     utf8::utf8to16(path_str.begin(), path_str.end(), std::back_inserter(directory_native));
-    if (BOOST_VERSION >= 106300)
-        return fs::path(directory_native).generic_path();
+#if (BOOST_VERSION >= 106300)
+    return fs::path(directory_native).generic_path();
+#else
     return fs::path(directory_native);
+#endif
 }
 
 #else // defined(FREEORION_WIN32)
--- a/util/Directories.cpp
+++ b/util/Directories.cpp
@@ -512,6 +512,8 @@ fs::path FilenameToPath(const std::string& path_str) {
     // convert UTF-8 directory string to UTF-16
     boost::filesystem::path::string_type directory_native;
     utf8::utf8to16(path_str.begin(), path_str.end(), std::back_inserter(directory_native));
+    if (BOOST_VERSION >= 106300)
+        return fs::path(directory_native).generic_path();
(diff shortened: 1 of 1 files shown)


server/SaveLoad.cpp
```diff
@@ -540,7 +540,7 @@ void LoadEmpireSaveGameData(const std::string& filename, std::map<int, SaveGameE
     try {
         fs::path path = FilenameToPath(filename);
         DebugLogger() << "LoadEmpireSaveGameData: filename: " << filename << " path:" << path;
-        fs::ifstream ifs(path, std::ios_base::binary);
+        fs::ifstream ifs(path.generic_path(), std::ios_base::binary);
```
dbenage-cx: Would need to bump minimum boost version to 1.63 (note this is commented as experimental).

path.generic [first appeared](https://github.com/boostorg/filesystem/commit/6e92c9a8b293c4b94969dd309405c80b67dacfdd) in boost version 1.60, with the rename to generic_path in 1.63

~[make_preferred](https://www.boost.org/doc/libs/1_58_0/libs/filesystem/doc/reference.html#path-modifiers)?~ (Edit: would not work either)

geoffthemedio: Maybe
````
#ifdef FREEORION_LINUX
        fs::ifstream ifs(path.native(), std::ios_base::binary);
#else
        fs::ifstream ifs(path, std::ios_base::binary);
#endif
````

dbenage-cx: In this case path.native == path.string
If the path separator is suspected to be the issue, may want to convert to path.generic on client side before sending to server (most/all such functions on linux perform no conversion).
`IsExistingFile` could be used to validate at server.

geoffthemedio:
Try that now?

o01eg:
@geoffthemedio Currently I don't have a Windows OS.

Could someone build 0.4.8 branch with this patch on Windows and try to choose savegame?

geoffthemedio:
> ... try to choose savegame?

Could you clarify what you mean by that?

o01eg:
I mean to load it in the lobby.

geoffthemedio:
In master + this patch, I can load a multiplayer save in the MP lobby, on Windows, with non-Latin characters in the path, while connected to a localhost server.

server/SaveLoad.cpp
Vezzra: As #2205 has been reported for the release branch, a fix for it needs to be cherry picked. However, on the release branch we still work with v8 of the SDK, meaning, IIRC, the min boost version there is 1.58 or 1.59. I really don't want to have to upgrade the SDK/boost versions for the release branch, so a patch for #2205 must not rely on newer boost versions.

geoffthemedio:
Restored use of `generic_path()` when the local boost version supports it. Might need to switch to `#ifdef` though...

o01eg:
Looks like server-side `SaveLoad.cpp` fix don't work. Could someone test this PR with Windows client 0.4.8 branch and my server?

geoffthemedio:
"don't work" how?

It should have been redundant, since the client-side fix is still in place.

dbenage-cx:
@o01eg The game seemed to load ok (with Win10), might check logs.  Am rebuilding linux for local test if there is other issue.

dbenage-cx:
As far as testing, LGTM.

As it is fairly straight forward, could implement our own utility function to replace using generic_path.

Vezzra (changes requested):
@geoffthemedio,

> Restored use of generic_path() when the local boost version supports it. Might need to switch to #ifdef though...

Yep, definitely. If I try to build this PR applied to the release branch I get build errors (because `generic_path` is an unknown symbol).

Vezzra:
@o01eg,

> Could someone build 0.4.8 branch with this patch on Windows and try to choose savegame?

With this PR applied to the release branch, I can load savegames on your server. Without the PR applied I can't (trying to load a game just won't do anything, don't know what happens, can't check the logs on your server).

So, as far as I can tell, this PR fixes things at least insofar as loading a savegame cross-platform works with that patch, and doesn't without it.

geoffthemedio:
> If I try to build this PR applied to the release branch I get build errors (because generic_path is an unknown symbol).

Switched to `#if` instead of `if`

Vezzra:
Ok, builds fine now with the release branch.

Wanted to make sure and run another quick test with @o01eg's multiplayer server, but couldn't (someone else was logged into the multiplayer lobby, but didn't respond, so I couldn't try to load a game, and then I lost connection to the server and couldn't reconnect).

However, the final adjustment has been trivial enough, so I don't expect any issues.

I'm going to merge and cherry pick this, so it can go into todays test builds.

o01eg:
Good. Tomorrow I will be with my PC and could publish fixed version.

> someone else was logged into the multiplayer lobby, but didn't respond, so I couldn't try to load a game

This bug didn't allow even choose save game in the lobby.

PR #2208, opened 2018-07-07
Upgrade TravisCI pytest version to 3.6.3

Morlic-fo:
In currently used pytest version 3.4.0, there are some quirks with the ```match``` parameter in ```pytest.raises``` which may give unprecise error messages if the test fails. This was adressed in 3.4.1 
https://docs.pytest.org/en/latest/changelog.html#pytest-3-4-1-2018-02-20

3.6.3 is the latest version at the time of this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -132,7 +132,7 @@ jobs:
       language: python
       python: 2.7.14
       before_install:
-        - pip install pytest==3.4.0
+        - pip install pytest==3.6.3
       script:
         - pytest
       env:


Cjkjvfnby:
Devs who use python tests should be informed to update their environment.  (`pip install pytest -U`), Who are this devs? How should I inform them?

@adrianbroher , @geoffthemedio , @Dilvish-fo, @Vezzra   do you have any ideas?

geoffthemedio:
Forum post? Pull requests are somewhat like push messages anyway...

Dilvish-fo:
I agree with Geoff, the devs most in need of notice here probably get watch notifications for all activity, such as this PR.  It would have been a bit more notice for a broader group if it had been left open for a few days before merging (you could have rebased your 'improve tests' PR to be based off this one in the meantime), but that's not a big deal.

You should add a reference about the min required version into our testing docs somewhere, also.

Vezzra:
> Devs who use python tests should be informed to update their environment. (pip install pytest -U), Who are this devs? How should I inform them?

@Cjkjvfnby, while all the relevant devs are most likely notified of that change by the notification emails sent by github, I still think the proper place for an announcement like this is the [Programming subforum](http://www.freeorion.org/forum/viewforum.php?f=9). Posting there should be sufficient.

PR #2217, opened 2018-07-15
Fix ResourceSupplyConnectedByEmpire condition

### LINKED ISSUE #2215: Belts with Asteroid Reformation Processor under blockage do not count as connected to it
* **FreeOrion Version**: build 2018-07-13.9a03676 MSCV 2015 (0.4.8 RC2, binary release)
* **Operating System**: Windows 7

### Description
An asteroid belt colonised with exobots and equiped with Asteroid Processor (AP) and Asteroid Reformation Processor (ARP) was building scattered asteroid hulls. Once the system has got blockaded by an Enemy, the asteriod belt fails to meet the building requirements, which is wrong. Check attached image
fo_cant_built_asteroid_hulls_on_belts
I've checked that asteroid hulls with the AP requirement (and not ARP) are available to be built in that same asteroid belt, and that the ARP counts as present in the asteroid belt (can't build a second one because it does not meet the requirement "there is no ARP in the planet"). I don't see why the FOCS code could be failing here.

### Expected Result
The hull should be allowed in the asteroid belt under blockage, since it has all required buildings. More specifically, the building requirements should be acknowledge as met, which are not.

### Steps to reproduce
Load attached save game. Check system Adhil Gamma. Check the items being built in Adhil Gamma Asteroids. There is a batch of 1 "Swarm Carrier" about to be finished. It's due date is "never". If you check the unavailable ship designs you'll see the "Swarm Carrier" design there, the asteroid belt fails to meet the Asteroid Reformation Processor building requirement, while it is actually built in that asteroid belt.

[FreeOrion_Sicofonte_Shy_0238_20180715_131612.zip](https://github.com/freeorion/freeorion/files/2195881/FreeOrion_Sicofonte_Shy_0238_20180715_131612.zip)

Dilvish-fo:
- so that it handles the fact that blockaded planets are still resource-connected to themselves

Some potential for confusion remains, in that:
- the related matter of blockaded planets having their production points available to them is
  currently handled in ResourcePool::Update() solely with regards to the empire's
      m_resource_pools[RE_INDUSTRY].m_connected_object_groups_resource_output
  without making corresponding changes to
      m_resource_pools[RE_INDUSTRY].m_connected_system_groups
  (nor to GetSupplyManager().ResourceSupplyGroups(m_id))
  and perhaps m_connected_system_groups should be renamed to be m_connected_object_groups and
    get a similar update

Fixes #2215

### DIFF
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -8228,10 +8228,41 @@ namespace {
                 return false;
 
             // is candidate object connected to a subcondition matching object by resource supply?
+            // first check if candidate object is (or is a building on) a blockaded planet
+            // "isolated" objects are anything not in a non-blockaded system
+            bool is_isolated = true;
+            for (const auto& group : groups) {
+                if (group.count(candidate->SystemID())) {
+                    is_isolated = false;
+                    break;
+                }
+            }
+            if (is_isolated) {
+                // planets are still supply-connected to themselves even if blockaded
+                auto candidate_planet = std::dynamic_pointer_cast<const Planet>(candidate);
+                std::shared_ptr<const ::Building> building;
+                if (!candidate_planet && (building = std::dynamic_pointer_cast<const ::Building>(candidate)))
+                    candidate_planet = GetPlanet(building->PlanetID());
+                if (candidate_planet) {
+                    int candidate_planet_id = candidate_planet->ID();
+                    // can only match if the from_object is (or is on) the same planet
+                    for (auto& from_object : m_from_objects) {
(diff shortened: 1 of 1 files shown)


agrrr3:
~~Hm. Without digging deeper ... shouldnt the fix go into SupplyManager::Update ?~~
After digging a bit deeper - SupplyManager currently only talks about systems, so it shouldnt concern individual planets. 
I feel there should be a more central place to handle this case but my cpp is not deep enough to suggest a reasonably efficient implementation.

PR #2225, opened 2018-07-25
Add EncyclopediaDetailPage.

flaviojs:
This is a refactoring PR.

I used "page" instead of "item" in the name because it's clearer.
Q1: Should I change other comments, variables, and function names that refer to a page as an item?

Originally I made these changes so that I could introduce async loading during or after #2219, but I realized that async loading won't be needed in most cases. Selective refresh should be enough to prevent most freeze times caused by the encyclopedia, and the turn number appears to be a good reference.

Q2: Is there anything you can do during a turn that should force a page to immediately refresh?  I'm not sure, but I have a feeling everything you do during your turn only takes effect (forces a page refresh) after you are done with the turn.

### DIFF
--- a/UI/EncyclopediaDetailPanel.cpp
+++ b/UI/EncyclopediaDetailPanel.cpp
@@ -557,10 +557,29 @@ namespace {
     };
 }
 
-std::list<std::pair<std::string, std::string>>
-    EncyclopediaDetailPanel::m_items = std::list<std::pair<std::string, std::string>>(0);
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Modifying (eg. renaming) an existing ship design causes that design's page to update during a turn. There are also articles with lists of objects, including the list of fleets, which can be created and destroyed during a turn. Various objects can also be renamed during a turn.

flaviojs:
I'm not be around to continue this so maybe it should be closed?

Issue #2201, opened 2018-07-03
Pedia lookup of a ship produces errors


flaviojs:
### Environment
* **FreeOrion Version**: v0.4.8+ (master at ae39545)
* **Operating System**: Linux Mint 18 Sarah
* **Graphic card used**:Advanced Micro Devices, Inc. [AMD/ATI] RV710/M92 [Mobility Radeon HD 4530/4570/545v]
* **Fetched as**
  * Compiled from source

### Description
Doing a pedia lookup of a ship produces errors in the log file `freeorion.log`.
```
08:51:19.640508 [error] client : ValueRef.cpp:86 : FollowReference : Unable to get system for object
08:51:19.640925 [error] client : ValueRef.cpp:903 : Variable<int>::Eval unable to follow reference: LocalCandidate.System.LastTurnBattleHere :  | Local Candidate: Ship -2 (  )  |  System (-1):  |  LastTurnBattleHere  | 
08:51:19.641374 [error] client : ValueRef.cpp:905 : source: OBJ_PLANET 2248 ( Sole γ I )
```

### Expected Result
No errors.

### Steps to reproduce
1. right click on a ship in the map or production windows
2. choose the option to open the pedia page

geoffthemedio:
Probably need to check that a ship is in a system before testing the last turn that system had a battle...

Dilvish-fo:
> choose the option to open the pedia page

@flaviojs just to be clear, you are referring to the right-click option in the fleet window listing, to open the pedia page for the Ship Design corresponding to the selected ship, yes?

flaviojs:
Right click on an existing ship or a ship in the production window. Let me add images to be clearer...

ship being produced:
![ship_being_produced](https://user-images.githubusercontent.com/1009600/43390482-3b0f9228-93e6-11e8-99de-e9e393d9e634.png)

ship that can be produced:
![ship_can_be_produced](https://user-images.githubusercontent.com/1009600/43390519-4d6e0e0e-93e6-11e8-9ef4-cbbf2830f100.png)

ship that exists:
![ship_that_exists](https://user-images.githubusercontent.com/1009600/43390590-7052506a-93e6-11e8-82cd-bd9bccd375e8.png)

geoffthemedio:
I think that's looking up the design, using a temporary ship object, in all cases...?

flaviojs:
The pedia page shows "[Game - Ship Design]" in all cases.

PR #2226, opened 2018-07-28
Fix ValueRef error doing a pedia lookup of a ship after learning SHP_ADV_DAM_CONT.

flaviojs:
Fixes: #2201

### DIFF
--- a/default/scripting/techs/ship_parts/damage_control/SHP_ADV_DAM_CONT.focs.txt
+++ b/default/scripting/techs/ship_parts/damage_control/SHP_ADV_DAM_CONT.focs.txt
@@ -15,6 +15,7 @@ Tech
             scope = And [
                 Ship
                 OwnedBy empire = Source.Owner
+                InSystem
                 Stationary
                 Turn low = LocalCandidate.System.LastTurnBattleHere + 1
             ]


Dilvish-fo:
The 'Stationary' condition really should already cover the 'Insystem' constraint, it seems to me, so I think that fixing Stationary would be the better fix rather than adding a should-be-redundant InSystem check.  Though I'll confess I'm not immediately seeing the bug with the Stationary code (but I'll keep looking a bit later).

flaviojs:
Stationary checks fleet movement.
A ship that doesn't exist probably does not have a fleet, so it would default to true.

Stationary means "not moving", so true seems like an appropriate result for non existent ships.
(edit) It can also mean "is staying in a system", so false could also be appropriate... or out of scope since movement is impossible when a ship does not exist.

Dilvish-fo:
I'm not following you.  By "non existent ship" do you mean something that is not a ship, or a ship that has been destroyed, or what?  And what is the relevance here?

flaviojs:
I open the pedia page from the production window, so the ship does not exist yet.
Maybe it isn't a ship? maybe it's a design? I'm not sure how ProductionWnd does it.

Dilvish-fo:
> ...pedia page... so the ship does not exist yet.

In general, for all Pedia pages that include info on ship stats, there will be an actual ship.  For pages about ship designs, the Pedia makes temporary ship objects to read stats from.  And in your example in the cited Issue, to my understanding there is an actual ship that you are clicking on.

flaviojs:
The error happens in both cases, that's why I put "right click on a ship in the map or production windows" on the steps to reproduce.

I tried an existing ship again and the pedia page has "[Game - Ship Design]", so apparently it always creates a temporary ship.

geoffthemedio:
> The 'Stationary' condition really should already cover the 'Insystem' constraint, it seems to me

As far as I can tell, a temporary ship created in order to have an object to run effects on to estimate the meters for a ship design article will not be placed into a system or into a fleet. Since it's not in a system, it won't match InSystem, and since it doesn't have a fleet, it will be considered stationary.

I don't see why there would be an issue with existing ships, though.

Dilvish-fo:
> and since it doesn't have a fleet, it will be considered stationary

Ah, good point, thanks for clarifying that.  So (depending on whether we change how Stationary handles such temporary ships) either we'd need to (at least generally, for safety) plan to pair any use of 'Stationary' with 'InSystem' or else pair any use of 'Not Stationary' with "Not InSystem'.   Our content currently has 14 uses of Stationary, and 5 uses of "Not Stationary'.  I would prefer to adjust Stationary (consider such temporary ships as not stationary, falling through the void), but I'll grant its a rather subjective decision.

The following fixes the particular cited error for me (but as noted, should probably also be accompanied by a 'Not InSystem' adjustment for those 5 cases of 'Not Stationary', or at least have those 5 scripts checked):

```diff
diff --git a/universe/Condition.cpp b/universe/Condition.cpp
index c087011..cc3fdf6 100644
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -7987,9 +7987,15 @@ bool Stationary::Match(const ScriptingContext& local_context) const {
     // attempt to cast the candidate object to a fleet or ship, and if it's a ship
     // get the fleet of that ship
     auto fleet = std::dynamic_pointer_cast<const Fleet>(candidate);
-    if (!fleet)
-        if (auto ship = std::dynamic_pointer_cast<const Ship>(candidate))
+    if (!fleet) 
+        if (auto ship = std::dynamic_pointer_cast<const Ship>(candidate)) {myself
             fleet = GetFleet(ship->FleetID());
+            // if a ship does not have a fleet, then it must be a temporary object made for Pedia
+            // purposes, and we have concluded it works best for scripting purposes if such ships
+            // fail the Stationary condition
+            if (!fleet)
+                return false;
+        }
 
     if (fleet) {
         // if a fleet is available, it is "moving", or not stationary, if it's
```

geoffthemedio:
I would leave fleetless ships as passing Stationary, since I would consider that the default state of a ship, and the case when most effects that care if a ship is stationary would probably be active.

The question also seemingly remains about why the errors are generated for extant ships...

Dilvish-fo:
> I would leave fleetless ships as passing Stationary

OK

> The question also seemingly remains about why the errors are generated for extant ships...

I just asked for a [clarification in the referenced Issue](https://github.com/freeorion/freeorion/issues/2201#issuecomment-408746575).  To my understanding, the extant ships here are only involved as a way of pulling up the corresponding design pedia page (which always create temporary ships to read stats from).

So it looks to me like this PR is probably, depending on the result from that clarification request, fine as is.

PR #2231, opened 2018-07-29
Fix collision counter in case some player names are reserved with cookies

o01eg:
`ServerApp::IsAvailableName` checks names in cookies as well so it possible loop doesn't find available name if counter use only player from lobby data. PR defines addition tries depending on size of stored cookie.

This PR fixes issue in v0.4.8 so should be cherry-picked there.

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -852,7 +852,9 @@ sc::result MPLobby::react(const JoinGame& msg) {
 
         bool collision = true;
         std::size_t t = 1;
-        while (t <= m_lobby_data->m_players.size() + 1 && collision) {
+        while (t <= m_lobby_data->m_players.size() + server.Networking().GetCookiesSize() + 1 &&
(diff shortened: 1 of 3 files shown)


server/ServerFSM.cpp
```diff
@@ -852,7 +852,9 @@ sc::result MPLobby::react(const JoinGame& msg) {
 
         bool collision = true;
         std::size_t t = 1;
-        while (t <= m_lobby_data->m_players.size() + 1 && collision) {
+        while (t <= m_lobby_data->m_players.size() + server.Networking().GetCookiesSize() + 1 &&
```
geoffthemedio: reorder to
````
(collision &&
 t <= m_lobby_data ...
````

and indent to match

Vezzra:
@geoffthemedio, @o01eg, anything preventing this from being merged?

geoffthemedio:
no

PR #2233, opened 2018-07-29
System messages in chat

### LINKED ISSUE #2207: Treat chat messages from non user as a system messages
### Environment
* **FreeOrion Version**: Custom based on 2018-06-24.4983374 0.4.8 pre-RC2
* **Fetched as** 
  * Compiled from source

### Description
I'm trying to implement welcome message on multiplayer server. When I send message with `INVALID_PLAYER_ID` it appears as `Player -1: ...`
Could we change this behaviour for `INVALID_PLAYER_ID`?

### Expected Result
Don't try to show player name if player id is `INVALID_PLAYER_ID`.

o01eg:
Fix #2207. Chat messages with empty player name treats as a system message. History messages treats as well.
Also adds server-side check so no one could login with empty name.

The issue was assigned to `post v0.4.8` milestone but could this PR picked to 0.4.8 too so welcome messages will be available for release?

### DIFF
--- a/UI/ChatWnd.cpp
+++ b/UI/ChatWnd.cpp
@@ -385,7 +385,10 @@ void MessageWnd::HandlePlayerChatMessage(const std::string& text,
     const std::string&& formatted_timestamp = ClientUI::FormatTimestamp(timestamp);
     if (IsValidUTF8(formatted_timestamp))
         wrapped_text += formatted_timestamp;
-    wrapped_text += player_name + ": " + filtered_message + "</rgba>";
+    if (player_name.empty())
+        wrapped_text += filtered_message + "</rgba>";
+    else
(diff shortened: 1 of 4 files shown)


geoffthemedio:
I am reluctant to add new features to the release now.

Vezzra:
To add to what @geoffthemedio said: cherry picking is limited strictly to bugfixes at this point. There is always going to be this or that small feature/tweak that would be nice to squeeze into the release, but we have to draw the line somewhere, or we'll never get that release out.

0.4.8 originally was due January/February. And now it's already the end of July. We're more than half a year delayed. That shouldn't happen, even with our relaxed pace of development. We really need to finish this.

Otherwise Ragnarok might actually beat us to 1.0... :wink:

geoffthemedio:
Assuming you tested it...

PR #2234, opened 2018-07-30
Python API for storing and loading chat history

o01eg:
Introduce dummy implementation for storing and loading chat history via python script which could be altered to use an actual DBMS or so.

Introduce class `ChatProvider` implements methods `load_history` to read chat history from external source and `put_history_entity` to store it.

P.S. Example for psycopg2:
```python
from logging import info

from common.configure_logging import redirect_logging_to_freeorion_logger

# Logging is redirected before other imports so that import errors appear in log files.
redirect_logging_to_freeorion_logger()

import freeorion as fo

import psycopg2
import psycopg2.extensions
psycopg2.extensions.register_type(psycopg2.extensions.UNICODE)
psycopg2.extensions.register_type(psycopg2.extensions.UNICODEARRAY)


class ChatHistoryProvider:
    def __init__(self):
        """
        Initializes ChatProvider. Doesn't accept arguments.
        """
        dsn = ""
        with open(fo.get_user_config_dir() + "/db.txt", "r") as f:
            for line in f:
                dsn = line
                break
        self.conn = psycopg2.connect(dsn)
        info("Chat initialized")

    def load_history(self):
        """
        Loads chat history from external sources.

        Returns list of tuples:
            (unixtime timestamp, player name, text, text color of type freeorion.GGColor)
        """
        info("Loading history...")
        res = []
        with self.conn:
            with self.conn.cursor() as curs:
                curs.execute(""" SELECT date_part('epoch', ts)::int, player_name, text,
                    text_color / 256 / 256 / 256 % 256,
                    text_color / 256 / 256 % 256,
                    text_color / 256 % 256,
                    text_color % 256
                    FROM chat_history
                    ORDER BY ts
                    LIMIT 1000 """)
                for r in curs:
                    c = fo.GGColor(r[3], r[4], r[5], r[6])
                    e = (r[0], str(r[1].encode('utf-8')), str(r[2].encode('utf-8')), c)
                    res.append(e)
        return res

    def put_history_entity(self, timestamp, player_name, text, text_color):
        """
        Put chat into external storage.

        Return True if successfully stored. False otherwise.

        Arguments:
        timestamp -- unixtime, number of seconds from 1970-01-01 00:00:00 UTC
        player_name -- player name
        text -- chat text
        text_color -- freeorion.GGColor
        """
        info("Chat %s: %s %s" % (player_name, text, text_color))
        with self.conn:
            with self.conn.cursor() as curs:
                curs.execute(""" INSERT INTO chat_history (ts, player_name, text, text_color)
                    VALUES (to_timestamp(%s) at time zone 'utc', %s, %s, %s)""",
                    (timestamp, player_name, text,
                        256 * (256 * (256 * text_color.r + text_color.g) + text_color.b) + text_color.a))
        return True
```

### DIFF
--- a/default/python/chat/chat.py
+++ b/default/python/chat/chat.py
@@ -6,7 +6,7 @@ from common.configure_logging import redirect_logging_to_freeorion_logger
 redirect_logging_to_freeorion_logger()
 
 
-class ChatProvider:
+class ChatHistoryProvider:
     def __init__(self):
         """
         Initializes ChatProvider. Doesn't accept arguments.
--- a/default/python/chat/chat.py
+++ b/default/python/chat/chat.py
@@ -1,0 +1,22 @@
+from logging import info
+
+from common.configure_logging import redirect_logging_to_freeorion_logger
+
+# Logging is redirected before other imports so that import errors appear in log files.
+redirect_logging_to_freeorion_logger()
+
+
+class ChatProvider:
+    def __init__(self):
+        info("Chat initialized")
+
+    def load_history(self):
+        info("Loading history...")
+        # c = fo.GGColor(255, 128, 128, 255)
+        # e = (123456789012, "P1", "Test1", c)
+        # return [e]
+        return []
+
+    def put_history_entity(self, timestamp, player_name, text, text_color):
+        info("Chat %s: %s %s" % (player_name, text, text_color))
+        return True
--- a/default/python/chat/chat.py
+++ b/default/python/chat/chat.py
@@ -8,9 +8,18 @@ redirect_logging_to_freeorion_logger()
 
 class ChatProvider:
     def __init__(self):
+        """
+        Initializes ChatProvider. Doesn't accept arguments.
+        """
         info("Chat initialized")
 
     def load_history(self):
+        """
+        Loads chat history from external sources.
+
+        Returns list of tuples:
+            (unixtime timestamp, player name, text, text color of type freeorion.GGColor)
+        """
         info("Loading history...")
         # c = fo.GGColor(255, 128, 128, 255)
         # e = (123456789012, "P1", "Test1", c)
@@ -18,5 +27,16 @@ class ChatProvider:
         return []
 
     def put_history_entity(self, timestamp, player_name, text, text_color):
+        """
+        Put chat into external storage.
+
+        Return True if successfully stored. False otherwise.
+
+        Arguments:
+        timestamp -- unixtime, number of seconds from 1970-01-01 00:00:00 UTC
+        player_name -- player name
+        text -- chat text
+        text_color -- freeorion.GGColor
+        """
         info("Chat %s: %s %s" % (player_name, text, text_color))
         return True
--- a/default/python/README.md
+++ b/default/python/README.md
@@ -11,6 +11,8 @@ an execute_turn_events() function returning a boolean (successful completion).
 sub-module of the resource directory and can be changed with the --ai-path flag.
 * auth/  -  Python code which manages auth information stored either in a file
 or in a database.
+* chat/  -  Python code which manages chat history stored either in a file or in
+a database.
 * common/  -  Common files for code utilized by both the AI and the server.
 * handlers/  -  see handlers/README.md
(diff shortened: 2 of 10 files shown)


Cjkjvfnby:
Why is it important?

default/python/chat/chat.py
```diff
@@ -0,0 +1,22 @@
...
+        info("Chat initialized")
+
+    def load_history(self):
+        info("Loading history...")
+        # c = fo.GGColor(255, 128, 128, 255)
+        # e = (123456789012, "P1", "Test1", c)
+        # return [e]
+        return []
+
+    def put_history_entity(self, timestamp, player_name, text, text_color):
```
Cjkjvfnby, line 29: Add docstring.

Cjkjvfnby (changes requested):
Can we store chat history in savegame?

default/python/chat/chat.py
o01eg: Added

default/python/chat/chat.py
```diff
@@ -0,0 +1,22 @@
...
+# Logging is redirected before other imports so that import errors appear in log files.
+redirect_logging_to_freeorion_logger()
+
+
+class ChatProvider:
+    def __init__(self):
+        info("Chat initialized")
+
+    def load_history(self):
+        info("Loading history...")
```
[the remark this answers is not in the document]
o01eg: Added

o01eg:
> Why is it important?

Players use chat to arrange next games. Currently the server doesn't show them anything they said in chat after restarts or failures.

> Can we store chat history in savegame?

It mostly used in lobby where savegames aren't available.

default/python/chat/chat.py
```diff
@@ -0,0 +1,42 @@
+from logging import info
+
+from common.configure_logging import redirect_logging_to_freeorion_logger
+
+# Logging is redirected before other imports so that import errors appear in log files.
+redirect_logging_to_freeorion_logger()
+
+
+class ChatProvider:
```
geoffthemedio: What exactly is this providing, or will it provide in general? Just chat history? If so, maybe name it more specifically?

o01eg: Will `ChatHistoryProvider` be better name?

o01eg: Changed.

Cjkjvfnby:
> Players use chat to arrange next games. Currently the server doesn't show them anything they said in chat after restarts or failures.

I am not familiar with this chat at all.  `chat history` is the ambiguous term for me, it references to lobby chat and game chat.   Please add `lobby` prefix (or some other)  to all documentation. 

Can we just have one provider, that will store this history in the file in the logs directory  (or maybe in multiple files in the subdirectory, to have good pagination)?  This should be the game feature that not require external tools. 

PS Which platform do you use?  On Windows python shipped with the game and it fails on my machine if I try to use some external libraries (as I recall all code that use _sockets.py affected). I expect that `psycopg2` will not work for me.

o01eg:
> I am not familiar with this chat at all. chat history is the ambiguous term for me, it references to lobby chat and game chat. Please add lobby prefix (or some other) to all documentation.

Chat is common for lobby and game.

> Can we just have one provider, that will store this history in the file in the logs directory (or maybe in multiple files in the subdirectory, to have good pagination)? This should be the game feature that not require external tools.

I'm not sure how to manage and clean up multiple files. Also I'm not sure if this storage is really needed for regular games.

Dummy implementation doesn't require any external tools.

> PS Which platform do you use? On Windows python shipped with the game and it fails on my machine if I try to use some external libraries (as I recall all code that use _sockets.py affected). I expect that `psycopg2` will not work for me.

I use Linux. Script with `psycopg2`  is only example how it can be altered. It's not for shipping with the game.

geoffthemedio:
Is the current version of his supposed to actually output a db.txt and restore it when restarted? I don't get such a file in my FreeOrion user settings directory, and chat isn't restored and resent when starting a new local server.

o01eg:
No, `db.txt` is a file for an example how it could be altered. "Current version" in branch for merge don't require neither`db.txt` file nor `psycopg2` package. Actually it doesn't load and store as well.

Issue #2222, opened 2018-07-24
Fleet movement orders broken in a couple ways


Dilvish-fo:
On master (currently at 4b5c70f2b1c0416f0591a0a636b9340b98fa5fdb) the ability to enter fleet movement orders is broken in a couple ways.
- [ ] 1) @adrianbroher   after 058e12a52f673843c93e10de26ed9e15465e093f can no longer append to the current path-- the fleet movement projection lines will display as if the new travel will be appended to the current path, but the actual movement path that the fleet winds up with is instead (a) in most cases the most direct route from its current location to the final destination or (b) after  83674122283ec4617d114e3c3cbd2cb8c9cee3f2 it seems in cases where the new final destination has never been Partially visible, the total amended travel path wound up rejected and the fleet remained where it was
- [ ] 2) After 83674122283ec4617d114e3c3cbd2cb8c9cee3f2, once a fleet has move orders, can no longer order it to stop at the first system in its movepath (including the situation where the fleet is currently in a system and was just ordered to leave it, it can no longer be ordered to instead stay).

adrianbroher:
Please give proper instructions to reproduce the issues.

Dilvish-fo:
I had thought I could assume a certain level of familiarity with our standard UI.  

On Windows and Linux we set a fleet path by left-clicking on a fleet to select it, and then right-clicking on a destination system to set the path.  We append to the current path of a selected fleet by shift-right-clicking on the new destination system.  I would guess that on MacOS it's an option-click or ctrl-click instead of right-click, but I'm not sure about that.

If there is something different that is not clear to you, please be more specific.

adrianbroher:
> I had thought I could assume a certain level of familiarity with our standard UI.

No you can't. I may have some time to do some maintenance and refactoring, but taking time for actual playing the game are gone for years.

Dilvish-fo:
If a person doesn't even strive to maintain enough familiarity with the operation of the program to know how to test changes to a section of code, then I don't see how it could be a responsible action to commit changes to that section directly to master instead of posting a PR and noting the lack of testing.

adrianbroher:
You're right. I should stop bothering this project with subpar code.

Bye.

Dilvish-fo:
[...]

PR #2236, opened 2018-07-31
Fix FleetMoveOrder not generating a correct append route.

flaviojs:
Fixes the 1st part of #2222


Dilvish-fo:
Hi, this looks fine to me, and tests out fine, thanks.

PR #2237, opened 2018-07-31
Allow the FleetMoveOrder destination to be the current/next system.

flaviojs:
Fixes the 2nd part of #2222

Note that this change is dubious... considering code design, how should a "stay in the current system" order be represented in the code? and how would the AI issue it?

If a "move fleet to this system" order is an acceptable representation, then this change is probably ok.

### DIFF
--- a/util/Order.cpp
+++ b/util/Order.cpp
@@ -301,11 +301,6 @@ bool FleetMoveOrder::Check(int empire_id, int fleet_id, int dest_system_id, bool
         return false;
     }
 
-    if (dest_system_id != INVALID_OBJECT_ID && dest_system_id == start_system) {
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
This PR does seem to test out fine, but I'd like to review the code more closely than I've had time for this past week, not just the code of this PR but the code of the recent commits that made the changes that led to  the need for this fix.  Thanks for your patience.  

I'll still be pretty busy for several days, though, so I don't have any objection if someone else reviews the code & makes a merge decision, I could still follow up after that.

Dilvish-fo:
Looking at it more these changes make plenty of sense, and have tested out fine for me, so I'll go ahead and merge...

Vezzra:
With this PR merged, have all issues of #2222 been addressed? Can we close #2222?

Dilvish-fo:
Yes, that slipped my mind; it's done now.

PR #2238, opened 2018-08-02
Fixed ru.txt fr.txt en.txt

DmStrelok:
Fixed: ru.txt, fr.txt, en.txt.
Incorrect damage and amount of interceptors.


geoffthemedio:
looks ok

PR #2245, opened 2018-08-11
Reformat files with the Black  [Not for merge]

Cjkjvfnby:
This PR is not for the merge, but for discussion. 

All changes were done by the external tool.    [Black](https://github.com/ambv/black)

Important feature:
```
Also, as a temporary safety measure, Black will check that the reformatted code still produces a valid AST that is equivalent to the original. This slows it down. If you're feeling confident, use --fast.
```
Formatting can be safely applied to code, without review. 

I don't like this formatting but it is ok for me to switch to it and yes it confronts some our style rules.  Good point that no need to write and follow any rules, just press the button.

So maybe we can switch to this code style for the project? 
@Dilvish-fo , @Morlic-fo , what do you think?

There is space for automation, but let's not discuss it in this thread.

PS. files with `print ` will not be reformated.

### DIFF
--- a/default/python/tests/AI/character_test/character_strings_test.py
+++ b/default/python/tests/AI/character_test/character_strings_test.py
@@ -1,8 +1,5 @@
 from character.character_module import Trait, Character, Aggression
-from character.character_strings_module import (
-    _make_aggression_based_function,
-    get_trait_name_aggression,
-)
+from character.character_strings_module import _make_aggression_based_function, get_trait_name_aggression
 
 
 def test_make_aggression_based_function():
@@ -40,7 +37,4 @@ def test_get_trait_name_aggression():
     assert get_trait_name_aggression(character) == "UserString GSETUP_TURTLE"
 
     nonaggressive_character = Character([UnknownTrait()])
-    assert (
-        get_trait_name_aggression(nonaggressive_character)
-        == "UserString UNKNOWN_VALUE_SYMBOL"
-    )
+    assert get_trait_name_aggression(nonaggressive_character) == "UserString UNKNOWN_VALUE_SYMBOL"
--- a/default/python/tests/AI/character_test/character_strings_test.py
+++ b/default/python/tests/AI/character_test/character_strings_test.py
@@ -1,5 +1,8 @@
 from character.character_module import Trait, Character, Aggression
-from character.character_strings_module import _make_aggression_based_function, get_trait_name_aggression
+from character.character_strings_module import (
+    _make_aggression_based_function,
+    get_trait_name_aggression,
+)
 
 
 def test_make_aggression_based_function():
@@ -37,4 +40,7 @@ def test_get_trait_name_aggression():
     assert get_trait_name_aggression(character) == "UserString GSETUP_TURTLE"
 
     nonaggressive_character = Character([UnknownTrait()])
-    assert get_trait_name_aggression(nonaggressive_character) == "UserString UNKNOWN_VALUE_SYMBOL"
+    assert (
+        get_trait_name_aggression(nonaggressive_character)
+        == "UserString UNKNOWN_VALUE_SYMBOL"
+    )
--- a/default/python/tests/AI/character_test/character_test.py
+++ b/default/python/tests/AI/character_test/character_test.py
@@ -9,9 +9,7 @@ class LeftTrait(Trait):
 class RightTrait(Trait):
     """A test Trait that injects values to be found by combiners"""
 
-    def may_explore_system(
-        self, monster_threat
-    ):  # pylint: disable=no-self-use,unused-argument
+    def may_explore_system(self, monster_threat):  # pylint: disable=no-self-use,unused-argument
         """Always reject. 1 parameter"""
         return False
 
@@ -19,9 +17,7 @@ class RightTrait(Trait):
         """Always reject. 0 parameters"""
         return False
 
-    def preferred_research_cutoff(
-        self, alternatives
-    ):  # pylint: disable=no-self-use,unused-argument
+    def preferred_research_cutoff(self, alternatives):  # pylint: disable=no-self-use,unused-argument
         """Pick 1st"""
         return alternatives[1]
 
@@ -79,10 +75,6 @@ class TestCharacter(object):
         assert rejection_character.preferred_research_cutoff([10, 11, 12]) == 11
 
     def test_character_must_be_composed_of_traits(self):
-        with pytest.raises(
-            TypeError,
-            message="Expect TypeError",
-            match="All traits must be sub-classes of Trait",
-        ):
+        with pytest.raises(TypeError, message="Expect TypeError", match="All traits must be sub-classes of Trait"):
             not_a_trait = int(1)
             Character([LeftTrait, not_a_trait])
--- a/default/python/tests/AI/character_test/character_test.py
+++ b/default/python/tests/AI/character_test/character_test.py
@@ -9,7 +9,9 @@ class LeftTrait(Trait):
 class RightTrait(Trait):
     """A test Trait that injects values to be found by combiners"""
 
-    def may_explore_system(self, monster_threat):  # pylint: disable=no-self-use,unused-argument
+    def may_explore_system(
+        self, monster_threat
+    ):  # pylint: disable=no-self-use,unused-argument
         """Always reject. 1 parameter"""
         return False
 
@@ -17,7 +19,9 @@ class RightTrait(Trait):
         """Always reject. 0 parameters"""
         return False
 
-    def preferred_research_cutoff(self, alternatives):  # pylint: disable=no-self-use,unused-argument
+    def preferred_research_cutoff(
+        self, alternatives
+    ):  # pylint: disable=no-self-use,unused-argument
         """Pick 1st"""
         return alternatives[1]
 
@@ -75,6 +79,10 @@ class TestCharacter(object):
         assert rejection_character.preferred_research_cutoff([10, 11, 12]) == 11
 
     def test_character_must_be_composed_of_traits(self):
-        with pytest.raises(TypeError, message='Expect TypeError', match='All traits must be sub-classes of Trait'):
+        with pytest.raises(
+            TypeError,
+            message="Expect TypeError",
+            match="All traits must be sub-classes of Trait",
+        ):
             not_a_trait = int(1)
             Character([LeftTrait, not_a_trait])
--- a/default/python/tests/AI/test_parse_ai_tag_grade.py
+++ b/default/python/tests/AI/test_parse_ai_tag_grade.py
@@ -2,16 +2,16 @@ from freeorion_tools import get_ai_tag_grade
 
 
 def test_parse_weapon_tag_positive1():
-    assert 'GREAT' == get_ai_tag_grade(['GREAT_WEAPONS'], 'WEAPONS')
+    assert "GREAT" == get_ai_tag_grade(["GREAT_WEAPONS"], "WEAPONS")
 
 
 def test_parse_weapon_tag_positive2():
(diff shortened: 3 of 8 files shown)


Dilvish-fo:
> This PR is not for the merge, but for discussion... There is space for automation, but let's not discuss it in this thread.

Eh? My best guess of what you want is... actually, I had to erase my best guess because it wasn't making enough sense to me.  I'll just discuss the things that seem sensible to me to discuss, and you can then say if you think some part would be better as a separate discussion.  

> I don't like this formatting but it is ok for me

I like some of the changes just fine, others less so, at least so far.  I take it you produced this PR using the default setting of line-length 88; I think it would be a more useful reference for our discussion if you re-ran with ```--line-length 120```.  

I see plenty of value for adopting a code-style and corresponding tool like this that can both be used within PyCharm and used at the command line level both for individuals doing coding, and for a ```--check``` check by Travis.  (Have you looked into whether that should replace or instead augment the pylint check?  I am suspecting that it should probably augment pylint, but then we'd probably need pylint configuration settings to ensure pylint respected the Black formatting).

They seem to have put a lot of good thought into their formatting rules, but I am not seeing any indication of good thought in their opposition to allowing for configurable rules, it rather looks like an ego issue:
> Before you suggest a new feature or configuration knob...if your answer is "because I don't like a particular formatting" then you're not ready to embrace Black yet.

Their licensing ("MIT license") is unfortunately apparently [a bit ambiguous according to gnu.org](https://www.gnu.org/licenses/license-list.en.html); both of the possible referents specifically identified by gnu.org are GPL compatible, and the primary need is to just use the product from their code, so most of the licensing details don't really matter.  But the gnu.org writeup seems to indicate their may be even more different licenses that could be called an "MIT license", so it would probably be best to push Black to be more specific about their license.

It would be a much easier decision about adopting this tool if they allowed for the possibility for greater configuration of project formatting rules.

Dilvish-fo:
I think that running with --line-length 120 ___after___ having previously run with --line-length 88 is leading to a bunch of lines being indicated as changed (due to some slight change in quotation mark type, it appears to me) that would probably not have been changed if you simply ran with --line-length 120 from the get-go. 

If you want to be able to compare the two different results, then you could have https://github.com/freeorion/freeorion/pull/2245/commits/c6397f2aa81afbc09379d938a8ef87c7cbd91696 and then an additional commit that reverts https://github.com/freeorion/freeorion/pull/2245/commits/c6397f2aa81afbc09379d938a8ef87c7cbd91696 prior to running with --line-length 120

Or else you could just reset to immediately prior to https://github.com/freeorion/freeorion/pull/2245/commits/c6397f2aa81afbc09379d938a8ef87c7cbd91696  and then run with  --line-length 120

Cjkjvfnby:
> Have you looked into whether that should replace or instead augment the pylint check?

It will not replace checks for variable names.

> They seem to have put a lot of good thought into their formatting rules, but I am not seeing any indication of good thought in their opposition to allowing for configurable rules, it rather looks like an ego issue:

And what they said for that:

```
In return, Black gives you speed, determinism, 
and freedom from pycodestyle nagging about formatting.
You will save time and mental energy for more important matters.  
```
I think it might be a good idea.

> Their licensing ("MIT license") ...

I am not good at licensing.   I understand only a few words (`licensing details don't really matter`).

Cjkjvfnby:
> I think that running with --line-length 120 after having previously run with --line-length 88 is leading to a bunch of lines being indicated as changed 

Black ignores the previous formatting.  So I expect the exact same result.  (Tested, it is same)

Dilvish-fo:
> And what they said for that:

Those look to me like reasons to have a tool generally like Black, but I see nothing about them that is directed to the issue of not allowing any configuration beyond line length.

Cjkjvfnby:
> to the issue of not allowing any configuration beyond line length.

I see that it is the main feature of the Black to not allowing any configuration beyond line length.

Dilvish-fo:
> I see that it is the main feature of the Black to not allowing any configuration beyond line length.

How is the lack of choice/configuration a feature?

It seems to me that the makers have simply invoked a kind of coding marketing-machismo ```then you're not ready to embrace Black yet``` while somehow ignoring that in the original quote from which they draw their name, the absurdity of pretending it was great to only offer the color black was a core part of the humor. 

It's just as absurd here, and will significantly and unnecessarily impede the adoption of their tool.  Even PEP8 and the Google Python Style Guide both acknowledge that at least in some cases it's more important to stay consistent with a project code style than with the guide recommendations.  This no-configuration aspect of Black prevents anyone in a project from using it unless they get signoff to totally change  the entire project style over to that of Black, which will inherently slow down and/or outright prevent adoption in a lot of cases. 

Sure, I suppose that if they allowed configuration then they'd have to change their name and lose their amusing pun.  But that still seems mostly like an ego issue, and carries a nontrivial price.  Since they seem to have trouble reassessing previously made decisions, it doesn't make it very appealing to just rely on their judgement about style coding rules.    

But suppose we decide we could live with their style decisions -- what do we really get from that that we couldn't already get from just using something like PyCharm's autoformatting? (Which has seemed mostly-decent-but-sometimes-needing-help.) Perhaps we'd like Black's style choices better than PyCharm's?  But at least with PyCharm's autoformatting you can just apply it to particular lines, and can manually format other lines where they autoformatting didn't work out so well.

Cjkjvfnby:
> How is the lack of choice/configuration a feature?

Everything that planed is a feature. 

>  This no-configuration aspect of Black prevents anyone in a project from using it unless they get signoff to totally change the entire project style over to that of Black, which will inherently slow down and/or outright prevent adoption in a lot of cases.

We have code style and code base did not match it (long lines). Black is one of the possible solutions. You can just press the button and got code that 100% matches code style (if this code style is black).

Current code style check and reformating are not 100% automated. Black is 100% automated in checks and formatting. 

It this point in my life, easy maintenance is more valuable than my personal preferences to code style.

> But suppose we decide we could live with their style decisions -- what do we really get from that that we couldn't already get from just using something like PyCharm's autoformatting? 

Because I cannot apply Pycharm reformatting globally.  Only to the specific place in the code. I cannot setup it to match our code style. 

Since Pycharm has settings, apply it to some part of the code may produce unexpected diff with other developers reformating, because of the different settings or Pycharm version. 

>  But at least with PyCharm's autoformatting you can just apply it to particular lines, and can manually format other lines where they autoformatting didn't work out so well.

Why do you want to do manually thing that can be done by computer? 

Pycharm use case: 
- select
-  press hotkey
- check result (undo unexpected changes)
- commit

Black use case:
- press hotkey or use pre-commit hook
- commit

Dilvish-fo:
> Why do you want to do manually thing that can be done by computer?

I think my words pretty clearly only referred to the situation where the computer didn't do the task well.

It does sound like Black is generally considered to mostly do the task well, though when it's adding spaces to a series of operations it doesn't group things nearly as well as a human could, and it seems like their whole set-in-stone attitude means they wouldn't even try to improve that.  Perhaps it's still worth accepting it as-is, I'll mull on it more.

If we/re going to consider adopting the use of an autoformatter, we should probably also at least also consider [yapf (Yet Another Python Formatter)](https://github.com/google/yapf), which is maintained by Google.  It is configurable, and has a handful of preset styles.

Cjkjvfnby:
> If we/re going to consider adopting the use of an autoformatter, we should probably also at least also consider yapf (Yet Another Python Formatter), 

Looks like it handles the most important case with a binary operator. :) Will take a closer look at it. 

> which is maintained by Google

It is on the bottom of the page you referenced:
```
YAPF is not an official Google product (experimental or otherwise), it is just code that happens to be owned by Google. 
```

Cjkjvfnby:
- I did not find if `yapf` compare bytecode before and after formatting.  Thie feature is important for me since codebase is huge and this allows to skip the review. 
- Work with binary operators are done in a dumb way, you just specify set of operators, that will not have spaces around.  Nothing about grouping them logically.
- `black` is deterministic, the result does not take the previous formatting in the account, `yapf` depends on the previous formatting.
- I did not found features that present in `black` and present in `yapf`.  
  - Format list with items, each on a separate line with trailing comma.   This should reduce diff in case of adding new items.
  - Quote consistency (always use double quotes, except cases when the double quote is in literal.)

If you want I can create some script that will produce the human-readable output of format examples.

Dilvish-fo:
> I did not find if yapf compare bytecode before and after formatting. Thie feature is important for me since codebase is huge and this allows to skip the review.

As you noted in your intro at the top, Black apparently only temporarily has that extra testing since their code is new, and they apparently don't plan to keep it once they (and people in general) are satisfied with the reliability.  yapf has been around getting tested a fair bit longer; I don't recall seeing anyone cite concerns about yapf's reliability.   But still, I suppose that for the period of time that Black still does that, it is a bit of a merit for Black, so long as you'd actually have some reasonable handling for the case that it fails their check. 

> Work with binary operators are done in a dumb way, you just specify set of operators, that will not have spaces around. Nothing about grouping them logically.

I'm pretty sure I've seen examples that could not have resulted from such a simple rule.  I'll try to look into it more.

> black is deterministic, the result does not take the previous formatting in the account, yapf depends on the previous formatting.

It's my understanding that it is only in rather rare situations that yapf(yapf(source)) != yapf(source), but yes, that is still a bit of a ding against yapf.

> I ~~did~~ ~~not~~ found features that _are_ present in black and _not_ present in yapf.

I assume that's what you meant?

>  - Format list with items, each on a separate line with trailing comma. This should reduce diff in case of adding new items.

One of the google guys who maintain yapf has [invited a PR on that, but noted that their experience indicates that kind of change to be "fraught with problems" and would need to be well tested.](https://github.com/google/yapf/issues/274#issuecomment-251029262)  That was a while ago with no movement, but perhaps the existence of Black will now help prompt a PR (assuming no bugs get discovered with Black's handling).  But in the meantime, yes it does appear to be an advantage for Black.

> Quote consistency (always use double quotes, except cases when the double quote is in literal.)

That aspect of Black actually seems to get as much criticism as approval.  We don't currently have a kind of quote policy that leads people to particularly  dislike that aspect of Black, but I don't think that converts it into an actual advantage for Black either.  When I look over all the places this diff shows Black changing single quotes to double quotes, I don't react with "Oh, I'm glad it's doing that!" for any of them.

> It is on the bottom of the page you referenced:
"YAPF is not an official Google product (experimental or otherwise), it is just code that happens to be owned by Google. "

I don't understand your point.  yapf was developed internally at Google, for their own use, and then made available openly.   Does it concern you if they don't consider it an "official product" and don't promise to necessarily keep maintaining it?

Cjkjvfnby:
I am writing code with jupyter to visualize difference, I don't know how to share this code it yet.

Example with binary operators.

*original*
```python
2+2*8
2*2*4
```
*black*
```python
2 + 2 * 8
2 * 2 * 4
```
*yapf*
```python
2 + 2*8
2*2*4
```

Cjkjvfnby:
@Dilvish-fo , are you common with Jupyter notebook (former IPython notebook)?

[Here is](https://gist.github.com/Cjkjvfnby/ac225978204d8d0a323cd7ce5d6dec5d) the document created with it. If you want to play with it, you need to [install jupyter](http://jupyter.org/install.html),  download gist and open it in Jupyter  (run `jupyter notebook`  in the same directory as the file)

Dilvish-fo:
> you common with Jupyter notebook

I have used them some.  Little sidenote-- you can run them directly from within Pycharm: it's a little tricky, because even when I was sure I had the server token right pycharm would not be able to connect, but it turns out you can just delete the token part of the server URL it first suggests for you, and then when that fails up by the title tab will be a clickable offer for Pycharm to start up a jupyter server to run the notebook.  

I'll play around a bit with your notebook, thanks.

Vezzra:
@Cjkjvfnby, which is the superseding PR? CAn you put a reference to it here?

Cjkjvfnby:
> which is the superseding PR? CAn you put a reference to it here?

Not by PR but by issue #2306. This PR was done not for merging but for discussion with code examples.

PR #2248, opened 2018-08-14
Populate lobby with AIs from saved game

### LINKED ISSUE #1810: On multiplayer loaded games, AI should re-appear in the multiplayer setup window
### Environnement
freeorion version 0.4.7.1

### Description
when in multiplayer mode, when i load a game, I have to manually add the AI just like they were here before.

### Expected Result
 I'd like the AI fields/lines to be pre-populated when I load a game (in my case : 2 human players + 2AI)

thanks

o01eg:
Fix #1810.
Reads information from saved game and populates lobby with AIs.

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -1273,7 +1273,9 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
             psd.second.m_save_game_empire_id = ALL_EMPIRES;
         }
 
-        // remove all AIs
+        // remove all AIs from current lobby data,
+        // so that when the save is loaded no AI state as appropriate,
+        // without having potential extra AIs lingering from the previous
         m_lobby_data->m_players.remove_if([](const std::pair<int, PlayerSetupData>& plr) {
             return plr.second.m_client_type == Networking::CLIENT_TYPE_AI_PLAYER;
         });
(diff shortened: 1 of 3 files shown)


o01eg:
Actually it looks more `component:network` then `component:UI` as I made zero changes in UI.

Vezzra:
Ok, added "component:network", but "component:UI" should stay in place. From a user's perspective this is more of an UI issue, which is why #1810 is tagged as such.

server/ServerFSM.cpp
```diff
@@ -1273,11 +1273,40 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
...
         // refresh save game empire data
         boost::filesystem::path save_dir(GetServerSaveDir());
+        std::vector<PlayerSaveHeaderData> player_save_header_data;
         try {
             LoadEmpireSaveGameData((save_dir / m_lobby_data->m_save_game).string(),
-                                   m_lobby_data->m_save_game_empire_data);
+                                   m_lobby_data->m_save_game_empire_data,
+                                   player_save_header_data);
+
+            for (const auto& pshd : player_save_header_data) {
```
geoffthemedio: add a comment before this block explaining its purpose

server/ServerFSM.cpp
```diff
@@ -1273,11 +1273,40 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
             psd.second.m_save_game_empire_id = ALL_EMPIRES;
         }
 
+        // remove all AIs
```
geoffthemedio: clarify: from current lobby data, so that when the save is loaded they can be re-added from a fresh / no AI state as appropriate, without having potential extra AIs lingering from the previous / abandoned game setup...

server/ServerFSM.cpp
```diff
@@ -1273,11 +1273,40 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
...
+                                   player_save_header_data);
+
+            for (const auto& pshd : player_save_header_data) {
+                if (pshd.m_client_type == Networking::CLIENT_TYPE_AI_PLAYER) {
+                    const auto& empire_data_it = m_lobby_data->m_save_game_empire_data.find(pshd.m_empire_id);
+                    if (empire_data_it != m_lobby_data->m_save_game_empire_data.end()) {
+                        PlayerSetupData player_setup_data;
+                        player_setup_data.m_player_id =     Networking::INVALID_PLAYER_ID;
+                        // don't use original names to prevent collision with manually added AIs
+                        player_setup_data.m_player_name =   UserString("AI_PLAYER") + "_" + std::to_string(m_ai_next_index++);
```
geoffthemedio: would be better to use a substitution rather than appending for combining "AI Player" with a number, but that can be done separately.

server/ServerFSM.cpp
```diff
@@ -1273,11 +1273,40 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
...
         boost::filesystem::path save_dir(GetServerSaveDir());
+        std::vector<PlayerSaveHeaderData> player_save_header_data;
         try {
             LoadEmpireSaveGameData((save_dir / m_lobby_data->m_save_game).string(),
-                                   m_lobby_data->m_save_game_empire_data);
+                                   m_lobby_data->m_save_game_empire_data,
+                                   player_save_header_data);
+
+            for (const auto& pshd : player_save_header_data) {
+                if (pshd.m_client_type == Networking::CLIENT_TYPE_AI_PLAYER) {
```
geoffthemedio: prefer
````
if (false)
    continue;

// more stuff not indented
````

File: server/ServerFSM.cpp [hunk 5]
```diff
@@ -1273,11 +1273,40 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
...
+                    const auto& empire_data_it = m_lobby_data->m_save_game_empire_data.find(pshd.m_empire_id);
+                    if (empire_data_it != m_lobby_data->m_save_game_empire_data.end()) {
```
geoffthemedio: same; avoid unnecessary indenting

server/ServerFSM.cpp
o01eg: Added.

server/ServerFSM.cpp

server/ServerFSM.cpp
o01eg: It's a same expression used in other functions where new AI was appended.

server/ServerFSM.cpp

server/ServerFSM.cpp

server/ServerFSM.cpp
geoffthemedio: I realize... doesn't mean it can't be improved for better multilingual compatibility.

o01eg: I suppose it better to do in other PR.

geoffthemedio: yes; that's what I meant by "can be done separately"

PR #2249, opened 2018-08-18
Reroll wrong targets with preferred prey conditions

agrrr3:
This pull request implements target preferences for weapons and fighters by retrying a different enemy if a wrong enemy is in scope.

Target preferences and a precision are specified in the weapon and hangar parts. 
Preferences are specified as condition to match against the enemy vessel in scope.

Multiple source can contribute weighted preference conditions. In the scope of this PR are detection ship parts (apply to all weapon systems), hangar parts (apply to the hangar's  boats) and short range weapon parts (apply to the part's weapon).

[design discussion in forum](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11048) 

Content setup:
planets - kind of GOOD precision ignoring space boats
Flak (SR_WEAPON_0_1) - GREAT precision hunting space boats
Bombers (FT_HANGAR_3) - GOOD precision hunting space ship
Fighters (FT_HANGAR_2) - AVERAGE precision at their worst enemy - ships with flak
Interceptors (FT_HANGAR_1) - GOOD precision hunting space boats
Spinal cannon (SR_SPINAL_ANTIMATTER) - GOOD Precision ignoring ships with low structure
Short range weapons (SR_WEAPON_1,2,3,4) - AVERAGE precision ignoring decoys
Everything else - no preferences

The code is functional ~~but not in a good shape. It needs cleanup but also in order to get the implementation going i was stumbling against the parser library and automagic pointers~~.


geoffthemedio:
I would rather implement prioritized targeting using parts in the ship design, as discussed in various places on the forums...

agrrr3:
@geoffthemedio well there were different things discussed including multiple sources of preferences e.g. targetting part, fleet preferences etc.. i think that for specialised weapon subsystems like flak (and probably fighters) targetting should not be determined by other ship parts. So i think weapon based preferences would still be necessary even with targetting parts. Would you mind joining the discussion in the forum or is this design to far from inclusion?

combat/CombatSystem.cpp
```diff
@@ -748,9 +760,9 @@ namespace {
                 int shots = static_cast<int>(ship->CurrentPartMeterValue(METER_SECONDARY_STAT, part_name)); // secondary stat is shots per attack)
                 if (part_attack > 0.0f && shots > 0) {
                     // attack for each shot...
-		  Targetting::TriggerConditions preferredTargets(Targetting::findPreferredTargets(part_name), part->Precision());
-		  for (int shot_count = 0; shot_count < shots; ++shot_count)
-		      retval.push_back(PartAttackInfo(part_class, part_name, part_attack, part->Precision(), preferredTargets));
+                    auto preferredTargets = Targetting::combine(design_preferred_prey, Targetting::findPreferredTargets(part_name), part->Precision());
```
geoffthemedio: Use `underscore_separated_local_variable_names`

agrrr3:
Ok, i think this again functional complete after adding precision/preferredPrey for detection parts.

Now i guess i should move from "make it work" to "make it right". Some code needs to be pruned - the notion that a universe object has a target-type is not used anymore. Also the way TriggerConditions are cut is not as helpful as i intended. Also i need to look again object ownership etc.

I plan to move most of the targetting code from CombatSystem into CombatTargetting.

The anti-comsat clause probably needs to be refined.

So a rough code review would be great. How can i get one?

combat/CombatSystem.cpp
```diff
@@ -273,22 +275,44 @@ namespace {
             part_type_name(part_name_),
             part_attack(part_attack_),
             fighters_launched(0),
-            fighter_damage(0.0f)
+            fighter_damage(0.0f),
```
geoffthemedio: prefer moving default values such as this (with any preexisting) into the declarations rather than extending the initialization list.

combat/CombatSystem.cpp
```diff
@@ -273,22 +275,44 @@ namespace {
...
         {}
 
         ShipPartClass   part_class;
         std::string     part_type_name;
         float           part_attack;        // for direct damage parts
         int             fighters_launched;  // for fighter bays, input value should be limited by ship available fighters to launch
         float           fighter_damage;     // for fighter bays, input value should be determined by ship fighter weapon setup
+        int             precision;          //
+
+       Targetting::TriggerConditions preferred_targets;
```
geoffthemedio: weird indenting

combat/CombatSystem.cpp
```diff
@@ -273,22 +275,44 @@ namespace {
...
+
+        bool AimsFor(std::shared_ptr<UniverseObject> target) const
+        {
```
geoffthemedio: move all onto single line with `{ ... }`

combat/CombatSystem.cpp
```diff
@@ -273,22 +275,44 @@ namespace {
...
         ShipPartClass   part_class;
         std::string     part_type_name;
         float           part_attack;        // for direct damage parts
         int             fighters_launched;  // for fighter bays, input value should be limited by ship available fighters to launch
         float           fighter_damage;     // for fighter bays, input value should be determined by ship fighter weapon setup
+        int             precision;          //
+
+       Targetting::TriggerConditions preferred_targets;
+
+        bool AimsFor(std::shared_ptr<UniverseObject> target) const
```
geoffthemedio: meaning of this function name unclear

combat/CombatSystem.cpp
```diff
@@ -704,13 +728,27 @@ namespace {
...
         float fighter_attack = 0.0f;
+        Targetting::TriggerCondition fighter_preferred_prey;
+        Targetting::Precision fighter_precision = 0;
         std::map<std::string, int> part_fighter_launch_capacities;
 
         // determine what ship does during combat, based on parts and their meters...
-        for (const auto& part_name : design->Parts()) {
+        for (const auto& part_name : design->Parts()) { // first round - collect information
+            const PartType* part = GetPartType(part_name);
+            if (!part)
```
geoffthemedio: should probably have an error logged in this case, unless `part_name` is empty.

combat/CombatSystem.cpp
```diff
@@ -722,15 +760,18 @@ namespace {
                 int shots = static_cast<int>(ship->CurrentPartMeterValue(METER_SECONDARY_STAT, part_name)); // secondary stat is shots per attack)
                 if (part_attack > 0.0f && shots > 0) {
                     // attack for each shot...
+                    auto preferredTargets = Targetting::combine(design_preferred_prey, Targetting::findPreferredTargets(part_name), part->Precision());
```
geoffthemedio: `preferredTargets` -> `preferred_targets`

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+            return BoatPrey;
+        }
+        if (part_name.rfind("SR_", 0) == 0) {
+            return ShipPrey;
+        }
+        // shouldnt happen
+        return AllPrey;
+}
+
+Targetting::TriggerCondition Targetting::findPreferredTargets(const std::string& part_name) {
```
geoffthemedio: `CamelCaseFunctionNames`

combat/CombatSystem.cpp
```diff
@@ -1344,15 +1414,16 @@ namespace {
...
                                 << " with damage " << weapon.fighter_damage
+				<< " with precision: " << weapon.precision
                                 << " for empire id: " << attacker_owner_id
                                 << " from ship id: " << attacker->ID();
 
+
             std::vector<int> new_fighter_ids =
                 combat_state.AddFighters(weapon.fighters_launched, weapon.fighter_damage,
-                                         attacker_owner_id, attacker->ID(), species_name);
+                                         attacker_owner_id, attacker->ID(), species_name, weapon.precision, weapon.preferred_targets);
```
geoffthemedio: move new params to new line to avoid making the line overly long

combat/CombatSystem.cpp
```diff
@@ -1344,15 +1414,16 @@ namespace {
                 continue;
 
             int attacker_owner_id = attacker->Owner();
-
             DebugLogger(combat) << "Launching " << weapon.fighters_launched
                                 << " with damage " << weapon.fighter_damage
+				<< " with precision: " << weapon.precision
```
geoffthemedio: align indentation

combat/CombatSystem.cpp
```diff
@@ -825,13 +868,14 @@ namespace {
         }
 
         std::vector<int> AddFighters(int number, float damage, int owner_empire_id,
-                                     int from_ship_id, const std::string& species)
+                                     int from_ship_id, const std::string& species,
+                                     Targetting::Precision precision, const Targetting::TriggerConditions& preferred_prey)
```
geoffthemedio: move last param to new line

combat/CombatSystem.cpp
```diff
@@ -825,13 +868,14 @@ namespace {
...
         {
             std::vector<int> retval;
 
             for (int n = 0; n < number; ++n) {
                 // create / insert fighter into combat objectmap
-                auto fighter_ptr = std::make_shared<Fighter>(owner_empire_id, from_ship_id, species, damage);
+                auto fighter_ptr = std::make_shared<Fighter>(owner_empire_id, from_ship_id, species, damage, precision, preferred_prey);
```
geoffthemedio: move new params to new line to avoid overly long line

combat/CombatSystem.cpp
```diff
@@ -1266,12 +1310,38 @@ namespace {
...
             // select target object
-            int target_idx = RandInt(0, valid_target_ids.size() - 1);
-            int target_id = *std::next(valid_target_ids.begin(), target_idx);
+            std::shared_ptr<UniverseObject> target;
+            DebugLogger() << "Target precision: " << std::to_string(weapon.precision);
+            for (int t = 1; t < weapon.precision ; t++) {
+                // Do all but the last round of finding a valid target; the simple last round is separate for performance reasons
+                DebugLogger() << "Find target - Round: " << std::to_string(t) << " (try to lock it if it matches preferredPrey conditions)";
+                int target_idx = RandInt(0, valid_target_ids.size() - 1);
+                int target_id = *std::next(valid_target_ids.begin(), target_idx);
```
geoffthemedio, line 1314: is this OK if `valid_target_ids` has just one element?

combat/CombatSystem.cpp
```diff
@@ -1266,12 +1310,38 @@ namespace {
...
 
-            auto target = combat_state.combat_info.objects.Object(target_id);
+                if (weapon.AimsFor(target)) {
+                    // shooting
+                    break;
+                }
+                target = nullptr; // signal that last round should be executed
+            }
+            if (!target) {
+                DebugLogger() << "Find target - Round: " << std::to_string(weapon.precision) << " (Final round ignores preferredPrey condition)";
```
geoffthemedio: comment this block?

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "../universe/Ship.h"
+#include "../universe/ShipDesign.h"
+#include "../universe/Fighter.h"
+#include "../universe/ValueRef.h"
+
+#include "../util/Logger.h"
+#include "../util/Random.h"
+
+bool Targetting::isPreferredTarget(Targetting::TriggerConditions conditions,
+                                   Targetting::Target target) {
```
geoffthemedio: `{` to newline

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+    const Precision highest_weight = Targetting::findPrecision(conditions);
+    for (long unsigned int i = 0; i < conditions.conditions.size(); i++) {
```
geoffthemedio, line 19: prefer `for (const auto& condition : conditions.conditions) {` unless you need the `i` index specifically...

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+        const auto& condition = conditions.conditions[i];
+        if (!condition) {
+            DebugLogger() << "No preferences. Target is perfect." ;
+            return true;
+        }
+
+        if (Targetting::isPreferredTarget(condition, target)) {
```
geoffthemedio: capitalize `Is`

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "../util/Logger.h"
+#include "../util/Random.h"
+
+bool Targetting::isPreferredTarget(Targetting::TriggerConditions conditions,
+                                   Targetting::Target target) {
+    const Precision highest_weight = Targetting::findPrecision(conditions);
+    for (long unsigned int i = 0; i < conditions.conditions.size(); i++) {
+        const auto& condition = conditions.conditions[i];
+        if (!condition) {
+            DebugLogger() << "No preferences. Target is perfect." ;
```
geoffthemedio, line 22: Target is perfect? Not target is irrelevant?

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+            return true;
+        }
+
+        if (Targetting::isPreferredTarget(condition, target)) {
+            const int upper = std::round(highest_weight * 1.5f); // HEURISTIC
+            const int improbability = RandInt(0, upper);
+            const int luck = conditions.weights[i];
+            DebugLogger() << "Prioritized locking attempt" <<  [&improbability, &luck, &upper](){
```
geoffthemedio: don't think you need a lambda for this... you're just using `<<` into a `stringstream` but the logger is already a stream.

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+            } else {
+                DebugLogger() << "Couldnt acquire lock." ;
+            }
+        }
+    }
+    return false;
+}
+
+bool Targetting::isPreferredTarget(Targetting::TriggerCondition condition,
+                                   Targetting::Target target) {
```
geoffthemedio: `{` to newline

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+    bool is_preferred = condition->Eval(target);
+    auto obj_type = target->ObjectType();
+    if ( is_preferred ) {
+        DebugLogger() << "This " << obj_type << " ship should die! " ;
+    } else {
+        DebugLogger() << "Don't care about this " << obj_type << " - reroll if possible. " ;
+    }
+    return is_preferred;
+}
+Targetting::TriggerCondition Targetting::preyAsTriggerCondition(const Targetting::Prey prey) {
```
geoffthemedio: separate function definitions with blank lines

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+        switch( prey ) {
```
geoffthemedio: be consistent with `switch(...)` or `switch (...)`. I forget which is preferred based on existing code.

`( prey )` -> `(prey)`

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                   Targetting::Target target) {
+    if (!condition) {
+            DebugLogger() << "No preferences. Target is perfect." ;
+            return true;
+    }
+
+    DebugLogger() << "Evaluate preferred prey condition" << condition->Description();
+    bool is_preferred = condition->Eval(target);
+    auto obj_type = target->ObjectType();
+    if ( is_preferred ) {
```
geoffthemedio: remove spaces around `is_preferred`

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+        case Targetting::NoPreference:
+                return std::make_shared<Condition::All>();
+                break;
+        default:
+                // error
+                return nullptr;
+        }
+}
+Targetting::TriggerConditions Targetting::preyAsTriggerConditions(const Targetting::Prey prey) {
+    return Targetting::TriggerConditions(Targetting::preyAsTriggerCondition(prey), 1);
```
geoffthemedio: `{ }` on same line as single line function definition

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+Targetting::Prey Targetting::findTargetTypes(const std::string& part_name) {
+        if (part_name.rfind("FT_", 0) == 0) {
```
geoffthemedio: don't need `{ }`

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+            return BoatPrey;
+        }
+        if (part_name.rfind("SR_", 0) == 0) {
```
geoffthemedio: look up `PartType` and check its class; don't test the content scripted name prefix.

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+            return ShipPrey;
+        }
+        // shouldnt happen
```
geoffthemedio: shouldn't

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+Targetting::Precision Targetting::findPrecision(const Targetting::TriggerConditions& conditions) {
+    auto result = std::max_element(conditions.weights.begin(), conditions.weights.end());
+    return *result;
+}
+
+Targetting::Precision Targetting::findPrecision(const Targetting::HunterType hunting_, const std::string& part_name) {
+    Targetting::HunterType hunting(hunting_);
+    if ( Targetting::HunterType::Unknown == hunting_ ) {
```
geoffthemedio: remove spaces between `(` and `)' and contained expression.

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+            hunting = Targetting::HunterType::Ship;
+        }
+    }
+    switch ( hunting ) {
+    case Targetting::HunterType::Boat:
+            return 2;
+    case Targetting::HunterType::Ship:
+            if ( (part_name == "SR_WEAPON_0_1" ) || ( part_name =="SR_WEAPON_1_1") ) {
+                    const PartType* part = GetPartType(part_name);
+                    if (part) {
```
geoffthemedio: unnecessary `{ }`

combat/CombatTargetting.cpp
```diff
@@ -0,0 +1,150 @@
...
+    case Targetting::HunterType::Unknown:
+            return 1;
+    default:
+            ErrorLogger() << "Got a lazy hunter weapon: " << part_name;
+            return 1;
+    }
+
+}
+
+Targetting::TriggerConditions Targetting::combine(const Targetting::TriggerConditions& condition, const Targetting::TriggerConditions& another_condition) {
```
geoffthemedio: break params over multiple lines, put `{` on its own line

combat/CombatTargetting.h
```diff
@@ -0,0 +1,69 @@
...
+//       Interceptors (FT_HANGAR_1) - mostly hunting space boats
+//    Attacking planets should be reconsidered (E.g. let ships prefer ships with precision 2); also glass cannon
+//       
+//    Content Setup Language
+//       Precion          'precision'     value in FOCS script for ship_parts
+//       TriggerCondition 'preferredPrey' condition in FOCS script for ship_parts
+//    The content may actually not be hardcoded, so this needs to change before merge.
+
+namespace Targetting {
+    enum PreyType { NoPreference = 7, AllPrey = 7, PlanetPrey = 1, ShipPrey = 2, BoatPrey = 4 };
```
geoffthemedio: weird indenting

combat/CombatTargetting.h
```diff
@@ -0,0 +1,69 @@
...
+          weights(weights_)
+      {}
+
+      TriggerConditions(const Targetting::TriggerCondition& condition_, const int weight_) :
+            conditions({condition_}),
+            weights({weight_})
+      {}
+    };
+
+    bool isPreferredTarget(TriggerCondition condition, Target target);
```
geoffthemedio: Capitalize first letter of FunctionNames

combat/CombatTargetting.h
```diff
@@ -0,0 +1,69 @@
...
+//       
+//    Content Setup Language
+//       Precion          'precision'     value in FOCS script for ship_parts
+//       TriggerCondition 'preferredPrey' condition in FOCS script for ship_parts
+//    The content may actually not be hardcoded, so this needs to change before merge.
+
+namespace Targetting {
+    enum PreyType { NoPreference = 7, AllPrey = 7, PlanetPrey = 1, ShipPrey = 2, BoatPrey = 4 };
+    enum class HunterType { Unknown, Planet, Boat, Ship }; 
+    typedef PreyType                           Prey;
```
geoffthemedio: unnecessary typedefs...?

universe/ShipDesign.h
```diff
@@ -90,7 +91,12 @@ class FO_COMMON_API PartType {
     PartType(ShipPartClass part_class, double capacity, double stat2,
              CommonParams& common_params, const MoreCommonParams& more_common_params,
              std::vector<ShipSlotType> mountable_slot_types,
-             const std::string& icon, bool add_standard_capacity_effect = true);
+             const std::string& icon, bool add_standard_capacity_effect = true,
+             int precision = 2
+             ,
```
geoffthemedio: ?

universe/ShipDesign.h
```diff
@@ -107,6 +113,10 @@ class FO_COMMON_API PartType {
     const std::vector<ShipSlotType>&
                             MountableSlotTypes() const  { return m_mountable_slot_types; }
 
+    int                     Precision() const            { return m_precision; }
+//    const Condition::ConditionBase* PreferredPrey() const{ return m_preferred_prey.get(); }          ///< returns the condition XXXX
+    const Targetting::TriggerCondition PreferredPrey() const{ return m_preferred_prey; }          ///< returns the condition XXXX
```
geoffthemedio: space after `const` before `{` unless aligning to other text

util/Random.cpp
```diff
@@ -40,6 +40,11 @@ GaussianDistType GaussianDist(double mean, double sigma) {
     return GaussianDistType(gen, boost::normal_distribution<>(mean, sigma));
 }
 
+DiscreteDistType RandWeightedInt(const std::list<int> & weights) {
```
geoffthemedio: no space between `>` and `&`

combat/CombatSystem.cpp
agrrr3: Well that was already the old code. And playtesting it worked if there was only one enemy. So I'd say its OK ;)

combat/CombatSystem.cpp
agrrr3: i add something for the final round if that is the block you mean

combat/CombatTargetting.cpp
agrrr3: i need it for the conditions.weights[i] - maybe i get round to refactor that structure, but dunno

combat/CombatTargetting.cpp
agrrr3: If you dont have preferences, any target is good to shoot at.

combat/CombatSystem.cpp
agrrr3: That constructor wasn't necessary actually. So i removed it.

agrrr3:
Reformatting is done. I will have a stab at refactoring probably at the weekend.

Dilvish-fo:
Since we've been having so many spurious CI fails lately, I'll explicitly note that this one appears to me to be caused by an actual error.

agrrr3:
@Dilvish-fo The failures started with e0b52abc267a71054ff8c7252184c9fff549f709, which is mostly reformatting. I guess a change in the build system is responsible? Locally on my fedora and debian systems it builds just fine.
The only non-formatting change seems to be the addition of
> if (!part_name.empty())
>     ErrorLogger() << "ShipWeaponsStrengths couldn't get part " << part_name;

combat/CombatSystem.cpp
```diff
@@ -1464,12 +1531,13 @@ namespace {
             }
 
         } else if (attack_planet) {     // treat planet defenses as direct fire weapon
+            Targetting::TriggerConditions preferred_targets(Targetting::PreyAsTriggerCondition(Targetting::ShipPrey), 3);
             weapons.push_back(PartAttackInfo(PC_DIRECT_WEAPON, UserStringNop("DEF_DEFENSE"),
-                                             attack_planet->CurrentMeterValue(METER_DEFENSE)));
+                                             attack_planet->CurrentMeterValue(METER_DEFENSE),3,preferred_targets));
```
geoffthemedio: spaces after `,`

Dilvish-fo:
> @Dilvish-fo The failures started with e0b52ab, which is mostly reformatting.

@agrrr3  No, you're misunderstanding how/when the CI builds get done & reported-- when several commits are pushed at the same time (or close enough together that a new CI build had not yet started), then the CI build is done with respect to everything up to the point of the last commit.  

In this PR, the errors start appearing with very first the CI build, reported for 54b37fb (which is a trivially small commit unrelated to the actual error), and there is not a successful CI build in this PR after that point. So we know from that that the error starts someplace in the initial bunch of commits, but we don't know which one, though the actual error(s) will generally give some hints at least.

> Locally on my fedora and debian systems it builds just fine.

This PR does not build for me on Ubuntu, using gcc version 5.4.0 and boost 1.58; I appear to be getting the same error being reported by the CI system.  

```
parse/ShipPartsParser.cpp:113:17:   required from here
/usr/include/boost/fusion/container/vector/convert.hpp:27:13: error: invalid use of incomplete type ‘struct boost::fusion::detail::barrier::as_vector<11>’
             type;
             ^
```
The reported line 113 is the start of a parsing structure that spans a number of lines, I think the actual error is probably caused by the added code a few lines below 113

agrrr3:
After introducing a new function in CombatTargetting.(cpp|h) I suddenly have a linker error which I do not understand:

[ 61%] Linking CXX executable freeorion
CMakeFiles/freeorion.dir/combat/CombatSystem.cpp.o: In function `(anonymous namespace)::ShipWeaponsStrengths(std::shared_ptr<Ship const>)':
CombatSystem.cpp:(.text+0x7223): undefined reference to `Targetting::WrapCondition(Condition::ConditionBase const*&&)

agrrr3:
After introducing a new function in CombatTargetting.(cpp|h) I suddenly have a linker error which I do not understand:

[ 61%] Linking CXX executable freeorion
CMakeFiles/freeorion.dir/combat/CombatSystem.cpp.o: In function `(anonymous namespace)::ShipWeaponsStrengths(std::shared_ptr<Ship const>)':
CombatSystem.cpp:(.text+0x7223): undefined reference to `Targetting::WrapCondition(Condition::ConditionBase const*&&)

o01eg:
`*&&` looks very strange. Are you sure there should be r-value (movable) reference to pointer to `Condition::ConditionBase`?

agrrr3:
@o01eg i get the same problem if i use Condition::ConditionBase* , and well the compile has no problem... only the linker

agrrr3:
I made sure now first that the CI runs successful. Then i isolated that adding the parsing of the preferred prey condition using the existing conditon_parser works for me but breaks the CI. Here is the culprit:
`> -(label(tok.PreferredPrey_) > condition_parser)`

So CI without it ( fb5bfc6 ) is green, with it ( ed938a7 ) is red. And the template monsters lurk in the error reports.
I do not have a clue why this happens so I need the help of the wise freeorion developers :)
How can I proceed?

geoffthemedio:
Maybe it's just too complicated of a parser, and needs to be broken into smaller bits?

agrrr3:
In the meantime I removed as a test the parsing of weapon/damage, and then the compile went fine in travis. Now trying to increase PHOENIX_LIMIT a bit further, but i guess it wont work. ~~Edit: that actually worked~~, Edit2: no it did not

i guess that the #define PHOENIX_LIMIT in my local build works and in the CI doesnt. But that PHOENIX_LIMIT is in there for at least boost 1.53.
@Dilvish-fo  Does the current version compile on your ubuntu?

besides that i think it makes sense to split the weapon/capacity/damage thing into a small subparser, where [i need some help](https://www.freeorion.org/forum/viewtopic.php?f=9&t=11087&p=93734#p93734)

edit2: had originally forgotten to decomment the preferredPrey condition parsing, so first i thought uppnig PHOENIX_LIMIT did work. But it seems it has not.

Dilvish-fo:
> Dilvish Does the current version compile on your ubuntu?

I am in the middle of moving, and just working off my Windows laptop, probably for at least a few more days.

parse/ShipPartsParser.cpp
```diff
@@ -1,15 +1,10 @@
+#define FUSION_MAX_VECTOR_SIZE 20
```
geoffthemedio, line 1: It may be necessary to build Boost with this define set differently, not just define it in a single file. Which would mean you can't change it, as most people won't have built boost with a different value set.

agrrr3: So in short i think the define is not for building boost but for configuring it. (But i my be wrong)

geoffthemedio: I didn't look into that define specifically, but at least some other similar macros worked that way.

agrrr3: As far as i understand after reading up on building with boost, the used stuff is mostly headers. Especially it seems fusion and spirit are headers-only and don't require a build. Can you remember an example for the define you mentioned? Maybe serialization?

agrrr3: Oh, CI went fine with the manual fusion_max_vector_size :)

agrrr3:
I concentrated on cleanup and making the CI work. Now i am basically back at fixing memory holes with the more exact configuration. With -fsanitize=address -fno-omit-frame-pointer flags i get quite a good impression of where the things go downhill, but actually i am not sure this is worth the effort.
Would be going back to using shared_ptr everywhere (which works but is maybe suboptimal performance wise) acceptable? The basic problem is that i have dynamic memory access because of the virtual conditions and there is not a clear ownership of the conditions; for the ship parts the ownership should be in those, but for the fighters and planets there is no good place for ownership.

Dilvish-fo:
> ...would be going back to using shared_ptr everywhere...

Your use of "back" here leaves me wondering, are there places where you changed preexisting uses of shared_ptr to something else, or are you referring to the course of changes within this PR?  (At least for me, this PR is large enough I haven't been able to track/remember the ongoing development closely.)  I do see some use of unique_ptr in Targetting::PreyAsTriggerCondition, for example, is that the kind of thing you are talking about?

>  ...for the ship parts the ownership should be in those, but...

Hmm, are you thinking that each ship part would have its own instance of the preferred prey condition?  Aren't there going to be a great many identical conditions instances that way?  That sounds to me like it could get rather wasteful, it would be like if each planet and ship owned its own instance of all the species and tech Conditions and Effects that applied to it rather than just looking those up from a central location when needed.  Perhaps have a TargettingManager that could own a single copy of the various Conditions that could be needed, and could look up and provide shared_ptrs (or probably, rather, raw pointers in that case, since the Targetting Mgr would be expected to outlive all client uses of such pointers) to those as needed?  Perhaps I haven't really looked closely enough at how you're using all these, though, so take my suggestion with a big grain of salt.

agrrr3:
> > ...would be going back to using shared_ptr everywhere...
> 
> Your use of "back" here leaves me wondering, are there places where you changed preexisting uses of shared_ptr to something else, or are you referring to the course of changes within this PR?

Only inside of this PR. I had a working version which used shared_ptr before (which is a bit wasteful, but probably not much) and tried to make the implementation more "correct" by using unique_ptr etc to express real ownership. (And actually i am not fighting memory holes ATM but heap-use-after-free issues, so accessing already deleted conditions)

I could probably could go back to shared_ptr and do a followup PR when/if i figure out the more correct pointer types.

> 
> > ...for the ship parts the ownership should be in those, but...
> 
> Hmm, are you thinking that each ship part would have its own instance of the preferred prey condition? Aren't there going to be a great many identical conditions instances that way?

No and Yes. So first the No: there is one instance of ship part ("shared" by all ships using that ship part), so one condition per ship part, but not one condition per ship. 
So the less relevant Yes: There is probably a chance to share conditions (especially the primitive ones) like you suggested.

> Perhaps have a TargettingManager that could own a single copy of the various Conditions that could be needed, and could look up and provide shared_ptrs (or probably, rather, raw pointers in that case, since the Targetting Mgr would be expected to outlive all client uses of such pointers) to those as needed?

That would probably solve at least part of my ownership problem and would save quite some resources. The Conditions should start living when the focs files are parsed and die when the game ends.
Is there an example where i would create/add the manager? and how I would feed the manager in the parser?

I probably will try the manager route for condition ownership and see if it also solves my heap-use-after-free issues.

combat/CombatSystem.cpp
```diff
@@ -744,13 +784,15 @@ namespace {
         if (available_fighters > 0 && !part_fighter_launch_capacities.empty()) {
             for (auto& launch : part_fighter_launch_capacities) {
                 int to_launch = std::min(launch.second, available_fighters);
-
+                std::string part_name = launch.first;
```
geoffthemedio: probably `const std::string&`

parse/Tokens.h
```diff
@@ -310,7 +310,9 @@
     (Population)                                \
     (PopulationCenter)                          \
     (Position)                                  \
+    (Precision)                                 \
     (PreferredFocus)                            \
+    (PreferredPrey)                             \
```
geoffthemedio: Rename to `PriorityTargets` or similar. "Prey" is not really applicable.

universe/ShipDesign.cpp
```diff
@@ -305,7 +306,9 @@ PartType::PartType() :
 PartType::PartType(ShipPartClass part_class, double capacity, double stat2,
                    CommonParams& common_params, const MoreCommonParams& more_common_params,
                    std::vector<ShipSlotType> mountable_slot_types,
-                   const std::string& icon, bool add_standard_capacity_effect) :
+                   const std::string& icon, bool add_standard_capacity_effect,
+                   int precision,
+                   std::unique_ptr<const ::Condition::ConditionBase>&& preferredPrey) :
```
geoffthemedio: name parameters `with_underscores` not `CamelCase`

universe/ShipDesign.h
```diff
@@ -90,7 +94,10 @@ class FO_COMMON_API PartType {
     PartType(ShipPartClass part_class, double capacity, double stat2,
              CommonParams& common_params, const MoreCommonParams& more_common_params,
              std::vector<ShipSlotType> mountable_slot_types,
-             const std::string& icon, bool add_standard_capacity_effect = true);
+             const std::string& icon, bool add_standard_capacity_effect = true,
+             int precision = 2,
```
geoffthemedio, line 99: Is a precision value necessary? There are random pick and priority sort conditions that would allow more flexible control than whatever hardcoded precision mechanisms you implement...

combat/CombatSystem.cpp
```diff
@@ -267,28 +269,40 @@ void CombatInfo::ForceAtLeastBasicVisibility(int attacker_id, int target_id)
 ////////////////////////////////////////////////
 namespace {
     struct PartAttackInfo {
+        typedef Targetting::TriggerConditions PartTriggerConditions;
```
geoffthemedio: seems unnecessary

Dilvish-fo:
> Is there an example where i would create/add the manager? and how I would feed the manager in the parser?

Techs, Species and ShipDesigns all use a corresponding Manager, not identical to what you want but similar enough to be a pretty helpful starting point I would think, including for how the parser feeds them info.

agrrr3:
@geoffthemedio i need a place to own/manage the conditions for the non ship parts (planets, combined conditions) else we will get memory leaks. The currently logical place for that is the PartAttackInfo. So in own case it should take ownership, and in the other case (ship parts) it should not. So having a "global" manager is actually a good solution.

The space savings would not be only space savings, but also save a lot creating/copying/deleting condition objects. Not sure if that is important in any way, but it is a nice-to-have on the side.

universe/ShipDesign.h
agrrr3: You mean picking and priority via FOCS? I do not think its a good idea to put the targetting algorithm distributed into FOCS parts.

The precision is there to supply weights for e.g. priority sort. I dont like the "hardcoded" fighting mechanism we have, I would prefer that it would be scripted in python and pluggable.

agrrr3:
I made the combination so you can have multiple sources of targetting, to
allow for e.g. species hatred besides regular targetting and to specify
that e.g. the spinal antimatter is really bad suited to shoot at small
targets, even if you add a specific battle scanner.

Geoff <notifications@github.com> schrieb am Fr., 12. Okt. 2018, 17:27:

> I'm not clear why you're combining conditions, actually. Why doesn't each
> part do its targeting independently according to its targeting condition?

agrrr3:
> I made the combination so you can have multiple sources of targetting, to allow for e.g. species hatred besides regular targetting and to specify that e.g. the spinal antimatter is really bad suited to shoot at small targets, even if you add a specific battle scanner.

Actually one thing to note: i am not combining the *Conditions; i am combining the targetting conditions. That makes it moot to have a manager for the combined conditions.

> > Geoff <notifications@github.com> schrieb am Fr., 12. Okt. 2018, 17:27:
> > […](#)
> > I'm not clear why you're combining conditions, actually. Why doesn't each part do its targeting independently according to its targeting condition?

I simplified the memory model now by removing ownership of Conditions from the PartAttackInfo. So ship parts own their conditions and for the very special case of a planet attacking, there is in CombatTargetting.cpp a const local which takes care of the lifetime of that specific Condition. Manager is not necessary for this use case any more. Implementation is fine IMHO.

universe/ShipDesign.h
agrrr3: I am not sure I understand your point regarding FOCS. I would have used weighted conditions if such existed. I had a lot of trouble with the parser already, so that was a stretch goal. 

I wouldnt mind if somebody moved weights to FOCS conditions. 
I have the feeling that I would mind if the exact algorithm is specified in ship parts. I think the battle system should be pluggable, so the interpretation of the values could be different from one battle system to the next. The interpretation should not happen in the contents like the parts. 

Precision is used as in a double role at the moment. Higher precision gives a weight for target selection if there a multiple conditions and also increases the chance that a condition actually locks a target. I expect precision to express a level of actually getting a lock on a target (and e.g. a good tactical engine could contribute to this) and maybe preferences weights move to its own specification. If you have a better suggestion, i will refactor the naming of course (its the best i could come up with). 

I think the design is reasonable enough even with its debatable choices to be included in master (at least i hope so). 

For design changes/expansion we should move this discussion to the forum i think. For example i think it would be a good idea to separate precision from condition weights. But that is definitly out of scope for my implementation. I think it is time to clean up the code and put the design into the wild and then improve.

parse/Tokens.h
agrrr3: ok, that was from a prototype before. i was looking for nomenclature from hunting to differentiate between things you would like to shoot, a thing which is already in your scope depending if you want to shoot it or not. So i tried to say the thing you want to shoot is prey. And things in your scope are targets. PriorityTargets works for me.

agrrr3: (in the prototype you could query the targets PreyType and compare with the PreferredPreyCondition, but since i switched to the flexible FOCS conditions it was not necessary anymore to assign a PreyType)

agrrr3: There is a small part in the code where the prey type stuff is applicable - the part where the planet defenses construct the condition to prefer shooting at ships. I'll change prey to priorityTargets everywhere else.

combat/CombatSystem.cpp
agrrr3: actually that local is unused, will remove

agrrr3:
I un-uncommented a line of code in c8479c9 which breaks the build. 
I really do not understand what is happening with the CheckSumCombine, it seems clang and gcc are unable to find the function `void CheckSumCombine(unsigned int&, const HullType::Slot&)` which is declared in `ShipDesign.h` and implemented in `ShipDesign.cpp`. Playing around with namespace&access and the place of declaration/implementation did not help. It seems HullType type is not found from included `CheckSum.h`.

MSVC does not seem to have a problem.

This started probably shortly before be9f360d8682e4abf07edf78c2ca7638f0acde3b but i never could find the reason for this to occur

universe/ShipDesign.h
geoffthemedio: The combat system can be "pluggable", but the way a part picks its targets is part of the design of that part, and not something that should take a few vaguely-defined values from the part and interpret in different ways.

I would drop the whole precision concept for this iteration. The point should just be to have a way to script what targets a part can attack, and apply that condition if present, to pick a part's allowed targets.

For the next iteration, you don't really need "weighted conditions". If species and part type both need to have input on target selection, then they can do so in a fixed order... eg. If there is a part targeting condition, filter the potential targets with that condition. Then, if there is a species targeting condition, apply that. For the latter, assuming a primary application would be prioritizing by species of the target, a sorting condition can be used, which checks for the species of the target and selects those targets first.

agrrr3:
Can anybody knowledgable about CheckSumCombine please review my last commit c6a21ce ?

geoffthemedio:
c6a21ce is a klunky workaround that shouldn't be necessary. It was a hassle to get all the CheckSumCombine functions to be recognized, though, I recall...

My initial thought was that it was an API exporting issue, but the declaration is
`FO_COMMON_API void CheckSumCombine(unsigned int& sum, const HullType::Slot& slot);`
so that should be fine.

Possibly it's an issue of the order of includes...?

agrrr3:
> Possibly it's an issue of the order of includes...?

Maybe. I do not grep the mechanics which could lead to this failure and i spent quite some time fiddling in the dark. So I rather would gofor a the working workaround of defining the CheckSum in a leaf type than for the complexity of template resolution/namespace declaration/API export/inclusion order. If nobody comes up with an explanation I would rather have your help making the workaround less klunky (e.g. now the CheckSumCombine is probably redundant with the CheckSum function and also the CheckSum implementation is probably not best practice).

agrrr3:
@geoffthemedio should i add the possibleTargets here or should i drop the PR?

edit: priorityTargets -> possibleTargets

or should we finish and merge this and add possibleTargets in another PR?

agrrr3:
Hm, i did some rebasing (updating to master and rearranging commits). And now the clang build says "/Users/travis/.travis/job_stages: line 373: shell_session_update: command not found"

Is that a unrelated bleep?

geoffthemedio:
> Is that a unrelated bleep?

Haven't looked through your changes but it doesn't seem like that sort of error would be relevant to most changes you'd likely have made...

geoffthemedio:
Superceded by #2384

geoffthemedio:
Superceded by #2384

PR #2251, opened 2018-08-19
Make ORBITAL_DRYDOCK directly enqueable

agrrr3:
This PR makes orbital drydock directly enqueable if a basic shipyard is enqueued. This spares one from revisiting the planet to build the drydock (which is in some playstyles at least quite common).

[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=2&t=11043&p=93364#p93364)

The standard time to build the orbital drydock is increased by the four turns a basic shipyard needs to be constructed.

All the hulls which require the drydock already require the basic shipyard, so no changes are necessary with those.

Note the advanced engineering bay, geo-integration, and nanorobotic processing still need the basic shipyard to be finished to show up in the production.

It would be better to make the buildtime dependent on the fact if there is an already finished shipyard  on the planet but my FOCS-foo wasnt good enough for this.


Dilvish-fo:
Though the CI error has nothing to do with the contents of this PR, I'll mention how strange it seems-- the problem was that it could find most, but not quite all, of the required Boost libraries; the missing one was Boost python.

As for the time adjustment, I'll try to work that out a bit later when I have more time.  Right now I am just on my laptop and need to get FO rebuilt, which is quite a bit more nuisance in Windows than it is in Linux, because of ccache and the lack thereof on Windows.  It's making me start to think about setting this laptop up for dual booting just to make rebuilding FO easier lol.

Dilvish-fo:
``` diff
diff --git a/default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt b/default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
index ab31bee90..5b3b3c922 100644
--- a/default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
+++ b/default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt
@@ -10,7 +10,10 @@ BuildingType
     name = "BLD_SHIPYARD_ORBITAL_DRYDOCK"
     description = "BLD_SHIPYARD_ORBITAL_DRYDOCK_DESC"
     buildcost = 20 * [[BUILDING_COST_MULTIPLIER]]
-    buildtime = 9
+    // If a completed shipyard is present, the Drydocks take 5 turns to build.
+    // This base buildtime of 21 is so that even after building for 4 turns in parallel with a shipyard, once shipyard finishes the
+    // drydocks will still need 5 more turns: (4/nturns < 1/5) ==> (nturns > 20)
+    buildtime = 21 - 16 * Statistic If condition = AND [ Source Contains And [ Building name = "BLD_SHIPYARD_BASE" OwnedBy empire = Source.Owner ] ]
     tags = "ORBITAL"
     location = And [
         Planet
```

In case you look to do more of these, in general if you have a current building with base buildtime ```cur_base_turns``` (5 in this case) and a prereq building with base buildtime ```prereq_base_turns```  (4 in this case) then the clause could be
```
buildtime = (prereq_base_turns * cur_base_turns + 1) - ((prereq_base_turns - 1) * cur_base_turns + 1) * Statistic If...
```

Note that this formula only ensures that two buildings built in parallel like this will still take the same total amount of time.  If you had 3 buildings you tried stringing together like this, like if you had adjusted  the geo-integration shipyard buildtime like this if it was being built in parallel with drydocks, but then you built a shipyard, drydocks and geo-integration yard in parallel, the total completion time would probably be 1 turn less than the sum of the base completion times.

agrrr3:
@Dilvish-fo thanks for figuring that out, i updated the PR

Oberlus:
I can't make this work.
With the above changes to ORBITAL_DRYDOCK.focs.txt I am able to queue the Orb. Dry. right after queueing the basic shipyard, but the remaining turns are 5 (instead of ~~9~~21). If I change the buildtime line by a simple "buildtime = 9" (without the "Statistic If ...") the remaining time is 9, so looks to me the Statistic If isn't working as expected.

geoffthemedio:
21 - 16 = 5 doesn't it? Or what "above changes" are you referring to?

Oberlus:
It gives 5 regardless of the presence of a finished shipyard. My recollection is it should be 21-16*0=21 (ah, yes, not 9... I hadn't understood Dilvish formula) if the shipyard is queued but not finished.

agrrr3:
@Oberlus the intent of the code is that finishing shipyard and drydock together - needs nine turns regardless when you queue the drydock. please see the comment in the code (see 'files changed') makes that understandable.

Oberlus:
@Ophiuchus, I understand it. But that is not my point. The point is #2251 does not seem to work on my machine:

I have applied the changes on default/scripting/buildings/shipyards/ORBITAL_DRYDOCK.focs.txt as of #2251, loaded an ongoing game, put a shipyard and an orbital drydock at the top of the building queue (on a planet that has neither), and the turns to finish were 4 for shipyard and 5 for orbital drydock. That is, I can finish the drydock right one turn after the shipyard.

If that is what is intended and what should be understood from the comments, then all is right.

geoffthemedio:
Try with `LocalCandidate` instead of `Source`

Oberlus:
> Try with LocalCandidate instead of Source

Done (applied to the last pre-RC3 build):
`buildtime = 21 - 16 * Statistic If condition = AND [ Source Contains And [ Building name = "BLD_SHIPYARD_BASE" OwnedBy empire = LocalCandidate.Owner ] ]`
I get the same:
odtime5

geoffthemedio:
I meant the condition `AND [ Source Contains ...`

Also probably `Target` instead of `Source` or `LocalCandidate`.

You want to check the potential production location in the Statistic If sampling condition, which should be target object within the evaluation of a building production time expression, and thus matched by a Target condition, and not the source object, which would be matched by a Source condition. If you test the source object, that's probably the capital planet of the empire, which will usually already have a shipyard. You could check that with the old expression by deleting that building and seeing if it affects the result of enqueuing elsewhere.

https://github.com/freeorion/freeorion/blob/master/universe/Building.cpp#L287

Oberlus:
Ok, that works.

To be specific (@agrrr3):
`buildtime = 21 - 16 * Statistic If condition = AND [ Target Contains And [ Building name = "BLD_SHIPYARD_BASE" OwnedBy empire = Source.Owner ] ]`

With that change (the first `Source` of that line changed by `Target`), building an orbital drydock on a planet with an already built base shipyard takes 5 turns, and queueing the o.d. at the same time than the shipyard makes it require 21 turns which becomes 5 once the shipyard is finished but less than 20% (1/5) of the PPs have been invested so it will take 5 more turns to finish, for a total of 9 turns as intended with Dilvish's formula. Great work.

Dilvish-fo:
@geoffthemedio 
> probably Target instead of Source or LocalCandidate...  If you test the source object, that's probably the capital planet of the empire, which will usually already have a shipyard. 

Ah, right, Target is the right one here, somehow I had misremembered the matchup for Source and Target for these.  My testing was just on the starting homeworld so that didn't reveal the issue.

agrrr3:
Any concerns against merging this?

Vezzra:
While the proposed implementation might work as intended, to me that still looks like a hack, not like an ideal solution. Of course, if only working with what's currently possible with FOCS, and without any changes to the backend C++ code, that's probably the best we can do.

But the right, clean solution for the intended objective would be to extend the capabilities of FOCS to handle such cases correctly.

The objective, if I understand correctly, is simple: if building A depends on building B being present, we want to be able to enqueue building A as soon as building B has been enqueued, and not have to wait for building B to get finished before we can enqueue building A, to avoid tedious and unecessary micromanagement.

Which means we need to be able to specify location conditions in more detailed ways: instead of just being able to specify when an item can be produced, we need two separate conditions: one that specifies if an item can be *enqueued*, and another one that specifies if an enqueued item *actually receives PP*.

Of course, in most cases those two conditions will be the same, so there needs to be a simple way to specify the condition once and use it for both.

With that in place, the issue this PR tries to address can be solved much simpler and cleaner. For the "can enqueue" condition check if the prereq building is already present or at least enqueued. For the "receive PP" condition check if the prereq building is already present.

Then you can enqueue both buildings, and the second one will be produced only once the first has been finished, without the need for complicated formulas for the build time (which is a high maintenance and highly error prune approach).

geoffthemedio:
> we need two separate conditions: one that specifies if an item can be enqueued, and another one that specifies if an enqueued item actually receives PP.

Why? Both to that specific comment, and this whole pull request...

There are already two conditions: a Location condition, and an EnqueueLocation condition. 

The Location condition determines on which planets a building can be allocated PP (ie. can be produced).

The EnqueueLocation condition determines on which planets a building can be enqueued (independent of where it can be allocated PP). If there is no EnqueueLocation condition specified, then the Location condition is used for that test.

However, in BuildDesignatorWnd, it checks both before deciding whether to show something:

https://github.com/freeorion/freeorion/blob/master/UI/BuildDesignatorWnd.cpp#L771

I suspect what should be done is to make the EnqueueLocation for drydocks check if a shipyard is present or enqueued at the production location, and get rid of the test of Location in BuildDesignatorWnd when deciding whether to allow something to be enqueued.

Vezzra:
> There are already two conditions: a Location condition, and an EnqueueLocation condition...

Meaning, what I've suggested actually already exists (cool! 😃). However:

> However, in BuildDesignatorWnd, it checks both before deciding whether to show something

Which obviously is a bug, plain and simple.

Meaning, the right fix for the issue at hand would be to properly fix the BuildDesignatorWnd bug and set up the Location and EnqueueLocation conditions accordingly (as @geoffthemedio suggested).

And all the other cases where we have buildings depending on the presence of other buildings, where we currently need to wait for the prereq building to finish production before we can queue the dependent building can be easily addressed the same way.

So, can we close this PR (as "invalid" or "superseded") and make a new one which implements the proper fix?

geoffthemedio:
> Which obviously is a bug, plain and simple.

I think it was implemented that way intentionally, but can be changed.

agrrr3:
> So, can we close this PR (as "invalid" or "superseded") and make a new one which implements the proper  fix?

Maybe have a look in the forum discussion?

> Ophiuchus wrote: ↑
> Sun Aug 19, 2018 12:24 am
> another more general approach could be to allow to enqueue items for which the location condition fails but the enqueueLocation succeeds; that item could would be added to the queue, but production would start/proceed not before the prerequisites are fulfilled

And an answer from dilvish
> poor alternate ... would tremendously clutter the BuildDesignator (because a fourth of buildings have no EnqueueLocation condition, and of the ones that do, about half are simply [[[ENQUEUE_BUILD_ONE_PER_PLANET]] so 3/4 of buildings would be showing up as enqueueable even if not buildable), unless we wrote new EnqueueLocation conditions for every single building-- the revised EnqueueLocation would have to include the current EnqueueLocation condition plus all of the Location conditions other than any local building requirements. It would be much simpler to just edit the handful of applicable Location conditions (and in some cases probably buildtime also) in the manner suggested for Drydocks.

So if somebody steps up for the cleaner solution, be my guest.

Vezzra:
Just because the mess is apparently bigger than it looked at first, and the clean solution would take considerably more effort than whatever quick stop-gap fix doesn't mean we shouldn't implement the clean solution, IMO.

On the contrary, just patching things up without fixing the underlying problem will only compound the entire mess, and make it even more difficult to clean it up later. So, my proposal is still to try the clean solution. We're just after a stable release, at the beginning of a new release cycle, what better time to attempt major fixes... 😉

agrrr3:
> Just because the mess is apparently bigger than it looked at first, and the clean solution would take considerably more effort than whatever quick stop-gap fix doesn't mean we shouldn't implement the clean solution, IMO.

@vezzra Of course we should implement the clean solution no matter how much work it is. And by 'we' I mean the same as you do: 'somebody else'. But your prompted dilvish to actually come up with another solution. So thanks for that :)

@Dilvish-fo Thank you, looks great. If it works i will have a look for (some of) the other buildings as well.

Vezzra:
> Of course we should implement the clean solution no matter how much work it is. And by 'we' I mean the same as you do: 'somebody else'.

Yeah, I know. I'll have to concede that... :wink:

agrrr3:
Hm the other shipyards have some pecularities - like being able to be build if there is a shipyard OR a orbital drydock - should they be streamlined away?
Im not sure why they are there. So you can kill of the basic shipyard?

agrrr3:
Also the change to xenocoordination  exemplifies my "problems" - with the change i did it is back to old school - the orbital incubator has to be finished before start on the xeno starts. But maybe the old new way was better - so i should only use the macro on the basic shipyard and leave the part for the orbital incubator like it was.

agrrr3:
Besides the prescheduling for the medium level shipyards I kept the old behaviour mostly intact, meaning the advanced shipyards can be build in parallel. I think this helps the production predictability (because items enqueued but not buildable are not part of the production plan).

Vezzra:
@Dilvish-fo:

> @Vezzra You seem to still be thinking that the current implementation is bugged; please note Geoff's response to that above ...

I did note that response, I just understood it more like meaning that the check in `BuildDesignatorWnd` isn't just some typo or bugged implementation, but that whoever implemented this made it the way it is now for a reason. Which means, if we decide to change the implementation, we need to determine what that reason was to not break something else.

However, that doesn't mean the implementation isn't bugged, whoever wrote that code could have overlooked the side effect implementing it this way has, or had a different idea of what the `Location` and `EnqueueLocation` conditions are supposed to be/do.

Anyway, my assessment of the current implementation being bugged has been made specifically in the context of [the explanation @geoffthemedio gave](https://github.com/freeorion/freeorion/pull/2251#issuecomment-418324059) in response to [my proposal for a better/cleaner solution](https://github.com/freeorion/freeorion/pull/2251#issuecomment-418309987) of the issue at hand.

He basically stated that what I suggested already exists, that the `Location` and `EnqueueLocation` conditions are supposed to do what I want, the only reason why the cleaner solution I proposed currently won't work is because of the way the check in `BuildDesignatorWnd` is implemented.

So, _under the premise that the `Location` and `EnqueueLocation` conditions are supposed to work the way @geoffthemedio described_, the current implementation of the check in `BuildDesignatorWnd` is bugged. Because it prevents the `Location` and `EnqueueLocation` conditions from being used as intended.

Furthermore, @geoffthemedio himself suggested to change this implementation:

> get rid of the test of Location in BuildDesignatorWnd when deciding whether to allow something to be enqueued

All I said above is moot of course if the `Location` and `EnqueueLocation` conditions are _not_ supposed to be/do exactly what I suggested (one condition strictly for determining if you can _enqueue_ at a location, the other strictly for _if you can spend PP_ on a certain item at a  location).

In that case I'd argue that we proably should redefine those conditions to do exactly that, because I think this strict separation is the best approach. Which would be subject to discussion of course, my assumption that this is the best approach might be flawed after all.

Personally I'd suggest a adjustment of these conditions anyway: I suggest making the condition that determines if an item can be enqueued at a loation the standard/required condition, and the condition determining if PP can be spent on an item an _optional extension_ of the main one.

Meaning, an item can only receive PP if the enqueue _and_ an optional extra condition ("Production" condition) is met. That removes the necessity to repeat the entire `Location` condition in the `EnqeueLocation` condition if those are not the same like it has to be done now (IIRC).

After all, I can't think of a case where you wouldn't have to meet all conditions required to enqueue an item in order for it to receive PP. If anything, the condition determining if an item can receive PP will always be a superset, not a subset of the enqueue condition, or do I miss something obvious here?

Dilvish-fo:
> After all, I can't think of a case where you wouldn't have to meet all conditions required to enqueue an item in order for it to receive PP. If anything, the condition determining if an item can receive PP will always be a superset, not a subset of the enqueue condition, or do I miss something obvious here?

You seem to me to be forgetting that the whole motivation for adding the EnqueueLocation to begin with was to provide a way to limit items from showing up on the BuildDesignatorWnd ***without*** preventing them from receiving PP once they were already on the queue, as as with the ubiquitous ```[[ENQUEUE_BUILD_ONE_PER_PLANET]]```

 EnqueueLocation is really just a UI-helper-filter thing, as opposed to Location which is a substantive game mechanic thing.   I just submitted a PR (noted just above) to clarify EnqueueLocation and remove the bit of wrong code from it.

default/scripting/common/enqueue.macros
```diff
@@ -1,3 +1,33 @@
...
+            And [
+                Enqueued type = Building name = "@1@"
+                Not Enqueued type = Building name = CurrentContent
+            ]
+        ]'''
+/* Allows the current building to be build if the given prerequisite is built or enqueued.
+   Takes the prerequisite name as parameter; Usage:
+        [[LOCATION_ALLOW_BUILD_IF_ENQUEUED(BLD_SHIPYARD_ORBITAL_DRYDOCK)]]
+*/
+LOCATION_ALLOW_BUILD_IF_ENQUEUED
```
Dilvish-fo: I don't think we should really have/use this macro/requirement -- even though this may correspond to how some things had previously worked, that it was fully sufficient if the prereq were simply enqueued whether or not it were ever finished, I think that was just a bit sloppy, and we allowed it only because we hadn't yet figured out a better way to get the flexibility that we actually wanted.  IMO prereqs should have to be completed before the dependent structures can make progress (i.e., use the other new macro), or else they just shouldn't be prereqs.   (Perhaps that's what you asked about before and I just hadn't read closely enough.)

default/scripting/common/enqueue.macros
```diff
@@ -1,3 +1,33 @@
+/* Allows the current building to be enqueued if the given prerequisite is built or enqueued.
+   Takes the prerequisite name as parameter; Usage:
+        [[LOCATION_ALLOW_ENQUEUE_IF_ENQUEUED(BLD_SHIPYARD_BASE)]]
+*/
+LOCATION_ALLOW_ENQUEUE_IF_ENQUEUED
```
Dilvish-fo: I think making this macro name slightly longer would make it much more clear-- 
```
LOCATION_ALLOW_ENQUEUE_IF_PREREQ_ENQUEUED
```

Dilvish-fo:
> Ok, agreed, but in that case the Location and EnqueueLocation conditions do not really match what I proposed above, or do they?  What I want is a simple way to script production items that can be queued, but won't receive any PP until certain (additional) conditions are met.

@Vezzra I believe that this PR now shows that what you wanted can be easily accomplished via the Location condition.  I think that the ```LOCATION_ALLOW_ENQUEUE_IF_ENQUEUED```  / ```LOCATION_ALLOW_ENQUEUE_IF_PREREQ_ENQUEUED``` macro now in this PR demonstrates an easy template that could be applied to other build constraints as well.

default/scripting/common/enqueue.macros
agrrr3: I know that the way it worked was we tried to achieve what we actually achieve with the changes in this thread. But there may be a downside to using ALLOW_ENQUEUE_IF_ENQUEUED everywhere which we should check first: is the location condition correctly reevaluated when calculating future production?

default/scripting/common/enqueue.macros
agrrr3: sounds good to me

agrrr3:
@vezzra @Dilvish-fo  Hm. If you guys want to continue your discussion about the way location/enqueueLocation should/could work, you should rather do that on a thread in the game design forum, right?

default/scripting/common/enqueue.macros
Dilvish-fo: Something enqueued in reliance on its prereq being enqueued, with ALLOW_ENQUEUE_IF_ENQUEUED in its location condition, will be listed in the production queue as Never completing, until such point as the prereq is completed and then it will start being able to receive PP and will start showing a predicted completion time.  The prediction code is not going to be able to simulate the future completion of prereqs (at least, not without a significant restructuring/recoding of how building prereqs are specified and a very significant recoding of the production simulator).

I don't think that the ability to predict completion time before a prereq is completed is a sufficient reason to weaken the prereq condition down to only requiring that the prereq be enqueued.

geoffthemedio: > The prediction code is not going to be able to simulate the future completion of prereqs

To elaborate, the location conditions test the gamestate, including objects in the universe and some other non-object-associated values. Simulating the production queue doesn't create any new objects in the universe, so a condition testing if an object exists can't change part way through the simulation.

agrrr3: > I don't think that the ability to predict completion time before a prereq is completed is a sufficient reason to weaken the prereq condition down to only requiring that the prereq be enqueued.

I do not think this weakens the prereq in this case... we are talking shipyards and the real prerequisites are not the prerequisites of the yards but of the ability to build the ships. As the ships depend on all prerequisites you cant build a ship with one prereq missing.

I think the only possible down sides of being able to build the prerequisites at the same time is that building in sequence takes longer (which i think is ok) and that the user has to understand which yards are necessary to build a ship (which i think is sufficiently understandable from the descriptions and from non enqueable condition info).
So i think these have to be balanced against the benefit of a more correct build queue prediction.

Vezzra: > I think the only possible down sides of being able to build the prerequisites at the same time is that building in sequence takes longer (which i think is ok)

That issue might be more serious than you think. The minimum build time is as important a factor to correctly balance a game element as the build costs are. It makes sure that producing a certain item takes a certain minimum amount of turns, and that you can't circument this by throwing more PP at it.

Meaning, when it comes to shipyards, those minimum build times ensure that it takes you a certain time to build up a fully upgraded shipyard with all extensions, and that you cannot do this too quickly just by investing more PP.

The moment you allow the various shipyard buildings to be built in parallel, you effectively break that premise, making balancing the shipyard extension buildings more difficult.

Sure, we can just increase the minimum build time of the shipyard extenion buildings to account for the ability to build in parallel, but that will totally screw the case where you just want to add an extension to an already existing shipyard, as the build times would be much too long in case you do *not* build in parallel.

So I think this comes down to how important we think it is to maintain that it the time it takes you to build a fully upgraded shipyard must not be cut short, but stay at a certain level (to ensure proper game balance).

Vezzra:
> I believe that this PR now shows that what you wanted can be easily accomplished via the Location condition. I think that the LOCATION_ALLOW_ENQUEUE_IF_ENQUEUED / LOCATION_ALLOW_ENQUEUE_IF_PREREQ_ENQUEUED macro now in this PR demonstrates an easy template that could be applied to other build constraints as well.

I don't have the time currently to analyze how these macros work, so I'm just going to trust your assessment on this. :smiley: No more objections from my side.

default/scripting/common/enqueue.macros
agrrr3: @Vezzra Keeping the parallel build means not changing the balance. Dilvish was suggesting to go back to an older style, which would change balance. 

Back to the general discussion how we want the mechanic to work, I would rephrase what you said: it comes down to how important we think it is to be able to build advanced buildings faster if the prerequisites are already build

I also think that nobody complained about the balance concerning advanced shipyards since the parallel build was introduced, so i take that as a data point indicating that this is not serious.
So, i repeat myself: I think it is OK if we do not enforce sequential build for advanced shipyards.

Vezzra: Sorry if I was a bit unclear: I merely wanted to point out that the issue of being able to build in parallel is a quite significant one in general, I wasn't specifically referring to the current change. The current state of things with the shipyards is a big mess anyway. Hence the lack of complaints about balance you mentioned - there was not much of well thought out balance to mess up to begin with.

For a proper fix/solution however what I said is relevant, IMO.

Dilvish-fo (approved):
OK, well, I think my comment about possibly doing away with the second macro here is really more of a proposal for a change that is related to but independent of the change intended to be accomplished by this PR, and I'm OK with just putting further discussion of that off for later; it seems it shouldn't really further delay this PR.

PR #2254, opened 2018-08-29
Give more correct descriptions of negated And conditions

agrrr3:
The ::description() of FOCS And conditions gave the same output no matter if the condition is negated or not.

This PR handles uses de morgans law for transformation in the Condition.cpp, describing NOT AND as an OR and pushing the negation to the operands. Same goes for NOT OR which gets transformed to AND.

```
Not And [  DesignHasHull name ="SH_COLONY_BASE"     ( PartsInShipDesign design = LocalCandidate.DesignID < 1 ) ]
(either not that contains the hull Colony Base Hull, or not when PartsInShipDesign < 1 )
```

Also NOT NOT nullifies itself now by negation switching.

A previous commit used de morgans law directly on the stringtable level, which allowed to use 'neither'/'nor' in english, but as it prevented negative operands to cancel the negation out i found that to be less clear in the end. 

Also note that one could pull up AND conditions inside of AND conditions and OR conditions inside of OR condition, making the description more clear. But probably in those cases one should fix the condition instead of the description, so its actually good to keep it smelly like it is.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -10040,13 +10040,13 @@ DESC_ORDERED_BOMBARDED_NOT
 ''' that is not being bombarded by an object %1%'''
 
 DESC_AND_BEFORE_OPERANDS
-''''''
+'''['''
 
 DESC_AND_BETWEEN_OPERANDS
 ''' and'''
 
 DESC_AND_AFTER_OPERANDS
-''''''
+''']'''
 
 DESC_AND_BEFORE_SINGLE_OPERAND
 ''''''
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Please give some before and after examples of actual output.

agrrr3:
```Not And [  DesignHasHull name ="SH_COLONY_BASE"     ( PartsInShipDesign design = LocalCandidate.DesignID < 1 ) ]```

without the PR:
`that contains the hull Colony Base Hull and when PartsInShipDesign < 1`

with the PR:
`(not that contains the hull Colony Base Hull or not when PartsInShipDesign < 1 )`

File: default/stringtables/en.txt [hunk 1]
```diff
@@ -10039,12 +10039,42 @@ DESC_ORDERED_BOMBARDED
 DESC_ORDERED_BOMBARDED_NOT
 ''' that is not being bombarded by an object %1%'''
 
+DESC_AND_BEFORE_OPERANDS
+''''''
```
Dilvish-fo: I think it would be most clear, and most consistent between negated and not negated, to render ```A And B``` as ```(A and B)```

so this entry would be 
```'''('''```

File: default/stringtables/en.txt [hunk 2]
```diff
@@ -10039,12 +10039,42 @@ DESC_ORDERED_BOMBARDED
...
+
 DESC_AND_BETWEEN_OPERANDS
 ''' and'''
 
+DESC_AND_AFTER_OPERANDS
```
Dilvish-fo: ```''')'''```

default/stringtables/en.txt
```diff
@@ -10039,12 +10039,42 @@ DESC_ORDERED_BOMBARDED
...
+''''''
+
+DESC_AND_BEFORE_SINGLE_OPERAND
+''''''
+
+DESC_AND_AFTER_SINGLE_OPERAND
+''''''
+
+DESC_NOT_AND_BEFORE_OPERANDS
+''' (not'''
```
Dilvish-fo: '''(not ('''

default/stringtables/en.txt
```diff
@@ -10039,12 +10039,42 @@ DESC_ORDERED_BOMBARDED
...
+
+DESC_NOT_AND_BETWEEN_OPERANDS
+'''[[DESC_OR_NOT_BETWEEN_OPERANDS]]'''
+
+DESC_NOT_AND_AFTER_OPERANDS
+''' )'''
```
Dilvish-fo: '''))'''

default/stringtables/en.txt
```diff
@@ -10039,12 +10039,42 @@ DESC_ORDERED_BOMBARDED
...
+''' not'''
+
+DESC_NOT_AND_AFTER_SINGLE_OPERAND
+''''''
+
 DESC_OR_BETWEEN_OPERANDS
 ''' or'''
 
+DESC_OR_NOT_BETWEEN_OPERANDS
+''' or not'''
```
Dilvish-fo: ''') or not ('''

Dilvish-fo (commented):
I think that this set of changes would accomplish my proposed adjustment

Dilvish-fo:
Hmm, thinking about this a bit more, it is seeming possible to me that, so long as corresponding changes were being made to OR, the only change to the PR that might be needed would be the very first to I propose, so that ```AND [A B]``` renders as ```(A and B)```, also with the corresponding change to OR so that ```OR [A B]``` renders as ```(A or B)```

agrrr3:
How about we always use parenthesis, but only for the OR operation. In boolean algebra the AND conjunction is the multiplication and thus binds stronger than addition, so you would not have to use parenthesis. Also the OR operation actually is mixture between addition and multiplication, so it makes sense to put parenthesis around it anyway.
> OR [ 
>  AND [ Source OR [ Ship Planet ] Species ]
>  Building
> ]

> --> That either (is the source and that either ( is a ship, or is a planet ) and has a species, or is a building)

or lets use different parenthesis for AND and OR
> --> That either ( [is the source and that either ( is a ship, or is a planet ) and has a species], or is a building)

Dilvish-fo:
I am fine with using regular parentheses for OR and using square brackets for AND (and it does seem like it has promise to make things more clear).

Regarding the "either" in your latest examples, I prefer ```"(either A, or B)"``` over ```either (A, or B)```

agrrr3:
I fixed that now accordingly in commit a4b44327b780d084d7721dc9abf3f6e0f38fc9a5

But actually i found its more worthwhile to actually push the negation through to the operands because that gives the possibility for more helpful messages as double negation has a chance to cancel itself out (instead of '(neither an object that is not a ORB_SY...' you get a '(an object that is a ORB_SY and ..)'. Also in english this is really fine.

PR #2261, opened 2018-09-03
[python] Use print function in common modules

Cjkjvfnby:
Add compatibility with python3 for common modules.  These modules can be imported before logging is configured so I used `print_function`

### DIFF
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 """configure_logging redirects print and the logger to use the freeorion server.
 
 Output redirected to the freeorion server prints in the appropriate log:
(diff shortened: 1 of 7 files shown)


Vezzra:
Is that really AI specific? To me it looks like these changes affect the common/general parts of the python scripting...?

Cjkjvfnby:
> Is that really AI specific? To me it looks like these changes affect the common/general parts of the python scripting...?

@Vezzra Should I replace label or use both?

`The Issue/PR deals with the Python AI decision making code or affects it.`
`The Issue/PR deals with the FOCS language, turn events, the universe generator and probably game mechanics.`

Dilvish-fo (commented):
Although you note in your intro that "These modules can be imported before logging is configured so I used print_function", it looks to me like at least one of these files does import logging first, but overall this looks fine to me and seems like a good general step.

Dilvish-fo (approved):

Vezzra:
> Should I replace label or use both?

I'd only use "component:content scripting", as those changes contain nothing that looks AI specific. The AI modules might use these (utility etc.) functions, but they are not specific for the AI.

I'll go ahead and adjust the labels.

PR #2262, opened 2018-09-03
Fix Part Exclusions

### LINKED ISSUE #2260: Ship part exclusions not working for certain part ordering
### Environment
Master as of 9439931

### Description
Consider a part A which excludes part B:
```
exclusions = ["B"]
```

When trying to add a new design, the "Add Finished Design" button is only grayed out if the excluded part B is in an earlier slot than the excluding part A.
This means you can actually add a design which has both part A and part B in it.

### Expected Result
Exclusions are respected regardless of part/slot order. You can not add a design with Part A and B in it if part A excludes part B.

### Steps to reproduce
1) Apply following diff:
```
diff --git a/default/scripting/ship_parts/Colony/CO_COL_OUTPOST.focs.txt b/default/scripting/ship_parts/Colony/CO_COL_OUTPOST.focs.txt
index a3affb1e8..bb16c6eca 100644
--- a/default/scripting/ship_parts/Colony/CO_COL_OUTPOST.focs.txt
+++ b/default/scripting/ship_parts/Colony/CO_COL_OUTPOST.focs.txt
@@ -1,6 +1,10 @@
 Part
     name = "CO_OUTPOST_POD"
     description = "CO_OUTPOST_POD_DESC"
+    exclusions = [
+        "CO_COLONY_POD" 
+        "CO_SUSPEND_ANIM_POD"
+    ]
     class = Colony
     capacity = 0
     mountableSlotTypes = Internal
```
2) Open the ship design screen
3) Select an empty Camouflage Asteroid Hull (or any other hull with at least 3 slots).
4) Add an outpost pod to one of the middle slots.
5) Add a colony pod to its left. You are unable to add the design as expected from the exclusions.
6) Move the colony pod to the right of the outpost pod. You are now able to save the design. This shouldn't happen.

geoffthemedio:
Hopefully resolves https://github.com/freeorion/freeorion/issues/2260

### DIFF
--- a/universe/ShipDesign.cpp
+++ b/universe/ShipDesign.cpp
@@ -1131,7 +1131,7 @@ ShipDesign::MaybeInvalidDesign(const std::string& hull_in,
     auto parts = parts_in;
 
     // ensure hull type exists
-    const HullType* hull_type = GetHullTypeManager().GetHullType(hull);
+    auto hull_type = GetHullTypeManager().GetHullType(hull);
     if (!hull_type) {
         is_valid = false;
         if (produce_log)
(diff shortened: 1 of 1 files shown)


universe/ShipDesign.cpp
```diff
@@ -1185,34 +1185,38 @@ ShipDesign::MaybeInvalidDesign(const std::string& hull_in,
...
-            part_name.clear();
             continue;
         }
 
         for (const auto& excluded : part_type->Exclusions()) {
-            if (already_seen_component_names.count(excluded)) {
+            // confict if a different excluded part is present, or if there are
+            // two or more of a part that excludes itself
+            if ((excluded == part_name && component_name_counts[excluded] > 1) ||
+                component_name_counts[excluded] > 0)
```
Morlic-fo: Second or-condition may override first. Should instead read
```
if(component_name_counts[excluded] > (excluded == part_name ? 1 : 0))
```
or similar...

Morlic-fo (changes requested):
Can't build designs with self-excluding parts with this diff.

Morlic-fo:
Seems to work fine with following additional diff...
```
diff --git a/universe/ShipDesign.cpp b/universe/ShipDesign.cpp
index 60804d08a..55427f00f 100644
--- a/universe/ShipDesign.cpp
+++ b/universe/ShipDesign.cpp
@@ -1210,8 +1210,8 @@ ShipDesign::MaybeInvalidDesign(const std::string& hull_in,
         for (const auto& excluded : part_type->Exclusions()) {
             // confict if a different excluded part is present, or if there are
             // two or more of a part that excludes itself
-            if ((excluded == part_name && component_name_counts[excluded] > 1) ||
-                component_name_counts[excluded] > 0)
+            unsigned int max_count = excluded == part_name ? 1 : 0;
+            if (component_name_counts[excluded] > max_count)
             {
                 is_valid = false;
                 if (produce_log)

```

geoffthemedio:
How about like this?

Morlic-fo (approved):
Seems to work correctly now

PR #2263, opened 2018-09-04
[RULES] Simplify rule description

Cjkjvfnby:
Make the description more simple for users. PRNG is not the thing many users know.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1074,7 +1074,7 @@ RULE_RESEED_PRNG_SERVER
 Random Reseeding
 
 RULE_RESEED_PRNG_SERVER_DESC
-With reseeding, replaying the same save or using the same settings and seed is not expected to produce the same random results. This is possibly preferable if players want to be sure that none of them can predict future random results based on previous random results
+With reseeding, replaying the same save or using the same settings and seed is not expected to produce the same random results. This is possibly preferable if players want to be sure that none of them can predict future random results based on previous random results or if a player wants to 're-roll' the results of a battle or random effect when re-playing a turn, and not get the game results as a previous attempt
 
 RULE_SHIP_SPEED_FACTOR
 Ship Speed Scaling
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1071,10 +1071,10 @@ RULE_NUM_COMBAT_ROUNDS_DESC
 How many rounds of combat occur each turn.
 
 RULE_RESEED_PRNG_SERVER
-More reproducible random generation
+Random Reseeding
 
 RULE_RESEED_PRNG_SERVER_DESC
-Give more predictable and reproducible values.
+With reseeding, replaying the same save or using the same settings and seed is not expected to produce the same random results. This is possibly preferable if players want to be sure that none of them can predict future random results based on previous random results
 
 RULE_SHIP_SPEED_FACTOR
 Ship Speed Scaling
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I prefer it as is...

default/stringtables/en.txt
```diff
@@ -1071,10 +1071,10 @@ RULE_NUM_COMBAT_ROUNDS_DESC
 How many rounds of combat occur each turn.
 
 RULE_RESEED_PRNG_SERVER
-Random Reseeding
+More reproducible random generation
 
 RULE_RESEED_PRNG_SERVER_DESC
-Random number generation is re-seeded frequently to give more predictable and reproducible values.
+Give more predictable and reproducible values.
```
Dilvish-fo: I agree with Geoff in preferring the original here, and see no reason to change it.

default/stringtables/en.txt
```diff
@@ -1071,10 +1071,10 @@ RULE_NUM_COMBAT_ROUNDS_DESC
 How many rounds of combat occur each turn.
 
 RULE_RESEED_PRNG_SERVER
-Random Reseeding
+More reproducible random generation
```
Dilvish-fo: Here the original text does seem unnecessarily prone to generating confusion in those not familiar with operational details of RNGs.   But I think that the proposed "random generation" is similarly a bit too incomplete & ungrammatical.

I would propose the somewhat more explicit
```
Random Number Generator Reseeding
```
or possibly
```
More reproducible random number generation
```
Although I think that for the current purpose the latter may be getting a bit too long.
@geoffthemedio thoughts on these?

geoffthemedio: The name can and should remain short. The description can elaborate... ie. how it is already.

geoffthemedio: Why would we remove it? It's a useful option, depending whether or not players want reproducible random results, eg. https://github.com/freeorion/freeorion/issues/1230

Cjkjvfnby: > depending whether or not players want reproducible random results

And in the description, we have `more predictable and reproducible values`. Since we talk about user interface each word matters. 

And according to the issue title, we always have the reproducible result but random number generator produce alike result for different seeds.

>  It's a useful option  

Why is one random better than the other? 
Can you explain in simple words what will be changed in the gameplay when I switch this option?  Treat me us user, who don't know anything about inner implementation.

geoffthemedio: The description is not nearly as size-constrained as the option name. The description only appears in a tooltip, so doesn't need to fit between other widget labels like the names do.

Without reseeding, the randomness should be reproducible if things are working correctly. That is, reloading the same save or restarting with the same galaxy settings and seed, if all players issue the same orders, then the results should be the same. This might not work in practice due to threading issues, but ideally it would.

With reseeding, replaying the same save or using the same settings and seed is not expected to produce the same random results. The PRNG is occasionally re-seeded from an effectively unpredictable source of randomness (the clock or some other random device). This is possibly preferable if players want to be sure that none of them can predict future random results based on previous random results. However, as noted in the issue, some people don't like how this is implemented for various reasons, and in some cases it's preferable to have reproducible random results.

Cjkjvfnby:
I revert option name and update description from comment.

Dilvish-fo:
Looks to me like Geoff had a typo in his last comment which got copy-pasted into the final commit and then merged:
"and not get the game results " probably should be "and not get the same results "

@geoffthemedio if you confirm that was a typo, then @Cjkjvfnby perhaps you could just push a commit with that correction.

geoffthemedio:
yes, typo

Cjkjvfnby:
Done https://github.com/freeorion/freeorion/commit/6e7521fed9328f97746cf67b072ea53e4d697df9

PR #2264, opened 2018-09-05
[python wrappers] Add user config and data dirs

Cjkjvfnby:
Add the config directory and option access. I need this code to have the way to generate AI stubs for that interfaces like it is done for  [AI](https://github.com/freeorion/freeorion/blob/master/default/python/AI/freeOrionAIInterface.pyi#L1)

I know a little about C++ and don't compile code at all. 

@o01eg  can you take care of this PR?

### DIFF
--- a/python/UniverseWrapper.cpp
+++ b/python/UniverseWrapper.cpp
@@ -18,6 +18,7 @@
 #include "../util/Directories.h"
 #include "../util/Logger.h"
 #include "../util/MultiplayerCommon.h"
+#include "../util/OptionsDB.h"
 #include "../util/GameRules.h"
 
 #include <boost/mpl/vector.hpp>
@@ -241,6 +242,17 @@ namespace {
     bool RuleExistsWithType(const GameRules& rules, const std::string& name, GameRules::Type type)
     { return rules.RuleExists(name, type); }
 
+    boost::python::object GetOptionsDBOptionStr(std::string const &option)
+    { return GetOptionsDB().OptionExists(option) ? boost::python::str(GetOptionsDB().Get<std::string>(option)) : boost::python::str(); }
+
+    boost::python::object GetOptionsDBOptionInt(std::string const &option)
+    { return GetOptionsDB().OptionExists(option) ? boost::python::object(GetOptionsDB().Get<int>(option)) : boost::python::object(); }
+
+    boost::python::object GetOptionsDBOptionBool(std::string const &option)
+    { return GetOptionsDB().OptionExists(option) ? boost::python::object(GetOptionsDB().Get<bool>(option)) : boost::python::object(); }
+
+    boost::python::object GetOptionsDBOptionDouble(std::string const &option)
+    { return GetOptionsDB().OptionExists(option) ? boost::python::object(GetOptionsDB().Get<double>(option)) : boost::python::object(); }
      boost::python::str GetUserConfigDirWrapper()
     { return boost::python::str(PathToString(GetUserConfigDir())); }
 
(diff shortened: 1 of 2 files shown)


o01eg:
Does it need additional methods in both `ServerWrapper` and `UniverseWrapper`? I suppose only one should be enough.

python/UniverseWrapper.cpp
```diff
@@ -239,6 +241,23 @@ namespace {
     { return rules.RuleExists(name); }
     bool RuleExistsWithType(const GameRules& rules, const std::string& name, GameRules::Type type)
     { return rules.RuleExists(name, type); }
+
+    boost::python::object GetOptionsDBOptionStr(std::string const &option)
```
geoffthemedio: `std::string const &option` -> `const std::string& option`

Cjkjvfnby: It is copypaste from AIWrapper

Cjkjvfnby:
> Does it need additional methods in both ServerWrapper and UniverseWrapper? I suppose only one should be enough.

I Checked usages in turn events and universe generation, looks like both uses methods defined in servers.

python/server/ServerWrapper.cpp
```diff
@@ -1392,4 +1409,15 @@ namespace FreeOrionPython {
         def("planet_make_colony",                   PlanetMakeColony);
         def("planet_cardinal_suffix",               PlanetCardinalSuffix);
     }
+
+        /////////////////
```
o01eg, line 1413: The code doesn't work outside functions.

python/server/ServerWrapper.cpp
```diff
@@ -1392,4 +1409,15 @@ namespace FreeOrionPython {
...
+        //   Configs   //
+        /////////////////
+        def("get_options_db_option_str",    GetOptionsDBOptionStr,     return_value_policy<return_by_value>(), "Returns the string value of option in OptionsDB or None if the option does not exist.");
+        def("get_options_db_option_int",    GetOptionsDBOptionInt,     return_value_policy<return_by_value>(), "Returns the integer value of option in OptionsDB or None if the option does not exist.");
+        def("get_options_db_option_bool",   GetOptionsDBOptionBool,    return_value_policy<return_by_value>(), "Returns the bool value of option in OptionsDB or None if the option does not exist.");
+        def("get_options_db_option_double", GetOptionsDBOptionDouble,  return_value_policy<return_by_value>(), "Returns the double value of option in OptionsDB or None if the option does not exist.");
+
+        def("get_user_config_dir",         GetUserConfigDirWrapper,        /* no return value policy, */ "Returns path to directory where FreeOrion stores user specific configuration.");
```
o01eg, line 1421: It's already defined in `WrapServer`

Vezzra (changes requested):
In their current form, the proposed changes won't work. I've provided some comments about what to change, I'm also going to provide a PR for your branch where you can see how I'd change things.

Morlic-fo:
Is there a reason to copy-paste the AIWrapper code to the ServerWrapper rather than moving it into a common source file?

Vezzra:
I wasn't aware that this wrapper code has been copied from the AIWrapper. If that's the case, then yes, instead of duplicating the code this should definitely be moved into the common wrappers code.

There are already existing source files for common wrappers, so that's where this should be moved to.

Cjkjvfnby:
Hi,   thanks for collaborating. 

Looks like I should not code C++. (especially in notepad).   I am not common with project structure too. 

I am not compiling, so I cannot test it now. I use Monday builds.

Feel free to push to my repo or abandon this branch and push to master in other way (Looks like nothing from my code is left).

o01eg:
@Cjkjvfnby Are you going to close this PR or accept changes from @Vezzra or me?

Vezzra:
> @Cjkjvfnby Are you going to close this PR or accept changes from @Vezzra or me?

Actually both this PR and my proposed changes should be discarded, and @Morlic-fo's suggestion implemented instead (move the functions copied over from the AIWrapper to the common wrappers), to avoid code duplication.

Vezzra:
@Cjkjvfnby, @o01eg: I have provided a PR which implements moving the wrappers from the AI to the common wrappers instead of duplicating them for the server: #2271.

Dilvish-fo:
I don' think we need to wait for the other PR to be completed/accepted before closing this one as superseded.

PR #2268, opened 2018-09-07
Sanitize string input before calling python exec.

Morlic-fo:
```exec``` is dangerous. Let's be safe.

### DIFF
--- a/python/CommonFramework.cpp
+++ b/python/CommonFramework.cpp
@@ -4,6 +4,7 @@
 #include "../util/Logger.h"
 #include "CommonWrappers.h"
 
+#include <boost/filesystem.hpp>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: python/CommonFramework.cpp [hunk 1]
```diff
@@ -130,13 +132,21 @@ void PythonBase::Finalize() {
 }
 
 void PythonBase::SetCurrentDir(const std::string dir) {
+	if (!fs::exists(dir)) {
+		ErrorLogger() << "Tried setting current dir to non-existing dir: " << dir;
```
geoffthemedio: replace tabs with four spaces

Vezzra (approved):

PR #2271, opened 2018-09-09
Move some Python interface functions from AI to common wrappers

Vezzra:
This PR is intended as an alternative implementation of #2264, which moves the Python wrappers from the AI wrappers to the common wrappers instead of duplicating them for the server.

If accepted, will supersede #2264.

### DIFF
--- a/python/ConfigWrapper.cpp
+++ b/python/ConfigWrapper.cpp
@@ -1,0 +1,56 @@
+#include "../util/Directories.h"
+#include "../util/OptionsDB.h"
+
+#include <boost/python.hpp>
+
+#include <string>
+
+using boost::python::def;
+using boost::python::return_value_policy;
+using boost::python::return_by_value;
+
+namespace {
+    boost::python::object GetOptionsDBOptionStr(std::string const &option)
+    { return GetOptionsDB().OptionExists(option) ? boost::python::str(GetOptionsDB().Get<std::string>(option)) : boost::python::str(); }
+
+    boost::python::object GetOptionsDBOptionInt(std::string const &option)
+    { return GetOptionsDB().OptionExists(option) ? boost::python::object(GetOptionsDB().Get<int>(option)) : boost::python::object(); }
+
+    boost::python::object GetOptionsDBOptionBool(std::string const &option)
+    { return GetOptionsDB().OptionExists(option) ? boost::python::object(GetOptionsDB().Get<bool>(option)) : boost::python::object(); }
+
+    boost::python::object GetOptionsDBOptionDouble(std::string const &option)
+    { return GetOptionsDB().OptionExists(option) ? boost::python::object(GetOptionsDB().Get<double>(option)) : boost::python::object(); }
+
+    boost::python::str GetUserConfigDirWrapper()
+    { return boost::python::str(PathToString(GetUserConfigDir())); }
+
+    boost::python::str GetUserDataDirWrapper()
+    { return boost::python::str(PathToString(GetUserDataDir())); }
+}
+
+namespace FreeOrionPython {
+    void WrapConfig()
+    {
+        // We provide the Python wrappers both with camelCase and with snake_case names,
+        // as these wrappers are used both by the AI and the universe generation scripts.
+        // The former still uses camelCase, while the latter uses snake_case, so we need
+        // to support both styles for the time being.
+
+        def("getOptionsDBOptionStr",    GetOptionsDBOptionStr,     return_value_policy<return_by_value>(), "Returns the string value of option in OptionsDB or None if the option does not exist.");
+        def("getOptionsDBOptionInt",    GetOptionsDBOptionInt,     return_value_policy<return_by_value>(), "Returns the integer value of option in OptionsDB or None if the option does not exist.");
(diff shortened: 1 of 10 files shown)


Dilvish-fo (approved):
Builds and runs fine for me on Ubuntu Linux.  Sounds like a good plan, looks fine, though I'll note that I simply did a rather cursory comparison of the moved parts.

Is it not necessary to make changes to the MS and OSX project files to specify ConfigWrapper.cpp, or do they now pick up changes specified for cmake?

python/ConfigWrapper.cpp
```diff
@@ -0,0 +1,56 @@
...
+    { return boost::python::str(PathToString(GetUserDataDir())); }
+}
+
+namespace FreeOrionPython {
+    void WrapConfig()
+    {
+        // We provide the Python wrappers both with camelCase and with snake_case names,
+        // as these wrappers are used both by the AI and the universe generation scripts.
+        // The former still uses camelCase, while the latter uses snake_case, so we need
+        // to support both styles for the time being.
```
Morlic-fo: While a unified naming convention would certainly be preferable, can we use preprocessor defines to avoid namespace pollution in the meantime?
```
#ifdef FREEORION_BUILD_SERVER
#ifdef FREEORION_BUILD_AI
```

Vezzra:
> Is it not necessary to make changes to the MS and OSX project files to specify ConfigWrapper.cpp, or do they now pick up changes specified for cmake?

Argh, of course, for the MSVC projects that's still needed. On OSX everything has been switched to cmake already.

@geoffthemedio, I have only MSVC 2015 installed, so I could only make the changes to the MSVC 2015 project. Can you make the changes to the MSVC 2017 project and push them to the PR branch (or provide a PR [on my fork](https://github.com/Vezzra/freeorion) for the PR branch)?

python/ConfigWrapper.cpp
Vezzra: Yep, sounds reasonable. Fixed.

Cjkjvfnby: Migration procedure is easy and can be done in steps. 

- add python wrapper to AI that will provide old and new version (We already patch AI [here ](https://github.com/freeorion/freeorion/blob/master/default/python/AI/freeorion_tools/extend_freeorion_AI_interface.py))
- update usage
- update C wrapper

Maybe we should create an issue about it and do some discussion?

Vezzra: Migrating all of the Python code to snake_case is certainly something that should be done, but is beyond the scope of this PR.

Opening an issue for this is a good idea, the discussion should happen on the forum though.

Vezzra:
> @geoffthemedio, I have only MSVC 2015 installed, so I could only make the changes to the MSVC 2015 project. Can you make the changes to the MSVC 2017 project and push them to the PR branch (or provide a PR on my fork for the PR branch)?

@geoffthemedio, bump? Or shall I go ahead and merge this PR, and you'll make the required changes to the MSVC 2017 project afterwards?

geoffthemedio:
@Vezzra Sorry, there's not much distinction between being pinged and generic updates to pull requests in the notifications. I don't want to deal with figuring out how to get the right branches checked out and referencing eachother, so just merge and I can try to add the files later on master...

Vezzra:
Ok!

Dilvish-fo:
> I don't want to deal with figuring out how to get the right branches checked out and referencing eachother, so just merge and I can try to add the files later on master...

The drawback to that is that (I am pretty sure) now all appveyor CI checks will fail until @geoffthemedio gets those changes made, requiring people to look very closely at the report to sort out if the failure is from this or from some problem in their own PR.  Although we've had periodic bouts of superfluous CI failures, they are overall helpful and, for example, I think that agrrr3 is relying on them as he tries sorting out problem with his PR 2249 which compiles fine on his machine but not for the CI (nor for me, but I don't think I have time just now to try helping him solve that, perhaps later).  So, I hope Geoff (or someone else) will be able to make those project adjustments soon, and I would urge that we **not** take that merge-first-fix-later approach in the future, not for problems that trigger CI fails.

Making a PR into Vezzra's branch could have been done by (i) add Vezzra's fork as a remote, (ii) pull Vezzra's ```move-ai-wrappers-to-common-wrappers``` branch (identified at the top of this PR), (iii) make the required changes, (iv) Geoff pushes that modified branch to his own fork (or to a branch on this repo if Geoff must really do all his work via this repo instead of using his own fork at all), and then (v) make a PR from that new branch back to Vezzra's branch (the github interface makes it easy to specify repo and branch for both ends of the PR).

Vezzra:
> The drawback to that is that (I am pretty sure) now all appveyor CI checks will fail until @geoffthemedio gets those changes made

I thought our AppVeyor CI checks still work with MSVC 2015...?

Morlic-fo:
No: https://github.com/freeorion/freeorion/blob/master/appveyor.yml#L27

Vezzra:
Ok, not good. In that case @Dilvish-fo is right, we should avoid merging something we know will break the CI checks.

Looks like I need to upgrade to MSVC 2017 soon... 😓

PR #2273, opened 2018-09-09
[AI] Replace prints

Cjkjvfnby:
I want to replace print to logging. 

`print` has the same logging level as  `debug`.  We have about 500 entries.

The fast way is to replace them all and when we really need that replace some debug to info.
The long way is to discuss if we need info or debug on each item.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -287,9 +287,9 @@ def generateOrders():  # pylint: disable=invalid-name
     debug("Generating Orders")
     debug("EmpireID: {empire.empireID}"
           " Name: {empire.name}_{empire.empireID}_pid:{p_id}_{p_name}RIdx_{res_idx}_{aggression}"
-          " Turn: {turn}").format(empire=empire, p_id=fo.playerID(), p_name=fo.playerName(),
-                                  res_idx=ResearchAI.get_research_index(), turn=turn,
-                                  aggression=aggression_name.capitalize())
+          " Turn: {turn}".format(empire=empire, p_id=fo.playerID(), p_name=fo.playerName(),
+                                 res_idx=ResearchAI.get_research_index(), turn=turn,
+                                 aggression=aggression_name.capitalize()))
     debug("EmpireColors: {0.colour.r} {0.colour.g} {0.colour.b} {0.colour.a}".format(empire))
     if planet:
         debug("CapitalID: " + str(planet_id) + " Name: " + planet.name + " Species: " + planet.speciesName)
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -65,7 +65,7 @@ def startNewGame(aggression_input=fo.aggression.aggressive):  # pylint: disable=
     Should clear any pre-existing state and set up whatever is needed for AI to generate orders."""
     empire = fo.getEmpire()
     if empire is None:
-        print "This client has no empire. Ignoring new game start message."
+        info("This client has no empire. Ignoring new game start message.")
         return
 
     if empire.eliminated:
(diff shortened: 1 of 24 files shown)


Dilvish-fo (approved):
> The fast way is to replace them all and when we really need that replace some debug to info.

I'm fine with them all being replaced with the debug function for now at least; individual cases could always be adjusted later if someone thinks them more or less important.

I can only guess that you set the milestone as gateway to void because you were thinking you'd resubmit a new PR covering all the files, but rather than needlessly multiply the number of PRs please just add commits here, and you can request a new review when done.

Or else please explain why you think it best to just void this PR.

Cjkjvfnby:
> Or else please explain why you think it best to just void this PR.

Checked contributing guide. Looks like I should leave it empty.

Vezzra:
That's probably something that should be done for all Python scripts, not only for the AI ones, but it's probably a good idea to make separate PRs for the different "areas", to avoid one overly massive PR...

Cjkjvfnby:
At some point, I understand that logging already has formatting capability. We can remove `%` and pass format arguments as args.

Deep in logging, it will be formatted:
```
        if self.args:
            msg = msg % self.args
```

```diff
- print "Requeueing %s to front of build queue, with result %d" % (building_name, res)
+ debug("Requeueing %s to front of build queue, with result %d", building_name, res)
```

default/python/AI/FreeOrionAI.py
```diff
@@ -64,7 +64,7 @@ def startNewGame(aggression_input=fo.aggression.aggressive):  # pylint: disable=
     Should clear any pre-existing state and set up whatever is needed for AI to generate orders."""
     empire = fo.getEmpire()
     if empire is None:
-        print "This client has no empire. Ignoring new game start message."
+        info("This client has no empire. Ignoring new game start message.")
```
Dilvish-fo: This seems like enough of an unexpected situation that it should be debug or warn rather than merely info.

default/python/AI/FreeOrionAI.py
```diff
@@ -102,15 +102,15 @@ def startNewGame(aggression_input=fo.aggression.aggressive):  # pylint: disable=
 def resumeLoadedGame(saved_state_string):  # pylint: disable=invalid-name
     """Called by client to when resume a loaded game."""
     if fo.getEmpire() is None:
-        print "This client has no empire. Doing nothing to resume loaded game."
+        info("This client has no empire. Doing nothing to resume loaded game.")
```
Dilvish-fo: As above, this seems like enough of an unexpected situation that it should be debug or warn rather than merely info.

default/python/AI/FreeOrionAI.py
```diff
@@ -146,7 +146,7 @@ def prepareForSave():  # pylint: disable=invalid-name
     in the state string so that they can be restored if the game is loaded."""
     empire = fo.getEmpire()
     if empire is None:
-        print "This client has no empire. Doing nothing to prepare for save."
+        info("This client has no empire. Doing nothing to prepare for save.")
```
Dilvish-fo: As above, this seems like enough of an unexpected situation that it should be debug or warn rather than merely info.

default/python/AI/FreeOrionAI.py
```diff
@@ -172,7 +172,7 @@ def handleChatMessage(sender_id, message_text):  # pylint: disable=invalid-name
     message_text is the text of the sent message."""
     empire = fo.getEmpire()
     if empire is None:
-        print "This client has no empire. Doing nothing to handle chat message."
+        info("This client has no empire. Doing nothing to handle chat message.")
```
Dilvish-fo: As above, this seems like enough of an unexpected situation that it should be debug or warn rather than merely info.

default/python/AI/InvasionAI.py
```diff
@@ -43,7 +43,7 @@ def get_invasion_fleets():
                                  (sys_id != INVALID_ID) and universe.systemsConnected(sys_id, home_system_id,
                                                                                       empire_id)]
     else:
-        print "Warning: Empire has no identifiable homeworld; will treat all visible planets as accessible."
+        warn("Empire has no identifiable homeworld; will treat all visible planets as accessible.")
```
Dilvish-fo: It seems not erroneous that an empire could still have ships but no remaining homeworld, so this should be just debug not warn, I think.

Dilvish-fo (changes requested):
Mostly looks good.  There are a few spots where I think a different logger level would be more appropriate, as noted in individual comments.  Could you please confirm you've also done some runtime testing?

default/python/AI/FreeOrionAI.py
Morlic-fo: > it should be debug or warn

If we can't get an empire after just creating a game, it seems we should log as ```fatal``` and terminate the AI process to be honest.

Cjkjvfnby: I will set it fatal but will not terminate anything,  this AI log will have only this messages, so it does not matter what error level we use.

default/python/AI/InvasionAI.py
Cjkjvfnby: I am a simple man, I see `Warning` I use `warn`.

Cjkjvfnby:
I like linter, it watches me. Looks like I need to set up some GitHub hooks to get report earlier.


PS. Do you know that you can pass `.\AI\fleet_orders.py:62:105` to `navigate to file` action to navigate to exact char.

Dilvish-fo (approved):

PR #2276, opened 2018-09-14
Clarify enqueue location

Dilvish-fo:
As discussed in #2251 the EnqueuLocation building methods currently have descriptions that do not perfectly align with how they are actually used, and the substantive check includes a wrong reference to the Location condition. 

This PR eliminates the wrong bit of code, and clarifies the UI-filter-helper nature of the EnqueueLocation methods, and how it is intended to work in conjunction with the Location condition (and need not have any overlap with it).


geoffthemedio:
haven't tested this myself... but is there some other reason to not yet merge?

Dilvish-fo:
I think I've just been distracted with a bunch of stuff.

PR #2279, opened 2018-09-15
Clean up unused player status.

o01eg:
As it was noted on forum multiple times (https://freeorion.org/forum/viewtopic.php?f=6&t=9732 andhttps://freeorion.org/forum/viewtopic.php?f=6&t=11048) interactive combat isn't planned in FreeOrion.

### DIFF
--- a/UI/PlayerListWnd.cpp
+++ b/UI/PlayerListWnd.cpp
@@ -49,10 +49,6 @@ namespace {
         static std::shared_ptr<GG::Texture> retval = ClientUI::GetTexture(ClientUI::ArtDir() / "icons" / "ready.png");
         return retval;
     }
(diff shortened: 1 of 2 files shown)


Vezzra:
Although the general consensus seems to lean heavily toward that direction (to abandon plans for any kind of interactive combat resolution), there hasn't really been a final decision yet.

So, not sure if we should remove these code bits already - @geoffthemedio?

geoffthemedio:
I'm fine with removing speculative stuff like this that's not being used. In the event it's needed, it can be easily added back...

Vezzra:
Alright, no further objections from my side, as far as I'm concerned, this can be merged.

PR #2280, opened 2018-09-16
Add option to enable/disable empire statistics.

o01eg:
First way to solve disclosure empire statistics in multiplayer game. Adds server option `network.server.publish-statistics` to overall on/off empire statistics.

Forum thread: https://freeorion.org/forum/viewtopic.php?f=6&t=11069

### DIFF
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1098,7 +1098,7 @@ namespace {
                                      " cause: " + m_specific_cause_name);
 
             // skip inactive sources
-            // FIXME: is it safe to move this out of the loop? 
+            // FIXME: is it safe to move this out of the loop?
             // Activation condition must not contain "Source" subconditions in that case
             const auto* activation = m_effects_group->Activation();
             if (activation && !activation->Eval(source_context, source))
@@ -1578,7 +1578,7 @@ void Universe::ExecuteEffects(const Effect::TargetsCauses& targets_causes,
                     Effect::TargetsAndCause& targets_and_cause = targets_it->second;
                     Effect::TargetSet& targets = targets_and_cause.target_set;
 
-                    // this is a set difference/union algorithm: 
+                    // this is a set difference/union algorithm:
                     // targets              -= non_stacking_targets
                     // non_stacking_targets += targets
                     for (auto object_it = targets.begin();
@@ -2262,7 +2262,7 @@ namespace {
 
                         // special case for fleets: grant partial visibility if
                         // visible contained object is partially or better visible
-                        // this way fleet ownership is known to players who can 
+                        // this way fleet ownership is known to players who can
                         // see ships with partial or better visibility (and thus
                         // know the owner of the ships and thus should know the
                         // owners of the fleet)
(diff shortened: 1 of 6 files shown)


universe/Universe.cpp
```diff
@@ -2953,6 +2953,10 @@ void Universe::UpdateStatRecords() {
     }
 }
 
+void Universe::SetStatRecordsPublish(bool publish /*= true*/) {
```
geoffthemedio: `{ }` and single code line on same line

o01eg: Fixed.

util/SerializeUniverse.cpp
```diff
@@ -151,8 +151,13 @@ void Universe::serialize(Archive& ar, const unsigned int version)
         }
     }
 
-    ar  & BOOST_SERIALIZATION_NVP(m_stat_records);
-    DebugLogger() << "Universe::serialize : " << serializing_label << " " << m_stat_records.size() << " types of statistic";
+    if (Archive::is_saving::value && (!m_stat_records_publish)) {
```
geoffthemedio: will this also stop stat records from begin saved in a save game file? that might be unexpected behaviour...

o01eg: Is `Universe::serialize` used in save game as well? How can I test if I'm in the serialization to file or to network?

geoffthemedio: Yes.

I think checking if `GetUniverse().EncodingEmpire() == ALL_EMPIRES` would work. It's used in various serialization code to check if info is being sent to a particular empire / player.

o01eg: Fixed,

universe/Universe.h
```diff
@@ -520,6 +523,7 @@ class FO_COMMON_API Universe {
 
     std::map<std::string, std::map<int, std::map<int, double>>>
                                     m_stat_records;                     ///< storage for statistics calculated for empires. Indexed by stat name (string), contains a map indexed by empire id, contains a map from turn number (int) to stat value (double).
+    bool                            m_stat_records_publish;             ///< send empire statistics to the players.
```
geoffthemedio: is this variable beneficial vs. just polling the OptionsDB each time?

o01eg: `OptionsDB` isn't included in `SerializeUniverse.cpp`

geoffthemedio: I'm hesitant to add it as a member of Universe... seems like it should be a parameter to the function (though that's probably not possible in practice) or stored elsewhere. That said, there is already `m_encoding_empire`, which works sort of the same way. That said having one thing implemented questionably isn't a great reason to add another...

o01eg: Does `boost::serialize` have something like context?

geoffthemedio: Don't know... Is there a reason to not include OptionsDB in SerializeUniverse?

o01eg: I suppose it possible to transfer additional parameters in derived `Archive` class: https://www.boost.org/doc/libs/1_58_0/libs/serialization/doc/derivation.html
But it will definitely require amount of refactoring.

geoffthemedio: Seems unnecessary...

o01eg: Added OptionsDB instead of `m_stat_records_publish`.

geoffthemedio (commented):
Looks OK

PR #2281, opened 2018-09-17
Reduce population meter upon conquest of a planet

Morlic-fo:
The fact that population survives an invasion means that compared to 1-3 pop colonizations, 20+ turns of growth are effectively skipped. In particular, capturing natives early may lead to snowballing.

This PR proposes to slow down the invasion-snowball by killing some population upon conquest.


Morlic-fo:
Forum thread for discussion:
https://www.freeorion.org/forum/viewtopic.php?p=93608#p93608

Morlic-fo:
Adding WIP for missing AI work if the idea is generally found acceptable.

PR #2283, opened 2018-09-18
Don't ask client save data without alive empire

o01eg:
Remove unneeded requests. Prepares for "ingame lobby".


geoffthemedio:
You've tested this with some non-alive empires and re-loaded the resulting saves without issues?

o01eg:
Yep, I conceded one of player and then tried to load game after it without issues.

PR #2285, opened 2018-09-19
Remove AIInterface::DoneTurn() calls from Python AI

Morlic-fo:
There is a single logical place to call AIInterface::DoneTurn(). Let's call it at that particular time and avoid worrying about it in python code...

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -222,9 +222,13 @@ def handleDiplomaticStatusUpdate(status_update):  # pylint: disable=invalid-name
 
 @listener
 def generateOrders():  # pylint: disable=invalid-name
-    """Called once per turn to tell the Python AI to generate and issue orders to control its empire.
-    at end of this function, fo.doneTurn() should be called to indicate to the client that orders are finished
-    and can be sent to the server for processing."""
+    """
+    Called once per turn to tell the Python AI to generate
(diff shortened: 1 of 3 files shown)


Dilvish-fo (commented):
I think this will likely have some minor conflicts with #2273 , which has more commits and has been pending longer so I've like to let CJ wrap that up first, then have you deal with any conflicts here before merging this.

Also, just to note, there have been discussions on the forums, focused on multiplayer games, about making the turn-end process still allow player to submit more orders until everyone is done.  If that gets implemented, it could be beneficial to try having the AI split order generation into high-priority/fast orders and lower-priority and/or super-carefully-checked orders, the latter of which could be interrupted in some way if they drag on past a couple minutes if the human player is already finished.  That might easily call for reverting some of this, but would also involve a lot of other related changes.  So I don't really see that prospect as a reason to not do this, but still felt it worth noting.

Cjkjvfnby:
@Morlic, please resolve conflicts

geoffthemedio:
@Morlic-fo see the previous comment with (probably) the wrong name linked.

Cjkjvfnby:
@Morlic-fo do you plan to merge this?

Morlic-fo:
Will come back to this, thanks for reminder.

geoffthemedio:
needs rebase...

geoffthemedio:
@Morlic-fo ?

PR #2286, opened 2018-09-19
Don't set any roles when return player to lobby.

o01eg:
Fix issue when returned back to lobby players acquired roles he shouldn't to have or lost his roles.

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -586,13 +586,6 @@ MPLobby::MPLobby(my_context c) :
                     player_setup_data.m_starting_species_name = sm.SequentialPlayableSpeciesName(player_id);
 
                 m_lobby_data->m_players.push_back({player_id, player_setup_data});
-
-                player_connection->SetAuthRoles({
-                                Networking::ROLE_CLIENT_TYPE_MODERATOR,
-                                Networking::ROLE_CLIENT_TYPE_PLAYER,
-                                Networking::ROLE_CLIENT_TYPE_OBSERVER,
-                                Networking::ROLE_GALAXY_SETUP
-                                });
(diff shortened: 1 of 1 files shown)


o01eg:
@geoffthemedio  Found it while testing ingame lobby. If player don't get roles to become Moderator or Observer after current game is finished he is returned back to lobby and here get additional roles he mustn't obtain.

geoffthemedio:
The player gets the extra roles only if they don't have them initially, or regardless of the roles they had before the game? What if the player has the extra roles before the game? Do they retain them when being returned to the lobby? What if a player doesn't have galaxy setup adjustment roles before a game, do they then get them after the game?

o01eg:
> The player gets the extra roles only if they don't have them initially, or regardless of the roles they had before the game?

Regardless previous roles.

> What if the player has the extra roles before the game? Do they retain them when being returned to the lobby?

He will lose them as well.

> What if a player doesn't have galaxy setup adjustment roles before a game, do they then get them after the game?

In a hostless mode all players should capable to adjust galaxy setup. But, if someone was revoked this role, he obtain them here. Maybe I shouldn't change any roles when return to lobby. Thanks.

geoffthemedio:
So now if a player is in a game then returns to the lobby, they lose all roles? It doesn't check what roles the player should have if defined in `auth.txt` and they then instead have no roles...?

o01eg:
> So now if a player is in a game then returns to the lobby, they lose all roles?

No. After changes player have his roles untouched.

> It doesn't check what roles the player should have if defined in `auth.txt` and they then instead have no roles...?

Player connection obtains roles on establish and then never changes its roles.

geoffthemedio:
Your various replies seem to contradict eachother, but presumably the latest is accurate.

PR #2289, opened 2018-09-20
[AI] Rename Universe object to target

Cjkjvfnby:
Once upon a time, I introduced these names during refactoring. 

Now I think that the chosen names lead to confusion with AIIntrface objects.

### DIFF
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -13,7 +13,7 @@ import MilitaryAI
 import InvasionAI
 import CombatRatingsAI
 from aistate_interface import get_aistate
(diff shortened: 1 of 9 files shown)


Dilvish-fo:
I don't have much time during the next week to review anything carefully.  But at a quick glance I agree with the sentiment and approach of this PR.

Morlic-fo (commented):
Sure, why not.

Cjkjvfnby:
> I don't have much time during the next week to review anything carefully. But at a quick glance I agree with the sentiment and approach of this PR.

I am not in a hurry. I have run this code, it works without error.

Dilvish-fo (approved):

PR #2290, opened 2018-09-20
[AI] Docstring cleanup

Cjkjvfnby:
The reason for this PR is to show that there is the alternative syntax for type declaration in docstring.
I think we can use both, ways.  It produces some inconsistency, but I don't want to do a migration. 

Example:
```diff
-  :param state: savegame state, modified in function
-  :type state: dict
-  :param version: Version to convert to
-  :type version: int
+ :param dict state: savegame state, modified in function
+ :param int version: Version to convert to
```

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -36,6 +36,7 @@ class ConversionError(Exception):
 
     Automatically logs and chats to the host if raised.
     """
+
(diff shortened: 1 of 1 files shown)


Dilvish-fo (approved):
Sure, looks like a more concise and probably more easily readable format that would be good to use going forward.

PR #2292, opened 2018-09-26
[wrapper] Remove duplicated declaration

Cjkjvfnby:
The second copy of the method was introduced in the commit.
https://github.com/freeorion/freeorion/commit/d8088cd82c25aba17827461bcfe856b5843de137#diff-058ed30f9c557ea7d8387a675bb3ce01R53

### DIFF
--- a/python/server/ServerWrapper.cpp
+++ b/python/server/ServerWrapper.cpp
@@ -1296,7 +1296,6 @@ namespace FreeOrionPython {
         def("user_string",                          make_function(&UserString,      return_value_policy<copy_const_reference>()));
         def("roman_number",                         RomanNumber);
         def("get_resource_dir",                     GetResourceDirWrapper);
-        def("get_user_config_dir",                  GetUserConfigDirWrapper);
 
         def("all_empires",                          AllEmpires);
         def("invalid_object",                       InvalidObjectID);


Cjkjvfnby:
> though I also seem to recall some plan to separate such by compiler directives so they wouldn't both be exposed at the same time

Did we just discuss to change AI name style to match with Universe one?

Cjkjvfnby:
More errors for the universe generation:
```
Error in "fieldType.dump" with error: Python argument types in
    None.None(fieldType)
did not match C++ signature:
    None(class FieldType {lvalue}, unsigned short)
```
```
Error in "buildingType.dump" with error: Python argument types in
    None.None(buildingType)
did not match C++ signature:
    None(class BuildingType {lvalue}, unsigned short)
```
```
Error in "special.dump" with error: Python argument types in
    None.None(special)
did not match C++ signature:
    None(class Special {lvalue}, unsigned short)
```
```
Error in "species.dump" with error: Python argument types in
    None.None(species)
did not match C++ signature:
    None(class Species {lvalue}, unsigned short)
```

Dilvish-fo:
> Did we just discuss to change AI name style to match with Universe one?

I'm not sure the word 'just' fits in there, but in general [that is as I understand the plan for what to eventually do](https://github.com/freeorion/freeorion/pull/2271#discussion_r216379110), and looking into that more closely now it seems that the change of this PR would be both the wrong one to delete, and rather premature/incomplete for when we do move forward with harmonizing


> Do we change our policy to unused code?

Could you explain your thought more than just linking to a 6 year old commit/line?


> More errors for the universe generation:

When do you get those errors, and how are they related to this PR?

Vezzra (approved):
Oops, good catch, oversight on my part. I forgot to remove that declaration from `ServerWrapper.cpp` in PR #2271 (where I moved the of the Python interface functions to the common wrappers).

Vezzra:
@Dilvish-fo:

> I notice that for both this and a similar method just after it, the commit provides two exposures for the underlying method, one with CamelCase and one with snake_case.

Yes, but both of these are in the new `ConfigWrapper.cpp`, the one @Cjkjvfnby wants to remove in this PR here is the declaration in `ServerWrapper.cpp`, which is supposed to be replaced by the ones in `ConfigWrapper.cpp`.

Cjkjvfnby:
> Could you explain your thought more than just linking to a 6 year old commit/line?

This method is unused.   My question was do we plan to remove unused methods or keep them.
I linked to the line where it is defined (it was open in blame mode when I copy link).

> When do you get those errors, and how are they related to this PR?
Not related.  Methods that have arguments are bound to python as properties. 

What you will suggest me to do with them:
1) remove, because nobody uses them
2) make an issue
3) Make PR where `add_property` replaced with `def`

Dilvish-fo:
> This method is unused. My question was do we plan to remove unused methods or keep them.

If you'd like to double check the code and confirm that  ```getSystemNeighborsMap```  just provides the same information as ```getImmediateNeighbors``` in conjunction with ```linearDistance```  then sure, it's redundant, not particularly more convenient and could be removed.

> What you will suggest me to do with them:

I'd suggest either fixing them per your third choice, or opening a thread in the programming forum to discuss the possibility of taking them out if that was your strong preference

Cjkjvfnby:
> If you'd like to double check the code and confirm that getSystemNeighborsMap just provides the same information as getImmediateNeighbors in conjunction with linearDistance then sure, it's redundant, not particularly more convenient and could be removed.

Out of my expertise.   It is not enough documented to be useful.

Cjkjvfnby:
> I'd suggest either fixing them per your third choice

Submitted PR to initial dump author.
https://github.com/freeorion/freeorion/pull/2296

PR #2293, opened 2018-09-27
Add universe_generator stub generator

Cjkjvfnby:
Now you can generate both AI and universe generation in the single command. 

```
<game>\freeorion.exe --ai-config <repo>\freeorion\default\python\handlers\inspect_interface_config.ini
```


default/python/handlers/inspect_universe_generation.py
```diff
@@ -0,0 +1,44 @@
+from common.listeners import register_post_handler
+from stub_generator import inspect
+
+
+def inspect_universe_generation_interface(*args, **kwargs):
+    import freeorion as fo
+    tech = fo.getTech('LRN_ARTIF_MINDS')
+    universe = fo.get_universe()
+    # empire = fo.get_empire(0)
```
Dilvish-fo: Were you wanting to include an empire in your instances below?  It seems to me you would.  I think you just need to specify empire 1 rather than empire 0.

Cjkjvfnby: Updated.

Dilvish-fo (approved):

PR #2294, opened 2018-09-27
Update freeorion.pyi

Cjkjvfnby:


Dilvish-fo:
It seems there are a lot of changes flagged simply because of sorting alphabetically in the new version.  Is there anything in particular you'd like to call my attention to?

Cjkjvfnby:
> It seems there are a lot of changes flagged simply because of sorting alphabetically in the new version. Is there anything in particular you'd like to call my attention to?

No.  I think I will push such changes myself.    

At my job merge is restricted without approving, so it reflexes to assigh reviewers to PR.

PR #2295, opened 2018-09-27
Update stub generator

Cjkjvfnby:
- fixed crash when multiple functions with same name and arguments
- fix crash on logging
- print trace in case of errors
- improve logging

### DIFF
--- a/default/python/stub_generator/parse_docs.py
+++ b/default/python/stub_generator/parse_docs.py
@@ -146,7 +146,7 @@ def merge_args(name, raw_arg_types, is_class):
     return ['%s=None' % name for name in names] if use_keyword else names, zip(names, types)
 
 
-def normilize_rtype(rtype):
+def normalize_rtype(rtype):
     if rtype == 'iterator':
         return 'iter'
     return rtype
@@ -188,7 +188,7 @@ class Docs(object):
         args, rtypes, infos = zip(*res)
         if len(set(rtypes)) != 1:
             error("[%s] Different rtypes", name)
-        self.rtype = normilize_rtype(rtypes[0])
+        self.rtype = normalize_rtype(rtypes[0])
 
         # cut of first and last string if they are empty
         # we cant cut off all empty lines, because it can be inside docstring
--- a/default/python/stub_generator/parse_docs.py
+++ b/default/python/stub_generator/parse_docs.py
@@ -143,7 +143,7 @@ def merge_args(name, raw_arg_types, is_class):
               '\n    '.join(', '.join('(%s)%s' % (tp, name) for tp, name in arg_set) for arg_set in raw_arg_types))
         names, types = get_argument_names(raw_arg_types[0], is_class)
         use_keyword = False
-    return ['%s=None' % name for name in names] if use_keyword else names, zip(names, types)
+    return ['%s=None' % arg_name for arg_name in names] if use_keyword else names, zip(names, types)
 
 
 def normalize_rtype(rtype):
--- a/default/python/stub_generator/parse_docs.py
+++ b/default/python/stub_generator/parse_docs.py
@@ -63,7 +63,8 @@ normalization_dict = {
     'ShipPartMeterMap': 'ship_part_meter_map',
     'ShipSlotVec': 'ship_slot_vec',
     'special': 'special',
-    'IntFltMap': 'int_flt_map'
+    'IntFltMap': 'int_flt_map',
+    'ruleType': 'rule_type',
 }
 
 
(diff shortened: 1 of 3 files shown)


default/python/stub_generator/parse_docs.py
```diff
@@ -110,16 +111,37 @@ def parse_name(txt):
...
+    Single argument set is used as is.
+    If we have two unique argument sets, and on of them is empty, use keywords.
+    In other cases log error and use first one.
+
+    :param str name:
+    :param list[tuple] raw_arg_types:
+    :param bool is_class:
+    :rtype: (list[str], list[(str, str)])
+    """
+    # If wrapper define functions that have same name, and same arguments but different return types,
```
Dilvish-fo, line 127: Is there an example of where this code is needed with our wrappers?  I am having a hard time seeing how it would not be a mistake for the wrapper to try specifying two function definitions with the same names and arguments but different return types.

default/python/stub_generator/parse_docs.py
```diff
@@ -110,16 +111,37 @@ def parse_name(txt):
...
+    # it will come here with len(arg_types) >= 2, where all arguments set are the same.
+    size = len(raw_arg_types)
+    arg_types = sorted(set(raw_arg_types))
+    if len(arg_types) != size:
+        error("[%s] Duplicated argument types", name)
```
Dilvish-fo, line 132: I'm not understanding why it should be an error if two of the arguments have the same type-- could you explain?

default/python/stub_generator/parse_docs.py
```diff
@@ -164,9 +186,9 @@ def parse_signature(line):
         self.resources = res
 
         args, rtypes, infos = zip(*res)
-        rtypes = set(rtypes)
-        assert len(rtypes) == 1, "Different rtypes for: %s in: %s" % (rtypes, text)
-        self.rtype = normilize_rtype(rtypes.pop())
+        if len(set(rtypes)) != 1:
+            error("[%s] Different rtypes", name)
+        self.rtype = normilize_rtype(rtypes[0])
```
Dilvish-fo: Since we are here, please also correct the spelling of this function name to ```normalize_rtype```

I also see a number of places where 'routine' is misspelled as 'rutine'; it's all done consistently so it doesn't cause errors, but it would be nice to get that corrected also.

Dilvish-fo (commented):
Some preliminary questions/comments...

default/python/stub_generator/parse_docs.py
Cjkjvfnby: > I am having a hard time seeing how it would not be a mistake for the wrapper to try specifying two function definitions with the same names and arguments but different return types.

It is a mistake which compiles.   I did not ever think that this is possible.
https://github.com/freeorion/freeorion/pull/2292

default/python/stub_generator/parse_docs.py
Cjkjvfnby: `arg_types` is a list of argument tuples when method overloaded.

Example message when method have two sets of arguments and they are different.
```
[canBuild] Cannot merge, use first argument group from:
(empire)arg1, (buildType)arg2, (str)arg3, (int)arg4
(empire)arg1, (buildType)arg2, (int)arg3, (int)arg4
```

This line prevents crash when they all overloads have same signature.
This can happen when there is an error in wrapper with a duplicated declaration.
Old code also crashed in case of `arg_types= [(), ()]`

default/python/stub_generator/parse_docs.py
Dilvish-fo: > It is a mistake which compiles. I did not ever think that this is possible. # 2292

That example you cite is not really an example of having different return types, it seems to me.  But I suppose it's close enough to show that the problem of semi-redundant declarations with different return types *could* happen.  And in case that did happen, then the error logging here would at least flag that in a log file, ok.

Dilvish-fo:
OK, thanks for the explanations.  Also please note that the CI fail in this case wasn't just another unrelated fail, it was a pylint fail.

Cjkjvfnby:
>  Also please note that the CI fail 

Do we have some way to get email on failure?  Or some prepared git hooks?

Dilvish-fo:
> Do we have some way to get email on failure?  Or some prepared git hooks?

If you set up an account with [Travis](https://travis-ci.org) and follow their instructons to link it to your github account, then one of your acount settings there lets you get emails with the status of your builds.  Hmm, although, looking into it more while writing this, and checking my email logs, it is sort of seeming to me that Travis only counts it as 'my build' if I manually restarted the build after an initial failure.  Perhaps it's because our travis.yml file does not (to my review) seem to have general email notifications turned on.  [It appears that should be easy enough to do.](https://docs.travis-ci.com/user/notifications/)  You could add yourself to the recipient list and just make some sort of filter to jettison any notifications except for fails on your own commits.

Dilvish-fo (approved):

PR #2296, opened 2018-09-28
Change dump from property to function

Cjkjvfnby:
Fix wrapper for the dump.  This method requires argument, and cannot be property. 
I don't sure if this method is useful since it was broken for a long time and nobody complains about it. 

PS. I did not compile this code.

### DIFF
--- a/python/UniverseWrapper.cpp
+++ b/python/UniverseWrapper.cpp
@@ -334,7 +334,7 @@ namespace FreeOrionPython {
         class_<Meter, noncopyable>("meter", no_init)
             .add_property("current",            &Meter::Current)
             .add_property("initial",            &Meter::Initial)
-            .def("dump",                        &Meter::Dump)
+            .def("dump",                        &Meter::Dump, return_value_policy<return_by_value>(), "Returns string with debug information, use '0' as argument.")
         ;
 
         ////////////////////
@@ -444,7 +444,7 @@ namespace FreeOrionPython {
                                                     return_value_policy<return_by_value>()
                                                 ))
 
-            .def("dump",                        &DumpObjects)
+            .def("dump",                        &DumpObjects, return_value_policy<return_by_value>(), "Returns string with debug information, use '0' as argument.")
         ;
 
         ////////////////////
@@ -474,7 +474,7 @@ namespace FreeOrionPython {
             .def("hasTag",                      &UniverseObject::HasTag)
             .add_property("meters",             make_function(ObjectMeters,                 return_internal_reference<>()))
             .def("getMeter",                    make_function(ObjectGetMeter,               return_internal_reference<>()))
-            .def("dump",                        &UniverseObject::Dump)
+            .def("dump",                        &UniverseObject::Dump,                      return_value_policy<return_by_value>(), "Returns string with debug information, use '0' as argument.")
         ;
 
         ///////////////////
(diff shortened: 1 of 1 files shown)


Dilvish-fo:
It's a debugging-aid type function, the one that also is used for the Dump command from the Object Window.  I certainly use that sometimes, and it seems it could also be useful to be able to get that info in the AI logs when debugging things, so I'm inclined to think it should stay.

But it would probably also be best to clarify what the required argument is (since my recollection is that in python we can't rely on the fact that in the underlying c++ function the argument has a default value).  In this case it's an integer ntabs (number of tabs?, can't be clear on that since it actually appears to me to be unused in all of the current Dump() implementations, and yes that would probably be something best cleaned up if someone found the time for it), so the comment should probably for now just say to specify a zero value.   

For an example of specifying a docstring, you could look at ```getSpecial``` in the UniverseWrapper, although for this function I think we'll need to specify ```return_value_policy<return_by_value>()```

Cjkjvfnby:
Can we stop aligning as table and split to multiline rows?

```
.def("dump",                        &UniverseObject::Dump,                      return_value_policy<return_by_value>(), "Returns string with debug information, use '0' as argument")
```

```
.def("dump",
     &UniverseObject::Dump,
     return_value_policy<return_by_value>(), 
     "Returns string with debug information, use '0' as argument."
)
```

python/UniverseWrapper.cpp
```diff
@@ -444,7 +444,7 @@ namespace FreeOrionPython {
                                                     return_value_policy<return_by_value>()
                                                 ))
 
-            .def("dump",                        &DumpObjects)
+            .def("dump",                        &DumpObjects, return_value_policy<return_by_value>(), "Returns string with debug information, use '0' as argument.")
```
geoffthemedio: https://github.com/freeorion/freeorion/blob/master/python/UniverseWrapper.cpp

`DumpObjects` is different from standard content and object `Dump` functions.

geoffthemedio:
> Can we stop aligning as table and split to multiline rows?

In selected locations sure, but in general the Python wrapping has amny repetetive lines like this and would be much longer and, in my opinion, harder to read with each property or function wrapper spread over several lines.

python/UniverseWrapper.cpp
Cjkjvfnby: > DumpObjects is different from standard content and object Dump functions.

Can you provide more clear instructions?

geoffthemedio: Look at the signature of the functions, don't assume they're all the same because they have Dump in the name....

https://github.com/freeorion/freeorion/blob/master/python/UniverseWrapper.cpp#L52

no 0 argument, returns nothing

Cjkjvfnby: I will just rollback this.

python/UniverseWrapper.cpp
```diff
@@ -733,7 +733,7 @@ namespace FreeOrionPython {
             .add_property("description",        &Special::Description)
             .add_property("spawnrate",          make_function(&Special::SpawnRate,      return_value_policy<return_by_value>()))
             .add_property("spawnlimit",         make_function(&Special::SpawnLimit,     return_value_policy<return_by_value>()))
-            .add_property("dump",               &Special::Dump)
+            .def("dump",                        &Special::Dump)
```
geoffthemedio, line 736: why no return value or note about parameter for this one?

python/UniverseWrapper.cpp
```diff
@@ -579,7 +579,7 @@ namespace FreeOrionPython {
                                                 ))
 
             .def("productionLocationForEmpire", &ShipDesign::ProductionLocation)
-            .add_property("dump",               &ShipDesign::Dump)
+            .def("dump",                        &ShipDesign::Dump, return_value_policy<return_by_value>(), "Returns string with debug information, use '0' as argument.")
```
geoffthemedio, line 582: would prefer consistent indenting with preceeding code...

geoffthemedio:
@Cjkjvfnby ?

geoffthemedio:
@Cjkjvfnby ?

geoffthemedio:
Resolved my comments in https://github.com/freeorion/freeorion/commit/bf54787a9ec10070ecf98a224559d6a5f27e109d

PR #2298, opened 2018-10-15
Add debugging helper function assertion_fails

Morlic-fo:
Introduces a helper function which semi-automates error logging for sanity checks and gives context on failure by providing a traceback. It is intended to be used as a graceful alternative to the assert-statement.

An optional logging level can be specified to avoid error messages in chat.

Example usage:
```
for fleet_id in fleets_which_should_exist:
    fleet = universe.getFleet()
    # if not fleet:
    #    do some manual error logging
    #    continue
    if assertion_fails(fleet is not None):
        continue

    # do something on fleet object
```

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -341,6 +341,9 @@ def generateOrders():  # pylint: disable=invalid-name
                    ResourcesAI.generate_resources_orders,
                    ]
 
+    from freeorion_tools import assertion_fails
+    print assertion_fails(False)
+    print assertion_fails(True)
     for action in action_list:
         try:
             main_timer.start(action.__name__)
--- a/default/python/AI/freeorion_tools/_freeorion_tools.py
+++ b/default/python/AI/freeorion_tools/_freeorion_tools.py
@@ -4,6 +4,7 @@ import cProfile
 import logging
 import pstats
 import re
+import traceback
 from collections import Mapping
 from functools import wraps
 from logging import debug, error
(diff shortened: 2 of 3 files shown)


default/python/AI/FreeOrionAI.py
```diff
@@ -341,6 +341,9 @@ def generateOrders():  # pylint: disable=invalid-name
                    ResourcesAI.generate_resources_orders,
                    ]
 
+    from freeorion_tools import assertion_fails
+    print assertion_fails(False)
+    print assertion_fails(True)
```
Dilvish-fo, line 346: Are these leftover from some initial testing?

Dilvish-fo (commented):
Besides the minor comment about a few lines that seem to me to have been probably intended as temporary, I have mixed feelings about this.  It seems like the benefit is to eliminate (with each use) one-line for an explicit call to a logger, but at the cost of slightly obscuring the same logging and also slightly the polarity of the condition being tested, which I am not really sure is a good trade-off.  But perhaps that's all just a matter of me getting more used to this, so I don't object if you are convinced of the value.

default/python/AI/FreeOrionAI.py
Dilvish-fo: @Morlic-fo  I presume I am right that these were just accidentally left over, and they are making spurious message window spam with each turn, so I've pushed a commit now removing these lines.

Morlic-fo: yes, sorry and thanks for fixing.

PR #2301, opened 2018-10-16
Make sure SetShipPartMeter::Execute() takes effect.

### LINKED ISSUE #2300: Dynamic cast fails in Clang 6: SetShipPartMeter::Execute couldn't cast simple increment ValueRef to an Operation
### Environment

* **FreeOrion Version**: compiled from git feb528b3 = master Oct 7 15:48, 2018
* **Operating System**: OpenBSD 6.4-current
* **Graphic card used**: probably irrelevant, not graphics-related, but here you go:

  pchb0 at pci0 dev 0 function 0 "Intel Core 5G Host" rev 0x09
  inteldrm0 at pci0 dev 2 function 0 "Intel HD Graphics 5500" rev 0x09
  drm0 at inteldrm0
  inteldrm0: msi
  inteldrm0: 1920x1080, 32bpp

### Description

When the player researches an upgrade like "Mass Driver 2", the damage done by the respective weapons is supposed to increase by one point, and similar for other upgrades of this kind, as soon as the ships having these weapons come back to supply.  However, with some compilers, this does not happen; instead, nothing happens, the damage remains unchanged.  With the two attached patches, i see the following log entry (using "effect" log level "trace"):

  [error] server : Effect.cpp:693 : SetShipPartMeter::Execute couldn't cast simple increment ValueRef to an Operation. Reverting to standard execute.
  [error] server : Effect.cpp:695 : typeid(ValueRefBase): N8ValueRef12ValueRefBaseIdEE [13398446319136]
  [error] server : Effect.cpp:697 : typeid(Operation): N8ValueRef9OperationIdEE [13398446319168]
  [error] server : Effect.cpp:699 : typeid(m_value): N8ValueRef9OperationIdEE [13397535772400]

So the problem is that at least with the following compiler, the preceding dynamic_cast fails and returns 0:

  $ uname -a
  OpenBSD isnote.usta.de 6.4 GENERIC.MP#341 amd64
  $ which cc 
  /usr/bin/cc
  $ /usr/bin/cc --version                                        
  OpenBSD clang version 6.0.0 (tags/RELEASE_600/final) (based on LLVM 6.0.0)
  Target: amd64-unknown-openbsd6.4 
  Thread model: posix
  InstalledDir: /usr/bin

I'm not completely sure why the RTTI downcast fails:  The source class has virtual functions and is only a parent of the target class in one way, and the SimpleIncrement() check right before seems to indicate that the object is indeed of type ::Operation.  But from the above debugging output, it appears that while the typeid *names* match, the *hashes* do not, which may indicate some kind of subtle type mismatch somewhere deep in the guts of the parser, which is not exactly easy to read, so i'm sorry to admit that i was unable to find the root cause.

The *lower* attached patch is an immediate bugfix in the sense of a workaround, to commit right away.  The same workaround is already present at other places in the code.  The upper patch is *not* intended for commit but to help you get started finding the root cause in the autogenerated parser code.

Thanks, Ingo.

[SetShipPartMeter.patch.txt](https://github.com/freeorion/freeorion/files/2481017/SetShipPartMeter.patch.txt)

Vezzra:
Pull request for patch provided by @ischwarze in #2300.

Fixes #2300

### DIFF
--- a/universe/Effect.cpp
+++ b/universe/Effect.cpp
@@ -689,7 +689,8 @@ void SetShipPartMeter::Execute(const ScriptingContext& context, const TargetSet&
         // deep inspection single ValueRef evaluation
         auto op = dynamic_cast<ValueRef::Operation<double>*>(m_value.get());
         if (!op) {
-            ErrorLogger() << "SetShipPartMeter::Execute couldn't cast simple increment ValueRef to an Operation...";
+            ErrorLogger() << "SetShipPartMeter::Execute couldn't cast simple increment ValueRef to an Operation. Reverting to standard execute.";
+            EffectBase::Execute(context, targets);
             return;
         }
         // RHS should be a ConstantExpr


Vezzra:
Superseded by b0abc77d414d5a4922ed29acafdd7c3f6cd85c1f.

Issue #2297, opened 2018-10-02
Memory leak? AddressSanitizer fails


agrrr3:
### Environment
* Branch: master
* OS: debian buster
* **Fetched as**
  * Compiled from source; cmake; gcc (debian buster package 4:8.1.0-1) 

### Description

I tried to hunt down a suspected memory leak for the preferred targetting branch and enable the address and undefined behavior sanitizers in the build. That found a suspected problem, but the problem is also in current master:
```
[100%] Built target freeorion
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2895==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x7f2774350ac0 bp 0x7ffddaa231a0 sp 0x7ffddaa23180 T0)
==2895==The signal is caused by a READ memory access.
==2895==Hint: address points to the zero page.
    #0 0x7f2774350abf in boost::serialization::void_cast_detail::void_caster_primitive<BoutBeginEvent, CombatEvent>::void_caster_primitive() (/home/axel-testing/freeorion/freeorion_build_sanitized/libfreeorioncommon.so+0x68bbabf)
    #1 0x7f27743299bf in boost::serialization::detail::singleton_wrapper<boost::serialization::void_cast_detail::void_caster_primitive<BoutBeginEvent, CombatEvent> >::singleton_wrapper() (/home/axel-testing/freeorion/freeorion_build_sanitized/libfreeorioncommon.so+0x68949bf)
    #2 0x7f2774329a9a in boost::serialization::singleton<boost::serialization::void_cast_detail::void_caster_primitive<BoutBeginEvent, CombatEvent> >::get_instance() (/home/axel-testing/freeorion/freeorion_build_sanitized/libfreeorioncommon.so+0x6894a9a)
    #3 0x7f2774248aea in __static_initialization_and_destruction_0(int, int) (/home/axel-testing/freeorion/freeorion_build_sanitized/libfreeorioncommon.so+0x67b3aea)
    #4 0x7f2774252ea2 in _GLOBAL__sub_I_CombatEvents.cpp (/home/axel-testing/freeorion/freeorion_build_sanitized/libfreeorioncommon.so+0x67bdea2)
    #5 0x7f277f8130c9  (/lib64/ld-linux-x86-64.so.2+0xf0c9)
    #6 0x7f277f8131d5  (/lib64/ld-linux-x86-64.so.2+0xf1d5)
[...]

PR #2303, opened 2018-10-17
Dont binary deserialize xml files when looking for headers

agrrr3:
This fixes memory leak issue #2297 and does some cleanup by assuming that we have two cases: xml files starting with an xml tag and binary files.


geoffthemedio:
Looks reasonable. Equivalent not necessary when actual loading as well?

PR #2305, opened 2018-10-22
[UI/About] clean up

VPeruS:
Buttons now have shared layout code.
Created difference between border_size and space between info text and buttons.

### DIFF
--- a/UI/About.cpp
+++ b/UI/About.cpp
@@ -40,14 +40,14 @@ About::About():
 void About::CompleteConstruction() {
     CUIWnd::CompleteConstruction();
 
-    m_done_btn = Wnd::Create<CUIButton>(UserString("DONE"));
+    m_done = Wnd::Create<CUIButton>(UserString("DONE"));
     m_license = Wnd::Create<CUIButton>(UserString("LICENSE"));
     m_vision = Wnd::Create<CUIButton>(UserString("VISION"));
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Looks reasonable. Built and ran OK for me on Wiindows, so disregarding the failed AppVeyor test.

VPeruS:
Maybe it was better to remove destructor completely, as it's not marked as virtual and mark class as final?

geoffthemedio:
Apparently marking derived class destructors as virtual is not recommended: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final

> If a base class destructor is declared virtual, one should avoid declaring derived class destructors virtual or override.

I don't immediately see why they suggest that, though... it doesn't seem to be addressed in the surrounding points.

Making the class final seems reasonable.

VPeruS:
Yeah, if inheritance chain is long then to ensure that class definition is correct you will need to find base class, and it will be tedious

VPeruS:
There is some disscussion about c128 https://github.com/isocpp/CppCoreGuidelines/issues/721

> 1.  If the base class dtor is virtual, both virtual and override on the derived class dtor are used only for the documentary purposes and from this point of view they're equivalent. However the latter is more consistent with the normal methods and hence, IMHO, slightly preferable.
> 2. If the base class dtor is not virtual, the use of virtual is most likely a programmer error and hides a bug such as one illustrated in a previous comment, while the use of override is a compile-time error and so is strongly preferable.

geoffthemedio:
sounds more reasonable

Issue #2308, opened 2018-10-26
Fails to build with Boost 1.69 (develop snapshot)


jbeich:
After https://github.com/boostorg/boost/pull/189 build fails, see [error log](https://ptpb.pw/0Qs1) and [missing files](https://reviews.freebsd.org/differential/changeset/?ref=585402)
```
CMake Error at /usr/local/share/cmake/Modules/FindBoost.cmake:2049 (message):
  Unable to find the requested Boost libraries.

  Boost version: 1.69.0

  Boost include path: /usr/local/include

  Could not find the following Boost libraries:

          boost_signals
```
Note, according to [Boost calendar](https://www.boost.org/development/) 1.69 release is planned sometime in December.

o01eg:
Looks like it will affect 0.4.8 as well.

geoffthemedio:
I thought the signals dependency was removed some time ago. Does this just require a change to a CMakeLists.txt ? Perhaps to change a default from signals to signals2 ?

Vezzra:
Has this been fixed sufficiently by #2310? Can we close it as resolved?

geoffthemedio:
Seemingly yes

https://github.com/freeorion/freeorion/pull/2310#issuecomment-433535906

jbeich:
After boostorg/serialization@af46ac2da3e9 (post-Beta1) build fails. See [error log](https://ptpb.pw/mNxv). Including `<string>` just after `<memory>` in `combat/CombatEvent.h` appears to be enough to fix it.

```c++
In file included from combat/CombatEvent.cpp:1:
combat/CombatEvent.h:42:25: error: implicit instantiation of undefined template 'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
    virtual std::string CombatLogDetails(int viewing_empire_id) const
                        ^
/usr/include/c++/v1/iosfwd:193:32: note: template is declared here
    class _LIBCPP_TEMPLATE_VIS basic_string;
                               ^
In file included from combat/CombatEvent.cpp:1:
combat/CombatEvent.h:43:14: error: implicit instantiation of undefined template 'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
    { return std::string(""); }
             ^
/usr/include/c++/v1/iosfwd:193:32: note: template is declared here
    class _LIBCPP_TEMPLATE_VIS basic_string;
                               ^
```

PR #2310, opened 2018-10-26
Remove CMake boost::signals dependencies

geoffthemedio:
Testing what happens in CMake build when removing signals dependencies, given https://github.com/freeorion/freeorion/issues/2308

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -204,7 +204,6 @@ find_package(Boost ${MINIMUM_BOOST_VERSION}
         log_setup
         regex
         serialization
-        signals
         system
         thread
     REQUIRED)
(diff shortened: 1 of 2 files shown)


o01eg:
https://stackoverflow.com/questions/27252213/cmake-can-not-find-boostsignals2

Looks like `signals2` doesn't require component as it is header-only now.

geoffthemedio:
> Looks like signals2 doesn't require component as it is header-only now.

Tried again by just removing all references to signals in CMakeLists.txt files.

jbeich:
[Builds fine](https://ptpb.pw/CDkM) with Boost 1.69.

geoffthemedio:
Would make sense to include in a v0.4.8 bugfix point release...

o01eg:
Is it planned? I suppose some more fixes should be picked to the bugfix point release.

geoffthemedio:
Not yet, but not being able to build against the next boost version might warrant one if various distros start including it by default.

Vezzra:
@geoffthemedio,

> Do you know if these changes are enough for OSX as well?

The latest test builds already contained that change, and built and ran ok on OSX, so I guess we're good...

Vezzra:
> Would make sense to include in a v0.4.8 bugfix point release...

Yep. And you've already tagged it for cherry picking too, excellent...

Issue #2302, opened 2018-10-16
Obliterating an enemy AI prevents to finish turn


Oberlus:
* **FreeOrion Version**: 2018-10-07.feb528b
* **Operating System**: Windows 7
* **Graphic card used**: GTX 660
* **Fetched as**
  * Weekly development build

### Description
On the turn you completely eliminate an enemy AI empire, on hitting next turn game will stay waiting for that dead AI to finish before saving, so game won't continue. Clicking on resign exits without giving you time to click on the "exit without saving".
This happens consistently since at least a week before this weekly dev. build was published. The attached game is a second, different-settings game with latest week dev. build, but it also happened to me on Linux using compiled source from master around 10 days ago.

### Expected Result
Games doesn't wait on dead empires and continues.

### Steps to reproduce
Load attached game and click on next turn until issue happens (should be on turn 143, save is from turn 141 when combat on enemy (last) homeworld is going on).
[FreeOrion_Jugador_Humano_El_0141_20181016_174736.zip](https://github.com/freeorion/freeorion/files/2483923/FreeOrion_Jugador_Humano_El_0141_20181016_174736.zip)

urshmost:
I just experienced this again in build 2018-10-27, in the situation where you invade a species' last planet, on the same turn that they build a last ship there.  
Example:
-turn 98: your fleet arrives at enemy capital planet and destroys their last fleet.
-turn 99: your troop fleet arrives and you click invade, and a last ship by the enemy is built.
-turn 100: the enemy's capital, their last planet, is captured, and then their last ship is destroyed.

The Messages window does not progress past "Waiting for Other Players" and the Empire window shows all players other than the one that you just defeated are finished their turns.  It also shows a 'Head on a Spike" icon for that empire that you just defeated.

This Situation Report shows, in sequence:

The Syndicate has been defeated!
The capital of the Syndicate empire has fallen.
On Royal Cor a I : there was a ground battle with Syndicate.
Royal Cor a I was captured by Terran Imperium.
At Cor a: there was combat with Syndicate
At Cor a: a Griffin was destroyed

Hope that helps!

o01eg:
I suppose defeated empire shouldn't be processed after https://github.com/freeorion/freeorion/pull/2283

Do you have a server log?

urshmost:
[hang on defeated empire.zip](https://github.com/freeorion/freeorion/files/2544176/hang.on.defeated.empire.zip)
[...]

PR #2319, opened 2018-11-13
Fix hang (one possible choice) after an empire is defeated

Dilvish-fo:
This narrowly resolves the hang #2302 (for me at least; I couldn't get the savegame posted in that issue to load for me).  But I hold off from committing this as-is because this still leaves a problem from #2284 (which is the first bad commit with the hang) and either worsens or creates another problem  -- specifically, this fixes the problem of the game hanging when one attempts to progress past a turn in which an AI (or any empire in multiplayer I suppose) is eliminated, however, if in a single player game the human player is eliminated (not merely defeated) then (i) because of the timing involved the human client never even gets a final sitrep saying that they have been eliminated (if they lose their last planet before losing their last ship, then they get a message about being defeated when they lost the planet, but on the turn they lose the last of their joint ships and planets they get no final sitrep saying such), and (ii) then the game just goes into full spin mode, with the AI(s) zipping through turns and the human client not even getting an update.  

Having written this and explored the remaining problems, I think that a better fix is just to revert a portion of #2284 -- the ServerApp parts that refrain from asking for orders from an eliminated client.   That may seem very slightly wasteful, but I have now tested it also and it seems to me to have none of the problems of this fix / partial-fix.  I'll still post this for reference though, in case @o01eg  wants to work on a solution that keeps the code my other solution reverts (and so I include another test file below in case the one from 2302 likewise did not load for o)leg).

[test_defeated_ai_hang.sav.zip](https://github.com/freeorion/freeorion/files/2574844/test_defeated_ai_hang.sav.zip)

Once someone agrees that the other solution is preferable, we can mark this as superseded and close.

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -70,6 +70,15 @@ namespace {
     }
 };
 
+namespace {
+    /** Returns true if \a empire has been eliminated by the applicable
+      * definition of elimination.  As of this writing, elimination means
+      * having no ships and no planets. */
+    bool EmpireEliminated(int empire_id) {
+          return (Objects().FindObjects(OwnedVisitor<Planet>(empire_id)).empty() &&    // no planets
+                  Objects().FindObjects(OwnedVisitor<Ship>(empire_id)).empty());      // no ship
+      }
+}
 
 ////////////////////////////////////////////////
 // ServerApp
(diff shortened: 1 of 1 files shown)


o01eg:
> Having written this and explored the remaining problems, I think that a better fix is just to revert a portion of #2284 -- the ServerApp parts that refrain from asking for orders from an eliminated client.

It was extracted to separate PR https://github.com/freeorion/freeorion/pull/2283

If it will be reverted it better to drop player connection instead entering him to ingame lobby because server will wait for turn orders from those players.

Dilvish-fo:
@o0leg -- I think that you'll probably want to take a look at also doing this, but only so long as there is either a moderator or observer or a surviving human-player-controlled empire.  That way the game won't spin if a single-player human is beaten, but in your multiplayer games you won't be stuck needing the eliminated players to advance turn (or leave their clients running with auto-advance set).

o01eg:
@Dilvish-fo  I suppose all processing ends before decision which empires have to send orders and while empires send orders no one could be eliminates. Only way to do so is ELIMINATE_SELF which directly removes the empire from waiting list. So I don't think it is needed to check empire for elimination while waiting orders.

o01eg:
I think it better to do something like #2283.

Vezzra:
Superseded by #2321.

PR #2320, opened 2018-11-13
Safer Fix for hang stemming from empire defeat

### LINKED ISSUE #2302: Obliterating an enemy AI prevents to finish turn

Dilvish-fo:
I think this is a superior (at least in terms of reliability and completeness) fix for #2302 than the first partial fix (#2319) I just posted, and should probably supersede it (but I posted both in case o0leg would prefer to pursue the approach of the other one).

This one simply reverts a portion of #2284, which portion had been intended to be efficient, but which also led to undesirable consequences.

Resolves #2302

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -3109,21 +3109,13 @@ void ServerApp::PreCombatProcessTurns() {
     m_networking.SendMessageAll(TurnProgressMessage(Message::DOWNLOADING));
 
     // send partial turn updates to all players after orders and movement
-    // exclude those without alive empire and who are not Observer or Moderator
     for (auto player_it = m_networking.established_begin();
          player_it != m_networking.established_end(); ++player_it)
     {
         PlayerConnectionPtr player = *player_it;
-        int empire_id = PlayerEmpireID(player->PlayerID());
-        const Empire* empire = GetEmpire(empire_id);
-        if ((empire && !empire->Eliminated()) ||
-            player->GetClientType() == Networking::CLIENT_TYPE_HUMAN_MODERATOR ||
-            player->GetClientType() == Networking::CLIENT_TYPE_HUMAN_OBSERVER)
-        {
-            bool use_binary_serialization = player->ClientVersionStringMatchesThisServer();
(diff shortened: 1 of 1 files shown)


o01eg:
There I've got idea what I missed. I've disabled `TurnPartialUpdateMessage` due disclosure game information.

I think if allow to send `TurnPartialUpdateMessage` to eliminated empire while don't send to those players without any empire then it should works.

Dilvish-fo:
Note: the Travis build had stalled; I've restarted it now.

> There I've got idea what I missed. I've disabled TurnPartialUpdateMessage due disclosure game information.  I think if allow to send TurnPartialUpdateMessage to eliminated empire while don't send to those players without any empire then it should works.

If these have given you some ideas for a more satisfying fix, then sure put that up.  Although since @Vezzra will likely be putting out a new weekly build either today or tomorrow, and this is a rather nuisancesome bug, I would be inclined to suggest that Vezzra go ahead and merge this fix before he does his build, and then you could follow up with your more satisfying & complex fix.

o01eg:
I've prepared fix but testing it now.

Vezzra:
Superseded by #2321.

PR #2326, opened 2018-12-01
Fix monster shield regeneration (#2325)

### LINKED ISSUE #2325: Monsters do not regenerate shields
We moved meter growth into species effects. For ship shields there is
https://github.com/freeorion/freeorion/blob/master/default/scripting/species/common/shields.macros#L13

As monsters do not have a species, they lack that effect and do not regrow shields after losing them e.g. due to a molecular cloud.

Originally reported by UrshMost in forums:
https://www.freeorion.org/forum/viewtopic.php?f=28&p=94121#p94121

Morlic-fo:
Attempts to fix #2325: After moving the ship shield growth from C++ to a FOCS species effect, monster ships did not regenerate shields as they generally do not have a species.

If we provide a bugfix release, we probably want to include this patch.

### DIFF
--- a/default/scripting/ship_hulls/monster/monster.macros
+++ b/default/scripting/ship_hulls/monster/monster.macros
@@ -1,3 +1,11 @@
+MONSTER_SHIELD_REGENERATION
+'''EffectsGroup  // Increase shields to max when not in battle
+            scope = Source
+            activation = (Source.LastTurnActiveInBattle < CurrentTurn)
+            priority = [[AFTER_ALL_TARGET_MAX_METERS_PRIORITY]]
+            effects = SetShield value = Target.MaxShield
+'''
+
 MONSTER_MOVE_ALWAYS
 '''EffectsGroup
             scope = And [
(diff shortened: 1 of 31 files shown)


default/scripting/ship_hulls/monster/monster.macros
```diff
@@ -1,3 +1,11 @@
+MONSTER_SHIELD_REGENERATION
+'''EffectsGroup  // Increase shields to max when not in battle
+            scope = Source
+            activation = (Source.LastTurnActiveInBattle < CurrentTurn)
```
geoffthemedio, line 4: Should this also test that the ship is unowned to avoid duplicating the effect when an empire owns it?

Morlic-fo: Unowned seems to be the wrong condition.

The critical part is that the ship has no species (as ship shields are regenerated by a species effect). If a player owns it is irrelevant to this. Adding ```Unowned``` condition would actually break a variety of player-owned monsters.

Could check for no species instead, I guess. Or add a stacking group. Or just ignore the possible duplication as it has no noticeable impact.

geoffthemedio:
Looks reasonable

PR #2329, opened 2018-12-02
Add option to enable/disable network binary serialization

o01eg:
This PR adds option `network.server.binary.enabled` same as `save.format.binary.enabled`. Useful for debugging when client and server have a same version but it need to view serialized message or for public server when client could have same version string but on a different CPU architecture.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1915,6 +1915,9 @@ Automatically quits once any turns specified by --auto-advance-n-turns are compl
 OPTIONS_DB_BINARY_SERIALIZATION
 Use Binary serialization for saving games. Binary serialization is faster to save and load, but may not be possible to load on a different operating system.
 
+OPTIONS_DB_SERVER_BINARY_SERIALIZATION
+The server will use Binary serialization for client-server interaction if the client's version matches the server's version. Binary serialization is faster to transfer, but may not be compatible between different operating systems and architectures.
+
 OPTIONS_DB_XML_ZLIB_SERIALIZATION
 When saving games with XML serialization, compress most of the XML before writing the file. Compression substantially reduces save file sizes, but may make saves unloadable due to memory requirements to decompress the save data.
 
--- a/network/ServerNetworking.cpp
+++ b/network/ServerNetworking.cpp
@@ -212,7 +212,7 @@ void PlayerConnection::SetCookie(boost::uuids::uuid cookie)
 const std::string& PlayerConnection::ClientVersionString() const
 { return m_client_version_string; }
 
-bool PlayerConnection::IsBinarySerializationEnabled() const {
+bool PlayerConnection::IsBinarySerializationUsed() const {
     return GetOptionsDB().Get<bool>("network.server.binary.enabled")
         && !m_client_version_string.empty()
         && m_client_version_string == FreeOrionVersionString();
--- a/network/ServerNetworking.cpp
+++ b/network/ServerNetworking.cpp
@@ -212,8 +212,11 @@ void PlayerConnection::SetCookie(boost::uuids::uuid cookie)
 const std::string& PlayerConnection::ClientVersionString() const
 { return m_client_version_string; }
 
-bool PlayerConnection::ClientVersionStringMatchesThisServer() const
-{ return !m_client_version_string.empty() && m_client_version_string == FreeOrionVersionString(); }
+bool PlayerConnection::ClientVersionStringMatchesThisServer() const {
+    return GetOptionsDB().Get<bool>("network.server.binary.enabled")
+        && !m_client_version_string.empty()
+        && m_client_version_string == FreeOrionVersionString();
+}
 
 PlayerConnectionPtr PlayerConnection::NewConnection(boost::asio::io_service& io_service,
                                                     MessageAndConnectionFn nonplayer_message_callback,
--- a/network/ServerNetworking.cpp
+++ b/network/ServerNetworking.cpp
@@ -212,7 +212,7 @@ void PlayerConnection::SetCookie(boost::uuids::uuid cookie)
 const std::string& PlayerConnection::ClientVersionString() const
 { return m_client_version_string; }
 
-bool PlayerConnection::ClientVersionStringMatchesThisServer() const {
+bool PlayerConnection::IsBinarySerializationEnabled() const {
     return GetOptionsDB().Get<bool>("network.server.binary.enabled")
         && !m_client_version_string.empty()
         && m_client_version_string == FreeOrionVersionString();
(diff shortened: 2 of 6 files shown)


default/stringtables/en.txt
```diff
@@ -1915,6 +1915,9 @@ Automatically quits once any turns specified by --auto-advance-n-turns are compl
 OPTIONS_DB_BINARY_SERIALIZATION
 Use Binary serialization for saving games. Binary serialization is faster to save and load, but may not be possible to load on a different operating system.
 
+OPTIONS_DB_SERVER_BINARY_SERIALIZATION
+Use Binary serialization for client-server interaction. Binary serialization is faster to transfer, but may not be possible to read message on a different operating system and architectures.
```
Dilvish-fo: It looks to me like this is purely a server-side option, and this help message should probably indicate that.

network/ServerNetworking.h
```diff
@@ -227,7 +227,7 @@ class PlayerConnection :
     const std::string& ClientVersionString() const;
 
     /** Returns true iff the client's specified version string matches the
-      * version string of this server process. */
+      * version string of this server process and the client could accept binary serialization. */
```
Dilvish-fo: Likewise, since the added option is purely server-side, the change does not seem to have to do with whether the client could accept binary serialization -- it looks to me like the initial messages establishing the connection and sharing the client version number are done via xml serialization, and I don't see that there has been any check yet done at this point to confirm that the client really could exchange compatible binary messages.  Or am I missing/misreading something?

o01eg: I've added some description because this function is used only to enable binary serialization when sending turn update and game start messages.

default/stringtables/en.txt

network/ServerNetworking.h
Dilvish-fo: I am not understanding how that really relates to my concern -- it seems to me like the code comment change here indicates that the function is checking whether "the client could accept binary serialization", but it does not look to me like the code actually checks that in any way.

o01eg: Tried to rephrase it.

default/stringtables/en.txt
```diff
@@ -1915,6 +1915,9 @@ Automatically quits once any turns specified by --auto-advance-n-turns are compl
 OPTIONS_DB_BINARY_SERIALIZATION
 Use Binary serialization for saving games. Binary serialization is faster to save and load, but may not be possible to load on a different operating system.
 
+OPTIONS_DB_SERVER_BINARY_SERIALIZATION
+Allows server to use Binary serialization for client-server interaction. Binary serialization is faster to transfer, but may not be possible to read message on a different operating system and architectures.
```
geoffthemedio: ...but may not be compatible between different operating systems and...

network/ServerNetworking.cpp
```diff
@@ -213,7 +213,11 @@ const std::string& PlayerConnection::ClientVersionString() const
 { return m_client_version_string; }
 
 bool PlayerConnection::ClientVersionStringMatchesThisServer() const
-{ return !m_client_version_string.empty() && m_client_version_string == FreeOrionVersionString(); }
+{
```
geoffthemedio: move `{` to end of previous line

default/stringtables/en.txt
geoffthemedio: This is still misleading, as the game checks the version strings before using binary serialization, and if they don't match, reverts to XML, so when will this be relevant for different OS or versions?

If you want an option to force XML, that's fine, but phrase it as that, and don't claim it's helpful for cases where it isn't (shouldn't?) be...

network/ServerNetworking.h
```diff
@@ -227,7 +227,7 @@ class PlayerConnection :
     const std::string& ClientVersionString() const;
 
     /** Returns true iff the client's specified version string matches the
-      * version string of this server process. */
+      * version string of this server process and the will enable binary serialization. */
```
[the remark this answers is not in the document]
o01eg: rephrased

network/ServerNetworking.cpp

default/stringtables/en.txt
o01eg: Tried to rewrite it.

default/stringtables/en.txt
```diff
@@ -1915,6 +1915,9 @@ Automatically quits once any turns specified by --auto-advance-n-turns are compl
 OPTIONS_DB_BINARY_SERIALIZATION
 Use Binary serialization for saving games. Binary serialization is faster to save and load, but may not be possible to load on a different operating system.
 
+OPTIONS_DB_SERVER_BINARY_SERIALIZATION
+The server will use Binary serialization for client-server interaction if the client's version match the server's version. Binary serialization is faster to transfer, but may not be compatible between different operating systems and architectures.
```
Dilvish-fo: One least tweak needed here, for proper verb form:
"if the client's version match" --> "if the client's version matches"

network/ServerNetworking.cpp
```diff
@@ -212,8 +212,11 @@ void PlayerConnection::SetCookie(boost::uuids::uuid cookie)
 const std::string& PlayerConnection::ClientVersionString() const
 { return m_client_version_string; }
 
-bool PlayerConnection::ClientVersionStringMatchesThisServer() const
-{ return !m_client_version_string.empty() && m_client_version_string == FreeOrionVersionString(); }
+bool PlayerConnection::ClientVersionStringMatchesThisServer() const {
```
geoffthemedio: Need to rename this or move the new logic to where it is used.

o01eg: Is `IsBinarySerializationUsed` correct name?

o01eg: Changed to new name.

network/ServerNetworking.h
```diff
@@ -227,8 +227,9 @@ class PlayerConnection :
     const std::string& ClientVersionString() const;
 
     /** Returns true iff the client's specified version string matches the
-      * version string of this server process. */
-    bool ClientVersionStringMatchesThisServer() const;
+      * version string of this server process so the server will enable binary serialization if it
+      * is allowed. */
```
geoffthemedio: Comment now misleading; should / could just note that it indicates whether the server will use binary serialization, not how it determines this.

network/ServerNetworking.h
```diff
@@ -227,8 +227,9 @@ class PlayerConnection :
...
+    bool IsBinarySerializationEnabled() const;
```
geoffthemedio: name a bit misleading... enabling not the same as using, as it needs to be enabled and the version strings need to match for this to return true.

PR #2330, opened 2018-12-06
Condition Combat Targetting

geoffthemedio:
Work in progress. Pull request to provoke test builds.


agrrr3:
You seem to make great progress :+1: 
I would like to give input in form of a code review, but i have never done that in github and cant find a button for that. Maybe i lack some rights to do that?

I do not quite understand the "// unowned target, when attacker is owned by an empire, and target is visible to that empire" part .. first i thought it is mostly for targetting monsters but it looks too complicated for that and also  Unowned AND (Source.Owner != LocalCandidate.Owner) seem to contradict each other. Could you please clarify?

I know its a matter of style, but i would prefer the generic FOCS part for filtering to visible enemy ships to be a macro so it gets a name and bugfixes do not need to be applied to all hangars and weapons each time maintenance is necessary.

Do you intend to specialise the fighter combatTargets in this branch? I could whip up a suggestion in the form of combatTarget conditions.

geoffthemedio:
>  cant find a button for that

This is how I made a comment on a pull request in the GitHub interface:
github pr commenting

I find the Start a review method confusing, and prefer to just add a series of single comments.

> Could you please clarify?

The complicated targeting condition deals with various cases and limitations of conditions. This way it will target empire ships if the ship is unowned, and enemy empire or unowned ships if the ship is owned by an empire.

> i would prefer the generic FOCS part for filtering to visible enemy ships to be a macro

Sounds reasonable / have thought about doing this.

> Do you intend to specialise the fighter combatTargets in this branch?

Hadn't really thought about it. Feel free to suggest stuff.

geoffthemedio:
@agrrr ?

default/scripting/ship_parts/FighterHangar/FT_HANGAR_3.focs.txt
```diff
@@ -5,6 +5,10 @@ Part
     class = FighterHangar
     capacity = 2
     damage = 5
+    combatTargets = And [
+        [[COMBAT_TARGETS_VISIBLE_ENEMY]]
+        Or [ Ship Fighter ]
```
agrrr3: I think bombers (and heavy bombers) should not target Fighters.
We could add another layer which targets Fighters if there are no Ships if you prefer, but i do not think this is really necessary.
There is an efficient but unprecise FOCS implementation for the target-fighters-iff-no-ships condition:
```
And [
    [[COMBAT_TARGETS_VISIBLE_ENEMY]]
    Or·[
        Ship
        // probability to hit fighter is only high if there are few enemy ships
        NumberOf·number·=·1·condition·=·Fighter
    ]
]
```
Or a precise, more verbose and maybe costly(?) variant which needs to calculate the number of enemy ships:
```
And [
    [[COMBAT_TARGETS_VISIBLE_ENEMY]]
    Or [
        Ship
        And [
// picking a single random fighter should speed up calculation and is ok: it will be the only target as it will be only selected if there are no Ship targets
            NumberOf·number·=·1·condition·=·Fighter
            Number low = 0 high = 0 condition = And [
                Object id = RootCandidate.ID
                [[COMBAT_TARGETS_VISIBLE_ENEMY]] // not sure if this needs to be reevaluated
                Ship
            ]
         ]
     ]
]
```
Caveat: I did not check if the conditions work yet

edit1: actually the second variant did not work so well

geoffthemedio: such tweaks / balance / optimization to parts can be done after the mechanisms to allow them are in place

agrrr3:
I playtested when fiddling with combatTarget conditions. 

I think i hit a bug - my fighters destroyed the only eligible target in the second round. In the third round there was no eligible target anymore .... so my fighters again damaged the destroyed target (?!)

edit1: i first thought that the behavior has changed - it seemed destroyed objects were not eligible for targetting anymore. This is not true

geoffthemedio:
> destroyed objects are not eligible for targetting

What does that mean?

default/scripting/ship_parts/FighterHangar/FT_HANGAR_3.focs.txt
agrrr3: >>    Do you intend to specialise the fighter combatTargets in this branch?
>
> Hadn't really thought about it. Feel free to suggest stuff.

Working on the parts at least checks if the mechanism can fullfill the intentions and if it actually works

geoffthemedio:
I've removed destroyed fighters from combat, and added some nonzero structure conditions for ship targeting (the value for which should be updated only once per combat round, so targeting within a round is still parallel / independent).

Seems to be an issue with ships not losing any fighters after combat, though. They are gone immediately after combat, but are being restored at some point during the post-combat turn processing, so ships always have full fighter complements.

geoffthemedio:
Any other problems / bugs with what's currently implemented?

geoffthemedio:
> I got some strange behavior when trying to use RootCandidate and LocalCandidate in the Number condition

As discussed here:
 
https://github.com/freeorion/freeorion/blob/master/universe/Condition.cpp#L373

Number doesn't test the objects that it is matching, but rather another set of objects, the result of which determines if all or none of the candidate objects of Number will be matched. Thus, `LocalCandidate` and (if not already defined by a higher-level condition) `RootCandidate` aren't defined within Number.

geoffthemedio:
> Didnt have the time to do code review but i did not find another bug.

I'm going to merge this then... If there are code grooming issues, they can be dealt with later.

PR #2333, opened 2018-12-09
Reduce AI invasion overkill/waste

Dilvish-fo:
Currently, when the AI submits an invasion order, it uses all of the troopships in the respective fleet, regardless of whether they are actually currently needed; if fleets had been combined because of a common system destination or if the target planet had lost troopers since the mission was started (due to other invasions, or dropping defense focus, or whatnot) then the AI would sometimes waste a lot of troops.  This PR reduces the maximum overkill (at least in the most common situation) to 2 troopers.   This also adds some TODO's for further related improvement.

### DIFF
--- a/default/python/AI/fleet_orders.py
+++ b/default/python/AI/fleet_orders.py
@@ -460,38 +460,51 @@ class OrderInvade(AIFleetOrder):
         planet = self.target.get_object()
         fleet = self.fleet.get_object()
 
-        invasion_roles = (ShipRoleType.MILITARY_INVASION,
-                          ShipRoleType.BASE_INVASION)
+        invasion_roles = (ShipRoleType.BASE_INVASION,
+                          ShipRoleType.MILITARY_INVASION)
 
         debug("Issuing order: %s fleet: %s target: %s" % (self.ORDER_NAME, self.fleet, self.target))
         # will track if at least one invasion troops successfully deployed
-        result = False
+        result = True
         aistate = get_aistate()
-        for ship_id in fleet.shipIDs:
-            ship = universe.getShip(ship_id)
-            role = aistate.get_ship_role(ship.design.id)
-            if role not in invasion_roles:
-                continue
-
-            debug("Ordering troop ship %d to invade %s" % (ship_id, planet))
-            result = fo.issueInvadeOrder(ship_id, planet_id) or result
+        overkill_margin = 2  # TODO: get from character module; allows a handful of extra troops to be immediately
+        #                            defending planet
+        troops_wanted = planet.currentMeterValue(fo.meterType.troops) + overkill_margin
+        troops_already_assigned = 0  # TODO: get from other fleets in same system
+        troops_assigned = 0
+        # Todo: evaluate all local troop ships (including other fleets) before using any, make sure base invasion troops
+        #       are used first, and that not too many altogether are used (choosing an optimal collection to use).
+        for invasion_role in invasion_roles:  # first checks base troops, then regular
             if not result:
-                shields = planet.currentMeterValue(fo.meterType.shield)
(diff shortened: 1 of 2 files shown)


default/python/AI/fleet_orders.py
```diff
@@ -460,38 +460,51 @@ def issue_order(self):
...
+                        aistate.needsEmergencyExploration.append(fleet.systemID)
+                        debug("Due to trouble invading, adding system %d to Emergency Exploration List" % fleet.systemID)
+                    self.executed = False
+                    # debug(universe.getPlanet(planet_id).dump())  # TODO: fix fo.UniverseObject.dump()
+                    break
+                troops_assigned += ship.troopCapacity
+        # TODO: split off unused troop ships into new fleet and give new orders this cycle
         if result:
-            debug("Successfully ordered troop ship(s) to invade %s" % planet)
+            debug("Successfully ordered %d troop ship(s) to invade %s" % (troops_assigned, planet))
```
Morlic-fo: ```troops_assigned``` tracks the number of troops deployed, not the number of troop ships.

default/python/AI/fleet_orders.py
```diff
@@ -460,38 +460,51 @@ def issue_order(self):
...
                 break
+            for ship_id in fleet.shipIDs:
+                if troops_already_assigned + troops_assigned >= troops_wanted:
+                    break
+                ship = universe.getShip(ship_id)
+                if aistate.get_ship_role(ship.design.id) != invasion_role:
+                    continue
+    
+                debug("Ordering troop ship %d to invade %s" % (ship_id, planet))
+                result = fo.issueInvadeOrder(ship_id, planet_id) or result
```
Morlic-fo: ```result``` starts as True and will never be set to False here (or anywhere else).

Morlic-fo (changes requested):

Morlic-fo:
Functional change looks fine. Never noticed the issue in practice but I haven't paid attention to it either. Does this happen outside of edge cases? I can see how this could happen if the defending troops were reduced after sending the invasion fleets (dropping defense focus, incomplete invasions, change of ownership). Or is there another potential issue with our troop ship allocation?

Should probably also add a ```# TODO actually try and minimize overhead by selecting optimum ships```.

Morlic-fo (approved):

Dilvish-fo:
>  ```# TODO actually try and minimize overhead by selecting optimum ships```

I was considering that to be an aspect of 
```
Todo: evaluate all local troop ships ... make sure ... and that not too many altogether are used
```
but I'll go ahead and add a specific reference

Dilvish-fo:
> Does this happen outside of edge cases? 

Well, it seems I've seen it happen more than just rarely, that the AI would arrive with a lot more troopships than needed for the (primary) target and then would use them all up, sometimes with so many extra troops that they would be fully wasted (more extras than the max troops meter for the planet).  It was particularly notable when there were two invadable planets, and the AI had enough troops to get them both but used them all on a single planet.  

Doing a quick check right now it looks like my theory about fleet mergers may be wrong, because it looks to me like AIFleetMission.check_mergers() should only merge troop fleets if they have the same target planet not just same system.  So that leaves me thinking the causes could be from the planet having defense focus when the troopers_needed calc was made but then focus changes before invasion, or it could be from someone other than that AI having camped the planet and stopped troop regen which the AI had been expecting.  But sometimes the excess troops have been much more than I would expect even from those, there may be a buggy fleet merger or troop calculation going on someplace.

The case of multiple targets being present is something I had meant to add a specific TODO for, that the excess troops could be redeployed for another target in the same system immediately rather than waiting for reassignment next turn.  ~~Ill go ahead and push a commit with just that TODO comment.~~ edit: nevermind, I did already get that in, just a bit more generally -- ```# TODO: split off unused troop ships into new fleet and give new orders this cycle```

PR #2334, opened 2018-12-09
Limit AI periodic resetting of military missions

Dilvish-fo:
In the past, AI military fleets would sometimes get 'stuck' in a deployment that was not properly getting marked as completed; as a partial workaround for that a periodic (every 30 turns) reset of military missions was put into place.   The hope had been that generally the fleets would still get reassigned to their same mission when appropriate (due to proximity), but many times that is not the case and a fleet that was in the middle of an important task will instead get sent off on some other mission (sometimes leaving related unarmed fleets vulnerable).

I think we've made good progress in clearing up those situations where fleet missions would not properly terminate and probably no longer need this automatic periodic reset.  This PR stops that by adding a boolean gate for these resets, which defaults to off, the idea being that we would take a period of time to assess whether this code is truly unneeded now.  (But I could just as easily see us as ready to simply delete this block of code right now.)


Morlic-fo (commented):
No objections. Will see what happens.

PR #2335, opened 2018-12-13
Only select fleets with sufficient fuel for a mission

Morlic-fo:
The AI currently may allocate fleets which have insufficient fuel for a mission, sometimes only with insufficient fuel for return. In particularly, I noticed this with military fleets. 

Such a fleet will refuse to move as the pathfinding actually detects that it has insufficient supply but the mission isn't released. The fleet is subsequently stuck until supply reach of the empire expands or it is released from its mission externally.

The proposed solution in this PR is to only select fleets for a mission that have the fuel requirements as determined by AI pathfinding algorithm. 

This is an incomplete solution as if supply situation changes later on, the fleet may still be stuck. 
There is an existing TODO comment in the move order generation to release the fleet from its mission if it has insufficient fuel. So, that one is still open.


Dilvish-fo (approved):
Sounds & looks good.

PR #2341, opened 2019-01-03
Target research pedia links and definition

Ouaz:
- Add [[metertype METER_TARGET_RESEARCH]] when "effects = SetTargetResearch value" in FOCS files
- Keep [[metertype METER_RESEARCH]] when "effects = SetResearch value" in FOCS files (only one: Interspecies Academy building)
- Click on [[metertype METER_TARGET_RESEARCH]] sends to the Research Pedia Article
- Add paragraph about Target Research in Research Pedia Article

As usual, feel free to reword it if necessary.

Note: I plan to do this for each meter type (i.e. target population, max supply, etc. ) to clearly make the difference in the Pedia (links and article) between the meter itself and the corresponding target/max value , so let me know if it's relevant or not.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -8090,7 +8090,7 @@ ANCIENT_RUINS_DEPLETED_SPECIAL
 Ancient Ruins (Excavated)
 
 ANCIENT_RUINS_DEPLETED_SPECIAL_DESC
-'''Increases [[metertype METER_RESEARCH]] on this planet by 1 per Population when Focus is set to Research.
+'''Increases [[metertype METER_TARGET_RESEARCH]] on this planet by 1 per Population when Focus is set to Research.
 
 This planet holds the ruins of an unknown, advanced ancient race. A valuable discovery has already been made here.'''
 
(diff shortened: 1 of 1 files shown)


default/stringtables/en.txt
```diff
@@ -5437,6 +5437,8 @@ Research
 METER_RESEARCH_VALUE_DESC
 '''The Research (or RP "Research Points") of all planets in the empire is combined to discover new [[encyclopedia ENC_TECH]].
 
+The Target Research is an extrapolated value of the Research output on a planet, making it fluctuate on several turns, according to different bonus/malus caused by some [[encyclopedia ENC_BUILDING_TYPE]]s built on the planet, some [[encyclopedia ENC_TECH]] researched by the empire, some [[encyclopedia ENC_SPECIAL]]s linked to the planet, and other various game effects (e.g. species traits).
```
geoffthemedio: I don't like this description of Target meters. I'd rather say they are the stable output level the planet will approach given the current factors influencing that resource. If buildings, techs, or focus change, then the target level could shift.

Ouaz: OK, thanks for the better wording. Let me know if it's correct:

`The Target Research is the stable Research output level a planet will approach, given the current factors influencing the Research resources. Its value is connected to [[encyclopedia ENC_BUILDING_TYPE]]s built on the planet,  [[encyclopedia ENC_TECH]] researched by the empire,  [[encyclopedia ENC_SPECIAL]]s linked to the planet, and other various game effects (e.g. species traits or planetary focus). If these factors change, then the Target Research value is likely to shift.`

geoffthemedio: The list of things that ends in "and other various" and then an example is awkward. Just include the example in the list, and / or "various other".

Issue #2340, opened 2019-01-02
Building FreeOrion from source is very slow under VS 2017


OlafPettersson:
* **FreeOrion Version**: 0.4.8
* **Operating System**: Windows 10
* **Fetched as** 
  * Compiled from source

### Description
When Building FreeOrion from source under VS 2017, on my system it can take 20 to 25 minutes. Incremental builds with only small changes can also take > 5 min. This slows development enormously.

Are there any reasons why precompiled headers are not configured? On my system, after configuring precompiled headers - in VS this can be done without changing any source files - the build is much quicker.

I believe gcc also supports precompiled headers, though only with changes to the source files: https://gcc.gnu.org/onlinedocs/gcc/Precompiled-Headers.html.

Vezzra:
@OlafPettersson, has this been sufficiently addressed by #2342? Can we close it as resolved?

PR #2342, opened 2019-01-03
enable precompiled headers for FreeOrion VS 2017 compilation

OlafPettersson:
As per issue #2340, compiling FreeOrion under Windows with VS 2017 takes a lot of time, also on incremental updates. 

This PR improves that issue, reducing compilation time by 2/3 or something. 

This PR  forces the usage of precompiled headers in the projects Common,Parsers,FreeOrion,FreeOrionD and FreeOrionCA when using VS 2017, without altering any source files.

I believe in the long run it might be better to proactively use precompiled headers, i.e. by including 
 "StdAfx.h" in all source files. This could allow using precompiled headers with other compilers, though I am no expert on this. 

Note that 3 source files in the parser project are excluded from using precompiled headers, since they define some boost macros before including other headers.

### DIFF
--- a/msvc2017/FreeOrion/StdAfx.h
+++ b/msvc2017/FreeOrion/StdAfx.h
@@ -1,0 +1,104 @@
+#pragma once
+
+// I generously include all external headers I found in the project header files,
+// including std, boost and GG headers.
+
+// https://hownot2code.com/2016/08/16/stdafx-h/
+
+
+#include <bitset>
+#include <cmath>
+#include <deque>
+#include <functional>
+#include <initializer_list>
+#include <iostream>
+#include <map>
+#include <memory>
+#include <queue>
+#include <set>
+#include <string>
+#include <tuple>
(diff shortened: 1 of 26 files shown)


msvc2017/FreeOrion/StdAfx.h
```diff
@@ -0,0 +1,104 @@
+#pragma once
+
+// I generously include all external headers I found in the project header files,
+// including std, boost and GG headers.
```
geoffthemedio: omit the GG code, I think. It is not a stable third party library like boost or std.

OlafPettersson: Thanks for your advice. I updated the PR to treat GiGi as an internal library. That means ommiting it from StdAfx.h, while using precompiled headers to compile it. I restrained from using precompiled headers for the GiGiSDL-Project, as it consists of only a single compilation unit.

PR #2344, opened 2019-01-06
Sit rep sensor upgrades

agrrr3:
In order to help players see that their enemies increased detection level one gets a sitrep message. Forum discussion: [SitRep for enemy sensor tech advances?](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11150)

In order to work, this implementation applies detection tech effects one turn after the tech was researched. 

Notes:    
    * in order to make sitrep work, the activation of detection changes is delayed one turn after the tech gets researched
    * probably needs somewhere a stringtable/encyclopedia change to document this behavior(?)
    * maybe should only tell enemy empires
    * maybe sitrep effect could be
    * agrrr/Sitrep_Sensor_Upgrades_Unshifted has a defunct version without the detection turn shift
      that one would need effects processing to be changed in order to work correctly

### DIFF
--- a/default/scripting/techs/spy/DETECT_1.focs.txt
+++ b/default/scripting/techs/spy/DETECT_1.focs.txt
@@ -14,20 +14,22 @@ Tech
                 OwnedBy empire = Source.Owner
             ]
             activation = And [
-                Not OwnerHasTech name = "SPY_DETECT_2"
-                Not OwnerHasTech name = "SPY_DETECT_3"
-                Not OwnerHasTech name = "SPY_DETECT_4"
-                Not OwnerHasTech name = "SPY_DETECT_5"
+                [[NOT_OWNER_HAS_TECH_EFFECTIVE(SPY_DETECT_2)]]
+                [[NOT_OWNER_HAS_TECH_EFFECTIVE(SPY_DETECT_3)]]
+                [[NOT_OWNER_HAS_TECH_EFFECTIVE(SPY_DETECT_4)]]
+                [[NOT_OWNER_HAS_TECH_EFFECTIVE(SPY_DETECT_5)]]
             ]
             effects = SetDetection value = Value + 50
 
         EffectsGroup
             scope = Source
             activation = And [
-                Not OwnerHasTech name = "SPY_DETECT_2"
-                Not OwnerHasTech name = "SPY_DETECT_3"
-                Not OwnerHasTech name = "SPY_DETECT_4"
-                Not OwnerHasTech name = "SPY_DETECT_5"
+                [[NOT_OWNER_HAS_TECH_EFFECTIVE(SPY_DETECT_2)]]
+                [[NOT_OWNER_HAS_TECH_EFFECTIVE(SPY_DETECT_3)]]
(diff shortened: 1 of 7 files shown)


o01eg:
Will it work if server won't send enemy's research progress?

agrrr3:
Also Answered in the forums; i guess it should work. If it doesn't it won't
break anything but indicates a bug in sending sitreps.
I dont have setup for testing your use case

Am So., 6. Jän. 2019, 16:31 hat O01eg <notifications@github.com>
geschrieben:

o01eg:
I'll try to check this PR on my server's settings tomorrow.

Vezzra:
@agrrr3, commits eec67288ae292d6ca192b7a32515866e06d0629e and c07f8086ea67df7b5499e231604cd00fdf991b7c don't seem to belong to this PR, I guess they slipped in accidentially?

Dilvish-fo:
> * in order to make sitrep work, the activation of detection changes is delayed one turn after the tech gets researched... probably needs somewhere a stringtable/encyclopedia change to document this behavior

I am not a fan of delaying the benefit of a researched tech by a turn just to support a convenience sitrep for the other players.   If we really did take that route, then yes, some very clear Pedia documentation of such would be absolutely needed.

agrrr3:
@Dilvish-fo please raise your concerns in the discussion thread, thank you

@Vezzra no idea why those commits are there - i think i simply need a rebase

agrrr3:
@o01eg did you try the PR in your server environment?

o01eg:
I'm out of my PC. Now it possible to test it with `RULE_SHOW_DETAILED_EMPIRES_DATA` set to off.

agrrr3:
I tried yesterday and disabled in multiplayer settings something like
'extensive information about other players'. Still works :)

Dilvish-fo:
> Dilvish-fo please raise your concerns in the discussion thread, thank you

I wasn't simply raising a design concern -- I was also saying that even if the design decision were accepted, this PR should include clear pedia documentation, which it appears to me it still does not.

agrrr3:
@Dilvish-fo Added pedia info

agrrr3:
> I am not a fan of delaying the benefit of a researched tech by a turn just to support a convenience sitrep for the other players.

I think the convenience sitrep is just a symptom. The "delay" is just relative to how it was before (which can be addressed by reducing research times for techs). There is no absolute delay. I think actually the UI will be better than before: You get an info you have a new tech and then you see its effect. Also the current behaviour confused the hell out of me in some FOCS scripting in the past ("I have the tech researched. Something does not really work. Ok lets add a sitrep for debugging. What is happening? Where is my sitrep?")

I can shorten research times of all techs by one if you think this change unbalances the game (or shift the research message to one turn earlier, which has the same effect). That would restore the turn number the tech gets applied.

> some very clear Pedia documentation of such would be absolutely needed.

I added entries to the pedia to address this. Note that there was no previous documentation how tech research and effect application actually works so I consider the current way informal.

agrrr3:
Superceded by #2570

Issue #2243, opened 2018-08-06
Allow player to disconnect from the server without stopping the game


o01eg:
### Description

Current architecture doesn't support disconnected player:

1. `PlayerInfo` is sent based on connected players. If someone disconnected on turn change he won't be included in players' list even with active empire.
1. `ServerSaveGameDataRequestMessage` requested from connected players and saved on `ClientSaveData` response. If someone disconnected on turn change he won't be included in save.
    * `PlayerSaveHeaderData` also have information about client type of connected player.
1. The server stops game on any player has disconnected.

First two of them require to change client-server interaction. Third could be implemented quickly but depends on previous issues:

- [x] Change player list so it will be empire-based not player-based. Also it should include Observers and Moderators without empires (#2347).
- [x] Change messages with `PlayerStatus` to make them empire-based not player-based (#2352).
- [x] Send `ClientSaveData` with turn orders and remember them until save (#2364).
- [x] There should be an option to manage conditions when the server stops or continue game with disconnected player (#2379).
- [x] Allow to load saved game without all players (#2400).
- Here will be possible to launch 'longturn' server.
- [x] Allow to save games with partial orders and readiness status so players won't be required to have game continuously running to make a turn (#2404). 

Forum thread: https://freeorion.org/forum/viewtopic.php?f=28&t=11080

### Expected Result
Player will be allowed to make turn orders, disconnect and then continue the game without stopping the game for other players.

Dilvish-fo:
Just to explicitly point out, I think that the "manage conditions" part should include that for a single player game,disconnecting ___does___ cause the server to terminate.

dbenage-cx:
> for a single player game,disconnecting does cause the server to terminate.

Might not be always wanted for external server.

Maybe a server startup option pointing to the monitored client(s)?
For a normal single player game, we just change human client to start server with new option.
Multiplayer could then be any number of specific players and/or some dummy "observer" for connection monitoring.

o01eg:
> Just to explicitly point out, I think that the "manage conditions" part should include that for a single player game,disconnecting does cause the server to terminate.

I suppose single-player game doesn't need to continue game on disconnection.

> Might not be always wanted for external server.
> Maybe a server startup option pointing to the monitored client(s)?

What about host player?

dbenage-cx:
My thought is, who should be host with an external/headless server. There may be multiple local clients, or the owner of a server may wish to temporary use a local client.

[...]

PR #2352, opened 2019-01-13
Send status based on empire instead of player

o01eg:
Next step of  #2243

`PlayerStatus` actually mean status of empire if it have orders to execute. This PR changes base of `PlayerStatus` from player to empire.
This allows player to reconnect and make turns so other player will see his empire makes turn even if them don't know his new player id.

### DIFF
--- a/UI/PlayerListWnd.cpp
+++ b/UI/PlayerListWnd.cpp
@@ -597,7 +597,7 @@ void PlayerListWnd::HandleEmpireStatusUpdate(Message::PlayerStatus player_status
     for (auto& row : *m_player_list) {
         if (PlayerRow* player_row = dynamic_cast<PlayerRow*>(row.get())) {
             if (about_empire_id == ALL_EMPIRES) {
-                player_row->SetStatus(player_status);
+                WarnLogger() << "PlayerListWnd::HandleEmpireStatusUpdate got no empire";
             } else if (player_row->EmpireID() == about_empire_id) {
                 player_row->SetStatus(player_status);
                 return;
--- a/UI/PlayerListWnd.cpp
+++ b/UI/PlayerListWnd.cpp
@@ -593,12 +593,12 @@ std::set<int> PlayerListWnd::SelectedPlayerIDs() const {
     return retval;
 }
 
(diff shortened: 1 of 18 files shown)


server/ServerFSM.cpp
```diff
@@ -2808,26 +2811,20 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
     server.ProcessCombats();
     server.PostCombatProcessTurns();
 
-    // update players that other players are now playing their turn
-    for (auto player_it = server.m_networking.established_begin();
-         player_it != server.m_networking.established_end();
-         ++player_it)
+    // update players that other empires are now playing their turn
+    for (const auto& empire : server.Empires())
     {
```
geoffthemedio: move to end of `for` line

geoffthemedio:
sounds reasonable; the statuses are only relevant for empire-controlling players anyway...

o01eg:
I suppose it enough for this PR. It tracks any changes in empire readiness via empire-based `PlayerStatus`.

UI/PlayerListWnd.cpp
```diff
@@ -593,12 +593,12 @@ std::set<int> PlayerListWnd::SelectedPlayerIDs() const {
     return retval;
 }
 
-void PlayerListWnd::HandlePlayerStatusUpdate(Message::PlayerStatus player_status, int about_player_id) {
+void PlayerListWnd::HandleEmpireStatusUpdate(Message::PlayerStatus player_status, int about_empire_id) {
     for (auto& row : *m_player_list) {
         if (PlayerRow* player_row = dynamic_cast<PlayerRow*>(row.get())) {
-            if (about_player_id == Networking::INVALID_PLAYER_ID) {
+            if (about_empire_id == ALL_EMPIRES) {
```
geoffthemedio, line 599: Does this make sense? If a status update comes in for `ALL_EMPIRES`, is that likely to apply to all players, or none, or just the non-empire players...? Maybe just output an error to the log in this case? The logic for empire IDs isn't really the same as for player IDs in this case...

o01eg: Fixed. Changed to warning as it isn't client error to get such value.

geoffthemedio:
No further comments, though haven't tested it.

PR #2356, opened 2019-01-15
Fix SmokeTestHostless on multiple games

o01eg:
`SmokeTestHostless` doesn't start second and other games because it decides the game was already started.
This PR fixes it and start next games.


geoffthemedio:
Is this done now? As far as I can see the relevant tests aren't run online anyway?

o01eg:
It's done. By default on CI `SmokeTestHostless` runs two games. Second game is ignored now on CI because of this error.

PR #2361, opened 2019-01-23
Target Industry pedia links and definition

Ouaz:
Same as #2341 , but this time for the Target Industry meter.

Note: I'll write a macro for each target_meter paragraph -  it's almost the same wording - (Research, Industry, Population, Infrastructure, Happiness), once all of them will be committed (easier to check for the reviewer)


geoffthemedio:
If you're going to write them all pretty much the same, then they don't need separate reviews, I think... I'm not carefully checking all the macro replacements as well.

PR #2364, opened 2019-01-26
Repeal save game data request

### LINKED ISSUE #2275: Server freezing on saving
### Environment
* **FreeOrion Version**: v0.4.8 83518e19f1cd3643dfe04213e289d162552877d1 based on 26f16b0 (v0.4.8 release).
* **Operating System**: Ubuntu 18.04.1 x64
* **Fetched as** 
  * Compiled from source

### Description
Sometimes in game server freezes after "Saving..." messages. Then after freeze ends the game stopped and players moved to lobby.
I suppose it could be related to connection issues with some players. 
While freeze server accept network messages but they are not processed by FSM until freeze ends.

Server logs just after "Saving...": 
[freeoriond.log.2018-09-13T20-25-45.zip](https://github.com/freeorion/freeorion/files/2381036/freeoriond.log.2018-09-13T20-25-45.zip)

I also got core dump after "Saving..." but I have to kill server:

```gdb
(gdb) thread apply all bt

Thread 1 (Thread 0x7fdce9199740 (LWP 1183)):
#0  0x00007fdce5c4cbb7 in epoll_wait (epfd=4, events=events@entry=0x7ffdde32cc30, maxevents=maxevents@entry=128, timeout=-1)
    at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
#1  0x00005556253f579b in boost::asio::detail::epoll_reactor::run (this=0x555626b44870, block=block@entry=true, ops=...)
    at /usr/include/boost/asio/detail/impl/epoll_reactor.ipp:438
#2  0x00005556253d1028 in boost::asio::detail::task_io_service::do_run_one (ec=..., this_thread=..., lock=..., this=0x555626b44530)
    at /usr/include/boost/asio/detail/impl/task_io_service.ipp:356
#3  boost::asio::detail::task_io_service::run_one (ec=..., this=0x555626b44530) at /usr/include/boost/asio/detail/impl/task_io_service.ipp:170
#4  boost::asio::io_service::run_one (this=0x7ffdde32d4c8) at /usr/include/boost/asio/impl/io_service.ipp:72
#5  ServerApp::Run (this=0x7ffdde32d4c0) at ./server/ServerApp.cpp:347
#6  0x000055562525811f in main (argc=<optimized out>, argv=<optimized out>) at ./server/dmain.cpp:86

(gdb) f 5
#5  ServerApp::Run (this=0x7ffdde32d4c0) at ./server/ServerApp.cpp:347
347	            if (m_io_service.run_one())
(gdb) p *m_fsm
(gdb) p *m_fsm
$3 = {<boost::statechart::state_machine<ServerFSM, Idle, std::allocator<void>, boost::statechart::null_exception_translator>> = {<boost::noncopyable_::noncopyable> = {<No data fields>}, _vptr.state_machine = 0x55562571d888 <vtable for ServerFSM+16>, eventQueue_ = empty std::__cxx11::list, 
    deferredEventQueue_ = std::__cxx11::list = {[0] = {px = 0x7fdcd459b960}, [1] = {px = 0x7fdcd4779d10}, [2] = {px = 0x7fdcd422c190}}, 
    currentStates_ = std::__cxx11::list = {[0] = {px = 0x7fdcc83894e0}}, currentStatesEnd_ = {px = 0x1}, pOutermostState_ = 0x7fdcd4592430, 
    isInnermostCommonOuter_ = false, pOutermostUnstableState_ = {px = 0x0}, translator_ = {<No data fields>}, performFullExit_ = true, 
    shallowHistoryMap_ = std::map with 0 elements, deepHistoryMap_ = std::map with 0 elements, pTriggeringEvent_ = 0x0}, 
  m_lobby_data = std::shared_ptr<MultiplayerLobbyData> (empty) = {get() = 0x0}, 
  m_single_player_setup_data = std::shared_ptr<SinglePlayerSetupData> (empty) = {get() = 0x0}, 
  m_player_save_game_data = std::vector of length 0, capacity 5, m_server_save_game_data = std::shared_ptr<ServerSaveGameData> (empty) = {get() = 0x0}, 
  m_server = @0x7ffdde32d4c0}
```

### Expected Result
Disconnect broken connection if there some. Process network messages without long pauses.

### Steps to reproduce
* I've seen this first so not sure but I got it with multiplayer 3 player + 2 AI.

o01eg:
Another step of #2243. Fixes #2275.

Currently the server requests turn orders and save data from the client after get turn orders before from the client which requires the client to be online when server decide to save game.
After this PR server gets turn orders and save data together from the client so it became possible to client to disconnect after sending turn orders.

To compensate lacking server-side request the client sends partial order changes on some event.

Also allows to send UI data if client connected to playing game.

Breaks client-server compatibility!

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -6445,6 +6445,8 @@ void MapWnd::HideSidePanel() {
 void MapWnd::RestoreSidePanel() {
     if (m_sidepanel_open_before_showing_other)
         ReselectLastSystem();
+    // send order changes could be made in research, production or other windows
+    HumanClientApp::GetApp()->SendPartialOrders();
 }
 
 void MapWnd::ShowResearch() {
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -5597,7 +5597,7 @@ std::vector<int> MapWnd::FleetIDsOfFleetButtonsOverlapping(int fleet_id) const {
         int vis_turn = -1;
         if (vis_turn_map.find(VIS_BASIC_VISIBILITY) != vis_turn_map.end())
             vis_turn = vis_turn_map[VIS_BASIC_VISIBILITY];
-        ErrorLogger() << "Couldn't find a FleetButton for fleet " << fleet_id 
+        ErrorLogger() << "Couldn't find a FleetButton for fleet " << fleet_id
                       << " with last basic vis turn " << vis_turn;
         return fleet_ids;
     }
@@ -6137,7 +6137,12 @@ bool MapWnd::EndTurn() {
     if (m_ready_turn) {
         HumanClientApp::GetApp()->UnreadyTurn();
     } else {
-        HumanClientApp::GetApp()->StartTurn();
+        ClientUI* cui = ClientUI::GetClientUI();
+        if (!cui)
+            return false;
+        SaveGameUIData ui_data;
+        cui->GetSaveGameUIData(ui_data);
+        HumanClientApp::GetApp()->StartTurn(ui_data);
     }
     return true;
 }
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -6138,8 +6138,10 @@ bool MapWnd::EndTurn() {
         HumanClientApp::GetApp()->UnreadyTurn();
     } else {
         ClientUI* cui = ClientUI::GetClientUI();
-        if (!cui)
+        if (!cui) {
+            ErrorLogger() << "MapWnd::EndTurn: No client UI available";
             return false;
+        }
         SaveGameUIData ui_data;
         cui->GetSaveGameUIData(ui_data);
         HumanClientApp::GetApp()->StartTurn(ui_data);
--- a/client/ClientApp.cpp
+++ b/client/ClientApp.cpp
@@ -123,7 +123,10 @@ void ClientApp::StartTurn(const std::string& save_state_string)
 { m_networking->SendMessage(TurnOrdersMessage(m_orders, save_state_string)); }
 
 void ClientApp::SendPartialOrders() {
-    m_networking->SendMessage(TurnPartialOrdersMessage(m_orders.ExtractChanges()));
+    auto changes = m_orders.ExtractChanges();
+    if (changes.first.empty() && changes.second.empty())
+        return;
+    m_networking->SendMessage(TurnPartialOrdersMessage(changes));
 }
 
 void ClientApp::HandleTurnPhaseUpdate(Message::TurnProgressPhase phase_id) {
--- a/client/ClientApp.cpp
+++ b/client/ClientApp.cpp
@@ -116,8 +116,13 @@ void ClientApp::SetEmpireStatus(int empire_id, Message::PlayerStatus status) {
     m_empire_status[empire_id] = status;
 }
 
-void ClientApp::StartTurn()
-{ m_networking->SendMessage(TurnOrdersMessage(m_orders)); }
+void ClientApp::StartTurn() {
+    if (m_ui_data) {
+        m_networking->SendMessage(TurnOrdersMessage(m_orders, *m_ui_data));
+    } else {
+        m_networking->SendMessage(TurnOrdersMessage(m_orders, m_ai_data));
+    }
+}
 
 void ClientApp::HandleTurnPhaseUpdate(Message::TurnProgressPhase phase_id) {
     switch (phase_id) {
--- a/client/ClientApp.cpp
+++ b/client/ClientApp.cpp
@@ -122,6 +122,10 @@ void ClientApp::StartTurn(const SaveGameUIData& ui_data)
 void ClientApp::StartTurn(const std::string& save_state_string)
 { m_networking->SendMessage(TurnOrdersMessage(m_orders, save_state_string)); }
 
+void ClientApp::SendPartialOrders() {
+    m_networking->SendMessage(TurnPartialOrdersMessage(m_orders.ExtractChanges()));
+}
+
 void ClientApp::HandleTurnPhaseUpdate(Message::TurnProgressPhase phase_id) {
     switch (phase_id) {
     case Message::WAITING_FOR_PLAYERS:
--- a/client/ClientApp.cpp
+++ b/client/ClientApp.cpp
@@ -116,13 +116,11 @@ void ClientApp::SetEmpireStatus(int empire_id, Message::PlayerStatus status) {
     m_empire_status[empire_id] = status;
 }
 
-void ClientApp::StartTurn() {
-    if (m_ui_data) {
-        m_networking->SendMessage(TurnOrdersMessage(m_orders, *m_ui_data));
-    } else {
-        m_networking->SendMessage(TurnOrdersMessage(m_orders, m_ai_data));
-    }
-}
+void ClientApp::StartTurn(const SaveGameUIData &ui_data)
+{ m_networking->SendMessage(TurnOrdersMessage(m_orders, ui_data)); }
+
+void ClientApp::StartTurn(const std::string& save_state_string)
+{ m_networking->SendMessage(TurnOrdersMessage(m_orders, save_state_string)); }
 
 void ClientApp::HandleTurnPhaseUpdate(Message::TurnProgressPhase phase_id) {
     switch (phase_id) {
--- a/client/ClientApp.cpp
+++ b/client/ClientApp.cpp
@@ -116,7 +116,7 @@ void ClientApp::SetEmpireStatus(int empire_id, Message::PlayerStatus status) {
     m_empire_status[empire_id] = status;
 }
 
-void ClientApp::StartTurn(const SaveGameUIData &ui_data)
+void ClientApp::StartTurn(const SaveGameUIData& ui_data)
 { m_networking->SendMessage(TurnOrdersMessage(m_orders, ui_data)); }
 
 void ClientApp::StartTurn(const std::string& save_state_string)
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -908,7 +908,7 @@ void HumanClientApp::Reinitialize() {
 float HumanClientApp::GLVersion() const
 { return GetGLVersion(); }
 
-void HumanClientApp::StartTurn(const SaveGameUIData &ui_data) {
+void HumanClientApp::StartTurn(const SaveGameUIData& ui_data) {
     DebugLogger() << "HumanClientApp::StartTurn";
 
     if (const Empire* empire = GetEmpire(EmpireID())) {
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -908,7 +908,7 @@ void HumanClientApp::Reinitialize() {
 float HumanClientApp::GLVersion() const
 { return GetGLVersion(); }
 
-void HumanClientApp::StartTurn() {
+void HumanClientApp::StartTurn(const SaveGameUIData &ui_data) {
     DebugLogger() << "HumanClientApp::StartTurn";
 
     if (const Empire* empire = GetEmpire(EmpireID())) {
@@ -930,7 +930,7 @@ void HumanClientApp::StartTurn() {
         Autosave();
     }
 
-    ClientApp::StartTurn();
+    ClientApp::StartTurn(ui_data);
     m_fsm->process_event(TurnEnded());
 }
 
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -975,7 +975,6 @@ void HumanClientApp::HandleMessage(Message& msg) {
     case Message::JOIN_GAME:                m_fsm->process_event(JoinGame(msg));                break;
     case Message::HOST_ID:                  m_fsm->process_event(HostID(msg));                  break;
     case Message::LOBBY_UPDATE:             m_fsm->process_event(LobbyUpdate(msg));             break;
-    case Message::SAVE_GAME_DATA_REQUEST:   m_fsm->process_event(SaveGameDataRequest(msg));     break;
     case Message::SAVE_GAME_COMPLETE:       m_fsm->process_event(SaveGameComplete(msg));        break;
     case Message::CHECKSUM:                 m_fsm->process_event(CheckSum(msg));                break;
     case Message::GAME_START:               m_fsm->process_event(GameStart(msg));               break;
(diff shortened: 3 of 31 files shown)


server/ServerApp.h
```diff
@@ -292,6 +284,14 @@ class ServerApp : public IApp {
       * between two empires. Updates those empires of the change. */
     void    HandleDiplomaticMessageChange(int empire1_id, int empire2_id);
 
+    /**  Adds an existing empire to turn processing. The position the empire is
```
geoffthemedio, line 291: why...?

o01eg: It is never used outside `ServerApp` and I decide to move it to private.

client/ClientApp.cpp
```diff
@@ -116,8 +116,11 @@ void ClientApp::SetEmpireStatus(int empire_id, Message::PlayerStatus status) {
     m_empire_status[empire_id] = status;
 }
 
-void ClientApp::StartTurn()
-{ m_networking->SendMessage(TurnOrdersMessage(m_orders)); }
+void ClientApp::StartTurn(const SaveGameUIData &ui_data)
```
geoffthemedio: space after & not before

client/human/HumanClientApp.cpp
```diff
@@ -906,7 +906,7 @@ void HumanClientApp::Reinitialize() {
 float HumanClientApp::GLVersion() const
 { return GetGLVersion(); }
 
-void HumanClientApp::StartTurn() {
+void HumanClientApp::StartTurn(const SaveGameUIData &ui_data) {
```
geoffthemedio: space after & not before

client/human/HumanClientApp.h
```diff
@@ -52,7 +52,7 @@ class HumanClientApp :
 
     /** \name Mutators */ //@{
     void HandleTurnUpdate() override;           ///< Handle background events that need starting when the turn updates
-    void StartTurn() override;
+    void StartTurn(const SaveGameUIData &ui_data) override;
```
geoffthemedio: space after & not before

network/Message.h
```diff
@@ -72,7 +72,7 @@ class FO_COMMON_API Message {
         LOBBY_EXIT,             ///< sent to server by clients when a player leaves the multiplayer lobby, or by server to clients when a player leaves the multiplayer lobby
         START_MP_GAME,          ///< sent to server (by the "host" client only) when the settings in the MP lobby are satisfactory and it is time to start the game
         SAVE_GAME_INITIATE,     ///< sent to server (by the "host" client only) when a game is to be saved
-        SAVE_GAME_DATA_REQUEST, ///< sent to clients by the server when the game is being saved and the server needs save state info from clients
+        SAVE_GAME_DATA_REQUEST, ///< sent to clients by the server when the game is being saved and the server needs save state info from clients (removed)
```
geoffthemedio: why not actually removed?

geoffthemedio: Also, isn't this still useful when doing a manual save in multiplayer?

network/Message.h
```diff
@@ -407,18 +396,20 @@ FO_COMMON_API void ExtractJoinGameMessageData(const Message& msg, std::string& p
...
+                                                std::string& save_state_string);
 
 FO_COMMON_API void ExtractTurnUpdateMessageData(const Message& msg, int empire_id, int& current_turn, EmpireManager& empires,
                                                 Universe& universe, SpeciesManager& species, CombatLogManager& combat_logs,
                                                 SupplyManager& supply, std::map<int, PlayerInfo>& players);
 
 FO_COMMON_API void ExtractTurnPartialUpdateMessageData(const Message& msg, int empire_id, Universe& universe);
 
-FO_COMMON_API void ExtractClientSaveDataMessageData(const Message& msg, OrderSet& orders,
+/*FO_COMMON_API void ExtractClientSaveDataMessageData(const Message& msg, OrderSet& orders,
```
geoffthemedio: remove rather than comment out

network/ServerNetworking.cpp
```diff
@@ -251,7 +251,7 @@ namespace {
         case Message::TURN_ORDERS:              return "Turn Orders";
         case Message::TURN_PROGRESS:            return "Turn Progress";
         case Message::PLAYER_STATUS:            return "Player Status";
-        case Message::CLIENT_SAVE_DATA:         return "Client Save Data";
+        //case Message::CLIENT_SAVE_DATA:       return "Client Save Data";
```
geoffthemedio: ...

server/SaveLoad.cpp
```diff
@@ -59,6 +59,8 @@ namespace {
         preview.number_of_empires = empire_save_game_data.size();
         preview.save_time = boost::posix_time::to_iso_extended_string(boost::posix_time::second_clock::local_time());
 
+        DebugLogger() << "player_save_game_data " << player_save_game_data.size();
```
geoffthemedio: add "size: " to the string

server/ServerApp.cpp
```diff
@@ -1288,11 +1292,15 @@ void ServerApp::LoadGameInit(const std::vector<PlayerSaveGameData>& player_save_
...
         m_player_empire_ids[player_id] = empire_id;
+    }
 
-
+    for (const auto& psgd : player_save_game_data) {
+        int empire_id = psgd.m_empire_id;
         // add empires to turn processing, and restore saved orders and UI data or save state data
-        if (GetEmpire(empire_id)) {
-            AddEmpireTurn(empire_id);
+        Empire* empire = GetEmpire(empire_id);
```
geoffthemedio: can put this line into the `if ( )` directly

server/ServerApp.cpp
```diff
@@ -1763,7 +1771,8 @@ int ServerApp::AddPlayerIntoGame(const PlayerConnectionPtr& player_connection) {
     m_player_empire_ids[player_connection->PlayerID()] = empire_id;
 
     const OrderSet dummy;
-    const OrderSet& orders = orders_it->second.second ? *orders_it->second.second : dummy;
+    const OrderSet& orders = static_cast<bool>(orders_it->second.second) && static_cast<bool>(orders_it->second.second->m_orders) ? *(orders_it->second.second->m_orders) : dummy;
```
geoffthemedio: why the extra explicit casting? should be implicitly if in a boolean test...

server/ServerApp.cpp
```diff
@@ -1799,6 +1808,18 @@ bool ServerApp::IsHostless() const
 const boost::circular_buffer<ChatHistoryEntity>& ServerApp::GetChatHistory() const
 { return m_chat_history; }
 
+std::vector<PlayerSaveGameData> ServerApp::GetPlayerSaveGameData() const {
+    std::vector<PlayerSaveGameData> player_save_game_data;
+    for (const auto& m_save_data : m_turn_sequence) {
+        DebugLogger() << "Empire " << m_save_data.first << " ready " << m_save_data.second.first
+            << " save_game_data " << m_save_data.second.second.get();
```
geoffthemedio: indent to match first `<<` on preceeding line

server/ServerFSM.cpp
```diff
@@ -2560,7 +2567,9 @@ sc::result WaitingForTurnEnd::react(const TurnOrders& msg) {
 
         TraceLogger(FSM) << "WaitingForTurnEnd.TurnOrders : Received orders from player " << player_id;
 
-        server.SetEmpireTurnOrders(empire_id, std::move(order_set));
+        server.SetEmpireSaveGameData(empire_id, boost::make_unique<PlayerSaveGameData>(sender->PlayerName(), empire_id,
+                                                                   order_set,       ui_data,    save_state_string,
```
geoffthemedio: dedent to line up with `empire_id`

server/ServerFSM.cpp
```diff
@@ -2677,125 +2686,29 @@ sc::result WaitingForTurnEndIdle::react(const SaveGameRequest& msg) {
         return discard_event();
     }
 
-    context<WaitingForTurnEnd>().m_save_filename = message.Text();  // store requested save file name in Base state context so that sibling state can retreive it
-
-    return transit<WaitingForSaveData>();
-}
+    const std::string& save_filename = message.Text(); // store requested save file name in Base state context so that sibling state can retreive it
```
geoffthemedio: I think this returns by value, so you shouldn't store it in a reference

server/ServerFSM.cpp
```diff
@@ -2677,125 +2686,29 @@ sc::result WaitingForTurnEndIdle::react(const SaveGameRequest& msg) {
...
-        const auto& existing_orders = server.m_turn_sequence[empire->EmpireID()].second;
-        if (existing_orders)
-            order_set.reset(new OrderSet(*existing_orders));
-        else
-            order_set.reset(new OrderSet(received_orders));
-    } else {
-        ErrorLogger(FSM) << "WaitingForSaveData::react(const ClientSaveData& msg) couldn't get empire for player " << player_id;
-        order_set.reset(new OrderSet(received_orders));
+    } catch (const std::exception& error) {
+        ErrorLogger(FSM) << "Catch std::exception: " << error.what();
```
geoffthemedio: "While saving, caught std::exception: "

server/ServerApp.h
geoffthemedio: OK

geoffthemedio:
Basic idea of including save game data in turn orders messages is reasonable. Dunno about removing the message type entirely, though...

network/Message.h
o01eg: If you remove it it will shift numbers of other messages. If it is OK I could remove it.
It should work without server-side request in multiplayer too else it cause #2275

geoffthemedio: I don't think the numbers need to be consistent if the change breaks compatibility anyway.

I'm not saying that there should be a separate save data request after sending turn orders in multiplayer. I'm saying that it should still be possible to manually initiate a save in multiplayer in the middle of a turn, which presumably will require the server to request save state info from the human players.

o01eg: Good. Wonder if it possible to reset boost serialization and cleanup deprecated code.

Server stores save state between turn in `ServerApp::m_turn_sequence` and updates it on player sends turn orders so it doesn't need to request it by itself.

server/ServerApp.cpp

server/ServerApp.cpp

server/ServerApp.cpp

server/ServerFSM.cpp

server/ServerFSM.cpp

server/ServerFSM.cpp

network/ServerNetworking.cpp

network/Message.h
geoffthemedio: *in the middle of a turn* ... ie. after some orders have been issued, but before the turn is ended by all players.

o01eg: The game has an `UNREADY` message which simply marks empire unready and doesn't clear orders set sent before so them could be saved. May be better to split readiness and committing orders?

geoffthemedio: That doesn't seem relevant. The point is what happens if a save is initiated while a player is still playing their turn. It should save the orders they have issued so far at that time. It's irrelevant what happens if they are "ready" or have become unready again after being ready.

network/Message.cpp
```diff
@@ -942,15 +916,29 @@ void ExtractJoinAckMessageData(const Message& msg, int& player_id,
     }
 }
 
-void ExtractTurnOrdersMessageData(const Message& msg, OrderSet& orders) {
+void ExtractTurnOrdersMessageData(const Message& msg, OrderSet& orders, bool& ui_data_available,
+                                      SaveGameUIData& ui_data, bool& save_state_string_available,
```
geoffthemedio: dedent to align with `const`

geoffthemedio: should probably break the parameters into one-per-line, or move the first down to the next line, as it goes too far to the right to read easily now

network/Message.cpp
```diff
@@ -942,15 +916,29 @@ void ExtractJoinAckMessageData(const Message& msg, int& player_id,
...
+                                      std::string& save_state_string) {
```
geoffthemedio: `{` to newline

geoffthemedio: `{` to newline

geoffthemedio:
If I start a single-player game, issue a few orders, save the game, resign, then load that save, none of my issued orders are retained. They should be.

UI/MapWnd.cpp
```diff
@@ -6137,7 +6137,12 @@ bool MapWnd::EndTurn() {
     if (m_ready_turn) {
         HumanClientApp::GetApp()->UnreadyTurn();
     } else {
-        HumanClientApp::GetApp()->StartTurn();
+        ClientUI* cui = ClientUI::GetClientUI();
+        if (!cui)
```
geoffthemedio: should log an error in this case...

client/human/HumanClientApp.cpp
```diff
@@ -908,7 +908,7 @@ void HumanClientApp::Reinitialize() {
 float HumanClientApp::GLVersion() const
 { return GetGLVersion(); }
 
-void HumanClientApp::StartTurn() {
+void HumanClientApp::StartTurn(const SaveGameUIData& ui_data) {
```
geoffthemedio, line 911: is this parameter necessary, or can the relevant data be gotten within this function?

network/Message.h
o01eg: It looks like turn order design issue so I've opened thread: https://freeorion.org/forum/viewtopic.php?f=6&t=11198

o01eg:
Yep, I think there should be explicit way for players to inform the server they have orders they want to be saved but don't want to advance turn.

geoffthemedio:
I don't think players should be making a distinction between issuing an order and telling the server to save it. The client should send a message for each order, and orders should be numbered so the server knows it didn't miss one if it receives the next.

client/human/HumanClientApp.cpp
o01eg: Yes, `HumanClientApp::StartTurn` doesn't send TurnOrder itself, it calls `ClientApp::Start` where it required to have `SaveGameUIData`. But it also overrides `ClientApp::Start` so them have to have same signature.

o01eg:
It suppose make change in `OrderSet` to send each order to server on each `IssueOrder` or `RescindOrder` so I decided to store in `OrderSet` changed ids and then sometimes (I'm still thinking where it better be placed) send pack of them.

o01eg:
@geoffthemedio Is it correct and desirable that `OrderSet::erase` doesn't do "undo" order as `OrderSet::RescindOrder` https://github.com/freeorion/freeorion/commit/b3e58c0627a096227b5310a183e89b0960e63b4b ?

As I see in python code only `size` is used for `OrderSet` object.

UI/MapWnd.cpp
o01eg: Added log here.

network/Message.cpp
o01eg: Fixed.

network/Message.cpp
o01eg: Made it one-per-line.

geoffthemedio:
> Is it correct and desirable that OrderSet::erase doesn't do "undo" order as OrderSet::RescindOrder b3e58c0 ?

I don't see the relevance of that commit... Not sure what you're asking. Erasing and undoing an order are different things, though...

o01eg:
Rebased. Started to implement partial orders updates.

o01eg:
Added sending partial orders when player press "Save" button. Should fix mentioned issue.

o01eg:
Added sending partial orders on closing fleet window, switching planets, closing research and production windows.

o01eg:
No, I suppose it finished. I set sending changes when player "lost focus" on windows I remember. Even if I forgot something it still sends entire turn orders on end turn to prevent desyncs.

PR #2368, opened 2019-01-29
Population meters pedia links and definition

Ouaz:
See commits for description.

One question though: I see that sometimes a key alone is used to replace a full word.
For example:
```
# %1% name of the species.
TT_SPECIES_POPULATION
%1% [[METER_POPULATION]]
```
I hesitated to do the same thing (using the key without the pedia link, as it's not necessary if the pedia link is already in the same paragraph).

Ex:

> A green number following the suitability data indicates the maximum **[[metertype METER_POPULATION]]** the planet could achieve if colonized by this species, whereas a red number indicates that the **Population** will stay idle or will decrease.

The second `Population` refers also to the meter, but no need to use again the pedia link. Is it better to replace it with `[[METER_POPULATION]]` or to keep the full written word?

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -5429,7 +5429,12 @@ METER_POPULATION_VALUE_LABEL
 Population
 
 METER_POPULATION_VALUE_DESC
-An abstract value for the size of a colony, relative to differing species.
+'''The Population represents the size of a colony on a planet. Its maximum value depends on the [[encyclopedia ENC_SPECIES]] and how suitable the planet's [[encyclopedia ENVIRONMENT_TITLE]] is for that species. Many [[metertype METER_INDUSTRY]] and [[metertype METER_RESEARCH]] output bonuses are linked to the Population size: larger Population gives more bonuses.
+
+A Population of at least three and a minimum [[metertype METER_HAPPINESS]] of 5 is necessary for a colony to act as a source of settlers for colonizing other planets. A colony ship has a population of Colonists, which is determined by the [[encyclopedia PC_COLONY]] capacity of the ship design. When colonizing a planet, the number of Colonists aboard the ship will be the starting Population on the newly settled planet.
+
+The Target Population is the stable Population size a planet will approach, given the current factors influencing the planetary Population. Its value is connected to [[encyclopedia ENC_BUILDING_TYPE]]s built on the planet, [[encyclopedia ENC_TECH]] researched by the empire, [[encyclopedia ENC_SPECIAL]]s linked to the planet, and various other game effects. If these factors change, then the Target Population value is likely to shift.
+'''
 
 METER_INDUSTRY_VALUE_LABEL
 Industry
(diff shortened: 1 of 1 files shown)


default/stringtables/en.txt
```diff
@@ -5429,7 +5429,10 @@ METER_POPULATION_VALUE_LABEL
 Population
 
 METER_POPULATION_VALUE_DESC
-An abstract value for the size of a colony, relative to differing species.
+'''An abstract value for the size of a colony on a planet, relative to differing species. The Population on [[encyclopedia PC_COLONY]] ships is referred to as Colonists. When colonizing a planet, the number of Colonists aboard the ship will be the starting Population on the newly settled planet.
```
geoffthemedio: meaning of abstract and relative are unclear

geoffthemedio: I'd probably put some information about what population does in the first paragraph, and move the info about colony ships later.  eg. many bonuses to resource output depend on population, and

default/stringtables/en.txt
```diff
@@ -5681,7 +5684,7 @@ Each [[encyclopedia ENC_SPECIES]] has its own environmental preferences, which d
 
 [[PE_UNINHABITABLE]] < [[PE_HOSTILE]] < [[PE_POOR]] < [[PE_ADEQUATE]] < [[PE_GOOD]]
 
-The planet suitability report can be displayed by right-cliking on a planet in the system sidepanel. A green number following the suitability data indicates the max population value if the species colonizes the planet, whereas a red number indicates that the population will stay idle or will decrease, to finally perish if the species attempts to colonize the planet.
+The planet suitability report can be displayed by right-cliking on a planet in the system sidepanel. A green number following the suitability data indicates the maximum [[metertype METER_POPULATION]] the planet could achieve if colonized by this species, whereas a red number indicates that the Population will stay idle or will decrease, to finally perish if the species attempts to colonize the planet.
```
geoffthemedio: that the colony will be unviable, and population will decrease over several turns until the the colonists have all died off and the colony is lost.

default/stringtables/en.txt
```diff
@@ -9066,7 +9069,7 @@ Colonization
 PC_COLONY_DESC
 '''Colony parts integrate with a ship to provide the necessary equipment to establish ownership on a new planet.
 
-These parts may also increase the [[metertype METER_POPULATION]] on a ship.
+These parts may also increase the [[metertype METER_POPULATION]] on a ship (referred to as Colonists).
```
geoffthemedio, line 9074: are they referred to as colonists much outside of these edits?

default/stringtables/en.txt
Ouaz: Tell me if it's better:

`'''The Population represents the size of a colony on a planet. Its base value depends of the [[encyclopedia ENC_SPECIES]] inherent characteristics and the planet's [[encyclopedia ENVIRONMENT_TITLE]] suitability for a given species. Many [[metertype METER_INDUSTRY]] and [[metertype METER_RESEARCH]] output bonuses are linked to the Population size: bigger is the number, greater are the bonuses.`

`A Population of at least three and a minimum [[metertype METER_HAPPINESS]] of 5 is necessary for a colony to act as a source of settlers for colonizing other planets.  When producing a [[encyclopedia PC_COLONY]] ship, it has its own Population referred to as Colonists, and its value is fixed according the Colonization capacity of the ship design.  When colonizing a planet, the number of Colonists aboard the ship will be the starting Population on the newly settled planet.'''`

default/stringtables/en.txt
Ouaz: Mainly in the Fleet Window panel: 
colonists

It's to make clear for new players, that Colonists are the Population value for a ship, and works in the same way.

Ouaz:
> It's debatable... For isolated words, it's probably useful to do the substitution, so that it only needs to be translated once and can then appear in various contexts. For whole paragraphs, it's probably OK either way... It just the single word is translated, then that doens't help a non-english speaker much.

OK, I leave it for the moment in my to-do list. I just thought it would make the stringtables a little tedious to read, with all these in-text keys.^^

default/stringtables/en.txt
Ouaz: Ok, I will modify the last sentence.

default/stringtables/en.txt
geoffthemedio: What is the "base value"...? The initial population after colonization, or the target value?

...`depends on the species and how suitable the planet's environment is for that species.`

`: larger population gives more bonuses.`

`A colony ship has a population of colonists, which is determined by the colonzation capacity of the ship design.`

default/stringtables/en.txt
geoffthemedio: ...?

geoffthemedio:
> I just thought it would make the stringtables a little tedious to read, with all these in-text keys.

In that case, you can leave the single words in big blocks of text as just text and not use a macro...

Also, @Ouaz bump

Ouaz:
> Also, @Ouaz bump

I force-pushed a commit adding your suggestions 9 days ago, it seems you didn't receive a notification?

geoffthemedio:
Didn't see the extra commit

Issue #2355, opened 2019-01-15
Improvement - Make SitRep effects on Tech research turn possible


agrrr3:
As well broadly discussed in [TurnTechResearched not working the first turn](https://www.freeorion.org/forum/viewtopic.php?f=9&t=11158&) it is not possible to send a custom sitrep in the same turn a technology gets researched (e.g. suitable planets for gas giant generators) due to the way the server does effect evaluation.

Gathering my current understanding I propose the following change:

~~- Introduce priority AFTER_RESEARCH_PRIORITY higher than anything we have currently~~
~~- Limit effect evaluation before processing research to effects with priority value lower than AFTER_RESEARCH_PRIORITY (adding priority min/max parameters to effect evaluation functions)~~
~~- Limit effect evaluation after processing research to effects with priority value higher than AFTER_RESEARCH_PRIORITY (adding priority min/max parameters to effect evaluation functions)~~
~~- Add extra tech initialization effects (currently i know only of sitreps for gas giant generators and  #2344 )~~

~~This is a low-risk approach moving effect evaluation in the direction of a priority based system.~~

Priorities and effects is a bit complicated so for the moment I rather suggest the approach implemented in #2370 : #2370 does not fiddle with priorities but removes the dubious first tech effect application so in effect tech does not behave differently for effect application between turns and is available consistently to FOCS coders.

I think the mentioned thread on the forum is ok for design discussion.

agrrr3:
I started working on this and i am not sure the way is the right one. 
As one can see in Effect::EffectsGroup::Execute(..), the code has quite some history. 
```
Execute(const` TargetsCauses& targets_causes,
                    AccountingMap* accounting_map = nullptr,
                    bool only_meter_effects = false,
                    bool only_appearance_effects = false,
                    bool include_empire_meter_effects = false,
                    bool only_generate_sitrep_effects = false) const;
```
I guess the code was once concise and clear and then a lot necessities were discovered. So the effects' application is cut into different kind of effects which happens sometimes multiple times and leads to very confusing FOCS scripting .

I am not sure that ordering effects execution along priorities will clean this really up.
Maybe it does, but for the tech research case it might be possible to reduce complexity instead:
 just move the tech research to the end of the turn, so that it the only effect tech makes on the research turn is the sitrep informing one that the tech was successfully researched. None of our techs really needs initialization (like ships do).
[...]

PR #2370, opened 2019-02-04
CheckResearch as is but add innovations at end of turn processing

agrrr3:
This PR relates to issue #2355 and moves the addition of newly researched techs ('innovations') to the end of turn processing while keeping the calculation of researched techs where it is (CheckResearch call in ServerApp).  This removes unnessecary complexity from effects and lessens the burden of understanding of FOCS scripters.

>  it might be possible to reduce complexity instead:
just move the tech research to the end of the turn, so that it the only effect tech makes on the research turn is the sitrep informing one that the tech was successfully researched. None of our techs really needs initialization (like ships do).
The reduction of complexity stems from the fact that currently tech is researched in the middle of turn processing, leading to two different periods inside turn processing (effect application before tech is researched and effect application after tech is researched) instead of one (effect application is always after tech is researched).

Regarding implementation: C++ gurus please check if the collection of the innovations is c++ idiomatic.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -598,6 +598,7 @@ void Empire::Eliminate() {
     // uncleared after elimination
 
     m_capital_id = INVALID_OBJECT_ID;
+    // m_new_techs
     // m_techs
     m_research_queue.clear();
     m_research_progress.clear();
@@ -976,7 +977,7 @@ void Empire::SetResourceStockpile(ResourceType resource_type, float stockpile) {
 }
 
 void Empire::PlaceTechInQueue(const std::string& name, int pos/* = -1*/) {
-    if (name.empty() || TechResearched(name) || m_techs.count(name))
+    if (name.empty() || TechResearched(name) || m_techs.count(name) || m_new_techs.count(name))
         return;
     const Tech* tech = GetTech(name);
     if (!tech || !tech->Researchable())
@@ -1336,21 +1337,36 @@ void Empire::ConquerProductionQueueItemsAtLocation(int location_id, int empire_i
     }
 }
 
-void Empire::AddTech(const std::string& name) {
+void Empire::AddNewTech(const std::string& name) {
     const Tech* tech = GetTech(name);
     if (!tech) {
-        ErrorLogger() << "Empire::AddTech given and invalid tech: " << name;
+        ErrorLogger() << "Empire::AddNewTech given an invalid tech: " << name;
         return;
     }
 
-    if (!m_techs.count(name))
+    if (m_techs.count(name))
+        return;
+
+    auto result = m_new_techs.insert(name);
+    if (result.second)
         AddSitRepEntry(CreateTechResearchedSitRep(name));
+}
 
-    for (const ItemSpec& item : tech->UnlockedItems())
-        UnlockItem(item);  // potential infinite if a tech (in)directly unlocks itself?
+void Empire::ApplyNewTechs() {
+    for (auto new_tech : m_new_techs) {
+        const Tech* tech = GetTech(new_tech);
+        if (!tech) {
+            ErrorLogger() << "Empire::ApplyNewTech has an invalid entry in m_new_techs: " << new_tech;
+            return;
+        }
 
-    if (!m_techs.count(name))
-        m_techs[name] = CurrentTurn();
+        for (const ItemSpec& item : tech->UnlockedItems())
+            UnlockItem(item);  // potential infinite if a tech (in)directly unlocks itself?
+
+        if (!m_techs.count(new_tech))
+            m_techs[new_tech] = CurrentTurn();
+    }
+    m_new_techs.clear();
 }
 
 void Empire::UnlockItem(const ItemSpec& item) {
@@ -1368,7 +1384,7 @@ void Empire::UnlockItem(const ItemSpec& item) {
         AddShipDesign(GetPredefinedShipDesignManager().GetDesignID(item.name));
         break;
     case UIT_TECH:
-        AddTech(item.name);
+        AddNewTech(item.name);
         break;
     default:
         ErrorLogger() << "Empire::UnlockItem : passed ItemSpec with unrecognized UnlockableItemType";
@@ -1571,14 +1587,14 @@ namespace {
     }
 }
 
-void Empire::CheckResearchProgress() {
+std::vector<std::string> Empire::CheckResearchProgress() {
     SanitizeResearchQueue(m_research_queue);
 
     float spent_rp{0.0f};
     float total_rp_available = m_resource_pools[RE_RESEARCH]->TotalAvailable();
 
     // process items on queue
-    std::vector<std::string> to_erase;
+    std::vector<std::string> to_erase_and_add;
     for (auto& elem : m_research_queue) {
         const Tech* tech = GetTech(elem.name);
         if (!tech) {
@@ -1589,11 +1605,9 @@ void Empire::CheckResearchProgress() {
         float tech_cost = tech->ResearchCost(m_id);
         progress += elem.allocated_rp / std::max(EPSILON, tech_cost);
         spent_rp += elem.allocated_rp;
-        if (tech->ResearchCost(m_id) - EPSILON <= progress * tech_cost)
-            AddTech(elem.name);
-        if (GetTechStatus(elem.name) == TS_COMPLETE) {
+        if (tech->ResearchCost(m_id) - EPSILON <= progress * tech_cost) {
             m_research_progress.erase(elem.name);
-            to_erase.push_back(elem.name);
+            to_erase_and_add.push_back(elem.name);
         }
     }
 
@@ -1659,14 +1673,14 @@ void Empire::CheckResearchProgress() {
         rp_left_to_spend -= consumed_rp;
 
         if (tech->ResearchCost(m_id) - EPSILON <= m_research_progress[cost_tech.second] * tech_total_cost)
-            AddTech(cost_tech.second);
+            to_erase_and_add.push_back(cost_tech.second);
 
         //DebugLogger() << "... allocated: " << consumed_rp << " to increase progress by: " << progress_increase;
     }
 
     // remove completed items from queue (after consuming extra RP, as that
     // determination uses the contents of the queue as input)
-    for (const std::string& tech_name : to_erase) {
+    for (const std::string& tech_name : to_erase_and_add) {
         auto temp_it = m_research_queue.find(tech_name);
         if (temp_it != m_research_queue.end())
             m_research_queue.erase(temp_it);
@@ -1674,6 +1688,7 @@ void Empire::CheckResearchProgress() {
 
     // can uncomment following line when / if research stockpiling is enabled...
     // m_resource_pools[RE_RESEARCH]->SetStockpile(m_resource_pools[RE_RESEARCH]->TotalAvailable() - m_research_queue.TotalRPsSpent());
+    return to_erase_and_add;
 }
 
 void Empire::CheckProductionProgress() {
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -598,7 +598,7 @@ void Empire::Eliminate() {
     // uncleared after elimination
 
     m_capital_id = INVALID_OBJECT_ID;
-    // m_new_techs
+    // m_newly_researched_techs
     // m_techs
     m_research_queue.clear();
     m_research_progress.clear();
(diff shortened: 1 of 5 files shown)


geoffthemedio:
> C++ gurus please check if the collection of the innovations is c++ idiomatic.

The test builds failed, so consult those error messages first.

https://ci.appveyor.com/project/freeorion/freeorion/builds/22103782#L242

agrrr3:
Hm the travis CI fail does not look like its coming from my changes(?)
```
14:56:19.200368 [error] python : configure_logging.py:199 :     handlers = split(get_option_dict()[HANDLERS])
14:56:19.200385 [error] python : configure_logging.py:199 : KeyError: 'handlers'
14:56:19.200401 [error] python : configure_logging.py:199 : 
14:56:19.200584 [error] ai : AIClientApp.cpp:196 : AIClientApp failed due to error in python AI code for AI_2.  Exiting Soon. id = -1
```

server/ServerApp.cpp
```diff
@@ -3333,12 +3333,13 @@ void ServerApp::PostCombatProcessTurns() {
     // Consume distributed resources to planets and on queues, create new
     // objects for completed production and give techs to empires that have
     // researched them
+    std::vector<std::pair<Empire *,std::vector<std::string>>> empires_innovations;
```
geoffthemedio: better to store `int` empire ID than the pointer, but if you store a pointer, write it `Empire*, std::vector<std::string>` with no space before `*` and a space after `,`

agrrr3: Trying to understand your reasoning:
This is local vector and I need a reference/pointer later in the same function.
Doing a lookup using the empire ID would be better because the empire could be deleted in between(?)

geoffthemedio: It's probably fine here, but in general I prefer to keep track of IDs rather than raw pointers to stuff. Perhaps someone could be inspired by this bit of code to do something more problematic elsewhere.

agrrr3:
One important question: which turn number should be registered for TurnTechResearched? The current turn or current turn plus one?

If it stays like it is implemented, TurnTechResearched will never match CurrentTurn. If it stays like this, the "fix" is that I would document in the wiki that the first turn the tech has an effect is TurnTechResearch plus one.

If current turn plus one is registered, the automatic tech researched sitrep would appear on turn TurnTechResearched minus one.

I do not really have a preference, one could argue both ways.

agrrr3:
"Fallback clouds stack is too big." sais the AppVeyor build, pretty sure this is coming from infrastructure and not from my changes.

Dilvish-fo:
I've made some comments in the underlying issue; while the discussion could go either there or here I think its a bit more suited to there.

server/ServerApp.cpp
```diff
@@ -3371,6 +3372,14 @@ void ServerApp::PostCombatProcessTurns() {
     m_universe.BackPropagateObjectMeters();
     empires.BackPropagateMeters();
 
+    // add new techs
```
geoffthemedio: Moving the adding of new techs to here, after scripted turn events and executing meters on newly-created objects means that tech research won't have any effect on new objects created during the turn processing after a player clicks turn. That could be potentially annoying or problematic in some cases, where having the tech researched would have some benefit to ships produced when the tech is researched. Similarly, any scripted turn events would execute as though the tech was not finished yet, which means one needs to finish researching techs a full turn prior to getting their benefit / effect in that context.

I'm unsure if either of these are really substantial issues though, so are not major concerns to me now...

geoffthemedio:
> One important question: which turn number should be registered for TurnTechResearched? The current turn or current turn plus one?

Does adding the techs to the empire's list of researched techs where you have it now do anything different than would adding them after `m_current_turn` is incremented? I didn't see anything between these that would depend on the researched techs directly (vs. indirectly through meter values such as for visibility checks). So presumably the techs could be added to the empire immediately after the turn increment for the same result?

If so, then I think it's potentially useful to have the first turn a tech's effects could fire be equal to the recorded turn the tech was researched, so some first-turn effects can be more clearly scripted.

Edit: I suppose "the first turn a tech's effects could fire" is a bit unclear given the changes in this pull request. If the player presses turn on turn N, a tech accumulates enough research and is added to the empire during that interturn, and then get a sitrep on turn N+1 saying the tech is researched, then it's reasonable to record turn N or turn N+1 as the turn the tech is researched, but not turn N+2, which after this pull request, would apparently be the first turn the tech's effects could fire. I'd be fine with recording turn N+1 for the turn researched.

server/ServerApp.cpp
```diff
@@ -3371,6 +3372,14 @@ void ServerApp::PostCombatProcessTurns() {
...
+    for (const auto& empire_techs : empires_innovations) {
```
geoffthemedio: What's the reason for adding techs to empires here / now, though?

It's moved from before scripted turn events, a meter update, pop growth production research phase (which currently just does planet depopulation), and clamping and back-propagation, to after all of those. What specifically is it being moved after to do something usefully different? I'm guessing it's the `ApplyMeterEffectsAndUpdateMeters` call? But why specifically?

Edit: I now think the main difference is the moving to after the backpropagation /Edit

Adding the techs is also still before an empire capital check, various visibility and object knowledge tests and empire elimination... Why not later? Having it before the turn number increment seems contradictory to what you wanted to accomplish (though I'm still vague on what that is exactly)...

server/ServerApp.cpp
geoffthemedio: If I understand correctly after reading a bit more of the related and background threads, I think you intend to delay the effects from a newly-researched tech from executing by one turn.

So, the player would see the "tech researched" sitrep when receiving the updates for turn N, but the effects of that tech won't contribute to the actual meter values (and resource output or detection range) that the player sees and gets until turn N+1. Presumably the tech effects would still be apparent in the next turn predicted meter values the player sees, though.

The code difference is adding of techs to the empire after, instead of before, the `ApplyAllEffectsAndUpdateMeters` and `BackPropagateObjectMeters` calls. The back propagation sets the actual meter values that will be used for the subsequent turn processing, and the `UpdateMeterEstimates` below will just update the expected next turn meter values (which may change when the player submits orders).

This has an advantage, in that the expected next turn meter values are more accurate... researching a tech won't cause them to change between when the player submits orders and when they get back the results of the next turn (although other unknown or unpredictable factors may make the estimates inaccurate).

It does also mean that it takes an extra turn between being informed a tech is researched and getting the benefits of the tech. This is potentially annoying / disappointing for players, although a bit less so in that they should be able to see the differences in the expected next turn meter values on the same turn the player gets the tech researched sitrep.

server/ServerApp.cpp
geoffthemedio: I suppose it doesn't matter much if techs are actually added to the empire before or after the turn number increment as far as the player is concerned; the number can be tweaked for record keeping purposes either way.

But it is worth considering what will happen if an effect grants an empire a tech. Presumably that should be recorded as being researched on the same turn as a researched tech regardless of the mechanism by which the tech is added. But, I expect that tech-granting effects will fire earlier in the turn sequence, right after the comment `// execute all effects and update meters prior to production, research, etc.`, and will directly add the tech to the empire, before the `CheckResearchProgress` calls even happen (or the subsequent `ApplyMeterEffectsAndUpdateMeters` call and meter back propagation)...

agrrr3: So I guess the tech granting effect should in the same way "delay" tech application as the techs granted by research?

geoffthemedio: There'd need to be a list of techs to grant, and the effect would add to that, rather than granting them directly... Similar to effects that directly or indirectly destroy objects, or setting object visibility, which add those objects to lists that are processed later.

agrrr3: > What's the reason for adding techs to empires here / now, though?

The important part is that it is later than the  `ApplyMeterEffectsAndUpdateMeters` call because that one will have a different result than the first effect invocation.
A good example is #2555 the first effect invocation happens before the better mine tech is added. So it does the lower tech version and generate the appriopiate sitrep. If the tech is added before the second effect application, it will apply the higher tech version damage but not generate any sitrep. So the player will get the wrong sitrep. 

> Adding the techs is also still before an empire capital check, various visibility and object knowledge tests and empire elimination... Why not later? Having it before the turn number increment seems contradictory to what you wanted to accomplish (though I'm still vague on what that is exactly)...

I guess it could/should be later. I will push the addition of techs after turn increment. 

Tech adding effects are also still on the TODO list for this PR. So is this work-in-progress?

agrrr3: Done/Ready-for-review. 
Turn tech researched for initial tech is turn -32768 (on my machine)
For all other tech it is the turn in which the SitRep is shown.

Note GiveTechToEmpire works delayed; this gives an opportunity to add a SitRep to the turn the "Tech researched" message is shown; e.g. "Conquering the planet of the ancient FooBar race provides you with the wisdom of the 'Banana Peeling' technology."

Empire/Empire.cpp
```diff
@@ -976,7 +977,7 @@ void Empire::SetResourceStockpile(ResourceType resource_type, float stockpile) {
 }
 
 void Empire::PlaceTechInQueue(const std::string& name, int pos/* = -1*/) {
-    if (name.empty() || TechResearched(name) || m_techs.count(name))
+    if (name.empty() || TechResearched(name) || m_techs.count(name) || m_new_techs.count(name))
```
geoffthemedio: meaning is unclear and `m_new_techs` lacks context. Please explain with a comment.

Empire/Empire.cpp
```diff
@@ -1336,21 +1337,36 @@ void Empire::ConquerProductionQueueItemsAtLocation(int location_id, int empire_i
     }
 }
 
-void Empire::AddTech(const std::string& name) {
+void Empire::AddNewTech(const std::string& name) {
```
geoffthemedio: "New" in what sense? Maybe `AddNewlyResearchedTechToGrantAtStartOfNextTurn`?

Empire/Empire.cpp
```diff
@@ -1336,21 +1337,36 @@ void Empire::ConquerProductionQueueItemsAtLocation(int location_id, int empire_i
...
+        ErrorLogger() << "Empire::AddNewTech given an invalid tech: " << name;
         return;
     }
 
-    if (!m_techs.count(name))
+    if (m_techs.count(name))
+        return;
+
+    auto result = m_new_techs.insert(name);
+    if (result.second)
```
geoffthemedio, line 1353: explain what this does / when it doesn't do anything...

Empire/Empire.cpp
```diff
@@ -1336,21 +1337,36 @@ void Empire::ConquerProductionQueueItemsAtLocation(int location_id, int empire_i
...
+}
 
-    for (const ItemSpec& item : tech->UnlockedItems())
-        UnlockItem(item);  // potential infinite if a tech (in)directly unlocks itself?
+void Empire::ApplyNewTechs() {
+    for (auto new_tech : m_new_techs) {
+        const Tech* tech = GetTech(new_tech);
+        if (!tech) {
+            ErrorLogger() << "Empire::ApplyNewTech has an invalid entry in m_new_techs: " << new_tech;
+            return;
```
geoffthemedio: continue rather then fully aborting...

Empire/Empire.cpp
```diff
@@ -1571,14 +1587,14 @@ namespace {
...
-void Empire::CheckResearchProgress() {
+std::vector<std::string> Empire::CheckResearchProgress() {
     SanitizeResearchQueue(m_research_queue);
 
     float spent_rp{0.0f};
     float total_rp_available = m_resource_pools[RE_RESEARCH]->TotalAvailable();
 
     // process items on queue
-    std::vector<std::string> to_erase;
+    std::vector<std::string> to_erase_and_add;
```
geoffthemedio: add comment... to erase from queue and add to empire

Empire/Empire.h
```diff
@@ -217,7 +217,8 @@ class FO_COMMON_API Empire {
     void ResumeProduction(int index);
     void AllowUseImperialPP(int index, bool allow=true);  ///< Allows or disallows the use of the imperial stockpile for production
 
-    void AddTech(const std::string& name);           ///< Inserts the given Tech into the Empire's list of available technologies.
+    void AddNewTech(const std::string& name);           ///< Inserts the given Tech into the Empire's list of innovations. Call ApplyAddedTech to make it effective.
+    void ApplyNewTechs();                          ///< Moves all Techs from the Empire's list of innovations into the Empire's list of available technologies.
```
geoffthemedio: align nearby comment starts

PR #2373, opened 2019-02-04
Improve AI military logic

Morlic-fo:
* Increase standard enemy weapon damage to MD4 levels (old value leads to drastic overevaluation of shields in the early game)
* Make the AI more conservative when choosing its battles by tweaking the military_safety_factor. The goal is to avoid unnecessary losses in higher aggression settings and also to give more robustness against errors in our combat ratings heuristic.

### DIFF
--- a/default/python/AI/CombatRatingsAI.py
+++ b/default/python/AI/CombatRatingsAI.py
@@ -33,7 +33,7 @@ def default_ship_stats():
     :return: Some weak standard ship
     :rtype: ShipCombatStats
     """
-    attacks = (4.0, 1)
+    attacks = (6.0, 1)
     structure = 15
     shields = 0
     fighters = 0


Dilvish-fo:
I've been busy with other things for the last few weeks or so, and am catching up on things.  Have there been changes to the weapons numbers?  9 points looks like  maybe MD4 for Eaxaw to me, but Laser 3 for average species.  I'm not immediately convinced it's best to have the turn 1 default planning enemy be that high.   If the point is just to make them stop doing defense grids, then it may not matter if it really matches a likely enemy.  Defense grids can be pretty handy though against a number of the monsters that are often encountered before engaging with enemy empires, so it's also not immediately obvious to me we want to entirely block them.   What sort of game-effectiveness testing have you done for this separate from the accompanying change to safety factor?  (or with it also)

As for the safety factor increase, that looks like a really large increase.  And while it's not handy for me to double check at the moment, it's in my head that not so long ago I had slightly reduced those numbers in order to avoid so much of a competing doomstacks problem the AI had been having.  Now, I think I had also intended to further help the doomstack problem by having the required safety factor go down as the total fleet size went up, but I don't recall if I got anything in on that.  And even for smaller battles, where a bigger margin is more likely merited, I'm doubtful that a 50% - 100% increase in required safety margin would really make the AI more effective, and it seems at risk of making it less effective.  

It would be great if we could around to putting together an AI test-harness (of AI vs AI games where the AI's are loaded with different configurations) to automate collecting comparison data over many games, but in the meantime we are left with personal playtesting -- what can you report on that front for this?

Morlic-fo:
Not sure how I got the 9, you are of course correct and it should be 6 for MD4.

Argument for increasing the number is as follows:
1) If the AI makes late contact, it is expected that the enemy will have researched at least MD4 if not higher weapon levels.
2) If the AI makes early contact, it will be able to figure out the actual weapon strength anyway.
A better approach might be to get a value depending on current turn.



> As for the safety factor increase, that looks like a really large increase.

It's a factor of 2, i.e. requiring larger fleets by a factor of sqrt(2), i.e. about 40%. Maybe overvalued for low aggression settings, I tried to keep scaling between different settings mostly constant. On aggressive/maniacal, the AI currently keeps suiciding its fleets. Current problems:
a) our combat rating estimate is rather crude and makes assumptions about the distribution of structure/weapons which aren't always met in practice
b) the AI doesn't usually consider the actual enemy fleet it will fight
c) the AI doesn't consider reinforcements
d) the AI doesn't retreat in lost battles
e) the AI doesn't force scouting before moving into target systems
f) no AI adjustments to new targeting mechanics (#2330)
g) the AI does not consider cannon fodder

Each of them justifies a safety factor > 1.

Even if ratings were perfect, trading fleets 1 for 1 is a rather crippling move considering that your expected losses are pretty much your entire fleet by definition. There may be tactical value in that and it may even be a strategically fine move in a 1vs1 scenario - but considering there are N enemies in the game, (N-1) enemies just got a huge advantage on you (and on your other enemy). Especially if it was a large fleet.

The suiciding issue became very clear to me when testing #2374 which increases the number of times the AI actually makes aggressive moves (and sometimes resolves doomstack situations if one AI got ahead). While that particular issue could be adressed by an additional safety factor in the PROTECT_REGION mission, I think the issue is a global problem and thus should be adressed globally.

While doomstacks are bad, I am not sure that simply suiciding the doomstacks is the correct approach.

geoffthemedio:
@Morlic-fo since @Dilvish-fo seems MIA, would you merge this now or wait for further changes?

Morlic-fo:
Will test a bit more.

geoffthemedio:
@Morlic-fo Have you had a chance to test a bit more?

Morlic-fo:
Dropped the safety factor increase for now.

PR #2374, opened 2019-02-04
Introduce new AI Mission Type: PROTECT_REGION

Morlic-fo:
This PR introduces a new MissionType for the AI: PROTECT_REGION.

In this mission type, the fleet will not defend its target system passively but rather try to actively engage and intercept nearby (currently: neighboring) enemy fleets if there is no immediate threat to the (primary) target system. 

This PR also assigns the PROTECT_REGION mission to idle fleets as a filler mission (targeting the closest owned system).

### DIFF
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -514,6 +514,13 @@ class AIFleetMission(object):
                               "to secure system %d (targeted for %s), "
                               "may release a portion of ships" % (self.fleet.id, last_sys_target, secure_type))
                         clear_all = False
+
+                # for PROTECT_REGION missions, only release fleet if no more threat
+                if self.type == MissionType.PROTECT_REGION:
+                    # use military logic code below to determine if can release
+                    # any or even all of the ships.
+                    clear_all = False
+
                 fleet_id = self.fleet.id
                 if clear_all:
                     if orders:
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -148,7 +148,9 @@ class AIFleetMission(object):
         :param context: Context of the function call for logging purposes
         :type context: str
         """
+        debug("Considering to merge %s", self.fleet)
         if self.type not in MERGEABLE_MISSION_TYPES:
+            debug("Not mergable.")
             return
         universe = fo.getUniverse()
         empire_id = fo.empireID()
@@ -156,6 +158,7 @@ class AIFleetMission(object):
         main_fleet = universe.getFleet(fleet_id)
         system_id = main_fleet.systemID
         if system_id == INVALID_ID:
+            debug("Invalid system")
             return  # can't merge fleets in middle of starlane
         aistate = get_aistate()
         system_status = aistate.systemStatus[system_id]
@@ -168,6 +171,7 @@ class AIFleetMission(object):
         other_fleets_here = [fid for fid in system_status.get('myFleetsAccessible', []) if fid != fleet_id and
                              fid not in destroyed_list and universe.getFleet(fid).ownedBy(empire_id)]
         if not other_fleets_here:
+            debug("No other fleets here")
             return
 
         target_id = self.target.id if self.target else None
@@ -175,12 +179,15 @@ class AIFleetMission(object):
         for fid in other_fleets_here:
             fleet_role = aistate.get_fleet_role(fid)
             if fleet_role not in COMPATIBLE_ROLES_MAP[main_fleet_role]:
+                debug("Local candidate %d not compatible", fid)
                 continue
             fleet = universe.getFleet(fid)
 
             if not fleet or fleet.systemID != system_id or len(fleet.shipIDs) == 0:
+                debug("Local candidate %d invalid", fid)
                 continue
             if not (fleet.speed > 0 or main_fleet.speed == 0):  # TODO(Cjkjvfnby) Check this condition
+                debug("Local candidate %d invalid", fid)
                 continue
             fleet_mission = aistate.get_fleet_mission(fid)
             do_merge = False
@@ -192,6 +199,7 @@ class AIFleetMission(object):
                 if main_fleet_role == fleet_role:
                     do_merge = False  # TODO: could allow merger if both orb invaders and both same target
             elif not fleet_mission and (main_fleet.speed > 0) and (fleet.speed > 0):
+                debug("Local candidate %d invalid", fid)
                 do_merge = True
             else:
                 if not self.target and (main_fleet.speed > 0 or fleet.speed == 0):
@@ -204,6 +212,7 @@ class AIFleetMission(object):
                         # TODO: should probably ensure that fleetA has aggression on now
                         do_merge = float(min(main_fleet.speed, fleet.speed))/max(main_fleet.speed, fleet.speed) >= 0.6
                     elif main_fleet.speed > 0:
+                        debug("different targets")
                         neighbors = aistate.systemStatus.get(system_id, {}).get('neighbors', [])
                         if target == system_id and target_id in neighbors and self.type == MissionType.SECURE:
                             # consider 'borrowing' for work in neighbor system  # TODO check condition
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -142,86 +142,44 @@ class AIFleetMission(object):
 
     def check_mergers(self, context=""):
         """
-        If possible and reasonable, merge this fleet with others.
+        Merge local fleets with same mission into this fleet.
 
         :param context: Context of the function call for logging purposes
         :type context: str
         """
-        debug("Considering to merge %s", self.fleet)
+        debug("Considering to merge %s", self.__str__())
         if self.type not in MERGEABLE_MISSION_TYPES:
-            debug("Not mergable.")
+            debug("Mission type does not allow merging.")
             return
+
+        if not self.target:
+            debug("Mission has no valid target - do not merge.")
+            return
+
         universe = fo.getUniverse()
-        empire_id = fo.empireID()
         fleet_id = self.fleet.id
         main_fleet = universe.getFleet(fleet_id)
-        system_id = main_fleet.systemID
-        if system_id == INVALID_ID:
-            debug("Invalid system")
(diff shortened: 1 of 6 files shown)


Dilvish-fo:
Looks like a great idea, and I like very much that you've implemented the neighbor threat check to help protect against getting lured into ambushes.  I didn't go through the connections with the regular code enough to determine just when the protection mission would end; could you summarize that for me?

Morlic-fo:
> I like very much that you've implemented the neighbor threat check to help protect against getting lured into ambushes

That isn't actually implemented yet. For now, it will still move to the target system if there is an overwhelming threat nearby (hopefully, often enough the military allocation code was correct in assuming the fleet and planetary defenses in target system are sufficient). While avoiding ambushes is definitely a good idea, that should be implemented and aimed for all mission types not this mission type in particular and so is somewhat beyond the scope of this PR.


> I didn't go through the connections with the regular code enough to determine just when the protection mission would end; could you summarize that for me?

For now, the mission will (i.e. should) end when the last order is executed - either by arriving at the target system or at a neighboring system. The idea is to let the MilitaryAI reallocate (portions of) the fleet to the same system if necessary or abandon the system if higher priority targets must be defended.

Maybe should add additional checks to verify the threat has been neutralized...

Morlic-fo:
There is some issue related with fleet mergers which should be adressed before merging this. Working on a fix.

Morlic-fo:
Note that this PR resolves a huge issue with doomstacks.

A common case is where empires A and B have neighboring colonies. Each of them only wants to protect its own colony so feedback will result in them stacking up defenses but noone makes a move (as no AI ever intended to attack).
With this PR, the runaway AI will make an attack move into the neighbor system, eliminating the enemy doomstack. Afterwards, it can reuse its fleet for other missions.

Take a look at this example savegame: [example_protect_region_doomstacks_resolved.zip](https://github.com/freeorion/freeorion/files/2862137/example_protect_region_doomstacks_resolved.zip)
The doomstack situation occurs in 2 locations:
1) Sakuru alpha (1071) and Hagalaz (987)
2) Yarrow beta (868) and Yarrow gamma (1001)

With the full PR, the respectively leading AI will merge its doomstack and engage the enemy.

Not ready for merge yet - AI keeps merging and splitting fleets multiple times at this point, looking for best fix.

default/python/AI/EnumsAI.py
```diff
@@ -102,6 +102,7 @@ class MissionType(Enum):
     ORBITAL_DEFENSE = 12
     ORBITAL_INVASION = 13
     ORBITAL_OUTPOST = 14
+    PROTECT_REGION = 15
     # ORBITAL_COLONISATION = 15 Not implemented yet
```
genestack-solomatin, line 105: Change value for `PROTECT_REGION` or for `ORBITAL_COLONISATION`

default/python/AI/MilitaryAI.py
```diff
@@ -921,10 +923,46 @@ def assign_military_fleets_to_systems(use_fleet_id_list=None, allocations=None,
...
+                elif get_system_jump2_threat(_sys_id):
+                    weight = 1
+                else:
+                    weight = 1 / float(state.get_distance_to_enemy_supply(_sys_id))**1.25
+                return float(weight) / (jump_distance+1)
+
+            for fid in avail_mil_fleet_ids:
+                fleet = universe.getFleet(fid)
+                FleetUtilsAI.get_fleet_system(fleet)
+                sys_distances = sorted([(system_score(fid, _sys_id), _sys_id)
```
genestack-solomatin:
```python
systems = state.get_empire_planets_by_system().keys()
if not systems: 
    continue
sys_id = max(systems,  key=lambda x: system_score(fid, x))

**Review by genestack-solomatin — CHANGES_REQUESTED** [review]

File: default/python/AI/AIFleetMission.py [hunk 3]
```diff
@@ -139,79 +143,53 @@ def _get_fleet_order_from_target(self, mission_type, target):
...
             return
+
+        if not self.target:
+            debug("Mission has no valid target - do not merge.")
+            return
+
         universe = fo.getUniverse()
         empire_id = fo.empireID()
+
+        # only merge PROTECT_REGION if there is any threat near target
```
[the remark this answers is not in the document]
Morlic-fo: If fleet A and fleet B have different mission types or different targets, then they will not be merged.

Morlic-fo:
Overall happy with how this plays now - ready to merge from my side.

geoffthemedio:
@Morlic-fo I'll give @Dilvish-fo or @Cjkjvfnby a few days to comment, though don't expect any given recent inactivity.

default/python/AI/AIFleetMission.py
```diff
@@ -139,79 +135,53 @@ def _get_fleet_order_from_target(self, mission_type, target):
...
-                            need_left = 1.5 * sum(aistate.systemStatus.get(nid, {}).get('fleetThreat', 0)
-                                                  for nid in neighbors if nid != target_id)
-                            fleet_rating = CombatRatingsAI.get_fleet_rating(fid)
-                            if need_left < fleet_rating:
-                                do_merge = True
-            if do_merge:
-                FleetUtilsAI.merge_fleet_a_into_b(fid, fleet_id, need_left,
-                                                  context="Order %s of mission %s" % (context, self))
-        return
+            if fleet_mission.type != self.type or fleet_mission.target != self.target:
```
Cjkjvfnby, line 181: FYI: you can replace

```python
fleet_mission.type != self.type or fleet_mission.target != self.target
``` 
with
```python
(fleet_mission.type, fleet_mission.target ) != (self.type, self.target)
```

Sometimes this approach can improve readability.

default/python/AI/AIFleetMission.py
```diff
@@ -139,79 +135,53 @@ def _get_fleet_order_from_target(self, mission_type, target):
...
-        # merging if there is no local threat (it tends to lead to fleet object churn)
-        if self.type in COMBAT_MISSION_TYPES and len(self.orders) == 1 and not system_status.get('totalThreat', 0):
+        main_fleet_system_id = main_fleet.systemID
+        if main_fleet_system_id == INVALID_ID:
+            debug("Can't merge: fleet in middle of starlane")
             return
-        destroyed_list = list(universe.destroyedObjectIDs(empire_id))
+
+        # only merge PROTECT_REGION if there is any threat near target
+        if self.type == MissionType.PROTECT_REGION:
```
Cjkjvfnby, line 163: Consider adding TODO:
```
Logic about cheking if we can merge certain mission types can be moved to MissionType class itself.
```


PS.  It can be easily implemented via inheritance or via mixins.

Mixin approach:

```python

class Base:
    def mergable(self, other):
        return False
    
# has a single method!
class Mergable:
    def mergable(self, other):
        return True
    
class TypaA(Base):
    pass

class TypeB(Base, Mergable):
    pass


class TypeC(Base):
    def mergable(self, other):
        ...
```

default/python/AI/AIFleetMission.py
```diff
@@ -564,6 +590,41 @@ def issue_fleet_orders(self):
                         MilitaryAI.assign_military_fleets_to_systems(use_fleet_id_list=new_military_fleets,
                                                                      allocations=allocations)
 
+    def _portion_of_fleet_needed_here(self):
+        """Calculate the portion of the fleet needed in target system considering enemy forces."""
+        # TODO check rating against planets
+        if assertion_fails(self.type in COMBAT_MISSION_TYPES):
+            error("%s" % self)
```
Cjkjvfnby: Can be simplified to `error(self)`.

You can add mesage to check.
`assertion_fails(self.type in COMBAT_MISSION_TYPES, str(self)):`

default/python/AI/AIFleetMission.py
```diff
@@ -564,6 +590,41 @@ def issue_fleet_orders(self):
...
+            return 0
+        if assertion_fails(self.target and self.target.id != INVALID_ID):
+            error("%s" % self)
```
Cjkjvfnby: same here

default/python/AI/AIFleetMission.py
```diff
@@ -660,3 +725,85 @@ def __str__(self):
                                                                             (fleet and len(fleet.shipIDs)) or 0,
                                                                             CombatRatingsAI.get_fleet_rating(fleet_id),
                                                                             self.target or 'no target')
+
+    def _get_target_for_protection_mission(self):
+        """Get a target for a PROTECT_REGION mission.
+
+        1) If primary target (system target of this mission) is under attack, move to primary target.
```
Cjkjvfnby, line 730: what if not strong enough?

default/python/AI/AIFleetMission.py
```diff
@@ -660,3 +725,85 @@ def __str__(self):
...
+        2) If neighbors of primary target have local fleets weaker than this fleet, may move to attack
```
Cjkjvfnby: only if neighbors are attacked?

default/python/AI/AIFleetMission.py
```diff
@@ -660,3 +725,85 @@ def __str__(self):
...
+        3) If no neighboring fleets or strongest enemy fleet is too strong, move to primary target
```
Cjkjvfnby: `strongest enemy fleet is too strong` should we consider all fleets?

Cjkjvfnby: Stopped here.

default/python/AI/CombatRatingsAI.py
```diff
@@ -291,6 +291,10 @@ def get_fleet_rating_against_planets(fleet_id):
     return FleetCombatStats(fleet_id, consider_refuel=False).get_rating_vs_planets()
 
 
+def get_ship_rating(ship_id, enemy_stats=None):
```
Cjkjvfnby, line 294: Do we need one line function for single use?

Cjkjvfnby:
> Overall happy with how this plays now - ready to merge from my side.

I did not check how it affects game balance, just code organization.  All my comments are optional.

default/python/AI/AIFleetMission.py
Morlic-fo: As is the case with current mission types, it will finally be handled by movement check code at later stage where more information is available.

Issue #2372, opened 2019-02-04
Travis CI is broken


agrrr3:
Looks the Travis CI is broken. See e.g.  https://travis-ci.org/freeorion/freeorion/jobs/488304756
```
01:29:42.293117 [error] python : configure_logging.py:199 : KeyError: 'handlers'
01:29:42.293141 [error] python : configure_logging.py:199 : 
01:29:42.293364 [error] ai : AIClientApp.cpp:196 : AIClientApp failed due to error in python AI code for AI_1.  Exiting Soon. id = -1
terminate called after throwing an instance of 'std::runtime_error'
```
Note there may be more issues in there.

Morlic-fo:
Relevant part of the log:
```
01:29:42.292859 [error] python : configure_logging.py:199 : Uncaught exception: Traceback (most recent call last):
01:29:42.293037 [error] python : configure_logging.py:199 :   File "/freeorion/default/python/AI/FreeOrionAI.py", line 374, in <module>
01:29:42.293065 [error] python : configure_logging.py:199 :     init_handlers(fo.getOptionsDBOptionStr("ai-config"), fo.getAIDir())
01:29:42.293084 [error] python : configure_logging.py:199 :   File "/freeorion/build/default/python/common/handlers.py", line 11, in init_handlers
01:29:42.293101 [error] python : configure_logging.py:199 :     handlers = split(get_option_dict()[HANDLERS])
01:29:42.293092 [debug] server : Message.cpp:895 : ExtractJoinGameMessageData() from Default_Player_Name_in_WaitingForSPGameJoiners::react(const JoinGame& msg) client type -1
01:29:42.293117 [error] python : configure_logging.py:199 : KeyError: 'handlers'
01:29:42.293141 [error] python : configure_logging.py:199 : 
01:29:42.293364 [error] ai : AIClientApp.cpp:196 : AIClientApp failed due to error in python AI code for AI_1.  Exiting Soon. id = -1
terminate called after throwing an instance of 'std::runtime_error'
  what():  PythonAI failed to initialize.
```

geoffthemedio:
I don't see the errors mentioned earlier in comments, but the "travis-ci/push" builds seem to randomly fail or not fail the "C++" test. eg.
https://travis-ci.org/freeorion/freeorion/builds/585164449
and
https://travis-ci.org/freeorion/freeorion/builds/585170881

one of which passed and one didn't pass, apparently due to
`3 - SmokeTestHostless (Failed)`
due to
[...]

PR #2375, opened 2019-02-05
Bandaid-fix for random AI CI errors

Morlic-fo:
We keep seeing the apparently random error reported in #2372 which breaks CI builds but hasn't been observed outside of it.
So, while a proper fix would be preferable, handling the uncaught exception should be sufficient as interim solution.


geoffthemedio:
Doesn't fix the tests, but at least seems to run more of them / longer than the existing state...

PR #2336, opened 2018-12-14
Big pictures of species (Trenchers, Abaddoni, Scylior, Ugmors, Banforo, Phinnert).

Bestary:
Big illustrations of: Trenchers, Abaddoni, Scylior, Ugmors, Banforo, Phinnert, Gysache, Chato, Happybirthdays, Trith.
Also updated stringtables to insert images to articles.

<s>Unfortunatly, they looks not very good now. My dream was do wrapping on the right side, where are a lot of free space, but I'm not a programmer. ^^' Geoff said, there are no specific tools for it at this moment</s>.

Thanks Ouaz for all awesome answers. :)

[Forum topic is here.](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11095)

PR #2378, opened 2019-02-08
Rearrange Species Article and Add Art

geoffthemedio:
Related to suggestions on forums: https://www.freeorion.org/forum/viewtopic.php?p=94253#p94253

Example image of resizing art in pedia article...

chato_article_art


Ouaz:
Nice!
Some thoughts:

- Is it possible to dynamically display the image without adding the image filepath directly in the stringtable, like with the environment image implementation? (https://github.com/freeorion/freeorion/blob/master/UI/EncyclopediaDetailPanel.cpp#L2699)

- As the image is dynamically resized when the window size changes, we need to define a new default size for each species image (banner + full artwork), keeping in mind to have the most light-weight possible images, but large enough for people who play with an high screen resolution. 

Currently, in #2336 , we have roughly 300Kb per species. There are 46 species in FO (not all active), so we would have 46*300= 13.8Mb total size for all species images (approximately +7.5% of FO installation size -currently 183Mb on Windows)

I suppose the 2400x800px and 3000x3000px images in this PR are just placeholders (you took it from the FO assets repository), because with an added weight of 4Mb per species, we would reach a mind-blowing 184Mb only for the species arts... (it's just to give an idea, I know this size won't be used ^^)

I think we can go for 500Kb per species (~23Mb total size) and define which image dimensions for both the banners and the full artworks will give this total size. For the banners, 600x200px (always a 3:1 ratio) will output roughly a 250Kb image size, for example.

- There's already `*speciesname*.png` image files in _/data/art/icons/species_ (the black&white species icons used in the upper right corner of the pedia panel).
In order to not have files with identical names (even if it's in two different folders), we should name each file like this, according its folder:

`speciesname.png` = _/data/art/icons/species_  (pedia species icon)
`speciesname_banner.png` = _/data/art/species/preview_ (banner at the beginning of a species article)
`speciesname_art.png` = _/data/art/species_ (full artwork at the end of a species article)

geoffthemedio:
> Is it possible to dynamically display the image without adding the image filepath directly in the stringtable...

Yes, but I'd rather have it be in the stringtable. The more of the article content / layout that is in the stringtable entry, the more easily it can be edited / rearranged / changed. If you dislike the markup in the translated text, a macro could hide most of it.

> ... have the most light-weight possible images, but large enough for people who play with an high screen resolution.

What would help a lot is images without transparency. Then they could be saved as moderately compressed .jpg instead of .png, which is substantially bigger for these images... about 4-fold for saving at 95% quality when saving in Paint.net.

836 KB PNG with transparency:
chato_1080
202 KB JPG without transparency:
chato_1080

I also tried [crushing](https://pngcrush.com/) the png, but that saved only about 10%

> In order to not have files with identical names (even if it's in two different folders), we should name each file like this...

I don't see a problem with same-named files that are in different directories. Certainly seems better than adding redundant info to the filename that the directory structure already indicates, or is too generic to be helpful, like '_art'.

Bestary:
@geoffthemedio, thanks so much, it's looks very awesome to me!! * ___ * You made my day so much happier ^^ It's really good news!

@Ouaz, thanks so much to show me this PR! (searching English forums hard for me some, I'll can miss important news, unfortunatly).

I also wanted to offer re-save pictures in JPG and can do it with minimal loss of quality and minimal size, but there will black background will appear in big art. Also there may can appear gray lines on borders (maybe not). I don't think it's a big problem. Even if FreeOrion will change his design oneday to another color, there are still png-sourses that have transparency.

geoffthemedio:
@Bestary perhaps you could save the cropped 3x1 aspect ratio banner images as JPG, tweaking to remove any transparency / avoid issues with needing to pick a background colour, and keep the square images as PNG but with a smaller resolution. I'd rather not save the square images with (currently) transparent background as jpg with fixed black background, as the background colour in the UI is user-adjustable, and shouldn't be assumed to be currently always black.

Bestary:
@geoffthemedio, oh, ok. Yes, previews needs to be turn to JPG unconditionally. I didn't knew that there are settings for bg color, that awesome. :)

geoffthemedio:
I committed the pedia re-arrangement part of this. Adding image files and updating the stringtable entries can be done separately.

https://github.com/freeorion/freeorion/commit/58cef69f91c299ecad1b2bc772bc1922c30c03d2

PR #2379, opened 2019-02-09
Allow human players to disconnect without stopping game

o01eg:
Next step of #2243 

Adds two options. The first is to define minimal connected human player to don't stop playing game. The second is to define maximum disconnected human player to don't stop playing game.
Default values acts as current setting when server stops game if no human player left or any alive empire was disconnected.

PR #2382, opened 2019-02-11
Protect empires of authenticated players

### LINKED ISSUE #2357: Disallow another player to play empire controlled by authorized player in load game.
### Environment
* **FreeOrion Version**: 0.4.8
* **Fetched as** 
  * Binary release
  * Weekly development build
  * Compiled from source

### Description
When authentication used to prevent other player to play as someone authenticated player it still possible to choose empire he played on loading saved game and get information or even make some turns in this game.

### Expected Result
When someone tries to load saved game with some authenticated players server should match current player name and previous player name otherwise return error.

### Steps to reproduce
* Start multiplayer game with authenticated players.
* Make a save.
* Load save without authenticated players and with different names.

o01eg:
Fix #2357 

Adds new flag to save game if player who controls empire was authenticated.
Forbids players with different player's name to choose this empire on loading.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -34,10 +34,11 @@ Empire::Empire() :
 { Init(); }
 
 Empire::Empire(const std::string& name, const std::string& player_name,
-               int empire_id, const GG::Clr& color) :
+               int empire_id, const GG::Clr& color, bool authenticated) :
     m_id(empire_id),
     m_name(name),
     m_player_name(player_name),
+    m_authenticated(authenticated),
     m_color(color),
     m_research_queue(m_id),
     m_production_queue(m_id)
@@ -67,6 +68,9 @@ const std::string& Empire::Name() const
 
 const std::string& Empire::PlayerName() const
 { return m_player_name; }
+
+bool Empire::IsAuthenticated() const
+{ return m_authenticated; }
 
 int Empire::EmpireID() const
 { return m_id; }
@@ -746,7 +750,7 @@ void Empire::UpdateSupplyUnobstructedSystems(const std::set<int>& known_systems,
     // fleets were not (meaning that the enemy fleets were continuing an existing blockade)
     // Friendly fleets can preserve available starlane accesss even if they are trying to leave the system
 
-    // Unrestricted lane access (i.e, (fleet->ArrivalStarlane() == system->ID()) ) is used as a proxy for 
+    // Unrestricted lane access (i.e, (fleet->ArrivalStarlane() == system->ID()) ) is used as a proxy for
     // order of arrival -- if an enemy has unrestricted lane access and you don't, they must have arrived
     // before you, or be in cahoots with someone who did.
     std::set<int> systems_containing_friendly_fleets;
@@ -837,7 +841,7 @@ void Empire::UpdateSupplyUnobstructedSystems(const std::set<int>& known_systems,
 
 void Empire::RecordPendingLaneUpdate(int start_system_id, int dest_system_id) {
     if (!m_supply_unobstructed_systems.count(start_system_id))
-        m_pending_system_exit_lanes[start_system_id].insert(dest_system_id); 
+        m_pending_system_exit_lanes[start_system_id].insert(dest_system_id);
     else { // if the system is unobstructed, mark all its lanes as avilable
         auto system = GetSystem(start_system_id);
         for (const auto& lane : system->StarlanesWormholes()) {
@@ -1067,7 +1071,7 @@ void Empire::PlaceProductionOnQueue(BuildType build_type, const std::string& nam
     }
 
     ProductionQueue::Element build(build_type, name, m_id, number, number, blocksize, location);
-    
+
     if (pos < 0 || static_cast<int>(m_production_queue.size()) <= pos)
         m_production_queue.push_back(build);
     else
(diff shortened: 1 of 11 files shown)


server/ServerFSM.cpp
```diff
@@ -1188,7 +1194,41 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
             // limit count of AI
             auto max_ai = GetOptionsDB().Get<int>("network.server.ai.max");
             if (ai_count <= max_ai || max_ai < 0) {
-                m_lobby_data->m_players    = incoming_lobby_data.m_players;
+                if (sender->HasAuthRole(Networking::ROLE_HOST)) {
+                    // don't check host.
+                    m_lobby_data->m_players    = incoming_lobby_data.m_players;
+                } else {
+                    // check player doesn't set protected empire to yourself
```
geoffthemedio: yourself -> themselves ?

not sure what this means...

server/ServerFSM.cpp
```diff
@@ -1188,7 +1194,41 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
...
+                                incorrect_empire = true;
+                            }
+                        }
+                        break;
+                    }
+                    if (incorrect_empire) {
+                        has_important_changes = true;
+                        player_setup_data_changed = true;
+                    } else {
+                        // ToDo: non-host player could change only AI and himself
```
geoffthemedio: could -> should only be able to ?

server/ServerFSM.cpp
```diff
@@ -1188,7 +1194,41 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
...
+                        m_lobby_data->m_players    = incoming_lobby_data.m_players;
```
geoffthemedio: if it's not lining up with something else, you don't need the extra spaces before `=`

Empire/Empire.cpp
```diff
@@ -68,6 +69,9 @@ const std::string& Empire::Name() const
 const std::string& Empire::PlayerName() const
 { return m_player_name; }
 
+bool Empire::IsAuthenticated() const
```
geoffthemedio, line 72: does `m_authenticated` make sense as part of empire, and not of the server players list?

geoffthemedio:
Some additional documentation in the code indicating what an empire being authenticated means would be helpful.

Empire/Empire.h
```diff
@@ -327,6 +328,8 @@ class FO_COMMON_API Empire {
 
     void UpdateOwnedObjectCounters();
 
+    void SetAuthenticated(bool authenticated=true);
```
geoffthemedio: prefer spaces around `=`

Empire/Empire.h
```diff
@@ -376,6 +379,7 @@ class FO_COMMON_API Empire {
     int                             m_id = ALL_EMPIRES;         ///< Empire's unique numeric id
     std::string                     m_name;                     ///< Empire's name
     std::string                     m_player_name;              ///< Empire's Player's name
+    bool                            m_authenticated;            ///< Empire's Player's authentication status
```
geoffthemedio: maybe this is the wrong thing to call this flag? It's not so much that the empire (or the player) is authenticated, but rather that the empire can only be played by the player who is authenticated with that name.

Empire/Empire.cpp
o01eg: To use it in multiplayer lobby I need to have it in `SaveGameEmpireData` which populated from `Empire`. Also I can't use server player list because after #2379 it could be possible someone would be disconnected and its authentication status would be unreachable.

server/ServerFSM.cpp
o01eg: Fixed. It means player could set protected empire only if player have requested name.

server/ServerFSM.cpp

server/ServerFSM.cpp

Empire/Empire.h
o01eg: Added more documentation here.

Empire/Empire.h
o01eg: Fixed.

server/ServerApp.cpp
```diff
@@ -1747,16 +1752,18 @@ void ServerApp::DropPlayerEmpireLink(int player_id)
 { m_player_empire_ids.erase(player_id); }
 
 int ServerApp::AddPlayerIntoGame(const PlayerConnectionPtr& player_connection) {
+    Empire* empire_ptr = nullptr;
```
geoffthemedio: prefer just `empire`. that it is a pointer is apparently clear enough from the use of `->` to access the member functions.

server/ServerFSM.cpp
```diff
@@ -1188,7 +1194,42 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
             // limit count of AI
             auto max_ai = GetOptionsDB().Get<int>("network.server.ai.max");
             if (ai_count <= max_ai || max_ai < 0) {
-                m_lobby_data->m_players    = incoming_lobby_data.m_players;
+                if (sender->HasAuthRole(Networking::ROLE_HOST)) {
+                    // don't check host
```
geoffthemedio: meaning the host can set any player to any empire, regardless of it nominally requiring an authenticated matching player name?

o01eg: Yes, I suppose Host is a `root` of the server.

geoffthemedio: please clarify this in the comment

server/ServerApp.cpp

server/ServerFSM.cpp

PR #2384, opened 2019-02-13
Topmost matches condition

agrrr3:
This PR implements a OrderedAlternativesOf (was named TopmostMatches before)  FOCS condition in order to implement the result of the  [Fighter targetting poll](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11204&p=94826#p94826).

It provides matching the matches of the topmost condition which has at least one match, so one can specify tiers/a hierarchy of preferred matches.

State of PR: works, tested fighters targeting and surrounding Or, Not, And

### DIFF
--- a/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
+++ b/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
@@ -26,7 +26,10 @@ Part
     location = OwnedBy empire = Source.Owner
     effectsgroups = [
         EffectsGroup
-            scope = [[EMPIRE_OWNED_SHIP_WITH_PART(FT_BAY_1)]]
+            scope = And [
+                Source
+                [[EMPIRE_OWNED_SHIP_WITH_PART(FT_BAY_1)]]
+            ]
             activation = Source
             stackinggroup = "INTERCEPTOR_FAST_LAUNCH_EFFECT"
             effects = [
--- a/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
+++ b/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
@@ -12,7 +12,7 @@ Part
             MaximumNumberOf number = [[COMBAT_TARGETS_SPREAD_HANGAR]]
                 sortkey = LocalCandidate.Attack
                 condition = Fighter
-            Ship
+            [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]]
         ]
     ]
     mountableSlotTypes = Internal
--- a/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
+++ b/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
@@ -7,7 +7,13 @@ Part
     damage = 1
     combatTargets = And [
         [[COMBAT_TARGETS_VISIBLE_ENEMY]]
-        Fighter
+        // Target "Bombers" first. If there are no enemy fighters present, target Ships
+        OrderedAlternativesOf [
+            MaximumNumberOf number = [[COMBAT_TARGETS_SPREAD_HANGAR]]
+                sortkey = LocalCandidate.Attack
+                condition = Fighter
+            Ship
+        ]
     ]
     mountableSlotTypes = Internal
     buildcost = 15 * [[FLEET_UPKEEP_MULTIPLICATOR]] * [[SHIP_PART_COST_MULTIPLIER]]
--- a/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
+++ b/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
@@ -8,19 +8,16 @@ Part
     combatTargets = And [
         // Consider only visible enemies as target
         [[COMBAT_TARGETS_VISIBLE_ENEMY]]
-        // Target "Bombers" first.
         OrderedAlternativesOf [
-            // Takes the N most dangerous enemy fighters as target
-            // N is the number of own interceptors multiplied by a spread multiplier
-            //  a higher spread ensures that interceptors rarely waste shots
-            //  a lower spread makes it more likely that the most dangerous enemies are taken down (while wasting more shots)
-            //  a spread of 1.2 has a good effectiveness
-            //  here we estimate the number of interceptors by the number of interceptor carriers (as you currently cant count the fighter objects)
-            //  estimating 4 interceptors per carrier usually will be too high in the last combat turn
-            //    we choose 3 instead, so at the beginning of the battle hitting the most dangerous targets for sure is emphasized, later spread is higher
-            MaximumNumberOf number = ( 1.2 * 3 * Statistic Count condition = [[OWN_INTERCEPTOR_CARRIER]] )
-                sortkey = LocalCandidate.Attack
-                condition = Fighter
+            // Target Bombers and Heavy Bombers first
+            And [
+                Fighter
+                Or [
+                   DesignHasPart name = "FT_HANGAR_3"
+                   DesignHasPart name = "FT_HANGAR_4"
+                ]
+            ]
+            Fighter
             // if no fighters: target enemy ships
             [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]]
         ]
@@ -40,33 +37,6 @@ Part
             ]
     ]
     icon = "icons/ship_parts/fighter06.png"
-
-// Trying to consider only with the same targetting conditions (i.e. launched by the same hangar type)
-// Currently this is implemented as workaround which accepts wrong values in rare cases
-// There would be a workaround to find out the relative balance of types of own hangar parts and use that for scaling
-// But probably its better to query the launching hangar part like: DesignHasPart name = CurrentContent
-// For own interceptors it is enough to check damage for a highest bound attack.
-//   fighters are always up-to-date (as they get recreated each turn)
-//   fighter damage DOES depend on species *_WEAPONS, taking 0.6 margin makes it work for most common cases
-//     GOOD FT_HANGAR_1 NO TECH           -> 1.5 Damage
-//     GRET FT_HANGAR_1 NO TECH           -> 2   Damage
-//     BAD  FT_HANGAR_2 NO TECH           -> 2   Damage
-//     AVRG FT_HANGAR_2 NO TECH           -> 3   Damage
-//
-//     GRET FT_HANGAR_1 SHP_FIGHTERS_2    -> 4   Damage
-//     BAD  FT_HANGAR_2 SHP_FIGHTERS_2    -> 4   Damage
-//     AVRG FT_HANGAR_2 SHP_FIGHTERS_2    -> 5   Damage
-//   THIS CONDITION WILL BE WRONG FOR SOME CASES (namely if you have GREAT and BAD pilots of different carrier types in the same battle)
-//   higher tech level makes this fail for less cases
-
-OWN_INTERCEPTOR_CARRIER
-'''And [ InSystem id = RootCandidate.SystemID Ship OwnedBy empire = Source.Owner DesignHasPart name = "FT_HANGAR_1" ]'''
-
-// 2019-02-27: currently will not match anything in count condition: Statistic Count condition = Fighter
-OWN_INTERCEPTOR
-'''And [ Fighter OwnedBy empire = Source.Owner DesignHasPart name "FT_HANGAR_1" ]'''
-
-
 
 #include "/scripting/common/upkeep.macros"
 #include "/scripting/techs/techs.macros"
--- a/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
+++ b/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
@@ -5,22 +5,19 @@ Part
     class = FighterHangar
     capacity = 4
     damage = 1
-    combatTargets = And [
-        // Consider only visible enemies as target
-        [[COMBAT_TARGETS_VISIBLE_ENEMY]]
-        OrderedAlternativesOf [
-            // Target Bombers and Heavy Bombers first
-            And [
-                Fighter
-                Or [
-                   DesignHasPart name = "FT_HANGAR_3"
-                   DesignHasPart name = "FT_HANGAR_4"
-                ]
-            ]
+    combatTargets = OrderedAlternativesOf [
+        // Target Bombers and Heavy Bombers first
+        And [
+            [[COMBAT_TARGETS_VISIBLE_ENEMY]]
             Fighter
-            // if no fighters: target enemy ships
-            [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]]
+            Or [
+               DesignHasPart name = "FT_HANGAR_3"
+               DesignHasPart name = "FT_HANGAR_4"
+            ]
         ]
+        And [ [[COMBAT_TARGETS_VISIBLE_ENEMY]]  Fighter ]
+        // if no fighters: target enemy ships
+        And [ [[COMBAT_TARGETS_VISIBLE_ENEMY]]  [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]] ]
     ]
     mountableSlotTypes = Internal
     buildcost = 15 * [[FLEET_UPKEEP_MULTIPLICATOR]] * [[SHIP_PART_COST_MULTIPLIER]]
--- a/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
+++ b/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
@@ -6,12 +6,22 @@ Part
     capacity = 4
     damage = 1
     combatTargets = And [
+        // Consider only visible enemies as target
         [[COMBAT_TARGETS_VISIBLE_ENEMY]]
-        // Target "Bombers" first. If there are no enemy fighters present, target Ships
+        // Target "Bombers" first.
         OrderedAlternativesOf [
-            MaximumNumberOf number = [[COMBAT_TARGETS_SPREAD_HANGAR]]
+            // Takes the N most dangerous enemy fighters as target
+            // N is the number of own interceptors multiplied by a spread multiplier
+            //  a higher spread ensures that interceptors rarely waste shots
+            //  a lower spread makes it more likely that the most dangerous enemies are taken down (while wasting more shots)
+            //  a spread of 1.2 has a good effectiveness
+            //  here we estimate the number of interceptors by the number of interceptor carriers (as you currently cant count the fighter objects)
+            //  estimating 4 interceptors per carrier usually will be too high in the last combat turn
+            //    we choose 3 instead, so at the beginning of the battle hitting the most dangerous targets for sure is emphasized, later spread is higher
+            MaximumNumberOf number = ( 1.2 * 3 * Statistic Count condition = [[OWN_INTERCEPTOR_CARRIER]] )
                 sortkey = LocalCandidate.Attack
                 condition = Fighter
+            // if no fighters: target enemy ships
             [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]]
         ]
     ]
@@ -30,6 +40,33 @@ Part
             ]
     ]
     icon = "icons/ship_parts/fighter06.png"
+
+// Trying to consider only with the same targetting conditions (i.e. launched by the same hangar type)
+// Currently this is implemented as workaround which accepts wrong values in rare cases
+// There would be a workaround to find out the relative balance of types of own hangar parts and use that for scaling
+// But probably its better to query the launching hangar part like: DesignHasPart name = CurrentContent
+// For own interceptors it is enough to check damage for a highest bound attack.
+//   fighters are always up-to-date (as they get recreated each turn)
+//   fighter damage DOES depend on species *_WEAPONS, taking 0.6 margin makes it work for most common cases
+//     GOOD FT_HANGAR_1 NO TECH           -> 1.5 Damage
+//     GRET FT_HANGAR_1 NO TECH           -> 2   Damage
+//     BAD  FT_HANGAR_2 NO TECH           -> 2   Damage
+//     AVRG FT_HANGAR_2 NO TECH           -> 3   Damage
+//
+//     GRET FT_HANGAR_1 SHP_FIGHTERS_2    -> 4   Damage
+//     BAD  FT_HANGAR_2 SHP_FIGHTERS_2    -> 4   Damage
+//     AVRG FT_HANGAR_2 SHP_FIGHTERS_2    -> 5   Damage
+//   THIS CONDITION WILL BE WRONG FOR SOME CASES (namely if you have GREAT and BAD pilots of different carrier types in the same battle)
+//   higher tech level makes this fail for less cases
+
+OWN_INTERCEPTOR_CARRIER
+'''And [ InSystem id = RootCandidate.SystemID Ship OwnedBy empire = Source.Owner DesignHasPart name = "FT_HANGAR_1" ]'''
+
+// 2019-02-27: currently will not match anything in count condition: Statistic Count condition = Fighter
+OWN_INTERCEPTOR
+'''And [ Fighter OwnedBy empire = Source.Owner DesignHasPart name "FT_HANGAR_1" ]'''
+
+
 
 #include "/scripting/common/upkeep.macros"
 #include "/scripting/techs/techs.macros"
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -10211,7 +10211,7 @@ DESC_NOT_OR_AFTER_SINGLE_OPERAND
 
 # OrderedAlternativesOf Descriptions for expressions 'OrderedAlternativesOf [ A B C ]'
 DESC_ORDERED_ALTERNATIVES_BEFORE_OPERANDS
-''' {'''
+'''that matches the first matching condition {'''
 
 DESC_ORDERED_ALTERNATIVES_BETWEEN_OPERANDS
 ''', else'''
@@ -10228,19 +10228,19 @@ DESC_ORDERED_ALTERNATIVES_AFTER_SINGLE_OPERAND
 # Expressions like  'NOT OrderedAlternativesOf [ A B C ]'
 # Note: returned matches rely on (NOT A), (NOT B).. and are handled in the descriptions of A, B.. respectively
 DESC_NOT_ORDERED_ALTERNATIVES_BEFORE_OPERANDS
-'''that matches the first matching conditions' non-matching candidates: n{'''
+'''that matches the first matching conditions' non-matching candidates: {'''
 
 DESC_NOT_ORDERED_ALTERNATIVES_BETWEEN_OPERANDS
 ''', else'''
 
 DESC_NOT_ORDERED_ALTERNATIVES_AFTER_OPERANDS
-''' }n'''
+''' }'''
 
 DESC_NOT_ORDERED_ALTERNATIVES_BEFORE_SINGLE_OPERAND
-'''n{'''
+'''{'''
 
 DESC_NOT_ORDERED_ALTERNATIVES_AFTER_SINGLE_OPERAND
-'''(iff it also has non-matches)}n'''
+'''(iff it also has non-matches)}'''
 
 
 # %1% textual description of the lower turn limit.
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -10208,6 +10208,41 @@ DESC_NOT_OR_AFTER_SINGLE_OPERAND
 '''[[DESC_AND_AFTER_SINGLE_OPERAND]]'''
 
 
+
+# OrderedAlternativesOf Descriptions for expressions 'OrderedAlternativesOf [ A B C ]'
+DESC_ORDERED_ALTERNATIVES_BEFORE_OPERANDS
+''' {'''
+
+DESC_ORDERED_ALTERNATIVES_BETWEEN_OPERANDS
+''', else'''
+
+DESC_ORDERED_ALTERNATIVES_AFTER_OPERANDS
+''' }'''
+
+DESC_ORDERED_ALTERNATIVES_BEFORE_SINGLE_OPERAND
+'''{'''
+
+DESC_ORDERED_ALTERNATIVES_AFTER_SINGLE_OPERAND
+'''(iff it also has non-matches)}'''
+
+# Expressions like  'NOT OrderedAlternativesOf [ A B C ]'
+# Note: returned matches rely on (NOT A), (NOT B).. and are handled in the descriptions of A, B.. respectively
+DESC_NOT_ORDERED_ALTERNATIVES_BEFORE_OPERANDS
+'''that matches the first matching conditions' non-matching candidates: n{'''
+
+DESC_NOT_ORDERED_ALTERNATIVES_BETWEEN_OPERANDS
+''', else'''
+
+DESC_NOT_ORDERED_ALTERNATIVES_AFTER_OPERANDS
+''' }n'''
+
+DESC_NOT_ORDERED_ALTERNATIVES_BEFORE_SINGLE_OPERAND
+'''n{'''
+
+DESC_NOT_ORDERED_ALTERNATIVES_AFTER_SINGLE_OPERAND
+'''(iff it also has non-matches)}n'''
+
+
 # %1% textual description of the lower turn limit.
 # %2% textual description of the upper turn limit.
 DESC_TURN
--- a/parse/Tokens.h
+++ b/parse/Tokens.h
@@ -521,7 +521,7 @@
     (TopPriorityEnqueuedTech)                   \
     (TopPriorityResearchableTech)               \
     (TopPriorityTransferrableTech)              \
-    (OrderedAlternativesOf)                            \
+    (OrderedAlternativesOf)                     \
     (Toxic)                                     \
     (Trade)
 
--- a/parse/Tokens.h
+++ b/parse/Tokens.h
@@ -521,6 +521,7 @@
     (TopPriorityEnqueuedTech)                   \
     (TopPriorityResearchableTech)               \
     (TopPriorityTransferrableTech)              \
+    (OrderedAlternativesOf)                            \
     (Toxic)                                     \
     (Trade)
 
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -10063,32 +10063,34 @@ void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context,
         // stop if the operand condition matches at least one item
         // if the operand condition does not match, restore the matches set by moving the temporary non-matches back
         ObjectSet temp_non_matches;
-        temp_non_matches.reserve(matches.size());
+        temp_non_matches.reserve(non_matches.size());
+        matches.reserve(std::max(matches.size(),non_matches.size()));
         // try matching operand conditions until one mateches
         // use a fresh set of possible matches
         for (auto& operand : m_operands) {
-            // move back temp_non_matches
+            // move back temp_non_matches from checking previous operand
             FCMoveContent(temp_non_matches, matches);
-            // move non matches from_matches to temp_non_matches
+
+            // try to apply the current operand
             operand->Eval(local_context, matches, temp_non_matches, MATCHES);
             if (!matches.empty()) {
-                break;
+                // Select this operand, temp_non_matches do not match
+                FCMoveContent(temp_non_matches, non_matches);
+                return;
             }
-            // Check if other LocalCandidates (in the non_matches set) match
+            // look for matches in the non_matches input set to figure out if we need to apply this operand
             operand->Eval(local_context, matches, non_matches, NON_MATCHES);
             if (!matches.empty()) {
-                // This operand should return the result, but no matches exist
-                // So all objects should go into the non_matches set
-                // move all matches to non_matches here
+                // This operand should be applied, but no matches exist in the matches innput set
+                // So all objects need to be moved into the non_matches set
                 FCMoveContent(matches, non_matches);
-                // stop finding the operand
-                // temp_non_matches must still be moved to non_matches
-                break;
+                FCMoveContent(temp_non_matches, non_matches);
+                return;
             }
         }
 
-        // non matching items were already moved from matches to temp_non_matches
-        // move non matching items from temp_non_matches to non_matches
+        // No operand condition was selected, or there are no matches. Move all objects to non_matches
+        // matches input set is already empty, else it's content would need to be moved
         FCMoveContent(temp_non_matches, non_matches);
     }
 }
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -10075,7 +10075,8 @@ void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context,
                 FCMoveContent(temp_non_matches, non_matches);
                 return;
             }
-            // Select this operand if there are matching objects in the non_matches input set. 
+            // Select this operand if there are matching objects in the non_matches input set.
+            // Note: the empty matches set is used to temporarily store matches from the non_matches input set
             operand->Eval(local_context, matches, non_matches, NON_MATCHES);
             if (!matches.empty()) {
                 // Select and apply this operand. But no matching objects exist in the matches input set,
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -5111,20 +5111,15 @@ namespace {
             if (!candidate)
                 return false;
 
-            auto fighter = std::dynamic_pointer_cast<const ::Fighter>(candidate);
             const ShipDesign* design = nullptr;
-            if (fighter) {
+            if (auto fighter = std::dynamic_pointer_cast<const ::Fighter>(candidate)) {
                 // it is a fighter
-                auto carriers = Objects().FindObjects<Ship>( (std::vector<int> ) { fighter->LaunchedFrom() } );
-                if (carriers.empty())
-                    return false;
-                design = carriers.front()->Design();
-            } else {
-                // is it a ship?
-                auto ship = std::dynamic_pointer_cast<const ::Ship>(candidate);
-                if (!ship)
-                    return false;
+                design = Objects().Object<Ship>(fighter->LaunchedFrom())->Design();
+            } else if (auto ship = std::dynamic_pointer_cast<const ::Ship>(candidate)) {
+                // ir is a ship
                 design = ship->Design();
+            } else {
+                return false;
             }
             // with a valid design?
             if (!design)
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -9964,10 +9964,12 @@ unsigned int Not::GetCheckSum() const {
     TraceLogger() << "GetCheckSum(Not): retval: " << retval;
     return retval;
 }
+
 ///////////////////////////////////////////////////////////
 // OrderedAlternativesOf
 ///////////////////////////////////////////////////////////
-OrderedAlternativesOf::OrderedAlternativesOf(std::vector<std::unique_ptr<ConditionBase>>&& operands) :
+OrderedAlternativesOf::OrderedAlternativesOf(
+    std::vector<std::unique_ptr<ConditionBase>>&& operands) :
     ConditionBase(),
     m_operands(std::move(operands))
 {}
@@ -9989,8 +9991,9 @@ bool OrderedAlternativesOf::operator==(const ConditionBase& rhs) const {
     return true;
 }
 
-void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context, ObjectSet& matches,
-               ObjectSet& non_matches, SearchDomain search_domain/* = NON_MATCHES*/) const
+void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context,
+                                 ObjectSet& matches, ObjectSet& non_matches,
+                                 SearchDomain search_domain/* = NON_MATCHES*/) const
 {
     std::shared_ptr<const UniverseObject> no_object;
     ScriptingContext local_context(parent_context, no_object);
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -9969,11 +9969,8 @@ unsigned int Not::GetCheckSum() const {
 // OrderedAlternativesOf
 ///////////////////////////////////////////////////////////
 void FCMoveContent(Condition::ObjectSet& from_set, Condition::ObjectSet& to_set) {
-    for (auto it = from_set.begin(); it != from_set.end(); ) {
-        to_set.push_back(*it);
-        *it = from_set.back();
-        from_set.pop_back();
-    }
+    to_set.insert(to_set.end(), from_set.begin(), from_set.end());
+    from_set.clear();
 }
 
 OrderedAlternativesOf::OrderedAlternativesOf(
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -10028,36 +10028,36 @@ void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context,
     // or if no candidate matches A or B but any candidate matches C, then all candidates that do not match C are matched.
     // If no candidate matches A, B, or C, then all candidates are matched.
     if (search_domain == NON_MATCHES) {
-        // check all operand conditions on all objects in the non matches set
-        // stop if the operand condition has at least one match
-        // Return non-matches if there are any
+        // check each operand condition on objects in the input matches and non_matches sets, until an operand condition matches an object.
+        // if an operand condition is selected, apply it to the input non_matches set, moving matching candidates to matches.
+        // if no operand condition is selected, because no candidate is matched by any operand condition, then do nothing.
         ObjectSet temp_objects;
         temp_objects.reserve(std::max(matches.size(),non_matches.size()));
         // try matching operand conditions until one matches
         for (auto& operand : m_operands) {
-            // recreate state before checking last operand. Do it here to save moves if no operand gets selected
-            FCMoveContent(temp_objects, matches);
-
             operand->Eval(local_context, temp_objects, non_matches, NON_MATCHES);
             if (!temp_objects.empty()) {
-                // have selected the current operand condition to use, so apply it to the NON_MATCHES candidate set.
-                // in this case we alread did the application, so we use the results
+                // Select the current operand condition. Apply it to the NON_MATCHES candidate set.
+                // We alread did the application, so we use the results
                 FCMoveContent(temp_objects, matches);
                 return;
             }
-            // Also have to look at the other local candidates if there was a match
+            // Check if the operand condition matches an object in the other input set
             operand->Eval(local_context, matches, temp_objects, MATCHES);
             if (!matches.empty()) {
-                // have selected the current operand condition to use, so apply it to the NON_MATCHES candidate set.
-                // We already did the application before, but there were no matches. Move all objects to non_matches.
-                break;
+                // Select the current operand condition. Apply it to the NON_MATCHES candidate set.
+                // We already did the application before, but there were no matches.
+                // restore state before checking the operand.
+                FCMoveContent(temp_objects, matches);
+                return;
             }
 
+            // restore state before checking the operand.
+            FCMoveContent(temp_objects, matches);
             // try the next operand
         }
-        // No operand condition was selected, or there are no matches. Move all objects to non_matches
-        FCMoveContent(matches, non_matches);
-        FCMoveContent(temp_objects, non_matches);
+
+        // No operand condition was selected, nothing changed, done.
     } else /*(search_domain == MATCHES)*/ {
         // check all operand conditions on all objects in the matches set, collecting non matches temporarily
         // stop if the operand condition matches at least one item
@@ -10068,9 +10068,6 @@ void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context,
         // try matching operand conditions until one mateches
         // use a fresh set of possible matches
         for (auto& operand : m_operands) {
-            // move back temp_non_matches from checking previous operand
-            FCMoveContent(temp_non_matches, matches);
-
             // try to apply the current operand
             operand->Eval(local_context, matches, temp_non_matches, MATCHES);
             if (!matches.empty()) {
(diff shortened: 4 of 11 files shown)


default/stringtables/en.txt
```diff
@@ -10208,6 +10208,28 @@ DESC_NOT_OR_AFTER_SINGLE_OPERAND
...
+
+DESC_TOPMOST_AFTER_OPERANDS
+''' }'''
+
+DESC_TOPMOST_BEFORE_SINGLE_OPERAND
+''''''
+
+DESC_TOPMOST_AFTER_SINGLE_OPERAND
+''''''
+# Not topmost expressions like  'NOT TopmostMatches [ A B C ]'
```
agrrr3: descriptions untested, also check description after NOT TopmostMatches works

universe/Condition.cpp
```diff
@@ -9955,6 +9955,203 @@ unsigned int Not::GetCheckSum() const {
...
+std::string TopmostMatches::Dump(unsigned short ntabs) const {
+    std::string retval = DumpIndent(ntabs) + "TopmostMatches [\n";
+    for (auto& operand : m_operands)
+        retval += operand->Dump(ntabs+1);
+    retval += DumpIndent(ntabs) + "]\n";
+    return retval;
+}
+
+void TopmostMatches::GetDefaultInitialCandidateObjects(const ScriptingContext& parent_context, ObjectSet& condition_non_targets) const {
+    // XXX unsure what this does
```
agrrr3: need to understand if GetDefaultInitialCandidateObjects is correct, then remove comment

universe/Condition.cpp
```diff
@@ -9955,6 +9955,203 @@ unsigned int Not::GetCheckSum() const {
...
+    if (!m_operands.empty()) {
+        m_operands[0]->GetDefaultInitialCandidateObjects(parent_context, condition_non_targets); // gets condition_non_targets from first operand condition
+    } else {
+        ConditionBase::GetDefaultInitialCandidateObjects(parent_context, condition_non_targets);
+    }
+}
+
+void TopmostMatches::SetTopLevelContent(const std::string& content_name) {
+    // XXX unsure what this does
```
agrrr3: need to understand if SetTopLevelContent is right, then remove comment

universe/Condition.cpp
```diff
@@ -9955,6 +9955,203 @@ unsigned int Not::GetCheckSum() const {
...
+    for (auto& operand : m_operands) {
+        if (!operand) {
+            ErrorLogger() << "TopmostMatches::Eval given null operand!";
+            return;
+        }
+    }
+
+    // probably i need to figure out which condition matches first
+    // by searching both domains
+    // ... or actually i can propagate the search domain and look for the first match, probably i can
```
agrrr3: implementation does only work for TopMostMatches with a single operand (the Not case is completely untested)

default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
```diff
@@ -7,7 +7,15 @@ Part
     damage = 1
     combatTargets = And [
         [[COMBAT_TARGETS_VISIBLE_ENEMY]]
-        Fighter
+        TopmostMatches [
+            And [
```
agrrr3: This is intended to match bombers first no other fighters. Not how to determine that the target is a bomber. Maybe hardcoded attack value is ok (but will be wrong with tech advances). I could try to somehow look up the design of the ship the fighter started from and see if it contains the right hangar parts. Or somehow tag the bomber fighters when launching in cpp. Or it could try to figure out what the highest fighter Attack value is and target those first

geoffthemedio:
It's probably possible to do what you want using the existing SortedNumberOf condition, with a suitable sort-key that ranks objects based on their type.

universe/Condition.cpp
geoffthemedio: `SetTopLevelContent` is used to set `m_top_level_content`, which is declared here:

https://github.com/freeorion/freeorion/blob/master/universe/ValueRef.h#L207

and which is used here:

https://github.com/freeorion/freeorion/blob/master/universe/ValueRef.cpp#L552

to return the top level bit of content in which this ValueRef is scripted. For example, in a species script, it would be the name of the species, returned as a `ValueRef::Constant<std::string>` when `ThisSpecies` is written into the script. This is helpful for writing generic macros that can be used in different contexts.

All the Effects, Conditions, and ValueRefs all have `SetTopLevelContent` functions, as they can contain eachother recursively and the final leaf ValueRef needs to have its top level content set by its enclosing object.

default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
geoffthemedio: "Bombers" are fighters with damage > 1

agrrr3: damage > 1 includes "Fighters"

agrrr3: and also includes "Interceptors" after the first tech is researched (and probably(?) good weapon species interceptors are also damage > 1)

geoffthemedio: OK, I thought they just got more shots. Apparently not, though. But in that case, if interceptors only target fighters, then the extra damage is pretty useless.

universe/Condition.cpp
agrrr3: Thanks for the explanation

default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
agrrr3: Yes, more shots/hangar capacity makes more sense for interceptors. Maybe i could take N most dangerous enemy fighters, where N is the count of own interceptors * 1.2

agrrr3:
Sort on LocalCandidate.Attack did not seem to work for ships as sortkey. LocalCandidate.Structure did though. I did not check if LocalCandidate.Attack works for fighters.

I probably would need a composite search-key or nested SortedNumberOf conditions to what i want.
It would be much better if i could make TopmostMatches work.

default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
agrrr3: Implemented a spread macro for targeting  Still figuring out the numbers, but 1.2 seems ok as heuristic. Could be adapted to species traits WIDE_TARGET_SPREAD for laid back or similar NARROW_TARGET_SPREAD for neurotic species.

universe/Condition.cpp
agrrr3: Works, i simply instantiated the wrong condition on parsing (d'oh)

default/stringtables/en.txt
agrrr3: tested all four description cases

geoffthemedio:
Could rename this `AlternativeOr`? Definitely shouldn't have "matches" in the name.

universe/Condition.cpp
geoffthemedio: `GetDefaultInitialCandidateObjects` narrows the search space of objects when no list of candidates is provided.

If a single object or a containers of matches and non-matches are passed to `Eval`, then just those objects are tested by the condition. But there is an `Eval` overload that just takes a set of matches, which it immediately clears, and then fills with any objects in the universe that the condition matches.

To make this faster, conditions have the `GetDefaultInitialCandidateObjects` function to initialize the search space of objects to test, rather than always testing all objects in the `Universe`. For example, `Source` initializes with just the `ScriptingContext`'s source object, and `Type` will initialize based on what object type is being matched (eg. ships for `OBJ_SHIP`).

For this case, it's acting like an `Or` condition, and thus shouldn't narrow the search space at all, so like `Or` you shouldn't overload `GetDefaultInitialCandidateObjects` and instead just use the base class behaviour.

geoffthemedio:
Since you don't want to match a particular number of objects, a separate condition is reasonable in this case, I think.

`SortedNumberOf` could be set up to match unlimited objects, and then you could probably set up the sort key to use the `If` statistic to test whether anything matches various subconditions, but it would be relatively awkward and complicated.

universe/Condition.cpp
```diff
@@ -9955,6 +9955,203 @@ unsigned int Not::GetCheckSum() const {
...
+    // Expressions like  'NOT TopmostMatches [ A B C ]' matches all candidates which do not match the topmost matching condition
+    // which means the same as TopmostMatches [ (Not A) (Not B) (Not C) ]
```
geoffthemedio: If the candidates set is `{ship 1, planet 2, building 3}`...

If the test is `Not TopmostMatches [All Ship]`, it should match nothing.

If the test is `TopmostMatches [Not All Not Ship]`, it should match `{planet 2, building 3}`.

Those aren't the same...

universe/Condition.cpp
```diff
@@ -9955,6 +9955,203 @@ unsigned int Not::GetCheckSum() const {
...
+        }
+
+        // operand matching items were already moved from non_matches to temp_matches
+        // move operand matching items from temp_matches to matches
+        matches.insert(matches.end(),
+                       std::make_move_iterator(temp_matches.begin()),
+                       std::make_move_iterator(temp_matches.end()));
+    } else /*(search_domain == MATCHES)*/ {
+        // check all operand conditions on all objects in the matches set
+        // stop if the operand condition matches at least one item
```
geoffthemedio: Don't think that makes sense... You'd need to check if anything in the matches or non_matches is matched by the current subcondition to decide which subcondition to use, then go back and move stuff out of matches that doesn't match the selected subcondition. You can't just test the matches to pick which subcondition to apply to matches.

universe/Condition.cpp
```diff
@@ -9955,6 +9955,203 @@ unsigned int Not::GetCheckSum() const {
...
+
+    // probably i need to figure out which condition matches first
+    // by searching both domains
+    // ... or actually i can propagate the search domain and look for the first match, probably i can
+    // Expressions like  'NOT TopmostMatches [ A B C ]' matches all candidates which do not match the topmost matching condition
+    // which means the same as TopmostMatches [ (Not A) (Not B) (Not C) ]
+    // TODO use logic like in base eval for both domains
+    if (search_domain == NON_MATCHES) {
+        // check all operand conditions on all objects in the non matches set
+        // stop if the operand condition has at least one non match
```
geoffthemedio: Similarly here, I think that if you have `Or [Ship TopmostMatches [Ship Planet]]` and you apply it to an initial input set non_matches containing `{ship 1, planet 2}`, it should match just `{ship 1}`...?

universe/Condition.cpp
agrrr3: Yeah you are right. I think NOT TopmostMatches [ A B C ] in FOCS terms should be the same as
```
TopmostMatches [ 
    And [
        Number low = 1 condition = A
        Not A
    ]
    And [
        Number low = 1 condition = B
        Not B
    ]
    And [
        Number low = 1 condition = C
        Not C
    ]
```
In the description, which is usually used to describe a mismatch it should probably read like:
```{ <<Not A>> iff there is at least one opposite match, else <<Not B>> iff there is at least one opposite match, else <<Not A>> iff there is at least one opposite match }```

or something like:
```matches the first conditions which has matching and non-matching candidates: { <<Not A>>, else <<Not B>>, else <<Not C>> }```

File: universe/Condition.cpp (hunk 7 above)
Comment (agrrr3) [reply universe/Condition.cpp]: I tried to make the documentation better for the MATCHES implementation

File: universe/Condition.cpp (hunk 8 above)
Comment (agrrr3) [reply universe/Condition.cpp]: Yes, that need fixing

**Comment by agrrr3** [comment]
Besides being the FOCS code becoming more convoluted, I also do not want to use the Sorted* conditions because it actually needs to sort (which is much more expensive).

> Could rename this `AlternativeOr`? Definitely shouldn't have "matches" in the name.

Not sure if `AlternativeOr [ A B C ]` and `Not AlternativeOr [ A B ]` are helpful.

I am still looking for a better name. ExclusiveOr is not good, because it actually is different from the established XOR/XNOR. 
TopOr, FirstOr, TopTier, ConditionTiers, TopCondition, TopmostCondition, TopNonEmpty... 
Is there a term for a set which has at least one member?

File: universe/Condition.cpp (hunk 6 above)
Comment (geoffthemedio) [reply universe/Condition.cpp]: Keep in mind that `Number` tests its subcondition on all objects in the universe, not (just) those in the candidate `matches` and `non_matches` sets.

https://github.com/freeorion/freeorion/blob/master/universe/Condition.cpp#L415

**Comment by geoffthemedio** [comment]
Maybe `OrderedPreferenceOf` or `FirstOf`, by analogy to `NumberOf`?

The meaning isn't going to be really clear from any reasonably-not-long name, I think. But having "matches" or "condition" in the name isn't good.

File: universe/Condition.cpp (hunk 6 above)
Comment (agrrr3) [reply universe/Condition.cpp]: So that woiuld mean rather ```Number low = 1 condition = And [ Object id = RootCandidate.ID  A ]```?

Hm. Not sure RootCandidate is available though.

Comment (agrrr3) [reply universe/Condition.cpp]: Would Filtering for Object I'd  = LocalCandidate.ID inside the Number condition suffice?

Comment (agrrr3) [reply universe/Condition.cpp]: Would Filtering for Object I'd  = LocalCandidate.ID inside the Number condition suffice?

Comment (agrrr3) [reply universe/Condition.cpp]: Would Filtering for Object I'd  = LocalCandidate.ID inside the Number condition suffice?

Comment (geoffthemedio) [reply universe/Condition.cpp]: I don't think that makes sense, and it's probably not supported. `LocalCandidate` within a subcondition of `Number` would be the objects being matched by that condition, and has nothing to do with objects being matched by `Number` itself. `Number` doesn't even pass a `LocalCandidate` to its subcondition. In some cases you might be able to do something with `RootCandidate`, but I don't advise it.

Does this matter, though? I'm not suggesting you should implement this condition using a composition of the existing conditions.

**Comment by agrrr3** [comment]
> Maybe `OrderedPreferenceOf` or `FirstOf`, by analogy to `NumberOf`?
> 
> The meaning isn't going to be really clear from any reasonably-not-long name, I think. But having "matches" or "condition" in the name isn't good.

I think I like `OrderedPreferenceOf` best yet. `Not OrderedPreferenceOf` sounds complicated enough so one does not assume some simple condition. `Not FirstOf` sounds like it could have a lot of wrong interpretations.

File: universe/Condition.cpp (hunk 6 above)
Comment (agrrr3) [reply universe/Condition.cpp]: No, you are right, it doesnt matter for the implementation. I think the implementaton for `matches` and `non_matches` will be quite similar in the end and not depend on other conditions.

But i think it would be helpful for documentation purposes to express it in FOCS terms.

So I probably would need a hypethetical `OuterCandidate` which allows to pump the matches of the surrounding (so outta Number in this case) into the inner condition
```
TopmostMatches [ 
    And [
        Number low = 1 condition = And [ 
            Object id = OuterCandidate.ID
            A
        ]
        Not A
    ]
```

File: universe/Condition.cpp (hunk 7 above)
Comment (agrrr3) [reply universe/Condition.cpp]: NON_MATCHES and MATCHES domain work now

File: universe/Condition.cpp (hunk 8 above)
Comment (agrrr3) [reply universe/Condition.cpp]: fixed

**Comment by agrrr3** [comment]
Ok it finally became the OrderedAlternativesOf condition. 
Branch is ready for review, i think this will spur more changing after playtesting.
In the end, AI needs some heuristics or simulator to estimate combat strength / build better fleets.

File: default/stringtables/en.txt [hunk 9]
```diff
@@ -10208,6 +10208,41 @@ DESC_NOT_OR_AFTER_SINGLE_OPERAND
...
+DESC_ORDERED_ALTERNATIVES_BEFORE_SINGLE_OPERAND
+'''{'''
+
+DESC_ORDERED_ALTERNATIVES_AFTER_SINGLE_OPERAND
+'''(iff it also has non-matches)}'''
+
+# Expressions like  'NOT OrderedAlternativesOf [ A B C ]'
+# Note: returned matches rely on (NOT A), (NOT B).. and are handled in the descriptions of A, B.. respectively
+DESC_NOT_ORDERED_ALTERNATIVES_BEFORE_OPERANDS
+'''that matches the first matching conditions' non-matching candidates: n{'''
```
geoffthemedio: `n` ?

parse/Tokens.h
```diff
@@ -521,6 +521,7 @@
     (TopPriorityEnqueuedTech)                   \
     (TopPriorityResearchableTech)               \
     (TopPriorityTransferrableTech)              \
+    (OrderedAlternativesOf)                            \
```
geoffthemedio: remove extra spaces so `\` line up

default/stringtables/en.txt
agrrr3: leftover from testing

agrrr3:
> Need to explain this sorting condition.

When having a look at finding a better explanation i tried to make it more restrictive like i originally wanted.
Found a few issues when testing though. I think the MaximumNumberOf Fighters does not work inside 
OrderedAlternativesOf. I am pretty sure i tested this outside of OrderedAlternativesOf. As a heads up - I will implement DesignHasPart for fighters (which will match if the launching ship's design has that part) as well in this branch. If MaximumNumberOf works outside of OrderedAlternativesOf i probably need to switch to DesignHasPart anyway.
(Maybe also PartsInShipDesign valueref)

agrrr3:
Ok, what does not work is the Count Statistic. I guess it initialises its context in a way which does not include Fighters.

So e.g. this fails: MaximumNumberOf number = ( 1.2 * Statistic Count condition = And [ Fighter OwnedBy empire = Source.Owner ] )
while this works:
MaximumNumberOf number = ( 0.2 * Statistic Count condition = And [ Ship OwnedBy empire = Source.Owner ] )

I think i will open an issue for this and as i dont have a sensible way to find out how many interceptors the shooter has, i will target only bombers and not use the numberOf condition.

universe/Condition.cpp
```diff
@@ -5110,12 +5111,22 @@ namespace {
...
 
-            // is it a ship?
-            auto ship = std::dynamic_pointer_cast<const ::Ship>(candidate);
-            if (!ship)
-                return false;
+            auto fighter = std::dynamic_pointer_cast<const ::Fighter>(candidate);
+            const ShipDesign* design = nullptr;
+            if (fighter) {
+                // it is a fighter
+                auto carriers = Objects().FindObjects<Ship>( (std::vector<int> ) { fighter->LaunchedFrom() } );
```
geoffthemedio: Just use `Objects().Object<Ship>(fighter->LaunchedFrom())` to get a single ship.

universe/Condition.h
```diff
@@ -1964,6 +1964,30 @@ struct FO_COMMON_API Not final : public ConditionBase {
     void serialize(Archive& ar, const unsigned int version);
 };
 
+/** Matches all objects that match the first matching Condition in \a operands. */
```
geoffthemedio: Maybe: Tests conditions in \a operands in order, to find the first condition that matches at least one candidate object. Matches candidate objects that match that condition, ignoring any conditions listed later. If no candidate matches any of the conditions, returns nothing.

universe/Condition.cpp
```diff
@@ -5110,12 +5111,22 @@ namespace {
             if (!candidate)
                 return false;
 
-            // is it a ship?
-            auto ship = std::dynamic_pointer_cast<const ::Ship>(candidate);
-            if (!ship)
-                return false;
+            auto fighter = std::dynamic_pointer_cast<const ::Fighter>(candidate);
+            const ShipDesign* design = nullptr;
+            if (fighter) {
```
geoffthemedio: could put the `auto fighter = ...` into the `( )` here.

universe/Condition.cpp
```diff
@@ -5110,12 +5111,22 @@ namespace {
...
+                // it is a fighter
+                auto carriers = Objects().FindObjects<Ship>( (std::vector<int> ) { fighter->LaunchedFrom() } );
+                if (carriers.empty())
+                    return false;
+                design = carriers.front()->Design();
+            } else {
```
geoffthemedio, line 5118: and make this `else if (auto ship = ...)`

and have `else { return; }` at end.

default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
geoffthemedio: I think this is better as written now, with the simpler condition. The 1.2 factor and taking N most is overly complicated when written with the subcondition.

If you wanted, though, you could extended `OrderedAlternativesOf` to take an optional `number` or `atleast` parameter, which indicates the number of matches that need to have been found to stop trying more alternate conditions. Not sure if that should be adding more objects to the matches from each successively tested subcondition, or only matching with a subcondition if at least the minimum objects match the subcondition, though.

geoffthemedio:
> Ok, what does not work is the Count Statistic. I guess it initialises its context in a way which does not include Fighters.

That's accurate, as far as I'm aware. Subcondition candidates within a Statistic don't depend on the objects being matched by a condition that contains the Statistic, so initialized using (an appropriate subset of) the main game universe, which doesn't have fighters, not the combat objectset, which does. The objects actually being matched do come from the combat objectset, so include fighters, even though the statistic subcondition candidates don't. Confusing and inconsistent, unfortunately... I'm not super keen on also including a "full universe objects for purposes of evaluating this condition/valueref" parameter in the context, though...

But as mentioned in a resolved conversation reply above, I don't think the complicated numberof subcondition is necessary anyway. The simpler ordered alternatives of subconditions is fine...

universe/Condition.cpp
```diff
@@ -5111,20 +5111,15 @@ namespace {
...
-                auto carriers = Objects().FindObjects<Ship>( (std::vector<int> ) { fighter->LaunchedFrom() } );
-                if (carriers.empty())
-                    return false;
-                design = carriers.front()->Design();
-            } else {
-                // is it a ship?
-                auto ship = std::dynamic_pointer_cast<const ::Ship>(candidate);
-                if (!ship)
-                    return false;
+                design = Objects().Object<Ship>(fighter->LaunchedFrom())->Design();
```
geoffthemedio: should probably not assume the `Ship` pointer is valid without checking...

universe/Condition.cpp
```diff
@@ -5111,17 +5111,20 @@ namespace {
...
 
-            const ShipDesign* design = nullptr;
+            std::shared_ptr<const Ship> ship = nullptr;
             if (auto fighter = std::dynamic_pointer_cast<const ::Fighter>(candidate)) {
                 // it is a fighter
-                design = Objects().Object<Ship>(fighter->LaunchedFrom())->Design();
-            } else if (auto ship = std::dynamic_pointer_cast<const ::Ship>(candidate)) {
-                // ir is a ship
-                design = ship->Design();
+                ship = std::move(Objects().Object<Ship>(fighter->LaunchedFrom()));
```
geoffthemedio: don't think `std::move` makes sense here...

agrrr3: cause ... its already an rvalue?

geoffthemedio: `ObjectMap::Object<T>` returns a `shared_ptr<T>` by value, and that value doesn't have a name here, and it falls out of scope immediately after this line, so yes, I think?

agrrr3: I guess the same reasoning applies to the `std::dynamic_pointer_cast<const ::Ship>(candidate)`.
The candidate is an lvalue, so I exptect the cast will return a new shared_ptr.

agrrr3: If I get it right if `candidate` would be a rvalue the cast would  actually move starting with c++20 and using candidate after that line would become illegal.

geoffthemedio:
I checked out this branch, built it, then added a custom sitrep effectsgroup to test the condition in default/customizations/custom_sitreps.txt:
````
EffectsGroup
    scope = Source
    activation = Source
    effects =
        GenerateSitRepMessage
            message = "Matches: %rawtext:number%   Non-Matches: %rawtext:number2%   All: %rawtext:number3%"
            label = "CUSTOM_2"
            NoStringtableLookup
            icon = "icons/building/gas-giant-generator.png"
            parameters = [
                tag = "number" data = Statistic Count condition =
                    OrderedAlternativesOf [
                        Ship
                        Building
                        Planet
                        All
                    ]                
                tag = "number2" data = Statistic Count condition =
                    Not OrderedAlternativesOf [
                        Ship
                        Building
                        Planet
                        All
                    ]
                tag = "number3" data = Statistic Count condition =
                    All
            ]
            empire = Source.Owner
````
It doesn't appear to work correctly, though... Sitreps from before and after scrapping all my ships:
orderalternativesof sitrep test

File: universe/Condition.cpp [hunk 17]
```diff
@@ -9955,6 +9964,249 @@ unsigned int Not::GetCheckSum() const {
     TraceLogger() << "GetCheckSum(Not): retval: " << retval;
     return retval;
 }
+///////////////////////////////////////////////////////////
```
geoffthemedio, line 9968: add blank line

universe/Condition.cpp
```diff
@@ -9955,6 +9964,249 @@ unsigned int Not::GetCheckSum() const {
...
+// OrderedAlternativesOf
+///////////////////////////////////////////////////////////
+OrderedAlternativesOf::OrderedAlternativesOf(std::vector<std::unique_ptr<ConditionBase>>&& operands) :
```
geoffthemedio: move parameter to newline; it's too long to read easily placed there

universe/Condition.cpp
```diff
@@ -9955,6 +9964,249 @@ unsigned int Not::GetCheckSum() const {
...
+        return false;
+    for (unsigned int i = 0; i < m_operands.size(); ++i) {
+        CHECK_COND_VREF_MEMBER(m_operands.at(i))
+    }
+
+    return true;
+}
+
+void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context, ObjectSet& matches,
+               ObjectSet& non_matches, SearchDomain search_domain/* = NON_MATCHES*/) const
```
geoffthemedio: indent to match `const`

agrrr3:
> I checked out this branch, built it, then added a custom sitrep effectsgroup to test the condition in default/customizations/custom_sitreps.txt:
> 
> ```
> EffectsGroup
>     scope = Source
>     activation = Source
>     effects =
>         GenerateSitRepMessage
>             message = "Matches: %rawtext:number%   Non-Matches: %rawtext:number2%   All: %rawtext:number3%"
>             label = "CUSTOM_2"
>             NoStringtableLookup
>             icon = "icons/building/gas-giant-generator.png"
>             parameters = [
>                 tag = "number" data = Statistic Count condition =
>                     OrderedAlternativesOf [
>                         Ship
>                         Building
>                         Planet
>                         All
>                     ]                
>                 tag = "number2" data = Statistic Count condition =
>                     Not OrderedAlternativesOf [
>                         Ship
>                         Building
>                         Planet
>                         All
>                     ]
>                 tag = "number3" data = Statistic Count condition =
>                     All
>             ]
>             empire = Source.Owner
> ```
> 
> It doesn't appear to work correctly, though... Sitreps from before and after scrapping all my ships:
> orderalternativesof sitrep test

Love the sitrep effectsgroup :1st_place_medal:  . The negative version is not working correctly obviously :(

I guess it is failing in the buildings case after finding no ships.
The missing number between not-matches and all objects i guess is the number of buildings.

If I do not make the negative version working today i would prefer this to be merged anyway (i can not imagine a use of the negative version anyway) so we can get feedback on the fighter conditions. I would open an issue that the negative version is broken in that case.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,252 @@ unsigned int Not::GetCheckSum() const {
...
+            // move those non matching the current operand to temp_non_matches
+            //            operand->Eval(local_context, matches, temp_non_matches, MATCHES);
+            // move those matching the current non-negated operand from non_matches to temp_non_matches
+            operand->Eval(local_context, non_matches, temp_non_matches, MATCHES);
+            if (!(non_matches.empty())) {
+                // i.e. the current operand should be used negated to move 
+                found_sth = true;
+                // move back temp_non matches
+                non_matches.insert(non_matches.end(),
+                               std::make_move_iterator(temp_non_matches.begin()),
```
geoffthemedio: indent

universe/Condition.cpp
```diff
@@ -9956,6 +9965,252 @@ unsigned int Not::GetCheckSum() const {
...
+        ErrorLogger() << "OrderedAlternativesOf::Eval given no operands!";
+        return;
+    }
+    for (auto& operand : m_operands) {
+        if (!operand) {
+            ErrorLogger() << "OrderedAlternativesOf::Eval given null operand!";
+            return;
+        }
+    }
+
```
geoffthemedio, line 10018: Before explaining the Not case, should probably explain the normal case...

For OrderedAlternativesOf [ A B C ], if any candidate matches A, then all candidates that match A are matched, or if no candidate matches A but any candidate matches B, then all candidates that match B are matched, or if no candidate matches A or B but any candidate matches C, then all candidates that match C are matched. If no candidate matches A, B, or C, then nothing is matched.

Then for the Not case...

For Not OrderedAlternativesOf [ A B C ], if any candidate matches A, then all candidates that do not match A are matched, or if no candidates match A but any candidate matches B, then all candidates that do not match B are matched, or if no candidate matches A or B but any candidate matches C, then all candidates that do not match C are matched. If no candidate matches A, B, or C, then all candidates are matched.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,252 @@ unsigned int Not::GetCheckSum() const {
...
+    //     And [
+    //         Number low = 1 condition = And [
+    //             Object id = OuterCandidate.ID
+    //             C
+    //         ]
+    //         Not C
+    //     ]
+    // ]
+    if (search_domain == NON_MATCHES) {
+        // check all operand conditions on all objects in the non matches set
```
geoffthemedio: Need to check candidates in matches and non_matches, no just in matches, to determine which subcondition to use. This should be the same regardless of the `search_domain`.

agrrr3: Not sure what you refer to. But i already did in both cases. In the non_matches case i only have to restore the original state after figuring out the first matching operand condition

agrrr3: probably i need to reread your comment and rethink

agrrr3:
> Love the sitrep effectsgroup 1st_place_medal . The negative version is not working correctly obviously :(
It also failed in the positive case. But the fighters combatTargets were working..

So i figured out the positive matches are not always correctly moved. If i added a `Or [ Ship Building Planet]` in your count condition in front of OrderedAlternativesOf the positive count was correct instead of zero.
As I already had one problem where the insert with move iterator did not seem to work, I use now a for loop and push and pop everywhere and the positive case works again :)

universe/Condition.cpp
agrrr3: Ok. So what would be a testcase?
```
And [
    Not Ship
    OrderedAlternativesOf [
        Ship
        Building
    ]
]
```
Hm. Lets say there are buildings and ships in the candidates. So you would say the Ship operand condition of OrderedAlternativesOf should be chosen for evaluation? So in such a case the And condition will always match nothing. 

Or should the OrderedAlternativesOf be context sensitive and fail the Ship operand condition, moving further to evaluate the Building condition? I think I would actually prefer this behavior.

geoffthemedio: > ... the Ship operand condition of OrderedAlternativesOf should be chosen for evaluation?

Yes; the result of `And [A B]` should be the same as the result of `And [B A]`.

But more so, conditions like `And` can pass candidate sets as `matches` or `non_matches` and adjust `search_domain` as an implementation / optimization detail. The subcondition (in this case `OrderedAlternativesOf`) should match each candidate in the candidate set corresponding to the `search_domain` with the same criteria, regardless of which `search_domain` is specified.

For your example, you could rewrite it
````
OrderedAlternativesOf [
    And [Ship Not Ship]
    Building
]
````
Or just `Building`

Comment (agrrr3) [reply universe/Condition.cpp]: Ok, I will see that non_matches and matches get searched. I think I have  probably can skip the matching generation in the other case (e.g. when searching non_matches in the MATCHES domain).

Correct me, but this means that for deciding the alternative, all LocalCandidates are used, no matter what the other conditions do, right?

Concerning the example, of course using Ship and Not Ship does not make sense but it could make sense to have different conditions which sometimes leads to a similar situation. (So the just `Building` rewrite would miss the point.)

I guess a more exact rewrite for the generic case would be
```
OrderedAlternativesOf [
    And [ Ship [[MAYBE_NOT_SHIP]] ]
    And [ Building [[MAYBE_NOT_SHIP]] ]
]
```

**Comment by agrrr3** [comment]
NON_MATCHES case also works now, tested with your example and my test case.

I made a free helper function for moving entries of the sets. I am pretty sure that could be bad style, if so please give me a better suggestion.

File: universe/Condition.cpp (hunk 22 above)
Comment (agrrr3) [reply universe/Condition.cpp]: Meh, so this caused the combatTarget conditions to fail, they get pretty bloated now.

File: universe/Condition.cpp [hunk 23]
```diff
@@ -9968,6 +9968,14 @@ unsigned int Not::GetCheckSum() const {
...
 // OrderedAlternativesOf
 ///////////////////////////////////////////////////////////
+void FCMoveContent(Condition::ObjectSet& from_set, Condition::ObjectSet& to_set) {
+    for (auto it = from_set.begin(); it != from_set.end(); ) {
+        to_set.push_back(*it);
+        *it = from_set.back();
+        from_set.pop_back();
+    }
+}
+
```
geoffthemedio, line 9977: Something like https://github.com/freeorion/freeorion/blob/master/universe/Condition.cpp#L423 doesn't work?

agrrr3: it works and (Un)surprisingly also the move iterator does work now... must have been something else before

universe/Condition.cpp
geoffthemedio: > Meh, so this caused the combatTarget conditions to fail

Did you fix that?

agrrr3: Yes, fixed. I am not too happy with the situation. 
```
    combatTargets = OrderedAlternativesOf [
        // Target Bombers and Heavy Bombers first
        And [
            [[COMBAT_TARGETS_VISIBLE_ENEMY]]
            Fighter
            Or [
               DesignHasPart name = "FT_HANGAR_3"
               DesignHasPart name = "FT_HANGAR_4"
            ]
        ]
        And [ [[COMBAT_TARGETS_VISIBLE_ENEMY]]  Fighter ]
        // if no fighters: target enemy ships
        And [ [[COMBAT_TARGETS_VISIBLE_ENEMY]]  [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]] ]
    ]
```
I was entertaining the thought that future-me will implement some WithFilteredCandidates condition like the following:
```
    combatTargets = WithFilteredCandidates 
        scope = [[COMBAT_TARGETS_VISIBLE_ENEMY]]
        condition = OrderedAlternativesOf [
            // Target Bombers and Heavy Bombers first
            And [
                Fighter
                Or [
                   DesignHasPart name = "FT_HANGAR_3"
                   DesignHasPart name = "FT_HANGAR_4"
                ]
            ]
            Fighter
            // if no fighters: target enemy ships
            [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]]
        ]
```
So for the purposes of this PR, it is fixed and tested.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+            // recreate previous state by moving back temp_non matches to non_matches
+            FCMoveContent(temp_non_matches, non_matches);
+
+            if (!found_sth) {
+                 // Also have to look at the other local candidates if there was a match
+                ObjectSet temp_matches;
+                temp_matches.reserve(matches.size());
+                operand->Eval(local_context, temp_matches, matches, NON_MATCHES);
+                if (!temp_matches.empty()) {
+                    // i.e. the current operand should be used negated to move
```
geoffthemedio: I don't understand this comment

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+                found_sth = true;
+            }
+            // recreate previous state by moving back temp_non matches to non_matches
+            FCMoveContent(temp_non_matches, non_matches);
+
+            if (!found_sth) {
+                 // Also have to look at the other local candidates if there was a match
+                ObjectSet temp_matches;
+                temp_matches.reserve(matches.size());
+                operand->Eval(local_context, temp_matches, matches, NON_MATCHES);
```
geoffthemedio: Is there a reason for testing input `matches` as `NON_MATCHES` but input `non_matches` as `MATCHES`? Might be more efficient to merge them all into one set, as if subconditions are all invariant, they would then only need to be evaluated once.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+        temp_non_matches.reserve(non_matches.size());
+        // try matching operand conditions until one matches
+        bool found_sth = false;
+        for (auto& operand : m_operands) {
+            // move those non matching the current operand to temp_non_matches
+            //            operand->Eval(local_context, matches, temp_non_matches, MATCHES);
+            // move those matching the current non-negated operand from non_matches to temp_non_matches
+            operand->Eval(local_context, non_matches, temp_non_matches, MATCHES);
+            if (!non_matches.empty()) {
+                // i.e. the current operand should be used negated to move
```
geoffthemedio: I don't understand this comment

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+        // Return non-matches if there are any
+        ObjectSet temp_non_matches;
+        temp_non_matches.reserve(non_matches.size());
+        // try matching operand conditions until one matches
+        bool found_sth = false;
+        for (auto& operand : m_operands) {
+            // move those non matching the current operand to temp_non_matches
+            //            operand->Eval(local_context, matches, temp_non_matches, MATCHES);
+            // move those matching the current non-negated operand from non_matches to temp_non_matches
+            operand->Eval(local_context, non_matches, temp_non_matches, MATCHES);
```
geoffthemedio: could probably search with `NON_MATCHES` here, with `non_matches` as the 3rd param and a temp container as the 2nd. Then if anything is in the temp container, can just move them into `matches` and don't need to touch `non_matches` again.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+    if (search_domain == NON_MATCHES) {
+        // check all operand conditions on all objects in the non matches set
+        // stop if the operand condition has at least one match
+        // Return non-matches if there are any
+        ObjectSet temp_non_matches;
+        temp_non_matches.reserve(non_matches.size());
+        // try matching operand conditions until one matches
+        bool found_sth = false;
+        for (auto& operand : m_operands) {
+            // move those non matching the current operand to temp_non_matches
```
geoffthemedio: // apply current operand condition to contents of non_matches, moving any that match into temp_non_matches.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+    //
+    // Not OrderedAlternativesOf [ A B C ] finds the topmost condition which has matches and then matches its non-matches.
+    // If any candidate matches A, then all candidates that do not match A are matched,
+    // or if no candidates match A but any candidate matches B, then all candidates that do not match B are matched,
+    // or if no candidate matches A or B but any candidate matches C, then all candidates that do not match C are matched.
+    // If no candidate matches A, B, or C, then all candidates are matched.
+    if (search_domain == NON_MATCHES) {
+        // check all operand conditions on all objects in the non matches set
+        // stop if the operand condition has at least one match
+        // Return non-matches if there are any
```
geoffthemedio: // check each operand condition on objects in the input matches and non_matches sets, until an operand condition matches an object. if an operand condition is selected, apply it to the input non_matches set, moving matching candidates to matches. if no operand condition is selected, because no candidate is matched by any operand condition, then do nothing.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+            if (!non_matches.empty()) {
+                // i.e. the current operand should be used negated to move
+                found_sth = true;
+            }
+            // recreate previous state by moving back temp_non matches to non_matches
+            FCMoveContent(temp_non_matches, non_matches);
+
+            if (!found_sth) {
+                 // Also have to look at the other local candidates if there was a match
+                ObjectSet temp_matches;
```
geoffthemedio: Could / should probably create a single `temp` set outside the loop / if, and reserve space in it equal to the larger of the sizes of `matches` and `non_matches`, rather than redoing the `reserve` call on each loop iteration here and separately in the `NON_MATCHES` case (there not in the loop).

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+                if (!temp_matches.empty()) {
+                    // i.e. the current operand should be used negated to move
+                    found_sth = true;
+                }
+                // recreate previous state
+                FCMoveContent(temp_matches, matches);
+            }
+
+            if (found_sth) {
+                // descent into subcondition for NON_MATCHES
```
geoffthemedio: // have selected the current operand condition to use, so apply it to the input candidate sets.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+                break;
+            }
+            // try the next operand
+        }
+        // If NON_MATCHES could not find any matches it should match all
+        if (!found_sth)
+            FCMoveContent(matches, non_matches);
+    } else /*(search_domain == MATCHES)*/ {
+        // check all operand conditions on all objects in the matches set, collecting non matches temporarily
+        // stop if the operand condition matches at least one item
```
geoffthemedio: // check each operand condition on objects in the input matches and non_matches sets, until an operand condition matches an object. if an operand condition is selected, apply it to the input matches set, moving non-matching candidates to non_matches. if no operand condition is selected, because no candidate is matched by any operand condition, then move all of the input matches into non_matches.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+        // if the operand condition does not match, restore the matches set by moving the temporary non-matches back
```
geoffthemedio: this comment would be better inside the loop where it happens

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+        ObjectSet temp_non_matches;
+        temp_non_matches.reserve(matches.size());
+        // try matching operand conditions until one mateches
```
geoffthemedio: comment seems redundant

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+        // use a fresh set of possible matches
```
geoffthemedio: this comment seems odd here... would be more applicable inside the loop

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+        for (auto& operand : m_operands) {
+            // move back temp_non_matches
```
geoffthemedio: // restore input matches set by moving back any that were moved into temp_non_matches on previous loop iteration

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+            FCMoveContent(temp_non_matches, matches);
+            // move non matches from_matches to temp_non_matches
```
geoffthemedio: // test current operand condition on input matches set

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+            operand->Eval(local_context, matches, temp_non_matches, MATCHES);
+            if (!matches.empty()) break;
```
geoffthemedio: // if anything in the input matches set was matched by the current operand condition, then the current operand condition is the one that will be applied.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+            // Check if other LocalCandidates (in the non_matches set) match
```
geoffthemedio: // test current operand condition on input non_matches set.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+            operand->Eval(local_context, matches, non_matches, NON_MATCHES);
```
geoffthemedio: it's a bit confusing to reuse the matches set here, rather than a temp set

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+            if (!matches.empty()) {
+                // This operand should return the result, but no matches exist
```
geoffthemedio: One of the input non_matches set objects matched the current operand condition, so this condition will be used. None of the input matches objects matched this condition, though, so everything in the input matches and non_matches sets should be moved into non_matches.

// Input matches are currently in temp_non_matches, and some of the input non_matches are currently in matches. Move the contents to matches back into non_matches here. The contents of temp_non_matches are moved into non_matches below.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+                // So all objects should go into the non_matches set
+                // move all matches to non_matches here
+                FCMoveContent(matches, non_matches);
+                // stop finding the operand
+                // temp_non_matches must still be moved to non_matches
+                break;
```
geoffthemedio: // have selected a operand condition to use, so can stop looping over operand conditions.

universe/Condition.cpp
```diff
@@ -9956,6 +9965,249 @@ unsigned int Not::GetCheckSum() const {
...
+            }
+        }
+
+        // non matching items were already moved from matches to temp_non_matches
+        // move non matching items from temp_non_matches to non_matches
```
geoffthemedio: // Objects in the input matches set that did not match the chosen operand condition were moved above into temp_non_matches. Move them into non_matches (since they aren't matched by this condition).

geoffthemedio:
@agrrr3 Note https://github.com/freeorion/freeorion/blob/master/combat/CombatSystem.cpp#L1406

There is no an optional species targeting condition applied before the weapon condition. I debated whether to put the species condition before or after the weapon condition. After, and the species could pick between possible targets of the weapon. Before, the weapon can do prioritized targetting among the targets the species is willing to attack. Since you've planning to do prioritized targetting of weapons with this pull request, I figured that before made more sense.

universe/Condition.cpp
agrrr3: Hm. I think my code is more complex than the problem. I would like to refactor to something more simple. I am halfway through thinking, please correct my reasoning:
the search domains are there for execution optimization (so you usually do not have to e.g. check an object for a match if you already know it fails to match a sibling in the surrounding AND condition. And you do not need to check an object for a non-match if you know it matches a sibling in the surrounding OR condition).  
 For the OrderedAlternativesOf in worst case I have to check all in the matches and non_matches set for a match, to find out if a operand condition is the one to determine matches (as determining the correct operand condition is not about optimization but determines the actual result).

I think I can optimize a just a little bit: 

If i am in the MATCHING domain, i can search the matches first and if there is a match I can add the temp_non_matches to the non_matches and i am finished.

If i am in the NON_MATCHING domain, i can search the non_matches first and if there is a match I can add the temp_matches to the matches and i am finished.

geoffthemedio: I thought that was what it does already, except checking both `matches` and `non_matches` when deciding which subcondition to use...

universe/Condition.cpp
agrrr3: Do you propose a change in semantics here? Probably my current implementation is actually wrong.

Because at the moment it is: if no operand condition is selected, because no candidate is matched by any operand condition, then move all objects from the matches input set into the non_matches set.

I will test using an OR condition.

geoffthemedio: Not sure if it is a change. I was suggesting more details for the comment about how I thought / expected it did or should work.

agrrr3: yes, confirmed there was a bug, fixed

universe/Condition.cpp
agrrr3: i know but if i introduce another local variable i also need an extra move, and besides 'matches' is actually a good name

universe/Condition.cpp
agrrr3: I guess I do not know how the invariance really works. Keeping the sets separated means i can often skip checking many objects.

universe/Condition.cpp
agrrr3: I was using the Not condition to reason about the NON_MATCHING domain. So that comment is not really useful. I cleaned up

agrrr3:
> @agrrr3 Note https://github.com/freeorion/freeorion/blob/master/combat/CombatSystem.cpp#L1406
> 
> There is no an optional species targeting condition applied before the weapon condition. I debated whether to put the species condition before or after the weapon condition. After, and the species could pick between possible targets of the weapon. Before, the weapon can do prioritized targetting among the targets the species is willing to attack. Since you've planning to do prioritized targetting of weapons with this pull request, I figured that before made more sense.

I think for species we will have rather preferences than restrictions. The weapon condition will really say "Flak can shoot only fighters". But we probably wont say "Trith can only shoot headache-making species (all but Trith or exobot)".

Preferences need to be applied after the restrictions. The other way round there may be no targets left, because they were not preferred.

I am certainly not tackling generic composition of conditions in this PR. I thought you wanted to make other sources of targeting conditions available as a kind of reference one would manually in FOCS mix into the ship part combatTargets condition.

agrrr3:
Note I processed all the review comments. Some of my answers do not show up here.

I think the implementation is now sufficiently clean and documented.

universe/Condition.cpp
geoffthemedio: I wrote an explanation of invariance, but I think it doesn't actually matter. The question would be whether an expensive subcondition needs to be evaluated multiple times. For example, if the OrderedAlternativesOf contained a Number subcondition, which itself contained a complicated subsubcondition for matching objects in the universe, that subsubcondition would need to be evaluated every time Number is evaluated on a set of objects, so if you evaluated it separately for input matches and non-matches, that could take longer. If all the subconditions are relatively simple, though, and there aren't many subsubconditions that are evaluated on objects other than the actual candidates being matched, then it shouldn't matter.

geoffthemedio:
> I think for species we will have rather preferences than restrictions. The weapon condition will really say "Flak can shoot only fighters". But we probably wont say "Trith can only shoot headache-making species (all but Trith or exobot)".

I thought the opposite. You already added some prioritized preferences for weapons, so also having them for species would be a bit complicated to combine. And I would expect some species would refuse to target particular other species, such as themselves, or perhaps just other biological / robotic / lithic species on moral grounds. Whether these restrictions are enforced might depend on species happiness levels or imperial policies as well.

agrrr3:
@geoffthemedio So it looks like reasonable order and more general the way how conditions combine  depends on the actual content (in order to support both restrictions and preferences). As you did not like the statistical combination I offered before, I guess you do not like auto resolution here. While I see more sophisticated ways to specify and combine preferences and restrictions I think a general solution would put this combination in the hands of the FOCS scripter. I think for ships and planets actually one can get reasonable information about species traits (via tags) for scripting species preferences in a generic way in the ship parts. This is currently not possible for fighters. 

Does the PR have open issues from your point of view?

universe/Condition.cpp
```diff
@@ -9956,6 +9965,238 @@ unsigned int Not::GetCheckSum() const {
...
+    // If any candidate matches A, then all candidates that do not match A are matched,
+    // or if no candidates match A but any candidate matches B, then all candidates that do not match B are matched,
+    // or if no candidate matches A or B but any candidate matches C, then all candidates that do not match C are matched.
+    // If no candidate matches A, B, or C, then all candidates are matched.
+    if (search_domain == NON_MATCHES) {
+        // Check each operand condition on objects in the input matches and non_matches sets, until an operand condition matches an object.
+        // If an operand condition is selected, apply it to the input non_matches set, moving matching candidates to matches.
+        // If no operand condition is selected, because no candidate is matched by any operand condition, then do nothing.
+        ObjectSet temp_objects;
+        temp_objects.reserve(std::max(matches.size(),non_matches.size()));
```
geoffthemedio, line 10035: Should probably reserve space in `matches` as well, as is done below when searching domain `MATCHES`, in case the content of `non_matches` is Tx into `matches`

universe/Condition.cpp
```diff
@@ -9956,6 +9965,238 @@ unsigned int Not::GetCheckSum() const {
...
+        }
+
+        // No operand condition was selected, nothing changed, done.
+    } else /*(search_domain == MATCHES)*/ {
+        // Check each operand condition on objects in the input matches and non_matches sets, until an operand condition matches an object.
+        // If an operand condition is selected, apply it to the input matches set, moving non-matching candidates to non_matches.
+        // If no operand condition is selected, because no candidate is matched by any operand condition, then move all of the input matches into non_matches.
+        ObjectSet temp_non_matches;
+        temp_non_matches.reserve(non_matches.size());
+        matches.reserve(std::max(matches.size(),non_matches.size()));
```
geoffthemedio: Should this actually be reserving in `non_matches` since that is the container objects can be moved into?

Ah, you re-use `matches` as the temp set below... I still find that confusing, but I guess it should be both being reserved then?

universe/Condition.cpp
```diff
@@ -9956,6 +9965,238 @@ unsigned int Not::GetCheckSum() const {
...
+
+        for (auto& operand : m_operands) {
+            // Apply the current operand optimistically. Select it if there are any matching objects in the input sets 
+            operand->Eval(local_context, matches, temp_non_matches, MATCHES);
```
geoffthemedio: If you evaluated this with the `NON_MATCHES` domain and `matches` as the second / source input set and `temp_objects` as the first / destination, then matches would be left containing the stuff that doesn't match the operand condition... You would still need to swap the contents if nothing in `non_matches` was matched by that operand condition, but it could save a temp set, I think?

agrrr3: I think you are right. Looks i even got the implementation right first try and it looks more structured. There is one extra move though, but it is the happy path anyway (so just a single time happening) and it is based on moving, so i think the cleaner implementation is worth it.

universe/Condition.cpp
agrrr3: used your suggestion. matches temp set use be gone!

PR #2377, opened 2019-02-06
Some fixes for MSVC projects generated by cmake

o01eg:
Tried to build FreeOrion on win64 with x64 version of sdk. Command used is `cmake -DBUILD_TESTING=On -DCMAKE_BUILD_TYPE=RelWithDebInfo -G "Visual Studio 15 2017 Win64" -T v141,host=x64 -DPYTHON_EXECUTABLE=..\python.exe ..`
It still doesn't work but fixed some errors.

PR #2386, opened 2019-02-17
Fix Boost auto-linking

o01eg:
Follow up  #2377. Correctly disables boost auto-linking on windows.


geoffthemedio:
assuming you know what you're doing...

PR #2393, opened 2019-02-27
Enter/leave chat messages

o01eg:
Adds support to show in chat "Player entered game" and "Player left game" messages.

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -287,7 +287,7 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
         }
         // add "player left game" message
         boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
-        std::string data = player_connection->PlayerName() + " [[" + UserStringNop("PLAYER_LEFT_GAME") + "]]";
+        std::string data = (boost::format("[[PLAYER_LEFT_GAME,%s]]") % player_connection->PlayerName()).str();
         m_server.PushChatMessage(data, "", GG::CLR_WHITE, timestamp);
 
         // send message to other players
@@ -504,7 +504,7 @@ bool ServerFSM::EstablishPlayer(const PlayerConnectionPtr& player_connection,
         {
             // add "player enter game" message
             boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
-            std::string data = player_name + " [[" + UserStringNop("PLAYER_ENTERED_GAME") + "]]";
+            std::string data = (boost::format("[[PLAYER_ENTERED_GAME,%s]]") % player_connection->PlayerName()).str();
             m_server.PushChatMessage(data, "", GG::CLR_WHITE, timestamp);
 
             // send message to other players
@@ -994,7 +994,7 @@ sc::result MPLobby::react(const Disconnection& d) {
 
         // add "player left game" message
         boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
-        std::string data = player_connection->PlayerName() + " [[" + UserStringNop("PLAYER_LEFT_GAME") + "]]";
+        std::string data = (boost::format("[[PLAYER_LEFT_GAME,%s]]") % player_connection->PlayerName()).str();
         server.PushChatMessage(data, "", GG::CLR_WHITE, timestamp);
 
         // send message to other players
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -490,6 +490,19 @@ bool ServerFSM::EstablishPlayer(const PlayerConnectionPtr& player_connection,
             client_type == Networking::CLIENT_TYPE_HUMAN_OBSERVER ||
             client_type == Networking::CLIENT_TYPE_HUMAN_PLAYER)
         {
+            // add "player enter game" message
+            boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
+            std::string data = player_name + " [[" + UserStringNop("PLAYER_ENTERED_GAME") + "]]";
+            m_server.PushChatMessage(data, "", GG::CLR_WHITE, timestamp);
+
+            // send message to other players
+            for (auto it = m_server.m_networking.established_begin();
+                 it != m_server.m_networking.established_end(); ++it)
+            {
+                if (player_connection != (*it))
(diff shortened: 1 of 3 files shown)


geoffthemedio:
This should be implemented as a network message that the client responds to by generating the message text in chat locally, not as a server-side composed text message that will be the same regardless of players' local language settings.

o01eg:
> This should be implemented as a network message that the client responds to by generating the message text in chat locally,

I want those messages could be stored.
> not as a server-side composed text message that will be the same regardless of players' local language settings

It translates on the client depending on client's string table file.

server/ServerFSM.cpp
```diff
@@ -456,6 +468,19 @@ bool ServerFSM::EstablishPlayer(const PlayerConnectionPtr& player_connection,
             client_type == Networking::CLIENT_TYPE_HUMAN_OBSERVER ||
             client_type == Networking::CLIENT_TYPE_HUMAN_PLAYER)
         {
+            // add "player enter game" message
+            boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
+            std::string data = player_name + " [[" + UserStringNop("PLAYER_ENTERED_GAME") + "]]";
```
geoffthemedio: It translates `PLAYER_ENTERED_GAME` when the chatbox sees the `[[` and `]]`, but this is doing the string composition in code with `+` instead of substitution with `FlexibleFormat`, which would allow the stringtable / translation to control where and how the name appears in the string.

o01eg: Changed to boost::format

server/ServerFSM.cpp
```diff
@@ -274,7 +274,7 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
         }
         // add "player left game" message
         boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
-        std::string data = player_connection->PlayerName() + " [[" + UserStringNop("PLAYER_LEFT_GAME") + "]]";
+        std::string data = (boost::format("%s [[PLAYER_LEFT_GAME]]") % player_connection->PlayerName()).str();
```
geoffthemedio: This is no different from using `+`.

The point is that the string composition should be done client-side so the stringtable of the client can determine how and where to put the parameter into the string.

server/ServerFSM.cpp
geoffthemedio: You missed the part about using the stringtable entry to determine where and how the name appears in the string. Just using `boost::format` instead of `+` but keeping the position of the name in the string the same is worse than it was before.

server/ServerFSM.cpp
geoffthemedio: Not all languages have Subject-Verb-Object ordering, so you shouldn't assume the player name will appear first in the string. If the substitution is done on the client, and has a playername parameter, the stringtable can specify where the substituted name should appear in the displayed text.

o01eg: Does `[[STRINGTABLE_ENTRY]]` have arguments support so it could be written like `[[PLAYER_LEFT_GAME,Player name]]`?

geoffthemedio: No, see here:

https://github.com/freeorion/freeorion/blob/master/UI/ChatWnd.cpp#L37

It's a bit hard to follow, but it scans through the input chat message, copying everything to the filtered message until it sees `[[`, then it searches for a stringtable entry that matches what comes next before the following `]]`. It then copies the stringtable entry's value instead of the `[[` and `]]` and the stuff between them.

No format call occurs, so no formatting gets done like for `VarText`, and no recursive macro substitution is done, like when loading the stringtable or scripts and processing macros.

o01eg: Ok, I'll postpone this PR until some stringtable arguments would be implemented.

geoffthemedio: https://github.com/freeorion/freeorion/commit/783fa637b5733a2730aa69e18ff441daa4aaad1c should support client-side string or stringtable entry substitutions.

o01eg: Thanks, updated PR according to this change.

o01eg:
Rebased. Used new argument support for enter/leave messages.

server/ServerFSM.cpp
```diff
@@ -285,6 +285,18 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
                 m_server.DropPlayerEmpireLink(id);
             }
         }
+        // add "player left game" message
+        boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
+        std::string data = (boost::format("[[PLAYER_LEFT_GAME,%s]]") % player_connection->PlayerName()).str();
```
geoffthemedio: `boost::format` isn't necessary here. It'd be fine to do just:
`std::string data = "[[PLAYER_LEFT_GAME," + player_connection->PlayerName()).str() + "]]";`

That said, it's probably slightly easier to read the current version, so OK.

server/ServerFSM.cpp
```diff
@@ -285,6 +285,18 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
                 m_server.DropPlayerEmpireLink(id);
             }
         }
+        // add "player left game" message
+        boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
+        std::string data = "[[PLAYER_LEFT_GAME," + player_connection->PlayerName() + "]]";
```
geoffthemedio: arguably the `PLAYER_LEFT_GAME` should be wrapped in `UserStringNoOp` ...

server/ServerFSM.cpp
```diff
@@ -504,7 +504,7 @@ bool ServerFSM::EstablishPlayer(const PlayerConnectionPtr& player_connection,
         {
             // add "player enter game" message
             boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
-            std::string data = "[[PLAYER_ENTERED_GAME," + player_connection->PlayerName() + "]]";
+            std::string data = std::string("[[") + UserStringNop("PLAYER_ENTERED_GAME") + "," + player_connection->PlayerName() + "]]";
```
geoffthemedio, line 507: Is the `std::string` wrapping `"[["` neccessary?

o01eg: Yes, without it I get `const char* + const char*` error.

PR #2396, opened 2019-03-02
Let Interspecies academy research boost change target research

agrrr3:
This is a bugfix. Research boost form interspecies academy should affect target research (+5 if research focus), not the change rate of research.


geoffthemedio:
I've never really understood the idea / point of this building, but it was implemented by you, so presumably you have some idea how it was supposed to work...

PR #2400, opened 2019-03-07
Allow to load game without all human player connected

o01eg:
Next step of #2243.
Make new restrictions work in lobby.
All empires considered as human if there no AI player.
It works even if Observer player launch saved game without any player.

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -278,8 +278,8 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
 
     // count of active (non-eliminated) empires, which currently have a connected human players
     int empire_connected_plr_cnt = 0;
-    // count of active (non-eliminated) empires, which currently have a disconnected human players
-    int empire_disconnected_plr_cnt = 0;
+    // count of active (non-eliminated) empires, which currently have a unconnected human players
+    int empire_unconnected_plr_cnt = 0;
     for (const auto& empire : Empires()) {
         if (!empire.second->Eliminated()) {
             switch (m_server.GetEmpireClientType(empire.first)) {
@@ -287,7 +287,7 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
                 empire_connected_plr_cnt++;
                 break;
             case Networking::INVALID_CLIENT_TYPE:
-                empire_disconnected_plr_cnt++;
+                empire_unconnected_plr_cnt++;
                 break;
             case Networking::CLIENT_TYPE_AI_PLAYER:
                 // ignore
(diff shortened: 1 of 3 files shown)


server/ServerFSM.cpp
```diff
@@ -748,6 +748,14 @@ void MPLobby::ValidateClientLimits() {
             ai_count++;
     }
 
+    const int human_connected_count = human_count;
+
+    // for load game consider as human all non-AI empires
+    // because human player could connect later in game
+    if (!m_lobby_data->m_new_game) {
+        human_count = m_lobby_data->m_save_game_empire_data.size() - ai_count;
```
geoffthemedio: does this matter? if there are fewer players than empires, won't it already prevent the game from starting?

server/ServerFSM.cpp
```diff
@@ -762,11 +770,22 @@ void MPLobby::ValidateClientLimits() {
...
         m_lobby_data->m_start_lock_cause = UserStringNop("ERROR_NOT_ENOUGH_HUMAN_PLAYERS");
+    } else if (human_connected_count < min_connected) {
+        m_lobby_data->m_start_locked = true;
+        m_lobby_data->m_start_lock_cause = UserStringNop("ERROR_NOT_ENOUGH_CONNECTED_HUMAN_PLAYERS");
+    } else if (max_disconnected > 0 &&
+        !m_lobby_data->m_new_game &&
+        static_cast<int>(m_lobby_data->m_save_game_empire_data.size()) - ai_count - human_connected_count >= max_disconnected)
+    {
+        m_lobby_data->m_start_locked = true;
+        m_lobby_data->m_start_lock_cause = UserStringNop("ERROR_TOO_MANY_DISCONNECTED_HUMAN_PLAYERS");
```
geoffthemedio: What does this mean / do in the lobby? I sort of understand stopping the server if humans disconnect from a game in progress, but why would disconnections from the lobby matter? Or when is this number reset...?

server/ServerFSM.cpp
```diff
@@ -748,6 +748,14 @@ void MPLobby::ValidateClientLimits() {
             ai_count++;
     }
 
+    const int human_connected_count = human_count;
+
+    // for load game consider as human all non-AI empires
+    // because human player could connect later in game
```
geoffthemedio, line 754: what is the consequence of counting human players this way...?

o01eg: This allows to check restrictions on human player count. As an example if it capped by 5 but 4 human players loads game with 10 human empires the game will exceed limit without this check when all players connect to the game.

server/ServerFSM.cpp
o01eg: It is to prevent violations of restrictions. Without this check it will be possible to load the game so there will be more disconnected players that allowed in options at start.

geoffthemedio: I don't understand at all. What is a "disconnected player" when in the lobby? What is a "disconnected player" at all? A player that was connected and then disconnects, or a player slot of some sort that doesn't currently have a player in it? If the latter, then something needs to be renamed, as the meaning isn't clear.

server/ServerFSM.cpp
geoffthemedio: I thought the idea was that a multi-player save could be loaded with just a single player connected, who could play their turn then disconnect, and then another player could join and play their turn, until all the turns are played, but with only a single player ever connected. Why shouldn't this be possible, even if there is a limit (of more than one) for how many players can be connected at once?

english tip: don't write "allows to check something", but rather "allows something to be checked" or "allows something to check something else"

server/ServerFSM.cpp
o01eg: > What is a "disconnected player" when in the lobby? What is a "disconnected player" at all?

Each non-eliminated empire without connected player considered to be one. May be better to rename it "disconnected empire" or so?
> or a player slot of some sort that doesn't currently have a player in it

Yes, it is.

geoffthemedio: The word "disconnected" is the problem; to me it means "was previously connected but isn't now", but you're using it to mean "doesn't have a connected player".

A better word would be "unconnected".

Also, preferably add "human" and/or "players" to the variable and option names, so
````
    int min_connected_human_empire_players = GetOptionsDB().Get<int>("network.server.conn-human-empire-players.min");
    int max_unconnected_human_empire_players = GetOptionsDB().Get<int>("network.server.unconn-human-empire-players.max");
````

server/ServerFSM.cpp
o01eg: I suppose restriction works on playing game should be checked before playing game as well.
> I thought the idea was that a multi-player save could be loaded with just a single player connected, who could play their turn then disconnect, and then another player could join and play their turn, until all the turns are played, but with only a single player ever connected. 

Actually it won't work before last step of #2243 would be implemented. If server set up to stop on no players connected then next load resets ready status of empires.

> Why shouldn't this be possible, even if there is a limit (of more than one) for how many players can be connected at once?

Because the server's owner sets up this limit.

server/ServerFSM.cpp
o01eg: Renamed.

PR #2402, opened 2019-03-07
Automatically save game before exit

o01eg:
Save game in hostless mode if server stops playing the game.
Prerequisite for last step of #2243 so partial save will be available.

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -323,6 +323,25 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
     if (must_quit) {
         ErrorLogger(FSM) << "Unable to recover server terminating.";
         if (m_server.IsHostless()) {
+            if (GetOptionsDB().Get<bool>("save.auto.hostless.enabled")) {
+                // save game on exit
+                std::string save_filename = GetAutoSaveFileName();
+                ServerSaveGameData server_data(m_server.CurrentTurn());
+                int bytes_written = 0;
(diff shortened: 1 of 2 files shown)


server/ServerFSM.cpp
```diff
@@ -473,6 +492,18 @@ bool ServerFSM::EstablishPlayer(const PlayerConnectionPtr& player_connection,
...
 }
 
+std::string ServerFSM::GetAutoSaveFileName() {
+    std::string subdir = GetGalaxySetupData().GetGameUID();
+    boost::filesystem::path autosave_dir_path = GetServerSaveDir() / (subdir.empty() ? "auto" : subdir);
+    const auto& extension = MP_SAVE_FILE_EXTENSION;
+    // Add timestamp to autosave generated files
+    std::string datetime_str = FilenameTimestamp();
+
+    std::string save_filename = boost::io::str(boost::format("FreeOrion_%04d_%s%s") % m_server.CurrentTurn() % datetime_str % extension);
```
geoffthemedio: is accessing `m_server.CurrentTurn()` the only reason this function is a member of `ServerFSM`?

o01eg: No, this code to generate save file name is used in two places now so it is also refactored to reduce code repetition. `ServerFSM` is a common space for both usages.

geoffthemedio: my point was whether that was the only reason for it not to be a file-scope free function that doesn't appear in the header

PR #2403, opened 2019-03-08
Notify players that eliminated empire finished its orders

o01eg:


server/ServerApp.cpp
```diff
@@ -1744,6 +1744,16 @@ bool ServerApp::EliminatePlayer(const PlayerConnectionPtr& player_connection) {
     // break link between player and empire
     m_player_empire_ids.erase(player_id);
 
+    // notify other player that this empire finished orders
```
geoffthemedio, line 1747: Could you clarify why this is useful and how it works? Does the eliminated player just stay in waiting mode on the other players' clients, for the rest of the session, or until the eliminated player disconnects? Are they disconnected automatically at some point? Is there any remaining empire to player connection after the elimination happens?

o01eg: There only change is to show other player that eliminated empire made its choice.
> Could you clarify why this is useful and how it works?

Currently other player see waiting status of eliminated empire until turn advance. They could think player of eliminated empire made its turn too long and postpone their turns as well.
> Does the eliminated player just stay in waiting mode on the other players' clients, for the rest of the session, or until the eliminated player disconnects?

On turn advance other player will get elimination sitrep and this empire get elimination mark in the empires' list.
> Are they disconnected automatically at some point?

No, but they stopped to get turn updates and to act in the game.
> Is there any remaining empire to player connection after the elimination happens?

No, link between empire and player is erased after elimination.

geoffthemedio: could you then clarify in the comment why this is being done?

o01eg: Added more commentary.

PR #2404, opened 2019-03-09
Save and load readiness status

### LINKED ISSUE #2243: Allow player to disconnect from the server without stopping the game

o01eg:
Save readiness status. Revoke it on load for online players so server wouldn't advance turn because saves are made when all players sent orders and became ready.
Fixes #2243.
Updates save game format.

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -1345,9 +1345,9 @@ void ServerApp::LoadGameInit(const std::vector<PlayerSaveGameData>& player_save_
             ErrorLogger() << "LoadGameInit got inconsistent empire ids between player save game data and result of PlayerEmpireID";
         }
 
-        // Revoke readiness only for online players so them could re-make orders of current turn.
-        // Without it server would advance turn because saves are made when all players sent orders
-        // and became ready.
+        // Revoke readiness only for online players so they can redo orders for the current turn.
+        // Without doing it, server would immediatly advance the turn because saves are made when
+        // all players sent orders and became ready.
         RevokeEmpireTurnReadyness(empire_id);
 
         // restore saved orders.  these will be re-executed on client and
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -693,7 +693,8 @@ void ServerApp::NewGameInitConcurrentWithJoiners(
         if (GetEmpire(empire_id))
             AddEmpireTurn(empire_id, PlayerSaveGameData(player_id_and_setup.second.m_player_name, empire_id,
                                                         nullptr, nullptr, std::string(),
-                                                        player_id_and_setup.second.m_client_type));
+                                                        player_id_and_setup.second.m_client_type,
+                                                        false));
     }
 
     // update visibility information to ensure data sent out is up-to-date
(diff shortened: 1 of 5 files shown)


util/MultiplayerCommon.h
```diff
@@ -169,30 +169,36 @@ struct FO_COMMON_API PlayerSaveGameData : public PlayerSaveHeaderData {
...
         m_ui_data(ui_data),
-        m_save_state_string(save_state_string)
+        m_save_state_string(save_state_string),
+        m_ready(ready)
     {}
 
     std::shared_ptr<OrderSet>       m_orders;
     std::shared_ptr<SaveGameUIData> m_ui_data;
     std::string                     m_save_state_string;
+    bool                            m_ready;
```
geoffthemedio, line 192: specify default `= false;` value here preferably... (just noticed it is set in the initializer list though)

server/ServerApp.cpp
```diff
@@ -1344,6 +1345,11 @@ void ServerApp::LoadGameInit(const std::vector<PlayerSaveGameData>& player_save_
             ErrorLogger() << "LoadGameInit got inconsistent empire ids between player save game data and result of PlayerEmpireID";
         }
 
+        // Revoke readiness only for online players so them could re-make orders of current turn.
```
geoffthemedio: so they can redo order for the current turn.

util/MultiplayerCommon.h
o01eg: Should it be changed? I don't see in other classes other than initializer list initialization.

server/ServerApp.cpp

util/MultiplayerCommon.h
geoffthemedio: They are both used in various places. I'm currently preferring to move towards default-at-declare rather than in-initializer-list, but it's not important for a single line if the surrounding code uses the other method.

server/ServerApp.cpp
```diff
@@ -1344,6 +1345,11 @@ void ServerApp::LoadGameInit(const std::vector<PlayerSaveGameData>& player_save_
             ErrorLogger() << "LoadGameInit got inconsistent empire ids between player save game data and result of PlayerEmpireID";
         }
 
+        // Revoke readiness only for online players so them can redo orders for the current turn.
```
geoffthemedio: them -> they

server/ServerApp.cpp
```diff
@@ -1344,6 +1345,11 @@ void ServerApp::LoadGameInit(const std::vector<PlayerSaveGameData>& player_save_
...
+        // Without it server would advance turn because saves are made when all players sent orders
```
geoffthemedio: Without doing this, the server would immediately advance the turn...

server/ServerApp.cpp

PR #2407, opened 2019-03-10
Timed auto save

o01eg:
Follows up #2405.
Add periodic submitted orders backup support via savegame in case when turn lasts too long.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1891,6 +1891,9 @@ Enable autosave when resigning or closing the game.
 OPTIONS_DB_AUTOSAVE_HOSTLESS
 Enable autosave in hostless mode.
 
+OPTIONS_DB_AUTOSAVE_HOSTLESS_TIMEOUT
+Duration in seconds when game will be saved in hostless mode after turn started. Prevents losing player orders on server crash. 0 if disabled.
+
 OPTIONS_DB_UI_MOUSE_LR_SWAP
 Swaps results of clicking left and right mouse buttons.
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1891,7 +1891,7 @@ Enable autosave when resigning or closing the game.
 OPTIONS_DB_AUTOSAVE_HOSTLESS
 Enable autosave in hostless mode.
 
-OPTIONS_DB_AUTOSAVE_TIMEOUT
+OPTIONS_DB_AUTOSAVE_INTERVAL
 Delay in seconds after the most recent turn start or autosave until the next autosave. Prevents losing player orders on server crash. 0 if disabled.
 
 OPTIONS_DB_UI_MOUSE_LR_SWAP
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1891,8 +1891,8 @@ Enable autosave when resigning or closing the game.
 OPTIONS_DB_AUTOSAVE_HOSTLESS
 Enable autosave in hostless mode.
 
-OPTIONS_DB_AUTOSAVE_HOSTLESS_TIMEOUT
-Duration in seconds when game will be saved in hostless mode after turn started. Prevents losing player orders on server crash. 0 if disabled.
+OPTIONS_DB_AUTOSAVE_TIMEOUT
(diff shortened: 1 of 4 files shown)


default/stringtables/en.txt
```diff
@@ -1891,6 +1891,9 @@ Enable autosave when resigning or closing the game.
 OPTIONS_DB_AUTOSAVE_HOSTLESS
 Enable autosave in hostless mode.
 
+OPTIONS_DB_AUTOSAVE_HOSTLESS_TIMEOUT
+Seconds when autosave happened in hostless mode. 0 if disabled.
```
geoffthemedio: meaning unclear

geoffthemedio: why are the timed autosaves only in hostless mode?

geoffthemedio: would also be good to have an automatic turn timer, though that's a separate issue

o01eg: Tried to rephrase it.

o01eg: The host can manually save game. I could remove this restriction although.

o01eg: Now timed autosave works in any mode.

o01eg: Turn timer requires some time sync with client so player would know correct time before turn change.

default/stringtables/en.txt
```diff
@@ -1891,6 +1891,9 @@ Enable autosave when resigning or closing the game.
 OPTIONS_DB_AUTOSAVE_HOSTLESS
 Enable autosave in hostless mode.
 
+OPTIONS_DB_AUTOSAVE_TIMEOUT
+Duration in seconds when game will be saved after turn started. Prevents losing player orders on server crash. 0 if disabled.
```
geoffthemedio: Delay from turn start until the single autosave occurs, or the delay after the most recent turn start or autosave until the next autosave?

util/MultiplayerCommon.cpp
```diff
@@ -34,6 +34,7 @@ namespace {
         db.Add("save.format.binary.enabled",                UserStringNop("OPTIONS_DB_BINARY_SERIALIZATION"),   false);
         db.Add("save.format.xml.zlib.enabled",              UserStringNop("OPTIONS_DB_XML_ZLIB_SERIALIZATION"), true);
         db.Add("save.auto.hostless.enabled",                UserStringNop("OPTIONS_DB_AUTOSAVE_HOSTLESS"),      true);
+        db.Add<int>("save.auto.timeout",           UserStringNop("OPTIONS_DB_AUTOSAVE_TIMEOUT"), 0, Validator<int>());
```
geoffthemedio: indent to match

geoffthemedio: "timeout" is misleading... maybe "interval"?

o01eg: Renamed.

default/stringtables/en.txt

server/ServerFSM.cpp
```diff
@@ -2931,6 +2941,20 @@ sc::result WaitingForTurnEnd::react(const SaveGameRequest& msg) {
...
+void WaitingForTurnEnd::SaveTimedoutHandler(const boost::system::error_code& error) {
+    if (error) {
+        DebugLogger() << "Save timed out cancelled";
+        return;
+    }
+
+    DebugLogger() << "Save timed out.";
+    PlayerConnectionPtr dummy_connection = nullptr;
+    Server().m_fsm->process_event(SaveGameRequest(HostSaveGameInitiateMessage(GetAutoSaveFileName(Server().CurrentTurn())), dummy_connection));
+    m_timeout.expires_from_now(std::chrono::seconds(GetOptionsDB().Get<int>("save.auto.interval")));
```
geoffthemedio: what happens when you pass `0` to `expires_from_now`? I can't tell from the documentation. Is there a reason to make this call if the option is set to 0?

o01eg: Handler called when "save.auto.interval" was more than 0. Is it possible option change its value after the server started?

geoffthemedio: I think client option updates can propagate to servers during a game... Would certainly be preferable in a case like this that that happens.

o01eg: Added check.

PR #2408, opened 2019-03-11
Add option to start hostless server from saved game

o01eg:
Add option `load` to start hostless server from saved game. Starts game without any human players connected so requires to have zero minimum of connected human player.
Supposed to be part of script which choose latest save file and re-launches freeoriond if it stopped.

Changes order of FSM initialization and background parsing. Without it `LoadGame` fails on `SpeciesManager`.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1262,6 +1262,9 @@ Start server in single player host mode. This only allows clients on localhost t
 OPTIONS_DB_HOSTLESS
 Start server in hostless mode. Server accepts players in the multiplayer lobby, and returns to the lobby after the game session ends.
 
+OPTIONS_DB_HOSTLESS_AUTOSTART_LOAD
+Start server in hostless mode and load and start game from provided save file. Server accepts players in the playing game, and returns to the lobby after the game session ends.
+
 OPTIONS_DB_SKIP_CHECKSUM
 Skip comparing the resource directory checksums.  This allows faster startup when client and server are on the same machine, using the same resource directory.
 
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1262,9 +1262,6 @@ Start server in single player host mode. This only allows clients on localhost t
 OPTIONS_DB_HOSTLESS
 Start server in hostless mode. Server accepts players in the multiplayer lobby, and returns to the lobby after the game session ends.
 
-OPTIONS_DB_HOSTLESS_AUTOSTART_LOAD
-Start server in hostless mode and load and start game from provided save file. Server accepts players in the playing game, and returns to the lobby after the game session ends.
-
 OPTIONS_DB_SKIP_CHECKSUM
 Skip comparing the resource directory checksums.  This allows faster startup when client and server are on the same machine, using the same resource directory.
 
@@ -1931,7 +1928,7 @@ OPTIONS_DB_UI_SITREP_ICONSIZE
 Sets the sitrep icon width and height; default 16 (min 12, max 64).
 
 OPTIONS_DB_LOAD
-Loads the specified single-player save game.
+Loads and starts the specified single-player or multi-player save game.
 
 OPTIONS_DB_EFFECTS_THREADS_UI_DESC
 Specifies number of threads to use for effects processing in the user interface. More than one thread may lead to unpredictable crashes.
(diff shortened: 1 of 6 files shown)


geoffthemedio:
Why is this option necessary when there already a hostless, autostart, and load options?

o01eg:
It wasn't implemented for the server itself. I could rename `hostless.autostart.load` to `load` although.
Actually I didn't even check client code so miss there already was option for it.

geoffthemedio:
The load option should work as similarly as possible between the client and server.. not sure if autostart is assumed with load as is. Hostless is probably independent, though.

o01eg:
Made `load` common option for client and server.

server/ServerFSM.cpp
```diff
@@ -622,6 +622,59 @@ sc::result Idle::react(const Error& msg) {
 }
 
 sc::result Idle::react(const Hostless&) {
+    std::string autostart_load_filename = GetOptionsDB().Get<std::string>("load");
+    if (!autostart_load_filename.empty()) {
```
geoffthemedio: prefer to check that the filename is empty, and if so early exit, rather than having all the following code indented a level with nothing substantial after the indented block

server/ServerFSM.cpp
```diff
@@ -622,6 +622,59 @@ sc::result Idle::react(const Error& msg) {
...
+                lobby_data->m_players.push_back({Networking::INVALID_PLAYER_ID, player_setup_data});
+            }
+
+            // copy locally stored data to common server fsm context so it can be
+            // retreived in WaitingForMPGameJoiners
+            context<ServerFSM>().m_lobby_data = lobby_data;
+            context<ServerFSM>().m_player_save_game_data = player_save_game_data;
+            context<ServerFSM>().m_server_save_game_data = server_save_game_data;
+        } catch (const std::exception& e) {
+            throw e;
```
geoffthemedio: should be some logging here or at the start of this function to record what it's doing / what file was passed

server/ServerFSM.cpp
```diff
@@ -622,6 +622,59 @@ sc::result Idle::react(const Error& msg) {
 }
 
 sc::result Idle::react(const Hostless&) {
+    std::string autostart_load_filename = GetOptionsDB().Get<std::string>("load");
```
geoffthemedio, line 627: so the `--load "filename"` option only works in hostless mode on the server? Should indicate this in the help string. Even better, throw an error if `--load "something"` is specified when launching the server without `--hostless` also specified, so the user knows there's a problem.

server/ServerFSM.cpp

server/ServerFSM.cpp

server/ServerFSM.cpp
o01eg: Added both.

default/stringtables/en.txt
```diff
@@ -1928,7 +1928,7 @@ OPTIONS_DB_UI_SITREP_ICONSIZE
 Sets the sitrep icon width and height; default 16 (min 12, max 64).
 
 OPTIONS_DB_LOAD
-Loads the specified single-player save game.
+Loads and starts the specified single-player or multi-player save game. In a multi-player it requires hostless mode and no restrictions on a number of connected players because server will start without one.
```
geoffthemedio: In multi-player, it requires hostless mode and no restriction on the minimum number of connected players, because the server will start with no connected players.

server/ServerFSM.cpp
```diff
@@ -622,7 +623,63 @@ sc::result Idle::react(const Error& msg) {
...
 
 sc::result Idle::react(const Hostless&) {
-    return transit<MPLobby>();
+    TraceLogger(FSM) << "(ServerFSM) Idle.Hostless";
+    std::string autostart_load_filename = GetOptionsDB().Get<std::string>("load");
+    if (autostart_load_filename.empty())
+        return transit<MPLobby>();
+
+    if (GetOptionsDB().Get<int>("network.server.conn-human-empire-players.min") > 0) {
+        throw std::invalid_argument("Autostart load file was choosed but the server wasn't allowed to play game without connected players");
```
geoffthemedio: A save file to load and autostart in hostless mode was specified, but the server has a non-zero minimum number of connected players, so cannot be started without a connected player.

geoffthemedio: might be better to make a customized exception for this, and catch it in dmain and output something translated from the stringtable.

default/stringtables/en.txt

server/ServerFSM.cpp

PR #2413, opened 2019-03-27
Match nothing in OrderedAlternativesOf if no operand condition applied

agrrr3:
Oberlus found a bug which manifested because of the species target condition matching:
[Traitor fighter attacks its fleet](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11232&p=95060#p95060)

Somewhere along the way the OrderedAlternativesOf condition left the matching state untouched if no operand condition matched. It should actually move all matches to non_matches in this case.

### DIFF
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -10057,7 +10057,7 @@ void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context,
             // try the next operand
         }
 
-        // No operand condition was selected. State is restored...
+        // No operand condition was selected. State is restored. Nothing should be moved to matches input set
     } else /*(search_domain == MATCHES)*/ {
         // Check each operand condition on objects in the input matches and non_matches sets, until an operand condition matches an object.
         // If an operand condition is selected, apply it to the input matches set, moving non-matching candidates to non_matches.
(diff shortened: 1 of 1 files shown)


universe/Condition.cpp
```diff
@@ -10092,8 +10092,11 @@ void OrderedAlternativesOf::Eval(const ScriptingContext& parent_context,
             FCMoveContent(temp_objects, matches);
         }
 
-        // No operand condition was selected. Nothing needed to be applied. State is restored. Done.
+        // No operand condition was selected. State is restored...
     }
+    // ...  Nothing should match, so move all to non_matches input set.
+    non_matches.reserve(matches.size() + non_matches.size());
+    FCMoveContent(matches, non_matches);
```
geoffthemedio: Does it make sense to always do the samething here, independent if thesearch domain? If searching `NON_MATCHES`, the contents of `matches` shouldn't be modified.

agrrr3: Hm. Probably you are right. For NON-MATCHES it should be a no-op. Not sure what a test case would be

agrrr3: Hm. Probably you are right. For NON-MATCHES it should be a no-op. Not sure what a test case would be

agrrr3: Found a test case for NON_MATCHES domain
`Or [ Building OrderedAlternativesOf [ Ship ] ]`

and the no-op fixes the bug

PR #2415, opened 2019-04-04
Remove ALLOW_NO_STARLANES

o01eg:
I'm going to move `setup.starlane.frequency` option to common so I got issues with this compiled-in parameter. I don't see it ever changed or supported so I decided to remove it.

Forum thread: https://freeorion.org/forum/viewtopic.php?f=9&t=11236


geoffthemedio:
Looks to me like the Python universe generation scripts don't support the `GSETUP_NONE` value for the starlane density anyway.

https://github.com/freeorion/freeorion/blob/master/default/python/universe_generation/universe_tables.py#L140

PR #2420, opened 2019-04-10
Fix test in case if empire id isn't equal player id

o01eg:


geoffthemedio:
dunno what this should be doing, but I'll assume you do...

o01eg:
It was bad idea to use ignored data. I'm going to remove it next.

PR #2421, opened 2019-04-14
Remove unused player ID from PLAYER_STATUS message.

o01eg:
Breaks networking compatibility.

### DIFF
--- a/network/Message.cpp
+++ b/network/Message.cpp
@@ -404,15 +404,13 @@ Message TurnProgressMessage(Message::TurnProgressPhase phase_id) {
     return Message(Message::TURN_PROGRESS, os.str());
 }
 
-Message PlayerStatusMessage(int about_player_id,
-                            Message::PlayerStatus player_status,
+Message PlayerStatusMessage(Message::PlayerStatus player_status,
                             int about_empire_id)
 {
     std::ostringstream os;
     {
         freeorion_xml_oarchive oa(os);
-        oa << BOOST_SERIALIZATION_NVP(about_player_id)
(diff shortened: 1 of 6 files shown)


network/Message.cpp
```diff
@@ -404,15 +404,13 @@ Message TurnProgressMessage(Message::TurnProgressPhase phase_id) {
     return Message(Message::TURN_PROGRESS, os.str());
 }
 
-Message PlayerStatusMessage(int about_player_id,
-                            Message::PlayerStatus player_status,
+Message PlayerStatusMessage(Message::PlayerStatus player_status,
```
geoffthemedio, line 407: Swapping the order of these causes a lot of changes elsewhere. Is it necessary or beneficial?

geoffthemedio:
Could probably restore network compatibility with some player ID empire ID lookups, but is thete much reason to bother with it?

network/Message.cpp
o01eg: It was player ID, now there empire ID and both have type int. So I decide to swap order to don't send player ID as empire ID somewhere.

o01eg:
Current {player id, status, empire id} was compatible with {player id, status}. I decided to remove player ID part after #2420

geoffthemedio:
Not sure what that means re: my question... Point was you could look up player or empire IDs from the other.

o01eg:
Using player id for `PlayerStatus` is deprecated in https://github.com/freeorion/freeorion/pull/2352
This PR is a cleanup.

> Point was you could look up player or empire IDs from the other.

It possible to have players (observers and moderators) without empire and empire without connected players.

PR #2417, opened 2019-04-10
Hostless quickstart

o01eg:
Implements `quickstart` options for multiplayer hostless server.

PR #2422, opened 2019-04-14
Refactor game initialization so it allowed to have disconnected player

o01eg:
Extracted from #2417

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -540,8 +540,8 @@ void ServerApp::NewMPGameInit(const MultiplayerLobbyData& multiplayer_lobby_data
     // available (human) and names (for AI clients which didn't have an ID
     // before now because the lobby data was set up without connected/established
     // clients for the AIs.
-    std::map<int, PlayerSetupData> player_id_setup_data;
     const auto& player_setup_data = multiplayer_lobby_data.m_players;
+    std::vector<PlayerSetupData> psds;
 
     for (const auto& entry : player_setup_data) {
         const PlayerSetupData& psd = entry.second;
@@ -561,13 +561,23 @@ void ServerApp::NewMPGameInit(const MultiplayerLobbyData& multiplayer_lobby_data
                 const PlayerConnectionPtr player_connection = *established_player_it;
                 if (player_connection->PlayerID() == player_id)
                 {
-                    player_id_setup_data[player_id] = psd;
+                    PlayerSetupData new_psd = psd;
+                    new_psd.m_player_id = player_id;
+                    psds.emplace_back(std::move(new_psd));
                     found_matched_id_connection = true;
                     break;
                 }
             }
-            if (!found_matched_id_connection)
-                ErrorLogger() << "ServerApp::NewMPGameInit couldn't find player setup data for human player with id: " << player_id;
+            if (!found_matched_id_connection) {
+                if (player_id != Networking::INVALID_PLAYER_ID) {
+                    ErrorLogger() << "ServerApp::NewMPGameInit couldn't find player setup data for human player with id: " << player_id << " player name: " << psd.m_player_name;
+                } else {
+                    // There is no player currently connected for the current setup data. A player
+                    // may connect later, at which time they may be assigned to this data or the
+                    // corresponding empire.
+                    psds.push_back(psd);
+                }
+            }
 
         } else if (psd.m_client_type == Networking::CLIENT_TYPE_AI_PLAYER) {
             // All AI player setup data, as determined from their client type, is
@@ -585,7 +595,9 @@ void ServerApp::NewMPGameInit(const MultiplayerLobbyData& multiplayer_lobby_data
                 {
                     // assign name-matched AI client's player setup data to appropriate AI connection
                     int player_id = player_connection->PlayerID();
-                    player_id_setup_data[player_id] = psd;
+                    PlayerSetupData new_psd = psd;
+                    new_psd.m_player_id = player_id;
+                    psds.emplace_back(std::move(new_psd));
                     found_matched_name_connection = true;
                     break;
                 }
@@ -600,12 +612,6 @@ void ServerApp::NewMPGameInit(const MultiplayerLobbyData& multiplayer_lobby_data
             // created for them.
             ErrorLogger() << "ServerApp::NewMPGameInit skipping unsupported client type in player setup data";
         }
-    }
-
-    std::vector<PlayerSetupData> psds;
-    for (auto& id_and_psd : player_id_setup_data) {
-        id_and_psd.second.m_player_id = id_and_psd.first;
-        psds.push_back(id_and_psd.second);
     }
 
     NewGameInitConcurrentWithJoiners(multiplayer_lobby_data, psds);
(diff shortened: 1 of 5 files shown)


server/ServerFSM.cpp
```diff
@@ -2167,62 +2164,53 @@ sc::result WaitingForMPGameJoiners::react(const JoinGame& msg) {
...
-            std::string ai_prefix = UserString("AI_PLAYER") + "_";
-            while (player_name.compare(0, ai_prefix.size(), ai_prefix) == 0)
-                player_name.erase(0, ai_prefix.size());
-            if(player_name.empty())
-                player_name = "_";
+        // Remove AI prefix to distinguish Human from AI.
+        std::string ai_prefix = UserString("AI_PLAYER") + "_";
+        while (player_name.compare(0, ai_prefix.size(), ai_prefix) == 0)
+            player_name.erase(0, ai_prefix.size());
+        if(player_name.empty())
```
geoffthemedio: space between `if` and `(`

server/ServerApp.cpp
```diff
@@ -550,13 +550,21 @@ void ServerApp::NewMPGameInit(const MultiplayerLobbyData& multiplayer_lobby_data
...
                     break;
                 }
             }
-            if (!found_matched_id_connection)
-                ErrorLogger() << "ServerApp::NewMPGameInit couldn't find player setup data for human player with id: " << player_id;
+            if (!found_matched_id_connection) {
+                if (player_id != Networking::INVALID_PLAYER_ID) {
+                    ErrorLogger() << "ServerApp::NewMPGameInit couldn't find player setup data for human player with id: " << player_id << " player name: " << psd.m_player_name;
+                } else {
+                    // player connect later
```
geoffthemedio: There is no player currently connected for the current setup data. A player may connect later, at which time they may be assigned to this data or the corresponding empire.

server/ServerFSM.cpp
```diff
@@ -2167,62 +2164,53 @@ sc::result WaitingForMPGameJoiners::react(const JoinGame& msg) {
...
-                    collision = true;
-                } else {
-                    for (std::pair<int, PlayerSetupData>& plr : m_lobby_data->m_players) {
-                        if (plr.second.m_empire_name == new_player_name) {
-                            collision = true;
-                            break;
-                        }
+        bool collision = true;
+        std::size_t t = 1;
+        while (collision &&
```
geoffthemedio, line 2220: Maybe a comment here explaining what is going on with name collisions and authorization to resolve them.

server/ServerFSM.cpp
```diff
@@ -2167,62 +2164,53 @@ sc::result WaitingForMPGameJoiners::react(const JoinGame& msg) {
...
+                new_player_name = player_name + std::to_string(++t); // start alternative names from 2
+        }
 
-            if (!player_connection->HasAuthRole(Networking::ROLE_CLIENT_TYPE_PLAYER)) {
-                player_connection->SendMessage(ErrorMessage(UserStringNop("ERROR_CLIENT_TYPE_NOT_ALLOWED"), true));
-                server.Networking().Disconnect(player_connection);
-                return discard_event();
-            }
+        if (collision) {
+            player_connection->SendMessage(ErrorMessage(str(FlexibleFormat(UserString("ERROR_PLAYER_NAME_ALREADY_USED")) % original_player_name),
```
geoffthemedio, line 2244: This isn't great as the error message is being formatted server-side, so a player with a different language set won't see the message in their chosen language...

server/ServerFSM.cpp
```diff
@@ -2167,62 +2164,53 @@ sc::result WaitingForMPGameJoiners::react(const JoinGame& msg) {
...
+                                                        true));
+            server.Networking().Disconnect(player_connection);
```
geoffthemedio, line 2246: Why disconnecting here if code above assigns a unique name?

server/ServerFSM.cpp
```diff
@@ -2167,62 +2164,53 @@ sc::result WaitingForMPGameJoiners::react(const JoinGame& msg) {
...
+            return discard_event();
+        }
 
-            fsm.EstablishPlayer(player_connection, new_player_name, client_type, client_version_string, roles);
+        if (!player_connection->HasAuthRole(Networking::ROLE_CLIENT_TYPE_PLAYER)) {
+            player_connection->SendMessage(ErrorMessage(UserStringNop("ERROR_CLIENT_TYPE_NOT_ALLOWED"), true));
+            server.Networking().Disconnect(player_connection);
```
geoffthemedio: Unless a player connection has no roles allowed, could assign a new role instead of disconnecting.

universe/UniverseGenerator.cpp
```diff
@@ -872,13 +872,12 @@ void InitEmpires(const std::map<int, PlayerSetupData>& player_setup_data)
...
-        if (player_id == Networking::INVALID_PLAYER_ID)
-            ErrorLogger() << "InitEmpires player id (" << player_id << ") is invalid";
+        int         empire_id =     entry.first;
+        if (empire_id == ALL_EMPIRES)
+            ErrorLogger() << "InitEmpires empire id (" << empire_id << ") is invalid";
 
-        // use player ID for empire ID so that the calling code can get the
-        // correct empire for each player ID  in player_setup_data
-        int         empire_id =     player_id;
+        // use map key for empire ID so that the calling code can get the
```
geoffthemedio: Should this comment be before the line above that assigns `empire_id`?

universe/UniverseGenerator.h
```diff
@@ -81,7 +81,7 @@ void GenerateStarlanes(int max_jumps_between_systems, int max_starlane_length);
 bool SetEmpireHomeworld(Empire* empire, int planet_id, std::string species_name);
 
 /** Creates Empires objects for each entry in \a player_setup_data with
- * empire id equal to the specified player ids (so that the calling code
+ * empire id equal to the specified map key (so that the calling code
```
geoffthemedio: empire ids equal to the corresponding map keys (...

server/ServerFSM.cpp

server/ServerApp.cpp

server/ServerFSM.cpp
o01eg: Added.

server/ServerFSM.cpp
o01eg: This collision check code is the same as `MPLobby::JoinGame`.
To implement client-side formatting there should be something like mentioned here: https://github.com/freeorion/freeorion/pull/2393#discussion_r261364566

server/ServerFSM.cpp
o01eg: If `t` overflow `m_lobby_data->m_players.size() + server.Networking().GetCookiesSize() + 1` then loop will exit with collision.

server/ServerFSM.cpp

universe/UniverseGenerator.h
o01eg: Fixed

universe/UniverseGenerator.cpp
o01eg: Fixed.

server/ServerFSM.cpp
geoffthemedio: I generally assume the such strings should / would be sent as a stringtable key or enum and the string formatted client-side, but not using the `[[KEY]]` type markup.

server/ServerFSM.cpp
geoffthemedio: add a comment and perhaps a log message explaining that?

server/ServerFSM.cpp
o01eg: What if send array of string instead of one string in the ERROR_MSG message?

server/ServerFSM.cpp
o01eg: Added comment.

PR #2424, opened 2019-04-15
Options game rules

o01eg:
For each game rule creates two options:

- `setup.rules.{NAME}` - value of the same type to override default in `game_rules.focs.txt`
- `setup.rules.server-locked.{NAME}` - if the rule locked on the server

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -1325,7 +1325,8 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
             m_lobby_data->m_native_freq    = incoming_lobby_data.m_native_freq;
             m_lobby_data->m_ai_aggr        = incoming_lobby_data.m_ai_aggr;
 
-            // copy only allowed rules
+            // copy rules from incoming lobby data to server lobby data, only if those rules are
+            // not locked by the server
             for (const auto& incoming_rule : incoming_lobby_data.m_game_rules) {
                 if (GetOptionsDB().OptionExists("setup.rules.server-locked." + incoming_rule.first) &&
                     !GetOptionsDB().Get<bool>("setup.rules.server-locked." + incoming_rule.first))
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -9,6 +9,7 @@
 #include "../network/ServerNetworking.h"
 #include "../network/Message.h"
 #include "../util/Directories.h"
+#include "../util/GameRules.h"
 #include "../util/i18n.h"
 #include "../util/Logger.h"
 #include "../util/LoggerWithOptionsDB.h"
@@ -747,6 +748,7 @@ MPLobby::MPLobby(my_context c) :
     ServerApp& server = Server();
     server.InitializePython();
     server.LoadChatHistory();
+    m_lobby_data->m_game_rules = GetGameRules().GetRulesAsStrings();
     const SpeciesManager& sm = GetSpeciesManager();
     if (server.IsHostless()) {
         DebugLogger(FSM) << "(ServerFSM) MPLobby. Fill MPLobby data from the previous game.";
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Not sure "editabe" is the right name... The options are all editable, as otherwise they wouldn't be options. The point is whether to show them in the GUI options screen when generating that programatically?
I don't like having to add `.value` to every option if specifying on the command line...

o01eg:
> Not sure "editabe" is the right name... The options are all editable, as otherwise they wouldn't be options. The point is whether to show them in the GUI options screen when generating that programatically?

Yes. Main idea of "editable" I have is to allow server's owner decide which rules could be changed. What if name it "locked"?

> I don't like having to add .value to every option if specifying on the command line...

Does it work in options XML if there text of value and some data in children elements?

geoffthemedio:
`server-locked` perhaps?

I think XMLDoc supports children or text in an element but not both.

o01eg:
> I don't like having to add .value to every option if specifying on the command line...

Fixed. I've changed order of option's path part.

o01eg:
Implemented server checks.

server/ServerFSM.cpp
```diff
@@ -1322,7 +1324,15 @@ sc::result MPLobby::react(const LobbyUpdate& msg) {
             m_lobby_data->m_monster_freq   = incoming_lobby_data.m_monster_freq;
             m_lobby_data->m_native_freq    = incoming_lobby_data.m_native_freq;
             m_lobby_data->m_ai_aggr        = incoming_lobby_data.m_ai_aggr;
-            m_lobby_data->m_game_rules     = incoming_lobby_data.m_game_rules;
+
+            // copy only allowed rules
```
geoffthemedio: elaborate comment a bit...

copy rules from incoming lobby data to server lobby data, only if those rules are not locked by the server.

util/GameRules.h
```diff
@@ -90,7 +90,9 @@ class FO_COMMON_API GameRules {
     //@}
 
     /** \name Mutators */ //@{
-    /** adds a rule, optionally with a custom validator */
+    /** adds a rule, optionally with a custom validator.
+        Register option setup.rules.{NAME} to override default value and
```
geoffthemedio: Registering a rule doesn't "override" the default value, it creates the rule and sets the default value.

Also, it's not clear if "Register option" is an instruction to the reader for something they could do, or if it is meant to indicate what this function does. Mixing "adds" and "register", and "rule" and "option", which appear to meant the same things, is also confusing.

util/GameRules.h
```diff
@@ -100,9 +102,17 @@ class FO_COMMON_API GameRules {
         auto it = m_game_rules.find(name);
         if (it != m_game_rules.end())
             throw std::runtime_error("GameRules::Add<>() : Rule " + name + " was added twice.");
-        m_game_rules[name] = Rule(RuleTypeForType(T()), name, default_value, default_value,
-                                  description, validator.Clone(), engine_interal, category);
-        DebugLogger() << "Added game rule named " << name << " with default value " << default_value;
+        if (!GetOptionsDB().OptionExists("setup.rules.server-locked." + name)) {
+            GetOptionsDB().Add<bool>("setup.rules.server-locked." + name, description, false);
```
geoffthemedio, line 106: Is there any security risk if the `server-locked` rules exist on the client and server, if the client sends lobby data that contains a `server-locked` rule set to false or true? Could a modified client add aa rule with `server-locked` in the name, even if one doesn't exist on the server, and thereby cause the lobby data sent to the server to copy that rule into the server's rules, and thus lock another rule that wasn't supposed to be locked on the server?

o01eg: `server-locked.*` is an option, not an game rule. It isn't transferred between client and server as I know.

server/ServerFSM.cpp

util/GameRules.h
o01eg: Game rules and options are different things. I've removed "Register".

util/GameRules.h
geoffthemedio: Debug level options do transfer, but that is set up explicitly in the logging options UI code and has a dedicated client server message type.

PR #2427, opened 2019-04-16
Add EXTERNAL_RESERVED messages for external applications

o01eg:
Reserves message type so other application like web-server could interact with freeoriond server without changes in the protocol.

### DIFF
--- a/network/Message.h
+++ b/network/Message.h
@@ -104,7 +104,8 @@ public:
         ELIMINATE_SELF,         ///< sent by client to server if the player wants to resign
         UNREADY,                ///< sent by client to server to revoke ready state of turn orders and sent by server to client to acknowledge it
         TURN_PARTIAL_ORDERS,    ///< sent to the server by a client that has changes in orders to be stored
-        TURN_TIMEOUT            ///< sent by server to client to notify about remaining time before turn advance
+        TURN_TIMEOUT,           ///< sent by server to client to notify about remaining time before turn advance
+        EXTERNAL_RESERVED       ///< sent to the server by an external application for custom reaction
     )
 
     GG_CLASS_ENUM(TurnProgressPhase,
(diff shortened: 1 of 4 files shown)


geoffthemedio:
What is the rationale for a gerneric external source message type? Why not have various messages that can be (somehow) sent from an external program, which do specific things?

o01eg:
New messages require protocol changes, don't they? I have no such external applications yet but when I'll get them I'll prefer to use them without waiting for new stable release.

geoffthemedio:
Probably, but a generic message doesn't make much sense to me...  If the new message is the highest enum value , it might not affect the others anyway...

o01eg:
Rebased. Fixed collisions.

geoffthemedio:
Delay is not related to merge conflicts...

geoffthemedio:
I think external application messaging would be better done by extending the UDP discovery server, rather than having an external application connect as a player by TCP. One can already set up, for example, a Python script to send a UDP message to the IP and port of a freeoriond process, with the server inquiry message, and can get back the UDP reply message. The discovery server can be extended to supply an interface for other queries.
 
In the Python interpreter:
````
>>> import socket
>>> UDP_IP = '127.0.0.1'
>>> UDP_PORT = 12345
>>> MESSAGE = 'Yo, can I play Free-O here, dog?'.encode('utf-8')
>>> s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
>>> s.sendto(MESSAGE, (UDP_IP, UDP_PORT))
32
>>> data, addr = s.recvfrom(1024, 0)
>>> print(data)
b'Word!Fiona'
````

In freeoriond log:
````
15:02:10.061132 {0x00003020} [debug] network : ServerNetworking.cpp:432 : Discovery server received from: 127.0.0.1:50434
message: Yo, can I play Free-O here, dog?
replied: Word!Fiona
````

geoffthemedio:
https://github.com/freeorion/freeorion/pull/2536 probably supercedes this

PR #2431, opened 2019-04-25
AI Unit tests

Morlic-fo:
Opening PR to provoke CI.
Will eventually contain AI unit tests.

### DIFF
--- a/default/python/AI/test/abstract_test_bases.py
+++ b/default/python/AI/test/abstract_test_bases.py
@@ -2,15 +2,8 @@ import unittest
 
 
 class PropertyTester(unittest.TestCase):
-    # hack: we don't want unittest to run this abstract test base.
-    # Therefore, we override this classes run function to do nothing.
     class_to_test = None
     properties = {}
-
-    def __init__(self, *args, **kwargs):
-        super(PropertyTester, self).__init__(*args, **kwargs)
-        if self.__class__ == PropertyTester:
-            self.run = lambda self, *args, **kwargs: None
 
     @classmethod
     def generate_tests(cls):
--- a/default/python/AI/test/abstract_test_bases.py
+++ b/default/python/AI/test/abstract_test_bases.py
@@ -4,6 +4,7 @@ import unittest
 class PropertyTester(unittest.TestCase):
     class_to_test = None
     properties = {}
+    longMessage = True
 
     @classmethod
     def generate_tests(cls):
--- a/default/python/AI/test/abstract_test_bases.py
+++ b/default/python/AI/test/abstract_test_bases.py
@@ -6,6 +6,10 @@ class PropertyTester(unittest.TestCase):
     properties = {}
     longMessage = True
 
+    TYPE = "type"
+    READ_ONLY = "read_only"
+    ADDITIONAL_TESTS = "additional_tests"
+
     @classmethod
     def generate_tests(cls):
         for property in cls.properties:
--- a/default/python/AI/test/abstract_test_bases.py
+++ b/default/python/AI/test/abstract_test_bases.py
@@ -42,3 +42,20 @@ class PropertyTester(unittest.TestCase):
 
         for test in additional_tests:
             test[0](self, return_value, *test[1:])
+
+
+class DumpTester(unittest.TestCase):
+
+    def test_dump(self):
+        for obj in self.objects_to_test:
+            retval = obj.dump()
+            self.assertIsInstance(retval, str)
+
+    def test_dump_takes_no_args(self):
+        for obj in self.objects_to_test:
+            for arg in (int(), float(), str(), None):
+                self.assertRaises(Exception, obj.dump, arg)
+
+    def setUp(self):
+        self.objects_to_test = []
+        raise NotImplementedError
--- a/default/python/AI/test/abstract_test_bases.py
+++ b/default/python/AI/test/abstract_test_bases.py
@@ -1,0 +1,46 @@
+import unittest
+
+
+class PropertyTester(unittest.TestCase):
+    # hack: we don't want unittest to run this abstract test base.
+    # Therefore, we override this classes run function to do nothing.
+    class_to_test = None
+    properties = {}
+
+    def __init__(self, *args, **kwargs):
+        super(PropertyTester, self).__init__(*args, **kwargs)
+        if self.__class__ == PropertyTester:
+            self.run = lambda self, *args, **kwargs: None
+
+    @classmethod
+    def generate_tests(cls):
+        for property in cls.properties:
+            setattr(cls, "test_property_%s" % property,
+                    PropertyTester._test_case_generator(property, cls.properties[property]))
+
+    @staticmethod
+    def _test_case_generator(property, kwargs):
+        def _test_case(self):
+            for obj in self.objects_to_test:
+                self.__test_property(property, obj, **kwargs)
+        return _test_case
+
+    def __test_property(self, property_to_test, obj, type=None, read_only=True, additional_tests=[]):
+        self.assertIn(property_to_test, dir(obj))
+        property_to_test = getattr(self.class_to_test, property_to_test)
+        return_value = property_to_test.fget(obj)
+        if type is not None:
+            self.assertIsInstance(return_value, type)
+        else:
+            self.assertIsNone(return_value)
+
+        if read_only:
+            self.assertIsNone(property_to_test.fset)
+        else:
+            property_to_test.fset(obj, type())
+            self.assertEqual(property_to_test.fget(obj), type())
+            property_to_test.fset(obj, return_value)
+            self.assertEqual(property_to_test.fget(obj), return_value)
+
+        for test in additional_tests:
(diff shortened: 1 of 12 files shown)


o01eg:
It could be enabled with `--testing` server option so will work only in CI.

Cjkjvfnby:
21 century and we still cannot use the power of code reuse.

As a simple solution, it is possible to add `pytest` files to the AI folder.

Cjkjvfnby:
You can try to split the generated test and handwritten tests to different classes. It will make the code more simple.

Cjkjvfnby:
Please add a comment to PR with the following:

- test output in logs
- example of the test failures for generated tests and for handwritten

Morlic-fo:
> 21 century and we still cannot use the power of code reuse.
> 
> As a simple solution, it is possible to add `pytest` files to the AI folder.

While ```pytest``` is great, I couldn't bring it to work on windows from inside the FreeOrionCA process which I think is the place to run the tests which deal with the python API. It would probably require changes to the shipped python version. ```unittest``` is part of the standard library instead of a 3rd party library and works without further changes. Note that ```unittest``` tests are supported in ```pytest```.

If you can think of a better approach to this, please let me know with more details.

o01eg:
Is it abandoned?  It would be nice to have AI unit tests to guide python3 migration.

o01eg:
I've fixed conflicts and rebased it onto master https://github.com/o01eg/freeorion/tree/AIUnitTests

geoffthemedio:
Do you need help with something for this? Is it still work in progress?

python/UniverseWrapper.cpp
```diff
@@ -473,7 +473,7 @@ namespace FreeOrionPython {
             .def("hasTag",                      &UniverseObject::HasTag)
             .add_property("meters",             make_function(ObjectMeters,                 return_internal_reference<>()))
             .def("getMeter",                    make_function(ObjectGetMeter,               return_internal_reference<>()))
-            .def("dump",                        &UniverseObject::Dump,                      return_value_policy<return_by_value>(), "Returns string with debug information, use '0' as argument.")
+            .def("dump",                        +[](const UniverseObject& obj) {return obj.Dump(0); })
```
o01eg: Looks like it required for GCC, but fails at MSVC. Can it be rewritten to work with both compilers?

geoffthemedio: Not sure, but could use an `#ifdef` if nothing else works

geoffthemedio:
````
diff --git a/python/UniverseWrapper.cpp b/python/UniverseWrapper.cpp
index a8b9c530e..2cdd5ac0e 100644
--- a/python/UniverseWrapper.cpp
+++ b/python/UniverseWrapper.cpp
@@ -51,6 +51,9 @@ namespace boost {
 namespace {
     void                    DumpObjects(const Universe& universe)
     { DebugLogger() << universe.Objects().Dump(); }
+    std::string             ObjectDump(const UniverseObject& obj) {
+        return obj.Dump(0);
+    }
 
     // We're returning the result of operator-> here so that python doesn't
     // need to deal with std::shared_ptr class.
@@ -212,7 +215,7 @@ namespace {
         }
         return results;
     }
-    boost::function<std::vector<int> (const ShipDesign&)> AttackStatsFunc =                 &AttackStatsP;
+    boost::function<std::vector<int> (const ShipDesign&)> AttackStatsFunc =                     &AttackStatsP;
 
     std::vector<ShipSlotType> HullSlots(const HullType& hull) {
         std::vector<ShipSlotType> retval;
@@ -473,7 +476,7 @@ namespace FreeOrionPython {
             .def("hasTag",                      &UniverseObject::HasTag)
             .add_property("meters",             make_function(ObjectMeters,                 return_internal_reference<>()))
             .def("getMeter",                    make_function(ObjectGetMeter,               return_internal_reference<>()))
-            .def("dump",                        +[](const UniverseObject& obj) {return obj.Dump(0); })
+            .def("dump",                        make_function(&ObjectDump,                  return_value_policy<return_by_value>()),    "Returns string with debug information.")
         ;
 
         ///////////////////
````

Comment (o01eg) [reply python/UniverseWrapper.cpp]: Yep, it looks better.

I've tried to split `Dump` without arguments and `Dump` with offset to two different methods but it requires massive changes.

Comment (geoffthemedio) [reply python/UniverseWrapper.cpp]: Offset-parameterized `Dump` isn't needed for the Python API, I think...

Comment (Morlic-fo) [reply python/UniverseWrapper.cpp]: That's funny. IIRC, this did pass CI back when I initially commited this and the lambda expression builds fine both on MSVC2015 and MSVC2017 on my system.

**Comment by Morlic-fo** [comment]
> Do you need help with something for this?

Just need to find time.

**Comment by geoffthemedio** [comment]
If I understand correctly, this is wanted or needed for getting Python3 migration done, and I'd like ro have that working soon since support for Python2 is ending and I suspect we'll start getting complaints or issues about that.

**Comment by o01eg** [comment]
@geoffthemedio It isn't required to migrate, but it better to have unit tests before migration.

File: python/UniverseWrapper.cpp (hunk 1 above)
Comment (o01eg) [reply python/UniverseWrapper.cpp]: @Morlic-fo Can you try proposed changes?

Comment (geoffthemedio) [reply python/UniverseWrapper.cpp]: Not sure what you mean. I think I tested with the diff above before posting it...

Comment (o01eg) [reply python/UniverseWrapper.cpp]: @geoffthemedio I mean if @Morlic-fo add it to this PR.

**Comment by o01eg** [comment]
@Morlic-fo bump!

**Comment by o01eg** [comment]
Should be `ReadOnlyDict` items compared as `set` instead of `list`?

```python
=================================== FAILURES ===================================
___________________________ test_conversion_to_dict ____________________________

    def test_conversion_to_dict():
        read_only_dict = ReadOnlyDict(dict_content)
        normal_dict = dict(read_only_dict)
        assert len(normal_dict) == len(dict_content)
>       assert list(normal_dict.items()) == list(dict_content.items())
E       AssertionError: assert [(1, -1), ('a..., 1, (1, 2)))] == [(1, -1), ('ab..., (1.09, 1.1)]
E         At index 2 diff: (1.09, 1.1) != ((1, 2, 3), ('a', 1, (1, 2)))
E         Use -v to get the full diff
```

Morlic-fo:
> Should be `ReadOnlyDict` items compared as `set` instead of `list`?

Yes, seems like that one instance was missed in #2711 while fixed at other places.

default/python/AI/test/test_universe.py
```diff
@@ -503,6 +505,278 @@ def setUp(self):
...
+            self.assertEquals(total_slots, obj.numSlots)
+
+            with self.assertRaises(Exception):
+                obj.numSlotsOfSlotType(2)
+            with self.assertRaises(Exception):
+                obj.numSlotsOfSlotType("")
+
+    def test_productionCost(self):
+        for obj in self.objects_to_test:
+            retval = obj.productionCost(fo.empireID(), INVALID_ID)
```
geoffthemedio: I'll probably merge some changes soon that add another `INVALID_ID` parameter to this and similar calls: https://github.com/freeorion/freeorion/commit/6f797a073b3e7aada63f8b577997383007becc5e

Morlic-fo: Thanks for the heads-up, might spare me some headache trying to figure out why some tests may suddenly fail.

o01eg:
@Morlic-fo Could you rebase it to see if it passes CI?

default/python/AI/test/abstract_test_bases.py
```diff
@@ -0,0 +1,61 @@
+import unittest
+
+
+class PropertyTester(unittest.TestCase):
```
Cjkjvfnby, line 4: Please add a docstring. 

Mention that object is mutated.
In my mock generator, I have encountered a problem that not all classes can be instantiated by python.  I used function to get real objecst from the game.   In that case mutation might be a problem and require ever cleanup or game exit after tests.

default/python/AI/test/test_galaxy_setup_data_wrapper.py
```diff
@@ -0,0 +1,127 @@
...
+            TYPE: int,
+            ADDITIONAL_TESTS: [greater_or_equal(1)],
+        },
+        "shape": {
+            TYPE: fo.galaxyShape,
+            ADDITIONAL_TESTS: [is_not_equal(fo.galaxyShape.invalid)],
+        },
+        "age": {
+            TYPE: fo.galaxySetupOption,
+            ADDITIONAL_TESTS: [is_not_equal(fo.galaxyShape.invalid)],
```
Cjkjvfnby, line 31: You use constant from one class to test result of the other.  `fo.galaxyShape.invalid`  for `fo.galaxySetupOption`.  Same in other tests.

Cjkjvfnby (commented):
Looks great. Only some cosmetic comments.

Cjkjvfnby:
@Morlic-fo, Bump!

PR #2433, opened 2019-04-28
Try to fix Bad file descriptor error in CI

o01eg:
Fixes CI faliure:
```
15:39:25.175479 [error] network : ClientNetworking.cpp:416 : ConnectToServer() : unable to connect to server at localhost due to exception: close: Bad file descriptor
/freeorion/test/system/SmokeTestHostless.cpp:142: error: fatal error in "hostless_server": critical check ConnectToServer("localhost") failed
```

It simply initializes new socket instead of failed one.


geoffthemedio:
Did this do what you hoped?

o01eg:
Yes, I've launched tests with this change three times and it didn't fail.

PR #2434, opened 2019-04-29
Add testing flag to AI to use it in tests

o01eg:
For #2431 to launch AI unittests in test environment only.


Morlic-fo:
Sure, useful for the first implementation. As far as simple tests are concerned which do not involve issuing orders and then checking next turn game state, this PR is good enough. 

My long-term plan was to introduce an additional layer of tests where the AI exclusively runs its unit tests rather than its normal code, possibly in a predefined handcoded universe where orders like colonization, invasion, bombardment, movement etc. could be verified. The current smoke tests would then be used exclusively to test the actual AI decision-making code for errors.

Issue #2412, opened 2019-03-24
Error - The connection to the server has been lost [v0.4.8 2018-08-23 win]


iapicca:
HI, 
when I go for a new turn sistem show me the error below,
I'm also attaching the .sav file if you wanna try yourself.
I don't mind losing the game, I just hope this will help you somehow.
Keep up the good work :)
<img width="960" alt="error_freeOrion" src="https://user-images.githubusercontent.com/45637986/54886800-b0368480-4e94-11e9-964c-e12ac81357f3.PNG">

[9.2.sav.zip](https://github.com/freeorion/freeorion/files/3000941/9.2.sav.zip)

Vezzra:
@iapicca, sorry for the much delayed response, looks like the majority of our team is currently occupied with real life things.

Anyway, thanks for the report. However, you're using the last stable release, there has been done a lot of work since then. Can you try and reproduce the issue with a current test release?

o01eg:
Without debug symbols I got:
```c++
Thread 1 "freeoriond" received signal SIGSEGV, Segmentation fault.
0x00007ffff7802850 in Empire::SetProductionQuantityAndBlocksize(int, int, int) () from /usr/lib64/freeorion/libfreeorioncommon.so
(gdb) bt
#0  0x00007ffff7802850 in Empire::SetProductionQuantityAndBlocksize(int, int, int) () from /usr/lib64/freeorion/libfreeorioncommon.so
#1  0x00007ffff7a2b898 in ProductionQueueOrder::ExecuteImpl() const () from /usr/lib64/freeorion/libfreeorioncommon.so
#2  0x00007ffff7a28a20 in Order::Execute() const () from /usr/lib64/freeorion/libfreeorioncommon.so
#3  0x0000555555789739 in ?? ()
#4  0x00005555557c3267 in ?? ()
#5  0x00005555557d8ae5 in ?? ()
#6  0x00005555557ad63b in ?? ()
#7  0x00005555557ad7a8 in ?? ()
#8  0x0000555555790562 in ?? ()
#9  0x00005555557a1318 in ?? ()
#10 0x00005555556480e5 in ?? ()
#11 0x0000555555637866 in ?? ()
#12 0x000055555577e222 in ?? ()
#13 0x0000555555611a91 in ?? ()
#14 0x00007ffff43934eb in __libc_start_main () from /lib64/libc.so.6
#15 0x0000555555613c6a in _start ()
```

geoffthemedio:
[...]

PR #2436, opened 2019-05-03
Fix potential out of bounds access in Production/ResearchQueue

Morlic-fo:
It was possible to access the ProductionQueue/ResearchQueue out of bounds in Release mode. For example, following test wouldn't necessarily raise an error (but possibly crash the AI) before the fix. Now, as expected, an IndexError is raised instead.
```
[03 Mai 22:29:36] AI_1: Entering debug mode
Print 'stop' to exit.
Local variables:
  ai  aistate
  u   universe
  e   empire
[03 Mai 22:29:45] Menschlicher_Spieler: e.productionQueue[999]
[03 Mai 22:29:45] AI_1: Traceback (most recent call last):
  File "<input>", line 1, in <module>
IndexError: Tried to access ProductionQueue element out of bounds
```

### DIFF
--- a/Empire/ResearchQueue.cpp
+++ b/Empire/ResearchQueue.cpp
@@ -145,7 +145,8 @@ ResearchQueue::const_iterator ResearchQueue::find(const std::string& tech_name) 
 }
 
 const ResearchQueue::Element& ResearchQueue::operator[](int i) const {
-    assert(0 <= i && i < static_cast<int>(m_queue.size()));
+    if (i < 0 || i >= static_cast<int>(m_queue.size()))
+        throw std::out_of_range("Tried to access ResearchQueue element out of bounds");
     return m_queue[i];
 }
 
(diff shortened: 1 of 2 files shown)


o01eg:
Does it fix  #2412? I suppose issue with using index to address item in the queues while it depends on orders' sorting.

Empire/ResearchQueue.cpp
```diff
@@ -323,7 +324,8 @@ void ResearchQueue::insert(iterator it, const std::string& tech_name, bool pause
 { m_queue.insert(it, Element(tech_name, m_empire_id, 0.0f, -1, paused)); }
 
 void ResearchQueue::erase(iterator it) {
-    assert(it != end());
+    if (it == end())
```
geoffthemedio, line 327: is this necessarily a crash and/or preferable to throw for, rather than it being a no-op?

Morlic-fo: I have no opinion on this. I implemented the throw considering the previous assert statement. If you think a no-op return would make more sense here, I could change it.

Morlic-fo:
> Does it fix #2412?

I don't think so. It may or may not throw an std::out_of_range exception with this PR rather than an assertion failure or the seemingly optimized out [std::runtime_error](https://github.com/freeorion/freeorion/blob/master/Empire/Empire.cpp#L1148).

Without exception handling, I would still expect a crash.

PR #2437, opened 2019-05-03
Add check if building is enqueueable before issuing an enqueue order

Morlic-fo:
Note that the AI did check if it was producible. However, items might not be enqueueable even if they are producible.

Attempt to fix underlying cause of #2412. Not sure if actually relevant.

### DIFF
--- a/AI/AIInterface.cpp
+++ b/AI/AIInterface.cpp
@@ -262,6 +262,11 @@ namespace AIInterface {
             return 0;
         }
 
+        if (!empire->EnqueuableItem(BT_BUILDING, item_name, location_id)) {
+            ErrorLogger() << "IssueEnqueueBuildingProductionOrder : specified item_name and location_id that don't indicate an item that can be enqueued at that location";
+            return 0;
+        }
+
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Description is misleading, I think. This doesn't "let" the AI do anything; it adds a check within the Python API implementation that rejects an order to enqueue, before it is issued, if the enqueue would fail.

Vezzra:
What's the status of this? Can it be merged?

Morlic-fo:
Reworded.

AI/AIInterface.cpp
```diff
@@ -262,6 +262,11 @@ namespace AIInterface {
             return 0;
         }
 
+        if (!empire->EnqueuableItem(BT_BUILDING, item_name, location_id)) {
+            ErrorLogger() << "IssueEnqueueBuildingProductionOrder : specified item_name and location_id that don't indicate an item that can be enqueued at that location";
```
geoffthemedio, line 266: Would perhaps be more useful to also output the relevant item name and location in the log message.

geoffthemedio:
Consider adding details to the log message in another commit...

PR #2441, opened 2019-05-05
Initial turn timeout implementation

o01eg:
This PR add three options to manage turn timeout:

* `network.server.turn-timeout.first-turn-time` - absolute time in format "2002-01-20 23:59:59" when first turn advance occured. Requires `fixed-interval` to be enabled
* `network.server.turn-timeout.max-interval` - maximum interval in seconds between two subsequent turn advances
* `network.server.turn-timeout.fixed-interval` - if enabled intervals should be strictly equals to have predicted turn advance time. Without it turn interval could be shorter if all players are ready or longer if turn processing time exceeds timeout.

Each new turn or for each new player connection it sends TURN_TIMEOUT message with seconds remaining before turn advance or 0, if disabled.

New UI control shows remain time and update it each second: fo-timeout

This PR doesn't allow to edit option.

Forum thread: https://freeorion.org/forum/viewtopic.php?f=9&t=11234

PR #2442, opened 2019-05-07
Show label with remain time before timeout

o01eg:
Adds clock to count down seconds before turn advance:

fo-timeout


o01eg:
I've decided to move it into #2441

PR #2454, opened 2019-05-23
Encode server name to save it in the XML element

### LINKED ISSUE #2449: Storing cookie for server by IP address violates XML standart
### Environment

* **FreeOrion Version**: 0.4.8, master
* **Operating System**: Any
* **Graphic card used**:
* **Fetched as** 
  * Binary release
  * Weekly development build
  * Compiled from source

### Description
When cookies stored for server addresses by IP for example "147.135.99.147" it becomes 

```xml
    <server>
      <cookie>
        <147>
          <135>
            <99>
              <147>Some long UUID</147>
            </99>
          </135>
        </147>
      </cookie>
      <uri>localhost</uri>
    </server>
```

This breaks options parsing and violates XML standard: https://www.w3schools.com/xml/xml_elements.asp

> Element names must start with a letter or underscore

### Expected Result
Servers addressed by IP doesn't break options.

I suppose it should be some encoding applied to server address so it will become correct xml element.

### Steps to reproduce
* Login to server by IP.
* Open options.xml

o01eg:
Fixes #2449

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -40,6 +40,7 @@
 #include <GG/Cursor.h>
 #include <GG/utf8/checked.h>
 
+#include <boost/algorithm/string/replace.hpp>
 #include <boost/algorithm/string/trim.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/filesystem/operations.hpp>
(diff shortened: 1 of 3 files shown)


client/human/HumanClientApp.cpp
```diff
@@ -200,6 +201,10 @@ void HumanClientApp::AddWindowSizeOptionsAfterMainStart(OptionsDB& db) {
     db.Add("video.windowed.top", UserStringNop("OPTIONS_DB_APP_TOP_WINDOWED"),    DEFAULT_TOP,         RangedValidator<int>(-max_height_plus_one, max_height_plus_one));
 }
 
+std::string HumanClientApp::EncodeServerAddressOption(const std::string& server) {
+    return "network.server.cookie._" + boost::replace_all_copy(server, ".", "_");
```
geoffthemedio: are / will ipv6 addresses be supported? if so, also replacing `:` with `_` would be good.

o01eg: Looks like `:` works even without replacing. Still the only mention is:

> Avoid ":". Colons are reserved for namespaces (more later).

I've replaced it too.

PR #2457, opened 2019-05-30
extend UI to notify player about diplomatic changes

TheSilentOne1:
Features:
- adds a "mail" icon to the player list window, positioned to the right of the diplomatic status icon, if a diplomatic offer has been received (that needs answering) and
- opens PlayerListWnd that flashes until the offer has been accepted or rejected
- sends message to all players when diplomatic status between any empires has changed, and
- opens ChatWnd that flashes if client player is affected by the diplomatic status change
- opens ChatWnd that flashes if client player received a message

### DIFF
--- a/UI/ChatWnd.cpp
+++ b/UI/ChatWnd.cpp
@@ -487,6 +487,18 @@ void MessageWnd::HandlePlayerChatMessage(const std::string& text,
 
     *m_display += wrapped_text + "\n";
     m_display_show_time = GG::GUI::GetGUI()->Ticks();
+
+    // if client empire is target of message, show message window
+    const ClientApp* app = ClientApp::GetApp();
+    if (!app) {
+        ErrorLogger() << "MessageWnd::HandleDiplomaticStatusChange couldn't get client app!";
+        return;
+    }
+    int client_empire_id = app->EmpireID();  
+    if (recipient_player_id == client_empire_id) {
+        Flash();
+        Show();
+    }
 }
 
 void MessageWnd::HandleTurnPhaseUpdate(Message::TurnProgressPhase phase_id, bool prefixed /*= false*/) {
@@ -548,10 +560,39 @@ void MessageWnd::HandleLogMessage(const std::string& text) {
 void MessageWnd::HandleDiplomaticStatusChange(int empire1_id, int empire2_id) {
     const ClientApp* app = ClientApp::GetApp();
     if (!app) {
-        ErrorLogger() << "PlayerListWnd::PlayerRightClicked couldn't get client app!";
+        ErrorLogger() << "MessageWnd::HandleDiplomaticStatusChange couldn't get client app!";
         return;
     }
+
     int client_empire_id = app->EmpireID();
+    DiplomaticStatus status = Empires().GetDiplomaticStatus(empire1_id, empire2_id);
+    std::string text;
+
+    const Empire* empire1 = GetEmpire(empire1_id);
+    const Empire* empire2 = GetEmpire(empire2_id);
+
+    std::string empire1_str = GG::RgbaTag(empire1->Color()) + empire1->Name() + "</rgba>";
+    std::string empire2_str = GG::RgbaTag(empire2->Color()) + empire2->Name() + "</rgba>";
+
+    switch (status) {
+    case DIPLO_WAR:
+        text = boost::str(FlexibleFormat(UserString("MESSAGES_WAR_DECLARATION"))
+                   % empire1_str % empire2_str);
+        break;
+    case DIPLO_PEACE:
+        text = boost::str(FlexibleFormat(UserString("MESSAGES_PEACE_TREATY"))
+                   % empire1_str % empire2_str);
+        break;
+    case DIPLO_ALLIED:
+        text = boost::str(FlexibleFormat(UserString("MESSAGES_ALLIANCE"))
+                   % empire1_str % empire2_str);
+        break;
+    default:
+        ErrorLogger() << "MessageWnd::HandleDiplomaticStatusChange: no valid diplomatic status found.";
+    }
+
+    *m_display += text;
+    m_display_show_time = GG::GUI::GetGUI()->Ticks();
 
     // if client empire is target of diplomatic status change, show message window
     if (empire2_id == client_empire_id) {
@@ -640,6 +681,7 @@ void MessageWnd::MessageEntered() {
     }
 
     m_edit->Clear();
+    StopFlash();
 }
 
 void MessageWnd::MessageHistoryUpRequested() {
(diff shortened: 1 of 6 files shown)


TheSilentOne1:
I'd also like to add a sitrep message, but don't know where I should place the code to generate the sitrep.

TheSilentOne1:
(Guess here: custom_sitreps.txt)

geoffthemedio:
Works for alliance and peace proposals. Green colour seems a bit odd. There is no notification for messages that don't require a response, like war declarations or ending an alliance.

youve got mail

TheSilentOne1:
> There is no notification for messages that don't require a response, like war declarations or ending an alliance.

And I guess that makes sense, with the mail icon meaning "message waiting for you, awaiting response". However, we then need a different method of communicating war declarations (and alliance breakups).
Some ideas, to be mulled over:
- use "messages" window to display war declarations etc. ("XYZ declared war on our empire")
- show some dialog window in the middle of the screen that asks for confirmation it has been read (safer way to make sure the player knows he has been declared war)
- show a different icon than the mail one, e. g. crossed out alliance or peace symbol

o01eg:
If diplomatic statuses are public either way what if generate real chat message like it done in https://github.com/freeorion/freeorion/pull/2393?

TheSilentOne1:
Thanks for the suggestion, I'll look into it.

o01eg:
You don't need to use `UserString` on the server. Send `[[STRING_ID,arg1,arg2]]` to the client.

TheSilentOne1:
Updated description in the first post. Tested and working.

server/ServerApp.cpp
```diff
@@ -70,6 +70,16 @@ namespace {
         }
         return ALL_EMPIRES;
     }
+
+    std::string ColorToRgbTag(const GG::Clr color) {
+        std::string color_str = "<rgba ";
+        color_str += std::to_string(static_cast<int>(color.r)) + " ";
+        color_str += std::to_string(static_cast<int>(color.b)) + " ";
+        color_str += std::to_string(static_cast<int>(color.g)) + " ";
+        color_str += std::to_string(static_cast<int>(color.a)) + ">";
```
TheSilentOne1: Suggestions for a better conversion method welcome.

TheSilentOne1: Never mind.

TheSilentOne1:
This should be ready for review.

server/ServerApp.cpp
```diff
@@ -3641,11 +3651,42 @@ void ServerApp::HandleDiplomaticStatusChange(int empire1_id, int empire2_id) {
...
 
+    boost::posix_time::ptime timestamp = boost::posix_time::second_clock::universal_time();
+    std::string text;
+
+    const Empire* empire1 = GetEmpire(empire1_id);
+    const Empire* empire2 = GetEmpire(empire2_id);
+
+    std::string empire1_clr_str = ColorToRgbTag(empire1->Color());
+    std::string empire2_clr_str = ColorToRgbTag(empire2->Color());
+    std::string white_clr_str = "<rgba 255 255 255 255>";
```
geoffthemedio: close the tags rather than opening more.

the default text colour is a UI setting, and cannot be assumed to be solid white anyway.

server/ServerApp.cpp
```diff
@@ -3641,11 +3651,42 @@ void ServerApp::HandleDiplomaticStatusChange(int empire1_id, int empire2_id) {
...
+    default:
+        ErrorLogger() << "ServerApp::HandleDiplomaticStatusChange: no valid diplomatic status found.";
+    }
+
     for (auto player_it = m_networking.established_begin();
          player_it != m_networking.established_end(); ++player_it)
     {
         PlayerConnectionPtr player = *player_it;
         player->SendMessage(DiplomaticStatusMessage(update));
+        player->SendMessage(ServerPlayerChatMessage(Networking::INVALID_PLAYER_ID, timestamp, text));
```
geoffthemedio: does this really need to be a chat message? I dislike this feature creep, with even more UI stuff being done on the server. This message could be generated client side in response to the / an actual network message.

TheSilentOne1: I had handled the messages within the ChatWnd code before, would that be acceptable or do you have a better idea where to move the code?

geoffthemedio: Depends what you mean. Something like how turn progress updates work would be good, I think.

TheSilentOne1:
Moved the code from ServerApp to ChatWnd, where progress updates are handled as well.

UI/ChatWnd.cpp
```diff
@@ -479,6 +487,18 @@ void MessageWnd::HandlePlayerChatMessage(const std::string& text,
 
     *m_display += wrapped_text + "\n";
     m_display_show_time = GG::GUI::GetGUI()->Ticks();
+
+    // if client empire is target of message, show message window
```
geoffthemedio, line 491: What messages are players receiving that aren't targeted to them?

geoffthemedio:
I think `StopFlash` needs to be called more often / in different places. I have to game windows open, both in a multiplayer game, and after sending a peace proposal from one to another, clicking and dragging the players list doesn't stop it flashing. After accepting, the messages window in the first player flashes, but doesn't stop when dragged / clicked on. Possibly the game window not having focus when the messages are sent is part of it.

UI/ChatWnd.cpp
TheSilentOne1: E. g. the chat window of the multiplayer lobby will start flashing once a player enters the game.

TheSilentOne1:
> ... after sending a peace proposal from one to another, clicking and dragging the players list doesn't stop it flashing.

Left-clicking now stops the flash of the PlayerListWnd; should dragging stop the window flash as well? To me dragging doesn't necessarily implicate the player wants the flashing to stop.

> Possibly the game window not having focus when the messages are sent is part of it.

Right, the first click sets focus to the game window, the second one stops the flash. This seems like it is expected behaviour. Alterntatively, flashing could stop once the game window gains focus, does that seem preferable to you?

geoffthemedio:
> Left-clicking now stops the flash of the PlayerListWnd; should dragging stop the window flash as well? To me dragging doesn't necessarily implicate the player wants the flashing to stop.

Probably yes... dragging is, to me, a more substantial action than just clicking on a window. It suggests the player's attention is on it anyway...

> Right, the first click sets focus to the game window, the second one stops the flash. This seems like it is expected behaviour.

I'll have to check again, but it felt like the flashing just never stopped if it started when the game window didn't have focus, even after subsequent clicks on the flashing window. Maybe that's already been fixed, though.

> Alterntatively, flashing could stop once the game window gains focus, does that seem preferable to you?

Game window gaining focus shouldn't stop the flashing, as then players could miss the notification without having seen the window.

geoffthemedio:
Clicking to stop flashing works now.

Really feels like dragging should work the same as clicking for stopping flashing, but it can go in later...

PR #2464, opened 2019-06-06
Don't send message after write error again

o01eg:
Due asynchronous error processing it possible to try to send message multiple times before connection will be considered broken. Each error will processed with new `Disconnection` FSM event.
This PR marks connection marks connection as invalid on write error and don't process it again.

### DIFF
--- a/network/ServerNetworking.h
+++ b/network/ServerNetworking.h
@@ -326,6 +326,7 @@ private:
     bool                            m_authenticated;
     Networking::AuthRoles           m_roles;
     boost::uuids::uuid              m_cookie;
+    bool                            m_valid;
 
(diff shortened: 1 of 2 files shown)


network/ServerNetworking.h
```diff
@@ -326,6 +326,7 @@ class PlayerConnection :
     bool                            m_authenticated;
     Networking::AuthRoles           m_roles;
     boost::uuids::uuid              m_cookie;
+    bool                            m_valid;
```
geoffthemedio, line 329: no need for threaded access guards?

o01eg: Server uses boost asio in a single threaded event loop.

PR #2465, opened 2019-06-06
Load chat history in quickstart and load mode

o01eg:


server/ServerFSM.cpp
```diff
@@ -707,9 +707,11 @@ sc::result Idle::react(const Hostless&) {
     std::shared_ptr<MultiplayerLobbyData> lobby_data(new MultiplayerLobbyData(std::move(server.m_galaxy_setup_data)));
     std::shared_ptr<ServerSaveGameData> server_save_game_data(new ServerSaveGameData());
     std::vector<PlayerSaveGameData> player_save_game_data;
+    server.InitializePython();
+    server.LoadChatHistory();
     if (autostart_load_filename.empty()) {
         DebugLogger(FSM) << "Start new game";
-        server.InitializePython();
+
```
geoffthemedio, line 714: Is this change relevant to the commit description...?

o01eg: Yes, I need to initialize python before loading chat and I don't need to initialize python twice.

PR #2466, opened 2019-06-07
add diplomatic status indicator

TheSilentOne1:
Adds an indicator of diplomatic relations to the PlayerListWindow. There are three icons representing a diplomatic status (war, peace, allied) each followed by colored squares representing empires, like below.

player_list_wnd

Unfortunately, I'm still not able to play with AI oponents with my own builds, so this is barely tested. There is one unresolved issue with the diplo icons being too large which I'm still looking into.

### DIFF
--- a/UI/PlayerListWnd.cpp
+++ b/UI/PlayerListWnd.cpp
@@ -92,6 +92,86 @@ namespace {
 
 
     ////////////////////////////////////////////////
+    // DiplomaticStatusIndicator
+    ////////////////////////////////////////////////
+    /** Shows specified diplomatic status from specified empire to each other empire. */
+    class DiplomaticStatusIndicator : public GG::Control {
+    public:
+        DiplomaticStatusIndicator(GG::X w, GG::Y h, int empire_id, DiplomaticStatus diplo_status) :
+            Control(GG::X0, GG::Y0, w, h, GG::NO_WND_FLAGS),
+            m_empire_id(empire_id),
+            m_empire_ids(),
+            m_diplo_status(diplo_status),
+            m_icon(nullptr)
+        {}
+
+        void Update() {
+            const ClientApp* app = ClientApp::GetApp();
+            if (!app) {
+                ErrorLogger() << "DiplomaticStatusIndicator::Render couldn't get client app!";
+                return;
+            }
+
+            // pick icon and tooltip text for diplo status
+            switch (m_diplo_status) {
+            case DIPLO_WAR:
+                m_icon = WarIcon();
+                break;
+            case DIPLO_PEACE:
+                m_icon = PeaceIcon();
+                break;
+            case DIPLO_ALLIED:
+                m_icon = AlliedIcon();
+                break;
+            default:
+                m_icon = UnknownIcon();
+                break;
+            }
+
+            // add color for each empire with specified diplomatic status
+            m_empire_ids.clear();
+
+            for (const auto& empire : Empires()) {
+                int empire_id = empire.second->EmpireID();
+
+                if (empire.second->Eliminated()) continue;
+                if (m_empire_id == empire_id) continue;
+
+                if (Empires().GetDiplomaticStatus(empire_id, m_empire_id) == m_diplo_status) {
+                    m_empire_ids.emplace_back(empire_id);
+                }
+            }
+        }
+
+        void Render() override {
+            int ICON_SIZE = IconSize();
+            GG::Pt ul = UpperLeft();
+            const GG::Clr& border_clr = ClientUI::WndOuterBorderColor();
+
+            // diplomatic status icon
+            glColor(GG::CLR_WHITE);
+            m_icon->OrthoBlit(ul, ul + GG::Pt(GG::X(ICON_SIZE), GG::Y(ICON_SIZE)));
+
+            // empire color squares
+            for (const auto& empire_id : m_empire_ids) {
+                const GG::Clr& square_color = GetEmpire(empire_id)->Color();
+                GG::Pt square_ul = ul + GG::Pt(GG::X(empire_id * (ICON_SIZE + PAD)), GG::Y0);
+                GG::FlatRectangle(square_ul, square_ul + GG::Pt(GG::X(ICON_SIZE), GG::Y(ICON_SIZE)), square_color, border_clr, DATA_PANEL_BORDER);
+            }
+        }
+
+    private:
+        int                                        m_empire_id;
+        std::vector<int>                           m_empire_ids;
+        DiplomaticStatus                           m_diplo_status;
+        std::shared_ptr<GG::Texture>               m_icon;
+
+        int IconSize() const { return Value(Height()); }
+        int PAD = 3;
+    };
+
+
+    ////////////////////////////////////////////////
     // PlayerDataPanel
     ////////////////////////////////////////////////
     /** Represents a player.  This class is used as the sole Control in
@@ -109,6 +189,9 @@ namespace {
             m_empire_production_text(nullptr),
             m_empire_research_text(nullptr),
             m_empire_detection_text(nullptr),
+            m_war_indicator(nullptr),
+            m_peace_indicator(nullptr),
+            m_allied_indicator(nullptr),
             m_diplo_status(INVALID_DIPLOMATIC_STATUS),
             m_player_status(Message::PLAYING_TURN),
             m_player_type(Networking::INVALID_CLIENT_TYPE),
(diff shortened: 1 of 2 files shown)


o01eg:
You didn't changed any common or server code so I think you could try to connect to the longturn server and see diplomatic status.
Also I though server doesn't send diplomatic status of the other empires.

TheSilentOne1:
> You didn't changed any common or server code so I think you could try to connect to the longturn server and see diplomatic status.

Leads to crash unfortunately.

o01eg:
> > You didn't changed any common or server code so I think you could try to connect to the longturn server and see diplomatic status.
> 
> Leads to crash unfortunately.

I think it would be very helpful to recompile it with debug info and get stacktrace or even run in debugger. Also it better to open new issue about crash.

TheSilentOne1:
Okay, I will try once I have spare time.

o01eg:
How it will look in a game with 20 empires? Do indicators show eliminated empires?

TheSilentOne1:
> How it will look in a game with 20 empires?

There will be (empire_count -1) colorsquares per line.

>Do indicators show eliminated empires?

I can't tell immediately. If diplomatic status is reset for eliminated empires, no. Anyhow, I will look into it and make sure eliminated empires will not be shown.

geoffthemedio:
I think that with so many icons of vague meaning, it will be necessary to have working tooltips in this window... Do I recall correctly that it currently doesn't have them?

TheSilentOne1:
That's something I'm planning to add for the color squares and icons (I was thinking about at least showing the name of the empire). As for the other elements of the PlayerListWindow, I can look into it sometime.

geoffthemedio:
Need to do something about the layout code. When resizing the window, it hangs the GUI for a half second or so for me with the computer set to power saving mode.

There appear to be some layout issues as well... after getting peace with a beginner AI, then resizing the empires window, some of the icons reposition oddly:

empires with diplo layhout odd

There should be tooltips for all the new icons, not just a small subset (appraently?) I'm not sure what the various icons are supposed to mean.... red with a yellow stripe at the bottom...? Box with red or cyan colour oddly to the left?

TheSilentOne1:
Slowness issue: Before PlayerDataPanel::DoLayout() is called, DiplomaticStatusIndicator::Update() needs to be called once so m_color_squares is initialised. I was calling DSI::Update() from within DSI::Width() and DSI::DoLayout() to make that sure (which speed-wise apparently is a very bad idea). I see two relatively simple solution: 1) Call DSI::Update() once in PDP::CompleteConstruction(); OR 2) Switch around PDP::Update() and PDP::DoLayout(), which seems more reasonable because DSI::Update() will only be called once; but I'm not sure switching them won't cause other problems.

geoffthemedio:
Resizing the window is much smoother now. The layout of the various icons still changes when resizing for no apparent reason. The meaning of the rightmost two columns of icons is still unclear.

TheSilentOne1:
I don't want to bother you with this more than necessary, so I will continue work here once I'm able to test the code myself (once the multiplayer server is up again).

o01eg:
@TheSilentOne1 I didn't stop multiplayer server. You could connect to it.

TheSilentOne1:
Doesn't the patch break compatibility?

o01eg:
@TheSilentOne1 There are only UI changes. They don't break compatibility.

geoffthemedio:
You don't need to connect to an internet server... just start a single player game with a few AIs and the max AI aggression at beginner. You can propose peace and they will accept, changing their diplomatic status, and then you can re-declare war.

TheSilentOne1:
> @TheSilentOne1 There are only UI changes. They don't break compatibility.

Currently, I can't log into the server without crash. So I would need to apply Geoff's or your patch, do they not break compatibility?

> You don't need to connect to an internet server... just start a single player game with a few AIs and the max AI aggression at beginner. You can propose peace and they will accept, changing their diplomatic status, and then you can re-declare war.

I wish it were that easy ;)  Unfortunately, if I compile FO myself, once I add an AI I can't enter the game. I described the issue here: https://freeorion.org/forum/viewtopic.php?f=25&t=11145
I wasn't able to resolve the issue.

geoffthemedio:
Can you run two clients locally and connect to a local server? I can do this in Windows by just running freeorion once, hosting a multiplayer game, then starting freeorion again and joining a game at localhost.

You can apply https://github.com/freeorion/freeorion/pull/2480 but it won't help if the server is still sending you messages that still have a boost::optional being serialized. Your client should try to deserialize the older format message, but if it's actually the problem, it would still fail.

TheSilentOne1:
> Can you run two clients locally and connect to a local server? I can do this in Windows by just running freeorion once, hosting a multiplayer game, then starting freeorion again and joining a game at localhost.

I'll try that.

TheSilentOne1:
Thanks for that very useful hint. I've fixed the layout issue. This is how it looks now:

psi

geoffthemedio:
Have the positions of the coloured squares for each empire remain the same, so all the same-colour squares are in a column, not shifting left / right if something is removed or added to their left. Eg. the yellow squares in this:
war_peace_icons

TheSilentOne1:
Color squares now sorted by empire id. Squashed commits.

geoffthemedio:
I thought partial tooltips were working earlier, but now I get none...

Otherwise looks better.

TheSilentOne1:
The right-click popupmenu, which allows to make peace etc., is handled by PlayerListWnd. To pass the right-clicks to PlayerListWnd, I've set the diplomatic status indicators to GG::NO_WND_FLAGS. So tooltips would need to be handled in PlayerListWnd, which seems complicated and I'm not sure is worth the trouble.

geoffthemedio:
It's not especially difficult / complicated. You can have a parent `Wnd` add an `EventFilter` to a child `Wnd` which handles things like click events, while the child `Wnd` is otherwise still interactive and can have tooltips. See https://github.com/freeorion/freeorion/commit/bb98a21514c449527d8819744918084465a35e18

TheSilentOne1:
Thanks. I'll add tooltips to the dsi's in a couple of days.

TheSilentOne1:
Haven't got this working yet, DSI::Eventfilter() doesn't seem to be called despite
`m_war_indicator->InstallEventFilter(shared_from_this());
            m_peace_indicator->InstallEventFilter(shared_from_this());
            m_allied_indicator->InstallEventFilter(shared_from_this());`

geoffthemedio:
-`DiplomaticStatusIndicator` is constructed with `GG::NO_WND_FLAGS`, so will not be interactive, so won't receive any events, but rather then mouseover event will be passed to parent directly.
-`PlayerDataPanel` is also similarly constructed to be non-interactive, so won't get mouseover events if moused over elsewhere (than the diplo status indicator).
-`PlayerDataPanel` doesn't define `EventFilter`, so even if it intercepted an event from one of the diplo status indicators, it wouldn't do anything with it.
-The point of installing an event filter is to capture events from another `Wnd` that would normally handle them, particularly in cases where one doesn't want to modify the other `Wnd`'s class or made a specialized derived class, but in this case, the `DiplomaticStatusIndicator` is already a specialized class, which could just be modified to forward a bunch of events to its parent, even if it's interactive, so that the mouseover and other interactive behaviour of the parent is in general maintained, except when the specialized class does something special.
-What you want to do is add a browse wnd to `DiplomaticStatusIndicator`, probably depending on which `DiplomaticStatus` is passed to it when constructed.

UI/PlayerListWnd.cpp
```diff
@@ -21,6 +21,16 @@
 namespace {
     const int           DATA_PANEL_BORDER = 1;
 
+    std::string ColorToRgbTag(const GG::Clr color) {
```
TheSilentOne1: Rather move to Empire::RgbaColorTag()?

TheSilentOne1:
> What you want to do is add a browse wnd to DiplomaticStatusIndicator, probably depending on which DiplomaticStatus is passed to it when constructed.

That is the easy part. What I don't know is how to pass on the RightClick event to PlayerListWnd. If I introduce
`void RClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_keys){ForwardEventToParent();}`, I can only pass the event until `PlayerDataPanel`, but that will not pass it on to `PlayerRow` (from where I could emit the RowRightClickedSignal).

UI/PlayerListWnd.cpp
geoffthemedio: Colour tags are not empire-related, so no.

geoffthemedio:
Does `RClick` need to go all the way back to the list window? Regardless, why can't you make `PlayerDataPanel` pass it further or emit a signal?

TheSilentOne1:
> Does `RClick` need to go all the way back to the list window?

At least that's where the right-clicks are currently handled: PlayerListWnd::PlayerRightClicked().

> Regardless, why can't you make PlayerDataPanel pass it further or emit a signal?

For a reason I'm not aware of, RClick's aren't passend from PlayerDataPanel to PlayerRow. I tried to fix it with `AttachChild(m_panel)` in `PlayerRow::CompleteConstruction()`, but that didn't help.

UI/PlayerListWnd.cpp
TheSilentOne1: Can you think of some other suitable place for this code so it doesn't need to be reproduced in any .cpp file it is needed?

geoffthemedio: Just use `GG::RgbaTag`

geoffthemedio:
I think the issue is that you also need to forward `RButtonDown` events, not just `RClick` events. The right-click mechanism needs both to function.

With those event handlers,
````
        void RButtonDown(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_keys) override
        { ForwardEventToParent(); }

        void RClick(const GG::Pt& pt, GG::Flags<GG::ModKey> mod_keys) override
        { ForwardEventToParent(); }
````

in `DiplomaticStatusIndicator` the right-click and tooltip work for me. Both should be `public`, also. `PlayerRow` and `PlayerDataPanel` don't appear to need them.

TheSilentOne1:
Thanks for your help.

PR #2469, opened 2019-06-08
Try to find suffixed boost_python

o01eg:
Some Linux distributives switched to suffixed boost python for versions <= 1.66.
Use major and minor python version from used python.

Cherry-picked #2462


geoffthemedio:
Can the CI build failure be ignored? Not sure if there's any difference in what the testing is doing for the release branch vs. master...

o01eg:
Yes, it's timeout issue because ccache isn't populated  with 0.4.8 branch.

Vezzra:
As there won't be a maintenance release for 0.4.8, I've set the milestone to "Gateway to the Void" - "Next Release" doesn't really fit in this case ("0.4.8.1" would, but alas...).

PR #2471, opened 2019-06-10
Fix cmake compatibility

o01eg:
GREATER_EQUAL was introduced in 3.7 https://stackoverflow.com/questions/16667017/cmake-express-the-greater-or-equal-statement
Freeorion still supports 3.1


geoffthemedio:
CMake AppVeyor build fail can be ignored?

o01eg:
Yes, it's a timeout issue.

PR #2474, opened 2019-06-11
Add more diagnostic output

o01eg:


network/Message.cpp
```diff
@@ -873,9 +873,11 @@ void ExtractGameStartMessageData(const Message& msg, bool& single_player_game, i
                 ia >> BOOST_SERIALIZATION_NVP(ui_data_available);
                 if (ui_data_available)
                     ia >> BOOST_SERIALIZATION_NVP(ui_data);
+                TraceLogger() << "ExtractGameStartMessage UI data " << ui_data_available << " deserialization time " << (deserialize_timer.elapsed() * 1000.0);
```
geoffthemedio, line 876: OK I guess but when do you expect deserializing a single bool will take any substantial amount of time?

o01eg: If this bool is true, then there will be deserialization of more data. Also this log also shows steps of game start message deserialization in the case of failure on some step.

PR #2478, opened 2019-06-13
Add more logs about restoring UI data and ship designs

o01eg:
Adds some additional logs following `TheSilentOne` crash.


o01eg:
Caught error with it.

PR #2481, opened 2019-06-15
Add option to set game UID

o01eg:
Useful for standalone server with the only game.
Allows game UID to be accessible from python `AuthProvider.list_players` so different server with different game could share same database.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1768,6 +1768,9 @@ OPTIONS_DB_GAMESETUP_SEED
 '''The seed used for randomly generating the galaxy.
 Galaxies generated with the same settings and the same seed will be identical.'''
 
+OPTIONS_DB_GAMESETUP_UID
+'''The game UID for new games. Server-only option can be used for predictable game name on hostless servers.'''
+
 OPTIONS_DB_GAMESETUP_STARS
 '''The approximate number of systems in the galaxy to be generated.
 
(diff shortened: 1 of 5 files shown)


geoffthemedio:
Could you elaborate on the description? I'm not clear how this is useful...

o01eg:
I mean if there single user database but different servers start different games with players from the database, it is simpler to use game UID to choose players participate each game. So I've introduced option to set game UID with configuration only.

geoffthemedio:
Still not clear. What does the new `setup.game.uid` option have to do with picking players? Why is it defined in common and not just server side? What does this have to do with Python?  How is a game UID related to what players to put into a lobby? Why does the UID need to be manually changed or be "predictable" for this to work?

o01eg:
> What does the new setup.game.uid option have to do with picking players?

`AuthProvider.list_players` can get this value from `fo.get_galaxy_setup_data().gameUID` and use here.

> Why is it defined in common and not just server side?

I've placed it where other galaxy setup options are placed.

> What does this have to do with Python?

Before this PR game UID initializes before universe generation. The PR made it available from the server start.

> How is a game UID related to what players to put into a lobby?

If hostless server starts new game with `quickstart` option without lobby it calls `AuthProvider.list_players` to get players for the new game. It can be rewrote to use DBMS.
And if there different servers with different games use same database, they need a way to select only players participate in each game.

> Why does the UID need to be manually changed or be "predictable" for this to work?

If we don't know which game UID will be assigned, we can't provide enough data to database query to select players for the given game.

o01eg:
Rebased. Fixed conflicts.

o01eg:
Actually I already use predefined game UID to place saved games into known folder which is used to automatically load latest, but I set this value in the universe generator. Next I found I need this value earlier.

geoffthemedio:
Still not clear why this is an option. Wouldn't a server-only command line parameter make more sense? ie. call `freeoriond --game-uid 83838383potato` or similar, used only when quickstarting / hostless as applicable?

o01eg:
It is part of `GalaxySetupData` so I decided place it into `setup.*` options. I could make it server-only although.

geoffthemedio:
Why is it part of `GalaxySetupData` though? It's not really about the galaxy; it's other info about the game, like player names or rules...

Being server-only would make more sense, but is it necessary / appropriate as an option?

o01eg:
Option can be added to `persistent_config.xml`.

o01eg:
Rebased. Made it server-only.

PR #2479, opened 2019-06-14
Don't break game if UI data corrupted

o01eg:
Fixes #2475 
Breaks compatibility due reordering game start message fields.
Shows non-fatal error to user if UI data corrupted instead of crash.

PR #2486, opened 2019-06-20
Avoid serializing of std::unordered_map in SaveGameUIData

o01eg:
Yet another alternative to #2479

### DIFF
--- a/util/SerializeMultiplayerCommon.cpp
+++ b/util/SerializeMultiplayerCommon.cpp
@@ -120,7 +120,28 @@ void SaveGameUIData::serialize(Archive& ar, const unsigned int version)
     }
     ar  & BOOST_SERIALIZATION_NVP(ordered_ship_hull_and_obsolete);
     TraceLogger() << "SaveGameUIData::serialize ship hull processed";
-    ar  & BOOST_SERIALIZATION_NVP(obsolete_ship_parts);
+    if (Archive::is_saving::value || version >= 4) {
+        std::map<std::string, int> ordered_obsolete_ship_parts;
+
+        if (Archive::is_saving::value) {
+            // populate temp container
+            ordered_obsolete_ship_parts = std::move(std::map<std::string, int>(obsolete_ship_parts.begin(), obsolete_ship_parts.end()));
+            // serialize into archive
+            ar & BOOST_SERIALIZATION_NVP(ordered_obsolete_ship_parts);
+        } else {
+            // deserialize into temp container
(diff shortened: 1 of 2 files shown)


util/SerializeMultiplayerCommon.cpp
```diff
@@ -120,7 +120,28 @@ void SaveGameUIData::serialize(Archive& ar, const unsigned int version)
     }
     ar  & BOOST_SERIALIZATION_NVP(ordered_ship_hull_and_obsolete);
     TraceLogger() << "SaveGameUIData::serialize ship hull processed";
-    ar  & BOOST_SERIALIZATION_NVP(obsolete_ship_parts);
+    if (Archive::is_saving::value || version >= 4) {
+        std::map<std::string, int> ordered_obsolete_ship_parts;
+
+        if (Archive::is_saving::value) {
+            // populate temp container
+            ordered_obsolete_ship_parts = std::move(std::map<std::string, int>(obsolete_ship_parts.begin(), obsolete_ship_parts.end()));
```
geoffthemedio, line 128: I think you could declare `ordered_obsolete_ship_parts` here, and simplify this (visually) with
`std::map<std::string, int> oosp{obsolete_ship_parts.begin(), obsolete_ship_parts.end()};`

util/SerializeMultiplayerCommon.cpp
```diff
@@ -120,7 +120,28 @@ void SaveGameUIData::serialize(Archive& ar, const unsigned int version)
...
+            // serialize into archive
+            ar & BOOST_SERIALIZATION_NVP(ordered_obsolete_ship_parts);
+        } else {
+            // deserialize into temp container
```
geoffthemedio, line 132: would also need to declare `ordered_obsolete_ship_parts` here if doing the above.

util/SerializeMultiplayerCommon.cpp
```diff
@@ -120,7 +120,28 @@ void SaveGameUIData::serialize(Archive& ar, const unsigned int version)
...
+            ar & BOOST_SERIALIZATION_NVP(ordered_obsolete_ship_parts);
+
+            // extract from temp container
+            obsolete_ship_parts = std::move(std::unordered_map<std::string, int>(ordered_obsolete_ship_parts.begin(), ordered_obsolete_ship_parts.end()));
```
geoffthemedio, line 136: but I suppose there's no suitable `operator=` for `unordered_map` that takes an iterator range or `map` for here... so I guess it's fine...

PR #2490, opened 2019-06-23
Replace bits8 and bits16 removed in Boost 1.70

o01eg:
I've tried to fix GG for boost 1.70.
Could someone test it against 1.58 boost?


Oberlus:
> Could someone test it against 1.58 boost?

Compiles fine with latest master.

Tested single player, issued orders and advanced two turns, no problems.

Tested against multiplayer slow-game server, connects fine but pressing Turn gives you
"Server cannot properly read messages from your client, most likely due to incompatible game versions."

o01eg:
> Tested single player, issued orders and advanced two turns, no problems.

Good.

> Tested against multiplayer slow-game server, connects fine but pressing Turn gives you
"Server cannot properly read messages from your client, most likely due to incompatible game versions."

It is expected. I wait for next weekly test build to update server.

PR #2491, opened 2019-06-24
Replace bits8 and bits16 removed in Boost 1.70

o01eg:
Cherry-picked #2490


agrrr3:
Was it intended that this went into the old (0.4.8) release?

o01eg:
Only if there will be 0.4.8.1 release or for distributive maintainers.

Vezzra:
0.4.8.1 won't happen, we are already trying to get ready for 0.4.9.

PR #2492, opened 2019-06-24
Hull size dependent fuel

agrrr3:
This implements fuel efficiency based on hull "size".  [Forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11245&p=96040#p96037)

It distinguishes between Great 400% efficiency, Good 200% efficiency, Average 100% efficiency, and BAD 60% efficiency. These efficiency levels mostly correspond to small, medium, large and huge hulls.

As also the base fuel of all hulls was halved, small hulls effectively double jump distance, medium hulls have the old fuel behaviour, large hulls can get half the distance.

This means very small ships (e.g. scouts, flux disruptors) can go very far, huge hulls are mostly dependent on supply networks/need very good fuel tech.

Implementation is done by a late effect which applies the appropiate scaling multiplier.
A hull needs a tag (so AI can do correct ship design) and the corresponding macro effect.

If a ship is missing the correct tag the effect is not applied so it is easy to catch the bug. 

```
$ for f in GREAT GOOD AVERAGE BAD; do echo $f ; grep --exclude ship_hulls.macros -e "tags.*${f}_FUEL" -r default/scripting/ship_hulls/ -l  ; done
GREAT
default/scripting/ship_hulls/robotic/SH_SPACE_FLUX_BUBBLE.focs.txt
default/scripting/ship_hulls/robotic/SH_SPATIAL_FLUX.focs.txt
default/scripting/ship_hulls/SH_BASIC_SMALL.focs.txt
GOOD
default/scripting/ship_hulls/asteroid/SH_MINIASTEROID_SWARM.focs.txt
default/scripting/ship_hulls/asteroid/SH_SMALL_ASTEROID.focs.txt
default/scripting/ship_hulls/asteroid/SH_SMALL_CAMOUFLAGE_ASTEROID.focs.txt
default/scripting/ship_hulls/energy/SH_COMPRESSED_ENERGY.focs.txt
default/scripting/ship_hulls/energy/SH_ENERGY_FRIGATE.focs.txt
default/scripting/ship_hulls/organic/SH_ENDOMORPHIC.focs.txt
default/scripting/ship_hulls/organic/SH_ENDOSYMBIOTIC.focs.txt
default/scripting/ship_hulls/organic/SH_ORGANIC.focs.txt
default/scripting/ship_hulls/organic/SH_PROTOPLASMIC.focs.txt
default/scripting/ship_hulls/organic/SH_STATIC_MULTICELLULAR.focs.txt
default/scripting/ship_hulls/organic/SH_SYMBIOTIC.focs.txt
default/scripting/ship_hulls/robotic/SH_TRANSSPATIAL.focs.txt
default/scripting/ship_hulls/SH_BASIC_MEDIUM.focs.txt
AVERAGE
default/scripting/ship_hulls/asteroid/SH_ASTEROID.focs.txt
default/scripting/ship_hulls/asteroid/SH_CAMOUFLAGE_ASTEROID.focs.txt
default/scripting/ship_hulls/asteroid/SH_CRYSTALLIZED_ASTEROID.focs.txt
default/scripting/ship_hulls/energy/SH_FRACTAL_ENERGY.focs.txt
default/scripting/ship_hulls/energy/SH_QUANTUM_ENERGY.focs.txt
default/scripting/ship_hulls/organic/SH_BIOADAPTIVE.focs.txt
default/scripting/ship_hulls/organic/SH_RAVENOUS.focs.txt
default/scripting/ship_hulls/organic/SH_SENTIENT.focs.txt
default/scripting/ship_hulls/robotic/SH_NANOROBOTIC.focs.txt
default/scripting/ship_hulls/robotic/SH_ROBOTIC.focs.txt
default/scripting/ship_hulls/SH_BASIC_LARGE.focs.txt
BAD
default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
default/scripting/ship_hulls/asteroid/SH_SCATTERED_ASTEROID.focs.txt
default/scripting/ship_hulls/energy/SH_SOLAR.focs.txt
default/scripting/ship_hulls/robotic/SH_LOGITICS_FACILITATOR.focs.txt
default/scripting/ship_hulls/robotic/SH_SELF_GRAVITATING.focs.txt
default/scripting/ship_hulls/robotic/SH_TITANIC.focs.txt
```

Originally 25% efficiency was intended for BAD. Instead 30% efficiency was chosen because else we would have two minor digits in fuel values (e.g. now 1.3 instead of 1.25) 
Bubble flux hull has GREAT efficiency but base fuel is nerfed.
Check base fuel/effect display (looks good now)
Update stringtables/encyclopedia about fuel efficiency & effects (done)


TheSilentOne1:
Nice work! But: how will the UI reflect this?

geoffthemedio:
Rename the "hull sizes" to something fuel-specific. These concepts shouldn't be merged by definition, at least in the script. There will / should be cases where "size" isn't correlated to "fuel efficiency".

TheSilentOne1:
To be more precise: with hulls having different fuel and fuel consumption rates, fleets composed of different hulls will be range limited by the ship with the lowest fuel / fuel consumption rate. Optimally, there would be an indicator that shows how many jumps a fleet can travel in its current composition.
For ships there should be the same, or alternatively an indicator of fuel consumption per jump.

agrrr3:
> To be more precise: with hulls having different fuel and fuel consumption rates, fleets composed of different hulls will be range limited by the ship with the lowest fuel / fuel consumption rate. Optimally, there would be an indicator that shows how many jumps a fleet can travel in its current composition.
> For ships there should be the same, or alternatively an indicator of fuel consumption per jump.

No, different fuel consumption/fleet composition is not an issue. Ships do not have different consumption rates. Just the max fuel gets adjusted. So a ship with 4 fuel can do 4 jumps (+1 the infamous jump-out-of-supply). But e.g. a huge ship with an antimatter tank has maybe 2 fuel, while a small ship with an antimatter tank may have 12 fuel.

If we wanted we could also adjust refill rates but actually i think that is not really important.

TheSilentOne1:
Ah, understood. Thanks for clearing that up.

default/scripting/ship_hulls/ship_hulls.macros
```diff
@@ -171,4 +171,32 @@ AVERAGE_BASE_FUEL_REGEN
...
+
+GOOD_FUEL_HULL
+'''
+'''
+
+AVERAGE_FUEL_HULL
+'''[[HULL_FUEL_EFFECTSGROUP(LARGE,Value * 0.5)]]
+'''
+
+VERY_AVERAGE_FUEL_HULL
```
geoffthemedio: very average?

default/scripting/ship_hulls/ship_hulls.macros
```diff
@@ -171,4 +171,32 @@ AVERAGE_BASE_FUEL_REGEN
...
+GREAT_FUEL_HULL
+'''[[HULL_FUEL_EFFECTSGROUP(SMALL,Value * 2)]]
+'''
+
+GOOD_FUEL_HULL
+'''
+'''
+
+AVERAGE_FUEL_HULL
+'''[[HULL_FUEL_EFFECTSGROUP(LARGE,Value * 0.5)]]
```
geoffthemedio: Are the `LARGE` labels still used somewhere?

agrrr3:
Mostly happy with the PR now. 

Should I switch *_FUEL_HULL to *_HULL_FUEL ?

default/python/AI/AIDependencies.py
```diff
@@ -782,6 +784,22 @@
     "GRO_ENERGY_META": {FUEL: 2},
 }
 
+DEFAULT_FUEL_EFFICIENCY = 1
+HULL_TAG_EFFECTS = {
```
geoffthemedio, line 788: Presumably these tags could also be added to a part, so they don't really need to have "hull" in the names, do they?

geoffthemedio:
Maybe don't mention `HULL` at all, as the tags could presumably also be added to a part and function...

default/python/AI/ShipDesignAI.py
```diff
@@ -854,6 +861,15 @@ def update_stats(self, ignore_species=False):
         self.production_cost = local_cost_cache.get(self.hull.name, self.hull.productionCost(fo.empireID(), self.pid))
         self.production_time = local_time_cache.get(self.hull.name, self.hull.productionTime(fo.empireID(), self.pid))
 
+        # calculate hull stat effects
+        if self.species and not ignore_species:
```
Morlic-fo: the ``` if self.species``` check should probably be done in the function itself.

default/python/AI/ShipDesignAI.py
```diff
@@ -805,6 +805,13 @@ def _set_stats_to_default(self):
         self.production_cost = 9999
         self.production_time = 1
 
+    def _hull_efficiency(self):
+        for tag in AIDependencies.HULL_TAG_EFFECTS:
+            if self.hull.hasTag(tag):
+                if AIDependencies.FUEL_EFFICIENCY in AIDependencies.HULL_TAG_EFFECTS[tag]:
+                    return AIDependencies.HULL_TAG_EFFECTS[tag][AIDependencies.FUEL_EFFICIENCY]
+        return AIDependencies.DEFAULT_FUEL_EFFICIENCY
+
```
Morlic-fo, line 814: ```_hull_fuel_efficiency()``` ?

default/scripting/ship_hulls/ship_hulls.macros
```diff
@@ -171,4 +171,32 @@ AVERAGE_BASE_FUEL_REGEN
...
+GREAT_FUEL_HULL
+'''[[HULL_FUEL_EFFECTSGROUP(GREAT,Value * 2)]]
+'''
+
+GOOD_FUEL_HULL
+'''
+'''
+
+AVERAGE_FUEL_HULL
+'''[[HULL_FUEL_EFFECTSGROUP(AVERAGE,Value * 0.5)]]
```
Morlic-fo: The "average" efficiency should probably correspond to a factor of 1.

default/python/AI/AIDependencies.py
```diff
@@ -782,6 +784,22 @@
...
+        FUEL_EFFICIENCY: 2,
+    },
+    "GOOD_FUEL_HULL": {
+        FUEL_EFFICIENCY: 1,
+    },
+    "AVERAGE_FUEL_HULL": {
+        FUEL_EFFICIENCY: 0.5,
+    },
+    "BAD_FUEL_HULL": {
+        FUEL_EFFICIENCY: 0.25,
```
Morlic-fo: Value does not correspond to FOCS data

default/python/AI/ShipDesignAI.py
```diff
@@ -854,6 +861,15 @@ def update_stats(self, ignore_species=False):
...
         self.production_time = local_time_cache.get(self.hull.name, self.hull.productionTime(fo.empireID(), self.pid))
 
+        # calculate hull stat effects
+        if self.species and not ignore_species:
+            self.design_stats.fuel += _get_species_fuel_bonus(self.species)
+        # TODO tech hull fuel effects
+        if self.design_stats.fuel < 0:
+            self.design_stats.fuel = 0
+        else:
+            self.design_stats.fuel *= self._hull_efficiency()
```
Morlic-fo: Seems more maintainable / in alignment with FOCS script to multiply the hull efficiency at a single point at the end of this function rather than applying it to the hull and to the parts separately unless I missed something.

default/python/AI/ShipDesignAI.py
```diff
@@ -854,6 +861,15 @@ def update_stats(self, ignore_species=False):
         self.production_cost = local_cost_cache.get(self.hull.name, self.hull.productionCost(fo.empireID(), self.pid))
         self.production_time = local_time_cache.get(self.hull.name, self.hull.productionTime(fo.empireID(), self.pid))
 
+        # calculate hull stat effects
+        if self.species and not ignore_species:
+            self.design_stats.fuel += _get_species_fuel_bonus(self.species)
+        # TODO tech hull fuel effects
+        if self.design_stats.fuel < 0:
+            self.design_stats.fuel = 0
```
Morlic-fo: Is max fuel meter enforced to be >= 0 at all times by the backend?
If not, this will easily lead to wrong calculations if there are negative interim results and fuel parts are used in the design.

Morlic-fo (changes requested):

default/python/AI/AIDependencies.py
agrrr3: Mixing here all kinds of tags - I would call that premature optimisation.  Probably the effects for tags on parts would be very different from hull tags. E.g. one could have multiple parts with the same or differing tags. E.g. having GOOD_FUEL_HULL and BAD_FUEL_HULL in the same ship does not make sense.

So i suggest to keep the HULL_TAG_EFFECTS until the need for something more general arises.

default/scripting/ship_hulls/ship_hulls.macros
agrrr3: The problem here is that I would need to introduce a VERY_BAD level for the 30% and there is actually never a need for something better than GREAT.

Hmm... i could multiply all base fuel capacity by 0.5 and double the effects. Would that be a good idea?

default/python/AI/ShipDesignAI.py
agrrr3: You missed that I intended another effect implementation which would not have applied to the hull base fuel capacity. But that implementation was not really feasable and i am happy with how the current one turned out.

So, yes, so be it

default/python/AI/ShipDesignAI.py
agrrr3: Backend is not involved here. This only happens for NO_FUEL effect which is the effect happening directly before the fuel efficiency. So if you have NO_FUEL, your fuel will be 0.
I add a source comment.

default/scripting/ship_hulls/ship_hulls.macros
agrrr3: As large hulls have average fuel efficiency, rescaling everything so that average fuel is 100% means also that  for medium size hulls 0.5 fuel capacity means one jump.

agrrr3:
Note I implemented rescaling average fuel efficiency to 100% (by halving fuel from hulls, parts and effects and doubling the efficiency effect) in e519f967b740d1b36e62594950b4677a8f4312e8

Not sure that it is worth this though.

default/stringtables/en.txt
```diff
@@ -15128,9 +15132,37 @@ Cloak Cross-Interference
...
+
+FUEL_EFFICIENCY_TITLE
+Fuel Efficiency
+
+FUEL_REFUEL_EFFICIENCY_TEXT
+'''[[metertype METER_FUEL]] allows ships to travel the starlanes outside of [[metertype METER_SUPPLY]] of your empire. Fuel is consumed for each starlane traversed outside of supply; stopping within [[metertype METER_SUPPLY]] lines will fully refuel a ship. If a ship remains stationary beyond friendly [[metertype METER_SUPPLY]] lines, fuel will slowly regenerate.
+
+The base fuel capacity of a <encyclopedia ENC_SHIP_DESIGN>Ship Design</encyclopedia> depends on the [[encyclopedia ENC_SHIP_HULL]] used, but it can also be increased using some [[encyclopedia ENC_SHIP_PART]]s.
+
+Ship hulls have a great (400%), good (200%), normal (100%), or bad (60%) efficiency which determines how effective the base fuel capacity can be used. 60% efficiency means for five base fuel you get three jumps. 200% efficiency means you can do two jumps per base fuel capacity. Fuel efficiency typically decreases with hull mass, so only very small ships have a high fuel efficiency and huge ships struggle outside of [[metertype METER_SUPPLY]]. The game always show the effective fuel levels.'''
```
geoffthemedio: always show the effective fuel levels -> will always show the effective fuel levels, after scaling for hull fuel efficiency

default/stringtables/en.txt
```diff
@@ -10874,7 +10878,7 @@ GRO_ENERGY_META
 Pure-Energy Metabolism
 
 GRO_ENERGY_META_DESC
-'''Increases max [[metertype METER_FUEL]] on all ships by 2, max [[metertype METER_DEFENSE]] on all planets by 5, max [[metertype METER_SHIELD]] on all planets by 5, [[metertype METER_TARGET_INDUSTRY]] on all planets with the Industry focus by 0.2 per [[metertype METER_POPULATION]], and [[metertype METER_TARGET_RESEARCH]] on all planets with the Research focus by 0.5 per [[metertype METER_POPULATION]].
+'''Increases max [[metertype METER_FUEL]] on all ships by 1, max [[metertype METER_DEFENSE]] on all planets by 5, max [[metertype METER_SHIELD]] on all planets by 5, [[metertype METER_TARGET_INDUSTRY]] on all planets with the Industry focus by 0.2 per [[metertype METER_POPULATION]], and [[metertype METER_TARGET_RESEARCH]] on all planets with the Research focus by 0.5 per [[metertype METER_POPULATION]].
```
geoffthemedio: ... by 1 (before scaling by [[encyclopedia FUEL_EFFICIENCY_TITLE]]), max ...

default/stringtables/en.txt
```diff
@@ -5588,7 +5588,11 @@ METER_FUEL_VALUE_LABEL
...
+'''Fuel allows ships to travel the starlanes outside of [[metertype METER_SUPPLY]] of your empire. Fuel is consumed for each starlane traversed outside of supply; stopping within [[metertype METER_SUPPLY]] lines will fully refuel a ship. If a ship remains stationary beyond friendly [[metertype METER_SUPPLY]] lines, fuel will slowly regenerate.
+
+The effective amount of fuel a <encyclopedia ENC_SHIP_DESIGN>Ship Design</encyclopedia> has depends mainly on the [[encyclopedia FUEL_EFFICIENCY_TITLE]] of the [[encyclopedia ENC_SHIP_HULL]]. Fuel efficiency modifies the base fuel capacity of a <encyclopedia ENC_SHIP_DESIGN>Ship Design</encyclopedia> which depends on the [[encyclopedia ENC_SHIP_HULL]] used, but it can also be increased using some [[encyclopedia ENC_SHIP_PART]]s.
```
geoffthemedio: ... has depends mainly on the ... -> ... has is scaled by the ...

default/scripting/ship_hulls/ship_hulls.macros
agrrr3: Implemented the rescaling.

agrrr3:
@Morlic-fo does this look good to you?

agrrr3:
If morlic is MIA how do we proceed?

Vezzra:
I'm getting worried - @Dilvish-fo is MIA, now it seems @Morlic-fo too. Is anyone of the AI department left? Getting out 0.4.9 will become even more challenging this way...

agrrr3:
@geoffthemedio @Morlic-fo Had another round of review right now and i think the solutions are good/in line with the requested changes. I found a small bad value and fixed tech maxfuel effect calculation. Morlic did not answer for a month or so so I doubt I get feedback soon.

There is a small change/underspecification as the energy metabolism tech is affecting monsters, monsters have no designated fuel efficiency and the effect was scaled down monsters only get one extra max fuel compared to the previous two. I probably could go and set fuel efficiency for all monster hulls(?).

agrrr3:
Looks the unit test do not fit the code anymore. Rebasing

geoffthemedio:
Do you want to merge now, or fix the issue with monster fuel efficiency?

agrrr3:
Rather merge now. I am not convinced tame monsters should get an energy
metabolism bonus at all.

Geoff <notifications@github.com> schrieb am Fr., 26. Juli 2019, 00:49:

PR #2493, opened 2019-06-26
add CUIWnd::Flash

TheSilentOne1:
If Flash() is called, the window will periodically brighten up. The duration of the period is specified by SetFlashDuration(). The purpose of this is to have a method to draw the player's attention to a window, e. g. to the message window if a message is received.
In preparation of diplomacy UI changes.

### DIFF
--- a/UI/CUIWnd.cpp
+++ b/UI/CUIWnd.cpp
@@ -318,14 +318,13 @@ void CUIWnd::Render() {
         glDrawArrays(GL_LINE_LOOP,      m_buffer_indices[0].first, m_buffer_indices[0].second);
 
     } else {
-        if (m_flashing && GG::GUI::GetGUI()->Ticks() % (m_flash_duration * 2) > m_flash_duration)
-            glColor(GG::LightColor(ClientUI::WndColor()));
-        else
-            glColor(ClientUI::WndColor());
+        bool flashing = m_flashing && GG::GUI::GetGUI()->Ticks() % (m_flash_duration * 2) > m_flash_duration;
+
+        flashing ? glColor(GG::LightColor(ClientUI::WndColor())) : glColor(ClientUI::WndColor());
         glDrawArrays(GL_TRIANGLE_FAN,   m_buffer_indices[1].first, m_buffer_indices[1].second);
-        glColor(ClientUI::WndOuterBorderColor());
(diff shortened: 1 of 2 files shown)


UI/CUIWnd.cpp
```diff
@@ -318,7 +318,10 @@ void CUIWnd::Render() {
         glDrawArrays(GL_LINE_LOOP,      m_buffer_indices[0].first, m_buffer_indices[0].second);
 
     } else {
-        glColor(ClientUI::WndColor());
+        if (m_flashing && GG::GUI::GetGUI()->Ticks() % (m_flash_duration * 2) > m_flash_duration)
+            glColor(GG::LightColor(ClientUI::WndColor()));
```
geoffthemedio: Won't work well if the WndColor is already fairly light...

UI/CUIWnd.cpp
```diff
@@ -318,7 +318,10 @@ void CUIWnd::Render() {
...
+        else
+            glColor(ClientUI::WndColor());
         glDrawArrays(GL_TRIANGLE_FAN,   m_buffer_indices[1].first, m_buffer_indices[1].second);
         glColor(ClientUI::WndOuterBorderColor());
```
geoffthemedio: Should probably also flash the border oulines...

geoffthemedio: Should probably also flash the border outlines

UI/CUIWnd.cpp
TheSilentOne1: Did some test with bright window background and I don't really think it is an issue. Did you have something specific in mind?

UI/CUIWnd.cpp
TheSilentOne1: Sounds good.

UI/CUIWnd.cpp
geoffthemedio: Idea would be to make the colour darker instead of lighter if it's already more than ~70% white.

geoffthemedio:
Best I can tell, the CI errors can be ignored...

UI/CUIWnd.cpp
TheSilentOne1: Sounds okay to me, currently I'm hindered, I can add it later.

Issue #2488, opened 2019-06-23
Planet invasion only possible on odd turns if planet does not have defence regeneration


Oberlus:
### Environment

* **FreeOrion Version**: 0.4.8 and later.
* **Operating System**: all
* **Fetched as**: all

### Description
After the first combat in a system you plan to invade (if you got enough fire power), the planetary defences and shields are knocked down and you can invade.
If the planet has defence regeneration, on the following turns there will be some planetary defence and another combat will trigger, the shields will keep at zero and so the invasion will be possible anytime.
However, if the planet does not have defence regeneration, on the following turn there will be no planetary defences, combat will not trigger, and the planet will regenerate 1 point of shield, stopping your planned invasion on the following turn. Then, after that turn without combat, the planet will have some defence, trigger combat and enable invasion for the third turn. So you can invade this kind of planets on the first, third, fifth, etc. turns, but not on the second, fourth, etc. turns,

Forum thread: https://www.freeorion.org/forum/viewtopic.php?f=25&t=11297

PR that brought in this undesired behaviour (meter growth migration from C++ to FOCS): https://github.com/freeorion/freeorion/pull/1941

I remember there was discussion about this here in github or somewhere in the forum, but I can't find it.

### Expected Result
On all following turns after the first combat and as long as there are armed ships ready to attack, all enemy detected planets keep shields at zero, regardless of the defence techs researched by its empire, and so can be invaded on every turn after the first combat.

### Steps to reproduce
There is a game uploaded to the forum thread linked above.
To reproduce this on a different game, it is easier to find when invading native planets, that will not have defence regeneration regardless of the game turn.

Edit: issue reappeared or was never truly fixed. Heraclion provided a new [save game here](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11600&p=100534#p100534).

agrrr3:
Hi all, do we have a tag for issues which concern effect application order? Or should I open an issue in which i reference relevant issues?

geoffthemedio:
`component:content scripting` seems relevant

agrrr3:
Would a workaround be acceptable? Like always growing shield to minimum value one?

Oberlus:
> Would a workaround be acceptable? Like always growing shield to minimum value one?

Better to always attack (trigger combat with) enemy planets even when they have no planetary defences?

agrrr3:
The default target condition for only considered armed planets valid targets (not sure why).
Will send a PR. @Oberlus please test

[...]

PR #2495, opened 2019-07-05
ShortRange weapons target enemy fighters, ships, or planets

agrrr3:
Before planets were only targeted if armed. Not sure actually why.
This should fix issue #2488


geoffthemedio:
So that ships don't waste turns attacking unarmed neutral planets or planets that can't be damaged... It would be like shooting an already-destroyed ship.

agrrr3:
So rather #2498

PR #2496, opened 2019-07-07
remove gap from diplomatic status indicator after one or more empires were eliminated

TheSilentOne1:
If an empire has been eliminated, that currently leaves an empty gap between non-eliminated empires on the diplomatic status indicator. Instead, the remaining empires should be shown next to each other, without gap, which is what this PR does.

### DIFF
--- a/UI/PlayerListWnd.cpp
+++ b/UI/PlayerListWnd.cpp
@@ -194,7 +194,7 @@ namespace {
                 if (empire.second->Eliminated()) continue;
                 square_position++;
                 if (std::find(m_empire_ids.begin(), m_empire_ids.end(), empire.first) != m_empire_ids.end()) {
-                    const GG::Clr& square_color = GetEmpire(empire.first)->Color();
+                    const GG::Clr& square_color = empire.second->Color();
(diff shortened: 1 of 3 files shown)


UI/PlayerListWnd.cpp
```diff
@@ -190,8 +190,14 @@ namespace {
...
             for (const auto& empire_id : m_empire_ids) {
+                int eliminated_empires_to_left = 0;
+                for (const auto& empire : Empires()) {
+                    if (empire.second->EmpireID() >= empire_id) break;
+                    if (empire.second->Eliminated()) eliminated_empires_to_left++;
+                }
+
                 const GG::Clr& square_color = GetEmpire(empire_id)->Color();
-                GG::Pt square_ul = ul + GG::Pt(GG::X(empire_id * (ICON_SIZE + PAD)), GG::Y0);
+                GG::Pt square_ul = ul + GG::Pt(GG::X((empire_id - eliminated_empires_to_left) * (ICON_SIZE + PAD)), GG::Y0);
```
geoffthemedio: should probably not assume empire ids are sequential starting from 0 anyway...

TheSilentOne1: Ok, I'll switch to a different method.

TheSilentOne1:
Switched to rendering method that doesn't rely on empire id.

PR #2498, opened 2019-07-07
Target infrastructure by ShortRange weapons and ensure combat

agrrr3:
…happens so planets may be invaded

Intended as Fix for #2488 

Infrastructure (construction) boosts planetary defense so ship weapons should target all planets which have infrastructure left.

Enforcing a minimal value of one for construction via FOCS is KISS and re-triggers combat every turn which allows troops to invade the planet.
Only a single bout is wasted on "already defeated" planets, which is acceptable.

I tested that construction one always is set and after attacking a shield-less defense-less planet one can invade every turn. 
Somebody probably should have a look what happens with monsters and test with force energy structures.

### DIFF
--- a/combat/CombatSystem.cpp
+++ b/combat/CombatSystem.cpp
@@ -341,6 +341,11 @@ namespace {
                             boost::make_unique<Condition::MeterValue>(
                                 METER_SHIELD,
                                 nullptr,
+                                boost::make_unique<ValueRef::Constant<double>>(0.0))),
+                        boost::make_unique<Condition::Not>(
+                            boost::make_unique<Condition::MeterValue>(
+                                METER_CONSTRUCTION,
+                                nullptr,
                                 boost::make_unique<ValueRef::Constant<double>>(0.0)))))));
 
     const std::unique_ptr<Condition::ConditionBase> if_source_is_planet_then_ships_else_all =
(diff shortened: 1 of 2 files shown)


agrrr3:
Is there a tag for like help-wanted-testing ?

combat/CombatSystem.cpp
```diff
@@ -331,17 +331,11 @@ namespace {
...
-                                METER_DEFENSE,
-                                nullptr,
-                                boost::make_unique<ValueRef::Constant<double>>(0.0))),
-                        boost::make_unique<Condition::Not>(
-                            boost::make_unique<Condition::MeterValue>(
-                                METER_SHIELD,
-                                nullptr,
-                                boost::make_unique<ValueRef::Constant<double>>(0.0)))))));
+                    boost::make_unique<Condition::Not>(
+                        boost::make_unique<Condition::MeterValue>(
```
geoffthemedio: Why remove the shield and defense tests?

agrrr3: The implicit assumption is that you always have infrastructure if you have either defense or shields.

But that is maybe not necessarily so. So at least for robustness i will expand the condition

geoffthemedio:
If you want we can merge this... Let me know.

agrrr3:
@geoffthemedio ok, please merge

PR #2503, opened 2019-07-17
add /help command to message_wnd

TheSilentOne1:
Adds a /help command for the message window that gives information about available commands, which are currently /zoom and /pedia.

PR #2504, opened 2019-07-17
Add /pm command to message_wnd

TheSilentOne1:
Add /pm command that sends a private message to a single player, as an alternative to selecting a player from the player list wnd.
Based on [#2503](https://github.com/freeorion/freeorion/pull/2503)

### DIFF
--- a/UI/ChatWnd.cpp
+++ b/UI/ChatWnd.cpp
@@ -469,17 +469,21 @@ void MessageWnd::HandlePlayerChatMessage(const std::string& text,
                                          const std::string& player_name,
                                          GG::Clr text_color,
                                          const boost::posix_time::ptime& timestamp,
-                                         int recipient_player_id)
+                                         int recipient_player_id,
+                                         bool pm)
 {
     std::string filtered_message = StringtableTextSubstitute(text);
     std::string wrapped_text = RgbaTag(text_color);
+    std::string pm_text = "";
+    if (pm)
+        pm_text = UserString("MESSAGES_WHISPER");
     const std::string&& formatted_timestamp = ClientUI::FormatTimestamp(timestamp);
     if (IsValidUTF8(formatted_timestamp))
         wrapped_text += formatted_timestamp;
     if (player_name.empty())
         wrapped_text += filtered_message + "</rgba>";
     else
-        wrapped_text += player_name + ": " + filtered_message + "</rgba>";
+        wrapped_text += player_name + pm_text + ": " + filtered_message + "</rgba>";
     TraceLogger() << "HandlePlayerChatMessage sender: " << player_name
                   << "  sender colour rgba tag: " << RgbaTag(text_color)
                   << "  filtered message: " << filtered_message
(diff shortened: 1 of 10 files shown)


o01eg:
1) How it will work if the player is disconnected?
2) Does server show it only to one player? If message will be processed by `ChatHistoryProvider.put_history_entity`, it will be shown public after server restart.

TheSilentOne1:
1. In our last games, I sent some messages which don't seem to have been transmitted. So my guess is, the current message implementation was designed for and in its current form is probably suited only for players playing simultaneously.
2. I don't know, but if you can read `ChatHistoryProvider.put_history_entity`, we can easily test if messages become public after server restart. I can send you a private message on our current game by selecting you on PlayerListWnd (but probably both of us should be online at the same time to be sure?).

o01eg:
@TheSilentOne1 I fixed disclosure of private messages in #2511 . Could you fix conflicts and rebase?

TheSilentOne1:
Rebased.

UI/ChatWnd.cpp
```diff
@@ -622,31 +622,79 @@ namespace {
...
+    }
+    else if (boost::iequals(command, "help")) {
+        *m_display += UserString("MESSAGES_HELP_COMMAND") + "\n";
+        m_display_show_time = GG::GUI::GetGUI()->Ticks();
+    }
+    else if (boost::iequals(command, "pm")) {
+        int player_id = ExtractPlayerID(params);
+        std::string message = ExtractMessage(params);
+
+        if (player_id != -1) {
```
geoffthemedio: replace `-1` with `Networking::INVALID_PLAYER_ID`

geoffthemedio:
This doesn't work in the multiplayer lobby. `ClientApp::Players` returns an empty map, even though I've got two human clients connected to the server.

geoffthemedio:
Works in-game, but it would probably be good to indicate in the message whether it was a pm or public message... like
````
[04 Aug 12:32:32] Human_Player [to All]: message to all players
[04 Aug 12:32:43] Human_Player [to OtherPlayer]: message just to one player
````

**Comment by geoffthemedio** [comment]
@TheSilentOne1 Any plans to do more for this?

**Comment by TheSilentOne1** [comment]
Yes, in the foreseeable future.

**Comment by geoffthemedio** [comment]
@TheSilentOne1 poke

**Comment by TheSilentOne1** [comment]
Soon, patience please :)

**Comment by geoffthemedio** [comment]
Typing `/pm playername` (with no message after) crashes the client. As does just `/pm`.

**Comment by TheSilentOne1** [comment]
Ouch. I was busy lately, but I'll have time to look into this and overally improve it later this week.

File: UI/ChatWnd.cpp [hunk 2]
```diff
@@ -611,42 +613,104 @@ void MessageWnd::OpenForInput() {
...
+        std::string recipient_names = "";
 
         if (recipients.empty()) {
             net.SendMessage(PlayerChatMessage(text));
         } else {
-            recipients.insert(HumanClientApp::GetApp()->PlayerID());   // ensure recipient sees own sent message
-            for (int recipient_id : recipients)
-                net.SendMessage(PlayerChatMessage(text, recipient_id));
+            for (int recipient_id : recipients) {
+                net.SendMessage(PlayerChatMessage(UserString("MESSAGES_WHISPER") + " " + text, recipient_id));
```
geoffthemedio: (Whisper) probably shouldn't be baked into the actual message text. Rather, use a string format substutution on the recipient client when displaying

UI/ChatWnd.cpp
```diff
@@ -611,42 +613,104 @@ void MessageWnd::OpenForInput() {
...
+
+                auto player_it = players.find(recipient_id);
+                if (player_it != players.end())
+                    recipient_names += player_it->second.name + "; ";
+            }
+            recipient_names = recipient_names.substr(0, recipient_names.size() - 2);
+            net.SendMessage(PlayerChatMessage("[" + UserString("TO") + ": " + recipient_names + + "] " + text, HumanClientApp::GetApp()->PlayerID()));
```
geoffthemedio: Same here... the message metadata shouldn't be baked in on the sender's side

UI/ChatWnd.cpp
TheSilentOne1: Could you give an example where or how that is done?

geoffthemedio: eg. https://github.com/freeorion/freeorion/blob/master/UI/CombatReport/GraphicalSummary.cpp#L358 or https://github.com/freeorion/freeorion/blob/master/UI/ChatWnd.cpp#L80 or search for `boost::io::str` for lots of examples.

UI/ChatWnd.cpp
TheSilentOne1: Should be fixed.

PR #2529, opened 2019-08-12
Fix AI not keeping track of split fleets

Morlic-fo:
Fixes the AI not considering newly split fleets in various parts of the AI code (for the remainder of the turn when it was split)


Cjkjvfnby:
Looks good, @Morlic-fo can you rebase to trigger CI?

PR #2530, opened 2019-08-12
Fix AI fighter calculations

Morlic-fo:
First commit fixes a bug where all parameters were offset by 1. Implying AI always considered fighter damage to be zero, the launch rate to be equal the actual damage and the capacity to be equal to the launch rate.

Second commit fixes a bug where the wrong capacity was calculated: The currentPartMeterValue for hangar capacity already returns the sum of all parts. Summing N parts up in AI code will then result in N times the actual capacity...

### DIFF
--- a/default/python/AI/CombatRatingsAI.py
+++ b/default/python/AI/CombatRatingsAI.py
@@ -3,6 +3,7 @@ from logging import warn
 
 import freeOrionAIInterface as fo
 import FleetUtilsAI
+from aistate_interface import get_aistate
 from EnumsAI import MissionType
 from freeorion_tools import get_ai_tag_grade, dict_to_tuple, tuple_to_dict, cache_by_session
 from ShipDesignAI import get_part_type
@@ -167,7 +168,9 @@ class ShipCombatStats(object):
     def get_rating(self, enemy_stats=None, ignore_fighters=False):
         """Calculate a rating against specified enemy.
 
-        :param enemy_stats: Enemy stats to be rated against
+        If no enemy is specified, will rate against the empire standard enemy
+
+        :param enemy_stats: Enemy stats to be rated against - if None
         :type enemy_stats: ShipCombatStats
         :param ignore_fighters: If True, acts as if fighters are not launched
         :type ignore_fighters: bool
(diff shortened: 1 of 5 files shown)


geoffthemedio:
> currentPartMeterValue already returns the sum of all parts

That's true for some meters, but not others, I think... But the true cases probably include fighter part meters, so OK for this context.

Morlic-fo:
> That's true for some meters, but not others, I think

You are of course right, I was talking about the specific case of capacity of fighter hangars. 

I think the commit message is specific enough, so not updating unless requested.

default/python/AI/CombatRatingsAI.py
```diff
@@ -77,13 +77,10 @@ def __str__(self):
...
     class FighterStats(object):
         """ Stores fighter-related stats """
-        def __init__(self, stat_tuple=None, capacity=0, launch_rate=0, damage=0):
-            if stat_tuple and isinstance(stat_tuple, tuple):
-                self.capacity, self.launch_rate, self.damage = stat_tuple
-            else:
-                self.capacity = capacity
-                self.launch_rate = launch_rate
-                self.damage = damage
+        def __init__(self, capacity=0, launch_rate=0, damage=0):
```
genestack-solomatin: All use cases use positional notation, let's remove keyword arguments.

geoffthemedio:
Do you need / want more feedback about this?

Morlic-fo:
The fix in this PR revealed some significant inconsistencies issues how AI ratings are handled throughout the code. 

Sometimes, they are weighted against an enemy composition, sometimes not. If fighters are involved (but also to a lesser degree for shields), that significantly skewes the AI's approximations of fleet strength in different contexts. 

I want to fix those inconsistencies before merging this PR. Setting status to WIP for now.

geoffthemedio:
@Morlic-fo status update on this?

Morlic-fo:
still WIP, will hopefully find time to finalize this weekend.

geoffthemedio:
@Morlic-fo How about now? Done? Still WIP?

Morlic-fo:
Sorry for delay. Unexpected real life stuff happened.

A complete fix for the issue mentioned earlier is beyond my current scope. But the fix I pushed should resolve the most dramatic issues (but introduces a few others). Overall, with the last commit seemed better than without the last commit, so I am going with this for now.

Morlic-fo:
CI failure unrelated.

PR #2531, opened 2019-08-12
Raise AI minimum fuel requirements for military ship designs to 2

Morlic-fo:
Playtesting of the recent fuel changes to hulls indicates that the AI currently doesn't handle fleets with low fuel capacity very well - especially considering it will generally spam only 1 type of ship. In the future, the AI should probably build a few dedicated high-fuel ships for long-range missions.

For now, require each design to have at least 2 fuel (generally easily achievable using the internal slots).

### DIFF
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -1635,6 +1635,8 @@ class MilitaryShipDesignerBaseClass(ShipDesigner):
 
     def __init__(self):
         super(MilitaryShipDesignerBaseClass, self).__init__()
+        self.additional_specifications.minimum_fuel = 2
+        self.additional_specifications.minimum_speed = 30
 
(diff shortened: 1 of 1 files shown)


agrrr3:
Rather raise it to 3 in order to get a 2 jump hop out and back range)?

Two fuel does not help a lot as 1 fuel and 2 fuel as well mean a 1 jump hop
out and back range).

Morlic-fo:
I initially considered to go with 3 fuel for the reason you mention but I think it is a bad idea for the AI for the following reasons:

1) When they move out of supply, they intend to take a planet. If the plan succeeds, they can resupply. If the plan does not succeed, they will most likely not retreat anyway.

2) Since 3 fuel is not trivially reached for some hulls until fuel techs are researched, such a requirement will reduce overall fighting power significantly (e.g. the AI would not use the large and robotic hull in the early game).

PR #2535, opened 2019-08-17
fix FramebufferFailedException

Aplaus228:


GG/src/SDL/SDLGUI.cpp
```diff
@@ -382,7 +382,7 @@ namespace {
                 default:
                     std::stringstream ss;
                     ss << "Framebuffer creation failed. Status: " << m_status;
-                    return ss.str().c_str();
+                    return to_cstr(std::move(ss).str());
```
geoffthemedio, line 385: How about `return "Framebuffer creation failed. Status: " + std::to_string(m_status);`?

`GLenum` appears to be just a typedef of `unsigned int`.

Issue #2497, opened 2019-07-07
Provide a way for external application to get information about playing game


o01eg:
### Environment
* **FreeOrion Version**: any
* **Operating System**: any

### Description

I want to get information from the server about playing game current turn, time before turn advance, game UID and so on to show it on both my servers. Currently there no API for external application to interact with the server.

### Expected Result

The server provides API to get playing game status.

geoffthemedio:
Any ideas about how to implement this? Just output to monitor the server, or polling and response messages? Network-like messages similar to the client-server communication? Server status file being written every time something changes?

o01eg:
> Any ideas about how to implement this?

Just a vague ideas. Maybe something Boost.Beast and some compatible Json serialization, but Boost.Beast isn't available in the minimal supported boost version and Boost.Serialize doesn't support JSON at all.

> Just output to monitor the server, or polling and response messages?

I suppose some polling is better. It doesn't have to be often.

> Network-like messages similar to the client-server communication?

Maybe, but Boost.Serialize is unsuitable for interaction with non-C++ application.

> Server status file being written every time something changes?

No. I think it works poorly with turn remaining time.

geoffthemedio:
Network messages don't need to use Boost serialization... Depending on the sort of data being requested, just returning the text strings, binary-encoded numbers, or numbers as text should cover most cases, except if lists or more complex data structures are needed.

Oberlus:
> > Server status file being written every time something changes?
> 
> No. I think it works poorly with turn remaining time.

Even if what you save is the **date** and time for next turn?
(edited)

o01eg:
Yes, it looks better. I'll try to use Boost.Property to get some JSON.

geoffthemedio:
As mentioned here, https://github.com/freeorion/freeorion/pull/2427

External application messaging can be done by extending the UDP discovery server.

One can set up, for example, a Python script to send a UDP message to the IP and discovery port of a freeoriond process, with the server discovery inquiry message, and can get back the UDP reply message. 
 
In the Python interpreter:
````
>>> import socket
>>> UDP_IP = '127.0.0.1'
>>> UDP_PORT = 12345
>>> MESSAGE = 'Yo, can I play Free-O here, dog?'.encode('utf-8')
>>> s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
>>> s.sendto(MESSAGE, (UDP_IP, UDP_PORT))
32
[...]

PR #2536, opened 2019-08-17
UDP FOCS

geoffthemedio:
Allows UDP messages to server to provide more information than just whether the server can host a game. Simple context-free FOCS queries are parsed and evaluated and returned by UDP message to the sender.

### DIFF
--- a/parse/Parse.h
+++ b/parse/Parse.h
@@ -72,12 +72,16 @@ namespace parse {
 
     /** Find all FOCS scripts (files with .focs.txt suffix) in \p path.  If \p allow_permissive =
         true then if \p path is not empty and there are no .focs.txt files allow all files to qualify.*/
-    FO_PARSE_API std::vector< boost::filesystem::path > ListScripts(const boost::filesystem::path& path, bool permissive = false);
+    FO_PARSE_API std::vector<boost::filesystem::path> ListScripts(const boost::filesystem::path& path, bool permissive = false);
 
     FO_PARSE_API void file_substitution(std::string& text, const boost::filesystem::path& file_search_path);
-    FO_PARSE_API void process_include_substitutions(std::string& text, 
-                                                    const boost::filesystem::path& file_search_path, 
+    FO_PARSE_API void process_include_substitutions(std::string& text,
+                                                    const boost::filesystem::path& file_search_path,
                                                     std::set<boost::filesystem::path>& files_included);
+
+    FO_PARSE_API bool int_free_variable(std::string& text);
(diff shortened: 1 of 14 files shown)


geoffthemedio:
Not really but perhaps somewhat relevant to https://github.com/freeorion/freeorion/issues/2497

o01eg:
Is it safe to execute FOCS scripts from unknown source?

o01eg:
Also it could lead to the game disclosure.

geoffthemedio:
The usable "scripts" are quite limited... Just single free variables of int, double, or string. This means stuff that doesn't reference any object... for example `CurrentTurn`, `GalaxySeed`, or `UniverseWidth`. No conditions or effects. No `Source.Population` or `SetPopulation`. No statistics or arithmetic.

That said, I can add a bunch more stuff within or outside the FOCS portion, to provide additional info that would be useful. Ideally stuff that just needs a single word to describe, ie. no parameters, and can be returned as a string, either by the result naturally being a string, or by composition of multiple strings, or casting a number to a string as is already done for the non-string free variables after evaluating them.

o01eg:
Ok, then it looks good. Could multiple variables be requested with a single request?

geoffthemedio:
> Could multiple variables be requested with a single request?

In principle, something could be set up to do a more complicated parse of the input, but why not just send multiple queries?

o01eg:
> In principle, something could be set up to do a more complicated parse of the input, but why not just send multiple queries?

I suppose single query is a simple for external application. But this PR is fine although.

parse/Parse.h
```diff
@@ -72,12 +72,16 @@ namespace parse {
...
     FO_PARSE_API void file_substitution(std::string& text, const boost::filesystem::path& file_search_path);
-    FO_PARSE_API void process_include_substitutions(std::string& text, 
-                                                    const boost::filesystem::path& file_search_path, 
+    FO_PARSE_API void process_include_substitutions(std::string& text,
+                                                    const boost::filesystem::path& file_search_path,
                                                     std::set<boost::filesystem::path>& files_included);
+
+    FO_PARSE_API bool int_free_variable(std::string& text);
+    FO_PARSE_API bool double_free_variable(std::string& text);
+    FO_PARSE_API bool string_free_variable(std::string& text);
```
Morlic-fo, line 84: I think the compilation units where these functions are defined do not to include ```parse.h``` and thus will never see the FO_PARSE_API declaration. Subsequently, the function visibility is still set to hidden and the linker won't find them.

geoffthemedio: Looks like that was it, thanks.

Issue #2534, opened 2019-08-15
Use after free in FramebufferFailedException 


mgehre:
On latest clang trunk, the code
```
std::stringstream ss;
ss << "Framebuffer creation failed. Status: " << m_status;
return ss.str().c_str();
```
in https://github.com/freeorion/freeorion/blob/master/GG/src/SDL/SDLGUI.cpp#L385

causes the warning 
```
warning: object backing the pointer will be destroyed at the end of the full-expression [-Wdangling]
```
because `str()` returns a temporary `std::string`, and the function returns a `const char*` into that temporary, which is destroyed before the function returns.

Aplaus228:
I would  try to fix this bug today but later

PR #2537, opened 2019-08-17
fix FramebufferFailedException

Aplaus228:


GG/src/SDL/SDLGUI.cpp
```diff
@@ -381,8 +380,10 @@ namespace {
                 case GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT_EXT:
                     return "One of the framebuffer attachments is incomplete.";
                 default:
-                	return "Framebuffer creation failed. Status: " + std::to_string(m_status);
-
+                	std::stringstream ss;
+                    ss << "Framebuffer creation failed. Status: " << m_status;
+					static thread_local std::string sloc = std::move(ss).str();
```
geoffthemedio: replace tabs with 4 spaces

geoffthemedio: This looks rather fishy... I think the clean solution is to return a string literal for all possible values of `m_status` and not do any string concatenation in local temporaries.

Aplaus228: sry, but I don't know how to convert from unsigned int directly into const char* without string.

Aplaus228:
someone can clarify for me why Travis did not finish with success.

Morlic-fo:
The function is marked as ```noexcept``` but may throw ```std::bad_alloc```.

GG/src/SDL/SDLGUI.cpp
```diff
@@ -380,9 +380,8 @@ namespace {
                 case GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT_EXT:
                     return "One of the framebuffer attachments is incomplete.";
                 default:
-                    std::stringstream ss;
-                    ss << "Framebuffer creation failed. Status: " << m_status;
-                    return ss.str().c_str();
+                    std::string * s= new std::string("Framebuffer creation failed. Status: " + std::to_string(m_status));
+                    return s->c_str();
```
geoffthemedio: This will leak memory each time it's executed, as you create a new string and don't delete it.

geoffthemedio: Again, instead of trying to create and return a pointer to a local that goes out of scope, just handle all the possible, or at least some more of the possible, values of `m_status`. See https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glCheckFramebufferStatus.xhtml 

Or more simply, if you don't want to handle all the possible cases, just do
````
default:
    return "Framebuffer creation failed somehow.";
````

File: GG/src/SDL/SDLGUI.cpp (hunk 1 above)
Comment (geoffthemedio) [reply GG/src/SDL/SDLGUI.cpp]: You don't convert an `int` to a `const char*`. You extend the `switch` block to return a suitable `"string literal"` for each possible value of `m_status`.

Comment (geoffthemedio) [reply GG/src/SDL/SDLGUI.cpp]: The point is you can't do the string concatenation of local temporary variables that will go out of scope when the function returns, as you're returning a `char*` that will point to the location of the result of that concatenation, which will not be a valid pointer after the function returns. Using a `stringstream` or combining `"string literals"` and `std::string`s with `+` doesn't change this, if the result is stored in a local and you can't return the result by value or rvalue reference, which you can't because the function signature is just `const char*`.

File: GG/src/SDL/SDLGUI.cpp [hunk 3]
```diff
@@ -379,10 +379,14 @@ namespace {
                     return "The requested framebuffer format was unsupported";
                 case GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT_EXT:
                     return "One of the framebuffer attachments is incomplete.";
+                case GL_FRAMEBUFFER_UNDEFINED:
+                    return "The default framebuffer does not exist.";
+                case GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT_EXT:
+                    return "One of the image framebuffer does not have attached to it.";
```
geoffthemedio: does not have what attached to it?

Aplaus228: thank you for your poise and advise's

GG/src/SDL/SDLGUI.cpp
```diff
@@ -382,7 +382,7 @@ namespace {
                 case GL_FRAMEBUFFER_UNDEFINED:
                     return "The default framebuffer does not exist.";
                 case GL_FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT_EXT:
-                    return "One of the image framebuffer does not have attached to it.";
+                    return "At least one picture must be attached to еру framebuffer.";
```
geoffthemedio: еру?

GG/src/SDL/SDLGUI.cpp
o01eg: I suppose other way is to do string concatenation in the exception's constructor and save result in a field.

GG/src/SDL/SDLGUI.cpp
Aplaus228: sry, but I immediately fixed it.

GG/src/SDL/SDLGUI.cpp
geoffthemedio: Could work, but more allocation that could throw in an exception constructor might not be ideal...

geoffthemedio:
Can you squash into a single commit?

Aplaus228:
yes, I have done it.

Vezzra:
@Aplaus228, @mgehre, @geoffthemedio, am I correct in assuming this is supposed to fix #2534? If yes, can we close #2534 as resolved?

Issue #2521, opened 2019-08-03
Fuel Efficiency Issues with Self Grav Hulls


urshmost:
### Environment
Build 2019-7-30 win32

### Description
The self grav hulls now has a fuel rating of 1.5, and fuel efficiency of 60% which gives them a max fuel value of 0.9. This somehow allows them to travel within your supply range, but as soon as they leave it they are stranded.

Sitting at .9 fuel, every self grav ship, on every turn, generates a message stating that they now have sufficient fuel to travel again, but because it appears to be capped at .9, they really don't.

### Expected Result
I would expect the message that a ship has enough fuel to travel to not be generated when the ship ultimately will not have sufficient fuel.
I would expect that any ship design should have at least a minimum fuel of 2 or 3, otherwise you are effectively forcing extra fuel tanks to always be necessary, and at that point what is the reason over just giving the design more fuel and taking away a slot?

agrrr3:
The need for the supply network for less than one fuel is intended. The big hulls need their supply networks (or good fuel tech) to operate.

The message should definitley be fixed. 

Besides of this this is a UI and balancing issue. A part of the balancing was kept as fuel efficiency was applied to all ships' base fuel (so the big hulls only have a fourth of fuel before). 

Difference between taking away a slot while adding base fuel vs adding a fuel part is that fuel part bonus increases better fuel tech. I am undecided if it is possible to design ships with less than one max fuel are an issue or not.
Maybe all ships designs should be able to return to supply network. Maybe in many hulls this is a tradeoff between having an extra internal slot. And having good fuel pilots or energy metabolism tech also can kick the hull over the 1 fuel barrier without sacrificing the slot.

geoffthemedio:
Could perhaps use `floor(...)` to round down rather than having fractional fuel values.

agrrr3:
> Could perhaps use `floor(...)` to round down rather than having fractional fuel values.

That is great idea for 0.x max fuel  (floor it to zero max fuel) - for other fractionals i would like to get some feedback from playtesting first (see the discussion in [New Fuel Settings](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11321&p=96805#p96805)

agrrr3:
Oberlus brought one real change of concept up - before it was always possible to refuel and do a jump if you are out of supply. This is currently not possible anymore - you would have to extend your supply to reach the ship or research energy metabolism.

geoffthemedio:
Did #2538 resolve this?

[...]

PR #2538, opened 2019-08-21
Ensure that no hull has less than 1 fuel

agrrr3:
This implements a solution to #2521 - [forum post](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11321&p=96805#p96860); with this fix all hulls have more than 1 effective fuel so it is impossible to get ships completely stranded
* heavy asteroid hull looses one internal slot and gains 0.6 effective fuel
* logistics facilitator looses one external slot and gains 0.3 effective fuel
* self gravitating hull looses one external slot and gains 0.3 effective fuel

### DIFF
--- a/default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
@@ -2,17 +2,16 @@ Hull
     name = "SH_HEAVY_ASTEROID"
     description = "SH_HEAVY_ASTEROID_DESC"
     speed = 60
-    fuel = 1
+    fuel = 2
     stealth = 5
     structure = 50
     slots = [
-        Slot type = External position = (0.30, 0.25)
         Slot type = External position = (0.45, 0.25)
         Slot type = External position = (0.60, 0.25)
+        Slot type = External position = (0.30, 0.50)
         Slot type = External position = (0.75, 0.50)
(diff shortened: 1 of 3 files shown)


Morlic-fo:
Reasoning for the slot changes? Seems like a pretty substantive nerf.

agrrr3:
> Reasoning for the slot changes? Seems like a pretty substantive nerf.

From the  linked discussion:
> I was thinking of exchange rate of a single internal slot for 2 base fuel - that would translate for the bad efficiency hulls to 0.6 extra fuel in exchange.
> But for self-grav ships it would be rather crippling to remove one of the two internal slots. How about also exchange a single external slot for 1 base fuel (so 0.3)...

Exchanging slots for fuel is for keeping the relative balance between the hulls. The nerf in a way happened when the fuel efficiency was introduced, i did some playtesting and the balance looked reasonable to my eyes. All bad efficiency hulls have only a quarter of their effective jump distance and with this PR the player does not have to add a fuel tank to these three hulls in order to have at least 1 maximum fuel. I am not sure how much the absolute balance makes sense as i do not use these hulls so much. I also tried to get feedback from the community but no response.

And as this is a design issue, discussion should probably happen in the forum thread(?).

Vezzra:
> And as this is a design issue, discussion should probably happen in the forum thread(?).

Yes, please. :smiley:

geoffthemedio:
Linked thread has some additional balance suggestions, but apparently no objections to this change, so OK.

PR #2542, opened 2019-09-02
Implement team support at start

o01eg:
Basic support for teams. Each players in the same team are initialized to be in alliance.

### DIFF
--- a/default/python/auth/auth.py
+++ b/default/python/auth/auth.py
@@ -9,6 +9,9 @@ import sys
 
 import freeorion as fo
 
+# Constants defined by the C++ game engine
+NO_TEAM_ID = -1
+
 
 class AuthProvider:
     def __init__(self):
@@ -72,5 +75,6 @@ class AuthProvider:
                 psd.player_name = player_name
                 psd.empire_name = player_name
                 psd.starting_species = "RANDOM"
+                psd.starting_team = NO_TEAM_ID
                 players.append(psd)
         return players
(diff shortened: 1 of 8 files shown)


default/python/auth/auth.py
```diff
@@ -72,5 +72,6 @@ def list_players(self):
                 psd.player_name = player_name
                 psd.empire_name = player_name
                 psd.starting_species = "RANDOM"
+                psd.starting_team = -1
```
geoffthemedio: Probably better to initialize with `INVALID_PLAYER_ID` to avoid "magic number" in code.

util/MultiplayerCommon.h
```diff
@@ -235,7 +235,8 @@ struct PlayerSetupData {
         m_save_game_empire_id(ALL_EMPIRES),
         m_client_type(Networking::INVALID_CLIENT_TYPE),
         m_player_ready(false),
-        m_authenticated(false)
+        m_authenticated(false),
+        m_starting_team(-1)
```
geoffthemedio: ...

geoffthemedio:
otherwise seems OK

default/python/auth/auth.py
o01eg: For python `INVALID_ID = -1` defined in the `AIDependencies.py`. Also it isn't player id at all. Maybe add new constant?

geoffthemedio: Another constant also OK.

util/MultiplayerCommon.h
o01eg: Fixed.

default/python/auth/auth.py
o01eg: I've added `NO_TEAM_ID` constant.

PR #2543, opened 2019-09-03
Add game rule to manage diplomatic interaction between players

o01eg:
Uses set of allowed values to be extended in the future like in FreeCiv:

freeciv-d

### DIFF
--- a/Empire/EmpireManager.cpp
+++ b/Empire/EmpireManager.cpp
@@ -2,6 +2,8 @@
 
 #include "Empire.h"
 #include "../util/Directories.h"
+#include "../util/GameRules.h"
+#include "../util/i18n.h"
 #include "../util/Logger.h"
 #include "../util/XMLDoc.h"
 #include "../universe/Enums.h"
@@ -225,6 +227,9 @@ void EmpireManager::RemoveDiplomaticMessage(int sender_id, int recipient_id) {
 void EmpireManager::HandleDiplomaticMessage(const DiplomaticMessage& message) {
     int sender_empire_id = message.SenderEmpireID();
     int recipient_empire_id = message.RecipientEmpireID();
+
+    if (!message.IsAllowed())
+        return;
 
     DiplomaticStatus diplo_status = GetDiplomaticStatus(sender_empire_id, recipient_empire_id);
     bool message_from_recipient_to_sender_available = DiplomaticMessageAvailable(recipient_empire_id, sender_empire_id);
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1227,6 +1227,18 @@ Show detailed empires data
 RULE_SHOW_DETAILED_EMPIRES_DATA_DESC
 Show research and production, researched buildings, parts, and hulls of all empires.
 
+RULE_DIPLOMACY
+Allowed diplomacy
+
+RULE_DIPLOMACY_DESC
+Manages diplomatic restrictions between players. By default all interactions are allowed.
+
+RULE_DIPLOMACY_ALLOWED_FOR_ALL
+All Allowed
+
+RULE_DIPLOMACY_FORBIDDEN_FOR_ALL
+All Forbidden
+
 ##
 ## Command-line and options database entries
 ##
--- a/Empire/Diplomacy.cpp
+++ b/Empire/Diplomacy.cpp
@@ -1,6 +1,22 @@
 #include "Diplomacy.h"
 
+#include "../util/GameRules.h"
+#include "../util/i18n.h"
+
 FO_COMMON_API extern const int ALL_EMPIRES;
+
+namespace {
+    void AddRules(GameRules& rules) {
+        // determine if diplomacy allowed
+        rules.Add<std::string>("RULE_DIPLOMACY", "RULE_DIPLOMACY_DESC",
+                               "MULTIPLAYER", "ALLOWED_FOR_ALL", false,
+                               DiscreteValidator<std::string>(std::set<std::string>({
+                                   "RULE_DIPLOMACY_ALLOWED_FOR_ALL",
+                                   "RULE_DIPLOMACY_FORBIDDEN_FOR_ALL"
+                               })));
+    }
+    bool temp_bool = RegisterGameRules(&AddRules);
+}
 
 DiplomaticMessage::DiplomaticMessage() :
     m_sender_empire(ALL_EMPIRES),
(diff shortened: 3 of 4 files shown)


Empire/EmpireManager.cpp
```diff
@@ -226,6 +238,9 @@ void EmpireManager::HandleDiplomaticMessage(const DiplomaticMessage& message) {
     int sender_empire_id = message.SenderEmpireID();
     int recipient_empire_id = message.RecipientEmpireID();
 
+    if (GetGameRules().Get<std::string>("RULE_DIPLOMACY") == "RULE_DIPLOMACY_FORBIDDEN_FOR_ALL")
```
geoffthemedio: String literals that are used as stringtable entry keys, if not being passed as a parameter to a function, should generally be wrapped in `UserStringNoOp` (or whatever it is, I can't check the spelling right now)

Empire/EmpireManager.cpp
```diff
@@ -2,13 +2,25 @@
...
 #include <boost/filesystem/fstream.hpp>
 
 namespace {
+    void AddRules(GameRules& rules) {
+        // determine if diplomacy allowed
+        rules.Add<std::string>("RULE_DIPLOMACY", "RULE_DIPLOMACY_DESC",
+                               "MULTIPLAYER", "ALLOWED_FOR_ALL", false,
+                               DiscreteValidator<std::string>(std::set<std::string>({
+                                   "RULE_DIPLOMACY_ALLOWED_FOR_ALL",
+                                   "RULE_DIPLOMACY_FORBIDDEN_FOR_ALL"
```
geoffthemedio: Might also make sense to wrap the literals here... not sure what standard practice for this would be as I don't think there are many cases of `DiscreteValidator` with strings...

Empire/EmpireManager.cpp
```diff
@@ -2,13 +2,25 @@
...
 #include "../util/Directories.h"
+#include "../util/GameRules.h"
 #include "../util/Logger.h"
 #include "../util/XMLDoc.h"
 #include "../universe/Enums.h"
 
 #include <boost/filesystem/fstream.hpp>
 
 namespace {
+    void AddRules(GameRules& rules) {
```
geoffthemedio: Why is this here and not in Diplomacy.cpp?

default/stringtables/en.txt
```diff
@@ -1227,6 +1227,18 @@ Show detailed empires data
 RULE_SHOW_DETAILED_EMPIRES_DATA_DESC
 Show research and production, researched buildings, parts, and hulls of all empires.
 
+RULE_DIPLOMACY
+Manage diplomatic restrictions
```
geoffthemedio: Maybe "Allowed diplomacy"

Empire/EmpireManager.cpp
o01eg: Because it is used only here.

Empire/EmpireManager.cpp
o01eg: In the `CombatEvents.cpp` nothing is wrapped.

Empire/EmpireManager.cpp
o01eg: Moved code to `Dimplomacy.cpp`

default/stringtables/en.txt

Empire/EmpireManager.cpp
o01eg: Wrapped.

Empire/EmpireManager.cpp
geoffthemedio: Those aren't using `DiscreteValidator` though.

Looking over code a bit, it looks like in (almost?) all cases, calls to `OptionsDB::Add` wrap their string-literal parameters for option name in `UserStringNop`, but in no case I saw does a call to `GameRules::Add` do so. It probably should in all cases, though it doesn't need to be done so for this pull request specifically.

geoffthemedio:
Some UI issues...
Diplomacy_Rule_Setup_Layout

o01eg:
@geoffthemedio Test rule for discrete validator looks alike.

geoffthemedio:
Test rule doesnt have the same text overflow issues for me.

o01eg:
It looks like combobox issue. You could try to add characters to `en.txt` to reproduce it.

geoffthemedio:
The issue is that the box is not wide enough for your text to fit, so one or both need to change size. Can you do that, or do you need / want me to?

o01eg:
@geoffthemedio I think it's a dropdownlist issue: https://github.com/freeorion/freeorion/issues/2545

geoffthemedio:
DropLists, including `StringRuleWidget` be made to now allow row contents to render outside them, but it doesn't really help much with this issue...
diplo_drop
I think your text is just too long for this space... even if the droplist is made a bit wider, more than one word is difficult to fit and all the text starts with the same word. Much wider and the description won't fit well...

Edit: I made the droplists (and the whole setup window and rules panel) wider, so the abbreviated text suggested below can fit.

default/stringtables/en.txt
```diff
@@ -1227,6 +1227,18 @@ Show detailed empires data
...
 Show research and production, researched buildings, parts, and hulls of all empires.
 
+RULE_DIPLOMACY
+Allowed diplomacy
+
+RULE_DIPLOMACY_DESC
+Manages diplomatic restrictions between players. By default all interactions are allowed.
+
+RULE_DIPLOMACY_ALLOWED_FOR_ALL
+Diplomacy allowed for all
```
geoffthemedio: `All Allowed`

default/stringtables/en.txt
```diff
@@ -1227,6 +1227,18 @@ Show detailed empires data
...
+
+RULE_DIPLOMACY_FORBIDDEN_FOR_ALL
+Diplomacy forbidden for all
```
geoffthemedio: `Always War` or similar

o01eg: I suppose it better to name `All forbidden` because for the same team #2542 it supposed to prevent breaking an alliance.

default/stringtables/en.txt

default/stringtables/en.txt
o01eg: I suppose it better to name `All forbidden` because for the same team #2542 it supposed to prevent breaking an alliance.

geoffthemedio: It's not really clear from the name that it means "no changes after start" and not "always war between everyone"

PR #2544, opened 2019-09-03
Notify clients about changes in the player list

o01eg:
Make connections and disconnection player visible. Fixes chat issues in the multiplayer game where client doesn't know player data of connected player.

### DIFF
--- a/network/Message.cpp
+++ b/network/Message.cpp
@@ -687,6 +687,15 @@ Message EliminateSelfMessage()
 Message UnreadyMessage()
 { return Message(Message::UNREADY, DUMMY_EMPTY_MESSAGE); }
 
+Message PlayerInfoMessage(const std::map<int, PlayerInfo>& players) {
+    std::ostringstream os;
+    {
+        freeorion_xml_oarchive oa(os);
+        oa << BOOST_SERIALIZATION_NVP(players);
+    }
+    return Message(Message::PLAYER_INFO, os.str());
+}
+
 ////////////////////////////////////////////////
 // Message data extractors
 ////////////////////////////////////////////////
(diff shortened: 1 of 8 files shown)


network/Message.cpp
```diff
@@ -1309,3 +1318,16 @@ void ExtractAuthResponseMessageData(const Message& msg, std::string& player_name
...
+void ExtractPlayerInfoMessageData(const Message &msg, std::map<int, PlayerInfo>& players) {
+    try {
+        std::istringstream is(msg.Text());
+        freeorion_xml_iarchive ia(is);
+        ia >> BOOST_SERIALIZATION_NVP(players);
+    } catch(const std::exception& err) {
+        ErrorLogger() << "ExtractPlayerInfo(const Message &msg, std::map<int, PlayerInfo>& players) failed!  Message:\n"
+                      << msg.Text() << "\n"
+                      << "Error: " << err.what();
+        throw err;
```
Morlic-fo: Probably should rethrow original exception with ```throw;``` rather than throwing a sliced copy?

o01eg: Fixed

geoffthemedio:
I'm surprised player connections aren't already included in lobby updates... but if not then OK.

network/Message.cpp
```diff
@@ -1309,3 +1318,16 @@ void ExtractAuthResponseMessageData(const Message& msg, std::string& player_name
...
+void ExtractPlayerInfoMessageData(const Message &msg, std::map<int, PlayerInfo>& players) {
+    try {
+        std::istringstream is(msg.Text());
+        freeorion_xml_iarchive ia(is);
+        ia >> BOOST_SERIALIZATION_NVP(players);
+    } catch(const std::exception& err) {
+        ErrorLogger() << "ExtractPlayerInfo(const Message &msg, std::map<int, PlayerInfo>& players) failed!  Message:\n"
+                      << msg.Text() << "\n"
+                      << "Error: " << err.what();
+        throw;
```
geoffthemedio: Maybe better to catch and log the error in the calling code instead or as well so that a serialization error doesn't crash the client

server/ServerFSM.cpp
```diff
@@ -452,15 +452,19 @@ void ServerFSM::UpdateIngameLobby() {
     }
     dummy_lobby_data.m_start_lock_cause = UserStringNop("SERVER_ALREADY_PLAYING_GAME");
 
-    // send it to all those without empire
-    // and who are CLIENT_TYPE_HUMAN_PLAYER
+    auto player_info_map = m_server.GetPlayerInfoMap();
```
geoffthemedio, line 455: Can / should this be a const reference `const auto&` ?

network/Message.cpp
o01eg: Yes. It isn't critical message.

server/ServerFSM.cpp
o01eg: No, it cann't. `GetPlayerInfoMap` doesn't return existing map but builds new one and return it by value.

o01eg:
In real and ingame lobby it works. Only players who are actually playing the game weren't notified.

Issue #2482, opened 2019-06-18
Can't start new game with weekly build on Mac OS X


jsena210:
### Environment

* **FreeOrion Version**: 0.4.8+ 2019-06-11
* **Operating System**: Mac OS X 10.12.6
* **Graphic card used**: NVIDIA GeForce GT 750M
* **Fetched as** 
  * Weekly development build 2019-06-11

### Description
When creating a new game (not "continue" or "quick start"), when you start game, game crashes without GUI error.

If launched via command line, this is what you see when you hit "OK":

main() caught exception(std::exception): boost::bad_any_cast: failed conversion using boost::any_cast
GUIController failed to setup or run the GUI.

### Expected Result
Game starts

### Steps to reproduce
Open app
Hit Single Player
Hit OK

jsena210:
FYI, This bug appeared in weekly build 2019-04-22 and was not manifesting in the build before that.

geoffthemedio:
Any change with build [2019-06-16.ecb7a95](http://sourceforge.net/projects/freeorion/files/FreeOrion/Test/FreeOrion_2019-06-16.ecb7a95_MacOSX_10.9.dmg/download) ?

jsena210:
No change with 2019-06-16

geoffthemedio:
Please post freeorion.log from after attempting to open the galaxy setup window / clicking single player using the latest version. There's new / additional logging that might be helpful, particularly the stuff after the last "parse_file" line.

jsena210:
Here is the stuff past the parse_file:

18:23:04.505278 {0x00007fffe03bc3c0} [info] log : Logger.cpp:198 : Added logger named "network"
18:23:04.505354 {0x00007fffe03bc3c0} [info] log : Logger.cpp:331 : Setting "network" logger threshold to "debug".
18:23:04.505370 {0x00007fffe03bc3c0} [debug] log : LoggerWithOptionsDB.cpp:105 : Configure log source "network" from optionsDB using threshold debug
18:23:04.505388 {0x00007fffe03bc3c0} [debug] network : ClientNetworking.cpp:349 : Attempt to connect to server at one of these addresses:
18:23:04.505458 {0x00007fffe03bc3c0} [debug] network : ClientNetworking.cpp:351 :   tcp::resolver::iterator host_name: 127.0.0.1  address: 127.0.0.1  port: 12346
[...]

PR #2546, opened 2019-09-11
GalaxySetupWnd OK Tweaks

geoffthemedio:
Wrapped stuff in try-catch blocks, logging, and related tweaks to diagnose and maybe help with https://github.com/freeorion/freeorion/issues/2482


geoffthemedio:
pushed commit directly to master

PR #2552, opened 2019-09-14
add new graphic for ion storm

TheSilentOne1:


geoffthemedio:
Seems too close in colour to the molecular cloud and galaxy gas... all are similar blue-green colours.

TheSilentOne1:
Don't think so:

fields

geoffthemedio:
I'm not saying they can't be told apart, but I'm not colourblind and I know there is a difference between them. For a new player or someone who can't distinguish colours as well, it might be more difficult.

TheSilentOne1:
If you feel colors need to be altered to make the fields more distinctive, I'd rather change the molecular cloud. The ion storm lightning color seems right to me. I'll post some suggestions on the forums.

geoffthemedio:
The apparent size of the ion storm is quite a bit smaller than its area of effect, partly because the graphic proposed doesn't fill the texture size...
ion size

TheSilentOne1:
Here's the cropped version.

geoffthemedio:
Slightly but not much better...
small ion appearance

TheSilentOne1:
Not much I can do without giving the storm a solid border. I think it might be a good idea to add a marker to fleet buttons (similar to the blockade marker) and update the planet status indicator if a planet or fleet gets affected by a field. Would that be possible with a reasonable amount of work?

PR #2562, opened 2019-09-23
Launching fighters uncloaks stealth carrier

agrrr3:
This PR intends to fix the "feature" that stealth carriers (ships with launch bays, hangars and no ShortRange weapons) are basically invincible as long as the ship stealth is higher than enemy's detection.

This PR makes launching fighters on par with firing a weapon from a stealthed ship - it makes the ship known to the enemies.

This is WIP. The carriers correctly unstealth but there is no event shown in the combat report. I do not grok the event code fully yet so I need some assistance/review how to proceed with the implementation.

### DIFF
--- a/combat/CombatSystem.cpp
+++ b/combat/CombatSystem.cpp
@@ -1746,6 +1746,32 @@ namespace {
                                launches_event);
 
                 DebugLogger(combat) << "Attacker: " << attacker->Name();
+
+                // Set launching carrier as at least basically visible to other empires.
+                if (launches_event) {
+                    for (auto detector_empire_id : combat_info.empire_ids) {
+                        Visibility initial_vis = combat_info.empire_object_visibility[detector_empire_id][attacker_id];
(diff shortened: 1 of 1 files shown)


agrrr3:
Ok, implementation now works as intended - SitRep that foreign empires can detect . Besides stealth carrriers I also checked for attack carriers (having ShortRange weapon besides fighters).

I had a look at the failing CI reports but those looked unrelated - so I would say the fails are bleeps.

So... ready-for-review.

combat/CombatSystem.cpp
```diff
@@ -1746,6 +1747,31 @@ namespace {
                                launches_event);
 
                 DebugLogger(combat) << "Attacker: " << attacker->Name();
+
+                // Set launching carrier as at least basically visible to other empires.
+                if (launches_event) {
```
geoffthemedio: does this test do anything?

agrrr3: It should have tested that there were actually launch events

agrrr3:
#2569 should fix the CI errors.

PR #2570, opened 2019-09-26
Inform other empires about detection increase

agrrr3:
As the SitReps for newly researched tech are generated on the same turn as the meter effects this is a simple implementation to inform players that they observed another empire upgrading their detection tech. 

So this supersedes #2344

[SitRep for enemy sensor tech advances?](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11150)

### DIFF
--- a/default/customizations/custom_sitreps.txt
+++ b/default/customizations/custom_sitreps.txt
@@ -104,6 +104,33 @@ EffectsGroup
             ]
             empire = Source.Owner
 
+EffectsGroup
+    scope = And [
+        Capital
+        Not OwnedBy empire = Source.Owner
+        Number low = 1 condition = And [
+            // maybe speed up by ORing subcategories
+            OwnedBy empire = Source.Owner
+            VisibleToEmpire empire = RootCandidate.Owner
+        ]
+    ]
+    activation = Or [
+       (CurrentTurn == TurnTechResearched empire = Source.Owner name = "SPY_DETECT_2")
+       (CurrentTurn == TurnTechResearched empire = Source.Owner name = "SPY_DETECT_3")
+       (CurrentTurn == TurnTechResearched empire = Source.Owner name = "SPY_DETECT_4")
+       (CurrentTurn == TurnTechResearched empire = Source.Owner name = "SPY_DETECT_5")
+    ]
+    priority = [[END_CLEANUP_PRIORITY]]
+    effects =
+        GenerateSitRepMessage
+            message = "SITREP_EMPIRE_TECH_RESEARCHED_DETECTION"
+            label = "SITREP_EMPIRE_TECH_RESEARCHED_DETECTION_LABEL"
+            icon = "icons/tech/active_radar.png"
+            parameters = [
+                tag = "empire"    data = Source.Owner
+                tag = "dstrength" data = EmpireMeterValue empire = Source.Owner meter = "METER_DETECTION_STRENGTH"
+            ]
+            empire = Target.Owner
 
 EffectsGroup
     scope = And [
--- a/default/customizations/custom_sitreps.txt
+++ b/default/customizations/custom_sitreps.txt
@@ -104,15 +104,11 @@ EffectsGroup
             ]
             empire = Source.Owner
 
+// Tell all the others that you have better detection tech (the information is public anyway)
 EffectsGroup
     scope = And [
         Capital
         Not OwnedBy empire = Source.Owner
-        Number low = 1 condition = And [
-            // maybe speed up by ORing subcategories
-            OwnedBy empire = Source.Owner
-            VisibleToEmpire empire = RootCandidate.Owner
-        ]
     ]
     activation = Or [
        (CurrentTurn == TurnTechResearched empire = Source.Owner name = "SPY_DETECT_2")
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -6576,7 +6576,7 @@ SITREP_EMPIRE_TECH_RESEARCHED_DETECTION
 The %empire% [[encyclopedia DETECTION_TITLE]] has upgraded to %rawtext:dstrength%
 
 SITREP_EMPIRE_TECH_RESEARCHED_DETECTION_LABEL
-Enemy Tech Upgrade
+Empire Detection Tech Upgrade
 
 SITREP_TECH_UNLOCKED
 %tech% has been unlocked and can now be researched.
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -6572,6 +6572,12 @@ SITREP_TECH_RESEARCHED
 SITREP_TECH_RESEARCHED_LABEL
 Tech Researched
 
+SITREP_EMPIRE_TECH_RESEARCHED_DETECTION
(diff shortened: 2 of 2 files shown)


default/customizations/custom_sitreps.txt
```diff
@@ -104,6 +104,45 @@ EffectsGroup
...
+            parameters = [
+                tag = "empire"  data = Source.Owner
+            ]
+            empire = Target.Owner
+
+EMPIRE_TECH_RESEARCHED
+'''And [
+        (CurrentTurn > 1)
+        Or [
+            (CurrentTurn == TurnTechResearched empire = Source.Owner name = "@1@")
```
geoffthemedio: Why both the turn the tech is researched and the turn after? Does this generate two sitreps saying the same thing, one turn apart?

agrrr3: This is a "leftover" from working on shifting the tech researched turn.
The FOCS implementation works for different backend implementations.

With the old and the also new backend implementation CurrentTurn == TurnTechResearched will never fire.

I left it in intentionally as i think if there are again changes, a double sitrep will more likely be noticed than a missing sitrep so we get a timely bug report in the worst case.

geoffthemedio: Add a comment explaining this then...

agrrr3: With the new backend implementation this actually changed. Now i get two sitrep messages.

Currently (11a579785a9e1a067b12d5fb2c75c2cbdcf1608f)  it is like:
If "tech researched" sitrep is shown on turn 5, the "upgraded detection technology" is shown on turn 6, by the  CurrentTurn == TurnTechResearched (plus nothing) condition. The TurnTechResearched for the tech is turn 5. For an initial tech it is -32768 and for an unresearched tech it is zero. 

Besides the break in the continuum (initial tech < no tech < normal tech) this looks quite sensible.

geoffthemedio: I don't understand you point in that comment...

agrrr3: It means i need to update this PR because the backend changed and now one gets the sitrep info twice.

agrrr3: And it would be good if initial TurnResearched < normal TurnResearched < not researched TurnResearched. 
So for unresearched tech TurnResearched could/should return maximum int. 
So that currentTurn >= TurnResearched means that the tech is actually researched

geoffthemedio: It would be somewhat awkward to add such a special case in the C++ code specifically for `TurnTechResearched`

agrrr3: Then we always have to check awkwardly on use site (FOCS)

agrrr3:
The CI complains something about something going wrong with joining/disconnecting players on a test case game. I say it is not connected with the changes in this PR (FOCS/stringtables only).

default/customizations/custom_sitreps.txt
agrrr3: > It would be somewhat awkward to add such a special case in the C++ code specifically for `TurnTechResearched`

I checked the code and now I see what you mean. Because this is generic valueref handling of maps it would be hard to add a value-to-return-if-not-found in a generic and effective way. Probably a separate path in ComplexVariable<int>::Eval  which passes the value-to-return-if-not-found to GetIntEmpirePropertySingleKey.

default/stringtables/en.txt
```diff
@@ -6572,6 +6572,12 @@ SITREP_TECH_RESEARCHED
 SITREP_TECH_RESEARCHED_LABEL
 Tech Researched
 
+SITREP_EMPIRE_TECH_RESEARCHED_DETECTION
+The %empire% upgraded their detection technology to %rawtext:dstrength% [[encyclopedia DETECTION_TITLE]]
```
geoffthemedio: Rephrase more passive... The %empire% detection strength has upgraded to... or somesuch.

default/customizations/custom_sitreps.txt
geoffthemedio: It should be possible to write a macro in FOCS to check if a tech us researched and return a value dependent on that. I think there is an empire has tech researched condition which could be combined with the If ValueRef.

default/customizations/custom_sitreps.txt
```diff
@@ -104,6 +104,33 @@ EffectsGroup
             ]
             empire = Source.Owner
 
+EffectsGroup
+    scope = And [
+        Capital
+        Not OwnedBy empire = Source.Owner
+        Number low = 1 condition = And [
```
geoffthemedio: What does this Number condition do?

default/stringtables/en.txt
```diff
@@ -6572,6 +6572,12 @@ SITREP_TECH_RESEARCHED
 SITREP_TECH_RESEARCHED_LABEL
 Tech Researched
 
+SITREP_EMPIRE_TECH_RESEARCHED_DETECTION
+The %empire% [[encyclopedia DETECTION_TITLE]] has upgraded to %rawtext:dstrength%
+
+SITREP_EMPIRE_TECH_RESEARCHED_DETECTION_LABEL
+Enemy Tech Upgrade
```
alleryn: Should probably specify Enemy _Detection_ Tech Upgrade

default/customizations/custom_sitreps.txt
agrrr3: To every empire [but the researching empire].. (Capital Not OwnedBy ..) 
which can detect an object owned by the researching empire (Number low ...)
send a sitrep

But probably you usually know the detection strength of unsee empires as well. Does Olegs rule prevent knowing the detection strength?

default/stringtables/en.txt
agrrr3: The question here is if other things should have the same label (as it is used to filter sitreps).
I am not sure though if the label name or the label text is used for grouping/filtering (probably the label name).

alleryn: One way or the other, the label name should match the label text.  Either both should be particular to detection, or neither should.

default/customizations/custom_sitreps.txt
o01eg: @agrrr3 No, empire detection strength is known with rules about tech and statistics enabled.

default/stringtables/en.txt
agrrr3: If it was the label text the grouping could be the translators decision

default/customizations/custom_sitreps.txt
agrrr3: Ok then this is not strictly necessary - i will remove it for the time being. It might be though that one usually is only interested in empires in reach. But if you do not have a lot of empires in the game i dont think this will spam too much.

geoffthemedio: Please then add comment(s) saying the scope includes capital objects owned by empires other than the empire that researched the tech, if there is at least one object that is owned by the empire that researched the tech and the owner of the scope candidate can see.

default/customizations/custom_sitreps.txt
geoffthemedio: Decided to add a special case for TurnTechResearched...

https://github.com/freeorion/freeorion/commit/d4e5b76ff0fe450a7cfa0dad66f454cb9fc81580

PR #2574, opened 2019-09-27
Fix loading multiplayer game with eliminted empire

o01eg:
I've tried to load my saves with me as eliminated empire and found out I have to assign someone to my empire while it was eliminated.
This PR prevents eliminated empires from choosing in empires drop down list and fixes calculations of game limits.


geoffthemedio:
looks OK

PR #2579, opened 2019-09-29
Fix initial unlock SitReps

agrrr3:
Geoff partly fixed this while I was working on it simultaneously, this adds some finetuning of initial unlock sitreps.

As technology research was shifted after turn increase, SitReps for items unlocked by tech
are shown one turn after they are actually available (e.g. one turn after you can actually
enqueue Automated History Analyser).

This implementation fixes it by shifting the target turn for the unlock SitReps by minus one.
This makes a problem with the messages for initially unlocked items because those are registered
in minimal integer turn (e.g. -32767) which is 1+BEFORE_FIRST_TURN.

In order to fix this for turn zero or smaller I set the sitrep to a specific value.
If I set it to -1 , there are actually shown two different initial turns in the SitRep (one for
the real initial turn and one for -1)

This implementation fixes it by setting the sitrep turn to BEFORE_FIRST_TURN + 1

But usually for all that SitRep code, CurrentTurn()+1 is used so in future there might be a code
path for which the used CurrentTurn()+0 is wrong.

Also manipulating the turn number seems hackish.

So a better implementation could be to reverse the timing of the unlock by shifting it from
ApplyNewTechs to AddNewlyResearchedTechToGrantAtStartOfNextTurn.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1357,6 +1357,10 @@ void Empire::AddNewlyResearchedTechToGrantAtStartOfNextTurn(const std::string& n
     auto result = m_newly_researched_techs.insert(name);
     if (result.second)
         AddSitRepEntry(CreateTechResearchedSitRep(name));
+
+    // Unlock items now as the SitReps get generated already for CurrentTurn()+1
+    for (const ItemSpec& item : tech->UnlockedItems())
+        UnlockItem(item);  // potential infinite if a tech (in)directly unlocks itself?
 }
 
 void Empire::ApplyNewTechs() {
@@ -1367,9 +1371,8 @@ void Empire::ApplyNewTechs() {
             continue;
         }
 
-        for (const ItemSpec& item : tech->UnlockedItems())
-            UnlockItem(item);  // potential infinite if a tech (in)directly unlocks itself?
-
+        // Items are already unlocked
+        // Register the techs with CurrentTurn() as this gets called after turn increase
         if (!m_techs.count(new_tech))
             m_techs[new_tech] = CurrentTurn();
     }
--- a/util/SitRepEntry.cpp
+++ b/util/SitRepEntry.cpp
@@ -103,7 +103,7 @@ SitRepEntry CreateBuildingBuiltSitRep(int building_id, int planet_id) {
 SitRepEntry CreateTechUnlockedSitRep(const std::string& tech_name) {
     SitRepEntry sitrep(
         UserStringNop("SITREP_TECH_UNLOCKED"),
-        CurrentTurn(),
+        CurrentTurn() <= 0 ? BEFORE_FIRST_TURN + 1 : CurrentTurn(),
         "icons/sitrep/tech_unlocked.png",
         UserStringNop("SITREP_TECH_UNLOCKED_LABEL"), true);
     sitrep.AddVariable(VarText::TECH_TAG,          tech_name);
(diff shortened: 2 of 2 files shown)


geoffthemedio:
I don't understand why you set it to `BEFORE_FIRST_TURN + 1` and not just `BEFORE_FIRST_TURN` ...?

agrrr3:
I tried first using BEFORE_FIRST_TURN. But then the initial SitReps for the tech unlocks do not align with the initial buildings and part unlocks.

geoffthemedio:
Should probly make stuff be unlocked on `BEFORE_FIRST_TURN` then, rather than some other turn number calculated from it...

agrrr3:
> Should probly make stuff be unlocked on `BEFORE_FIRST_TURN` then, rather than some other turn number calculated from it...

Agreed and I already suggested:
> So a better implementation could be to reverse the timing of the unlock by shifting it from
ApplyNewTechs to AddNewlyResearchedTechToGrantAtStartOfNextTurn.

geoffthemedio:
Those are not the same thing, though... I'm suggesting to make initial stuff be unlocked on `BEFORE_FIRST_TURN` and not `BEFORE FIRST TURN + 1`, but still do the sitrep generation in the function that unlocks, not the function that marks stuff to be unlocked later. Nothing should be recorded as happening on `BEFORE FIRST TURN + 1`.

Empire/Empire.cpp
```diff
@@ -1357,6 +1357,10 @@ void Empire::AddNewlyResearchedTechToGrantAtStartOfNextTurn(const std::string& n
     auto result = m_newly_researched_techs.insert(name);
     if (result.second)
         AddSitRepEntry(CreateTechResearchedSitRep(name));
+
+    // Unlock items now as the SitReps get generated already for CurrentTurn()+1
+    for (const ItemSpec& item : tech->UnlockedItems())
```
geoffthemedio, line 1362: This partly undoes the change of moving tech adding to empires to after the turn number increment and effects, since those effects can depend on what stuff is unlocked. It is also strange to have separate paths for how to set when techs vs other stuff are unlocked.

agrrr3:
@geoffthemedio Oh, I didnt see that one comment of yours. Yes, probably registering initial techs and items on BEFORE_FIRST_TURN is the right thing to do. But the initialisation code is a bit fiddly and it probably would require introducing new code paths for generating SitReps with CurrentTurn vs CurrentTurn+1 (or given turn N) - I do not think I have the resources to get it right right now.

Empire/Empire.cpp
agrrr3: Well the root question is when do we want to have the unlocks and the effects available.

I think it is quite fine to have the effects start doing their effects in the turn after the "researched tech" message but the unlocks available in the turn of the  "researched tech" message.

The SitReps should align with that - so unlocks in Turn X and effects' SitReps (e.g. enemy detection increase) in Turn X+1.

agrrr3:
We could also do an ugly guard in all the SitRep generation functions to check if the given turn is BEFORE_FIRST_TURN (or given turn is less than zero) and do CurrentTurn() for that specific case, else CurrentTurn()+1.

Ugly but easy and would fit the bill, as you said
```Nothing should be recorded as happening on BEFORE FIRST TURN + 1.```

geoffthemedio:
https://github.com/freeorion/freeorion/commit/11a579785a9e1a067b12d5fb2c75c2cbdcf1608f fixes it for me

agrrr3:
> [11a5797](https://github.com/freeorion/freeorion/commit/11a579785a9e1a067b12d5fb2c75c2cbdcf1608f) fixes it for me

i will check :)
if it works, i am pretty sure all the items will register at  BEFORE FIRST TURN + 1

agrrr3:
Your commit is perfect. It seems the current state even fixed TurnTechResearched so it works in the turn one researches the tech. I'll need to adapt my other PR.

PR #2577, opened 2019-09-28
Use a working condition to see if fuel should be zeroed (see #2576)

agrrr3:
This is a fix - using initial MaxFuel value - and a workaround for the issues described in #2576

Issue #2576, opened 2019-09-28
Value() in scope condition is useless and misleading


agrrr3:
This issue was named 'Less-than operator for ValueRefs seems broken. Also Value() in scope condition.'
I found out the less-than operator is not broken and that the Value() does not return the current value is to be expected. 

So this issue is now what to do with Value() in scope.. conditions. 

The original text:
I found an unexpected issue with the latest commit which sets the max fuel to zero if the ship has 0.x fuel:
 The max fuel in the ship design window is always zero. 

I was thinking and concluded that I my effect was using the initial value but needed the current one.

So I tried to fix it by using 
`(Value(Source.MaxFuel) < 1)` instead of `MaxFuel high = 0.9999` in the condition.

I also tried `effects = If condition = (Value(Source.MaxFuel < 1)) effects = SetMaxFuel value = 0` instead of filtering in the scope condition.

But it seems Value(Source.MaxFuel) < 1 is always evaluating to true.

Trying further I found the following logically equivalent condition to be working:
`effects = If condition = Not (Value(Source.MaxFuel) >= 1.0) effects = SetMaxFuel value = 0`
as well as `effects = If condition = Not (Value(Source.MaxFuel) >= 1) effects = SetMaxFuel value = 0`

Using that in the scope condition did not work:
```
                //Not (Value(Source.MaxFuel) >= 1.0) // Always is true
                // Not (Value(Source.MaxFuel) >= 1) // Always is true
                // (Value(Source.MaxFuel) < 1) // Always is true
```

It seems the less-than operator is buggy!

And - I am not sure - usage of Value in the scope condition does not seem to work.
I guess it should (?)

geoffthemedio:
Is this an issue with less than or with `Value( ... )` wrapping?

I suspect that using `Value( ... )` wrappers in scope and activation conditions for a Max meter will always return 0. Max and Target meters are reset to 0 before effects are evaluated, so at the time of evaluating scope and activation conditions, they will be 0. Even if using priorities to order the execution, the scope and activation conditions are all evaluated in a separate earlier step. So probably don't do that.

The result of using `Value( ... )` wrapping within an effect will depend on when it's evaluated, relative to other effects that modify the meter.

agrrr3:
I had a look again at the C++ code and it is completely good and sane.
So i tried again with (Value(Source.MaxFuel) <1 ) in the If condition.... and it works.

Luckily I copy'n'pasted the non-working string here into the issue. 
[...]

PR #2580, opened 2019-09-29
Fix bug introduced by #2577

agrrr3:
MaxFuel in scope condition checks against the initial value which is always zero when calculating fuel for display in the encyclopedia. So this sets fuel always to zero. This condition commit was by mistake pushed to #2577 and merged as a fix for #2576 .

This PR fixes this by checking against the current value in effect evaluation.

### DIFF
--- a/default/scripting/ship_hulls/ship_hulls.macros
+++ b/default/scripting/ship_hulls/ship_hulls.macros
@@ -204,11 +204,12 @@ HULL_FUEL_EFFICIENCY_EFFECTSGROUP
             scope = And [
                 Ship
                 Source
-                MaxFuel high = 0.9999
+                // Issue #2567: using Not (Value(Source.MaxFuel) >= 1.0) here in the scope condition wrongly matches any object
             ]
             accountinglabel = "MAX_FUEL_LESS_THAN_ONE_LABEL"
             priority = [[METER_OVERRIDE_PRIORITY]]
-            effects = SetMaxFuel value = 0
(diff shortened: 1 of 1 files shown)


default/scripting/ship_hulls/ship_hulls.macros
```diff
@@ -204,11 +204,11 @@ HULL_FUEL_EFFICIENCY_EFFECTSGROUP
             scope = And [
                 Ship
                 Source
```
geoffthemedio, line 209: Put the `Source` condition first, or just leave out the `Ship` condition entirely. No sense checking all ships to see if they're the source when you could just start with the source only as the initial candidates set.

default/scripting/ship_hulls/ship_hulls.macros
```diff
@@ -202,13 +202,13 @@ HULL_FUEL_EFFICIENCY_EFFECTSGROUP
...
             scope = And [
-                Ship
                 Source
-                MaxFuel high = 0.9999
+                Ship
             ]
             accountinglabel = "MAX_FUEL_LESS_THAN_ONE_LABEL"
             priority = [[METER_OVERRIDE_PRIORITY]]
-            effects = SetMaxFuel value = 0
+            // Have to use Value() in effects, as all scope and activation conditions are evaluated before applying any effect
```
geoffthemedio: This comment doesn't make much sense outside the context of this pull request and related discussion...  Maybe write what you're doing first, then that that has to be done with a test / filter in the effect execution and not in the activation or scope because no effects have not yet been applied yet at the time the scope and activation are evaluated.

PR #2584, opened 2019-10-03
Don't clean empire readiness on player list changes

o01eg:
Fixes introduced in #2544 issue when player doesn't see ready status of other empires. Related thread https://freeorion.org/forum/viewtopic.php?p=97587#p97587.

### DIFF
--- a/UI/PlayerListWnd.cpp
+++ b/UI/PlayerListWnd.cpp
@@ -503,6 +503,9 @@ namespace {
         void            SetStatus(Message::PlayerStatus player_status)
         { m_player_status = player_status; }
 
+        Message::PlayerStatus Status() const
+        { return m_player_status; }
+
     private:
         int             IconSize() const   { return Value(Height()) - 2; }
 
@@ -657,6 +660,12 @@ namespace {
             return m_empire_id;
         }
 
+        Message::PlayerStatus Status() const {
+            if (m_panel)
+                return m_panel->Status();
+            return Message::PLAYING_TURN;
+        }
+
         void    Update() {
             if (m_panel)
                 m_panel->Update();
@@ -749,7 +758,7 @@ void PlayerListWnd::CompleteConstruction() {
         boost::bind(&PlayerListWnd::PlayerListWnd::HandleDiplomaticMessageChange, this, _1, _2));
     DoLayout();
 
-    Refresh();
+    Refresh(true);
 }
 
 std::set<int> PlayerListWnd::SelectedPlayerIDs() const {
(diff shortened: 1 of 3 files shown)


UI/PlayerListWnd.cpp
```diff
@@ -765,6 +774,23 @@ std::set<int> PlayerListWnd::SelectedPlayerIDs() const {
     return retval;
 }
 
+std::set<int> PlayerListWnd::ReadyEmpiresIDs() const {
```
geoffthemedio, line 777: comment explaining what this does and why?

UI/PlayerListWnd.cpp
```diff
@@ -818,8 +844,9 @@ void PlayerListWnd::Update() {
     }
 }
 
-void PlayerListWnd::Refresh() {
+void PlayerListWnd::Refresh(bool clean_empire_status) {
```
geoffthemedio, line 847: what's the purpose of the `clean_empire_status` parameter?

o01eg: On new game and new turn empire status must be cleared.

UI/PlayerListWnd.cpp
o01eg: Added comment.

UI/PlayerListWnd.cpp
geoffthemedio: Why not always?

o01eg: Because players list also updates when player enters/leaves the game. It shouldn't change empire status.

UI/PlayerListWnd.cpp
```diff
@@ -835,6 +862,8 @@ void PlayerListWnd::Refresh() {
     for (const auto& empire : Empires()) {
         int player_id = app->EmpirePlayerID(empire.first);
         auto player_row = GG::Wnd::Create<PlayerRow>(row_size.x, row_size.y, player_id, empire.first);
+        if (!clean_empire_status && (ready_empires.find(empire.first) != ready_empires.end()))
```
geoffthemedio, line 865: There are other statuses... why is this one handled differently?

o01eg: There only two statuses and WAITING means empire ready with its orders and PLAYING_TURN means otherwise.

UI/PlayerListWnd.cpp
geoffthemedio: Would it change the status if the player enters/leaves? Is a non-connected empire not considered waiting? Seems like an arbitrary special case...

UI/PlayerListWnd.cpp
geoffthemedio: Seems like the `PlayerRow` constructor or init should take or determine the status of players, and set itself internally when being constructed, rather than needing this separate set call outside...

o01eg: This value doesn't store anywhere except `PlayerRow`. Only if make it part of `Empire` class, but it would be a huge refactoring just before planned release.

UI/PlayerListWnd.cpp
o01eg: Currently if player enters game empire become unready because I think if player enters game, then he wants to change something, but it could be changed.
If player leaves game or empire is non-connected then status leaves unchanged. Moreover status is saved in the savegames end restored from them.
Only player or new turn can make empire status unready, and only player can make empire status ready.

UI/PlayerListWnd.cpp
geoffthemedio: If `PlayerRow` can get the info for itself given the player or empire ID, then that's fine. Point is it's weird to have it as a separate call to set the status after...

o01eg: Status can be changed without refreshing players list.

PR #2586, opened 2019-10-07
Made CMAKE_PREFIX_PATH configurable

mmoderau:
This enabled building outside SDK directory on Windows.

Kept the old hard-coded value as default.

---

Rationale:
When I was trying to build freeorion outside the SDK directory (even outside freeorion source directory, but that's beside the point), almost everything worked by supplying correct paths through CMake gui.
The only thing that didn't work was copying various dependencies, like python27.dll, because they were searched for in paths like: `${CMAKE_PREFIX_PATH}/bin/python27.dll`

`${CMAKE_PREFIX_PATH}` was hard-coded as `"${CMAKE_SOURCE_DIR}/.."`, which made the assumption, that freeorion is checked out inside SDK directory.

Some people like to keep sources and build artifacts in separate directory hierarchies. For such people being able to *customize* the `CMAKE_PREFIX_PATH` will be quite useful.

For others, this MR keeps the default value.

If discussion on forum is needed, pleasue use [this topic](https://www.freeorion.org/forum/viewtopic.php?f=24&t=11262).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,7 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 
 if(WIN32)
     # Location of the FreeOrionSDK root
-    set(CMAKE_PREFIX_PATH "${CMAKE_SOURCE_DIR}/..")
+    set(CMAKE_PREFIX_PATH "${CMAKE_SOURCE_DIR}/.." CACHE PATH "Location of the FreeOrionSDK root")
     # Search dependency DLLs inside FreeOrionSDK
     target_dependencies_add_search_path("${CMAKE_PREFIX_PATH}/bin")
     set(CMAKE_INSTALL_BINDIR ".")


mmoderau:
OK, the build on AppVeyor failed because:

> Build execution time has reached the maximum allowed time for your plan (60 minutes).

Does that happen often?

geoffthemedio:
Similar issues happen fairly often with the automated builds, yes.

eg. after a quick look at recent commits, https://ci.appveyor.com/project/freeorion/freeorion/builds/27881819

geoffthemedio:
Unless @adrianbroher would be kind enough to comment, or @Vezzra has any experience with the CMake builds, I don't have any basis to review this...

ghost:
> Unless @adrianbroher would be kind enough to comment

Yes, here. Replying under my job account.

The hardcoded value reflects the setup used by the Visual Studio projects provided within the freeorion/freeorion repository. I decided intentionally against allowing a more flexible setup as it would require additional effort to test against both cmake and VS project files.  As @geoffthemedio is unwilling to drop support of the VS projects within freeorion/freeorion and switch to cmake for all platforms those kind of workarounds, platform specific conditionals and hardcoded values are the consequence.

mmoderau:
@marcelmetz , I think I understand, what you mean. Still, this PR does not change much for the default setup. It adds a new configurable path, but the default value is the same as the old hard-coded.
The configurable path is displayed *only* on Win32 builds.

Are you wary of adding additional degrees of freedom, because probably someone somewhere will change this value and it will blow up and someone will have to support it? Is that it?

This PR enabled me building FO outside SDK, but if I need to keep the change to CMakeList.txt in git stash instead and apply anytime I need to run cmake again... well, I can live with it, I guess.

geoffthemedio:
To be clear, as long as CMake generates MSVC project files that work more or less the same as the current hand-maintained ones, and then those can be regularly committed and updated after CMake changes, it would be fine. The main issue is that it shouldn't be required for someone using MSVC to also have CMake to generate their project files, but rather they should (unless the CMake files were just updated) be able to directly load the MSVC project files and build.

ghost:
@macmodrov 

> Are you wary of adding additional degrees of freedom, because probably someone somewhere will change this value and it will blow up and someone will have to support it? Is that it?

I'm not associated with this project anymore. Feel free to break it every way you seem to be fit.

mmoderau:
To be fair, this is this is my first time doing anything with a CMake set up, so review would be a good idea.

On the other hand, what I did is I promoted a variable that was already in the CMakeList.txt to be cached in CMakeCache.txt. This way, if a developer *wants* to change it for their setup, they can. For all others they will only see another entry in cmake-gui with already a working value provided.

Also, again, this only affects Win32 CMake users, as for linux and macos this variable will not appear.

geoffthemedio:
@Vezzra do you use Windows CMake for weekly builds? If not then I'm not as concerned about merging this... if it is a problem / bad style or whatever standard practice is for CMake stuff, then it can be reverted...

Vezzra:
@geoffthemedio, no, I don't use CMake on Windows, just the manually maintained MSVC projects.

My proficiency with CMake  is, well, quite limited, but the change looks simple enough, I don't think it very likely that it will break anything. And as you said, it could be easily reverted anyway.

So I'd say go ahead and merge...

PR #2590, opened 2019-10-12
Fix handling for gifting objects.

swaqvalley:
This fixes failed planet/fleet gifting and also crashing when cancelling a fleet gift that I found during testing.


alleryn:
I helped with testing a little and in case it's useful i'll just leave the relevant commits here when gifting last worked and was broken:
Working in 400086437
Broken in subsequent commit d90855027

PR #2591, opened 2019-10-13
Add macros for easier adjustment of native detection, defense, shield…

alleryn:
…s, and troops

There are no adjustments to actual mechanics/numbers in this branch, just added functionality for future changes.

Any discussion can be addressed here or in the forum thread: https://www.freeorion.org/forum/viewtopic.php?f=15&t=11355


default/scripting/species/common/native_fortification.macros
```diff
@@ -0,0 +1,33 @@
...
+NATIVE_FORTIFICATION_MEDIUM
+'''[[NATIVE_FORTIFICATION_CUSTOM(20,20,20,20)]]'''
+
+//currently unused
+NATIVE_FORTIFICATION_LOW
+'''[[NATIVE_FORTIFICATION_CUSTOM(10,10,10,10)]]'''
+
+//currently unused
+NATIVE_FORTIFICATION_MINIMAL
+'''[[NATIVE_PLANETARY_SHIELDS(1)]]'''
```
geoffthemedio, line 28: 1 troops would seem more sensible than 2 shield as the minimum...

alleryn: Well, native colonies all (currently, afaik) already have troops.  They have a default shield value of 0.

Adding a shield value of 1 would force the player to bring a combat ship to the planet they want to invade, instead of being able to use only troop ships.

That is my thinking at least.

(Note that the macros i implement here are increments not "set to", so it isn't as though any of them would reduce the troops to zero or anything like that).

Edit: Those 'presets' (Fortification_high, Fortifcation_minimal, etc) are intended as examples/placeholders anyways (and are currently unused).  I expect it to take some testing to find appropriate values. /Edit

Edit2: Is your point that a shield without any troops does nothing to oppose invasion while a troop without a shield has some 'stopping power'? /Edit2

geoffthemedio: more so that a minimally defensive planet wouldn't likely have a shield set up, but it would sensibly have a token garrison of troops.

alleryn: > more so that a minimally defensive planet wouldn't likely have a shield set up, but it would sensibly have a token garrison of troops.

I think this makes sense in terms of 'fluff', but it's not clear to me that adding a single troop has any qualitative effect on how difficult it is to penetrate a planet's defenses (at most it's one more troop ship to bring along).

I would expect this case to be covered by simply omitting any Native_Fortification (or implementing an empty macro NATIVE_FORTIFICATION_NONE), and relying on the population's 'default' troops (i.e. those provided by [[BASIC_DEFENSE_TROOPS]] via [[GREAT_DEFENSE_TROOPS]] / [[AVERAGE_DEFENSE_TROOPS]] / etc).

(Aside: Fluffwise i envision these Native_Fortification defenses as factions loyal to their homeworld but who lack interest in planetary defense after being subsumed by an empire.  This explains why the bonus disappears when the planet becomes "owned".)

geoffthemedio: In that case, I'd add a weak attack rather than weak shields.

alleryn: > In that case, I'd add a weak attack rather than weak shields.

Again, i don't think it's clear that really would do anything significant to oppose invasion.  A planet with an attack value and no shield can be directly invaded (InvadeOrder::Check in util/Order.cpp checks only that planet's shield value is zero, not planet's attack value), so (presuming it's a small attack value of say, 1) the invasion force could still consist of only troop ships.

Also note that the shield value of 1 is what all empire planets have as default defense (from the starting tech DEF_ROOT_DEFENSE in https://github.com/freeorion/freeorion/blob/master/default/scripting/techs/defense/Defense.focs.txt ).

Vezzra:
As this is intended as prep work for later adjustments, I've assigned it to the "post release" milestone. Just in case it breaks anything, better postpone merging it after the release branch has been created.

Issue #2551, opened 2019-09-14
Ask for confirmation to leave without sending orders in the multiplayer game


o01eg:
### Environment

* **FreeOrion Version**: master
* **Fetched as** 
  * Binary release
  * Weekly development build
  * Compiled from source

### Description
In the multiplayer game the player could forgot to send his orders before close client: https://www.freeorion.org/forum/viewtopic.php?p=97201#p97201
I suppose there should be some confirmation to ensure player want to disconnect from the game without set his empire to ready.

### Expected Result
In the multiplayer game show confirmation dialog to ask if the player want to leave the game if his empire is unready.

geoffthemedio:
This is a would-be-nice but not a requirement for v0.4.9, I think.

Vezzra:
Ok

o01eg:
It looks like it was fixed in #2594

PR #2594, opened 2019-10-14
Issue #2551 changes

swaqvalley:
For Issue #2551 created by @o01eg. These changes prompt the user for confirmation when leaving a multiplayer game without setting their status to ready.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -2579,6 +2579,9 @@ The Save Game button is disabled, either because this is a multiplayer game in w
 GAME_MENU_REALLY_CONCEDE
 Confirm conceding? This empire will be removed from the game.
 
+GAME_MENU_CONFIRM_NOT_READY
+Confirm quit? Orders not finalized.
+
 
 ##
 ## Save game dialog
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -1436,13 +1436,31 @@ void HumanClientApp::ExitSDL(int exit_code)
 { SDLGUI::ExitApp(exit_code); }
 
 void HumanClientApp::ResetOrExitApp(bool reset, bool skip_savegame, int exit_code /* = 0*/) {
+    if (m_exit_handled) return;
+    m_exit_handled = true;
     DebugLogger() << (reset ? "HumanClientApp::ResetToIntro" : "HumanClientApp::ExitApp");
 
     auto was_playing = m_game_started;
     m_game_started = false;
 
-    // Only save if not exiting due to an error.
+    // Only save or allow user to cancel if not exiting due to an error.
     if (!skip_savegame) {
+        // Check if this is a multiplayer game and the player has not set status to ready
+        if (was_playing && !m_single_player_game && m_empire_status[m_empire_id] != Message::WAITING) {
+            std::shared_ptr<GG::Font> font = ClientUI::GetFont();
+            auto prompt = GG::GUI::GetGUI()->GetStyleFactory()->NewThreeButtonDlg(
+                GG::X(275), GG::Y(75), UserString("GAME_MENU_CONFIRM_NOT_READY"), font,
+                ClientUI::CtrlColor(), ClientUI::CtrlBorderColor(), ClientUI::CtrlColor(), ClientUI::TextColor(),
+                2, UserString("YES"), UserString("CANCEL"));
+            prompt->Run();
+            if (prompt->Result() != 0) {
+                // User aborted exit/resign, reset variables
+                m_game_started = was_playing;
+                m_exit_handled = false;
+                return;
+            }
+        }
+
         if (was_playing && GetOptionsDB().Get<bool>("save.auto.exit.enabled"))
             Autosave();
 
@@ -1480,6 +1498,8 @@ void HumanClientApp::ResetOrExitApp(bool reset, bool skip_savegame, int exit_cod
         after_server_shutdown_action = std::bind(&HumanClientApp::ExitSDL, this, exit_code);
 
     m_fsm->process_event(StartQuittingGame(m_server_process, std::move(after_server_shutdown_action)));
+    
+    m_exit_handled = false;
 }
 
 void HumanClientApp::InitAutoTurns(int auto_turns) {
--- a/client/human/HumanClientApp.h
+++ b/client/human/HumanClientApp.h
@@ -180,6 +180,7 @@ private:
 
     bool m_single_player_game = true;   ///< true when this game is a single-player game
     bool m_game_started = false;        ///< true when a game is currently in progress
+    bool m_exit_handled = false;        ///< true when the exit logic is already being handled
     bool m_connected = false;           ///< true if we are in a state in which we are supposed to be connected to the server
     int  m_auto_turns = 0;              ///< auto turn counter
     bool m_have_window_focus = true;


alleryn:
You should probably try to squash/fixup those commits (see https://help.github.com/en/articles/using-git-rebase-on-the-command-line).

That should work at least for the bottom two commits.  Something like:
`git rebase -i HEAD~3`
```
pick 6ae7457 update
pick 3981e1c Prompt user when leaving multiplayer game without setting status to ready.
fixup 49b18145c
```
That first update shouldn't really be in there.  I'm not totally sure how to get rid of that. 
@dbenage-cx could you help us with this?

swaqvalley:
Yeah, I'm still trying to figure out GitHub. I was wondering if I was supposed to delete my fork and recreate it to avoid the merge commit.

swaqvalley:
I was able to combine the two commits with actual changes. No idea how to go about removing the merge commit. If it's a problem I can start over my fork from scratch and do a new pull request.

geoffthemedio:
Rebase you branch on master and then force-push.

swaqvalley:
I tried that but now it says "This branch is 14 commits ahead, 17 commits behind freeorion:master."

swaqvalley:
I think I did it! I followed an example for how to force a fork to be even with the upstream then rebased the branch and it appears to just have the one commit now.

default/stringtables/en.txt
```diff
@@ -2579,6 +2579,9 @@ The Save Game button is disabled, either because this is a multiplayer game in w
 GAME_MENU_REALLY_CONCEDE
 Confirm conceding? This empire will be removed from the game.
 
+GAME_MENU_CONFIRM_NOT_READY
+Confirm quit in unready state?
+
```
alleryn: I would maybe suggest rewording this.  "Unready" seems a little ambiguous to me.

How would you feel about "Confirm quit?  Orders are not finalized."  Or something like that?

swaqvalley: As far as I am aware you don't need to click the turn button to finalize or submit your orders, that happens automatically. So that make the purpose of clicking the button to indicate to the server that you are ready for the turn to advance so that it can proceed when all the players are ready without waiting for the timer to expire.

That said I don't think my wording is ideal, I was just trying to be concise. The name of that feature is not really given in the game. Maybe the closest thing is "waiting" which is used in the code for the enum and the text in the messages box "Waiting for Other Players".

Some other possible options:
"Confirm quit? You are not waiting."
"Confirm quit? You are not waiting for turn advance."
"Confirm quit? Waiting status not set."
"Confirm quit? Ready status not set."
"Confirm quit without setting ready status?"

alleryn: Oh, i thought you did need to click the turn button for your orders to process, but that was just an assumption on my part.  My mistake.

Maybe just something to indicate what the player isn't ready for?
"Confirm quit?  Still revising orders."
"Confirm quit?  Turn advance not set."
"Confirm quit?  Still revising orders, turn advance not set."
"Confirm quit without setting readiness for next turn?"

swaqvalley: I like "Confirm quit? Still revising orders." That matches up with how the button changes when you click it which would help indicate what the user needs to do.

alleryn: I'm happy with that.

swaqvalley: String has been updated.

default/stringtables/en.txt
```diff
@@ -2579,6 +2579,9 @@ The Save Game button is disabled, either because this is a multiplayer game in w
 GAME_MENU_REALLY_CONCEDE
 Confirm conceding? This empire will be removed from the game.
 
+GAME_MENU_CONFIRM_NOT_READY
+Confirm quit? Still revising orders.
```
geoffthemedio: This message is also displayed if orders have never been sent for the current turn, not just if they were sent and then the revise orders button was clicked.

swaqvalley: True. Would you prefer different wording?

geoffthemedio: Ideally it would check whether orders have ever been sent for the turn currently being played, or if there are orders issued but not sent since the revise orders button was clicked, or of no orders have ever been sent for the current turn, or if orders were sent and revise clicked but no no additional orders were then issued, and adjust the message accordingly. Alternatively, in a multiplayer game it could say that the server is waiting for orders, or in single player it could say orders have not yet been sent. Or in all cases it could say orders have not been sent... even though if revising then orders were sent (and just the revisions haven't been sent)...

alleryn: This kind of brings me back to my original suggestion:
"Confirm quit?  Orders have not been finalized."
This appears to cover both the case in which orders have never been sent for the current turn, and the case in which orders were sent but are currently in the state of revision.

The objection was that you do not need to click the End Turn button to finalize or submit your orders.  But after thinking it through, while your orders may finalize and submit if you allow the timer to expire, it is still accurate to say that the orders have not been _finalized_ if they are currently in the unsubmitted state or the revising state.  (I think)

This does not cover this case:

>  in single player it could say orders have not yet been sent

but i think that is outside the scope of this PR and related issue, which pertain only to multiplayer.  I'm not sure a confirmation is really needed for singleplayer,  where exiting without advancing the turn is (i think) more "standard" behavior than it is in a situation with multiplayer turns where the player has periods of "forced hiatus".

swaqvalley: The message will not be displayed in a single player game.

I would prefer a single wording for the dialog than trying to check whether the user has submitted then cancelled to start revising. It would get extra confusing when a user logs back in to the multiplayer server and it automatically starts revising. So a consistent message regardless of that state would be easier and probably better for the user to understand.

I am also fine with alleryn's suggestion "Confirm quit? Orders have not been finalized." Or if we wanted a more concise version: "Confirm quit? Orders not finalized."

Thoughts?

geoffthemedio: > The message will not be displayed in a single player game.

Are you sure / is that true now? I can't test immediately but I think I got the message when reseigning or closing the window in a single player game...

alleryn: > > The message will not be displayed in a single player game.
> 
> Are you sure / is that true now? I can't test immediately but I think I got the message when reseigning or closing the window in a single player game...

I'm unable to reproduce this.  I only get the dialog box when i try to exit a multiplayer game.  Exiting a singleplayer game is as before.

I think the && !m_single_player_game is intended to prevent this.

While testing i did produce an error however; i'll comment below.

alleryn:
I found an error while testing:

Reproduce:
* Start client.  Host multiplayer game.
* Start second instance of client.  Join multiplayer game at localhost.
* Accept and start game in the clients.
* In first client, Resign and confirm quit.
* This causes the second client to receive message no longer connected to server.  (Should probably exit to main menu at this point).
* Second client also receives confirm quit orders not ready dialog.
* Cancel quit in second client.
* Second client receives connection to server lost.
* Resign and confirm quit in second client.
* Second client receives 'save game in progress' and hangs.

The first time i found this bug, the client completely hung.  Following these steps, i can still press the quit to main menu without saving from the save dialog.  I'm trying to reproduce the worse crash.

Edit: I've been unable to reproduce the complete hang, but the bug is still there: the client should not remain open after it disconnects from the server.  (At least i think that is desired behavior).

swaqvalley:
Thanks for testing, alleryn. I was doing all my testing with a persistent server so I didn't run into this issue.

I can reproduce the save game in progress hang by reversing which player quits first. If the player hosting the game quits second they get the save window. But I found that it is a soft hang, as cancelling the save drops back to the main menu.

What is the preferred behavior here? Just kick the rest of the players out to the main menu whenever one person quits?

default/stringtables/en.txt
swaqvalley: Geoff, I have never had the message come up for a single player game in my tests. As alleryn stated, I explicitly check that it is not a single player game before showing the dialog box.

swaqvalley:
Okay, it looks like I can use the skip_savegame flag to determine when the reason for reset/exit is because of an error like server disconnect.

default/stringtables/en.txt
geoffthemedio: Possibly I tested with a nominally multiplayer game with only one human player, which does produce the popup.

But I think the multiplayer/singleplayer distinction isn't important really...

Regardless, "Confirm quit? Orders not finalized." seems OK if you don't want to adjust for the exactly turn-send-revise-resend state.

geoffthemedio:
I think ideally it would do an autosave whenever quitting, without expecting save info from not-connected players, but that's probably beyond the scope of this pull request...

swaqvalley:
I agree that changing the behavior of the autosave is outside of the scope for this PR.

I believe the commit 019f24f addresses/matches the latest discussion here. Is there anything you are waiting on me for?

geoffthemedio:
This also shows the popup for observers... but probably shouldn't.

swaqvalley:
Ah, that is definitely an oversight. That shouldn't be too hard to fix. I assume I should make a new pull request to handle that?

geoffthemedio:
yes

Issue #2541, opened 2019-08-29
Pedia gives incorrect information for fuel of a hull


Morlic-fo:
Since #2492, the pedia description of hulls is incorrect as far as fuel is concerned. 

1) For example the [basic small hull](https://github.com/freeorion/freeorion/blob/master/default/scripting/ship_hulls/SH_BASIC_SMALL.focs.txt) has base fuel 2. That is also the value which appears in the pedia. Effectively, however, the hull has 8 fuel due to the ```[[GREAT_FUEL_EFFICIENCY]]``` effect.

2) The fuel efficiency of a hull is never mentioned in the pedia.

This makes it impossible to figure out how much fuel the hull actually has (or how much adding a fuel pod will help) from the pedia.

Expected result:
Show fuel efficiency for all hulls with the corresponding scaling factor.

Probably preferably, the fuel after fuel efficiency applied is presented to the player. The alternative, showing base fuel only and then letting the player do the calculation seems rather clumsy. In that case, there should be rewording or an additional sentence to make clear that the base fuel of the hull will be multiplied by the fuel efficiency factor.

agrrr3:
Already started discussing documentation issues and solutions in [Forum thread](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11321#p96865)

alleryn:
@Morlic-fo I'm attempting to address some of your concerns via https://github.com/freeorion/freeorion/pull/2596.  This would correct the fuel value in the Pedia for the hulls, but won't explicitly state the Fuel Efficiency for each hull.  (My understanding is there are concerns over accessing the FOCS files from the C++ code in order to do this -- see https://www.freeorion.org/forum/viewtopic.php?p=97692#p97692)

Thoughts on this approach?

geoffthemedio:
> won't explicitly state the Fuel Efficiency for each hull.

This can be added to the hull's description text in the stringtable.

Vezzra:
Fixed by #2621.

PR #2596, opened 2019-10-15
Adjust fuel values and efficiency formula so hull info is correct in …

alleryn:
…pedia

### DIFF
--- a/default/scripting/ship_hulls/SH_BASIC_MEDIUM.focs.txt
+++ b/default/scripting/ship_hulls/SH_BASIC_MEDIUM.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_BASIC_MEDIUM"
     description = "SH_BASIC_MEDIUM_DESC"
     speed = 60
-    fuel = 1.5
+    fuel = 3
     stealth = 5
     structure = 10
     slots = [
--- a/default/scripting/ship_hulls/SH_BASIC_SMALL.focs.txt
+++ b/default/scripting/ship_hulls/SH_BASIC_SMALL.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_BASIC_SMALL"
     description = "SH_BASIC_SMALL_DESC"
     speed = 75
-    fuel = 2
+    fuel = 8
     stealth = 5
     structure = 5
     slots = [
--- a/default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_HEAVY_ASTEROID.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_HEAVY_ASTEROID"
     description = "SH_HEAVY_ASTEROID_DESC"
     speed = 60
-    fuel = 2
+    fuel = 1.2
     stealth = 5
     structure = 50
     slots = [
--- a/default/scripting/ship_hulls/asteroid/SH_MINIASTEROID_SWARM.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_MINIASTEROID_SWARM.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_MINIASTEROID_SWARM"
     description = "SH_MINIASTEROID_SWARM_DESC"
     speed = 60
-    fuel = 2
+    fuel = 4
     stealth = 35
     structure = 11
     slots = [
--- a/default/scripting/ship_hulls/asteroid/SH_SCATTERED_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_SCATTERED_ASTEROID.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_SCATTERED_ASTEROID"
     description = "SH_SCATTERED_ASTEROID_DESC"
     speed = 60
-    fuel = 2
+    fuel = 1.2
     stealth = 5
     structure = 140
     slots = [
--- a/default/scripting/ship_hulls/asteroid/SH_SMALL_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_SMALL_ASTEROID.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_SMALL_ASTEROID"
     description = "SH_SMALL_ASTEROID_DESC"
     speed = 60
-    fuel = 1
+    fuel = 2
     stealth = 5
     structure = 15
     slots = [
--- a/default/scripting/ship_hulls/asteroid/SH_SMALL_CAMOUFLAGE_ASTEROID.focs.txt
+++ b/default/scripting/ship_hulls/asteroid/SH_SMALL_CAMOUFLAGE_ASTEROID.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_SMALL_CAMOUFLAGE_ASTEROID"
     description = "SH_SMALL_CAMOUFLAGE_ASTEROID_DESC"
     speed = 60
-    fuel = 2
+    fuel = 4
     stealth = 35
     structure = 15
     slots = [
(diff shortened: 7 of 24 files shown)


alleryn:
Hopefully fixes https://github.com/freeorion/freeorion/issues/2541

agrrr3:
I see that you updated 22 hull files. Which fits the expected count (3+13+6). 
Code looks also good to me :) Let's merge ;)

```
for f in GREAT GOOD AVERAGE BAD; do echo $f ; grep --exclude ship_hulls.macros -e "tags.*${f}_FUEL" -r default/scripting/ship_hulls/ -l  | wc -l; done
GREAT
3
GOOD
13
AVERAGE
11
BAD
6
```

agrrr3:
Not sure if pedia should be updated somewhere to explain this(?)

alleryn:
> Not sure if pedia should be updated somewhere to explain this(?)

What is 'this'?

alleryn:
Should i try changing the fuel values to something like, e.g. scout with 2 base fuel and great = 4* efficiency:

BaseFuel = 2
fuel = BaseFuel * [(If HasTag Great_Effic) * 4 + (If HasTag Good_Effic)*2 + ...]

?

Just a thought.  Probably not really useful, and not sure if it would cause any problems.

alleryn:
> > Should i try changing the fuel values...
> 
> What would be the reason for that? Looks like you want to test for all tags instead of just the one the hull is expected to have? Why?

My intention was to improve readability.  RIght now the fuel efficiency is masked in the value of fuel ascribed to the hull, and the "Base Fuel" for that hull is never defined.  

I skipped a step above but i was thinking something like

BaseFuel = 2
Fuel = BaseFuel*[[Fuel Efficiency]]
Fuel Efficiency = '''(If HasTag Great_Effic) * 4 + (If HasTag Good_Effic)*2 + ... '''

I don't know if there's any real reason to retain that BaseFuel value (there isn't any reason now... it would be 'futureproofing'), but that was my thought process.

> 
> One change I would test is replacing `Object id = Source.ID` with just `Source`.

I can test that ~~later today~~ in the near future.
 
> And I didn't really understand the purpose of the `HasTag name = "@1@_FUEL_EFFICIENCY"` other than it was a safety check for scripting errors, but could that condition be put into the activation condition of the effectsgroup instead?

@agrrr3 ?

geoffthemedio:
> I don't know if there's any real reason to retain that BaseFuel value (there isn't any reason now... it would be 'futureproofing')

Seems unnecessary

agrrr3:
> > Not sure if pedia should be updated somewhere to explain this(?)
> 
> What is 'this'?

This is: Change to how fuel / fuel efficiency works.

I grepped myself and both FUEL_REFUEL_EFFICIENCY_TEXT and METER_FUEL_VALUE_DESC refer to base fuel which gets scaled which is not true anymore.

agrrr3:
> > And I didn't really understand the purpose of the `HasTag name = "@1@_FUEL_EFFICIENCY"` other than it was a safety check for scripting errors

You understood the purpose. Only the emphasis on hard-to-catch scripting errors is missing in your description. 


> > , but could that condition be put into the activation condition of the effectsgroup instead?
>
> @agrrr3 ?

As with this PR base fuel is gone, the check is not really helpful anymore. Without base fuel modified by efficiency you only see the discrepancy if you have additional fuel effects and care about those.

So I would rather put it in the activation condition of another effect writing a sitrep. The feature I am trying to emulate is a fail-fast assert which checks the condition after loading the FOCS files and stops the game from starting - but we do not have such a feature.

agrrr3:
> As with this PR base fuel is gone, the check is not really helpful anymore. Without base fuel modified by efficiency you only see the discrepancy if you have additional fuel effects and care about those.

I read the macro again and I was wrong - the fuel should still be set to zero if right tag is missing. So it can stay like it is (besides the simplifying the source id check) and suffices IMHO

> > > , but could that condition be put into the activation condition of the effectsgroup instead?

No. If that check was in the activation condition max fuel would not be set to zero but would stay at hull fuel level and we would miss the error. 

It could be put in the activation condition of another effect but I guess that would be overkill.

agrrr3:
Change in METER_FUEL_VALUE_DESC
```The effective amount of fuel a <encyclopedia ENC_SHIP_DESIGN>Ship Design</encyclopedia> is scaled by the [[encyclopedia FUEL_EFFICIENCY_TITLE]] of the [[encyclopedia ENC_SHIP_HULL]]. Fuel efficiency modifies the base fuel capacity of a <encyclopedia ENC_SHIP_DESIGN>Ship Design</encyclopedia> which depends on the [[encyclopedia ENC_SHIP_HULL]] used, but it can also be increased using some [[encyclopedia ENC_SHIP_PART]]s.```
to 
```The effective amount of fuel a <encyclopedia ENC_SHIP_DESIGN>Ship Design</encyclopedia> depends on the base fuel of the hull and additional fuel effects scaled by the [[encyclopedia FUEL_EFFICIENCY_TITLE]] of the [[encyclopedia ENC_SHIP_HULL]]. Fuel efficiency modifies the effective fuel capacity of fuel effects for example extra fuel given by [[encyclopedia ENC_SHIP_PART]]s.```

geoffthemedio:
>  Fuel efficiency modifies the effective fuel capacity of fuel effects for example extra fuel given by [[encyclopedia ENC_SHIP_PART]]s.

Fuel efficiency changes the effective capacity of fuel [[encyclopedia ENC_SHIP_PART]]s and will multiply any other fuel-modifying effects.

alleryn:
> One change I would test is replacing `Object id = Source.ID` with just `Source`.

Seemed to work fine.

alleryn:
> I grepped myself and both FUEL_REFUEL_EFFICIENCY_TEXT and METER_FUEL_VALUE_DESC refer to base fuel which gets scaled which is not true anymore.

Added agrrr3 and geoffthemedio's changes to METER_FUEL_VALUE_DESC and edited FUEL_REFUEL_EFFICIENCY_TEXT.

alleryn:
Still to do (at least):

- 
> And I didn't really understand the purpose of the `HasTag name = "@1@_FUEL_EFFICIENCY"` other than it was a safety check for scripting errors, but could that condition be put into the activation condition of the effectsgroup instead?

There was discussion here: https://www.freeorion.org/forum/viewtopic.php?p=97786#p97786 but it's unclear to me where it landed.

- Add Fuel Efficiency to stringtable for every hull

Morlic-fo:
This will also need AI adjustments:
https://github.com/freeorion/freeorion/blob/master/default/python/AI/ShipDesignAI.py#L1037

agrrr3:
> This will also need AI adjustments:
> https://github.com/freeorion/freeorion/blob/master/default/python/AI/ShipDesignAI.py#L1037

Should be something like changing `self.design_stats.fuel *= self._hull_fuel_efficiency()`
to `self.design_stats.fuel = (self.design_stats.fuel - self.hull.fuel ) * self._hull_fuel_efficiency() + self.hull.fuel`

agrrr3:
> `HasTag name = "@1@_FUEL_EFFICIENCY"`

Looks fine to me

>     * Add Fuel Efficiency to stringtable for every hull

just adjust the string tables entries of all hulls accordingly. I think it would be good for players to have the fuel efficiency information in 0.4.9 and I do not see a more generic solution coming up in the next week(s)

alleryn:
> just adjust the string tables entries of all hulls accordingly. I think it would be good for players to have the fuel efficiency information in 0.4.9 and I do not see a more generic solution coming up in the next week(s)

It's unlikely i'll have time to do this in the next month.

Do the other changes look acceptable?  Perhaps i should rebase and we can merge, and someone else can pick up the rest?

agrrr3:
It would be great if you could apply and test the AI change.

I think the current changes look good and I consider it an acceptable risk that the hull descriptions would be not perfect in 0.4.9

alleryn:
> It would be great if you could apply and test the AI change.

I don't foresee having the time to do this in at least the next two weeks (could be longer).

agrrr3:
Note this PR also mitigates a problem I just found. A negative fuel bonus scales the wrong way (being better means less max fuel). I just thought we do not have negative fuel bonus, but forgot species BAD_FUEL is there (which for an great efficiency ship should lead to -1 max fuel). 
For Scylior without the PR this should lead to - with the PR it is an acceptable -2 fuel (so 6 instead of 8)
I will open a separate issue for fixing BAD_FUEL with fuel efficiency.

agrrr3:
I put the AI changes into #2621 and tested. I suggest merging and doing a followup PR for hull info.

Note if the hull info is done manually it should probably happen after fixing #2620

geoffthemedio:
Superceded by #2621

PR #2597, opened 2019-10-16
Add more Diplomacy articles to Pedia

alleryn:
This PR also cleans up the System Blockade article, formatting it into paragraphs and removing some superfluous (imo) information that wasn't specific to Blockades.  (Specifically the following lines:

> Also, if the blockaded empire posesses colonies within the system, the production generated by these colonies can only be used on the planet it is generated on. If a combat is initiated concomitantly, effects that only work if no combat has occurred, such as [[encyclopedia ORBITAL_DRYDOCK_REPAIR_TITLE]] and [[metertype METER_TROOPS]] meter growth, are suspended for that turn.

)

The blockaded colony production limitation is a function of the breaking of supply propagation (which is already discussed in the article).

The limits to repair, troop growth are a function of combat and unrelated to the blockade.


alleryn:
@TheSilentOne1 Looks like you originally wrote the System Blockade Pedia article https://github.com/freeorion/freeorion/commit/1f7d9fe6a9025871fd8c3d80b69870d186888ce4

Are you okay with these changes to it?

TheSilentOne1:
Yes.

PR #2600, opened 2019-10-21
Apply newly added techs before final meter estimation

agrrr3:
Move ApplyNewTechs call Before Final Meter Estimate Update in PostCombatProcessTurns in order to do correct predictions for next turn values on server side.

[Forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11388&p=97773#p97773) 

Test target meter estimation and it looks good for empire and planetary output.
When i play humans i have 9 RP for all, i research first Algorithmic Elegance and get the "tech researched" message on turn 4. Planetary values are: currently 9RP, target 11RP and next turn 10RP. Empire values are 10RP used/output and 11RP target. Then the values grow as expected each turn until 11RP.

By moving it up we also move it before the following calls
    UpdateEmpireSupply(true);
    m_universe.UpdateEmpireLatestKnownObjectsAndVisibilityTurns();
    m_universe.UpdateStatRecords();
    for (auto& entry : empires)
        entry.second->UpdateOwnedObjectCounters();
    GetSpeciesManager().UpdatePopulationCounter();

Reasoning that there are no unintended side-effects by this move:
  supply update is meter based - OK (meter backpropagation already happened)
  visibility is based on stealth and detection meters - OK
  stat records - probably OK; FOCS statistics are evaluated here those should not change game state. Number of technologies would include newly added techs (which is ok)
  update object counters - should be OK, tech does not directly add/remove/change objects
  population counters - should be OK, tech does not directly add/remove population

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -3548,8 +3548,18 @@ void ServerApp::PostCombatProcessTurns() {
     m_universe.UpdateEmpireObjectVisibilities();
 
 
+    DebugLogger() << "ServerApp::PostCombatProcessTurns applying Newly Added Techs";
+    // apply new techs
+    for (auto& entry : empires) {
+        Empire* empire = entry.second;
+        if (empire && !empire->Eliminated())
+            empire->ApplyNewTechs();
+    }
+
+
     TraceLogger(effects) << "ServerApp::PostCombatProcessTurns Before Final Meter Estimate Update: ";
     TraceLogger(effects) << objects.Dump();
+
 
     // redo meter estimates to hopefully be consistent with what happens in clients
     m_universe.UpdateMeterEstimates(false);
@@ -3570,14 +3580,6 @@ void ServerApp::PostCombatProcessTurns() {
     for (auto& entry : empires)
         entry.second->UpdateOwnedObjectCounters();
     GetSpeciesManager().UpdatePopulationCounter();
-
-
-    // apply new techs
-    for (auto& entry : empires) {
-        Empire* empire = entry.second;
-        if (empire && !empire->Eliminated())
-            empire->ApplyNewTechs();
-    }
 
 
     // indicate that the clients are waiting for their new gamestate


agrrr3:
Somebody broke the build - or we moved to c++14 without me noticing.

Condition.cpp:9550:23: Fehler: Verwendung von »auto« in Lambda-Parameterdeklaration ist nur mit »-std=c++14« oder »-std=gnu++14« verfügbar

agrrr3:
Reset to the weekly-test-build and cherry-picked the commit in order to evade the c++14 issue. Builds and test fine now

geoffthemedio:
> Somebody broke the build - or we moved to c++14 without me noticing.

Was me.  Should be able to change the return od the lambda to `std::string`.

geoffthemedio:
Probably more testing reports would be good

agrrr3:
I have the impression people only test code if it is in the test builds. 
And I sadly do not have time to play the game to look for issues.
I'll ask on the forum *sigh*

agrrr3:
Or is there a please-test-everybody tag on our github?

Vezzra:
@agrrr3,

> Or is there a please-test-everybody tag on our github?

Yep, see the label that @o01eg has added.

geoffthemedio:
If we can't get any test feedback by asking, then this can be merged to force this issue, though I'd prefer pre-merge testing.

agrrr3:
No feedback for a week. I guess we had some implicit testing because we put the code closer to the state before moving the application so far down the code (in order to be after turn increase).

Oberlus:
I've tested the effect on population techs.
Initial population is 20.
I got on turn 7 the sitrep for Planetary Ecology researched.
In that turn, target population is 21 (correct) and predicted population for next turn is 20 (no increase, I don't know if this is correct because tech needs one more turn to kick in or if it is wrong).
In turn 8 population is 20 (consistent with prediction from previous turn), target population is 21, predicted population is 20.4.
In turn 9 population is 20.4 (consistent with predicted), predicted population in 20.73.
In turn 10 population is 20.73.

Everything seems right, unless we should be getting the first population increase in the turn right after the tech sitrep (turn 8).

This probably can be merged.

Vezzra:
For the sake of making progress with the release, I'm going to merge this.

PR #2601, opened 2019-10-21
Improve SitRepPanel::SizeMove() performance

Morlic-fo:
Before this PR, resizing the SitRep window causes the UI to freeze/lag when multiple (N>10) sitrep entries are displayed (i.e. not filtered out).

When resizing the SitRep window, the entries contained in the ListBox are not expected to change. It is therefore unnecessary to rebuild all its contents, i.e. to call the very expensive SitRepPanel::Update().

Tested and seems to work fine ingame: No noticable slowdown of the UI while resizing and all entries are displayed / responding as expected.


geoffthemedio:
Seems reasonable

PR #2603, opened 2019-10-23
Cache GetResourceDir() results

Morlic-fo:
In current master, every call to GetResourceDir() verifies that the directory exists which is expensive.

The solution proposed in this PR is to check for validity of the ResourceDir only on changes. While the user could, in principle, delete their resource dir in between calls, handling this case here, seems rather excessive.

Profiled with MSVC2017 using default msvc2017 solution (Release build). Test case was sitting ingame and auto-advancing turns (i.e. watching AI play).
Current master: GetResourceDir() takes roughly 10% of CPU time in FreeOrion.exe. 
With this PR, it is less than 1%.

### DIFF
--- a/util/Directories.cpp
+++ b/util/Directories.cpp
@@ -440,16 +440,30 @@ void CompleteXDGMigration() {
 const fs::path GetResourceDir() {
     // if resource dir option has been set, use specified location. otherwise,
     // use default location
+
+    static fs::path resource_dir_cached;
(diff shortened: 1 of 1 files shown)


Morlic-fo:
Not sure what I thought when pushing this, this is obviously not thread-safe.

geoffthemedio:
Would locking a mutex before accessing those statics not resolve the threading issues?

Issue #2607, opened 2019-10-25
The server freezes when sending messages on new turn


o01eg:
### Environment

* **FreeOrion Version**: v0.4.8+ https://github.com/o01eg/freeorion/commit/b45f7873724b7274a3278efd3aad13a61c9558f4 based on 4f372cb
* **Operating System**: Ubuntu Server 18.04.3 x64
* **Graphic card used**: No card
* **Fetched as** 
  * Compiled from source

### Description
When server send new turn messages it can freeze on network hiccups due synchronous sending.

Thread dump from core made while server was freezed:
```c++
(gdb) thread apply all bt

Thread 1 (Thread 0x7f6a37286740 (LWP 678)):
#0  0x00007f6a33d4ebf9 in __GI___poll (fds=fds@entry=0x7ffdbf87c628, nfds=nfds@entry=1, timeout=timeout@entry=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
#1  0x000055a253006dde in poll (__timeout=-1, __nfds=1, __fds=0x7ffdbf87c628) at /usr/include/x86_64-linux-gnu/bits/poll2.h:46
#2  boost::asio::detail::socket_ops::poll_write (ec=..., state=0 '\000', s=20) at /usr/include/boost/asio/detail/impl/socket_ops.ipp:1858
#3  boost::asio::detail::socket_ops::sync_send (ec=..., all_empty=<optimized out>, flags=0, count=<optimized out>, bufs=0x7ffdbf87c670, 
    state=<optimized out>, s=20) at /usr/include/boost/asio/detail/impl/socket_ops.ipp:1210
#4  boost::asio::detail::reactive_socket_service_base::send<boost::asio::detail::consuming_buffers<boost::asio::const_buffer, std::vector<boost::asio::const_buffer, std::allocator<boost::asio::const_buffer> > > > (flags=0, this=<optimized out>, ec=..., buffers=<synthetic pointer>..., impl=...)
    at /usr/include/boost/asio/detail/reactive_socket_service_base.hpp:184
#5  boost::asio::stream_socket_service<boost::asio::ip::tcp>::send<boost::asio::detail::consuming_buffers<boost::asio::const_buffer, std::vector<boost::asio::const_buffer, std::allocator<boost::asio::const_buffer> > > > (this=<optimized out>, ec=..., flags=0, buffers=<synthetic pointer>..., impl=...)
    at /usr/include/boost/asio/stream_socket_service.hpp:318
[...]

PR #2609, opened 2019-10-26
Async write returns nothing so remove bool result

o01eg:
A little refactoring for #2607 in preparation for asynchronous message sending.


geoffthemedio:
I can't really comment on the functional changes...

PR #2618, opened 2019-10-27
Don't automatically choose empire based on player name in the in-game lobby

o01eg:
Extracted from #2617 to separately manage UI and protocol changes.

As it is a required changes for #2617 which is a important feature for long multiplayer games I prefer to have it 0.4.9 or even next test build.

Actually its a bug when a player see himself in the in-games lobby if he doesn't play this empire.

### DIFF
--- a/UI/MultiplayerLobbyWnd.cpp
+++ b/UI/MultiplayerLobbyWnd.cpp
@@ -364,11 +364,12 @@ namespace {
 
     // Row for player info when loading a game
     struct LoadGamePlayerRow : PlayerRow {
-        LoadGamePlayerRow(const PlayerSetupData& player_data, int player_id, const std::map<int, SaveGameEmpireData>& save_game_empire_data, bool disabled) :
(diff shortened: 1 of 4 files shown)


Vezzra (commented):
As I'm not familiar with the code sections in question, I can't really give a meaningful code review. I have to defer to @geoffthemedio for that...

Vezzra:
@o01eg,

> As it is a required changes for #2617 which is a important feature for long multiplayer games I prefer to have it 0.4.9 or even next test build.

I can see where this is an important and desired feature, so ok. However, we need to keep the last minute addition of new features that should make it into the release to a minimum, otherwise we'll never get the release out.

There will always be yet another good/desireable/important feature that should be included, we need to draw the line somewhere.

Concerning getting this into the next test builds: no problem, if it gets merged in time.

o01eg:
@Vezzra This is why I split it into this PR and server-side.

o01eg:
I'll take a risk and merge it. At least it compiles and tested on CI and I tested it for in-game lobby.

Issue #2604, opened 2019-10-23
Some Tracelogger statements affect performance despite loglevel debug


Morlic-fo:
### Environment
* **FreeOrion Version**: master as of eb42b0e
* **Operating System**: Win10
* **Fetched as**
  * Compiled from source

### Description
Despite having all log levels set to debug, at least some TraceLogger statements are significantly impacting performance. In particular, there are the statements added by @geoffthemedio in https://github.com/freeorion/freeorion/commit/b7cd0fef4f44652d5520632f5b210d9958be1d7e - not sure if there are any other.

The following profiling results are for a the standard MSVC2017 solution Release build. Setup was starting a game with default settings, enqueue SuperTester building to get full map vision. After this setup, profiler was attached to FreeOrion.exe process and turns auto-advanced. Profiling duration was 15 seconds.
The percentages displayed are relative to total CPU time spent in FreeOrion.exe.
![TraceLogger](https://user-images.githubusercontent.com/11706951/67437941-6eb8b400-f5f2-11e9-85bc-610b9ef81ec8.PNG)

In a next test, I used the following diff to very that the lambda is indeed executed and this is not a profiling error:
```
diff --git a/universe/Condition.cpp b/universe/Condition.cpp
index 6397d6b24..8ba07303c 100644
--- a/universe/Condition.cpp
+++ b/universe/Condition.cpp
@@ -9549,6 +9549,7 @@ void And::Eval(const ScriptingContext& parent_context, ObjectSet& matches,

     auto ObjList = [](const ObjectSet& objs) -> std::string {
         std::stringstream ss;
+        ErrorLogger() << "This should not be called";
         for (const auto& obj : objs)
             ss << obj->Name() << " (" << std::to_string(obj->ID()) << ")  ";
         return ss.str();
```
and sure enough
```
23:24:03.773264 {0x00000764} [error] client : condition.cpp:9552 : This should not be called
```

### Expected Result
Logger statements should be lazily evaluated and not affect performance if the relevant log level is not set.

geoffthemedio:
At least the trivial logger is supposed to be lazy: https://www.boost.org/doc/libs/1_67_0/libs/log/doc/html/log/tutorial/trivial_filtering.html

Possibly in general it doesn't though?: 
https://www.boost.org/doc/libs/1_62_0/libs/log/doc/html/log/rationale/why_not_lazy_streaming.html

Are you sure the inner error logger doesn't override the lazy evaluation filtering somehow?

[...]

PR #2619, opened 2019-10-28
Fix log filters not applying while setting a new log level

Morlic-fo:
By resetting the filter before registering the new and updated filter, there was a short time frame where logging filters were not applied (as logging is handled asynchronously). As such, e.g. TraceLogger statements could be (partially) logged while log level was set to Debug and higher, see e.g. #2604.


Vezzra:
#2604 fixed by this, or does it need more work?

Morlic-fo:
This PR eliminated the spurious and incorrectly called trace-log statements but has no significant impact on the (inactive) trace logger overhead, so not fixed.

PR #2621, opened 2019-10-29
Alleryn fuel pedia fix

agrrr3:
This is alleryn's #2596 plus a single commit with necessary AI changes.

I had a look at the calculated values in game and it looked sane.

So the hull descriptions need updating but I think that could be done in a followup PR.
Alleryn could either cherry-pick my change or merge this one (?)

### DIFF
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -1041,7 +1041,7 @@ class ShipDesigner(object):
         if self.design_stats.fuel < 0:
             self.design_stats.fuel = 0
         else:
-            self.design_stats.fuel *= self._hull_fuel_efficiency()
+            self.design_stats.fuel = (self.design_stats.fuel - self.hull.fuel) * self._hull_fuel_efficiency() + self.hull.fuel
 
     def add_design(self, verbose=True):
         """Add a real (i.e. gameobject) ship design of the current configuration.
--- a/default/scripting/ship_hulls/SH_BASIC_MEDIUM.focs.txt
+++ b/default/scripting/ship_hulls/SH_BASIC_MEDIUM.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_BASIC_MEDIUM"
     description = "SH_BASIC_MEDIUM_DESC"
     speed = 60
-    fuel = 1.5
+    fuel = 3
     stealth = 5
     structure = 10
     slots = [
--- a/default/scripting/ship_hulls/SH_BASIC_SMALL.focs.txt
+++ b/default/scripting/ship_hulls/SH_BASIC_SMALL.focs.txt
@@ -2,7 +2,7 @@ Hull
     name = "SH_BASIC_SMALL"
     description = "SH_BASIC_SMALL_DESC"
     speed = 75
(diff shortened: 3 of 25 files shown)


alleryn:
> Alleryn could either cherry-pick my change or merge this one (?)

Merging this one is fine with me.

default/python/AI/ShipDesignAI.py
```diff
@@ -1041,7 +1041,7 @@ def parse_tokens(tokendict, is_hull=False):
         if self.design_stats.fuel < 0:
             self.design_stats.fuel = 0
         else:
-            self.design_stats.fuel *= self._hull_fuel_efficiency()
+            self.design_stats.fuel = (self.design_stats.fuel - self.hull.fuel) * self._hull_fuel_efficiency() + self.hull.fuel
```
Morlic-fo, line 1044: Not introduced in this PR, so I could also fix this separately after this is merged... There is another missing fuel effect here:
```
if self.design_stats.fuel < 1: 
    self.design_stats.fuel = 0
```

Vezzra:
@Morlic-fo, @agrrr3, @alleryn, @geoffthemedio: #2596 was supposed to fix #2541. With this merged, can #2541 be considered properly addressed and closed as resolved?

alleryn:
> @Morlic-fo, @agrrr3, @alleryn, @geoffthemedio: #2596 was supposed to fix #2541. With this merged, can #2541 be considered properly addressed and closed as resolved?

A far as i'm concerned, probably.  @Morlic-fo wrote

> 1. For example the [basic small hull](https://github.com/freeorion/freeorion/blob/master/default/scripting/ship_hulls/SH_BASIC_SMALL.focs.txt) has base fuel 2. That is also the value which appears in the pedia. Effectively, however, the hull has 8 fuel due to the `[[GREAT_FUEL_EFFICIENCY]]` effect.
> 2. The fuel efficiency of a hull is never mentioned in the pedia.

Point 1 has been addressed (i believe).  As for  point 2, fuel efficiency is mentioned as a general concept (not sure if that was added to the pedia before or after the issue), but the fuel efficiency of each hull is not explicitly stated.  It's open to debate whether that is a concern or not.

geoffthemedio:
> the fuel efficiency of each hull is not explicitly stated. It's open to debate whether that is a concern or not.

That seems like a concern, though a different one from the original issue this addressed, which was that the actual fuel a hull gave wasn't what was shown in the pedia due to the efficiency being applied on top of that. Now the issue is, presumably, that it's not clearly documented what fuel efficiencies hulls have, and thus how much fuel capacity parts in designs using that hull will contribute.

Vezzra:
So, close #2541 and open a separate issue for the remaining problems/shortcomings, or just leave #2541 open (until the remaining problems have been addressed)?

@Morlic-fo, what do you think (as the one who opened #2541)?

Morlic-fo:
No strong opinion, your call.

Vezzra:
Ok, I went ahead and closed #2541 as resolved.

Please open a new issue for the remaining problems with pedia fuel information.

alleryn:
> Please open a new issue for the remaining problems with pedia fuel information.

#2637

PR #2622, opened 2019-10-29
Ship weapons stop targeting comsats/empty-hull-chaff

agrrr3:
A comsat is a ship which has no maximum speed and no dangerous ship parts.
Ship part classes which are considered dangerous/worthwhile targets:
+ PC_TROOPS
+ PC_FIGHTER_HANGAR if also PC_FIGHTER_BAY in the ship
+ PC_COLONY,
+ PC_GENERAL,
+ PC_BOMBARD,

[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11402)
[older forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=10854#p91265)

Please somebody add the testing requested label

### DIFF
--- a/combat/CombatSystem.cpp
+++ b/combat/CombatSystem.cpp
@@ -315,7 +315,7 @@ namespace {
 
             std::unique_ptr<Condition::ConditionBase>{VisibleEnemyOfOwnerCondition()});
 
-    const std::unique_ptr<Condition::ConditionBase> is_enemy_ship_fighter_or_armed_planet =
+    const std::unique_ptr<Condition::ConditionBase> is_enemy_ship_fighter_or_armed_planet_no_comsat =
         boost::make_unique<Condition::And>(
             std::unique_ptr<Condition::ConditionBase>{VisibleEnemyOfOwnerCondition()},  // enemies
             boost::make_unique<Condition::Or>(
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Should probably add these sorts of filters in FOCS, not C++.

Will this attack unarmed ships once there are no more targets?

agrrr3:
> Should probably add these sorts of filters in FOCS, not C++.

Yes, I also think so. But what to do with the default targeting? We do not have composition of target conditions so every standard weapon would need to have an explicit targetCondition. 

Introducing something like named conditions like CD_SHORT_RANGE_DEFAULT_TARGETS, CD_FIGHTER_DEFAULT_TARGETS, CD_PLANET_DEFAULT_TARGETS 
which would be used by the C++ code as default would probably work. These would then be defined in FOCS. We could also add a `Condition name = <CONDITION_NAME>` condition in order to compose conditions where macros do not suffice.

Also I forgot that at least bombers should ignore comsats.

What are your preferences?

> 
> Will this attack unarmed ships once there are no more targets?

Currently not. In the last poll I made on the forums the majority had a preference for doing away with fallback targeting.

edit1: also note that other weapon systems still attack comsats

Also Oberlus reminded me that the small energy hull costs also just 5PP, so probably that should be also considered a comsat although one could argue that the ability to build the base locally in the combat system makes it a comsat. I mainly introduced the speed condition so effect processing can skip checking the six-times-nested-OR.

So I guess this is again work in progress.

agrrr3:
> > Oberlus reminded me that the small energy hull costs also just 5PP, **so probably that should be also considered a comsat**
> 
> I don't follow you. The compressed energy hull can be armed, while the base hull can't (no external slot).

The targeting in this PR ignores ships without weapons.., so I meant harmless small energy hull.

I didnt want to move the discussion here, lets go back to the forums.

Vezzra:
> lets go back to the forums.

Yes, please!

However, there is a technical issue with including this in 0.4.9: @freeorion/ai-team, how much does the AI rely on/employ comsats for cannon fodder tactics? Meaning, how much would the AI be affected if this gets merged?

Edit: Unless I get confirmation from the AI team that this won't affect the AI, I'll tag it as post release.

geoffthemedio:
> where macros do not suffice.

Where do macros not suffice?

> In the last poll I made on the forums the majority had a preference for doing away with fallback targeting.

OK, but it has to be handled appropriately. Having specialized parts that target with restrictions is fine, but not making it the default. I don't think it's sensible to have broad classes of ship that just don't get attacked ever even if there is nothing else to attack, particularly if there is no in-game explanation why they can't be. Target preferences with fallback or probabilistic selection are valid, though debatable, ways to avoid that. Guns that just won't shoot at something for no reason probably isn't. A specialized part that's intended to only attack certain stuff is fine, though, as it can be given a reason for working that way.

> how much does the AI rely on/employ comsats for cannon fodder tactics?

This does seem like a rather too large change for a hopefully imminent release... At least in the test game I ran earlier today, the AIs had substantial numbers of "Comsat" ships built.

agrrr3:
#2628 implements a balancing by doubling base hull cost. That PR should definitly go in first.

Also #2629 is related as it fixes the keep-your-comsats-as-surprise-without-having-to-pay-upkeep-for-it micromanagement loophole.

Vezzra:
> This does seem like a rather too large change for a hopefully imminent release

I agree. So this will stay assigned to post release.

geoffthemedio:
Is this relevant after https://github.com/freeorion/freeorion/pull/2665 ?

Oberlus:
> Is this relevant after #2665 ?

Comsats are still ships (not fighters), so they are still targetted by "big guns".
However, I'd say this isn't relevant anymore because
- #2628 adjusted cost of base hulls, they are not as inexpensive as they were and can't be exploited, and
- #2688 introduced a weapon that is quiet effective against Comsats (and any other kind of cheap, weak ships used as chaff).

@agrrr3 ?

agrrr3:
I think this PR can be closed until we get evidence that chaff-by-numbers is OP. I already asked for feedback but did not get. I doubt that this will occur specifically for the base hull so I renamed the pull request to include "chaff". We can reopen if need arises.

PR #2612, opened 2019-10-26
field graphics improvements

TheSilentOne1:
- clearer image of ion storm and molecular cloud (better export method)
- new accretion disk image
- new images for star forming nebulae
- extend star forming nebula mechanic: increase in stealth by 20, decrease of speed and detection by 20

(Star forming nebulae should have some game play impact beyond sometimes creating a sun. It seems intuitive that they affect stealth, detection and speed.)

PR #2624, opened 2019-11-01
Revision and extension of fields

TheSilentOne1:
WIP:
- introduces new fields: nanite swarm, meteor blizzard, void rift
- updated graphics for ion storm and molecular cloud
- new graphics for star forming nebulae and accretion disk
- gameplay changes: fields spawn closer to the center of the galaxy, but move less inwards; reduce occurence of fields; field size adjustments; star forming nebula affect speed, detection, stealth
- misc. minor changes

Finalised graphics are molecular cloud, accretion disk and the star forming nebulae, other graphics are subject to change.

@Vezzra: since, if I'm not mistaken, you implemented fields in the first place, I'd be thankful for your approval of the gameplay changes. The reason for changing the spawning rate and placement of fields is that, especially for larger galaxies, they would only appear in the periphery and the center would be empty. Also, in my tests, the spawning rate for fields seemed too high so that fields would often overlap. On top of that I've reduced the size of some fields because (with a more discernible border) they appeard too large.

Testing and comments appreciated!

Supercedes https://github.com/freeorion/freeorion/pull/2612.


Vezzra (commented):
> The reason for changing the spawning rate and placement of fields is that, especially for larger galaxies, they would only appear in the periphery and the center would be empty.

That has been intentional. The idea was to create more diversity for different "galactic regions", so to speak: the outer rim would be more exposed to molecular clouds and ion storms, while the center would be free of them.

> Also, in my tests, the spawning rate for fields seemed too high so that fields would often overlap.

I don't see the overlapping of fields as a particular issue, but if the general consenus of the current playtesters is that too many fields get spawned, sure, feel free to reduce the spawn rates. I have never done extensive playtesting, so my values are more or less just a rough first try supposed to be fine tuned anyway.

> On top of that I've reduced the size of some fields because (with a more discernible border) they appeard too large.

Again, if that's the general consensus of the playtesters, fine with me (what I said above applies to these values too). I'd just try to still have sufficiently large differences in size between the different fields to maintain diversity. If every type of field is roughly the same size, that would be boring IMO.

I've seen you simplified my placement algorithm and tweaked the movement speed of the fields. I don't have the time to playtest the changes, so I rely on you having sufficiently tested the results these changes produce.

I recall having had to tinker quite a lot with the placement, movement and growth speed, and the duration how long the fields last before they dissipate, because it wasn't easy to get all the factors/parameters right. Either the fields grew too fast or too slow over their lifespans, or they moved too fast to the center, or they dissipated to soon (so they only ever touched the outermost systems) or too late (which resulted in cumulating fields in the center) etc.

Which is why I'm a bit weary if that simplifications will work well. If they do however, sure, go ahead. My implementation was only meant as a starting point to get the feature reasonably working at all, it has always been meant to be tweaked, fine-tuned and optimized anyway.

Oberlus:
I've tried to test this. My cmake is giving me warnings:

> Imported targets not available for Boost version 106501

I don't have that problem with other branch I've tested. Are we supposed to update to a newer Boost?

[cmake_output.txt](https://github.com/freeorion/freeorion/files/3813817/cmake_output.txt)

o01eg:
@Oberlus It's only warnings, try to build and run project.

TheSilentOne1:
C++ code is (basically) unchanged, you should not have any troubles.

Oberlus:
All right, I'm on it.

Amount of fields seems right. They seem to wander in many directions.
Overlapping means that you can't check by right-clicking the borders of the field that is below the other. That's a problem, but solvable by making a careful adjustment of the graphics area and the effects area.

**Nebula Cloud**
I love the effects: -20 speed and detection range, +20 stealth. And the graphics.
The effect happens a bit too much outside the graphic. In a cloud (the graphic) of radius 100 uu, the effect was still felt at 105 uu from the centre. I suggest reducing the area of effect a 5 or 7% with respect to the graphic area.

**Meteor Blizzard**
The damage effect, 5 structure per turn, is huge for early game and negligible for late big hulls. I suggest turning it into a percentage of the ship max structure, like 10 or 20%.
The borders are easier to see without right-clicking, but a bit outside of the graphic area, I'd reduce the effect area 5%.

**Nanite Swarm**
Not so sure here, but maybe it would be better to make the repair rate also a percentage.
Borders quite easy to recognise, but again a 3 or 5% reduction of the area effect with respect graphics would be welcome.

**Ion Storm**
5% reduction on the effects area would be good.
The color is quite similar to the nanite swarm, but the texture is different so distinguishable. However, making it pink as it was before could be a good idea.

**Molecular Cloud**
The colors are too close to the nebulae, almost indistinguishable. I retire my previous suggestion about making ion storms pink. Molecular clouds could be pink-ish to differentiate from nebulae.
Also a 5 to 10% reduction on the effect area would be great.

**Void Rift**
Another cool effect: -50% speed.
The borders are quite difficult to realise unless you make a lot of right-clicks, because there is a lot of "black" space (intuitively out of the effect) that is actually within the void rift.
I suggest reducing the area of effect with respect to the graphics area, but much more than for the others. Not sure how much, I'd try -33% for starters.
Check out the image: that Void Rift is predating almost 75% of the meteor blizzard although you could expect its effect to fade away way sooner. 
voidRiftTooWide

TheSilentOne1:
Thanks for your feedback, Oberlus, I've adressed most of your points, void rift and nano swarm images are still WIP. Would definitely like to hear your opinion about field sizes, too.

TheSilentOne1:
Graphics are finalised, script is working. Aside from fine-tuning, this should be good to go. More fields can go in later.

Oberlus:
Tested this again, two games (150 systems) with seeds 0 and random, until round 100.
Second game got only nanite swarms and one ion storm, but the game with seed 0 shows you all fields.
Wandering of fields seems to follow a clock-wise pattern around galaxy centre (I really like that).

Void rift's area of effect is much more close to the graphics area. Good job.

I still feel all fields have an area of effect slightly bigger than what is the visible graphics area.
In general, I would prefer the opposite: area of effect slightly smaller than graphics, but ideally to match perfectly.
I guess another reduction of 5% is convenient.

I think this "should" be merged for 0.4.9:
- It's a cool addition.
- The new effects are small enough to not be a source of concern regarding gameplay.

TheSilentOne1:
> Which is why I'm a bit weary if that simplifications will work well. If they do however, sure, go ahead. My implementation was only meant as a starting point to get the feature reasonably working at all, it has always been meant to be tweaked, fine-tuned and optimized anyway.

I admire the (python) work you've done. I hope you do approve of my changes when you get to try them out in-game.

> I guess another reduction of 5% is convenient.

Applied. Thanks again for your help with testing.

@geoffthemedio, @Vezzra: as far as I'm concerned this can be merged before the release. Okay for you?

geoffthemedio:
I think adding several new fields should be held to be post-release.

TheSilentOne1:
PR with graphics only: https://github.com/freeorion/freeorion/pull/2636

Vezzra:
@TheSilentOne1,

> as far as I'm concerned this can be merged before the release. Okay for you?

Not a particularly strong opinion, my main concern would be how much those new fields might impact/hamper the AI. AFAIK it doesn't care about the already existing fields at all, which of course is not optimal, and adding more fields with potentially stronger effects might compound that problem.

If @geoffthemedio prefers to postpone this, I think we should.

agrrr3:
What happens with AI in a blizzard? Does it move it ships to repair or is the fleet dead after 10 turns?

TheSilentOne1:
It doesn't know about the blizzard, that much is for sure. It may react to the ships taking damage though - @Morlic-fo, is there something like an AI fleet repair mission?

Morlic-fo:
There are repair missions but they are probably insufficient in this context because the AI may easily try to move through the field to repair or only return after it passed more than 50% of the field (i.e. dies on the way back).

The other fields should be fine (i.e. not worse than current fields which aren't really handled by AI either).

geoffthemedio:
Are the AI issues with this sufficiently resolved?

TheSilentOne1:
The damage effect was removed from meteors. Other than that @Morlic-fo didn't have any objections. I plan to work some more on the graphics sometime soon, but it may be a while until I get to it.

adrianbroher:
Can you fix the turn_event script to let Travis-CI run properly?

TheSilentOne1:
Hardly knowing any python, I don't know what to do. Does it have to do with the ongoing change of the python version?

o01eg:
@TheSilentOne1 No, it was python2 lint error. Also I would like you to rebase PR to try it against python3 lint checker as well.

adrianbroher:
> Hardly knowing any python, I don't know what to do. Does it have to do with the ongoing change of the python version?

The error message is pretty obvious:

```
$ flake8
./turn_events/turn_events.py:8:1: F401 'math.hypot' imported but unused
The command "flake8" exited with 1.
```

As seen here: https://travis-ci.org/freeorion/freeorion/jobs/619167544#L221-L223

Vezzra:
As @o01eg pointed out, that's not an error in `turn_events.py`, but an issue with the checker. It falsely reports `math.hypot` as not used in the code, but that function actually *is* used (line 31, only one occurrence, but still).

adrianbroher:
It is not, check again.

Vezzra:
Argh, sorry, you're right, I checked the wrong branch. This PR removed the one usage of `hypot`.

@TheSilentOne1: You just need to remove `hypot` from the `from math import` statement [on line 8](https://github.com/TheSilentOne1/freeorion/blob/532e0720932042247160618f81d8ac0218f7dd7e/default/python/turn_events/turn_events.py#L8).

o01eg:
@Vezzra I mean it was an error found by python2 lint checker, not the error in checker itself.

TheSilentOne1:
> The error message is pretty obvious: ...

I didn't think a warning like that would actually throw an error. Fixed.

TheSilentOne1:
> Also I would like you to rebase PR to try it against python3 lint checker as well.

@o01eg: I will do it, but not too soon, too busy currently.

adrianbroher:
> I didn't think a warning like that would actually throw an error. Fixed.

The whole point of the CI is to ensure enhanced code quality. Pendantic rule enforcement is a quality assurance feature, not a flaw.

geoffthemedio:
Please rebase on master; don't merge master into your feature branch.

TheSilentOne1:
That's what you get for using for using the web-interface... will fix it soon.

TheSilentOne1:
Superceded by https://github.com/freeorion/freeorion/pull/2978

geoffthemedio:
You can force-push an existing branch, without need for a new one or a new pull request.

TheSilentOne1:
I did just now, however I can't reopen this PR.

PR #2625, opened 2019-11-01
Fix exception warning

Vezzra:
There are nine places where exceptions are caught by value instead of by reference. This patch changes these to const&. Patch provided by spikethehobbit on the forum.

### DIFF
--- a/UI/OptionsWnd.cpp
+++ b/UI/OptionsWnd.cpp
@@ -133,7 +133,7 @@ namespace {
             fs::path path = FilenameToPath(file);
             return boost::algorithm::ends_with(file, STRINGTABLE_FILE_SUFFIX) &&
                 fs::exists(path) && !fs::is_directory(path);
-        } catch (std::exception ex) {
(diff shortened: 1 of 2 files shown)


UI/OptionsWnd.cpp
```diff
@@ -133,7 +133,7 @@ namespace {
             fs::path path = FilenameToPath(file);
             return boost::algorithm::ends_with(file, STRINGTABLE_FILE_SUFFIX) &&
                 fs::exists(path) && !fs::is_directory(path);
-        } catch (std::exception ex) {
+        } catch (const std::exception&) {
```
geoffthemedio, line 136: none of the catch blocks in this file actually use the exception, so I don't think the change matters... but if updating, could switch to just `catch (...)`

Vezzra: If I understand correctly, the purpose of this patch is to silence compiler warnings, which should be achieved, regardless of the catch blocks using the exceptions or not.

So, good to merge?

PR #2628, opened 2019-11-03
Balance - double base hull cost, but triple internal slots

agrrr3:
[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11402) 

As the upping the production cost is KISS and AI can handle it better, we try this first before trying to change targeting priorities.
 
* +1 structure
* Doubling the cost should make comsats more expensive cannon fodder
  +3PP cost for colonisers or outpost does not make much difference
* Three internal slots makes for about 2PP hull cost per troop.
  Note that a while ago this was only 1PP hull cost per troop and that
  was not considered a problem AFAIK
* you can still trigger combat

Another TODO for making comsat spam less a problem:
* count unfinished ships in the production queue for hull and part upkeep
* maybe increase bouts per combat to four - this nerfs all cannonfodder by maybe 50%

### DIFF
--- a/default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
+++ b/default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
@@ -10,7 +10,7 @@ Hull
         Slot type = Internal position = (0.50, 0.50)
         Slot type = Internal position = (0.62, 0.50)
     ]
-    buildCost = 6
+    buildCost = 6 * [[FLEET_UPKEEP_MULTIPLICATOR]] * [[SHIP_HULL_COST_MULTIPLIER]]
     buildTime = 3
     tags = [ "PEDIA_HULL_LINE_GENERIC" ]
     location = OwnedBy empire = Source.Owner
@@ -20,3 +20,5 @@ Hull
     graphic = "hulls_design/colony_base_hull.png"
 
 #include "ship_hulls.macros"
+
+#include "/scripting/common/upkeep.macros"
--- a/default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
+++ b/default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
@@ -4,9 +4,13 @@ Hull
     speed = 0
     fuel = 0
     stealth = 5
-    structure = 2
-    slots = Slot type = Internal position = (0.50, 0.50)
-    buildCost = 3
+    structure = 3
+    slots = [
+        Slot type = Internal position = (0.38, 0.50)
+        Slot type = Internal position = (0.50, 0.50)
+        Slot type = Internal position = (0.62, 0.50)
+    ]
+    buildCost = 6
     buildTime = 3
     tags = [ "PEDIA_HULL_LINE_GENERIC" ]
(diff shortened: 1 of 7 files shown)


Vezzra:
@freeorion/ai-team, any AI adjustments required for these changes (considering the AI's current use of comsats)?

agrrr3:
Not a specialist, but diving into the code: the decoy specific code is in ShipDesignAI.py#OrbitalDefenseShipDesigner and the _rating function returns:
`(1+self._total_dmg_vs_shields()*self.design_stats.structure)/self._adjusted_production_cost())`

This is guarded by a speed less than 10 which means it always chooses base hull.
That probably means that structure gets ignored if you do not add a weapon which is not possible for base hull anyway. 

With the new cost balance it might be more efficient to build small asteroids or energy ships if you already have the shipyard. AI will not do that.

I may be wrong, but I also think that all the _rating_functions use the wrong formula. In the recent discussion it was important (at least for war ships) that the cost square is used instead of simple production cost and the code does not support that currently. *scared*

l29ah:
It would be nice if the upkeep multiplier will go in as well, as seen in https://github.com/Voker57/freeorion/commit/8a2478bb3b0875c276edece8e9ae4e4048a5c2fa

default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
```diff
@@ -4,9 +4,13 @@ Hull
...
-    structure = 2
-    slots = Slot type = Internal position = (0.50, 0.50)
-    buildCost = 3
+    structure = 3
+    slots = [
+        Slot type = Internal position = (0.38, 0.50)
+        Slot type = Internal position = (0.50, 0.50)
+        Slot type = Internal position = (0.62, 0.50)
+    ]
+    buildCost = 6
```
Oberlus: Should this cost be multiplied by upkeep factor? (also ship hull cost multiplier)
buildCost = 6 * [[FLEET_UPKEEP_MULTIPLICATOR]] * [[SHIP_HULL_COST_MULTIPLIER]]
Otherwise, spamming comsats will regain efficiency with larger fleets.

Morlic-fo:
> @freeorion/ai-team, any AI adjustments required for these changes (considering the AI's current use of comsats)?

I don't think this is significant enough to require AI changes. They are mostly used for triggering combat which this still allows.

default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
agrrr3: done.

o01eg:
Can it be landed in v0.4.9?

Vezzra:
@agrrr3, please also adjust the predefined ship designs based on the Base Hull accordingly, where appropriate.

agrrr3:
> @agrrr3, please also adjust the predefined ship designs based on the Base Hull accordingly, where appropriate.

Done, note I upped the troop part size to three. I think that is the most sensible default at a hull cost point of 6.

Vezzra:
As far as I can tell, looks good now. Can this be merged?

PR #2629, opened 2019-11-03
Add ShipDesignsInProduction ValueRef and use it for upkeep

agrrr3:
ShipDesignsInProduction returns the (count of) empire's ship batches in the production queue
if work on that project already started i.e. if you already spent PP on it.

[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11414)

NOTE: As we do not (yet) count the ship parts in the queue, part upkeep assumes the same cost
per ship in the queue as for ship based upkeep (equating to four ship parts).
This makes upkeep especially for unfinished decoys quite more expensive than for finished decoys.
It might make sense to penalize keeping unfinished projects in the queue.

This also calls UpdateOwnedObjectCounters when loading the game.
It is possible that other update calls are missing e.g. for statistics

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -2224,6 +2224,16 @@ void Empire::UpdateOwnedObjectCounters() {
         m_ship_designs_owned[ship->DesignID()]++;
     }
 
+    // ships in the queue for which production started
+    m_ship_designs_in_production.clear();
+    for (const auto& elem : m_production_queue) {
+        ProductionQueue::ProductionItem item = elem.item;
+
+        if ((item.build_type == BT_SHIP) && (elem.progress > 0.0f)) {
+            m_ship_designs_in_production[item.design_id] += elem.blocksize;
+        }
+    }
+
     // update ship part counts
     m_ship_part_types_owned.clear();
     m_ship_part_class_owned.clear();
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1354,7 +1354,7 @@ void Empire::AddNewlyResearchedTechToGrantAtStartOfNextTurn(const std::string& n
         return;
 
     // Mark given tech to be granted at next turn. If it was already marked, skip writing a SitRep message
-    auto result = m_newly_researched_techs.insert(name);
+    m_newly_researched_techs.insert(name);
 }
 
 void Empire::ApplyNewTechs() {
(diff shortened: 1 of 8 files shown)


agrrr3:
If this is post release should the included fix for UpdateOwnedObjectCounters go into 0.4.9 ?

Vezzra:
@agrrr3, which issue does this fix for `UpdateOwnedObjectCounters` address?

agrrr3:
> @agrrr3, which issue does this fix for `UpdateOwnedObjectCounters` address?

If you load a game, UpdateOwnedObjectCounters is currently not called which means that the counters are probably wrong in the turn when you load the game; this PR needs the update so the counter for ships in production is set. 
But as nobody complained about wrong counters this is certainly not very important.

geoffthemedio:
`UpdateOwnedObjectCounters` is called once per turn during `PostCombatProcessTurns`. Why would an extra call be needed when loading a saved game?

Vezzra:
@agrrr3, ok, I understand. In cases like this it would be better anyway to make a separate PR for the bugfix, and have the PR that needs that bugfix mark as dependent on the bugfix PR (simply by putting a proper note in the PR description).

It seems however we first need to clarify if there actually is a bug that needs to be fixed?

agrrr3:
Just a note that reading AI code I stumbled over upkeep calc there and that is very simple right now.
```def get_fleet_upkeep():
    # TODO: Use new upkeep calculation                                                                                                                                                                                                       
    return 1 + AIDependencies.SHIP_UPKEEP * get_aistate().shipCount
```
This could be improved by using new upkeep calculation (and counting currently built ships in the queue as well). 

@Dilvish-fo  @Morlic-fo @freeorion/ai-team I think AI does not shuffle production around(?), so build queue upkeep is always minimal so it does not need to be adressed in this PR IMHO.

Morlic-fo:
It does somewhat shuffle production around (i.e. move high priority items to front of the queue) but it does not intentionally delay completing an item.

agrrr3:
> `UpdateOwnedObjectCounters` is called once per turn during `PostCombatProcessTurns`. Why would an extra call be needed when loading a saved game?

The call in PostCombatProcess turns is too late for normal effect evaluation. The question is if the value is persisted or not. When I was developing the value was always zero for a loaded game. I'll check once more.

agrrr3:
So I checked. If I remove that counter update and I add three ships in production in turn 8 and load that game, the sitrep is like it was, in turn 9 ShipDesignsInProduction returns zero, and in turn 10 I see the correct value.

It works for ShipDesignsOwned though. ~~So any idea why that is?~~
edit: That is because I forgot how serialization works. Adding m_ship_designs_in_production to SerializeEmpire.cpp should do the trick.

PR #2632, opened 2019-11-05
Let METER_MAX_FUEL hull effect execute very very late

agrrr3:
This executes the max fuel effect of hulls after all other fuel effects, especially fuel efficiency.

This makes it possible to remove the workaround in ship_hulls.macros
and the effect accounting order is more sensible.

The priority change for the addition of hull fuel property was hardcoded into the cpp code in the first implementation. After discussions with geoff i pushed a new implementation which moves the effect to the hulls' FOCS definitions.  

As a followup PR one could shift the species fuel levels by e.g. 0.5 or 1.0 so that BAD_FUEL is 0 or 0.5 and there are no negative values to multiply by efficiency. For balancing all hulls would need to be loose 0.5/1.0 times efficiency max fuel.


universe/ShipDesign.cpp
```diff
@@ -51,7 +51,11 @@ namespace {
     // by the result of evalulating \a increase_vr
     std::shared_ptr<Effect::EffectsGroup>
     IncreaseMeter(MeterType meter_type,
-                  std::unique_ptr<ValueRef::ValueRefBase<double>>&& increase_vr)
+                  std::unique_ptr<ValueRef::ValueRefBase<double>>&& increase_vr,
+                  const std::string& accounting_label = "",
+                  const std::string& stacking_group = "", int priority = 0,
```
geoffthemedio: default priority should probably be `DEFAULT_PRIORITY` which is currently 100.

geoffthemedio: actually, no, this is fine, as it is consistent with existing behaviour.

universe/ShipDesign.cpp
```diff
@@ -603,7 +607,7 @@ HullType::~HullType()
 
 void HullType::Init(std::vector<std::unique_ptr<Effect::EffectsGroup>>&& effects) {
     if (m_fuel != 0)
-        m_effects.push_back(IncreaseMeter(METER_MAX_FUEL,       m_fuel));
+        m_effects.push_back(IncreaseMeter(METER_MAX_FUEL,       m_fuel, 500)); // FIXME MAGIC NUMBER
```
geoffthemedio: I think this should actually be done in FOCS rather than C++ code.

The currently proposed change would bake a specific priority into the C++, which otherwise isn't done (?) or at least isn't done very much. The relative values of the priorities are all set from the values in scripts, mainly in priorities.macros, so putting a value into the C++ that is determined mainbly based on what those scripts contain seems like a bad idea.

So in that case, I'd disable to autogenerated effect by adding `NoDefaultCapacityEffect` to the hull scripts, and instead add a custom scripted fuel effect to the hulls with whatever priority is desired. The fuel property will remain in the script and be displayed in the UI, but just won't be used to generate a default capacity effect for the affect hull parts.

Adding the disable capacity effects flag will also require tweaking the hull parser to recognize that flag... as I suspect it only works presently for parts. And probably doing that in a way that lets just the fuel effects be disabled, since hulls also have several other autogenerated effects, though those could also all be custom-scripted in the interim.

The custom scripted hull fuel effect probably needs to be a new [[VERY_VERY_LATE_PRIORITY]] or similar, since the fuel effeciency scaling effects are currently [[VERY_LATE_PRIORITY]]... though perhaps all the fuel effects could be bumped down one priority step so the unscaled hull contribution could be put a [[VERY_LATE_PRIOIRITY]].

geoffthemedio: Given all that, I'd push this to post-release. It sound like the current issue this is intended to fix is cosmetic, regarding ordering in effect accounting?

agrrr3: I would rather have the priority be a functional value in priorities.macros and slurp that into c++
Adding language features to selectively disable parts of default effects sound like a lot of headaches (and much work as well).

> Given all that, I'd push this to post-release. It sound like the current issue this is intended to fix is cosmetic, regarding ordering in effect accounting?

You can not really understand the effect from looking at the effect values I think; but maybe the current state is acceptable. Have a look and decide:
201910__2620__negative_fuel_effects_accounting__great_looks_bad

Intended look:
201910__2620__negative_fuel_effects_accounting__reordered_effects

geoffthemedio: > I would rather have the priority be a functional value in priorities.macros and slurp that into c++

I don't know what that means.

> Have a look and decide:

The effect accounting doesn't make sense / isn't clear regardless of the ordering of the list, I think.

agrrr3: > > I would rather have the priority be a functional value in priorities.macros and slurp that into c++
> 
I meant we have e.g. FUNCTIONAL_* stringtable entries using the stringtable mechanism to use values from FOCS in C++. We could introduce something similar to read a priority defined in priorities.macros e.g. AFTER_FUEL_EFFICIENCY_PRIORITY or APPLY_HULL_FUEL_PRIORITY = 500
Then we do not have the magic value in the code.


> The effect accounting doesn't make sense / isn't clear regardless of the ordering of the list, I think.

The point here is the scaling also for positive values. I expect that the fuel efficiency 400% is applied to the effects above because that is how it usually works. And with the intended UI that expectation is matched. 

edit1: make distinction between stringtables and FOCS more clear

agrrr3:
This could be a result with good/positive fuel efficiency and showing the relative effect, I think understandability is the best we could get:
201910__2620__positive_fuel_effects_accounting__plus_percents_reordered

geoffthemedio:
That is clearer, but I don't think the difference is a critical release-blocking issue, and don't like the suggestion of relying on stringtable entries for something like this. 

Disabling the autogenerated effects is an existing scripting feature (for parts) that is intended to address this specific issue (the default behaviour autogenerated effects not doing what the scripter wants), so I would use that for this situation as well.

agrrr3:
> Disabling the autogenerated effects is an existing scripting feature (for parts) that is intended to address this specific issue (the default behaviour autogenerated effects not doing what the scripter wants), so I would use that for this situation as well.

But is fuel efficiency not the default behaviour from now on, as it applies to all ships?

agrrr3:
So if I get you right we should redo all those four effects in FOCS?
```    if (m_fuel != 0)
        m_effects.push_back(IncreaseMeter(METER_MAX_FUEL,       m_fuel, 500)); // FIXME MAGIC NUMBER                                                                                                                                         
    if (m_stealth != 0)
        m_effects.push_back(IncreaseMeter(METER_STEALTH,        m_stealth));
    if (m_structure != 0)
        m_effects.push_back(IncreaseMeter(METER_MAX_STRUCTURE,  m_structure,    "RULE_SHIP_STRUCTURE_FACTOR"));
    if (m_speed != 0)
        m_effects.push_back(IncreaseMeter(METER_SPEED,          m_speed,        "RULE_SHIP_SPEED_FACTOR"));
```

geoffthemedio:
Adding `NoDefaultCapacityEffect` or perhaps something similar but not "Capacity"-specific to hull parsing to disable those stat-derived effects and then adding them back as FOCS would be reasonable. Separate flags for each of the hull stats could also work.

agrrr3:
> Adding `NoDefaultCapacityEffect` or perhaps something similar but not "Capacity"-specific to hull parsing to disable those stat-derived effects and then adding them back as FOCS would be reasonable. Separate flags for each of the hull stats could also work.

Is your general intention to move stuff to FOCS? In that case I should do a single flag.

As I remember extending the parsers to hurt a lot I think I won't be able to implement this before christmas holidays. Maybe I could give it a fast stab and put it on hold if i run into problems.

geoffthemedio:
@agrrr3 Did you run into problems?

The idea is to do something "simple" and sensible by default with the stats-derived effects, so that stats are reasonably consistent with how a part actually functions and thus useful to display in the pedia to summarize the part. But if something more specific / bespoke / complicated is needed, then there should be a way to override / disable the autogenerated effects so that something customized can be added as needed.

For priorities, hard coding them i to C++ should be avoided as they are primarily or nearly exclusively defined and relevant relative to other FOCS defined content, and I'd try to avoid imposing a particular priority value as happening at a particular relative time in the API itself.

agrrr3:
@geoffthemedio i am mainly confused about the requirements and i do not have the time on my hands for further experiments, so i (still) need your input.

So if I get the requirements right:

1. we want simple and sensible default effects based on stats (e.g. fuel meter refill)
2. it should be possible to override these effects if something more complicated is necessary
3. effect priorities should not be hard coded in C++ as they are relative to other FOCS effects

So regarding the topic of this PR.
1. we should provide a default effect that ships in supply should be refueled (i.e. a ship hull should not have to add that effect)
2. overriding these effects is currently not possible, but the aim of this PR is to establish default behavior, not override it - so adding the capability to override is only tangential to the topic of this PR (but definitly nice-to-have)
3. this i take that at least the priorities of the default effects should be defined in FOCS content

So disabling the autogenerated effects can not be the way to go (because of 1.) - i would have to add the default effects via a macro to every single hull type (and probably simply remove the default effects from c++ completely).

I suggest instead to somewhere specify the priorities for the default effects in FOCS content. Maybe even to define EffectGroups for the default effects. Defining default effects in FOCS would allow a FOCS scripter to actually see the default effects and take that as inspiration (or even change the default effects to ones liking). And back on topic it would allow for setting an explicit priority for each default effectsGroup. This would fulfill all three requirements (We have default effects on hulls. It is possible to override the default effects. Priorities are not hardcoded)

One implementation sketch for moving the default effects to a FOCS specification: 
Recreating the four default effects of a hull in a FOCS file should be easy. 
I would create a ship Hull name = "DEFAULT_HULL_PROTOTYPE", adding two effectGroups (one for each priority) with the appropiate effects. I would need to enforce that that DEFAULT_HULL_PROTOTYPE is parsed first before normal hulls are instantiated and copy/reference the effect groups of the prototype hull when initialising hulls. 
Maybe keep it completely out of the normal PartTypeManager hulls and add it maybe as member variable to the HullType.
I am fairly (but not completely) sure referencing the prototype effects should work.

geoffthemedio:
I prefer leaving the current behaviour as the default capacity effect and making the fuel efficiency stuff be implemented as a custom scripted effect with the defaults disabled. The new behavior is too complicated and novel and inconsistent with other meters and stats to use as the default capacity effect, even if you want to script it into all hulls. This judgement is also partly based on the complicated process of implementing it and the associated issues.

Adding a script somewhere for the default effects would overcomplicated them, I think. They are supposed to be a simple default behaviour that can be overridden as needed, not another layer of required scripting that complicates the parsing. Using macros to give all hulls the desired non-default behavior is preferable to that, I think.

agrrr3:
Force-pushed a new implementation. From the commits:
 Removed default hull fuel property effect.
 Readded hull fuel property effect to all hulls in FOCS.

Also added fuel efficiency to Xentronium hull in case there are fuel effects.

Only exceptions are immobile hulls (no effect added):

monster/SH_EXP_OUTPOST_HULL.focs.txt
monster/SH_GUARD_0_BODY.focs.txt
monster/SH_GUARD_1_BODY.focs.txt
monster/SH_GUARD_3_BODY.focs.txt
monster/SH_GUARD_MONSTER_BODY.focs.txt
monster/SH_IMMOBILE_FACTOR.focs.txt
monster/SH_TREE_BODY.focs.txt
SH_COLONY_BASE.focs.txt

geoffthemedio:
https://github.com/freeorion/freeorion/commit/ec416b457d2e68ef5069cbf7cffeb415e9048f9d adds a mechanism to disable each of the default effects from hull stats. Please use that in the hull scripts.

agrrr3:
> [ec416b4](https://github.com/freeorion/freeorion/commit/ec416b457d2e68ef5069cbf7cffeb415e9048f9d) adds a mechanism to disable each of the default effects from hull stats. Please use that in the hull scripts.

Done. All hulls have NoDefaultFuelEffect added.

universe/ShipDesign.cpp
```diff
@@ -1003,7 +1003,7 @@ std::vector<std::string> ShipDesign::Parts(ShipSlotType slot_type) const {
         ErrorLogger() << "Design hull not found: " << m_hull;
         return retval;
     }
-    const auto& slots = hull->Slots();
+    const std::vector<HullType::Slot>& slots = hull->Slots();
```
geoffthemedio: ?

parse/StringValueRefParser.cpp
```diff
@@ -43,6 +43,7 @@ namespace parse {
             |    tok.BuildingType_
             |    tok.Focus_
             |    tok.PreferredFocus_
+            |    tok.Hull_
```
geoffthemedio, line 46: Is this doing something for this pull request...?

agrrr3: it seems the merge went wrong in a big way

agrrr3: Actually this one is ok, i need to access the fuel value of the hull for the effect which replaces the default hull fuel effect:
 
effects = SetMaxFuel value = ( Value + HullFuel name = Source.Hull )

geoffthemedio: Could also have used `ThisHull`

agrrr3: Ah, true

geoffthemedio: but exposing Hull is also OK

aschuch247:
I am using 6311d93 compiled from source, and scout ships now have a fuel of 40. Isn't this a bit too much?

geoffthemedio:
Not sure about specifically 40, but try updating to the latest, or at least https://github.com/freeorion/freeorion/commit/57319af16ce01b707ebd4ddfbcee10128a344ff1

PR #2638, opened 2019-11-11
Use empire ID instead of incorrect player ID for diplomacy

### LINKED ISSUE #2633: diplomatic message indicator not appearing (longturn server)
On the longturn server, send a peace or alliance proposal to a player your empire is at war with. The player receiving the offer should see the diplo status indicator in the empires window, but it doesn*t appear.

o01eg:
Fixes #2633 missing diplomacy messages and flashing.

### DIFF
--- a/UI/PlayerListWnd.cpp
+++ b/UI/PlayerListWnd.cpp
@@ -320,7 +320,7 @@ namespace {
                 ErrorLogger() << "PlayerDataPanel::Render couldn't get client app!";
                 return;
             }
-            if (m_player_id != app->PlayerID()) {
+            if (m_empire_id != app->EmpireID()) {
                 // render diplomacy icon
                 switch (m_diplo_status) {
                 case DIPLO_WAR:                 WarIcon()->OrthoBlit(    UpperLeft() + m_diplo_status_icon_ul, UpperLeft() + m_diplo_status_icon_ul + ICON_SIZE); break;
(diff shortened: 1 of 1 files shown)


UI/PlayerListWnd.cpp
```diff
@@ -320,7 +320,7 @@ namespace {
                 ErrorLogger() << "PlayerDataPanel::Render couldn't get client app!";
                 return;
             }
-            if (m_player_id != app->PlayerID()) {
+            if (m_empire_id != app->EmpireID()) {
```
geoffthemedio, line 323: Is the player ID actually incorrect here?

UI/PlayerListWnd.cpp
```diff
@@ -333,7 +333,7 @@ namespace {
 
             // render incoming diplomatic message icon, if there is one
             const DiplomaticMessage& incoming_message_to_client =
-                Empires().GetDiplomaticMessage(m_player_id, app->PlayerID());
+                Empires().GetDiplomaticMessage(m_empire_id, app->EmpireID());
```
geoffthemedio, line 336: ... or just here?

UI/PlayerListWnd.cpp
o01eg: I suppose it doesn't change anything as I tested with observers and moderators. Maybe better also test for `ALL_EMPIRES`.

PR #2640, opened 2019-11-11
Fallback to IPv4 if cannot open UPD or TCP socket

### LINKED ISSUE #2626: Server unconditionally listens on a UDPv6 port for discovery requests
The following issue has been reported by @spikethehobbit on the forum:

### Environment

* **FreeOrion Version**: 6e933988d858e7ccf5832db72ac0afcd63bc1f9f
* **Fetched as**
  * Compiled from source

### Description

Since pull request #2536 was merged, the server unconditionally listens on a UDPv6 port for discovery requests. This causes two problems:

First, single player games should not be listening for server discovery since they should not be accepting LAN connections.

Second, requiring IPv6 breaks single player on IPv4-only machines because the server crashes on startup:

```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  what():  open: Address family not supported by protocol
Aborted
```
The problem commit is 6e933988d858e7ccf5832db72ac0afcd63bc1f9f.

Forum thread: https://www.freeorion.org/forum/viewtopic.php?f=9&t=11408

o01eg:
Fixes #2626

### DIFF
--- a/network/ServerNetworking.cpp
+++ b/network/ServerNetworking.cpp
@@ -30,8 +30,34 @@ namespace {
 class DiscoveryServer {
 public:
     DiscoveryServer(boost::asio::io_context& io_context) :
-        m_socket(io_context, udp::endpoint(udp::v6(), Networking::DiscoveryPort()))
-    { Listen(); }
+        m_socket(io_context)
+    {
+        // use a dual stack (ipv6 + ipv4) socket
+        udp::endpoint discovery_endpoint(udp::v6(), Networking::DiscoveryPort());
+
+        if (GetOptionsDB().Get<bool>("singleplayer")) {
+            // when hosting a single player game only accept connections from
+            // the localhost via the loopback interface instead of the any
+            // interface.
+            // This should prevent unnecessary triggering of Desktop Firewalls as
+            // reported by various users when running single player games.
+            discovery_endpoint.address(boost::asio::ip::address_v4::loopback());
+        }
+
+        try {
+            m_socket = udp::socket(io_context, discovery_endpoint);
+        } catch (const std::exception &e) {
+            ErrorLogger(network) << "DiscoveryServer cannot open IPv6 socket: " << e.what()
+                                 << ". Fallback to IPv4";
+            discovery_endpoint = udp::endpoint(udp::v4(), Networking::DiscoveryPort());
+            if (GetOptionsDB().Get<bool>("singleplayer"))
+                discovery_endpoint.address(boost::asio::ip::address_v4::loopback());
+
+            m_socket = udp::socket(io_context, discovery_endpoint);
+        }
+
+        Listen();
+    }
 
 private:
     void Listen() {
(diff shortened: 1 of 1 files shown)


network/ServerNetworking.cpp
```diff
@@ -30,17 +30,24 @@ namespace {
...
-        m_recv_buffer.fill('\0');
-        m_socket.async_receive_from(
-            boost::asio::buffer(m_recv_buffer),
-            m_remote_endpoint,
-            boost::bind(&DiscoveryServer::HandleReceive, this,
-                        boost::asio::placeholders::error));
+        // use a dual stack (ipv6 + ipv4) socket
+        udp::endpoint discovery_endpoint(udp::v6(), Networking::DiscoveryPort());
+
+        if (!GetOptionsDB().Get<bool>("singleplayer")) {
```
geoffthemedio: Why? This would prevent querying a single player server for its status from the same machine.

o01eg: Should single player server be answering to that queries?

geoffthemedio: Why not?

network/ServerNetworking.cpp
```diff
@@ -30,17 +30,24 @@ namespace {
...
+            m_socket = udp::socket(m_service, discovery_endpoint);
```
geoffthemedio: Instead of checking if it's a single or multi-player game, maybe put a try-catch block around this, or whichever line throws the exception that was causing the issue? Only setting up the socket in MP won't actually help if the MP server only have IPV4, but try-catch would avoid a crash in either case. Could fall-back to an IPV4 endpoint as well, if setting up the IPV6 throws.

network/ServerNetworking.cpp
o01eg: Made it listen loopback in singleplayer mode.

network/ServerNetworking.cpp
o01eg: Added try-catch block both for message and discovery ports.

network/ServerNetworking.cpp
```diff
@@ -30,11 +30,35 @@ namespace {
...
+        // use a dual stack (ipv6 + ipv4) socket
+        udp::endpoint discovery_endpoint(udp::v6(), Networking::DiscoveryPort());
+
+        if (GetOptionsDB().Get<bool>("singleplayer")) {
+            // when hosting a single player game only accept connections from
+            // the localhost via the loopback interface instead of the any
+            // interface.
+            // This should prevent unnecessary triggering of Desktop Firewalls as
+            // reported by various users when running single player games.
+            discovery_endpoint.address(boost::asio::ip::address_v4::loopback());
```
geoffthemedio, line 44: IPV4 loopback works with IPV6 endpoints?

o01eg: It works same way as `message_endpoint`.

geoffthemedio: That doesn't answer the question, which is whether you can use an IPV4 loopback with an IPV6 endpoint.

o01eg: Looks like IPv6 is actually dual-stack endpoint.

geoffthemedio: In that it works for both when both are present but also requires both to be present to work at all?

o01eg: Yes, it looks like it opens IPv6 port and allows IPv4 connections/packages be accepted here.

geoffthemedio:
With this applied, attempting to get info from the server over UDP crashes it, whereas it works in master.

In Python while a server is running on localhost:
````
import socket
UDP_IP = '127.0.0.1'
UDP_PORT = 12345
MESSAGE = 'CurrentTurn'.encode('utf-8')
s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.sendto(MESSAGE, (UDP_IP, UDP_PORT))
data, addr = s.recvfrom(1024, 0)
print(data)
````

**Comment by o01eg** [comment]
@geoffthemedio  I've missed point `Listen` shouldn't connect again. Fixed it.

File: network/ServerNetworking.cpp [hunk 4]
```diff
@@ -30,8 +30,35 @@ namespace {
 class DiscoveryServer {
 public:
     DiscoveryServer(boost::asio::io_context& io_context) :
-        m_socket(io_context, udp::endpoint(udp::v6(), Networking::DiscoveryPort()))
-    { Listen(); }
+        m_service(io_context),
```
geoffthemedio: `m_service` is only used in the constructor, so does it need to be stored as a member reference?

PR #2651, opened 2019-11-20
201911 fix basic hull descriptions

agrrr3:
Small hull description was wrong after the fuel efficiency changes.

[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11245&start=15#p98293)


geoffthemedio:
Can you do this with just description fixes and not also the balance changes?

agrrr3:
> Can you do this with just description fixes and not also the balance changes?

Sorry, I branched from the current workspace instead of freeorion/master.

Automatic failure is a cryptic timeout as usual.

Issue #1583, opened 2017-05-24
Add Expandable Combat Forces Breakdown to Combat Log


geoffthemedio:
In the combat log, it's a bit confusing at first to see ships mentioned as doing thing, without being first listed as participating in the battle. For the initial and lost forces numbers, a clickable expand button could be added that would show a list of all the relevant ships when expanded, similar to the expandable details about individual attacks in the log.
![combat_log_forces_breakdown](https://cloud.githubusercontent.com/assets/11386908/26398088/b9f06bf2-4078-11e7-9d07-4cd23900d0fa.png)

mmoderau:
I have a *WIP* on this, but could really use a save game with interesting enough battle to test this out (combat with multiple factions, planets, ships, casualties, etc). Should I just go and ask on the forums about it?

I don't really have time to try and create such scenario myself.

mmoderau:
Also, to make this look good, I would need to use the string formatting functions, that are in `CombatEvents.cpp` now.. and marked with a **TODO** to move them somewhere else.

Should I create a a separate MR for this before I create a MR for #1583?

geoffthemedio:
This save as a hopefully interesting combat sitrep:
![combat_sitreps](https://user-images.githubusercontent.com/11386908/68850876-16e30980-06d5-11ea-8c14-3b27c012034a.png)
[combat_sitreps.zip](https://github.com/freeorion/freeorion/files/3846040/combat_sitreps.zip)

The save also has some ship design errors during loading, but hopefully that's not a problem.

geoffthemedio:
I'd just implement your feature, then worry about code reorganization later.

mmoderau:
> This save as a hopefully interesting combat sitrep:

Woah, that's precisely what I needed, thank you!

> I'd just implement your feature, then worry about code reorganization later.

This will mean literally copying and pasting code from one anonymous name space to another in different file.

I guess I can mark the copied code with a `//TODO` comment...

mmoderau:
Here's my WIP using your savegame:
![obraz](https://user-images.githubusercontent.com/11045567/68891424-15d7c980-0721-11ea-85c3-1f347e225f6b.png)

*TODO:*
- Sort by object type (planets, ships) or divide into additional accordion subsections
- Sort by name
- Linkify

mmoderau:
Also, I'll add another line break before "Lost:".

mmoderau:
[...]

PR #2654, opened 2019-11-23
Added combat forces accordion panels

mmoderau:
Added expandable accordion panels for each empire that has forces in combat.

When expanded, the panel lists combat forces grouped by object type and sorted by displayed name and ID.

### DIFF
--- a/UI/CombatReport/CombatLogWnd.cpp
+++ b/UI/CombatReport/CombatLogWnd.cpp
@@ -66,52 +66,145 @@ public:
 };
 
 namespace {
-    typedef std::shared_ptr<LinkText> LinkTextPtr;
 
-    const std::string EMPTY_STRING;
+    // TODO: Function adapted from CombatEvents.cpp, will need to be extracted to a common library
+    const std::string& LinkTag(UniverseObjectType obj_type) {
+        static const std::string EMPTY_STRING("");
 
-    // Counts how many ships or populated planets in \a objects are owned by
-    // each empire. Assumes ownership and populations now are consistent with
-    // whenever the result is supposed to indicate, not accounting for 
-    std::map<int, int> CountByOwner(const std::set<int>& owners, const std::set<int>& objects) {
-        std::map<int, int> objects_per_owner;
-        for (int owner_id : owners)
-            objects_per_owner[owner_id] = 0;
-        for (int obj_id : objects) {
-            auto object = Objects().Object(obj_id); // gets destroyed objects, so this function can be used for initial combat forces even if some are destroyed during the combat
-            if (object && (
-                    object->ObjectType() == OBJ_SHIP || (
-                        object->GetMeter(METER_POPULATION) &&
-                        object->InitialMeterValue(METER_POPULATION) > 0.0f)))
-            {
-                int owner_id = object->Owner();
-                if (!objects_per_owner.count(owner_id))
-                    objects_per_owner[owner_id] = 0;
-                ++objects_per_owner[owner_id];
-            }
+        switch (obj_type) {
+        case OBJ_SHIP:
+            return VarText::SHIP_ID_TAG;
+        case OBJ_FLEET:
+            return VarText::FLEET_ID_TAG;
+        case OBJ_PLANET:
+            return VarText::PLANET_ID_TAG;
+        case OBJ_BUILDING:
+            return VarText::BUILDING_ID_TAG;
+        case OBJ_SYSTEM:
+            return VarText::SYSTEM_ID_TAG;
+        case OBJ_FIELD:
+        case OBJ_FIGHTER:
+        default:
+            return EMPTY_STRING;
         }
-        return objects_per_owner;
     }
 
-    // converts input empire ids and numbers of objects to text with names and
-    // links to empires and numbers-as-text of how many objects they own
-    std::string CountsToText(const std::map<int, int>& count_per_empire,
-                             const std::string& delimiter = ", ")
-    {
-        std::stringstream ss;
-        for (auto it = count_per_empire.begin(); it != count_per_empire.end(); ) {
-            std::string owner_string;
-            if (const Empire* owner = GetEmpire(it->first))
-                owner_string = GG::RgbaTag(owner->Color()) + "<" + VarText::EMPIRE_ID_TAG + " "
-                    + std::to_string(owner->EmpireID()) + ">" + owner->Name()
-                    + "</" + VarText::EMPIRE_ID_TAG + ">" + "</rgba>";
-            else
-                owner_string = GG::RgbaTag(ClientUI::DefaultLinkColor()) + UserString("NEUTRAL")  + "</rgba>";
-            ss << owner_string << ": " << it->second;
-            ++it;
-            if (it != count_per_empire.end())
-                ss << delimiter;
+    // TODO: Function adapted from CombatEvents.cpp, will need to beextracted to a common library
+    std::string WrapWithTagAndId(const std::string& meat, const std::string& tag, int id) {
+        return boost::str(boost::format("<%1% %2%>%3%</%1%>") % tag % id % meat);
+    }
+
+    /// Returns true, if the \a object should be counted in initial and/or destroyed empire forces
+    bool ShouldObjectBeShownInEmpireForces(std::shared_ptr<UniverseObject> object) {
+        return object && (object->ObjectType() == OBJ_SHIP ||
+                          (object->GetMeter(METER_POPULATION) &&
+                           object->InitialMeterValue(METER_POPULATION) > 0.0f));
+    }
+
+    using EmpireForces = std::vector<std::shared_ptr<UniverseObject>>;
+
+    /// Selects ships and populated planets in \a objects and stores them in collections
+    /// for each empire.
+    std::map<int, EmpireForces> ForcesByOwner(const std::set<int>& owners,
+                                              const std::set<int>& objects) {
+        std::map<int, EmpireForces> forces_by_owner;
+
+        for (int object_id : objects) {
+            // gets destroyed objects, so this function can be used for initial combat forces even
+            // if some are destroyed during the combat
+            auto object = Objects().Object(object_id);
+            if (ShouldObjectBeShownInEmpireForces(object)) {
+                forces_by_owner[object->Owner()].push_back(object);
+            }
         }
+
+        return forces_by_owner;
+    }
+
+    std::string EmpireIdToText(int empire_id) {
+        if (const Empire* empire = GetEmpire(empire_id))
+            return GG::RgbaTag(empire->Color()) + "<" + VarText::EMPIRE_ID_TAG + " " +
+                   std::to_string(empire->EmpireID()) + ">" + empire->Name() + "</" +
+                   VarText::EMPIRE_ID_TAG + ">" + "</rgba>";
+        else
+            return GG::RgbaTag(ClientUI::DefaultLinkColor()) + UserString("NEUTRAL") + "</rgba>";
+    }
+
+    /// converts to "Empire_name: n" text
+    std::string CountToText(int empire_id, int forces_count) {
+        std::stringstream ss;
+        ss << EmpireIdToText(empire_id) << ": " << forces_count;
+        return ss.str();
+    }
+
+    /// Used to sort empire foces for display
+    class ForcesSorter {
+    public:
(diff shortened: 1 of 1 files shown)


UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -66,52 +66,145 @@ class CombatLogWnd::Impl {
...
+    /// Used to sort empire foces for display
+    class ForcesSorter {
+    public:
+        ForcesSorter(int viewing_empire_id_)
+            : viewing_empire_id(viewing_empire_id_) {
+            }
+                    
+        bool operator()(const std::shared_ptr<UniverseObject>& lhs,
+                        const std::shared_ptr<UniverseObject>& rhs) {
+            if (TypeComparison(lhs->ObjectType(), rhs))
```
geoffthemedio: The `TypeComparison` function seems unnecessary and hard to read... Why not just:
````
if (lhs->ObjectType() < rhs->ObjectType())
    return true;
if (rhs->ObjectType() < lhs->ObjectType())
    return false;
````
or
````
if (lhs->ObjectType() != rhs->ObjectType())
    return lhs->ObjectType() < rhs->ObjectType();
````

File: UI/CombatReport/CombatLogWnd.cpp [hunk 2]
```diff
@@ -66,52 +66,145 @@ class CombatLogWnd::Impl {
...
+                return true;
+
+            if (TypeComparison(rhs->ObjectType(), lhs))
+                return false;
+
+            const std::string lhs_public_name = lhs->PublicName(viewing_empire_id);
```
geoffthemedio: `PublicName` returns a const reference so you should probably store it in one.

UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -66,52 +66,145 @@ class CombatLogWnd::Impl {
...
+            const std::string rhs_public_name = rhs->PublicName(viewing_empire_id);
+            if (lhs_public_name < rhs_public_name)
```
geoffthemedio: For string comparisons, you should use a comparison operator that is aware of UTF8 encoding / wide chars, eg. like: https://github.com/freeorion/freeorion/blob/master/UI/ObjectListWnd.cpp#L1862

UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -66,52 +66,145 @@ class CombatLogWnd::Impl {
...
+            return lhs->ID() < rhs->ID();
+        }
+
+        static bool TypeComparison(UniverseObjectType type,
+                                   const std::shared_ptr<UniverseObject>& object) {
+            return type < object->ObjectType();
+        }
+
+    private:
+        int viewing_empire_id;
```
geoffthemedio: should default initilalize with `ALL_EMPIRES`

UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -187,6 +280,79 @@ namespace {
...
+    /// and can be expanded to enumerate all forces of this empire that take part in the combat.
+    class EmpireForcesAccordionPanel : public AccordionPanel {
+    public:
+        EmpireForcesAccordionPanel(GG::X w,
+                                   CombatLogWnd::Impl& log_,
+                                   int viewing_empire_id_,
+                                   int empire_id,
+                                   EmpireForces forces_);
+
+        ~EmpireForcesAccordionPanel();
```
geoffthemedio, line 306: should probably be virtual (even though equivalent code doesn't do so...)

UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -66,52 +66,145 @@ class CombatLogWnd::Impl {
...
+
+    // TODO: Function adapted from CombatEvents.cpp, will need to beextracted to a common library
+    std::string WrapWithTagAndId(const std::string& meat, const std::string& tag, int id) {
+        return boost::str(boost::format("<%1% %2%>%3%</%1%>") % tag % id % meat);
+    }
+
+    /// Returns true, if the \a object should be counted in initial and/or destroyed empire forces
+    bool ShouldObjectBeShownInEmpireForces(std::shared_ptr<UniverseObject> object) {
+        return object && (object->ObjectType() == OBJ_SHIP ||
+                          (object->GetMeter(METER_POPULATION) &&
```
geoffthemedio: Not sure about the test being if an object has population > 0

More relevant would probably be something similar to the tests used during combat for incapacitation of planets: https://github.com/freeorion/freeorion/blob/master/combat/CombatSystem.cpp#L1161

UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -66,52 +66,145 @@ class CombatLogWnd::Impl {
...
+                           object->InitialMeterValue(METER_POPULATION) > 0.0f));
+    }
+
+    using EmpireForces = std::vector<std::shared_ptr<UniverseObject>>;
```
geoffthemedio: Calling it `EmpireForces` when it could also be non-empire ships / planets is perhaps misleading?

UI/CombatReport/CombatLogWnd.cpp
```diff
@@ -66,52 +66,145 @@ class CombatLogWnd::Impl {
...
+
+    /// Selects ships and populated planets in \a objects and stores them in collections
+    /// for each empire.
+    std::map<int, EmpireForces> ForcesByOwner(const std::set<int>& owners,
```
geoffthemedio: The `owners` parameter appears to be unused.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: This was the initial implementation, but the partitioning code later required the same criteria to be used in `ForcesToText`.

```c++
                std::upper_bound(first_of_type, forces_sorted.end(), (*first_of_type)->ObjectType(),
                                 ForcesSorter::TypeComparison);
```

For now it's just comparing types by numerical value, but should another order be desired (planets first, for example), then one change will apply to both sorting and partitioning.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: OK, I missed that. I assumed it is returning by value.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: Indeed. Will do.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: You mean the constructor parameter? There's no point in defaulting the field if there's no default constructor.

mmoderau: I should make the constructor explicit, while I'm at it.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: I'm fairly sure, that destructors of classes derived from types with virtual destructors have automatically virtual destructors too, but I'll check with the standard later.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: That's the condition that was used before. I just extracted it into a separate function.

UI/CombatReport/CombatLogWnd.cpp
geoffthemedio: I also think they are, but marking the derived classes is sometimes done for documentation reasons, I gather...

UI/CombatReport/CombatLogWnd.cpp
geoffthemedio: The one param as object and the other as the ObjectType still seems odd, though...

mmoderau: I agree. As I will be making some changes either way, I'll create a class or function that performs sorting and partitioning in one place.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: Is there a domain-specific term? "Side" (empire or unowned)?

UI/CombatReport/CombatLogWnd.cpp
mmoderau: Yes. It was there already when I started refactoring. I should be paying more attention.

mmoderau:
@geoffthemedio, should I add a commit to this PR, or create a new, squished commit and a new PR?

geoffthemedio:
add one or more commits to this

UI/CombatReport/CombatLogWnd.cpp
geoffthemedio: I don't have a good suggestion for such a name. Normally I'd not bother with the typedef at all...  or maybe call it ObjectVec or similar...

UI/CombatReport/CombatLogWnd.cpp
mmoderau: Either way, I was mimicking local code style (for lack of better references). If you believe, that `virtual` should be always explicit for virtual destructors, I will make it so.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: OK, I'll see if the type alias will be necessary after I make other changes.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: Actually, the "forces" is a bit ambiguous. Does it aim to enumerate all "active" forces (that is, able to fire), all objects that could be potentially fired upon (all owned planets and ships)?

I fail to see, how I could reuse the code you linked as a general condition over all objects in the system. There's a call to `ObjectCanAttack`, for example. Should I check if any object in the system *can potentially attack* a planet when deciding if that planet should be counted among its owner's forces?

geoffthemedio: The main thing I meant to point to was the tests of various other meters to decide if a planet is incapacitated:
````
                if (target->CurrentMeterValue(METER_SHIELD) <= 0.0f &&
                    target->CurrentMeterValue(METER_DEFENSE) <= 0.0f &&
                    target->CurrentMeterValue(METER_CONSTRUCTION) <= 0.0f)
````
It seemed like the same criteria for deciding that a planet is no longer a valid attacker could be used for marking the planet as "lost" or "disabled" in the summary.

UI/CombatReport/CombatLogWnd.cpp
geoffthemedio: It's not very important... Since there are other cases of similar code that don't, then being consistent with the surrounding style is fine.

UI/CombatReport/CombatLogWnd.cpp
mmoderau: The point is, that the label is "Lost" - in case of colonies I would consider only colonies that were completely destroyed (ownership lost) to be "lost".

If you want, I *can* change the forces summary to contain separate sections for "Ships lost" and "Colonies disabled" (or incapacitated).

This will require more changes, including string tables.

-

So far I tried to preserve the object sets reported, just increase verbosity of the report. If you want to change the meaning of the report or the object sets reported - sure, I can do that. I would really like acceptance criteria on the functional side.

I haven't played FO enough to have my own opinion on this topic.

mmoderau: I'd also like to point out, that "Lost" objects are populated from `CombatLog::destroyed_object_ids` . I'm not sure it contains (or should contain, for that matter) incapacitated planets.

mmoderau: @geoffthemedio , I didn't have the time to revisit this since, but I think I'll refactor this to a more easily extendable form and leave the current (previous) criteria.

I'll make it so that the criteria/groups can be changed in future iterations.

mmoderau:
I improved the implementation a bit. Now it will be easier to adjust, what you want to see in the accordion panels.

geoffthemedio:
Sorry for the delay in following up on this... holiday travel and imminent release issues have taken priority.

mmoderau:
@geoffthemedio, I figured. No problem. :grin:

Vezzra:
~~With this merged, can we close #1583 as resolved?~~

EDIT: I just saw that #1583 apparently isn't completely implemented yet.

PR #2661, opened 2019-12-02
Ship weapons target dangerous, vulnerable, damaged ships first

agrrr3:
This is WIP and mostly for furthering discussion.

[forum post by telos](https://www.freeorion.org/forum/viewtopic.php?p=98396#p98396)

For all ship weapons (including Bombers and monster weapons) but flak cannon:
Targeting a number of enemy vessels. Possible targets are fighters (10% chance of being included) and N most-juicy ships.

N is an approximation of the total number of shots shot by the attacking fleet.

most-juicy targets are ships which have a high Attack value and low structure (Attack divided by structure). 
Also damaged ships ared juicy: Half as juicy as attack value is the number of already lost structure. 

There is also a small change which makes interceptor targeting less rigid.

Implemented but untested. No idea if this works.

Play-Testing wanted if this is more suitable than the current targeting scheme.

### DIFF
--- a/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
+++ b/default/scripting/ship_parts/FighterHangar/FT_HANGAR_1.focs.txt
@@ -6,14 +6,19 @@ Part
     capacity = 4
     damage = 1
     combatTargets = OrderedAlternativesOf [
-        // Target Bombers and Heavy Bombers first
+        // Rather target Bombers and Heavy Bombers
         And [
             [[COMBAT_TARGETS_VISIBLE_ENEMY]]
             Fighter
             Or [
                DesignHasPart name = "FT_HANGAR_3"
                DesignHasPart name = "FT_HANGAR_4"
+               // Low chance to intercept other fighters. Higher if few bombers present
+               Random probability = 0.1
             ]
+            // If there are few bombers intercept other fighters
+            // NumberOf doesnt work for fighters, also number should be fleet dependent
+            NumberOf number = 3 condition = Fighter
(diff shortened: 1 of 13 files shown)


geoffthemedio:
Is this relevant after https://github.com/freeorion/freeorion/pull/2665 ?

Oberlus:
> Is this relevant after #2665 ?

It is probably relevant nevertheless. The focus of this PR is on (or could develop into) using different targetting logics away from the type of target: target first the ships that will take less effort to destroy and/or that will save more damage on allied ships if quickly destroyed.

It could be the foundation for computer targetting parts.

The problem with all this is it introduces a more complex targetting system that could work as intended (and be an advantage for fleets that use it, call them "attackers") or become a liability if the enemy ("defenders") could exploit it by carefully balancing the number and type of its own assets with respect to the attackers fleet composition in order to make the attacker concentrate many shots in very few defender targets, allowing this way a huge advantage for the defender in the form of very cheap damage soakers.

Out of KISS principle, I'm wary of using it (despite I was one of the proponents), and I'm quite happy with current targetting system (despite I having serious loses in current multiplayer game due to my enemy being lucky enough to lose 80% of total fleet structure and not a single ship, I would have enjoyed a targetting scheme that prioritises heavily damaged armed ships on second bout).

All in all, I would not close this.

agrrr3:
I think we can leave this as a draft. The topic will certainly come up again. We will direct the poster to try this code. If the poster does not even volunteer to try we can probably ignore the post.

If we get playtesting feedback it would be valuable into furthering the discussion.

PR #2664, opened 2019-12-03
lower the min unfocused FPS to 0.1

l29ah:
It lowered the CPU consumption considerably on my laptop compared with 1.
Couldn't test with smaller values because of the rounding in the UI.

### DIFF
--- a/UI/ClientUI.cpp
+++ b/UI/ClientUI.cpp
@@ -481,7 +481,7 @@ namespace {
         db.Add("video.fps.max.enabled",                                 UserStringNop("OPTIONS_DB_LIMIT_FPS"),                      true);
         db.Add("video.fps.max",                                         UserStringNop("OPTIONS_DB_MAX_FPS"),                        60.0,                           RangedStepValidator<double>(1.0, 0.0, 240.0));
         db.Add("video.fps.unfocused.enabled",                           UserStringNop("OPTIONS_DB_LIMIT_FPS_NO_FOCUS"),             true);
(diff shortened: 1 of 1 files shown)


Oberlus:
I got this when starting freerion after last update, and it seems related to this PR:

`[0x00007f018687b8c0] [info]    OptionsDB::SetFromXMLRecursive() : while processing config.xml the following exception was caught when attempting to set option "video.fps.unfocused": bad lexical cast: source type value could not be interpreted as target`

Using Ubuntu 16.04.

geoffthemedio:
Does that continue happening even after deleting config.xml ?

I don't get that line, but it may be related to number conversions working differently with different compilers or standard libraries...

Oberlus:
> Does that continue happening even after deleting config.xml ?

After deleting config.xml I no longer get that message.
Then, after rearranging the windows to my like and exiting FO I get the message again when I restart FO.
The game works fine anyway and I don't feel FO is using more or less CPU than before (I mean, fans are not noisy).

geoffthemedio:
Try after https://github.com/freeorion/freeorion/commit/f15151b6961cd4f7520c6722593abce0819a2435

Oberlus:
> Try after [f15151b](https://github.com/freeorion/freeorion/commit/f15151b6961cd4f7520c6722593abce0819a2435)

That solved my problem.

Vezzra:
@geoffthemedio, I assume f15151b6961cd4f7520c6722593abce0819a2435 should be cherry picked into the release branch, too?

PR #2673, opened 2019-12-11
Suse readme

mslacken:
* added correct link


o01eg:
You could update previous PR, but OK.

PR #2678, opened 2019-12-20
Reset Rules to Defaults in Client when Returning to Intro Screen

geoffthemedio:
Resets rules to default value when returning to the intro screen menu. This prevents an issue where after playing a game with modified rules, such as a loaded save gamed, the default setting for a new game and the rule values in the intro screen pedia would be those of the previously-played game, rather than the defined rule default values.


geoffthemedio:
Probably OK to cherry pick...?

Vezzra:
Done: ba5a262bebd57ece4955a2ff85ff5877cf65dc6c

PR #2685, opened 2019-12-29
Fix spelling/grammar in pedia.

aschuch247:


default/stringtables/en.txt
```diff
@@ -12742,7 +12742,7 @@ BLD_GATEWAY_VOID
 Gateway to the Void
 
 BLD_GATEWAY_VOID_DESC
-'''The ability to close off a system almost entirely can be useful for an empire on the defensive. Any ships who enter this system without passing through on the same turn are destroyed, and all colonies within are immediately reduced to [[encyclopedia OUTPOSTS_TITLE]] status. The [[metertype METER_STEALTH]] of all objects within the system is increased by 1000, making them almost impossible to detect.
+'''The ability to close off a system almost entirely can be useful for an empire on the defensive. Any ships which enter this system without passing through on the same turn are destroyed, and all colonies within are immediately reduced to [[encyclopedia OUTPOSTS_TITLE]] status. The [[metertype METER_STEALTH]] of all objects within the system is increased by 1000, making them almost impossible to detect.
```
geoffthemedio: Should be `that` I think

aschuch247: For persons `who`, for things `which` and if you do not care or are not sure, `that`. Using `which` communicates that you are not talking about a person, but about a thing.

aschuch247: https://www.grammarly.com/blog/which-vs-that/

Well, it is a defining clause, so yes, `that` is the correct choice.

geoffthemedio: yes; I do not prefer "who" over "which".

PR #2687, opened 2019-12-29
Make Chato bad pilots

### LINKED ISSUE #2680: Chato slightly OP
### Environment
* **FreeOrion Version**: v0.4.9 v0.4.9 2019-12-15

### Description
Chato has broad planet tolerance, great research (and average production), plus bad troops (both) and phototrophic.
Any species can have a bad start (little to none good/adequate planets), but this one tends to do pretty well most often. If it is lucky, it's OP; if it is unlucky, it can overcome the problem faster than other thanks to the great research: Exobots are almost on the opposite side of the environment wheel, so once it gets Xenological Genetics (access to poor) it can colonise pretty much everything.

### Suggestion
~~Remove the broad planet tolerance.~~
Add Bad Pilots to Chato.
An species with no production penalty and great research is hard to not be OP, but at least this way it's not that bad.

~~PS: I thought this is pretty much obvious (once I thought of it) and thought it would be good to post it directly as an issue. Should this kind of "issues" be posted first in the forums?~~

o01eg:
Fixes #2680

### DIFF
--- a/default/scripting/species/SP_CHATO.focs.txt
+++ b/default/scripting/species/SP_CHATO.focs.txt
@@ -6,7 +6,7 @@ Species
     CanProduceShips
     CanColonize
 
-    tags = [ "PHOTOTROPHIC" "GREAT_RESEARCH" "AVERAGE_SUPPLY" "BAD_ATTACKTROOPS" "PEDIA_PHOTOTROPHIC_SPECIES_CLASS" ]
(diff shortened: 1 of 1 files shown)


Vezzra (approved):
Looks good.

geoffthemedio:
I have no useful opinion. If there is vague consensus amongst regular players who commented, go ahead.

PR #2688, opened 2020-01-01
Add Gamma Burst Slingshot SR weapon

agrrr3:
Add gamma burst slingshot ship weapon part
 
Designed to be cost efficient at killing decoys.
[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=15&t=11473&sid=b70e8c0049338e9d9894b6fea5b6435f)

In absense of shields this weapon can get almost as good as Plasma. Against shields its pretty useles

Fluff:
Reactor energy is used to excite matter inside a warhead, pumps it up into a semistable state.
A missile stabilizes the state and delivers the warhead into combat range, pointing into direction of the target.
When the missile is coming into point defense range, the warhead gets triggered.
Stabilization ceases, the matter falls back into stable state, releasing gamma rays in direction of the target.

Game mechanic/Balance reasoning:
// BURST_1   12RP  4T  allow 40PP part unlock with 3 shots with 2 damage - comparable to two MD_1, but 50% more shots
// BURST_2  120RP  8T  upgrade to 3 shots with 4 damage - comparable damage to two MD_4 at the same PP cost
// BURST_3  720RP 12T  upgrade to 3 shots with 7 damage - comparable damage and cost to P_5; comparable to one L_4 against shields-3 (but 33% more expensive)
// No xx_WEAPON pilot bonus to damage for balance reasons (also fits the fluff, as this is automated/missile based)
// Weapon seems like a good choice to support invasion

// Maybe we should add a MD_5 at 100RP (which would total damage at 7) as a slight buff to mass drivers

// It should be able to support MD invasion - should at least double the shots at the same price point(?). (Early comsats)
// Upgrading it should bring the comsat downing to late comsats.

// Increasing shots with steady 3 damage is plasma gun level.
The usual level bonus for pilot skill does not work so well here.
The normal appliance would make great pilots completely OP (+6 effective damage)
One could link great pilot bonus to BURST_2 and good pilot bonus to BURST_3 which would provide an acceptable nerf and delay.
But that would be hard to communicate


o01eg:
I suppose this PR shouldn't contain networking commit.

geoffthemedio:
Discussion appears to be ongoing on the forums, so letting this sit...

agrrr3:
@freeorion/ai-team how do make AI pick up the weapon? I registered it in AIDependencies. it should be a good choice for bad_pilots at least. I started a with halved PP cost, so the weapon is definitly doing more total damage than mass driver at the same cost. But AI would not research the tech and not build ships with it

hm that tag does not look like it works

Morlic-fo:
> AI would not research the tech

You need to add the tech at an appropriate location in this file:
https://github.com/freeorion/freeorion/blob/master/default/python/AI/TechsListsAI.py

agrrr3:
> > AI would not research the tech
> 
> You need to add the tech at an appropriate location in this file:
> https://github.com/freeorion/freeorion/blob/master/default/python/AI/TechsListsAI.py

Ok, that did (kind of) work. As far as I get it the current research approach are the blocks and tiers of predefined TechLists and some fast tracking in ResearchAI. I did not see anything to delay/prevent research of certain techs. As the gamma burst slingshot gives an alternative to the normal weapon line tree this means a lot of waste of research (like researching all the hull lines). Gysache should go for GBS-1 and basic fighters, instead of MD-4 and Lasers it could/should go for GBS-2. Probably then it should research lasers so it is able to upgrade to plasma weapons fast. If the enemy uses lots of shielded ships or there is at least a good pilot species which can produce ships, plasma should be researched and probably upgraded; if there is still GBS fleet around research GBS-3. Not sure what kind of framework would be good to address that.. I was thinking about using something like fuzzy logic.

agrrr3:
I started a [forum thread](https://www.freeorion.org/forum/viewtopic.php?f=9&t=11508&p=99229#p99229) about the AI. This will probably lead to a followup PR.

agrrr3:
Any reason not to merge this?

geoffthemedio:
What's the current state of discussion about it? As commented above, last I checked, it was still ongoing.

Oberlus:
> What's the current state of discussion about it? As commented above, last I checked, it was still ongoing.

The numbers seem to be right for initial testing, once the base cost of the part is changed to 40 PP (as per [this](https://www.freeorion.org/forum/viewtopic.php?f=15&t=11473&p=99839#p99829)).

My only concerns now regard the fluff, commented [here](https://www.freeorion.org/forum/viewtopic.php?f=15&t=11473&p=99853#p99853).

agrrr3:
Fluff is fixed, did a smoketest after last change.

Oberlus:
It would be great to test this in next multiplayer game.

geoffthemedio:
So is the design / testing for this settled now?

Oberlus:
> So is the design / testing for this settled now?

No testing done from my part (and [I won't be able to do it](https://www.freeorion.org/forum/viewtopic.php?f=24&t=11581#p100184) any time soon).

agrrr3:
Cost is back at 40PP as requested. Should be good to merge.

Oberlus:
Design discussion is over and implementation looks good to merge.

PR #2689, opened 2020-01-03
Supply Obstruction Update Before Fleet Movement

### LINKED ISSUE #2675: Incorrect pass through
Enemy stealthy 5 fleet has passed through my ship, without triggering a combat.
My version is : c0eb3bb, win32 version
[Save](https://github.com/jinlan/issue/raw/master/FreeOrion_incorrect_pass_through_jinlanid.7z)

Don't just press "end turn" after load this save, order the robotic hull ship to stand still, then end turn.

geoffthemedio:
Updates after orders, gifting, invasion, colonization, scrapping, and bombardment.

May fix #2675


Vezzra:
> May fix #2675

It does on macOS. Didn't notice any side effects when testing, but did only a quick test (build and run the provided save).

geoffthemedio:
probably no better way to get testing than to ship it.

PR #2694, opened 2020-01-06
Use boost signals to handle context menu in StatisticIcon

adrianbroher:
Instead of using the GG filter event mechanism to handle the context menu of the various information panels, this PR now uses boost signals and callbacks.  This removes the need to find the associated context menu destination via some id lookup and simplifies the context menu callback.

### DIFF
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -1080,8 +1080,6 @@ public:
     mutable boost::signals2::signal<void (const std::vector<int>&)> NewFleetFromShipsSignal;
 
 protected:
-    bool EventFilter(GG::Wnd* w, const GG::WndEvent& event) override;
-
     void DropsAcceptable(DropsAcceptableIter first, DropsAcceptableIter last,
                          const GG::Pt& pt, GG::Flags<GG::ModKey> mod_keys) const override;
 
@@ -1339,64 +1337,6 @@ void FleetDataPanel::SizeMove(const GG::Pt& ul, const GG::Pt& lr) {
     //std::cout << "FleetDataPanel::SizeMove new size: (" << Value(Width()) << ", " << Value(Height()) << ")" << std::endl;
     if (old_size != Size())
         DoLayout();
-}
-
-bool FleetDataPanel::EventFilter(GG::Wnd* w, const GG::WndEvent& event) {
-    //std::cout << "FleetDataPanel::EventFilter " << EventTypeName(event) << std::endl << std::flush;
-
-    if (w == this || !w)
-        return false;
-
-    switch (event.Type()) {
-    case GG::WndEvent::RClick: {
-        MeterType meter_type = INVALID_METER_TYPE;
-        for (const auto& meter_type_stat_icon_pair : m_stat_icons) {
-            if (!meter_type_stat_icon_pair.second || meter_type_stat_icon_pair.second.get() != w)
-                continue;
-            meter_type = meter_type_stat_icon_pair.first;
-            break;
-        }
-        if (meter_type == INVALID_METER_TYPE)
-            return false;
-
-        std::string meter_string = boost::lexical_cast<std::string>(meter_type);
-        std::string meter_title;
-        if (UserStringExists(meter_string))
-            meter_title = UserString(meter_string);
-
-        bool retval = false;
-        auto zoom_article_action = [&retval, &meter_string]() { retval = ClientUI::GetClientUI()->ZoomToMeterTypeArticle(meter_string);};
-
-        auto pt = event.Point();
-        auto popup = GG::Wnd::Create<CUIPopupMenu>(pt.x, pt.y);
-
-        if (!meter_title.empty()) {
-            std::string popup_label = boost::io::str(FlexibleFormat(UserString("ENC_LOOKUP")) %
-                                                                    meter_title);
-            popup->AddMenuItem(GG::MenuItem(popup_label, false, false, zoom_article_action));
-        }
-        popup->Run();
-
-        return retval;
-        break;
-    }
-
-    case GG::WndEvent::DragDropEnter:
-    case GG::WndEvent::DragDropHere:
-    case GG::WndEvent::CheckDrops:
-    case GG::WndEvent::DragDropLeave:
-    case GG::WndEvent::DragDroppedOn:
-        if (w == this) {
-            ErrorLogger() << "FleetDataPanel::EventFilter w == this";
-            return false;
-        }
-        HandleEvent(event);
-        return true;
-        break;
-
-    default:
-        return false;
-    }
 }
 
 void FleetDataPanel::ToggleAggression() {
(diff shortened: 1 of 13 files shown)


UI/PopulationPanel.cpp
```diff
@@ -70,7 +70,30 @@ void PopulationPanel::CompleteConstruction() {
...
-        meter_stat.second->InstallEventFilter(shared_from_this());
+        meter_stat.second->RightClickedSignal.connect([this, &meter_stat](const GG::Pt& pt) {
+            MeterType meter_type = meter_stat.first;
+
+            std::string meter_string = boost::lexical_cast<std::string>(meter_type);
+
+            auto popup = GG::Wnd::Create<CUIPopupMenu>(pt.x, pt.y);
+            auto pc = this->GetPopCenter();
+            if (meter_type == METER_POPULATION && pc) {
+                std::string species_name = pc->SpeciesName();
```
geoffthemedio, line 81: Zoom to species doesn't seem to work in my tests. Quickstarted a game, right click the population icon on my planet and get a two-command popup, pick "Pedia Lookup Human" but nothing happens. Other meter pedia lookup commands do work, though.

UI/FleetWnd.cpp
```diff
@@ -978,7 +977,18 @@ namespace {
             auto icon = GG::Wnd::Create<StatisticIcon>(entry.second, 0, 0, false, StatIconSize().x, StatIconSize().y);
             m_stat_icons.push_back({entry.first, icon});
             AttachChild(icon);
-            icon->InstallEventFilter(shared_from_this());
+            icon->RightClickedSignal.connect([entry](const GG::Pt& pt) {
+                std::string meter_string = boost::lexical_cast<std::string>(entry.first);
+
+                auto zoom_article_action = [&meter_string]() { ClientUI::GetClientUI()->ZoomToMeterTypeArticle(meter_string);};
```
geoffthemedio: This results in double popup menus from both the icon and the containing fleet panel. First the icon Pedia Lookup X command appears, then after that's closed, a list of fleet commands appears. The latter shouldn't appear if the click was on the icon.

UI/FleetWnd.cpp
```diff
@@ -1842,7 +1756,20 @@ void FleetDataPanel::Init() {
             m_stat_icons.push_back({std::get<0>(entry), icon});
             icon->SetBrowseModeTime(tooltip_delay);
             icon->SetBrowseText(UserString(std::get<2>(entry)));
-            icon->InstallEventFilter(shared_from_this());
+            icon->RightClickedSignal.connect([entry](const GG::Pt& pt){
+                MeterType meter_type = std::get<0>(entry);
+
+                std::string meter_string = boost::lexical_cast<std::string>(meter_type);
+
+                auto zoom_article_action = [&meter_string]() { ClientUI::GetClientUI()->ZoomToMeterTypeArticle(meter_string);};
```
geoffthemedio: same for this... double popups

UI/PopulationPanel.cpp
geoffthemedio: OK, not sure if this issue is in the PopulationPanel or the MultiIconValueIndiactor. It happens when the population panel is NOT expanded, but it works fine when it is expanded.

adrianbroher: Cannot reproduce.  Seed 0, Race human, quick start game. German localization.

UI/FleetWnd.cpp
adrianbroher: Confirmed.

UI/FleetWnd.cpp
adrianbroher: Confirmed.

adrianbroher: Fixed via 4574e50ccff2134536e036a3478bec08c7cfd80e

UI/FleetWnd.cpp
adrianbroher: Fixed via 4574e50ccff2134536e036a3478bec08c7cfd80e

UI/PopulationPanel.cpp
geoffthemedio: To illustrate the issue, I'm trying to use this popup command:

FailedLookupPopup

Switching to the de stringtable doesn't change the result.

adrianbroher: Still not reproducible, still works as expected.

UI/PopulationPanel.cpp
```diff
@@ -70,7 +70,30 @@ void PopulationPanel::CompleteConstruction() {
...
+                if (!species_name.empty()) {
+                    auto zoom_species_action = [&species_name]() { ClientUI::GetClientUI()->ZoomToSpecies(species_name); };
```
geoffthemedio: Seemingly the problem is here, when capturing `species_name` by reference, not by value. It's a local, so is probably out of scope and/or reset to an empty string by a later loop iteration by the time the lambda gets executed.

Removing the `&` from the capture fixes it for me.

geoffthemedio: should probably do all the captures in these loops by value...?

adrianbroher: Updated PR according to suggestion.

UI/PopulationPanel.cpp
```diff
@@ -70,7 +70,30 @@ void PopulationPanel::CompleteConstruction() {
...
+                std::string species_name = pc->SpeciesName();
+                if (!species_name.empty()) {
+                    auto zoom_species_action = [species_name]() { ClientUI::GetClientUI()->ZoomToSpecies(species_name); };
+                    std::string species_label = boost::io::str(FlexibleFormat(UserString("ENC_LOOKUP")) %
+                                                                              UserString(species_name));
+                    popup->AddMenuItem(GG::MenuItem(species_label, false, false, zoom_species_action));
+                }
+            }
+
+            auto pedia_meter_type_action = [&meter_string]() { ClientUI::GetClientUI()->ZoomToMeterTypeArticle(meter_string); };
```
geoffthemedio: still by reference here intentionally?

adrianbroher: No, that was an oversight.  Fixed via 17c6807c8d6a026690322d74e173afd3a2a359a3

geoffthemedio:
I think this is causing another issue: drag-dropping onto statistic icons on fleet panels now doesn't work. It's a bit hard to tell because the drag-over highlighting is a bit quirky in master... In master, as far as I can tell, the first statistic icon one drags another fleet panel onto will cause the whole drag-target fleet panel to highlight, but dragging over another statistic icon during the same drag will not. Regardless, dropping seems to still work regardless of whether it's over a statistic icon or the base fleet panel. But with this pull request applied, I can't drop a fleet onto a statistic icon of the target fleet panel; it doesn't execute the drop and fleet merge, and there is also no highlighting of the drop target when dragging over the first (or any) statistic icons of the target fleet panel.

Comparisons in screenshots...
Fleet Panel Statistic Icon Drag Highlighting

adrianbroher:
Bah, the signal delegation in GG sucks.  Added forwarding of drag related signals to StatisticIcon as stop gap for now.

PR #2698, opened 2020-01-09
Add CombatBout ValueRef

agrrr3:
You can use this CombatBout ValueRef in combatTargets conditions to query the current combat bout.
E.g. (CombatBout == 1) will match all objects if combat is in the first bout

The ScriptingContext constructor which is used only in combat gets the current bout passed as a new parameter.

### DIFF
--- a/universe/ScriptingContext.h
+++ b/universe/ScriptingContext.h
@@ -2,7 +2,6 @@
 #define _ScriptingContext_h_
 
 #include "Universe.h"
-#include "../combat/CombatSystem.h"
 
 #include <boost/any.hpp>
 
@@ -10,11 +9,21 @@
 
 class UniverseObject;
 
+struct FO_COMMON_API ScriptingCombatInfo {
+    ScriptingCombatInfo() {}
+    ScriptingCombatInfo(int bout_, const Universe::EmpireObjectVisibilityMap& vis) :
+        bout(bout_),
+        empire_object_visibility(vis)
+    {}
+
+    int bout = 0;
+    Universe::EmpireObjectVisibilityMap empire_object_visibility;
+};
+
 struct ScriptingContext {
     /** Empty context.  Useful for evaluating ValueRef::Constant that don't
       * depend on their context. */
-    ScriptingContext()
-    {}
+    ScriptingContext() = default;
 
     /** Context with only a source object.  Useful for evaluating effectsgroup
       * scope and activation conditions that have no external candidates or
@@ -27,7 +36,7 @@ struct ScriptingContext {
       * conditions. Useful in combat resolution when the visibility of objects
       * may be different from the overall universe visibility. */
     ScriptingContext(std::shared_ptr<const UniverseObject> source_,
-                     const CombatInfo& combat_info_) :
+                     const ScriptingCombatInfo& combat_info_) :
         source(source_),
         combat_info(combat_info_)
     {}
@@ -85,12 +94,15 @@ struct ScriptingContext {
         current_value(current_value_)
     {}
 
+    static const ScriptingCombatInfo s_empty_combat_info;
+
     std::shared_ptr<const UniverseObject>   source;
     std::shared_ptr<UniverseObject>         effect_target;
     std::shared_ptr<const UniverseObject>   condition_root_candidate;
     std::shared_ptr<const UniverseObject>   condition_local_candidate;
     const boost::any                        current_value;
-    const CombatInfo&                       combat_info = s_empty_combat_info;
+    //
+    const ScriptingCombatInfo&              combat_info = ScriptingCombatInfo();
 };
 
 #endif // _ScriptingContext_h_
--- a/universe/ScriptingContext.h
+++ b/universe/ScriptingContext.h
@@ -9,22 +9,6 @@
 
 class UniverseObject;
 
-struct BackgroundContext {
-public:
-    BackgroundContext()
-    {}
-
-    BackgroundContext(int bout_, Universe::EmpireObjectVisibilityMap& empire_object_visibility_) :
-        bout(bout_),
-        empire_object_visibility(empire_object_visibility_)
-    {}
-
-    int bout;
-    Universe::EmpireObjectVisibilityMap empire_object_visibility;       ///< indexed by empire id and object id, the visibility level the empire has of each object.  may be increas
-    static const BackgroundContext s_empty_scripting_context_background;
-};
-const BackgroundContext BackgroundContext::s_empty_scripting_context_background = {};
-
 struct ScriptingContext {
     /** Empty context.  Useful for evaluating ValueRef::Constant that don't
       * depend on their context. */
@@ -42,9 +26,9 @@ struct ScriptingContext {
       * conditions. Useful in combat resolution when the visibility of objects
       * may be different from the overall universe visibility. */
     ScriptingContext(std::shared_ptr<const UniverseObject> source_,
-                     const BackgroundContext& background_) :
+                     const Universe::EmpireObjectVisibilityMap& visibility_map) :
         source(source_),
-        background(background_)
+        empire_object_vis_map_override(visibility_map)
     {}
 
     ScriptingContext(std::shared_ptr<const UniverseObject> source_,
@@ -70,7 +54,7 @@ struct ScriptingContext {
         condition_root_candidate(parent_context.condition_root_candidate),
         condition_local_candidate(parent_context.condition_local_candidate),
         current_value(current_value_),
-        background(parent_context.background)
+        empire_object_vis_map_override(parent_context.empire_object_vis_map_override)
     {}
 
     /** For recursive evaluation of Conditions.  Keeps source and effect_target
@@ -86,7 +70,7 @@ struct ScriptingContext {
                                             condition_local_candidate_),    // if parent context doesn't already have a root candidate, the new local candidate is the root
         condition_local_candidate(      condition_local_candidate_),        // new local candidate
         current_value(                  parent_context.current_value),
-        background(                     parent_context.background)
+        empire_object_vis_map_override( parent_context.empire_object_vis_map_override)
     {}
 
     ScriptingContext(std::shared_ptr<const UniverseObject> source_,
@@ -105,8 +89,7 @@ struct ScriptingContext {
     std::shared_ptr<const UniverseObject>   condition_root_candidate;
     std::shared_ptr<const UniverseObject>   condition_local_candidate;
     const boost::any                        current_value;
-
-    const BackgroundContext&                background = BackgroundContext::s_empty_scripting_context_background;
+    Universe::EmpireObjectVisibilityMap     empire_object_vis_map_override;
 };
 
 #endif // _ScriptingContext_h_
--- a/universe/ScriptingContext.h
+++ b/universe/ScriptingContext.h
@@ -9,6 +9,20 @@
 
 class UniverseObject;
 
+struct BackgroundContext {
+public:
+    BackgroundContext()
+    {}
+
+    BackgroundContext(int bout_, Universe::EmpireObjectVisibilityMap& empire_object_visibility_) :
+        bout(bout_),
+        empire_object_visibility(empire_object_visibility_)
+    {}
+
+    int bout;
+    Universe::EmpireObjectVisibilityMap empire_object_visibility;       ///< indexed by empire id and object id, the visibility level the empire has of each object.  may be increas
+};
+
 struct ScriptingContext {
     /** Empty context.  Useful for evaluating ValueRef::Constant that don't
       * depend on their context. */
@@ -26,9 +40,9 @@ struct ScriptingContext {
       * conditions. Useful in combat resolution when the visibility of objects
       * may be different from the overall universe visibility. */
     ScriptingContext(std::shared_ptr<const UniverseObject> source_,
-                     const boost::any& background_) :
-    source(source_)//,
-    //        background(boost::any(std::ref(background_)))
+                     const BackgroundContext& background_) :
+        source(source_),
+        background(background_)
     {}
 
     ScriptingContext(std::shared_ptr<const UniverseObject> source_,
@@ -70,7 +84,7 @@ struct ScriptingContext {
                                             condition_local_candidate_),    // if parent context doesn't already have a root candidate, the new local candidate is the root
         condition_local_candidate(      condition_local_candidate_),        // new local candidate
         current_value(                  parent_context.current_value),
-        background( parent_context.background)
+        background(                     parent_context.background)
     {}
 
     ScriptingContext(std::shared_ptr<const UniverseObject> source_,
@@ -89,7 +103,7 @@ struct ScriptingContext {
     std::shared_ptr<const UniverseObject>   condition_root_candidate;
     std::shared_ptr<const UniverseObject>   condition_local_candidate;
     const boost::any                        current_value;
-    const boost::any                        background = boost::any();
+    const BackgroundContext                 background;
 };
 
 #endif // _ScriptingContext_h_
--- a/universe/ScriptingContext.h
+++ b/universe/ScriptingContext.h
@@ -21,9 +21,9 @@ public:
 
     int bout;
     Universe::EmpireObjectVisibilityMap empire_object_visibility;       ///< indexed by empire id and object id, the visibility level the empire has of each object.  may be increas
-
+    static const BackgroundContext s_empty_scripting_context_background;
 };
-static const BackgroundContext theEmptyBackground = BackgroundContext();
+const BackgroundContext BackgroundContext::s_empty_scripting_context_background = {};
 
 struct ScriptingContext {
     /** Empty context.  Useful for evaluating ValueRef::Constant that don't
@@ -106,7 +106,7 @@ struct ScriptingContext {
     std::shared_ptr<const UniverseObject>   condition_local_candidate;
     const boost::any                        current_value;
 
-    const BackgroundContext&               background = theEmptyBackground;
+    const BackgroundContext&                background = BackgroundContext::s_empty_scripting_context_background;
 };
 
 #endif // _ScriptingContext_h_
--- a/universe/ScriptingContext.h
+++ b/universe/ScriptingContext.h
@@ -21,7 +21,9 @@ public:
 
     int bout;
     Universe::EmpireObjectVisibilityMap empire_object_visibility;       ///< indexed by empire id and object id, the visibility level the empire has of each object.  may be increas
+
 };
+static const BackgroundContext theEmptyBackground = BackgroundContext();
 
 struct ScriptingContext {
     /** Empty context.  Useful for evaluating ValueRef::Constant that don't
@@ -103,7 +105,8 @@ struct ScriptingContext {
     std::shared_ptr<const UniverseObject>   condition_root_candidate;
     std::shared_ptr<const UniverseObject>   condition_local_candidate;
     const boost::any                        current_value;
-    const BackgroundContext                 background;
+
+    const BackgroundContext&               background = theEmptyBackground;
 };
 
 #endif // _ScriptingContext_h_
--- a/universe/ScriptingContext.h
+++ b/universe/ScriptingContext.h
@@ -9,6 +9,8 @@
 
 class UniverseObject;
 
+/** combat/CombatInfo extends this ScriptingCombatInfo in order
+  * to give Conditions and ValueRefs access to combat related data */
 struct FO_COMMON_API ScriptingCombatInfo {
     ScriptingCombatInfo() {}
     ScriptingCombatInfo(int bout_, const Universe::EmpireObjectVisibilityMap& vis) :
@@ -16,9 +18,10 @@ struct FO_COMMON_API ScriptingCombatInfo {
         empire_object_visibility(vis)
     {}
 
-    int bout = 0;
-    Universe::EmpireObjectVisibilityMap empire_object_visibility;
+    int bout = 0; ///< current combat bout, used with CombatBout ValueRef for implementing bout dependent targeting. First combat bout is 1
+    Universe::EmpireObjectVisibilityMap empire_object_visibility; ///< indexed by empire id and object id, the visibility level the empire has of each object.  may be increased during battle
 };
+FO_COMMON_API const ScriptingCombatInfo EMPTY_COMBAT_INFO{};
 
 struct ScriptingContext {
     /** Empty context.  Useful for evaluating ValueRef::Constant that don't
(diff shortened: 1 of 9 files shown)


agrrr3:
@geoffthemedio Can we please have this merged to master ASAP? I need this in windows test builds to be able to further the discussion about missiles and attack ranges.

geoffthemedio:
I've been hesitating about this because I'm reluctant to add more data to `ScriptingContext`. More data in the context means every time one is created or copied - which can be quite a lot - that extra data needs to be handled, including for the majority of cases where that extra data is completely irrelevant (eg. outside of combat). It's probably not a big deal for a single int, but it starts setting a precedent that I'd rather avoid.

For this case, is it possible to look up the current combat round similar to how the current game turn is looked up, without adding it to the context? Possibly not, as it would need to be added to the app interface `IApp` and that wouldn't make sense in the case of parallel evaluation of combats (unlike the current game turn which is single-valued per server and only changes in one place...)

agrrr3:
If think the bout value is hard to share. I am not sure what other kind of data one would wish to include in the scripting context. 

(Why) is the scripting context copied/switched/changed during condition/valueRef evaluation?

Does the empire_object_vis_map_override change during condition evaluation? It could be put together with the bout value and data to come in a struct and be passed by reference. This would also make copying unnessecary.

geoffthemedio:
> If think the bout value is hard to share. I am not sure what other kind of data one would wish to include in the scripting context.

High on my list of stuff to do is to include an `ObjectMap` override, or perhaps mandatory parameter, in the `ScriptingContext`. Rather than getting objects from the main gamestate, this would be used for getting objects for condition or valueref evaluations, and for effect executions that modify objects. This would be helpful for combat, as the combat-specific object map could be passed, which would include fighters.

> (Why) is the scripting context copied/switched/changed during condition/valueRef evaluation?

A common case is when evaluating a condition that contains a valueref. If the valueref refers to `LocalCandidate` in its script, then the local candidate info in the scripting context needs to be updated for each candidate object being tested by the condition before that context is passed into the valueref's eval function. Typically that's done by constructing a new scripting context from the parent context and the local candidate object that the condition is matching.

> Does the empire_object_vis_map_override change during condition evaluation? It could be put together with the bout value and data to come in a struct and be passed by reference. This would also make copying unnessecary.

The visibility override shouldn't change during condition evaluation, and could probably be stored by reference on its own, rather than constructing an empty map each time. A CombatScriptingContext struct is also reasonable, which could be stored by reference, with the reference pointing to a static empty one in case of non-combat construction of ScriptingContext.

Oberlus:
> I've been hesitating about this because I'm reluctant to add more data to `ScriptingContext`

Combat bout is not necessary in `ScriptingContext` if we do the range and missiles implementation in the backend. And it seems necessary to do such implementation at least for missiles, because FOCS-only isn't satisfactory enough.

Related forum posts:
[General discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11474&p=99320&sid=7f002287f6c595504ec91730bb2dbf1d#p99312)
[Specific to backend](https://www.freeorion.org/forum/viewtopic.php?f=9&t=11518&sid=7f002287f6c595504ec91730bb2dbf1d)

The same that the backend makes a special treatment for fighters (so that, e.g., they are not launched on the last bout), it can make special treatment for each type of ranged weapon.
In the FOCS weapon definition, we would set the `class` to `LongRange`, `CloseRange` and the already present `ShortRange`. The backend will allow to shot only LongRange on first bout, also ShortRange from bout 2, and all direct-damage weapons from bout 3, without any need to specify this in FOCS.

agrrr3:
@oberlus i need this feature, not only for missiles but also for combat ranges and you just started your travel in freeorion cpp, so i doubt we will have a cpp based prototype soon and certainly not soon in master

agrrr3:
> High on my list of stuff to do is to include an `ObjectMap` override, or perhaps mandatory parameter, in the `ScriptingContext`. 

Yes, that would have been very useful back with targeting. I think also it would speed up combat calculation a bit if it would only contain the local system. Are you planning to put the whole universe in there or only the local system?

> The visibility override shouldn't change during condition evaluation, and could probably be stored by reference on its own, rather than constructing an empty map each time. A CombatScriptingContext struct is also reasonable, which could be stored by reference, with the reference pointing to a static empty one in case of non-combat construction of ScriptingContext.

I will give it a fast stab.

Ah s* somebody moved all the basic files around. Have to redo this.

agrrr3:
@geoffthemedio would it be ok to pass a const reference to the existing combat_info into the scripting context? it contains both values (the visibility override and also the current bout). Would be KISS but either ScriptingContext needs to know about about combat info structure. Also it should be possible to to do some lesser magic with templates to remove that dependency (so the real type of the additional info is only known inside CombatSystem.cpp)

geoffthemedio:
> > High on my list of stuff to do is to include an `ObjectMap` override, or perhaps mandatory parameter, in the `ScriptingContext`.
> 
> Yes, that would have been very useful back with targeting. I think also it would speed up combat calculation a bit if it would only contain the local system. Are you planning to put the whole universe in there or only the local system?

Whatever makes sense in each case... For combat, it could be just the objects involved in the combat, using the existing objectmap in the combatinfo.

geoffthemedio:
> @geoffthemedio would it be ok to pass a const reference to the existing combat_info into the scripting context? it contains both values (the visibility override and also the current bout). Would be KISS but either ScriptingContext needs to know about about combat info structure. Also it should be possible to to do some lesser magic with templates to remove that dependency (so the real type of the additional info is only known inside CombatSystem.cpp)

A reference to the whole combatinfo might work for combat specific stuff, but passing an objectmap is potentially useful in other situations as well, so should be passed independently...  It could be set up so that when constrructing a scriptingcontext, one can pass a combatinfo from which the objectmap is taken, or one can pass just an objectmap.

It'd probably be necessary for scriptingcontext to include the combatinfo headers in the .cpp file, but if the interface / constructors only deal with references, then it might ne enough to forward declare class combatinfo in the scriptingcontext header.

agrrr3:
The info i need is actually AutoResolveInfo.

I had a small stab at a generic context payload using templates and using nullptr_t as default. Then I found this sends type ripples through the whole code base. I stopped shortly before hitting the parsers in CommonParams. Then I started rethinking and found that i templates are probably the wrong thing today even besides the ripples - I would need to create trees of Condition<AutoResolveInfo> instances out of  trees of Condition<>. For almost no benefit. I guess my experience with java generics got the better of me here.

Also agree that the ObjectMap is a different problem to solve even if the AutoResolveInfo contains the necessary data. I am also happy with just the objects involved in combat - it would be a rather strange weapon which changes targeting based on things outside the combat (e.g. buildings in neighboring system).

So thanks for the input, i will have another stab now.

geoffthemedio:
> it would be a rather strange weapon which changes targeting based on things outside the combat (e.g. buildings in neighboring system).

Not implausible to have buildings that affecting targeting in other systems...

Oberlus:
Could a reasonably small subset of the universe be passed? Something that includes all objects within N hops or within X uu distance from the system of the battle (assuming no effects could take play based on further away objects), but not the whole universe. X and N would be game rules maybe.
Parts that are boosted by nearby buildings/parts, combat effects that take into account the presence of certain species...
Any effect that comes from further away (mopre than X distance and N jumps) would still be possible as long as they do not change during combat.
Also, being able to base the effects of local parts based on the rest of parts present in the battle, and being able to change that during the battle (like losing certain effect if the capital ship that was giving it is destroyed in first bout) would be great.

Does all this discusion merit for a forum thread? I could copy your messages there if you like.

geoffthemedio:
> Could a reasonably small subset of the universe be passed? 

Plausibly yes, but it would involve setting up a temporary ObjectMap and populating it with the desired subset if objects. But why bother? You could just pass a reference the whole gamestate objectmap as is...  But I 'd just use the combat objectmap and not worry about including any other objects outside the system. I just meant it was plausible one might want to.

> Also, being able to base the effects of local parts based on the rest of parts present in the battle, and being able to change that during the battle (like losing certain effect if the capital ship that was giving it is destroyed in first bout) would be great.

I'm very reluctant to complicate combat resolution with effects and other complicated tests and calculations beyond targetting conditions.

> Does all this discusion merit for a forum thread?

The technical parts, no. The design issues perhaps.

universe/ScriptingContext.h
```diff
@@ -9,6 +9,20 @@
 
 class UniverseObject;
 
+struct BackgroundContext {
```
geoffthemedio: why "Background" ?

universe/ScriptingContext.h
```diff
@@ -9,6 +9,20 @@
...
+    BackgroundContext()
+    {}
+
+    BackgroundContext(int bout_, Universe::EmpireObjectVisibilityMap& empire_object_visibility_) :
+        bout(bout_),
+        empire_object_visibility(empire_object_visibility_)
+    {}
+
+    int bout;
+    Universe::EmpireObjectVisibilityMap empire_object_visibility;       ///< indexed by empire id and object id, the visibility level the empire has of each object.  may be increas
```
geoffthemedio: you're storing a copy of the input map here. it should be possible to store a reference or pointer which points to a static empty map if not specified in the constructor.

universe/ScriptingContext.h
agrrr3: Couldnt find a good name. We have the context info which changes while the conditions get evaluated. The background is what does not change. Maybe i meant rather backdrop(?).

universe/ScriptingContext.h
agrrr3: the main problem is that i still stored a copy of the background in each ScriptingContext.
Changed that to a reference and tried to create an empty background only once and reuse that if no special info is given  (already pushed).

universe/ScriptingContext.h
geoffthemedio: Not sure what you mean by the scripting context changing "while the conditions get evaluated"... how so? It's passed by const reference to condition classes `Eval` functions. It's sometimes used to create a new scripting context for sub-evaluations of conditions or value refs, but the original doesn't change. I don't think this needs a separate struct.

universe/ScriptingContext.h
geoffthemedio: This still makes a copy of the visibility map whenever you construct a background context. Should store it by reference as well... and probably just get rid of the background context and store directly in the scripting context.

universe/ScriptingContext.h
```diff
@@ -9,6 +9,22 @@
...
+
+};
+static const BackgroundContext theEmptyBackground = BackgroundContext();
```
geoffthemedio: Make this a vis map static class variable `s_empty_vis_map`. Don't need to assign it any value; the default empty state should work for the dummy / default thing to bind the reference to.

universe/ScriptingContext.h
agrrr3: > This still makes a copy of the visibility map whenever you construct a background context. Should store it by reference as well... and probably just get rid of the background context and store directly in the scripting context.

Uhm sounds like you want the original implementation (vis map and bout in the ScriptingContext)? You told me you do not want more things in the scripting context. So where should the bout go?

geoffthemedio: Ideal would perhaps be moving the combat round tracking into combat info or autoresolve info, which also has a visibility map in it, and then pass the info struct to the scripting context constructor and store reference to it while also extracting and using the vis info to set a separate reference in scripting context. But doing that requires having a constructor for scripting context that takes a combat info struct of some sort, which might require putting that struct in common... So that can maybe be worked around by making a base class for combat info that contains the combat round and the vis info only, and putting that in common but not the full combat info. So hopefully then you could initialize a scripting context with a combat info (including combat round) without the extra stuff going into the common library.

But at this point I thought for just this issue, it's probably fine to have the combat round stored by value in the scripting context directly and worry about all the rest uf and when more combat info type stuff is to be made accessible from the context. More important is to avoid needing to copy and initialize whole new empty visibility maps with every scripting context and to avoid copying them unnecessarily when passing one to create a scripting context.

universe/ScriptingContext.h
agrrr3: If I got you right you said the same visibility map gets handed through to the sub-evals. The same would be true for a bout value. The idea is that i can put all that data that stays the same in the same data structure and use a single reference to that.

So I should have said "During the evaluation of a condition/valueRef tree there are different values for e.g. localCandidate necessary while some values are always the same."

With reference to vis map and int val in ScriptingContext: every time a ScriptingContext for a subevaluation gets created, both get copied (with the old implementation i think even the whole map got copied every time)

With a reference to a const data structure in ScriptingContext where the data structure contains bout and map : every time a ScriptingContext for a subevaluation gets created, only one reference gets copied. There is are more addresses to resolve, but only once for each Visibility condition or CombatBout condition which is neglible. 

While the reasoning for two values this might be overengineering it is easy now to add more values cost efficiently if we need to.

geoffthemedio: Allocating and copying one int and one reference isn't enough to warrant wrapping in a separate struct yet, I think. If a bunch more gets added, then perhaps in future it would be worthwhile, especially in the wrapped up stuff is conceptually related.

universe/ScriptingContext.h
agrrr3: I tried with AutoresolveInfo before doing the struct. In order to get anywhere i had to expose tons of structure and had add dummy creation ... so i said that is flawed do something simple

See the last commit in [agrrr3/freeorion/202001_CombatBoutCondition_09_stab_at_AutoresolveInfo](https://github.com/agrrr3/freeorion/commits/202001_CombatBoutCondition_09_stab_at_AutoresolveInfo) ( fecb3f0525d3bf2f10c90ca2cbc8a4ef8245a20f )

geoffthemedio: That's why I mentioned making a base class that is public that AutoresolveInfo derives from, so you wouldn't need to make the full contents of AutoresolveInfo public...

agrrr3:
The travis build was killed before it could finish.

universe/ScriptingContext.h
agrrr3: I could not make the static map work. After lots of trials i have a working version where i pass a reference to the CombatInfo (where i added a int bout member). 

Could not make a static const class member work as dummy either it got multiple defined, not initialised or the symbol was hidden. Works a global though.

universe/ScriptingContext.h
agrrr3: Also did not work, I went for the CombatInfo instead.

geoffthemedio:
Nor fully tested, but you can do something like:
````
 combat/CombatSystem.h       |  7 ++-----
 universe/Effects.cpp        |  1 +
 universe/ScriptingContext.h | 21 ++++++++++++++++-----
 3 files changed, 19 insertions(+), 10 deletions(-)

diff --git a/combat/CombatSystem.h b/combat/CombatSystem.h
index eaac44f62..491f4e412 100644
--- a/combat/CombatSystem.h
+++ b/combat/CombatSystem.h
@@ -2,6 +2,7 @@
 #define _CombatSystem_h_
 
 #include "../universe/Universe.h"
+#include "../universe/ScriptingContext.h"
 #include "../util/AppInterface.h"
 #include "CombatEvent.h"
 
@@ -12,7 +13,7 @@
 
 /** Contains information about the state of a combat before or after the combat
   * occurs. */
-struct CombatInfo {
+struct CombatInfo : public ScriptingCombatInfo {
 public:
     /** \name Structors */ //@{
     CombatInfo() = default;
@@ -37,8 +38,6 @@ public:
     std::set<int>                       damaged_object_ids;             ///< ids of objects damaged during this battle
     std::set<int>                       destroyed_object_ids;           ///< ids of objects destroyed during this battle
     std::map<int, std::set<int>>        destroyed_object_knowers;       ///< indexed by empire ID, the set of ids of objects the empire knows were destroyed during the combat
-    Universe::EmpireObjectVisibilityMap empire_object_visibility;       ///< indexed by empire id and object id, the visibility level the empire has of each object.  may be increased during battle
-    int                                 bout = 0;
     std::vector<CombatEventPtr>         combat_events;                  ///< list of combat attack events that occur in combat
 
     float   GetMonsterDetection() const;
@@ -125,6 +124,4 @@ void CombatInfo::load(Archive & ar, const unsigned int version)
     combat_events.swap(           filtered_combat_events);
 }
 
-static const CombatInfo s_empty_combat_info = {};
-
 #endif // _CombatSystem_h_
diff --git a/universe/Effects.cpp b/universe/Effects.cpp
index 2bcc5a518..a89b8f7ef 100644
--- a/universe/Effects.cpp
+++ b/universe/Effects.cpp
@@ -37,6 +37,7 @@ namespace {
 using boost::io::str;
 
 FO_COMMON_API extern const int INVALID_DESIGN_ID;
+FO_COMMON_API extern const ScriptingCombatInfo EMPTY_COMBAT_INFO{};
 
 namespace {
     /** creates a new fleet at a specified \a x and \a y location within the
diff --git a/universe/ScriptingContext.h b/universe/ScriptingContext.h
index 23fe49342..135d75198 100644
--- a/universe/ScriptingContext.h
+++ b/universe/ScriptingContext.h
@@ -2,19 +2,30 @@
 #define _ScriptingContext_h_
 
 #include "Universe.h"
-#include "../combat/CombatSystem.h"
 
 #include <boost/any.hpp>
 
 #include <memory>
 
+FO_COMMON_API struct ScriptingCombatInfo {
+    ScriptingCombatInfo() {}
+    ScriptingCombatInfo(int bout_, const Universe::EmpireObjectVisibilityMap& vis) :
+        bout(bout_),
+        empire_object_visibility(vis)
+    {}
+
+    int bout = 0;
+    Universe::EmpireObjectVisibilityMap empire_object_visibility;
+};
+
+FO_COMMON_API extern const ScriptingCombatInfo EMPTY_COMBAT_INFO;
+
 class UniverseObject;
 
 struct ScriptingContext {
     /** Empty context.  Useful for evaluating ValueRef::Constant that don't
       * depend on their context. */
-    ScriptingContext()
-    {}
+    ScriptingContext() = default;
 
     /** Context with only a source object.  Useful for evaluating effectsgroup
       * scope and activation conditions that have no external candidates or
@@ -27,7 +38,7 @@ struct ScriptingContext {
       * conditions. Useful in combat resolution when the visibility of objects
       * may be different from the overall universe visibility. */
     ScriptingContext(std::shared_ptr<const UniverseObject> source_,
-                     const CombatInfo& combat_info_) :
+                     const ScriptingCombatInfo& combat_info_) :
         source(source_),
         combat_info(combat_info_)
     {}
@@ -90,7 +101,7 @@ struct ScriptingContext {
     std::shared_ptr<const UniverseObject>   condition_root_candidate;
     std::shared_ptr<const UniverseObject>   condition_local_candidate;
     const boost::any                        current_value;
-    const CombatInfo&                       combat_info = s_empty_combat_info;
+    const ScriptingCombatInfo&              combat_info = EMPTY_COMBAT_INFO;
 };
 
 #endif // _ScriptingContext_h_


````

**Comment by agrrr3** [comment]
> o something like:

Ok, i will try that. It would surprise me though if that got rid of the linking error.

**Comment by agrrr3** [comment]
With the patch I still get linker errors when trying to link freeoriond: undefined reference to `ScriptingContext::s_empty_combat_info'

I do not know enough about linking. It seems it fails in CMakeFiles/freeoriond.dir/link.txt and there is CombatSystem.cpp.o mentioned in that link.txt, but not ScriptingContext.

Is there a way to add ScriptingContext there (and would it help)?

**Comment by geoffthemedio** [comment]
Try playing around with `FO_COMMON_API extern const ScriptingCombatInfo EMPTY_COMBAT_INFO{};` in Effects.cpp. Maybe remove `FO_COMMON_API extern ` or similar?

**Comment by agrrr3** [comment]
A non-extern global constant initialisation worked. Should the bout value in CombatInfo be serialized?

**Comment by geoffthemedio** [comment]
> Should the bout value in CombatInfo be serialized?

Probably yes, although as far as I know `CombatInfo` is never serialized...

**Comment by adrianbroher** [comment]
> Try playing around with FO_COMMON_API extern const ScriptingCombatInfo EMPTY_COMBAT_INFO{}; in Effects.cpp. Maybe remove FO_COMMON_API extern or similar?

You're asking for external linking (i.e. telling the compiler that the instance is available somewhere else AND asking for the creation of an instance at the same time (e.g. {}-Initialization).  The compiler starts barfing repeatedly because of this:

```
Building CXX object CMakeFiles/freeorioncommon.dir/Empire/ProductionQueue.cpp.o
In file included from Projects/freeorion/freeorion.git/Empire/../universe/ValueRef.h:4,
                 from Projects/freeorion/freeorion.git/Empire/ProductionQueue.cpp:7:
Projects/freeorion/freeorion.git/Empire/../universe/ScriptingContext.h:24:41: warning: 'visibility' attribute ignored [-Wattributes]
   24 | FO_COMMON_API const ScriptingCombatInfo EMPTY_COMBAT_INFO{};
      |                                         ^~~~~~~~~~~~~~~~~
```

Please fix this properly.

**Comment by agrrr3** [comment]
> > Try playing around with FO_COMMON_API extern const ScriptingCombatInfo EMPTY_COMBAT_INFO{}; in Effects.cpp. Maybe remove FO_COMMON_API extern or similar?
> 
> You're asking for external linking (i.e. telling the compiler that the instance is available somewhere else AND asking for the creation of an instance at the same time (e.g. {}-Initialization). The compiler starts barfing repeatedly because of this:
> 
> ```
> Building CXX object CMakeFiles/freeorioncommon.dir/Empire/ProductionQueue.cpp.o
> In file included from Projects/freeorion/freeorion.git/Empire/../universe/ValueRef.h:4,
>                  from Projects/freeorion/freeorion.git/Empire/ProductionQueue.cpp:7:
> Projects/freeorion/freeorion.git/Empire/../universe/ScriptingContext.h:24:41: warning: 'visibility' attribute ignored [-Wattributes]
>    24 | FO_COMMON_API const ScriptingCombatInfo EMPTY_COMBAT_INFO{};
>       |                                         ^~~~~~~~~~~~~~~~~
> ```
> 
> Please fix this properly.

I thought extern is telling the compiler that the instance is available somewhere else (which i did not use). I couldnt find the FO_COMMON_API definition though, where is it?

**Comment by adrianbroher** [comment]
> I thought extern is telling the compiler that the instance is available somewhere else (which i did not use).

That is correct and what I mentioned in the first half of the sentence.  However you are also creating an instance by using curly-braces initialization (`{}`).  The compiler error stems from the fact that FO_COMMON_API resolves with GCC and Clang to:

```c++
__attribute__((__visibility__("default"))) const ScriptingCombatInfo EMPTY_COMBAT_INFO{};
```

If you want to properly expose the instance do something like:

```c++
// in header.h

FO_COMMON_API extern const ScriptingCombatInfo EMPTY_COMBAT_INFO;
```

```c++
// in implementation.cpp

const ScriptingCombatInfo EMPTY_COMBAT_INFO{};
```

> I couldnt find the FO_COMMON_API definition though, where is it?

util/Export.h

PR #2700, opened 2020-01-11
Fix finalization segfaults for upcoming python3

o01eg:
In Python3 it's fobidden to call object desructors after python finalizaton.
Fix finalization segfaults by enforcing object finalization before Python interpreter.

Extracted from #2653


geoffthemedio:
I don't have much input on this... I ran a quick test game with the Python2 SDK and see no relevant issues.

PR #2704, opened 2020-01-12
Fix G++ warnings

adrianbroher:
This PR fixes some problematic code, which causes G++ to emit warnings:

* Remove unused and deprecated boost header for endian detection, which emits a warning since boost 1.69.0.
* Remove unreferenced functions.
* Remove unused local variables.
* Catch exceptions by reference in GG Font implementation.
* Add pragma stament block to std_image.h to let GCC ignore multiple "misleading indentations" (talking about pendantic).

The remaining warning about the std::move call in MovableEnvelope was kept as I couldn't yet read into possible consequences of that change.

The warning about violated strict aliasing rules in Effect.cpp require some bigger redesign of the EffectBase interface to be fixed.

### DIFF
--- a/GG/src/Font.cpp
+++ b/GG/src/Font.cpp
@@ -2242,7 +2242,7 @@ void Font::HandleTag(const std::shared_ptr<FormattingTag>& tag, double* orig_col
                     } else {
                         well_formed_tag = false;
                     }
-                } catch (boost::bad_lexical_cast) {
+                } catch (const boost::bad_lexical_cast&) {
                     try {
                         double color[4];
                         color[0] = lexical_cast<double>(tag->params[0]);
@@ -2259,7 +2259,7 @@ void Font::HandleTag(const std::shared_ptr<FormattingTag>& tag, double* orig_col
                         } else {
                             well_formed_tag = false;
                         }
-                    } catch (boost::bad_lexical_cast) {
+                    } catch (const boost::bad_lexical_cast&) {
                         well_formed_tag = false;
                     }
                 }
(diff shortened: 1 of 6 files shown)


GG/src/nanovg/stb_image.h
```diff
@@ -1375,6 +1378,8 @@ static unsigned char *stbi__convert_format(unsigned char *data, int img_n, int r
          default: STBI_ASSERT(0);
       }
       #undef CASE
+
+      #pragma GCC diagnostic push
```
o01eg: Shouldn't be there `#pragma GCC diagnostic pop` to pair up with `push` above?

o01eg (changes requested):

GG/src/nanovg/stb_image.h
adrianbroher: Yes it should

adrianbroher: Fixed via 4dd097c66614fe09845c4b4f4d535a62376e11c0

GG/src/Font.cpp
```diff
@@ -2242,7 +2242,7 @@ void Font::HandleTag(const std::shared_ptr<FormattingTag>& tag, double* orig_col
                     } else {
                         well_formed_tag = false;
                     }
-                } catch (boost::bad_lexical_cast) {
+                } catch (const boost::bad_lexical_cast&) {
```
geoffthemedio, line 2245: could make these just `catch (...)`

adrianbroher: It should only catch the `boost::bad_lexical_cast` emitted by the `boost::lexical_cast`, not every exception.

PR #2705, opened 2020-01-12
Make UniverseObject's dump Python API don't accept arguments

o01eg:
Changes for #2431

### DIFF
--- a/python/UniverseWrapper.cpp
+++ b/python/UniverseWrapper.cpp
@@ -51,6 +51,8 @@ namespace boost {
 namespace {
     void                    DumpObjects(const Universe& universe)
     { DebugLogger() << universe.Objects().Dump(); }
+    std::string             ObjectDump(const UniverseObject& obj)
+    { return obj.Dump(0); }
 
     // We're returning the result of operator-> here so that python doesn't
     // need to deal with std::shared_ptr class.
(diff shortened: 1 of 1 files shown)


Morlic-fo (approved):

python/UniverseWrapper.cpp
```diff
@@ -473,7 +475,7 @@ namespace FreeOrionPython {
             .def("hasTag",                      &UniverseObject::HasTag)
             .add_property("meters",             make_function(ObjectMeters,                 return_internal_reference<>()))
             .def("getMeter",                    make_function(ObjectGetMeter,               return_internal_reference<>()))
-            .def("dump",                        &UniverseObject::Dump,                      return_value_policy<return_by_value>(), "Returns string with debug information, use '0' as argument.")
+            .def("dump",                        make_function(&ObjectDump,                  return_value_policy<return_by_value>()), "Returns string with debug information.")
```
geoffthemedio, line 478: If it's a no-parameter function, it could be instead defined as a property to the API with `add_property`...?

o01eg: @Morlic-fo calls it as function.

geoffthemedio: Is there a reason to prefer that over treating it as a property?

o01eg: I suppose it better to ask @Morlic-fo

Morlic-fo: It could be defined as a property and it is ultimately a matter of taste/style.

Typically, you would use a property where it would make sense if that property was an actual member attribute of the class/object (or actually refers to some corresponding member attribute but you need to apply some additional calculations/constraints or want to provide a read-only interface).

Having the dump-string as an actual member attribute doesn't sound very intuitive to me (and also isn't what happens on the C++ side), so a function seems to be more natural.

PR #2710, opened 2020-01-12
[prototype] fighter based missile FOCS-only

agrrr3:
[prototype] Add SHP_FIGHTERS_MISSILE tech FT_BAY_ and FT_HANGAR_MISSILE too

This should not go into master/a release. It is a prototype so one can try a new type of weapon.

Note this depends on another commit/PR #2698.

[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11474)
    
    This provides a FOCS-only prototype for missile weapon
    
    Gameplay
    * missiles launch on bout 1, do nothing in bout 2, and do 3 shield-piercing damage in bout 3
    * missiles target ships and fighters (no planets)
    * as many 20PP FT_BAY_MISSILE external parts as you like (provide each 4 bout-1 launch capacity and 4 missiles munition)
    * up to one 20PP FT_HANGAR_MISSILE internal part per ship (necessary, also provides 16 missiles munition)
    
    Technical:
    * uses MISSILE_STORAGE_CAPACITY special to track actual missile storage
    * hangar capacity is set to launch capacity so missiles only launch in bout 1
    * target condition only targets enemies in bout 3 (i.e. bout 1 lauch, bout 2 shot-me-down, bout 3 damage others)
    * recaptured missiles are effectively ignored as storage gets always lowered by launch capacity after battle
    * Note this cuts MaxCapacity down to launch rate to prevent unwanted resupply
      else MaxCapacity could be used to show current capacity to user
    * Own specific resupply effect is ignored because of resupply hardcoded in backend (in Fleet AFAICR)
    
    Tactical:
    * can easily build designs which launch enormous amounts of missiles
    * enemy can shield ships by fielding fighters
    * enemy wants to shoot down missiles in bout two
    * enemy's Interceptors are both effective as chaff as well as taking down misisles
    * you can shield your missiles using Interceptors
    * you can shield your Fighters/Bombers by shooting missiles
    * you have to take care of your supply lines if
    * if you want to use missiles in enemy territory or longer battles you can put fewer launchers
    
    Balancing questions
    * probably munition rate of hangar should be higher. maybe launchers should not carry munition (e.g. hangar: 32 launcher: 0 )
    * having only one launch bay is not really juicy - its comparable..
    ** offensively to a Fighter bay/hangar but doing 25% percent less damage (if three combat bouts and not shot down before)
    ** defensively to a slightly expensive Interceptor bay/hangar
    ** ... so maybe prefer missiles over Interceptors if you do not want to target enemy's fighters
    * starting with two launch bays you do not have to supply extra hangars so damage and chaff potential goes up
    * robo hull probably needs to loose an external part slot if this is in (4 launchers seems OP)
    
    Ugly
    * in order to not change current code, part exclusions do not work if adding fighter parts afterwards
    * no UI currently to show missile storage
    * missiles will provide cover after bout 3 even if "exploded"
    * unconditional launching
    * no specific resupply condition
    * are shown as fighters in combat log and also get recaptured at end of battle
    * damage is not scaled accordingly to number of bouts

### DIFF
--- a/default/scripting/ship_parts/FighterBay/FT_BAY_MISSILE.focs.txt
+++ b/default/scripting/ship_parts/FighterBay/FT_BAY_MISSILE.focs.txt
@@ -1,0 +1,14 @@
+Part
+    name = "FT_BAY_MISSILE"
+    description = "FT_BAY_MISSILE_DESC"
+    exclusions = [ "FT_HANGAR_0" "FT_HANGAR_1" "FT_HANGAR_2" "FT_HANGAR_3" "FT_HANGAR_4" "FT_BAY_1" ]
+    class = FighterBay
+    capacity = 4
+    mountableSlotTypes = External
+    buildcost = 20 * [[FLEET_UPKEEP_MULTIPLICATOR]] * [[SHIP_PART_COST_MULTIPLIER]]
+    buildtime = 1
+    tags = [ "PEDIA_PC_FIGHTER_BAY" ]
+    location = OwnedBy empire = Source.Owner
+    icon = "icons/ship_parts/nuclear_missile.png"
+
+#include "/scripting/common/upkeep.macros"
--- a/default/scripting/ship_parts/FighterHangar/FT_HANGAR_MISSILE.focs.txt
+++ b/default/scripting/ship_parts/FighterHangar/FT_HANGAR_MISSILE.focs.txt
@@ -1,0 +1,89 @@
+Part
+    name = "FT_HANGAR_MISSILE"
+    description = "FT_HANGAR_MISSILE_DESC"
+    exclusions = [ "FT_HANGAR_0" "FT_HANGAR_1" "FT_HANGAR_2" "FT_HANGAR_3" "FT_HANGAR_4" "FT_HANGAR_MISSILE" "FT_BAY_1" ]
+    class = FighterHangar
+    // Capacity is set to launch capacity per launch bay (4)
+    // MaxCapacity is also set to that value
+    // "Real" missile capacity (storage) is four times launch capacity plus once launch capacity per bay (4 per launch bay)
+    // As default for UI we put four times launch capacity
+    capacity = 16
+    damage = 3
+    // There is a problem with storage/resupply and launching:
+    //   resupply happens in Fleet.cpp in movement phase and sets all parts capacity to maximum capacity
+    //   but hangar capacity has to be equal to launch capacity (because missile should only launch in bout 1)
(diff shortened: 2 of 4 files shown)


agrrr3:
@geoffthemedio should i add someting like a `NoDefaultSupply` flag to suppress standard supply effect and be able to script part specific supply?

geoffthemedio:
Not sure what specifically you mean...

agrrr3:
> Not sure what specifically you mean...

@geoffthemedio Hm I think I also need to figure out what did not work when i implemented this. I wrote something like: 'Resupply sets the capacity to maximum capacity for all parts which overrides my definition of resupply'. So I wanted to ask if it is ok for you to include a flag in the part definition which allows to skip setting a part capacity to maximum capacity when supply happens. 

Oh think I got it. The launch capacity should be the same as hangar capacity so missiles are only launched in the first bout. I wanted to use the hangar MaxCapacity in order to show how much many missiles are stored, but the default supply didnt allow the capacity to be less than maxcapacity inside of supply.

agrrr3:
It probably makes more sense to rather add a generic launchCondition for launch bay parts for having sensible supply and hangar max capacity. Launch capacity and hangar max capacity would work the normal way. For simulating exploding munitions one still needs a meter or special to count and set the current capacity (or implement something in the backend to remove retrieval for this)

geoffthemedio:
> For simulating exploding munitions one still needs a meter or special to count and set the current capacity

Could you explain that? The relevance us unclear.

agrrr3:
> > For simulating exploding munitions one still needs a meter or special to count and set the current capacity
> 
> Could you explain that? The relevance us unclear.

This draft prototype is "simulating" missile systems using existing FOCS functionality. There are quite some mismatches which make this awkward. If the missiles would work as intended they would be used up when they hit something and probably would not be retrieved if they do not hit anything. 

When using fighters as missiles and four fighters start at the end of combat zero to four fighters are retrieved no matter if they found a target. But the hangar capacity should be four less after that combat compared to before - so this needs to be tracked seperatly.

Cjkjvfnby:
Can you share a few statistics of fighting out of the supply zone?  Does it happen often?

Could this weapon be used as a good defense weapon? (It will have eighter high damage or reduce attacker strength by requiring more AA weapons)).



PS. Fighter-based Missle **KS-1 Komet** and **MiG-15** in the background.

image

agrrr3:
> Can you...

this discussion belongs in the forums, not on github. I will post there

Issue #2690, opened 2020-01-05
issueFleetMoveOrder() changes already enqueued route of a fleet


aschuch247:
### Environment

* **FreeOrion Version**: 456a9b64510faf81adfc637b9eb5c106f2d0139e, compiled from source
* **Fetched as**: Compiled from source

### Description

* This is an overview of the situation: ![changed-routing](https://user-images.githubusercontent.com/13714468/71782339-c4d1ac80-2fd0-11ea-8bbb-992bb0e5e7e0.png)
* The marked fleet is at system 879.
* My AI issues the following commands, using `issueFleetMoveOrder()`:
  * Ordering fleet 2072 to move to system 543 with result 1.
  * Ordering fleet 2072 to move to system 777 with result 1.
  * Ordering fleet 2072 to move to system 825 with result 1.
  * Ordering fleet 2072 to move to system 288 with result 1.
  * Ordering fleet 2072 to move to system 903 with result 1.
  * Ordering fleet 2072 to move to system 843 with result 1.
  * Ordering fleet 2072 to move to system 189 with result 1.
  * Ordering fleet 2072 to move to system 909 with result 1.
  * Ordering fleet 2072 to move to system 309 with result 1.
  * Ordering fleet 2072 to move to system 894 with result 1.
* But the server changes the path:
  * Order.cpp:346 : FleetMoveOrder::ExecuteImpl Setting route of fleet 2072 to  879 543
  * Order.cpp:346 : FleetMoveOrder::ExecuteImpl Setting route of fleet 2072 to  879 543 777
  * Order.cpp:346 : FleetMoveOrder::ExecuteImpl Setting route of fleet 2072 to  879 273 825
    * **Here, the path gets changed automatically!**
  * Order.cpp:346 : FleetMoveOrder::ExecuteImpl Setting route of fleet 2072 to  879 273 825 288
  * Order.cpp:346 : FleetMoveOrder::ExecuteImpl Setting route of fleet 2072 to  879 273 825 288 903
  * Order.cpp:346 : FleetMoveOrder::ExecuteImpl Setting route of fleet 2072 to  879 273 825 288 903 843
  * Order.cpp:346 : FleetMoveOrder::ExecuteImpl Setting route of fleet 2072 to  879 273 825 288 903 843 189
  * Order.cpp:346 : FleetMoveOrder::ExecuteImpl Setting route of fleet 2072 to  879 273 825 288 903 843 189 909
[...]

PR #2712, opened 2020-01-14
Fix FleetMoveOrder::Check when appending a route

Morlic-fo:
Related to #2222, the ```FleetMoveOrder::Check``` is currently broken when appending a system to a route.

This does not show in the human client, because the ```append``` parameter is actually never passed to the function (and defaults to false) - but it causes issues for the AI (#2690).

The check removed in this PR was useful when the starting system for a FleetMoveOrder had to be explicitly passed and thus needed to be verified. Nowadays, when ```append=true```, the starting system for the pathfinding/route to append will always be the last system in the current route.

### DIFF
--- a/util/Order.cpp
+++ b/util/Order.cpp
@@ -293,26 +293,10 @@ bool FleetMoveOrder::Check(int empire_id, int fleet_id, int dest_system_id, bool
         return false;
     }
 
-    int start_system = fleet->SystemID();
-    if (start_system == INVALID_OBJECT_ID)
-        start_system = fleet->NextSystemID();
-
     auto dest_system = GetEmpireKnownSystem(dest_system_id, empire_id);
     if (!dest_system) {
         ErrorLogger() << "Empire with id " << empire_id << " ordered fleet to move to system with id " << dest_system_id << " but no such system is known to that empire";
         return false;
-    }
-
-    // verify fleet route first system
-    if (append && !fleet->TravelRoute().empty()) {
-        // We should append and there is something to append to
-        int last_system = fleet->TravelRoute().back();
-        if (last_system != start_system) {
-            ErrorLogger() << "Empire with id " << empire_id
-                          << " ordered a fleet to continue from system with id " << start_system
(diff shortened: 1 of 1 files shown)


Morlic-fo:
Seems I am not too compatible with the Github web interface for creating commits. I will push an actually useful commit message once I have access to a machine with git.

adrianbroher:
> Seems I am not too compatible with the Github web interface for creating commits.

The commit message editor below the file editor.

geoffthemedio:
> Nowadays, when `append=true`, the starting system for the pathfinding/route to append will always be the last system in the current route.

What about when `append=false`?

Morlic-fo:
> What about when `append=false`?

Unrelated to this PR, but it uses the fleet's current location (or next system when on the starlane).

In the constructor:
https://github.com/freeorion/freeorion/blob/0a3702c67b91461832addb654d61c29eda566acb/util/Order.cpp#L269-L273

util/Order.cpp
```diff
@@ -293,28 +293,12 @@ bool FleetMoveOrder::Check(int empire_id, int fleet_id, int dest_system_id, bool
...
-        start_system = fleet->NextSystemID();
-
     auto dest_system = GetEmpireKnownSystem(dest_system_id, empire_id);
     if (!dest_system) {
         ErrorLogger() << "Empire with id " << empire_id << " ordered fleet to move to system with id " << dest_system_id << " but no such system is known to that empire";
         return false;
     }
 
-    // verify fleet route first system
-    if (append && !fleet->TravelRoute().empty()) {
```
geoffthemedio, line 307: I think this was the only use of `append` so it could be removed from the function signature...?

Morlic-fo: The AIWrapper currently expects same parameters for the constructor and ```::Check```
https://github.com/freeorion/freeorion/blob/4a704df4d0fd2510b1d6ffbe7da502604cedaabb/python/AI/AIWrapper.cpp#L95-L105

PR #2716, opened 2020-01-17
Let AI scrap Gateway to the void

### LINKED ISSUE #2715: AI owned Gateway to the Void causes self-pwnage
[Bug reported by spikethehobbit on the forum](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11490):

### Environment

* **FreeOrion Version**: 2019-12-16.cf05826

### Description

An AI player captured an outpost on the same turn that a Gateway to the Void was completed there. The AI then kept trying to reinforce that system with predictable results.

### Suggestion

If the AI manages to capture a GttV, it should sell it if it doesn't want it.

Morlic-fo:
Resolves #2715

### DIFF
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -1113,13 +1113,14 @@ def generate_production_orders():
         else:
             warn("Failed enqueueing %s at planet %s, got result %d" % (building_name, planet, res))
 
-    building_name = "BLD_EVACUATION"
+    buildings_to_scrap = ("BLD_EVACUATION", "BLD_GATEWAY_VOID")
     for pid in state.get_inhabited_planets():
         planet = universe.getPlanet(pid)
         if not planet:
             continue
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I don't consider this release blocking, but if there's an RC2, then this could presumably go in? @Vezzra not sure how to tag accordingly.

TheSilentOne1:
Should the AI also scrap concentration camps if it conquers them? In multiplayer it's often used as a strategy to prevent an enemy from acquiring your species.

Morlic-fo:
There is already dedicated code for Concentration Camp handling which may also scrap it.

Vezzra:
> I don't consider this release blocking, but if there's an RC2, then this could presumably go in? @Vezzra not sure how to tag accordingly.

Sounds reasonable. Tagging as "cherry-pick for release" is fine in such cases.

Issue #2719, opened 2020-01-18
Supply line connection disappears


kaposvarigellert:
### Environment

* **FreeOrion Version**: FreeOrion_v0.4.9-RC1_2020-01-13
* **Operating System**:  Windows 10 Pro 64bit
* **Graphic card used**: AMD Radeon R7 M260
* **Fetched as** 
  * Binary release

### Description

My supply lines collapsed after clicking end turn. I can't produce ships, buildings etc.
Ship refueling does not work either.

I recently used a Black Hole Collapser and the generated map has active Experimentors.

[FreeOrion_Lord_Of_Terra_Terran_0297_20200118_055839.zip](https://github.com/freeorion/freeorion/files/4080783/FreeOrion_Lord_Of_Terra_Terran_0297_20200118_055839.zip)
![bug](https://user-images.githubusercontent.com/20242050/72665414-8db7be00-3a08-11ea-8080-04ad0b9526d1.JPG)

### Steps to reproduce

* First step.
* Second step.
* Another step.
* Issue occurs.

Vezzra:
Confirmed on OSX.

Depending on if this is only a very rare edge case that only occurs under very special circumstances or can happen more frequently, this is potentially a very serious bug.

Shall we consider it as release-blocking?

@kaposvarigellert, do you have a save from one turn earlier? Before the supply lines disappear? That might be helpful for debugging...

kaposvarigellert:
Sure
[FreeOrion_Lord_Of_Terra_Terran_0296_20200118_055738.zip](https://github.com/freeorion/freeorion/files/4083593/FreeOrion_Lord_Of_Terra_Terran_0296_20200118_055738.zip)

I'm not a game dev, but it might have something to do with the Black Hole Collapser.
I think this because my team and the green player is the only one affected.

I play this game since 0.3.2 and this is the first bug I've ever found in any release candidate.
(I only play the game a few times after every major update.)

Vezzra:
I just tried the save of turn 296 on my Mac and my Windows machines. I can reproduce the issue on both. I have no idea what's going on here, though.

The supply meters all show the values you'd expect, but for some strange reason all the systems are not considered connected. The issue also does not go away after advancing a couple of turns.

Declaring the official release is scheduled for coming Sunday (26th). @geoffthemedio, shall I proceed with that? Or should we consider this issue serious enought to be release-blocking? See my comment above:

[...]

PR #2721, opened 2020-01-19
Fix compile error in GCC 5.4.0

### LINKED ISSUE #2720: FreeOrion does not compile in gcc 5.4.0
# Environment

* **FreeOrion Version**: Git 4195b0c
* **Operating System**: Ubuntu
* gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609

### Description

```
freeorion/UI/CUIControls.cpp: In constructor ‘StatisticIcon::StatisticIcon(std::shared_ptr<GG::Texture>, double, int, bool, GG::X, GG::Y)’:
freeorion/UI/CUIControls.cpp:1304:42: error: converting to ‘const value_type {aka const std::tuple<double, int, bool>}’ from initializer list would use explicit constructor ‘constexpr std::tuple< <template-parameter-1-1> >::tuple(_UElements&& ...) [with _UElements = {double&, int&, bool&}; <template-parameter-2-2> = void; _Elements = {double, int, bool}]’
     m_values(1, {value, digits, showsign})
```

### Expected Result

I expect the source code to compile.

### Steps to reproduce

* Compile the mentioned FreeOrion version with the mentioned compiler.

geoffthemedio:
Hopefully fixes #2720

### DIFF
--- a/UI/CUIControls.cpp
+++ b/UI/CUIControls.cpp
@@ -1301,7 +1301,7 @@ StatisticIcon::StatisticIcon(const std::shared_ptr<GG::Texture> texture,
                              double value, int digits, bool showsign,
                              GG::X w /*= GG::X1*/, GG::Y h /*= GG::Y1*/) :
     GG::Control(GG::X0, GG::Y0, w, h, GG::INTERACTIVE),
-    m_values(1, {value, digits, showsign})
+    m_values(1, std::tuple<double, int, bool>{value, digits, showsign})
 {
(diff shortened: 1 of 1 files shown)


Vezzra:
In case we do a RC2, should this be cherry picked?

Just in case. As no issues have turned up until now, it seems very unlikely that we're going to need another release candidate...

geoffthemedio:
I think the changes are a fix to changes that weren't cherry picked to release, so picking this fi doesn't make sense / isn't actually possible.

There is https://github.com/freeorion/freeorion/issues/2719 which may be narrow enough to ignore, but which I'm still investigating.

Vezzra:
Ok.

PR #2722, opened 2020-01-19
Simplify ObjectMap interface

adrianbroher:
This PR simplifies the ObjectMap class interface:

* Use established naming conventions to mimic std containers.
  * NumObjects - size
  * Empty -> empty
  * Object -> get
  * FindObjects (with some query parameter) -> find
  * FindObjects (without query parameter) -> all
  * const_{begin,end} -> {begin, end} with const modifier
  * Insert -> insert
  * Remove -> erase
  * Clear -> clear
* Remove non-templates variants if they can be substituted by the template equivalent
  * begin
  * end
  * const_end
  * const_begin
  * FindObjects
  * Object
  * NumObjects
* Reimplement ObjectMap::find(const UniverseObjectVisitor&) as template member function to query only a subtype with an applied visitor.
* Remove internal ObjectMap helpers from class interface.
* Replace FindObjectIDs convenience functions with the corresponding all or find equivalent.

### DIFF
--- a/AI/AIInterface.cpp
+++ b/AI/AIInterface.cpp
@@ -173,7 +173,7 @@ namespace AIInterface {
             // ownership to all, while remembering which planets this is done
             // to.
             universe.InhibitUniverseObjectSignals(true);
-            for (auto& planet : universe.Objects().FindObjects<Planet>()) {
+            for (auto& planet : universe.Objects().all<Planet>()) {
                  if (planet->Unowned()) {
                      unowned_planets.push_back(planet);
                      planet->SetOwner(player_id);
--- a/UI/EncyclopediaDetailPanel.cpp
+++ b/UI/EncyclopediaDetailPanel.cpp
@@ -297,7 +297,7 @@ namespace {
 
                 // occupied planets
                 std::vector<std::shared_ptr<const Planet>> species_occupied_planets;
-                for (auto& planet : Objects().FindObjects<Planet>()) {
+                for (auto& planet : Objects().all<Planet>()) {
                     if ((planet->SpeciesName() == entry.first) && !known_homeworlds.count(planet->ID()))
                         species_occupied_planets.push_back(planet);
                 }
@@ -365,7 +365,7 @@ namespace {
             }
 
         } else if (dir_name == "ENC_SHIP") {
-            for (auto& ship : Objects().FindObjects<Ship>()) {
+            for (auto& ship : Objects().all<Ship>()) {
                 const std::string& ship_name = ship->PublicName(client_empire_id);
                 sorted_entries_list.insert({ship_name,
                                             {LinkTaggedIDText(VarText::SHIP_ID_TAG, ship->ID(), ship_name) + "  ",
@@ -373,7 +373,7 @@ namespace {
             }
 
         } else if (dir_name == "ENC_MONSTER") {
-            for (auto& ship : Objects().FindObjects<Ship>()) {
+            for (auto& ship : Objects().all<Ship>()) {
                 if (!ship->IsMonster())
                     continue;
                 const std::string& ship_name = ship->PublicName(client_empire_id);
@@ -395,7 +395,7 @@ namespace {
             }
 
         } else if (dir_name == "ENC_FLEET") {
-            for (auto& fleet : Objects().FindObjects<Fleet>()) {
+            for (auto& fleet : Objects().all<Fleet>()) {
                 const std::string& flt_name = fleet->PublicName(client_empire_id);
                 sorted_entries_list.insert({flt_name,
                                             {LinkTaggedIDText(VarText::FLEET_ID_TAG, fleet->ID(), flt_name) + "  ",
@@ -403,7 +403,7 @@ namespace {
             }
 
         } else if (dir_name == "ENC_PLANET") {
-            for (auto& planet : Objects().FindObjects<Planet>()) {
+            for (auto& planet : Objects().all<Planet>()) {
                 const std::string& plt_name = planet->PublicName(client_empire_id);
                 sorted_entries_list.insert({plt_name,
                                             {LinkTaggedIDText(VarText::PLANET_ID_TAG, planet->ID(), plt_name) + "  ",
@@ -411,7 +411,7 @@ namespace {
             }
 
         } else if (dir_name == "ENC_BUILDING") {
-            for (auto& building : Objects().FindObjects<Building>()) {
+            for (auto& building : Objects().all<Building>()) {
                 const std::string& bld_name = building->PublicName(client_empire_id);
                 sorted_entries_list.insert({bld_name,
                                             {LinkTaggedIDText(VarText::BUILDING_ID_TAG, building->ID(), bld_name) + "  ",
@@ -419,7 +419,7 @@ namespace {
             }
 
         } else if (dir_name == "ENC_SYSTEM") {
-            for (auto& system : Objects().FindObjects<System>()) {
+            for (auto& system : Objects().all<System>()) {
                 const std::string& sys_name = system->ApparentName(client_empire_id);
                 sorted_entries_list.insert({sys_name,
                                             {LinkTaggedIDText(VarText::SYSTEM_ID_TAG, system->ID(), sys_name) + "  ",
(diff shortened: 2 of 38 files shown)


geoffthemedio:
I don't like the `Object` -> `at` renaming, as I find that name very unintuitive, and would much prefer something like `get`, and because I think `ObjectMap`'s function works differently from most (?) STL containers' `at` functions, in that if there is no such item in the container, it doesn't throw an exception, but rather returns a `null_ptr`.

GG/GG/ZList.h
```diff
@@ -59,14 +59,14 @@ class GG_API ZList
     /** RenderOrderIterable can be iterated in back to front render order. */
     struct RenderOrderIterable {
         RenderOrderIterable(const std::list<std::shared_ptr<Wnd>>& list) :
-            m_list(list.crbegin(), list.crend())
+            m_list(list.rbegin(), list.rend())
```
geoffthemedio: seems unrelated to the subject of this pull request...

UI/OptionsWnd.cpp
```diff
@@ -1279,7 +1279,7 @@ namespace {
 
         const std::regex dot { "\\.+" };
         const std::vector<std::string> nodes {
-            std::sregex_token_iterator(name.cbegin(), name.cend(), dot, -1),
+            std::sregex_token_iterator(name.begin(), name.end(), dot, -1),
```
geoffthemedio: also seems unrelated

python/UniverseWrapper.cpp
```diff
@@ -72,20 +72,14 @@ namespace {
...
-    std::vector<int>        PlanetIDs(const Universe& universe)
-    { return Objects().FindObjectIDs<Planet>(); }
-    std::vector<int>        ShipIDs(const Universe& universe)
-    { return Objects().FindObjectIDs<Ship>(); }
-    std::vector<int>        BuildingIDs(const Universe& universe)
-    { return Objects().FindObjectIDs<Building>(); }
+    template<typename T>
+    std::vector<int> ObjectIDs(const Universe& universe)
+    {
+        std::vector<int> result;
```
geoffthemedio: probably worth reserving space before repeatedly pushing a known number of items into container

UI/OptionsWnd.cpp
adrianbroher: See #2730

GG/GG/ZList.h
adrianbroher: See #2730

adrianbroher:
Renamed `ObjectMap::at` to `ObjectMap::get`.

python/UniverseWrapper.cpp
adrianbroher: Done

combat/CombatSystem.cpp
```diff
@@ -1303,7 +1301,7 @@ namespace {
         }
     };
 
-    std::vector<PartAttackInfo> GetWeapons(std::shared_ptr<UniverseObject>& attacker) {
+    std::vector<PartAttackInfo> GetWeapons(std::shared_ptr<UniverseObject> attacker) {
```
geoffthemedio, line 1304: Reason for passing by value and not reference? (Or better: const reference)

universe/Universe.cpp
```diff
@@ -856,12 +856,17 @@ void Universe::UpdateMeterEstimatesImpl(const std::vector<int>& objects_vec, boo
...
         obj->BackPropagateMeters();
 }
 
 void Universe::BackPropagateObjectMeters()
-{ BackPropagateObjectMeters(m_objects.FindObjectIDs()); }
+{
+    std::vector<int> result;
+    for (const auto& obj : m_objects.all())
+        result.push_back(obj->ID());
+    BackPropagateObjectMeters(result);
```
geoffthemedio: This version of `BackPropagateObjectMeters` should probably just do the `obj->BackPropagateMeters()` call directly on all the objects, rather than looping over objects to get their ids, push them into a vector, pass that vector into the overload, which then calls `find` with the vector of ids to get the objects again, in order to do the actual calls.

AI/AIInterface.cpp
```diff
@@ -173,7 +173,7 @@ namespace AIInterface {
             // ownership to all, while remembering which planets this is done
             // to.
             universe.InhibitUniverseObjectSignals(true);
-            for (auto& planet : universe.Objects().FindObjects<Planet>()) {
+            for (auto& planet : universe.Objects().all<Planet>()) {
                  if (planet->Unowned()) {
                      unowned_planets.push_back(planet);
```
geoffthemedio, line 178: not related to this pull request, but another case where reserving space before pushing into vector would be good...

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -297,7 +297,7 @@ namespace {
 
                 // occupied planets
                 std::vector<std::shared_ptr<const Planet>> species_occupied_planets;
```
geoffthemedio, line 299: also here...

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -1671,10 +1671,9 @@ namespace {
 
         // Fleets
         std::vector<std::shared_ptr<const UniverseObject>> nonempty_empire_fleets;
-        for (auto& maybe_fleet : Objects().FindObjects(OwnedVisitor<Fleet>(empire_id))) {
-            if (auto fleet = std::dynamic_pointer_cast<const Fleet>(maybe_fleet))
-                if (!fleet->Empty())
-                    nonempty_empire_fleets.push_back(fleet);
+        for (const auto& fleet : Objects().find<Fleet>(OwnedVisitor<Fleet>(empire_id))) {
+            if (!fleet->Empty())
+                nonempty_empire_fleets.push_back(fleet);
```
geoffthemedio, line 1676: and here... probably there are a lot that should be handled separately

universe/Universe.cpp
adrianbroher: The vector&lt;int&gt; variant is only used at this location so it can be dropped entirely.

AI/AIInterface.cpp
adrianbroher: There are multiple location where universe object id vector is not nessecary, but the size of the MR is already big enough as it is.

combat/CombatSystem.cpp
adrianbroher: GetWeapons ist not called with shared_ptr&lt;UniverseObject&gt; all the time, but shared_ptr&lt;Ship&gt;.

I'm also not sure it we are even using the shared_ptr as intended.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-resource
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-smartptrparam

universe/Universe.cpp
adrianbroher: Done

combat/CombatSystem.cpp
geoffthemedio: > GetWeapons ist not called with shared_ptr all the time, but shared_ptr.

?

adrianbroher: Fixed the comment. Github ate the angle brackets.

geoffthemedio: > GetWeapons ist not called with shared_ptr<UniverseObject> all the time, but shared_ptr<Ship>.

Not sure of the relevance... pointer-to-derived should be passable as pointer-to-base either way.

My thinking is that passing by value, if not optimized away, would require a copy of the pointer, which involves incrementing the reference count, while passing by const reference would not.

adrianbroher: You don't pass a pointer-to-derived, but a shared_ptr&lt;T&gt; as shared_ptr&lt;U&gt;&, which are unrelated types.  The fact that T is-a U doesn't matter for the compiler here.

geoffthemedio:
As in https://github.com/freeorion/freeorion/commit/1ac218d51a93641e6367d5cd7130307ce1af9d7d it's currently awkward to iterate over the type-filtered contents of `ObjectMap` (ie. only systems, not all objects), since `all` creates a vector and fills it with objects and returns it, and range-based for doesn't have a template parameter to select a templated `begin` overload.

Do you have a better suggestion than replacing
````
for (auto& system : Objects().all<System>()) {
````
with
````
for (auto sys_it = objects.begin<System>(); sys_it != objects.end<System>(); ++sys_it) {
    auto& system = *sys_it;
````
?

adrianbroher:
The idea was to adopt boost ranges in the next changesets for this.  However to adopt ranges I first got rid of constructs like you just commited.

geoffthemedio:
OK, then I won't do any more and feel free to revert when an alternative is implemented.

PR #2723, opened 2020-01-24
corrected Colony Base hull description re slots

l29ah:


default/stringtables/en.txt
```diff
@@ -14289,7 +14289,7 @@ SH_COLONY_BASE
 Colony Base Hull
 
 SH_COLONY_BASE_DESC
-'''A hull designed to create a colony on another planet in system. It cannot move between systems or move quickly in-system and consequently has only one internal slot.
+'''A hull designed to create a colony on another planet in system. It cannot move between systems or move quickly in-system and consequently has only three internal slots.
```
TheSilentOne1, line 14292: "... and consequently has only three internal slots": this doesn't really make sense to me. Should we just leave it at "It cannot move between systems."?

PR #2728, opened 2020-01-26
Design Add Rule

### LINKED ISSUE #2725: Allow players to disable getting default ship designs for people who prefers to use own ship designs
Some people like me prefer personalized galactic warfare. Some people like using own custom ship designs tailored to personal style of gameplay or situation in current universe. However in this case, getting default ship designs gets in way. Every time I research new technology I get design or two and it clogs ship list fast. Especially new hull types gives way too much designs. It would be very nice to have in settings toggle that would disable getting default designs at all

geoffthemedio:
Adds rule to control whether to add default designs to human players.

Should resolve #2725

### DIFF
--- a/default/python/universe_generation/empires.py
+++ b/default/python/universe_generation/empires.py
@@ -448,7 +448,7 @@ def compile_home_system_list(num_home_systems, systems, gsd):
     return home_systems
 
 
-def setup_empire(empire, empire_name, home_system, starting_species, player_name):
+def setup_empire(empire, empire_name, home_system, starting_species, player_name, add_premade_ship_designs=True):
     """
     Sets up various aspects of an empire, like empire name, homeworld, etc.
     """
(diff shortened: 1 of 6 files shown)


geoffthemedio:
@Morlic-fo poke

Morlic-fo:
Will hopefully find time to look into this tomorrow.

Morlic-fo (approved):
It looks like AI shouldn't have any problems with this.

geoffthemedio:
As far as I can tell, the current code adds a bunch of ship designs in Python, then optionally removes them based on the human client option. Apparently the AI doesn't use the designs either, so I'd rather rework things so the option enables adding the designs, rather than removing them client side. The rule itself would be unnecessary, though.

o01eg:
@geoffthemedio poke

geoffthemedio:
I'd like to fix this properly, but probably not until after the next release...

PR #2733, opened 2020-01-28
[AI] Add mypy config scripts and ignore comment for non game files

Cjkjvfnby:
Another tool that can help with python3 migration. 

I don't like its output, it has a lot of false error reported. I am not planning to fix them all before the migration.

It still highligh an important class of errors:  Python 2 and Python 3 will produce a different result for 3/2. If we use integer division intentionally in our code it will work in Python 3 in a slightly different way. I want to be sure that we checked all such places.

Here is a report example:
```
AI\MilitaryAI.py:412: error: Incompatible types in assignment (expression has type "float", base class "Allocator" defined the type as "int")
...
Found 228 errors in 28 files (checked 67 source files)
```

After the migration, I plan to tune the check to make it pass on python3 and add it as a mandatory code check with the help of Travis.


Morlic-fo:
I have no opinion on this. If this helps you, go ahead.

geoffthemedio:
@Cjkjvfnby Do you need someone else to comment or do the merge?

PR #2734, opened 2020-01-30
Design mainwindow layout

rafafelp:
I find this layout more natural (buttons at the bottom right). And it can be squeezed in less space.

Ideally I think add and update should be a single button that changes when the name is new, with no text just an icon. And the panels should be moved (pedia moved left of main panel and parts spread laterally).

### DIFF
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -3948,6 +3948,8 @@ public:
     /** Replace an existing design.*/
     void ReplaceDesign();
 
+    void ToggleDescriptionEditor();
+  
     void HighlightSlotType(std::vector<ShipSlotType>& slot_types);   //!< renders slots of the indicated types differently, perhaps to indicate that that those slots can be drop targets for a particular part?
 
     /** Track changes in base type. */
@@ -4015,8 +4017,8 @@ private:
     std::shared_ptr<GG::StaticGraphic>          m_background_image = nullptr;
     std::shared_ptr<GG::Label>                  m_design_name_label = nullptr;
     std::shared_ptr<GG::Edit>                   m_design_name = nullptr;
-    std::shared_ptr<GG::Label>                  m_design_description_label = nullptr;
-    std::shared_ptr<GG::Edit>                   m_design_description = nullptr;
+    std::shared_ptr<GG::StateButton>            m_design_description_toggle = nullptr;
+    std::shared_ptr<GG::MultiEdit>              m_design_description_edit = nullptr;
     std::shared_ptr<GG::Button>                 m_replace_button = nullptr;
     std::shared_ptr<GG::Button>                 m_confirm_button = nullptr;
     std::shared_ptr<GG::Button>                 m_clear_button = nullptr;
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Attaching after and before screenshots of the layout to a pull request like thos would be appreciated.

rafafelp:
Screenshot from 2020-01-31 14-20-50

Here is a screenshot after. The different panels have been moved around from the default config, the changes to the code only affect what's inside the main panel. In this pic the description is open, if you clic the button again it's hidden.

adrianbroher:
That's better.  Most other windows haver their confirm button also at the bottom right.

geoffthemedio:
In a related matter, feel free to post a pull request that moves the search box and navigation buttons for the pedia to the top of the text area...

UI/DesignWnd.cpp
```diff
@@ -4504,45 +4507,38 @@ void DesignWnd::MainPanel::DoLayout() {
...
-
-    GG::X edit_right = ClientWidth();
-    GG::X confirm_right = ClientWidth() - PAD;
-
-    GG::Pt lr = GG::Pt(confirm_right, BUTTON_HEIGHT) + GG::Pt(GG::X0, GG::Y(PAD));
-    GG::Pt ul = lr - GG::Pt(BUTTON_WIDTH, BUTTON_HEIGHT);
-    m_confirm_button->SizeMove(ul, lr);
+	
+	GG::Pt ul,lr,ll,ur,mus;
+	lr = ClientSize() - GG::Pt(GG::X(PAD), GG::Y(PAD));
```
geoffthemedio, line 4513: replace tabs with spaces

UI/DesignWnd.cpp
```diff
@@ -4965,6 +4961,10 @@ void DesignWnd::MainPanel::ReplaceDesign() {
     DesignChangedSignal();
 }
 
+void DesignWnd::MainPanel::ToggleDescriptionEditor() {
+  if (m_design_description_toggle->Checked()) { m_design_description_edit->Show() ; }
```
geoffthemedio, line 4965: indent 4 spaces

UI/DesignWnd.cpp
```diff
@@ -4504,45 +4507,38 @@ void DesignWnd::MainPanel::DoLayout() {
...
-    lr.x = edit_right;
-    m_design_name->SizeMove(ul, lr);
+	ul= GG::Pt(m_design_name_label->RelativeLowerRight().x+PAD, GG::Y(PAD));
+    m_design_name->SizeMove(ul, GG::Pt(GG::X(ClientWidth()-PAD), ul.y+m_design_name->MinUsableSize().y));
 
-    ul.x = GG::X(PAD);
-    ul.y += (m_design_name->Height() + PAD);
-    lr = ul + GG::Pt(LABEL_WIDTH, m_design_name->MinUsableSize().y);
-    m_design_description_label->SizeMove(ul, lr);
+	ul=GG::Pt(GG::X(PAD), GG::Y(m_design_name->RelativeLowerRight().y+PAD));
```
geoffthemedio, line 4533: prefer spaces around `=`

UI/DesignWnd.cpp
```diff
@@ -4504,45 +4507,38 @@ void DesignWnd::MainPanel::DoLayout() {
...
-    const GG::X BUTTON_WIDTH = PTS_WIDE*GUESSTIMATE_NUM_CHARS_IN_BUTTON_TEXT;
-
-    GG::X edit_right = ClientWidth();
-    GG::X confirm_right = ClientWidth() - PAD;
-
-    GG::Pt lr = GG::Pt(confirm_right, BUTTON_HEIGHT) + GG::Pt(GG::X0, GG::Y(PAD));
-    GG::Pt ul = lr - GG::Pt(BUTTON_WIDTH, BUTTON_HEIGHT);
-    m_confirm_button->SizeMove(ul, lr);
+	
+	GG::Pt ul,lr,ll,ur,mus;
```
geoffthemedio: "mus" refers to what?

geoffthemedio:
This indirectly introduces a crash bug, when entering text into the description field which includes some formatting tags like `<i>`. I'm not sure exactly how to reproduce, but it seems to involve erasing all the text, then starting typing with `<`.

The text editing with such formatting tags is also quite quirky even when it doens't crash, as they are hidden and take-effect in the text as it's being edited, which is problematic as there's no way to see them. The current master `CUIEdit` appears to not process the formatting tags in its input text, so doesn't cause the issue.

rafafelp:
I'll try to test with adding GG::Font::ClearKnownTags(). But according to the doc <i> is among the built-in tags.

geoffthemedio:
I added some safety checks in the function that was crashing (https://github.com/freeorion/freeorion/commit/78af64d27a88f4825ba5e244487e7bd30f797eed) but the issue still appears when the description text in the pedia for the incomplete design has errors with parsing formatting tags. But revert that commit if you want to reproduce the crash.

rafafelp:
Oh, \<i> is among the builtin tags on github too...

It didn't crash on my test but anyway it's better to deactivate them with SetTextFormat. I made an other change to add a button to change that (although if tag handling can crash, maybe we shouldn't propose it). I'm not sure how to commit the new change? I guess I need to pull your changes first?

geoffthemedio:
Fetch and rebase your branch on master.

The box to edit descriptions needs to ignore formatting tags entirely, just showing them as raw text, I think, independent of issues like crashes when editing text in a box that also displays it with the tags enacted...

rafafelp:
I tried `git fetch` and `git rebase` and it didn't do anything, I probably have something to configure. I don't feel like investigating this at the moment, can you just add `m_design_description_edit->SetTextFormat(m_design_description_edit->GetTextFormat() | GG::FORMAT_IGNORETAGS);`
just after its creation (in MainPanel::CompleteConstruction)?

geoffthemedio:
> just add `m_design_description_edit->SetTextFormat(m_design_description_edit->GetTextFormat() | GG::FORMAT_IGNORETAGS);`

Seems to work.

geoffthemedio:
Pushed directly to master https://github.com/freeorion/freeorion/commit/c9eff09dde78aaf39db6b32fc7c6773811ca3da2

PR #2736, opened 2020-02-02
Expose option to append fleet move orders.

aschuch247:
The internally used method already supports appending of fleet move
orders, but the 'append' parameter is not exposed to the Python
interface. Add this parameter with a default value to keep
compatibility.

### DIFF
--- a/python/AI/AIWrapper.cpp
+++ b/python/AI/AIWrapper.cpp
@@ -129,8 +129,11 @@ namespace {
     int IssueChangeFocusOrder(int planet_id, const std::string& focus)
     { return Issue<ChangeFocusOrder>(planet_id, focus); }
 
-    int IssueFleetMoveOrder(int fleet_id, int destination_id)
+    int IssueFleetMoveOrder2(int fleet_id, int destination_id)
     { return Issue<FleetMoveOrder>(fleet_id, destination_id); }
+
+    int IssueFleetMoveOrder3(int fleet_id, int destination_id, bool append)
+    { return Issue<FleetMoveOrder>(fleet_id, destination_id, append); }
 
     int IssueColonizeOrder(int ship_id, int planet_id)
     { return Issue<ColonizeOrder>(ship_id, planet_id); }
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Have you tested using this from Python? I'm unsure if the API generating code will properly handle an optional parameter like this.

Morlic-fo:
https://www.boost.org/doc/libs/1_62_0/libs/python/doc/html/tutorial/tutorial/functions.html#tutorial.functions.default_arguments
> Boost.Python wraps (member) function pointers. Unfortunately, C++ function pointers carry no default argument info.

You want to use ```BOOST_PYTHON_FUNCTION_OVERLOADS``` macro as explained in the link above to correctly handle default arguments.

aschuch247:
@geoffthemedio: I tested this. I recompiled FreeOrion, and it works as expected.

If I undo the change, I get:

> [error] ai : CommonFramework.cpp:153 : ArgumentError: Python argument types in
    freeOrionAIInterface.issueFleetMoveOrder(int, int, bool)
did not match C++ signature:
    issueFleetMoveOrder(int, int)

If I remove the third parameter when calling `issueFleetMoveOrder()` from Python, the error goes away, but appending does not work - the fleet gets killed by the monster due to taking an unexpected route.

> [error] ai : CommonFramework.cpp:153 : ArgumentError: Python argument types in
    freeOrionAIInterface.issueFleetMoveOrder(int, int)
did not match C++ signature:
    issueFleetMoveOrder(int, int, bool)

Okay, I now see that the default parameter does not work. There are just a handful of calls to `issueFleetMoveOrder()`, so that those can be adjusted for a third parameter.

geoffthemedio:
Apparently there are also some changes to the Python scripts that would need to be committed along with the API change, then.

aschuch247:
@Morlic-fo: I did not get `BOOST_PYTHON_FUNCTION_OVERLOADS` working. I failed at

> def("foo", foo, foo_overloads());

Because the third argument is a text already. But I succeeded with the 'manual' method of registering the method twice, once with 2 and once with 3 parameters.

I updated the merge request. How about this?

geoffthemedio:
@Morlic-fo could you comment further?

python/AI/AIWrapper.cpp
```diff
@@ -202,7 +205,8 @@ namespace FreeOrionPython {
         def("updateResearchQueue",                  AIInterface::UpdateResearchQueue);
         def("updateProductionQueue",                AIInterface::UpdateProductionQueue);
 
-        def("issueFleetMoveOrder",                  IssueFleetMoveOrder, "Orders the fleet with indicated fleetID (int) to move to the system with the indicated destinationID (int). Returns 1 (int) on success or 0 (int) on failure due to not finding the indicated fleet or system.");
+        def("issueFleetMoveOrder",                  IssueFleetMoveOrder2, "Orders the fleet with indicated fleetID (int) to move to the system with the indicated destinationID (int). Returns 1 (int) on success or 0 (int) on failure due to not finding the indicated fleet or system.");
+        def("issueFleetMoveOrder",                  IssueFleetMoveOrder3, "Orders the fleet with indicated fleetID (int) to move to the system with the indicated destinationID (int). Append (true) the destination to an already present route, or set the destination as route (false). Returns 1 (int) on success or 0 (int) on failure due to not finding the indicated fleet or system.");
```
Cjkjvfnby: Can we just name it `appendFleetMoveOrder`?

aschuch247: I now renamed the methods. My AI still works as expected (after adjusting the method call).

Morlic-fo:
> Because the third argument is a text already.

There is an [example provided in boost docs](https://www.boost.org/doc/libs/1_62_0/libs/python/doc/html/reference/function_invocation_and_creation/boost_python_overloads_hpp.html#function_invocation_and_creation.boost_python_overloads_hpp.example)
```
def("f", f,
        f_overloads(
            args("x", "y", "z"), "This is f's docstring"
        ));
```

Manual overloading looks fine though.

PR #2740, opened 2020-02-03
Force py3 build

xnox:
Trying to force a python3 build.

Had to port Init functions for python3 builds.
Had to port make_versioncpp.py to python3.

Then tweaked CMake to force using python3 for everything.

The game builds! It even starts up to the main menu! But then starting a single user game fails. Somehow I think that boost-python/AI stuff is broken, as I get an error that I am "disconnected from server" instead of the world getting populated / game starting.

Is freeorion capable of operating in pure python3? If not, I shall be requesting removal of freeorion from Ubuntu 20.04 LTS due in April.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,12 +227,13 @@ endif()
 ## Collect project dependencies.
 ##
 
-set(MINIMUM_PYTHON_VERSION 2.7)
+set(MINIMUM_PYTHON_VERSION 3.6)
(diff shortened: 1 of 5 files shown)


geoffthemedio:
> Is freeorion capable of operating in pure python3?

Not presently.

Python3 migration is ongoing, such as in https://github.com/freeorion/freeorion/pull/2653

PR #2741, opened 2020-02-04
Use boost::ranges for ObjectMap::all function

adrianbroher:
This PR:

* prevents the generation of an intermediate std::vector for ObjectMap::all calls and returns an boost range instead.
* Accepts a boost::any_range<int> as an input for the ObjectMap::find function to prevent code duplication.
* Drops the ObjectMap::{begin,end} interface and replaces it with the corresponding ObjectMap::{all,find} calls.

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -2723,7 +2723,7 @@ void MapWnd::InitTurn() {
     ObjectMap& objects = Objects();
 
     TraceLogger(effects) << "MapWnd::InitTurn initial:";
-    for (auto obj : objects)
+    for (auto obj : objects.all())
         TraceLogger(effects) << obj->Dump();
 
     timer.EnterSection("system graph");
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -473,9 +473,7 @@ namespace {
                 return;
 
             const auto& destroyed_objects = GetUniverse().EmpireKnownDestroyedObjectIDs(m_empire_id);
-            for (auto ship_it = Objects().begin<Ship>(); ship_it != Objects().end<Ship>(); ++ship_it) {
-                auto ship = *ship_it;
-
+            for (auto& ship : Objects().all<Ship>()) {
                 if (!ship->OwnedBy(m_empire_id) || destroyed_objects.count(ship->ID()))
                     continue;
                 m_values[FLEET_DETAIL_SHIP_COUNT]++;
@@ -2753,8 +2751,7 @@ void MapWnd::InitTurn() {
 
     timer.EnterSection("fleet signals");
     // connect system fleet add and remove signals
-    for (auto sys_it = objects.begin<System>(); sys_it != objects.end<System>(); ++sys_it) {
-        auto system = *sys_it;
+    for (auto& system : objects.all<System>()) {
         m_system_fleet_insert_remove_signals[system->ID()].push_back(system->FleetsInsertedSignal.connect(
             boost::bind(&MapWnd::FleetsInsertedSignalHandler, this, _1)));
         m_system_fleet_insert_remove_signals[system->ID()].push_back(system->FleetsRemovedSignal.connect(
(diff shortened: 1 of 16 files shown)


adrianbroher:
ping

geoffthemedio:
Why do you prefer to require `all()` instead of having that be the result of iterating over just `Objects()` ?

adrianbroher:
It's inconsistent and cannot be templated. It's an eyesore.

python/UniverseWrapper.cpp
```diff
@@ -76,9 +76,10 @@ namespace {
     std::vector<int> ObjectIDs(const Universe& universe)
     {
         std::vector<int> result;
-        result.reserve(universe.Objects().size());
-        for (auto it = universe.Objects().begin<T>(); it != universe.Objects().end<T>(); ++it)
-            result.push_back(it->ID());
+        auto objs = universe.Objects().all<T>();
```
geoffthemedio: Is there a purpose or need for storing the result of calling all in a local rather than putting the all call into the loop declaration directly?

universe/UniverseGenerator.cpp
```diff
@@ -658,9 +658,11 @@ void GenerateStarlanes(int max_jumps_between_systems, int max_starlane_length) {
     std::map<int, std::set<int>> potential_system_lanes;
 
     // get systems
-    auto sys_vec = Objects().all<System>();
+    auto sys_rng = Objects().all<System>();
+    std::vector<std::shared_ptr<System>> sys_vec;
```
geoffthemedio: sys_vec seems redundant now

util/ModeratorAction.cpp
```diff
@@ -157,8 +157,7 @@ namespace {
     std::string GenerateSystemName() {
         static std::vector<std::string> star_names = UserStringList("STAR_NAMES");
 
-        const ObjectMap& objects = Objects();
-        std::vector<std::shared_ptr<const System>> systems = objects.all<System>();
+        auto systems = Objects().all<System>();
```
geoffthemedio: Can probably iterate over systems directly with `all<System>` below without this

python/UniverseWrapper.cpp
adrianbroher: I usually do those mass substitutions via `sed` and let the compiler decide if the change was good or not.  When I see some place for optimization when I check in I usually do so, but I miss sometimes stuff.

geoffthemedio: I assumed something like that, but thought I should ask in case there was a reason for it still.

util/ModeratorAction.cpp
adrianbroher: See d9fb4f9bf58fb845bf4547dd4b9e1f53bc7df8e0

python/UniverseWrapper.cpp
adrianbroher: See d9fb4f9bf58fb845bf4547dd4b9e1f53bc7df8e0

UI/MapWnd.cpp
```diff
@@ -2753,13 +2753,37 @@ void MapWnd::InitTurn() {
     // connect system fleet add and remove signals
     for (auto& system : objects.all<System>()) {
         m_system_fleet_insert_remove_signals[system->ID()].push_back(system->FleetsInsertedSignal.connect(
-            boost::bind(&MapWnd::FleetsInsertedSignalHandler, this, _1)));
+            [this](const std::vector<std::shared_ptr<Fleet>>& fleets) {
```
geoffthemedio, line 2756: is this relevant to this commit?

geoffthemedio: I don't object, but it seems out of place given the commit message...

adrianbroher: Yes, changing to boost::range would have required to update the method signature of the now-removed methods. As those were only called here I moved the logic into this place to avoid figuring out the correct signature for those methods.

PR #2742, opened 2020-02-04
Replace boost::timer

geoffthemedio:
As reported [here](https://www.freeorion.org/forum/viewtopic.php?f=9&t=11501) `boost::timer` is deprecated. This adapts `ScopedTimer` to replace it.

### DIFF
--- a/UI/TechTreeWnd.cpp
+++ b/UI/TechTreeWnd.cpp
@@ -12,6 +12,7 @@
 #include "../util/Logger.h"
 #include "../util/OptionsDB.h"
 #include "../util/Directories.h"
+#include "../util/ScopedTimer.h"
 #include "../universe/Tech.h"
 #include "../universe/Effect.h"
 #include "../universe/ValueRef.h"
@@ -28,8 +29,6 @@
 #include <GG/StaticGraphic.h>
 
 #include <algorithm>
-
-#include <boost/timer.hpp>
 
 namespace {
     const std::string RES_PEDIA_WND_NAME = "research.pedia";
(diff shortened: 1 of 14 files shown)


UI/TechTreeWnd.cpp
```diff
@@ -1868,9 +1869,9 @@ void TechTreeWnd::TechListBox::Update(bool populate /* = true */) {
         auto& tech_row = row.second;
         if (TechVisible(tech_row->GetTech(), m_categories_shown, m_tech_statuses_shown)) {
             tech_row->Update();
-            insertion_timer.restart();
+            insertion_timer.start();
             auto listbox_row_it = Insert(tech_row);
-            insertion_elapsed += insertion_timer.elapsed();
+            insertion_elapsed += ElapsedSeconds(insertion_timer);
```
geoffthemedio: probably need an optional parameter to control rounding to avoid issues when accumulating the results like this...

geoffthemedio:
@adrianbroher @o01eg Some help with CMake would be appreciated. Apparently adding boost timer to the linked and required libs fixes the XCode build but breaks GCC? (Assuming the Travis CI build for XCode depends on CMake defs, which perhaps it doesn't, in which case I'm even more confused...)

adrianbroher:
> @adrianbroher @o01eg Some help with CMake would be appreciated. Apparently adding boost timer to the linked and required libs fixes the XCode build but breaks GCC? (Assuming the Travis CI build for XCode depends on CMake defs, which perhaps it doesn't, in which case I'm even more confused...)

Seems like the boost timer library is not installed within the Docker image of the SDK.

o01eg:
@geoffthemedio @adrianbroher I suppose we should test boost version before add timer component. Also, should we add this PR to release?

adrianbroher:
Why? The CI setup has the package not installed, it's not like it is available in older boost releases.

adrianbroher:
@geoffthemedio Why don't you use the Scoped{,Section}Timer?  I don't see any place in this change where it could not be used.

geoffthemedio:
> @geoffthemedio Why don't you use the Scoped{,Section}Timer?

Done. Needed to add to the interface of `ScopedTimer` to be equivalent to `boost::timer`.

Also adapted `SectionedScopedTimer` to replace to re-use of `ScopedTimer`.

Also tweaked output formatting and options.

PR #2743, opened 2020-02-05
cmake: make make_versioncpp.py be py2&3 compatible

xnox:


geoffthemedio:
Works for me. @Morlic-fo please comment if there are any issues...

PR #2744, opened 2020-02-05
cmake: Fix config for CMP0093

xnox:
See https://cmake.org/cmake/help/v3.15/policy/CMP0093.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,7 +246,15 @@ find_package(Boost ${MINIMUM_BOOST_VERSION}
         system
         thread
     REQUIRED)
-if(${Boost_VERSION} GREATER 106699) # boost >= 1.67
+
+# New CMake correct how Boost versions are reported
+if(POLICY CMP0093)
+  cmake_policy(SET CMP0093 NEW)
+  set(BOOST167 1.66.99)
+else()
+  set(BOOST167 106699)
+endif()
+if(${Boost_VERSION} GREATER ${BOOST167}) # boost >= 1.67
     find_package(Boost COMPONENTS python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR} REQUIRED)
     set(Boost_PYTHON_LIBRARY ${Boost_PYTHON${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}_LIBRARY})
 else()


xnox:
Without this v0.4.9 will fail to build from source on Debian Ubuntu and all other up to date linux distributions.

geoffthemedio:
@adrianbroher or @agrrr3 any comment?

adrianbroher:
Icky.  Won't merge as there is no explicit policy selection for OLD or NEW behaviour in the patch.

adrianbroher:
Also the change in CMake is too recent for me, so setting the policy to OLD would my preferred solution.

xnox:
> Also the change in CMake is too recent for me, so setting the policy to OLD would my preferred solution.

But implementation of OLD may and will be removed at any time.

I guess i can add a `get()` call to check what the policy is set to, if it is available to cover all cases.

xnox:
Also not that this will safe freeorion from removal in debian/ubuntu as both are dropping python2, and freeorion fails to work under boost-python38 / python3 only.

adrianbroher:
> Also not that this will safe freeorion from removal in debian/ubuntu as both are dropping python2, and freeorion fails to work under boost-python38 / python3 only.

We are aware of that.  However this fact will not change the current schedule.

adrianbroher:
> But implementation of OLD may and will be removed at any time.

Yes, and what is your point?  The CMake setup is already convoluted as it is.  I don't want to see any more crude fixes in there.

xnox:
See my updated merge proposal that works with old/new cmakes, irrespective of cmake preset policy and is forward-proof.

geoffthemedio:
@adrianbroher any better now?

adrianbroher:
No. The same can be achieved with the _MAJOR, _MINOR variables in a single line and without caring for CMP0093.

Vezzra:
@xnox,

> Without this v0.4.9 will fail to build from source on Debian Ubuntu and all other up to date linux distributions.

As @adrianbroher said, we are aware of this. The problem is, getting FO to work with Python3 requires a lot of adjustments, as all the AI, universe generation and event turn scripts are Python. Migrating the interface code as well as all the scripts is a non-trivial task we decided not to do for 0.4.9, and accept the fact that this will get FO removed from all Linux distros that do not support Python2 anymore (for the time being).

Getting the Python3 migration into 0.4.9 would cause another huge delay for that release, which is already massively delayed (we wanted to get that release out almost a year ago originally!).

We decided to dedicate the entire next release cycle to the Python3 migration, which we will try to get done asap, so FO can again be included in all Linux distros.

Until then, players on Linux are recommended to use our FO snap package. The release announcement will include a respective note.

Vezzra:
Superseded by #2751.

PR #2702, opened 2020-01-11
Separate focs base class declarations into dedicated header files

adrianbroher:
This PR moves the ScriptingContext, ValueRef, Condition and Effect base class declarations into seperate headers and removes the `Base` suffix.

PR #2747, opened 2020-02-09
Move AI client, general client and server sources to place of use

adrianbroher:
This PR moves some of the randomly sprinkled source files to the directory of the corresponding component.

```
$ git diff --summary master 
 delete mode 100644 AI/CMakeLists.txt
 rename {python => client}/AI/AIFramework.cpp (98%)
 rename {python => client}/AI/AIFramework.h (78%)
 rename {AI => client/AI}/AIInterface.cpp (96%)
 rename {AI => client/AI}/AIInterface.h (99%)
 rename {python => client}/AI/AIWrapper.cpp (99%)
 rename {python => client}/AI/AIWrapper.h (71%)
 rename {network => client}/ClientNetworking.cpp (99%)
 rename {network => client}/ClientNetworking.h (99%)
 delete mode 100644 python/AI/CMakeLists.txt
 delete mode 100644 python/server/CMakeLists.txt
 delete mode 100644 python/server/ServerWrapper.h
 rename {python/server => server}/ServerFramework.cpp (98%)
 rename {python/server => server}/ServerFramework.h (90%)
 rename {network => server}/ServerNetworking.cpp (100%)
 rename {network => server}/ServerNetworking.h (99%)
 rename {python/server => server}/ServerWrapper.cpp (98%)
 create mode 100644 server/ServerWrapper.h
```


geoffthemedio:
You don't think it's reasonable to keep all the Python API code in the same place?

adrianbroher:
> You don't think it's reasonable to keep all the Python API code in the same place?
No.

Feel free to answer the following questions to yourself:

* Who is the corresponding consument of the moved python wrappers? Where does the component live inside of the source tree?
* Does not moving reduce or improve the API or ABI of the appication components?
* Do the python wrappers need specific compiler flags that can be better managed when the Python files are bundled in a directory?

geoffthemedio:
> Who is the corresponding consument of the moved python wrappers? Where does the component live inside of the source tree?

My thought was that the current organization isn't random as asserted; it uses what-is-it rather than what-uses-it for top-level categorizing of the source files. The latter keeps similar code together, which can be useful. There are also various common / shared files as part of the Python API that are also similar and which don't make sense in the AI or server directories, so presumably there will still be a python dir with those in it, and it might be confusing if there are other python API files located elsewhere.

But I don't object to the reorganization if you prefer it that way... just wanted to know why. Location makes no difference to me for compiler flag management, and I don't think the ABI / API will be affected regardless.

adrianbroher:
> There are also various common / shared files as part of the Python API that are also similar and which don't make sense in the AI or server directories, so presumably there will still be a python dir with those in it, and it might be confusing if there are other python API files located elsewhere.

Guess what: there is already a component called common which would be an excellent place for for those file.  However the sources for that common componet are sprayed all over the place which was not my doing.  But thanks to those continous efforts of sticking to arbitary boundaries (like what library is used or how the file is formatted) that reflect in the source tree we can now enjoy things like compiling the same code twice or even more often, finding high level logic right beside network message calls in the UI and seeing the very same but subtle differnt logic code in 3 different locations in both clients and the server.  Or having 2 different mechanism to send instructions from clients to servers.  Or having two incompatible filter mechanism for game entities.  Or having tons of unneccessary cacading includes, because it is a good idea™ to put the base classes and their 500 different subclasses into the same header to let every translation unit rebuild when even the smallest change in a header happended to put the compile times into new, unknown heights.  Or placing very specific game specific components into the `utilities` directory.

adrianbroher:
Also lets don't forget the fact that its almost impossible to do any more advanturous changes like trying another library with possible better performance thanks to the fact that its completely neccessary to maintain 3 different build systems.

geoffthemedio:
> there is already a component called common which would be an excellent place for for those file.

Not sure I follow... We're talking about directory sorting of source files, not what code should be compiled into the common library vs. into the separate executable binaries.

But regardless of that, the Python API isn't used in the human client, which I think is why those files aren't put into the common library. If that can be done without introducing a direct Python dependence into the human client, then I guess it'd be fine.

> put the base classes and their 500 different subclasses into the same header

Assuming that refers, at least in part, to https://github.com/freeorion/freeorion/pull/2702, I will merge that. As noted, I wanted to delay it to get the government branch merged first, but it looks like government will be delayed another release so we can get Python3 support out faster. As such, the base class separate can be re-prioritized ahead of government instead.

> its almost impossible to do any more advanturous changes like trying another library with possible better performance thanks to the fact that its completely neccessary to maintain 3 different build systems.

Please don't let the existence of MSVC or XCode project files stop you... If you find a better alternative, we can adapt the non-CMake project files as needed; that doesn't need to be done by you.

Also, to reiterate: You've complained extensively about the inability to generate MSVC project files from CMake due to maintaining them separately and related directory organization and such, but I have no objection to generating them from CMake and the related reorganization. I just want the MSVC project files to also be committed after being generated, so that people using MSVC don't need to use CMake to generate the project files. There's no specific requirement to maintain them independent of CMake, though.

adrianbroher:
> I just want the MSVC project files to also be committed after being generated, so that people using MSVC don't need to use CMake to generate the project files.

6+ years and still not grasping what cmake is or what it does.

Vezzra:
> I just want the MSVC project files to also be committed after being generated, so that people using MSVC don't need to use CMake to generate the project files.

Why? CMake isn't *that* hard to use? I don't see the problem with expecting anyone who wants to build FO by themselves to be able to handle CMake. Keeping files which get generated (for whatever purpose by whatever means) in version control kind of defeats the purpose of generating them...?

I mean, there is a reason why it's usually not done...

geoffthemedio:
The point of generating them from CMake is/would be that one only needs to edit CMake files to do the project file updates for all build systems, including windows, and can do so even without having MSVC to open and do the edits.

The point of generating the MSVC project files and committing them is that people who are using MSVC don't have to generate new project files whenever the CMake files change; the updated files will already be present at whatever commit one is currently on.

I frequently jump around in the commit log and switch between branches. If I have to check every time whether I need to generate new project files, and then do so, that is a major delay and disruption in workflow. If I can just switch branches and be sure that the MSVC project files are already in working order, then that's much easier to work with. To clarify, I work within MSVC mostly, not command line except for some git stuff that doesn't work well within TortoiseGit (mainly pulling branches from pull requests), so I'm not switching branches using console command most of the time, and expect MSVC to immediately update itself after I switch branches or revert to a previous commit using the popup menus.

Having the MSVC project files already available also means that someone who wants to work with them doesn't need to install and use CMake. It's not a matter of whether they can learn to do so, but rather a matter of it being an annoying hassle to need to do so.

Apparently I don't understand CMake, so perhaps the build happens with XCode and GCC without a separate project file generation step. Even if that's the case, it doesn't matter for me, as I'd still need to generate the project files as a separate step in order to see the result of changing the project files. Even if MSVC can be configured to run CMake when building a project or a solution, that doesn't fix the issue, as I want Visual Studio to update before I start a build.

adrianbroher:
> If I have to check every time whether I need to generate new project files, and then do so, that is a major delay and disruption in workflow.

The solution is refreshed on every build when required without manual intervention.  MSVC will at most ask for a reload of the solution.

> Having the MSVC project files already available also means that someone who wants to work with them doesn't need to install and use CMake. It's not a matter of whether they can learn to do so, but rather a matter of it being an annoying hassle to need to do so.

So its the best course of action to load the additional effort required on long time contributors and increasing the maintaince effort for everything related around the build system like documentation, consistency accross platforms and automation efforts. Add another dimension of pain by 'supporting' multiple versions of MSVC at the same time. 10 out of 10 prioritisation.

> Apparently I don't understand CMake, so perhaps the build happens with XCode and GCC without a separate project file generation step.

Again, apart from the initial build tree, this is done without manual intervention. This is done on every platform and switching between branches is handled just fine.

> Even if that's the case, it doesn't matter for me, as I'd still need to generate the project files as a separate step in order to see the result of changing the project files.

What kind of change do you expect to see that needs to be reviewed?

o01eg:
With CMake we have other issue, it requires more time to build on AppVeyor than MSVS project.

Vezzra:
This isn't really the place for a discussion about how to handle the CMake/MSVC project files on Windows. I have opened a [thread on our forum](https://www.freeorion.org/forum/viewtopic.php?f=9&t=11535) and copied the relevant posts there, please continue the discussion in that thread.

PR #2750, opened 2020-02-11
Ignore failures of unittest

o01eg:


adrianbroher:
Please drop this althogether.  The current "test" only reduces confidence in the CI by permanently failing.  Also it should not be in the same stage as the build.

o01eg:
@adrianbroher I've moved `unittest` to `after_success` block so fail here will be ignored.

geoffthemedio:
Do the unit tests still get run? I don't see them at all in the tests for this pull request.

o01eg:
@geoffthemedio Yes, it is after `before_cache` block in c++ job.

geoffthemedio:
It doesn't mark the test as failed now, which is better, but it seems overly hidden with no indication unless you go into the full log. Is there some way to mark a test yellow as passing important tests but giving a warning that the low-importance tests failed?

o01eg:
@geoffthemedio No, I don't remember of some semi-failed status in Travis CI.

adrianbroher:
> @adrianbroher I've moved `unittest` to `after_success` block so fail here will be ignored.

I mean that the test should be part of the test stage and not the build stage.  Currently working on a way to use the gdrive as inter-stage storage.

geoffthemedio:
Should this be merged now or do you have a better solution that will be available soon enough to skip this version entirely?

adrianbroher:
No, not soon.  The solution needs some testing and an updated Docker image.

geoffthemedio:
So is this pull request worth merging now to avoid the misleading test failures?

geoffthemedio:
Lacking further response, will merge this now. Better solutions can be applied when ready.

PR #2751, opened 2020-02-11
Simplify CMake buildsystem

adrianbroher:
This PR:

* raises the minimum required version to CMake 3.5, which is the version provided by Ubuntu 16.04.
* Lets CMake decide which is the correct standard library to link on MacOS, when we request a specific C++ version.
* Hide all non explicit exported symbols on unix-like platforms by default, which reduces the library size of GiGi and GiGiSDL (other components had this flag already enabled for some time).
* Use the suggested modern CMake approach of using export targets to define dependencies to external targets instead of manually fiddeling with include directories, compile defines and libraries to link.
* Use a more recent CMake upstream FindBoost.cmake to locate the correct Boost Python component, respecting the Python 2/3 version while keeping platform naming oddities out of sight and mind.


adrianbroher:
Supersedes also #2744

adrianbroher:
@Vezzra please verify if the built .app is still working as intended.

Vezzra:
@adrianbroher,

> please verify if the built .app is still working as intended.

It is 👍

geoffthemedio:
So is this OK to merge then?

o01eg:
@adrianbroher  GG doesn't see Glew headers on windows.

adrianbroher:
CMake is not officially supported on Windows anyway, so what is the point?

geoffthemedio:
Presumably the windows issue can be fixed when needed...

PR #2755, opened 2020-02-12
Use objectmap interface

adrianbroher:
This PR replaces and removes various "convenience" Get* functions and replaces them with the corresponding `ObjectMap::get`, `ObjectMap::find` or `ObjectMap::all` call.  Using a consistent interface prevents misuse (e.g. `GetPlanet` in a for each loop, where `ObjectMap::find<Planet>` would be more consistent) and gives better insight, searchability and comparision capabilities for further API improvements.

### DIFF
--- a/UI/SidePanel.cpp
+++ b/UI/SidePanel.cpp
@@ -682,7 +682,7 @@ public:
 
     void Refresh() {
         ScopedTimer timer("RotatingPlanetControl::Refresh", true);
-        auto planet = GetPlanet(m_planet_id);
+        auto planet = Objects().get<Planet>(m_planet_id);
         if (!planet) return;
 
         // these values ensure that wierd GLUT-sphere artifacts do not show themselves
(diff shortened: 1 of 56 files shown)


UI/SidePanel.cpp
```diff
@@ -1287,10 +1286,12 @@ std::shared_ptr<const Ship> ValidSelectedColonyShip(int system_id) {
         return nullptr;
 
     // is there a valid selected ship in the active FleetWnd?
-    for (int ship_id : FleetUIManager::GetFleetUIManager().SelectedShipIDs())
-        if (auto ship = GetShip(ship_id))
-            if (ship->SystemID() == system_id && ship->CanColonize() && ship->OwnedBy(HumanClientApp::GetApp()->EmpireID()))
-                return ship;
+    for (const auto& ship : Objects().find<Ship>(FleetUIManager::GetFleetUIManager().SelectedShipIDs())) {
```
geoffthemedio, line 1289: This is probably fine in practice, but might warrant a bit more consideration, since getting all the selected ship IDs is not in general necessary, since the first found valid selected colony ship is returned.

adrianbroher: Now with having the ObjectMap in place for most interfaces it would be feasible to check where filter or transformation ranges could be applied.

PR #2759, opened 2020-02-15
Produce error if checksum on the server and the client or AI differs

o01eg:


default/stringtables/en.txt
```diff
@@ -1049,6 +1049,9 @@ Cannot concede from a game with only one human player.
 ERROR_INCOMPATIBLE_VERSION
 Server cannot properly read messages from your client, most likely due to incompatible game versions.
 
+ERROR_CHECKSUM_MISMATCH
+COntent checksum on the server and the client differs.
```
geoffthemedio: cApitalization

client/AI/AIClientApp.cpp
```diff
@@ -401,7 +401,11 @@ void AIClientApp::HandleMessage(const Message& msg) {
 
     case Message::CHECKSUM: {
         TraceLogger() << "(AIClientApp) CheckSum.";
-        VerifyCheckSum(msg);
+        bool result = VerifyCheckSum(msg);
+        if (!result) {
+            ErrorLogger() << "Wrong checksum";
+            throw std::runtime_error("AI got incorrect checksum.");
```
geoffthemedio, line 407: Is this caught somewhere or do you intend to just crash / kill the AI if it has a bad checksum...?

o01eg: I intend to kill AI.

default/stringtables/en.txt

client/AI/AIClientApp.cpp
geoffthemedio: Have you had cases where an AI had a bad checksum? They usually (only?) run on the same machine as the server, I think... so wouldn't expect this to happen.

Would this kill them when setting up a multiplayer game, so attempting to add a buggy AI will fail and drop the connection?

Presumably this would break starting single-player games in the event that an AI got a bad checksum somehow.

o01eg: @geoffthemedio No yet but in case of using alternative content we are testing now it possible. I also remember having issues with checksum in tests which are definitely run on the same machine as the server.

AI launches at game start which will be failed.

PR #2760, opened 2020-02-15
Prune stale translations

adrianbroher:
This PR drops stale translation entries from the unmaintained translation tables.

To identify stale entries one of the following condition applies:
* The entry key does not exist in en.txt
* The entry value is identical to the corresponding en.txt entry
* The entry exists despite the corresponding en.txt entry is only a substitution reference (the whole value is enclosed via `[[]]`).
* The entry value has a different amount of lines than the corresponding en.txt entry

Except from the last condition I would consider those unambigious.  For the last one I would rather remove potentially good (only difference by e.g. whitespace shuffling) over keeping potentially bad translation entries (describing outdated gameplay elements wrongly) as those leave the everlasting first bad impression to players.  Also the application layout expects the format used in the en.txt so its more likely to see bad rendering when using different layouts.

The PR only covers stringtables without any active maintance so the default action was the removal of entries with some exceptions where the change was obviously of cosmetic nature.

### DIFF
--- a/default/stringtables/cz.txt
+++ b/default/stringtables/cz.txt
@@ -1234,10 +1234,6 @@ METER_POPULATION
 populace
 
 # Meter types
-METER_HEALTH
-zdraví
-
-# Meter types
 METER_INDUSTRY
 průmysl
 
--- a/default/stringtables/cz.txt
+++ b/default/stringtables/cz.txt
@@ -151,9 +151,6 @@ COMMAND_LINE_USAGE
 COMMAND_LINE_DEFAULT
 '''Standardní: '''
 
-OPTIONS_DB_HELP
-Vytisknout tuto nápovědu.
-
 OPTIONS_DB_GENERATE_CONFIG_XML
 Využívá všech nastavení z jakékoli stávající config.xml souboru a ty, uvedené na příkazové řádce vytvořit config.xml souboru. Tím se přepíše stávající soubor config.xml, pokud existuje.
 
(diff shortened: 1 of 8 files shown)


geoffthemedio:
@Ouaz do you have any comment?

Vezzra:
@geoffthemedio, the French stringtable is not affected (and it shouldn't be, according to the PR description: "The PR only covers stringtables without any active maintance"), so I'm not sure what @Ouaz could comment on...?

geoffthemedio:
I thought he might have concerns about the criteria for removing stuff, but if you think it's fine, then fine...

Vezzra:
Well, the first three criteria sound reasonable, the last one might be debateable (as @adrianbroher pointed out in the PR description).  I admin I can't offer a opinion on that last criteria myself - @adrianbroher's reasoning makes sense to me, that's all I can say.

As @Ouaz only ever was concerned with the French translation, and never with the more general procedures/handling of the stringtables, I don't expect he'll have much of an opinion on this PR.

Ouaz:
> I thought he might have concerns about the criteria for removing stuff, but if you think it's fine, then fine...

FR file is not affected, so it's OK. Some language stringtables are not maintained since ages, so it won't be an annoyance for possible future translators (they'll have to start the translation almost from scratch anyway, i.e. line by line.)

I don't know for russian translators, but for me, the most important thing is that the number of lines stays exactly the same between EN and FR stringtables, otherwise it would break the diff process, and make the translation update impossible (as the diff would suggest to import all missing lines, even the identical ones which not require a translation).

I wrote a post years ago about how I proceed to update the FR translation, and why it is crucial that the number of lines match between EN and FR files:
https://freeorion.org/forum/viewtopic.php?f=27&t=9463#p77199

Cjkjvfnby:
> The entry value has a different amount of lines than the corresponding en.txt entry

Looks dangerous.  I have never consider that requirement when do a translation.   Can you add this to translation guide?

adrianbroher:
@Cjkjvfnby can you point out the translation guide you are referring to?

> Looks dangerous.

Well it is as noted in the OP.  But currently there is no other way to identify deficiencies aside from per-line git author date comparision and proficient speaking the language in question.  I can do only the former with scripts.

> I have never consider that requirement when do a translation.

That was one of the reasons why I wrote st-tool in the first place: to do some sanity checks within the CI:
* Does <language>.txt contain valid string table syntax?
* Does the <language>.txt contain keys without a matching key in en.txt?
   ⇒ author forgot to rename the key or the entry was removed.
* Are all internal stingtable references within `[[]]` have a corresponding entry?
   ⇒ author forgot to rename the key or the entry was removed.
* Does the layout still match (what @Ouaz noted as "matching number of lines")
  ⇒ the string tables should be synchronized via `st-tool sync`.

Cjkjvfnby:
> That was one of the reasons why I wrote st-tool in the first place: to do some sanity checks within the C

I totally missed intention about using it in CI. 

> author forgot to rename the key 
It should not be an author, it should be a person who renames/deletes key.  

> Does the layout still match 

Person who is far away from the translation added a new line to wiki entry. (`This species know as deadly herbivores` )  .  What should happen with other translations?

PR #2761, opened 2020-02-15
Check both python versions lints

o01eg:


.travis.yml
```diff
@@ -19,6 +19,20 @@ jobs:
...
+      dist: bionic
+      language: python
+      python: 3.5
+      before_install:
+        - pip install flake8-putty==0.4.0
+      before_script:
+        - cd default/python
+      script:
+        - flake8
+        - python -m compileall -q .
```
adrianbroher: Why are you running python2?

.travis.yml
```diff
@@ -19,6 +19,20 @@ jobs:
         - python -m compileall -q .
       env:
         - NAME="Lint Python script"
+    - stage: lint
+      os: linux
+      dist: bionic
+      language: python
+      python: 3.5
+      before_install:
+        - pip install flake8-putty==0.4.0
```
adrianbroher: Why are you installing into the python2 site-packages?

.travis.yml
```diff
@@ -19,6 +19,20 @@ jobs:
...
+      before_script:
+        - cd default/python
+      script:
+        - flake8
+        - python -m compileall -q .
+      env:
+        - NAME="Lint Python script"
```
adrianbroher: You are linting python3 syntax, that should be reflected in the name.

adrianbroher (changes requested):
The change only runs Python 2 linting.

.travis.yml
adrianbroher: Also the env section was just a workaround for the at-that-time missing [job name](https://docs.travis-ci.com/user/build-matrix/#job-names) feature.

.travis.yml
o01eg: Fixed to use python3

.travis.yml
o01eg: Fixed to use python3

.travis.yml
o01eg: Fixed names to distinguish both.

adrianbroher:
LGTM.

PR #2764, opened 2020-02-16
Simplify AI client

adrianbroher:
This PR replaces the `AIBase` interface and subclasses with direct calls from boost::python.  That interface were assumingly used during the C++ to Python AI implementation migration before v0.3 times.


geoffthemedio:
I think the idea was to have an implementation-independent interface for AIs, rather than baking in / assuming that the only supported AI is one implemented in Python.

adrianbroher:
Do you see any non Python-AI implementation?  Do you see a need for a non-Python AI implementation?  What is the gain of having complex code over something that most certainly will never be created?

Keep it simple.

geoffthemedio:
I wasn't objecting to the changes, but explaining the idea behind the old design...

adrianbroher:
> I wasn't objecting to the changes, but explaining the idea behind the old design...

Ah, ok.  Sorry, I misunderstood your reply.

PR #2766, opened 2020-02-17
Python3 compatibility fixes

o01eg:
Fixes error happening in python 3

### DIFF
--- a/default/python/AI/character/character_module.py
+++ b/default/python/AI/character/character_module.py
@@ -575,6 +575,16 @@ def _make_single_function_combiner(funcnamei, f_combo):
     return func
 
 
+def _maxmin_not_none(f_combo):
+    """Make a combiner that collects not None items and applies min or max"""
+    def func(llin):
+        ll = [x for x in llin if x is not None]
+        if not ll:
+            return 0
+        return f_combo(ll)
+    return func
+
+
 # Create combiners for traits that all must be true
 for funcname in ["may_explore_system", "may_surge_industry", "may_maximize_research", "may_invade",
                  "may-invade_with_bases", "may_build_building", "may_produce_troops",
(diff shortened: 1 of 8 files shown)


default/python/universe_generation/starnames.py
```diff
@@ -177,7 +177,7 @@ def name_star_systems(system_list):
     indiv_systems = []
 
     # remove groups with only one non-deep-system
-    for groupindex, group_list in star_groups.items():
+    for groupindex, group_list in list(star_groups.items()):
```
Cjkjvfnby, line 180: Why do we need `list` wrapper for iteration?

default/python/AI/character/character_module.py
```diff
@@ -575,6 +575,16 @@ def func(self, *args, **kwargs):
     return func
 
 
+# Create combiners for traits that max not None results
```
Cjkjvfnby: Use docstring instead of comment

Cjkjvfnby (changes requested):

default/python/universe_generation/starnames.py
o01eg: Because `star_groups` mutates in the loop which result in the error `RuntimeError: dictionary changed size during iteration`

default/python/AI/character/character_module.py
o01eg: Fixed.

Cjkjvfnby (approved):

Cjkjvfnby:
@o01eg , please merge after CI build success.

default/python/universe_generation/starnames.py
```diff
@@ -221,8 +221,8 @@ def name_star_systems(system_list):
     random.shuffle(potential_group_names)
     random.shuffle(individual_names)
     random.shuffle(group_names)
-    num_for_indiv = min(max(len(potential_group_names) / 2, num_individual_stars + 1 - len(individual_names)),
-                        len(potential_group_names))
+    num_for_indiv = int(min(max(len(potential_group_names) / 2, num_individual_stars + 1 - len(individual_names)),
```
Cjkjvfnby, line 224: `len(potential_group_names) / 2` is an example of a place I am looking for. Hardest part of migration. It is floor division in python2 (int) and true division in python3 (float).  In some cases such behavior may be hidden. After I will check/fix al such cases, python migration will be finished.

 I think that changing it to  `len(potential_group_names) // 2`  will be more readable.  Will fix it in my PR.

PR #2767, opened 2020-02-17
Yet another set of python fixes.

o01eg:
I've tried to load game and fix some errors from here.

### DIFF
--- a/default/python/AI/FleetUtilsAI.py
+++ b/default/python/AI/FleetUtilsAI.py
@@ -26,6 +26,10 @@ def stats_meet_reqs(stats, requirements):
     :rtype: bool
     """
     for key in requirements:
+        if key not in stats:  # skip requirements not related to stats
+            if key != "target_system":  # expected not to be in stats
+                warn("Requirement %s not in stats", key)
+            continue
         if stats.get(key, 0) < requirements[key]:
             return False
     return True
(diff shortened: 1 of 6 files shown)


default/python/AI/FleetUtilsAI.py
```diff
@@ -26,7 +26,7 @@ def stats_meet_reqs(stats, requirements):
     :rtype: bool
     """
     for key in requirements:
-        if stats.get(key, 0) < requirements[key]:
+        if key not in stats or stats.get(key, 0) < requirements[key]:
```
Morlic-fo: ```TypeError: '<' not supported between instances of 'int' and 'TargetSystem'``` sounds like an underlying issue with the code which should not be silenced like this.

o01eg: `target_system` key contains `TargetSystem` type. I suppose in python2 `<` returned `False` for such comparsion.

Cjkjvfnby: Split condition to two separate ifs.  First one  `if key not in stats: # skip requirements not related to stats`

o01eg: @Cjkjvfnby Should I use `continue` in this case?

Cjkjvfnby (approved):

default/python/AI/FleetUtilsAI.py
Morlic-fo: ```requirements``` should logically be a subset of ```stats``` and ```target_system``` should not be in the dict but passed around differently. This is beyond the scope of this PR though.

For now, we should probably log a warning.
```
if key not in stats:
    if key != "target_system":  # expected not to be in stats
        warn("Requirement %s not in stats", key)
    continue
```

o01eg: Fixed.

Morlic-fo (approved):

PR #2771, opened 2020-02-18
[AI] Fix pylint warnings

Cjkjvfnby:
- Check if division is true or floor
- Make floor division explicit
- Remove some unused methods (class still have ~900 lines)
- Fix argument for constructor of BasicStats, because `initialMeterValue` always return float

### DIFF
--- a/default/python/AI/CombatRatingsAI.py
+++ b/default/python/AI/CombatRatingsAI.py
@@ -60,8 +60,8 @@ class ShipCombatStats(object):
             :type shields: int|None
             :return:
             """
-            self.structure = 1 if structure is None else structure
-            self.shields = 0 if shields is None else shields
+            self.structure = 1.0 if structure is None else structure
+            self.shields = 0.0 if shields is None else shields
             self.attacks = {} if attacks is None else tuple_to_dict(attacks)  # type: dict[float, int]
 
         def get_stats(self, hashable=False):
@@ -149,7 +149,7 @@ class ShipCombatStats(object):
                         # TODO: Depending on future implementation, might actually need to handle this case.
                         warn("Multiple hangar types present on one ship, estimates expected to be wrong.")
                     fighter_damage = max(fighter_damage, part_damage)
-        self._basic_stats = self.BasicStats(attacks, int(structure), int(shields))
+        self._basic_stats = self.BasicStats(attacks, structure, shields)
         self._fighter_stats = self.FighterStats(fighter_capacity, fighter_launch_rate, fighter_damage)
 
     def get_basic_stats(self, hashable=False):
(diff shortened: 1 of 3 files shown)


default/python/AI/CombatRatingsAI.py
```diff
@@ -148,7 +149,7 @@ def __get_stats_from_ship(self):
                         # TODO: Depending on future implementation, might actually need to handle this case.
                         warn("Multiple hangar types present on one ship, estimates expected to be wrong.")
                     fighter_damage = max(fighter_damage, part_damage)
-        self._basic_stats = self.BasicStats(attacks, structure, shields)
+        self._basic_stats = self.BasicStats(attacks, int(structure), int(shields))
```
Morlic-fo: should fix these in the constructor to be ```float``` instead considering they are floats in the game engine.

default/python/AI/ShipDesignAI.py
```diff
@@ -1760,7 +1740,7 @@ def _starting_guess(self, available_parts, num_slots):
 
     def _calc_rating_for_name(self):
         self.update_stats(ignore_species=True)
-        return self.design_stats.structure*self._total_dmg()*(1+self.design_stats.shields/10)
+        return self.design_stats.structure*self._total_dmg()*(1+self.design_stats.shields//10)
```
Morlic-fo: should be true division.

default/python/AI/ShipDesignAI.py
```diff
@@ -1830,7 +1810,7 @@ def _filling_algorithm(self, available_parts, verbose=False):
         return best_rating, best_partlist
 
     def _calc_rating_for_name(self):
-        base_rating = self.design_stats.structure*self._total_dmg()*(1+self.design_stats.shields/10)
+        base_rating = self.design_stats.structure*self._total_dmg()*(1+self.design_stats.shields//10)
```
Morlic-fo: should be true division.

Morlic-fo (changes requested):

default/python/AI/CombatRatingsAI.py
```diff
@@ -52,7 +53,7 @@ def __init__(self, attacks, structure, shields):
             """
 
             :param attacks:
-            :type attacks: dict|None
+            :type attacks: dict[float, int]|None
```
o01eg, line 56: Is it have sense to use float as a key of dict?

Morlic-fo: Yes, the key is the damage of a weapon which is a float in the game engine.

default/python/AI/CombatRatingsAI.py
Cjkjvfnby: It is documented as int, so class users expect to get int here.   But I don't have any strong preferences here.

default/python/AI/ShipDesignAI.py
Cjkjvfnby: My goal was to make code work same as it.  Lets change it to true division.

default/python/AI/CombatRatingsAI.py
Morlic-fo: int is a bug.

Cjkjvfnby: I have already pushed commit with the fix

PR #2776, opened 2020-02-23
Fix Windows compilation and unittest dependencies

o01eg:


o01eg:
@adrianbroher I've added here search for python27 component if FindBoost works with python27 as well.

Cjkjvfnby:
@o01eg , ping.

o01eg:
I've removed move to `python27`, so it works and can be merged now.

Cjkjvfnby:
@o01eg can you assign a person for review?

o01eg:
It should be @adrianbroher but I cannot set him for review.

Vezzra (commented):
See my comment in freeorion/freeorion-sdk#57:

This PR helped with the issue with the `Boost_PYTHON_LIBRARY` symbol, but now another `cmake -E copy_if_different` operation fails, this time because apparently `Boost_SYSTEM_LIBRARY` isn't properly set (resolves to an empty string, exactly as `Boost_PYTHON_LIBRARY` before).

Vezzra:
@o01eg, now I can build FO with the new SDK, but the app crashes because the boost system library is missing. The `boost_locale` library apparently depends on `boost_system`, so `boost_system` has to be re-added to the SDK (or the FO build scripts or whereever `boost_system` has been removed...).

o01eg:
@Vezzra OK, I've restored this component.

Vezzra (approved):
Yep, now it works - with this PR applied, I can now build and run FO with the new SDK on macOS.

o01eg:
Nice, it doesn't affect MSVC on windows so I think I can merge it.

o01eg:
It's strange, but github doesn't allow to "Create a merge commit" and says "Not enabled for this branch".

geoffthemedio:
> It's strange, but github doesn't allow to "Create a merge commit" and says "Not enabled for this branch".

Probably because I just enabled a setting about that, at @Cjkjvfnby's suggestion. I thought this would just prevent merging branches that already had merge commits in them... But apparently it prevents merging with a merge commit. In that case, I'll turn it off...

In particular, it was "Require linear history - Prevent merge commits from being pushed to matching branches."

Cjkjvfnby:
Sorry for the necroposting, I just saw `Require linear history - Prevent merge commits from being pushed to matching branches` once again and got that this is the perfect match to our workflow. 

Basically, this check forbids you not only to merge PR with merge in the history but also to merge it via creating a merge commit. 

When you want to merge PR, you could use one of 3 options how to do it. 
изображение

In the repository settings option to merge with merge commit could be disabled, so another option will be shown by default.
изображение


Squash and merge and rebase 100% match our recommended workflow.   

Personally, I'd prefer to keep only squash and merge, in that case, we won't be caring if the user has merges in his branch or not.

geoffthemedio:
> Personally, I'd prefer to keep only squash and merge,

That would force merges to squash all commits into a single commit. That's not always desirable, as I / we prefer to keep logically separate commits separate, particularly when the branch being merged is clean (ie. ideally with every commit working and not mixing unrelated changes into a single commit more than reasonable / necessary).

> In the repository settings option to merge with merge commit could be disabled

I don't see any such option.

Cjkjvfnby:
> That would force merges to squash all commits into a single commit. That's not always desirable, as I / we prefer to keep logically separate commits separate, particularly when the branch being merged is clean (ie. ideally with every commit working and not mixing unrelated changes into a single commit more than reasonable / necessary).


If you could merge commit independently, you could create PRs for them.  If they are dependent a squash is ok.  I am a big fan of iterative development: fast feedback, and easy review.  But it might be pretty hard when testing is mostly manual.


> I don't see any such option.

These are branch protection rules, for the branch you target.  You will see master rules in the place of the red shape.
изображение

PR #2783, opened 2020-02-28
Remove seemingly unnecessary use of deprecated std::unary_function.

### LINKED ISSUE #2782: GG / ListBox.h uses deprecated std::unary_function
https://github.com/freeorion/freeorion/blob/master/GG/GG/ListBox.h#L217

That line uses `std::unary_function` which was [deprecated in C++11](https://en.cppreference.com/w/cpp/utility/functional/unary_function)

It should probably be replaced.

geoffthemedio:
Seems to resolve https://github.com/freeorion/freeorion/issues/2782

### DIFF
--- a/GG/GG/ListBox.h
+++ b/GG/GG/ListBox.h
@@ -214,7 +214,8 @@ public:
         bool operator()(const iterator& lhs, const iterator& rhs) const;
     };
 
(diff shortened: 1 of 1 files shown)


adrianbroher:
LGTM.

PR #2789, opened 2020-03-09
Updating MSVC 2017 Project Configurations

geoffthemedio:
...

### DIFF
--- a/msvc2017/Common/Common.vcxproj
+++ b/msvc2017/Common/Common.vcxproj
@@ -5,13 +5,17 @@
       <Configuration>Debug</Configuration>
       <Platform>Win32</Platform>
     </ProjectConfiguration>
-    <ProjectConfiguration Include="Release-XP|Win32">
-      <Configuration>Release-XP</Configuration>
-      <Platform>Win32</Platform>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
     </ProjectConfiguration>
     <ProjectConfiguration Include="Release|Win32">
       <Configuration>Release</Configuration>
       <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
     </ProjectConfiguration>
   </ItemGroup>
   <PropertyGroup Label="Globals">
@@ -27,6 +31,12 @@
     <CharacterSet>Unicode</CharacterSet>
     <PlatformToolset>v141</PlatformToolset>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
+    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <UseDebugLibraries>true</UseDebugLibraries>
+    <CharacterSet>Unicode</CharacterSet>
+    <PlatformToolset>v141</PlatformToolset>
+  </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>StaticLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
@@ -34,12 +44,12 @@
     <CharacterSet>Unicode</CharacterSet>
     <PlatformToolset>v141</PlatformToolset>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release-XP|Win32'" Label="Configuration">
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>StaticLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
     <WholeProgramOptimization>true</WholeProgramOptimization>
     <CharacterSet>Unicode</CharacterSet>
-    <PlatformToolset>v140_xp</PlatformToolset>
+    <PlatformToolset>v141</PlatformToolset>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <ImportGroup Label="ExtensionSettings">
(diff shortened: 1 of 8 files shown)


geoffthemedio:
@Vezzra any objections?

msvc2017/Common/Common.vcxproj
```diff
@@ -27,19 +31,25 @@
     <CharacterSet>Unicode</CharacterSet>
     <PlatformToolset>v141</PlatformToolset>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
+    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <UseDebugLibraries>true</UseDebugLibraries>
+    <CharacterSet>Unicode</CharacterSet>
+    <PlatformToolset>v141</PlatformToolset>
```
adrianbroher, line 38: Is this even the correct platform toolset?

geoffthemedio: Not sure what you mean... v141 is the MSVC 2017 compiler toolset. These are MSVC 2017 project files.

adrianbroher: Isn't there a distinction between a 64 bit toolset and a 32 bit toolset?

geoffthemedio: Toolset (compiler version) and platform (x64 or Win32) are independent settings in the project configuration interface.

msvc2017/Common/Common.vcxproj
```diff
@@ -5,20 +5,24 @@
...
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
     </ProjectConfiguration>
     <ProjectConfiguration Include="Release|Win32">
       <Configuration>Release</Configuration>
       <Platform>Win32</Platform>
     </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
```
geoffthemedio, line 18: platform is specified here

msvc2017/Common/Common.vcxproj
```diff
@@ -5,20 +5,24 @@
...
+    </ProjectConfiguration>
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectGuid>{BEDF460A-EAE9-4E20-AFB2-2C8434051150}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>Common</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.15063.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>
```
Vezzra, line 25: @geoffthemedio, does this mean we're raising the minimum required Windows version to 8.1? Not just Win7?

Vezzra:
> @Vezzra any objections?

No. Personally I'd raise the minimum required Windows version to Win7, because I think there are still many still using Win7 out there, but if you prefer to go straight to Win8, I don't object.

I'll postpone the next test builds until this has been merged.

adrianbroher:
@Vezzra Windows 7 is completely EOL since two month, Windows 8.1 has at least extended support until 2023-01-10.  I don't see a reason to support OS, which have been abandoned.

msvc2017/Common/Common.vcxproj
geoffthemedio: Because I didn't have an option for Win7.... just various `10.0.X` and `8.1`. I'm not actually sure what this setting does though, since the old project has this set to `10.0.15063.0` but it still runs on Win8.1 for me. But I seemed to need to change it to get it to run, although I may have confused the problem with the GLEW library in the SDK issue.

Vezzra:
@adrianbroher, I'm aware of that, however, that obviously hasn't been a criteria until now - we've been supporting FO on XP up until now, which is way more outdated than Win7. The reason being, IIRC, because XP has still been in use by so many people, despite having reached EOL long ago.

That most likely does not apply to XP anymore (I mean, that thing is *ancient* by now), but Win7 probably still has a significant user base. *If* our criteria is to support FO on platforms that are still in sufficiently widespread use, dropping support for Win7 is not a good idea.

If however our criteria is to only support platforms that have not reached EOL, then continuing support only for Win8 and later makes sense.

I'm fine with both approaches.

msvc2017/Common/Common.vcxproj
Vezzra: Ok, I see. Well, in that case we don't have much of a choice... I just wonder, because we've been able to support XP up until now, and suddenly we can't support Win7 anymore? Seems weird...

geoffthemedio:
Also, the change in question is a change to the x64 configuration. If the Win32 configuation worked with Win7 before, then that shouldn't have changed with these changes. I also assumed Vezzra would continue making 32-bit builds, and that these would not be affected. Only if someone builds the x84 vesion for themselves would they potentially see the difference.

Vezzra:
Oh, ok, didn't notice that. However, that raises the question, where can I see/how can I determine which Windows versions are supported by what I build with the MSVC solution...? Is it the toolkit version?

Vezzra:
I'm asking because we need to update those parts of our wiki and other docs where the minimum Windows version required to run FO is stated. XP is no longer correct, so that needs to be updated accordingly...

Morlic-fo:
Windows SDK 8.1 is backward compatible to those platforms. In addition, you should define the appropriate value for _WIN32_WINNT in targetver.h. For Windows 7, that's 0x0601. See Modifying WINVER and _WIN32_WINNT.
```

geoffthemedio:
I'm not sure if there's an easy way to test what Windows versions a program supports without testing on them. It may depend on the setting when building and what Windows SDK you have installed to build against, though.

That said, [this post](https://social.msdn.microsoft.com/Forums/vstudio/en-US/08aaf316-f901-4385-82b2-537c2f6b2cde/change-target-platform-version-to-windows-7?forum=visualstudiogeneral) has a comment that suggests the 8.1 target should work on windows 7, possibly with some #define modifications as described in the [linked msdn page](https://docs.microsoft.com/en-us/previous-versions/8x480de8(v=vs.140)). Without those modifications, maybe it targets 8.1?

Also, I think I was wrong... the 8.1 setting isn't specific to the x64 configuration, but rather is set for each project independent of the x64 or Win32 config setting. So possibly right now the Win32 build also requires 8.1 minimum.

geoffthemedio:
I looked into the project files. I see that several (all I checked) projects have `_WIN32_WINNT=0x0600` in the preprocessor defines. That should be targetting Windows Vista, based on [this page](https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=vs-2017). So I think it should still work on Win7.

That appears to be set [here](https://github.com/freeorion/freeorion/blob/master/msvc2017/win32.dependency.props#L10). I might update it to `0x0601` actually (Win7), in case there's some advantage to the slightly newer Windows API.

PR #2790, opened 2020-03-12
Raise minimum Python requirements in build instructions to 3.5

adrianbroher:
This is a follow up to #2653 , updating some bits of build documentation.

@o01eg was there also a change in the minimum required boost version thanks to the python migration?

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -24,7 +24,7 @@ FreeOrion depends on the following software to build:
   * A C++14 compliant compiler - Other Operating Systems
     * [GNU GCC] - 5.0 or later
     * [Clang] - 3.9 or later (4.0 or later on FreeBSD)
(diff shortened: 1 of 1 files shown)


o01eg:
@adrianbroher Boost 1.58 could be build with python3 support, but I don't know about Linux distributive with such version.

geoffthemedio:
Presumably all relevant distros should be building every version of Boost they include against Python 3 by now?

o01eg:
@geoffthemedio Yes, for debian-based it's distros started at Debian 9 Stretch and Ubuntu 18.04 Bionic.

adrianbroher:
> Presumably all relevant distros should be building every version of Boost they include against Python 3 by now?

Since ages.

PR #2697, opened 2020-01-07
Enable shared library build with with CMake on Windows platform

adrianbroher:
This PR is an updated version of #1416

PR #2791, opened 2020-03-12
Fix headless compilation

o01eg:
For server-side deploy freeorion should be able to build without graphic libraries. Common and server code depends on `GG::Clr` and `GG_ENUM`.

Before it was enough to use include files from GiGi without it compilation, but after https://github.com/freeorion/freeorion/pull/2751 it stared to require GiGi project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,6 +118,24 @@ jobs:
         - cmake -DBUILD_TESTING=ON ..
         - cmake --build . -- -j 2
 
+    - name: Build headless FreeOrion on Ubunutu 18.04 (Bionic)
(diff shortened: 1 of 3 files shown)


o01eg:
@adrianbroher I cannot choose you in reviewers. Could you check this PR, it's going to conflict with your https://github.com/freeorion/freeorion/pull/2697

adrianbroher:
@o01eg To be honest I don't care about this PR or the idea behind it.

Vezzra:
Sorry for getting back to this too late, but I wouldn't have been able to offer much useful feedback anyway...

PR #2803, opened 2020-03-26
Don't expect CMake policy CMP0044 OLD behaviour anymore

adrianbroher:
The build system is already [CMP0044](https://cmake.org/cmake/help/latest/policy/CMP0044.html) compliant and starting with CMake 3.11 the OLD policy value introduces a deprecation warning during build configuration time.

### DIFF
--- a/GG/CMakeLists.txt
+++ b/GG/CMakeLists.txt
@@ -25,10 +25,6 @@
 
 cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
-if(POLICY CMP0044)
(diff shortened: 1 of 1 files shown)


Vezzra:
Um, the "component" tag should be "build system", not "infrastructure"? If I'm mistaken, feel free to change back the tagging.

geoffthemedio:
You can just fix it if you thunk it's wrong... no need to comment also...

Vezzra:
I'm a bit reluctant to make such corrections/changes without any comment/explanation. I mean, I could be the one who's wrong... 😉

PR #2804, opened 2020-03-27
Improve invasion description to account for manual and automatic ship…

Oberlus:
… selection

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -9362,6 +9362,8 @@ PC_TROOPS_DESC
 
 To successfully invade a planet, the total number of [[metertype METER_TROOPS]] from all invading ships must exceed the total defending (ties are awarded to the defender).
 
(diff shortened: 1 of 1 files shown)


default/stringtables/en.txt
```diff
@@ -9362,6 +9362,8 @@ PC_TROOPS_DESC
 
 To successfully invade a planet, the total number of [[metertype METER_TROOPS]] from all invading ships must exceed the total defending (ties are awarded to the defender).
 
+When invading a planet, the troop ships selected in the fleet window will be used for the invasion. If no troop ships were selected, enough troop ships will be automatically selected.
```
geoffthemedio: remove "enough" as there might not be that many available

geoffthemedio: were -> are

geoffthemedio: maybe add "from the available ships and based on the number of defending troops".

Oberlus: Fixed as per your suggestions.

PR #2805, opened 2020-03-27
Remove GLU dependency

adrianbroher:
This PR replaces all existing `glu*` function calls with functional equivalents.

This is a small step towards cutting ties with the fixed function pipeline and allowing the usage of EGL, which in turn would allow the usage of Wayland, building on Android, and the new Linux OpenGL ABI (GLVND).

### DIFF
--- a/GG/src/SDL/SDLGUI.cpp
+++ b/GG/src/SDL/SDLGUI.cpp
@@ -363,8 +363,6 @@ void SDLGUI::SDLInit()
 
 void SDLGUI::GLInit()
 {
-    double ratio = Value(m_app_width) * 1.0 / Value(m_app_height);
-
     glEnable(GL_LIGHTING);
     glEnable(GL_LIGHT0);
     glEnable(GL_DEPTH_TEST);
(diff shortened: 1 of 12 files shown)


GG/src/SDL/SDLGUI.cpp
```diff
@@ -375,7 +373,22 @@ void SDLGUI::GLInit()
     glViewport(0, 0, Value(m_app_width), Value(m_app_height));
     glMatrixMode(GL_PROJECTION);
     glLoadIdentity();
-    gluPerspective(50.0, ratio, 1.0, 10.0);
+
+    // calculate perspective with vertical FOV of 50°. 1:1 application
+    // window ratio, near plane of 1.0 and far plane of 10.0
+    float ratio = Value(m_app_width) * 1.0f / Value(m_app_height);
+    float radians = 50.0f * M_PI / 90.0f;
+    float cotangent = std::cos(radians) / std::cos(radians);
```
geoffthemedio: second should be sin probly?

adrianbroher: Yes, it should be. I've updated the PR.

GG/test/acceptance/runner/Application.cpp
```diff
@@ -181,19 +181,39 @@ void MinimalGGApp::Render() {
...
+    projection[1][1] = cotangent;
+    projection[2][2] = -11.0f / 9.0f;
+    projection[2][3] = -1.0f;
+    projection[3][2] = -20.0f / 9.0f;
+    projection[3][3] = 0.0f;
+
+    glMultMatrixf(&projection[0][0]);
+
+    GLfloat view[4][4] = { 0.0 };
+
```
geoffthemedio: seemingly unnecessary blank line

GG/test/acceptance/runner/Application.cpp
```diff
@@ -181,19 +181,39 @@ void MinimalGGApp::Render() {
...
+    // calculate perspective with vertical FOV of 50°. 1:1 application
+    // window ratio, near plane of 1.0 and far plane of 10.0
+    float ratio = Value(AppWidth() * 1.0) / Value(AppHeight());
+    float radians = 50.0f * M_PI / 90.0f;
+    float cotangent = std::cos(radians) / std::sin(radians);
+
+    float projection[4][4] = { 0.0f };
+    projection[0][0] = cotangent / ratio;
+    projection[1][1] = cotangent;
+    projection[2][2] = -11.0f / 9.0f;
```
geoffthemedio: Lot of unexplained "magic" numbers here...  a comment indicating where they came from might be good. Maybe also rephrase "calculate" to "set up" above, since not much calculation is actually happening...

geoffthemedio: That said, I don't consider this test code particularly important, so if you'd prefer to not bother commenting it more, it's fine with me.

GG/test/acceptance/runner/Application.cpp
adrianbroher: Removed line and added a comment for camera placement.

GG/test/acceptance/runner/Application.cpp
adrianbroher: Replaced the non-magic magic numbers with variables within the matrix set up.

PR #2665, opened 2019-12-05
[default-only] KISS hard targeting

agrrr3:
This implements KISS hard targeting - weapons can only shoot at a certain class of targets. Each target has an equal probability to be hit.

Mass drivers, Lasers, Plasmas, Death Rays, and Spinal antimatter cannon attack ships and planets (no fighters)
Flak cannon only attacks fighters
Interceptors attack enemy fighters only. Fighter tech increases capacity and launch rate of interceptor fighters instead of damage
Fighters attack ships and fighters.
Bombers attack ships only.

Also balancing for the fighters. Making launch bay and hangars cheaper and slightly less effective (but more cost efficient). Also the structure is now more uniform so it is easier to compare
10 PP: Launch bay  (was 20PP)
25 PP: 1 Heavy Bomber, 12 damage (was deactivated)
20 PP: 2 Bombers, 6 damage, get a +3 damage bonus per tech level. (was 25PP, 5 damage, but had higher damage for death ray fighters)
15 PP: 3 Fighters, 4 damage, get a +2 damage bonus per tech level. Ignoring launching and targeting Fighters have a comparable power level to Bombers. (was  20PP, 3 damage, but had higher damage for death ray fighters)
10PP: 3 Interceptors only kill fighters. Can launch 50% faster compared to Fighter. Each tech gives one extra Interceptor and launch capacity. (Was 15 PP, 4 interceptors, but had damage increase against ships)

Interceptors are not affected by pilot trait. Fighters/Bombers/Heavy Bombers get the same damage malus/bonus per fighter: Bad/Avrg/Good/Great/Ultimate -1/+0/+1/+2/+3

This means Fighters Hangars benefit most from pilot trait as those launch three fighters (and Heavy Bombers are the least affected).

As the base damage is "the same", the types of fighters are differentiated by their launch patterns, concentration, targeting, and pilot traits.

Flak vs Interceptors is a great low-tech option, an ok mid-tech good-pilot option, and late game not really interesting. Fighters need more external slots but can do real damage. Interceptors are probably better if the enemy employs Fighters - they act as chaff, but they need more research and need extra internal slots.

[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11352&sid=6e51eec70784a37bed608c2bab8aad2f)

There are no client/server changes. So using the default directory of this branch with your freeorion 0.4.9 would work but is a bit outdated. To get the release version use. #2756 which  applies this code to the 0.4.9 stream

PR #2810, opened 2020-03-28
Make AI aware of allowed combat targets

### LINKED ISSUE #2786: AI only make interceptor
I mean they don't make fighter or bomber, and don't make flak.
I don't know whether it's a bug.
I thought game was designed that way.
Maybe possible that, it's because I always play in some way that make AI only make interceptor?
Ophiuchus said that it's a bug so I opened this issue.

And I played many different version, all the same, my latest version: FreeOrion_2020-03-02.89280c7_Test_Win32_Setup

Morlic-fo:
This PR allows the AI to understand combat targets better. More work is required to make full use of it but some improvements are:

* The AI should understand now that Flaks and Interceptors will not deal damage to ships
* The AI should understand now that Decoy spamming is no longer effective
* The AI should have a more accurate estimation of its strength against planets

Adapted to changes in #2665. Should resolve #2786.

Known limitations: The AI does not build Flak or Interceptors yet to specifically counter enemy carriers.

### DIFF
--- a/default/python/AI/AIDependencies.py
+++ b/default/python/AI/AIDependencies.py
@@ -758,6 +758,40 @@ SYSTEM_SHIP_FACILITIES = frozenset((
 # <editor-fold desc="Shipdesign/-parts">
 PART_KRILL_SPAWNER = "SP_KRILL_SPAWNER"
 
+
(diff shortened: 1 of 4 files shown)


geoffthemedio:
Still work in progress...?

PR #2818, opened 2020-04-01
Fix behaviour of cookies storage with new config policy

o01eg:
Make actual values and default value to be different so they will be stored in the config file.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -687,7 +687,8 @@ void HumanClientApp::MultiPlayerGame() {
             if (!GetOptionsDB().OptionExists(cookie_option + ".cookie"))
                 GetOptionsDB().Add<std::string>(cookie_option + ".cookie", "OPTIONS_DB_SERVER_COOKIE", "");
             if (!GetOptionsDB().OptionExists(cookie_option + ".address"))
-                GetOptionsDB().Add<std::string>(cookie_option + ".address", "OPTIONS_DB_SERVER_COOKIE", server_dest);
+                GetOptionsDB().Add<std::string>(cookie_option + ".address", "OPTIONS_DB_SERVER_COOKIE", "");
+            GetOptionsDB().Set(cookie_option + ".address", server_dest);
             std::string cookie_str = GetOptionsDB().Get<std::string>(cookie_option + ".cookie");
             boost::uuids::string_generator gen;
             cookie = gen(cookie_str);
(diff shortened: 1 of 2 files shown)


client/human/HumanClientApp.cpp
```diff
@@ -687,7 +687,8 @@ void HumanClientApp::MultiPlayerGame() {
             if (!GetOptionsDB().OptionExists(cookie_option + ".cookie"))
                 GetOptionsDB().Add<std::string>(cookie_option + ".cookie", "OPTIONS_DB_SERVER_COOKIE", "");
             if (!GetOptionsDB().OptionExists(cookie_option + ".address"))
-                GetOptionsDB().Add<std::string>(cookie_option + ".address", "OPTIONS_DB_SERVER_COOKIE", server_dest);
+                GetOptionsDB().Add<std::string>(cookie_option + ".address", "OPTIONS_DB_SERVER_COOKIE", "");
```
geoffthemedio, line 691: why just `""` here but `boost::uuids::to_string(boost::uuids::nil_uuid())` on the client?

client/human/HumanClientFSM.cpp
```diff
@@ -351,7 +352,8 @@ boost::statechart::result WaitingForMPJoinAck::react(const JoinGame& msg) {
             try {
                 std::string cookie_option = HumanClientApp::EncodeServerAddressOption(Client().Networking().Destination());
                 GetOptionsDB().Remove(cookie_option + ".cookie");
-                GetOptionsDB().Add(cookie_option + ".cookie", "OPTIONS_DB_SERVER_COOKIE", boost::uuids::to_string(cookie));
+                GetOptionsDB().Add(cookie_option + ".cookie", "OPTIONS_DB_SERVER_COOKIE", boost::uuids::to_string(boost::uuids::nil_uuid()));
```
geoffthemedio: not relevant to these changes, but why `Remove` then `Add` rather than the `OptionExists` test used on the server?

Might also be clearer to `<std::string>` parameterize `Add` like on the server as well.

o01eg: As I remember `Remove`+`Add` was done to reset default value as well. Looks like it isn't issue anymore.

client/human/HumanClientApp.cpp
o01eg: `server_dest` isn't uuid value but server host address. I suppose empty for addresses and `nil_uuid` for uuids are reasonable default values.

client/human/HumanClientFSM.cpp
o01eg: Fixed

geoffthemedio:
if it works for you, fine with me...

PR #2828, opened 2020-04-05
Move SDLGUI into FreeOrion client

adrianbroher:
This PR build the GG::SDLGUI as part of the FreeOrion client.

This simplifies the packaging as libGiGiSDL.so does not exist as a separate file anymore, it reduces the overall file size of the release and makes it easier in the long run to move the game loop control into the game client, to reduce and avoid jumping back and forward  between GG and the client application control flow for several actions like application resizing, application closing or network handling.


Vezzra:
I assume that means `GiGiSDL.dll` also needs to be removed from the "Bibaries Only" weekly test build archive provided for Windows, once this gets merged.

Note to self: don't forget to do that. 😎

adrianbroher:
@Vezzra how do you package the "binaries only" release anyway? Manually or do you use a script for that?

Vezzra:
@adrianbroher: I've written a small Python script for that.

PR #2832, opened 2020-04-07
Replace XMLElement::NoSuchChild with std::out_of_range

adrianbroher:
Small PR to reduce dependency between common component and GG.

### DIFF
--- a/util/XMLDoc.cpp
+++ b/util/XMLDoc.cpp
@@ -103,7 +103,7 @@ const XMLElement& XMLElement::Child(const std::string& tag) const {
         [&tag] (const XMLElement& e) { return e.m_tag == tag; });
 
     if (match == children.end())
-        throw NoSuchChild("XMLElement::Child(): The XMLElement \"" + Tag() + "\" contains no child \"" + tag + "\".");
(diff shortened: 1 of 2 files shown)


util/XMLDoc.cpp
```diff
@@ -103,7 +103,7 @@ const XMLElement& XMLElement::Child(const std::string& tag) const {
         [&tag] (const XMLElement& e) { return e.m_tag == tag; });
 
     if (match == children.end())
-        throw NoSuchChild("XMLElement::Child(): The XMLElement \"" + Tag() + "\" contains no child \"" + tag + "\".");
+        throw std::out_of_range("XMLElement::Child(): The XMLElement \"" + Tag() + "\" contains no child \"" + tag + "\".");
```
geoffthemedio, line 106: Is `out_of_range` appropriate in this context? The children aren't stored in any particular order, I think... Maybe `invalid_argument` would fit better? (I'm not sure... feel free to disregard...)

adrianbroher: I'm following the general semantics of `std::map::at` here.  `XMLElement::Child` returns an element accoring to a key, similar to `std::map::at`. Because `XMLElement::Child` has no other way to expose a non found child (e.g. end() iterator) due to the return type or circumventing this state by generating a new child (similar to `std::map::operator[]`), `std::out_of_range` is thrown.

PR #2834, opened 2020-04-08
Call application method when winner is determined

### LINKED ISSUE #2401: Stop the game after having winners
### Environment
* **FreeOrion Version**: 0.4.8, master
* **Fetched as** 
  * Binary release
  * Weekly development build
  * Compiled from source

### Description
Currently the game continues after some players wins the game.

### Expected Result
Server notifies all player about winners and their victory reason.
Server shutdowns or went back to lobby depending on mode.

### Steps to reproduce
* Win the game.
* Make next turns.

o01eg:
Fixes #2401.

Disables turn timeout when the game get a winner.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -631,6 +631,10 @@ bool Empire::Won() const {
 
 void Empire::Win(const std::string& reason) {
     if (m_victories.insert(reason).second) {
+        auto app = IApp::GetApp();
+        if (app != nullptr) {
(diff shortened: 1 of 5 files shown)


geoffthemedio:
I don't like implementing this as changing the option setting. Better would be to explicitly state in the option name that it will only auto-cycle turns if no player has won, and test for that where the cycling happens, but leave the option setting unchanged.

PR #2835, opened 2020-04-08
Remove GG:::ListBox::Row constructor default parameters

adrianbroher:
Most subclasses only use a specific combination of settings, which can
be set via setters.  Also the parameterized constructor client to set
the drag and drop datatype always to `""` (not draggable), even if the
row is not draggable.

### DIFF
--- a/GG/GG/ListBox.h
+++ b/GG/GG/ListBox.h
@@ -134,7 +134,7 @@ public:
 
         /** \name Structors */ ///@{
         Row();
-        Row(X w, Y h, const std::string& drag_drop_data_type, Alignment align = ALIGN_VCENTER, unsigned int margin = 2);
+        Row(X w, Y h);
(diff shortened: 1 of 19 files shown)


geoffthemedio:
A few minor merge conflicts, relating to `,` at the end of a line that were removed when you also edited the line. Attempted to resolve here: https://github.com/freeorion/freeorion/tree/resolve-conflicts-simplify-gg-listbox

adrianbroher:
> Attempted to resolve here: https://github.com/freeorion/freeorion/tree/resolve-conflicts-simplify-gg-listbox

No idea what do want to show me here.  I will rebase this branch.

geoffthemedio:
The hope was that it would be slightly helpful to have my version of the conflict resolution to refer to. If not, then ignore it.

PR #2837, opened 2020-04-08
better words for various sorts of exiting the game

l29ah:


geoffthemedio:
I don't like either of these renamings.
"Disconnect" doesn't make sense if one isn't connected to a remove server.
"Surrender" sounds like something one would do in the game, rather than the act of leaving the game and giving up playing / competing.

swaqvalley:
I prefer "Concede" over "Surrender" as well.
For "Resign" I think a replacement that would be more accurate is "Quit to Menu" or "Exit to Menu".

geoffthemedio:
@l29ah Will you update this?

l29ah:
done

magnate:
So they are completely different concepts. One is quitting the app, the other is quitting the game. Exit to Menu is pretty universal for the former, nobody would worry that they're conceding the game. For the latter, Resign/Concede/Surrender all have roughly the same meaning and any would do. I personally think Surrender would be the clearest, since Resign and Concede have already been used. But no harm in sticking with Concede if that's what people want.

PR #2845, opened 2020-04-10
Add options to disable disclosure of galaxy seed

### LINKED ISSUE #2820: Add option to hide real galaxy in multiplayer game
### Environment
* **FreeOrion Version**: any
* **Operating System**: any

### Description
Currently player could get galaxy settings from the server and expose entire map using same seed, which make visibility of starts and planets useless.

Forum thread: https://freeorion.org/forum/viewtopic.php?p=99852#p99852

### Expected Result
With some options enables the server hides real value of seed from players, but still saves it so the game can be reproduced.

o01eg:
Fixes #2820

### DIFF
--- a/server/SaveLoad.cpp
+++ b/server/SaveLoad.cpp
@@ -106,7 +106,7 @@ std::map<int, SaveGameEmpireData> CompileSaveGameEmpireData() {
 int SaveGame(const std::string& filename, const ServerSaveGameData& server_save_game_data,
              const std::vector<PlayerSaveGameData>& player_save_game_data, const Universe& universe,
              const EmpireManager& empire_manager, const SpeciesManager& species_manager,
-             const CombatLogManager& combat_log_manager, const GalaxySetupData& galaxy_setup_data,
+             const CombatLogManager& combat_log_manager, GalaxySetupData& galaxy_setup_data,
              bool multiplayer)
 {
     SectionedScopedTimer timer("SaveGame");
@@ -172,6 +172,7 @@ int SaveGame(const std::string& filename, const ServerSaveGameData& server_save_
         pos_before_writing = ofs.tellp();
 
         bool save_completed_as_xml = false;
+        galaxy_setup_data.EncodingEmpire() = ALL_EMPIRES;
 
         if (!use_binary) {
             if (use_zlib_for_zml) {
--- a/server/SaveLoad.cpp
+++ b/server/SaveLoad.cpp
@@ -106,7 +106,7 @@ std::map<int, SaveGameEmpireData> CompileSaveGameEmpireData() {
 int SaveGame(const std::string& filename, const ServerSaveGameData& server_save_game_data,
              const std::vector<PlayerSaveGameData>& player_save_game_data, const Universe& universe,
              const EmpireManager& empire_manager, const SpeciesManager& species_manager,
-             const CombatLogManager& combat_log_manager, GalaxySetupData& galaxy_setup_data,
+             const CombatLogManager& combat_log_manager, GalaxySetupData galaxy_setup_data,
              bool multiplayer)
 {
     SectionedScopedTimer timer("SaveGame");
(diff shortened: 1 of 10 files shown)


util/MultiplayerCommon.h
```diff
@@ -51,6 +51,13 @@ struct FO_COMMON_API GalaxySetupData {
     /** \name Mutators */ //@{
     void                SetSeed(const std::string& seed);
     void                SetGameUID(const std::string& game_uid);
+
+    /** HACK! This must be set to the encoding empire's id when serializing a
+      * Universe, so that only the relevant parts of the Universe are
+      * serialized.  The use of this global variable is done just so I don't
+      * have to rewrite any custom boost::serialization classes that implement
+      * empire-dependent visibility. */
+    int& EncodingEmpire();
```
geoffthemedio: This is a struct with public members, so there's no point to this getter... Just access `.m_encoding_empire` directly.

server/SaveLoad.cpp
```diff
@@ -172,6 +172,7 @@ int SaveGame(const std::string& filename, const ServerSaveGameData& server_save_
         pos_before_writing = ofs.tellp();
 
         bool save_completed_as_xml = false;
+        galaxy_setup_data.EncodingEmpire() = ALL_EMPIRES;
```
geoffthemedio: Rather than passing the `GalaxySetupData` by non-const reference and changing the interfaces everywhere, I'd pass it by value and modify the local, or by const reference and make a local copy.

o01eg: Fixed

util/MultiplayerCommon.h
o01eg: Fixed.

util/MultiplayerCommon.h
```diff
@@ -75,6 +68,12 @@ struct FO_COMMON_API GalaxySetupData {
     std::map<std::string, std::string>
                         m_game_rules;
     std::string         m_game_uid;
+
+    /** HACK! This must be set to the encoding empire's id when serializing a
+      * Universe, so that only the relevant parts of the Universe are
+      * serialized.  The use of this global variable is done just so I don't
```
geoffthemedio: update this comment... unlike the main encoding empire global, this is not a global...

o01eg: Fixed.

util/SerializeMultiplayerCommon.cpp
```diff
@@ -15,8 +16,14 @@
 template <class Archive>
 void GalaxySetupData::serialize(Archive& ar, const unsigned int version)
 {
-    ar  & BOOST_SERIALIZATION_NVP(m_seed)
-        & BOOST_SERIALIZATION_NVP(m_size)
+    if (Archive::is_saving::value && m_encoding_empire != ALL_EMPIRES && (!GetOptionsDB().Get<bool>("network.server.publish-seed"))) {
+        std::string dummy = "hidden";
```
geoffthemedio: Is this "hidden" text meant to be shown to the player on the galaxy info pedia page or multiplayer lobby? If so, there shouldn't be localized text in the source code...

o01eg: I'm not sure what to place here as any localization will happen on the server side with some default language instead of client side. Maybe send "-" or some sign to the client?

adrianbroher: Unless an empty string has already another special meaning use that and let the client handle the seed representation.

geoffthemedio: In the client UI, setting an empty seed sets it to generate a random seed. The universe generation also treats an empty seed as a sign to generate a random one server-side. But neither of those mean an empty seed can't be sent back to the client later...

o01eg: So I suppose empty seed is fine to send to the client.

PR #2848, opened 2020-04-11
Format en stringtable

adrianbroher:
This PR:

formats the english string table with the `st-tool format` command.  With that 

* Non-escaped whitespace was removed.
* Unneccessary quotes were removed.
* Actual section titles were marked with double hashes.
* Missing empty lines between some entries were added.

It also adds a formatting check for the english stringtable to the `Lint stringtables` CI job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,7 @@ jobs:
       python: 3.5
       script:
         - ./check/st-tool.py check --reference default/stringtables/en.txt default/stringtables/*.txt
+        - diff -u default/stringtables/en.txt <(check/st-tool.py format default/stringtables/en.txt) || { echo "String table is not properly formatted"; exit 1; }
 
     - name: Build C++ API documentation
       stage: build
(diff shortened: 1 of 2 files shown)


Cjkjvfnby:
@adrianbroher, I need an idea of what should be done with other languages. 

Problems: 
- The first comment section is removed (authors)
- Empty spaces after comments are removed and comments are joined with the wrong keys.

`ru.txt`
```
# Used as a placeholder for unexpanded content in the combat log window.
EMPIRE
Империя
```

`en.txt`
```
# Used as a placeholder for unexpanded content in the combat log window.
ELLIPSIS
...

EMPIRE
Empire
```

Sample diff:
https://github.com/freeorion/freeorion/compare/master...Cjkjvfnby:ru-format?expand=1

adrianbroher:
> The first comment section is removed (authors)

We have the credits.xml file, no need to duplicate this information.

> Empty spaces after comments are removed and comments are joined with the wrong keys.

Use the `sync` verb first to place all keys and comments including untranslated keys into the russian translation file at the equivalent location to the english to allow proper diffs.  The tool cannot read intent, it applies structure on given data.

PR #2859, opened 2020-04-15
Make COMBAT_TARGETS_PLANET_WITH_DEFENSE macro consider infrastructure

### LINKED ISSUE #2488: Planet invasion only possible on odd turns if planet does not have defence regeneration

Oberlus:
Fixes #2488 as suggested by @agrrr3 in [this post](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11600&p=100633#p100612).

Tested with mass drivers against an Acirema HW, Furthest HW with Moderate Tech Special, and a Gysache AI HW without Planetary Defense Regeneration.

### DIFF
--- a/default/scripting/ship_parts/targeting.macros
+++ b/default/scripting/ship_parts/targeting.macros
@@ -22,11 +22,12 @@ COMBAT_TARGETS_NOT_DESTROYED_SHIP
 '''
 
 COMBAT_TARGETS_PLANET_WITH_DEFENSE
-'''         And [
+'''        And [
             Planet
             Or [
-               Not Shield high = 0
-               Not Defense high = 0
+                Not Shield high = 0
+                Not Defense high = 0
+                Not Construction high = 0
             ]
         ]
 '''


agrrr3:
@geoffthemedio As Oberlus playtested this and it is more consistent with the c++ backend, please merge (i think else the slow server game gets delayed by two weeks)

default/scripting/ship_parts/targeting.macros
```diff
@@ -22,11 +22,12 @@ COMBAT_TARGETS_NOT_DESTROYED_SHIP
...
 COMBAT_TARGETS_PLANET_WITH_DEFENSE
-'''         And [
+'''        And [
             Planet
             Or [
-               Not Shield high = 0
-               Not Defense high = 0
+                Not Shield high = 0
+                Not Defense high = 0
+                Not Construction high = 0
```
o01eg, line 30: Does construction somehow affect it?

Oberlus: This way it mimics the way it is done in the backend. So the answer must be yes.

agrrr3: Yes in the past it was too messy to combine regeneration for defense and shield for triggering combat with the usual balance and intended feature (zero defense and zero shields should be possible on a colony). So we reused the unused construction for triggering combat by ensuring minimum construction one (on colonies).

PR #2860, opened 2020-04-15
Remove python2 compatilility

Cjkjvfnby:
Remove compatibility with python2.

This PR can be checked commit by commit.

PR #2861, opened 2020-04-15
Add unit test for AI savegame compression

Morlic-fo:
The compression&decompression was previously not covered by unit tests. 

Catches an error introduced in #2860.


Morlic-fo:
Superseded by #2865

PR #2864, opened 2020-04-16
Don't install unneeded for headless build files

o01eg:


geoffthemedio:
@adrianbroher I know you don't care about headless builds, but this seemingly touches the general-case script, so you might want to have a look at it...

adrianbroher:
It should work, but I don't see the point the current implementation of `BUILD_HEADLESS`.  That's what CPack components are intended for.

PR #2865, opened 2020-04-16
[AI] Remove old save compatibility

Cjkjvfnby:
- Remove constructor argument to compress file, it was never used.  
- Simplify save decompression: drop support for uncompressed and old saves (partially compressed)
- Add tests for compression-decompression

### DIFF
--- a/default/python/AI/savegame_codec/_decoder.py
+++ b/default/python/AI/savegame_codec/_decoder.py
@@ -13,8 +13,10 @@ to provide a __setstate__ method to verify and possibly sanitize the content of 
 """
 import binascii
 import json
+from typing import Union
 
 import EnumsAI
+from AIstate import AIstate
 from freeorion_tools import profile
 
 from ._definitions import (ENUM_PREFIX, FALSE, FLOAT_PREFIX, INT_PREFIX, InvalidSaveGameException, NONE, PLACEHOLDER,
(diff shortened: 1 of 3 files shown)


o01eg:
Does it break compatibility with saves made before this PR, or with some very old saves?

default/python/AI/savegame_codec/_decoder.py
```diff
@@ -11,81 +11,41 @@
...
-        s = s.decode(encoding, errors)
-    return s
+    pass
 
 
 @profile
 def load_savegame_string(string: str):
     """
     :rtype: AIstate
+    :raises: SaveUncompressException, InvalidSaveGameException
```
Morlic-fo: SaveDecompressException

default/python/AI/savegame_codec/_decoder.py
```diff
@@ -11,81 +11,41 @@
...
     """
     import base64
     import zlib
 
-    new_string = string
     try:
         new_string = base64.b64decode(string.encode("utf-8"))
```
Morlic-fo: the encode seems unnecessary here. ```b64decode``` allows to pass a ```str``` and will  use ```encode('ascii')``` internally (which the string  must comply with if it is indeed base64 encoded).

default/python/AI/savegame_codec/_decoder.py
```diff
@@ -11,81 +11,41 @@
...
-        # probably an uncompressed string
-        debug("Will try decoding savestate string without base64 or zlib compression.")
-
-    # We need to check uncompressed string, since zip archive of the empty string is not the empty string.
-    if not string:
-        raise Exception("Save game string is empty")
-    return decode(string)
+        new_string = zlib.decompress(new_string)
+    except zlib.error as e:
+        raise SaveDecompressException("Fail to ungzip savestate %s" % e) from e
```
Morlic-fo: typo: unzip

Morlic-fo (commented):
besides typos, looks good.

default/python/AI/savegame_codec/_decoder.py
o01eg: I want to mention we can use `typing` syntax instead of comments because minimal supported python version allows it.

default/python/AI/savegame_codec/_decoder.py
Cjkjvfnby: zlib docs:
The functions in this module allow compression and decompression using the
zlib library, which is based on GNU zip.

So it should be `gunzip` like corresponding tool.   But I think I will use decompress to reduce confusion.

Cjkjvfnby:
> Does it break compatibility with saves made before this PR, or with some very old saves?

Only with old saves.  Saves prior the PR work fine.

Morlic-fo (approved):

o01eg (approved):

Issue #2862, opened 2020-04-16
New colonies should be happier than invaded planets


swaqvalley:
PR #2745 causes new colonies to have no industry or research when created because they have less than 5 happiness. This new mechanic handicaps new colonies just as much (or more) than invaded planets.

Oberlus suggests starting a colony with more happiness when the habitability is better for the species. Forum discussion thread:
https://www.freeorion.org/forum/viewtopic.php?f=28&t=11604

Vezzra:
Sounds like a quite serious issue for gameplay. Should be addressed before a new release.

Oberlus:
This can be closed.
Edit: it's been tested and works fine.

PR #2866, opened 2020-04-17
Use new LastTurnColonized property for #2862 fix

swaqvalley:
Adjusts the fix for Issue #2862 to use the new LastTurnColonized property. My testing shows the expected behavior is identical.

### DIFF
--- a/default/scripting/species/common/happiness.macros
+++ b/default/scripting/species/common/happiness.macros
@@ -16,9 +16,9 @@ AVERAGE_HAPPINESS
             activation = And [
                 Planet
                 (Source.LastTurnConquered < CurrentTurn - 5)
-                Stockpile high = 0
+                (Source.LastTurnColonized = CurrentTurn)
             ]
-            effects = SetHappiness value = max(Value, 
+            effects = SetHappiness value = max(Value,
                 1*(Statistic If Condition = And [Target Planet environment = Poor]) +
                 2*(Statistic If Condition = And [Target Planet environment = Adequate]) +
                 4*(Statistic If Condition = And [Target Planet environment = Good]))


default/scripting/species/common/happiness.macros
```diff
@@ -16,9 +16,9 @@ AVERAGE_HAPPINESS
             activation = And [
                 Planet
                 (Source.LastTurnConquered < CurrentTurn - 5)
```
agrrr3, line 18: Should be: current turn not equal to LastTurnConquered I guess

swaqvalley: That would work too. Though with the new LastTurnColonized property we could probably just remove the check for LastTurnConquered entirely since I don't think they can happen on the same turn anyway.

agrrr3: Yes, you are right

agrrr3:
Parsing fails for me when using LastTurnColonized

geoffthemedio:
> Parsing fails for me when using LastTurnColonized

What FreeOrion git commit are you using? Support was added in https://github.com/freeorion/freeorion/commit/5976f772939329bb0361f298ac449204be90a46c#diff-601d5b80352a1ab4acdf05496215f901

agrrr3:
> > Parsing fails for me when using LastTurnColonized
> 
> What FreeOrion git commit are you using? Support was added in [5976f77#diff-601d5b80352a1ab4acdf05496215f901](https://github.com/freeorion/freeorion/commit/5976f772939329bb0361f298ac449204be90a46c#diff-601d5b80352a1ab4acdf05496215f901)

That was on top of 49d325931b9a5db58d44442aecbf0d7d6784dcc8 and I can also see the necessary parts in Tokens.h etc and also the implementation from this PR (e286b001c1b77e7659eb23fab404686419f409f6). 5976f772939329bb0361f298ac449204be90a46c is also there as expected. I started freeorion to test my stuff and it complained about wrong all species  and the parsing breaking at Source.LastTurnColonized (it wanted the closing parenthesis for the And condition). I changed that to Source.LastTurnConquered and everything was dandy.

geoffthemedio:
You rebuilt from source after updating?

agrrr3:
I did (or so I think). I jumped forward some revisions and rebuild and now it worked (so probably not rebuild successfully). So everything seems good.

PR #2867, opened 2020-04-17
Do not inherit python classes from object

Morlic-fo:
This is no longer needed in Python3


Cjkjvfnby:
Golegue just sended me a link about tool that can do simular:
https://github.com/asottile/pyupgrade/

Cjkjvfnby (approved):

PR #2869, opened 2020-04-18
Rename PartType class to ShipPart and move to separate TU files

adrianbroher:
As advertised in the PR title.

The naming is now more consistent and more unambigious.  I decided to go with `ShipPart` as FOCS knows already the corresponding token and the name matches with the already existing `ShipPartClass` enum.


geoffthemedio:
Why the separate file / TU?

adrianbroher:
It's to reduce the dependencies of unrelated changes between classes, otherwise if you change only one of the classes you may have recompile the code any class defined in the cpp and heder file, you have to load the include file to get the dependencies of both, even time the file is compiled.  When the classes are separated you only have to recompile what are really depends on this class.

It's a measurement to reduce compile time.  But to take advantage of this this has to be applied rigorusly accross the code and I have to start somewhere.

geoffthemedio:
But presumably having many more source files / translation units may also increase compile time overall though, particularly for FreeOrion code, as it has to resolve all the includes and re-instantiate templates for each (vs. unity builds)...?

Separating the headers to reduce header dependencies across source files is reasonable, though in the case of ship part and ship hull, is probably not of much benefit since the two are almost always used together, so almost anything that includes one will include the other and need recompiling regardless. I recently split the two of them out from ShipDesign.h because there are several places where only ShipDesign is needed. But I left them in a single .cpp because of the previous paragraph; in general other places nclude the headers take a lot longer to recompile than the classes the headers are about...

... and for ShipPart in particular, there is a substantial amount of other stuff included to be resolved and templates to process. UnlockableItem is much simpler so probably less affected? Not sure how best to evaluate such tradeoffs...

adrianbroher:
> But presumably having many more source files / translation units may also increase compile time overall though, particularly for FreeOrion code, as it has to resolve all the includes and re-instantiate templates for each (vs. unity builds)...?

I can only speak from my experience that:

1. You do way more often do incremental builds than compiling from scratch.
2. That slow compile times in other projects were most of the time significantly reduced when consequently splitting out class implementations into separated translation units.

I personally consider unity builds and precompiled headers crutches for a bad code layout.

geoffthemedio:
Does "compile time" include link time, and how much did those previous projects make use of templated boost and standard library stuff compared with FreeOrion?

Regardless, I suppose it's easy enough to #include things into one file later if desired.

util/SerializeEmpire.cpp
```diff
@@ -186,7 +186,7 @@ void Empire::serialize(Archive& ar, const unsigned int version)
             & BOOST_SERIALIZATION_NVP(m_research_progress)
             & BOOST_SERIALIZATION_NVP(m_production_queue)
             & BOOST_SERIALIZATION_NVP(m_available_building_types)
-            & BOOST_SERIALIZATION_NVP(m_available_part_types)
+            & BOOST_SERIALIZATION_NVP(m_available_ship_parts)
```
geoffthemedio: Breaks save loading: `LoadGame(...) failed!  Error: XML start/end tag mismatch - m_available_ship_parts`. Need to increment the serialization class version and have a case to handle the older field name using `make_nvp`.

adrianbroher:
> Does "compile time" include link time, and how much did those previous projects make use of templated boost and standard library stuff compared with FreeOrion?

Link time was never a problem for me, but I meant compile time only. The projects use std lib and a good amount of project specific written templating code.

util/SerializeEmpire.cpp
adrianbroher: It's probably better to use `make_nvp` and preserve the tag instead of changing the format.

geoffthemedio: Sounds reasonable for just renaming without changing the data being encoded.

PR #2870, opened 2020-04-18
Rename ItemSpec class to UnlockableItem and move to separate TU files

adrianbroher:
As advertised in the PR title.

The naming is now more consistent and more unambigious.  I decided to go with `UnlockableItem` as it matches with the already existing `UnlockableItemType` enum.


universe/UnlockableItem.h
```diff
@@ -0,0 +1,61 @@
...
+//! (UIT_SHIP_PART, "Death Ray")).
+struct FO_COMMON_API UnlockableItem {
+    UnlockableItem();
+    UnlockableItem(UnlockableItemType type_, const std::string& name_) :
+        type(type_),
+        name(name_)
+    {}
+
+    //! Returns a data file format representation of this object
+    auto Dump(unsigned short ntabs = 0) const -> std::string;
```
geoffthemedio, line 34: What's the purpose of using `auto` and `-> std::string` here?

adrianbroher: It's for consistent indentation.  `void` and `auto` have the same width and by using trailing return types there is no need for whitespace fuckery (or "grooming").

adrianbroher:
Also another reason which I forgot to mention: Seems like MSVC doesn't like forward declarations when exporting symbols for dll.  I had another variant of this commit in #2697 as trying to export the ItemSpec without a full declaration failed in the CI:

https://ci.appveyor.com/project/freeorion/freeorion/builds/30174742/job/35yoyf8rbu1citnk#L295

PR #2871, opened 2020-04-18
Simplify ShipDesignAI

Morlic-fo:
Removes some ancient workaround for missing interface functions by providing and using the relevant interface functions.

### DIFF
--- a/default/python/AI/ShipDesignAI.py
+++ b/default/python/AI/ShipDesignAI.py
@@ -55,7 +55,7 @@ import CombatRatingsAI
 import FleetUtilsAI
 from AIDependencies import INVALID_ID, Tags
 from aistate_interface import get_aistate
-from freeorion_tools import UserString, get_species_tag_grade, tech_is_complete
+from freeorion_tools import UserString, get_species_tag_grade, tech_is_complete, assertion_fails
 from turn_state import state
 
 # Define meta classes for the ship parts  TODO storing as set may not be needed anymore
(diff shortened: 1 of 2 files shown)


default/python/AI/ShipDesignAI.py
```diff
@@ -2307,26 +2093,23 @@ def _get_design_by_name(design_name, update_invalid=False, looking_for_new_desig
...
 
-def get_part_type(partname):
+@cache_for_session
+def get_part_type(part_name):
     """Return the partType object (fo.getPartType(partname)) of the given partname.
 
     As the function in late game may be called some thousand times, the results are cached.
 
-    :type partname: str
+    :type part_name: str
```
Cjkjvfnby: I think we don't need type strings anymore.   If you change it just use annotations.   Optional for this PR.

python/UniverseWrapper.cpp
```diff
@@ -604,6 +626,7 @@ namespace FreeOrionPython {
             .def("canMountInSlotType",          &PartType::CanMountInSlotType)
             .add_property("costTimeLocationInvariant",
                                                 &PartType::ProductionCostTimeLocationInvariant)
+            .def("productionLocation",          &PartTypeProductionLocation)
```
Cjkjvfnby: Any docstring?

Cjkjvfnby (commented):
It looks like a great simplification.  I am not sure that I understand correctly how old code worked. 
All my comments are optional.

PR #2875, opened 2020-04-19
Use typename keyword within template statements

adrianbroher:
Functional identical to the `class` keyword.  However using `typename` makes grepping the source code easier when searching for classes.


adrianbroher:
Is there anything wrong this PR (aside from the now fixed merge conflict)?

PR #2879, opened 2020-04-19
Let outposts have minimum construction one, in order to be attacked

agrrr3:
Let outposts have minimum construction one, in order to be attacked by enemies

This sets minimum construction value for empire outposts. For neutral outposts strangeness may occur.
But neutral outposts are also strange - not sure if that is actually a thing.

This is an alternative approach in order to solve #2488


geoffthemedio:
This will cause combat to occur every turn. Why not instead stop shield / defense growth if they are both 0 and there are armed aggressive enemy ships in the system?

Oberlus:
Tested:

- Native planets (as an example of planets without any defense or shields) get attacked every turn. They don't get attacked in more than one bout every combat. OK!
- Outposts does not shot infrastructure in the GUI but they indeed have 1 infrastructure and are targetted once per combat on every turn. OK!
- Empire colonies also get attacked every turn (as expected from previous fix). OK!

From above, issue #2488 is resolved and can be closed, and this PR should be merged before tomorrow's weekly test build (@geoffthemedio)

On a side note, the relatively great infrastructure of empire colonies make them able to attract and soak many small-damage shots even when their shields&defenses were zeroed at combat start. This is not caused by this fix but by the previous one (#2859), and it was happening also to Arc Disruptors before that PR, since the initial fix for issue #2488 sometime in 2019.
This is not necessarily bad, it just shifts a bit combat advantage to the defending side, more noticeably early game (when weapons do less damage). I would not bother addressing this in any way since conquering planets is still seen as OP when compared to colonization, specially early game, and it does not really hampers blitzkriegs. But if veteran player would complain from the unexpected change of behaviour (to which such complaints I would disagree), it could be addressed by lowering initial infrastructure (not for v0.4.10 release though).

Oberlus:
> This will cause combat to occur every turn. Why not instead stop shield / defense growth if they are both 0 and there are armed aggressive enemy ships in the system?

Well, that would be an option too. But the gain in terms of avoiding combat on every turn would be ephemeral since empires can get relatively soon "Defense Regeneration" that will trigger combat on every turn. I don't think this would be worth.

PR #2885, opened 2020-04-22
Exclude engines from colony base hull

Oberlus:
Fixes #2882 

Debate about whether that issue is really a bug to be fixed or a new feature is still open.

[Forum thread](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11625).

PR #2886, opened 2020-04-22
Make AI aware of excluded parts in SH_COLONY_BASE

Morlic-fo:
AI Adaptations for #2885.

Should be implemented cleaner eventually by exposing and using FOCS exclusions but for works for now.


adrianbroher:
> Should be implemented cleaner eventually by exposing and using FOCS exclusions but for works for now.

Wouldn't it be more useful to ask @Oberlus directly, so he can integrate this into the exclusions PR?

PR #2887, opened 2020-04-22
Update stub generator and stubs

Cjkjvfnby:
A lot of changes, but I don't sure that it should be reviewed.   I plan to do more improvements to the generator code later.  And also need to work with getting more info for the stub. 

I put stubs and generator to different commits.

This release includes: 
- migrate to python3 stub style (more compact)
- fix minor issues
- tons of refactoring

PS. I know rules about ere, but this PR will be squash and merged anyway. 

Maybe anyone can help me with the C++ side. 

1) I got a problem with properties. They don't have any info about the return type.   Can you do something about it?  Regular functions have an autogenerated signature in the docstring. Is it possible to produce signatures for property get functions too?

2) Is it possible to add any real docstring to the property?  

You can just put it in C++ code and I will get the next win build and try it.


adrianbroher:
Would be excellent if this can be put somewhere else in the tree and not into the default directory, especially if this is a utility script and not something required for running the game.

o01eg:
>     1. I got a problem with properties. They don't have any info about the return type.   Can you do something about it?  Regular functions have an autogenerated signature in the docstring. Is it possible to produce signatures for property get functions too?

Not sure what happened when property declares. It calls https://docs.python.org/3/c-api/type.html#c.PyTypeObject with getter, setter or null, null, and docstring.

>     2. Is it possible to add any real docstring to the property?

Non-static property declaration have `doc` parameter:
https://www.boost.org/doc/libs/1_66_0/libs/python/doc/html/reference/high_level_components.html#high_level_components.boost_python_class_hpp

Cjkjvfnby:
> Would be excellent if this can be put somewhere else in the tree and not into the default directory, especially if this is a utility script and not something required for running the game

Will see what can be done in the next PR.

Cjkjvfnby:
> Not sure what happened when property declares. It calls https://docs.python.org/3/c-api/type.html#c.PyTypeObject with getter, setter or null, null, and docstring.

On the hight abstract level, the setter is a function.  And functions can have autogenerated docs with signature. 
 Is it possible to do it for getter?  I can access getter via `fget` attribute but it never has a docstring or anything useful.

adrianbroher:
@Cjkjvfnby can you extract the correct return f the Empire.planetsWithWastedPP property?

This is one of the properties with an explicit signature.

o01eg:
> Is it possible to do it for getter? I can access getter via `fget` attribute but it never has a docstring or anything useful.

Code of getters is placed here https://github.com/boostorg/python/blob/master/include/boost/python/data_members.hpp#L139 Not sure if it possible to have docstring here.

Cjkjvfnby:
> @Cjkjvfnby can you extract the correct return f the Empire.planetsWithWastedPP property?
This is one of the properties with an explicit signature.

I cannot get property information from the empire class.. For stub, I need to pass the instance of the empire to get a property return type.  It is not a very convenient way.

o01eg:
> I cannot get property information from the empire class.. For stub, I need to pass the instance of the empire to get a property return type. It is not a very convenient way.

Maybe it is defined as attribute instead of property?

Vezzra:
If I understand correctly, the stub generator and stubs are helper tools to provide code completion for Python IDEs (like PyCharm). While this is certainly important and useful for AI coders, is it really necessary to consider an update to these tools release-blocking?

Cjkjvfnby:
> release-blocking

When I created PR there were only `next-release` and `Gateway`.  Making it release-blocking is a kind of side effect of the release procedure.

Vezzra:
@Cjkjvfnby, we've kind of re-designed how we use milestones and labels for the release process.

You do not need to assign a milestone anymore, that is only necessary for issues/PRs which need to be tagged/marked as either mandatory for an upcoming release (in which case you assign the "next release" or the "vX.Y.Z" milestone) or as being postponed after the next release (in which case you assign the "post release milestone).

The vast majority of issues/PRs usually is optional, and is indicated by not assigning a milestone, which IMO makes most sense for this PR too.

Cjkjvfnby:
@Vezzra , can we merge it now?

Vezzra (commented):
> can we merge it now?

No objections from my side. Can't offer a meaningful review of the proposed changes though.

PR #2890, opened 2020-04-24
Cheaper shields

Oberlus:
With the addition of fighters (that ignore shields), shields became considerably less useful, to the point of they being a liability. The main problem being its cost: you can fare much better with more, cheaper, unshielded ships. Also competing with hangars for internal slots.

This PR reduces shield part costs roughly a 33%, except for three cases:
- Black Shield (-25%), because it is end-game shield capable of significantly reduce the damage of every weapon but fighters, coupled with late game hulls with many external slots to prorate the cost. In ships with huge armor stats that can be OP.
- Multispectral shield (-20%), because it comes with a big boost to stealth (+60) and was already cheap compared to Plasma shield (9 vs 10 strength).
- Robotic Interface Shield (-42%), because it was even less popular than the other shields and seemed to deserve a bigger bump.

Edit: Deflector shield (strength 5) goes from 50 to 30 (-40%). Maybe 35 PP would make more sense...

[Forum thread](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11547)

### DIFF
--- a/default/scripting/ship_parts/Shield/SH_BLACK.focs.txt
+++ b/default/scripting/ship_parts/Shield/SH_BLACK.focs.txt
@@ -5,7 +5,7 @@ Part
     class = Shield
     capacity = 15
     mountableSlotTypes = Internal
-    buildcost = 150 * [[FLEET_UPKEEP_MULTIPLICATOR]] * [[SHIP_PART_COST_MULTIPLIER]]
+    buildcost = 100 * [[FLEET_UPKEEP_MULTIPLICATOR]] * [[SHIP_PART_COST_MULTIPLIER]]
     buildtime = 6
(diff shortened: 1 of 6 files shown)


Morlic-fo:
> except for three cases

SH_DEFLECTOR  is reduced by 40%.

Oberlus:
> > except for three cases
> 
> SH_DEFLECTOR is reduced by 40%.

Yes... Do you think 35 PP (-30%) would make more sense?

Meanwhile description updated.

magnate:
I just popped by to say bravo Oberlus, great minds think alike (except about chaff!)

Oberlus:
Cost of deflector shield set to 35 instead of 30 to make it more consistent with the other costs reductions. Ready to merge.

PR #2894, opened 2020-04-25
Add codeowners for russian translation

Cjkjvfnby:
This feature allows to auto-assign reviewers.  This is useful for external collaborators, who can do a PR but have no permission to request a review. 

If multiple persons are assigned, the first one should leave a comment about his plans to avoid double review.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
+# Code owners for translation stringtables
 ru.txt @Cjkjvfnby @o01eg
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+ru.txt @Cjkjvfnby @o01eg


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+ru.txt @Cjkjvfnby @o01eg
```
adrianbroher: Please use an absolute path to the file as you are referring to that specific translation file.

```suggestion
/default/stringtables/ru.txt @Cjkjvfnby @o01eg
```

adrianbroher (changes requested):

.github/CODEOWNERS
Cjkjvfnby: I reference all of them.  If for some reason this file will appear in the other place, I want to be a reviewer too.

o01eg (approved):

.github/CODEOWNERS
geoffthemedio: @adrianbroher is that reasonable?

adrianbroher: It is, but I personal would prefer explicit pathes due to their explicit nature, however a title comment like `# Code owners for translation stringtables` before this block would be fine too.  `ru.txt` will probably be unique within the tree, even when moving.

geoffthemedio: @Cjkjvfnby Could you add a title comment for the section as proposed?

PR #2896, opened 2020-04-26
Set ship speed for new ships to be affected by Interstellar Logistics

Oberlus:
As per @geoffthemedio [suggestion](https://www.freeorion.org/forum/viewtopic.php?f=28&t=11639&p=101086#p101086).
I don't know what should I do for this to appear as authored by Geoff.

Tested and works fine.
Not extensively tested to see if it might cause any trouble.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1974,7 +1974,7 @@ void Empire::CheckProductionProgress() {
                 // have been applied, letting new ships start with maxed
                 // everything that is traced with an associated max meter.
                 ship->SetShipMetersToMax();
-                // set ship speed so that it can be affected by Interstellar Logistic check for non-zero speed
+                // set ship speed so that it can be affected by non-zero speed checks
                 if (auto* design = GetShipDesign(elem.item.design_id))
                     ship->GetMeter(METER_SPEED)->Set(design->Speed(), design->Speed());
                 ship->BackPropagateMeters();
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1974,6 +1974,9 @@ void Empire::CheckProductionProgress() {
                 // have been applied, letting new ships start with maxed
                 // everything that is traced with an associated max meter.
                 ship->SetShipMetersToMax();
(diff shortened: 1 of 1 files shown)


geoffthemedio:
> Not extensively tested to see if it might cause any trouble.

A bit of such testing would be good...

Oberlus:
> A bit of such testing would be good...

I'm in a SP game with 6 AIs, turn 124. None of the AIs have researched Interstellar Logistics yet, but so far there was no error I could see.
The only issue I found is that tamed monsters do not benefit from the change (i.e. they still need one turn to get the +20 speed).

geoffthemedio:
I think (tamed and most other) monsters are created by effects, not the empire production system, so that would make sense...

You might need to script an after-creation effect, to increase the newly-created monster's speed, for the CreateShip effect that creates the monster.

It might also depend on effect priorities, or might not be possible if the monster is being created without a more-complicated after-creation effect that explicitly checks for the logistics tech and applies the equivalent effect to the monster... The effectsgroup scope conditions that decide when to generate monsters are probably being evaluated at the same time (ie. before monster creation) as the scope conditions that decide where to apply the logistics effect, so the monster is created after it is decided which objects to apply the logistics effect to, on a given turn.

Oberlus:
> You might need to script an after-creation effect, to increase the newly-created monster's speed, for the CreateShip effect that creates the monster.

I'll have a look at that today. But it might take a while.
In any case, I could live with monsters taking an extra turn, because they are seldom used compared to regular ships, despite the weird difference that introduces between ships and monsters.

So I would prefer merging this as is and dealing with the monster part in another PR.

Empire/Empire.cpp
```diff
@@ -1974,6 +1974,9 @@ void Empire::CheckProductionProgress() {
                 // have been applied, letting new ships start with maxed
                 // everything that is traced with an associated max meter.
                 ship->SetShipMetersToMax();
+                // set ship speed so that it can be affected by Interstellar Logistic check for non-zero speed
```
geoffthemedio: not a huge fan of the comment referring to specific scripted content...

Oberlus: Hopefully fixed

PR #1020, opened 2016-10-03
Add documentation for the correct usage of GitHub Issue labels

adrianbroher:
This was requested in the Forums by @dbenage-cx .

@Vezzra There are still some unclear labels for me. Those are marked with `???`.  Maybe you can help me with coming up for a better explanation.

PR #2897, opened 2020-04-26
Add GitHub project label workflow

adrianbroher:
This PR is a first test of using [GitHub workflows](https://github.com/features/actions).  It makes the current project labels managable by configuration files within the repository.

The added workflow will run on every commit and use the newly introduced `.github/project-labels.yml` file to create, update, rename or delete the existing pull request and issue labels based on the configured values.

When providing a PR on said file the update only consists of a dry run, indicating what labels will be modified. Only after merging to master the changes will be applied.

This PR also moves the existing `labels.md` documentation into the  `doc` folder and formats them.

I had to change the label description length to 100 characters or less, so I made them more concise.


adrianbroher:
Sample run on my personal repository: https://github.com/adrianbroher/freeorion/actions/runs/88337226

.github/project-labels.yml
```diff
@@ -0,0 +1,142 @@
...
+    or similar documentation.
+- name: component:translation
+  color: "d93f0b"
+  description: >
+    The Issue/PR deals with the game terminology and the translation process
+    in general.
+- name: component:UI
+  color: "d93f0b"
+  description: >
+    The Issue/PR deals with the game interface, a consistent representation
```
geoffthemedio: Confused by `a consistent representation style in-game`

.github/project-labels.yml
```diff
@@ -0,0 +1,142 @@
...
+    infrastructure.
+- name: component:technical docs
+  color: "d93f0b"
+  description: >
+    The Issue/PR deals with technical documentation like API, style, workflow
+    or similar documentation.
+- name: component:translation
+  color: "d93f0b"
+  description: >
+    The Issue/PR deals with the game terminology and the translation process
```
geoffthemedio: The Issue/PR provides or updates translations or deals with the translation process in general.

.github/project-labels.yml
```diff
@@ -0,0 +1,142 @@
...
+    distribution or helper tools.
+- name: component:internal
+  color: "d93f0b"
+  description: >
+    The Issue/PR deals with any project component that has no explicit
+    `component` label.
+- name: component:network
+  color: "d93f0b"
+  desscription: >
+    The Issue/PR deals with TCP/IP networking and establishing a test network
```
geoffthemedio: The Issue/PR deals with network communication to the server and between clients and the server.

.github/project-labels.yml
```diff
@@ -0,0 +1,142 @@
...
+    The topic of the Issue/PR is or was already covered by another Issue/PR.
+- name: status:superseded
+  color: "c5def5"
+  description: >
+    The PR implementation is insufficient and a follow up implementation PR
+    exists.
+- name: status:help wanted
+  color: "c5def5"
+  description: >
+    Other developers or external contributors are welcome to take over this
```
geoffthemedio: can ask for help with part of something without wanting someone to take it over completely...

.github/project-labels.yml
```diff
@@ -0,0 +1,142 @@
...
+    underlying bug.
+- name: status:testing requested
+  color: "c5def5"
+  description: >
+    The Implementation can't be tested sufficiently by the developer and
+    require support.
+- name: status:won't fix
+  color: "c5def5"
+  description: >
+    The Issue won't be fixed as implementation or maintainance effort required
```
geoffthemedio, line 126: can probably omit the (likely incomplete) list of reasons why

.github/project-labels.yml
adrianbroher: Updated.

.github/project-labels.yml
adrianbroher: Further simplified to
```
The Issue/PR deals with network communication between clients and server.
```

.github/project-labels.yml
adrianbroher: ```Other developers or external contributors are welcome to support developing this Issue/PR.```

.github/project-labels.yml
adrianbroher: I copied the description from the previous documentation and I tried to condense it to the essentials, which is usually "it's too much effort for the gain".

.github/project-labels.yml
adrianbroher: Not sure what is meant by that, tbh.

adrianbroher:
@Vezzra any opinion on the descriptions?

Otherwise I would consider this PR done.

Vezzra:
Yes, concerning the "category:*" labels, specifically "category:feature", "category:tweak" and "category:refactoring". As you have been the one introducing this label scheme, my understanding what these categories are supposed to mean might be wrong, but I always understood them like this:

*Feature*: Something that adds new functionality

*Tweak*: Something that modifies/changes/expands existing functionality

*Refactoring*: Something that only changes internal implementation, but does not change existing functionality in any way nor introduces any new functionality (like code reorganisation/restructuring etc.)

Scope (that is, how big or small the features/changes/whatever in question are) is not of any relevance when it comes to these categories.

Meaning, I understand the categories in a purely technical sense regarding how "functionality" is affected. That would make most sense to me, but you (or @geoffthemedio) might have a different idea...

Of course, this discussion is actually beyond the scope of this PR, but it affects it nonetheless.

adrianbroher:
The `category:tweak` label was proposed by @geoffthemedio as seen in [this discussion](https://github.com/freeorion/freeorion/pull/1020#discussion_r81595117) and I have still no idea where this should be applied properly as I still consider the other categories sufficient.

I agree on your definition for the other categories however.

geoffthemedio:
The key aspect of a "tweak" is that it's small. It's not big or new or different enough to be a feature. It's not substantial enough to be refactoring. It may or may not change how something works, balance, or appearance. I did and still want this category because I find it very useful, as many requests don't fit in the other categories, but do seem well-described as a "tweak".

.github/project-labels.yml
geoffthemedio: You don't know what I meant, or don't know what the description means?

Maybe `The Issue/PR deals with the game user interface, graphical or other`

Vezzra:
I think at this point we need to clarify what purpose exactly the "category:*" labels should serve, by which I mean what aspect(s) are supposed to get "categorized".

Based on the current descriptions in `labels.md` it looks like an inconsistent mix now:

"category:feature" refers to functionality (adding new functionality), but not to scope. 

"category:tweak" (apparently) refers to "change" in general, without regard if functionality is affected or not, *and* scope (only small/insignificant).

"category:refactoring"  again refers to functionality, and "change": it means changes that do not affect functionality. Scope again is not relevant.

Correct me if I misunderstood any of these definitions in `labels.md`.

I think we need to decide what aspects these labels are supposed to categorize (how functionality is affected, scope, "type" of change: addition/modification), and then define the categories accordingly. Otherwise assignment of these labels will stay a somewhat subjective matter (as it is now).

This discussion should probably happen on the forum.

.github/project-labels.yml
adrianbroher: I'm talking about the label description itself.  The `consistent representation` part does not fit into the listing. Going with your suggestion.

Vezzra:
> I think at this point we need to clarify what purpose exactly the "category:*" labels should serve, by which I mean what aspect(s) are supposed to get "categorized". [...] This discussion should probably happen on the forum.

As an afterthought, this probably should not block the PR from being merged. We can merge the PR with the current descriotions for the category labels. Whenever we find the time and will to dive into the discussion we can update the descriptions according to the consensus once we reach one.

PR #2898, opened 2020-04-26
Use more GitHub issue template features

adrianbroher:
This PR updates the existing issue templates to utilize the newest capabilities of GitHub:

* Users are now presented a selection for picking the correct issue template:

issue-template-chooser

* Issue metadata are now encoded in the YAML frontmatter.  This allows to populate the issue with informations like:
  * A default title
  * Issue labels
  * An assignee

I didn't look if the templates still represent reality or if the template chooser has the correct entries available.  I would appreciate feedback here.


geoffthemedio:
>  I would appreciate feedback here.

Can or should something be tested in operation, or should someone just look at the script changes?

adrianbroher:
> Can or should something be tested in operation, or should someone just look at the script changes?

There is nothing to test in operation, so a text review should be fine. However my request is about if the current issue templates are sufficient or if there should be a separate template for feature requests or if there is another topic, which should be covered by a new issue template.

geoffthemedio:
I'm not really sure how this will look or work, but a "Request Feature" button sounds reasonable, as it would likely warrant different info being provided.

Are there also templates for making pull requests?

adrianbroher:
> I'm not really sure how this will look or work, but a "Request Feature" button sounds reasonable, as it would likely warrant different info being provided.

You create a `.github/ISSUE_TEMPLATE/feature.md` file, add a [YAML frontmatter](https://github.com/freeorion/freeorion/pull/2898/commits/ae133711a3b3ac43d4785ae52a19714f685bd12e#diff-0a9248bc9364aa2bae3116ca7a945c5dR1-R5) in front of the document, followed by a regular markdown file.

The file name decides on the order in the selection menu, the title and description are determined by the `name` and `about` frontmatter keys. The `title`, `labels` and `assignees` frontmatter determine the prepopulated values in the issue.  See [this manual](https://help.github.com/en/github/building-a-strong-community/manually-creating-a-single-issue-template-for-your-repository).

> Are there also templates for making pull requests?

Yes there are, they follow the same conventions as the issue templates, but live in the `.github/PULL_REQUEST_TEMPLATE` directory.  However there is no user selection menu, but you have to use instead a url query parameter to select the template.

Vezzra:
@adrianbroher, assuming you just moved the existing templates into the `ISSUE_TEMPLATE` subfolder, and after a quick glance at them, I'd say those more or less still represent reality.

The template for bug reports seems fine, but as @geoffthemedio said, a separate one tailored for feature requests will probably make sense, as the current template for issues is definitely for bug reports, not so much for feature requests.

The template for the release issue is still what I use when I open the tracking issue for a release.

adrianbroher:
@Vezzra and @geoffthemedio Can you throw some ideas in the room about the feature request template?  I have hardly any idea on what the template should/should not contain as feature descriptions are pretty much free form.

Vezzra:
Good point. You're right, feature requests are much more "free form" than bug reports, there isn't really much a template can reasonably provide.

After a bit of mulling over, this is what I can think of:

First and most obvious of course setting the "category:feature" label.

Second would be to provide a simple two-section structure: "Description" and "Link to corresponding forum thread". While feature requests can and should be made here on the issue tracker, we do not want design discussions take place in the comment sections of issues/PRs on github.

Currently it happens again and again that design discussions start in comment sections, especially those of feature requests (naturally), and we have to move these discussions to the forum. *If* we move them, sometimes we just let it slip.

Having that simple structure provided by a template, with proper captions and comments explaining what we expect could help nudging people to open corresponding design threads on the forum and putting a link in the feature request.

adrianbroher:
One way to think about this would be that the feature request acts blueprint, while the thread contains  the discussion for working out the blueprint.

geoffthemedio:
My main reason for wanting a separate template for feature requests was that it should not contain all the requests for info that a bug report does, which don't really apply to most feature requests.

A free-form description and a link to a forum thread are reasonable to include.

adrianbroher:
I have updated the bug template a bit by removing the references to features and giving some additional instructions about save game uploads and I have added a template for feature requests.  You can take a look at their usage on my personal repository:

https://github.com/adrianbroher/freeorion/issues/new/choose

geoffthemedio:
I'd rephrase the feature request description to something like "Propose a new or improved feature for FreeOrion."

Vezzra (approved):

Issue #2901, opened 2020-04-28
Structure scaling rework and four bouts


agrrr3:
I open this for tracking a structure rescaling work. This will broaden the design space for multishot and heighten influence of ship design on combat. When introducing the disruptor arc it was almost impossible to find a design spot which is not either underpowered or overpowered or has other strong drawbacks (even the arc disruptor is restricted for balance reasons - it does not scale with pilot skill).

Forum discussion with benefits/tradeoffs etc: [Four combat bouts as default](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11665)

Game Aim:
* Higher structure values so integer based damages can scale more finely - currently a multishot weapon with four shots has a minimum damage upgrade like the best ship weapon (4 damage for one step)
* Increase default bouts to four from three while keeping current balance (so battles take about the same amount of turns)

Basic implementation idea:
- multiply structure by 6 because we often have 2 and 3 divisions [Re: Base weapon stats for (mostly) integer values](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11563&p=99935&hilit=bouts#p99906) and multiply the weapon damage accordingly
- multiply structure by 4/3 to spread the structure to account for the ship weapon damage done by one extra bout
- the resulting structure multiplier is 8 (== 6 * 4 / 3 )

Implementation Plan (always checking stringtables and AI in each step):
- [x] set default number of bouts to 4
- [x] armor and hull structure scaling by 8 (easy via hip_hulls.macros:STRUCTURE_FACTOR)
- [x] weapon scaling (probably adding SHIP_WEAPON_DAMAGE_FACTOR 6 and FIGHTER_WEAPON_DAMAGE_FACTOR 6 and related rules)
- [x] weapon upgrade techs (scale with SHIP_WEAPON_DAMAGE_FACTOR)
- [x] damage control techs (scale with SHIP_STRUCTURE_FACTOR)
- [x] ship shields, including robotic shields, GOOD_SHIELDS species and ship hulls 2x asteroid swarms (SetMaxShield bonus scale with SHIP_WEAPON_DAMAGE_FACTOR)
- [x] mine effects (SHIP_STRUCTURE_FACTOR)
[...]

PR #2908, opened 2020-04-29
Four bouts and structure rescaling

agrrr3:
Implementation for #2901 

Not so sure that I should be applying the scale factors so much in the FOCS effects.

Maybe SetMaxStructure and similar effects should always directly apply the appropiate scaling.

@geoffthemedio what do you think - apply the scaling factors in FOCS or directly e.g. in the FOCS effect implementation?


universe/ShipPart.cpp
```diff
@@ -47,21 +47,21 @@ namespace {
...
         typedef std::vector<std::unique_ptr<Effect::Effect>> Effects;
         auto scope = std::make_unique<Condition::Source>();
         auto activation = std::make_unique<Condition::Source>();
 
         auto value_vr = std::make_unique<ValueRef::Operation<double>>(
-            ValueRef::PLUS,
-            std::make_unique<ValueRef::Variable<double>>(
-                ValueRef::EFFECT_TARGET_VALUE_REFERENCE, std::vector<std::string>()),
-            std::make_unique<ValueRef::Constant<double>>(increase)
+             ValueRef::PLUS,
```
geoffthemedio: why extra indenting?

universe/ShipPart.cpp
```diff
@@ -100,6 +100,41 @@ namespace {
...
+        auto increase_vr = std::make_unique<ValueRef::Constant<double>>(fixed_increase);
+        return IncreaseMeter(meter_type, part_name, std::move(increase_vr), allow_stacking);
+    }
+
+    // create effectsgroup that increases the value of the part meter
+    // of type \a meter_type for part name \a part_name by the fixed
+    // amount \a base_increase and the value of the game
+    // rule of type double with the name \a scaling_factor_rule_name
+    std::shared_ptr<Effect::EffectsGroup>
+    IncreaseMeter(MeterType meter_type, const std::string& part_name,
```
geoffthemedio: perhaps call it something different... like `IncreaseMeterRuleScaled` or so...?

universe/ShipPart.cpp
```diff
@@ -234,7 +277,15 @@ float ShipPart::Capacity() const {
 }
 
 float ShipPart::SecondaryStat() const
-{ return m_secondary_stat; }
+{
```
geoffthemedio: `{` to previous line

geoffthemedio:
> what do you think - apply the scaling factors in FOCS or directly e.g. in the FOCS effect implementation?

I think rules for scaling content make more sense in the scripts than in the engine internal C++ making scripts do something other than they were written to do...

agrrr3:
Curse your refactorings adrian

adrianbroher:
> Curse your refactorings adrian

Stop doing pointless whitespace edits and use [`git mergetool`](https://www.git-scm.com/docs/git-mergetool).

agrrr3:
While I implemented mines:
* not sure if rawtext without quotes works - need to test this
* also do not know if one can escape chars in macro params - i wanted to use parenthesis around the GameRule
 
> > what do you think - apply the scaling factors in FOCS or directly e.g. in the FOCS effect implementation?
> 
> I think rules for scaling content make more sense in the scripts than in the engine internal C++ making scripts do something other than they were written to do...

Not sure I follow. But I read it as: using rules for scaling in the backend for the values set by FOCS-content-specification is ok (like already done before for speed and ship structure), but for FOCS-effects it is not good. 

So for this PR in the end I add the scaling rules in FOCS  to all SetMaxStructure, SetMaxShields, SetMaxDefense, SetMaxDamage and to some SetMaxCapacity cases. Also I need to set it to all non-max-relative calls like SetStructure +=5 becomes SetStructure += SHIP_STRUCTURE_FACTOR*5 .

I will go forward with that. Or should I look for a less invasive solution?

agrrr3:
> > Curse your refactorings adrian
> 
> Stop doing pointless whitespace edits and use [`git mergetool`](https://www.git-scm.com/docs/git-mergetool).

And that will magically move the hull code to the new files? I really look forward to it

adrianbroher:
No, but it will open the conflicting files with a difftool where you can pick the different hunks for copying them over.  With vimdiff you even pick foreign files for diff comparision.

adrianbroher:
You can even do it in meld by selecting another file within the diff panel.

agrrr3:
Hm, I got here some pytests failing because of not being able to find a function in freeOrionAIInterface.
My python-foo is at a loss here also freeOrionAIInterface seems to be specially handled for pytest. Hm @adrianbroher any ideas?

agrrr3:
Ok, the tweak part is done.
Can somebody add testing-requested tag so I can get feedback on the balance?

There is one particular ugly thing and that is that I now hardcoded the base hull structure as one cant use calculation for structure value and as the value should be the base damage of the Arc Disruptor it needs to be calculated. I could probably script up the correct calculation by adding a effect but that is also ugly.

universe/ShipPart.cpp
agrrr3: I probably aligned with the variable definition above, will make it four spaces.

universe/ShipPart.cpp
agrrr3: Yes, probably a good idea to differentiate the call signatures a bit more.

agrrr3:
Implemented the review feedback and rebased on current weekly build. 

I probably add the base hull structure effect I mentioned in the previous post as a crutch. It is better than hardcoding the value for one scenario I think.

@geoffthemedio I would like to get the refactoring part into master soonish (before the next multiplayer game starts, so the players can try higher bout numbers), that means i would add a commit which sets the rule defaults to current values (three bouts and scaling rules 1.0) and after getting positive feedback I do another PR with changed defaults.

@Vezzra  would you want four bouts as default in 0.4.10 if feedback is in time? Else I suggest to merge that later part into master after 0.4.10 is split.

Vezzra:
@agrrr3, this PR as well as the corresponding issue it implements (#2901) have been assigned to the "post release" milestone, which means they are not supposed to go into the upcoming 0.4.10 release.

We're in a phase where we're trying to wrap the release up, and incorporating/merging big new features or major revisions/reworks needs to be hold off (unless they address a serious issue), otherwise we'll never get the release out.

So, unless there is some very important reason why this has to go into 0.4.10, I strongly recommend to hold off merging until the release branch has been created.

agrrr3:
The reason why this should get into 0.4.10 is too have it in the
multiplayer build to get feedback. In the current form it is a refactoring
(no game changes if you do not set the new rules) with low risk.

But if 0.4.10 gets out of the door soon there is no need to get it into
that release.

agrrr3:
@o01eg @Oberlus will (some of) the next multiplayer game(s) be based on 0.4.10 ? Should the feature to be able to do rescaling/four bouts go in or wait after release?

Oberlus:
> @o01eg @Oberlus will (some of) the next multiplayer game(s) be based on 0.4.10 ? Should the feature to be able to do rescaling/four bouts go in or wait after release?

Since in its current state it is just refactoring, I really belive it should go into v0.4.10 for playtesting. The sooner the better.

o01eg:
> @o01eg @Oberlus will (some of) the next multiplayer game(s) be based on 0.4.10 ? Should the feature to be able to do rescaling/four bouts go in or wait after release?

Yes, multiplayer games will follow release branch, and at least one game will use stable 0.4.10 release.

default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
```diff
@@ -15,11 +16,23 @@ Hull
     buildTime = 3
     tags = [ "PEDIA_HULL_LINE_GENERIC" ]
     location = OwnedBy empire = Source.Owner
-    effectsgroups =
+    effectsgroups = [
         [[REGULAR_HULL_DETECTION]]
+
+        // There are weapons designed to kill a base hull with a single shot (Arc Disruptor)
+        // Also mines should be able to down base hull
+        // So we scale the base hull with the lower of those scaling factors
```
geoffthemedio, line 24: why lower not higher?

default/scripting/ship_parts/Shield/SH_ROBOTIC_INTERFACE_SHIELDS.focs.txt
```diff
@@ -25,7 +25,7 @@ Part
             activation = ([[ROBOTIC_SHIELD_EFFECT]] >= [[BEST_SHIELD_EFFECT]])
             stackinggroup = "ROBOTIC_SHIELD_PART_STACK"
             accountinglabel = "ROBOTIC_INTERFACE_EFFECT"
-            effects = SetMaxShield value = Value - [[BEST_SHIELD_EFFECT]] + [[ROBOTIC_SHIELD_EFFECT]]
+            effects = SetMaxShield value = Value - [[BEST_SHIELD_EFFECT]] + ( [[ROBOTIC_SHIELD_EFFECT]] * (GameRule name = "RULE_SHIP_WEAPON_DAMAGE_FACTOR") )
```
geoffthemedio: Why `GameRule name = "RULE_SHIP_WEAPON_DAMAGE_FACTOR"` not `[[SHIP_WEAPON_DAMAGE_FACTOR]]` or `[[SHIP_SHIELD_FACTOR]]` ?

default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
Oberlus: So that both (mines damage and arc disruptor single shot) are able to do the 1-shot kill.

default/scripting/ship_parts/Shield/SH_ROBOTIC_INTERFACE_SHIELDS.focs.txt
agrrr3: Did not know what style to prefer and if macro expansion would work everywhere (was before asking the parens question in the forum). Later on it became clear that using a macro name is more appropriate as I was reusing the same rules for different purposes (e.g. damage rule for shield scaling) so decoupling intent and the used rule became necessary. Will change all those rules use to macros for consistency.

default/scripting/ship_hulls/SH_COLONY_BASE.focs.txt
agrrr3: Another variant would be to not scale the base structure of the colony base hull at all. Reinforced hull scales with the weapon damage so that would always be correct (but that is probably a detail for followup PR when we actually set new default values)

default/scripting/ship_parts/Shield/SH_ROBOTIC_INTERFACE_SHIELDS.focs.txt
agrrr3: Refactor to macros done

Vezzra:
Ok, if the defaults for the new game rules are set in a way that current behavior remains unchanged, I guess this can go in. "post release" assignment has been removed.

default/python/AI/CombatRatingsAI.py
```diff
@@ -299,11 +299,24 @@ def _rating():
...
+        num_bouts = fo.getGameRules().getInt("RULE_NUM_COMBAT_ROUNDS")
+        for firing_bout in range(num_bouts - 1):
+            if firing_bout < full_launch_bouts:
+                flying_fighters += (flying_fighters * survival_rate) + launch_rate
+            elif firing_bout == full_launch_bouts:
+                # now handle a bout with lower capacity launch
+                flying_fighters += (flying_fighters * survival_rate) + (capacity % launch_rate)
+            else:
+                flying_fighters += (flying_fighters * survival_rate)
+            total_fighter_damage += fighter_damage * flying_fighters
```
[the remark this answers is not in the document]
agrrr3: Thanks for finding a bug. I need to keep two values I think, aggregated shots and flying_fighters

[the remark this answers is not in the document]
agrrr3: Opened issue #2973 for this.

default/scripting/ship_hulls/organic/SH_BIOADAPTIVE.focs.txt
```diff
@@ -47,7 +47,7 @@ Hull
         EffectsGroup
             scope = Source
             accountinglabel = "ORGANIC_GROWTH"
-            effects = SetMaxStructure value = Value + min(Source.Age*0.5, 25)
+            effects = SetMaxStructure value = Value + ( min(Source.Age*0.5, 25) * [[SHIP_STRUCTURE_FACTOR]] )
```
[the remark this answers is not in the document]
agrrr3: Also noted in #2973

Oberlus:
I spotted an undesired effect of putting the damage scale in the backend instead of in the FOCS files:

Flak cannon damage is misleading now, you have to divide by the damage factor to get the actual number of flak shots. For example, a regular pilot species get 18 damage flaks, 18/6=3 PD shots. Not cool.

But it seems there are worse problems than that. This is a Sentinel ship:
FlakCannon21
How many flak shots does it have? 3.5?

The weapon stats of some monsters are modified in their hull definitions:
https://github.com/freeorion/freeorion/blob/1f2ee39854af60db4d4325be401b9ee7115cae0c/default/scripting/ship_hulls/monster/SH_GUARD_MONSTER_BODY.focs.txt#L27-L30
I don't know if that code is correct, regardless of the damage and structure scaling problem. Shouldn't the Capacity and MaxCapacity of SR_WEAPON_0_1 be changed instead of Damage and MaxDamage?

agrrr3:
Afaicr damage and capacity are the same for ship weapon parts.

It makes sense for fighter-only targeting to show only the number of shots.
Quick fix could be to suppress the default capacity effect.
I would rather fix this in the UI though (UI depending on targeting, also
add an option to switch between different target sets).

And the monsters should get something in exchange for their nerfed flak.

agrrr3:
Also if I am not mistaken the SetDamage effects should not automatically
scale.
So 3 flak * (1dam * 6scale + 1effect) == 21

Monsters need love

Oberlus:
> Afaicr damage and capacity are the same for ship weapon parts.

Well, then whatever stat is the one that determines fire rate. SecondaryStat?

> It makes sense for fighter-only targeting to show only the number of shots.

If at some point someone changes the targetting conditions in FOCS, how will backend handle that?

> Quick fix could be to suppress the default capacity effect. I would rather fix this in the UI though (UI depending on targeting, also add an option to switch between different target sets).

I would move all damage/structure/shield scaling to FOCS, since part of it is already there. Currently it's a bit inconsistent.

agrrr3:
> whatever stat is the one that determines fire rate. SecondaryStat?
>
Yes, that would be a buff.

It makes sense for fighter-only targeting to show only the number of shots.
>
> If at some point someone changes the targetting conditions in FOCS, how
> will backend handle that?
>
That can not be completely automated. Big topic.
Some options
Part specification contains a UI-damage hint which chooses the right widget
Automatic support for well known target conditiobs
Scriptable association from target condition name to UI layout
Scripting creation of a target dummy object, so the backend can actually
simulate the damage against that

It's on my 0.5 list of want to do%s

> Quick fix could be to suppress the default capacity effect. I would rather
> fix this in the UI though (UI depending on targeting, also add an option to
> switch between different target sets).
>
> I would move all damage/structure/shield scaling to FOCS, since part of it
> is already there. Currently it's a bit inconsistent.
>
Think so too. But the parsers do not support that right now (valueref
definition of hull/part structure/damage). Implementing that could be a
huge pain.
Suppressing the default effect is rather straightforward.

And also for sensible damage estimation you need to consider the targets
anyway.

PR #2880, opened 2020-04-20
Simplify public logger interface

adrianbroher:
The current header file `util/Logger.h` leaks some implementation details, which are actually should not be exposed as they increase the compile time for every file which includes it.

This PR tries to sweep some of it under the rug to improve compile time for users of the logger.

There is also potential to remove the boost::signal2 header from the logger but that requires extra work and care so this PR can be merged as is.

Some statistics:

I've used the `g++ -ftime-report` but left out the compilation steps, which were below 10% contributing to the overall time. Also keep in mind that the display of time variables is maybe a subpart of another bigger step or there was waiting for some IO operation, which is why the the percentage maybe does not sum up.

Mostly relevant is the `usr` column or `user` row, which where the compiler were actually working, `sys` indicates times where the operating system were doing work for the compiler (e.g. opening files or allocating memory). And the `wall` column or `real` row  is the overall real time (wall clock) spent.

The other g++ command uses the `-H` flag, which prints out all includes files recursively.  It does a bit of post processing and prints out the included files, as the preprocessor would find it, this may or may not repeat subtrees.

master: Overall build time (clean build tree and empty ccache):
```
$ time ninja
[...]
real	33m44,170s
user	183m17,328s
sys	13m58,895s
```

simplify-logger: Overall build time (clean build tree and empty ccache):
```
$ time ninja
[...]
real	27m25,150s
user	165m45,789s
sys	11m40,391s
```

master:parse/ConditionParser7.cpp (A template heavy implementation file)
```
$ g++ -ftime-report   -DBINPATH="/usr/local/bin" -DBOOST_ALL_DYN_LINK -DBOOST_GIL_IO_ENABLE_GRAY_ALPHA -DENABLE_BINRELOC -DFONS_USE_FREETYPE -DFREEORION_BUILD_PARSE -DFREEORION_LINUX -DNDEBUG -DSHAREPATH="/usr/local/share" -I/home/adrian/Projects/freeorion/freeorion.git/GG -IGG -I/usr/include/freetype2  -O2 -DNDEBUG -fPIC -fvisibility=hidden   -Wall -std=c++14 -o CMakeFiles/freeorionparseobj.dir/parse/ConditionParser7.cpp.o -c /home/adrian/Projects/freeorion/freeorion.git/parse/ConditionParser7.cpp

Time variable                                   usr           sys          wall               GGC
 phase setup                        :   0.00 (  0%)   0.00 (  0%)   0.01 (  0%)    1352 kB (  0%)
[...]

PR #2909, opened 2020-04-29
Improve python logger interface

Morlic-fo:
Simplifications to the logger interface and fix for some formatting issue introduced in #2880: Since that change, the "python :" part was printed twice in logs, e.g.

```01:37:50.761509 {0x00002b1c} [debug] python : InvasionAI.py:451 : python : Invasion eval of P_2338<Adams γ I>```

### DIFF
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -136,7 +136,7 @@ class _stdXLikeStream:
             (filename, line_number, function_name) = ("", "", "")
 
         try:
-            self.logger(msg, "python", str(os.path.split(filename)[1]), str(function_name), str(line_number))
+            self.logger(msg, str(os.path.split(filename)[1]), str(line_number))
 
         finally:
             # Explicitly del references to the caller's frame to avoid persistent reference cycles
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: default/python/common/configure_logging.py [hunk 1]
```diff
@@ -136,7 +136,7 @@ def write(self, msg):
             (filename, line_number, function_name) = ("", "", "")
```
adrianbroher: Unused `function_name`.

python/LoggingWrapper.cpp
```diff
@@ -49,91 +46,27 @@ namespace {
...
-    void PythonLoggerError(const std::string& msg, const std::string& logger_name,
-                           const std::string& filename, const std::string& function_name, const std::string& lineno)
-    {
-        PythonLoggerWrapper(LogLevel::error, msg, logger_name, filename, /*function_name,*/ lineno);
+        auto logger_func = [&](const std::string& arg) {
+            FO_LOGGER(log_level, python) << boost::log::add_value("SrcFilename", filename)
+                                         << boost::log::add_value("SrcLinenum", lineno)
+                                         << arg;
+        };
+        send_to_log(log_stream, msg, logger_func);
```
geoffthemedio, line 59: When is `logger_func` called? Are you sure you want to capture `filename` by reference? Capturing everything you need explicitly might be clearer in this case given the macro use...

adrianbroher: `send_to_log` splits up the python logger input into lines and uses the logger func to call back onto the boost::log logger.

https://github.com/freeorion/freeorion/pull/2909/files/67c4d8f9d33d90bbd997306763d09dc9a2bb6edd#diff-518a46164dacf94517c4776f1d790ab0R30

PR #2914, opened 2020-04-30
Always create a sentinel defending ancient ruins

l29ah:
No one should be able to get such a powerful artifact for free.

### DIFF
--- a/default/scripting/specials/planet/ANCIENT_RUINS.focs.txt
+++ b/default/scripting/specials/planet/ANCIENT_RUINS.focs.txt
@@ -35,7 +35,6 @@ Special
             scope = Source
             activation = AND [
                 Turn high = 0
-                Random probability = 0.9
                 (GalaxyMaxAIAggression >= 1)
                 (GalaxyMonsterFrequency >= 1)
             ]


Oberlus:
Does this work when monsters are set to None?

l29ah:
Most probably not.

geoffthemedio:
> Does this work when monsters are set to None?

Define "work"? The effects group that creates the guard explicitly checks for `GalaxyMonsterFrequency >= 1` to be active: https://github.com/freeorion/freeorion/blob/master/default/scripting/specials/planet/ANCIENT_RUINS.focs.txt#L40

Oberlus:
> Define "work"?

I meant if the PR made this special to be guarded even when Monsters was set to None. 
You answered it already: no, the Ancient Ruin won't be guarded if Monsters are set to None.
Maybe it would be better to remove that `GalaxyMonsterFrequency >= 1` ?

geoffthemedio:
> Maybe it would be better to remove that `GalaxyMonsterFrequency >= 1` ?

That would mean there would be monsters when the player specifically requested no monsters.

Oberlus:
> > Maybe it would be better to remove that `GalaxyMonsterFrequency >= 1` ?
> 
> That would mean there would be monsters when the player specifically requested no monsters.

In single player games it makes all the sense to give the player control over this.

However, in multiplayer games (which IMO are more important to be balanced) it is common that some players request less than low monsters because even "low" is too much, trapping some empires behind a few monsters than won't let them expand as fast as other empires, and so leaving them in disadvantage for the whole game.
Setting specials to low and monsters to none is not a solution because then some empires could get a huge advantage from a nearby, unguarded Ancient Ruins.
Setting specials also to none gives extra advantage to Self-sustaining and Phototrophic species because Organic, Robotic and Lithic won't be able to get any growth special.

A middle point could be to spawn Ancient Guardians with High Tech special on the Ancient Ruins planet when monsters are set to none.

o01eg:
> However, in multiplayer games (which IMO are more important to be balanced) it is common that some players request less than low monsters because even "low" is too much, trapping some empires behind a few monsters than won't let them expand as fast as other empires, and so leaving them in disadvantage for the whole game.

What if place monsters, that there exists path between home systems without monsters?

Oberlus:
> What if place monsters, that there exists path between home systems without monsters?

IMO, that would be best solution. But also far more complicated to implement.

Morlic-fo:
How about a rule which toggles whether (all) guards are spawned when monsters are set to None?

Vezzra:
> A middle point could be to spawn Ancient Guardians with High Tech special on the Ancient Ruins planet when monsters are set to none.

The Ancient Guardian species have been introduced specifically for that reason: to provide a sufficiently serious roadblock to get to a powerful resource as e.g. Ancient Ruins, without having to resort to a space monster that will not only block that resource, but also the entire system it resides in (leading to the issue of boxed in starting positions).

So Ancient Guardians should be the proper solution to the problem at hand. If you consider them insufficient, why? Are they too weak? In that case I'd propose to either give Ancient Guardians another buff, or make a more powerful version of the Ancient Guardians for more powerful specials (the same way there are different guardian monsters) and use those.

But do not create space monsters when the space monsters setting is "None", even for multiplayer games. Some may want to have specials, but no monsters, not even as guardians for whatever, and we should not remove that option.

l29ah:
Yes, Ancient Guardians are much weaker than the Sentinel or Warden (in case of the Honeycomb special).

Vezzra:
In this case I propose to introduce more powerful versions of the Ancient Guardians, and use those to defend more powerful specials like Ancient Ruins.

PR #2915, opened 2020-04-30
Fix 2900 tweak universe generation constants

### LINKED ISSUE #2900: Improve equilibrium of colonisation opportunities nearby HWs in universe generation
### Environment
* **FreeOrion Version**: v0.4.8, v0.4.9+

### Description
Universe generation is too random regarding starting positions and the opportunities of colonisation for the different empires.

Monsters, natives, paths/chokepoints, specials or any other factor that can affect game balance is out of the scope of this feature request. All those could be addressed in different ways but there is always the chance to disable monsters, natives or specials and to use galaxy shapes and starlane densities that minimize the chances of unfair starting positions.

This is only about colonization opportunities.

The following is a counting of available planets around each empire's HW from ninth multiplayer game in @o01eg's server, for the player with less available good&adeq. planets and the player with more. For each player and planet tolerance, the figures correspond to number of planets at 0, 1, 2, 3, 4, 5 and 6 jumps away from HW, not including HW itself.

- Lucky (swaq)
```
Good: 2 / 0 / 3 / 2 / 0 / 1 / 0
Adeq: 0 / 0 / 2 / 1 / 3 / 3 / 4
Poor: 0 / 0 / 1 / 2 / 1 / 3 / 1
Host: 0 / 0 / 5 / 0 / 2 / 2 / 5
Belt: 1 / 1 / 4 / 3 / 1 / 3 / 7
GG:   1 / 0 / 5 / 1 / 0 / 4 / 7
```

- Unlucky (Oberlus)
```
Good: 0 / 0 / 0 / 1 / 0 / 0 / 2
Adeq: 0 / 0 / 0 / 0 / 2 / 1 / 3
Poor: 0 / 0 / 0 / 2 / 1 / 0 / 1
Host: 2 / 0 / 0 / 0 / 0 / 1 / 9
Belt: 2 / 0 / 0 / 0 / 3 / 0 / 3
GG:   1 / 0 / 0 / 0 / 0 / 0 / 0
```

There are other players rather lucky (danyspin96) or unlucky (Magnate) in the same game, but these two are examples good enough to illustrate how different and unbalanced can be the starting conditions for two players. One on one, an unlucky empire have no chance at all of winning the game against a lucky empire, even if the lucky empires make wrong choices and the unlucky ones do a perfect game.

Once the players begin to realize their situation, it is rather discouraging to know that s/he cannot win (or lose, to a lesser extend).

### Feature request

Improve the universe generation algorithm to disallow too unbalanced colonisation opportunities.

Currently the algorithm ensures that every empire has a minimum of 8 systems (empty or not) in the in the vicinity (3 jumps from HW), a minimum number of planets (regardless of environment!) in that vicinity equal to the minimum between 10 and the number of systems in its vicinity. That seems flawed, looking at the results, on one hand because it does not consider environments, and second because a HW starting in an area with many starlanes will have more available planets (ensured by the algorithm) than a HW starting in a more densely connected area (in the example above, lucky empire had 14 systems in the vicinity, unlucky had 8).

The conditions should not only check for minimums but also ensure that luckiest empire does not have much better colonizable opportunities than the unluckiest empire.

Oberlus:
This is intended to improve balance of colonization opportunities of empires at universe generation.

It reduces differences in raw number of planets availability between different empires due to randomness of number of systems in vicinity and number of planets per system in that vicinity. It does so by including into calculations the galaxy starlane and planet densities and making them all dependent on systems per empire.
It also ensures that empires in big galaxies can start further apart from each other in big galaxies, by allowing longer minimum jump distances between HW systems (up to a minimum of 20 jumps for galaxies with 55 systems per empire), but still allowing the algorithm to fall back to shorter minimum distances if the desired distances are unattainable.

It does **not** address the great variability in planet environments that can happen between different empires. That is, the problem of some empires starting with many good/adequate planets nearby while others only get poor/hostile planets is still present. This will be further alleviated by an incoming rebalancing of the Exobot species based on Voker57's proposal (https://github.com/Voker57/freeorion/commit/8bd4af7b43896c77fe168da99cb88c1c8ed1eda8).

Generating galaxies of 1000 systems and 20 empires take ~2 seconds to generate in an Intel i7-4710HQ laptop (2.5-3.5 GHz) with 8 GB RAM.

Fixes #2900

### DIFF
--- a/default/python/universe_generation/options.py
+++ b/default/python/universe_generation/options.py
@@ -30,25 +30,18 @@ NAMING_LARGE_GALAXY_SIZE = 200  # a galaxy with 200+ star systems is considered 
 # These options are needed in the home system selection/placement process. They determine the minimum number of
 # systems and planets that a home system must have in its near vicinity, define the extend of this "near vicinity", etc.
 
-# The following two options are used to determine the minimum number of systems and planets. This limit is
-# HS_MIN_SYSTEMS_IN_VICINITY systems and HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM planets per system within the near
-# vicinity of a home system, capped at HS_MIN_PLANETS_IN_VICINITY_TOTAL
-# HS_MIN_SYSTEMS_IN_VICINITY = 12
-# HS_MIN_PLANETS_IN_VICINITY_TOTAL = 12
-# HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM = 1
-
-# This option defines the extend of what is considered the "near vicinity" of a home system. This are all systems that
-# are within the number of jumps specified by HS_VICINITY_RANGE.
+# These two options define the minimum and maximum extend of what is considered the "near vicinity" of a home system.
+# This are all systems that are within the number of jumps specified by vicinity_range(systems_per_empire).
 HS_MIN_VICINITY_RANGE = 2
 HS_MAX_VICINITY_RANGE = 7
 
-# This options sets the maximum starting value for the minimum jump distance limit required between home systems.
+# This option sets the maximum starting value for the minimum jump distance limit required between home systems.
 # With large galaxies an excessive amount of time can be used in failed attempts to select home systems, so defining
 # an upper limit for the home system selection process to use when calculating the starting value for the minimum
 # jump distance limit is reasonable.
 HS_MAX_JUMP_DISTANCE_LIMIT = 20
 
-# This options defines the minimum jump distance limit between home systems that should be considered high priority.
+# This option defines the minimum jump distance limit between home systems that should be considered high priority.
 # As long as the jump distance limit which home systems must at least be apart does not get reduced below this limit
 # during the home system selection process, the minimum systems in home system vicinity requirement takes
 # precedence over the jump distance limit. If the jump distance limit drops below this minimum jump distance limit,
--- a/default/python/universe_generation/options.py
+++ b/default/python/universe_generation/options.py
@@ -58,5 +58,5 @@ HS_MIN_DISTANCE_PRIORITY_LIMIT = 4
 # These two options define which types of planets are counted when determining the number of planets in the near
 # vicinity of a home system. HS_ACCEPTABLE_PLANET_SIZES is actually only needed for the process of adding planets
 # to the near vicinity of a home system in case that's needed to meet the limit.
-HS_ACCEPTABLE_PLANET_TYPES = planet_types_real + (fo.planetType.asteroids,fo.planetType.gasGiant,)
-HS_ACCEPTABLE_PLANET_SIZES = planet_sizes_real + (fo.planetSize.asteroids,fo.planetSize.gasGiant,)
+HS_ACCEPTABLE_PLANET_TYPES = planet_types_real + (fo.planetType.asteroids, fo.planetType.gasGiant,)
+HS_ACCEPTABLE_PLANET_SIZES = planet_sizes_real + (fo.planetSize.asteroids, fo.planetSize.gasGiant,)
--- a/default/python/universe_generation/options.py
+++ b/default/python/universe_generation/options.py
@@ -33,29 +33,30 @@ NAMING_LARGE_GALAXY_SIZE = 200  # a galaxy with 200+ star systems is considered 
 # The following two options are used to determine the minimum number of systems and planets. This limit is
 # HS_MIN_SYSTEMS_IN_VICINITY systems and HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM planets per system within the near
 # vicinity of a home system, capped at HS_MIN_PLANETS_IN_VICINITY_TOTAL
-HS_MIN_SYSTEMS_IN_VICINITY = 8
-HS_MIN_PLANETS_IN_VICINITY_TOTAL = 10
-HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM = 1
+# HS_MIN_SYSTEMS_IN_VICINITY = 12
+# HS_MIN_PLANETS_IN_VICINITY_TOTAL = 12
+# HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM = 1
 
 # This option defines the extend of what is considered the "near vicinity" of a home system. This are all systems that
 # are within the number of jumps specified by HS_VICINITY_RANGE.
-HS_VICINITY_RANGE = 3
+HS_MIN_VICINITY_RANGE = 2
+HS_MAX_VICINITY_RANGE = 7
 
 # This options sets the maximum starting value for the minimum jump distance limit required between home systems.
 # With large galaxies an excessive amount of time can be used in failed attempts to select home systems, so defining
 # an upper limit for the home system selection process to use when calculating the starting value for the minimum
 # jump distance limit is reasonable.
-HS_MAX_JUMP_DISTANCE_LIMIT = 10
+HS_MAX_JUMP_DISTANCE_LIMIT = 20
 
 # This options defines the minimum jump distance limit between home systems that should be considered high priority.
 # As long as the jump distance limit which home systems must at least be apart does not get reduced below this limit
 # during the home system selection process, the minimum systems in home system vicinity requirement takes
 # precedence over the jump distance limit. If the jump distance limit drops below this minimum jump distance limit,
 # the process is restarted giving the jump distance limit precedence.
-HS_MIN_DISTANCE_PRIORITY_LIMIT = 5
+HS_MIN_DISTANCE_PRIORITY_LIMIT = 4
 
 # These two options define which types of planets are counted when determining the number of planets in the near
 # vicinity of a home system. HS_ACCEPTABLE_PLANET_SIZES is actually only needed for the process of adding planets
 # to the near vicinity of a home system in case that's needed to meet the limit.
-HS_ACCEPTABLE_PLANET_TYPES = planet_types_real + (fo.planetType.asteroids,)
-HS_ACCEPTABLE_PLANET_SIZES = planet_sizes_real + (fo.planetSize.asteroids,)
+HS_ACCEPTABLE_PLANET_TYPES = planet_types_real + (fo.planetType.asteroids,fo.planetType.gasGiant,)
+HS_ACCEPTABLE_PLANET_SIZES = planet_sizes_real + (fo.planetSize.asteroids,fo.planetSize.gasGiant,)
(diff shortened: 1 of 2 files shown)


Oberlus:
@Vezzra I think you authored these algorithms. It would be nice if you can do a quick review (dunno how to ask for reviewers).

default/python/universe_generation/options.py
```diff
@@ -33,29 +33,30 @@
 # The following two options are used to determine the minimum number of systems and planets. This limit is
 # HS_MIN_SYSTEMS_IN_VICINITY systems and HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM planets per system within the near
 # vicinity of a home system, capped at HS_MIN_PLANETS_IN_VICINITY_TOTAL
-HS_MIN_SYSTEMS_IN_VICINITY = 8
-HS_MIN_PLANETS_IN_VICINITY_TOTAL = 10
-HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM = 1
+# HS_MIN_SYSTEMS_IN_VICINITY = 12
+# HS_MIN_PLANETS_IN_VICINITY_TOTAL = 12
+# HS_MIN_PLANETS_IN_VICINITY_PER_SYSTEM = 1
```
Vezzra: If those constants are obsolete with the updated algorithm, just remove them instead of commenting them out. This just leads to commented out stuff accumulating in the code base. These values are already retained by the revision system.

default/python/universe_generation/options.py
```diff
@@ -33,29 +33,30 @@
...
-HS_MIN_DISTANCE_PRIORITY_LIMIT = 5
+HS_MIN_DISTANCE_PRIORITY_LIMIT = 4
 
 # These two options define which types of planets are counted when determining the number of planets in the near
 # vicinity of a home system. HS_ACCEPTABLE_PLANET_SIZES is actually only needed for the process of adding planets
 # to the near vicinity of a home system in case that's needed to meet the limit.
-HS_ACCEPTABLE_PLANET_TYPES = planet_types_real + (fo.planetType.asteroids,)
-HS_ACCEPTABLE_PLANET_SIZES = planet_sizes_real + (fo.planetSize.asteroids,)
+HS_ACCEPTABLE_PLANET_TYPES = planet_types_real + (fo.planetType.asteroids, fo.planetType.gasGiant,)
+HS_ACCEPTABLE_PLANET_SIZES = planet_sizes_real + (fo.planetSize.asteroids, fo.planetSize.gasGiant,)
```
Vezzra, line 55: This is potentially very problematic. Although I understand why you want to include them, so that players starting with Sly also get covered, this will lead to issues for everyone else.

Gas giants have been excluded for a reason, because they are very common, and if you count them as eligible vicinity planets for the evaluation of starting positions, all empires except those starting with Sly will most likely end up with far worse starting conditions than before.

When the original algorith has been implemented, gas giants haven't been valid colonization targets for any race. That changed with the introduction of Sly. Not including them may put Sly on a disadvantage, however, considering that gas giants are more common than other planets types, I think that is a smaller issue than counting gas giants as valid planet types for all other empires, for which they are useless (for colonization).

So I'd remove this particular change.

The proper fix of course would be to consider the species when calculating the merit of a starting position, but that will require a complete rewrite of the homeworld selection algorithm. The current algorithm can't do that at all, because the way it is designed it just chooses a certain requested number of homeworlds.

Assignment from that pool to specific empires happens in a subsequent step, and only then the information what starting species the empire (that gets a homeworld assigned) has is available.

Which means any algorithm that wants to ensure that not only a proper amount of (theoretically) colonizable planets is present near a starting position, but also a proper amount of the right *type* has to be a complete rewrite of the current one, as it needs to consider the starting species. And it will most likely have to be far more complex...

Vezzra (changes requested):
At a first glance most of the changes seem reasonable, AFAICT (there is one change I'm very wary of, see my comment at the respective code section). You have to take my assessment with a grain of salt however.

Because, although I am the one who wrote the algorithm originally, it has been tweaked and optimized by more competent coders subsequently, as my implementation hadn't been very efficient (on slower machines it caused the time it took for larger maps to generate increase exponentially).

Which means I can't really assess how your changes will actually impact starting position selection. I propose sufficient testing to see if those changes get you the results you desire.

default/python/universe_generation/options.py
Oberlus: > This is potentially very problematic.

In some games I started as Sly, the number of GG at less than 6 hops from my HW was very low (1, 2). For Sly, anything but GG are hostile, which means that not having enough GG around is the most disgraced possible situation (compared to any other species). If a Sly starts with few planets around plus few or no GG, and the algorithm fills the missing planets with anything but GG, the Sly is irremediably doomed. This is very problematic, without the "potentially", and thus deserves some attention.

Although GG are usually more common than any other planetary environment (that is, GG is in general more frequent than Radiated, Terran, etc.), there will be always many more regular planets than GG (i.e. the sum of the nine regular environments is way more frequent than GG), which means that on average regular species will always have more available good+adequate planets than Sly. Except for Narrow Tolerance species, this means than on average regular species always have more early-game colonizable opportunities than Sly.

I assume that the function to add extra planets uses a similar environment frequency than the initial universe generation function, an so the average distribution of planets will stay as more favorable for non-Sly species even if we add GG as an option in that function.

What is worse here? A regular species with few planets around that gets some extra GG (instead of more desirable good/adequate planets) that can be used as a source of extra production with the GG Generator? Or a Sly empire with no good/adequate planets around that get more hostile planets and never a good one?
My impression is that not allowing GG in the function to add extra planets is potentially worse than allowing GG.

> Which means any algorithm that wants to ensure that not only a proper amount of (theoretically) colonizable planets is present near a starting position, but also a proper amount of the right type has to be a complete rewrite of the current one, as it needs to consider the starting species. And it will most likely have to be far more complex...

Totally agree. My initial idea was to reorder the different function in universe generation to accomplish exactly what you say here. The problem is my python skills are close to NULL. This is the first time I do any python programming in my life, so I have to relly on the source code I already have. And so I discarded the more complex solution. However, the problem about unfair starting positions is a already an urging and serious problem. We have players that concede early on in the multiplayer games because their starting positions makes impossible for them to win (and that gives a great advantage for the nearby empires, in detriment of the farther away empires, so the problem cascades), and other players that remain in the game in an effort to not disappoint other players or abandon their team mates, but they stay in a rather discouraged mood which is also bad for everyone else. Some players have stated their decision to not play any more multiplayer games until this is solved. That is serious. Giving an incomplete solution to that issue is arguably better than not giving any solution at all, as long as the partial solution is not making things worse for other cases (but I am rather confident this is not the case).

Summing up, I believe letting that change to stay is better than reverting it. However, I do want your blessing for this change (after you consider the above reasoning), or I'll revert it. At least, the changes are able to improve the situation for most playable species, which is already good to have.

Vezzra: I guess you have to run several test games to see how the actual results are, and how much the inclusion of gas giants here affect the presence of usable planets for non-Sly species.

All we can do without testing is trying to make more or less educated guesses, which is probably not sufficient... 😉 

I won't strictly object to keeping your change, if it doesn't work it can always be reverted anyway. But you should keep an eye on it.

Vezzra: Otherwise, agreed, we definitely need to revise the placement algorithm to consider species. With the inclusion of species like Sly that became necessary.

Oberlus: I'm working on getting it more tested.
Up to now I've started ~10 games with supertesters to see whole galaxy at turn 3 and see all starting positions. Didn't find a "I have many less planets than you" situation yet, but I found several "I have many less good/adequate planets than you". I'll keep an eye on the "I got mostly GG and I'm not Sly, FML" situation, which haven't appeared yet.

Oberlus:
This is not ready to merge, more a WIP. Current deficiencies:

- The changes force to include too many planets, which in turn makes it unlikely to have empty systems.
- The lack of consideration of environment tolerance of the species makes it a clearly insufficient solution for issue #2900, even with PR #2916.

adrianbroher:
> This is not ready to merge, more a WIP.

Then mark it as draft.

Vezzra:
@adrianbroher,

> Then mark it as draft.

Correct me if I'm wrong, but with the introduction of this "draft" feature and us starting to use it, the "status:work in progress" label is now redundant and could/should be dropped, right?

adrianbroher:
@Vezzra I don't know if you can properly filter for drafts, but I would rather connect those two with a workflow that automatically labels/unlabels the PR according to state.

agrrr3:
I'm not sure what the official definition for GitHub drafts is, but for me
it is not the same as work in progress.

I also use it for prototypes which are never intended to be merged while
work in progress means if finished it should be merged

It definitely has the meaning: do not merge now

I also did not try to switch from a draft to a regular pr if that is
possible

Vezzra <notifications@github.com> schrieb am So., 3. Mai 2020, 11:53:

Vezzra:
Well, I'd consider the text github displays for drafts as the "official" definition: "This pull request is still a work in progress". And the actual functionality, which is that you cannot merge a PR as long as it is set to "draft".

So this is pretty much exactly what I think our "work in progress" label is supposed to be (sans the preventing merge functionality).

Using that label for prototypes has made sense for the reason you cited, as it means "do not merge". However, it might make sense to introduce a separate label ("status:prototype") to mark PRs which are supposed to be only that, prototypes.

Or a more general "status:testing only" label for all PRs which are only supposed to experiment with/test things, and are not intended to be ever merged.

geoffthemedio:
@Oberlus any progress?

Oberlus:
> any progress?

Not in this PR. My lack of knowledge of python and the codebase is getting in the way, but also the complexity of the subject. Current state:

- By just increasing HS_MAX_JUMP_DISTANCE_LIMIT there is an improvement in homeworld allocation (appear further away when possible, which is not often, see [here](https://www.freeorion.org/forum/viewtopic.php?f=2&t=11711&start=60#p102510)), but that doesn't tackle the planet environment distribution: https://github.com/freeorion/freeorion/blob/f7795ac1b35ac24e71909b8ad72b3b06efa6481f/default/python/universe_generation/options.py#L42

Mi intention is to write several functions to:
- Associate each non-HW system to the nearest (jump distance) HW system and label as frontier the ones that are at the same distance from more than one HW to compose the vicinity of each HW.
- Compute the number of planets of each environment on each HW's vicinity to compute some sort of score.
- Add planets to vicinities with too few planets and remove from vicinities with too many.
- Change environments of planets in vicinities with too many or two few good/adeq planets for its associated HW.

No prospects of finishing this any time soon. I'm learning python at a slow pace.

o01eg:
> 
> 
>     * Associate each non-HW system to the nearest (jump distance) HW system and label as frontier the ones that are at the same distance from more than one HW to compose the vicinity of each HW.

@Oberlus I wrote this function already to place teams: 

https://github.com/freeorion/freeorion/blob/1d197e6bec90b3e724a66fb52131c7dc761ff5db/default/python/universe_generation/empires.py#L550

PR #2919, opened 2020-05-02
Boolify Condition invariants

adrianbroher:
This PR:

- Replaces the `Condition::Invariance` enum with a boolean flag. `Condition`s are now variant unless configured otherwise.
- Moves all `Condition::Condition` subclass constructors into the `universe/Conditions.cpp` implementation file.
- Removes unused `Condition::Condition` subclass constructors.
- Uses delegated constructors on `Condition::Condition` subclasses, where an appropriate target constructor already existed.
- Lets the `Condition::Condition` invariance getters return the corresponding attribute instead of being overwritten by subclasses.
- Lets the `Condition::Condition` subclass constructors assign their invariances instead of quering it on evaluation time.
- Prevents the `Condition::Condition` invariance getters from being overwritten by subclasses.

### DIFF
--- a/universe/Condition.h
+++ b/universe/Condition.h
@@ -55,7 +55,7 @@ struct FO_COMMON_API Condition {
     //! condition is a subcondition to another Condition or a ValueRef, this
     //! condition may be evaluated once and its result used to match all local
     //! candidates to that condition.
-    virtual bool RootCandidateInvariant() const
+    bool RootCandidateInvariant() const
     { return m_root_candidate_invariant; }
 
     //! (Almost) all conditions are varying with local candidates; this is the
@@ -67,12 +67,12 @@ struct FO_COMMON_API Condition {
 
     //! Returns true iff this condition's evaluation does not reference the
     //! target object.
-    virtual bool TargetInvariant() const
+    bool TargetInvariant() const
     { return m_target_invariant; }
 
     //! Returns true iff this condition's evaluation does not reference the
     //! source object.
-    virtual bool SourceInvariant() const
+    bool SourceInvariant() const
     { return m_source_invariant; }
 
     virtual std::string Description(bool negated = false) const = 0;
--- a/universe/Condition.h
+++ b/universe/Condition.h
@@ -16,12 +16,6 @@ namespace Condition {
 
 typedef std::vector<std::shared_ptr<const UniverseObject>> ObjectSet;
 
-enum Invariance : int {
-    UNKNOWN_INVARIANCE, ///< This condition hasn't yet calculated this invariance type
-    INVARIANT,          ///< This condition is invariant to a particular type of object change
-    VARIANT             ///< This condition's result depends on the state of a particular object
-};
-
 enum SearchDomain : int {
     NON_MATCHES,    ///< The Condition will only examine items in the non matches set; those that match the Condition will be inserted into the matches set.
     MATCHES         ///< The Condition will only examine items in the matches set; those that do not match the Condition will be inserted into the nonmatches set.
@@ -88,9 +82,9 @@ struct FO_COMMON_API Condition {
     { return 0; }
 
 protected:
-    Invariance m_root_candidate_invariant = UNKNOWN_INVARIANCE;
-    Invariance m_target_invariant = UNKNOWN_INVARIANCE;
-    Invariance m_source_invariant = UNKNOWN_INVARIANCE;
+    bool m_root_candidate_invariant = false;
+    bool m_target_invariant = false;
+    bool m_source_invariant = false;
 
 private:
     struct MatchHelper;
--- a/universe/Condition.h
+++ b/universe/Condition.h
@@ -50,30 +50,30 @@ struct FO_COMMON_API Condition {
     virtual void GetDefaultInitialCandidateObjects(const ScriptingContext& parent_context,
                                                    ObjectSet& condition_non_targets) const;
 
-    /** Returns true iff this condition's evaluation does not reference
-      * the RootCandidate objects.  This requirement ensures that if this
-      * condition is a subcondition to another Condition or a ValueRef, this
-      * condition may be evaluated once and its result used to match all local
-      * candidates to that condition. */
+    //! Returns true iff this condition's evaluation does not reference
+    //! the RootCandidate objects.  This requirement ensures that if this
+    //! condition is a subcondition to another Condition or a ValueRef, this
+    //! condition may be evaluated once and its result used to match all local
+    //! candidates to that condition.
     virtual bool RootCandidateInvariant() const
-    { return false; }
+    { return m_root_candidate_invariant; }
 
-    /** (Almost) all conditions are varying with local candidates; this is the
-      * point of evaluating a condition.  This funciton is provided for
-      * consistency with ValueRef, which may not depend on the local candidiate
-      * of an enclosing condition. */
+    //! (Almost) all conditions are varying with local candidates; this is the
+    //! point of evaluating a condition.  This funciton is provided for
+    //! consistency with ValueRef, which may not depend on the local candidiate
+    //! of an enclosing condition.
     bool LocalCandidateInvariant() const
     { return false; }
 
-    /** Returns true iff this condition's evaluation does not reference the
-      * target object.*/
+    //! Returns true iff this condition's evaluation does not reference the
+    //! target object.
     virtual bool TargetInvariant() const
-    { return false; }
+    { return m_target_invariant; }
 
-    /** Returns true iff this condition's evaluation does not reference the
-      * source object.*/
+    //! Returns true iff this condition's evaluation does not reference the
+    //! source object.
     virtual bool SourceInvariant() const
-    { return false; }
+    { return m_source_invariant; }
 
     virtual std::string Description(bool negated = false) const = 0;
     virtual std::string Dump(unsigned short ntabs = 0) const = 0;
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -2737,6 +2737,11 @@ unsigned int CreatedOnTurn::GetCheckSum() const {
 ///////////////////////////////////////////////////////////
 // Contains                                              //
 ///////////////////////////////////////////////////////////
+Contains::Contains(std::unique_ptr<Condition>&& condition) :
+    Condition(),
+    m_condition(std::move(condition))
+{}
+
 bool Contains::operator==(const Condition& rhs) const {
     if (this == &rhs)
         return true;
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -6701,7 +6701,12 @@ OwnerHasShipDesignAvailable::OwnerHasShipDesignAvailable(
     Condition(),
     m_id(std::move(design_id)),
     m_empire_id(std::move(empire_id))
-{}
+{
+    auto operands = {m_id.get(), m_empire_id.get()};
+    m_root_candidate_invariant = boost::algorithm::all_of(operands, [](auto& e){ return !e || e->RootCandidateInvariant(); });
+    m_target_invariant = boost::algorithm::all_of(operands, [](auto& e){ return !e || e->TargetInvariant(); });
+    m_source_invariant = boost::algorithm::all_of(operands, [](auto& e){ return !e || e->SourceInvariant(); });
+}
 
 OwnerHasShipDesignAvailable::OwnerHasShipDesignAvailable(int design_id) :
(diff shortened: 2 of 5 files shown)


universe/Conditions.cpp
```diff
@@ -26,6 +26,7 @@
 #include "../Empire/EmpireManager.h"
 #include "../Empire/Supply.h"
 
+#include <boost/algorithm/cxx11/all_of.hpp>
```
geoffthemedio, line 29: why boost not std?

adrianbroher: It uses the range style interface instead of a begin, end pair.

adrianbroher:
I forgot to add the serialization of the invariance state.

universe/Conditions.cpp
```diff
@@ -640,7 +633,21 @@ SortedNumberOf::SortedNumberOf(std::unique_ptr<ValueRef::ValueRef<int>>&& number
     m_sort_key(std::move(sort_key_ref)),
     m_sorting_method(sorting_method),
     m_condition(std::move(condition))
-{}
+{
+    m_root_candidate_invariant =
+        (!m_number || m_number->SourceInvariant()) &&
```
geoffthemedio: Should probably be calling `RootCandidateInvariant` not `SourceInvariant` here. (Not original in your changes, but should be fixed)

universe/Conditions.cpp
```diff
@@ -640,7 +633,21 @@ SortedNumberOf::SortedNumberOf(std::unique_ptr<ValueRef::ValueRef<int>>&& number
...
+        (!m_sort_key || m_sort_key->SourceInvariant()) &&
+        (!m_condition || m_condition->SourceInvariant());
+    m_target_invariant =
+        (!m_number || m_number->SourceInvariant()) &&
```
geoffthemedio: Same here; should call `TargetInvariant`

universe/Conditions.cpp
```diff
@@ -1041,6 +1033,14 @@ unsigned int SortedNumberOf::GetCheckSum() const {
 ///////////////////////////////////////////////////////////
 // All                                                   //
 ///////////////////////////////////////////////////////////
+All::All() :
+    Condition()
+{
+    m_root_candidate_invariant = true;
```
geoffthemedio, line 1039: Why `true`? Previous implementation would default to base class `false`.

universe/Conditions.cpp
```diff
@@ -1078,6 +1078,14 @@ unsigned int All::GetCheckSum() const {
 ///////////////////////////////////////////////////////////
 // None                                                  //
 ///////////////////////////////////////////////////////////
+None::None() :
+    Condition()
+{
+    m_root_candidate_invariant = true;
```
geoffthemedio, line 1084: same... should be all `false` I think? Doesn't matter what the root candidate, source, or target are; nothing will be matched.

universe/Conditions.cpp
```diff
@@ -1358,6 +1359,13 @@ unsigned int EmpireAffiliation::GetCheckSum() const {
 ///////////////////////////////////////////////////////////
 // Source                                                //
 ///////////////////////////////////////////////////////////
+Source::Source() :
+    Condition()
+{
+    m_root_candidate_invariant = true;
```
geoffthemedio, line 1361: inconsistent with later cases where all three are explicitly set

universe/Conditions.cpp
```diff
@@ -1477,7 +1501,11 @@ Homeworld::Homeworld() :
 Homeworld::Homeworld(std::vector<std::unique_ptr<ValueRef::ValueRef<std::string>>>&& names) :
     Condition(),
     m_names(std::move(names))
-{}
+{
+    m_root_candidate_invariant = boost::algorithm::all_of(m_names, [](auto& e){ return e->RootCandidateInvariant(); });
```
geoffthemedio, line 1502: can't use `SetConditionVariance` with `m_names`?

universe/Conditions.cpp
```diff
@@ -2216,7 +2221,13 @@ HasSpecial::HasSpecial(std::unique_ptr<ValueRef::ValueRef<std::string>>&& name,
     m_name(std::move(name)),
     m_since_turn_low(std::move(since_turn_low)),
     m_since_turn_high(std::move(since_turn_high))
-{}
+{
+    auto operands = {m_since_turn_low.get(), m_since_turn_high.get()};
+    SetConditionVariance(operands, m_root_candidate_invariant, m_target_invariant, m_source_invariant);
+    m_root_candidate_invariant = m_root_candidate_invariant || (!m_name || m_name->RootCandidateInvariant());
```
geoffthemedio: Could perhaps add a non-templated `ValueRef` base class that has the `*Invariant` functions so that all could be put into the same range here?

Or even a common untemplated base class for `Condition` and `ValueRef` for the same reason?

adrianbroher: > Or even a common untemplated base class for Condition and ValueRef for the same reason?

Sounds reasonable, but I would say this is out of scope for this PR.

universe/Conditions.cpp
```diff
@@ -2586,7 +2570,10 @@ CreatedOnTurn::CreatedOnTurn(std::unique_ptr<ValueRef::ValueRef<int>>&& low,
     Condition(),
     m_low(std::move(low)),
     m_high(std::move(high))
-{}
+{
+    auto operands ={m_low.get(), m_high.get()};
```
geoffthemedio: space after = perhaps?

universe/Conditions.cpp
adrianbroher: This class was updated before I templated the `SetConditionVariance` function.

universe/Conditions.cpp
adrianbroher: The `ValueRef`s need to be updated before that too.

universe/Conditions.h
```diff
@@ -1169,17 +1025,10 @@ struct FO_COMMON_API EmpireStockpileValue final : public Condition {
     EmpireStockpileValue(ResourceType stockpile,
                          std::unique_ptr<ValueRef::ValueRef<double>>&& low,
                          std::unique_ptr<ValueRef::ValueRef<double>>&& high);
-    EmpireStockpileValue(std::unique_ptr<ValueRef::ValueRef<int>>&& empire_id,
```
geoffthemedio: should be implemented, not removed...

universe/Conditions.cpp
adrianbroher: I replaced `SetConditionVariance` with `boost::algorithm::all_of` calls.  When the new base class emerges the interface of `SetConditionVariance` will be unusable anyway, so better ditch it right now.

universe/Conditions.h
adrianbroher: Why? It is not called anywhere.

adrianbroher:
Updated the PR according to review remarks, left the discussions without consensus open.

universe/Conditions.h
geoffthemedio: It should be implemented and called from the parser

universe/Condition.h
```diff
@@ -88,9 +82,10 @@ struct FO_COMMON_API Condition {
     { return 0; }
 
 protected:
-    Invariance m_root_candidate_invariant = UNKNOWN_INVARIANCE;
-    Invariance m_target_invariant = UNKNOWN_INVARIANCE;
-    Invariance m_source_invariant = UNKNOWN_INVARIANCE;
+    bool m_root_candidate_invariant = false;
+    bool m_local_candidate_invariant = false;
```
geoffthemedio: What's the purpose of `m_local_candidate_invariant` ? I think it will always be false...

geoffthemedio:
otherwise seems OK

universe/Condition.h
adrianbroher: Some `Condition` subclasses calculate the local candidate invariance in their `Condition::Eval` method. Calculating the local candidate invariance should also be done in the constructor, which requires the serialization of the invariances.  As I am already introducing serialization for the other invariances I didn't want to break the serialization twice in short succession.

universe/Conditions.h
adrianbroher: Maybe, I don't know. The problem is that this is not part of the scope for this PR. I can undo the deletion commit.

universe/Condition.h
geoffthemedio: You mean the `simple_eval_safe` tests based on subconditions and valuerefs?

That's not the same as a the whole condition being local candidate invariant, and probably doesn't need to be exposed with a getter (or at least one named just `LocalCandidateInvariant`).

A condition or valueref being _source_ invariant means that whether a local candidate is matched doesn't depend on what the source object is.

Only a few conditions, like perhaps `All` or `None`, might qualify as _local candidate_ invariant, but I'd have to think a bit about whether that actually makes sense... It might depend on how condition local candidate invariance would be used.... which it presently isn't (in the sense that source or target invariance are by calling code).

Regardless, even if a condition's valuerefs and subconditions are all local candidate invariant, that doesn't mean the condition as a whole is. A condition like `Armed` has no internal state, but still matches some local candidates and not others depending on the details of the local candidate, so is not local candidate invariant.

Assuming the simple eval safe test is what you mean, it also probably shouldn't be part of the base class, as its an implementation detail of some of the derived classes, but not needed for others. It might also be replaced on a case-by-case basis with something more complicated... eg. there could be conditions with multiple subconditions or valuerefs, some of which are local candidate invariant and some which aren't, and the `Eval` function could re-evaluate the variant ones but not the invariant ones, rather than summarizing all of them as invariant or not.

That said, adding `m_simple_eval_safe` for caching, to conditions that use it, would make sense. That needs to be done carefully though, as you still have to test `(parent_context.condition_root_candidate || RootCandidateInvariant())` in `Eval` each time, as that depends on the input context, not just the condition itself... So maybe name it something else... I suppose it could be `m_valuerefs_and_subconditions_local_candidate_invariant` or somesuch...

adrianbroher: > You mean the simple_eval_safe tests based on subconditions and valuerefs?

Yes, I was talking about those.

The problem I have is that the implementation is overcomplicated.  You cannot understand 10.000 + 5.000 lines of code.

geoffthemedio: It is a huge complicated system. But it's not as bad as it might seem, as there are a lot of just-slightly-different functions and code between the various classes.

If you want to get into slightly more in-depth refactoring, it might make sense to combine some conditions that have similar interfaces into a single multi-purpose condition, that uses an enum to control how it functions.

For example, I think `RootCandidate`, `Target`, `Capital`, `Monster`, `Armed`, `Stationary`, `Aggressive`, `CanColonzie`, `CanProduceShips`, `All`, `None`, and `Source` don't have any internal state and don't take any parameters when constructed, so don't really need to be separate classes...

Sort like how `ValueRef` classes work, the `Eval` and `Match` functions would check the enum value to decide how to evaluate their input, and might get fairly big. But the `Description` or `Dump` would just refer to some members storing what strings to use, which would set during construction. Some external code that relies on casting pointers would need to further inspect the condition, like in the objects list window, as well.

universe/Conditions.h
adrianbroher: Removed corrsponding commit.

universe/Condition.h
adrianbroher: Removed `Condition::Condition::m_local_candidate_invariant` attribute again.

adrianbroher: > If you want to get into slightly more in-depth refactoring, it might make sense to combine some conditions that have similar interfaces into a single multi-purpose condition, that uses an enum to control how it functions.

Not now, this PR was mind-numbing enough for now.

> For example, I think RootCandidate, Target, Capital, Monster, Armed, Stationary, Aggressive, CanColonzie, CanProduceShips, All, None, and Source don't have any internal state and don't take any parameters when constructed, so don't really need to be separate classes...

> Sort like how ValueRef classes work, the Eval and Match functions would check the enum value to decide how to evaluate their input, and might get fairly big. But the Description or Dump would just refer to some members storing what strings to use, which would set during construction. Some external code that relies on casting pointers would need to further inspect the condition, like in the objects list window, as well.

Sounds good and matches what I had in mind but I would prefer to use member function pointers or functors for this instead of enums.  Trying to mimicking the interface of boost::python function bindings and removing most of the Condition/ValueRef implementations.

Also there are various implementations like `Building`, `Planet{Type,Size}` or Species, which are pretty much identical, comparing an entity property against a list of values.  Things like that are certainly not neccessary and can be replaced within the parser by binding the matching member function/functor to a more generalized `InList` Condition.

PR #2920, opened 2020-05-03
Improve chat debugging

Cjkjvfnby:
It is a reference PR for architecture approaches.

Let's discuss the overall structure first and then drill down to implementation.

I decided to refactor this part, because it is not critical, easy to test, and can wait until merging without conflicts for a longer time (It is release time now).

In total about +400 lines of code (190 of them are tests).

The idea behind these changes is to make code simpler, maintainable, testable, reusable.
 
- Hide implementation from the AI:
  It has two functions setup debug and handle the message.  It makes AI code simpler.
- Split single debug function to the separate classes  use composition to make them work together:
  Composition helps us to follow the single responsibility principle. On class responsible for getting info from the `fo`,  one class is responsible for configuring debug shell, one class responsible for sending messages to the chat.
- Dependency injection:
  Instead of hardcoding dependencies in the code, I used "dependency injection" in the simplest possible way: inject them via a constructor.
  This simplifies managing dependencies and refactoring.  Also, it allows us to write tests in simple way:  no need to use Mocks or monkeypatching. Just inherit from some abstract classes and inject them to the constructor. Tests are as simple as they can be. We still cannot auto-test the implementation of the injected classes for gameplay, but they are almost logicless.
- Abstract classes:
  They give us a runtime check that all methods are implemented. Push us to write docs and type annotation. You cannot say "No need to explain, it is oblivious form the code."
- Dependency management:
 `chat_tools` is an absolutely independent module from all other code.  For that exact module it does not provide any special bonuses,  but it leads to better code reusability(use it anywhere), testability(you can write test), and maintainability(changes in AI code does not affect this module).

Using this approach will allow us to write more structured and maintainable code. And increase or test coverage.

Commit message

```
Refactor debug chat and add tests

* split monolithic code to small chunks
* shell formatting updated
```


adrianbroher:
Is this PR only about the application of programming patterns for Python?

Cjkjvfnby:
> Is this PR only about the application of programming patterns for Python?

Yes. 

I also fixed annoying things like printing where location file is present and debug message formatting.

geoffthemedio:
Do you need some feedback or a review?

Cjkjvfnby:
> Do you need some feedback or a review?

I have already requested a review from @Morlic-fo.

Morlic-fo:
Yeah, sorry. I am currently swamped by RL stuff. I will try to get to this as soon as I find some spare time.

Cjkjvfnby:
> Yeah, sorry. I am currently swamped by RL stuff. I will try to get to this as soon as I find some spare time.

I don't expect any conflicts, feel free to keep this PR at the bottom of your review list.

geoffthemedio:
@Morlic-fo reminder in case you have some time

Morlic-fo:
> The idea behind these changes is to make code simpler, maintainable, testable, reusable.

I do not agree that patterns like dependency injection, class composition and abstract classes achieve any of those except testability in this case.

All these patterns inherently add one or even more level of indirection, require more LOC both initially and for trivial changes, reduce readability of the code and make life hard(er) for IDEs.

If all that is ever going to be used is a single configuration, then there is no reason to bloat the code by writing some abstract base class and dependency injection into your user classes. Just define the configuration and use it directly. 

Now, if there is reasonable expectation that in the future you will have multiple configurations which will have to be maintained and used with the same class or that you must provide some user of some library that ability, then sure. Use all these patterns.

Otherwise, this is pretty much a case of premature refactoring.


That being said, it does undoubtedly provide a simpler testing interface. I would argue that simpler testing alone is no argument to make your actual code base harder to maintain: Hide the complexity in the test code.

Cjkjvfnby:
> I do not agree that patterns like dependency injection,  ... achieve any of those except testability in this case.

To be on the same page with `dependency injection`

Passing an argument via constructor **is** `dependency injection`
Setting an argument via setter **is** `dependency injection`
Reading argument for the outer scope **is not** `dependency injection`.

> make code simpler

- With DI you have `in`, `out`.   
- Without DI you have `in`, `out`, and some variables in outer scope you depend on (a typical example is `fo`).

You need to keep in mind fewer items.  That is `simpler` for me.

> maintainable

If you can test it, it is more maintainable. If you change an item, you will change/test everything that depends on it.  But using Dependency inversion (via Dependency injection) you can control it. I will provide an example when talks about abstract classes.

> reusable

If you function depends on freeOrionAIInterface can you reuse it in turn_envent, universe generator, tests?  
If your function depends on variable from the outer scope, can you reuse it in a different part of the code?

To be continued...

Morlic-fo:
> * With DI you have `in`, `out`.
> 
> * Without DI you have `in`, `out`, and some variables in outer scope you depend on (a typical example is `fo`).

```some variables``` seems misleading. I agree that global/external variables are (mostly) bad and should be avoided.

Most of the times it is an external interface / function call.

Let's discuss a simple example
```
def add(a, b):
    return a + b

def add_1(a):
    return add(a, 1)

def add_1_with_DI(a, adder_function):
    return adder_function(a, 1)

add_1(10)
add_1_with_DI(10, add)
```

Would you say ```add_1_with_DI``` is "simpler" than add_1 because it does not rely on "some variables in outer scope"?

**Comment by Cjkjvfnby** [comment]
> some variables seems misleading. I agree that global/external variables are (mostly) bad and should be avoided.

+1

> Would you say add_1_with_DI is "simpler" than add_1 because it does not rely on "some variables in outer scope"?

Both cases are "simple". The first one is a little simpler.  DI does not have any benefits here in terms of `simplicity`.

DI is about crossing boundaries.  The module inside is a monolith. DI is a way to make monoliths work together without merging them into a single one.  

Monolithic vs. modular - where should we draw a line? It is the second programming question after the variable names. This PR is a part of my quest for knowledge.

**Comment by Cjkjvfnby** [comment]
> All these patterns inherently add one or even more level of indirection, require more LOC both initially and for trivial changes, reduce readability of the code and make life hard(er) for IDEs.

The idea of the interfaces is to break programs to chunks. You have more layers in total but you don't need to go beyond interfaces. For example, logging call some Freorion C++ code. But it does not matter for me, module documentation is the final destination. 

It does not matter ho many LOC in your program.  It matters how many LOC in code you need to read.  At that point it has the same effect as the extraction of the logic to separate methods, these methods just in different classes.

The interface should act as boundaries for the changes. So if abstraction is done right, trivial changes will affect a couple of modules.  If `fo` is changed, `DebugChatHandler` should not be affected, changes will be stopped at `DataProvider`.  And `DataProvider` itself is very simple class, it should not be hard to changes it. 

It affects readability in both ways.  More classes with better cohesion.  I feel like a fish in the water in a multimodule environment so for me it 's a benefit.

IDE life should be hard :)
Do you know about inheritance navigation?
image

PS. These answers help me to looks at my ideas from a different angle.

**Comment by Cjkjvfnby** [comment]
> Now, if there is reasonable expectation that in the future you will have multiple configurations which will have to be maintained and used with the same class or that you must provide some user of some library that ability, then sure. Use all these patterns.

Unit tests should be considered as first-class citizens.  So I already have 2 configurations.

If I will remove abstract classes, I will need to mock a lot of stuff in tests.
Both mocking and implementing abstract classes require some effort.  Which way is easier?

IMHO abstract classes are.  It is as trivial as an inheritance. It is an easy code, it is transparent. A little bulky, but it consists of simple blocks. We will have slightly more code in AI, and slightly more code in tests.  

We already have a global mock of freeOrionAIInterface. But if we need to have different output and not break other tests, we need to [monkeypath](https://docs.pytest.org/en/latest/monkeypatch.html) and [MagicMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock) them.  And we got two layers of mocks.  When tests have a lot of magic, nobody wants to write and support them.

Actually, it is an internal library.  It has a simple facade to use.  It has a documented integration procedure. And it has tests.

**Comment by geoffthemedio** [comment]
@Morlic-fo status of this...?

**Comment by Cjkjvfnby** [comment]
I have simplified it. 

I did not actally tested it, since I blocked by absence of the Windows build.

**Comment by geoffthemedio** [comment]
> I did not actally tested it

What needs testing?

**Comment by Cjkjvfnby** [comment]
> What needs testing?

I haven't run this code after my last changes. And if I don't run code, it will fail.  
I hope it will be a windows build soon.

This code only affects debug chat.

**Comment by Morlic-fo** [comment]
My apologies for not getting back to this earlier. 

While I am still not convinced that the refactoring is actually beneficial if we are looking at AI code only, I think this refactoring will be helpful when we want to eventually implement chat-based python debugging for the server as well.
So go ahead if you want to.

**Comment by geoffthemedio** [comment]
@Cjkjvfnby plans to get back to this? needs a rebase against mater (not merging master into it) as well...

**Comment by Cjkjvfnby** [comment]
> plans to get back to this?

Definitely.


> needs a rebase against mater (not merging master into it) as well...

I'd prefer squash and merge :)


> While I am still not convinced that the refactoring is actually beneficial 

It is not about chat itself, it is about the development approach.  How to organize code, how to manage dependencies, how to test.

And I managed to get state when I don't need FreeOrion to do changes. Just change code, run tests, and get instant feedback ( 15 passed in 0.06s).  This is really great feeling.

**Comment by geoffthemedio** [comment]
I could squash while merging, but it needs to be rebased before that can happen.

**Comment by Cjkjvfnby** [comment]
> I could squash while merging, but it needs to be rebased before that can happen.

If you squash, it does not matter, if I rebased or not,  it will be a single commit. Can you explain why rebase required?

PS. Do you know if this enabled in `settings/branch_protection_rules`? 
image

**Comment by geoffthemedio** [comment]
Whether it's a single commit isn't the issue (though I'd actually prefer to keep relevant history of commits if they are separable). The issue is that we can't trust you or anyone else to have properly done the merging, and previously it has cause issues when it wasn't done properly. So, the rule is that we don't merge branches that have merged master. This ensures nothing in the commit history could have been messed up.

**Comment by geoffthemedio** [comment]
commit message could include a bit of detail about what this does...

otherwise is it done?

**Comment by Cjkjvfnby** [comment]
> though I'd actually prefer to keep relevant history of commits if they are separable.

I try to keep one feature per PR, in that case, it is pretty Ok to have one commit. Usually first commit is a feature and all following are just fixes/cleanups.   


> The issue is that we can't trust you or anyone else to have properly done the merging, and previously it has cause issues when it wasn't done properly. So, the rule is that we don't merge branches that have merged master. This ensures nothing in the commit history could have been messed up.

If it is a fast-forward merge there should not be any caveats.  Changes you see are the changes that will be in a single commit. 

And no,  I am not rebasing my commits to master step by step, it is a waste of time.  I just do `git reset master` and `git add` my changes back.  So if some merges were done with errors, I don't track it during removing merge commits from my history.  I use self-review to avoid such cases.  I got rid of merges, but in the case of squash and merge it does not bring additional value. 

Clean history force pushed.

**Comment by Cjkjvfnby** [comment]
> commit message could include a bit of detail about what this does...

Good point. For my work and pet projects, I postpone this until the last moment.  When I do squash and merge I collect all commit messages and make a final one.

If this will be merged not by me, it is a trouble. 

I can add a message to the PR description, so it can be used for sqashuing, or merge it myself.

**Comment by geoffthemedio** [comment]
merging by yourself works

**Comment by Cjkjvfnby** [comment]
I got pretty nice coverage. 

```
Name                                                     Stmts   Miss  Cover
----------------------------------------------------------------------------
AI\freeorion_tools\chat_handler\__init__.py                  2      0   100%
AI\freeorion_tools\chat_handler\base_chat_handler.py         4      0   100%
AI\freeorion_tools\chat_handler\chat_formatter.py           25      4    84%
AI\freeorion_tools\chat_handler\debug_chat_handler.py       82      3    96%
AI\freeorion_tools\chat_handler\interpreter.py              36      0   100%
AI\freeorion_tools\chat_handler\normal_chat_handler.py       4      1    75%
AI\freeorion_tools\chat_handler\shell_variable.py            9      0   100%
----------------------------------------------------------------------------
TOTAL                                                      162      8    95%
```

PR #2921, opened 2020-05-03
Make Chato BAD_POPULATION

Oberlus:
Chato are OP.

Arguably, Scylior are also OP in the long run because they have good population and the same research/industry skills.
Scylior have bad supply, Chato have bad troops (offensive and defensive). This means Scylior expands slower but is better at conquering natives early on.
However Chato also have broad tolerance, with means more colonization opportunities early on, and Scylior start at a medium planet (less PP and RP), which means some slowdown early on. 

This PR tries to fix Chato advantage by giving them bad population.
There is no GOOD/GREAT_RESEARCH playable species that is also bad population, so this change does no make them too similar to any other species.
This also makes Chato somewhat the opposite of Egassem (except for the supply).


geoffthemedio:
Any AI Adjustments needed @freeorion/ai-team ?

Was there a forum discussion?

geoffthemedio:
?

Oberlus:
There was no forum discussion, only in the XMPP chatroom.
No idea if this needs AI adjustments.

Morlic-fo:
Should be fine for AI.

PR #2926, opened 2020-05-07
Const-ify Empire statistic getters

adrianbroher:
Class attributes should only be modified by the owning class.

### DIFF
--- a/universe/Universe.cpp
+++ b/universe/Universe.cpp
@@ -1641,34 +1641,6 @@ void Universe::ExecuteEffects(std::map<int, Effect::SourcesEffectsTargetsAndCaus
     }
 }
 
-namespace {
-    static const std::string EMPTY_STRING;
-
-    const std::string& GetSpeciesFromObject(std::shared_ptr<const UniverseObject> obj) {
-        switch (obj->ObjectType()) {
-        case OBJ_PLANET: {
-            auto obj_planet = std::static_pointer_cast<const Planet>(obj);
-            return obj_planet->SpeciesName();
-            break;
-        }
-        case OBJ_SHIP: {
-            auto obj_ship = std::static_pointer_cast<const Ship>(obj);
-            return obj_ship->SpeciesName();
-            break;
-        }
-        default:
-            return EMPTY_STRING;
-        }
-    }
-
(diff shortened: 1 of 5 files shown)


universe/Universe.cpp
```diff
@@ -1641,65 +1641,18 @@ void Universe::ExecuteEffects(std::map<int, Effect::SourcesEffectsTargetsAndCaus
...
-            source_empire->SpeciesShipsDestroyed()[species_for_obj]++;
-    }
-
-    if (Empire* obj_empire = GetEmpire(empire_for_obj_id)) {
-        if (!species_for_obj.empty())
-            obj_empire->SpeciesShipsLost()[species_for_obj]++;
-
-        if (design_for_obj_id != INVALID_DESIGN_ID)
-            obj_empire->ShipDesignsLost()[design_for_obj_id]++;
+    if (auto shp = std::static_pointer_cast<const Ship>(obj)) {
```
geoffthemedio: why static and not dynamic cast?

adrianbroher: That's a c&p error stemming from `GetSpeciesFromObject`.

universe/Universe.cpp
```diff
@@ -1641,65 +1641,18 @@ void Universe::ExecuteEffects(std::map<int, Effect::SourcesEffectsTargetsAndCaus
...
-
-    int GetDesignIDFromObject(std::shared_ptr<const UniverseObject> obj) {
-        if (obj->ObjectType() != OBJ_SHIP)
-            return INVALID_DESIGN_ID;
-        auto shp = std::static_pointer_cast<const Ship>(obj);
-        return shp->DesignID();
-    }
-}
-
 void Universe::CountDestructionInStats(int object_id, int source_object_id) {
```
geoffthemedio, line 1644: I think `CountDestructionInStats` is also called from `ExecuteEffects` for any object that is destroyed by an effect, so this might actually fix a bug where a planet destroyed by an effect could be counted as a ship of its species being destroyed

adrianbroher: When implementing this I was also a bit confused why destroyed planets would be recorded as destroyed ships but good to know that this is a bug.

PR #2927, opened 2020-05-07
ValueRef::Variable separate property dispatch

adrianbroher:
This PR separates the dispatching of object properties into binding (converting the property string into a callable binding and the actual dispatching of the call with the current evaluated object.

This is a preparation to move the binding step out of the `ValueRef::Variable::Eval` member function to avoid the string comparision, which happens currently during every `ValueRef::Variable` evaluation and to overall simplify the `ValueRef` implementations.

Contains a not yet merged PR, therefor blocked-by: #2926


adrianbroher:
Rebased onto master.

unblocked-by: #2926

PR #2928, opened 2020-05-07
ValueRef::ComplexVariable separate property dispatch

adrianbroher:
This PR separates the dispatching of object properties into binding (converting the property string into a callable binding and the actual dispatching of the call with the current evaluated object.

This is a preparation to move the binding step out of the ValueRef::ComplexVariable::Eval member function to avoid the string comparision, which happens currently during every ValueRef::ComplexVariable evaluation and to overall simplify the ValueRef implementations.

I have separated this from PR #2927 as this PR also contains some more complex unfolding of helper functions.  I would suggest to check every commit by itself or even create separate PRs for each commit.

Contains a not yet merged PR, therefor blocked-by: #2927 #2926

### DIFF
--- a/universe/ValueRefs.cpp
+++ b/universe/ValueRefs.cpp
@@ -28,6 +28,8 @@
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/split.hpp>
+#include <boost/range/adaptor/map.hpp>
+#include <boost/range/numeric.hpp>
 
 #include <functional>
 #include <iomanip>
@@ -1489,18 +1491,6 @@ namespace {
         return EMPTY_INT_FLOAT_MAP;
     }
 
-    int GetIntEmpirePropertyNoKeyImpl(int empire_id, const std::string& parsed_property_name) {
-        Empire* empire = GetEmpire(empire_id);
-        if (!empire)
-            return 0;
-
-        if (parsed_property_name == "OutpostsOwned")
-            return empire->OutpostsOwned();
-        // todo: add all the various OwnerWhatever ValueRef stuff here
-
-        return 0;
-    }
-
     // gets property for a particular map key string for one or all empires
     int GetIntEmpirePropertySingleKey(int empire_id, const std::string& parsed_property_name,
                                       const std::string& map_key)
@@ -1649,21 +1639,6 @@ namespace {
                 return design->ID();
         }
         return -1;
-    }
-
-    // gets unkeyed property for one or all empires
-    int GetIntEmpirePropertyNoKey(int empire_id, const std::string& parsed_property_name) {
-        int sum = 0;
-
-        if (empire_id != ALL_EMPIRES) {
-            // single empire
-            return GetIntEmpirePropertyNoKeyImpl(empire_id, parsed_property_name);
-        }
-
-        // all empires summed
-        for (const auto& empire_entry : Empires())
-            sum += GetIntEmpirePropertyNoKeyImpl(empire_entry.first, parsed_property_name);
-        return sum;
     }
 }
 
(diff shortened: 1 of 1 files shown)


adrianbroher:
Rebased onto #2927.

unblocked-by: #2926

geoffthemedio:
I think this needs a rebase...?

adrianbroher:
Rebased onto 931c5bf70ad4df4f16403911d6efeb82fcdf600b
unblocked-by: #2927

universe/ValueRefs.cpp
```diff
@@ -1409,340 +1412,128 @@ Visibility ComplexVariable<Visibility>::Eval(const ScriptingContext& context) co
...
-    {
-        if (boost::istarts_with(parsed_property_name, "ShipDesign")) {
-            // look up ship design id corresponding to specified predefined ship design name
-            const ShipDesign* design = GetPredefinedShipDesign(key_string);
-            if (design)
-                return design->ID();
-        }
-        return -1;
-    }
+        if (empire)
```
geoffthemedio, line 1493: should this be `!empire` ?

universe/ValueRefs.cpp
```diff
@@ -2015,26 +1775,64 @@ double ComplexVariable<double>::Eval(const ScriptingContext& context) const
...
-            return GetFloatEmpirePropertySingleKey(empire_id, variable_name, key_int);
-        }
+        std::function<bool (const std::map<int, float>::value_type&)> key_filter;
+        key_filter = [](auto k){ return true; };
+
+        if (m_int_ref2)
+            key_filter = [k = m_int_ref2->Eval(context)](auto e){ return k == e.first; };
+
+        if (empire)
+            return boost::accumulate(empire_property(*empire) | filtered(key_filter) | map_values, 0.0f);
```
geoffthemedio, line 1830: Is this clear and beneficial? Do you anticipate some extension in which something other than all-keys and just-one-key are used? If not this seems overly complicated and hard to quickly understand when using `accumulate`.

adrianbroher: What do you mean?  The general pattern like [those lines](https://github.com/freeorion/freeorion/pull/2928/files/17c050391dc6d63e6997b8b78a868dd0984aeaaa#diff-a3119b793e03ce2507143dbe73a6554dR1829-R1835) or this specific line?

> Is this clear and beneficial?

I would argure yes, this is clear and concise. Applying the same pattern for all properties would make it possible to have a generalized implementation, which only requires to bind the property on construction of the ValueRef while having the evaluation doesn't need those massive if else branches anymore.

> Do you anticipate some extension in which something other than all-keys and just-one-key are used? 

No, but both variants can be handled with the same statement whilst allowing composability whenever the need arises.

> If not this seems overly complicated and hard to quickly understand when using accumulate.

Why exactly?  `accumulate` is a standard library function and should be known by most people. It sums up the content of a container.  This container here are the extracted values of a map, with an optional applied filter by the map key.

geoffthemedio: The general pattern.

The issue I have with `accumulate` is not what it does internally, but rather that it is somewhat misleading when used here with the two cases of "all" and "just one". Accumulating with a filter to pick just one item in the map is not, to me, intuitive, and suggests there is something more complicated going at first glace. The previous implementation was, to me, clearer, in that it only did the loop and sum when all the entries were being summed. Looping over all the entries to test each instead of using the built-in map lookup seems strange (and is slightly inefficient, but likely not enough so to matter).

Presumably you could just do the accumulate step when summing all the values from the map, and skip it when lookup up just one?

The same pattern can be applied for all properties... I don't have any issue with the function object to do the lookup, which has a clear path to future optimization as you've described.

adrianbroher: > Accumulating with a filter to pick just one item in the map is not, to me, intuitive, and suggests there is something more complicated going at first glace.

Well, then I just have to disagree here.  I don't see what make the one empire case so special that it requires a different code path when it can be handled as a container with `size() == 1`. 

> Presumably you could just do the accumulate step when summing all the values from the map, and skip it when lookup up just one?

Sure I do it that way if it prevents merging this PR.

However your suggestion needs to be adopted for both the "all empires" and "that empire" cases so it's 4 branches for each property type.  I still believe this can be implemented in a less convoluted way.

geoffthemedio: will get back to this...

PR #2933, opened 2020-05-08
Make utility scripts  use Python3 and become PEP-8 compliant

adrianbroher:
This PR takes the various utilitiy scripts located across the source tree and makes them PEP-8 compliant according to the Flake8 configuration used in the CI linting job.

Those scripts are currently not covered by CI as the job only checks the files within the `default/python` directory.

### DIFF
--- a/cmake/make_versioncpp.py
+++ b/cmake/make_versioncpp.py
@@ -60,7 +60,7 @@ class Generator(object):
         try:
             with open(self.infile) as template_file:
                 template = Template(template_file.read())
-        except:
+        except IOError:
             print("WARNING: Can't access %s, %s not updated!" % (self.infile, self.outfile))
             return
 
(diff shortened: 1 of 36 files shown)


cmake/make_versioncpp.py
```diff
@@ -60,7 +60,7 @@ def execute(self, version, branch, build_no, build_sys):
         try:
             with open(self.infile) as template_file:
                 template = Template(template_file.read())
-        except:
+        except IOError:
```
Cjkjvfnby, line 63: Print error too:
```
as e: 
   print("...." % (...., e))

```

default/scripting/buildings/colonies/SP_ABADDONI.focs.txt
```diff
@@ -64,7 +64,7 @@ BuildingType
     ]
     effectsgroups = [
         [[LIFECYCLE_MANIP_POPULATION_EFFECTS("SP_ABADDONI")]]
-
```
Cjkjvfnby, line 67: This is not python script.

Cjkjvfnby (approved):
Minor comments.

default/scripting/buildings/colonies/SP_ABADDONI.focs.txt
adrianbroher: This was generated by a python script and the PEP-8 conformance directly impacts that generated output.

PR #2941, opened 2020-05-11
Remove keymaps configuration file

adrianbroher:
The current use case is already covered by configurable keyboard
shortcuts and the localization of keyboard layout provided by SDL2.

The functionality was not exposed via UI anyway and the configuration
used the last years didn't do any remapping.

@geoffthemedio you were the initial author of this code.  Can you
confirm this?


ChangeLog.md
```diff
@@ -7,6 +7,13 @@ Notable changes to the FreeOrion project will be documented in this file.
 [v0.4.9+] - TBA
 ---------------
 
+### Detailed Changelog
+
+#### Removed features
+
+- Removed the scripting/keymaps.inf configuration file, which allowed users to remap one keycode to another.
+  This functionality is already exposed in a differnt way via configurable keyboard shortcuts.
```
geoffthemedio: differnt -> different

geoffthemedio:
If you mean the in-game hotkeys, then I don't think those replicate the functionality of the keymap.

I also don't know how SDL2's localization works on what it can or can't do.

The motivation for the keymaps addition is discussed here: https://www.freeorion.org/forum/viewtopic.php?f=26&t=2348&p=54176&hilit=keymap&sid=3fd3891aac6d6cf08a71d183022f883a#p54176

If SDL2 automatically handles switching keyboard layouts in the OS, including Dvorak support, then that's probably sufficient.

adrianbroher:
So the purpose was to provide a manual keyboard layout mapping?

> I also don't know how SDL2's localization works on what it can or can't do.

SDL2 separates the physical location of a key (identified by the scancode, following the USB HID specification, more exactly the usage table 0x08) from the emitted character/label on the key (identified by the keycode).  The former is relevant for configuration serialization, the latter for representation in the UI.

> If SDL2 automatically handles switching keyboard layouts in the OS, including Dvorak support, then that's probably sufficient.

This is covered by the separation above and the `SDL_TEXTINPUT` event.

ChangeLog.md
adrianbroher: Rewritten to describe the original intent of the code.

PR #2959, opened 2020-05-17
Replace std::bind with boost::bind

geoffthemedio:
In many / most places. Boost version seems less crash prone.

### DIFF
--- a/GG/GG/Spin.h
+++ b/GG/GG/Spin.h
@@ -443,12 +443,9 @@ void Spin<T>::SetEditTextFromValue()
 template <typename T>
 void Spin<T>::ConnectSignals()
 {
-    m_edit->FocusUpdateSignal.connect(
-        std::bind(&Spin::ValueUpdated, this, std::placeholders::_1));
-    m_up_button->LeftClickedSignal.connect(
-        std::bind(&Spin::IncrImpl, this, true));
-    m_down_button->LeftClickedSignal.connect(
-        std::bind(&Spin::DecrImpl, this, true));
+    m_edit->FocusUpdateSignal.connect(boost::bind(&Spin::ValueUpdated, this, boost::placeholders::_1));
+    m_up_button->LeftClickedSignal.connect(boost::bind(&Spin::IncrImpl, this, true));
+    m_down_button->LeftClickedSignal.connect(boost::bind(&Spin::DecrImpl, this, true));
 }
 
 template <typename T>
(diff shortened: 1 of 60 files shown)


agrrr3:
If switching from boost::bind to std::bind is not so easy, can you c++-savvy guys suggest a way to go forward? Is the underlying problem understood? 

The internet recommends to use lambdas or [named function objects](http://www.stroustrup.com/C++11FAQ.html#lambda)  over std::bind

So I would imagine starting to move to lambdas bit by bit(?)

geoffthemedio:
> ... a way to go forward?

Just using `boost::bind` should be fine...? The original problem was that in newer versions of Boost, the definitions of the placeholders like `_1` were moved in such a way that they were ambiguous with the standard library equivalents. Fixing this doesn't require switching to `std::bind`.

> The internet recommends to use lambdas or [named function objects](http://www.stroustrup.com/C++11FAQ.html#lambda) over std::bind

The relevant uses of bind are not creating a new local function to implement a custom comparison operator or similar, but rather calling existing functions for particular objects in response to a signal. Equivalent lambdas could be written, but whether there is any benefit to doing so, I'm not sure...
`WindowResizedSignal.connect(boost::bind(&HumanClientApp::HandleWindowResize,this, _1, _2));`
vs
`WindowResizedSignal.connect([this](auto w, auto h) { this->HandleWindowResize(w, h); });`
Possibly the lambda version is better due to potential inlining?

adrianbroher:
> Is the underlying problem understood?

The problem is boost::signals2 automatic slot lifetime management.

We use boost::signals2 to notify one or more objects (the receivers) of another objects (emitter) state change (e.g. a button as pressed: play a sound and execute the associated button action).   Signals decoupled the emitter from the receivers by a slot (which can be any callable like a function, a bind, a functor, a lambda or another signal).  The problem arises whenever a emitter outlives one or more receivers.  As the emitter does not know that the slot became invalid and may call a function on an object which does not exist anymore.

boost::signals2 circumvent this by looking into the boost::bind parameters and checks if any of those are inherited from the boost::signals2::trackable class.  If that's true the signal registers the trackable as a weak_ptr  so it can disconnect the slot on object destruction.

This lookup is only implemented for boost::bind.

boost::signals2 slots can instead use the track or track_foreign member function to attach smart pointers for tracking an objects lifetime.


> but whether there is any benefit to doing so, I'm not sure...
> WindowResizedSignal.connect(boost::bind(&HumanClientApp::HandleWindowResize,this, _1, _2));

It's disgusting. It discards parameters without warning. Also the dispatch logic needs to be separated from the location of call.

> vs
> WindowResizedSignal.connect([this](auto w, auto h) { this->HandleWindowResize(w, h); });

It's neat and concise and the syntax is easy to understand.

PR #2932, opened 2020-05-07
Rebalance exobots: make them bad at everything but let them live anywhere

Voker57:
Fixed-up version of Oberlus's improvements on my patch #2931

New stats: 50% population
Bad supply
Bad industry
Bad research

But: adequate everywhere but asteroids (still poor)
Move up in research tree

PR #2961, opened 2020-05-18
exobots everywhere AI upgrade

agrrr3:
This makes MINDLESS_SPECIES tag effect from #2932  known to the AI.

Regarding AI i think classic AI is the default.
For classic AI approach maybe some ordering in default/python/AI/TechsListsAI.py  or some fast-tracking in ResearchAI should be done - needs game analysis/heuristics though.

New AI approach looks more sensible than the classic one, but I am not sure anyone actually works on it currently.

Somebody please tag this for the AI team.

### DIFF
--- a/default/python/AI/AIDependencies.py
+++ b/default/python/AI/AIDependencies.py
@@ -125,6 +125,9 @@ POP_MOD_LIGHTSENSITIVE_STAR_MAP = {fo.starType.blue: -2, fo.starType.white: -1}
 # the percentage of normal population that a species with Gaseous tag has
 GASEOUS_POP_FACTOR = 0.50
 
+# the percentage of normal population that a species with MINDLESS_SPECIES tag has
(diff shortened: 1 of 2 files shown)


Cjkjvfnby (approved):
Am I right that this is still correct?  https://github.com/freeorion/freeorion/blob/master/default/python/AI/ColonisationAI.py#L130

Oberlus:
> Am I right that this is still correct? https://github.com/freeorion/freeorion/blob/master/default/python/AI/ColonisationAI.py#L130

Yes (if I understand it).
These Exobots are adequate everywhere (poor asteroids), and never good.

Morlic-fo (approved):

Vezzra:
As this PR has been tagged as release-blocking, and been approved by all requested reviewers, please merge it before Sunday.

PR #2962, opened 2020-05-18
Set rule defaults to 4 bouts and rescale structure and damage

agrrr3:
Set defaults for 4-bouts scaled damage and structure (3-1-1-1 to 4-8-6-6) for #2901

So 4 bouts and following multipliers: 8 structure, 6 main weapon damage, and 6 fighter damage 

This is intended 0.4.10 post-release and should provide feedback.

### DIFF
--- a/combat/CombatEvents.cpp
+++ b/combat/CombatEvents.cpp
@@ -22,7 +22,7 @@ namespace {
     // to exist on client and server.
     void AddRules(GameRules& rules) {
         rules.Add<int>("RULE_NUM_COMBAT_ROUNDS", "RULE_NUM_COMBAT_ROUNDS_DESC",
-                       "", 3, true, RangedValidator<int>(2, 20));
+                       "", 4, true, RangedValidator<int>(2, 20));
         rules.Add<bool>("RULE_AGGRESSIVE_SHIPS_COMBAT_VISIBLE", "RULE_AGGRESSIVE_SHIPS_COMBAT_VISIBLE_DESC",
                         "", false, true);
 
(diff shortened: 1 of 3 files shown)


o01eg:
@agrrr3 I suppose changed values in stringtables should be calculated. It is possible players set other values for game rules.

Vezzra:
As the release branch has been created, this can be merged (if ready).

agrrr3:
@oberlus do you think we should add the structure scaling for multi-shot now (like done in this PR)? Or wait until feedback from the 11th game?

Oberlus:
> do you think we should add the structure scaling for multi-shot now (like done in this PR)? Or wait until feedback from the 11th game?

I'm not sure I understand.
If you mean the 8x structure scaling done in here:
https://github.com/freeorion/freeorion/blob/d112e5e6053b8292e0ca40157a73f620d2474ac0/universe/ShipHull.cpp#L17

Then yes, as it is necessary for the balance, but I don't think you mean that.
Could you explain what do you mean?

agrrr3:
> Could you explain what do you mean?

The scaling in this PR is 4-8-6-6 . A smaller change to accomodate four bouts is 4-4-3-3 . 

We were talking that it would be good for multishot if the current structure and weapon damage would be scaled up and with factors of 2 and 3 . 4-8-6-6 contains that scaling but of course we could also rescale at any point in time and do that later.

But as the difference is only factor 2 I would actually say let's do it now, players will need some time to adjust to the new values.

Oberlus:
Yes, I think 4-8-6-6 is the way to go. Also, testing it now will let us see if it causes some undesired effects in the GUI or whatnot.

agrrr3:
Ok, this can be merged into master then.

PR #2963, opened 2020-05-19
Rebalance flat bonuses, add one refinement level to NAI and AA

Oberlus:
This is a follow up to #2943 

Gas Giant Generator down to +5 and +3 (from +10 and +5)

Nascent AI down to +1 from +2.
New tech refinement Artificial Minds +3.

Adaptive Automation down to +2 from +5.
New tech refinement Sentient Automation +6.

AI updated accordingly to get the new tech refinements.
I could not find any place in the AI code to tell it the new values for these techs.
@agrrr3  @Cjkjvfnby @Morlic-fo please review this changes.

[Related thread](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11040&p=101798&sid=3f5e517fcafd674bbd677c1122909dc5#p101703)

PR #2943, opened 2020-05-12
Nerf pop-based bonuses (-50% to -80%) to reduce snowballing

Oberlus:
Early, small bonuses are halved.
Big bonuses are divided by five.
Some are in between.

Techs like Quantum Networking or Solar Orb. Gen. (in blue star) no longer cause a sharp spike in the output graph.

Maximum pop-based bonus goes down from aprox. 3.2 to 1.0 (including the initial 0.2 per pop for average species).

In the following tables:
- RP prev. refers to the sum of RP required to unlock the tech/building.
- RP and turns refer to the actual tech that grants the bonus or unlocks the building that does it.
- Turns prev. refers to the minimum number of turns required to unlock the tech/building (hence it is not the sum of turns of all prerequisites). For Black Hole Power Generator this assumes PR #2940.
- Focus: needs research/industry focus.
- Bld.: it is a building.
- Rstrc.: special restrictions apply for the bonus to be granted.
- Bonus (and Before): calculated with respect to standard RESEARCH_PER_POP and INDUSTRY_PER_POP (currently both are 0.2).
- Total: sum of bonus for end-game considering the expected average in case of system-dependent bonuses and the best bonuses in case of mutually-exclusive (total accounts for the bonuses in italics).

RESEARCH

Tech or Building | RP prev. | Turns prev. | RP | Turns | Focus | Bld. | Rstrc. | Bonus | Before
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
*Algorithmic Elegance* | 0 | 0 | 24 | 3 | Yes | No | No | 0.05 | 0.1
*Quantum Networking* | 334 | 17 | 400 | 6 | Yes | No | No | 0.1 | 0.5
*Distrib. Thought Comp.* | 656 | 11 | 150 | 5 | No | No | No | 0.1 | 0.1
Distrib. Thought Comp. (psi) | 206 | 11 | 150 | 5 | No | No | No | 0.1 | 0.1
Stellar Tomography (black hole) | 1134 | 23 | 250 | 6 | Yes | No | No | 0.2 | 1
(neutron) | 1134 | 23 | 250 | 6 | Yes | No | No | 0.15 | 0.75
*(blue/white)* | 1134 | 23 | 250 | 6 | Yes | No | No | 0.1 | 0.5
(yellow/orange/red) | 1134 | 23 | 250 | 6 | Yes | No | No | 0.05 | 0.2
[...]

PR #2965, opened 2020-05-19
Rebalance flat bonus techs

agrrr3:
This is basically #2963 with some bugfixes and refactoring.

> This is a follow up to #2943
> 
> Gas Giant Generator down to +5 and +3 (from +10 and +5)
> 
> Nascent AI down to +1 from +2.
> New tech refinement Artificial Minds +3.
> 
> Adaptive Automation down to +2 from +5.
> New tech refinement Sentient Automation +6.
> 
> AI updated accordingly to get the new tech refinements.
> I could not find any place in the AI code to tell it the new values for these techs.
> @agrrr3 @Cjkjvfnby @Morlic-fo please review this changes.
> 
> [Related thread](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11040&p=101798&sid=3f5e517fcafd674bbd677c1122909dc5#p101703)

Also micrograv gets nerfed down to +2 from +5 (see comments below)


geoffthemedio:
Meant for post release?

Oberlus:
> Meant for post release?

I recommend merging it now, it helps balancing flat vs pop-dep bonuses.

Currently, AA and GGG are overpowered early game, and AI don't give them enough priority to keep up with human players. With this PR that is at least alleviated.

default/python/AI/TechsListsAI.py
```diff
@@ -452,6 +453,7 @@ def __init__(self):
             self.economy,
             self.economy,
             self.economy,
+            self.economy,
```
Oberlus, line 456: Does the order within TechGroup3.economy matter when calling self.economy in TechGroup3A?

agrrr3: AFAIK it takes the next entry from economy dict

agrrr3:
History analyzer is missing, isnt it? Should probably have increased research/industry-per-pop instead.

default/python/AI/TechsListsAI.py
Oberlus: To be sure: that means "yes, the order in TechGroup3 does matter", right?
I'm sorry, I don't know what is a dict or how it works, if it is ordered, if "next" takes something at random, etc.

Oberlus:
> 
> 
> History analyzer is missing, isnt it? Should probably have increased research/industry-per-pop instead.

I wouldn't bother on that for now. It's a mere 5 RP for the whole game, can't influence snowballing, which is the point of this PR.

default/python/AI/TechsListsAI.py
agrrr3: Ah sorry. yes, order does matter. The tech names are in a "list" (using square brackets), which is ordered. The researchAI uses (unordered) sets and lists,  techs_to_add is the relevant one which is uses to issue tech enqueue orders so i think the order gets preserved front to back.

 A "dict" is a key-value datatype and I think ordered but not in all pythons (using squibbly brackets).

default/python/AI/AIDependencies.py
```diff
@@ -299,10 +301,6 @@ class Tags:
 
 EXOBOT_TECH_NAME = "PRO_EXOBOTS"
 
-SHP_WEAPON_ARC_DISRUPTOR_1 = "SHP_WEAPON_ARC_DISRUPTOR_1"
-SHP_WEAPON_ARC_DISRUPTOR_2 = "SHP_WEAPON_ARC_DISRUPTOR_2"
-SHP_WEAPON_ARC_DISRUPTOR_3 = "SHP_WEAPON_ARC_DISRUPTOR_3"
```
Vezzra, line 304: How do these changes relate to this PR (which is about rebalancing flat bonus techs)?

Oberlus: they don't relate to rebalance, they are grooming. Should they be removed from the PR?

Vezzra: I don't have a particular strong opinion on the matter, but generally I think it's better not to include such grooming stuff in PRs if it's not related in any way. Such things can be committed directly, without putting them into PRs.

IMO it is potentially confusing for reviewers when there are completely unrelated changes here and there in a PR...

No need to remove these changes now from this PR, but try to avoid it in the future.

Oberlus:
I forgot to change the flat bonus from Microgravity Industry.
It should be now +2 (not +3 since asteroid belts are more common than GG and also can get the bonus on the asteroid belt itself thanks to Exobots).

@agrrr3 can you do it?

agrrr3:
Will add micrograv industry.

@Morlic-fo Just dont know what new value to return here:  
```
def _base_asteroid_mining_val():
    return 5 if tech_is_complete("PRO_MICROGRAV_MAN") else 3
```

agrrr3:
> Will add micrograv industry.
> 
> @Morlic-fo Just dont know what new value to return here:
> 
> ```
> def _base_asteroid_mining_val():
>     return 5 if tech_is_complete("PRO_MICROGRAV_MAN") else 3
> ```

After thinking a while I guess the idea was that asteroids with the tech boost industry by 5 and if you do not have the tech yet (but you will soon) it is also valueable, but not so much.

So I will change that to 2 and 1.

agrrr3:
code review would be appreciated else this is good to go from my viewpoint

Oberlus:
I've tested this single-player till turn 150 twice.
AI performs well when looking at the PP&RP graphs. In fact, better than before pop-based bonuses were merged.
Without it, Nascent AI, Adaptive Auto., Micro. Ind. (when having belts) and GG Generators (when having GGs) are gamebreaking, so huge boosts compared to pop-dependent bonuses that you can set all colonies but HW to defense and perform similarly.
@geoffthemedio It's good to merge IMO
And before @Vezzra makes the release branch.

PR #2970, opened 2020-05-21
ExtractJoinGameMessageData Safety

### LINKED ISSUE #2966: freeoriond crashes on malformed player name
Toss the attached file at freeoriond tcp socket and it will abort.
malformed

geoffthemedio:
Try to prevent server crash upon receiving a malformed message. Might fix #2966

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -1138,7 +1138,14 @@ sc::result MPLobby::react(const JoinGame& msg) {
     Networking::ClientType client_type;
     std::string client_version_string;
     boost::uuids::uuid cookie;
-    ExtractJoinGameMessageData(message, player_name, client_type, client_version_string, cookie);
+    try {
+        ExtractJoinGameMessageData(message, player_name, client_type, client_version_string, cookie);
+    } catch (const std::exception& e) {
+        ErrorLogger(FSM) << "MPLobby::react(const JoinGame& msg): couldn't extract data from join game message";
+        player_connection->SendMessage(ErrorMessage(UserString("ERROR_INCOMPATIBLE_VERSION"), true));
+        server.Networking().Disconnect(player_connection);
+        return discard_event();
+    }
 
     Networking::AuthRoles roles;
     bool authenticated;


o01eg:
I suppose each `Extract*` function should be called such way.

geoffthemedio:
> I suppose each `Extract*` function should be called such way.

Wrapping them all in try-catch blocks is easy enough, but how to respond when an exception is thrown is a bit more involved. If a message is to be sent back to the connection that sent the malformed data, it needs to be picked out or added to the stringtable. And then something needs to be done after the exception is caught... Just disconnecting every connection that sends any malformed message might not always be the best response.

geoffthemedio:
pushed into master

o01eg:
Won't it be pushed to 0.4.10?

Vezzra:
Should these changes be cherry picked to the release branch? If yes, please assign to the 0.4.10 milestone and tag with "cherry-pick for release". If no, please assign to the post release milestone.

Vezzra:
bump?

server/ServerFSM.cpp
```diff
@@ -1138,7 +1138,14 @@ sc::result MPLobby::react(const JoinGame& msg) {
     Networking::ClientType client_type;
     std::string client_version_string;
     boost::uuids::uuid cookie;
-    ExtractJoinGameMessageData(message, player_name, client_type, client_version_string, cookie);
+    try {
+        ExtractJoinGameMessageData(message, player_name, client_type, client_version_string, cookie);
+    } catch (const std::exception& e) {
+        ErrorLogger(FSM) << "MPLobby::react(const JoinGame& msg): couldn't extract data from join game message";
+        player_connection->SendMessage(ErrorMessage(UserString("ERROR_INCOMPATIBLE_VERSION"), true));
```
adrianbroher, line 1145: I don't believe `ERROR_INCOMPATIBLE_VERSION` is the correct message here.  The client could also be broken or act malicious.  So unless you can identify the version of the client reliable I would rather suggest to name it like `ERROR_BAD_REQUEST` in style of the HTTP 4XX client error response codes (to be exact `HTTP 400 - Bad Request`).

geoffthemedio: sure, sounds reasonable...

PR #2971, opened 2020-05-21
Rebalance Exobots to be adequate in all planets

Oberlus:
Supersedes #2932 

* Exobots
  - are adequate in all nine planet environments,
  - are very bad (50%) at research and industry,
  - are bad at everything else, and
  - require Nascent AI and Nanotech Prod. instead of Adaptive Automation (can be researched earlier).

* Sentient Automation requires Alg. Eleg. instead of Nascent AI (decoupling of research and industry flat bonuses for more varied gameplay).

The main difference with #2932 is that the hacky "mindless species" (-50% population) is substituted by very bad industry and research. This makes them slightly less interesting mid-to-late game and saves us from having to create a fluff for "mindless species".


default/scripting/species/common/research.macros
```diff
@@ -1,5 +1,8 @@
+NO_RESEARCH
+'''[[DESCRIPTION_EFFECTSGROUP_MACRO(NO_RESEARCH_DESC)]]'''
+
```
Oberlus, line 3: Grooming: moved to top of file NO_RESEARCH macro to match the industry.macros ordering.

Oberlus:
This found a _problem_ with /check/st-tool.py:
```
$ ./check/st-tool.py check --reference default/stringtables/en.txt default/stringtables/*.txt
Traceback (most recent call last):
  File "./check/st-tool.py", line 1036, in <module>
    sys.exit(args.action(args))
  File "./check/st-tool.py", line 835, in check_action
    reference_st = StringTable.from_file(args.reference)
  File "./check/st-tool.py", line 314, in from_file
    raise RuntimeError("{}: git blame did not collect any matching author times for key {}".format(fpath, entry.key))
RuntimeError: default/stringtables/en.txt: git blame did not collect any matching author times for key BAD_RESEARCH_DESC
The command "./check/st-tool.py check --reference default/stringtables/en.txt default/stringtables/*.txt" exited with 1.
```

@adrianbroher The problem in this stringtable was that I had two keys with same name. The output error didn't let me see that, and `./check/st-tool.py format` didn't help on that neither.

Oberlus:
@geoffthemedio 
This PR will conflict with #2965.
https://github.com/freeorion/freeorion/blob/0a868811f981cd4e0514400e0791e6f12370a5ca/default/scripting/techs/production/EXOBOTS.focs.txt#L9
If you merge first #2965, I'll be able to resolve the conflict in this PR.

agrrr3:
Did you check the balance for asteroids (are they worth colonising)? If not maybe make the environment adequate (although that means less tech necessary)? Maybe increase habitable size?

Oberlus:
> Did you check the balance for asteroids (are they worth colonising)? If not maybe make the environment adequate (although that means less tech necessary)? Maybe increase habitable size?

They are now less productive, but I don't see that as bad:
- You get them earlier, and adequate for all planets. Even with the Very Bad maluses, that's more production than before.

Oberlus:
> Did you check the balance for asteroids (are they worth colonising)? If not maybe make the environment adequate (although that means less tech necessary)? Maybe increase habitable size?

They are now less productive per in asteroid belts, but I don't see that as bad: You get them (the Exobots) earlier in tech tree, and adequate for all planets. Even with the Very Bad maluses, that's more production than before, specially when taking into account the (nerfed) flat bonuses and the general pop-based nerf.
However, since the colonization of asteroid belts is not necessary to get Microgravity Industry bonus, letting Exobots colonize asteroids as adequate should not be that much of a boost. In fact, it might make sense for the original purpose of this PR: help unlucky empires.

I would not touch asteroid size in any case. We might come up with asteroid-dwelling species in the future, and having asteroid belts with average habitable size (3) seems the best.

Optionally, I would change GG size to 4 (from 6), remove the Gaseous population modifier from Sly (not needed if they have GG's size 4), and let Exobots live in GGs as poor.

agrrr3:
> Optionally, I would change GG size to 4 (from 6), remove the Gaseous population modifier from Sly (not needed if they have GG's size 4), and let Exobots live in GGs as poor.

No easily available colonizing species for GG. For deep hidden peaceful expansion you need empty spots. 

GG size 4 instead of Gaseous on its own would be a huge buff to Sly - we could try, but my guts tell me they would be OP.

My main fear for exobots-everywhere is that the environments become the same. That would be really a huge waste. But lets try it. And we should think of ways to make environments more unique and document it so it can become an "official" part of species design.

Oberlus:
> GG size 4 instead of Gaseous on its own would be a huge buff to Sly - we could try, but my guts tell me they would be OP.

Because how pre- and post-scaling priority bonuses work, the change would be a nerf instead of a buff. I did the math for target populations:
  | HW at start | Early game | End game
-- | -- | -- | --
Sly GG4 | 20 | 16 | 32
Sly GG6 -50% | 21 | 15 | 36

Early game is with Sub. Hab. With Sym. Bio. you add +4 to _Sly GG4_ and +3 to _Sly GG6 -50%_.

So I still recommend doing that change.
Later, if we think Sly needs a buff, we could make GG size 4.5. Values in the above table become 22.5, 18 and 36.

> My main fear for exobots-everywhere is that the environments become the same.

I mostly agree. The need to help balance starting conditions for more fair/fun MP games needs to be balanced with this. I hope the bad performance of exobots and the increasing colonization costs will make players still consider differently each environment depending on species availability.
I still like more the three artificial species approach, but this flat-exobots approach can be tested for a release cycle to gather wide players' opinion.

Oberlus:
Conflicts with PR #2965 resolved. Playtested, works as intended.
This is good to merge (pre- or post- v0.4.10 release, as you all wish).

PR #2978, opened 2020-05-24
Revision and extension of fields (rebased)

TheSilentOne1:
- adds new fields: nanite swarm, meteor blizzard, void rift, star forming nebula 3
- star forming nebulae slow ship speed and reduce detection, increase stealth
- fields are generated not only on the galaxy periphery, but also nearer to its center
- adjusts existing fields occurence, size and speed


geoffthemedio:
````
*** Error compiling './turn_events/turn_events.py'...
Sorry: IndentationError: unexpected indent (turn_events.py, line 28)
````

Could force push the original branch and keep the old PR.

File: default/python/turn_events/turn_events.py [hunk 1]
```diff
@@ -17,21 +17,22 @@ def execute_turn_events():
...
     systems = fo.get_systems()
     radius = fo.get_universe_width() / 2.0
-    if random() < max(0.0003 * radius, 0.03):
-        if random() < 0.4:
-            field_type = "FLD_MOLECULAR_CLOUD"
-            size = 5.0
-        else:
-            field_type = "FLD_ION_STORM"
-            size = 5.0
+    field_types = {
```
adrianbroher: No need for a dict, a list can achieve the same here.

default/python/turn_events/turn_events.py
```diff
@@ -17,21 +17,22 @@ def execute_turn_events():
...
+        1: "FLD_MOLECULAR_CLOUD",
+        2: "FLD_ION_STORM",
+        3: "FLD_NANITE_SWARM",
+        4: "FLD_METEOR_BLIZZARD",
+        5: "FLD_VOID_RIFT"
+    }
 
+    if random() < max(0.00015 * radius, 0.03):
+        field_type = field_types.get(randint(1, 5), "FLD_ERROR")
```
adrianbroher: Don't hardcode indices for selecting the field type. Use `field_types[0, len(field_types)]` or `random.choice(field_types)` instead, where the latter is even more comprehensible.

PR #2981, opened 2020-05-25
Sort includes for #2982

adrianbroher:
This PR sorts the include statements in files affected by #2982  

Sorting was done via piping the text block through `sort -V`.

PR #2982, opened 2020-05-25
Separate UniverseObjectVisitor into separate TU files

adrianbroher:
This PR separates the `UniverseObjectVisitor` in a corresponding h, cpp file pair with the same name.

It also renames the existing `Predicates.{h,cpp}` file pair to `UniverseObjectVisitors.{h,cpp}` and removes unused includes to the `Predicates.h` header.

Blocked-by: #2981


adrianbroher:
What is the status on this PR?

geoffthemedio:
>What is the status on this PR?

Blocked-by: #2981 which is going to have to wait for a bit because it will cause too many merge conflict resolution issues.

adrianbroher:
Rebased to master

PR #2984, opened 2020-05-25
Replace FOCS object description files with YAML files

adrianbroher:
This PR replaces the current zoo of entity declarations used within FreeOrion with YAML declaration files. This covers for now:

* PlanetType and system star representation (used in SidePanel; currently using XML)
* Credits listing (currently using XML)
* Empire colors (currently using XML)
* Initial fleets buildings and techs (currently using some custom file format)
* Monster fleets spawn rules (currently using some custom file format)
* Encyclopedia Articles (currently using FOCS)
* Some odd UI configuration values (currently mistreating the string tables)
* ShipPart definitions

This is an incomplete, untested work in progress and I'm still figuring out some concepts and responsibilities but the general goals are clear for me:

* Reduce the boost spirit parser to the minimum required, which means it:
  - accepts a string range only and not files anymore
  - Is only used to evaluate `ValueRef` or `Condition` expressions.
* Top level concepts like `Tech`s, `BuildingType`s or `Species` declarations are represented through YAML maps.
* This PR will keep the currently implemented macro substitution, but limit it to YAML scalars.
* This PR will only implement a 1 to 1 mapping and try to avoid optimizations like implementing a better YAML representation the aforementioned top level concepts or merging/rearanging logical connected concepts (e.g. merge the connected Tech, the ShipPart and the Encyclopedia Article into a single file).
* This PR will convert all of the FOCS files into their YAML counterpart.
* This PR will convert all of the XML files into their YAML counterpart.
* This PR will convert all of the .inf fiels into their YAML counterpart.
* It will drop the `#include` keyword and general mistreatment for the StringTables.


o01eg:
If we are going to use it, then `yaml-cpp` should become part of SDK for Windows and MacOS.

Cjkjvfnby:
You can add yml validator to the Travis checks.

> This PR will convert all of the XML files into their YAML counterpart.

I would say that splitting these changes into parts will greatly speed up the review process and make everyone happier.

You can start from: `Yaml introduction and credits.xml`

adrianbroher:
> I would say that splitting these changes into parts will greatly speed up the review process and make everyone happier.

This is not the intent of this PR.  Checking for general feasibility, run- and compiletime performance and implementation complexity/user convenience comes first.  Auxiliary tasks stuff like CI integration can be tackled at a later point.

adrianbroher:
That said, @o01eg and @Cjkjvfnby: you should probably unassign yourself from this PR as reviewer because of the codeowners file and that I've edited the ru.txt translation table.  This PR is far from done and I don't know if you want to recieve the spam associated to changes on this PR.

Cjkjvfnby:
> This is not the intent of this PR.

You can use the Draft feature for such purposes. Geoff has already changed PR to it.

Cjkjvfnby:
Recently I have found [TOML](https://github.com/toml-lang/toml).

This language is used in Python and Rust for [packaging](https://www.python.org/dev/peps/pep-0518/#file-format). 

> This format was chosen as it is human-usable (unlike JSON [7]), it is flexible enough (unlike configparser [9]), stems from a standard (also unlike configparser [9]), and it is not overly complex (unlike YAML [8]). The TOML format is already in use by the Rust community as part of their Cargo package manager [14] and in private email stated they have been quite happy with their choice of TOML.

I don't know if it fits all cases and have support for C++.

adrianbroher:
> TOML

Horrible syntax, not interested, no proper libraries available, less tooling, I want the "overcomplexity" like anchors or node tags.

o01eg:
Other idea was to use Python's `ast` module to parse python syntax instead of FOCS.

agrrr3:
Refactoring can be nice for reducing clutter. Having configuration available in formats which are directly readable by multiple languages (in our case that means at least cpp, python and i would also like to have web support i.e. javascript) actually could provide some value.

Use for FOCS content though is dubious. FOCS content is usually not just configuration. E.g. YAML does not by itself provide includes so this is not language-agnostic.
One thing to look at in your prototype would be if FOCS scripting errors are better reported than in the current parser.

What we really need though is some way to access the same game values in AI, stringtables, scripting and backend. And that probably means support for includes and certainly means support for basic arithmetics.

Discussion here: [RFC: A single source for default values (FOCS, python, stringtables, backend)](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11427)

adrianbroher:
@agrrr3 I'm sick about all talk and no outcome.  If you feel like there should be something like that feel free to implement it yourself but don't ask me for solving your problems in the way you prefer.

agrrr3:
> @agrrr3 I'm sick about all talk and no outcome. If you feel like there should be something like that feel free to implement it yourself but don't ask me for solving your problems in the way you prefer.

I did not expect you to solve "my" problems. I just think you are not adding much value with what you try to achieve in this PR as far as I can see (looks like "I hob zwoar ka ohnung wo i hinfoahr. Aber dafür bin i gschwinder duat"). And I give some context to that claim.

geoffthemedio:
> I just think you are not adding much value

So far as I understand, the purpose of these changes is not to add features. Your requests for added functionality is reasonable in of itself, but not really relevant to this pull request, I think.

But there still are, or could be, benefits from refactoring and switching formats for representing game content, though. Faster compile times, faster parse times, easier extension or alteration of the format, perhaps automated generation of description of the format for documentation purposes, availability of standardized tools to view or edit the scripts, and probably others...

agrrr3:
> So far as I understand, the purpose of these changes is not to add features. Your requests for added functionality is reasonable in of itself, but not really relevant to this pull request, I think.

The relevance was that these are sorely missing features for a maintainable base and some small language might have the necessary features for our use cases. But of course, one could switch afterwards away from FOCS embedded in YAML+macro to whatever that language base that may be. Just dont see much value in an intermediate step if it does not buy something else. Especially if the stated purpose is to reduce the language zoo.

What I would have expected if somebody tackles the configuration zoo: research the use cases and choose a fitting language. 

Also I do not see a rationale for using YAML besides there is some tool support.

> But there still are, or could be, benefits from refactoring and switching formats for representing game content, though. Faster compile times, faster parse times, easier extension or alteration of the format, perhaps automated generation of description of the format for documentation purposes, availability of standardized tools to view or edit the scripts, and probably others...

I can see the "could be" component. It could also become worse than it is right now. A prototype is a good thing to find that out if it gets measured.

Vezzra:
Looks interesting. Replacing the current patchwork of configuration formats, files and locations with a more consistent and unified solution, and using YAML for that sounds like a good idea.

The only thing where I'm not sure if that really makes sense is moving FOCS stuff to YAML, considering our plan to move all content scripting to Python at some point. No strong opinion, though. Worst case, some effort will have been wasted...

adrianbroher:
> considering our plan to move all content scripting to Python at some point

Hm? Didn't you forget that I made the initial suggestion regarding this topic? And I don't believe that Python is the solution for this.

Vezzra:
No, I didn't forget that, but I wasn't sure if you maybe have discarded the idea since, and wanted to point out that (IMO) that plan is still on the roadmap.

Python is certainly not the solution for most of what you're trying to replace here with YAML, only for replacing FOCS for the content scripting stuff.

Anyway, as I said, no strong opinion. The only thing I *do* have a strong opinion of or preference for, is getting rid of FOCS in the long run. The parser lib is a PITA...

adrianbroher:
Trying to figure out some quirks together with yaml-cpp upstream: https://github.com/jbeder/yaml-cpp/issues/871

PR #2988, opened 2020-05-26
make_versioncpp.py: don't get pissed at commit signatures

l29ah:
that get shown by default if user desires so in their gitconfig

### DIFF
--- a/cmake/make_versioncpp.py
+++ b/cmake/make_versioncpp.py
@@ -133,8 +133,8 @@ try:
         branch = ""
     else:
         branch += " "
-    commit = check_output(["git", "show", "-s", "--format=%h", "--abbrev=7", "HEAD"], universal_newlines=True).strip()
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Could you elaborate on what issue this fixes?

l29ah:
https://github.com/github/hub/issues/1379

Vezzra:
This can be potentially problematic for people who (for whatever reasons) use a very old version of git (2.4.0 or older):

https://github.com/golang/go/issues/26844

That aside, the link you, @l29ah, provided, didn't really really make it clear to me what the issue is. Does `make_versioncpp.py` fail or break in any way if users set that option in their gitconfig?

l29ah:
Yes, it stops the compilation, otherwise i wouldn't have bothered to fix it.

geoffthemedio:
OK then...

Issue #2964, opened 2020-05-19
Users can sniff server-client handshake and get sensible info that should be unknown to their clients


Oberlus:
Bug Report
==========

Environment
-----------
* **FreeOrion Version**: 9eb021f
  * Compiled from source

Description
-----------
Server sends unnecessary info to clients, which allows the users to sniff the packets and deduce information that should be unknown to them or to their clients.
This includes items in the production and research queues. 
The following is an extract of a conversation in the freeorion XMPP chatroom in which user @l29ah revealed such kind of information that should not be possible for him to know.

> (17:57:48) L29Ah: achtung, danyspin97 builds an orbital incubator
> (17:58:58) L29Ah: and O01eg has interstellar logistics research enqueued
> (18:00:16) L29Ah: ThinkSome: watch out for Oberlus, he enqueued lifecycle manipulation
> (18:00:33) L29Ah: and has a sexbomb
> (18:15:29) Querens: L29Ah: describe sexbomb
> (18:16:33) L29Ah: Querens: i can't 
> (18:17:38) Querens: L29Ah: why?
> (18:18:32) L29Ah: dunno how to resolve design ids
> (18:18:48) L29Ah: probably they're private to the empire
> (18:19:07) L29Ah: or i need a separate call for that
> (18:21:52) Querens: I thought you actually see it
> (18:22:59) Voker57: hax
> (18:30:54) Oberlus: L29Ah: so no scout? Would be good to know what you are doing to get that info so that it can be fixed
> (18:31:35) L29Ah: i'm just looking at the freeorion handshake with a sniffer
> (18:31:54) L29Ah: it sends a megabyte worth of xml
> (18:32:04) L29Ah: with various universe info
> (18:36:26) Oberlus: O01eg: I think it's a bit buggy that empire A gets info from empire B that was unknown to it. Because of the obvious hacks and because of unnecessary bandwidth comsumption
> (18:37:35) Oberlus: L29Ah: did you get this by connecting with user "L29Ah" or trying other players' names?
> (18:42:35) O01eg: Oberlus: it's a buggy but someone need to implement server-side calculation for ship's meters.
> (18:43:24) Oberlus: You mean calculations are done in clients?
> (18:43:53) Oberlus: How come then that having the wrong default folder doesn't fool the server?
> (18:44:30) Oberlus: I mean, I can connect with different FOCS files, server says checksum is bad, and I can play
[...]

PR #2991, opened 2020-05-27
Hide some sensible info from enemy empires

o01eg:
Partially solves https://github.com/freeorion/freeorion/issues/2964

Removes unneeded game rule.
Allows allies to see empire data.
Enemy techs are still used in ship meters calculation on the client.

Most notable thing is enemy won't see queued buildings.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1249,12 +1249,6 @@ Drydocks and Fleet Field Repair repair allied ships
 RULE_ENABLE_ALLIED_REPAIR_DESC
 Allow Drydocks building and Fleet Field Repair tech to repair ships of allied empires.
 
-RULE_SHOW_DETAILED_EMPIRES_DATA
-Show detailed empires data
-
-RULE_SHOW_DETAILED_EMPIRES_DATA_DESC
-Show research and production, researched buildings, parts, and hulls of all empires.
-
 RULE_DIPLOMACY
 Allowed diplomacy
 
--- a/default/stringtables/es.txt
+++ b/default/stringtables/es.txt
@@ -1233,12 +1233,6 @@ Escalado de coste de partes
 #*RULE_ENABLE_ALLIED_REPAIR_DESC
 #* <not translated 1970-01-01 01:00:00>
 
-#*RULE_SHOW_DETAILED_EMPIRES_DATA
-#* <not translated 1970-01-01 01:00:00>
-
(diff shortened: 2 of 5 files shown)


Voker57:
not seeing queued buildings is fine with me, not ideal but if it is a way to solve the problem in reasonable time, ok

geoffthemedio:
This may lead to more "unknown" meter contributions on clients. Probably not too many, though...

o01eg:
> This may lead to more "unknown" meter contributions on clients. Probably not too many, though...

We tested it one time before and it looks like only techs are used for client-side meters calculation.

o01eg:
@geoffthemedio I found it possible to break if client doesn't know if it get additional data because of alliance.
Can it be ensured that diplomatic statuses known before empires serialization/deserialization?

geoffthemedio:
alliance timing is tricky now as it occurs independent of turn timing. it needs to be reworked as a set of diplomatic orders with connected messages to players.

o01eg:
I think I just shouldn't send information from `if (visible)` block for allied empires for now.

geoffthemedio:
> I think I just shouldn't send information from `if (visible)` block for allied empires for now.

not sure exactly what you mean, but you can probably do as you prefer...

geoffthemedio:
> Enemy techs are still used in ship meters calculation on the client.

Really? Doesn't look like they get sent to non-allies.

Also empires will now immediately forget former allies' tech if they un-ally...

geoffthemedio:
This pull request appears to remove sending techs under the `RULE_SHOW_DETAILED_EMPIRES_DATA` rule and adds sending techs when allied. How will clients access enemy techs?

o01eg:
@geoffthemedio No, there only compatibility code left for techs. Branch with sending incomplete techs was removed with rule.

geoffthemedio:
So why does the description say "Enemy techs are still used in ship meters calculation on the client." if those techs aren't sent to the enemy clients?

o01eg:
@geoffthemedio Those techs are sent because they used.

geoffthemedio:
> Those techs are sent because they used.

What does it mean for a tech to be "used", and where is the code that sends them to enemy players?

o01eg:
https://github.com/freeorion/freeorion/blob/ff86758cf02d21fe06df97630f2c0534e2a7b82a/util/SerializeEmpire.cpp#L151

Issue #3001, opened 2020-06-03
Sly & Gaia special - planet type cycling, colonization weighting


rrobgill:
Description

Prevent planet types cycling when Sly inhabit a planet with Gaia special, and have AI devalue Gaia special when evaluating planets for Colonization

Gain

1. At present the planet cycles through the planet types never stopping. Clearly Sly will never reach good suitability on a non-gaseous planet.

2. When evaluating planets for Colonization, the AI presently treats a Gaia special as favorable for Sly, even though Sly receives no benefit from it.

Exobots also receive no benefit from Gaia specials, a Gaia special does not transform the planet type when inhabited by them, and the AI devalues the Gaia special when evaluating planets for Colonization.

Discussion

https://www.freeorion.org/forum/viewtopic.php?f=6&t=11685

agrrr3:
great catch!

Vezzra:
That isn't a feature request, but a bug report. The current behavior should be considered a bug.

Vezzra:
@rrobgill, with #3002 merged, can this be closed as resolved?

rrobgill:
Yes, resolved by #3002

PR #3002, opened 2020-06-03
Sly & Gaia special - prevent type cycling, reweight colonization

rrobgill:
1.  alter the Gaia special to prevent it from cycling through
    planet types when occupied by Sly (similar to Exobots)

2.  alter the Colonization AI to devalue the Gaia special as Sly derive
   no benefit from it (similar to Exobots)

Related Issue #3001 

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -133,8 +133,8 @@ def calc_max_pop(planet, species, detail):
         detail.append("%s (maxPop%+.1f)" % (_special, this_mod))
         base_pop_not_modified_by_species += this_mod
 
-    #  exobots can't ever get to good environ so no gaia bonus, for others we'll assume they'll get there
-    if "GAIA_SPECIAL" in planet_specials and species.name != "SP_EXOBOT":
+    #  exobots and sly can't ever get to good environ so no gaia bonus, for others we'll assume they'll get there
+    if "GAIA_SPECIAL" in planet_specials and species.name != "SP_EXOBOT" and species.name != "SP_SLY":
         base_pop_not_modified_by_species += 3
         detail.append("Gaia_PSM_late(3)")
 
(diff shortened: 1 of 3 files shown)


agrrr3:
I would prefer if you could refactor that in FOCS to a more general solution:

Not sure if it works, but you should be able to check in the activation condition if Source.NextBetterPlanetType is the current planet type and only fire the effect if it is not.
Then remove the special cases for Sly and exobots

rrobgill:
@agrrr3 it seems that NextBetterPlanetType (as visible within FOCS) will never stabilize for Sly, but will always be the current type +1 (clockwise). There is no way at this level to distinguish between this case (continual cycling with no goal) and ordinary clockwise progression towards a goal for another species.

(In /universe/Planet.cpp and universe/ValueRefs.cpp it seems there is no logic in NextBetterPlanetType() to handle a species that doesn't have a favoured planet falling amongst the terrestrial planets.)

The other thought that occured was adding a _preferredenvironment_ class for each species (could be done en-masse for most species in env.macros, would only need additions to ancient guardians, exobots and sly (or add macros for their environmental tolerances to env.macros as well) but this would also involve backend changes (SpeciesParser.cpp at the least).

geoffthemedio:
> In /universe/Planet.cpp and universe/ValueRefs.cpp it seems there is no logic in NextBetterPlanetType() to handle a species that doesn't have a favoured planet falling amongst the terrestrial planets.

That should not be difficult to fix...

rrobgill:
The BetterPlanet detection logic has been reworked to only consider terrestrial planets.

Using @agrrr3's suggested method allowed removing the species specific tests for the Gaia special.

geoffthemedio:
@agrrr3 is this satisfactory, or would you prefer something else...?

geoffthemedio:
@agrrr3 ?

agrrr3:
I think for the moment this suffices well

geoffthemedio:
not sure about whether it should go into v0.4.10...

Vezzra:
It's a bugfix, so I'd say it should.

Issue #2997, opened 2020-06-01
Windows' resize bottom-right button is unusable if window becomes too small


Oberlus:
Bug Report
==========

Environment
-----------
* **FreeOrion Version**: v0.4.10 and every other one before.
* **Operating System**: Ubuntu 18.04, and probably any other OS.
* **Graphic card used**: GTX 860M
* **Fetched as**
  * Compiled from source

Description
-----------
This "button":
![Corner_resize_button](https://user-images.githubusercontent.com/28650914/83383467-768cd200-a3e5-11ea-8024-e4b2c648b8d1.png)

It becomes unselectable in the Design Details window (the one that shows the hull with the part slots) when that window is too small:
![Corner_resize_button_unusable_if_window_too_small](https://user-images.githubusercontent.com/28650914/83383552-a340e980-a3e5-11ea-8a23-7ab454a87a82.png)

That is, placing the mouse pointer over it will not highlight the 2 oblique lines and user won't be able to drag the window's corner to resize the window.

Note that this happens only with the Design Details window. In the above screenshot, I can select the Pedia's corner button to resize it to normal dimensions.

A secondary and maybe related problem is that the main panels (if that is the name, I'm not a GUI designer) of all windows overlaps their resize corner buttons:
![Corner_resize_button_should_overlap_main_panel](https://user-images.githubusercontent.com/28650914/83383814-31b56b00-a3e6-11ea-904a-a1c04d1036af.png)

A tertiary and also maybe also related problem is that some elements of a too-small window are drawn outside its window (see the Pedia's window in the second screenshot attached above).

Expected Result
---------------

1. Windows' bottom-right resize button is usable no matter the size of the window.
2. That button always overlaps any other window element.
3. No window element is drawn outside the window (see the Pedia's window in the second screenshot attached above).

Steps to reproduce
------------------
* Start a new game and go to Ship Design.
* Drag the bottom-right corner button of Design Details window and resize it to its minimum possible size.
* Issue occurs (user can't undo the resizing).

Workaround
----------
After the Design Details window got too small (and unusable) for whatever reason, you will find a section like the one below in config.xml file (I've deleted the irrelevant parameters and subsections):
```
    <design>
      <edit>
        <windowed>
          <height>93</height>
          <width>50</width>
        </windowed>
      </edit>
    </design>
```
You can give bigger numbers to height and width to get back a usable Design Details window.

Related issue: #2987

Oberlus:
@geoffthemedio 
[...]

PR #3006, opened 2020-06-04
 Set minimum window sizes on design screen 

rrobgill:
Fixes issue raised by bug #2997 (Several windows on the design screen had no minimum size specified, which caused issues of loss of resize control and rendering glitches if the window was inadvertently shrunk to very small sizes.)


geoffthemedio:
Please rebase against master. Pull requests should not contain merges or apparent duplicate commits.

PR #3007, opened 2020-06-04
Set minimum window sizes on design screen

rrobgill:
Fixes issue raised by bug #2997 (Several windows on the design screen had no minimum size specified, which caused issues of loss of resize control and rendering glitches if the window was inadvertently shrunk to very small sizes.)

Replaces PR #3006 

Signed-off-by: Rob Gill rrobgill@protonmail.com

### DIFF
--- a/UI/DesignWnd.cpp
+++ b/UI/DesignWnd.cpp
@@ -1896,6 +1896,8 @@ void DesignWnd::PartPalette::CompleteConstruction() {
 
     CUIWnd::CompleteConstruction();
 
+    SetMinSize(GG::Pt(PART_CONTROL_WIDTH + 70 ,PART_CONTROL_HEIGHT + 70));
+
     DoLayout();
     SaveDefaultedOptions();
 }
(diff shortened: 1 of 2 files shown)


UI/DesignWnd.cpp
```diff
@@ -1896,6 +1896,8 @@ void DesignWnd::PartPalette::CompleteConstruction() {
 
     CUIWnd::CompleteConstruction();
 
+    SetMinSize(GG::Pt(PART_CONTROL_WIDTH + 70 ,PART_CONTROL_HEIGHT + 70));
```
adrianbroher: Where does that magical `70` come from and why is it not `75` or `65`?

rrobgill: I borrowed the magic from the options window, (where they used 20/70)

geoffthemedio: Could have a separately defined `const GG::Pt PALETTE_MIN_SIZE{GG::X{PART_CONTROL_WIDTH + 70}, GG::Y{PART_CONTROL_HEIGHT + 70}};` to clarify?

Issue #2777, opened 2020-02-23
SP_TRITH always have unknown population modifier


aschuch247:
### Environment

* **FreeOrion Version**: 3ac143d
* **Fetched as**
  * Compiled from source

### Description

Enemy colonies of `SP_TRITH` always show unknown population modifier.

### Expected Result

The reason of the population minus shall be shown.

### Steps to reproduce

* Start a new game, using seed `9` and only a single AI, also chose `SP_TRITH` as species.
* The own capital planet has a population of 32 (large radiated).
* Build super testers.
* The enemy capital has a target population of 27.4 (large radiated).
* Furthermore, all enemy colonies of `SP_TRITH` show a negative unknown population modifier.

geoffthemedio:
Does it only happen after https://github.com/freeorion/freeorion/pull/2774 ?

aschuch247:
@geoffthemedio: No. I just compiled be647f6c3c6c43845525242a8fd019ae2319e225 and the same happens.

aschuch247:
![trith](https://user-images.githubusercontent.com/13714468/75123003-8fc20c00-569b-11ea-8f6a-82291cfc4388.png)

A Derthrean colony exists in the same system as the Trith capital. My guess is that the xenophobic frenzy is just misplaced in the client.

PR #3008, opened 2020-06-05
Fix Trith population target accounting

rrobgill:
As raised in Bug #2777, script contained incorrect accounting tag, leading to population malus
being attributed to 'unknown'.

### DIFF
--- a/default/scripting/species/SP_TRITH.focs.txt
+++ b/default/scripting/species/SP_TRITH.focs.txt
@@ -32,7 +32,7 @@ Species
 
         [[XENOPHOBIC_SELF]]
         [[XENOPHOBIC_OTHER(TRITH)]]
-        [[TELEPATHIC_DETECTION(5)]]
+        [[TELEPATHIC_DETECTION(5,2)]]
 
         // not for description
         [[AVERAGE_PLANETARY_SHIELDS]]
--- a/default/scripting/species/common/telepathic.macros
+++ b/default/scripting/species/common/telepathic.macros
@@ -1,3 +1,17 @@
+PRECOGNITIVE_DETECTION
+'''     EffectsGroup
+            description = "PRECOGNITIVE_DETECTION_DESC"
+            scope = And [
+                PopulationCenter
+                Not VisibleToEmpire empire = Source.Owner
+                Not Source
+                Not OwnedBy empire = Source.Owner
+                WithinStarlaneJumps jumps = @1@ condition = Source
+            ]
+            activation = OwnedBy affiliation = AnyEmpire
+            effects = SetVisibility empire = Source.Owner visibility = max(Basic, Value)
+'''
+
 TELEPATHIC_DETECTION
 '''     EffectsGroup
             description = "TELEPATHIC_DETECTION_DESC"
@@ -10,5 +24,9 @@ TELEPATHIC_DETECTION
                 WithinStarlaneJumps jumps = @1@ condition = Source
             ]
             activation = OwnedBy affiliation = AnyEmpire
-            effects = SetVisibility empire = Source.Owner visibility = max(Basic, Value)
+            effects = [
+                SetVisibility empire = Source.Owner visibility = max(Basic, Value)
+                If Condition = WithinStarlaneJumps jumps = @2@ condition = Source
+                    effects = SetVisibility empire = Source.Owner visibility = max(Partial, Value)
+            ]
 '''
--- a/default/scripting/species/common/xenophobic.macros
+++ b/default/scripting/species/common/xenophobic.macros
@@ -31,7 +31,7 @@ XENOPHOBIC_SELF
                 SetTargetResearch value = Value * 0.9
                 SetTargetHappiness value = Value - (
                     Statistic Count condition = And [
-                        Planet
+                        PopulationCenter
                         OwnedBy empire = Source.Owner
                         Not OR [ 
                             Species name = Source.Species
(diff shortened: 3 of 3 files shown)


geoffthemedio:
Could you add a sentence to the pull request (if not the commit) explaining what the bug in the script actually was and why / how this fixes it?

geoffthemedio:
Doesn't seem to work for me...
Trith_Unknown_Pop_Effect
[TrithTest.zip](https://github.com/freeorion/freeorion/files/4742083/TrithTest.zip)

rrobgill:
Interesting. It seems the actual target is including all nearby systems, whether invisible or unexplored, but in its current form XENOPHOBIC_SELFSUSTAINING_QUALIFYING_PLANET_COUNT seems to be only including planets which are visible to the empire. (Once all the planets are in visible, they are all included in the Xenophobic Frenzy count).

**I'm curious if this was the original intention, that it would only be accounted for once the species has become visible ?**

The same game let to run 3 occupied planets are now visible and their effect is included within the -8.74 attributed to Xenophobic frenzy.

geoffthemedio:
My guess is that on the server, the effects are applied correctly considering the nearby planets that the Trith planet's owner has visibility of. On the human client, the player empire can see all the planets, but has no knowledge of what planets the other empires that own the planets can see, so the player client visibility condition thinks that the owners of other empires' planets can't see any planets, so it doesn't count any planets as being visible and in range for the effect.

I think a reasonable solution would be to remove the visibility condition from the effect. There will be some cases where a not-visible other-species planet nearby produces the frenzy effect on a player's own planets, which might actually be more visible than the problem noted in this thread... But perhaps also the frenzy effect should give visibility of the planets that cause it, in order to avoid that problem?

o01eg:
> I think a reasonable solution would be to remove the visibility condition from the effect. There will be some cases where a not-visible other-species planet nearby produces the frenzy effect on a player's own planets, which might actually be more visible than the problem noted in this thread... But perhaps also the frenzy effect should give visibility of the planets that cause it, in order to avoid that problem?

@geoffthemedio I suppose disclosure of planets is inappropriate. What if the server send to client how it calculated effects instead of trying to calculate them same way as the server?

rrobgill:
Is there a way in FOCS to check level of visibilty (Invisible, basic, partial, full) of the PopulationCenter ? 

It can be set in this granular way with SetVisibilityLevel. Is there an equivalent way to read it?

eg:
````
XENOPHOBIC_SELFSUSTAINING_QUALIFYING_PLANET_COUNT
'''
Statistic Count condition = And [
            PopulationCenter
            Not OR [
                Species name = Source.Species
                Species name = "SP_EXOBOT"
            ]
            Not Population high = 0
            WithinStarlaneJumps jumps = 2 condition = Source
            **VisibleToEmpire empire = Source.Empire visiblility = Full**
            ]
'''
````

geoffthemedio:
`VisibleToEmpire empire = Source.Owner` matches objects that are visible to an empire, but can't currently check for level of visibility (ie. >= basic visibility).

There is also a ValueRef to look up an object's visibility to an empire as a value (not as a condition), though it wasn't being used and had an unnoticed typo in it until just now: https://github.com/freeorion/freeorion/commit/01980bfe0052a69e931fec526b24615b79da8eb8 with parsing here: https://github.com/freeorion/freeorion/blob/master/parse/VisibilityValueRefParser.cpp#L34

That might, but probably wouldn't, work with a value comparison like https://github.com/freeorion/freeorion/blob/master/default/scripting/fields/FLD_MOLECULAR_CLOUD.focs.txt#L47 with constants like `Invisible`, `Basic`, `Partial`, and `Full`.

And you can also refer to the target object's visibility when setting it with `Value` as in https://github.com/freeorion/freeorion/blob/master/default/scripting/species/common/telepathic.macros#L12

rrobgill:
Unfortunately it does not seem to work in a value comparison.

Likewise I can't find a way to access the planet's detection Strength (rather than range) for comparison within FOCS.

eg

````
XENOPHOBIC_SELFSUSTAINING_QUALIFYING_PLANET_COUNT
'''
Statistic Count condition = And [
            PopulationCenter
            Not OR [
                Species name = Source.Species
                Species name = "SP_EXOBOT"
            ]
            Not Population high = 0
            WithinStarlaneJumps jumps = 2 condition = Source
            Stealth high = Source.**DETECTION_STRENGTH** - 1
            ]
'''
````

**Testing setting the stealth level manually seems to generate targets in agreement with the server.**

But unless I'm missing something obvious the only FOCS friendly workaround I can think of would involve the ability to select between 18 different QUALIFYING_PLANET_COUNT  calculations depending on technology level and buildings present on the Trith planet. (spy_detection 1, and each level of a spy_detection tech (2-5) against relevant presence of Scanning Facility, all against presence of Panopticon special).

In each case the QUALIFYING_PLANET_COUNT would then need to use a heuristic to convert back from the planetary detection range back to the detection strength. 

It's too inefficient to consider, and would break in the event of any changes to detection technologies, and any additional detection related buildings or specials would cause exponential expansion of the number of calculations needed.

geoffthemedio:
> Unfortunately it does not seem to work in a value comparison.

Makes sense... the `ValueTest` condition only works for ints, doubles, and strings. This could be made to work by adding an `IntCast` valueref to convert the various enum-valued valuerefs into ints, if needed.

But what I think would best suit this application is to extend the `VisibleToEmpire` condition to take a level parameter, such that it will only match objects that have at least the specified level of visibility, or perhaps to have min/max parameters to specify a range of visibility levels.

Or just ignore visibility when evaluating the effectsgroup scope and set the planets visible to the owner of the Trith planet as part of the effect. Might be a few off-by-one-turn issues, but it makes sense to have those planets be visible anyway, I think...

> Source.**DETECTION_STRENGTH**

For the planet's owner empire detection strength, that is an empire meter, accessible with something like `EmpireMeterValue empire = Source.Owner meter = "METER_DETECTION_STRENGTH"`

I don't like checking stealth vs. detection strength though, as there are other ways a planet could be visible to an empire.

geoffthemedio:
> @geoffthemedio I suppose disclosure of planets is inappropriate. 

In this case it would make sense, I think. It's already a rare case that these planets wouldn't be visible, I'd think, so not a big issue. And since the Trith are being telepathically annoyed by the presence of another species nearby, being able to see where they are would make sense for gameplay and fluff explanation... Getting some extra visibility is a nice tradeoff for the penalty the nearby species produce.

> What if the server send to client how it calculated effects instead of trying to calculate them same way as the server?

That would require substantial reworking of how effect accounting is done, and sort of misses the point of the unknown effects, which are to document the differences between what an empire knows and what's really happening. Players aren't supposed to know the source of all effects, unless they can detect and understand that source.

rrobgill:
I don't know my way around the code well enough yet to be extending the FOCS interface, that would be for a longer term. 

The ability to determine visibility level by VisibleToEmpire would be handy, but will unfortunately run into the same problem here. 

For example (especially with Sly) where the planet itself is VisibleToEmpire = Basic (gained by telepathic detection), but the inhabitants are stealthy, the Population is unknown to the Trith. (The Sly planet in the save game above is excluded by the "Not Population high = 0"). The server can see the population, the client cannot.

From your earlier suggestion of increasing Visibility, https://github.com/freeorion/freeorion/pull/3008/commits/2fd521be7be55b7d7754103baa1e815e02417f34 alters Telepathic Detection in Self Sustaining Species to grant Partial rather than Basic visibility to planets within two jumps, (remaining at basic for 3-5 jumps). This allows the Trith to see the population and species of the nearest planets and prevents the need to exclude based upon stealth level, and eliminates the need to check for detection and stealth.

geoffthemedio:
> The ability to determine visibility level by VisibleToEmpire would be handy, but will unfortunately run into the same problem here.
> 
> For example (especially with Sly) where the planet itself is VisibleToEmpire = Basic (gained by telepathic detection), but the inhabitants are stealthy, the Population is unknown to the Trith. (The Sly planet in the save game above is excluded by the "Not Population high = 0"). The server can see the population, the client cannot.

In that case, you could require Partial visibility, and thus knowledge of the species, for the xeno frenzy effect to activate.... That's the point of the condition extension. Or am I missing something?

rrobgill:
Not missing anything that's essentially the same conclusion I reached.

Better coders than me have decided not to pursue this: 
https://github.com/freeorion/freeorion/blob/e4f72d84fb4699a066c1e0eeee60166a0d6b0635/parse/VisibilityValueRefParser.cpp#L70-L71 
So I think I'll check out of this one at this point too. I don't have enough familiarity with the backend code at this time to complete the extension to EmpireObjectVisiblity.

geoffthemedio:
There's still no reason to have direct `Object.Visibility` be supported in `ValueRef::Variable<Visibility>`. Except perhaps in some particular case that's not immediately necessary, object visibility would only make sense as a `ComplexVariable` because it needs parameters like for-which-empire and is not just a simple property of an object. Hence https://github.com/freeorion/freeorion/blob/master/parse/VisibilityValueRefParser.cpp#L34

PR #3010, opened 2020-06-06
Random Refactoring

geoffthemedio:
-Replaces `std::random_shuffle`, which is removed in C++17
-Moves all access of the random generator behind mutex locks
-Replaces `boost` random generator and distributions with `std` equivalents

### DIFF
--- a/UI/ClientUI.h
+++ b/UI/ClientUI.h
@@ -226,6 +226,10 @@ private:
     std::shared_ptr<SaveFileDialog>         m_savefile_dialog;
     std::shared_ptr<PasswordEnterWnd>       m_password_enter_wnd;   //!< the authentication window
 
+    //!< map key represents a directory and first part of a texture filename.
+    //!< when textures are looked up with GetPrefixedTextures, the specified
+    //!< dir is searched for filenames that start with the prefix. pointers
+    //!< to the Texture objects for these files are stored as the mapped value.
     std::map<std::string, std::vector<std::shared_ptr<GG::Texture>>>
                                             m_prefixed_textures;
 
--- a/UI/ClientUI.h
+++ b/UI/ClientUI.h
@@ -215,14 +215,6 @@ public:
     //!@}
 
 private:
-    typedef std::pair<std::vector<std::shared_ptr<GG::Texture>>,
-                      std::shared_ptr<SmallIntDistType>>    TexturesAndDist;
-    typedef std::map<std::string, TexturesAndDist>          PrefixedTextures;
-
-    TexturesAndDist PrefixedTexturesAndDist(const boost::filesystem::path& dir,
-                                            const std::string& prefix,
-                                            bool mipmap);
-
     void HandleSizeChange(bool fullscreen) const;
     void HandleFullscreenSwitch() const;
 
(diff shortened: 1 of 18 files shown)


util/Random.cpp
```diff
@@ -1,56 +1,62 @@
...
+    gen.seed(static_cast<GeneratorType::result_type>(diff.total_milliseconds()));
 }
 
-SmallIntDistType SmallIntDist(int min, int max) {
-    boost::mutex::scoped_lock lock(s_prng_mutex);
-    return SmallIntDistType(gen, boost::uniform_smallint<>(min, max));
+int RandInt(int min, int max) {
+    if (min <= max)
+        return min;
+    std::uniform_int_distribution<> dis(min, max);
```
adrianbroher: C++11 does not guarrantee reproducible distributions accross implementations:

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2059r0.pdf

util/Random.cpp
```diff
@@ -1,56 +1,62 @@
...
-IntDistType IntDist(int min, int max) {
-    boost::mutex::scoped_lock lock(s_prng_mutex);
-    return IntDistType(gen, boost::uniform_int<>(min, max));
+double RandZeroToOne()
+{ return RandDouble(0.0, 1.0); }
+
+double RandDouble(double min, double max) {
+    if (min <= max)
+        return min;
+    std::uniform_real_distribution<> dis(min, max);
```
adrianbroher: Same applies here

util/Random.cpp
```diff
@@ -1,56 +1,62 @@
...
+    }
 }
 
-DoubleDistType DoubleDist(double min, double max) {
-    boost::mutex::scoped_lock lock(s_prng_mutex);
-    return DoubleDistType(gen, boost::uniform_real<>(min, max));
+double RandGaussian(double mean, double sigma) {
+    if (sigma <= 0.0)
+        return mean;
+    std::normal_distribution<> dis(mean, sigma);
```
adrianbroher: And here

adrianbroher:
Seems like you can combine the boost random distribution with the std generator:

https://github.com/Orphis/RandoBlazer/blob/master/Random.cpp

Cjkjvfnby:
Could you update docstring for Random module with information about the distribution and stability of the result on different platforms?

>  -Moves all access of the random generator behind mutex locks

How may random calls do we have? Does it affect performance?

geoffthemedio:
> > -Moves all access of the random generator behind mutex locks
> 
> How may random calls do we have?

I found 26 `RandInt` calls, 12 `RandZeroToOne` calls, and 7 `RandDouble` calls in C++ code. Some of those are in loops over objects in a set, and how often some are used depends on scripted content and Python..

> Does it affect performance?

Probably not much. The previous implementation was using similar locks in most of the same places, but in a way that didn't actually always usefully guard the generator in some cases.

Some during universe generation, some (probably unnecessary) during some `Condition` evaluations, and a few in explici

geoffthemedio:
reworked using the boost distributions

UI/ClientUI.h
```diff
@@ -234,7 +226,8 @@ class ClientUI {
     std::shared_ptr<SaveFileDialog>         m_savefile_dialog;
     std::shared_ptr<PasswordEnterWnd>       m_password_enter_wnd;   //!< the authentication window
 
-    PrefixedTextures                        m_prefixed_textures;
+    std::map<std::string, std::vector<std::shared_ptr<GG::Texture>>>
```
adrianbroher, line 233: Would be useful to document what the key refers to. Filename? Path? Prefix? Something else?

adrianbroher:
LGTM otherwise.

Issue #2983, opened 2020-05-25
AI gives up colonizing after some time


Oberlus:
Bug Report
==========

Environment
-----------
* **FreeOrion Version**: v0.4.10+ [build 2020-05-24.cfef763] CMake
  * Compiled from source
* **Operating System**: Ubuntu 18.04

Description
-----------
Check out screenshot:
![AI_gives_up_colonizing](https://user-images.githubusercontent.com/28650914/82807619-6d04e680-9e88-11ea-97c5-2c84a91a6e1d.png)

Light blue is my empire (Human). The continuous increase in colonies is from peaceful colonization, very few planet invasions.
There is no obvious or reasonable reason for the AIs to stop devoting part of their production to occupy more colonies in their safe space.

Expected Result
---------------
AIs do not stop devoting part of their production to occupy more colonies in their safe space.

Steps to reproduce
------------------
This happens in every game if played long enough.
I attach two save files from my current game, turns 92 and 184, and log files from the most recent turns up to 187.
[AI_not_colonizing.zip](https://github.com/freeorion/freeorion/files/4676882/AI_not_colonizing.zip)

Oberlus:
More on this issue:

This Trith AI did well at start but stopped expanding seemingly when it could not find any more planets within its supply. Check out screenshot. Wheeler has two interesting planets to colonize, but it would need using a colony ship. That AI did nothing but produce military for the last 100 turns, although it has not met any of its neighbours (this is a 4 AI 200 system disc galaxy).
@Morlic-fo @Cjkjvfnby Maybe AI is relying only on outpost ships and thus gets stuck?

![TrithExobotsIgnoringRadiatedSpecials](https://user-images.githubusercontent.com/28650914/82880811-34741400-9f3f-11ea-9f5d-7d98d4dfef76.png)

Cjkjvfnby:
Good question. I'll try to find time for the investigation. Feel free to ping me in a week.

Oberlus:
In another similar game I found that an Egassem AI that also stopped colonizing had an Inferno world within its supply, so maybe it is not related to relying only on outposts.

Oberlus:
Update:
After fixing #2993, AIs seems to behave _slightly_ better regarding colonization of available planets. Their graphs for number of colonies no longer becomes flat for hundreds of turns, they might get some more colonies, eventually.
[...]

PR #3018, opened 2020-06-09
Nerf flat bonuses

Oberlus:
Artificial Minds +1 research focused

Sentient Automation +3 industry focuses

[Forum discussion.](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11040&start=30#p102137)

Intended for v0.4.10


Oberlus:
Screenshots from a SP game, 200 systems, 7 AIs.

NFB_Colonies
NFB_Population
NFB_Research
NFB_Production

No sharp spikes.
The nerf might seem huge a priory but Sentient Auto. and Artif. Minds are still quite important boost thanks to the many colonies one have when researching them and also to the immediate effect after founding a colony.

geoffthemedio:
Pre- and post-change graphs would be more useful than just post-graphs, for validation / justification of the proposed changes.

Also FYI if you right click on the graph, there is a log-scale option.

Any AI issues?

Oberlus:
> Pre- and post-change graphs would be more useful than just post-graphs, for validation / justification of the proposed changes.

I'll keep that in mind next time. I didn't save graphs of the pre-change games.

> Any AI issues?

None that I can see. Edit: Updated hardcoded value for Sentient Auto.

BTW, I don't know if it is because of the Exobots everywhere or the nerf on research/industry boosts (pop. and flat) or a combination, but AIs now are quite more challenging. I am greatly pleased. They only begin dragging behind after turn 200-250, because of their reluctance to keep colonizing planets after turn 150 and focus instead on military expansion (see #2983).

PR #3030, opened 2020-06-12
Effect Condition Fixes

geoffthemedio:
-Homeworld condition wasn't setting its invariance in some cases
-A script was using a Target reference within a Conditional effect condition, which isn't allowed
-Moved test of Conditional effect condition target invariance into constructor instead of Eval

### DIFF
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -1501,8 +1501,7 @@ unsigned int Target::GetCheckSum() const {
 // Homeworld                                             //
 ///////////////////////////////////////////////////////////
 Homeworld::Homeworld() :
-    Condition(),
-    m_names()
+    Homeworld(std::vector<std::unique_ptr<ValueRef::ValueRef<std::string>>>{})
 {}
 
 Homeworld::Homeworld(std::vector<std::unique_ptr<ValueRef::ValueRef<std::string>>>&& names) :
@@ -2215,8 +2214,8 @@ HasSpecial::HasSpecial() :
                std::unique_ptr<ValueRef::ValueRef<int>>{})
 {}
 
-HasSpecial::HasSpecial(const std::string& name) :
-    HasSpecial(std::make_unique<ValueRef::Constant<std::string>>(name),
+HasSpecial::HasSpecial(std::string name) :
+    HasSpecial(std::make_unique<ValueRef::Constant<std::string>>(std::move(name)),
                std::unique_ptr<ValueRef::ValueRef<int>>{},
                std::unique_ptr<ValueRef::ValueRef<int>>{})
 {}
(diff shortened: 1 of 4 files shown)


universe/Conditions.cpp
```diff
@@ -1501,14 +1501,14 @@ unsigned int Target::GetCheckSum() const {
...
-    Condition(),
-    m_names()
+    Homeworld(std::vector<std::unique_ptr<ValueRef::ValueRef<std::string>>>{})
 {}
 
 Homeworld::Homeworld(std::vector<std::unique_ptr<ValueRef::ValueRef<std::string>>>&& names) :
     Condition(),
     m_names(std::move(names))
 {
+    // all_of should return true if m_names is empty
```
adrianbroher: Pointless comment, that behaviour is already documented in the boost algorithm  API/the described behavior according to the C++ specification.

https://www.boost.org/doc/libs/1_52_0/libs/algorithm/doc/html/algorithm/CXX11.html

> all_of and all_of_equal both return true for empty ranges, no matter what is passed to test against. When there are no items in the sequence to test, they all satisfy the condition to be tested against.

universe/Conditions.cpp
```diff
@@ -1501,14 +1501,14 @@ unsigned int Target::GetCheckSum() const {
 // Homeworld                                             //
 ///////////////////////////////////////////////////////////
 Homeworld::Homeworld() :
-    Condition(),
-    m_names()
+    Homeworld(std::vector<std::unique_ptr<ValueRef::ValueRef<std::string>>>{})
```
adrianbroher, line 1504: AFAIK `{}` should be sufficient here.

universe/Conditions.cpp
geoffthemedio: It's not obvious from the name `all_of` and I initially guessed the problem was that it should be `m_names.empty() || boost::algorithm::all_of(...)`

universe/Conditions.cpp
geoffthemedio: I thought the same, but it can't resolve the overload with the move constructor:
````
1>  c:\users\geoff\desktop\fosdk11\freeorionalt2\universe\conditions.cpp(1504): error C2668: 'Condition::Homeworld::Homeworld': ambiguous call to overloaded function
1>  c:\users\geoff\desktop\fosdk11\freeorionalt2\universe\conditions.h(250): note: could be 'Condition::Homeworld::Homeworld(Condition::Homeworld &&)'
1>  c:\users\geoff\desktop\fosdk11\freeorionalt2\universe\conditions.h(234): note: or       'Condition::Homeworld::Homeworld(std::vector<std::unique_ptr<ValueRef::ValueRef<std::string>,std::default_delete<_Ty>>,std::allocator<std::unique_ptr<_Ty,std::default_delete<_Ty>>>> &&)'
1>          with
1>          [
1>              _Ty=ValueRef::ValueRef<std::string>
1>          ]
1>  c:\users\geoff\desktop\fosdk11\freeorionalt2\universe\conditions.cpp(1505): note: while trying to match the argument list '(initializer list)'
````

PR #3031, opened 2020-06-12
Groom UniverseObject and subclasses

adrianbroher:
This PR grooms the following classes:

 * `UniverseObject`
 * `Building`
 * `Field`
 * `Fleet`

by applying the following changes:
 
  * Move primitive getters into header.
  * Remove dead code.
  * Use `//!` comments instead for documentation.
  * Rewrite some comments.
  * Use trailing return types.
  * Use east const for function parameters and return types.
  * Removing unused includes from both header and implementation file.

It also rewrites the `MovePathNode` struct to use aggregate initialization instead of user-provided constructors.

### DIFF
--- a/UI/MapWnd.cpp
+++ b/UI/MapWnd.cpp
@@ -4620,9 +4620,11 @@ void MapWnd::SetProjectedFleetMovementLine(int fleet_id, const std::list<int>& t
     // We need the route to contain the current system
     // even when it is empty to switch between non appending
     // and appending projections on shift changes
-    if (path.empty()) {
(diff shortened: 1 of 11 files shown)


geoffthemedio:
> Use east const for function parameters and return types.

I'd prefer to keep to the existing general precedent in the code of `const type&`.

adrianbroher:
To avoid reiteration of arguments: https://isocpp.org/wiki/faq/const-correctness#const-ref-alt

geoffthemedio:
I don't think the arguments for "east const" are important enough to warrant converting the code base.

Issue #3016, opened 2020-06-08
Graph Axis Label Float Precision Quirk


geoffthemedio:
![GraphAxisLabelsRounding](https://user-images.githubusercontent.com/11386908/84059288-9f4f3180-a9ba-11ea-8fec-5a3b36beb7c9.png)

The axis labels in the pedia graphs have some buggy displays of non-integer numbers when the range of what's being graphed in small.

Vezzra:
Did #3032 sufficiently address the issue? Can we close this as resolved?

PR #3032, opened 2020-06-13
Fix encyclopedia graph axis label precision.

rrobgill:
Stringstream used to specify precision for output string, to
prevent extraneous decimals when converting axis label values
to text.

Solves bug #3016

image

### DIFF
--- a/UI/GraphControl.cpp
+++ b/UI/GraphControl.cpp
@@ -197,12 +197,10 @@ void GraphControl::Render() {
         auto font = ClientUI::GetFont();
         glColor(ClientUI::TextColor());
         for (auto label : m_y_scale_ticks) {
-            std::string labeltext = "";
             std::stringstream rndlabel;
             rndlabel.precision(12);
             rndlabel << label.second;
-            labeltext += rndlabel.str();
-            font->RenderText({ul.x + GG::X1, lr.y + label.first}, labeltext);
+            font->RenderText({ul.x + GG::X1, lr.y + label.first}, rndlabel.str());
         }
     }
 
--- a/UI/GraphControl.cpp
+++ b/UI/GraphControl.cpp
@@ -196,8 +196,14 @@ void GraphControl::Render() {
         glEnable(GL_TEXTURE_2D);
         auto font = ClientUI::GetFont();
         glColor(ClientUI::TextColor());
-        for (auto label : m_y_scale_ticks)
-            font->RenderText({ul.x + GG::X1, lr.y + label.first}, boost::lexical_cast<std::string>(label.second));
+        for (auto label : m_y_scale_ticks) {
+            std::string labeltext = "";
+            std::stringstream rndlabel;
+            rndlabel.precision(12);
+            rndlabel << label.second;
+            labeltext += rndlabel.str();
+            font->RenderText({ul.x + GG::X1, lr.y + label.first}, labeltext);
+        }
     }
 
     glPushMatrix();
--- a/UI/GraphControl.cpp
+++ b/UI/GraphControl.cpp
@@ -197,10 +197,8 @@ void GraphControl::Render() {
         auto font = ClientUI::GetFont();
         glColor(ClientUI::TextColor());
         for (auto label : m_y_scale_ticks) {
-            std::stringstream rndlabel;
-            rndlabel.precision(12);
-            rndlabel << label.second;
-            font->RenderText({ul.x + GG::X1, lr.y + label.first}, rndlabel.str());
+            auto roundedlabel = boost::format("%|1$.12|") % label.second;
+            font->RenderText({ul.x + GG::X1, lr.y + label.first}, roundedlabel.str());
         }
     }
 


UI/GraphControl.cpp
```diff
@@ -196,8 +196,14 @@ void GraphControl::Render() {
...
         auto font = ClientUI::GetFont();
         glColor(ClientUI::TextColor());
-        for (auto label : m_y_scale_ticks)
-            font->RenderText({ul.x + GG::X1, lr.y + label.first}, boost::lexical_cast<std::string>(label.second));
+        for (auto label : m_y_scale_ticks) {
+            std::string labeltext = "";
+            std::stringstream rndlabel;
+            rndlabel.precision(12);
+            rndlabel << label.second;
+            labeltext += rndlabel.str();
```
geoffthemedio: is `labeltext` needed? maybe just pass `rndlabel.str()` to `RenderText`?

UI/GraphControl.cpp
```diff
@@ -196,8 +196,14 @@ void GraphControl::Render() {
         glEnable(GL_TEXTURE_2D);
         auto font = ClientUI::GetFont();
         glColor(ClientUI::TextColor());
-        for (auto label : m_y_scale_ticks)
-            font->RenderText({ul.x + GG::X1, lr.y + label.first}, boost::lexical_cast<std::string>(label.second));
```
adrianbroher, line 200: `boost::format` achieves the same in a single line:

`boost::format("%1|.1|") % label.second`

Not sure what the exact format specifier you want, as your selected `12` is random and strange and the bug report is not precise on the expected output.

https://www.boost.org/doc/libs/1_66_0/libs/format/doc/format.html#syntax

geoffthemedio: There is no precise expectation to give... it should look reasonable and mot have excessive digits or obvious float precision issues like the current code.

I'd prefer consistent numbers of digits for all numbers shown at the same time, though... so `2.0, 1.8, 1.6, 1.4, 1.2, 1.0, 0.8, 0.6, 0.4, 0.2` rather than `2, 1.8, 1.6, 1.4, 1.2, 1, 0.8, 0.6, 0.4, 0.2`.

But to be clear, there's nothing special about 2 digits, so `0.10, 0.08, 0.06, 0.04, 0.02` not (truncated) `0.1, 0.0, 0.0, 0.0, 0.0` or (rounded) `0.1, 0.1, 0.1, 0.0, 0.0`.

rrobgill: > `boost::format` achieves the same in a single line:
> 
> `boost::format("%1|.1|") % label.second`
> 
> Not sure what the exact format specifier you want, as your selected `12` is random and strange and the bug report is not precise on the expected output.
> 
> https://www.boost.org/doc/libs/1_66_0/libs/format/doc/format.html#syntax

Using that method would fix it at 1 decimal, even once the axis was denoting thousands or tens of thousands. 

The std:string method species the maximum number, excluding digits beyond 12 which is where the floating point equivalents of small integers start to diverge. Essentially rounding it to close to the integer, or approximating low numbers of decimals with similar divergence effects - once the numbers get higher, the terminal digits are excluded and the presented number has no decimal point or digits after it. Eg, in this figure labels like 20, 18, 16 etc instead of 20.0, 18.0, 16.0

image

adrianbroher: > Using that method would fix it at 1 decimal.

No it won't, the default representation is general, not fixed.

> precision (preceded by a point), sets the stream's precision
> When outputting a floatting type number, it sets the maximum number of digits
> after decimal point when in fixed or scientific mode
> in total when in default mode ('general mode', like %g)

As I said:

> Not sure what the exact format specifier you want, as your selected `12` is random and strange and the bug report is not precise on the expected output.

UI/GraphControl.cpp
rrobgill: done.

UI/GraphControl.cpp
geoffthemedio: > `boost::format("%1|.1|") % label.second`

I tried

````
            auto text = boost::format("%1|.1|") % label.second;
            font->RenderText({ul.x + GG::X1, lr.y + label.first}, text.str());
````

and various other variations of the format string, but I keep getting `boost::bad_format_string: format-string is ill-formed`

I've spent 15 min trying to find a string that works, but I keep getting the same exception and I don't want to waste more time on it.

@adrianbroher can you provide a pull request or working block of code that implements your preferred version? If not, I'll accept this pull request soon, as I care more about fixing this issue promptly and this discussion seems to have stalled.

Comment (adrianbroher) [reply UI/GraphControl.cpp:200]: > can you provide […] working block of code that implements your preferred version?

Functional equivalent:

```
// g++ -o format format.cpp && ./format
#include <iostream>
#include <sstream>
#include <boost/format.hpp>

int main()
{
    std::cout << "boost::format:" << std::endl;

    // `1$` specifies the parameter position, here first given parameter; optional in this context.
    // `.12` speficies the precision, here in general float mode; More verbose '.12g'.
    auto formatter = boost::format("%|1$.12|");
    std::cout << formatter % 9 << std::endl;
    std::cout << formatter % 1.9 << std::endl;
    std::cout << formatter % 10.9 << std::endl;
    std::cout << formatter % 100.09 << std::endl;
    std::cout << formatter % 1000.009 << std::endl;
    std::cout << formatter % 10000.0009 << std::endl;
    std::cout << formatter % 100000.00009 << std::endl;

    std::cout << std::endl << "stringstream:" << std::endl;

    std::stringstream ss;
    ss.precision(12);
    ss << 9 << std::endl;
    ss << 1.9 << std::endl;
    ss << 10.9 << std::endl;
    ss << 100.09 << std::endl;
    ss << 1000.009 << std::endl;
    ss << 10000.0009 << std::endl;
    ss << 100000.00009 << std::endl;
    std::cout << ss.str() << std::flush;
}
```

Output
```
$ g++ -o format format.cpp && ./format 
boost::format:
9
1.9
10.9
100.09
1000.009
10000.0009
100000.00009

stringstream:
9
1.9
10.9
100.09
1000.009
10000.0009
100000.00009
```

Both versions will fall over to scientific notation as soon as the float exceeds 12 digits.

Comment (rrobgill) [reply UI/GraphControl.cpp:200]: The floats being returned by label in the program are above 12 digits in length (See screenshot in #3016). It is beyond that the errors are introduced due to the nature of ieee float representation

Stringstream rounds at the specified point preventing the problem. Does the boost library handle it in the same way?

Comment (adrianbroher) [reply UI/GraphControl.cpp:200]: > Does the boost library handle it in the same way?

`boost::format` uses the same standard library I/O stream interface as `std::stringstream`, why should it behave differently?

Comment (rrobgill) [reply UI/GraphControl.cpp:200]: No particular reason that it should be different. But if it would convert those same floating point numbers from the screenshot into scientific notation, rather than rounding them then it would have to be handling it differently.

Comment (adrianbroher) [reply UI/GraphControl.cpp:200]: > But if it would convert those same floating point numbers from the screenshot into scientific notation, rather than rounding them then it would have to be handling it differently.

If rounding occurs your platform then behaves differently and the outcome is not reliable behaviour accross systems.  Adopting the previous code to precision `.1` leads to scientific notation with GNU C++ library:

```
boost::format:
9
2
1e+01
1e+02
1e+03
1e+04
1e+05

stringstream:
9
2
1e+01
1e+02
1e+03
1e+04
1e+05
```

PR #3037, opened 2020-06-14
Increase pop when Population Policy is applied

rrobgill:
* Increase rate of change to population by 50%,
  if population on planet is lower than target population.

* No change to rate if population is decreasing


Oberlus:
Population growth is controlled in https://github.com/freeorion/freeorion/blob/ed213fba84d31196934ccf43c1a54d2246e73668/default/scripting/species/common/population.macros#L29
It already considers the case on negative growth and does not apply the policy effect in that case.
This PR should not be merged.

geoffthemedio:
> This PR should not be merged.

Unless @rrobgill has some other insight I've missed, I agree...

rrobgill:
can't believe i missed that. Sorry.

geoffthemedio:
Scripts can be in unexpected places for various reasons.

PR #3039, opened 2020-06-14
Colonization policy gives +3 base fuel to colonization-only ships.

geoffthemedio:
As proposed [here](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11444&start=15#p102203) or earlier in the thread, makes the colonization policy give bonus fuel to colony ships. Does not affect armed ships.


agrrr3:
Looks good to me. Will certainly try for my next distributed empire

Oberlus:
As commented in the thread, I would rather make a new policy for the extra fuel in colony ships, with only +2 fuel, separated from the cheaper colony policy.

PR #3041, opened 2020-06-14
Fix typo in condition to make Military Command empire-unique

### LINKED ISSUE #3040: Military Command can be constructed more than once
Bug Report
==========

Environment
-----------
* **FreeOrion Version**:  ed213fb + b62be60
* **Fetched as** 
  * Compiled from source

Description
-----------
Military Command building, which grants an extra military policy slot, can be constructed in more than one planet.
I haven't tested, but it seems constructing more than one would make no effect (due to the stacking group MILITARY_COMMAND_SLOT_METER_EFFECT)
This is probably a bug in the reference to the building name ("_DESC" is wrong): https://github.com/freeorion/freeorion/blob/28d3d6458d255e039010c967e211cb4961cee86b/default/scripting/buildings/MILITARY_COMMAND.focs.txt#L13

Expected Result
---------------
Since the effect is not supply-bound and the fluff/name is about headquarters, the building should be unique for each empire, as the Imperial Palace.

Oberlus:
Fixes #3040

### DIFF
--- a/default/scripting/buildings/MILITARY_COMMAND.focs.txt
+++ b/default/scripting/buildings/MILITARY_COMMAND.focs.txt
@@ -10,7 +10,7 @@ BuildingType
         Not Contains Building name = "BLD_MILITARY_COMMAND_DESC"
         TargetPopulation low = 1
         Number low = 0 high = 0 condition = And [
-            Building name = "BLD_MILITARY_COMMAND_DESC"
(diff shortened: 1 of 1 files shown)


geoffthemedio:
already fixed https://github.com/freeorion/freeorion/commit/6a1035d8f372ff463c056fe9b9fd490a00bbdc84 (including the 2nd one you didn't...)

PR #3000, opened 2020-06-02
Reference Named ValueRefs in stringtables

agrrr3:
This implements a solution to [A single source for default values - forum thread](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11427)

This adds 
- a configuration file `default/scripting/common/named_values.focs.txt` for defining named value refs
- a `NamedInteger` definition in that file that takes a `string` name and a `ValueRef<int>` value as parameter.
- a `NamedReal` definition in that file that takes a `string` name and a `ValueRef<double>` value as parameter.
- a `NamedEnvironment` definition in that file that takes a `string` name and a `ValueRef<PlanetEnvironment>` value as parameter. (this and the following NamedPlanetType are mostly proof-of-concept how this works with enums).
- a `NamedPlanetType` definition in that file that takes a `string` name and a `ValueRef<PlanetType>` value as parameter.
- a encyclopedia `[Test - Named Value Refs]` entry which lists all the named value refs in the game (name/ID, registration type, stringtable entry, calculation, calculated value, and pattern of invariances)

Also in normal focs and macro files one can use:
- a `NamedRef` ValueRef called `Lookup` in FOCS which takes a string name as parameter. This can be used to look up and eval a registered value ref.
- a name-in-the-middle-ValueRef definition named `NamedReal` which takes a `string` name and a `ValueRef<double>` value as parameters. This registers the given value ref and inserts a `NamedRef`instead.
- a name-in-the-middle-ValueRef definition named `NamedInteger` which takes a `string` name and a `ValueRef<integer>` value as parameters. This registers the given value ref and inserts a `NamedRef`instead.

Currently use is showcased for fuel and reinforced hull techs. The named_values.focs.txt is used for the reinforced hull bonus (although that tech should use a name-in-the-middle valuref as that tech is the single place where the value is used). The name-in-the-middle functionality is shown for fuel tech in `fuel.macros`-using named_values one would have to make an explicit entry for each of the techs.
```
default/scripting/techs/ship_parts/damage_control/SHP_REINFORCED_HULL.focs.txt:            effects = SetMaxStructure value = Value + Lookup name = "SHP_REINFORCED_HULL_BONUS"
default/scripting/common/named_values.focs.txt:NamedReal name = "SHP_REINFORCED_HULL_BONUS" value = (5 * [[SHIP_STRUCTURE_FACTOR]]) 
```

[...]

PR #3046, opened 2020-06-15
ValueRef Optimization

geoffthemedio:
- determine ValueRef invariance in constructors
- determine whether an Operation is a simple increment in the constructor

### DIFF
--- a/universe/ValueRef.h
+++ b/universe/ValueRef.h
@@ -9,10 +9,37 @@
 
 namespace ValueRef {
 
-/** The base class for all ValueRef classes.  This class provides the public
-  * interface for a ValueRef expression tree. */
+//! The common base class for all ValueRef classes. This class provides
+//! some the return-type-independent interface.
+struct FO_COMMON_API ValueRefBase {
+    ValueRefBase() = default;
+    virtual ~ValueRefBase() = default;
+
+    bool RootCandidateInvariant() const  { return m_root_candidate_invariant; }
+    bool LocalCandidateInvariant() const { return m_local_candidate_invariant; }
+    bool TargetInvariant() const         { return m_target_invariant; }
+    bool SourceInvariant() const         { return m_source_invariant; }
+    virtual bool SimpleIncrement() const { return false; }
+    virtual bool ConstantExpr() const    { return false; }
+
+    virtual std::string Description() const = 0;                    //! Returns a user-readable text description of this ValueRef
+    virtual std::string Dump(unsigned short ntabs = 0) const = 0;   //! Returns a textual representation that should be parseable to recreate this ValueRef
+
+    virtual void SetTopLevelContent(const std::string& content_name) {}
+
+    virtual unsigned int GetCheckSum() const { return 0; }
+
+protected:
+    bool m_root_candidate_invariant = false;
+    bool m_local_candidate_invariant = false;
+    bool m_target_invariant = false;
+    bool m_source_invariant = false;
+};
+
+//! The base class for all ValueRef classes returning type T. This class
+//! provides the public interface for a ValueRef expression tree.
 template <typename T>
-struct FO_COMMON_API ValueRef
+struct FO_COMMON_API ValueRef : public ValueRefBase
 {
     ValueRef() = default;
     virtual ~ValueRef() = default;
@@ -32,41 +59,6 @@ struct FO_COMMON_API ValueRef
       * to objects such as the source, effect target, or condition candidates
       * that exist in the tree. */
     virtual T Eval(const ScriptingContext& context) const = 0;
-
-    bool RootCandidateInvariant() const
-    { return m_root_candidate_invariant; }
-
-    bool LocalCandidateInvariant() const
-    { return m_local_candidate_invariant; }
-
-    bool TargetInvariant() const
-    { return m_target_invariant; }
-
-    bool SourceInvariant() const
-    { return m_source_invariant; }
-
-    virtual bool SimpleIncrement() const
-    { return false; }
-
-    virtual bool ConstantExpr() const
-    { return false; }
-
-    virtual std::string Description() const = 0;
-
-    /** Returns a text description of this type of special. */
-    virtual std::string Dump(unsigned short ntabs = 0) const = 0;
-
-    virtual void SetTopLevelContent(const std::string& content_name)
-    {}
-
-    virtual unsigned int GetCheckSum() const
-    { return 0; }
-
-protected:
-    bool m_root_candidate_invariant = false;
-    bool m_local_candidate_invariant = false;
-    bool m_target_invariant = false;
-    bool m_source_invariant = false;
 };
 
 enum StatisticType : int {
--- a/universe/ValueRef.h
+++ b/universe/ValueRef.h
@@ -14,8 +14,8 @@ namespace ValueRef {
 template <typename T>
 struct FO_COMMON_API ValueRef
 {
-    virtual ~ValueRef()
-    {}
+    ValueRef() = default;
+    virtual ~ValueRef() {}
 
     virtual bool operator==(const ValueRef<T>& rhs) const;
 
@@ -33,17 +33,17 @@ struct FO_COMMON_API ValueRef
       * that exist in the tree. */
     virtual T Eval(const ScriptingContext& context) const = 0;
 
-    virtual bool RootCandidateInvariant() const
-    { return false; }
+    bool RootCandidateInvariant() const
+    { return m_root_candidate_invariant; }
 
-    virtual bool LocalCandidateInvariant() const
-    { return false; }
+    bool LocalCandidateInvariant() const
+    { return m_local_candidate_invariant; }
 
-    virtual bool TargetInvariant() const
-    { return false; }
+    bool TargetInvariant() const
+    { return m_target_invariant; }
 
-    virtual bool SourceInvariant() const
-    { return false; }
+    bool SourceInvariant() const
+    { return m_source_invariant; }
 
     virtual bool SimpleIncrement() const
     { return false; }
@@ -61,6 +61,12 @@ struct FO_COMMON_API ValueRef
 
     virtual unsigned int GetCheckSum() const
     { return 0; }
+
+protected:
+    bool m_root_candidate_invariant = false;
+    bool m_local_candidate_invariant = false;
+    bool m_target_invariant = false;
+    bool m_source_invariant = false;
 };
 
 enum StatisticType : int {
--- a/universe/ValueRef.h
+++ b/universe/ValueRef.h
@@ -15,7 +15,7 @@ template <typename T>
 struct FO_COMMON_API ValueRef
 {
     ValueRef() = default;
-    virtual ~ValueRef() {}
+    virtual ~ValueRef() = default;
 
     virtual bool operator==(const ValueRef<T>& rhs) const;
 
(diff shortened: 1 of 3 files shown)


geoffthemedio:
@adrianbroher review would be appreciated

universe/ValueRef.h
```diff
@@ -14,8 +14,8 @@ namespace ValueRef {
 template <typename T>
 struct FO_COMMON_API ValueRef
 {
-    virtual ~ValueRef()
-    {}
+    ValueRef() = default;
+    virtual ~ValueRef() {}
```
adrianbroher: Please use `default` for the destructor.

universe/ValueRefs.h
```diff
@@ -710,7 +695,21 @@ Statistic<T>::Statistic(std::unique_ptr<ValueRef<T>>&& value_ref, StatisticType
     m_stat_type(stat_type),
     m_sampling_condition(std::move(sampling_condition)),
     m_value_ref(std::move(value_ref))
-{}
+{
+    this->m_root_candidate_invariant = (!m_sampling_condition || m_sampling_condition->RootCandidateInvariant()) &&
```
adrianbroher, line 681: For clarification: you are leaving out the `Variable<T>::*Invariant` call because `Statistic<T>` initializes `Variable<T>` with `NON_OBJECT_REFERENCE`, right?

universe/ValueRefs.h
```diff
@@ -1114,7 +1079,31 @@ ComplexVariable<T>::ComplexVariable(S&& variable_name,
     m_int_ref3(std::move(int_ref3)),
     m_string_ref1(std::move(string_ref1)),
     m_string_ref2(std::move(string_ref2))
-{}
```
adrianbroher, line 1117: Can we merge those constructors somehow?

universe/ValueRefs.h
```diff
@@ -1695,34 +1633,53 @@ Operation<T>::Operation(OpType op_type, std::vector<std::unique_ptr<ValueRef<T>>
...
     CacheConstValue();
 }
 
 template <typename T>
-void Operation<T>::DetermineIfConstantExpr()
+void Operation<T>::InitConstInvariants()
 {
     if (m_op_type == RANDOM_UNIFORM || m_op_type == RANDOM_PICK) {
         m_constant_expr = false;
+        this->m_root_candidate_invariant = false;
```
adrianbroher, line 1588: Is there anything shadowing `m_root_candidate_invariant` and friends?  If not I don't see a need for `this->`.

universe/ValueRefs.h
```diff
@@ -1695,34 +1633,53 @@ Operation<T>::Operation(OpType op_type, std::vector<std::unique_ptr<ValueRef<T>>
...
+        [](const auto& operand) { return operand && operand->RootCandidateInvariant(); });
+    this->m_local_candidate_invariant = std::all_of(m_operands.begin(), m_operands.end(),
+        [](const auto& operand) { return operand && operand->LocalCandidateInvariant(); });
+    this->m_target_invariant = std::all_of(m_operands.begin(), m_operands.end(),
+        [](const auto& operand) { return operand && operand->TargetInvariant(); });
+    this->m_source_invariant = std::all_of(m_operands.begin(), m_operands.end(),
+        [](const auto& operand) { return operand && operand->SourceInvariant(); });
+
+    // determine if this is a simple incrment operation
+    if (m_op_type != PLUS && m_op_type != MINUS)
```
adrianbroher: While this seems to be correct it is strange to read.  The previous code assigned some values to variables based on conditions.  Here you are leaving the function early based on conditions, the previous implementation had at least return true or false, which made it easier to grasp.  I would suggest to change it either to:

* Set `m_simple_increment` to false before each return statement.
* Assign the combination of all conditions to `m_simple_increment` with positive notation.

universe/ValueRefs.h
```diff
@@ -1695,34 +1633,53 @@ Operation<T>::Operation(OpType op_type, std::vector<std::unique_ptr<ValueRef<T>>
...
+        return;
+    if (m_operands.size() < 2 || !m_operands[0] || !m_operands[1])
```
adrianbroher: This allows 3 and more operands.  Can those still be a simple increment?

universe/ValueRefs.h
geoffthemedio: GCC won't compile without it, in the header-defined cases.

adrianbroher: Can you be more specific than "won't compile"?

universe/ValueRefs.h
geoffthemedio: 3+ operands shouldn't occur, at least with `PLUS` or `MINUS`, and would be ignored if passed. So it would still be a simple increment if it happened, I guess...

universe/ValueRefs.h
geoffthemedio: How about explicitly default initial setting `m_simple_increment = false;` above this block?

universe/ValueRefs.h
geoffthemedio: Yes, the result `true` is already know and doesn't matter.

universe/ValueRefs.h
geoffthemedio: https://travis-ci.org/github/freeorion/freeorion/jobs/698338097#L368

adrianbroher: Oh, I wasn't aware of this behaviour: https://isocpp.org/wiki/faq/templates#nondependent-name-lookup-members

universe/ValueRefs.h
geoffthemedio: With the `const char* variable_name` overload or to avoid the perfect forwarding template?

It could take a `string` by value and always move but as far as I understand that's not optimal if passed an lvalue.

adrianbroher: The main gripe with this is the duplicated code for calculating the invariance, not the parameters.

universe/ValueRefs.h
adrianbroher: That would be better than the current implementation, but still no good IMO.

universe/ValueRefs.h
geoffthemedio: Moved into an init function and added a common `ValueRefBase` struct to facilitate ranged processing.

(`Condition` could also share the same base class which would similarly clean up some condition invariant init code.)

universe/ValueRef.h
```diff
@@ -9,10 +9,37 @@
 
 namespace ValueRef {
 
-/** The base class for all ValueRef classes.  This class provides the public
-  * interface for a ValueRef expression tree. */
+//! The common base class for all ValueRef classes. This class provides
+//! some the return-type-independent interface.
+struct FO_COMMON_API ValueRefBase {
```
adrianbroher, line 14: This class is named horrible and I doubt that it should exist in the first place. Why do you do that? Why would you need that?

geoffthemedio: So different `ValueRef<T>` can have a common base interface and be stored in a simple container of pointers-to-base. This allows invariance tests on a range of valuerefs of different types (int, string, double) to be simplified with a single range `all_of` call.

It could also be used as a base class / interface for `Condition` so they could also be stored in the same container with valuerefs, further simplifying the invariance determination code.

It would probably also allow some other code that passes around a `ValueRef` but doesn't need to know what type is returned to do so with a less bulky #include 

It also [avoids redundant instantiation](https://books.google.de/books?id=m2cnCgAAQBAJ&pg=PA19) of all of the functions that were moved from the templated class into the non-templated base class, which [don't depend on the template](https://blog.feabhas.com/2014/06/template-inheritance/) type (eg. simple getters of non-template-dependent member bool flags).

@agrrr3 also wants to do something quite similar: https://github.com/freeorion/freeorion/pull/3000/files#diff-b516c63990c167d4ef5caaea60ce33b0

The name can be changed, but I think `ValueRefBase` sums it up pretty well, though not if it is also used as a base for `Condition`.

universe/ValueRefs.h
geoffthemedio: hopefully acceptable now?

universe/ValueRef.h
geoffthemedio: @adrianbroher any further comments about this?

adrianbroher: > So different ValueRef<T> can have a common base interface and be stored in a simple container of pointers-to-base. This allows invariance tests on a range of valuerefs of different types (int, string, double) to be simplified with a single range all_of call.

> It would probably also allow some other code that passes around a ValueRef but doesn't need to know what type is returned to do so with a less bulky #include

And then you have to distinguish them again with expensive casts when you want to evaluate the value of the expression.  I don't see any gain here.  `ValueRef`s are always bound to the domain with a certain type (e.g. the ships hull structure is always expressed as floating point number, a systems star type is expressed as `StarType` enum, etc).  There is no common ground between those in their nature that could be abstracted out.

> It could also be used as a base class / interface for Condition so they could also be stored in the same container with valuerefs, further simplifying the invariance determination code.

If you want to avoid duplicated code because some common semantic exists, which is not backed by the domain a mixin would be the tool of choice.  This is different to a base class as it encapsulates a concept away without using a base class for, by domain, unrelated classes.

> The name can be changed, but I think ValueRefBase sums it up pretty well, though not if it is also used as a base for Condition.

`focs::Expression` but as mentioned before the idea of a typeless expression is pointless.

geoffthemedio: > > It would probably also allow some other code that passes around a ValueRef but doesn't need to know what type is returned to do so with a less bulky #include
> 
> And then you have to distinguish them again with expensive casts when you want to evaluate the value of the expression.

That would be a case where you do need to know what the type is. If you need to evaluate a `ValueRef`, you probably wouldn't pass it as a base non-evaluable type pointer, and that's not what I'm referring to.

> I don't see any gain here. `ValueRef`s are always bound to the domain with a certain type (e.g. the ships hull structure is always expressed as floating point number, a systems star type is expressed as `StarType` enum, etc). There is no common ground between those in their nature that could be abstracted out.

This pull request contains examples in which several `ValueRef`s of different types are passed and treated the same way and where it's convenient to store them in a single container. The `ValueRefBase` class shows the commonality between `ValueRef`s of different type specialization (the functions and storage that don't depend on the template type) which is used by that code. Specifically: when determining if a `Condition` or `ValueRef` that has subconditions or subvaluerefs is Source/Target/LocalCandidate/RootCandidate invariant, the template type doesn't matter and the `ValueRef`s aren't evaluated.

It doesn't matter if the sub-expressions are `Condition` or `ValueRef<T>` or what `T` is; these all share the same invariance properties. That the `Condition` or `ValueRef<T>` can be evaluated on object sets or to get a number, string, or enum doesn't matter in that code, as that code doesn't evaluate them.

> > It could also be used as a base class / interface for Condition so they could also be stored in the same container with valuerefs, further simplifying the invariance determination code.
> 
> If you want to avoid duplicated code because some common semantic exists, which is not backed by the domain a mixin would be the tool of choice. This is different to a base class as it encapsulates a concept away without using a base class for, by domain, unrelated classes.

So you propose having an `InvarianceInfo` struct that `Condition` and `ValueRef<T>` have member variables of, and when doing the invariance tests, use a list of pointers to `InvarianceInfo`, which is initialized by `value_ref->GetInvarianceInfo()` and `condition->GetInvarianceInfo()` calls? I suppose that could work, but it seems awkward compared to storing the base class pointers. It also doesn't avoid instantiating the common invariance getters that don't depend on the template type, which was another advantage of the common base class pattern.

adrianbroher: Do whatever you want then.  I will push an improvement PR then when I have time.

PR #2910, opened 2020-04-29
Add local variables to python exception logging

Morlic-fo:
This PR extracts the local variables from the last traceback frame and adds them to the log record. Makes it a bit easier to understand the context of errors in reports where a savegame can not be supplied.

Example log output:
```
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 : Exception division by zero occurred during generateOrders
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 : Traceback (most recent call last):
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 :   File "C:/FOPy3\32\FreeOrion\default\python\AI\FreeOrionAI.py", line 70, in _error_handler
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 :     res = func(*args, **kwargs)
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 :   File "C:/FOPy3/32/FreeOrion/default/python\common\listeners.py", line 42, in wrapper
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 :     res = funct(*args)
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 :   File "C:/FOPy3\32\FreeOrion\default\python\AI\FreeOrionAI.py", line 369, in generateOrders
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 :     failing_func(2)
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 :   File "C:/FOPy3\32\FreeOrion\default\python\AI\FreeOrionAI.py", line 382, in failing_func
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 :     print(x/z)  # throws for x = 2
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 : ZeroDivisionError: division by zero
21:55:21.173441 {0x00002a04} [error] python : FreeOrionAI.py:73 : Local context: {'y': 2, 'z': 0, 'x': 2}
```

PR #3049, opened 2020-06-15
Ensure python log messages end with \n

Morlic-fo:
Messages must end with newline to ensure log flushing.

Fixes defect introduced in #2910.

### DIFF
--- a/default/python/common/configure_logging.py
+++ b/default/python/common/configure_logging.py
@@ -158,7 +158,9 @@ class _LoggerHandler(logging.Handler):
 
     def emit(self, record):
         msg = self.format(record)
-        self.logger(msg, str(record.filename), record.lineno)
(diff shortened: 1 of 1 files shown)


default/python/common/configure_logging.py
```diff
@@ -158,7 +158,9 @@ def __init__(self, level):
 
     def emit(self, record):
         msg = self.format(record)
-        self.logger(msg, str(record.filename), record.lineno)
+        if not msg.endswith("\n"):
```
Cjkjvfnby, line 161: IMHO it will be nice to leave a comment in the code about the reason.

Morlic-fo: https://github.com/freeorion/freeorion/commit/e405949c3a44d0b06d5f93e521dead5e361537c8

PR #3050, opened 2020-06-15
Fix Python logging memory leak

### LINKED ISSUE #3038: AI High Memory Usage
Bug Report
There appears to be an AI memory lead in recent v0.4.10+ builds including v0.4.10+ build 2020-06-13.ed213fb as reported here: https://www.freeorion.org/forum/viewtopic.php?f=9&t=11700&p=102260#p102260

Reloading a save reportedly substantially reducing the AI memory usage.

Morlic-fo:
The stringstream buffer was never cleared and eventually contained the entire log.

Note that this memory leak was recently amplified by malformed logs (see #3049) but already present in older versions. @Vezzra: This should subsequently be cherry-picked into release.

Resolves #3038

### DIFF
--- a/python/LoggingWrapper.cpp
+++ b/python/LoggingWrapper.cpp
@@ -33,6 +33,7 @@ namespace {
 
         // If ss is good, store any partial line of text for the next call to send_to_log.
         if (ss.eof()) {
+            ss.str("");
             ss.clear();
             ss << line;
         }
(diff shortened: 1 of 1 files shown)


geoffthemedio:
> The stringstream buffer was never cleared

There are `ss.clear()` right after your added lines, so I'm confused by the description and the fix...

Cjkjvfnby:
> There are ss.clear() right after your added lines, so I'm confused by the description and the fix...

Confusing names are meant to confuse. 
http://www.cplusplus.com/reference/ios/ios/clear/

PR #3051, opened 2020-06-15
Use non-intrusive serialization for util/MultiplayerCommon.h structs

adrianbroher:
This PR removes the need for including any serialization related headers to `util/MultiplayerCommon.h` or exposing the implementation of the to-be-serialized classes to `util/Serialize.h` by using the non-intrusive `boost::serialization` interface.

### DIFF
--- a/util/MultiplayerCommon.h
+++ b/util/MultiplayerCommon.h
@@ -5,8 +5,8 @@
 #include "../network/Networking.h"
 #include "Export.h"
 #include "OptionsDB.h"
+#include "OrderSet.h"
 #include "Pending.h"
-#include "Serialize.h"
 
 #include <GG/Clr.h>
 #include <GG/ClrConstants.h>
@@ -15,7 +15,6 @@
 #include <set>
 #include <vector>
 #include <map>
-#include <boost/serialization/access.hpp>
 #include <boost/date_time/posix_time/posix_time_types.hpp>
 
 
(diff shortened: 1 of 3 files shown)


adrianbroher:
Fixed linker error on MSVC caused by the mismatch between struct and class of the forward declaration and struct definition.

geoffthemedio:
With this applied, the server disconnects the player after pressing turn. Server log has
````
11:38:12.442736 {0x00006f98} [debug] server : message.cpp:990 : deserializing orders
11:38:12.442736 {0x00006f98} [debug] server : message.cpp:992 : checking for ui data
11:38:12.442736 {0x00006f98} [debug] server : message.cpp:995 : deserializing UI data
11:38:12.443737 {0x00006f98} [error] server : serializemultiplayercommon.cpp:134 : Deserializing ship design ids and obsoletes failed. Skipping hull order and obsoletion, and obsolete ship parts.
11:38:12.443737 {0x00006f98} [error] server : message.cpp:1006 : ExtractTurnOrdersMessageData(const Message& msg, ...) failed! Message probably long, so not outputting to log.
Error: input stream error
11:38:12.443737 {0x00006f98} [debug] FSM : serverfsm.cpp:303 : ServerFSM::HandleNonLobbyDisconnection : Lost connection to player #1, named "Human_Player".
````

o01eg:
`Input stream error` could be caused by desync of server and client protocols. Could you catch it in the debugger and see stacktrace?

adrianbroher:
> With this applied, the server disconnects the player after pressing turn.

Updated implementation to make `BOOST_CLASS_VERSION` visible for invocations of `BOOST_SERIALIZATION_NVP` at other locations like the network `Message` factory functions.

Also replaced the catch-all exception block with a more concrete `std::exception` catch block for better diagnostic messages.

PR #3054, opened 2020-06-16
Fields affect target meter, add standard meter change for ships

### LINKED ISSUE #3048: Nanite Swarm and Meteor Blizzard PP/RP harvesting effects do not expire outside the fields
Bug Report
==========

Environment
-----------
* **FreeOrion Version**: v0.4.10+ build 2020-06-14 9af9a91 CMake
* **Operating System**: Ubuntu 18.04
* **Fetched as** 
  * Compiled from source

Description
-----------
The haversting PP effect of the Nanite Swarm field does not fade off outside of the field, making ships constant producers of PP for the rest of the game.

Check out the ships at Watt or at Perelandra:
NaniteSwarmPPNotFading

Expected Result
---------------
The effect disappears after leaving the area of effect of the field.

[logs.zip](https://github.com/freeorion/freeorion/files/4780877/logs.zip)
[FreeOrion_Oberlus_Abyss_0129_132.zip](https://github.com/freeorion/freeorion/files/4780878/FreeOrion_Oberlus_Abyss_0129_132.zip)

Edit: the same happens with the Meteor Blizzard field that grants RPs to the ships.

Oberlus:
Makes Nanite and Meteor fields affect Target meters of ships (not monsters) instead of current meter.
Adds standard meter growth/decay for ships.

Fixes #3048 

Tested with the save file in the linked issue. Works fine for ships with species.
Monsters in that save that already had PP output won't lose it outside of fields, but new monsters won't get the resource output with this fix.

### DIFF
--- a/default/scripting/fields/FLD_METEOR_BLIZZARD.focs.txt
+++ b/default/scripting/fields/FLD_METEOR_BLIZZARD.focs.txt
@@ -29,10 +29,11 @@ FieldType
             scope = And [
                 Ship
                 WithinDistance distance = Source.Size * 0.9 condition = Source
+                Not Monster
             ]
             stackinggroup = "METEOR_BLIZZARD_SHIP_EFFECTS"
             effects = [
-                SetResearch value = max(1, Value)
+                SetTargetResearch value = Value + 1
             ]
 
         EffectsGroup    // after reaching a certain age, dissipate when small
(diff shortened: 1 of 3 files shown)


TheSilentOne1:
Thank you, Oberlus, for taking something on I would have felt responsible for.

agrrr3:
Hm, the ships which had output should also shrink, shouldn't they?
The max meters should be zero

geoffthemedio:
> Hm, the ships which had output should also shrink, shouldn't they?
> The max meters should be zero

They are, but for monsters don't have a species, so don't get the effect from the `STANDARD_METER_GROWTH` macro used in species definitions.

PR #3055, opened 2020-06-18
Build gas giant generators only in inhabited systems

rrobgill:
As their effect is limited to the system they are in, they don't do anything if there are no planets within the system

* Changes location specification so that there must be an inhabited planet in the same system

* Encyclopedia entry text updated


Oberlus:
What if the player is building them in advance? Maybe the player is moving a colony to that system, or is building colonies in some of the planets already there.

I don't see the gain from this change.

geoffthemedio:
> I don't see the gain from this change.

The gain would presumably be players not accidentally or unknowingly producing the building where it isn't useful, but you've also identified cases where they'd want to pre-build it which is also reasonable... so I'm inclined to reject.

PR #3056, opened 2020-06-18
Require shipyard (built or enqueued) in system to build asteroid processor

rrobgill:
As their effect is limited to the system they are in, they don't do anything if there are no shipyards within the system

* Changes location specification to that there must be a shipyard
  (either built or enqueued) within the same system

* Encyclopedia entry text updated

### DIFF
--- a/default/scripting/buildings/shipyards/ASTEROID.focs.txt
+++ b/default/scripting/buildings/shipyards/ASTEROID.focs.txt
@@ -8,6 +8,17 @@ BuildingType
         Not Contains Building name = "BLD_SHIPYARD_AST"
         Planet type = Asteroids
         OwnedBy empire = Source.Owner
+        ContainedBy And [
+            System
+            Contains And [
+                Planet
+                OwnedBy empire = Source.Owner
+                Or [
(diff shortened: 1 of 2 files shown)


Oberlus:
Their effect is not limited to the system they are in: supply connected Asteroid Processors are required to mount the asteroid armor parts.

Edit: the actual requirement for the asteroid armor parts is BLD_SHIPYARD_AST_REF, but that one in turn requires BLD_SHIPYARD_AST.

geoffthemedio:
@Oberlus so does this pull request make sense or not?

Oberlus:
IMO, it could be annoying more than helpful.
I sometimes go for organics with asteroid armor, so I don't necessarily build shipyards in the same system I have the asteroid reformation processor.

The only case in which I think this change (PR) could be helpful is when a rookie player builds the asteroid shipyards in a system without the shipyard because s/he wants to build asteroid hulls there. S/he will learn "the lesson" at that time and not incur into the same mistake again, but having to build the shipyard in an asteroid system (maybe a system with just an asteroid belt on it) where you won't be building any ships would be annoying every time.

I would reject this PR.

PR #3058, opened 2020-06-18
Remove CombatLogManager pointer-to-implementation pattern.

adrianbroher:
This PR removes the PIMPL pattern from the `CombatLogManager` class. PIMPLs are usually used to separte complex implementation dependencies from the class declaration.  This is not true for the `CombatLogManager` and makes the introduction of  non-intrusive serialization harder than necessary.


geoffthemedio:
This breaks combat logs for me. I get a sitrep saying there was a combat, but clicking the link does nothing. Setting a breakpoint, it appears that the combat logs are empty.

Without this applied, combat log links in sitreps work.

adrianbroher:
Undone the `std::map<>::insert(begin, end)` calls.

PR #3065, opened 2020-06-22
Use non-intrusive serialization for combat related classes

adrianbroher:
This PR removes the need for including any serialization related headers to `combat/CombatLogManager.h` by using the non-intrusive `boost::serialization` interface.


adrianbroher:
Added missing function forward declaration.

Issue #3059, opened 2020-06-18
Incorrect amount of troops on outposts


l29ah:
Bug Report
==========

Environment
-----------

@o01eg 's server, should be e352120

Description
-----------

Player `swaq` has evacuated all his planets, and the troops went to 13 on the outposts as expected. But three turns ago they've grown to 26 for some unknown reasons. Outposts of other players are at 13 as expected. The meter says bullshit:
![2020-06-18-224701_555x360_scrot](https://user-images.githubusercontent.com/31852/85065389-bf9a9f80-b1b5-11ea-88f1-59f8c6e8c5bf.png)

Expected Result
---------------

13 troops

Steps to reproduce
------------------

dunno

o01eg:
Looks like his troops aren't affected by outpost effect.

geoffthemedio:
As far as I can tell, the output troop effect would come from [this script](https://github.com/freeorion/freeorion/blob/master/default/scripting/species/common/troops.macros#L67) which is included into various species through some macros. If there are no valid source objects that actually exist, but your client thinks there is one due to visibility limitations, maybe that could cause this? Not sure why this would happen unless there are also no ships in the empire.

Might help to move the effect to a tech so that it doesn't depend on a planet with population being owned by the empire, if that's the case...

o01eg:
>  If there are no valid source objects that actually exist, but your client thinks there is one due to visibility limitations, maybe that could cause this?

@geoffthemedio  Is there something that couldn't be seen by L29Ah's empire that cause this effect no being applied to swaq's outposts?

geoffthemedio:
Player B's empire seeing or not seeing something wouldn't cause the server effects evaluation to do anything different to player A's empire.

But if player B previously observed some objects that player A owned that could act as a source for an effect, but player B now doesn't know that those objects have been destroyed or otherwise changed since player B last had visibility of them, then player B's client might think that those those objects could act as a source for an effect. Player B's client would then calculate effects accounting for player A's outpost assuming that those source objects still exist. But on the server, no such object exists, and thus the effect isn't applied. The final meter totals that the server calculates are sent to player B, but these don't match what player B's client expects based on what player B has observed in the universe, so there is then an "unknown" contribution to the meter accounting for player A's objects in player B's client.

[...]

PR #3068, opened 2020-06-22
Treat empire without population and ships as eliminated

o01eg:
Forum thread: https://freeorion.org/forum/viewtopic.php?f=6&t=11708

Workaround for https://github.com/freeorion/freeorion/issues/3059


geoffthemedio:
Can be revised if a preferred solution is provided, but for now this is probably better.

PR #3070, opened 2020-06-23
Add CMake GiGi::header import target to simplify headless build

adrianbroher:
This PR adds a CMake INTERFACE target to GG, called header.  This target contains all build settings and source references required to use common classes like `GG::Clr` or the `GG_ENUM` macro within the `freeorioncommon` targets without the need to depend on the GG library target itself.  This simplifies the `headless` build logic somewhat and is a prerequisite for Docker-less TravisCI builds.


geoffthemedio:
Want feedback or just merge? @o01eg ?

o01eg:
CI passed so I'm fine with this PR.

PR #3079, opened 2020-06-27
Exclude data for universe generator from checksum

o01eg:
This allows to tune universe generator without desyncing checksum with client.

### DIFF
--- a/universe/Special.cpp
+++ b/universe/Special.cpp
@@ -167,8 +167,6 @@ unsigned int Special::GetCheckSum() const {
     CheckSums::CheckSumCombine(retval, m_description);
     CheckSums::CheckSumCombine(retval, m_stealth);
     CheckSums::CheckSumCombine(retval, m_effects);
-    CheckSums::CheckSumCombine(retval, m_spawn_rate);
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I'm uncertain about whether this makes sense. The spawn rate and limit are part of the parsed data, so do make some sense to include in the checksums. Part of the point is to verify that everything is scripted as expected and not modded, which includes span conditions during universe generation. That said, the spawn rate and limit don't affect the game after galaxy gen, and galaxy gen could be modded without affecting any scripted content checksums...

o01eg:
> Part of the point is to verify that everything is scripted as expected and not modded,

I suppose it's OK for the game to be modded nowadays.

geoffthemedio:
> it's OK for the game to be modded

Modded unexpectedly, I mean. If the client has the same mod applied, it will give the same checksum.

agrrr3:
I think it could make a little sense to have a way to check the generated galaxy if you know the galaxy seed. if the seed, the generation scripts and input do create a different galaxy you could diff it.

I think a "wrong" galaxy should not prevent from playing.

The use case I guess o01eg wants to support is modding the galaxy without having to push the code to all clients (which is a hassle currently)

Issue #3042, opened 2020-06-14
Removing batch of ships from construction queue removes the wrong one


Oberlus:
Bug Report
==========

Environment
-----------
* **FreeOrion Version**: v0.4.10+ build 2020-06-14 9af9a91 CMake
* **Operating System**: Ubuntu 18.04
* **Fetched as** 
  * Compiled from source

Description
-----------
Having two batches of the same kind in the Production Queue and removing the newest (bottom one) causes instead the removal of the oldest (top one).
Queue before removal:
![before](https://user-images.githubusercontent.com/28650914/84592245-609afa80-ae44-11ea-9185-8ce2adc6be1a.png)
Queue after removal of the last (bottom) item:
![after](https://user-images.githubusercontent.com/28650914/84592247-62fd5480-ae44-11ea-8031-ffb6a45b7fc9.png)
Notice the progress of the outpost ship has been lost, presumably because the one removed was actually the one with some progress instead of the selected one.

This is not necessarily limited to items of the same kind, but I have only seen it happen for that situation.
[logs.zip](https://github.com/freeorion/freeorion/files/4776312/logs.zip)
[FreeOrion_Oberlus_Abyss_0039_20200614_124409.zip](https://github.com/freeorion/freeorion/files/4776313/FreeOrion_Oberlus_Abyss_0039_20200614_124409.zip)

Steps to reproduce
------------------
* Load attached game.
* Go to Production Queue and remove the last item.
* Issue occurs.

Oberlus:
Update:

This does not only affects "Remove from Queue", also "Pause", "Move to Queue Top", etc.

One can drag&drop the top one after (under) the bottom one and it will work, the queue will be reordered, but the bottom one cannot be drag&dropped in a different queue spot: it will be the top one which gets relocated. 

So it seems the item ID is wrong for most operations except for drag&drop the top one to a lower queue position.

If adding more outpost ships to the queue, the new ones won't be affected by the bug: you can place one of the new ones in any queue spot and pause or remove them without interfering with the older outpost ships in the queue. So the ID problem might come from a previous turn.

o01eg:
Could it be reproduced for 0.4.10?

Oberlus:
> Could it be reproduced for 0.4.10?

I don't have a save file, but I have encountered this before in MP games, prior to v0.4.10.

Vezzra:
In that case this bug should be considered release-blocking. While not outright game-breaking, it's certainly serious enough.

[...]

PR #3080, opened 2020-06-28
Forbid use of repeated element UUID in production queue

o01eg:
Should help to find cause of #3042


geoffthemedio:
I included that commit in #3081 as it wouldn't want to use it without related adjustments.

PR #3088, opened 2020-07-01
Fix Sly refueling

### LINKED ISSUE #3086: Sly ships don't get Gas Giant extra refuelling
Bug Report
==========

Environment
-----------
* **FreeOrion Version**: 2020-06-23.a2f5c56
  * Compiled from source

Description
-----------
Sly ships don't get Gas Giant extra refuelling.

Steps to reproduce
------------------
* Start a game as Sly.
* Wander around with a scout and let it spent at least 1 fuel, stop in a system without gas giant and check out refuel per turn: it is +0.1.
* Move the scout to a system with a gas giant, stop in there and check out refuel per turn: it is +0.1.

rrobgill:
Correct detection of systems containing gas giants for sly
refueling bonus.

Fixes #3086

### DIFF
--- a/default/scripting/species/SP_SLY.focs.txt
+++ b/default/scripting/species/SP_SLY.focs.txt
@@ -46,8 +46,7 @@ Species
                 Ship
                 Stationary
                 ContainedBy And [
-                    Object id = RootCandidate.SystemID
-                    System
+                    Object id = Source.SystemID
                     Contains condition = And [
                         Planet type = GasGiant
                         Not OwnedBy affiliation = EnemyOf empire = RootCandidate.Owner


agrrr3:
Um. I do not see how the fix could have fixed the behaviour oberlus describes.

@geoffthemedio did something change in RootCandidate handling? Is there a bug?

rrobgill:
> Um. I do not see how the fix could have fixed the behaviour oberlus describes.

Save:

[zzz.sav.gz](https://github.com/freeorion/freeorion/files/4862533/zzz.sav.gz)

image


Before:

image
image

After:

image
image

agrrr3:
I believe you possibly fixed the "wrong" bug

Rob Gill <notifications@github.com> schrieb am Do., 2. Juli 2020, 09:02:

> Um. I do not see how the fix could have fixed the behaviour oberlus
> describes.
>
> See
>

geoffthemedio:
> I believe you possibly fixed the "wrong" bug

Could you be specific? In the posted screenshots in this thread, like my test before merging, it seems to fix the issue with Sly ship refueling in or away from gas giant systems.

agrrr3:
> > I believe you possibly fixed the "wrong" bug
> 
> Could you be specific? In the posted screenshots in this thread, like my test before merging, it seems to fix the issue with Sly ship refueling in or away from gas giant systems.

The condition as it was worked before. So something probably in evaluation changed. So we should figure out if the not-working condition is correct or not and not just patch it to something which works. Also the patch is inconsistent IMHO.

The fix here changes the RootCandidate to the Source in the middle-nested ContainedBy, but not the deep-nested one in the Contains condition

```
ContainedBy And [
            Object id = RootCandidate.SystemID  // patch changes this to Source.SystemID
            System
            Contains condition = And [
                Planet type = GasGiant
                Not OwnedBy affiliation = EnemyOf empire = RootCandidate.Owner
```

We are here in an activation condition so in my understanding RootCandidate and Source should be the same universe object under all circumstances. So the merged patch should not change anything.

Also the oberlus observation that the refuel effect was inverted (working on opposite of the intended targets) means something weird is going on.

geoffthemedio:
> @geoffthemedio did something change in RootCandidate handling?

Not intentionally, and probably not in the v0.4.10 release branch, which this issue reportedly also affected.

agrrr3:
> > @geoffthemedio did something change in RootCandidate handling?
> 
> Not intentionally, 

Thats why I am curious there might be bug in condition handling

> and probably not in the v0.4.10 release branch, which this issue reportedly also affected.

I was thinking - you changed the passing in of "local universes sets", right? that changes should  be in v0.4.10 and master as well.

Thinking again about this ContainedBy is the topmost condition having its own universe set of objects, so the meaning of RootCandidate is the same as LocalCandidate here(?) - basically each object in the universe should match, so the first two lines of the And would find all objects where the SystemID is the id, i.e. all Systems in the universe(?). The nested Contains would remove all systems which do not contain gasgiants and it would also remove all systems which are owned by an enemy of the owner of the system. I am not sure what the owner of a system is. If the owner of a system is "unowned" I have to check what OwnedBy does, but I would expect that unowned is not considered to be an enemy, so that line `Not OwnedBy..` would not remove any candidates.
So maybe you could refill on enemy gasgiants (that is quite possible but was not intended).

Anyway I see no logical case where this should give a fuel boosts to objects in systems without gas giants.

PR #3090, opened 2020-07-01
Remove useless code comments

adrianbroher:
This PR removes various source code comments that do not provide any documentation value:

* Removes the Doxygen grouping for *structors, accessors, mutators and similar groupings.
* ~~Removes class banners.~~
* Removes `template implementation`-start here comments.
* Removes include guard endif comments.


universe/Conditions.cpp
```diff
@@ -480,9 +476,7 @@ unsigned int Number::GetCheckSum() const {
     return retval;
 }
 
-///////////////////////////////////////////////////////////
-// Turn                                                  //
-///////////////////////////////////////////////////////////
+
```
geoffthemedio: I think some of these are rather helpful actually... Particularly in files with a bunch of different but similar implementations.

Don't necessarily need the lines of `///////////` around if those are objectionable to you, though.

adrianbroher: Those are not helpful in any way.  Methods names are already required by the language to be prefixed with the class/struct name.

geoffthemedio: Rephrased: I find them helpful when scanning through the file looking for particular groups of code, _specifically in cases like this file_. The "bunch of different but similar implementations" aren't distinctive enough to easily to see the separations between and to know when one group ends and the next begins without the separator comments. Having the `////` comments makes it a lot easier to navigate the file. Most other cases are unnecessary and not helpful, though.

adrianbroher: Instead of relying on such clutches put them in separate files.  That also untangles internal dependencies and reduced the amount of recomplications of unrelated class implementations.

adrianbroher: Which also give additional gain to use difftools on "similar implementations" to find more potential for refactorings and code deduplications.

geoffthemedio: Separate files would also work. I just prefer to keep these comments while all these implementations are in the same file.

adrianbroher: Removed the class banner commit.

adrianbroher:
Rebased to resolve merge conflict.

PR #3093, opened 2020-07-05
Move FOCS related TU into focs directory; move classes into `focs` namespace.

adrianbroher:
See subject.


geoffthemedio:
What is your definitions of "FOCS"?

I think of it as the scripting language (implemented by the parser) or a script written in it (in a .focs.txt file or similar).

I don't think of it as a subset (Conditions, ValueRefs, Effects) of lower-level functional stuff used in the top level scripted content (eg. BuildingType, Species, Specials, Techs, etc) that gets created when parsing the scripts.

That top level stuff seems like it would be as much "FOCS" as the functional stuff, but is also used in contexts where there is no parsing required... so doesn't seem reasonable to refer to with a name that includes "scripting" in it...

What's your categorization...?

adrianbroher:
I was tempted to answer for a second this but gee, why do I even bother trying refactor this crap?

geoffthemedio:
I assumed you had an answer about why you wanted to refactor this way, which is why I asked. "I don't understand the logic here, please explain" doesn't mean "this clearly has no logic".

Just renaming the namespace might be all that's needed for it to make sense (to me).

Vezzra:
@adrianbroher, closed as...?

adrianbroher:
> @adrianbroher, closed as...?

"I don't give a fuck", so invalid I guess.

PR #3100, opened 2020-07-11
Refactor Python bindings

adrianbroher:
This PR refactors the implementation of the Python client, server and common bindings.  This covers:

* Remove storing the function pointers of helper functions in a variable when a direct function pointer is sufficient.
* Use trailing return types for helper functions for easier conversion into lambdas.
* Use helper functions instead of `boost::bind` or function pointer casts.
* Use lambdas for signature translation only helper functions.  [Apply unary plus operator to decay lambdas into functions.](https://stackoverflow.com/questions/18889028/a-positive-lambda-what-sorcery-is-this).
* Remove `boost::python::make_function`  on `boost::python::def` and `boost::python::class_<T>::def` calls, as the construction already happens for those compared to `boost::python::class_<T>::add_attribute`.

### DIFF
--- a/client/AI/AIFramework.cpp
+++ b/client/AI/AIFramework.cpp
@@ -80,7 +80,8 @@ BOOST_PYTHON_MODULE(freeOrionAIInterface)
 //////////////////////
 //     PythonAI     //
 //////////////////////
-bool PythonAI::Initialize() {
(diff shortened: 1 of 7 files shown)


geoffthemedio:
seems to work... can't spare time to do a detailed check of every line, though.

PR #3101, opened 2020-07-11
Remove unnecessary step (check-python)

Cjkjvfnby:
`check-python` step is obsolete, it is already checked inside `check-pep8`


o01eg (approved):
I'm not sure about it so I'll approve it. If PPA build catch compileall issue I'll return this step.

PR #3103, opened 2020-07-13
 WIP: Add new GG::Measure class to enable UI scaling 

rah2501:
This is the approach I've come up with to supporting UI scaling, adding a new class to handle static UI measurements and scaling them according to font size.  There's still a lot of classes that need changing to make use of it but I wanted to get feedback before I continue.

### DIFF
--- a/GG/GG/BrowseInfoWnd.h
+++ b/GG/GG/BrowseInfoWnd.h
@@ -32,6 +32,7 @@
 
 #include <GG/Wnd.h>
 #include <GG/Font.h>
+#include <GG/Measure.h>
 #include <GG/GLClientAndServerBuffer.h>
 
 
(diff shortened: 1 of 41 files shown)


adrianbroher:
There is no need to define a dedicated type, X and Y have a 1:1 pixel size relation only by convention and there is no reason to handle them as arbitary screen coordinates with a non one to one mapping towards the pixels.  The scaling should happen at rendering level and rely only on the target display HiDPI capabilities.

rah2501:
> The scaling should happen at rendering level

Could you say a bit more about what you mean by "rendering level"?  There are OpenGL calls everywhere, there doesn't seem to be any distinct area of the code where rendering is done.  Are you saying all of the OpenGL rendering calls should have scaling involved?  Assuming so, I've never touched OpenGL before, could you say a little more about that could be achieved?

> and rely only on the target display HiDPI capabilities.

I disagree with this; display HiDPI capability is not the only reason to want to scale the UI.

adrianbroher:
> Assuming so, I've never touched OpenGL before, could you say a little more about that could be achieved?

OpenGL or rendering API in general allows to set a transformation matrix or a stack of matrices to manipulate the viewport during a drawing operation.

rah2501:
> set a transformation matrix or a stack of matrices to manipulate the viewport during a drawing operation

Could you possibly say how that's done?  Again, I've never touched OpenGL.  Could you provide an example?

PR #3105, opened 2020-07-14
WIP: Add new scaling helper functions in GiGi

rah2501:
This is a cargo-culted attempt to implement scaling at "the rendering level" as per [comments](https://github.com/freeorion/freeorion/pull/3103#issuecomment-657819800) in my other merge request.  It doesn't work.  I don't understand why.  Help is welcome.

### DIFF
--- a/GG/GG/CMakeLists.txt
+++ b/GG/GG/CMakeLists.txt
@@ -58,6 +58,7 @@ target_sources(GiGi
         ${CMAKE_CURRENT_LIST_DIR}/MultiEditFwd.h
         ${CMAKE_CURRENT_LIST_DIR}/MultiEdit.h
         ${CMAKE_CURRENT_LIST_DIR}/PtRect.h
+        ${CMAKE_CURRENT_LIST_DIR}/Scale.h
         ${CMAKE_CURRENT_LIST_DIR}/Scroll.h
         ${CMAKE_CURRENT_LIST_DIR}/ScrollPanel.h
         ${CMAKE_CURRENT_LIST_DIR}/Slider.h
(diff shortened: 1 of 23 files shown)


geoffthemedio:
I'm not sure what adrian_broher meant, but I think you would set the scale with a matrix push and then do the rest of the rendering independent of that. I don't think every class's rendering code needs rewriting...

rah2501:
> I think you would set the scale with a matrix push and then do the rest of the rendering independent of that

You'll have to forgive my ignorance but I have no idea how to make that work.  I don't understand how FreeOrion makes use of the different matrix stacks or how the different stacks, which according to the documentation for `glPushMatrix` correspond to different modes, interact with each other or with mode switches.

Can anyone suggest where in FreeOrion might be a suitable place to add a scaling matrix?

hifron:
On [Khronos Getting Started](https://www.khronos.org/opengl/wiki/Getting_Started) - OpenGL Standard maker there are links to some tutorials and for [OpenGL official docs reference](https://www.khronos.org/opengl/wiki/OpenGL_Reference) there is function [glViewport](https://www.khronos.org/registry/OpenGL-Refpages/gl4//html/glViewport.xhtml) since OpenGL 2.0 although from [OpenGL4 index](https://www.khronos.org/registry/OpenGL-Refpages/gl4//) , but I don't have such display to test.

But the document is licensed under the SGI Free Software B License.

HW Graphics maker deliver their implementing drivers, but purpose of it is to deliver application, not for drivers itself, but standards like OpenGL is hard way to invent and also to implement and there should be used best industry practices, but also user interest to make such applications and GG could be something like QT(with special permissions of QT for project like kde.org), but GG won't become such library that way. At least official.

PR #3098, opened 2020-07-08
Migrate Python related CI jobs onto GitHub actions CI

adrianbroher:
This PR migrates the existing Travis CI jobs related to Python

* Syntax check (compileall)
* PEP-8 conformance test
* Unit test

onto GitHub Actions CI infrastructure.

The PR contains some intentional bugs to show the integration of GitHub Check annotations, which display errors and warnings within the PR file diff for easier reviewing.

I'm not sure if the annotations are already created on this initial PR, so you can take a look into a PR in private repository: adrianbroher/freeorion#8.  Most relevant aspects are the `Checks` tab for looking into the CI run logs and the `Files` tab to see the in-place annotations for the provoked error messages.

PR #3106, opened 2020-07-15
Use github actions for checks

Cjkjvfnby:
This is a rework of the https://github.com/freeorion/freeorion/pull/3098

This PR is ready for merge but will require more steps after that.  I expect the following workflow:

- merge this PR, action will become available and will be triggered on each build.
- configure this action as required for PRs to master. [Enabling required status checks](https://docs.github.com/en/github/administering-a-repository/enabling-required-status-checks) 

## Important changes
I have removed a pytest run on Python 3.8.  Python backward compatibility is good enough. 

## Benefits
- GitHub action has a cool feature, it is possible to leave a comment at the line that caused an error. It is already enabled for flake8 and pytest.
- It takes less time to finish tests.
- Code is simpler (we don't need to put C++ and python configuration into the same file, as we do with Travis)

## Next steps
- Add annotation to the string tables check
- Allow skipping checks if no corresponding files are affected.
- Move C++ checks to GitHub actions
- Automate the creation of the windows build

### DIFF
--- a/.github/workflows/python-check.yml
+++ b/.github/workflows/python-check.yml
@@ -1,0 +1,35 @@
+name: AI and string tables checks
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.6', '3.9']
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install development dependencies
+        run: |
+          pip install -r default/python/requirements-dev.txt
+      - name: Lint with flake8
+        run: |
+          flake8 default/python --config=default/python/tox.ini --format="::error file=%(path)s,line=%(row)d,col=%(col)d::%(path)s:%(row)d:%(col)d: %(code)s %(text)s"
+      - name: Test with pytest
+        run: |
+           pytest -v default/python/tests
+      - name: Check string tables formatting
+        run: |
+          diff -u default/stringtables/en.txt <(check/st-tool.py format default/stringtables/en.txt) || { echo "String table is not properly formatted"; exit 1; }
+      - name: Validate string tables
+        run: |
+          ./check/st-tool.py check --reference default/stringtables/en.txt default/stringtables/*.txt
--- a/.github/workflows/python-check.yml
+++ b/.github/workflows/python-check.yml
@@ -19,17 +19,25 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install development dependencies
-        run: |
-          pip install -r default/python/requirements-dev.txt
+        run: pip install -r default/python/requirements-dev.txt
       - name: Lint with flake8
-        run: |
-          flake8 default/python --config=default/python/tox.ini --format="::error file=%(path)s,line=%(row)d,col=%(col)d::%(path)s:%(row)d:%(col)d: %(code)s %(text)s"
+        # Run from root to preserve full path in output
+        run: >
+          flake8 default/python
+          --config=default/python/tox.ini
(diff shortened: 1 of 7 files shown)


o01eg:
I suppose it is very actual because of issues with Travis CI.

geoffthemedio:
please rebase on master, rather than merging master into your branch...

Cjkjvfnby:
I think it is ready to be merged.

o01eg (changes requested):
Could we test it against different python versions?
I see it uses matrix.python-version but I don't see its definition.

Cjkjvfnby:
> Could we test it against different python versions?

I think we can test only the minimum required version.  

There are 2 reasons for it:
- Python preserves compatibility and we don't expect any failures.
- I don't remember any compatibility problem since we introduce testing on two versions.

Testing only one Python version saves resources and make thing a bit easier.

o01eg:
As I know Blender discourages to use python 3.9 due compatibility issues https://bugs.gentoo.org/737388 . Also I see it takes less a minute to execute python tests.

Also we could use matrix to choose version so it will be simpler to switch to next versons.

.github/workflows/python-check.yml
```diff
@@ -0,0 +1,37 @@
...
+      - opened
+      - edited
+      - reopened
+      - synchronize
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.6', '3.9']
```
o01eg, line 14: It gets `No steps defined in `steps` for the following jobs: build` error. Maybe it should use options without quotes?

o01eg (changes requested):

o01eg (approved):
Looks good

.github/workflows/python-check.yml
```diff
@@ -0,0 +1,35 @@
...
+          flake8 default/python --config=default/python/tox.ini --format="::error file=%(path)s,line=%(row)d,col=%(col)d::%(path)s:%(row)d:%(col)d: %(code)s %(text)s"
+      - name: Test with pytest
+        run: |
+           pytest -v default/python/tests
+      - name: Check string tables formatting
+        run: |
+          diff -u default/stringtables/en.txt <(check/st-tool.py format default/stringtables/en.txt) || { echo "String table is not properly formatted"; exit 1; }
+      - name: Validate string tables
+        run: |
+          ./check/st-tool.py check --reference default/stringtables/en.txt default/stringtables/*.txt
```
o01eg: @Cjkjvfnby You should use `??.txt` pattern as I moved includes here too.

o01eg (changes requested):

Cjkjvfnby:
I have split the script into 2 workflows.  One is run with multiple pythons (check and linter) and another one with the single one (string tables check).

o01eg:
> I have split the script into 2 workflows. One is run with multiple pythons (check and linter) and another one with the single one (string tables check).

It possible to use conditions.

Cjkjvfnby:
> It possible to use conditions.

This will make the code a bit dirty. I'd prefer to have a simple and straightforward code, which is easy to understand modify and support.

o01eg (approved):

geoffthemedio:
@Cjkjvfnby @o01eg Is this resolved and ready to merge or still under discussion?

o01eg:
@geoffthemedio It's fine for me.

Cjkjvfnby:
@geoffthemedio , since these new jobs are never failed for no reason and always executed we could make them mandatory for merge.

See https://github.com/freeorion/freeorion/settings/branch_protection_rules

This will look like:
image

geoffthemedio:
I'm hesitant to make test builds mandatory for merging if they can take 30+ min to run, even if they don't fail randomly...

Cjkjvfnby:
> I'm hesitant to make test builds mandatory for merging if they can take 30+ min to run, even if they don't fail randomly...

Check that introduced here takes less than a minute.  Look ok for me to make them required. 

Looks like I have messed with task names, they are not informative. I will make new PR to fix it.

geoffthemedio:
> Check that introduced here takes less than a minute. Look ok for me to make them required.

OK then.

PR #3109, opened 2020-07-18
Make imperial stockpile stringtable entries slightly better

agrrr3:
As new players regularly have problems understanding the imperial stockpile this makes the entries slightly more consistent and clear. 

* In detail entries, when production points are referenced first, write it as: "Production Points (PP)"
* Some consistency in casing of technical terms
* Some consistency in naming of extraction limit (e.g. never calling it extraction limit)
* a little bit more clarity for stockpiling focus
* a little bit more clarity for stockpile liquidation policy

Actually we should find a better term than stockpiling which expresses the following:
* planets have a local max-capacity, which may be increased by policy, buildings, species..
* this local capacity adds up to a global/imperial value which describes a limit
* there is also an imperial value which tracks a resource
** excess industry gets converted into it
** it can be used to build buildings and ships anywhere in your empire ("convert it back in PP")
** the imperial limit limits the amount which may be converted/used for that

The current difficulty arises that you increase a local capacity, which actually increases your
conversion limit somewhere else...

ideas: smuggling, hidden reserve, shipping, teleporting, transport, deep space, deep aid, courier...
...stash...

Should also be cherry-picked for 0.4.10

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -5679,7 +5679,7 @@ There are two ways two store PP in the imperial stockpile:
 * PP explicitly allocated to a [[encyclopedia PROJECT_BT_STOCKPILE]] are also transferred to the stockpile.
 
 The imperial stockpile use limit is the total of the [[metertype METER_STOCKPILE]] of each planet.
-Species stockpiling ability, buildings, technologies, and setting the [[encyclopedia STOCKPILE_FOCUS_TITLE]] increase it.
+Species stockpile distribution ability, buildings, technologies, and setting the [[encyclopedia STOCKPILE_FOCUS_TITLE]] increase it.
 
 Information about imperial stockpile, the use, and the use limit can be seen in the top of the [[encyclopedia MAP_WINDOW_ARTICLE_TITLE]].
 Details for production can be seen in the [[encyclopedia PRODUCTION_WINDOW_ARTICLE_TITLE]].
@@ -5835,7 +5835,7 @@ Industry
 METER_INDUSTRY_VALUE_DESC
 '''Industry represents the production and modification of physical goods. It is needed to produce buildings, space ships and other projects.
 
-The Industry of a planet can be used for any production projects of [[metertype METER_SUPPLY]] line connected planets. Unallocated Industry gets added to the Imperial Stockpile.
+The Industry of a planet can be used for any production projects of [[metertype METER_SUPPLY]] line connected planets. Unallocated Industry gets added to the [[encyclopedia STOCKPILE_TITLE]].
 
 The Target Industry is the stable Industry output level a planet will approach, given the current factors influencing the Industry resources. Its value is connected to [[encyclopedia ENC_BUILDING_TYPE]]s built on the planet, [[encyclopedia ENC_TECH]] researched by the empire, [[encyclopedia ENC_SPECIAL]]s linked to the planet, and various other game effects. If these factors change, then the Target Industry value is likely to shift.
 
@@ -5940,12 +5940,12 @@ Supply lines can also be used to refill the [[metertype METER_FUEL]] supply of s
 Supply lines will be obstructed by armed enemy ships belonging to empires with which an empire is at [[encyclopedia WAR_TITLE]]. At the opposite end of the spectrum, empires in an [[encyclopedia ALLIANCE_TITLE]] will share supply lines and all their associated benefits.'''
 
 METER_STOCKPILE_VALUE_LABEL
-Stockpile
+Stockpile Distribution
 
 METER_STOCKPILE_VALUE_DESC
-'''The stockpile meter of a planet represents how much it contributes to the [[encyclopedia STOCKPILE_TITLE]] use limit. The imperial stockpile use limit indicates an empire's ability to use stockpiled Production Points (PP). An empire's stockpile use limit is the sum of all its planets' stockpile meters plus technology bonus. The total determines how many PP can be taken out of the stockpile each turn to be used for production.
+'''The stockpile distribution meter of a planet represents how much it contributes to the [[encyclopedia STOCKPILE_TITLE]] use limit. The imperial stockpile use limit indicates an empire's ability to use stockpiled Production Points (PP). An empire's stockpile use limit is the sum of all its planets' stockpile distribution meters plus technology bonus. The total determines how many PP can be taken out of the stockpile each turn to be used for production.
 
-Species have various innate levels of stockpiling ability.'''
+Species have various innate levels of stockpile distribution ability.'''
 
 METER_TROOPS_VALUE_LABEL
 Troops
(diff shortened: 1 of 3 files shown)


geoffthemedio:
I think it's better to describe PP as stored than as converted to yet another something.

agrrr3:
> I think it's better to describe PP as stored than as converted to yet another something.

Ok. That is a very valid description for the stockpile metaphor.

Vezzra:
> Should also be cherry-picked for 0.4.10

As this isn't critical stuff, but also only text changes which can't break anything, I won't do another release candidate just for this, but am willing to include it if another one is needed after this has been merged.

So, if this gets merged before the release has been declared, tag it for cherry-picking.

default/stringtables/en.txt
```diff
@@ -5914,9 +5914,9 @@ Supply
...
 
-Supply lines are shown by coloring star-lanes with an empire's color. Planets connected by supply lines form a 'Resource Group' and can share physical resources. Production Points created on one planet can be used on any connected planet to build ships or structures.
+Supply lines are shown by coloring star-lanes with an empire's color. Planets connected by supply lines form a 'Resource Group' and can share physical resources. Production Points (PP) created on one planet can be used on any connected planet to build ships or structures.
 
-A set of core supply starlanes for an empire (the set of least-jump starlanes connecting the resource-producing systems of each Resource Group) will have greater thickness than the non-core starlanes. If PP is currently being wasted within a Resource Group, the outer bands of the core starlanes will be a contrasting color.
+A set of core supply starlanes for an empire (the set of least-jump starlanes connecting the resource-producing systems of each Resource Group) will have greater thickness than the non-core starlanes. If PP are currently unalloted within a Resource Group, the outer bands of the core starlanes will be a contrasting color.
```
geoffthemedio: unalloted -> unallocated or unused

Cjkjvfnby:
Suggestions:

`consumption rate`
`per turn consumption`
`per turn usage`

default/stringtables/en.txt
```diff
@@ -11053,10 +11053,10 @@ PLC_STOCKPILE_LIQUIDATION
 Stockpile Liquidation
 
 PLC_STOCKPILE_LIQUIDATION_DESC
-Doubles stockpiling on stockpile-focused planets.
+Doubles stockpile use capacity added by stockpile-focused planets.
 
 PLC_STOCKPILE_LIQUIDATION_SHORT_DESC
-Increases stockpiling
+Increases Stockpile Use Limit
```
Cjkjvfnby, line 11076: Inconsistent term use with `PLC_STOCKPILE_LIQUIDATION_DESC`

geoffthemedio:
Re: terminology, it needs to be something like "limit", as it's not a rate in of itself, but a limit on how much can be used. Perhaps "Stockpile Extraction Limit" or "Stockpile Extraction Capacity".

Cjkjvfnby:
>Re: terminology, it needs to be something like "limit", as it's not a rate in of itself, but a limit on how much can be used. Perhaps "Stockpile Extraction Limit" or "Stockpile Extraction Capacity".

When I see capacity and stockpile together, I think about the total amount of goods that can be combined in that stockpile.  
"Limit" is too broad. Everything has a limit. 
"Extraction" looks like some complex action. But we are doing something simple. Just take something that was put there for taking. 


`Take out limit`.  (It is already present in PR)
`Bring limit`
`Fetch limit`
`Supply`
`Deliver`

> as it's not a rate in of itself

But the maximum rate is technically a limit. I like this word since it associates with `something per time` like `PP per turn`. In our case, it is more narrow than capacity or limit terms.

PS. My suggestions are a kind of brainstorming. I am not sure that I will like them tomorrow morning :)

geoffthemedio:
> > as it's not a rate in of itself
> 
> But the maximum rate is technically a limit.

Yes. The point is it can't just be a "rate", it has to have something like "limit" in it as well or instead.

Maybe `Stockpile Withdrawal Limit`?

default/stringtables/en.txt
```diff
@@ -11053,7 +11053,7 @@ PLC_STOCKPILE_LIQUIDATION
 Stockpile Liquidation
 
 PLC_STOCKPILE_LIQUIDATION_DESC
-Doubles stockpile use capacity added by stockpile-focused planets.
+Doubles planetary stockpile use capacity of stockpile-focused planets which increases the imperial stockpile use limit.
```
geoffthemedio: This is more confusing after the change... It looks like "planetary stockpile" a thing...

Maybe `Doubles the contribution of stockpile-focused planets to the imperial stockpile use limit.`

Should probably add links in that to stockpile focus and whatever stockpiling article there is in the pedia.

agrrr3: did some crosslinking, fixed up the map window description and added a new encyclopedia entry for imperial stockpile

stockpile meter and imperial stockpile article are now the nexi of stockpile information

agrrr3:
Should i rename the stockpile focus to something like  "Remote Production" ?

Cjkjvfnby:
> Stockpile Withdrawal Limit

I like it!

Cjkjvfnby:
> Should i rename the stockpile focus to something like "Remote Production" ?

Remote is a too broad term. It means that I produce PP on one planet and use them on the other. Both supply lines and Imperial stockpile fits.

If we will stick with `Withdrawal Limit` than it should be something like `Deposit focus`.

agrrr3:
> > Stockpile Withdrawal Limit
> 
> I like it!

it sounds like a bank account to me

But the main problem is it is too long for most UI. There needs to be a shorter term. 

Maybe a stash, where you transfer your PP to and your planets have a smuggling focus and that is the smuggling limit? (Such a change is outside the scope of this PR - it would also mean to rewrite all the fluff for stockpile buildings/tech)

"Remote Production" seemed to be more fitting than "Unconnected Production" or telework. And the normal usual terminology sais you have local Industry and send the produce via supply network to the building/ship production site. So there is no remote production in a supply network.

Deposit focus is exactly not what the stockpile focus does (and the reason why it should be renamed).

agrrr3:
Hm AI has problems with invalid ship designs... is that coming from master?

geoffthemedio:
> smuggling focus

No. It is not, at least in general, implied to be a "smuggling" type process.

> it sounds like a bank account to me

The concept it similar. It's an account that stores your extra money / PP until needed. You can put in as much as you want, but you have a limit per day / turn on how much you can take out.

> Deposit focus is exactly not what the stockpile focus does (and the reason why it should be renamed).

It's exactly the opposite of what it does. The focus enables your empire to withdrawal more, not deposit more. I'm not sure what your point is.

agrrr3:
> > smuggling focus
> 
> No. It is not, at least in general, implied to be a "smuggling" type process.

Yes. For such a change in terminology all the fluff would need to be rewritten (and why it is out of scope of this PR; Smuggling would partly explain why extraction is limited and happens somewhere else and not through the official channels (supply network). This was just an example for changing the base terminology so the mechanic becomes more intuitive).

> 
> > it sounds like a bank account to me
> 
> The concept it similar. It's an account that stores your extra money / PP until needed. You can put in as much as you want, but you have a limit per day / turn on how much you can take out.

Yes and it may help clarifying the mechanic. It just does not feel space-opera that your empire has a savings account. (i do not know the correct financial terms here - pretty sure savings account is wrong)

> 
> > Deposit focus is exactly not what the stockpile focus does (and the reason why it should be renamed).
> 
> It's exactly the opposite of what it does. The focus enables your empire to withdrawal more, not deposit more. I'm not sure what your point is.

Yes, that was my point. "exactly not what it does" == "opposite of what it does". in case that equality is not valid in english it was a germanism

Or to be clearer the "Stockpile Focus" is misleading:

> It is counterintuitive that one planet focused to stockpile, with no local production (or just the focus-independent bonuses), is helping a far away planet to build its stuff. It's in fact focused to help (others) extraction.

agrrr3:
After discussion on the forum:

The meters: (max) Stockpile --> (max) Stockpile Distribution
The focus and activity/species ability: Stockpiling (Focus) -> Stockpile Distribution (Focus) 

 IMHO this can be merged into master.

PR #3115, opened 2020-07-24
Fix Experimentors sitrep about spotted monsters

Oberlus:
Fixes the number of reported monsters created by the Experimentors on each way (before it was 3 on the first phase of each monster, now it's always 2).

Cherrypick for release, it won't need playtesting or a RC2.


geoffthemedio:
if there's another release candidate, can be cherry picked

Vezzra:
> Cherrypick for release, it won't need playtesting or a RC2.

The moment we make another build, it requires playtesting, just to make sure nothing went wrong in the build process (very unlikely, but possible). Which is why the release candidates do not have a "RC#" anywhere in their version string, because it would require a rebuild to change the version string to something fitting of "official release".

That rebuild, although nothing changed but the version string, would again require testing. However, doing it the way we do, we can just rename the last RC build and declare it the official release.

Issue #3111, opened 2020-07-21
AI Empty List Warnings Bug?


geoffthemedio:
In master, when quickstarting a game, I see various `[warn]` lines in the AI log files about `Try to enqueue tech from empty list` and `NoneType: None`. An excerpt:
````
09:56:10.368129 {0x000076a0} [debug] ai : aiclientapp.cpp:257 : Extracted GameStart message for turn: 1 with empire: 3
09:56:10.368129 {0x000076a0} [debug] ai : aiclientapp.cpp:261 : Message::GAME_START loaded_game_data: false
09:56:10.368129 {0x000076a0} [debug] ai : aiclientapp.cpp:267 : Message::GAME_START Starting New Game!
09:56:10.368129 {0x000076a0} [debug] ai : aiclientapp.cpp:297 : Message::GAME_START getting Zone AI aggression, RNG Seed: 2644641040
09:56:10.368129 {0x000076a0} [debug] ai : aiclientapp.cpp:315 : Message::GAME_START setting AI aggression as 5 (from rnd 5; max aggression 5)
09:56:10.368129 {0x000076a0} [debug] python : FreeOrionAI.py:94 : Initializing AI state...
09:56:10.369130 {0x000076a0} [debug] python : character_module.py:494 : EmpireIDTrait initialized.
09:56:10.370130 {0x000076a0} [debug] python : character_strings_module.py:56 : CharacterTable returns Maniacal
09:56:10.370130 {0x000076a0} [debug] python : FreeOrionAI.py:99 : New game started, AI Aggression level 5 (Maniacal)
09:56:10.375133 {0x000076a0} [info] python : AIstate.py:912 : Fleet Summary Turn 1
09:56:10.375133 {0x000076a0} [info] python : AIstate.py:912 : =============================================================================
09:56:10.375133 {0x000076a0} [info] python : AIstate.py:912 : | Fleet                | Rating   | Troops | Location         | Destination |
09:56:10.375133 {0x000076a0} [info] python : AIstate.py:912 : =============================================================================
09:56:10.375133 {0x000076a0} [info] python : AIstate.py:912 : | F_2744<Scout Fleet>  |     0.00 |   0.00 | S_2144<Mimosa ε> | -           |
09:56:10.375133 {0x000076a0} [info] python : AIstate.py:912 : | F_2792<Colony Fleet> |     0.00 |   0.00 | S_2144<Mimosa ε> | -           |
[...]

PR #3117, opened 2020-07-24
Add tech group tests

Cjkjvfnby:
Changing tech lists in https://github.com/freeorion/freeorion/pull/2965 introduced warnings: https://github.com/freeorion/freeorion/issues/3111

@agrrr3 , @Oberlus I have changed TechLists.enqueue to eliminate warnings.  These changed do not modify what was added to the research queue. 

PS this PR is based on https://github.com/freeorion/freeorion/pull/3116

### DIFF
--- a/default/python/AI/TechsListsAI.py
+++ b/default/python/AI/TechsListsAI.py
@@ -82,12 +82,10 @@ class TechGroup:
                     # Do not display error message as those should be shown only once per game session
                     # by the initial test_tech_integrity() call.
                     msg = "[step %s]: Try to enqueue tech from empty list" % step
-                    warning(msg)
                     self._errors.append(msg)
                     continue
             if tech_name in self._tech_queue:
                 msg = "[step %s]: Tech is already in queue: %s" % (step, tech_name)
-                warning(msg)
                 self._errors.append(msg)
             else:
                 self._tech_queue.append(tech_name)
--- a/default/python/AI/TechsListsAI.py
+++ b/default/python/AI/TechsListsAI.py
@@ -184,7 +184,6 @@ class TechGroup1SparseB(TechGroup1):
         super(TechGroup1SparseB, self).__init__()
         self.enqueue(
             self.economy,
-            self.economy,
             self.weapon,
             self.hull,
             "SHP_ZORTRIUM_PLATE",
@@ -216,7 +215,6 @@ class TechGroup1SparseC(TechGroup1):
     def __init__(self):
         super(TechGroup1SparseC, self).__init__()
         self.enqueue(
-            self.economy,
             self.economy,
             "SHP_ORG_HULL",
             "SHP_MULTICELL_CAST",
@@ -298,7 +296,6 @@ class TechGroup2A(TechGroup2):
             self.weapon,
             self.weapon,
             self.weapon,
-            self.economy,
             self.hull,
         )
 
@@ -343,11 +340,8 @@ class TechGroup2SparseA(TechGroup2):
             self.economy,
             self.economy,
             Dep.LRN_ARTIF_MINDS_1,
-            self.defense,
-            self.defense,
             self.weapon,
             self.weapon,
-            self.defense,
             self.weapon,
             self.weapon,
             self.weapon,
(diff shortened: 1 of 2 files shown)


geoffthemedio:
@agrrr3 @Oberlus @Morlic-fo any comments?

agrrr3:
The fix seems dubious. In #2965 introduced one learning tech (Dep.LRN_ARTIF_MINDS_2) and one production tech (Dep.PRO_AUTO_2), both were added to TechGroup3 base in the economy dictionary. If counts are off because of #2965 it should be at those places.
The "hull" and "defense" dicts should not be involved IMHO.

Cjkjvfnby:
> The fix seems dubious

It is not a fix.  It is how the game actually works. 

@agrrr3 , my bad, this was introduced far long before that changes.  

Nobody just noticed it for a couple of years.
https://github.com/freeorion/freeorion/pull/2338

Cjkjvfnby:
Close and open to trigger build again.

Cjkjvfnby:
Close and open to trigger build again.

Cjkjvfnby:
@agrrr3 , I cannot assign you as a reviewer via UI. Can you review and leave your comments?

default/python/AI/TechsListsAI.py
```diff
@@ -346,11 +340,8 @@ def __init__(self):
             self.economy,
             self.economy,
             Dep.LRN_ARTIF_MINDS_1,
-            self.defense,
```
agrrr3, line 349: These removed self.defense entries are in TechGroup2SparseA (the others are in TechGroup2SparseB) - both inherit from TechGroup2 which does not contain a defense list, so it is ok to remove the entries.
Also in a sparse setup you do not need planetary defense soon, makes also sense for the AI.

default/python/AI/TechsListsAI.py
```diff
@@ -467,7 +455,7 @@ def __init__(self):
             self.hull,
             self.misc,
             self.economy,
-            self.hull,
```
agrrr3, line 470: TechGroup3 have asteroid reformation, heavy asteroid hulls and contragravitational maintenance techs in the self.hull list. That is not very effective. Asteroid reformation gets researched very soon. Heavy asteroid hulls are currently not a great deal compared to normal asteroid hulls. IMHO if one researches scattered asteroid hulls, heavy asteroid hull get automatically researched as well - so personally I would rather drop the heavy asteroid hulls  from TechGroup3 as well or maybe switch order for contragrav and heavy asteroid hulls and not remove this self.hull reference, but the one before (so heavy asteroid hulls are quite late to be researched).

default/python/AI/TechsListsAI.py
```diff
@@ -522,7 +510,6 @@ def __init__(self):
...
```
agrrr3, line 525: as above, rather remove the self.hull in 522

default/python/AI/TechsListsAI.py
```diff
@@ -573,7 +560,6 @@ def __init__(self):
...
```
agrrr3, line 576: as above, rather remove the entry in 573

default/python/AI/TechsListsAI.py
agrrr3: so, instead one could remove the entry in line 467

agrrr3:
Ok i added some half-informed opinions. Did not have the time for checking the economy removals. Could you give a reasoning why you want to remove those self.economy entries and not other ones?

Cjkjvfnby:
> Could you give a reasoning why you want to remove those self.economy entries and not other ones?

I just remove the latest entry, because it was not enqueued.  

Since this changed were introduced a long time ago, they can be considered a feature (I don't know author intent and nobody sends a complaint about bad AI). So I just did c cleanup, without actual changes in AI behavior.

I don't have the expertise to reason about techs, that's why I need your help.

I prefer to merge this PR as is, and do you suggested changes in the next one. 

PS.  How do you check techs? I can try to build some visualization if you need it.

geoffthemedio:
@agrrr3 any further comments / objections...?

agrrr3:
merge, we can do this later

PR #3137, opened 2020-08-11
Re-organize sources layout

o01eg:
Re-organizes sources in the next folders:

- src/ - all sources
- - src/server/ - sources for server-side binaries (AI and server)
- - - src/server/ai/ - sources for `freeorionca`
- - - src/server/server/ - sources for `freeoriond`
- - src/client/ - sources for clients
- - - src/client/gigi - sources for GG-based client
- - src/common/ - common sources


Vezzra (commented):
While I understand why you'd want to group the AI client sources together with the server sources from a "deployment point of view", IMO it does not make sense from the "sourcecode point of view".

The AI client is a _client_, _not_ a server component. AFAIK, the AI client does not share any code/modules with the server the human client doesn't share also, but there is code/there are modules the AI and the human client share with each other, but not with the server.

Therefore IMO the current layout makes sense, and should be left as it is.

No opinion on moving the GiGi sources into the client tree.

However, I defer decision on that to @adrianbroher and @geoffthemedio.

geoffthemedio:
@adrianbroher reportedly has some plans for source code reorganization, so I'd probably leave it to him to make such re-arrangements...

o01eg:
As I see @adrianbroher did something with GG in the September

o01eg:
> AFAIK, the AI client does not share any code/modules with the server the human client doesn't share also

@Vezzra Python wrappers. At least we aren't going to use python at the client.

o01eg:
Rebased.

Vezzra:
> Python wrappers. At least we aren't going to use python at the client.

Ah yes, forgot about those. Although, if we go ahead and replace FOCS with Python at some point, the human client too will include the Python wrappers.

o01eg:
If we won't use YAML to replace FOCS.

Vezzra:
That's another option. However, personally I lean towards keeping the required frameworks/parsers for the various config/script files to a minimum. If we can do everything properly with Python, then lets do everything with Python.

Less dependencies to maintain.

o01eg:
I had idea not to execute python from data but use it as parser so server's and AI's code for python couldn't be used here.

Vezzra:
@o01eg, closed as abandoned/invalid/superseded/merged...?

o01eg:
Yes, it's abandoned.

Issue #3148, opened 2020-08-24
Fleets Get Into Glitched State With No Previous Or Next System When Starlane Removed


raptor:
Bug Report
==========
Environment
-----------
* **FreeOrion Version**: 0.4.10
* **Operating System**: Linux x86_64 (openSUSE 15.1)
* **Graphic card used**: Intel i915
* **Fetched as** 
  * Compiled from source

Description
-----------
If you load the attached save game, then select the single ship that is shown in the Deep Space circle between Kent Zeta and Arkab Beta, the game will crash with the following error:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  system1_id invalid
[2020-08-23 20:49:57.537367] [0x00007fc5802d0300] [info]    ServerApp::~ServerApp
[2020-08-23 20:49:57.537414] [0x00007fc5802d0300] [info]    Server exited cleanly.
Aborted (core dumped)
buckd@obsidian:~> [2020-08-23 20:49:57.552693] [0x00007fc5802d0300] [info]    Cleaned up FreeOrion Python interface
[2020-08-23 20:49:57.561195] [0x00007fc5802d0300] [info]    Reset id allocators with highest object id = -1 and highest design id = 65309
```
![Screenshot_20200823_204846](https://user-images.githubusercontent.com/6452412/90998937-7afafe00-e582-11ea-94fd-637fb4400236.png)

[freeoriond.log](https://github.com/freeorion/freeorion/files/5115560/freeoriond.log)
[freeorion.log](https://github.com/freeorion/freeorion/files/5115562/freeorion.log)

[FreeOrion_ghjhg_empireripme_0349_20200823_194800.zip](https://github.com/freeorion/freeorion/files/5115567/FreeOrion_ghjhg_empireripme_0349_20200823_194800.zip)

geoffthemedio:
I can stop the crash with some extra safety checks, but it leaves that fleet in a weird glitched state.

Do you have a save from the turn before?

raptor:
Sadly, I do not - the auto-save rotation lost it as I was trying to duplicate it again with the other Nova 1 ship.  But I think I did duplicate the glitch in the below file set, or pretty close to it.

[FreeOrion_ghjhg_empireripme_0350_20200824_070434.zip](https://github.com/freeorion/freeorion/files/5117956/FreeOrion_ghjhg_empireripme_0350_20200824_070434.zip)

[...]

PR #3150, opened 2020-08-24
Fix Fleet Crash

geoffthemedio:
Avoids crash but not cause of #3148


Vezzra:
@geoffthemedio, as this is only a workaround, not a real fix for #3148, shall we reassign #3148 to 0.5? This fix here should be good enough for the bugfix release, or do you want to provide a proper fix also for 0.4.10.1?

geoffthemedio:
There are other priorities, so I suppose it doesn't need to block the bugfix release to fully resolve the issue...

Issue #3143, opened 2020-08-19
[AI] Improve TechList technology enque


Cjkjvfnby:
I do an investigation for the AI TechGroups. Tech group might riquire adjustments as we discussed in the https://github.com/freeorion/freeorion/pull/3117.

This is important part of the AI but it is not very transparent how tech are added.

I create this ticket to collect a list of notices I have about it.  It may take some time, and I want to save progress in a persistent place.

Cjkjvfnby:
This function completely breaks an order of the execution, so SHP_WEAPON_1_3 can be added before the SHP_WEAPON_1_2.

```
    def _add_remaining(self):
        """Add all remaining techs in the tech group to self._tech_queue if not already contained."""
        all_needed_techs = set(self.economy + self.weapon + self.armor
                               + self.misc + self.defense + self.hull)
        remaining_techs = list(all_needed_techs - set(self._tech_queue))
        self._tech_queue.extend(remaining_techs)

```

The solution is just rewrite it to preserve order.   This will change AI behavior.

Cjkjvfnby:
Requirements are not listed in the code.  Therefore TechGroup does not list all tech that should be added. 

For example TechGroup1
```
List in TechGroup                       Tech list with respect of requirements
--------------------                       --------------------------------------------
LRN_PHYS_BRAIN                          LRN_PHYS_BRAIN
GRO_PLANET_ECOL                         GRO_PLANET_ECOL
LRN_ALGO_ELEGANCE                       LRN_ALGO_ELEGANCE
GRO_SUBTER_HAB                          GRO_SUBTER_HAB
LRN_NASCENT_AI                          LRN_NASCENT_AI
                                        SHP_ROOT_AGGRESSION
SHP_WEAPON_ARC_DISRUPTOR_1              SHP_WEAPON_ARC_DISRUPTOR_1
                                        DEF_ROOT_DEFENSE
DEF_GARRISON_1                          DEF_GARRISON_1
DEF_DEFENSE_NET_1                       DEF_DEFENSE_NET_1
SHP_FIGHTERS_1                          SHP_FIGHTERS_1
                                        SHP_WEAPON_1_2
SHP_WEAPON_1_3                          SHP_WEAPON_1_3
PRO_ROBOTIC_PROD                        PRO_ROBOTIC_PROD
SHP_WEAPON_1_2                          SHP_WEAPON_1_2
SHP_WEAPON_1_4                          SHP_WEAPON_1_4
[...]

PR #3155, opened 2020-08-27
Fix tech list filling.

Cjkjvfnby:
Added stable adding remaining items implementation. 

This is a fix for the issue mentioned in https://github.com/freeorion/freeorion/issues/3143#issuecomment-676805330

I have not tested it yet, I want to get some feedback before it.  (I have not played the game for some time, so I am not an expert in AI tuning)

This PR definitely will change AI behavior, but I m not sure if it will cause any noticiblae changes.

### DIFF
--- a/default/python/AI/TechsListsAI.py
+++ b/default/python/AI/TechsListsAI.py
@@ -3,11 +3,12 @@ The TechsListAI module provides functions that describes dependencies between
 various technologies to help the AI decide which technologies should be
 researched next.
 """
-from logging import warning, debug
-from typing import List, Union
+from itertools import zip_longest
+from logging import debug, warning
+from typing import Iterator, List, Union
 
-import freeOrionAIInterface as fo  # pylint: disable=import-error
 import AIDependencies as Dep
+import freeOrionAIInterface as fo  # pylint: disable=import-error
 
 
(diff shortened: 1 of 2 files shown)


geoffthemedio:
@agrrr3 any comment?

agrrr3:
i do not completely understand what this is supposed to fix/what the problem is.

geoffthemedio:
@Cjkjvfnby could you elaborate on the purpose of this?

geoffthemedio:
@Cjkjvfnby could you elaborate on the purpose of this?

Cjkjvfnby:
> i do not completely understand what this is supposed to fix/what the problem is.

This fixes the order or adding the remaining techs.  

The old version of the code added all remaining tech in pseudo-random order by converting the `list` to `set`. This approach may provide different orders of tech on different machines.

Before [w1, w2, w3] could be added as one of `[w1, w2, w3], [w1, w3, w2], [w2, w1, w3], [w2, w3, w1], [w3, w1, w2], [w3, w2, w1]`

After changes [w1, w2, w3] will be always added in [w1, w2, w3] order. 

Since new order is one of the possible old orders, this PR does not need extra balancing.

PR #3160, opened 2020-09-04
README.md: Add link to Flathub

Siosm:


geoffthemedio:
looks reasonable... will assume it is

PR #3184, opened 2020-10-09
Haiku support proposal.

devnexen:


geoffthemedio:
What is Haiku? Googling such a well-known term is not helpful.

devnexen:
Haiku OS is BeOs inheritant.

PR #3185, opened 2020-10-10
Colo threshold sitrep takes growth rate from macro

Oberlus:
When PLC_NO_GROWTH and PLC_POPULATION effects were added to species/common/population.macros, the base rate factor was changed from 0.01 to 0.005. This wasn't updated in techs/ship_hulls/SHP_GAL_EXPLO.focs.txt, which continued to use the old 0.01 value. This causes premature SitReps about colonizing threshold.

This PR fixed that.


Oberlus:
Playtested.
Good to merge.

Issue #3188, opened 2020-10-10
NamedReal/NamedInteger value declaration precedence is odd


Oberlus:
Bug Report
==========

Environment
-----------

* **FreeOrion Version**: e8b1d40
* **Operating System**: All

Description
-----------

When writing this:
(NamedReal name = "GARRISON_2_MAXTROOPS_EFFECT" value = 2 * [[TROOPS_PER_POP]]) 

This is intended:
(NamedReal name = "GARRISON_2_MAXTROOPS_EFFECT" value = (2 * [[TROOPS_PER_POP]])) 

But this is what we get:
((NamedReal name = "GARRISON_2_MAXTROOPS_EFFECT" value = 2) * [[TROOPS_PER_POP]]) 

[Forum thread.](https://www.freeorion.org/forum/viewtopic.php?p=103678#p103678)

agrrr3:
It seems precedence is working for integer, but not for floats:

`(NamedReal name = "TWO_BY_THREE" value = 2 * 3)` registers 2.0

`(NamedReal name = "TWO_BY_THREE" value = 2.0 * 3.0)` registers 2.0

`(NamedInteger name = "TWO_BY_THREE" value = 2 * 3)` registers 6

Also it seems this precedence is working for top level named value refs, so the only thing known to be broken is name-in-the-middle Real value refs.

agrrr3:
I made a PR which seems to fix the issue for NamedReal

PR #3194, opened 2020-10-18
Use expr instead of primary_expr when parsing NamedReal in the middle…

agrrr3:
… of another valueref - seems to fix precedence

Not sure what exactly is the difference, but it seems to fix the precedence and not to break anything.

Intended to fix issue #3188 
  NamedReal/NamedInteger value declaration precedence is odd

Issue:

When writing this:
(NamedReal name = "GARRISON_2_MAXTROOPS_EFFECT" value = 2 * [[TROOPS_PER_POP]])

This is intended:
(NamedReal name = "GARRISON_2_MAXTROOPS_EFFECT" value = (2 * [[TROOPS_PER_POP]]))

But this is what we get:
((NamedReal name = "GARRISON_2_MAXTROOPS_EFFECT" value = 2) * [[TROOPS_PER_POP]])

Analysis was:

It seems precedence is working for integer, but not for floats:

(NamedReal name = "TWO_BY_THREE" value = 2 * 3) registers 2.0

(NamedReal name = "TWO_BY_THREE" value = 2.0 * 3.0) registers 2.0

(NamedInteger name = "TWO_BY_THREE" value = 2 * 3) registers 6

Also it seems this precedence is working for top level named value refs, so the only thing known to be broken is name-in-the-middle Real value refs.

### DIFF
--- a/parse/DoubleValueRefParser.cpp
+++ b/parse/DoubleValueRefParser.cpp
@@ -123,7 +123,7 @@ parse::double_parser_rules::double_parser_rules(
     named_real_valueref
         = (     tok.Named_ >> tok.Real_
             >>  label(tok.Name_) > tok.string
-             >  label(tok.Value_) > primary_expr
+             >  label(tok.Value_) > expr
           ) [
              // Register the value ref under the given name by lazy invoking RegisterValueRef
              parse::detail::open_and_register_as_string_(_2, _3, _pass),


geoffthemedio:
Based on https://github.com/freeorion/freeorion/blob/master/parse/ArithmeticRules.cpp#L202 , an `expr` is an mathematical expression, which can include various operators like `+` or `*` that depend on order of operations to resolve.

Based on https://github.com/freeorion/freeorion/blob/master/parse/DoubleValueRefParser.cpp#L142 , `primary_expr` is an indivisible expression, which may be because it's wrapped in `( ... )` or because it's a simple constant like `12` or because it's another type of valueref like a `Target.Population` or a `Statistic Sum ...` expression.

The main distinction is whether or not the expression has sub-expressions connected by math operators. So, `2 * [[TROOPS_PER_POP]]` is a valid `expr` but not a `primary_expr` due to the `*` and two sub-expressions on either side.

PR #3195, opened 2020-10-19
Named Valuerefs for Research modifiers

Oberlus:
And some others mentioned in the same stringtable entries, to not duplicate active translators work.

Depends on #3206.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -8793,9 +8793,12 @@ ANCIENT_RUINS_DEPLETED_SPECIAL
 Ancient Ruins (Excavated)
 
 ANCIENT_RUINS_DEPLETED_SPECIAL_DESC
-'''Increases [[metertype METER_TARGET_RESEARCH]] on this planet by 1 per [[metertype METER_POPULATION]] when Focus is set to Research.
+'''Increases [[metertype METER_TARGET_RESEARCH]] on this planet by [[value ANCIENT_RUINS_TARGET_RESEARCH_PERPOP]] per [[metertype METER_POPULATION]] when Focus is set to Research.
 
 This planet holds the ruins of an unknown, advanced ancient race. A valuable discovery has already been made here.'''
+
+ANCIENT_RUINS_TARGET_RESEARCH_PERPOP
+per population
 
 ANCIENT_RUINS_SPECIAL
 Ancient Ruins
@@ -8864,7 +8867,7 @@ ECCENTRIC_ORBIT_SPECIAL
 Eccentric Orbit
 
 ECCENTRIC_ORBIT_SPECIAL_DESC
-'''Increases [[metertype METER_TARGET_RESEARCH]] by 3, but [[metertype METER_SUPPLY]] is decreased by 2.
+'''Increases [[metertype METER_TARGET_RESEARCH]] by [[value ECCENTRIC_ORBIT_TARGET_RESEARCH_FLAT]], but [[metertype METER_SUPPLY]] is decreased ([[value ECCENTRIC_ORBIT_MAX_SUPPLY_FLAT]]).
 
 This planet's orbit is very eccentric. It has a large distance between its closest and farthest approaches to its star. Total insolation varies significantly throughout the year. The varied conditions inhibit the supply the planet can provide, but are a beneficial platform for research.'''
 
@@ -8885,7 +8888,7 @@ TEMPORAL_ANOMALY_SPECIAL
 Temporal Anomaly
 
 TEMPORAL_ANOMALY_SPECIAL_DESC
-'''Increases [[metertype METER_TARGET_RESEARCH]] on this planet by 5 per [[metertype METER_POPULATION]] when Focus is set to [[metertype METER_RESEARCH]]. Decreases [[metertype METER_TARGET_POPULATION]] (regardless of Focus) depending on the planet size:
+'''Increases [[metertype METER_TARGET_RESEARCH]] on this planet by [[value TEMPORAL_ANOMALY_TARGET_RESEARCH_PERPOP]] per [[metertype METER_POPULATION]] when Focus is set to [[metertype METER_RESEARCH]]. Decreases [[metertype METER_TARGET_POPULATION]] (regardless of Focus) by planet size per [[value TEMPORAL_ANOMALY_TARGET_POPULATION_PERSIZE]]:
 • [[SZ_TINY]] (-5)
 • [[SZ_SMALL]] (-10)
 • [[SZ_MEDIUM]] (-15)
@@ -8894,6 +8897,12 @@ TEMPORAL_ANOMALY_SPECIAL_DESC
 
 The flow of time on this planet is gravely disturbed. Seemingly random local occurences of increases and decreases of the pace of time, induces grave disarrangement in any living being. A social or productive life is nearly impossible. Extreme adaptability and precaution is needed to establish a colony on this planet.
 On the other hand the temporal anomaly has a huge potential for scientific experiments beyond what is possible on normal planets.'''
+
+TEMPORAL_ANOMALY_TARGET_RESEARCH_PERPOP
+per population
+
+TEMPORAL_ANOMALY_TARGET_POPULATION_PERSIZE
+per planet size
 
 RESONANT_MOON_SPECIAL
 Resonant Moon
@@ -8907,9 +8916,12 @@ COMPUTRONIUM_SPECIAL
 Computronium Moon
 
 COMPUTRONIUM_SPECIAL_DESC
-'''When a planet with a Computronium Moon is focused on [[metertype METER_RESEARCH]], all research focused planets in the empire gain 0.1 [[metertype METER_TARGET_RESEARCH]] per [[metertype METER_POPULATION]].
+'''When a planet with a Computronium Moon is focused on [[metertype METER_RESEARCH]], all research focused planets in the empire gain [[value COMPUTRONIUM_TARGET_RESEARCH_PERPOP]] [[metertype METER_TARGET_RESEARCH]] per [[metertype METER_POPULATION]].
 
 Left by some vanished civilization, the entire moon - from surface to core - is a computing device of awesome power. It can complete the most complicated simulations and equations faster than they can be entered.'''
+
+COMPUTRONIUM_TARGET_RESEARCH_PERPOP
+per population
 
 HONEYCOMB_SPECIAL
 Honeycomb
@@ -8923,7 +8935,7 @@ PHILOSOPHER_SPECIAL
 Philosopher Planet
 
 PHILOSOPHER_SPECIAL_DESC
-'''As long as the Philosopher Planet is uninhabited, all inhabited planets in the same system have their [[metertype METER_TARGET_RESEARCH]] increased by 5. [[metertype METER_TARGET_CONSTRUCTION]] on the Philosopher Planet is decreased by 20.
+'''As long as the Philosopher Planet is uninhabited, all inhabited planets in the same system have their [[metertype METER_TARGET_RESEARCH]] increased by [[value PHILOSOPHER_TARGET_RESEARCH_FLAT]]. [[metertype METER_TARGET_CONSTRUCTION]] on the Philosopher Planet is decreased ([[value PHILOSOPHER_TARGET_CONSTRUCTION_FLAT]]).
 
 The lone remnant of a planet-sized space wandering species has found his final rest orbiting in this system like a normal planet. Having lived for thousands of years he lost the will and ability to travel and resorts to pondering about the mysteries of the universe he has seen and learned about. Eccentric and slightly demented he refuses to communicate with anything that isn't the size and shape of his kin, leaving scientists no choice but to ask questions via the planet they are living on. Understandably colonizing the Philosopher Planet will disturb him gravely and he will fall silent. Construction on his surface is difficult as is to be expected from a living being.'''
 
@@ -11555,9 +11567,12 @@ LRN_ALGO_ELEGANCE
 Algorithmic Elegance
 
 LRN_ALGO_ELEGANCE_DESC
-'''Increases [[metertype METER_TARGET_RESEARCH]] on all Research focused planets by 0.05 per [[metertype METER_POPULATION]].
+'''Increases [[metertype METER_TARGET_RESEARCH]] on all Research focused planets by [[value LRN_ALGO_ELEGANCE_TARGET_RESEARCH_PERPOP]] per [[metertype METER_POPULATION]].
 
 With greater and greater difficulty of data analysis problems, traditional measures of algorithm performance become less useful due to the limitations of irreducible complexity. At this stage, other measures of algorithm form and function become significant; aesthetically and metaphorically, the elegance of the solution must be optimized.'''
+
+LRN_ALGO_ELEGANCE_TARGET_RESEARCH_PERPOP
+per population
 
 LRN_TRANSLING_THT
 Translinguistics
@@ -11577,7 +11592,7 @@ LRN_NASCENT_AI
 Nascent Artificial Intelligence
 
 LRN_NASCENT_AI_DESC
-'''Increases [[metertype METER_TARGET_RESEARCH]] on all planets by 1.
+'''Increases [[metertype METER_TARGET_RESEARCH]] on all planets by [[value LRN_NASCENT_AI_TARGET_RESEARCH_FLAT]].
 
 While traditional computers have nearly unfathomable intelligence and computational abilities, they lack the essential qualities of self-awareness, consciousness or sentience. While these traits remain elusive, reasonable facsimiles may be produced. These investigations open new avenues of research into cognitive sciences and novel metaparadigms.
 
@@ -11587,7 +11602,7 @@ LRN_ARTIF_MINDS
 Artificial Minds
 
 LRN_ARTIF_MINDS_DESC
-'''Increases [[metertype METER_TARGET_RESEARCH]] on all Research focused planets by 1.
+'''Increases [[metertype METER_TARGET_RESEARCH]] on all Research focused planets by [[value LRN_ARTIF_MINDS_TARGET_RESEARCH_FLAT]].
 
 Artificial minds are the epitome of the research on cognitive sciences. These are not facsimiles of self-awareness, consciousness or sentience, but the true attainment of such qualities within an artificial being. This sentient devices can expand on their own the limits of science and research, identifying by themselves new ways of data processing and deduction of knowledge. Each planet can contain such structures, if properly maintained and assisted by researchers.'''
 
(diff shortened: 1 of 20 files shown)


default/stringtables/en.txt
```diff
@@ -8864,10 +8867,16 @@ ECCENTRIC_ORBIT_SPECIAL
 Eccentric Orbit
 
 ECCENTRIC_ORBIT_SPECIAL_DESC
-'''Increases [[metertype METER_TARGET_RESEARCH]] by 3, but [[metertype METER_SUPPLY]] is decreased by 2.
+'''Increases [[metertype METER_TARGET_RESEARCH]] by [[value ECCENTRIC_ORBIT_MAXRESEARCH_FLAT]], but [[metertype METER_SUPPLY]] is decreased by [[value ECCENTRIC_ORBIT_MAXSUPPLY_FLAT]].
```
agrrr3: ECCENTRIC_ORBIT_MAXSUPPLY_FLAT is -2, not 2

default/stringtables/en.txt
```diff
@@ -8923,10 +8940,16 @@ PHILOSOPHER_SPECIAL
 Philosopher Planet
 
 PHILOSOPHER_SPECIAL_DESC
-'''As long as the Philosopher Planet is uninhabited, all inhabited planets in the same system have their [[metertype METER_TARGET_RESEARCH]] increased by 5. [[metertype METER_TARGET_CONSTRUCTION]] on the Philosopher Planet is decreased by 20.
+'''As long as the Philosopher Planet is uninhabited, all inhabited planets in the same system have their [[metertype METER_TARGET_RESEARCH]] increased by [[value PHILOSOPHER_MAXRESEARCH_FLAT]]. [[metertype METER_TARGET_CONSTRUCTION]] on the Philosopher Planet is decreased by [[value PHILOSOPHER_MAXCONSTRUCTION_FLAT]].
```
agrrr3: PHILOSOPHER_MAXCONSTRUCTION_FLAT is -20

agrrr3:
I saw the change for the artifical minds. Is there any other functional difference?

There is the question how to handle the malus. ECCENTRIC_ORBIT_MAXSUPPLY_FLAT, PHILOSOPHER_MAXCONSTRUCTION_FLAT, and TEMPORAL_ANOMALY_MAXPOPULATION_FLAT all have negative values but currently need the positive/absolute value in the stringtables. "increased by -20" is pretty shitty and I do not think there is a good way to express such a thing. In a perfect world there would something like a function/formatting on UI/stringtable level to have this correctly rendered; e.g. something similar to `[[increasedby-value PHILOSOPHER_MAXCONSTRUCTION_FLAT]]` which would render to `increased by 5` for 5 and to `decreased by 5` for -5.

I suggest to register the positive value as PHILOSOPHER_MAXCONSTRUCTION_MALUS and use that in the stringtables.

Else this is good.

Oberlus:
> I saw the change for the artifical minds. Is there any other functional difference?


It is not an actual functional change: `5 * [[RESEARCH_PER_POP]]` equals `1`.

I noticed that most flat bonuses for most meters do not use a [[<METER>_PER_POP]] macro. And I like that, since the flat bonuses are not PER_POP, it makes no sense to use that macro there, and it has the disadvantage that one can not rebalance flat and perpop bonuses by playing with the default values of the PER_POP macros.
So I removed that macro and updated the value accordingly.
I'm planning to do the same with any other flat bonus (not multiplied by TargetPopulation) that includes a PER_POP macro.


> There is the question how to handle the malus.
> "increased by -20" is pretty shitty

Oh, I thought it was good enough (like what we see in the Stellar Tomography or the Solar Orb. Gen., were some players believe it's gonna multiply the number by zero dot something). Anyways, fair enough.

> I suggest to register the positive value as PHILOSOPHER_MAXCONSTRUCTION_MALUS and use that in the stringtables.
I don't like that for the already commented reasons. Scripting-wise, I really want first to know from the name that it is a flat or per pop bonus rather than to know if it is positive or negative (that can be inferred from the sign of the value). 

What about these?
`Increases [[metertype METER_TARGET_RESEARCH]] by [[value ECCENTRIC_ORBIT_MAXRESEARCH_FLAT]], but reduces [[metertype METER_SUPPLY]] ([[value ECCENTRIC_ORBIT_MAXSUPPLY_FLAT]]).`

`[[metertype METER_TARGET_CONSTRUCTION]] on the Philosopher Planet is decreased ([[value PHILOSOPHER_MAXCONSTRUCTION_FLAT]]).`

What I committed for Temporal Anomaly stringtables is just wrong, I didn't finish it, I must fix it.
I'd like to provide a solution that is independent of planet size values and of the malus scripted in the special, something like this:

`Decreases [[metertype METER_TARGET_POPULATION]] by [[value TEMPORAL_ANOMALY_MAXPOPULATION_FLAT]] per planet size` and linking "planet size" to a pedia page with the habitable sizes for current settings.

agrrr3:
> I'm planning to do the same with any other flat bonus (not multiplied by TargetPopulation) that includes a PER_POP macro.

fair enough - i think the only scenario where the macro would make sense is like "a scripter wants to change all bonus at the same time by simple changing the macro". dont think that is a valid scenario.

> I don't like that for the already commented reasons. Scripting-wise, I really want first to know from the name that it is a flat or per pop bonus rather than to know if it is positive or negative (that can be inferred from the sign of the value).

I  would expect _MALUS to be used only for the flat case. For a positive value which is used to decrease a value. So in AI I would use a minus instead of a plus to calculate the effect.
_FLAT is totally ok for a positive or negative value to be added - so using a plus to calculate the effect even if it decreases the value. With the current state of things one just needs to adjust the stringtables if you change a _FLAT from positive to negative or vice versa. I find your proposal (see below) good enough.

> What about these?
> `Increases [[metertype METER_TARGET_RESEARCH]] by [[value ECCENTRIC_ORBIT_MAXRESEARCH_FLAT]], but reduces [[metertype METER_SUPPLY]] ([[value ECCENTRIC_ORBIT_MAXSUPPLY_FLAT]]).`
> 
> `[[metertype METER_TARGET_CONSTRUCTION]] on the Philosopher Planet is decreased ([[value PHILOSOPHER_MAXCONSTRUCTION_FLAT]]).`

will "break" if one only changes the value and not the stringtable entry, but works for me.  OTOH i am not native.

Oberlus:
[Forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=15&t=11806) about naming convention for Named values.

default/stringtables/en.txt
```diff
@@ -8885,7 +8894,7 @@ TEMPORAL_ANOMALY_SPECIAL
 Temporal Anomaly
 
 TEMPORAL_ANOMALY_SPECIAL_DESC
-'''Increases [[metertype METER_TARGET_RESEARCH]] on this planet by 5 per [[metertype METER_POPULATION]] when Focus is set to [[metertype METER_RESEARCH]]. Decreases [[metertype METER_TARGET_POPULATION]] (regardless of Focus) depending on the planet size:
+'''Increases [[metertype METER_TARGET_RESEARCH]] on this planet by [[value TEMPORAL_ANOMALY_TARGET_RESEARCH_PERPOP]] per [[metertype METER_POPULATION]] when Focus is set to [[metertype METER_RESEARCH]]. Decreases [[metertype METER_TARGET_POPULATION]] (regardless of Focus) by planet size per [[TEMPORAL_ANOMALY_TARGET_POPULATION_PERSIZE]]:
 • [[SZ_TINY]] (-5)
```
Oberlus, line 8892: This won't update if TEMPORAL_ANOMALY_TARGET_POPULATION_PERSIZE named value changes, although the previous line allows the reader to infer the real values (if planet sizes are known).
I've considered removing this bullet points and linking to the values of the planet sizes (not sure if there is a Pedia page for that already or not), or adding five new named values, one per planet size (and more new named values for other stringtable entries like this one with per planet size modifiers. Thoughts?

Oberlus:
All issues addressed, except the one I point out in the last comment.
Otherwise good to merge.

default/stringtables/en.txt
geoffthemedio: `[[TEMPORAL_ANOMALY_TARGET_POPULATION_PERSIZE]]` is a stringtable lookup macro, isn't it?

default/stringtables/en.txt
```diff
@@ -12439,24 +12487,39 @@ LRN_DISTRIB_THOUGHT
...
-* [[STAR_BLUE]] or [[STAR_WHITE]] (x0.1)
-* [[STAR_YELLOW]], [[STAR_ORANGE]] or [[STAR_RED]] (x0.05)
+* [[STAR_BLACK]] (x[[value LRN_STELLAR_TOMO_BLACK_TARGET_RESEARCH_PERPOP]])
+* [[STAR_NEUTRON]] (x[[value LRN_STELLAR_TOMO_NEUTRON_TARGET_RESEARCH_PERPOP]])
+* [[STAR_BLUE]] or [[STAR_WHITE]] (x[[value LRN_STELLAR_TOMO_BRIGHT_TARGET_RESEARCH_PERPOP]])
+* [[STAR_YELLOW]], [[STAR_ORANGE]] or [[STAR_RED]] (x[[value LRN_STELLAR_TOMO_DIME_TARGET_RESEARCH_PERPOP]])
 
 A network of specialized satellites can use tomographic methods to reconstruct the activities occurring deep within a star. These otherwise costly-to-produce conditions are ideal for a variety of research endeavors, with larger benefits for more exotic star types.'''
 
+LRN_STELLAR_TOMO_BLACK_TARGET_RESEARCH_PERPOP
```
geoffthemedio, line 12487: what are these repeated strings for...? is it necessary to have so many different stringtable entries that contain the same "per population" text?

default/stringtables/en.txt
Oberlus: > `[[TEMPORAL_ANOMALY_TARGET_POPULATION_PERSIZE]]` is a stringtable lookup macro, isn't it?

I don't think so, but I don't have the concepts clear.
AFAIK, it is a NamedReal defined in https://github.com/freeorion/freeorion/blob/0ce3083f5f5b351f2667af7a48c7d76a2332a4a5/default/scripting/specials/planet/TEMPORAL_ANOMOLY.focs.txt#L20

For some reason that @agrrr3 might understand (I don't), it is not enough to declare the named value in the FOCS file: each named value used in the stringtables must have a corresponding stringtable entry (macro?) with a text that will show in the tootips over the stringtable calls to `[[value <named value name>]]` and in the [[Test - NamedValues]] Pedia page.

default/stringtables/en.txt
Oberlus: Since Stellar Tomo has different values for each star type group, one named value per star type group is required. And hence, as per the previous comment about how named values work in the stringtables, you need to declare one stringtable entry (macro?) per each referenced named value.

It is indeed quite annoying. I would very much prefer that the mere inclusion of `[[value <named_value_name>]]` was the only thing needed. @agrrr3 ?

default/stringtables/en.txt
geoffthemedio: There might be a "TEMPORAL_ANOMALY_TARGET_POPULATION_PERSIZE" named valueref, but here you're just doing a stringtable substitution macro as far as I can tell. There is no `value` in `[[TEMPORAL_ANOMALY_TARGET_POPULATION_PERSIZE]]`.

Oberlus: The line you see is before I fixed it with the latest force-push. Current state of the PR has that missing "value" fixed:
https://github.com/freeorion/freeorion/blob/0ce3083f5f5b351f2667af7a48c7d76a2332a4a5/default/stringtables/en.txt#L8897

default/stringtables/en.txt
geoffthemedio: What is the purpose of these strings? Presumably it's to explain as text what the number that the valueref calculates is for or how it is calculated? Your stringtable entries just contain "per population" which doesn't explain anything.

@agrrr3 What happens if no stringtable entry is provided for a named valueref? If that is treated as an error, can it instead behave as if an emptry string was added to the stringtable so all these extra string entries need not be created when there's nothing relevant to explain in the stringtable? Are they needed for the UI to work, in which case presumably the current "per population" strings aren't explaining what they should be?

Oberlus: I don't know about the purpose or implementation reasons for the entries like
```
LRN_STELLAR_TOMO_BLACK_TARGET_RESEARCH_PERPOP
per population
```
or 
```
BLD_AUTO_HISTORY_ANALYSER_TARGET_RESEARCH_FLAT
''''''
```
I only know Pedia pages with named values will not work if you don't include these kind of entries in the stringtable.



> Your stringtable entries just contain "per population" which doesn't explain anything.

When parsing named values in the AI code, or when working with FOCS scripts, having that kind of extra info in the named value names allows to differentiate between flat bonuses/maluses (`Value = Value + <Named value>`), population based bonuses7maluses (`Value = Value + Target.TargetPopulation * <Named value>`) and scaling bonuses (`Value = Value * <Named value>`).
The same for indicating the affected meter. By including that info in the named value names, the AI code could parse (e.g.) TargetResearch values including "TARGET_RESEARCH" in their names and automatically be aware of every tech affecting that meter regardless of the FOCS content scripters adjusting the AI python code (assuming content scripters do follow that naming convention).

geoffthemedio: > By including that info in the named value names, the AI code could parse

I'm referring to the "per population" contents of the string, which would normally be localized, not the name (or "key") of the stringtable entry. There are arguments for having a semi-standardized format for naming of valuerefs, but that doesn't have anything to do with the contents of the stringtable.

default/stringtables/en.txt
geoffthemedio: OK.

Regarding the original question, a proper phrasing in the description before the list of planet sizes should be adequate. Could add a bunch more value-refs, but they'd also be a maintainance issue if not actually used in the FOCS that generates the relevant effects in-game.

default/stringtables/en.txt
Oberlus: Oh, right. I don't have an informed answer for that.
At first thought writing `''''''` for all those strings since I didn't care about them actually (seem to me like a necessary evil), but I tried to follow @agrrr3 suggestions for the UI. Those "per population" appear there also, although they don't seem necessary there neither since you can see the string names in there so you know if they are per pop. or not.

agrrr3: > > By including that info in the named value names, the AI code could parse
> 
> I'm referring to the "per population" contents of the string, which would normally be localized, not the name (or "key") of the stringtable entry. There are arguments for having a semi-standardized format for naming of valuerefs, but that doesn't have anything to do with the contents of the stringtable.

This is a bit conincedental, as the tags did not work without the entries. Then I tried to use the entries for providing description of the calculation and of the intended context ("per population"). Oberlus has a UI suggestion how to better use the context.

If we want to keep this, we should probably use a macro for the stringtable contents  `[[NAMED_VALUE_PER_POP]]` so it works well with localisation

Oberlus: > the tags did not work without the entries

Is it easy to understand why?

agrrr3: > > the tags did not work without the entries
> 
> Is it easy to understand why?

Not for me. Somewhere a UserString lookup gets performed, but i couldnt pinpoint it. UI/LinkText.cpp LinkTaggedText looked promising as format fits the one i get handed, but it is only used in the pedia panel
```
std::string LinkTaggedText(const std::string& tag, const std::string& stringtable_entry)
{ return "<" + tag + " " + stringtable_entry + ">" + UserString(stringtable_entry) + "</" + tag + ">"; }
```

Oberlus: I guess I won't work on introducing more named values until the issue about stringtable cluttering from the unwanted stringtable entries is solved or accepted as a necessary evil.

agrrr3: I implemented a workaround to ignore missing stringtable entries for named value refs. See #3206

agrrr3: So for me it would be ok not to add the empty entries.

Oberlus: I'll remove them then. Less lines in the stringtable files.
I'll wait for PR  #3206 to be merged.

Oberlus: With the empty strings removed, the lint check fails (referenced keys not found). I guess it will work once #3206 is merged.

agrrr3:
The linter is unrelated (and does not know about the named value refs), so
this will still fail.

Oberlus:
> The linter is unrelated (and does not know about the named value refs), so this will still fail.

`default/stringtables/en.txt:8869: Referenced key 'ECCENTRIC_ORBIT_TARGET_RESEARCH_FLAT' in value of 'ECCENTRIC_ORBIT_SPECIAL_DESC' was not found.`

`ECCENTRIC_ORBIT_TARGET_RESEARCH_FLAT` is one of those empty strings (`''''''`) that I removed as per the above discussion. If I understand you correctly, linter would fail for everyone once the removed empty strings get into master? Would that mean that linter need fixing, or that these PRs can't be merged yet and discussion about empty strings keeps open?

agrrr3:
> If I understand you correctly, linter would fail for everyone once the removed empty strings get into master? Would that mean that linter need fixing, or that these PRs can't be merged yet and discussion about empty strings keeps open?

I am adding support of "optional" keys to the linter (and make the [[value XXX]] tags optional). This probably means that this PR depends on #3206 to make the CI happy

Oberlus:
Good to merge.

geoffthemedio:
````
22:19:03.371830 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: BLD_AUTO_HISTORY_ANALYSER_TARGET_RESEARCH_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.371830 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: BLD_CULTURE_ARCHIVES_TARGET_RESEARCH_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.371830 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: BLD_CULTURE_LIBRARY_TARGET_RESEARCH_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.372831 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: ECCENTRIC_ORBIT_TARGET_RESEARCH_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.372831 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: ECCENTRIC_ORBIT_MAX_SUPPLY_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.373830 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: GRO_ENERGY_META_MAX_DEFENSE_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.373830 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: GRO_ENERGY_META_MAX_SHIELD_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.373830 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: LRN_ARTIF_MINDS_TARGET_RESEARCH_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.373830 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: LRN_NASCENT_AI_TARGET_RESEARCH_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.374831 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: PHILOSOPHER_TARGET_RESEARCH_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.374831 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: PHILOSOPHER_TARGET_CONSTRUCTION_FLAT in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
22:19:03.375831 {0x00001e10} [error] client : stringtable.cpp:248 : Unresolved reference: SHP_REINFORCED_HULL_BONUS in: C:/Users/g_top/Desktop/FOSDK11/FreeOrionAlt/default/stringtables/en.txt.
````

agrrr3:
@geoffthemedio is something up with your workspace? the CI lint went fine

geoffthemedio:
Not CI lint errors. These are runtime errors in the logs.

PR #3196, opened 2020-10-21
Use std::array instead of GG::Clr in common and server code

o01eg:
Removes usage of GG in the non-client code.

### DIFF
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -569,10 +569,10 @@ void HumanClientApp::NewSinglePlayerGame(bool quickstart) {
 
     // DB stores index into array of available colours, so need to get that array to look up value of index.
     // if stored value is invalid, use a default colour
-    const std::vector<GG::Clr>& empire_colours = EmpireColors();
+    const std::vector<EmpireColor>& empire_colours = EmpireColors();
     int colour_index = GetOptionsDB().Get<int>("setup.empire.color.index");
     if (colour_index >= 0 && colour_index < static_cast<int>(empire_colours.size()))
-        human_player_setup_data.empire_color = empire_colours[colour_index];
+        human_player_setup_data.empire_color = GG::Clr(empire_colours[colour_index]);
     else
         human_player_setup_data.empire_color = GG::CLR_GREEN;
 
--- a/client/human/HumanClientApp.cpp
+++ b/client/human/HumanClientApp.cpp
@@ -572,9 +572,9 @@ void HumanClientApp::NewSinglePlayerGame(bool quickstart) {
     const std::vector<EmpireColor>& empire_colours = EmpireColors();
     int colour_index = GetOptionsDB().Get<int>("setup.empire.color.index");
     if (colour_index >= 0 && colour_index < static_cast<int>(empire_colours.size()))
-        human_player_setup_data.empire_color = GG::Clr(empire_colours[colour_index]);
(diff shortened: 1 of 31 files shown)


CMakeLists.txt
```diff
@@ -684,18 +684,19 @@ endif()
...
 ##
 
-add_subdirectory(client)
-add_subdirectory(combat)
-add_subdirectory(Empire)
-add_subdirectory(network)
-add_subdirectory(parse)
-add_subdirectory(python)
-add_subdirectory(server)
+add_subdirectory(src/common/client)
```
geoffthemedio: can you separate the source layout changes?

o01eg: It's here #3137

o01eg: Removed sources layout changes.

server/UniverseGenerator.cpp
```diff
@@ -875,18 +875,20 @@ void InitEmpires(const std::map<int, PlayerSetupData>& player_setup_data) {
...
             } else {
                 // as a last resort, make up a colour
-                empire_colour = GG::FloatClr(static_cast<float>(RandZeroToOne()),
-                                             static_cast<float>(RandZeroToOne()),
-                                             static_cast<float>(RandZeroToOne()),
-                                             1.0f);
+                empire_colour = {RandInt(0, 255),
+                                 RandInt(0, 255),
+                                 RandInt(0, 255),
+                                 255};
```
geoffthemedio, line 891: https://ci.appveyor.com/project/freeorion/freeorion/builds/35875836/job/78bhaug75kmcby0b#L275
````
error C2397: conversion from 'int' to '_Ty' requires a narrowing conversion [C:\projects\freeorion\msvc2017\FreeOrionD\FreeOrionD.vcxproj]
          with
          [
              _Ty=unsigned char
          ]
````

Comment (o01eg) [reply server/UniverseGenerator.cpp:891]: Should be fixed with `static_cast`.

**Comment by geoffthemedio** [comment]
Can this load saves that serialized `GG::Clr`?

**Comment by o01eg** [comment]
@geoffthemedio No, that why I asked about breaking compatibility on forum.

**Comment by geoffthemedio** [comment]
That needs to be fixed for the medium term. After the next version, the compatibility can be dropped, but it should load old saves for now. Workaround could involve a local `Clr` definition in the serialization code, if that helps....

**Comment by o01eg** [comment]
@geoffthemedio Restored compatibility.

File: client/human/HumanClientApp.cpp [hunk 3]
```diff
@@ -569,12 +569,12 @@ void HumanClientApp::NewSinglePlayerGame(bool quickstart) {
...
     // DB stores index into array of available colours, so need to get that array to look up value of index.
     // if stored value is invalid, use a default colour
-    const std::vector<GG::Clr>& empire_colours = EmpireColors();
+    const std::vector<EmpireColor>& empire_colours = EmpireColors();
     int colour_index = GetOptionsDB().Get<int>("setup.empire.color.index");
     if (colour_index >= 0 && colour_index < static_cast<int>(empire_colours.size()))
         human_player_setup_data.empire_color = empire_colours[colour_index];
     else
-        human_player_setup_data.empire_color = GG::CLR_GREEN;
+        human_player_setup_data.empire_color = {GG::CLR_GREEN.r, GG::CLR_GREEN.g, GG::CLR_GREEN.b, GG::CLR_GREEN.a};
```
geoffthemedio, line 577: should probably add a conversion operator later

PR #3197, opened 2020-10-21
Removes usage of GG_ENUM in common and server code.

o01eg:
Backed by @adrianbroher's enum macros in https://github.com/freeorion/freeorion/commit/d6142533e1832949cd5f631a0e687f8464bf0011


Empire/Diplomacy.h
```diff
@@ -3,17 +3,18 @@
...
 //! diplomatic statuses
-GG_ENUM(DiplomaticStatus,
-    INVALID_DIPLOMATIC_STATUS = -1,
-    DIPLO_WAR,
-    DIPLO_PEACE,
-    DIPLO_ALLIED,
-    NUM_DIPLO_STATUSES
+FO_ENUM(
+    (DiplomaticStatus),
+    ((INVALID_DIPLOMATIC_STATUS, "INVALID_DIPLOMATIC_STATUS", -1))
```
geoffthemedio: why do the invalid type enums need explicit string representations when the rest don't?

UI/ClientUI.h
```diff
@@ -8,7 +8,9 @@
 #include <boost/date_time/posix_time/posix_time.hpp>
 
 #include "SaveFileDialog.h"
-#include "../universe/EnumsFwd.h"
+#include "../universe/Enums.h"
```
geoffthemedio: universe/EnumsFwd.h was a lot lighter than universe/Enums.h is. Can the forward declare only alternative be retained?

util/Enum.h
```diff
@@ -0,0 +1,161 @@
...
+
+/** @brief Implementation detail for FO_ENUM */
+#define FO_DEF_ENUM_VALUE(r, data, elem) \
+    BOOST_PP_IF(BOOST_PP_EQUAL(BOOST_PP_TUPLE_SIZE(elem), 3), \
+        BOOST_PP_TUPLE_ELEM(0, elem) = BOOST_PP_TUPLE_ELEM(2, elem), \
+        BOOST_PP_TUPLE_ELEM(0, elem)),
+
+/** @brief Implementation detail for FO_ENUM */
+#define FO_DEF_ENUM(typeName, values) \
+enum \
```
geoffthemedio: can this define `enum class`, like `GG_ENUM`, instead of old style `enum`s?

Empire/Diplomacy.h
o01eg: I suppose it because 2-tuple is used to define identifier and string representation. I suppose we doesn't need it.

geoffthemedio: > I suppose it because 2-tuple is used to define identifier and string representation.

Not sure what your point is... You can optionally include a string representation for each enum value, which you're doing here just for the invalid enum values. The question is why are you doing that? It seemed intentional...

o01eg: Fixed

util/Enum.h
o01eg: Fixed.

UI/ClientUI.h
o01eg: Restored it back.

geoffthemedio:
numerous errors in MSVC builds, eg: https://ci.appveyor.com/project/freeorion/freeorion/builds/35921193/job/sgmfe306iw3w7ck9#L7516

o01eg:
@geoffthemedio I'm trying to get which changes caused it.

o01eg:
@geoffthemedio Fixed it.

PR #3202, opened 2020-10-23
Initial Godot Android support

o01eg:
Builds only libraries as Android supposed to used them from java application.
Directories code doesn't work but it's better than nothing.

Forum thread: https://freeorion.org/forum/viewtopic.php?f=24&t=5593

Upd: I switched it to Godot branch and I've managed to build GDNative library for Android and export apk from Godot Editor.

Follows up: #3237


geoffthemedio:
> Directories code doesn't work

How so?

o01eg:
> How so?

I didn't test it, only compiled it, but I know directories on Android is very different from other UNIX systems.

o01eg:
Rebased.

geoffthemedio:
I have no useful input / comment on this.

o01eg:
Log from device: [freeorion-godot.log](https://github.com/freeorion/freeorion/files/5592754/freeorion-godot.log)

Vezzra (commented):
@o01eg, I'm afraid I'm out of my depth here. I can see nothing that looks obviously wrong or buggy to me, so I'm just going to assume that you know what you're doing.

My only question would be why invest so much effort into Android support at this point? Do you need that for your dev work perhaps?

o01eg:
@Vezzra No, I was interested in it long before, but Godot opened a way to get working application to test it.

agrrr3:
> @Vezzra No, I was interested in it long before, but Godot opened a way to get working application to test it.

I am also totally intrigued. Porting to android is a bit pointless with our legacy client. But now when switching to another framework which supports android, it totally makes sense to explore it right now.

Being able to check the empire state, do simple orders and press the next turn button is a heaven-sent to speed up slow game multiplayer (as long as thinksome is not in, actually he is speeding up).

o01eg:
> I am also totally intrigued. Porting to android is a bit pointless with our legacy client.

@agrrr3 No, it also possible with SDL Android support but GG need to support GLES.

PR #3208, opened 2020-10-26
Server's python refactoring

o01eg:
Moves python initialization before background parser so it could be used if Python became FOCS replacement.


geoffthemedio:
What does this have to do with "Small server's"?

o01eg:
@geoffthemedio Nothing, I meant it a small change of code.

PR #3210, opened 2020-10-28
case sensitive value refs

agrrr3:
This PR intends to implement case sensitive parsing in FOCS.

[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=9&t=11816)

State: halfway through (?) - opened the PR in order to trigger build and to create awareness


agrrr3:
how do we proceed with this? cant really playtest much, probably some corner cases will be broken. squash and merge?

agrrr3:
Also rules are:
* labels/parameter names lowercase (e.g. name = "BLA", buildcost)
* definition names CamelCase (e.g. ShipDesign)
* containers are CamelCase Source, Target, *Candidate
* valuerefs are CamelCaseID - e.g. Source.X, Target.DesignID - also immediate value flags Value and Value()
* arithmetic operations are lowercase (abs,min,max, ceil, round, cos, log, sign...) - may be these should be different. Max(enums) is not treated as arithmetics but like valuerefs.
* conditionas are CamelCase, e.g. Ship, HasSpecialSinceTurn
* effects are CamelCase, e.g. SetEmpireMeter
* statistics are CamelCase - e.g. Statistics Max - could also be upper Statistics MAX
* And, Or, Not - could also be AND, OR, NOT - especially NOT could make sense as it is easy to miss
* Enums are CamelCase: Hostile, Tiny, GasGiant, Basic,...


probably i should change combatTargets to combattargets

geoffthemedio:
> how do we proceed with this? cant really playtest much, probably some corner cases will be broken.

I don't think much "play" testing is needed. Main thing is whether the parse is successful, and are all the checksums the same. The classes created and enum values set shouldn't have changed, and I don't think you changed anything in ValueRefs.cpp so all the matched strings should be consistent for anything that was working...

> squash and merge?

No need to squash. Just needs to be verified that everything parses correctly now.

agrrr3:
> I don't think much "play" testing is needed. Main thing is whether the parse is successful, and are all the checksums the same. The classes created and enum values set shouldn't have changed, and I don't think you changed anything in ValueRefs.cpp so all the matched strings should be consistent for anything that was working...

everything parses. but how do i check the checksums?

geoffthemedio:
I guess comment out https://github.com/freeorion/freeorion/blob/master/client/human/HumanClientApp.cpp#L478 rebuild, quickstart, and check the server log for lines with `checksum` in them, like
````
22:08:01.698528 {0x000024e0} [debug] server : namedvaluerefmanager.cpp:111 : NamedValueRefManager checksum: 2045214
22:08:01.734536 {0x000024e0} [debug] server : shiphull.cpp:296 : ShipHullManager checksum: 6216434
22:08:01.756541 {0x000024e0} [debug] server : shippart.cpp:444 : ShipPartManager checksum: 439320
22:08:01.756541 {0x000024e0} [debug] server : shipdesign.cpp:865 : PredefinedShipDesignManager checksum: 862040
22:08:01.908575 {0x000024e0} [debug] server : species.cpp:588 : SpeciesManager checksum: 7766527
22:08:01.920578 {0x000024e0} [debug] server : tech.cpp:682 : TechManager checksum: 4875169
````

o01eg:
Unittests are also failing in the case of mismatched checksums.

geoffthemedio:
There wouldn't be a mismatch if the parser used by both client and server are modified the same way...

agrrr3:
ok, checksums are the same if applying the PR or not

so ready to merge i guess

agrrr3:
got merge conflicts. will rebase and repush

agrrr3:
good to go IMHO

PR #3212, opened 2020-10-30
Cleanup GiGi usage from MSVC project

o01eg:


geoffthemedio:
I get
`\util\process.cpp(15): fatal error C1083: Cannot open include file: 'GG/utf8/checked.h': No such file or directory`

o01eg:
@geoffthemedio Replaced with `codecvt` conversion.

Issue #3182, opened 2020-10-07
Server-side auto-turn mode


o01eg:
Feature request
===============

Idea
----

Allow player to disconnect from the game and left his empire to make auto-turns. We can allow him to set number of auto-turns empires will made and disable auto-turns if he enters back to the game.
Information about it should be stored in saves so it will continue to work in case of restarting server.

Gain
----

In MP games player could leave the game only if he concedes its empire or delegate its empire to someone else. The first has disadvantages in additional resources for neighbors enemies, even leftover empire's assets don't act as defense because effect processing doesn't work with eliminated empires. The second highly depends on other people and could be impossible if no one will to take over other empire.
If the player simply never connects to the game it tooks turn to last all available time which slows the game.

Discussion
----------

https://freeorion.org/forum/viewtopic.php?f=6&t=11792

Oberlus:
Aside from the convenience for active players in MP games, recently this has been suggested as a solution for players of MP games that can't play any more (better than current concede mechanics that are full of quirks).

o01eg said he has trouble fixing the UI part. Can anyone chime in?

PR #3213, opened 2020-10-31
Introduce server-side auto-turn

o01eg:
Server-side part for https://github.com/freeorion/freeorion/issues/3182

Adds counter which make empire automatically ready:
- -1 means infinity auto-turn.
- 0 means no auto-turn.

Other values are automatically decreasing each turn.

### DIFF
--- a/Empire/Empire.cpp
+++ b/Empire/Empire.cpp
@@ -1017,6 +1017,18 @@ bool Empire::Ready() const
 void Empire::SetReady(bool ready)
 { m_ready = ready; }
 
+void Empire::AutoTurnSetReady() {
+    if (m_auto_turn_count > 0) {
+        m_auto_turn_count --;
+    }
+    SetReady(m_auto_turn_count != 0);
+}
+
+void Empire::SetAutoTurn(int turns_count) {
+    m_auto_turn_count = turns_count;
+    SetReady(m_auto_turn_count != 0);
+}
+
 void Empire::UpdateSystemSupplyRanges(const std::set<int>& known_objects) {
     //std::cout << "Empire::UpdateSystemSupplyRanges() for empire " << this->Name() << std::endl;
     m_supply_system_ranges.clear();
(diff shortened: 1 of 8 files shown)


geoffthemedio:
Looks mostly fine so far, though does need a UI.

I wonder whether this and "ready" status fit as part of the empire, rather than the player, though.

o01eg:
> I wonder whether this and "ready" status fit as part of the empire, rather than the player, though.

Because player could connect and disconnect so it won't be same other time it reconnects.

server/ServerFSM.cpp
```diff
@@ -3349,14 +3414,14 @@ sc::result ProcessingTurn::react(const ProcessTurn& u) {
     server.PostCombatProcessTurns();
 
     // update players that other empires are now playing their turn
-    for (const auto& empire : server.Empires()) {
+    for (auto& empire : server.Empires()) {
```
agrrr3: shouldnt that still be `const auto& empire` ?
Ready() has a const function body.

Empire/Empire.cpp
```diff
@@ -1017,6 +1017,18 @@ bool Empire::Ready() const
 void Empire::SetReady(bool ready)
 { m_ready = ready; }
```
agrrr3, line 1018: I did not check all the SetReady code paths, but would it possible that a client sends further messages after sending AUTO_TURN ?

I mean like pressing the turn button (or reentering the game) in order to signal that you are not ready anymore/want to change orders. In that case m_auto_turn_count needs to be set to zero i guess.

o01eg: Yes, if player connected and made orders auto-turn should be disabled.

agrrr3: > Yes, if player connected and made orders auto-turn should be disabled.

Ok; Probably this should be handled on the server and not the client.
m_auto_turn_count only gets set zero only when getting AUTO_TURN message from client, so I guess that means it is missing

server/ServerFSM.cpp

Empire/Empire.cpp
agrrr3: a thanks, I see the newest commit should fix this

o01eg:
I wonder if it's good idea to add corresponding button and entry to `InGameMenu` window.

PR #3216, opened 2020-11-01
Named Values for most Industry modifiers

Oberlus:
Follow up of #3195

### DIFF
--- a/default/scripting/buildings/CONC_CAMP.focs.txt
+++ b/default/scripting/buildings/CONC_CAMP.focs.txt
@@ -23,7 +23,9 @@ BuildingType
                 OwnedBy empire = Source.Owner
             ]
             priority = [[TARGET_AFTER_SCALING_PRIORITY]]
-            effects = SetTargetIndustry value = Value + Target.Population * 10 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
+                        * (NamedReal name = "BLD_CONC_CAMP_TARGET_INDUSTRY_PERPOP"
+                                     value = 10 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
             scope = And [
@@ -55,7 +57,7 @@ BuildingType
                     icon = "icons/building/concentration-camp.png"
                     parameters = tag = "planet" data = Source.PlanetID
                     empire = Source.Owner
-                ]
+            ]
 
         EffectsGroup
             scope = And [
@@ -69,7 +71,8 @@ BuildingType
             ]
             priority = [[CONCENTRATION_CAMP_PRIORITY]]
             effects = [
-                SetPopulation value = min(Value, Target.Population - 3)
+                SetPopulation value = min(Value, Target.Population + (NamedReal
+                                                 name = "BLD_CONC_CAMP_POPULATION_FLAT" value = -3))
                 SetIndustry value = Target.TargetIndustry
             ]
 
--- a/default/scripting/buildings/BLACK_HOLE_POW_GEN.focs.txt
+++ b/default/scripting/buildings/BLACK_HOLE_POW_GEN.focs.txt
@@ -24,7 +24,9 @@ BuildingType
             activation = Star type = BlackHole
             stackinggroup = "BLD_BLACK_HOLE_POW_GEN_PRIMARY_EFFECT"
             priority = [[TARGET_AFTER_SCALING_PRIORITY]]
-            effects = SetTargetIndustry value = Value + Target.Population * 1.0 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
+                        * (NamedReal name = "BLD_BLACK_HOLE_POW_GEN_TARGET_INDUSTRY_PERPOP"
+                                     value = 1.0 * [[INDUSTRY_PER_POP]])
     ]
     icon = "icons/building/blackhole.png"
 
--- a/default/scripting/buildings/GAS_GIANT_GEN.focs.txt
+++ b/default/scripting/buildings/GAS_GIANT_GEN.focs.txt
@@ -25,7 +25,9 @@ BuildingType
             activation = ContainedBy Population high = 0
             stackinggroup = "GAS_GIANT_GEN_STACK"
             priority = [[TARGET_AFTER_2ND_SCALING_PRIORITY]]
-            effects = SetTargetIndustry value = Value + 25 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value
+                          + (NamedReal name = "BLD_GAS_GIANT_GEN_OUTPOST_TARGET_INDUSTRY_FLAT"
+                                       value = 25 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
             scope = And [
@@ -38,7 +40,9 @@ BuildingType
             activation = ContainedBy Not Population high = 0
             stackinggroup = "GAS_GIANT_GEN_STACK"
             priority = [[TARGET_AFTER_2ND_SCALING_PRIORITY]]
-            effects = SetTargetIndustry value = Value + 15 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value
+                          + (NamedReal name = "BLD_GAS_GIANT_GEN_COLONY_TARGET_INDUSTRY_FLAT"
+                                       value = 15 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
             scope = Source
--- a/default/scripting/buildings/HYPER_DAM.focs.txt
+++ b/default/scripting/buildings/HYPER_DAM.focs.txt
@@ -27,7 +27,9 @@ BuildingType
             ]
             stackinggroup = "BLD_HYPER_DAM_BONUS"
             priority = [[TARGET_AFTER_2ND_SCALING_PRIORITY]]
-            effects = SetTargetIndustry value = Value + Target.Population * 1.0 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
+                        * (NamedReal name = "BLD_HYPER_DAM_TARGET_INDUSTRY_PERPOP"
+                                     value = 1.0 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
             scope = And [
@@ -43,7 +45,8 @@ BuildingType
             ]
             stackinggroup = "BLD_HYPER_DAM_MALUS"
             priority = [[TARGET_POPULATION_AFTER_SCALING_PRIORITY]]
-            effects = SetTargetPopulation value = Value - Target.HabitableSize
+            effects = SetTargetPopulation value = Value + Target.HabitableSize
+                        * (NamedReal name = "BLD_HYPER_DAM_TARGET_POPULATION_PERSIZE" value = -1)
     ]
     icon = "icons/building/blackhole.png"
 
--- a/default/scripting/buildings/INDUSTRY_CENTER.focs.txt
+++ b/default/scripting/buildings/INDUSTRY_CENTER.focs.txt
@@ -25,7 +25,9 @@ BuildingType
                 Not OwnerHasTech name = "PRO_INDUSTRY_CENTER_II"
             ]
             stackinggroup = "INDUSTRY_CENTER_STACK"
-            effects = SetTargetIndustry value = Value + Target.Population * 0.25 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
+                        * (NamedReal name = "BLD_INDUSTRY_CENTER_1_TARGET_INDUSTRY_PERPOP"
+                                     value = 0.25 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
             scope = And [
@@ -39,7 +41,9 @@ BuildingType
                 Not OwnerHasTech name = "PRO_INDUSTRY_CENTER_III"
             ]
             stackinggroup = "INDUSTRY_CENTER_STACK"
-            effects = SetTargetIndustry value = Value + Target.Population * 0.5 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
+                        * (NamedReal name = "BLD_INDUSTRY_CENTER_2_TARGET_INDUSTRY_PERPOP"
+                                     value = 0.5 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
             scope = And [
@@ -50,7 +54,9 @@ BuildingType
             ]
             activation = OwnerHasTech name = "PRO_INDUSTRY_CENTER_III"
             stackinggroup = "INDUSTRY_CENTER_STACK"
-            effects = SetTargetIndustry value = Value + Target.Population * 0.75 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
+                        * (NamedReal name = "BLD_INDUSTRY_CENTER_3_TARGET_INDUSTRY_PERPOP"
+                                     value = 0.75 * [[INDUSTRY_PER_POP]])
     ]
     icon = "icons/tech/industrial_centre_ii.png"
 
--- a/default/scripting/buildings/SOL_ORB_GEN.focs.txt
+++ b/default/scripting/buildings/SOL_ORB_GEN.focs.txt
@@ -23,7 +23,9 @@ BuildingType
             activation = Star type = [Blue White]
             stackinggroup = "BLD_SOL_ORB_GEN_EFFECT"
             priority = [[TARGET_AFTER_2ND_SCALING_PRIORITY]]
-            effects = SetTargetIndustry value = Value + Target.Population * 0.75 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
+                        * (NamedReal name = "BLD_SOL_ORB_GEN_BRIGHT_TARGET_INDUSTRY_PERPOP"
+                                     value = 0.75 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
             scope = And [
@@ -42,7 +44,9 @@ BuildingType
             ]
             stackinggroup = "BLD_SOL_ORB_GEN_EFFECT"
             priority = [[TARGET_AFTER_2ND_SCALING_PRIORITY]]
-            effects = SetTargetIndustry value = Value + Target.Population * 0.5 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
+                        * (NamedReal name = "BLD_SOL_ORB_GEN_NORMAL_TARGET_INDUSTRY_PERPOP"
+                                     value = 0.5 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
             scope = And [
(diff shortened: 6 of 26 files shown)


Oberlus:
This is not good to merge.

Runtime errors (not needed to start a game, just run the client):

```
17:40:38.571388 {0x00007fdddc819700} [debug] server : NamedValueRefManager.cpp:204 : ctor(NamedRef<T>): N8ValueRef8NamedRefIdEE value_ref_name: ANCIENT_RUINS_TARGET_RESEARCH_PERPOP
17:40:38.571466 {0x00007fdddc819700} [debug] server : NamedValueRefManager.cpp:263 : NamedRef<T>::GetValueRef() look for registered valueref for "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP"
17:40:38.571496 {0x00007fdddc819700} [error] server : NamedValueRefManager.cpp:56 : NamedValueRefManager::GetValueRef found no registered double valueref for "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP"
17:40:38.571509 {0x00007fdddc819700} [debug] server : NamedValueRefManager.cpp:263 : NamedRef<T>::GetValueRef() look for registered valueref for "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP"
17:40:38.571519 {0x00007fdddc819700} [error] server : NamedValueRefManager.cpp:56 : NamedValueRefManager::GetValueRef found no registered double valueref for "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP"
17:40:38.571529 {0x00007fdddc819700} [debug] server : NamedValueRefManager.cpp:263 : NamedRef<T>::GetValueRef() look for registered valueref for "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP"
17:40:38.571538 {0x00007fdddc819700} [error] server : NamedValueRefManager.cpp:56 : NamedValueRefManager::GetValueRef found no registered double valueref for "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP"
17:40:38.571565 {0x00007fdddc819700} [debug] server : NamedValueRefManager.cpp:263 : NamedRef<T>::GetValueRef() look for registered valueref for "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP"
17:40:38.571578 {0x00007fdddc819700} [error] server : NamedValueRefManager.cpp:56 : NamedValueRefManager::GetValueRef found no registered double valueref for "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP"
17:40:38.571587 {0x00007fdddc819700} [error] server : NamedValueRefManager.cpp:257 : Unexpected call of SetTopLevelContent(ANCIENT_RUINS_DEPLETED_SPECIAL) on a NamedRef - unexpected because no value ref registered yet. This should not happen
```
It repeats with some changes several times in the log.

These are different:
```
17:40:38.579269 {0x00007fddde81d700} [info] server : NamedValueRefManager.cpp:132 : Register double valueref for BLD_SOL_ORB_GEN_DIM_TARGET_INDUSTRY_PERPOP: target's Population * 0.25 * 0.20
17:40:38.633368 {0x00007fddde81d700} [error] server : NamedValueRefManager.cpp:140 : Currently only invariant value refs can be named. BLD_SOL_ORB_GEN_DIM_TARGET_INDUSTRY_PERPOP
[...]
17:40:38.633554 {0x00007fdddd01a700} [debug] server : NamedValueRefManager.cpp:135 : Number of registered double ValueRefs: 21
17:40:38.633585 {0x00007fddc7fff700} [error] server : NamedValueRefManager.cpp:140 : Currently only invariant value refs can be named. BASIC_DEFENSE_TROOPS_MAX_TROOPS_PERPOP
```
Several other instances as well, for different named values.

@agrrr3 do you know what should I do to fix those runtime errors?

default/scripting/buildings/CONC_CAMP.focs.txt
```diff
@@ -39,12 +42,13 @@ BuildingType
...
         EffectsGroup
             scope = Source
-            activation =    ContainedBy And [
-                                Object id = Source.PlanetID
-                                Population high = 0
-                            ]
+            activation = ContainedBy And [
+                    Object id = Source.PlanetID
+                    Population high = 0
+                ]
```
geoffthemedio: weird indenting.  line up the `]` with `a` of `activation` and indent the contained lines by 4 spaces

default/scripting/buildings/CULTURE_ARCHIVES.focs.txt
```diff
@@ -17,8 +17,11 @@ BuildingType
             ]
             priority = [[VERY_LATE_PRIORITY]]
             effects = [
-                SetTargetResearch value = Value + (NamedReal name = "BLD_CULTURE_ARCHIVES_TARGET_RESEARCH_FLAT" value = 5)
-                SetTargetIndustry value = Value + (NamedReal name = "BLD_CULTURE_ARCHIVES_TARGET_INDUSTRY_PERPOP" value = (Target.Population / 2))
+                SetTargetResearch value = Value
+                    + (NamedReal name = "BLD_CULTURE_ARCHIVES_TARGET_RESEARCH_FLAT" value = 5)
+                SetTargetIndustry value = Value
+                    + (NamedReal name = "BLD_CULTURE_ARCHIVES_TARGET_INDUSTRY_PERPOP"
+                                 value = (Target.Population / 2))
```
geoffthemedio: I think you can't use a `Target` reference in a named valueref like this.

default/scripting/buildings/CONC_CAMP.focs.txt
```diff
@@ -23,7 +23,10 @@ BuildingType
                 OwnedBy empire = Source.Owner
             ]
             priority = [[LATE_PRIORITY]]
-            effects = SetTargetIndustry value = Value + Target.Population * 10 * [[INDUSTRY_PER_POP]]
+            effects = SetTargetIndustry value = Value + Target.Population
```
geoffthemedio, line 26: rather do it like this with the `Target` reference outside the named valueref

default/scripting/buildings/CULTURE_ARCHIVES.focs.txt
Oberlus: Solved (here in several other places).

default/scripting/buildings/CONC_CAMP.focs.txt
Oberlus: Fixed.

Oberlus:
Placing ANCIENT_RUINS_TARGET_RESEARCH_PERPOP first in the named_values file did the trick for that error. Weird.

Rest of runtime errors were caused by bugs in my coding, all fixed now.

default/scripting/buildings/CULTURE_ARCHIVES.focs.txt
agrrr3: if i remember rigt i think it might work, but is not officially supported ;)
that is why there are the runtime error messages

of course the pedia should give some error about this, because source/target does not exist.

agrrr3: also this breaks the current convention - PERPOP is a value you have to multiply with population and not a value which was already multiplied by population so BLD_CULTURE_ARCHIVES_TARGET_INDUSTRY_PERPOP should probably be 0.5

agrrr3:
> Placing ANCIENT_RUINS_TARGET_RESEARCH_PERPOP first in the named_values file did the trick for that error. Weird.
> 
> Rest of runtime errors were caused by bugs in my coding, all fixed now.

so ANCIENT_RUINS_TARGET_RESEARCH_PERPOP was defined further down in the named_values file when this was not working?

agrrr3:
I would have appreciated what the logs said if ANCIENT_RUINS_TARGET_RESEARCH_PERPOP was registered or not and if something was shown for ANCIENT_RUINS_TARGET_RESEARCH_PERPOP in named value pedia entry.

agrrr3:
My wild guess was that at the time this wasnt working you broke the parsing somewhere so definition was not parsed ;)

agrrr3:
> I would have appreciated what the logs said if ANCIENT_RUINS_TARGET_RESEARCH_PERPOP was registered or not and if something was shown for ANCIENT_RUINS_TARGET_RESEARCH_PERPOP in named value pedia entry.

not sure what happens there, could reproduce. when starting the game, the value ref is registered in the real-registry.

agrrr3:
ok, kind of race condition. named_values.focs.txt needs to parsed before. 
As CurrentContent is only supported for named-in-the-middle value refs, I could also simply make this an info log (or skip the log entry) for lookups.

Oberlus:
> > Placing ANCIENT_RUINS_TARGET_RESEARCH_PERPOP first in the named_values file did the trick for that error. Weird.
> > Rest of runtime errors were caused by bugs in my coding, all fixed now.
> 
> so ANCIENT_RUINS_TARGET_RESEARCH_PERPOP was defined further down in the named_values file when this was not working?

Yes, I first put it at the end (before includes), then at the start. Gonna try these and other orders, now that I'm sure there is no broken parsing elsewhere, to see what happens.

Oberlus:
No errors with this:
```
NamedReal name = "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP" value = (5 * [[RESEARCH_PER_POP]])

NamedReal name = "IMPERIAL_GARRISON_MAX_TROOPS_FLAT" value = 6

NamedReal name = "SHP_REINFORCED_HULL_BONUS" value = (5 * [[SHIP_STRUCTURE_FACTOR]]) 

NamedInteger name = "NUM_COMBAT_ROUNDS" value = (GameRule name = "RULE_NUM_COMBAT_ROUNDS")

NamedInteger name = "NUM_COMBAT_ROUNDS_FIGHTERS" value = (GameRule name = "RULE_NUM_COMBAT_ROUNDS") - 1
```

With this (put before first integer value)
```
NamedReal name = "IMPERIAL_GARRISON_MAX_TROOPS_FLAT" value = 6

NamedReal name = "SHP_REINFORCED_HULL_BONUS" value = (5 * [[SHIP_STRUCTURE_FACTOR]]) 

NamedReal name = "ANCIENT_RUINS_TARGET_RESEARCH_PERPOP" value = (5 * [[RESEARCH_PER_POP]])

NamedInteger name = "NUM_COMBAT_ROUNDS" value = (GameRule name = "RULE_NUM_COMBAT_ROUNDS")

NamedInteger name = "NUM_COMBAT_ROUNDS_FIGHTERS" value = (GameRule name = "RULE_NUM_COMBAT_ROUNDS") - 1
```
I get several errors: client complains can't found register value `ANCIENT_RUINS_TARGET_RESEARCH_PERPOP`; server complains on some variant within named values that are actually invariant (I already removed `Target` from the named value it is complaining).

So it has nothing to do with the order real-integer.

Oberlus:
Good to merge. The above discussion is about debugging of backend code for Named Values. This PR is only about FOCS changes, tested without runtime errors in its current form.

geoffthemedio:
So are you planning to make basically every value in every effect a named ref? Seems excessive and probably a substantial performance hit...Are they all necessary useful to reference from various other places?

Oberlus:
The idea is to turn into named values anything referenced from pedia/stringtables that might be subject to rebalance (either locally as in personal modding or when working on permanent game changes).
If it is going to hit on performance... I could finish all the work and make tests before merging.

Oberlus:
@geoffthemedio How is this going to cause any performance hit?
NamedValues won't change after a game has started, they would require calculation only before turn 1 (or after game load/start). Am I wrong? Or there are other sources of overload?

Oberlus:
Smoketest failed in the Travis-CI. And the log dump is too long (Travis stops it).
It works well in my local, AIs were killing each other nicely. I got repeated instances of this kind of error though:
```
12:22:19.557388 {0x00007f503200c740} [error] server : Pathfinder.cpp:831 : GeneralizedLocation of Notus (2356) is between -1 and -1
```

Oberlus:
I've completed three test games (multiplayer in localhost, moderator plus three random species AI) with 150 systems, random seeds. All three without errors apart from the following, which seems unrelated to this PR:

In freeoriond.log
```
15:54:40.117026 {0x00007f35d48a8740} [error] server : ServerApp.cpp:1605 : ServerApp::GetPlayerInfoMap: couldn't find an empire for player with id 1
15:54:40.117035 {0x00007f35d48a8740} [error] server : ServerApp.cpp:1611 : ServerApp::GetPlayerInfoMap found player connection with unsupported client type.
```

In freeorion.log

At start:
`15:54:40.164157 {0x00007f4bd557cf80} [error] client : DesignWnd.cpp:1118 : Unable to initialize ShipDesignManager because empire id, -1, is invalid`

Every turn:
```
15:56:09.220032 {0x00007f4bd557cf80} [debug] client : SitRepPanel.cpp:768 : SitRepPanel::Update()
15:56:09.220409 {0x00007f4bd557cf80} [error] client : VarText.cpp:219 : Substitute::operator(): No substitution executed for tag: ship from token: %ship%
```

This would be good to merge, unless @geoffthemedio thinks it's better to have a full prototype with all the named values in place to compare performance.
Also [this comment](https://github.com/freeorion/freeorion/pull/3216#issuecomment-722308221) needs answer.

geoffthemedio:
> @geoffthemedio How is this going to cause any performance hit?
> NamedValues won't change after a game has started, they would require calculation only before turn 1 (or after game load/start). Am I wrong? Or there are other sources of overload?

I was thinking of expressions like `effects = SetTargetIndustry value = Value + 15 * [[INDUSTRY_PER_POP]]` which have a Value-reference +/- something target-invariant, and are (supposed to be) optimized, but I suppose the named-lookup equivalent would still be target-invariant.

In general though, named references require an extra lookup to evaluate each time, rather than having direct pointers to the rest of the expression tree. Whether it's actually substantially slower can be measured with the effect execution timing in the logs...

Oberlus:
> Whether it's actually substantially slower can be measured with the effect execution timing in the logs...

I don't think I'm doing this right.
The best way I found for now is as follows:
1. Create a testing branch from master and remove every instance of NamedInteger/NamedReal/*Lookup.
2. Test master, testing and release-0.4.10 branches following this steps:
-- Set log options to [info] except for timer, which is at [trace].
-- Start single player game.
-- Set Monsters, Specials and Natives to [none], and unmark [random reseeding].
-- Begin game, set autoturn and press Turn.
-- Stop game at turn 80 and exit freeorion.
-- Save a backup of the logs.
3. Compare time elapsed between game start and server exit for each branch.

I get 53 (+/- 0.5) seconds for master and testing branches and 45 seconds for release-0.4.10.

So it seems introducing hundreds of named values in the FOCS files is not affecting performance, assuming in both cases the backend code for named values management.

There are many changes between 0.4.10 and master, including more meters and quite a bunch of new effects, so one can't know how much performance loss is due to backend code for named values, but I ran it out of curiosity.

I guess a better way to compare master with and without named values would be removing all commits about named values, to be able to distinguish between the performance loss coming from having the backend code (without actually including named values in any FOCS file).

@geoffthemedio Should I proceed in any way about this or just continue with PRs for the remaining named values to be introduced?

Oberlus:
> Whether it's actually substantially slower can be measured with the effect execution timing in the logs...

I don't think I'm doing this right, but here is what I did:

- Create a branch "testing" out of master and remove every instance of NamedInteger/NamedReal/*Lookup in the FOCS files.

- For each branch release-0.4.10, master and testing, do:
-- Start freeorion, set timer log options to trace, rest to error, and autosave to 2 per turn.
-- Single player game: set all options to none/low, unmark random reseeding.
-- Start game, set autoturn, press turn, stop game at turn 80, exit freeorion.
-- Backup freeorion*.log files.

- Then compare for each branch the time elapsed between game start and server exit. I got:
-- master and testing: 53 seconds.
-- release.0.4.10: 45 seconds.

So it seems there is no apparent performance loss from adding hundreds of named values in the FOCS files, compared to the same backend code (with named reference manager and the such) with no named values in the FOCS files.

The comparison with 0.4.10 branch is mostly out of curiosity, since it doesn't allow to distinguish how much performance loss comes from the backend code specific to named values and how much from the rest of new code for influence (new meters, many more effects...).

I guess I could try and reverse every named-value-related commit from master and compare with that.

@geoffthemedio Should I proceed in any way regarding this performance testing or continue with more PRs for remaining to-be-introduced named values?

agrrr3:
> I guess a better way to compare master with and without named values would be removing all commits about named values, to be able to distinguish between the performance loss coming from having the backend code (without actually including named values in any FOCS file).

There should be no runtime hit for having named values in the backend. This only gets triggered if the value refs are used. Parsing an empty named_values.txt takes probably some milliseconds on startup.

As long as the invariants work I would also not expect a relevant runtime impact for using a hundred named value refs. The time spend extra for lookup of course also depends on how often a named value ref is used, e.g. if used in combattargets condition a value ref might be looked up really often.

Oberlus:
Then I'll continue with the following PR.
Building/tech/policy/species effects changing meters are not referenced many times per turn (and so are safe), are they?

geoffthemedio:
Time taken to auto-play X turns isn't a great metric, as it depends on a lot of other things.

A probably better way to test would be to have a fairly large (hundreds of systems) game on turn 100+ with multiple players / AIs. Run a turn, and then check the logs for lines like
````
22:30:40.068965 {0x000078b0} [debug] timer : scopedtimer.cpp:48 : StoreTargetsAndCausesOfEffectsGroup < 966 >  cause type: ECT_SPECIAL  specific cause: TIDAL_LOCK_SPECIAL  sources: 83  scope: And [    Source    Focus name = "FOCUS_INDUSTRY"] time:     10.3 ms
22:30:40.068965 {0x00004d64} [debug] timer : scopedtimer.cpp:48 : StoreTargetsAndCausesOfEffectsGroup < 959 >  cause type: ECT_SPECIAL  specific cause: RESONANT_MOON_SPECIAL  sources: 67  scope: And [    InSystem id = Source.SystemID    Ship    OwnedBy empire = Source.Owner] time:     10.5 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - species         time:       28 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - specials        time:        1 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - techs           time:        5 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - policies        time:        0 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - buildings       time:       11 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - ship hull/parts time:       10 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - fields          time:        2 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - eval waiting    time:        3 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:259 : Effect TargetSets Evaluation - reordering      time:        9 ms
22:30:40.101973 {0x00006538} [debug] timer : scopedtimer.cpp:269 : Effect TargetSets Evaluation                   time:     74.8 ms
22:30:40.182991 {0x00006538} [debug] timer : scopedtimer.cpp:48 : Universe::ExecuteEffects time:     80.9 ms
22:30:40.188993 {0x00006538} [debug] timer : scopedtimer.cpp:48 : Universe::UpdateMeterEstimatesImpl on 5532 objects time:      164 ms
````
which are indicating how log it took to evaluate scope conditions and to execute effects, and lists the details of any particularly slow condition evaluations.

PR #3218, opened 2020-11-02
Named Values for Influence and follow-ups

Oberlus:
- Named Values for Influence modifiers
- Follow-ups and grooming
- Addition of Research Focus encyclopedia page
- Improvement of descriptions of some policies

Follows up #3216

### DIFF
--- a/default/scripting/buildings/IMPERIAL_PALACE.focs.txt
+++ b/default/scripting/buildings/IMPERIAL_PALACE.focs.txt
@@ -54,12 +54,14 @@ BuildingType
             ]
             effects = [
                 SetEmpireCapital
-                SetTargetInfluence value = Value + 3
+                SetTargetInfluence value = Value
+                     + (NamedReal name = "BLD_IMPERIAL_PALACE_TARGET_INFLUENCE_FLAT" value = 3) 
             ]
 
         EffectsGroup
             scope = Source
-            effects = SetEmpireMeter empire = Source.Owner meter = "ECONOMIC_CATEGORY_NUM_POLICY_SLOTS" value = Value + 1
+            effects = SetEmpireMeter empire = Source.Owner
+                                     meter = "ECONOMIC_CATEGORY_NUM_POLICY_SLOTS" value = Value + 1
 
         EffectsGroup
             scope = And [
@@ -68,8 +70,8 @@ BuildingType
                 OwnedBy empire = Source.ProducedByEmpireID
             ]
             stackinggroup = "IMPERIAL_PALACE_SUPPLY_EFFECT"
-            effects =
-                SetMaxSupply value = Value + 2
+            effects = SetMaxSupply value = Value
+                        + (NamedReal name = "BLD_IMPERIAL_PALACE_MAX_SUPPLY_FLAT" value = 2)
 
         EffectsGroup
             scope = And [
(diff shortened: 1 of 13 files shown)


Oberlus:
Ehmm... This includes the commits from #3216. I'll rebase after it has been merged.

Oberlus:
The smoketests failed because the job exceeded the maximum log length and was terminated.
The logs are huge due to lots of named value debugging info.

o01eg:
@Oberlus Successful tests don't produce logs, I suppose there was something failing.

Oberlus:
> Successful tests don't produce logs, I suppose there was something failing.

Smoketests run in my local finished with no errors. Dunno what's going on.

o01eg:
I suppose it isn't issue with PR but with master. I've launched tests many times and sometimes I get 
```
[warn] ai : ClientApp.cpp:173 : Checksum for NamedValueRefManager on server 215410 != client 2208271

[warn] ai : ClientApp.cpp:173 : Checksum for NamedValueRefManager on server 215410 != client 6123322

[warn] test : ClientApp.cpp:173 : Checksum for NamedValueRefManager on server 2208271 != client 215410
```

agrrr3:
> I suppose it isn't issue with PR but with master. I've launched tests many times and sometimes I get
> 
> ```
> [warn] ai : ClientApp.cpp:173 : Checksum for NamedValueRefManager on server 215410 != client 2208271
> 
> [warn] ai : ClientApp.cpp:173 : Checksum for NamedValueRefManager on server 215410 != client 6123322
> 
> [warn] test : ClientApp.cpp:173 : Checksum for NamedValueRefManager on server 2208271 != client 215410
> ```

do i understand correctly that that means that there were differences in the contents of the three registries?

Oberlus:
@agrrr3 I can reproduce starting multiplayer game in localhost.

o01eg:
> do i understand correctly that that means that there were differences in the contents of the three registries?

@agrrr3 No, for the same content manager produce different checksums.

agrrr3:
> > do i understand correctly that that means that there were differences in the contents of the three registries?
> 
> @agrrr3 No, for the same content manager produce different checksums.

my guess is that the invariants are part the checksums. but these are subject to race conditions without a bigger refactoring.
if that is so a reasonable workaround would be to take the invariants out of the checksums as those are derived anyway.
the clean solution would be to initialise the invariants after parsing is done. another solution would be to initialise invariants on first use - introducing a "m_invariants_initialized" flag into all valuerefs

agrrr3:
Actually the invariants are not part of the checksum. But m_constant_expr in Operation is and that is derived like the invariants. So I will remove that from the checksums for consistency.

geoffthemedio:
> another solution would be to initialise invariants on first use - introducing a "m_invariants_initialized" flag into all valuerefs

not a great idea as it could a probably inlinable / simple function into a test / branch...

Oberlus:
There is a problem with stringtables: the empty strings I removed (the ones ending with _FLAT) are not being substituted by the calculated values. It seems the fix for that didn't work or I'm confused and it wasn't merged. @agrrr3 @geoffthemedio should I add the empty strings back?

geoffthemedio:
Just fix it properly... No need for a temporary workaround.

Oberlus:
> Just fix it properly... No need for a temporary workaround.

Agree. I'm sorry I can't help with that.
@agrrr3 

Edit: meanwhile, this is good to merge. The named values not working will be fixed with someone else's PR addressing the empty strings issue in the backend.

PR #3230, opened 2020-11-11
Fix display of named value refs without string table entry

agrrr3:
Stringtable entries for named value refs are optional (info gets logged instead of reporting errors).

But display did not work as `[[value bla]]` needs to be converted to `<value bla></value>` in StringTables lookup, 
so the LinkText can do formatting and tooltip. With this PR it converts correctly.

@oberlus reported this bug but i dont see an github Issue for this

### DIFF
--- a/util/StringTable.cpp
+++ b/util/StringTable.cpp
@@ -245,11 +245,15 @@ void StringTable::Load(std::shared_ptr<const StringTable> fallback) {
                     entry.second.replace(position, match.length(), substitution);
                     position += substitution.length();
                 } else {
-                    if (match[1] == "value")
+                    if (match[1] == "value") {
                         InfoLogger() << "Unresolved optional value reference: " << match[2] << " in: " << m_filename << ".";
-                    else
+                        const std::string substitution = "<value " + match[2].str() + "></value>";
+                        entry.second.replace(position, match.length(), substitution);
+                        position += substitution.length();
+                    } else {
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I can't really tell if this is working or what it's supposed to look like...
Nanite per

Oberlus:
f69f667 in Ubuntu 18.04, PR #3230 on top of master (fa112c8)

PR3230_ubuntu

Decimals are still ugly, but that's a different issue.

Someone should tackle the Windows rendering issue...

agrrr3:
> I can't really tell if this is working or what it's supposed to look like...

sigh. it should read "Structure by 40 per turn" and "Industry by 1 per". 

Thanks oberlus for the screenshot.

@geoffthemedio Could you please screenshot the named value ref testpage in the pedia showing the values for the FLD_NANITE_... entries (its listed alphabetically)?

geoffthemedio:
Nanite Named Refs

agrrr3:
Thanks for the screenshots geoff. Valuerefs are registered - probably correctly as the invariants are right. And also fetching the invariants works. So only the UI is botched. 

This seems to be geoffs known weird ms windows issue. Will open an issue for that and work on it after getting the concurrent parse right to fix the remaining checksum error.

Oberlus confirmed that the UI fix works on his linux. So good to merge?

PR #3234, opened 2020-11-13
Self-sustaining metabolism nerf - population boost only on good planets

agrrr3:
Nerfs early colonisation of poor planets strongly.

Need terraforming to make the effect best out of the effect.

https://www.freeorion.org/forum/viewtopic.php?f=15&t=11844&p=104297


agrrr3:
@geoffthemedio  can we merge this? oberlus trith nerf went into master - i want to check out trith balance/power level with the next test build.

PR #3205, opened 2020-10-25
Build GDNative client library

o01eg:
Additional option to build GDNative library to use in Godot.

Adds option `BUILD_CLIENT_GODOT` (default **OFF**) and `GODOT_CPP_ROOT` to point at `godot-cpp` repository.

Exposes property `optionsDB` with methods to access FreeOrion's OptionsDB.

Exposes property `networking` with methods to connect to the server. Incoming messages are emits `ping` signal.

Exposes some incoming network messages as signals on main node include `game_start` message.

To test add 

```
p
GodotPlayer:p:1
otherPlayer2:p:1
```
into `auth.txt` config file and start the server with `./freeoriond --hostless -q --log-file /proc/self/fd/1 --setup.ai.player.count 0 --network.server.conn-human-empire-players.min 0 --network.server.unconn-human-empire-players.max 0`. This will start the game on the server and the only thing client need to do is to connect to the server.

To test single player game place client binary into the same folder as freeoriond binary.

C++ sources of it got from https://github.com/Vezzra/fo-godot-prototype

Forum thread: https://freeorion.org/forum/viewtopic.php?f=9&t=11744

PR #3235, opened 2020-11-17
Build GDNative client library

o01eg:
_Edit by @Vezzra: supersedes #3205 /edit_

Additional option to build GDNative library to use in Godot.

Adds option `BUILD_CLIENT_GODOT` (default **OFF**) and `GODOT_CPP_ROOT` to point at `godot-cpp` repository.

Exposes property `optionsDB` with methods to access FreeOrion's OptionsDB.

Exposes property `networking` with methods to connect to the server. Incoming messages are emits `ping` signal.

Exposes some incoming network messages as signals on main node include `game_start` message.

To test add 

```
p
GodotPlayer:p:1
otherPlayer2:p:1
```
into `auth.txt` config file and start the server with `./freeoriond --hostless -q --log-file /proc/self/fd/1 --setup.ai.player.count 0 --network.server.conn-human-empire-players.min 0 --network.server.unconn-human-empire-players.max 0`. This will start the game on the server and the only thing client need to do is to connect to the server.

To test single player game place godot binary into the same folder as freeoriond binary.

C++ sources of it got from https://github.com/Vezzra/fo-godot-prototype

Forum thread: https://freeorion.org/forum/viewtopic.php?f=9&t=11744


Vezzra:
@o01eg, thanks, perfect. I'll now try to integrate the Godot client prototype app into this branch.

o01eg:
@Vezzra Godot-cpp was included here as is. Is it intended?

Vezzra:
Yes.

Vezzra:
Ok, done. I've copied the entire Godot client prototype app (the extended version by @o01eg, see [here](https://github.com/o01eg/fo-godot-prototype/tree/use-freeorion-library)) into a new "client" subfolder within the "godot" folder.

I also made a few adjustments, so developing the Godot client app within the Godot IDE works a bit smoother:

- I've copied the `godot-cpp` dependency into the "godot" folder
- I've adjusted the cmake build scripts accordingly, so that the `godot-cpp` dependency is found by default
- I've adjusted the cmake build scripts so that when building the library for the Godot client, all the required binaries get copied to the "godot/client/bin/x11" folder, where the Godot client app expects them
- I've changed the default value for the resource directory to "../../default", to avoid having to place a link to the "default" folder in "godot/client" (don't know how git would handle that)

These are of course more or less crude hacks, but for prototype development these should make life a bit easier. Otherwise everyone would have to provide the `godot-cpp` dependency themselves and point cmake to it, copy the required libraries for the GDNative scripts to the proper locations every time the C++ code is modified and needs to be rebuild, etc.

With my setup, you basically only need to set the BUILD_GODOT_CLIENT option to ON, build FO, open the Godot project in "godot/client" with the Godot IDE and run it.

There are of course issues, things don't work as expected yet. Apparently parsing fails, there is a error message in the logs about not being able to access the resource directory at "../../default/../../default/......." - why the resource path gets duplicated here I don't know. This happend during background parsing of the content scripts.

@o01eg, take a look at what I've done, I hope I got everything reasonably right. Feel free to suggest corrections to my tinkerings.

godot/gdfreeorion.cpp
```diff
@@ -0,0 +1,237 @@
...
+            !boost::filesystem::exists(GetResourceDir() / "data" / "art" / "misc" / "missing.png"))
+        {
+            DebugLogger() << "Default Resources directory missing or does not contain expected files. Cannot start game.";
+            throw std::runtime_error("Unable to load game resources at default location: " +
+                                     PathToString(GetResourceDir()) + " : Install may be broken.");
+        }
+    }
+
+    // initialize any variables here
+    app = std::make_unique<GodotClientApp>();
```
Vezzra, line 234: @o01eg, I've moved that line back here. You were right after all, the resource folder needs to be correctly set first, because it's needed when initializing `GodotClientApp`. When the "default" folder can't be found and needs to be reset (which will happen with my hack to set it to "../../default", that happens to late and all the stuff called by the constructor of `GodotClientApp` fails.

Having no error message output in the Godot client app when the resource dir can't be found is the lesser evil in that case.

Vezzra (approved):
So far so good. Only one comment, see below - I had to move the initialization of `GodotClientApp` back to where you originally placed it.

Otherwise I now can at least build the GDNative library and start the Godot client app. Still issues (see my post above), but it's a start. Now we need to start fixing and improving things... :wink:

Vezzra:
As the purpose of this branch/PR is developing a protoype Godot client app, I've set the BUILD_GODOT_CLIENT build option to ON by default.

o01eg:
> I've adjusted the cmake build scripts accordingly, so that the godot-cpp dependency is found by default

Does freeorion's cmake build godot-cpp?

I supposed if accepted godot-cpp will became part of MacOS and Windows SDKs. On LInux it's simple to build it like it done in CI.

> I've changed the default value for the resource directory to "../../default", to avoid having to place a link to the "default" folder in "godot/client" (don't know how git would handle that)

I think client should be launched from project root directory instead.

godot/gdfreeorion.cpp
o01eg: > I've moved that line back here. You were right after all, the resource folder needs to be correctly set first, because it's needed when initializing GodotClientApp.

Yes, it's a same other clients do.

Vezzra:
> Does freeorion's cmake build godot-cpp?

Not yet, which is also one of the issues - this setup currently only works with Linux, for which the `godot-cpp` libraries have already been prebuilt (and incorporated into the branch). That's of course not ideal.

Ideally, building the `godot-cpp` libs should be invoked from the FO build scripts when necessary (so that all platforms are supported). That's one of the things which need to be fixed, but I don't know how the cmake scripts actually need to be adjusted.

How to do that correctly is the problem...

o01eg:
That why I think it better to haven't `godot-cpp` in FO's sources. It's hardly usable and in case of cross-compilation the cohesion of `scons` and `cmake` looks very tough.

Vezzra:
> I supposed if accepted godot-cpp will became part of MacOS and Windows SDKs.

Most likely, yes.

> I think client should be launched from project root directory instead.

Difficult to do from within the IDE (as currently the Godot project file is within "godot/client"). Development of a Godot app is best done with the Godot IDE, having to export for each test run and then run the app from the source root is cumbersome.

How to do that correctly is another issue which needs to be addressed. I guess the only proper way to do this is by placing the Godot project file in the source root. But if we do that, a lot of paths in the current app needs to be adjusted.

Vezzra:
> That why I think it better to haven't godot-cpp in FO's sources. It's hardly usable and in case of cross-compilation the cohesion of scons and cmake looks very tough.

Probably - we'll whatever approach is best in the end, if we really adopt Godot. But for prototyping I wanted to make things as painless as possible, and not require our devs to have to care for the dependency manually. I guess the best for now will be to prebuild the godot cpp libs for all platforms and add them to the branch.

Vezzra:
@o01eg, I've finally managed to get `godot-cpp` to be built from FO's cmake scripts (I found a very useful cmake feature called "ExternalProject" that fits the purpose perfectly). I've put up a PR so you can review it: #3236.

This should accomplish what you've asked for:

> Does freeorion's cmake build godot-cpp?

With those changes that should now work correctly. On Linux it does, on Windows I haven't tested yet, and on macOS it doesn't yet (but then, on macOS everything is much more complicated anyway when it comes to building FO...).

In that PR I've also cleaned up the `godot-cpp` folder structure and `.gitignore`, so that the build artefacts that get placed in the source tree don't get into version control (which is pretty messy). That problem should be solved too now.

Please take a look at what I've cobbled together and let me know what you think.

o01eg:
@Vezzra Is `godot-cpp` supposed to be rebuilded each time?

o01eg:
Yet another disadvantage in having own copy of sources is missing version information which could be required to fill issues for `godot-cpp`.

o01eg:
@Vezzra Which version of `godot-cpp` we use? It looks like `master` should be used for 3.2.3 Godot: https://github.com/godotengine/godot-cpp/pull/476#issuecomment-734479675

Vezzra:
@o01eg,

> Is godot-cpp supposed to be rebuilded each time?

No - unless you clean out the FO build directory. In that case `godot-cpp` gets rebuilt.

> Yet another disadvantage in having own copy of sources is missing version information which could be required to fill issues for `godot-cpp`.

For prototyping it seemed good enough for me to just incorporate the version of the `godot-cpp` sources that corresponds to the stable version of the Godot engine we use (which was 3.2. until recently, now 3.2.3 is out).

> Which version of godot-cpp we use? It looks like master should be used for 3.2.3 Godot

I pulled the 3.2 branch, assuming that this branch would contain the version of `godot-cpp` which corresponds to the current stable 3.2.x releases.

Apparently that's not the case. And judging by the comments you linked to, there is neither a proper, transparent structure to the `godot-cpp` repo that makes it clear which branch/tag to use, nor a proper documentation which could help you find out that actually important info.

What do you suggest? What I want to avoid is for devs to have to manually maintain their own copy of `godot-cpp` somwhere independently of the FO source tree. When you build FO, building the GDNative FO client library and all required dependencies should happen automatically, without devs having to manually tweak things.

We can of course, instead of maintaining a copy of the `godot-cpp` sources within the FO source tree, clone `godot-cpp` from their repo, and build and maintain it entirely within the FO build directory. FO cmake scripts need to be adjusted accordingly of course.

As `godot-cpp` master apparently is supposed to be used for the current stable 3.2.x releases of Godot, we can only hope that `godot-cpp` master is always in a stable, never in a broken state. I have to admit, I don't feel very comfortable with that approach. If only they'd at least maintained proper version tags to indicate stable releases of `godot-cpp`...

Well, maybe `godot-cpp` master is supposed to be stable all the time, and they only merge things into master that don't break anything. Again, some documentation or info anywhere to give some indication about that would be nice...

Well, what would you prefer? @geoffthemedio, what's your opinion on this?

o01eg:
> What do you suggest? What I want to avoid is for devs to have to manually maintain their own copy of godot-cpp somwhere independently of the FO source tree. When you build FO, building the GDNative FO client library and all required dependencies should happen automatically, without devs having to manually tweak things.

@Vezzra Could we use sub-repository for it? At least it looks simpler to manage compared to sources copy.

> Well, what would you prefer? @geoffthemedio, what's your opinion on this?

Yes, if we end up with merging it into `master` branch how we should deal with godot-cpp dependency? At least for Windows and MacOS it could be built in SDK.

agrrr3:
> > What do you suggest? What I want to avoid is for devs to have to manually maintain their own copy of godot-cpp somwhere independently of the FO source tree. When you build FO, building the GDNative FO client library and all required dependencies should happen automatically, without devs having to manually tweak things.
> 
> @Vezzra Could we use sub-repository for it? At least it looks simpler to manage compared to sources copy.

I have some experience with git submodules. where i was using it we had two repos which were corelated/developed in parallel. That was a bad idea there, should have gone for a monolithic repo instead.

We loosely follow a dependency (and rarely change it), so **here a submodule makes a lot of sense**. Definitly a lot more than copying the source into our repo. 

Main benefits: 
* you do not have to download the godot source if you do not need it
* godot code and code changes does not bloat our git
* a submodule is basically a link to a revision in another repo - with source you either have no clue which revision you are using or you have to track it explicitly somewhere else (which is prone to error)

geoffthemedio:
Is godot-cpp something that we can rely on being available on most Linux systems with reasonably consistent versions? Probably not, given how variable and usually out of date the versions of Boost are... Seems awkward to have a different dependency mechanism for just this code, though. But keeping a manually maintained copy within the FreeOrion git repository is also awkward. So if the "submodule" mechanism works without much complication to pull a specific version of godot-cpp when checking out freeorion, I suppose it's fine.

Vezzra:
> Could we use sub-repository for it? At least it looks simpler to manage compared to sources copy.

It is, or better, it should be, which is why, when I fiddled around with my prototype, I took that road (it's also what's recommended in the Godot documentation).

However, either I did something wrong, or they have something messed up in how they have set up their repo (or, more precisely, their `.gitignore` files), because when I built `godot-cpp` and my test GDNative library, I ended up with some changes to files in the `godot-cpp` source tree which git insisted I need to commit (which I can't and won't).

I could not get the setup to work smoothly and ended up ripping the git submodule out, and using a copy of `godot-cpp`.

Aside from that, for a protoype branch I did not want to include a git submodule into the FO main repo. Pulling and building `godot-cpp` via CMake's feature `ExternalProject` looks like the better approach to me at this point, if we want to avoid maintaining a copy of `godot-cpp` in FO's source tree.

But I defer to @geoffthemedio for the final call on this.

Vezzra:
> Is godot-cpp something that we can rely on being available on most Linux systems with reasonably consistent versions?

No, I don't think this is maintained in the repos of the various Linux distros around at all. It's an external dependency on Linux we need to pull in one way or another. AFAICT from what I know currently, the entire Godot dependency will be an issue on Linux if we decide to switch to Godot, because we definitely will have to handle it very differently than all the other dependencies (which are provided by the distro repos).

Vezzra:
> Yes, if we end up with merging it into master branch how we should deal with `godot-cpp` dependency? At least for Windows and MacOS it could be built in SDK.

Good question, see above. On Mac and Windows you're right, the only sensible way to handle this is incorporating it into the SDK. But on Linux things will most likely become more difficult than they are now.

o01eg:
> Is godot-cpp something that we can rely on being available on most Linux systems with reasonably consistent versions?

@geoffthemedio Even Godot itself rarely get into Linux repositories.

Calinou:
> Even Godot itself rarely get into Linux repositories.

Indeed, I wouldn't rely on up-to-date Godot (or worse, godot-cpp) being readily available in Linux distribution repositories.

o01eg:
As discussed on online chat next step is to try to package it at least with snap and debian packages.

o01eg:
@Vezzra Could we somehow move Godot project directory so `default/` folder will be visible and could be exported?

Vezzra:
@o01eg, I got a notification about a comment you posted:

> Could we somehow move Godot project directory so default/ folder will be visible and could be exported?

However, I can't see it here. I'm going to respond regardless: I agree, but I have no idea what's the best way to do this. Of course, we could make the FO root folder the project root for the FO Godot client, but that would mean to place the Godot project file in the root folder, which is in conflict with the idea to contain all the Godot client related stuff in it's own dedicated subfolder.

If we put all Godot client related stuff in it's own dedicated subfolder, I don't know how to tell Godot to use assets which are located outside of a Godot project's root folder. Is there a way to maintain a symbolic link within a git repo that is handled correctly by git across different platforms?

@geoffthemedio, any ideas?

o01eg:
@Vezzra I solved it in #3367 by creating default directory in a godot project and making links to `scripting` and `stringtables`.

Vezzra:
Ok

o01eg:
Closed in favor #3572

PR #3237, opened 2020-11-21
Some Godot fixes

o01eg:
- Removes unneeded CI check for separate `godot-cpp` library.
- Uses subfolders for Godot architectures.
- Fixes default resources search.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -639,11 +639,17 @@ if(BUILD_CLIENT_GG)
 endif()
 
 if(BUILD_CLIENT_GODOT)
+    if(CMAKE_BUILD_TYPE MATCHES Debug)
+        set(GODOT_CPP_BUILD_TYPE Debug)
+    else()
+        set(GODOT_CPP_BUILD_TYPE Release)
+    endif()
+
     include(ExternalProject)
     ExternalProject_Add(godotcpp
         SOURCE_DIR ${PROJECT_SOURCE_DIR}/godot/godot-cpp
         BINARY_DIR ${PROJECT_SOURCE_DIR}/godot/godot-cpp/build
-        CMAKE_ARGS -D CMAKE_BUILD_TYPE=$<IF:$<CONFIG:Debug>,Debug,Release>
+        CMAKE_ARGS -D CMAKE_BUILD_TYPE=${GODOT_CPP_BUILD_TYPE}
         INSTALL_COMMAND ""
     )
     
@@ -675,7 +681,7 @@ if(BUILD_CLIENT_GODOT)
     endif(CMAKE_SIZEOF_VOID_P EQUAL 8)
 
     string(TOLOWER ${CMAKE_SYSTEM_NAME} SYSTEM_NAME)
-    string(TOLOWER ${CMAKE_BUILD_TYPE} BUILD_TYPE)
+    string(TOLOWER ${GODOT_CPP_BUILD_TYPE} BUILD_TYPE)
     if(CMAKE_VERSION VERSION_GREATER "3.13")
         target_link_directories(freeoriongodot
             PRIVATE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -759,13 +759,13 @@ endif()
 if(BUILD_CLIENT_GODOT)
     add_custom_command(TARGET freeoriongodot POST_BUILD
         COMMAND
-           ${CMAKE_COMMAND} -E make_directory ${CMAKE_SOURCE_DIR}/godot/client/bin/
+           ${CMAKE_COMMAND} -E make_directory ${CMAKE_SOURCE_DIR}/godot/client/bin/${SYSTEM_NAME}.${BITS}
         COMMAND
-           ${CMAKE_COMMAND} -E copy_if_different "$<TARGET_FILE:freeorioncommon>" "${CMAKE_SOURCE_DIR}/godot/client/bin/"
+           ${CMAKE_COMMAND} -E copy_if_different "$<TARGET_FILE:freeorioncommon>" "${CMAKE_SOURCE_DIR}/godot/client/bin/${SYSTEM_NAME}.${BITS}"
         COMMAND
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -666,18 +666,27 @@ if(BUILD_CLIENT_GODOT)
...
+    # Synchronized with godot-cpp's CMakeLists.txt
+
+    set(BITS 32)
+    if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+        set(BITS 64)
+    endif(CMAKE_SIZEOF_VOID_P EQUAL 8)
+
+    string(TOLOWER ${CMAKE_SYSTEM_NAME} SYSTEM_NAME)
+    string(TOLOWER ${CMAKE_BUILD_TYPE} BUILD_TYPE)
     if(CMAKE_VERSION VERSION_GREATER "3.13")
```
Vezzra, line 685: Generating `BUILD_TYPE` like this just from `CMAKE_BUILD_TYPE` does not work in all cases, as FO has more valid build types than `godot-cpp`. `godot-cpp` only knows/allows `Debug` and `Release`, whereas FO also knows/allows e.g. `RelWithDebInfo` (which actually is the default, at least when I run the build from KDevelop on my Linux system without changing anything).

Meaning, if you build FO e.g. with `RelWithDebInfo`, the `freeoriongodot` target fails to build (because it tries to link with `libgodot-cpp.linux.relwithdebinfo.64.a`, the `godot-cpp` build scripts generated a `libgodot-cpp.linux.release.64.a` static lib though).

So we either revert to the old approach (just using the generator expressions), or, if you want to stick with copying what the `godot-cpp` cmake script does, I recommend the following:

Use a variable like `GODOT_CPP_BUILD_TYPE` to store the build type `godot-cpp` should use depending on the build type used in FO's cmake script, like so:
```
    if(CMAKE_BUILD_TYPE MATCHES Debug)
        set(GODOT_CPP_BUILD_TYPE Debug)
    else()
        set(GODOT_CPP_BUILD_TYPE Release)
    endif()
```
And refer to that in the `ExternalProject_Add` call:
```
    ExternalProject_Add(godotcpp
        SOURCE_DIR ${PROJECT_SOURCE_DIR}/godot/godot-cpp
        BINARY_DIR ${PROJECT_SOURCE_DIR}/godot/godot-cpp/build
        CMAKE_ARGS -D CMAKE_BUILD_TYPE=${GODOT_CPP_BUILD_TYPE}
        INSTALL_COMMAND ""
    )
```
As well as further down, when generating the `godot-cpp` library name:
```
    string(TOLOWER ${GODOT_CPP_BUILD_TYPE} BUILD_TYPE)
```
The entire section would look like this:
```
if(BUILD_CLIENT_GODOT)
    if(CMAKE_BUILD_TYPE MATCHES Debug)
        set(GODOT_CPP_BUILD_TYPE Debug)
    else()
        set(GODOT_CPP_BUILD_TYPE Release)
    endif()
    
    include(ExternalProject)
    ExternalProject_Add(godotcpp
        SOURCE_DIR ${PROJECT_SOURCE_DIR}/godot/godot-cpp
        BINARY_DIR ${PROJECT_SOURCE_DIR}/godot/godot-cpp/build
        CMAKE_ARGS -D CMAKE_BUILD_TYPE=${GODOT_CPP_BUILD_TYPE}
        INSTALL_COMMAND ""
    )
    
    add_library(freeoriongodot SHARED "")

    set_target_properties(freeoriongodot PROPERTIES CXX_VISIBILITY_PRESET default)

    target_compile_definitions(freeoriongodot
        PRIVATE
            -DFREEORION_BUILD_HUMAN_GODOT
    )

    target_include_directories(freeoriongodot SYSTEM
        PRIVATE
            ${GODOT_CPP_ROOT}/include
            ${GODOT_CPP_ROOT}/include/core
            ${GODOT_CPP_ROOT}/include/gen
            ${GODOT_CPP_ROOT}/godot_headers
    )
    
    add_dependencies(freeoriongodot godotcpp)

    # Create the correct name (godot.os.build_type.system_bits)
    # Synchronized with godot-cpp's CMakeLists.txt

    set(BITS 32)
    if(CMAKE_SIZEOF_VOID_P EQUAL 8)
        set(BITS 64)
    endif(CMAKE_SIZEOF_VOID_P EQUAL 8)

    string(TOLOWER ${CMAKE_SYSTEM_NAME} SYSTEM_NAME)
    string(TOLOWER ${GODOT_CPP_BUILD_TYPE} BUILD_TYPE)
    if(CMAKE_VERSION VERSION_GREATER "3.13")
        target_link_directories(freeoriongodot
            PRIVATE
                ${GODOT_CPP_ROOT}/bin/
        )
        target_link_libraries(freeoriongodot
            godot-cpp.${SYSTEM_NAME}.${BUILD_TYPE}.${BITS}
        )
    else()
        target_link_libraries(freeoriongodot
            ${GODOT_CPP_ROOT}/bin/libgodot-cpp.${SYSTEM_NAME}.${BUILD_TYPE}.${BITS}.a
        )
    endif()

    target_link_libraries(freeoriongodot
        freeorioncommon
        freeorionparse
    )

    target_dependencies_copy_to_build(freeoriongodot)
endif()
```

Vezzra (changes requested):
Aside from one issue with generating the `godot-cpp` library name (see my comment below), looks fine to me and works.

So, after applying the suggested fix (or reverting to the former approach to generate the `godot-cpp` library name, whatever you deem better), this should be good to merge.

CMakeLists.txt
o01eg: Used last approach.

Vezzra (approved):
Looks good. As I've already tested the fix before exactly as you have applied it, I'd say this is good to merge.

o01eg:
Should we rebase entire `build-godot` branch to fix MacOS CI issue?

geoffthemedio:
> Should we rebase entire `build-godot` branch to fix MacOS CI issue?

sounds like a good idea

PR #3239, opened 2020-11-21
Fix slow MacOS homebrew update

o01eg:


geoffthemedio:
37 min... seems better...

Vezzra:
I wouldn't have been able to provide any useful input anyway, so fine with me...

PR #3240, opened 2020-11-24
Fix named value display on WIN32 (initialisation issue)

agrrr3:
It seems VarText::FOCS_VALUE_TAG was not initialised on WIN32 when the match happened,
so it tried to match on `</>` instead of `</value>`.

### DIFF
--- a/UI/LinkText.cpp
+++ b/UI/LinkText.cpp
@@ -95,23 +95,21 @@ namespace {
         return retval;
     }
 
-    const std::string FOCS_VALUE_TAG_OPEN_PRE("<" + VarText::FOCS_VALUE_TAG);
-    const std::string FOCS_VALUE_TAG_CLOSE("</" + VarText::FOCS_VALUE_TAG + ">");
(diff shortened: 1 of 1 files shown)


geoffthemedio:
works now, though I feel like showing a tooltip with the expanded value derivation would be better than replacing the text when mousing over, as it wouldn't disrupt the layout of the rest of the text

also, I see numbers displayed like "40.000000" which is obviously awkward in most cases.

geoffthemedio:
mouseover

looks a bit silly...

geoffthemedio:
I also have issues with tooltips...
NameValueRefs_in_ToolTips

Vezzra:
@geoffthemedio, shouldn't we open an issue for the tooltip problems? Here in the comment section of a merged PR it's likely to be forgotten...

geoffthemedio:
> @geoffthemedio, shouldn't we open an issue for the tooltip problems? Here in the comment section of a merged PR it's likely to be forgotten...

Sure

Vezzra:
@geoffthemedio, have you already opened an issue for that? If not, can you do it? I can't reproduce the problem on my Mac.

Vezzra:
>  I can't reproduce the problem on my Mac.

Oops, correction, I can. As explained [here](https://github.com/freeorion/freeorion/issues/3268#issuecomment-752970073), I got a little confused... 😊

PR #3241, opened 2020-11-25
Fix empire readiness when unset auto turn

### LINKED ISSUE #3246: Turn advance is broken - freeze
Bug Report
==========
Turn advance is broken - I checked single player with 6 AI and single player with zero AI.

Environment
-----------
current master build locally, and weekly build snap on fedora 32/33

Description
-----------

Turn advance on turn 1 freezes.

Expected Result
---------------

Turn two is ready for orders.

Steps to reproduce
------------------
* Start a single player game (with zero or more AI)
* Click turn advance

o01eg:


agrrr3:
Ah great catch.
I find that SetAutoTurn calls SetReady very surprising. How about one of the following?

- call it SetAutoTurnAndUpdateReady
- do not call SetReady inside of the SetReady function but do that explicitly on each use site instead
- do not call SetReady inside of the SetReady function but call AutoTurnSetReady afterwards on each use site instead

o01eg:
Second looks better. `AutoTurnSetReady` should be called only on new turn.

agrrr3:
looks good to me

agrrr3:
I also compiled and works on fedora32

Issue #3204, opened 2020-10-25
Modding support


o01eg:
Feature request
===============

Idea
----

Add some layered file system to access game resources based on enabled mods.
For list of enabled mods we can use option like `mod1,mod2,mod3,...`. Then when we ask for file `default/dir1/file1` we check files in order `default/dir1/file1`, `~/.local/share/freeorion/mods/mod1/dir1/file1`, `~/.local/share/freeorion/mods/mod2/dir1/file1` so on and choose latest exists. If latest is empty then file considered empty.
Directories listing also could use same scan.

Gain
----

Simplifies to mod the game, test new features and balance.

Discussion
----------

https://freeorion.org/forum/viewtopic.php?f=9&t=11790

Cjkjvfnby:
> test new features and balance

It is already here.

I use intalled Freeorion for development, so I just point the game to my repository folder i:/projects/freeorion via persistent config.  (I use windows and don't have any compiler)

PPS. As I know all this options can be defined via command-line.  Output  was changed since I used it last time, I don't understand how to use it.

As quick solution you can create a wrapper that will scan folders and generate a commandline.  (I have never tried it but it can do a UI https://github.com/peterbrittain/asciimatics)



persistent_config.xml
```
<?xml version="1.0"?>
<XMLDoc>
  <resource>
    <path>i:/projects/freeorion/default</path>
    <stringtable>
      <path>i:/projects/freeorion/default/stringtables/en.txt</path>
    </stringtable>
  </resource>
  <save>
    <path>i:/games/FreeOrion/save</path>
    <server>
      <path>i:/games/FreeOrion/save</path>
    </server>
  </save>
  </XMLDoc>
```

o01eg:
@Cjkjvfnby There also issue with mod distribution. It's simpler to distribute only changed files than entire `default` folder.

agrrr3:
> @Cjkjvfnby There also issue with mod distribution. It's simpler to distribute only changed files than entire `default` folder.

Especially if one changes the freeorion revision (e.g. when developing or following the weekly builds) an overlay or similar mechanism which keeps the plugin code in its own directories is much more helpful

o01eg:
Yet another reason to have some layered filesystem is to get access to files bundled in Android APK which aren't accessible directly as file.

PR #3242, opened 2020-11-25
Move FOCS file reading to Directory code

o01eg:
Should simplify access to data unavailable via file operations like Android assets #3202 or use it indirection to implement modding support #3204.

### DIFF
--- a/util/Directories.cpp
+++ b/util/Directories.cpp
@@ -734,3 +734,25 @@ auto IsExistingFile(const fs::path& path) -> bool
 
     return false;
 }
+
+auto ReadFile(boost::filesystem::path const& path, std::string& file_contents) -> bool
+{
+    boost::filesystem::ifstream ifs(path);
+    if (!ifs)
+        return false;
+
+    // skip byte order mark (BOM)
+    for (int BOM : {0xEF, 0xBB, 0xBF}) {
+        if (BOM != ifs.get()) {
+            // no header set stream back to start of file
+            ifs.seekg(0, std::ios::beg);
+            // and continue
+            break;
+        }
+    }
+
+    std::getline(ifs, file_contents, '\0');
+
+    // no problems?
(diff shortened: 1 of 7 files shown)


util/StringTable.cpp
```diff
@@ -57,7 +57,7 @@ void StringTable::Load(std::shared_ptr<const StringTable> fallback) {
     auto path = FilenameToPath(m_filename);
     std::string file_contents;
 
-    bool read_success = parse::read_file(path, file_contents);
+    bool read_success = ReadFOCScript(path, file_contents);
```
geoffthemedio: A stringtable isn't FOCS

util/Directories.cpp
```diff
@@ -734,3 +734,25 @@ auto IsExistingFile(const fs::path& path) -> bool
 
     return false;
 }
+
+auto ReadFOCScript(boost::filesystem::path const& path, std::string& file_contents) -> bool
```
geoffthemedio: this isn't FOCS specific... the old name was more accurate

util/StringTable.cpp
o01eg: Aren't it parsed for FOCS substitutions?

util/Directories.cpp
o01eg: OK, changed to `ReadFile`.

o01eg:
I'll remove `ReadFile` from `ShaderProgram`.

util/StringTable.cpp
geoffthemedio: Macro substitutions aren't [FO] content scripts. Even if they were, this function doesn't do them.

o01eg: OK, I removed FOCS mention here. Moreover I suddenly fixed code duplication in `ShaderProgram`.

PR #3245, opened 2020-11-25
Fix segfault when use multiple signal arguments

o01eg:
Fixes `godot_variant` size mismatch on some architectures where `uint64_t` and `void*` have different size like Android armeabi-v7a.

Related to https://github.com/godotengine/godot-cpp/issues/473 in our copy of `godot-cpp` sources.

### DIFF
--- a/godot/godot-cpp/godot_headers/gdnative/variant.h
+++ b/godot/godot-cpp/godot_headers/gdnative/variant.h
@@ -37,7 +37,7 @@ extern "C" {
 
 #include <stdint.h>
 
-#define GODOT_VARIANT_SIZE (16 + sizeof(void *))
+#define GODOT_VARIANT_SIZE (16 + sizeof(int64_t))
 
 #ifndef GODOT_CORE_API_GODOT_VARIANT_TYPE_DEFINED
 #define GODOT_CORE_API_GODOT_VARIANT_TYPE_DEFINED


godot/godot-cpp/godot_headers/gdnative/rid.h
```diff
@@ -37,7 +37,7 @@ extern "C" {
 
 #include <stdint.h>
 
-#define GODOT_RID_SIZE sizeof(void *)
+#define GODOT_RID_SIZE sizeof(int64_t)
```
bruvzg: This change is most likely wrong, RID is `int64_t` only in master. It's pointer in Godot 3.2 https://github.com/godotengine/godot/blob/803f6d2388f64d069810554ab1b0a57d6a5a8087/core/rid.h#L60

o01eg: Thanks, I've re-check `gdnative/rid.h` from Godot Engine 3.2.3 and it definitely have `void *`. Wonder why Godot Engine doesn't use `godot_headers` itself.

Vezzra (approved):
I see. If there are bugs in the `godot-cpp` branch we use that only get fixed in `godot-cpp` master, I guess you are right, we'll have to switch to `godot-cpp` master.

As a quick fix it should be ok to merge this, but the more permanent solution has to be to use the more stable `godot-cpp` branch.

geoffthemedio:
> I see. If there are bugs in the `godot-cpp` branch we use that only get fixed in `godot-cpp` master, I guess you are right, we'll have to switch to `godot-cpp` master.

Can a specific git revision of godot-cpp master be set as the dependency, with a switch to a stable release once one is available, with the relevant bugs fixed?

Vezzra:
> Can a specific git revision of godot-cpp master be set as the dependency, with a switch to a stable release once one is available, with the relevant bugs fixed?

I guess. Should be possible with CMake's `ExternalProject` feature, and I guess `git submodule` can pull a specific revision/commit too... @agrrr3 apparently has already experience with this, so he can probably comment on it - @agrrr3?

geoffthemedio:
Status of this? @agrrr3 Your input was requested...

agrrr3:
We talked about this in a monthly developer meeting. The decision was rather to just copy the godot sources into the freeorion source tree

geoffthemedio:
So close this then?

Vezzra:
@geoffthemedio, I guess yes, this can be closed, but a short confirmation by @agrrr3 would be good nonetheless... @agrrr3?

o01eg:
Changes was merged to `godot-cpp` `master` branch.

PR #3253, opened 2020-12-02
Remove broken on MacOS mutex

### LINKED ISSUE #3252: Crash to desktop when trying to launch FO on macOS
Bug Report
==========

Environment
-----------

* **FreeOrion Version**: 0.4.10+, build 81807c6a864d51fefdd8433d821609cbc4008da8
* **Operating System**: macOS 10.14.6
* **Fetched as** 
  * Compiled from source

Description
-----------

﻿﻿When I try to launch FO on macOS, the app immediately crashes. On the terminal I get this output:
```
Assertion failed: (e == 0), function ~recursive_mutex, file /BuildRoot/Library/Caches/com.apple.xbs/Sources/libcxx/libcxx-400.9.4/src/mutex.cpp, line 64.
Abort trap: 6
```
freeorion.log: [freeorion.log.zip](https://github.com/freeorion/freeorion/files/5625216/freeorion.log.zip)

o01eg:
Should fix #3252 but supposedly will return back issues with checksum.

### DIFF
--- a/util/Pending.h
+++ b/util/Pending.h
@@ -38,7 +38,6 @@ namespace Pending {
 
         boost::optional<std::future<T>> pending = boost::none;
         std::string filename;
-        std::recursive_mutex m_mutex;
     };
 
     /** Wait for the \p pending parse to complete.  Set pending to boost::none
@@ -46,7 +45,6 @@ namespace Pending {
         Return boost::none on errors.*/
     template <typename T>
     boost::optional<T> WaitForPending(boost::optional<Pending<T>>& pending, bool do_not_care_about_result = false) {
-        std::lock_guard<std::recursive_mutex> lock(pending->m_mutex);
         if (!pending)
             return boost::none;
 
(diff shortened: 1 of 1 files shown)


agrrr3:
@Vezzra can you try if this fixes it for you?

@o01eg can you convert this to a draft? As applying this quickfix will return the checksum sync issues this should not be merged.

Regarding the code it does not have to be a recursive mutex, but then this has to be refactored accordingly that all the entrypoints use the mutex on the same level.

o01eg:
@agrrr3 I've tried to use `std::mutex` but then it freezes on Linux.

agrrr3:
> @agrrr3 I've tried to use `std::mutex` but then it freezes on Linux.

yes, it needs refactoring to work. a version of WaitForPending without locking which is used by all entry point functions which have to lock on the std::mutex.

agrrr3:
@Vezzra can you try if this PR fixes the crash on start for you?

Vezzra (commented):
With this PR, I can launch FO (most of the time), but starting a game fails.

The build seems unstable. Once it crashed already on launch with a segmentation fault, output on the terminal:
```
Segmentation fault: 11
```
The other times the app started, but when I tried to launch a quickstart game, I either get an error message after some time that launching and connecting to the server timed out, or I'm stuck at a completely blank screen. The audio plays, but nothing more happens.

I've zipped up the log files:  [Logs.zip](https://github.com/freeorion/freeorion/files/5635917/Logs.zip)

geoffthemedio:
Is this still needed after #3254 ?

o01eg:
@geoffthemedio No.

PR #3263, opened 2020-12-29
Use Github Actions to build FreeOrion

o01eg:


geoffthemedio:
Does using GitHub actions require not using AppVeyor?

o01eg:
@geoffthemedio Not, it just moves Windows CI from Appveyor to Github Actions.

geoffthemedio:
> ... moves Windows CI from Appveyor to Github Actions.

Why do you want to do that?

o01eg:
I remember we had a plan to move from Appveyor like FreeOrion SDK did.

.github/workflows/build.yml
```diff
@@ -0,0 +1,130 @@
...
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+  push:
+    branches:
+      - master
```
geoffthemedio, line 11: does this mean it will only rub test builds on the master branch when it is pushed an no others unless they are part of a pull request?

o01eg: Yes. Do we need to build pushes always?

geoffthemedio: I find it handy to have a test branch that I can push to test build without needing a pull request always open. Could just add the [refactor_test](https://github.com/freeorion/freeorion/tree/refactor_test) branch for that, though, I suppose.

Vezzra (commented):
I'm sorry, but I don't think I can provide any meaningful input here - my knowledge about all the CI stuff is next to non-existent.

For what it's worth, looks fine from what I can see.

.github/workflows/build.yml
Cjkjvfnby: We can add a special pattern for that.  For example `ci_*`

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

.github/workflows/mono-uninstall.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+#
+# Uninstalling Mono on macOS.
```
Cjkjvfnby, line 3: Could you share more info on why do we need this step?

Cjkjvfnby (commented):
I don't understand this line in logs.  Is it expected?

image

Cjkjvfnby:
The game builds on Ubuntu in `25m 11s`  is it normal numbers?

geoffthemedio:
> The game builds on Ubuntu in `25m 11s` is it normal numbers?

yes, eg: https://travis-ci.org/github/freeorion/freeorion/builds/753497352

Cjkjvfnby:
> yes, eg: https://travis-ci.org/github/freeorion/freeorion/builds/753497352

What is the timing on your machine?

geoffthemedio:
> What is the timing on your machine?

About 4 min with parallel builds on a 12-core desktop...
````
1>------ Build started: Project: Parsers, Configuration: Release Win32 ------
2>------ Build started: Project: GiGi, Configuration: Release Win32 ------
1>Build started 2021-01-10 12:30:58 PM.
2>Build started 2021-01-10 12:30:58 PM.
2>Time Elapsed 00:00:30.86
1>Time Elapsed 00:01:39.36
3>------ Build started: Project: Common, Configuration: Release Win32 ------
3>Build started 2021-01-10 12:32:37 PM.
3>Time Elapsed 00:00:39.40
4>------ Build started: Project: FreeOrion, Configuration: Release Win32 ------
5>------ Build started: Project: FreeOrionD, Configuration: Release Win32 ------
6>------ Build started: Project: FreeOrionCA, Configuration: Release Win32 ------
5>Build started 2021-01-10 12:33:17 PM.
4>Build started 2021-01-10 12:33:17 PM.
6>Build started 2021-01-10 12:33:17 PM.
4>Time Elapsed 00:01:03.35
6>Time Elapsed 00:01:13.77
5>Time Elapsed 00:01:24.65
========== Build: 6 succeeded, 0 failed, 0 up-to-date, 0 skipped ==========
````
My older 4-core laptop took 30-40 min or so, I think.

**Comment by Cjkjvfnby** [comment]
Also, we need to add paths that should trigger build.  There is no sense to run compilation if the translation is changed.

We should cover cpp and build scrips(cmake, github actions, etc)

Since build takes a lot of time this thing worth investment.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

**Comment by o01eg** [comment]
We could cover content as well because it could be tested with unit tests.

File: .github/workflows/build.yml [hunk 3]
```diff
@@ -0,0 +1,130 @@
...
+        run: |
+          mkdir build
+          cd build
+          cmake -DBUILD_TESTING=ON ..
+          cmake --build .
+      - name: Test
+        run: |
+          cd build
+          cmake --build . --config Release --target unittest || echo Failure
+  build-linux-headless:
```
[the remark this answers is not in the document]
geoffthemedio: Whether the human client is built, mainly, I think?

Cjkjvfnby:
> We could cover content as well because it could be tested with unit tests.

`unit tests` is a very confusing term, it is the name of the library that used to do unit, component/integration, and end-to-end tests.  If we are talking about real unit tests most of them are checking logic and can be run only on one system (don't need to run them on all builds).

Cjkjvfnby:
I have remembered that path filtering doesn't work with protection: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-required-status-checks

.github/workflows/build.yml
```diff
@@ -17,27 +17,46 @@ jobs:
...
+        id: build-linux-ccache-timestamp
+        shell: cmake -P {0}
+        run: |
+          string(TIMESTAMP current_date "%Y-%m-%d-%H;%M;%S" UTC)
+          message("::set-output name=timestamp::${current_date}")
+      - name: ccache files
+        uses: actions/cache@v1.1.0
+        with:
+          path: ${{ runner.temp }}/ccache
+          key: build-linux-ccache-${{ steps.build-linux-ccache-timestamp.outputs.timestamp }}
```
Cjkjvfnby, line 41: The timestamp for a cache looks strange to me. I don't understand how it could be reused.

I am not sure what is really cached, so maybe I will tell you some nonsense. 

In general, I expect 2 cache levels. 

1) Dependencies. Key is md5 of file/files with dependencies.  This thing is not changed often, so it could be reused between builds.  
2) Freeorion generated cache.  We can use a commit as a key (this will allow us to reuse cache during actions run on the same commit).  A more elegant way is to use the last commit that touches C++/build files. In that case, we could reuse cache for PRs that change only the content of `default` folder.  The best way is to cache modules separately.

o01eg: As I understand it uses two caches with different action version to cache dependencies and ccache. And `restore-key` allows to use ccache with different timestamp.

o01eg: It's based on https://github.com/freeorion/freeorion/pull/3270 and https://github.com/airvzxf/cache-anything-new-action/issues/1

o01eg:
I moved Linux build only because I'm not sure what broken on MacOS and Windows.

.github/workflows/mono-uninstall.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
Cjkjvfnby, line 1: This file is not part of linux build, could you remove it from this PR?  Probably you need to save it for later.

.github/workflows/build.yml
```diff
@@ -0,0 +1,170 @@
...
+          cache: ${{ runner.temp }}/cache-build-linux-install
+          snapshot: '/'
+          exclude: '/boot /data /dev /mnt /proc /run /sys'
+      - name: Prepare ccache timestamp
+        id: build-linux-ccache-timestamp
+        shell: cmake -P {0}
+        run: |
+          string(TIMESTAMP current_date "%Y-%m-%d-%H;%M;%S" UTC)
+          message("::set-output name=timestamp::${current_date}")
+      - name: ccache files
```
Cjkjvfnby: I'd recommend rewording it to clarify that `ccached` is not a typo but a tool name.

`Cache files with ccache`

.github/workflows/build.yml
```diff
@@ -0,0 +1,170 @@
...
+          restore-keys: |
+            build-linux-ccache-
+      - name: Build
+        env:
+          CACHE_NAME: linux
+          CCACHE_DIR: ${{ runner.temp }}/ccache
+          CMAKE_BUILD_PARALLEL_LEVEL: 2
+        run: |
+          echo "----------------------------------------"
+          ccache -s || true
```
Cjkjvfnby: Let's use the full name for arguments in that script. This increases readability for people who is not familiar with that tool.   
`--show-stats`

.github/workflows/build.yml
```diff
@@ -0,0 +1,170 @@
...
+          echo "----------------------------------------"
+          mkdir build
+          cd build
+          cmake -DBUILD_TESTING=ON ..
+          cmake --build .
+          echo "----------------------------------------"
+          ccache -s || true
```
Cjkjvfnby: We have 2 stats call. They are run in different directories, does it affects anything?

Could we move them to separate steps before and after the build step?  There might be many reasons to see the build log and checking cache is only one of them.  In most cases, cache statistics are not interesting for developers.  Also, you could remove echo parts since they will be in different sections.  Moreover, it would be much easier to compare them because build logs are huge.

Steps in UI:
image

.github/workflows/install-build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
```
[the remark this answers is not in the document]
Cjkjvfnby: @o01eg , could you rename this file?

.github/workflows/install-build.sh
```diff
@@ -0,0 +1,15 @@
...
+	libglew-dev \
+	libogg-dev \
+	libopenal-dev \
+	libsdl2-dev \
+	libvorbis-dev \
+	cppcheck \
+	doxygen \
+	python3-pip \
+	ccache
+python3 -m pip install flake8==3.7.9
```
Cjkjvfnby: Is it a part of the Linux build?  We have a separate build for linters, probably we don't need this here.

Cjkjvfnby (commented):
My comments do not block the merge, most of them could be done later.

@geoffthemedio ,  someone should set build as required for PRs after merging. Could you check if you can do it?  See Require status checks to pass before merging in documentation https://docs.github.com/en/github/administering-a-repository/managing-a-branch-protection-rule

.github/workflows/mono-uninstall.sh
o01eg: It's a part of MacOS build.

.github/workflows/build.yml
o01eg: OK, fixed.

.github/workflows/build.yml
o01eg: Fixed.

.github/workflows/build.yml
o01eg: Moved to separate step.

.github/workflows/install-build.sh
o01eg: Renamed to `linux-install-build.sh` and removed linter installation.

.github/workflows/build.yml
```diff
@@ -0,0 +1,182 @@
...
+      - name: Show ccache statistics after build
+        env:
+          CACHE_NAME: linux
+          CCACHE_DIR: ${{ runner.temp }}/ccache
+        run: |
+          ccache --show-stats || true
+      - name: Test
+        run: |
+          cd build
+          cmake --build . --config Release --target unittest || echo Failure
```
Cjkjvfnby, line 69: @o01eg , failure on the test step does not fail the build. Is it intended?  `33% tests passed, 2 tests failed out of 3`

.github/workflows/build.yml
```diff
@@ -0,0 +1,182 @@
...
+          cd build
+          cmake -DBUILD_TESTING=ON ..
+          cmake --build .
+      - name: Show ccache statistics after build
+        env:
+          CACHE_NAME: linux
+          CCACHE_DIR: ${{ runner.temp }}/ccache
+        run: |
+          ccache --show-stats || true
+      - name: Test
```
Cjkjvfnby: Lets name it `Run C++ unit tests` to avoid confusion with Python unit tests.

.github/workflows/build.yml
o01eg: Yes, it is intended. We get some strange failure here sometimes.

.github/workflows/build.yml
o01eg: Renamed.

.github/workflows/build.yml
Cjkjvfnby: This not how tests supposed to work. If they do not fail it is the same as not having them, but longer.

Do you know if someone has a plan to fix it? 

Anyway, this is definitely out of the scope of this PR.

Cjkjvfnby:
@o01eg , do we need to do a check that if compilation is failed, this action is failed too?

.github/workflows/build.yml
o01eg: I cann't reproduce it so maybe someone else could fix it.

o01eg:
No, I disabled MacOS and Windows because broken compilation made action to fail.

geoffthemedio:
> @geoffthemedio , someone should set build as required for PRs after merging. 

Do you mean before merging? From what I see, there are configurable rules that branches must meet to be merged, but that doesn't seem related to automatically starting a test build *after* merging...

> Could you check if you can do it? See Require status checks to pass before merging in documentation https://docs.github.com/en/github/administering-a-repository/managing-a-branch-protection-rule

I see:
branch_protection_rule

Not sure what you want, though.

Cjkjvfnby:
@geoffthemedio , 

This checkbox should be checked after merging this PR, to make build required for all branches before merging. 
image

geoffthemedio:
> This checkbox should be checked after merging this PR, to make build required for all branches before merging.

As long as it doesn't time out ~40% of the builds...

Cjkjvfnby (approved):

Cjkjvfnby:
> As long as it doesn't time out ~40% of the builds...

Am I right that build failures are already a part of the process and this PR does not change how things merged now?

That means that if we merge these changes and something went wrong, it won't be a huge trouble, so we can fix it without a strict deadline.

hifron:
One huge Github rules them all... ?

Vezzra (commented):
I'm afraid I still don't know what meaningful input I could give here.

Can only repeat what I said before: looks ok to me, but that doesn't count for much given I don't really know much about this stuff.

geoffthemedio:
Is there a way to tell CMake or the test build to use some extra include directories to fix [this](https://github.com/freeorion/freeorion/pull/3263/checks?check_run_id=1947523282#step:8:354)?
````
In file included from /usr/include/x86_64-linux-gnu/SDL2/_real_SDL_config.h:33:
/usr/include/x86_64-linux-gnu/SDL2/SDL_platform.h:179:10: fatal error: 'begin_code.h' file not found
#include "begin_code.h"
         ^~~~~~~~~~~~~~
````
As discussed [here](https://bugs.launchpad.net/ubuntu/+source/libsdl2-ttf/+bug/1872023)...

**Comment by o01eg** [comment]
Docs tell about CXXFLAGS and CFLAGS https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_INIT.html#variable:CMAKE_%3CLANG%3E_FLAGS_INIT

**Comment by geoffthemedio** [comment]
Tried that, doesn't seem to help: https://github.com/freeorion/freeorion/commit/4d567c178ee6504f0e0a98ffcc0ce0437b488280

**Comment by o01eg** [comment]
@geoffthemedio It works for me, should be `CXXFLAGS` not `CPPFLAGS`.

**Comment by geoffthemedio** [comment]
> @geoffthemedio It works for me, should be `CXXFLAGS` not `CPPFLAGS`.

Can you tell specifically which include directory is needed for Ubuntu 20.04?

**Comment by o01eg** [comment]
@geoffthemedio I used same from `.travis.yml`.

```
-I/usr/include/SDL2 -I/usr/include/x86_64-linux-gnu/SDL2/ -I/usr/include/x86_64-linux-gnu/
```

geoffthemedio:
> @geoffthemedio I used same from `.travis.yml`.
> 
> ```
> -I/usr/include/SDL2 -I/usr/include/x86_64-linux-gnu/SDL2/ -I/usr/include/x86_64-linux-gnu/
> ```

My question was which of those is really needed?Probably not all are.

o01eg:
I think I can only find out it with trial and error.

geoffthemedio:
It would probably also be better if it could be fixed directly in CMakeLists.txt rather than in the CI config.

o01eg:
@geoffthemedio As I understand it is clang issue which is supposed to be fixed later.

geoffthemedio:
> @geoffthemedio As I understand it is clang issue which is supposed to be fixed later.

Does that change how it should be fixed now for FreeOrion?

o01eg:
@geoffthemedio Yes, I think so.

o01eg:
Superseded by #3411 , #3409 and #3408

PR #3211, opened 2020-10-29
Introduce python parser

o01eg:
Preparation for using python's parser.

Moves minimal python initialization and finalization to common library.

Should be usable in both cases if we use python ast or python interpretation.

Follows #3342

PR #3265, opened 2020-12-29
Switch game rules to python script

o01eg:
Follows-up #3211

Uses supposed format for Python FreeOrion Content Script is function with named arguments.
Enables flake8 to check scripts.
Uses deferred futures because python must use the same thread.

It looks like big speed up at least for game rules. Boost parser required 19-21 ms, but python parser uses around 1 ms.


o01eg:
It fails on Ubuntu Bionic 18.04 LTS (Travis CI) but works correctly on my Gentoo and Ubuntu Focal 20.04 LTS.

Boost changed from 1.62 to 1.71 and python from 3.6.9 to 3.8.5

o01eg:
Raising boost version to 1.65 on Ubuntu Bionic doesn't help.

o01eg:
Failed on 

```cpp
type_int = eval("int", dict());
```

o01eg:
Looks like python 3.6.9 cannot evaluate `int`, but it somehow works in REPL.

```
   27	  if (local.is_none()) local = global;
   28	  // should be 'char const *' but older python versions don't use 'const' yet.
   29	  char *s = python::extract<char *>(string);
> 30	  PyObject* result = PyRun_String(s, Py_eval_input, global.ptr(), local.ptr());
   31	  if (!result) throw_error_already_set();
   32	  return object(detail::new_reference(result));
```

geoffthemedio:
Is this implemented as Python function calls to an exposed API, or as a Python syntax tree parse? (Looks like function calls to me)

> Boost changed from 1.62 to 1.71 and python from 3.6.9 to 3.8.5

Ideally it will still work with a wide range of Boost versions.

o01eg:
@geoffthemedio It's implemented on Python calls.

> Ideally it will still work with a wide range of Boost versions.

Looks like some Python issue.

```
[2020-12-30 10:14:26.265488] [0x00007f085361b100] [info]    Traceback (most recent call last):
[2020-12-30 10:14:26.265502] [0x00007f085361b100] [info]      File "<string>", line 1, in <module>
[2020-12-30 10:14:26.265507] [0x00007f085361b100] [info]    NameError: name 'int' is not defined
```

geoffthemedio:
> @geoffthemedio It's implemented on Python calls.

This approach is then not elegantly usable for any FOCS replacement that needs more-complex data structures to be parsed, like ValueRefs, Conditions, and similar. It's debatable whether just speeding up the game rules parsing a few 10s of ms is worth the extra complexity / different format...

o01eg:
@geoffthemedio We could use operation override to generate complex `ValueRefs`.

geoffthemedio:
> @geoffthemedio We could use operation override to generate complex `ValueRefs`.

Sounds plausible...

o01eg:
Fixed it with copy `builtins` to `exec` globals.

Vezzra (changes requested):
I can build and launch FO successfully with this PR, however, when I try to start a "Quickstart" game, it fails with "Connection to server lost".

Logs: [logs.zip](https://github.com/freeorion/freeorion/files/5757411/logs.zip)

Vezzra:
@o01eg,

> It's implemented on Python calls.

So you decided to try the execution approach after all? Not that I'm complaining, it just sounded like you'd prefer the approach where we'd use Python only as parser in the online chat meeting on Monday.

So, it will be possible to do something like this:
```
for a_variant in set_of_some_variants:
    Building(
        param1 = whatever,
        param2 = something else,
        ...
        paramM = a_variant,
        ...
        paramN
    )
```
?

geoffthemedio:
> So, it will be possible to do something like this:

To be determined / implemented. This pull request still needs more stuff to be supported to allow non-trivial (eg. ValueRefs, not just strings or numbers) to be passed.

o01eg:
> So, it will be possible to do something like this:

Yes

o01eg:
> I can build and launch FO successfully with this PR, however, when I try to start a "Quickstart" game, it fails with "Connection to server lost".
> 
> Logs: [logs.zip](https://github.com/freeorion/freeorion/files/5757411/logs.zip)

I forgot to make change .txt to .py in the game code. Fixed now.

Vezzra (approved):
Yep, works now. I can build and run FO, and launch a Quickstart game. From what I can see everything seems to be there, buildings etc. exist, so the parsing apparently has been successful.

Checked the logs, the only error message related to parsing stuff has been this in `freeorion.log`:
```
17:15:32.654523 {0x0000700000a77000} [debug] client : Pending.h:67 : Dont care for result of parsing "common". Have to get() once to release shared state in pending future.
17:15:32.654590 {0x0000700000a77000} [error] client : Pending.h:78 : Parsing of "common" failed with error: The state of the promise has already been set.
```
Don't know if that has anything to do with this PR though.

geoffthemedio:
> Yep, works now. I can build and run FO, and launch a Quickstart game. From what I can see everything seems to be there, buildings etc. exist, so the parsing apparently has been successful.

Main thing to check is, I think, game rules. They appear to be the only thing parsed with Python so far.

Vezzra:
Ah, ok. Just ran another test game, this time I used the Single Player Game Setup - enabled cheap and fast techs, buildings and policies, all of which worked as expected.

default/scripting/game_rules.focs.py
```diff
@@ -0,0 +1,134 @@
+GameRule(
```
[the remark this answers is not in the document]
o01eg: I just reproduce FOCS with minimal changes. I had idea to infer type from default value to remove `type` argument. If argument have no sense we can just not use it.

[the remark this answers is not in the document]
Cjkjvfnby: > If argument have no sense we can just not use it.

If it is possible to add an argument that has no sense it will be added. It always works in that way. And of cause, it will confuse the user, because of the mismatch of the expectations and reality. We can protect ourselves with documentation and strict validation.  I'm not aware if documentation is in plans. But validation should not be very hard to implement.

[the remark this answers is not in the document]
o01eg: I've setup pep8 check. Not sure what could be done for stricter validation.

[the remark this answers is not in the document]
Cjkjvfnby: I mean about the contnet.  Something like that.

Each type expects its own set of arguments. Some copy-paste issues and misunderstandings will get some problems.  I'd like to catch them on the CI stage.

```python
GameRule(
    name="XXX",
    description="XXX",
    category="BALANCE",
    type=str,
    max="x")
    
GameRule(
    name="RULE_TEST_STRING",
    description="RULE_TEST_STRING_DESC",
    category="TEST",
    type=bool,
    allowed=[true, false, "maybe"])
```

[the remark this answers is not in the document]
o01eg: We run unit tests for parsing code now so could check it here.

util/Pending.h
```diff
@@ -127,6 +127,52 @@ namespace Pending {
...
             std::async(std::launch::async, parser, path), path.filename().string());
     }
+
+    /** Return a Pending<T> constructed with \p parser, \p arg1, and \p path*/
+    template <typename Func, typename Arg1>
+    auto Parsed(const Func& parser, const Arg1& arg1, const boost::filesystem::path& path)
+        -> Pending<decltype(parser(arg1, path))>
+    {
+        auto f = std::async(std::launch::deferred, parser, arg1, path);
+        f.wait();
```
Cjkjvfnby: > auto f = std::async(std::launch::deferred, parser, arg1, path);

The deferred future is created

> f.wait();

Deferred future starts executing. 
Can we just use std::launch::async policy?

Also, `wait` is blocking, so this thread will wait until the future is finished.  Could we exit this method without waiting until the task finished?

util/PythonCommon.h
```diff
@@ -0,0 +1,50 @@
...
+       and no longer available.
+
+       Call PyErr_Print() if the exception is an error.
+
+       HandleErrorAlreadySet is idempotent, calling it multiple times
+       won't crash or hang the process.
+     */
+    void HandleErrorAlreadySet();
+
+    bool         Initialize();        // initializes and runs the Python interpreter, prepares the Python environment
```
Cjkjvfnby, line 33: Can you move the comment to a function name?
`InitializeInterpretorAndPreparePythonEnvironmet` is a little bit verbose, but we can get rid of the comment.

If you put the documentation into the code objects you'll get self-documenting code.

util/PythonCommon.h
```diff
@@ -0,0 +1,50 @@
...
+
+    virtual bool InitCommonImports(); // initializes Python imports, should be implemented by derived classes
```
Cjkjvfnby: > should be implemented by derived classes

Contract via comment does not always work.  Maybe you can make it abstract and make the compiler care about overriding.  Or put text inside the default implementation as an exception, then you will be warned in runtime.

msvc2017/Common/Common.vcxproj
```diff
@@ -114,7 +114,7 @@
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <IntrinsicFunctions>true</IntrinsicFunctions>
       <PreprocessorDefinitions>NDEBUG;_USRDLL;FREEORION_WIN32;_DLL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <AdditionalIncludeDirectories>../../../include/;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>../../../include/;../../../include/python3.5/;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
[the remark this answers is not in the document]
geoffthemedio: It's what's in the Windows SDK, so it's fine for the MSVC project files. It's also how the AI project has its include directories specified, so probably a separate issue if that's to be changed.

util/PythonCommon.h
geoffthemedio: > make it abstract

pure virtual function `= 0` would make sense

util/PythonCommon.h
geoffthemedio: In general yes, but in this case, does it really clarify anything more usefully with the longer name...?

util/Pending.h
o01eg: No, because of python we cannot use threads other than python interpreter was initialized.

o01eg:
This PR is very big, could you first review and merge followed one #3211 ?

client/human/GGHumanClientApp.cpp
```diff
@@ -1549,7 +1561,16 @@ void GGHumanClientApp::UpdateFPSLimit() {
 void GGHumanClientApp::HandleResoureDirChange() {
     if (!m_game_started) {
         DebugLogger() << "Resource directory changed.  Reparsing universe ...";
-        StartBackgroundParsing();
```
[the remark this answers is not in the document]
o01eg: Python in client and python in AI and server are very different. Not sure how to simplify code.

util/PythonCommon.h
o01eg: Client and test doesn't override it because don't use python imports. I'll write comment differently.

geoffthemedio:
> This PR is very big, could you first review and merge followed one #3211 ?

I don't want to merge any of the Python parsing stuff until it's clear it will be used for all the parsing. As far as I know you've just implemented the relatively simple GameRules parsers, but as yet not the more difficult stuff like a full Tech script with nested Condition, ValueRef, and Effects scripted.

util/Pending.h
Cjkjvfnby: > No, because of python we cannot use threads other than python interpreter was initialized.

Is it about using async instead of deffered or about blocking call?

Also, the name Pending is a little bit confusing.  At this point, we are not pending anything we already called the wait method.

util/PythonCommon.h
Cjkjvfnby: We have at least 2 places in code declaration and usage. And only one place has this comment. 

So the long name will spread the documentation along with the code.

util/Pending.h
o01eg: Yes, it just about to use the same code to work with parsing result as before.

Vezzra (approved):
I can build and run the PR on macOS. A quick test game also showed no apparent issues.

o01eg:
@geoffthemedio Does it stuck or not?

o01eg:
Also I rebased it.

geoffthemedio:
> @geoffthemedio Does it stuck or not?

Runs seemingly OK

o01eg:
Nice, then something broken in https://github.com/freeorion/freeorion/pull/3373

geoffthemedio:
Hangs with
````
09:32:05.499699 {0x00009f34} [debug] client : Pending.h:54 : Waiting for parse of "common" to complete.
09:32:05.499699 {0x00009f34} [debug] client : Pending.h:60 : WaitForPendingUnlocked another wait_for round
````

o01eg:
Nice, and could you check previous #3211 ?

Vezzra:
@o01eg, as this PR has been merged (at least Github shows it as merged), the status label should also be set to "merged", not "superseded". The commits/branch of this PR have/has not been discarded, but merged into master (even if that happened as part of another PR).

PR #3267, opened 2020-12-30
Treat empty files as failure

o01eg:


geoffthemedio:
Why?

I don't have an opinion, otherwise.

o01eg:
Mostly for https://github.com/freeorion/freeorion/issues/3204 to skip overridden empty files.

Issue #3257, opened 2020-12-19
Exception float division by zero occurred during generateOrders


a2geek:
Bug Report
==========
Am seeing divide by zero errors in logs. Uncertain conditions required. Game is random setting, 50 planets, 3 AI's at Turtle mode (yeah, I'm still stinky at this game!).

```
[19 Dec 15:27:11] AI_3: AI_Error: AI script error : FreeOrionAI.py:_error_handler():73  - Exception float division by zero occurred during generateOrders
Traceback (most recent call last):
  File "/snap/freeorion/128/share/freeorion/default/python/AI/FreeOrionAI.py", line 70, in _error_handler
    res = func(*args, **kwargs)
  File "/snap/freeorion/128/share/freeorion/default/python/common/listeners.py", line 42, in wrapper
    res = funct(*args)
  File "/snap/freeorion/128/share/freeorion/default/python/AI/FreeOrionAI.py", line 336, in generateOrders
    aistate.prepare_for_new_turn()
  File "/snap/freeorion/128/share/freeorion/default/python/AI/AIstate.py", line 998, in prepare_for_new_turn
    self.__report_last_turn_fleet_missions()
  File "/snap/freeorion/128/share/freeorion/default/python/AI/AIstate.py", line 952, in __report_last_turn_fleet_missions
    CombatRatingsAI.get_fleet_rating(fleet_id),
  File "/snap/freeorion/128/share/freeorion/default/python/AI/CombatRatingsAI.py", line 395, in get_fleet_rating
    return FleetCombatStats(fleet_id, consider_refuel=False).get_rating(enemy_stats)
  File "/snap/freeorion/128/share/freeorion/default/python/AI/CombatRatingsAI.py", line 370, in get_rating
    return combine_ratings_list([x.get_rating(enemy_stats, ignore_fighters) for x in self.__ship_stats])
  File "/snap/freeorion/128/share/freeorion/default/python/AI/CombatRatingsAI.py", line 370, in <listcomp>
    return combine_ratings_list([x.get_rating(enemy_stats, ignore_fighters) for x in self.__ship_stats])
  File "/snap/freeorion/128/share/freeorion/default/python/AI/CombatRatingsAI.py", line 284, in get_rating
    shield_factor = e_total_attack / e_net_attack
ZeroDivisionError: float division by zero
```

Environment
-----------

* **FreeOrion Version**: v0.4.10.1
* **Operating System**: Ubuntu 20.04, installed via snap
[...]

PR #3271, opened 2021-01-09
Fix zero division in shield factor calculation see #3257

Cjkjvfnby:
It is a rare issue that was reported by user. I have extracted logic to a separate function and add a protection for case when number of shots is zero.  It is not clear why does it happens, maybe  because of some species treats. Issue was introduced on turn 721, maybe it was some impure which was almost destroyed and its stats was not very actual.

### DIFF
--- a/default/python/AI/CombatRatingsAI.py
+++ b/default/python/AI/CombatRatingsAI.py
@@ -270,19 +270,11 @@ class ShipCombatStats:
         # The fighter rating calculations are heavily based upon the enemy stats.
         # So, for now, we compare at least against a certain standard enemy.
         enemy_stats = enemy_stats or get_aistate().get_standard_enemy()
-
         my_attacks, my_structure, my_shields = self.get_basic_stats()
         # e_avg_attack = 1
         if enemy_stats:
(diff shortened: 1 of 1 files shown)


geoffthemedio:
I don't think we plan another v0.4.10 bugfix release, so not sure what you mean with the cherry-pick-for-release tag...

Cjkjvfnby:
> I don't think we plan another v0.4.10 bugfix release

This issue was reproduced in the released version.

geoffthemedio:
I guess we can pick it into the v0.4.10 release branch anyway, even if no release is planned? @Vezzra ?

Vezzra:
Well, of course we could cherry pick it to the 0.4.10 release branch regardless if we actually do another bugfix release for 0.4.10. However, in that case what should we set the milestone to? 0.4.10.2 doesn't exist yet. 0.5 would be misleading if we do 0.4.10.2.

It's probably better to just keep the "cherry-pick for release" tag, which I'm going to check whatever the next release will be. At that point we'll know for sure and can set the right milestone (and do the cherry picking then).

PR #3282, opened 2021-01-20
Reduce Conc. Camps industry boost

Oberlus:
Reduce concentration camps industry bonus from 2 x population to 0.75 x population.
It should have been reduced when the rest of bonuses.

Discussion thread: https://www.freeorion.org/forum/viewtopic.php?p=104842#p104842

### DIFF
--- a/default/scripting/buildings/CONC_CAMP.focs.txt
+++ b/default/scripting/buildings/CONC_CAMP.focs.txt
@@ -25,7 +25,7 @@ BuildingType
             priority = [[TARGET_AFTER_SCALING_PRIORITY]]
             effects = SetTargetIndustry value = Value + Target.Population
                         * (NamedReal name = "BLD_CONC_CAMP_TARGET_INDUSTRY_PERPOP"
-                                     value = 10 * [[INDUSTRY_PER_POP]])
+                                     value = 3.75 * [[INDUSTRY_PER_POP]])
 
         EffectsGroup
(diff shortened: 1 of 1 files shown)


default/scripting/buildings/CONC_CAMP.focs.txt
```diff
@@ -25,7 +25,7 @@ BuildingType
             priority = [[TARGET_AFTER_SCALING_PRIORITY]]
             effects = SetTargetIndustry value = Value + Target.Population
                         * (NamedReal name = "BLD_CONC_CAMP_TARGET_INDUSTRY_PERPOP"
-                                     value = 10 * [[INDUSTRY_PER_POP]])
+                                     value = 3.75 * [[INDUSTRY_PER_POP]])
```
geoffthemedio, line 28: A bit confusing how this relates to the description and .75 vs 2x pop...

Oberlus: I'm not sure I see your confusion.
INDUSTRY_PER_POP == 0.2
3.75 * 0.2 == 0.75
10 * 0.2 == 2.0

No body ever changes INDUSTRY_PER_POP, RESEARCH_PER_POP...
I've been thinking on changing their values to 1.0 and update the factors accordingly (so in this case I would set 0.75*[[INDUSTRY_PER_POP]]).

PR #3285, opened 2021-01-25
Unlock Divine Authority policy at Transcendend Architecture tech

### LINKED ISSUE #3284: Divine Authority is not unlocked by any tech
Bug Report
==========

* **FreeOrion Version**: 60a10c5

No tech unlocks PLC_DIVINE_AUTHORITY:
```
freeorion$ grep -rl DIVINE default/scripting/*
default/scripting/policies/DIVINE_AUTHORITY.focs.txt
freeorion$
``` 

Maybe, fluff wise, it could be done in TRANSCEND_ARCH. It is not a perfect match, and has no reasoning behind about if that's the right place balance-wise, but I don't see anything better than that.
Alternatively, place it at PSIONICS, but that one is already overloaded of policies (I would relocate some of those, but that's another issue), and you don't need fantastical mind powers to hold divine authority among your people.

Oberlus:
Fixes #3284

### DIFF
--- a/default/scripting/techs/construction/TRANSCEND_ARCH.focs.txt
+++ b/default/scripting/techs/construction/TRANSCEND_ARCH.focs.txt
@@ -10,7 +10,11 @@ Tech
         "CON_ORGANIC_STRC"
         "CON_NDIM_STRC"
     ]
-    unlock = Item type = Building name = "BLD_MEGALITH"
+    unlock = 
+    unlock = [
+        Item type = Building name = "BLD_MEGALITH"
+        Item type = Policy name = "PLC_DIVINE_AUTHORITY"
+    ]
     effectsgroups = EffectsGroup
         scope = Source
         effects = SetEmpireMeter empire = Source.Owner meter = "SOCIAL_CATEGORY_NUM_POLICY_SLOTS" value = Value + 1


geoffthemedio:
I feel like this should be unlocked much earlier, if not from the start, unless there's some special future spin on it... The concept is historical more so than sci-fi.

Oberlus:
I feel the same. Let's give it some more thought to other techs. But unlocked by start might be fine.

hifron:
There is branch Pre-warp start. No development of DA policy at TA techs? When there were no cars, no proper roads, doesn't mean there were no movement. But to develop cars means not only tech, but something more... Way of life and thinking. And to want something more than develop and travel. But without development there is something like people in Africa. Have techs, but are not developed...

geoffthemedio:
can move later... better to have it unlocked at least somewhere until then

default/scripting/techs/construction/TRANSCEND_ARCH.focs.txt
```diff
@@ -10,7 +10,11 @@ Tech
         "CON_ORGANIC_STRC"
         "CON_NDIM_STRC"
     ]
-    unlock = Item type = Building name = "BLD_MEGALITH"
+    unlock =
```
geoffthemedio, line 13: Didn't notice earlier or something went weird with merging, but this extra line is a problem...

Oberlus: Oops. Yes, that was wrong in my PR, sorry.

Oberlus:
> can move later... better to have it unlocked at least somewhere until then

Yes, it must be moved. Early game, maybe just at start.

Maybe there should be some sort of Social tech unlocked at start, like "Galactic Government/Sociology", with pre-unlocked policies attached to it, and create some, few, new "social" techs to have more space for political/social policies, with two main branches, one for oppression and one for liberty.

PR #3286, opened 2021-01-25
Convert turn state to submodules

Cjkjvfnby:
@geoffthemedio as I understand that there is no one who can review this code now, but could you check the module description.  It is about code organization and I want to get some feedback if my idea is clearly explained.  It is located in the `default/python/AI/turn_state/__init__.py`


default/python/AI/turn_state/__init__.py
```diff
@@ -0,0 +1,39 @@
+"""
+This module contains tools to work with pre-aggregated information about AI state.
+This information is not stored in the save game and recalculated each turn.
```
geoffthemedio: a bit unclear... "is not (stored in the same game and recalculated)" or "is not (stored in the save game), but rather is recaulcated"

default/python/AI/turn_state/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+The main purpose of this module is to separate code that aggregates information and code that makes a decision.
+Decision-making code is smart and this is the dumb module,
+that means that we try to make this module API as comfortable as possible.
+
+Module API:
+Module API is represented as several functions.
+All public functions just lazy objects.
```
geoffthemedio: ... functions are just...

default/python/AI/turn_state/__init__.py
```diff
@@ -0,0 +1,39 @@
+"""
+This module contains tools to work with pre-aggregated information about AI state.
+This information is not stored in the save game and recalculated each turn.
+
+The main purpose of this module is to separate code that aggregates information and code that makes a decision.
+Decision-making code is smart and this is the dumb module,
+that means that we try to make this module API as comfortable as possible.
```
geoffthemedio: that means that we try to -> which means that we will try to   (or)  which means that we have tried to

geoffthemedio: meaning of "comfortable" is unclear

default/python/AI/turn_state/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+Module API:
+Module API is represented as several functions.
+All public functions just lazy objects.
+They initialized with the first request and cache its value till the end of the turn.
```
geoffthemedio: They are intialized upon the first request each turn, and the value is then cached until the end of that turn.

default/python/AI/turn_state/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+Avoiding classes that are exposed to outer scope requires some additional calls inside functions,
+ but this price is not too high to get a simple API.
+
+The module is split into submodules by the idea that each submodule works with the same set of data.
```
geoffthemedio: The module is split into sub-modules, based around the idea that each submodule should work with the same set of data.

default/python/AI/turn_state/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+This allows to achieve the following:
+- Each submodule is relatively small, so it is easy to understand and easy to extended;
```
geoffthemedio: - Each submodule is relatively small, and is therefore easy to understand and extend.

geoffthemedio:
Overall, what you wrote is understandable, but rather conceptual or motivation, and not very specific about what it really means. There are some modules that are useful for accessing or manipulating AI state, and each module can easily work on a different kind of pack of data. You probably are intentionally not giving such specific details about how thing are organized in this bit of text, though.

Cjkjvfnby:
Thank you for your review.

> Overall, what you wrote is understandable, but rather conceptual or motivation, and not very specific about what it really means. There are some modules that are useful for accessing or manipulating AI state, and each module can easily work on a different kind of pack of data. You probably are intentionally not giving such specific details about how thing are organized in this bit of text, though.

My idea is to make the AI code more structured.  Introducing structure - is introducing rules.  So I try to split code into different buckets. One bucket is this module: it is responsible for data aggregation.  And it has a simple rule: If you have data that can be recalculated from querying freeOrionAIInterface, it a good idea to put it here.  Where put the new staff inside the module actually does not matter, because this decision is cheap to change.  It is easy to reshuffle this module inside without breaking its public API.

I think I need to state this more clear in the module documentation.

This type of code organization looks good for me now.  Probably better vision with emerge after some time, but having small modules with little dependencies means that reorganization can be done very fast.

geoffthemedio:
so merge now or after edits?

Cjkjvfnby:
> so merge now or after edits?  

I have updated the docstring.

Could you use `squash and merge` button for my PRs?  It makes history looks better.

PR #3288, opened 2021-01-28
Remove getOrders from AI interface

Cjkjvfnby:
This code was added for debugging purposes, so now it can be removed.
Only `size` method of OrderSet is working, conversion set to list leads to error: `No to_python (by-value) converter found for C++ type: class std::shared_ptr<class Order>.`

Python code was introduced with the commit message: "Added debug output of size of issued orders when loading and when done issuing orders."

PS. I have not compiled it myself.

### DIFF
--- a/client/AI/AIWrapper.cpp
+++ b/client/AI/AIWrapper.cpp
@@ -13,7 +13,6 @@
 #include "../../util/i18n.h"
 #include "../../util/MultiplayerCommon.h"
 #include "../../util/OptionsDB.h"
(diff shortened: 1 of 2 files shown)


geoffthemedio:
Could probably fix it rather than removing, if it is needed / useful.

Cjkjvfnby:
> Could probably fix it rather than removing, if it is needed / useful.

This is dead code.  Nobody using it, so fixing and testing it will require much more time than removing it.

PR #3292, opened 2021-01-30
[AI] Cleanup code

Cjkjvfnby:
- Replace `rtype` and `type` with annotations
- Reduce union types according to actual usage
- Simplify function arguments to reduce module dependencies
- Move function definition after its requirements


geoffthemedio:
AI unit tests fail... Should it be fixed?

PR #3293, opened 2021-01-31
[AI] Remove situation report binding from AI interface

Cjkjvfnby:
- This code is not used by AI.
- Some sitrep object methods are broken, for example: `getTags`.
- Probably AI could get some useful information from situation reports, but this general implementation is too general.

This binding introduced in the commit https://github.com/freeorion/freeorion/commit/e01c6fbc757206c1ae661c76c328fccff0630309

### DIFF
--- a/default/python/AI/AIstate.py
+++ b/default/python/AI/AIstate.py
@@ -610,12 +610,12 @@ class AIstate:
                 debug("In system %s: Ignoring lost fleets since known threats could cause it.", system)
                 lost_fleet_rating = 0
 
-            # TODO use sitrep combat info rather than estimating stealthed enemies by fleets lost to them
             # TODO also only consider past stealthed fleet threat to still be present if the system is still obstructed
             # TODO: track visibility across turns in order to distinguish the blip of visibility in (losing) combat,
             #       which FO currently treats as being for the previous turn,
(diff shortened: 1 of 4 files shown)


geoffthemedio:
> this general implementation is too general

What do you mean by that?

Cjkjvfnby:
>What do you mean by that?

For example, to get information about combat, the API is too general. 

You can use `getType/getTags/getTurn` for filtering, and `getDataString` as a source of the information. 

The situation report's purpose is user informing. Adding additional responsibility to it leads to possible errors. 
 People who make better UX for users may not be aware that this string is parsed in AI.

agrrr3:
I guess a sitrep report binding would mainly useful for centaurs (player which use some kind of python automation) and very maybe for machine learning stuff.
 
I thought combat report would be only interesting for AI if one works on machine learning. 
But i see in the code changes there was the idea to estimate combat strength of stealthed enemies via the combat report. I think combat reports need their own interface.

Else are there any known events in the sitrep which are not available through the game state?
(sitrep can definitly do that, but i am not sure if there is such a case). Because sitrep also indicates events.
To identify the information in sitreps the sitrep id, and the tags/ tag values is a valid abstraction. These usually have semantic meaning and do not need to be parsed out of string.

PR #3295, opened 2021-02-02
Make influence upkeep tractable

Oberlus:
Influence upkeep is untractable currently: after a certain number of planets you can't keep positive influence.

This PR deals with that:

- Move influence upkeep maluses to a tech (custom advisories, in lack of a proper tech for empire upkeep/management) from species influence macros and centralization policy.
- Make upkeep based in number of colonies use square root to make it tractable later game.
- Make Centralization incur into penalization stronger than usual and negate number-based upkeep.
- Make Confederation negate distance-based upkeep and supply-disconnected malus.
- Make Centralization penalize colony supply and bonus capital supply.

### DIFF
--- a/default/customizations/custom_sitreps.txt
+++ b/default/customizations/custom_sitreps.txt
@@ -5,16 +5,12 @@
 // a suitable place to start is http://www.freeorion.org/index.php/Free_Orion_Content_Script_(FOCS)
 
 
-// Important, the following line "effectsgroups = [" is functional and important, do not change it
-effectsgroups = [
-
-
 // normally, for a custom sitrep the player will write their own message here, and label, and not look them up in the Stringtable.  For this first custom sitrep,
 // however, since it is the primary way we introduce the custom sitrep capability to players, we use stringtable references so that the message and label may be more 
 // readily translated into multiple languages as part of our standard distribution.
 
-
-EffectsGroup
+CUSTOM_SITREPS
+'''EffectsGroup
     scope = Source
     activation = Turn low = 0 high = 0
     effects =
@@ -160,7 +156,7 @@ EffectsGroup
                 Unowned
             ]
             VisibleToEmpire empire = Source.Owner
-            Turn low = [[ETA_NEXT_TURN]] high = [[ETA_NEXT_TURN]]
+            Turn low = (CurrentTurn + LocalCandidate.ETA - 1) high = (CurrentTurn + LocalCandidate.ETA - 1)
             (RootCandidate.ID = LocalCandidate.NextSystemID)
         ]
         effects = GenerateSitRepMessage
@@ -171,9 +167,6 @@ EffectsGroup
                 tag = "system"  data = Target.ID
             ]
             empire = Source.Owner
-
-ETA_NEXT_TURN
-'''(CurrentTurn + LocalCandidate.ETA - 1)'''
 
 
 //*********************************************************************************************************************************************************
@@ -313,8 +306,6 @@ EffectsGroup
             ]
             empire = Source.Owner
     ]
-
-// Note, the "]" on the following line is functional and important; it must be present and all EffectsGroups must be above it.
-]
+'''
 
 #include "/scripting/common/priorities.macros"
--- a/default/scripting/techs/spy/CUSTOM_ADVISORIES.focs.txt
+++ b/default/scripting/techs/spy/CUSTOM_ADVISORIES.focs.txt
@@ -6,4 +6,28 @@ Tech
     researchcost = 1
     researchturns = 1
     tags = [ "PEDIA_SPY_CATEGORY" ]
+    effectsgroups = [
+        [[CUSTOM_SITREPS]]
+
+        EffectsGroup      // colonies consume influence proportional to how many planets the empire controls
+            scope = And [
+                Planet
+                OwnedBy empire = Source.Owner
+                Population low = 0.001
+                Not Capital
+            ]
+            stackinggroup = "IMPERIAL_PALACE_MANY_PLANETS_INFLUENCE_PENALTY"
+            accountinglabel = "COLONY_ADMIN_COSTS_LABEL"
+            effects = SetTargetInfluence value = Value
+                + Statistic If condition = Not EmpireHasAdoptedPolicy empire = Source.Owner name = "PLC_CENTRALIZATION"
+                    * (NamedReal name = "TARGET_INFLUENCE_PER_COLONY" value = -0.2)
+                    * (Statistic Count condition = And [
+                        Planet
+                        OwnedBy empire = Source.Owner
+                    ])^0.5
+                + Statistic If condition = Not EmpireHasAdoptedPolicy empire = Source.Owner name = "PLC_CONFEDERATION"
+                    * (NamedReal name = "TARGET_INFLUENCE_PER_JUMP" value = -0.2)
+                    * JumpsBetween object = Source.ID object = Target.ID
+    ]
+
 #include "/customizations/custom_sitreps.txt"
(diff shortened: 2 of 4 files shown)


default/customizations/custom_sitreps.txt
```diff
@@ -160,7 +156,7 @@ EffectsGroup
                 Unowned
             ]
             VisibleToEmpire empire = Source.Owner
-            Turn low = [[ETA_NEXT_TURN]] high = [[ETA_NEXT_TURN]]
+            Turn low = (CurrentTurn + LocalCandidate.ETA - 1) high = (CurrentTurn + LocalCandidate.ETA - 1)
```
geoffthemedio, line 159: seems unrelated?

default/scripting/techs/spy/CUSTOM_ADVISORIES.focs.txt
```diff
@@ -6,4 +6,28 @@ Tech
     researchcost = 1
     researchturns = 1
     tags = [ "PEDIA_SPY_CATEGORY" ]
+    effectsgroups = [
+        [[CUSTOM_SITREPS]]
+
+        EffectsGroup      // colonies consume influence proportional to how many planets the empire controls
```
geoffthemedio, line 12: this does not fit here

geoffthemedio:
> Move influence upkeep maluses to a tech (custom advisories, in lack of a proper tech for empire upkeep/management) from species influence macros and centralization policy.

What does that have to do with the motivating problem of influence costs being too high after some number of planets?

> Make upkeep based in number of colonies use square root to make it tractable later game.

Sounds plausibly reasonable and definitely relevant to the motivationg issue.

> Make Centralization incur into penalization stronger than usual and negate number-based upkeep.

Don't undersatnd what that means.

> Make Confederation negate distance-based upkeep and supply-disconnected malus.

Didn't Confederation policy already compensate for the supply-disconnection penalty?

> Make Centralization penalize colony supply and bonus capital supply.

Why should Centralization boost capital supply, thereby making the capital more easily connected to other colonies? What does that have to do with the influence costs issue?

default/customizations/custom_sitreps.txt
Oberlus: I'm not sure why I removed the macro and used its content instead, can't remember what trouble it caused to me.
The macro wasn't used in any other place.
I will try putting it back again (in a couple of days) unless you tell me to leave it as is.

default/scripting/techs/spy/CUSTOM_ADVISORIES.focs.txt
Oberlus: You mean it makes little sense to have influence upkeep code in CUSTOM_ADVISORIES.focs.txt, right?
I fully agree.
Should I create a new tech to be the root of influence upkeep (and related) stuff?

Oberlus:
> What does that have to do with the motivating problem of influence costs being too high after some number of planets?

Due my lack of FOCS skills, I wasn't able to query the distance to capital from the influence.macros file, but I could from a tech, so I moved the number-based upkeep to a tech.

Then I wanted to make Centralization the alternative to Confederation (instead of something that you can only use during first 20-30 turns and then drop forever), so that Centralization removes the number-based upkeep, the Confederation removes the distance-based upkeep, and having none of those makes colonies pay for both types of upkeep.
The way it was before, Centralization only increased upkeep costs of all planets in exchange of a fixed pop-based bonus in the capital that is outperformed by the lack of influence and so would never be used if we manage to make Influence a must.
Empires should be able to be competitive for a while without investing in some key features (in this case influence), but not forever.


> > Make Centralization incur into penalization stronger than usual and negate number-based upkeep.
> 
> Don't undersatnd what that means.

Centralization did add extra influence upkeep (distace-based) on top of starting influence upkeep (number-based). I made it substitute it. Then I realised that wasn't enough to make Centralization a good policy late game, so I did that thing of making starting upkeep have both components, and make Centralization nullify the number-based upkeep and Confederation nullify distance-based upkeep.
That is not enough for big games. We still need extra techs to reduce the constant factor for the number-based upkeep of Confederation won't work late game. Centralization could work indefinitely as long as you can build Administration Centers (or whatever).


> Didn't Confederation policy already compensate for the supply-disconnection penalty?

Yes. That is the only thing it does currently apart from extra troops. I didn't change that part, only make it nullify distance-based upkeep.


> Why should Centralization boost capital supply, thereby making the capital more easily connected to other colonies? What does that have to do with the influence costs issue?

I talked about this in the forum. The -1 supply to capital becomes irrelevant after your first colonies (first one if close enough to homeworld, otherwise you might need an outpost or wait for certain population tech, but that's it, your empire will be connected). Also, fluff-wise, all central hubs of any sort of centralized system have more connections/logistics/lanes/bandwidth than the non-central nodes. It just makes no sense at all that the capital of the centralized empire is the one with least connections. But the important thing for me is that it is more punishing to place the malus in the colonies: that makes it harder to expand during the rest of the game. Much more interesting, IMO, than current form.

Edit: anyways, we could give the capital no bonus and keep a malus in the colonies. Better?

default/customizations/custom_sitreps.txt
geoffthemedio: The macro is maybe slightly clearer that just its expansion, but if you really want to remove it, then I don't really object. But if so, it should be a separate commit at least...

geoffthemedio:
> anyways, we could give the capital no bonus [to supply range] and keep a malus in the colonies. Better?

I think so.

default/scripting/techs/spy/CUSTOM_ADVISORIES.focs.txt
geoffthemedio: Would be better than having it here. Probably there will need to be a few other influence or similar themed techs anyway.

geoffthemedio:
> ... Centralization removes the number-based upkeep, the Confederation removes the distance-based upkeep, and having none of those makes colonies pay for both types of upkeep.

I don't think "removing number-based upkeep" really fits with "Centralization". It seems like that should be another policy related to administration or something...

Also, is it really a good idea to remove number-based upkeep entirely? Seems like it could perhaps reduce it, perhaps changing the cost calculation to be less for large numbers of planets, but not completely taking it away...

> Centralization did add extra influence upkeep (distace-based) on top of starting influence upkeep (number-based). I made it substitute it. Then I realised that wasn't enough to make Centralization a good policy late game...

Centralization is intended to be a policy that focuses on the capital at the expense of other areas. So if an empire is quite large, then it generally shouldn't make much sense to be centralized, but rather it would be useful for small empires with very few planets. Perhaps there should be a distinction between "Centralized and capital-focused but still multi-planet, with the other planets supporting the capital" and "Really anything but the capital is trash and gets dumped on", though. Centralized might work well when the whole empire has consistent focus settings, for example, as the whole empire is following the central leadership closely.

> Centralization could work indefinitely as long as you can build Administration Centers (or whatever).

That does seem weird, though... Centralization working well by having lots of regional administration? I suppose they are "Administration Centers" so could synergize with "Centralization", but it still seems a bit odd.

Oberlus:
I'll think more about this, putting it together with the other, related policies (implemented and ideas), and make a post about it in the forum.

Oberlus:
Forum thread (ongoing at this time): https://www.freeorion.org/forum/viewtopic.php?f=6&t=11920&sid=29a10bd24349362dafacd2653bafce8d

geoffthemedio:
Is this still relevant or in progress?

Oberlus:
> Is this still relevant or in progress?

Yes, still relevant indeed, and still on progress. I am quite busy for the last month and haven't find the time to finish **and** playtest it, but I hope to do it in no more than two weeks.

Oberlus:
This has been superseded by several other PRs.

PR #3296, opened 2021-02-04
Remove python wrapper for class that is not used

Cjkjvfnby:
This class is present in Python bindings but there is no way to access it.


o01eg:
@Cjkjvfnby `list_players` in `auth.py` returns list of `PlayerSetupData` to initialize quick hostless game.

Cjkjvfnby:
@o01eg , thank you for the fix. 

I'm I right that we need a C++ object because this function is called from C++?

o01eg:
Yes and it's actively used by long turn server.

PR #3298, opened 2021-02-06
Remove influence queue

Cjkjvfnby:
Remove unused code.


geoffthemedio:
Not now. Maybe later if influence is implemented finally without a queue at all.

Cjkjvfnby:
> Not now. 

I have reverted the removal of cpp code.  I searched case sensitive and missed usages.

> Maybe later if influence is implemented finally without a queue at all.

sounds like never :)

Applying policies takes one turn, so AI can track submitted policies itself. No need to put this to the interface.

PR #3299, opened 2021-02-06
Remove effect accounting

Cjkjvfnby:
Remove classes that are not accessible to the AI interface.


geoffthemedio:
This really seems like something that should be left accessible to the AI. The accounting info could provide important indication of the cause of meter modifications, including unknown causes, which the AI could notice and act on... Like an enemy planet nearby with bioterror effect, or reductions in stability due to species disliking policies...

Cjkjvfnby:
> The accounting info could provide important indication of the cause of meter modifications

It will be easy to find it via git and restore.  It was added 3 years ago. 

Aslo after 130 turns `[x for x in fo.getUniverse().effectAccounting]` does not return a single value.  So this code is broken.

geoffthemedio:
> It will be easy to find it via git and restore. It was added 3 years ago.

Nobody will remember it was there or search through all the previous git commits to check if there used to be an implementation of something. It'll just get reimplemented, or more likely forgotten entirely as a possibility.

> Aslo after 130 turns `[x for x in fo.getUniverse().effectAccounting]` does not return a single value. So this code is broken.

On AI clients, the meter accounting isn't calculated unless `updateMeterEstimates` is called while the `"effects.accounting.enabled"` option is true.

Cjkjvfnby:
> Nobody will remember it was there or search through all the previous git commits to check if there used to be an implementation of something. It'll just get reimplemented...

This code is so simple that implementation is not a problem.

> On AI clients, the meter accounting isn't calculated unless updateMeterEstimates is called while the "effects.accounting.enabled" option is true.

This is a good reason to remove it.  Too many `if` to run this code. And the set of people who knows how to enable it and set of people who can [find](https://github.com/freeorion/freeorion/pulls?q=is%3Apr+Effects+) this PR is probably the same.  

`effects.accounting.enabled` is True by default. 

This is named temporal coupling, you need to call methods in a specific order. This considered a bad practice. Here is an [example](https://github.com/freeorion/freeorion/pull/3301#issuecomment-774528612) of how to deal with it 

Following your advice, I  managed to get effects and printing the value gives me the:
```
[error] ai : commonframework.cpp:150 : TypeError: No to_python (by-value) converter found for C++ type: class boost::container::flat_map<enum MeterType,class std::vector<struct Effect::AccountingInfo,class std::allocator<struct Effect::AccountingInfo> >,struct std::less<enum MeterType>,void>
```

I am against having broken code in a project. This makes the project less attractive to developers.  So I'd like to remove it or fix it and test that it works.

geoffthemedio:
> This is named temporal coupling, you need to call methods in a specific order. This considered a bad practice.

It should be trivial to make this calculation run every turn (in the C++ if you prefer) upon receiving the update from the server, so that you don't need to call the update function separately. 

It is set up as it is, with the separate call requried to do the meter accounting calculations, because running them when they are not desired would take a bit of time that would slightly slow down AI turn order generation and would be wasted if the result isn't actually wanted.

> ```
> [error] ai : commonframework.cpp:150 : TypeError: No to_python (by-value) converter found for C++ type: class boost::container::flat_map<enum MeterType,class std::vector<struct Effect::AccountingInfo,class std::allocator<struct Effect::AccountingInfo> >,struct std::less<enum MeterType>,void>
> ```

Looks like I need to fix / update the code that exposes those classes to the Python API.

...

I don't have an easy test Python script, but something like...
````
        ///////////////////////////
        //   Effect Accounting   //
        ///////////////////////////
        py::class_<Effect::AccountingInfo>("AccountingInfo")
            .add_property("causeType",          &Effect::AccountingInfo::cause_type)
            .def_readonly("specificCause",      &Effect::AccountingInfo::specific_cause)
            .def_readonly("customLabel",        &Effect::AccountingInfo::custom_label)
            .add_property("sourceID",           &Effect::AccountingInfo::source_id)
            .add_property("meterChange",        &Effect::AccountingInfo::meter_change)
            .add_property("meterRunningTotal",  &Effect::AccountingInfo::running_meter_total)
        ;
        py::class_<std::vector<Effect::AccountingInfo>>("AccountingInfoVec")
            .def(py::vector_indexing_suite<std::vector<Effect::AccountingInfo>, true>())
        ;

        py::class_<Effect::AccountingMap::mapped_type::value_type>("MeterTypeAccountingInfoVecPair")
            .add_property("meterType",          &Effect::AccountingMap::mapped_type::value_type::first)
            .add_property("accountingInfo",     &Effect::AccountingMap::mapped_type::value_type::second)
        ;
        py::class_<Effect::AccountingMap::mapped_type>("MeterTypeAccountingInfoVecMap")
            .def(py::map_indexing_suite<Effect::AccountingMap::mapped_type, true>())
        ;

        py::class_<Effect::AccountingMap::value_type>("IntMeterTypeAccountingInfoVecMapPair")
            .add_property("targetID",           &Effect::AccountingMap::value_type::first)
            .add_property("meterAccounting",    &Effect::AccountingMap::value_type::second)
        ;
        py::class_<Effect::AccountingMap>("TargetIDAccountingMapMap")
            .def(py::map_indexing_suite<Effect::AccountingMap, true>())
        ;
````

Also, to hopefully clarify, this should be a

````
map(object_id,
    map(meterType,
        vec(AccountingInfo)
    )
)
````
where `AccountingInfo` should be a struct of several fields that indicate information about an effect that modified the indicated `meterType` meter on the object with id `object_id`, in order of application to that meter.

**Comment by Cjkjvfnby** [comment]
@geoffthemedio , I have a question about exposing Python API.  

Some objects return vectors and some python lists https://github.com/freeorion/freeorion/blob/c66fe2e94db92b3b803e041e0dceef91753496bf/server/ServerWrapper.cpp#L408

What the difference from C++ side? Is using python wrappers a performance hit or require some extra code?
From a Python side, it is a little bit more comfortable to work with python objects.

**Comment by geoffthemedio** [comment]
> What the difference from C++ side? Is using python wrappers a performance hit or require some extra code?

I don't think I wrote any of the functions that return native Python data structures from within C++, but it appears to require writing a bit of extra code to do the conversion. I doubt there's a substantial performance impact.

For that particular case, there is no existing C++ function that returns exactly what is wanted (a list of names of monster ship designs), so it's no extra work to loop over the designs and append the names into a python list vs. appending them into a C++ vector of strings that would then be returned, and would need to be exposed as a class to the Python API (if not already so exposed).

For other cases, where it's not such a simple data structure, then doing the conversion might be more complicated to set up. Like if I've got a map of int to a map of string to double, it's not immediately clear how to set up a corresponding Python data structure, or whether it would even retain all the relevant info, like ordering.

For cases where the returned data is convertible directly to a Python data structure, then it could be set up to do so, though it would likely require some converter code to be written. It looks like a lot of the code in that section is doing something like that.

**Comment by Cjkjvfnby** [comment]
And there are some issues in AI logs related to effect accounting. 

```
22:38:56.939315 {0x00000554} [error] python : warnings.py:18 : _frozen_importlib:222: RuntimeWarning: to-Python converter for struct std::pair<enum MeterType,class std::vector<struct Effect::AccountingInfo,class std::allocator<struct Effect::AccountingInfo> > > already registered; second conversion method ignored.
22:38:56.940315 {0x00000554} [error] python : warnings.py:18 : _frozen_importlib:222: RuntimeWarning: to-Python converter for struct std::pair<int const ,class boost::container::flat_map<enum MeterType,class std::vector<struct Effect::AccountingInfo,class std::allocator<struct Effect::AccountingInfo> >,struct std::less<enum MeterType>,void> > already registered; second conversion method ignored.
```

Cjkjvfnby:
@geoffthemedio , could you take a look into errors in AI console?

https://github.com/freeorion/freeorion/pull/3299#issuecomment-814392960

Vezzra:
@Cjkjvfnby, closed as what? Invalid? Superseded? Abandoned?

Cjkjvfnby:
Sorry forget about tags.
 
I'd say:  Superseded by https://github.com/freeorion/freeorion/commit/9415a21f62e335f3e7d8cd5a5b077168618d31f9

PR #3306, opened 2021-02-11
bi-annual monster revamp

agrrr3:
Updating monsters to current state of the game

[forum thread](https://www.freeorion.org/forum/viewtopic.php?p=104865#p104865)

Polished basically all monsters we have.

What i did not do: krill-eating predators, flight behaviour

Only place I changed behaviour: tree are not aggressive anymore, but obstructive

What is missing: a graphic (and maybe better naming etc) for SR_GRAV_PULSE, i reused the grav pulse bombardment weapon icon

### DIFF
--- a/default/scripting/game_rules.focs.txt
+++ b/default/scripting/game_rules.focs.txt
@@ -133,3 +133,12 @@ GameRule
     category = "MULTIPLAYER"
     type = Toggle
     default = Off
+
+GameRule
+    name = "RULE_FIRST_COMBAT_ROUND_IN_CLOSE_TARGETING_RANGE"
+    description = "RULE_FIRST_COMBAT_ROUND_IN_CLOSE_TARGETING_RANGE_DESC"
+    category = "BALANCE"
+    type = Integer
+    default = 3
+    min = 1
+    max = 20
--- a/default/scripting/monster_designs/SM_BLACK_KRAKEN.focs.txt
+++ b/default/scripting/monster_designs/SM_BLACK_KRAKEN.focs.txt
@@ -5,7 +5,7 @@ ShipDesign
     hull = "SH_BLACK_KRAKEN_BODY"
     parts = [
         "SR_SPINES"
-        "SR_SPINES"
+        "SR_TENTACLE"
         "SR_SPINES"
         "SP_CHAOS_WAVE"
         "ST_CLOAK_1"
--- a/default/scripting/ship_parts/ShortRange/SR_TENTACLE.focs.txt
+++ b/default/scripting/ship_parts/ShortRange/SR_TENTACLE.focs.txt
@@ -3,6 +3,30 @@ Part
     description = "SR_TENTACLE_DESC"
     class = ShortRange
     damage = 5
+    shots = 2
+    combatTargets = And [
+        (CombatBout > 1)
+        [[COMBAT_TARGETS_VISIBLE_ENEMY]]
+        OrderedAlternativesOf [
+           And [
+               // Attack fighters in bout 2
+               (CombatBout = 2)
+               [[COMBAT_TARGETS_VISIBLE_ENEMY]] Fighter
+           ]
+           And [
+               (CombatBout > 2) // Ships are also in reach starting from bout 3
+               [[COMBAT_TARGETS_VISIBLE_ENEMY]] [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]]
+           ]
+           And [
+               (CombatBout > 2) // if no ships, fighters and planets are fair game
+               [[COMBAT_TARGETS_VISIBLE_ENEMY]]
+               Or [
+                    Fighter
+                    [[COMBAT_TARGETS_PLANET_WITH_DEFENSE]]
+               ]
+           ]
+        ]
+    ]
     mountableSlotTypes = External
     buildcost = 30 * [[FLEET_UPKEEP_MULTIPLICATOR]] * [[SHIP_PART_COST_MULTIPLIER]]
     buildtime = 2
@@ -13,3 +37,4 @@ Part
 #include "shortrange.macros"
 
 #include "/scripting/common/upkeep.macros"
+#include "/scripting/ship_parts/targeting.macros"
--- a/default/scripting/ship_parts/ShortRange/SR_TENTACLE.focs.txt
+++ b/default/scripting/ship_parts/ShortRange/SR_TENTACLE.focs.txt
@@ -5,25 +5,26 @@ Part
     damage = 5
     shots = 2
     combatTargets = And [
-        (CombatBout > 1)
         [[COMBAT_TARGETS_VISIBLE_ENEMY]]
         OrderedAlternativesOf [
            And [
-               // Attack fighters in bout 2
-               (CombatBout = 2)
-               [[COMBAT_TARGETS_VISIBLE_ENEMY]] Fighter
-           ]
-           And [
-               (CombatBout > 2) // Ships are also in reach starting from bout 3
+               // Ships are also in reach starting from bout 3
+               (CombatBout >= NamedIntegerLookup name = "FIRST_COMBAT_ROUND_IN_CLOSE_TARGETING_RANGE")
                [[COMBAT_TARGETS_VISIBLE_ENEMY]] [[COMBAT_TARGETS_NOT_DESTROYED_SHIP]]
            ]
            And [
-               (CombatBout > 2) // if no ships, fighters and planets are fair game
+               // if no ships, fighters and planets are fair game
+               (CombatBout >= NamedIntegerLookup name = "FIRST_COMBAT_ROUND_IN_CLOSE_TARGETING_RANGE")
                [[COMBAT_TARGETS_VISIBLE_ENEMY]]
                Or [
                     Fighter
                     [[COMBAT_TARGETS_PLANET_WITH_DEFENSE]]
                ]
+           ]
+           And [
+               // Attack fighters in bout 2
+               (CombatBout = 2)
+               [[COMBAT_TARGETS_VISIBLE_ENEMY]] Fighter
            ]
         ]
     ]
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -1124,6 +1124,12 @@ Combat Rounds
 RULE_NUM_COMBAT_ROUNDS_DESC
 How many rounds of combat occur each turn.
 
+RULE_FIRST_COMBAT_ROUND_IN_CLOSE_TARGETING_RANGE
+Close Range Targeting Combat Round
+
+RULE_FIRST_COMBAT_ROUND_IN_CLOSE_TARGETING_RANGE_DESC
+The first combat round in which close range weapons are able to shoot at targets.
+
 RULE_AGGRESSIVE_SHIPS_COMBAT_VISIBLE
 Aggressive Ships Combat Visible
 
@@ -14266,12 +14272,6 @@ Krill Swarm
 FT_BAY_KRILL_DESC
 When Krill swarms are big enough, single Krill leave the swarm to attack ships the swarm encounters.
 
-SHORT_RANGE_COMBAT_ROUND
-2
-
-CLOSE_RANGE_COMBAT_ROUND
-3
-
 SR_WEAPON_0_1
 Flak Cannon
 
@@ -14344,7 +14344,7 @@ Graviton Pulsar
 SR_GRAV_PULSE_DESC
 '''Graviton pulse causing damage to hull integrity to every target caught in the blast. While still deadly to fighters if the blast is spread, it causes only minor damage to ships or planet defenses.
 
-Graviton pulse inflicts [[encyclopedia DAMAGE_TITLE]] only once in combat, in combat round [[SHORT_RANGE_COMBAT_ROUND]].'''
+Graviton pulse inflicts [[encyclopedia DAMAGE_TITLE]] only once in combat, in combat round 2.'''
 
 SR_JAWS
 Jaws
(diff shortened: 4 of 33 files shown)


agrrr3:
there is a 18ab4bbc8fd41108b6cb3e27302f14b451bfd858 in this history, not sure why the commit is suddenly 01b18055e2055cdd36671950eec61615c9deb155 in the master (@geoffthemedio did you force-push to master?), so I kick that 18ab commit out of the history and again do a rebase (actually the contents is the same and it is the current tip so there is no reason for a repush, dunno)

geoffthemedio:
> @geoffthemedio did you force-push to master?

Not intentionally or that I know of.

agrrr3:
AFAICT this is good for merge

geoffthemedio:
With this applied, I get some AI chat error messages:
`[13 Feb 13:12:11] AI_3: AI_Error: AI script error : _targets.py:get_allowed_targets():14  - AI has no targeting information for weapon part %s. Will assume any target allowed.Please update CombatTarget.PART_ALLOWED_TARGETS in AIDependencies.py`

agrrr3:
> With this applied, I get some AI chat error messages:

thanks for catching this

agrrr3:
I was playtesting this a bit and i have a freeoriond crash if i send a carrier into battle. Not sure how this is possible (enemy has only  mass driver, so my changes should not have an influence on the combat at all).

I went and build 0437322 (weekly-build) and older revisions locally. And I get the same behaviour *shrug*

edit:
So it looks like a31333b17f8b70d6abec608ce85d3dac82c5dca4 broke this, but i know geoff revert part of that in 9f6b9286d94e285cf84949801b93a12bcb604238 (so something probably missing).

I think the cause is the m_species_name reference in Fighter.cpp

default/scripting/monster_designs/SM_BLACK_KRAKEN.focs.txt
```diff
@@ -5,7 +5,7 @@ ShipDesign
     hull = "SH_BLACK_KRAKEN_BODY"
     parts = [
         "SR_SPINES"
-        "SR_SPINES"
+        "SR_TENTACLE"
```
Cjkjvfnby, line 8: If you add one tentacle by a revision and do it bi-annual you'll get 8 tentacles in 4 years and Kraken will become an octopus :)

default/scripting/ship_parts/ShortRange/SR_TENTACLE.focs.txt
```diff
@@ -3,6 +3,30 @@ Part
     description = "SR_TENTACLE_DESC"
     class = ShortRange
     damage = 5
+    shots = 2
+    combatTargets = And [
+        (CombatBout > 1)
+        [[COMBAT_TARGETS_VISIBLE_ENEMY]]
+        OrderedAlternativesOf [
+           And [
```
Cjkjvfnby, line 10: I am not familiar with scripting language and the combat process.  Could you explain to me a little about `OrderedAlternativesOf`?

According to the description, ships should be attacked first, but when CombatBout is 2 fighters is a priority.

How CombatBout is decremented?  When all ships are destroyed or unconditionally?

geoffthemedio: The OrderedAlternativesOf condition matches all objects that match the first subcondition, unless nothing matches, in which case it moves to the next, and keeps trying subconditions until something is matched. From the code comment:
````
    // OrderedAlternativesOf [ A B C ] matches all candidates which match the topmost condition.
    // If any candidate matches A, then all candidates that match A are matched,
    // or if no candidate matches A but any candidate matches B, then all candidates that match B are matched,
    // or if no candidate matches A or B but any candidate matches C, then all candidates that match C are matched.
    // If no candidate matches A, B, or C, then nothing is matched.
    //
    // Not OrderedAlternativesOf [ A B C ] finds the topmost condition which has matches and then matches its non-matches.
    // If any candidate matches A, then all candidates that do not match A are matched,
    // or if no candidates match A but any candidate matches B, then all candidates that do not match B are matched,
    // or if no candidate matches A or B but any candidate matches C, then all candidates that do not match C are matched.
    // If no candidate matches A, B, or C, then all candidates are matched.
````

Every main game turn, there is a new combat independent of any previous combats in a system, and the CombatBout is reset to 1.

Comment (agrrr3) [reply default/scripting/ship_parts/ShortRange/SR_TENTACLE.focs.txt:10]: > I am not familiar with scripting language and the combat process. Could you explain to me a little about `OrderedAlternativesOf`?

There is also an entry in [FOCS_Scripting_Details](https://www.freeorion.org/index.php/FOCS_Scripting_Details) in the wiki.
From combat bout 3 on ships are primarily targetted . Only if there are no ship enemies, tentacles attack fighters or planets.

> 
> How CombatBout is decremented?

Each battle has (usually) four combat bouts. First bout 1, second bout 2,... You can see this in the combat reports.


> 
> According to the description, ships should be attacked first, but when CombatBout is 2 fighters is a priority.

It also sais: "Tentacles start to destroy fighters when they come in range in [[SHORT_RANGE_COMBAT_ROUND]]."

I introduce here a design/fluff model of "shrinking distances". First bout is "long" distance, second bout is "short" distance, and starting from third bout it is "close". Fighters are supposed to closer to the ships (they shoot from inside the ship shields, fluff said), so in second bout the "close" distance attacking tentacle can not attack ships, but it can reach the fighters which moved already closer. Note this is not done consistently for other weapons (jaws cant attack fighters in bout two), as the tentacles can "stretch". Mostly adding flavor and some utility.

edit1: use the right quote

File: default/stringtables/en.txt [hunk 3]
```diff
@@ -14332,23 +14338,45 @@ Spinal Antimatter Cannon
...
+Spines inflict [[encyclopedia DAMAGE_TITLE]] starting from combat round [[CLOSE_RANGE_COMBAT_ROUND]] as the targets are too far away in earlier rounds.'''
 
 SR_TENTACLE
 Tentacle
 
 SR_TENTACLE_DESC
-Tentacle of a kraken, can damage ships.
+'''Tentacle of a kraken, primarily do damage to ships. Is also able to attack fighters and planets.
+
+Tentacles can inflict [[encyclopedia DAMAGE_TITLE]] to ships starting from combat round [[CLOSE_RANGE_COMBAT_ROUND]] as the targets are too far away in earlier rounds. Tentacles start to destroy fighters when they come in range in [[SHORT_RANGE_COMBAT_ROUND]]. Tentacles only attack fighters or planets if there are no ships to target.'''
```
Cjkjvfnby:
```suggestion
Tentacles can inflict [[encyclopedia DAMAGE_TITLE]] to ships, fighters, and planets. In the first combat round, they cannot reach anything and shake the tentacles in anger, at the [[SHORT_RANGE_COMBAT_ROUND]] round they can smash fighters. Later on, they can reach any targets but prefer crushing ships first.
```

I'd prefer removing [[SHORT_RANGE_COMBAT_ROUND]] and [[CLOSE_RANGE_COMBAT_ROUND]] from string tables. 
Since we can count rounds from the beginning it is ok to embed them into the translation.

short = 2 and close = 3 could be confusing to translater (as it was for me at the beginning).  Using first, second and third as names is probably a better option.

default/scripting/ship_parts/ShortRange/SR_TENTACLE.focs.txt
Cjkjvfnby: Thank you for the explanation.  I think I got an idea.  

I tried to rephrase the description in the string tables.  I try to make it a story rather than a list of facts.  In the evening my brain not good at interpreting facts.

default/stringtables/en.txt
agrrr3: I see your pain. Not completly convinced if THIRD_COMBAT_ROUND is good though. The whole point of the macro is to avoid magic numbers. What if the close range round is changed from 3 to e.g. 5? Then all the translators would need to change all occurences of THIRD_COMBAT_ROUND to FIFTH_COMBAT_ROUND

agrrr3: How about something like THIRD_WAVE_COMBAT_ROUND (at the moment 3, but could also become e.g. 5) or stressing that the contents of the macro is a number by using something like SHORT_RANGE_COMBAT_ROUND_NUMBER.
One could also make this clearer by shifting this from a macro to a value ref (or a rule) using a `[value SHORT_RANGE_COMBAT_ROUND]` expression in the stringtables.

Cjkjvfnby: > The whole point of the macro is to avoid magic numbers. What if the close range round is changed from 3 to e.g. 5? 

It is hardcoded in the script: `(CombatBout > 1)`   Also that means that close range for tentacles is not the same close range for Jaws.    

Terms *RANGE_COMBAT and *WAVE_COMBAT_ROUND are not a part of focs-script language (at least of parts we are talking about).  So adding new terms with new meanings may be confusing to the translations.   Actually, I don't know how to communicate this information to them. This example is not enough to share an idea behind waves: `THIRD_WAVE_COMBAT_ROUND = 5`.

I totally agree that a new layer of abstractions could make an update of the translation much easier, but we cannot guarantee that it will be maintained properly.  And keep in mind having them in a translation and not having them in focs-scripts make things even more complicated than using combat rounds in the text.

> using a [value SHORT_RANGE_COMBAT_ROUND]

Can we use this ref in focs-scripts?

agrrr3: > > The whole point of the macro is to avoid magic numbers. What if the close range round is changed from 3 to e.g. 5?
> 
> It is hardcoded in the script: `(CombatBout > 1)` Also that means that close range for tentacles is not the same close range for Jaws.

well. as i said fighter fluff sais those are really close/able to penetrate shields starting in bout 2 (while all other weapons are not able to do that). so those two weapons are consistent (jaws are not targeting fighters at all, while tentacles and  only target fighters in bout 2). Spines are like jaws. SR_THRASHING_BODY is not consistent as it also targets fighters, but only starting from bout 2. I could fluff that away or make it the same complex attack pattern like tentacles.

> 
> Terms *RANGE_COMBAT and *WAVE_COMBAT_ROUND are not a part of focs-script language (at least of parts we are talking about). So adding new terms with new meanings may be confusing to the translations. Actually, I don't know how to communicate this information to them. This example is not enough to share an idea behind waves: `THIRD_WAVE_COMBAT_ROUND = 5`.

The right place to talk about complex topics would be in an encyclopedia entry. The closest we have currently is default/scripting/encyclopedia/game_concepts/ship_management/DAMAGE.focs.txt i guess, so the DAMAGE_TITLE and DAMAGE_TEXT, but we could write another entry about combat/targeting/ranges/waves.

Note that FOCS also has no notion of colony/outpost. One is a planet with some population, the other without. So I expect more knowledge from somebody who is coding FOCS.

> 
> And keep in mind having them in a translation and not having them in focs-scripts make things even more complicated than using combat rounds in the text.

wasnt sure if i should introduce that into FOCS; could be done by a macro, a rule or a named value ref. Or some combination of rule and macro/named value ref.
The almost-zero-overhead version is the macro variant.

@geoffthemedio is it ok for you if i introduce a rule for defining the first combat round in which the targets come into range? like RULE_FIRST_COMBAT_ROUND_IN_CLOSE_TARGETING_RANGE and RULE_FIRST_COMBAT_ROUND_IN_SHORT_TARGETING_RANGE. The upside of having this as game rules is that you can adjust it if you adjust the RULE_NUM_COMBAT_ROUNDS.

Adjusting the number of combat rounds strongly shifts the damage output (three bouts is a strong nerf for unhidden close range weapons, while making hidden ones "invincible"; more than four bouts means close range weapons get a boost). The current implementation follows the idea that we mainly stick with four bouts as was discussed in the forums. If we have the game rules one can set e.g. 8 combat rounds, short targeting start in bout 3 and close targeting in bout 5 and the overall damage would be restored. If you want to compensate first-strike advantage you could make it 1/2-3/4-7 instead of 1-2/3-4/5-8.

Something like RULE_COMBAT_ROUND_WHEN_SECOND_ATTACK_WAVE_STARTS would also be possible but I think the wave metaphor is worse than the targeting range one.

> 
> > using a [value SHORT_RANGE_COMBAT_ROUND]
> 
> Can we use this ref in focs-scripts?

Yes. Actually this can only be defined in the focs-scripts; it can be read from stringtables.

Cjkjvfnby:
> Adjusting the number of combat rounds 

Making things configurable has one huge drawback.  It takes time and resources to implement and support. And it pays off only if you did it correctly.

At this point, I'd suggest hardcoding these combat rounds (both in text and focs). This will make things simpler.  And later on when/if we decide to change how combat works, we can just manually fix all weapon scripts and translation entries.  At least we will have a better understanding of why we need to do it. Actually, I think that 4 rounds and existing focus features give a lot of flexibility.

agrrr3:
> > Adjusting the number of combat rounds
> 
> Making things configurable has one huge drawback. It takes time and resources to implement and support. And it pays off only if you did it correctly.

Adjustable combat rounds is in the game since forever (and it doesnt play nicely with "ranges"). Adding the values for ranges takes me about ten minutes I think. I just dont know if geoff wants it or not.

> At this point, I'd suggest hardcoding these combat rounds (both in text and focs). 

so basically you take back your suggestion and you agree the PR can be merged in its current state?

geoffthemedio:
> @geoffthemedio is it ok for you if i introduce a rule for defining the first combat round in which the targets come into range?

I don't really care / have an opinion about it. Keep the rule defined in FOCS, not C++, though.

agrrr3:
> > @geoffthemedio is it ok for you if i introduce a rule for defining the first combat round in which the targets come into range?
> 
> I don't really care / have an opinion about it. Keep the rule defined in FOCS, not C++, though.

Cool, done

One upside of using the value refs in stringtables is that you can see in the encyclopedia that you can adjust the value using a game rule.

default/scripting/game_rules.focs.txt
```diff
@@ -133,3 +133,12 @@ GameRule
     category = "MULTIPLAYER"
     type = Toggle
     default = Off
+
+GameRule
```
Cjkjvfnby, line 137: I see one game rule, should we have 2 of them?

Cjkjvfnby:
> so basically you take back your suggestion and you agree the PR can be merged in its current state?

My suggestion was to remove `*COMBAT_ROUND*` entries and just hardcode values in the translation.  Anyway, I am not blocking the merge, it's just a suggestion.

default/scripting/game_rules.focs.txt
agrrr3: Actually i am not using short range targeting yet, (the other targeting of tentacle depends on when fighters launch, which is not scriptable in focs). If there is an instance of shortrange weapon i will add it.

Cjkjvfnby: Please remove string tables entries: RULE_FIRST_COMBAT_ROUND_IN_SHORT_TARGETING_RANGE*

agrrr3: done

geoffthemedio:
is this done?

agrrr3:
> is this done?

Yes.

PR #3309, opened 2021-02-14
Parse techs with python parser

o01eg:
Follows-up #3265


o01eg:
@geoffthemedio I tried to test techs parsing but catch segfault with:

```
#0  0x00005555555d8669 in CurrentTurn () at /mnt/another/srcs/GIT/freeorion/universe/../Empire/../util/AppInterface.h:145
#1  ScriptingContext::ScriptingContext (this=0x7ffff34c3160) at /mnt/another/srcs/GIT/freeorion/universe/ScriptingContext.h:18
#2  0x000055555563ce59 in ValueRef::Operation<double>::CacheConstValue (this=this@entry=0x7fffec3b3ba0) at /mnt/another/srcs/GIT/freeorion/parse/../universe/ValueRefs.h:1736
#3  0x0000555555640484 in ValueRef::Operation<double>::Operation (this=0x7fffec3b3ba0, op_type=<optimized out>, operand1=..., operand2=...) at /mnt/another/srcs/GIT/freeorion/parse/../universe/ValueRefs.h:1661
#4  0x000055555564211d in boost::phoenix::new_eval::operator()<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<boost::phoenix::detail::target<ValueRef::Operation<double> > >, 0l>, boost::phoenix::actor<boost::spirit::local_variable<2> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::phoenix::detail::tag::function_eval, boost::proto::argsns_::list3<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<parse::detail::deconstruct_movable>, 0l>, boost::phoenix::actor<boost::spirit::local_variable<0> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<boost::phoenix::argument<3> >, 0l> > >, 3l> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::phoenix::detail::tag::function_eval, boost::proto::argsns_::list3<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<parse::detail::deconstruct_movable>, 0l>, boost::phoenix::actor<boost::spirit::argument<0> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<boost::phoenix::argument<3> >, 0l> > >, 3l> >, boost::phoenix::vector2<boost::phoenix::vector4<boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::assign, boost::proto::argsns_::list2<boost::phoenix::actor<boost::spirit::local_variable<1> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::phoenix::detail::tag::function_eval, boost::proto::argsns_::list2<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<parse::detail::construct_movable>, 0l>, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::phoenix::tag::new_, boost::proto::argsns_::list4<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<boost::phoenix::detail::target<ValueRef::Operation<double> > >, 0l>, boost::phoenix::actor<boost::spirit::local_variable<2> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::phoenix::detail::tag::function_eval, boost::proto::argsns_::list3<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<parse::detail::deconstruct_movable>, 0l>, boost::phoenix::actor<boost::spirit::local_variable<0> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<boost::phoenix::argument<3> >, 0l> > >, 3l> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::phoenix::detail::tag::function_eval, boost::proto::argsns_::list3<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<parse::detail::deconstruct_movable>, 0l>, boost::phoenix::actor<boost::spirit::argument<0> >, boost::phoenix::actor<boost::proto::exprns_::basic_expr<boost::proto::tagns_::tag::terminal, boost::proto::argsns_::term<boost::phoenix::argument<3> >, 0l> > >, 3l> > >, 4l> > >, 2l> > >, 2l> > const*, boost::fusion::vector<parse::detail::MovableEnvelope<ValueRef::ValueRef<double> >&>&, boost::spirit::context<boost::fusion::cons<parse::detail::MovableEnvelope<ValueRef::ValueRef<double> >&, boost::fusion::nil_>, boost::fusion::vector<parse::detail::MovableEnvelope<ValueRef::ValueRef<double> >, parse::detail::MovableEnvelope<ValueRef::ValueRef<double> >, ValueRef::OpType, std::vector<parse::detail::MovableEnvelope<ValueRef::ValueRef<double> >, std::allocator<parse::detail::MovableEnvelope<ValueRef::ValueRef<double> > > > > >&, bool&>&, boost::phoenix::default_actions const&> > (ctx=..., a3=..., a2=..., a1=..., this=<optimized out>)
    at /usr/include/boost/phoenix/object/detail/cpp03/preprocessed/new_eval_10.hpp:101
...
#438 _ZN5boost6spirit2qi12phrase_parseINS0_3lex7lexertl8iteratorINS4_7functorINS4_14position_tokenIN9__gnu_cxx17__normal_iteratorIPKcNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEENS_3mpl6vectorIbidSH_N4mpl_2naESM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_EENSL_5bool_ILb1EEEmEENS4_6detail4dataESI_SP_SP_EEEENS_5proto7exprns_4exprINSV_6tagns_3tag8terminalENSV_7argsns_4termINS1_25parameterized_nonterminalINS1_4ruleISU_NS1_16in_state_skipperINS3_6detail10lexer_def_INS3_5lexerINS4_11actor_lexerISQ_SI_ST_EEEEEESB_EEFvRNS_11multi_index21multi_index_containerISt10unique_ptrI4TechSt14default_deleteIS1H_EENS1E_10indexed_byINS1E_18ordered_non_uniqueINS1E_3tagIN11TechManager13CategoryIndexESM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_EENS1E_13const_mem_funIS1H_RKSH_XadL_ZNKS1H_8CategoryB5cxx11EvEEEESM_EENS1E_14ordered_uniqueINS1N_INS1O_9NameIndexESM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_EENS1R_IS1H_S1T_XadL_ZNKS1H_4NameB5cxx11EvEEEESM_EESM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_SM_EESaIS1K_EEEENS0_11unused_typeES26_EENS_6fusion6vectorIJNS_7phoenix5actorINSW_10basic_exprIS10_NS12_INS_17reference_wrapperIS23_EEEELl0EEEEEEEEEEEELl0EEENSX_INSZ_9subscriptENS11_5list2IRNSX_IS10_NS12_INS0_11terminal_exINS0_3tag8in_stateENS29_IJRA3_SA_EEEEEEELl0EEERKS1C_EELl2EEEEEbRT_S32_RKT0_RKT1_NS1_9skip_flagE (post_skip=boost::spirit::qi::skip_flag::postskip, skipper=..., expr=..., last=..., 
    first=...) at /usr/include/boost/spirit/home/qi/parse.hpp:120
--Type <RET> for more, q to quit, c to continue without paging--
#439 parse::detail::parse_file<(anonymous namespace)::grammar, boost::multi_index::multi_index_container<std::unique_ptr<Tech>, boost::multi_index::indexed_by<boost::multi_index::ordered_non_unique<boost::multi_index::tag<TechManager::CategoryIndex>, boost::multi_index::const_mem_fun<Tech, std::__cxx11::basic_string<char> const&, &Tech::Category> >, boost::multi_index::ordered_unique<boost::multi_index::tag<TechManager::NameIndex>, boost::multi_index::const_mem_fun<Tech, std::__cxx11::basic_string<char> const&, &Tech::Name> > > > > (lexer=..., path=..., arg1=...) at /mnt/another/srcs/GIT/freeorion/parse/ParseImpl.h:187
#440 0x00005555558cc78e in parse::techs<std::tuple<boost::multi_index::multi_index_container<std::unique_ptr<Tech, std::default_delete<Tech> >, boost::multi_index::indexed_by<boost::multi_index::ordered_non_unique<boost::multi_index::tag<TechManager::CategoryIndex, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, boost::multi_index::const_mem_fun<Tech, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, &(Tech::Category[abi:cxx11]() const)>, mpl_::na>, boost::multi_index::ordered_unique<boost::multi_index::tag<TechManager::NameIndex, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, boost::multi_index::const_mem_fun<Tech, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, &(Tech::Name[abi:cxx11]() const)>, mpl_::na>, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na, mpl_::na>, std::allocator<std::unique_ptr<Tech, std::default_delete<Tech> > > >, std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::unique_ptr<TechCategory, std::default_delete<TechCategory> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::unique_ptr<TechCategory, std::default_delete<TechCategory> > > > >, std::set<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > > >(boost::filesystem::path const&) (path=...)
    at /mnt/another/srcs/GIT/freeorion/parse/TechsParser.cpp:249
...
```

Look like it somehow tried to get current turn on parser stage. Is it intended and I should initialize `IApp` to run parser?

geoffthemedio:
It is intended that `ScriptingContext` initializes itself by getting references to gamestate objects and info (like the current turn) when it is constructed, which will fail if there is no `IApp` initialized.

Is initializing `IApp` a problem?

If yes, another way to initialize `ScriptingContext` with other dummy values in it could be added that doesn't require looking up stuff via `IApp`...

o01eg:
@geoffthemedio No, it is not a problem. I just asked if I understand it correctly.

test/parse/TestPythonParser.cpp
```diff
@@ -0,0 +1,101 @@
...
+    Pending::Pending<TechManager::TechParseTuple> techs_p = Pending::StartParsing(parse::techs<TechManager::TechParseTuple>, m_scrpiting_dir / "techs");
+    auto techs_tuple = *Pending::WaitForPendingUnlocked(std::move(techs_p));
+    auto techs = std::get<0>(std::move(techs_tuple));
+    auto tech_categories = std::get<1>(std::move(techs_tuple));
+    auto categories_seen = std::get<2>(std::move(techs_tuple));
+    BOOST_REQUIRE(!techs.empty());
+    BOOST_REQUIRE(!tech_categories.empty());
+    BOOST_REQUIRE(!categories_seen.empty());
+
+    for (const auto& cat : {"PRODUCTION_CATEGORY", "GROWTH_CATEGORY"}) {
```
Cjkjvfnby: It is preferable to avoid any cycles in the tests, tests should be simple and linear.  
  I don't know if you get a good message if that part is failed if it will be clear which `cat` is a culprit, it's OK for me to leave it as is.

test/parse/TestPythonParser.cpp
```diff
@@ -0,0 +1,101 @@
...
+        BOOST_REQUIRE_EQUAL(1, categories_seen.count(cat));
+        BOOST_REQUIRE_EQUAL(1, tech_categories.count(cat));
+    }
+
+    const auto cat_it = tech_categories.find("CONSTRUCTION_CATEGORY");
+    BOOST_REQUIRE(tech_categories.end() != cat_it);
```
Cjkjvfnby, line 40: What is your intention about this line?  Can it be broken just by rearranging entries in the python file?

test/parse/TestPythonParser.cpp
```diff
@@ -0,0 +1,101 @@
...
+        BOOST_REQUIRE_EQUAL("", tech_effect_gr->StackingGroup());
+        BOOST_REQUIRE_EQUAL("", tech_effect_gr->GetDescription());
+        BOOST_REQUIRE_EQUAL("", tech_effect_gr->AccountingLabel());
+        BOOST_REQUIRE_EQUAL(17, tech_effect_gr->Priority());
+
+        BOOST_REQUIRE_EQUAL(1, (*tech_it)->Prerequisites().size());
+        BOOST_REQUIRE_EQUAL(1, (*tech_it)->Prerequisites().count("PRO_MICROGRAV_MAN"));
+    }
+
+    // test it last
```
Cjkjvfnby, line 191: In general, it is a bad practice to do multiple tests at once. 

In Python, I can solve it with fixtures:  create a session scope fixture (execute once, reuse in multiple tests).
Quick googling shows that it is possible to create only test level fixtures. 

You mention that the parser is fast, so probably we can run it on each test.

test/parse/TestPythonParser.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "util/Pending.h"
+
+#include "ParserAppFixture.h"
+
+BOOST_FIXTURE_TEST_SUITE(TestPythonParser, ParserAppFixture)
+
+BOOST_AUTO_TEST_CASE(parse_game_rules) {
+    PythonParser parser(m_python);
+
+    Pending::Pending<GameRules> game_rules_p = Pending::Parsed(parse::game_rules, parser, m_scrpiting_dir / "game_rules.focs.py");
```
Cjkjvfnby: Actually, I'd like to have two test sets, grouped by change reason:

- First will parse the real game files and assert nothing.  This test set will check if a real game file is syntactically correct.
  You don't want to change the test just because someone adds a new rule or rename it.
- The second set will have its own test files (I'd recommend having different files for the different tests).  You can add any case you want, you can put self-explanatory values inside, you can have erroneous testcases. And the only reason to change these files is the parser update.

test/parse/TestPythonParser.cpp
o01eg: No, it only checks if `CONSTRUCTION_CATEGORY` present in categories.

test/parse/TestPythonParser.cpp
o01eg: These tests are for old parser and it took 18s to parse techs. I will switch to python parser after I write enough test to cover all accessible fields.

test/parse/TestPythonParser.cpp
o01eg: We have first test set which parse real game files in system tests where the client and the server call all parsers.

Second sounds reasonable, so we need to have to copy some subset of default to test and use it for tests.

test/parse/TestPythonParser.cpp
o01eg: Made in linear.

test/parse/TestPythonParser.cpp
Cjkjvfnby: I'd prefer not copying, but creating from scratch.  There is no reason to make it look like a game object, it should represent a minimum field required for tests.  It is also a good practice to have an independent fixture for each tests, but for the first implementation, it is not an option.

geoffthemedio:
@o01eg can you tell if a similar issue with `IApp` and `ScriptingContext` is causing the SmokeTestGame tests to fail / not return anything after this commit: https://github.com/freeorion/freeorion/commit/738b0c8c1be7a3a2f198e44552036623f6aba129
eg: https://travis-ci.com/github/freeorion/freeorion/jobs/484076599#L1503

o01eg:
@geoffthemedio No, it's a some strange freeze. I got segfault because application singleton was NULL pointer but smoke test have it initialized.

o01eg:
@geoffthemedio I built `tweak_test` with debug symbols and got segault 
```
p IApp::s_app 
$2 = (IApp *) 0x0
(gdb) bt
#0  0x00007ffff7167f09 in CurrentTurn () at /mnt/another/srcs/GIT/freeorion/parse/../universe/../Empire/../util/AppInterface.h:143
#1  ScriptingContext::ScriptingContext (this=0x7ffff1ce2fe0) at /mnt/another/srcs/GIT/freeorion/parse/../universe/ScriptingContext.h:18
#2  0x00007ffff7168279 in ValueRef::Operation<double>::CacheConstValue (this=this@entry=0x7fffcc3fc960) at /mnt/another/srcs/GIT/freeorion/parse/../universe/ValueRefs.h:1736
#3  0x00007ffff716b7fd in ValueRef::Operation<double>::Operation (this=0x7fffcc3fc960, op_type=<optimized out>, operand=...) at /mnt/another/srcs/GIT/freeorion/parse/../universe/ValueRefs.h:1672
...
#333 parse::detail::parse_file<(anonymous namespace)::grammar, std::map<std::__cxx11::basic_string<char>, std::unique_ptr<Species> > > (arg1=std::map with 31 elements = {...}, path=..., lexer=...)
    at /mnt/another/srcs/GIT/freeorion/parse/ParseImpl.h:187
#334 parse::species[abi:cxx11](boost::filesystem::path const&) (path=...) at /mnt/another/srcs/GIT/freeorion/parse/SpeciesParser.cpp:404
```

Could you open PR to discuss it?

geoffthemedio:
Here you go: https://github.com/freeorion/freeorion/pull/3313

test/parse/TestPythonParser.cpp
o01eg: I've tried it to do but got issue where to place test resources. Tests binaries are generated in `CMAKE_BINARY_DIR` and `default` folder is linked here from `CMAKE_SOURCE_DIR` but I not sure if it good idea to place it here.
Maybe I should create `default-test` folder and then link it before call `unittest` target or after test binaries built.

o01eg:
I've tried to get access to internal structure of parsed techs but it looks like it doesn't works. `ValueRef` turn and cost fields are private and inaccessible outside, effect scope `Condition` doesn't returns it's operands.

https://github.com/freeorion/freeorion/blob/2b7471820b72de49fca9700f1047fe0f1bdbb0e4/universe/Conditions.cpp#L9929-L9935

It doesn't fill `retval` variable so I cannot check tree deeper.

Maybe I should just call `Dump` and compare strings?

geoffthemedio:
> `ValueRef` turn and cost fields are private and inaccessible outside

should be simple to add necessary getters

Cjkjvfnby:
> > ValueRef turn and cost fields are private and inaccessible outside
   
> should be simple to add necessary getters

Unittesting should test public API.  If a private variable does not affect any call of public API probably we can remove it.  

Adding getters is not a good idea because you extend that API just for the reason of testing.  So it will be harder to change internal implementation since they are exposed. This leads to fragile tests and increases the number of actions required to change something.

The idea of unittests is to check that unit does what you want, not check how it does it.

geoffthemedio:
> > should be simple to add necessary getters
> 
> Unittesting should test public API. If a private variable does not affect any call of public API probably we can remove it.
>
> The idea of unittests is to check that unit does what you want, not check how it does it.

As [suggested here](https://github.com/freeorion/freeorion/pull/3332#issuecomment-787043520), a reason to implement the tests by providing `bool Tech::operator==(const Tech&) const` and using it to compare results.

probably also need a few fixes like https://github.com/freeorion/freeorion/commit/e92683333246fb357455c76a281d800e8b8f2b8b

edit: and several other `operator==` are broken due to just checking for equality of pointer members

edit2: fixed `Operation` and `Statistic`, hopefully. Don't see any other broken ValueRef `operator==`

Cjkjvfnby:
> As suggested here, a reason to implement the tests by providing bool Tech::operator==(const Tech&) const and using it to compare results.

You can provide a function that gets two Techs and do the comparison. You don't need to extend Tech API for that. 
The private variable should somehow affect the return value.  

I have checked the Tech API, It requires some refactoring to be testable.

geoffthemedio:
> > As suggested here, a reason to implement the tests by providing bool Tech::operator==(const Tech&) const and using it to compare results.
> 
> You can provide a function that gets two Techs and do the comparison.

That's what `operator==` does... it compares an object, such as a tech, to another object and returns true if they are equal by whatever criteria are relevant for those objects.

> You don't need to extend Tech API for that. The private variable should somehow affect the return value.

I'm not sure what you have in mind, then... You objected to adding the getters for the internal `ValueRef`s, and you object to adding `operator==` that would do the comparison within a class member function and thus avoid the need to make the `ValueRef`s publically accessible with getters. A free function that compares two techs could be written, but that's just another way to implement `operator==` and is awkward / more complicated to write for classes with private data. There are also `GetCheckSum` functions, but those aren't guaranteed to produce unique results for different `Tech` objects.

> I have checked the Tech API, It requires some refactoring to be testable.

Such as?

Cjkjvfnby:
> I'm not sure what you have in mind, then... 

I read a bit more about `operator==`  and now I understand how it works. I am happy with that approach.

> Such as?

It's is not required for this task, because it is solved with oeprator==

To write a unittest for a Tech methods like `float Tech::ResearchCost(int empire_id)`  you need to mock 2 things: 

```cpp
GetGameRules().Get<bool>("RULE_CHEAP_AND_FAST_TECH_RESEARCH")
Empires().GetSource(empire_id);   // TODO: pass ScriptingContext and use here
```

For example, you could add 2 booleans to the constructor.  Instead of boolean, you can pass some "callable objects".

When you write a unittest you don't want to create a GetGameRules() and configure its values. You have to mock it. 
If you inject this value in the constructor you don't need any tools and can create an object manually. 

I don't have a good example but have a bad one.  Instead of passing factories to the constructor, I mock all of them.
https://github.com/freeorion/freeorion/blob/323f689df6014abb9294da34bc2b350465bb2a31/default/python/tests/AI/test_debug_chat.py#L62

geoffthemedio:
Looks like implementing `Tech::operator==` fully requires implementing some other `operator==` including `EffectsGroup`, `Effect`, and various `Effect`-derived classes. I've done some of that... https://github.com/freeorion/freeorion/commits/Tech_operator_equals

> ```c++
> GetGameRules().Get<bool>("RULE_CHEAP_AND_FAST_TECH_RESEARCH")
> Empires().GetSource(empire_id);   // TODO: pass ScriptingContext and use here
> ```
> 
> For example, you could add 2 booleans to the constructor. Instead of boolean, you can pass some "callable objects".

What do you mean by "the constructor"?

> When you write a unittest you don't want to create a GetGameRules() and configure its values. You have to mock it.

Depends what / how you're unit testing. If you want to test that a `Condition` or `ValueRef` returns the right value, then you'd need to evaluate them, and would need to pass a `ScriptingContext` with a known and relevant contents, depending on exactly what calculation is being tested.

If you instead want to test the parsers, which is what @o01eg is doing, then you don't need to evalulate what you've generated from the parser, except in a few cases where there is an optimization for constant values that get evaluated when being constructed, which I think is what's leading to the extra stuff @o01eg had to set up for these tests.

Those cases are specifically for constant results, which don't depend on the input `ScriptingContext` at all. But they are still evaluated using the same common interface that expects a `ScriptingContext`, so it just passes a default-constructed `ScriptingContext`. To construct that, it collects references to various other structures that the app interface should provide, like the game rules or empires container, even though it doesn't really need them to evaluate a constant valueref expression.

If this is really a problem, then `ScriptingContext` could be given a way to set up a version of itself that doesn't depend on any app interface, or a separate set of `ConstExprEvalImpl` functions could be implemented in `ValueRef::Operation` to avoid any calls that use a `ScriptingContext`. But @o01eg said above that setting up the app interface wasn't really a problem, so I didn't do either of these.

Also, to be clear, the example you posted a bit of has a comment about `// TODO: pass ScriptingContext and use here`. That's a reference to what I've been working on recently, involving replacing calls to global getters like `GetGameRules()` and `Empires()` with accesses through a passed in `ScriptingContext`. Until that's done, there's no real way to avoid implementing the app interface so that those global getters have something to get.

> If you inject this value in the constructor you don't need any tools and can create an object manually.

So you propose something like storing a `ScriptingContext` in every effect or condition or valueref, or possibly higher level stuff like techs, to be used in some situation instead of taking one as a parameter when evaluating valuerefs or conditions?

Or maybe passing a `ScriptingContext` to the constructors of all those classes to use if it needs to evaluate anything during construction? That could work, but would require a lot of modifiations of the consturctors to pass along the `ScriptingContext`, which also might be tricky to get working right with the parsers that actually do the construction from nested-to-outer, not outer-to-nested. You'd also still need to initialize everything to assemble a `ScriptingContext` in that case, so why not do it in the same place that it usually happens in the app interface so the existing code all just works?

> I don't have a good example but have a bad one. Instead of passing factories to the constructor, I mock all of them.
> 
> https://github.com/freeorion/freeorion/blob/323f689df6014abb9294da34bc2b350465bb2a31/default/python/tests/AI/test_debug_chat.py#L62

I can't tell what you're referring to in that code.

o01eg:
Looks like usage of `std::unique_ptr` in `ValueRef` makes it literally impossible to wrap it for boost python code https://github.com/boostorg/python/issues/202

geoffthemedio:
> Looks like usage of `std::unique_ptr` in `ValueRef` makes it literally impossible to wrap it for boost python code [boostorg/python#202](https://github.com/boostorg/python/issues/202)

Can you point me to the bit of code that doesn't work? I don't know what it means to "wrap it for boost python code".

o01eg:
@geoffthemedio I cannot move out `std::unique_pr<ValueRef::ValueRef<T>>` from python. It requires copy-able to return value from function or reference but I need to own it.

Maybe we could add `virtual std::uniqur_ptr<ValueRef::ValueRer<T>> ValueRef::ValueRer::copy() const` method so I could copy out this values?

Edit: Same for `Conditions` too.

geoffthemedio:
> @geoffthemedio I cannot move out `std::unique_pr<ValueRef::ValueRef<T>>` from python. It requires copy-able to return value from function or reference but I need to own it.

Do you need to? Can you store with raw pointers, and move into a `unique_ptr` in C++?

> Maybe we could add `virtual std::uniqur_ptr<ValueRef::ValueRer<T>> ValueRef::ValueRer::copy() const` method so I could copy out this values?

Can be done, though is a bit of a hassle to set up for all the needed cases... Something like
````
 universe/Condition.h    |  4 ++++
 universe/Conditions.cpp | 33 +++++++++++++++++++++++++++++++++
 universe/Conditions.h   |  5 +++++
 3 files changed, 42 insertions(+)

diff --git a/universe/Condition.h b/universe/Condition.h
index 20a79dcc5..12611dc7f 100644
--- a/universe/Condition.h
+++ b/universe/Condition.h
@@ -29,6 +29,8 @@ struct FO_COMMON_API Condition {
     Condition() = default;
     virtual ~Condition();
 
+    [[nodiscard]] virtual std::unique_ptr<Condition> Clone() const = 0;
+
     virtual bool operator==(const Condition& rhs) const;
     bool operator!=(const Condition& rhs) const
     { return !(*this == rhs); }
@@ -94,6 +96,8 @@ struct FO_COMMON_API Condition {
     { return 0; }
 
 protected:
+    virtual void Copy(const Condition& rhs);
+
     bool m_root_candidate_invariant = false;
     bool m_target_invariant = false;
     bool m_source_invariant = false;
diff --git a/universe/Conditions.cpp b/universe/Conditions.cpp
index d5a7a2c95..d2b839a64 100644
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -256,6 +256,14 @@ struct Condition::MatchHelper {
 
 Condition::~Condition() = default;
 
+void Condition::Copy(const Condition& rhs) {
+    if (&rhs == this)
+        return;
+    m_root_candidate_invariant = rhs.m_root_candidate_invariant;
+    m_target_invariant = rhs.m_target_invariant;
+    m_source_invariant = rhs.m_source_invariant;
+}
+
 bool Condition::operator==(const Condition& rhs) const {
     if (this == &rhs)
         return true;
@@ -392,6 +400,31 @@ Number::Number(std::unique_ptr<ValueRef::ValueRef<int>>&& low,
         boost::algorithm::all_of(operands, [](auto& e){ return !e || e->SourceInvariant(); });
 }
 
+std::unique_ptr<Condition> Number::Clone() const {
+    auto retval = std::make_unique<Number>();
+    retval->Copy(*this);
+    return retval;
+}
+
+void Number::Copy(const Condition& rhs) {
+    if (&rhs == this)
+        return;
+    if (typeid(*this) != typeid(rhs))
+        return;
+
+    Condition::Copy(rhs);
+    const Number& rhs_ = static_cast<const Number&>(rhs);
+
+    if (rhs_.m_low) {
+        //m_low = rhs_.m_low->Clone();
+    }
+    if (rhs_.m_high) {
+        //m_high = rhs_.m_high->Clone();
+    }
+    if (rhs_.m_condition)
+        m_condition = rhs_.m_condition->Clone();
+}
+
 bool Number::operator==(const Condition& rhs) const {
     if (this == &rhs)
         return true;
diff --git a/universe/Conditions.h b/universe/Conditions.h
index 740a5bc76..91723383b 100644
--- a/universe/Conditions.h
+++ b/universe/Conditions.h
@@ -76,7 +76,9 @@ struct FO_COMMON_API Number final : public Condition {
            std::unique_ptr<ValueRef::ValueRef<int>>&& high,
            std::unique_ptr<Condition>&& condition);
 
+    [[nodiscard]] std::unique_ptr<Condition> Clone() const override;
     bool operator==(const Condition& rhs) const override;
+
     void Eval(const ScriptingContext& parent_context, ObjectSet& matches,
               ObjectSet& non_matches, SearchDomain search_domain = SearchDomain::NON_MATCHES) const override;
     std::string Description(bool negated = false) const override;
@@ -85,6 +87,9 @@ struct FO_COMMON_API Number final : public Condition {
     unsigned int GetCheckSum() const override;
 
 private:
+    Number() = default;
+    void Copy(const Condition& rhs) override;
+
     bool Match(const ScriptingContext& local_context) const override;
 
     std::unique_ptr<ValueRef::ValueRef<int>> m_low;
````

**Comment by o01eg** [comment]
> Do you need to? Can you store with raw pointers, and move into a `unique_ptr` in C++?

It possible but I not sure if it could be ensured python doesn't store pointer somewhere else and tries to use if even after destruction. I was suggested to use pybind11 but it also doesn't support std::unique_ptr.

**Comment by o01eg** [comment]
#3345 approach works for me here.

**Comment by o01eg** [comment]
Finished minimal techs parser for two techs.

All other techs aren't converted but it shows how Python parser looks like.

At least it passed parser tests locally.

**Comment by o01eg** [comment]
I have implemented python parsing tech in parallel with old parser. Next I'll convert `Self Defense` tech which uses complex macros from `mines.macro`.

**Comment by o01eg** [comment]
@geoffthemedio  How `Planet` token parsed in:

```python
        EffectsGroup
            scope = And [
                Planet
                OwnedBy empire = Source.Owner
            ]
            stackinggroup = "PLANET_SHIELDS_STACK_ROOT"
            effects = SetMaxShield value = Value + 1 accountinglabel = "DEF_ROOT_DEFENSE"
```

I cannot find appropriate code for it in Conditions.

geoffthemedio:
> @geoffthemedio How `Planet` token parsed in:

It's an object `Type` condition...

https://github.com/freeorion/freeorion/blob/master/parse/ConditionParser6.cpp#L110
https://github.com/freeorion/freeorion/blob/master/parse/UniverseObjectTypeValueRefParser.cpp#L21

o01eg:
@geoffthemedio In `EffectParser.cpp` that line means default priority is 100?

```cpp
            > ((label(tok.priority_)         > tok.int_ [ _e = _1 ]) | eps [ _e = 100 ])
```

o01eg:
@geoffthemedio Also in `Defense.focs.txt` what means and how to parsed

```
Value(Target.MaxShield)
```

geoffthemedio:
I don't think the presence or absence of `boost::none` should depend on whether it is `Value` wrapped... Otherwise yes.

o01eg:
OK, I'm struck on https://stackoverflow.com/questions/19772058/boost-python-global-and-local-dictionary issue.

`Defense.focs.py` sees and can import `SYSTEM_MINES_SCOPE` and `EG_SYSTEM_MINES` but when `EG_SYSTEM_MINES` is called it doesn't see `SYSTEM_MINES_SCOPE`.

o01eg:
@geoffthemedio  As I see `GenerateSitRepMessage` accepts `ValueRef<std::string>` in parameters but `Target` fields here are `ValueRef<int>`.

Does that means it casts parameters with `StringCast`?

```txt
                GenerateSitRepMessage
                    message = "EFFECT_MINES"
                    label = "EFFECT_MINES_LABEL"
                    icon = "icons/sitrep/combat_damage.png"
                    parameters = [
                        tag = "empire" data = Target.Owner
                        tag = "fleet" data = Target.ID
                        tag = "rawtext" data = @1@
                        tag = "system" data = Target.SystemID
                    ]
                    empire = Source.Owner
```

o01eg:
@geoffthemedio @Vezzra I converted `tech/defense/Defense`.

geoffthemedio:
> Does that means it casts parameters with `StringCast`?

yes:
https://github.com/freeorion/freeorion/blob/master/parse/EffectParser1.cpp#L220

parse/TechsParser.cpp
```diff
@@ -207,16 +210,490 @@ namespace {
...
+
+        return boost::python::object();
+    }
+
+    template <ValueRef::OpType O>
+    boost::python::object insert_minmaxoneof_(const py_grammar_techs& g, const boost::python::tuple& args, const boost::python::dict& kw) {
+        if (args[0] == g.m_parser.type_int) {
+            std::vector<std::unique_ptr<ValueRef::ValueRef<int>>> operands;
+            operands.reserve(boost::python::len(args) - 1);
+            for (size_t i = 1; i < boost::python::len(args); i++) {
```
geoffthemedio: `size_t` -> `auto`

parse/TechsParser.cpp
```diff
@@ -207,16 +210,490 @@ namespace {
...
+                    operands.push_back(ValueRef::CloneUnique(arg().value_ref));
+                } else {
+                    operands.push_back(std::make_unique<ValueRef::Constant<int>>(boost::python::extract<int>(args[i])()));
+                }
+            }
+            return boost::python::object(value_ref_wrapper<int>(std::make_shared<ValueRef::Operation<int>>(O, std::move(operands))));
+        } else if (args[0] == g.m_parser.type_float) {
+            std::vector<std::unique_ptr<ValueRef::ValueRef<double>>> operands;
+            operands.reserve(boost::python::len(args) - 1);
+            for (size_t i = 1; i < boost::python::len(args); i++) {
```
geoffthemedio: `size_t` -> `auto`

geoffthemedio:
Need
````
diff --git "a/msvc2019/Test/Test.vcxproj" "b/msvc2019/Test/Test.vcxproj"
index 7e6bcd4ec..79b14801f 100644
--- "a/msvc2019/Test/Test.vcxproj"
+++ "b/msvc2019/Test/Test.vcxproj"
@@ -118,7 +118,7 @@
       <FunctionLevelLinking>true</FunctionLevelLinking>
       <IntrinsicFunctions>true</IntrinsicFunctions>
       <PreprocessorDefinitions>NDEBUG;FREEORION_BUILD_HUMAN;FREEORION_WIN32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <AdditionalIncludeDirectories>../../;../../GG/;../../../include/;../include/;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>../../;../../GG/;../../../include/;../include/;../../../include/python3.5/;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PrecompiledHeaderFile>StdAfx.h</PrecompiledHeaderFile>
       <PrecompiledHeaderOutputFile>$(IntDir)$(TargetName).pch</PrecompiledHeaderOutputFile>
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
````

**Comment by geoffthemedio** [comment]
Currents gets stuck on startup with repetitions of this in the client log:
`10:35:32.190207 {0x00009e80} [debug] client : Pending.h:54 : Waiting for parse of "common" to complete.
10:35:32.190207 {0x00009e80} [debug] client : Pending.h:60 : WaitForPendingUnlocked another wait_for round`

**Comment by o01eg** [comment]
@geoffthemedio Does previous PR https://github.com/freeorion/freeorion/pull/3373 stuck as well?

**Comment by o01eg** [comment]
@geoffthemedio Could you check it again? I removed incorrect `set_exception_at_thread_exit` usage.

File: msvc2017/Parsers/Parsers.vcxproj [hunk 7]
```diff
@@ -261,6 +262,10 @@
     <ClCompile Include="..\..\parse\UniverseObjectTypeValueRefParser.cpp" />
     <ClCompile Include="..\..\parse\ValueRefParser.cpp" />
     <ClCompile Include="..\..\parse\VisibilityValueRefParser.cpp" />
+    <ClCompile Include="..\..\parse\PythonParser.cpp" />
```
geoffthemedio: this duplicates line 225, which breaks the project file

geoffthemedio: after fixing that, seems to start ok with a702bfb7418a

o01eg: Fixed it

geoffthemedio:
needs rebase

o01eg:
@geoffthemedio Yes, I see. Maybe if @Vezzra confirm this PR we could already start merge with previous PRs and rebase over them.

Vezzra (changes requested):
I can build and run this PR, but when I try to start a quickstart game, I get the "Errors occurred during universe generation..." message box, and I end up on the main menu screen. However, the game is somehow running in the background, when I try to exit FO I get the message "A savegame is in progress", which stays until I press the "Exit without saving" button.

Logfiles:  [logfiles.zip](https://github.com/freeorion/freeorion/files/6255127/logfiles.zip)

o01eg:
Do you have same issue with previous PR #3373 ?

o01eg:
Looks like it somehow missing `__import__` in `builtins` when initializes `globals` for techs parsing.

I suppose it related to https://github.com/pybind/pybind11/pull/2616

o01eg:
@Vezzra Could you try again https://github.com/freeorion/freeorion/pull/3309/commits/b3fb3d5510a8ae6b66ce01f4cf215dee167f4398 ?

Vezzra:
@o01eg,

> Could you try again https://github.com/freeorion/freeorion/pull/3309/commits/b3fb3d5510a8ae6b66ce01f4cf215dee167f4398 ?

That's actually commit 164a43c0ebd5eec1f891675a0b3e9f4ad1687035, right? That commit works fine. Can build, run and start a game now, no immediately obvious issues.

> Do you have same issue with previous PR #3373 ?

As this PR (#3309) works, I guess I don't need to check #3373, correct?

o01eg:
> @o01eg,
> 
> > Could you try again [b3fb3d5](https://github.com/freeorion/freeorion/commit/b3fb3d5510a8ae6b66ce01f4cf215dee167f4398) ?
> 
> That's actually commit [164a43c](https://github.com/freeorion/freeorion/commit/164a43c0ebd5eec1f891675a0b3e9f4ad1687035), right? That commit works fine. Can build, run and start a game now, no immediately obvious issues.

Yes, I decided to rebase it recently.

> 
> > Do you have same issue with previous PR #3373 ?
> 
> As this PR (#3309) works, I guess I don't need to check #3373, correct?

Yes, I told about previous PR in case issues found.

Could we start merging python parser PR as it works now with all supported OS?

geoffthemedio:
This parses all the Defense category techs? Any others?

o01eg:
@geoffthemedio It parses only `techs/defense/Defense.focs.py`, `techs/growth/ORBITAL_HAB.focs.py`, and `techs/learning/ALGO_ELEGANCE.focs.py`.

Vezzra (approved):
Builds and runs fine on macOS. No immediately obvious issues.

geoffthemedio:
seems fine.

need to fix up the tech tree controls better than [master](https://github.com/freeorion/freeorion/commit/082f01a93e9cd3d643c755f6916bdf13570aba01), after #3379 as [noted](https://github.com/freeorion/freeorion/pull/3403#issuecomment-819735668) but that is pre-existing...

geoffthemedio:
> seems fine.

@o01eg Fine with me to merge if you're ready. Still marked as Draft though...

o01eg:
@geoffthemedio I was expected to merge it step by step but if it's OK to merge them all then I'm fine with it too.

PR #3325, opened 2021-02-23
Create codeql-analysis.yml

geoffthemedio:
Might do a better job than LGTM, which can't seem to handle all the C++ code.


Cjkjvfnby:
I prefer to have issues reported before pushing to matter rather than after. 

It's a matter of who must deal with it. If we check before the merge, it is the responsibility of the PR author. If we check after merge nobody is responsible for that.

geoffthemedio:
> I prefer to have issues reported before pushing to matter rather than after.
> 
> It's a matter of who must deal with it. If we check before the merge, it is the responsibility of the PR author. If we check after merge nobody is responsible for that.

Not sure what your point is with regard to this pull request...

Cjkjvfnby:
> Not sure what your point is with regard to this pull request...

```
on:
  push:
    branches: [ master ]
```

geoffthemedio:
The next two (non-comment) lines cover pull requests on master as well:
````
on:
  push:
    branches: [ master ]
  pull_request:
    # The branches below must be a subset of the branches above
    branches: [ master ]
````
But this pull request is current just testing to see if it will work at all.

File: .github/workflows/codeql-analysis.yml [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+            libfreetype6-dev \
+            libpng-dev \
+            libogg-dev \
+            libvorbis-dev \
+            libsdl2-dev \
+            cppcheck \
+
+    - name: Build C++
+      run: |
+       export CPLUS_INCLUDE_PATH=/usr/include/python3.6
```
geoffthemedio, line 58: @o01eg do you know a better way to get this path included? Without it [can't find pyconfig.h](https://github.com/freeorion/freeorion/runs/1988712963#step:6:235) but this method seems too hard coded / inflexible...  Maybe it's fine as this will only be run on github for code analysis, though...?

o01eg: `$ python3-config --includes` should return some `-I<path>` arguments

geoffthemedio:
superseeded by #3333

PR #3327, opened 2021-02-24
Create artifacts_cleanup_expired.yml

geoffthemedio:
Script that should apparently help clean up expired build artifacts from GitHub Actions...

### DIFF
--- a/.github/workflows/artifacts_cleanup_expired.yml
+++ b/.github/workflows/artifacts_cleanup_expired.yml
@@ -1,0 +1,12 @@
+name: 'Purge expired artifacts'
+on:
+  schedule:
+    - cron: '0 * * * *'
+jobs:
+  purge-artifacts:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: LKP-RnD/purge-expired-artifacts-action@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          repo_to_purge: "freeorion/freeorion"


.github/workflows/artifacts_cleanup_expired.yml
```diff
@@ -0,0 +1,12 @@
+name: 'Purge expired artifacts'
+on:
+  schedule:
+    - cron: '0 * * * *'
```
Cjkjvfnby, line 4: https://crontab.guru/every-hour

How long does it take for an artifact to become expired?  What does expired mean in our case?

geoffthemedio: I don't know. See https://github.com/LKP-RnD/purge-expired-artifacts-action

The CodeQL analysis failed because t[oo much disk space was taken up](https://github.com/github/codeql-action/issues/402#issuecomment-785246631), so I googled for how to clean up artifacts, and found a few scripts and am trying one out.

Cjkjvfnby: > To set an expiry time for artifacts you can use an artifact upload action that supports this or change the default expiry time for your repository.

What kind of artifacts we have?  I guess it could be some caches for builds and some CodeQL or LGTM artifacts.

Could you check the default expiration time?  https://github.com/freeorion/freeorion/settings/actions
It's 90 days by default.

geoffthemedio: > What kind of artifacts we have? I guess it could be some caches for builds and some CodeQL or LGTM artifacts.

GitHub actions builds and CodeQL, but not LGTM as that's a separate site that wouldn't be affected by this file within the GitHub Workflows directories.

> Could you check the default expiration time? https://github.com/freeorion/freeorion/settings/actions
> It's 90 days by default.

Was 90. I set it now to 14. Could go down to 1 day.

PR #3328, opened 2021-02-25
Add new building: Abandon Outpost System

agrrr3:
this PR adds a building which can be used to make an outpost unowned again

it has a timer built in, cant abandon the outpost if there is battle in the system and gives useful feedback using sitrep messages

it could maybe be improved by adding a special to carry the timer (because that cant be removed), but i think this is good enough and the implementation with the special would be way more ugly

[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11933)

### DIFF
--- a/default/scripting/techs/construction/OUTPOST.focs.txt
+++ b/default/scripting/techs/construction/OUTPOST.focs.txt
@@ -5,6 +5,7 @@ Tech
     category = "CONSTRUCTION_CATEGORY"
     researchcost = 1
     researchturns = 1
+    unlock = Item type = Building name = "BLD_ABANDON_OUTPOST"
     tags = [ "PEDIA_CONSTRUCTION_CATEGORY" ]
 
     // Effects for outposts
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -13431,10 +13431,10 @@ BLD_EVACUATION_DESC
 Remove [[metertype METER_POPULATION]] from this planet over several turns according to an [[encyclopedia EVACUATION_TITLE]]. If a suitable alternate planets are available, with the same species and sufficient extra capacity, Population will move there.
 
 BLD_ABANDON_OUTPOST
-Abandon System
+Abandon
 
 BLD_ABANDON_OUTPOST_DESC
-Abandon the outpost on this planet. The Abandon System building needs two turns to prepare dismantling the outpost. If it is prepared, it needs an uninterupted turn without battles happening to completely abandon the outpost. The building gets used up in the process and the planet will be unowned afterwards.
+Abandon the outpost on this planet. The building needs two turns to prepare dismantling the outpost. If dismantling is prepared, it takes an uninterupted turn without battles happening to completely abandon the outpost. The building gets used up in the process and the planet will be unowned afterwards.
 
 BLD_OBSERVATORY
 Observatory
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -6149,7 +6149,7 @@ OUTPOSTS_TITLE
 Outpost
 
 OUTPOSTS_TEXT
-Outposts are unmanned stations, which can be founded in places too hostile for a colony to survive. They have no [[metertype METER_POPULATION]], and normally produce no resources. But they create [[metertype METER_SUPPLY]] (with the right tech) and provide fuel and vision. Colonies can be founded on planets that already have an Outpost. Building an Outpost requires a ship with an [[shippart CO_OUTPOST_POD]]. Outposts are also formed when the Population of a colony drops to zero for any reason. You can abandon an outpost by building a [[building BLD_ABANDON_OUTPOST]].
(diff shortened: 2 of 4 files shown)


default/stringtables/en.txt
```diff
@@ -13415,6 +13430,12 @@ Evacuation System
 BLD_EVACUATION_DESC
 Remove [[metertype METER_POPULATION]] from this planet over several turns according to an [[encyclopedia EVACUATION_TITLE]]. If a suitable alternate planets are available, with the same species and sufficient extra capacity, Population will move there.
 
+BLD_ABANDON_OUTPOST
+Abandon System
```
geoffthemedio: Remove / replace "System" in the name. Makes purpose unclear.

agrrr3: Should I also remove "System" from the "Evacuation System" building? (I think the system part was introduced to gap the difference between a building and an activity)

geoffthemedio: For consistency, probably yes.

To be clear, in the case of "Evacuation System", it's not ambiguous whether "System" refers to a star system vs. a means and process for doing something (ie. the evacuation).

default/stringtables/en.txt
```diff
@@ -13410,11 +13425,17 @@ Additional experience and facilities improvements further extend the size of ind
...
 
 BLD_EVACUATION
-Evacuation System
+Evacuation
 
 BLD_EVACUATION_DESC
 Remove [[metertype METER_POPULATION]] from this planet over several turns according to an [[encyclopedia EVACUATION_TITLE]]. If a suitable alternate planets are available, with the same species and sufficient extra capacity, Population will move there.
 
+BLD_ABANDON_OUTPOST
+Abandon
```
geoffthemedio: probably should be "Abandon Outpost"

default/stringtables/en.txt
```diff
@@ -13410,11 +13425,17 @@ Additional experience and facilities improvements further extend the size of ind
 ##
 
 BLD_EVACUATION
-Evacuation System
+Evacuation
```
geoffthemedio: Could be "Evacuate Planet" in consistent style with suggested "Abdandon Outpost"

default/stringtables/en.txt
```diff
@@ -6149,7 +6149,7 @@ OUTPOSTS_TITLE
 Outpost
 
 OUTPOSTS_TEXT
-Outposts are unmanned stations, which can be founded in places too hostile for a colony to survive. They have no [[metertype METER_POPULATION]], and normally produce no resources. But they create [[metertype METER_SUPPLY]] (with the right tech) and provide fuel and vision. Colonies can be founded on planets that already have an Outpost. Building an Outpost requires a ship with an [[shippart CO_OUTPOST_POD]]. Outposts are also formed when the Population of a colony drops to zero for any reason.
+Outposts are unmanned stations, which can be founded in places too hostile for a colony to survive. They have no [[metertype METER_POPULATION]], and normally produce no resources. But they create [[metertype METER_SUPPLY]] (with the right tech) and provide fuel and vision. Colonies can be founded on planets that already have an Outpost. Building an Outpost requires a ship with an [[shippart CO_OUTPOST_POD]]. Outposts are also formed when the Population of a colony drops to zero for any reason. You can abandon an outpost by building a [[building BLD_ABANDON_OUTPOST]].
```
geoffthemedio: needs to be `[[buildingtype `

geoffthemedio:
Could you also add a building that abandons a colony with population, reverting to or making it become neutral but populated?

agrrr3:
> Could you also add a building that abandons a colony with population, reverting to or making it become neutral but populated?

Can do. Should i make the defense levels "sticky" - keeping the current defense/shield levels so the missing techs dont make it too cheap a target to reinvade?

geoffthemedio:
> Should i make the defense levels "sticky" - keeping the current defense/shield levels so the missing techs dont make it too cheap a target to reinvade?

Max or current meters? If max, how would you do that? If the latter, sure, just let the current meters adjust to the new situation naturally...

agrrr3:
> > Should i make the defense levels "sticky" - keeping the current defense/shield levels so the missing techs dont make it too cheap a target to reinvade?
> 
> Max or current meters? If max, how would you do that? If the latter, sure, just let the current meters adjust to the new situation naturally...

i thought about max meters. adding buildings or specials which save the target meter values as capacity and set the max meter to that (and remove that stuff upon invasion).

geoffthemedio:
probably not then

default/scripting/techs/construction/OUTPOST.focs.txt
```diff
@@ -5,6 +5,7 @@ Tech
     category = "CONSTRUCTION_CATEGORY"
     researchcost = 1
     researchturns = 1
+    unlock = Item type = Building name = "BLD_ABANDON_OUTPOST"
```
geoffthemedio, line 8: this is malformed, needs to go after tags. breaks parsing.

geoffthemedio: https://github.com/freeorion/freeorion/pull/3340

geoffthemedio: https://github.com/freeorion/freeorion/commit/ce332ea9b5961fa21dd744238b627ab96ef81d91

PR #3332, opened 2021-02-27
Add getters for ValueRef fields of Tech

o01eg:


universe/Tech.h
```diff
@@ -95,6 +95,9 @@ class FO_COMMON_API Tech {
     const std::vector<UnlockableItem>&    UnlockedItems() const
     { return m_unlocked_items; }
 
+    const ValueRef::ValueRef<double>* ResearchCost() const { return m_research_cost.get(); } //!< return value ref of research cost
```
geoffthemedio: name as `ResearchCostRef` or similar to disambiguate with the same-name functions above

o01eg: Fixed

geoffthemedio:
You could also add `bool Tech::operator==(const Tech&) const` to do the comparison. Might be cleaner.

PR #3336, opened 2021-02-28
Add BLD_COLONY_INDEPENDENCE_DECREE to unown your colony (MAX DEFENSE version)

agrrr3:
[forum discussion](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11952)

Geoff asked for a "unown colony building" in  #3328

This also contains a feature which allows to keep the defense levels of the colony as opposed to #3337 where defense/shields/troops drop when going independent

This feature needs a discussion though, but one can check out this pr to see it work.


default/scripting/common/priorities.macros
```diff
@@ -111,6 +111,9 @@ LATE_AFTER_ALL_TARGET_MAX_METERS_PRIORITY
 FOCUS_CHANGE_PENALTY_PRIORITY
 '''600'''
 
+COLONY_INDEPENDENCE_PRIORITY
+'''650'''
+
```
Oberlus: Is it necessary to introduce another priority level? Can be done with existing levels? Maybe using `METER_OVERRIDE_PRIORITY` is enough.

hifron:
Shouldn§t it be to declare own parliament and possibly race under current administration to gain some profits(maybe in culture and maybe to advance current culture into another consideration)?

default/scripting/common/priorities.macros
agrrr3: probably its not necessary.

geoffthemedio:
An issue is that the planet still appears as owned by the player, but just just visible by the player, after the builiding takes effect. Perhaps the building could also have a one-turn visibility modifier to make sure the player is aware of the new ownership by empire -1 ?

turn_after

hifron:
Turkish solution???

agrrr3:
> An issue is that the planet still appears as owned by the player, but just just visible by the player, after the builiding takes effect. Perhaps the building could also have a one-turn visibility modifier to make sure the player is aware of the new ownership by empire -1 ?

interesting. done.

geoffthemedio:
Since you've added specials to maintain defense, shields, and troops, what about population? Currently if a planet's target population is > 0 only because of empire-dependent effects, independent planets will depopulate over a few turns if their species doesn't inherently have enough target population on that planet to remain viable...

agrrr3:
> Since you've added specials to maintain defense, shields, and troops, what about population?

done

geoffthemedio:
another issue... most player-colonized planets aren't homeworlds, so don't have the usual +5 to stability from homeworlds that is present on native planets. independent colonies thus have 0 target stability by default, so don't actually have any way to grow their defense meters...

geoffthemedio:
Also maybe need something similar for planets that go independent due to rebel troops, as they will otherwise sometimes then depopulate over the next few turns due to having target population <= 0 without their former empire's support. Maybe that's not a problem, though?

PR #3342, opened 2021-03-03
Make python into freeorioncommon dependency

o01eg:
Also uses python library version to find MacOS python folder.

Extracted from #3211.

Follows up #3371 but could be merged separately.


geoffthemedio:
Is Python actually a common library dependency?

o01eg:
It wasn't but it will be one after migration to python parser. Also I found out another usage of python here for MacOS so I decided to extract into separate PR.

Actually the only case when server and AI disabled is Android experimental build. So it isn't some additional dependency.

Vezzra (approved):
I can build, run and launch a default quickstart game with this PR on macOS.

geoffthemedio:
seems fine

PR #3344, opened 2021-03-06
Allow AI to delegate player

o01eg:
Allows locally launched AI to skip authentication.
Allows player or AI to delegate someone without lobby if they have single empire to play.

It still requires to manually launch AI process although.

Could be back-ported to 0.4.10 branch.

### DIFF
--- a/server/ServerApp.cpp
+++ b/server/ServerApp.cpp
@@ -1859,8 +1859,6 @@ int ServerApp::AddPlayerIntoGame(const PlayerConnectionPtr& player_connection, i
     int empire_id = ALL_EMPIRES;
     std::list<std::string> delegation = GetPlayerDelegation(player_connection->PlayerName());
     if (target_empire_id == ALL_EMPIRES) {
-        if (!delegation.empty())
-            return ALL_EMPIRES;
         // search empire by player name
         for (auto& e : Empires()) {
             if (e.second->PlayerName() == player_connection->PlayerName()) {
@@ -1868,6 +1866,20 @@ int ServerApp::AddPlayerIntoGame(const PlayerConnectionPtr& player_connection, i
                 empire = e.second;
                 break;
             }
+        }
+        // Assign player to empire if he doesn't have own empire and delegates signle
+        if (delegation.size() == 1 && empire == nullptr) {
+            for (auto e : Empires()) {
+                if (e.second->PlayerName() == delegation.front()) {
+                    empire_id = e.first;
+                    empire = e.second;
+                    break;
+                }
+            }
+        }
+        if (!delegation.empty()) {
+            DebugLogger() << "ServerApp::AddPlayerIntoGame(...): Player should choose between delegates.";
+            return ALL_EMPIRES;
         }
     } else {
         // use provided empire and test if it's player himself or one of delegated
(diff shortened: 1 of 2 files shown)


server/ServerApp.cpp
```diff
@@ -1869,6 +1867,20 @@ int ServerApp::AddPlayerIntoGame(const PlayerConnectionPtr& player_connection, i
                 break;
             }
         }
+        // Assign player to empire if he doesn't have own empire and delegates signle
```
geoffthemedio, line 1870: meaning unclear... assign the new player to an empire if that player doesn't already have an empire "and delegates signle" ...?  Maybe "Assign the newly-connected player to an empire that doesn't already have a player and that has delegated (only) the name of the new player to play it"....?

Why can't an empire be delegated to multiple players, and the server assign any of those players to that empire when the player joins?

What if multiple empires have delegated the same player to play them? How does it pick which empire to assign the player to? That seems like the more problematic situation to address...

server/ServerFSM.cpp
```diff
@@ -1158,7 +1158,8 @@ sc::result MPLobby::react(const JoinGame& msg) {
         if (authenticated)
             player_connection->SetAuthenticated();
     } else {
-        if (client_type != Networking::ClientType::CLIENT_TYPE_AI_PLAYER && server.IsAuthRequiredOrFillRoles(player_name, roles)) {
+        const bool relaxed_auth = player_connection->IsLocalConnection() && client_type != Networking::ClientType::CLIENT_TYPE_AI_PLAYER;
```
geoffthemedio, line 1161: I don't understand this... Why the not-an-AI requirement? Would be good to comment to explain the logic / expectations...

And what does relaxed authentication mean, exactly? What is the alternative? (No authentication required, or no connecton possible?)

server/ServerApp.cpp
```diff
@@ -1869,6 +1867,20 @@ int ServerApp::AddPlayerIntoGame(const PlayerConnectionPtr& player_connection, i
...
+        if (delegation.size() == 1 && empire == nullptr) {
+            for (auto e : Empires()) {
+                if (boost::algorithm::to_lower_copy(e.second->PlayerName()) == boost::algorithm::to_lower_copy(delegation.front())) {
```
geoffthemedio: `boost::iequals` might be less verbose?

server/ServerApp.cpp
o01eg: > meaning unclear... assign the new player to an empire if that player doesn't already have an empire "and delegates signle" ...? Maybe "Assign the newly-connected player to an empire that doesn't already have a player and that has delegated (only) the name of the new player to play it"....?

No, this function either chooses single empire for player so he enter to the game and controls that empire or doesn't choose any so player is sent to lobby. Before this PR player who doesn't own empire but have one empire he delegates was sent to lobby where he had choose this empire. AI could not manage it. So after this PR such player automatically assigned to that empire.

> Why can't an empire be delegated to multiple players, and the server assign any of those players to that empire when the player joins?

Yes, it possible but server doesn't search player for the empire, it searches empire for the player.

> What if multiple empires have delegated the same player to play them? How does it pick which empire to assign the player to? That seems like the more problematic situation to address...

In this case player isn't assigned to empire so he is sent to lobby. Then in lobby he choose empire for himself like he did it was open saved game. This function is called again with chosen empire id in `target_empire_id` and else branch checks and assigns chosen empire for the player.

server/ServerFSM.cpp
o01eg: > I don't understand this... Why the not-an-AI requirement? Would be good to comment to explain the logic / expectations...

Because AI could not answer passwords to authentication requests.

> And what does relaxed authentication mean, exactly? What is the alternative? (No authentication required, or no connecton possible?)

It means local AI can join the game without authentication check, no authentication required for it. Not sure what you mean for alternative

server/ServerApp.cpp
o01eg: No, I just blindly copied condition from my server code with case insensitive player names.

geoffthemedio:
some elaboration and explanation of what's going on in comments would be helpful, still...

PR #3345, opened 2021-03-06
Add clone method to value ref and condition

o01eg:
Follows up #3265. Requirement for #3309 .

Workaround for missing `std::unique_ptr` support in Boost.Python https://github.com/boostorg/python/issues/202

### DIFF
--- a/universe/Condition.h
+++ b/universe/Condition.h
@@ -93,6 +93,9 @@ struct FO_COMMON_API Condition {
     virtual unsigned int GetCheckSum() const
     { return 0; }
 
+    //! Makes a clone of this Condition in a new owning pointer. Required for Boost.Python, which
+    //! doesn't supports move semantics for returned values.
+    [[nodiscard]] virtual std::unique_ptr<Condition> Clone() const = 0;
 protected:
     bool m_root_candidate_invariant = false;
     bool m_target_invariant = false;
--- a/universe/Condition.h
+++ b/universe/Condition.h
@@ -97,6 +97,9 @@ struct FO_COMMON_API Condition {
     //! doesn't supports move semantics for returned values.
     [[nodiscard]] virtual std::unique_ptr<Condition> Clone() const = 0;
 protected:
+    //! Copies invariants from other Condition
+    Condition(const Condition& rhs) = default;
+
     bool m_root_candidate_invariant = false;
     bool m_target_invariant = false;
     bool m_source_invariant = false;
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -2164,7 +2164,7 @@ void Type::GetDefaultInitialCandidateObjects(const ScriptingContext& parent_cont
                 AddSystemSet(parent_context.ContextObjects(), condition_non_targets);
                 break;
             case UniverseObjectType::OBJ_FIGHTER:   // shouldn't exist outside of combat as a separate object
-            default: 
+            default:
                 found_type = false;
                 break;
         }
@@ -2418,6 +2418,19 @@ HasSpecial::HasSpecial(std::unique_ptr<ValueRef::ValueRef<std::string>>&& name,
     m_source_invariant = boost::algorithm::all_of(operands, [](auto& e){ return !e || e->SourceInvariant(); });
 }
 
+HasSpecial::HasSpecial(const HasSpecial& rhs) :
+    Condition(),
+    m_name(ValueRef::CloneUnique(rhs.m_name)),
+    m_capacity_low(ValueRef::CloneUnique(rhs.m_capacity_low)),
+    m_capacity_high(ValueRef::CloneUnique(rhs.m_capacity_high)),
+    m_since_turn_low(ValueRef::CloneUnique(rhs.m_since_turn_low)),
+    m_since_turn_high(ValueRef::CloneUnique(rhs.m_since_turn_high))
+{
+    m_root_candidate_invariant = rhs.m_root_candidate_invariant;
+    m_target_invariant = rhs.m_target_invariant;
+    m_source_invariant = rhs.m_source_invariant;
+}
+
 bool HasSpecial::operator==(const Condition& rhs) const {
     if (this == &rhs)
         return true;
@@ -2608,15 +2621,7 @@ unsigned int HasSpecial::GetCheckSum() const {
 }
 
 std::unique_ptr<Condition> HasSpecial::Clone() const {
-    auto retval = std::make_unique<HasSpecial>(ValueRef::CloneUnique(m_name),
-                                               ValueRef::CloneUnique(m_since_turn_low),
-                                               ValueRef::CloneUnique(m_since_turn_high));
-    retval->m_capacity_low = ValueRef::CloneUnique(m_capacity_low);
-    retval->m_capacity_high = ValueRef::CloneUnique(m_capacity_high);
-    retval->m_root_candidate_invariant = m_root_candidate_invariant;
-    retval->m_target_invariant = m_target_invariant;
-    retval->m_source_invariant = m_source_invariant;
-    return retval;
+    return std::make_unique<HasSpecial>(*this);
 }
 
 ///////////////////////////////////////////////////////////
@@ -2908,13 +2913,13 @@ namespace {
         ContainsSimpleMatch(const ObjectSet& subcondition_matches) :
             m_subcondition_matches_ids()
         {
-            // We need a sorted container for efficiently intersecting 
+            // We need a sorted container for efficiently intersecting
             // subcondition_matches with the set of objects contained in some
             // candidate object.
             // We only need ids, not objects, so we can do that conversion
             // here as well, simplifying later code.
-            // Note that this constructor is called only once per 
-            // Contains::Eval(), its work cannot help performance when executed 
+            // Note that this constructor is called only once per
+            // Contains::Eval(), its work cannot help performance when executed
             // for each candidate.
             m_subcondition_matches_ids.reserve(subcondition_matches.size());
             // gather the ids
@@ -4532,6 +4537,20 @@ Enqueued::Enqueued(BuildType build_type,
     m_source_invariant = boost::algorithm::all_of(operands, [](auto& e){ return !e || e->SourceInvariant(); });
 }
 
+Enqueued::Enqueued(const Enqueued& rhs) :
+    Condition(),
+    m_build_type(rhs.m_build_type),
+    m_name(ValueRef::CloneUnique(rhs.m_name)),
+    m_design_id(ValueRef::CloneUnique(rhs.m_design_id)),
+    m_empire_id(ValueRef::CloneUnique(rhs.m_empire_id)),
+    m_low(ValueRef::CloneUnique(rhs.m_low)),
+    m_high(ValueRef::CloneUnique(rhs.m_high))
+{
+    m_root_candidate_invariant = rhs.m_root_candidate_invariant;
+    m_source_invariant = rhs.m_source_invariant;
+    m_target_invariant = rhs.m_target_invariant;
+}
+
 bool Enqueued::operator==(const Condition& rhs) const {
     if (this == &rhs)
         return true;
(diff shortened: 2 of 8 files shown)


universe/Condition.h
```diff
@@ -93,6 +93,10 @@ struct FO_COMMON_API Condition {
     virtual unsigned int GetCheckSum() const
     { return 0; }
 
+    //! Copies value into new ownening pointer. Required for Boost.Python which
+    //! doesn't supports move semantic for returned values.
+    virtual std::unique_ptr<Condition> Copy() const = 0;
+
```
geoffthemedio: Should be named `Clone()`

geoffthemedio:
When working, make the `Clone()` commits a separate stand-alone pull request.

See also: https://github.com/freeorion/freeorion/pull/3309#issuecomment-791911615

universe/Condition.h
o01eg: Renamed it

universe/Conditions.cpp
```diff
@@ -666,6 +674,12 @@ unsigned int Turn::GetCheckSum() const {
     return retval;
 }
 
+std::unique_ptr<Condition> Turn::Clone() const override {
+    auto retval = std::make_unique<Turn>(m_low ? std::move(m_low->Clone()) : nullptr,
```
geoffthemedio: `std::move` probably unnecessary since it's already an rvalue

universe/Condition.h
```diff
@@ -93,6 +93,9 @@ struct FO_COMMON_API Condition {
     virtual unsigned int GetCheckSum() const
     { return 0; }
 
+    //! Makes a clone of value into new owning pointer. Required for Boost.Python which
```
geoffthemedio: Makes a clone of this Condition in a new owning pointer. Required for Boost.Python, which doesn't support move semantics for returned values.

universe/ValueRefs.h
```diff
@@ -377,6 +424,15 @@ struct FO_COMMON_API Operation final : public ValueRef<T>
 
     unsigned int GetCheckSum() const override;
 
+    std::unique_ptr<ValueRef<T>> Clone() const override {
+        std::vector<std::unique_ptr<Operation<T>>> operands(m_operands.size());
```
geoffthemedio: if you init with a size, it creates the objects and default inits. if you then emplace back, you'll have twice as many, the first half being all `nullptr`. you probably want to default init the `vector` and call `reserve(m_operands.size())` and then `push_back` all the clones.

universe/ValueRefs.h
```diff
@@ -377,6 +424,15 @@ struct FO_COMMON_API Operation final : public ValueRef<T>
...
+        for (const auto& operand : m_operands) {
+            operands.emplace_back(operand ? std::move(std::make_unique<ValueRef<T>>(operand->Clone())) : nullptr);
+        }
+        auto retval = std::make_unique<Operation<T>>(m_op_type, std::move(operands));
+        return std::move(retval);
```
geoffthemedio: probably should not `std::move` to return; I believe it prevents move elision. rather just `return std::make_unique<...>(...)`

universe/Conditions.cpp
o01eg: Fixed.

universe/ValueRefs.h
o01eg: Fixed.

universe/ValueRefs.h
o01eg: Fixed.

universe/Conditions.cpp
```diff
@@ -524,6 +524,13 @@ unsigned int Number::GetCheckSum() const {
     return retval;
 }
 
+std::unique_ptr<Condition> Number::Clone() const {
+    return std::make_unique<Number>(m_low ? std::move(m_low->Clone()) : nullptr,
+                                    m_high ? std::move(m_high->Clone()) : nullptr,
```
geoffthemedio: most / all the other `move(...)` are unnecessary as well... `Clone()` returns an rvalue.

universe/Conditions.cpp
```diff
@@ -1788,6 +1838,19 @@ unsigned int Homeworld::GetCheckSum() const {
     return retval;
 }
 
+std::unique_ptr<Condition> Homeworld::Clone() const {
+    std::vector<std::unique_ptr<ValueRef::ValueRef<std::string>>> names;
+    names.reserve(m_names.size());
+    for (const std::unique_ptr<ValueRef::ValueRef<std::string>>& name : m_names) {
```
geoffthemedio: can do just `for (const auto& name : m_names)` probably

universe/Conditions.cpp
```diff
@@ -1788,6 +1838,19 @@ unsigned int Homeworld::GetCheckSum() const {
...
+        if (name) {
```
geoffthemedio: Can collapse to `names.push_back(name ? name->Clone() : nullptr);` or with `emplace_back` (should be the same result).

universe/ValueRefs.h
```diff
@@ -196,6 +211,14 @@ struct FO_COMMON_API ComplexVariable final : public Variable<T>
     const ValueRef<std::string>* StringRef2() const;
     unsigned int GetCheckSum() const override;
 
+    std::unique_ptr<ValueRef<T>> Clone() const override {
+        return std::make_unique<ComplexVariable<T>>(Variable<T>::PropertyName()[0],
```
geoffthemedio, line 215: Can't use `operator[]` on the return result of `PropertyName()` because `operator[]` can mutate what it operates on and `PropertyName()` returns a const reference.

Also, not sure if that will really fully clone `this` in all cases if you just access `[0]`.

Also, not sure if it's safe to assume that `PropertyName()` returns a non-empty `vector` reference.

Also, don't need to, as there are `Variable<T>` constructors taking `std::vector<std::string>` (const or &&).

geoffthemedio: These non-trivial `Clone()` implementations should probably be in the .cpp not the .h

universe/ValueRefs.h
```diff
@@ -377,6 +417,19 @@ struct FO_COMMON_API Operation final : public ValueRef<T>
...
+        std::vector<std::unique_ptr<Operation<T>>> operands;
+        operands.reserve(m_operands.size());
+        for (const std::unique_ptr<ValueRef<T>>& operand : m_operands) {
+            if (operand) {
+                operands.push_back(std::move(operand->Clone()));
+            } else {
+                operands.emplace_back(nullptr);
+            }
+        }
+        return std::make_unique<Operation<T>>(m_op_type, std::move(operands));
```
geoffthemedio: Would perhaps be nice to clone all the invariants directly rather than relying on the constructor re-determining them.

universe/Condition.h
o01eg: Fixed.

universe/ValueRefs.h
o01eg: It's a template so it cannot be in *.cpp. 

`ComplexVariable` constructor fills single variable name as I understand so PropertyName here should have only one value.

geoffthemedio: > It's a template so it cannot be in *.cpp.

good point

> `ComplexVariable` constructor fills single variable name as I understand so PropertyName here should have only one value.

perhaps, but no reason to not just pass the whole `vector`, and a couple other reasons to do so.

universe/ValueRefs.h
o01eg: Could consructor determine them differently?

geoffthemedio: If you construct a new ValueRef / Condition using the standard constructors, it just gets pointers to sub-refs and sub-condtions and needs to check them all to determine its own invariance.

You could also set up a base class virtual `Copy` method and use that in `Clone` as suggested here: https://github.com/freeorion/freeorion/pull/3309#issuecomment-791911615 and copy the invariants in the base class implementations.

I don't think it makes sense to also set up constructor overloads that take invariants and pass them to the base classes similar overloads.

universe/Conditions.cpp
o01eg: Fixed.

universe/Conditions.cpp
o01eg: Fixed.

universe/Conditions.cpp
o01eg: Fixed.

o01eg:
This commits builds fine for me

universe/Conditions.cpp
```diff
@@ -9385,6 +9675,24 @@ unsigned int ValueTest::GetCheckSum() const {
     return retval;
 }
 
+std::unique_ptr<Condition> ValueTest::Clone() const {
+    auto retval = std::make_unique<ValueTest>(ValueRef::CloneUnique(m_value_ref1),
```
geoffthemedio: for cases like this, I'd rather add a new private constructor that takes no parameters and doesn't init the invariants. it seems odd and assymmetric to use the 3-int constructor, which does all the invariant initialization checks for those input valuerefs, and then overwrite that by copying from `this` anyway.

or perhaps a private copy constructor that takes `const ValueTest&` and initializes all the members, including the invariants, without redoing the calculation of them...
````
private:
    ValueTest(const ValueTest& rhs);

...

ValueTest::ValueTest(const ValueTest& rhs) :
    Condition(),
    m_value_ref1(ValueRef::CloneUnique(rhs.m_value_ref1)),
    m_value_ref2(ValueRef::CloneUnique(rhs.m_value_ref2)),
    m_value_ref3(ValueRef::CloneUnique(rhs.m_value_ref3)),
    m_string_value_ref1(ValueRef::CloneUnique(rhs.m_string_value_ref1)),
    m_string_value_ref2(ValueRef::CloneUnique(rhs.m_string_value_ref2)),
    m_string_value_ref3(ValueRef::CloneUnique(rhs.m_string_value_ref3)),
    m_int_value_ref1(ValueRef::CloneUnique(rhs.m_int_value_ref1)),
    m_int_value_ref2(ValueRef::CloneUnique(rhs.m_int_value_ref2)),
    m_int_value_ref3(ValueRef::CloneUnique(rhs.m_int_value_ref3)),
    m_compare_type1(rhs.m_compare_type1),
    m_compare_type2(rhs.m_compare_type2)
{
    m_root_candidate_invariant = rhs.m_root_candidate_invariant;
    m_target_invariant = rhs.m_target_invariant;
    m_source_invariant = rhs.m_source_invariant;
}
````

o01eg: Not sure why, but `std::make_unique` somehow doesn't accept private constructors.

geoffthemedio: I was hoping one could make `std::make_unique` a friend function of `ValueTest` but that apparently isn't portably possible due to STD implementation details.

I guess it's fine to just make the custom copy constructor public...

````
public:
   [...]
    ValueTest(const ValueTest& rhs); // deep copy / clone of contained unique_ptr
````

I'd also add a protected copy constructor to base class `Condition`, which when defaulted should just copy the invariant members:
````
protected:
    Condition(const Condition& rhs) = default; // not virtual, intended for use within derived class copy constructors
````

and use that within the `ValueTest` copy constructor, to avoid the explicit invariant assignments:
````
ValueTest::ValueTest(const ValueTest& rhs) :
    Condition(rhs),
    m_value_ref1(ValueRef::CloneUnique(rhs.m_value_ref1)),
    m_value_ref2(ValueRef::CloneUnique(rhs.m_value_ref2)),
    m_value_ref3(ValueRef::CloneUnique(rhs.m_value_ref3)),
    m_string_value_ref1(ValueRef::CloneUnique(rhs.m_string_value_ref1)),
    m_string_value_ref2(ValueRef::CloneUnique(rhs.m_string_value_ref2)),
    m_string_value_ref3(ValueRef::CloneUnique(rhs.m_string_value_ref3)),
    m_int_value_ref1(ValueRef::CloneUnique(rhs.m_int_value_ref1)),
    m_int_value_ref2(ValueRef::CloneUnique(rhs.m_int_value_ref2)),
    m_int_value_ref3(ValueRef::CloneUnique(rhs.m_int_value_ref3)),
    m_compare_type1(rhs.m_compare_type1),
    m_compare_type2(rhs.m_compare_type2)
{}
````

o01eg: Made it

geoffthemedio: from edit:

> I'd also add a protected copy constructor to base class `Condition`, which when defaulted should just copy the invariant members:

PR #3347, opened 2021-03-09
Implement Clone for Effect

o01eg:
Another requirement for #3309

### DIFF
--- a/universe/Effects.cpp
+++ b/universe/Effects.cpp
@@ -464,6 +464,10 @@ unsigned int NoOp::GetCheckSum() const {
     return retval;
 }
 
+std::unique_ptr<Effect> NoOp::Clone() const {
+    return std::make_unique<NoOp>();
+}
+
 
 ///////////////////////////////////////////////////////////
 // SetMeter                                              //
@@ -679,6 +683,12 @@ unsigned int SetMeter::GetCheckSum() const {
     return retval;
 }
 
+std::unique_ptr<Effect> SetMeter::Clone() const {
+    return std::make_unique<SetMeter>(m_meter,
+                                      ValueRef::CloneUnique(m_value),
+                                      m_accounting_label);
+}
+
 
 ///////////////////////////////////////////////////////////
 // SetShipPartMeter                                      //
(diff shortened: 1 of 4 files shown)


universe/Effects.cpp
```diff
@@ -998,6 +1014,13 @@ unsigned int SetEmpireMeter::GetCheckSum() const {
     return retval;
 }
 
+std::unique_ptr<Effect> SetEmpireMeter::Clone() const {
+    auto meter = m_meter;
```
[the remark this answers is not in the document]
o01eg: Yes, and when I tried to change reference type to `&&` and got errors in parsers.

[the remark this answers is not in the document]
geoffthemedio: > 
> 
> Yes, and when I tried to change reference type to `&&` and got errors in parsers.

which is why it's `&`

universe/Effects.cpp
```diff
@@ -1863,6 +1936,16 @@ unsigned int CreateShip::GetCheckSum() const {
     return retval;
 }
 
+std::unique_ptr<Effect> CreateShip::Clone() const {
+    auto retval = std::make_unique<CreateShip>(ValueRef::CloneUnique(m_design_name),
+                                               ValueRef::CloneUnique(m_empire_id),
+                                               ValueRef::CloneUnique(m_species_name),
+                                               ValueRef::CloneUnique(m_name),
+                                               ValueRef::CloneUnique(m_effects_to_apply_after));
+    retval->m_design_id = ValueRef::CloneUnique(m_design_id);
```
geoffthemedio, line 1945: I suppose this is OK since the base class `Effect` has no state and `CreateShip` has no other state or initialization upon construction besides these `unique_ptr`. Still a bit awkward to have the one `CloneUnique` separately. Not sure if it would be better to skip the constructor with parameter calls...

universe/ValueRef.h
```diff
@@ -132,6 +132,16 @@ template<typename T>
     return retval;
 }
 
+template<typename T>
+[[nodiscard]] inline std::vector<std::pair<std::string, std::unique_ptr<T>>> CloneUnique(const std::vector<std::pair<std::string, std::unique_ptr<T>>>& vec) {
+    std::vector<std::pair<std::string, std::unique_ptr<T>>> retval;
+    retval.reserve(vec.size());
+    for (const auto& val : vec) {
+        retval.push_back(std::make_pair(val.first, CloneUnique(val.second)));
```
geoffthemedio: should be able to `emplace_back` without `make_pair`.

o01eg: Fixed.

PR #3353, opened 2021-03-11
Kick out other players controls same empire

o01eg:
Additionally kills AI process if other player was AI.


geoffthemedio:
I don't have an easy way to test this, so if you think it's fine, then fine...

o01eg:
I made `auth.py` `get_player_delegation` returns `AI_1` to be delegated. Then started first client and host multiplayer game with two 2, then I started second client and in lobby chosen empire of `AI_1`.

geoffthemedio:
So you tested it and it seemed to work?

o01eg:
Yes

PR #3356, opened 2021-03-12
Add totalFighterDamage and totalShipDamage calculations

agrrr3:
I am working on combat damage estimation based on valuerefs. 

In the end I want to have two numbers shown in the UI. 
One is the maximum damage a ship can do against a ship in a combat. The other is the maximum count of fighters a ship can destroy in a combat.

In some places we currently show the per-bout damage, but in many cases I think it makes more sense to give the damage per combat as that is directly comparable to enemy structure for a turn-advance.

Implementation is based on providing two optional fields in weapon and hangar definitions: totalFighterDamage and totalShipDamage. There you can override the default implementation.
Note the backend implementation drifts towards using the term "destroy" capacity for the number of fighters an attacker can shoot as opposed to "damage" (used in its most specific sense) meaning the amount of damage to structure against ship targets.

Currently implemented as examples are the SR_WEAPON_0_1 (Flak) which has totalShipDamage = 0, SR_WEAPON_1_1 (mass driver) which has totalFighterDamage = 0 and SR_JAWS which only shoots in close range bouts (usually 2 out of 4 bouts -> half the normal damage).

- [ ] add to documentation: need to use immediate values (e.g. `Value(Source.MaxStructure)`), else the info in the design window is wrong (as the initial values are zero)
- [ ] add to documentation: dont use ThisPart here, but the part name instead (maybe open an issue?)
- [x] use the totalFighterDamage in the UI
- [x] check has weapon/blockade implications (weapons able to shoot down only fighters should also count for blockade?)
- [x] (my) decision: blockade should only be possible by ships which are able to damage ships (so destroying fighters is not enough), it doesnt matter if that damage comes from fighters (strikers,(heavy)bombers) are direct weapons. If a blockade is enforced by a fighter force, eliminating that fighter force (e.g. using flak) can break the blockade.
- [x] UI decision: should fighter caused damage be included in "destroy tally"? min/max issue? - yes, including it seems the better default
- [x] UI decision: should the fleet window show whole-combat damage? min/max issue with fighters? - whole-combat damage. min/max damage is out of scope for this PR, so showing only max damage.
- [x] decision: should the lower level backend functions calculate combat-damage or bout-damage as default (those are convertible on ship level/knowing the targeting)? the targeting can turn damage for a bout on/off. fighter damage depends also on launch behaviour (which needs launch bay and hangar capacities).
- [x] adapt all the weapons
- [x] maybe switch in all that backend-code to full combat damage instead of per-bout
- [x] also check for shields (target-dependent ShipDamage) (currently half implemented/not working)
- [x] see followups... ~~also check for Attack property, it currently triggers an endless recursion. using it should currently produce a warning at least (so the game crashes and you can easily see that using the Attack valueref was the cause).~~ Will crash for scope of this PR.
- [x] https://www.freeorion.org/forum/viewtopic.php?f=6&t=11985 IsArmed/HasArmedShips discussion and implementation -> going for CanDamageShips for supply and blockade

Primary test cases
- Strikers against shields (design screen), ARC_DISRUPTOR (design screen), Capital against shields, Starting frigate ship&fleet,

Secondary test cases
- SR Jaws (design screen), Mass Driver, Flak, Bomber, Interceptor, shields yes/no, CR weapon (ship)

Followups/Out of scope for this PR
- [ ] adapt planet defense UI to show full combat damage(?) - problem is defense dual role - one might want to know both the "shield equivalent" and the "total structural damage!
- [ ] show total fighter hangar(?) damage in ship damage popup - it may make sense to group multiple weapons of a kind
- [ ] find a good UI to give a feeling for real combat strength. Max total damage assumes that you loose no ships nor fighters, which is far from what is usually happening, so a span would be nice but hard to do nicely in UI
- [ ] using METER_CAPACITY and METER_MAX_CAPACITY for structural damage and fighter destruction leads to "strange" labels in the UI (as the meter descriptions are used) - I think that happens in MeterBrowseWnd
- [ ] asteroid snail should not have a SR_WEAPON_3_1, it blows the cover. Maybe some close range arc disruption instead or/and a jaws upgrade?
- [ ] per bout damage info on fleet (and maybe system) level as a mouse-over
- [ ] preventing/warning scripters not to use TotalFighterDamageEstimation in totalFighterDamageEstimation
- [ ] preventing/warning scripters not to use TotalShipDamageEstimation in totalShipDamageEstimation
- [ ] rename FighterDamage -> FighterDestruction (in backend and FOCS)
- [ ] maybe rename ShipDamage -> StructuralDamage (in backend and FOCS)
- [ ] AI interface 

[developer chat notes](https://www.freeorion.org/forum/viewtopic.php?p=105197#p105197)

### DIFF
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -504,14 +504,16 @@ void ShipDamageBrowseWnd::UpdateSummary() {
     float breakdown_total = ship->TotalWeaponsShipDamage(0.0f, false);
     float breakdown_total_augmentation = ship->TotalWeaponsFighterDamage(false);
     std::string breakdown_meter_name = UserString("SHIP_DAMAGE_STAT_TITLE");
-
+    std::string breakdown_meter_augmentation_name = UserString("SHIP_DESTRUCTION_STAT_TITLE");
+    
 
     // set accounting breakdown total / summary
     if (m_meter_title)
         m_meter_title->SetText(boost::io::str(FlexibleFormat(UserString("TT_BREAKDOWN_AUGMENTED_SUMMARY")) %
                                               breakdown_meter_name %
                                               DoubleToString(breakdown_total, 3, false) %
-                                              DoubleToString(breakdown_total_augmentation, 1, false)));
+                                              breakdown_meter_augmentation_name %
+                                              (int)breakdown_total_augmentation));
 }
 
 void ShipDamageBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -501,7 +501,7 @@ void ShipDamageBrowseWnd::UpdateSummary() {
         return;
 
     // unpaired meter total for breakdown summary
-    float breakdown_total = ship->TotalWeaponsDamage(0.0f, false);
+    float breakdown_total = ship->TotalWeaponsShipDamage(0.0f, false); // FIXME do something with ship->TotalWeaponsFighterDamage
     std::string breakdown_meter_name = UserString("SHIP_DAMAGE_STAT_TITLE");
 
 
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -1,6 +1,7 @@
 #include "MeterBrowseWnd.h"
 
 #include "../util/i18n.h"
+#include "../util/GameRules.h"
 #include "../util/Logger.h"
 #include "../universe/Building.h"
 #include "../universe/Effect.h"
@@ -501,19 +502,15 @@ void ShipDamageBrowseWnd::UpdateSummary() {
         return;
 
     // unpaired meter total for breakdown summary
-    float breakdown_total = ship->TotalWeaponsShipDamage(0.0f, false);
-    float breakdown_total_augmentation = ship->TotalWeaponsFighterDamage(false);
-    std::string breakdown_meter_name = UserString("SHIP_DAMAGE_STAT_TITLE");
-    std::string breakdown_meter_augmentation_name = UserString("SHIP_DESTRUCTION_STAT_TITLE");
-    
-
+    float total_structure_damage = ship->TotalWeaponsShipDamage(0.0f, false);
+    float total_fighters_destroyed = ship->TotalWeaponsFighterDamage(false);
+    int num_bouts = GetGameRules().Get<int>("RULE_NUM_COMBAT_ROUNDS");
     // set accounting breakdown total / summary
     if (m_meter_title)
-        m_meter_title->SetText(boost::io::str(FlexibleFormat(UserString("TT_BREAKDOWN_AUGMENTED_SUMMARY")) %
-                                              breakdown_meter_name %
-                                              DoubleToString(breakdown_total, 3, false) %
-                                              breakdown_meter_augmentation_name %
-                                              (int)breakdown_total_augmentation));
+        m_meter_title->SetText(boost::io::str(FlexibleFormat(UserString("TT_DAMAGE_BREAKDOWN_SUMMARY")) %
+                                              num_bouts %
+                                              DoubleToString(total_structure_damage, 3, false) %
+                                              (int)total_fighters_destroyed));
 }
 
 void ShipDamageBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -546,7 +546,7 @@ void ShipDamageBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
             continue;
 
         // get the attack power for each weapon part
-        const ScriptingContext context(ship);
+        const ScriptingContext context{ship};
         float part_attack = ship->WeaponPartShipDamage(part, context);
         float part_fighters_shot = ship->WeaponPartFighterDamage(part, context);
 
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -501,15 +501,17 @@ void ShipDamageBrowseWnd::UpdateSummary() {
         return;
 
     // unpaired meter total for breakdown summary
-    float breakdown_total = ship->TotalWeaponsShipDamage(0.0f, false); // FIXME do something with ship->TotalWeaponsFighterDamage
+    float breakdown_total = ship->TotalWeaponsShipDamage(0.0f, false);
+    float breakdown_total_augmentation = ship->TotalWeaponsFighterDamage(false);
     std::string breakdown_meter_name = UserString("SHIP_DAMAGE_STAT_TITLE");
 
 
     // set accounting breakdown total / summary
     if (m_meter_title)
-        m_meter_title->SetText(boost::io::str(FlexibleFormat(UserString("TT_BREAKDOWN_SUMMARY")) %
+        m_meter_title->SetText(boost::io::str(FlexibleFormat(UserString("TT_BREAKDOWN_AUGMENTED_SUMMARY")) %
                                               breakdown_meter_name %
-                                              DoubleToString(breakdown_total, 3, false)));
+                                              DoubleToString(breakdown_total, 3, false) %
+                                              DoubleToString(breakdown_total_augmentation, 1, false)));
 }
 
 void ShipDamageBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
@@ -544,18 +546,20 @@ void ShipDamageBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
             continue;
 
         // get the attack power for each weapon part
-        float part_attack = ship->CurrentPartMeterValue(MeterType::METER_CAPACITY, part_name) *
-                            ship->CurrentPartMeterValue(MeterType::METER_SECONDARY_STAT, part_name);
+        const ScriptingContext context(ship);
+        float part_attack = ship->WeaponPartShipDamage(part, context);
+        float part_fighters_shot = ship->WeaponPartFighterDamage(part, context);
+
         auto text = boost::io::str(FlexibleFormat(label_template) % name % UserString(part_name));
 
         auto label = GG::Wnd::Create<CUILabel>(std::move(text), GG::FORMAT_RIGHT);
         label->MoveTo(GG::Pt(GG::X0, top));
-        label->Resize(GG::Pt(MeterBrowseLabelWidth(), m_row_height));
+        label->Resize(GG::Pt(MeterBrowseLabelWidth() - MeterBrowseValueWidth(), m_row_height));
         AttachChild(label);
 
-        auto value = GG::Wnd::Create<CUILabel>(ColouredNumber(part_attack));
-        value->MoveTo(GG::Pt(MeterBrowseLabelWidth(), top));
-        value->Resize(GG::Pt(MeterBrowseValueWidth(), m_row_height));
+        auto value = GG::Wnd::Create<CUILabel>(ColouredNumber(part_attack) + " / " + ColouredNumber(part_fighters_shot, 0));
+        value->MoveTo(GG::Pt(MeterBrowseLabelWidth() - MeterBrowseValueWidth(), top));
+        value->Resize(GG::Pt(2*MeterBrowseValueWidth(), m_row_height));
         AttachChild(value);
         m_effect_labels_and_values.emplace_back(std::move(label), std::move(value));
 
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -4,6 +4,7 @@
 #include "../util/Logger.h"
 #include "../util/GameRules.h"
 #include "../universe/Building.h"
+#include "../universe/Condition.h"
 #include "../universe/Effect.h"
 #include "../universe/Planet.h"
 #include "../universe/PopCenter.h"
@@ -721,13 +722,29 @@ namespace {
     };
 
     /** Populate fighter state quantities and damages for each combat round until @p limit_to_bout */
-    std::map<int, FighterBoutInfo> ResolveFighterBouts(int bay_capacity, int current_docked,
+    std::map<int, FighterBoutInfo> ResolveFighterBouts(const std::shared_ptr<Ship> ship, const Condition::Condition* combat_targets,
+                                                       int bay_capacity, int current_docked,
                                                        float fighter_damage, int limit_to_bout = -1)
     {
+        // FIXME this should be moved to combat probably
         std::map<int, FighterBoutInfo> retval;
         const int NUM_BOUTS = GetGameRules().Get<int>("RULE_NUM_COMBAT_ROUNDS");
         int target_bout = limit_to_bout < 1 ? NUM_BOUTS : limit_to_bout;
+
+        // FIXME duplicates code from Ship.cpp
+        //auto target = std::make_shared<Fighter>();
+        // XXX DesignID valid?
+        auto target = std::make_shared<Ship>(ALL_EMPIRES, ship->DesignID(), "SP_HUMAN");
+        // target needs to have an ID != -1 to be visible, stealth should be low enough
+        // structure must be higher than zero to be valid target
+        target->SetID(-1000000); // XXX magic number AutoresolveInfo.next_fighter_id starts at -1000001 counting down
+        target->GetMeter(MeterType::METER_STRUCTURE)->Set(100.0f, 100.0f);
+        target->GetMeter(MeterType::METER_MAX_STRUCTURE)->Set(100.0f, 100.0f);
+        GetUniverse().SetEmpireObjectVisibility(ship->Owner(), target->ID(), Visibility::VIS_FULL_VISIBILITY);
+        ScriptingContext context(ship, target);
+
         for (int bout = 1; bout <= target_bout; ++bout) {
+            context.combat_bout = bout;
             // init current fighters
             if (bout == 1) {
                 retval[bout].qty.docked = current_docked;
@@ -740,7 +757,11 @@ namespace {
                 retval[bout].total_damage = retval[bout - 1].total_damage;
             }
             // calc damage this bout, apply to total
-            retval[bout].damage = retval[bout].qty.attacking * fighter_damage;
+            int shots_this_bout = retval[bout].qty.attacking;
+            if (combat_targets && !combat_targets->Eval(context, context.effect_target)) {
+                shots_this_bout = 0;
+            }
+            retval[bout].damage = shots_this_bout * fighter_damage;
             retval[bout].total_damage += retval[bout].damage;
             // launch fighters
             if (bout < NUM_BOUTS) {
@@ -791,7 +812,7 @@ void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
     int hangar_current_fighters = 0;
     int hangar_total_capacity = 0;
     int bay_total_capacity = 0;
-
+    const Condition::Condition* combat_targets = nullptr;
     // populate values from hangars and bays
     for (std::string part_name : parts) {
         const ShipPart* part = GetShipPart(part_name);
@@ -810,6 +831,7 @@ void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
         } else if (part_class == ShipPartClass::PC_FIGHTER_HANGAR) {
             if (hangar_part.first.empty()) {
                 hangar_part.first = part_name;
+                combat_targets = part->CombatTargets();
             } else if (hangar_part.first != part_name) {
                 ErrorLogger() << "Ship " << ship->ID() << "contains different hangar parts: "
                               << hangar_part.first << ", " << part_name;
@@ -860,7 +882,7 @@ void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
     if (!m_show_all_bouts) {
         // Show damage for first wave (2nd combat round)
         std::map<int, FighterBoutInfo> bout_info = ResolveFighterBouts(
-            bay_total_capacity, hangar_current_fighters, fighter_damage, 2);
+            ship, combat_targets, bay_total_capacity, hangar_current_fighters, fighter_damage, 2);
         FighterBoutInfo first_wave = bout_info.rbegin()->second;
         GG::Clr highlight_clr = bout_info[1].qty.launched < bay_total_capacity ? HANGAR_COLOR : BAY_COLOR;
         std::string launch_text = ColouredInt(first_wave.qty.attacking, false, highlight_clr);
@@ -923,7 +945,7 @@ void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
         // TODO Add list of effects on hangar(fighter) damage
 
         std::map<int, FighterBoutInfo> bout_info = ResolveFighterBouts(
-            bay_total_capacity, hangar_current_fighters, fighter_damage);
+            ship, combat_targets, bay_total_capacity, hangar_current_fighters, fighter_damage);
         const FighterBoutInfo& last_bout = bout_info.rbegin()->second;
 
         // damage summary text
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -2,9 +2,7 @@
 
 #include "../util/i18n.h"
 #include "../util/Logger.h"
-#include "../util/GameRules.h"
 #include "../universe/Building.h"
-#include "../universe/Condition.h"
 #include "../universe/Effect.h"
 #include "../universe/Planet.h"
 #include "../universe/PopCenter.h"
@@ -13,6 +11,7 @@
 #include "../universe/ShipPart.h"
 #include "../universe/UniverseObject.h"
 #include "../Empire/Empire.h"
+#include "../combat/CombatDamage.h"
 #include "../client/human/GGHumanClientApp.h"
 #include "ClientUI.h"
 #include "CUIDrawUtil.h"
@@ -708,71 +707,6 @@ void ShipFightersBrowseWnd::UpdateSummary() {
                                               DoubleToString(breakdown_total, 3, false)));
 }
 
-namespace {
-    /** Container for return values of ResolveFighterBouts */
-    struct FighterBoutInfo {
-        struct StateQty {
-            int         docked;
-            int         attacking;
-            int         launched;       // transitioning between docked and attacking
-        };
-        float           damage;
-        float           total_damage;   // damage from all previous bouts, including this bout
-        StateQty        qty;
-    };
-
-    /** Populate fighter state quantities and damages for each combat round until @p limit_to_bout */
-    std::map<int, FighterBoutInfo> ResolveFighterBouts(const std::shared_ptr<Ship> ship, const Condition::Condition* combat_targets,
-                                                       int bay_capacity, int current_docked,
-                                                       float fighter_damage, int limit_to_bout = -1)
-    {
-        // FIXME this should be moved to combat probably
-        std::map<int, FighterBoutInfo> retval;
-        const int NUM_BOUTS = GetGameRules().Get<int>("RULE_NUM_COMBAT_ROUNDS");
-        int target_bout = limit_to_bout < 1 ? NUM_BOUTS : limit_to_bout;
-
-        // FIXME duplicates code from Ship.cpp
-        //auto target = std::make_shared<Fighter>();
-        // XXX DesignID valid?
-        auto target = std::make_shared<Ship>(ALL_EMPIRES, ship->DesignID(), "SP_HUMAN");
-        // target needs to have an ID != -1 to be visible, stealth should be low enough
-        // structure must be higher than zero to be valid target
-        target->SetID(-1000000); // XXX magic number AutoresolveInfo.next_fighter_id starts at -1000001 counting down
-        target->GetMeter(MeterType::METER_STRUCTURE)->Set(100.0f, 100.0f);
-        target->GetMeter(MeterType::METER_MAX_STRUCTURE)->Set(100.0f, 100.0f);
-        GetUniverse().SetEmpireObjectVisibility(ship->Owner(), target->ID(), Visibility::VIS_FULL_VISIBILITY);
-        ScriptingContext context(ship, target);
-
-        for (int bout = 1; bout <= target_bout; ++bout) {
-            context.combat_bout = bout;
-            // init current fighters
-            if (bout == 1) {
-                retval[bout].qty.docked = current_docked;
-                retval[bout].qty.attacking = 0;
-                retval[bout].qty.launched = 0;
-            } else {
-                retval[bout].qty = retval[bout - 1].qty;
-                retval[bout].qty.attacking += retval[bout].qty.launched;
-                retval[bout].qty.launched = 0;
-                retval[bout].total_damage = retval[bout - 1].total_damage;
-            }
-            // calc damage this bout, apply to total
-            int shots_this_bout = retval[bout].qty.attacking;
-            if (combat_targets && !combat_targets->Eval(context, context.effect_target)) {
-                shots_this_bout = 0;
-            }
-            retval[bout].damage = shots_this_bout * fighter_damage;
-            retval[bout].total_damage += retval[bout].damage;
-            // launch fighters
-            if (bout < NUM_BOUTS) {
-                retval[bout].qty.launched = std::min(bay_capacity, retval[bout].qty.docked);
-                retval[bout].qty.docked -= retval[bout].qty.launched;
-            }
-        }
-        return retval;
-    }
-}
-
 void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
     // clear existing labels
     for (unsigned int i = 0; i < m_effect_labels_and_values.size(); ++i) {
@@ -881,9 +815,9 @@ void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
 
     if (!m_show_all_bouts) {
         // Show damage for first wave (2nd combat round)
-        std::map<int, FighterBoutInfo> bout_info = ResolveFighterBouts(
-            ship, combat_targets, bay_total_capacity, hangar_current_fighters, fighter_damage, 2);
-        FighterBoutInfo first_wave = bout_info.rbegin()->second;
+        std::map<int, Combat::FighterBoutInfo> bout_info = Combat::ResolveFighterBouts(
+            std::static_pointer_cast<const Ship>(ship), combat_targets, bay_total_capacity, hangar_current_fighters, fighter_damage, 2);
+        Combat::FighterBoutInfo first_wave = bout_info.rbegin()->second;
         GG::Clr highlight_clr = bout_info[1].qty.launched < bay_total_capacity ? HANGAR_COLOR : BAY_COLOR;
         std::string launch_text = ColouredInt(first_wave.qty.attacking, false, highlight_clr);
         std::string damage_label_text = boost::io::str(FlexibleFormat(UserString("TT_FIGHTER_DAMAGE"))
@@ -944,9 +878,9 @@ void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
         // Damage summary labels
         // TODO Add list of effects on hangar(fighter) damage
 
-        std::map<int, FighterBoutInfo> bout_info = ResolveFighterBouts(
+        std::map<int, Combat::FighterBoutInfo> bout_info = Combat::ResolveFighterBouts(
             ship, combat_targets, bay_total_capacity, hangar_current_fighters, fighter_damage);
-        const FighterBoutInfo& last_bout = bout_info.rbegin()->second;
+        const Combat::FighterBoutInfo& last_bout = bout_info.rbegin()->second;
 
         // damage summary text
         auto& damage_total_text = UserString("SHIP_FIGHTERS_DAMAGE_TOTAL");
@@ -974,7 +908,7 @@ void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
         GG::X left = GG::X0;
         int previous_docked = hangar_current_fighters;
         for (auto& current_bout : bout_info) {
-            const FighterBoutInfo& bout = current_bout.second;
+            const Combat::FighterBoutInfo& bout = current_bout.second;
             // combat round label
             std::string bout_text = boost::io::str(FlexibleFormat(UserString("TT_COMBAT_ROUND"))
                                                    % IntToString(current_bout.first));
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -817,8 +817,8 @@ void ShipFightersBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
         std::map<int, Combat::FighterBoutInfo> bout_info = Combat::ResolveFighterBouts(
             std::static_pointer_cast<const Ship>(ship), combat_targets, bay_total_capacity, hangar_current_fighters, fighter_damage, 2);
         Combat::FighterBoutInfo first_wave = bout_info.rbegin()->second;
-        GG::Clr highlight_clr = bout_info[1].qty.launched < bay_total_capacity ? HANGAR_COLOR : BAY_COLOR;
-        std::string launch_text = ColouredInt(first_wave.qty.attacking, false, highlight_clr);
+        GG::Clr highlight_clr = bout_info[1].launched < bay_total_capacity ? HANGAR_COLOR : BAY_COLOR;
+        std::string launch_text = ColouredInt(first_wave.attacking, false, highlight_clr);
         std::string damage_label_text = boost::io::str(FlexibleFormat(UserString("TT_FIGHTER_DAMAGE"))
                                                        % launch_text % fighter_damage_text);
         // damage formula label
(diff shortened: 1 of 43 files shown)


universe/Ship.cpp
```diff
@@ -456,6 +457,10 @@ namespace {
         int available_fighters = 0;
 
         retval.reserve(parts.size() + 1);
+        //const ScriptingContext context(ship); // XXX expects a shared_ptr ???
+        std::shared_ptr<Ship> ship_clone(ship->Clone()); // XXX ugly way to create a shared_ptr necessary for the context
+        const ScriptingContext context(ship_clone);
```
agrrr3: @geoffthemedio is there a better way to get a shared_ptr for a ship so I can use it as Source in context evaluation?

universe/Ship.cpp
```diff
@@ -464,7 +469,11 @@ namespace {
             ShipPartClass part_class = part->Class();
 
             // get the attack power for each weapon part.
-            if (part_class == ShipPartClass::PC_DIRECT_WEAPON) {
+            if (part->TotalShipDamage()) { // FIXME TotalFighterDamage
+                // there is a special computation. overrides the default one
+                double estimation = part->TotalShipDamage()->Eval(context);
```
agrrr3: @geoffthemedio here I use the context for the evaluation of the totalShipDamage valueref. But it seems it does not work - I try to access the ship inside the context via Source, but it always returns zero (e.g. Source.Structure)

geoffthemedio: should be able to use `shared_from_this()`

universe/Ship.cpp
geoffthemedio: should be able to use `shared_from_this()`

universe/Ship.cpp
```diff
@@ -438,7 +438,7 @@ float Ship::TotalWeaponsShipDamage(float shield_DR, bool include_fighters) const
 }
 
 namespace {
-    std::vector<float> WeaponShipDamageImpl(const Ship* ship, const ShipDesign* design,
+    std::vector<float> WeaponShipDamageImpl(const std::shared_ptr<Ship> ship, const ShipDesign* design,
```
geoffthemedio: you want to pass `std::shared_ptr<const Ship>` not `const std::shared_ptr<Ship>`

universe/Ship.cpp
```diff
@@ -531,7 +529,7 @@ std::vector<float> Ship::AllWeaponsMaxShipDamage(float shield_DR , bool include_
     if (!design)
         return retval;
 
-    return WeaponShipDamageImpl(this, design, shield_DR, true, include_fighters);
+    return WeaponShipDamageImpl(std::const_pointer_cast<Ship>(std::static_pointer_cast<const Ship>(shared_from_this())), design, shield_DR, true, include_fighters);
```
geoffthemedio: the `const_pointer_cast` looks questionable... change what you pass above.

universe/Ship.cpp
agrrr3: done.

Still doesnt work, always get back zero if I try to access Source properties (e.g. Source.MaxStructure). 

In the end I want the following to work in e.g. SR_JAWS
```(ShipPartMeter part = ThisPart meter = Capacity object = Source.ID)```

Comment (agrrr3) [reply universe/Ship.cpp]: It does actually work for normal ships. Just the temporary one used in the ship design does not work.

Comment (agrrr3) [reply universe/Ship.cpp]: Maybe it worked before (not sure). ThisPart was a culprit at least in the design screen. Using "SR_JAWS" instead worked.

Source.MaxStructure still does not work in the design screen but on in-game ships.

Comment (agrrr3) [reply universe/Ship.cpp]: Ok, that is because the initial value is zero. One needs to use `Value(Source.MaxStructure)`

**Comment by agrrr3** [comment]
@geoffthemedio i did a prototypical default implementation for fighter damage estimation in cpp using the combatTargets conditions in the last commit aaed01a13a71fad8f45978fb9155c5b05a582894 . I think the launch patterns are too complicated to script in FOCS. The default implementation works for the existing fighters, but if one uses more complicated combatTarget conditions one probably wont be able to calculate a correct damage estimation.

**Comment by geoffthemedio** [comment]
> I think the launch patterns are too complicated to script in FOCS.

What do you need to calculate / consider for that? Perhaps a function could be written in C++ and exposed to FOCS to do that calculation, if it's too complicated to script with just FOCS expressions.

**Comment by agrrr3** [comment]
> > I think the launch patterns are too complicated to script in FOCS.
> 
> What do you need to calculate / consider for that? Perhaps a function could be written in C++ and exposed to FOCS to do that calculation, if it's too complicated to script with just FOCS expressions.

@geoffthemedio the damage estimation value refs calculate the total damage per combat (because e.g. only half damage for close range weapons). for calculating the total maximum damage a carrier does you need the launch patterns - and the pattern is interleaved with the damage, so you would need to know how many fighters are launched in a certain bout. 

So I guess the valueref would look like this: `MaxNumberOfFighterShots countIf = <condition>`

And the striker damage estimation would look something like this
```
shipDamageEstimation = MaxNumberOfFighterShots *  ShipPartMeter part = "FT_HANGAR_2" meter = SecondaryStat 
fighterDamageEstimation = MaxNumberOfFighterShots
```  

And a fighter type only shooting in bout three would use `MaxNumberOfFighterShots countIf = (CombatBout == 3)`

Else estimating the damage of a single fighter in a certain bout number would also work - but that is basically what the combatTargets condition does (turning damage on/off). So in the backend I could evaluate that combatTargets condition once for each bout number.

Note that there is in my opinion not really a use case for shooting only in a certain bout. The prototype for missiles did that but also launched all missiles in the same bout and the total damage estimation for that is easy (launch capacity * fighter damage). I imagine that usually you would rather want to use the number of bouts since launch. If fighters had such a property (and it should be easy to add) one could use it in the combatTargets condition and in the estimation e.g. for a missile doing damage exactly two bouts after launch: `MaxNumberOfFighterShots countIf = (CombatBout - Source.LaunchBout > 2)`

So what you say? Evaluating combatTargets multiple times or MaxNumberOfFighterShots valueref or delay until we need it?

edit1: MaxNumberOfLaunchedFighters -> MaxNumberOfFighterShots

**Comment by agrrr3** [comment]
SOTA update: i switched to full combat damage as a default, i also fixed a bug in hangar part display, for fighters i implemented both a `TotalFighterShots condition = OPTIONAL_TARGET_FILTER` condition (necessary to use if one uses CombatBout) and a fallback which uses the hangar part combatTargets if such exists (so the current hangar content is fine without using TotalFighterShots).

Damage for hangars calculates maximum damage. I think that suffices for this PR.

I guess depending on content using TotalFighterShots can be much faster than using the fallback. TotalFighterShots could be optimized for the most common cases (e.g. hangar capacity * combat bouts IFF launch capacity >= hangar capacity). Probably not worth it. As a use case I could update the missile prototype #2710 to give correct damage estimations.

So this is feature-ready, time to check the last issues and clean up the code.

**Comment by agrrr3** [comment]
@geoffthemedio i need advice on some topics, hope you can give me insight

1a) using the Attack valueref/object property inside of damage estimation leads to an endless recursion. The value ref calls for total damage calculation which uses the damage estimation if it exists. I am not sure how to prevent that in a good way. Would it be good enough if it logged a helpful warning/error when it crashes? I could pass a flag to the damage calculation that using damage estimation might be dangerous so I could log the warning if it goes to that code path. I could also stop the estimation at that point at the cost of a wrong Attack value. Also the only use currently is `default/scripting/empire_statistics/MILITARY_STRENGTH.focs.txt`. So maybe a documentation hint might be enough as well. 

1b) Also I am not sure how to figure in combat value of anti-fighter weapons and (of fighters as well) in a combined Attack value (on a basic level one must assume if and when fighters get shot down). maybe it makes more sense to split into MinimalStructuralDamage, MaximalStructuralDamage, MinimalFighterDestruction & MinimalFighterDestruction values. Note I will use the maximum structural damage done by ship+fighters in this PR and ignore destruction.

~~2) the code which knows about fighter launches and targeting etc (it also creates a fighter/ship dummy as target) should not reside in multiple places in the CombatSystem, Ship, UI (MeterBrowseWnd), ShipDesign (::AdjustedAttack) or valueref (TotalFighterShots) but in a central combat system related place. Should this become a public helper function in CombatSystem.cpp or something like CombatSystemDetails.cpp(?)~~
It became CombatDamage.cpp

~~3) i dont understand why the windows build is broken. the ubuntu 18.04 build seems to hang in the tests, 20.04 build seems to missing SDL2 headers. windows has a linker error (`Common.lib(Ship.obj) : error LNK2019: unresolved external symbol "public: bool __thiscall Condition::Condition::Eval(class ScriptingContext const &,class std::shared_ptr<class UniverseObject const >)const "` - i had a look at the msvc files, i would guess that the implementation should be there as Condition.cpp is in the common lib)~~
Seems the forward declaration was not enough. Including ScriptingContext.h in Ship.h seems to have worked

**Comment by geoffthemedio** [comment]
> 1a) using the Attack valueref/object property inside of damage estimation leads to an endless recursion. 

So don't do that then? I'm not sure what the problem is...

> The value ref calls for total damage calculation which uses the damage estimation if it exists. I am not sure how to prevent that in a good way.

Decide on an order of things that can call other things, and don't call in the other order. Perhaps this requires rewriting `Attack` so it doesn't call another function that can call `Attack` again.

> Would it be good enough if it logged a helpful warning/error when it crashes? 

No; it shouldn't crash.

> I could pass a flag to the damage calculation that using damage estimation might be dangerous so I could log the warning ...

Sounds overly complicated.

> Also the only use currently is `default/scripting/empire_statistics/MILITARY_STRENGTH.focs.txt`. 

If `Attack` is really only used there, can it just be replaced with your new calculation?

> I am not sure how to figure in combat value of anti-fighter weapons and (of fighters as well) in a combined Attack value (on a basic level one must assume if and when fighters get shot down). maybe it makes more sense to split into MinimalStructuralDamage, MaximalStructuralDamage, MinimalFighterDestruction & MinimalFighterDestruction values. 

I'm not sure what you mean by all that, but I think the point of `Attack` is to give a single summarizing attack strength value for the ship, which accounts for all its weapons and their various strengths. If such a number doesn't make sense anymore, then perhaps the `Attack` ValueRef doesn't make sense either.

**Comment by agrrr3** [comment]
> > 1a) using the Attack valueref/object property inside of damage estimation leads to an endless recursion.
> 
> So don't do that then? I'm not sure what the problem is...

I do not have control what scripters do. If Attack valueref needs to return the correct value, it needs access to the damageEstimation valueref. If a scripter uses the Attack valueref in the damageEstimation it creates a loop and crashes.

> > I am not sure how to figure in combat value of anti-fighter weapons and (of fighters as well) in a combined Attack value (on a basic level one must assume if and when fighters get shot down). maybe it makes more sense to split into MinimalStructuralDamage, MaximalStructuralDamage, MinimalFighterDestruction & MinimalFighterDestruction values.
> 
> I'm not sure what you mean by all that, but I think the point of `Attack` is to give a single summarizing attack strength value for the ship, which accounts for all its weapons and their various strengths. If such a number doesn't make sense anymore, then perhaps the `Attack` ValueRef doesn't make sense either.

It makes some sense (depending on context). If I have a single Attack value i have to put a damage value on destruction of fighters. If I use zero, I say flak does not contribute to Attack. This is correct if the enemies do not have fighters. If they have a lot of bombers flak may provide a lot of value in combat. Attack valueref could take an optional parameter for calculating that value (and it could depend on time as research increases bomber value). Or split Attack into AttackShipDamage and AttackFighterDestruction in the statistic instead of `(LocalCandidate.Attack * (LocalCandidate.Structure + 2 * LocalCandidate.MaxShield))` something like
`((LocalCandidate.AttackShipDamage + DESTRUCTED_FIGHTER_VALUE * LocalCandidate.AttackFighterDestruction) * (LocalCandidate.Structure + 2 * LocalCandidate.MaxShield))`

**Comment by geoffthemedio** [comment]
If a single `Attack` ref only makes sense in a context dependent way, then I don't think it makes sense at all, as it's supposed to be a single context-free value. You can probably construct an equivalent with a FOCS `max(...)` expression with several sub-expressions that cover those cases explicitly...

**Comment by agrrr3** [comment]
hm, waiting for feedback on code since two weeks.
@geoffthemedio  can we merge this?

**Comment by geoffthemedio** [comment]
This tooltip is unclear... (why two numbers?)
weapon_damage_tooltip

**Comment by geoffthemedio** [comment]
This ship either has 1.1k or 0.0 damage, depending whether I look at the icon or tooltip or pedia.

damage_1 1k_or_0 0

File: UI/MeterBrowseWnd.cpp [hunk 5]
```diff
@@ -544,18 +546,20 @@ void ShipDamageBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
             continue;
 
         // get the attack power for each weapon part
-        float part_attack = ship->CurrentPartMeterValue(MeterType::METER_CAPACITY, part_name) *
-                            ship->CurrentPartMeterValue(MeterType::METER_SECONDARY_STAT, part_name);
+        const ScriptingContext context(ship);
```
geoffthemedio: prefer `{}` brace initialization.

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+// damage calculation (shortrange, fighters, shields)
+
+namespace {
+    ScriptingContext CombatContextWithShipTarget(std::shared_ptr<const Ship> source, float shields = 0.0f) {
+        // FIXME find default enemy - i use the given design as that at least exists (but might be non-targetable because of stealth)
+        //ErrorLogger() << "DESIGN " << GetPredefinedShipDesignManager().GetDesignID("08a58b08-0929-496d-84fc-faa91424ca21");
+        //auto target = std::make_shared<Ship>(ALL_EMPIRES, GetPredefinedShipDesignManager().GetDesignID("SD_MARK_1"), "SP_HUMAN");
+        //auto target = std::make_shared<Fighter>();
+        // XXX DesignID valid?
+        auto target = std::make_shared<Ship>(ALL_EMPIRES, source->DesignID(), "SP_HUMAN");
```
geoffthemedio: you cannot assume the `SP_HUMAN` species exists in C++ code.

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+        // target needs to have an ID != -1 to be visible, stealth should be low enough
+        // structure must be higher than zero to be valid target
+        target->SetID(-1000000); // XXX magic number AutoresolveInfo.next_fighter_id starts at -1000001 counting down
+        target->GetMeter(MeterType::METER_STRUCTURE)->Set(100.0f, 100.0f);
+        target->GetMeter(MeterType::METER_MAX_STRUCTURE)->Set(100.0f, 100.0f);
+        // Shield value is used for structural damage estimation
+        target->GetMeter(MeterType::METER_SHIELD)->Set(shields,shields);
```
geoffthemedio: space after `,`

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    ScriptingContext CombatContextWithFighterTarget(std::shared_ptr<const Ship> source) {
+        auto target = std::make_shared<Fighter>();
+        target->SetID(-1000000); // XXX magic number AutoresolveInfo.next_fighter_id starts at -1000001 counting down
+        GetUniverse().SetEmpireObjectVisibility(source->Owner(), target->ID(), Visibility::VIS_FULL_VISIBILITY);
+        return ScriptingContext(source, target);
+    }
+
+    std::vector<float> WeaponDamageCalcImpl(std::shared_ptr<const Ship> ship, const ShipDesign* design,
```
geoffthemedio: purpose of `max` parameter is unclear

why is there both a ship and a ship design passed? purpose unclear

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+                                            bool max, bool include_fighters, bool target_ships, const ScriptingContext&& context) {
```
geoffthemedio: move `{` to new line

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+        for (const auto& part_name : parts) {
+            const ShipPart* part = GetShipPart(part_name);
+            if (!part)
+                continue;
+            ShipPartClass part_class = part->Class();
+
+            // get the attack power for each weapon part.
+            if (part_class == ShipPartClass::PC_DIRECT_WEAPON) {
+                if (target_ships)
+                    retval.emplace_back(ship->WeaponPartShipDamage(part, context));
```
geoffthemedio: prefer `push_back` to `emplace_back` for a single existing object. use `emplace_back` to forward parameters to a constructor

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+                    include_fighters = false;
+                } else if (part->TotalFighterDamage() && !target_ships) {
+                    retval.emplace_back(part->TotalFighterDamage()->Eval(context));
+                    // as TotalFighterDamage contains the damage from all fighters, do not further include fighter
+                    include_fighters = false;
+                } else if (part->CombatTargets() && context.effect_target && part->CombatTargets()->Eval(context, context.effect_target)) {
+                    fighter_damage = ship->CurrentPartMeterValue(SECONDARY_METER, part_name);
+                    available_fighters = std::max(0, static_cast<int>(ship->CurrentPartMeterValue(METER, part_name)));  // stacked meter
+                } else {
+                    std::vector<const Condition::Condition*> target_conditions;
```
geoffthemedio: could directly initialized `target_conditions` with a single pointer, rather than defalut init and then push

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+            --remaining_bouts;
+        }
+
+        // how much damage does a fighter shot do?
+        fighter_damage = std::max(0.0f, fighter_damage);
+
+        if (target_ships)
+            retval.emplace_back(fighter_damage * fighter_shots);
+        else
+            retval.emplace_back((float)fighter_shots);
```
geoffthemedio: is the explicit cast necessary?

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+    else
+        return WeaponDamageCalcImpl(ship, design, max, include_fighters, target_ships, CombatContextWithFighterTarget(ship));
+}
+
+
+/** Populate fighter state quantities and damages for each combat round until @p limit_to_bout */
+std::map<int, Combat::FighterBoutInfo> Combat::ResolveFighterBouts(std::shared_ptr<const Ship> ship,
+                                                           const Condition::Condition* combat_targets,
+                                                   int bay_capacity, int current_docked,
+                                                   float fighter_damage, int limit_to_bout)
```
geoffthemedio: weird indenting

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+{
+        // FIXME this should be moved to combat probably
```
geoffthemedio: meaning of comment unclear

combat/CombatDamage.cpp
```diff
@@ -0,0 +1,211 @@
...
+            // launch fighters
+            if (bout < NUM_BOUTS) {
+                retval[bout].qty.launched = std::min(bay_capacity, retval[bout].qty.docked);
+                retval[bout].qty.docked -= retval[bout].qty.launched;
+            }
+        }
+        return retval;
+}
+
+int Combat::TotalFighterShots(const ScriptingContext& context, const Ship& ship, const Condition::Condition* sampling_condition)
```
geoffthemedio, line 183: indicate purpose of function and what the Condition is for in a comment probably...

combat/CombatDamage.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "../universe/Condition.h"
+#include "../universe/EnumsFwd.h"
+#include "../universe/Ship.h"
+
+namespace Combat {
+    /** Container for return values of ResolveFighterBouts */
+    struct FighterBoutInfo {
+        struct StateQty {
+            int         docked;
```
geoffthemedio: give these all default values?

combat/CombatDamage.h
```diff
@@ -0,0 +1,35 @@
...
+#define _CombatDamage_h_
+
+#include "../universe/Condition.h"
+#include "../universe/EnumsFwd.h"
+#include "../universe/Ship.h"
+
+namespace Combat {
+    /** Container for return values of ResolveFighterBouts */
+    struct FighterBoutInfo {
+        struct StateQty {
```
geoffthemedio: is the sub-struct necessary?

combat/CombatDamage.h
```diff
@@ -0,0 +1,35 @@
...
+            int         launched;       // transitioning between docked and attacking
+        };
+        float           damage;
+        float           total_damage;   // damage from all previous bouts, including this bout
+        StateQty        qty;
+    };
+
+    /** Populate fighter state quantities and damages for each combat round until @p limit_to_bout */
+    FO_COMMON_API std::map<int, FighterBoutInfo> ResolveFighterBouts(std::shared_ptr<const Ship> ship,
+                                                               const Condition::Condition* combat_targets,
```
geoffthemedio: odd indenting

default/scripting/ship_parts/ShortRange/SR_WEAPON_1_1.focs.txt
```diff
@@ -4,6 +4,7 @@ Part
     class = ShortRange
     damage = 3
     NoDefaultCapacityEffect
+    totalFighterDamageEstimation = 0
```
geoffthemedio: why is this 0?

default/stringtables/en.txt
```diff
@@ -4081,7 +4087,10 @@ FW_FLEET_STRUCTURE_SUMMARY
 Fleet Structure - Equal to the total structure of ships in the fleet.
 
 FW_FLEET_DAMAGE_SUMMARY
-Fleet Damage - Equal to the total direct fire damage of ships in the fleet.
+Fleet Damage - Equal to the total maximum structural damage of ships and fighters in the fleet.
+
+FW_FLEET_DESTROY_SUMMARY
+Fleet Destruction - Equal to the total maximum fighter destruction of ships and fighters in the fleet.
```
geoffthemedio: I don't understand this description.

What's the difference between "fleet destruction" and "fighter destruction"?

Is it "fighter (destructions of ships and fighters)" or "(fighter destruction) of ships and fighters"?

I think "structural damage" is relatively clear, but what is "destruction"?

(I'm asking rhetorically / as the player...)

parse/DoubleValueRefParser.cpp
```diff
@@ -116,6 +117,10 @@ parse::double_parser_rules::double_parser_rules(
         =   int_rules.statistic_expr [ _val = construct_movable_(new_<ValueRef::StaticCast<int, double>>(deconstruct_movable_(_1, _pass))) ]
         ;
 
+    int_total_fighter_shots_cast
```
geoffthemedio, line 120: is an int -> double cast not automatic?

combat/CombatDamage.cpp
agrrr3: > you cannot assume the `SP_HUMAN` species exists in C++ code.

Hm I need a ship instances for evaluation. 

Damage estimation probably should not rely on target species (even if weapon combatTargets could rely on species). Else one would always need to specify target species in some way - e.g. in case of a weapon only hurting lithic species.

I could use the source's species if that sounds better.

universe/ValueRefs.cpp
```diff
@@ -1322,6 +1331,83 @@ std::string Statistic<std::string, std::string>::Eval(const ScriptingContext& co
     return max->first;
 }
 
+///////////////////////////////////////////////////////////
+// TotalFighterShots (of a carrier during one battle)    //
+///////////////////////////////////////////////////////////
+TotalFighterShots::TotalFighterShots(std::unique_ptr<Condition::Condition>&& sampling_condition) :
```
geoffthemedio: Seems unliklely that you need a whole extra ValueRef class for this, and if you do, it's probably a more general concept than what you've called it here. Can this be a type of `Statistic<int>` ?

parse/IntValueRefParser.cpp
```diff
@@ -149,16 +150,25 @@ parse::int_arithmetic_rules::int_arithmetic_rules(
           ]
         ;
 
+    total_fighter_shots
```
geoffthemedio, line 154: do you use `TotalFighterShots` in a script? If not, why is it parseable? And why does it need to be a ValueRef at all?

universe/Ship.cpp
```diff
@@ -232,12 +238,19 @@ bool Ship::IsMonster() const { // TODO: pass in ScriptingContext
         return false;
 }
 
+bool Ship::CanDamageShips(float target_shields) const {
+    return TotalWeaponsShipDamage(target_shields, true) > 0.0f;
```
geoffthemedio: move `{}` to same line with `return`

universe/ShipPart.cpp
```diff
@@ -282,6 +286,8 @@ bool ShipPart::operator==(const ShipPart& rhs) const {
     CHECK_COND_VREF_MEMBER(m_production_cost)
     CHECK_COND_VREF_MEMBER(m_production_time)
     CHECK_COND_VREF_MEMBER(m_location)
+    CHECK_COND_VREF_MEMBER(m_total_fighter_damage)
```
geoffthemedio, line 289: do these two actually affect how the part functions, or just the UI representation?

if not, it perhaps doesn't need to be checksummed...

universe/ValueRefs.h
```diff
@@ -176,6 +176,33 @@ struct FO_COMMON_API Statistic final : public Variable<T>
     std::unique_ptr<ValueRef<V>>          m_value_ref;
 };
 
+/** The variable TotalFighterShots class.   The value returned by this node is
+  * computed from the gamestate; the number of shots of a launched fighters
+  * of the source carrier is counted (and added up) for all combat bouts
```
geoffthemedio: is combat bout limits the only use of the sampling condition?

universe/ValueRefs.h
```diff
@@ -176,6 +176,33 @@ struct FO_COMMON_API Statistic final : public Variable<T>
...
+    std::string Dump(unsigned short ntabs = 0) const override;
+    void        SetTopLevelContent(const std::string& content_name) override;
+
+    const Condition::Condition* GetSamplingCondition() const
+    { return m_sampling_condition.get(); }
+
+    unsigned int GetCheckSum() const override;
+
+    std::unique_ptr<ValueRef<int>> Clone() const override {
+        return std::make_unique<TotalFighterShots>(CloneUnique(m_sampling_condition));
```
geoffthemedio: `{}` to same line as `return`

combat/CombatDamage.cpp
geoffthemedio: > Damage estimation probably should not rely on target species

I would avoid imposing such limits on conditions for targetting

> I could use the source's species if that sounds better.

if it makes sense, sure...

universe/ValueRefs.cpp
geoffthemedio: I'm also unclear why you need the sampling condition to be passed in and can't get it from parts in the design...

And the design of this ValueRef is in general odd... it seems like it should be a property of the ship and accessible with `Source.TotalFighterShots` or similar as a bound property value ref, and not a separate type of reference. It's not clear from this design what object's properties are actually being accessed...

parse/DoubleValueRefParser.cpp
agrrr3: not sure if/how automatic cast is supposed to work - but it sure does not work without that extra cast rules

default/stringtables/en.txt
agrrr3: fleet destruction is to fighter destruction what fleet damage is to structural damage.

i'm not a native speaker, i tried to use destruction only when talking about fighter targets and talking about structural damage when talking about ship targets.
Maybe a clearer term exists. 

Probably fleet destruction could be worded much better, maybe: Fleet Destruction - Equal to the maximal total destruction of fighters by the fleet.
('by the fleet' is probably a germanism)
Or Fleet Destruction - The maximum number of enemy fighters the fleet is able to destroy in a turn.

default/scripting/ship_parts/ShortRange/SR_WEAPON_1_1.focs.txt
agrrr3: A mass driver destroys exactly zero fighters per battle. If the combatTargets were not restricted, it would destroy up to four fighters.

combat/CombatDamage.h
agrrr3: Probably not I guess, ResolveFighterBouts is old code which I moved from MeterBrowseWnd.

agrrr3:
> This tooltip is unclear... (why two numbers?)
> weapon_damage_tooltip

Would changing the label suffice? Like `Structural Damage / Destroyed Fighters:  100.0 / 0`

> This ship either has 1.1k or 0.0 damage, depending whether I look at the icon or tooltip or pedia.
> 

weird. do you have save-game?

parse/IntValueRefParser.cpp
agrrr3: That is for calculating the number of fighter shots of a carrier for use inside of the damage estimation valuerefs of hangar parts. Launch patterns are too complicated for FOCS (you would need something like a loop and variables). So this valueref can calculate the correct number of shots.

For current content it is not necessary. But if one adds a e.g. close-range fighter  (so only combatTargets = bout > 2) one needs this.

universe/ValueRefs.cpp
agrrr3: I am not sure it is clear what this about (see next comment about IntValueRefParser.cpp). 

In principle, if fighter-launches were universe objects parameterized with the combat bout one could sum those up with a statistic. But that would be much more horrible.

combat/CombatDamage.cpp
agrrr3: > > Damage estimation probably should not rely on target species
> 
> I would avoid imposing such limits on conditions for targetting
> 

Yes, targeting is free to do whatever one can script up.

Different from that, damage estimation is often without context, so a weapon only targeting psionic crewed ships should still do non-zero damage when estimating fleet values.

default/scripting/ship_parts/ShortRange/SR_WEAPON_1_1.focs.txt
geoffthemedio: maybe this needs a clearer name then... like `maxFightersDestroyedPerBattleEstimate` (though that's a bit long...)

default/stringtables/en.txt
geoffthemedio: Using words consistently doesn't mean the reader will immediately pick up on the distinction when similar expressions are used with a word substituted but used in a different way. In particular, I think the use of "fighter destruction" to mean fighters destroyed and "fleet destruction" to mean destruction by a fleet is confusing.

"destruction of fighters" and "destruction of fighters by fleet" would be clearer, or just "fighter destruction" and "fighter destruction by fleet" to save a few characters.

parse/DoubleValueRefParser.cpp
geoffthemedio: OK

universe/ValueRefs.cpp
geoffthemedio: I meant use the `Statistic` framework to specify a ValueRef with a condition parameter, and add a TotalFighterShots statistic type and move all the related logic into the appropriate case when evaluating. But I guess the logic being applied doesn't really use the condition like a statitic would (to filter objects to evalute the valueref on...)

parse/IntValueRefParser.cpp
geoffthemedio: Why does this need a subcondition parameter?

I'd have thought you could do `Source.TotalFighterShots` and get all the info needed from the object whose property is being queried...

combat/CombatDamage.cpp
agrrr3: that are also remnants. design is now redundant, will only pass ship instance.

max is used to use max*meters instead of *meters.

combat/CombatDamage.cpp
agrrr3: target_conditions was a left-over; removed

combat/CombatDamage.cpp
geoffthemedio: > max is used to use max*meters instead of *meters.

please make that clearer in the code

geoffthemedio:
> > This tooltip is unclear... (why two numbers?)
> > weapon_damage_tooltip
> 
> Would changing the label suffice? Like `Structural Damage / Destroyed Fighters: 100.0 / 0`

Better.

Maybe `Structural Damage: 100.0   Destroyed Fighters: 0`

Also, are these still referring to damage / destroyed per combat round? And are these the max possible fighters destroyed, assuming all shots hit fighters, or how is that calculated?

> > This ship either has 1.1k or 0.0 damage, depending whether I look at the icon or tooltip or pedia.
> 
> weird. do you have save-game?

[stab up down save.zip](https://github.com/freeorion/freeorion/files/6500190/stab.up.down.save.zip)
affected fleet

universe/ValueRefs.cpp
agrrr3: > I'm also unclear why you need the sampling condition to be passed in and can't get it from parts in the design...

It was intended to use a simplified sampling condition as opposed to the combatTargets (e.g. visibility checks can/should be omitted for damage estimation).

> it seems like it should be a property of the ship and accessible with `Source.TotalFighterShots` or similar as a bound property value ref, and not a separate type of reference.

If the sampling condition is separate, having it as a bound property value ref is not really feasible (?).

> It's not clear from this design what object's properties are actually being accessed...

Yes, it is only in the documentation. 

Maybe rename to SourcesTotalFighterShots? or add a carrier id parameter?

or should I unpower it and always use the slower combatTargets condition? 

it could also be decoupled from a ship; you would need to pass in launch capacity, hangar capacity and sampling condition.

parse/IntValueRefParser.cpp
agrrr3: Note I intended to use the `TotalFighterShots` it in the missile prototype, but as i tightly control the launch pattern there (launching missiles only in bout 1) I actually know how many shots to expect without the `TotalFighterShots` valueref.

universe/ShipPart.cpp
agrrr3: It is supposed to be used in UI and AI

universe/ValueRefs.h
agrrr3: > for fighters i implemented both a TotalFighterShots condition = OPTIONAL_TARGET_FILTER condition (necessary to use if one uses CombatBout) and a fallback which uses the hangar part combatTargets if such exists (so the current hangar content is fine without using TotalFighterShots).

Combat bout limits is the only use case which is not handled by the fallback (using combatTargets). Mostly for performance concerns I did not want to eval the target condition for every bout in all cases but only once per carrier per combat. (The performance could be also gained by introducing BoutInvariance for valuerefs).

Also see [comment above](https://github.com/freeorion/freeorion/pull/3356#issuecomment-805274270)

universe/Ship.h
```diff
@@ -74,11 +78,15 @@ class FO_COMMON_API Ship : public UniverseObject {
...
+    float                       TotalWeaponsFighterDamage(bool include_fighters = true) const; ///< versus an fighter enemy
+    float                       TotalWeaponsShipDamage(float shield_DR = 0.0f, bool include_fighters = true) const; ///< versus an enemy ship with a given shields DR
     float                       FighterCount() const;
     float                       FighterMax() const;
-    std::vector<float>          AllWeaponsDamage(float shield_DR = 0.0f, bool include_fighters = true) const;   ///< any nonzero weapons strengths after adjustment versus an enemy with a given shields DR
-    std::vector<float>          AllWeaponsMaxDamage(float shield_DR = 0.0f, bool include_fighters = true) const;///< any nonzero weapons strengths, assuming the ship has been refueled recently, after adjustment versus an enemy with a given shields DR
+    std::vector<float>          AllWeaponsFighterDamage(bool include_fighters = true) const;   ///< any shots against enemy fighters
+    std::vector<float>          AllWeaponsShipDamage(float shield_DR = 0.0f, bool include_fighters = true) const;   ///< any nonzero weapons strengths after adjustment versus an enemy with a given shields DR
```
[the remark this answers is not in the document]
agrrr3: difference is "assuming the ship has been refueled recently," (should be "resupplied" actually)
should i write "uses Max*Meters" instead?

universe/ValueRefs.h
agrrr3: Note this would be the FOCS version when not using the fallback for strikers
```
totalFighterDamageEstimation = TotalFighterShots condition = All
totalShipDamageEstimation = TotalFighterShots condition = All * ShipPartMeter part = "FT_HANGAR_3" meter = Capacity object = Source.ID * [[FIGHTER_DAMAGE_FACTOR]]
```

agrrr3:
> > > This tooltip is unclear... (why two numbers?)
> > > weapon_damage_tooltip
> > 
> > 
> > Would changing the label suffice? Like `Structural Damage / Destroyed Fighters: 100.0 / 0`
> 
> Better.
> 
> Maybe `Structural Damage: 100.0 Destroyed Fighters: 0`
> 
> Also, are these still referring to damage / destroyed per combat round? And are these the max possible fighters destroyed, assuming all shots hit fighters, or how is that calculated?

These are currently supposed to show damage/destroyed per battle (for carriers a single bout as representative does not make sense much sense). I would like to have a bout based list somewhere.

Damage and destruction are both maximal values, so assuming the source ship/fighter are not shot down, and assuming no overshooting (at destroyed ships/fighters, not hitting with more damage than structure is available).

Of course especially for fighters (and also low structure ships), this assumption is usually wrong.

agrrr3:
> > > This ship either has 1.1k or 0.0 damage, depending whether I look at the icon or tooltip or pedia.
> > 
> > 
> > weird. do you have save-game?
> 

Ah ok, the tooltip currently only shows ship weapon parts and so the summary also only contains damage from weapon parts and not hangar parts. Probably should also show a summary for the hangar parts there(?)

agrrr3:
note i also rebased

universe/ValueRefs.cpp
geoffthemedio: Looking at it again, it could possibly be implemented as a Statistic. The condition would be your combatTargets condition, and the single value could be the ID of the ship to inspect. You'd then need a new statistic type for this specific use, which would be handled differently in `Statistic<T, V>::Eval` at the start, by not doing a search for condition matches or a call to `ReduceData<T, V>`.

Not sure if that's worth the complication / limits, though...

geoffthemedio:
> These are currently supposed to show damage/destroyed per battle. [...] Damage and destruction are both maximal values, so assuming the source ship/fighter are not shot down, and assuming no overshooting (at destroyed ships/fighters, not hitting with more damage than structure is available).

The tooltip text needs to be elaborated to make this clear... probably also indicating the number of combat rounds... maybe like "For 4 bout battle, total structure damage: XXX  total fighter destruction: Y"

universe/Ship.h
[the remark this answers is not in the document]
geoffthemedio: maybe, and also add to the non-max case further explanation that it might be less than max if the ship has less than a full complement of fighters

universe/ShipPart.cpp
geoffthemedio: I guess if it's used in the AI then it can be included in the calculation

universe/ValueRefs.cpp
agrrr3: > Looking at it again, it could possibly be implemented as a Statistic. 

Would you imagine usage like this?
```
totalFighterDamageEstimation = Statistic TotalFighterShots value = Source.ID condition = And [ [[COMBAT_TARGETS_VISIBLE_ENEMY]] Fighter ]
```
I do not see what is the benefit of doing that. It would be more flexible, because it could technically be passed more than a single object, but then actually it is probably a sum/needs a statistic type.

Also the misuse of the value parameter would be quite confusing.

The following would be probably closer to usual Statistic usage, but still calculating a Sum over multiple bouts instead of over multiple filtered objects 
```
totalFighterDamageEstimation = CombatStatistic Sum value = Source.LaunchedFightersCount condition = And [ [[COMBAT_TARGETS_VISIBLE_ENEMY]] Fighter ]
```
Note this assumes there is at maximum one candidate, else one would need something like  `...condition = NumberOf number = 1 condition = And ...`

default/scripting/ship_parts/ShortRange/SR_WEAPON_1_1.focs.txt
agrrr3: made it destroyFightersPerBattleMax and damageStructurePerBattleMax

parse/IntValueRefParser.cpp
agrrr3: Added a parameter `carrier = SHIP_ID` for specifying the subject of the valuerref (instead of implicitly using the source)

Cjkjvfnby:
There is how the shipping damage contributes to the AI: https://github.com/freeorion/freeorion/blob/694a408a29a784fd83094022125960f3eecd5578/default/python/AI/CombatRatingsAI/_ship_combat_stats.py#L80

It takes current design stats and the best enemy shield. 
Attacks are a map from weapon to its number. This code is used when the design is not created, we try to find the best design with weapons/pilots we could.  

So looks like the API that we need could be the following:

```python
def get_ship_damage(hull_name, part_names, enemy_sields):
```

I am not sure if the hull affects damage.

PS.  hull_name, part_names could have the same type as `issueCreateShipDesignOrder` (https://github.com/freeorion/freeorion/blob/694a408a29a784fd83094022125960f3eecd5578/client/AI/AIWrapper.cpp#L537)

agrrr3:
> So looks like the API that we need could be the following:
> ```python
> def get_ship_damage(hull_name, part_names, enemy_sields):
> ```
> 
> I am not sure if the hull affects damage.

Hull doesnt effect damage currently, but in principle it could, so i guess the AI should send as much information as possible. 

Actually the target heavily influences the outcome (e.g. if the enemy is expected to include fighters) so it would be good to send the whole enemy combat stats, not only the shield value. 

Or we only expose the *Damage calculation per part and keep the combined calculation in python as needed.

@Cjkjvfnby Anyway we should open a separate issue (and related PR) for AI and probably discuss it rather in the forums it is not lost if we switch away from github one day.

Lets merge this PR ASAP - I want to get some more feedback from players.

agrrr3:
@geoffthemedio can we merge this?

geoffthemedio:
A remaining minor problem is the "Total Weapon Damage" (key: "ATTACK") in the objects list, which is created here: https://github.com/freeorion/freeorion/blob/master/UI/ObjectListWnd.cpp#L279 which is broken because the "Attack" value-ref has been removed, so just show 0.0 for all ships.

Can there instead be columns for total ship damage and total fighter destruction or similar?

PR #3357, opened 2021-03-13
Add option to allow human player to take over any AI

o01eg:


default/stringtables/en.txt
```diff
@@ -1515,6 +1515,9 @@ Limit of client message size in bytes.
 OPTIONS_DB_DROP_EMPIRE_READY
 Drop empire's readiness on joining to the playing game.
 
+OPTIONS_DB_TAKE_OVER_AI
+Allow human player to delegate any AI player.
```
geoffthemedio: maybe clarify that this is just a server option...

o01eg: Fixed

default/stringtables/en.txt
```diff
@@ -1515,6 +1515,9 @@ Limit of client message size in bytes.
 OPTIONS_DB_DROP_EMPIRE_READY
 Drop empire's readiness on joining to the playing game.
 
+OPTIONS_DB_TAKE_OVER_AI
+Server option to allow human player to delegate any AI player.
```
geoffthemedio: I would interpret this as saying that the human player can say that any AI player can replace the player.

I think you mean that this option would allow any human player to replace any AI player?

o01eg: Yes, that what I mean.

geoffthemedio: Is a player with no assigned empire automatically assigned an AI empire if there is one or multiple AI empires, or do they get put into the lobby in order to pick one?

o01eg: Player without assigned empire will automatically be assigned to a AI empire if there is only one AI empire. This behaviour could be changed.

geoffthemedio: What if there are multiple AI empires?

o01eg: Player will get to lobby to choose.

geoffthemedio: OK, then I guess
````
OPTIONS_DB_TAKE_OVER_AI
Server option to allow human players to take control of AI empires.
````

Comment (o01eg) [reply default/stringtables/en.txt]: Changed.

**Comment by o01eg** [comment]
@geoffthemedio I also change empire's player name so after reloading game player could still connect to the empire.

File: server/ServerApp.cpp [hunk 3]
```diff
@@ -1932,6 +1941,10 @@ int ServerApp::AddPlayerIntoGame(const PlayerConnectionPtr& player_connection, i
             const std::string previous_player_name = (*previous_it)->PlayerName();
             m_networking.Disconnect(previous_player_id);
             if (previous_client_type == Networking::ClientType::CLIENT_TYPE_AI_PLAYER) {
+                // change empire's player so after reload the player still could connect
+                // to the empire
```
geoffthemedio, line 1945: Does this let them reconnect to the same empire, or does it prevent other players from connecting to that empire? If you didn't set the name, couldn't the same player just reconnect to the same empire anyway...?

o01eg: Empire lost its AI client type so other player couldn't connect to it, but after reload empire will be started without assigned AI process so if it continue to have old `AI_#` player name no one could connect to it.

Issue #3350, opened 2021-03-10
Add title screen song that is different from the in-game song


magnusviri:
Feature request
===============

Idea
----

Please add something in the title screen code that looks for a different song than the current one. If it's not defined, have it go back to the current behavior (where it plays only one song). This allows the option of adding a title screen song, but if a title screen song doesn't show up, then it wont be silent.

Gain
----

A title screen song would be nice. I hope to write it. I haven't written it yet.

Discussion
----------

[Forum link](https://www.freeorion.org/forum/viewtopic.php?f=6&t=11959)

PR #3358, opened 2021-03-13
Separate title and in-game music track options

geoffthemedio:
Implements https://github.com/freeorion/freeorion/issues/3350

### DIFF
--- a/UI/ClientUI.cpp
+++ b/UI/ClientUI.cpp
@@ -492,7 +492,8 @@ namespace {
         db.Add("video.fps.unfocused",                                   UserStringNop("OPTIONS_DB_MAX_FPS_NO_FOCUS"),               15.0,                           RangedStepValidator<double>(0.125, 0.125, 30.0));
 
         // sound and music
-        db.Add<std::string>("audio.music.path",                         UserStringNop("OPTIONS_DB_BG_MUSIC"),                       (GetRootDataDir() / "default" / "data" / "sound" / "artificial_intelligence_v3.ogg").string());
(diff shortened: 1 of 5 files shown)


geoffthemedio:
Linux and MacOS test reports would be good... That it compiles doesn't really ensure that it works.

geoffthemedio:
Issue: music doesn't restart if toggled off and back on in the options

Vezzra:
> Linux and MacOS test reports would be good

Works on macOS.

> Issue: music doesn't restart if toggled off and back on in the options

Same on macOS.

magnusviri:
When it starts it plays the theme song for about 2 seconds, stops, then it starts over when the title screen shows. So it's calling the song twice. I don't think it did this before, did it?

hifron:
On current master there is stopped audio already:

```
[2022-06-16 23:54:40.188142] [0x000001cc] [info]    NamedValueRefManager::GetValueRef found no registered double valueref for "AUGMENTATION_FULL_GROWTH_INFRASTRUCTURE_REQUIREMENT". This should not happen once "#3225 Refactor initialisation of invariants in value refs to happen after parsing" is implemented
[2022-06-16 23:54:40.203755] [0x0000218c] [info]    Audio disabled.
[2022-06-16 23:54:40.203755] [0x000001cc] [info]    NamedRef<T>::NamedRefInitInvariants() could not find value ref, will sleep a bit and retry.
```

Should be fixed by https://github.com/freeorion/freeorion/pull/3922

hifron:
@magnusviri ping

magnusviri:
I have no idea how to test this mainly because it's been over a year since I built it and I don't remember how to do it. I was going to try to use the GitHub Actions to build it for me but I'm not quite sure how to make sure both https://github.com/freeorion/freeorion/pull/3922 and https://github.com/freeorion/freeorion/pull/3358 are included in the build. I imagine I have to rebase or something. I'm a lot better with git than I was a year ago but I'm not that advanced yet.

PR #3365, opened 2021-03-28
Move single content files to appropriate folders

o01eg:
Actually I need it to export required game content for Android with only two folders.

I could pick all those files with theirs paths but reorganization of location looks simpler.

Also I deleted `customizations` folder because it lost its customization behavior after introducing checksum.


geoffthemedio:
Seems fine...

 `content_specific_parameters.txt` isn't really stringtable-related and should probably be replaced with a few client-side options, but that's a separate issue.

I think the customizations are just meant for single player games, for which the checksum would be the same, not multiplayer games, as they need to run on the server to generate sitreps, so client-side-only modifications aren't expected to work for the multiplayer case.

Vezzra (commented):
Can't comment on the code changes, but does not break anything on macOS.

PR #3369, opened 2021-03-29
Fix first turn advance time without fixed turn interval

o01eg:


geoffthemedio:
if you've tested this, then ok...?

o01eg:
Yes

PR #3371, opened 2021-03-29
Test number of all parsed techs

o01eg:
Checks if all techs are parsed without check of internal structure

### DIFF
--- a/test/parse/TestScriptParser.cpp
+++ b/test/parse/TestScriptParser.cpp
@@ -192,5 +192,22 @@ BOOST_AUTO_TEST_CASE(parse_techs) {
     BOOST_REQUIRE_EQUAL(2, categories_seen.size());
 }
 
+BOOST_AUTO_TEST_CASE(parse_techs_full) {
+    auto scripting_dir = boost::filesystem::system_complete(GetBinDir() / "default/scripting");
+    BOOST_REQUIRE(scripting_dir.is_absolute());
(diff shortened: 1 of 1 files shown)


test/parse/TestScriptParser.cpp
```diff
@@ -192,5 +192,22 @@ BOOST_AUTO_TEST_CASE(parse_techs) {
...
+    BOOST_REQUIRE(boost::filesystem::exists(scripting_dir));
+    BOOST_REQUIRE(boost::filesystem::is_directory(scripting_dir));
+
+    Pending::Pending<TechManager::TechParseTuple> techs_p = Pending::StartAsyncParsing(parse::techs<TechManager::TechParseTuple>, scripting_dir / "techs");
+    auto [techs, tech_categories, categories_seen] = *Pending::WaitForPendingUnlocked(std::move(techs_p));
+    BOOST_REQUIRE(!techs.empty());
+    BOOST_REQUIRE(!tech_categories.empty());
+    BOOST_REQUIRE(!categories_seen.empty());
+
+    BOOST_REQUIRE_EQUAL(207, techs.size());
```
geoffthemedio, line 207: Is this just planned to be used during FOCS -> Python transition? Obviously difficult to keep up to date if adding / removing techs later...

o01eg: Yes, while both FOCS and Python parser coexists it is better to check we don't miss techs. It won't affect CI because tests allowed to fail.

PR #3375, opened 2021-03-31
Detect boost python library from python library instead of python executable

### LINKED ISSUE #3334: Build error with 0.4.10.1: cmake cannot find boost_python library from boost 1.61.0
Bug Report
==========
Trying to do a local build of freeorion 0.4.10.1, from the sources from github.
The build fails at cmake configure stage with the error message: "CMake Error at cmake/FindBoost.cmake:2161 (message):
Unable to find the requested Boost libraries. Boost version: 1.61.0 Boost include path: /usr/include 
Could not find the following Boost libraries: boost_python35
No Boost libraries were found."

Environment
-----------
* **FreeOrion Version**: 0.4.10.1
* **Operating System**: Rosa Linux Fresh 2016.1 (a Mandriva derivative) x86_64
* **Graphic card used**:  Radeon RX460
* **Fetched as**: Compiled from source
* **Compiler**: gcc 5.5.0
* **Boost release installed**: 1.61.0
* **Python releases installed**: 3.5.9 and 2.7.14
* **CMake release installed**: 3.16.2

Description
-----------
Cmake command line used:
*********************************
/usr/bin/cmake .. -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_INSTALL_LIBDIR:PATH=/usr/lib64 -DINCLUDE_INSTALL_DIR:PATH=/usr/include -DLIB_INSTALL_DIR:PATH=/usr/lib64 -DSYSCONF_INSTALL_DIR:PATH=/etc -DSHARE_INSTALL_PREFIX:PATH=/usr/share -DLIB_SUFFIX=64 -DCMAKE_SKIP_RPATH:BOOL=ON -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DBUILD_SHARED_LIBS:BOOL=ON -DBUILD_STATIC_LIBS:BOOL=OFF '-DCMAKE_MODULE_LINKER_FLAGS= -Wl,--as-needed -Wl,--no-undefined -Wl,-z,now -Wl,-z,relro -Wl,-O1 -Wl,--build-id -Wl,--enable-new-dtags -Wl,--hash-style=gnu -Wl,--unresolved-symbols=ignore-all' -DCMAKE_BUILD_TYPE=Release -DLIB_SUFFIX=64 -DBUILD_SHARED_LIBS=OFF -DENABLE_TIFF_TEXTURES=ON
*********************************

In my system Boost 1.61.0 libraries are in /usr/lib64 and header files in /usr/include/boost; the installed python3 libraries in /usr/lib64 are: lib64boost_python3_1.61.0 and libboost_python3.so (there are also lib64boost_python1_1.61.0 and libboost_python1.so, supporting python2).

Cmake searches for and finds various boost libraries (filesystem, iostreams, locale, log, log_setup, serialization, system, thread, regex, date_time, chrono and  atomic), but fails to find the libboost_python3 ones with this error message:
***********************************
CMake Error at cmake/FindBoost.cmake:2161 (message):
Unable to find the requested Boost libraries.
Boost version: 1.61.0
Boost include path: /usr/include
Could not find the following Boost libraries:
         boost_python35
No Boost libraries were found.  You may need to set BOOST_LIBRARYDIR to the directory containing Boost libraries or BOOST_ROOT to the location of Boost.
Call Stack (most recent call first):
CMakeLists.txt:240 (find_package)
***********************************

The culprit is at the lines 240-241 of the main CMakeLists.txt file:
*********************
find_package(Boost COMPONENTS python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR} REQUIRED)
set(Boost_PYTHON_SUFFIX "${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
*********************
AFAICT the last line instructs cmake to search for a library called libboost_python35, while the installed one is name libboost_pyhton3.
Since the minimum boost version supported is set to "1.58.0" (see in the line 223 of the main CMakelists.txt file), that statement looks wrong to me: as stated in the initial comment of the provided FindBoost.cmake file (lines 44-51), the boost python libraries have a version suffix in the form of major.minor only in boost release 1.67.0 and above... In boost <= 1.66.0 the version suffix is distribution specific: in my case they are "1" for python2 and "3" for python3.

The issue can be easily worked around by patching the main CMakeLists.txt file and cmake/make_versioncpp.py to remove the mention of the minor version number (see attached patch), but this in turn will cause trouble for any boost >= 1.67 and cannot be the proper solution... As an alternative, the minimum boost version required should be bumped to "1.67.0".
[freeorion-0.4.10.1-fix-boost_python-detection.txt](https://github.com/freeorion/freeorion/files/6054460/freeorion-0.4.10.1-fix-boost_python-detection.txt)

Expected Result
---------------
The build continues.

Steps to reproduce
------------------
* Install the 0.4.10.1 sources.
* Run cmake with the command line reported above.
* Look at the error message.

o01eg:
Fix #3334 

Allows to have different python library and python executable in case of cross-compilation.

### DIFF
--- a/cmake/make_versioncpp.py
+++ b/cmake/make_versioncpp.py
@@ -13,22 +13,6 @@ from glob import glob
 
 INVALID_BUILD_NO = "???"
 
-required_boost_libraries = [
-    "boost_chrono",
-    "boost_date_time",
-    "boost_filesystem",
-    "boost_iostreams",
-    "boost_locale",
-    "boost_log",
-    "boost_log_setup",
-    "boost_python%d%d" % (sys.version_info.major, sys.version_info.minor),
-    "boost_regex",
-    "boost_serialization",
-    "boost_signals",
-    "boost_system",
-    "boost_thread",
-]
-
 
 class Generator(object):
     def __init__(self, infile, outfile):
@@ -105,13 +89,34 @@ class NsisInstScriptGenerator(Generator):
                 FreeOrion_PYTHON_VERSION="")
 
 
-if 3 != len(sys.argv):
+if len(sys.argv) not in (3, 4):
     print("ERROR: invalid parameters.")
-    print("make_versioncpp.py <project rootdir> <build system name>")
+    print("make_versioncpp.py <project rootdir> <build system name> [<boost python suffix>]")
     quit()
 
 os.chdir(sys.argv[1])
 build_sys = sys.argv[2]
+if 4 == len(sys.argv):
+    boost_python_suffix = sys.argv[3]
+else:
+    boost_python_suffix = "%d%d" % (sys.version_info.major, sys.version_info.minor)
+
+required_boost_libraries = [
+    "boost_chrono",
+    "boost_date_time",
+    "boost_filesystem",
+    "boost_iostreams",
+    "boost_locale",
+    "boost_log",
+    "boost_log_setup",
+    "boost_python%s" % boost_python_suffix,
+    "boost_regex",
+    "boost_serialization",
+    "boost_signals",
+    "boost_system",
+    "boost_thread",
+]
+
 
 # A list of tuples containing generators
 generators = [
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,6 +231,9 @@ find_package(Threads)
 find_package(PythonInterp ${MINIMUM_PYTHON_VERSION} REQUIRED)
 if(BUILD_AI OR BUILD_SERVER)
     find_package(PythonLibs ${MINIMUM_PYTHON_VERSION} REQUIRED)
+    message(STATUS "Python library version detected ${PYTHONLIBS_VERSION_STRING}")
+    string(REGEX REPLACE "^([0-9]+)\\.([0-9]+)\\..*" "\\1\\2" Boost_PYTHON_SUFFIX "${PYTHONLIBS_VERSION_STRING}")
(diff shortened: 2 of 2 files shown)


cmake/make_versioncpp.py
```diff
@@ -105,13 +89,34 @@ def compile_output(self, template, version, branch, build_no, build_sys):
                 FreeOrion_PYTHON_VERSION="")
 
 
-if 3 != len(sys.argv):
+if 3 != len(sys.argv) and 4 != len(sys.argv):
```
Cjkjvfnby: Probably it is time to start using [argparse](https://docs.python.org/3/library/argparse.html#module-argparse). But we could do it later, after merging these changes.

```suggestion
if len(sys.argv) not in (3, 4):
```

o01eg: Fixed.

hifron:
This could be Enterprise edition of Python like is in Windows with various systems environments... Which should not be guaranteed to have... But in game industry there should be some support(maybe in future) in paid games(like in distribution for Windows games on Linux(currently very buggy and not recommended, need development and maybe also commercial development, but not currently lucrative)).

In others there should be some negotiation between vendors, FO? and who uses that on something... Maybe this needs a lot of time to be fixed to have proper solution...

PR #3378, opened 2021-04-02
Fix multiplayer issues

o01eg:
Correctly get revious player id. Before could get same player just connected.

Fix crash if connection to the game could cause switch to lobby.


geoffthemedio:
seems fine...

PR #3380, opened 2021-04-03
[AI] Simplify enums

Cjkjvfnby:
- Fix an issue when enum item from other class could be accepted by has_items
- Remove range method
- Simplify metaclass
- Add documentation
- Add tests

It was not clear for me how to use some enum features so I did some research and document result in code and tests.


geoffthemedio:
seems fine... want to merge...?

PR #3384, opened 2021-04-04
Remove unused enum value from EnumWrapper

Cjkjvfnby:
AI does not use these values. Entries in *.pyi could be ignored, I'll deal with them on the next stubs update.

### DIFF
--- a/python/EnumWrapper.cpp
+++ b/python/EnumWrapper.cpp
@@ -97,7 +97,6 @@ namespace FreeOrionPython {
             .value("targetConstruction",MeterType::METER_TARGET_CONSTRUCTION)
             .value("targetHappiness",   MeterType::METER_TARGET_HAPPINESS)
 
-            .value("maxDamage",         MeterType::METER_MAX_CAPACITY)
             .value("maxCapacity",       MeterType::METER_MAX_CAPACITY)
             .value("maxSecondaryStat",  MeterType::METER_MAX_SECONDARY_STAT)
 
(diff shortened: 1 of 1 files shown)


geoffthemedio:
sure, if you prefer...

Cjkjvfnby:
> sure, if you prefer...

It is not a preference, `MeterType::METER_MAX_CAPACITY` is bind to the 2 enum fields. Probably some artifact of the past.

```c++
.value("maxDamage",         MeterType::METER_MAX_CAPACITY)
.value("maxCapacity",       MeterType::METER_MAX_CAPACITY)
```

o01eg:
@Cjkjvfnby It looke like typo and should use `MeterType::METER_MAX_DAMAGE`.

Cjkjvfnby:
> It looks like typo and should use MeterType::METER_MAX_DAMAGE

Actually, it is more complex. `MeterType::METER_MAX_DAMAGE` does not exist anymore.  I assume this piece of code is left for backward compatibility. We could remove it now.

https://github.com/freeorion/freeorion/commit/378c1703cdb04a093e379a03775e5ee73c056c44

geoffthemedio:
It's not an error or legacy code; it is intentional and consistent with the FOCS API. The "capacity" meter is also used as weapon damage for weapon parts, so the alternative name for that stat was also provided.

If you prefer not to do that for the Python API, it's fine, though.

Cjkjvfnby:
> If you prefer not to do that for the Python API, it's fine, though.

Sometimes, duplication produces more confusion than benefits. So I'd prefer to have only one enum item with this value.

geoffthemedio:
> If you prefer not to do that for the Python API, it's fine

PR #3397, opened 2021-04-11
Build Location Regional Administration

Grummel7:
Only distance to own capital or regional admin should be considered for build location


Grummel7:
I reverted the second commit. Wasn't aware that any commit to the same branch is automatically added to the pull request.

geoffthemedio:
Please reset to the earlier commit, rather than adding a new revert commit immediately after the unintentional commit. Revert commits are fine if there is a bunch of stuff between them, but in this pull request there is nothing between so you can just remove it from the branch commit log.

Grummel7:
Sorry, still learning to use git properly.
Looks better now.

geoffthemedio:
looks OK.

still need to add some exploit mitigation for swapping planets between allies to avoid these restrictions, though.

PR #3402, opened 2021-04-13
Refactored distance to capital stability

Grummel7:
Fixed the happiness calculation for distance from capital or regional administration.
It should use the minimum of all supply connected candidates.
Also used to named refs to make it easier to maintain.

### DIFF
--- a/default/scripting/species/common/happiness.macros
+++ b/default/scripting/species/common/happiness.macros
@@ -71,7 +71,7 @@ COMMON_HAPPINESS_EFFECTS
 
 SUPPLY_CONNECTED_TO_CAPITAL_OR_REGAD
 '''
-		// this should ensure that there is a capital / admin for JUMPS_TO_CAPITAL_OR_REGAD
+        // this should ensure that there is a capital / admin for JUMPS_TO_CAPITAL_OR_REGAD
         ResourceSupplyConnected empire = Source.Owner condition = And [
             Or [Capital Building name = "BLD_REGIONAL_ADMIN"]
             OwnedBy empire = Source.Owner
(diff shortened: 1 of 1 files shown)


default/scripting/species/common/happiness.macros
```diff
@@ -69,84 +69,62 @@ COMMON_HAPPINESS_EFFECTS
...
+JUMPS_TO_CAPITAL_OR_REGAD
+'''
+		Statistic Min
+		    // Replace JumpsBetween by JumpsBetweenByEmpireSupplyConnections once it is implemented
+            value = JumpsBetween
+            object = Source.ID
+            object = LocalCandidate.ID
+            condition = And [
+                Or [Capital Building name = "BLD_REGIONAL_ADMIN"]
+                OwnedBy empire = Source.Owner
```
geoffthemedio, line 90: why not use `SUPPLY_CONNECTED_TO_CAPITAL_OR_REGAD` here?

Grummel7: Which part do you want to replace, the lines above your comment or the one line below?
The condition must identify the Capital and all RAs, but only if they are connected to the Source, i.e. the planet to be affected by the happiness modification. The Capital itself is of course always supply connected to the Capital.

default/scripting/species/common/happiness.macros
```diff
@@ -69,84 +69,62 @@ COMMON_HAPPINESS_EFFECTS
             ]
 '''
 
+SUPPLY_CONNECTED_TO_CAPITAL_OR_REGAD
+'''
+		// this should ensure that there is a capital / admin for JUMPS_TO_CAPITAL_OR_REGAD
```
geoffthemedio: replace tabs with spaces everywhere

default/scripting/species/common/happiness.macros
geoffthemedio: Actually, neither. I was misreading the structure of the condition in the macro above due to linebreaks, in part due to the tab characters causing a lot of indentation.

Grummel7: Oh yeah, fun with tabs. At work we use an SVN hook to stop any check-in containing a tab.
The default settings for text editors in eclipse apparently create them. I changed the settings and updated the commit.

PR #3404, opened 2021-04-14
Add cache directory access

o01eg:
Adds access to cache directory provided with `XDG_CACHE_HOME`.

Mostly intended for Android to extract standard python library into filesystem.


geoffthemedio:
Nothing to comment on for me, I think...

PR #3339, opened 2021-03-02
Fixes for Godot deb packaging

o01eg:
Some fixes I made to prepare .deb packages

Branch with deb package files is https://github.com/o01eg/freeorion/tree/ppa-godot

It builds for Ubuntu 20.04 Focal LTS but I'm not yet tested it.

PR #3405, opened 2021-04-14
Add options to set path to freeoriond executable

o01eg:
Could be useful if it cannot be detected automatically like https://github.com/freeorion/freeorion/pull/3339#issuecomment-818591171

### DIFF
--- a/client/human/GGHumanClientApp.cpp
+++ b/client/human/GGHumanClientApp.cpp
@@ -433,7 +433,7 @@ void GGHumanClientApp::StartServer() {
         throw LocalServerAlreadyRunningException();
     }
 
-    std::string SERVER_CLIENT_EXE = GetOptionsDB().Get<std::string>("freeoriond.path");
+    std::string SERVER_CLIENT_EXE = GetOptionsDB().Get<std::string>("misc.server-local-binary.path");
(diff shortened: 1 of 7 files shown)


o01eg:
Could it be merged? We solved related issue other way but this PR could be helpful too.

client/human/GGHumanClientApp.cpp
```diff
@@ -443,7 +433,7 @@ void GGHumanClientApp::StartServer() {
         throw LocalServerAlreadyRunningException();
     }
 
-    std::string SERVER_CLIENT_EXE = ServerClientExe();
+    std::string SERVER_CLIENT_EXE = GetOptionsDB().Get<std::string>("freeoriond.path");
```
geoffthemedio: option should probably be something like `misc.server-local-binary.path`.

(I assume this can't ever be used to start a remote server)

o01eg: No, I suppose support of remote process execution via ssh or something is outside of FreeOrion scope.

o01eg: Renamed option.

geoffthemedio:
seems OK

PR #3408, opened 2021-04-16
Use Github Actions for Ubuntu builds

o01eg:


geoffthemedio:
Can you make the various Ubuntu builds have names that appear more different at the start of the names, so they are more distinct in lists where the ends are truncated?
ubuntu builds in list

geoffthemedio:
And could we leave the TravisCI builds running for a bit, so I can get a sense of how they compare?

o01eg:
@geoffthemedio It takes name from workflow name and adds jobs name

I could make it something like "Ubuntu / 18.04 Build FreeOrion ..." and "Ubuntu / 20.04 Build FreeOrion ...".

geoffthemedio:
> I could make it something like "Ubuntu / 18.04 Build FreeOrion ..." and "Ubuntu / 20.04 Build FreeOrion ...".

Looks like it would do what I want.

o01eg:
@geoffthemedio Chaged it.

geoffthemedio:
Are you manually cancelling the AppVeyor builds for some reason...?

otherwise seems fine...

o01eg:
@geoffthemedio Yes, because it unparalleled and slows down check for other PRs.

PR #3409, opened 2021-04-16
Use Github Actions for Windows CMake build

o01eg:


geoffthemedio:
no comment on most, and no objections to adding an include you need

.github/workflows/build-windows.yml
```diff
@@ -0,0 +1,49 @@
+name: "Windows build"
+
+on:
+  push:
```
Cjkjvfnby, line 4: Should we really build it on each push?

As I recall Geoff uses this reap to push his development branches and probably all other forks will be affected.

.github/workflows/build-windows.yml
```diff
@@ -0,0 +1,49 @@
...
+    name: Build FreeOrion on Windows
+    runs-on: windows-2019
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - name: Download SDK
+        id: download-sdk
+        uses: suisei-cn/actions-download-file@v1
+        with:
+          url: https://github.com/freeorion/freeorion-sdk/releases/download/v11/FreeOrionSDK_11_MSVC-v141-Win32.zip
```
Cjkjvfnby, line 32: Should we cache it?  https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows  ()

o01eg: Hard to say if Github Cache quicker than Github Releases file. It better to have something like ccache for MSVS.

.github/workflows/build-windows.yml
o01eg: Current Travis, Appveyor and Github CI build project on each push.

geoffthemedio: Is there a reason not to?

Vezzra (commented):
Can't really provide any meaningful input... as far as I can tell, looks fine...

geoffthemedio:
I suppose the caching issue can be discussed separately...

PR #3411, opened 2021-04-16
Use Github Actions for MacOS build

o01eg:


.github/workflows/build-macos.yml
```diff
@@ -0,0 +1,56 @@
...
+      - 'check/**'
+      - 'doc/**'
+      - 'packaging/**'
+      - 'snap/**'
+      - '*.md'
+
+jobs:
+  build-macos:
+    name: Build FreeOrion on MacOS
+    runs-on: macos-10.15
```
Vezzra: This needs to be at least `macos-10.14`. We need to maintain backward compatibility at least for that macOS version.

Vezzra (dismissed):
I don't understand a lot of the stuff here, what I do understand looks fine.

The min supported macOS version needs to be 10.14 at least though.

.github/workflows/build-macos.yml
o01eg: The only possible `macos` choice on Github Actions is `macos-11.0` and `macos-10.15` https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

Vezzra: In that case, can we use `macos-11.0`? Or wouldn't that work?

Providing backward compatibility only for the two most recent macOS versions is a big problem IMO...

o01eg: > In that case, can we use macos-11.0? Or wouldn't that work?

I'll try it.

> Providing backward compatibility only for the two most recent macOS versions is a big problem IMO...

Yes, so we cannot remove MacOS build from Travis CI.

o01eg: Looks like Github either renamed it, or deleted it:

```
No runner matching the specified labels was found: macos-11.0.
```

Vezzra: Oops, sorry, `macos-11.0` wouldn't be any good anyway, I confused that with `macos-10.11`. If GitHub offers no older macOS version than 10.15, I guess we'll have to go with that.

Vezzra (approved):

PR #3412, opened 2021-04-17
Fixed Exobot Productivity effect

Grummel7:


default/scripting/policies/EXOBOT_PRODUCTIVITY.focs.txt
```diff
@@ -19,9 +19,10 @@ Policy
                 TargetIndustry low = 0
                 Focus type = "FOCUS_INDUSTRY"
             ]
+            // Very bad multiplier is applied before, so it must be reverted before applying bad multiplier.
             effects = SetTargetIndustry value = Value
                         * (NamedReal name = "PLC_EXOBOT_PRODUCTIVITY_TARGET_INDUSTRY_SCALING"
```
Grummel7, line 24: Note sure what the value of this NamedReal should be.
With the actual values of the multipliers it becomes 1.5, changing TargetIndustry from 0.5 * base production to 0.75 * base production

geoffthemedio:
Could you clarify in the description what the issue is / was, including a link to an existing issue, if there is one, or a forum post?

Grummel7:
Perhaps my comment distracts from the actual change.
The problem is that the old code just multiplies Production with (0.75 - 0.5) = 0.25 resulting in even less production. The code now does what the Pedia article says it should, modifying the production as if it were only bad.

My only Problem is that I do not know what the NamedReal is supposed to be.

default/scripting/policies/EXOBOT_PRODUCTIVITY.focs.txt
agrrr3: > My only Problem is that I do not know what the NamedReal is supposed to be.
 
It defines a name for the calculation to be used in the description. Not used currently, but you could add similar to `In effect, the target industry gets multiplied by [[value PLC_EXOBOT_PRODUCTIVITY_TARGET_INDUSTRY_SCALING]]`

And maybe move the division to the end to be a little bit more clear ([[BAD_MULTIPLIER]] / [[VERY_BAD_MULTIPLIER]]). The tooltip would read something like `0.75 / 0.5`

Grummel7: The problem with these multiplying effects is that you have to know in which place it happens.
E.g. Industrialism only multiplies with 1.25, but may add a lot more, because it does so after a lot of other bonuses are added.

Perhaps re-writing the effect in absolute terms makes a better fix:
`
            effects = SetTargetIndustry value = Value + Target.Population
                        * (NamedReal name = "PLC_EXOBOT_PRODUCTIVITY_TARGET_INDUSTRY_SCALING"
                                     value = ([[BAD_MULTIPLIER]] - [[VERY_BAD_MULTIPLIER]]) * [[INDUSTRY_PER_POP]])
`
That would also make the calculation less dependent on the order of effects.

Haven't notice the tooltip feature in the Pedia yet. Cool!

geoffthemedio: The design of the good and bad industry or other resource output traits is as a multiplicative adjustment. The result of them is thus intended to be after some other effects and to also multipy those. Changing how that works is possible, but out of the scope of a fix like this.

default/scripting/policies/EXOBOT_PRODUCTIVITY.focs.txt
```diff
@@ -19,9 +19,10 @@ Policy
...
-                                     value = [[BAD_MULTIPLIER]] - [[VERY_BAD_MULTIPLIER]])
+                                     value = (1 / [[VERY_BAD_MULTIPLIER]]) * [[BAD_MULTIPLIER]])
```
geoffthemedio: Can this just be
````
            effects = SetTargetIndustry value = Value
                        * (NamedReal name = "PLC_EXOBOT_PRODUCTIVITY_TARGET_INDUSTRY_SCALING"
                                     value = [[BAD_MULTIPLIER]] / [[VERY_BAD_MULTIPLIER]])
````

Comment (Grummel7) [reply default/scripting/policies/EXOBOT_PRODUCTIVITY.focs.txt]: Of course, don't know why I did not write it like that in the first place.
Does it help when I suggest the change here?
```suggestion
                                     value = ([[BAD_MULTIPLIER]] / [[VERY_BAD_MULTIPLIER]])
```

geoffthemedio: seemingly yes, there is a button

geoffthemedio:
There's a parse error...
````
ReportParseError.cpp:92 : /home/travis/build/freeorion/freeorion/build/default/scripting/policies/EXOBOT_PRODUCTIVITY.focs.txt:48:4: Parse error.  Expected ) here:
             ]
             // Very bad multiplier is applied before, so it must be reverted before applying bad multiplier.
             effects = SetTargetIndustry value = Value
                         * (NamedReal name = "PLC_EXOBOT_PRODUCTIVITY_TARGET_INDUSTRY_SCALING"
                                      value = (0.75 / 0.5)
     ]
     ^
     graphic = "icons/policies/exobot_productivity.png"
````

File: default/scripting/policies/EXOBOT_PRODUCTIVITY.focs.txt [hunk 3]
```diff
@@ -19,9 +19,10 @@ Policy
                 TargetIndustry low = 0
                 Focus type = "FOCUS_INDUSTRY"
             ]
+            // Very bad multiplier is applied before, so it must be reverted before applying bad multiplier.
             effects = SetTargetIndustry value = Value
                         * (NamedReal name = "PLC_EXOBOT_PRODUCTIVITY_TARGET_INDUSTRY_SCALING"
-                                     value = [[BAD_MULTIPLIER]] - [[VERY_BAD_MULTIPLIER]])
+                                     value = ([[BAD_MULTIPLIER]] / [[VERY_BAD_MULTIPLIER]])
```
geoffthemedio: needs another `)` at the end, I think...

Grummel7: Of course you are right. That's the problem with proposing single line changes: it's hard to read it in context
```suggestion
                                     value = ([[BAD_MULTIPLIER]] / [[VERY_BAD_MULTIPLIER]]))
```

PR #3417, opened 2021-04-18
Enable testing and Godot for MacOS CI

o01eg:


Vezzra (approved):
Makes sense I guess...

PR #3421, opened 2021-04-21
Use CMake to generate NSIS installer

o01eg:
This PR is intended to replace manual weekly test build with automatic. Also it uses CMake to include Godot client in future.


o01eg:
It looks like `TARGET_DEPENDENCIES_COPY_TO_BUILD` can find dependencies of binaries but not sure how to get it from separate cmake process.

o01eg:
@Vezzra @geoffthemedio It looks like get all dependencies but requires quite fresh CMake version for Windows and maybe miss some functionality from current NSIS script. Could you check this installer?

geoffthemedio:
> Could you check this installer?

How? Is it downloadable, or do I need to build it myself locally?

The GitHub action "Windows build" produces the https://github.com/freeorion/freeorion/suites/2556127167/artifacts/55621774 artifact, which is a .zip file that contains a .exe so I guess that?

Assuming yes: The installer seems to work fine. The uninstaller also runs, but doesn't remove `/default/python/` from the installation directory, probably because it contains a bunch of .pyc files generated when running the game.

o01eg:
> Assuming yes: The installer seems to work fine. The uninstaller also runs, but doesn't remove `/default/python/` from the installation directory, probably because it contains a bunch of .pyc files generated when running the game.

I suppose it need to add some commands to CMake NSIS options to clean them too.

Vezzra (changes requested):
The installer that got created as build artifact basically seems to work, but there are a couple of issues:

- The installer file name needs adjustment (e.g. lacks the build number).
- The build is run on a checkout of detached HEAD, causing the version string to contain "HEAD", which is not desired for the weekly test builds
- The installer did not try to unsinstall an already existing FO installation, although that is probably due to that installation having been done with the "manually" created NSIS installer.
- For the weekly builds, also the "binaries only" archive needs to be created. Only creating the NSIS installer automatically isn't of much use.

These are things that have been immediately obvious. There might be other issues, but let's start with those.

o01eg:
> The installer file name needs adjustment (e.g. lacks the build number).

I think we could get build number with CMake.

> The build is run on a checkout of detached HEAD, causing the version string to contain "HEAD", which is not desired for the weekly test builds

It because CI merges `master` with this PR.

> The installer did not try to unsinstall an already existing FO installation, although that is probably due to that installation having been done with the "manually" created NSIS installer.

That is hard. I barely understand logic in NSIS script but not sure how to integrate it into CMake/CPack.

> For the weekly builds, also the "binaries only" archive needs to be created. Only creating the NSIS installer automatically isn't of much use.

It could be solved with components.

Vezzra:
As discussed in the last monthly voice chat, we can still use this to establish automated "nightly builds", and try to solve the issues step by step. We just can't use it as a replacement for the manually created weekly test builds as long as not all of these issues are solved (one way or the other).

hifron:
> > The build is run on a checkout of detached HEAD, causing the version string to contain "HEAD", which is not desired for the > > The installer did not try to unsinstall an already existing FO installation, although that is probably due to that installation having been done with the "manually" created NSIS installer.

[variable:CPACK_NSIS_ENABLE_UNINSTALL_BEFORE_INSTALL](https://cmake.org/cmake/help/latest/cpack_gen/nsis.html?highlight=nsis#variable:CPACK_NSIS_ENABLE_UNINSTALL_BEFORE_INSTALL) from [NSIS search of cmake docs](https://cmake.org/cmake/help/latest/search.html?q=NSIS) from [CMake](https://cmake.org/documentation/)

There probably could MAYBE be nighly, stable, ... branches to enable various parallel installation, but each should be updatable via correct installer...

PR #3422, opened 2021-04-22
Initial GDNative library and Godot project

o01eg:
Sets up initial Godot and GDNative projects with linters, formatters and unit tests on CI.


geoffthemedio:
not sure what I should comment on...

o01eg:
@geoffthemedio C++ code and maybe files organization.

Not sure what to do with Godot's snake_case and FreeOrion's CameCase. `_init` and `register_methods` are enforces by `godot-cpp`.For our methods could use 

```cpp
    register_method("get_version", &FreeOrionNode::GetVersion);
```

but I suppose it less readable.

geoffthemedio:
> 
> 
> @geoffthemedio C++ code and maybe files organization.

There's not much C++ code to comment on... I'd make a few style tweaks like `{}` instead of `{ }`, but that's not really important.

All the Godot stuff being in `/godot/` seems reasonable. Within that, I don't know what makes sense for organizing a Godot project's internal files.

> Not sure what to do with Godot's snake_case and FreeOrion's CameCase. `_init` and `register_methods` are enforces by `godot-cpp`.For our methods could use
> 
> ```c++
>     register_method("get_version", &FreeOrionNode::GetVersion);
> ```
> 
> but I suppose it less readable.

This is essentially exposing an API to Godot, similar to the Python API?

For the names that Godot sees, I'd use the Godot naming standard.

For cases where you need to add a function in C++ that only exists to be exposed to Godot, I don't have a strong opinion about whether to use `FreeOrionCamelCase` or `godot_underscores`. You've used the latter, which is fine. If a function has substantial use in C++ as well, then probably I'd make those `CamelCase`.

Vezzra:
@o01eg, @geoffthemedio, I've just added a "component:godot-client" label, please use for issues/PRs dealing with the Godot client.

o01eg:
@Vezzra Thanks

o01eg:
Looks like snapcraft build fails even in master.

Issue #3247, opened 2020-11-26
Named value refs should format float values better


agrrr3:
Currently, value refs defined as NamedReal are displayed like `12.000000`

Example: ![mouseover](https://user-images.githubusercontent.com/11386908/100291247-8abd2080-2f7d-11eb-807e-ab91e979c674.png)

Depending on circumstances it should be `12`, or `12.00` or `12,00` or `1200%`

That formatting decision should be made in the stringtables file where it is used.
So it would be great if we had something like a formatting syntax to apply to the values.

Or the poor mans version: make it display `12.00`

Oberlus:
I'd like to set a maximum number of decimals (say 2), and then

When there are no decimals (it can be represented by an integer), remove all decimal zeros:
12.000000 -> 12

When there are decimals, remove all decimal zeros after last non-zero decimal and truncate to maximum number of decimals if needed:
12.5 -> 12.5
12.333333 -> 12.33
12.000400 -> 12.00

With roundings?
12.666667 -> 12.67

agrrr3:
@Oberlus yes, the format suggestions look good for a poor mans implementation. if we make it a bit more fancy by allowing the client to set the number of decimals I guess we should put that into the UI rules section (or make it a functional stringtable entry?)

What would be the canonical way to do that formatting in cpp?

PR #3423, opened 2021-04-24
Display ValueRef floating points with only 3 digits in UI

Morlic-fo:
Currently, standard ```std::to_string``` format results in excessive number of digits, e.g. ```std::to_string(3.0)``` resulting in 3.000000.

This PR uses existing functionality to return only 3 digits. This number of digits was chosen for consistency with (most) other parts of the UI.

Arguably, ```EvalAsString``` doesn't specify it is only to be used in UI but appears to currently be. Should be trivial to introduce another function reproducing the current behavior if required.

### DIFF
--- a/universe/ValueRef.h
+++ b/universe/ValueRef.h
@@ -4,6 +4,7 @@
 #include "ScriptingContext.h"
 #include "../util/Enum.h"
 #include "../util/Export.h"
+#include "../util/i18n.h"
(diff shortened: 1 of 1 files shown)


geoffthemedio:
probably has some edge cases where it won't work well, but seem OK in practice for now...

agrrr3:
Relevant to issue #3247

PR #3430, opened 2021-05-03
Rename GitHub Actions jobs

Cjkjvfnby:
Improve code readability:
- split into modules
- replace dictionaries with objects

These changes do not affect the result of the generator. 
Since this code is technically and output was not changed no review required.


geoffthemedio:
no comment from me...

PR #3432, opened 2021-05-04
Let AI explore systems before attempting colonization

### LINKED ISSUE #3424: Precognitive Detection leads AI colony ships to reckless behavior
The species effect Precognitive Detection gives basic visibility of planets even in unexplored systems.
If such a planet is found and it is suitable, the AI will attempt to colonize it. 

The issue is that the good planets are often guarded by some stationary space monster. Detecting the planet via Precognitive Detection means that the guard is not scouted, often resulting in the loss of the outpost/colony ship.

A sane behaviour would be to prioritize an exploration mission to the system of interest to check for any guards before sending in some much more expensive colony ship.

For reference:
https://github.com/freeorion/freeorion/blob/8a995e1585ef6ea410b2d4c3dcf37a438729daa8/default/scripting/species/common/telepathic.macros#L1-L13

Morlic-fo:
Fixes #3424.


default/python/AI/ColonisationAI.py
```diff
@@ -1280,6 +1281,15 @@ def send_colony_ships(colony_fleet_ids, evaluated_planets, mission_type):
                 universe.getSystem(sys_id), aistate.systemStatus[sys_id]['monsterThreat']))
             already_targeted.append(planet_id)
             continue
+
+        # do not try to colonize in unexplored systems - there often are guards!
+        # Note that we can have vision of a planet in an unexplored system due to e.g. some species effects
```
Cjkjvfnby:
```suggestion
        # Note that we can have a vision of a planet in an unexplored system due to e.g. some species effects
        # for example "Precognitive Detection"
````

File: default/python/AI/ColonisationAI.py [hunk 2]
```diff
@@ -1280,6 +1281,15 @@ def send_colony_ships(colony_fleet_ids, evaluated_planets, mission_type):
...
+        has_explored_system = get_partial_visibility_turn(sys_id) > 0
+        if not has_explored_system:
```
Cjkjvfnby: I am not sure who has, maybe we can rename it to the system_is_explored?

```suggestion
        system_is_explored = get_partial_visibility_turn(sys_id) > 0
        if not system_is_explored :
```

As an alternative solution, we could put the code, TODO, and comment to the separated function:

```python
if system_may_be_guarded():
```

default/python/AI/ColonisationAI.py
```diff
@@ -1280,6 +1281,15 @@ def send_colony_ships(colony_fleet_ids, evaluated_planets, mission_type):
                 universe.getSystem(sys_id), aistate.systemStatus[sys_id]['monsterThreat']))
             already_targeted.append(planet_id)
             continue
+
+        # do not try to colonize in unexplored systems - there often are guards!
```
Cjkjvfnby:
```suggestion
        # Do not try to colonize unexplored systems - there are often guards!
```

Cjkjvfnby (changes requested):
Looks great.  

I have left some cosmetic comments.

Cjkjvfnby (approved):

PR #3436, opened 2021-05-08
Make AI aware of self-sustaining restrictions

### LINKED ISSUE #3433: AI not aware of self-sustaining metabolism change in 7ec95db
Bug Report
==========

Environment
-----------
* **FreeOrion Version**: 768af09
* **Operating System**: All
* **Fetched as** 
  * Weekly development build

Description
-----------
There has been a change on self-sustaining metabolism that makes it to apply only on Good planets (7ec95db).
AI seems to not be aware of this change, and it colonizes planets with zero target population.

Expected Result
---------------
AI is aware of the new effects of self-containment metabolism and only colonizes planets when target pop. is 1+.

Morlic-fo:
Fix #3433


Cjkjvfnby (approved):
Looks good.

PR #3438, opened 2021-05-09
Avoid negative effect from Industrialism

Grummel7:
Industrialism bonus is calculated after -5 from Environmentalism,
so the value can be negative at that point and simply multiplying
it with 1.25 can result in making it worse.

### DIFF
--- a/default/scripting/policies/INDUSTRIALISM.focs.txt
+++ b/default/scripting/policies/INDUSTRIALISM.focs.txt
@@ -19,8 +19,8 @@ Policy
             ]
             priority = [[TARGET_2ND_SCALING_PRIORITY]]
             effects =
-                SetTargetIndustry value = Value + abs(Value)
-                    * ((NamedReal name = "PLC_INDUSTRIALISM_TARGET_INDUSTRY_SCALING" value = 1.25) - 1)
+                SetTargetIndustry value = Value
+                    + abs(Value * (NamedReal name = "PLC_INDUSTRIALISM_TARGET_INDUSTRY_PERCENT" value = 25) / 100)
 
         EffectsGroup
             scope = And [
--- a/default/scripting/policies/INDUSTRIALISM.focs.txt
+++ b/default/scripting/policies/INDUSTRIALISM.focs.txt
@@ -19,8 +19,8 @@ Policy
             ]
             priority = [[TARGET_2ND_SCALING_PRIORITY]]
             effects =
-                SetTargetIndustry value = Value
-                    * (NamedReal name = "PLC_INDUSTRIALISM_TARGET_INDUSTRY_SCALING" value = 1.25)
+                SetTargetIndustry value = Value + abs(Value)
+                    * ((NamedReal name = "PLC_INDUSTRIALISM_TARGET_INDUSTRY_SCALING" value = 1.25) - 1)
 
         EffectsGroup
             scope = And [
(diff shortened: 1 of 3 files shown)


default/scripting/policies/INDUSTRIALISM.focs.txt
```diff
@@ -19,8 +19,8 @@ Policy
             ]
             priority = [[TARGET_2ND_SCALING_PRIORITY]]
             effects =
-                SetTargetIndustry value = Value
-                    * (NamedReal name = "PLC_INDUSTRIALISM_TARGET_INDUSTRY_SCALING" value = 1.25)
+                SetTargetIndustry value = Value + abs(Value)
```
geoffthemedio: just do `Value + abs(Value * (NamedReal ... ))`

default/scripting/policies/INDUSTRIALISM.focs.txt
```diff
@@ -19,8 +19,8 @@ Policy
...
+                    * ((NamedReal name = "PLC_INDUSTRIALISM_TARGET_INDUSTRY_SCALING" value = 1.25) - 1)
```
Grummel7: Well, if you prefer it like that, here you are:
```suggestion
                SetTargetIndustry value = Value + abs(Value
                    * ((NamedReal name = "PLC_INDUSTRIALISM_TARGET_INDUSTRY_SCALING" value = 1.25) - 1))
```

Since this kind of non-negative multiplication (or non-positive for bad values) is done over and over again, I wonder whether we could have a macro for it. I quickly tried the following, but when using it with "Value = [[ABS_MULTIPLY(Value, factor)]]" I get the error message "Parse error.  Expected real number expression here."
ABS_MULTIPLY
'''
    @1@ + abs(@1@) * (@2@ - 1)
'''
Note that the macro would need the braces like that to work also with factors < 0.

Grummel7: Well, in this case we'd have to re-write the description for Industrialism.
en.txt contains:
`'''Industry-focused planets get a boost to [[metertype METER_TARGET_INDUSTRY]] (x[[value PLC_INDUSTRIALISM_TARGET_INDUSTRY_SCALING]]) and....`

geoffthemedio: I think "increases Thing by 25%" is better than "multiplies Thing by 1.25", anyway...

Grummel7: Oh my, I only wanted to fix a small problem...

I agree however that "increased by 25%" sounds more natural.
However with a NamedReal it would read 25.0% and the script becomes more complicated again:
`SetTargetIndustry value = Value + abs(Value * (NamedReal name = "PLC_INDUSTRIALISM_TARGET_INDUSTRY_PERCENT" value = 25) / 100)`
Using a NamedInteger doesn't work, or is there a way to cast values in the scripts?

geoffthemedio: > with a NamedReal it would read 25.0%

seems fine...

Grummel7: I've updated the branch accordingly. My French is hardly existing anymore, by I think my adaption to fr.txt can't be too wrong.

When the policies have settled, I may update de.txt, German is my native language. At the moment it looks too much like a Sisyphean task.

PR #3441, opened 2021-05-10
Add missed entries which are referenced from the stringtables file

Cjkjvfnby:
Add strings references in ru.txt into ru.txt. This effectively reverts back to only omitting entries from translation stringtable if they are present in the default en.txt stringtable. This resolves an issue where stringtable references in a translation were not substituted correctly if the referenced string was present only in the default stringtable but not in the translated stringtable.

### DIFF
--- a/default/stringtables/ru.txt
+++ b/default/stringtables/ru.txt
@@ -1890,6 +1890,9 @@ OPTIONS_DB_SECTION_MISC
 OPTIONS_DB_SECTION_NETWORK
 Сервер и сеть
 
+OPTIONS_DB_SECTION_RAW
+All options with only raw description keys
+
 OPTIONS_DB_SECTION_SAVE
 Сохранение игры
 
@@ -4646,6 +4649,12 @@ ENC_SHIP_DESIGN_DESCRIPTION_STATS_STR
             (<i>противник с атакой %17$.1f и щитами %18$.1f</i>):  %19$.f   (%20$.2f за PP)
 '''
 
+ENC_SDD_HANGAR
+[[encyclopedia PC_FIGHTER_HANGAR]] [[METER_CAPACITY]]
+
+ENC_SDD_BAY
+[[encyclopedia PC_FIGHTER_BAY]] [[METER_CAPACITY]]
+
 ENC_UNLOCKED_BY
 '''<u>Открытые технологии:</u>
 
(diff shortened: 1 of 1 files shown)


geoffthemedio:
You added a few strings... what does that have to do with removing untranslated entries?

Cjkjvfnby:
> You added a few strings... what does that have to do with removing untranslated entries?

These entries MUST be present in `ru.txt`, because they are referenced from that file.  So only specific untranslated entries could be removed.    

Probably we should deal with it.  I'd prefer to implement a fallback to the en.txt for such entries, but since I could not do it myself, I can add a check for python stringtables check tool.

geoffthemedio:
> These entries MUST be present in `ru.txt`, because they are referenced from that file. So only specific untranslated entries could be removed.

OK, so the description should be something like "Add strings references in ru.txt into ru.txt. This effectively reverts back to only omitting entries from translation stringtable if they are present in the default en.txt stringtable. This resolves an issue where stringtable references in a translation were not substituted correctly if the referenced string was present only in the default stringtable but not in the translated stringtable." ...?
 
> Probably we should deal with it. I'd prefer to implement a fallback to the en.txt for such entries, but since I could not do it myself, I can add a check for python stringtables check tool.

I think we probably discussed how string substitutions should work, and decided to not do substitutions between tables. Likely it would be confusing if a string is present in both, as it would then behave differently if a substitution of the same string is done in the default or in the translation? 

I suspect that part of the issue is that all the substitutions are done when a stringtable is loaded, so all the default english substitutions are done with other default english strings, and all the other translations are done with strings from within those stringtables as they are loaded.

For it to be possible to rely on a default english string reference resolving to a string in a translation when loading that translation, it won't work, since the english substitutions will have already happened internally to the english stringtable. The default english and any translated stringtables are loaded and stored separately. The fallback to english, if a string isn't present in the selected stringtable, only happening when the key lookup is done, which is every time a bit of text is generated for the UI, and not when the stringtables are loaded...

It might be possible to have it work differently for translated stringtables, so that substitutions can fall back to the english, but it's a lot simpler to understand what's going on if all the substitutions are kept internal to each stringtable as it's loaded.

Cjkjvfnby:
> I think we probably discussed how string substitutions should work,

I'll forget it after a couple of years. :(

It's is not a problem to update stingtable check to catch such typo of issue.

geoffthemedio:
> 
> 
> It's is not a problem to update stingtable check to catch such typo of issue.

Not sure what you mean by that...

Cjkjvfnby:
> Not sure what you mean by that...

I mean that script https://github.com/freeorion/freeorion/blob/master/check/st-tool.py which is run via GitHub Actions on each PR.

Need to remove the second condition in this line https://github.com/freeorion/freeorion/blob/master/check/st-tool.py#L846

Cjkjvfnby:
@geoffthemedio , can we merge it?

o01eg:
@Cjkjvfnby I'm not sure about merging non-translated string to translation files.

Cjkjvfnby:
> @Cjkjvfnby I'm not sure about merging non-translated string to translation files.

Still much better that that: 
image

o01eg:
@Cjkjvfnby Maybe add commentary in file that those lines not translated.

geoffthemedio:
> @Cjkjvfnby Maybe add commentary in file that those lines not translated.

will you do that now, later, or not?

Cjkjvfnby:
> Maybe add commentary in file that those lines not translated.

I am not sure why.  Probably the person who will open this translation could notice that. I am not aware of any other use case.

PR #3446, opened 2021-05-13
Launch tests on MacOS

o01eg:


Vezzra (commented):
I don't really know how all that CI stuff works, so I'm afraid my input isn't worth much.

But as far as I can tell, looks fine...

o01eg:
Somehow it doesn't find tests.

Vezzra:
I'm sorry, but I really don't know how I can help here...

hifron:
Maybe missing export PATH and so on...

BUT on ubuntu build there is no ctest, only something with
`cmake --build . --config Release --target unittest`

o01eg:
It looks like for each test binary dependencies should be copied same way as they copied for `freeorion` target.

geoffthemedio:
once these tests work, it'd be good to test before and after d69f91ab518fabdf89a0315597eb1e6fb73f6a67 re: #4644

o01eg:
> once these tests work, it'd be good to test before and after https://github.com/freeorion/freeorion/commit/d69f91ab518fabdf89a0315597eb1e6fb73f6a67 re: https://github.com/freeorion/freeorion/issues/4644

We need to launch universe generator in test. Not sure how to make it, separate server process fails to start.

o01eg:
Looks like I need to do something with `// search bundle_path for a directory named "FreeOrion.app", exiting if not found, else constructing a path to application bundle contents` in `util/Directories.cpp` for test case to be able to use `InitDirs` in tests.
I'll try to do with it in the next PR.

Vezzra:
@o01eg, sorry for not responding earlier, but I got sick and have been out of commission for the last couple of days.

PR #3447, opened 2021-05-13
Refactored Fleet Aggression Handling

Grummel7:
Default aggression was used inconsistantly. When a fleet was
split, depending on the splitting method split-off battle ships
became aggressive or just obstrusive.
I defined alias values for default aggression of armed
and unarmed ships and used the new values where appropriate.
Also did some further cleanup in those methods.

### DIFF
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -3529,17 +3529,25 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
         auto split_action = [this, &ship_ids_set]() {
             ScopedTimer split_fleet_timer("FleetWnd::SplitFleet", true);
 
+            FleetAggression new_aggression_setting = FleetAggression::INVALID_FLEET_AGGRESSION;
+            if (m_new_fleet_drop_target)
+                new_aggression_setting = m_new_fleet_drop_target->GetFleetAggression();
+
             // assemble container of containers of ids of fleets to create.
             // one ship id per vector
             for (int ship_id : ship_ids_set) {
                 CreateNewFleetFromShips(
                     std::vector<int>{ship_id},
-                    FleetAggression::INVALID_FLEET_AGGRESSION);
+                    new_aggression_setting);
             }
         };
 
         auto split_per_design_action = [this, fleet]() {
-            CreateNewFleetsFromShipsForEachDesign(fleet->ShipIDs(), FleetAggression::INVALID_FLEET_AGGRESSION);
+            FleetAggression new_aggression_setting = FleetAggression::INVALID_FLEET_AGGRESSION;
+            if (m_new_fleet_drop_target)
+                new_aggression_setting = m_new_fleet_drop_target->GetFleetAggression();
+
+            CreateNewFleetsFromShipsForEachDesign(fleet->ShipIDs(), new_aggression_setting);
         };
 
         popup->AddMenuItem(GG::MenuItem(UserString("FW_SPLIT_FLEET"),             false, false, split_action));
--- a/UI/FleetWnd.cpp
+++ b/UI/FleetWnd.cpp
@@ -174,13 +174,13 @@ namespace {
     }
 
     FleetAggression AggressionForFleet(FleetAggression aggression_mode, const std::vector<int>& ship_ids) {
-        if (aggression_mode <= FleetAggression::FLEET_AGGRESSIVE &&
-            aggression_mode >= FleetAggression::FLEET_PASSIVE)
+        if (aggression_mode < FleetAggression::NUM_FLEET_AGGRESSIONS &&
+            aggression_mode > FleetAggression::INVALID_FLEET_AGGRESSION)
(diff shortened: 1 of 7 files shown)


UI/FleetWnd.cpp
```diff
@@ -3532,29 +3528,18 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
     {
         auto split_action = [this, &ship_ids_set]() {
             ScopedTimer split_fleet_timer("FleetWnd::SplitFleet", true);
-            // remove first ship from set, so it stays in its existing fleet
-            auto ship_id_it = ship_ids_set.begin();
-            ship_ids_set.erase(ship_id_it);
```
Grummel7, line 3537: This code means that the oldest ship in the fleet always kept the old aggression value. I don't see why it should.

UI/FleetWnd.cpp
```diff
@@ -3532,29 +3528,18 @@ void FleetWnd::FleetRightClicked(GG::ListBox::iterator it, const GG::Pt& pt, con
...
-
-            FleetAggression new_aggression_setting = FleetAggression::INVALID_FLEET_AGGRESSION;
-            if (m_new_fleet_drop_target)
-                new_aggression_setting = m_new_fleet_drop_target->GetFleetAggression();
```
Grummel7: I don't think there'll ever be a drop target when a fleet is split. Same in split_per_design below.

geoffthemedio: This isn't checking if the drop target was used as part of the splitting action, but rather if there is a drop target in the window. It can be present without being part of the action that leads to the check. The point is for players to be able to adjust how the new fleet aggression is determined based on what the aggression widget in the drop target is set to.

I guess this is "inconsistent", but it's intentionally so.

util/SerializeUniverse.cpp
```diff
@@ -367,7 +367,7 @@ void serialize(Archive& ar, Fleet& obj, unsigned int const version)
     if (Archive::is_loading::value && version < 5) {
         bool aggressive = false;
         ar  & make_nvp("m_aggressive", aggressive);
-        obj.m_aggression = aggressive ? FleetAggression::FLEET_AGGRESSIVE : FleetAggression::FLEET_DEFENSIVE;
```
geoffthemedio: this is fallback code for old versions of the serialization format, and the results should not change even if the UI behaviour or default when creating a new fleet changes

UI/FleetWnd.cpp
Grummel7: Ah! I've never used that feature, but I could see how it works now. Will revert that part.

util/SerializeUniverse.cpp
Grummel7: I guess you are right, will revert that line.

UI/FleetWnd.cpp
Grummel7: As far as I could make out, the oldest ship is always the first, but that is not the point.
There is nothing special about a fleet, they are just collections of ships with number.
Keeping the first ship in the fleet means that they player may end up with an aggressive scout just because this scout happened to be the first ship in its fleet.

UI/FleetWnd.cpp
```diff
@@ -196,8 +196,6 @@ namespace {
         //       of not at all.
         Sound::TempUISoundDisabler sound_disabler;
 
-        std::set<int> original_fleet_ids; // ids of fleets from which ships were taken
```
geoffthemedio, line 199: removing this is find since it doesn't seem to be actually used for anything

UI/FleetWnd.cpp
geoffthemedio: I think in this case, when splitting a fleet in a way that treats all ships the same, it's reasonable that this action would effectively (or really) make new fleets for each ship (or each design) and thus not treat the one ship that remains in the original fleet differently from the others. It can set the aggression of the resulting fleets independently, based on their design, if the UI is set to auto-determine fleet aggression, but if it's set to a particular aggression for new fleets, then they should all get that aggression, not all but the original fleet.

universe/Fleet.h
```diff
@@ -46,6 +46,11 @@ FO_ENUM(
     ((NUM_FLEET_AGGRESSIONS))
 )
 
+namespace FleetDefaults {
+    constexpr auto FLEET_DEFAULT_ARMED = FleetAggression::FLEET_AGGRESSIVE;
```
geoffthemedio, line 50: this is just so you can put the alias in all the places where a new fleet is created and change the default consistently by changing the single value here?

Grummel7: That's the idea. Plus it is more verbose to use an explicit default value and may avoid inconsistencies in the future.

geoffthemedio:
@Grummel7  Do you have a full / real name you'd like to be listed with in the credits (in the style used for others therein)?

Grummel7:
Well, why not. When I used to work on open source projects 20 years ago, we used nothing but the real name.
My name is Frank Muzzulini.

PR #3451, opened 2021-05-16
Add unused references check

Cjkjvfnby:
I've started as cleanup of the check script and ended with adding a check.  I don't expect any feedback for the Python part.

Most interesting part of this code is the additional check and ist implementation:
```python
exit_code += _check_key_usage(source_st, reference_st, references)
```

What changes I have done and how they affect the development. 

I've added a check for keys that are not referenced from this file and does not present in en.txt.  This will solve a couple of issues:

- When the developer removes something from en.txt he will be reminded that this key present in other translations.
- When the developer renames some references he will be reminded that there is something else to do.  This is the most important part of this check because this information is shared in a complicated way (git history) and usually is not communicated to the translators.

### DIFF
--- a/check/st-tool.py
+++ b/check/st-tool.py
@@ -16,17 +16,17 @@ import webbrowser
 from http.server import HTTPServer, BaseHTTPRequestHandler
 from collections import namedtuple, OrderedDict
 
-STRING_TABLE_KEY_PATTERN = r'^[A-Z0-9_]+$'
+STRING_TABLE_KEY_PATTERN = re.compile(r'^[A-Z0-9_]+$')
 # Provides the named capture groups 'ref_type' and 'key'
-INTERNAL_REFERENCE_PATTERN = r'\[\[(?P<ref_type>(?:(?:encyclopedia|(?:building|field|meter)type|predefinedshipdesign|ship(?:hull|part)|special|species|tech|policy|value) )?)(?P<key>{})\]\]'
-OPTIONAL_REF_TYPES = [ "value" ];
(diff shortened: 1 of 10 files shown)


geoffthemedio:
sounds reasonable...

o01eg (approved):

Cjkjvfnby:
@geoffthemedio,  I see some warning that [[value XXX]] is not found in string tables.  Am I right that this reference is always not in stringtable and this case should not be printed?

```
default\stringtables\en.txt:14815: Optional referenced key 'SHP_ANTIMATTER_TANK_EFFECT_MULT' in value of 'FU_BASIC_TANK_DESC' was n
ot found. Reference was [[value SHP_ANTIMATTER_TANK_EFFECT_MULT]].
```

geoffthemedio:
It's a named valueref label, which can be referenced in the stringtable like `[[value NAME_OF_REF]]` to be replaced by whatever is defined in FOCS. It specically says it's an "Optional referenced key" in that message, so it's not an error... just a warning...  But I think it can be included in the stringtable in order to give a more customized descdription of the reference in the UI, rather than just its value.

Cjkjvfnby:
>  so it's not an error... just a warning...

In that context of the `check` call, it looks like a problem from my point of view.  Script reports (add lines to console) about using the feature.

For me check tools should produce zero output if everything is OK.  There is a single reason why I watch this tool output: I need to act.  Looking into lines that need to be addressed among lines that don't require any actions consumes my mental capacity.

That's why I'd prefer removing these "warnings".

geoffthemedio:
fine with me...

PR #3454, opened 2021-05-17
Add missed references to de.txt

Cjkjvfnby:
Similar changes that were done for ru.txt in https://github.com/freeorion/freeorion/pull/3441

I plan to enable a check about that case as soon as all translations will be fixed.


geoffthemedio:
looks fine for what it is supposed to be, checks pass...

PR #3472, opened 2021-05-31
Test build with G++ 11

o01eg:


geoffthemedio:
I don't have any relevant comment...

.github/workflows/build-windows.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Add msbuild to PATH
         uses: microsoft/setup-msbuild@v1.0.2
         with:
-          vs-version: '[16.0,16.10)'
+          vs-version: '[16.0,16.30)'
```
geoffthemedio, line 37: > I suppose only the newest MSVS installed on CI so I couldn't choose older one.

What does this line do then?

o01eg: Searches any msbuild with that version and add it to `PATH`.

geoffthemedio: I meant what does changing from `.10` to `.30` do?

o01eg: To allow `MSVC 19.29.30037.0` be included.

geoffthemedio: If "only the newest" version is installed, why do you need to specify a version at all?

Or is it the newest version of MSVC2019 and you want to specify not MSVC2017?

o01eg: > Or is it the newest version of MSVC2019 and you want to specify not MSVC2017?

Yes.

geoffthemedio: Makes sense then.

Otherwise, as above,
> I don't have any relevant comment...

PR #3480, opened 2021-06-03
Change "maximum" for "target" in Populaton Pedia article

Oberlus:
- Remove reference to \"maximum population\" in Population Pedia, amend it slightly (only EN)
- Add named values to min colonization size and happiness (EN, ES, FR, DE)"


chomwitt:
Thanks!

default/stringtables/en.txt
```diff
@@ -6023,11 +6023,11 @@ METER_POPULATION_VALUE_LABEL
...
 
-The Target Population is the stable Population size a planet will approach, given the current factors influencing the planetary Population. Its value is connected to [[encyclopedia ENC_BUILDING_TYPE]]s built on the planet, [[encyclopedia ENC_TECH]] researched by the empire, [[encyclopedia ENC_SPECIAL]]s linked to the planet, and various other game effects. If these factors change, then the Target Population value is likely to shift.
+A Population of at least three and a minimum [[metertype METER_HAPPINESS]] of 5 is necessary for a colony to act as a source of settlers for colonizing other planets. A colony ship has a population of Colonists, which is determined by the [[encyclopedia PC_COLONY]] capacity of the ship design. When colonizing a planet, the number of Colonists aboard the ship will be the starting Population on the newly settled planet.
```
geoffthemedio: perhaps `three` should be a value reference?

Oberlus: I added named integers (in the four languages that has that label defined).
I tried but couldn't make it work with a macro substitution (I remember doing it before, but forgot what was the trick).
Ready to merge, I guess.

PR #3488, opened 2021-06-10
Combine list of identical building effects into one

Grummel7:
Identical means all but the planets, thus avoiding long
lists of liked or disliked buildings within a large empire.

See also [https://www.freeorion.org/forum/viewtopic.php?f=28&t=12016](url)

### DIFF
--- a/UI/MeterBrowseWnd.cpp
+++ b/UI/MeterBrowseWnd.cpp
@@ -329,7 +329,6 @@ void MeterBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
 
     // helpers for combining building effects into a single line
     std::vector<std::string> combined_names;
-    float combined_meter_change = 0.0f;
     // add label-value pairs for each alteration recorded for this meter
     for (auto it = maybe_info_vec->begin(); it != maybe_info_vec->end(); ++it) {
         auto info = *it;
@@ -359,18 +358,17 @@ void MeterBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
                 if (const auto& planet = Objects().get<Planet>(building->PlanetID()))
                     name = planet->Name();
             // Some effects are triggered by every building in the own empire.
-            // To avoid excessive effect list, we combine same effects.
-            // Local (same planet or system) buildings have a stronger effect
-            // and are kept apart.
-            // So this changes e.g.
+            // To avoid excessive effect lists, we combine identical effects.
+            // This changes e.g.
             // Likes <own planet> building Interstellar Lighthouse +4.00
             //    Likes <other 1> building Interstellar Lighthouse +0.71
             //    Likes <other 2> building Interstellar Lighthouse +0.71
             // to
             // Likes <own planet> building Interstellar Lighthouse +4.00
             //          Likes 2 x building Interstellar Lighthouse +1.41
-            // Note that if there is more than one of the building on other
-            // planets in the same system, you will get two combined lines.
+            // There is one line per effect strength and it shows the
+            // number of buildings involved and the summed effect of
+            // all those buildings.
             auto next = it + 1;
             if (next != maybe_info_vec->end() &&
                 next->cause_type == info.cause_type &&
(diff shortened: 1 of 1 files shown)


UI/MeterBrowseWnd.cpp
```diff
@@ -327,8 +327,12 @@ void MeterBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
     if (!maybe_info_vec)
         return;
 
+    // helpers for combining building effects into a single line
+    std::vector<std::string> combinedNames;
+    float combinedChange = 0.0;
```
geoffthemedio: prefer `combined_names` and `combined_change` format

put `0.0f` for float constants

UI/MeterBrowseWnd.cpp
```diff
@@ -354,12 +358,36 @@ void MeterBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
...
                     name = planet->Name();
-            const std::string& label_template = (info.custom_label.empty()
-                ? UserString("TT_BUILDING")
-                : UserString(info.custom_label));
-            text += boost::io::str(FlexibleFormat(label_template)
-                % name
-                % UserString(info.specific_cause));
+            // same effect from several planets?
+            auto next = it + 1;
+            if (next != maybe_info_vec->end() and
```
geoffthemedio: use the standard `&&` and `!` spellings, consistent with (as far as I know) every other use in the codebase...

UI/MeterBrowseWnd.cpp
```diff
@@ -354,12 +358,36 @@ void MeterBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
...
+                next->cause_type == EffectsCauseType::ECT_BUILDING and
+                // better not compare floats with ==
+                std::fabs(next->meter_change - info.meter_change) < 0.001 and
```
geoffthemedio: explain purpose of this check with a comment... ensuring the grouped changes are all the same size?

UI/MeterBrowseWnd.cpp
```diff
@@ -354,12 +358,36 @@ void MeterBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
...
+                next->specific_cause == info.specific_cause)
+            {
+                combinedChange += info.meter_change;
+                combinedNames.emplace_back(std::move(name));
+                continue;
+            }
+            else
```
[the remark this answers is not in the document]
Grummel7: I've fixed the code style and added a long comment to explain what is done.

UI/MeterBrowseWnd.cpp
```diff
@@ -358,28 +358,43 @@ void MeterBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
...
+            // Local (same planet or system) buildings have a stronger effect
+            // and are kept apart.
+            // So this changes e.g.
+            // Likes <own planet> building Interstellar Lighthouse +4.00
+            //    Likes <other 1> building Interstellar Lighthouse +0.71
+            //    Likes <other 2> building Interstellar Lighthouse +0.71
+            // to
+            // Likes <own planet> building Interstellar Lighthouse +4.00
+            //          Likes 2 x building Interstellar Lighthouse +1.41
+            // Note that if there is more than one of the building on other
```
geoffthemedio: meaning of this comment is unclear... is this an issue specific to how the liking / disliking effects are scripted, or a more fundamental issue to how effect accounting entries are combined?

UI/MeterBrowseWnd.cpp
```diff
@@ -358,28 +358,43 @@ void MeterBrowseWnd::UpdateEffectLabelsAndValues(GG::Y& top) {
...
+            if (next != maybe_info_vec->end() &&
+                next->cause_type == info.cause_type &&
                 // better not compare floats with ==
-                std::fabs(next->meter_change - info.meter_change) < 0.001 and
+                std::fabs(next->meter_change - info.meter_change) < 0.001 &&
                 next->specific_cause == info.specific_cause)
             {
-                combinedChange += info.meter_change;
-                combinedNames.emplace_back(std::move(name));
+                // Combined if with next, if next exists, is also a building,
```
geoffthemedio: is there an extra `if` in there?

UI/MeterBrowseWnd.cpp
Grummel7: I guess before I upload more changes, let's discuss what the outcome should be. So far the code produces one line per effect strength, so e.g. you'll get:
- Likes Orion I building Interstellar Lighthouse +4.00 // own planet -> strongest effect
- Likes 2 x building Interstellar Lighthouse +2.0 // Orion II and III, same system 1 per building
- Likes 2 x building Interstellar Lighthouse +1.41 // Other planets, 1 / sqrt(num) per planet

The code knows nothing of effect groups so you may also get:
- Likes Orion I building Interstellar Lighthouse +4.00 // own planet -> strongest effect
- Likes 2 x building Interstellar Lighthouse +2.0 // Orion II and Antares I (a single remote also gives 1.0)

If the list can be unfolded (note that the list is displayed via mouse-over, so this may be tricky), I could just as well combine all effects of one building type into one line and collect the values together with the planet names to be shown when the list is unfolded.

If unfolding is not done, I think keeping separate lines makes the list easier to understand. The code could then be simplified by only counting the number of planets, thus replacing combined_names and combined_meter_change by a single integer.

geoffthemedio: Separate lines for separate effect sizes is good / fine I think.

My only concern is whether 3 effects that each give +2 for a total of +6 should be shown as
`Likes 3x building Interstellar Lighthouse +2.0` or `Likes 3x building Interstellar Lighthouse +6.0` where the +YY number is the total, not the per-effect contribution. I lean toward the latter, but not strongly.

Grummel7: Well, so far you could always add up the numbers shown in on the right column to get the target effect. This will no longer work and when we combine several lines, but do not add their effects strengths. That's why I'd rather show the combined effect

Grummel7: Of course with a little more display hacking, we could also make it something like that:
Likes 3x building Interstellar Lighthouse 3 x +2.0 = 6.0

geoffthemedio: > That's why I'd rather show the combined effect

I think we're saying the same thing...?

Grummel7: I hope the comments are better now. I also simplified the code by removing combined_meter_change. When combining only identical effect, multiplication does the job.
I also added a small script fix. Looking at the numbers it occurred to me that sqrt(5) is not 2.04!

geoffthemedio:
Not sure if it's working correctly... At least it's hard to understand why there are separate items listed for subsets of the same building...
effect accounting
vs
effect accounting master
It doesn't seem to be combining the contributions consistently, or considering same-valued contributions as same-valued for purposes of deciding what to combine...

Grummel7:
That looks like the effects aren't identical anymore. I have a suspicion that they may have been caused by the script fix.
I've reset this change, can you try again?
If the script problem isn't as simple as it looked, its better handled in an other pull request anyway.

geoffthemedio:
looks OK now
meter accounting again

PR #3489, opened 2021-06-12
Avoid linking against libvorbisenc

sevu:
Due to ${VORBIS_LIBRARIES} it was also linking against libvorbisenc.

As for how to easily see if it's needed or not:
libvorbisenc.dll is not in the list for packaging at the end of the file, only libvorbis.dll and libvorbisfile.dll. So, if it would be needed, you would have had problems and noticed it already.


geoffthemedio:
@o01eg any comment?

o01eg:
@geoffthemedio Looks fine for me

PR #3490, opened 2021-06-12
cmake: hide PATH-Variables in the advanced section

sevu:
Maybe you know the cmake-gui or ccmake.
This image is form the cmake-gui:
cmake-gui
All the variables for setting paths are hidden behind the switch to the advanced options … except of a few. Changing that.
The removed entries in the diff will still stay hidden after this change.


geoffthemedio:
@o01eg your comment on this would also be good...

o01eg:
@geoffthemedio Looks nice

PR #3491, opened 2021-06-13
Introduce fo specific type annotations

Cjkjvfnby:
This  PR tries to address the issue that the same type is used in different contexts, namely `int` serves as different ids (System, Planet, etc.). 

Basically, a group of different objects is represented by the same type.  AI almost always expects a specific type.
Having different types of ids, will make the code more readable and allows to cache logical errors via typing.

Python typing module already has a feature for that case: it is called [NewType](https://docs.python.org/3/library/typing.html#newtype).  All we need is to declare some types and use them as annotations. 

The caveat is getting values from C++ code.  FO stubs are not aware of subtypes of ints so will not return proper types. 
This could be done via hack (manually create a map between code objects and proper return types) or via adding these types to C++ binding code (does not look like a hard task, but may take some time due to the massive number of API objects). 

@geoffthemedio could you take a look into this idea?


geoffthemedio:
IDs of planets, systems, ships, fields, etc. are all IDs of a common UniverseObject base class, so treating them as unrelated types is awkward, at least in C++. They'd need to have built-in typesafe polymorphism to be drop-in usable in all required contexts...

One could in principle distinguish between UniverseObject IDs and eg. Empire IDs and other int types. I've considered making a separate typedef for those and similar IDs, but never thought it worth the trouble.

It's also more awkward with ValueRefs, as I don't want to make a bunch of separate variants of those to accept and return different int-derived types...

Cjkjvfnby:
> IDs of planets, systems, ships, fields, etc. are all IDs of a common UniverseObject base class, so treating them as unrelated types is awkward,

Treat a ship the same way as a planet that is awkward.   `UniverseObject`  is not a thing from the AI domain, it more like technical detail that used for typing and code organization.  

> They'd need to have built-in typesafe polymorphism to be drop-in usable in all required contexts...

These cases in quite rare in Python code. Things that accept general type inside look like this (`if isinstance(target, TargetSystem):`).  


C++ is a bit more interesting ist Visitors or switch: 

```c++
    const std::string& LinkTag(UniverseObjectType obj_type) {
        static const std::string EMPTY_STRING;

        switch (obj_type) {
        case UniverseObjectType::OBJ_SHIP:
            return VarText::SHIP_ID_TAG;
        case UniverseObjectType::OBJ_FLEET:
            return VarText::FLEET_ID_TAG;
        case UniverseObjectType::OBJ_PLANET:
            return VarText::PLANET_ID_TAG;
        case UniverseObjectType::OBJ_BUILDING:
            return VarText::BUILDING_ID_TAG;
        case UniverseObjectType::OBJ_SYSTEM:
            return VarText::SYSTEM_ID_TAG;
        case UniverseObjectType::OBJ_FIELD:
        case UniverseObjectType::OBJ_FIGHTER:
        default:
            return EMPTY_STRING;
        }
    }
```

> It's also more awkward with ValueRefs,

My proposal is mostly about Python bindings modules.  But probably I will hack it on the stub generation side.

As the next step I'll try to improve stub generators a bit.  Probably some guessing based on function names could do a trick.

geoffthemedio:
> My proposal is mostly about Python bindings modules.

A bunch of `TypedID` functions could probably be added to `System`, `Planet`, `Ship`, etc. in the Python API, which return a bunch of typedefed ints that are specific for those object types, if that would be useful. It would require converting those back to int to use any of the API calls that expect object IDs as int parameters, though...

Cjkjvfnby:
> A bunch of TypedID functions could probably be added to System, Planet, Ship, etc. in the Python API, which return a bunch of typedefed ints that are specific for those object types, if that would be useful. It would require converting those back to int to use any of the API calls that expect object IDs as int parameters, though...

Doesn't look like a hard task but will consume a lot of time, maybe start adding this for a new code.  

At this point, I just updated the skeleton generator so it will return the required type.  Not all entries are updated but I think the state is usable enough to be merged.

geoffthemedio:
> maybe start adding this for a new code

not sure what you mean by that...

Cjkjvfnby:
> not sure what you mean by that...

Sorry my bad, Python is a super flexible language.  If you inherit these DomainSpecific types from int then python will not notice any changes.  Probably small inconvenience if this value should be passed back to the C++.  

If we decide to move C++ wrapper in this direction we could start by introducing it to features we add to the code.

PR #3496, opened 2021-06-16
Fix compilation

o01eg:
Fixes missing `FLT_MAX` declaration error

### DIFF
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -1,6 +1,7 @@
 #include "Conditions.h"
 
 #include <array>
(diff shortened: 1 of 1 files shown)


geoffthemedio:
> missing `FLT_MAX` declaration error

Does this appear in any of the CI builds? Latest master has all checkmarks...

o01eg:
It appeared on my Gentoo after boost and gcc update.

sevu:
Same here. Gentoo and Arch are always ahead with software versions. If there's a change, they are hit first.

PR #3497, opened 2021-06-16
Return false from auth.py result to mark unallowed player names

o01eg:
Could be used to allow only registered players

### DIFF
--- a/default/python/auth/auth.py
+++ b/default/python/auth/auth.py
@@ -54,7 +54,8 @@ class AuthProvider:
         return roles
 
     def is_require_auth_or_return_roles(self, player_name):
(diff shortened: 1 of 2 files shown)


Oberlus:
@o01eg Would this allow players introducing their name with a typo to be notified? Something like "This player name is not registered in current game, review the name you introduced or contact the server admin."

o01eg:
Not yet, I need to refactor queries for it. Currently it just sends error instead of showing password window.

Oberlus:
> Not yet, I need to refactor queries for it. Currently it just sends error instead of showing password window.

That's clearly an improvement.

PR #3504, opened 2021-06-26
document stability requirements

Grummel7:
Added lots of MIN_STABILITY values and used them in en.txt
Also added some missing documentation.
Finally did two minor fixes/improvement to the scripting:
I modified the effect of Industry Center to make sure
you never lose production after researching a refinement.
The Influence effect of Diversity did not match the description.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -9075,6 +9075,7 @@ Eccentric Orbit
 
 ECCENTRIC_ORBIT_SPECIAL_DESC
 '''Increases [[metertype METER_TARGET_RESEARCH]] by [[value ECCENTRIC_ORBIT_TARGET_RESEARCH_FLAT]], but [[metertype METER_SUPPLY]] is decreased ([[value ECCENTRIC_ORBIT_MAX_SUPPLY_FLAT]]).
+The research bonus requires a stability of [[value ECCENTRIC_ORBIT_MIN_STABILITY]].
 
 This planet's orbit is very eccentric. It has a large distance between its closest and farthest approaches to its star. Total insolation varies significantly throughout the year. The varied conditions inhibit the supply the planet can provide, but are a beneficial platform for research.'''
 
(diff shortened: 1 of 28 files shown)


Grummel7:
Btw. there are two things I skipped because I feel they should not remain as they are:
- Ancient Ruins: When depleted, they require stability 15 for the research bonus, originally only 10.
- Divine Authority gives: Stability +4, Research -5 and Influence +1, Requiring Stability 10. Something that increases stability shouldn't have a stability requirement. Otherwise you can end up with two identical planets, one having stability 8 and the other having 12, both staying at there level. Researching more at lower stability is also questionable. Perhaps the influence bonus alone can have a stability requirement.

default/stringtables/en.txt
```diff
@@ -11594,8 +11600,9 @@ Diversity
 
 PLC_DIVERSITY_DESC
 '''Gives bonuses or penalties to research, stability, and capital influence, dependent on the number of species in the empire.
+The research bonus requires a stability of [[value PLC_DIVERSITY_MIN_STABILITY]].
 
-Gives bigger bonuses or penalities the more above or below 4 species that are present.'''
+All bonuses or penalties are [value PLC_DIVERSITY_SCALING]] per species above or below 4 species that are present.'''
```
geoffthemedio, line 11605: please double-check the changes...

diversity

Grummel7: I opened all pages, but missed two mistakes anyway. Updated the branched.

Any idea regarding the two points I made in my comment?

geoffthemedio:
> Ancient Ruins: When depleted, they require stability 15 for the research bonus, originally only 10.

I don't have a strong opinion about this. It makes some sense that it would be more difficult to get a research bonus from "depleted" ruins than pristine ruins, though I don't know the details of the relevant mechanics for depletion of ruins.

> Something that increases stability shouldn't have a stability requirement. [...] Researching more at lower stability is also questionable.

A flat +4 stability and -5 research, independent of target stability, seems reasonable.

> Perhaps the influence bonus alone can have a stability requirement.

sure, also sounds reasonable.

PR #3505, opened 2021-06-28
Didn't kick out current player connection

o01eg:
Fixes case when client send multiple Lobby Update messages to join as empire and couldn't issue orders: https://freeorion.org/forum/viewtopic.php?p=107021#p107021


geoffthemedio:
I don't know enough about what's going on with this to have much to say... seems plausibly fine?

PR #3509, opened 2021-06-30
Add Godot methods to connect to remote server and get universe objects

o01eg:


geoffthemedio:
not sure what feedback I should give...

o01eg:
@geoffthemedio C++ code. Is it's fine I'll merge it.

hifron:
Question may be if Godot is proper way to incorporate into this project, especially when C++ interface is not in good shape.

There is also possible to use [OGRE](https://www.ogre3d.org/download/sdk/sdk-ogre) with Android or with SDL.  However there is [OGRE Resource Manager](https://ogrecave.github.io/ogre/api/latest/_resource-_management.html). Supported is also Blender 3D for creators of 3D graphics... But to have MacOS support is currently problem, it is under-developed.

There is also PRO engine [Unity](https://unity.com), but not on open source project.

For a lot of programmer is problem to program in C++. Godot scripts or OGRE python may help, but I don´t know what is proper way or preferred. Godot for this project seems particular premature.

Issue #3500, opened 2021-06-22
Link error on macOS


Vezzra:
Bug Report
==========

Environment
-----------

* **FreeOrion Version**: a52063bf9224295d34a3da8377238227ee0524f6
* **Operating System**: macOS 10.14.6
* **Fetched as** 
  * Compiled from source

Description
-----------

```
ld: can't map file, errno=22 file '/Users/user/SoftwareProjekte/FO/main/build/dep/Frameworks/vorbis.framework' for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

geoffthemedio:
related to https://github.com/freeorion/freeorion/pull/3489 ?

try reverting b9c464b

Vezzra:
Yep, reverting that commit (which, btw, is d79e0d1e9fdce57ca8a83e251c71068e01d13f2e in master) fixes the link error on macOS.

Should the revert be commited?

geoffthemedio:
@sevu can you fix it in ~24 hours?

sevu:
My best hint is 0732058. There's some extra treatment for frameworks … but that applies not to targets, only variables.

Does it work if you exclude the line `Vorbis::VorbisFile` from d79e0d1.

Vezzra:
> Does it work if you exclude the line `Vorbis::VorbisFile` from d79e0d1

No, unfortunately not. Same link error.

As a temporary fix I'm going to revert d79e0d1e9fdce57ca8a83e251c71068e01d13f2e. @sevu, can you try to provide a fixed version of #3489 that works on macOS?

Vezzra:
Fixed by 24a08d6c597ce8afc90ccec485a937982b6c3683 (revert of d79e0d1e9fdce57ca8a83e251c71068e01d13f2e).

sevu:
@Vezzra I made a new PR, could you confirm if it works, after the CI passed.

I'm pretty sure it failed because I had removed the include directory.
On the CI the vorbis installation is probably in a path already included by another dependency, and thus it passed last time.

PR #3512, opened 2021-07-05
cmake-handling of libvorbis

sevu:
See #3500

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -653,7 +653,8 @@ if(BUILD_CLIENT_GG)
         $<IF:$<BOOL:${OPENGL_opengl_LIBRARY}>,${OPENGL_opengl_LIBRARY},${OPENGL_gl_LIBRARY}>
         ${OPENAL_LIBRARY}
         ${OGG_LIBRARIES}
-        ${VORBIS_LIBRARIES}
+        ${VORBIS_LIBRARY}
+        ${VORBISFILE_LIBRARY}
         ${ICONV_LIBRARY}
     )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,7 +276,7 @@ if(BUILD_CLIENT_GG)
     find_package(OpenGL REQUIRED)
     find_package(OpenAL REQUIRED)
     find_package(Ogg REQUIRED)
-    find_package(Vorbis REQUIRED)
+    find_package(Vorbis COMPONENTS Vorbis VorbisFile REQUIRED)
 endif()
 
 if(APPLE)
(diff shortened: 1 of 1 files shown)


geoffthemedio:
@Vezzra ping

Vezzra:
Unfortunately trying to build this PR on my Mac still fails, same link error:
```
ld: can't map file, errno=22 file '/Users/user/SoftwareProjekte/FO/main/build/dep/Frameworks/vorbis.framework' for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
The CI build on macOS 10.15 however succeeds, so this is either a problem with older macOS versions (I run on 10.14), or there is something amiss with my specific setup. Although I really have no idea what the problem could be...

sevu:
Thank you.

For the intended change of this PR, the first commit would be enough.
The second is fancy stuff which I found by reading the comments. As it checks conditions before building, I think it's not the reason here.
And for the third, can you build without that one?

For getting more info about the fist one, can you add this at the end of the file:
```
message(STATUS "Value of VORBIS_INCLUDE_DIRS is: ${VORBIS_INCLUDE_DIRS}")
message(STATUS "Value of VORBIS_LIBRARY is: ${VORBIS_LIBRARY}")
message(STATUS "Value of VORBISENC_LIBRARY is: ${VORBISENC_LIBRARY}")
message(STATUS "Value of VORBISFILE_LIBRARY is: ${VORBISFILE_LIBRARY}")
```
This would be printed somewhere at the beginning before building, easy to miss.

To get some more info from the error message, setting `-v` for clang could be done by setting the environment variable `CXXFLAGS=-v`.
Another idea would be to set the verbose mode of cmake. There's different ways to do it, I'm not sure which one applies to mac, setting the environment variable `VERBOSE=1` would be a way with cmake ≥3.14.

However, my hopes are on keeping the first commit and dropping the last one. No need to dig too deep if that works.
And it's good that it fails for you, that way more setups can be covered.

Vezzra:
@sevu, if I only try to build and run the first commit, a1d4154c8f958e771a4d20f74e7a46772bc0da8e, everything works (as far as I can see).

I'm short on time, I'll try the other commits too within the next couple of days.

Vezzra:
@sevu, fbf25611c68d4951179c1f9c73bf4abed405e9a2 builds and runs ok too. So it seems only the last commit (d426b99349518c342702ab086aea7f10bccf33ac) is causing the issues.

Vezzra:
@sevu, just to make sure I tried to build d426b99349518c342702ab086aea7f10bccf33ac once more right now, and as expected, it failed. So it really looks like the issue is with the changes of that commit.

geoffthemedio:
So what is the status of this?

PR #3513, opened 2021-07-06
Added a script to check language files

Grummel7:
As it is, the script works only on Linux or other Unix-like systems, maybe Cygwin as well.
I think only be the find command in line 19 that needs to be replaced for other systems, but that must be done by someone else. I have no experience in scripting for Windows or Mac.

### DIFF
--- a/check/st-tool.py
+++ b/check/st-tool.py
@@ -172,7 +172,7 @@ class StringTable(object):
                 entry.value_times = value_times[entry.key]
 
     @staticmethod
-    def from_file(fhandle, with_blame=True):
+    def from_file(fhandle, with_blame=True, check_brackets=False):
         fpath = fhandle.name
 
         is_quoted = False
@@ -193,8 +193,23 @@ class StringTable(object):
         untranslated_key = None
         untranslated_keyline = None
         untranslated_lines = []
+        if check_brackets:
+            # compile them once to speed up the loop
+            reBrackets = re.compile(r'\[.*\]')
+            reOneBracket = re.compile(r'\[[^\[\]]*\]')
+            reBadBracket = re.compile(r'[\[\]]')
 
         for fline, line in enumerate(fhandle, 1):
+            if check_brackets:
+                # hack for DESC_AND_BEFORE/AFTER_OPERANDS: only check, if there is at least one pair
+                if (reBrackets.search(line)):
+                    aux = line;
+                    while True:
+                        (aux, n) = reOneBracket.subn('', aux)
+                        if n == 0: break;
+                    if reBadBracket.search(aux):
+                        print("{}:{}: Unbalanced brackets: {}".format(fpath, fline, line))
+
             if not is_quoted and not line.strip():
                 # discard empty lines when not in quoted value
                 if section:
@@ -839,7 +854,7 @@ def check_action(args):
     if args.reference:
         reference_st = StringTable.from_file(args.reference, with_blame=False)
     for source in args.sources:
-        source_st = StringTable.from_file(source, with_blame=False)
+        source_st = StringTable.from_file(source, with_blame=False, check_brackets=True)
 
         references = set()
         for key in source_st:


Cjkjvfnby:
Hi!

We already have the tool to check language files and it is already a part of the merge pipeline. 
https://github.com/freeorion/freeorion/blob/master/check/st-tool.py

I'm not familiar with Perl, does your script find something new in string tables (language files)?

geoffthemedio:
I believe it's specifically for named value references and macro insertions?

I plan to comment regarding exporting those to stdout when requested. It's not trivial to do without also launching the GUI, I think...

Grummel7:
> Hi!
> 
> We already have the tool to check language files and it is already a part of the merge pipeline.
> https://github.com/freeorion/freeorion/blob/master/check/st-tool.py

Well, first of all it would have been nice to know that there is a checker already. I've some changes to en.txt without knowing. My script finds typos like:
- unbalanced square brackets as in [value MIN_COLONY_SIZE]]
- typos in the names of referenced value of texts, i.e [[value MIN_CLONY_SIZE]]

I started working on finding open references as well (it showed me e.g. that RESEARCH_ARTICLE_SHORT_DESC is missing in the current en.txt), but that is getting some false positives still.

I'll have a better look at st-tool.py, but from its results there seems to be one crucial thing that it does not: read the FOCS files. It only checks the language file itself, and so it cannot find inconsistencies between scripting and the language files.

Cjkjvfnby:
> Well, first of all it would have been nice to know that there is a checker already. 

Do you have any suggestion where I could put the reference to it?

> unbalanced square brackets as in [value MIN_COLONY_SIZE]]
> typos in the names of referenced value of texts, i.e [[value MIN_CLONY_SIZE]]

Probably this should be found. If not it is an issue in the tool.

 > started working on finding open references as well 

Probably we could start from check that this word is present anywhere in FOCS files (txt and Python).  I am not sure about accuracy, but it will be a useful check.  

@o01eg, you work a lot with value references lately, will it work?


@Grummel7 Are you familiar with Python? Do you feel comfortable updating st-tool?

o01eg:
> @o01eg, you work a lot with value references lately, will it work?

I didn't work with stringtables

Cjkjvfnby:
Having stringtables checked before commit could be useful.
https://github.com/freeorion/freeorion/pull/3519

Cjkjvfnby:
> I didn't work with stringtables

In string tables we could reference values like that:  `[[value FLD_NANITE_SWARM_STRUCTURE_FLAT]]`

The starting idea is to do something like `grep FLD_NANITE_SWARM_STRUCTURE_FLAT default/scripts/????` 
Add fail if the value is not found.

Cjkjvfnby:
> unbalanced square brackets as in [value MIN_COLONY_SIZE]]

This check was missed: https://github.com/freeorion/freeorion/pull/3520

geoffthemedio:
From a forum private message from Grummel7:
> Checking for missing braces isn't too hard. Checking that all displayed values are actually defined in scripting is trickier. I thought I could find the values in the script by simple regexp parsing, but then I found cases like SHP_DEUTERIUM_TANK_EFFECT_MULT with is defined by "token pasting" in PART_UPGRADE_MAXFUEL_EFFECTS.
> [...]
> [Could] we add e.g. a command line option to the client *or and extra executable) to print all defined NamedReal, NamedInteger, possible Meter Type, etc., then exit.

I looked into this for a bit. It's not trivial to add like the `-h` help flags, as it needs the parsers to have run first before it can output all of the results. In the client, that happens in the `IApp::StartBackgroundParsing` and also requires Python to be initialized and needs some thread-related stuff to be passed in. Maybe instead adding it to the server would make more sense and be easier, but I didn't look into that yet.

Grummel7:
> Do you have any suggestion where I could put the reference to it?

Well, instead of putting it into a parallel directory I would have put the script into the stringtables directory.
But I guess putting a readme.txt, explaining how to work the language files, into the stringtable directory would do just as well.

> @Grummel7 Are you familiar with Python? Do you feel comfortable updating st-tool?

I am more familiar with perl, but I've work with python, too. The actual checks shouldn't be to complicated, the big question is how to get the data from the FOCS files. With perl and find it was almost trivial, but that is not a portable solution. See also geoff's post, if he can add an output function to the server, then getting the data into python shouldn't be too complacated.

Cjkjvfnby:
>  But I guess putting a readme.txt, explaining how to work the language files, into the stringtable directory would do just as well.

I put a note to myself to do it.

>  With perl and find it was almost trivial, but that is not a portable solution.

Easy in Perl -> easy in Python. 

> See also geoff's post, if he can add an output function to the server,

That's complicated. To get all values that could be referenced you need to evaluate FOCS. And to evaluate FOCS you need to compile parsers. This adds an extra layer of complexity to CI and local development.

Just a suggestion:
Probably we could move parsers to a separate module. And write an executable program that uses that library to produce some output (list of referenced values). 
Since parsers are not changed often, we could just compile that program and use it as a standalone binary.

Using a different binary than the server will make CI and local process a bit easier.  But I am not aware of how hard it will be to extract parsers from other code.

geoffthemedio:
> Probably we could move parsers to a separate module. And write an executable program that uses that library to produce some output (list of referenced values).
> Since parsers are not changed often, we could just compile that program and use it as a standalone binary.
> 
> Using a different binary than the server will make CI and local process a bit easier.

Why is a separate program much better than using the server with a special flag that causes it to parse, output the results to stdout, and then exit?

Grummel7:
I just saw that Cjkjvfnby has done a lot of coding on st-tool.
I suggest to close this pull request, but keep working on a FOCS parser.
As soon as we have it, either him or me implements some cross checks between scripting and language files.

Cjkjvfnby:
> Why is a separate program much better than using the server with a special flag that causes it to parse, output the results to stdout, and then exit?


1) Separation of concerns: The server is responsible for the game. The focs_parser is responsible for FOCS validation. We already have a lot of flags for the server. Let's try to keep things small.
2) You can compile focs_parser once and reuse it later for CI and local checks. (We could use it in the same way as st-tool). As I know parser is not updated often.

In general, it is about architecture and modularity. At this stage of the project, they make things less complicated.

I don't know if it is an easy task. Looking at AI code I guess that extracting a module might be a problem.

Cjkjvfnby:
> I just saw that Cjkjvfnby has done a lot of coding on st-tool.

Not so much, and thank you for bringing these problems to my attention.

> I suggest to close this pull request, but keep working on a FOCS parser.
> As soon as we have it, either him or me implements some cross checks between scripting and language files.

Sounds like a plan.

geoffthemedio:
> You can compile focs_parser once and reuse it later for CI and local checks. (We could use it in the same way as st-tool). As I know parser is not updated often.

Doesn't the same apply if it was a feature of the server binary?

> In general, it is about architecture and modularity. At this stage of the project, they make things less complicated.

To me, adding a flag to the server binary is simpler than adding another binary.

Issue #950, opened 2016-09-04
Evacuation will evacuate a new colony


l29ah:
- start http://dump.bitcheese.net/files/furyfub/bug-colony.mps as L29Ah
- scrap evacuation at (already evacuated) Navi 5 I
- colonize Navi 5 I with the colony ship hanging there
- end turn
- it will vanish the next turn and evacuation will still be there, so it will be wasted

Vezzra:
Well, that's not actually a bug, but, lets say, expected behaviour. When you issue the scrap command, the building gets _marked_ for scrapping, the actual scrapping occurs during turn execution. The colonizing command works the same way - you issue the command, actual command execution occurs during turn execution.

If you issue both commands on the same turn, the order in which those commands are executed matters. Obviously colonizing and the effects of the Evacuation building happen before scrapping, so your pop gets evacuated before the building is scrapped.

You need to wait a turn before you colonize that planet.

I don't know if that issue can be fixed at all. Of course we could try to change the order in which things are done during turn execution, but that will probably lead to other issues, where yet another order would be preferable. I doubt that you can come up with an ordering that will be optimal for all possible different settings/situations.

But I don't want to dismiss this as wontfix right away, maybe others can offer more insight what can possibly done here.

l29ah:
Is there an article where i can read about the order of order executions?

l29ah:
Also why isn't the evacuation get auto-scrapped the turn it evacuates all the population?

geoffthemedio:
See: https://github.com/freeorion/freeorion/blob/master/server/ServerApp.cpp#L2820

l29ah:
I wonder if i should make a pedia article on this.

MatGB:
It would make sense to expand the Evacuation and Concentration Camp articles a bit—note that it's not a problem for the outpost building method of colonising, and we're trying to encourage that method over colonise by ship.

The turn order is displayed in the message window, invasions and colonisation happen first, before pretty much everything, and effects are processed near the end. Building scrapping, IIRC, happens before effects—it might be possible now we have a priority system to have the buildings destroy themselves on the turn they complete but I've not looked into trying.

l29ah:
On Mon, Sep 05, 2016 at 09:10:25PM -0700, Mat Bowles wrote:

> It would make sense to expand the Evacuation and Concentration Camp articles a bit—note that it's not a problem for the outpost building method of colonising, and we're trying to encourage that method over colonise by ship.

[...]

PR #3514, opened 2021-07-07
Ignore freshly established colonies for evacuation building effects

agrrr3:
Ignoring freshly established colonies on BLD_EVACUATION (i.e. do destroy the building and do not decrease population if the building is on such a colony).

Solves #950

### DIFF
--- a/default/scripting/buildings/EVACUATION.focs.txt
+++ b/default/scripting/buildings/EVACUATION.focs.txt
@@ -63,7 +63,7 @@ BuildingType
                 OwnedBy empire = Source.Owner
                 // evacuation building needs to exist one turn for technical reasons.
                 // it should not destroy a new colony
-                (LocalCandidate.TurnsSinceColonization > 1)
+                (LocalCandidate.TurnsSinceColonization > 0)
             ]
             priority = [[CONCENTRATION_CAMP_PRIORITY]]
             effects = [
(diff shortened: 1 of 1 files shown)


geoffthemedio:
You've tested this, including establishing a colony (in supply to main empire) and immediately producing an evacuation on it, and all works as expected?

agrrr3:
> You've tested this, including establishing a colony (in supply to main empire) and immediately producing an evacuation on it, and all works as expected?

nice catch, there was an issue:
* colonizing and 1-turn building evacuation building does increase **but not decrease pop**

after shifting the turn numbers, this is fixed and following is tested:
* evacuation - usually population goes down by two, pop on another planet goes up by two
* completely evacuated planet still has building, turning destroys it. colonizing on the turn the building still exists does not make it decrease/increase population.
* colonizing and 1-turn building evacuation building does decrease/increase pop and thus immediately evacuates the colony

geoffthemedio:
> * colonizing and 1-turn building evacuation building does decrease/increase pop and thus immediately evacuates the colony

what about with a cryo colony pod (3 pop to start, I think) ?

agrrr3:
> > * colonizing and 1-turn building evacuation building does decrease/increase pop and thus immediately evacuates the colony
> 
> what about with a cryo colony pod (3 pop to start, I think) ?

as expected -2 pop decrease (and +2 pop increase) and building still there. scrapping or turning works as expected

agrrr3:
is it possible to have automated tests for such effects?

geoffthemedio:
> is it possible to have automated tests for such effects?

Probably, but constructing the test scenario and conditions for pass/fail for something like this is complicated...

PR #3526, opened 2021-07-15
Don't save game with incorrect turn number

o01eg:
Could be backported to 0.4.10.2

### DIFF
--- a/server/ServerFSM.cpp
+++ b/server/ServerFSM.cpp
@@ -402,7 +402,8 @@ void ServerFSM::HandleNonLobbyDisconnection(const Disconnection& d) {
         ErrorLogger(FSM) << "Unable to recover server terminating.";
         if (m_server.IsHostless()) {
             if (GetOptionsDB().Get<bool>("save.auto.hostless.enabled") &&
-                GetOptionsDB().Get<bool>("save.auto.exit.enabled"))
+                GetOptionsDB().Get<bool>("save.auto.exit.enabled") &&
+                m_server.CurrentTurn() > 0)
             {
                 // save game on exit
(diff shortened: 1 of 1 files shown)


geoffthemedio:
could you elaborate on or make more specific the commit message? as written, it's not clear what is meant and it can be interpreted different ways...  Some info about when this could occur, or that you don't know when but have observed that for some reason something happens that leads to an "incorrect" turn number being saved, and what actually "incorrect" means, would be better...

o01eg:
@geoffthemedio Sometimes the server saves game with -65535 as turn number. It often happens if the server lacks free space, and saves incomplete file, and then loads it.

Vezzra:
> Could be backported to 0.4.10.2

Is the issue this PR addresses something sufficiently serious? There are probably a lot of things that could be backported, but for 0.4.10.2 we're trying to keep things to a minimum, so...

geoffthemedio:
> @geoffthemedio Sometimes the server saves game with -65535 as turn number. It often happens if the server lacks free space, and saves incomplete file, and then loads it.

OK, can you put that in the commit message and probably also in a comment in the code where the check is done, saying that in the case that such an invalid turn number is found, saving is skipped since something has already gone wrong, so it doesn't make sense to create a broken save...?

o01eg:
@Vezzra No, it mostly fixes consequences of other issues.

Vezzra:
@o01eg, well, I'll leave the decision whether to backport this to the 0.4.10 release branch or not to @geoffthemedio and you. I don't think it's necessary, but if you want to include it, I won't object.

o01eg:
@Vezzra Yes, it doesn't fix anything very important.

PR #3529, opened 2021-07-19
Make policy adoption costs based on #planets or #population use square root

Oberlus:
Due to scarcity of influence points for bigger empires, using absolute values of planets or populations (even with a constant factor reducing it) makes it too expensive to adopt most of the policies once a certain empire size has been reached. This PR deals with that by using the square root of such magnitudes. Costs for starting empires remain very close to what they were.

Also add macros for #ships, #planets and #population in policies.macros and change base costs in a few policies:
- Augmentation: #population instead of 100 -> make it easier for smaller empires, harder for bigger ones
- Bureaucracy: #population instead of #planets -> better for fluff
- Checkpoints: #planets instead of #population -> better for fluff
- Design Simplicity: #ships instead of #population -> better for fluff, also rewards getting it earlier in army-growth process instead of earlier in empire-growth process
- Indoctrination: #population instead of 10 -> better for fluff
- Martial Law: #population instead of 10 -> better for fluff

### DIFF
--- a/default/scripting/policies/ALLIED_REPAIR.focs.txt
+++ b/default/scripting/policies/ALLIED_REPAIR.focs.txt
@@ -3,7 +3,7 @@ Policy
     description = "PLC_ALLIED_REPAIR_DESC"
     short_description = "PLC_ALLIED_REPAIR_SHORT_DESC"
     category = "MILITARY_CATEGORY"
-    adoptioncost = floor(5 + 0.25 * Statistic Sum value = LocalCandidate.Population condition = And [ Planet OwnedBy empire = Source.Owner ])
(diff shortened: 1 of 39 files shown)


Oberlus:
Is there anything I can do to improve chances of this being merged?

There are no errors. I tested it before making the PR. It goes in line with what Vezzra proposed about better abundance than scarcity of IPs. It makes sense with respect to what has been playtested already (several players stated it's too hard to get policies late game). It is mathematically sound: IP output decreases with empire size, so it makes sense that IP requirements do not increase superlinearly with empire size.

Vezzra:
Well, sounds good to me - @geoffthemedio, anything preventing this from being merged?

PR #3530, opened 2021-07-23
Use unexpanded snapcraft.yaml

agrrr3:
Weekly build failed to publish. I retried building using the unexpanded version in the local directory. 

Before i compared the checked-in expanded version to sanely match the unexpanded version after expansion.
Build and publishing went fine (i.e. the snapcraft.io tools upgraded their snap/snapcraft versions) and the game runs.


geoffthemedio:
want / need feedback? I have none to give...

agrrr3:
Thanks, no need for input

PR #3531, opened 2021-07-25
Add black formatter

Cjkjvfnby:
Add black formatter and reformat some files.

I use [black](https://github.com/psf/black) for my personal projects, on my previous job, and when working with students. I am really happy with it. 

Not the best style, but zero effort for support and config it. 

This is the first part, I just added all required configs and reformated all python files outside the default diectory.


geoffthemedio:
looks like it's all Python changes. what review do you want?

Cjkjvfnby:
> looks like it's all Python changes. what review do you want?

Adding formatter is a project-level change.  I want to be sure that you are aware of it.

test-scripting/techs/Categories.inf.py
```diff
@@ -1,44 +1,17 @@
...
-Category(
-    name="SHIP_WEAPONS_CATEGORY",
-    graphic="ship_weapons.png",
-    colour=(255, 139, 85, 255))
-
-Category(
-    name="SPY_CATEGORY",
-    graphic="spy.png",
-    colour=(168, 0, 255, 255))
+Category(name="LEARNING_CATEGORY", graphic="learning.png", colour=(54, 202, 229))
```
geoffthemedio: I think this was more readable before the reformat

test-scripting/game_rules.focs.py
```diff
@@ -5,7 +5,8 @@
     type=float,
     default=1.0,
     min=0.1,
-    max=10.0)
+    max=10.0,
```
geoffthemedio, line 8: I'm surprised this (`,` with nothing after before the `)`) isn't a format error in Python...

msvc2017/update_stdafx.py
```diff
@@ -141,18 +147,17 @@ def update_stdafx_h(stdafx_h_filename, header_only_vcxproj_files, vcxproj_files)
 
         lines.append(include_directive)
 
-    with open(stdafx_h_filename, 'w') as f:
+    with open(stdafx_h_filename, "w") as f:
```
geoffthemedio, line 150: why is `"` preferred over `'` ?

test-scripting/techs/Categories.inf.py
Cjkjvfnby: This line fits in one line (120) so it is inlined. 

This is the disadvantage of such tools.  They are better only for 90% of cases.

test-scripting/game_rules.focs.py
Cjkjvfnby: Trailing comma is ok for function calls and list literals.  For tuples with a single element, it is mandatory. 

```
(1) == 1  # True
(1, ) == (1, )  # True
```

I find this feature very useful, since it reduce diff when adding new argument/element.

msvc2017/update_stdafx.py
Cjkjvfnby: Just to use the same style evrythere. 

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings

test-scripting/techs/Categories.inf.py
Cjkjvfnby: I went to docs to answer other questions and found that this could be fixed.

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma

I will fix it tomorrow.

test-scripting/game_rules.focs.py
geoffthemedio: > it reduce diff when adding new argument/element.

You mean just when the params are on separate lines?

Cjkjvfnby: On separate lines with coma:

```diff
- a
+ a,
+ b
```

```diff
+ b,
```

Cjkjvfnby:
@geoffthemedio, I think I have addressed all your comments.   I plan to merge it at a weekend.

PR #3545, opened 2021-08-14
Reformat code with black

Cjkjvfnby:
If some lest are squashed to a single line, it's ok.  Will fix it later.


default/python/AI/AIDependencies.py
```diff
@@ -102,30 +103,36 @@ class Tags:
 POP_SIZE_MOD_MAP_MODIFIED_BY_SPECIES = {
     # "Effect": [uninhab, hostile, poor, adequate, good]
     "environment_bonus": [0, -4, -2, 0, 3],
-    "GRO_SYMBIOTIC_BIO": [0,  0,  1, 1, 1],
```
geoffthemedio, line 105: extra whitespace was clearly intentional for layout / readability... you prefer to remove?

default/python/AI/AIDependencies.py
```diff
@@ -652,78 +659,76 @@ class Tags:
...
-    "MISIORLA"
-]
+EXTINCT_SPECIES = ["BANFORO", "KILANDOW", "MISIORLA"]
 # </editor-fold>
 
 # <editor-fold desc="Piloting traits">
 # TODO (Morlic): Consider using only 1 dict with tuple for (Capacity, SecondaryStat) effects
 PILOT_DAMAGE_UNSCALED_MODIFIER_DICT = {
     # TRAIT:    {weapon_name: effect, weapon_name2: effect2,...}
-    "NO":       {},
```
geoffthemedio, line 673: same

default/python/AI/AIDependencies.py
Cjkjvfnby: I am not sure that we lost a lot of readability.  I don't want to make an exclusion for a couple of variables.

**The Zen of Python**
> Special cases aren't special enough to break the rules.

geoffthemedio: as you prefer for Python code...

default/python/AI/ResourcesAI.py
```diff
@@ -317,10 +360,11 @@ def assess_protection_focus(pinfo):
...
-                          min(2, len(sys_status.get('enemies_nearly_supplied', []))))
+    threat_from_supply = (
+        0.25 * aistate.empire_standard_enemy_rating * min(2, len(sys_status.get("enemies_nearly_supplied", [])))
+    )
     debug("%s has regional+supply threat of %.1f", this_planet, threat_from_supply)
-    regional_threat = sys_status.get('regional_threat', 0) + threat_from_supply
+    regional_threat = sys_status.get("regional_threat", 0) + threat_from_supply
     if not regional_threat:  # no need for protection
         if pinfo.current_focus == PROTECTION:
             debug("Advising dropping Protection Focus at %s due to no regional threat", this_planet)
```
st-pa-fo, line 370: after looking at the code for almost half an hour, i came to this point here, and i believe this may be a candidate cause for why the volp-uglush and the setinon and the nightsiders native planets do not default to defense focus but instead industry. shouldnt there be some sort of checking here if the local species like or dislike that focus? i'll look some more at the code, maybe i'm completely wrong ...

st-pa-fo: see also https://github.com/freeorion/freeorion/pull/3665#issuecomment-1013861481

Cjkjvfnby: > after looking at the code for almost half an hour

You are quite fast in reading tons of lines.  This code is overcomplicated. :)

This is code for the AI player to handle its empire. That means that this planet is already conquered by the empire.  Changes here will not address your problem that natives are not defensive enough.

PR #3548, opened 2021-08-29
[AI] Use caches to speedup invasion calculations

Cjkjvfnby:
Cache repeating operations related to system network.  This is expensive operations on the server side, so it's easier to deal with them on AI side.

### DIFF
--- a/default/python/AI/ExplorationAI.py
+++ b/default/python/AI/ExplorationAI.py
@@ -10,6 +10,7 @@ from common.fo_typing import SystemId
 from EnumsAI import MissionType
 from freeorion_tools import get_fleet_position, get_partial_visibility_turn
 from target import TargetSystem
+from universe.system_network import systems_connected
 
 graph_flags = set()
 border_unexplored_system_ids = set()
@@ -167,7 +168,7 @@ def follow_vis_system_connections(start_system_id, home_system_id):
             status_info = [system_header]
 
         has_been_visible = get_partial_visibility_turn(cur_system_id) > 0
-        is_connected = universe.systemsConnected(cur_system_id, home_system_id, -1)  # self.empire_id)
+        is_connected = systems_connected(cur_system_id, home_system_id)
         status_info.append("    -- is%s partially visible" % ("" if has_been_visible else " not"))
         status_info.append("    -- is%s visibly connected to homesystem" % ("" if is_connected else " not"))
         if has_been_visible:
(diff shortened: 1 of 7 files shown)


geoffthemedio:
Seems to help. Dropped the longest turn time in a test from 200 s to 22 s

FYI, the profiler indicates it's still spending 60-70% of the AI's turn generating time in the C++ pathfinder code.

Cjkjvfnby:
> FYI, the profiler indicates it's still spending 60-70% of the AI's turn generating time in the C++ pathfinder code.

I have cached 2 functions here.  Do we have more API call that touches path finding?

I made a quick look into the Pathfinder.  

1) During AI turn this should be immutable.  So we can remove all mutexes for getters used by AI.
2) LeastJumpsPathImpl uses BFSVisitor, it doesn't looks like the best algorithm here. Dijkstra or A* suit it better.  
3) Some positive path are implemented as exceptions: `throw FoundDestination();`  Are exceptions  slow in this version of C++?

geoffthemedio:
> LeastJumpsPathImpl uses BFSVisitor, it doesn't looks like the best algorithm here. Dijkstra or A* suit it better.

It's counting jumps between systems, which are always the same "weight", so BFS = Dijkstra in this case. A heuristic based on distance between the start and end nodes could be implemented for A*, though whether it really helps would require testing.

> Some positive path are implemented as exceptions: `throw FoundDestination();`  Are exceptions  slow in this version of C++?

Possibly, or [possibly not](https://blog.plan99.net/what-s-wrong-with-exceptions-nothing-cee2ed0616?gi=6acdda8bf946), but it will only throw once when the search is done, so it shouldn't be a major timesink.

Really knowing why it is slow requires profiling, though. Changing the container types used in `EdgeVisibilityFilter` will be the first thing to try.

geoffthemedio:
After some additional tweaks, the profiler shows the following. It might also help to cache the results of finding the shortest path, as it spends about 26% of the time in `ShortestPath`.
AI Profiler HotPath ShortestPathImpl

It also spends ~5% of time in `ImmediateNeighbours`.

Cjkjvfnby:
Great. I will do the next optimization in the next branch, lets merge this one to get speed up of the next test build.

Cjkjvfnby:
Just to be clear: 

We have 2 function in the AI interface:
```
    def shortestPath(self, number1: int, number2: int, number3: int) -> IntVec: ...
    def shortestPathDistance(self, number1: int, number2: int) -> float: ...
```
In our case `shortestPathDistance` is a case, since `shortestPath` is not used.

PR #3552, opened 2021-08-29
Preparation to have multiple prefixes for codecs

Cjkjvfnby:
For turn 284 we might have about 25K prefixes.  
I want to use shorter ones.  This will save us a bit of space in save game. (I'd say a kilobyte or two)

I want to save with new and load with both, so I have extracted all prefix related code to a separate functions.   So basically I will need to change prefixes from list to tuple, update a 3 functions and add a bit of tests. 

image

### DIFF
--- a/default/python/AI/savegame_codec/_decoder.py
+++ b/default/python/AI/savegame_codec/_decoder.py
@@ -38,7 +38,17 @@ class SaveDecompressException(Exception):
     Exception class for troubles with decompressing save game string.
     """
 
-    pass
+
+def _starts_with_prefix(prefix: str, candidate: str) -> bool:
+    return candidate.startswith(prefix)
+
+
+def _extract_value(prefix: str, value: str):
+    return value[len(prefix) :]
+
+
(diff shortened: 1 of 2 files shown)


geoffthemedio:
>  This will save us a bit of space in save game. (I'd say a kilobyte or two)

This is a negligible amount compared to most compressed-xml saves I have... well under 1%, so don't prioritize it just for space savings.

Cjkjvfnby:
> This is a negligible amount compared to most compressed-xml saves I have... well under 1%, so don't prioritize it just for space savings.

Also save string will look a bit more accurate.    

Also there is some selfish point to do that: I am looking for ways on how to organize code in a way it will be easy adaptable and maintainable.

Just simple encapsulation:  only a couple of functions care about content of prefixes.  So, the prefix values are perfectly encapsulated. This gives as maintainability and adaptability.  I need to change a fewer places if I want to update them.  Information about format is still spread across the whole file, so this is not a perfect core organization but it's enough for now :).

PR #3553, opened 2021-08-30
Cache shortest distance

Cjkjvfnby:
@geoffthemedio,  I use the same approach for caching.


Cjkjvfnby:
I have cached top function that slows down the `generateOrders`, looks like other optimization places are in the Python code.

Cjkjvfnby:
These 3 lines on the first turn after loading takes about 10-15 times more then on second turn after load (I use 284 turn save.  first after loading: 10s, second after loading and next: 0.6s )
```
fo.initMeterEstimatesDiscrepancies()
fo.updateMeterEstimates(False)
fo.updateResourcePools()
```

PR #3554, opened 2021-08-31
Likes Kraken Nest and Metropoles error in UI

Grummel7:
LIKE-macros were missing in monster nests specials

MapWnd::SelectSystem called Universe::UpdateMeterEstimates(int object_id... which restricted the context to objects with the system. Metropoles looks for the 6 biggest most populous planets and this only works when the context contains all planets. So when clicking at a system with planets >20, they all showed the Metropoles bonus, even if not amongst the 6 biggest.


Grummel7:
With a fully visible standard map, 2/3 of all planets settled, the delay was about half a second. And my machine is not exactly brand new. Shouldn't be worth showing wrong values.

Note that the existing code already does recalculate everything when you changed the focus. Which just made me think: do we need this "recalculate when selected" at all? Recalculating everything when and only when something changes should be the natural way.

geoffthemedio:
>  do we need this "recalculate when selected" at all? Recalculating everything when and only when something changes should be the natural way.

It's not been strictly enforced, but the general idea was a mix of update on changes and update before showing, to be safe. If only updating when changing, then that relies on everything that could change a meter updating all the meters...

PR #3556, opened 2021-09-06
Colonize after conquer

Grummel7:
As discussed in the forum, these two should not require much discussion.

Note that the ? : operator in FOCS does not work as expected, that's why I had to turn it into two effects separated by scope conditions.
When I tried to replace the simple condition (Target.TurnsSinceLastConquered != 0) by a boolean expression using "And", I get a parse error "Expected real number expression here". I guess my first version only worked because the boolean expression I've used happened to start with a numeric value.

### DIFF
--- a/default/scripting/policies/INDOCTRINATION.focs.txt
+++ b/default/scripting/policies/INDOCTRINATION.focs.txt
@@ -10,25 +10,45 @@ Policy
         [[SPECIES_LIKES_OR_DISLIKES_POLICY_STABILITY_EFFECTS]]
 
         // makes all planets more stable over time at the cost of influence
+        // First variant: Current population was colonized.
+        // Planet was never conquered or it was conquered, emptied and re-colonised.
+        // Bonus is based on only how long indoctrination has been adapted.
         EffectsGroup
         scope = And [
             Planet
             OwnedBy empire = Source.Owner
             Species
+            Or [(LocalCandidate.TurnsSinceLastConquered == 0)
+                (LocalCandidate.TurnsSinceLastConquered > LocalCandidate.TurnsSinceColonization)]
         ]
         effects = [
             SetTargetHappiness value = Value + 
                 min((NamedReal name = "PLC_INDOCTRINATION_STRENGTH_MAX" value = 10.0),
-                    (Target.TurnsSinceLastConquered != 0 ?
-                     min(Target.TurnsSinceLastConquered,
-                         TurnsSincePolicyAdopted empire = Source.Owner name = ThisPolicy) :
-                     TurnsSincePolicyAdopted empire = Source.Owner name = ThisPolicy) *
-                     (NamedReal name = "PLC_INDOCTRINATION_PER_TURN" value = 0.25))
+                    (TurnsSincePolicyAdopted empire = Source.Owner name = ThisPolicy) *
+                    (NamedReal name = "PLC_INDOCTRINATION_PER_TURN" value = 0.25))
+            SetTargetInfluence value = Value - 1.0
+        ]
+        // Second variant: Current population was conquered.
+        // Bonus is additonally limited to turns since conquered.
+        EffectsGroup
+        scope = And [
+            Planet
+            OwnedBy empire = Source.Owner
+            Species
+            (LocalCandidate.TurnsSinceLastConquered != 0)
+            // Note: TurnsSinceColonization cannot be equal to TurnsSinceLastConquered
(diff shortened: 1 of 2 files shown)


geoffthemedio:
> the ? : operator in FOCS does not work as expected

Could you elaborate on that...?  Just the parsing is the issue?

> When I tried to replace the simple condition (Target.TurnsSinceLastConquered != 0) by a boolean expression using "And"

I don't know what this means. Could you give an example?

> I get a parse error "Expected real number expression here". I guess my first version only worked because the boolean expression I've used happened to start with a numeric value.

It's possible that if it expects a floating point number and you write an integer, it could have parse issues, though it generally should convert ints to floats when the latter is expected. You can easily add `.0` to the end of an int constant if that's the situaton, though...

default/scripting/policies/INDOCTRINATION.focs.txt
```diff
@@ -10,25 +10,45 @@ Policy
...
+        ]
+        // Second variant: Current population was conquered.
+        // Bonus is additonally limited to turns since conquered.
+        EffectsGroup
+        scope = And [
+            Planet
+            OwnedBy empire = Source.Owner
+            Species
+            // Note: TurnsSinceColonization cannot be equal to TurnsSinceLastConquered
+            And [(LocalCandidate.TurnsSinceLastConquered != 0)
```
geoffthemedio: what is the nested `And` for?

Grummel7: See comments before the two effect groups.
Perhaps I should move the comment into the scope, right in front of the Or/And to make it clearer?

Grummel7:
Okay, here's an example. This works fine:
`SetTargetHappiness value = Value + (Target.Population > 12 ? 2 : 1)`

So does this:
`SetTargetHappiness value = Value + (12 < Target.Population ? 2 : 1)`

This however yields "Parse error.  Expected real number expression here:" with the C of Capital marked as the parse position.
`SetTargetHappiness value = Value + (Capital ? 2 : 1)`

Just like "And", Capital is a boolean, which should be fine for the ternary operator. But the parser stops before even finding the operator. In the good examples the expressions start with numbers.

default/scripting/policies/INDOCTRINATION.focs.txt
geoffthemedio: The `Or` on line 21 is fine, but here you've essentially got `And [ C1 And [ C2 C3 ] ]` which is the same logic as `And [ C1 C2 C3 ]`

geoffthemedio:
> This however yields "Parse error. Expected real number expression here:" with the C of Capital marked as the parse position.
> `SetTargetHappiness value = Value + (Capital ? 2 : 1)`
> 
> Just like "And", Capital is a boolean, which should be fine for the ternary operator. But the parser stops before even finding the operator. In the good examples the expressions start with numbers.

This is due to a misunderstanding of how FOCS expressions work. `Capital` is a condition that matches objects, but here the parser expects a real-number valued expression, in the form of a value reference, like `12` or `Source.Owner` or `cos(CurrentTurn / 3.0)`. The ternary operator `? :` can be part of a number-valued expression, but it expects all three of its parameters to also be number-valued expressions, not conditions, including the one before `?`.

You can convert a condition into a number that can be used as a boolean value by using a statistic expression, like `Statistic If condition = Capital`. That will check if there are any objects in the universe that are a capital of an empire, and if there are, return 1, and if none are, return 0.

As written, that's probably not what's intended in most cases, so it's better to narrow the scope with an additional condition, like `Statistic If condition = And [ Target Capital ]`. This tests just the target object of the effect to see if it's a capital.

As a condition, `Capital` is used alone (not in a Statistic) in `scope =` or `activation =` or similar clauses in content definitions, where the game needs a criterion by which to filter or test objects, not to calculate a number.

There are a variety of much more interesting statistics than `Statistic If` available, which involve a condition to sample objects, and then an expression to evaluate on each of the objects that match the condition, and then a way to reduce the data (calculate a statistic) from those numbers. But `Statistic If` is a particularly useful one, in which the data reduction is just checking if there were any matches to the sampling condition.

default/scripting/policies/INDOCTRINATION.focs.txt
Grummel7: Ah, now I see. Well, I basically reverted the Or from the first condition, overlooking the fact that I am within an And condition. Will update it.

geoffthemedio:
sorry for the delay... was travelling and then sick and then distracted

PR #3558, opened 2021-09-11
Balance policy Metropoles

Grummel7:
Halved production and research bonus.
Require stability of 12.
No more than 1 in 4 of an empires populated planets can become a metropole.
Increased adaption cost.

### DIFF
--- a/default/scripting/policies/METROPOLES.focs.txt
+++ b/default/scripting/policies/METROPOLES.focs.txt
@@ -3,28 +3,36 @@ Policy
     description = "PLC_METROPOLES_DESC"
     short_description = "PLC_METROPOLES_SHORT_DESC"
     category = "ECONOMIC_CATEGORY"
-    adoptioncost = 1.0 * floor(5 + 0.25 * Statistic Sum value = LocalCandidate.Population
-                                          condition = And [ Planet OwnedBy empire = Source.Owner ])
+    adoptioncost = 1.0 * floor(15 + 0.25 * Statistic Sum value = LocalCandidate.Population
+                                           condition = And [ Planet OwnedBy empire = Source.Owner ])
     exclusions = [ "PLC_CONFEDERATION" "PLC_COLONIZATION" ]
     effectsgroups = [
         [[SPECIES_LIKES_OR_DISLIKES_POLICY_STABILITY_EFFECTS]]
 
-        // TODO: limit of planets based on number of regional admins?
-
-        // boosts to largest-population planets...
+        // Boost largest-population planets, provided they have enough population and stability.
+        // Metropoles need some "umland", so no more than 1 in 4 can become a metropole.
+        // Note that to raise a planet to a metropole, you have to reach the minimum stability
+        // while suffering from the non-metropole penalty below.
         EffectsGroup
-            scope = MaximumNumberOf number = (NamedInteger name = "METROPOLES_PLANET_LIMIT" value = 6) sortkey = LocalCandidate.Population condition = And [
+            scope = MaximumNumberOf number = min((NamedInteger name = "METROPOLES_PLANET_LIMIT" value = 6),
+                                                 floor(Statistic Count condition = And [
+                                                           Planet
+                                                           OwnedBy empire = Source.Owner
+                                                           Species
+                                                       ] / (NamedInteger name = "METROPOLES_MIN_PLANETS" value = 4)))
+                                    sortkey = LocalCandidate.Population condition = And [
                 Planet
                 OwnedBy empire = Source.Owner
                 Population low = (NamedReal name = "METROPOLES_MINIMUM_POPULATION" value = 20.0)
+                Happiness low = (NamedReal name = "METROPOLES_MIN_STABILITY" value = 12)
             ]
             effects = [
                 SetTargetIndustry value = Value + Target.Population
                     * (NamedReal name = "METROPOLES_TARGET_INDUSTRY_PERPOP"
-                                 value = 2.0 * [[INDUSTRY_PER_POP]])
+                                 value = [[INDUSTRY_PER_POP]])
                 SetTargetResearch value = Value + Target.Population
                     * (NamedReal name = "METROPOLES_TARGET_RESEARCH_PERPOP"
-                                 value = 2.0 * [[RESEARCH_PER_POP]])
+                                 value = [[RESEARCH_PER_POP]])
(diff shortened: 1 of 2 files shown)


geoffthemedio:
> No more than 1 in 4 of an empires populated planets can become a metropole.

What do you think about a quadratic scaling of the number of populated planets? So the max number would depend on the square root of the number of populated planets, not linearly with it.

This is supposed to be a policy for small empires, so (even with the maximum of 6 total), so adding a scaling requirement per additional metropole might make sense...?

Grummel7:
> > No more than 1 in 4 of an empires populated planets can become a metropole.
> 
> What do you think about a quadratic scaling of the number of populated planets? So the max number would depend on the square root of the number of populated planets, not linearly with it.
> 
> This is supposed to be a policy for small empires, so (even with the maximum of 6 total), so adding a scaling requirement per additional metropole might make sense...?

Well, the problem I see with this is that when you can have one metropole from the start, it again becomes almost compulsory to get it as fast as you can. Even with the bonuses reduced, getting about +5/+5 in turn 10 is just too good to ignore. If we require one additional colony, it's even worse, because then depends only on whether you find a planet to drop you initial colony ship to.

Therefore I suggest to require at least 3, better 4 for the first metropole. Of course the number could grow. E.g. we could require one more planet for each further metropole, i.e. 4 for 1, 9 for 2, 15 for 3, etc. or 3 for 1, 7 for 2, 12 for 3, etc.

default/stringtables/en.txt
```diff
@@ -11519,15 +11519,16 @@ PLC_METROPOLES
 Metropoles
 
 PLC_METROPOLES_DESC
-'''Boosts the output of up to [[value METROPOLES_PLANET_LIMIT]] planets with population of at least [[value METROPOLES_MINIMUM_POPULATION]]. The highest-[[metertype METER_POPULATION]] of the qualified planets will receive, regardless of focus:
-• Increases [[metertype METER_TARGET_RESEARCH]] by [[value METROPOLES_TARGET_RESEARCH_PERPOP]] per [[metertype METER_POPULATION]].
+'''Boosts the output of up to [[value METROPOLES_PLANET_LIMIT]] planets (but no more than 1 / [[value METROPOLES_MIN_PLANETS]] or your populated planets) with [[metertype METER_POPULATION]] of at least [[value METROPOLES_MINIMUM_POPULATION]] and [[metertype METER_HAPPINESS]] of at least [[value METROPOLES_MIN_STABILITY]].
```
geoffthemedio, line 11522: don't refer to the player in the second person, eg. "your planets".

default/scripting/policies/METROPOLES.focs.txt
```diff
@@ -3,28 +3,36 @@ Policy
...
+                                                 floor(Statistic Count condition = And [
+                                                           Planet
+                                                           OwnedBy empire = Source.Owner
+                                                           Species
+                                                       ] / (NamedInteger name = "METROPOLES_MIN_PLANETS" value = 4)))
+                                    sortkey = LocalCandidate.Population condition = And [
                 Planet
                 OwnedBy empire = Source.Owner
                 Population low = (NamedReal name = "METROPOLES_MINIMUM_POPULATION" value = 20.0)
+                Happiness low = (NamedReal name = "METROPOLES_MIN_STABILITY" value = 12)
```
geoffthemedio, line 27: I'm a bit concerned about having the selection of which planets from those eligible be just based on population, when also filtering based on population and happiness. Not sure if it will lead to a weird / unexpected situation, or if it will be fine...

default/stringtables/en.txt
```diff
@@ -11519,15 +11519,16 @@ PLC_METROPOLES
...
+The highest-populated of the qualified planets will receive, regardless of focus:
+• Increases [[metertype METER_TARGET_RESEARCH]] by [[value METROPOLES_TARGET_RESEARCH_PERPOP]] per population.
 • Increases [[metertype METER_TARGET_INDUSTRY]] by [[value METROPOLES_TARGET_INDUSTRY_PERPOP]] per population.
 • Increases [[metertype METER_TARGET_CONSTRUCTION]] by [[value METROPOLES_TARGET_CONSTRUCTION_FLAT]].
 • Increases [[metertype METER_MAX_SUPPLY]] by [[value METROPOLES_MAX_SUPPLY_FLAT]].
 but also has a penalty:
 • Decreases [[metertype METER_MAX_STOCKPILE]] by [[value METROPOLES_MAX_STOCKPILE_FLAT]].
 
-Additionally, other populated planets (with lower than the highest population planets, including all with less than [[value METROPOLES_MINIMUM_POPULATION]] [[metertype METER_POPULATION]]) have a penalty applied:
+Additionally, other planets (with lower than the highest population planets, including all with less than [[value METROPOLES_MINIMUM_POPULATION]] population or less than [[value METROPOLES_MIN_STABILITY]] stability) have a penalty applied:
```
geoffthemedio, line 11531: why do you remove the metertype link for population (and not add one for stability)?

geoffthemedio:
> I suggest to require at least 3, better 4 for the first metropole.

No objections there...

default/stringtables/en.txt
Grummel7: Well, that sentence will probably change anyway. How you you write it? ...of all populated planets)

default/scripting/policies/METROPOLES.focs.txt
Grummel7: I found it odd that when you conquer a planet, you get the bonus immediately. Almost all bonuses have a minimum stability, why not this? You have to watch two values instead of one, but Metropoles are important enough to take a little care.

default/stringtables/en.txt
Grummel7: Wikipedia style: Link a word only where it appears for the first time.

default/scripting/policies/METROPOLES.focs.txt
geoffthemedio: I don't have a specific problem, just a vague concern, so it's fine for now...

default/stringtables/en.txt
geoffthemedio: refer to "of an empire" or "the empire's" or similar, rather than "your"

geoffthemedio:
https://github.com/freeorion/freeorion/commit/cceb39c8cc2ba24daf57f76e34f16abb6a9fdc5b

PR #3560, opened 2021-09-11
Fix star based colonization value calculation

Morlic-fo:
The larger boni should apply if the AI does not yet have a star of that type. The defect was introduced in #3484


Cjkjvfnby (approved):
Good catch!

PR #3568, opened 2021-09-29
Fix python parser on Android

o01eg:


BUILD.md
```diff
@@ -203,9 +203,14 @@ the `freeorion-project/build` directory.
 
 Install Android NDK.
 
-Build [Python-For-Android].
+Build [Python-For-Android]. Pack standart library with
```
geoffthemedio:
```suggestion
Build [Python-For-Android]. Pack standard library with
```

util/Directories.cpp
```diff
@@ -103,6 +104,27 @@ void RedirectOutputLogAndroid(int priority, const char* tag, int fd)
     });
     background.detach();
 }
+
+void CopyInitialResourceAndroid(const std::string& rel_path)
```
geoffthemedio, line 108:
```suggestion
void CopyInitialResourceAndroid(std::string_view rel_path)
```

util/Directories.cpp
```diff
@@ -476,6 +498,10 @@ void InitDirs(std::string const& argv0)
 
     RedirectOutputLogAndroid(ANDROID_LOG_ERROR, "stderr", 2);
     RedirectOutputLogAndroid(ANDROID_LOG_INFO, "stdout", 1);
+
+    s_python_home = s_cache_dir / "python";
+    fs::create_directories(s_python_home / "lib");
+    CopyInitialResourceAndroid("lib/python36.zip");
```
geoffthemedio, line 504: called here with `char*` literal so passing as `const string&` does redundant `string` construction probly

util/Directories.cpp
o01eg: Then I got `error: invalid operands to binary expression ('fs::path' and 'std::string_view' (aka 'basic_string_view<char>'))`

BUILD.md
o01eg: Fixed.

util/Directories.cpp
o01eg: Maybe start to use `std::filesystem::path` instead of `boost::filesystem::path` to get support for string views?

geoffthemedio: There might be compiler support issues still, eg. on older Apple Clang versions.

Edit: `std::filesystem` requires Apple Clang 11 apparently: https://en.cppreference.com/w/cpp/compiler_support

o01eg: I see, Android toolchain have only Clang 9.

o01eg: @geoffthemedio Maybe merge it as is while boost doesn't supports string view for paths?

geoffthemedio: > Android toolchain have only Clang 9.

Apple Clang and other Clang are separarte. That same table lists Clang as supporting `std::filesystem` from version 7.

> merge it as is

feel free...

PR #3570, opened 2021-10-03
Add support for Python's asyncio

o01eg:
Adds periodical call to asyncio event loop so scheduled callbacks and coroutines could be executed. Disabled by default.


geoffthemedio:
some comments explaining what code does and why might help. otherwise I don't have any opinion. not sure what it should do...

o01eg:
@geoffthemedio Added more comments and link to https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_forever

Should be usable with based on asyncio code where we don't wait network request to be completed within API call, but could start it in API call and be completed in asyncio event loop.

geoffthemedio:
Don't really understand but it's probly fine...

PR #3574, opened 2021-10-05
Fix reading game rules values from options

### LINKED ISSUE #3573: Parsed game rules couldn't be set from options
Bug Report
==========

Environment
-----------

* **FreeOrion Version**: 2021-10-05.980fd8e
* **Operating System**: Ubuntu Linux 20.04
* **Fetched as** 
  * Compiled from source

Description
-----------

After https://github.com/freeorion/freeorion/commit/9a0240fc04fb87d719f6c9bbce2203aff19963c3 game rules values from options got ignored

Expected Result
---------------

Games reads game rules values from options.

Steps to reproduce
------------------

* Start the game with `--setup.rules.RULE_ENABLE_SUPER_TESTER 0` options.
* Check rules in encyclopedia.

o01eg:
Fixes #3573

### DIFF
--- a/util/GameRules.cpp
+++ b/util/GameRules.cpp
@@ -139,6 +139,7 @@ void GameRules::Add(Pending::Pending<GameRulesTypeMap>&& future)
 { m_pending_rules = std::move(future); }
 
 namespace {
+    /** Sets default validator, ensures option and sets value to value of the option. */
     template <typename T>
     void CheckValidatorAndAddRuleOption(GameRule& rule)
     {
--- a/util/GameRules.cpp
+++ b/util/GameRules.cpp
@@ -146,9 +146,10 @@ namespace {
             rule.validator = std::make_unique<Validator<T>>();
         auto option_name = "setup.rules." + rule.name;
         if (!GetOptionsDB().OptionExists(option_name))
-            GetOptionsDB().Add<T>(std::move(option_name), rule.description,
+            GetOptionsDB().Add<T>(option_name, rule.description,
                                   boost::any_cast<T>(rule.default_value),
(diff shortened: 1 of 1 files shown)


util/GameRules.cpp
```diff
@@ -146,9 +146,10 @@ namespace {
             rule.validator = std::make_unique<Validator<T>>();
         auto option_name = "setup.rules." + rule.name;
         if (!GetOptionsDB().OptionExists(option_name))
-            GetOptionsDB().Add<T>(std::move(option_name), rule.description,
+            GetOptionsDB().Add<T>(option_name, rule.description,
                                   boost::any_cast<T>(rule.default_value),
                                   rule.validator->Clone());
+        rule.value = GetOptionsDB().Get<T>(option_name);
```
geoffthemedio, line 153: Does this always need to be set, or just if there was already an existing option for that option name? If the latter, would be clearer to put it in an `else` case...

geoffthemedio: also should add a comment above this function that indicates that it may modify the passed-in `rule` by changing the value if an option already exists for that rule's value.

o01eg: @geoffthemedio It should be set because even if options was already registered its value could be different.

o01eg: Added comment about modification.

geoffthemedio: My point was that if there was no option yet registered, then is there any need to set the rule value from the newly-registered option?

o01eg: Yes, because options from XML not registered yet but already have value and marked with `recognized = false`. To access have still required to call `Add` method.

PR #3575, opened 2021-10-12
Halved effect of good/bad stability

Grummel7:


geoffthemedio:
I'd prefer without the numbers with fractional parts after the decimal, but OK...

Issue #3578, opened 2021-10-15
PLC_FLANKING stationary bonus is broken


agrrr3:
a bug in the flanking policy - defense bonus only applies if ETA=1and stationary. That can never happen.

Issue #3577, opened 2021-10-14
Give Combat Bonus from Policies also to Fighters and Arc Disruptor lines (and maybe monsters?)


agrrr3:
Feature request
===============

Idea
----
I just recognized the combat policies FLANKING and CHARGE only work for the ship weapons, not fighters nor arc disruptors (also not special monster weapons). Probably that is wrong. The bonus is 1*scaling - for mass drivers this means 1/3rd vs mass drivers zero.

That would directly translate to 2/3*scaling for arc disruptors, but one could add a shot instead.

For flak one could also think of adding a shot, but it does not fit the theme so I rather wouldn't. Actually CHARGE could/should lower the number of flak shots.

For fighters strikers have a bit higher base damage than mass drivers (4 vs 3), so adding 1x scaling is rather conservative. If taking bombers (6)as a reference point, adding 2x scaling would keep the balance. So I'd suggest adding 2x scaling for all fighter types for PLC_FLANKING. So I suggest increasing launch capacity by 1 instead. This will make CHARGE especially good for heavier designs and FLANKING good for lighter fighters. 
To stress the good-for-lighter-fighters, FLANKING interceptors should get a +2 launch capacity bonus.

For monster weapons dunno. Fluff-wise a small bonus would make sense (trained monsters not following tactics/orders as well as your pilots do). But also no bonus or normal bonus could be argued.

Meta
----
Alternative to giving the bonus to all weapon lines is having specialised policies. Or distinguish a weapon line being more/less dependent on the policies (e.g. like arc disruptors do not depend on pilot skill) but balance that with extra power for those who are not able to get a bonus. 

Also like Oberlus wrote, the policies currently boost the lower tier ship weapons more significantly (which I think is good and interesting).

The three weapon trees are orthogonal. So a typical empire is expected to invest in one of those heavily and investing in multiple lines has slighter return on investment. It makes sense NOT to have specialised policies.

Again for monster weapons dunno.

The CHARGE policy applies a malus - that means a flak-only ship will get a malus but no bonus. But to be efficient on point defense fluff-wise the ship has to charge with the others. For a pure carrier, it might make sense to stay behind, so having no bonus would be more dubious.

Fluffwise it makes not much sense that a faster carrier increases the damage of the fighters, so we increase launch rate instead.

[...]

PR #3582, opened 2021-10-17
policy combat bonus for other weapons

agrrr3:
discussion started in the forums https://freeorion.org/forum/viewtopic.php?p=108138#p108138

Features:
- implementation for #3577 (Give Combat Bonus from Policies also to Fighters and Arc Disruptor lines)
- bugfix for #3578 (PLC_FLANKING stationary bonus is broken)
- also adds pincer attack description for PLC_FLANKING

Changes:
PLC_FLANKING
- Weapon damage for strikers, bombers and heavy bombers is increased by 12 per fighter
- Launch rate for interceptors is increased by 2 per bay
- Weapon shots for arc disruptors is increased by 1 
- Defensive flanking does not expect ETA1 anymore (bugfix)

PLC_CHARGE
- Launch rate for any type of fighter is increased by 1 per bay
- Weapon shots for arc disruptors is increased by 1 
- Weapon shots for flak is decreased by 1 

Following text adapted from the
Feature request
===============

Idea
----
I just recognized the combat policies FLANKING and CHARGE only work for the ship weapons, not fighters nor arc disruptors (also not special monster weapons). Probably that is wrong. The bonus is 1*scaling - for mass drivers this means 1/3rd vs mass drivers one tech.

That would directly translate to 2/3*scaling for arc disruptors, but one could add a shot instead.

For flak one could also think of adding a shot, but it does not fit the theme so I rather wouldn't. Actually CHARGE could/should lower the number of flak shots.

For fighters strikers have a bit higher base damage than mass drivers (4 vs 3), so adding 1x scaling is rather conservative. If taking bombers (6)as a reference point, adding 2x scaling would keep the balance. So I'd suggest adding 2x scaling for all fighter types for PLC_FLANKING. And for CHARGE I suggest increasing launch capacity by 1 per bay instead. This will make CHARGE especially good for heavier designs and FLANKING good for lighter fighters. Combining CHARGE and FLANKING makes a sweet spot for striker fighters (a bay able to launch a complete hangar and also getting the most damage out of FLANKING).
To stress the good-for-lighter-fighters, FLANKING interceptors should get a +2 launch capacity bonus.

For monster weapons dunno. Fluff-wise a small bonus would make sense (trained monsters not following tactics/orders as well as your pilots do). But also no bonus or normal bonus could be argued.

Meta
----
Alternative to giving the bonus to all weapon lines is having specialised policies. Or distinguish a weapon line being more/less dependent on the policies (e.g. like arc disruptors do not depend on pilot skill) but balance that with extra power for those who are not able to get a bonus. 

Also like Oberlus wrote, the policies currently boost the lower tier ship weapons more significantly (which I think is good and interesting).

The three weapon trees are orthogonal. So a typical empire is expected to invest in one of those heavily and investing in multiple lines has slighter return on investment. It makes sense NOT to have specialised policies.

Again for monster weapons dunno.

The CHARGE policy applies a malus - that means a flak-only ship will get a malus but no bonus. But to be efficient on point defense fluff-wise the ship has to charge with the others. For a pure carrier, it might make sense to stay behind, so having no bonus would be more dubious.

Fluffwise it makes not much sense that a faster carrier increases the damage of the fighters, so we increase launch rate instead.

Fluffwise, a charge bonus could interact with "range", so e.g. a charging vessel could shoot a bout earlier (and carriers field interceptors earlier). Seeing this as a nice-to-have, the feature would be too complex (interaction with stealth, fighter launch,..) and also hard to script.

Fluffwise, a charge bonus probably should wear off when already in close combat range. Also not worth it - players would probably surprised that the bonus does not kick in.

It is also not clear yet how the policies interact with damage estimation. Probably the game should show the bonus when it is expected to apply, but distinct such a "temporary effect" from the "base effects". I think thats not possible currently.

@geoffthemedio Also NOTE defensive flanking basically always applies if no enemy is present, so if a larger force arrives, the first turn the defender will get the bonus (which is probably not what we want?).

### DIFF
--- a/default/scripting/policies/CHARGE.focs.txt
+++ b/default/scripting/policies/CHARGE.focs.txt
@@ -7,7 +7,8 @@ Policy
     effectsgroups = [
         [[SPECIES_LIKES_OR_DISLIKES_POLICY_STABILITY_EFFECTS]]
 
-        // fast ships do more damage
+        // This is policy which trains your military to efficiently plan and execute charge attacks, i.e.:
+        // fast ships do more damage and fast carriers field fighters more efficiently/closer to the enemy
         EffectsGroup
             scope = And [
                 Ship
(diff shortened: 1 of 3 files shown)


geoffthemedio:
> @geoffthemedio Also NOTE defensive flanking basically always applies if no enemy is present, so if a larger force arrives, the first turn the defender will get the bonus (which is probably not what we want?).

There's not really an easy fix for that, as it effectively calculates the bonus before movement, and it's generally preferable to have numbers for things like ship damage be accurately predicted in the client, even if it doesn't know what other empires will do with their fleets before the next combat happens (though not necessarily always, as stealth can impact results of client-side predictions, but there should be "Unknown" meter factors then...). Generally the numbers in the UI are meant to be accurate for the next interturn, though.

Also, it's not so unreasonable... The defender can position their ships around the starlane entry point, regardless of how many there are coming out of it.

geoffthemedio:
> It is also not clear yet how the policies interact with damage estimation. Probably the game should show the bonus when it is expected to apply, but distinct such a "temporary effect" from the "base effects". I think thats not possible currently.

It would probably depend on how the effect is scripted and what owner empire is specified for the test ship used for damage estimation calculations. If the test ship's empire is a valid empire ID, then a test for if the ship's owner has a policy adopted should match that ship. If it's not a valid empire o ris another empire that doesn't (or isn't known to) have the policy adopted, then the ship wouldn't match and wouldn't get the bonus in the estimate, I suppose.

geoffthemedio:
Is this done or do you plan to do more tweaks before it should be merged?

agrrr3:
> > @geoffthemedio Also NOTE defensive flanking basically always applies if no enemy is present, so if a larger force arrives, the first turn the defender will get the bonus (which is probably not what we want?).
> 
> There's not really an easy fix for that, as it effectively calculates the bonus before movement, and it's generally preferable to have numbers for things like ship damage be accurately predicted in the client,

three options: 

1. keep it like it is, maybe speed up the calculation for the default case (if there is no enemy ship in the system apply the bonus); maybe rename the policy
2. delay the bonus a turn, base it on the information from the battle; fluff would be that it takes a while to organise the fleet according to the enemy - this handles visibility much better
3. introduce "combat preparation" effects in the backend etc which are only processed when a battle starts or rather when a bout starts (and for estimation)

The combat prepartion effects would be the most work, but suits flanking really well. It would also be easy to treat combat preparation effects differently in the GUI. 
CHARGING is KISS and good as it is in my opinion and does not really benefit from such though. But stationary flanking is complicated because in the estimation, you should not count stealthed enemies, but in combat, stealthed enemies should count.

For stationary flanking it would correctly handle visibility (only ships participating in battle would be counted) and could, if we want, be dynamic (that is, if you loose your advantage in numbers in bout two because your enemy shot most of your fleet down, you dont get the bonus anymore).
It would also make the pincer move effect less awkward, you would not have to be in-flight a turn before in order to get the bonus (so you one-jump pincer moves would work as well).

If we keep it as it is. Would a OR-ing Statistic Not If condition = (VISIBLE_)ENEMY_SHIP_IN_SYSTEM speed up things?

> 
> Also, it's not so unreasonable... The defender can position their ships around the starlane entry point, regardless of how many there are coming out of it.

Yes, but maybe we should rename it. The effect is a bonus as long as your defense is not overrun by sheer amount of enemies - so the smell is rather a wall than a dynamic manoveur. OTOH the pincer move would not fit the policy fluff anymore.

Worst thing about this is that the presence of stealthed enemy ship leaks through, or if we take visibility into account there are cases where the stealthed fleet is not able to reap the effect it should (that is being able to overwhelm the FLANKING enemy right from the start of battle).

> Is this done or do you plan to do more tweaks before it should be merged?

I fixed a minor issue. Else, for the moment, yes, please merge.

PR #3586, opened 2021-10-19
Python parser refactor

o01eg:


parse/ConditionPythonParser.h
```diff
@@ -5,6 +5,12 @@
 
 #include "../universe/Condition.h"
 
+namespace boost {
```
geoffthemedio: should be able to do `namespace boost::python {`

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        std::vector<std::pair<std::string, std::unique_ptr<ValueRef::ValueRef<std::string>>>> parameters;
+        if (kw.has_key("parameters")) {
+            boost::python::dict param_args = boost::python::extract<boost::python::dict>(kw["parameters"]);
+            boost::python::stl_input_iterator<std::string> p_begin(param_args.keys()), p_end;
+            for (auto it = p_begin; it != p_end; ++it) {
+                auto p_arg = boost::python::extract<value_ref_wrapper<std::string>>(param_args[*it]);
+                if (p_arg.check()) {
+                    parameters.push_back(std::make_pair(*it, ValueRef::CloneUnique(p_arg().value_ref)));
```
geoffthemedio: should be able to replace `make_pair(` with `pair{` in most/all cases

or do `parameters.emplace_back(*it, ValueRef::CloneUnique(p_arg().value_ref));`

parse/ConditionPythonParser.h
o01eg: Fixed.

parse/EffectPythonParser.cpp
o01eg: Fixed.

o01eg:
@geoffthemedio  Is it ready to be merged?

parse/ConditionPythonParser.cpp
```diff
@@ -22,3 +29,135 @@ return condition_wrapper(std::make_shared<Condition::Not>(
...
+    condition_wrapper insert_contained_by_(const condition_wrapper& cond) {
+        return condition_wrapper(std::make_shared<Condition::ContainedBy>(ValueRef::CloneUnique(cond.condition)));
+    }
+
+    condition_wrapper insert_contains_(const condition_wrapper& cond) {
+        return condition_wrapper(std::make_shared<Condition::Contains>(ValueRef::CloneUnique(cond.condition)));
+    }
+
+    template <MeterType M>
+    condition_wrapper insert_meter_value_(const boost::python::tuple& args, const boost::python::dict& kw) {
```
geoffthemedio: why `MeterType M` a template parameter, rather than a standard function parameter?

parse/ValueRefPythonParser.cpp
```diff
@@ -66,3 +73,88 @@ condition_wrapper operator==(const value_ref_wrapper<int>& lhs, const value_ref_
...
+
+void RegisterGlobalsValueRefs(boost::python::dict& globals, const PythonParser& parser) {
+    globals["NamedRealLookup"] = boost::python::raw_function(insert_named_lookup_<double>);
+    globals["Value"] = value_ref_wrapper<double>(std::make_shared<ValueRef::Variable<double>>(ValueRef::ReferenceType::EFFECT_TARGET_VALUE_REFERENCE));
+    globals["CurrentTurn"] = value_ref_wrapper<int>(std::make_shared<ValueRef::Variable<int>>(ValueRef::ReferenceType::NON_OBJECT_REFERENCE, "CurrentTurn"));
+
+    std::function<boost::python::object(const boost::python::tuple&, const boost::python::dict&)> f_insert_game_rule = [&parser](const boost::python::tuple& args, const boost::python::dict& kw) { return insert_game_rule_(parser, args, kw); };
+    globals["GameRule"] = boost::python::raw_function(f_insert_game_rule);
+    std::function<boost::python::object(const boost::python::tuple&, const boost::python::dict&)> f_insert_min = [&parser](const boost::python::tuple& args, const boost::python::dict& kw) { return insert_minmaxoneof_<ValueRef::OpType::MINIMUM>(parser, args, kw); };
+    globals["Min"] = boost::python::raw_function(f_insert_min, 3);
```
geoffthemedio, line 156: I guess this is refactoring, so maybe this is better for a separate change, but `"Min"` seems overly general and nonspecific here. Maybe store it with the key `"MinOperation"`?

geoffthemedio:
This seems to still run. It's refactoring, so various nitpicks I have that also apply to the previous version aren't really necessary to deal with as part of this, and would rather be handled in another set of changes. I don't want to / think I need to re-read all the code that was just moved around...

parse/ValueRefPythonParser.cpp
o01eg: Currently I prefer to reproduce FOCS as close as possible so I tries to remain with same tokens,

parse/ConditionPythonParser.cpp
o01eg: It was simple to write. I've changed it now.

parse/ValueRefPythonParser.cpp
geoffthemedio: I didn't think the key affected how the parser worked...

geoffthemedio:
OK then

PR #3587, opened 2021-10-19
Test techs checksum

o01eg:
Before I continue to convert techs to Python this PR adds check for techs chcksum

### DIFF
--- a/test/system/TestChecksum.cpp
+++ b/test/system/TestChecksum.cpp
@@ -7,7 +7,8 @@ BOOST_FIXTURE_TEST_SUITE(TestChecksum, ClientAppFixture)
 BOOST_AUTO_TEST_CASE(compare_checksum) {
     auto checksums = CheckSumContent();
 
-    BOOST_WARN_EQUAL(checksums["NamedValueRefManager"], 1042814);
+    BOOST_WARN_EQUAL(checksums["NamedValueRefManager"], 42021);
+    BOOST_WARN_EQUAL(checksums["TechManager"], 9095847);
 }
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/test/system/TestChecksum.cpp
+++ b/test/system/TestChecksum.cpp
@@ -8,8 +8,7 @@ void TestCheckSumFromEnv(const char* env, unsigned int def, unsigned int expecte
     bool force = false;
     unsigned int value = def;
 
-    const char *env_value = std::getenv(env);
-    if (env_value) {
+    if (const char *env_value = std::getenv(env)) {
         force = true;
         try {
             value = boost::lexical_cast<unsigned int>(env_value);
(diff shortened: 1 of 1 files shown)


geoffthemedio:
consistent with my checksums

test/system/TestChecksum.cpp
```diff
@@ -4,10 +4,37 @@
...
+    }
+    if (force) {
+        BOOST_REQUIRE_MESSAGE(expected == value, env << " expected " << expected << " was " << value);
+    } else {
+        BOOST_WARN_MESSAGE(expected == value, env << " expected " << expected << " was " << value);
+    }
+}
+
+/**
+ * - Enforces named value reference checksum test if FO_CHECKSUM_NAMED_VALUEREF set.
```
geoffthemedio:
```suggestion
 * - Enforces named value reference checksum test if FO_CHECKSUM_NAMED_VALUEREF is set.
```

test/system/TestChecksum.cpp
```diff
@@ -4,10 +4,37 @@
...
+ * - Enforces techs checksum test if FO_CHECKSUM_TECH set.
```
geoffthemedio:
```suggestion
 * - Enforces techs checksum test if FO_CHECKSUM_TECH is set.
```

test/system/TestChecksum.cpp
```diff
@@ -4,10 +4,37 @@
...
+ * - If environment variable set to integer value that value test with calculated checksum.
```
geoffthemedio:
```suggestion
 * - Setting each of these environment variables to an integer value will test that value against the corresponding parsed content checksum.
```

test/system/TestChecksum.cpp
```diff
@@ -4,10 +4,37 @@
 
 BOOST_FIXTURE_TEST_SUITE(TestChecksum, ClientAppFixture)
 
+void TestCheckSumFromEnv(const char* env, unsigned int def, unsigned int expected) {
+    bool force = false;
+    unsigned int value = def;
+
+    const char *env_value = std::getenv(env);
+    if (env_value) {
```
geoffthemedio:
```suggestion
    if (const char *env_value = std::getenv(env)) {
```

test/system/TestChecksum.cpp
```diff
@@ -4,10 +4,37 @@
...
+        force = true;
```
geoffthemedio, line 12: should `force` be set before the `try` and in the case that the `lexical_cast` throws?

test/system/TestChecksum.cpp
```diff
@@ -4,10 +4,37 @@
 
 BOOST_FIXTURE_TEST_SUITE(TestChecksum, ClientAppFixture)
 
+void TestCheckSumFromEnv(const char* env, unsigned int def, unsigned int expected) {
```
geoffthemedio, line 7: so `def` is a backup value to use for the check if the `getenv` lookup fails or the cast throws?

o01eg: Yes, it is a default value.

test/system/TestChecksum.cpp
o01eg: No, in that case it should force check default value.

test/system/TestChecksum.cpp
o01eg: Fixed.

test/system/TestChecksum.cpp
o01eg: Fixed.

test/system/TestChecksum.cpp
o01eg: Fixed.

geoffthemedio:
otherwise OK

PR #3588, opened 2021-10-20
Make FreeOrionNode global so it can be used in other scenes

o01eg:


client/godot/FreeOrionNode.cpp
```diff
@@ -391,3 +401,8 @@ godot::Dictionary FreeOrionNode::get_fleets() const {
     return fleets;
 }
 
+void FreeOrionNode::send_chat_message(godot::String text) {
+    std::string text8 = std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>, wchar_t>{}.to_bytes(text.unicode_str());
```
geoffthemedio: Does Godot use UTF-16 internally on all platforms?

geoffthemedio:
Otherwise no comment

client/godot/FreeOrionNode.cpp
o01eg: No, it uses `wchar_t` but `std::codecvt_utf8_utf16` looks like works both with Linux 4-bytes and Windows UTF-16 representation.

o01eg: Refactored it.

PR #3589, opened 2021-10-21
Convert some construction techs to Python

o01eg:


parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+        std::vector<std::pair<std::string, std::unique_ptr<ValueRef::ValueRef<std::string>>>> parameters;
+        if (kw.has_key("parameters")) {
+            boost::python::dict param_args = boost::python::extract<boost::python::dict>(kw["parameters"]);
+            boost::python::stl_input_iterator<std::string> p_begin(param_args.keys()), p_end;
+            for (auto it = p_begin; it != p_end; ++it) {
+                auto p_arg = boost::python::extract<value_ref_wrapper<std::string>>(param_args[*it]);
+                if (p_arg.check()) {
+                    parameters.emplace_back(*it, ValueRef::CloneUnique(p_arg().value_ref));
+                } else {
```
geoffthemedio: this cascading multiple nesting of ifs is a bit awkward... could probably avoid by using if initializers like
````
                if (auto p_arg = boost::python::extract<value_ref_wrapper<std::string>>(param_args[*it]); p_arg.check()) {
                    parameters.emplace_back(*it, ValueRef::CloneUnique(p_arg().value_ref));
                } else if (auto p_arg_double = boost::python::extract<value_ref_wrapper<double>>(param_args[*it]); p_arg_double.check()) {
                    parameters.emplace_back(*it, std::make_unique<ValueRef::StringCast<double>>(ValueRef::CloneUnique(p_arg_double().value_ref)));
                } else if (auto p_arg_int = boost::python::extract<value_ref_wrapper<int>>(param_args[*it]); p_arg_int.check())
                    parameters.emplace_back(*it, std::make_unique<ValueRef::StringCast<int>>(ValueRef::CloneUnique(p_arg_int().value_ref)));
                } else {
                    parameters.emplace_back(*it, std::make_unique<ValueRef::Constant<std::string>>(boost::python::extract<std::string>(param_args[*it])));
                }
````

File: parse/EffectPythonParser.cpp [hunk 2]
```diff
@@ -0,0 +1,175 @@
...
+#include <boost/python/stl_iterator.hpp>
+
+#include "../universe/Effects.h"
+#include "../universe/Enums.h"
+
+#include "EnumPythonParser.h"
+#include "ValueRefPythonParser.h"
+
+namespace {
+    effect_wrapper insert_generate_sit_rep_message_(const boost::python::tuple& args, const boost::python::dict& kw) {
```
geoffthemedio: probably warrants `using namespace boost::python;` here to save a lot of redundant namespace specifications in this code.

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,175 @@
...
+            affiliation = boost::python::extract<enum_wrapper<EmpireAffiliationType>>(kw["affiliation"])().value;
+        }
+
+        std::unique_ptr<Condition::Condition> condition;
+        if (kw.has_key("condition")) {
+            condition = ValueRef::CloneUnique(boost::python::extract<condition_wrapper>(kw["condition"])().condition);
+        }
+
+        return effect_wrapper(std::make_shared<Effect::GenerateSitRepMessage>(
+                                    message,
```
geoffthemedio: `message` and `icon` can probably be moved from...?

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,175 @@
...
+    effect_wrapper insert_generate_sit_rep_message_(const boost::python::tuple& args, const boost::python::dict& kw) {
+        auto message = boost::python::extract<std::string>(kw["message"])();
+        auto label = boost::python::extract<std::string>(kw["label"])();
+
+        bool stringtable_lookup = true;
+        if (kw.has_key("NoStringtableLookup")) {
+            stringtable_lookup = false;
+        }
+
+        std::string icon = "";
```
geoffthemedio: remove ` = ""`

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    template <MeterType M>
+    effect_wrapper insert_set_meter_(const boost::python::tuple& args, const boost::python::dict& kw) {
+        auto value = boost::python::extract<value_ref_wrapper<double>>(kw["value"])();
+
+        boost::optional<std::string> accountinglabel = boost::none;
+        if (kw.has_key("accountinglabel")) {
+            accountinglabel = boost::python::extract<std::string>(kw["accountinglabel"])();
+        }
+        return effect_wrapper(std::make_shared<Effect::SetMeter>(M,
```
geoffthemedio: probly move `M` to a new line, indented once

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,175 @@
...
+        auto effects_args = boost::python::extract<boost::python::list>(kw["effects"]);
+        if (effects_args.check()) {
+            boost::python::stl_input_iterator<effect_wrapper> effects_begin(effects_args), effects_end;
+            for (auto it = effects_begin; it != effects_end; ++it)
+                effects.push_back(ValueRef::CloneUnique(it->effect));
+        } else {
+            effects.push_back(ValueRef::CloneUnique(boost::python::extract<effect_wrapper>(kw["effects"])().effect));
+        }
+
+        std::string stackinggroup = "";
```
geoffthemedio: no ` = ""`

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,175 @@
...
+        if (kw.has_key("stackinggroup")) {
+            stackinggroup = boost::python::extract<std::string>(kw["stackinggroup"]);
+        }
+
+        std::unique_ptr<Condition::Condition> activation = nullptr;
```
geoffthemedio: don't need ` = nullptr` I think

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,175 @@
...
+                                    std::move(parameters),
+                                    std::move(empire),
+                                    affiliation,
+                                    std::move(label),
+                                    stringtable_lookup
+                              ));
+    }
+
+
+    template <MeterType M>
```
geoffthemedio: This seems like an unnecessary template since `M` is just passed on by value. A lambda instead of `insert_set_meter_<MeterType::METER_MAX_SHIELD>);` might arguably be better, but it is rather awkward to repeatedly write out
`boost::python::raw_function([](auto& args, auto& kw) { insert_set_meter_(MeterType::METER_MAX_SHIELD, args, kw)});` and similar, so I guess the template is OK...

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,175 @@
...
+    effect_wrapper insert_set_meter_(const boost::python::tuple& args, const boost::python::dict& kw) {
+        auto value = boost::python::extract<value_ref_wrapper<double>>(kw["value"])();
+
+        boost::optional<std::string> accountinglabel = boost::none;
```
geoffthemedio: ` = boost::none` probably unnecessary

parse/EffectPythonParser.cpp
o01eg: `auto message`  and `auto icon` are already reference and generates error with `std::move`:
```
/mnt/another/srcs/GIT/freeorion/parse/EffectPythonParser.cpp:67:78:   required from here
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/include/g++-v11/ext/new_allocator.h:162:11: error: cannot bind non-const lvalue reference of type «std::string&» {aka «std::__cxx11::basic_string<char>&»} to an rvalue of type «std::__cxx11::basic_string<char>»
```

parse/EffectPythonParser.cpp
o01eg: Fixed in parent PR.

parse/EffectPythonParser.cpp
o01eg: I suppose it will be possible to just iterate over `MeterType` enum but it will require to change identifiers so I postpone it after complete Python migration.

parse/EffectPythonParser.cpp
o01eg: Fixed in parent PR.

parse/EffectPythonParser.cpp
o01eg: Fixed in parent PR.

parse/EffectPythonParser.cpp
o01eg: Fixed in parent PR.

parse/EffectPythonParser.cpp
o01eg: Fixed in parent PR.

parse/EffectPythonParser.cpp
o01eg: Fixed in parent PR.

parse/EffectPythonParser.cpp
o01eg: Fixed in parent PR.

parse/EffectPythonParser.cpp
```diff
@@ -0,0 +1,172 @@
...
+                    parameters.emplace_back(*it, std::make_unique<ValueRef::StringCast<int>>(ValueRef::CloneUnique(p_arg_int().value_ref)));
+                } else {
+                    parameters.emplace_back(*it, std::make_unique<ValueRef::Constant<std::string>>(py::extract<std::string>(param_args[*it])));
+                }
+            }
+        }
+
+        std::unique_ptr<ValueRef::ValueRef<int>> empire;
+        if (kw.has_key("empire")) {
+            auto empire_args = py::extract<value_ref_wrapper<int>>(kw["empire"]);
```
geoffthemedio: there are probably cases there are unnecessary copies being made of whatever is returned from `py::extract`, which I think can return a const reference. not a high priority to fix, though.

o01eg:
Rebased it against `master`.

geoffthemedio:
seems OK

PR #3591, opened 2021-10-21
Make necessity require research focus

Grummel7:


default/scripting/policies/NECESSITY.focs.txt
```diff
@@ -12,7 +12,8 @@ Policy
             scope = And [
                 ProductionCenter
                 OwnedBy empire = Source.Owner
-                Happiness high = (NamedReal name = "PLC_NECESSITY_MAX_STABILITY" value = 6)
+                Focus type = "FOCUS_RESEARCH"
+                Happiness high = (NamedReal name = "PLC_NECESSITY_MAX_STABILITY" value = 5.99)
```
geoffthemedio: `Not Happiness low = 6` might be better, as the `5.99` might show up in the UI and be somewhat odd looking...

Grummel7: I noticed the problem myself, but your fix is better, will update...

PR #3592, opened 2021-10-21
Three minor fixes

Grummel7:
First is definitely an oversight.
Not sure about the second, but I cannot see a good reason why e.g. unstable exobots should be better reasearchers. In fact the way it is now, Exobots with stability 7 are better at research than e.g. Replicon with stability 7 since very bad triggers later then bad research.
Last is a grammatical fix of the Metropole description.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -11528,7 +11528,7 @@ PLC_METROPOLES
 Metropoles
 
 PLC_METROPOLES_DESC
-'''Boosts the output of up to [[value METROPOLES_PLANET_LIMIT]] planets. At most 1 / [[value METROPOLES_MIN_PLANETS]] of all populated planets with [[metertype METER_POPULATION]] of at least [[value METROPOLES_MINIMUM_POPULATION]] and [[metertype METER_HAPPINESS]] of at least [[value METROPOLES_MIN_STABILITY]] will be boosted.
+'''Boosts the output of up to [[value METROPOLES_PLANET_LIMIT]] planets, but not more than a fraction of 1 / [[value METROPOLES_MIN_PLANETS]] of all the empire's populated planets. To qualify for being boosted, planets must have [[metertype METER_POPULATION]] of at least [[value METROPOLES_MINIMUM_POPULATION]] and [[metertype METER_HAPPINESS]] of at least [[value METROPOLES_MIN_STABILITY]].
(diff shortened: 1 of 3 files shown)


default/stringtables/en.txt
```diff
@@ -11528,7 +11528,7 @@ PLC_METROPOLES
 Metropoles
 
 PLC_METROPOLES_DESC
-'''Boosts the output of up to [[value METROPOLES_PLANET_LIMIT]] planets. At most (1 / [[value METROPOLES_MIN_PLANETS]] or populated planets) with [[metertype METER_POPULATION]] of at least [[value METROPOLES_MINIMUM_POPULATION]] and [[metertype METER_HAPPINESS]] of at least [[value METROPOLES_MIN_STABILITY]] will be boosted.
+'''Boosts the output of up to [[value METROPOLES_PLANET_LIMIT]] planets. At most 1 / [[value METROPOLES_MIN_PLANETS]] of all populated planets with [[metertype METER_POPULATION]] of at least [[value METROPOLES_MINIMUM_POPULATION]] and [[metertype METER_HAPPINESS]] of at least [[value METROPOLES_MIN_STABILITY]] will be boosted.
```
geoffthemedio:
```suggestion
'''Boosts the output of up to [[value METROPOLES_PLANET_LIMIT]] planets. The number of planets boosted will be, at most, a fraction of 1 / [[value METROPOLES_MIN_PLANETS]] of all the empire's populated planets that have [[metertype METER_POPULATION]] of at least [[value METROPOLES_MINIMUM_POPULATION]] and [[metertype METER_HAPPINESS]] of at least [[value METROPOLES_MIN_STABILITY]].
```

Grummel7: Your version can be interpreted to say that only 1/4 of the qualified planets are boosted.
I think we better handle the maximum number in one sentence and the condition in another.
How about:

```suggestion
'''Boosts the output of up to [[value METROPOLES_PLANET_LIMIT]] planets, but not more than a fraction of 1 / [[value METROPOLES_MIN_PLANETS]] of all the empire's populated planets. To qualify for being boosted, planets must have [[metertype METER_POPULATION]] of at least [[value METROPOLES_MINIMUM_POPULATION]] and [[metertype METER_HAPPINESS]] of at least [[value METROPOLES_MIN_STABILITY]].
```

geoffthemedio: looks OK

PR #3593, opened 2021-10-21
Rework and rename Balance policy to Moderation

geoffthemedio:
Reworks Balance policy as Moderation. Formula for bonuses to capital influence and stability now depend on the ratio of the number of planets in the empire and the number using the most-commonly-used focus in the empire.

<strike>Tweak Balance policy focus-dependent formula so that the effect is additive rather than multiplicative, and having only a single focus used, and thus a spread of 0 between most and least common used focus, can't give a big bonus.</strike>


Grummel7:
The idea you described in the forum was a little different:
Divide the number of planet by max used focus.
||Planets With Focus|Planets with Focus|Planets with Focus|Planets with Focus|Planets with Focus|Planets with Focus|
|---|---|---|---|---|---|---|
|Focus A|1|8|2|8|5|5|
|Focus B |2 |  |2 |8 |5|4|
|Focus C |3 |  |2 |8 | |1|
|Focus D |1 |  |2 |8 | | |
|Focus E |1 |  |2 |8 | | |
|  |  |  |  |  | | |
|histo max |3 |8 |2 |8 |5|5|
|histo spread |2 |0 |0 |0 |0|4|
|unique |5 |1 |5 |5 |2|3|
|all|9|8|10|40|10|10|
|  |  |  |  |  | | |
|(unique-1)/(spread + 1) |1.33 |0.00 |**4.00** |**4.00** |**1.00**|0.40|
|unique /(spread+1) |1.67 |1.00 |5.00 |5.00 |2.00|0.60|
|(unique-1)/(max+1) |1.00 |0.00 |1.33 |0.44 |0.17|0.34|
|(all/max)  |2.67|1.00|5.00|5.00|2.00|2.00|
|(all/max)-1|1.67|0.00|4.00|4.00|1.00|1.00|

I wonder how you got values as high as 9, I've corrected some entries.

The last one would be my choice.
The problem with dividing by spread is that it the number may go down a lot when you add another focus on a single planet (compare last two columns)

geoffthemedio:
> I wonder how you got values as high as 9, I've corrected some entries.

I was testing scaling up the numbers a bit and didn't revert the chage before copy-pasting the results.

> The idea you described in the forum was a little different:
> Divide the number of planet by max used focus.

I misremembered and misread my own post.

> The problem with dividing by spread is that it the number may go down a lot when you add another focus on a single planet (compare last two columns)

That was intentional, to fit with the "theme". If you have 8, 8, 8, 7, and 1 planet on each of 5 focus settings, that's rather unbalanced compared with 8, 8, 8, and 8, which is perfectly balanced.

Perhaps the gameplay outweighs the conceptual here, though I'm hesitant to make all but the highest-used focus not matter at all.

Grummel7:
> That was intentional, to fit with the "theme". If you have 8, 8, 8, 7, and 1 planet on each of 5 focus settings, that's rather unbalanced compared with 8, 8, 8, and 8, which is perfectly balanced.

Well, 8, 8, 8, 8 is only looks perfectly balanced because you ignore all the zeroes (=unused focuses).

Also note that with 4x8 the bonus will also drop drastically when you extend your empire with a 33th planet, no matter what focus you chose for it. Thematically it may be nice to reward reaching such a nice-looking setup, but for a bigger empire it will require a lot of counting (there is no direct way to see the numbers in the game) and often its simply not possible.

geoffthemedio:
As far as I can tell, it should currently count the non-focused planets as having a distinct focus for the calculation of the number of unique focuses used, as the sampling condition doesn't exclude them.

For the case of 4x8 = 32 and adding one, you could instead use 3x11.

Alternatively, it could be based on the unique count / stdev + 1, which would much favours the "more balanced" 8,8,8,9 over "unbalanced" 8,8,8,8,1. Arguably breaks down a bit for all one focus, though that could be considered "balanced" in a sense...

Also, if I calculated the fcorrectly, (all/max) - 1 actually favours 8,8,8,8,1 over 8,8,8,9 since all is the same and max is lower for the former, which seems wrong, as the latter seem much more "balanced". As well, 9,8,8,8 would be scored the same as 9,9,9,2,2,2 which seems odd. There could be a policy around not using any single focus a lot more than all the others, perhaps called "Moderation", though I don't think that behaviour really fits for "Balance".

<html><body>
<google-sheets-html-origin>

  | Planets with Focus | Planets with Focus | Planets with Focus | Planets with Focus | Planets with Focus | Planets with Focus
-- | -- | -- | -- | -- | -- | --
Focus A | 8 | 8 | 8 | 6 | 32 | 33
Focus B | 8 | 8 | 8 | 7 | 1 |  
Focus C | 8 | 8 | 8 | 7 |   |  
Focus D | 8 | 9 | 8 | 7 |   |  
Focus E |   |   | 1 | 6 |   |  
  |   |   |   |   |   |  
max | 8 | 9 | 8 | 7 | 32 | 33
spread | 0 | 1 | 7 | 1 | 31 | 0
unique | 4 | 4 | 5 | 5 | 2 | 1
stdev | 0.00 | 0.50 | 3.13 | 0.55 | 21.92 | 0.00
rms | 8.00 | 8.26 | 7.17 | 6.62 | 22.64 | 33.00
mean | 8 | 8.25 | 6.6 | 6.6 | 16.5 | 33
all | 32 | 33 | 33 | 33 | 33 | 33
  |   |   |   |   |   |  
(unique-1)/(spread + 1) | 3.00 | 1.50 | 0.50 | 2.00 | 0.03 | 0.00
unique /(spread+1) | 4.00 | 2.00 | 0.63 | 2.50 | 0.06 | 1.00
(unique-1)/(max+1) | 0.33 | 0.30 | 0.44 | 0.50 | 0.03 | 0.00
(all/max) - 1 | 3.00 | 2.67 | 3.13 | 3.71 | 0.03 | 0.00
(unique)/(stdev + 1) | 4.00 | 2.67 | 1.21 | 3.23 | 0.09 | 1.00
4*exp(mean-rms) | 4.00 | 3.95 | 2.26 | 3.93 | 0.01 | 4.00


</body>
</html>

Grummel7:
> As far as I can tell, it should currently count the non-focused planets as having a distinct focus for the calculation of the number of unique focuses used, as the sampling condition doesn't exclude them.

With zeroes I meant all the potential focuses that no planet using.
Outposts on the other hand are another problem, since you can only change the focus by building a colony on them.

> For the case of 4x8 = 32 and adding one, you could instead use 3x11.

Well, the next number may be prime. More serious, you cannot simply re-adjust a third of your planets only to maximise the output of Balance. In other cases you always try to avoid micro management, but what is that?

> Alternatively, it could be based on the unique count / stdev + 1, which would much favours the "more balanced" 8,8,8,9 over "unbalanced" 8,8,8,8,1. Arguably breaks down a bit for all one focus, though that could be considered "balanced" in a sense...

I thought the original idea behind balance was to provide an incentive for using different focuses. Of course this could again be achieved by using (unique-1)/(stdev+1) instead. The stdev version is definitely better than the spread one.

agrrr3:
The moves one can do for outposts:
- add one on an unowned planet, takes preparation or time, some cost, will increase total number of planets (e.g. for balance counting and influence upkeep), spot will usually be easy to find (maybe not so easy in late game)
- add one by reducing a colony (prohibitive costly), keeps number of planets steady
- add one by invasion (usually costly, difficult), will increase number of owned planets
- remove one by abandoning it (takes some time, has a delay), will decrease number of owned planets
- remove one by colonizing it,  will keep number of owned planets

This is one or two orders of magnitude more complicated than changing focus. A very different beast.

The main problem I see though is that UI fails: you don't get correct estimates if you order to outpost (I think), and for sure you dont get the right estimate if you order to abandon it; same for colonisation
I guess that would be ok if the change in numbers is not too high.

And if we keep influence as a max meter we should at least add a 1-turn delay to focus adoption - as you have no control about the default focus you should be able to tweak it before the policy is affected. i would prefer a target meter though.

One idea for more robust bonus: I think we could reinterpret "balance" to "natural order" and look for deviation from a gauss distribution. With the best fit giving the highest bonus multiplier, multiplied by the highest count of focus. So the best could be 0 1 2 4 4 2 1 0 (e.g. 2 * 4), and the worst 0 0 0 14 0 0 0 0 (e.g. -2 * 14). My hunch is switching a single value wont change too much even if it is "extreme". Not completely sure how to put that into a FOCS formula though.

So I guess I suggest, lets make influence a target meter, dont count newly colonised/outposted planets for the first turn, and choose a "robust" bonus.

side note: Also a bit sad this design discussion/reasoning is not on the forums - we will loose it.

geoffthemedio:
> ...look for deviation from a gauss distribution. With the best fit giving the highest bonus multiplier, multiplied by the highest count of focus. So the best could be 0 1 2 4 4 2 1 0 (e.g. 2 * 4), and the worst 0 0 0 14 0 0 0 0 (e.g. -2 * 14).

Both of those are somehat like a gaussian distribution, but with different variance. Rather non-gaussian would be things like uniform 2 2 2 2 2 2 2 2 or bimodal 6 1 1 0 0 1 1 6 or skewed/asymmetric 0 1 2 3 4 6 0 0. But bimodality or skew or "normality" (gaussianness) calculations don't really make sense for this application because the histogram values aren't ordered, so there's no difference between 6 1 1 0 0 1 1 6 and  0 1 1 6 6 1 1 0, and there are no possible values outside the set of discrete focuses to check for.

> My hunch is switching a single value wont change too much even if it is "extreme". Not completely sure how to put that into a FOCS formula though.

Depends what exactly you mean...

agrrr3:
> > ...look for deviation from a gauss distribution. With the best fit giving the highest bonus multiplier, multiplied by the highest count of focus. So the best could be 0 1 2 4 4 2 1 0 (e.g. 2 * 4), and the worst 0 0 0 14 0 0 0 0 (e.g. -2 * 14).
> 
> Both of those are somehat like a gaussian distribution, but with different variance. 

right. in the broken(?) example for worst i intended that the normal distribution should have a fixed spread/variance (on the number of available focus types?). Although that introduces another can of worms - is there a prefixed value for that or does that spread increase when you unlock a new focus type. (If we cant specify the expected variance we could skew the data (e.g. scale by 100 and add one) I guess, so 1 201 301 401 401 301 201 1 and 101 101 101 1501 101 101 101 101).

> Rather non-gaussian would be things like uniform 2 2 2 2 2 2 2 2 or bimodal 6 1 1 0 0 1 1 6 or skewed/asymmetric 0 1 2 3 4 6 0 0
> 

the focus types should not matter, so it would need a sorting step beforehand. so bimodal, skewed, and asymmetric would not exist (0 1 1 6 6 1 1 0 and 0 1 3 6 4 2 0 ). besides uniform, a split would rather non-gaussian i guess 0 0 4 4 4 4 0 0 )

> > My hunch is switching a single value wont change too much even if it is "extreme". Not completely sure how to put that into a FOCS formula though.
> 
> Depends what exactly you mean...

My hunch could be better described as: humans have a good natural grasp of gaussian distribution so they can spot deviation and estimate changes easily (adding the ability to show a histogram of planet focus via object view would make that even more simple).

For the formula after a bit of google-foo ( https://towardsdatascience.com/6-ways-to-test-for-a-normal-distribution-which-one-to-use-9dcf47d8fa93 ): well, one could use something like a Shapiro–Wilk test to get the probability that the given measurements could be produced by a perfect normal distribution (which does not exist in reality, so the test is not so useful for real-life data as soon as you have a lot of data points). The player is in control, so it is ok to strive for that a perfect normal distribution.
It seems its hinted that for Kolmogorov Smirnov test you can specify a variance. Or maybe rather not get the probability of the test vs normal distribution but directly the Kolmogorov Smirnov D statistic (which AFAIU is maximum absolute deviation from the cumulated normal curve).

So the policy would be rather a "Perfect Normality", striving for something rather unnatural.

geoffthemedio:
> is there a prefixed value for that or does that spread increase when you unlock a new focus type.

There's no way at present to get a number for how many focus types an empire has available, and that number would vary from planet to planet anyway, so is somewhat meaningless anyway...

> ... Shapiro–Wilk test ... Kolmogorov Smirnov test ... Kolmogorov Smirnov D statistic (which AFAIU is maximum absolute deviation from the cumulated normal curve).

I think if it needs to run these sorts of statistical tests to calculate a bonus to a meter, it's gotten a bit too far from player understandability.

At this point, I'm more inclined to go with @Grummel7's suggestion of something based on (count of planets / number using most popular focus) and then to rename the policy to Moderation.

agrrr3:
> > ... Shapiro–Wilk test ... Kolmogorov Smirnov test ... Kolmogorov Smirnov D statistic (which AFAIU is maximum absolute deviation from the cumulated normal curve).
> 
> I think if it needs to run these sorts of statistical tests to calculate a bonus to a meter, it's gotten a bit too far from player understandability.

i think it breaks expectations less than what we have now. (like: ah shit, i have to abandon this newly built outpost immediately)

> 
> At this point, I'm more inclined to go with @Grummel7's suggestion of something based on (count of planets / number using most popular focus) and then to rename the policy to Moderation.

is also fine, pretty robust

My main point was: lets make influence a target meter, dont count newly colonised/outposted planets for the first turn, and choose a "robust" bonus.

geoffthemedio:
> something based on (count of planets / number using most popular focus) and then to rename the policy to Moderation.

done...

@Grummel7 further comments?

Grummel7:
> @Grummel7 further comments?

I don't mind keeping the name and the formula for was my suggestion anyway.

The scaling is currently too low, I think. Bonus for Influence is now 2 * (all / max - 1), i.e. all/max=4, which is a rather high value in practice, give just 6 points. I suggest using a factor of 4 instead. Stability to capital is less important, but could also be doubled, I think.

And since the formula is now rather simple, why not put it into the help text, instead of just telling which factors are used?

agrrr3:
DISCLAIMER: Ah I forgot about outposts... Also I sometimes took the non-doubled influence output.. need to edit this... WIP 

...
ok, after recalculation I think doubling influence output would be too high

agrrr3:
With the new formula, regarding balancing there are six types of focus (outpost, production, research, influence, growth, defense) in the beginning and you can go easily for logistics and stockpile extraction (so 8). There are 14 focus types in the game.

So best reachable ratio is probably eight planets with eight foci (8:1), so a +14 being the maximum in the game with perfect balance (same result for 16:2, 24:3...).
This breaks easily, getting 15 planets with maximum 3 same foci out of 8 focus seems possible (15:3), that would give 8 influence.

If one sets a planet to "useless" foci instead of influence that will probably reduce your net influence output.(?)
So in a peaceful situation if you have rather 3 fully useable foci (influence, production, research) a half-usable (outposts) and some planets which you could set to supply and rarely to growth. 
So in a 8 planet situation, put 2 to influence,production,research and 1 to outpost and 1 to supply (or 2 outposts). That nets 6 influence from the policy. Setting a planet more to influence would net only 3.3 from the policy, depending on population and trait that could be offset by the influence boost from the extra focus (and artisan workshop as a policy would alone give 4 extra influence there).
If you have 9 planets and put them to 3 production, 3 research, 3 influence you get 4 influence from the policy. Adding "cheap" outposts would increase the influence output by 2 (and increase the influence upkeep by ~1.5).

So I think doubling the influence output would be probably too high, +50% (so multiplier 3) could be ok. I am not sure on interaction with other multipliers (e.g. if you get another 25% on top if you have good influence species on the capital).

agrrr3:
@Grummel7 can you have a look at my balancing comment and comment if you concur

geoffthemedio:
changed `"PLC_MODERATION_INFLUENCE_SCALING"` to `value = 3.0`

Grummel7:
> @Grummel7 can you have a look at my balancing comment and comment if you concur

Sorry, didn't see your comment in time. 3 may be okay.
There is no such thing as a perfect value, but my feeling was that 2 was too little for a focus that denies industrialism. Yes, you can get very high ratios, but in practice this will usually cost you more than the influence bonus is worth.

PR #3510, opened 2021-06-30
Enable Godot for snap package

o01eg:

PR #3601, opened 2021-11-06
Enable Godot for snap package (FIXED)

agrrr3:
fixed snap build based on #3510 so one can start the godot client

edit: last cleanup verified.

== Fixes ==
* icon path without $SNAP - Icon paths in the desktop file will be rewritten to use ${SNAP}/<file> if specified as desktop file in snapcraft.yaml
* use a library path in gdnlib which is known at build time and accessible at run time
** it seems this gets hardcoded when exporting via godot-server
** layout adds a symlink for runtime

== Refactor ==
* use ${SNAPCRAFT_ARCH_TRIPLET}
* format fixes for yamllint

### DIFF
--- a/.github/snap-xvfb-launch.sh
+++ b/.github/snap-xvfb-launch.sh
@@ -8,7 +8,7 @@ echo = Launching freeorion =
 
 export SDL_VIDEODRIVER=x11
 
-freeorion &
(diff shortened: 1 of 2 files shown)


o01eg:
I suppose it is ready to be merged? @agrrr3

agrrr3:
> I suppose it is ready to be merged? @agrrr3

yes

PR #3602, opened 2021-11-12
Some improvements to BUREAUCRACY

Grummel7:
Also do not penalise focus change right after conquering
Avoid duplicate value definition
Fixed numbers in message

### DIFF
--- a/default/scripting/policies/BUREAUCRACY.focs.txt
+++ b/default/scripting/policies/BUREAUCRACY.focs.txt
@@ -16,7 +16,8 @@ Policy
                 OwnedBy empire = Source.Owner
                 Species
                 (LocalCandidate.TurnsSinceFocusChange == 0)
-                (LocalCandidate.TurnsSinceColonization > 1) // all changes on first turn or two after colonization without penalty
+                (LocalCandidate.TurnsSinceColonization > 1) // all changes on first turn after colonization without penalty
+                Not (LocalCandidate.TurnsSinceLastConquered == 1) // same for newly conquered, note that 0 means never conquered
             ]
             priority = [[FOCUS_CHANGE_PENALTY_PRIORITY]]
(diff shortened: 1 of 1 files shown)


default/scripting/policies/BUREAUCRACY.focs.txt
```diff
@@ -66,7 +67,7 @@ Policy
                         ]
                         OwnedBy empire = Source.Owner
                     ]
-                    tag = "needbuildings" data = (TurnsSincePolicyAdopted empire = Source.Owner name = ThisPolicy) /
+                    tag = "needbuildings" data = (TurnsSincePolicyAdopted empire = Source.Owner name = ThisPolicy + 1) /
```
Grummel7, line 70: There is still a problem with this, but I have no idea how to fix it:
The message is generated the turn before, but you may finish another RegAd in the same turn and never actually lose the stability bonus.

geoffthemedio: I suppose it doesn't need to only appear on turns when it expects the newly increased requirement to not be met...  It could instead appear on turns when the requirement will soon increase by 1, regardless of whether the empire has enough admin buildings already.

Grummel7: You means something like:
  `activation = ((turns + 2) % 10) == 0`
and message:
  `Next turn you'll need N, currently having M.`
I guess that could work, just have to add operator %.

geoffthemedio: https://github.com/freeorion/freeorion/commit/ca9c4cb5f3230e8ee901a2879348ac5019b5aa05

geoffthemedio: fixing this can be a separate pull request, though...

PR #3605, opened 2021-11-18
AI entries for FLUX techs, hulls, and lance

agrrr3:
not much used by AI though. it will research spatial flux hull and so will get access to the lance. 

if it does not go organic it probably should research the flux bubble hull, tried to adjust hull research in ResearchAI , not sure it will work


agrrr3:
the lint fails seem dubious.
i tested the implementation. found one bug: shipdesigners try to add multiple flux lances part

agrrr3:
So if only the string lint fails, it is not related to this PR (no stringtable changes)

Also that looks "old":
./check/st-tool.py check --reference default/stringtables/en.txt default/stringtables/??.txt
default/stringtables/en.txt:6718: Referenced key 'browsepath PATH_CONFIG' in value of 'CONFIG_GUIDE_TEXT' was not found.
Error: Process completed with exit code 1.

geoffthemedio:
> So if only the string lint fails, it is not related to this PR (no stringtable changes)

hopefully fixed... https://github.com/freeorion/freeorion/commit/0da0365d976d4abe6a948ae77b4ce290c4e6acab

rebase this?

agrrr3:
everything green after rebase

Issue #3527, opened 2021-07-17
Abandoned Outposts still considered a combat target


MichaelCourtney:
Bug Report
==========

MichaelCourtney:
I failed at using the issue interface again.

If you build an outpost then abandon it, the planet is still a target if combat occurs in the system.

hifron:
Maybe outpost should be changed to become some artefacts...

agrrr3:
which version? can you provide a save game/screenshot?

geoffthemedio:
Does this apply just on the turn the outpost in abandoned, or for several turns or forever after?

**Abandon Outpost** is a building, and its effect don't fire until after combat during a turn cycle.

MichaelCourtney:
16th Multiplayer game 2021-05-10.4f04bdc It occurred well after the outpost was abandoned. I'd probably have to recreate it to get a screenshot. I'll look through the old thread after work, maybe I left a turn number there.

MichaelCourtney:
Re: Sixteen game on the multiplayer slow game server
Looking at the combat log for turn 119 Demi alpha. LienRag attacked Demi alpha 1 doing 1 pt of damage and incapacitating it. Thing is there is no outpost or colony on Demi alpha 1, but there was earlier in the game. So there is still a problem with abandon outpost.

o01eg:
Saves placed here to reproduce: https://freeorion-test.dedyn.io/FO0016-saves.7z

geoffthemedio:
possibly related to abandoned outposts having lingering nonzero defense, shield, or infrastructure meters from before they were abandoned?

MichaelCourtney:
> possibly related to abandoned outposts having lingering nonzero defense, shield, or infrastructure meters from before they were abandoned?

Probably. I've tried to reproduce in current master and the planet was not attacked.

PR #3607, opened 2021-11-20
Treat neutral (==ALL_EMPIRES) forces like an empire for visibility

agrrr3:
Without this fix, stealth does not work against neutral ships (aka monsters):
2021-11-16 6c3d9f1_combat_report

With the fix, the hidden ship Guilt gets attacked correctly starting from bout 4

The implementation only changes visibility resolution in battle

Detailed Design:

- the underlying model is: neutral forces per default can see everything
- this gets modified/interpreted differently in many places; usually: local monster detection has to be higher than the object's stealth

Special situations handling of detection:

- initiating combat (ServerApp::GetFleetsVisibleToEmpireAtSystem)
- targeting during combat via FOCS; ship weapons use VisibleToEmpire condition in default/scripting/ship_parts/targeting.macros (this is necessary as the visibility changes while fighting)
- Bombard parts use Stealth high = Source.Detection

Changed in this PR: VisibleToEmpire Condition checks if context.combat_bout > 0 to see if we are in a battle. If not, we return true as before.

No change: Universe::EmpireObjectVisibility does not have a context, so I am unable to decide anything, returns always full visibility.

Changed in this PR: The unused EmpireObjectVisibility complex ValueRef now is consistent with the VisibleToEmpire Condition. I.e. it checks the visibility map if inside combat, else it assumes full vision for unowned objects.

### DIFF
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -7842,8 +7842,6 @@ namespace {
         bool operator()(const std::shared_ptr<const UniverseObject>& candidate) const {
             if (!candidate)
                 return false;
-            if (m_empire_id == ALL_EMPIRES)
-                return true;
             if (m_vis == Visibility::VIS_NO_VISIBILITY)
                 return true;
 
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -7844,6 +7844,8 @@ namespace {
                 return false;
             if (m_vis == Visibility::VIS_NO_VISIBILITY)
                 return true;
+            if (m_empire_id == ALL_EMPIRES && m_context.combat_bout < 1)
+                return true; // outside of battle neutral forces have full visibility per default
 
             if (m_since_turn == INVALID_GAME_TURN) {
                 // no valid game turn was specified, so use current universe state
--- a/universe/ScriptingContext.h
+++ b/universe/ScriptingContext.h
@@ -313,7 +313,7 @@ struct ScriptingContext {
     const CurrentValueVariant&            current_value = EMPTY_CURRENT_VALUE;
 
     // general gamestate info
-    int                                            combat_bout = 0;
+    int                                            combat_bout = 0; // first round of battle is combat_bout == 1
     int                                            current_turn = CurrentTurn();
     int                                            in_design_id = INVALID_DESIGN_ID;
     int                                            production_block_size = 1;
--- a/universe/ValueRefs.cpp
+++ b/universe/ValueRefs.cpp
@@ -1475,8 +1475,8 @@ Visibility ComplexVariable<Visibility>::Eval(const ScriptingContext& context) co
         int empire_id = ALL_EMPIRES;
         if (m_int_ref1) {
             empire_id = m_int_ref1->Eval(context);
-            if (empire_id == ALL_EMPIRES)
-                return Visibility::VIS_NO_VISIBILITY;
+            if (empire_id == ALL_EMPIRES && context.combat_bout < 1)
+                return Visibility::VIS_FULL_VISIBILITY; // outside of battle neutral forces have full visibility per default
         }
 
         int object_id = INVALID_OBJECT_ID;


agrrr3:
using git blame i can see  no reason why the check for ALL_EMPIRES was introduced (which returned false first and prevented neutrals to attack at all, then got (partly) corrected to true which means always attack). So i guess when that was introduced ALL_EMPIRES was interpreted as an error case rather than a neutral candidate; this should get merged ASAP

geoffthemedio:
> then got (partly) corrected to true which means always attack

You're seemingly only considering this in the context of target selection in combat, but this is a general use condition that could affect how a variety of scripted content works in various other contexts.

> this should get merged ASAP

This needs to be carefully considered with regard to where else it might impact. If this check is removed, then the next / fallback case will probably be https://github.com/freeorion/freeorion/pull/3607/files#diff-4432a96f2c01579f3e2784641296a0bff3d3c556a0ff1fdb096620692deee6c3R7857-R7858 which will return false in cases where it would currently return true, if the specified empire is -1 (`ALL_EMPIRES`), which could happen for various cases like uses of `Source.Onwer` when the source object is unowned, which can be the case for buildings or species or ship parts or hulls.

For example, does monster planet bombardment still function?

Are the results consistent with `Universe::GetObjectVisibilityByEmpire` ?

agrrr3:
> > then got (partly) corrected to true which means always attack
> 
> You're seemingly only considering this in the context of target selection in combat, but this is a general use condition that could affect how a variety of scripted content works in various other contexts.

AFAIR we switched it from always false to always true only because of combat and we did not notice a difference in behavior beyond combat(?). 09a00663ae2edd14804ed41f6219ee04399073a9

> 
> > this should get merged ASAP
> 
> This needs to be carefully considered with regard to where else it might impact.

i just thought if it is in master it will be playtested this week and we will probably find major troubles if it causes any (and fix it or roll back). 
i think you are the only one who is able to do the necessary consideration - knows the visibility map code. you know better than me.

> If this check is removed, then the next / fallback case will probably be https://github.com/freeorion/freeorion/pull/3607/files#diff-4432a96f2c01579f3e2784641296a0bff3d3c556a0ff1fdb096620692deee6c3R7857-R7858 ...

i only see the check i removed there. probably the link is not working(?)

> For example, does monster planet bombardment still function?

dunno. 
are monsters currently able to bombard hidden planets they should not be able to see?

> 
> Are the results consistent with `Universe::GetObjectVisibilityByEmpire` ?

no: for ALL_EMPIRES Universe::GetObjectVisibilityByEmpire returns full visibility for any object. 

edit1: Universe::EmpireKnownObjects returns all existing objects, no matter if anybody sees those

Universe::EmpireVisibleObjectIDs seems to treat ALL_EMPIRES as any-empire. It returns the union of at least basic visibility of all the empires in the set given as parameter. And the function seems to be dead code/unused.

if we interpret ALL_EMPIRES everywhere slightly differently, we should probably introduce another value to distinguish those cases(?)

what is the theoretical model for neutral forces visibility? i guess it would be ok to treat as an empire without common detection level. i.e. sharing the results of detection. Every turn adding to the same visibility map.

Or: no monster and no neutral planet shares visibility information; but we do not want to track object-to-object visibility because of combinatorial complexity, so the game does not track that at all. only when battle starts they work as a unit. Is this the status quo?


second edit;batch of extra stuff:

Current FOCS provides: 

- EmpireObjectVisibility complex ValueRef takes a single empire (and single object).
- Visibility simple ValueRef. 
- EnumValueRefs for the visibility levels. 
- And finally the VisibleToEmpire Condition, which takes a single empire ID ValueRef, optional turn ValueRef, and optional visibility ValueRef. For a normal empire it looks at a single empire's visibility map. 

You can apply set operations on the Condition and the ValueRefs as well (if you wrap them in a condition!?), so we have to think about Set Theory here.

from generic conditons/ Set Theory point it is OK to treat ALL_EMPIRES like an empire. Because objects Not OwnedBy id = X for a normal empire contains all objects from the all the other empires (including the neutral empire). Same for Not OwnedBy id = ALL_EMPIRES, it contains all objects from all the other empires.

The main Set operation for visibility is shared vision for alliances (a best-visibility-union) and that is only possible between normal empires. The communal vision also includes neutral objects, but that is also ok because that is no set operation - it adds visibility for every relevant object to every relevant empire. I am also not sure if shared alliance vision is a set operation or works by manipulating all empires (probably the latter). 

Applying best-of-vision when querying for any level of visibility means: the union of all objects having at least that level of visibility by any of the given empires. The complement to that is the intersect of all objects with lower visibility (i.e. all the objects which cant be seen by any of the given empires)

So I am not sure what way to go here - I think visibility needs to be cleaned up a bit, have a better foundation. 

Some ideas

1. we say neutral forces have no memory/visibility tracking but in combat. Outside of combat we always assume perfect vision (i.e. effect evaluation) for special cases (initiating combat, bombardment) we can code and script some specific stuff using object detection meter. We can introduce something like a CombatVisibleToEmpireCondition, CombatVisibilityValueRef and EmpireObjectVisibilityValueRef OR check inside those if we are in a combat context or not.
2. we say usually neutral forces have no memory, but for specific cases they can, so before returning always true/perfect vision, we check first if there is an entry in the visibility map or not and use the entry if it exists.
3. we say neutral works like a normal single empire and we always use its visibility map (so in the beginning neutral tracks almost all objects in the universe)

currently i am inclined to try 1) and detect if we are in combat or not

edit3:
I think I remember we had a mismatch between initiating combat and actual combat with monsters once - combat was started, but the monsters could not shoot all (because of hidden ships?). So I think I found where that initiating combat happens (linearized pseudo/code):
```
 ServerApp::AssembleSystemCombatInfo
   for (system : universe.Objects().all<System>())
     return collect(combat_info if CombatConditionsInSystem(system))

 ServerApp::CombatConditionsInSystem
      ...
      for (aggressive_empire: empires_with_aggressive_fleets_here):
        // in ServerApp::GetFleetsVisibleToEmpireAtSystem(.., aggressive_empire, ..)
        ... // handling fleets visibile to empire ships and planets 
        // now considering only fleets visible to monsters       
        monster_detection_strength_here = ... // get best detection here.  Use monster detection meters for this...
        // test each ship in each fleet for visibility by best monster detection here 
        for (ship in fleet in objects.find<Fleet>(fleet_ids)):
           fleets_visible_to_empire_at_system.emplace(fleet.id) if (monster_detection_strength_here >= ship.initial_stealth_meter)
        for (fleets_visible_to_empire_at_system):
          if (at_war_with_empire):
            return true;
```
It only uses monster ships (so neutral colonies are ignored). I think this should not cause issues with 1)

edit4: a side observation; if we allow weapons to always do damage even if no enemies are detected checking for visibility won't suffice. but that is basically what mines are doing, and i think the effect approach can handle that sufficiently

edit5: EmpireObjectVisibility complex valueref is inconsistent (always returns NO visibility) and also dead (not used in contents)

edit6: Bombard parts use Stealth high = Source.Detection

edit7: the Visibility ValueRef is also dead code: As of this writing, there are no properties of objects that directly return a Visibility, as it will normally need to be queried for a particular empire

agrrr3:
State of PR: 
* the underlying model is: neutral forces per default can see everything
* this gets modified/interpreted differently in many places; usually: local monster detection has to be higher than the object's stealth

Special handling:
* initiating combat (ServerApp::GetFleetsVisibleToEmpireAtSystem)
* targeting during combat via FOCS; ship weapons use VisibleToEmpire condition in default/scripting/ship_parts/targeting.macros (this is necessary as the visibility changes while fighting)
* Bombard parts use Stealth high = Source.Detection

VisibleToEmpire Condition checks if context.combat_bout > 0 to see if we are in a battle. If not, we return true as before.

Universe::EmpireObjectVisibility does not have a context, so I am unable to decide anything. No change, returns always full visibility.

The unused EmpireObjectVisibility complex ValueRef now is consistent with the VisibleToEmpire Condition. I.e. it checks the visibility map if inside combat, else it assumes full vision for unowned objects.

@geoffthemedio good to be merged? 

another question is if we should remove the dead code (Visibility ValueRef, Universe::EmpireVisibleObjectIDs, and maaybe EmpireObjectVisibility complex ValueRef)?

hifron:
Shouldn't have neutral forces have diplomacy, ... and in future maybe some scenarios and missions?

agrrr3:
Please discuss future ideas in the forum's

hifron ***@***.***> schrieb am Fr., 24. Dez. 2021, 09:42:

hifron:
It is somehow on [Roadmap](https://freeorion.org/index.php/Roadmap), originally on old roadmap it was more clear and this depends on this project development and decisions.

agrrr3:
@hifron this is a bugfix PR. feature discussion (even if on roadmap) should not be here.

agrrr3:
@geoffthemedio can we merge this or do i need to address something?

hifron:
Is [issue #3527](https://github.com/freeorion/freeorion/issues/3527) related somehow with this?

geoffthemedio:
> @geoffthemedio can we merge this or do i need to address something?

I don't like how it's implemented and I'm looking into something more complicated / involved but general...

agrrr3:
side-note: it took me a while to get used to the current detection model. It is ugly in a sense but I can see benefits (not having to manage so much state). the fix is consistent with the detection model and not a hack, although it looks like one.

> > @geoffthemedio can we merge this or do i need to address something?
> 
> I don't like how it's implemented and I'm looking into something more complicated / involved but general...

fair enough but can we merge the fix anyway and throw it all out when you come up with the more general solution?

PR #3608, opened 2021-11-25
New policy icons and slots

TheSilentOne1:
WIP on the new policy icons.


UI/GovernmentWnd.cpp
```diff
@@ -243,28 +245,35 @@ void PolicyControl::CompleteConstruction() {
     GG::Control::CompleteConstruction();
     if (!m_policy)
         return;
+    int empire_id = GGHumanClientApp::GetApp()->EmpireID();
+    auto name = UserString(m_policy->Name());
```
geoffthemedio:
```suggestion
    const auto& name = UserString(m_policy->Name());
```

geoffthemedio:
Icons look pretty good. The layout probably needs some adjusting, as I don't have enough vertical screen to see all three rows of policy slots...

new policy icons

TheSilentOne1:
Good point, I'll look into it.

TheSilentOne1:
Less distance between policy slots and no new row after category looks like this:

poliies

geoffthemedio:
Looks pretty good overall. Black Markets looks more like a spying-related policy.

TheSilentOne1:
Agreed, several other icons will also get another pass, then some clean-up and it should be ready.
How do you feel about the adjusted policy layout?

geoffthemedio:
It looks decent, but it requires a very large government window to see all the available slots later in games when the player gets more of them, which will be a problem on not huge screens, particularly any laptops. A way to shrink the cards and slots is probably needed.
government_wnd_big_cards_layout

TheSilentOne1:
Smaller policy card size is done, some icon changes and this should be ready
tinies
.

geoffthemedio:
The small versions of the cards are too small for me, and the automatic layout deciding what size to use seems to do wonky things, changin between small and big and small again as I rezie the window.

I'd rather have an option to control the policy card scale, like: https://github.com/freeorion/freeorion/commit/8bfa45ea1099c2ae30aa9b7aec90b7a8df957bbd 

Still needs some bugfixing as sometimes the rightmost slot extends outside the window.

I'd also make the policly palette's cards adjust in size the same as the slot controls below.

There could also be a slider or set of radio buttons to adjust the card size in the display directly, rather than having to add something to the options window.

TheSilentOne1:
Almost there, but I'm stuck because I don't know how to access all the items of a PoliciesListBox. I should probably iterate through rows and them items, but I don't know how to do that. Help? :)

geoffthemedio:
`PoliciesListBox` is a `CUIListBox` which is a `GG::ListBox` which has `begin` and `end` iterators over its `Row`s, which have `at` and `size`. What specifically can't you access?

TheSilentOne1:
I just couldn't find the right syntax for `for`, so now I used `while`, but would be thankful how to do it with `for`.

```
auto it = begin();
    while (it != end()) {
        auto row = *it;
        int itt = 0;
        while (itt < row->size()) {
            auto policy_control = dynamic_cast<PolicyControl*>(row->at(itt));
            if (policy_control)
                policy_control->Resize(sz, pts);
            ++itt;
        }
        ++it;
    }
```

TheSilentOne1:
The main panel (bottom one) works as intended, but I'm kinda lost with the policy palette. I don't have much of an idea how to apply the zoom factor there; the current solution does work somehow but seems like a hack.

geoffthemedio:
Something like
````
    for (auto& row : *m_policies_list) {
        for (auto row_idx = 0; row_idx < row->size(); ++row_idx) {
            if (auto policy_control = dynamic_cast<PolicyControl*>(row->at(row_idx)))
                policy_control->Resize(sz, pts);
        }
    }
````

File: UI/GovernmentWnd.h [hunk 2]
```diff
@@ -26,9 +27,11 @@ class GovernmentWnd : public CUIWnd {
 
     void CloseClicked() override;
     void DoLayout();
+    void PolicySizeButtonClicked(std::size_t idx);
 
-    std::shared_ptr<PolicyPalette>  m_policy_palette = nullptr;
-    std::shared_ptr<MainPanel>      m_main_panel = nullptr;
+    std::shared_ptr<PolicyPalette>           m_policy_palette = nullptr;
```
geoffthemedio: remove the ` = nullptr`. Is redundant.

UI/GovernmentWnd.cpp
```diff
@@ -1519,9 +1560,20 @@ void GovernmentWnd::CompleteConstruction() {
     m_policy_palette->ClearPolicySignal.connect(
         boost::bind(&GovernmentWnd::MainPanel::ClearPolicy, m_main_panel, _1));
 
+    m_policy_size_buttons = GG::Wnd::Create<GG::RadioButtonGroup>(GG::Orientation::HORIZONTAL);
+    m_policy_size_buttons->AddButton(
+        GG::Wnd::Create<CUIStateButton>("100%", GG::FORMAT_LEFT, std::make_shared<CUIRadioRepresenter>()));
```
geoffthemedio: these % are somewhat meaningless and arbitrary... just some icon for big, medium, and small would be more appropriate

UI/GovernmentWnd.cpp
```diff
@@ -416,14 +412,21 @@ const std::set<std::string>& PoliciesListBox::GetCategoriesShown() const
 void PoliciesListBox::SizeMove(const GG::Pt& ul, const GG::Pt& lr) {
     GG::Pt old_size = GG::Wnd::Size();
 
+    double zoom_factor = 1;
```
geoffthemedio: could probably make a separate function for getting the policy width and height as a pair, rather than repeating this and similar code in a few places

TheSilentOne1: I did this, it doesn't save much space but looks cleaner.

UI/GovernmentWnd.cpp
```diff
@@ -1468,15 +1596,51 @@ void GovernmentWnd::Refresh() {
...
     const GG::Pt main_lr(ClientWidth(), ClientHeight() - GG::Y(INNER_BORDER_ANGLE_OFFSET));
 
     m_main_panel->SizeMove(main_ul, main_lr);
     m_policy_palette->SizeMove(palette_ul, palette_lr);
+
+    const GG::Pt size_buttons_ul = main_ul + GG::Pt(GG::X(PAD), - BUTTON_HEIGHT/2);
+    const GG::Pt size_buttons_lr = main_ul + GG::Pt((GG::X(PAD) + POLICY_SIZE_BUTTON_WIDTH) * num_size_buttons, BUTTON_HEIGHT/2);
+
+    m_policy_size_buttons->SizeMove(size_buttons_ul, size_buttons_lr);
+    // m_policy_size_buttons->ExpandButtons(true);       // leads to crash
```
TheSilentOne1: Inexplicably, this leads to a crash - any idea why @geoffthemedio?

geoffthemedio: Not sure. The error message for me in Visual Studio says heap corruption.

Stack looks like:
````
 	GiGi.dll!GG::Wnd::DetachChild(GG::Wnd * wnd) Line 622	C++
>	GiGi.dll!GG::Layout::Remove(GG::Wnd * wnd) Line 542	C++
 	GiGi.dll!GG::Layout::DetachChildCore(GG::Wnd * wnd) Line 734	C++
 	GiGi.dll!GG::Wnd::DetachChild(GG::Wnd * wnd) Line 622	C++
 	GiGi.dll!GG::Layout::ResizeLayout(unsigned int rows, unsigned int columns) Line 582	C++
 	GiGi.dll!GG::RadioButtonGroup::RemoveButton(GG::StateButton * button) Line 737	C++
 	GiGi.dll!GG::RadioButtonGroup::ExpandButtons(bool expand) Line 749	C++
 	FreeOrion.exe!GovernmentWnd::DoLayout() Line 1639	C++
 	FreeOrion.exe!GovernmentWnd::SizeMove(const GG::Pt & ul, const GG::Pt & lr) Line 1585	C++
 	[Inline Frame] FreeOrion.exe!CUIWnd::ValidatePosition() Line 237	C++
 	FreeOrion.exe!CUIWnd::CompleteConstruction() Line 172	C++
 	FreeOrion.exe!GovernmentWnd::CompleteConstruction() Line 1578	C++
 	FreeOrion.exe!GG::Wnd::Create<GovernmentWnd,std::string const &>(const std::string & <args_0>) Line 348	C++
 	FreeOrion.exe!MapWnd::CompleteConstruction() Line 1464	C++
 	FreeOrion.exe!GG::Wnd::Create<MapWnd>() Line 348	C++
````

Comment (TheSilentOne1) [reply UI/GovernmentWnd.cpp]: Found it: had to call RadioButtonGroup::ExpandButtons right after creating the RadioButtonGroup.

**Comment by TheSilentOne1** [comment]
I'm almost through with touching up the icons. Now would be a good time for any final feedback, especially on the code.
Don't comment on the ugly black market icon. Thanks 😁

File: UI/GovernmentWnd.cpp [hunk 6]
```diff
@@ -229,35 +238,56 @@ void PolicyControl::CompleteConstruction() {
     GG::Control::CompleteConstruction();
     if (!m_policy)
         return;
+    int empire_id = GGHumanClientApp::GetApp()->EmpireID();
+    const ScriptingContext context;
+    const auto& name = UserString(m_policy->Name());
+    const auto& cost = static_cast<int>(m_policy->AdoptionCost(empire_id, context));
```
geoffthemedio: don't need a reference to store an int returned by value

geoffthemedio:
```suggestion
    const auto cost = static_cast<int>(m_policy->AdoptionCost(empire_id, context));
```

UI/GovernmentWnd.cpp
```diff
@@ -229,35 +238,56 @@ void PolicyControl::CompleteConstruction() {
...
 }
 
+void PolicyControl::Resize(const GG::Pt& sz, const int pts) {
+    m_background->Resize(sz);
+    m_icon->Resize(GG::Pt(sz.x, sz.y * 2/3));
+
+    std::shared_ptr<GG::Font> font = ClientUI::GetBoldFont(pts);
+
+    m_name_label->SetFont(font);
+    m_cost_label->SetFont(font);
```
geoffthemedio:
```suggestion
    m_cost_label->SetFont(std::move(font));
```
if you don't use the poniter again, can move it

UI/GovernmentWnd.cpp
```diff
@@ -524,6 +571,20 @@ void PoliciesListBox::HideAllCategories(bool refresh_list) {
         Populate();
 }
 
+void PoliciesListBox::ResizePolicies(const GG::Pt& sz, const int pts) {
+    auto it = begin();
```
geoffthemedio, line 575: didn't like / want to use something like my version?

UI/GovernmentWnd.cpp
```diff
@@ -1468,15 +1599,50 @@ void GovernmentWnd::Refresh() {
...
+    case 2: return 0.5;
+    }
+
+    DebugLogger() << "GovernmentWnd::GetPolicyZoomFactor(): m_policy_buttons no button checked.";
+    return 1;
+}
+
+GG::Pt GovernmentWnd::GetPolicySlotSize() {
+    double zoom_factor = GetPolicyZoomFactor();
+    const GG::X slot_width = static_cast<GG::X>(SLOT_CONTROL_WIDTH * zoom_factor);
```
geoffthemedio: `static_cast<GG::X>` seems odd... normally they can be created by calling the constructor.

UI/GovernmentWnd.cpp
```diff
@@ -1468,15 +1599,50 @@ void GovernmentWnd::Refresh() {
...
+    const GG::Y slot_height = static_cast<GG::Y>(SLOT_CONTROL_HEIGHT * zoom_factor);
+    return GG::Pt(slot_width, slot_height);
+}
+
+int GovernmentWnd::GetPolicyTextSize() {
+    double zoom_factor = GetPolicyZoomFactor();
+    const int text_pts = static_cast<int>(ClientUI::Pts() * zoom_factor);
+    return text_pts;
```
geoffthemedio:
```suggestion
    return static_cast<int>(ClientUI::Pts() * zoom_factor);
```

geoffthemedio:
A couple quirks. The clear button gets hidden. Might be OK I suppose. The policies adopted still overlap the right edge sometimes.

policies_layout

geoffthemedio:
A couple quirks. The clear button gets hidden. Might be OK I suppose. The policies adopted still overlap the right edge sometimes.

policies_layout

UI/GovernmentWnd.cpp
TheSilentOne1: No, I tried, but didn't manage to access the list's items.

TheSilentOne1:
I guess I could just move the clear button to the top, so that it is to the right of the size buttons.

TheSilentOne1:
I guess I could just move the clear button to the top, so that it is to the right of the size buttons.

TheSilentOne1:
I rearranged the buttons, but now there is a small gap above them. If I move the buttons towards the top, they become cut off by the policy palette, if I extend the policy palette towards the bottom, the upper half of the clear button becomes unclickable, so I settled for this. (I guess it might be avoided by moving the clear button into the GovermentWnd, should I do that?)

gap

geoffthemedio:
Why the difference? The previous version had no gap...

TheSilentOne1:
The policy buttons are a member of GovernmentWnd and work even if the are positioned more to the top. For the sizebutton, member of Mainpanel, it is a different matter: move it to the top and the upper half is visible but can't be clicked (is behind the policy palette then, I presume).

geoffthemedio:
the clear button can be a child of GovernmentWnd, then

TheSilentOne1:
ok

TheSilentOne1:
Looks good now I think. Gonna do some last icon polishing tomorrow and squash commits then.

TheSilentOne1:
Needs one final clean-up before merging.

TheSilentOne1:
Good to go, I think.

Oberlus:
Oooooof, I'm eager to play with this.

PR #3615, opened 2021-12-05
Refactored and fixed starlane bores

### LINKED ISSUE #3599: Starlane bore are random and may create very long lanes
Pedia for Starlane bore says:
Generates a single new starlane to the nearest system that can be connected without crossing existing starlanes or passing too close to another system.

Well, that was definitely not the nearest one:
LongStarlane

Code was today's master, but I think I've seen unexpected results from starlane bores before.
Never such a dramatic one for, though ;-)

Grummel7:
- Fixed ObjectTooCloseToLane
- Starlane Nexus no works sequentially so that each new lane
  is also checked against the already created ones
- Bore and Nexus have both a positive and a negative
  SitRepMessage

### DIFF
--- a/default/scripting/buildings/STARLANE_BORE.focs.txt
+++ b/default/scripting/buildings/STARLANE_BORE.focs.txt
@@ -5,27 +5,38 @@ BuildingType
     buildtime = 1 + Target.System.NumStarlanes
     location = Planet
     effectsgroups = [
+        [[SPECIES_LIKES_OR_DISLIKES_BUILDING_STABILITY_EFFECTS]]
+
         EffectsGroup
             scope = Object id = Source.SystemID
+            activation = Not [[ BORE_POSSIBLE ]]
             effects = [
-                AddStarlanes endpoint = MinimumNumberOf
-                number = 1
-                sortkey = DirectDistanceBetween object = Source.ID object = LocalCandidate.ID
-                condition = And [
-                    System
-                    CanAddStarlanesTo condition = Source
-                ]
+                GenerateSitRepMessage
+                    message = "EFFECT_STARLANE_NEXUS_FAILED"
+                    label = "EFFECT_STARLANE_NEXUS_LABEL"
+                    icon = "icons/tech/n-dimensional_structures.png"
+                    parameters = tag = "system" data = Target.ID
+                    empire = Source.Owner
+            ]
+
+        EffectsGroup
+            scope = Object id = Source.SystemID
+            activation = [[ BORE_POSSIBLE ]]
+            effects = [
+                [[ DO_STARLANE_BORE ]]
                 GenerateSitRepMessage
                     message = "EFFECT_STARLANE_BORE"
                     label = "EFFECT_STARLANE_BORE_LABEL"
                     icon = "icons/tech/n-dimensional_structures.png"
                     parameters = tag = "system" data = Target.ID
                     empire = Source.Owner
-                ]
+            ]
+
         EffectsGroup
             scope = Source
             effects = Destroy
     ]
     icon = "icons/tech/n-dimensional_structures.png"
 
+#include "/scripting/buildings/buildings.macros"
 #include "/scripting/common/base_prod.macros"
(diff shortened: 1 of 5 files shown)


default/scripting/buildings/STARLANE_BORE.focs.txt
```diff
@@ -5,27 +5,38 @@ BuildingType
     buildtime = 1 + Target.System.NumStarlanes
     location = Planet
     effectsgroups = [
+        [[SPECIES_LIKES_OR_DISLIKES_BUILDING_STABILITY_EFFECTS]]
```
Grummel7, line 8: Note sure whether having (dis-)likes for temporary buildings makes much sense, but Bore and Nexus should either both support them or both do not.

default/scripting/buildings/buildings.macros
```diff
@@ -104,3 +104,22 @@ SPECIES_LIKES_OR_DISLIKES_BUILDING_STABILITY_EFFECTS
...
                 ) * [[POLICY_LIKE_DISLIKE_SCALING]] * SpeciesContentOpinion species = Target.Species name = ThisBuilding
 '''
+
+DO_STARLANE_BORE
+'''AddStarlanes endpoint = MinimumNumberOf
+       number = 1
+       sortkey = DirectDistanceBetween object = Source.ID object = LocalCandidate.ID
+       condition = And [
+           System
+           CanAddStarlanesTo condition = Source
```
Grummel7, line 114: Nexus had a maximum length, Bore did not. I think they should both do the same, so if there shall be a limits, we add it here (and in BORE_POSSIBLE below)

default/scripting/buildings/STARLANE_BORE.focs.txt
```diff
@@ -5,27 +5,38 @@ BuildingType
...
+
+        EffectsGroup
+            scope = Object id = Source.SystemID
+            activation = Not [[ BORE_POSSIBLE ]]
```
Grummel7, line 12: There is still a minor issue here: If multiple bores trigger in the same turn, all report success, even if some of them could not create a new lane.

default/scripting/buildings/STARLANE_BORE.focs.txt
geoffthemedio: probably not necessary

universe/Conditions.cpp
```diff
@@ -8386,18 +8386,19 @@ namespace {
...
-        //  /a___|___
+        //     / |   1a
+        //    /  |d  |
+        //   /   |   |
+        //  /a___|___|
         // 1         2
-        // (1O).(12) = |1O| |12| cos(a)
-        // d = |1O| cos(a) = (1O).(12) / |12|
-        // d = -(O1).(12 / |12|)
+        // (1O).(12) = |1O| |12| sin(a)
```
geoffthemedio: Need some explanation here...  `A•B = |A||B|cos(α)` where **α** is the angle between the vectors **A** and **B**.

What is **1a** and how is it different from **O** and what does it have to do with **d** and why isn't it just **3** ?

In the older version, **O** is the location of the object, and **1** and **2** are the ends of a starlane that we want to calculate how far is away from **O** at the closest point to the line through those stars. It tries to use a vector projection, but is probably calculating the projection of **1->0** onto **1->2** rather than the perpendicular distance.

I think your calculation is OK, but it doesn't need the confusing **1a** in the comment...

rather maybe:
````
        // distance to point from line from vector projection / dot products
        //       O
        //      /|
        //     / |
        //    /  |d
        //   /   |
        //  /a___|___
        // 1         2
        // (1O)x(12) = |1O| |12| sin(a)
        // d = |1O| sin(a) = (1O)x(12) / |12|
        // d = (10)x(12 / |12|)
        constexpr float MIN_PERP_DIST = 20; // magic limit, in units of universe units (uu)

        float perp_dist = std::abs(v_o1_x*v_12_y - v_o1_y*v_12_x);
````

default/scripting/buildings/STARLANE_BORE.focs.txt
geoffthemedio: probably less of an issue than what it's replacing...

universe/Conditions.cpp
Grummel7: I haven't done much geometry in the last 30 years. I had to figure out what a dot product was and now after checking for cross product I agree that your explanation is better. Will replace the comment.

PR #3619, opened 2021-12-22
Check If Producible Before Enqueuing in AI

geoffthemedio:
Adds functions to the Python API to check of buildings or ships are enqueuable or produbile. Can be used rather than attempting to enqueue them.

Added the checks before attempting to enqueue Scanning Facility and Orbital Drydocks, to reduce error spam in the AI log.

### DIFF
--- a/client/AI/AIWrapper.cpp
+++ b/client/AI/AIWrapper.cpp
@@ -358,6 +358,30 @@ namespace {
 
     }
 
+    auto IsProducibleBuilding(const std::string& item_name, int location_id) -> bool
+    {
+        ScriptingContext context;
+        int empire_id = AIClientApp::GetApp()->EmpireID();
+        auto empire = context.GetEmpire(empire_id);
+        if (!empire) {
+            ErrorLogger() << "IsProducibleBuilding : couldn't get empire with id " << empire_id;
+            return false;
+        }
+        return empire->ProducibleItem(BuildType::BT_BUILDING, item_name, location_id, context);
+    }
+
+    auto IsEnqueuableBuilding(const std::string& item_name, int location_id) -> bool
+    {
+        ScriptingContext context;
+        int empire_id = AIClientApp::GetApp()->EmpireID();
+        auto empire = context.GetEmpire(empire_id);
+        if (!empire) {
+            ErrorLogger() << "IsEnqueuableBuilding : couldn't get empire with id " << empire_id;
+            return false;
+        }
+        return empire->EnqueuableItem(BuildType::BT_BUILDING, item_name, location_id, context);
+    }
+
     auto IssueEnqueueBuildingProductionOrder(const std::string& item_name, int location_id) -> int
     {
         ScriptingContext context;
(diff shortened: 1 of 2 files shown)


geoffthemedio:
@Morlic-fo your comment on this would also be helpful...

default/python/AI/ProductionAI.py
```diff
@@ -1082,6 +1087,10 @@ def generate_production_orders():
                 planet = universe.getPlanet(pid)
                 if not planet:
                     continue
+                if not fo.isProducibleBuilding(building_name, pid):
+                    continue
```
Morlic-fo, line 1091: I think silent failure here will make debugging hard, so please add some logging statement before the continue statement (also below).

Morlic-fo (changes requested):

Morlic-fo:
Overall definitely helpful to have in the API but see comment re usage.

default/python/AI/ProductionAI.py
geoffthemedio: The idea here isn't that it's a failure, but that it's being used to search for valid places to produce the building. Prior to this change, it just attempts to issue the enqueue order at a bunch of places until one succeeds.

client/AI/AIWrapper.cpp
```diff
@@ -389,6 +413,31 @@ namespace {
         return 1;
     }
 
+    auto IsProducibleShip(int design_id, int location_id) -> bool
```
Cjkjvfnby: According to current usage `IsEnqueuableShip` will be enough.  So let's keep the API small and remove this function. 

Having two functions implies that one or both cases are valid `enqueueable and not producible` and `not enqueueable and producible`.   This makes things much complex, since knowledge about that difference should be shared (via methods documentation).

geoffthemedio: There is a difference: buildings can be producible but not enqueuable. This is used for cases like only letting the player have one of something on the queue at a particular location. That is, another one can't be enqueued, but the one on the queue will still received PP allocations (is producible). The same functions were added for ships for consistency with buildings. Can be removed if you prefer, though, I suppose...

Cjkjvfnby: > The same functions were added for ships for consistency with buildings.

Let's remove it from ships. It has no sense in that context.


For buildings it has sense, but at current time it will be enough to have only one of them.  And I prefer to trade flexibility for simplicity and leave a single function as well. 

If you prefer to keep them please add a documentation string:

"Return True if a building could be queued. The building could be enqueued if it is producible and not already in the queue at this location."

"Return True if a building producible.  This implies that unlocking technology is researched and location has all requirements for it (planet type, prerequired buildings are built or enqueued). To check that building is enqueuable at this location please use IsEnqueuableShip"

I am not sure that my descriptions are correct. And this is a problem a new developer will face as well.

geoffthemedio: done?

Cjkjvfnby (approved):

geoffthemedio:
https://github.com/freeorion/freeorion/commit/a7e5d216b78043a86ce2f4cbeaa865a1787360fa

PR #3620, opened 2021-12-22
Fix timeout behaviour in ConnectToServer()

Morlic-fo:
Previous to this patch, small timeout values were effectively ignored. This happens because boost::asio::io_context::run() is a blocking call and the attempt to async_connect() may time out after e.g. 2 seconds on my machine. This means any shorter timeout parameters will effectively be ignored and any longer timeout still only has a resolution of, e.g., 2 seconds.

The fix proposed here is to introduce an asynchronous timer which will cancel pending operations once timeout occurs by closing the socket.

The improved behavior is most easily noticed when pressing the "Single Player" button in the intro menu. This results in a check for existing local servers implemented by a connection attempt with a small timeout parameter (i.e. 100 ms). But since that parameter may be effectively ignored due to the existing bug, a significantly longer freeze of the UI happens (i.e. 2 seconds for me).
With the patch applied, the Galaxy Setup Window will come up after the expected ~100 ms (assuming parsers have finished).

### DIFF
--- a/client/ClientNetworking.cpp
+++ b/client/ClientNetworking.cpp
@@ -227,8 +227,10 @@ public:
 
 private:
     void HandleException(const boost::system::system_error& error);
-    void HandleConnection(boost::asio::ip::tcp::resolver::iterator* it,
-                          const boost::system::error_code& error);
+    void HandleConnection(const boost::system::error_code& error,
+                          tcp::resolver::iterator endpoint_it);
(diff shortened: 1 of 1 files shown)


geoffthemedio:
@Morlic-fo Is this intentionally still marked work in progress? Seems to work for me, and does seem to make the galaxy setup appear faster...

geoffthemedio:
@Morlic-fo poke

Morlic-fo:
Cleaned up commits and removed WIP label.

PR #3621, opened 2021-12-23
Buff population growth to thirdway where it was before

agrrr3:
Since the last nerf, normal colony growth is pale in comparison to invading natives.

forum discussion started here https://www.freeorion.org/forum/viewtopic.php?p=108840#p108840
and there is discussion of changing [Population curves](https://www.freeorion.org/forum/viewtopic.php?f=6&t=12222)

Before introducing growth policy in 5c9b5023f00e047f1b33e260677a13f21d4f6c2b,

base growth factor was 0.01
Then it was 0.005, but one could improve it by 50% using the policy
New formula kept that balance for planets of target population 16 (using a base growth factor of 0.08).

with this PR, base growth factor is 0.1; 25% higher (so upping growth a bit)

(Note: with old formula i suggested improving it to 0.008 which would have been ~50% more)

### DIFF
--- a/default/scripting/species/common/population.macros
+++ b/default/scripting/species/common/population.macros
@@ -1,7 +1,7 @@
 
 GROWTH_RATE_FACTOR
 '''
-0.08 *
+0.1 *
 (1 - (Statistic If condition = And [Target EmpireHasAdoptedPolicy name = "PLC_NO_GROWTH"])) *       // no growth with no-growth policy
 (1 + 0.5*(Statistic If condition = And [Target EmpireHasAdoptedPolicy name = "PLC_POPULATION"])) *  // +50% growth with population policy
 (1 + (Statistic If condition = And [Target EmpireHasAdoptedPolicy name = "PLC_AUGMENTATION"]) *     // slower growth with augmentation on low-infrastructure planets


agrrr3:
the failing check is unrelated - the snapcraft(?) service was not available it seems after building. And well, I changed a number in FOCS.

Grummel7:
Could you perhaps include [this](https://www.freeorion.org/forum/viewtopic.php?f=6&t=12222) discussion?

Using the formula suggested by Oberlus, even with 0.008, large planets would take longer become full, but small one would fill a lot quicker.

agrrr3:
as discussion is ongoing (and where it belongs - in the forums), i put this on hold / as a draft for the moment

agrrr3:
a better local formula was implemented, but AFAIU it does not really speed up growth

 @Grummel7 is that following comment about growth correct?

it is the same as 0.005 for the old formula, if the planet has a max population of 16; for smaller planets it is faster and for bigger planets it is even slower. i suggest using 0.1 for a starter (+25% growth)

Grummel7:
> @Grummel7 is that following comment about growth correct?
> 
> it is the same as 0.005 for the old formula, if the planet has a max population of 16; for smaller planets it is faster and for bigger planets it is even slower. i suggest using 0.1 for a starter (+25% growth)

Yes, and with 0.1 it would be the same for max 20. That's okay for me.

geoffthemedio:
@agrrr3 what's the status of this?

agrrr3:
> @agrrr3 what's the status of this?

taking grummel's comment into account - and personal opinion, i think we should simply merge

OTOH i found the comments in [forum](https://www.freeorion.org/forum/viewtopic.php?p=111604#p111604) not in complete agreement.

for ramp down to 0.5 i will concentrate on peaceful hidden expansion, colony growth probably helps with that

PR #3622, opened 2021-12-23
PLC_LIBERTY kicks in at stability 6 iff target_stability without liberty >= 8

agrrr3:
https://www.freeorion.org/forum/viewtopic.php?f=28&t=11768&p=108838#p108838

Stability 9 may invite to some focus (or other) dancing trying to keep stability at 8 which is NOT ok .
one could make it though something like this (curStability is current stability when the effect gets applied); making getting all the 8+ stability gated bonus harder:

late effect, after second scaling:
* (stability >= 6)&&(targetStability >= 8 ) -> +2 targetResearch, -2 targetStability

### DIFF
--- a/default/scripting/policies/LIBERTY.focs.txt
+++ b/default/scripting/policies/LIBERTY.focs.txt
@@ -16,13 +16,19 @@ Policy
             Planet
             OwnedBy empire = Source.Owner
             Species
-            Happiness low = (NamedReal name = "PLC_LIBERTY_MIN_STABILITY" value = 8)
+            Happiness low = (NamedReal name = "PLC_LIBERTY_MIN_STABILITY" value = 8.0)
+                          - (NamedReal name = "PLC_LIBERTY_STABILITY_PENALTY_FLAT" value = 2.0)
         ]
+        priority = [[TARGET_AFTER_2ND_SCALING_PRIORITY]]
         effects = [
-            SetTargetResearch value = Value + (NamedReal name = "PLC_LIBERTY_RESEARCH_BONUS_FLAT" value = 2.0)
-            SetTargetHappiness value = Value - (NamedReal name = "PLC_LIBERTY_STABILITY_PENALTY_FLAT" value = 2.0)
+            If condition = (Value(RootCandidate.TargetHappiness) >= (NamedRealLookup name = "PLC_LIBERTY_MIN_STABILITY"))
+            effects = [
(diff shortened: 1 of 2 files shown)


agrrr3:
not tested yet

geoffthemedio:
Can just have two separate effects groups, where the penalty doesn't have a minimum stability threshold to be applied, but the bonus does.

Grummel7:
`NamedRealLookup name = "PLC_LIBERTY_STABILITY_PENALTY_FLAT" value = 2.0`
I'm surprised this even works, but eanyway, the value should be removed. Plus in the effect you use a definition without a value.

If you don't want to go below the threshold, why not use:
`SetTargetHappiness value = max(Value - NamedRealLookup name = "PLC_LIBERTY_STABILITY_PENALTY_FLAT", (NamedRealLookup name = "PLC_LIBERTY_MIN_STABILITY"))`

agrrr3:
> Can just have two separate effects groups, where the penalty doesn't have a minimum stability threshold to be applied, but the bonus does.

ah, i thought the only-a-downside-if-there-is-an-upside was intended (i found that quite nice actually). but liberty is anyway very powerful, so destabilizing the whole empire with it is probably ok.

agrrr3:
> If you don't want to go below the threshold, ...

i do want to go below the threshold - liberty makes other 8 stability bonus harder to get, which is good because liberty is also very powerful

agrrr3:
> `NamedRealLookup name = "PLC_LIBERTY_STABILITY_PENALTY_FLAT" value = 2.0` I'm surprised this even works, but eanyway, the value should be removed. Plus in the effect you use a definition without a value.

yeah, it doesn't work. must have not really playtested after doing last cleanup-refactoring

agrrr3:
> > Can just have two separate effects groups, where the penalty doesn't have a minimum stability threshold to be applied, but the bonus does.
> 
> ah, i thought the only-a-downside-if-there-is-an-upside was intended (i found that quite nice actually). but liberty is anyway very powerful, so destabilizing the whole empire with it is probably ok.

@geoffthemedio do you prefer the unconditional separate effectgroups to the only-a-downside-if-there-is-an-upside?

geoffthemedio:
The simpler implementation with an unconditional effect is probably better, I think...

agrrr3:
> The simpler implementation with an unconditional effect is probably better, I think...

ok, i will do another PR with that and leave this as a draft for the moment

PR #3627, opened 2021-12-26
Automatically flattens And and Or conditions

o01eg:
Make conditions `And [ A B C ]`, `And [ And [ A B ] C ]`, and pythons `A & B & C` equals.

### DIFF
--- a/universe/Conditions.cpp
+++ b/universe/Conditions.cpp
@@ -10311,8 +10311,13 @@ And::And(std::unique_ptr<Condition>&& operand1, std::unique_ptr<Condition>&& ope
     Condition()
 {
     // would prefer to initialize the vector m_operands in the initializer list, but this is difficult with non-copyable unique_ptr parameters
-    if (operand1)
-        m_operands.push_back(std::move(operand1));
+    if (operand1) {
+        if (And* operand1_and = dynamic_cast<And*>(operand1.get())) {
+            m_operands = std::move(operand1_and->m_operands);
+        } else {
+            m_operands.push_back(std::move(operand1));
+        }
+    }
     if (operand2)
         m_operands.push_back(std::move(operand2));
     if (operand3)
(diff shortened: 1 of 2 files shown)


geoffthemedio:
I think it will work OK, though only for nested `And` or `Or` conditions where the first subcondition is the same as the enclosing one, and not for any of the other subconditions.

o01eg:
With this PR applied #3613 produces same checksum before and after.

geoffthemedio:
Sounds like it's working, for what (limited) it does, then.

agrrr3:
@o01eg  cool! do i understand correctly? you did some optimization on the python AST and that made the checksums mismatch, so you changed the conditions in cpp to have the same structure after parsing? i am really looking forward to the AST power. how far are you with the python parsing? after switching we can do some fancy optimizations I think.

o01eg:
@agrrr3 No, I simply made optimization to common universe code which independent of parser but also unifies same but not equal structures.

agrrr3:
> @agrrr3 No, I simply made optimization to common universe code which independent of parser but also unifies same but not equal structures.

@o01eg oh, I got carried away. I am really looking forward to the AST.

universe/Conditions.cpp
```diff
@@ -10311,8 +10311,13 @@ And::And(std::unique_ptr<Condition>&& operand1, std::unique_ptr<Condition>&& ope
     Condition()
 {
     // would prefer to initialize the vector m_operands in the initializer list, but this is difficult with non-copyable unique_ptr parameters
-    if (operand1)
-        m_operands.push_back(std::move(operand1));
+    if (operand1) {
```
geoffthemedio, line 10314: @o01eg any reason not to also consider other subconditions besides `operand1` ?

o01eg: Other cases are rare as we don't use nested And or Or in FOCS. I could add it in other PRs. Maybe it means we don't need separate flattening function.

PR #3633, opened 2021-12-30
Restrict Metropoles

geoffthemedio:
-reduce number of metropoles to at most 3
-require support planets to have at least 5 population and to not have been invaded or colonized in the last 10 turns

### DIFF
--- a/default/scripting/policies/METROPOLES.focs.txt
+++ b/default/scripting/policies/METROPOLES.focs.txt
@@ -14,12 +14,15 @@ Policy
         // Note that to raise a planet to a metropole, you have to reach the minimum stability
         // while suffering from the non-metropole penalty below.
         EffectsGroup
-            scope = MaximumNumberOf number = min((NamedInteger name = "METROPOLES_PLANET_LIMIT" value = 6),
+            scope = MaximumNumberOf number = min((NamedInteger name = "METROPOLES_PLANET_LIMIT" value = 3),
                                                  floor(Statistic Count condition = And [
-                                                           Planet
-                                                           OwnedBy empire = Source.Owner
-                                                           Species
-                                                       ] / (NamedInteger name = "METROPOLES_MIN_PLANETS" value = 4)))
+                                                    Planet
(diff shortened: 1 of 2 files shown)


hifron:
Political parties in dominant countries have systems which make dominant 2 political parties, but that doesn't mean that that system is much mandatory advanced. Dominion in some areas sometimes means that something requires something more, less or some another way, but is not possible in that system. Maybe Metropoles policy needs further research...

Grummel7:
TurnsSinceLastConquered returns 0 when the planet was never conquered, so with the current implementation only conquered planets can support at all.

You could also add a comment that, while there is no direct time requirement for the metropoles themselves, the minimum stability basically acts as one. If you conquer a big planet, it will take at least 12 turns until it becomes a metropole. Perhaps with kickstart and evacuation you can turn a new colony into a metropole in less, but that is rather a theoretic issue.

hifron:
Or there could be option at generation of universe that setup such policy, support planets maybe could have stacks...

Oberlus:
> require support planets to have at least 5 population

That's a substantial change. Currently, to get your first metropole you need to have 3 colonies (even just colonized with 1 pop). Making it require pop 5 for the supporting colonies makes players with tiny/poor planets around their HW unable to use the later until later when they expand further or enlarge population limits in current space via techs.
What's the aim of this change? Is Metropoles OP?

agrrr3:
i find metropoles in the current state certainly not overpowered

Vezzra:
What's the status of this PR anyway? It has been opened a year ago, got the last few comments about a month later, and has been idle ever since... @geoffthemedio? Are the proposed changes still relevant?

geoffthemedio:
I don't remember what the motivation was or if it's still relevant.

Vezzra:
@geoffthemedio, well, it's your PR, so it's up to you what you want to do with it...

geoffthemedio:
I suppose if no-one is commenting / complaining about that policy's balance now, the pull request can be abandoned.

PR #3634, opened 2021-12-31
Replace Terraforming (reversion) by targeted Terraforming.

Grummel7:
This allows Exobots to do Terraforming, but no longer allows to do several steps in parallel.
See also the following [Complete rework of Terraforming](https://www.freeorion.org/forum/viewtopic.php?f=6&p=109043#p109043)

The only problem at the moment is that the client shows wrong cost for the advanced step when more than one step is enqueued. I guess I can somehow calculate the correct costs by taking into account all other enqueued and built terraformings, but that would explode the already complicated formula for buildcost.

What I need is a way to determine the distance of a given type from the planets original type. So instead of saying `1 + Target.DistanceFromOriginalType +/- <queued changes>` I could simply say `Distance type1 = Target.OriginalType type2 = @2@` or even `Target.OriginalType <> @2@`. Okay, it would make reverse terraforming a little cheaper, but I doubt that would be a game breaker.


geoffthemedio:
Based on the linter, need to remove the now-unused BLD_TERRAFORM, BLD_TERRAFORM_DESC, BLD_TERRAFORM_REVERT, and BLD_TERRAFORM_REVERT_DESC

Grummel7:
Note sure whether I was supposed to modify all the translation files. Done it now.
Any ideas for the best way to determine the distance between two environment types?

geoffthemedio:
> Any ideas for the best way to determine the distance between two environment types?

It's difficult...

`ValueRef::Operation` return the same type as their input type, so it wouldn't be possible to specify two `PlanetType` values and get back an int or double value representing their difference using a standard difference expression.

There are type-casting ValueRefs, but those just cast, without doing an operation before or after.

There are `ValueRef::Complex` that can take parameters of one type and return another type, but those don't support parameters of type `PlanetType`. It could be extended to support that, or possibly it could be somewhat akwardly worked around by having something that takes an arbitrary enum ValueRef type and casts it to an int-typed ValueRef, and then knows to cast it back to an enum for doing the evaluation, and then convert that back into an int to return. I'm not sure if `ValueRef::StaticCast` can support enum-valued to int-valued casting, but it could probably be made to. I'd probably prefer that to adding more types to the internals of `ValueRef::Complex`.

Grummel7:
Well, a type cast to int or float may not be ideal, but it would allow to calculate the difference in a macro roughly like this:
`ENVIRONMENT_DISTANCE
'''Min((Int(@1@) - Int(@2@)) % 9, (Int(@2@) - Int(@1@)) % 9)'''`

Perhaps that would be the easiest way, if we stick to this terraforming solution at all.

geoffthemedio:
Something like that could work as long as it doesn't ever need to handle any of the planet types not in the loop, like gas giant or asteriods.

But that also doesn't really properly handle the loop arrangement, as say `Swamp` = 0 and `Ocean` = 8, but they should be just 1 apart, not 8.

Grummel7:
Argh, % in C++ can return negative numbers, but then the macro will yield -8, not 8.
Anyway, adding 9 before the modulo will fix it.

geoffthemedio:
probably `abs(...)` would make sense...

Grummel7:
Well, no, abs would exactly do the wrong thing:
abs(0-8)%9 = abs((0-8)%9) = 8


What we need is:
`ENVIRONMENT_DISTANCE '''Min((9 + Int(@1@) - Int(@2@)) % 9, (9 + Int(@2@) - Int(@1@)) % 9)'''`
(9+0-8)%9 = 1

geoffthemedio:
I was thinking something like `min(abs(Int(@2@) - Int(@1@)), 9 - abs(Int(@2@) - Int(@1@)))`

effectively the same as using `%` I suppose

default/scripting/buildings/TERRAFORM.focs.txt
```diff
@@ -1,32 +1,55 @@
...
+[[TARGET_TERRAFORMING(Ocean,Swamp,Toxic,TERRAN,OCEAN,SWAMP,TOXIC,INFERNO)]]
+[[TARGET_TERRAFORMING(Swamp,Toxic,Inferno,OCEAN,SWAMP,TOXIC,INFERNO,RADIATED)]]
+[[TARGET_TERRAFORMING(Toxic,Inferno,Radiated,SWAMP,TOXIC,INFERNO,RADIATED,BARREN)]]
+[[TARGET_TERRAFORMING(Inferno,Radiated,Barren,TOXIC,INFERNO,RADIATED,BARREN,TUNDRA)]]
+[[TARGET_TERRAFORMING(Radiated,Barren,Tundra,INFERNO,RADIATED,BARREN,TUNDRA,DESERT)]]
+[[TARGET_TERRAFORMING(Barren,Tundra,Desert,RADIATED,BARREN,TUNDRA,DESERT,TERRAN)]]
+[[TARGET_TERRAFORMING(Tundra,Desert,Terran,BARREN,TUNDRA,DESERT,TERRAN,OCEAN)]]
+
+
+// @1@ cw previous type
```
agrrr3: would write out cw to clockwise etc..

agrrr3 (commented):
as far as i understand, you show the two possible environments one can terraform to directly.
sounds fine UI wise. A minor gripe is that you do not get feedback if you are going in the wrong direction/no preview for an existing population. What would be really great is if it would write in the build window "Terraform to Desert (adequate)". But that is not really possible currently (it is possible but not in a sensible way). Could add a sitrep for building in a worse direction.

I would also allow outposts to terraform (no need to drop exobots first). Could increase build turns if no/not enough population if you really want to. But why do you want to?

Grummel7:
> as far as i understand, you show the two possible environments one can terraform to directly. sounds fine UI wise. A minor gripe is that you do not get feedback if you are going in the wrong direction/no preview for an existing population. What would be really great is if it would write in the build window "Terraform to Desert (adequate)".

I think all that requires is a new SiteRep variable type (or rather two), but that should not be too complicated, I can give it a try.

> But that is not really possible currently (it is possible but not in a sensible way). Could add a sitrep for building in a worse direction.

Possibly, but it would be a little bit late if player is informed after completing the wrong step. Or is there a way to create a SiteRep while when a building is started being built? Note that as soon as you add one step, you see what the next one would be, that should help players finding the right direction.

> I would also allow outposts to terraform (no need to drop exobots first). Could increase build turns if no/not enough population if you really want to. But why do you want to?

Well, that's a point I'd rather discuss in the forum. TargetPopulation low = 1 was there before and changing it is more or less unrelated this refactoring.

o01eg (approved):
I suppose it's fine.

geoffthemedio:
`Evacuate Planet` is phrased like an imperative verb, so probably the terraforming buildings should be phrased like `Terraform to Desert`

Grummel7:
> `Evacuate Planet` is phrased like an imperative verb, so probably the terraforming buildings should be phrased like `Terraform to Desert`

Yes, and more importantly the descriptions were using the wrong verb form, too. Fixed it.

Grummel7:
> I was thinking something like `min(abs(Int(@2@) - Int(@1@)), 9 - abs(Int(@2@) - Int(@1@)))`
> 
> effectively the same as using `%` I suppose

That formula didn't occur to me, it does the same indeed.

Can you help me with this int-cast? The only thing that I've found that looks like an explicit cast is part_class_as_int, but it is only used as a sub-expression in a Complex Int. Or did you think about castable_as_int_parser_rules?

Perhaps we can also merge this PR as it is now and deal with this in due time. It's only the forecast of future costs that is wrong right now.

agrrr3:
> Int(@2@)

Explicit casts/numeric type transformations do not exist in FOCS as far as I know. Are they necessary here? ( Maybe there are implicit transformations when using % so that could work.)

Grummel7:
> > Int(@2@)
> 
> Explicit casts/numeric type transformations do not exist in FOCS as far as I know. Are they necessary here? ( Maybe there are implicit transformations when using % so that could work.)

Well, what I do need is a way to determine the distance between two planet types.

Geoff mentioned that casts exist and that would allow me to do it, whether explicit or implicit. An specialised complex int function would also be fine, but adding planet type as an argument to complex int is beyond my code knowledge, too.

agrrr3:
ok. maybe i got it now. The question is how you get an integer out of planettype? 
Or actually you would rather want `(Counter)ClockwiseDistance early = <PLANETTYPE> later = <PLANETTYPE>` complex valueref?

i had a short look at the parsers, i probably wont be able to add such a `ClockwiseDistance` soonish. probably adding an `Int` valueref is easier. actually adding a `PlanetTypeAsInt` should be the easiest to implement

geoffthemedio:
Be aware of https://github.com/freeorion/freeorion/blob/master/universe/ValueRefs.cpp#L1986

Also note that it's relatively easy to add a [parser that converts a constant enum expression](https://github.com/freeorion/freeorion/blob/master/parse/IntComplexValueRefParser.cpp#L155) of a PlanetType like **Ocean** or **Inferno** to an int, but somewhat more complicated to get it working for arbitrary non-constant expressions like **Target.OriginalType**.

agrrr3:
> Be aware of https://github.com/freeorion/freeorion/blob/master/universe/ValueRefs.cpp#L1986

thanks for adding it. I guess the parser part is still missing(?)

agrrr3:
just for your info, i am trying to add a parser for that PlanetTypeDifference complex valueref, and currently in boost parser induced cpp depency hell.

geoffthemedio:
as expected...

> somewhat more complicated to get it working for arbitrary non-constant expressions

agrrr3:
> as expected...
> 
> > somewhat more complicated to get it working for arbitrary non-constant expressions

yes, i think i can manage anyway - i hope some forward declaration will be the solution. anyway cant wait for c++20 modules to be available to us

edit1: unluckily forward declaration needs rather a reference (or pointer), but I do not have a good place to initialize the star_type_parser_rules and pass it into the IntComplexValueRefParser... so I am trying to get around by declaring a unique_ptr with forward declared type and creating and saving it in the int_complex_parser_grammar struct ctor. Not sure what the lifelines will be though. Also it seems the uniqueptr needs to know the memory layout/size in advance to be able to auto-delete, so getting around the incomplete type did not really work. ...

edit3: ok using raw pointer or std_shared ptr works; else for unique_ptr it seems we would have to declare a custom deletor

geoffthemedio:
I got about this far: https://github.com/freeorion/freeorion/commit/d338edb13b45111eb63f83d0bb86c0b7a4f058fa

agrrr3:
> I got about this far: [d338edb](https://github.com/freeorion/freeorion/commit/d338edb13b45111eb63f83d0bb86c0b7a4f058fa)

i get it that your planet_type_as_int does not compile, right?
mine is similar, i tried to go the string identifier route but cant get the conversion to work.

geoffthemedio:
`planet_type_as_int` compiles on its own, but doesn't work when used in `planet_type_difference`

agrrr3:
made it work, will post PR after cleanup. #3681 

did the typing/transformations in a simple cpp function and then BOOST_PHOENIX_ADAPT_FUNCTION in order to use it in the parser action.

Grummel7:
> made it work, will post PR after cleanup. #3681
> 
> did the typing/transformations in a simple cpp function and then BOOST_PHOENIX_ADAPT_FUNCTION in order to use it in the parser action.

Great! Is there a way I can merge your not-yet-merged pull request into my branch to try it? I should I just wait until it is merged?

geoffthemedio:
> Is there a way I can merge your not-yet-merged pull request into my branch

Can access it with git? GitHub shows me these command line instructions:
````
git checkout -b Grummel7-terraforming master
git pull https://github.com/Grummel7/freeorion.git terraforming
````

If you can get a local copy, then you can cherry pick or merge locally.

**Comment by Grummel7** [comment]
> If you can get a local copy, then you can cherry pick or merge locally.

I am still rather new to git... Of course the pull request is based on a branch in agrrr3's repo, which I can access like any other.

But since you already created a branch in the official repo, I've merged planet_type_difference_parsing and now I can make it work:
`                    * (Target.HabitableSize * max(1, PlanetTypeDifference from = Target.OriginalType to = @2@)) * [[BUILDING_COST_MULTIPLIER]]`

Note that the max(1, ...) is needed to avoid terraforming back to the orignal type having cost 0. Could possibly use 0.5 instead to make it cheaper than going away from the original.

File: util/VarText.h [hunk 2]
```diff
@@ -179,6 +179,10 @@ class FO_COMMON_API VarText {
     static constexpr std::string_view METER_TYPE_TAG = "metertype";
     //! Variable value is the name of a registed ValueRef.
     static constexpr std::string_view FOCS_VALUE_TAG = "value";
+    //! Variable value is Planet::ID(), gives environment of that planet (Good, Poor...)
+    static constexpr std::string_view ENVIRONMENT_TAG = "environment";
+    //! Variable value is a user_string (translation string from <lang>.txt)
+    static constexpr std::string_view USER_STRING_TAG = "user_string";
```
geoffthemedio, line 185: seems unrelated...?

Grummel7: That is needed to implement agrrr3's suggestion for the SitRep (see en.txt):
`%planet% has been terraformed to %user_string% (%environment%).`

geoffthemedio:
> since you already created a branch in the official repo, I've merged planet_type_difference_parsing and now I can make it work:

Are you referring to https://github.com/freeorion/freeorion/commits/planet_type_difference_parsing ?

If so, should that be merged as is into master?

Grummel7:
> Are you referring to https://github.com/freeorion/freeorion/commits/planet_type_difference_parsing ?
> 
> If so, should that be merged as is into master?

Yes that is what I used.

I haven't looked at the details, but PlanetTypeDifference does work as expected.

geoffthemedio:
merged the parsing stuff to master

Grummel7:
Rebased on this one master

agrrr3:
PR looks good to me :)

geoffthemedio:
I think this missed some references to terraforming related content eg: https://github.com/freeorion/freeorion/blob/master/default/scripting/buildings/GAIA_TRANS.focs.txt#L17-L22

PR #3642, opened 2022-01-04
more names for ships and stars:

st-pa-fo:
- add few more names for ships (mainly mineralogy based)
- add myself as dev ship name
- add few more names for stars (star trek memory alpha)
- alphabetic reordering of some ship names
- reordering some ship names (as per #FIXME)
- correcting some typos in ship names (mainly umlauts of proper names)
- filled in "(this description is missing)" where the test script failed before

PR #3644, opened 2022-01-04
- add few more names for ships (mainly mineralogy based)

st-pa-fo:
- add myself as dev ship name
- add few more names for stars (star trek memory alpha)
- alphabetic reordering of some ship names
- reordering some ship names (as per #FIXME)
- correcting some typos in ship names (mainly umlauts of proper names)
- including feedback from https://github.com/freeorion/freeorion/pull/3642


st-pa-fo:
one more thing, scrolling was not fun, so i added 'support' for folding .. see screenshot


code folding thx 2 '{' + '}' in #'s

st-pa-fo:
darn it, the string test failing again. i can revert this, or add more newlines to trick the test script, or whatever. later.

Cjkjvfnby:
> darn it, the string test failing again. i can revert this, or add more newlines to trick the test script, or whatever. later.

You could reformat it with the same tool, which is used for format check.

If you use mac or linux it will look like this: 
```
./check/st-tool.py format default/stringtables/en.txt
```

hifron:
> You could reformat it with the same tool, which is used for format check.

Time to move to some online translators...

st-pa-fo:
> > darn it, the string test failing again. i can revert this, or add more newlines to trick the test script, or whatever. later.
> 
> You could reformat it with the same tool, which is used for format check.
> 
> If you use mac or linux it will look like this:
> 
> ```
> ./check/st-tool.py format default/stringtables/en.txt
> ```

thanx indeed, i'll try that next time, but it seems i already got it right by just trying often enough

st-pa-fo:
i have recombined it all into one commit with a cleaner message

PR #3646, opened 2022-01-06
purely cosmetic changes in english texts

st-pa-fo:
- typo corrections in existing texts
- beautifications of species-descriptions-tooltips / pedia-pages:
	-- less wrapped lines -> easier readable in game start screen
	-- added a neutral symbol ø to preexisting positive + and negative -

not sure what branch this would belong in, so master it is then.


st-pa-fo:
here example screenshots before and after:
before
after

geoffthemedio:
not sure about ø

what happens to the layout if you make the font substantially bigger or smaller?

st-pa-fo:
i did not try bigger/smaller fonts, but i found out the current font has limited support for unicode characters … all of these very simply displayed as space signs (so not displayed at all): ・⸰▫︎∘⚬✧▻♢⋄⁘⁖⸬⋯⧊෴꩜ – if the ø is to exotic, maybe just an ordinary o instead? it really looks better in line with those plusses and minusses

st-pa-fo:
i just maxed the font size to 40 and took a look. … and dude, there's really a lot more problems with the layout at bigger font sizes than just the ø – lots of stuff overlapping, cut off, buttons moving out of nonresizable windows, etc. ...

geoffthemedio:
That there are problems, particularly with extreme font sizes, doesn't mean layout issues can otherwise be ignored.

The hard-coded linebreaks definitely have problems. Notably, the species descriptions aren't just used in tooltips on the game setup screen, but also in the pedia, where the linebreaks look a bit odd...
species_pedia_article

geoffthemedio:
• works in-game, but is hard to get to line up nicely, due to variable-width font layout

maybe one of ◊!~»§

st-pa-fo:
there exist unicode characters with conditional or allowed breaks which are only applied when applyable, i'll try if the used font(engine) supports those, and i like the »

geoffthemedio:
> i'll try if the used font(engine) supports those

I expect it won't

st-pa-fo:
right, [they](https://www.fileformat.info/info/unicode/char/0082/index.htm) come through as spaces

st-pa-fo:
also i noticed crazy jumping text bits when you try for example to mark a range of text containing tabs by holding mouse pressed, and alas the ≫ doesnt align either. how about just o or x or back to nothing as it was up to now?

george-with-o
human-with-o's

st-pa-fo:
double • aligns perfectly with + and – but i think it looks not as good as the plain o

double-dot

geoffthemedio:
"problem" with `o` is that it's not vertically centred like + or - so seems a bit misaligned. Hence the `◊` suggestion.

st-pa-fo:
nah, the rhombus is als a bit more leftish than +- and too high for my taste. i'd say take a pick please
a) o
b) •• double dot
c)

(c is back to as it was without extra neutral symbol)

geoffthemedio:
I don't like `o` but otherwise don't have a strong opinion. Maybe post on the forums?

st-pa-fo:
so we can at least agree on the double dot? i'll put that in the commit then, and i'll ask on the forum. we can search/replace that thing again later if need be.

st-pa-fo:
picked, squashed, fixed and rebased it, force pushed, et voilá .. although i must admit that previously i had not been that big a fan of rewriting history.
please note, i've picked up on the lowest hanging fruit tickets i could find and dabbled at fixing wish-issues

UI/EncyclopediaDetailPanel.cpp
```diff
@@ -260,7 +260,7 @@ namespace {
         }
         else if (dir_name == "ENC_BUILDING_TYPE") {
             for (const auto& [building_name, building_type] : GetBuildingTypeManager()) {
-                if (!exclude_custom_categories_from_dir_name || DetermineCustomCategory(building_type->Tags()).empty()) {
+                if (!boost::starts_with(building_name,"BLD_COL_") && (!exclude_custom_categories_from_dir_name || DetermineCustomCategory(building_type->Tags()).empty())) {
```
geoffthemedio: this is adding some special case logic that depends on particular naming of content in scripts, so doesn't fit in this pull request and may not be acceptable at all (vs. specific-purpose tags or somesuch).

st-pa-fo: understood.

st-pa-fo: i have reverted that and made a tidy commit message. i promise to read up on tags and learn from older commits before trying something like that again.

st-pa-fo:
i do believe that failing test phase 'ubuntu build' has nothing to do with my changes, looks like some network problems to me. and the cancelled tests … no idea, might be related to the force-pushing i did?

geoffthemedio:
Tests are in some cases cancelled automatically if another in the same group fails.

I've restrated the group's tests.

Update: The [failing tests](https://github.com/freeorion/freeorion/runs/4747545106?check_suite_focus=true#step:13:8744) are content checksum tests that @o01eg added to check the Python parser. Since you are changing the content scripts, it's not surprising that the checksums also change:
````
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_NAMED_VALUEREF expected 42021 was 8272309
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_TECH expected 9095847 was 9161042
````

To fix that, you'd need to update the expected values of the checksums: https://github.com/freeorion/freeorion/blob/master/test/system/TestChecksum.cpp#L35-L36

**Comment by st-pa-fo** [comment]
not sure how to do that right, in what order?
1. pullrequest with content changed (see abobve)
2. test failed -> copy new expected checksum
3. clone the test repo
4. insert new checksum in test repo
5. pull request test repo with link to original PR from 1.?
sorry i'm still new to this ..

**Comment by st-pa-fo** [comment]
sorry, i got it now. it just did not show up in the xcode project . is all there ... just a sec

**Comment by geoffthemedio** [comment]
You can run the test locally and observe the results. Not sure exactly how, though.

You can also push a branch, make a pull request, note the failed test, commit a change to the test to the branch, and re-push it to re-run the updated test.

File: UI/EncyclopediaDetailPanel.cpp (hunk 1 above)
Comment (st-pa-fo) [reply UI/EncyclopediaDetailPanel.cpp]: seems to me that at some other places in the code something similar is done, e.g. see screenshots attached.
スクリーンショット 2022-01-08 18 00 49
スクリーンショット 2022-01-08 17 58 00
and yes, i know, just because others did wrong does not allow me to do wrong, am just saying, is all .. am after all still quite noobish around here ;)

Comment (geoffthemedio) [reply UI/EncyclopediaDetailPanel.cpp]: that's python, not the core C++ engine... different rules or guidelines apply.

**Comment by st-pa-fo** [comment]
whoopsie, left my work mail address in there by mistake, changed it now, nevermind no need to rewrite commit history, was not committing during work hours ^^

File: default/stringtables/en.txt [hunk 2]
```diff
@@ -6646,27 +6646,37 @@ GREETINGS_INTRO
 Greetings Esteemed Entity, and Welcome to the FreeOrion Universe!
 
 GREETINGS_GUIDE_REFERENCE
-Newcomers to this universe may particularly wish to consult the [[encyclopedia QUICK_START_GUIDE_TITLE]].
+To help you understand how these three areas of government - ships, planets, and research - are managed in your empire, please have a look at the [[encyclopedia QUICK_START_GUIDE_TITLE]] which we have prepared for you.
```
geoffthemedio: If you really want to keep this general change, it needs substantial rewriting. Avoid second person reference to the player ("you", "your") or first person ("our") in general, and remove all the overly specific story info like "elected Emperor" and "new-founded". Basically, in-game text does not address the player as a person, but rather states information.

st-pa-fo: acknowledged. will comply. also still thinking about those darned trenchers and their missing description ...

st-pa-fo:
again, there are other places, where you can find 'you', that were there before.
preexisting-you'ses
i will not try to correct those unless told to do so

geoffthemedio:
in the interest of getting this merged reasonably fast, please:
-put translations into a separate pull request
-put font file changes in a separate pull request
-remove the extra linebreaks, as they [don't work in the pedia](https://github.com/freeorion/freeorion/pull/3646#issuecomment-1007260927).
-move the QUICK_START_GUIDE_TEXT stuff to a separate pull request

st-pa-fo:
i've checked the strings file locally. no idea, why the test fails again. it showed more useful output when failing earlier so i suppose … i do not know what to suppose to be honest

what-now-with-the-string-test-again

st-pa-fo:
i spent >15 minutes looking at this, but no idea why this failed. now i'm calling it a night.
usr-bin-sg exit code 2

geoffthemedio:
The Snapcraft build is failing for a bunch of branches, including master, so I think it's an issue with that build, not the code or any specific branch.

geoffthemedio:
From the stingtable check:
````
default/stringtables/en.txt:BLD_COL_EAXAW: Unbalanced brackets: 'Colony of [[SP_EAXAW]'
default/stringtables/en.txt:BLD_COL_EGASSEM: Unbalanced brackets: 'Colony of [[SP_EGASSEM]'
default/stringtables/en.txt:BLD_COL_ETTY: Unbalanced brackets: 'Colony of [[SP_ETTY]'
````

File: default/stringtables/en.txt [hunk 3]
```diff
@@ -14619,187 +14631,187 @@ BLD_SCRYING_SPHERE_DESC
...
 BLD_COL_DERTHREAN
-Derthrean Colony
+Colony of [[SP_DERTHREAN]]
 
 BLD_COL_DERTHREAN_DESC
 [[BLD_COL_PART_1]] [[species SP_DERTHREAN]] colony. [[BLD_COL_PART_2]] Derthrean colony [[BLD_COL_PART_3]].
 
 BLD_COL_EAXAW
-Eaxaw Colony
+Colony of [[SP_EAXAW]
```
geoffthemedio: !

Cjkjvfnby:
> i've checked the strings file locally. no idea, why the test fails again. it showed more useful output when failing earlier so i suppose … i do not know what to suppose to be honest

I hope this is solved.  Looks like the format needs an option to do formatting in place. This will make things a bit simpler.

default/stringtables/en.txt
st-pa-fo: triple whoopsies … well i ran the check before catching the overlooking and what can i say except that i'll compress those commits again

st-pa-fo:
> Looks like the format needs an option to do formatting in place. This will make things a bit simpler.

is not really a problem for me, if that really would bug me i'd file an issue report …

geoffthemedio:
> Looks like the format needs an option to do formatting in place.

Not sure what you mean by this...

st-pa-fo:
> > Looks like the format needs an option to do formatting in place.
> 
> Not sure what you mean by this...

probably because i showed a screenshot where i pipe the output of "format en.txt" to en2.txt because piping it into itself breaks the format function of st-tool.py as it as at the moment, but as i said, i can live with that

st-pa-fo:
i only merged because it showed up as conflict. i know, the checksums will have to be adjusted again after a merge into master.

geoffthemedio:
A pull request won't be accepted if it merges master into the branch. It needs to be rebased on master. There is also still a merge conflict...

st-pa-fo:
> A pull request won't be accepted if it merges master into the branch. It needs to be rebased on master. There is also still a merge conflict...

yes. i justed pressed some button here on the page and it did that. and then i reverted that and force pushed to pre-merge. now i rebased. but the checksums probably fail again and i know i'd been told how to compute them locally before but forgot where and anyway the next failling test will probably show the expected checksum and i'll copy that

geoffthemedio:
Looks like you also combined the changes into a single commit with just "- rebased on master" as the commit message. A commit message that actually describes what the commit changes would be a lot better...

PR #2055, opened 2018-03-16
Census Report

st-pa-fo:
Census Report: new column with the number of settled world per species / per characteristics plus two new footer lines summarizing that information

Census Rows: Population values > 1000 were displayed as 1.234k before and are now displayed as 1234 without the k so as to better align with values < 1000

see also http://freeorion.org/forum/viewtopic.php?f=26&t=10907&p=91901#p91901

*Edit by @Vezzra:* Supersedes #2049.

PR #3648, opened 2022-01-08
adding number of settled worlds to census report

st-pa-fo:
for now without column labels. i'll post maybe another discussion on the forum about how necessary those would be, since i think it should be clear from the early game phase, when you see those numbers rising, what they meen. see screenshots attached.

seems pretty obvious, like the label below the table says, especially with the one world in the beginning and then there's that ten and one can see quite clearly that the 9 and 1 add up to that … so, yeah. here it is again, pretty much as it was back in https://github.com/freeorion/freeorion/pull/2055

### DIFF
--- a/UI/CensusBrowseWnd.cpp
+++ b/UI/CensusBrowseWnd.cpp
@@ -23,25 +23,25 @@ namespace {
     constexpr GG::Y ICON_BROWSE_ICON_HEIGHT(64);
 
     const GG::X BrowseTextWidth()
-    { return GG::X(FontBasedUpscale(200)); }
+    { return GG::X(FontBasedUpscale(230)); }
 }
 
 class CensusRowPanel : public GG::Control {
 public:
-    CensusRowPanel(GG::X w, GG::Y h, const std::string& name, double census_val, bool show_icon) :
+    CensusRowPanel(GG::X w, GG::Y h, const std::string& name, double census_val, int worlds, bool show_icon) :
         GG::Control(GG::X0, GG::Y0, w, h, GG::NO_WND_FLAGS),
-        m_show_icon(show_icon)
+        m_show_icon(show_icon),
+        m_worlds()
     {
         if (m_show_icon)
             m_icon = GG::Wnd::Create<GG::StaticGraphic>(
                 ClientUI::SpeciesIcon(name), GG::GRAPHIC_FITGRAPHIC);
 
         m_name = GG::Wnd::Create<CUILabel>(UserString(name), GG::FORMAT_RIGHT);
-
-        int num_digits = census_val < 10 ? 1 : 2; // this allows the decimal point to line up when there number above and below 10.
-        num_digits =    census_val < 100 ? num_digits : 3; // this allows the decimal point to line up when there number above and below 100.
-        num_digits =   census_val < 1000 ? num_digits : 4; // this allows the decimal point to line up when there number above and below 1000.
-        m_census_val = GG::Wnd::Create<CUILabel>(DoubleToString(census_val, num_digits, false), GG::FORMAT_RIGHT);
+        std::stringstream ss;
+        ss << std::fixed << std::setprecision(1) << census_val;
+        m_census_val = GG::Wnd::Create<CUILabel>(ss.str(), GG::FORMAT_RIGHT);
+        m_worlds = GG::Wnd::Create<CUILabel>(std::to_string(worlds), GG::FORMAT_RIGHT);
     }
 
     void CompleteConstruction() override {
@@ -54,6 +54,7 @@ public:
 
         AttachChild(m_name);
         AttachChild(m_census_val);
+        AttachChild(m_worlds);
 
         DoLayout();
     }
@@ -83,6 +84,7 @@ private:
     void DoLayout() {
         const GG::X SPECIES_NAME_WIDTH(ClientUI::Pts() * 9);
         const GG::X SPECIES_CENSUS_WIDTH(ClientUI::Pts() * 5);
+        const GG::X SPECIES_WORLDS_WIDTH(ClientUI::Pts() * 3);
 
         GG::X left(GG::X0);
         GG::Y bottom(MeterIconSize().y - GG::Y(EDGE_PAD));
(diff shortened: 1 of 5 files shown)


st-pa-fo:
… and the string test still failing at the one newline that should be two, i suppose.

st-pa-fo:
damn it, still needs a final touch to reconsile those numbers ...

108-or-109-or-111-what-the-heck

st-pa-fo:
it was the usual FO thing with current vs. target, fixed it, pushing soon.

UI/CensusBrowseWnd.cpp
```diff
@@ -23,25 +23,25 @@ namespace {
...
 }
 
 class CensusRowPanel : public GG::Control {
 public:
-    CensusRowPanel(GG::X w, GG::Y h, const std::string& name, double census_val, bool show_icon) :
+    CensusRowPanel(GG::X w, GG::Y h, const std::string& name, double census_val, int worlds, bool show_icon) :
         GG::Control(GG::X0, GG::Y0, w, h, GG::NO_WND_FLAGS),
-        m_show_icon(show_icon)
+        m_show_icon(show_icon),
+        m_worlds(nullptr)
```
geoffthemedio: `shared_ptr` default initializes to `nullptr` or empty, so this is unnecessary

UI/CensusBrowseWnd.cpp
```diff
@@ -216,6 +238,31 @@ void CensusBrowseWnd::CompleteConstruction() {
 
     m_tags_list->Resize(GG::Pt(BrowseTextWidth(), top2 -top -ROW_HEIGHT - HALF_HEIGHT + (EDGE_PAD*3)));
 
+    top2 += 0.5 * ROW_HEIGHT;
+
```
geoffthemedio: in the editor I see 4 spaces here, though I can't tell if they are here in the web view. if so, please make it a blank line instead

geoffthemedio:
otherwise seems fine

UI/CensusBrowseWnd.cpp
st-pa-fo: done

UI/CensusBrowseWnd.cpp
st-pa-fo: and done

geoffthemedio:
please rebase on master and force push. merging master into your branch isn't acceptable.

UI/CensusBrowseWnd.cpp
st-pa-fo: they were not visible here in the web view for me, i made it a blank line anyways.

UI/CensusBrowseWnd.cpp
st-pa-fo: i was trying to add something along the lines of 'Please do not explicitly initialize pointers to nullptr since this is redundent' under #Pointers at https://www.freeorion.org/index.php?title=Code_Standards&action=edit but i do not have permission to do that
schirmschuß

UI/CensusBrowseWnd.cpp
```diff
@@ -23,25 +23,25 @@ namespace {
...
 }
 
 class CensusRowPanel : public GG::Control {
 public:
-    CensusRowPanel(GG::X w, GG::Y h, const std::string& name, double census_val, bool show_icon) :
+    CensusRowPanel(GG::X w, GG::Y h, const std::string& name, double census_val, int worlds, bool show_icon) :
         GG::Control(GG::X0, GG::Y0, w, h, GG::NO_WND_FLAGS),
-        m_show_icon(show_icon)
+        m_show_icon(show_icon),
+        m_worlds()
```
geoffthemedio: also unnecessary to explicitly initialize something that can be default initialized

UI/MapWnd.cpp
```diff
@@ -6939,17 +6939,20 @@ void MapWnd::RefreshIndustryResourceIndicator() {
...
 
 void MapWnd::RefreshPopulationIndicator() {
+    float target_population = 0.0;
     Empire* empire = GetEmpire(GGHumanClientApp::GetApp()->EmpireID());
     if (!empire) {
         m_population->SetValue(0.0);
         return;
-    }
-    m_population->SetValue(empire->GetPopulationPool().Population());
+    } else target_population = empire->GetPopulationPool().Population();
```
geoffthemedio:
```suggestion
    } else {
        target_population = empire->GetPopulationPool().Population();
    }
```

UI/CensusBrowseWnd.cpp
geoffthemedio:
```suggestion
        m_show_icon(show_icon)
```

UI/CensusBrowseWnd.h
```diff
@@ -8,11 +8,14 @@
...
-                    std::vector<std::string> census_order);
-
+    CensusBrowseWnd(std::string                  title_text
+                   ,float                        target_population // take as parameter so as not to have to import empire etc.
+                   ,std::map<std::string, float> population_counts
+                   ,std::map<std::string, int>   population_worlds
+                   ,std::map<std::string, float> tag_counts
+                   ,std::map<std::string, int>   tag_worlds
+                   ,std::vector<std::string>     census_order
+                   );
```
geoffthemedio:
```suggestion
    CensusBrowseWnd(std::string                  title_text,
                   float                        target_population, // take as parameter so as not to have to import empire etc.
                   std::map<std::string, float> population_counts,
                   std::map<std::string, int>   population_worlds,
                   std::map<std::string, float> tag_counts,
                   std::map<std::string, int>   tag_worlds,
                   std::vector<std::string>     census_order);
```

UI/MapWnd.cpp
```diff
@@ -6939,17 +6939,20 @@ void MapWnd::RefreshIndustryResourceIndicator() {
 }
 
 void MapWnd::RefreshPopulationIndicator() {
+    float target_population = 0.0;
```
geoffthemedio:
```suggestion
    float target_population = 0.0f;
```

UI/CensusBrowseWnd.h
geoffthemedio: more consistent with other code formatting

UI/CensusBrowseWnd.cpp
geoffthemedio: I'd rather add a coding standards document to the git repo than worry about the wiki...

PR #3651, opened 2022-01-09
code folding for texts and translations

st-pa-fo:
i realize, there might be some initial merging pains. but bear with me, i propose that code foldability has its long term merrits as a potential time saver when working with those files. see attached screenshot. and i might need to get into contact with the localizers about this one.
search replace+regex=folding

### DIFF
--- a/check/st-tool.py
+++ b/check/st-tool.py
@@ -155,13 +155,14 @@ class StringTable(object):
             elif isinstance(entry, list):
                 result += "\n"
                 result += "\n"
-                result += "##\n"
+                result += "## }\n"
                 for line in entry:
                     if line:
-                        result += "## " + line + "\n"
+                        if (line!="}" and line!="{"):
+                            result += "## " + line + "\n"
                     else:
                         result += "##\n"
-                result += "##\n"
+                result += "## {\n"
 
         if self.includes:
             result += "\n"
@@ -1053,6 +1054,11 @@ if __name__ == "__main__":
         * Section titles are prefixed with a block comment (prefix '##').
         * Section titles have two leading whitespaces and one trailing
           whitespace.
+        * Additional parentheses are inserted in the block comments like
+          such:  ## }
+                 ## Comment...
+                 ## {
+          in order to facilitate code folding.
         """
         ),
         formatter_class=argparse.RawTextHelpFormatter,
--- a/check/st-tool.py
+++ b/check/st-tool.py
@@ -158,7 +158,7 @@ class StringTable(object):
                 result += "## }\n"
                 for line in entry:
                     if line:
-                        if (line!="}" and line!="{"):
+                        if (line != "}" and line != "{"):
                             result += "## " + line + "\n"
                     else:
                         result += "##\n"
--- a/default/stringtables/cz.txt
+++ b/default/stringtables/cz.txt
@@ -20,7 +20,7 @@ Czech
 
 ##
 ## Common phrases
-##
+## {
 
 CANCEL
 Zpět
@@ -81,14 +81,14 @@ DEFAULT_PLAYER_NAME
 Lidský_Hráč
 
 
-##
+## }
 ## Major errors
-##
+## {
 
 
-##
+## }
 ## Predefined Ship Designs (located in default/scripting/ship_designs/)
-##
+## {
 
 SD_SCOUT
 Průzkumník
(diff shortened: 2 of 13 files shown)


st-pa-fo:
i would have to redo that commit anyway, because i just noticed that i 'managed' to change all tabs to spaces by mistake .. whoopsie, but other than that, i still think it's a good idea

Cjkjvfnby:
@st-pa-fo , for the bulk changes, let's use a single file.   It's will be much easier to do experiments with it.  

st-tool has UI mode.  Maybe you could use it instead of folding. 
```
./check/st-tool.py edit default/stringtables/en.txt default/stringtables/ru.txt
```

PS. Do not forget to click `save` at the top of the screen.

st-pa-fo:
"Run black . --check -q
Error: Process completed with exit code 1."
no idea what that meens. i'll try the interactive mode. i do not understand your comment about one big file, can you please clarify

st-pa-fo:
wheee that is some nice interactive mode indeed. although alas not fully functional for me, i get some
"urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1122)>"

and i think i know what you ment now, that i should not apply the changes to all of the txt files at once but pick just one, while we are still experimenting

st-pa-fo:
@Cjkjvfnby yeah well when working purely as translator, that interactive mode seems great (provided is fully functional), on the other hand there is a lot of programming-related stuff in there, like the parameters and the markup and stuff, and when adding new keys for new text due to new features, or when generally maintaining that stuff, it is also useful to have the code folding as a programmer who needs to work with those same files. question is, will the interactive mode still work for you (i suppose you have your ssl certificate thingy set up properly so the tool works for you=)

Cjkjvfnby:
> and i think i know what you ment now, that i should not apply the changes to all of the txt files at once but pick just one, while we are still experimenting

It saves a lot of time.  
For the translation, this trick works as well.  Each time you sit for translation start a new branch, and make a PR when you are about to stand up.  10 PR with 10 translations each will be merged faster when one with 100.

> i suppose you have your ssl certificate thingy set up properly so the tool works for you=

It's a `localhost`, nobody does a certificate for a `localhost`.  Could you check that you use `http` for that?  Please submit an issue, I will check it.

Cjkjvfnby:
The best thing to know/remember how things work is to read the documentation.  If there is no documentation just write it. @st-pa-fo feel free to join, you have a fresh view, so you could provide insight about things that are too obvious for people who get used to that things.  https://github.com/freeorion/freeorion/pull/3652

st-pa-fo:
> nobody does a certificate for a localhost

i'd-try-and-get-it-to-work-for-me-but

nevermind, i can imagine it works great when it works. although me as translator would had preferred smth. like Ωt with more translator-support stuff like built in style checking and translation memories and fuzzy matching the okapi automation stuff and so on, but i'm not here as a translator right now :)

seems though, the st-tool.py's edit would not complain about some extra parentheses, so applying that search/replace from the initial screenshot in this PR would not obstruct your normal workflow, right?
parentheses-seem-to-be-okay

st-pa-fo:
working with all the new species, i could really use the code folding about now over there in the other pullrequests .. so please, tell me who else to notify for approval or are we gonna do this only in english and let each translator decide for themselves if they want to make the programmers'es work that tiny little bit more easier when working with those giant text files? @Cjkjvfnby mentioning you by name because i still am not sure how the whole notification thing works here on this platform ..

Cjkjvfnby:
The certificate problem is not related to localhost.  The server tries to download the file from https://code.jquery.com/jquery-3.4.1.min.js.  This link looks fine, you could check this at the https://dnschecker.org/ssl-certificate-examination.php

`Self-signed certificate in the chain` might be related to your network. Like your internet provider just catch all your traffic, decrypt it, and encrypt with it's own certificate.  In that case, I expect the same problem with all other HTTP services, or you just have installed the certificate to your browser, but the server code ignores it.

Cjkjvfnby:
Code foldability might be good, but I'd prefer splitting it to separate files. I am not working with translation, I just keep an eye on st-tool.   The only rule: do not touch the `fr.txt`. 


Which tool do you use?  

My text editor is Sublme Text.  It works fast on big files, has search and local folding.
So you could just fold something exactly for file on disc. https://wesbos.com/sublime-text-code-folding

It does not support curly braces, for automatic folding.  Adding some syntax that works only with a specific text editor, does not look good for me. 

Need to think about how we could deal with it.

st-pa-fo:
> Adding some syntax that works only with a specific text editor, does not look good for me.

agreed indeed it does not. am using xcode, it has the blame and the diff and the autocomplete and the folding and the multiple-cursor-simultanious-editing thing and the bells and the whistles and is for free for whoever works on macs .. and it wouldnt mind any other kind of brackets braces or parentheses and i guess i can live without the folding, just thought might be neat and maybe save some cursing and yelling at the editor ...

PR #3652, opened 2022-01-09
Add README to st-tool

Cjkjvfnby:
I'd like to collect all info about string tables in the repository.  

This is the first step. 

The following steps are (as separate PRs):
- Move content of the https://www.freeorion.org/forum/viewtopic.php?f=27&t=9275
- Add some description about the translation processes.  
- Put references to this file from other places (forum, other readmes, etc).

For the review I expect: logic check, I might misunderstand/forget something, spelling check, and formatting check.  For formatting, I'd recommend opening the file in a separate tab when it is rendered by GitHub https://github.com/Cjkjvfnby/freeorion/tree/sa/refactor-st-tool/check.

@geoffthemedio , could you add https://github.com/st-pa-fo as a collaborator to the repo, so I could use mention?

### DIFF
--- a/check/README.md
+++ b/check/README.md
@@ -5,7 +5,7 @@
 - `string table` a file with translation
 - `reference table` en.txt string table.
   English is the development language, so any changes related to text should be in that file.
-  Also, this table us used as fallback for other translations, if a needed key is missing.
+  Also, this table is used as a fallback for other translations, if a key is missing.
   This file is mostly maintained by people who change game code,
   so the list of entries in this file should be the most up-to-date.
 - `entry` one piece of translation, it consists of
@@ -100,7 +100,7 @@ Supported types
 [[buildingtype BLD_SHIPYARD_ORBITAL_DRYDOCK]]
 ```
 
-- `value` inserts value from the script to the text,
+- `value` inserts value from content scripts into the text,
   eg. `[[value FLD_NANITE_SWARM_STRUCTURE_FLAT]]` will look up the value from
   the named valueref scripted as
   `NamedReal name = "FLD_NANITE_SWARM_STRUCTURE_FLAT" value = 5 * [[SHIP_STRUCTURE_FACTOR]])`
--- a/check/README.md
+++ b/check/README.md
@@ -7,20 +7,20 @@
   English is the development language, so any changes related to text should be in that file.
   Also, this table us used as fallback for other translations, if a needed key is missing.
   This file is mostly maintained by people who change game code,
-  so the list of entries in this file should be the most actual.
+  so the list of entries in this file should be the most up-to-date.
 - `entry` one piece of translation, it consists of
   - `key` a sting that represents entry in the codebase (C++, Python).
     This string must be unique.
   - `value` actual translation that will be shown to the user.
 
 ### Key
-Only English chars in uppercase and dash are allowed.
+Only uppercase Latin alphabet (A-Z) and underscore (_) are allowed.
 Renaming a key should be done with caution.
 Many of the keys in the stringtable are used in the game source code explicitly,
 or in content scripts.
 Some keys are also generated programatically, such as by adding a prefix to other text.
-In C++ `UserString` and `UserStringNop` are used to get keys values,
-In Python `fo.userString` and `fo.userStringList` are used.
+C++ uses UserString(...) or UserStringNop(...) to look up strings or to mark stringtable keys that are used in code.
+Python uses `fo.userString` and `fo.userStringList` for look up.
 
 ### Value syntax
 #### Simple (single line)
@@ -80,7 +80,7 @@ Hello world
 
 This feature should be used to keep the translation in a consistent state.
 You define term once and use it everywhere.
-For most common used things like tech, building, etc. we have more advanced type of reference: link
+For various kinds of game content, like techs, buildings, or species, etc., there are links
 
 - `link` inserts a link to a wiki page about some scripted content or a ship design in the game universe.
 Supported types
@@ -111,4 +111,4 @@ Supported types
 ## Maintain string tables
 Formatting and basic checks are maintained by tool: `st-tool.py`.
 See `.check/st-tool.py format --help` and `.check/st-tool.py check --help`
-This two commands are part of GitHub automatic checks.
+These two commands are part of GitHub automatic checks.
--- a/check/README.md
+++ b/check/README.md
@@ -19,8 +19,8 @@ Renaming a key should be done with caution.
 Many of the keys in the stringtable are used in the game source code explicitly,
 or in content scripts.
 Some keys are also generated programatically, such as by adding a prefix to other text.
-C++ use `UserString` and `UserStringNop` to retrieve keys,
-Python `fo.userString`, and `fo.userStringList`.
+In C++ `UserString` and `UserStringNop` are used to get keys values,
+In Python `fo.userString` and `fo.userStringList` are used.
 
 ### Value syntax
 #### Simple (single line)
--- a/check/README.md
+++ b/check/README.md
@@ -60,10 +60,9 @@ world
 ```
 
 #### Tags
-You can customize text with special markup tags. Not full list of tags:
+You can customize text with special markup tags. Not a full list of tags:
 
-- `[[]]` inserts text from another entry.
-  You add a key in double square braces and user will see the value of that key.
+- `[[KEY]]` will be replaced with the text in the entry with key `KEY`.
   Substitution is recursive, if text has another substitution it will be resolved too.
   If key was not found in this table, it will be looked in the reference table. 
 
@@ -108,6 +107,35 @@ Supported types
   which might be just a constant in some cases,
   but in this case will depend on the value of a game rule.
 
+### Comments
+
+Comments can be added to stringtables by starting a line with `#`.
+Comment lines are ignored when processing the stringtable. 
+Comments are often written immediatley before a key line, 
+and may contain information about parameters that will be substituted by 
+the game engine into that string at runtime.
+
+```text
+# %1% amount of research points required to research the technology.
+# %2% number of turns required to research the technology.
+TECH_TOTAL_COST_ALT_STR
+%1% RP / %2% Turns
+```
+
+Groups of related stringtable entries are often preceeded 
+by a section marker comment, which is formatted with two initial blank lines,
+and then three lines starting with `##`, with the middle ## line having a section title, eg.
+
+```text
+
+
+##
+## TechTreeWnd
+##
+
+TECH_DISPLAY
+Display
+```
 
 ## Maintain string tables
 Formatting and basic checks are maintained by tool: `st-tool.py`.
--- a/check/README.md
+++ b/check/README.md
@@ -37,7 +37,7 @@ value
 
 #### Full (single or multiline)
 Start and end with triple single quotes.
-You can put anything except triple singe quotes inside.
+You can put anything except triple single quotes inside.
 
 ```
 KEY_WITH_LEADING_SPACE
@@ -61,8 +61,11 @@ world
 
 #### Tags
 You can customize text with special markup tags. Not full list of tags:
-- `reference` special syntax that allows to put value from another entry.
+
+- `[[]]` inserts text from another entry.
   You add a key in double square braces and user will see the value of that key.
+  Substitution is recursive, if text has another substitution it will be resolved too.
+
 ```
 KEY_WORLD
 world
@@ -77,10 +80,9 @@ Hello world
 
 This feature should be used to keep the translation in a consistent state.
 You define term once and use it everywhere.
-For most common used things like tech, building, etc we have more advanced type of reference: link
+For most common used things like tech, building, etc. we have more advanced type of reference: link
 
-
-- `link` special syntax that will be rendered as a link to a wiki page about some scripted content or a ship design in the game universe.
+- `link` inserts a link to a wiki page about some scripted content or a ship design in the game universe.
 Supported types
   - encyclopedia
   - buildingtype
@@ -93,7 +95,17 @@ Supported types
   - species
   - tech
   - policy
-  - value
+```
+[[buildingtype BLD_SHIPYARD_ORBITAL_DRYDOCK]]
+```
+
+- `value` inserts value from the script to the text,
+  eg. `[[value FLD_NANITE_SWARM_STRUCTURE_FLAT]]` will look up the value from
+  the named valueref scripted as
+  `NamedReal name = "FLD_NANITE_SWARM_STRUCTURE_FLAT" value = 5 * [[SHIP_STRUCTURE_FACTOR]])`
+  and be replaced by whatever `5 * [[SHIP_STRUCTURE_FACTOR]]` equals,
+  which might be just a constant in some cases,
+  but in this case will depend on the value of a game rule.
 
 
 ## Maintain string tables
--- a/check/README.md
+++ b/check/README.md
@@ -9,7 +9,8 @@
   This file is mostly maintained by people who change game code,
   so the list of entries in this file should be the most up-to-date.
 - `entry` one piece of translation, it consists of
-  - `key` a sting that represents entry in the codebase (C++, Python).
+  - `key` a sting that represents entry. It could be referenced from code: C++, Python, FOCS.
+     Also, new key could be added just for substitution purposes to the string table.
     This string must be unique.
   - `value` actual translation that will be shown to the user.
 
@@ -58,13 +59,13 @@ world
 
 ```
 
-
 #### Tags
 You can customize text with special markup tags. Not full list of tags:
 
 - `[[]]` inserts text from another entry.
   You add a key in double square braces and user will see the value of that key.
   Substitution is recursive, if text has another substitution it will be resolved too.
+  If key was not found in this table, it will be looked in the reference table. 
 
 ```
 KEY_WORLD
--- a/check/README.md
+++ b/check/README.md
@@ -1,0 +1,99 @@
+# Tool for work with string tables
+
+## Glossary
+
+- `string table` a file with translation
+- `reference table` en.txt string table.
+  English is the main game language, so any changes related to text should be in that file.
+  Also this table us used as fallback for other translation.
+  This file is mostly maintained by people who change game code,
+  so the list of entries in this file should be the most actual.
+- `entry` one piece of translation, it consists of
+  - `key` a sting that represents entry in the codebase (C++, Python).
+    This string must be unique.
+  - `value` actual translation that will be shown to the user.
+
+
+### Key
+Only English chars in uppercase and dash are allowed.
+Rename of a key should be done with caution. Usually key used as is in code,
+but sometimes it could be used as prefix and code logic based on name could be present.
+
+### Value syntax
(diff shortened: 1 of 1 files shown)


geoffthemedio:
> @geoffthemedio , could you add https://github.com/st-pa-fo as a collaborator to the repo, so I could use mention?

maybe done / invited to "contributors"

Ouaz:
I don't undertstand exactly what it's all about ^^, but as long as the localized stringtables (i.e. FR for me) keep the same number of lines than the EN file, it's ok for me (I need that to sync new translation entries).

st-pa-fo:
@ouaz please have a look at https://github.com/freeorion/freeorion/pull/3651

Cjkjvfnby:
> I don't undertstand exactly what it's all about ^^, but as long as the localized stringtables (i.e. FR for me) keep the same number of lines than the EN file, it's ok for me (I need that to sync new translation entries).

I want to document the translation process, and you one of the active contributers to the translation.  So if I'll miss something in the doc, you can help me to catch it.

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+  - fieldtype
+  - metertype
+  - predefinedshipdesign
+  - shiphull
+  - shippart
+  - special
+  - species
+  - tech
+  - policy
+  - value
```
geoffthemedio: `[[value NAMED_VALUEREF_LABEL]]` doesn't make a link. Rather, it inserts the value from the script into the text.

eg. `[[value FLD_NANITE_SWARM_STRUCTURE_FLAT]]` will look up the value from the named valueref scripted as `NamedReal name = "FLD_NANITE_SWARM_STRUCTURE_FLAT" value = 5 * [[SHIP_STRUCTURE_FACTOR]])` and be replaced by whatever `5 * [[SHIP_STRUCTURE_FACTOR]]` equals, which might be just a constant in some cases, but in this case will depend on the value of a game rule.

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+world
+<<<
+
+```
+
+
+#### Tags
+You can customize text with special markup tags. Not full list of tags:
+- `reference` special syntax that allows to put value from another entry.
+  You add a key in double square braces and user will see the value of that key.
```
geoffthemedio: This is misleading. `[[KEY]]` does a substitution, and if there is a stringtable entry with the key `KEY`, it will be simply replaced. But those substitutions are recursive, so if there is a stringtable entry
````
KEY
Some text [[ANOTHER_KEY]] and perhaps a [[tech LRN_PSYONICS]]
````
Then the references and link generation in the stringtable entry with key `KEY` will be resolved first, then substituted into the place where `[[KEY]]` was written.

File: check/README.md [hunk 3]
```diff
@@ -0,0 +1,99 @@
...
+value
+
+>>>
+value
+<<<
+```
+
+#### Full (single or multiline)
+Start and end with triple single quotes.
+You can put anything except triple singe quotes inside.
```
geoffthemedio: singe -> single

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+KEY
+value
+
+>>>
+value
+<<<
+```
+
+#### Full (single or multiline)
+Start and end with triple single quotes.
```
geoffthemedio, line 40: maybe **triple** -> **three**

maybe **single quotes** -> **apostrophes**

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+Hello [[KEY_WORLD]]!
+
+>>>
+Hello world
+<<<
+```
+
+This feature should be used to keep the translation in a consistent state.
+You define term once and use it everywhere.
+For most common used things like tech, building, etc we have more advanced type of reference: link
```
geoffthemedio: **etc** -> **etc.**

geoffthemedio: For various kinds of game content, like techs, buildings, or species, there are links

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+  This file is mostly maintained by people who change game code,
+  so the list of entries in this file should be the most actual.
+- `entry` one piece of translation, it consists of
+  - `key` a sting that represents entry in the codebase (C++, Python).
+    This string must be unique.
+  - `value` actual translation that will be shown to the user.
+
+
+### Key
+Only English chars in uppercase and dash are allowed.
```
geoffthemedio: English chars in uppercase and dash -> uppercase Latin alphabet (**A-Z**) and underscore (**_**)

check/README.md
```diff
@@ -0,0 +1,99 @@
+# Tool for work with string tables
+
+## Glossary
+
+- `string table` a file with translation
+- `reference table` en.txt string table.
+  English is the main game language, so any changes related to text should be in that file.
+  Also this table us used as fallback for other translation.
+  This file is mostly maintained by people who change game code,
+  so the list of entries in this file should be the most actual.
```
geoffthemedio: actual -> current or up-to-date

check/README.md
```diff
@@ -0,0 +1,99 @@
+# Tool for work with string tables
+
+## Glossary
+
+- `string table` a file with translation
+- `reference table` en.txt string table.
+  English is the main game language, so any changes related to text should be in that file.
```
geoffthemedio: main game language -> development language

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+  Also this table us used as fallback for other translation.
```
geoffthemedio: Also -> Also,

other translation -> other translations, if a needed key is missing

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+  so the list of entries in this file should be the most actual.
+- `entry` one piece of translation, it consists of
+  - `key` a sting that represents entry in the codebase (C++, Python).
+    This string must be unique.
+  - `value` actual translation that will be shown to the user.
+
+
+### Key
+Only English chars in uppercase and dash are allowed.
+Rename of a key should be done with caution. Usually key used as is in code,
```
geoffthemedio: Rename of a key -> Renaming a key

geoffthemedio: Many of the keys in the stringtable are used in the game source code explicitly, or in content scripts. Some keys are also generated programatically, such as by adding a prefix to other text.

geoffthemedio: Could mention the `UserString("KEY")` or `UserStringNop("KEY")` expressions that appear in C++ code whenever a stringtable key is or could be looked up explicitly.

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+Hello world
+<<<
+```
+
+This feature should be used to keep the translation in a consistent state.
+You define term once and use it everywhere.
+For most common used things like tech, building, etc we have more advanced type of reference: link
+
+
+- `link` special syntax that will be rendered as a link to an object wiki page.
```
geoffthemedio: **to an object wiki page** -> **to a wiki page about some scripted content or a ship design in the game universe**

check/README.md
```diff
@@ -0,0 +1,99 @@
...
+  - species
+  - tech
+  - policy
+  - value
+
+
+## Maintain string tables
+Formatting and basic checks are maintained by tool: `st-tool.py`.
+See `.check/st-tool.py format --help` and `.check/st-tool.py check --help`
+This too commands are mandatory and part of our GitHub automatic checks.
```
geoffthemedio: don't know what **This too commands** means

Cjkjvfnby: too is two

check/README.md
```diff
@@ -4,20 +4,23 @@
...
-
 ### Key
 Only English chars in uppercase and dash are allowed.
-Rename of a key should be done with caution. Usually key used as is in code,
-but sometimes it could be used as prefix and code logic based on name could be present.
+Renaming a key should be done with caution.
+Many of the keys in the stringtable are used in the game source code explicitly,
+or in content scripts.
+Some keys are also generated programatically, such as by adding a prefix to other text.
+C++ use `UserString` and `UserStringNop` to retrieve keys,
```
geoffthemedio: C++ uses `UserString(...)` or `UserStringNop(...)` to look up strings or to mark stringtable keys that are used in code.

check/README.md
Cjkjvfnby: triple is often used in Python [documents](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).

> For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. For Unicode docstrings, use u"""Unicode triple-quoted strings""".

In the [Python documentation about string](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str) this is called `Triple quoted`, but I find it measleading.

check/README.md
```diff
@@ -96,4 +99,4 @@ Supported types
 ## Maintain string tables
 Formatting and basic checks are maintained by tool: `st-tool.py`.
 See `.check/st-tool.py format --help` and `.check/st-tool.py check --help`
-This too commands are mandatory and part of our GitHub automatic checks.
+This two commands are part of GitHub automatic checks.
```
geoffthemedio: These two

check/README.md
geoffthemedio: OK... I just found the phrase "triple single quoted" confusing...

check/README.md
```diff
@@ -19,8 +19,8 @@ Renaming a key should be done with caution.
 Many of the keys in the stringtable are used in the game source code explicitly,
 or in content scripts.
 Some keys are also generated programatically, such as by adding a prefix to other text.
-C++ use `UserString` and `UserStringNop` to retrieve keys,
-Python `fo.userString`, and `fo.userStringList`.
+In C++ `UserString` and `UserStringNop` are used to get keys values,
```
geoffthemedio: `UserStringNop` doesn't actually get a key's value... it just marks that a string is a key. It actually is a No-Op... does nothing and returns what it was passed, at the preprocessor stage.

Cjkjvfnby: So basically `UserStringNop("KEY")` is a special syntax to write "KEY".  The only purpose of it is readability?

I use that same idea for the AI code for ids. https://github.com/freeorion/freeorion/blob/master/default/python/common/fo_typing.py

geoffthemedio: The purpose is to be able to scan through the scource code and identify strings that are used that are expected to be present in the stringtable by the UI code. I don't know if anything actually does check this, but that's the idea.

check/README.md
```diff
@@ -7,20 +7,20 @@
   English is the development language, so any changes related to text should be in that file.
   Also, this table us used as fallback for other translations, if a needed key is missing.
   This file is mostly maintained by people who change game code,
-  so the list of entries in this file should be the most actual.
+  so the list of entries in this file should be the most up-to-date.
 - `entry` one piece of translation, it consists of
   - `key` a sting that represents entry in the codebase (C++, Python).
     This string must be unique.
   - `value` actual translation that will be shown to the user.
```
st-pa-fo, line 15: sometimes the comments contain instructions or parameter descriptions as well

check/README.md
```diff
@@ -0,0 +1,114 @@
...
+  - buildingtype
+  - fieldtype
+  - metertype
+  - predefinedshipdesign
+  - shiphull
+  - shippart
+  - special
+  - species
+  - tech
+  - policy
```
st-pa-fo, line 97: i'll add a new type 'auto' soon that will try to determine the appropriate matching content type from raw string, see also https://github.com/freeorion/freeorion/pull/3665/commits/88abdd6be3107669efe360a6b0fb6212308ec12c#diff-c6e7742666839efb5adf4f6a1fea0c22d9e28991d1f88c76e9062ccef54223f9

st-pa-fo:
> 
> @geoffthemedio , could you add https://github.com/st-pa-fo as a collaborator to the repo, so I could use mention?

oops i just noticed an email from too long ago to accept the invitation. can we try that again please? @geoffthemedio
invitation expired

check/README.md
st-pa-fo: @geoffthemedio, i am afraid it is not used consequently enough. there are still thousands of places where we need to put those in, in order for e.g. the japanese translation (which someone is working on right now) to look and feel japanese-ish, all the punctuation and spaces and things that are just programmatically done to and with strings ...

st-pa-fo: maybe can write a test script to check that each and every single string literal is marked as either `UserString(...)` or `UserStringNop(...)` but not let the build fail, just to get a list of warnings

check/README.md
st-pa-fo: well as i see it, it informs human readers of the source code, that _**yes, someone thought about this, and decided that, no this is not for internationalization, so we put a mark on it, so noone has to think about that one again**_ or yes maybe some outomated testing in the future to make sure that all string literals are marked as either to be internationalized or marked as to be excluded from internationalization. the latter should really be very few, things like internal enum values.

Vezzra:
@st-pa-fo, I checked the organization settings and found a "failed invitation", because you didn't reply to the invitation within 7 days. I then clicked "retry invitation", which made the invitation disappear completely. When I tried to make a new invitation for you, it said that you already are a member of the organization. I checked that, and yes, there you are, as a member of the "Contributors" team.

So apparently you've been succesfully added to the FO organzation.

check/README.md
```diff
@@ -0,0 +1,114 @@
...
+## Glossary
+
+- `string table` a file with translation
+- `reference table` en.txt string table.
+  English is the development language, so any changes related to text should be in that file.
+  Also, this table us used as fallback for other translations, if a needed key is missing.
+  This file is mostly maintained by people who change game code,
+  so the list of entries in this file should be the most up-to-date.
+- `entry` one piece of translation, it consists of
+  - `key` a sting that represents entry in the codebase (C++, Python).
```
geoffthemedio: stringtable keys aren't just stuff from C++ and Python. They can also be used in FOCS and can be added just for substitution purposes via `[[KEY_INSERT_MACROS]]` within other strings in the same stringtable, and should also look in the reference (en) stringtable if not found in the current stringtable.

To be clear, a translation can add its own keys that it uses for substitutions, which don't need to also appear in the english stringtable.

check/README.md
geoffthemedio: > in order for e.g. the japanese translation (which someone is working on right now) to look and feel japanese-ish, all the punctuation and spaces and things that are just programmatically done to and with strings ...

These are mostly unrelated issues, aren't they? Yes, it's impossible to do a proper looking translation by replacing substrings of english text with another language when there are incompatible built-in assumptions about grammar and formatting and such in the code. But how many of the strings that are used in the C++ code have `UserStringNop` wrappers is mostly independent of that.

geoffthemedio: > maybe can write a test script to check that each and every single string literal is marked as either `UserString(...)` or `UserStringNop(...)`

Not all string literals in the code are expected to have translations. The point is to flag or mark the ones that should, when they appear in the C++ code explicitly and exactly as there should be corresponding keys found in the stringtable.

There are actually numerous incorrect uses, like [here](https://github.com/freeorion/freeorion/blob/master/client/AI/AIClientApp.cpp#L54), where the key is not written explicitly in code but is being programatically generated. In those cases, wrapping with `UserStringNop` isn't necessary or useful, as there's no trivial way to determine from that C++ code what specific strings should be present.

It would probably be useful to have a test script that checks through all the C++ code for appearances of `UserString("...")` and `UserStringNop("...")` and makes a list, and then checks if the corresponding strings are present in the stringtables. That was the motivation for adding the `Nop` version of the wrapper, at least as far as I recall.

check/README.md
geoffthemedio: > yes, someone thought about this, and decided that, no this is not for internationalization, so we put a mark on it, so noone has to think about that one again

The opposite: Someone marked a string specifically as needing a user-readable stringtable entry. Ideally all strings that are written explicitly in code and that need a stringtable entry would be wrapped by `UserString` or `UserStringNop`.

If a string in C++ code doesn't need localization, then there is presently no way to mark it as such. And I don't think there should be, as there are many small internal-use string literals that don't need localization and marking them all would be highly intrusive and distracting throughout the code.

check/README.md
```diff
@@ -0,0 +1,115 @@
+# Tool for work with string tables
+
+## Glossary
+
+- `string table` a file with translation
+- `reference table` en.txt string table.
+  English is the development language, so any changes related to text should be in that file.
+  Also, this table us used as fallback for other translations, if a needed key is missing.
```
geoffthemedio:
```suggestion
  Also, this table is used as a fallback for other translations, if a key is missing.
```

check/README.md
```diff
@@ -0,0 +1,115 @@
...
+world
+<<<
+
+```
+
+#### Tags
+You can customize text with special markup tags. Not full list of tags:
+
+- `[[]]` inserts text from another entry.
+  You add a key in double square braces and user will see the value of that key.
```
geoffthemedio:
```suggestion
- `[[]]` inserts text from another entry.
  `[[KEY]] will be replaced with the text in the entry with key `KEY`.
```

check/README.md
```diff
@@ -0,0 +1,115 @@
...
+  - shippart
+  - special
+  - species
+  - tech
+  - policy
+```
+[[buildingtype BLD_SHIPYARD_ORBITAL_DRYDOCK]]
+```
+
+- `value` inserts value from the script to the text,
```
geoffthemedio:
```suggestion
- `value` inserts value from content scripts into the text,
```

check/README.md
```diff
@@ -0,0 +1,115 @@
...
+  eg. `[[value FLD_NANITE_SWARM_STRUCTURE_FLAT]]` will look up the value from
+  the named valueref scripted as
+  `NamedReal name = "FLD_NANITE_SWARM_STRUCTURE_FLAT" value = 5 * [[SHIP_STRUCTURE_FACTOR]])`
+  and be replaced by whatever `5 * [[SHIP_STRUCTURE_FACTOR]]` equals,
+  which might be just a constant in some cases,
+  but in this case will depend on the value of a game rule.
+
+
```
geoffthemedio, line 139: Suggestion that I can't format as a suggestion:

## Comments
Comments can be added to stringtables by starting a line with `#`. Comment lines are ignored when processing the stringtable. Comments are often written immediatley before a key line, and may contain information about parameters that will be substituted by the game engine into that string at runtime.
```
# %1% amount of research points required to research the technology.
# %2% number of turns required to research the technology.
TECH_TOTAL_COST_ALT_STR
%1% RP / %2% Turns
```

Groups of related stringtable entries are often preceeded by a section marker comment, which is formatted with two initial blank lines, and then three lines starting with `##`, with the middle `##` line having a section title, eg.
```


##
## TechTreeWnd
##

TECH_DISPLAY
Display
```

check/README.md
Cjkjvfnby: Is it possible to create a recursive cycle?   do we have a recursion limit?

geoffthemedio: There is [a check for cyclic recursion](https://github.com/freeorion/freeorion/blob/master/util/StringTable.cpp#L247) that will generate an error message or abort when it finds one. I don't think there is an explicitly coded limit to recursive (non-cyclic) macro substitutions. At some point it will run out of memory, though, I suppose.

check/README.md
geoffthemedio: then **this too** -> **these two**

Ouaz (approved):

PR #3657, opened 2022-01-09
Add dump of tested tech to Boost.Test log

o01eg:


geoffthemedio:
anything in particular I should look at?

o01eg:
@geoffthemedio No, only if it raises some concerns.

PR #3660, opened 2022-01-11
adding unicode glyph U+22C4 "Diamond Operator" to roboto font

st-pa-fo:
... and adjusting font version number and copyright-notice

--> https://www.freeorion.org/forum/viewtopic.php?p=109183#p109183


geoffthemedio:
doesn't seem to work... ⋄ shows up as a blank space for me

st-pa-fo:
on what kind of system? (not sure if that would make a difference though) and i admit, i did not even bother to try to close that glyph's polygon after seeing the first apparently working version .. soo, i'll look into that some more. edit: mentioning you by name @geoffthemedio

geoffthemedio:
Windows 10

st-pa-fo:
latest macos here. please tell me if you can see those test lines i put in the pedia index with the most recent commit
?
dot operator

edit: should of mixed em up more for verticel comparability and should of used hyphen stead of dash .. if it not displayed only on windows .. hm no idea

geoffthemedio:
text test

st-pa-fo:
should be like that
like that and no i did not try to morse code anything

geoffthemedio:
The ⋄ appears very small and barely distinguishable from ⋅

st-pa-fo:
and aligns well enough when single signs are in a column which was the point of the exercise i think ... there'd be other unicodepoints with bigger glyphs that we could use, i'd hesitate to make the 'dot' bigger than, well, a dot. is the dot operator from math, so .. something circular with emptiness in the middle, maybe the ⚬? as in +–⚬–+ as opposed to o⚬o

st-pa-fo:
> The ⋄ appears very small and barely distinguishable from ⋅

agreed. not exactly a masterpiece. but for the very first try .. can be improved upon, no doubt

st-pa-fo:
> +–⚬–+ ... o⚬o

disclaimer: any similarities to copyrighted computergame material pre-90's is purely unintentional and by accident ... sorry, couldnt help it, just looked a little like a tie bomber and a y-wing ..

st-pa-fo:
ping @geoffthemedio – unclear how i should proceed.
a) make the diamond hollow*
b) use the dot instead**
c) no change***
d) something else
either way, this should not affect the other pr with the minor tweakage, since the diamond operator used there will simply be rendered as space sign in the between time

*:different fonts out there in the wild will render that diamond operator differenty. some really make it big and hollow, in others it is more like the adjacent dot operator but just a bit sharpened to the edges.
**:need to rewrite the minor tweakage in that case
***:not sure why this should not work without further changes

geoffthemedio:
If you want to add ⋄ to the font, then I suggest doing so in a way that makes it distinguishable from other characters that it's being specifically added to be distinguishable from. Bigger and hollow sounds like a reasonable way to do that.

st-pa-fo:
done. see screenshots attached, one with normal font size and one with max font size.

font size 15
font size 40

geoffthemedio:
Better, though still rather small I think. In the font of this site / my browser, ⋄ is about the same size as + rather than as ⋅

st-pa-fo:
made it bigger. i would then proceed to compress the commits into a single commit and rebase on master. right?

bigger diamond thanks to the scaling tool

geoffthemedio:
doesn't need a rebase, but should get rid of the stringtable changes. combining into one commit is fine, but I could also squash and merge.

st-pa-fo:
done. reverted string change. did not rebase.

st-pa-fo:
good compromise i'd say, and i know, here on githup they render it as a big diamond pebble, but as i said, in other places it shows up small. attaching screenshots. nevermind though, all is good.

diamond operator in font forge

diamond operator in source tree

PR #3664, opened 2022-01-12
Update st-toot to format files inplace

Cjkjvfnby:
This behavior should be much more convenient for users.

PS. Added em.txt changes to check the CI


Cjkjvfnby:
@st-pa-fo , FYI st tool now formats string table file in place.

PR #3668, opened 2022-01-16
Detection Tweaks

geoffthemedio:
Based on https://freeorion.org/forum/viewtopic.php?p=109150#p109150
-make snowflake bodies have good instead of excellent vision
-make detection techs increment range in steps of 25 instead of 25 or 50
-use value reference for tech detection ranges in pedia
-reduce living hull detection boost from 50 to 40
-reduce scanning facility range boost from 75 to 50


Oberlus:
If this gets merged before tomorrow, we'll playtest it in next MP game.

PR #3669, opened 2022-01-16
Print address to console, exit service gracefully

Cjkjvfnby:
@st-pa-fo, could you give another try to st-tool edit mode. 

Could you start the server and try to open the link `http://localhost:8080/` in different browsers?  

You can use a version from the master.  These changes are mostly cosmetical.

### DIFF
--- a/check/st-tool.py
+++ b/check/st-tool.py
@@ -910,8 +910,12 @@ def edit_action(args):
     server = HTTPServer(("localhost", 8080), EditServerHandler)
     server.reference_st = reference_st
     server.source_st = source_st
+    print("Starting server at http://localhost:8080/")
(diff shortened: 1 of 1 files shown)


st-pa-fo:
oh sorry i am just now going through hundreds of mails accumulated the last few days - i'll give it a try soon. but, to be perfectly honest, i'd even much rather spend two days or three setting up something so i could do the translations with omegaT instead and try to integrate that with the existings workflows .. just seems more sensible to use fully fledged open source translation tool for translation of fully fledged open source projects is all..

edit; i picture a separate translations repository for every target localization, omegaT supports so called 'team projects' and does the git thing itself, commits of translations would be easily reviewable, and Ωt brings many possibilities for translators to increase productivity and quality of work

PR #3670, opened 2022-01-17
Species likes/dislikes tweaks

MichaelCourtney:


default/scripting/species/SP_FULVER.focs.txt
```diff
@@ -42,6 +42,15 @@ Species
         [[LARGE_PLANET]]
         [[NARROW_EP]]
         [[STANDARD_SHIP_SHIELDS]]
+		
+		EffectsGroup
```
geoffthemedio: Weird indenting. Are there tabs used?

MichaelCourtney: yeah looks like its 2 tabs, notepad++ just defaulted to that when I started a new line. Its in line with [[STANDARD_SHIP_SHIELDS]] when I look at it in npp

MichaelCourtney:
Better?

default/scripting/species/SP_FULVER.focs.txt
geoffthemedio: should replace tabs with 4 spaces

MichaelCourtney: should be fixed below

geoffthemedio:
Since you're changing scripts that are visible to the checksums, could you also update the expected checksums for the tests:
````
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_NAMED_VALUEREF expected 2351423 was 2465801
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_TECH expected 5454078 was 5355498
````

MichaelCourtney:
Is that correct now? In general how do I know when the expected checksums need adjusting and what the correct values are?

geoffthemedio:
You can check the result of the tests on github for your pull request, or run the test locally. It's likely to change the checksums if you modify content scripts.

PR #3672, opened 2022-01-19
Test all checksum and remove some unused tokens

o01eg:
Unused tokens get with bash line:

```bash
grep -Po "^.*\(\K.+(?=\))" parse/Tokens.h | while read TOKEN ; do { grep -Rni --exclude=*.py "${TOKEN}" default/scripting/* 2>&1 1>/dev/null ; USED=$? ; if [ "1" == "${USED}" ] ; then echo $TOKEN ; fi } done
```

### DIFF
--- a/parse/Tokens.h
+++ b/parse/Tokens.h
@@ -8,8 +8,6 @@
     (abs)                                       \
     (accountinglabel)                           \
     (activation)                                \
-    (AddedBefore)                               \
-    (AddedSince)                                \
     (AddSpecial)                                \
     (AddStarlanes)                              \
     (Adequate)                                  \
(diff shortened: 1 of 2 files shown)


geoffthemedio:
looks reasonable.

would be good to add those missing features to the parser, though.

geoffthemedio:
I'm pondering if it would be useful to add `vector<unsigned int> GetCheckSumsList()` to various manager classes, which will return a list of the checksums of each of the things in the manager. If the checksum for a manager doesn't match, you can check the checksums of all the things in it. Not sure how to get a maintainable list of "correct" values, though. Could probably add a new message type between client and server to get more detailed checksums, though, so exactly what doesn't match between them can be reported in the error log...

PR #3675, opened 2022-01-20
Species policy preferences

MichaelCourtney:


hifron:
What about some maluses, overhead it could not be alone...

MichaelCourtney:
That should be the lot of them. Some are a bit arbitrary but I guess that can be fixed later if anyone disagrees

hifron:
researchable...

geoffthemedio:
@Oberlus any comment?

I guess you already 👍

Issue #3676, opened 2022-01-20
Invading the planet with Asteroid Coating special should not remove the bonus from all ships


MichaelCourtney:
![asteroidcoating](https://user-images.githubusercontent.com/43386669/150377390-9478906d-7c4e-44c4-805f-0d6e1ede9cc7.png)

geoffthemedio:
There are two specials with the same name [in the scripts](https://github.com/freeorion/freeorion/tree/master/default/scripting/specials). Not sure why, but seems to have something to do with visibility.

hifron:
Coating of Arms?

agrrr3:
> There are two specials with the same name [in the scripts](https://github.com/freeorion/freeorion/tree/master/default/scripting/specials). Not sure why, but seems to have something to do with visibility.

Looks like it becomes visible once one actually "uses it". So it might be you stumble upon it when you actually build an outpost on an asteroid belt (which is ok), and also makes it visible to enemies in that case (also fine). Not really necessary, but ok.

ASTEROID_COATING_SPECIAL_DESC could be Asteroid Coating (Hidden) or (Unexploited) or (Unclaimed).
And description/sitrep should tell about that change in visibility.

Another issue: Actually i think we said the coating should actually add a special to supplied ships, so invading the special does not remove the bonus but cutting it off keeps new ships from getting the bonus.

And a kind of serious issue: Also it should boost the current structure when it gets applied the first time, else it will create of tons "Ship has been repaired" sitreps, also every time you actually build a new ship. This had been discussed on the forum.

agrrr3:
the two are different in the pedia in master -  "Asteroid Coating" and  "Asteroid Coating (Owned)"

can somebody close or reword this issue?

geoffthemedio:
> reword this issue?

To what?

agrrr3:
> > reword this issue?
> 
> To what?

"Invading the planet with Asteroid Coating special should not remove the bonus from all ships"

even if we want that changed (using a special), i would rather keep getting the bonus without supply lines - else getting the bonus is quite awkward for disconnected empires.

PR #3677, opened 2022-01-20
Combine ASTEROID_COATING into single file

rrobgill:
The various effects of the ASTEROID_COATING special were split across two files, which have now been combined into a single file.

This avoids duplication of entries in the pedia described in  #3676 


agrrr3:
this changes not only the extra entry, but changes the behaviour, check the discussion #3676 
would be great if you could fix the implementation issues I mentioned there

rrobgill:
The behaviour should be the same after this. The second (deleted) file reduced stealth to 5 and awarded the bonus to the ships when the asteroids were occupied. These functions were rolled into the first file.

I'll have a look at the other issues.

geoffthemedio:
Is the key thing here the merging into a single file or a single special definition?

default/scripting/specials/ASTEROID_COATING.focs.txt
```diff
@@ -24,10 +24,16 @@ Special
                 OwnedBy affiliation = AnyEmpire
             ]     
             activation = Source
-            effects = [
-                AddSpecial name = "ASTEROID_COATING_OWNED_SPECIAL"
-                RemoveSpecial name = "ASTEROID_COATING_SPECIAL"
+            effects = SetStealth value = 5
```
geoffthemedio, line 27: This sets the stealth of the object that has the special attached, which is not the same as having two specials with different stealth levels for the specials themselves, which was, I think, the main reason that there were two separate special definitions. The result of this is that the object with this special will have its stealth set to 5 at some middleish point during effect exectution, but the special itself will have stealth 25 instead of 5.

default/scripting/specials/ASTEROID_COATING_OWNED.focs.txt
```diff
@@ -1,30 +0,0 @@
-Special
-    name = "ASTEROID_COATING_OWNED_SPECIAL"
-    description = "ASTEROID_COATING_OWNED_SPECIAL_DESC"
-    stealth = 5
```
geoffthemedio, line 4: this is the stealth of the special, not the object with the special attached

default/scripting/specials/ASTEROID_COATING.focs.txt
agrrr3: we do not track stealth per special currently; a special "instance" is actually an entry in the `std::map<std::string, std::pair<int, float>> m_specials;` in UniverseObject. (tracks added-on-turn and capacity)

Also the role of stealth of specials is not really well defined. E.g. ion storms should not be able to affect the stealth of a special.

so currently the implementation using two specials is the only option. not sure we should change something in the backend (e.g. adding stealth values to the m_specials map)

geoffthemedio: alternative is to make special stealth a valueref.

agrrr3: > alternative is to make special stealth a valueref.

so one could return in this case a different value based on ownership; clever

default/scripting/specials/ASTEROID_COATING_OWNED.focs.txt
rrobgill: thanks got it. the two files workaround makes sense now

rrobgill:
current situation seems more suitable

PR #3678, opened 2022-01-21
Updated info about provide at ArchLinux

ghost:


geoffthemedio:
@Vezzra @o01eg seems reasonable?

o01eg:
@geoffthemedio I don't use Arch Linux but looks fine.

Vezzra:
Same as @o01eg - I've never used ArchLinux, so I can't really provide meaningful input, but AFAICT looks ok...

PR #3679, opened 2022-01-21
Add option for Title screen music

rrobgill:
Switches between main background music in gameplay and adds option of different
music for the title screen.

Additional menu entry in settings to select file for title screen music.

NB. No music is included, defaults to using existing music file.

Discussion in Issue #3350 and at https://www.freeorion.org/forum/viewtopic.php?p=105422

### DIFF
--- a/client/human/HumanClientFSM.cpp
+++ b/client/human/HumanClientFSM.cpp
@@ -17,6 +17,7 @@
 #include "../../UI/MultiplayerLobbyWnd.h"
 #include "../../UI/PasswordEnterWnd.h"
 #include "../../UI/MapWnd.h"
+#include "../../UI/Sound.h"
 
 #include <boost/format.hpp>
 #include <boost/uuid/uuid_io.hpp>
@@ -128,6 +129,7 @@ IntroMenu::IntroMenu(my_context ctx) :
     Base(ctx)
 {
     TraceLogger(FSM) << "(HumanClientFSM) IntroMenu";
+    Sound::GetSound().PlayTitleMusic();
     Client().GetClientUI().ShowIntroScreen();
     GetGameRules().ResetToDefaults();
 }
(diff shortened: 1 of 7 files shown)


client/human/HumanClientFSM.cpp
```diff
@@ -138,23 +140,27 @@ IntroMenu::~IntroMenu()
...
     TraceLogger(FSM) << "(HumanClientFSM) IntroMenu.HostSPGameRequested";
     Client().Remove(Client().GetClientUI().GetIntroScreen());
+    Sound::GetSound().PlayBackgroundMusic();
     return transit<WaitingForSPHostAck>();
 }
 
 boost::statechart::result IntroMenu::react(const HostMPGameRequested& a) {
     TraceLogger(FSM) << "(HumanClientFSM) IntroMenu.HostMPGameRequested";
     Client().Remove(Client().GetClientUI().GetIntroScreen());
+    Sound::GetSound().PlayBackgroundMusic();
```
geoffthemedio, line 150: I think it makes more sense to put calls to start playing music when constructing an FSM state, rather than when reacting to an event in a state that is being exited. eg. https://github.com/freeorion/freeorion/pull/3358/files#diff-5926c0324401788b2baa5920aff99a5f1d8938f2694222070badba11218400d6

edit: although with my version it restarts the music was transiting back from MP lobby to title screen without going into a game, which seems awkward.

It's probably in general easier to keep the music state consistent when setting it when constructing FSM states, but somewhat awkward to do so now since there are multiple top-level states other than `PlayingGame` and whether or not one is in `PlayingGame` is currently the primary way to decide what music should be playing. Restarting the music when going into `IntroMenu` is awkward due to `MPLobby` being a separate state. Can't just set music when going in and out of `PlayingGame` as the title music needs to start up at the start of the game. Could possibly add another top-level state that includes all the states other than `PlayingGame`...

geoffthemedio: I'd also start the background music when actually in the game, not when transiting to the MP lobby.

geoffthemedio:
another issue with this: when the title music is set so something different from background music, when in the intro screen options menu, if the music is toggled off and on, it restarts with the background music, not the title music.

geoffthemedio:
@rrobgill any chance you'll take another look at this soon?

rrobgill:
Redundant to  #3358

PR #3681, opened 2022-01-24
PlanetTypeDistance valueref for terraform

agrrr3:
access to planet type difference in FOCS

### DIFF
--- a/parse/DoubleValueRefParser.cpp
+++ b/parse/DoubleValueRefParser.cpp
@@ -82,7 +82,7 @@ parse::double_parser_rules::double_parser_rules(
     int_rules(tok, label, condition_parser, string_grammar),
     simple_int_rules(tok),
     simple_double_rules(tok),
-    int_complex_grammar(tok, label, int_rules, string_grammar),
+    int_complex_grammar(tok, label, int_rules, condition_parser, string_grammar),
     double_complex_grammar(tok, label, condition_parser, string_grammar)
 {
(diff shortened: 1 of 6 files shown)


agrrr3:
note that i did not change content nor checksum calculation, so the CI fail seems bogus

geoffthemedio:
> note that i did not change content nor checksum calculation, so the CI fail seems bogus

If it's SmokeTestHostless failing, then it can be ignored presently. It just seems to randomly fail 75% of the time or so.

I've restarted the Ubuntu tests, which sometimes gets it to work.

geoffthemedio:
I combined your stuff into one commit, then reworked a bit as https://github.com/freeorion/freeorion/tree/planet_type_difference_parsing

I'd rather use ints than strings internally, since it's just an enum and the conversion is easier, and it should matching and storing the FOCS strings like Ocean rather than the enum string representations like PT_OCEAN anyway.

I also got rid of the shared_ptr by rearraning include order a bit.

geoffthemedio:
@agrrr3 does that also work for you?

geoffthemedio:
pushed my version to master https://github.com/freeorion/freeorion/commit/4a7eeb01d5eb6f76ffd4c399ff70943d26596924

agrrr3:
Yes, I want mostly something what works. So, works for me

Geoff ***@***.***> schrieb am So., 30. Jän. 2022, 22:23:

PR #3686, opened 2022-01-30
add influence stockpile to PlayerListWnd

TheSilentOne1:
Adds influence stockpile to Playerlist window. Any ideas if it should be influence output or stockpile?
[Think I pushed this to origin by mistake. Should I delete and repush?]

### DIFF
--- a/UI/PlayerListWnd.cpp
+++ b/UI/PlayerListWnd.cpp
@@ -80,6 +80,10 @@ namespace {
         static std::shared_ptr<GG::Texture> retval = ClientUI::GetTexture(ClientUI::ArtDir() / "icons" / "meter" / "research.png");
         return retval;
     }
+    std::shared_ptr<GG::Texture> InfluenceIcon() {
+        static std::shared_ptr<GG::Texture> retval = ClientUI::GetTexture(ClientUI::ArtDir() / "icons" / "meter" / "influence.png");
+        return retval;
+    }
     std::shared_ptr<GG::Texture> DetectionIcon() {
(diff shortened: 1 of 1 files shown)


geoffthemedio:
The pull request itself looks fine... not sure what you think you did wrong.

Ah, I see, it's in a freeorion branch rather than one of your branches, like how when I push a branch. Should be fine regardless.

Edit: Doesn't really work:
influence_on_empires

geoffthemedio:
I think the calculation of these stats client-side should be removed anyway, and instead it should show numbers from the empire statistics as seen in the empire graphs, which are calculated server-side.

TheSilentOne1:
Hm, I guess I didn't think this through. Ofc player will not know the other players stockpile(s), at least not by deafult. For now I think I'll switch to showing influence generation as known by intel.

agrrr3:
> Hm, I guess I didn't think this through. Ofc player will not know the other players stockpile(s), at least not by deafult. For now I think I'll switch to showing influence generation as known by intel.

i think showing influence generation from intel is good. At least on multiplayer we hide the server stats, so showing a sum of what is visible/known helps.

TheSilentOne1:
Will resume work on this soon.

geoffthemedio:
Probably would be useful to show the empire statistics, same as seen in the graphs screen, if they are available, and a calculation based on the known universe if they aren't. Not sure how best to set that up, though. I don't like hard-coding a particular statistic reference in the GUI, but that might be reasonable for this case...

PR #3688, opened 2022-02-02
Apply defensive flanking bonus only if there are at least two defenders

agrrr3:
defensive flanking bonus should not be applied when there is a single defending ship.

this matters mostly because of 1-turn delay and also of stealthed enemies. 

With the current system we cant exempt stealthed attackers from the defensive flanking as it would give away the position. At least for the single defender we can correct this.

### DIFF
--- a/default/scripting/policies/FLANKING.focs.txt
+++ b/default/scripting/policies/FLANKING.focs.txt
@@ -34,7 +34,7 @@ Policy
                                Stationary
                                Armed
                            ]
-                        ) > (
+                        ) > max( 1, // no bonus if there is only a single defender
                            Statistic Count condition = And [
                                Ship
                                InSystem id = RootCandidate.SystemID


geoffthemedio:
update checksums?
https://github.com/freeorion/freeorion/runs/5055794418?check_suite_focus=true#step:13:5842
````
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_BUILDING expected 3444689 was 5972872
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: info: check 'FO_CHECKSUM_ENCYCLOPEDIA expected 1078929 was 1078929' has passed
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: info: check 'FO_CHECKSUM_FIELD expected 3780088 was 3780088' has passed
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_POLICY expected 1239846 was 1243515
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: info: check 'FO_CHECKSUM_SHIP_HULL expected 9937526 was 9937526' has passed
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: info: check 'FO_CHECKSUM_SHIP_PART expected 9683018 was 9683018' has passed
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: info: check 'FO_CHECKSUM_SHIP_DESIGN expected 872567 was 872567' has passed
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_SPECIES expected 2811615 was 6513018
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: info: check 'FO_CHECKSUM_SPECIALS expected 6000305 was 6000305' has passed
5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_TECH expected 5355498 was 5367038
````

agrrr3:
> update checksums? https://github.com/freeorion/freeorion/runs/5055794418?check_suite_focus=true#step:13:5842
> 

did neither change buildings, species, nor techs (as errors indicate). guess that is/was from master.

but I did change a policy. how do i get the correct checksum for that?

geoffthemedio:
> but I did change a policy. how do i get the correct checksum for that?

It tells you when the test is run:
`5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_POLICY expected 1239846 was 1243515`

agrrr3:
> > but I did change a policy. how do i get the correct checksum for that?
> 
> It tells you when the test is run: `5: /home/runner/work/freeorion/freeorion/test/system/TestChecksum.cpp:22: error: warning: in "TestChecksum/compare_checksum": FO_CHECKSUM_POLICY expected 1239846 was 1243515`

well if i simply copy from the CI whenever it changes then we definitely do not need the test; it would be just adding bureaucracy for exactly no gain. So how do i get the checksum locally?

geoffthemedio:
> if i simply copy from the CI whenever it changes then we definitely do not need the test; it would be just adding bureaucracy for exactly no gain.

The test isn't for you; it's for verifying that the Python parsed content matches the Boost / FOCS parsed content.

> So how do i get the checksum locally?

I think you'd need to add something to output them, eg. in https://github.com/freeorion/freeorion/blob/master/universe/Universe.cpp#L3331

agrrr3:
> The test isn't for you; it's for verifying that the Python parsed content matches the Boost / FOCS parsed content.
> 

yes,  part of that i understand. one can use it as development tool. 

so if i change something i make first sure the checksum is the correct one and then i make my changes and rerun the test for regression. no need to keep all checksums updated all the time for that.

a better test for the stated intent would be parsing the contents with both parsers and then expect the same checksums.

geoffthemedio:
> a better test for the stated intent would be parsing the contents with both parsers and then expect the same checksums.

Converting a .focs.txt to a .focs.py [replaces the old file with the new one](https://github.com/freeorion/freeorion/commit/01d78a71f46687d68cbb1f1fe49ed992d4b41143#diff-37a8eec1ce19687d132fe29051dca629d164e2c4958ba141d5f4133a33f0688f). It doesn't add a second script to produce the same content. If both were present, it would probably give an error about duplicate content, and even if it didn't, it would only produce a checksum for whichever of the two is retained, not both, so there would be nothing to compare to, except an external record of what the checksum should be (ie. the test).

agrrr3:
> there would be nothing to compare to, except an external record of what the checksum should be (ie. the test).

ok, valid. So if that is the only reason for the test it is also the only point in time where the checksums matter. No reason to keep it updated on every content change and creating conflicts any time the same kind of content (e.g. building types) gets updated in master and the branch.

Issue #3689, opened 2022-02-03
Flux Bubble Hull doesn't improve colony building speed


MichaelCourtney:
Bug Report
==========
This was presumably just missed when the flux line was split from robotics

geoffthemedio:
resolved by https://github.com/freeorion/freeorion/commit/bb82da1287443cb89129388dc1dd985cb338be69

PR #3690, opened 2022-02-03
Flux bubble should decrease colony building time

MichaelCourtney:


MichaelCourtney:
Issue #3689

default/scripting/buildings/colonies/SP_BANFORO.focs.txt
```diff
@@ -9,22 +9,23 @@ BuildingType
...
+               And [
+                   Species name = "SP_BANFORO"
+                   Population low = [[MIN_RECOLONIZING_SIZE]]
+                   Happiness low = 5
+                ]
+                And [
+                    HasSpecial name = "EXTINCT_BANFORO_SPECIAL"
+                    Contains Building name = "BLD_XENORESURRECTION_LAB"
                 ]
+            ]
```
agrrr3: unintended whitespace changes?

agrrr3 (commented):
besides that whitespace thingy LGTM

agrrr3:
note i think the AI has no clue that there is a way to speed up colonization, but i also think it does not need to know

default/scripting/buildings/colonies/SP_BANFORO.focs.txt
MichaelCourtney: Yeah. I didn't manually change this file, I just changed the python script that generates it

MichaelCourtney:
Spatial flux hulls are still showing up under the robotic line in the pedia. If I work out how to change that I'll fix that too

default/scripting/buildings/colonies/SP_BANFORO.focs.txt
MichaelCourtney: Checked this. The spacing for the indent of Banforo was edited in master by Geoff over here: https://github.com/freeorion/freeorion/commit/0741af7426f291a1718108d2920725f562d843dc

The spacing for the indent in this commit is consistent with the other extinct species and the spacing generated by col_bld_gen.py

geoffthemedio: Probably col_bld_gen.py should be adjusted to fix the indenting.

My editing the the .focs.txt file was a bad idea and is indicating as such in the comment at the start of the file. So reverting that change is fine.

agrrr3 (commented):
only thing which could be missing is the colony building for exobots, else it is good to go

MichaelCourtney:
> only thing which could be missing is the colony building for exobots, else it is good to go

Looks like exobot colony building isn't affected as it has no source colony, its always 5 turns to build

PR #3693, opened 2022-02-04
Try to hide CI SmokeTestHostless failure

### LINKED ISSUE #3684: SmokeTestHostless Randomly Fails
The SmokeTestHostless test is presently less than useless, in that it seems to randomly fail for no reason, or no reason related to whatever the most recent commits are about, leading to reported CI failures that make it difficult to tell if a change actually broke something.

An example: https://github.com/freeorion/freeorion/runs/4896276619?check_suite_focus=true#step:13:8907
Or: https://github.com/freeorion/freeorion/runs/4873208942?check_suite_focus=true#step:13:5587
Or: https://github.com/freeorion/freeorion/pull/3681#issuecomment-1022550569

Re-running the tests several times will often result in a different one of the 4 Ubuntu tests failing, with the others then being cancelled. Eventually it usually manages to run them all. Other times they all work on the first try.

o01eg:
Should fix #3684

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -105,6 +105,8 @@ jobs:
         run: |
           ccache --show-stats || true
       - name: Run C++ unit tests
+        env:
+          FO_TEST_HOSTLESS_GAMES: 1
         run: |
           cd build
           cmake --build . --config Release --target unittest


geoffthemedio:
Could you briefly comment how this works?

Otherwise seems OK... in that all the tests pass... but I'm not sure if or which of them would otherwise have been a failure.

o01eg:
CI somehow corrupts sockets in reconnection so `FO_TEST_HOSTLESS_GAMES` limits numbers of games in hostless test to 1 (no reconnection for the next game)

PR #3696, opened 2022-02-09
Stabilitytweaks

MichaelCourtney:
This seems cleaner to me. The old version was giving +10 stability to Sly and Laenfa to compensate for a penalty they did not have. Also you were penalized for a poor supply connection, but not penalized for no supply connection.

### DIFF
--- a/default/stringtables/en.txt
+++ b/default/stringtables/en.txt
@@ -11984,8 +11984,8 @@ Confederation
 PLC_CONFEDERATION_DESC
 '''Planets of the empire are organized in a confederacy of independent systems.
 
-• Increases [[metertype METER_TARGET_INFLUENCE]] on colonies disconnected from the capital by [[value PLC_CONFEDERATION_SUPPLY_DISCONNECTED_INFLUENCE_FLAT]]
-• Increases [[metertype METER_TARGET_HAPPINESS]] on colonies disconnected from the capital and regional administrative centres by [[value DISCONNECTED_FROM_CAPITAL_AND_REGIONAL_ADMIN_STABILITY_PENALTY]]
+• Removes [[metertype METER_TARGET_INFLUENCE]] penalty (baseline for most species) on colonies disconnected from the [[buildingtype BLD_IMPERIAL_PALACE]]
+• Removes [[metertype METER_TARGET_HAPPINESS]] penalty (baseline for most species) on colonies disconnected from the [[buildingtype BLD_IMPERIAL_PALACE]] and [[buildingtype BLD_REGIONAL_ADMIN]]
 • Reduces [[metertype METER_TARGET_HAPPINESS]] ([[value PLC_CONFEDERATION_TARGET_HAPPINESS_FLAT]]) on all planets
 • Increases [[metertype METER_MAX_TROOPS]] by [[value PLC_CONFEDERATION_MAX_TROOPS_FLAT]]'''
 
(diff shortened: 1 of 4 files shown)


default/stringtables/en.txt
```diff
@@ -11984,8 +11984,8 @@ Confederation
 PLC_CONFEDERATION_DESC
 '''Planets of the empire are organized in a confederacy of independent systems.
 
-• Increases [[metertype METER_TARGET_INFLUENCE]] on colonies disconnected from the capital by [[value PLC_CONFEDERATION_SUPPLY_DISCONNECTED_INFLUENCE_FLAT]]
-• Increases [[metertype METER_TARGET_HAPPINESS]] on colonies disconnected from the capital and regional administrative centres by [[value DISCONNECTED_FROM_CAPITAL_AND_REGIONAL_ADMIN_STABILITY_PENALTY]]
+• Removes [[metertype METER_TARGET_INFLUENCE]] penalty on colonies disconnected from the capital
```
geoffthemedio: Could be misleading... It compensates for a specific penalty, but doesn't prevent another source of such a penalty, and doesn't do anything if that penalty wasn't present.

MichaelCourtney: Hmm I'll have a think about it and see if I can come up with clearer wording. If stability penalties were in the pedia I could link to the penalty. Maybe its worth creating a pedia entry for stability modifiers? That might be worthwhile in itself.

Grummel7: It was possibly not intended, but Sly and Laenfa were a lot happier with confederation on disconnected planets. They got the bonus without having the penalty.
They won't like your change ;-)

MichaelCourtney: > It was possibly not intended, but Sly and Laenfa were a lot happier with confederation on disconnected planets. They got the bonus without having the penalty. They won't like your change ;-)

Its a big part of why I suggested changing it. Sly and laenfa independent stability was added in after confederation and suddenly they got +10 stability pretty much for free. I believe trying to add 10 to compensate the penalty, rather then just cancelling is going to be prone to these sort of bugs.

default/stringtables/en.txt
```diff
@@ -11984,8 +11984,8 @@ Confederation
...
+• Removes [[metertype METER_TARGET_HAPPINESS]] penalty on colonies disconnected from the capital and regional administrative centres
```
geoffthemedio: Maybe `• Removes [[metertype METER_TARGET_HAPPINESS]] penalty (baseline for most species) on colonies disconnected from the capital and [[buildingtype BLD_REGIONAL_ADMIN]]`

MichaelCourtney: ok will change it. There's probably a buildingtype for the capital too, imperial palace. I was planning on adding a modifier for planet size (https://www.freeorion.org/forum/viewtopic.php?f=6&t=12214) and maybe one for planet suitability to the PR. Maybe still will. I assume you can selectively merge the commits, or I can remove contentious commits that don't pass.

MichaelCourtney: guess, I'll just leave any extra commits out for now pending any discussion that happens in the forum thread. Heading out, will fix the description when I get home and it'll be ready

MichaelCourtney:
left out a "from". Fixed now. One thing I noticed is the influence penalty for being disconnected from capital, only checks the palace and not regional admins. Intentional?

PR #3697, opened 2022-02-10
Fix mutual exclusions between policies, keep list ordered

Oberlus:
Corrects unpaired exclusions between policies.
Grooming (keep lists of policy exclusions ordered for easier maintaining).

Discussion: https://www.freeorion.org/forum/viewtopic.php?p=109889#p109889


st-pa-fo:
just wondering, would be nice to have a specialized policy graph window akin to the research tech tree? where the locks and unlocks and exclusions are marked as vertices and the policies and techs and what ever else as nodes in the tree? or maybe as additional display option for the tech tree. beyond the scope of this pr here tho and did not want to put it in the nineteenth online game channel.. probably safe to ignore this comment.

PR #3698, opened 2022-02-10
add some new field graphics (WIP)

TheSilentOne1:
Work in progress on some new fields.


st-pa-fo:
can we have some colour variants of the nanite swarm please while you are at it? i was gonna make a bad nanite swarm of corrosive nanites to be attracted by a soon to be new core slot ship part called the Rotten Core module … thx ^.^

TheSilentOne1:
Sounds good, I'll see what I can do.

default/scripting/fields/FLD_CORROSIVE_NANITE_SWARM.focs.txt
```diff
@@ -0,0 +1,66 @@
...
+                ])
+
+        EffectsGroup    // shrink when ships are in...
+            scope = Source
+            activation = Number low = 1 condition = And [
+                Ship
+                WithinDistance distance = Source.Size * 0.9 condition = Source
+                Not Monster
+            ]
+            effects = SetSize value = Value - Statistic Count condition = And [
```
geoffthemedio, line 49: I'm not sure if this whole script is just for demo purposes, but I would have the swarm grow rather than shrink when eating...

TheSilentOne1: Right, scripting is not finalized, @st-pa-fo probably has some plans?

st-pa-fo: yes well nothing tangible yet, the colour image variant would be great as a stand alone, but feel free to do the scripting as well ^..^ i probably would just have either copy/pasted existing nanites or made them parametrized for the effect like two variables one for the affected meter and one for the inflicted value of affection - i.e. the hull plus for the existing variant, and then hull minus for a new one, and maybe one that eats only fighters out of the bays and one other that rebuilds them. lame. i know. but i had no time to think about it much

st-pa-fo: maybe flesh eating nanites/microbes? only affect organic hulls and monsters. or maybe like the nanites from red dwarf where they resurrect dead populations or embiggen small hull space ships to huge hulls .. and that emotion kraken from the JMCS Esperanto i believe it was called, that induces hallucinations and lets ..
stop. off topic. thanks for asking. i'll get back to that.

geoffthemedio:
@TheSilentOne1 poke

TheSilentOne1:
Mmh, my plans are to add some more nebulae now and then. I'd prefer if @st-pa-fo did the scripting on the evil nanites himself, but if he doesn't feel up to it I will do it sometime.

PR #3700, opened 2022-02-11
Python API Tweaks

geoffthemedio:
Fixes how policy adoption cost is exposed to Python API, which before this requires an impossible-to-provide `ScriptingContext` parameter.

@Cjkjvfnby @Morlic-fo I'm not sure how to regenerate [freeOrionAIInterface.pyi](https://github.com/freeorion/freeorion/blob/master/default/python/AI/freeOrionAIInterface.pyi)

### DIFF
--- a/python/EmpireWrapper.cpp
+++ b/python/EmpireWrapper.cpp
@@ -36,14 +36,13 @@ namespace {
 
     auto obstructedStarlanes(const Empire& empire) -> std::vector<std::pair<int, int>>
     {
-        const std::set<std::pair<int, int>>& laneset = GetSupplyManager().SupplyObstructedStarlaneTraversals(empire.EmpireID());
-        std::vector<std::pair<int, int>> retval;
+        const auto& laneset = GetSupplyManager().SupplyObstructedStarlaneTraversals(empire.EmpireID());
+        static_assert(!std::is_same_v<std::decay_t<decltype(laneset)>, std::vector<std::pair<int, int>>>); // if are the same, don't need to explicitly construct the return value...
         try {
-            for (const std::pair<int, int>& lane : laneset)
-            { retval.push_back(lane); }
+            return {laneset.begin(), laneset.end()};
         } catch (...) {
+            return {};
         }
-        return retval;
     }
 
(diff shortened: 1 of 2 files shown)


python/EmpireWrapper.cpp
```diff
@@ -485,7 +485,9 @@ namespace FreeOrionPython {
             .add_property("description",            make_function(&Policy::Description,         py::return_value_policy<py::copy_const_reference>()))
             .add_property("shortDescription",       make_function(&Policy::ShortDescription,    py::return_value_policy<py::copy_const_reference>()))
             .add_property("category",               make_function(&Policy::Category,            py::return_value_policy<py::copy_const_reference>()))
-            .def("adoptionCost",                    &Policy::AdoptionCost)
+            .add_property("adoptionCost",           +[](const Policy& p)                       { return p.AdoptionCost(AppEmpireID(), ScriptingContext{}); })
```
Grummel7: The property does not work. The interface generator seems to ignore it an when I add a matching property manually and print `str(policy.adoptionCost)` I get `<bound method adoptionCost of <freeOrionAIInterface.policy object at 0x7ff14c069400>>` instead of a float

When I changed it to...
```suggestion
            .def("adoptionCost",                    +[](const Policy& p)                       { return p.AdoptionCost(AppEmpireID(), ScriptingContext{}); })
```
...I got a member function that works, which does the job for me.

python/EmpireWrapper.cpp
```diff
@@ -485,7 +485,9 @@ namespace FreeOrionPython {
...
+            .def("adoptionCost",                    +[](const Policy& p, const Empire& empire) { return p.AdoptionCost(empire.EmpireID(), ScriptingContext{}); })
+            .def("adoptionCost",                    +[](const Policy& p, int empire_id)        { return p.AdoptionCost(empire_id, ScriptingContext{}); })
         ;
```
Grummel7, line 491: For other empires, this will be guesses based on the AIs information?

geoffthemedio: probably yes. this might be useful in contexts other than AI scripts, though.

PR #3704, opened 2022-02-15
Popular Terraform using NextBestPlanetType valueref

agrrr3:
while the target terraforming is a huge improvement, some little things got lost.
this adds another terraform building which brings back some of the original functionality 
* mass-terraform using object window
* simple way to terraform if you dont know the species environment preferences

it leaves out a lot of stuff, because the targeted terraform buildings provide that much better:
* possibility of doing transformation in steps
* terraformation to something

also quite some cleanup and polishing

[forum discussion](https://freeorion.org/forum/viewtopic.php?f=28&t=12278)


default/scripting/buildings/TERRAFORM.focs.txt
```diff
@@ -84,5 +84,80 @@ TARGET_TERRAFORMING
     icon = "icons/building/terraform.png"
 '''
 
+BuildingType
+    name = "BLD_TERRAFORM_BEST"
+    description = "BLD_TERRAFORM_BEST_DESC"
+    buildcost =  75 * (1 - 0.4 * Statistic If condition = And [Source EmpireHasAdoptedPolicy empire = Source.Owner name = "PLC_TERRAFORMING"])
```
Grummel7: This one costs only 75% (and 2/3 time) of the targeted version. This is to avoid small wins by "micro-managment"? I think 90% would do for that.

default/scripting/buildings/TERRAFORM.focs.txt
```diff
@@ -84,5 +84,80 @@ TARGET_TERRAFORMING
...
+                    * Target.HabitableSize
+                      // calculating multiple steps: 1+2+..+N == (1+N)*N/2
+                    * max(1, (1 + (PlanetTypeDifference from = Target.OriginalType to = Target.NextBestPlanetType))
+                             * (PlanetTypeDifference from = Target.OriginalType to = Target.NextBestPlanetType) / 2 )
```
Grummel7: That calculation can be quite wrong if the current type is not the original type. Especially it means that reverting to the original type costs one step, no matter how far it is away.

default/scripting/buildings/TERRAFORM.focs.txt
```diff
@@ -84,5 +84,80 @@ TARGET_TERRAFORMING
...
+        Not HasSpecial name = "GAIA_SPECIAL"
+        // Hide building if the species can not improve (Check species' highest available target environment tag)
+        Not HasTag name = "NO_TERRAFORM"
+        Not And [
+            HasTag name = "TERRAFORM_TO_POOR"
+            Or [ Planet environment = Adequate Planet environment = Poor ]
+        ]
+        Not And [
+            HasTag name = "TERRAFORM_TO_ADEQUATE"
+            Or [ Planet environment = Adequate ]
```
Grummel7: Can't you just specify something like:
(PlanetTypeDifference from = LocalCandidate.PlanetType to = LocalCandidate.NextBestPlanetType) > 0

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: there are real wins for micromanaging multisteps, because the population can grow much sooner

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: sounds better

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: one could probably implement multisteps by having an effectsgroup checking if the building is enqueued and how far progress is. if somebody does that, we can go to 90% or maybe even remove
i am not sure how much shortening the buildtime lessens the problem; i would also accept a count-of-step-based buildcost (e.g. 0% discount for 1 step, 10% discount for 2 steps, 21% for 3, 33% for 4, 46% for 5...)

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: reverting is not possible with that building

agrrr3: but i certainly meant the current type, not the original one

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: so id say 10% linear discount, it should only maximum 4 steps anyway.

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: Parse error.  Expected PlanetType expression here:
```
PlanetTypeDifference from = Target.PlanetType to = Target.NextBestPlanetType))
                            ^
```

agrrr3: ah found, parser bug

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: ok, cant do anything with floats there, it seems the parser is broken round(xxx) is not accepted as integer expression

default/scripting/buildings/TERRAFORM.focs.txt
Grummel7: > reverting is not possible with that building

Of course it is, you just have to change the species to one that likes the original type. diff(Current, NextBest) is definitely better.

There is still the following issue, though: for 4 steps cost 10 x base cost, 3 steps cost 6, regardless of the distance form the original type. So if I want to terraform e.g. from Tundra to Swamp, I can use Terrafrom to Desert, followed by Popular Terraforming and its much cheaper then using the Popular Terraforming straight away.

default/scripting/buildings/TERRAFORM.focs.txt
geoffthemedio: not broken, rather a missing feature

default/scripting/buildings/TERRAFORM.focs.txt
```diff
@@ -87,36 +87,33 @@ TARGET_TERRAFORMING
...
         Not HasTag name = "NO_TERRAFORM"
-        Not And [
-            HasTag name = "TERRAFORM_TO_POOR"
-            Or [ Planet environment = Adequate Planet environment = Poor ]
-        ]
-        Not And [
-            HasTag name = "TERRAFORM_TO_ADEQUATE"
-            Or [ Planet environment = Adequate ]
-        ]
+        (0 < PlanetTypeDifference from = LocalCandidate.PlanetType to = LocalCandidate.NextBestPlanetType)
```
geoffthemedio: Thus can probably just be an == test between the next best and current types

agrrr3: Parser sais no after the following change
```
-        (0 < PlanetTypeDifference from = LocalCandidate.PlanetType to = LocalCandidate.NextBestPlanetType)
+        (LocalCandidate.PlanetType == LocalCandidate.NextBestPlanetType)
```
```
default/scripting/buildings/TERRAFORM.focs.txt:885:8: Parse error.  Expected ] here:
        Not Enqueued type = Building name = "BLD_TERRAFORM_INFERNO"
        Not Enqueued type = Building name = "BLD_TERRAFORM_RADIATED"
        Not Enqueued type = Building name = "BLD_TERRAFORM_BARREN"
        // Hide building if the species can not improve (Check species' highest available target environment tag)
        Not HasTag name = "NO_TERRAFORM"
        (LocalCandidate.PlanetType == LocalCandidate.NextBestPlanetType)
        ^
```

agrrr3: probably we dont have comparisons for enum valuerefs? only numerics/real number expressions

agrrr3: if we had arithmetics for enums we could also kick out PlanetTypeDifference everywhere and write instead
`abs(LocalCandidate.PlanetType - LocalCandidate.NextBestPlanetType)`

agrrr3: so for this PR i wontfix this

geoffthemedio: Try with latest master? It should work for those PlanetType valuerefs...

agrrr3: syntax check OK
```
(LocalCandidate.PlanetType == LocalCandidate.NextBestPlanetType)
```

syntax check FAIL (Parse error.  Expected real number expression here):
```
((LocalCandidate.PlanetType) == (LocalCandidate.NextBestPlanetType))
  ^
```

geoffthemedio: Do you particularly like adding extra `()` around sub-expressions without ambiguous operator precedence issues?

Probly could be fixed, but seemingly isn't really necessary...

agrrr3: Yes, this works. Though I am a little uneasy of the class of new bugs being possible (e.g. comparing values of different enum types).

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: ok, i need to check how the costs are actually intended

default/scripting/buildings/TERRAFORM.focs.txt
geoffthemedio: So far only PlanetType enums are supported by the parser.

agrrr3: > Do you particularly like adding extra `()` around sub-expressions without ambiguous operator precedence issues?

yes, thats my personal hobby. 

no, those were left over from a previous more complex calc; i really don't care about those, just wanted to provide feedback to you. also note i did not open an issue.

and in this case it might even be helpful that it does not work, as doing a complicated calculation using enums is probably the wrong thing to do. it will of course only show up sporadically (when it is used where an integer expression is necessary)

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: @Grummel7  could you help me figure out a correct formula? i am pretty much swamped. i guess in the old formula, each step cost was based on the distance to OriginalType. so the base should probably be calculating cost for OriginalType to best type and then subtract the cost of OriginalType to CurrentType. 

Dont know what to do about being on the wrong side of the wheel - e.g. you have been terraforming should it become more expensive?
e.g. original type being ocean, and best type terran (so one step counterclock to best). But you are currently at Radiated (so 4 steps clockwise or 5 steps counterclockwise. Cost of reversions was static at roughly 150% of a single step of terraforming. So going the 5 steps (1.5+1.5+1.5+1.5+1) was cheaper than going the 4 steps (4+3+2+1).

also the with the current terraform* buildings, reversions are not cheaper. so, e.g. transforming to an e.g. Terran planet type is the same cost coming a single step clockwise or counter-clockwise. With current content going the 5 steps would still be cheaper (3+2+1+1+1 vs 4+3+2+1)

So here is the plan: i will not implement cheaper reversion. and the automation should still directly go to terran (clockwise) no matter what the best cost is - popular opinion is not always right and this will be a rare case - you still can manually manage if you think it is doing the wrong thing.

Grummel7: Actually your calculation for the old formula is wrong, it would have been 5+4+3+2, the cost was DistanceFromOriginal + 1. But then the old Reverse Terraforming did not get cheaper by having Policy Terraforming. I think we better forget about that old formulas.

At the moment, with the targeted version its a trade off. Going the direct way costs a sum of 10xbase and takes 4 time slots, while going the long way costs 8xbase and takes 5 time slots, through most of which you get no intermediate improvement. That's fine, I think. There's also a more extreme case: it could also be 4+4+3+2 vs. 2+1+1+1+2...

Popular terraforming should always go the direct way, going away from the preferred type will not be popular with the planets inhabitants.

As for the formula, the more I ponder on it, the more think we should add an extra interface function for it so we can calculate it in C++, where we can simple use a loop to add the step costs. I think it should be possible to do it in FOCS by manually adding 4 values, each but the first multiplied by 1 or 0 based on whether Next(Next) != Next(Next(Next)), etc. I guess you get the point. But we'll end up with one hell of a formula.

We could also use that new interface function to simplify the calculations for targeted terraforming, and, if we want, we can then change the cost so that going back isn't cheaper.

Grummel7: Actually the sum formula cannot be done in FOCS. Planet has NextBetterPlanetType, but that doesn't allow you to get nextBetter(nextBetter), it's always nextBetter(current).

geoffthemedio: The `NextBetterPlanetType` property uses the species that's on the planet to determine what the next better type is, so a free function just taking a single planet type would not be able to determine that. Instead there'd need to be a `ComplexVariable<PlanetType>` that takes a planet type enum and a string for the species to consider, and that could return another planet type and be chained. This would be similar to the case of `ComplexVariable<int>` that takes two planet types and returns their distance.

Grummel7: Well, of course such a function could be added.
My suggestion however is the following:
Add a ComplexInt `TerraformingCostMultiplier planetId=<id> to=<destinationType>`
This would do all those complicated calculation that are in fact not complicated at all when you do them in C++.
E.g. for a Planet with Current = Original = Terran, and to=Inferno it'll return 1+2+3+4=10.
For the way back it would return 3+2+1+1 if we stick to the current rules, but could just as well return 4+3+2+1, if we decide that the way back should cost the same.

agrrr3: TerraformingCostMultiplier is not really acceptable i think as it ties content into the backend.

agrrr3: I see two options currently: one specialised difference function taking three planet types (so distance with a hop in the middle).

Or a difference function with two planet types as parameters, something which is akin to a loop, so you pass the cost function as a valueref and the loop applies the cost function on each step and sums up the values.

universe/Species.h
```diff
@@ -108,6 +108,7 @@ class FO_COMMON_API Species {
     const std::string&              DefaultFocus() const                { return m_default_focus; }         ///< returns the name of the planetary focus this species defaults to. Used for new colonies and uninvaded natives.
     const std::map<PlanetType, PlanetEnvironment>& PlanetEnvironments() const { return m_planet_environments; } ///< returns a map from PlanetType to the PlanetEnvironment this Species has on that PlanetType
     PlanetEnvironment               GetPlanetEnvironment(PlanetType planet_type) const;                     ///< returns the PlanetEnvironment this species has on PlanetType \a planet_type
+    PlanetType                      NextBestPlanetType(PlanetType initial_planet_type) const;             ///< returns the next better PlanetType for this species from the \a initial_planet_type specified
```
geoffthemedio: needs clarification how NextBest and NextBetter are different

geoffthemedio:
-needs rebase
-please remove or at least detangle the unrelated changes
-please consolidate changes and then later fixes or updates to those changes

util/VarText.cpp
```diff
@@ -194,13 +194,30 @@ namespace {
...
+                    try {
+                        auto planet = IApp::GetApp()->GetUniverse().Objects().get<Planet>(boost::lexical_cast<int>(data));
+                        if (planet)
+                            return UserString(to_string(planet->EnvironmentForSpecies()));
+                    } catch (boost::bad_lexical_cast &e) {}
                     return UserString("UNKNOWN_PLANET");
                 }},
             {VarText::USER_STRING_TAG, [](const std::string& data)
                 { return UserString(data); }},
+            {VarText::PLANET_TYPE_TAG, [](const std::string& data)
```
geoffthemedio: after rebasing:

```suggestion
            {VarText::PLANET_TYPE_TAG, [](const std::string& data, const ScriptingContext& context)
```

util/VarText.cpp
```diff
@@ -194,13 +194,30 @@ namespace {
...
+                {
+                    // Assume that we have no userstring which is also a number
+                    if (UserStringExists(data))
+                        return UserString(data);
+                    try {
+                        auto planet = IApp::GetApp()->GetUniverse().Objects().get<Planet>(boost::lexical_cast<int>(data));
```
geoffthemedio: after rebasing
```suggestion
                        auto planet = context.ContextObjects().get<Planet>(boost::lexical_cast<int>(data));
```

universe/Species.h
```diff
@@ -137,6 +138,9 @@ class FO_COMMON_API Species {
 private:
     void Init();
 
+    template <typename Func>
```
geoffthemedio, line 156: need some comment explaining what this does and how the Func param is used

geoffthemedio:
What's with the untitled ship part files?

util/VarText.cpp
```diff
@@ -194,13 +194,30 @@ namespace {
...
             {VarText::ENVIRONMENT_TAG, [](const std::string& data)
                 {
-                    auto planet = IApp::GetApp()->GetUniverse().Objects().get<Planet>(boost::lexical_cast<int>(data));
-                    if (planet)
-                        return UserString(to_string(planet->EnvironmentForSpecies()));
+                    // Assume that we have no userstring which is also a number
+                    if (UserStringExists(data))
+                        return UserString(data);
+                    try {
+                        auto planet = IApp::GetApp()->GetUniverse().Objects().get<Planet>(boost::lexical_cast<int>(data));
```
geoffthemedio: after rebasing, use `context.ContextObjects()`

agrrr3:
> What's with the untitled ship part files?

thanks for catching, did a bad git add

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: This is not really resolved (or it needs a followup). @Grummel7 do you think the implementation in this branch (after rebase) is good enough/better than what we have in master?

Grummel7: I've just merged it to actually try it out. I found two issues:

1. Build time calculation is wrong, since it is done with integers. E.g. 2 * 90 / 100 = 1. Just remove that part, I do not see why popular terraforming should be faster and cheaper.

2. My original point is still open: Your cost multiplier is N*(N+1)/2*(1.1-N/10), e.g. for 3 steps that give 6*0.8 = 4.8. Doing the first step with targeted terraforming, you pay 1 + 3*0.9 = 3.7.
I thought you were looking for a solution for that after your last comment in February. Btw, if you want a specialised difference function, why not using the one I already suggested? Just give it another name e.g. WeightedPlanetTypeDifference

agrrr3: > I've just merged it to actually try it out. I found two issues:
> 
>     1. ...why popular terraforming should be faster and cheaper.

it is the less micro version, it should be obviously better in the normal case

> 
>     2. My original point is still open: Your cost multiplier is N*(N+1)/2*(1.1-N/10), e.g. for 3 steps that give 6_0.8 = 4.8. Doing the first step with targeted terraforming, you pay 1 + 3_0.9 = 3.7.

ah, yes. we could get rid of the complexity by dropping supported use cases. e.g. it only has to be right/cheaper if the planet was not terraformed before (and is more expensive than targeted  terraforming in other cases; by e.g. adding some Max calculation there). in that case using OriginalType and NextBestPlanetType would suffice.

how about 
```
max((PlanetTypeDifference from = Target.OriginalType to = Target.NextBestPlanetType))
                             * (PlanetTypeDifference from = Target.OriginalType to = Target.NextBestPlanetType) / 2 ),
(PlanetTypeDifference from = Target.CurrentType to = Target.NextBestPlanetType))
                             * (PlanetTypeDifference from = Target.CurrentType to = Target.NextBestPlanetType) / 2 ))
```

>        I thought you were looking for a solution for that after your last comment in February. Btw, if you want a specialised difference function, why not using the one I already suggested? Just give it another name e.g. WeightedPlanetTypeDifference

The one you suggested sadly does not suffice. The name is not the problem with your suggested formula. The backend should not know about terraforming costs; that should be scriptable.

i did not find time to implement anything since february, so i expect i wont be able to implement the best solution here in the next half year or so.

Grummel7: > > 1. ...why popular terraforming should be faster and cheaper.
> 
> it is the less micro version, it should be obviously better in the normal case

Then I suggest that you multiply by 6/12 before applying the discount. At the moment, due to integer division, two steps take the same time as one, while three and four step take twice the time of one step.

> ah, yes. we could get rid of the complexity by dropping supported use cases. e.g. it only has to be right/cheaper if the planet was not terraformed before (and is more expensive than targeted terraforming in other cases; by e.g. adding some Max calculation there). in that case using OriginalType and NextBestPlanetType would suffice.
> 
> how about
> 
> ```
> max((PlanetTypeDifference from = Target.OriginalType to = Target.NextBestPlanetType))
>                              * (PlanetTypeDifference from = Target.OriginalType to = Target.NextBestPlanetType) / 2 ),
> (PlanetTypeDifference from = Target.CurrentType to = Target.NextBestPlanetType))
>                              * (PlanetTypeDifference from = Target.CurrentType to = Target.NextBestPlanetType) / 2 ))
> ```

That should do, I think.
Using a macro to make it more readable would probably a good idea.

agrrr3: implemented the max formula and did some testing; looks ok so far; next step clean up

agrrr3: @Grummel7 if you have time i would appreciate if you also test it now and check the two issues are gone

default/scripting/buildings/TERRAFORM.focs.txt
```diff
@@ -84,5 +84,92 @@ TARGET_TERRAFORMING
...
+            scope = Source
+            effects = Destroy
+    ]
+    icon = "icons/building/terraform.png"
+
+
+// @1@ from planet type
+// @2@ to planet type
+AGGREGATED_STEPS_FROM_TO
+'''((PlanetTypeDifference from = @1@ to = @2@) * (PlanetTypeDifference from = @1@ to = @2@) / 2 )'''
```
Grummel7:
```suggestion
'''((1 + PlanetTypeDifference from = @1@ to = @2@) * (PlanetTypeDifference from = @1@ to = @2@) / 2 )'''
```

Grummel7 (commented):
You forgot a +1, but after adding it, it more or less works as intended now. Targeted Terraforming is also nice for creating test cases :)

You don't need the 1 in max, the other two aggregated costs cannot be 0 at the same time. With targeted Terraforming I needed it for the case of someone going back to the original type.

There are some a theoretical case, where popular T. cost only about half of targeted T.: E.g. someone did 4 steps, then the species was replaced by one that likes the type halfway between current and original. In this case, both Original and Current are 2 steps away from Best, but since current is 4 away from Original, Steps with Targeted are more expensive. We could avoid it by adding `AGGREGATED_STEPS_FROM_TO(Target.OriginalType,Target.PlanetType)` to the max. I am undecided on whether this is needed.

default/scripting/buildings/TERRAFORM.focs.txt
agrrr3: ok, i trust you on this one

agrrr3:
> You don't need the 1 in max, the other two aggregated costs cannot be 0 at the same time. With targeted Terraforming I needed it for the case of someone going back to the original type.

lets keep it anyway for robustness; if the planet types are incompatible, the distance is zero.

> 
> There are some a theoretical case, where popular T. cost only about half of targeted T.: E.g. someone did 4 steps, then the species was replaced by one that likes the type halfway between current and original. In this case, both Original and Current are 2 steps away from Best, but since current is 4 away from Original, Steps with Targeted are more expensive. We could avoid it by adding `AGGREGATED_STEPS_FROM_TO(Target.OriginalType,Target.PlanetType)` to the max. I am undecided on whether this is needed.

i think that case is fine. getting the end result is still more expensive then going "directly" there - so the better optimisation is probably not doing the first terraform (and i expect it to be rare). and as said in the beginning, it is ok if popular is cheaper than targeted terraforming.

Grummel7:
> lets keep it anyway for robustness; if the planet types are incompatible, the distance is zero.

Well, the building shouldn't be enqueuable in that case. But keeping it surely breaks nothing... 

> i think that case is fine. getting the end result is still more expensive then going "directly" there - so the better optimisation is probably not doing the first terraform (and i expect it to be rare). and as said in the beginning, it is ok if popular is cheaper than targeted terraforming.

Fine for me, it definitely cannot be abused.

agrrr3:
cleaned up and rebased

agrrr3:
tested after clean up, good to be merged

agrrr3:
tested after clean up, good to be merged

Grummel7:
Why did you add NO_TERRAFORM to Exobots, while removing the only use the Tag had ever had?
Allowing Exobots to terraform a planet was one of the objectives of targeted terraforming.

agrrr3:
> Why did you add NO_TERRAFORM to Exobots, while removing the only use the Tag had ever had? Allowing Exobots to terraform a planet was one of the objectives of targeted terraforming.

exobots are still able to do targeted terraforming with this PR

the point of targeted terraforming was that any species could do it. so NO_TERRAFORM now guards the popular terraformig for those species which are not able to make their environment any better for themselves. we could rename it though if that helps

Grummel7:
> the point of targeted terraforming was that any species could do it. so NO_TERRAFORM now guards the popular terraformig for those species which are not able to make their environment any better for themselves. we could rename it though if that helps

If a species cannot improve its environment, `(LocalCandidate.PlanetType != LocalCandidate.NextBestPlanetType)` should stop them from using popular terraforming, so if you want to allow even Sly to use targeted terraforming, I think the tag is simply no longer needed.

agrrr3:
> I think the tag is simply no longer needed.

strictly speaking this is true; it is a computational shortcut though (maybe premature)

Grummel7:
> > I think the tag is simply no longer needed.
> 
> strictly speaking this is true; it is a computational shortcut though (maybe premature)

Well, it surely won't break anything. If you want to keep it, than keep it.
We should finally get this PR merged.

agrrr3:
>  We should finally get this PR merged.

+1 

@geoffthemedio could you merge please?

PR #3708, opened 2022-02-17
Hack to allow multi-line commands in debug chat

Grummel7:
Simple, but effective.

Allows entering loops like:
`for p in e.availablePolicies: print(p);`
Or even nested loops like:
`for cat in e.categoriesSlotPolicies:; print(cat.key()); for slot in cat.data(): print(' ', slot);`

### DIFF
--- a/default/python/AI/freeorion_tools/chat_handler/interpreter.py
+++ b/default/python/AI/freeorion_tools/chat_handler/interpreter.py
@@ -18,6 +18,8 @@ class DebugInterpreter:
 
         sys.stdout = StringIO()
         sys.stderr = StringIO()
+        if msg.endswith(";"):
+            msg = msg.replace(";", "\n")
         handler = logging.StreamHandler(sys.stdout)
         logging.getLogger().addHandler(handler)
         self._interpreter.runsource(msg)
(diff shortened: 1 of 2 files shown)


default/python/AI/freeorion_tools/chat_handler/interpreter.py
```diff
@@ -18,6 +18,8 @@ def eval(self, msg):
 
         sys.stdout = StringIO()
         sys.stderr = StringIO()
+        if msg.endswith(';'):
+            msg = msg.replace(';', '\n')
```
geoffthemedio: This seems potentially risky, but since it's effectively opt-in, I suppose it's fine. If @Morlic-fo or @Cjkjvfnby object it can be reverted...

geoffthemedio: I'm reluctant to merge if the black linter is throwing an error, even though there's no indication what the problem is...

Grummel7: Just installed black. That tool isn't very talkative:
```
> black AI/freeorion_tools/chat_handler/interpreter.py --verbose
Using configuration from /home/muzz/freeorion-project/Grummel7/pyproject.toml.
reformatted AI/freeorion_tools/chat_handler/interpreter.py
All done! ✨ 🍰 ✨
1 file reformatted.
```

It didn't like the single-quotes. I will update the readme. It already says that you should use flake8, but black it not mentioned.

Grummel7:
Btw. I've figured out that ai_config works in multiplayer-mode, if I specify an absolute path. I have not yet figured out where it changes the directory. Perhaps the game could use std::filesystem::canonical on the supplied file name.

PR #3710, opened 2022-02-18
fix bug where policy icons would change size on switching category

TheSilentOne1:


Grummel7:
The fix works for me.

PR #3712, opened 2022-02-20
no more "Deep Space" in sitrep --> use normal names plus suffix "(empty)" instead

st-pa-fo:
makes it easier to distinguish all the deep spaces, all the depths of space and all the deepnesses upon the dark. expecially when there are lots of them. see screenshot attached.

deep space id's

since i am new to cpp and not really fluent in its ins and outs, i'm open to suggestions how to improve the implementation. e.g. i found no quick way to keep the & and make it work, compiler always complained about 'Reference to stack memory associated with local variable returned' ..


o01eg:
As I know they have real names assigned to them. Maybe just not suppress it.

geoffthemedio:
Returning `const std::string&` means you're claiming to return an address that points to a string in memory that will persist and can be used some time later. Your implementation creates a new string on the stack and programmatically fills that string with stuff, but that string will be destructed / become invalid when the function returns, meaning the address you return will point to a destructed object. If you want to do something like this, do it to the return value at the call site of a function like this, not within that function, eg. https://github.com/freeorion/freeorion/blob/master/UI/SystemIcon.cpp#L209

Also, it would need to be controlled with an option, like the existing "ui.name.id.shown" (which is controlled in the GUI at the bottom of the Options window UI tab, there is an "IDs after names" option), which applies to everything, not just systems.

st-pa-fo:
there is not necessarily a relation to the "show all IDs" option, i tried galactic coordinates as well but that got too long, it could also be some short random hex numbering scheme remeniscent of some stellar catalogue numbers. with the main purpose of making empty spaces distinguishable from one another in the sitrep.

> meaning the address you return will point to a destructed object

this is why i am not returning an address to something (which would be the and operator) but instead the something itself.
seemed at least to build and run for a turn, long enough to make that screenshot, which i assume would not have been possible, if the constructed names were already removed from memory.

st-pa-fo:
> As I know they have real names assigned to them. Maybe just not suppress it.

"Deep Space" as name is fine, does not really need a fullblown name, only needs to be a little different from the next "Deep Space" and the next "Deep Space" when they appear close to each other in sitrep so the player does not get disoriented as easily from scrolling through a list of seemingly identical deep spaces. introducing a column with line numbers to the sitrep instead might be another solution to the percieved problem, but this here seemed to be the easier path.  (edit; typo)

st-pa-fo:
> do it to the return value at the call site of a function

i am quite sure i do not understand that correctly. because i doubt you want me to do the checking for empty systems at those nine call sites instead of in the one single function that is called, and do the exact same string operations in nine repeated places instead of the one.

ApparentName is used nine times

geoffthemedio:
> > meaning the address you return will point to a destructed object
> 
> this is why i am not returning an address to something (which would be the and operator) but instead the something itself.

I was explaining why you were getting errors returning a function-local temporary as a const reference before you switched to by-value return.

agrrr3:
Showing IDs can be long and not very helpful (system ids are quite spread and non-local).

As a default, I suggest showing "Deep Space (<real name>)" in the sitrep messages, if the global show id option is off. 
Usually that real name already includes a location information (the ~cluster name). Writing out "Deep Space" explains why you cant see it on the map.

If the global show id option is on, i would not show the real name (only the id) like it is done on the map.

universe/System.cpp
```diff
@@ -195,13 +195,13 @@ const std::string System::ApparentName(int empire_id, const Universe& u,
 
         if (m_star == StarType::INVALID_STAR_TYPE) {
             std::stringstream s;
-            s << UserString("UNEXPLORED_REGION") << " #" << std::fixed << std::setprecision(0) << ID();
+            s << UserString("UNEXPLORED_REGION") << " #" << std::to_string(ID());
```
geoffthemedio: use `std::string` instead of `std::stringstream`, call `.reserve(...)` with a suitable size first, and then use `+=` or `.append(...)` to avoid extra allocations.

universe/System.h
```diff
@@ -73,7 +73,7 @@ class FO_COMMON_API System : public UniverseObject {
       * yet explored might be "Unexplored Region", rather than an empty string
       * for the name.  This is distinct from PublicName functions, which filter
       * the name based on ownership. */
-    [[nodiscard]] const std::string&      ApparentName(int empire_id, const Universe& u, bool blank_unexplored_and_none = false) const;
+    [[nodiscard]] const std::string       ApparentName(int empire_id, const Universe& u, bool blank_unexplored_and_none = false) const;
```
geoffthemedio: If making it return by-value, get rid of the `const`.

Also need to check all the places `ApparentName` is called to store the result by value and not const reference.

universe/System.cpp
st-pa-fo: acknowledged. will do.

geoffthemedio: https://quick-bench.com/q/-h8v4haY3lbY-EGQMNExY-2ohLU

o01eg:
Maybe we should just assign required UniverseObject::m_name in universe creator instead of any logic?

geoffthemedio:
> Maybe we should just assign required UniverseObject::m_name in universe creator instead of any logic?

I believe empty systems actually do get assigned a name, which is a normal system name like **Klendathu β**, but it just shows **Deep Space** in the GUI when there is no star. If a planet later gets a star, such as by a Nebula collapsing, it then displays the name it had all along.

Oberlus:
Maybe show it like "Klendathu β empty space"? Too long...
"Klendathu β (empty)" ?

st-pa-fo:
i've stuck to the previous idea with just a number, but no longer using the universe-id but instead a dedicated system counter variable so that star systems get a number from zero up to the number of stars, stored as 'catalogue name', and used for empty space, unexplored region and unexplored system. looks pretty similar to the first commit in the attached screenshot, but numbers are now not as semi-random as universe-id was and do not get too long. i am sufficiently happy with how this looks, altho smth like 'CXFG-2471' .. 'KQRF-7887' .. 'TZPG-4012' would be nicer to look at but more work to get it right so going for minimally sufficient which is voila.

deep space numbered catalogue names

agrrr3:
> i've stuck to the previous idea with just a number, but no longer using the universe-id but instead a dedicated system counter variable so that star systems

is this number permanent? else i would rather prefer the system name to be used. to be more consistent we could show the system name also in other contexts instead of "empty system" (i.e. system panel)

st-pa-fo:
> > i've stuck to the previous idea with just a number, but no longer using the universe-id but instead a dedicated system counter variable so that star systems
> 
> is this number permanent? else i would rather prefer the system name to be used. to be more consistent we could show the system name also in other contexts instead of "empty system" (i.e. system panel)

yes permanent. stays the same between turns, stays the same after saving closing opening loading game.

Oberlus:
And permanent when an empty space gets a new star or a planet is brought from another system?

geoffthemedio:
In the call a couple hours ago, we discussed just showing the real system name, when known, and dropping the "Deep Space" label. The real name actually shows in the objects list already...

st-pa-fo:
implemented the suggested change. see screenshot attached. also rebased on master. when all feedbacks are accounted for, can compress commits for clearer message.

empty systems can have names now

st-pa-fo:
commits are squashed. edit; i messed up the message though ...

default/stringtables/en.txt
```diff
@@ -204,13 +204,13 @@ NEW_ASTEROIDS_SUFFIX
 [[PT_ASTEROIDS]]
 
 EMPTY_SPACE
-Deep Space
+''' (empty)'''
```
geoffthemedio: This is possibly misleading... there could be ships in the system, and possibly planets even with no star. What is shown in those cases?

universe/System.cpp
```diff
@@ -194,10 +193,17 @@ const std::string& System::ApparentName(int empire_id, const Universe& u,
         if (blank_unexplored_and_none)
             return EMPTY_STRING;
 
-        if (m_star == StarType::INVALID_STAR_TYPE)
-            return UserString("UNEXPLORED_REGION");
-        else
-            return UserString("UNEXPLORED_SYSTEM");
+        if (m_star == StarType::INVALID_STAR_TYPE) {
+            std::string s;
+            s.append(m_name).append(UserString("UNEXPLORED_REGION"));
```
geoffthemedio: for just two parts, you can just do return `m_name + UserString(...)`

`.append(...)` or `+=` are more useful when there are numerous strings added to the existing string, where chaining `+` would lead to an allocation for each

default/stringtables/en.txt
st-pa-fo: right, the same. so - change to " (no star)" instead?

universe/System.cpp
st-pa-fo: aye

default/stringtables/en.txt
st-pa-fo: wait a moment .. no, normal pre-change behaviour should apply, so only the really empty systems that show no name on the map will get the suffix and it will not show up in starless systems that were not called Deep Space before. still, might be fleets there so still somewhat misleading and still i'll put the no-star-suffix in next. and still not hewn in stone.

st-pa-fo: did not find a starless planet right now.

empty systems 'no star' no star

geoffthemedio:
I feel like it would be better without the **(no star)** tag in the sitrep and sidepanel. If the startype is important for the sitrep, then it could be shown for all systems. For the sidepanel, the star type is already shown just below the system name.
NoStar

Also, if you could combine all the commits other than the override one, that would be preferable.

st-pa-fo:
so back to Deep Space then?

geoffthemedio:
> so back to Deep Space then?

No, just show the system name without any annotation, since there's already an indication of the star type right below on the sidepanel, and it looks weird in the sitrep. Or does that not accomplish what you wanted?

Vezzra:
@st-pa-fo, closed as invalid, abandoned, superseded...?

geoffthemedio:
> @st-pa-fo, closed as invalid, abandoned, superseded...?

seemingly as frustrated...

more-or-less merged as https://github.com/freeorion/freeorion/commit/f07d589e18bb681ddc5f07b62fafd372332b3580

PR #3716, opened 2022-02-22
[AI] Update python skeletons

Cjkjvfnby:


Grummel7:
Have you updated to master before regenerating the interfaces?
At least this function `def adoptionCost(self, number: int, obj: object) -> float: ...` is outdated.

See https://www.freeorion.org/forum/viewtopic.php?f=9&t=12274

Cjkjvfnby:
> Have you updated to master before regenerating the interfaces?

No, I use windows builds.  So new changes come with delay.  
I'll merge this PR and will update stubs to the recent changes in the next one.

Issue #3718, opened 2022-02-23
When an AI civilisation has ships, but no planets, message log shows big error every turn.


psypherium:
FreeOrion built from git on Arch Linux, AMD drivers.

When an AI civilisation has ships, but no planets, message log shows big error every turn.

As soon as their final planet was taken over, the messages started to appear. When the enemy ships were all destroyed, the error no longer appeared.

[23 Feb 05:01:37] AI_4: AI_Error: AI script error : FreeOrionAI.py:generateOrders():365  - Exception cannot unpack non-iterable NoneType object while trying to calculate_priorities
Traceback (most recent call last):
  File "/home/alistair/Games/freeorion/freeorion/default/python/AI/FreeOrionAI.py", line 362, in generateOrders
    action()
  File "/home/alistair/Games/freeorion/freeorion/default/python/AI/PriorityAI.py", line 65, in calculate_priorities
    InvasionAI.get_invasion_fleets()  # sets AIstate.opponentPlanetIDs, and AIstate.invasionTargetedPlanetIDs
  File "/home/alistair/Games/freeorion/freeorion/default/python/AI/InvasionAI.py", line 226, in get_invasion_fleets
    evaluated_planets = assign_invasion_values(evaluated_planet_ids)
  File "/home/alistair/Games/freeorion/freeorion/default/python/AI/InvasionAI.py", line 285, in assign_invasion_values
    planet_values[pid] = neighbor_values.setdefault(pid, evaluate_invasion_planet(pid))
  File "/home/alistair/Games/freeorion/freeorion/default/python/AI/InvasionAI.py", line 447, in evaluate_invasion_planet
    least_jumps_path, max_jumps = _get_path_from_capital(planet)
TypeError: cannot unpack non-iterable NoneType object
[...]

PR #3719, opened 2022-02-25
Fixed some AI exceptions

Grummel7:
This should fix the exceptions mentioned in #3718 

Btw, can anybody tell me, whether old AIFleetMission.py line 608 makes sense:
`if assertion_fails(new_fleet_id and new_fleet_id != INVALID_ID):`
Common sense says checking for INVALID_ID should do.

### DIFF
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -596,8 +596,9 @@ class AIFleetMission:
                             )
                             allowance = CombatRatingsAI.rating_needed(fleet_remaining_rating, fleet_min_rating)
                             debug("May release ships with total rating of %.1f", allowance)
-                            ship_ids = list(self.fleet.get_object().shipIDs)
-                            for ship_id in ship_ids:
+                            for ship_id in self.fleet.get_object().shipIDs:
+                                if len(self.fleet.get_object().shipIDs) == 1:
+                                    break
                                 ship_rating = CombatRatingsAI.get_ship_rating(ship_id)
                                 debug("Considering to release ship %d with rating %.1f", ship_id, ship_rating)
                                 if ship_rating > allowance:
@@ -605,7 +606,7 @@ class AIFleetMission:
                                     continue
                                 debug("Splitting from fleet.")
                                 new_fleet_id = FleetUtilsAI.split_ship_from_fleet(fleet_id, ship_id)
-                                if assertion_fails(new_fleet_id and new_fleet_id != INVALID_ID):
+                                if assertion_fails(new_fleet_id not in (None, INVALID_ID)):
                                     break
                                 new_fleets.append(new_fleet_id)
                                 fleet_remaining_rating = CombatRatingsAI.rating_difference(
--- a/default/python/AI/InvasionAI.py
+++ b/default/python/AI/InvasionAI.py
@@ -326,15 +326,14 @@ def _get_path_from_capital(planet: "fo.planet") -> Tuple[Sequence[SystemId], int
     If there is no path, return empty sequence and default distance.
     """
     universe = fo.getUniverse()
-    capitol_id = PlanetUtilsAI.get_capital()
-    if capitol_id:
-        homeworld = universe.getPlanet(capitol_id)
+    capital_id = PlanetUtilsAI.get_capital()
+    if capital_id != INVALID_ID:
+        homeworld = universe.getPlanet(capital_id)
         if homeworld and homeworld.systemID != INVALID_ID and planet.systemID != INVALID_ID:
             least_jumps_path = list(universe.leastJumpsPath(homeworld.systemID, planet.systemID, fo.empireID()))
             max_jumps = len(least_jumps_path)
             return least_jumps_path, max_jumps
     else:
-
         return [], 8
 
 
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -135,6 +135,7 @@ def generate_production_orders():
     building_ratio = aistate.character.preferred_building_ratio([0.4, 0.35, 0.30])
     capital_id, homeworld, capital_system_id = _get_capital_info()
     current_turn = fo.currentTurn()
+    production_queue = empire.productionQueue
     if not homeworld:
         debug("if no capital, no place to build, should get around to capturing or colonizing a new one")  # TODO
     else:
@@ -178,7 +179,6 @@ def generate_production_orders():
 
             debug("")
             debug("Buildings already in Production Queue:")
-            production_queue = empire.productionQueue
             capital_queued_buildings = []
             for element in [e for e in production_queue if (e.buildType == EmpireProductionTypes.BT_BUILDING)]:
                 building_expense += element.allocation


agrrr3:
> Btw, can anybody tell me, whether old AIFleetMission.py line 608 makes sense: `if assertion_fails(new_fleet_id and new_fleet_id != INVALID_ID):` Common sense says checking for INVALID_ID should do.

not sure what you mean with common sense here. 
split_ship_from_fleet returns without values in some places, so the check is necessary

Grummel7:
> not sure what you mean with common sense here. split_ship_from_fleet returns without values in some places, so the check is necessary

Well, my "common sense" thinking was that if there is a constant named INVALID_ID (-1), then 0 should be a valid fleet id and the assertion is bad. May not be a issue, I guess a split off fleet will never get id 0, but the code shouldn't rely on such an implementation detail.

Indeed I missed the fact that fleet_id could also be None, but then the check should probably better written as:
`if assertion_fails(new_fleet_id != None and new_fleet_id != INVALID_ID):`

default/python/AI/InvasionAI.py
```diff
@@ -327,14 +327,13 @@ def _get_path_from_capital(planet: "fo.planet") -> Tuple[Sequence[SystemId], int
     """
     universe = fo.getUniverse()
     capitol_id = PlanetUtilsAI.get_capital()
-    if capitol_id:
+    if capitol_id != INVALID_ID:
```
geoffthemedio: don't also need to test for `None`?

geoffthemedio:
>... the check should probably better written as: `if assertion_fails(new_fleet_id != None and new_fleet_id != INVALID_ID):`

I'm informed that `if assertion_fails(new_fleet_id and new_fleet_id != INVALID_ID):` would be the stanard way to write that...

default/python/AI/InvasionAI.py
Grummel7: Argh, these checks are a pain in the ass, some functions return None, some INVALID_ID, perhaps some do both. At the moment get_capital() does not return None, but I agree it's safer to check both.

And there is still the question whether 0 is a valid id, because the "standard check" `if id and id != INVALID_ID` also yields false if id is 0

geoffthemedio: There won't normally be a planet with id 0. Might be a system 0, though.

geoffthemedio:
> > ... the check should probably better written as: `if assertion_fails(new_fleet_id != None and new_fleet_id != INVALID_ID):`
> 
> I'm informed that `if assertion_fails(new_fleet_id and new_fleet_id != INVALID_ID):` would be the stanard way to write that...

after further discussion, according to PEP8, it should be `if assertion_fails(new_fleet_id is not None and new_fleet_id != INVALID_ID):`

Cjkjvfnby:
> after further discussion, according to PEP8, it should be if assertion_fails(new_fleet_id is not None and new_fleet_id != INVALID_ID):


I'd say if our linter does not catch it, its OK. 

I'd prefer something like `new_fleet_id  not in (None, INVALID_ID)`

default/python/AI/AIFleetMission.py
```diff
@@ -596,16 +596,19 @@ def issue_fleet_orders(self):
                             )
                             allowance = CombatRatingsAI.rating_needed(fleet_remaining_rating, fleet_min_rating)
                             debug("May release ships with total rating of %.1f", allowance)
-                            ship_ids = list(self.fleet.get_object().shipIDs)
+                            org_fleet = self.fleet.get_object().shipIDs
```
Cjkjvfnby: `org_fleet` is a bit confusing name, since it collection of ships.  

`for ... in ...` and `list()` uses the same things inside, which implies that we could just skip wrapping to list. 

```python
for ship_id  in self.fleet.get_object().shipIDs:
```

default/python/AI/InvasionAI.py
```diff
@@ -327,14 +327,13 @@ def _get_path_from_capital(planet: "fo.planet") -> Tuple[Sequence[SystemId], int
     """
     universe = fo.getUniverse()
     capitol_id = PlanetUtilsAI.get_capital()
-    if capitol_id:
+    if capitol_id and capitol_id != INVALID_ID:
```
Cjkjvfnby: `PlanetUtilsAI.get_capital()` will return id or INVALID_ID, so we could drop` capitol_id and` part

default/python/AI/ProductionAI.py
```diff
@@ -135,6 +135,7 @@ def generate_production_orders():
     building_ratio = aistate.character.preferred_building_ratio([0.4, 0.35, 0.30])
     capital_id, homeworld, capital_system_id = _get_capital_info()
     current_turn = fo.currentTurn()
+    production_queue = empire.productionQueue
```
Cjkjvfnby, line 138: I'd put it inside else branch.

Cjkjvfnby (approved):
Looks good. I left a couple of minor comments.

Grummel7:
> `org_fleet` is a bit confusing name, since it collection of ships.

Perhaps I should simply work without an extra name, it is used only once anyway.

> `for ... in ...` and `list()` uses the same things inside, which implies that we could just skip wrapping to list.
> 
> ```python
> for ship_id  in self.fleet.get_object().shipIDs:
> ```

As far as I understand it, the list() wrapper is a way to create a copy, it was there before. Whether the copy is necessary is beyond my python expertise. The problem is that self.fleet is modified within the loop.

> I'd prefer something like `new_fleet_id not in (None, INVALID_ID)`

I like that one, could use it for capital case below as well.

> `PlanetUtilsAI.get_capital()` will return id or INVALID_ID, so we could drop `capitol_id and` part

Originally I only checked for INVALID_ID, but Geoff complained about it. And I think he's right, in doubt we should rather check both values, since it is less likely to be affected by future changes in other parts of the code.

> I'd put it inside else branch.

That's were it was, but the name is accessed after the if...else, that's what I put it outside.

Cjkjvfnby:
@Grummel7 It will be much easier if you will pull answers in the comment thread.  In that case, I can just resolve them.  Also, they will be close to the code we are talking about.

> As far as I understand it, the list() wrapper is a way to create a copy, it was there before. 
Whether the copy is necessary is beyond my python expertise. The problem is that self.fleet is modified within the loop.

> it was there before

In this codebase, it's not a good argument. For many reasons, a lot of strange things are here. Keeping them does not make code better. 

> The problem is that self.fleet is modified within the loop.

We use `fleet.get_object().shipIDs` which is already a copy. Not critical for that PR

> Originally I only checked for INVALID_ID, but Geoff complained about it. And I think he's right, in doubt we should rather check both values, since it is less likely to be affected by future changes in other parts of the code.

`PlanetUtilsAI.get_capital()` never returns None. And changing it will require checking all usages. I don't expect this change will ever happen, so we could keep things simple for that case. Instead of that other places that return INVALID_ID or None, should be changed to return only INVALID id, but this is out of the scope of this PR.

> That's were it was, but the name is accessed after the if...else, that's what I put it outside.

Now I see it.  Big functions are a bad thing.

default/python/AI/InvasionAI.py
Grummel7: Actually yes, there is a system with ID 0. Just had a test game in which one AI started in it. It never left the system. Haven't yet figured out why. Log says:

`21:08:28.839817 {0x00007fa656944780} [warn] python : target.py:19 : Target is invalid inaccessible system_-1<>`
`21:08:28.839890 {0x00007fa656944780} [debug] python : MoveUtilsAI.py:51 : Requesting path for fleet F_3902<Recon Fleet 3902> from inaccessible sys>`
`21:08:28.839993 {0x00007fa656944780} [warn] python : pathfinding.py:204 : Requested path between invalid systems.`

But this is a separate issue...

Grummel7:
> @Grummel7 It will be much easier if you will pull answers in the comment thread. In that case, I can just resolve them. Also, they will be close to the code we are talking about.

Okay, will do next time. There'll surely be bigger PRs

> We use `fleet.get_object().shipIDs` which is already a copy. Not critical for that PR

Actually, if it is a copy, my length check works on a copy, making it pointless.
But I guess you are right, since the object comes straight from C++.
So I simplified it by not using extra names at all.

> `PlanetUtilsAI.get_capital()` never returns None. And changing it will require checking all usages. I don't expect this change will ever happen, so we could keep things simple for that case. Instead of that other places that return INVALID_ID or None, should be changed to return only INVALID id, but this is out of the scope of this PR.

Well, I agree on that as well, so I changed it back.

geoffthemedio:
I'm not clear about the status of this, but @Cjkjvfnby can you merge when it's ready?

Cjkjvfnby (approved):

geoffthemedio:
I'm getting some AI errors after loading a save, which are plausibly related to these changes...
````
[05 Mar 23:50:56] AI_2: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_186269<Fleet 186269> has no valid system.
[05 Mar 23:50:56] AI_1: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_183699<Battle Fleet 183699> has no valid system.
[05 Mar 23:50:56] AI_3: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_186207<Battle Fleet 186207> has no valid system.
[05 Mar 23:50:56] AI_2: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_186819<Battle Fleet 186819> has no valid system.
[05 Mar 23:50:56] AI_1: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_186764<Battle Fleet 186764> has no valid system.
[05 Mar 23:50:56] AI_3: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_188091<Fleet 188091> has no valid system.
[05 Mar 23:50:56] AI_2: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_189108<Battle Fleet 189108> has no valid system.
[05 Mar 23:50:56] AI_1: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_189053<Battle Fleet 189053> has no valid system.
[05 Mar 23:50:56] AI_3: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_188494<Battle Fleet 188494> has no valid system.
[05 Mar 23:50:56] AI_3: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_188660<Fleet 188660> has no valid system.
[05 Mar 23:50:56] AI_3: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_188671<Fleet 188671> has no valid system.
````

Grummel7:
> I'm getting some AI errors after loading a save, which are plausibly related to these changes...
> 
> ```
> [05 Mar 23:50:56] AI_2: AI_Error: AI script error : AIstate.py:__clean_fleet_roles():938  - Fleet F_186269<Fleet 186269> has no valid system.
> ...
> ```

This happened with current master? I do not see how this PR should affect fleet positions.
Or have you merged #3723 in your environment? That one at least does change code related to fleet positions.

geoffthemedio:
I thought it was master, but I'm not able to reproduce the problem now...

PR #3720, opened 2022-02-27
Balance colonization and centralization

Oberlus:
- Colonization and Centralization cost factors apply to colony buildings and outpost ship parts (https://www.freeorion.org/forum/viewtopic.php?p=110448#p110448)
- Exobot colony cost decreased from 70 to 60 (there's been discussion on whether Exobot should still cost more than regular species, this is something in between).
- Colonization and Centralization cost factors decreased from 0.5 to to 1/3 (50% of a colony ship is comparable to +~80% PP per turn early game; 33% is probably still OP, but this is at least a step in the good direction and will be good to have feedback).

### DIFF
--- a/default/scripting/buildings/colonies/SP_ABADDONI.focs.txt
+++ b/default/scripting/buildings/colonies/SP_ABADDONI.focs.txt
@@ -3,7 +3,7 @@
 BuildingType
     name = "BLD_COL_ABADDONI"
     description = "BLD_COL_ABADDONI_DESC"
-    buildcost = 50 * [[COLONY_UPKEEP_MULTIPLICATOR]] * [[BUILDING_COST_MULTIPLIER]]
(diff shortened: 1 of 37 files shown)


geoffthemedio:
> 50% of a colony ship is comparable to +~80% PP per turn early game

I think the policy only affects the cost of the colony pod, not the whole ship...

Oberlus:
> > 50% of a colony ship is comparable to +~80% PP per turn early game
> 
> I think the policy only affects the cost of the colony pod, not the whole ship...

Yes, that's why I said +80% and not +100%. Hulls cost much less than colo pods.

Medium basic hull: 20 PP
Organic hull: 8 PP
Colony pod: 120 PP

Basic colo ship with / without Colonization: 80 / 140 (so you save 43% of each 140 PP, equivalent to having +75% PP when building colo ships).
Organic colo ship: 68 / 128 (+88% PP when building colo ships).

Oberlus:
https://freeorion.org/forum/viewtopic.php?f=15&t=12312

PR #3723, opened 2022-03-04
More basic AI stuff

Grummel7:
No I have figured out why the AI cannot handle a system with ID 0.
It also was unable to detect how bad Exobots have become, since the tag parser couldn't handle VERY_BAD.


default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -37,11 +37,13 @@ def get_ai_tag_grade(tag_list, tag_type):
     Checks for the first tag in the list (if any), for tag_type "TYPE",
     having the structure X_TYPE
     and then returns 'X'
-    X is most commonly (but not necessarily) one of [NO, BAD, AVERAGE, GOOD, GREAT, ULTIMATE]
+    X is most commonly (but not necessarily) one of [NO, VERY_BAD, BAD, AVERAGE, GOOD, GREAT, ULTIMATE]
     If no matching tags, returns empty string (which for most types should be considered equivalent to AVERAGE)
     """
     for tag in [tag_ for tag_ in tag_list if tag_.count("_") > 0]:
-        parts = tag.split("_", 1)
+        parts = tag.split("_", 2)
```
Cjkjvfnby: Looks like our tag name convention does not support machine parcing. 

Changing tags does not look like an easy task.  So we need a hack. 


```python
    for tag in [tag_ for tag_ in tag_list if tag_.count("_") > 0]:
        # special case when tag modifier has underscore in name
        if tag.startswith(f"VERY_BAD_{tag_type.upper()}"):
            return "VERY_BAD"
        else:
            parts = tag.split("_", 1)
            if parts[1] == tag_type.upper():
                return parts[0]
    return ""
```

And don't forget to add the test.

```python
def test_very_bad_tag_parsing_find_correct_modifier():
    assert "VERY_BAD" == get_ai_tag_grade(["VERY_BAD_WEAPONS"], "WEAPONS")
```

Grummel7: I thought about that way, too, but there is now also EXTREMELY_BAD_POPULATION. On the other hand, the second part the AI is interested in only consists of single words, that's why I did it that way.
I wasn't aware of the tests, will have to fix it, of course.

Grummel7: Actually I was wrong, there are e.g. BAD_DEFENSIVE_TROOPS.
How about:
```
        parts = tag.split("_", 1)
        if parts[1] == "BAD_":
            parts = [parts[0] + "_BAD", parts[4:]]
```

Cjkjvfnby: Please add a test I have mentioned.

> On the other hand, the second part the AI is interested in only consists of single words,

Sounds like a plan.  Lets use `rsplit()`.

PS. It would be nice to add some checks for the `Tags` class that all attributes do not have any dashes. I guess enums or metaclasses could be the solution. Do you want to try it yourself?

Cjkjvfnby (approved):

PR #3725, opened 2022-03-05
Aligned species tags with effects and provide a checking tool

Grummel7:
I found that the fact that the AI couldn't handle VERY_BAD tags was not the only reason, it had a lot of wrong ideas about species.


check/checkSpeciesTags.py
```diff
@@ -0,0 +1,106 @@
...
+    for focus in foci_list:
+        if ("NO_" + focus) not in alltags:
+            print("%s: Not HAS_%s_FOCUS, NO_%s missing in tags" % (file.name, focus, focus))
+    if tags:
+        for tag in tags:
+            print("%s: %s in tags, missing in effectgroups" % (file.name, tag))
+
+
+if __name__ == "__main__":
+    if len(argv) != 2:
```
Cjkjvfnby, line 98: I'd recommend using https://docs.python.org/3/library/argparse.html.

check/checkSpeciesTags.py
```diff
@@ -0,0 +1,106 @@
...
+foci_to_check = [
+    "INDUSTRY",
+    "RESEARCH",
+    "INFLUENCE",
+]
+
+
+def check_skill(name):
+    if (
+        name.startswith("AVERAGE")
```
Cjkjvfnby: https://docs.python.org/3/library/stdtypes.html?highlight=str%20startswith#str.startswith

> prefix can also be a tuple of prefixes to look for. 

You could simplify this code.

check/checkSpeciesTags.py
```diff
@@ -0,0 +1,106 @@
...
+            return skill
+    return ""
+
+
+def check_species(file):
+    effects = False
+    foci = False
+    tags = set()
+    foci_list = set(foci_to_check)
+    with open(file, "r", encoding="utf-8") as f:
```
Cjkjvfnby, line 57: I'll try to look into this tomorrow. 

I like the state machine approach to solving such kinds of tasks.

check/checkSpeciesTags.py
Grummel7: As long as I only need exactly one value, having no options, this seemed to be the simplest solution.

check/checkSpeciesTags.py
Grummel7: You live and learn... :)
I will update it.

Grummel7:
Actually, it occurred to me that we should evaluate another way to solve the issue:
Pedia articles are based on description values of the effect groups.
Can we allow the AI to access these effects groups as well? Then we could get rid of the redundancy and do not need an additional check to keep redundant information consistent.

check/checkSpeciesTags.py
Cjkjvfnby: Maybe a few more lines of code, but it will greatly improve readability. Still ok to leave it as is.

Cjkjvfnby:
> Actually, it occurred to me that we should evaluate another way to solve the issue:
Pedia articles are based on description values of the effect groups.
Can we allow the AI to access these effects groups as well? Then we could get rid of the redundancy and do not need an additional check to keep redundant information consistent.

@geoffthemedio is it hard to implement this?

geoffthemedio:
> @geoffthemedio is it hard to implement this?

Assuming it means giving the AI access to named valuerefs from scripts, no, it shouldn't be hard.

@agrrr3 FYI

Grummel7:
> > @geoffthemedio is it hard to implement this?
> 
> Assuming it means giving the AI access to named valuerefs from scripts, no, it shouldn't be hard.
> 
> @agrrr3 FYI

We definitely need the valuerefs, but here I am talking about the effects. e.g.:
```
VERY_BAD_RESEARCH
'''[[BASIC_RESEARCH]]

        EffectsGroup
            description = "VERY_BAD_RESEARCH_DESC"
            scope = Source
            activation ...
```
Pedia uses the description attribute and python could use it as well to avoid to redundantly defining the same information in the tags.

geoffthemedio:
Internal content implementation classes like EffectsGroups that various content has could probably also be exposed, but I don't like the idea of using something inteded for UI display purposes to guide AI decisions...

Grummel7:
Well, we can stick to the PR.
CheckSpeciesTags should then be added to one of the Github hooks. That would ensure that the information the AI gets about species abilities stays aligned with what is implemented.

Cjkjvfnby:
My conclusion:  I am too lazy to dive deep into that script.  But this should not be a problem since we can treat the script by its results.  Also, it's pretty isolated from the codebase.  Probably we will not need it after migration to PythonScripts. 

I am not competent in the  AI script so if someone could review them it would be great.

Looks like it would be a good strategy to merge this script and changes in tags and polish them later.

I'll take care of adding a CI check after this PR will be merged. (I might forget so feel free to ping me.)

geoffthemedio:
I'm unclear about the status of thing... Some stuff can be done in future, but is this ready and good or still a work in progress or flawed...?

Grummel7:
> I'm unclear about the status of thing... Some stuff can be done in future, but is this ready and good or still a work in progress or flawed...?

I think it is finished work. I've added a checker and used it to fix some inconsistencies, mostly by adding or correcting tags that are only read by the AI.

I modified ancient guardians to have industry focus, which shouldn't break anything.
And I changed some races, e.g. Khaturians from BAD_INFLUENCE to NO_INFUENCE since they do not have influence focus. Having a skill level without the matching focus makes no sense, does it?

geoffthemedio:
> Having a skill level without the matching focus makes no sense, does it?

It could or could not... depends on context and details.

geoffthemedio:
So it looks like this is reading in the .focs.txt script files and parsing them for tags and similar. Will this break if moving all the scripts to Python, or will it be easily adapted in that case, or does it already handle them OK?

o01eg:
I think it could be adapted to Python. Maybe with Python AST library.

Grummel7:
Well, the way it is written now, it will simply ignore python files. So nothing will break, but the consistency check will be lost when the species definition are moved to python.

Adapting the check shouldn't be too hard, but cannot be prepared before the move.

geoffthemedio:
I also dislike the basic design of parsing the script files, FOCS or Python, to extract tags and even go into effectsgroups to look for labels. It's very fragile and the AI should really be getting its info from the API.

I assume the problem is that the effectsgroups that are defined in macros and inserted into content definitions don't automatically come with tags that get attached to that content, so they would need to be manually added to that content's tags list? If that's the main problem, then I'd rather have a tags field for an effectsgroup that adds the specified tags to whatever content has that effectsgroup in it, and then use tags to pass all this info to the AI.

Or is this really just checking to verify but not actually getting tags from the direct parsing?

Grummel7:
> I also dislike the basic design of parsing the script files, FOCS or Python, to extract tags and even go into effectsgroups to look for labels. It's very fragile and the AI should really be getting its info from the API.

The script is not part of the AI, it is only a consistency checker, that should be applied to commits via github hooks. The AI still gets the tags via freeOrionAIInterface.pyi -> species.tags(). The problem is that it gets a lot of wrong information that way, because currently the tags are not aligned with the effects.

> I assume the problem is that the effectsgroups that are defined in macros and inserted into content definitions don't automatically come with tags that get attached to that content, so they would need to be manually added to that content's tags list? If that's the main problem, then I'd rather have a tags field for an effectsgroup that adds the specified tags to whatever content has that effectsgroup in it, and then use tags to pass all this info to the AI.

Well, that would be another way to fix it, although I think we may as well use that script for now. When converting the species files to python, adding effects and tags from one source should be rather simple in the python code. Then we can discard the checker again.

Grummel7:
> > Having a skill level without the matching focus makes no sense, does it?
> 
> It could or could not... depends on context and details.

Do you know any case where the skill level is applied when using another focus?

Also the only species that had a skill level other then "NO_" without the matching focus have been added [recently](https://github.com/freeorion/freeorion/commit/fd820239b4733537c4d7db3774b781badc61d21c). It rather looks like oversight to me. Maybe @st-pa-fo can tell us whether it was intended.

Grummel7:
@geoffthemedio This is the PR we talked about in the dev meeting. Could you please merge it?
Plus, since I see no other way to reach you: could you please re-activate my forum account? I've tried to change the email address, which apparently locks it until confirm something sent to the new address. Maybe I mis-typed the, but I never got anything.

Vezzra:
> ...could you please re-activate my forum account? ...

I've reactivated your forum account just now. The registered email address does look a bit unusual, but not invalid/malformed. Please try to log into the forum, check your profile, and report back.

PR #3727, opened 2022-03-05
Remove unused check

Cjkjvfnby:
Discussed in the thread: https://www.freeorion.org/forum/viewtopic.php?p=110431#p110431

@geoffthemedio , could you add @Grummel7 to collaborators, so I could request a review from him?

### DIFF
--- a/default/python/AI/ColonisationAI.py
+++ b/default/python/AI/ColonisationAI.py
@@ -712,7 +712,7 @@ class OrbitalColonizationPlan:
 
         # make sure source is valid
         source = universe.getPlanet(self.source)
-        if not (source and source.ownedBy(fo.empireID()) and can_build_colony_for_species(source.speciesName)):
+        if not (source and source.ownedBy(fo.empireID())):
             return False
 
         # appears to be valid


Grummel7:
Note that this checks the species of the source planet, i.e. the planet that is producing the ship. For building outposts, that planet doesn't even need to fulfill the colonisation requirements.

There should perhaps be a check for the species intended to settle the target, but the AI has much bigger issues right now. Removing this check definitely allows the AI to finish Outpost Bases again.

Morlic-fo:
@Cjkjvfnby @Grummel7 
Are you guys absolutely sure that the check is redundant?

When I wrote the initial code, it checked whether the source planet species is capable of colonization.  Git blame indicates [refactoring](https://github.com/Cjkjvfnby/freeorion/commit/63e8a2fdf0cc5cb0a6b0c39ca98dd52816721deb#diff-9e4bee169814fc6dc8262a5a708121118aff358cf61e9d5b248f54980508a48dL699) made the meaning of this particular check less clear and probably introduced the bug you are seeing right now.

The reason for that check was that Outpost Pods (i.e. the ship part) could only be built by species capable of colonization. 
If after the creation of the colonization plan, the species changed (possibly wiped out?), then the plan could no longer be executed: It was impossible to build an outpost base on that planet. Subsequently, that colonization plan should be deleted so that an actually valid plan can then be made to colonize the target planet. That is/was necessary as no new colonization plan for a particular target planet may be created as long as another one exists.

Can outpost bases be constructed by any species nowadays? If not, the check should probably be kept.

If it actually can be built by any species, the [check when creating a plan in the first place](https://github.com/Cjkjvfnby/freeorion/blob/d9d790c524d06274885d05ae18bacaa3c3608a9a/default/python/AI/ColonisationAI.py#L258) should be adjusted as well:

Grummel7:
There maybe a lot of things that need adjustment and I would appreciate any help by someone who understands the AI better than I do.

Regarding the removed check: Thank you for the hint. Perhaps we should re-add it in a simplified form:
`if not (source and source.ownedBy(fo.empireID()) and source.speciesName):`
After all there is no way a species can be replaced without the planet being empty for at least one turn.

In fact, the check was wrong from the beginning: You do not a colonising species, but you do not need a population of 3 to build a colony base, so this check may have removed a perfectly working plan.

geoffthemedio:
> After all there is no way a species can be replaced without the planet being empty for at least one turn.

Other than Super Tester Takeover, that might be true with current content, but noe that it's at least possible to cause with effects.

Cjkjvfnby:
@Morlic-fo , Regardless of this check, there is a bug.  `can_build_colony_for_species` is useless, before we do the survey universe, and we call it before the survey universe.  

I'd say we should do the same check when we create a plan and when we validate it.

I'll try to address the temporal coupling and unify checks.

Morlic-fo:
> I'd say we should do the same check when we create a plan and when we validate it.

Yes, I agree. Extract the check into a method so we do not have to keep track of it at two places.

> I'll try to address the temporal coupling and unify checks.

As Grummel7 correctly points out, ```can_build_colony_for_species``` is the wrong check here. A simple ```species and species.canColonize``` check is probably more correct in this context.

PR #3728, opened 2022-03-05
Fixes some unused variable warnings

Oberlus:
Fixes some unused variable warnings.

### DIFF
--- a/UI/GovernmentWnd.cpp
+++ b/UI/GovernmentWnd.cpp
@@ -478,13 +478,11 @@ void PoliciesListBox::Populate() {
     ScopedTimer scoped_timer("PoliciesListBox::Populate");
 
     GG::Pt slot_size = GG::Pt(SLOT_CONTROL_WIDTH, SLOT_CONTROL_HEIGHT);
-    int text_pts = ClientUI::Pts();
 
     auto policy_palette = Parent();
     auto gov_wnd = std::dynamic_pointer_cast<GovernmentWnd>(policy_palette->Parent());
     if (gov_wnd) {
         slot_size = gov_wnd->GetPolicySlotSize();
-        text_pts = gov_wnd->GetPolicyTextSize();
     }
 
     const GG::X TOTAL_WIDTH = ClientWidth() - ClientUI::ScrollWidth();
@@ -667,7 +665,6 @@ void GovernmentWnd::PolicyPalette::CompleteConstruction() {
             UserStringNop("POLICY_LIST_UNAFFORDABLE"), UserStringNop("POLICY_LIST_RESTRICTED"),
             UserStringNop("POLICY_LIST_LOCKED")}};
     for (size_t n = 0; n < 5; ++n) {
-        Availability avail = static_cast<Availability>(n);
         auto& button = m_availabilities_buttons[n];
         button = GG::Wnd::Create<CUIStateButton>(UserString(BUTTON_LABELS[n]),
                                                  GG::FORMAT_CENTER, std::make_shared<CUILabelButtonRepresenter>());
--- a/UI/LinkText.cpp
+++ b/UI/LinkText.cpp
@@ -589,14 +589,14 @@ void TextLinker::MarkLinks() {
 }
 
 std::string LinkStringIfPossible(const std::string &raw, const std::string &user_string) {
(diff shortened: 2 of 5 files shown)


Oberlus:
Most were fixed by removing an unused variable.
Three cases require review:
- UI/LinkText.cpp: removing the "auto x =" in the if conditions makes the warning disappear, but maybe they were needed for something?
- UI/GovernmentWnd.cpp:676: the unused variable, now used, is making a static cast that maybe was unnecesary.
- UI/MapWnd.cpp: idem.

universe/ObjectMap.cpp
```diff
@@ -133,10 +133,11 @@ namespace {
             return false;
 
         } else {
+            // TODO Implement or remove
             // shouldn't be possible to instantiate with this block enabled, but if so,
             // try to generate some useful compiler error messages to indicate what
             // type MappedObjectType is
-            using GenerateCompileError = typename MappedObjectType::not_a_member_zi23tg;
+            //using GenerateCompileError = typename MappedObjectType::not_a_member_zi23tg;
```
geoffthemedio: no, this is intentional, hence the name

I'd prefer to have a static_assert in here, but they apparently get evaluated even if that constexpr if / else block isn't used

universe/ObjectMap.cpp
```diff
@@ -133,10 +133,11 @@ namespace {
             return false;
 
         } else {
+            // TODO Implement or remove
```
geoffthemedio: no, this is here to generate an error if an incompatible type is used

... like what the following comment says ...

UI/LinkText.cpp
```diff
@@ -589,14 +589,14 @@ void TextLinker::MarkLinks() {
 }
 
 std::string LinkStringIfPossible(const std::string &raw, const std::string &user_string) {
-    if      (auto x = GetBuildingType(raw)) return LinkTaggedPresetText(VarText::BUILDING_TYPE_TAG, raw, user_string);
```
geoffthemedio, line 592: the `x` aren't used and then go out of scope, so removing them is fine...

UI/MapWnd.cpp
```diff
@@ -4141,7 +4141,7 @@ void MapWnd::InitVisibilityRadiiRenderingBuffers() {
 
             GG::Pt circle_centre = GG::Pt{GG::X(X), GG::Y(Y)};
             int radius_i = static_cast<int>(radius);
-            GG::Pt rad_pt{GG::X(radius), GG::Y(radius)};
```
geoffthemedio, line 4144: probably just remove `radius_i`

universe/ObjectMap.cpp
Oberlus: I misunderstood the comment due to my lack of knowledge.
I have no idea what that "using GenerateCompileError..." line does, and judging by the compiler warning, it seemed to me like a declaration of a type that was later not used. But I assume it's doing more than that but the compiler is confused.

UI/GovernmentWnd.cpp
```diff
@@ -673,8 +671,8 @@ void GovernmentWnd::PolicyPalette::CompleteConstruction() {
                                                  GG::FORMAT_CENTER, std::make_shared<CUILabelButtonRepresenter>());
         AttachChild(button);
         button->CheckedSignal.connect(
-            boost::bind(&GovernmentWnd::PolicyPalette::ToggleAvailability, this, Availability(n)));
-        button->SetCheck(m_availabilities_state.GetAvailability(Availability(n)));
```
geoffthemedio: I think the `static_cast` does the same as the constructor-like `Availability(n)` anyway. Probably works out to be the same code generation...

Oberlus: In that case I think it's better to remove it.

UI/MapWnd.cpp
Oberlus: Will do.

universe/ObjectMap.cpp
geoffthemedio: It doesn't do anything, really. `using X = Y;` is equiavlent to `typedef Y X;`

The point of the line is that if that constexpr else case gets compiled, it will generate a compile error.

Oberlus:
All comments addressed.
Hastily playtested: loaded a turn 20 game, advanced a couple of turns, checked no warnings or errors in message window or command line output.
Good to merge.

PR #3702, opened 2022-02-12
Removed active AI focus dithering

Grummel7:
See https://www.freeorion.org/forum/viewtopic.php?p=109993&sid=403d00698660ab7fdeeceb7b52b204a4#p109993

PR #3729, opened 2022-03-06
First AI with policies plus minor adoptions

Grummel7:
There is still a lot to do, but it's time to submit a first version.

This PR supersedes #3702 and #3723, I suggest to close at least 3702. I can re-base this one, if 3723 get merged before. It does not depend on #3725, although without it some AI decision are based on wrong information.


default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def __init__(self):
+        # resourceAvailable includes this turns production, but that is wrong for influence
+        self.empire = fo.getEmpire()
+        self.universe = fo.getUniverse()
+        self.ip = self.empire.resourceAvailable(fo.resourceType.influence) - self.get_infl_prod()
+        self.adopted = set(self.empire.adoptedPolicies)
+        self.adoptable = self.get_adoptable()
+        empire_owned_planet_ids = PlanetUtilsAI.get_owned_planets_by_empire(self.universe.planetIDs)
+        self.planet_ids = PlanetUtilsAI.get_populated_planet_ids(empire_owned_planet_ids)
```
Cjkjvfnby: Lets name it `self.populated_planet_ids`

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,235 @@
...
+        self.aistate = get_aistate()
+
+    def generate_orders(self):
+        """generate policy orders"""
+        debug(
+            "Start turn %d IP: %.2f + %.2f, adoptable: %s, adopted: %s",
+            fo.currentTurn(),
+            self.ip,
+            self.get_infl_prod(),
+            str(self.adoptable),
```
Cjkjvfnby: You don't need to `str`, because `%s` will convert it to string.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,235 @@
...
+        self.ip = self.empire.resourceAvailable(fo.resourceType.influence) - self.get_infl_prod()
+        self.adopted = set(self.empire.adoptedPolicies)
+        self.adoptable = self.get_adoptable()
+        empire_owned_planet_ids = PlanetUtilsAI.get_owned_planets_by_empire(self.universe.planetIDs)
+        self.planet_ids = PlanetUtilsAI.get_populated_planet_ids(empire_owned_planet_ids)
+        self.num_populated = len(self.planet_ids)
+        self.num_outposts = len(empire_owned_planet_ids) - self.num_populated
+        self.aistate = get_aistate()
+
+    def generate_orders(self):
```
Cjkjvfnby, line 47: This function is very huge. Let's deal with it. 

We could keep it as an index of operation and move all low-level stuff into separate methods. 

```python
def generate_orders():
    debug(...)
    self._adopt_base_policies()
    self._process_bureaucracy()
    ...
```

Cjkjvfnby (commented):
This is an initial review.  I have not read the whole code. 

General comments:
- Python does not have real public and private items, but it has a similar convention: 
Attribute (instance variable or method) started from a single underscore, considered private, and not recommended for use outside the class. For example: `self._empire`. The idea is to keep public API as small as possible. This will give us the flexibility for future changes.
- We use standard Python documentation convention.  Let's keep a bare minimum: Start with a capital letter and ends with a period.  A more detailed description could be found here https://www.python.org/dev/peps/pep-0257/.  I'd prefer to keep more detailed talk-about docs out of this PR.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,235 @@
...
+            else:
+                self.try_adopt_bureaucracy()
+            self.check_deadopt_centralization()
+            self.techno_or_industry()
+        new_production = self.get_infl_prod(True)
+        debug("End of turn IP: %.2f + %.2f", self.ip, new_production)
+        # priority shouldn't only be based on current value
+        forecast = self.ip + 3 * new_production
+        if infra1 in self.adopted and bureaucracy not in self.adopted and centralization not in self.adopted:
+            ccost = fo.getPolicy(centralization).adoptionCost()
```
Cjkjvfnby: Let's use full names: centralization_cost, bureaucracy_cost.  

That short name style comes from the last century when people have small screens and printed the code on paper to do the code review.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,235 @@
...
+            bcost = fo.getPolicy(bureaucracy).adoptionCost()
+            forecast -= self.num_populated + ccost + bcost
+        if forecast < 40:
+            prio = 60 - forecast
```
Cjkjvfnby: When I get the not word in the code, you have to either guess or look into the code.   So let's just use full names.

We already have some conventions for names like `pid` (planet id), but I try to avoid adding others.

Grummel7: I think I've resolved all your comments, but this one I do not understand.

default/python/AI/PolicyAI.py
Grummel7: Well, it was quite short compared to other functions in the AI, but it surely can't hurt to split it now, it will surely grow.

Cjkjvfnby: > Well, it was quite short compared to other functions in the AI

There is a nice metric called [cognitive complexity](https://pypi.org/project/flake8-cognitive-complexity/), it's a measurement of how hard it to read the code.  For my personal (single) project I keep it at 5.   AI has hundreds and thousands.   

So referencing AI code as an example might not be a good idea.

default/python/AI/PolicyAI.py
Cjkjvfnby: `prio` is not a word.  So I need to guess it. Looks like a priority, but it's not a good match with the word forecast.  So I need to break reading the code and do a forward lookup on what this variable is. 

This decreases the readability.   

It does not slow down me a lot, but fixing it is also very easy.  By fixing I mean not only change this variable, but also prevent other such cases in the future.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def _process_bureaucracy(self):
+        if bureaucracy in self._adopted:
+            max_turn = 19  # AI doesn't build Regional Admins yet (TBD count, palace may be gone, too)
+            if fo.currentTurn() >= self._empire.turnsPoliciesAdopted[bureaucracy] + max_turn:
+                self._deadopt(bureaucracy)
+                self._try_adopt_centralization()
+        else:
+            self._try_adopt_bureaucracy()
+        self._check_deadopt_centralization()
```
Cjkjvfnby, line 81: This line is always executed.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def _try_adopt_bureaucracy(self):
+        """Try to adopt bureaucracy, starting with centralization if necessary."""
+        if bureaucracy in self._adoptable:
+            self._adopt(bureaucracy)
+        elif centralization not in self._adopted:
+            self._try_adopt_centralization()
+        elif self._empire.policyPrereqsAndExclusionsOK(bureaucracy):
+            if self._bureaucracy_cost <= self._ip:
+                # when we are here, no free slot should be the only reason, so we
```
Cjkjvfnby, line 105: This looks like adopting with extra steps.  Dependency on adopting `centralization` looks wired.  It's not only for AI, this also makes Human player life more complicated.   I'd prefer to remove this dependency and unlock the administration by `bureaucracy` or `centralization`.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+            self._deadopt(centralization)
+
+    def _techno_or_industry(self):
+        research_prio = self._aistate.get_priority(PriorityType.RESOURCE_RESEARCH)
+        production_prio = self._aistate.get_priority(PriorityType.RESOURCE_PRODUCTION)
+        may_switch_to_techno = self._can_adopt(technocracy, industrialism)
+        may_switch_to_industry = self._can_adopt(industrialism, technocracy)
+        # do not switch too often
+        if technocracy in self._adopted:
+            research_prio *= 1.25
```
Cjkjvfnby, line 129: This multiplication does not have any sense, because when `may_switch_to_industry = False` priority check will not be executed.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+            may_switch_to_techno,
+            may_switch_to_industry,
+        )
+
+    def _determine_influence_priority(self, new_production):
+        # priority shouldn't only be based on current value
+        forecast = self._ip + 3 * new_production
+        if infra1 in self._adopted and bureaucracy not in self._adopted and centralization not in self._adopted:
+            forecast -= self._num_populated + self._centralization_cost + self._bureaucracy_cost
+        if forecast < 40:
```
Cjkjvfnby, line 153: This is some magic, I don't understand.  Let's extract it to function and add documentation to it.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+            prio = 60 - forecast
+        else:
+            prio = 800 / forecast
+        prio *= (3 + self._num_populated + self._num_outposts / 10) / 10
+        # if not infra3 in self._adopted: prio += 10
```
Cjkjvfnby: Let's remove unused code.   If you want to add a message to the future, just add a comment.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+        """Get / Update IP production."""
+        if update:
+            self._universe.updateMeterEstimates(self._populated_planet_ids)
+            fo.updateResourcePools()
+        return self._empire.resourceProduction(fo.resourceType.influence)
+
+    def _can_adopt(self, name, replace_other=None):
+        policy = fo.getPolicy(name)
+        ret = name in self._empire.availablePolicies and policy.adoptionCost() <= self._ip
+        if replace_other and replace_other in self._adopted:
```
Cjkjvfnby, line 186: Should we check `self._empire.policyPrereqsAndExclusionsOK(name)` here?

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def _get_adoptable(self):
+        """List of adoptable policies with the still available IP."""
+        ret = set()
+        for p in self._empire.availablePolicies:
+            if p in self._adopted:
+                continue
+            if self._can_adopt(p):
+                ret.add(p)
+        return ret
```
Cjkjvfnby: Set comprehension is a great thing.
```suggestion
        def can_adopt(policy):
            return policy not in self._adopted and self._can_adopt(policy)

        return {p for p in self._empire.availablePolicies if can_adopt(p)}
```

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+    def status(self):
+        # only for interactive debugging
+        print("\nAdoptable Policies:")
+        for p in self._get_adoptable():
+            print("  ", p)
+        print("Empty Slots:")
+        for s in self._empire.emptyPolicySlots:
+            print("  ", s)
+        print("Adopted:")
+        for cat in self._empire.categoriesSlotPolicies:
```
Cjkjvfnby: [meow](https://en.wikipedia.org/wiki/Cat_(Unix))!  Let's rename to `category`.

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+            if cat:
```
Cjkjvfnby: self._empire.categoriesSlotPolicies returns a map-like object.  Could it return None during iteration?

default/python/AI/PolicyAI.py
```diff
@@ -0,0 +1,242 @@
...
+        print("Influence:", self._ip)
+        print("Infl. Prod.:", self._get_infl_prod())
+        print("Num Planet (pop./outp.): %d/%d" % (self._num_populated, self._num_outposts))
+
+
+def generate_policy_orders():
+    PolicyManager().generate_orders()
+
+
+def status():
```
Cjkjvfnby: Maybe rename it to `print_status` it will highlight debug purposes of this function.

default/python/AI/PolicyAI.py
Grummel7: That is intended. I'll add a comment...

default/python/AI/PolicyAI.py
Grummel7: That one's better discussed in the forum, I think.

default/python/AI/PolicyAI.py
Grummel7: > This multiplication does not have any sense, because when `may_switch_to_industry = False` priority check will not be executed.

I only set at most one of the may_switch_to... to false, the one that would reapply what is already adopted.
Later on I check whether I should switch to the other:
```
        if may_switch_to_industry and research_prio < production_prio:
           ...
        elif may_switch_to_techno and research_prio > production_prio:
```

default/python/AI/PolicyAI.py
Grummel7: I'll have to double-check, but I think you are right.

default/python/AI/PolicyAI.py
Grummel7: Actually I think _can_adopt should return False, for a policy that is already adopted, that would simplify other functions, too.

I don't know why I didn't think about comprehension before (well, apart from the fact that I've never heard the name, but I have used it). Will change.

default/python/AI/PolicyAI.py
Grummel7: Actually, you are wrong. Take the _can_adopt(industrialism, technocracy) for example. industrialism excludes technocracy, so policyPrereqsAndExclusionsOK would return False, that's why I do not call it her. Note that I do call it if replace_other is None.

The only safe way to implement _can_adopt with replace_other at the moment would be to temporarily deadopt replace_other, call policyPrereqsAndExclusionsOK, than readopt it. I do not like that, not the least because it would create lots of pseudo orders. Perhaps we can add replace_other as an argument to policyPrereqsAndExclusionsOK. For the moment the function works good enough, I'll add a comment.

default/python/AI/PolicyAI.py
Cjkjvfnby: What is your personal opinion, does this dependency bring you problems when you play games as a user?

default/python/AI/colonization/calculate_planet_colonization_rating.py
```diff
@@ -576,15 +580,32 @@ def _calculate_planet_colonization_rating(
...
+
         retval += (
-            max(ind_val + asteroid_bonus + gas_giant_bonus, research_bonus, growth_val)
+            max(ind_val + asteroid_bonus + gas_giant_bonus, research_bonus, growth_val, infl_val)
             + fixed_ind
             + fixed_res
             + supply_val
         )
+        detail.append(
+            " max(%.1f+%.1f+%.1f, %.1f, %.1f, %.1f) + %.1f + %.1f + %.1f"
```
Cjkjvfnby: Let's use [f-string](https://realpython.com/python-f-strings/) here.  It will be much more readable. 

PS some python hint: you could put one string on multiple lines with open and close quotes with whitespace in between. This helps with the formatting of long lines.

```python
a = []

a.append(
    "word1"
    " word2"
    " word3" ' word4'
)

print(a)
# ['word1 word2 word3 word4']

File: default/python/AI/colonization/rate_influence.py [hunk 17]
```diff
@@ -0,0 +1,43 @@
...
+
+INFLUENCE_SPECIALS = [
+    "FRACTAL_GEODES_SPECIAL",
+    "MIMETIC_ALLOY_SPECIAL",
+    "SHIMMER_SILK_SPECIAL",
+    "SPARK_FOSSILS_SPECIAL_DESC",
+    "SUCCULENT_BARNACLES_SPECIAL",
+]
+# NamedReal name = "SPECIAL_INFLUENCE_FOCUS_BONUS" value = 3.0
+# NamedReal name = "ARTISANS_INFLUENCE_FLAT_FOCUS" value = 4.0
```
Cjkjvfnby: Let's replace it with "Hardcoded values from the scripts ARTISANS_INFLUENCE_FLAT_FOCUS"
And put a comment before the variable.

default/python/AI/colonization/rate_influence.py
```diff
@@ -0,0 +1,43 @@
...
+SPECIAL_FLAT = 3.0
+ARTISAN_FLAT = 4.0
+# TBD artisan bonus for non-focused planets.
+# well, first we'll have to teach the AI to use workshops at all...
+PLC_ARTISAN = "PLC_ARTISAN_WORKSHOPS"
```
Cjkjvfnby, line 23: Is it used anywhere?

default/python/AI/colonization/rate_influence.py
```diff
@@ -0,0 +1,43 @@
...
+    Give a rating how useful the colony would be for generating influence.
+    This should be used as an alternative for industry/research.
+    """
+    if not species:
+        return 0.0
+    infl_tag_mod = SPECIES_INFLUENCE_MODIFIER.get(get_species_tag_grade(species.name, Tags.INFLUENCE), 1)
+    production = sqrt(max_pop) * infl_tag_mod
+    for special in planet.specials:
+        if special in INFLUENCE_SPECIALS:
+            production += SPECIAL_FLAT
```
Cjkjvfnby: The same idea as comprehension, but it is called generator expression.

```suggestion
    production += sum(SPECIAL_FLAT for special in planet.specials if special in INFLUENCE_SPECIALS)
```

default/python/AI/PolicyAI.py
Grummel7: no

python/EmpireWrapper.cpp
```diff
@@ -324,6 +324,8 @@ namespace FreeOrionPython {
 
             .def("policyAvailable",                 +[](const Empire& e, const std::string& policy) { return e.PolicyAvailable(policy); })
 
+            .def("policyPrereqsAndExclusionsOK",    +[](const Empire& e, const std::string& policy) { return e.PolicyPrereqsAndExclusionsOK(policy); })
```
Cjkjvfnby, line 327: Maybe name it  `policyCouldBeAdopted()`

default/python/AI/ResourcesAI.py
```diff
@@ -84,6 +96,39 @@ def bake_future_focus(self, pid, focus, update=True):
...
         pinfo = self.raw_planet_info.get(pid)
+        if (focus == INDUSTRY or focus == RESEARCH) and not force:
+            idx = 0 if focus == INDUSTRY else 1
+            # check for influence instead
+            debug("possible_output of %s: %s", pinfo.planet.name, str(pinfo.possible_output))
+            focus_gain = pinfo.possible_output[focus][idx] - pinfo.possible_output[INFLUENCE][idx]
+            influence_gain = pinfo.possible_output[INFLUENCE][2] - pinfo.possible_output[focus][2]
+            if influence_gain * self.priority[2] > focus_gain * self.priority[idx]:
+                debug(
+                    "Chosing influence over %s. %.2f * %.1f > %.2f * %.1f",
```
Cjkjvfnby: Let's use f-string here. 

Float formatting looks lie that:
```python
a = 0.3333
f"{a:.2}"
'0.33'
```

default/python/AI/PolicyAI.py
Grummel7: Well, there is currently a small issue with Centralization: IP production forecast isn't updated properly, so you have a leave a safety margin. I've already addressed it during one of the developer meetings.

Apart form that, my personal opinion is: problems are there to be solved. I don't mind one way or the other, but I guess some people would be glad to get rid of it.

default/python/AI/ResourcesAI.py
```diff
@@ -116,36 +162,73 @@ def calculate_planet_infos(self, pids):
             if INDUSTRY in planet.availableFoci and planet.focus != INDUSTRY:
                 fo.issueChangeFocusOrder(pid, INDUSTRY)  # may not be able to take, but try
 
+        def merge(po1, po2):  # merge too partially determined possible_outputs
```
Cjkjvfnby: Let's converts comment into the docstring.

default/python/AI/ResourcesAI.py
```diff
@@ -116,36 +162,73 @@ def calculate_planet_infos(self, pids):
...
+            return (
+                po1[0] if po1[0] else po2[0],
```
Cjkjvfnby:
```suggestion
                po1[0] or po2[0]
```

default/python/AI/PolicyAI.py
Grummel7: Well, the whole _determine_influence_priority function is "magic", and surely not finalised. I doubt that splitting it makes it better. I've added some comments.

default/python/AI/PolicyAI.py
Grummel7: Actually the code became simpler after I changed _can_adopt to return False when already adopted.

default/python/AI/ResourcesAI.py
Grummel7: Hey, that looks almost like jinja, the one thing I like about ansible ;-)
But we need an extra f:
```
a=123.456
f"{a:.2}"
'1.2e+01'
f"{a:.2f}"
'123.46'
```

python/EmpireWrapper.cpp
Grummel7: Well, I simply stuck to the custom of forwarding the C++ function to a python function with the same name. Plus couldBeAdopted sounds like it also checks for required IP, the current name specifies exactly what it is doing.

default/python/AI/colonization/rate_influence.py
Grummel7: Well, as the comment says, so far the AI does not adopt it yet. But that doesn't make the calculation wrong.

default/python/AI/ResourcesAI.py
Cjkjvfnby: > Hey, that looks almost like jinja, the one thing I like about ansible ;-)
 But we need an extra f:

If you do Jinja2 directly it requires a bit more than f.  Same for `c-style` and `.format`. 

To read this entry I need to go back of forth to the variable list.  `%s. %.2f * %.1f > %.2f * %.1f`, f string allows to read it in sequence.

default/python/AI/colonization/rate_influence.py
Cjkjvfnby: I follow the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle. 

If we don't need it to this PR we should not add it to the code.  Ideas, priorities are changed over time and sometimes these items are left abandoned.

Cjkjvfnby (approved):
@Grummel7 , great job! 

We could go forward and merge this code.

python/EmpireWrapper.cpp
geoffthemedio: The current name is probably better as it's more specific. As @Grummel7 notes, whether a policy can be adopted is more than that its prereqs and exclusions are satisfied.

PR #3730, opened 2022-03-08
Ensure that tags has valid name

Cjkjvfnby:
@Grummel7 , here is some magic that will help with tag parsing. 

Metaclasses in Python is a hook into the class creation process. 
The code will be executed when the file is imported and the class code is evaluated. 

PS. I have not tested this code yet, since I use windows compiled build and it's not yet compatible with the master.  The concept is working, but I could mess up with usages.

 Also, there is another discussion about getting that info in a different way, so this code could become obsolete.   So I set PR as Draft.

### DIFF
--- a/default/python/AI/AIDependencies.py
+++ b/default/python/AI/AIDependencies.py
@@ -20,6 +20,7 @@ Example usage:
     my_industry = AIDependencies.INDUSTRY_PER_POP * my_population
 """
 import freeOrionAIInterface as fo
+from enum import Enum, EnumMeta
 from typing import Dict
 
 # Note re common dictionary lookup structure, "PlanetSize-Dependent-Lookup":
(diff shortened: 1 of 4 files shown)


Grummel7:
Well, if you look at #3725, I've just added a tag with an underscore.

The problem was that the tags were all named like the effect groups implementing them, except for effects OFFENSE_TROOPS, which was named ATTACKTROOPS in the tags. Since this was neither documented nor checked, some species had  ATTACKTROOPS in their tags other had OFFENSE_TROOPS. Keeping the names identical makes life easier for the script writers and also makes it easier to check the consistency.

Apart from that, I guess using Enum for Tags and many other definition in AIDependencies may be a good idea, especially if we could check the the values in the game are actually know to the AI. Writing a functions to provide a list of all techs, all policies, ship parts, etc. shouldn't be too hard. Then at the start the AI could check for both values not defined in the AI as well as for defined values that are no longer used.

Cjkjvfnby:
> Writing a functions to provide a list of all techs, all policies, ship parts, etc. shouldn't be too hard. 

Converting dependencies to enums and do validation for that names sound like a great idea.  I'll try it.

As for this PR. I will leave the enum part but will remove metaclass with the check.

Cjkjvfnby:
Not actual anymore.

PR #3737, opened 2022-03-13
Use default WindowsTargetPlatformVersion

Morlic-fo:
Considering that mainstream windows 8.1 support ended over 4 years ago, I propose to update the target platform version.
If a fixed version is preferred rather than using default value, I propose we should at least target 10.0 (i.e. latest version of windows 10 SDK) rather than 8.1...

### DIFF
--- a/msvc2022/Common/Common.vcxproj
+++ b/msvc2022/Common/Common.vcxproj
@@ -22,7 +22,6 @@
     <ProjectGuid>{BEDF460A-EAE9-4E20-AFB2-2C8434051150}</ProjectGuid>
     <Keyword>Win32Proj</Keyword>
     <RootNamespace>Common</RootNamespace>
-    <WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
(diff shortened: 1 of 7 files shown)


geoffthemedio:
We still have people being surprised that Win7 isn't supported any more... If there's not a need, then I don't see a point to updating. It was bumped to 8.1 due to this being necessary to use a newer MSVC toolset version, which was needed to support some newer C++ features.

Morlic-fo:
Visual Studio 2019/2022 do not include Windows SDK 8.1 in their installers. It is of course still possible to find it on the legacy download page on MS website. Requiring it seems weird though if we explicitly provide project files for msvc 2019/2022.

Updating the projects files manually (like done here) is cumbersome since it changes files tracked by git and subsequently regularly requires a ```git stash``` to e.g. rebase.


Or are you worried about backwards compatibility of the precompiled binaries?

geoffthemedio:
Partly the issue is I'm about to go on a trip with a Win8.1 laptop, so need the build to still work on that machine. I suppose since barely anyone still uses Win8.1 rather than Win10, it's not otherwise that that important to still support (vs. supporting 7, which is apparently still ~4x more popular than 8.1).

Morlic-fo:
If I understand correctly, you can install and build with the Win10 SDK on a (fully updated) Win8.1 system (https://developer.microsoft.com/en-us/windows/downloads/windows-sdk/)

To my understanding, it is also possible to target older windows installations (up to Win7/SP1) using the Win10 SDK by using the relevant [defines](https://docs.microsoft.com/en-us/windows/win32/winprog/using-the-windows-headers)

geoffthemedio:
https://github.com/freeorion/freeorion/commit/b3b85121c9dcb01c483cbef33fbc71428e7a4233

PR #3741, opened 2022-03-13
Let split_ship_from_fleet return INVALID_ID on failure

Morlic-fo:
Avoids the awkward error checking and prevents potential bugs when using 0 as return code in C++ side...


geoffthemedio:
looks fine

PR #3744, opened 2022-03-14
[AI] Ensure colony_buildes are initialized before usage

Cjkjvfnby:
This PR address the issue that it's hard to track if colony_buildes  is fully initialized before usage.

This code uses [publisher-subscriber](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern). We did not use this pattern before for AI. 

Not sure if we should crash or just return the default value and log error.  I am open to suggestions.

@Grummel7, @Morlic-fo, could you take a look?

### DIFF
--- a/default/python/AI/empire/colony_builders.py
+++ b/default/python/AI/empire/colony_builders.py
@@ -2,8 +2,14 @@ from typing import Dict, List, Mapping, Sequence, Union
 
 import AIDependencies
 from common.fo_typing import PlanetId, SpeciesName
+from common.listeners import register_pre_handler
 from freeorion_tools import tech_is_complete
 from freeorion_tools.caching import cache_for_current_turn
+from freeorion_tools.lazy_initializer import InitializerLock
+
+colony_builder_lock = InitializerLock("colony_builder")
+
+register_pre_handler("generateOrders", colony_builder_lock.lock)
 
 
 def set_colony_builders(species_name: SpeciesName, yards: Sequence[PlanetId]):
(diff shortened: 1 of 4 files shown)


default/python/AI/empire/colony_builders.py
```diff
@@ -2,8 +2,15 @@
...
 from common.fo_typing import PlanetId, SpeciesName
+from common.listeners import register_post_handler, register_pre_handler
 from freeorion_tools import tech_is_complete
 from freeorion_tools.caching import cache_for_current_turn
+from freeorion_tools.lazy_initializer import InitializerLock
+
+_colony_builder_lock = InitializerLock("colony_builder")
+
+register_pre_handler("generateOrders", _colony_builder_lock.lock)
+register_post_handler("survey_universe", lambda *args: _colony_builder_lock.unlock())
```
Grummel7, line 13: Phew, this looks almost like obfuscated python to me. But I keep learning...

Can't you change register_pre_handler / register_post_handler to accept a function instead of a string? I've learned to hate such string based interfaces. A typo or a change in the function name will simply means the the handler isn't called and code analysers will not identify these strings as a usage of the function.

Morlic-fo: > Can't you change register_pre_handler / register_post_handler to accept a function instead of a string

I agree. In doubt, just internally convert to the `function.__name__` if needed...

```
>>> def print_function_name(fun):
...     print(fun.__name__)
...
>>> print_function_name(print)
print
```

Cjkjvfnby: > Phew, this looks almost like obfuscated python to me

Feel free to ask questions. 

> Can't you change register_pre_handler / register_post_handler to accept a function instead of a string? 

yes, it's possible.  Let's do it in a separate PR.

Issue #3745, opened 2022-03-15
AI invasion handling


Grummel7:
I'm sure many have wondered why the AI produces so many troop ships, flies them to random places and then seems to ignore them.
I've dug into it and the answer is complicated. #3735 was a part of it, but that was rather new. The other reasons go deeper.
The following multiplayer save file demonstrate a lot of them (check the Green AI): [GreenNotUsingTroopers.mps.gz](https://github.com/freeorion/freeorion/files/8256287/GreenNotUsingTroopers.mps.gz)

It was created before the fix of #3735. When you play it now with the current master, the AI will immediately send a lot of troop ships (much more than needed) to Sunce α I and successfully conquer it. Then it stop invading again. (btw, having systems called Sunce and Sonce next to each other can be confusing ;-)

Reasons for this behaviour:
1. When assigning troop fleets, `FleetUtilsAI.get_fleets_for_mission` is called with `target_stats = {"rating": 10, "troopCapacity": troops_needed + _TROOPS_SAFETY_MARGIN}`. Troop ships have no battle rating, so the target is never reached and the function assigns all available troop ships to Sunce α I and then finds that they at least fulfill the min_stats.
2. `AIFleetMission._check_retarget_invasion` does not check whether the mission target is still valid. After successful invasion extra troop ships remain with a mission to invade Sunce α I (but the don't fly there, since it now has shields, see below).
3. Some troop fleets are scheduled to invade the Acirema planet Alfirk α I. The planet has shields, but there is a ship keeping them down. The problem is that the ship's mission is MissionType.PROTECT_REGION, not MissionType.SECURE: `fleet_orders.py:634 : trooper_move_reqs_met() holding Invasion fleet 19467 before leaving supply because target (P_426<Alfirk α I>) has nonzero max shields and there is not yet a military fleet assigned to secure the target system.`

Point 1 could be fixed easily by removing the rating condition. It looks intentional, though. Does anyone have an idea?

Point 2: _check_retarget_invasion cries for refactoring. It has lots of returns and TODOs and only if no return is triggered, it does some retargeting. Also there is this:
```
        last_sys_target = INVALID_ID
        if orders:
            last_sys_target = orders[-1].target.id
        if last_sys_target == fleet.systemID:
            return  # TODO: check for best local target
```
The second if never triggers, because the last target of an invasion mission is a planet.

[...]

PR #3754, opened 2022-03-19
Improve AI invasion handling

Grummel7:
This should fix the three points described in #3745. It should conquer more successful and waste less troopers now.
The AI still builds too many troopers (priority calculation sometimes produces strange results) and may ignore good planets (planetary rating also needs work), but these are for another PR.

### DIFF
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -260,6 +260,7 @@ class AIFleetMission:
                     debug("EspionageAI predicts we can no longer detect %s, will abort mission" % fleet_order.target)
                     planet_stealthed = True
         if target_is_planet and not planet_stealthed:
+            # TBD use fleet_order.is_valid or add fleet_order.should_abort
             if isinstance(fleet_order, OrderColonize):
                 if planet.initialMeterValue(fo.meterType.population) == 0 and (
                     planet.ownedBy(fo.empireID()) or planet.unowned
@@ -288,8 +289,6 @@ class AIFleetMission:
         # canceling fleet orders
         debug("   %s" % fleet_order)
         debug("Fleet %d had a target planet that is no longer valid for this mission; aborting." % self.fleet.id)
-        self.clear_fleet_orders()
-        self.clear_target()
         FleetUtilsAI.split_fleet(self.fleet.id)
         return True
 
@@ -414,6 +413,7 @@ class AIFleetMission:
             if isinstance(fleet_order, (OrderColonize, OrderOutpost, OrderInvade)) and self._check_abort_mission(
                 fleet_order
             ):
+                self.clear_fleet_orders()
                 self.clear_target()
                 return
         aistate = get_aistate()
--- a/default/python/AI/AIFleetMission.py
+++ b/default/python/AI/AIFleetMission.py
@@ -268,7 +268,19 @@ class AIFleetMission:
             elif isinstance(fleet_order, OrderOutpost):
                 if planet.unowned:
                     return False
-            elif isinstance(fleet_order, OrderInvade):  # TODO add substantive abort check
+            elif isinstance(fleet_order, OrderInvade):
+                if planet.owner == fo.getEmpire().empireID:
+                    debug(f"Abandoning invasion for {planet}, it already belongs to us")
+                    return True
+                if planet.currentMeterValue(fo.meterType.maxShield):
+                    military_support_fleets = MilitaryAI.get_military_fleets_with_target_system(planet.systemID)
+                    if not military_support_fleets:
+                        # Maybe try again later...
+                        debug(
+                            f"Abandoning Invasion mission for {planet} because target has nonzero max shields "
+                            f"and there is no military fleet assigned to secure the target system."
+                        )
+                        return True
                 return False
             else:
                 return False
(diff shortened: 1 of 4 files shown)


default/python/AI/AIFleetMission.py
```diff
@@ -268,7 +268,19 @@ def _check_abort_mission(self, fleet_order: AIFleetOrder):
...
+                    debug(f"Abandoning invasion for {planet}, it already belongs to us")
+                    return True
+                if planet.currentMeterValue(fo.meterType.maxShield):
+                    military_support_fleets = MilitaryAI.get_military_fleets_with_target_system(planet.systemID)
+                    if not military_support_fleets:
+                        # Maybe try again later...
+                        debug(
+                            f"Abandoning Invasion mission for {planet} because target has nonzero max shields "
+                            f"and there is no military fleet assigned to secure the target system."
+                        )
```
Cjkjvfnby, line 283: Please do not make big and complicated functions bigger and more complicated.

Just extract this code to a method. `_check_abort_mission_for_invarions_fleet`.

PS.  good OOP pattern is to replace these is instance calls with a single call of the fleet_order.should_abort() and move code from branches to classes.  But this is out of scope of this PR.

default/python/AI/AIFleetMission.py
```diff
@@ -500,15 +512,15 @@ def issue_fleet_orders(self):
                         return  # colonize order must not have completed yet
                 clear_all = True
                 last_sys_target = INVALID_ID
+                secure_targets = set(
+                    AIstate.colonyTargetedSystemIDs
+                    + AIstate.outpostTargetedSystemIDs
+                    + AIstate.invasionTargetedSystemIDs
+                )
```
Cjkjvfnby, line 519: I don't have old python installed.  This works in Pyhton3.9, but I am not sure if it works in 3.6

You can create set with unpacking variable.
```python
a = [1,2, 3]
b = [3,4, 5]
{*a, *b}
{1, 2, 3, 4, 5}
```

default/python/AI/AIFleetMission.py
```diff
@@ -533,6 +545,11 @@ def issue_fleet_orders(self):
                     last_sys_target = self.target.id
                     debug("Check if PROTECT_REGION mission with target %d is finished.", last_sys_target)
 
+                # for SECURE missions, only release fleet if no settle or invade mission is ongoing
+                if self.type == MissionType.SECURE and self.target.id in secure_targets:
+                    debug(f"Secure mission for {self.target} not yet finished")
```
Cjkjvfnby, line 550: Maybe print other info about the mission? For example fleet.   You can try to print `self`.  This will cal `__str__` method implicitly.

default/python/AI/InvasionAI.py
```diff
@@ -543,27 +527,45 @@ def evaluate_invasion_planet(planet_id):
...
+        f" - popval: {colony_base_value:.1f}\n"
+        f" - bldval: {bld_tally}\n"
+        f" - enemyval: {enemy_val}\n"
+        f" - system secured: {system_secured}"
     )
-    debug(" - system secured: %s" % system_secured)
     return [planet_score, planned_troops]
 
 
+def secure_system(system_id: SystemId, secure_now: bool):
```
Cjkjvfnby: You could specify return type with annotations.
```suggestion
def secure_system(system_id: SystemId, secure_now: bool) -> bool:
```

default/python/AI/InvasionAI.py
```diff
@@ -543,27 +527,45 @@ def evaluate_invasion_planet(planet_id):
...
+    """Check whether we have any military fleets in the system.
+    If secure_now is True, set one of the fleets to MissionType.SECURE
+    If secure_now is False, return true if there is any fleet that secure it.
```
Cjkjvfnby:
```suggestion
    If secure_now is False, return True if there is any fleet that secures it.
```

default/python/AI/InvasionAI.py
```diff
@@ -543,27 +527,45 @@ def evaluate_invasion_planet(planet_id):
...
+        fleet = universe.getFleet(mission.fleet.id)
+        if not fleet or fleet.systemID != system_id:
+            continue
+        target_obj = mission.target.get_object()
+        if target_obj is not None and target_obj.id in secure_targets:
+            if secure_now and mission.type != MissionType.SECURE:
+                mission.set_target(MissionType.SECURE, mission.target)
+            secured = system_status.get("myFleetRating", 0) > 0
+            break
+    return secured
```
Cjkjvfnby, line 568: We could try to remove the `secured` variable. 

```suggestion
            return secured
    return False
```

Should we stop on the first fleet to we should keep trying?
If this felt is not

Cjkjvfnby (commented):
Looks great, a couple of minor comments.

default/python/AI/AIFleetMission.py
Grummel7: Well, I just moved this code, because I needed the set outside if clause. There are surely more important things to refactor than this.

default/python/AI/AIFleetMission.py
Grummel7: There is lots of debug outputs about the mission before, reading the logs you do not miss that information.

default/python/AI/InvasionAI.py
Grummel7: actually it should be ... that could secure it.

default/python/AI/InvasionAI.py
Grummel7: I know that could isn't perfect. I copied the "myFleetRating" stuff. It seems to work in by and large...
I think I should rather check whether the fleet I just set to secure actually does the job, but I have to figure out how to do that. I'll add a TBD to make that clearer.

default/python/AI/AIFleetMission.py
Grummel7: Looking a fleet orders I wonder whether we could simply use fleet_order.is_valid. There is definitely some code duplication here.
But that is not a trivial refactoring, I suggest merging this as it is. _check_abort_mission is not that bad yet.

Cjkjvfnby (approved):

geoffthemedio:
@Cjkjvfnby why did you approve but not merge?

Cjkjvfnby:
> @Cjkjvfnby why did you approve but not merge?

Sorry.  At my job author of PR merges it, so I constantly forget to merge in the FreeOrion project.

PR #3755, opened 2022-03-19
Two fixes

Grummel7:


default/python/AI/ResourcesAI.py
```diff
@@ -96,6 +96,8 @@ def bake_future_focus(self, pid, focus, update=True, force=False):
         Return success or failure
         """
         pinfo = self.raw_planet_info.get(pid)
+        if not pinfo:
+            return
```
Cjkjvfnby: It's better to return False.  None and False behaves similarly in some cases, but different in others (`if variable` vs `variable is False`)

Cjkjvfnby (changes requested):

default/python/AI/ResourcesAI.py
Grummel7: IMHO, and code that checks "variable is False" indicates that there is a problem.
Nevertheless you are right, all other returns return a bool, so I should stick to it. I updated the commit.

Cjkjvfnby (approved):

PR #3762, opened 2022-03-24
Move survey_universe to empire module and call it earlier

### LINKED ISSUE #3760: get_colony_builders, etc.
The new InitializerLock has done its job and identified another premature usage of get_colony_builders. While analysing it, I found a can of worms.

The problem was caused by this [commit](https://github.com/freeorion/freeorion/commit/63e8a2fdf0cc5cb0a6b0c39ca98dd52816721deb). A couple of dicts in ColonizationAI.py have been moved to extra files. Before this commit, the variables have simply been recalculated in `ColonizationAI.survey_universe`. This meant that before survey_universe they had their old values, but they were usually not empty. Accessing those variable before `survey_universe` was always wrong, but it became worse since the values are now cached_for_turn, meaning they are cleared at the beginning of each turn and completely empty before `survey_universe`. 

But they are used before. E.g. `PlanetUtils.get_capital` uses both get_colony_builders and get_ship_builders when the AI has lost its capital and it is called several times before survey_universe, starting right in `FreeOrionAI.generate_orders`. So once an AI has lost its capital, it will throw an exception every turn.

I tried to move `survey_universe` to the top of `generateOrder`, after the following lines and it seems to work. Does anyone see a problem with that move?
```
    if empire.eliminated:
        info("This empire has been eliminated. Aborting order generation.")
        return
```

Grummel7:
This should fix #3760 for good.

It also makes sense as a refactoring, since survey_universe mostly sets values in the empire module and was completely independent from the rest of ColonisationAI.

It does also set some value in the colonization module (nice mix of spelling ;-) which should probably be protected by locks as well, but this can be done with a later PR, this one is big enough and is needed as a fix.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -44,6 +44,7 @@ from character.character_strings_module import (
 )
 from common.handlers import init_handlers
 from common.listeners import listener
+from empire.survey_universe import survey_universe
 from freeorion_tools.timers import AITimer
 
 turn_timer = AITimer("full turn")
@@ -270,6 +271,9 @@ def generateOrders():  # pylint: disable=invalid-name
         info("This empire has been eliminated. Aborting order generation.")
         return
 
+    # results of this function are needed in many places...
+    survey_universe()
+
     generate_order_timer.start("Update states on server")
     # This code block is required for correct AI work.
     info("Meter / Resource Pool updating...")
@@ -343,7 +347,6 @@ def generateOrders():  # pylint: disable=invalid-name
     debug("Calling AI Modules")
     # call AI modules
     action_list = [
-        ColonisationAI.survey_universe,
         ShipDesignAI.Cache.update_for_new_turn,
         PriorityAI.calculate_priorities,
         ExplorationAI.assign_scouts_to_explore_systems,
--- a/default/python/AI/empire/survey_universe.py
+++ b/default/python/AI/empire/survey_universe.py
@@ -1,0 +1,191 @@
+import freeOrionAIInterface as fo
+from logging import info
+
+import AIDependencies
+import AIstate
+from AIDependencies import Tags
+from aistate_interface import get_aistate
+from colonization import rate_piloting_tag, special_is_nest
+from colonization.calculate_planet_colonization_rating import empire_metabolisms
+from colonization.calculate_population import active_growth_specials
+from common.listeners import register_pre_handler
+from common.print_utils import Bool, Sequence, Table, Text
+from empire.buildings_locations import set_building_locations
+from empire.colony_builders import can_build_colony_for_species, set_colony_builders
+from empire.colony_status import (
+    set_colonies_is_under_attack,
+    set_colonies_is_under_treat,
+)
+from empire.growth_specials import set_growth_special
+from empire.pilot_rating import (
+    get_pilot_ratings,
+    set_pilot_rating_for_planet,
+    summarize_pilot_ratings,
+)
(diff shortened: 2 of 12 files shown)


default/python/AI/empire/survey_universe.py
```diff
@@ -126,12 +121,9 @@ def survey_universe():
                 set_colonies_is_under_treat()
 
     survey_universe_lock.unlock()
-    # main lock must be unlocked before get_pilot_ratings is called
-    rating_list = sorted(get_pilot_ratings().values(), reverse=True)
-    summarize_pilot_ratings(rating_list)
-    survey_universe_lock2.unlock()
     survey_timer.stop()
 
+    # lock must be released before get_pilot_ratings is called
```
Cjkjvfnby: Not sure about this comment.

default/python/AI/FreeOrionAI.py
```diff
@@ -343,7 +347,6 @@ def generateOrders():  # pylint: disable=invalid-name
     debug("Calling AI Modules")
     # call AI modules
     action_list = [
-        ColonisationAI.survey_universe,
```
Cjkjvfnby, line 346: Maybe just move `aistate.prepare_for_new_turn()` into the action list after `ColonisationAI.survey_universe`?  

Most of the code before `action_list` is timers and debug prints, so I'd prefer survey universe calls after that.


Actually `prepare_for_new_turn()` is a number of small functions without arguments,  so we could just "inline" it here and put all calls to the list directly.   This will give us the flexibility to adjust orders.

The survey universe does a single cycle over everything and collects information.   
 Algorithmically doing this cycle twice does not look good.  But this could simplify the logic.   If we will have 2 or more cycles we could deal with temporal coupling in much easy way.

default/python/AI/freeorion_tools/lazy_initializer.py
```diff
@@ -21,7 +20,6 @@ def unlock(self):
     def __call__(self, function: Callable):
         @wraps(function)
         def wrapper(*args, **kwargs):
-            print("use zzzxxcc3", fo.currentTurn())
```
Cjkjvfnby, line 24: Oops, my bad.  
This is my favorite prefix for debug prints, it's easy to find it in logs.

Cjkjvfnby (changes requested):

default/python/AI/FreeOrionAI.py
Grummel7: survey_universe must be called before this [line](https://github.com/freeorion/freeorion/blob/df647152ccc177103d82d64e4c19ed9e7591a998/default/python/AI/FreeOrionAI.py#L312) because get_capital needs the information.

I wonder whether we can get rid of survey_universe completely by doing the calculation in the cache_for_current_turn-functions, like it did it for _get_pilot_summary (see second commit). We wouldn't even need the lock anymore then, but of course we would have a dozen loops instead of one.

default/python/AI/empire/survey_universe.py
Grummel7: Actually it is nonsense now, I should have changed it with my second commit.

Cjkjvfnby (approved):

Cjkjvfnby:
@Grummel7 , please apply isort to the code. 

https://github.com/freeorion/freeorion/runs/5702180393?check_suite_focus=true

Grummel7:
> @Grummel7 , please apply isort to the code.
> 
> https://github.com/freeorion/freeorion/runs/5702180393?check_suite_focus=true

Done it. You can merge it now, I cannot do that...

PR #3767, opened 2022-03-27
Add doctest runner

Cjkjvfnby:
Doctest is a quite nice feature to put the test into the documentation. 
https://docs.python.org/3/library/doctest.html

There is a nice part about doctests in that video.  https://www.youtube.com/watch?v=ARKbfWk4Xyw

Since we import runtime libraries, we could not use doctest with that files, since it leads to complicated exclusion rules.   In the next iteration, I'll try to deal with it by mocks.

### DIFF
--- a/.github/workflows/python-check.yml
+++ b/.github/workflows/python-check.yml
@@ -46,7 +46,7 @@ jobs:
           --format="::error file=%(path)s,line=%(row)d,col=%(col)d::%(path)s:%(row)d:%(col)d: %(code)s %(text)s"
       - name: Test with pytest
         # Run from root to preserve full path in output
-        run: pytest -v default/python/tests
+        run: pytest -v default/python
       - name: Test st-tool with pytest
         # Run from root to preserve full path in output
         run: pytest -v check --doctest-modules --doctest-continue-on-failure
--- a/default/python/common/print_utils/_table.py
+++ b/default/python/common/print_utils/_table.py
@@ -16,7 +16,7 @@ def print_in_columns(items: Collection[Any], columns=2, printer=print):
     """
     Split flat list to columns and print them.
 
-    >>> print_in_columns(['a', 'b', 'c', 'd', 2])
+    >>> print_in_columns(['a', 'b', 'c', 'd'], 2)
     a   c
     b   d
     """
--- a/default/python/conftest.py
+++ b/default/python/conftest.py
@@ -1,0 +1,38 @@
+import os
+from pathlib import Path
+
+excluded_paths = tuple(
+    os.path.join(*x)
+    for x in (
+        ("python", "AI"),
+        ("python", "universe_generation"),
+        ("python", "turn_events"),
+        ("python", "common", "configure_logging.py"),
+        ("python", "tests", "fixtures"),
+        ("python", "tests", "htmlcov"),
+        ("python", "auth"),
+        ("python", "chat"),
+        (".pyi",),
+        (".txt",),
+        (".pytest_cache",),
+        (".md",),
+        (".coverage",),
+        (".ini",),
+    )
+)
+
+
+def pytest_ignore_collect(path: Path):
+    """
+    Filter out modules that imports C++ bindings.
+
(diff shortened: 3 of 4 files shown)


Cjkjvfnby:
@Grummel7  could you take a look? 

Docstring allows to write test much easier in simple cases.

hifron:
Are modules independent of each other to make tests from command line possible?

Because code should document itself by its variables, methods, comments, styling to allow easy understanding and also modification when its something wrong and also is its complex value important to easy comprehension.

And if tests are written before its implementation, it also means that it could be tested under conditions not yet implemented or known to be implemented, but also means that test driven development, which in this case FreeOrion is probably not that example.

But AI is currently script way AI under data oriented comprehension, but in future it could be scenario driven and also no single truth way, in which tests this way as docstring seems strange. And if loading user oriented AI competition like in StarCraft way may mean that testing single file(maybe with docstring), but only to allow method docs, not to test it when tests are not primary part of FreeOrion and probably never be...

Cjkjvfnby:
> Are modules independent of each other to make tests from command line possible?

Dependency on each other is not usually a problem, but dependency on freeorion boost libs is.
conftest.py is preventing some modules to be executed for the testing. Unfirtiantly this is the most number of them.

> Because code should document itself by its variables, methods, comments, styling to allow easy understanding and also modification when its something wrong and also is its complex value important to easy comprehension.

And tested.   Sometimes it's a bit hard to express why this code is written via code itself.  In that case, it's nice to have some documentation with examples.  In addition to testing, doctest ensures that your code samples are valied. 

> And if tests are written before its implementation, it also means that it could be tested under conditions not yet implemented or known to be implemented, but also means that test driven development, which in this case FreeOrion is probably not that example.

TDD is not a thing that is visible at the PR stage. In which order you write tests and code does not matter, when you finish. 

> But AI is currently script way AI under data oriented comprehension, but in future it could be scenario driven and also no single truth way, in which tests this way as docstring seems strange.

Tests are only about the present.   they help me to be sure that what I want and what I got are the same thing.  Doctest is just syntax sugar to write unittests. Quite limited way, so it could not replace all unittest.

> And if loading user oriented AI competition like in StarCraft way may mean that testing single file(maybe with docstring), but only to allow method docs, not to test it when tests are not primary part of FreeOrion and probably never be...

Not sure what do you mean.  

We already use unittests for AI, and this is a part of the development process, they run on each PR.  Although coverage is quite small.

hifron:
AI Unit test https://github.com/freeorion/freeorion/pull/2431 as test driven development, but there is also BDD(Behavior-driven development)(Red-Orange-Green semaphore testing)

Cjkjvfnby:
> AI Unit test https://github.com/freeorion/freeorion/pull/2431 as test driven development,

Will take look into that PR, totally forgot about it.  TTD is the process of writing tests. I am not sure that this approach is applicable to existing code with low coverage.  Do you have experience with it?

agrrr3:
> > AI Unit test #2431 as test driven development,
> 
> Will take look into that PR, totally forgot about it. TTD is the process of writing tests. I am not sure that this approach is applicable to existing code with low coverage. Do you have experience with it?

BDD is basically TDD but not concentrating on the tests but on the "behavior" (aka protocols; simulating/mocking lower-layers'/backend-services responses). 
Depending on the task it can be a useful way to write unit tests. So yes, it can be a good tool for existing code regardless of low coverage. It works for development, regression testing and documentation of expectations, especially if interaction is important (e.g. pop after push should return the pushed element) and not getting complex data structures right.
Payoff of writing tests for working/stable code is "always" lower than integrating it into the development cycle early.

Not sure how BDD can help when developing our AI as i do not think we have a lot of backend services there?

Cjkjvfnby:
@hifron @agrrr3 , lets get back to the main topic. 

Doctest is a framework for writing tests.  It gives us syntax to write tests, test discovery, and test runner. 

This PR adds one more way to write tests. A bit limited but when applicable, it gives better results when pytest.

Cjkjvfnby:
Closing as outdated.

PR #3768, opened 2022-03-28
Abandoned first try on: Fix population reducing effect of planetary drive

### LINKED ISSUE #3715: Planetary drive not reducing population
I have just tested moving planets in single-player and it seems that there is no population reduction happening (even though it is announced when you switch to starlane drive focus). - Thinksome

(https://www.freeorion.org/forum/viewtopic.php?p=110269#p110269)

Either this is broken or the message and description are wrong.

19th Multiplayer game

Grummel7:
This should fix #3715.

Sadly the population forecast is now all wrong. The client shows no reduction before the jump, but indicates one afterwards.

I tried keeping the original effect for the forecast, but when doing consecutive jumps, the original effect actually worked and so you'd get the population quartered.

Perhaps we could explain the error in the buildings description...


Grummel7:
#3769 works better...

PR #3769, opened 2022-03-29
Fix population reducing effect of planetary drive

### LINKED ISSUE #3715: Planetary drive not reducing population

Grummel7:
Another attempt to fix #3715, this time using an invisible special. With the second commit, now even the forecast is correct.

The special should possibly get a different picture (single black pixel?).

### DIFF
--- a/default/scripting/specials/planet/STARLANE_DRIVE_ACTIVATED.focs.txt
+++ b/default/scripting/specials/planet/STARLANE_DRIVE_ACTIVATED.focs.txt
@@ -5,31 +5,11 @@ Special
     effectsgroups = [
         EffectsGroup
             scope = Source
-            activation = Not And [
-                Planet
-                Focus type = "FOCUS_PLANET_DRIVE"
-                // FOCUS_PLANET_DRIVE halfs the population, but it only works if the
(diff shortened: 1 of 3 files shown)


geoffthemedio:
tested and works?

Grummel7:
> tested and works?

Not sure who you are asking here. I've tested it with three consecutive jumps and in each case the forecast matched the result, which was normal population growth followed by a loss of half the population.

The two effects have different priority, they'll have to be. One is adding the special that triggers the other, so I guess the special must trigger later, doesn't it? Perhaps I should split the effect in advanced_focus.macros into one that adds the special with earlier priority, so that both the population effect can trigger with the same. And I should put the condition into a macro to avoid repeating it...

geoffthemedio:
So you consider this still a work in progress then?

Grummel7:
Now I consider it finished.

PR #3780, opened 2022-04-03
Convert techs/construction/OUTPOST.focs

o01eg:


parse/EffectPythonParser.cpp
```diff
@@ -188,15 +216,19 @@ void RegisterGlobalsEffects(py::dict& globals) {
 
     globals["SetConstruction"] = py::raw_function(insert_set_meter_<MeterType::METER_CONSTRUCTION>);
     globals["SetDefense"] = py::raw_function(insert_set_meter_<MeterType::METER_DEFENSE>);
+    globals["SetHappiness"] = py::raw_function(insert_set_meter_<MeterType::METER_HAPPINESS>);
```
geoffthemedio, line 219: maybe take this chance to rename it SetStability ?

o01eg: It's still used in old parser as `SetHappiness` so for simpler converting I'd like to leave it without changes.

PR #3784, opened 2022-04-04
Update minimum required boost version to 1.69

o01eg:
Also drop Ubuntu 18.04 which provides 1.65 boost


GG/CMakeLists.txt
```diff
@@ -61,7 +61,7 @@ endif()
 ## Collect project dependencies.
 ##
 
-set(MINIMUM_BOOST_VERSION 1.60.0)
+set(MINIMUM_BOOST_VERSION 1.69.0)
```
Cjkjvfnby, line 64: Whi minimum is not 1.71.0?

o01eg: We decided on that version: https://www.freeorion.org/forum/viewtopic.php?p=111046#p111046

It allows to support Fedora 32 distributive. Do you need something from 1.71?

hifron: Ubuntu 18.04 is the last Ubuntu which supports 32bit platform and it is supported until 26. 4. 2023. After then, only Debian(if hw firmware and soft possible) and although for 32bit platform is maybe time for upgrade, maybe there is some ppa or to compile boost for 18.04 and decision should be for user to decide, but at least there should be some documentation or to use older release...

o01eg: We are going to raise python version too so Ubuntu 18.04 still will be dropped.

.github/workflows/build-ubuntu.yml
```diff
@@ -24,19 +24,15 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04]
-        client-gg: ['ON', 'OFF']
-        client-godot: ['OFF']
+        os: [ubuntu-20.04]
+        client-gg: ['ON']
```
Cjkjvfnby, line 28: Looks like client-gg is always ON.  Should we just remove it from the matrix?

.github/workflows/build-ubuntu.yml
```diff
@@ -24,19 +24,15 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04]
-        client-gg: ['ON', 'OFF']
```
Cjkjvfnby, line 28: Can we leave it `['ON', 'OFF']` and remove `include:` section?

.github/workflows/build-ubuntu.yml
o01eg: We disable it when test for `clang` but it could be tested too.

.github/workflows/build-ubuntu.yml
o01eg: I've changed variants we use.

Cjkjvfnby (approved):

GG/CMakeLists.txt
hifron: In Debian 11 there is Boost 1.74 and Python 3.9

- Therefore it would require manual compile options for Ubuntu 18.04 or
- download current Python and Boost with setting default /etc/alternatives or compile FO dirs

Or upgrade hw...

.github/workflows/build-ubuntu.yml
```diff
@@ -111,7 +107,7 @@ jobs:
           cd build
           cmake --build . --config Release --target unittest
       - name: Run Godot unit tests
-        if: matrix.client-godot == 'ON' && matrix.os == 'ubuntu-20.04'
+        if: matrix.client-godot == 'ON'
```
geoffthemedio, line 110: with line 29 https://github.com/freeorion/freeorion/pull/3784/files#diff-3dc91c5652448e68371fa6d23e2a83da91a66ad49d85a828dcaf8bcab84ded5fR29 is this ever not `ON`?

o01eg: I left variant with Godot OFF for clang.

geoffthemedio:
not sure what I should comment on...

Vezzra (commented):
I can't really contribute much - I can't see anything that looks wrong or broken, at least nothing that would be obvious to me.

The only thing I noticed is that in the `snapcraft.yaml` file you updated the boost libraries to 1.71 instead of 1.69, but I guess that's intentional...?

o01eg:
> The only thing I noticed is that in the `snapcraft.yaml` file you updated the boost libraries to 1.71 instead of 1.69, but I guess that's intentional...?

It was 1.67 which is lower than proposed minimum version and Ubuntu Focal have only 1.67 and 1.71 to choose.

Vezzra (approved):
> It was 1.67 which is lower than proposed minimum version and Ubuntu Focal have only 1.67 and 1.71 to choose.

Ah, ok. That's fine then.

GG/CMakeLists.txt
o01eg: Debian 11 is stable so I think it's fine to switch to it for x86.

hifron: According https://github.com/actions/virtual-environments from [Github docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-software) there is no support for Debian, only by [Docker image](https://hub.docker.com/_/debian?tab=tags).

BTW, https://packages.debian.org could help, but maybe as new pull request(currently is FO on Windows on amd64_x86 arch and things theoretically could broke on x86 architecture unnoticed(as on some other architecture...))

PR #3787, opened 2022-04-05
Python API Stuff

geoffthemedio:
Exposes Species likes/dislikes, and named reals and ints to Python API.

### DIFF
--- a/python/UniverseWrapper.cpp
+++ b/python/UniverseWrapper.cpp
@@ -11,6 +11,7 @@
 #include "../universe/Field.h"
 #include "../universe/FieldType.h"
 #include "../universe/Fleet.h"
+#include "../universe/NamedValueRefManager.h"
 #include "../universe/Pathfinder.h"
 #include "../universe/Planet.h"
 #include "../universe/PopCenter.h"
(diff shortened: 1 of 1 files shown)


geoffthemedio:
@Grummel7 Do you also need enumeration named refs? Might be a bit trickier if all the necessary enum types aren't already exposed.

python/UniverseWrapper.cpp
```diff
@@ -272,6 +273,42 @@ namespace FreeOrionPython {
...
+        //////////////////////////////////////////////
+        py::def("namedRealDefined",
+                +[](const std::string& name) -> bool { return GetValueRef<double>(name, true); },
+                "Returns true/false (boolean) whether there is a defined double-valued "
+                "scripted constant with name (string).");
+        py::def("getNamedReal",
+                +[](const std::string& name) -> double {
+                    auto ref = GetValueRef<double>(name, true);
+                    if (!ref)
+                        return 0.0;
```
Morlic-fo: Can we please not return 0.0 here and in the int case below?

This will inevitably introduce bugs where FOCS constants are renamed but some AI code will just silently use zero in calculations instead of throwing an error (which could be rather impactful like silently multiplying with zero).

I'd rather return ```None``` and have the python code throw if an outdated constant is used in a math context.

IIRC, it should be possible by returning a ```boost::python::object``` instead of a double/int - specifically ```return py::object()``` on failure and wrapping the int/double return value like ```return py::object(retval)``` on success.

python/UniverseWrapper.cpp
```diff
@@ -272,6 +273,42 @@ namespace FreeOrionPython {
...
+                    if (!ref)
+                        return 0;
+                    if (ref->ConstantExpr())
+                        return ref->Eval();
+                    const ScriptingContext context;
+                    return ref->Eval(context);
+                },
+                "Returns the int-valued value of the scripted constant with name (string). "
+                "If no such named constant exists, returns 0.");
+
```
Morlic-fo, line 305: Also should python user really ever care and have to know whether it's an int or a double?
Otherwise, we could combine both for easier usage... Something like this?
```
py::object getFocsConstant(const std::string& name) {
    auto eval = [](auto&& ref) -> py::object {
        if (ref->ConstantExpr()) 
            return py::object(ref->Eval());
        return py::object(ref->Eval(ScriptingContext{}));
    };
    
    if (const auto ref = GetValueRef<double>(name, true))
        return eval(ref);
    else if (const auto ref = GetValueRef<int>(name, true))
        return eval(ref);
    else
        return py::object();
}
```

python/UniverseWrapper.cpp
geoffthemedio: hopefully OK now

Grummel7:
> @Grummel7 Do you also need enumeration named refs? Might be a bit trickier if all the necessary enum types aren't already exposed.

What do you mean by enumeration named refs? Well, we can still add something later on, but with that extensions I have a lot to do for a while.

Just for my understanding: getNamedValue will now return either an int, a double or None, right? Then I don't think I'll need the ...Defined functions. Just like Morlic said: the AI will use the values and if one of them is removed from FOCS, a python exception will tell us the AI must be adapted.

geoffthemedio:
> What do you mean by enumeration named refs?

The are named ref managers for refs of types PlanetEnvironment, PlanetType, Visibility, PlanetSize, UniverseObjectType, and StarType.

Grummel7:
> The are named ref managers for refs of types PlanetEnvironment, PlanetType, Visibility, PlanetSize, UniverseObjectType, and StarType.

The AI interface has the enums (except for UniverseObjectType), that's all I need, I do not want to access the values via strings.

I've merged the PR into my local copy and looks good. You can merge it, I'll update the pyi-files with my next PR.

PR #3645, opened 2022-01-06
Unown outposts with Chaos Wave bombardement weapon

o01eg:

PR #3789, opened 2022-04-06
Also remove ownership of buildings when abandoning an outpost

Grummel7:
A minor fix: abandoning an outpost to safe on influence while keeping its gas giant generator feels like a cheat. Code mostly copied from COLONY_INDEPENDENCE_DECREE.

### DIFF
--- a/default/scripting/buildings/ABANDON_OUTPOST.focs.txt
+++ b/default/scripting/buildings/ABANDON_OUTPOST.focs.txt
@@ -62,6 +62,15 @@ BuildingType
                     empire = Source.Owner
                 SetVisibility empire = Source.Owner visibility = Max(Value, Partial)
             ]
+         EffectsGroup
+            scope = And [
+                (LocalCandidate.PlanetID == Source.PlanetID)
+                Building
+                Not Building name = ThisBuilding
+            ]
+            activation = Turn low = max(1 + Source.System.LastTurnBattleHere, 3 + Source.CreationTurn)
+            priority = [[POPULATION_OVERRIDE_PRIORITY]]
+            effects = SetOwner empire = [[UNOWNED_EMPIRE_ID]]
 
         EffectsGroup
             scope = Source


default/scripting/buildings/ABANDON_OUTPOST.focs.txt
```diff
@@ -62,6 +62,15 @@ BuildingType
                     empire = Source.Owner
                 SetVisibility empire = Source.Owner visibility = Max(Value, Partial)
             ]
+         EffectsGroup
```
geoffthemedio, line 65: does this need to be a separate effects group?

Grummel7: > does this need to be a separate effects group?

I do not want to add SitReps for the building, plus Target.ID wouldn't be a planet-ID for the buildings, so the additional SitReps would be broken. Or the stackinggroup would prevent the effect to be executed more than once, not sure what it is good for in the planet effect. I don't think we can create one SitRep, but change multiple objects in one EffectGroup, can we?

o01eg:
Should buildings have ownership other than ownership of their planets?

Grummel7:
> Should buildings have ownership other than ownership of their planets?

Well, Owner is an attribute of UniverseObject. Maybe we could make Owner virtual and make Building return the planet's owner. But at the moment they have an owner and other FOCS is specified accordingly.

default/scripting/buildings/ABANDON_OUTPOST.focs.txt
geoffthemedio: one can with If/else effects, but it's not the preferred implementation

o01eg:
It means I need to unown buildings here as well #3645

agrrr3:
> Should buildings have ownership other than ownership of their planets?

Yes, i think we currently need it for auto destroy in some cases (e.g. gifting). 
For future we probably want projects to be able to target other planets (e.g. building outpost/colony in supply range without ships, or sabotage one enemy planets). Not sure if we need building objects for that if we introduce projects as new category though.

PR #3791, opened 2022-04-06
Fix _get_queued_base_troopers

Grummel7:
@Cjkjvfnby You were too quick, I just realised my mistake and pushed and update, but you had already merged it.


Cjkjvfnby:
It happens :)

PR #3792, opened 2022-04-06
[AI] Cleanup research

Cjkjvfnby:
A couple of small improvement, it's better to check code commit by commit. 

@geoffthemedio , could you add @Grummel7 to the project, so I could request a review from him?

### DIFF
--- a/default/python/AI/ResearchAI.py
+++ b/default/python/AI/ResearchAI.py
@@ -909,11 +909,9 @@ def generate_classic_research_orders():
                     ]:
                         res = fo.issueEnqueueTechOrder(supply_tech, num_techs_accelerated)
                         debug(
-                            (
-                                "Empire has poor colony/outpost prospects, so attempted to fast-track %s, got result %d",
-                                supply_tech,
-                                res,
-                            )
+                            "Empire has poor colony/outpost prospects, so attempted to fast-track %s, got result %d",
+                            supply_tech,
+                            res,
                         )
             else:
                 pass
@@ -957,11 +955,9 @@ def generate_classic_research_orders():
                         res = fo.issueEnqueueTechOrder(xenoTech, insert_idx)
                         num_techs_accelerated += 1
                         debug(
-                            (
-                                "ANCIENT_RUINS: have an ancient ruins, so attempted to fast-track %s to enable LRN_XENOARCH, got result %d",
-                                xenoTech,
-                                res,
-                            )
+                            "ANCIENT_RUINS: have an ancient ruins, so attempted to fast-track %s to enable LRN_XENOARCH, got result %d",
+                            xenoTech,
+                            res,
                         )
                 research_queue_list = get_research_queue_techs()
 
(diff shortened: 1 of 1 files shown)


default/python/AI/ResearchAI.py
```diff
@@ -704,36 +702,28 @@ def exclude_tech(tech_name):
...
     """generate research orders"""
-    report_adjustments = False
     empire = fo.getEmpire()
     empire_id = empire.empireID
     aistate = get_aistate()
     enemies_sighted = aistate.misc.get("enemies_sighted", {})
     galaxy_is_sparse = ColonisationAI.galaxy_is_sparse()
-    debug("Research Queue Management:")
+
+    empire = fo.getEmpire()
```
Grummel7: That's a duplicate of line 705

default/python/AI/ResearchAI.py
```diff
@@ -849,10 +839,7 @@ def generate_classic_research_orders():
                 ):
                     res = fo.issueEnqueueTechOrder(cc_tech, insert_idx)
                     msg = "Empire is very aggressive, so attempted to fast-track %s, got result %d" % (cc_tech, res)
-                    if report_adjustments:
-                        chat_human(msg)
-                    else:
-                        debug(msg)
+                    debug(msg)
```
Grummel7: You may as well remove msg and simply call:
`debug("Empire is very aggressive, so attempted to fast-track %s, got result %d" % (cc_tech, res))`

Same for all the other cases of course.

Grummel7 (changes requested):
If I ever want to use chat_human like this, I'll put it into a function that that debug and if enabled also chat_human. I don't like if ... else blocks copied all over the place like this.

I think it definitely simplifies the code without removing anything useful. Just go all the way and remove the unneeded msg definitions as well.

default/python/AI/ResearchAI.py
Cjkjvfnby: This operation is called inline. It's a part of the refactoring menu in the IDE.

default/python/AI/ResearchAI.py
```diff
@@ -930,14 +908,13 @@ def generate_classic_research_orders():
...
                         res = fo.issueEnqueueTechOrder(supply_tech, num_techs_accelerated)
-                        msg = (
-                            "Empire has poor colony/outpost prospects, so attempted to fast-track %s, got result %d"
-                            % (supply_tech, res)
+                        debug(
+                            (
+                                "Empire has poor colony/outpost prospects, so attempted to fast-track %s, got result %d",
+                                supply_tech,
+                                res,
+                            )
```
Grummel7, line 916: I'm surprised that neither pycharm nor black handles that:
You could remove one pair of brackets here. Same in line 960 and 1145.

default/python/AI/ResearchAI.py
```diff
@@ -930,14 +908,13 @@ def generate_classic_research_orders():
...
                         )
```
Grummel7: I'm surprised that neither pycharm nor black handles that:
You could remove one pair of brackets here. Same in line 960 and 1145.
(updated code reference)

Grummel7 (approved):

PR #3793, opened 2022-04-06
[AI] Remove alternative approach

Cjkjvfnby:
About 7 years ago there was an attempt to implement an alternative research approach.

It was not finished.  

I suggest removing it.

@Grummel7 , @Morlic-fo , what do you think?

PS. It's a bit hard to track history because of refactorings and other stuff, here is an initial commit  https://github.com/freeorion/freeorion/commit/1fd44bf0382b0bbff8872b4d42bb264c9f5b442e


Grummel7:
I found out only a few days ago. Having a lot of dead code surely doesn't improve anything.
So my opinion is: Go ahead!

agrrr3:
it has been a while since i had a look at the code.
but the "classic" approach (playbook based, mixed snippets and other stuff) is badly performing and hard to maintain (gets worse with every change in contents).
if i remember correctly the alternative approach was using a utility based approach which is much more composable and easier to maintain (similar how the ship designs are evaluated).

of course if nobody makes it work it is not worth anything; i would rather wish for the playbooks to go/keep themselves to the early game
side note: i did not have a look at the implementation of policies but if that is utility based, one could leverage that for utility based research.

Grummel7:
Well, the code isn't gone, if I or anyone else wants to give it another try, we can always get it from the git history. And if I do that one day, I will definitely put it into a different python file.

The policies are rather simple yet, but I doubt that you find an approach that works for both anyway. With research you're dealing with a big tree and you always go forward, the AI has to decide for a direction and how important research is. With policies you are dealing with a limited number of slots, you have much less choices, but you can go back, i.e. you can and sometimes must deadopt a policy for various reasons. Most of the time PolicyAI does nothing but setting the priority for influence.

PR #3794, opened 2022-04-07
Rework Liberty and Conformance

Oberlus:
- Liberty doubles dislike effects and increases RP based on stability: +(stability-5)/5
- Conformance halves dislike effects and increases stability (+2)
- Removed RP malus and infrastructure&stockpiling bonus from Conformance, and stability malus from Liberty
- Grooming: effectsgroups for content like/dislike only need to query SpeciesContentOpinion in scope, not in effects

Discussion: https://www.freeorion.org/forum/viewtopic.php?f=15&t=12335

### DIFF
--- a/default/scripting/buildings/buildings.macros
+++ b/default/scripting/buildings/buildings.macros
@@ -1,14 +1,21 @@
-// sets stability (happiness) on planets in empire that has adopted the policy this macro effect is added to, if those planets' species like or dislike the policy
+// sets stability (happiness) on planets in empire that have the building this macro effect is added to, if those planets' species like or dislike the building
 
-POLICY_LIKE_DISLIKE_SCALING
-'''(1.0 + (Statistic If condition = And [Source EmpireHasAdoptedPolicy empire = Target.Owner name = "PLC_LIBERTY"])
-    - 0.5*(Statistic If condition = And [Source EmpireHasAdoptedPolicy empire = Target.Owner name = "PLC_CONFORMANCE"]))'''
+// Policy Liberty doubles de effect, policy Conformance halves it, values are declared in common/named_values.focs.txt
+POLICY_DISLIKE_SCALING
+'''(1.0 * (NamedRealLookup name = "PLC_LIBERTY_DISLIKE_FACTOR")^(Statistic If condition = And [Source EmpireHasAdoptedPolicy empire = Target.Owner name = "PLC_LIBERTY"])
+        * (NamedRealLookup name = "PLC_CONFORMANCE_DISLIKE_FACTOR")^(Statistic If condition = And [Source EmpireHasAdoptedPolicy empire = Target.Owner name = "PLC_CONFORMANCE"]))'''
 
-STABILITY_PER_LIKED_OR_DISLIKED_BUILDING_ON_PLANET
-'''4.0 * [[POLICY_LIKE_DISLIKE_SCALING]]'''
+STABILITY_PER_LIKED_BUILDING_ON_PLANET
+'''4.0'''
 
-STABILITY_PER_LIKED_OR_DISLIKED_BUILDING_IN_SYSTEM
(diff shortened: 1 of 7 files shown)


Oberlus:
This didn't get any "bad" review. My playtesting didn't show any undesired side effects. Ready to merge.

default/scripting/buildings/buildings.macros
```diff
@@ -31,8 +37,7 @@ SPECIES_LIKES_OR_DISLIKES_BUILDING_STABILITY_EFFECTS
                 (SpeciesContentOpinion species = LocalCandidate.Species name = ThisBuilding < 0)
             ]
             accountinglabel = "DISLIKES_BUILDING_LABEL"
-            effects =
-                SetTargetHappiness value = Value + [[STABILITY_PER_LIKED_OR_DISLIKED_BUILDING_ON_PLANET]] * SpeciesContentOpinion species = Target.Species name = ThisBuilding
+            effects = SetTargetHappiness value = Value - [[STABILITY_PER_DISLIKED_BUILDING_ON_PLANET]]
```
geoffthemedio, line 40: this and similar changes aren't mentioned anywhere...

default/stringtables/en.txt
```diff
@@ -12198,26 +12198,21 @@ PLC_CONFORMANCE
...
+Reduces species dislikes effects and increases stability
 
 PLC_LIBERTY
 Liberty
 
 PLC_LIBERTY_DESC
-'''• Increases the impact of species likes and dislikes on planet stability due to buildings and policies.
-• Increases [[metertype METER_TARGET_RESEARCH]] by [[value PLC_LIBERTY_RESEARCH_BONUS_FLAT]] and decreases [[metertype METER_TARGET_HAPPINESS]] by [[value PLC_LIBERTY_STABILITY_PENALTY_FLAT]] on all planets with a stability of at least [[value PLC_LIBERTY_MIN_STABILITY]].'''
+'''• Doubles the impact of species dislikes on planet stability due to buildings, policies or planetary focus.
+• Increases [[metertype METER_TARGET_RESEARCH]] by 1 RP per each [[value PLC_LIBERTY_RESEARCH_BONUS_DENOMINATOR]] points of [[metertype METER_TARGET_HAPPINESS]] above [[value PLC_LIBERTY_MIN_STABILITY]] on all planets regardless of focus.'''
```
geoffthemedio: the 1 should still be a named value lookup

ideally also "Doubles" but that's a bit awkward to set up

Oberlus: I'll rework that part.

default/scripting/buildings/buildings.macros
Oberlus: It ain't an actual change because result is the same:
Because each effectsgroup was only for liked or disliked content, never both at the same time, the sign in SpeciesContentOpinion is always known (edit: because of the scope), so you can just adjust the operator ('-' instead of '+' for the dislikes effects).
Should I mark it as grooming?

Oberlus:
Reworked as per Geoff's review, and rebased.

Oberlus:
I realized I didn't check the results in the pedia before pushing, I'll find the time today.

Plus I broke the equation for policy dislikes...

Oberlus:
This works. Ready for merge.

Check out the new equation for policy dislike scaling in buildings.macros, policies.macros or happiness.macros.

Old equation was:
1.0 + (LibertyAdopted?) - 0.5*(ConformanceAdopted?)

New equation is:
1.0 * 2^(LibertyAdopted?) * 0.5^(ConformanceAdopted?)

Results are the same, but this allows me to define a named value for the "2" (or whatever) of Liberty to use in the pedia, so that I can show "Increases dislike effects (x[[value LibertyMultiplier]])", which I could not do with the old formula because it uses (LibertyMultiplier-1).

PR #3796, opened 2022-04-10
ProductionAI preparations

Grummel7:
This PR doesn't really change anything yet, but is probably worth a review.

@Cjkjvfnby Can we perhaps have a chat on an evening next week?
I'd like to refactor ProductionAI.generate_production_orders, before including species preferences and new buildings.

### DIFF
--- a/default/python/AI/PlanetUtilsAI.py
+++ b/default/python/AI/PlanetUtilsAI.py
@@ -61,7 +61,7 @@ def get_capital() -> PlanetId:
                 "Nominal Capitol %s does not appear to be owned by empire %d %s"
                 % (homeworld.name, empire_id, empire.name)
             )
-    empire_owned_planet_ids = get_owned_planets_by_empire(universe.planetIDs)
+    empire_owned_planet_ids = get_owned_planets_by_empire()
     peopled_planets = get_populated_planet_ids(empire_owned_planet_ids)
     if not peopled_planets:
         if empire_owned_planet_ids:
@@ -109,16 +109,15 @@ def get_planets_in__systems_ids(system_ids):
     return list(planet_ids)
 
 
-def get_owned_planets_by_empire(planet_ids):
+@cache_for_current_turn
+def get_owned_planets_by_empire() -> List[PlanetId]:
     """
-    Return list of planets owned by empire.
-    :param planet_ids: list of planet ids
-    :return: list of planet ids
+    Return list of all planets owned by empire.
     """
     universe = fo.getUniverse()
     empire_id = fo.getEmpire().empireID
     result = []
-    for pid in planet_ids:
+    for pid in universe.planetIDs:
         planet = universe.getPlanet(pid)
         # even if our universe says we own it, if we can't see it we must have lost it
         if (
--- a/default/python/AI/PlanetUtilsAI.py
+++ b/default/python/AI/PlanetUtilsAI.py
@@ -1,6 +1,6 @@
 import freeOrionAIInterface as fo
 from logging import debug, error
-from typing import Iterable, List, Sequence, Union
+from typing import Iterable, List, Sequence, Set, Union
 
 from AIDependencies import INVALID_ID
 from common.fo_typing import PlanetId, SystemId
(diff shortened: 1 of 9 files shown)


default/python/AI/PlanetUtilsAI.py
```diff
@@ -166,3 +163,40 @@ def get_systems(planet_ids: Sequence[PlanetId]) -> Sequence[SystemId]:
...
+    """
+    Returns sets of empire planets that like, are neutral and dislike the given feature
+    """
+    # default: feature not in any like or dislike set, all neutral
+    default = (set(), set(get_owned_planets_by_empire()), set())
+    return _calculate_get_planet_opinions().get(feature, default)
+
+
+@cache_for_current_turn
+def _calculate_get_planet_opinions():  # -> Dict[str, (Set[PlanetId], Set[PlanetId], Set[PlanetId])]:
```
Grummel7: I couldn't make the typing work, but for a private function it is not that important anyway.

Cjkjvfnby: `Union`?

Cjkjvfnby:
I will be on low internet until Wednesday (traveling).  Let's do it when I get back.

default/python/AI/PlanetUtilsAI.py
Cjkjvfnby:
```suggestion
def _calculate_get_planet_opinions(): -> Dict[str, Tuple[Set[PlanetId], Set[PlanetId], Set[PlanetId]]]``` 
 ```

Comment (Cjkjvfnby) [reply default/python/AI/PlanetUtilsAI.py]: Nut actually we could replace tuple with named tuple.  It's lightweight as a tuple but convenient as a class. 

So instead of accession attributes by index, you could do it with name.

Python 3.6 has a very convenient way to declare it. 
https://docs.python.org/3/library/typing.html#typing.NamedTuple

```python
from typing import NamedTuple, Set

from fo_typing import PlanetId


class Preferences(NamedTuple):
    likes: Set[PlanetId]
    neutral: Set[PlanetId]
    dislikes: Set[PlanetId]
```

default/python/AI/ProductionAI.py
```diff
@@ -1690,12 +1690,14 @@ def build_ship_facilities(bld_name, top_locs=None):
...
     if bld_name in AIDependencies.SYSTEM_SHIP_FACILITIES:
-        current_locs = get_systems_with_building(bld_name)
+        current_locs = get_systems_with_facilities(bld_name)
         current_coverage = current_locs.union(universe.getPlanet(planet_id).systemID for planet_id in queued_bld_locs)
         open_systems = set(
             universe.getPlanet(pid).systemID for pid in get_best_pilot_facilities("BLD_SHIPYARD_BASE")
         ).difference(current_coverage)
-        try_systems = open_systems.intersection(get_systems_with_building(prereq_bldg)) if prereq_bldg else open_systems
+        try_systems = (
+            open_systems.intersection(get_systems_with_facilities(prereq_bldg)) if prereq_bldg else open_systems
```
Cjkjvfnby, line 1699: Maybe move `if` to the upper level and simplify case `open_systems.intersection(open_systems)`.

Cjkjvfnby (commented):
Looks great to merge, left a couple of minor comments.

default/python/AI/PlanetUtilsAI.py
Grummel7: Thank you for the hint, that definitely better.

default/python/AI/ProductionAI.py
Grummel7: Well, I did not touch that line, except that I renamed the function and black did some reformatting.
Real changes in this file will come later.
Btw., would you be available for a chat this evening, e.g. 4pm GMT (7pm Estonian time)? Or maybe even now, if you read it quickly.

Cjkjvfnby: I am not a quick reader. 

@Grummel7 , I'll be in our weekly call meeting for a while.  If you have time jump in.

Cjkjvfnby (approved):

PR #3798, opened 2022-04-15
[AI] Convert ProductionAI to a package

Cjkjvfnby:
This is a [reparation to further refactoring. 

I have converted the module as a file to the module as a package. 

All usages remain the same, but now it's much easier to extract some parts of the code to separate files.  1800 lines are way too much for a single file.


Cjkjvfnby:
Actually, we don't have many usages and already have a production package. 

So we could just move there.

PR #3803, opened 2022-04-19
[AI] Add BuildingType enum

Cjkjvfnby:
Sample PR on how we could use ENuMS to represent the building.

I am not still 100% sure that this approach is much better than just using constants.


agrrr3:
@o01eg Maybe there is some synergy from focs.py transition? I imagine the complete content (e.g. building types) will be available. So maybe AI code could directly reference those?

o01eg:
@agrrr3 No, currently I reproduce tokens from FOCS as close as possible and AI/Server python code isolated from FOCS python code.

default/python/AI/buildings.py
```diff
@@ -0,0 +1,48 @@
+import freeOrionAIInterface as fo
+from enum import Enum
+from typing import List
+
+from common.fo_typing import PlanetId
+from freeorion_tools.caching import cache_for_current_turn, cache_for_session
+
+
+class Building(Enum):
```
Grummel7: Yeah, why not start to use some real objects.
BuildingType would be a better name for this one, I think.

default/python/AI/buildings.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @cache_for_session
+    def get_type(self) -> "fo.buildingType":
+        return fo.getBuildingType(self.value)
+
+    @cache_for_current_turn
+    def available(self, empire: "fo.empire" = None) -> bool:
+        if empire is None:
+            empire = fo.getEmpire()
+        return empire.buildingTypeAvailable(self.value)
```
Grummel7: Is it really worth caching this two functions?

default/python/AI/buildings.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class Building(Enum):
+    """
+    Enum to represent fo building template.
+
+    Pros:
+    - Shorten code a bit
+    - Allow to have shorter names for ids,
+      we could get read of BLD prefix, since it always use with Building. prefix.
```
Grummel7: Well, using the class name as a prefix, names rather get longer. But that is negligible.
Using python names instead of strings has lots of advantages. E.g. IDE can use name completion and show you were a building type is used, typos can be detected by flake.

default/python/AI/buildings.py
```diff
@@ -0,0 +1,48 @@
...
+    - Possible to add automated check that buildings exists and all building are matched.
+
+    Cons:
+    - fo code for getting id returns strings, probably will need to make an adapters.
+
+    Thoughts:
+    - We could replace it in iterative way one constant at the time,
+      this will require to use `value` a lot. When all ids are converted we could get rid of `value` usage.
```
Grummel7: Well, when we wrap building releated fo-functions (as you've done it with available and queue_in), most of the value calls will be done by the class itself.

Grummel7 (approved):
Go ahead with it.
I started working on ProductionAI. When you merge this PR, I can add the buildings I am working on, so my new code is using this class.

default/python/AI/buildings.py
Cjkjvfnby: Probably not.  I just tested that caching works.

default/python/AI/buildings.py
Cjkjvfnby: The idea is to use value only inside the class.  Right now we have the code that uses buildings as a string, and to change this code we should convert chunk of strings to enum.

default/python/AI/ProductionAI.py
```diff
@@ -436,17 +437,14 @@ def generate_production_orders():
...
-                    )
+                    res = building.enqueue(pid)
+                    debug("Enqueueing %s at planet %s, with result %d" % (building, universe.getPlanet(pid).name, res))
                     if res:
                         queued_building_locs.append(pid)
                         cost, time = empire.productionCostAndTime(production_queue[production_queue.size - 1])
                         building_expense += cost / time  # production_queue[production_queue.size -1].blocksize *
                         res = fo.issueRequeueProductionOrder(production_queue.size - 1, 0)  # move to front
-                        debug("Requeueing %s to front of build queue, with result %d", building_name, res)
+                        debug("Requeueing %s to front of build queue, with result %d", building, res)
```
Grummel7, line 447: You could add an option at_front to BuildingType.enqueue and make it return a float, debugging should also be done there, then the whole block could be replaced by:
`building_expense += building.enqueue(pid, at_front=True)`

default/python/AI/buildings.py
```diff
@@ -0,0 +1,68 @@
...
+    It should not contain any decision logic, only provide information that is easy to handle.
+
+    It should not accept or expose any fo.* object.
+
+    Note: It does not required to have exactly the same name and value.
+    """
+
+    SHIPYARD_ENRG_COMP = "BLD_SHIPYARD_ENRG_COMP"
+    SHIPYARD_AST = "BLD_SHIPYARD_AST"
+    ...
```
Grummel7: I guess these dots should be changed into a commented

Grummel7 (approved):
I'll need some more functions, but I can add them myself. Please merge this soon, so I can adapt my code to use it.

default/python/AI/ProductionAI.py
Cjkjvfnby: Let's create a separate function for that in the production code

default/python/AI/buildings.py
Cjkjvfnby: This is an ellipsis, it is often used as noop.

Issue #3800, opened 2022-04-17
Autosave python error at turn 1 load - AI reset to aggressive


abmpicoli2:
* **FreeOrion Version**: 0.4.10.2 and master
* **Operating System**: Windows

Description
-----------
When reloading a single player autosave game, the AI, that was originally set to typical, is reset to aggressive. 

In the logs, the error messages that appear are:

```
17 abr 14:26:24] AI_11: AI_Error: AI script error : FreeOrionAI.py:resumeLoadedGame():133  - AI given empty state-string to resume from; this is expected if the AI is assigned to an empire previously run by a human, but is otherwise an error. AI will be set to Aggressive.
```

Expected Result
---------------
The expected result from an autosave is that the AI selected should be roughly the same. And not show the python errors in the logs.

[FreeOrion-logs.zip](https://github.com/freeorion/freeorion/files/8502158/FreeOrion-logs.zip)

Steps to reproduce
------------------
During single player creation, choose "Typical" in the AI level.
Then reload the autosave from turn 1.

geoffthemedio:
This also seems to still affect master.

As far as I can tell, this only affects the turn 1 start of turn autosave. Autosaves from later turns and manual saves appear to work correctly for me.

Also seemingly working is the autosave from the end of the first turn, which can be enabled in the options.

geoffthemedio:
Problem internally is that the autosave runs without polling the players for their state information. Normally this is fine, as the server has previous state info from players to put into the save, but on the first turn after galaxy generation, AIs haven't yet sent any state info. The server thus saves an empty state string for th AIs. When that save is later loaded, the AIs receive an empty state string, and thus don't know that they were previously set to some other aggression.

@Cjkjvfnby @Morlic-fo @Grummel7 I'm not sure how to get the right save state info. I'd have to get it from the AIs themselves, and can't generate it on the server. The AIs could perhaps send that info back to the server immediately upon starting up, with the initial post-galaxy-creation autosave deferred until that info is received. Or there could be (agin) a message from the server to the AIs that requests a state string...

Morlic-fo:
How about setting save string server side to some special value and if that is passed, just let the AI behave as if it is a new game, i.e. regenerate the aggression.

At least in the current implementation, RNG is seeded by galaxy seed and empire name so it should yield same results when loading the game.

I.e. change
[...]

PR #3804, opened 2022-04-20
Save AI state right after start of a game

Grummel7:
This is a partial fix for #3800. AI now calls setSaveStateString right after starting up.
@geoffthemedio: If you delay the initial autosave until all AIs have called it, that should do.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -81,6 +81,23 @@ def error_handler(func):
     return _error_handler
 
 
+def _save_ai_state():
+    """Serialize (convert to string) global state dictionary and send to AI client to be stored in save file."""
+    import savegame_codec
+
+    info("Preparing for game save by serializing state")
+    try:
+        dump_string = savegame_codec.build_savegame_string()
+        fo.setSaveStateString(dump_string)
+    except Exception as e:
+        error(
+            "Failed to encode the AIstate as save-state string. "
+            "The resulting save file should be playable but the AI "
+            "may have a different aggression. The error raised was: %s" % e,
+            exc_info=True,
+        )
+
+
 def _pre_game_start(empire_id, aistate):
     """
     Configuration that should be done before AI start operating.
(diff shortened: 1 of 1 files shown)


geoffthemedio:
Delaying the initial autosave seems tricky to get right... It's initiated by the human client, which would thus need to know how long to wait... Or the server could wait after getting the message, but depending how long the AIs take to send in info, other messages could arrive during the wait, and it would require setting up something on the server to set a flag that it should do an autosave later when it gets more info, and to check for that flag whenever that info is received...

Also, just setting the AI state string on the AI client doesn't communicate that to the server, so some additional network message will be needed, unless the server queries the AIs for state info during the autosave... perhaps just in cases when it doesn't already have info from them...?

Grummel7:
I wonder if this is even the right way to tackle the issue.

The Max aggression is stored with the game settings, isn't it? Couldn't we just start the AIs as if it was a new game when the same game is from turn 1 and they haven't stored a status yet?

geoffthemedio:
Depending how aggression is picked given the max aggression, picking a new aggression when loading would potentially result in a different assignment of aggressions to AIs on each reload.

Morlic-fo:
> Depending how aggression is picked given the max aggression, picking a new aggression when loading would potentially result in a different assignment of aggressions to AIs on each reload.

Sorry, your discussion happened while I was reading source code for initial issue. Current implementation appears to be properly seeded by saved values. I don't think there is a strong argument to consider future implementations that may not properly seed their RNG...

Grummel7:
Well, even if the aggression level changes, it is just as random as before. When somebody reloads a turn 1 save game, the AIs have not yet done anything after all.

Vezzra:
@Grummel7, closed as abandoned, invalid, ...?

Grummel7:
> @Grummel7, closed as abandoned, invalid, ...?

This idea for a fix was too complicated, see Geoffs comment.
The discussion is ongoing in the linked issue.

PR #3807, opened 2022-04-24
Improve AI handling of invalid savegame strings

### LINKED ISSUE #3800: Autosave python error at turn 1 load - AI reset to aggressive

Morlic-fo:
Fix #3800

This PR does 2 things:
1) Add some special values to save_state_string to allow handling special cases.
2) Move AI aggression settings from C++ to python which allows us to address #3800 python-side.

### DIFF
--- a/default/python/AI/FreeOrionAI.py
+++ b/default/python/AI/FreeOrionAI.py
@@ -45,6 +45,7 @@ from character.character_strings_module import (
 from common.handlers import init_handlers
 from common.listeners import listener
 from empire.survey_universe import survey_universe
+from freeorion_tools import chat_human
 from freeorion_tools.timers import AITimer
 from generate_orders import (
     empire_is_ok,
@@ -81,10 +82,11 @@ def error_handler(func):
     return _error_handler
 
 
-def _pre_game_start(empire_id, aistate):
+def _pre_game_start(empire_id):
     """
     Configuration that should be done before AI start operating.
     """
+    aistate = get_aistate()
     aggression_trait = aistate.character.get_trait(Aggression)
     diplomatic_corp_configs = {
         fo.aggression.beginner: DiplomaticCorp.BeginnerDiplomaticCorp,
@@ -96,8 +98,25 @@ def _pre_game_start(empire_id, aistate):
     configure_debug_chat(empire_id)
 
 
+def _choose_aggression():
+    galaxy_setup_data = fo.getGalaxySetupData()
+    aggression = int(galaxy_setup_data.maxAIAggression)
+
+    if aggression > 0:
+        rng = random.Random()
+        galaxy_seed = hash(galaxy_setup_data.seed)
+        empire_seed = 3 * hash(fo.getEmpire().name)
+        rng.seed(galaxy_seed * empire_seed)
+
+        random_number = rng.randint(0, 99)
+        if random_number > 74:
+            aggression -= 1
+
+    return fo.aggression(aggression)
+
+
 @error_handler
-def startNewGame(aggression_input=fo.aggression.aggressive):  # pylint: disable=invalid-name
+def startNewGame():  # pylint: disable=invalid-name
     """Called by client when a new game is started (but not when a game is loaded).
     Should clear any pre-existing state and set up whatever is needed for AI to generate orders."""
     empire = fo.getEmpire()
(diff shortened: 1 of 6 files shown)


default/python/AI/FreeOrionAI.py
```diff
@@ -122,49 +140,56 @@ def startNewGame(aggression_input=fo.aggression.aggressive):  # pylint: disable=
...
-                " to 'aggressive'. The behaviour of the AI may be different"
-                " than in the original session. The error raised was: %s" % e,
-                exc_info=True,
+                "Failed to load the AIstate from the savegame: %s"
+                " AI can continue but behaviour may be different from the previous session.", e,
+                exc_info=True
             )
-    _pre_game_start(fo.getEmpire().empireID, aistate)
+    if aistate is None:
+        info("Creating new ai state due to failed load.")
```
Grummel7, line 194: Shouldn't this be an error, too?

Grummel7 (approved):
That are more changes than i expected, but it looks like some reasonable cleanup to me

default/python/AI/FreeOrionAI.py
Morlic-fo: I guess it can be argued either way. There are cases where we fail to load and must create a new state without it being a "real" error, e.g. being assigned an empire previously controlled by a human player.

My intention was to have error logs above where we fail unexpectedly but just info here when we recovered.

default/python/AI/FreeOrionAI.py
```diff
@@ -96,8 +98,25 @@ def _pre_game_start(empire_id, aistate):
...
+    galaxy_setup_data = fo.getGalaxySetupData()
+    aggression = int(galaxy_setup_data.maxAIAggression)
+
+    if aggression > 0:
+        rng = random.Random()
+        galaxy_seed = hash(galaxy_setup_data.seed)
+        empire_seed = 3 * hash(fo.getEmpire().name)
+        rng.seed(galaxy_seed * empire_seed)
+
+        random_number = rng.randint(0, 99)
```
Cjkjvfnby, line 111: https://docs.python.org/3/library/random.html#random.choices

```python
aggression _modifier = choices((0, -1), cum_weights=(74, 100))[0]
```

default/python/AI/FreeOrionAI.py
Cjkjvfnby: We implement robustness.  If something went wrong we still try to do our best.  This problem could happen because of some save game incompatibility.

Morlic-fo: @Cjkjvfnby Not sure if you are arguing for ```error``` or ```info``` here.

Any string that isn't specially handled above is sent to the decoder. If invalid/incompatible, an Exception will be raised and logged accordingly. All of that happens before reaching here.

default/python/AI/FreeOrionAI.py
Morlic-fo: While I like to use this function a lot when there are multiple choices or multiple samples involved, I find it adds quite a bit of cognitive overhead in this simple case and prefer the if-statement.

If you have a strong preference, I am of course willing to change it.

Morlic-fo:
I probably won't find the time to come back to this in the next 1-2 weeks. So any open issues will have to wait a bit.

geoffthemedio:
@Morlic-fo reminder

geoffthemedio:
@Morlic-fo reminder

Morlic-fo:
I'm back slightly later than expected, sorry for that.
Can I assume the lack of further comments means this is good to merge? As far as I am concerned, it would be...

PR #3808, opened 2022-04-24
First AI Building refactoring

Grummel7:
Scanning Facility: Don't build it where it is disliked
Gas Giant Generator: Completely new logic
Nearly Universal Translator: Build logic added


default/python/AI/ResearchAI.py
```diff
@@ -546,6 +549,30 @@ def generate_classic_research_orders():
...
+                        and aistate.character.may_research_tech_classic(dt_ech)
+                    ):
+                        res = fo.issueEnqueueTechOrder(dt_ech, insert_idx)
+                        num_techs_accelerated += 1
+                        insert_idx += 1
+                        fmt_str = "Empire wants to build translators, so attempted to fast-track %s (got result %d)"
+                        fmt_str += " with current target_RP %.1f and current pop %.1f, on turn %d"
+                        debug(fmt_str, dt_ech, res, resource_production, empire.population(), fo.currentTurn())
+                research_queue_list = get_research_queue_techs()
+    #
```
Grummel7: This is the same ugly style used by the rest of ResearchAI, I cannot refactor everything I touch.

default/python/AI/buildings.py
```diff
@@ -64,3 +87,29 @@ def production_time(self, planet: PlanetId, empire: EmpireId = THIS_EMPIRE) -> i
         if empire is THIS_EMPIRE:
             empire = fo.empireID()
         return fo.getBuildingType(self.value).productionTime(empire, planet)
+
+    def turn_cost(self, planet: PlanetId, empire: EmpireId = THIS_EMPIRE) -> float:
+        return self.production_cost(planet, empire) / self.production_time(planet, empire)
```
Grummel7: I wonder whether we really need the empire arguments. I doubt that the AI will ever use it.
I know, I should have asked this in the review of the last PR...

default/python/AI/PlanetUtilsAI.py
```diff
@@ -180,10 +199,47 @@ def get_planet_opinion(feature: str) -> Opinion:
...
+@cache_for_current_turn
+def _planned_species() -> Mapping[PlanetId, str]:
+    universe = fo.getUniverse()
+    production_queue = fo.getEmpire().productionQueue
+    planned_species = {}
+    building_prefix = "BLD_COL_"
+    species_prefix = "SP_"
+    colonisation_plans = get_aistate().colonisablePlanetIDs
+    for element in production_queue:
+        if element.name.startswith(building_prefix):
```
Cjkjvfnby: These string operations are mixed with prod inspection queue logic, Let's move it to a separate function.  _is_building(element).

default/python/AI/PlanetUtilsAI.py
```diff
@@ -180,10 +199,47 @@ def get_planet_opinion(feature: str) -> Opinion:
...
+            planned_species[element.locationID] = species_prefix + element.name[len(building_prefix) :]
```
Cjkjvfnby: Lets extract it to a function `_getspecies_for_colony_building(element)`

default/python/AI/PlanetUtilsAI.py
```diff
@@ -180,10 +199,47 @@ def get_planet_opinion(feature: str) -> Opinion:
...
+    for pid in get_owned_planets_by_empire():
+        planet = universe.getPlanet(pid)
+        for building in map(universe.getBuilding, planet.buildingIDs):
+            if building.name.startswith(building_prefix):
```
Cjkjvfnby: What kind of planet is it?  Outpost with colony building, which will be converted on next turn?

default/python/AI/ProductionAI.py
```diff
@@ -1812,3 +1724,204 @@ def get_number_of_existing_outpost_and_colony_ships() -> int:
...
+        planet = universe.getPlanet(pid)
+        if not planet:
+            error(f"Got pid {pid} in candidate, which does not seem to be a planetID")
+            continue
+        if not ignore_dislike and pid in opinion.dislikes:
+            continue
+        if not building_type.can_be_produced(pid) or not building_type.can_be_enqueued(pid):
+            continue
+        if pid in opinion.likes:
+            preferred_locations.append((planet.currentMeterValue(fo.meterType.maxTroops), pid))
```
Cjkjvfnby: Let's extract `planet.currentMeterValue(fo.meterType.maxTroops)` to a function `_location_value(planet)`

default/python/AI/buildings.py
```diff
@@ -39,12 +46,28 @@ def available(self, eid: EmpireId = THIS_EMPIRE) -> bool:
             empire_object = fo.getEmpire(eid)
         return empire_object.buildingTypeAvailable(self.value)
 
+    def get_opinions(self) -> Opinion:
+        return get_planet_opinion(self.value)
```
Cjkjvfnby, line 44: Let's remove this method and change `get_planet_opinion` to accept enum and string(temporary).

Cjkjvfnby (commented):
I'll continue tomorrow

default/python/AI/PlanetUtilsAI.py
```diff
@@ -180,10 +199,47 @@ def get_planet_opinion(feature: str) -> Opinion:
...
+
+
+def _planet_species(pid: PlanetId) -> Optional[fo.species]:
+    universe = fo.getUniverse()
+    planet = universe.getPlanet(pid)
+    species_name = planet.speciesName
+    if not species_name:
+        species_name = _planned_species().get(pid)
+    if species_name:
+        return fo.getSpecies(species_name)
```
Cjkjvfnby, line 251: This function returns a value, so let's explicitly return None.

```suggestion
    if species_name:
        return fo.getSpecies(species_name)
    else:
        return None
```

default/python/AI/PlanetUtilsAI.py
```diff
@@ -180,10 +199,47 @@ def get_planet_opinion(feature: str) -> Opinion:
...
+
+
 @cache_for_current_turn
 def _calculate_get_planet_opinions() -> Dict[str, Opinion]:
     universe = fo.getUniverse()
-    all_species = {universe.getPlanet(pid).speciesName for pid in get_owned_planets_by_empire()}
+    all_species = [universe.getPlanet(pid).speciesName for pid in get_owned_planets_by_empire()]
+    all_species += get_extra_colony_builders()
```
Cjkjvfnby, line 259: `get_extra_colony_builders()`  is it colonies in progress?

default/python/AI/ProductionAI.py
```diff
@@ -1812,3 +1724,204 @@ def get_number_of_existing_outpost_and_colony_ships() -> int:
...
+    # this can be improved a lot, taking into account value of planets, actual stability and
+    # what effects the change would have. For the moment, keep it simple.
+    # Note that the strongest effect is always on the building's planet itself.
+    opinion = building_type.get_opinions()
+    universe = fo.getUniverse()
+    if len(opinion.likes) >= len(opinion.dislikes):
+        like_candidates = opinion.likes - building_type.built_at(True)
+        # plans may change, so consider only actual colonies that like it
+        candidates = [pid for pid in like_candidates if universe.getPlanet(pid).speciesName]
+        return _try_enqueue(building_type, candidates)
```
Cjkjvfnby, line 1820: Should we consider new_turn_cost here?

default/python/AI/ProductionAI.py
```diff
@@ -1812,3 +1724,204 @@ def get_number_of_existing_outpost_and_colony_ships() -> int:
...
+    return 0.0
+
+
+def _build_scanning_facility() -> float:
```
Cjkjvfnby, line 1824: Have you considered using the class approach here? 

```python
class BaseBuildingEnqueue(ABC):
     @abstractmethod
     def get_building_type() -> BuildingType: ...

     def avaliable(self): return self.get_building_type().available()  
     
     @abstractmethod
     def _process_buiding_type(self): ...
     
     def process(self) -> float:
         if not self.avaliable():
            return 0.0   
         return slelf._process()

```        

The `process` will do the work.  

It will have zero type-specific code, just call its methods.   There are two kinds of methods:  abstract, like `_process`.  It will contain code that is unique for all building types.  And with reasonable defaults, so for most classes, it will be the same, but you could still override it for special cases. 

We could merge this code as is, just to have some examples of what use-cases might look like, and convert this to class later.

Cjkjvfnby: With the class approach, we actually could iterate all these objects twice.  The first one is for normal enqueue, and the second for the `_may_enqueue_for_stability`,  to utilize resources that are left from the normal process.

default/python/AI/ProductionAI.py
```diff
@@ -1812,3 +1724,204 @@ def get_number_of_existing_outpost_and_colony_ships() -> int:
...
+    max_scanner_builds = max(1, int(empire.productionPoints / 30)) - len(building_type.queued_in())
+    scanner_systems = building_type.built_at_sys(True)
+    debug(
+        "Considering building %s, found current and queued systems %s, planets that like it %s, #dislikes: %d",
+        building_type,
+        PlanetUtilsAI.sys_name_ids(scanner_systems),
+        PlanetUtilsAI.sys_name_ids(opinion.likes),
+        len(opinion.dislikes),
+    )
+    for sys_id in get_owned_planets().keys():
```
Cjkjvfnby:
```suggestion
    for sys_id in get_owned_planets():
```

Dict is iterated by keys, you don't need explicitly call it here.

default/python/AI/buildings.py
```diff
@@ -39,12 +46,28 @@ def available(self, eid: EmpireId = THIS_EMPIRE) -> bool:
...
+            ret.update(set(self.queued_in()))
+        return ret
+
+    def built_at_sys(self, include_queued_buildings: bool = False) -> Set[SystemId]:
+        ret = _get_building_locations()[self.value].systems
+        if include_queued_buildings:
+            universe = fo.getUniverse()
+            ret.update(set(universe.getPlanet(pid).systemID for pid in self.queued_in()))
+        return ret
+
```
Cjkjvfnby:
```suggestion
        return {universe.getPlanet(pid).systemID for pid in self.built_at(include_queued_buildings=include_queued_buildings)}
```

default/python/AI/PlanetUtilsAI.py
Grummel7: That's species we can build without having a colony, i.e. Exobots, Banforo, etc. The function's doc string explains it.

default/python/AI/buildings.py
Grummel7: Well, actually I think I'll revert this, _get_building_locations()[self.value].systems returns a cached set, using it should be faster then building a new set with every call, shouldn't it?

default/python/AI/ProductionAI.py
Grummel7: Let's do that later. When I've done more buildings, it's easier to see what they all have in common.

default/python/AI/ProductionAI.py
Grummel7: Check if at the top of the function: it returns immediately if new_turn_cost is not 0.

default/python/AI/buildings.py
Grummel7: Looking at my implementation again, I found it was actually broken, since it modified the cached set, so every later call would include the queued locations, regardless of include_queued_buildings.
Updated...

Cjkjvfnby: > Well, actually I think I'll revert this, _get_building_locations()[self.value].systems returns a cached set, using it should be faster then building a new set with every call, shouldn't it?

Cashe is not free, set creation is fast. Keep it simple.    We could add caching later, let's focus on the logic.

Cjkjvfnby (approved):
@Grummel7  How confident are you at merging it?

default/python/AI/buildings.py
Grummel7: I've done the caching already and if we implement your suggestion, we'd use the same cached function indirectly via built_at.

Grummel7:
> @Grummel7 How confident are you at merging it?

I've run several test games and it is definitely an improvement. I'd like to go on now, using the additional policy slot and adding Regional Adminstrations.

I'll have to work on GGG a little, Sly still build too many of them, but at least they don't do it everywhere. Sly in general are not handled very well, although in a pure AI games they survive because it takes virtually forever until any AI gets enough detection to see their colonies. But that is all beyond this PR.

PR #3811, opened 2022-04-29
remove explicit inheritance from object

geoffthemedio:
According to a video I watched, explicitly inheriting from object should be unnecessary after switching to Python3.

Just did a find and replace; didn't test.


Cjkjvfnby (approved):

Cjkjvfnby:
>According to a video I watched, explicitly inheriting from an object should be unnecessary after switching to Python3.

100% true.

Issue #3809, opened 2022-04-25
Regression: Ship values in ship design view is broken (including Estimated combat strength)


agrrr3:
Pedia view for a human frigate object; Both for species  Generic and species human:
```
Estimated combat strength   (ignoring shields):  239   (5.11 per PP)
            (enemy with attack 3.1 and shields 0.0):  239   (5.11 per PP)
```

In design view estimation is zero; Both for species  Generic and species human:
```
Estimated combat strength   (ignoring shields):  0   (0.00 per PP)
            (enemy with attack 3.1 and shields 0.0):  0   (0.00 per PP)
```

agrrr3:
related from the log; 

`
[error] client : CombatDamage.cpp:168 : Combat::WeaponDamageImpl passed source ship without a valid design ID: OBJ_SHIP -2:   at: (-100000.000000, -100000.000000) nearest (System 897) owner: Erdimperium created on turn: 1 specials:   Meters: TargetIndustry: Cur: 0.00 Init: 0.00  TargetResearch: Cur: 0.00 Init: 0.00  TargetInfluence: Cur: 0.00 Init: 0.00  MaxFuel: Cur: 0.00 Init: 0.00  MaxShield: Cur: 0.00 Init: 0.00  MaxStructure: Cur: 0.00 Init: 0.00  Industry: Cur: 0.00 Init: 0.00  Research: Cur: 0.00 Init: 0.00  Influence: Cur: 0.00 Init: 0.00  Fuel: Cur: 0.00 Init: 0.00  Shield: Cur: 0.00 Init: 0.00  Structure: Cur: 0.00 Init: 0.00  Stealth: Cur: 0.00 Init: 0.00  Detection: Cur: 0.00 Init: 0.00  Speed: Cur: 0.00 Init: 0.00   design id: -1 fleet id: -1 species name: SP_HUMAN produced by empire id: 1 arrived on turn: 1 last resupplied on turn: 1
`
a) that is a temporary object (ID -2) and the code returns because there is no valid design ID (either we should not return for the temporary or set a temporary design id)
b) the values are basically unusable

agrrr3:
also those zero values are shown in the UI, only PP cost seems to do the right thing

[...]

PR #3813, opened 2022-04-30
use INCOMPLETE_DESIGN_ID instead of INVALID for the design view

agrrr3:
this solves #3809 by using ID -4 for the incomplete design from the design view and treating it different from invalid designs

it would be better if someone bisected/did analysis to find out how this got broken in the first place (in order to see if something else broke); and there may be better implementations; but this works


agrrr3:
note that this worked before doing merging and cleanup; currently building locally, i hope i can retest this today

agrrr3:
ok, seems to work

agrrr3:
@geoffthemedio can we merge this? i guess it is important for the slow game starting tomorrow (dont have the rights to do that)

PR #3817, opened 2022-05-01
Make changes in MaxPopulation affect growth without delay

Grummel7:
I always considered it a minor issue that populations started to grow only in the second turn after a new tech or exported resource increased a planet's MaxPopulation. Now I learned the reason and found it is rather easy to fix.

### DIFF
--- a/default/scripting/species/common/population.macros
+++ b/default/scripting/species/common/population.macros
@@ -30,12 +30,12 @@ BASIC_POPULATION
                 // growth is GROWTH_RATE_FACTOR * population * (maximum + 1 - population) / maximum
                 // which gives factor for population 1, best growth around half full and aproaching
                 // factor again when almost full
-                ((Value < Target.TargetPopulation) *
-                    min(Target.TargetPopulation, Value + Value*[[GROWTH_RATE_FACTOR]]*
-                                                               (1 + (1 - Value)/Target.TargetPopulation)))
+                ((Value < Value(Target.TargetPopulation)) *
+                    min(Value(Target.TargetPopulation), Value + Value*[[GROWTH_RATE_FACTOR]]*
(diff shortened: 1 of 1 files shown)


agrrr3:
you use the immediate value at priority 40; did you check the priorities of the SetTargetPopulation effects? some hints probably here: https://freeorion.org/index.php/EffectsGroup_Priority_Standard_Values

another question - is the growth prediction correct with your changes?

Grummel7:
> you use the immediate value at priority 40; did you check the priorities of the SetTargetPopulation effects? some hints probably here: https://freeorion.org/index.php/EffectsGroup_Priority_Standard_Values
> 
> another question - is the growth prediction correct with your changes?

The Wiki Page is outdated, plus I did not change the priority, but it seems okay: In fact priorities.macros specifies, that Effects that do SetTargetPopulation shall have priorities between 1 and 30.

I must however use Value() also in the other 4 places where TargetPopulation is accessed in the formula. Which means the PR gets a merge conflict with the other. Well, this one needs a little more closer inspection, I guess.

Forecasts are fine.

agrrr3:
> The Wiki Page is outdated, plus I did not change the priority,

if you do not use the immediate value (Value), priority is not so important, because last turns value will be used anyway

> but it seems okay: In fact priorities.macros specifies, that Effects that do SetTargetPopulation shall have priorities between 1 and 30.

so using the immediate value at 40 should work in principle.

Grummel7:
rebased on master

Grummel7:
@geoffthemedio I think this should go into 0.5, can you please merge it?

geoffthemedio:
I suspect there will be glitchy situations with this, but we can try it...

geoffthemedio:
thanks for patience...

PR #3819, opened 2022-05-02
AI code for regional administration

Grummel7:


default/python/AI/PolicyAI.py
```diff
@@ -45,7 +46,9 @@ def __init__(self):
         self._num_outposts = len(empire_owned_planet_ids) - self._num_populated
         self._centralization_cost = fo.getPolicy(centralization).adoptionCost()
         self._bureaucracy_cost = fo.getPolicy(bureaucracy).adoptionCost()
-        self._max_turn_bureaucracy = 19  # AI doesn't build Regional Admins yet (TBD count, palace may be gone, too)
+        admin_buildings = len(BuildingType.PALACE.built_at()) + len(BuildingType.REGIONAL_ADMIN.built_at())
+        turns_per_building = fo.getNamedValue("BUREAUCRACY_ADMINS_NEEDED_TURN_SCALING")
+        self._max_turn_bureaucracy = admin_buildings * turns_per_building + 9
```
Cjkjvfnby: Let's extract 9 to the variable.  Naming constants remove magic from them.

default/python/AI/ProductionAI.py
```diff
@@ -273,14 +274,9 @@ def generate_production_orders():
...
-            if ("BLD_IMPERIAL_PALACE" in possible_building_types) and (
-                "BLD_IMPERIAL_PALACE" not in (capital_buildings + queued_building_names)
-            ):
-                res = fo.issueEnqueueBuildingProductionOrder("BLD_IMPERIAL_PALACE", homeworld.id)
-                debug("Enqueueing BLD_IMPERIAL_PALACE at %s, with result %d" % (homeworld.name, res))
-                if res:
-                    res = fo.issueRequeueProductionOrder(production_queue.size - 1, 0)  # move to front
-                    debug("Requeueing BLD_IMPERIAL_PALACE to front of build queue, with result %d" % res)
+            building_type = BuildingType.PALACE
+            if building_type.available() and not building_type.built_at(True):
```
Cjkjvfnby: Let's call building_type.built_at() with keyword argument, I cannot guess what True means without looking at the declaration.

default/python/AI/ProductionAI.py
```diff
@@ -1822,7 +1819,7 @@ def _may_enqueue_for_stability(building_type: BuildingType, new_turn_cost: float
 
 
 def _build_scanning_facility() -> float:
-    """Consider building Scanning Facilities"""
+    """Consider building Scanning Facilities, returns added turn costs"""
```
Cjkjvfnby: https://peps.python.org/pep-0257/#one-line-docstrings
> The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

Also, both parts of the sentence should use the same style.

```suggestion
    """Consider building Scanning Facilities, return added turn costs"""
```

default/python/AI/ProductionAI.py
```diff
@@ -1948,3 +1945,69 @@ def _build_translator():
...
+    jumps_to_admin = [
+        (min(universe.jumpDistance(sys_id, admin) for admin in current_admin_sys), sys_id)
+        for sys_id in get_owned_planets()
+    ]
+    jumps_to_admin.sort(reverse=True)
+    # Currently 6 is required, this is supposed to change. Note that the game allows to enqueue several ones
+    # close to each other, but only one will get finished.
+    if jumps_to_admin[0][0] < 6:
+        return 0.0
+    debug(f"current_admin_sys: {current_admin_sys}, jumps_to_admin = {jumps_to_admin}")
```
Cjkjvfnby: admin and sys area bit confusing, let's use full words.  

As I understand this is a candidate system for the administration that is located {jumps} from another administration.

default/python/AI/ProductionAI.py
```diff
@@ -1948,3 +1945,69 @@ def _build_translator():
...
+        if difference > 0:
+            for pid in get_colonized_planets_in_system(other_sys_id):
+                planet = universe.getPlanet(pid)
+                if Tags.INDEPENDENT not in fo.getSpecies(planet.speciesName).tags:
+                    stability = planet.currentMeterValue(fo.meterType.targetHappiness)
+                    rating += difference
+                    # reaching 10 gives a lot of bonuses
+                    if stability < 0 or stability < 10 <= stability + difference:
+                        rating += 2
+    debug(f"admin rating {sys_id}({universe.getSystem(sys_id).name})={rating}")
```
Cjkjvfnby: You could just print `universe.getSystem(sys_id)` instead of id and name.

Cjkjvfnby (changes requested):
Looks great, some minor comments

default/python/AI/PolicyAI.py
Grummel7: That was nonsense, don't know what I was thinking.
It must be (admin_buildings + 1) * turns_per_building - 1.

default/python/AI/ProductionAI.py
Grummel7: I'll do it here, but don't want to change all places. In fact throughout ProductionAI, so far I always use include_queued_building=True, since obviously I don't want to enqueue a building that's already enqueued. Later, other parts of the AI maybe interested in the finished buildings.

default/python/AI/ProductionAI.py
Grummel7: I improved the variable names.
I'll stick to admin because it is used a lot all over the function and it should be rather obvious what it stands for.

default/python/AI/ProductionAI.py
Cjkjvfnby: So just drop this argument from the function and later we will add it back.  You could do it as a separate PR, after this one will be merged.

Grummel7: I decided to use two additional functions now. Note that PolicyAI already uses the basic version to determine the number of (finished) administrative buildings.

Cjkjvfnby (approved):

Issue #3820, opened 2022-05-03
Asteroid coating adding 10% multiple times


Grummel7:
I doubt that this was intended, was it?
![asteroid_coating](https://user-images.githubusercontent.com/20499830/166416882-6d3b9c16-f6eb-492a-b3be-7eb5d56e8ea4.png)
See also https://www.freeorion.org/forum/viewtopic.php?p=111365#p111365

agrrr3:
does this scale with the number of plates, or do you maybe have two specials in your empire?

PR #3824, opened 2022-05-04
Use ASTEROID_COATING_STACK stackinggroup to ensure effect happens once

agrrr3:
fix for #3820 
added stackinggroup to the effect; completely untested


agrrr3:
@Grummel7 can you test this?

Grummel7:
> @Grummel7 can you test this?

Works as expected.

PR #3828, opened 2022-05-07
AI liberty

Grummel7:


default/python/AI/PolicyAI.py
```diff
@@ -70,12 +74,119 @@ def generate_orders(self) -> None:
         debug("End of turn IP: %.2f + %.2f", self._ip, new_production)
         self._determine_influence_priority(new_production)
 
-    def _process_basics(self) -> None:
-        for policy in basics:
-            if policy in self._adoptable:
-                self._adopt(policy)
+    def _process_social(self) -> None:
+        """Process social policies."""
+        self._check_liberty()
```
Cjkjvfnby: `check` term is usually used in the context of validation.  maybe use `handle`?

default/python/AI/PolicyAI.py
```diff
@@ -70,12 +74,119 @@ def generate_orders(self) -> None:
...
+        # Note that this will adopt propaganda in turn 1
+        # TBD check for dislike? algo_research is currently disliked by Chato
+        if self._can_adopt(propaganda):
+            self._adopt(propaganda)
+        if self._can_adopt(algo_research):
+            self._adopt(algo_research)
+
+    def _check_liberty(self) -> None:
+        """Adopt or deadopt liberty, may replace propaganda."""
+        # Liberty is available very early, but to help setting up basic policies we keep propaganda at least
```
Cjkjvfnby: Just put the comment to the docstring.

default/python/AI/PolicyAI.py
```diff
@@ -70,12 +74,119 @@ def generate_orders(self) -> None:
...
+                self.population_stability,
+            )
+
+    def _calculate_empire_output(self) -> _EmpireOutput:
+        # TBD: here we could use the stability-adapted update, that would be much better than trying
+        # to rate the stability effects again production effects.
+        self._universe.updateMeterEstimates(self._populated_planet_ids)
+        result = self._EmpireOutput()
+        for pid in self._populated_planet_ids:
+            planet = self._universe.getPlanet(pid)
```
Cjkjvfnby: Move calculation for a single planet to a function.  To make it simple lets it return new planet output.  

By implementing def __add__(self, other: self._EmpireOutput) you could just add it to the existing (actually create new one as well).


```python
result = self._EmpireOutput(industry=0, reserach=0, influence=0, stability=0)
for pid in self._populated_planet_ids:
    result  += get_planet_output(pid)

File: default/python/AI/PolicyAI.py [hunk 4]
```diff
@@ -70,12 +74,119 @@ def generate_orders(self) -> None:
...
+                population += 1
+                stability -= 1
+            elif stability >= 10:
+                # 10 give a lot of bonuses, increases above 10 are less important
+                stability = 11 + (stability - 10) / 2
+            result.population_stability += population * stability
+        debug(f"Empire output: {result}")
+        return result
+
+    def _compare_empire_output(self, current: _EmpireOutput, new: _EmpireOutput, adoption_cost: float) -> bool:
```
Cjkjvfnby: Let's move it to _the EmpireOutput method.  We could name it `is_better()`

Cjkjvfnby (commented):
@Grummel7 , Looks good, left a couple of comments.

default/python/AI/PolicyAI.py
Grummel7: Actually I'll move it into the "if", since it specifically explains two exclusion I do there.

default/python/AI/PolicyAI.py
Grummel7: I've moved _EmpireOutput out of PolicyManager and moved almost all code to that class. That definitely looks better.

Grummel7:
Hm, there's a check for python3.6 compatibility.
@o01eg Haven't we bumped the requirement to 3.8?

o01eg:
@Grummel7  Not yet, I didn't get enough time to bump Windows/MacOS SDK patches.

default/python/AI/PolicyAI.py
Cjkjvfnby: Than extract code under if to the function and put this comment to docstring.

Grummel7: Well, I've moved the comment to the doc string and slightly amended it. It actually does looks better that way.

default/python/AI/PolicyAI.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import annotations
+# from __future__ import annotations
```
Cjkjvfnby, line 1: Please remove it.  No need to leave commented code.

default/python/AI/PolicyAI.py
```diff
@@ -407,4 +410,4 @@ def generate_policy_orders() -> None:
 
 
 def print_status() -> None:
-    PolicyManager().print_status()
+    PolicyManager(True).print_status()
```
Cjkjvfnby, line 413: Let's use kwargs here `PolicyManager(status_only=True)`.

Cjkjvfnby (approved):
Left a couple of minor comment, please address it in a separate PR.

PR #3829, opened 2022-05-08
Add CI check for CMake files

o01eg:


geoffthemedio:
Anything in particular I should comment on? I know about as much about CMake as @Vezzra

o01eg:
Not sure. I dislike how it formats `# ... #` comments.

hifron:
According it homepage, comments in cmake sources
```
# Maybe things like listing bugs in comment:
#   Bug 1
#   Bug 2
```

should be rewritten in Markdown syntax([cmake-format(GPL3 version)](https://github.com/cheshirekow/cmake_format)), **but easily readable**...

There is also option **to turn** formatting **off** for special comment, but problem is that I think that it is heading towards maybe annotation and comments specific for something like for special compilers, version, variable or something and there is already [some policy](https://cmake.org/cmake/help/latest/policy/CMP0141.html)  in cmake heading towards it, but who knows...

PR #3832, opened 2022-05-09
Add option to use pre-opened file descriptor to accept connections

o01eg:
Xinetd or systemd socket activation allows to start application only when someone connected to its port but the application itself should be able to get accepting socket from provided file descriptor.


geoffthemedio:
Don't understand what it does but code seems fine...

PR #3836, opened 2022-05-14
Minor ai stuff

Grummel7:
Mostly refactoring, plus a little improvement and a bug fix regarding species shield bonus
@Cjkjvfnby As usual, please review...


default/python/AI/AIDependencies.py
```diff
@@ -681,7 +682,9 @@ class Tags:
 SPECIES_POPULATION_MODIFIER = {"EXTREMELY_BAD": 0.25, "VERY_BAD": 0.5, "BAD": 0.75, "GOOD": 1.25}
 SPECIES_SUPPLY_MODIFIER = {"VERY_BAD": -1, "BAD": 0, "AVERAGE": 1, "GREAT": 2, "ULTIMATE": 3}
 SPECIES_FUEL_MODIFIER = {"NO": -100, "BAD": -0.5, "AVERAGE": 0, "GOOD": 0.5, "GREAT": 1, "ULTIMATE": 1.5}
-# missing: INFLUENCE, HAPPINESS, ...
+# values for both offense and defense
+SPECIES_TROOP_MODIFIER = {"NO": 0.0, "BAD": 0.5, "": 1.0, "GOOD": 1.5, "GREAT": 2.0, "ULTIMATE": 3.0}
```
Cjkjvfnby, line 686: Optional: 

I'd like to get away from exposing that stuff at a low level. 


```python
_SPECIES_TROOP_MODIFIER = {"NO": 0.0, "BAD": 0.5, "": 1.0, "GOOD": 1.5, "GREAT": 2.0, "ULTIMATE": 3.0}

def get_species_troop_modifier(modifier: str, default=0.0) -> float:
   return _SPECIES_TROOP_MODIFIER.get(modifier, default)
```

- Smaller API (dict has a lot of ways to interact with it, function only one.)
- Type hints for in and out
- Later you could change this function to reading from scripts, without changing any AI code.

default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -311,6 +313,28 @@ def get_species_tag_grade(species_name, tag_type):
     return get_ai_tag_grade(species.tags, tag_type)
 
 
+@cache_for_session
+def get_species_tag_value(species_name: str, tag_type: AIDependencies.Tags) -> float:
+    """Get a numeric value for a skill-like (GOOD, BAD...) Tag."""
+    grade = get_species_tag_grade(species_name, tag_type)
+    if tag_type == AIDependencies.Tags.INDUSTRY:
```
Cjkjvfnby, line 320: See my suggestion about function instead of the dicts.  Basically, you won't need this wrapper and could import what you need. 

As an alternative solution we could put all `AIDependencies.*_MODIFIER` to dict {AIDependencies.Tags.X, {...} } and expose only that function.

Cjkjvfnby (approved):
Left a couple of comments about future refactoring.

default/python/AI/AIDependencies.py
Grummel7: Perhaps we could replace get_species_tag_value by functions like
`get_species_troop_modifier(species_name: str) -> float`
and put the dicts into a functions. That would also allow us to read the value, if they are turned into NamedReals. That is, at least at the moment, not possible with globally defined dicts, because the AI is started before the server has finished parsing the game rules.

Note that the default is always the value for average, so it only depends on the type of modifier.

o01eg: > That is, at least at the moment, not possible with globally defined dicts, because the AI is started before the server has finished parsing the game rules.

It could be changed, I think. Also I though the server don't start the game until parse resources and doesn't start AI until initialized game.

PR #3837, opened 2022-05-14
AI diversity and artisans

Grummel7:
@Cjkjvfnby Now here is a little more to look at :)


o01eg:
`TypeError: 'type' object is not subscriptable`

Vezzra:
@Grummel7, what should this PR be categorized as? Bugfix, tweak, refactoring....?

Grummel7:
> @Grummel7, what should this PR be categorized as? Bugfix, tweak, refactoring....?

Tweak and refactoring, I guess.  I've thaught AI to make use or to more policies, which I could not do without some refactoring.

Also some kind of bug fix, as the documentation for Artisan Workshops was completely outdated. I only wanted to introduce NamedReals for values the AI was interested in, but then I noticed how wrong the current text was.

default/python/AI/PolicyAI.py
```diff
@@ -79,7 +82,7 @@ def add_planet(self, planet: fo.planet) -> None:
         elif stability >= 10:
             # 10 give a lot of bonuses, increases above 10 are less important
             stability = 11 + (stability - 10) / 2
-        self.population_stability += population * stability
+        self.population_stability += population * stability * 0.75
```
Cjkjvfnby: Please, try to avoid magic numbers.   Extract them to a variable.   Maybe add some comments to them.

default/python/AI/PolicyAI.py
```diff
@@ -195,6 +286,59 @@ def _calculate_empire_output(self) -> _EmpireOutput:
...
+        """Rate diversity."""
+        if diversity not in self._empire.availablePolicies:
+            return 0.0
+        diversity_value = len(get_empire_planets_by_species()) - fo.getNamedValue("PLC_DIVERSITY_THRESHOLD")
+        # diversity affects stability, but also gives a bonus to research-focused planets and a little influence,
+        diversity_scaling = fo.getNamedValue("PLC_DIVERSITY_SCALING")
+        research_priority = self._aistate.get_priority(PriorityType.RESOURCE_RESEARCH)
+        rating = (
+            self._rate_opinion(diversity)
+            # 4 for global influence
```
Cjkjvfnby: global_influence = 4

default/python/AI/PolicyAI.py
```diff
@@ -195,6 +286,59 @@ def _calculate_empire_output(self) -> _EmpireOutput:
...
+            + diversity_scaling * diversity_value * (self._num_populated + 4 + max(0, research_priority - 20))
+        )
+        debug(
+            f"_rate_diversity: rating={rating}. diversity_value={diversity_value}, "
+            f"research priority={research_priority}"
+        )
+        return rating
+
+    def _rate_artisans(self) -> float:
```
Cjkjvfnby, line 304: This function does not fit my screen.  Has nested loops and if condition.  Let's split it into smaller pieces.

default/python/AI/PolicyAI.py
```diff
@@ -336,19 +492,24 @@ def _get_infl_prod(self, update=False) -> float:
...
+        Can we adopt named policy, possibly by replacing (one of) replace?
+        Note that when replace_other is set, this function currently assumes that with
         replace_other removed, there are no other exclusions for policy(name).
         If this won't do, the function needs to temporarily deadopt replace_other or
         policyPrereqsAndExclusionsOK must be extended so that replace_other can be passed to it.
         """
         if name in self._adopted:
             return False
         policy = fo.getPolicy(name)
+        wanted = self._wanted_ip * (name not in self._prioritised_policies)
```
Cjkjvfnby: Boolean multiplication could be tricky.  

```suggestion
        wanted = self._wanted_ip if name in self._prioritised_policies else 0
```

default/python/AI/PolicyAI.py
```diff
@@ -336,19 +492,24 @@ def _get_infl_prod(self, update=False) -> float:
...
         ret = name in self._empire.availablePolicies and (
-            policy.adoptionCost() <= self._ip or name in self._originally_adopted
+            policy.adoptionCost() <= (self._ip - wanted) or name in self._originally_adopted
         )
-        if replace_other and replace_other in self._adopted:
+        if type(replace) == str and replace in self._adopted:
```
Cjkjvfnby: You could use `isinstance(replace, str)`

default/python/AI/PolicyAI.py
```diff
@@ -395,6 +556,24 @@ def _deadopt(self, name: str) -> None:
...
             debug(f"Issued deadoption order for {name} turn {fo.currentTurn()}, remaining IP: {self._ip}")
 
+    def _deadopt_one_of(self, names: Set[str]) -> None:
+        """
+        Deadopt one of the given policies and store its name.
+        This can be used after a decision is made, or to temporarily deadopt one for
+        decision-making. In the later case, use _readopt_selected_one() to revert it.
+        """
+        selection = [(self._rate_policy(name), name) for name in names]
+        selection.sort()
```
Cjkjvfnby:
```suggestion
        selection = sorted((self._rate_policy(name), name) for name in names)
```

default/python/AI/PolicyAI.py
```diff
@@ -395,6 +556,24 @@ def _deadopt(self, name: str) -> None:
...
+        debug(f"selection = {selection}")
+        for _, name in selection:
```
Cjkjvfnby, line 570: Should we stop the cycle after finding one_of?

Cjkjvfnby (changes requested):

default/python/AI/PolicyAI.py
Grummel7: Actually it was more dangerous in dislike_multiplier than here.
Unless I am mistaken the problem is that something that looks like a bool is actually something else. But the result of a "not in" operator will always be a bool, won't it? Anyway, the alternative works fine here, too.

default/python/AI/PolicyAI.py
Grummel7: Definitely, this is a real bug!

default/python/AI/PolicyAI.py
Grummel7: First I thought: why not...
```
for pid in planets:
    rating += self._rate_artisan_planet(...)
```
...but to do that I'll have to pass a whole screen line of arguments to that function.
I'd rather improve the documentation. I'll improved the function a bit, too.

Grummel7: Well, not if I put the getNamedValue calls into the function. The values should one day become global variables anyway.

Cjkjvfnby: > but to do that I'll have to pass a whole screen line of arguments to that function

This is a sign of complexity.  When you spread it across the big function it's not so visible, but when you start extraction, you will hit it hard.

That is exactly what I want to avoid in a new code.

default/python/AI/PolicyAI.py
Cjkjvfnby: It's not about technical stuff, it's about implementation and intention. 

`self._wanted_ip if name in self._prioritised_policies else 0` your intention is to define value by condition and implementation does exactly that. That is the simplest implementation.

Cjkjvfnby (approved):

PR #3838, opened 2022-05-16
Test FreeOrion with Fedora 32 in Docker

o01eg:
Add test for minimum supported Boost version 1.69 and minimum supported python version 3.8


geoffthemedio:
I see a ✔️

PR #3839, opened 2022-05-17
fix system ship unstealthiness

agrrr3:
lienrag found the fleet unstealthiness to misbehave if in transit. (e.g.  [screenshot in forum](https://freeorion.org/forum/viewtopic.php?p=111772#p111772) )

if a candidate is not in a system (has SystemID -1) it matches nothing. the -1 as target is used to signal "match any system" (i.e. an SystemID but -1). 

not sure if the implementation of InSystem is a bug or if it is correct and simply not helpful/not intuitive. Not Not InSystem id = xxx probably has to be the same as InSystem id = xxx for all xxx .

so for fleet unstealthiness code. the ship the stealth effect applies to is the target. the ships which are counted are the candidates.
according to my code dive the stealth of a ship on a starlane would be decreased depending on the count of all ships which are not on a starlane - which is clearly a bug.

the scope condition needs an InSystem condition (without id).

this PR contains three commits; one fixes a typo; one fixes the unstealthiness ; one moves the effect to a more fitting focs file

### DIFF
--- a/default/scripting/techs/spy/PLANET_STEALTH_MOD.focs.txt
+++ b/default/scripting/techs/spy/PLANET_STEALTH_MOD.focs.txt
@@ -20,6 +20,7 @@ Tech
         EffectsGroup
             scope = And [
                 Ship
+                InSystem
                 OwnedBy empire = Source.Owner
             ]
             accountinglabel = "FLEET_UNSTEALTHINESS"
--- a/default/scripting/techs/spy/PLANET_STEALTH_MOD.focs.txt
+++ b/default/scripting/techs/spy/PLANET_STEALTH_MOD.focs.txt
@@ -15,27 +15,7 @@ Tech
             ]
             stackinggroup = "PLANET_STEALTH_MOD_STACK"
             accountinglabel = "BASIC_PLANET_STEALTH"
-            effects = SetStealth value = Value + (NamedReal name = "PLANET_BASIC_STEALTH" value = 5.0)
-        
-        EffectsGroup
-            scope = And [
-                Ship
-                InSystem
-                OwnedBy empire = Source.Owner
-            ]
-            accountinglabel = "FLEET_UNSTEALTHINESS"
-            effects = SetStealth value = Value - (NamedReal name = "FLEET_UNSTEALTH_SHIPS_SCALING" value = 5.0) *
-                floor(
-                    max(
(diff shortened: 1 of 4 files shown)


geoffthemedio:
> Not Not InSystem id = xxx probably has to be the same as InSystem id = xxx for all xxx

Not sure what your point is...

agrrr3:
> > Not Not InSystem id = xxx probably has to be the same as InSystem id = xxx for all xxx
> 
> Not sure what your point is...

probably my point was i am not sure that the InOrIsSystem::Eval works correctly for the both search domains. but probably it does. and changing it (see below) would also be fine.

@geoffthemedio the underlying question was if we should change the behaviour of the condition not to interpret -1 as "match any system"; probably by using a different int constant (e.g. ANY_SYSTEM_ID -2)

geoffthemedio:
> probably my point was i am not sure that the InOrIsSystem::Eval works correctly for the both search domains.

What are the two search domains you mean?

> the underlying question was if we should change the behaviour of the condition not to interpret -1 as "match any system"; probably by using a different int constant (e.g. ANY_SYSTEM_ID -2)

This is how most / all conditions and valuerefs work. ID `-1` means "no particular object", and may result in matching anything or nothing, depending on the situation. Specifying or using a valueref that evaluates to `-1` is also often different from specifying nothing, and sometimes one needs to include a no parameter version of a condition or apply earlier filtering to be sure one is getting a valid object ID and not `-1` before passing it on...

agrrr3:
> > probably my point was i am not sure that the InOrIsSystem::Eval works correctly for the both search domains.
> 
> What are the two search domains you mean?

matches and non_matches

> 
> > the underlying question was if we should change the behaviour of the condition not to interpret -1 as "match any system"; probably by using a different int constant (e.g. ANY_SYSTEM_ID -2)
> 
> This is how most / all conditions and valuerefs work. ID `-1` means "no particular object", and may result in matching anything or nothing, depending on the situation. Specifying or using a valueref that evaluates to `-1` is also often different from specifying nothing, and sometimes one needs to include a no parameter version of a condition or apply earlier filtering to be sure one is getting a valid object ID and not `-1` before passing it on...

ok, so InSystem id = Target.SystemID matching any object on a starlane is expected behavior. I added documentation entries in the wiki. case closed.

PR #3841, opened 2022-05-18
Add common/opinion.macros for liberty/conformance dislike scaling effects

Oberlus:
Fixes duplicated key in buildings/policies/common.happiness.macros

### DIFF
--- a/default/scripting/buildings/buildings.macros
+++ b/default/scripting/buildings/buildings.macros
@@ -1,13 +1,9 @@
 // sets stability (happiness) on planets in empire that have the building this macro effect is added to, if those planets' species like or dislike the building
 
-// Policy Liberty doubles de effect, policy Conformance halves it, values are declared in common/named_values.focs.txt
-POLICY_DISLIKE_SCALING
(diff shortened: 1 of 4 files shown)


Oberlus:
This needs debate, maybe.
I chose to keep a single macro definition in a single file for the three use cases, instead of giving three different names to the same macro declared in three different files.

I added a new macros file because it seems incorrect to reuse any of the others (including unrelated stuff).

And I included that macros file in the three original macros files. I didn't find other uses of macros including macros, but it seems to work: Quickstarted a human game, got Liberty and Protection focus, the default malus for disliked focus (-2) was doubled (-4).

geoffthemedio:
> didn't find other uses of macros including macros

You mean like https://github.com/freeorion/freeorion/blob/master/default/scripting/ship_hulls/organic/organic.macros#L32 ?

Oberlus:
> > didn't find other uses of macros including macros
>
> You mean like https://github.com/freeorion/freeorion/blob/master/default/scripting/ship_hulls/organic/organic.macros#L32 ?

Yes. So it was meant to work OK.

Well, this is ready to merge, unless you prefer some other place to put the macro.

geoffthemedio:
I think you forgot to add and commit the new file

Oberlus:
Fixed.

PR #3843, opened 2022-05-19
Removed unused AVERAGE_OFFENSE_TROOPS macro

Oberlus:
This macro is not used anywhere, except in Fulver species where it does nothing.
And it is annoying that it matches searches for SetTargetResearch (used in the useless macro to not have an empty effect, ugly hack).
If average offense troops does ever need an actual effect, it can be created then.

Edit: made the effect NoOp, removed it from Fulver (unneded).

### DIFF
--- a/default/scripting/species/SP_FULVER.focs.txt
+++ b/default/scripting/species/SP_FULVER.focs.txt
@@ -65,7 +65,6 @@ Species
         [[BAD_HAPPINESS]]
         [[AVERAGE_SUPPLY]]
         [[BAD_DEFENSE_TROOPS]]
-        [[AVERAGE_OFFENSE_TROOPS]]
         [[GOOD_WEAPONS]]
         [[GREAT_FUEL]]
 
--- a/default/scripting/species/common/troops.macros
+++ b/default/scripting/species/common/troops.macros
@@ -279,13 +279,7 @@ BAD_OFFENSE_TROOPS
             ]
 '''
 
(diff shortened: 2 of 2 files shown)


geoffthemedio:
> it matches searches for SetTargetResearch (used in the useless macro to not have an empty effect, ugly hack).

Particularly as there is a `NoOp` effect...

geoffthemedio:
You could leave the macro but make it empty. Arguably that's documenting in the species script that it wasn't forgotten, but rather explicitly made empty.

Oberlus:
Ready to merge

geoffthemedio:
I'd rather keep the macro in the species description but make it empty... no need to run activation and scope checks for a NoOp, and having the macro is more self-documenting in the species script

Oberlus:
> I'd rather keep the macro in the species description but make it empty... no need to run activation and scope checks for a NoOp, and having the macro is more self-documenting in the species script

Are you suggesting to add the macro to all the species that are average offense troops?
Because only Fulver had the macro (I guess Ophi thought it had to be used when he created the species), but it makes no effect at all, not in the Pedia, not in the species descrioption, not anywhere.

geoffthemedio:
I didn't think there would be many species that have no relevant macro assigned, and that it would just be this one. Probably that's not necessary then and you can just remove it entirely.

Oberlus:
Done, completely removed the macro.

> documenting in the species script that it wasn't forgotten

I left a comment in place for the macro

PR #3844, opened 2022-05-20
Changed Bureaucracy after forum discussion.

Grummel7:
Another forum discussion: https://www.freeorion.org/forum/viewtopic.php?f=28&t=12420&sid=ccc146868760118459489fbe4702cf9a

### DIFF
--- a/default/python/AI/PolicyAI.py
+++ b/default/python/AI/PolicyAI.py
@@ -116,7 +116,7 @@ class PolicyManager:
         self._populated_planet_ids = PlanetUtilsAI.get_populated_planet_ids(empire_owned_planet_ids)
         self._num_populated = len(self._populated_planet_ids)
         self._num_outposts = len(empire_owned_planet_ids) - self._num_populated
-        self._max_turn_bureaucracy = self._calculate_max_turn_bureaucracy()
+        self._max_turn_bureaucracy = 9999
         self._centralization_cost = fo.getPolicy(centralization).adoptionCost()
         self._bureaucracy_cost = fo.getPolicy(bureaucracy).adoptionCost()
         self._wanted_ip = self._wanted_for_bureaucracy()
--- a/default/python/AI/PolicyAI.py
+++ b/default/python/AI/PolicyAI.py
@@ -8,7 +8,6 @@ from typing import Iterable, Optional, Set, Union
 import PlanetUtilsAI
 from AIDependencies import Tags
 from aistate_interface import get_aistate
-from buildings import BuildingType
 from common.fo_typing import PlanetId, SpeciesName
 from EnumsAI import FocusType, PriorityType
 from freeorion_tools import assertion_fails, get_species_tag_value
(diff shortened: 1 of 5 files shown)


default/stringtables/en.txt
```diff
@@ -12286,7 +12278,7 @@ Capital Markets
 
 PLC_CAPITAL_MARKETS_DESC
 '''• Planets with a luxury resource special and that are set to influence focus will export the resource to supply-connected planets. This boosts the stability of planets with species that like the resource by [[value SPECIAL_CAPITAL_MARKETS_STABILITY_FOCUS_BONUS]].
-• Increases [[metertype METER_TARGET_INFLUENCE]] on populated planets with at least [[value CAPITAL_MARKETS_INFLUENCE_MIN_STABILITY]] [[metertype METER_HAPPINESS]] by [[value CAPITAL_MARKETS_INFLUENCE_BONUS_SCALING]] per the lesser of [[metertype METER_CONSTRUCTION]] and [[metertype METER_POPULATION]].
+• Increases [[metertype METER_TARGET_INFLUENCE]] on influence-focused planets with at least [[value CAPITAL_MARKETS_INFLUENCE_MIN_STABILITY]] [[metertype METER_HAPPINESS]] by [[value CAPITAL_MARKETS_INFLUENCE_BONUS_SCALING]] per the lesser of [[metertype METER_CONSTRUCTION]] and [[metertype METER_POPULATION]].
```
Grummel7, line 12287: Unrelated documentation mistake. Here I am pretty sure which is right, at least granting that bonus without requiring influence focus sounds OP to me.

geoffthemedio:
CodeQL build fail should be fixed by a rebase on master

Grummel7:
> CodeQL build fail should be fixed by a rebase on master

This PR does not change any C++, I suspect it was tested against master before your commit "fix dummy parsers".

Grummel7:
What the hack went wrong now?
I updated to freeorion/master (reset --hard), re-added my changes as a single commit and now the PR show dozens of commits.

Grummel7:
Argh, git merge...
Now it looks better.

default/python/AI/PolicyAI.py
```diff
@@ -116,7 +116,7 @@ def __init__(self, status_only: bool = False):
         self._populated_planet_ids = PlanetUtilsAI.get_populated_planet_ids(empire_owned_planet_ids)
         self._num_populated = len(self._populated_planet_ids)
         self._num_outposts = len(empire_owned_planet_ids) - self._num_populated
-        self._max_turn_bureaucracy = self._calculate_max_turn_bureaucracy()
+        self._max_turn_bureaucracy = 9999
```
Cjkjvfnby: `_calculate_max_turn_bureaucracy` is not used anymore.  Maybe move 9999 inside it? 

@Grummel7 , whom would be the best person to review FOCS changes?

Cjkjvfnby (changes requested):

default/python/AI/PolicyAI.py
Grummel7: Well, this is just a quick fix to avoid the AI from regularly replacing bureaucracy.
Once the PR is merged, I want to overhaul the entire centralization / bureaucracy stuff.

Usually Geoff does checks this kind of FOCS changes, but I guess at least Oberlus or agrrr3 could review the FOCS change as well.

Cjkjvfnby: > Well, this is just a quick fix to avoid the AI from regularly replacing bureaucracy.

leaving dead code is bad practice, even if you have a plan to deal with it later.    So let's just move 9999 inside and add a comment on why it is so.     

Basically, things that you said in that comment could be just a part of the code.

Oberlus (approved):

Oberlus:
The only thing that IMO could need some tweaking is the actual equation for IP cost per turn of bureacracy. It seems OK to me as it is, we'll see with more playtesting.
I've played with this PR a game until turn 150, keeping Centralization also. I liked it, except that I see the extra IP cost from Centralization is still too high. But that is left for the overhaul @Grummel7 mentioned.

Cjkjvfnby (approved):

PR #3850, opened 2022-05-25
Remove checks and code for boost older 1.69

o01eg:


Vezzra (approved):
FO builds and runs with this PR on macOS.

PR #3851, opened 2022-05-26
AI military policies

Grummel7:


default/python/AI/PolicyAI.py
```diff
@@ -196,18 +232,38 @@ def _unlocked_by(name: str) -> Optional[str]:
             return diversity
         return None
 
+    _rating_functions = {
+        propaganda: lambda self: 20 + self._rate_opinion(propaganda),
+        algo_research: lambda self: self._rate_algo_research(),
```
Cjkjvfnby: Let's simplify this.

- exclude self from the arguments.  Since we are inside the class, we don't need to pass it once again.
- replace simple lambdas with the methods.


```suggestion
        algo_research: self._rate_algo_research,
```

default/python/AI/PolicyAI.py
```diff
@@ -196,18 +232,38 @@ def _unlocked_by(name: str) -> Optional[str]:
...
-        if name == propaganda:
-            return 20 + self._rate_opinion(name)
-        if name == algo_research:
-            return self._rate_algo_research()
-        if name == diversity:
-            return self._rate_diversity()
-        if name == artisans:
-            return self._rate_artisans()
-        raise (ValueError(f"_rate_policy {name} not yet supported"))
+        f = self._rating_functions.get(name)
```
Cjkjvfnby: `f` -> `rating_function`

default/python/AI/PolicyAI.py
```diff
@@ -196,18 +232,38 @@ def _unlocked_by(name: str) -> Optional[str]:
...
+        if assertion_fails(f, f"_rate_policy({name}) not yet supported"):
+            return 0.0
+        return f(self)
+
+    def _rate_costs(self, costs: float) -> float:
+        """
+        How do we rate the costs in comparison to a policies rating?
+        Generally costs / 3, but less if we have plenty if IP.
```
Cjkjvfnby:
```suggestion
        Generally costs / 3, but less if we have plenty of IP.
```

default/python/AI/PolicyAI.py
```diff
@@ -196,18 +232,38 @@ def _unlocked_by(name: str) -> Optional[str]:
...
+        """
+        unreserved_ip = self._ip - self._wanted_ip
+        cost_or_prio = (costs + self._aistate.get_priority(PriorityType.RESOURCE_INFLUENCE)) / 2
```
Cjkjvfnby: `cost_or_prio`   not sure what you mean by or in that name.

Cjkjvfnby (commented):
Look great, left a couple of cosmetic comments.

default/python/AI/PolicyAI.py
Grummel7: I'd have to change _rating_functions into a function for this work, which also means that the dict is created anew for each call. The good thing about making self an argument to the lamdas is, that the dict is static.

default/python/AI/PolicyAI.py
Grummel7: I had problems coming up with a good name, but actually when looking at it now, the formula made not sense anyway. Rewrote it and added a comment. It's probably one of the places that may be tweaked again.

default/python/AI/PolicyAI.py
Cjkjvfnby: You could move it to `__init__`, so it would be created once.

default/python/AI/PolicyAI.py
```diff
@@ -196,18 +232,39 @@ def _unlocked_by(name: str) -> Optional[str]:
...
+        return rating_function(self)
+
+    def _rate_costs(self, costs: float) -> float:
+        """
+        How do we rate the costs in comparison to a policies rating?
+        Generally costs / 3, but less if we have plenty of IP.
+        """
+        unreserved_ip = self._ip - self._wanted_ip
+        priority = max(self._aistate.get_priority(PriorityType.RESOURCE_INFLUENCE), 1)  # in turn 1 priority is 0
+        # For big numbers this makes 1.5 * priority an upper limit for the rating
```
Cjkjvfnby: Not sure that I got this comment, could you rephrase it?

Grummel7: I refactored the whole function. Using min instead of costs / max it should be easier to read and I don't need the special handling for turn 1 anymore.

default/python/AI/PolicyAI.py
Grummel7: > You could move it to `__init__`, so it would be created once.

That's a good idea, done...

default/python/AI/PolicyAI.py
```diff
@@ -122,6 +122,21 @@ def __init__(self, status_only: bool = False):
         self._wanted_ip = self._wanted_for_bureaucracy()
         self._adoptable = self._get_adoptable()
         self._available = set(self._empire.availablePolicies)
+        self._rating_functions = {
+            propaganda: lambda: 20 + self._rate_opinion(propaganda),
+            algo_research: self._rate_algo_research,
+            diversity: self._rate_diversity,
+            artisans: self._rate_artisans,
+            # military policies are mostly chosen by opinion, plus some rule-of-thumb values
+            allied_repair: lambda: self._rate_opinion(charge),  # no effect, unless we have allies...
```
Cjkjvfnby, line 131: For such cases you could use [partilal](https://docs.python.org/3/library/functools.html#functools.partial)
```suggestion
            allied_repair: partial(self._rate_opinion, charge),  # no effect, unless we have allies...
```

Cjkjvfnby (approved):

PR #3854, opened 2022-05-28
Compress message data

mlam19:
This adds compression/decompression for message data. When a network game is created, or when a client joins a game, the endpoints communicate whether they may accept/receive compressed data; uncompressed data is always accepted. The message header is expanded to include a flag indicating whether or not the data is compressed. Thus, an endpoint may receive a mix of compressed and uncompressed data.

By default, both client and server default to accepting compressed data. Client-side acceptance may be modified by the user.

Data will always be sent uncompressed - overriding default and user options - in the following cases:
1. The receiving endpoint is on localhost
2. The size of the data (in bytes) is smaller than Message::COMPRESSION_THRESHOLD (currently set to 64.)

### Notes
- The change to the message headers likely breaks backwards compatibility.
- The default zlib settings are used; the size of large data may be reduced by 70-80%.

### Resources
- [FreeOrion forum thread](https://www.freeorion.org/forum/viewtopic.php?f=9&t=12226)

### DIFF
--- a/client/ClientNetworking.cpp
+++ b/client/ClientNetworking.cpp
@@ -223,7 +223,7 @@ public:
     void SetHostPlayerID(int host_player_id);
 
     /** Sets compression usage for outgoing messages. */
-    void SetCompressionUse(bool use_cpr);
+    void SetUseCompression(bool use_compression);
 
     /** Get authorization roles access. */
     Networking::AuthRoles& AuthorizationRoles();
@@ -259,7 +259,6 @@ private:
     boost::asio::high_resolution_timer m_reconnect_timer;
     tcp::resolver::iterator            m_resolver_results;
     bool                               m_deadline_has_expired{ false };
-    bool                               m_localhost_server;
 
     // m_mutex guards m_incoming_message, m_rx_connected and m_tx_connected which are written by
     // the networking thread and read by the main thread to check incoming messages and connection
@@ -290,7 +289,6 @@ ClientNetworking::Impl::Impl() :
     m_socket(m_io_context),
     m_deadline_timer(m_io_context),
     m_reconnect_timer(m_io_context),
-    m_localhost_server(false),
     m_incoming_messages(m_mutex),
     m_use_compression(false)
 {}
@@ -431,7 +429,6 @@ bool ClientNetworking::Impl::ConnectToServer(
     }
     if (IsConnected()) {
         m_destination = ip_address;
-        m_localhost_server = m_socket.remote_endpoint().address().is_loopback();
     }
     TraceLogger(network) << "ClientNetworking::Impl::ConnectToServer() - Returning.";
     return IsConnected();
@@ -478,8 +475,8 @@ void ClientNetworking::Impl::SetPlayerID(int player_id) {
 void ClientNetworking::Impl::SetHostPlayerID(int host_player_id)
 { m_host_player_id = host_player_id; }
 
-void ClientNetworking::Impl::SetCompressionUse(bool use_cpr)
-{ m_use_compression = use_cpr; }
+void ClientNetworking::Impl::SetUseCompression(bool use_compression)
+{ m_use_compression = use_compression && !m_socket.remote_endpoint().address().is_loopback(); }
 
 Networking::AuthRoles& ClientNetworking::Impl::AuthorizationRoles()
 { return m_roles; }
@@ -489,7 +486,7 @@ void ClientNetworking::Impl::SendMessage(Message&& message) {
         ErrorLogger(network) << "ClientNetworking::SendMessage can't send message when not transmit connected";
         return;
     }
-    if (!m_localhost_server && m_use_compression && message.Size() >= Message::COMPRESSION_THRESHOLD) {
+    if (m_use_compression && message.Size() >= Message::COMPRESSION_THRESHOLD) {
         TraceLogger(network) << "ClientNetworking::SendMessage() : sending compressed message " << message;
         message.Compress();
     } else {
@@ -843,8 +840,8 @@ void ClientNetworking::SetPlayerID(int player_id)
 void ClientNetworking::SetHostPlayerID(int host_player_id)
 { return m_impl->SetHostPlayerID(host_player_id); }
 
-void ClientNetworking::SetCompressionUse(bool use_cpr)
-{ m_impl->SetCompressionUse(use_cpr); }
+void ClientNetworking::SetUseCompression(bool use_compression)
+{ m_impl->SetUseCompression(use_compression); }
 
 Networking::AuthRoles& ClientNetworking::AuthorizationRoles()
 { return m_impl->AuthorizationRoles(); }
--- a/client/ClientNetworking.cpp
+++ b/client/ClientNetworking.cpp
@@ -222,6 +222,9 @@ public:
     /** Sets Host player ID. */
     void SetHostPlayerID(int host_player_id);
 
+    /** Sets compression usage for outgoing messages. */
+    void SetCompressionUse(bool use_cpr);
+
     /** Get authorization roles access. */
     Networking::AuthRoles& AuthorizationRoles();
 
@@ -256,6 +259,7 @@ private:
     boost::asio::high_resolution_timer m_reconnect_timer;
     tcp::resolver::iterator            m_resolver_results;
     bool                               m_deadline_has_expired{ false };
+    bool                               m_localhost_server;
 
     // m_mutex guards m_incoming_message, m_rx_connected and m_tx_connected which are written by
     // the networking thread and read by the main thread to check incoming messages and connection
@@ -274,6 +278,8 @@ private:
     Message::HeaderBuffer           m_outgoing_header = {};
 
     std::string                     m_destination;
+
+    bool                            m_use_compression;
 };
 
 
@@ -284,7 +290,9 @@ ClientNetworking::Impl::Impl() :
     m_socket(m_io_context),
     m_deadline_timer(m_io_context),
     m_reconnect_timer(m_io_context),
-    m_incoming_messages(m_mutex)
+    m_localhost_server(false),
+    m_incoming_messages(m_mutex),
+    m_use_compression(false)
 {}
 
 bool ClientNetworking::Impl::IsConnected() const {
@@ -421,8 +429,10 @@ bool ClientNetworking::Impl::ConnectToServer(
         ErrorLogger(network) << "ConnectToServer() : unable to connect to server at "
                              << ip_address << " due to exception: " << e.what();
     }
-    if (IsConnected())
+    if (IsConnected()) {
         m_destination = ip_address;
+        m_localhost_server = m_destination == "localhost" || m_destination == "127.0.0.1";
+    }
     TraceLogger(network) << "ClientNetworking::Impl::ConnectToServer() - Returning.";
     return IsConnected();
 }
@@ -468,6 +478,9 @@ void ClientNetworking::Impl::SetPlayerID(int player_id) {
 void ClientNetworking::Impl::SetHostPlayerID(int host_player_id)
 { m_host_player_id = host_player_id; }
 
+void ClientNetworking::Impl::SetCompressionUse(bool use_cpr)
+{ m_use_compression = use_cpr; }
+
 Networking::AuthRoles& ClientNetworking::Impl::AuthorizationRoles()
 { return m_roles; }
 
@@ -476,7 +489,12 @@ void ClientNetworking::Impl::SendMessage(Message&& message) {
         ErrorLogger(network) << "ClientNetworking::SendMessage can't send message when not transmit connected";
         return;
     }
-    TraceLogger(network) << "ClientNetworking::SendMessage() : sending message " << message;
+    if (!m_localhost_server && m_use_compression && message.Size() >= Message::COMPRESSION_THRESHOLD) {
+        TraceLogger(network) << "ClientNetworking::SendMessage() : sending compressed message " << message;
+        message.Compress();
+    } else {
+        TraceLogger(network) << "ClientNetworking::SendMessage() : sending message " << message;
+    }
     m_io_context.post(boost::bind(&ClientNetworking::Impl::SendMessageImpl, this, std::move(message)));
 }
 
@@ -487,8 +505,14 @@ void ClientNetworking::Impl::SendSelfMessage(Message&& message) {
 
 boost::optional<Message> ClientNetworking::Impl::GetMessage() {
     auto message = m_incoming_messages.PopFront();
-    if (message)
-        TraceLogger(network) << "ClientNetworking::GetMessage() : received message " << *message;
+    if (message) {
+        if ((*message).Compressed()) {
+            (*message).Decompress();
+            TraceLogger(network) << "ClientNetworking::GetMessage() : received compressed message " << *message;
+        } else{
+            TraceLogger(network) << "ClientNetworking::GetMessage() : received message " << *message;
+        }
+    }
     return message;
 }
 
@@ -818,6 +842,9 @@ void ClientNetworking::SetPlayerID(int player_id)
 
 void ClientNetworking::SetHostPlayerID(int host_player_id)
 { return m_impl->SetHostPlayerID(host_player_id); }
+
+void ClientNetworking::SetCompressionUse(bool use_cpr)
+{ m_impl->SetCompressionUse(use_cpr); }
 
 Networking::AuthRoles& ClientNetworking::AuthorizationRoles()
 { return m_impl->AuthorizationRoles(); }
--- a/client/ClientNetworking.cpp
+++ b/client/ClientNetworking.cpp
@@ -427,9 +427,8 @@ bool ClientNetworking::Impl::ConnectToServer(
         ErrorLogger(network) << "ConnectToServer() : unable to connect to server at "
                              << ip_address << " due to exception: " << e.what();
     }
-    if (IsConnected()) {
+    if (IsConnected())
         m_destination = ip_address;
-    }
     TraceLogger(network) << "ClientNetworking::Impl::ConnectToServer() - Returning.";
     return IsConnected();
 }
--- a/client/ClientNetworking.cpp
+++ b/client/ClientNetworking.cpp
@@ -431,7 +431,7 @@ bool ClientNetworking::Impl::ConnectToServer(
     }
     if (IsConnected()) {
         m_destination = ip_address;
-        m_localhost_server = m_destination == "localhost" || m_destination == "127.0.0.1";
+        m_localhost_server = m_socket.remote_endpoint().address().is_loopback();
     }
     TraceLogger(network) << "ClientNetworking::Impl::ConnectToServer() - Returning.";
     return IsConnected();
@@ -506,8 +506,8 @@ void ClientNetworking::Impl::SendSelfMessage(Message&& message) {
 boost::optional<Message> ClientNetworking::Impl::GetMessage() {
     auto message = m_incoming_messages.PopFront();
     if (message) {
-        if ((*message).Compressed()) {
-            (*message).Decompress();
+        if (message->Compressed()) {
+            message->Decompress();
             TraceLogger(network) << "ClientNetworking::GetMessage() : received compressed message " << *message;
         } else{
             TraceLogger(network) << "ClientNetworking::GetMessage() : received message " << *message;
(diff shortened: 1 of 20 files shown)


network/Message.cpp
```diff
@@ -85,15 +94,74 @@ char* Message::Data() noexcept
...
+        m_is_compressed = true;
+    }
+}
+
+std::string Message::CompressImpl(const std::string& data) {
+    boost::iostreams::basic_array_source<char> src(data.c_str(), data.size());
+    boost::iostreams::stream<boost::iostreams::basic_array_source<char>> s_src(src);
+
+    std::string compressed;
+    compressed.reserve(std::pow(2.0, 26.0));
```
geoffthemedio: shouldn't be necessary to reserve more than the size of the data being compressed

network/Message.cpp
```diff
@@ -85,15 +94,74 @@ char* Message::Data() noexcept
...
     m_message_size = 0;
+    m_is_compressed = false;
     m_message_text.clear();
 }
 
+void Message::Compress() noexcept {
+    if (!m_is_compressed) {
+        std::string_view::size_type temp = m_message_size;
+        m_message_text = CompressImpl(m_message_text);
+        m_message_size = m_message_text.size();
```
geoffthemedio: could perhaps leave this as the uncompressed size, so that it can be used to allocate space when decompressing

network/Message.cpp
```diff
@@ -85,15 +94,74 @@ char* Message::Data() noexcept
...
+        m_is_compressed = false;
+    }
+}
+
+std::string Message::DecompressImpl(const std::string& data) {
+    boost::iostreams::basic_array_source<char> src(data.c_str(), data.size());
+    boost::iostreams::stream<boost::iostreams::basic_array_source<char>> s_src(src);
+
+    std::string decompressed;
+    decompressed.reserve(std::pow(2.0, 26.0));
```
geoffthemedio: would be better to send the message size and allocate appropriate space. maybe with a safety check / max message size to avoid trying to allocate too much.

network/Message.cpp
```diff
@@ -743,15 +825,33 @@ void ExtractErrorMessageData(const Message& msg, int& player_id, std::string& pr
...
 
-void ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string) {
+void ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string,
+                                  bool& use_compression) {
     try {
         std::istringstream is(msg.Text());
         freeorion_xml_iarchive ia(is);
         ia >> BOOST_SERIALIZATION_NVP(host_player_name)
-           >> BOOST_SERIALIZATION_NVP(client_version_string);
+           >> BOOST_SERIALIZATION_NVP(client_version_string)
```
geoffthemedio: should separately handle the new field so that, hopefully, it not being there doesn't prevent the other data from being used.

network/Message.h
```diff
@@ -139,22 +142,30 @@ class FO_COMMON_API Message {
...
+    bool               Compressed() const noexcept; ///< Returns the compression flag.
+    const char*        Data() const noexcept;       ///< Returns the underlying buffer.
+    const std::string& Text() const;                ///< Returns the underlying buffer as a std::string.
 
     void               Resize(std::size_t size);   ///< Resizes the underlying char buffer to \a size uninitialized bytes.
     char*              Data() noexcept;            ///< Returns the underlying buffer.
     void               Swap(Message& rhs) noexcept;///< Swaps the contents of \a *this with \a rhs.  Does not throw.
     void               Reset() noexcept;           ///< Reverts message to same state as after default constructor
 
+    void               Compress() noexcept;   ///< Compresses data, updates size and compression flag, if data is uncompressed.
```
geoffthemedio: I am skeptical about `noexcept` given the large allocations

client/ClientNetworking.cpp
```diff
@@ -487,8 +505,14 @@ void ClientNetworking::Impl::SendSelfMessage(Message&& message) {
 
 boost::optional<Message> ClientNetworking::Impl::GetMessage() {
     auto message = m_incoming_messages.PopFront();
-    if (message)
-        TraceLogger(network) << "ClientNetworking::GetMessage() : received message " << *message;
+    if (message) {
+        if ((*message).Compressed()) {
```
geoffthemedio: can you do `message->Compressed()` ? (I'm not sure, but would be nicer...)

network/Message.cpp
mlam19: Would using a caught exception be acceptable as a way of handling this? Something like:
```
// read everything else

try {
    ia >> BOOST_SERIALIZATION_NVP(use_compression);
} catch (const std::exception& dne) {
    use_compression = false;
}

// continue
```
I don't know if there are "better" ways to discover whether an element exists inside the `xml_iarchive`. Using an XML parser seems like overkill, and a substring search seems unreliable.

geoffthemedio: I'm unsure if it will work, but I would guess that it should

mlam19: After some experimentation, I regret to report that it doesn't work.

```
try {
    std::istringstream is(msg.Text());
    freeorion_xml_iarchive ia(is);

    try {
        ia >> BOOST_SERIALIZATION_NVP(use_compression);
    } catch (const boost::archive::archive_exception& err) {
        use_compression = false;
    }
    // When ia goes out of scope, another archive_exception is thrown
} catch (const boost::archive::archive_exception& last_resort) {
    // Doesn't catch the exception
}
```
The `archive_exception` thrown when the tag is not found is caught and handled successfully. But then another `archive_exception` is thrown when `ia` goes out of scope, and that exception seems to be impossible to catch. 

It looks like an existing Boost issue: https://github.com/boostorg/serialization/issues/109. 

The only thing I can think of now is to create a new message type that includes the `use_compression` property. This would also be applied to the other messages that have been extended to communicate `use_compression`.

geoffthemedio: Based on that Boost issue, put the `freeorion_xml_iarchive ia(is);` within the inner `try` (and get rid of the outer `try`), so that the archive is destructed immediately when the exception is thrown, rather than persisting into the immediately following `catch`. Doing this lets me catch such an exception without immediately aborting.

client/ClientNetworking.cpp
```diff
@@ -421,8 +429,10 @@ bool ClientNetworking::Impl::ConnectToServer(
         ErrorLogger(network) << "ConnectToServer() : unable to connect to server at "
                              << ip_address << " due to exception: " << e.what();
     }
-    if (IsConnected())
+    if (IsConnected()) {
         m_destination = ip_address;
+        m_localhost_server = m_destination == "localhost" || m_destination == "127.0.0.1";
```
o01eg: I'm sure we can use `m_socket->remote_endpoint().address().is_loopback()`

network/Message.cpp
```diff
@@ -85,15 +94,74 @@ char* Message::Data() noexcept
...
+
+    std::string compressed;
+    compressed.reserve(std::pow(2.0, 26.0));
+
+    boost::iostreams::back_insert_device<std::string> cmp(compressed);
+    boost::iostreams::stream<boost::iostreams::back_insert_device<std::string>> s_cmp(cmp);
+
+    boost::iostreams::filtering_ostreambuf out;
+    out.push(boost::iostreams::zlib_compressor());
+    out.push(boost::iostreams::base64_encoder());
```
o01eg: Does base64 necessary?

network/Message.cpp
o01eg: Maybe we should allow to break compatibility between versions?

o01eg (changes requested):

network/Message.cpp
geoffthemedio: probably only if it will be wrapped in another layer of XML, which I guess doesn't apply in this case

network/Message.cpp
geoffthemedio: I'd rather not if it can be avoided

network/Message.cpp
o01eg: We have to send compressed size as well so network part will know where message ends.

network/Message.cpp
o01eg: We have to send compressed size as well so network part will know where message ends.

network/Message.cpp
mlam19: > Based on that Boost issue, put the `freeorion_xml_iarchive ia(is);` within the inner `try` (and get rid of the outer `try`), so that the archive is destructed immediately when the exception is thrown, rather than persisting into the immediately following `catch`. Doing this lets me catch such an exception without immediately aborting.

Collapsing the exception handling into a single `try` let's it run.
```
void ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string,
                                  bool& use_compression) {
    try {
        std::istringstream is(msg.Text());
        freeorion_xml_iarchive ia(is);
        ia >> BOOST_SERIALIZATION_NVP(host_player_name)
           >> BOOST_SERIALIZATION_NVP(client_version_string)
           >> BOOST_SERIALIZATION_NVP(use_compression);
    } catch (const boost::archive::archive_exception& err) {
        use_compression = false;
    } catch (const std::exception& err) {
        ErrorLogger() << "ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string, bool& use_compression) failed!  Message:\n"
                      << msg.Text() << "\n"
                      << "Error: " << err.what();
        throw err;
    }
}
```
But it also makes it impossible to tell whether it was `BOOST_SERIALIZATION_NVP(use_compression)` causing the `archive_exception`, which doesn't seem good.

It also makes `Message::ExtractJoinGameMessageData()` tricky since it also uses exceptions to detect the existence/validity of a `dependencies` element in the archive. Given the difficulties discovered above, the implementation of that function may already be broken.

I will try implementing a new message type to communicate compression use. Hopefully that side-steps all of this craziness.

geoffthemedio: I'd be slightly surprised if there was just one case where a try/catch block is used to handle message deserialization like this that don't actually work, since I/we wasn't aware of the archive destructor throwing issue and probably added a few intended-to-be-optional message fields over time.

o01eg: > I'd rather not if it can be avoided

Message header changed so it quite hard compatibility break already.

test/system/ClientAppFixture.h
```diff
@@ -24,14 +24,15 @@ class ClientAppFixture : public ClientApp {
...
-    MultiplayerLobbyData m_lobby_data;     ///< Lobby data.
+    bool                 m_game_started;    ///< Is server started the game?
+    std::set<int>        m_ai_empires;      ///< Ids of AI empires in game.
+    std::set<int>        m_ai_waiting;      ///< Ids of AI empires not yet send orders.
+    bool                 m_turn_done;       ///< Is server processed turn?
+    bool                 m_save_completed;  ///< Is server saved game?
+    boost::uuids::uuid   m_cookie;          ///< Cookie from server login.
+    bool                 m_lobby_updated;   ///< Did player get updated lobby.
+    MultiplayerLobbyData m_lobby_data;      ///< Lobby data.
+    bool                 m_use_compression; ///< Compression use request flag from the server.
```
geoffthemedio: What does this track? If the server has requested that this client send back compressed messages?

network/Message.h
```diff
@@ -51,13 +51,16 @@ namespace Moderator {
...
-    enum Parts : std::size_t {TYPE = 0, SIZE, Parts_Count};
+    enum Parts : std::size_t {TYPE = 0, SIZE, CPR_USED, Parts_Count};;
 
     typedef std::array<int, Parts::Parts_Count> HeaderBuffer;
 
     constexpr static std::size_t HeaderBufferSize =
         std::tuple_size<HeaderBuffer>::value* sizeof(HeaderBuffer::value_type);
 
+    /** Commpression is only performed on message data of at least this size in bytes. */
+    const static int COMPRESSION_THRESHOLD = 64;
```
geoffthemedio:
```suggestion
    static constexpr int COMPRESSION_THRESHOLD = 64;
```

network/Message.h
```diff
@@ -51,13 +51,16 @@ namespace Moderator {
   * misbehave as well.) */
 class FO_COMMON_API Message {
 public:
-    enum Parts : std::size_t {TYPE = 0, SIZE, Parts_Count};
+    enum Parts : std::size_t {TYPE = 0, SIZE, CPR_USED, Parts_Count};;
```
geoffthemedio:
```suggestion
    enum Parts : std::size_t {TYPE = 0, SIZE, Parts_Count, CPR_USED};
```
new thing at end to maintain consistency of old values underlying representation

network/Message.cpp
```diff
@@ -743,15 +825,33 @@ void ExtractErrorMessageData(const Message& msg, int& player_id, std::string& pr
...
+
+    } catch (const std::exception& err) {
+        ErrorLogger() << "ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string, bool& use_compression) failed!  Message:\n"
+                      << msg.Text() << "\n"
+                      << "Error: " << err.what();
+        throw err;
+    }
+}
+
+void ExtractHostMPAckMessage(const Message& msg, int& player_id, bool& req_compression)
```
geoffthemedio: don't abbreviate to "req" as it could be "request" or "require"

default/stringtables/en.txt
```diff
@@ -3411,6 +3414,9 @@ Create Binary Save Files
 OPTIONS_USE_XML_ZLIB_SERIALIZATION
 Use Compression for XML Save Files
 
+OPTIONS_USE_MSG_ZLIB_CPR
```
geoffthemedio, line 3417: This and the `_DB_` strings are possibly being used incorrectly here. The `_DB_` version is the long / full description shown in the usage text output listing of options or as a tooltip, while the non-DB version is shown as the option name:

option_name_and_description

Here you've apparently used one for the server option and one for the client...?

network/Message.h
mlam19: I do not know if there is a way to add additional fields to the header without putting its identifier before `Parts_Count`; `Parts_Count` effectively sets the size of the header.

o01eg: @geoffthemedio It won't work that way.

default/stringtables/en.txt
mlam19: Changed `OPTIONS_DB_USE_MSG_ZLIB_CPR` to be an appropriate tooltip for `OPTIONS_USE_MSG_ZLIB_CPR` in https://github.com/freeorion/freeorion/pull/3854/commits/4ead6795ce2b51bce839955a679750840ada4831.

network/Message.cpp
mlam19: Function removed in https://github.com/freeorion/freeorion/pull/3854/commits/e937091c4540e765549e9147757117dfbe3f4bcf; I had previously added it so the removal is also a revert.

network/Message.h
mlam19: Change made in https://github.com/freeorion/freeorion/pull/3854/commits/4ead6795ce2b51bce839955a679750840ada4831.

test/system/ClientAppFixture.h
mlam19: Variable removed in https://github.com/freeorion/freeorion/pull/3854/commits/e937091c4540e765549e9147757117dfbe3f4bcf.

network/Message.cpp
mlam19: Removed the base64 encoding and decoding in https://github.com/freeorion/freeorion/pull/3854/commits/4ead6795ce2b51bce839955a679750840ada4831.

network/Message.cpp
mlam19: https://github.com/freeorion/freeorion/commit/e937091c4540e765549e9147757117dfbe3f4bcf re-implements the transmission of compression usage requests to use a new message type `USE_COMPRESSION`. The changes to the serialization/deserialization of the existing messages are reverted.

client/ClientNetworking.cpp
mlam19: Change found at https://github.com/freeorion/freeorion/pull/3854/commits/e937091c4540e765549e9147757117dfbe3f4bcf#diff-4f06b6aa7be808a707f9ce9f096345a34f4cfe138aa1d1605ee9961d7611475dL481 after the removal of the indicated line (https://github.com/freeorion/freeorion/pull/3854/commits/e937091c4540e765549e9147757117dfbe3f4bcf#diff-4f06b6aa7be808a707f9ce9f096345a34f4cfe138aa1d1605ee9961d7611475dL434.)

client/ClientNetworking.cpp
mlam19: Change made in https://github.com/freeorion/freeorion/pull/3854/commits/4ead6795ce2b51bce839955a679750840ada4831.

network/Message.h
mlam19: Removed `noexcept` from `Compress()` and `Decompress()` in https://github.com/freeorion/freeorion/pull/3854/commits/4ead6795ce2b51bce839955a679750840ada4831.

network/Message.cpp
mlam19: Changed reserve size to uncompressed size in https://github.com/freeorion/freeorion/pull/3854/commits/4ead6795ce2b51bce839955a679750840ada4831. Also added handling for exception if the reserve size is still too large; in that case, compression is cancelled, error is logged, and message is sent uncompressed.

network/Message.cpp
mlam19: In https://github.com/freeorion/freeorion/pull/3854/commits/4ead6795ce2b51bce839955a679750840ada4831:
- Header expanded - again - to include the expected uncompressed size of the data.
- `m_message_size` is used to store the uncompressed size, and `m_message_cpr_size` is used to store the compressed size.
- `Compress()` updates `m_message_cpr_size`

network/Message.cpp
mlam19: Changed reserve size to uncompressed size - as found in the received header - in https://github.com/freeorion/freeorion/pull/3854/commits/4ead6795ce2b51bce839955a679750840ada4831. If the reserve size is too large, an error is logged and an exception is thrown; I don't think there's much that can be done in this case.

o01eg:
Is it possible to not allocate memory for uncompressed data and instead provide decompressing stream to boost serialization?

geoffthemedio:
> Is it possible to not allocate memory for uncompressed data and instead provide decompressing stream to boost serialization?

I'm not able to construct a `boost::archive::xml_iarchive` or a `boost::iostreams::stream` directly from a `boost::iostreams::filtering_istreambuf`. Maybe there's some other class that can bridge between them, though...

Edit: Apparently one can make a `std::istream` though... This works in [`LoadGame`](https://github.com/freeorion/freeorion/blob/master/server/SaveLoad.cpp#L434):
````
                timer.EnterSection("decompression");
                // allocate buffers for compressed and deceompressed serialized gamestate
                DebugLogger() << "Allocating buffers for XML deserialization...";
                std::string compressed_str;
                try {
                    if (ignored_save_preview_data.compressed_text_size > 0) {
                        DebugLogger() << "Based on header info for compressed state string, attempting to reserve: " << ignored_save_preview_data.compressed_text_size << " bytes";
                        compressed_str.reserve(ignored_save_preview_data.compressed_text_size);
                    } else {
                        compressed_str.reserve(std::pow(2.0, 26.0));
                    }
                } catch (...) {
                    DebugLogger() << "Unable to preallocate full deserialization buffers. Attempting deserialization with dynamic buffer allocation.";
                }

                // extract compressed gamestate info
                xia >> BOOST_SERIALIZATION_NVP(compressed_str);

                // wrap compressed string in iostream::stream to extract compressed data
                typedef boost::iostreams::basic_array_source<char> SourceDevice;
                SourceDevice compressed_source(compressed_str.data(), compressed_str.size());
                boost::iostreams::stream<SourceDevice> c_source(compressed_source);

                // set up filter to decompress data
                boost::iostreams::filtering_istreambuf i;
                i.push(boost::iostreams::zlib_decompressor());
                if (ignored_save_preview_data.save_format_marker == XML_COMPRESSED_BASE64_MARKER)
                    i.push(boost::iostreams::base64_decoder());
                i.push(c_source);

                std::istream in(&i);

                // create archive with (preallocated) buffer...
                freeorion_xml_iarchive xia2(in);
````

o01eg:
I've posted alternative approach #3870

Vezzra:
@mlam19, closed as/why? Superseded by another PR, or are you giving up on it...?

*Edit*: Never mind, just saw your other PR which replaces this one (meaning, this one here is closed as superseded).

Vezzra:
Superseded by #3904

PR #3853, opened 2022-05-27
Python parser refactor and unused tokens clean up

o01eg:
Fully cleans not implemented tokens.

Follows up #3860

PR #3860, opened 2022-05-30
Refactor python parser for single-parameter math functions

o01eg:
Follows up #3862


geoffthemedio:
why?

o01eg:
@geoffthemedio Those tokens not used in existing FOCS files so I removes them in a hope to reduce needed compilation resources. Also I moves their implementation to Python parser so it won't be lost they will be needed somehow.

Vezzra:
@o01eg, superseded by which PR?

o01eg:
@Vezzra By #3853

PR #3868, opened 2022-06-01
Windows binaries archive in CI

o01eg:
It was asked in recent voice chat for an archive with binaries for AI team.

It's a partial approach of #3421 to provide artifact with game binaries for Win32. Maybe could be improved further to provide game installer later.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -78,6 +78,19 @@ jobs:
         run: |
           $env:Path += ";$($pwd.Path)\\build\\Release"
           .\build\Release\Godot_v3.2.3-stable_win32.exe --no-window --disable-render-loop --video-driver GLES2 --verbose -s --path godot addons/gut/gut_cmdln.gd -gdir=res://test/ -ginclude_subdirs -gexit
+      - name: Generate binaries only
+        run: |
+          pushd build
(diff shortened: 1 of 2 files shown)


o01eg:
That how artifacts looks like https://github.com/o01eg/freeorion/actions/runs/2423028484

o01eg:
How binaries looks https://github.com/o01eg/freeorion/actions/runs/2422118304

Vezzra (approved):
Can't really provide much useful input on the changes in the script files, but they seem to work. The resulting archive contains all the binaries my little Python script stores in the "Binaries Only" archive, and much more (all the Boost DLLs etc.).

Cjkjvfnby:
I am thinking about extracting this to a separate workflow with the manual run.

So you build binaries when you need them. 

I'll experiment in my repo since this workflow must be in master to run it manaully.
https://github.com/Cjkjvfnby/freeorion/blob/master/.github/workflows/windows-artefacts.yml

UPD. never mind, it took additional 10 seconds and generate 24 MB of files.  We could leave it here.

o01eg:
@Cjkjvfnby I suppose we can add `workflow_dispatch` to `on` block and check it in `if` parameter of appropriate step.

geoffthemedio:
> manual run.

why not for all master branch CI runs?

Cjkjvfnby:
It looks 10 seconds to build them, so building on master branches sounds reasonable.

Cjkjvfnby:
@o01eg , maybe we could merge this branch now?

PR #3869, opened 2022-06-01
Fix dummy parsers

o01eg:


geoffthemedio:
same as https://github.com/freeorion/freeorion/commit/2d3ed2160d94bfdfd03c560cd860829cd789cbbd I think

PR #3874, opened 2022-06-03
Add boost version to freeorion version string

o01eg:
Solves potential issue where binary serialization chosen between different boost versions.


geoffthemedio:
This seems like it should be sent separately from the FreeOrion version string, rather than as part of it.

geoffthemedio:
to be clear, you can / should say so when you disagree with me on such issues...

o01eg:
It's fine, I'll just use versions of dependencies which clients already sent. It looks more clean compared to all those changes to build scripts.

PR #3876, opened 2022-06-03
Skip codeql job if not files changed

Cjkjvfnby:


.github/workflows/codeql-analysis.yml
```diff
@@ -6,7 +6,16 @@ on:
...
-
+    paths-ignore:
+      - 'msvc*/**'
+      - 'check/**'
+      - 'doc/**'
+      - 'packaging/**'
+      - 'snap/**'
+      - '*.md'
+      - 'check/**'
+      - 'default/**'
```
geoffthemedio, line 17: CodeQL seems to only apply to C++ files, so I guess this is OK despite eg. https://github.com/freeorion/freeorion/tree/master/default/python

Although I suppose it could be set up to also check python files?

Cjkjvfnby: Let's keep it simple.  It's not checking python files now. 

Even if we decide to add it to Python, I'd prefer to keep it as a separate task with its own set of paths.

PR #3879, opened 2022-06-04
Split getNamedValue to getNamedReal/Int

Grummel7:
Also allow queuing unbuildable buildings and regenerated the python interfaces.

Using functions that return int / float helps with the python code analysis. Also the current function always logs a warning, when accessing a NameInteger.

### DIFF
--- a/client/AI/AIWrapper.cpp
+++ b/client/AI/AIWrapper.cpp
@@ -394,11 +394,6 @@ namespace {
             return 0;
         }
 
-        if (!empire->ProducibleItem(BuildType::BT_BUILDING, item_name, location_id)) {
-            ErrorLogger() << "IssueEnqueueBuildingProductionOrder : specified item_name and location_id that don't indicate an item that can be built at that location";
-            return 0;
-        }
-
         if (!empire->EnqueuableItem(BuildType::BT_BUILDING, item_name, location_id)) {
             ErrorLogger() << "IssueEnqueueBuildingProductionOrder : specified item_name and location_id that don't indicate an item that can be enqueued at that location";
             return 0;
--- a/default/python/AI/freeOrionAIInterface.pyi
+++ b/default/python/AI/freeOrionAIInterface.pyi
@@ -1301,9 +1301,14 @@ def getGameRules() -> GameRules:
     Returns the game rules manager, which can be used to look up the names (string) of rules are defined with what type (boolean / toggle, int, double, string), and what values the rules have in the current game.
     """
 
-def getNamedValue(string: str) -> object:
+def getNamedInt(string: str) -> int:
     """
-    Returns the named value of the scripted constant with name (string). If no such named constant exists, returns none.
+    Returns the named integer value of the scripted constant with name (string). If no such named constant exists, returns 0.
+    """
+
+def getNamedReal(string: str) -> float:
+    """
+    Returns the named real value of the scripted constant with name (string). If no such named constant exists, returns 0.0.
     """
 
 def getOptionsDBOptionBool(string: str) -> object:
--- a/default/python/AI/freeorion_tools/_freeorion_tools.py
+++ b/default/python/AI/freeorion_tools/_freeorion_tools.py
@@ -371,14 +371,10 @@ def get_named_int(name: str) -> int:
     Note that we do not raise and exception so that the AI can continue, as good as it can, with outdated information.
     This is also why we return 1, returning 0 could cause followup errors if the value is used as divisor.
     """
-    value = fo.getNamedValue(name)
-    if value is None:
-        error(f"Requested NamedInt {name}, which doesn't exist!")
-        value = 1
-    elif not isinstance(value, int):
-        error(f"Requested value {name} of type int got {type(value)}!")
-        value = 1
-    return value
+    if fo.namedIntDefined(name):
+        return fo.getNamedInt(name)
+    error(f"Requested integer {name} does not exist!")
+    return 1
 
 
 def get_named_real(name: str) -> float:
@@ -388,11 +384,7 @@ def get_named_real(name: str) -> float:
     Note that we do not raise and exception so that the AI can continue, as good as it can, with outdated information.
     This is also why we return 1, returning 0 could cause followup errors if the value is used as divisor.
     """
-    value = fo.getNamedValue(name)
-    if value is None:
-        error(f"Requested NamedReal {name}, which doesn't exist!")
-        value = 1.0
-    elif not isinstance(value, float):
-        error(f"Requested value {name} of type float got {type(value)}!")
-        value = 1.0
-    return value
+    if fo.namedRealDefined(name):
+        return fo.getNamedReal(name)
+    error(f"Requested integer {name} does not exist!")
+    return 1.0
--- a/python/UniverseWrapper.cpp
+++ b/python/UniverseWrapper.cpp
@@ -321,24 +321,30 @@ namespace FreeOrionPython {
                 +[](const std::string& name) -> bool { return GetValueRef<int>(name, true); },
                 "Returns true/false (boolean) whether there is a defined int-valued "
                 "scripted constant with name (string).");
-        py::def("getNamedValue",
-                +[](const std::string& name) -> py::object {
-                    auto eval = [](auto&& ref) -> py::object {
+        py::def("getNamedReal",
+                +[](const std::string& name) -> double {
+                    if (const auto ref = GetValueRef<double>(name, true)) {
                         if (ref->ConstantExpr()) 
-                            return py::object(ref->Eval());
+                            return ref->Eval();
                         const ScriptingContext context;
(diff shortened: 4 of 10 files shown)


client/AI/AIWrapper.cpp
```diff
@@ -394,11 +394,6 @@ namespace {
             return 0;
         }
 
-        if (!empire->ProducibleItem(BuildType::BT_BUILDING, item_name, location_id)) {
-            ErrorLogger() << "IssueEnqueueBuildingProductionOrder : specified item_name and location_id that don't indicate an item that can be built at that location";
-            return 0;
-        }
-
```
Grummel7, line 401: The human client allows it, I see no reason, why the AI should not be allowed to do it.

geoffthemedio: Also consider whether the order will execute on the server... If not, is being able to issue an broken order useful?

Grummel7: I got an error from when the AI issued an orbital dry dock right after the basic ship yard. As far as I know all cases where a building is enqueueable, but not producible are cases where a prerequisite is also in the queue.

default/python/AI/freeOrionAIInterface.pyi
```diff
@@ -1973,9 +1981,18 @@ def getTech(string: str) -> tech:
...
+
+
+@overload
+def getTechCategories()  -> object: ...
+
+
+def getTechCategories(*args) -> object:
     """
     Returns the names of all tech categories (StringVec).
+    Returns the names of all tech categories (StringVec).
```
Cjkjvfnby: It would be nice o to add information about arguments to C++ documentation.

Cjkjvfnby (approved):

Cjkjvfnby:
LGFM, @geoffthemedio , could you take a look?

python/UniverseWrapper.cpp
```diff
@@ -311,23 +311,29 @@ namespace FreeOrionPython {
...
-                        return eval(ref);
-                    else
-                        return py::object();
+                        return ref->Eval(context);
+                    }
+                    return 0.0;
                 },
-                "Returns the named value of the scripted constant with name (string). "
+                "Returns the named real value of the scripted constant with name (string). "
+                "If no such named constant exists, returns none.");
```
geoffthemedio: it appears to `return 0.0;` in the event there is no such named real valueref...

python/UniverseWrapper.cpp
```diff
@@ -311,23 +311,29 @@ namespace FreeOrionPython {
...
+        py::def("getNamedInt",
+                +[](const std::string& name) -> int {
+                    if (const auto ref = GetValueRef<int>(name, true)) {
+                        if (ref->ConstantExpr())
+                            return ref->Eval();
+                        const ScriptingContext context;
+                        return ref->Eval(context);
+                    }
+                    return 0;
```
geoffthemedio, line 344: same here...

python/UniverseWrapper.cpp
```diff
@@ -311,23 +311,29 @@ namespace FreeOrionPython {
...
-
-                    if (const auto ref = GetValueRef<double>(name, true))
-                        return eval(ref);
-                    else if (const auto ref = GetValueRef<int>(name, true))
-                        return eval(ref);
-                    else
-                        return py::object();
+                        return ref->Eval(context);
+                    }
+                    return 0.0;
```
geoffthemedio, line 332:
```suggestion
                    return py::object();
```

python/UniverseWrapper.cpp
geoffthemedio, line 344:
```suggestion
                    return py::object();
```

python/UniverseWrapper.cpp
Grummel7: Actually that does not work:
`error: cannot convert ‘boost::python::api::object’ to ‘double’ in return`

It almost works with:
```
        py::def("getNamedReal",
                +[](const std::string& name) -> boost::optional<double> {
                    if (const auto ref = GetValueRef<double>(name, true)) {
                        if (ref->ConstantExpr()) 
                            return ref->Eval();
                        const ScriptingContext context;
                        return ref->Eval(context);
                    }
                    return boost::optional<double>(boost::none);
                },
```

It compiles and generates the interface, but when I run it, I get an exception:
`No to_python (by-value) converter found for C++ type: boost::optional<double>`

geoffthemedio: I guess the lambda would need to have the signature `[](const std::string& name) -> py::object` as before then? Don't think you need `boost::optional` to get the equivalent functionality on the Python side of the API.

Grummel7: Well, using return type py::object would surely work, but defeats the purpose. Okay, I'll get rid of the warning, but PyCharm would still only see two functions that returns an object and complain when I use it like a number.

Well, I guess I could use a wrapper in python, that casts the value to the correct type...

geoffthemedio: an alternative is to have separate `IsNamedReal` function to query before using, and to return 0 if there isn't one, as you've tried to do, and then update the documentation to not claim it returns `none` if there isn't.

Grummel7: Well, doing this would also require wrapper functions.
I've now done it with return value py::object and wrapper functions in freeorion_tools.

default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -349,3 +349,31 @@ def get_ship_part(part_name: str):
...
+        value = 1
+    return value
+
+
+def named_real(name: str) -> float:
+    """
+    Returns a NamedReal from FOCS.
+    If the value does not exist, reports an error and returns 1.0.
+    Note that we do not raise and exception so that the AI can continue, as good as it can, with outdated information.
+    This is also why we return 1, returning 0 could cause followup errors if the value is used as divisor.
```
geoffthemedio, line 385: arguably better to add safety checks on the divisor wherever that error could happen...?

Grummel7: The only case where I have done it so far was dividing by BUREAUCRACY_ADMINS_NEEDED_TURN_SCALING, which hopefully is gone soon anyway. This value was also used as a divisor in FOCS itself.

In the end, when a value that is used by the AI is removed from FOCS, the AI should be adopted. But in doubt using a wrong value should still be better than throwing an exception, which means skipping all the later code for that turn, although most of it may still work correctly.

geoffthemedio: the point wouldn't be to throw an exception, but rather to check that the value that is returned from the API call won't cause an exception to be thrown, which could occur if the referenced named value is removed and as a result the returned value is different from what the AI code expected

default/python/AI/freeOrionAIInterface.pyi
geoffthemedio: there are none
https://github.com/freeorion/freeorion/blob/master/universe/Tech.h#L195
https://github.com/freeorion/freeorion/blob/master/python/EmpireWrapper.cpp#L470

default/python/AI/freeorion_tools/_freeorion_tools.py
Grummel7: That point is beyond the scope of this PR. I've chosen a default that is less likely to cause problems, while the fact that the default is used is reported by an error. There is currently not a single case where a named value is used as a divisor.

python/UniverseWrapper.cpp
Grummel7: > an alternative is to have separate `IsNamedReal` function to query before using, and to return 0 if there isn't one, as you've tried to do, and then update the documentation to not claim it returns `none` if there isn't.

Since I started using wrappers anyway, I've gone that way.

client/AI/AIWrapper.cpp
geoffthemedio: I suppose this should be a separate commit, if not a separate pull request, as it's seemingly unrelated to refactoring the getNamed* interface

geoffthemedio:
should probably combine into one commit with a better message, like:
````
* split getNamedValue to getNamedReal/Int in the Python API, which return 0 when no such named reference exists rather than a Python none object
* in Python get_named_int and get_named_real wrappers, check that the requested reference exists before evaluating, and return 1 if it doesn't
````

Grummel7:
> should probably combine into one commit with a better message, like:
> 
> ```
> * split getNamedValue to getNamedReal/Int in the Python API, which return 0 when no such named reference exists rather than a Python none object
> * in Python get_named_int and get_named_real wrappers, check that the requested reference exists before evaluating, and return 1 if it doesn't
> ```

Sigh, if this is what it takes to finally get this merged...
Of course you are right that outside the PR my first commit comment would have sounded rather pointless.

PR #3887, opened 2022-06-07
Add dependencies data to HOST_* messages

o01eg:
Make it same as JOIN_GAME message. Will allow to use those data when establish player connection.

### DIFF
--- a/client/human/GGHumanClientApp.cpp
+++ b/client/human/GGHumanClientApp.cpp
@@ -619,7 +619,7 @@ void GGHumanClientApp::NewSinglePlayerGame(bool quickstart) {
 
 
     TraceLogger() << "Sending host SP setup message";
-    m_networking->SendMessage(HostSPGameMessage(setup_data));
+    m_networking->SendMessage(HostSPGameMessage(setup_data, DependencyVersions()));
     m_fsm->process_event(HostSPGameRequested());
     TraceLogger() << "GGHumanClientApp::NewSinglePlayerGame done";
 }
@@ -668,7 +668,7 @@ void GGHumanClientApp::MultiPlayerGame() {
     }
 
     if (server_connect_wnd->GetResult().server_dest == "HOST GAME SELECTED") {
-        m_networking->SendMessage(HostMPGameMessage(server_connect_wnd->GetResult().player_name));
+        m_networking->SendMessage(HostMPGameMessage(server_connect_wnd->GetResult().player_name, DependencyVersions()));
         m_fsm->process_event(HostMPGameRequested());
     } else {
         boost::uuids::uuid cookie = boost::uuids::nil_uuid();
@@ -804,7 +804,7 @@ void GGHumanClientApp::LoadSinglePlayerGame(std::string filename/* = ""*/) {
     // leving setup_data.m_players empty : not specified when loading a game, as server will generate from save file
 
 
-    m_networking->SendMessage(HostSPGameMessage(setup_data));
+    m_networking->SendMessage(HostSPGameMessage(setup_data, DependencyVersions()));
     m_fsm->process_event(HostSPGameRequested());
 }
 
--- a/network/Message.cpp
+++ b/network/Message.cpp
@@ -130,25 +130,27 @@ Message ErrorMessage(const std::string& problem, bool fatal, int player_id) {
     return Message{Message::MessageType::ERROR_MSG, std::move(os).str()};
 }
 
-Message HostSPGameMessage(const SinglePlayerSetupData& setup_data) {
+Message HostSPGameMessage(const SinglePlayerSetupData& setup_data, const std::map<std::string, std::string>& dependencies) {
     std::ostringstream os;
     {
         std::string client_version_string = FreeOrionVersionString();
         freeorion_xml_oarchive oa(os);
         oa << BOOST_SERIALIZATION_NVP(setup_data)
-           << BOOST_SERIALIZATION_NVP(client_version_string);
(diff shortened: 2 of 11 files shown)


client/human/GGHumanClientApp.cpp
```diff
@@ -799,7 +799,7 @@ void GGHumanClientApp::LoadSinglePlayerGame(std::string filename/* = ""*/) {
     // leving setup_data.m_players empty : not specified when loading a game, as server will generate from save file
 
 
-    m_networking->SendMessage(HostSPGameMessage(setup_data));
+    m_networking->SendMessage(HostSPGameMessage(setup_data, DependencyVersions()));
```
geoffthemedio, line 807: I guess this is only relevant if launching a stand alone server locally that was built against different dependencies than the client? Normally they should be consistent if built together and the client launches the server it connects to for an SP game.

network/Message.cpp
```diff
@@ -743,12 +745,18 @@ void ExtractErrorMessageData(const Message& msg, int& player_id, std::string& pr
...
 
-void ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string) {
+void ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string, std::map<std::string, std::string>& dependencies) {
     try {
         std::istringstream is(msg.Text());
         freeorion_xml_iarchive ia(is);
         ia >> BOOST_SERIALIZATION_NVP(host_player_name)
            >> BOOST_SERIALIZATION_NVP(client_version_string);
+        dependencies.clear();
+        try {
```
geoffthemedio: this extra `try catch` won't work, based on recent discoveries. If there is some error when deserializing, such as due to lack of text for `dependencies` in the archive, the exception gets rethrown in the archive destructor, which causes an immediate abort.

You can just use the outer `try` that has the archive in its scope, so that the archive has already been destructed by the time the `catch` block is reached. You can then check if some of the deserialized-into containers are empty if you want to output to the log...

network/Message.cpp
```diff
@@ -743,12 +745,18 @@ void ExtractErrorMessageData(const Message& msg, int& player_id, std::string& pr
     }
 }
 
-void ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string) {
+void ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string, std::map<std::string, std::string>& dependencies) {
```
geoffthemedio, line 748:
```suggestion
void ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string, std::map<std::string, std::string>& dependencies) {
    dependencies.clear();
```

network/Message.cpp
```diff
@@ -743,12 +745,18 @@ void ExtractErrorMessageData(const Message& msg, int& player_id, std::string& pr
...
         std::istringstream is(msg.Text());
         freeorion_xml_iarchive ia(is);
         ia >> BOOST_SERIALIZATION_NVP(host_player_name)
            >> BOOST_SERIALIZATION_NVP(client_version_string);
+        dependencies.clear();
+        try {
+            ia >> BOOST_SERIALIZATION_NVP(dependencies);
+        } catch (...) {
+            DebugLogger() << "ExtractHostMPGameMessageData(...) didn't get a list of dependencies.";
+        }
```
geoffthemedio:
```suggestion
            ia >> BOOST_SERIALIZATION_NVP(dependencies);
```

network/Message.cpp
```diff
@@ -743,12 +745,18 @@ void ExtractErrorMessageData(const Message& msg, int& player_id, std::string& pr
...
 
     } catch (const std::exception& err) {
         ErrorLogger() << "ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string) failed!  Message:\n"
```
geoffthemedio, line 757:
```suggestion
        if (dependencies.empty())
            DebugLogger() << "ExtractHostMPGameMessageData(...) didn't get a list of dependencies.";
        ErrorLogger() << "ExtractHostMPGameMessageData(const Message& msg, std::string& host_player_name, std::string& client_version_string) failed!  Message:\n"
```

network/Message.cpp
```diff
@@ -1162,13 +1170,18 @@ void ExtractPlayerStatusMessageData(const Message& msg, Message::PlayerStatus& s
...
-void ExtractHostSPGameMessageData(const Message& msg, SinglePlayerSetupData& setup_data, std::string& client_version_string) {
+void ExtractHostSPGameMessageData(const Message& msg, SinglePlayerSetupData& setup_data, std::string& client_version_string, std::map<std::string, std::string>& dependencies) {
     try {
         std::istringstream is(msg.Text());
         freeorion_xml_iarchive ia(is);
         ia >> BOOST_SERIALIZATION_NVP(setup_data)
            >> BOOST_SERIALIZATION_NVP(client_version_string);
-
+        dependencies.clear();
+        try {
```
geoffthemedio: same as above

client/human/GGHumanClientApp.cpp
o01eg: It possible to use `network.server.external.force` and connect somewhere else.

network/Message.cpp
o01eg: Removed unneeded `try catch`.

network/Message.cpp
o01eg: Moved `clear` up in all functions.

PR #3890, opened 2022-06-09
Rename Min, Max operations to MinOf, MaxOf, and If effect to Conditional in python parser

o01eg:
This change allows to not conflict with statistic type enums.

### DIFF
--- a/default/scripting/techs/construction/FORCE_ENERGY_STRC.focs.py
+++ b/default/scripting/techs/construction/FORCE_ENERGY_STRC.focs.py
@@ -19,14 +19,14 @@ Tech(
                 If(
                     condition=(Value(LocalCandidate.Industry) <= Value(LocalCandidate.TargetIndustry)),
                     effects=SetIndustry(
-                        value=Min(
+                        value=MinOf(
                             float,
                             Value + NamedReal(name="FORCE_ENERGY_RATE_INCREASE", value=3.0),
                             Value(Target.TargetIndustry),
                         )
                     ),
                     else_=SetIndustry(
-                        value=Max(
+                        value=MaxOf(
                             float,
                             Value - NamedRealLookup(name="FORCE_ENERGY_RATE_INCREASE"),
                             Value(Target.TargetIndustry),
@@ -36,14 +36,14 @@ Tech(
                 If(
                     condition=(Value(LocalCandidate.Research) <= Value(LocalCandidate.TargetResearch)),
                     effects=SetResearch(
-                        value=Min(
+                        value=MinOf(
                             float,
                             Value + NamedRealLookup(name="FORCE_ENERGY_RATE_INCREASE"),
                             Value(Target.TargetResearch),
                         )
                     ),
                     else_=SetResearch(
-                        value=Max(
+                        value=MaxOf(
                             float,
                             Value - NamedRealLookup(name="FORCE_ENERGY_RATE_INCREASE"),
                             Value(Target.TargetResearch),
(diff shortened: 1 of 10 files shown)


parse/ValueRefPythonParser.cpp
```diff
@@ -350,10 +349,17 @@ void RegisterGlobalsValueRefs(boost::python::dict& globals, const PythonParser&
...
-    std::function<boost::python::object(const boost::python::tuple&, const boost::python::dict&)> f_insert_min = [&parser](const boost::python::tuple& args, const boost::python::dict& kw) { return insert_minmaxoneof_<ValueRef::OpType::MINIMUM>(parser, args, kw); };
-    globals["Min"] = boost::python::raw_function(f_insert_min, 3);
-    std::function<boost::python::object(const boost::python::tuple&, const boost::python::dict&)> f_insert_max = [&parser](const boost::python::tuple& args, const boost::python::dict& kw) { return insert_minmaxoneof_<ValueRef::OpType::MAXIMUM>(parser, args, kw); };
-    globals["Max"] = boost::python::raw_function(f_insert_max, 3);
+
+    // selection_operator
+    for (const auto& op : std::initializer_list<std::pair<const char*, ValueRef::OpType>>{
+            {"OneOf",   ValueRef::OpType::RANDOM_PICK},
+            {"MinOf",   ValueRef::OpType::MINIMUM},
+            {"MaxOf",   ValueRef::OpType::MAXIMUM}})
```
geoffthemedio, line 357: not sure if any better / different, but could also do:
```suggestion
    for (const auto& op : std::array<std::pair<const char*, ValueRef::OpType>, 3> {{
            {"OneOf",   ValueRef::OpType::RANDOM_PICK},
            {"MinOf",   ValueRef::OpType::MINIMUM},
            {"MaxOf",   ValueRef::OpType::MAXIMUM}}})
```

default/scripting/techs/construction/FORCE_ENERGY_STRC.focs.py
```diff
@@ -16,68 +16,68 @@
             accountinglabel="CON_TECH_ACCOUNTING_LABEL",
             priority=AFTER_ALL_TARGET_MAX_METERS_PRIORITY,
             effects=[
-                If(
+                EffectIf(
```
geoffthemedio: this seems awkward

Oberlus: `ConditionIf` ?

geoffthemedio: internally it's called `Conditional` ... Can't think of anything better than that.

default/scripting/techs/construction/FORCE_ENERGY_STRC.focs.py
```diff
@@ -16,68 +16,68 @@
...
                     condition=(Value(LocalCandidate.Industry) <= Value(LocalCandidate.TargetIndustry)),
```
geoffthemedio, line 20: could change the `condition=` to `if=` for consistency with the `else=` below perhaps?

default/scripting/techs/construction/FORCE_ENERGY_STRC.focs.py
```diff
@@ -16,68 +16,68 @@
...
                     effects=SetIndustry(
```
geoffthemedio, line 21: perhaps make `effects=` be `matching_effects=` and `else=` be `non_matching_effects=` ?

o01eg: I'd like to change original FOCS tokens and structure as little as possible.

parse/ValueRefPythonParser.cpp
o01eg: I use same loop in other places so not sure if forcing to define size is better.

default/scripting/techs/construction/FORCE_ENERGY_STRC.focs.py
o01eg: Changed to it.

Cjkjvfnby:
> This change allows to not conflict with statistic type enums.

Should we just rename statistic type enums?  I guess conditions will e used more often so they are deserve shorter names.

o01eg:
Not sure what if better, statistic type enums are expected to use with token `Statistic` which already long.

PR #3895, opened 2022-06-11
Update log dumper

Cjkjvfnby:
@Grummel7 , this is quite a massive refactoring of the log dumper. Not sure that it is worth reviewing it since it is helper code.  Just keep in mind that this tool is not compatible with the old logs.

Major features
- drop custom deserialization for specific kyes, and make it more general
- fix colors for empire plots
- move the player name to the sidebar


Grummel7:
Well, I tried it, the colors are now correct, and the side bar works as well. But I just see you already merged it anyway...

PR #3896, opened 2022-06-12
[AI] Rename log dumper to statistic

Cjkjvfnby:
Found a better name for log dumper


geoffthemedio:
maybe slightly confusingly similar to `ValueRef::Statistic` but I guess that doesn't really exist in Python land...

Cjkjvfnby:
> maybe slightly confusingly similar to ValueRef::Statistic but I guess that doesn't really exist in Python land...

It exists but is not used.

PR #3898, opened 2022-06-12
Do not run workflow if no C++ files changed

Cjkjvfnby:


.github/workflows/snapcraft.yml
```diff
@@ -7,14 +7,22 @@ on:
       - 'check/**'
       - 'doc/**'
       - 'packaging/**'
+      - 'snap/**'
```
o01eg: I'm not sure but maybe snap CI should build when snap files changed.

PR #3906, opened 2022-06-15
Speed up CMake CI by disabling changes checks

o01eg:


o01eg:
Looks like it doesn't help

PR #3908, opened 2022-06-16
Cleanup buildings

Cjkjvfnby:
I am playing with the idea to create dedicated enums for each building type. 

I have moved shipyard enums to a separate Enum, IMHO this looks better in the code since names became shorter without losing any information.

Enum could not be inherited, so I move base functions to a separate class.

@Grummel7 what do you think?

PS. These changes are related to the code, so they should be merged only after review.  There is no time pressure on that PR.

### DIFF
--- a/default/python/AI/ProductionAI.py
+++ b/default/python/AI/ProductionAI.py
@@ -23,7 +23,7 @@ import PlanetUtilsAI
 import PriorityAI
 from AIDependencies import INVALID_ID, Tags
 from aistate_interface import get_aistate
-from buildings import BuildingType, get_empire_drydocks
+from buildings import BuildingType, Shipyard, get_empire_drydocks
 from character.character_module import Aggression
 from colonization import rate_planetary_piloting
 from colonization.rate_pilots import GREAT_PILOT_RATING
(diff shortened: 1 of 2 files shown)


Grummel7:
I guess the new enum should be called ShipyardType to fit in.

In my current development, I'll iterator over all building types.  Is there an easy way to iterator over two enums? Otherwise you should provide maybe a tuple with all types.

The PR would be an elegant way to get rid of AIDependencies.SHIP_FACILITIES.
Logically it's a bit weird, though. Shipyards are buildings, after all. The doc string of BuildingType should at least mention that shipyards have their own class and ShipyardType should have a doc-string as well.

Code wise I see no problem with this.

Cjkjvfnby:
> 
> In my current development, I'll iterator over all building types. Is there an easy way to iterator over two enums? 


```python
from enum import Enum
from itertools import chain


class Foo(Enum):
    A = 1
    B = 2

class Boo(Enum):
    C = 3
    D = 4


for x in chain(Foo, Boo):
    print(x)
```

Cjkjvfnby:
> I guess the new enum should be called ShipyardType to fit in.

We use BuildingType to avoid confusion with the building class in C++ interface.  Shipyard is not used so we could claim it. The same for other specific building types.

Cjkjvfnby:
Reorganize dos in a slightly different way.

Cjkjvfnby:
@Grummel7 , could you take a second look?

Grummel7 (changes requested):
Since it is already merged, you could as well fix this line to use a chaining of both types:
https://github.com/freeorion/freeorion/blob/d75dd157b0feb76ac9be38120907a250eb2ef117/default/python/AI/colonization/calculate_stability.py#L85

Apart from that, I still think it is okay.

Cjkjvfnby:
Opps, forget to push the last commuit.

PR #3911, opened 2022-06-17
Add comment to dump about use of StaticCast and StringCast

o01eg:
Displays used casts in dump.

Research cost of Singularity of Transcendence from:

```
researchcost = 500 * GalaxySize * (GameRule name = "RULE_SINGULARITY_COST_FACTOR") * [[TECH_COST_MULTIPLIER]]
```

became:

```
researchcost = 500.000000 * /* StaticCast<i,d> */ GalaxySize /* -StaticCast */ * GameRule name = "RULE_SINGULARITY_COST_FACTOR" * GameRule name = "RULE_TECH_COST_FACTOR"
```

### DIFF
--- a/universe/ValueRefs.h
+++ b/universe/ValueRefs.h
@@ -1525,7 +1525,7 @@ std::string StaticCast<FromType, ToType>::Description() const
 
 template <typename FromType, typename ToType>
 std::string StaticCast<FromType, ToType>::Dump(unsigned short ntabs) const
-{ return m_value_ref->Dump(ntabs); }
+{ return "(" + m_value_ref->Dump(ntabs) + ") // StaticCast<" + typeid(FromType).name() + "," + typeid(ToType).name() + ">\n" + DumpIndent(ntabs + 1); }
 
 template <typename FromType, typename ToType>
 void StaticCast<FromType, ToType>::SetTopLevelContent(const std::string& content_name)
@@ -1647,7 +1647,7 @@ std::string StringCast<FromType>::Description() const
 
 template <typename FromType>
 std::string StringCast<FromType>::Dump(unsigned short ntabs) const
-{ return m_value_ref->Dump(ntabs); }
+{ return "(" + m_value_ref->Dump(ntabs) + ") // StringCast<" + typeid(FromType).name() + ">\n" + DumpIndent(ntabs + 1); }
 
 template <typename FromType>
 void StringCast<FromType>::SetTopLevelContent(const std::string& content_name) {


geoffthemedio:
Inserting the `/*` and `*/` in the middle of math expressions that use those operators is confusing. Perhaps it can be formatted better with spaces and some extra punctuation?

What's the difference between `StaticCast<i,d>` and `-StaticCast` ?

Some of the `(` and `)` seem to have gone missing as well?

o01eg:
> Inserting the `/*` and `*/` in the middle of math expressions that use those operators is confusing. Perhaps it can be formatted better with spaces and some extra punctuation?

Maybe try to use // and new line?

> What's the difference between `StaticCast<i,d>` and `-StaticCast` ?

It shows where part cast expression starts and finishes.

> Some of the `(` and `)` seem to have gone missing as well?

Yes, looks like they not used actually.

o01eg:
> Inserting the `/*` and `*/` in the middle of math expressions that use those operators is confusing. Perhaps it can be formatted better with spaces and some extra punctuation?

I've just tried to use it in Adaptive Automation and it doesn't produce error:

```
researchcost = 100 /* test */ * [[TECH_COST_MULTIPLIER]]
```

geoffthemedio:
> it doesn't produce error:

I didn't expect an error; it's just harder to read.

> > Some of the `(` and `)` seem to have gone missing as well?
> 
> Yes, looks like they not used actually.

Not necessary to parse but might also clarify for the reader.

Especially if moving to Python FOCS, the main purpose of the Dump output is a human reader...

so maybe representing the casts as functions, something like `value = 500.0 * todouble(Target.Age - 10)` or `value = 500.0 * int_to_double(Target.Age - 10)` would be better...?

o01eg:
I'd like to produce something still compatible with old FOCS.

Maybe something like?
```
researchcost = 500.000000 * ( GalaxySize ) // StaticCast<i,d>
    * GameRule name = "RULE_SINGULARITY_COST_FACTOR" * GameRule name = "RULE_TECH_COST_FACTOR"
```

Converted expression placed in round brackets and description of case placed in comment after closing bracket, next part is placed on next line according to `ntabs` argument.

geoffthemedio:
at least that example is clearer

o01eg:
Produced dump looks like:

```
Tech
    name = "LRN_TRANSCEND"
    description = "LRN_TRANSCEND_DESC"
    short_description = "VICTORY_SHORT_DESC"
    category = "LEARNING_CATEGORY"
    researchcost = 500.000000 * (GalaxySize) // StaticCast<i,d>
         * GameRule name = "RULE_SINGULARITY_COST_FACTOR" * GameRule name = "RULE_TECH_COST_FACTOR"
    researchturns = 20
...
```

geoffthemedio:
looks OK.

could probably do something compile-time to get the typenames, but that's probably not necessary

PR #3914, opened 2022-06-18
Add policies to charting

Cjkjvfnby:
Show table with policies in charting.

Page became overpopulated, I'll think about a new layout for it.


default/python/AI/PolicyAI.py
```diff
@@ -659,6 +658,7 @@ def _adopt(self, name: str) -> None:
         if fo.issueAdoptPolicyOrder(name, category, slot):
             if name not in self._originally_adopted:
                 self._ip -= policy.adoptionCost()
+            stats.adopt_policy(name)
```
Grummel7, line 664: This produces a lot of noise since PolicyAI adopts and deadopts several policies temporarily.
I suggest to do the output at the end of generate_orders by comparing _adopted with _originally_adopted

Grummel7 (commented):
Just merged it and tries it. This could actually be useful.

default/python/AI/PolicyAI.py
Cjkjvfnby: This will make things a bit more complicated.  145 items for 164 turns does not look a lot.

PR #3916, opened 2022-06-19
Planet rating

Grummel7:
This task is far from being finished, and I'll have to rebase it when done, too.
Anyway, it is time for others, especially @Cjkjvfnby, to have a look at it.


default/python/AI/AIDependencies.py
```diff
@@ -49,6 +49,15 @@
 TECH_COST_MULTIPLIER = 2.0
 FOCUS_CHANGE_PENALTY = 1
 STABILITY_PER_LIKED_FOCUS = 2.0
+HOMEWORLD_BONUS = 5.0
+STABILITY_PER_LIKED_BUILDING_ON_PLANET = 4.0
```
Oberlus, line 55: Can values like these be queried from the scripting files where they are defined?
I'm thinking of ease of maintenance. Modders (like me) changing the named values in the scripting files might completely miss the corresponding AI files.

Cjkjvfnby: Let's bring validation of the constants used by AI to the next voice meeting.

Grummel7: Since April the AI can read NamedInt and NamedReal values from the scripts, but not macros or hard-coded values.

I fact I found that named values can also be variable, so in theory, even formulas can be encoded into named values. But then you would need a context to get the correct value and at least so far the AI cannot specify a context.

geoffthemedio: You'd need a way to specify a context if you want to evaluate a non-constant valueref using something other than a default empty context. The API already does the default empty context case when called for a non-constant value ref. That could already be useful / work for something that can be evaluated with a default empty context, which might be useful for a formula (a combination of math operations) that depends on just constant values, on the current game turn, or other general gamestate info that does not reference `Source` or `Target` objects or a `Value` reference. Hard-coded species, hull, part, special, and similar names could be used, and probably statistics, like how many ships are there in the universe (as known to the client where this is running), or multi-parameter "complex" variables that use hard-coded names, like HullFuel or GameRule lookups in an expression.

default/python/AI/colonization/colony_score.py
```diff
@@ -1,15 +1,30 @@
...
 
 
 def debug_rating(*args, **kwargs):
     if DEBUG_COLONY_RATING:
         debug(*args, **kwargs)
+
+
+@cache_for_session
+def use_new_rating() -> bool:
+    config_value = eval(get_option_dict().get("use_new_colony_rating", "1"))
```
Cjkjvfnby: Eval sounds terrible.  Arbitrary Python code could be executed. 

I'd suggest to make check like that:  `get_option_dict().get("use_new_colony_rating", "false").lower() == 'true'`

default/python/AI/colonization/nest_rating.py
```diff
@@ -9,5 +9,5 @@ def special_is_nest(special_name: str) -> bool:
     return special_name in _nest_ratings
 
 
-def get_nest_rating(special_name: str, default: float) -> float:
-    return _nest_ratings.get(special_name, default)
+def get_nest_rating(special_name: str) -> float:
+    return _nest_ratings.get(special_name, 0)
```
Cjkjvfnby: This looks a bit more consistent, with return annotation
```suggestion
    return _nest_ratings.get(special_name, 0.0)
```

default/python/AI/colonization/rate_pilots.py
```diff
@@ -1,26 +1,120 @@
...
+}
+
+BAD_PILOT_RATING = 0.4
+GOOD_PILOT_RATING = 2.0
+GREAT_PILOT_RATING = 3.0
+ULT_PILOT_RATING = 4.0
+GOOD_SHIELD_BONUS = 0.3
+DETECTION_FACTOR = 0.08  # detection values -1 to 3
+FUEL_FACTOR = 0.1  # fuel values: -0.5 to 1.5
+# troop values range from 0.5 to 2 (or 3, but there is currently no species with ult. troops)
```
Cjkjvfnby: Reminds me some old joke: 

In the state of emergency value of cos could be 2, or 3 if required.

default/python/AI/colonization/rate_pilots.py
```diff
@@ -1,26 +1,120 @@
...
 
-@cache_for_session
 def rate_piloting_tag(species_name: str) -> float:
+    """
+    Wrapper while we have both rating functions.
+    This ensures survey_universe uses the same rating function like colonization.
+    """
+    if use_new_rating:
+        return rate_piloting(species_name)
+    return rate_piloting_old(species_name)
```
Cjkjvfnby: Although this approach is shorter, I prefer the full version with else.  It's much easy to track execution paths, because code and logic are aligned.   

```suggestion
    if use_new_rating:
        return rate_piloting(species_name)
    else:
        return rate_piloting_old(species_name)
```

default/python/AI/colonization/rate_pilots.py
```diff
@@ -1,26 +1,120 @@
...
+
+@cache_for_session
+def rate_piloting_old(species_name: str) -> float:
+    """
+    Almost the old function.
+    Does the shield bonus stuff here instead of only in survey_universe, though.
+    """
+    weapon_grade_tag = get_species_tag_grade(species_name, Tags.WEAPONS)
+    result = _pilot_tags_rating_old.get(weapon_grade_tag, 1.0)
+    if result == "SP_ACIREMA":
```
Cjkjvfnby:
```suggestion
    if species_name == "SP_ACIREMA":
```

default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -337,7 +337,7 @@ def get_species_tag_value(species_name: str, tag_type: AIDependencies.Tags) -> f
     if tag_type == AIDependencies.Tags.POPULATION:
         return AIDependencies.SPECIES_POPULATION_MODIFIER.get(grade, 1.0)
     if tag_type == AIDependencies.Tags.SUPPLY:
-        return AIDependencies.SPECIES_SUPPLY_MODIFIER.get(grade, 1.0)
+        return AIDependencies.SPECIES_SUPPLY_MODIFIER.get(grade, 1)
```
Cjkjvfnby, line 340: ?

Cjkjvfnby: Python float and int are pretty compatible.  But you could end up with a case when the expected float/float division will be the actual int/int division.

Cjkjvfnby (changes requested):

default/python/AI/colonization/colony_score.py
Grummel7: Yeah, you can break the AI by writing a wrong config file, but whoever can do that can also modify the python files, so what?

The original idea was that I could write something like `use_new_colony_rating=fo.empireID%2`.
Didn't work and the way I am using it now, I can just as well cast it to int and have all the flexibility I want.
I hope that is more acceptable.

default/python/AI/colonization/nest_rating.py
Grummel7: Well, then let's make the actual ratings floats as well.

default/python/AI/colonization/rate_pilots.py
Grummel7: Not sure what you are mean, except perhaps I should improve my comments.
Factors are multiplied by values to add up as ratings, the comments were intended to show how much these values will affect the overall ratings. Perhaps _scaling instead of _factor would be better, too.

default/python/AI/freeorion_tools/_freeorion_tools.py
Grummel7: Supply specifies number of jumps, perhaps its better to make sure it becomes an integer when it matters. I actually got an exception once when doing something like `for i in range(0, supply)`
Will revert that change.

default/python/AI/colonization/colony_score.py
Cjkjvfnby: > Yeah, you can break the AI by writing a wrong config file

You could execute arbitrary code with the wrong config using eval, like rm the home folder of the user.  This is a security vulnerability. 

I am ok with int.

default/python/AI/freeorion_tools/_freeorion_tools.py
Cjkjvfnby: So some tags are float and other ints?

I'd prefer to have a separate function instead of each switch case.  I'll do it in separate PR.

Cjkjvfnby (approved):

PR #3917, opened 2022-06-19
Fix log file detection for charting

Cjkjvfnby:


Grummel7 (approved):
The review would have been easier if you had described what problem you tried to fix.

Indeed there is no guarantee that the user interface log ends at the same time as the AI log, but the daemon log does. I guess replacing freeorion.log by freeoriond.log would solved it, too.

Apart from that the change looks fine to me.

PR #3918, opened 2022-06-20
Added helper functions for names values and stability modifier

Grummel7:
Small preparatory work for the planet rating as requested by @Cjkjvfnby


default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -349,3 +351,31 @@ def get_ship_part(part_name: str):
         warning("Could not find part %s" % part_name)
 
     return part_type
+
+
+def named_int(name: str) -> int:
```
Cjkjvfnby: Maybe `get_int_reference`?  It would be nice to include the verb to the function name.

default/python/AI/freeorion_tools/_freeorion_tools.py
```diff
@@ -349,3 +351,31 @@ def get_ship_part(part_name: str):
...
+    """
+    Returns a NamedReal from FOCS.
+    If the value does not exist, reports an error and returns 1.
+    Note that we do not raise and exception so that the AI can continue, as good as it can, with outdated information.
+    This is also why we return 1, returning 0 could cause followup errors if the value is used as divisor.
+    """
+    value = fo.getNamedValue(name)
+    if not isinstance(value, int):
```
Cjkjvfnby: Let's handle separately two cases:

1. Value not found (is in None?)
2. Value is found, but it is of the wrong type.

Cjkjvfnby (changes requested):
Left a couple of comments.

default/python/AI/freeorion_tools/_freeorion_tools.py
Grummel7: Well, my original idea was #3879, which had two different interface functions. But since nobody wants to merge this and I was fed up with see PyCharm complain about every use of a named value, I made this simple wrappers.

default/python/AI/freeorion_tools/_freeorion_tools.py
Grummel7: In FOCS they are called NamedReal and NamedInt, maybe get_named_int?

My long term idea is to mostly use them to initialise global variables in AIDependencies.py. So far the AI is started too early, however, I simply get None when I use them at global level.

Cjkjvfnby: `get_named_int`  looks good.

default/python/AI/freeorion_tools/_freeorion_tools.py
Cjkjvfnby:
```python
value = fo.getNamedValue(name)
if value is None:
    error(f"Requested NamedInt {name}, which doesn't exist!")
elif not isinstance(value, int):
    error(f"Requested value {name} of type int got {type(value)}!")
```

Grummel7: Well, if you insist on it...

Cjkjvfnby: These are different errors and different reactions from user are expected for them.

PR #3919, opened 2022-06-20
Added world tree and ast. coating to empire resources.

Grummel7:
More preparatory work for the planet rating

### DIFF
--- a/default/python/AI/AIDependencies.py
+++ b/default/python/AI/AIDependencies.py
@@ -219,9 +219,15 @@ for metab, boosts in metabolismBoostMap.items():
 # Please see the Note at top of this file regarding PlanetSize-Dependent-Lookup
 # Regardless of whether the sub-dictionary here has PlanetSize keys, the final
 # value will be applied as a *fixed-size mod* to the max population
+WORLDTREE_SPECIAL = "WORLDTREE_SPECIAL"
 POP_FIXED_MOD_SPECIALS = {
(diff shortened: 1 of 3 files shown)


default/python/AI/AIDependencies.py
```diff
@@ -219,10 +219,16 @@ class Tags:
...
 # Regardless of whether the sub-dictionary here has PlanetSize keys, the final
 # value will be applied as a *fixed-size mod* to the max population
+WORLDTREE_SPECIAL = "WORLDTREE_SPECIAL"
 POP_FIXED_MOD_SPECIALS = {
-    "DIM_RIFT_MASTER_SPECIAL": -4,
+    WORLDTREE_SPECIAL: 1,  # Not for SP_KHAKTURIAN...
 }
 
+
+def not_affect_by_special(special: str, species: str) -> bool:  # stupid special handlings...
```
Cjkjvfnby, line 228: This is not stupid, this is one of the nice ways to handle special cases: in the isolation from other code.

Cjkjvfnby (approved):

default/python/AI/AIDependencies.py
Grummel7: > This is not stupid, this is one of the nice ways to handle special cases: in the isolation from other code.

I was talking about the fact that those Khak... creatures are not affected by world trees.
And this function only partially handles it, you still have to make sure to use it in the right place(s).

PR #3923, opened 2022-06-20
Added within_n_jumps plus annotation

Grummel7:
another small preparation PR

### DIFF
--- a/default/python/AI/universe/system_network.py
+++ b/default/python/AI/universe/system_network.py
@@ -102,7 +102,7 @@ def within_n_jumps(system_id: SystemId, n: int) -> FrozenSet[SystemId]:
     if n < 1:
         return frozenset({system_id})
     elif n == 1:
-        return frozenset({system_id} & get_neighbors(system_id))
+        return frozenset({system_id} | get_neighbors(system_id))
     tier_minus_2 = within_n_jumps(system_id, n - 2)
     tier_minus_1 = within_n_jumps(system_id, n - 1)
     result = set(tier_minus_1)
--- a/default/python/AI/universe/system_network.py
+++ b/default/python/AI/universe/system_network.py
@@ -1,5 +1,5 @@
 import freeOrionAIInterface as fo
-from typing import Sequence, Set
+from typing import FrozenSet, Sequence, Set
 
 from AIDependencies import INVALID_ID
 from common.fo_typing import SystemId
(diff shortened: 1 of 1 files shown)


default/python/AI/universe/system_network.py
```diff
@@ -92,3 +92,14 @@ def systems_connected_to_system(system_id: SystemId) -> Set[SystemId]:
...
+
+
+@cache_for_current_turn
+def within_n_jumps(system_id: SystemId, n: int) -> FrozenSet[SystemId]:
+    if n < 1:
+        return frozenset({system_id})
+    inner_tier = within_n_jumps(system_id, n - 1)
+    result = set(inner_tier)
+    for sys_id in inner_tier:
+        result.update(get_neighbors(sys_id))
```
Cjkjvfnby, line 110: We will call `get_neighbors` for system_id n times.  For each neighbor n-1 times.   
We use memorization for `get_neighbors`, so it won't affect performance very much.

Some experiments:
If we see 11 systems around it will be 90 calls
When we see 78 it's will 354 calls.

Classical BFS without recursion with check of visited systems will be a bit faster. 

It still pretty fast and 10 should be extream number.

Grummel7: Well, at least duplicates can be avoided, it should be a little better now.

Cjkjvfnby: I mean something like that,

```python
def within_n_jumps(system_id: SystemId, n: int) -> FrozenSet[SystemId]:
    current_layer = [system_id]
    next_layer = []
    visited = {system_id}

    for _ in range(n):
        for sid in current_layer:
            neighbors = get_neighbors(sid)
            not_visited = neighbors - visited
            next_layer.extend(not_visited)
            visited.update(not_visited)
        current_layer = next_layer
        next_layer = []

    return frozenset(visited)
```

When I was preparing for a job interview, I spent a lot of time solving tasks like that.

Grummel7: Your version is indeed a little fast then mine, maybe 20%, but I had to create an extra big universe to measure that.

In fact when intermediate results are needed, my version is faster overall, thanks to caching. And I do need the results for different numbers when comparing supply effects of different species.

Cjkjvfnby: Lets get back to this a bit later, I\ll merge it as is now.

PR #3931, opened 2022-06-25
Add NamedValues to Nebula, Ion Storm and Molecular Cloud for Pedia

Oberlus:


Oberlus:
With this PR applied to current master (e3d8204), consulting the Pedia for first time after starting FreeOrion needs half a minute maybe (in a high performance rig). Without the PR (no recompilation needed), consulting Pedia is instantaneous.

geoffthemedio:
Check the client log. Probably there are a bunch (about 10 is the limit I think) errors about missing strings. If those are the issue, then usually after the limit is reached, searches should be faster as it stops logging the errors and has already populated the "ERROR: WHATEVER" string table.

o01eg:
@geoffthemedio Could it be made into unit test about missing strings?

geoffthemedio:
Probably, though the error log includes a stack trace that might be difficult to get into a unit test result...?

Oberlus:
The problem is in the NamedRealLookup in two of the three FLD_NEBULA_*.
The NamedReal in the first nebula file gives no trouble, the other two create errors for a while in the log until resolved.
I'll try declaring the value in named_values.focs.txt and use the lookup in the three, or merging the three nebulae in a single FOCS file. ~~Not today.~~

Oberlus:
Declaring the values in named_values.focs and using lookups in the three nebula focs files did the trick.

Ready for merge.

default/scripting/fields/FLD_NEBULA_1.focs.txt
```diff
@@ -51,9 +51,12 @@ FieldType
             ]
             stackinggroup = "NEBULA_SHIP_EFFECTS"
             effects = [
-                SetSpeed value = max(5, Value - 20)
-                SetStealth value = Value + 20
-                SetDetection value = Value - 20
+                //SetSpeed value = max(5, Value - (NamedReal name = "FLD_NEBULA_SPEED_MALUS" value = 20))
+                //SetStealth value = Value + NamedReal name = "FLD_NEBULA_STEALTH_BONUS" value = 20
+                //SetDetection value = Value - NamedReal name = "FLD_NEBULA_DETECTION_MALUS" value = 20
```
geoffthemedio:
```suggestion

```

Oberlus: Ouch... I forgot to remove the commented code.

agrrr3:
> The problem is in the NamedRealLookup in two of the three FLD_NEBULA_*. The NamedReal in the first nebula file gives no trouble, the other two create errors for a while in the log until resolved. 

@oberlus was that the message that it is waiting for a value ref not yet registered? and how did you open the pedia?

@o01eg is the python parsing running unconditionally (or is it lazy)?

Oberlus:
> was that the message that it is waiting for a value ref not yet registered?

Yes

> and how did you open the pedia?

I started FO and clicked on Pedia. Was OK.
Then I clicked on FieldType (I was testing this PR) and FO freezed for some 30 seconds.

I guess if I wait a minute after starting FO before using Pedia, there won't be any delay.

o01eg:
> @o01eg is the python parsing running unconditionally (or is it lazy)?

Python executes FOCS files uncoditionally.

PR #3939, opened 2022-07-01
calculate research and stability, tech_soon_available

Grummel7:
This should be the last pre-PR, hopefully this weekend I can finish a first version of the rating stuff.

### DIFF
--- a/default/python/AI/colonization/calculate_production.py
+++ b/default/python/AI/colonization/calculate_production.py
@@ -1,16 +1,16 @@
 import freeOrionAIInterface as fo
-from typing import Iterable, NamedTuple
+from logging import debug
+from typing import List, NamedTuple
 
 import AIDependencies
 from buildings import BuildingType
-from freeorion_tools import get_named_real, get_species_tag_value, tech_is_complete
+from colonization.colony_score import DEBUG_COLONY_RATING
+from freeorion_tools import get_named_real, get_species_tag_value, tech_soon_available
 from freeorion_tools.caching import cache_for_current_turn
 from turn_state import have_honeycomb
 
 
-def calculate_production(
-    planet: fo.planet, species: fo.species, max_population: float, stability: float, research_list: Iterable
-) -> float:
+def calculate_production(planet: fo.planet, species: fo.species, max_population: float, stability: float) -> float:
     """
     Calculate how much PP the planet's population could generate with industry focus.
     This only considers values that actually rely on industry focus, those that do not are handled by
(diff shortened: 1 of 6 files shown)


default/python/AI/colonization/calculate_research.py
```diff
@@ -0,0 +1,213 @@
...
+    skill_multiplier = get_species_tag_value(species.name, AIDependencies.Tags.RESEARCH)
+    bonus_by_policy = _get_research_bonus_modified_by_policy(stability)
+    flat_by_policy = _get_research_flat_modified_by_policy(planet, stability)
+    policy_multiplier = _get_policy_multiplier(stability)
+    bonus_unmodified = _get_research_bonus_unmodified(planet, stability)
+
+    result = max_population * (AIDependencies.INDUSTRY_PER_POP + bonus_modified) * skill_multiplier
+    result = (result + max_population * bonus_by_policy + flat_by_policy) * policy_multiplier
+    result += max_population * bonus_unmodified
+    if DEBUG_COLONY_RATING:
```
Cjkjvfnby: Maybe move to function debug_rating?

default/python/AI/colonization/calculate_research.py
```diff
@@ -0,0 +1,213 @@
...
+
+def _get_research_bonus_modified(planet: fo.planet, stability: float) -> float:
+    """
+    Calculate bonus research per population which would be added before multiplication with the species skill value.
+    """
+    result = 0.0
+    for bonus in _get_modified_research_bonuses(planet):
+        if bonus.available and stability >= bonus.min_stability:
+            result += bonus.value
+    return result
```
Cjkjvfnby: There is an interesting approach to designing objects responsible for data. 
Instead of doing logic outside the object, we can do the logic inside it. 

Making _ResearchBonus to work in that way  will make this code could looks like that:
```suggestion
    return sum(bonus.get_bonus(stability) for for bonus in _get_modified_research_bonuses(planet))
```

default/python/AI/colonization/calculate_production.py
```diff
@@ -47,47 +54,45 @@ def _get_modified_industry_bonuses():
     centre_bonus1 = get_named_real("BLD_INDUSTRY_CENTER_1_TARGET_INDUSTRY_PERPOP")
     centre_bonus2 = get_named_real("BLD_INDUSTRY_CENTER_2_TARGET_INDUSTRY_PERPOP")
     centre_bonus3 = get_named_real("BLD_INDUSTRY_CENTER_3_TARGET_INDUSTRY_PERPOP")
-    # use fo.getEmpire().researchQueue directly here to simplify caching this function
-    top3_research = [fo.getEmpire().researchQueue[i] for i in range(0, 3)]
     return [
         _ProductionBonus(
-            tech_is_complete("PRO_FUSION_GEN") or "PRO_FUSION_GEN" in top3_research,
+            tech_soon_available("PRO_FUSION_GEN", 3),
```
Cjkjvfnby, line 60: This is a nice change. 

Also, this is a good example of why having small focused PR is great. 

Having it in a separate PR with a comment, that this is refactoring and small would make it merge faster.  
I always could spend a minute on small reviews when I browse my mail.

In general, it's ok to have some cleanup together with features, but I want to build the process when it happens very rarely. 

Backporting changes from the development branch with git and Pycharm is not time-consuming. Maybe we should meet and I'll screen share some tricks and tips related to that process.

Cjkjvfnby (commented):
New changes come during the review process, will check it next time.

default/python/AI/colonization/calculate_research.py
Grummel7: That's definitely better

default/python/AI/colonization/calculate_research.py
Grummel7: why not...

Cjkjvfnby (approved):

default/python/AI/colonization/calculate_production.py
```diff
@@ -142,37 +141,33 @@ def _get_production_bonus_unmodified(planet: fo.planet, stability: float):
             0.0,
             get_named_real("HONEYCOMB_TARGET_INDUSTRY_PERPOP"),
         ),
+        # TBD: Collective Thought Network? Ignored for now, AI doesn't know how to handle it.
+        # It shouldn't make a big difference for selecting which planets to colonise anyway.
     ]
-    for bonus in bonuses:
-        if bonus.available and stability >= bonus.min_stability:
```
Cjkjvfnby, line 147: @Grummel7 
`bonus available ` is no longer taken into account.

I found it during small refactoring.
https://github.com/freeorion/freeorion/pull/3942

PR #3940, opened 2022-07-01
[AI] Impove chart view

Cjkjvfnby:
A bunch of improvements

- Graphs could be zoomable/dragable.
- Make graph more curly (this is side effect, don't know hot to fix it)
- You can select on empire by clicking on its legend
- Increase time for log span

Zoomed graph
изображение
One empire selected
изображение

### DIFF
--- a/default/python/charting/file_reader.py
+++ b/default/python/charting/file_reader.py
@@ -25,7 +25,7 @@ def _validate_data_dir(data_dir: Path) -> bool:
     return os.path.exists(data_dir / "freeorion.log")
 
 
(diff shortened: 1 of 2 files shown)


Grummel7 (approved):
Looks good.

Some comments:
- Actually the best way to determine which log files belong to the last game would be to parse the first line which contains a time stamp of when the AI has been started.
-  For longer games it may be useful to have an option of displaying wide graphs and use a scroll bar as before. Perhaps also in full screen mode.
- Also for longer games where numbers get very high, a logarithmic option could be added.

PR #3942, opened 2022-07-03
[AI] Refactor bonus calculation, fix check

Cjkjvfnby:
I have joined duplicated code together. 

While writting the tests I found and fix the bug in the `get_bonus`.


default/python/tests/AI/bonus_calculator_tests.py
```diff
@@ -0,0 +1,17 @@
...
+        (Bonus(True, 1, 5.5), 0),
+        (Bonus(True, 2, 5.5), 0),
+        (Bonus(True, 0, 5.5), 5.5),
+        (Bonus(False, 0, 5.5), 0),
+        (Bonus(False, 1, 5.5), 0),
+        (Bonus(False, 2, 5.5), 0),
+    ),
+)
+def test_get_bonus_for_stability_1(bonus, expected):
+    assert bonus.get_bonus(1) == expected
```
Grummel7:
```suggestion
    ("bonus", "stability", "expected"),
    (
        (Bonus(True, 1, 5.5), 1, 5.5),
        (Bonus(True, 2, 5.5), 1.9, 0),
        (Bonus(True, 0, 5.5), 0.1, 5.5),
        (Bonus(False, 0, 5.5), 0, 0),
        (Bonus(False, 1, 5.5), 0, 0),
        (Bonus(False, 2, 5.5), 0, 0),
    ),
)
def test_get_bonus_for_stability_1(bonus, stability, expected):
    assert bonus.get_bonus(stability) == expected
```
When I looked at the table I wondered how you got to the value until I saw the test hard-coded stability. Looks more natural for me that way.

Grummel7 (changes requested):
Thanks, that is better...
But I added another bug: stability must be compared with >=, not

default/python/AI/freeorion_tools/bonus_calculation.py
```diff
@@ -0,0 +1,12 @@
...
+
+class Bonus(NamedTuple):
+    available: bool
+    min_stability: float
+    value: float
+
+    def get_bonus(self, stability: float) -> float:
+        if not self.available:
+            return 0.0
+        return self.value if stability > self.min_stability else 0.0
```
Grummel7:
```suggestion
        if not self.available or stability < self.min_stability:
            return 0.0
        return self.value
```

default/python/tests/AI/bonus_calculator_tests.py
Cjkjvfnby: The problem with parameterized tests is when you get too many arguments, you will need to look into the code anyway.   So I tried to reduce the number of them, hardcoding some in the test and the test name.

default/python/AI/freeorion_tools/bonus_calculation.py
Cjkjvfnby: There is a simple rule, expressions with `not` are more challenging for humans to read.

 `not self.available or stability < self.min_stability` -> `self.available or stability >= self.min_stability`

Cjkjvfnby:
@Grummel7 

- rename attribute
- simplify bonus usage
- update tests

Grummel7 (approved):

PR #3946, opened 2022-07-05
Tech short desc

Grummel7:
Some polishing, many tech short descriptions were outdated, especially some said theoretical prerequisite, also the tech did unlock something now.

This is not complete and I am not even sure its the right thing to do. There are lots of technologies that do something and also unlock a policy.
Does it make sense to have a short description like this?
`Increases Starlane Speed & Unlocks Policy`

### DIFF
--- a/default/scripting/techs/construction/ARCH_PSYCH.focs.py
+++ b/default/scripting/techs/construction/ARCH_PSYCH.focs.py
@@ -3,7 +3,7 @@ from common.base_prod import TECH_COST_MULTIPLIER
 Tech(
     name="CON_ARCH_PSYCH",
     description="CON_ARCH_PSYCH_DESC",
-    short_description="THEORY_SHORT_DESC",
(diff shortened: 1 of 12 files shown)


geoffthemedio:
If the policy primarily also increases speed, then it doesn't need to separate mention the unlocking.

But in general, I think the "main" effect can be mentioned in the short description, even if there are other effects omitted.

Grummel7:
> If the policy primarily also increases speed, then it doesn't need to separate mention the unlocking.
> 
> But in general, I think the "main" effect can be mentioned in the short description, even if there are other effects omitted.

Well, maybe we can just keep the rest as it is. With this PR:
- There should be no more "Theoretical Prerequisites" that actually give something
- All Techs that give a policy slot have it in their short description.

geoffthemedio:
needs rebase due to tech moved to python. otherwise seems OK

Grummel7:
@geoffthemedio Can you merge this now? If you keep waiting, Oleg will move more techs and I'll have to rebase again...

PR #3947, opened 2022-07-06
Meter serialize

geoffthemedio:
Customized and hopefully faster serialization of UniverseObject Meters.

Some testing reports from others would be good before merging this...


o01eg (approved):
Looks fine. Maybe we could use integer as fixed-point float meter further to speed up.

Vezzra (approved):
Builds and runs fine on macOS.

PR #3949, opened 2022-07-06
[AI] Break dependency of buildings on empire

Cjkjvfnby:
Building is low level concept and it should depend only on the fo and tools.


Grummel7 (approved):
You once said that in the end, you want only BuildingType itself use value, that's why I've introduced that function. I guess you are right, we better use type.value for a while then to add more dependencies like this to building.py.

I think in the long run, empire should include building and get_best_pilot_facilities should take an argument of type Shipyard.

PR #3950, opened 2022-07-06
Add unique name for binary artefact

Cjkjvfnby:
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

> GITHUB_RUN_NUMBER | A unique  number for each run of a particular workflow in a repository. This  number begins at 1 for the workflow's first run, and increments with  each new run. This number does not change if you re-run the workflow  run. For example, 3.

I choose this variable because it's short, unique, and ordered.   We could implement this with date, branch name and short SHA, but this wont be short and will be much more complicated.


o01eg (approved):
Fine.

